PR #3817, opened 2020-05-09
[secure-storage] GitHub Wrapper

davidiw:
This introduces GitHub as another solution to secure-storage and
identifies several other interesting bits along the way:

* GitHub inserts newlines every 60 characters in their base64 encodings -- I don't understand but it means we need to remove them before decoding.
* In order to recursively delete everything, we add a delete_directory and also annotate directories in the "get_directory" query with an extra "/" -- which is exactly how vault does it
* Added a proper liveness check via branch checking
* The actual GitHubStorage is actually rather trivial -- but all tests pass
* Removed a duplicate test on key generation, rotation, exporting

This builds on previous PR: https://github.com/libra/libra/pull/3812

### DIFF
--- a/secure/storage/github/src/lib.rs
+++ b/secure/storage/github/src/lib.rs
@@ -67,8 +67,14 @@ impl Client {
     }
 
     /// Delete a file from a GitHub repository
-    pub fn delete(&self, path: &str) -> Result<(), Error> {
-        let hash = self.get_sha(path)?;
+    pub fn delete_file(&self, path: &str) -> Result<(), Error> {
+        // Occasionally GitHub sends us back delayed results and the file is already deleted.
+        let hash = match self.get_sha(path) {
+            Ok(hash) => hash,
+            Err(Error::NotFound(_)) => return Ok(()),
+            Err(e) => return Err(e),
+        };
+
         let resp = self
             .upgrade_request(ureq::delete(&self.url(path)))
             .send_json(json!({ "message": "libra-secure", "sha": hash }));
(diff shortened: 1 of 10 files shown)


secure/storage/src/github.rs
```diff
@@ -0,0 +1,53 @@
+// Copyright (c) The Libra Core Contributors
+// SPDX-License-Identifier: Apache-2.0
+
+use crate::{CryptoKVStorage, Error, GetResponse, KVStorage, Value};
+use libra_github_client::Client;
+use libra_secure_time::{RealTimeService, TimeService};
+
+/// GitHubStorage leverages a GitHub repository to provide a file system approach to key / value
+/// storage.  This is not intended for storign private data but for organizing public data.
```
JoshLind, line 9: nit: storign -> storing

secure/storage/github/src/lib.rs
```diff
@@ -232,8 +279,8 @@ mod tests {
         let value = std::str::from_utf8(&b64_value).unwrap();
         assert_eq!(value, value2);
 
-        // Verify delete
-        github.delete(path).unwrap();
+        // Verify delete_file
```
zekun000, line 282: does the OWNER, REPOSITORY actually work?

secure/storage/src/tests/suite.rs
```diff
@@ -185,43 +184,6 @@ fn test_create_and_get_non_existent_version(storage: &mut dyn Storage) {
...
-/// are correctly rotated in storage.
-fn test_create_rotate_and_check_key_pair(storage: &mut dyn Storage) {
-    // Create new key pair, fetch both public and private keys and verify relationship
-    let public_key = storage
-        .create_key(CRYPTO_NAME, &Policy::public())
-        .expect("Failed to create a test Ed25519 key pair!");
-    let private_key = storage
-        .export_private_key(CRYPTO_NAME)
-        .expect("Failed to get the private key for a key pair that should exist!");
-    assert_eq!(private_key.public_key(), public_key);
```
JoshLind, line 257: nit: if we're going to remove this test, can we add these relationship checks to the test below? i.e., ensure that the public and private keys for a specific key have a relationship? "assert_eq!(private_key.public_key(), public_key);"

secure/storage/github/src/lib.rs
```diff
@@ -79,13 +85,47 @@ impl Client {
...
+                continue;
+            }
+            self.delete_file(&file)?;
+        }
+        Ok(())
+    }
+
+    /// Retrieve a list of branches, this is effectively a status check on the repository
+    pub fn get_branches(&self) -> Result<Vec<String>, Error> {
+        let url = format!("{}/repos/{}/{}/branches", URL, self.owner, self.repository);
```
zekun000, line 103: this "URL/repose/owner/repository" pattern probably should be abstracted

secure/storage/github/src/lib.rs
davidiw: Hahaha, you can see the magic: https://github.com/davidiw/secure-store-test

secure/storage/github/src/lib.rs
davidiw: so then we would potentially call format twice :O

I already have one for general purpose accesses.

Issue #3679, opened 2020-04-30
[Move] Reconsider `Transaction::sender()` and `move_to_sender`


sblackshear:
Take a look at RFC PR https://github.com/libra/libra/pull/3673 which:
- Introduces the `Sender::T` resource, which is just a wrapper around an account address. It exposes a function that allows reading the inner address. You can think of `Sender::T` as the "sender capability"; it is the permission to know the sender's address + take actions that require  knowing this address.
- Eliminates every use of `Transaction::sender()` in the core modules by using a functional discipline: thread an explicit `&Sender` parameter everywhere that needs to read the sender. This is similar to `ViewerContext` (and related design patterns) in a codebase that is probably familiar to folks here.
- Similarly, adds a call to `Sender::move_to(&Sender::T)` (a newly introduced no-op function) before every use of `move_to_sender`. This doesn't actually do anything for now, but gives an idea of how the code would look if we replaced the `move_to_sender<R>(R)` bytecode with `move_to<R>(R, &Sender)`. 

Now, the obvious next question is: how do I get a `Sender::T` resource given that the 	`Sender` module exposes no API for creating one? There are a few possible answers:

1. Add a `sender: Option<Sender::T>` field to the `LibraAccount::T` struct that will be instantiated with `Sender { addr }` upon a call to `LibraAccount::create(addr)`. Then, expose `extract` and `fill` APIs for the `Option` that allow the account owner to take their `Sender` resource out/put it back. This does not require any changes to the language or the Libra/Move adapter.
2. Have the Libra/Move adapter create a `Sender { addr }` resource (where `addr` is the transaction sender) in Rust, then pass it in as a parameter of the transaction script (e.g., `main(sender: &Sender::T)` or `main(sender: Sender::T): Sender::T`). This requires a small change to the Libra/Move adapter and the bytecode verifier for transaction script signatures (for the `sender` param).

## Ways forward

I am not suggesting that we commit this change as-is; it was more of an experiment to understand whether this sort of programming discipline would be reasonable. Having completed it, I feel that the result is reasonable to write, easier to read, and has some significant benefits (more on that in the next section). There are several paths we could consider from here:

I. Decide we don't like this and leave everything as-is.   
II. Decide we like this as a design pattern, but want to keep both `Transaction::sender()` and `move_to_sender`.  
[...]

PR #3819, opened 2020-05-10
[move core modules] simplifying scripts and stdlib with `move_to<T>(&signer, T)`

sblackshear:
This is a proof-of-concept PR to show some of the simplifications that become possible if we add a `move_to<T>(&sender, T)` instruction (as explained in https://github.com/libra/libra/issues/3679).

For a high-level summary of the simplifications:
- Each of the different account types can be created/instantiated in a single transaction instead of  today's three-tx create/apply/approve process. This is an oft-requested feature from custody, and should make life easier for AOS too.
- This cuts 35% (14) of the user-facing scripts in `transaction_scripts.rs`. "Number of tx scripts" is the the best single number metric for client complexity (IMO).
- The stdlib is 11% shorter (-350 lines) after the changes in this PR, and I suspect it would be possible to cut another hundred or so using the simplifications below.

Here's a more fine-grained summary of what the PR does:
- Single function to create each of RootVASPs, ChildVASPs, and Association accounts
- This allows us to eliminate the `AccountType` module and significantly simplify the `VASP`, `Association`, `AccountLimits`, `LibraConfig`, and `Genesis` modules
- The `AccountType` and `AccountTracking` modules are no longer needed

Further simplifications are possible, but I have not attempted them in this PR:

- We can likely simplify `LibraConfig` to get rid of the two-step create/claim pattern there
- We can add a flag to create_account that lets the creator choose to pre-populate the account with all currency types (very useful for custody). Without this, setting up an account is actually a  3 + N step process (where N is the number of currencies)


dariorussi:
I think we should talk a bit more about this because I am confused a bit.
Here we have a `create_sender()`, `destroy_sender()` private to a module that have infinite power (so to speak ;)). That seems really odd.
I am not sure how we are thinking to materialize that in the future. I remember conversations around VM native type and bytecodes and I am not clear what kind of privileges we are thinking of, etc.
The code seems fine though it is removing a ton of stuff which is somewhat surprising...

language/stdlib/modules/libra_account.move
```diff
@@ -636,6 +746,14 @@ module LibraAccount {
...
             let transaction_fee_balance = borrow_global_mut<Balance<Token>>(0xFEE);
             Libra::deposit(&mut transaction_fee_balance.coin, transaction_fee);
+            Transaction::assert(
+                AccountLimits::update_deposit_limits<Token>(
+                    transaction_fee_amount,
+                    0xFEE,
+                    &borrow_global<AccountOperationsCapability>(0xA550C18).limits_cap
+                ),
+                9
+            );
```
tzakian, line 756: I don't believe this code is there anymore so can be taken away probably.

language/stdlib/modules/vasp.move
```diff
@@ -133,225 +74,54 @@ module VASP {
...
-    // Return whether `root_vasp_addr` and `child_address` have a valid
-    // root-child VASP relationship.
-    public fun is_root_child_vasp(root_vasp_addr: address, child_address: address): bool {
-        if (is_root_vasp(root_vasp_addr) && is_child_vasp(child_address)) {
-            AccountType::root_address<ChildVASP>(child_address) == root_vasp_addr
-        } else {
-            false
-        }
+    // Create a child VASP resource for the sender
+    // Aborts if the sender is not a RootVASP
```
tzakian, line 82: So we no longer have the concept of a "parent account" for a VASP with this. Just want to confirm that that's OK.

language/stdlib/modules/libra_account.move
```diff
@@ -409,70 +458,131 @@ module LibraAccount {
...
+    /// Create an account with the ParentVASP role at `new_account_address` with authentication key
+    /// `auth_key_prefix` | `new_account_address`.
+    public fun create_parent_vasp_account<Token>(
+        new_account_address: address,
+        auth_key_prefix: vector<u8>,
+        human_name: vector<u8>,
+        base_url: vector<u8>,
+        compliance_public_key: vector<u8>,
+    ) {
+        Association::assert_sender_is_association();
```
tzakian, line 558: We can do this in a different PR, but do we want this to be the root account? Or some specific privilege?

sblackshear: Good question and I'm not sure (probably the TC account?). Will add a TODO

PR #3827, opened 2020-05-11
[mempool] gc by block timestamp for successful txn commits

christinacelee:
## Motivation

Mempool wasn't gc'ing by block timestamp. Make sure mempool actually gc's txns by block timestamp (both with successful and failed txns)


davidiw:
Closing this out due to lack of traction, feel free to re-open.

PR #3831, opened 2020-05-11
[storage] enable "libradb/fuzzing" feature for using `LibraDB::new_for_test()`

msmouse:
## Motivation

`LibraDB::new_for_test()` is guarded by cfg(any(test, feature="fuzzing)), we need to enable the fuzzing feature when used in tests of packages other than libradb itself.

It works before this fix for `cargo xtest` because it has --all-features enabled, but not for individual packages like `cargo xtest -p executor`.

yes

## Test Plan
existing coverage

## Related PRs

### DIFF
--- a/execution/execution-correctness/Cargo.toml
+++ b/execution/execution-correctness/Cargo.toml
@@ -28,7 +28,7 @@ thiserror = "1.0"
 [dev-dependencies]
 config-builder = { path = "../../config/config-builder", version = "0.1.0" }
 executor-utils = { path = "../executor-utils", version = "0.1.0" }
-libradb = { path = "../../storage/libradb", version = "0.1.0"}
+libradb = { path = "../../storage/libradb", version = "0.1.0", features = ["fuzzing"] }
 storage-interface = { path = "../../storage/storage-interface", version = "0.1.0" }
 storage-service = { path = "../../storage/storage-service", version = "0.1.0" }
 tempfile = "3.1.0"
--- a/execution/executor/Cargo.toml
+++ b/execution/executor/Cargo.toml
@@ -38,7 +38,7 @@ config-builder = { path = "../../config/config-builder", version = "0.1.0" }
 executor-utils = { path = "../executor-utils", version = "0.1.0" }
 libra-config = { path = "../../config", version = "0.1.0" }
 libra-temppath = { path = "../../common/temppath", version = "0.1.0" }
-libradb = { path = "../../storage/libradb", version = "0.1.0" }
+libradb = { path = "../../storage/libradb", version = "0.1.0", features = ["fuzzing"] }
 stdlib = { path = "../../language/stdlib", version = "0.1.0" }
 transaction-builder = { path = "../../language/transaction-builder", version = "0.1.0" }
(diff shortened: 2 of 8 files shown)


execution/execution-correctness/Cargo.toml
```diff
@@ -28,7 +28,7 @@ thiserror = "1.0"
 [dev-dependencies]
 config-builder = { path = "../../config/config-builder", version = "0.1.0" }
 executor-utils = { path = "../executor-utils", version = "0.1.0" }
-libradb = { path = "../../storage/libradb", version = "0.1.0"}
+libradb = { path = "../../storage/libradb", version = "0.1.0", features = ["fuzzing"] }
```
wqfish, line 31: I suspect these changes may need to go to the `[features]` section below instead. I'm not sure.

execution/executor/Cargo.toml
```diff
@@ -38,7 +38,7 @@ config-builder = { path = "../../config/config-builder", version = "0.1.0" }
 executor-utils = { path = "../executor-utils", version = "0.1.0" }
 libra-config = { path = "../../config", version = "0.1.0" }
 libra-temppath = { path = "../../common/temppath", version = "0.1.0" }
-libradb = { path = "../../storage/libradb", version = "0.1.0" }
+libradb = { path = "../../storage/libradb", version = "0.1.0", features = ["fuzzing"] }
```
lightmark, line 41: Would this be added to line 48 instead? I think you are enabling fuzzing by default here?

execution/execution-correctness/Cargo.toml
lightmark: I think so, too.

msmouse: You do that when libradb is in the main "[dependency]" section

execution/executor/Cargo.toml
msmouse: This is the dev-dependence, I AM enabling fuzzing by default (to get the hidden LibraDB::new_for_test).

adding it to [features] requires LibraDB to be in the main [dependencies] (although it can be "optinoal = true".

execution/execution-correctness/Cargo.toml
davidiw: Actually I'm not sure the guidance is very clear in this space.

The reality is if you specify this under dev-dependencies you could accidentally enable the feature via transitive dependencies. So despite it seeming somewhat unclear.. I think the suggestions are correct.

cc: @bmwill

msmouse: "The reality is if you specify this under dev-dependencies you could accidentally enable the feature via transitive dependencies." elaboration? @davidiw

rexhoffman: @sunshowers is likely to make a lint at some point soon, so not sure accidental enablement will be a concern, but not there yet.  Unfortunately they are out this week as is @bmwill, so @metajack thoughts?

metajack: This is fine because it does not also appear in [dependencies], which means it won't get merged together. dev-dependencies are not considered except for the current crate, so it won't get merged into another crate either.

PR #3744, opened 2020-05-05
[LEC] Create ExecutionCorrectness service

lightmark:
## Motivation
2 commits:

1 . We need to move the traits to executor_types to tidy up the dependencies.
Also, a customized `Error` type is necessary for sending results over wire for the future use of this trait in execution correctness service.

2. Create the ExecutionCorrectness Service (LEC service) similar to safety-rules. No asymmetric cryptography involved yet.

### Have you read the [Contributing Guidelines on pull requests](https://github.com/libra/libra/blob/master/CONTRIBUTING.md#pull-requests)?

Y

## Test Plan

CI

PR #3832, opened 2020-05-11
[consensus] let consensus call LEC

lightmark:
## Motivation

According to #3744 , consensus has to talk to LEC directly instead of building a block_executor.

Y

## Test Plan

CI

### DIFF
--- a/config/src/config/execution_config.rs
+++ b/config/src/config/execution_config.rs
@@ -45,7 +45,7 @@ impl Default for ExecutionConfig {
         ExecutionConfig {
             genesis: None,
             genesis_file_location: PathBuf::new(),
-            service: ExecutionCorrectnessService::Local,
+            service: ExecutionCorrectnessService::Thread,
         }
     }
 }
--- a/consensus/Cargo.toml
+++ b/consensus/Cargo.toml
@@ -31,6 +31,7 @@ channel = { path = "../common/channel", version = "0.1.0" }
 consensus-types = { path = "consensus-types", version = "0.1.0", default-features = false }
 crash-handler = { path = "../common/crash-handler", version = "0.1.0" }
 debug-interface = { path = "../common/debug-interface", version = "0.1.0" }
+execution-correctness = { path = "../execution/execution-correctness", version = "0.1.0" }
 executor = { path = "../execution/executor", version = "0.1.0" }
 executor-types = { path = "../execution/executor-types", version = "0.1.0" }
 lcs = { path = "../common/lcs", version = "0.1.0", package = "libra-canonical-serialization" }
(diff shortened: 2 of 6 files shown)


consensus/Cargo.toml
```diff
@@ -46,6 +47,7 @@ libra-types = { path = "../types", version = "0.1.0" }
 libra-workspace-hack = { path = "../common/workspace-hack", version = "0.1.0" }
 network = { path = "../network", version = "0.1.0" }
 safety-rules = { path = "safety-rules", version = "0.1.0" }
+simple-storage-client = { path = "../storage/simple-storage-client", version = "0.1.0" }
```
zekun000, line 50: where do you use this?

consensus/src/state_computer.rs
```diff
@@ -21,14 +22,17 @@ use std::{
 /// Basic communication with the Execution module;
 /// implements StateComputer traits.
 pub struct ExecutionProxy {
-    executor: Mutex<Box<dyn BlockExecutor>>,
+    lec: Mutex<Box<dyn BlockExecutor + Send + Sync>>,
```
zekun000, line 25: this is a worse name

consensus/src/consensus_provider.rs
```diff
@@ -42,7 +41,11 @@ pub fn start_consensus(
         .expect("Failed to create Tokio runtime!");
     let storage = Arc::new(StorageWriteProxy::new(node_config, libra_db));
     let txn_manager = Box::new(MempoolProxy::new(consensus_to_mempool_sender));
-    let state_computer = Arc::new(ExecutionProxy::new(executor, state_sync_client));
+    let execution_correctness_manager = ExecutionCorrectnessManager::new(node_config);
```
davidiw, line 44: the copypasta is strong with this one.

config/src/config/execution_config.rs
```diff
@@ -45,7 +45,7 @@ impl Default for ExecutionConfig {
         ExecutionConfig {
             genesis: None,
             genesis_file_location: PathBuf::new(),
-            service: ExecutionCorrectnessService::Local,
+            service: ExecutionCorrectnessService::Thread,
```
davidiw, line 48: I will admit I'm a little reserved about this one.... I think flipping it now to evaluate is okay, but we should revisit where we are in a few weeks.

lightmark: What's wrong with this than local? At least this one includes ser/de and I think `safety_rules` uses it too.

consensus/Cargo.toml
lightmark: I think I forget to remove it

PR #3668, opened 2020-04-30
[crypto] Gates batch verification under a feature

huitseeker:
This allows us to (de)-activate features relying on ed25519's "batch" if needed.

Workaround for #3567.

cc @n4ss 
review @kchalkias

PR #3671, opened 2020-04-30
[ci] Cross compile transaction-builder for powerpc

metajack:
This is need to ensure libra-types, libra-crypto, and lcs compile on powerpc
for some upstream users.

This should be ready to land as soon as CI is green.

PR #3837, opened 2020-05-12
[crypto] Load the batch verification feature, but only in little endian …

huitseeker:
…contexts

ed25519 becomes a target-specific dependency of crypto/crypto.
Context: this can be seen as a better version of  https://github.com/libra/libra/pull/3668
tested with `cargo --target powerpc-unkown-linux-gnu` 

I would appreciate a quick look from @metajack  on this one for the conditional dependency.

### DIFF
--- a/crypto/crypto/Cargo.toml
+++ b/crypto/crypto/Cargo.toml
@@ -14,7 +14,6 @@ anyhow = "1.0"
 bytes = "0.5.4"
 curve25519-dalek = { git = "https://github.com/calibra/curve25519-dalek.git", branch = "fiat2", default-features = false }
 digest = "0.8.1"
-ed25519-dalek = { git = "https://github.com/calibra/ed25519-dalek.git", branch = "fiat2", features = ["serde"], default-features = false }
 hex = "0.4.2"
 hmac = "0.7.1"
 once_cell = "1.3.1"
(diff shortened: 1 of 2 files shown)


huitseeker:
> Unfortunately cargo will unify these features sets together unless you are using nightly cargo and opting into the new resolver (-Z features=all).

I see. Or we could opt into [`-Zfeatures=itarget`](https://github.com/rust-lang/cargo/issues/7914) right?

> We'll need to remove the batch feature completely from Cargo.toml and then pass it explicitly when building for powerpc with --features batch or something.

That's done already — nobody loads batch any more, see my comment on #3671

> Or I can prioritize finishing the new resolver PR and we can switch resolvers. This would only break stable cargo in the case of trying to build for big endian platforms.

I'm afraid breaking a big endian build is not an option for the moment.

metajack:
> I'm afraid breaking a big endian build is not an option for the moment.

I think you might have misunderstood me. If we switch to the new cargo resolver, our CI would always use that to build. But a random developer cloning libra/libra would probably build with the old resolver if they failed to do any extra setup to mix nightly cargo with stable rustc. In that case, libra would still build, except on powerpc. Powerpc in generally would still always be buildable, just not with stable cargo (until -Z features stabilizes).

huitseeker:
@metajack I see. Closing until one of these stabilizes, then (or, failing that, until we fix the issue upstream).

PR #3840, opened 2020-05-12
Consolidated protos

agclement:
## Motivation

This _draft_ PR is a first step towards addressing the proto contamination of types/admission-control/storage and is intended to address a problem raised by Hui:

> Goal: Move all proto code out of types and storage  and isolate it in admission-control. This allows us to isolate our gRPC service, much like we did with our json service, without infecting the rest of the code.

The PR, in its current state, is not appropriate for submission. It is put forth as a draft to gather feedback and to ascertain if the strategy aligns with the intent and to get guidance from the team for a number of high level aesthetic questions. 

At a high level, this PR does the following four things:

1.  Create a _libra-protos/_ directory 
1.  Within _libra-protos/_, create separate crates for protos associated with _admission-control/_, _types/_, and _storage/_.   These crates are separated because the first two contain rpc services requiring tonic in addition to Prost.  Additionally, _libra-protos/_ contains *only* the proto files.  The associated _From_ and _TryFrom_ functionality and non-proto struct definitions remain in the original locations.
1.  Consolidate the _TryFrom, From_ functions that were previously a part of _storage-proto_ into a module of _storage-interface/_.  This decision sidesteps a set of orphaned type complexities that otherwise prevent the storage proto definitions from being moved out of the storage crates.

In the interest of gathering feedback and sanity checking direction, the PR leaves a couple of things unaddressed

Key questions
- Does this PR have the right shape for the desired end state?  
- A key distinction that I made in the PR is between the _proto_ files and definitions themselves and the _rust_ code written to wrap those protos.  I have consciously left the rust code in place.  Are there any strong reasons to keep the rust wrapper code with the protos?
- There are several options for the proto consolidation.  They can be placed in an umbrella directory in individual crates (as is done here);  they can be placed in an umbrella directory in a single crate;  they can be gathered under admission-control.  I find the first two options more attractive, but may not have full context.
- Are there any directory/crate/package naming conventions relevant to this approach that may not be obvious?

Assuming that the general direction is good and that there are consensus answers for the key questions, my plan is to refactor the changes into a clean and curated (set of) PRs for submission consideration.

yes

## Test Plan

Refactor;  don't break existing unit/integration tests.

## Related PRs

This is a draft PR to sanity check the approac


agclement:
Questions about have been answered; rolling this back to start with a clean slate.

PR #3842, opened 2020-05-12
[fuzzing] coverage needs to run with fuzzing flag enabled

mimoo:
just updating the README

### DIFF
--- a/testsuite/libra-fuzzer/README.md
+++ b/testsuite/libra-fuzzer/README.md
@@ -71,7 +71,7 @@ cargo run --bin investigate -- -i artifacts/compiled_module/crash-5d7f403f
 The following command will run libfuzzer on the relevant target with your input:
 
 ```
-// build single fuzzer for target using instruction in the 'google oss-fuzz integration' section
+# build single fuzzer for target using instruction in the 'google oss-fuzz integration' section
 ./fuzzer input
 ```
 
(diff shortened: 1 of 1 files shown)


testsuite/libra-fuzzer/README.md
```diff
@@ -71,7 +71,7 @@ cargo run --bin investigate -- -i artifacts/compiled_module/crash-5d7f403f
 The following command will run libfuzzer on the relevant target with your input:
 
 ```
-// build single fuzzer for target using instruction in the 'google oss-fuzz integration' section
+# build single fuzzer for target using instruction in the 'google oss-fuzz integration' section
```
davidiw, line 74: why the change?

testsuite/libra-fuzzer/README.md
```diff
@@ -86,7 +86,7 @@ See [Google OSS-Fuzz's documentation on reproducing bugs](https://google.github.
 To test coverage of our fuzzers you can run the following command with [grcov](https://github.com/mozilla/grcov):
 
 ```
-CORPUS_PATH=fuzz/corpus cargo xtest --html-cov-dir <some path for html output> -p libra-fuzzer -- coverage
+RUSTFLAGS='--cfg feature="fuzzing"' CORPUS_PATH=fuzz/corpus cargo xtest --html-cov-dir <some path for html output> -p libra-fuzzer -- coverage
```
davidiw, line 89: we should code mod this to testing since it is used for both purposes.

rexhoffman: Pretty sure that xtest, which turns on all features makes this unnecessary.   Also confident that x will ignore this set of flags and not pass them through.   Confident that this is a no op.

testsuite/libra-fuzzer/README.md
mimoo: that's a bash comment

PR #3843, opened 2020-05-12
[move-prover] Bound size of serialization output to work around termination problem

wrwg:
This introduces a new flag `--serialize-bound=N` which bounds the length of the vector created by LCS::to_bytes. It turns out that this lets the non-termination in `regression/performance_200511` go away.

The value defaults to 4 right now. Experiments on the above example show:

- bound=4 ==> takes a few seconds
- bound=8 ==> takes about 40 seconds
- bound=16 ==> takes hundreds of seconds
- no bound ==> does not terminate

The above behavior is independent of whether we keep the axioms for serialization or not.

Why this seems to fix the problem is unclear. Notice that termination is really about finding an error; if termination happens, a valid counter example is printed out, even though Z3 marks the problem as inconclusive (`(:reason-unknown "(incomplete quantifiers)")`). So this might not be the last word for this problem, but it isolates it further and unblocks us.

Restricting the size to 4 should be (mostly) sound because in theory, a perfect serializer could manage to do something like this, as long as properties like injectivity et. al are satisfied.

As part of this PR we also introduce a new flag `--generate-smt=true|false` which let the prover generate an smtlib file per verification problem. This aids debugging on Z3 level.

## Motivation

Performance.

Yes

## Test Plan

Regression test now passes.

## Related PRs

NA

### DIFF
--- a/language/move-prover/src/cli.rs
+++ b/language/move-prover/src/cli.rs
@@ -304,7 +304,8 @@ impl Options {
                     .require_equals(true)
                     .help(
                         "a bound for the length of the result vector of LCS::to_bytes(). \
-                        Currently, solver runtime increases exponentially with the size of this value."
+                        Currently, solver runtime increases exponentially with the size of this value. \
+                        Use 0 for no bound"
                     ),
             )
             .arg(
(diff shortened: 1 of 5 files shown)


junkil-park:
Could we have an option to specify N to be unbounded (e.g., infty)?
It's because that's the most accurate to the original semantics of `LCS::to_bytes` although the verification does not terminate when it's unbounded?

wrwg:
> Could we have an option to specify N to be unbounded (e.g., infty)?
> It's because that's the most accurate to the original semantics of `LCS::to_bytes` although the verification does not terminate when it's unbounded?

Done. One can now use `--serialize-bound=0` to remove any bound.

junkil-park:
Nice!!

PR #3845, opened 2020-05-12
[LCS hashing] Improve LCS-based implementations of CryptoHash

matbd:
## Motivation

* Refactor LCS to avoid memory allocations when hashing bytes.
* Add a new procedural macro `LCSCryptoHash` that implements `CryptoHash` using the existing `CryptoHasher` framework plus the new API `lcs::serialize_into`
* Use it to remove many manual implementations of `CryptoHash` where `lcs::to_bytes` was used (this should be faster now)
* Not trying to convert manual hashing into LCS hashing yet until we figure out why benchmark results are so different.

## Test Plan

I used the benchmark below to test that hashing with `lcs::serialize_into` is faster than with the old `lcs::to_bytes`.

```
cargo x test -p libra-crypto
cargo bench -p libra-crypto hash
```

### DIFF
--- a/common/lcs/src/error.rs
+++ b/common/lcs/src/error.rs
@@ -11,6 +11,8 @@ pub type Result<T, E = Error> = std::result::Result<T, E>;
 pub enum Error {
     #[error("unexpected end of input")]
     Eof,
+    #[error("I/O error: {0}")]
+    Io(String),
     #[error("exceeded max sequence length")]
     ExceededMaxLen(usize),
     #[error("expected boolean")]
@@ -37,6 +39,12 @@ pub enum Error {
     NonCanonicalUleb128Encoding,
     #[error("ULEB128-encoded integer did not fit in the target size")]
     IntegerOverflowDuringUleb128Decoding,
+}
+
+impl From<std::io::Error> for Error {
+    fn from(err: std::io::Error) -> Self {
+        Error::Io(err.to_string())
+    }
 }
 
 impl ser::Error for Error {
--- a/common/lcs/src/ser.rs
+++ b/common/lcs/src/ser.rs
@@ -51,9 +51,42 @@ where
     T: ?Sized + Serialize,
 {
     let mut output = Vec::new();
-    let serializer = Serializer::new(&mut output);
-    value.serialize(serializer)?;
+    serialize_into(&mut output, value)?;
     Ok(output)
+}
+
+/// Same as `to_bytes` but write directly into an `std::io::Write` object.
+pub fn serialize_into<W, T>(write: &mut W, value: &T) -> Result<()>
+where
+    W: std::io::Write,
+    T: ?Sized + Serialize,
+{
+    let serializer = Serializer::new(write);
+    value.serialize(serializer)
+}
+
+struct WriteCounter(usize);
+
+impl std::io::Write for WriteCounter {
+    fn write(&mut self, buf: &[u8]) -> std::io::Result<usize> {
+        let len = buf.len();
+        self.0 += len;
+        Ok(len)
+    }
+
+    fn flush(&mut self) -> std::io::Result<()> {
+        Ok(())
+    }
+}
+
+/// Same as `to_bytes` but only return the size of the serialized bytes.
+pub fn serialized_size<T>(value: &T) -> Result<usize>
+where
+    T: ?Sized + Serialize,
+{
+    let mut counter = WriteCounter(0);
+    serialize_into(&mut counter, value)?;
+    Ok(counter.0)
 }
 
 pub fn is_human_readable() -> bool {
@@ -63,13 +96,16 @@ pub fn is_human_readable() -> bool {
 }
 
 /// Serialization implementation for LCS
-struct Serializer<'a> {
-    output: &'a mut Vec<u8>,
+struct Serializer<'a, W> {
+    output: &'a mut W,
 }
 
-impl<'a> Serializer<'a> {
+impl<'a, W> Serializer<'a, W>
+where
+    W: std::io::Write,
+{
     /// Creates a new `Serializer` which will emit LCS.
-    fn new(output: &'a mut Vec<u8>) -> Self {
+    fn new(output: &'a mut W) -> Self {
         Self { output }
     }
 
@@ -77,11 +113,11 @@ impl<'a> Serializer<'a> {
         while value >= 0x80 {
             // Write 7 (lowest) bits of data and set the 8th bit to 1.
             let byte = (value & 0x7f) as u8;
-            self.output.push(byte | 0x80);
+            self.output.write_all(&[byte | 0x80])?;
             value >>= 7;
         }
         // Write the remaining bits of data and set the highest bit to 0.
-        self.output.push(value as u8);
+        self.output.write_all(&[value as u8])?;
         Ok(())
     }
 
@@ -98,14 +134,17 @@ impl<'a> Serializer<'a> {
     }
 }
 
-impl<'a> ser::Serializer for Serializer<'a> {
+impl<'a, W> ser::Serializer for Serializer<'a, W>
+where
+    W: std::io::Write,
+{
     type Ok = ();
     type Error = Error;
     type SerializeSeq = Self;
     type SerializeTuple = Self;
     type SerializeTupleStruct = Self;
     type SerializeTupleVariant = Self;
-    type SerializeMap = MapSerializer<'a>;
+    type SerializeMap = MapSerializer<'a, W>;
     type SerializeStruct = Self;
     type SerializeStructVariant = Self;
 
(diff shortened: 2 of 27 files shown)


crypto/crypto/src/hash.rs
```diff
@@ -605,7 +612,104 @@ impl<T: ser::Serialize + ?Sized> TestOnlyHash for T {
...
+        hasher.finish()
+    }
+}
+
+/// Alternative to CryptoHash
+pub trait SimpleHash {
+    /// The hash method
+    fn hash(&self) -> HashValue;
+
+    #[cfg(any(test, feature = "fuzzing"))]
```
davidiw, line 625: not sure we really need this behind a testing flag.

crypto/crypto/src/hash.rs
```diff
@@ -605,7 +612,104 @@ impl<T: ser::Serialize + ?Sized> TestOnlyHash for T {
...
-        hasher.write(&bytes);
+        hasher.update(&bytes);
+        hasher.finish()
+    }
+}
+
+/// Alternative to CryptoHash
+pub trait SimpleHash {
+    /// The hash method
+    fn hash(&self) -> HashValue;
```
jolestar, line 623: Whether to consider using an alternative name to avoid conflict with the std Hash trait?

matbd: I remember a long discussion where I was supporting `digest` for this reason, but I'm not sure I want to re-open the question :)

common/lcs/src/ser.rs
```diff
@@ -197,7 +192,7 @@ impl<'a> ser::Serializer for &'a mut Serializer {
     where
         T: ?Sized + Serialize,
     {
-        self.serialize_u8(1)?;
+        self.output.push(1);
```
bmwill: This seems like a step back in terms of readability, we should probably try to ensure that all writing of data goes through the "serialize" methods.

common/lcs/src/error.rs
```diff
@@ -11,6 +11,8 @@ pub type Result<T, E = Error> = std::result::Result<T, E>;
 pub enum Error {
     #[error("unexpected end of input")]
     Eof,
+    #[error("I/O error: {0}")]
+    Io(String),
```
bmwill, line 15: Drop the custom from impl and do the following instead, unless you only wanted to store the string and not the full io error:

```suggestion
    Io(#[from] std::io::Error),
```

crypto/crypto/src/hash.rs
```diff
@@ -605,3 +616,101 @@ impl<T: ser::Serialize + ?Sized> TestOnlyHash for T {
...
 }
+
+/// Alternative to CryptoHash
+pub trait SimpleHash {
+    /// The hash method
+    fn hash(&self) -> HashValue;
+
+    #[cfg(any(test, feature = "fuzzing"))]
+    /// The salt used for hashing this type.
+    fn salt() -> Vec<u8>;
```
bmwill, line 627: The salt should probably be something that is unchanging based on internal state, right? So this should instead have a signature of:

```suggestion
    fn salt() -> &'static [u8];
```

crypto/crypto/src/hash.rs
```diff
@@ -605,3 +616,101 @@ impl<T: ser::Serialize + ?Sized> TestOnlyHash for T {
...
+            Network => b"Network",
+        }
+    }
+}
+
+fn compute_canonical_salt<'de, T>() -> Vec<u8>
+where
+    T: CanonicalHash + serde::Serialize + serde::Deserialize<'de>,
+{
+    println!("Called for {}", std::any::type_name::<T>());
```
bmwill, line 676: Change this to a log line?

common/lcs/src/ser.rs
matbd: `serialize_u8` is a user-facing API that consumes `self`. If needed, we should be able to create internal helper function, though.

crypto/crypto/src/hash.rs
matbd: oh my bad, I was supposed to remove this

matbd:
FYI I'm going to try to make this PR a complete but non-breaking code-refactoring and postpone the new seeding mechanism (based on `serde-name`) to a next PR.

crypto/crypto/src/hash.rs
matbd: We would have to memoize it as well then.

common/lcs/src/error.rs
matbd: I get this compilation error:
```
binary operation `!=` cannot be applied to type `std::io::Error`
```

matbd:
> Fantastic!!! Please mark this as breaking changes.

@davidiw This PR is not breaking anything (or at least should not be). For now, I have also lowered my ambitious and chosen not to completely remove manual implementations of `CryptoHash` (then CryptoHasher`) to avoid any performance regressions.

crypto/crypto/src/hash.rs
```diff
@@ -66,26 +66,15 @@
...
-//!     fn hash(&self) -> HashValue {
-//!         let mut state = Self::Hasher::default();
-//!         state.write(b"Struct serialized into bytes here");
-//!         state.finish()
-//!     }
-//! }
+//! # use libra_crypto_derive::*;
+//! # use serde::Serialize;
+//! #[derive(Serialize, CryptoHasher, LCSCryptoHash)]
+//! struct MyNewStruct { /* ... */ }
```
huitseeker, line 77: That's great, but can we sketch the use of the hash?

crypto/crypto/src/unit_tests/cryptohasher.rs
```diff
@@ -1,62 +1,84 @@
...
+use libra_crypto_derive::{CryptoHasher, LCSCryptoHash};
+use serde::{Deserialize, Serialize};
 use tiny_keccak::{Hasher, Sha3};
 
-/// This file tests the CryptoHasher derive macro defined in the crypto-derive
-/// crate. This macro, as the comments there indicate, is meant to derive an
-/// instance of CryptoHasher (see `crate::hash`) for any type, which
-/// domain-separation tag is precisely the fully-qualified name of the
-/// struct. Here, we create a struct, manually "compute" its tag and compare it
-/// to the generated one.
```
huitseeker, line 17: It's unfortunate that we lose the comment here. People will not expect this
test to actually be for a different crate.

matbd: I tried to summarize this comment in the new line above and with the realistic example `Foo` below.

crypto/crypto/src/hash.rs
matbd: Good idea, thanks!

common/lcs/src/error.rs
davidiw: ahh yes, the partialeq

crypto/crypto-derive/src/lib.rs
```diff
@@ -366,12 +366,48 @@ pub fn hasher_dispatch(input: TokenStream) -> TokenStream {
...
+
+    );
+    out.into()
+}
+
+#[proc_macro_derive(LCSCryptoHash)]
+pub fn lcs_crypto_hash_dispatch(input: TokenStream) -> TokenStream {
+    let ast = parse_macro_input!(input as DeriveInput);
+    let name = &ast.ident;
+    let hasher_name = Ident::new(&format!("{}Hasher", &name.to_string()), Span::call_site());
```
davidiw, line 395: I take it this call site is the module / struct wherein this is derived?

crypto/crypto-derive/src/lib.rs
```diff
@@ -366,12 +366,48 @@ pub fn hasher_dispatch(input: TokenStream) -> TokenStream {
...
+    let error_msg = syn::LitStr::new(
+        &format!("LCS serialization of {} should not fail", name.to_string()),
+        Span::call_site(),
+    );
+    let out = quote!(
+
```
davidiw, line 401: interesting choice of whitespace, I would have imagined a newline above let out and none before quote..

crypto/crypto/src/unit_tests/cryptohasher.rs
```diff
@@ -1,62 +1,84 @@
...
+        .as_bytes()
+        .to_vec();
+    name.extend_from_slice(LIBRA_HASH_SUFFIX);
 
+    let value = Foo { a: 5, b: 1025 };
     let expected = {
+        let mut digest = Sha3::v256();
+        digest.update(HashValue::from_sha3_256(&name[..]).as_ref());
+        digest.update(&value.a.to_le_bytes());
+        digest.update(&value.b.to_le_bytes());
```
davidiw, line 71: nothing like a little hand implemented LCS... why not use LCS here?

matbd: Sure, we could do that too

crypto/crypto-derive/src/lib.rs
matbd: I made it look more like the code above but no strong opinion otherwise.

crypto/crypto-derive/src/lib.rs
matbd: That is part of the macro expansion.

crypto/crypto/src/unit_tests/cryptohasher.rs
matbd: done

PR #3846, opened 2020-05-12
[language, move] Introduce public key validation

huitseeker:
TL;DR: Introduce a native function for public key validation, and use it in TR key rotation.

This solves the problem that any 32-byte value would be accepted as a public key in the stdlib contracts, despite the length check being insufficient to protect against [invalid point attacks, small subgroup attacks](https://link.springer.com/content/pdf/10.1007/3-540-44598-6_8.pdf), and typos. 

In the last case, before this PR, a user would typically rotate their key to an invalid public key unwittingly, and then fail to produce signatures against that new key, ~~losing the balance of their account in the process~~ losing the ability to demonstrate the legality of their transactions. We'd rather have early failure.

This introduces a native function for checking the validity of this public key, and uses it upon rotation. It would probably be useful to call this public key validation function in a few more places - in fact, anywhere externally-controlled key material is registered on-chain. I have not looked beyond the key rotation contracts — Edit : and publication — for now.

A further step would be an opt-in proof of ownership (TBD).

### DIFF
--- a/language/move-lang/tests/functional/natives/signature.move
+++ b/language/move-lang/tests/functional/natives/signature.move
@@ -1,0 +1,21 @@
+// Test fot public key validation
+
+script {
+use 0x0::Signature;
+use 0x0::Transaction;
+
+fun main() {
+
+    // from RFC 8032
+    let valid_pubkey =  x"3d4017c3e843895a92b70aa74d1b7ebc9c982ccf2ec4968cc0cd55f12af4660c";
+    let short_pubkey = x"100";
+    // concatenation of the two above
+    let long_pubkey = x"1003d4017c3e843895a92b70aa74d1b7ebc9c982ccf2ec4968cc0cd55f12af4660c";
+    let invalid_pubkey = x"0000000000000000000000000000000000000000000000000000000000000000";
+
+    Transaction::assert(Signature::ed25519_validate_pubkey(valid_pubkey), 9003);
+    Transaction::assert(!Signature::ed25519_validate_pubkey(short_pubkey), 9003);
+    Transaction::assert(!Signature::ed25519_validate_pubkey(long_pubkey), 9003);
+    Transaction::assert(!Signature::ed25519_validate_pubkey(invalid_pubkey), 9003);
+}
+}
--- a/language/move-lang/tests/functional/shared_ed25519_public_key/shared_key.move
+++ b/language/move-lang/tests/functional/shared_ed25519_public_key/shared_key.move
@@ -10,7 +10,8 @@ use 0x0::SharedEd25519PublicKey;
 use 0x0::Transaction;
 fun main() {
     let old_auth_key = LibraAccount::authentication_key({{default}});
-    let pubkey1 = x"0000000000000000000000000000000000000000000000000000000000000000";
+    // from RFC 8032
+    let pubkey1 = x"3d4017c3e843895a92b70aa74d1b7ebc9c982ccf2ec4968cc0cd55f12af4660c";
     SharedEd25519PublicKey::publish(copy pubkey1);
     let new_auth_key = LibraAccount::authentication_key({{default}});
 
@@ -24,7 +25,8 @@ fun main() {
     Transaction::assert(copy new_auth_key != old_auth_key, 3003);
 
     // now rotate to another pubkey and redo the key-related checks
-    let pubkey2 = x"1000000000000000000000000000000000000000000000000000000000000000";
+    // from RFC 8032
+    let pubkey2 = x"d75a980182b10ab7d54bfed3c964073a0ee172f3daa62325af021a68f707511a";
     SharedEd25519PublicKey::rotate_sender_key(copy pubkey2);
     Transaction::assert(SharedEd25519PublicKey::key({{default}}) == pubkey2, 3004);
     // make sure the auth key changed again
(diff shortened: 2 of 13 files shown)


language/move-lang/tests/functional/shared_ed25519_public_key/shared_key.move
```diff
@@ -61,3 +64,19 @@ fun main() {
 }
 // check: ABORTED
 // check: 7000
+
+// rotating to a key with a good length but bad contents should fail
```
valerini, line 68: Do we want to have separate tests for failure cases:
1) point not on curve,
2) point of small order,
3) point not canonically serialized?

language/stdlib/modules/approved_payment.move
```diff
@@ -65,7 +65,15 @@ module ApprovedPayment {
 
     // Rotate the key used to sign approved payments. This will invalidate any approved payments
     // that are currently in flight
-    public fun rotate_key(approved_payment: &mut T, new_public_key: vector<u8>) {
+    public fun rotate_key(approved_payment: &mut T, new_public_key: vector<u8>)
+    {
```
valerini, line 69: Might want to return { to the previous line.

language/move-lang/tests/functional/shared_ed25519_public_key/shared_key.move
huitseeker: Not for now, as the native function does not expose them.
 It may be an improvement if Move has the error code breadth.

language/move-prover/tests/sources/stdlib/modules/signature.move
```diff
@@ -1,6 +1,7 @@
 address 0x0 {
 
 module Signature {
+    native public fun ed25519_validate_pubkey(public_key: vector<u8>): bool;
```
sblackshear, line 4: Nit: comment would be good here (I know it's missing for the others, boo).

language/stdlib/modules/shared_ed25519_public_key.move
```diff
@@ -34,6 +35,11 @@ module SharedEd25519PublicKey {
 
     fun rotate_key(shared_key: &mut T, new_public_key: vector<u8>) {
         Transaction::assert(Vector::length(&new_public_key) == 32, 7000);
```
sblackshear, line 37: Can probably kill this check now (since `validate_pubkey` presumably returns false for a bad length)?

language/move-lang/tests/functional/shared_ed25519_public_key/shared_key.move
sblackshear: I think it would be useful to have these tests + maybe a few others (e.g. key too long, key too short) in `functional/natives/signature.move`. I don't think we want/need a more sophisticated error codes to write these tests--checking that the function says `false` for all of them is enough.

junkil-park:
I confirm that the latest commit (f1aeaab) (with modifying `prelude.bpl`) passed the Prover test in CI (https://github.com/mimoo/mirai-bot/runs/668772188?check_suite_focus=true#step:7:1). Miraibot is silent when it finds no errors.

davidiw:
Thanks for tackling this important endeavor!

language/stdlib/modules/shared_ed25519_public_key.move
huitseeker: The nice thing you have with this is the consistent error code for a length error.

language/move-lang/tests/functional/shared_ed25519_public_key/shared_key.move
huitseeker: Done

language/move-lang/tests/functional/natives/signature.move
```diff
@@ -0,0 +1,21 @@
...
+    // concatenation of the two above
+    let long_pubkey = x"1003d4017c3e843895a92b70aa74d1b7ebc9c982ccf2ec4968cc0cd55f12af4660c";
+    let invalid_pubkey = x"0000000000000000000000000000000000000000000000000000000000000000";
+
+    Transaction::assert(Signature::ed25519_validate_pubkey(valid_pubkey) == true, 9003);
+    Transaction::assert(Signature::ed25519_validate_pubkey(short_pubkey) == false, 9003);
+    Transaction::assert(Signature::ed25519_validate_pubkey(long_pubkey) == false, 9003);
+    Transaction::assert(Signature::ed25519_validate_pubkey(invalid_pubkey) == false, 9003);
+}
+}
```
sblackshear, line 21: <3

language/move-lang/tests/functional/natives/signature.move
```diff
@@ -0,0 +1,21 @@
...
+fun main() {
+
+    // from RFC 8032
+    let valid_pubkey =  x"3d4017c3e843895a92b70aa74d1b7ebc9c982ccf2ec4968cc0cd55f12af4660c";
+    let short_pubkey = x"100";
+    // concatenation of the two above
+    let long_pubkey = x"1003d4017c3e843895a92b70aa74d1b7ebc9c982ccf2ec4968cc0cd55f12af4660c";
+    let invalid_pubkey = x"0000000000000000000000000000000000000000000000000000000000000000";
+
+    Transaction::assert(Signature::ed25519_validate_pubkey(valid_pubkey) == true, 9003);
```
sblackshear, line 16: Nit: no need for `== true` (and can do `!` for `== false`).

PR #3847, opened 2020-05-12
[language][file format] implement self module handle index

vgao1996:
## Summary
This adds a field `self_module_handle_idx` to the compiled module and gets rid of IMPLEMENTED_MODULE_INDEX.

With this change, the compiler will also no longer generate a self module handle for scripts.

## Test Plan
cargo test


language/vm/src/serializer.rs
```diff
@@ -1088,6 +1098,19 @@ impl ModuleSerializer {
...
+        module: &CompiledModuleMut,
+    ) -> Result<()> {
+        self.common.table_count = self.common.table_count.wrapping_add(1); // the count will bound to a small number
+        self.misc.0 = check_index_in_binary(binary.len())?;
+        // self module handle index
+        write_u16_as_uleb128(binary, module.self_module_handle_idx.0)?;
+        self.misc.1 = checked_calculate_table_size(binary, self.misc.0)?;
+        Ok(())
+    }
+
```
vgao1996, line 1113: @dariorussi Probably the only interesting part of this PR: currently I store the self module handle index in a new table called "MISC", but I'm open to change:
- Do we need a separate table for this? Would it be messy if we just store the field in another table like struct definitions?
- Better name if we do want a new table?

dariorussi: I think we talked about this and for now let's go in and let's review soon.
Essentially the table is strange for a single element entry. 
Looking into making it an explicit "field" is probably the right thing

language/bytecode-verifier/bytecode-verifier-tests/src/unit_tests/duplication_tests.rs
```diff
@@ -9,6 +9,8 @@ proptest! {
     #[test]
     fn valid_duplication(module in CompiledModule::valid_strategy(20)) {
         let duplication_checker = DuplicationChecker::new(&module);
-        prop_assert!(duplication_checker.verify().is_ok());
+        let res = duplication_checker.verify();
+        println!("{:?}", res);
```
dariorussi, line 13: the `println` should probably go, right? in fact we could revert this, can we?

vgao1996: Ah my bad, forgot to clean this up. Will revert before landing.

PR #3849, opened 2020-05-12
[network] propagate noise pubkey in addr everywhere

phlip9:
+ For the main validator and full node network use cases, the noise and
handshake protocols should both be present in the `NetworkAddress`es
throughout the system.

+ The only discrepancy is that `NetworkConfig` currently contains the
"base" network address (e.g., `"/ip6/::1/tcp/6180"`) for both listen and
advertised addresses, which are then rendered into the "prod" address
(e.g., `"/ip6/::1/tcp/6180/ln-noise-ik/<pubkey>/ln-handshake/<version>"`)
just before they're handed to `NetworkBuilder`.

+ It would be cleaner and easier to specify protocol upgrades if the
addresses in configs were always fully rendered. Some future steps in
this direction include (1) modify ops code to always fully render
addresses, (2) configs and network should store a `Vec<NetworkAddress>`
for listen and advertisement, (3) support listening on multiple addrs.

+ With this change, we're now much closer to full noise ik integration,
since the noise pubkey is pushed down all the way to the transport when
dialing.

+ All on-chain network addrs are fully rendered.

+ There are some hacky bits since this PR is an intermediate step in the
overall `Transport` refactor. Specifically, the current `Transport` API
is not amenable to incrementally parsing the `NetworkAddress`.
Consequently, the parsing in this PR is pretty loose though this should
change soon when monolithic Transports are introduced. By "loose parsing"
I mean dialing on a Memory transport with
`/memory/123/ip4/1.2.3.4/dns/foo.example` which is clearly non-sensical
will actually succeed as the Memory transport is only parsing out the
`/memory/123` prefix while the suffix is ignored.

### DIFF
--- a/language/tools/vm-genesis/src/lib.rs
+++ b/language/tools/vm-genesis/src/lib.rs
@@ -385,15 +385,25 @@ pub fn validator_registrations(node_configs: &[NodeConfig]) -> Vec<ValidatorRegi
             let network = n.validator_network.as_ref().unwrap();
             let network_keypairs = network.network_keypairs.as_ref().unwrap();
             let signing_key = network_keypairs.signing_keypair.public_key();
-            let raw_advertised_address =
-                RawNetworkAddress::try_from(&network.advertised_address).unwrap();
+
+            let identity_key = network_keypairs.identity_keypair.public_key();
+            let handshake_version = network.handshake_version;
+
+            let advertised_address = network
+                .advertised_address
+                .clone()
+                .into_prod(identity_key, handshake_version);
+            let raw_advertised_address = RawNetworkAddress::try_from(&advertised_address).unwrap();
+
+            // TODO(philiphayes): do something with n.full_node_networks instead
+            // of ignoring them?
 
             let script = transaction_builder::encode_register_validator_script(
                 consensus_key.to_bytes().to_vec(),
                 signing_key.to_bytes().to_vec(),
-                network_keypairs.identity_keypair.public_key().to_bytes(),
+                identity_key.to_bytes(),
                 raw_advertised_address.clone().into(),
-                network_keypairs.identity_keypair.public_key().to_bytes(),
+                identity_key.to_bytes(),
                 raw_advertised_address.into(),
             );
             (account_key, script)
--- a/language/tools/vm-genesis/src/lib.rs
+++ b/language/tools/vm-genesis/src/lib.rs
@@ -37,6 +37,9 @@ const GENESIS_SEED: [u8; 32] = [42; 32];
 
 const GENESIS_MODULE_NAME: &str = "Genesis";
 
+// TODO(philiphayes): probably not the right place to put this...
+const HANDSHAKE_VERSION: u8 = 0;
+
 pub static GENESIS_KEYPAIR: Lazy<(Ed25519PrivateKey, Ed25519PublicKey)> = Lazy::new(|| {
     let mut rng = StdRng::from_seed(GENESIS_SEED);
     let private_key = Ed25519PrivateKey::generate(&mut rng);
@@ -385,14 +388,12 @@ pub fn validator_registrations(node_configs: &[NodeConfig]) -> Vec<ValidatorRegi
             let network = n.validator_network.as_ref().unwrap();
             let network_keypairs = network.network_keypairs.as_ref().unwrap();
             let signing_key = network_keypairs.signing_keypair.public_key();
-
             let identity_key = network_keypairs.identity_keypair.public_key();
-            let handshake_version = network.handshake_version;
 
             let advertised_address = network
                 .advertised_address
                 .clone()
-                .into_prod(identity_key, handshake_version);
+                .append_prod_protos(identity_key, HANDSHAKE_VERSION);
             let raw_advertised_address = RawNetworkAddress::try_from(&advertised_address).unwrap();
 
             // TODO(philiphayes): do something with n.full_node_networks instead
--- a/libra-node/src/main_node.rs
+++ b/libra-node/src/main_node.rs
@@ -192,6 +192,11 @@ pub fn setup_network(
             })
             .collect();
 
+        info!(
+            "network setup: role: {}, seed_peers: {:?}, network_peers: {:?}",
+            role, seed_peers, network_peers,
+        );
+
         let trusted_peers = if role == RoleType::Validator {
             // for validators, trusted_peers is empty will be populated from consensus
             HashMap::new()
--- a/libra-node/src/main_node.rs
+++ b/libra-node/src/main_node.rs
@@ -130,16 +130,15 @@ pub fn setup_network(
         .enable_all()
         .build()
         .expect("Failed to start runtime. Won't be able to start networking.");
-    let mut network_builder = NetworkBuilder::new(
-        runtime.handle().clone(),
-        config.peer_id,
-        config.listen_address.clone(),
-        role,
-    );
+
+    let mut network_builder = NetworkBuilder::new(runtime.handle().clone(), config.peer_id, role);
     network_builder
         .enable_remote_authentication(config.enable_remote_authentication)
-        .advertised_address(config.advertised_address.clone())
         .add_connection_monitoring();
+
+    // TODO(philiphayes): it might make more sense to refactor this so we have a
+    // config per "archetype"?
+
     if config.enable_remote_authentication {
         // If the node wants to run in permissioned mode, it should also have authentication and
         // encryption.
@@ -148,6 +147,8 @@ pub fn setup_network(
             "Permissioned network end-points should use authentication"
         );
         let seed_peers = config.seed_peers.seed_peers.clone();
+        let network_peers = config.network_peers.peers.clone();
+
         let network_keypairs = config
             .network_keypairs
             .as_mut()
@@ -162,20 +163,59 @@ pub fn setup_network(
             .identity_keypair
             .take_private()
             .expect("identity key should be present");
+        let identity_pubkey = identity_key.public_key();
+
+        let handshake_version = config.handshake_version;
+
+        // TODO(philiphayes): ideally config should contain full addrs
+        let listen_address = config
+            .listen_address
+            .clone()
+            .into_prod(identity_pubkey, handshake_version);
+        let advertised_address = config
+            .advertised_address
+            .clone()
+            .into_prod(identity_pubkey, handshake_version);
+
+        // TODO(philiphayes): seed peers should be fully rendered. then we can
+        // also get rid of `network_peers`. we'll need to make some changes to
+        // ops code since it only specifies the base address.
+        let seed_peers: HashMap<_, _> = seed_peers
+            .into_iter()
+            .map(|(peer_id, addrs)| {
+                let opt_id_pubkey = network_peers
+                    .get(&peer_id)
+                    .map(|info| info.identity_public_key);
+
+                match opt_id_pubkey {
+                    Some(id_pubkey) => {
+                        let addrs = addrs
+                            .into_iter()
+                            .map(|addr| addr.into_prod(id_pubkey, handshake_version))
+                            .collect();
+                        (peer_id, addrs)
+                    }
+                    // TODO(philiphayes): can skip these when transport actually
+                    // uses pubkey
+                    None => (peer_id, addrs),
+                }
+            })
+            .collect();
 
         let trusted_peers = if role == RoleType::Validator {
             // for validators, trusted_peers is empty will be populated from consensus
             HashMap::new()
         } else {
-            config.network_peers.peers.clone()
+            network_peers
         };
         network_builder
+            .listen_address(listen_address)
+            .advertised_address(advertised_address)
             .transport(TransportType::TcpNoise(Some(identity_key)))
-            .connectivity_check_interval_ms(config.connectivity_check_interval_ms)
             .seed_peers(seed_peers)
             .trusted_peers(trusted_peers)
             .signing_keypair((signing_private, signing_public))
-            .discovery_interval_ms(config.discovery_interval_ms)
+            .connectivity_check_interval_ms(config.connectivity_check_interval_ms)
             .add_connectivity_manager();
     } else if config.enable_noise {
         let identity_key = config
@@ -185,17 +225,37 @@ pub fn setup_network(
             .identity_keypair
             .take_private()
             .expect("identity key should be present");
+        let identity_pubkey = identity_key.public_key();
+
+        let listen_address = config
+            .listen_address
+            .clone()
+            .into_prod(identity_pubkey, config.handshake_version);
+        let advertised_address = config
+            .advertised_address
+            .clone()
+            .into_prod(identity_pubkey, config.handshake_version);
+
         // Even if a network end-point operates without remote authentication, it might want to prove
         // its identity to another peer it connects to. For this, we use TCP + Noise but without
         // enforcing a trusted peers set.
-        network_builder.transport(TransportType::TcpNoise(Some(identity_key)));
+        network_builder
+            .transport(TransportType::TcpNoise(Some(identity_key)))
+            .listen_address(listen_address)
+            .advertised_address(advertised_address);
     } else {
-        network_builder.transport(TransportType::Tcp);
+        // TODO(philiphayes): probably remove this branch since there are no
+        // current no noise or no client auth use cases.
+        network_builder
+            .transport(TransportType::Tcp)
+            .listen_address(config.listen_address.clone());
     }
 
     match config.discovery_method {
         DiscoveryMethod::Gossip => {
-            network_builder.add_gossip_discovery();
+            network_builder
+                .discovery_interval_ms(config.discovery_interval_ms)
+                .add_gossip_discovery();
         }
         DiscoveryMethod::Onchain => {
             setup_onchain_discovery(
--- a/libra-node/src/main_node.rs
+++ b/libra-node/src/main_node.rs
@@ -22,7 +22,7 @@ use libra_metrics::metric_server;
 use libra_types::{on_chain_config::ON_CHAIN_CONFIG_REGISTRY, waypoint::Waypoint, PeerId};
 use libra_vm::LibraVM;
 use libradb::LibraDB;
-use network::validator_network::network_builder::{NetworkBuilder, TransportType};
+use network::validator_network::network_builder::{AuthMode, NetworkBuilder, HANDSHAKE_VERSION};
 use onchain_discovery::{client::OnchainDiscovery, service::OnchainDiscoveryService};
 use state_synchronizer::StateSynchronizer;
 use std::{
@@ -131,10 +131,13 @@ pub fn setup_network(
         .build()
         .expect("Failed to start runtime. Won't be able to start networking.");
 
-    let mut network_builder = NetworkBuilder::new(runtime.handle().clone(), config.peer_id, role);
-    network_builder
-        .enable_remote_authentication(config.enable_remote_authentication)
-        .add_connection_monitoring();
+    let mut network_builder = NetworkBuilder::new(
+        runtime.handle().clone(),
+        config.peer_id,
+        role,
+        config.listen_address.clone(),
+    );
+    network_builder.add_connection_monitoring();
 
     // TODO(philiphayes): it might make more sense to refactor this so we have a
     // config per "archetype"?
@@ -144,7 +147,7 @@ pub fn setup_network(
         // encryption.
         assert!(
             config.enable_noise,
-            "Permissioned network end-points should use authentication"
+            "Permissioned network end-points must use authentication"
         );
         let seed_peers = config.seed_peers.seed_peers.clone();
         let network_peers = config.network_peers.peers.clone();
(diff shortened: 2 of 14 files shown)


network/netcore/src/transport/tcp.rs
```diff
@@ -74,18 +74,21 @@ impl Transport for TcpTransport {
...
-        let socket_addr = multiaddr_to_socketaddr(&addr)?;
+        let (socket_addr, addr_suffix) = parse_socketaddr(&addr)?;
         let config = self.clone();
+
         let listener = ::std::net::TcpListener::bind(&socket_addr)?;
-        let local_addr = NetworkAddress::from(listener.local_addr()?);
         let listener = TcpListener::try_from(listener)?;
 
+        let mut actual_addr = NetworkAddress::from(listener.local_addr()?);
+        actual_addr.extend_from_slice(addr_suffix);
```
davidiw, line 84: maybe extend_from_slice could consume and return

  extend_from_slice(val: mut T) -> T

network/netcore/src/transport/tcp.rs
```diff
@@ -193,36 +196,32 @@ impl AsyncWrite for TcpSocket {
...
-    })?;
-    let proto2 = iter.next().ok_or_else(|| {
-        io::Error::new(
-            io::ErrorKind::InvalidInput,
-            format!("Invalid NetworkAddress '{:?}'", addr),
-        )
-    })?;
-
-    if iter.next().is_some() {
+/// parse the `NetworkAddress` into the `/ip4/<ip>/tcp/<port>` or
```
davidiw, line 199: Same as mem socket, we should push this into NetworkAddress

network/network-address/src/lib.rs
```diff
@@ -271,12 +281,39 @@ impl NetworkAddress {
...
+
+    /// Given base advertised `NetworkAddress`, append production protocols and
+    /// return the modified `NetworkAddress`.
+    ///
+    /// ### Example
+    ///
+    /// `/dns/example.com/tcp/6180` =>
+    /// `/dns/example.com/tcp/6180/ln-noise-ik/<network_pubkey>/ln-handshake/<handshake_version>`
+    // TODO(philiphayes): handshake version type
+    pub fn into_prod(mut self, network_pubkey: x25519::PublicKey, handshake_version: u8) -> Self {
```
davidiw, line 296: hmm... naming is kinda rough, but I can't think of something better ... I wonder how much time you spent here :)

network/src/transport.rs
```diff
@@ -145,6 +145,43 @@ fn identity_key_to_peer_id(
     None
 }
 
+/// A temporary, hacky function to parse out the first `/ln-noise-ik/<pubkey>` from
+/// a `NetworkAddress`. We can remove this soon, when we move to the interim
+/// "monolithic" transport model.
+fn parse_noise_pubkey(addr: NetworkAddress) -> Result<x25519::PublicKey, io::Error> {
```
davidiw, line 151: this should be in the NetworkAddress

network/src/validator_network/network_builder.rs
```diff
@@ -82,11 +82,14 @@ pub enum TransportType {
 /// MempoolNetworkHandler and ConsensusNetworkHandler are constructed by calling
 /// [`NetworkBuilder::build`].  New instances of `NetworkBuilder` are obtained
 /// via [`NetworkBuilder::new`].
+// TODO(philiphayes): refactor NetworkBuilder and libra-node; current config is
```
davidiw, line 85: Aaron Zhou is looking at this problem too maybe you can share some insights?

language/tools/vm-genesis/src/lib.rs
```diff
@@ -385,15 +385,25 @@ pub fn validator_registrations(node_configs: &[NodeConfig]) -> Vec<ValidatorRegi
...
-                RawNetworkAddress::try_from(&network.advertised_address).unwrap();
+
+            let identity_key = network_keypairs.identity_keypair.public_key();
+            let handshake_version = network.handshake_version;
+
+            let advertised_address = network
+                .advertised_address
+                .clone()
+                .into_prod(identity_key, handshake_version);
+            let raw_advertised_address = RawNetworkAddress::try_from(&advertised_address).unwrap();
```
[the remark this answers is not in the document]
phlip9: i'm not sure i understand this comment? `RawNetworkAddress` is tackling a totally orthogonal issue:

```rust
/// A `RawNetworkAddress` is the serialized, unverified, on-chain representation
/// of a [`NetworkAddress`]. Specifically, a `RawNetworkAddress` is usually an
/// [`lcs`]-serialized `NetworkAddress`.
///
/// This representation is useful because:
///
/// 1. Move does't understand (and doesn't really need to understand) what a
///    `NetworkAddress` is, so we can just store an opaque `vector<u8>` on-chain,
///    which we represent as `RawNetworkAddress` in Rust.
/// 2. We want to deserialize a `Vec<NetworkAddress>` but ignore ones that don't
///    properly deserialize. For example, a validator might advertise several
///    `NetworkAddress`. If that validator is running a newer version of the node
///    software, and old versions can't understand one of the `NetworkAddress`,
///    they would be able to ignore the one that doesn't deserialize for them.
///    We can easily do this by storing a `vector<vector<u8>>` on-chain, which
///    we deserialize as `Vec<RawNetworkAddress>` in Rust. Then we deserialize
///    each `RawNetworkAddress` into a `NetworkAddress` individually.
///
/// Note: deserializing a `RawNetworkAddress` does no validation, other than
/// deserializing the underlying `Vec<u8>`.
```

i also don't understand the comment about not "leveraging the non-human-readable ser/de". all network addresses stored on-chain and sent over-the-wire (the important places) are in the compact binary format. for our prod addresses, the compact format is about 35-40% the size of the human readable format

```rust
let addr_str = "/ip4/123.123.123.123/tcp/55555/ln-noise-ik/080e287879c918794170e258bfaddd75acac5b3e350419044655e4983a487120/ln-handshake/233";
let addr = NetworkAddress::from_str(addr_str).unwrap();
let raw_addr = RawNetworkAddress:try_from(&addr).unwrap();

addr_str.len()
// => 124 bytes

raw_addr.len()
// => 45 bytes

45f / 124
// => 0.362..
```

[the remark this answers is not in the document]
davidiw: Yes, this is totally orthogonal to the PR, but I feel like we should explore merging RawNetworkAddress concepts directly into NetworkAddress, it isn't clear to me why we want both? For example in other code, you convert from NetworkAddress to RawNetworkAddress using a try_from... but that should be infallible afaiu. if it were all self-contained this would be more self-evident.

phlip9:
not letting me reply inline 😭 

> Yes, this is totally orthogonal to the PR, but I feel like we should explore merging RawNetworkAddress concepts directly into NetworkAddress, it isn't clear to me why we want both? For example in other code, you convert from NetworkAddress to RawNetworkAddress using a try_from... but that should be infallible afaiu. if it were all self-contained this would be more self-evident.

yeah the only reason `NetworkAddress` => `RawNetworkAddress` is b/c `lcs::to_bytes` is apparently (?) fallible :'(

i would love it if it weren't haha

as for why even have `RawNetworkAddress` in the first place, here is an example of ignoring `RawNetworkAddress`es that fail to deserialize:

https://github.com/libra/libra/blob/master/network/onchain-discovery/src/types.rs#L144

```rust
#[derive(Clone, Debug, Eq, PartialEq)]
pub struct DiscoverySetInternal(pub HashMap<PeerId, DiscoveryInfoInternal>);

impl DiscoverySetInternal {
    pub fn from_validator_set(role_filter: RoleType, validator_set: ValidatorSet) -> Self {
        Self(
            validator_set
                .into_iter()
                .filter_map(|validator_info| {
                    let peer_id = *validator_info.account_address();
                    let res_info =
                        DiscoveryInfoInternal::try_from_validator_info(role_filter, validator_info);

                    // ignore network addresses that fail to deserialize
                    res_info
                        .map_err(|err| {
                            debug!(
                                "failed to deserialize addresses from validator: {}, err: {}",
                                peer_id.short_str(),
                                err
                            )
                        })
                        .map(|info| (peer_id, info))
                        .ok()
                })
                .collect::<HashMap<_, _>>(),
        )
    }
}
```

libra-node/src/main_node.rs
```diff
@@ -168,20 +169,59 @@ pub fn setup_network(
...
+            .into_prod(identity_pubkey, handshake_version);
+        let advertised_address = config
+            .advertised_address
+            .clone()
+            .into_prod(identity_pubkey, handshake_version);
+
+        // TODO(philiphayes): seed peers should be fully rendered. then we can
+        // also get rid of `network_peers`. we'll need to make some changes to
+        // ops code since it only specifies the base address.
+        let seed_peers: HashMap<_, _> = seed_peers
```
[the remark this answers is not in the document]
phlip9: seems tricky since modifying the `seed_peers` on load would cause it to save the modified seed peers no?

phlip9:
changes with these fixups:

1. listen_address is just the base address (e.g., `"/ip6/::1/tcp/1234"`) inside network. though we return the fully rendered address to the caller in `NetworkBuilder::build()`.
2. reworked `TransportType` => `AuthMode` so that `listen_address` + `AuthMode` fully specifies the transport. for example `"/memory/0"` + `AuthMode::Mutual(key)` => `build_memory_noise_transport()` + `listen_address = "/memory/0/ln-noise-ik/<key.pubkey()>/ln-handshake/0"`.
3. i initially tried to enforce that all dial addrs must include `"/ln-noise-ik/<pubkey>/ln-handshake/0"` but it seems i actually need to convert seed_peers so they're always fully rendered first, since the `network_peers` field i was relying on for the pubkey is usually empty in validator configs.

libra-node/src/main_node.rs
```diff
@@ -131,24 +131,28 @@ pub fn setup_network(
...
+    // TODO(philiphayes): it might make more sense to refactor this so we have a
+    // config per "archetype"?
+
     if config.enable_remote_authentication {
         // If the node wants to run in permissioned mode, it should also have authentication and
         // encryption.
         assert!(
             config.enable_noise,
-            "Permissioned network end-points should use authentication"
+            "Permissioned network end-points must use authentication"
```
mimoo, line 151: :D

libra-node/src/main_node.rs
```diff
@@ -164,19 +168,50 @@ pub fn setup_network(
             .take_private()
             .expect("identity key should be present");
 
+        // TODO(philiphayes): seed peers should be fully rendered. then we can
```
mimoo, line 171: makes me think that we have a bunch of `TODO(abhayb)` as well, not sure who is taking care of them?

network/netcore/src/transport/tcp.rs
```diff
@@ -193,36 +198,32 @@ impl AsyncWrite for TcpSocket {
...
+        [Protocol::Ip6(ip), Protocol::Tcp(port)] => SocketAddr::new((*ip).into(), *port),
+        _ => {
+            return Err(io::Error::new(
+                io::ErrorKind::InvalidInput,
+                format!("Invalid NetworkAddress '{:?}'", addr),
+            ))
+        }
+    };
+
+    Ok((socket_addr, suffix))
```
mimoo, line 226: I think in programming language this pattern is usually called a tail and not a suffix : o

network/network-address/src/lib.rs
```diff
@@ -263,12 +267,79 @@ impl NetworkAddress {
...
+    ///
+    /// `/dns/example.com/tcp/6180` =>
+    /// `/dns/example.com/tcp/6180/ln-noise-ik/<network_pubkey>/ln-handshake/<handshake_version>`
+    // TODO(philiphayes): use handshake version enum
+    pub fn append_prod_protos(
+        self,
+        network_pubkey: x25519::PublicKey,
+        handshake_version: u8,
+    ) -> Self {
+        self.push(Protocol::NoiseIk(network_pubkey))
```
mimoo, line 306: nit: `NoiseIk` -> `NoiseIK`

network/network-address/src/lib.rs
```diff
@@ -263,12 +267,79 @@ impl NetworkAddress {
...
+        self.push(Protocol::PeerIdExchange)
+            .push(Protocol::Handshake(handshake_version))
+    }
+
+    /// A temporary, hacky function to parse out the first `/ln-noise-ik/<pubkey>` from
+    /// a `NetworkAddress`. We can remove this soon, when we move to the interim
+    /// "monolithic" transport model.
+    pub fn find_noise_proto(&self) -> Option<&x25519::PublicKey> {
+        self.0.iter().find_map(|proto| match proto {
+            Protocol::NoiseIk(pubkey) => Some(pubkey),
```
mimoo, line 329: `x25519::PublicKey` are `Copy` so you should be able to return a value not a reference :o

mimoo: what's the "monolithic" transport model?

network/src/validator_network/network_builder.rs
```diff
@@ -66,31 +66,55 @@ pub const MAX_CONCURRENT_NETWORK_REQS: usize = 100;
...
-    MemoryNoise(Option<x25519::PrivateKey>),
-    PermissionlessMemoryNoise(Option<x25519::PrivateKey>),
-    Tcp,
-    TcpNoise(Option<x25519::PrivateKey>),
-    PermissionlessTcpNoise(Option<x25519::PrivateKey>),
+// TODO(philiphayes): remove when transport returns fully rendered network address
+pub const HANDSHAKE_VERSION: u8 = MessagingProtocolVersion::V1 as u8;
+
+#[derive(Debug)]
+pub enum AuthMode {
```
mimoo, line 73: suggest fully spelling `AuthenticationMode`

network/src/validator_network/network_builder.rs
```diff
@@ -66,31 +66,55 @@ pub const MAX_CONCURRENT_NETWORK_REQS: usize = 100;
...
+    /// Inbound and outbound connections are unauthenticated and unencrypted, using
+    /// a simple PeerId exchange protocol to share each peer's PeerId in-band.
+    /// This mode is obviously insecure and should never be used in production.
+    // TODO(philiphayes): remove? gate with cfg(test)? should never use this in prod.
+    Unauthed,
```
mimoo, line 78: same here, `Unauthenticated`

network/src/validator_network/network_builder.rs
```diff
@@ -455,48 +484,59 @@ impl NetworkBuilder {
...
-            TransportType::PermissionlessMemoryNoise(ref mut keys) => {
-                let keys = keys.take().expect("Identity keys not set");
-                self.build_with_transport(build_unauthenticated_memory_noise_transport(
-                    keys,
-                    supported_protocols,
-                ))
-            }
-            TransportType::Tcp => {
-                self.build_with_transport(build_tcp_transport(peer_id, supported_protocols))
+        let auth_mode = self.auth_mode.take().expect("Authentication Mode not set");
```
mimoo, line 493: random idea: we could assert somewhere that the auth_mode is mutual on validator network, server side on peer network, and unauthenticated only if the config is marked as "test" (we could have the first field of the config be something set to production or test)

network/src/validator_network/network_builder.rs
```diff
@@ -66,31 +66,55 @@ pub const MAX_CONCURRENT_NETWORK_REQS: usize = 100;
...
+    /// Inbound and outbound connections are unauthenticated and unencrypted, using
+    /// a simple PeerId exchange protocol to share each peer's PeerId in-band.
+    /// This mode is obviously insecure and should never be used in production.
+    // TODO(philiphayes): remove? gate with cfg(test)? should never use this in prod.
+    Unauthed,
+    /// Inbound and outbound connections are secured with NoiseIK; however, only
+    /// clients/dialers will authenticate the servers/listeners. More specifically,
+    /// dialers will pin the connection to a specific, expected pubkey while
+    /// listeners will accept any inbound dialer's pubkey.
+    ClientOnly(x25519::PrivateKey),
```
mimoo, line 83: that should be `ServerOnly`

network/src/validator_network/network_builder.rs
```diff
@@ -455,48 +484,59 @@ impl NetworkBuilder {
...
-                self.build_with_transport(build_unauthenticated_memory_noise_transport(
-                    keys,
-                    supported_protocols,
-                ))
-            }
-            TransportType::Tcp => {
-                self.build_with_transport(build_tcp_transport(peer_id, supported_protocols))
+        let auth_mode = self.auth_mode.take().expect("Authentication Mode not set");
+
+        // formatting the listen address here is a temporary hack until `transport.rs`
```
mimoo, line 495: lots of these temporary hack :p

testsuite/generate-format/tests/staged/network.yaml
```diff
@@ -103,6 +103,8 @@ Protocol:
     8:
       Handshake:
         NEWTYPE: U8
+    9:
+      PeerIdExchange: UNIT
```
mimoo, line 107: what's a UNIT oO?

mimoo:
Thanks for pushing that out! Make sure to look at my comments, and I guess squash all these fix commits :o 

@bors-libra delegate+

libra-node/src/main_node.rs
phlip9: i think abhay mostly did plain `TODO:` : D

network/network-address/src/lib.rs
phlip9: been sketching this out... probably will look something like this 

```rust
impl MemoryNoiseMutualAuthTransport {
    pub fn new(
        identity_key: x25519::PrivateKey,
        trusted_peers: TrustedPeersMap,
        application_protocols: SupportedProtocols,
    ) -> Self {
        let mut own_handshake = HandshakeMsg::new();
        own_handshake.add(SUPPORTED_MESSAGING_PROTOCOL, application_protocols);

        Self {
            noise_config: Arc::new(NoiseConfig::new(identity_key)),
            own_handshake,
            trusted_peers,
        }
    }

    pub async fn dial(
        &self,
        addr: NetworkAddress,
    ) -> io::Result<Connection<NoiseSocket<MemorySocket>>> {
        use libra_network_address::Protocol::*;
        let origin = ConnectionOrigin::Outbound;

        let (port, pubkey, version) = match addr.as_slice() {
            [Memory(port), NoiseIk(pubkey), Handshake(version)] => (*port, *pubkey, *version),
            _ => {
                return Err(io::Error::new(
                    io::ErrorKind::InvalidData,
                    format!(
                        "Unexpected NetworkAddress: '{}', expected: \
                         '/memory/<port>/ln-noise-ik/<pubkey>/ln-handshake/<version>'",
                        addr
                    ),
                ))
            }
        };
        verify_handshake_version_supported(version)?;

        let socket = MemorySocket::connect(port)?;

        let (remote_static_key, socket) =
            self.noise_config.upgrade_connection(socket, origin).await?;
        verify_pubkeys_match(&remote_static_key, &pubkey)?;

        let peer_id = identity_key_to_peer_id(&self.trusted_peers, &remote_static_key)
            .ok_or_else(|| io::Error::new(io::ErrorKind::Other, "Not a trusted peer"))?;

        perform_handshake(peer_id, socket, addr, origin, &self.own_handshake).await
    }

    pub fn listen_on(&self, addr: NetworkAddress) -> io::Result<()> {
        use libra_network_address::Protocol::*;

        let origin = ConnectionOrigin::Inbound;

        let port = match addr.as_slice() {
            [Memory(port)] => *port,
            _ => {
                return Err(io::Error::new(
                    io::ErrorKind::InvalidData,
                    format!(
                        "Unexpected NetworkAddress: '{}', expected: '/memory/<port>'",
                        addr
                    ),
                ))
            }
        };

        let mut listener = MemoryListener::bind(port)?;
        let port = listener.local_addr();

        Ok(listener.incoming().and_then(|socket| async move {
            let (remote_static_key, socket) =
                self.noise_config.upgrade_connection(socket, origin).await?;
            verify_pubkeys_match(&remote_static_key, &pubkey)?;

            let peer_id = identity_key_to_peer_id(&self.trusted_peers, &remote_static_key)
                .ok_or_else(|| io::Error::new(io::ErrorKind::Other, "Not a trusted peer"))?;

            perform_handshake(peer_id, socket, addr, origin, &self.own_handshake).await
        }))

        todo!()
    }
}
```

testsuite/generate-format/tests/staged/network.yaml
phlip9: yeah `"/ln-peerid-ex` has no arguments, hence "()" is the enum value I guess

I makes sense that these two examples are isomorphic

```rust
enum Foo {
    Bar,
}

enum Foo {
    Bar(()),
}
```

network/src/validator_network/network_builder.rs
phlip9: 😭

network/network-address/src/lib.rs
mimoo: ah yes! You already explained that to me :) I'm all for it!

mimoo: let me get my noise PR in first though :P (I already verify public keys in the noise upgrade_connection itself)

network/src/validator_network/network_builder.rs
phlip9: yeah, I think the configs as a whole need some broader refactoring work :'(

also to help reduce some of these invalid states

PR #3852, opened 2020-05-13
change type_tag_parser to pub

liangping:
There's pub func `parse_type_tags` in type_tag_parser,
It's useful for developer to parse type tags in their third-party tools


davidiw:
Hey language folks, could you take a look at this. It seems pretty reasonable to me, but I know you're in an effort to straighten out your crates.

dariorussi:
it looks fine to me, letting @vgao1996 comment on this

PR #3853, opened 2020-05-13
[move-prover] Disable two-state invariants temporarily

shazqadeer:
## Motivation

This PR temporarily disables two-state invariants in the test cases to make it easier to land 
the new memory model. Specifically:
1. Two-state invariants in invariants.move and nested-invariants.move have been commented out. invariants.exp is updated and the old file checked in as invariants.exp.old.  nested-invariants.exp did not change because the two-state invariants inside it did not result in errors.
2. In all other files, the use of two-state invariants was redundant since it was already covered by invariant pack/unpack.  Deletion resulted in shifted line numbers and .exp files were updated.
3. Disabled VarUpdate and InvariantUpdate in annotations in spec-lang/src/ast.rs. Consequently, commented out "invariant update" in the unit tests there.  Two .exp files changed.  I checked in their old version: invariants_err.exp.old, variables_err.exp.old.

After landing the new memory model, we will revisit two-state invariants.

Yes.

## Test Plan

Existing tests.

### DIFF
--- a/language/move-prover/spec-lang/src/ast.rs
+++ b/language/move-prover/spec-lang/src/ast.rs
@@ -82,16 +82,13 @@ impl ConditionKind {
     /// Returns true if this condition is allowed on a public function declaration.
     pub fn allowed_on_public_fun_decl(&self) -> bool {
         use ConditionKind::*;
-        matches!(
-            self,
-            Requires | RequiresModule | AbortsIf | Ensures | VarUpdate(..)
-        )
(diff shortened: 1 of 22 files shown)


shazqadeer:
> This looks great, but can you also please deactivate those in the type checker? This will also give you confidence that you do not missed something.
> 
> It's easy. Go to `spec-lang/src/ast.rs`'. Search for mentioning of `VarUpdate` and `InvariantUpdate` Remove it from all functions `allowed_XXX`. The context checker will then produce errors when two-state invariants are used.

Done.

PR #3865, opened 2020-05-13
[move vm] Added signer type

tnowacki:
## Motivation

- Added the primitive `signer` type as described in #3679 
  - Name is `signer` instead of `sender` based on in person discussions
- Operations over the type will come in a later PR
- Long term this system will replace `MoveToSender` and `GetTxnSenderAddress`

## Test Plan

- Updated prop tests
- Added MVIR tests to verify `signer` is treated as a resource


language/move-vm/types/src/values/values_impl.rs
```diff
@@ -318,6 +319,8 @@ impl ValueImpl {
             U128(x) => U128(*x),
             Bool(x) => Bool(*x),
             Address(x) => Address(*x),
+            // REVIEW copying resource?
+            Signer(_) => Invalid,
```
vgao1996, line 323: Good point. Although instead of Invalid you should probably Err as "Invalid" really means "unassigned local". I'd say let's make a real copy for now and keep the comment. I'm planning to review the whole vm story soon and will think about if and how we'll introduce protection for resources in general. (Remember right now resource structs can also be copied.)

language/ir-testsuite/tests/signer/invalid_move_to_sender.mvir
```diff
@@ -0,0 +1,17 @@
+module M {
+    t(s: signer) {
+        move_to_sender<signer>(move(s));
+        return;
+    }
+}
+// check: ParserError: Invalid Token
```
sblackshear, line 7: Interesting example! 

I guess a broader question is what the bytecode verifier should do with this code (i.e.,  should it accept or reject `move_to_sender<signer>`, `move_from<signer>`, `borrow_global<signer>`, etc). Is it possible to write these in bytecode (even if the IR parser won't allow them)? 

If so, it seems like the bytecode verifier ought to reject them (at least conceptually) because these instructions are supposed to work on a resource declared in the current module, and  `signer` is always declared elsewhere (by construction).

language/ir-testsuite/tests/signer/invalid_move_to_sender.mvir
```diff
@@ -0,0 +1,17 @@
...
+
+//! new-transaction
+module M {
+    resource R { s: signer }
+    t(s: signer) {
+        move_to_sender<R>(move(s));
+        return;
+    }
+}
+// check: MOVETOSENDER_TYPE_MISMATCH_ERROR
```
sblackshear, line 17: Maybe another test to show that `move_to_sender<R>(R { s: move(signer) })` works?

language/ir-testsuite/tests/signer/is_resource_transitive.mvir
```diff
@@ -0,0 +1,13 @@
...
+    }
+
+    t(s: signer) {
+        let x: Self.S<signer>;
+        x = S<signer> { f: move(s) };
+        return;
+    }
+}
+
+// check: UNSAFE_RET_UNUSED_RESOURCES
```
sblackshear, line 13: Nice test!

language/move-prover/src/boogie_helpers.rs
```diff
@@ -181,6 +183,8 @@ fn boogie_well_formed_expr_impl(
             PrimitiveType::Num => conds.push(format!("$IsValidNum({})", name)),
             PrimitiveType::Bool => conds.push(format!("is#Boolean({})", name)),
             PrimitiveType::Address => conds.push(format!("is#Address({})", name)),
+            // TODO fix this for a real boogie check
+            PrimitiveType::Signer => conds.push(format!("is#Address({})", name)),
```
sblackshear, line 187: A `panic!` here might be easier for Prover devs (cc @wrwg)?

language/move-vm/types/src/values/values_impl.rs
sblackshear: I agree with Victor on this--allowing the copy seems cleanest given that we do that for structs.

language/vm/src/file_format.rs
```diff
@@ -676,7 +680,7 @@ impl SignatureToken {
     pub fn is_primitive(&self) -> bool {
```
sblackshear, line 680: It looks like the only place this function is used is in checking for valid tx script signatures:

```
/// This function checks the extra requirements on the signature of the main function of a script.
pub fn verify_main_signature(script: &CompiledScript) -> VMResult<()> {
    let arguments = script.signature_at(script.as_inner().parameters);
    for arg_type in &arguments.0 {
        if !(arg_type.is_primitive()
            || *arg_type == SignatureToken::Vector(Box::new(SignatureToken::U8)))
        {
            return Err(VMStatus::new(StatusCode::INVALID_MAIN_FUNCTION_SIGNATURE));
        }
    }
    Ok(())
}
```

If this is the only use-case, it might make sense for us to repurpose this to `is_valid_tx_script_parameter` or similar?  Eventually, we want to allow tx scripts to accept `vector<u64>` etc. and `&signer`, but presumably not `signer`.

language/ir-testsuite/tests/signer/keyword.mvir
```diff
@@ -0,0 +1,7 @@
+module M {
+    foo(signer: address) {
+        return;
+    }
+}
+
+// check: ParserError: Invalid Token
```
sblackshear, line 7: Couple more tests for resource-y ness of signer that might be useful:

```
copy_signer_bad1(s:  signer): (signer * signer) {
  return (copy(s), move(s))
}

copy_signer_bad2(v: vector<signer>): vector<signer> {
  let s: signer;
  s = *Vector::borrow_loc(&v, 0);
}
```

language/vm/src/file_format.rs
dariorussi: even better would be to remove it completely and add that to `verify_main_signature` which as mentioned will change significantly

language/ir-testsuite/tests/signer/invalid_move_to_sender.mvir
tnowacki: The bytecode verifier can't reject them since it isn't expressible at the file format level. The instructions take a `StructDefinitionIndex`, so it will always be safe

language/ir-testsuite/tests/signer/keyword.mvir
tnowacki: Ah, yes forgot bout copies. I'll add these

language/move-prover/src/boogie_helpers.rs
tnowacki: Yes, I was unsure what to do here for the prover, let me know and I'll do whatever y'all suggest
@junkil-park @wrwg @DavidLDill @shazqadeer

language/vm/src/file_format.rs
tnowacki: I'll take a look in just killing it

language/move-prover/src/boogie_helpers.rs
wrwg: Because compilation of all stdlib sources to boogie is part of our tests (and the MIRAI bot), a panic might be actually not the best choice.

language/ir-testsuite/tests/signer/invalid_move_to_sender.mvir
sblackshear: Ah, nice! I knew it used to be that way but didn't realize this is still true after the file format refactor. Excellent news.

language/bytecode-verifier/src/verifier.rs
```diff
@@ -165,11 +165,26 @@ impl ScriptAccess for VerifiedScript {
 
 /// This function checks the extra requirements on the signature of the main function of a script.
 pub fn verify_main_signature(script: &CompiledScript) -> VMResult<()> {
+    fn is_valid_arg_type(arg_type: &SignatureToken) -> bool {
+        use SignatureToken as S;
+        match arg_type {
+            S::Bool | S::U8 | S::U64 | S::U128 | S::Address => true,
+            S::Vector(inner) => match &**inner {
+                S::U8 => true,
+                _ => false,
```
sblackshear, line 174: Nit: exhaustive pattern matching

language/ir-testsuite/tests/signer/invalid_move_to_sender.mvir
sblackshear: Nice! I didn't realize this was still true after the big file format refactor; excellent news.

Might be better to use storage op other than `move_to_sender` so we don't have to change this when we eliminate `move_to_sender`.

tnowacki: I'm going to keep it at move_to_sender (even though it will need to be refactored) as the test is a bit closer to what I want (trying to actually move a `signer` into global storage directly)

PR #3870, opened 2020-05-13
[config] Merge service and libranet port randomization

davidiw:
Not really sure why these are separated since we likely need them randomized for the same reasons.

### DIFF
--- a/network/onchain-discovery/src/test.rs
+++ b/network/onchain-discovery/src/test.rs
@@ -165,7 +165,7 @@ fn read_validator_set(libra_db: &Arc<dyn DbReader>) -> ValidatorSet {
 fn gen_configs(count: usize) -> Vec<NodeConfig> {
     config_builder::ValidatorConfig::new()
         .validators(count)
-        .build_common(true, false)
+        .build_common(false)
         .unwrap()
         .0
 }
--- a/config/config-builder/src/full_node_config.rs
+++ b/config/config-builder/src/full_node_config.rs
@@ -174,11 +174,10 @@ impl FullNodeConfig {
 
         // Don't randomize ports. Until we better separate genesis generation,
         // we don't want to accidentally randomize initial discovery set addresses.
-        let randomize_service_ports = false;
-        let randomize_libranet_ports = false;
+        let randomize_validator_ports = false;
         let (validator_configs, faucet_key) = self
             .validator_config
-            .build_common(randomize_service_ports, randomize_libranet_ports)?;
+            .build_common(randomize_validator_ports)?;
         let validator_config = validator_configs.first().ok_or(Error::NoConfigs)?;
 
         let mut rng = StdRng::from_seed(self.full_node_seed);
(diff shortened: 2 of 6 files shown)


phlip9:
> Not really sure why these are separated since we likely need them randomized for the same reasons.

iirc this is a side effect of cluster test using config generator independently on each V and VFN

from blame:

> Revert "Revert "[genesis][breaking] Integrate config gen w/ genesis discovery set""
> 
> This reverts commit 40dcdc313b8e883d47a04bb6be8e38ec81686034.
> 
> The original reverted commit missed the full node config generation
> for cluster test, which was accidentally randomizing the ports for the
> genesis DiscoverySet. This lead to an unintentional fork as full nodes
> had a different genesis than their validators.

it seems like config/management should replace the independent generation via config/generator for genesis creation.

for local swarm, we would have to get available ports in the very beginning, alongside sampling the keypairs. we then use these to derive genesis, which we later pass to each V and VFN so they listen on their assigned port.

for cluster test, each node actually advertises a DNS name like "val1.<cluster-test-aws-thing>.com" and should all listen on the same port, so we should have all we need for genesis creation via k8s+config

davidiw:
> > Not really sure why these are separated since we likely need them randomized for the same reasons.
> 
> iirc this is a side effect of cluster test using config generator independently on each V and VFN
> 
> from blame:
> 
> > Revert "Revert "[genesis][breaking] Integrate config gen w/ genesis discovery set""
> > This reverts commit [40dcdc3](https://github.com/libra/libra/commit/40dcdc313b8e883d47a04bb6be8e38ec81686034).
> > The original reverted commit missed the full node config generation
> > for cluster test, which was accidentally randomizing the ports for the
> > genesis DiscoverySet. This lead to an unintentional fork as full nodes
> > had a different genesis than their validators.
> 
> it seems like config/management should replace the independent generation via config/generator for genesis creation.
> 
> for local swarm, we would have to get available ports in the very beginning, alongside sampling the keypairs. we then use these to derive genesis, which we later pass to each V and VFN so they listen on their assigned port.
> 
> for cluster test, each node actually advertises a DNS name like "val1..com" and should all listen on the same port, so we should have all we need for genesis creation via k8s+config

Right! Though the fact is the only cases where we had a mismatch on port randomization was within integration tests, otherwise they were the same value (either true true or false false). So without digging into the current behavior of cluster test or other external tests, this PR does not change their behaviors.

network/onchain-discovery/src/test.rs
```diff
@@ -165,7 +165,7 @@ fn read_validator_set(libra_db: &Arc<dyn DbReader>) -> ValidatorSet {
 fn gen_configs(count: usize) -> Vec<NodeConfig> {
     config_builder::ValidatorConfig::new()
         .validators(count)
-        .build_common(true, false)
```
phlip9, line 168: makes sense. now that we're just using `LibraDB` directly w/o a network service in between, we don't need to randomize service ports. originally, the problem was that, since unit tests run in parallel, w/o port randomization the different test threads would each try to spin up a storage service on the same port

phlip9:
cool, lgtm! hopefully CT comes up positive : p

@bors-libra delegate+

network/onchain-discovery/src/test.rs
davidiw: Yes, I think if this were not the case, we would have to figure out how to spread this information. I think getting config/management to replace config-builder is not an overnight project :/

PR #3872, opened 2020-05-14
[fuzzing] adding storage structured fuzzer

mimoo:
This PR adds a new structured fuzzer for storage.
Meaning we can now fuzz functions that expect a struct as input (as opposed to a bytearray).

The way it works is that the fuzzer gives the function a random string of bytes,
and the function pretends that the random bytearray is the ouptut of an RNG it can use to construct the struct.
This is called a "passthrough" RNG and we use the proptest library to build such structs from random data.

The current behavior of a passthrough RNG (in the proptest library we depend on)
is to crash if the input is not large enough to build the struct.
This is not ideal, so to make it work the PR also patches proptest to make pass-through RNGs return 0s after they depleted the input given to them.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -146,3 +146,6 @@ lto = 'thin'
 
 [profile.bench]
 debug = true
+
+[patch.crates-io]
+proptest = { git = "https://github.com/mimoo/proptest.git", branch = "fuzzing" }
--- a/testsuite/libra-fuzzer/src/fuzz_targets/storage_save_blocks.rs
+++ b/testsuite/libra-fuzzer/src/fuzz_targets/storage_save_blocks.rs
@@ -1,0 +1,48 @@
+// Copyright (c) The Libra Core Contributors
+// SPDX-License-Identifier: Apache-2.0
+
+use crate::FuzzTargetImpl;
+use libra_proptest_helpers::ValueGenerator;
+use libradb::{test_helper::arb_blocks_to_commit, test_save_blocks_impl};
+use proptest::{
+    strategy::{Strategy, ValueTree},
+    test_runner::{self, TestRunner},
+};
+use rand::RngCore;
+
+#[derive(Clone, Debug, Default)]
+pub struct StorageSaveBlocks;
+
+impl FuzzTargetImpl for StorageSaveBlocks {
+    fn name(&self) -> &'static str {
+        module_name!()
+    }
+
+    fn description(&self) -> &'static str {
+        "Storage save blocks"
+    }
+
+    fn generate(&self, _idx: usize, _gen: &mut ValueGenerator) -> Option<Vec<u8>> {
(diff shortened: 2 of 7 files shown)


msmouse:
stamp @bors-libra delegate+

testsuite/libra-fuzzer/src/fuzz_targets/storage_save_blocks.rs
```diff
@@ -0,0 +1,48 @@
...
+
+    fn description(&self) -> &'static str {
+        "Storage save blocks"
+    }
+
+    fn generate(&self, _idx: usize, _gen: &mut ValueGenerator) -> Option<Vec<u8>> {
+        let mut output = vec![0u8; 4096];
+        let mut rng = rand::thread_rng();
+        rng.fill_bytes(&mut output);
+        Some(output)
```
sunshowers, line 29: Hmm, so does this generate any arbitrary sequence of bytes? Are all such sequences valid?

mimoo: yes (uses thread_rng) and yes (any random output is valid for a strategy), but a better question is: is this enough or too much output for the strategy?

For now I have no idea. 

A better way would be to ask proptest to generate a bunch of values, and then dump the  random bytes that it used to generate them, unfortunately there doesn't seem to be any API for that.

Some ideas:

* push a PR upstream to create this API
* generate a value using this `output` with proptest
  - if it couldn't create a value, try again with a random output twice the size (and again and again)
  - if it worked, see how much bytes are left in the `output` and discard them. 

what do you think?

Cargo.toml
```diff
@@ -146,3 +146,6 @@ lto = 'thin'
 
 [profile.bench]
 debug = true
+
+[patch.crates-io]
+proptest = { git = "https://github.com/mimoo/proptest.git", branch = "fuzzing" }
```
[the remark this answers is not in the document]
mimoo: upstream author told me it would get in there by end of the week, so I was hopping that this would be very temporary

testsuite/libra-fuzzer/src/fuzz_targets/storage_save_blocks.rs
sunshowers: Would the pass through API work for most of this?

mimoo: I don't think so, since we can't know how much of the seed was used. If the passthrough crashed (the behavior before my patch) we could do something sketchy where we generate a seed of length X, catch a panic, start again with length X+1, and so on.

metajack:
Did you discuss with upstream? I hope we planning on forking proptest long term.

mimoo:
yup, he told me he would accept my PR 2 weeks ago, never happened :D

PR #3878, opened 2020-05-14
[management] Improve test framework

davidiw:
This is an incremental step toward building smoke tests that leverage this framework -- we needed to be able to generate a genesis using an integration test friendly storage solution and the framework had to be reusable. This PR does that an just a tad more:

* Fixup constants to match the original style -- yes I am being anal here
* Swap json in for toml, toml is a really bad format for the way we are storing data in secure-storage and is actually broken for Transaction
* Replace vault for disk in the tests for management
* Move the API interaction for tests in management accessible (StorageHelper)

### DIFF
--- a/config/management/src/genesis.rs
+++ b/config/management/src/genesis.rs
@@ -1,12 +1,7 @@
 // Copyright (c) The Libra Core Contributors
 // SPDX-License-Identifier: Apache-2.0
 
-use crate::{
-    error::Error,
-    layout::Layout,
-    management_constants::{COMMON_NS, LAYOUT, VALIDATOR_CONFIG},
-    SingleBackend,
-};
+use crate::{error::Error, layout::Layout, SingleBackend};
 use libra_crypto::ed25519::Ed25519PublicKey;
 use libra_global_constants::{ASSOCIATION_KEY, OPERATOR_KEY};
 use libra_secure_storage::Storage;
@@ -62,11 +57,11 @@ impl Genesis {
         let mut common_config = self.backend.backend.clone();
         common_config
             .parameters
-            .insert("namespace".into(), COMMON_NS.into());
+            .insert("namespace".into(), crate::constants::COMMON_NS.into());
         let common: Box<dyn Storage> = common_config.try_into()?;
 
         let layout = common
-            .get(LAYOUT)
+            .get(crate::constants::LAYOUT)
             .map_err(|e| Error::RemoteStorageReadError(e.to_string()))?
             .value
             .string()
@@ -92,7 +87,7 @@ impl Genesis {
                 .map_err(|e| Error::RemoteStorageReadError(e.to_string()))?;
 
             let txn = validator
-                .get(VALIDATOR_CONFIG)
+                .get(crate::constants::VALIDATOR_CONFIG)
                 .map_err(|e| Error::RemoteStorageReadError(e.to_string()))?
                 .value;
             let txn = txn.transaction().unwrap();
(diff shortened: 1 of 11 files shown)


config/management/src/genesis.rs
```diff
@@ -62,11 +57,11 @@ impl Genesis {
         let mut common_config = self.backend.backend.clone();
         common_config
             .parameters
-            .insert("namespace".into(), COMMON_NS.into());
+            .insert("namespace".into(), crate::constants::COMMON_NS.into());
```
JoshLind, line 60: Interesting.. I originally took this approach when I created the global constants crate, but I eventually decided to import the full path for two reasons:
(1) quoting from the rust book [1] : "On the other hand, when bringing in structs, enums, and other items with use, it’s idiomatic to specify the full path. " I took this to mean constants as well (although I've yet to find a concrete argument for/against this specific issue).
(2) it seemed ugly to me to have to import each constant using the full path, especially for files that use the same constant more than once..

It would be nice if there was some form of consensus on the issue (although I suspect not)? Leaving up to your discretion!

[1] https://doc.rust-lang.org/book/ch07-04-bringing-paths-into-scope-with-the-use-keyword.html#creating-idiomatic-use-paths

config/management/src/storage_helper.rs
```diff
@@ -0,0 +1,212 @@
...
+    OWNER_KEY, PREFERRED_ROUND, VALIDATOR_NETWORK_KEY, WAYPOINT,
+};
+use libra_network_address::NetworkAddress;
+use libra_secure_storage::{NamespacedStorage, OnDiskStorage, Policy, Storage, Value};
+use libra_types::{account_address::AccountAddress, transaction::Transaction};
+use std::fs::File;
+use structopt::StructOpt;
+
+pub struct StorageHelper {
+    temppath: libra_temppath::TempPath,
```
JoshLind, line 17: nit: why not just import TempPath?

config/management/src/storage_helper.rs
```diff
@@ -0,0 +1,212 @@
...
+                management
+                association-key
+                --local backend={backend};\
+                    path={path};\
+                    namespace={local_ns}
+                --remote backend={backend};\
+                    path={path};\
+                    namespace={remote_ns}\
+            ",
+            backend = crate::secure_backend::DISK,
```
JoshLind, line 70: nit: why not import directly?

config/management/src/lib.rs
```diff
@@ -409,81 +405,86 @@ pub mod tests {
...
-        initialize_storage(local.as_mut());
+        helper.initialize(local_ns.clone());
+        let local = helper.storage(local_ns.clone());
         let local_key = local.get_public_key(key_name).unwrap().public_key;
 
-        let mut remote = default_storage(remote_ns.clone());
-        remote.reset_and_clear().unwrap();
-
-        let output_key = op(&local_ns, &remote_ns).unwrap();
+        println!("{}", helper.path_string());
```
JoshLind, line 485: Guessing this is a loose end from debugging?

config/management/src/genesis.rs
davidiw: I share your thoughts... my refinements is does importing it add more lines of code, if not don't import it.

config/management/src/storage_helper.rs
davidiw: this doesn't add lines of code, so the extra import seems unnecessary.

config/management/src/storage_helper.rs
davidiw: same, plus I guess I kinda see constants more in line with functions

config/management/src/lib.rs
davidiw: yuck, woops

davidiw:
Thanks for the late night review!!

PR #3879, opened 2020-05-14
[language][easy] Refactor a few iterator / Result patterns in language

huitseeker:
Please refer to the individual commit messages, which are sorted by pattern.

The collect-to-result instances may offer a way to delete unwraps (if
it's proper to ignore failures), but for now they keep the original
behavior.

### DIFF
--- a/language/stdlib/src/lib.rs
+++ b/language/stdlib/src/lib.rs
@@ -48,8 +48,9 @@ static STAGED_MOVELANG_STDLIB: Lazy<Vec<VerifiedModule>> = Lazy::new(|| {
     let modules = lcs::from_bytes::<Vec<Vec<u8>>>(STAGED_STDLIB_BYTES)
         .unwrap()
         .into_iter()
-        .map(|bytes| CompiledModule::deserialize(&bytes).unwrap())
-        .collect();
+        .map(|bytes| CompiledModule::deserialize(&bytes))
+        .collect::<Result<_, _>>()
+        .unwrap();
     batch_verify_modules(modules)
 });
 
--- a/language/vm/src/access.rs
+++ b/language/vm/src/access.rs
@@ -114,10 +114,11 @@ pub trait ModuleAccess: Sync {
     fn function_def_at(&self, idx: FunctionDefinitionIndex) -> &FunctionDefinition {
         let result = &self.as_module().as_inner().function_defs[idx.into_index()];
         assumed_postcondition!(result.function.into_index() < self.function_handles().len()); // invariant
-        assumed_postcondition!(match &result.code {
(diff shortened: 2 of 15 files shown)


language/vm/src/access.rs
```diff
@@ -114,10 +114,11 @@ pub trait ModuleAccess: Sync {
...
         assumed_postcondition!(result.function.into_index() < self.function_handles().len()); // invariant
-        assumed_postcondition!(match &result.code {
-            Some(code) => code.locals.into_index() < self.signatures().len(),
-            None => true,
-        }); // invariant
+        assumed_postcondition!(result
+            .code
+            .as_ref()
+            .map_or(true, |code| code.locals.into_index()
+                < self.signatures().len())); // invariant
```
tnowacki, line 121: Oof, rustfmt really looks awful here. Any way to improve this?

language/stdlib/src/lib.rs
```diff
@@ -48,8 +48,9 @@ static STAGED_MOVELANG_STDLIB: Lazy<Vec<VerifiedModule>> = Lazy::new(|| {
     let modules = lcs::from_bytes::<Vec<Vec<u8>>>(STAGED_STDLIB_BYTES)
         .unwrap()
         .into_iter()
-        .map(|bytes| CompiledModule::deserialize(&bytes).unwrap())
-        .collect();
+        .map(|bytes| CompiledModule::deserialize(&bytes))
+        .collect::<Result<_, _>>()
+        .unwrap();
```
tnowacki, line 53: I think the previous way was a bit easier to read. But I liked the collecting into Result in the other cases

dariorussi:
fwiw, I would really appreciate if no change of this kind were ever pushed to `language`.
Having said so `language` is a lot of things ;) and I don't want to speak for everybody so whatever... I am sure there are a lot of great fans of those changes... I think is best if I stay out of this as at the end is not a big deal either way.
Sorry @huitseeker, take this for what is worth, and I'm fine with @tnowacki and @vgao1996 to drive this.

tnowacki:
If I had to take my pick, yeah I'd agree with @dariorussi and I'd never want to see PRs like this. I find them invasive in a way that often makes the code harder to read; on top of that, they often cause merge conflicts with very large PRs we are working on. 

That being said, I will try to compromise if the rest of Libra really wants these style changes for whatever reason.

tnowacki:
I think the only one I'd want to keep in this PR is the one `json-rpc/src/methods.rs`. This converts an `unwrap` into a `?`, which seems like a win. 

All the rest fit my previous complaint

huitseeker:
Fair enough. Those are made to be useful. If they aren't, I get that.

PR #3887, opened 2020-05-14
[client] Add a command to upgrade stdlib modules

runtian-zhou:
## Motivation

Implement a command to upgrade/rebuild the stdlib modules on chain.

Yes

## Test Plan

I ran a one shot test as following:
1. Run a libra_swarm locally via 
`cargo run -p libra-swarm -- -s --enable-logging`
2. Create two accounts and transfer some funds between them.
3. Add an assertion such that deposit will abort when the transfer amount is equal to a magic number.
4. Run `dev upgrade_stdlib`
5. Transfer the same amount and see payment will only fail when you try to send that special transfer amount.

### DIFF
--- a/language/transaction-builder/src/lib.rs
+++ b/language/transaction-builder/src/lib.rs
@@ -4,15 +4,20 @@
 #![forbid(unsafe_code)]
 
 use libra_types::{
+    access_path::AccessPath,
     account_address::AccountAddress,
     block_metadata::BlockMetadata,
     on_chain_config::{LibraVersion, VMPublishingOption},
-    transaction::{authenticator::AuthenticationKey, Script, Transaction, TransactionArgument},
+    transaction::{
+        authenticator::AuthenticationKey, ChangeSet, Script, Transaction, TransactionArgument,
+    },
+    write_set::{WriteOp, WriteSetMut},
 };
 use mirai_annotations::*;
 use move_core_types::language_storage::TypeTag;
 use std::convert::TryFrom;
-use stdlib::transaction_scripts::StdlibScript;
+use stdlib::{transaction_scripts::StdlibScript, StdLibOptions};
+use vm::access::ModuleAccess;
 
 fn validate_auth_key_prefix(auth_key_prefix: &[u8]) {
     let auth_key_prefix_length = auth_key_prefix.len();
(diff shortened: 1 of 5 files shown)


sblackshear:
> This command is not tested right now. The reason is that the command will only rebuild the existing staged stdlib and thus cannot be tested at this time as it's going to be guaranteed to be equivalent to the on chain bytes. To test this feature we need to manually change some modules and rebuild the stdlib.

Did you try a manual test on a local instance of Libra?

language/transaction-builder/src/lib.rs
```diff
@@ -564,3 +569,26 @@ encode_txn_script! {
     script: RemoveParentAccount,
     doc: "Removes the parent account at `parent_address`. It can be recertified in the future however."
 }
+
+//...........................................................................
+// WriteSets
+//...........................................................................
+
+pub fn encode_stdlib_upgrade_transaction(option: StdLibOptions) -> ChangeSet {
```
sblackshear, line 577: Would it be possible to reuse the code that genesis uses to build this `ChangeSet` to ensure that the implementations don't diverge?

language/transaction-builder/src/lib.rs
```diff
@@ -564,3 +569,26 @@ encode_txn_script! {
...
+    let mut write_set = WriteSetMut::new(vec![]);
+    let stdlib = stdlib::stdlib_modules(option);
+    for module in stdlib {
+        let mut bytes = vec![];
+        module
+            .serialize(&mut bytes)
+            .expect("Failed to serialize module");
+        write_set.push((
+            AccessPath::code_access_path(&module.self_id()),
```
sblackshear, line 586: If the on-chain stdlib previously contained a module named `Foo`, but `Foo` is deleted in the local stdlib, will `Foo` still exist on-chain after running this upgrade transaction?

language/transaction-builder/src/lib.rs
jolestar: The stdlib upgrade is only updating code, or also update resource? genesis code also generates the resource.

sblackshear: This approach only updates code. But you raise an interesting point: it seems very important to check that data format of existing resources has not been changed.

language/transaction-builder/src/lib.rs
runtian-zhou: I guess there's no way for us to capture that just in the `transaction-builder` right? As there's no way for us to know what the "old" stdlib modules are.

runtian-zhou:
I really enjoy the versioning conversation over here. It really seems to me that we need to checkin every version of stdlib.mv file into the codebase in the future so that we can build the tooling around deprecating old modules and making sure the formats remains the same.

language/transaction-builder/src/lib.rs
sblackshear: Not in `transaction-builder` directly, but perhaps we could read this list from the blockchain in the client + pass it to `transaction-builder` (or do the check there)?

dariorussi:
I guess the question raised is whether any push of stdlib should be blind or go through rigorous verification steps. Right now it seems to me to be a bit too lax.
Thoughts?

sblackshear:
> I guess the question raised is whether any push of stdlib should be blind or go through rigorous verification steps. Right now it seems to me to be a bit too lax.
> Thoughts?

I agree--there are a lot of scary or hard-to-debug issues that a blind update could cause. Good motivation for us to define backward compatibility!

runtian-zhou:
> Run verifier too?
Just checked the code. `build_stdlib` function will verify the modules emitted so we don't need to invoke verifier again.

Issue #3868, opened 2020-05-13
[cluster_test] Client side latency statistics


andll:
One of functions of cluster test is to generate transaction load and gather statistics about it.
Transaction generation can be tried on libra swarm cluster locally, you need to run cluster with `cargo run -p libra-swarm` and it will print out command to run cluster test locally.

As you may notice, cluster test prints statics, like TPS and how many transactions are submitted.

One important piece that is currently missing is latency information, e.g. how much time does transaction take from submitting to committing.

We want to add this statistics

1) At first, lets add average latency - this is the simplest thing.
2) As a second PR we can try adding p90 for latency

Code pointers:
* Most of transaction generation is happening in `tx_emitter.rs`.
* Stats are printed in [main.rs:emit_tx](https://github.com/libra/libra/blob/master/testsuite/cluster-test/src/main.rs#L351)

Breakdown on sub issues

- [x] Calculate client side latency in tx_emitter and print in swarm mode
- [x] Use client side latency in experiments([1](https://github.com/libra/libra/blob/master/testsuite/cluster-test/src/experiments/performance_benchmark.rs#L156), [2](https://github.com/libra/libra/blob/master/testsuite/cluster-test/src/experiments/performance_benchmark_three_region_simulation.rs#L80)) instead of latency from prometheus. Probably can submit intermediate PR that prints both, use `./scripts/cti --pr <PR> --perf-run` to run it and see if there is difference
- [x] Switch TPS to client side estimation in experiments
- [x] Calculate p90 latency and populate it in swarm mode and in experiments

PR #3888, opened 2020-05-14
[cluster_test] Client side latency statistics #3868

snakech:
Added average latency stats for each transaction to take from submitting to committing

## Motivation

(Write your answer here.)

## Test Plan

## Related PRs

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -270,7 +270,9 @@ version = "0.1.0"
 dependencies = [
  "anyhow 1.0.30 (registry+https://github.com/rust-lang/crates.io-index)",
  "async-trait 0.1.31 (registry+https://github.com/rust-lang/crates.io-index)",
+ "backup-service 0.1.0",
  "byteorder 1.3.4 (registry+https://github.com/rust-lang/crates.io-index)",
+ "bytes 0.5.4 (registry+https://github.com/rust-lang/crates.io-index)",
  "futures 0.3.5 (registry+https://github.com/rust-lang/crates.io-index)",
  "hex 0.4.2 (registry+https://github.com/rust-lang/crates.io-index)",
  "itertools 0.9.0 (registry+https://github.com/rust-lang/crates.io-index)",
@@ -284,6 +286,7 @@ dependencies = [
  "libradb 0.1.0",
  "proptest 0.9.6 (registry+https://github.com/rust-lang/crates.io-index)",
  "rand 0.7.3 (registry+https://github.com/rust-lang/crates.io-index)",
+ "reqwest 0.10.4 (registry+https://github.com/rust-lang/crates.io-index)",
  "storage-client 0.1.0",
  "storage-interface 0.1.0",
  "storage-service 0.1.0",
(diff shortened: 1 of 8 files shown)


testsuite/cluster-test/src/tx_emitter.rs
```diff
@@ -376,6 +379,7 @@ impl SubmissionWorker {
         while !self.stop.load(Ordering::Relaxed) {
             let requests = self.gen_requests();
             let num_requests = requests.len();
+            let cur_time = Instant::now();
```
andll, line 382: You probably want some better name here, something like `start_time`. It is not really 'current time' on the next lines

testsuite/cluster-test/src/tx_emitter.rs
```diff
@@ -398,6 +402,9 @@ impl SubmissionWorker {
                     self.stats
                         .expired
                         .fetch_add(uncommitted.len() as u64, Ordering::Relaxed);
+                    self.stats
```
andll, line 405: You don't need to count latency for uncommitted transactions

testsuite/cluster-test/src/tx_emitter.rs
```diff
@@ -729,16 +739,18 @@ impl StatsAccumulator {
...
             committed: self.committed.load(Ordering::Relaxed),
             expired: self.expired.load(Ordering::Relaxed),
+            latency: self.latency.load(Ordering::Relaxed),
         }
     }
 }
 
 impl TxStats {
-    pub fn rate(&self, window: Duration) -> TxStatsRate {
+    pub fn rate(&self, window: Duration, total_tx_num: u64) -> TxStatsRate {
```
andll, line 748: You don't need to supply `total_tx_num` as a parameter, I think you have all the information in the function

testsuite/cluster-test/src/tx_emitter.rs
snakech: I think for this part, we count committed tx by num_requests - uncommitted.len(), so I assume we still need to count the latency for batch of txs with part of uncommitted? Or just as you said, we totally do not count latency if batch of txs just have partial commited?

testsuite/cluster-test/src/tx_emitter.rs
snakech: My understanding for latency is per tx, so in this part, we pass in duration because we need tx number per sec, but for latency, I think we should calculate it by time.

for example, if we have 1000 txs in 10mins, we should divide total latency by 1000? So we can have a average latency for each transaction in last 10 mins?

Please correct me if I am misunderstanding.

testsuite/cluster-test/src/tx_emitter.rs
andll: You can count latency for committed transactions only, and then calculate average as `total_committed_txn_time / num_committed_txn`

testsuite/cluster-test/src/tx_emitter.rs
snakech: I c, I can use self directly instead of passing by value. Please ignore my last comment. Not very familiar with Rust ^_^ Sorry~~

testsuite/cluster-test/src/tx_emitter.rs
snakech: for line #401, we also add up committed txn inside this if condition, do we really want to ignore the latency occurred by those committed txn?

andll: Oh, ok, you are actually right :)
Ok, I think for this first diff we can just include uncommitted txn's and count them too in latency.

When we get to p90, we probably will need to refactor `wait_for_accounts_sequence` to account for every transaction time separately, and then we can fix this to count only committed transactions

snakech: ok, sounds good, I am going to update the pr. Thx @andll

PR #3889, opened 2020-05-14
[Move] Introduce freeze/unfreeze move scripts

moeziniaf:
## Motivation

Freeze/Unfreeze scripts for AOS - move functions already exist to freeze txn operations from an account

Yes

## Test Plan

cargo test -- freeze_unfreeze_account
Transaction_builder.rs freeze/unfreeze test
Tests account can not execute txns once frozen.


language/stdlib/src/transaction_scripts.rs
```diff
@@ -37,6 +37,7 @@ pub enum StdlibScript {
     CreateAccount,
     CreateEmptyAccount,
     EmptyScript,
+    Freeze,
```
runtian-zhou, line 40: FreezeAccount?

sblackshear, line 40: Nit: `FreezeAccount` would be slightly more explicit

andll:
Can we also add UnfreezeAccount script in same PR?

sblackshear:
Would you mind adding tests for each new script [here](https://github.com/libra/libra/blob/master/language/e2e-tests/src/tests/transaction_builder.rs) to make sure the arguments are type-correct + the script does what we expect?

language/stdlib/src/transaction_scripts.rs
tzakian: Not entirely sure why `stdlib.mv` would need to be updated with this. That should only change if there are changes to the stdlib modules unless I'm mistaken (and I don't see any changes to them).

tzakian:
I don't think there should be a staged `stdlib.mv` under transaction scripts, (can happen if you run it from the wrong directory). You'll probably want to remove it and do a `cargo run` from `language/stdlib`

language/stdlib/transaction_scripts/freeze_account.move
```diff
@@ -0,0 +1,7 @@
+script {
+use 0x0::LibraAccount;
+// Script for freezing account by authorized initiator
```
sblackshear, line 3: Nit: align comment with `fun main`

language/e2e-tests/src/tests/transaction_builder.rs
```diff
@@ -65,6 +67,47 @@ fn register_preburn_burn() {
...
+    // Grant freezing privilege to assocation account
+    executor.execute_and_apply(association.signed_script_txn(
+        encode_grant_association_privilege(lbr_ty, *association.address()),
+        2,
+    ));
+    // Execute freeze on account
+    executor.execute_and_apply(
+        association.signed_script_txn(encode_freeze_account(*account.address()), 3),
+    );
+
```
sblackshear, line 104: Not strictly necessary, but could be useful to test that sending a tx from `account` fails after freeze and succeeds after unfreeze.

sblackshear: Any transaction will do for testing this, but key rotation is probably the easiest: https://github.com/libra/libra/blob/master/language/e2e-tests/src/tests/rotate_key.rs

moeziniaf:
Included relevant more robust test that frozen account can not perform txns

PR #3890, opened 2020-05-14
[backup] move away from gPRC storage server

msmouse:
implemented the other API needed on the HTTP server.

## Motivation

We are about ready to remove gRPC server entirely.

yes

## Test Plan

tests


lightmark:
@bors-libra r+
kudos to @msmouse

config/src/config/test_data/random.complete.node.config.toml
```diff
@@ -61,7 +61,7 @@ max_chunk_limit = 1000
 max_timeout_ms = 120000
 
 [storage]
-address = "127.0.0.1:6184"
```
msmouse, line 64: @davidiw  is it okay to fix it like this? context: we removed "address" for grpc, and renamed "simple_address" for the SimpleStorageService to "address", which originally had the value of 6666

davidiw: that's how we roll!

yes, it is okay

lightmark: 6666 is just a magic number I thought of. I think it's the same for 6184...

PR #3896, opened 2020-05-14
[language] add support for u8 & u128 as transaction arguments

vgao1996:
## Summary
This adds support for u8 and u128 as transaction arguments in tests & client cli. Transaction arguments are now parsed using a proper parser. 

## Test Plan
cargo test


dariorussi:
I wonder about adding the parser to move-core-types. I guess is fine for now and we'll see in the future but this seems too much of a special code for certain kind of clients....

vgao1996:
@dariorussi yeah that's a fair point, although I can't think of where else that code could go... (transaction_builder?). Anyways it's a fairly isolated component and we'll see if we need to move it soon.

PR #3897, opened 2020-05-14
[move vm] Add move_to that uses signer

tnowacki:
## Motivation

- Added `MoveTo` and `MoveToGeneric` that operate over a `signer` instead of the `GetTxnSenderAddress`
- This (plus a new native function) will open up refactors in LibraAccount

## Test Plan

- new tests


language/bytecode-verifier/src/reference_safety/mod.rs
```diff
@@ -259,6 +259,12 @@ fn execute_inner(
         | Bytecode::MoveToSenderGeneric(_) => {
             checked_verify!(verifier.stack.pop().unwrap().is_value());
         }
+        Bytecode::MoveTo(_) | Bytecode::MoveToGeneric(_) => {
+            // resource value
+            checked_verify!(verifier.stack.pop().unwrap().is_value());
+            // signer reference
+            state.release_value(verifier.stack.pop().unwrap());
+        }
```
sblackshear, line 267: #fresh

language/ir-testsuite/tests/signer/move_to_non_signer.mvir
```diff
@@ -0,0 +1,21 @@
...
+        return;
+    }
+}
+// check: MOVETO_TYPE_MISMATCH_ERROR
+
+//! new-transaction
+
+module N {
+    resource R<T> { f: T }
+    t0<T>(s: address) {
```
sblackshear, line 16: `T` needed here?

language/ir-testsuite/tests/signer/move_to_extra_arg.mvir
```diff
@@ -0,0 +1,23 @@
...
+        return;
+    }
+}
+// check: POSITIVE_STACK_SIZE_AT_BLOCK_END
+
+//! new-transaction
+
+module N {
+    resource R<T> { f: T }
+    t0<T>(s: &signer, a: address) {
```
sblackshear, line 18: `T` needed here (same question in a couple of other places below)? Makes the tests a bit  hard to read because `T` is unused here but used above

tnowacki: Ah yeah, bad copy paste. I'll remove them

tnowacki:
> Man, it is challenging to test the interpreter part of this without the ability to create a sender...

All things in time :) I'm not too worried about any bugs or anything, its an iterative process and if we can't test one because we can't create one, then no one can break anything

language/move-vm/runtime/src/interpreter.rs
```diff
@@ -390,6 +390,24 @@ impl<'txn> Interpreter<'txn> {
...
+    fn move_to(
+        resource: Struct,
+    ) -> Box<
+        dyn FnOnce(
+            &mut Interpreter,
+            &mut dyn InterpreterContext,
+            AccessPath,
+            &FatStructType,
+        ) -> VMResult<AbstractMemorySize<GasCarrier>>,
+    > {
```
vgao1996, line 403: You don't necessarily have to Box the closure. The following would work:
```
    /// MoveTo opcode.
    fn move_to(
        resource: Struct,
    ) -> impl FnOnce(
        &mut Interpreter,
        &mut dyn InterpreterContext,
        AccessPath,
        &FatStructType,
    ) -> VMResult<AbstractMemorySize<GasCarrier>> {
        |_interpreter, context, ap, struct_ty| {
            let size = resource.size();
            context.move_resource_to(&ap, struct_ty, resource)?;
            Ok(size)
        }
    }
```

vgao1996: No it's not. `dyn Foo` is fundamentally different from impl Foo.

IMO `dyn Foo` represents a sum type of all types that implement trait `Foo` and `impl Foo` at return position is more like an existential type. In other words, `impl Foo` means "I'm gonna return a type but nah I'm not gonna tell you what that type is, but you do know it implements trait Foo". 

```
trait Foo {}

struct Bar;
struct Baz;

impl Foo for Bar {}
impl Foo for Baz {}

fn f(b: bool) -> impl Foo {
    if b {
        Bar
    }
    else {
        Baz
    }
    // compile error, there's no way you can solve these set of equations:
    //     Bar ~ x, Baz ~ x, x implements Foo
}

fn g(b: bool) -> Box<dyn Foo> {
    if b {
        Box::new(Bar)
    }
    else {
        Box::new(Baz)
    }
    // good
}

fn main() {
    let v: Vec<Box<dyn Foo>> = vec![Box::new(Bar), Box::new(Baz)];
    // good
}
```

Now in practice, in order to actually manage the memory for dyn types, you'll want it to be boxed almost all the time to meet the Sized requirement, and yes calling a method on a dyn type requires a vtable lookup. (dynamic dispatch)

Existential types on the other hand are statically dispatched. The compiler actually knows what the type is. Rust doesn't allow you to spell out the type of a closure, but imagine if can, you will actually be able to just write down the actual type instead of `impl Foo`, just that the type maybe super long and complex. (Another example is returning iterator combinators.)

tnowacki: >  The compiler actually knows what the type is. Rust doesn't allow you to spell out the type of a closure

I now understand how it is doing it. Thanks!

I've seen a similar thing before when compiling closures to monomorphize all of them to different instances. So the impl is saying, hey! I have some instance but you don't which
The Box<dyn _> is more that you have the "parent" of all of those instances (from a subtyping perspective)

vgao1996: Yeah subtyping is another way to think about this.

PR #3899, opened 2020-05-15
[management] smoke test!

davidiw:
Per title, this introduces an end-to-end smoke test for management.
The test completes because all nodes wre able to get connected.
this shows that there are still a few missing pieces:
* waypoint generation / insertion
* making sense of fullnodes
* strip out network signing key or evaluate if we want it there

some cleanups:
* a better framework for converting ed25519 to x25519 so we don't need
to export the ed25519 private key
* better integration or replacement of libra swarm

### DIFF
--- a/config/management/src/smoke_test.rs
+++ b/config/management/src/smoke_test.rs
@@ -1,0 +1,169 @@
+// Copyright (c) The Libra Core Contributors
+// SPDX-License-Identifier: Apache-2.0
+
+use crate::{layout::Layout, storage_helper::StorageHelper};
+use config_builder::{BuildSwarm, SwarmConfig};
+use executor::db_bootstrapper;
+use libra_config::config::{
+    DiscoveryMethod, NetworkConfig, NetworkKeyPairs, NodeConfig, OnDiskStorageConfig, RoleType,
+    SecureBackend,
+};
+use libra_crypto::{ed25519::Ed25519PrivateKey, x25519, Uniform};
+use libra_secure_storage::Value;
+use libra_swarm::swarm::{LibraSwarm, LibraSwarmDir};
+use libra_temppath::TempPath;
+use libra_types::account_address::AccountAddress;
+use libra_vm::LibraVM;
+use libradb::LibraDB;
+use std::{convert::TryInto, path::PathBuf};
+use storage_interface::DbReaderWriter;
+
+struct ManagementBuilder {
+    configs: Vec<NodeConfig>,
+    faucet_key: Ed25519PrivateKey,
+}
+
+impl BuildSwarm for ManagementBuilder {
+    fn build_swarm(&self) -> anyhow::Result<(Vec<NodeConfig>, Ed25519PrivateKey)> {
+        Ok((self.configs.clone(), self.faucet_key.clone()))
+    }
+}
+
+#[test]
+fn smoke_test() {
+    let helper = StorageHelper::new();
+    let num_validators = 5;
+    let shared = "_shared";
+    let association = "association";
+    let association_shared = association.to_string() + shared;
+
+    // Step 1) Prepare the layout
+    let mut layout = Layout::default();
+    layout.association = vec![association_shared.to_string()];
+    layout.operators = (0..num_validators)
+        .map(|v| (v.to_string() + shared))
+        .collect();
+
+    let mut common_storage = helper.storage(crate::constants::COMMON_NS.into());
+    let layout_value = Value::String(layout.to_toml().unwrap());
+    common_storage
+        .set(crate::constants::LAYOUT, layout_value)
+        .unwrap();
+
+    // Step 2) Set association key
+    helper.initialize(association.into());
+    helper
+        .association_key(&association, &association_shared)
+        .unwrap();
+
+    // Step 3) Prepare validators
+    let mut configs = Vec::new();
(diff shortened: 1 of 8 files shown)


huitseeker:
See also #3904

consensus/safety-rules/src/safety_rules.rs
```diff
@@ -120,7 +120,7 @@ impl<T: Payload> SafetyRules<T> {
                 .set_waypoint(&Waypoint::new_epoch_boundary(ledger_info)?)?;
             self.persistent_storage.set_last_voted_round(0)?;
             self.persistent_storage.set_preferred_round(0)?;
-            self.persistent_storage.set_epoch(ledger_info.epoch() + 1)?;
+            self.persistent_storage.set_epoch(next_epoch)?;
```
zekun000, line 123: you can use epoch_info.epoch and skip the next_epoch var

config/management/src/smoke_test.rs
```diff
@@ -0,0 +1,167 @@
...
+            .unwrap();
+        let identity_key = identity_key.to_bytes().as_ref().try_into().unwrap();
+        // This is a deprecated field
+        let signing_key = Ed25519PrivateKey::generate_for_testing();
+        fullnode_network.network_keypairs = Some(NetworkKeyPairs::load(identity_key, signing_key));
+        let fullnode_network_address = fullnode_network.listen_address.clone();
+
+        configs.push(config);
+
+        helper.operator_key(&ns, &ns_shared).unwrap();
```
zekun000, line 100: those helper functions are so hard to read, spend a few mins realizing this is to fetch from local and submit to remote

config/management/src/smoke_test.rs
```diff
@@ -0,0 +1,167 @@
...
+    for i in 0..num_validators {
+        let storage = helper.storage(i.to_string());
+        let ns = i.to_string();
+        let ns_shared = ns.clone() + shared;
+        helper.initialize(ns.clone());
+
+        let mut config = NodeConfig::default();
+
+        let mut network = NetworkConfig::default();
+        network.discovery_method = DiscoveryMethod::Onchain;
```
zekun000, line 70: milestone!

config/management/src/smoke_test.rs
```diff
@@ -0,0 +1,167 @@
...
+
+    for (i, mut config) in configs.iter_mut().enumerate() {
+        let waypoint_value = Value::String(waypoint.to_string());
+        helper
+            .storage(i.to_string())
+            .set(libra_global_constants::WAYPOINT, waypoint_value)
+            .unwrap();
+
+        let mut to = swarm_path.clone();
+        to.push("secure_store_for_".to_string() + &i.to_string());
```
zekun000, line 136: use format?

config/management/src/smoke_test.rs
```diff
@@ -0,0 +1,167 @@
...
+        std::fs::copy(helper.path_string(), &to).unwrap();
+
+        let mut storage_config = OnDiskStorageConfig::default();
+        storage_config.path = to;
+        storage_config.set_data_dir(PathBuf::from(""));
```
zekun000, line 141: is this not used?

config/management/src/smoke_test.rs
```diff
@@ -0,0 +1,167 @@
...
+    // Step 3) Prepare validators
+    let mut configs = Vec::new();
+    for i in 0..num_validators {
+        let storage = helper.storage(i.to_string());
+        let ns = i.to_string();
+        let ns_shared = ns.clone() + shared;
+        helper.initialize(ns.clone());
+
+        let mut config = NodeConfig::default();
+
```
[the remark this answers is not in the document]
davidiw: it is a separate clause from network and has nothing to do with ns initialization

config/management/src/smoke_test.rs
```diff
@@ -0,0 +1,167 @@
...
+        let mut network = NetworkConfig::default();
+        network.discovery_method = DiscoveryMethod::Onchain;
+        config.validator_network = Some(network);
+
+        let mut network = NetworkConfig::default();
+        network.discovery_method = DiscoveryMethod::Onchain;
+        config.full_node_networks = vec![NetworkConfig::default()];
```
davidiw, line 75: woops, good call, I added fullnode support but don't use it, so it is untested... I'll clean this up

config/management/src/smoke_test.rs
davidiw: ahh, the intent is to ensure the command-line works as possible, I guess I could add comments to storage_helper... thoughts on what the confusion stemmed from?

config/management/src/smoke_test.rs
davidiw: it defaults to something, but no, it is not, if the path is a full path, this is ignored, but I want to be careful to not leave some unused data here.

consensus/safety-rules/src/safety_rules.rs
davidiw: oh, this must have appeared after your code refactor, it wasn't available before :)

config/management/src/smoke_test.rs
zekun000: I'd expect something like `submit_[operator_key|validator_config]`, `operator_key` feels the function only returns the operator key without doing anything else

PR #3901, opened 2020-05-15
[move-prover] Updating the Boogie translation to reflect new memory model

shazqadeer:
## Motivation

This PR moves Boogie translation to the new memory model.  In addition to bug fixes in the core transforms and analyses, the bulk of the changes are in bytecode_translator.rs.  The interaction between the new memory model and explicit PackRef/UnpackRef and resource invariants is not fully sorted out.  Specifically, there are several discrepancies in the output for the test case tests/sources/functional/mut_ref_accross_modules.exp.  I have noted down these discrepancies for further investigation.
 
Yes

## Test Plan

Existing tests

### DIFF
--- a/language/move-prover/tests/sources/functional/aborts-if.exp
+++ b/language/move-prover/tests/sources/functional/aborts-if.exp
@@ -8,8 +8,6 @@ error:  A postcondition might not hold on this return path.
     │
     =     at tests/sources/functional/aborts-if.move:32:5: abort2_incorrect (entry)
     =     at tests/sources/functional/aborts-if.move:32:5: abort2_incorrect (exit)
-    =         _x = <redacted>,
-    =         _y = <redacted>
 
 error:  A postcondition might not hold on this return path.
 
@@ -20,8 +18,6 @@ error:  A postcondition might not hold on this return path.
     │
     =     at tests/sources/functional/aborts-if.move:47:5: abort4_incorrect (entry)
     =     at tests/sources/functional/aborts-if.move:48:9: abort4_incorrect
-    =         x = <redacted>,
-    =         y = <redacted>
     =     at tests/sources/functional/aborts-if.move:47:5: abort4_incorrect (exit)
 
 error: abort not covered by any of the `aborts_if` clauses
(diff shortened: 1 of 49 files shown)


language/move-prover/tests/sources/functional/aborts-if.exp
```diff
@@ -8,8 +8,6 @@ error:  A postcondition might not hold on this return path.
     │
     =     at tests/sources/functional/aborts-if.move:32:5: abort2_incorrect (entry)
     =     at tests/sources/functional/aborts-if.move:32:5: abort2_incorrect (exit)
-    =         _x = <redacted>,
```
wrwg, line 11: Looks like all debug info is gone here. It's likely OK we submit this, but we need to get this back ASAP, given that its needed for folks working on the specs.

language/move-prover/tests/sources/functional/mut_ref_accross_modules.move
```diff
@@ -76,22 +76,22 @@ module TestMutRefs {
     }
 
     // For a public function, the data invariant for a mut ref is expected to hold.
-    public fun data_invariant(x: &mut T) {
+    public fun data_invariant(x: &mut T) { // Shaz: This function verifies because resource invariant is assumed on the inout value modeling x
```
wrwg, line 79: Is this a comment or an issue we need to fix? If it is an issue please clearly mark a TODO here and state it needs to be fixed. If not I'm wondering what the comment is about. if its a clarification, please remove redundant information ("Shaz") and put the comment before the function. (Which also says something similar?) Same for other cases.

I remember you said that where issues with some tests but I can't find them documented in the PR.

shazqadeer: All the issues were in these "Shaz:" comments in this one test.  I have improved the comments to explain the issue more precisely.

language/move-prover/tests/sources/functional/aborts-if.exp
shazqadeer: Agreed.  Debug info and the other test you commented on are the two major outstanding issues once this PR lands.

PR #3904, opened 2020-05-15
[crypto] Adds conversion to/from X25519 from/to Ed25519 public keys

huitseeker:
This is meant to allow compensating for deficiencies of key storage engines and help produce back-convertible serialized assets, rather than actually allow reuse of distinct keys.

### DIFF
--- a/crypto/crypto/src/ed25519.rs
+++ b/crypto/crypto/src/ed25519.rs
@@ -117,6 +117,39 @@ impl Ed25519PublicKey {
             Err(_) => Err(CryptoMaterialError::DeserializationError),
         }
     }
+
+    /// Deserialize an Ed25519PublicKey from its representation as an x25519
+    /// public key, along with an indication of sign. This is meant to
+    /// compensate for the poor key storage capabilities of key management
+    /// solutions, and NOT to promote double usage of keys under several
+    /// schemes, which would lead to BAD vulnerabilities.
+    /// Arguments:
+    /// - `x25519_bytes`: bit representation of a public key in clamped
+    ///            Montgomery form, a.k.a. the x25519 public key format.
+    /// - `negative`: whether to interpret the given point as a negative point,
+    ///               as the Montgomery form erases the sign byte. By XEdDSA
+    ///               convention, if you expect to ever convert this back to an
+    ///               x25519 public key, you should pass `false` for this
+    ///               argument.
+    pub fn from_x25519_bytes(
+        x25519_bytes: &[u8],
+        negative: bool,
+    ) -> Result<Self, CryptoMaterialError> {
+        if x25519_bytes.len() != 32 {
+            return Err(CryptoMaterialError::DeserializationError);
+        }
+        let key_bits = {
+            let mut bits = [0u8; 32];
+            bits.copy_from_slice(x25519_bytes);
+            bits
+        };
+        let mtg_point = curve25519_dalek::montgomery::MontgomeryPoint(key_bits);
+        let sign = if negative { 1u8 } else { 0u8 };
+        let ed_point = mtg_point
+            .to_edwards(sign)
+            .ok_or(CryptoMaterialError::DeserializationError)?;
+        Ed25519PublicKey::try_from(&ed_point.compress().as_bytes()[..])
+    }
 }
 
 impl Ed25519Signature {
--- a/crypto/crypto/src/ed25519.rs
+++ b/crypto/crypto/src/ed25519.rs
@@ -123,6 +123,7 @@ impl Ed25519PublicKey {
     /// compensate for the poor key storage capabilities of key management
     /// solutions, and NOT to promote double usage of keys under several
     /// schemes, which would lead to BAD vulnerabilities.
+    ///
     /// Arguments:
     /// - `x25519_bytes`: bit representation of a public key in clamped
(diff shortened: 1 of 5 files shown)


davidiw:
Nice, I think the only remaining component to add is to verify that the Ed25519PrivateKey -> X25519PrivateKey::public_key == Ed25519PublicKey::to_montgomery()

Having not looked at what is happening behind the scenes, it is unclear, but can be easily tested.

huitseeker:
@davidiw can you reformulate your equation? Given that this PR does not include a comparison between private keys yet, it's hard to interpret which of several possible meanings you intend.

huitseeker:
@davidiw 

I've added a test to exemplify what you seem to suggest.

`x25519::PublicKey::from(x25519::PrivateKey::try_from(&ed25519_privkey.to_bytes()[..]).unwrap()) == x25519PublicKey::from_ed25519_bytes(&ed25519_pubkey.to_bytes()[..]).unwrap())`

is false, because that includes a conversion of a general Ed25519 private key into the restrictive x25519 private key format, which while well-defined, mangles some bits irreversibly.

However, storing genuine x25519 key bytes as an ed25519 key pair (a "looser" type from all points of view) allows you to retrieve precisely your pristine x25519 key bytes, as shown by `ed25519_and_x25519_privkeys`.

crypto/crypto/src/x25519.rs
```diff
@@ -93,13 +96,56 @@ impl PrivateKey {
...
     pub fn as_slice(&self) -> &[u8] {
         &self.0
     }
+
+    /// Deserialize an X25119 PublicKey from its representation as an
+    /// Ed25519PublicKey, following the XEdDSA approach. This is meant to
+    /// compensate for the poor key storage capabilities of key management
+    /// solutions, and NOT to promote double usage of keys under several
+    /// schemes, which would lead to BAD vulnerabilities.
+    pub fn from_ed25519_bytes(ed25519_bytes: &[u8]) -> Result<Self, CryptoMaterialError> {
```
valerini, line 139: By the name of this function, it's not very clear, what bytes these are...
May be call it: `from_ed25519_public_key_bytes` ?

crypto/crypto/src/x25519.rs
```diff
@@ -93,13 +96,56 @@ impl PrivateKey {
         let shared_secret = self.0.diffie_hellman(&remote_public_key);
         shared_secret.as_bytes().to_owned()
     }
+
+    /// Deserialize an X25119 PrivateKey given a canonical representation as an
```
valerini, line 100: as an bytes -> as bytes

crypto/crypto/src/x25519.rs
```diff
@@ -93,13 +96,56 @@ impl PrivateKey {
...
+    /// bytes, following the XEdDSA approach.
+    ///
+    /// This will FAIL if the passed-in byte representation converts to a
+    /// non-canonical scalar in the X25519 sense (and thus cannot correspond to
+    /// a X25519 valid key without bit-mangling).
+    ///
+    /// This is meant to compensate for the poor key storage capabilities of
+    /// key management solutions, and NOT to promote double usage of keys under
+    /// several schemes, which would lead to BAD vulnerabilities.
+    pub fn from_scalar_bytes(scalar_slice: &[u8]) -> Result<Self, CryptoMaterialError> {
```
valerini, line 110: Why is the argument called scalar_slice? Is this function simply generates an x25519 private key from a random byte array?

huitseeker: The function will return `Err` on precisely those inputs whose ms-half-of-sha512 image are not a canonical X25519 Scalar.

crypto/crypto/src/ed25519.rs
```diff
@@ -117,6 +117,39 @@ impl Ed25519PublicKey {
...
+    /// schemes, which would lead to BAD vulnerabilities.
+    /// Arguments:
+    /// - `x25519_bytes`: bit representation of a public key in clamped
+    ///            Montgomery form, a.k.a. the x25519 public key format.
+    /// - `negative`: whether to interpret the given point as a negative point,
+    ///               as the Montgomery form erases the sign byte. By XEdDSA
+    ///               convention, if you expect to ever convert this back to an
+    ///               x25519 public key, you should pass `false` for this
+    ///               argument.
+    pub fn from_x25519_public_bytes(
```
valerini, line 134: Why do we need x25519 -> ed25519? For a complete picture?
But since these are "dangerous" APIs, may be we should not expose conversion in this direction since we don't need it?

crypto/crypto/src/x25519.rs
valerini: Don't we want instead a function ed25519 sk -> x25519 sk?
Where the ed25519 sk is an UN-expanded secret key (i.e. pre SHA512)?

crypto/crypto/src/ed25519.rs
huitseeker: For unit tests IIUC. I'll gate it.

crypto/crypto/src/x25519.rs
valerini: I was just confused by the naming of the function's argument. Makes sense now!

davidiw:
Thanks a lot!!!

@bors-libra r+

PR #3905, opened 2020-05-15
[crypto] Convert ed25519 keys to x25519 keys

kchalkias:
## Motivation

Convert an ed25519 key (public or private) to its corresponding x25519 key. 
This might be required in cases where KMS (key management service) can only store Ed25519 keys (i.e., Vault or HSM), but we use it for x25519 noise (DH) protocol communication.

YES

## Test Plan

`convert_from_ed25519_publickey()`

## Related PRs

https://github.com/libra/libra/pull/3899

### DIFF
--- a/crypto/crypto/src/unit_tests/noise_test.rs
+++ b/crypto/crypto/src/unit_tests/noise_test.rs
@@ -9,8 +9,27 @@ use crate::{
     x25519, Uniform as _,
 };
 
+use crate::ed25519::{Ed25519PrivateKey, Ed25519PublicKey};
 use rand::SeedableRng;
 use serde::*;
+use std::convert::TryFrom;
+
+#[test]
+fn convert_from_ed25519_publickey() {
+    let mut rng = ::rand::rngs::StdRng::from_seed(TEST_SEED);
+    let ed25519_private_key = Ed25519PrivateKey::generate(&mut rng);
+    let ed25519_public_key: Ed25519PublicKey = Ed25519PublicKey::from(&ed25519_private_key);
+    let x25519_public_key = x25519::PublicKey::try_from(&ed25519_public_key);
+    assert!(x25519_public_key.is_ok());
+
+    // Let's construct an x25519 private key from the ed25519 private key.
+    let x25519_privatekey = x25519::PrivateKey::try_from(&ed25519_private_key).unwrap();
+
+    // Now derive the public key from x25519_privatekey and see if it matches the public key that
+    // was created from the Ed25519PublicKey.
+    let x25519_publickey_2 = x25519_privatekey.public_key();
+    assert_eq!(x25519_public_key.unwrap(), x25519_publickey_2)
+}
 
 #[test]
 fn simple_handshake() {
--- a/crypto/crypto/src/traits.rs
+++ b/crypto/crypto/src/traits.rs
@@ -38,6 +38,8 @@ pub enum CryptoMaterialError {
     PointNotOnCurveError,
     /// BitVec errors in accountable multi-sig schemes.
     BitVecError(String),
+    /// Key conversion related errors (i.e., from ed25519 to x25519).
+    KeyConversionError,
 }
 
 /// The serialized length of the data that enables macro derived serialization and deserialization.
--- a/crypto/crypto/src/x25519.rs
+++ b/crypto/crypto/src/x25519.rs
@@ -47,6 +47,10 @@ use proptest_derive::Arbitrary;
 // This makes it easier to uniformalize build dalek-x25519 in libra-core.
 //
 
+use crate::ed25519::{Ed25519PrivateKey, Ed25519PublicKey};
+use crate::CryptoMaterialError;
+use curve25519_dalek::edwards::CompressedEdwardsY;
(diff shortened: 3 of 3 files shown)


crypto/crypto/src/unit_tests/noise_test.rs
```diff
@@ -9,8 +9,36 @@ use crate::{
...
+    // Let's try to reconstruct an x25519 private key from the ed25519 private key.
+    let ed25519_secretkey = SecretKey::from_bytes(&ed25519_private_key.to_bytes()).unwrap();
+    let expanded_ed25519_privatekey: ed25519_dalek::ExpandedSecretKey =
+        ed25519_dalek::ExpandedSecretKey::from(&ed25519_secretkey);
+
+    // Note that the first 32 bytes of the ExpandedSecretKey correspond to the actual private key.
+    let mut expanded_keypart = [0u8; 32];
+    let bytes = &expanded_ed25519_privatekey.to_bytes()[..32];
+    expanded_keypart.copy_from_slice(bytes);
+    let x25519_privatekey = x25519::PrivateKey::from(expanded_keypart);
```
davidiw, line 35: could we hide this in a function within Ed25519PrivateKey? I'd like to not expose this outside of this crate.

kchalkias: Do you prefer a  try_from inside x25519 (as I did for public key) or because it's a private key (sensitive) you prefer this being a function of Ed25519PrivateKey?

davidiw: I mean... this API comes with its risks, I would say try_from is good enough with the caveat stating that this is intended to only be used for "cases where KMS (key management service) can only store Ed25519 keys (i.e., a Vault or HSM), but we use it for x25519 noise (DH) protocol communication." or the like.

crypto/crypto/src/unit_tests/noise_test.rs
```diff
@@ -9,8 +9,27 @@ use crate::{
     x25519, Uniform as _,
 };
 
+use crate::ed25519::{Ed25519PrivateKey, Ed25519PublicKey};
 use rand::SeedableRng;
 use serde::*;
+use std::convert::TryFrom;
+
+#[test]
+fn convert_from_ed25519_publickey() {
```
huitseeker, line 18: This test will fail on some inputs, specifically on inputs that are not [X25519-canonical](https://tools.ietf.org/html/rfc7748#section-5). I reimplemented it as a proptest on top of my #3904 here:

https://github.com/huitseeker/libra/compare/crypto-shenanigans...huitseeker:this_test_doesnt_work?expand=1

The test output, including a counterexample:
https://gist.github.com/1c93555d176f731af485f740a92c8615

kchalkias: Yes we could add this test here as well. I do catch it on try_from, but didn't test it here.

kchalkias: @huitseeker note that in your referenced test, you use the Ed private key bytes directly (w/o expansion)
`let x25519_privatekey = x25519::PrivateKey::try_from(&keypair.private_key.to_bytes()[..]).unwrap();` which will always fail.

, while in my approach I produce the expanded key first (after sha512), then Ed25519 key clamping, then get the first 32bytes. Based on how clamping of the most significant bits works on Ed25519 and x25519 (and because Ed25519 is more restrictive as it sets to zero the 2nd most significant bit), this conversion is always possible.

However, I get it that this does not conform 100% w/ the x25519 standard (due to this 2nd bit twiddling), but this is the price to pay if we want to support what @davidiw expects here.

huitseeker: Oh yes indeed! :facepalm: Yeah, I messed up the test. But you can still work around the bit-clamping of X25519 by rejecting inputs where it has to apply.

I also have the conversion in my PR (`ed25519::Ed25519PrivateKey::from_ed25519_private_bytes`)
and I outline the difference in type of conversions in the tests `convert_from_ed25519_publickey` and `ed25519_and_x25519_privkeys`. 

In the former, you take in ed25519 bits, and you *reinterpret* what is the "moral" key in there, i.e. the scalar that links the public key to the private key. I reject those scalars that cannot be represented as x25519 scalars unmodified, because in our block chain a signature verification failure can have dire consequences (e.g. if your public noise signature is not what's expected, you may be considered as an atacker). That's one approach I call "reinterpretation" for the sake of discussion.

In `ed25519_and_x25519_privkeys` I consider you have marshalled an x25519 keypair inside an ed25519 storage device, knowing what you're doing. In that case the public and private key ed25519 key do not "match" each other, but the x25519 keypair you can extract from that is valid in every case. That's what I call marshalling.

As discussed with @valerini over review, both approaches are valid, and the inputs to them are genuinely untyped: without user input, there really is two ways to see an ed25519 key pair as x25519. Hence why I now think more strongly that a serialization/deserialization approach works better than a conversion.

kchalkias:
Closing this as we agreed w/ @huitseeker and @valerini to go with the serialization method in [3904](https://github.com/libra/libra/pull/3904)

PR #3907, opened 2020-05-15
[testcompile] fix cargo toml to allow test compilation

rexhoffman:
## Motivation

Allow this crate to compile on it's own (and if tests are ever added test on its own).

Yes

## Test Plan

Failed local compile in executor-util, fix, successful compile.

## Related PRs

None

### DIFF
--- a/execution/executor-utils/Cargo.toml
+++ b/execution/executor-utils/Cargo.toml
@@ -27,3 +27,4 @@ libra-types = { path = "../../types", version = "0.1.0" }
 
 [features]
 testing = []
+fuzzing = [ "libra-types/fuzzing" ]


execution/executor-utils/Cargo.toml
```diff
@@ -25,5 +25,8 @@ libra-workspace-hack = { path = "../../common/workspace-hack", version = "0.1.0"
 libra-crypto = { path = "../../crypto/crypto", version = "0.1.0" }
 libra-types = { path = "../../types", version = "0.1.0" }
 
+[dev-dependencies]
+libra-types = { path = "../../types", version = "0.1.0", features = ["fuzzing"] }
```
davidiw, line 29: why not add the features flag to line 26?

rexhoffman: Because line 24 is commented out with a '#', and isn't the start of a dev dependencies section.

metajack: This will cause feature unification. You need to enable this from the testing feature below than then pass the feature explicitly or have dependent packages turn it on.

rexhoffman: good point, thanks!   was tired when getting these to compile.

PR #3910, opened 2020-05-15
[language] Separate cost and data API for the MoveVM

dariorussi:
Here we are breaking up the API for gas and data in the MoveVM.
We remove few traits as a side effect (e.g. `ChainState`, `InterpreterContext`, `ExecutionContext`) in favor of 2 API: `CostStrategy` and `DataStore`.
Names should be reviewed and suggestions welcome (waiting for @tzakian rage ;)).

The reason for the break up is that _cost_ and _data_ have different lifetime and characteristics for clients and they are independent of each other, as such there is no reason to bundle them in a single API.

I was also trying to remove traits as much as possible and lock the API inside the MoveVM. That is not the long term strategy we want, but at the same time we are not in a good position, at the moment, to define a correct and consistent plug in story for the Move VM. In that sense restricting implementation is the right thing for V1. However because of genesis and the resource viewer we are forced to have a trait at the moment.
Once we clean up the Move Value story we may come up with something else and be able to remove the trait.

`move-vm/state` is gone because what was there were either removed traits or concrete API that could be split across Libra and Move.

I was planning to put this up as an RFC but it's ready and something I believe we should do.
In any case feel free to ask question and express your concerns, if you have any


language/libra-vm/src/data_cache.rs
```diff
@@ -0,0 +1,98 @@
+// Copyright (c) The Libra Core Contributors
```
dariorussi, line 1: this is effectively the Libra implementation of the data cache.
As the comment says it is also used by some random client for convenience and we may remove that later.
However I'd be fine to go out as is too given the only public usage is to create one.
It makes sense to me for this to live in the libra layer given it is exclusively a libra or client concept

language/move-vm/runtime/src/data_cache.rs
```diff
@@ -1,75 +1,25 @@
 // Copyright (c) The Libra Core Contributors
```
dariorussi, line 1: this moved from `state` to `runtime` as it was the only thing left in `state` and it was required by the `runtime`.
Users of this are users of the runtime so it does not even force people to extra dependencies.
`RemoteCache` is the only trait that could be moved elsewhere (e.g. `move-vm/types`) and we'll see.
If I could (and maybe we will be able to after @vgao1996 work on values) I would remove `DataStore` and force this concrete implementation in the API.
For now clients should use this and certainly that is what Libra uses

language/move-vm/types/src/data_store.rs
```diff
@@ -0,0 +1,56 @@
+// Copyright (c) The Libra Core Contributors
```
dariorussi, line 1: this is the trait and the data store API that has to be provided to the Move VM.
Maybe `RemoteCache` could be moved here?

language/move-vm/types/src/gas_schedule.rs
```diff
@@ -6,21 +6,117 @@
...
     file_format::{
         Bytecode, ConstantPoolIndex, FieldHandleIndex, FieldInstantiationIndex,
         FunctionHandleIndex, FunctionInstantiationIndex, StructDefInstantiationIndex,
         StructDefinitionIndex, NUMBER_OF_NATIVE_FUNCTIONS,
     },
-    file_format_common::instruction_key,
+    file_format_common::{instruction_key, Opcodes},
 };
 
+/// The Move VM implementation for gas charging.
```
dariorussi, line 28: this is the gas Api to be provided to the Move VM.
It is a concrete implementation and it's forced to clients.
It's the most reasonable way I could see this working for now.
I think this is much better than any trait story

dariorussi:
> Hey you! Here is the Move coverage report
> 
> ```
> undefined
> 
> You can get more details here: https://github.com/mimoo/mirai-bot/runs/693121445?check_suite_focus=true#step:7:1
> ```

anybody any idea about this? the link does not really help for me: @hermanventer, @junkil-park, @sausagee, @wrwg

dariorussi:
@runtian-zhou, @tzakian, @vgao1996 please take a look and let me know your thoughts. Happy to chat about this as needed

tzakian:
Woops! my fault about the bot...It should be working once #3995 lands.

My apologies for the churn!

language/move-vm/types/src/data_store.rs
vgao1996: I feel like the remote cache isn't a concept that needs to be exposed to the Move VM. To me it's specific to Libra and therefore should stay in the Libra/Move adapter layer.

language/move-vm/types/src/gas_schedule.rs
vgao1996: I think the concrete implementation is good at least for now and probably for v1, but I don't see the downsides of a trait story either. Could you elaborate? My gut feeling is that a trait story might help us move certain concepts out of the Move VM and make it possible to enable/disable gas metering via monomorphization.

language/libra-vm/src/data_cache.rs
```diff
@@ -0,0 +1,98 @@
...
+// Adapter to convert a `StateView` into a `RemoteCache`.
+pub struct RemoteStorage<'a>(&'a dyn StateView);
+
+impl<'a> RemoteStorage<'a> {
+    pub fn new(state_store: &'a dyn StateView) -> Self {
+        Self(state_store)
+    }
+}
+
+impl<'a> RemoteCache for RemoteStorage<'a> {
```
runtian-zhou, line 92: BTW I would encourage us to look into the different abstractions of `RemoteCache`. We have `StateView` and `ConfigStorage` IIRC that basically did the same thing.

language/move-vm/runtime/src/data_operations.rs
```diff
@@ -0,0 +1,109 @@
...
+//
+// Provides an implementation for data store bytecodes and guarantees proper invariants
+// for each operation. It uses the `DataStore` for the data operation but it makes sure
+// all operations are consistent to Move semantic.
+//
+
+// Publish a resource if the resource does not exists at that location.
+// Return an error otherwise.
+pub(crate) fn move_resource_to(
+    data_store: &mut dyn DataStore,
```
runtian-zhou, line 26: same here

language/libra-vm/src/data_cache.rs
```diff
@@ -0,0 +1,98 @@
...
+}
+
+impl<'block> RemoteCache for StateViewCache<'block> {
+    fn get(&self, access_path: &AccessPath) -> VMResult<Option<Vec<u8>>> {
+        StateViewCache::get(self, access_path)
+    }
+}
+
+// Adapter to convert a `StateView` into a `RemoteCache`.
+pub struct RemoteStorage<'a>(&'a dyn StateView);
```
vgao1996, line 84: I've been nagging about this for a long time but I still I don't think `dyn` should be used unless you have a heterogeneous collection or other esoteric use cases. 

With that being said let's not worry about this in this PR but think about whether we can replace them with parametric polymorphism.

language/move-vm/runtime/src/data_operations.rs
```diff
@@ -0,0 +1,109 @@
...
+    data_store::DataStore,
+    loaded_data::types::FatStructType,
+    values::{GlobalValue, Struct, Value},
+};
+use vm::errors::{vm_error, Location, VMResult};
+
+//
+// Provides an implementation for data store bytecodes and guarantees proper invariants
+// for each operation. It uses the `DataStore` for the data operation but it makes sure
+// all operations are consistent to Move semantic.
```
runtian-zhou, line 20: Just out of curiosity why do we want to have some logics to be in DataStore and the rest over here? Can we implement those logics directly in `data_cache.rs`?

language/move-vm/types/src/data_store.rs
vgao1996, line 1: Hmm so data_store.rs is in move-vm-types not move-vm-state?

language/move-vm/runtime/src/data_cache.rs
vgao1996: My take on this: 
- Concrete implementations like TransactionDataCache belongs to the runtime so I'm definitely fine moving them there. 
- I'm not too sure about the removal of the traits and forcing TransactionDataCache onto the clients. The way I see this is that TransactionDataCache is an additional layer/adapter on top of whatever DataStore the client passes in and it's kinda internal. Leaking it to the client seems a bit weird to me at the moment. Also consider that we now have folks trying to use Move & at least part of the Move VM on other blockchains, and I think a trait will provide more flexibility AND better isolation. With that being said, I'm 100% supportive of having a concrete implementation rather than a trait for things that are in the Libra VM/Adapter.

language/move-vm/runtime/src/data_operations.rs
dariorussi: the idea here is that those operations are "vital" to the system and should not and are not provided by the cache and particularly by the trait (given the trait is not providing an implementation).
So basically the functions here are internal to the VM and are called by the bytecodes or the natives and give the guarantees of proper implementation over the `DataStore` or data caches.
Caches and in general implementations of `DataStore` offer a much more direct and simple managing of the cache deprived of semantic that is a VM concern

dariorussi:
> Any specific reason why we pick `dyn DataStore` vs type parameter? IMHO we don't need to have the `dyn` here as we always pass in the very same struct all the time.

Fair point and I hear you!
I had issues with lifetime and strange stuff when I was doing the type parameter. Later on I found a lifetime issue unrelated to the type parameter vs `dan DataStore`. Now that everything works I can go back and see if adding the type parameter works just fine. So I'll let you know.
Overall though and as a side issue the `dyn` here is pretty immaterial in terms of implementation and performance, most operation on the data store are heavy (event being the only "simple" one) so I doubt the cost of the indirection is going to be significant.

vgao1996:
@runtian-zhou yeah I think it was more like a legacy -- the old code uses `dyn` and we always had higher priorities than trying to convert them into type parameters during refactoring like this.

But yeah I also think `dyn`s are unnecessary and confusing in this case and we should get rid of it in a follow up PR.

vgao1996:
@dariorussi that's a good point. I don't think performance is going to suck in practice because `dyn` in this case. Although I would argue
- However small it is, it's an unnecessary hit.
- It may be very confusing to a Rust programmer -- to me a `dyn` to some extend implies an esoteric use case like heterogeneous collections, and makes me dig through the code and try to find it, only to realize it's not the case here.

language/libra-vm/src/data_cache.rs
dariorussi: `RemoteCache` is a subset of `StateView` as you know and we can certainly use `StateView` instead. It is just a simpler API, it returns a `VMResult` and it is "internal" to the VM which why we put it there in the first place.
`ConfigStorage` is a different deal and certainly not what we want. The only API there is `fetch_config` (and besides the name it has the same signature as the `StateView`).
I'll see to remove `RemoteCache` in favor of `StateView` and we'll see where that goes

language/move-vm/runtime/src/data_operations.rs
dariorussi: sounds good (as above)

language/libra-vm/src/data_cache.rs
dariorussi: well let me look into it...

language/move-vm/types/src/data_store.rs
dariorussi: `move-vm-state` is gone, it's no more....
I am not sure about the `move-vm-types` name but that was always a mix of different things and a bit the standalone API that could be used with and without the runtime.
So not sure about the name but I could move it to `move-vm-runtime`

dariorussi:
for reasons that are a bit of a mystery to me I cannot reply to some of @vgao1996 comments so I am going to add something here...
traits vs concrete implementation: I agree with the principle but I also think that traits are better defined when the API is clear and the responsibilities are clear. I am fine locking the API and making it more strict in V1, which is where a bunch of this work is going.
The problem with all of this is that gas, data and events are not well defined and separated (client responsibilities, natives, VM guarantees) so the intent was to create a more restricted model first and then look at the result. If we see clear path to open things up we will, otherwise my inclination is to ship a more restricted system for V1. 
Having a concrete implementation gives you guarantees of what you are calling and make the code more static and obvious to review and audit for correctness.
So I would have been fine having no traits at all as a starter.

External users of move will be penalized a bit by this, but again not having a clear story for using Move does not help them already and they would have to hack something anyway. My hope is that even with a more restrictive story we can make it very clear which files to touch in order to have alternative semantic

dariorussi:
@runtian-zhou, @vgao1996 so I tried a couple of things and was not very successful so I gave up and maybe we'll do later. If any of you wanted to try and let me know I'd appreciate.
1- Removing `dyn DataStore` I cannot make it happen, there are a couple of place (a return of a function which take a DataStore and the function context in native_function) where that is problematic. I get errors I am not sure I understand and if it's not everywhere I don't see the value. So I'll look into this later if you guys don't have a chance to look into it
2- `RemoteCache`: it makes sense to me to have that because that way we do not expose `StateView` to the Move VM. I tried to simplify `ConfigStorage` but that is a mess too. It's entirely used in the Libra adapter so @runtian-zhou if you could do the right thing there I'd appreciate. `ConfigStorage` makes no sense to the MoveVM and it's there only to serve the Libra adapter so if you could find a way to remove it that would be wonderful.

I may actually check this in and talk with you all offline to see what is next

language/libra-vm/src/libra_vm.rs
```diff
@@ -639,15 +676,19 @@ impl LibraVM {
...
     }
 
     /// Run the epilogue of a transaction by calling into `EPILOGUE_NAME` function stored
     /// in the `ACCOUNT_MODULE` on chain.
-    fn run_epilogue<T: ChainState>(
+    fn run_epilogue(
         &self,
-        chain_state: &mut T,
+        data_store: &mut TransactionDataCache,
+        cost_strategy: &mut CostStrategy,
```
runtian-zhou, line 691: Why did `run_prologue` and `run_epilogue` has argument `cost_strategy: &mut CostStrategy,` but not for `run_writeset_prologue`?

language/libra-vm/src/libra_vm.rs
```diff
@@ -306,50 +321,66 @@ impl LibraVM {
...
+                get_transaction_output(
+                    &mut data_store,
+                    &cost_strategy,
+                    txn_data,
+                    VMStatus::new(StatusCode::EXECUTED),
+                )
+            })
         })
         .unwrap_or_else(|err| {
             self.failed_transaction_cleanup(
```
runtian-zhou, line 381: This is a question for @sblackshear  and @tzakian . When a transaction fails should we just charge the claimed max gas amount sent by the user instead of the exact number of gas that was consumed? That would make this logic a lot cleaner. The difference is that for example when we hit an assertion failure in move, we will charge for full gas instead the gas used till the assertion failure.

language/libra-vm/src/libra_vm.rs
dariorussi: the way the code is arranged `run_writeset_prologue` makes the cost strategy inside the function rather than taking one in. This is just what the code was (so to speak).
I think we should review the usage of the move vm inside libra at some point given the way that API is evolving, but it's for another time...

vgao1996:
@dariorussi sounds good to me! Let me try to play with the dyn later and see if I can get it removed.

PR #3916, opened 2020-05-15
Add transaction fee distribution per-currency

tzakian:
Adds transaction fee distribution for each currency to the system. 

In order to do this, the transaction fee distribution needs to be decoupled from the block-prologue in Move since it needs to be called once per-currency (which we can only get at the Rust level in the VM).

## Tests
Adds an additional test to make sure that multiple currencies are being distributed successfully in the block prologue.


language/libra-vm/src/libra_vm.rs
```diff
@@ -457,6 +457,10 @@ impl LibraVM {
         txn_data.sender = account_config::CORE_CODE_ADDRESS;
         txn_data.max_gas_amount = GasUnits::new(std::u64::MAX);
 
+        let registered_currencies =
```
runtian-zhou, line 460: Was wondering if we should persist this in LibraVM for performance.

language/move-lang/tests/functional/translated_ir_tests/transaction_fee_distribution/multi_currency.move
```diff
@@ -0,0 +1,87 @@
...
+        Transaction::assert(LibraAccount::balance<Coin1::T>(0xFEE) > 0, 7);
+        Transaction::assert(LibraAccount::balance<Coin2::T>(0xFEE) > 0, 8);
+
+        Transaction::assert(LibraAccount::balance<LBR::T>({{vivian}}) == 0, 9);
+        Transaction::assert(LibraAccount::balance<Coin1::T>({{vivian}}) == 0, 10);
+        Transaction::assert(LibraAccount::balance<Coin2::T>({{vivian}}) == 0, 11);
+    }
+}
+// check: EXECUTED
+
```
runtian-zhou, line 69: Add a test to register/unregister a coin and see if it's distributed correctly?

language/libra-vm/src/libra_vm.rs
tzakian: It's a good question! At least for now I would say "no" for two reasons, but more than open to reconsidering:
1. The registered currencies aren't core to the operation of the VM (at least in my view). So they shouldn't be held as part of the VM state.
2. I'm not sure what happens in terms of the cache invalidation if a currency is registered. Does the VM need to be restarted in this case?

language/move-lang/tests/functional/translated_ir_tests/transaction_fee_distribution/multi_currency.move
tzakian: Sadly I don't think this is possible in the functional tests right now unless we do something seriously hacky. Once we can publish modules under 0x0 in the functional tests, I'll go ahead and add a test for this in :)

PR #3917, opened 2020-05-15
[move] Value and Constant refactor

runtian-zhou:
## Motivation

Refactored MoveValue into `move_core_types` and cleaned up the constants used in `bytecode-verifier`, `compiler` and `move-prover`. This allows us to get rid of the `move-vm` dependency in crates mentioned above.

Yes

### DIFF
--- a/language/move-vm/types/src/values/values_impl.rs
+++ b/language/move-vm/types/src/values/values_impl.rs
@@ -2440,12 +2440,6 @@ impl Value {
         let ty = Self::constant_sig_token_to_type(&constant.type_)?;
         Value::simple_deserialize(&constant.data, &ty).ok()
     }
-
-    pub fn serialize_constant(type_: SignatureToken, value: Value) -> Option<Constant> {
-        let ty = Self::constant_sig_token_to_type(&type_)?;
-        let data = value.simple_serialize(&ty)?;
-        Some(Constant { data, type_ })
-    }
 }
 
 /***************************************************************************************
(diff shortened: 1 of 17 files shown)


language/move-vm/types/src/values/values_impl.rs
```diff
@@ -2517,4 +2512,49 @@ pub mod prop {
             (Just(layout), value_strategy)
         })
     }
+
+    impl ValueImpl {
+        pub fn as_move_value(&self, ty: &FatType) -> MoveValue {
```
dariorussi, line 2517: is this needed?

language/vm/src/constant.rs
```diff
@@ -0,0 +1,50 @@
+// Copyright (c) The Libra Core Contributors
```
dariorussi, line 1: I am not clear this should be in vm but it's fine for now.
Any thoughts on this?

language/move-vm/types/src/values/values_impl.rs
runtian-zhou: This is used for proptest. We generate random ValueImpl and map it to MoveValue, and do two rounds of serialization/deserialization test and make sure the bytes remains the same.

language/vm/src/constant.rs
runtian-zhou: I'm fine having this in vm right now. The reason is that `Constant` is a file format related concept, thus whenever an upstream is trying to make sense of the constants in the serialized modules it should import the `Constant` in `file_format`.

PR #3919, opened 2020-05-15
[move source language] Added support for signer

tnowacki:
## Motivation

- Added signer as a primitive resource type
- Added move_to builtin
- Added rules around transaction arguments
- Blocked by #3913, so only look at the top commit 

## Test Plan

- migrated tests

### DIFF
--- a/language/move-lang/src/hlir/translate.rs
+++ b/language/move-lang/src/hlir/translate.rs
@@ -6,7 +6,7 @@ use crate::{
     expansion::ast::Fields,
     hlir::ast::{self as H, Block},
     naming::ast as N,
-    parser::ast::{BinOp_, Field, FunctionName, ModuleIdent, StructName, Value_, Var},
+    parser::ast::{BinOp_, Field, FunctionName, Kind_, ModuleIdent, StructName, Value_, Var},
     shared::{unique_map::UniqueMap, *},
     typing::ast as T,
 };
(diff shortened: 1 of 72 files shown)


language/move-lang/src/hlir/translate.rs
```diff
@@ -1183,33 +1186,73 @@ fn use_tmp(var: Var) -> H::UnannotatedExp_ {
     }
 }
 
+// let (bf, arg) = match *bf {
```
sblackshear, line 1189: Intentional?

language/move-lang/src/typing/translate.rs
```diff
@@ -90,32 +90,50 @@ fn script(context: &mut Context, nscript: N::Script) -> T::Script {
...
-        format_comma(&["u8", "u64", "u128", "bool", "address", "vector<u8>"]),
-    );
-    context.error(vec![(mloc, mmsg), (*loc, tmsg)])
+    if idx != 0 && subtype_no_report(context, ty.clone(), signer_ref.clone()).is_ok() {
+        let mmsg = format!(
+            "Invalid parameter for script function '{}'",
+            context.current_function.as_ref().unwrap()
+        );
+        let tmsg = format!(
+            "{} is only supported for the first argument to a script",
```
sblackshear, line 131:
```suggestion
            "{} must be the first argument to a script",
```

language/move-lang/src/hlir/translate.rs
tnowacki: Good catch! Removed

Issue #3885, opened 2020-05-14
[Bug] State sync stuck trying to sync forked chain


mgorven:
# 🐛 Bug

The pre-mainnet upgrade from 1a8c6c5 to e64989b kept network compatibility, but the genesis changed. When validators were upgraded they were stuck trying to sync to the old chain and never started consensus, even when enough were upgraded to form quorum.

## Expected Behaviour

Consensus starts on the new chain once enough validators are upgraded.

PR #3921, opened 2020-05-16
[state sync] introduce timeout to consensus requests

phoenix-antigravity:
related to issue https://github.com/libra/libra/issues/3885
now if state sync won't make any progress for configured period of time, it will notify consensus about failed sync request
test is added

### DIFF
--- a/config/src/config/state_sync_config.rs
+++ b/config/src/config/state_sync_config.rs
@@ -16,6 +16,8 @@ pub struct StateSyncConfig {
     pub max_chunk_limit: u64,
     // valid maximum timeout limit for sanity check
     pub max_timeout_ms: u64,
+    // default timeout for sync request
+    pub sync_request_timeout_ms: u64,
 }
 
 impl Default for StateSyncConfig {
@@ -26,6 +28,7 @@ impl Default for StateSyncConfig {
             long_poll_timeout_ms: 30000,
             max_chunk_limit: 1000,
             max_timeout_ms: 120_000,
+            sync_request_timeout_ms: 60_000,
         }
     }
 }
--- a/state-synchronizer/src/coordinator.rs
+++ b/state-synchronizer/src/coordinator.rs
@@ -38,12 +38,13 @@ pub(crate) struct SyncRequest {
     // reach the target (the LI in the storage remains unchanged as if nothing happened).
     pub callback: oneshot::Sender<Result<()>>,
     pub target: LedgerInfoWithSignatures,
+    pub last_progress_tst: SystemTime,
 }
 
 /// message used by StateSyncClient for communication with Coordinator
 pub(crate) enum CoordinatorMessage {
     // used to initiate new sync
-    Request(SyncRequest),
+    Request(Box<SyncRequest>),
     // used to notify about new txn commit
     Commit(
         // committed transactions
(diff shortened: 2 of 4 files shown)


state-synchronizer/src/coordinator.rs
```diff
@@ -382,6 +383,9 @@ impl<T: ExecutorProxyTrait> SyncCoordinator<T> {
         self.check_subscriptions();
         self.peer_manager.remove_requests(local_version);
 
+        if let Some(mut req) = self.sync_request.as_mut() {
+            req.last_progress_tst = SystemTime::now();
```
zekun000, line 387: what is tst?

config/src/config/state_sync_config.rs
```diff
@@ -26,6 +28,7 @@ impl Default for StateSyncConfig {
             long_poll_timeout_ms: 30000,
             max_chunk_limit: 1000,
             max_timeout_ms: 120_000,
+            sync_request_timeout_ms: 60_000,
```
zekun000, line 31: this is probably too long, I think 10s is enough for a single chunk

bmaurer:
Dumb question: how does the timeout help in the case of #3885? In other words, what was the victim of the issue trying to do that had an infinite timeout? What happens after the timeout occurs that gets them back on track?

What would be the consequence if this timeout fired in a case without misconfiguration

zekun000:
@bmaurer I should have filled more context in the issue, here's what happened

1. one node doesn't upgrade and is on the old-chain with version 5 million.
2. victims upgrade their nodes and are on version 0 but the they share the same keys with the old chain.
3. victims receive valid LedgerInfo about version 5M and try to sync (this is infinite loop until it reaches the target version and blocks consensus) but was unable to finish because execution recognizes it's a fork
4. once we add the timeout, consensus would bail the sync and timeout and start to make progress (although will still stuck for the timeout every time it receives the LedgerInfo from the old chain)

I think it's a good enough fix for now given it's still a problem of forking ourselves, alternatives would be adding a chain id in LedgerInfo which anchors the genesis but won't help if the genesis is the same but the chain forks.

bmaurer:
I see. So in other words, the victim recieves an, apparently valid, LI for version X that is inconsistent with the latest committed LI the victim has. By handing back to consensus we get a chance to recieve consensus messages from other chains / make progress.

In this process, are we able to see that GetChunkResponse is returning an accumulator that is inconsistent with our latest known LI? At least IMHO in production if we were to start downloading a chunk and see a signed LI that conflicts with our accumulator, we'd want to panic / wait for operator intervention

zekun000:
@bmaurer yes, we're able to see the error messages `Fork happens because the current synced_trees doesn't match the txn list provided.` due to our fix to previous sev

Good point, we definitely should make such error much more visible in prod

bmaurer:
That error message seems pretty deep into the codebase... Wondering if there are places closer to where the chunk is validated that would catch this.

Also, if we do want to panic on this in prod... is a timeout the right thing to do?

zekun000:
This is where the chunk being validated I think.
If we do want to panic, we probably should panic directly where the error happens. I wonder if we can make some macro which compile to panic in prod and error in debug mode.

PR #3922, opened 2020-05-16
[management] improvements

davidiw:
* Cleaned up management/src/lib
* Enabled token from file per @mgorven  and BisonTrail's recommendation
* Added CreateWaypoint ability to management

### DIFF
--- a/config/management/src/key.rs
+++ b/config/management/src/key.rs
@@ -1,0 +1,73 @@
+// Copyright (c) The Libra Core Contributors
+// SPDX-License-Identifier: Apache-2.0
+
+use crate::{error::Error, SecureBackends};
+use libra_crypto::ed25519::Ed25519PublicKey;
+use libra_secure_storage::{Storage, Value};
+use std::convert::TryInto;
+use structopt::StructOpt;
+
+#[derive(Debug, StructOpt)]
+pub struct AssociationKey {
+    #[structopt(flatten)]
+    pub secure_backends: SecureBackends,
+}
+
+impl AssociationKey {
+    pub fn execute(self) -> Result<Ed25519PublicKey, Error> {
+        submit_key(
+            libra_global_constants::ASSOCIATION_KEY,
+            self.secure_backends,
+        )
+    }
+}
+
+#[derive(Debug, StructOpt)]
+pub struct OperatorKey {
+    #[structopt(flatten)]
+    pub secure_backends: SecureBackends,
+}
+
+impl OperatorKey {
+    pub fn execute(self) -> Result<Ed25519PublicKey, Error> {
+        submit_key(libra_global_constants::OPERATOR_KEY, self.secure_backends)
+    }
+}
+
+#[derive(Debug, StructOpt)]
+pub struct OwnerKey {
+    #[structopt(flatten)]
+    pub secure_backends: SecureBackends,
+}
+
+impl OwnerKey {
+    pub fn execute(self) -> Result<Ed25519PublicKey, Error> {
+        submit_key(libra_global_constants::OWNER_KEY, self.secure_backends)
+    }
+}
+
+fn submit_key(key_name: &str, secure_backends: SecureBackends) -> Result<Ed25519PublicKey, Error> {
+    let local: Box<dyn Storage> = secure_backends.local.try_into()?;
+    if !local.available() {
+        return Err(Error::LocalStorageUnavailable);
+    }
+
+    let key = local
+        .get_public_key(key_name)
+        .map_err(|e| Error::LocalStorageReadError(e.to_string()))?
+        .public_key;
+
+    if let Some(remote) = secure_backends.remote {
+        let key = Value::Ed25519PublicKey(key.clone());
+        let mut remote: Box<dyn Storage> = remote.try_into()?;
+        if !remote.available() {
+            return Err(Error::RemoteStorageUnavailable);
+        }
+
+        remote
+            .set(key_name, key)
+            .map_err(|e| Error::RemoteStorageWriteError(e.to_string()))?;
+    }
+
+    Ok(key)
+}
--- a/config/management/src/lib.rs
+++ b/config/management/src/lib.rs
@@ -5,6 +5,7 @@
 
 mod error;
 mod genesis;
+mod key;
 mod layout;
 mod secure_backend;
 mod validator_config;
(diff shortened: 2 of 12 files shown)


config/management/src/lib.rs
```diff
@@ -177,77 +174,16 @@ impl Command {
...
-
-            Self::write_u64(storage.as_ref(), &mut buffer, EPOCH);
-            Self::write_u64(storage.as_ref(), &mut buffer, LAST_VOTED_ROUND);
-            Self::write_u64(storage.as_ref(), &mut buffer, PREFERRED_ROUND);
-            Self::write_waypoint(storage.as_ref(), &mut buffer, WAYPOINT);
-
-            writeln!(buffer, "=================================================").unwrap();
-
-            Ok(buffer)
+        if let Command::Verify(verify) = self {
```
JoshLind, line 194: Nit: I always found the name for the "Verify" command a little odd (seeing as though it doesn't actually perform any checks). Do you see the code being updated to perform verification checks? If not, "Export" might be a better name (if the entire purpose is to just print it out to the console)?

config/management/src/key.rs
```diff
@@ -0,0 +1,73 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+use crate::{error::Error, SecureBackends};
+use libra_crypto::ed25519::Ed25519PublicKey;
+use libra_secure_storage::{Storage, Value};
+use std::convert::TryInto;
+use structopt::StructOpt;
+
+#[derive(Debug, StructOpt)]
+pub struct AssociationKey {
```
JoshLind, line 11: Nit: What's the motivation behind adding another layer of indirection? i.e., we now have Command::*Key (which provides an execute() call), and we also have crate::Key::*Key which provides execute(). I can see that it makes a little bit more sense in other cases (e.g., genesis, waypoint) where there's more going on, but here it might be a little bit overkill? At the same time though, I can see why you might do it for consistency... (no changes required, I just found it a little bit of an interesting pattern!)

config/management/src/lib.rs
```diff
@@ -39,6 +40,8 @@ pub mod constants {
 pub enum Command {
     #[structopt(about = "Submits an Ed25519PublicKey for the association")]
     AssociationKey(crate::key::AssociationKey),
+    #[structopt(about = "Create a waypoint and optionally store place it in a store")]
```
JoshLind, line 43: Nit: "... optionally store place..."? --> "Create a waypoint and optionally place it into a storage backend"?

config/src/config/secure_config.rs
```diff
@@ -126,3 +168,79 @@ impl OnDiskStorageConfig {
...
+                namespace: None,
+                server: "127.0.0.1:8200".to_string(),
+                token: Token::FromConfig(TokenFromConfig {
+                    token: "test".to_string(),
+                }),
+            },
+        };
+
+        let text_from_config = "
+[vault]
```
JoshLind, line 195: nit: is it common practice to do this kind of indentation for multi line strings? Why not match the indentation of the let statement?

config/src/config/secure_config.rs
```diff
@@ -88,7 +88,7 @@ pub struct VaultConfig {
     /// Vault's URL, note: only HTTP is currently supported.
     pub server: String,
     /// The authorization token for access secrets
```
JoshLind, line 90: nit: "for accessing secrets"?

config/src/config/secure_config.rs
```diff
@@ -126,3 +168,79 @@ impl OnDiskStorageConfig {
...
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use std::io::Write;
+
+    #[test]
+    fn test_token_config_parsing() {
```
JoshLind, line 178: I'd maybe split this into two tests? e.g., one called "test_from_token_config_parsing()" and one for test_from_disk_parsing()"?

config/management/src/lib.rs
```diff
@@ -207,7 +207,7 @@ pub struct SecureBackends {
     /// The local secure backend, this is the source of data. Secure
     /// backends are represented as a semi-colon deliminted key value
     /// pair: "k0=v0;k1=v1;...".  The current supported formats are:
-    ///     Vault: "backend=vault;server=URL;token=TOKEN"
+    ///     Vault: "backend=vault;server=URL;token=PATH_TO_TOKEN"
```
JoshLind, line 210: nit : are we not supporting both the PATH_TO_TOKEN and the TOKEN?

config/management/src/key.rs
```diff
@@ -0,0 +1,73 @@
...
+    if !local.available() {
+        return Err(Error::LocalStorageUnavailable);
+    }
+
+    let key = local
+        .get_public_key(key_name)
+        .map_err(|e| Error::LocalStorageReadError(e.to_string()))?
+        .public_key;
+
+    if let Some(remote) = secure_backends.remote {
```
JoshLind, line 60: Silly question (and not directly related to this PR), but under what circumstances do we see a remote not being defined, but allowing this method to still be called? The name "submit_key" suggests to me it's going to submit the key (somewhere), but surely if remote doesn't exist, this method should error? Rather than just silently failing? Unless there's a use-case I'm missing here

config/management/src/key.rs
davidiw: Well actually at this point in time, we could unify the code in lib if all the executes returned the same type. I'm not sure it is worth it. The second part was just consistency. I think it is better to follow this type of formula than to try to have 40+ line exceptions.

config/management/src/key.rs
davidiw: Good question, all values are returned and printed to the screen. So this tool let's the human also perform the operation in case you only have a local secure end point.

config/management/src/lib.rs
davidiw: haha

config/management/src/lib.rs
davidiw: Well this has a couple means:
1) Verify you're initialized properly if you look at the proposal, we call this NodeVerifier
2) Given timelines, I'll actually be adding an optional parameter for genesis (and maybe a remote store) to confirm correct formation

config/management/src/lib.rs
davidiw: Not for this tool, I was told explicitly if you let them use it this way, they will mess up, they will accidentally share their secrets :P

davidiw: e.g., here's the command I used -- oh there's my token ooops

config/src/config/secure_config.rs
davidiw: You know, I don't know, I figured cargo xfmt would fix this if it wasn't right.

PR #3927, opened 2020-05-17
[language] Fix delete in data cache across blocks

dariorussi:
Delete in the data cache was simply removing the entry from
the cache, which is very problematic for a resource published
in a block and removed in another.
Basically the current behavior does not expose the remove
in the cache which would (missing the entry) retrieve the value
from the remote data store.
Basically we would be able to access a resource deleted in a block
multiple times in that same block as if it was not deleted.

### DIFF
--- a/language/e2e-tests/src/tests/data_store.rs
+++ b/language/e2e-tests/src/tests/data_store.rs
@@ -1,0 +1,360 @@
+// Copyright (c) The Libra Core Contributors
+// SPDX-License-Identifier: Apache-2.0
+
+use crate::{account::AccountData, compile::compile_script_with_address, executor::FakeExecutor};
+use bytecode_verifier::VerifiedModule;
+use compiler::Compiler;
+use libra_types::{
+    account_config::LBR_NAME,
+    transaction::{Module, SignedTransaction, Transaction, TransactionPayload, TransactionStatus},
+    vm_error::{StatusCode, VMStatus},
+};
+
+#[test]
+fn move_from_across_blocks() {
+    let mut executor = FakeExecutor::from_genesis_file();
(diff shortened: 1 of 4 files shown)


language/e2e-tests/src/tests/data_store.rs
```diff
@@ -0,0 +1,415 @@
+// Copyright (c) The Libra Core Contributors
```
dariorussi, line 1: all those tests fail in master without the change showing the problem described in the summary.
Essentially the blocks with multiple transactions all succeed!!

dariorussi:
the change is just few lines of code in `language/move-vm/state/src/data_cache.rs`
The rest is testing

dariorussi:
> Wow, great catch! How did you notice this?
> 
Just working on #3910 and reviewing code... but yeah, annoying...

> I guess the functional testing framework puts each transaction in its own block. This means that although we have many remove/access tests there, they will not catch this issue.
> 
> One tx per block seems like a good default for the functional tests given that real-world versions of the scenarios being tested will likely play out across multiple blocks. But maybe we could consider a functional tests mode that aggregates txes without an explicit `//! block_prologue` directives into a single block and check that the behavior of that mode matches the behavior of the default mode?

That's right, it is an interesting problem and certainly something that would be good to expose in functional tests (@vgao1996). 
I was even wondering about functional test running into a mode where transactions are "randomly" grouped together in blocks....

vgao1996:
@dariorussi @sblackshear Thanks for the feedback! I'll keep this in mind when I refactor functional tests and see what I can do. It seems to me that running the same sequence of transactions with different groups is the answer to this problem.

PR #3939, opened 2020-05-18
[network] improve error reporting during connect

### LINKED ISSUE #3884: [Bug] Improve error reporting of safety-rules communication
# 🐛 Bug

Pre-mainnet seems to be having communication problems with safety-rules, but there is no logging on the validator side to indicate this. 

## Expected Behavior

Clear error messages about communication failures with safety-rules.

gregnazario:
## Motivation

Currently, there isn't error reporting during repeated failures to
connect to a server in the network lib, which makes debugging failures
difficult.  The debug level message should prevent too much logging,
while allowing us to debug issues. resolves #3884

Yes

## Test Plan

As requested in #3884, there should be logging in the event of connection failures on a client-server connection.  There should be something to test with directly from the errors that caused this issue.

## Related PRs

None

### DIFF
--- a/secure/net/src/lib.rs
+++ b/secure/net/src/lib.rs
@@ -86,7 +86,8 @@ impl NetworkClient {
             let mut stream = TcpStream::connect(self.server);
 
             let sleeptime = time::Duration::from_millis(100);
-            while stream.is_err() {
+            while let Err(e) = stream {
+                debug!("Failed to connect to upstream {} {:?}", self.server, e);
                 thread::sleep(sleeptime);
                 stream = TcpStream::connect(self.server);
             }


secure/net/src/lib.rs
```diff
@@ -87,6 +87,7 @@ impl NetworkClient {
 
             let sleeptime = time::Duration::from_millis(100);
             while stream.is_err() {
+                debug!("Failed to connect to upstream {} {:?}", self.server, stream);
```
davidiw, line 90: Ahh, you'd like to print out the error?

I think you'd probably want to swap this to a loop:

```
loop {
  if let Some(stream) = stream {
    break;
  } else if let Err(e) = stream {
    debug!(..)
  }
  ...
}
```

You can also look into match statements.

gregnazario: I think `while let` is what I'm going for, which is a `while` with a `match`.

secure/net/src/lib.rs
```diff
@@ -86,7 +86,8 @@ impl NetworkClient {
             let mut stream = TcpStream::connect(self.server);
 
             let sleeptime = time::Duration::from_millis(100);
-            while stream.is_err() {
+            while let Err(e) = stream {
+                debug!("Failed to connect to upstream {} {:?}", self.server, e);
```
zekun000, line 90: this probably would spam the logs, I'd recommend to try a few times and bail out

davidiw: It is worth considering what the right policy is here... I'm okay with this for now.

gregnazario: I'm considering coming back and putting a timeout on these function.

The problem here is it will continue to poll forever, even if the destination is gone.  Rather error out, and handle in the outside logic how to find another server.

I couldn't test it failing on a unit test without either adding this or running a multi-threaded test that first starts with the service not existing, and then showing up.

davidiw: I personally wouldn't want to ship this version in that fashion, since we don't have the facilities for recovery in place just yet. If this failed out, we'd be stuck with a dead binary, which could be the result of one process starting later than another. Say I start Libra-node at time 0 and safety rules 2 minutes later. If our timeout window is 1 minute, the process would be dead :/.

phlip9: do what connectivity manager does and apply exp. backoff w/ a max delay.

https://github.com/libra/libra/blob/master/network/src/connectivity_manager/mod.rs#L522
https://github.com/libra/libra/blob/master/network/src/validator_network/network_builder.rs#L401

davidiw: I support that!

PR #3940, opened 2020-05-18
[move-prover][BUG?]

DavidLDill:
The file tests/sources/functional/generic_bug has a reduced version of a
problem I ran into in association.move.  It is probably the case that
spec schema and apply do exactly what we said we wanted them to do, but
maybe that's not what we want.

This file has generic functions to add and remove "Privileges" (which
are just resources). "apply_for_privilege<Privilege>" creates the
resource PrivilegedCapability<Privilege> and stores it at the sender's
address, and "remove_privilege<Privilege>" removes that if the sender
== root_address().

"initialize" stores a PrivilegedCapability<T> at the root_address().

I tried to specify and prove an invariant that "the root address
always has a PrivilegedCapability<T>", which the Prover happily proved.
I don't think it understands that someone could call
remove_privilege<GenericBug::T> to remove the privilege.

There is a commented-out function stupid_root that does
remove_privilege<T>(root_address()).  If this function is commented in,
the prover reports a violation of the invariant.

I don't fully understand this behavior. I think it has something to do
with universal quantification over generic parameters, but haven't
thought through the details.  The key is that remove_privilege can be
used to remove any PrivilegedCapability if the sender is root,
including the PrivilegedCapability<T> that appears in the specified
property, but I'm not sure how to tell Boogie about that.

## Motivation

(Write your answer here.)

## Test Plan

## Related PRs

### DIFF
--- a/language/move-prover/tests/sources/functional/generic_bug.move
+++ b/language/move-prover/tests/sources/functional/generic_bug.move
@@ -1,0 +1,70 @@
+address 0x0 {
+
+module GenericBug {
+    use 0x0::Transaction;
+
+    resource struct PrivilegedCapability<Privilege> { }
+
+    struct T { }
+
+    public fun initialize() {
+        let sender = Transaction::sender();
+        Transaction::assert(sender == root_address(), 1000);
+        move_to_sender(PrivilegedCapability<T>{ });
+    }
+
+    // Publish a specific privilege under the sending account.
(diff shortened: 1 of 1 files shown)


wrwg:
No need to merge this. I copied the example and have a fix.

wrwg:
Fixed with #3952 which also includes the example. Closing.

PR #3944, opened 2020-05-18
Adding tests to get us to ~100% test coverage of Move stdlib

tzakian:
This PR adds a number of tests to get us almost to 100% test coverage of the Move standard library (technically, to 99.14%, see below for breakdown). The missing percentages are dead code/inaccessible. Whether we want to delete these pieces of code is a question I would be interested in hearing from people on -- the ones I haven't removed are pieces that are duplicate checks on data passed in in a private function (or untestable at the moment due to genesis constraints). 

## Coverage Summary

```
+-------------------------+
| Move Coverage Summary   |
+-------------------------+
Module 00000000000000000000000000000000::Debug
>>> % Module coverage: NaN
Module 00000000000000000000000000000000::Testnet
>>> % Module coverage: 100.00
Module 00000000000000000000000000000000::LibraTimestamp
>>> % Module coverage: 100.00
Module 00000000000000000000000000000000::Association
>>> % Module coverage: 100.00
Module 00000000000000000000000000000000::AccountLimits
>>> % Module coverage: 96.87
Module 00000000000000000000000000000000::Vector
>>> % Module coverage: 100.00
Module 00000000000000000000000000000000::Empty
>>> % Module coverage: 100.00
Module 00000000000000000000000000000000::AccountType
>>> % Module coverage: 100.00
Module 00000000000000000000000000000000::VASP
>>> % Module coverage: 100.00
Module 00000000000000000000000000000000::Unhosted
>>> % Module coverage: 100.00
Module 00000000000000000000000000000000::AccountTrack
>>> % Module coverage: 88.96
Module 00000000000000000000000000000000::LCS
>>> % Module coverage: NaN
Module 00000000000000000000000000000000::Hash
>>> % Module coverage: NaN
Module 00000000000000000000000000000000::Authenticator
>>> % Module coverage: 100.00
Module 00000000000000000000000000000000::Offer
>>> % Module coverage: 100.00
Module 00000000000000000000000000000000::Event
>>> % Module coverage: 92.25
Module 00000000000000000000000000000000::LibraConfig
>>> % Module coverage: 100.00
Module 00000000000000000000000000000000::RegisteredCurrencies
>>> % Module coverage: 100.00
Module 00000000000000000000000000000000::FixedPoint32
>>> % Module coverage: 100.00
Module 00000000000000000000000000000000::Libra
>>> % Module coverage: 100.00
Module 00000000000000000000000000000000::Coin1
>>> % Module coverage: 100.00
Module 00000000000000000000000000000000::Coin2
>>> % Module coverage: 100.00
Module 00000000000000000000000000000000::Compare
>>> % Module coverage: 100.00
Module 00000000000000000000000000000000::Option
>>> % Module coverage: 100.00
Module 00000000000000000000000000000000::Signature
>>> % Module coverage: NaN
Module 00000000000000000000000000000000::LibraTransactionTimeout
>>> % Module coverage: 100.00
Module 00000000000000000000000000000000::LibraAccount
>>> % Module coverage: 100.00
Module 00000000000000000000000000000000::ValidatorConfig
>>> % Module coverage: 100.00
Module 00000000000000000000000000000000::LibraSystem
>>> % Module coverage: 98.98
Module 00000000000000000000000000000000::TransactionFee
>>> % Module coverage: 100.00
Module 00000000000000000000000000000000::LibraWriteSetManager
>>> % Module coverage: 100.00
Module 00000000000000000000000000000000::LibraBlock
>>> % Module coverage: 100.00
Module 00000000000000000000000000000000::LBR
>>> % Module coverage: 100.00
Module 00000000000000000000000000000000::Genesis
>>> % Module coverage: 100.00
Module 00000000000000000000000000000000::LibraVMConfig
>>> % Module coverage: 100.00
Module 00000000000000000000000000000000::LibraVersion
>>> % Module coverage: 100.00
Module 00000000000000000000000000000000::SharedEd25519PublicKey
>>> % Module coverage: 100.00
+-------------------------+
| % Move Coverage: 99.14  |
+-------------------------+
```

## Dead and Uncovered Code
The dead pieces of code, and the reasoning for why these are actually dead pieces of code: 

### LibraSystem:
The check that the index is within the bounds of the validator set in `update_ith_validator` is dead: 
* `update_ith_validator_info_` is a private function, only called from within this loop in `update_and_reconfigure`:
```rust
while (i < size) {
            let validator_info_update = update_ith_validator_info_(validators, i);

            configs_changed = configs_changed || validator_info_update;
            i = i + 1;
        }; 
```
where  `size = Vector::length(validators)`. Therefore for all possible calls to `update_ith_validator_` `i < Vector::length(validators)` and the following code is unreachable: 
```rust
 fun update_ith_validator_info_(validators: &mut vector<ValidatorInfo>, i: u64): bool {
        let size = Vector::length(validators);
        if (i >= size) {
            return false <<< THIS IS DEAD CODE
        };
        let validator_info = Vector::borrow_mut(validators, i);
        let new_validator_config = ValidatorConfig::get_config(validator_info.addr);
        // check if information is the same
        let config_ref = &mut validator_info.config;

        if (config_ref == &new_validator_config) {
            return false
        };
        *config_ref = *&new_validator_config;

        true
    }
```

### AccountLimits
The uncovered pieces of code are unused assertions about testnet. In particular:
```rust
Transaction::assert(0x0::Testnet::is_testnet(), 10047);
```
in `update_deposit_limits`, and 
```rust
 Transaction::assert(0x0::Testnet::is_testnet(), 10048);
```
in `update_withdrawal_limits`. 

These are dead in the current system since they can only be called via `AccountTrack::update_deposit_limits` and `AccountTrack::update_withdrawal_limits` respectively. However, these also perform these testnet checks. 

The reason we may not want to remove these checks is because these may allow the prover to reason locally about whether these functions can be called in mainnet.

### AccountTrack
* The following code is dead:
```rust
 fun tracking_info(addr: address): (address, AccountLimits::Window) {
        if (AccountType::is_a<Unhosted::T>(addr)) {
            (
                Unhosted::limits_addr(),
                Unhosted::account_limits(AccountType::account_metadata<Unhosted::T>(addr))
            )
        } else if (VASP::is_vasp(addr)) {
            abort 3003 <<< THIS IS DEAD
        } else if (AccountType::is_a<Empty::T>(addr)) {
            abort 3002
        } else { // Can add more logic here as we add more account types
            abort 3001
        }
    }
```
This is because we always return early if `is_unlimited_account` for the address returns true. However, `is_unlimited_account` is a straightforward check that the account type is a VASP. Therefore, we can never call `tracking_info` on an account where `VASP::is_vasp` will return true as the system stands today.

* The following code is dead:
```rust
fun update_info(addr: address, info: AccountLimits::Window)
    acquires AccountLimitsCapability {
        if (AccountType::is_a<Unhosted::T>(addr)) {
            let unhosted_info = Unhosted::update_account_limits(info);
            AccountType::update_with_capability<Unhosted::T>(
                addr,
                unhosted_info,
                &borrow_global<AccountLimitsCapability>(singleton_addr()).update_cap
            )
        } 
       [ else if (VASP::is_vasp(addr)) {
            abort 3003
        } else if (AccountType::is_a<Empty::T>(addr)) {
            abort 3002
        } else { // Can add more logic here as we add more account types
            abort 3001
        }
       ] <<< THIS IS ALL DEAD CODE
    }
```

In the current system this code is dead: `tracking_info` is always called on the address before `update_info` is called at any of this function's call sites. Therefore, this code cannot be exercised as the code stands today.

### Event
The uncovered event code is:
```rust
 Transaction::assert(Transaction::sender() == 0xA550C18 || Transaction::sender() == 0xF1A95, 0);
Transaction::assert(addr == 0xA550C18 || addr == 0xF1A95, 0);
```
in `Event::initialization`. We need to be able to run functional tests pre-genesis in order to test this functionality. 

## Future

Need to work on getting this integrated in with MIRAI-bot or the like so that we can make sure this coverage doesn't go down over time.


sblackshear:
Amazing accomplishment!

I haven't had time to review all of the new tests yet, but a few thoughts on the dead bits:

- `update_ith_validator_info_`: I don't think there's anything wrong with this function, but it seems like we could make the code shorter/simpler by inlining it in the loop + skipping the redundant check?
- `AccountLimits` testnet checks: agree with you that we should probably keep them.
- `AccountTrack`: not concerned because I think this code goes away once we have `signer`
- `Event`: oh no, this is the  circular dependencies hack :'(. I hope we can get rid of this eventually (maybe `signer` will help there too?), but seems ok for now

dariorussi:
I think is fine to have the dead code you mention to be honest and I would not worry much about it. It's just a safety net for different usage of the API.
Of course if you want to rewrite them in some form to make sure we get to 100% that is fine too, but I would not worry much about that.

Great work!!!!

PR #3951, opened 2020-05-19
[network] Noise IX -> Noise IK

mimoo:
This pull request integrates our Noise IK implementation with network,
replacing the current Noise IX code.

Notable changes:

1. The network/noise crate has been renamed to noise-wrapper and has been moved to a module inside the network crate (network/src/noise).
2. The NoiseSocket type has been renamed to NoiseSession type, as NoiseSocket is the name of a protocol framework (https://noisesocket.org/spec/noisesocket/).
3. Noise handshake messages are not prefixed with a length field anymore (handshake messages have a known length).
4. The first handshake message now prepends a u64 timestamp (and includes it into the prologue part of noise) in the first handshake message. When receiving a handshake initialization message, a server verifies that the timestamp not older than 60 seconds in the past, and not drifting in the future for more than 10 seconds.
5. The noise code now implements two additional validation checks after parsing the first client message:
    - if a trusted peer hashmap is passed, it will check that the client's public key is in there
    - it will detect replay attacks by checking if the prologue's u64 timestamp has been seen before for that same public key


network/socket-bench-server/src/lib.rs
```diff
@@ -9,7 +9,7 @@ use futures::{
     sink::SinkExt,
     stream::{Stream, StreamExt},
 };
-use libra_crypto::test_utils::TEST_SEED;
+use libra_crypto::{test_utils::TEST_SEED, x25519, Uniform as _};
```
davidiw, line 12: why import Uniform as _, why not just Uniform, I haven't seen this style anywhere else

network/socket-bench-server/src/lib.rs
```diff
@@ -76,21 +76,31 @@ impl Args {
...
-pub fn build_tcp_noise_transport() -> impl Transport<Output = NoiseSocket<TcpSocket>> {
+pub fn build_tcp_noise_transport() -> impl Transport<Output = NoiseSession<TcpSocket>> {
     TcpTransport::default().and_then(move |socket, _addr, origin| async move {
         let mut rng: StdRng = SeedableRng::from_seed(TEST_SEED);
-        let noise_config = Arc::new(NoiseConfig::new_random(&mut rng));
-        let (_remote_static_key, socket) = noise_config.upgrade_connection(socket, origin).await?;
+        let private = x25519::PrivateKey::generate(&mut rng);
+        let dummy_public = private.public_key();
+        let noise_config = Arc::new(NoiseWrapper::new(private));
+        // TODO: get this from addr
```
davidiw, line 100: where is the address coming from, can we actually trust it?

network/src/noise_wrapper/handshake.rs
```diff
@@ -0,0 +1,387 @@
...
+
+        // check the client timestamp
+        // TODO: security logging (mimoo)
+        let now = time::SystemTime::now()
+            .duration_since(time::UNIX_EPOCH)
+            .expect("system clock should work");
+        if client_timestamp > now
+            && client_timestamp - now > time::Duration::from_secs(MAX_FUTURE_TIMESTAMP)
+        {
+            println!("DEBUG: 2");
```
davidiw, line 206: nice

network/src/noise_wrapper/handshake.rs
```diff
@@ -0,0 +1,387 @@
...
+            return Err(io::Error::new(
+                io::ErrorKind::InvalidData,
+                format!(
+                    "noise: client initiated connection with a timestamp too far in the future: {}",
+                    client_timestamp_u64
+                ),
+            ));
+        } else if now.checked_sub(client_timestamp).unwrap()
+            > time::Duration::from_secs(EXPIRATION_TIMESTAMP)
+        {
```
davidiw, line 217: maybe push these condition checks into their own lines above so we have nice clean if too_far_into_the future { ... } else too_old { ... }

network/src/noise_wrapper/handshake.rs
```diff
@@ -0,0 +1,387 @@
...
+                )
+            })?;
+
+        // make sure the public key is a validator before continuing (if we're in the validator network)
+        if let Some(trusted_peers) = trusted_peers {
+            let found = trusted_peers
+                .read()
+                .unwrap()
+                .iter()
+                .find(|(_peer_id, public_keys)| public_keys.identity_public_key == their_public_key)
```
davidiw, line 250: seems a little interesting we don't really care if the peer knows its own identity lol, what's the point in passing this in?

network/src/noise_wrapper/handshake.rs
```diff
@@ -0,0 +1,387 @@
...
+            .0
+            .parse_client_init_message(&prologue, &client_init_message)
+            .map_err(|e| {
+                io::Error::new(
+                    io::ErrorKind::InvalidData,
+                    format!("noise: wrong message received {}", e),
+                )
+            })?;
+
+        // make sure the public key is a validator before continuing (if we're in the validator network)
```
davidiw, line 244: I guess the better phrasing here is if authentication is enabled, verify it

network/src/noise_wrapper/handshake.rs
```diff
@@ -0,0 +1,387 @@
...
+                    format!(
+                        "noise: validator client initiated connection with an unknown public key: {}",
+                        their_public_key
+                    ),
+                ));
+            }
+        }
+
+        // check the timestamp is not a replay
+        {
```
davidiw, line 266: why the namespacing?

network/src/noise_wrapper/handshake.rs
```diff
@@ -0,0 +1,387 @@
...
+                        "noise: client initiated connection with a timestamp already seen before: {}",
+                        client_timestamp_u64
+                    ),
+                ));
+                }
+            }
+        }
+
+        // construct and send the response
+        let mut rng = rand::rngs::OsRng;
```
davidiw, line 284: probably a good case for an HKDF?

network/src/noise_wrapper/handshake.rs
```diff
@@ -0,0 +1,387 @@
...
+            .map_err(|e| {
+                io::Error::new(
+                    io::ErrorKind::InvalidData,
+                    format!("noise: wrong message received {}", e),
+                )
+            })?;
+        socket.write_all(&server_response).await?;
+
+        // the connection succeeded, store the client timestamp for replay prevention
+        {
```
davidiw, line 298: why are we putting this into a namespace?

network/src/noise_wrapper/mod.rs
```diff
@@ -0,0 +1,85 @@
...
+//! use memsocket::MemorySocket;
+//! use crate::NetworkPublicKeys;
+//! use libra_crypto::{x25519, ed25519, Uniform, PrivateKey, test_utils::TEST_SEED};
+//! use rand::{rngs::StdRng, SeedableRng};
+//! use libra_types::PeerId;
+//! use libra_config::config::NetworkPeerInfo;
+//! use std::{collections::HashMap, sync::{Arc, RwLock}};
+//!
+//! # fn example() -> std::io::Result<()> {
+//! // create client and server NoiseWrapper
```
davidiw, line 25: maybe first generate client, then server and not interleave?

network/src/transport.rs
```diff
@@ -213,19 +213,40 @@ pub fn build_memory_noise_transport(
...
+                        io::ErrorKind::Other,
+                        "SHOULD NOT HAPPEN: remote public key not found",
+                    ))
+                }
+            };
+            let (remote_static_key, socket) = noise_config
+                .upgrade_connection(socket, origin, remote_public_key, Some(&trusted_peers))
+                .await?;
 
             // TODO(philiphayes): reenable after seed peers are always fully rendered
```
davidiw, line 235: How is this going to be the remote peers fully fleshed address? I think we still need to keep around a list of valid public keys and ensure that the authed pubkey is in the list.

crypto/crypto/src/noise.rs
```diff
@@ -680,8 +685,7 @@ impl NoiseSession {
         nonce.extend_from_slice(&self.read_nonce.to_be_bytes());
         let nonce = GenericArray::from_slice(&nonce);
 
-        let (buffer, authentication_tag) =
-            ciphertext.split_at_mut(ciphertext.len() - AES_GCM_TAGLEN);
+        let (buffer, authentication_tag) = message.split_at_mut(message.len() - AES_GCM_TAGLEN);
```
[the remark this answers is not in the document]
mimoo: since these are "encrypt in place" functions having it named ciphertext or plaintext is weird. I can split this commit in several commits if that helps reviews.

network/socket-bench-server/src/lib.rs
mimoo: in general I import traits like this, since you don't use it anywhere else you can't have a name collisions later on

network/socket-bench-server/src/lib.rs
mimoo: that's a bencher : o

network/src/noise_wrapper/handshake.rs
mimoo: tests are still not passing :P

network/src/noise_wrapper/handshake.rs
mimoo: not sure I understand your question : o

network/src/noise_wrapper/handshake.rs
mimoo: to release the lock once we're done

network/src/noise_wrapper/handshake.rs
mimoo: not sure what you mean here

network/src/noise_wrapper/handshake.rs
mimoo: see above

crypto/crypto/src/noise.rs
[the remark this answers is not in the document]
davidiw: nah, just a heads up for the future, that way I focus on the important stuff and not get distract... I don't have a good attention span :)

network/socket-bench-server/src/lib.rs
davidiw: On the one hand, I agree with you. Importing traits is dumb unless you intend to explicitly use that in a box / interface. Should we make this consistent across the code base? you know let's bike shed a little ;)

network/socket-bench-server/src/lib.rs
```diff
@@ -76,21 +76,31 @@ impl Args {
 }
 
 /// Build a MemorySocket + Noise transport
-pub fn build_memsocket_noise_transport() -> impl Transport<Output = NoiseSocket<MemorySocket>> {
+pub fn build_memsocket_noise_transport() -> impl Transport<Output = NoiseSession<MemorySocket>> {
```
davidiw, line 79: we should really figure out how to merge this kinda stuff with transport.rs

network/socket-bench-server/src/lib.rs
davidiw: I guess I still don't understand... there's no address here, so how do we intend on getting this from an address? I could imagine extracting an address from the outgoing socket... but I don't think that API really exists. More importantly, do we really care about extracting "addresses" from sockets?

network/src/noise_wrapper/handshake.rs
```diff
@@ -0,0 +1,387 @@
...
+        TSocket: AsyncRead + AsyncWrite + Unpin,
+    {
+        // Perform protocol negotiation
+        // TODO: should we really be doing this here?
+        let (socket, proto) = match origin {
+            ConnectionOrigin::Outbound => {
+                negotiate_outbound_interactive(socket, [NOISE_PROTOCOL]).await?
+            }
+            ConnectionOrigin::Inbound => negotiate_inbound(socket, [NOISE_PROTOCOL]).await?,
+        };
```
davidiw, line 97: No, this is explicitly supposed to be noise IK... if it fails, I would say it is like talking to an HTTP server using Gopher or some nonsense like that.

network/socket-bench-server/src/lib.rs
mimoo: honestly I wish we would, could be enforced with a linter too (cc @sunshowers )

network/socket-bench-server/src/lib.rs
mimoo: updated this part, like other parts of the code it now returns None if there's no public key, and only errors if there is no public key AND we are dialing

```
let remote_public_key = addr.find_noise_proto();
```

network/socket-bench-server/src/lib.rs
phlip9: idk it makes the imports kinda noisy : (
you get things like
```rust
use futures::{
    future::{Future as _, FutureExt as _},
    stream::{FusedStream as _, Stream as _, StreamExt as _},
};
```

network/src/noise_wrapper/handshake.rs
phlip9: yeah remove all the in-band negotation, since we don't need it anymore

in fact, we should remove all of netcore/negotiate

network/src/noise_wrapper/handshake.rs
```diff
@@ -0,0 +1,390 @@
...
+    /// Perform a protocol upgrade on an underlying connection. In addition perform the noise IX
+    /// handshake to establish a noise session and exchange static public keys. Upon success,
+    /// returns the static public key of the remote as well as a NoiseSession.
+    // TODO(mimoo, philp9): this code could be inlined in transport.rs once the monolithic network is done
+    pub async fn upgrade_connection<TSocket>(
+        &self,
+        socket: TSocket,
+        origin: ConnectionOrigin,
+        anti_replay_timestamps: Arc<RwLock<AntiReplayTimestamp>>,
+        remote_public_key: Option<x25519::PublicKey>,
```
phlip9, line 81: when you rebase on master, we should be able to remove the `Option`

network/src/noise_wrapper/handshake.rs
```diff
@@ -0,0 +1,390 @@
...
+        trusted_peers: Option<&Arc<RwLock<HashMap<PeerId, NetworkPeerInfo>>>>,
+    ) -> io::Result<(x25519::PublicKey, NoiseSession<TSocket>)>
```
phlip9, line 83: since the protocol is not symmetric, we should have two separate functions

`upgrade_inbound` for when you're a listener
`upgrade_outbound` for when you're a dialer

the args will also change (e.g., we don't need origin, we don't need replay timestamps if dialer, we don't need remote_public_key if listener)

network/socket-bench-server/src/lib.rs
mimoo: it'd be great if we could do 

```
use futures::{
    future::{_Future, _FutureExt},
    stream::{_FusedStream, _Stream, _StreamExt},
};
```

network/src/noise_wrapper/handshake.rs
mimoo: I can do this in another commit :D don't want to break anything here

network/src/noise_wrapper/session.rs
```diff
@@ -557,163 +426,205 @@ where
...
-            // -> e, s
-            self.receive().await?;
+/// Collection of buffers used for buffering data during the various read/write states of a
+/// NoiseSession
+struct NoiseBuffers {
+    /// A read buffer, used for both a received ciphertext and then for its decrypted content.
+    read_buffer: [u8; noise::MAX_SIZE_NOISE_MSG],
+    /// A write buffer, used for both a plaintext to send, and then its encrypted version.
+    write_buffer: [u8; noise::MAX_SIZE_NOISE_MSG],
+}
```
phlip9, line 444: awesome! only two buffers now! : p

network/src/noise_wrapper/handshake.rs
```diff
@@ -0,0 +1,390 @@
...
+                ));
+            }
+        }
+
+        // perform the noise handshake
+        let socket = match origin {
+            ConnectionOrigin::Outbound => {
+                let remote_public_key = match remote_public_key {
+                    Some(key) => key,
+                    None if cfg!(test) => unreachable!(),
```
phlip9, line 113: is this so errors will panic in tests/fuzzing? probably use `panic` instead of `unreachable`, or consider removing this.

network/src/noise_wrapper/handshake.rs
phlip9: let's remove the in-band negotiation in this diff and clean out netcore/negotiate later?

network/src/noise_wrapper/handshake.rs
```diff
@@ -0,0 +1,390 @@
...
+    /// Create a new NoiseConfig with the provided keypair
+    pub fn new(key: x25519::PrivateKey) -> Self {
+        Self(noise::NoiseConfig::new(key))
+    }
+
+    /// Perform a protocol upgrade on an underlying connection. In addition perform the noise IX
+    /// handshake to establish a noise session and exchange static public keys. Upon success,
+    /// returns the static public key of the remote as well as a NoiseSession.
+    // TODO(mimoo, philp9): this code could be inlined in transport.rs once the monolithic network is done
+    pub async fn upgrade_connection<TSocket>(
```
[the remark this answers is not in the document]
mimoo: yes! I was waiting for monolithic transport to remove this :)

network/src/noise_wrapper/handshake.rs
```diff
@@ -0,0 +1,390 @@
...
+// Noise by default is not aware of the above or lower protocol layers,
+// We thus need to build this wrapper around Noise to both:
+//
+// - fragment messages that need to be encrypted by noise (due to its maximum 65535-byte messages)
+// - understand how long noise messages we send and receive are,
+//   in order to pass them to the noise implementaiton
+//
+
+/// TODO: why do we expose this to the outer protocol?
+const NOISE_PROTOCOL: &[u8] = b"/noise_ik_25519_aesgcm_sha256/1.0.0";
```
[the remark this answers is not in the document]
mimoo: I'll delete this in the next PR to remove the negotiation if that's alright?

network/src/noise_wrapper/handshake.rs
```diff
@@ -0,0 +1,390 @@
...
+// this is not a measure against real validators spamming us
+
+/// we're willing to tolerate a client timestamp 10 seconds in the future
+const MAX_FUTURE_TIMESTAMP: u64 = 10;
+
+/// we store client timestamps (to prevent replay) for up to 60 seconds
+const EXPIRATION_TIMESTAMP: u64 = 60;
+
+/// hashmap to store client timestamps if a connection succeeds
+pub type AntiReplayTimestamp = HashMap<x25519::PublicKey, u64>;
```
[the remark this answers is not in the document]
mimoo: ah indeed! Hadn't think about that

network/src/noise_wrapper/session.rs
```diff
@@ -1,270 +1,102 @@
...
-/// A Noise session with a remote
+/// A Noise session with a remote peer.
 ///
 /// Encrypts data to be written to and decrypts data that is read from the underlying socket using
-/// the noise protocol. This is done by wrapping noise payloads in u16 (big endian) length prefix
-/// frames.
+/// the noise protocol. This is done by prefixing noise payloads with a u16 (big endian) length field.
 #[derive(Debug)]
-pub struct NoiseSocket<TSocket> {
+pub struct NoiseSession<TSocket> {
```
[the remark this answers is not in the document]
mimoo: due to noise socket being a protocol already : o could change that to EncryptedSession which is more descriptive

network/src/noise_wrapper/handshake.rs
mimoo: actually only in tests, but yeah I'd want that in fuzzing too I guess. This should be unreachable though, why not use unreachable? That should give us a pretty good warning if we ever reach that panic

network/src/noise_wrapper/handshake.rs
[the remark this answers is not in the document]
mimoo: actually, it should be pretty easy to remove here

network/src/noise_wrapper/handshake.rs
```diff
@@ -0,0 +1,390 @@
...
+        let client_timestamp_u64 = u64::from_le_bytes(prologue);
+        let client_timestamp = time::Duration::from_secs(client_timestamp_u64);
+
+        // check the client timestamp
+        // TODO: security logging (mimoo)
+        let now = time::SystemTime::now()
+            .duration_since(time::UNIX_EPOCH)
+            .expect("system clock should work");
+        if client_timestamp > now
+            && client_timestamp - now > time::Duration::from_secs(MAX_FUTURE_TIMESTAMP)
```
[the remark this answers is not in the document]
mimoo: what do you propose?

If we don't rely on the timestamp being correct then we do need to measure the drift to know when it can't be replayed OR we need to send them our timestamp before they start a handshake with us :D

network/src/noise_wrapper/handshake.rs
mimoo: ah, well best practice is to not use HKDF and use OsRng :)

network/src/noise_wrapper/handshake.rs
```diff
@@ -0,0 +1,390 @@
...
+        {
+            // if the client timestamp is too far in the future, abort
+            return Err(io::Error::new(
+                io::ErrorKind::InvalidData,
+                format!(
+                    "noise: client initiated connection with a timestamp too far in the future: {}",
+                    client_timestamp_u64
+                ),
+            ));
+        } else if now.checked_sub(client_timestamp).unwrap()
```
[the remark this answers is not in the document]
mimoo: ^ this is also not good, refactoring this a bit based on you and davidiw's comment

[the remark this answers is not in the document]
mimoo: replaced with a match, I think it's clearer now

network/src/noise_wrapper/handshake.rs
mimoo: replaced with a match, I think it's clearer now

network/src/noise_wrapper/handshake.rs
mimoo: isn't this more accurate? If I was reading the code and read "if authentication is enabled" I would then have more questions : D

network/src/noise_wrapper/handshake.rs
```diff
@@ -0,0 +1,390 @@
...
+                ));
+            }
+        }
+
+        // check the timestamp is not a replay
+        {
+            let timestamps = anti_replay_timestamps.read().unwrap();
+            if let Some(timestamp) = timestamps.get(&their_public_key) {
+                // TODO: security logging the ip + blocking the ip? (mimoo)
+                if timestamp == &client_timestamp_u64 {
```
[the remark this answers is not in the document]
mimoo: I thought about it, but why prevent someone to connect again with an older timestamp? It is kind of weird, and maybe it's a good idea to just prevent it anyway : o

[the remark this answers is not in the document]
mimoo: timestamp is in seconds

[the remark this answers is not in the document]
mimoo: so two things about a monotonic clock:

* I do currently accept dialers that present a timestamp in the past
* should we expect an honest dialer to dial twice in less than 60sec?
* We also currently just replace the previously seen timestamp when we see a connection from the same dialer, meaning that if one dials twice in less than 60sec then the first handshake message can be used in a replay attack

network/src/noise_wrapper/handshake.rs
```diff
@@ -0,0 +1,390 @@
...
+                )
+            })?;
+        socket.write_all(&server_response).await?;
+
+        // the connection succeeded, store the client timestamp for replay prevention
+        {
+            let mut timestamps = anti_replay_timestamps.write().unwrap();
+            timestamps
+                .entry(their_public_key)
+                .and_modify(|e| *e = client_timestamp_u64)
```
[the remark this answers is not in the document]
mimoo: as I said earlier, if you connect twice in less than 60sec, then the first handshake message can then be used in a replay attack.

so perhaps we can modify this to have a Vec of timestamp instead...

network/src/noise_wrapper/handshake.rs
phlip9: we use `trusted_peers` if we have mutual authentication enabled. which is true not just in the validator network, but also a validator's internal full node network.

docker/validator-dev/build.sh
```diff
@@ -28,7 +28,7 @@ else
   docker run --name libra-node-builder-container -d -i -t -v $DIR/../..:/libra $PROXY_RUN libra-node-builder bash
 fi
 
-docker exec -i -t libra-node-builder-container bash -c 'source /root/.cargo/env; cargo build --release -p libra-node --target-dir /target && /bin/cp /target/release/libra-node target/libra-node-builder/'
+docker exec -i -t libra-node-builder-container bash -c 'source /root/.cargo/env; RUSTFLAGS="-Ctarget-cpu=sandybridge -Ctarget-feature=+aes,+sse2,+sse4.1,+ssse3" cargo build --release -p libra-node --target-dir /target && /bin/cp /target/release/libra-node target/libra-node-builder/'
```
phlip9, line 31: is this exhaustive? there's probably more docker builds floating around no?

network/src/noise_wrapper/session.rs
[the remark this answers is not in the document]
mimoo: perfect

network/src/noise_wrapper/session.rs
```diff
@@ -1,270 +1,102 @@
...
-    match ready!(poll_read_exact(context, socket, buf, offset)) {
-        Ok(()) => Poll::Ready(Ok(Some(u16::from_be_bytes(*buf)))),
-        Err(e) => {
-            if *offset == 0 && e.kind() == io::ErrorKind::UnexpectedEof {
-                return Poll::Ready(Ok(None));
-            }
-            Poll::Ready(Err(e))
-        }
+    #[cfg(any(test, feature = "fuzzing"))]
+    pub fn take_socket(self) -> TSocket {
```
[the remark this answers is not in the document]
mimoo: makes sense : o

network/src/noise_wrapper/handshake.rs
```diff
@@ -0,0 +1,348 @@
...
+    sync::{Arc, RwLock},
+    time,
+};
+
+use libra_config::config::NetworkPeerInfo;
+use libra_crypto::{noise, x25519};
+use libra_types::PeerId;
+use netcore::transport::ConnectionOrigin;
+
+use crate::noise_wrapper::session::NoiseSession;
```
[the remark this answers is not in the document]
mimoo: you mean without the space before? I was splitting external/internal/crate imports

docker/validator-dev/build.sh
mimoo: oops, I didn't mean to add this in this commit. Yes there are more but not sure where it should be added exactly : o

network/src/noise_wrapper/handshake.rs
```diff
@@ -0,0 +1,348 @@
...
+/// If the client timestamp is valid, we store it.
+#[derive(Default)]
+pub struct AntiReplayTimestamps(HashMap<x25519::PublicKey, u64>);
+
+impl AntiReplayTimestamps {
+    /// Returns true if the timestamp has already been observed for this peer
+    /// or if it's an old timestamp
+    pub fn is_replay(&self, pubkey: x25519::PublicKey, timestamp: u64) -> bool {
+        if let Some(last_timestamp) = self.0.get(&pubkey) {
+            &timestamp <= last_timestamp
```
gregnazario, line 45: I'm new to rust.  Why not dereference last_timestamp and compare u64s vs &u64s?

network/src/noise_wrapper/handshake.rs
```diff
@@ -0,0 +1,348 @@
...
+    where
+        TSocket: AsyncRead + AsyncWrite + Unpin,
+    {
+        // on the validator network, send prologue as current timestamp (in seconds)
+        let mut prologue = [0u8; 8];
+        if validator_network {
+            let now: u64 = time::SystemTime::now()
+                .duration_since(time::UNIX_EPOCH)
+                .expect("system clock should work")
+                .as_secs();
```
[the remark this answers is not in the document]
gregnazario: I would rather us be at millisecond granularity at the network layer and set us up for being able to tune it later, we don't see it often now, but we may see it later.  Most networking happens on the order of milliseconds between datacenters.  What's the cost implication of trying again?  I'm still catching up a little on what this exactly does :)

crypto/crypto/src/noise.rs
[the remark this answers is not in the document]
mimoo: we do use it to calculate `MAX_WRITE_BUFFER_LENGTH` outside of crypto/noise.

I actually like this better inside of crypto/noise itself since we do use AES-GCM directly and this is more explicit : o

network/src/noise_wrapper/handshake.rs
mimoo: I'm actually not sure why I need to borrow here, I was expecting rust to do it automatically, but I wouldn't expect borrowing or dereferencing to change how anything gets compiled eventually.

docker/client/Dockerfile
```diff
@@ -16,6 +16,7 @@ FROM toolchain AS builder
 
 COPY . /libra
 
+ENV RUSTFLAGS "-Ctarget-cpu=sandybridge -Ctarget-feature=+aes,+sse2,+sse4.1,+ssse3"
```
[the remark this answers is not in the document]
phlip9: cc @mgorven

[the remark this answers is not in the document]
mgorven: Sandybridge is quite old, I'd be surprised if we're using it anyway, but that's far from a guarantee. I'd expect virtualisation to be fine though.

network/src/noise_wrapper/handshake.rs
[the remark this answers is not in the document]
phlip9: i'd rather we don't conflate rate limiting with replay attack detection.

instead of importing another crate, let's just do:

```rust
let now: u128 = time::SystemTime::now()
    .duration_since(time::UNIX_EPOCH)
    .expect("system clock should work")
    .as_nanos();
let payload: [u8; 16] = now.to_le_bytes();
```

let's also send the timestamp in the encrypted payload, since we're not actually using it until we parse the whole client init message. this would also hide highly structured data (nanos timestamp) from being visible over-the-wire.

docker/validator-dev/build.sh
mimoo: now added all of them in the last commit

docker/client/Dockerfile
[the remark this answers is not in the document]
mimoo: good question

network/src/noise_wrapper/session.rs
```diff
@@ -1,270 +1,102 @@
 // Copyright (c) The Libra Core Contributors
 // SPDX-License-Identifier: Apache-2.0
 
-//! Noise Socket
```
phlip9, line 4: rename file `session` -> `stream` : p

network/src/noise_wrapper/handshake.rs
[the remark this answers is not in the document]
phlip9: convention seems to be all imports in one block and let cargo fmt handle it

network/src/noise_wrapper/handshake.rs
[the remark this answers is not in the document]
mimoo: why nanos though? If I understand the dailer correctly it's only going to take milliseconds to have another dial right?

mimoo:
rebased

docker/client/Dockerfile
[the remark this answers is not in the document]
tarcieri: Hi, author of the `aes-gcm`, `ghash`, and `polyval` crates here.

Sandy Bridge was the first Intel microarchitecture which supported CLMUL, so it's effectively the minimum viable target. The instructions are forward-compatible.

That said, if possible I'd recommend optimizing for Skylake, which introduced an AVX-512 version of the CLMUL instruction (VCLMULQDQ) which will provide optimal performance.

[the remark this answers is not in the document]
mimoo: Thanks @tarcieri !

will change that to `skylake`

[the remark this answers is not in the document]
mimoo: according to [this thread](https://www.reddit.com/r/rust/comments/6ynm53/a_simple_tip_to_improve_rust_program_speed/) we should probably change to `target-cpu=native`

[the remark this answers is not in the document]
mimoo: actually it sounds like this is what `cargo build --release` does o.o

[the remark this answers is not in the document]
phlip9: `target-cpu=native` is probably too restrictive. i would only compile with `target-cpu=native` if the container is run on the exact same machine as it's compiled on. we can probably assume prod will be at least `skylake`, so i'm comfortable with that.

network/socket-bench-server/src/main.rs
```diff
@@ -34,13 +34,20 @@ fn main() {
...
 
+    // passed as TCP_NOISE_ADDR=/ip4/1.2.3.4/tcp/1234
     if let Some(addr) = args.tcp_noise_addr {
-        let addr = start_stream_server(&executor, build_tcp_noise_transport(), addr);
+        let (noise_transport, public_key) = build_tcp_noise_transport();
+        let addr = start_stream_server(&executor, noise_transport, addr);
+
+        // display full networkaddr
+        let handshake_version = 0;
+        let addr = addr.append_prod_protos(public_key, handshake_version);
```
phlip9, line 50: oops, you only want to append the `/ln-noise-ik` protocol, since these bench Transports are not running the `/ln-handshake` protocol.

mimoo: ah, well I removed this, I'll add this to another PR

network/Cargo.toml
```diff
@@ -38,10 +38,10 @@ libra-types = { path = "../types", version = "0.1.0" }
 libra-workspace-hack = { path = "../common/workspace-hack", version = "0.1.0" }
 memsocket = { path = "memsocket", version = "0.1.0" }
 netcore = { path = "netcore", version = "0.1.0" }
-noise = { path = "noise", version = "0.1.0" }
 num-variants = { path = "../common/num-variants", version = "0.1.0" }
 proptest = { version = "0.10.0", default-features = true, optional = true }
 stream-ratelimiter = { path = "../common/stream-ratelimiter", version = "0.1.0" }
+rand_core = { version = "0.5.1", optional = true }
```
huitseeker, line 44: I would have expected to see this one in `[dev-dependencies]`?

network/src/noise_wrapper/handshake.rs
```diff
@@ -0,0 +1,356 @@
+// Copyright (c) The Libra Core Contributors
+// SPDX-License-Identifier: Apache-2.0
+
+//! The handshake module implements the handshake part of the protocol.
+//! This module also implements additional anti-DoS mitigation,
```
huitseeker, line 5: I'm sceptical of the effectiveness of DoS mitigation through statefulness on an 'I' protocol (by which the initiator doesn't really need to prove anything but knowledge of the receiver's identity on entry) ... do we want an open issue to let admission control introduce some token delivery in the mid-term future, which would allow us to upgrade this to a KK framework?

network/src/noise_wrapper/handshake.rs
```diff
@@ -0,0 +1,356 @@
...
+                )
+            })?;
+
+        // make sure the public key is a validator before continuing (if we're in the validator network)
+        if let Some(trusted_peers) = trusted_peers {
+            let found = trusted_peers
+                .read()
+                .unwrap()
+                .iter()
+                .any(|(_peer_id, public_keys)| public_keys.identity_public_key == their_public_key);
```
huitseeker, line 221: That's a `find` you're looking for here.

network/src/noise_wrapper/handshake.rs
```diff
@@ -0,0 +1,356 @@
...
+                return Err(io::Error::new(
+                    io::ErrorKind::InvalidData,
+                    "noise: client initiated connection without an 8-byte timestamp",
+                ));
+            }
+            let mut client_timestamp = [0u8; PAYLOAD_SIZE];
+            client_timestamp.copy_from_slice(&payload);
+            let client_timestamp = u64::from_le_bytes(client_timestamp);
+
+            // check the timestamp is not a replay
```
huitseeker, line 248: 1. How does the replay detection interact with a retry? In particular, if the client for some reason does not receive the responser's message, is there a chance that its retry of the Handshake would be flagged as a replay?
2. If the answer to the previous question is yes, how do you recommend we proceed on responder message loss?
3. Is this replay detection meant to guard against the server cost of performing a few DH operations, or is it meant to protect the server against client impersonation?

network/src/noise_wrapper/handshake.rs
```diff
@@ -0,0 +1,356 @@
...
+            let mut anti_replay_timestamps = anti_replay_timestamps.write().unwrap();
```
huitseeker, line 249: This is returning a `Result`, would it be appropriate to replace the `unwrap()` by a `?`, here?

network/src/noise_wrapper/handshake.rs
```diff
@@ -0,0 +1,356 @@
...
+                    ),
+                ));
+            }
+
+            // store the timestamp
+            anti_replay_timestamps.store_timestamp(their_public_key, client_timestamp);
+        }
+
+        // construct the response
+        let mut rng = rand::rngs::OsRng;
```
huitseeker, line 266: is this something we want to make seedable in the `pub(crate)` api so that this can be unit-tested?

network/src/noise_wrapper/handshake.rs
```diff
@@ -0,0 +1,356 @@
...
+        let mut server_response = [0u8; noise::handshake_resp_msg_len(0)];
+        let session = self
+            .0
+            .respond_to_client(&mut rng, handshake_state, None, &mut server_response)
+            .map_err(|e| {
+                io::Error::new(
```
huitseeker, line 272: Would an ` impl From<ThisError> for ThatError` help here (and in multiple other places)?

network/src/noise_wrapper/mod.rs
```diff
@@ -0,0 +1,92 @@
+// Copyright (c) The Libra Core Contributors
+// SPDX-License-Identifier: Apache-2.0
+
+//! This crate implements wrappers around our [Noise][noise] implementation.
+//! Noise is a protocol framework to encrypt and authentication connections.
+//! We use Noise to secure connections between peers in Libra.
+//! Specifically, we use the [Noise IK][ik] handshake which is a one round-trip protocol
+//! (the client sends one message, then the server responds).
+//! For more information about Noise and our implementation, refer to the [crypto] crate.
```
huitseeker, line 9: I missed the part where the [crypto] crate (are you referring to `libra_crypto`?) contained information about Noise?

network/src/noise_wrapper/stream.rs
```diff
@@ -557,163 +426,204 @@ where
...
+    TSocket: AsyncRead,
+{
+    match ready!(poll_read_exact(context, socket, buf, offset)) {
+        Ok(()) => Poll::Ready(Ok(Some(u16::from_be_bytes(*buf)))),
+        Err(e) => {
+            if *offset == 0 && e.kind() == io::ErrorKind::UnexpectedEof {
+                return Poll::Ready(Ok(None));
+            }
+            Poll::Ready(Err(e))
+        }
```
huitseeker, line 514: let res = match ready!(poll_read_exact(context, socket, buf, offset)) {
      Ok(()) => Ok(Some(u16::from_be_bytes(*buf)))),
      Err(io::ErrorKind::UnexpectedEof) if *offset == 0 =>  Ok(None),
      Err(e) => Err(e),
   };
   Poll::Ready(res)
Would something like this help?

network/src/noise_wrapper/handshake.rs
mimoo: `any() =  find() + some()`

network/Cargo.toml
mimoo: fuzzing is not run as dev : o

network/src/noise_wrapper/handshake.rs
mimoo: the DoS mitigation is not about the handshake pattern, but about sending a payload as part of the first message to avoid replays : o

network/src/noise_wrapper/handshake.rs
mimoo: 1. good question! We thought about that so we made the timestamp accurate to the millisecond (and retries happen after X ms, with X configurable)

3. cost of performing a few DH operations

network/src/noise_wrapper/handshake.rs
mimoo: yes! Thanks for catching this : ) 

the only reason this would fail is because the lock has become poisoned. I don't see any reason this would happen since we don't run this in threads, but if it ever can happen we have no reason to panic and we should just log and continue to use our current connections.

network/src/noise_wrapper/handshake.rs
mimoo: yes! This is the plan for another PR (especially because this will be annoying for fuzzing)

network/src/noise_wrapper/mod.rs
mimoo: yeah libra_crypto contains our implementation of noise

network/src/noise_wrapper/stream.rs
mimoo: I'm actually just copying this code, and it works, so I'm not touching it : D

network/src/noise_wrapper/handshake.rs
mimoo: yes! Good one :) looking into this

mimoo: urg, io::Error is in a different crate, so can't do this.

network/src/noise_wrapper/handshake.rs
huitseeker: Yeah. But that's my point: you can, at modest cost, generate a large number of valid distinct first messages which will mandate an answer.

network/src/noise_wrapper/mod.rs
huitseeker: 1. `[crypto]`  will not doc-link to `libra_crypto`
2. how does your reader interpret the word `Noise` to refer to "Libra's single-pattern implementation of Noise"?

network/src/noise_wrapper/stream.rs
huitseeker: The answer is usually to put the change into a single commit, so that it's independently reversible.

network/src/noise_wrapper/handshake.rs
huitseeker: As a client implementer, where do I find the minimal latency for retries?

PR #3954, opened 2020-05-19
[crypto] Use Serde name for seeding

matbd:
## Motivation

* (breaking) Make hash seeds more specification-friendly and tooling-friendly by using the Serde name instead of Rust qualified name. Typically, a struct `Foo` has seed: `sha3_256("LIBRA::Foo")` instead of `sha3_256("some::rust::namespace::Foo@@$$LIBRA$$@@")`

* Add support for generics in derive macros CryptoHasher and LCSCryptoHash.
* Add a unit test to verify that Serde names don't clash between corpuses (or rather when they do, the Serde format must be identical).

## Test Plan

```
cargo x test -p libra-crypto -p generate-format
```

## Related PRs

https://github.com/libra/libra/pull/3845
https://github.com/libra/libra/pull/3895

### DIFF
--- a/crypto/crypto/src/hash.rs
+++ b/crypto/crypto/src/hash.rs
@@ -41,7 +41,7 @@
 //! ```
 //! use libra_crypto::hash::CryptoHash;
 //! use libra_crypto_derive::{CryptoHasher, LCSCryptoHash};
-//! use serde::{Serialize, Deserialize};
+//! use serde::{Deserialize, Serialize};
 //! #[derive(Serialize, Deserialize, CryptoHasher, LCSCryptoHash)]
 //! struct MyNewStruct { /*...*/ }
 //!
@@ -63,14 +63,14 @@
 //! use the derive macro [`CryptoHasher`](https://doc.rust-lang.org/reference/procedural-macros.html).
 //!
 //! ```
-//! # use libra_crypto_derive::CryptoHasher;
-//! # use serde::Deserialize;
+//! use libra_crypto_derive::CryptoHasher;
+//! use serde::Deserialize;
 //! #[derive(Deserialize, CryptoHasher)]
 //! #[serde(rename = "OptionalCustomSerdeName")]
 //! struct MyNewStruct { /*...*/ }
 //! ```
 //!
-//! The macro will define a hasher automatically called `MyNewStructHasher`, and derive a salt
+//! The macro `CryptoHasher` will define a hasher automatically called `MyNewStructHasher`, and derive a salt
 //! using the name of the type as seen by the Serde library. In the example above, this name
 //! was changed using the Serde parameter `rename`: the salt will be based on the value `OptionalCustomSerdeName`
 //! instead of the default name `MyNewStruct`.
--- a/crypto/crypto/src/hash.rs
+++ b/crypto/crypto/src/hash.rs
@@ -41,8 +41,8 @@
 //! ```
 //! use libra_crypto::hash::CryptoHash;
 //! use libra_crypto_derive::{CryptoHasher, LCSCryptoHash};
-//! use serde::Serialize;
-//! #[derive(Serialize, CryptoHasher, LCSCryptoHash)]
+//! use serde::{Serialize, Deserialize};
+//! #[derive(Serialize, Deserialize, CryptoHasher, LCSCryptoHash)]
 //! struct MyNewStruct { /*...*/ }
 //!
 //! let value = MyNewStruct { /*...*/ };
@@ -62,16 +62,18 @@
 //! For any new structure `MyNewStruct` that needs to be hashed, it is recommended to simply
 //! use the derive macro [`CryptoHasher`](https://doc.rust-lang.org/reference/procedural-macros.html).
 //!
-//! ```ignore
-//! #[derive(CryptoHasher)]
-//! struct MyNewStruct {
-//!   ...
-//! }
+//! ```
+//! # use libra_crypto_derive::CryptoHasher;
+//! # use serde::Deserialize;
+//! #[derive(Deserialize, CryptoHasher)]
+//! #[serde(rename = "OptionalCustomSerdeName")]
+//! struct MyNewStruct { /*...*/ }
 //! ```
 //!
-//! The macro will define a hasher automatically called `MyNewStructHasher`, and pick a salt
-//! equal to the full module path + "::"  + structure name, i.e. if
-//! `MyNewStruct` is defined in `bar::baz::quux`, the salt will be `b"bar::baz::quux::MyNewStruct"`.
+//! The macro will define a hasher automatically called `MyNewStructHasher`, and derive a salt
+//! using the name of the type as seen by the Serde library. In the example above, this name
+//! was changed using the Serde parameter `rename`: the salt will be based on the value `OptionalCustomSerdeName`
+//! instead of the default name `MyNewStruct`.
 //!
 //! ## Customized hashers
 //!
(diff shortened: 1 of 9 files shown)


crypto/crypto/src/hash.rs
```diff
@@ -62,16 +62,18 @@
...
 //! use the derive macro [`CryptoHasher`](https://doc.rust-lang.org/reference/procedural-macros.html).
 //!
-//! ```ignore
-//! #[derive(CryptoHasher)]
-//! struct MyNewStruct {
-//!   ...
-//! }
+//! ```
+//! # use libra_crypto_derive::CryptoHasher;
+//! # use serde::Deserialize;
```
huitseeker, line 67: We may actually want to print those imports in documentation.

crypto/crypto/src/hash.rs
```diff
@@ -62,16 +62,18 @@
...
+//! struct MyNewStruct { /*...*/ }
 //! ```
 //!
-//! The macro will define a hasher automatically called `MyNewStructHasher`, and pick a salt
-//! equal to the full module path + "::"  + structure name, i.e. if
-//! `MyNewStruct` is defined in `bar::baz::quux`, the salt will be `b"bar::baz::quux::MyNewStruct"`.
+//! The macro will define a hasher automatically called `MyNewStructHasher`, and derive a salt
+//! using the name of the type as seen by the Serde library. In the example above, this name
+//! was changed using the Serde parameter `rename`: the salt will be based on the value `OptionalCustomSerdeName`
+//! instead of the default name `MyNewStruct`.
```
huitseeker, line 76: While we're speaking of "the macro", perhaps we want to desambiguate between `CryptoHasher` and `LCSCryptoHash` (and maybe even have a sentence about which does what)?

crypto/crypto/src/unit_tests/cryptohasher.rs
```diff
@@ -33,51 +41,43 @@ impl CryptoHash for Bar {
...
-    );
+    assert_eq!(&expected, actual.as_ref());
+}
+
+#[test]
+fn test_lcs_cryptohash_with_generics() {
+    let value = Baz { a: 5u64, b: 1025 };
+    let expected = CryptoHash::hash(&Foo { a: 5, b: 1025 });
+    let actual = CryptoHash::hash(&value);
+    assert_eq!(expected.as_ref(), actual.as_ref());
```
huitseeker, line 82: Nice :) do we want support for lifetimes?

crypto/crypto/src/hash.rs
```diff
@@ -41,8 +41,8 @@
 //! ```
 //! use libra_crypto::hash::CryptoHash;
 //! use libra_crypto_derive::{CryptoHasher, LCSCryptoHash};
-//! use serde::Serialize;
-//! #[derive(Serialize, CryptoHasher, LCSCryptoHash)]
+//! use serde::{Serialize, Deserialize};
```
davidiw, line 44: maybe order alphabetically? otherwise it is rather unclear how we are ordering it

crypto/crypto/src/hash.rs
```diff
@@ -62,16 +62,18 @@
...
 //!
-//! ```ignore
-//! #[derive(CryptoHasher)]
-//! struct MyNewStruct {
-//!   ...
-//! }
+//! ```
+//! # use libra_crypto_derive::CryptoHasher;
+//! # use serde::Deserialize;
+//! #[derive(Deserialize, CryptoHasher)]
```
davidiw, line 68: same comment on ordering

crypto/crypto/src/hash.rs
```diff
@@ -109,7 +111,7 @@ use serde::{de, ser};
 use std::{self, convert::AsRef, fmt, str::FromStr};
 use tiny_keccak::{Hasher, Sha3};
 
-pub(crate) const LIBRA_HASH_SUFFIX: &[u8] = b"@@$$LIBRA$$@@";
+pub(crate) const LIBRA_HASH_PREFIX: &[u8] = b"LIBRA::";
```
davidiw, line 114: haha, finally...

thefallentree:
I suggest we add an explicit name attributes, the last thing you want is that to be source code dependent, that throw hitless refactoring out of the window!

Yucong Sun, Libra PE

matbd:
@thefallentree
* As you know, we are now logging Serde formats in unit-test-checked YAML files. The vision is that we will lock these files at some point to prevent a wide range of breaking changes in public-facing formats (including public-facing names).
* Therefore, I believe it is both desirable and totally adequate to use Serde names for seeding hash function.
    * This PR solves the current issue that fully qualified Rust names are indeed implicit and not stable under refactoring.
    * From a security point of view, I'm also introducing new tooling to detect seed collisions based on the YAML files.

bmwill:
Yeah I'm sort of with @thefallentree on this one. While having a mechanism which you can derive is super useful, it makes things error prone during refactoring.

matbd:
> we need to discuss how to prevent accidental breaking changes .
> 
> As implemented, any refactoring would change the hash, right?

Can we be specific about the notion of "refactoring"?
* Before this PR, changing the name of the struct OR any of the high-level namespaces (crate and modules) would silently change the seed.
* After this PR, if you change the name of the struct or equivalently, add `#[serde(rename = "something")]`, it will visibly change the format logged in the corresponding YAML file (by breaking the unit test).

davidiw:
@thefallentree, we share your sentiment, this is a huge step toward that direction. Furthermore, as @matbd stated, if someone changes the name of a struct or moves it, its hash would change. At least now, if they move it, its hash wouldn't change.

I believe the requested change was made due to a misunderstanding, please follow up by end of day.

davidiw:
> Is this PR actually an breaking change as is? Does the name matches the old one?
> 
> I agree, it seems like we are able to avoid breaking changes by adding serde_rename tags , why not make this change hitless by adding the rename tag too ?
> 
> As being said, I'm not blocking this change to go through so we can make progress, but I want to note that every time we don't practice "breakage avoiding" we are just saying we can do "non breaking changes" , but not actually doing it.

I think at this point in time, we should avoid accumulating tech debt and instead favor breaking changes. This PR is fantastically simplifying the means that we do hashing and makes it easier to implement clients.

It should absolutely be a breaking change and push that direction or it will not achieve its goal.

Thank you for rescinding your "request changes".

bmwill:
After taking another look I misunderstood the goal of this PR. Fixing the hash see to the "name" of the struct vs the fully qualified path of the struct is a huge leap in the right direction. It does require that all structs are appropriately "named" though as renaming a struct without  using an appropriate `#[serde(rename = "")]` would also result in a hash change. But that is much better than what we currently have.

Sorry for the misunderstanding.

matbd:
I updated the summary to clarify what is breaking in the PR. The longer explanation is that:
* I removed the very long Rust-centric namespaces (e.g. `network::network::v1::protocol::`) from the salts. We could totally maintain them using `#[serde(rename = ...)]` but I'd have to inspect each definition manually and could still make a mistake (no unit tests on seeds before this PR).
* I took this opportunity to further change the salting schema to use a clean looking `LIBRA::` prefix instead of a funny looking `@@$$LIBRA$$@@` suffix.
* Deploying in test net will definitely require trashing the blockchain but I didn't find the value `@@$$LIBRA$$@@` anywhere in the python Libra client (I suspect it is buried in the Rust/C library) so I suspect there is no extra work here.

crypto/crypto/src/unit_tests/cryptohasher.rs
matbd: I don't get the comment :) but it looks like I can remove the two `.as_ref()`.

crypto/crypto/src/hash.rs
matbd: I see

crypto/crypto/src/hash.rs
matbd: That one, I'm going to pass because afaik rustfmt does not care.

crypto/crypto/src/unit_tests/cryptohasher.rs
matbd: The current code does indeed fail on lifetimes :( I'm going to defer this to another PR, though.

PR #3955, opened 2020-05-19
[move-prover] Specification of association.move, partial specs for li…

DavidLDill:
[move-prover] Specification of association.move, partial specs for libra.move

For review, I'm interested in possible errors and, especially, omissions.  Thoughts
about style of both Move and specs are welcome.

Both are partially completed, but association.move is more complete
than libra.move.  They're not yet using the doc gen format.

Some problems in the prover were found and some issues in the move code
were found.  Comments describe these, along with explanations of what's
going on and a certain amount of rumination about style issues.

## Motivation

It's the whole point of the Move Prover project.

yes
## Test Plan

## Related PRs

### DIFF
--- a/language/stdlib/modules/association.move
+++ b/language/stdlib/modules/association.move
@@ -24,6 +24,10 @@ module Association {
 
     // A type tag to mark that this account is an association account.
     // It cannot be used for more specific/privileged operations.
+
+    // DD: The presence of an instance of T at and address, with is_certified,
+    // means that the address is an association address. I suggest giving "T"
+    // a more meaningful name (e.g., AssociationMember? AssociationPrivileges?)
     struct T { }
 
     // Initialization is called in genesis. It publishes the root resource
@@ -61,6 +65,7 @@ module Association {
     // Remove the `Privilege` from the address at `addr`. The sender must
     // be the root association account. The `Privilege` need not be
     // certified.
+    // DD: Perhaps should not allow root to remove itself from Association?
     public fun remove_privilege<Privilege>(addr: address)
     acquires PrivilegedCapability {
         assert_sender_is_root();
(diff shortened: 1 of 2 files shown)


language/stdlib/modules/association.move
```diff
@@ -109,6 +114,156 @@ module Association {
...
+    }
+    spec schema OnlyRootAddressHasRootPrivilege {
+        invariant root_invariant();
+    }
+    spec module {
+        // apply to all functions except "initialize" and functions that "initialize" calls
+        // before the invariant is established.
+        // Note: All the called functions OBVIOUSLY cannot affect the invariant.
+        // TODO: Try to find a better approach to this that does not require excepting functions.
+        apply OnlyRootAddressHasRootPrivilege to *<T>, *
```
wrwg, line 162: We should open a bug that one only needs to provide one `*` here. I think the rule should be that all type parameters on the left side must be bound by type parameters in the pattern. A schema with no type parameters can be matched against any function.

language/stdlib/modules/association.move
```diff
@@ -109,6 +114,156 @@ module Association {
...
+    spec module {
+        // enclose this in a function so we can easily repeat the same property in the invariant
+        // and as a post-condition to initialize (the base case).
+        // SPEC: Only the root address has a Root{} resource.
+        define root_invariant(): bool {
+            all(addresses(), |a| exists<Root>(a) ==> (a == spec_root_address()))
+        }
+    }
+    spec schema OnlyRootAddressHasRootPrivilege {
+        invariant root_invariant();
```
wrwg, line 155: I guess this should be `invariant module`.

language/stdlib/modules/association.move
```diff
@@ -109,6 +114,156 @@ module Association {
...
+    }
+    spec module {
+        // apply to all functions except "initialize" and functions that "initialize" calls
+        // before the invariant is established.
+        // Note: All the called functions OBVIOUSLY cannot affect the invariant.
+        // TODO: Try to find a better approach to this that does not require excepting functions.
+        apply OnlyRootAddressHasRootPrivilege to *<T>, *
+            except initialize, root_address, has_privilege, addr_is_association,
```
wrwg, line 163: Why do we need to exclude anything else than initialize? While those other functions do not violate the module invariant, it doesn't hurt to apply the module invariant to them.

I would expect that in general, we only exclude the special case of initialize. This gives us peace of mind regards soundness of that the module invariant can be assumed for external calles.

language/stdlib/modules/association.move
```diff
@@ -109,6 +114,156 @@ module Association {
...
+        ensures sender() == spec_root_address();
+    }
+
+    // ### authority to set is_certified flag
+
+    // SPEC: only grant_* functions can set the is_certified flag in
+    // PrivilegedCapability<Privilege>.
+    // Other functions can create and remove the Privilege object, so we have to condition
+    // everything on its existence.
+    spec schema OnlyGrantCanCertify<Privilege> {
```
wrwg, line 191: This schema would be associated with function `assert_sender_is_root` in the generated doc, so you probably want to move it into the module stuff before any function, where the other apply is.

language/stdlib/modules/association.move
```diff
@@ -109,6 +114,156 @@ module Association {
...
+       ensures all(addresses(),
+                   |a| old(exists<PrivilegedCapability<Privilege>>(a)
+                           ==> !global<PrivilegedCapability<Privilege>>(a).is_certified)
```
wrwg, line 194: This is really hard to read. Perhaps put the implication into a helper function? You can use old on helper functions: `old(helper()) ==> helper()`.

Also, is this really true? It is not the case that for all those functions, is_certified is false. What is true is that if it was false before, then they must not set it to true.

language/stdlib/modules/association.move
```diff
@@ -109,6 +114,156 @@ module Association {
...
+                        ==> (exists<PrivilegedCapability<Privilege>>(a)
+                             ==> !global<PrivilegedCapability<Privilege>>(a).is_certified));
+    }
+    spec module {
+        // By excepting only grant_*, we make sure only these two functions
+        // can change the is_certified from true to false.
```
wrwg, line 200: false to true?

language/stdlib/modules/association.move
```diff
@@ -109,6 +114,156 @@ module Association {
...
+        // TODO: Try deleting the association version to see if prover catches it.
+        apply OnlyGrantCanCertify<Privilege> to *<Privilege>
+            except grant_privilege, grant_association_address;
+    }
+
+    // ### Functions can't make privileges disappear, except remove_privilege
+
+    // SPEC: Only remove_privilege can remove privileges
+    spec schema OnlyRemoveCanRemovePrivileges<Privilege> {
```
wrwg, line 209: Again, this schema need to be move before any `spec fun` or `spec struct`.

language/stdlib/modules/association.move
```diff
@@ -109,6 +114,156 @@ module Association {
...
+         ensures any(addresses(), |a| (old(exists<PrivilegedCapability<Privilege>>(a))
+                                       && !exists<PrivilegedCapability<Privilege>>(a))
+                                       ==> sender() == spec_root_address());
+    }
+    spec module {
+        // FEATURE REQUEST: We need to be able to apply this to functions that don't have
```
wrwg, line 215: I think it would be clearer to instead allow quantification over types as part of the unbound quantification you already introduced for addresses, instead of letting the apply operator implicitly introduce one.

```
schema OnlyRemoveCanRemovePrivileges {
     invariant module all(types(), |t| ...t...)
}
```

That we need types as first class values is something Junkil already bumped on.

But this all depends on how well the unbound quantification actually works. From syntax viewpoint, once we have general unbounded quantification, we may also want to revisit the syntax.

language/stdlib/modules/association.move
```diff
@@ -109,6 +114,156 @@ module Association {
...
+    // ## Management of privileges
+    // TODO: With grant/remove etc., separately specify that only root may grant or remove privileges
+
+    // ### Management of Root {} marker.
+
+    // The root_address is marked by a Root{} object that is stored at that place only.
+    spec module {
+        // enclose this in a function so we can easily repeat the same property in the invariant
+        // and as a post-condition to initialize (the base case).
+        // SPEC: Only the root address has a Root{} resource.
```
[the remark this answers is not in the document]
DavidLDill: I'm trying to mark English versions of the formal specs.  I was experimenting with a convention where every logical spec is also written in English (separately from other explanation, etc.)

Does that seem like a good idea?

language/stdlib/modules/association.move
DavidLDill: There's just a bug in my spec, not in the Prover, right?

language/stdlib/modules/association.move
DavidLDill: Applying the module invariant puts the invariant in as a "requires".  But, when the function (e.g. root_address) is called from initialize, the invariant doesn't hold, and so their is an error reported because of the precondition that doesn't hold.

These functions only need the invariant as a precondition to prove the invariant as a post-condition.  So, they don't really care whether the invariant holds when they are called during initialization.

I think there is a better approach to this problem (I don't think it requires changes to the verifier, just to the specifications).  I will think about it some more.

language/stdlib/modules/association.move
DavidLDill: That seems like a good solution.

language/stdlib/modules/association.move
```diff
@@ -109,6 +114,156 @@ module Association {
...
+
+// # Specifications for association.move
+
+// This is preliminary.  This is one of the first real module libraries with global
+// specifications, and everything is evolving.
+
+    spec module {
+        pragma verify = true;
+
+        // This mirrors the Move function root_address()
```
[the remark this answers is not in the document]
JoelMarcey: The Google doc is good (I am going to check internally if we have our own guidelines written somewhere).

Maybe this has already been discussed at some other time - I am curious why we are not following the [Rust Documentation Comment format and policies](https://doc.rust-lang.org/book/ch14-02-publishing-to-crates-io.html#making-useful-documentation-comments), given how much relationship there is to Move and Rust.

Re: referring to another function - generally, yes, the function should be public if the documentation referencing it is public. And you could use [xxx](LINK) to point to the file that would be created when generating the documentation for that function.

language/stdlib/modules/association.move
```diff
@@ -109,6 +114,156 @@ module Association {
...
+        define spec_root_address(): address { 0xA550C18 }
```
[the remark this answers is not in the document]
DavidLDill: I wonder if it's worth having a slightly different name to remind the user that they are different functions. I'm worried about the case where someone changes root_address but not spec_root_address.

I suppose the ideal solution would be able to use pure functions from Move as functions in specs.  (Emma asked for this feature in a "wishlist" a long time ago, but it wasn't submitted as an issue).

Unfortunately, I'm not sure how to implement something like this easily.  We could translate a subset of the AST for the source code, and bail if there are any side effects?

language/stdlib/modules/association.move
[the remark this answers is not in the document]
DavidLDill: Ok, in this case I want to say it's the same as root_address, because results will be confusing if they diverge. I spelled out that it must be consistent. We don't have to have the source code for root_address in front of us to understand it.  I changed "This".

language/stdlib/modules/association.move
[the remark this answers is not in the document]
DavidLDill: I changed SPEC: to "Informally:".  I think it would be a good idea to restate every property in English.  It makes it more readable, and, if someone spots a discrepancy between the English and the spec language, it might mean there's an error in the formal spec.

Issue #3925, opened 2020-05-17
[state-sync] Make StateSynchronizer the source of truth for reconfiguration


davidiw:
StateSync provides a subscription service for sending off on chain config updates. Registration occurs [here](https://github.com/libra/libra/blob/master/libra-node/src/main_node.rs#L262) and delivery [here](https://github.com/libra/libra/blob/master/state-synchronizer/src/executor_proxy.rs#L218).

Network still depends on consensus to [update eligible peers](https://github.com/libra/libra/blob/master/consensus/src/epoch_manager.rs#L380), we need to eliminate this dependency. Also it is important to note that this API is incomplete as this only updates the networking keys and not the networking addresses.

To eliminate this, we need to funnel reconfiguration events into networking by adding a subscription service for networking that calls the appropriate APIs on connectivity manager: [UpdateEligibleAddresses and UpdateEligibleNodes](https://github.com/libra/libra/blob/master/network/src/connectivity_manager/mod.rs#L100-L102) or consider making a composite API -- but not recommended for MVP.

So we need a new service that has access to these endpoints, we'll call it simple_onchain_discovery and place it under libra/network/simple-onchain-discovery. The requirements are that it can either identify as a fullnode or a validator role. In the validator role, it acquires the changes for the validator nodes and pushes those to connectivity manager. In the fullnode case, it passes the fullnode updates. To instantiate this, we need a connection manager sender, which can be found [here](https://github.com/libra/libra/blob/master/network/src/validator_network/network_builder.rs#L106). We should export this via a new public interface on NetworkBuilder `pub fn connection_request_sender(&self) -> libra_channel::Sender<PeerId, ConnectionRequest>`. Or we can just use the existing [add_protocol_handler](https://github.com/libra/libra/blob/master/network/src/validator_network/network_builder.rs#L309) with empty fields.

Then we need to figure out when to pass updates, we can follow consensus [model](https://github.com/libra/libra/blob/master/consensus/src/epoch_manager.rs#L499) with the caveat that we can just set this up in an infinite loops on the await without needing a select because this has no other utility.

Note: This is a precursor to a larger interaction with Onchain discovery found in network/onchain-discovery. It is not intended to replace it, but to be an incremental improvement that we will integrate at a later time. My recommendation is to not investigate that code for the time being.

[...]

PR #2104, opened 2019-12-20
[consensus] fix safety rules storage issues

davidiw:
- Make all writes sync to ensure we don't leave partial or incomplete
data (TBD, needs cluster testing and announcement before enabling)
- swap between writing to files so that we always have one that is
correct

Addresses #2092

PR #3956, opened 2020-05-19
review comments

bmaurer:
@zekun000 — sending as a PR per your request

### DIFF
--- a/consensus/safety-rules/src/persistent_safety_storage.rs
+++ b/consensus/safety-rules/src/persistent_safety_storage.rs
@@ -18,6 +18,8 @@ pub struct PersistentSafetyStorage {
     internal_store: Box<dyn Storage>,
 }
 
+// @REVIEW: it seems that updates are made on a per key basis. What happens if the process crashes
+// mid update? Is there a need for atomic updates?
 impl PersistentSafetyStorage {
     pub fn in_memory(private_key: Ed25519PrivateKey) -> Self {
         let storage = InMemoryStorage::new_storage();
--- a/consensus/safety-rules/src/safety_rules.rs
+++ b/consensus/safety-rules/src/safety_rules.rs
@@ -158,6 +158,8 @@ impl<T: Payload> TSafetyRules<T> for SafetyRules<T> {
         }
     }
 
+    // @REVIEW Which of these TODOs are necessary for safety? Are all of the checks in this method
+    // strictly necessary for safety?
     /// @TODO verify signature on vote proposal
     /// @TODO verify QC correctness
     /// @TODO verify epoch on vote proposal
@@ -218,6 +220,7 @@ impl<T: Payload> TSafetyRules<T> for SafetyRules<T> {
         ))
     }
 
+    // @REVIEW: does this only impact liveness?
     /// @TODO only sign blocks that are later than last_voted_round and match the current epoch
     /// @TODO verify QC correctness
     /// @TODO verify QC matches preferred round
@@ -229,6 +232,8 @@ impl<T: Payload> TSafetyRules<T> for SafetyRules<T> {
             &self.validator_signer,
         ))
     }
+
+    // @REVIEW: does this only impact liveness?
 
     /// Only sign the timeout if it is greater than or equal to the last_voted_round and ahead of
     /// the preferred_round. We may end up signing timeouts for rounds without first signing votes
--- a/consensus/src/block_storage/block_store.rs
+++ b/consensus/src/block_storage/block_store.rs
@@ -84,6 +84,8 @@ fn update_counters_for_committed_blocks<T>(blocks_to_commit: &[Arc<ExecutedBlock
 ///             ├--------> C2
 ///             ╰--------------> D3
 pub struct BlockStore<T> {
+    // @REVIEW: is there actually concurrent access to this data structure? This is one of the few
+    // places that has concurrency in consensus?
     inner: Arc<RwLock<BlockTree<T>>>,
     state_computer: Arc<dyn StateComputer<Payload = T>>,
     /// The persistent storage backing up the in-memory data structure, every write should go
@@ -333,6 +335,8 @@ impl<T: Payload> BlockStore<T> {
             Some(executed_block) => {
                 ensure!(
                     executed_block.block_info() == *qc.certified_block(),
+                    // @REVIEW: As noted elsewhere -- this check doesn't seem to happen at restart
+                    // because we don't record the blockinfo that we vote on.
                     "QC for block {} has different {:?} than local {:?}",
                     qc.certified_block().id(),
                     qc.certified_block(),
@@ -342,6 +346,8 @@ impl<T: Payload> BlockStore<T> {
             None => bail!("Insert {} without having the block in store first", qc),
         }
 
+        // @REVIEW: is it actually necessary to save a QC that's not attached to a block on disk?
+        // i.e., is liveness threatened if this data is lost?
         self.storage
             .save_tree(vec![], vec![qc.clone()])
             .context("Insert block failed when saving quorum")?;
@@ -355,6 +361,8 @@ impl<T: Payload> BlockStore<T> {
         if tc.round() <= cur_tc_round {
             return Ok(());
         }
+        // @REVIEW: is liveness threatened if this data isn't fsync(). What would the consequence
+        // be if this data wasn't saved?
         self.storage
             .save_highest_timeout_cert(tc.as_ref().clone())
             .context("Timeout certificate insert failed when persisting to DB")?;
--- a/consensus/src/block_storage/block_tree.rs
+++ b/consensus/src/block_storage/block_tree.rs
@@ -225,6 +225,9 @@ where
         // qc1 == qc2 || qc1.round != qc2.round
         // The invariant is quadratic but can be maintained in linear time by the check
         // below.
+
+        // @REVIEW: If this is true, is there a reason this data structure isn't indexed by round?
+        // Also, What if this invariant is violated (e.g., due to byzantine cases? what would happen?)
         precondition!({
             let qc_round = qc.certified_block().round();
             self.id_to_quorum_cert.values().all(|x| {
@@ -254,6 +257,10 @@ where
 
         Ok(())
     }
+
+    // @REVIEW: Would it be simpler just to prune all blocks that have round < committed round?
+    // This seems like a lot of fairly complex code to optimize for purning slightly earlier when
+    // there are forks
 
     /// Find the blocks to prune up to next_root_id (keep next_root_id's block). Any branches not
     /// part of the next_root_id's tree should be removed as well.
@@ -298,6 +305,7 @@ where
     /// Note that we do not necessarily remove the pruned blocks: they're kept in a separate buffer
     /// for some time in order to enable other peers to retrieve the blocks even after they've
     /// been committed.
+    // REVIEW: How long do we need to persist them? Is this for correctness or an optimization?
     pub(super) fn process_pruned_blocks(
         &mut self,
         root_id: HashValue,
--- a/consensus/src/block_storage/sync_manager.rs
+++ b/consensus/src/block_storage/sync_manager.rs
@@ -189,6 +189,8 @@ impl<T: Payload> BlockStore<T> {
         Ok(())
     }
 
+
+    // @REVIEW: It feels like some comments would help here.
     pub async fn fast_forward_sync<'a>(
         highest_commit_cert: &'a QuorumCert,
         retriever: &'a mut BlockRetriever<T>,
--- a/consensus/src/consensus_provider.rs
+++ b/consensus/src/consensus_provider.rs
@@ -39,6 +39,8 @@ pub fn start_consensus(
         .enable_all()
         .build()
         .expect("Failed to create Tokio runtime!");
+    // REVIEW: Should MempoolProxy etc be the object passed to this function (vs passing a channel
+    // to send messages on).
     let storage = Arc::new(StorageWriteProxy::new(node_config, libra_db));
     let txn_manager = Box::new(MempoolProxy::new(consensus_to_mempool_sender));
     let execution_correctness_manager = ExecutionCorrectnessManager::new(node_config);
@@ -48,6 +50,8 @@ pub fn start_consensus(
     ));
     let time_service = Arc::new(ClockTimeService::new(runtime.handle().clone()));
 
+
+    // REVIEW: what bugs could arise if we hit the backpressure of 1024 elements? Is a deadlock possible?
     let (timeout_sender, timeout_receiver) =
         channel::new(1_024, &counters::PENDING_PACEMAKER_TIMEOUTS);
     let (self_sender, self_receiver) = channel::new(1_024, &counters::PENDING_SELF_MESSAGES);
--- a/consensus/src/consensusdb/mod.rs
+++ b/consensus/src/consensusdb/mod.rs
@@ -55,6 +55,9 @@ impl ConsensusDB {
     )> {
         let last_vote = self.get_last_vote()?;
         let highest_timeout_certificate = self.get_highest_timeout_certificate()?;
+        // @REVIEW: seems that we don't record the executed state ID of the block in storage. So
+        // in theory, a block without a QC could be executed with state hash X, then after restart
+        // (due to an error) we recompute it with state Y.
         let consensus_blocks = self
             .get_blocks()?
             .into_iter()
--- a/consensus/src/epoch_manager.rs
+++ b/consensus/src/epoch_manager.rs
@@ -67,6 +67,10 @@ impl<T: Payload> LivenessStorageData<T> {
     }
 }
 
+// @REVIEW: FWIW, before I read any code I assumed that an "EventProcessor" is global and that it
+// delegates to an EpochManager that is spawned per epoch. If other people are confused could be
+// worth considering the name
+
 // Manager the components that shared across epoch and spawn per-epoch EventProcessor with
 // epoch-specific input.
 pub struct EpochManager<T> {
@@ -151,6 +155,7 @@ impl<T: Payload> EpochManager<T> {
             .verifier
             .get_ordered_account_addresses_iter()
             .collect::<Vec<_>>();
+        // @REVIEW: Do we use all these?
         match self.config.proposer_type {
             ConsensusProposerType::MultipleOrderedProposers => {
                 Box::new(MultiProposer::new(epoch_info.epoch, proposers, 2))
@@ -206,6 +211,7 @@ impl<T: Payload> EpochManager<T> {
         };
     }
 
+    // @REVIEW: Different epochs for what? Method name could use more context
     async fn process_different_epoch(&mut self, different_epoch: u64, peer_id: AccountAddress) {
         match different_epoch.cmp(&self.epoch()) {
             // We try to help nodes that have lower epoch than us
@@ -286,6 +292,9 @@ impl<T: Payload> EpochManager<T> {
 
         info!("Update SafetyRules");
 
+        // @REVIEW: the "safety rules" system also has data that's required for liveness. Seems like it'd be nice
+        // to be able to recover just based on the consensus db / storage systems (at least if an operator chooses)
+
         let mut safety_rules = self.safety_rules_manager.client();
         let consensus_state = safety_rules
             .consensus_state()
@@ -342,6 +351,8 @@ impl<T: Payload> EpochManager<T> {
         info!("EventProcessor started");
     }
 
+    // @REVIEW: Under what circumstances would consensusdb not have enough data?
+
     // Depending on what data we can extract from consensusdb, we may or may not have an
     // event processor at startup. If we need to sync up with peers for blocks to construct
     // a valid block store, which is required to construct an event processor, we will take
@@ -351,6 +362,9 @@ impl<T: Payload> EpochManager<T> {
         ledger_recovery_data: LedgerRecoveryData,
         epoch_info: EpochInfo,
     ) {
+        // @REVIEW: might it be necessary to update the network config as sync happens?
+        // What if we are so far out of date that the network info is invalid? Is there a way
+        // to bootstrap?
         let network_sender = NetworkSender::new(
             self.author,
             self.network_sender.clone(),
@@ -377,6 +391,8 @@ impl<T: Payload> EpochManager<T> {
         };
         let validator_keys = validator_set.payload().to_vec();
         info!("Update Network about new validators");
+        // @REVIEW: Why does this use a different method to subscribe to onchain config changes than
+        // e.g., mempool (the reconfig_events system)
         self.network_sender
             .update_eligible_nodes(validator_keys)
             .await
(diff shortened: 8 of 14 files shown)


consensus/src/block_storage/block_tree.rs
```diff
@@ -255,6 +258,10 @@ where
         Ok(())
     }
 
+    // @REVIEW: Would it be simpler just to prune all blocks that have round < committed round?
+    // This seems like a lot of fairly complex code to optimize for purning slightly earlier when
+    // there are forks
```
zekun000, line 263: I think we were trying to make sure block tree is really a tree, it's one of the few chances we can write BFS lol

consensus/src/block_storage/block_tree.rs
```diff
@@ -298,6 +305,7 @@ where
     /// Note that we do not necessarily remove the pruned blocks: they're kept in a separate buffer
     /// for some time in order to enable other peers to retrieve the blocks even after they've
     /// been committed.
+    // REVIEW: How long do we need to persist them? Is this for correctness or an optimization?
```
zekun000, line 308: it's optimization to help nodes fall behind so that it won't stay behind forever like need to fetch block -> already committed -> sync -> need to fetch block ->

consensus/src/block_storage/sync_manager.rs
```diff
@@ -189,6 +189,8 @@ impl<T: Payload> BlockStore<T> {
         Ok(())
     }
 
+
+    // @REVIEW: It feels like some comments would help here.
```
zekun000, line 193: agree, it's extracted from `sync_to_highest_commit_cert`

consensus/src/consensus_provider.rs
```diff
@@ -39,6 +39,8 @@ pub fn start_consensus(
         .enable_all()
         .build()
         .expect("Failed to create Tokio runtime!");
+    // REVIEW: Should MempoolProxy etc be the object passed to this function (vs passing a channel
```
zekun000, line 42: we're writing rust, use channel! - I think mempool is running in a different runtime not like libradb which is shared via Arc

consensus/src/consensus_provider.rs
```diff
@@ -48,6 +50,8 @@ pub fn start_consensus(
     ));
     let time_service = Arc::new(ClockTimeService::new(runtime.handle().clone()));
 
+
+    // REVIEW: what bugs could arise if we hit the backpressure of 1024 elements? Is a deadlock possible?
```
zekun000, line 54: I don't think deadlock is possible, the sender is on the different thread dedicated to send. although I think it's good to have a LIFO queue with size 1 for this.

consensus/src/consensusdb/mod.rs
```diff
@@ -55,6 +55,9 @@ impl ConsensusDB {
     )> {
         let last_vote = self.get_last_vote()?;
         let highest_timeout_certificate = self.get_highest_timeout_certificate()?;
+        // @REVIEW: seems that we don't record the executed state ID of the block in storage. So
+        // in theory, a block without a QC could be executed with state hash X, then after restart
+        // (due to an error) we recompute it with state Y.
```
zekun000, line 60: it's intentional, if it has a QC, recompute would ensure it's the same otherwise panic, if it doesn't have QC, we'll never equivocate because of TCB.
I was hoping this helps in case we see a QC which doesn't match local compute (due to an error) and we panic, after restart recompute can fix the error and recover.

consensus/src/epoch_manager.rs
```diff
@@ -67,6 +67,10 @@ impl<T: Payload> LivenessStorageData<T> {
     }
 }
 
+// @REVIEW: FWIW, before I read any code I assumed that an "EventProcessor" is global and that it
+// delegates to an EpochManager that is spawned per epoch. If other people are confused could be
+// worth considering the name
```
zekun000, line 72: lol, EpochManger is global and spawns EventProcessor per epoch, suggestions are welcome!

consensus/src/epoch_manager.rs
```diff
@@ -151,6 +155,7 @@ impl<T: Payload> EpochManager<T> {
             .verifier
             .get_ordered_account_addresses_iter()
             .collect::<Vec<_>>();
+        // @REVIEW: Do we use all these?
```
zekun000, line 158: LeaderReputation is used in prod, others are test-only or unused I think

consensus/src/epoch_manager.rs
```diff
@@ -206,6 +211,7 @@ impl<T: Payload> EpochManager<T> {
         };
     }
 
+    // @REVIEW: Different epochs for what? Method name could use more context
```
zekun000, line 214: process messages from different epoch, trying to minimize the words

consensus/src/epoch_manager.rs
```diff
@@ -286,6 +292,9 @@ impl<T: Payload> EpochManager<T> {
 
         info!("Update SafetyRules");
 
+        // @REVIEW: the "safety rules" system also has data that's required for liveness. Seems like it'd be nice
+        // to be able to recover just based on the consensus db / storage systems (at least if an operator chooses)
```
zekun000, line 296: how to manage the fault tolerance between tcb and consensus is not clear yet. although it's probably not something like this (we don't do nice-to-have thing right now)

consensus/src/epoch_manager.rs
```diff
@@ -342,6 +351,8 @@ impl<T: Payload> EpochManager<T> {
         info!("EventProcessor started");
     }
 
+    // @REVIEW: Under what circumstances would consensusdb not have enough data?
```
zekun000, line 354: someone deletes consensusdb or fullnode promoted to validator

consensus/src/epoch_manager.rs
```diff
@@ -351,6 +362,9 @@ impl<T: Payload> EpochManager<T> {
         ledger_recovery_data: LedgerRecoveryData,
         epoch_info: EpochInfo,
     ) {
+        // @REVIEW: might it be necessary to update the network config as sync happens?
+        // What if we are so far out of date that the network info is invalid? Is there a way
+        // to bootstrap?
```
zekun000, line 367: I believe the solution for V1 is to restore backup to a state that's relatively new (have at least one peer)

consensus/src/epoch_manager.rs
```diff
@@ -377,6 +391,8 @@ impl<T: Payload> EpochManager<T> {
         };
         let validator_keys = validator_set.payload().to_vec();
         info!("Update Network about new validators");
+        // @REVIEW: Why does this use a different method to subscribe to onchain config changes than
+        // e.g., mempool (the reconfig_events system)
```
zekun000, line 395: legacy, https://github.com/libra/libra/issues/3925

consensus/src/epoch_manager.rs
```diff
@@ -495,6 +511,9 @@ impl<T: Payload> EpochManager<T> {
         mut network_receivers: NetworkReceivers<T>,
         mut reconfig_events: libra_channel::Receiver<(), OnChainConfigPayload>,
     ) {
+        // @REVIEW: Seems like you're trying to prioritize these events. Why
+        // only the first time when the processor is started?
```
zekun000, line 515: not prioritize, we need to bootstrap the processor upon start.

consensus/src/event_processor.rs
```diff
@@ -649,6 +651,8 @@ impl<T: Payload> EventProcessor<T> {
         }
         let block = executed_block.block();
 
+        // REVIEW: is this true now that we block during block retrieval?
```
zekun000, line 654: we always block during block retrieval

consensus/src/event_processor.rs
```diff
@@ -722,6 +726,7 @@ impl<T: Payload> EventProcessor<T> {
                 return;
             }
         } else {
+            // REVIEW: why?
```
zekun000, line 729: I asked the same question lol, see https://toroeng.slack.com/archives/CH9HSH0C9/p1584488500068600

consensus/src/event_processor.rs
```diff
@@ -840,6 +845,11 @@ impl<T: Payload> EventProcessor<T> {
                 .certified_block()
                 .round(),
         );
+        // @REVIEW: The strategy taken here (where you bootstrap by replaying old events) is very
+        // different than the one taken by EpochManager::start_event_processor (which does a fair
+        // bit of manual reconstruction). I wonder if start_event_processor would benefit from a similar
```
zekun000, line 850: I tried the same strategy (replay all blocks), the only manual reconstruction is the virtual genesis because we never persist it (plus we do some cleanup)

consensus/src/liveness/leader_reputation.rs
```diff
@@ -29,6 +29,9 @@ pub struct LibraDBBackend {
     window: Mutex<Vec<(u64, NewBlockEvent)>>,
 }
 
+// REVIEW: I couldn't understand how this worked from the comments. For example, how do you make sure that
+// for round R, everybody is using the same view (even if they are synced up to different points)
```
zekun000, line 33: I should have filled more comments for this, in short we can't guarantee everybody is using the same view(committed version) for round R.
In such cases, they'll time out and do a n^2 sync and hopefully next round they'd have the same view.

consensus/src/persistent_liveness_storage.rs
```diff
@@ -373,6 +373,7 @@ impl<T: Payload> PersistentLivenessStorage<T> for StorageWriteProxy {
             highest_timeout_certificate,
         ) {
             Ok(mut initial_data) => {
+                // @REVIEW: Why do we have to clean these things up at startup?
```
zekun000, line 376: we don't retry or panic if we fail to prune it normally, so there may be some garbage (dangling blocks) in storage, and we want to make sure it won't grow infinitely

consensus/src/persistent_liveness_storage.rs
```diff
@@ -395,6 +396,8 @@ impl<T: Payload> PersistentLivenessStorage<T> for StorageWriteProxy {
                 LivenessStorageData::RecoveryData(initial_data)
             }
             Err(e) => {
+                // @REVIEW: It seems like on this path you can potentially be byzantine (e.g., forget
+                // a previous vote). Should we allow this to happen without operator intervention?
```
zekun000, line 400: you can't be byzantine because of tcb. this is for the convenience when someone deletes consensusdb. I don't have strong opinion on this, this does add more complexity to the code.

consensus/src/util/time_service.rs
```diff
@@ -10,6 +10,8 @@ use std::{
 use thiserror::Error;
 use tokio::{runtime::Handle, time::delay_for};
 
+// REVIEW: note the existence of a TimeService in secure/time/src/lib.rs
```
zekun000, line 13: they copy us!

secure/storage/src/on_disk.rs
```diff
@@ -71,6 +71,11 @@ impl<T: TimeService> OnDiskStorageInternal<T> {
         let contents = serde_json::to_vec(data)?;
         let mut file = File::create(self.temp_path.path())?;
         file.write_all(&contents)?;
+        // @REVIEW: many filesystems will not guarantee that the contents of the file will be stored
```
zekun000, line 74: I think that's why we have temp_path and file_path swapping

state-synchronizer/src/synchronizer.rs
```diff
@@ -159,6 +159,10 @@ impl StateSyncClient {
                 ))
                 .await?;
 
+            // @REVIEW: Is one second reasonable? What happens if there is an error?
+            // in block_store this seems to cause a panic. Imagine that there was swapping / slow disk
+            // seems like this would cause a panic.
+            // Does this actually need to be synchronous?
```
zekun000, line 165: first time seeing this.
we probably should relax the timeout, reconfig publish depends on this (it probably doesn't matter for correctness although may make debug more difficult).

consensus/safety-rules/src/safety_rules.rs
```diff
@@ -158,6 +158,8 @@ impl<T: Payload> TSafetyRules<T> for SafetyRules<T> {
         }
     }
 
+    // @REVIEW Which of these TODOs are necessary for safety? Are all of the checks in this method
+    // strictly necessary for safety?
```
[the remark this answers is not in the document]
davidiw: strictly for safety? We honestly could strip out most of safety rules and just have a series of counters and the private key.

secure/storage/src/on_disk.rs
bmaurer: That doesn't work... in some cases if you do:
rm a b
echo 1 > a
sync
echo 2 > b
mv b a
# crash

cat a

it's possible that you'll get a zero length file.

consensus/src/persistent_liveness_storage.rs
bmaurer: I see... part of this is that we don't take the approach "remove all blocks prior to round X", so we have to ensure everything gets GC'd.

To be defensive, it could be worth considering something along those lines: e.g., prefix this data with the epoch and blow away all data for an epoch once its over

consensus/safety-rules/src/persistent_safety_storage.rs
```diff
@@ -18,6 +18,8 @@ pub struct PersistentSafetyStorage {
     internal_store: Box<dyn Storage>,
 }
 
+// @REVIEW: it seems that updates are made on a per key basis. What happens if the process crashes
+// mid update? Is there a need for atomic updates?
```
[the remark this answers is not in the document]
davidiw: Not sure atomic save would even be possible since that would need to be a primitive on the storage solutions. We could consider a bulk update (i.e., submitting several http requests together), but that is just a mitigation and not a resolution. The ordering does matter though to ensure safety. For example, if you're going to move into the next epoch, you should first set epoch -> next epoch and then set preferred and last voted to 0. If you crash, you're in a bad state but only until you catch up. It might be worth exploring some means to make this appear atomic, set all fields and then have a single value that points to the current set of values.

At the same time, if it does crash, we can log these operations a priori so that the admin could fix safety rules directly. This might be a good case to leverage a human operator.

consensus/src/util/time_service.rs
davidiw: The hope would be to unify them at some point in time, this one has support for async / tokio :(

secure/storage/src/on_disk.rs
davidiw: This isn't production code -- but this is surprising perhaps adding fsync back in but it kills performance :P

zekun000: I think this was discussed in https://github.com/libra/libra/pull/2104/files so probably the reason we don't do fsync is perf cc: @davidiw

consensus/safety-rules/src/persistent_safety_storage.rs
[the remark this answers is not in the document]
zekun000: one easy way to ensure atomic is to store a serialized ConsensusState in storage instead of separate fields.

consensus/safety-rules/src/safety_rules.rs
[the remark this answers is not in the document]
bmaurer: I think the risk here is that if safety rules becomes "safety rules. and some other checks" that either (1) people we duplicate all other checks outside of safety rules (2) in order to avoid duplication people will minimize the situation in (1) but this risks people misunderstanding what checks are in safety rules.

consensus/safety-rules/src/safety_rules.rs
```diff
@@ -218,6 +220,7 @@ impl<T: Payload> TSafetyRules<T> for SafetyRules<T> {
         ))
     }
 
+    // @REVIEW: does this only impact liveness?
```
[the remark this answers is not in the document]
bmaurer: I guess my question (builds on the comment in the convo above) is: "what checks should go in safety rules". You clearly mention that the key operation must happen. But if something isn't required for safety, when should other checks be done

consensus/src/block_storage/block_store.rs
```diff
@@ -355,6 +361,8 @@ impl<T: Payload> BlockStore<T> {
         if tc.round() <= cur_tc_round {
             return Ok(());
         }
+        // @REVIEW: is liveness threatened if this data isn't fsync(). What would the consequence
+        // be if this data wasn't saved?
```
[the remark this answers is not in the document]
bmaurer: Won't nodes re-broadcast their timeouts, eventually allowing us to reassemble the HTC?

Agree that rocksdb persists data. I'm just trying to understand why each piece of data we store is necessary for liveness.

consensus/src/block_storage/block_tree.rs
```diff
@@ -225,6 +225,9 @@ where
         // qc1 == qc2 || qc1.round != qc2.round
         // The invariant is quadratic but can be maintained in linear time by the check
         // below.
+
+        // @REVIEW: If this is true, is there a reason this data structure isn't indexed by round?
+        // Also, What if this invariant is violated (e.g., due to byzantine cases? what would happen?)
```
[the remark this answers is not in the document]
bmaurer: > if there's more than 1 QC in the same round, safety is violated. what would happen? well this chain is doomed we need writeset and blabla.

My point is is more: is the behavior when we run into this situation "panic and inform the system operator". Have we stored enough data to disk at this point to allow the operator to determine the exact block that caused this issue?

> re indexing, it's indexed by id so that we can easily query by block id, there can be multiple blocks in the same round.

But only one block per round can have a QC formed on it (and a block without a QC can never be committed). So it seems like we really only have to remember one block per round.

consensus/src/block_storage/block_tree.rs
bmaurer: That seems like it was more important when we used parent relationships to check the commit rule (vs when we embedded that info in the QC)...

Trying to save you from having to write BFS :-)

consensus/src/block_storage/block_tree.rs
bmaurer: It might be worth stress testing this by having a config where we are strict about deletion.

This seems like a way we could inadvertently cause a liveness problem (if we end up growing a liveness dependency on having this data / a corner case make us aggressive about deletion)

consensus/src/consensus_provider.rs
bmaurer: I see. With a LifoQueue of 1 you're basically treating this like a condition variable (i.e., you can tolerate one wakeup for multiple signals)

consensus/src/consensusdb/mod.rs
bmaurer: Let's say you get a block B1, you execute it and it has state S1. You vote on that state. Then you crash before a QC is formed on B1.

Later, you get a block B2 that build on B1. You execute B as part of computing B' and get a state S1'. When you vote on B2, you'll implicitly be endorsing S1' even though you voted on S1. Strictly speaking, not an equivocation / you aren't doing multiple votes per round... but not sure if it's the right behavior.

I realize this is a corner case that only happens with non-deterministic execution

consensus/src/epoch_manager.rs
bmaurer: Honestly, I'd reverse the two names. EventProcessor sounds like the main() loop for consensus events, EpochManager sounds like it handles a single epoch. But that's just what's intuitive to me...

Anyways, not a big deal, but just want to point out the confusion a first time reader had

secure/storage/src/on_disk.rs
bmaurer: Ok, if it's non-prod code would just suggest marking as such.

consensus/safety-rules/src/persistent_safety_storage.rs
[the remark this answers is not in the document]
davidiw: Hahaha, that is true, let's explore this especially for the triplet of integers that actually have this dependency.

consensus/safety-rules/src/safety_rules.rs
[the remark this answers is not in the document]
davidiw: Very fair point. We're going to work on a specification soon that we can use as a readme to this crate that that explicitly makes it clear what is for safety and what is for reliability / debugability.

consensus/safety-rules/src/safety_rules.rs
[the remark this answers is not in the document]
davidiw: Yeah, I think if we explore solutions within an HSM, maybe we are computation or space contained and do not want to necessarily perform these operations. The specification will be useful to really hammer home on the minimal set and motivation for a larger set of checks.

secure/storage/src/on_disk.rs
davidiw: It is on line 19/20 but I can add some additional emphasis.

consensus/src/block_storage/block_store.rs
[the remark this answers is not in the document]
zekun000: it'll re-broadcast, but they can also keep crashing before it persists. it's more like optimization that we don't lose known certificate even we're able to re-generate them

consensus/src/block_storage/block_tree.rs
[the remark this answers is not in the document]
zekun000: we only have to remember one block per round eventually, but when there's multiple proposals coming, you have to remember both until one of them gets QC

consensus/src/block_storage/block_tree.rs
zekun000: agree it's worth testing, it's not a liveness problem but rather fairness problem. 
we can't have more than 1/3 nodes fall behind, at that time they'd have chance to sync and get all the blocks.
but when there's only handful of nodes fall behind, they may stay behind until majority slows down

consensus/src/consensus_provider.rs
zekun000: I miss the #filetask feature we have, so we can file task easily and not risk forgetting during the auditing.

consensus/src/consensusdb/mod.rs
zekun000: it's true in the sense of transitive result, but not sure if it violates safety since the node still only votes once per round.

consensus/src/persistent_liveness_storage.rs
zekun000: I feel this is sufficient for now, it things get bad, oncall probably would have to do a manual GC

consensus/src/block_storage/block_tree.rs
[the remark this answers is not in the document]
bmaurer: When does the leader for the next round fetch blocks — after they get 2f+1 votes, or does it fetch every block it sees votes on?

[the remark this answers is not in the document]
zekun000: nope, it only fetches block after QC aggregated. you want to propose we only store the block we voted on and fetch blocks that have QC? I think that's probably doable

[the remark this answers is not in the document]
bmaurer: at least from my read that could substantially simplify assumptions (e.g., you could index by round -> you can use ranged deletes to GC). I don't think you have to do this.... but it seems like an interesting approach to consider

davidiw:
@zekun000 are we still working on the feedback here? or should we close it out?

consensus/safety-rules/src/persistent_safety_storage.rs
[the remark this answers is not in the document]
davidiw: So I recall, I specifically wrote the structure of the code so that any crash would be recoverable into a safe state (except one case I'll mention at the end): https://github.com/libra/libra/blob/master/consensus/safety-rules/src/safety_rules.rs#L251

Now the one case we do not handle is the scenario in which a generated vote is lost. Specifically, prior to signing a vote, we set our last_voted_round, sign the vote, and then return it. If there is a crash in safety rules after storing the value and before sending, or consensus crashes before sending it out, then we could have a liveness issue if there are only 2f + 1 honest nodes and this is considered one of the honest nodes.

davidiw:
Gonna close this out as its 2 months old.

PR #3958, opened 2020-05-19
[move-prover] Revised serialization axioms

shazqadeer:
## Motivation

The current axioms for $LCS_serialize_core seem complicated and expensive.  This PR attempts to simplify them.

Specifically:
- The congruence axiom can soundly assume == on the right side of the implication provided we ensure that the Boogie runtime of the Move Prover satisfies (forall v1, v2 :: IsEqual(v1, v2) ==> v1 == v2). 
- The injectivity axiom can be more efficiently encoded using a quantifier with a single bound variable vs two variables (linear vs potentially quadratic instantiations).

Yes

## Test Plan

Existing tests

### DIFF
--- a/language/move-prover/src/prelude.bpl
+++ b/language/move-prover/src/prelude.bpl
@@ -1134,16 +1134,11 @@ function {:inline} $LCS_serialize($m: Memory, $txn: Transaction, ta: TypeValue, 
 }
 
 function $LCS_serialize_core(v: Value): Value;
-
-// This says that $serialize respects isEquals (substitution property)
-// Without this, Boogie will get false positives where v1, v2 differ at invalid
-// indices.
-axiom (forall v1,v2: Value :: IsEqual(v1, v2) ==> IsEqual($LCS_serialize_core(v1), $LCS_serialize_core(v2)));
-
-
-// This says that serialize is an injection
-axiom (forall v1, v2: Value ::  IsEqual($LCS_serialize_core(v1), $LCS_serialize_core(v2))
-           ==> IsEqual(v1, v2));
+function $LCS_serialize_core_inv(v: Value): Value;
+// Needed only because IsEqual(v1, v2) is weaker than v1 == v2 in case there is a vector nested inside v1 or v2.
+axiom (forall v1, v2: Value :: IsEqual(v1, v2) ==> $LCS_serialize_core(v1) == $LCS_serialize_core(v2));
+// Injectivity
+axiom (forall v: Value :: $LCS_serialize_core_inv($LCS_serialize_core(v)) == v);
 
 // This says that serialize returns a non-empty vec<u8>
 {{#if (eq serialize_bound 0)}}


wrwg:
> * The congruence axiom can soundly assume == on the right side of the implication since there is never any computation on the serialized vector

There is actually computation on it. This is what is happening in the performance problem in sources/regression: the result of multiple serialize calls is concatenated and combined with another byte vector, then a certain length of the result is asserted. AFAICT there is no indexing of this vector, though. Does that make a difference?

shazqadeer:
@wrwg :  My comment about the congruence axiom in the PR description is nonsense; I will revise it.  

Let me first describe a potentially surprising implication of the axioms in my PR.  These axioms allow us to prove (forall v1, v2: Value :: IsEqual(v1, v2) ==> v1 == v2) (one application of each axiom).  This means that if we choose to land my PR, we are committed for consistency reasons to ensure that any Value created dynamically in our Boogie runtime satisfies this axiom.  For instance, we must make sure that vectors outside the valid range contain the default value.

We could then ask why we keep around the IsEqual predicate at all?  We already know from the definition of IsEqual that (forall v1, v2: Value :: v1 == v2 ==> IsEqual(v1, v2)).  The derived fact above allows us to conclude that IsEqual(v1,  v2) is identical to v1 == v2.  My take on this question is that IsEqual is needed simply because Z3 (or CVC4, for that matter) is incomplete, i.e., unable to prove all true facts, for the logic we are using in Move Prover.  The stratified definition of IsEqual is an aid to Z3 to help it prove more useful things.

wrwg:
@shazqadeer Your line of argument makes sense to me. We should be actually fine in having DefaultValue everywhere outside the range of a ValueArray. While debugging the performance problem, I checked all vector code in the prelude for this. It looked good.

DavidLDill:
> ... My take on this question is that IsEqual is needed simply because Z3 (or CVC4, for that matter) is incomplete, i.e., unable to prove all true facts, for the logic we are using in Move Prover. The stratified definition of IsEqual is an aid to Z3 to help it prove more useful things.

Are we sure IsEqual is contributing to completeness?  If so, how?  I think the first version was just intended to compare the valid entries of two ValueArrays without comparing the invalid ones.

I thought IsEqual had an implicit property that if all in-range elements of two ValueArrays were "IsEqual", then the two arrays are "IsEqual".  But, when I look at the formula now, I don't see it.

When I used the array theory, I left IsEqual in as a synonym for == so I didn't have to change code generation.

wrwg:
> > ... My take on this question is that IsEqual is needed simply because Z3 (or CVC4, for that matter) is incomplete, i.e., unable to prove all true facts, for the logic we are using in Move Prover. The stratified definition of IsEqual is an aid to Z3 to help it prove more useful things.
> 
> Are we sure IsEqual is contributing to completeness? If so, how? I think the first version was just intended to compare the valid entries of two ValueArrays without comparing the invalid ones.
> 
> I thought IsEqual had an implicit property that if all in-range elements of two ValueArrays were "IsEqual", then the two arrays are "IsEqual". But, when I look at the formula now, I don't see it.
> 
> When I used the array theory, I left IsEqual in as a synonym for == so I didn't have to change code generation.

Last time I try, lots of things stopped working when using the flag `--use-native-equality`, with or without `--use-array-theory`.

shazqadeer:
> I thought IsEqual had an implicit property that if all in-range elements of two ValueArrays were "IsEqual", then the two arrays are "IsEqual". But, when I look at the formula now, I don't see it.

I had a chat with @DavidLDill offline about the comment above.  We examined the current stratified definition of IsEqual together and came to the conclusion that indeed it has the property described above.

> Are we sure IsEqual is contributing to completeness? If so, how?

@DavidLDill and I discussed the above also.  We concluded that the stratified definition of IsEqual does appear to have an effect but we don't completely understand how it helps the solver.  

Overall, we decided to land the PR.  I will do a few more experiments separately to understand the effect of native equality and native array theory and report back.

PR #3962, opened 2020-05-19
[Move] Add freeze/unfreeze events in assoc account

moeziniaf:
## Motivation

To emit events on freeze and unfreeze invocations

## Test Plan

cargo test event emission in functional test

### DIFF
--- a/language/stdlib/modules/libra_account.move
+++ b/language/stdlib/modules/libra_account.move
@@ -83,9 +83,28 @@ module LibraAccount {
     // A privilege to allow the freezing of accounts.
     struct FreezingPrivilege { }
 
+    // Message for freeze account events
+    struct FreezeAccountEvent {
+        // The address that initiated freeze txn
+        initiator_address: address,
+        // The address that was frozen
+        frozen_address: address,
+    }
+
+
+    // Message for freeze account events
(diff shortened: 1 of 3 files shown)


language/stdlib/modules/libra_account.move
```diff
@@ -82,16 +82,37 @@ module LibraAccount {
...
+        initiator_address: address,
+        // The address that was unfrozen
+        unfrozen_address: address,
+    }
+
     resource struct AccountOperationsCapability {
         tracking_cap: AccountTrack::CallingCapability,
         event_creation_cap: Event::EventHandleGeneratorCreationCapability,
+        freeze_event_handle: Event::EventHandle<FreezeAccountEvent>,
+        unfreeze_event_handle: Event::EventHandle<UnfreezeAccountEvent>,
```
runtian-zhou, line 106: Question for @sblackshear @tzakian, should we have the two handles in the `AccountOperationsCapability` or a separate resource? I can see arguments for both.

tzakian: Good question, and I think either way would be fine. That being said, I personally think having it in this resource is better since this holds all of the singleton resources needed for privileged account operations in one place.

PR #3964, opened 2020-05-19
[fuzzing] removing google oss-fuzz hacks

mimoo:
According to this PR we can now remove these flags: https://github.com/google/oss-fuzz/pull/3840/files

There were present presumably to work with a rust codebase and Google's environment.


sausagee:
@bors-libra r+

LGTM

PR #3966, opened 2020-05-19
[consensus] verify epoch on vote proposal

gregnazario:
## Motivation

Add epoch verification on vote proposals to ensure correctness.  This makes sure that we don't accept vote proposals from an entirely different epoch, and to fail short before we do too much work.

Yes

## Test Plan

Added a test to verify that if the epoch has changed, there is in fact an error returned for it.

## Related PRs

None

### DIFF
--- a/consensus/safety-rules/src/safety_rules.rs
+++ b/consensus/safety-rules/src/safety_rules.rs
@@ -126,6 +126,16 @@ impl<T: Payload> SafetyRules<T> {
 
         Ok(())
     }
+
+    /// This checks the epoch given against storage for consistent verification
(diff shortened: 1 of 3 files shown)


consensus/safety-rules/src/safety_rules.rs
```diff
@@ -126,6 +126,16 @@ impl<T: Payload> SafetyRules<T> {
 
         Ok(())
     }
+
+    /// This checks the epoch given against storage for consistent verification
```
davidiw, line 130: oh, you and @JoshLind will get along swimmingly :)

PR #3967, opened 2020-05-19
[ci] Build move-vm-types for powerpc

metajack:
This was added as a dependency for custody so we need to make sure it
continues to build on powerpc.


thefallentree:
@n4ss FYI

n4ss:
<3

PR #3972, opened 2020-05-19
[client] Add multi-currency support to the client and faucet

tzakian:
This PR Adds multi currency support to the faucet and the cli interface.


thefallentree:
per offline convesation, we need additional change
1) add an default of LBR to save trouble of migrating all call sites
2) change the faucet amount to  currency base amount instead of "libra"  (but this would make 1) irrelevant since people need to change callsite anyway)

PR #3906, opened 2020-05-15
[testing] correct features to allow tests to pass in secure/key-manager crate

rexhoffman:
## Motivation

allow code coverage and individual tests to run against the  secure/key-manager crate.

### Have you read the [Contributing Guidelines on pull requests](https://github.com/libra/libra/blob/master/CONTRIBUTING.md#pull-requests)?

Yes

## Test Plan

breaks locally,  fixed, tested locally, wont appear as a broken crate list in slack the next day.
Also removed an unintentional fuzzing featured added in to a regular dependency.

## Related PRs

None.

PR #3973, opened 2020-05-19
Add linters to enforce separation between production and test-only crates

sunshowers:
@rexhoffman found an issue with the fuzzing linter #3906 -- it turned out that we weren't considering that crate because it wasn't included in the default build.

Add a bunch of linters to ensure that:
* we have a set of "production crates", which are exactly the set of crates that form publishable artifacts
* everything else is a test-only crate.


documentation/coding_guidelines.md
```diff
@@ -286,38 +290,70 @@ in tests/benchmarks](https://github.com/rust-lang/cargo/issues/2911), we rely on
...
 2. Annotate your test-only helper `foo` with both the `test` flag (for in-crate callers) and the `"fuzzing"` custom feature (for out-of-crate callers):
-    ```
+    ```rust
     #[cfg(any(test, feature = "fuzzing"))]
     fn foo() { ... }
     ```
 3. (optional) Use `cfg_attr` to make test-only trait derivations conditional:
-    ```
+    ```rust
     #[cfg_attr(any(test, feature = "testing"), derive(FooTrait))]
```
rexhoffman, line 321: should we update "testing" to "fuzzing" here?

devtools/x-core/src/lib.rs
```diff
@@ -1,15 +1,28 @@
 // Copyright (c) The Libra Core Contributors
 // SPDX-License-Identifier: Apache-2.0
 
-use guppy::{graph::PackageGraph, MetadataCommand};
+#[macro_use]
+extern crate rental;
```
bmwill, line 5: Has the rental crate not been updated to 2018 macros?

documentation/coding_guidelines.md
sunshowers: Done!

devtools/x-core/src/lib.rs
sunshowers: Unfortunately it hasn't been, it still requires this.

sunshowers:
TODO before landing: instead of saying that the test-only list isn't canonical, print out a warning explaining what to do.

Cargo.toml
```diff
@@ -127,14 +127,26 @@ members = [
...
+# never include crates that require fuzzing features or test features. These are the crates we want built with no extra
 # test-only code included.
+#
+# For more, see the "Conditional compilation for tests" section in documentation/coding_guidelines.md.
 default-members = [
+    "config/config-builder",
+    "config/management",
+    "client/libra-dev",
+    "execution/db-bootstrapper",
     "testsuite/cli",
```
sunshowers, line 140: TODO before landing: where should this go

PR #3980, opened 2020-05-20
[move-prover] Generalized quantifiers and types-as-values

### LINKED ISSUE #3708: [Feature Request] [move-prover] support type values and variables in spec
# 🚀 Feature Request

To support using type variables and type values in spec

## Motivation

To be able to verify a function like the following, using type variables and type values in the specification is necessary. There seems to be no easy workaround.

```
fun pseudo_create_new_account<Token>(fresh_address: address, coin_to_deposit: Libra::T<Token>)
{
	pseudo_save_account(Balance { coin: Libra::zero<LBR::T>() }, fresh_address);

	deposit(fresh_address, coin_to_deposit);
}
spec fun pseudo_create_new_account {
	…
	aborts_if Token != LBR::T && !exists<Balance<Token>>(fresh_address);
	…
}
```

wrwg:
This PR adds a lot of new fire power (also to shoot in your foot) for specifications with a few extensions:

1. We now support types as values. You can write `type<T>()` to construct a value which represents type T. The only operation available on type values  is equality. The type of types is called `type`, so one can e.g. do
   ```
   define helper(t: type, addr: address): bool { exists<R<t>>(addr) }
   ```
   Notice the use of the type value `t` in the instantiation of `exists`.

2. The quantification over addresses David added recently has beeen generalized to arbitrary types. You can write `all(domain<address>(), |addr| P(addr))` to quantify over addresses, for example. This also works for types using `domain<type>()`.

This was technically relative simple because the boogie encoding already deals with types as values.

The example in the test case `type_values.move` demonstrates that this mechanism can be used to express resource invariants, in the presence of generic types, as module invariants, which -- surprisingly -- seems to work well. Similarly, this should allow to write schemas which can apply to all functions in a module and do not require type parameters (resp. use old-style module invariants directly).

The syntax we have choosen for quantifiers may not be totally adequate any longer, as it was oriented towards bounded quantification over vectors or ranges. We may want to change it in the near future.

## Motivation

Global specifications.

Closes #3708 

Yes

## Test Plan

New Tests

## Related PRs

NA

### DIFF
--- a/language/move-prover/spec-lang/src/ast.rs
+++ b/language/move-prover/spec-lang/src/ast.rs
@@ -281,7 +281,6 @@ pub enum Operation {
     Result(usize),
     Index,
     Slice,
-    Addresses,
 
     // Binary operators
     Range,
@@ -312,6 +311,8 @@ pub enum Operation {
     Len,
     All,
     Any,
+    TypeValue,
+    TypeDomain,
     Global,
     Exists,
     Old,
--- a/language/move-prover/spec-lang/src/translate.rs
+++ b/language/move-prover/spec-lang/src/translate.rs
@@ -322,7 +322,6 @@ impl<'env> Translator<'env> {
         let num_t = &Type::new_prim(PrimitiveType::Num);
         let range_t = &Type::new_prim(PrimitiveType::Range);
         let address_t = &Type::new_prim(PrimitiveType::Address);
-        let addresses_t = &Type::new_prim(PrimitiveType::Addresses);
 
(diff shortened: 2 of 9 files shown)


junkil-park:
`cargo test` failed on `type_values.move`. I think the baseline file needs to be updated.

wrwg:
> `cargo test` failed on `type_values.move`. I think the baseline file needs to be updated.

I forgot to gid-add it. It should be now working.

PR #3984, opened 2020-05-20
[move-modules] Refine work-flow logic for Validator Operator setup

valerini:
This PR makes changes to the stdlib move modules: LibraAccount, LibraSystem and ValidatorConfig.
It introduces a different logic for account creation in LibraAccount that allows to avoid duplicating wrapper function in LibraAccount module. This logic is introduced at the bottom of LibraAccount module to not interfere with existing logic for VASPs and unhosted accounts. Once we agree on a unified approach for managing roles and account creations, we can merge the two logics together.
Note that throughout this, the Validator Operator is "Option::none", making the validator itself be its own validator operator.

The most important changes are made to the core move modules:
* `language/stdlib/modules/libra_account.move`
* `language/stdlib/modules/libra_system.move`
* `language/stdlib/modules/validator_config.move`

In rust most meaningful changes are made to:
* `language/tools/vm-genesis/src/lib.rs`,
* `language/stdlib/src/transaction_scripts.rs`,
* `language/e2e-tests/src/common_transactions.rs` and
* `types/src/validator_config.rs`.

.. and the CLI tool:  `testsuite/cli/src/client_proxy.rs` and `testsuite/tests/libratest/smoke_test.rs` 

The following transaction scripts are being called:
* `language/stdlib/transaction_scripts/create_validator_account.move` - called by the association to create an account for the validator and certify a validator role for this account
* `language/stdlib/transaction_scripts/register_validator.move` - called by the validator to set the consensus config and add itself to the set
* `language/stdlib/transaction_scripts/rotate_consensus_pubkey.move`, `language/stdlib/transaction_scripts/rotate_validator_config.move` - called by validator
* `language/stdlib/transaction_scripts/add_validator.move`, `language/stdlib/transaction_scripts/remove_validator.move` - called by validator add itself to the set

Each Validator has an option to set the ValidatorOperator (otherwise the Validator itself acts as a Validator Operator). The Operator is in charge of setting and rotating the consensus configs and running a consensus node, **adding/removing the validator from the set of active validators**. Note that prior to this PR, this functionality was only available to the Association itself, but now the Association may only grant/revoke a validator role to/from the account (certify/decertify methods of LibraAccount) and it's up to validator's operator to add/remove itself from the set (add_validator/remove_validator methods of LibraSystem).

The Association via functions in LibraAccount module may create an account for a validator (see `LibraAccount::create_validator_account`), it will publish a ValidatorConfig resource under it and establish a certified role (LibraAccount::Role_temp<LibraAccount::ValidatorRole>`) under this account; the association may decertify the role and certify it again indicating whether it wants to enable or disable a validator account (this is done because the resources cannot currently be removed from under another account).

### Registration to Validator Set
A new validator will need to give its auth key to the Association off-chain, the Association will create an account for it [1], then the validator will set the config and add itself to the set [2], those scripts are present:
* Association calls:
[1] `transaction_scripts/create_validator_account.move`
* Validator calls:
[2] `transaction_scripts/register_validator.move`

Alternatively the validator may first rotate-in it's validator operator:
* Association calls:
[1] `LibraAccount::create_validator_account`
* Validator calls:
[2] `ValidatorConfig::rotate_operator`
* Validator Operator calls:
[3] `ValidatorConfig::rotate_config`
[4] `LibraSystem::add_validator`

### Removal from the Validator Set
* The Validator Operator may decide to remove its validator from the set if it wants to pause running consensus, it calls:
`LibraSystem::remove_validator`
* The Association may decide to remove the validator (or validator operator) entirely, it then decertifies the validator (or operator) and calls the reconfiguration (reconfiguration induces the removal of invalid validators or validators with invalid operators):
[1] `LibraAccount::decertify<LibraAccount::ValidatorRole>` (or `LibraAccount::decertify<LibraAccount::ValidatorOperatorRole>`)
[2] `LibraSystem::update_and_reconfigure`
* The Validator may decide to remove itself from the set if it wants to change its validator operator or otherwise temporarily stop its validator node from operating, it first removes its operator and then removes itself from the set:
[1] `ValidatorConfig::remove_operator`
[2] `LibraSystem::remove_validator`

### Rotations
* The Validator rotating its operator should ensure that its old validator continues running the node until the new operator takes over. If there is no certainty on this, the validator should first remove the old operator, remove itself from the validator set, then rotate the operator, finally the operator should update the config and add the validator back to the set:
** Validator calls:
[1] `ValidatorConfig::remove_operator`
[2] `LibraSystem::remove_validator(self)`
[3] `ValidatorConfig::rotate_operator`
** Validator Operator calls:
[4] `ValidatorConfig::rotate_config`
[5] `LibraSystem::add_validator`

* The Validator Operator rotating config calls `ValidatorConfig::rotate_config`. In an emergency it can trigger the reconfiguration via `LibraSystem::update_and_reconfigure` (rate limits should apply - future work), but should not do so if this is a regular rotation and instead should await for the reconfiguration to be invoked automatically or by the Association. Note that update_and_reconfigure will induce a copy of configs from all of the Validator Accounts and may incur very large gas costs.

## Have you read the [Contributing Guidelines on pull requests](https://github.com/libra/libra/blob/master/CONTRIBUTING.md#pull-requests)?

Yes

## Test Plan

cargo xtest

## Related PRs

[PR 3819](https://github.com/libra/libra/pull/3819)


language/stdlib/modules/validator_config.move
```diff
@@ -1,156 +1,183 @@
...
-    // Returns true if addr has a published ValidatorConfig::T resource
-    public fun has(addr: address): bool {
-        exists<T>(addr)
+    ///////////////////////////////////////////////////////////////////////////
+    // Validator setup methods
+    ///////////////////////////////////////////////////////////////////////////
+
+    // This function instantiates a
+    // ValidatorConfig resourse under validator_address
+    public fun initialize(validator_address: address) {
```
davidiw, line 47: since eventually this will be part of the account create process, my recommendation would be to have the rotate_operator try to instantiate this object if it doesn't exist, if it does exist, then it just rotates the operator.

language/stdlib/modules/validator_operator_config.move
```diff
@@ -0,0 +1,41 @@
...
+
+module ValidatorOperatorConfig {
+    use 0x0::Association;
+    use 0x0::Transaction;
+
+    resource struct T {
+        is_certified: bool,
+    }
+
+    public fun initialize(validator_operator_address: address) {
```
davidiw, line 15: let's make this hidden, since it will be part of account creation later -- i.e., create it if it doesn't exist, otherwise move along...

language/stdlib/modules/validator_config.move
```diff
@@ -1,156 +1,183 @@
...
-    // Get Config
-    public fun get_config(addr: address): Config acquires T {
-        *&borrow_global<T>(addr).config
-    }
-
-    // Get consensus_pubkey from Config
-    public fun get_consensus_pubkey(config_ref: &Config): vector<u8> {
-        *&config_ref.consensus_pubkey
+    // This function sets is_certified flag to true
+    public fun certify(validator_address: address) acquires T {
```
davidiw, line 60: I thought the intent here was that the association has already created the account so it doesn't need to certify it. I guess this maps to the current behavior of add_validator?

language/stdlib/modules/libra_system.move
```diff
@@ -44,26 +43,32 @@ module LibraSystem {
...
     fun set_validator_set(value: T) acquires CapabilityHolder {
         LibraConfig::set_with_capability<T>(&borrow_global<CapabilityHolder>(LibraConfig::default_config_address()).cap, value)
     }
 
+    ///////////////////////////////////////////////////////////////////////////
+    // Methods operating the Validator Set config
+    // callable by Validator's operators
+    ///////////////////////////////////////////////////////////////////////////
+
+    // Adds a new validator, this validator should met the validity conditions
```
davidiw, line 57: I guess in the same script we perform the rotation we can call this? can we make this idempotent? so we can keep calling it without causing an assert error? or maybe we should have two scripts rotate and add and just rotate?

language/stdlib/modules/validator_config.move
valerini: Right, the challenge here is that the Association cannot instantiate a certified resource at the moment of account creation at the validator's account (because the `move_to` function is not available in move yet). But eventually, the Association who creates the account will certify it in the same transaction indeed, add_validator now is callable by the Association as a work-around...

language/stdlib/modules/libra_system.move
valerini: Which script do you mean by "the same script we perform the rotation" ?

davidiw: Good question! A transaction script

language/stdlib/modules/validator_operator_config.move
valerini: Where do you think this can be hidden?
In the validator_config we can create resource in the rotate_operator. But who will create resource here - for validator_operator_config?

language/stdlib/modules/libra_system.move
valerini: Good question... Do you mean rotate_config or rotate_operator?

davidiw: depends on whether or not you have a list of potential validators. if you do, we would likely just perform that operation here. If you don't, then after rotate_config and you have a valid config.

types/src/validator_config.rs
```diff
@@ -11,7 +11,8 @@ use serde::{Deserialize, Serialize};
 
 #[derive(Debug, Deserialize, Serialize)]
 pub struct ValidatorConfigResource {
-    pub validator_config: ValidatorConfig,
+    pub is_certified: bool,
+    pub validator_config: Vec<ValidatorConfig>,
     pub delegated_account: Vec<AccountAddress>,
 }
```
davidiw, line 17: Shouldn't these be Option<>? They should have the same representation.

language/stdlib/transaction_scripts/register_validator.move
```diff
@@ -1,35 +1,38 @@
...
+    // Validator Set.
+    fun main(
+        consensus_pubkey: vector<u8>,
+        validator_network_signing_pubkey: vector<u8>,
+        validator_network_identity_pubkey: vector<u8>,
+        validator_network_address: vector<u8>,
+        fullnodes_network_identity_pubkey: vector<u8>,
+        fullnodes_network_address: vector<u8>,
+    ) {
+        ValidatorConfig::rotate_config(Transaction::sender(),
```
davidiw, line 22: Transaction::sender() should be on the next line.

language/stdlib/modules/libra_system.move
```diff
@@ -151,11 +175,20 @@ module LibraSystem {
...
 
-    // Private helper functions
+    ///////////////////////////////////////////////////////////////////////////
+    // Private functions
+    ///////////////////////////////////////////////////////////////////////////
 
-    // The Association, the VM and the validator operator from the current validator set
-    // are authorized to update the set of validator infos
-    fun is_sender_authorized_(): bool {
+    fun is_valid_and_certified(addr: address): bool {
```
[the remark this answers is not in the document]
valerini: Sure! I'd also be happy to discuss this more if you find it should be done in some better way!
But currently:
* "valid" means that the validator has the config set to Option::some (because initially when the association creates an account for the validator, it's config is set to Option::none), so it's up to the validator to make itself valid,
* "certified" means that the association assigned a role to this account (when it calls LibraAccount::certify it sets the is_certified flag of the role to true, when it calls LibraAccount::decertify it sets the is_certified flag of the role to false).

So in order to be added to the ValidatorSet, the validator needs to be certified and valid. Also when LibraSystem::update_and_reconfigure is called, validators who are not valid_and_certified are kicked out of the set.

language/stdlib/modules/validator_config.move
```diff
@@ -1,156 +1,154 @@
...
-    }
+    ///////////////////////////////////////////////////////////////////////////
+    // Validator setup methods
+    ///////////////////////////////////////////////////////////////////////////
 
-    // Get Config
-    public fun get_config(addr: address): Config acquires T {
-        *&borrow_global<T>(addr).config
+    public fun publish(signer: &signer) {
+        Transaction::assert(Transaction::sender() == 0xA550C18, 1101);
```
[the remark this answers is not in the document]
valerini: I think here signer is the validator's account, so the address_of signer is not 0xA550C18, unless I am missing something.

language/stdlib/modules/validator_config.move
```diff
@@ -1,156 +1,154 @@
...
-    // to a delegated_account.
-    public fun set_delegated_account(delegated_account: address) acquires T {
-        Transaction::assert(LibraAccount::exists(delegated_account), 5);
-        // check delegated address is different from transaction's sender
-        Transaction::assert(delegated_account != Transaction::sender(), 6);
-        let t_ref = borrow_global_mut<T>(Transaction::sender());
-        t_ref.delegated_account = Option::some(delegated_account)
+    // Sets a new operator account, preserving the old config.
+    // If the resource does not exist, creates it.
+    public fun rotate_operator(operator_account: address) acquires T {
```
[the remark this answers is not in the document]
valerini: We should! I think this will be my next PR! The challenge here is that most of the tests now assume that "validator account" == "validator_operator account", and it's not possible to instantiate both roles under a single account, so I essentially don't instantiate the validator operator role now anywhere...

language/stdlib/modules/validator_config.move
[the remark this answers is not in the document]
valerini: Oh, may be this function should take two signers?

language/stdlib/modules/libra_account.move
```diff
@@ -863,6 +864,103 @@ module LibraAccount {
...
+                sent_events: Event::new_event_handle<SentPaymentEvent>(&new_account),
+                sequence_number: 0,
+                is_frozen: false,
+            }
+        );
+        // (2) publish Account::Role. it's the caller's job to publish resources with role-specific
+        //     configuration such as AccountLimits before calling this
+        move_to(&new_account, Role_temp<RoleType> { role_type, is_certified: true });
+        // (3) publish Balance resource(s)
+        move_to(&new_account, Balance<Token>{ coin: Libra::zero<Token>() });
```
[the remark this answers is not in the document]
valerini: Makes sense! But how would validators pay transaction fees if they don't have balances?

language/stdlib/modules/validator_config.move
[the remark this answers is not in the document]
valerini: I am probably doing something wrong with the signer - not sure how to pass it in as an argument to the transaction script - tried multiple ways... I'll leave it as is for now, but will be happy to fix it later!!

language/stdlib/modules/libra_system.move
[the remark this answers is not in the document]
sblackshear: Thanks, that makes sense!

language/stdlib/modules/libra_account.move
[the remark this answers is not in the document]
sblackshear: Ah, I think the idea is that validators have the prioritized transactions privilege (like Association accounts). They can always set `gas_price=0`, yet be assured of prompt execution.

language/stdlib/modules/validator_config.move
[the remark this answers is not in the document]
sblackshear: Sorry for the bad experience re signers and scripts! If  you add `&signer`  as the first argument of any script, the VM will take care of creating a `signer` for you + binding a reference to it to the first arg

>  Oh, may be this function should take two signers?

Yes, I think it should probably be something like:

```
public  fun publish(association: &signer, validator: &signer) {
         Transaction::assert(Signer::address_of(association) == 0xA550C18, 1101);
          move_to(validator, T {
            config: Option::none(),
            operator_account: Option::none(),
        });
}
```

language/stdlib/modules/libra_account.move
```diff
@@ -937,6 +938,56 @@ module LibraAccount {
...
+    }
+
+    public fun create_validator_account<Token>(
+        new_account_address: address,
+        auth_key_prefix: vector<u8>,
+    ) {
+        Association::assert_sender_is_association();
+        let new_account = create_signer(new_account_address);
+        Event::publish_generator(&new_account);
+        ValidatorConfig::publish(&new_account);
```
sblackshear, line 983: This will be `ValidatorConfig::publish(association, new_account)`

language/stdlib/modules/libra_account.move
```diff
@@ -937,6 +938,56 @@ module LibraAccount {
...
+        borrow_global_mut<Role_temp<RoleType>>(addr).is_certified = false;
+    }
+
+    public fun certify<RoleType: copyable>(addr: address) acquires Role_temp {
+        Association::assert_sender_is_association();
+        // Role can be only published under the account at its creation
+        borrow_global_mut<Role_temp<RoleType>>(addr).is_certified = true;
+    }
+
+    public fun create_validator_account<Token>(
```
sblackshear, line 976: This will be `create_validator_account<Toke>(association: &signer,  ...)`

language/ir-testsuite/tests/genesis/genesis_validator_set.mvir
```diff
@@ -9,9 +9,5 @@ main() {
     // check that unexpected addresses aren't validators
     assert(!LibraSystem.is_validator(0x88), 80);
 
-    // registering a new *candidate* validator shouldn't change the validator set size
```
davidiw, line 12: why didn't we just delete this test too?

language/move-lang/tests/functional/validator_set/add_one_rotate_by_delegate.move
```diff
@@ -2,16 +2,15 @@
...
 //! account: carrol, 1000000, 0, validator
 
 //! sender: bob
 script {
 use 0x0::ValidatorConfig;
 fun main() {
-    ValidatorConfig::register_candidate_validator(x"beefbeef", x"10", x"20", x"30", x"40", x"50");
     // register alice as bob's delegate
-    ValidatorConfig::set_delegated_account({{alice}});
+    ValidatorConfig::rotate_operator({{alice}});
```
davidiw, line 13: I think just calling this "set_operator" is probably good... or even "set_config"

rotate feels like a phrase that implies that the client is sending only a request with no additional data. not a requirement, just a thought on this topic...

language/move-lang/tests/functional/validator_set/add_validator.move
```diff
@@ -0,0 +1,53 @@
...
+use 0x0::LBR;
+use 0x0::LibraAccount;
+fun main() {
+    LibraAccount::create_validator_account<LBR::T>(0xAA, x"00000000000000000000000000000000");
+}
+}
+
+// check: EXECUTED
+
+// TODO(valerini): enable the following test once the sender format is supported
```
davidiw, line 30: should we fix this now? or is it still work in progress?

language/move-lang/tests/functional/validator_set/register_validator.move
```diff
@@ -0,0 +1,80 @@
...
+    // register Alice as a validator candidate
+    fun main() {
+        LibraAccount::create_validator_account<LBR::T>(0xAA, x"00000000000000000000000000000000");
+    }
+}
+
+// check: EXECUTED
+
+// TODO(valerini): enable the following test once the sender format is supported
+// //! new-transaction
```
davidiw, line 51: same here, also why 0xaa versus a name?

language/stdlib/modules/libra_system.move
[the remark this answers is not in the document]
davidiw: Comment pending?

language/stdlib/modules/validator_config.move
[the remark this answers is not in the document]
davidiw: even if we perform this check, it isn't guaranteed that the operator is still set afterward, so we'll need to check this along with valid every reconfiguration.

testsuite/cli/src/client_proxy.rs
```diff
@@ -514,40 +514,68 @@ impl ClientProxy {
...
         );
         let (account_address, _) =
             self.get_account_address_from_parameter(space_delim_strings[1])?;
-        match self.faucet_account {
-            Some(_) => self.association_transaction_with_local_faucet_account(
-                transaction_builder::encode_remove_validator_script(account_address),
-                is_blocking,
-            ),
-            None => unimplemented!(),
+        let private_key = Ed25519PrivateKey::from_encoded_string(space_delim_strings[2])?;
```
davidiw, line 522: Why are we changing this? It seems like the association should still have this ability? Although arguably -- at some point, we will have to figure out how we balance out the real governance from a test governance (which this is still, right?)

testsuite/cli/src/client_proxy.rs
```diff
@@ -514,40 +514,68 @@ impl ClientProxy {
...
         );
         let (account_address, _) =
             self.get_account_address_from_parameter(space_delim_strings[1])?;
-        match self.faucet_account {
-            Some(_) => self.association_transaction_with_local_faucet_account(
-                transaction_builder::encode_add_validator_script(account_address),
-                is_blocking,
-            ),
-            None => unimplemented!(),
+        let private_key = Ed25519PrivateKey::from_encoded_string(space_delim_strings[2])?;
```
davidiw, line 558: same here

language/stdlib/modules/libra_account.move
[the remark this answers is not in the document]
davidiw: My understanding is these accounts don't pay gas, they bypass these checks.

testsuite/tests/libratest/smoke_test.rs
```diff
@@ -1160,7 +1188,7 @@ fn test_client_waypoints() {
         .unwrap()
         .to_string();
     client_proxy
-        .remove_validator(&["remove_validator", &peer_id], true)
+        .remove_validator(&["remove_validator", &peer_id, &operator_private_str], true)
```
davidiw, line 1191: I think this operation should still be done by the association -- or we should chat briefly about the right format for this.

language/stdlib/modules/validator_config.move
```diff
@@ -1,156 +1,156 @@
...
+    ///////////////////////////////////////////////////////////////////////////
 
-        let t_ref = borrow_global_mut<T>(validator_account);
-        // Set the new key
-        t_ref.config.consensus_pubkey = new_consensus_pubkey;
+    // Returns true if all of the following is true:
+    // 1) there is a ValidatorConfig resource under the address, and
+    // 2) the config is set, and
+    // NB! currently we do not require the the operator_account to be set
+    public fun is_valid(addr: address): bool acquires T {
```
davidiw, line 117: and the validator operator is valid

davidiw:
I think we should try to move this forward a little more aggressively. I'd love to see the APIs not changed if possible. Beyond that, perhaps writing an issue that details what needs to be done to wrap up the MVP solution here and we can close this commit?

language/move-lang/tests/functional/validator_set/add_one_rotate_by_delegate.move
valerini: That's a very good idea! I wanted to rename rotate_ to set_, but forgot...

language/move-lang/tests/functional/validator_set/add_validator.move
valerini: I hope it will get fixed very soon - there is a discussion in slack #dev-platform. But it's still work in progress (by Victor). I can add a temporary function to the module to re-enable this test, but I'd rather wait for it to be fixed... Since especially the validators' creation functionality is being tested with creating validators for these functional tests themselves.

language/move-lang/tests/functional/validator_set/register_validator.move
valerini: The name can't be used because the account for the name already exists... I can't create another account say for {{alice}} address... I need a fresh name to create an address, and fresh names aren't supported for these tests yet.

testsuite/cli/src/client_proxy.rs
valerini: With this PR: the validator operator may add/remove its validator from the set. The Association may certify/decertify a validator. It's not a change in governance - just change in naming.

With that, it was simpler to change these tests to call add/remove validator by the validator's operator (previously - those were called by the Association). Because if I change those tests to call the certify/decertify by the Association, then to become active in the set again after the Association called "certify", the validator's operator needs to additionally call "LibraSystem::add_validator". So the test will become more complicated - but I can change it.

Are you saying, these new APIs are not how we envision governance to work?

language/stdlib/modules/validator_config.move
valerini: Right, I will add the ValidatorOperator role in the next PR (may be by then the role model will get settled on).

language/stdlib/modules/libra_account.move
```diff
@@ -947,6 +948,56 @@ module LibraAccount {
...
+    // Proof of concept code used for Validator and ValidatorOperator roles management
+    ///////////////////////////////////////////////////////////////////////////
+
+    // Defines an alternative role type for the proof-of-concept implementation
+    // that will avoid wrappers around module's functions.
+    resource struct Role_temp<RoleType: copyable> {
+        role_type: RoleType,
+        is_certified: bool, // this flag serves for the purpose of role's revocation
+    }
+    // Role types. This approach effectively mimics the enum,
```
davidiw, line 962: nit new line, maybe fix in next PR :P

language/stdlib/modules/libra_account.move
```diff
@@ -947,6 +948,56 @@ module LibraAccount {
...
+        borrow_global_mut<Role_temp<RoleType>>(addr).is_certified = false;
+    }
+
+    public fun certify<RoleType: copyable>(addr: address) acquires Role_temp {
+        Transaction::assert(Association::addr_is_association(Transaction::sender()), 1002);
+        // Role can be only published under the account at its creation
+        borrow_global_mut<Role_temp<RoleType>>(addr).is_certified = true;
+    }
+
+    public fun create_validator_account<Token>(
```
davidiw, line 986: This is a wonderful function!

language/stdlib/modules/libra_system.move
```diff
@@ -151,11 +175,20 @@ module LibraSystem {
...
+        ValidatorConfig::is_valid(addr) &&
+            LibraAccount::is_certified<LibraAccount::ValidatorRole>(addr)
+            // TODO(valerini): only allow certified operators, i.e. uncomment the line
+            // && LibraAccount::is_certified<LibraAccount::ValidatorOperatorRole>(ValidatorConfig::get_operator(addr))
```
davidiw, line 186: (y)

secure/key-manager/src/libra_interface.rs
```diff
@@ -145,6 +145,7 @@ impl LibraInterface for JsonRpcLibraInterface {
         match validator_config_resource {
             Ok(config_resource) => config_resource
                 .map(|config_resource| config_resource.validator_config)
```
davidiw, line 147: can we use a flat_map here?

PR #3986, opened 2020-05-20
[network] ensure seed peers are fully rendered libranet addrs

phlip9:
+ Follow up PR to #3849.

+ Previously, seed peer addresses were just base transport (e.g.,
`"/ip6/::1/tcp/1234"`). With this diff they are fully rendered and so
include the pubkey `"/ln-noise-ik/<pubkey>"` and handshake protocol
`"/ln-handshake/0"`.

+ All addresses in seed peer configs generated via config builder and
e2e smoke test config generators are now fully rendered libranet
addresses w/ noise pubkey and handshake protocol.

+ Remove hacky seed_peers formatting in `libra-node/src/main_node.rs`.

+ Re-enable transport checks to assert noise pubkeys are present in the
outbound dial addresses.

### DIFF
--- a/config/config-builder/src/full_node_config.rs
+++ b/config/config-builder/src/full_node_config.rs
@@ -247,13 +247,11 @@ impl FullNodeConfig {
     }
 
     fn build_seed_peers(&self, config: &NodeConfig) -> Result<SeedPeersConfig> {
-        let network = config
+        let seed_config = config
             .full_node_networks
             .last()
             .ok_or(Error::MissingFullNodeNetwork)?;
-        let mut seed_peers = HashMap::new();
-        seed_peers.insert(network.peer_id, vec![self.bootstrap.clone()]);
-        Ok(SeedPeersConfig { seed_peers })
+        seed_config.build_seed_peers(self.bootstrap.clone())
     }
 }
 
@@ -275,10 +273,13 @@ mod test {
         // should be the bootstrap
         let config = FullNodeConfig::new().build().unwrap();
         let network = &config.full_node_networks[0];
-        let (seed_peer_id, seed_peer_ips) = network.seed_peers.seed_peers.iter().next().unwrap();
+
+        network.seed_peers.verify_libranet_addrs().unwrap();
+        let (seed_peer_id, seed_addrs) = network.seed_peers.seed_peers.iter().next().unwrap();
+        assert_eq!(seed_addrs.len(), 1);
         assert_ne!(&network.peer_id, seed_peer_id);
-        // This is true because  we didn't update the DEFAULT_ADVERTISED
-        assert_eq!(network.advertised_address, seed_peer_ips[0]);
+        assert_ne!(&network.advertised_address, &seed_addrs[0]);
+
         assert_eq!(
             network.advertised_address,
             NetworkAddress::from_str(DEFAULT_ADVERTISED).unwrap()
@@ -287,6 +288,7 @@ mod test {
             network.listen_address,
             NetworkAddress::from_str(DEFAULT_LISTEN).unwrap()
         );
+
         assert!(config.execution.genesis.is_some());
     }
 
--- a/config/config-builder/src/full_node_config.rs
+++ b/config/config-builder/src/full_node_config.rs
@@ -6,6 +6,7 @@ use anyhow::{ensure, Result};
 use libra_config::{
     config::{
         NetworkPeersConfig, NodeConfig, PeerNetworkId, RoleType, SeedPeersConfig, UpstreamConfig,
+        HANDSHAKE_VERSION,
     },
     utils,
 };
@@ -252,7 +253,6 @@ impl FullNodeConfig {
             .last()
             .ok_or(Error::MissingFullNodeNetwork)?;
 
-        let seed_handshake = 0;
         let seed_pubkey = seed_config
             .network_keypairs
             .as_ref()
@@ -260,7 +260,7 @@ impl FullNodeConfig {
             .identity_keypair
             .public_key();
         let seed_base_addr = self.bootstrap.clone();
-        let seed_addr = seed_base_addr.append_prod_protos(seed_pubkey, seed_handshake);
+        let seed_addr = seed_base_addr.append_prod_protos(seed_pubkey, HANDSHAKE_VERSION);
 
         let mut seed_peers = SeedPeersConfig::default();
         seed_peers
--- a/config/config-builder/src/full_node_config.rs
+++ b/config/config-builder/src/full_node_config.rs
@@ -251,7 +251,22 @@ impl FullNodeConfig {
             .full_node_networks
             .last()
             .ok_or(Error::MissingFullNodeNetwork)?;
-        seed_config.build_seed_peers(self.bootstrap.clone())
+
+        let seed_handshake = 0;
+        let seed_pubkey = seed_config
+            .network_keypairs
+            .as_ref()
+            .ok_or(Error::MissingNetworkKeyPairs)?
+            .identity_keypair
+            .public_key();
+        let seed_base_addr = self.bootstrap.clone();
+        let seed_addr = seed_base_addr.append_prod_protos(seed_pubkey, seed_handshake);
+
+        let mut seed_peers = SeedPeersConfig::default();
+        seed_peers
+            .seed_peers
+            .insert(seed_config.peer_id, vec![seed_addr]);
+        Ok(seed_peers)
     }
 }
 
--- a/config/config-builder/src/full_node_config.rs
+++ b/config/config-builder/src/full_node_config.rs
@@ -4,11 +4,8 @@
 use crate::{BuildSwarm, Error, ValidatorConfig};
 use anyhow::{ensure, Result};
 use libra_config::{
-    config::{
-        NetworkPeersConfig, NodeConfig, PeerNetworkId, RoleType, SeedPeersConfig, UpstreamConfig,
-        HANDSHAKE_VERSION,
-    },
-    utils,
+    config::{NetworkPeersConfig, NodeConfig, PeerNetworkId, RoleType, UpstreamConfig},
+    generator, utils,
 };
 use libra_crypto::ed25519::Ed25519PrivateKey;
 use libra_network_address::NetworkAddress;
@@ -124,8 +121,14 @@ impl FullNodeConfig {
 
     pub fn build(&self) -> Result<NodeConfig> {
         let (mut configs, _) = self.build_internal(false)?;
+
         let validator_config = configs.last().ok_or(Error::NoConfigs)?;
-        let seed_peers = self.build_seed_peers(&validator_config)?;
+        let seed_config = &validator_config
+            .full_node_networks
+            .last()
+            .ok_or(Error::MissingFullNodeNetwork)?;
+        let seed_peers = generator::build_seed_peers(&seed_config, self.bootstrap.clone());
+
         let mut config = configs.swap_remove(self.full_node_index);
         let network = &mut config
             .full_node_networks
@@ -140,8 +143,14 @@ impl FullNodeConfig {
 
     pub fn extend_validator(&self, config: &mut NodeConfig) -> Result<()> {
         let (mut configs, _) = self.build_internal(false)?;
+
         let mut new_net = configs.swap_remove(configs.len() - 1);
-        let seed_peers = self.build_seed_peers(&new_net)?;
+        let seed_config = &new_net
+            .full_node_networks
+            .last()
+            .ok_or(Error::MissingFullNodeNetwork)?;
+        let seed_peers = generator::build_seed_peers(&seed_config, self.bootstrap.clone());
+
         let mut network = new_net.full_node_networks.swap_remove(0);
         network.advertised_address = self.advertised.clone();
         network.listen_address = self.listen.clone();
@@ -228,7 +237,8 @@ impl FullNodeConfig {
             .full_node_networks
             .last()
             .ok_or(Error::MissingFullNodeNetwork)?;
-        let seed_peers = self.build_seed_peers(&validator_full_node_config)?;
+        let seed_peers =
+            generator::build_seed_peers(&validator_full_node_network, self.bootstrap.clone());
         let upstream_peer_id = validator_full_node_network.peer_id;
         for config in configs.iter_mut() {
             let network = config
@@ -245,28 +255,6 @@ impl FullNodeConfig {
         }
 
         Ok((configs, faucet_key))
-    }
-
-    fn build_seed_peers(&self, config: &NodeConfig) -> Result<SeedPeersConfig> {
-        let seed_config = config
-            .full_node_networks
-            .last()
-            .ok_or(Error::MissingFullNodeNetwork)?;
-
-        let seed_pubkey = seed_config
-            .network_keypairs
-            .as_ref()
-            .ok_or(Error::MissingNetworkKeyPairs)?
-            .identity_keypair
-            .public_key();
-        let seed_base_addr = self.bootstrap.clone();
-        let seed_addr = seed_base_addr.append_prod_protos(seed_pubkey, HANDSHAKE_VERSION);
-
-        let mut seed_peers = SeedPeersConfig::default();
-        seed_peers
-            .seed_peers
-            .insert(seed_config.peer_id, vec![seed_addr]);
-        Ok(seed_peers)
     }
 }
 
--- a/config/config-builder/src/validator_config.rs
+++ b/config/config-builder/src/validator_config.rs
@@ -6,8 +6,8 @@ use anyhow::{ensure, format_err, Result};
 use executor::db_bootstrapper;
 use libra_config::{
     config::{
-        ConsensusType, NodeConfig, RemoteService, SafetyRulesService, SecureBackend,
-        SeedPeersConfig, Token, VaultConfig, HANDSHAKE_VERSION,
+        ConsensusType, NodeConfig, RemoteService, SafetyRulesService, SecureBackend, Token,
+        VaultConfig,
     },
     generator,
 };
@@ -141,11 +141,17 @@ impl ValidatorConfig {
     pub fn build(&self) -> Result<NodeConfig> {
         let mut configs = self.build_set()?;
 
-        // Extract and format first node's advertised address for the cluster
-        // bootstrap.
-        let seed_peers = self.build_seed_peers(&configs[0])?;
+        // Extract and format first node's advertised address to use as the seed
+        // peer for bootstrapping other validator nodes.
+        let seed_config = &configs[0]
+            .validator_network
+            .as_ref()
+            .ok_or(Error::MissingValidatorNetwork)?;
+        let seed_peers = generator::build_seed_peers(&seed_config, self.bootstrap.clone());
 
+        // Pull out this specific node from the generated validator configs.
         let mut config = configs.swap_remove(self.index);
+
         let validator_network = config
             .validator_network
             .as_mut()
@@ -276,28 +282,6 @@ impl ValidatorConfig {
         }
 
         Ok(())
-    }
-
-    fn build_seed_peers(&self, config: &NodeConfig) -> Result<SeedPeersConfig> {
-        let seed_config = config
-            .validator_network
-            .as_ref()
-            .ok_or(Error::MissingValidatorNetwork)?;
-
-        let seed_pubkey = seed_config
-            .network_keypairs
-            .as_ref()
-            .ok_or(Error::MissingNetworkKeyPairs)?
-            .identity_keypair
-            .public_key();
-        let seed_base_addr = self.bootstrap.clone();
-        let seed_addr = seed_base_addr.append_prod_protos(seed_pubkey, HANDSHAKE_VERSION);
-
-        let mut seed_peers = SeedPeersConfig::default();
-        seed_peers
-            .seed_peers
-            .insert(seed_config.peer_id, vec![seed_addr]);
-        Ok(seed_peers)
     }
 }
 
--- a/config/config-builder/src/validator_config.rs
+++ b/config/config-builder/src/validator_config.rs
@@ -7,7 +7,7 @@ use executor::db_bootstrapper;
 use libra_config::{
     config::{
         ConsensusType, NodeConfig, RemoteService, SafetyRulesService, SecureBackend,
-        SeedPeersConfig, Token, VaultConfig,
+        SeedPeersConfig, Token, VaultConfig, HANDSHAKE_VERSION,
     },
     generator,
 };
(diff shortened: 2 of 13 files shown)


network/network-address/src/lib.rs
```diff
@@ -321,6 +321,45 @@ impl NetworkAddress {
             .push(Protocol::Handshake(handshake_version))
     }
 
+    /// Check that a `NetworkAddress` looks like a typical LibraNet address with
+    /// associated protocols.
+    pub fn is_libranet_addr(&self) -> bool {
```
mimoo, line 326: is it something that potentially we should call all the time when deserializing a NetowkAddress?

mimoo: potentially having a `deserialize_unchecked_addr()`  function to deserialize without checking if this has a real use case (making the default behavior sane)

config/src/config/network_config.rs
```diff
@@ -191,6 +191,43 @@ impl NetworkConfig {
         };
         self.network_keypairs = Some(network_keypairs);
     }
+
+    /// Use this node config to build a seed peer address (using the given
+    /// `bootstrap_address` instead of the configured `self.advertised_address`).
+    pub fn build_seed_address(&self, bootstrap_address: NetworkAddress) -> Result<NetworkAddress> {
+        // TODO(philiphayes): use enum in wire crate when that gets extracted
+        let handshake_version = 0;
```
zekun000, line 199: we have a few HANDSHAKE_VERSION constant already, why not use one of them?

config/src/config/network_config.rs
```diff
@@ -191,6 +191,43 @@ impl NetworkConfig {
...
+
+        if self.enable_remote_authentication || self.enable_noise {
```
zekun000, line 201: this looks very similar to `append_libranet_protocols`

network/network-address/src/lib.rs
```diff
@@ -321,6 +321,45 @@ impl NetworkAddress {
             .push(Protocol::Handshake(handshake_version))
     }
 
+    /// Check that a `NetworkAddress` looks like a typical LibraNet address with
+    /// associated protocols.
```
zekun000, line 325: some comments about what is a typical address?

network/network-address/src/lib.rs
```diff
@@ -321,6 +321,45 @@ impl NetworkAddress {
...
+        } else {
+            return false;
+        };
+
+        match suffix {
+            [NoiseIK(_), Handshake(_)] | [PeerIdExchange, Handshake(_)] => (),
+            _ => return false,
+        }
+
+        true
```
zekun000, line 360: this functions looks like a regex, does rust support that?

network/network-address/src/lib.rs
phlip9: not all addrs include the later handshake protocols, like listen_address

mimoo: why don't they have the handshake protocol?

also, what about having different `from` functions then? a `from_libra_addr`, and so on.

I guess my comment is more about, how can we make this API harder to mis-use :)

network/src/transport.rs
```diff
@@ -218,12 +218,10 @@ pub fn build_memory_noise_transport(
...
 
     memory_transport
-        .and_then(move |socket, _addr, origin| async move {
+        .and_then(move |socket, addr, origin| async move {
             let (remote_static_key, socket) =
                 noise_config.upgrade_connection(socket, origin).await?;
-
-            // TODO(philiphayes): reenable after seed peers are always fully rendered
-            // expect_noise_pubkey(&addr, remote_static_key.as_slice(), origin)?;
+            expect_noise_pubkey(&addr, remote_static_key.as_slice(), origin)?;
```
mimoo, line 224: btw the name of this function makes me think of `expect` which panics. I'm not suggesting any changes though because the Noise IX -> IK PR will remove this check I believe.

network/network-address/src/lib.rs
```diff
@@ -321,6 +321,45 @@ impl NetworkAddress {
             .push(Protocol::Handshake(handshake_version))
     }
 
+    /// Check that a `NetworkAddress` looks like a typical LibraNet address with
+    /// associated protocols.
+    pub fn is_libranet_addr(&self) -> bool {
+        use Protocol::*;
+
+        let suffix = if self.len() == 3 {
```
mimoo, line 329: comments would help to avoid reverse engineering code :) (what's suffix, prefix)

also, it looks like the NetworkAddr only has 3 components if it's a test. Can we also add a `&& cfg!(test)` ?

network/network-address/src/lib.rs
mimoo: I would suggest renaming this to "canonical". We are defining what a proper libranet address is. (We should probably give a name to other types of address as well.)

network/network-address/src/lib.rs
```diff
@@ -321,6 +321,45 @@ impl NetworkAddress {
...
+            let (prefix, suffix) = self.as_slice().split_at(1);
+
+            match prefix {
+                [Memory(_)] => (),
+                _ => return false,
+            }
+
+            suffix
+        } else if self.len() == 4 {
+            let (prefix, suffix) = self.as_slice().split_at(2);
```
mimoo, line 339: so a canonical libra NetworkAddr looks like: prefix = ip4 or ip6 or dns + tcp and suffix = noiseIK or PeerIdExchange + Handshake

Two comments:

* why not just check that the first element is ip or dns, then that the second element is tcp, then that ... (so checking items one by one instead of two by two)
* looks like PeerIdExchange is also a test only thing right? So a `if cfg!(test)` flow might be better.

network/network-address/src/lib.rs
mimoo: I don't think it would here, it's also easy to mess a regex so I'm more comfortable with this kind of flow (even though it takes more space to write)

config/src/config/network_config.rs
phlip9: there were some problems with circular dependencies when i tried to unify, so i'm going to handle extracting this in a follow up

network/src/transport.rs
phlip9: right, should rename to `verify_noise_pubkey_present` or something. `expect` == panic in my mind to, which is what it was originally to make tests fail faster : p

should be removed soon anyway

network/network-address/src/lib.rs
phlip9: hmm i think you might be right... can probably simplify this with cfg(test)...

libra-node/src/main_node.rs
```diff
@@ -151,8 +149,14 @@ pub fn setup_network(
             config.enable_noise,
             "Permissioned network end-points must use authentication"
         );
-        let seed_peers = config.seed_peers.seed_peers.clone();
+        // Sanity check seed peer addresses.
+        config
+            .seed_peers
```
davidiw, line 154: better in the config space so we detect this earlier and in case it is used else where.

libra-node/src/main_node.rs
```diff
@@ -169,43 +173,18 @@ pub fn setup_network(
             .take_private()
             .expect("identity key should be present");
 
-        // TODO(philiphayes): seed peers should be fully rendered. then we can
-        // also get rid of `network_peers`. we'll need to make some changes to
-        // ops code since it only specifies the base address.
-        let seed_peers: HashMap<_, _> = seed_peers
-            .into_iter()
```
davidiw, line 176: wonderful

network/network-address/src/lib.rs
davidiw: This is a very good point... but what is a libranet address? :/
Perhaps for listen address we just use a SockAddr? There's no real benefit to using a multiaddr there.

phlip9: isn't `SocketAddr` too restrictive? can't listen on `"/memory/<port>"` anymore

config/src/config/network_config.rs
```diff
@@ -191,6 +191,43 @@ impl NetworkConfig {
...
+            Ok(bootstrap_address.append_prod_protos(pubkey, handshake_version))
+        } else {
+            Ok(bootstrap_address.append_test_protos(handshake_version))
+        }
+    }
+
+    /// Use this node config to build a `SeedPeersConfig` with this node as the
+    /// single seed peer (using the given `bootstrap_address` instead of the
+    /// configured `self.advertised_address`).
+    pub fn build_seed_peers(&self, bootstrap_address: NetworkAddress) -> Result<SeedPeersConfig> {
```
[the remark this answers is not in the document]
phlip9: i agree, but it doesn't seem so easy to extract, since `generator` as well as `config-builder`'s `validator_config` and `full_node_config` need this. the `config-builder` does some tricky things with the `bootstrap_address` as the base for the seed peer address

config/config-builder/src/full_node_config.rs
```diff
@@ -247,13 +247,26 @@ impl FullNodeConfig {
...
-        let network = config
+        let seed_config = config
             .full_node_networks
             .last()
             .ok_or(Error::MissingFullNodeNetwork)?;
-        let mut seed_peers = HashMap::new();
-        seed_peers.insert(network.peer_id, vec![self.bootstrap.clone()]);
-        Ok(SeedPeersConfig { seed_peers })
+
+        let seed_handshake = 0;
```
mimoo, line 255: should be a constant : o

config/src/config/network_config.rs
```diff
@@ -200,6 +201,23 @@ pub struct SeedPeersConfig {
     pub seed_peers: HashMap<PeerId, Vec<NetworkAddress>>,
```
mimoo, line 201: why not a tuple struct?

config/config-builder/src/full_node_config.rs
```diff
@@ -247,13 +247,26 @@ impl FullNodeConfig {
...
+            .ok_or(Error::MissingNetworkKeyPairs)?
+            .identity_keypair
+            .public_key();
+        let seed_base_addr = self.bootstrap.clone();
+        let seed_addr = seed_base_addr.append_prod_protos(seed_pubkey, seed_handshake);
+
+        let mut seed_peers = SeedPeersConfig::default();
+        seed_peers
+            .seed_peers
+            .insert(seed_config.peer_id, vec![seed_addr]);
```
mimoo, line 268: so we start with one peer that can have multiple addresses? Why not start with multiple peers?

config/config-builder/src/full_node_config.rs
```diff
@@ -275,10 +288,13 @@ mod test {
         // should be the bootstrap
         let config = FullNodeConfig::new().build().unwrap();
         let network = &config.full_node_networks[0];
-        let (seed_peer_id, seed_peer_ips) = network.seed_peers.seed_peers.iter().next().unwrap();
+
+        network.seed_peers.verify_libranet_addrs().unwrap();
```
mimoo, line 292: shouldn't we `verify_libranet_addrs()` in `build_seed_peers()` too?

config/config-builder/src/validator_config.rs
```diff
@@ -140,24 +140,19 @@ impl ValidatorConfig {
...
         let mut configs = self.build_set()?;
-        let first_peer_id = configs[0]
-            .validator_network
-            .as_ref()
-            .ok_or(Error::MissingValidatorNetwork)?
-            .peer_id;
-        let mut config = configs.swap_remove(self.index);
 
+        // Extract and format first node's advertised address for the cluster
+        // bootstrap.
```
mimoo, line 145: define "cluster bootstrap"

config/config-builder/src/validator_config.rs
```diff
@@ -140,24 +140,19 @@ impl ValidatorConfig {
...
+        let seed_peers = self.build_seed_peers(&configs[0])?;
+
+        let mut config = configs.swap_remove(self.index);
```
mimoo, line 148: I'd suggest adding a comment here, took me a while to figure out that we are removing ourselves

config/src/generator.rs
```diff
@@ -33,12 +33,8 @@ pub fn validator_swarm(
...
     }
 
-    let mut seed_peers = SeedPeersConfig::default();
-    let network = nodes[0].validator_network.as_ref().unwrap();
-    seed_peers
-        .seed_peers
-        .insert(network.peer_id, vec![network.listen_address.clone()]);
-
+    // set the first validator as every validators' initial configured seed peer.
+    let seed_peers = build_seed_peers(&nodes[0]);
```
mimoo, line 37: swarm is only used in tests right?

config/src/generator.rs
```diff
@@ -52,3 +48,22 @@ pub fn validator_swarm_for_testing(nodes: usize) -> ValidatorSwarm {
...
+        .public_key();
+    let seed_base_addr = seed_config.advertised_address.clone();
+    let seed_addr = seed_base_addr.append_prod_protos(seed_pubkey, seed_handshake);
+
+    let mut seed_peers = SeedPeersConfig::default();
+    seed_peers
+        .seed_peers
+        .insert(seed_config.peer_id, vec![seed_addr]);
+    seed_peers
+}
```
mimoo, line 69: there's 3 build_seed_peers functions pretty much doing the same thing, maybe worth it to consolidate?

zekun000: +1 on this, not sure why we need three mostly identical functions

network/network-address/src/lib.rs
```diff
@@ -643,6 +685,124 @@ impl Arbitrary for DnsName {
...
+/// parse canonical libranet protocols
+///
+/// See: `NetworkAddress::is_libranet_addr(&self)`
+fn parse_libranet_protos(protos: &[Protocol]) -> Option<&[Protocol]> {
+    // parse_ip_tcp
+    // <or> parse_dns_tcp
+    // <or> cfg(test) parse_memory
+
+    let opt_transport_suffix = parse_ip_tcp(protos).map(snd);
+    let opt_transport_suffix = opt_transport_suffix.or_else(|| parse_dns_tcp(protos).map(snd));
```
zekun000, line 780:
```
parse_ip_tcp(protos)
  .or_else(|| parse_dns_tcp(protos)
  .or_else(|| if cfg(test parse_memory(protos) else None)
  .map(snd)?;
```

network/network-address/src/lib.rs
```diff
@@ -643,6 +685,124 @@ impl Arbitrary for DnsName {
...
+/// unparsed `&[Protocol]` suffix.
+pub fn parse_handshake(protos: &[Protocol]) -> Option<(u8, &[Protocol])> {
+    match protos.split_first() {
+        Some((Protocol::Handshake(version), suffix)) => Some((*version, suffix)),
+        _ => None,
+    }
+}
+
+/// Extract the second item of a 2-tuple.
+fn snd<T, U>(arg: (T, U)) -> U {
```
zekun000, line 767: the name is confusing, probably more explicit

network/netcore/src/transport/tcp.rs
```diff
@@ -88,18 +87,22 @@ impl Transport for TcpTransport {
...
             },
             actual_addr,
         ))
     }
 
     fn dial(&self, addr: NetworkAddress) -> Result<Self::Outbound, Self::Error> {
-        let socket_addr = multiaddr_to_string(&addr)?;
-        let config = self.clone();
+        let (hostport, _addr_suffix) =
+            parse_hostport(addr.as_slice()).ok_or_else(|| invalid_addr_error(&addr))?;
```
mimoo, line 98: btw, at this point addr is still an untrusted input right?

network/netcore/src/transport/tcp.rs
```diff
@@ -88,18 +87,22 @@ impl Transport for TcpTransport {
...
         let f: Pin<Box<dyn Future<Output = io::Result<TcpStream>> + Send + 'static>> =
-            Box::pin(TcpStream::connect(socket_addr));
-        Ok(TcpOutbound { inner: f, config })
+            Box::pin(TcpStream::connect(hostport));
```
mimoo, line 100: (hostport makes it sound like it's just a port)

network/network-address/src/lib.rs
mimoo, line 767: do we really need a function for that :D?

network/network-address/src/lib.rs
```diff
@@ -643,6 +685,124 @@ impl Arbitrary for DnsName {
...
+    arg.1
+}
+
+/// parse canonical libranet protocols
+///
+/// See: `NetworkAddress::is_libranet_addr(&self)`
+fn parse_libranet_protos(protos: &[Protocol]) -> Option<&[Protocol]> {
+    // parse_ip_tcp
+    // <or> parse_dns_tcp
+    // <or> cfg(test) parse_memory
```
mimoo, line 777: what's <or> ?

mimoo:
LGTM, left some comments, @bors-libra delegate+

config/config-builder/src/full_node_config.rs
phlip9: at least for the internal full node network, there's no reason not to have all peers in your seed config. only issue i think is config generation needs to be modified a bit (it's just taking a single seed peer address vs. all peer addresses). if we want to do this, we should do it in a separate pr

config/src/config/network_config.rs
phlip9: i think it's so the actual seed_peers.toml looks nicer

```toml
# seed_peers.toml

[seed_peers]
# each seed peer info
```

config/config-builder/src/validator_config.rs
phlip9: bootstrapping the validator cluster

config/src/generator.rs
phlip9: right

config/src/generator.rs
phlip9: ... that's what i did previously, putting it in `NetworkConfig`

network/network-address/src/lib.rs
phlip9: ahh you're probably right... stole snd from haskell (http://hackage.haskell.org/package/base-4.14.0.0/docs/Data-Tuple.html) which is not exactly known for its user-friendly naming : p

config/src/generator.rs
zekun000: blame @davidiw who has different taste of code!

network/network-address/src/lib.rs
phlip9: this is about as close as i could get

```rust
    let transport_suffix = None
        .or_else(|| parse_ip_tcp(protos).map(snd))
        .or_else(|| parse_dns_tcp(protos).map(snd))
        .or_else(|| {
            if cfg!(test) {
                parse_memory(protos).map(snd)
            } else {
                None
            }
        })?;
```

not sure which one you prefer

network/network-address/src/lib.rs
phlip9: parser combinator : d

network/netcore/src/transport/tcp.rs
phlip9: yep

config/src/generator.rs
```diff
@@ -52,3 +48,22 @@ pub fn validator_swarm_for_testing(nodes: usize) -> ValidatorSwarm {
     config.test = Some(TestConfig::open_module());
     validator_swarm(&NodeConfig::default(), nodes, [1u8; 32], true)
 }
+
+fn build_seed_peers(config: &NodeConfig) -> SeedPeersConfig {
```
davidiw, line 52: I recommend we just delete this, there is no test cases that need seed peers from generator -- it is purely a feature intended /only/ for swarms.

config/src/generator.rs
davidiw: I think my feedback wasn't super clear :).

We don't need this for generator as mentioned above and the code in fullnode and validator swarm can be unified in config-builder. That would make it not so accessible, which is ideal since it has no practical use for non-test scenarios.

network/netcore/src/transport/tcp.rs
davidiw: sock_addr maybe?

network/network-address/src/lib.rs
```diff
@@ -643,6 +685,124 @@ impl Arbitrary for DnsName {
...
+
+    if protos.len() < 2 {
+        return None;
+    }
+
+    let (prefix, suffix) = protos.split_at(2);
+    match prefix {
+        [Dns(name), Tcp(port)] => Some(((name, *port), suffix)),
+        [Dns4(name), Tcp(port)] => Some(((name, *port), suffix)),
+        [Dns6(name), Tcp(port)] => Some(((name, *port), suffix)),
```
davidiw, line 734: Was looking at where this was used, we just pass in name into tokio. This means tokio might connect to either 4 or 6 but not necessarily what we specified. I recommend we consider leveraging Rust's ToSockAdder and select the right name if 4 or 6 is specified. If it is just Dns, we can delegate parsing to tokio. Thoughts?

network/network-address/src/lib.rs
davidiw: Are you moonlighting as a Haskell engineer?

network/network-address/src/lib.rs
davidiw: to be honest, I don't really care if we compile memory into production, let's not over engineer this.

network/netcore/src/transport/tcp.rs
mimoo: not too sure what to do with this, but a few suggestions:

* add a comment for reviewers to understand that this is untrusted, and to pay attention to the parsing
* change the name of the variable to reflect that it is untrusted
* create an Untrusted type wrapping the correct type

config/src/generator.rs
mimoo: just a general feedback, but reviewing these parts it's not often super clear that they're test only. A combination of no comments indicating this, and no `#[cfg(test)]` annotation. Maybe worth thinking about that, and maybe even moving some of the code to testsuite if it's possible.

config/config-builder/src/validator_config.rs
mimoo: what's the validator cluster :D?

network/network-address/src/lib.rs
zekun000: I see the ugliness comes from the `.map(snd)` because the first element is different for every function

config/src/generator.rs
phlip9: smoke tests appear to rely on `generator` setting the seed peers

network/network-address/src/lib.rs
phlip9: I think we can just use [`tokio::net::lookup_host`](https://docs.rs/tokio/0.2.21/tokio/net/fn.lookup_host.html) directly and filter the results, e.g., `Dns4` will only use `SocketAddrV4`, `Dns6` will only use `SocketAddrV6`, and `Dns` will use either.

PR #3988, opened 2020-05-20
[deps]. cargo update + kube version bump

rexhoffman:
## Motivation

See where it breaks, or keep kube up to date

y

## Test Plan

ci

## Related PRs

none

### DIFF
--- a/testsuite/cluster-test/Cargo.toml
+++ b/testsuite/cluster-test/Cargo.toml
@@ -46,5 +46,5 @@ futures = "0.3.5"
 tokio = { version = "0.2.21", features = ["full"] }
 async-trait = "0.1.31"
 
-kube = "0.32.0"
-k8s-openapi = { version = "0.7.1", default-features = false, features = ["v1_15"] }
+kube = "0.34.0"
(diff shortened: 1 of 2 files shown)


testsuite/cluster-test/Cargo.toml
```diff
@@ -46,5 +46,5 @@ futures = "0.3.5"
 tokio = { version = "0.2.21", features = ["full"] }
 async-trait = "0.1.31"
 
-kube = "0.32.0"
-k8s-openapi = { version = "0.7.1", default-features = false, features = ["v1_15"] }
+kube = "0.34.0"
+k8s-openapi = { version = "0.8.0", default-features = false, features = ["v1_15"] }
```
rexhoffman, line 50: It seems when their is an open k8s-openapi update (which will tend to fail due to the features section here), then kube will be unable to upgrade.

rexhoffman:
@metajack once mio gets upgraded to 0.7.0, the net2 will be replace by stream2, unfortunately latest tokio still depends on mio 0.6.X.

PR #3990, opened 2020-05-20
[backup] asyncify restore cli

msmouse:
## Motivation

in preparation for new Adaptor trait.

yes
## Test Plan
unit tests

### DIFF
--- a/storage/libradb/src/lib.rs
+++ b/storage/libradb/src/lib.rs
@@ -48,7 +48,7 @@ use crate::{
 };
 use anyhow::{ensure, format_err, Result};
 use itertools::{izip, zip_eq};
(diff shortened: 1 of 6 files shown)


storage/libradb/src/lib.rs
```diff
@@ -404,6 +404,14 @@ impl LibraDB {
         Ok(())
     }
 
+    pub fn get_state_restore_receiver(
```
wqfish, line 407: not sure if we should call this ***handler like the back up handler

msmouse: Great idea. Will add that layer when it handles more than state (i.e. Transactions / ledger_infos, etc)

PR #3993, opened 2020-05-20
[move core modules] mint/burn LBR only by minting underlying currencies

sblackshear:
We currently allow minting LBR without adding Coin1/Coin2 assets to the reserve and burning LBR without removing Coin1/Coin2 assets from the reserve. This shouldn't happen.

The underlying problem here was that the `MintCapability<LBR::T>` and `BurnCapability<LBR::T>` were not unique. There was one of each under 0xA550C18 and in the `LBR::Reserve` resource.

This PR fixes the problem by ensuring that the capabilities are unique and placing the unique capability inside the LBR::Reserve resource. However, this means that `Libra::mint<LBR::T>` and `Libra::burn<LBR::T>` no longer work; instead the new `LBR::mint` function must be used.

Ideally, we would not have `LBR::mint` at all. But the problem is that the tests and testnet rely heavily on the ability to mint LBR. LBR is the default currency in every test and lots of tests need to create new accounts and give them currency. In the long run, it would be good to switch the default currency to (e.g.) `Coin1` and get rid of `LBR::mint` altogether. But that is a much bigger task...

## Motivation

Make LBR minting correct.

Y

## Test Plan

Fixed tests that use `Libra::mint<LBR::T>`

## Related PRs

https://github.com/libra/libra/pull/3972


sblackshear:
> Looks good to me! Could you possibly add a transaction script for minting LBR as well with this?

Will do! I need to add functional tests for the new mint + checking that `Libra::mint/burn<LBR>` don't work too.

junkil-park:
Prover (in Miraibot) complains because the test was broken because of the conflict of two PR landed recently. It will be fine once this (https://github.com/libra/libra/pull/4001) is landed.

PR #3996, opened 2020-05-20
[move] Introducing module sliding_nonce

andll:
Sliding nonce is a resource, that can be used in custom scripts to allow for transaction deduplication with limited out-of-order execution.

Sliding keeps fixed size state allowing up to 128 out-of-order transaction to exist at any given time.

For sliding nonce to work effectively, nonce generator for transaction should produce nonce in sequential manner.


tzakian:
If you rebase the coverage reports should work now....

andll:
Rebased

language/move-lang/tests/functional/sliding_nonce/test.move
```diff
@@ -0,0 +1,217 @@
...
+
+// ****
+// Account setup - bob is account with nonce resource and alice is a regular account
+// ****
+
+//! new-transaction
+//! sender: bob
+script {
+    use 0x0::SlidingNonce;
+    // Unset testnet
```
sblackshear, line 12: Does this do something?

language/move-lang/tests/functional/sliding_nonce/test.move
```diff
@@ -0,0 +1,217 @@
...
+}
+// check: EXECUTED
+
+//! new-transaction
+//! sender: alice
+script {
+    use 0x0::SlidingNonce;
+    // Unset testnet
+    fun main() {
+        SlidingNonce::accept_nonce(0)
```
sblackshear, line 40: Duplicate of test above?

andll:
Build failure on this commit is a spurious infra failure, otherwise it's good for review

language/move-lang/tests/functional/sliding_nonce/test.move
```diff
@@ -0,0 +1,217 @@
...
+}
+// check: EXECUTED
+
+//! new-transaction
+//! sender: bob
+script {
+    use 0x0::SlidingNonce;
+    // Unset testnet
+    fun main() {
+        SlidingNonce::accept_nonce(0)
```
sblackshear, line 51: Might be easier to read/maintain if we combine all of the successful `bob` tests into one  (this is what we tend to do elsewhere in the tests):

```
 fun main() {
     SlidingNonce::create_nonce_resource_for_user();
     SlidingNonce::accept_nonce(0);
     SlidingNonce::accept_nonce(0); // duplicates ok
     SlidingNonce.accept_nonce(1);  
     SlidingNonce::accept_nonce(1000 + 127);
     SlidingNonce::accept_nonce(1000);
     SlidingNonce::accept_nonce(2000 + 128);
     SlidingNonce::accept_nonce(3000 + 128);
     SlidingNonce::accept_nonce(3000 + 127)
}
```

sblackshear: Also, one  test with a loop where you try all the values between `N` and 128 might be useful...

language/stdlib/modules/sliding_nonce.move
```diff
@@ -0,0 +1,45 @@
...
+    use 0x0::Transaction;
+
+    resource struct T {
+        min_nonce: u64,
+        nonce_mask: u128,
+    }
+
+    // Aborts transaction if nonce can not be accepted
+    public fun accept_nonce(seq_nonce: u64) acquires T {
+        if(seq_nonce == 0) {
```
sblackshear, line 12: Can you explain the motivation behind this a bit more? It seems like a potential invitation to replays if an approval gets tagged with nonce 0.

sblackshear: Nit: space between `if` and `(` (and same below). Sorry, we don't have an autoformatter yet...

language/stdlib/modules/sliding_nonce.move
```diff
@@ -0,0 +1,45 @@
...
+            return
+        };
+        let t = borrow_global_mut<T>(Transaction::sender());
+        Transaction::assert(t.min_nonce <= seq_nonce, 10001);
+        let jump_limit = 10000; // Don't allow giant leaps in nonce to protect against nonce exhaustion
+        Transaction::assert(t.min_nonce + jump_limit > seq_nonce, 10002);
+        let bit_pos = seq_nonce - t.min_nonce;
+        let nonce_mask_size = 128; // size of T::nonce_mask in bits. no constants in move?
```
sblackshear, line 20: Not yet, but they're coming!

language/stdlib/modules/sliding_nonce.move
```diff
@@ -0,0 +1,45 @@
...
+                t.min_nonce = t.min_nonce + shift;
+            }
+        };
+        let bit_pos = seq_nonce - t.min_nonce;
+        let set = 1u128 << (bit_pos as u8);
+        Transaction::assert(t.nonce_mask & set == 0, 100003);
+        t.nonce_mask = t.nonce_mask | set;
+    }
+
+    public fun create_nonce_resource_for_user() {
```
sblackshear, line 37: - Comment plz
- The idiomatic name for this kind of function is `publish`

language/stdlib/modules/sliding_nonce.move
```diff
@@ -0,0 +1,45 @@
+address 0x0 {
+module SlidingNonce {
+    use 0x0::Transaction;
+
+    resource struct T {
+        min_nonce: u64,
+        nonce_mask: u128,
+    }
+
+    // Aborts transaction if nonce can not be accepted
```
sblackshear, line 10: Could you add a more detailed comment explaining when the nonce will/won't be accepted?

language/stdlib/modules/sliding_nonce.move
```diff
@@ -0,0 +1,45 @@
...
+    public fun accept_nonce(seq_nonce: u64) acquires T {
```
sblackshear, line 11: Nit: `advance_nonce` or some name that suggests that this mutates might be useful.

language/stdlib/modules/sliding_nonce.move
```diff
@@ -0,0 +1,45 @@
+address 0x0 {
+module SlidingNonce {
+    use 0x0::Transaction;
+
+    resource struct T {
```
sblackshear, line 5: Would you mind adding some of the commentary from your excellent doc explaining this scheme as comments?

language/move-lang/tests/functional/sliding_nonce/test.move
andll: So problem here is that every time something fails it aborts the whole thing. How do I test that?
How do I do this check in the same fn main?
```
     SlidingNonce::accept_nonce(2000 + 128); // accepted
     SlidingNonce::accept_nonce(2000); // failed
```

sblackshear: Ah, that you cannot do. I was suggesting combining all of  the success tests into one fn (since most of these are success cases). 

For the failure ones, each needs their own separate fn, but in this case that seems reasonable because the code to build up the state to the failure case is so minimal. But ok with any test organization, just a suggestion!

sblackshear: I guess one more thing that you could do is factor out `accept_nonce(u64): boolean` a, call it from some new fn like `advance_nonce()`, and check check `accept_nonce() == false` instead of failure tests

andll:
@sblackshear Hopefully addressed all of your comments!

andll:
Move prover errors does not seem to be relevant to this pr(something about signer in lines that are not touched in this PR)

language/stdlib/modules/sliding_nonce.move
sblackshear: This is great; thank you!

PR #4000, opened 2020-05-20
[move] Add Signer module and native function

tnowacki:
## Motivation

- Added 0x0::Signer module
- Added native fun borrow_address
- Added user-land fun address_of

## Test Plan

- new tests
- updated old tests


language/ir-testsuite/tests/signer/borrow_address.mvir
```diff
@@ -0,0 +1,11 @@
+//! account: alice, 90000
```
sblackshear, line 1: Nit: `, 9000` not needed

language/ir-testsuite/tests/signer/borrow_address.mvir
```diff
@@ -0,0 +1,11 @@
...
+//! new-transaction
+//! sender: alice
+
+import 0x0.Signer;
+main(s: &signer) {
+    let x: address;
+    x = {{alice}};
+    assert(Signer.borrow_address(move(s)) == &x, 42);
+    return;
+}
```
sblackshear, line 11: Do we need both Move and mvir tests for these? I have found it painful to maintain two copies of the same test when we do this elsewhere...

language/stdlib/modules/signer.move
```diff
@@ -0,0 +1,9 @@
+address 0x0 {
+module Signer {
+    native public fun borrow_address(s: &signer): &address;
```
sblackshear, line 3: Nit: comments

language/ir-testsuite/tests/signer/borrow_address.mvir
tnowacki: I could move them just to Move, sure

PR #4003, opened 2020-05-20
[move prover] Add type requires for generated code

emmazzz:
## Motivation

This PR adds pre-conditions requiring that function parameters are well-formed in generated code. If `type-requires` option is set to true, `requires` is used; otherwise `free requires` is used.

Yes

## Test Plan

cargo test

### DIFF
--- a/language/move-prover/tests/sources/functional/invariants.exp
+++ b/language/move-prover/tests/sources/functional/invariants.exp
@@ -28,9 +28,9 @@ error:  This assertion might not hold.
 
 error:  This assertion might not hold.
 
-     ┌── tests/sources/functional/invariants.move:143:9 ───
+     ┌── tests/sources/functional/invariants.move:150:9 ───
      │
- 143 │         invariant x > 1;
+ 150 │         invariant y > 1;
      │         ^^^^^^^^^^^^^^^^
      │
      =     at tests/sources/functional/invariants.move:153:5: lifetime_invalid_S_branching (entry)
--- a/language/move-prover/src/boogie_helpers.rs
+++ b/language/move-prover/src/boogie_helpers.rs
@@ -259,6 +259,23 @@ pub fn boogie_well_formed_check(
     }
 }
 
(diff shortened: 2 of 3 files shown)


emmazzz:
@wrwg I can't add reviewers so I'm tagging you here. Can you add yourself as a reviewer?

emmazzz:
addressing issue @3989

emmazzz:
@wrwg please help me land it. Thanks!

language/move-prover/tests/sources/functional/invariants.exp
```diff
@@ -28,9 +28,9 @@ error:  This assertion might not hold.
 
 error:  This assertion might not hold.
 
-     ┌── tests/sources/functional/invariants.move:143:9 ───
+     ┌── tests/sources/functional/invariants.move:150:9 ───
      │
- 143 │         invariant x > 1;
+ 150 │         invariant y > 1;
```
wrwg, line 33: Do we understand why this has changed? 

We do have some non-determinism that when multiple conditions can fail, only one of them will be displayed, and which one is kind of non-deterministic. Can both old and new invariant fail here?

If so we may want to rewrite the test a little such that only one condition can fail, to make our tests less flaky. This is particular important as we are moving to make those tests run on CircleCi.

emmazzz: I'm not sure why this has changed. Error should only occur at line 150, and not line 143 though.

emmazzz: I meant I did re-generate the exp file after I saw that MIRAI bot failed here. I don't know why exp file showed it was expected to fail at line 143 before(which was wrong).

wrwg: Then something regressed. Can you please try to investigate? We can't merge without understanding what goes wrong here.

emmazzz: Ok.

PR #4004, opened 2020-05-20
Revert "Bump tokio from 0.2.13 to 0.2.21"

andll:
This reverts commit 968557657521b3ef47c6bab6c7bdedbce9c3ce7d.

Looks like we now have spurious failure in validator startup, reverting first, then investigating


davidiw:
Well this is ultimately disappointing but appreciate you jumping on top of it!

@bors-libra r+

andll:
Some problems with lock file

metajack:
Is that lockfile change really a revert of tokio? It touches so many packages and moves them back to old versions. cc @rexhoffman

PR #4008, opened 2020-05-21
[CI] stop e2e test prebuild

sausagee:
## Motivation
smoketest prebuilds libra-node inside libra-swarm. This breaks the
original assumption that CI can prebuild the test binary and run the
tests in parallel. It is possible to prebuild the libra-node binary,
but it requires saving and restoring the build artifacts in target/
between jobs, which takes longer than building from scratch.

## Test Plan
CI
image

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -234,28 +234,9 @@ jobs:
           name: Build release
           command: RUST_BACKTRACE=1 cargo build -j 8 --release
       - build_teardown
-  build-e2e-test:
-    executor: test-executor
-    description: Generate a list of E2E test targets to be distributed.
-    steps:
-      - build_setup
-      - attach_workspace:
-          at: /home/circleci/project
-      - restore_cargo_package_cache
-      - run:
-          name: Find all e2e tests
-          command: |
-            RUST_BACKTRACE=1 cargo x test --package testsuite -- --list | \
-              grep "::" | sed 's/: .*$//' > e2e_tests
-            cat e2e_tests
-      - persist_to_workspace:
-          root: /home/circleci/project
-          paths:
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -264,16 +245,18 @@ jobs:
       MESSAGE_PAYLOAD_FILE: "/tmp/message_payload"
     steps:
       - build_setup
-      - restore_cargo_package_cache
       - attach_workspace:
           at: /home/circleci/project
+      - restore_cargo_package_cache
```
rexhoffman, line 250: Got it, this is save since the cargo cache is placed in /usr/local/cargo, but this was an unnecessary change.  Puzzled on it for a bit.

.circleci/config.yml
```diff
@@ -264,16 +245,18 @@ jobs:
...
       - run:
           name: Determine test targets for this container.
           # NOTE Currently the tests are distributed by name order. Once test
           # metadata is enabled, the tests can be distributed by run time to
           # speed up this job.
           command: |
-            ls -all target/debug/*
+            RUST_BACKTRACE=1 cargo x test --package testsuite -- --list | \
```
rexhoffman, line 257: Should be safe since it's not executing developer code in generation of the list, and so far our build is deterministic.

.circleci/config.yml
sausagee: This restores the crate cache, not saving it. It is needed to speed up the build by avoiding fetching over network.

The two lines (`attach_workspace blah blah`) right above this are actually not needed anymore, as there is no more prebuilt binary to restore from a prev job.

sausagee: Fixed in #4016

PR #4014, opened 2020-05-21
[management] More progress!

davidiw:
This does a few things:
* Enables GitHub
* Lays out our initial genesis approach -- while work in smart contracts wrap up
* Fixes a bug that wasn't detected because we don't actually verify that a local peer's id maps to what their effective peer id is... perhaps we should remove peer_id from networking and infer from the chain data -- alternatively we should enforce that somewhere...

### DIFF
--- a/config/management/README.md
+++ b/config/management/README.md
@@ -67,3 +67,82 @@ While this is compiled as a single binary it provides several different faciliti
   storage -- leveraging the identity tool.
 * Converting a genesis configuration and a secure storage into a genesis.blob /
   genesis waypoint.
+
+## The Process
+
+* The association first sets up a secure storage, in this case, say GitHub. All
+  participants must have commit and read access to this repository. The
+  association will share out the GitHub repository and a distinct name for each
+  participant.
+* Each participant must retrieve an appropriate
+  [token](https://github.com/settings/tokens) for their account that allows
+  access to the `repo` scope. This token should be stored locally on their disk
+  in a file accessible to the management application.
+* Each participant should initialize their respective key: `association`,
+  `owner`, or `operator` in a secure storage solution, likely Vault. And
+  prepare another token file accessible to the management application.
+* Each participant will upload their key to GitHub specifying their namespace
+  as their distinct name provided earlier:
+  * Association:
+```
+management association-key \\
+    --local 'backend=vault;server=URL;token=PATH_TO_TOKEN' \\
+    --remote 'backend=github;owner=OWNER;repostiory=REPOSITORY;token=PATH_TO_TOKEN;namespace=NAME'
+```
+  * Owner:
+```
+management owner-key \\
+    --local 'backend=vault;server=URL;token=PATH_TO_TOKEN' \\
+    --remote 'backend=github;owner=OWNER;repostiory=REPOSITORY;token=PATH_TO_TOKEN;namespace=NAME'
+```
+  * Operator:
+```
+management association-key \\
+    --local 'backend=vault;server=URL;token=PATH_TO_TOKEN' \\
+    --remote 'backend=github;owner=OWNER;repostiory=REPOSITORY;token=PATH_TO_TOKEN;namespace=NAME'
+```
+* The operator will upload a signed validator-config (note: the owner address
+  is irrelevant in this run):
+```
+management validator-config \\
+    --owner-address 00000000000000000000000000000000 \
+    --validator-address '/dns/DNS/tcp/PORT' \
+    --fullnode-address '/dns/DNS/tcp/PORT' \
+    --local 'backend=vault;server=URL;token=PATH_TO_TOKEN' \
+    --remote 'backend=github;owner=OWNER;repostiory=REPOSITORY;token=PATH_TO_TOKEN;namespace=NAME'
+```
+* The association will publish a layout containing the distinct names and roles
+  of the participants, this is placed into a common namespace:
+```
+management set-layout \\
+  --path PATH_TO_LAYOUT \\
+  --backend 'backend=github;owner=OWNER;repostiory=REPOSITORY;token=PATH_TO_TOKEN;namespace=common'
+```
+* Each participant can now produce genesis, this requires no namespace:
+```
+management genesis \\
+  --path PATH_TO_LAYOUT \\
+  --backend 'backend=github;owner=OWNER;repostiory=REPOSITORY;token=PATH_TO_TOKEN'
+```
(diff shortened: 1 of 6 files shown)


config/management/README.md
```diff
@@ -67,3 +67,74 @@ While this is compiled as a single binary it provides several different faciliti
...
+management association-key \\
+    --local 'backend=vault;server=URL;token=PATH_TO_TOKEN' \\
+    --remote 'backend=github;owner=OWNER;repostiory=REPOSITORY;token=PATH_TO_TOKEN;namespace=NAME'
+```
+* The operator will upload a signed validator-config (note: the owner address
+  is irrelevant in this run):
+```
+management validator-config \\
+    --owner-address 00000000000000000000000000000000 \\
+    --validator-address '\\dns\\DNS\\tcp\\PORT' \\
```
mgorven, line 109: multiaddr uses forward slashes, not back slashes.

config/management/README.md
```diff
@@ -67,3 +67,74 @@ While this is compiled as a single binary it provides several different faciliti
...
+    --fullnode-address '\\dns\\DNS\\tcp\\PORT' \\
+    --local 'backend=vault;server=URL;token=PATH_TO_TOKEN' \\
+    --remote 'backend=github;owner=OWNER;repostiory=REPOSITORY;token=PATH_TO_TOKEN;namespace=NAME'
```
mgorven, line 112: What is the `owner` param? And `namespace`?

config/management/src/validator_config.rs
```diff
@@ -43,7 +43,7 @@ impl ValidatorConfig {
         let consensus_key = ed25519_from_storage(CONSENSUS_KEY, local.as_mut())?;
         let fullnode_network_key = x25519_from_storage(FULLNODE_NETWORK_KEY, local.as_mut())?;
         let validator_network_key = x25519_from_storage(VALIDATOR_NETWORK_KEY, local.as_mut())?;
-        let owner_key = ed25519_from_storage(OWNER_KEY, local.as_mut())?;
+        let operator_key = ed25519_from_storage(OPERATOR_KEY, local.as_mut())?;
```
mgorven, line 46: Does this remove the need for signing permission on the `owner` key? Or are partners going to be using `management owner-key` as well?

config/management/README.md
```diff
@@ -67,3 +67,74 @@ While this is compiled as a single binary it provides several different faciliti
...
+  participants must have commit and read access to this repository. The
+  association will share out the GitHub repository and a distinct name for each
+  participant.
+* Each participant must retrieve an appropriate
+  [token](https://github.com/settings/tokens) for their account that allows
+  access to the `repo` scope. This token should be stored locally on their disk
+  in a file accessible to the management application.
+* Each participant should initialize their respective key: `association`,
+  `owner`, or `operator` in a secure storage solution, likely Vault. And
+  prepare another token file accessible to the manaagement application.
```
JoshLind, line 83: typo: *manaagement*

config/management/README.md
```diff
@@ -67,3 +67,74 @@ While this is compiled as a single binary it provides several different faciliti
...
+```
+* Each participant can now produce genesis, this requires no namespace:
+```
+management genesis \\
+  --path PATH_TO_LAYOUT \\
+  --backend 'backend=github;owner=OWNER;repostiory=REPOSITORY;token=PATH_TO_TOKEN'
+```
+* Each participant can now produce a waypoint, this requires no namespace, also
+  note, in this case, the remote is the destination where the waypoint will be
+  stored derived from data in the local backend:
```
JoshLind, line 129: did you mean "stored and derived"?

config/management/README.md
```diff
@@ -67,3 +67,74 @@ While this is compiled as a single binary it provides several different faciliti
...
+```
+management create-waypoint \\
+    --local 'backend=github;owner=OWNER;repostiory=REPOSITORY;token=PATH_TO_TOKEN' \\
+    --remote 'backend=vault;server=URL;token=PATH_TO_TOKEN'
+```
+* Perform a verify that ensure the local store maps to Genesis and Genesis maps
```
JoshLind, line 135: typo: "ensure"?

config/management/README.md
```diff
@@ -67,3 +67,74 @@ While this is compiled as a single binary it provides several different faciliti
...
+* Each participant must retrieve an appropriate
+  [token](https://github.com/settings/tokens) for their account that allows
+  access to the `repo` scope. This token should be stored locally on their disk
+  in a file accessible to the management application.
+* Each participant should initialize their respective key: `association`,
+  `owner`, or `operator` in a secure storage solution, likely Vault. And
+  prepare another token file accessible to the manaagement application.
+* Each participant will upload their key to GitHub specifying their namespace
+  as their distinct name provided earlier:
+  * Association:
```
JoshLind, line 86: Having now been able to see the end to end flow a little more clearly, I'm wondering if it might make sense to rename some of these operations. For example, renaming the *-key commands to be upload-*-key instead (or submit-*key), as they simply just upload the key to the remote storage (e.g., Github account). Likewise for "validator-config". Regarding the genesis command, it might also make sense to update it to something like "produce-genesis"? My feeling is that we probably want these commands to be as obvious as possible -- especially if they're going to be run by non-experts. (Just my 2 cents!)

davidiw: so many more characters :octocat:

config/management/README.md
davidiw: ah good call. it is also worth calling out that my command line-based markdown displays differently than github... I cannot say one is better than the other... but text-based markdown is pretty cool

config/management/README.md
davidiw: owner would be the owner or project where the repo is held
namespace is the directory within the namespace

I'm actually thinking we should give a templatable script for this process specific to our genesis and shared out of band

config/management/README.md
davidiw: ahh, the language here is bad... but good call, I'll fix

config/management/src/validator_config.rs
davidiw: there are moving pieces, as long as we keep coordinated on the doc, we can seamlessly upgrade from milestone to milestone while minimizing the headache for partners.

I prefer we only simplify and never increase complexity -- we're not ready to simplify yet

config/management/README.md
mgorven: So it would actually be "libra"?

Issue #3663, opened 2020-04-30
[Bug] CLI should be more verbose about errors from the VM


bob-wilson:
# 🐛 Bug

This is a usability bug for the Libra CLI. When there is an error from the VM, it is way too easy to miss it because the only indication that a transaction fails is that no events are emitted. If at all possible, we should find a way to show some error messages.

## To reproduce

Follow the steps in the [Run Move Programs Locally document](https://developers.libra.org/docs/run-move-locally) up through the "Execute Transaction Script" section. Re-run the custom script without specifying any script arguments (or with completely bogus arguments like "abc"). You should see the following output:

```
transaction is stored!
no events emitted
Successfully finished execution
```

The only difference from the successful output is the "no events emitted". It is especially confusing because of the "Successfully finished execution" message.

## Expected Behavior

Whatever error is detected by the VM should be displayed in a more prominent way.

bob-wilson:
See also https://github.com/libra/libra/issues/3658 -- the problem is even worse for failures publishing modules because (AFAICS) there is no way at all to see if the operation was successful. The CLI shows "no events emitted" and "Successfully published module" regardless.

bob-wilson:
This is fixed in #4017

PR #4017, opened 2020-05-21
[client] Detect and report failed transactions

bob-wilson:
The client code has not been checking the VM status to determine if transactions executed successfully. This changes to check for a status of EXECUTED and reports an error otherwise. This exposed an issue where the external_transaction_signer test in smoke_test.rs has been failing because it ran out of gas, so this also adjusts the gas parameters for that test to make it pass.

https://github.com/libra/libra/issues/3663

Yes

## Test Plan

Ran existing tests and fixed the issues that were exposed

### DIFF
--- a/testsuite/cli/src/client_proxy.rs
+++ b/testsuite/cli/src/client_proxy.rs
@@ -623,17 +623,17 @@ impl ClientProxy {
                 .get_txn_by_acc_seq(account, sequence_number - 1, true)
             {
                 Ok(Some(txn_view)) => {
-                    if txn_view.vm_status != StatusCode::EXECUTED {
-                        break Err(format_err!(
-                            "transaction failed to execute; status: {:?}!",
-                            txn_view.vm_status
-                        ));
-                    } else {
-                        println!("transaction is stored!");
+                    if txn_view.vm_status == StatusCode::EXECUTED {
+                        println!("transaction executed!");
                         if txn_view.events.is_empty() {
                             println!("no events emitted");
                         }
                         break Ok(());
+                    } else {
+                        break Err(format_err!(
+                            "transaction failed to execute; status: {:?}!",
+                            txn_view.vm_status
+                        ));
                     }
                 }
                 Err(e) => {
--- a/testsuite/cli/src/client_proxy.rs
+++ b/testsuite/cli/src/client_proxy.rs
@@ -33,6 +33,7 @@ use libra_types::{
         parse_transaction_argument, Module, RawTransaction, Script, SignedTransaction,
         TransactionArgument, TransactionPayload, Version,
     },
+    vm_error::StatusCode,
     waypoint::Waypoint,
 };
 use libra_wallet::{io_utils, WalletLibrary};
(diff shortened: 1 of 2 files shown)


testsuite/cli/src/client_proxy.rs
```diff
@@ -618,11 +623,18 @@ impl ClientProxy {
...
-                    println!("transaction is stored!");
-                    if txn_view.events.is_empty() {
-                        println!("no events emitted");
+                    if txn_view.vm_status != StatusCode::EXECUTED {
+                        break Err(format_err!(
+                            "transaction failed to execute; status: {:?}!",
+                            txn_view.vm_status
+                        ));
+                    } else {
+                        println!("transaction is stored!");
```
thefallentree, line 632: I think you meant transaction is executed.

testsuite/cli/src/client_proxy.rs
```diff
@@ -618,11 +623,18 @@ impl ClientProxy {
                 .get_txn_by_acc_seq(account, sequence_number - 1, true)
```
thefallentree, line 623: I think you need to handle Network related issue here too. Otherwise one network issue and you will be done.

testsuite/cli/src/client_proxy.rs
```diff
@@ -598,13 +599,17 @@ impl ClientProxy {
...
     }
 
     /// Waits for the next transaction for a specific address and prints it
-    pub fn wait_for_transaction(&mut self, account: AccountAddress, sequence_number: u64) {
+    pub fn wait_for_transaction(
+        &mut self,
+        account: AccountAddress,
+        sequence_number: u64,
+    ) -> Result<()> {
         let mut max_iterations = 5000;
```
thefallentree, line 613: Instead of using fixed iteration, could we set an time deadline instead?  because the defualt timeout settings on the JsonRpcCient. this could be super long if we wait for 5000 * (max_timout + 10ms)

bob-wilson: All I'm changing here is to return a Result from wait_for_transaction. I see your point about using a time deadline, but that should be a separate change (preferably done by someone more familiar with the history of this code).

testsuite/cli/src/client_proxy.rs
bob-wilson: @phoenix-antigravity changed it from "completed" to "stored" in c433798c52. I'm just moving this code. But, maybe you have a point -- since we now know that the execution was successful, maybe the message should change as well. Phoenix, WDYT?

testsuite/cli/src/client_proxy.rs
bob-wilson: I'm not sure what you mean. From what I can tell, the intent of this code is to just keep trying until it either completes or times out.

bob-wilson:
> this will, incidentally will really affect faucet performance (because it uses CLI to mint)

I don't see how it will affect performance at all. I'm just adding a check of the VM status and reporting and error when a transaction fails to execute.

thefallentree:
my bad I thought this changed how mint transaction work, but it didn't

testsuite/cli/src/client_proxy.rs
bob-wilson: I'll go ahead and change it to "transaction executed". I'm also going to invert the comparison and put the successful case first in the if-statement.

PR #4019, opened 2020-05-21
[cluster-test] Add basic json rpc health check when waiting for cluster health

andll:
This diff adds additional check to wait_until_all_healthy - in addition to verifying consensus health, we also making sure that jsonrpc port is available on all nodes, including full nodes.
    
Motivation for this change is in tokio upgrade problem.
    
We've seen after tokio upgrade that sometimes validators report healthy, but then tx emitter fails to connect to jsonrpc port and fails cluster test.
    
There can be two explanations to this:
- With new tokio there is some race condition that makes jsonrpc on full node start little bit after consensus is healthy. In this case this PR will fix the problem as we now wait for json rpc to be available
    
- Maybe with new tokio jsonrpc do not star on some nodes - in this case this PR will highlight problematic node and give better visibility.

This diff also reverts the revert and puts tokio back to latest version - we can observe if there are failures over the weekend.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -45,7 +45,7 @@ dependencies = [
  "libra-types 0.1.0",
  "libra-workspace-hack 0.1.0",
  "prost 0.6.1 (registry+https://github.com/rust-lang/crates.io-index)",
- "tokio 0.2.13 (registry+https://github.com/rust-lang/crates.io-index)",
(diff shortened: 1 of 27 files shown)


rexhoffman:
@bors-libra delegate=andll

For the first option, there are two possibilities:  tokio upgrade introduces a new race condition, or makes an existing race condition more likely to fail.   Always fun stuff.

PR #4020, opened 2020-05-21
Grant mint capability to blessed

moeziniaf:
## Motivation

Let blessed have all coin mint capabilities

Oui

## Test Plan

Coming PR to test g


language/stdlib/modules/genesis.move
```diff
@@ -91,6 +91,9 @@ module Genesis {
         Libra::grant_burn_capability_for_sender<Coin1::T>();
         Libra::grant_burn_capability_for_sender<Coin2::T>();
         Libra::grant_burn_capability_for_sender<LBR::T>();
+        Libra::grant_mint_capability<Coin1::T>();
+        Libra::grant_mint_capability<Coin2::T>();
+        Libra::grant_mint_capability<LBR::T>();
```
andll, line 96: @dmitri-perelman / @sblackshear do we want LBR mint capability for treasure/compliance account?

dmitri-perelman: Good call, we don't want to mint LBR in general in prod, so no need to have this capability.

sblackshear: Agreed--the `MintCapability` for LBR should live in `LBR::Reserve` and nowhere else.

runtian-zhou:
Can you run `cargo run` in stdlib so that we can commit this change in the coming testnet.

language/stdlib/modules/genesis.move
andll: We probably should remove burn capability for LBR as well from TC account?
@moeziniaf since you are here, do you mind removing it? (line `grant_burn_capability_for_sender<LBR::T>`)

language/stdlib/modules/libra.move
```diff
@@ -138,6 +138,11 @@ module Libra {
         MintCapability<CoinType> { }
     }
 
+    public fun grant_mint_capability_for_sender<CoinType>() {
+        Transaction::assert(Transaction::sender() == 0xB1E55ED, 0);
```
tzakian, line 142: You probably want an additional assertion that `CoinType` is a registered currency here as well (`assert_is_coin`). I also just realized looking at this, that we also want to add this to `grant_burn_capability_for_sender` if you could add this same check to that as well.

PR #4021, opened 2020-05-21
[language/stdlib] Connect doc generator with stdlib.

wrwg:
This connects the doc generator to the program which generates the stdlib blob and scripts, and improves the generator to better work with a set of documentation files. It also checks in all generated docs.

Usage: run `cargo run -p stdlib` to regenerate all stdlib docs in addition to stdlib blob and staged scripts. Run `cargo run -p stdlib -- --doc-only` to only run documentation generation. Or use `--no-doc` to skip documentation generation.

stdlib:
- updated association.move to work with docgen. It now serves as an example how to work with specs and docgen.

stdlib generator:
- Added hooks for generating documentation for modules and scripts. Documentation is stored in `modules/doc` and transaction_scripts/doc`.
- Added command line flags for controlling operation (see `cargo run -p stdlib -- --help`).
- Script documentation generation is slow specifically for scripts if not compiled in `--release` mode (it's 4 seconds vs 100 seconds).

docgen:
- Generates now one .md for each .move file.
- Cross-reference linking between .md files.
- Can link to already generated .md from different runs.
- Spec fun, vars, and schemas now have links

## Motivation

Better docs.

Yes

## Test Plan

cargo test

## Related PRs

NA


wrwg:
@sblackshear @tzakian would be nice we can submit this before the long weekend. It simplifies live for the folks working on the specs as document generation is automated for them.

sblackshear:
This looks excellent; thank you!

A favor to ask: would you mind not landing this until I can get https://github.com/libra/libra/pull/3819 in? Both touch big portions of the standard library, but many changes here are fairly mechanical (I assume you ran a script to add the extra `/`?) and thus should should be easy to reapply. If I'm wrong about that and you made the changes by hand, feel free to go ahead.

wrwg:
> This looks excellent; thank you!
> 
> A favor to ask: would you mind not landing this until I can get #3819 in? Both touch big portions of the standard library, but many changes here are fairly mechanical (I assume you ran a script to add the extra `/`?) and thus should should be easy to reapply. If I'm wrong about that and you made the changes by hand, feel free to go ahead.

Bummer. No the `///` is manually (its not automatable as it stands -- you can once replace it but then have to fix all code comments back to `//` based on compiler errors) and by now done for all sources.  When do you plan to land this?

PR #4022, opened 2020-05-21
[management] doc update

davidiw:


config/management/README.md
```diff
@@ -70,74 +70,89 @@ While this is compiled as a single binary it provides several different faciliti
...
-  * Operator:
+
+### Validator Operators
+
+* Each Validator Operator member will upload their key to GitHub:
 ```
-management association-key \\
-    --local 'backend=vault;server=URL;token=PATH_TO_TOKEN' \\
+management operator-key \
+    --local 'backend=vault;server=URL;token=PATH_TO_TOKEN' \
```
mgorven, line 122: Replace URL with http://localhost:8200
PATH_TO_TOKEN with VAULT_TOKEN_FILE

config/management/README.md
```diff
@@ -70,74 +70,89 @@ While this is compiled as a single binary it provides several different faciliti
...
     --remote 'backend=github;owner=OWNER;repostiory=REPOSITORY;token=PATH_TO_TOKEN;namespace=NAME'
```
mgorven, line 123: What do they use for OWNER? and REPOSITORY? How do they get their Github TOKEN? Where is namespace from?

config/management/README.md
```diff
@@ -70,74 +70,89 @@ While this is compiled as a single binary it provides several different faciliti
...
-  --backend 'backend=github;owner=OWNER;repostiory=REPOSITORY;token=PATH_TO_TOKEN;namespace=common'
-```
-* Each participant can now produce genesis, this requires no namespace:
+* Upon receiving signal from the association, validator operators can now build
+  genesis, this requires no namespace:
 ```
-management genesis \\
-  --path PATH_TO_LAYOUT \\
+management genesis \
+  --path PATH_TO_LAYOUT \
```
mgorven, line 140: Where do they get PATH_TO_LAYOUT from?

config/management/README.md
davidiw: OWNER and REPO will be shared out of band...

Token retrieval is discussed in `The Process` which is general for everyone.

config/management/README.md
davidiw: Is this only for demo purposes? because in practice this is deployment specific.

config/management/README.md
davidiw: they generate a layout locally and ..

since we'll be doing this in the immediate future, I'll leave out the details :)

config/management/README.md
mgorven: Following my vault-init instructions, the vault server address will be http://localhost:8200. The token suggestion is to distinguish from the Github tokne.

PR #4024, opened 2020-05-21
Bootstrap designated dealer resource with tiered_mint function

moeziniaf:
## Motivation

setup DD and Tiers for tiered_mint flow from Blessed account flow initiation to DD holding minted coins under tiered restrictions.

## Test Plan

Introduces blessed as special account for functional tests
Tests all aspects of blessed account minting for DD -
minting/adding tiers etc tested.
cargo test tiered_mint

## Related PRs


language/stdlib/modules/designated_dealer.move
```diff
@@ -0,0 +1,150 @@
...
+            while (index < number_of_tiers) {
+                // INVALID_TIER_START
+                Txn::assert(new_upperbound < *Vector::borrow(&tiers.tiers, index), 4);
+                index = index + 1;
+            };
+            let tier_mut = Vector::borrow_mut(&mut tiers.tiers, tier_index);
+            *tier_mut = new_upperbound;
+        }
+
+        public fun apply_for_designated_dealer() {
```
sblackshear, line 70: Now that https://github.com/libra/libra/commit/e66e1fc5c9d61a5aace906ec01b60429beb6b485 is in, we could consider making a separate `Account::Role` for `DesignatedDealer`. I think this might make sense because otherwise we will have to pick one of the existing `Account::Role`s for DD and nothing that exists quite fits (VASP, association, unhosted)... In addition,  that would let us do a one-shot creation instead of using the apply/certify  pattern.

language/stdlib/modules/designated_dealer.move
```diff
@@ -0,0 +1,150 @@
+address 0x0 {
+    module DesignatedDealer {
```
sblackshear, line 2: Nit: we usually don't indent inside `address` blocks just to keep the overall indentation reasonable (and because multiple blocks in the same file are uncommon)

language/stdlib/modules/designated_dealer.move
```diff
@@ -0,0 +1,150 @@
...
+        use 0x0::LibraTimestamp;
+        use 0x0::Vector;
+        use 0x0::Transaction as Txn;
+        use 0x0::Libra;
+        use 0x0::LibraAccount;
+        use 0x0::Coin1;
+        use 0x0::Coin2;
+
+        resource struct Dealer {
```
sblackshear, line 11: @tzakian Is there enough overlap with `AccountLimits` to use some the type/logic from there?

language/stdlib/modules/designated_dealer.move
```diff
@@ -0,0 +1,150 @@
...
+            // Time window start in microseconds
```
sblackshear, line 12: Using `///`  comments here and on functions will augment the generated markdown documentation (if desired)

language/stdlib/modules/designated_dealer.move
```diff
@@ -0,0 +1,150 @@
...
+            window_start: u64,
+            // The minted inflow during this time window
+            window_inflow: u64,
+            // The burned outflow during this time window
+            window_outflow: u64,
```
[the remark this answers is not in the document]
moeziniaf: I'll remove it from this PR - still have to check whether a DD burning coin will deduct from mint flow

language/stdlib/modules/designated_dealer.move
moeziniaf: Is the plan to standardize all accounts to LibraAccount::T with some specific RoleData?

sblackshear: > Is the plan to standardize all accounts to LibraAccount::T with some specific RoleData?

Yes, that is the idea.

> Also, this introduces a cyclic dependency since mint_to_designated_dealer calls LibraAccount::deposit. Any suggestions of how to proceed that work with the current architecture?

I would address this by implementing `mint_to_designated_dealer` inside `LibraAccount` instead. In fact, I think you can just repurpose the existing `mint` to (1) check for the `DesignatedDealer` role (2) extract the DD `RoleData` and do the tier checks.

language/stdlib/modules/designated_dealer.move
```diff
@@ -0,0 +1,150 @@
...
+            })
+        }
+
+        public fun add_tier(next_tier_upperbound: u64)
+        acquires Tiers {
+            assert_sender_is_blessed();
+            let tiers = borrow_global_mut<Tiers>(singleton_addr());
+            let number_of_tiers: u64 = Vector::length(&tiers.tiers);
+            // INVALID TIER_ADDITION
+            Txn::assert(number_of_tiers <= 4, 3);
```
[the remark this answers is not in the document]
moeziniaf: For now yes, but there may be some changes :)

language/stdlib/modules/designated_dealer.move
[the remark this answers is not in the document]
moeziniaf: DDs will have per currency accounts and will be tracked that way too

language/stdlib/modules/designated_dealer.move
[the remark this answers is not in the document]
moeziniaf: Yup, no need for outflow at all

moeziniaf:
bumping this so i can unblock some script work :) @tzakian @sblackshear

language/stdlib/modules/designated_dealer.move
```diff
@@ -0,0 +1,129 @@
...
+        Txn::assert(tier_index <= 4, 3);
+        Txn::assert(tier_index < number_of_tiers, 3);
+        // Make sure that this new start for the tier is consistent
+        // with the tiers above it.
+        let index = tier_index + 1;
+        while (index < number_of_tiers) {
+            // INVALID_TIER_START
+            Txn::assert(new_upperbound < *Vector::borrow(tiers, index), 4);
+            index = index + 1;
+        };
```
dmitri-perelman, line 67: nit: it should be enough to check the next tier only, right?

language/stdlib/modules/libra_account.move
```diff
@@ -631,6 +633,68 @@ module LibraAccount {
...
+        let dealer =
+            &mut borrow_global_mut<Role<DesignatedDealer::Dealer>>(addr).role_data;
+        DesignatedDealer::add_tier(dealer, tier_upperbound)
+    }
+
+    public fun update_tier(addr: address, tier_index: u64, new_upperbound: u64) acquires Role {
+        let dealer =
+            &mut borrow_global_mut<Role<DesignatedDealer::Dealer>>(addr).role_data;
+        DesignatedDealer::update_tier(dealer, tier_index, new_upperbound)
+    }
```
dmitri-perelman, line 663: I might be off here, but why do we want these functions in LibraAccount?

language/stdlib/modules/designated_dealer.move
```diff
@@ -0,0 +1,129 @@
...
+    fun reset_window(dealer: &mut Dealer) {
+        assert_sender_is_blessed();
+        let current_time = LibraTimestamp::now_microseconds();
+        if (current_time > dealer.window_start + window_length()) {
+            dealer.window_start = current_time;
+            dealer.window_inflow = 0;
+        }
+    }
+
+    public fun assert_sender_is_blessed() {
```
sblackshear, line 115: We are in  the process of deprecating `Txn::sender()`--would you mind changing this to:

```
    public fun assert_account_is_blessed(account: &signer) {
        Txn::assert(Signer::address_of(account) == singleton_addr(), 0)
    }
```
?

language/stdlib/modules/libra_account.move
sblackshear: This is my fault: an artifact of the way that roles are currently implemented (every `LibraAccount::Role<T>` resource with a `T` for each role). @valerini and I were brainstorming a design pattern to implement this that won't require this pesky wrappers, but probably best to move forward with this approach for now.

language/stdlib/modules/designated_dealer.move
```diff
@@ -0,0 +1,129 @@
...
+            let tier_upperbound: u64 = *Vector::borrow(tiers, tier_index);
+            *tier_check = (cur_inflow + amount <= tier_upperbound);
+        };
+        if (*tier_check) {
+            dealer.window_inflow = cur_inflow + amount;
+        };
+        *tier_check
+    }
+
+    public fun remove_designated_dealer(dealer: &mut Dealer)
```
sblackshear, line 91: Wondering: what should a DD with `is_certified = false` be able to do? Would it be better/simpler to just freeze the account?

language/stdlib/modules/libra_account.move
moeziniaf: The role resource (maintaining Dealer state) is only accessible from within the LibraAccount for now - not just that wrappers are needed but this file will blow up if every function needing account state needs to originate here :)

PR #4025, opened 2020-05-21
[move-prover] Enabling `signer` in move prover and unblocking prover tests.

wrwg:
The type `signer` was already introduced before by the language team, but our type checker wasn't updated to actually recognize it, which broke our tests. Also the prelude did not had native function implementations for it.

For now `signer` is represented by its underlying address, so `address` and `signer` are identical on boogie level.

## Motivation

Unblock prover tests.

Yes

## Test Plan

Existing tests.

## Related PRs

NA


tnowacki:
> For now signer is represented by its underlying address, so address and signer are identical on boogie level.

You can probably keep that representation forever. Unless you need some other wrapping in boogie for resource related things

PR #4031, opened 2020-05-22
[secure-storage] Add support for certificates

davidiw:
This enables certificates specifically for self-hosted vault instances. While
overall pretty intuitive, there are some serious limitations:
* Rustls / WebPKI (ureq's TLS library) expect DNS names and do not support IP addresses
* If DNS resolves to IPv4 and IPv6 but Vault only listens to one of
them, ureq might not be able to connect to it

I'll poke at the ureq code to mitigate the second circumstance, but to
be clear, the SSL cert for Vault and Vault itself must be using a DNS
name and not an IP.

Test method:
Append to /etc/ssl/openssl.cnf:
[v3_ca]
subjectAltName = DNS:localhost

Commands as user:
openssl req -x509 -newkey rsa -keyout key.pem -out cert.pem -days 365 -extensions v3_ca
openssl rsa -in key.pem -out key.pem

Build the following Vault config file (name it config):
listener "tcp" {
    address = "127.0.0.1:8200"
    tls_cert_file = "/path/to/cert.pem"
    tls_key_file = "/path/to/key.pem"
}

vault server -config=config -dev-root-token-id=root_token -dev-listen-address=0.0.0.0:8200 -dev

VAULT_CACERT=/path/to/cert.pem VAULT_ADDR=https://localhost:8200 VAULT_TOKEN=root_token vault secrets enable transit

cargo run -- verify --backend "backend=vault;server=https://localhost:8200;certificate=/path/to/cert.pem;token=/path/to/vault/token"

This will say no values are set if it can connect otherwise it will
complain about storage unavailable.


secure/storage/vault/src/lib.rs
```diff
@@ -76,18 +77,34 @@ impl From<serde_json::Error> for Error {
...
+            let cert = rustls::Certificate(certificate.as_bytes().to_vec());
+            if tls_config.root_store.add(&cert).is_err() {
+                let certs = rustls::internal::pemfile::certs(&mut certificate.as_bytes()).unwrap();
+                tls_config.root_store.add(&certs[0]).unwrap();
+            }
+            Some(Arc::new(tls_config))
+        } else {
+            None
+        };
+        println!("{}", host);
```
davidiw, line 97: this needs to be cleaned up before landing...

secure/storage/vault/src/lib.rs
```diff
@@ -372,9 +365,9 @@ impl Client {
     /// Returns whether or not the vault is unsealed (can be read from / written to). This can be
     /// queried without authentication.
     pub fn unsealed(&self) -> Result<bool, Error> {
-        let resp = ureq::get(&format!("{}/v1/sys/seal-status", self.host))
-            .timeout_connect(TIMEOUT)
-            .call();
+        let request = ureq::get(&format!("{}/v1/sys/seal-status", self.host));
+        let resp = self.upgrade_request_without_token(request).call();
+        println!("{:?}", resp.synthetic_error());
```
davidiw, line 370: Same here, also worth considering using these errors if they exist... but that's another PR

config/config-builder/src/validator_config.rs
```diff
@@ -262,6 +262,7 @@ impl ValidatorConfig {
                 "in-memory" => SecureBackend::InMemoryStorage,
                 "on-disk" => safety_rules_config.backend.clone(),
                 "vault" => SecureBackend::Vault(VaultConfig {
+                    certificate: None,
```
mgorven, line 265: Can we include "CA" in the name (Vault can do client certificate auth, so we should distinguish this from a client cert). `ca_certificate` or `ca_cert` etc.

davidiw: makes sense, renamed to server_certificate

mgorven: Well, it's not the server certificate, it's the CA certificate.
It may actually be nice to use the same environment variables that Vault uses (`VAULT_CACERT` for this) because then users can set it once and Terraform, `vault` and this will all pick it up.

mgorven: (Although I suppose this doesn't work when the Association also uses Vault and there are two CA certs in play.)

davidiw: Yeah, not going to touch environmental variables in this case... as convenient as they are, they create nightmares in this case.

I renamed to ce_certificate. You are right, in my case, it was a "server_certificate" but it generalizes to a "ca_certificate" :).

davidiw:
Also this builds on https://github.com/libra/libra/pull/4035

if this is accepted, I'll just land the entire set via this commit...

PR #4032, opened 2020-05-22
[config] explicit getters for OnChainConfig

agc-quadra:
## Motivation

This is a small PR intended to make the usage of OnChainConfig more readable and explicit.

The change itself makes the parameterized get<T> function private and replaces it with explicit getters for the four config types currently contained in the OnChainConfig.

Yes.

## Test Plan

Cosmetic only change, ensure that existing tests continue to pass.

## Related PRs

### DIFF
--- a/consensus/src/epoch_manager.rs
+++ b/consensus/src/epoch_manager.rs
@@ -369,7 +369,7 @@ impl<T: Payload> EpochManager<T> {
 
     pub async fn start_processor(&mut self, payload: OnChainConfigPayload) {
         let validator_set: ValidatorSet = payload
-            .get()
+            .validator_set()
             .expect("failed to get ValidatorSet from payload");
         let epoch_info = EpochInfo {
             epoch: payload.epoch(),
--- a/state-synchronizer/src/tests/on_chain_config_tests.rs
+++ b/state-synchronizer/src/tests/on_chain_config_tests.rs
@@ -137,7 +137,7 @@ fn test_on_chain_config_pub_sub() {
 
     let receive_reconfig = async {
         let payload = reconfig_receiver.select_next_some().await;
-        let received_config = payload.get::<VMConfig>().unwrap();
+        let received_config = payload.vm_config().unwrap();
         assert_eq!(received_config.publishing_option, vm_publishing_option);
     };
 
--- a/types/src/on_chain_config/mod.rs
+++ b/types/src/on_chain_config/mod.rs
@@ -50,10 +50,10 @@ impl ConfigID {
 
 /// State sync will panic if the value of any config in this registry is uninitialized
 pub const ON_CHAIN_CONFIG_REGISTRY: &[ConfigID] = &[
-    VMConfig::CONFIG_ID,
     LibraVersion::CONFIG_ID,
-    ValidatorSet::CONFIG_ID,
     RegisteredCurrencies::CONFIG_ID,
+    ValidatorSet::CONFIG_ID,
+    VMConfig::CONFIG_ID,
 ];
 
 #[derive(Clone, Debug, PartialEq)]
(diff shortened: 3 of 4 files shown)


davidiw:
I don't have much skin in this game, passing off to @christinacelee, @phoenix-antigravity, @zekun000 ... the classical API is a little interesting :).

zekun000:
personally I don't feel this makes it more readable, the generics is quite explicit about the types (it needs either let foo: Foo = config.get() or let foo = config.get::<Foo>()) and avoid boiler-plate for future types.

agc-quadra:
> personally I don't feel this makes it more readable, the generics is quite explicit about the types (it needs either let foo: Foo = config.get() or let foo = config.get::<Foo>()) and avoid boiler-plate for future types.

Acknowledging that readability is heavily influenced by preference and that bikeshedding the topics is generally not a good use of time, I'll expand on why I believe generic methods (differentiated from generic types) are low on the readability scale.  A decisive :+1: or :-1: will dictate the future of this PR and won't be taken personally in either direction.

1.  [acknowledge] The proposal is a longer definition of the OnChainConfig type and does include additional boilerplate relative to the legacy implementation.  
1. the current interface relies allows for the following code snippet:
    `let config_piece_one = config.get();`
    `let config_piece_two = config.get();`
    `parse_configs(config_piece_one, config_piece_two);`
which type inference will correctly resolve and parse. The client author should use better variable names and `config.get::<VMConfig>()` and reviewers should squawk if that is not the case.  But that relies on human discipline moving forward, which is always dangerous.  (Note that for the author of client code, `config.vm_config()` has a marginally better code golf score.)
1.  The current `get<T>` implementation allows users of the OnChainConfig struct to invoke `get` for any type with a `CONFIG_ID`, even if that type is not a part of OnChainConfig.  This behavior can lead to maintenance pain as configs are added/removed over time as the resulting error is indistinguishable from a case where the requested type is stored but not available.  If unit tests are not sufficient, this error can very easily lead to production suffering.  The proposed interface does not eliminate this problem entirely as a field can be replaced without updating the explicit getter, but it dos mean that properly performed updates to OnChainConfig will result in compilation errors in any calling code.   This explicitly highlights a choice between consolidating the maintenance complexity of OnChainConfig to on_chain_config.rs or distributing that complexity amongst all calling locations.
1.  Generic types, i.e. `vector<T>`, also have function signatures akin to `T get<T>`.  The key difference between the generic type `get` of vector and the generic method `get` of OnChainConfig is that the former uses the generic to define the type for the collection while the latter uses the generic to select between different internal fields of the struct.  Generic type, are (per collection) straightforward and core knowledge for any Rust or C++ programmer at this point in time;  generic methods require expertise of level ~8 (out of 10) to understand and use well.  While there are occasional benefits to relying on expert language features from time to time, I've typically found it easier to (a) understand, (b) contribute to, and (c) maintain in production codebases which are restricted to language features at expertise level 6 or below.   

The underlying question of this PR is a calibration question to better understand the assumed language proficiency of people reading/contributing to/running the codebase.

zekun000:
@agc-quadra thanks for your thoughtful comment, here's some additional thoughts

2) type erasure, if the compiler is able to erase the type, I think it's reasonable for readers too (with a good var name)
3) better errors, the generic on the getter is bound by the trait OnChainConfig which should be the source of truth. In order to use the type, the author has to impl the trait first, otherwise it's a compile error. I'd argue it's easier to maintain one source of truth instead of both the trait and the explicit getter. thus the value of the explicit getter would be much more if we don't have such trait bound.
4) it's a heterogeneous generic use case instead of homogeneous one like vector.

agc-quadra:
@zekun000 Thanks for the additional thoughts.  The discussion certainly falls into a gray area with no clear right/wrong answer.   

Unless somebody speaks up in favor, I will close this PR in 24 hours.

agc-quadra:
Closing PR per previous comment.

Issue #4012, opened 2020-05-21
[network] disambiguate networks with `NetworkId` to fix network metrics/logs.


phlip9:
Replaces prev. issue #2397

Relevant issue #3341

### Problem

It's painful to view prometheus metrics or node logs for network when multiple network instances are running, e.g., a validator is connected to the validator swarm via one instance and their fullnode swarm via another instance. The issue is that network metrics don't have an extra label/dimension for the network type to ensure that the values aren't accidentally combining. Similarly, most log entries in network are difficult to disambiguate across multiple network instances. For example, the Rpc "cancelled requests" metric doesn't have a label for the network type, so it accidentally counts the total number of cancelled requests across all network instances without any way to refine the query to a specific network instance.

A separate (but somewhat orthogonal) problem is that application protocols like state sync or mempool have trouble disambiguating between network messages from different network instances. To unblock those protocols, we recently added some quick changes to state sync/mempool to deal with ambiguous network messages (#3568). However, we should probably find a long-term solution to this issue in network itself.

### Solution

Since different network instances are currently ambiguous, we can add a `NetworkId` to uniquely identify each one. This type will probably look something like `RoleType`. Unfortunately, we can't use `RoleType` to disambiguate networks as the existing `enum RoleType { Validator, FullNode }` describes a specific node's role inside a network instance, rather than the network instance itself.

Instead we can probably get by with exhaustively enumerating the network instance types in `NetworkId`:

```rust
// should probably look something like RoleType:
// https://github.com/libra/libra/blob/master/config/src/config/mod.rs#L110

#[repr(u8)]
enum NetworkType {
    Validator = 0,
    Public,
    Private,
}
struct NetworkId {
    network_type: NetworkType,
    network_name: String
    // ..
}
```

+ Each `NetworkConfig` should have a `network_id`. (https://github.com/libra/libra/blob/master/config/src/config/network_config.rs#L26)

+ `config-builder` should emit the right `network_id` for the `NetworkConfig`s it generates.
(https://github.com/libra/libra/blob/master/config/config-builder/src/validator_config.rs, https://github.com/libra/libra/blob/master/config/config-builder/src/full_node_config.rs, https://github.com/libra/libra/blob/master/config/src/generator.rs)

+ Will need some sanity checks in `NodeConfig` to check we don't have multiple `NetworkConfig`s with the same `NetworkId`.

[...]

PR #4034, opened 2020-05-22
[draft][network] simple on-chain discovery via state-sync reconfig

agc-quadra:
## Motivation

Draft work on  https://github.com/libra/libra/issues/3925

(Write your answer here.)

## Test Plan

## Related PRs

### DIFF
--- a/consensus/src/epoch_manager.rs
+++ b/consensus/src/epoch_manager.rs
@@ -375,12 +375,6 @@ impl<T: Payload> EpochManager<T> {
             epoch: payload.epoch(),
             verifier: (&validator_set).into(),
         };
-        let validator_keys = validator_set.payload().to_vec();
-        info!("Update Network about new validators");
-        self.network_sender
-            .update_eligible_nodes(validator_keys)
-            .await
-            .expect("Unable to update network's eligible peers");
 
         match self.storage.start() {
             LivenessStorageData::RecoveryData(initial_data) => {
--- a/libra-node/src/main_node.rs
+++ b/libra-node/src/main_node.rs
@@ -1,21 +1,6 @@
 // Copyright (c) The Libra Core Contributors
 // SPDX-License-Identifier: Apache-2.0
 
-use std::{
-    boxed::Box,
-    collections::HashMap,
-    net::ToSocketAddrs,
-    sync::Arc,
-    thread,
-    time::{Duration, Instant},
-};
-
-use futures::{channel::mpsc::channel, executor::block_on, stream::StreamExt};
-use tokio::{
-    runtime::{Builder, Handle, Runtime},
-    time::interval,
-};
-
 use admission_control_service::admission_control_service::AdmissionControlService;
 use backup_service::start_backup_service;
 use consensus::consensus_provider::start_consensus;
@@ -25,6 +10,7 @@ use debug_interface::{
 };
 use executor::{db_bootstrapper::bootstrap_db_if_empty, Executor};
 use executor_types::ChunkExecutor;
+use futures::{channel::mpsc::channel, executor::block_on, stream::StreamExt};
 use libra_config::{
     config::{DiscoveryMethod, NetworkConfig, NodeConfig, RoleType},
     utils::get_genesis_txn,
@@ -41,9 +27,21 @@ use network::validator_network::network_builder::{
 };
 use onchain_discovery::{client::OnchainDiscovery, service::OnchainDiscoveryService};
 use state_synchronizer::StateSynchronizer;
+use std::{
+    boxed::Box,
+    collections::HashMap,
+    net::ToSocketAddrs,
+    sync::Arc,
+    thread,
+    time::{Duration, Instant},
+};
 use storage_interface::{DbReader, DbReaderWriter};
 use storage_service::start_storage_service_with_db;
 use subscription_service::ReconfigSubscription;
+use tokio::{
+    runtime::{Builder, Handle, Runtime},
+    time::interval,
+};
 
 const AC_SMP_CHANNEL_BUFFER_SIZE: usize = 1_024;
 const INTRA_NODE_CHANNEL_BUFFER_SIZE: usize = 1;
--- a/libra-node/src/main_node.rs
+++ b/libra-node/src/main_node.rs
@@ -33,11 +33,11 @@ use libra_json_rpc::bootstrap_from_config as bootstrap_rpc;
 use libra_logger::prelude::*;
 use libra_mempool::MEMPOOL_SUBSCRIBED_CONFIGS;
 use libra_metrics::metric_server;
-use libra_types::{on_chain_config::ON_CHAIN_CONFIG_REGISTRY, PeerId, waypoint::Waypoint};
+use libra_types::{on_chain_config::ON_CHAIN_CONFIG_REGISTRY, waypoint::Waypoint, PeerId};
 use libra_vm::LibraVM;
 use libradb::LibraDB;
 use network::validator_network::network_builder::{
-    AuthenticationMode, HANDSHAKE_VERSION, NetworkBuilder,
+    AuthenticationMode, NetworkBuilder, HANDSHAKE_VERSION,
 };
 use onchain_discovery::{client::OnchainDiscovery, service::OnchainDiscoveryService};
 use state_synchronizer::StateSynchronizer;
@@ -69,10 +69,10 @@ fn setup_debug_interface(config: &NodeConfig) -> Runtime {
         "{}:{}",
         config.debug_interface.address, config.debug_interface.admission_control_node_debug_port,
     )
-        .to_socket_addrs()
-        .unwrap()
-        .next()
-        .unwrap();
+    .to_socket_addrs()
+    .unwrap()
+    .next()
+    .unwrap();
     rt.spawn(
         tonic::transport::Server::builder()
             .add_service(NodeDebugInterfaceServer::new(NodeDebugService::new()))
@@ -257,7 +257,7 @@ pub fn setup_environment(node_config: &mut NodeConfig) -> LibraHandle {
             false, /* readonly */
             node_config.storage.prune_window,
         )
-            .expect("DB should open."),
+        .expect("DB should open."),
     );
     let _simple_storage_service =
         start_storage_service_with_db(&node_config, Arc::clone(&libra_db));
@@ -313,9 +313,13 @@ pub fn setup_environment(node_config: &mut NodeConfig) -> LibraHandle {
         mempool_network_handles.push((network.peer_id, mempool_sender, mempool_events));
         validator_network_provider = Some((network.peer_id, runtime, network_builder));
         // Set up the network configuration event handling by establishing the subscription and starting the config change processor.
-        let (validator_subscription, validator_network_events) = ReconfigSubscription::subscribe(VALIDATOR_CONFIG_CHANGES);
+        let (validator_subscription, validator_network_events) =
+            ReconfigSubscription::subscribe(VALIDATOR_CONFIG_CHANGES);
         reconfig_subscription.push(validator_subscription);
-        let network_config_listener = ConfigurationChangeListener::new(network_builder.conn_mgr_reqs_tx(), validator_network_events);
+        let network_config_listener = ConfigurationChangeListener::new(
+            network_builder.conn_mgr_reqs_tx(),
+            validator_network_events,
+        );
         network_config_listener.run();
     }
 
@@ -346,9 +350,13 @@ pub fn setup_environment(node_config: &mut NodeConfig) -> LibraHandle {
         debug!("Network started for peer_id: {}", full_node_network.peer_id);
 
         // Set up the network configuration event handling by establishing the subscription and starting the config change processor.
-        let (full_node_network_subscription, full_node_network_events) = ReconfigSubscription::subscribe(VALIDATOR_CONFIG_CHANGES(network_id));
+        let (full_node_network_subscription, full_node_network_events) =
+            ReconfigSubscription::subscribe(VALIDATOR_CONFIG_CHANGES(network_id));
         reconfig_subscription.push(full_node_network_subscription);
-        let network_config_listener = ConfigurationChangeListener::new(network_builder.conn_mgr_reqs_tx(), full_node_network_events);
+        let network_config_listener = ConfigurationChangeListener::new(
+            network_builder.conn_mgr_reqs_tx(),
+            full_node_network_events,
+        );
         network_config_listener.run();
     }
 
--- a/libra-node/src/main_node.rs
+++ b/libra-node/src/main_node.rs
@@ -1,6 +1,21 @@
 // Copyright (c) The Libra Core Contributors
 // SPDX-License-Identifier: Apache-2.0
 
+use std::{
+    boxed::Box,
+    collections::HashMap,
+    net::ToSocketAddrs,
+    sync::Arc,
+    thread,
+    time::{Duration, Instant},
+};
+
+use futures::{channel::mpsc::channel, executor::block_on, stream::StreamExt};
+use tokio::{
+    runtime::{Builder, Handle, Runtime},
+    time::interval,
+};
+
 use admission_control_service::admission_control_service::AdmissionControlService;
 use backup_service::start_backup_service;
 use consensus::consensus_provider::start_consensus;
@@ -10,7 +25,6 @@ use debug_interface::{
 };
 use executor::{db_bootstrapper::bootstrap_db_if_empty, Executor};
 use executor_types::ChunkExecutor;
-use futures::{channel::mpsc::channel, executor::block_on, stream::StreamExt};
 use libra_config::{
     config::{DiscoveryMethod, NetworkConfig, NodeConfig, RoleType},
     utils::get_genesis_txn,
@@ -19,27 +33,17 @@ use libra_json_rpc::bootstrap_from_config as bootstrap_rpc;
 use libra_logger::prelude::*;
 use libra_mempool::MEMPOOL_SUBSCRIBED_CONFIGS;
 use libra_metrics::metric_server;
-use libra_types::{on_chain_config::ON_CHAIN_CONFIG_REGISTRY, waypoint::Waypoint, PeerId};
+use libra_types::{on_chain_config::ON_CHAIN_CONFIG_REGISTRY, PeerId, waypoint::Waypoint};
 use libra_vm::LibraVM;
 use libradb::LibraDB;
-use network::validator_network::network_builder::{AuthenticationMode, NetworkBuilder};
+use network::validator_network::network_builder::{
+    AuthenticationMode, HANDSHAKE_VERSION, NetworkBuilder,
+};
 use onchain_discovery::{client::OnchainDiscovery, service::OnchainDiscoveryService};
 use state_synchronizer::StateSynchronizer;
-use std::{
-    boxed::Box,
-    collections::HashMap,
-    net::ToSocketAddrs,
-    sync::Arc,
-    thread,
-    time::{Duration, Instant},
-};
 use storage_interface::{DbReader, DbReaderWriter};
 use storage_service::start_storage_service_with_db;
 use subscription_service::ReconfigSubscription;
-use tokio::{
-    runtime::{Builder, Handle, Runtime},
-    time::interval,
-};
 
 const AC_SMP_CHANNEL_BUFFER_SIZE: usize = 1_024;
 const INTRA_NODE_CHANNEL_BUFFER_SIZE: usize = 1;
@@ -65,10 +69,10 @@ fn setup_debug_interface(config: &NodeConfig) -> Runtime {
         "{}:{}",
         config.debug_interface.address, config.debug_interface.admission_control_node_debug_port,
     )
-    .to_socket_addrs()
-    .unwrap()
-    .next()
-    .unwrap();
+        .to_socket_addrs()
+        .unwrap()
+        .next()
+        .unwrap();
     rt.spawn(
         tonic::transport::Server::builder()
             .add_service(NodeDebugInterfaceServer::new(NodeDebugService::new()))
@@ -253,7 +257,7 @@ pub fn setup_environment(node_config: &mut NodeConfig) -> LibraHandle {
             false, /* readonly */
             node_config.storage.prune_window,
         )
-        .expect("DB should open."),
+            .expect("DB should open."),
     );
     let _simple_storage_service =
         start_storage_service_with_db(&node_config, Arc::clone(&libra_db));
@@ -308,6 +312,11 @@ pub fn setup_environment(node_config: &mut NodeConfig) -> LibraHandle {
         );
         mempool_network_handles.push((network.peer_id, mempool_sender, mempool_events));
         validator_network_provider = Some((network.peer_id, runtime, network_builder));
+        // Set up the network configuration event handling by establishing the subscription and starting the config change processor.
+        let (validator_subscription, validator_network_events) = ReconfigSubscription::subscribe(VALIDATOR_CONFIG_CHANGES);
+        reconfig_subscription.push(validator_subscription);
+        let network_config_listener = ConfigurationChangeListener::new(network_builder.conn_mgr_reqs_tx(), validator_network_events);
+        network_config_listener.run();
     }
 
     for mut full_node_network in node_config.full_node_networks.iter_mut() {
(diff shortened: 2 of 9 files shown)


agc-quadra:
@davidiw This is very much a draft PR trying to work out the structure for funneling updates from statesync into the network.   This reflects my current mental model and mappings.

Key pieces at the moment:

1. Very basic NetworkConfigListener added
1. removing the deprecated code from epoch_manager
1. sketching out the new subscription registration patterns

Blocking/Unresolved issues

- Is this structurally along the lines of what you are thinking or have I jumped the rails already?
- StateSync currently does not expose FullNodeNetworks as far as I can tell
- StateSync currently does not expose NodeAddress changes as far as I can tell
- There are probably a number of race conditions around registration, process starting, etc. that need to be plumed through
- Are there tensions around StateSync using the same network ConnectivityManager which we are attempting to update based on StateSync provided information?
- Are there timing concerns for consensus relative to the network and the LivenessStorage updates?

libra-node/src/main_node.rs
```diff
@@ -1,6 +1,21 @@
...
 // SPDX-License-Identifier: Apache-2.0
 
+use std::{
+    boxed::Box,
+    collections::HashMap,
+    net::ToSocketAddrs,
+    sync::Arc,
+    thread,
+    time::{Duration, Instant},
+};
```
davidiw, line 11: The normal order of things has been super, crate, alphabetical.

libra-node/src/main_node.rs
```diff
@@ -19,29 +33,17 @@ use libra_json_rpc::bootstrap_from_config as bootstrap_rpc;
 use libra_logger::prelude::*;
 use libra_mempool::MEMPOOL_SUBSCRIBED_CONFIGS;
 use libra_metrics::metric_server;
-use libra_types::{on_chain_config::ON_CHAIN_CONFIG_REGISTRY, waypoint::Waypoint, PeerId};
+use libra_types::{on_chain_config::ON_CHAIN_CONFIG_REGISTRY, PeerId, waypoint::Waypoint};
 use libra_vm::LibraVM;
 use libradb::LibraDB;
 use network::validator_network::network_builder::{
-    AuthenticationMode, NetworkBuilder, HANDSHAKE_VERSION,
+    AuthenticationMode, HANDSHAKE_VERSION, NetworkBuilder,
```
davidiw, line 40: in general I would expect cargo x fmt to call this out, was this done manually?

libra-node/src/main_node.rs
```diff
@@ -67,10 +69,10 @@ fn setup_debug_interface(config: &NodeConfig) -> Runtime {
         "{}:{}",
         config.debug_interface.address, config.debug_interface.admission_control_node_debug_port,
     )
-    .to_socket_addrs()
-    .unwrap()
-    .next()
-    .unwrap();
+        .to_socket_addrs()
```
davidiw, line 72: haha, I'm gonna send out an easy PR to fix this nonsense, we should be extracting a socket address directly from the config...

network/simple-onchain-discovery/src/lib.rs
```diff
@@ -0,0 +1,102 @@
...
+        mut reconfig_events: libra_channel::Receiver<(), OnChainConfigPayload>,
+    ) {
+        loop {
+            let pre_select_instant = Instant::now();
+            let _idle_duration;
+            select! {
+                  payload = reconfig_events.select_next_some() => {
+                      _idle_duration = pre_select_instant.elapsed();
+                      self.process_payload(payload).await
+              }
```
davidiw, line 96: probably just await unless we'll have multiple channels?

network/simple-onchain-discovery/src/lib.rs
```diff
@@ -0,0 +1,102 @@
...
+            .get()
+            .expect("failed to get ValidatorSet from payload");
+
+        let update_eligible_nodes_request =
+            validator_set_to_update_eligible_nodes_request(validator_set);
+        info!("Update Network about new validators");
+        self.conn_mgr_reqs_tx
+            .send(update_eligible_nodes_request)
+            .await
+            .expect("Unable to update network's eligible peers");
```
davidiw, line 78: Do we want to panic the system if this fails? Possibly, would be good to double check.

network/simple-onchain-discovery/src/lib.rs
```diff
@@ -0,0 +1,102 @@
...
+                        signing_public_key: validator.network_signing_public_key().clone(),
+                    },
+                )
+            })
+            .collect(),
+    )
+}
+
+impl ConfigurationChangeListener {
+    pub fn new(conn_mgr_reqs_tx: channel::Sender<ConnectivityRequest>) -> Self {
```
davidiw, line 62: I guess each one of these would specify if they are listening for updates to fullnodes or to validators and then pass the appropriate data sets via the ConnectivityRequest

davidiw: You can use the Role in a network config to help dictate which would be appropriate.

network/simple-onchain-discovery/src/lib.rs
```diff
@@ -0,0 +1,102 @@
+// Copyright (c) The Libra Core Contributors
+// SPDX-License-Identifier: Apache-2.0
+
+use futures::sink::SinkExt;
+use futures::{select, StreamExt};
+use once_cell::sync::Lazy;
```
davidiw, line 6: nit on ordering, while there is something more reasonable about your approach, it would result in much bike shedding :P

davidiw:
Coming along nicely!

libra-node/src/main_node.rs
agc-quadra: My IDE has a "optimize includes" button that I pressed, thinking it would remove unused includes (it didnt).  I'll revert this to the original orderings before moving this out of draft.

libra-node/src/main_node.rs
agc-quadra: see "optimize includes" previously.  I believe I ran cargo x fmt, but will redo to confirm.

network/simple-onchain-discovery/src/lib.rs
agc-quadra: baby steps.  but yes.

network/simple-onchain-discovery/src/lib.rs
agc-quadra: no to bike shedding.  divergence from standard is not intended here.

agc-quadra:
> the ValidatorInfo code is pretty rough and needs some TLC.

+10000.  there are a couple of pieces that are confusing me here:

- Why is the full node configuration a part of `ValidatorInfo`?  My understanding is that a node does not need to be a part of a validator network (or to be a validator) in order to be a member of a full node network?

-  How are different full node networks differentiated?  My understanding is that there are an unbounded number of full node networks, each with their own `NetworkBuilder` and `ConnectionManager`, which are configured independently.

- Are all full node networks maintained in the on chain config database?  Some?  Exactly 1 consisting only of validators?

The sensible way to proceed (and amount of technical debt to incur) seems to depend heavily on the answer to the last question.

network/simple-onchain-discovery/src/lib.rs
```diff
@@ -0,0 +1,168 @@
...
+            ConnectivityRequest::UpdateAddresses(
+                DiscoverySource::OnChain,
+                // TODO:  again, this seems wrong for the account address
+                full_node.account_address().clone(),
+                full_node.config().full_node_network_address().clone(),
+            )
+        })
+        .collect();
+    (addresses, nodes)
+}
```
agc-quadra, line 117: @davidiw helper function to extract the full_node relevant information.

There are a number of pieces that are very opaque here and feel like they might be wrong/misguided/incomplete.   

Key issues:
1). what is the account_address for a full_node?  Is it the same as for the validator on that machine?  (i.e., is the account_address actually a machine address????)
2).  are the full_node identity and signing keys guaranteed to be the same?

network/simple-onchain-discovery/src/lib.rs
agc-quadra: This is copied from epoch_manager and inheriting the same behaviors.

network/simple-onchain-discovery/src/lib.rs
```diff
@@ -0,0 +1,198 @@
...
+    let full_node_infos = full_node_set.payload().to_vec();
+
+    // Collect the set of updated addresses.
+    let updates: Vec<ConnectivityRequest> = full_node_infos
+        .into_iter()
+        .map(|full_node| {
+            ConnectivityRequest::UpdateAddresses(
+                DiscoverySource::OnChain,
+                // TODO:  this seems wrong for the account address.  Is it shared with the validator?
+                full_node.account_address().clone(),
```
davidiw, line 89: well this is truly a mess... we actually want this to be account_address::from_public_key where the public key is the identity key. We really need to land on how handle peers with inside connectivity manager otherwise we have a nice memory leak -- since that's currently a hash and not some cache limited capacity hash.

network/simple-onchain-discovery/src/lib.rs
```diff
@@ -0,0 +1,198 @@
...
+            .map(|full_node| {
+                (
+                    // TODO:  This seems like it is wrong and should be something else.  Or is the full_node account address the same as the validator?
+                    full_node.account_address().clone(),
+                    NetworkPublicKeys {
+                        identity_public_key: full_node
+                            .config()
+                            .full_node_network_identity_public_key
+                            .clone(),
+                        // TODO:  there are not separate keys for the full_node.  Are they therefore the same?
```
davidiw, line 111: These would be the same... our current approach is that a validator also runs a full node and that these are tightly coupled. Hence these are supposed to be trusted fullnodes that should map to a specific validator. Longer term, we might envision supporting 3rd party fullnodes on chain, but that is a bit more of utility / complexity that hasn't been properly evaluated.

network/simple-onchain-discovery/src/lib.rs
```diff
@@ -0,0 +1,198 @@
...
+                    info!(
+                        "Update {} network about new Node IDs",
+                        self.role.to_string()
+                    );
+                }
+                ConnectivityRequest::GetDialQueueSize(a) => {
+                    unimplemented!("Not expecting to receive this change!");
+                }
+            };
+            self.conn_mgr_reqs_tx.send(update).await.expect(&err_msg);
```
davidiw, line 177: is this an unhandlable error?

also consider join! or join_all

https://rust-lang.github.io/async-book/06_multiple_futures/02_join.html
https://docs.rs/futures/0.2.1/futures/future/fn.join_all.html

network/simple-onchain-discovery/src/lib.rs
```diff
@@ -0,0 +1,198 @@
...
+        }
+    }
+
+    /// Starts the listener to wait on reconfiguration events.  Creates an infinite loop.
+    pub async fn start(
+        mut self,
+        mut reconfig_events: libra_channel::Receiver<(), OnChainConfigPayload>,
+    ) {
+        loop {
```
davidiw, line 186: if you wanted to venture into an integration test, which I really don't care if you do, you could consider splitting this out and loop around a helper function that actually does this process.

network/simple-onchain-discovery/src/lib.rs
```diff
@@ -0,0 +1,198 @@
...
+        };
+
+        info!(
+            "Update {} Network about new Node IDs",
+            self.role.to_string()
+        );
+        for mut update in updates {
+            let err_msg;
+            match &update {
+                ConnectivityRequest::UpdateAddresses(a, b, c) => {
```
davidiw, line 154: maybe clearer names here? should we consider putting this logging into the place where we formatted the objects? is this duplicative of what is happening within the connection manager?

davidiw:
> > the ValidatorInfo code is pretty rough and needs some TLC.
> 
> +10000. there are a couple of pieces that are confusing me here:
> 
> * Why is the full node configuration a part of `ValidatorInfo`?  My understanding is that a node does not need to be a part of a validator network (or to be a validator) in order to be a member of a full node network?
We had planned to have two different data sets, but it added a lot of complexity to code when we only intended to ever host a single fullnode network and in the our initial release would be 1:1 with the validator network.
> * How are different full node networks differentiated?  My understanding is that there are an unbounded number of full node networks, each with their own `NetworkBuilder` and `ConnectionManager`, which are configured independently.
That's possible. In reality, there are likely to be 3 networks:
* Validator all-to-all
* (Validator)-fullnode-to-validator
* (Public)-fullnode-to-(validator)-fullnode
At some point in time I should write up a brief proposal that describes what I envision as one potential long term vision of producing a self-organizing Libra PFN, but not today...
> * Are all full node networks maintained in the on chain config database?  Some?  Exactly 1 consisting only of validators?
No, probably alluded to this many times. Onchain is a trusted set of fullnodes, they somehow have been trusted enough by the validators to be added to the blockchain. More specifically, they are the validators' fullnodes.
> 
> The sensible way to proceed (and amount of technical debt to incur) seems to depend heavily on the answer to the last question.

No tech debt, just not the problem we can reasonably hope to solve in the short term.

agc-quadra:
> > * How are different full node networks differentiated?  My understanding is that there are an unbounded number of full node networks, each with their own `NetworkBuilder` and `ConnectionManager`, which are configured independently.
> >   That's possible. In reality, there are likely to be 3 networks:
> 
> * Validator all-to-all
> * (Validator)-fullnode-to-validator
> * (Public)-fullnode-to-(validator)-fullnode
>   At some point in time I should write up a brief proposal that describes what I envision as one potential long term vision of producing a self-organizing Libra PFN, but not today...


3 networks or 3 types of networks?    I don't understand the descriptions above and I was actually asking a slightly simpler question.

Looking at [main_node.rs](https://github.com/libra/libra/blob/9bae2e911c653935a23862f31586f756b58ff675/libra-node/src/main_node.rs#L334), there are potentially multiple full_node networks instantiated each with its own [NetworkBuilder](https://github.com/libra/libra/blob/9bae2e911c653935a23862f31586f756b58ff675/libra-node/src/main_node.rs#L335) and each NetworkBuilder has its own [ConnectivityManager] (https://github.com/libra/libra/blob/9bae2e911c653935a23862f31586f756b58ff675/libra-node/src/main_node.rs#L142).   My question is how theses different networks are differentiated?  When a config change is registered, which of the networks changes?

agc-quadra:
> * Validator all-to-all
> * (Validator)-fullnode-to-validator
> * (Public)-fullnode-to-(validator)-fullnode
>   At some point in time I should write up a brief proposal that describes what I envision as one potential long term vision of producing a self-organizing Libra PFN, but not today...

Pushing on the three types of networks here and their use cases,

* validator all-to-all is the network that supports consensus and the mempool?
* for the other two network types, what does it mean for the network to be directional from the fullnode to the validator (or validator to fullnode)?

From experience with the config-builder and associated documentation, my expectation  was that there would be three types of networks:

1. the validator-to-validator network for consensus
1. a public facing full node network which mirrors the database for public consumption (and buffers the core network from most interactions)
1. a collection of privately run full node networks which "hang" off of members of one of the first two networks.

That understanding is driven partially by explicit documentation---points 1 and 3 are mostly implied by [Full Node Basics](https://libra.org/en-US/blog/full-node-basics/)---and my own intuitions---point 2.

network/simple-onchain-discovery/src/lib.rs
agc-quadra: Switched to join! and removed the expect.

I havent been thinking about error handling yet, and instead have been focusing on the data ingestion/conversion.

network/simple-onchain-discovery/src/lib.rs
agc-quadra: ahh, so this should be the validator public key?

network/simple-onchain-discovery/src/lib.rs
```diff
@@ -0,0 +1,198 @@
...
+                        signing_public_key: validator.network_signing_public_key().clone(),
+                    },
+                )
+            })
+            .collect(),
+    ));
+    updates
+}
+
+fn extract_full_node_updates(full_node_set: ValidatorSet) -> Vec<ConnectivityRequest> {
```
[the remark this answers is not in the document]
agc-quadra: The tension I see here is that the difference between these two functions lies in which fields of xxx.config() are extracted internal to the map.    

At my level of rust-foo, the two options are to write two separate functions (as currently) or to write one function which has the same code split with the match statement.

I lean towards the former.

network/simple-onchain-discovery/src/lib.rs
```diff
@@ -0,0 +1,198 @@
...
+    // Collect the set of address updates.
+    let updates: Vec<ConnectivityRequest> = validator_keys
+        .into_iter()
+        .map(|validator| {
+            ConnectivityRequest::UpdateAddresses(
+                DiscoverySource::OnChain,
+                validator.account_address().clone(),
+                vec![
+                    NetworkAddress::try_from(&validator.config().validator_network_address)
+                        .expect("WTF"),
```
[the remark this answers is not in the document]
agc-quadra: I'm torn.  On the one hand your instinct is much safer in production.  On the other hand, this is a value which is coming out of the config and should have been properly structured when it went in;  so if this try_from fails there are much bigger problems afoot....

network/simple-onchain-discovery/src/lib.rs
```diff
@@ -0,0 +1,198 @@
...
+
+fn extract_validator_updates(validator_set: ValidatorSet) -> Vec<ConnectivityRequest> {
+    let validator_keys = validator_set.payload().to_vec();
+    // Collect the set of address updates.
+    let updates: Vec<ConnectivityRequest> = validator_keys
+        .into_iter()
+        .map(|validator| {
+            ConnectivityRequest::UpdateAddresses(
+                DiscoverySource::OnChain,
+                validator.account_address().clone(),
```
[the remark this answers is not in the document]
agc-quadra: I had the same thought.   Taking a step back, what would be great to have is a handler which receives a NetworkConnectivityConfig from any source (on chain discovery, gossip discovery, seed file, ....) and propagates those changes into the ConnectivityManager.  That is sort of what is being done right now through the ConnectivityRequest interface, but we're writing multiple of these handlers where I think it would be simpler to write one handler that converst a NetworkConnectivityConfig into the update and a collection of ::From<> implementations that the various discovery sources implement to convert their datatype into the standard NetworkConnectivityConfig which the ConnectivityManager cares about.

davidiw:
> > > * How are different full node networks differentiated?  My understanding is that there are an unbounded number of full node networks, each with their own `NetworkBuilder` and `ConnectionManager`, which are configured independently.
> > >   That's possible. In reality, there are likely to be 3 networks:
> > 
> > 
> > 
> > * Validator all-to-all
> > * (Validator)-fullnode-to-validator
> > * (Public)-fullnode-to-(validator)-fullnode
> >   At some point in time I should write up a brief proposal that describes what I envision as one potential long term vision of producing a self-organizing Libra PFN, but not today...
> 
> 3 networks or 3 types of networks? I don't understand the descriptions above and I was actually asking a slightly simpler question.
> 
> Looking at [main_node.rs](https://github.com/libra/libra/blob/9bae2e911c653935a23862f31586f756b58ff675/libra-node/src/main_node.rs#L334), there are potentially multiple full_node networks instantiated each with its own [NetworkBuilder](https://github.com/libra/libra/blob/9bae2e911c653935a23862f31586f756b58ff675/libra-node/src/main_node.rs#L335) and each NetworkBuilder has its own [ConnectivityManager](https://github.com/libra/libra/blob/9bae2e911c653935a23862f31586f756b58ff675/libra-node/src/main_node.rs#L142). My question is how theses different networks are differentiated? When a config change is registered, which of the networks changes?

This really doesn't answer the question directly but here's a direction we're going to help disambiguate networks: https://github.com/libra/libra/issues/4012

In general, the only two networks that would be affected by config changes are:
* validator to validator (or validator in the issue)
* public to validator full node (or public in the issue)

Personally, I would focus on nailing down the validator network first and foremost and not care about the others in this PR. That way we can get this landed and clean up the rest as things come together.

network/simple-onchain-discovery/src/lib.rs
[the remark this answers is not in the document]
agc-quadra: If it is coming from the blockchain, then I would actually expect it to be fully vetted and parseable *before* the blockchain accepts the change including the new address.  The fact that it is on the blockchain is actually the only reason I would consier trusting that the data is parsable.

(Admittedly, such expectations are dangerous and double checking is always at least as safe as relying on a previously performed check)

network/simple-onchain-discovery/src/lib.rs
agc-quadra: It is duplicate-esque to what is happening in ConnectivityManager, but there is an asynchronous data transition between the two logging points.  Failure in one should correlate with failure in the other, but does not necessarily if some other part of the system happens to be borked as well.

consensus/src/epoch_manager.rs
```diff
@@ -375,12 +375,6 @@ impl<T: Payload> EpochManager<T> {
             epoch: payload.epoch(),
             verifier: (&validator_set).into(),
         };
-        let validator_keys = validator_set.payload().to_vec();
-        info!("Update Network about new validators");
-        self.network_sender
-            .update_eligible_nodes(validator_keys)
-            .await
-            .expect("Unable to update network's eligible peers");
```
agc-quadra, line 383: This segment for updating the eligible nodes is relocated to the simple-onchain handler.

Epoch_manager must continue  to monitor these changes from state sync in order to update LivenessStorage below.

network/config/src/node_config.rs
```diff
@@ -0,0 +1,16 @@
...
+}
+
+struct NetworkMembershipConfig {
+    network_id : string,
+    nodes: Map<AccountdAddress, NetworkNodeConfig>,
+}
+
+struct NetworkOperationalConfig {
+
+}
```
agc-quadra, line 16: File to be removed.  Part of scratch work.

network/simple-onchain-discovery/src/lib.rs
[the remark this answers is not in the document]
agc-quadra: https://github.com/libra/libra/pull/4097. consolidates the UpdateAddress interface.

network/config/Cargo.toml
```diff
@@ -0,0 +1,9 @@
+[package]
+name = "config"
+version = "0.1.0"
+authors = ["Allen Clement <aclement@gmail.com>"]
```
davidiw, line 4: well I suspect that this is just for testing, maybe?  these should just be copypasta of others due to CLA, afaiu.

network/config/src/node_config.rs
davidiw: Yeah, let's not do this work in this PR.

network/simple-onchain-discovery/src/lib.rs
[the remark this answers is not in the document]
phlip9: the `validator_network_address` on chain is just an opaque byte array (`vector<u8>` in move). there are several reasons for not parsing them further on chain:

1. we want to be able to upgrade network transport and handshake protocols w/o needing to do an upgrade of the consensus-critical on-chain validator set. instead, validators can just advertise new and old protocols simultaneously.

2. clients running an older binary might not understand some of the advertised addresses and will need to ignore them, which implies that we have to parse the addresses anyway. this is also a natural occurrence during an upgrade rollout.

3. move lang is not really mature enough yet to implement the parsing code. for example, i don't think it supports enums, among other things.

4. additional maintenance burden of support move parsing logic and ensuring it matches the rust logic.

so this code should probably just ignore network addresses it can't understand. we should probably log those that fail, however.

agc-quadra:
> In general, the only two networks that would be affected by config changes are:
> 
> * validator to validator (or validator in the issue)
> * public to validator full node (or public in the issue)
>

I would expect every network to be impacted by config changes, though only those two would be impacted by StateSync triggered config changes.

libra-node/src/main_node.rs
```diff
@@ -335,6 +346,16 @@ pub fn setup_environment(node_config: &mut NodeConfig) -> LibraHandle {
         // Start the network provider.
         let _listen_addr = network_builder.build();
         debug!("Network started for peer_id: {}", full_node_network.peer_id);
+
```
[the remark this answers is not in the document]
agc-quadra: That means not cutting the corresponding fucntionality out of epoch_manager yet.

network/config/Cargo.toml
agc-quadra: +1.

PR #4035, opened 2020-05-22
[secure-storage] some cleanups

davidiw:
This PR seeks to simplify some of the constructions in secure-storage to be a bit more accessible / less overhead for reviewers.

### DIFF
--- a/secure/storage/src/tests/suite.rs
+++ b/secure/storage/src/tests/suite.rs
@@ -1,7 +1,7 @@
 // Copyright (c) The Libra Core Contributors
 // SPDX-License-Identifier: Apache-2.0
 
-use crate::{Error, Policy, Storage, Value};
+use crate::{Error, Storage, Value};
 use libra_crypto::{ed25519::Ed25519PrivateKey, HashValue, PrivateKey, Signature, Uniform};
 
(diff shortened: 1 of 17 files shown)


secure/storage/src/tests/suite.rs
```diff
@@ -191,8 +191,9 @@ fn test_create_key_pair_and_perform_rotations(storage: &mut dyn Storage) {
         let new_public_key = storage.rotate_key(CRYPTO_NAME).unwrap();
         let new_private_key = storage.export_private_key(CRYPTO_NAME).unwrap();
 
-        let exported_key =
-            storage.export_private_key_for_version(CRYPTO_NAME, public_key).unwrap();
+        let exported_key = storage
```
JoshLind, line 194: I'm surprised the linter didn't catch this?

davidiw: what though?

davidiw:
landing this: https://github.com/libra/libra/pull/4031
which contains both PRs

secure/storage/src/tests/suite.rs
JoshLind: Ooh sorry -- I was confused about why this line changed wrapping? It seemed to be 2 lines before but then changed to 3 😕

PR #4036, opened 2020-05-23
[core modules] Regenerate stdlib and fix invalid event handles, get coverage back up

tzakian:
A previous PR added two new event handles to the `0xA550C18` account address, however the stdlib wasn't regenerated so we didn't catch that the event keys needed to be updated. 

This PR updates the event keys and updates the stdlib. It also adds additional tests to get our coverage back up.


language/move-lang/tests/functional/sliding_nonce/test.move
```diff
@@ -49,3 +50,29 @@ script {
...
+// check: 10003
+
+//! new-transaction
+//! sender: vivian
+script {
+    use 0x0::SlidingNonce;
+
+    fun main() {
+        SlidingNonce::publish_nonce_resource_for_user();
+        SlidingNonce::record_nonce_or_abort(128);
```
tzakian, line 75: @andll Are there additional checks that should be done here with a nonce above 128?

andll: Mm, I am not sure I understand context. Are those tests different from existing tests for bob?

tzakian: This follows a different code path in the sliding nonce module is all (since this is different for nonce >= 128), so I was just checking to see if this was any different from the `bob` checks is all.

andll: Got it. Thanks for adding extra test case, and glad that coverage caught that!

I think most of other >128 test cases are covered in bob tests

Issue #3893, opened 2020-05-14
[Bug] Bytecode verifier does not properly handle mismatched Definition vs Instantiation index


tnowacki:
# 🐛 Bug

- Passing a generic struct definition index to a non-generic instruction (e.g. `Pack`, `Unpack`, etc) can result in a panic instead of an error message

## To reproduce

MVIR snippet
```
module M {
    resource R<T> { f: T }
    t0(): R<bool> {
        return R { f: false };
    }
}
```

PR #4041, opened 2020-05-24
[language] Make verifier aware of generic vs non generic instructions

dariorussi:
This is a fix for issue #3893.
The verifier never checked for generic instructions correctly referring
to generic member and non generic ones referring to non generic members.
This fix it by adding that check in the type safety pass.
It is not clear that is the best pass to check for this property and
maybe a single pass for generic or something similar may be a better
choice. We will find out in time...

It's interesting that calling a generic with the non generic form crashed the verifier, while calling a non generic with the generic form worked just fine!!! Not sure which was more messed up...

### DIFF
--- a/language/bytecode-verifier/bytecode-verifier-tests/src/unit_tests/generic_ops_tests.rs
+++ b/language/bytecode-verifier/bytecode-verifier-tests/src/unit_tests/generic_ops_tests.rs
@@ -1,0 +1,738 @@
+// Copyright (c) The Libra Core Contributors
+// SPDX-License-Identifier: Apache-2.0
+
+use bytecode_verifier::InstructionConsistency;
+use libra_types::vm_error::StatusCode;
+use move_core_types::{account_address::AccountAddress, identifier::Identifier};
+use vm::file_format::*;
+
+// Make a Module with 2 structs and 2 resources with one field each, and 2 functions.
+// One of the struct/resource and one of the function is generic, the other "normal".
+// Also make a test function whose body will be filled by given test cases.
+fn make_module() -> CompiledModuleMut {
+    CompiledModuleMut {
+        module_handles: vec![
+            // only self module
+            ModuleHandle {
+                address: AddressIdentifierIndex(0),
+                name: IdentifierIndex(0),
+            },
+        ],
+        self_module_handle_idx: ModuleHandleIndex(0),
+        identifiers: vec![
+            Identifier::new("M").unwrap(),       // Module name
+            Identifier::new("S").unwrap(),       // Struct name
+            Identifier::new("GS").unwrap(),      // Generic struct name
+            Identifier::new("R").unwrap(),       // Resource name
+            Identifier::new("GR").unwrap(),      // Generic resource name
+            Identifier::new("f").unwrap(),       // Field name
+            Identifier::new("fn").unwrap(),      // Function name
+            Identifier::new("g_fn").unwrap(),    // Generic function name
+            Identifier::new("test_fn").unwrap(), // Test function name
+        ],
+        address_identifiers: vec![
+            AccountAddress::default(), // Module address
+        ],
+        struct_handles: vec![
+            StructHandle {
+                module: ModuleHandleIndex(0),
+                name: IdentifierIndex(1),
+                is_nominal_resource: false,
+                type_parameters: vec![],
+            },
+            StructHandle {
+                module: ModuleHandleIndex(0),
+                name: IdentifierIndex(2),
+                is_nominal_resource: false,
+                type_parameters: vec![Kind::Copyable],
+            },
+            StructHandle {
+                module: ModuleHandleIndex(0),
+                name: IdentifierIndex(3),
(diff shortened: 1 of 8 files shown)


dariorussi:
I have no idea what the coverage tool is asking. There is in fact a difference as you can see, however how to act on that is a mystery.
If I run the coverage on my machine on both master and this change they produce the same result (or so it seems) so I am really not sure.
Given the change I am also not lear why I should even be tagged with understanding why the coverage is changing, anyway.... putting this up and we'll see this week....

tzakian:
Good questions! 
* In this case you don't need to act on it -- it was a previous coverage regression and there is a PR out fixing this (#4036). Since both mater and yours produce the same result there is nothing further you need to worry about coverage wise.
* The reason why a coverage report was generated is because a file under `ir-testsuite` was changed.

vgao1996:
Wouldn't it make more sense to implement this as part of `signature.rs`? (You walk all instructions and error on any misuse of generic/non-generic operations.) IMO `type_safety.rs` is all about checking the validity of the transition of stack state, not about whether the instructions themselves are valid.

dariorussi:
> It seems to me this check is in the wrong pass. I think it does make more sense to implement this as part of `signature.rs`, You enumerate all the instructions and error on any misuse of generic/non-generic operations. If you implement the check as part of the abstract interpreter loop for type safety, you can perform the same check multiple times. IMO `type_safety.rs` is all about checking the validity of the transition of stack state, not about whether the instructions themselves are valid.

I thought about adding it to the signature check and it did not make sense to me because it was a signature check ;) but looking at the code it may be a better place to do it.
I don't think we would perform the check multiple times in the type safety but I'll look into moving it to the signature checker....

dariorussi:
> > It seems to me this check is in the wrong pass. I think it does make more sense to implement this as part of `signature.rs`, You enumerate all the instructions and error on any misuse of generic/non-generic operations. If you implement the check as part of the abstract interpreter loop for type safety, you can perform the same check multiple times. IMO `type_safety.rs` is all about checking the validity of the transition of stack state, not about whether the instructions themselves are valid.
> 
> I thought about adding it to the signature check and it did not make sense to me because it was a signature check ;) but looking at the code it may be a better place to do it.
> I don't think we would perform the check multiple times in the type safety but I'll look into moving it to the signature checker....

hmmm on a second thought I am not sure about the signature checker :)

vgao1996:
Well if you are bothered about the name "signature checker" then I would suggest you make it a separate pass. My point was more about that this check should be in a simple pass that enumerates all the bytecode instructions, rather than an abstract interpreter loop.

tnowacki:
A separate pass seems reasonable. Where do we check arity? i.e. if I have `struct S<T1, T2>` and `S<u64>` or `S<u64, bool, u8>` as invalid instantiations

dariorussi:
I am going for a separate pass for now and then we'll see

tnowacki:
Still, where is arity checked? Kinda confused this isn't checked there as I think about it more and more

dariorussi:
> Still, where is arity checked? Kinda confused this isn't checked there as I think about it more and more

I think signature is where it's checked if I understand your question and some aspects of arity are checked in bounds I think

dariorussi:
> > Still, where is arity checked? Kinda confused this isn't checked there as I think about it more and more
> 
> I think signature is where it's checked if I understand your question and some aspects of arity are checked in bounds I think

actually I think only bounds check

tnowacki:
> actually I think only bounds check

It feels like a really weird thing to do in bounds check. Maybe move all of these things into a single arity check?

dariorussi:
> > actually I think only bounds check
> 
> It feels like a really weird thing to do in bounds check. Maybe move all of these things into a single arity check?

yeah I think it's because we need to do a bound check for type parameter and we made that check for rarity too. 
I'll look into it starting next week though, I don't want to move things around right now, if that is ok with all of you

dariorussi:
I made a new pass called `instruction_consistency` which only has the "generic" checks for now and we'll see later. If there is nothing else to put there we can change the name of move the check in a more proper place.
I want to review all verifier checks once we make the `VerifiedModule` "obsolete".
Next week I will start working on that so I'd be happy to put this in as-is.

This PR still removes the `View` usage from the type checker (it was more work to remove it than to leave it there) so if you hate that let me know and I'll revert that part

language/bytecode-verifier/bytecode-verifier-tests/src/unit_tests/generic_ops_tests.rs
```diff
@@ -0,0 +1,738 @@
+// Copyright (c) The Libra Core Contributors
```
tnowacki, line 1: Nice tests!

FWIW these can be expressed in the IR. But if you prefer these, dealers choice

dariorussi: interesting... I should have known given the test error that was there....
I'll probably leave it as is for now...

PR #4052, opened 2020-05-26
Network/connectivity

agc-quadra:
## Motivation

NOT FOR SUBMISSION

This PR is an effort to consolidate the ConnectivityManager updates into one _UpdateConfiguration_ request as opposed to the current two _UpdateAddresses_ and _UpdateEligiblePeers_.

While this PR largely shows that it is mostly feasible, it highlights a couple of difficulties with the approach:

1).  Currently, different sources update only addresses or only peers.  It is no immediately obvious if this is an accident of implementation or fundamental to the design.
2).  the current configuration objects sometimes contain information about only validators, sometimes only full_nodes, and sometimes both.  Appropriately tracking that context is tricky at best.

*Takeaways:*
0) I do not think this approach is promising
1)  consolidating UpdateAddresses and UpdateEligiblePeers into a single Meta-Update appears to be very challenging and may be undesirable from a logical/safety perspective.
2)   Replacing `UpdateAddresses(peerId, vec<address>)` with `UpdateAddresses(map<peerId, vec<addresses>)` appears to be a very sensible step.

(Write your answer here.)

## Test Plan

## Related PRs

### DIFF
--- a/network/onchain-discovery/src/client.rs
+++ b/network/onchain-discovery/src/client.rs
@@ -338,7 +338,8 @@ where
                                 addrs.push(addr);
                             }
                         }
-
+// TODO:  This is the UpdateAddress that I cannot conver to UpdateConfiguration because the eligible peers are not available.
+                        // It could easily be consolidated into a UpdateGroupAddresses, however.
                         Some(ConnectivityRequest::UpdateAddresses(
                             DiscoverySource::OnChain,
                             peer_id,
--- a/network/onchain-discovery/src/client.rs
+++ b/network/onchain-discovery/src/client.rs
@@ -317,6 +317,7 @@ where
         // advertising new addresses. In an effort to maintain
         // connectivity, we will also merge in the previous advertised
         // addresses.
+        let change_detected = false;
         let update_addr_reqs = latest_discovery_set
             .0
             .into_iter()
--- a/network/src/protocols/discovery/test.rs
+++ b/network/src/protocols/discovery/test.rs
@@ -92,16 +92,14 @@ fn setup_discovery(
     )
 }
 
-async fn expect_address_update(
+async fn expect_configuration_update(
     conn_mgr_reqs_rx: &mut channel::Receiver<ConnectivityRequest>,
-    expected_peer_id: PeerId,
-    expected_addrs: &[NetworkAddress],
+    expected_partner_configuration: PartnerConfiguration,
 ) {
     match conn_mgr_reqs_rx.next().await.unwrap() {
-        ConnectivityRequest::UpdateAddresses(src, peer_id, addrs) => {
+        ConnectivityRequest::UpdateConfiguration(src, partner_configuration) => {
             assert_eq!(DiscoverySource::Gossip, src);
-            assert_eq!(expected_peer_id, peer_id);
-            assert_eq!(expected_addrs, &addrs[..]);
+            assert_eq!(expected_partner_configuration, partner_configuration);
         }
         req => {
             panic!("Unexpected request to connectivity manager: {:?}", req);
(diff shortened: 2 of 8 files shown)


agc-quadra:
@davidiw FYI on consolidating the ConnectivityRequest logic.

network/onchain-discovery/src/client.rs
```diff
@@ -337,7 +338,8 @@ where
                                 addrs.push(addr);
                             }
                         }
-
+// TODO:  This is the UpdateAddress that I cannot conver to UpdateConfiguration because the eligible peers are not available.
```
agc-quadra, line 341: @davidiw 

onchain-discovery is the current blocker on consolidating the interface as it does not appear to have access to the trusted/eligible peer set.

network/src/protocols/discovery/test.rs
```diff
@@ -230,8 +238,18 @@ fn inbound() {
         delivered_rx.await.unwrap();
 
         // Connectivity manager receives address of other peer.
-        expect_address_update(&mut conn_mgr_reqs_rx, new_peer_id, &new_addrs[..]).await;
-        expect_address_update(&mut conn_mgr_reqs_rx, other_peer_id, &other_addrs[..]).await;
+        let updated_configuration = PartnerConfiguration::new(
```
agc-quadra, line 241: There is a danger in simply re-using the existing trusted_peers set here as due to asynchrony we may end up with this update reverting to the old version after anotehr component updates teh eligible peers set concurrently.

@davidiw

davidiw: this is one of my biggest concerns about specifying trusted peers... I'll cc @phlip9 here so that he can track this PR and use it as inspiration for figuring out how he might handle this process -- since it has to be done to properly migrate to a single network identity (ip / port / public key) for peers.

agc-quadra:
> Yeah, there's quite a bit of stuff to wad through here :/
> 
> The only reasonable approach would be to have a hash map (ids) -> hash map (source) -> keys for tested_peers :/

that could work.  but seems very janky and is at odds with the notion that there is a trusted peer set.

versioning by source makes a lot of sense for the addresses.

I think that restricting trusted_peers updates to coming *only* from one source (state sync most likely) could work well.

But this tension very much highlights how the peerID->key and peerID -> networkAddress mappings are two very different things.

network/src/protocols/discovery/test.rs
agc-quadra: (note:  PR is closed) 

this PR is mostly a documentation of combining the two updates and highlighting where the combination becomes difficult/incompatible with the current implementation.

I'll go on record here as believing that the peerID -> Key mapping should be even more explicitly separated from the peerID -> Address mapping.   Taking that a step further, I would hope (?) that there is a signing/verifying component that maintains the peerID -> key mapping completely independent of the network component.  the network component should really only care about the peerID->address mapping and the set of allowed peerIds.

if/when handshakes are needed on connection formation or when evaluating incoming messages, those funcionalities can be invoked on the signing/verification component.

agc-quadra: >since it has to be done to properly migrate to a single network identity (ip / port / public key) for peers.

Note that I'm about 90% convinced that this is a dangerous path to consider.

that said, if there is a *per-network* id and not a unified node  id then my reservations may subside.  But even then I have some very specific concerns related to key rotations/revocations and administrative break glass properties.

To give a hint of those concerns:   if at any point it is decided that a participant is evil (tm), it is appropriate to disable their key with immediate effect with respect to relying on that node to transmit messages in the form of state, mempool proposals, etc.  Given consensus, it is infeasible to disable the key entirely as, in the context of consensus quorum validations, the nodes must discard the key at the same logical time (specifically, nodes which are behind the leading edge may still need to trust the validation of the known to be bad key as part of validating historical decisions.

PR #4053, opened 2020-05-26
abstraction above the file system for the compiler.

RIg410:
Implements the abstraction above the file system for the compiler.
Fixed memory leak.

## Motivation

We use the compiler as a service. For distributed compilation purposes. We are prevented by leak_str and the fact that the compiler uses files as a source. It would be nice to have abstraction above the file system.

yes.

## Test Plan
Unit test.

## Related PRs
--

### DIFF
--- a/language/move-core/types/src/fs.rs
+++ b/language/move-core/types/src/fs.rs
@@ -1,0 +1,535 @@
+// Copyright (c) The Libra Core Contributors
+// SPDX-License-Identifier: Apache-2.0
+
+use serde::{de, export::Formatter, Deserialize, Deserializer, Serialize, Serializer};
+use std::{
+    cell::RefCell,
+    cmp::Ordering,
+    collections::HashMap,
+    ffi::OsString,
+    fmt, fs,
+    fs::{File, OpenOptions},
+    io::{Error, ErrorKind, Read, Write},
+    path::{Path, PathBuf},
+};
+
+thread_local! {
+     static DESCRIPTORS: RefCell<Option<Descriptors>> = RefCell::new(None);
+}
+
+#[derive(Debug)]
+struct Descriptors {
+    fs_counter: usize,
+    paths: Vec<String>,
+}
+
+impl Descriptors {
+    pub fn in_context<A, R>(action: A) -> R
+    where
+        A: Fn(&[String]) -> R,
+    {
+        DESCRIPTORS.with(|descriptors| {
+            descriptors
+                .borrow()
+                .as_ref()
+                .map(|descriptors| action(&descriptors.paths))
+                .expect("Expected AFS in context.")
+        })
+    }
+
+    pub fn in_context_mut<A, R>(action: A) -> R
+    where
+        A: Fn(&mut Vec<String>) -> R,
+    {
+        DESCRIPTORS.with(|descriptors| {
+            descriptors
+                .borrow_mut()
+                .as_mut()
+                .map(|descriptors| action(&mut descriptors.paths))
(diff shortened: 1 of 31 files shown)


language/move-core/types/src/fs.rs
```diff
@@ -0,0 +1,535 @@
...
+    ffi::OsString,
+    fmt, fs,
+    fs::{File, OpenOptions},
+    io::{Error, ErrorKind, Read, Write},
+    path::{Path, PathBuf},
+};
+
+thread_local! {
+     static DESCRIPTORS: RefCell<Option<Descriptors>> = RefCell::new(None);
+}
```
tnowacki, line 18: Before I go too deep into the PR. I'd like to hammer out some details in this file. 

The reason I didn't initially do something like this is that at some point in the future, the Move source language compiler will want to be parallel/multi-threaded. At that time, I was worried any such implementation for interned file strings would fall apart. 

How will this work if the Move source language were to become multithreaded?

language/move-core/types/src/fs.rs
```diff
@@ -0,0 +1,535 @@
+// Copyright (c) The Libra Core Contributors
+// SPDX-License-Identifier: Apache-2.0
+
+use serde::{de, export::Formatter, Deserialize, Deserializer, Serialize, Serializer};
```
runtian-zhou, line 4: I would like to avoid adding this file to `move-core`. The reason is that it sits relatively low in libra's dependency tree, meaning that all crates in `libra` will basically depend on this crate. I would argue that moving this to `move-lang` seems to be a better location.

tnowacki: The source language and the IR use the same spanned information. So maybe in the IR types, `move-ir/types`?

language/move-core/types/src/fs.rs
RIg410: Now the implementation cannot work in a multithreaded environment. But it is quite easy to fix,
we can do something like that.

```
static DESCRIPTORS: Lazy<RwLock<Option<Descriptors>>> = Lazy::new(|| RwLock::new(None));

#[derive(Debug)]
struct Descriptors {
    /// Context counter.
    fs_counter: usize,
    /// Path table.
    paths: Vec<String>,
}
```
The table will be created together with the context at the beginning of compilation and destroyed at the end.
This will avoid memory leaks.
This is the main goal of our PR.
I would like to discuss ways to solve this problem.

tnowacki:
Taking a longer look at this, I don't think this is the right change at this time.

Quick summarized thoughts:
- Currently, the leaking isn't a big issue at the moment as the Move source language compiler is not intended to be hooked into any long lived process. It should only be invoked via a command line tool, and should not be integrated into another Rust service
- Long term, we will need to include some sort of intern table for these strings. That is: the Box::leak solution is an intentional, temporary solution that allows iteration on the compiler to move faster
- In the future, we will move the string table (along with other compiler instance specific information) inside of a context that is threaded through the compiler.
  - There is a lot of work needed to get to that point though, so it isn't likely something that can be done in the next few weeks

I appreciate the PR in that it addresses a known issue in the compiler, but it is not the solution we want to take.
In the future, I think it will be more helpful to first open a GitHub issue before making a PR. It will give you a better sense of how we are thinking about addressing these problems, and let us figure out the best solution going forward.
Nonetheless, I really do appreciate the effort put in here! These changes aren't quick or easy to do, and I love seeing external contributions for the source language. But right now, we are heavily focused on getting this first version fully working, so while I do love seeing them, our bandwidth might be limited for dealing with external contributions.

PR #4059, opened 2020-05-26
[move core modules] add_all_currencies parameter for parent/child vas…

sblackshear:
…p creation

One remaining inconvenience of parent and child VASP account creation is the need to send separate transactions to add new currencies to an account. This PR solves the problem by adding a add_all_currencies parameter to both the parent and child VASP creation scripts. When true, this initializes the freshly created account with all of the available currencies in the system.

VASPs that want to accept a subset of the currencies will still need to send an initial creation transaction followed by a few AddCurrency transactions, but that seems reasonable (and preferable to having a different script for every combination of currencies).

## Motivation

(Write your answer here.)

## Test Plan

## Related PRs


language/stdlib/modules/libra_account.move
```diff
@@ -573,7 +577,18 @@ module LibraAccount {
...
+            if (!::exists<Balance<Coin1::T>>(new_account_addr)) {
+                add_currency<Coin1::T>(&new_account);
+            };
+            if (!::exists<Balance<Coin2::T>>(new_account_addr)) {
+                add_currency<Coin2::T>(&new_account);
+            };
+            if (!::exists<Balance<LBR::T>>(new_account_addr)) {
+                add_currency<LBR::T>(&new_account);
+            };
+        };
```
tzakian, line 591: Overall this looks god, and I'm fine with this going in as-is. 

One higher-level question I had though was whether we want this logic in module code, or to instead have this be in the transaction script code (i.e. `create_account(...)` and then have this logic within the script). What do you think?

sblackshear: I think having it in the script could be good, but I don't see how to do it. The problem is that we need a `signer` to call `add_currency` on, but the only `signer` available in the script is the tx sender (not the `signer of the created account, which we have here). Do you have a scheme for this?

tzakian: ah! I didn't see this. It's doing a `move_to` into the newly created account (still wrapping my head around this `&signer` stuff...). This all makes a lot more sense now.

I can't think of any other way around doing this unless we allow `create_signer` to be a public function, which isn't a good idea.

PR #4060, opened 2020-05-26
[language] Remove `TransactionMetadata` from Move VM into the Libra A…

dariorussi:
…dapter

`TransactionMetadata` was not really needed by the Move VM once we
removed all the instructions that would return transaction info.
We removed it from `move-vm` into `libra-vm` and we may look to
remove it entirely.
The Move VM takes a sender and a transaction size now.


dariorussi:
@runtian-zhou please take a look at the libra_vm and let me know your thoughts.
Also if we should remove `TransactionMtadata` altogether and some point soon

language/libra-vm/src/libra_vm.rs
```diff
@@ -336,16 +336,18 @@ impl LibraVM {
...
                 let ret = self.move_vm.execute_script(
                     s,
                     ty_args,
                     args,
-                    &mut cost_strategy,
+                    txn_data.sender(),
+                    txn_data.transaction_size(),
                     &mut data_store,
-                    txn_data,
+                    &mut cost_strategy,
```
tzakian, line 350: Removing the idea of "transaction" from the interpreter it awesome!

I was wondering your thoughts on instead of passing the transaction size to the entrypoint, changing the `CostStrategy` to hold this?

To explain a bit more, `CostStrategy::transaction(...)` would take the `TransactionMetadata` and would have an additional field `min_gas_charge` in it that is calculated from the transaction size. The `CostStrategy::system(...)` would just set this to `0`.

Then you pass only the `CostStrategy` to the `Interpreter::entrypoint` and `charge_intrinsic_gas` would no longer take the transaction size, but use this stored `min_gas_charge` within it.

dariorussi: yes that is in line with the talk we had and I'd like to do it. But I was thinking in another PR so I can possibly come up with alternative solutions and we can see how we like it.
Thoughts?

tzakian: ah! For some reason I was thinking this was that PR....

Carry on! We can take a look at it in the upcoming PR

language/move-vm/runtime/src/interpreter.rs
```diff
@@ -54,53 +53,47 @@ macro_rules! debug_writeln {
...
     pub(crate) fn entrypoint(
-        data_store: &mut dyn DataStore,
-        loader: &Loader,
-        txn_data: &'txn TransactionMetadata,
-        cost_strategy: &mut CostStrategy,
         function: Arc<Function>,
         ty_args: Vec<Type>,
         args: Vec<Value>,
+        sender: AccountAddress,
+        txn_size: AbstractMemorySize<GasCarrier>,
```
runtian-zhou, line 73: Do we want to have the charging logic on caller side or callee side? IMHO we can do it over the caller side as we really need to charge the gas only when running the script body. But it's fine if we decide to do it in a separate PR.

dariorussi: the idea is to hav a separate PR where we do that only and see the alternative.
I think there are a couple of options and I think @tzakian may like one and @runtian-zhou (and possibly me ;)) another.
PR (or multiple ones) coming soon

PR #4063, opened 2020-05-26
[move-prover] Add support for signer and move_to

shazqadeer:
## Motivation

This PR adds support for the primitive signer  type and for the new bytecode instruction move_to.

Yes.

## Test Plan

Existing tests and another test.

## Related PRs

None.

### DIFF
--- a/language/move-prover/src/bytecode_translator.rs
+++ b/language/move-prover/src/bytecode_translator.rs
@@ -1181,15 +1181,29 @@ impl<'env> ModuleTranslator<'env> {
                         );
                         emitln!(self.writer, &update_and_track_local(dest, "$tmp"));
                     }
+                    MoveTo(mid, sid, type_actuals) => {
+                        let value = srcs[0];
+                        let signer = srcs[1];
+                        let resource_type =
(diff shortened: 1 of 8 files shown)


language/move-prover/src/bytecode_translator.rs
```diff
@@ -1181,15 +1181,29 @@ impl<'env> ModuleTranslator<'env> {
...
+                        let value = srcs[0];
+                        let signer = srcs[1];
+                        let resource_type =
+                            boogie_struct_type_value(self.module_env.env, *mid, *sid, type_actuals);
+                        emitln!(
+                            self.writer,
+                            "call $MoveTo({}, {}, {});",
+                            resource_type,
+                            str_local(value),
+                            str_local(signer),
```
sblackshear, line 1194: In the source language syntax, `signer` comes first (e.g., `move_to(account, some_resource)`). Fine if the prover always uses the stack order like this everywhere (i.e., reverse of the source syntax), but thought I would point out just in case...

shazqadeer: I tried to be consistent with the convention used in the rest of the translator for instructions with multiple arguments.

PR #4064, opened 2020-05-26
[terraform] Make Vault data persistence configurable

mgorven:
Make Vault data non-persistent if validator data is too.

### DIFF
--- a/terraform/vault.tf
+++ b/terraform/vault.tf
@@ -86,7 +86,7 @@ resource "aws_ecs_task_definition" "vault" {
 
   volume {
     name      = "vault-data"
-    host_path = "/vault"
+    host_path = var.persist_libra_data ? "/vault" : ""
   }
 
   placement_constraints {


rustielin:
What happens when Vault restarts, causing the keys to change?

mgorven:
> What happens when Vault restarts, causing the keys to change?

Dunno, hopefully it doesn't crash. But if the validators all start from scratch on a push we need Vault to do so too.

mgorven:
> Looks good to me. Quick q though: would it be possible to restore vault data? e.g. If the vault container crashes and restarts, we still have its old volume until ECS decides to clean it up. Could we just point vault to its old volume?

Theoretically yes. Why do you think this is necessary though? If the testnet is running without persistence we're assumedly okay with it losing data and restart arbitrarily.

rustielin:
> Theoretically yes. Why do you think this is necessary though? If the testnet is running without persistence we're assumedly okay with it losing data and restart arbitrarily.

Nonpersistent storage was added to improve devnet liveness, but if vault restarts and the validators themselves don't, then the network is stuck. Agreed we don't care about data loss and arbitrary restarts in general, but this will halt the entire network until at least 1 hr later until continuous push kicks in. Not too big of an issue overall though; just means we'll some manual work in the edge cases.

PR #4069, opened 2020-05-27
[core modules] Add on-chain VASP registry

tzakian:
This stores the list of parent VASP addresses on-chain. Note that some of these addresses may no longer be valid due to expiration of the parent VASP credential. This will need to be checked off-chain for each of these addresses. 

## Tests
Adds additional tests to make sure that the registry is updated when creating or decertifying a VASP.


tzakian:
Closing as this is no longer needed.

PR #4071, opened 2020-05-27
[move core modules] remove use of Transaction::sender in association …

sblackshear:
…module

## Motivation

Working on deprecating `Transaction::sender()` and `move_to_sender`

Y

## Test Plan

Existing tests pass.


language/move-vm/runtime/src/data_operations.rs
```diff
@@ -43,11 +43,14 @@ pub(crate) fn move_resource_to(
...
-            Location::new(),
-            StatusCode::CANNOT_WRITE_EXISTING_RESOURCE,
-        ))
+        warn!(
+            "[VM] Cannot write over existing resource type {:?} access path {}",
+            ty, ap
+        );
+        Err(
+            vm_error(Location::new(), StatusCode::CANNOT_WRITE_EXISTING_RESOURCE)
+                .with_sub_status(sub_status::DRE_GLOBAL_ALREADY_BORROWED),
```
tzakian, line 52: This might be right, but this sub status seems incorrect to me? It signifies that the resource was already borrowed, whereas this can happen even if the resource hasn't been borrowed but just exists in storage?

sblackshear: My bad, this change was just for debugging. Will revert.

As an aside, I noticed a bunch of these dynamic reference error codes that are no longer relevant--we should get rid of them.

PR #4072, opened 2020-05-27
[management] Support verification of node properties at least....

davidiw:
verify now ensure that
    1) operator account host validator credentials
    2) that those validator credentials are what you thought they would be
    3) that the computed waypoint makes sense

    Todo:
    1) check that the authenticator for operator account is correct
    2) check that the authenticator for owner account is correct

I also noticed that genesis wasn't actually storing the contents to the disk... If this looks good, I'll amend the readme and land those together.

This is an unfortunately ugly PR, but it does a lot of manual not easy to unify work...

This contains a series of commits that are irrelevant and currently landing, will update later.

### DIFF
--- a/config/management/src/lib.rs
+++ b/config/management/src/lib.rs
@@ -249,7 +249,10 @@ pub mod tests {
     use super::*;
     use crate::storage_helper::StorageHelper;
     use libra_types::account_address::AccountAddress;
-    use std::{fs::File, io::Write};
+    use std::{
+        fs::File,
+        io::{Read, Write},
+    };
 
     #[test]
     fn test_end_to_end() {
@@ -318,7 +321,14 @@ pub mod tests {
 
         // Step 4) Produce genesis
 
-        helper.genesis().unwrap();
+        let genesis_path = libra_temppath::TempPath::new();
+        genesis_path.create_as_file().unwrap();
+        helper.genesis(genesis_path.path()).unwrap();
+        let mut file = File::open(genesis_path.path()).unwrap();
+        let mut contents = Vec::new();
+        assert!(contents.is_empty());
+        file.read_to_end(&mut contents).unwrap();
+        assert!(!contents.is_empty());
     }
 
     #[test]
@@ -385,12 +395,20 @@ pub mod tests {
         let helper = StorageHelper::new();
         let namespace = "verify";
 
-        let output = helper.verify(namespace).unwrap().split("KeyNotSet").count();
+        let output = helper
+            .verify(namespace)
+            .unwrap()
+            .split("Key not set")
+            .count();
         assert_eq!(output, 10); // 9 KeyNotSet results in 9 splits
 
         helper.initialize(namespace.into());
 
-        let output = helper.verify(namespace).unwrap().split("KeyNotSet").count();
+        let output = helper
+            .verify(namespace)
+            .unwrap()
+            .split("Key not set")
+            .count();
         assert_eq!(output, 1); // 0 KeyNotSet results in 1 split
     }
 
--- a/config/management/src/smoke_test.rs
+++ b/config/management/src/smoke_test.rs
@@ -113,9 +113,11 @@ fn smoke_test() {
     }
 
     // Step 4) Produce genesis and introduce into node configs
-    let genesis = helper.genesis().unwrap();
+    let genesis_path = TempPath::new();
+    genesis_path.create_as_file().unwrap();
+    let genesis = helper.genesis(genesis_path.path()).unwrap();
 
-    // Step 6) Introduce waypoint and genesis into the configs
+    // Step 5) Introduce waypoint and genesis into the configs and verify along the way
     let temppath = TempPath::new();
     temppath.create_as_dir().unwrap();
     let swarm_path = temppath.path().to_path_buf();
@@ -123,6 +125,9 @@ fn smoke_test() {
     for (i, mut config) in configs.iter_mut().enumerate() {
         let ns = i.to_string();
         let waypoint = helper.create_waypoint(&ns).unwrap();
+        let output = helper.verify_genesis(&ns, genesis_path.path()).unwrap();
+        // 4 matches = 6 splits
+        assert_eq!(output.split("match").count(), 5);
 
         let mut to = swarm_path.clone();
         to.push("secure_store_for_".to_string() + &ns);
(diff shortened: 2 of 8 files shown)


config/management/src/verify.rs
```diff
@@ -110,7 +146,134 @@ fn write_waypoint(storage: &dyn Storage, buffer: &mut String, key: &str) {
...
+    write_assert(
+        buffer,
+        libra_global_constants::WAYPOINT,
+        actual_waypoint == expected_waypoint,
+    );
+
+    Ok(())
+}
+
+fn compute_genesis(
```
JoshLind, line 207: nit: this method looks more like it computes a waypoint and just reads genesis from the file? perhaps rename to "compute_waypoint"?

config/management/src/verify.rs
```diff
@@ -110,7 +146,134 @@ fn write_waypoint(storage: &dyn Storage, buffer: &mut String, key: &str) {
...
+        .map_err(|e| Error::LocalStorageReadError(VALIDATOR_NETWORK_KEY, e))?;
+    let expected_validator_key = validator_config.validator_network_identity_public_key;
+    write_assert(
+        buffer,
+        VALIDATOR_NETWORK_KEY,
+        actual_validator_key == expected_validator_key,
+    );
+
+    let actual_fullnode_key = x25519_from_storage(FULLNODE_NETWORK_KEY, storage)
+        .map_err(|e| Error::LocalStorageReadError(VALIDATOR_NETWORK_KEY, e))?;
```
JoshLind, line 183: copy paste error: VALIDATOR_NETWORK_KEY --> FULLNODE_NETWORK_KEY

config/management/src/verify.rs
```diff
@@ -110,7 +146,134 @@ fn write_waypoint(storage: &dyn Storage, buffer: &mut String, key: &str) {
...
+    let expected_fullnode_key = validator_config.full_node_network_identity_public_key;
+    write_assert(
+        buffer,
+        FULLNODE_NETWORK_KEY,
+        actual_fullnode_key == expected_fullnode_key,
+    );
+
+    let actual_waypoint = storage
```
JoshLind, line 191: nit: I'd be tempted to compare the waypoint first in this method as it means that "expected_waypoint" is used closer to where. This comparison also doesn't rely on validator_config

config/management/src/verify.rs
```diff
@@ -74,17 +91,36 @@ impl Verify {
...
     }
 }
 
-fn write_key(storage: &dyn Storage, buffer: &mut String, key: &str) {
-    let value = storage
-        .get_public_key(key)
-        .map(|c| c.public_key.to_string())
-        .unwrap_or_else(|e| format!("{:?}", e));
+fn write_assert(buffer: &mut String, name: &str, value: bool) {
+    let value = if value { "match" } else { "MISMATCH" };
```
JoshLind, line 105: nit: caps vs. no caps? Although, it seems like this probably more an intentional decision to highlight the error?

Some future food for thought: if we do end up comparing/verifying a lot of information, it might make sense to print some final result that is an aggregation of all assertions. Especially if we won't throw an error (which is really obvious to see).. this will make it less likely for human error when someone is trying to make sure state is as it should be (e.g., they don't miss a MISMATCH).

config/management/src/lib.rs
```diff
@@ -318,7 +318,9 @@ pub mod tests {
 
         // Step 4) Produce genesis
 
-        helper.genesis().unwrap();
+        let genesis_path = libra_temppath::TempPath::new();
+        genesis_path.create_as_file().unwrap();
+        helper.genesis(genesis_path.path()).unwrap();
```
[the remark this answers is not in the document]
davidiw: Hmm... let me think about it... there are other tests that verify this so at least we cannot have a regression, but that's some massive end to end test... not ideal

config/management/src/smoke_test.rs
```diff
@@ -113,16 +113,21 @@ fn smoke_test() {
     }
 
     // Step 4) Produce genesis and introduce into node configs
-    let genesis = helper.genesis().unwrap();
+    let genesis_path = TempPath::new();
+    genesis_path.create_as_file().unwrap();
+    let genesis = helper.genesis(genesis_path.path()).unwrap();
```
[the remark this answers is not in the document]
davidiw: We actually verify it on line 128

config/management/src/verify.rs
davidiw: Intentional as you can imagine. Yeah, this is a fair point, as we get feedback, I'm definitely going to iterate. It might also be worth pulling in termcolor or what not to paint green versus red.

config/management/src/verify.rs
```diff
@@ -74,17 +91,36 @@ impl Verify {
...
+    writeln!(buffer, "{} - {}", name, value).unwrap();
+}
+
+fn write_break(buffer: &mut String) {
```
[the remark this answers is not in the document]
davidiw: I only did it because this ended up being 5 lines for what should be a single line :P

config/management/src/verify.rs
davidiw: Yup, sigh...

config/management/src/verify.rs
davidiw: Interesting point....

config/management/src/verify.rs
davidiw: The DbReaderWriter contains the computation of genesis, I was wondering if I should leave comments, this indicates I should :)

config/management/src/verify.rs
```diff
@@ -110,7 +146,134 @@ fn write_waypoint(storage: &dyn Storage, buffer: &mut String, key: &str) {
...
+        .map_err(|e| Error::UnexpectedError(format!("Unable to parse genesis: {}", e)))?;
+
+    let waypoint = db_bootstrapper::bootstrap_db_if_empty::<LibraVM>(&db_rw, &genesis)
+        .map_err(|e| Error::UnexpectedError(e.to_string()))?
+        .ok_or_else(|| Error::UnexpectedError("Unable to generate a waypoint".to_string()))?;
+
+    Ok((db_rw, waypoint))
+}
+
+fn validator_config(
```
davidiw, line 230: Like this blob here is pretty common, but maybe still... comments...

PR #4073, opened 2020-05-27
[crypto] Signing based on LCS prefix, with less Sha3

huitseeker:
## *What this does*
TL;DR
Now, for a large class of structs, you get to write
```
        let hashable = MyNewStruct::new(x);
        let signature = private_key.sign(&hashable).expect("that should work");
```
and the same for verification. No `HashValue` involved.

This offers a new signing+verification api where the message, which formerly had to be a `sha3` `HashValue`, can now be any struct fulfilling a few trait bounds, namely:
- `libra_crypto::hash::CryptoHash`, ensuring there's a unique serialization prefix for that type,
- `serde::Serialize`, now ensuring there's a canonical byte representation for the object.

The signing and verification builds on these two items, but does *not* hash them, and rather passes the "right" bytes to dalek (which then Sha512-hashes them internally).

This builds on the *awesome* work in LCS +  `lcs::trace_name` tooling (see #3954 and #3845) that did not exist at the time the `HashValue` machinery was written (not to mention the `CryptoHasher` stuff). 

## *Why this is better*

No double hashing in signing / verification, massively more structs are signable / verifiable without writing a proxy method (e.g. see `RawTransaction::sign`) to hash & pass them off to signing. 

## *Why this is worse*
- breaking change, 

## *Tests & TODO*
- see small included unit tests.
  + `test_cryptohasher_salt_access`
  + `test_signature_verification_from_struct`
- we should continue guaranteeing that the only value that can pass a signature verification is the one that was used to produce it. There are still edge cases we have flagged (e.g. generics collapsing under the same prefix, versioning ..), though the current setup is a strict improvement on the extant.
- we can trigger both the derivation of `CryptoHasher` and `CryptoHash` under a single `derive` annotation. Do we still want to require `(CryptoHasher, LCSCryptoHash)` as we do now?
- ~~considering we don't exactly have a lot of distinct `Hasher` implementations, and that they're all macro-generated, do we still want a distinct machinery for `CryptoHash` and `CryptoHasher`, or can we collapse both under the same trait?~~ (see @matbd below: `CryptoHasher` has other uses)
- we should remove proxy methods for signing (see e.g. above),
- do we still need the manual `hash` macros?
- is `sign_arbitrary_message_for_testing` still relevant / useful? Or can we `NewType` the bytes that Move would want to sign for TR, thereby creating a trackable serialization name & opening access to the `sign(my_new_type)` api ?


crypto/crypto-derive/src/lib.rs
```diff
@@ -368,6 +368,11 @@ pub fn hasher_dispatch(input: TokenStream) -> TokenStream {
         }
 
         impl libra_crypto::hash::CryptoHasher for #hasher_name {
+            fn salt() -> std::borrow::Cow<'static, str> {
+                std::borrow::Cow::from(libra_crypto::_serde_name::trace_name::<#type_name #param>()
```
kchalkias, line 372: Where do we apply hashing in salts? This is required to ensure that no salt is the prefix of another salt.

crypto/crypto/src/hash.rs
```diff
@@ -442,6 +442,10 @@ pub trait CryptoHash {
 
 /// A trait for representing the state of a cryptographic hasher.
 pub trait CryptoHasher: Default + std::io::Write {
```
kchalkias, line 444: Not sure if w/o actual hashing happening, this should be still called a Hasher or CryptoSerializer or TypedSerializer.

crypto/crypto-derive/src/lib.rs
huitseeker: Yeah, we definitely need to enforce this in one way or another.

@matbd is this a guarantee we can enforce with tooling?

@sblackshear @davidiw How does the complexity increment from "pass these salt bytes and object bytes to EdDSA signing" to "pass these *hashed* bytes and object bytes to EdDSA signing" fare with our specification efforts?

crypto/crypto/src/hash.rs
huitseeker: The prime goal for `CryptoHasher` is to provide the `update`, `finish` functions of a real honest-to-god hasher, which is invoked when you call `<T as CryptoHash>.hash(self)`.

The fact I'm piggy-backing on it to add non-hashed functionality doesn't mean the structure's hashing purposes get invalidated :)

kchalkias: That’s true, but if we never sign hashes and this is only used for type-defined serialization, I don’t see why they should be called hashers vs typedLCS (and then we can use it for other purposes too where we specifically need LCS Serialization with type (domain separation) attached )

crypto/crypto/src/ed25519.rs
```diff
@@ -214,6 +218,21 @@ impl SigningKey for Ed25519PrivateKey {
     type VerifyingKeyMaterial = Ed25519PublicKey;
     type SignatureMaterial = Ed25519Signature;
 
+    fn sign<T: CryptoHash + Serialize>(
+        &self,
+        message: &T,
+    ) -> Result<Ed25519Signature, CryptoMaterialError> {
+        let mut bytes = Vec::<u8>::new();
+        bytes.extend_from_slice(&<T::Hasher as CryptoHasher>::salt().as_bytes());
```
valerini, line 226: The salt is an un-hashed type name? Shouldn't it be hashed here?

valerini: I guess, we want salts to be pre-hashed already, so we don't hash them over and over again...

BTW, if we had a suffix based separator (like the one we had before - something like $$$###LIBRA###$$$ that will clearly not be part of any type name) we might not need hashing of salts, they will be guaranteed to not be prefixes of each other. But we will loose the property of salts being of the same length in case we need it.

crypto/crypto-derive/src/lib.rs
matbd: Currently, the *salt* is defined as `format!("{}::{}", LIBRA_HASH_PREFIX, serde_name)`.
We seed the hashing using `sha3_256(salt)` to avoid worrying about prefix-collisions and offer a simplified programmer experience. (*seed* = obligatory first bytes to hash before any LCS payload).

@huitseeker I suspect you want to expose the seed here. The return type is likely to be `&'static [u8]` as we probably want to memoize the seed.

crypto/crypto/src/hash.rs
matbd: `CryptoHasher` is usable (and used) independently from LCS.

crypto/crypto/src/ed25519.rs
matbd: See my previous comment: I believe we should inject a fixed-length (i.e. hashed) seed here instead of the salt.

crypto/crypto/src/ed25519.rs
```diff
@@ -214,6 +218,21 @@ impl SigningKey for Ed25519PrivateKey {
...
+        let serialized_material =
+            lcs::to_bytes(&message).map_err(|_| CryptoMaterialError::SerializationError)?;
```
matbd, line 228: Please avoid memory allocations by using `lcs::serialize_into(..)` instead of `lcs:to_bytes(..)`.

crypto/crypto/src/multi_ed25519.rs
```diff
@@ -472,6 +493,19 @@ impl Signature for MultiEd25519Signature {
...
 
+    fn verify_struct_msg<T: CryptoHash + Serialize>(
+        &self,
+        message: &T,
+        public_key: &MultiEd25519PublicKey,
+    ) -> Result<()> {
+        let mut bytes = Vec::<u8>::new();
+        bytes.extend_from_slice(&<T as CryptoHash>::Hasher::salt().as_bytes());
+        let serialized_material =
+            lcs::to_bytes(&message).map_err(|_| CryptoMaterialError::SerializationError)?;
```
matbd, line 504: lcs::serialize_into

crypto/crypto/src/hash.rs
matbd: We could create a smaller trait (say `CryptoSeed`) but I'm not sure if it's worth the trouble.

crypto/crypto/src/ed25519.rs
matbd: The suffix of the salt is now a fixed prefix. It was never important for security as we were hashing the salt anyway (partly for performance reasons).

davidiw: on an unrelated note, this seems like we're creating to Vec<u8>

why not make it

```
let mut bytes = <T::Hasher as CryptoHasher>::salt().as_bytes();
```
or maybe appending an additional to_vec after that?

andll:
That's a bit large PR to review.
Right now we use signing like this:
```
        let transaction: RawTransaction;
        let key_pair: KeyPair<Ed25519PrivateKey, Ed25519PublicKey> = key.into();
        key_pair
            .sign_txn(transaction);
```
This PR would not change this code right? (If yes, then we are ok :D )

huitseeker:
A revision to expose a sha3-ed seed derived from the agreed upon salt is forthcoming, no need to spend too much time on review RN.

crypto/crypto/src/ed25519.rs
huitseeker: As argued (with a ton of sense) by @matbd, this now injects a 32-byte sha3-seed. And I added @davidiw 's `to_vec()`.

crypto/crypto-derive/src/lib.rs
huitseeker: Hashing of salts is [there](https://github.com/libra/libra/pull/4073/files#diff-7d2098dbb70aee9e033b266567340498R382-R392) and [there](https://github.com/libra/libra/pull/4073/files#diff-040d5f23675cd0d3adaa822575b670edR523-R529).

crypto/crypto-derive/src/lib.rs
```diff
@@ -368,6 +375,24 @@ pub fn hasher_dispatch(input: TokenStream) -> TokenStream {
         }
 
         impl libra_crypto::hash::CryptoHasher for #hasher_name {
+            fn seed() -> &'static [u8; 32] {
+                if let Some(seed_value) = #static_seed_name.get() {
```
matbd, line 379: Why not `get_or_init`?
https://docs.rs/once_cell/1.4.0/once_cell/sync/struct.OnceCell.html#method.get_or_init

crypto/crypto-derive/src/lib.rs
```diff
@@ -368,6 +375,24 @@ pub fn hasher_dispatch(input: TokenStream) -> TokenStream {
...
+                } else {
+                    use libra_crypto::_sha3::Hasher;
+
+                    let name = libra_crypto::_serde_name::trace_name::<#type_name #param>()
+                        .expect("The `CryptoHasher` macro only applies to structs and enums.").as_bytes();
+                    let salt: Vec<u8> = [libra_crypto::hash::LIBRA_HASH_PREFIX, name].concat();
+                    let mut sha3 = libra_crypto::_sha3::Sha3::v256();
+                    sha3.update(&salt);
+                    let mut output = [0u8; 32];
+                    sha3.finalize(&mut output);
```
matbd, line 390: This code is now duplicated in two places: here and in `DefaultHasher::new()`. I feel that we could avoid this and keep LIBRA_HASH_PREFIX `pub(crate)` simply by creating a utility function in `DefaultHasher`.

crypto/crypto-derive/src/lib.rs
huitseeker: :facepalm: ofc

crypto/crypto/src/hash.rs
```diff
@@ -103,15 +103,21 @@ use anyhow::{ensure, Error, Result};
...
+use once_cell::sync::{Lazy, OnceCell};
 #[cfg(any(test, feature = "fuzzing"))]
 use proptest_derive::Arbitrary;
 use rand::{rngs::OsRng, Rng};
 use serde::{de, ser};
 use std::{self, convert::AsRef, fmt, str::FromStr};
 use tiny_keccak::{Hasher, Sha3};
 
-pub(crate) const LIBRA_HASH_PREFIX: &[u8] = b"LIBRA::";
+/// A prefix used to begin the salt of every libra hashable structure. The salt
```
matbd, line 114: IHMO long explanations like this get out of sync too easily and are better grouped in one place.

crypto/crypto/src/hash.rs
```diff
@@ -506,6 +516,20 @@ macro_rules! define_hasher {
         }
 
         impl CryptoHasher for $hasher_type {
+            fn seed() -> &'static [u8;32] {
+                if let Some(seed_value) = $seed_name.get() {
```
matbd: `get_or_init`

crypto/crypto/src/hash.rs
```diff
@@ -506,6 +516,20 @@ macro_rules! define_hasher {
...
+                    seed_value
+                } else {
+                    let salt: Vec<u8> = [LIBRA_HASH_PREFIX, $salt].concat();
```
matbd, line 521: Although we are in the same file this time, I would still call into `Default::make_seed` or something like that.

crypto/crypto/src/lib.rs
```diff
@@ -22,8 +22,10 @@ mod unit_tests;
...
 pub use hash::HashValue;
 
-// Reexport once_cell and serde_name for use in CryptoHasher Derive implementation.
+// Reexport sha3, once_cell and serde_name for use in CryptoHasher Derive implementation.
 #[doc(hidden)]
 pub use once_cell as _once_cell;
 #[doc(hidden)]
 pub use serde_name as _serde_name;
+#[doc(hidden)]
+pub use tiny_keccak as _sha3;
```
matbd, line 31: looks like we could also avoid this?

matbd:
@huitseeker Thanks for the changes. I left a few style comments. (Sorry for the spam, I realize I should have grouped them by "starting a review")

crypto/crypto-derive/src/lib.rs
huitseeker: So, it's an excellent suggestion, and I have created a `DefaultHasher::prefixed_hash`, rather than touch `DefaultHasher::new()`.

I do not understand `DefaultHasher::new`, think it is seeding with a double sha-3 of the salt, and is therefore buggy as hell. @kchalkias @valerini @weilei0107 feel free to enlighten me.

matbd: Good catch! It totally looks like a bug due to the confusing name `HashValue::from_sha3_256`.

huitseeker: Thanks for your offline explanation, it seems the `DefaultHasher::new(typename)` is indeed OK, though cryptic.
https://github.com/libra/libra/blob/97c458e54545e01a7f04ee28ea27164368ea20ce/crypto/crypto/src/hash.rs#L484-L488

We start with 
`HashValue::from_sha3_256(&salt[..]) == seed`
then `state.update(seed)`

I think I'll tack on a couple additional commits that rename `from_sha3_256` and and just rephrase the existing `DefaultHasher::new()` in a hopefully more readable fashion.

huitseeker: Done! @davidiw I'd love your opinion on whether the last couple of commits in this PR are an improvement.

huitseeker:
Should we land this relatively soon, we can then start eradicating `private_key.sign_message(t.hash())` and replacing it with `private_key.sign(t)` wherever possible! (and same thing for verify of course).

crypto/crypto/Cargo.toml
```diff
@@ -52,7 +52,7 @@ serde_json = "1.0.53"
 default = ["std", "fiat_u64_backend"]
 assert-private-keys-not-cloneable = ["static_assertions"]
 cloneable-private-keys = []
-fuzzing = ["proptest", "proptest-derive", "cloneable-private-keys"]
+fuzzing = ["proptest", "proptest-derive", "cloneable-private-keys",]
```
matbd, line 55: remove?

crypto/crypto/src/hash.rs
```diff
@@ -457,6 +463,16 @@ pub struct DefaultHasher {
 }
 
 impl DefaultHasher {
+    #[doc(hidden)]
+    pub fn prefixed_hash(buffer: &[u8]) -> [u8; 32] {
```
matbd: FWIW I preferred a name like `make_seed` (because we are not returning a `HashValue` neither is the returned value useful in any other way).

crypto/crypto/src/ed25519.rs
```diff
@@ -214,6 +218,19 @@ impl SigningKey for Ed25519PrivateKey {
     type VerifyingKeyMaterial = Ed25519PublicKey;
     type SignatureMaterial = Ed25519Signature;
 
+    fn sign<T: CryptoHash + Serialize>(
+        &self,
+        message: &T,
+    ) -> Result<Ed25519Signature, CryptoMaterialError> {
+        let mut bytes = <T::Hasher as CryptoHasher>::seed().to_vec();
+        lcs::serialize_into(&mut bytes, &message)
+            .map_err(|_| CryptoMaterialError::SerializationError)?;
```
matbd, line 227: FYI in the rest of the codebase, we assume that LCS serialization into a vector never fails. (From the back of my mind, the only way it would fail is because of a non-supported type such as `char`, `f32`, `f64` or a clowny implementation of a `Serialize` trait for some user type.)

crypto/crypto/src/traits.rs
```diff
@@ -24,6 +24,8 @@ use thiserror::Error;
 #[derive(Clone, Debug, PartialEq, Eq, Error)]
 #[error("{:?}", self)]
 pub enum CryptoMaterialError {
+    /// Struct to be signed does not serialize correctly.
```
matbd, line 27: This should never happen.

aching:
@huitseeker, if this is a breaking change, can you please add the "breaking" label to this commit?

crypto/crypto/src/traits.rs
kchalkias: It can still happen if for example we try to serialize an array of 2^31+ elements. Let’s revisit that in an upcoming PR.

crypto/crypto/src/ed25519.rs
```diff
@@ -214,6 +218,19 @@ impl SigningKey for Ed25519PrivateKey {
     type VerifyingKeyMaterial = Ed25519PublicKey;
     type SignatureMaterial = Ed25519Signature;
 
+    fn sign<T: CryptoHash + Serialize>(
```
kchalkias, line 221: @huitseeker what happens if one invokes this accidentally with a bytes, actually Vec object, after they serialized the original struct.
It should be clear to devs that they don’t serialize the struct outside and this function takes as input the original object type. Maybe add some docs clarifying that.

huitseeker:
@aching It's not breaking. Strict addition in functionality.

crypto/crypto/src/hash.rs
huitseeker: The fact that this does not return a `HashValue` is inferrable from the type, but I'll add a docstring.

crypto/crypto/src/ed25519.rs
huitseeker: I'll let you and @kchalkias figure it out in a forthcoming PR, but I think that an arcane failure of `Serialization`, if it manifests as a signature failure, is sufficiently far from its root cause to warrant reporting on the error precisely, here.

crypto/crypto/src/traits.rs
huitseeker: See https://github.com/libra/libra/pull/4073#discussion_r433229305

crypto/crypto/src/ed25519.rs
huitseeker: The docstring is on the trait.

huitseeker:
@matbd @kchalkias This should be ready to go. As mentioned [above](https://github.com/libra/libra/pull/4073#issuecomment-636243078) I'll start working on a PR that changes all possible signing / verification behavior (`HashValue`-based) in favor of the new signing (`LCS`-based) asap, so that the sting of the breaking changes happens only once.

wrwg:
> Move prover errors are spurious here (changes in language are precisely a method renaming, and applied to an internal methods that's not exposed to the IR at that). /cc @wrwg @shazqadeer

Changes on the VM, specially if hashing is effected, can cause such errors. Could you run `UPBL=1 cargo test -p move-prover`? You need to have boogie and Z3 installed (at the right versions) and some env variables set. See move-prover/scripts for instructions. Note if you do not have those installed tests will silently pass but still fail in CI. Thanks!

huitseeker:
@wrwg There are no changes on the VM, hashing is not affected. The changes outside of the `crypto`   crate are limited to this single commit:
https://github.com/libra/libra/pull/4073/commits/2dd9b1deb7b7614e0d9df90f9395e103d5f93791

which changes the **name** of a function related to hashing, nothing else. No move-prover runs required.

davidiw:
Why isn't this a breaking change? Now when we sign messages, we don't hash with sha3. So a message signed with this branch is incompatible with the previous branch. Is my understanding incorrect?

kchalkias:
@davidiw the feature is introduced here, but not used yet. The next PR that will update everything to use this new logic and deprecate signing hashes will be a breaking change.

PR #4080, opened 2020-05-27
[vm] Report gas used instead of transaction fee in transaction output.

tzakian:
We were multiplying the gas used by the gas unit price, and we were actually reporting the transaction fee as opposed to the gas used for the transaction.


runtian-zhou:
Probably need to fix a unit test

Issue #3721, opened 2020-05-04
[move-prover][Feature Request] Warn about unused schemas


DavidLDill:
# 🚀 Feature Request

Warn about unused schemas

## Motivation

I defined a very nice schema, but forgot to apply it.  So I might have thought that I has proved a property when I didn't.

## Pitch

**Describe the solution you'd like**

**Describe alternatives you've considered**

**Are you willing to open a pull request?** (See [CONTRIBUTING](../../CONTRIBUTING.md))

PR #4085, opened 2020-05-27
[move prover] generate warnings about unused schemas

emmazzz:
## Motivation

This PR adds the generation of warnings about schemas that are defined but not used. Warnings are generated after spec checker phase. And unlike errors, Move Prover will not immediately exit after warnings.

Yes.

## Test Plan

Added a test `functional/unused_schema.move`.

### DIFF
--- a/language/move-prover/spec-lang/src/env.rs
+++ b/language/move-prover/spec-lang/src/env.rs
@@ -342,6 +342,12 @@ impl GlobalEnv {
         self.error_with_notes(loc, msg, vec![]);
     }
 
+    /// Adds a warning to this environment.
+    pub fn warn(&self, loc: &Loc, msg: &str) {
+        let diag = Diagnostic::new_warning(msg, Label::new(loc.file_id, loc.span, ""));
+        self.add_diag(diag);
+    }
+
     /// Returns the unknown location.
     pub fn unknown_loc(&self) -> Loc {
         self.unknown_loc.clone()
@@ -423,9 +429,34 @@ impl GlobalEnv {
             .any(|d| d.severity >= Severity::Error)
     }
 
-    /// Writes accumulated diagnostics to writer.
+    /// Returns true if diagnostics have warning severity or worse.
+    pub fn has_warnings(&self) -> bool {
+        self.diags
+            .borrow()
+            .iter()
+            .any(|d| d.severity >= Severity::Warning)
+    }
+
+    /// Writes accumulated errors to writer.
     pub fn report_errors<W: WriteColor>(&self, writer: &mut W) {
-        for diag in self.diags.borrow().iter() {
+        for diag in self
+            .diags
+            .borrow()
+            .iter()
+            .filter(|d| d.severity >= Severity::Error)
+        {
+            emit(writer, &Config::default(), &self.source_files, diag).expect("emit must not fail");
+        }
+    }
+
+    /// Writes accumulated diagnostics with warning severity or worse to writer.
+    pub fn report_warnings<W: WriteColor>(&self, writer: &mut W) {
+        for diag in self
+            .diags
+            .borrow()
+            .iter()
+            .filter(|d| d.severity >= Severity::Warning)
+        {
             emit(writer, &Config::default(), &self.source_files, diag).expect("emit must not fail");
         }
     }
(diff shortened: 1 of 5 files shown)


emmazzz:
This is to address the issue #3721 raised by @DavidLDill. @wrwg should probably also review.

language/move-prover/spec-lang/src/env.rs
```diff
@@ -342,6 +342,12 @@ impl GlobalEnv {
         self.error_with_notes(loc, msg, vec![]);
     }
 
+    /// Adds an warning to this environment.
```
wrwg, line 345: Isn't it "a warning"? I'm myself all the time confused.

language/move-prover/spec-lang/src/env.rs
```diff
@@ -423,9 +429,34 @@ impl GlobalEnv {
...
+            .diags
+            .borrow()
+            .iter()
+            .filter(|d| d.severity >= Severity::Error)
+        {
+            emit(writer, &Config::default(), &self.source_files, diag).expect("emit must not fail");
+        }
+    }
+
+    /// Writes accumulated warnings to writer.
```
wrwg, line 452: This actually seems to write warnings and errors. Do we want to change the doc or the impl?

language/move-prover/spec-lang/src/translate.rs
```diff
@@ -1168,6 +1173,17 @@ impl<'env, 'translator> ModuleTranslator<'env, 'translator> {
 
         // Perform post reduction of module invariants.
         self.reduce_module_invariants();
+
+        // Warn about unused schemas.
+        for (schema_name, loc) in &self.parent.unused_schema_table {
+            self.parent.env.warn(
+                loc,
+                &format!(
+                    "A schema is not used: {}",
```
wrwg, line 1182: I suggest to make the message "unused schema `{}`". We usually start error messages with lowercase and use backquotes for references to symbols.

language/move-prover/tests/sources/functional/unused_schema.exp
```diff
@@ -0,0 +1,25 @@
+Move prover returns: exiting with boogie verification errors
```
wrwg, line 1: Can we give a better message here like "exiting with warnings"?

language/move-prover/tests/sources/functional/unused_schema.exp
```diff
@@ -0,0 +1,25 @@
...
+    │
+ 21 │ ╭     spec schema AddsThree {
+ 22 │ │         i: num;
+ 23 │ │         result: num;
+ 24 │ │         ensures result == i + 3;
+ 25 │ │     }
+    │ ╰─────^
+    │
+
+error:  A postcondition might not hold on this return path.
```
wrwg, line 14: Ah! It is actually exiting with verification errors as well. I suggest we just remove those errors to keep the tests clean. We may(?) need to fix the test driver in `tests/testsuite.rs` a bit to handle the warnings case correctly.

language/move-prover/spec-lang/src/env.rs
emmazzz: Ah it was a copy/paste mistake! Original was "Adds an error"

language/move-prover/spec-lang/src/translate.rs
emmazzz: I was actually following the format of the error message below "A postcondition might not hold on this return path." Is this error message directly from Boogie? Maybe we should standardize the format of error messages too.

language/move-prover/tests/sources/functional/unused_schema.exp
emmazzz: I think we can leave as is right now and change the driver when more use cases for warnings come up.

wrwg: That's the kind of thing which leads to entropy in code. "Later" will then easily become "never". I also don't know whether you need to do anything at the driver at all, my main issue here is that you have a verification error in tests around warnings and that does not make sense -- or did you do it for some purpose?

wrwg: Also, if the prover prints "exiting with verification errors" when you actually have only warnings is kind of a bad user experience. So this is not just about the tests.

emmazzz: Well if there are only warnings the prover will not print out "exiting with verification errors". It will only print out the warnings. The case that might lead to confusion is when there are both warnings and errors. Then prover would say "exiting with errors", followed by warning messages and error messages.

emmazzz: I have updated the test case so there is only a warning. And the exp file only contains the warning message.

PR #4086, opened 2020-05-27
[safety-rules] address sign_proposal TODOs

zaxguo:
## Motivation

sign_proposal safety rules are for leaders to make safe proposal for other validators to vote on.

Yes

## Test Plan

The sign_proposal is called and exercised in the _consensus_ package which continously generates new proposal to sign by fuzzing.
The test only checks the method won't crash under fuzzing. It, however, does not check the _correctness_ of the methods (e.g. if the rules are correctly implemented).

## Related PRs

### DIFF
--- a/consensus/safety-rules/src/safety_rules.rs
+++ b/consensus/safety-rules/src/safety_rules.rs
@@ -127,6 +127,19 @@ impl SafetyRules {
             Ok(())
         }
     }
+
+    /// This checkes whether the author of one proposal is the validator signer
+    fn verify_author(&self, author: Option<Author>) -> Result<(), Error> {
+        let validator_signer_author = &self.validator_signer.author();
+        let author = author
+            .ok_or_else(|| Error::InvalidProposal("No author found in the proposal".into()))?;
+        if (*validator_signer_author) != author {
+            return Err(Error::InvalidProposal(
+                "Proposal author is not validator signer!".into(),
+            ));
+        }
+        Ok(())
+    }
 }
 
 impl TSafetyRules for SafetyRules {
@@ -221,19 +234,16 @@ impl TSafetyRules for SafetyRules {
         ))
     }
 
-    /// @TODO only sign blocks that are later than last_voted_round and match the current epoch
-    /// @TODO verify QC correctness
-    /// @TODO verify QC matches preferred round
-    fn sign_proposal(&mut self, block_data: BlockData) -> Result<Block, Error> {
+    fn sign_proposal(&mut self, block_data: BlockData<T>) -> Result<Block<T>, Error> {
         debug!("Incoming proposal to sign.");
         let proposal_epoch = block_data.epoch();
         let proposal_round = block_data.round();
         let last_voted_round = self.persistent_storage.last_voted_round()?;
 
-        // proposed epoch must match the current epoch
+        self.verify_author(block_data.author())?;
+
         self.verify_epoch(proposal_epoch)?;
 
-        // proposed round must be increasing
         if proposal_round <= last_voted_round {
             debug!(
                 "Block round is older than last_voted_round ({} <= {})",
@@ -245,7 +255,6 @@ impl TSafetyRules for SafetyRules {
             });
         }
 
-        // Done checking the block epochs, now checking and verifying QC correctness
         let qc = block_data.quorum_cert();
         let qc_round = qc.certified_block().round();
         let preferred_round = self.persistent_storage.preferred_round()?;
--- a/consensus/safety-rules/src/safety_rules.rs
+++ b/consensus/safety-rules/src/safety_rules.rs
@@ -133,7 +133,7 @@ impl SafetyRules {
         let validator_signer_author = &self.validator_signer.author();
         let author = author
             .ok_or_else(|| Error::InvalidProposal("No author found in the proposal".into()))?;
-        if (*validator_signer_author) != author {
+        if validator_signer_author != &author {
             return Err(Error::InvalidProposal(
                 "Proposal author is not validator signer!".into(),
             ));
@@ -234,37 +234,31 @@ impl TSafetyRules for SafetyRules {
         ))
     }
 
-    fn sign_proposal(&mut self, block_data: BlockData<T>) -> Result<Block<T>, Error> {
+    fn sign_proposal(&mut self, block_data: BlockData) -> Result<Block, Error> {
         debug!("Incoming proposal to sign.");
-        let proposal_epoch = block_data.epoch();
-        let proposal_round = block_data.round();
-        let last_voted_round = self.persistent_storage.last_voted_round()?;
-
         self.verify_author(block_data.author())?;
-
-        self.verify_epoch(proposal_epoch)?;
-
-        if proposal_round <= last_voted_round {
+        self.verify_epoch(block_data.epoch())?;
+        let last_voted_round = self.persistent_storage.last_voted_round()?;
+        if block_data.round() <= last_voted_round {
             debug!(
                 "Block round is older than last_voted_round ({} <= {})",
-                proposal_round, last_voted_round
+                block_data.round(),
+                last_voted_round
             );
             return Err(Error::OldProposal {
-                proposal_round,
+                proposal_round: block_data.round(),
                 last_voted_round,
             });
         }
 
         let qc = block_data.quorum_cert();
-        let qc_round = qc.certified_block().round();
-        let preferred_round = self.persistent_storage.preferred_round()?;
-
         self.verify_qc(qc)?;
-
-        if qc_round < preferred_round {
+        let preferred_round = self.persistent_storage.preferred_round()?;
+        if qc.certified_block().round() < preferred_round {
             debug!(
                 "QC round does not match preferred round {} < {}",
-                qc_round, preferred_round
+                qc.certified_block().round(),
+                preferred_round
             );
             return Err(Error::InvalidQuorumCertificate(
                 "QC's certified round is older than the preferred round".into(),
--- a/consensus/safety-rules/src/safety_rules.rs
+++ b/consensus/safety-rules/src/safety_rules.rs
@@ -226,6 +226,42 @@ impl TSafetyRules for SafetyRules {
     /// @TODO verify QC matches preferred round
     fn sign_proposal(&mut self, block_data: BlockData) -> Result<Block, Error> {
         debug!("Incoming proposal to sign.");
+        let proposal_epoch = block_data.epoch();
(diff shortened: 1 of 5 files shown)


consensus/safety-rules/src/safety_rules.rs
```diff
@@ -235,6 +235,39 @@ impl<T: Payload> TSafetyRules<T> for SafetyRules<T> {
...
+        if qc_round < preferred_round {
+            debug!(
+                "QC round does not match preferred round {} < {}",
+                qc_round, preferred_round
+            );
+            return Err(Error::InvalidQuorumCertificate(
+                "Preferred round too early".into(),
+            ));
+        }
+
```
davidiw, line 270: PR looks great!
Let's also verify the QC!

consensus/safety-rules/src/safety_rules.rs
```diff
@@ -235,6 +235,42 @@ impl<T: Payload> TSafetyRules<T> for SafetyRules<T> {
...
+        // proposed epoch must match the current epoch
+        self.verify_epoch(proposal_epoch)?;
+
+        // proposed round must be increasing
+        if proposal_round <= last_voted_round {
+            debug!(
+                "Block round is older than last_voted_round ({} <= {})",
+                proposal_round, last_voted_round
+            );
+            return Err(Error::OldProposal {
```
zekun000, line 251: not related to this pr, but this pattern of println + return Err makes me sad, can we have some magic that both prints and return the error?

we probably want another error type for this  to differentiate self-propose old proposal vs receiving remote old proposal

davidiw: I think the error is okay, because the context is different... and the text for the error is pretty neutral: https://github.com/libra/libra/blob/master/consensus/safety-rules/src/error.rs#L44

consensus/safety-rules/src/safety_rules.rs
```diff
@@ -235,6 +235,42 @@ impl<T: Payload> TSafetyRules<T> for SafetyRules<T> {
...
+
+        self.verify_qc(qc)?;
+
+        if qc_round < preferred_round {
+            debug!(
+                "QC round does not match preferred round {} < {}",
+                qc_round, preferred_round
+            );
+            return Err(Error::InvalidQuorumCertificate(
+                "Preferred round too early".into(),
```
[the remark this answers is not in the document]
davidiw: Yeah, this does sound very wrong!  Also the logging is a little interesting.

I would say the QCs certified round is older than the preferred round. Perhaps you can fix other interesting cases within the code :).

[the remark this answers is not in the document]
zaxguo: Thanks guys!I'll address the comments

zekun000:
I think you don't intend to include key verification in this PR?

zaxguo:
> I think you don't intend to include key verification in this PR?

Actually I thought about this and did this from two reasons:
1) the two tasks have dependency, which is caused by the change of validator_signer form <ValidatorSigner> to Option<ValidatorSigner> -- this will be used by the sign_proposal() in the 1st commit. But squashing them into one commit would be brutal, hence the two commits.
2) both tasks belong to the TODO items in the same safety_rules.rs file. Considering the lines of changes are not drastic, placing them under one PR may make sense as long as the PR name is revised to reflect the correct changes (i.e. "[safety-rules] clearing all TODO items").

But I have to admit I'm a complete Git noob in terms of engineering practice so I'm happy to split it into two PRs if necessary :)

zekun000:
it's fine I was just to confirm. if that's intended, you probably should call out and update the title/summary to reflect it, to set the right expectations for reviewers.

zaxguo:
> it's fine I was just to confirm. if that's intended, you probably should call out and update the title/summary to reflect it, to set the right expectations for reviewers.

David thinks it'll be better to separate them, so I think I'll just separate them.

consensus/safety-rules/src/safety_rules.rs
```diff
@@ -235,6 +235,42 @@ impl<T: Payload> TSafetyRules<T> for SafetyRules<T> {
     /// @TODO verify QC matches preferred round
     fn sign_proposal(&mut self, block_data: BlockData<T>) -> Result<Block<T>, Error> {
         debug!("Incoming proposal to sign.");
+        let proposal_epoch = block_data.epoch();
+        let proposal_round = block_data.round();
+        let last_voted_round = self.persistent_storage.last_voted_round()?;
+
+        // proposed epoch must match the current epoch
```
davidiw, line 242: I don't think these comments add clarity, but open for discussion

consensus/safety-rules/src/safety_rules.rs
```diff
@@ -235,6 +235,42 @@ impl<T: Payload> TSafetyRules<T> for SafetyRules<T> {
...
+        self.verify_epoch(proposal_epoch)?;
+
+        // proposed round must be increasing
```
davidiw, line 245: same here, we have a debug statement below describing why we have this block of code, so a comment is just more overhead

zaxguo: Leaving comments in the code is kind of a personal habbit. Will remove them thereafter..

zaxguo:
> remove to do
> and add a verification that the proposal comes from someone within the validator verifier
> and that the signature is correct

I think this is achieved by ```self.verify(qc)?;``` (safety_ruies.rs:262) which further calls into the ValidatorVerifier to check if the proposal comes from someone within the validator verifier and the signature matches. Do I miss anything?

davidiw:
> > remove to do
> > and add a verification that the proposal comes from someone within the validator verifier
> > and that the signature is correct
> 
> Are you referring to the TODOs in construct_and_sign_vote()?

Sorry, I made a mistake, Block has a signature but we are passing in BlockData here: https://github.com/libra/libra/blob/master/consensus/consensus-types/src/block.rs#L31

I think this is good enough, but maybe leave a TODO to consider changing the API to be Block.

CC: @zekun000, what are your thoughts? I think it is good hygiene. But definitely not in this PR.

zekun000:
we can verify the author in the BlockData matches the author in validator signer

consensus/safety-rules/src/safety_rules.rs
zekun000: I prefer code as comments too, only leave comments when the code is not obvious

zaxguo:
This sounds good to me -- we ensure the proposal is generated by the same validator, as in ProposalGenerator and the validator signer

consensus/safety-rules/src/safety_rules.rs
```diff
@@ -230,11 +243,44 @@ impl<T: Payload> TSafetyRules<T> for SafetyRules<T> {
         ))
     }
 
-    /// @TODO only sign blocks that are later than last_voted_round and match the current epoch
-    /// @TODO verify QC correctness
-    /// @TODO verify QC matches preferred round
+    /// @TODO consider the API change: BlockData<T> to Block<T>
```
zekun000, line 246: let's remove this, I don't think this is possible, the Block<T> is the return value which contains the signature

davidiw: hahaha, did I mix two apis up! I intended this to be for sign and construct vote :P

@zekun000 is right!

consensus/safety-rules/src/safety_rules.rs
```diff
@@ -230,11 +243,37 @@ impl<T: Payload> TSafetyRules<T> for SafetyRules<T> {
...
     }
 
-    /// @TODO only sign blocks that are later than last_voted_round and match the current epoch
-    /// @TODO verify QC correctness
-    /// @TODO verify QC matches preferred round
     fn sign_proposal(&mut self, block_data: BlockData<T>) -> Result<Block<T>, Error> {
         debug!("Incoming proposal to sign.");
+        self.verify_author(block_data.author())?;
+        self.verify_epoch(block_data.epoch())?;
+        let last_voted_round = self.persistent_storage.last_voted_round()?;
```
davidiw, line 250: would recommend a new line between 249 and 250, but this is okay, I have nitpicked a lot on this PR :P

consensus/safety-rules/src/tests/suite.rs
```diff
@@ -485,3 +486,89 @@ fn test_voting_potential_commit_id(func: RoundCallback) {
         a3.block().id(),
     );
 }
+
+fn test_sign_bad_proposals(func: RoundCallback) {
```
davidiw, line 490: I think this is new. We shouldn't have an 85 line test, no one can reasonably read this without focus. Tests and code in general should be accessible and intuitive. Can we break this up into one function per underlying test case?

consensus/safety-rules/src/safety_rules.rs
```diff
@@ -136,6 +136,19 @@ impl<T: Payload> SafetyRules<T> {
             Ok(())
         }
     }
+
+    /// This checkes whether the author of one proposal is the validator signer
+    fn verify_author(&self, author: Option<Author>) -> Result<(), Error> {
+        let validator_signer_author = &self.validator_signer.author();
```
davidiw, line 142: Maybe just put this inline unless we use it multiple times?

consensus/safety-rules/src/safety_rules.rs
```diff
@@ -136,6 +136,19 @@ impl<T: Payload> SafetyRules<T> {
...
+        let author = author
+            .ok_or_else(|| Error::InvalidProposal("No author found in the proposal".into()))?;
+        if (*validator_signer_author) != author {
```
davidiw, line 145: I think if you instead do:
`if validator_signer_author != &author`
you can avoid the double parentheses

consensus/safety-rules/src/tests/suite.rs
```diff
@@ -485,3 +486,89 @@ fn test_voting_potential_commit_id(func: RoundCallback) {
...
+    let (mut safety_rules, signer) = func();
+
+    let (proof, genesis_qc) = make_genesis::<Round>(&signer);
+    let round = genesis_qc.certified_block().round();
+    safety_rules.initialize(&proof).unwrap();
+
+    let a1 = test_utils::make_proposal_with_qc(round + 1, genesis_qc, &signer);
+    safety_rules.update(a1.block().quorum_cert()).unwrap();
+    let err = safety_rules.sign_proposal(a1.block().block_data().clone());
+    assert!(err.is_ok());
```
davidiw, line 502: recommend you just unwrap, that way if we get an err we can it is okay... also the variable name is a little strange since it is a result not an err or an ok

davidiw:
Great work! Thanks so much!

@bors-libra r+

PR #4087, opened 2020-05-27
[stdlib] Make writeset txn use the regular sequence number of an acco…

runtian-zhou:
…unt.

## Motivation

Previously WriteSet was using its own sequence number, which broke the invariant that sender address and sequence number has to be unique. This PR enforces the same safety check for accounts' sequence number and auth_key, and bumps the sequence number of the sender account to make the sequence number indeed unique.

Yes
## Test Plan

Added an extra test to make sure the sequence number is indeed bumped as expected.


language/libra-vm/src/system_module_names.rs
```diff
@@ -86,6 +86,8 @@ pub static CREATE_ACCOUNT_NAME: Lazy<Identifier> =
     Lazy::new(|| Identifier::new("create_unhosted_account").unwrap());
 pub static PROLOGUE_NAME: Lazy<Identifier> = Lazy::new(|| Identifier::new("prologue").unwrap());
 pub static EPILOGUE_NAME: Lazy<Identifier> = Lazy::new(|| Identifier::new("epilogue").unwrap());
+pub static BUMP_SEQUENCE_NAME: Lazy<Identifier> =
```
sblackshear, line 89:
```suggestion
pub static BUMP_SEQUENCE_NUMBER_NAME: Lazy<Identifier> =
```

language/stdlib/modules/libra_account.move
```diff
@@ -863,6 +863,13 @@ module LibraAccount {
             Libra::deposit(&mut transaction_fee_balance.coin, transaction_fee);
         }
     }
+
+    // Bump the sequence number of an account. This function should be used only for bumping the sequence number when
+    // a writeset transaction is committed.
+    fun bump_sequence_number() acquires T {
```
sblackshear, line 869: Could you use `&signer` here to get rid of the `Transaction::sender()` below?

language/stdlib/modules/libra_writeset_manager.move
```diff
@@ -9,7 +9,6 @@ module LibraWriteSetManager {
     use 0x0::LibraConfig;
```
sblackshear, line 9: This module is so small that it might be worth combining with `LibraAccount` to avoid the funkiness of `bump_sequence_number`....

runtian-zhou: My main concern is that this module will depend on `LibraConfig`. If we combine this with `LibraAccount` it might make our dependency story to be really weird.

Issue #3985, opened 2020-05-20
[network] Hardening


davidiw:
# LibraNet

This issue details ur vision for the initial LibraNet and the series of work that remains to get us there.

LibraNet consists of a set of framework to establish connections with remote peers over TCP connections enabling bidirectional communication, i.e., requests can be sent in either direction.

## Networks

The Libra Network contains 3 types of networks:

* a private validator network
* public access to the validator network via a set of fullnodes (validator fullnode network)
* many, distinct private networks, including, for example, one or more fullnodes connected directly to the validator

public => VFN => validator
private_external => private_validator => validator

Only the validator network has the notion of equal peers and forms a mesh network. The rest of the networks have clear downstream and upstream network endpoints, eventually leading up to the validators. There are no connections between peers at the same level.

## Protocols and data flow

These networks will speak the following protocols:

* Consensus (validators only)
* State Synchronization - retrieving the current state of the network
* Mempool - submitting transaction
* HealthCheck

The data flow across these networks isn't necessarily bidirectional. Libra has a designation of downstream / upstream. Downstream peers retrieve blockchain updates from upstream peers. Downstream peers push transactions to upstream peers. The network generalizes into a tree structure with the validator network as its root. Each node from further out is downstream from the node closer to the validators. Specifically,

* validator networks act as equal peers in which each peer plays the role as both an upstream and downstream peer to each other
* one or more fullnodes, connect via a private network, directly to the validator, this is known as the validator fullnode network or VFN. In this arrangement, the validator is upstream to these fullnodes
* the VFN provides public access to libra. Peers connected to this network are downstream to the VFN.

## Discovery

Each of these networks employs different means for discovery:

* validator network leverage the ledger, specifically, the current ValidatorSet, to learn of other validators network addresses and public keys, this is known as onchain discovery.
* similarly a peer desiring to connect to the VFN can learn of all VFNs from the current ValidatorSet.
* private networks must be manually configured leveraging the SeedPeers within the NodeConfig

## Security

LibraNet leverages noise to provide authentication and encryption. In practice, only authentication is a requirement. 

The expectation for each network is:

* validator network employs bidirection authentication via the data onchain
* VFN do not authenticate to validators but do authenticate validators (i.e., unidirectional authentication)
* nodes connecting to the VFN authenticate the VFN via the data onchain but those nodes themselves are not authenticated (i.e., unidirectional authentication)
* private networks can leverage SeedPeers for upstream authentication and network peers for (optional) downstream authentication

[...]

PR #4088, opened 2020-05-27
[network] add NetworkId to config to differentiate

### LINKED ISSUE #4012: [network] disambiguate networks with `NetworkId` to fix network metrics/logs.

gregnazario:
## Motivation

Each network device that is on a different network currently all log
the same way, which makes it difficult to determine which network that
one is looking at.  This allows us to differentiate both at a logging
level and later at a communication level between networks.

resolves #4012

Yes

## Test Plan

Added tests to check that network handshakes work with the new network Ids.  This also means that both sides need to be updated for communication to work.

## Related PRs

### DIFF
--- a/config/config-builder/src/full_node_config.rs
+++ b/config/config-builder/src/full_node_config.rs
@@ -7,7 +7,9 @@ use libra_config::{
     config::{
         NetworkPeerInfo, NetworkPeersConfig, NodeConfig, PeerNetworkId, RoleType, UpstreamConfig,
     },
-    generator, utils,
+    generator,
+    network_id::NetworkId,
+    utils,
 };
 use libra_crypto::ed25519::Ed25519PrivateKey;
 use libra_network_address::NetworkAddress;
@@ -175,6 +177,7 @@ impl FullNodeConfig {
                 .full_node_networks
                 .get_mut(0)
                 .ok_or(Error::MissingFullNodeNetwork)?;
+            network.network_id = NetworkId::vfn_network();
             network.listen_address = utils::get_available_port_in_multiaddr(true);
             network.advertised_address = network.listen_address.clone();
             network.enable_remote_authentication = self.enable_remote_authentication;
@@ -208,6 +211,7 @@ impl FullNodeConfig {
                 .full_node_networks
                 .last_mut()
                 .ok_or(Error::MissingFullNodeNetwork)?;
+            network.network_id = NetworkId::Public;
             network.network_peers = network_peers.clone();
             network.seed_peers = seed_peers.clone();
             let mut upstream = UpstreamConfig::default();
--- a/config/management/src/smoke_test.rs
+++ b/config/management/src/smoke_test.rs
@@ -3,9 +3,12 @@
 
 use crate::{layout::Layout, storage_helper::StorageHelper};
 use config_builder::{BuildSwarm, SwarmConfig};
-use libra_config::config::{
-    DiscoveryMethod, Identity, NetworkConfig, NodeConfig, OnDiskStorageConfig, RoleType,
-    SecureBackend,
+use libra_config::{
+    config::{
+        DiscoveryMethod, Identity, NetworkConfig, NodeConfig, OnDiskStorageConfig, RoleType,
+        SecureBackend,
+    },
+    network_id::NetworkId,
 };
 use libra_crypto::ed25519::Ed25519PrivateKey;
 use libra_secure_storage::Value;
@@ -69,11 +72,11 @@ fn smoke_test() {
         let validator_account = account_address::from_public_key(&operator_key);
         let mut config = NodeConfig::default();
 
-        let mut network = NetworkConfig::default();
+        let mut network = NetworkConfig::network_with_id(NetworkId::Validator);
         network.discovery_method = DiscoveryMethod::None;
         config.validator_network = Some(network);
 
-        let mut network = NetworkConfig::default();
+        let mut network = NetworkConfig::network_with_id(NetworkId::vfn_network());
         network.discovery_method = DiscoveryMethod::None;
         config.full_node_networks = vec![network];
         config.randomize_ports();
--- a/config/src/config/mod.rs
+++ b/config/src/config/mod.rs
@@ -5,6 +5,7 @@ use anyhow::{ensure, Result};
 use rand::{rngs::StdRng, SeedableRng};
 use serde::{de::DeserializeOwned, Deserialize, Serialize};
 use std::{
+    collections::HashSet,
     fmt,
     fs::File,
     io::{Read, Write},
@@ -44,6 +45,7 @@ pub use safety_rules_config::*;
 mod upstream_config;
 pub use upstream_config::*;
 mod test_config;
+use crate::network_id::NetworkId;
 use libra_types::waypoint::Waypoint;
 pub use test_config::*;
 
@@ -205,13 +207,23 @@ impl NodeConfig {
             );
         }
 
+        let mut network_ids = HashSet::new();
         let input_dir = RootPath::new(input_path);
         config.execution.load(&input_dir)?;
         if let Some(network) = &mut config.validator_network {
             network.load(&input_dir, RoleType::Validator)?;
+            network_ids.insert(network.network_id.clone());
         }
         for network in &mut config.full_node_networks {
             network.load(&input_dir, RoleType::FullNode)?;
+
+            // Validate that a network isn't repeated
+            let network_id = network.network_id.clone();
+            ensure!(
+                !network_ids.contains(&network_id),
+                format!("network_id {:?} was repeated", network_id)
+            );
+            network_ids.insert(network_id);
         }
         config.set_data_dir(config.data_dir().clone());
         Ok(config)
@@ -276,7 +288,8 @@ impl NodeConfig {
             );
 
             if self.validator_network.is_none() {
-                self.validator_network = Some(NetworkConfig::default());
+                let network_config = NetworkConfig::network_with_id(NetworkId::Validator);
+                self.validator_network = Some(network_config);
             }
 
             let validator_network = self.validator_network.as_mut().unwrap();
@@ -285,7 +298,8 @@ impl NodeConfig {
         } else {
             self.validator_network = None;
             if self.full_node_networks.is_empty() {
-                self.full_node_networks.push(NetworkConfig::default());
+                let network_config = NetworkConfig::network_with_id(NetworkId::Public);
+                self.full_node_networks.push(network_config);
             }
             for network in &mut self.full_node_networks {
                 network.random(rng);
--- a/config/src/config/test_data/random.complete.node.config.toml
+++ b/config/src/config/test_data/random.complete.node.config.toml
@@ -84,3 +84,6 @@ seed_peers_file = "31893204fa402143c11b26ce8a89ea1d.seed_peers.toml"
 type = "from_config"
 key = "60dd107034b4582a2ef42c5e1ea475f2fea477a10a9f1d75b3635243b2506b72"
 peer_id = "31893204fa402143c11b26ce8a89ea1d"
+
+[validator_network.network_id]
+type = "Validator"
--- a/config/src/network_id.rs
+++ b/config/src/network_id.rs
@@ -1,0 +1,126 @@
+// Copyright (c) The Libra Core Contributors
+// SPDX-License-Identifier: Apache-2.0
+use serde::{Deserialize, Deserializer, Serialize, Serializer};
+
+#[derive(Clone, Debug, Deserialize, Eq, Hash, PartialEq, Serialize)]
+pub struct NetworkInfo {
+    name: String,
+}
+
+/// A representation of the network being used in communication.
+/// There should only be one of each NetworkId used for a single node, and handshakes should verify
+/// that the NetworkId being used is the same during a handshake, to effectively ensure communication
+/// is restricted to a network.  Network should be checked that it is not the `DEFAULT_NETWORK`
+#[derive(Clone, Debug, Eq, Hash, PartialEq)]
+pub enum NetworkId {
+    Validator,
+    Public,
+    Private(NetworkInfo),
+}
+
+/// Default needed to handle downstream structs that use `Default`
+impl Default for NetworkId {
+    fn default() -> NetworkId {
+        NetworkId::Public
+    }
+}
+
+impl NetworkId {
+    /// Convenience function to specify the VFN network
+    pub fn vfn_network() -> NetworkId {
+        NetworkId::private_network("VFN")
+    }
+
+    /// Creates a private network so we don't have to keep track of `NetworkInfo` outside of here.
+    pub fn private_network(name: &str) -> NetworkId {
+        NetworkId::Private(NetworkInfo {
+            name: name.to_string(),
+        })
+    }
+}
+
+#[derive(Serialize, Deserialize)]
+#[serde(tag = "type", rename = "NetworkId")]
+enum HumanReadableNetworkId {
+    Validator,
+    Public,
+    Private(NetworkInfo),
+}
+
+#[derive(Serialize, Deserialize)]
+#[serde(rename = "NetworkId")]
+enum NonHumanReadableNetworkId {
+    Validator,
+    Public,
+    Private(NetworkInfo),
+}
+
+impl Serialize for NetworkId {
+    fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
+    where
+        S: Serializer,
+    {
+        if serializer.is_human_readable() {
+            match self {
+                NetworkId::Validator => HumanReadableNetworkId::Validator,
+                NetworkId::Public => HumanReadableNetworkId::Public,
+                NetworkId::Private(info) => HumanReadableNetworkId::Private(info.clone()),
+            }
+            .serialize(serializer)
+        } else {
+            match self {
+                NetworkId::Validator => NonHumanReadableNetworkId::Validator,
+                NetworkId::Public => NonHumanReadableNetworkId::Public,
+                NetworkId::Private(info) => NonHumanReadableNetworkId::Private(info.clone()),
+            }
+            .serialize(serializer)
+        }
+    }
+}
+
+impl<'de> Deserialize<'de> for NetworkId {
+    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
+    where
+        D: Deserializer<'de>,
+    {
+        if deserializer.is_human_readable() {
+            Ok(match HumanReadableNetworkId::deserialize(deserializer)? {
+                HumanReadableNetworkId::Validator => NetworkId::Validator,
+                HumanReadableNetworkId::Public => NetworkId::Public,
+                HumanReadableNetworkId::Private(info) => NetworkId::Private(info),
+            })
+        } else {
+            Ok(
+                match NonHumanReadableNetworkId::deserialize(deserializer)? {
+                    NonHumanReadableNetworkId::Validator => NetworkId::Validator,
+                    NonHumanReadableNetworkId::Public => NetworkId::Public,
+                    NonHumanReadableNetworkId::Private(info) => NetworkId::Private(info),
+                },
+            )
+        }
+    }
+}
+
+#[cfg(test)]
+mod test {
+    use super::*;
+
+    #[test]
+    fn test_serialization() {
+        let id = NetworkId::private_network("fooo");
+        let encoded = toml::to_string(&id).unwrap();
+        let decoded: NetworkId = toml::from_str(encoded.as_str()).unwrap();
+        assert_eq!(id, decoded);
+        let encoded = toml::to_vec(&id).unwrap();
+        let decoded: NetworkId = toml::from_slice(encoded.as_slice()).unwrap();
+        assert_eq!(id, decoded);
+
+        let id = NetworkId::Validator;
+        let encoded = toml::to_string(&id).unwrap();
+        let decoded: NetworkId = toml::from_str(encoded.as_str()).unwrap();
+        assert_eq!(id, decoded);
+        let encoded = toml::to_vec(&id).unwrap();
+        let decoded: NetworkId = toml::from_slice(encoded.as_slice()).unwrap();
+        assert_eq!(id, decoded);
+    }
+}
--- a/network/src/transport.rs
+++ b/network/src/transport.rs
@@ -10,6 +10,7 @@ use crate::{
     },
 };
 use futures::io::{AsyncRead, AsyncWrite};
+use libra_config::network_id::NetworkId;
 use libra_crypto::{traits::ValidCryptoMaterial, x25519};
 use libra_logger::prelude::*;
 use libra_network_address::NetworkAddress;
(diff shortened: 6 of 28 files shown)


types/src/network_id.rs
```diff
@@ -0,0 +1,41 @@
...
+// SPDX-License-Identifier: Apache-2.0
+use serde::{Deserialize, Serialize};
+
+/// A representation of the network being used in communication.
+/// There should only be one of each NetworkId used for a single node, and handshakes should verify
+/// that the NetworkId being used is the same during a handshake, to effectively ensure communication
+/// is restricted to a network.  Network should be checked that it is not the `DEFAULT_NETWORK`
+/// @TODO: Figure out why a normal struct with an enum and String doesn't work here for serialization
+#[derive(Clone, Debug, Deserialize, Eq, Hash, PartialEq, Serialize)]
+pub struct NetworkId(String);
```
davidiw, line 11: PR is coming along really well! Thought I had left a comment somewhere. I left a note in the issue. I think we want an enum here, it will make it much more flexible / usable going forward. Also I think it is okay for this to live in the config itself, was there a reason to have it types?

gregnazario: Wasn't quite sure where to put it so I put it where PeerId / AccountAddress was.  I did get a message about the enum, and planning to do it next :)  The Newtype struct was a stopgap to get the tests to run and not crash on previous attempts of an enum that didn't serialize to toml.

If its only used for config/handshake purposes i think config is fine.  if we use it past just handshakes, we might want to leave it in the types to be more general

davidiw: Sweet. In general, types is intended to be our dumping ground for client interfaces. The network id is probably outside the scope of that at least at this time , /I think/.

gregnazario: Sounds good, moved to config.

config/config-builder/src/full_node_config.rs
```diff
@@ -166,6 +168,7 @@ impl FullNodeConfig {
         let new_net = full_node_config.full_node_networks.swap_remove(0);
         for net in &config.full_node_networks {
             ensure!(new_net.peer_id != net.peer_id, "Network already exists");
+            // TODO do we want to distinguish between public network ids
```
davidiw, line 171: There is only one public network, in my opinion :D.

config/config-builder/src/full_node_config.rs
```diff
@@ -218,6 +221,8 @@ impl FullNodeConfig {
                 .full_node_networks
                 .get_mut(0)
                 .ok_or(Error::MissingFullNodeNetwork)?;
+            // @TODO: We need to determine a name for this Private network
```
davidiw, line 224: this is config-builder... if this is a validator node or a fullnode connected to a validator, then this should probably be "vfn"
otherwise this is the public network

let's just call all the networks in config-builder private "vfn" and let @mgorven help us adjust if possible.

I think @agclement has also spent time here, perhaps he can make a suggestion?

config/management/src/smoke_test.rs
```diff
@@ -71,10 +72,12 @@ fn smoke_test() {
         let mut network = NetworkConfig::default();
         network.discovery_method = DiscoveryMethod::Onchain;
         network.peer_id = validator_account;
+        network.network_id = NetworkId::validator_network();
         config.validator_network = Some(network);
 
         let mut network = NetworkConfig::default();
         network.discovery_method = DiscoveryMethod::None;
+        network.network_id = NetworkId::public_network();
```
davidiw, line 80: this is private_network "vfn"

config/src/config/mod.rs
```diff
@@ -278,7 +290,9 @@ impl NodeConfig {
             );
 
             if self.validator_network.is_none() {
-                self.validator_network = Some(NetworkConfig::default());
+                let mut network_config = NetworkConfig::default();
+                network_config.network_id = NetworkId::validator_network();
```
davidiw, line 294: maybe we should make a new function that takes in a network id? maybe not...

config/src/config/test_data/random.complete.node.config.toml
```diff
@@ -82,3 +82,7 @@ enable_remote_authentication = true
 network_peers_file = ""
 seed_peers_file = "31893204fa402143c11b26ce8a89ea1d.seed_peers.toml"
 identity_private_key = "60dd107034b4582a2ef42c5e1ea475f2fea477a10a9f1d75b3635243b2506b72"
+
+[validator_network.network_id]
+network_name = ""
```
davidiw, line 87: this shouldn't have a network_name :O
also I would just use name and type here otherwise it is a bit redundant

config/src/network_id.rs
```diff
@@ -0,0 +1,87 @@
+// Copyright (c) The Libra Core Contributors
+// SPDX-License-Identifier: Apache-2.0
+use serde::{Deserialize, Serialize};
+
+#[derive(Clone, Debug, Deserialize, Eq, Hash, PartialEq, Serialize)]
+pub enum NetworkType {
+    Validator = 0,
+    Public,
+    Private,
```
davidiw, line 9: Why didn't we follow the proposed format of

```
pub enum NetworkType {
  Validator,
  Public,
  Private(PrivateNetwork),
}

pub struct PrivateNetwork {
  name: String,
}
```

config/src/network_id.rs
```diff
@@ -0,0 +1,87 @@
...
+        }
+    }
+
+    /// Global Validator network
+    pub fn validator_network() -> NetworkId {
+        NetworkId {
+            network_type: NetworkType::Validator,
+            network_name: "".to_string(),
+        }
+    }
```
davidiw, line 62: your current approach requires this extra code.

network/src/transport.rs
```diff
@@ -397,3 +420,151 @@ pub fn build_tcp_transport(
         .with_timeout(TRANSPORT_TIMEOUT)
         .boxed()
 }
+
+#[cfg(test)]
```
davidiw, line 424: feel mixed about these tests, I don't know if we benefit from adding them to the code base, it is a lot to read and potentially modify... Ideally we cover these from a unit test perspective in the HandshakeMsg. Then we also handle it from our various integration and smoke tests. Thoughts? 150 lines of test code is a lot. Question, did this help you develop it?

config/config-builder/src/full_node_config.rs
mgorven: If public==unauthenticated then yes, but there may also be an authenticated fullnode-only network for trusted external fullnodes to use (e.g. fullnodes run by Novi in FB infra connecting to the fullnode of Novi's validator deployment).

config/config-builder/src/full_node_config.rs
mgorven: That sounds good enough for now.

config/config-builder/src/full_node_config.rs
agc-quadra: TL;DR --- the term _public_ is overloaded and _public network_ means different things to different people.

TL;DR(2) --- A clear, concise, consistent understanding of deployment targets is needed.


A couple of different definitions of _public_

* unauthenticated and open to full participation by anyone
* accessible to anyone

The former definition is most inline with traditional blockchain projects.  This definition goes hand in hand with unconstrained deployments.

The second definition allows for networks which are deployed in a controlled environment, but accessible to anyone. A reasonable analogue is the public front ends that stand between users  and the back end infrastructure.

Looking at the [discussion](https://github.com/libra/libra/issues/3985#issuecomment-636659952) in #3985, one irony is that the suggested "private" networks connected to the public ring are in fact the most inline with the first definition of public.

config/src/config/test_data/random.complete.node.config.toml
gregnazario: `type` is a reserved word in rust :(. So, to be consistent I put `network` before both.

config/src/network_id.rs
gregnazario: TOML serialization determines that an ENUM with any extra values was an UnsupportedType.  See: https://github.com/libra/libra/issues/4210

I can go back and try it, but this should get the ball rolling.

config/src/network_id.rs
gregnazario: I can cut this out, the reason it required extra code was I found myself constantly switching between enums and structs trying to figure out which one was actually allowed in TOML.  By abstracting it away it let me change it easier.

I really wanted to use the enum approach, but TOML caused a whole bunch of serialization issues.

network/src/transport.rs
gregnazario: It's a lot because I copied them from a different class mostly because I didn't know what I was doing.

The issue here is there are no tests actually checking if there are checks on the handshakes.  All of the tests are whether the handshake works at a wire level, but there are none at the transport level.

I think a negative test is important to know that we are actually validating our connections.  The smoke test can handle the positive test.

config/src/config/mod.rs
gregnazario: I can take a look into that, there's just a lot of defaults used, and I think when I started this I wasn't sure if the higher level configs call NetworkConfig default.

config/src/network_id.rs
gregnazario: For everyone's knowledge, toml, won't serialize the enum:
- toml: https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=0e206f00f9d42117e9c3dc1906ba1762
- yaml: https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=0e206f00f9d42117e9c3dc1906ba1762

gregnazario: Got it mostly working after putting in a tag.  Now just some side effects in the serde analyzer.

config/src/config/test_data/random.complete.node.config.toml
gregnazario: Solved with a serde tag.

config/src/config/mod.rs
gregnazario: I've combined these.  But, can't fully get rid of `NetworkConfig::default()` because it is used in deserialization.

config/src/network_id.rs
```diff
@@ -0,0 +1,78 @@
...
+pub struct NetworkInfo {
+    name: String,
+}
+
+/// A representation of the network being used in communication.
+/// There should only be one of each NetworkId used for a single node, and handshakes should verify
+/// that the NetworkId being used is the same during a handshake, to effectively ensure communication
+/// is restricted to a network.  Network should be checked that it is not the `DEFAULT_NETWORK`
+#[derive(Clone, Debug, Deserialize, Eq, Hash, PartialEq, Serialize)]
+#[serde(tag = "type")]
```
matbd, line 15: This option is likely the cause of the breakage in the `generate-format` unit test. It would not work with LCS in first place.

Any reason for using this? Is this meant to be serialized both in JSON and LCS? (then I don't have a solution but to accept a slightly uglier JSON)

davidiw: Can we eliminate the notion of "tag" from LCS, I don't think we need to serialize / deserialize this field... I don't know why serde would do this to us :P

matbd: OK so if we really want to share the definition of `NetworkId` between TOML and LCS, here is a solution:
```
#[derive(Clone, Debug, Eq, Hash, PartialEq)]
pub enum NetworkId {
    Validator,
    Public,
    Private(NetworkInfo),
}

#[derive(Serialize, Deserialize)]
#[serde(tag = "type", rename = "NetworkId")]
enum HumanReadableNetworkId {
    Validator,
    Public,
    Private(NetworkInfo),
}

#[derive(Serialize, Deserialize)]
#[serde(rename = "NetworkId")]
enum NonHumanReadableNetworkId {
    Validator,
    Public,
    Private(NetworkInfo),
}

impl Serialize for NetworkId {
    fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
    where
        S: Serializer
    {
        if serializer.is_human_readable() {
            match self {
                NetworkId::Validator => HumanReadableNetworkId::Validator.serialize(serializer),
                NetworkId::Public => HumanReadableNetworkId::Public.serialize(serializer),
                NetworkId::Private(info) => HumanReadableNetworkId::Private(*info).serialize(serializer),
            }
        } else {
            match self {
                NetworkId::Validator => NonHumanReadableNetworkId::Validator.serialize(serializer),
                NetworkId::Public => NonHumanReadableNetworkId::Public.serialize(serializer),
                NetworkId::Private(info) => NonHumanReadableNetworkId::Private(*info).serialize(serializer),
            }
        }
    }
}

impl<'de> Deserialize<'de> for NetworkId {

    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
    where
        D: Deserializer<'de>
    {
        if deserializer.is_human_readable() {
            match HumanReadableNetworkId::deserialize(deserializer)? {
                HumanReadableNetworkId::Validator => Ok(NetworkId::Validator),
                HumanReadableNetworkId::Public => Ok(NetworkId::Public),
                HumanReadableNetworkId::Private(info) => Ok(NetworkId::Private(info)),
            }
        } else {
            match NonHumanReadableNetworkId::deserialize(deserializer)? {
                NonHumanReadableNetworkId::Validator => Ok(NetworkId::Validator),
                NonHumanReadableNetworkId::Public => Ok(NetworkId::Public),
                NonHumanReadableNetworkId::Private(info) => Ok(NetworkId::Private(info)),
            }
        }
}
```

Obviously, it's not ideal but it works. If this happens often in the future, we could think of some procedural macros to help.

config/src/config/mod.rs
```diff
@@ -287,7 +300,8 @@ impl NodeConfig {
         } else {
             self.validator_network = None;
             if self.full_node_networks.is_empty() {
-                self.full_node_networks.push(NetworkConfig::default());
+                let network_config = NetworkConfig::network_with_id(NetworkId::Public);
```
davidiw, line 303: @mgorven is this the public network or is it the internal network for VFN or are we even going to use config-builder at all come launch?

davidiw: note: this is the FN directly on the validator

config/src/config/mod.rs
```diff
@@ -400,7 +414,6 @@ mod test {
         let root_dir = RootPath::new(expected.test.as_ref().unwrap().temp_dir().unwrap());
         let path = root_dir.full_path(&PathBuf::from("node.config.toml"));
         expected.save(&path).expect("Unable to save config");
-
```
davidiw, line 403: why do you hate whitespace?

config/src/network_id.rs
```diff
@@ -0,0 +1,136 @@
...
+/// that the NetworkId being used is the same during a handshake, to effectively ensure communication
+/// is restricted to a network.  Network should be checked that it is not the `DEFAULT_NETWORK`
+#[derive(Clone, Debug, Eq, Hash, PartialEq)]
+pub enum NetworkId {
+    Validator,
+    Public,
+    Private(NetworkInfo),
+}
+
+const INVALID: &str = "Invalid";
```
davidiw, line 21: why?

config/src/network_id.rs
```diff
@@ -0,0 +1,136 @@
...
+impl NetworkId {
+    /// Tell us whether the `NetworkId` was configured correctly.
+    /// Private networks must have a name, otherwise we can't differentiate.
+    pub fn is_valid(&self) -> bool {
+        match self {
+            NetworkId::Validator => true,
+            NetworkId::Public => true,
+            NetworkId::Private(network_info) => network_info.name == INVALID,
+        }
+    }
```
davidiw, line 39: they should all be valid imo, not sure what we get by having this?

config/src/network_id.rs
```diff
@@ -0,0 +1,136 @@
...
+
+#[derive(Serialize, Deserialize)]
+#[serde(rename = "NetworkId")]
+enum NonHumanReadableNetworkId {
+    Validator,
+    Public,
+    Private(NetworkInfo),
+}
+
+impl Serialize for NetworkId {
```
davidiw, line 68: very clever guys, thanks @matbd

network/src/transport.rs
```diff
@@ -397,3 +420,92 @@ pub fn build_tcp_transport(
...
+        let public_network = NetworkId::Public;
+
+        // Create client and server handshake messages.
+        let mut server_handshake = HandshakeMsg::new(validator_network);
+        server_handshake.add(
+            MessagingProtocolVersion::V1,
+            [
+                ProtocolId::ConsensusDirectSend,
+                ProtocolId::MempoolDirectSend,
+            ]
```
davidiw, line 459: we don't need to actually be this verbose, so maybe we can just strip out a lot of boiler plate stuff like this?

network/src/transport.rs
```diff
@@ -397,3 +420,92 @@ pub fn build_tcp_transport(
...
+            )
+            .await
+            .expect_err(
+                format!(
+                    "network_ids don't match own: {:?} received: {:?}",
+                    NetworkId::Validator,
+                    NetworkId::Public
+                )
+                .as_str(),
+            );
```
davidiw, line 487: probably could just do an unwrap_err here, unless you think we get additional signal from this? I don't think we do

network/src/transport.rs
```diff
@@ -397,3 +420,92 @@ pub fn build_tcp_transport(
...
+        let client = async move {
+            perform_handshake(
+                peer_id_2,
+                outbound,
+                network_addr_2,
+                ConnectionOrigin::Outbound,
+                &client_handshake,
+            )
+            .await
+            .expect_err(
```
davidiw, line 499: same a simple unwrap_err will strip out a lot of code

config/src/config/mod.rs
mgorven: I'm not sure when this code is run, but I'm guessing this is the internal validator-fullnode network on the fullnode side. For launch we need fullnodes to have two networks: the validator-fullnode network, and the public unauthenticated network.

I would prefer not to use `config-builder` in production, but we need a way to handle multiple fullnodes on the validator-fullnode network. If the fullnodes can all use the `fullnode_network` key from Vault with random peer IDs when talking to the validator then that should work. We just need to figure out the validator-side of that to trust that `fullnode_network` key.

davidiw: Let's talk offline about this and come up with a scheme that works. Networking is complicated :/

config/src/config/mod.rs
gregnazario: I actually really like whitespace.

I spent a lot of time putting printlns places to see what was going on, I may have deleted this line.

config/src/network_id.rs
gregnazario: I don't think this necessarily has to be a constant.

The purpose of this is when deserializing, it seemed I needed a default value (probably for the struct that I had before).  So, I added this to ensure that if it had been loaded with a lack of a configuration for a network, it wouldn't just all of a sudden connect to one of the three networks.

config/src/network_id.rs
gregnazario: Same as above.  I'll check if the Default value was because of the struct or an enum.

network/src/transport.rs
gregnazario: I stole this all from https://github.com/libra/libra/blob/1df4359a5fc4d2c82b636152987071b57021c768/network/src/protocols/identity.rs#L89-L138

I'll take a round to simplify it.

The only reason I added the test here and not in the identity protocol, is because it doesn't cover the application level check on equality of networkIds.

config/src/network_id.rs
gregnazario: The purpose of this was that there is a `Default` for `NetworkConfig` when deserializing.  However, I wanted to be explicit that if `NetworkConfig` fills in the default value, that we should not trust it because it loaded a default value from the `NetworkConfig`.

I'm not sure having a defaulting configuration is really what we want on a `NetworkConfig`

network/src/transport.rs
gregnazario: I think the idea was to ensure we aren't just getting some random other error

config/src/network_id.rs
```diff
@@ -0,0 +1,126 @@
...
+pub enum NetworkId {
+    Validator,
+    Public,
+    Private(NetworkInfo),
+}
+
+/// Default needed to handle downstream structs that use `Default`
+impl Default for NetworkId {
+    fn default() -> NetworkId {
+        NetworkId::private_network("Default")
```
davidiw, line 24: let's make the default the Public network or Validator network.

davidiw:
Don't forget to rebase!

gregnazario:
> Don't forget to rebase!

Let's hope that the 3rd rebase is the charm

PR #4092, opened 2020-05-28
[move-lang] fix offset in error message for invalid hexstring characters

bob-wilson:
The previous code had the wrong offset for hexstrings containing an even number of characters.

Yes

## Test Plan

Adjusted the existing tests so that we cover both cases of even and odd length hex strings. Verified that they now pass.


language/move-lang/src/parser/syntax.rs
```diff
@@ -434,14 +434,17 @@ fn parse_byte_string<'input>(tokens: &mut Lexer<'input>) -> Result<Vec<u8>, Erro
     let s = tokens.content();
     assert!(s.starts_with("x\""));
     let mut hex_string = String::from(&s[2..s.len() - 1]);
-    if hex_string.len() % 2 != 0 {
+    let adjust = if hex_string.len() % 2 != 0 {
         hex_string.insert(0, '0');
```
tnowacki, line 438: At some point, we will need to move this + the decoding into expansion, for higher fidelity for formatting + other tools

PR #4094, opened 2020-05-28
[Config Builder] Added configuration generation support for the Key Manager.

JoshLind:
## Motivation

This PR adds support for generating config files for the key manager. The purpose of this PR is so that the key manager can ultimately be created and deployed in a docker container (the next PR that will follow this one). For now, we add config-builder support for the key manager in two commits:

1. First, this PR adds the appropriate command line flags to the config builder for the key manager and uses the values held by those flags to generate key manager related configs. In this commit, we also refactor several dependent setter methods into a single method for safety rules. This makes things easier to reason about and avoids additional and unnecessary methods.
2. We update the config builder README.

Note: in commit (1) above, it appears we also fix a bug related to how secure storage is initialized for safety rules (and now the key manager). Currently, creating a safety-rules config using "on-disk" as the storage engine isn't supported (internally the config-builder just results in an in_memory implementation instead...?). 

Yes.

## Test Plan

Tested this PR locally by: (i) running the newly added test; (ii) compiling and running the config-builder binary with different flags and verified the generated configs by hand; (iii) running the safety-rules docker container to ensure that it starts;

## Related PRs

None.


config/config-builder/src/validator_config.rs
```diff
@@ -257,31 +346,48 @@ impl ValidatorConfig {
...
+    fn build_secure_backend(
+        &self,
+        storage_backend: &str,
+        server: Option<String>,
+        namespace: Option<String>,
+        token: Option<String>,
+    ) -> Result<SecureBackend> {
+        match storage_backend {
+            "in-memory" => Ok(SecureBackend::InMemoryStorage),
+            "on-disk" => Ok(SecureBackend::OnDiskStorage(OnDiskStorageConfig::default())),
```
JoshLind, line 345: Note, this change compared to the original version.

config/config-builder/src/main.rs
```diff
@@ -90,6 +92,43 @@ struct FullNodeArgs {
...
+    #[structopt(long)]
+    /// Specifies the token for secure storages that use credentials.
+    storage_token: Option<String>,
+    #[structopt(long)]
+    /// Specifies a unique namespace for the secure storage.
+    storage_namespace: Option<String>,
+    #[structopt(short = "o", long, parse(from_os_str))]
+    /// The output directory.
+    output_dir: PathBuf,
+    #[structopt(short = "t", long, parse(from_os_str))]
```
davidiw, line 124: I kinda regret having short options in this app, fwiw.

config/config-builder/src/main.rs
```diff
@@ -90,6 +92,43 @@ struct FullNodeArgs {
...
+    json_rpc_endpoint: Option<String>,
+    #[structopt(long)]
+    /// Specifies the rotation period for key rotations (in seconds).
+    rotation_period_secs: Option<u64>,
+    #[structopt(long)]
+    /// Specifies the length of time the key manager will periodically sleep (in seconds).
+    sleep_period_secs: Option<u64>,
+    #[structopt(long)]
+    /// Specifies the type of backend to use for the key manager: in-memory, on-disk, or vault.
+    storage_backend: Option<String>,
```
davidiw, line 111: I don't think this should be optional, it makes no sense to start this without a specific service and currently that service is vault ... right? perhaps we should just tailer this to vault?

config/config-builder/src/main.rs
```diff
@@ -305,10 +372,12 @@ fn safety_rules_common(args: &ValidatorCommonArgs) -> ValidatorConfig {
         .validator_index(args.validator_index)
         .validators(args.validators)
         .safety_rules_addr(args.safety_rules_addr.clone())
-        .safety_rules_backend(args.safety_rules_backend.clone())
-        .safety_rules_host(args.safety_rules_host.clone())
-        .safety_rules_namespace(args.safety_rules_namespace.clone())
-        .safety_rules_token(args.safety_rules_token.clone())
+        .safety_rules_backend(
```
davidiw, line 375: you so fancy

config/config-builder/src/validator_config.rs
```diff
@@ -30,6 +30,14 @@ pub struct ValidatorConfig {
...
     bootstrap: NetworkAddress,
     index: usize,
+    key_manager_json_rpc_endpoint: Option<String>,
+    key_manager_rotation_period_secs: Option<u64>,
+    key_manager_sleep_period_secs: Option<u64>,
+    key_manager_storage_backend: Option<String>,
+    key_manager_storage_host: Option<String>,
+    key_manager_storage_token: Option<String>,
+    key_manager_storage_namespace: Option<String>,
+    key_manager_txn_expiration_secs: Option<u64>,
```
davidiw, line 40: why is this part of ValidatorConfig? Wouldn't it make more sense for this to be in its own "Config"?

config/config-builder/src/validator_config.rs
```diff
@@ -160,27 +204,12 @@ impl ValidatorConfig {
...
 
         Ok(config)
     }
 
-    pub fn build_set(&self) -> Result<Vec<NodeConfig>> {
-        let (configs, _) = self.build_common(false)?;
-        Ok(configs)
-    }
-
-    pub fn build_faucet_client(&self) -> Result<(Ed25519PrivateKey, Waypoint)> {
```
davidiw, line 173: for example, this is only here, because validators / faucet are core to the system, but we could have had a faucet config and the validator config depended on the faucet config.

config/config-builder/src/validator_config.rs
```diff
@@ -257,31 +322,48 @@ impl ValidatorConfig {
...
+                server: server
+                    .as_ref()
+                    .ok_or_else(|| Error::MissingSafetyRulesHost)?
+                    .clone(),
+                ca_certificate: None,
+                token: Token::new_config(
+                    token
                         .as_ref()
-                        .ok_or_else(|| Error::MissingSafetyRulesHost)?
+                        .ok_or_else(|| Error::MissingSafetyRulesToken)?
```
davidiw, line 356: hahaha

config/config-builder/src/main.rs
JoshLind: I agree -- I've removed them from the key manager options, but I don't want to remove them from others (in case I break any 3rd party deps, e.g., the containers)

config/config-builder/src/main.rs
JoshLind: Thanks! As discussed offline, we'll assume that the key manager only works with vault. I've made only the namespace optional now

config/config-builder/src/validator_config.rs
JoshLind: Thanks! Moved to it's own config 😄

config/config-builder/src/validator_config.rs
JoshLind: Heheh.. Git diff is broke

JoshLind:
Thanks, @davidiw, this is ready for another look. I tested this locally by starting the safety-rules container (which internally runs config-builder) and everything looks okay (i.e., I don't see any diversions in behaviour between a clean repo and one that contains this PR), so we should be good for another look 😄

config/config-builder/src/key_manager_config.rs
```diff
@@ -0,0 +1,135 @@
...
+impl Default for KeyManagerConfig {
+    fn default() -> Self {
+        let template = NodeConfig::default();
+        Self {
+            json_rpc_endpoint: template.secure.key_manager.json_rpc_endpoint.clone(),
+            rotation_period_secs: None,
+            sleep_period_secs: None,
+            template,
+            txn_expiration_secs: None,
+            vault_host: "".to_string(),
```
JoshLind, line 27: @davidiw, the only thing I'm not certain about is defining defaults for vault_host and token? Sure, we could add garbage/test values for the host but what about the token? Is there really a default a token? (Similar in my mind to asking.. is there really a default private key? 😄)

config/config-builder/src/key_manager_config.rs
```diff
@@ -0,0 +1,135 @@
...
+        }
+    }
+}
+
+impl KeyManagerConfig {
+    pub fn new() -> Self {
+        Self::default()
+    }
+
+    pub fn json_rpc_endpoint(&mut self, json_rpc_endpoint: String) -> &mut Self {
```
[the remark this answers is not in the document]
JoshLind: Yeah, I did wonder that, too. Great, I'll handle that in a future refactor so that it doesn't block my container PR 😄

PR #4096, opened 2020-05-28
[cluster-test] Client side latency statistics P90 latency

snakech:
## Motivation

(Write your answer here.)

## Test Plan

## Related PRs

### DIFF
--- a/testsuite/cluster-test/src/atomic_histogram.rs
+++ b/testsuite/cluster-test/src/atomic_histogram.rs
@@ -1,0 +1,181 @@
+// Copyright (c) The Libra Core Contributors
+// SPDX-License-Identifier: Apache-2.0
+
+#![forbid(unsafe_code)]
+
+use std::{
+    fmt::Debug,
+    ops::Sub,
+    sync::atomic::{AtomicU64, Ordering},
+};
+
+const DEFAULT_HISTOGRAM_CAPACITY: usize = 1024;
+const DEFAULT_HISTOGRAM_STEP_WIDTH: u64 = 50;
+
+pub struct AtomicHistogramAccumulator {
+    capacity: usize,
+    step_width: u64,
+    buckets: Vec<AtomicU64>,
+}
+
+impl Default for AtomicHistogramAccumulator {
+    fn default() -> AtomicHistogramAccumulator {
+        AtomicHistogramAccumulator::new(DEFAULT_HISTOGRAM_CAPACITY, DEFAULT_HISTOGRAM_STEP_WIDTH)
+    }
+}
+
+impl AtomicHistogramAccumulator {
+    pub fn new(size: usize, step: u64) -> AtomicHistogramAccumulator {
+        let mut buf = Vec::with_capacity(size);
+        for _i in 0..size {
+            buf.push(AtomicU64::new(0));
+        }
+        Self {
+            capacity: size,
+            step_width: step,
+            buckets: buf,
+        }
+    }
+
+    pub fn snapshot(&self) -> AtomicHistogramSnapshot {
+        let mut buf = Vec::with_capacity(self.capacity);
+        for i in 0..self.capacity {
+            buf.push(self.buckets[i].load(Ordering::Relaxed));
+        }
+        AtomicHistogramSnapshot {
+            capacity: self.capacity,
+            step_width: self.step_width,
+            buckets: buf,
+        }
+    }
+
+    fn get_bucket_num(&self, data_value: u64) -> u64 {
+        let bucket_num = data_value / self.step_width;
+        if bucket_num >= self.capacity as u64 - 2 {
+            return self.capacity as u64 - 1;
+        }
+        bucket_num
+    }
+
+    pub fn record_data_point(&self, data_value: u64, data_num: u64) {
+        let bucket_num = self.get_bucket_num(data_value);
+        self.buckets[bucket_num as usize].fetch_add(data_num as u64, Ordering::Relaxed);
+    }
+}
+
+#[derive(Debug)]
+pub struct AtomicHistogramSnapshot {
+    capacity: usize,
+    step_width: u64,
+    buckets: Vec<u64>,
+}
+
+impl Default for AtomicHistogramSnapshot {
+    fn default() -> AtomicHistogramSnapshot {
+        AtomicHistogramAccumulator::default().snapshot()
+    }
+}
+
+impl Sub for &AtomicHistogramSnapshot {
+    type Output = AtomicHistogramSnapshot;
+
+    fn sub(self, other: &AtomicHistogramSnapshot) -> AtomicHistogramSnapshot {
+        assert_eq!(
+            self.buckets.len(),
+            other.buckets.len(),
+            "Histogram snapshots must have same size, prev: {}, cur: {}",
+            self.buckets.len(),
+            other.buckets.len()
+        );
+        let mut buf = Vec::with_capacity(self.capacity);
+        for i in 0..self.buckets.len() {
+            buf.push(self.buckets[i] - other.buckets[i]);
+        }
+        AtomicHistogramSnapshot {
+            capacity: self.capacity,
+            step_width: self.step_width,
+            buckets: buf,
+        }
+    }
+}
+
+impl AtomicHistogramSnapshot {
+    pub fn percentile(&self, numerator: u64, denominator: u64) -> u64 {
+        let committed: u64 = self.buckets.iter().sum();
+        let p_count = committed * numerator / denominator;
+        let mut counter = 0u64;
+        for i in 0..self.buckets.len() {
+            counter += self.buckets[i];
+            if counter >= p_count {
+                return i as u64 * self.step_width;
+            }
+        }
+        unreachable!()
+    }
+}
+
+#[cfg(test)]
+mod test {
+    use super::*;
+    use crate::tx_emitter::*;
+    // other use:: for test only
+
+    #[test]
+    pub fn test_default_atomic_histogram() {
+        let histogram = AtomicHistogramAccumulator::default();
+        assert_eq!(histogram.step_width, DEFAULT_HISTOGRAM_STEP_WIDTH);
+        assert_eq!(histogram.buckets.len(), DEFAULT_HISTOGRAM_CAPACITY);
+    }
+
+    #[test]
+    pub fn test_get_bucket_num() {
+        let histogram = AtomicHistogramAccumulator::default();
+        assert_eq!(histogram.get_bucket_num(0), 0);
+        assert_eq!(histogram.get_bucket_num(49), 0);
+        assert_eq!(histogram.get_bucket_num(50), 1);
+        assert_eq!(histogram.get_bucket_num(51), 1);
+        assert_eq!(histogram.get_bucket_num(200_000), 1023);
+    }
+
+    #[test]
+    pub fn test_sub() {
+        let mut cur_snap = AtomicHistogramSnapshot::default();
+        let mut cur_vec = Vec::new();
+        for i in 10..20 {
+            cur_vec.push(i);
+        }
+        cur_snap.buckets = cur_vec;
+
+        let mut pre_snap = AtomicHistogramSnapshot::default();
+        let mut prev_vec = Vec::new();
+        for i in 0..10 {
+            prev_vec.push(i);
+        }
+        pre_snap.buckets = prev_vec;
+        let res = &cur_snap - &pre_snap;
+        for &i in res.buckets.iter() {
+            assert_eq!(i, 10);
+        }
+    }
+
+    #[test]
+    pub fn test_percentile_latency() {
+        let histogram = AtomicHistogramAccumulator::default();
+        // set 10 commits, with latencies as:
+        // 100ms, 200ms, 300ms ... 900ms, 1000ms
+        // for p90 count is 9
+        // expected p90 = 900
+        for i in 1..11 {
+            histogram.record_data_point(i as u64 * 100, 1);
+        }
+        let stat = TxStats {
+            submitted: 0,
+            committed: 10,
+            expired: 0,
+            latency: 0,
+            latency_buckets: histogram.snapshot(),
+        };
+        let res = stat.latency_buckets.percentile(9, 10);
+        assert_eq!(res, 900);
+    }
+}
--- a/testsuite/cluster-test/src/tx_emitter.rs
+++ b/testsuite/cluster-test/src/tx_emitter.rs
@@ -3,7 +3,7 @@
 
 #![forbid(unsafe_code)]
 
-use crate::{cluster::Cluster, instance::Instance};
+use crate::{atomic_histogram::*, cluster::Cluster, instance::Instance};
 use std::{
     fmt, slice,
     sync::Arc,
@@ -64,6 +64,7 @@ struct StatsAccumulator {
     committed: AtomicU64,
     expired: AtomicU64,
     latency: AtomicU64,
+    latencies: Arc<AtomicHistogramAccumulator>,
 }
 
 #[derive(Debug, Default)]
@@ -72,6 +73,7 @@ pub struct TxStats {
     pub committed: u64,
     pub expired: u64,
     pub latency: u64,
+    pub latency_buckets: AtomicHistogramSnapshot,
 }
 
 #[derive(Debug, Default)]
@@ -80,6 +82,7 @@ pub struct TxStatsRate {
     pub committed: u64,
     pub expired: u64,
     pub latency: u64,
+    pub p90_latency: u64,
 }
 
 #[derive(Clone)]
@@ -389,6 +392,7 @@ impl SubmissionWorker {
                 {
                     let end_time = (Instant::now() - start_time).as_millis() as u64;
                     let num_committed = (num_requests - uncommitted.len()) as u64;
+                    let latency = end_time - tx_offset_time / num_requests as u64;
                     self.stats
                         .committed
                         .fetch_add(num_committed, Ordering::Relaxed);
@@ -399,22 +403,30 @@ impl SubmissionWorker {
                         // To avoid negative result caused by uncommitted tx occur
                         // Simplified from:
                         // end_time * num_committed - (tx_offset_time/num_requests) * num_committed
-                        (end_time - tx_offset_time / num_requests as u64) * num_committed as u64,
+                        // to
+                        // (end_time - tx_offset_time / num_requests) * num_committed
+                        latency * num_committed as u64,
                         Ordering::Relaxed,
                     );
+                    self.stats
+                        .latencies
+                        .record_data_point(latency, num_committed);
                     info!(
                         "[{:?}] Transactions were not committed before expiration: {:?}",
                         self.client, uncommitted
                     );
                 } else {
                     let end_time = (Instant::now() - start_time).as_millis() as u64;
+                    let latency = end_time - tx_offset_time / num_requests as u64;
                     self.stats
                         .committed
                         .fetch_add(num_requests as u64, Ordering::Relaxed);
-                    self.stats.latency.fetch_add(
-                        end_time * num_requests as u64 - tx_offset_time,
-                        Ordering::Relaxed,
-                    );
+                    self.stats
+                        .latency
+                        .fetch_add(latency * num_requests as u64, Ordering::Relaxed);
+                    self.stats
+                        .latencies
+                        .record_data_point(latency, num_requests as u64);
                 }
             }
         }
@@ -740,6 +752,7 @@ impl StatsAccumulator {
             committed: self.committed.load(Ordering::Relaxed),
             expired: self.expired.load(Ordering::Relaxed),
             latency: self.latency.load(Ordering::Relaxed),
+            latency_buckets: self.latencies.snapshot(),
         }
     }
 }
@@ -751,6 +764,7 @@ impl TxStats {
             committed: self.committed / window.as_secs(),
             expired: self.expired / window.as_secs(),
             latency: self.latency / self.committed,
+            p90_latency: self.latency_buckets.percentile(9, 10),
         }
     }
 }
@@ -764,6 +778,7 @@ impl Sub for &TxStats {
             committed: self.committed - other.committed,
             expired: self.expired - other.expired,
             latency: self.latency - other.latency,
+            latency_buckets: &self.latency_buckets - &other.latency_buckets,
         }
     }
 }
@@ -773,7 +788,7 @@ impl fmt::Display for TxStats {
         write!(
             f,
             "submitted: {}, committed: {}, expired: {}",
-            self.submitted, self.committed, self.expired
+            self.submitted, self.committed, self.expired,
         )
     }
 }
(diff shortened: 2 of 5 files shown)


snakech:
implemented client side statistic for P90(90% percentile) latency, prints out like below:

for swarm tests:
```
INFO 2020-05-28 04:16:34 testsuite/cluster-test/src/tx_emitter.rs:260 Tx emitter workers started
submitted: 13 txn/s, committed: 12 txn/s, expired: 0 txn/s, latency: 1174 ms
submitted: 12 txn/s, committed: 12 txn/s, expired: 0 txn/s, latency: 1136 ms
submitted: 13 txn/s, committed: 15 txn/s, expired: 0 txn/s, latency: 1001 ms
submitted: 13 txn/s, committed: 12 txn/s, expired: 0 txn/s, latency: 1054 ms
submitted: 15 txn/s, committed: 15 txn/s, expired: 0 txn/s, latency: 1010 ms
submitted: 13 txn/s, committed: 13 txn/s, expired: 0 txn/s, latency: 1013 ms
Total stats: submitted: 810, committed: 810, expired: 0, p90 latency: 1013
Average rate: submitted: 13 txn/s, committed: 13 txn/s, expired: 0 txn/s, latency: 1060 ms
```
for performance tests:
```
====json-report-end===
INFO 2020-05-28 03:57:15 testsuite/cluster-test/src/aws.rs:23 Scaling to min_desired_capacity : 0, buffer: 0, asg_name: ct-1-k8s-testnet-validators
all up : 1305 TPS, 3458 ms latency, 3404 ms p90 latency, no expired txns
10% down : 1279 TPS, 3174 ms latency, 3091 ms p90 latency, no expired txns
3 Region Simulation : 880 TPS, 4636 ms latency, 4486 ms p90 latency
**********
```

andll:
I see that p90 latency is lower than avg latency? Is this correct? How can this be?

snakech:
> I see that p90 latency is lower than avg latency? Is this correct? How can this be?

I have thought about this point, it has two reason:

1. for the first bucket, like 0 - 50ms, I count it as 0, so p90 latency is a kind of shifting one slot.
2. for average latency, we average total committed txn latency, for p90, we do not count the end 10%, if the last 10% are all outliner, p90 should be expected lower than average?

testsuite/cluster-test/src/tx_emitter.rs
```diff
@@ -67,12 +71,50 @@ struct StatsAccumulator {
     latency: AtomicU64,
 }
 
+pub struct AtomicHistogram {
+    buckets: Box<[AtomicU64]>,
```
andll, line 75: I think you should be able to just use `Vec<AtomicU64>` instead of `Box<[AtomicU64]>` here.

testsuite/cluster-test/src/tx_emitter.rs
```diff
@@ -57,6 +60,7 @@ pub struct EmitJob {
     workers: Vec<Worker>,
     stop: Arc<AtomicBool>,
     stats: Arc<StatsAccumulator>,
+    latencies: Arc<Mutex<AtomicHistogram>>,
```
andll, line 63: Are you sure you need Mutex here? Should be ok without it, and Mutex kind of dangerous here as it's a high rate place

snakech: I think we need to have a Mutex here to make consistency/atomic to read/write on this array

andll: But you don't write to this array (updating atomicu64 takes immutable reference for updating).
What happens if you remove mutex and use Vec?

snakech: I think it should also works, I can remove mutex

testsuite/cluster-test/src/tx_emitter.rs
```diff
@@ -313,7 +352,7 @@ impl TxEmitter {
                 .expect("TxEmitter worker thread failed");
             self.accounts.append(&mut accounts);
         }
-        job.stats.accumulate()
+        job.stats.accumulate((*job.latencies).clone())
```
andll, line 355: Why do you need to clone it? Why not just pass `&AtomicHistogram`?

testsuite/cluster-test/src/tx_emitter.rs
```diff
@@ -765,6 +845,7 @@ impl Sub for &TxStats {
             committed: self.committed - other.committed,
             expired: self.expired - other.expired,
             latency: self.latency - other.latency,
+            latency_buckets: AtomicHistogram::default().to_u64_array(),
```
andll, line 848: You can in theory substract histograms too, and therefore get p90 on each period, not just overall p90?

testsuite/cluster-test/src/tx_emitter.rs
```diff
@@ -67,12 +68,46 @@ struct StatsAccumulator {
...
+        let mut buf = Vec::with_capacity(HISTOGRAM_CAPACITY);
+        for _i in 0..HISTOGRAM_CAPACITY {
+            buf.push(AtomicU64::new(0));
+        }
+        Self { buckets: buf }
+    }
+
+    fn to_u64_array(&self) -> Vec<u64> {
+        let mut buf = Vec::with_capacity(HISTOGRAM_CAPACITY);
+        for _i in 0..HISTOGRAM_CAPACITY {
```
andll, line 86: This should be named `i`, not `_i` because this var is used

testsuite/cluster-test/src/tx_emitter.rs
```diff
@@ -67,12 +68,46 @@ struct StatsAccumulator {
...
+impl AtomicHistogram {
+    fn default() -> Self {
+        let mut buf = Vec::with_capacity(HISTOGRAM_CAPACITY);
+        for _i in 0..HISTOGRAM_CAPACITY {
+            buf.push(AtomicU64::new(0));
+        }
+        Self { buckets: buf }
+    }
+
+    fn to_u64_array(&self) -> Vec<u64> {
```
andll, line 84: `to_vec` would me more in line with how other methods in rust named

testsuite/cluster-test/src/tx_emitter.rs
```diff
@@ -67,12 +68,46 @@ struct StatsAccumulator {
...
+        let mut buf = Vec::with_capacity(HISTOGRAM_CAPACITY);
+        for _i in 0..HISTOGRAM_CAPACITY {
+            buf.push(self.buckets[_i].load(Ordering::Relaxed));
+        }
+
+        buf
+    }
+}
+
+impl Clone for AtomicHistogram {
```
andll, line 94: I think we should be able to make implementation of this without need to have Clone for histogram

testsuite/cluster-test/src/tx_emitter.rs
```diff
@@ -400,22 +441,34 @@ impl SubmissionWorker {
...
                         // Simplified from:
                         // end_time * num_committed - (tx_offset_time/num_requests) * num_committed
-                        (end_time - tx_offset_time / num_requests as u64) * num_committed as u64,
+                        // to
+                        // (end_time - tx_offset_time / num_requests) * num_committed
+                        latency * num_committed as u64,
                         Ordering::Relaxed,
                     );
+                    let latency_bucket_num = get_bucket_num(latency);
+                    {
```
andll, line 450: Why is this line wrapped as a separate block with `{}`?

testsuite/cluster-test/src/tx_emitter.rs
```diff
@@ -400,22 +441,34 @@ impl SubmissionWorker {
...
+                        self.latencies.buckets[latency_bucket_num as usize]
```
andll, line 451: Probably makes sense to make `AtomicHistogram` a self contained struct (perhaps move it out of this file to some new file like `histogram.rs`.
And then AtomicHistogram should have some method like 'record_latency(latency_ms)` and buckets should not be exposed directly?

testsuite/cluster-test/src/tx_emitter.rs
snakech: I can do subtract P90, I thought P90 is only for overall

testsuite/cluster-test/src/tx_emitter.rs
snakech: aha, forgot remove this with Mutex, which was using to limit mutex scope

testsuite/cluster-test/src/atomic_histogram.rs
```diff
@@ -0,0 +1,84 @@
...
+
+const HISTOGRAM_CAPACITY: usize = 2048;
+const HISTOGRAM_WIDTH_MILLIS: u64 = 50;
+
+pub struct AtomicHistogram {
+    buckets: Vec<AtomicU64>,
+}
+
+impl AtomicHistogram {
+    pub fn default() -> Self {
```
andll, line 17: Note that usually you want to implement trait `Default`, instead of just having method with this name (can be useful in other case)

testsuite/cluster-test/src/atomic_histogram.rs
```diff
@@ -0,0 +1,84 @@
...
+
+    fn get_bucket_num(&self, latency: u64) -> u64 {
+        let bucket_num = latency / HISTOGRAM_WIDTH_MILLIS;
+        if bucket_num >= HISTOGRAM_CAPACITY as u64 - 2 {
+            return HISTOGRAM_CAPACITY as u64 - 1;
+        }
+        bucket_num
+    }
+
+    pub fn record_latency_in_bucket(&self, latency: u64, num_requests: u64) {
```
andll, line 42: You may want to rename `latency` -> `latency_ms`
Also `record_latency_in_bucket` -> `record_latency_ms`

testsuite/cluster-test/src/atomic_histogram.rs
```diff
@@ -0,0 +1,84 @@
...
+            }
+        } else {
+            for i in 0..a.len() {
+                buf.push(a[i] - b[i]);
+            }
+        }
+        buf
+    }
+
+    pub fn calculate_p90(stats: &TxStats) -> u64 {
```
andll, line 61: You probably can have generic method here like `percentile(numerator: u64, denominator: u64)` and then you can call it as `.percentile(9, 10)`

testsuite/cluster-test/src/atomic_histogram.rs
```diff
@@ -0,0 +1,84 @@
...
+use std::sync::atomic::{AtomicU64, Ordering};
+
+const HISTOGRAM_CAPACITY: usize = 2048;
+const HISTOGRAM_WIDTH_MILLIS: u64 = 50;
+
+pub struct AtomicHistogram {
+    buckets: Vec<AtomicU64>,
+}
+
+impl AtomicHistogram {
```
andll, line 16: Do you want to add some unit tests for atomic histogram?

testsuite/cluster-test/src/atomic_histogram.rs
```diff
@@ -0,0 +1,84 @@
+// Copyright (c) The Libra Core Contributors
+// SPDX-License-Identifier: Apache-2.0
+
+#![forbid(unsafe_code)]
+
+use crate::tx_emitter::TxStats;
+use std::sync::atomic::{AtomicU64, Ordering};
+
+const HISTOGRAM_CAPACITY: usize = 2048;
```
andll, line 9: We can probably just have both of capacity and width(maybe 'step' is better?) configurable?
Have them as fields on AtomicHistogram struct. Default can read those constants, but for those who want customize it, they could use `new(capacity, size)`.

testsuite/cluster-test/src/tx_emitter.rs
```diff
@@ -57,6 +57,7 @@ pub struct EmitJob {
     workers: Vec<Worker>,
     stop: Arc<AtomicBool>,
     stats: Arc<StatsAccumulator>,
+    latencies: Arc<AtomicHistogram>,
```
andll, line 60: Would it make more sense to have `latencies: AtomicHistogram` as a field on `StatsAccumulator`?

testsuite/cluster-test/src/atomic_histogram.rs
```diff
@@ -0,0 +1,84 @@
...
+                counter += gap;
+            } else {
+                total_latency += num * HISTOGRAM_WIDTH_MILLIS * i as u64;
+                counter += num;
+            }
+        }
+        if p90_count == 0 {
+            return 0;
+        }
+        total_latency / p90_count
```
andll, line 82: This formula here is a bit unusual, are you sure it's correct? Is it some fancy histogram formula with improved precision?

I was thinking about something as simple as
```
for i in 0.. stats.latency_buckets.len() {
count+=stats.latency_buckets[i];
if count > p90_count { return i* HISTOGRAM_WIDTH_MILLIS}
}
```

snakech: I think yes, I construct this formula which works by steps:

1. count p90_count
2. iterate each bucket from low to high(from 50ms to ... n * 50ms), count all txns are fitted in this bucket, sum up totally latency
3. for special case, the last bucket count is more than total p90 count, we can not just add on the whole count of that bucket, I use ```gap``` to count the last piece of p90_count

this is important because for example, with larger txn total number, we will lost more and more accuracy, for worst case we will add extra total latency = (10% * total_txn - 1) * i * HISTOGRAM_WIDTH_MILLIS

andll: I see. Let's cover it with unit tests then :)

testsuite/cluster-test/src/atomic_histogram.rs
```diff
@@ -0,0 +1,182 @@
...
+        }
+        Self {
+            capacity: HISTOGRAM_CAPACITY,
+            step_millis: HISTOGRAM_STEP_MILLIS,
+            buckets: buf,
+        }
+    }
+}
+
+impl AtomicHistogram {
```
andll, line 32: Can you add `fn new(capacity, step_millis)` and call it from `default()`?
Also, can you rename HISTOGRAM_CAPACITY=>DEFAULT_ HISTOGRAM_CAPACITY and same for HISTOGRAM_STEP_MILLIS

testsuite/cluster-test/src/atomic_histogram.rs
```diff
@@ -0,0 +1,182 @@
...
+            }
+        } else {
+            for i in 0..a.len() {
+                buf.push(a[i] - b[i]);
+            }
+        }
+        buf
+    }
+
+    pub fn cal_percentile_latency(numerator: u64, denominator: u64, stats: &TxStats) -> u64 {
```
andll, line 69: I think this api got little bit mixed up.
We should not reference tx_emitter classes from here. Imagine histogram to be as a utility class that anyone can use. Maybe it will get moved outside of cluster test crate even.

I think what you want to do probably is to split this into two classes:
`AtomicHistogramAccumulator` that will hold vec<AtomicU64> and `AtomicHistogramSnapshot` that will hold vec<u64>.
Then existing API will change as
- instead of to_vec, AtomicHistogramAccumulator will have .snapshot() method.
- cal_percentile_latency will be on AtomicHistogramSnapshot
- Instead of `sub(a: &[u64], b: &[u64]) ` would would implement trait `Sub` on AtomicHistogramSnapshot
- In tx_emitter, `StatsAccumulator` will hold AtomicHistogramAccumulator and `TxStats` will hold `AtomicHistogramSnapshot`

testsuite/cluster-test/src/atomic_histogram.rs
```diff
@@ -0,0 +1,232 @@
...
+
+impl Default for AtomicHistogramSnapshot {
+    fn default() -> AtomicHistogramSnapshot {
+        AtomicHistogramAccumulator::new(DEFAULT_HISTOGRAM_CAPACITY, DEFAULT_HISTOGRAM_STEP_MILLIS)
+            .snapshot()
+    }
+}
+
+// reserving place for debug print in the future
+impl Debug for AtomicHistogramSnapshot {
```
andll, line 82: Mm, you can just add `#[dervie(Debug)]` on top of `pub struct AtomicHistogramSnapshot`

testsuite/cluster-test/src/atomic_histogram.rs
```diff
@@ -0,0 +1,232 @@
...
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(f, "",)
+    }
+}
+
+pub fn sub_vec(a: &[u64], b: &[u64]) -> Vec<u64> {
```
andll, line 88: This should not be public function? In fact, maybe it does not need to be separate function at all, can just be code inside AtomicHistogramSnapshot::sub?

Additionally, I think you can just enforce that snapshots should be of same length, e.g. just write `assert!`(a.len==b.len, "Histogram snapshots must have same size, a:{}, b:{}", a.len, b.len)`

testsuite/cluster-test/src/atomic_histogram.rs
```diff
@@ -0,0 +1,232 @@
...
+        AtomicHistogramSnapshot {
+            capacity: self.capacity,
+            step_millis: self.step_millis,
+            buckets: sub_vec(self.buckets.as_ref(), other.buckets.as_ref()),
+        }
+    }
+}
+
+impl AtomicHistogramSnapshot {
+    pub fn cal_percentile_latency(&self, numerator: u64, denominator: u64, stats: &TxStats) -> u64 {
```
andll, line 115: We should not pass TxStats here. Please make sure this file does not use anything from tx_emitter

testsuite/cluster-test/src/atomic_histogram.rs
snakech: we can not use assert here, because the first snapshot sub, we are using NONE to be the prev_snap

andll: Not sure I understand - what is None? Don't we just use AtomicHistogramSnapshot::default?

snakech: in main.rs, we record prev_stats for duration average:
```
        let mut prev_stats: Option<TxStats> = None;
        while Instant::now() < deadline {
            let window = Duration::from_secs(10);
            tokio::time::delay_for(window).await;
            let stats = emitter.peek_job_stats(&job);
            let delta = &stats - &prev_stats.unwrap_or_default();
            prev_stats = Some(stats);
            println!("{}", delta.rate(window));
        }

so for the first duration of stat substract, we do expect the length is 0

Comment (andll) [reply testsuite/cluster-test/src/atomic_histogram.rs:88]: `unwrap_or_default` returns Default, and default is not a zero-size snapshot, it's snapshot of default size filled with zeroes....

Comment (snakech) [reply testsuite/cluster-test/src/atomic_histogram.rs:88]: oh, I c, good to know ```unwrap_or_default``` return default

Comment (andll) [reply testsuite/cluster-test/src/atomic_histogram.rs:88]: Yep (this is also what name suggests :) )
You could not pass None in this function even if you wanted to - argument is `&Snapshot` - you can not pass None with this type (it has to be Option to allow for None)

File: testsuite/cluster-test/src/atomic_histogram.rs (hunk 19 above)
Comment (snakech) [reply testsuite/cluster-test/src/atomic_histogram.rs:82]: I tried, it will get error and ask to implement Debug for ```AtomicHistogramSnapshot```

Comment (andll) [reply testsuite/cluster-test/src/atomic_histogram.rs:82]: Mm, this is hard to believe :)
`#[dervie(Debug)]` is implementing debug for this type, what exactly error message that you saw?

Comment (snakech) [reply testsuite/cluster-test/src/atomic_histogram.rs:82]: ```error[E0277]: `atomic_histogram::AtomicHistogramSnapshot` doesn't implement `std::fmt::Debug`
  --> testsuite/cluster-test/src/tx_emitter.rs:77:5
   |
77 |     pub latency_buckets: AtomicHistogramSnapshot,
   |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ `atomic_histogram::AtomicHistogramSnapshot` cannot be formatted using `{:?}`
```

andll: Did you write `#[dervie(Debug)]` on top of `pub struct AtomicHistogramSnapshot` when you had this error message?
If yes, do you mind pushing version with this to github so we look together

snakech: sorry for the late reply, I think I figured out that I put it on top of wrong place. ...my bad

testsuite/cluster-test/src/atomic_histogram.rs
```diff
@@ -0,0 +1,230 @@
...
+        }
+    }
+
+    #[test]
+    pub fn test_percentile_latency() {
+        let histogram = AtomicHistogramAccumulator::default();
+        // set 10 commits, with latency as:
+        // 100ms, 200ms, 300ms ... 1000ms
+        // for p90 count is 9
+        // expected p90 = (100 + 200 + ... + 900) / 9 = 500
```
andll, line 184: So this is not what p90 is actually :)
For this test case, p90 is 900ms:
90% of transactions has faster or equal top90
10% of transactions are slower then p90

testsuite/cluster-test/src/atomic_histogram.rs
```diff
@@ -0,0 +1,191 @@
...
+        }
+    }
+}
+
+impl AtomicHistogramSnapshot {
+    pub fn cal_percentile_latency(
+        &self,
+        numerator: u64,
+        denominator: u64,
+        committed: u64,
```
andll, line 107: Do you want to just calculate this number by summing across all `snapshot.buckets`?

testsuite/cluster-test/src/atomic_histogram.rs
```diff
@@ -0,0 +1,191 @@
...
+        snapshot: &AtomicHistogramSnapshot,
```
andll, line 108: Why do you pass `snapshot` here? What is difference between `snapshot` and `self`?

testsuite/cluster-test/src/atomic_histogram.rs
snakech: if we can pass in the totally number which already have from outside, I prefer to save one operation here?

testsuite/cluster-test/src/atomic_histogram.rs
snakech: you are correct, will remove

testsuite/cluster-test/src/atomic_histogram.rs
andll: I understand preference to save some time, but really more convenient API outweights perf in this case, in my opinion.

snakech: for generic usage, yes, I agree, will update

snakech: I may also change some fn name such as ```record_latency_ms``` to a generic name like ```record_data_point```

testsuite/cluster-test/src/atomic_histogram.rs
```diff
@@ -0,0 +1,188 @@
...
+        AtomicHistogramSnapshot {
+            capacity: self.capacity,
+            step_width: self.step_width,
+            buckets: buf,
+        }
+    }
+}
+
+impl AtomicHistogramSnapshot {
+    pub fn cal_percentile_latency(
```
andll, line 103: nit: rename it to `percentile`?

testsuite/cluster-test/src/atomic_histogram.rs
```diff
@@ -0,0 +1,188 @@
...
+        let committed: u64 = self.buckets.iter().sum();
+        let p_count = committed * numerator / denominator;
+        let mut counter = 0u64;
+        for i in 0..self.buckets.len() {
+            counter += self.buckets[i];
+            if counter >= p_count {
+                return i as u64 * self.step_width;
+            }
+        }
+        0
```
andll, line 117: Should you use `unreachable!()` here? It should not be possible to enter this condition(and if it is possible, 0 is certainly not a valid return here)

PR #4101, opened 2020-05-28
[client] Fix bad argument index's in client_proxy. Update mint name in json-rpc views

tzakian:
Invalid argument index's meant that you couldn't add a currency while specifying the gas currency. This fixes this so you can now do 
```
a addc 0 LBR 0 100000 Coin1
```

assuming account 0 has a Coin1 balance, and no LBR balance.

This also updates the script names for minting so that the JSON-RPC views report this correctly. This should be switched over to not be a string comparison, but instead grab the script from the hash and then check that. But that's for a later PR.

### DIFF
--- a/json-rpc/types/src/views.rs
+++ b/json-rpc/types/src/views.rs
@@ -403,7 +403,7 @@ impl From<TransactionPayload> for ScriptView {
                     Err(format_err!("Unable to parse PeerToPeer arguments"))
                 }
             }
-            "mint_transaction" => {
+            "mint" | "mint_lbr_to_address" => {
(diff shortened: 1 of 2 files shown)


json-rpc/types/src/views.rs
```diff
@@ -403,7 +403,7 @@ impl From<TransactionPayload> for ScriptView {
                     Err(format_err!("Unable to parse PeerToPeer arguments"))
                 }
             }
-            "mint_transaction" => {
+            "mint" | "mint_lbr_to_address" => {
```
thefallentree, line 406: is this right though?  did someone change get_transcation_name() function ?

PR #4104, opened 2020-05-28
[consensus] fix potential crash loop after restart

zekun000:
## Motivation

We have a reconfiguration hack in execute_block (rollover the parent's result) where we don't follow in build_block_tree
where we execute normally and would crash because the result won't match QC.

This PR unifies how we handle the insertion and makes it hopefully less error prone.

Y

## Test Plan

Existing tests.

## Related PRs


zekun000:
@dmitri-perelman see these lines https://github.com/libra/libra/blob/master/consensus/src/block_storage/block_store.rs#L304-L322

if a block is a child of reconfiguration we can't change any state

PR #4105, opened 2020-05-28
[move-prover] Fixed a bug in translation of parameters in asserts/assumes

shazqadeer:
## Motivation

This PR does the following:
- Input parameters of a Move function are mutable but input parameters in a Boogie procedure used to model a Move function are not.  To handle mutable input parameters, the input must be copied to temporary local at the Boogie level and in asserts/assumes appropriate substitution must be performed.
- Added more test cases for spec blocks inside functions including loops.
- Renamed two test files to follow the naming convention for files.

Yes

## Test Plan

Existing tests and a new test


shazqadeer:
> Can't see the new loop test cases. Am I missing something or did you forget to add a file?

Indeed I had forgotten to check in the new test file.  I just did that.

PR #4107, opened 2020-05-29
Block Executor with Strategies

tzakian:
This PR proposes a block executor interface that allows user-defined strategies for testing out different block packings of transactions. You can see example strategy definitions in `basic_strategy`, `guided_strategy`, and `random_strategy`.

The basic premise of this is that it takes in a block and executes that block. From the caller's perspective, this is no different from executing a normal block.

However, it can also segment that block into different sub-blocks based upon the strategy used in order to provide different segmentations of that same block (given by the `schedule` function). 

You can see examples of how to use this in the new test, and examples strategies under `execution_strategies`.

### DIFF
--- a/language/e2e-tests/src/execution_strategies/basic_strategy.rs
+++ b/language/e2e-tests/src/execution_strategies/basic_strategy.rs
@@ -1,0 +1,54 @@
+// Copyright (c) The Libra Core Contributors
+// SPDX-License-Identifier: Apache-2.0
+
+#![forbid(unsafe_code)]
+
+use crate::{
+    execution_strategies::types::{Block, Executor, ExecutorResult, PartitionStrategy},
+    executor::FakeExecutor,
+};
+use libra_types::{transaction::SignedTransaction, vm_error::VMStatus};
+
+#[derive(Debug, Clone)]
+pub struct BasicStrategy;
+
+impl PartitionStrategy for BasicStrategy {
+    type Txn = SignedTransaction;
+    fn partition(&mut self, block: Block<Self::Txn>) -> Vec<Block<SignedTransaction>> {
+        vec![block]
+    }
+}
+
+#[derive(Debug)]
+pub struct BasicExecutor {
+    executor: FakeExecutor,
+    strategy: BasicStrategy,
+}
+
+impl Default for BasicExecutor {
+    fn default() -> Self {
+        Self::new()
+    }
+}
+
+impl BasicExecutor {
+    pub fn new() -> Self {
+        Self {
+            executor: FakeExecutor::from_genesis_file(),
+            strategy: BasicStrategy,
+        }
+    }
+}
+
+impl Executor for BasicExecutor {
+    type Txn = <BasicStrategy as PartitionStrategy>::Txn;
+    type BlockResult = VMStatus;
+    fn execute_block(&mut self, txns: Block<Self::Txn>) -> ExecutorResult<Self::BlockResult> {
+        let mut block = self.strategy.partition(txns);
+        let outputs = self.executor.execute_block(block.remove(0))?;
+        for output in &outputs {
+            self.executor.apply_write_set(output.write_set())
+        }
+        Ok(outputs)
+    }
+}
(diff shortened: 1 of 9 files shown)


vgao1996:
Sounds like a great idea to me! Had a brief look and I would suggest remodeling the strategies in the following way:
```
// Note: this is psudocode

// The high level executor trait.
// The protocol is simple: you take a bunch of transactions, execute them in whatever way you like 
// as long as the order is preserved and return the result.
trait Executor {
    fn execute_transactions(&mut self, txns: Vec<Transaction>) -> Result<Vec<Result<TransactionOutput>>>;
}

// A strategy to partition transactions into blocks.
trait PartitioningStrategy {
    fn partition(&mut self, txns: Vec<Transaction>) -> Vec<Vec<Transaction>>;
}



// Wrapper struct to avoid generic impl<T> Executor <T> and potential conflicting impls.
struct PartitioningExecutor<S> { s: S }

impl<S: PartitioningStrategy> PartitioningExecutor<S> {
    fn new(s: S) -> Self {
        Self { s }
    }
}

impl<S: PartitioningStrategy> Executor for PartitioningExecutor<S> {
    fn execute_transactions(&mut self, txns: Vec<Transaction>) -> Result<Vec<Result<TransactionOutput>>> {
        let blocks = self.s.partition(txns);
        execute the blocks and return the results
    }
}


// Defining an executor that puts all transactions into a single block
struct SingleBlockStrategy {}

impl PartitioningStrategy for SingleBlockStrategy {
    ...
}

type SingleBlockExecutor = PartitioningExecutor<SingleBlockStrategy>;


// Defining an executor that puts each transaction into a single block
struct OneTransactionPerBlockStrategy {}

impl PartitioningStrategy for OneTransactionPerBlockStrategy {
    ...
}

type OneTransactionPerBlockExecutor 
    = PartitioningExecutor<OneTransactionPerBlockStrategy>;
    
    
// Defining an executor that randomly partitions transactions into blocks
struct RandomizedStrategy {
    rng: SomeRng,
}

impl PartitioningStrategy for RandomizedStrategy {
    ...
}

type RandomizedExecutor = PartitioningExecutor<RandomizedStrategy>;



// An executor built by combining multiple executors.
// It will run the same transactions through all executors and return the results if they are all consistent, or err otherwise
struct ComposedExecutor {
    execs: Vec<Box<dyn Executor>>,
}

impl Executor for ComposedExecutor {
    fn execute_transactions(&mut self, txns: Vec<Transaction>) -> Result<Vec<Result<TransactionOutput>>> {
        ...
    }
}
```
I think this will give us more flexibility and make the code cleaner as you do not have to reason about concepts like workers or combining the results from multiple executors when you don't have to.

dariorussi:
just out of curiosity why not use proptest for this?
It should be relatively trivial is it? 
it should be a couple of numbers to define blocks and transaction within a block. We can then pick randomly for the number of blocks and randomly for each block to define the number of transactions. Finally we fill each block from the set of allowed transactions (which can be anything) with possibly some distribution across them.
Do we need this?

tzakian:
Good question! One primary reason why I don't think proptest is the best suited here is because we don't just want to pick random block packings, or have them "algorithmically" defined, but also allow "caller-guided" strategies. As an example see the `guided_strategy` which allows the caller to define how the block should be split up. This should be useful for the future work in functional tests where you want to specify precise settings for which transactions are in a block.

dariorussi:
don't know, it feels all of that could be accomplished via "normal" proptest and strategies defined there. Maybe at the end, you end up with something similar but I really think we should look into it and see if we get some stuff for free and be in a better place... though you may have looked at that already...

tzakian:
Your comments made me step back and think a bit a more. I think I agree that proptest could be useful here, and I think it can be used for certain, but not all strategies (e.g. `guided_strategy`), and for some strategies proptest doesn't really make sense (e.g. `basic_strategy`). But, strategies aren't limited to being proptests, nor are they limited to being non-proptests in this setting.

In particular, since both `PartitionStratey` and `Execution` are traits, the underlying implementation can be using proptest or some other mechanism. The strategies that are defined so far are meant as examples of how to implement and use a strategy, and I'll look at implementing a couple with proptest.

PR #4114, opened 2020-05-29
[key-manager] do not use is_ok in test

davidiw:
When an is_ok fails, you are left knowing it is not ok. Then you have to
change the test to unwrap the error. Then you know why it is not ok.

TL;DR, is ok is not ok in  test ;)

A little bit of humor for you @JoshLind


davidiw:
I know it does, so I tried to make the PR funny to try to raise your spirits!

davidiw:
Though honestly, assert!(T::is_ok()) versus T::unwrap() mean the same thing to me. Maybe it is an acquired taste?

JoshLind:
Yeah, perhaps it is an acquired taste 😄

Issue #2519, opened 2020-02-11
Libra Execution Correctness (LEC)


lightmark:
# LEC Design

## Motivation
This issue describes the plan to build Libra Execution Correctness TCB for the Exec/VM which is a building block of Libra Core TCB presented in issue #2326. The goal of TCB is to isolate critical properties in a finite space that collectively ensures the correctness of the Libra blockchain. As a crucial module to TCB, LEC has to guarantee the correctness of execution - All agreed upon transactions within a block execute in order and correctly (Always execute transactions correctly). The current implementation of Executor is not isolated from the rest modules of Libra Core so It has to be refactored.

### Assumptions
- TCB only requires Execution Correctness of blocks but doesn’t trust storage so committing is not its concern. Therefore LEC in TCB framework does not necessitate functionalities related to chunks in state sync procedure or committing any execution results to storage.
- The architecture of full nodes is simplified over validators in the sense that full nodes don’t need TCB. However, they should be able to monitor the progress of the blockchain and to catch up via some mechanism w/o TCB involved. 
- Both full nodes and validators will have a single storage.
- As the figure shows, LSR, LEC and other non-TCB components are isolated into separate processes, or even containers. Also, it is only LSR that is allowed to talk to LEC.
### Implications
- Anything functionalities in executor related to state_sync and committing will be in non-TCB part but that related to block execution will be the LEC.
- Full nodes can fully rely on the non-TCB executor to be functional.
- The validator will have two executors which must share a consistent state with one single storage.
Separated containers implies any shared-memory data exchange is not practical. The crash-recovery process has to be comprehensive since Consensus, LSR and LEC must be consistent in terms of multiple pieces of data, including block tree structure, latest committed state, etc.  The isolation turns them into individual micro services communicating through serializable messages, which adds complications onto the consistency between them because any one can crash at any time but the others are not aware of the crashing. The good news is, from the perspective of LEC, its state is in memory and easily recoverable from Consensus/LSR. A intuitive, simple but not optimized approach is just reconstruct the whole block tree if a block is not found. While the other executor in non-TCB is stateless so it is less concerned with.

![LEC](https://user-images.githubusercontent.com/1549573/75594827-0580fb80-5a3f-11ea-846c-4a9e818cb4a0.png)

## Plan
The introduction of LEC requires the factoring of multiple components of the Libra Core. To to specific, it entails multiple sub-projects:
- Breaking Executor into LEC and SyncExecutor(chunk execution and committing only used by StateSynchronizer)
[...]

PR #4118, opened 2020-05-29
[LEC] build the authenticated link between LSR and LEC

lightmark:
## Motivation

This PR setup the authenticated link between LSR and LEC. The basic rationale behind
this PR is the architecture of TCB https://github.com/libra/libra/issues/2519. LEC is in zone 2, safety-rules is in zone 1 and
consensus is in zone 3. Because LSR doesn't trust consensus, the computation result
from LEC cached in consensus should be authenticated by LSR to verify the integrity
to avoid any tampering from any third party including consensus.

To do this, we add a key pair `EXECUTION_KEY` into vault whose private key is only
accessible by LEC. Any state_compute_result returned from LEC will carry a signature
of a proposal associated with this block. When this block needs to be voted, the generated
vote proposal will be verified by LSR service using the public key.

Y

## Test Plan

CI


config/management/src/smoke_test.rs
```diff
@@ -137,7 +137,9 @@ fn smoke_test() {
         storage_config.path = to;
         storage_config.set_data_dir(PathBuf::from(""));
         storage_config.namespace = Some(ns);
-        config.consensus.safety_rules.backend = SecureBackend::OnDiskStorage(storage_config);
+        config.consensus.safety_rules.backend =
+            SecureBackend::OnDiskStorage(storage_config.clone());
+        config.execution.backend = SecureBackend::OnDiskStorage(storage_config.clone());
```
davidiw, line 142: shouldn't need clone twice, no?

config/src/config/test_config.rs
```diff
@@ -90,6 +97,11 @@ impl TestConfig {
         self.consensus_keypair = Some(ConsensusKeyPair::load(privkey));
     }
 
+    pub fn random_execution_key(&mut self, rng: &mut StdRng) {
```
davidiw, line 100: this will need a rebase, the API has changed recently :O

consensus/consensus-types/src/vote_proposal.rs
```diff
@@ -21,6 +23,17 @@ pub struct VoteProposal<T> {
     next_epoch_state: Option<EpochState>,
 }
 
+impl<T: Serialize> CryptoHash for VoteProposal<T> {
+    type Hasher = VoteProposalHasher;
+
+    fn hash(&self) -> HashValue {
+        let mut state = Self::Hasher::default();
+        lcs::serialize_into(&mut state, &self)
+            .expect("LCS serialization of VoteProposal should not fail");
```
davidiw, line 32: @matbd, serialization should not fail!

consensus/safety-rules/benches/safety_rules.rs
```diff
@@ -15,7 +16,12 @@ use tempfile::NamedTempFile;
 
 /// Execute an in order series of blocks (0 <- 1 <- 2 <- 3 and commit 0 and continue to rotate
 /// left, appending new blocks on the right, committing the left most block
-fn lsr(mut safety_rules: Box<dyn TSafetyRules<Vec<u8>>>, signer: ValidatorSigner, n: u64) {
+fn lsr(
+    mut safety_rules: Box<dyn TSafetyRules<Vec<u8>>>,
+    signer: ValidatorSigner,
+    n: u64,
+    prikey: Ed25519PrivateKey,
```
davidiw, line 23: prikey? executor_key?

consensus/safety-rules/benches/safety_rules.rs
```diff
@@ -55,52 +69,56 @@ fn lsr(mut safety_rules: Box<dyn TSafetyRules<Vec<u8>>>, signer: ValidatorSigner
...
+    lsr(safety_rules_manager.client(), signer, n, prikey);
 }
 
 fn thread(n: u64) {
     let signer = ValidatorSigner::from_int(0);
     let file_path = NamedTempFile::new().unwrap().into_temp_path().to_path_buf();
     let waypoint = test_utils::validator_signers_to_waypoints(&[&signer]);
+    let (prikey, pubkey) = test_utils::test_execution_keypair();
     let storage = PersistentSafetyStorage::initialize(
         Box::new(OnDiskStorage::new(file_path)),
```
davidiw, line 116: it seems like this execution keypair should be pushed into initialize and not into the other apis

consensus/safety-rules/src/local.rs
```diff
@@ -0,0 +1,91 @@
+// Copyright (c) The Libra Core Contributors
```
davidiw, line 1: You have a single commit in which you moved a file but didn't explain why you did. That makes it really hard to review. Ideally you would move this in one commit and modify it in the next. Perhaps you could do that during the rebase?

consensus/safety-rules/src/local.rs
```diff
@@ -0,0 +1,91 @@
...
+            .initialize(proof)
+    }
+
+    fn update(&mut self, qc: &QuorumCert) -> Result<(), Error> {
+        self.internal.write().unwrap().safety_rules.update(qc)
+    }
+
+    fn construct_and_sign_vote(
+        &mut self,
+        vote_proposal: &VoteProposal<T>,
```
davidiw, line 67: signature should be within vote proposal
it should be optional
we should configure whether or not we expect signatures via the safety rules config and executor config respectively

consensus/safety-rules/src/local.rs
davidiw: Alternatively, don't move this file.

consensus/safety-rules/src/process.rs
```diff
@@ -18,7 +19,8 @@ pub struct Process {
 
 impl Process {
     pub fn new(mut config: NodeConfig) -> Self {
-        let (author, storage) = safety_rules_manager::extract_service_inputs(&mut config);
+        let (author, storage, execution_pubkey) =
+            safety_rules_manager::extract_service_inputs(&mut config);
```
davidiw, line 23: we should not be receiving an execution pub key, we should retrieve it from storage within safety rules -- what we should potentially extract is whether or not we should expect signed vote proposals

consensus/safety-rules/src/process.rs
```diff
@@ -48,13 +51,19 @@ impl Process {
 
     fn start_internal<T: Payload>(&mut self) {
         let data = self.data.take().expect("Unable to retrieve ProcessData");
-        remote_service::execute::<T>(data.author, data.storage, data.server_addr);
+        remote_service::execute::<T>(
+            data.author,
+            data.storage,
+            data.execution_pubkey,
```
davidiw, line 57: remove

consensus/safety-rules/src/process.rs
```diff
@@ -48,13 +51,19 @@ impl Process {
...
+            data.server_addr,
+        );
     }
 }
 
 struct ProcessData {
     author: Author,
     server_addr: SocketAddr,
+    execution_pubkey: Ed25519PublicKey,
```
davidiw, line 66: remove

consensus/safety-rules/src/safety_rules.rs
```diff
@@ -136,10 +136,8 @@ impl<T: Payload> SafetyRules<T> {
             Ok(())
         }
     }
-}
 
-impl<T: Payload> TSafetyRules<T> for SafetyRules<T> {
-    fn consensus_state(&mut self) -> Result<ConsensusState, Error> {
```
davidiw, line 142: why? we implemented the interface, there's no point in not exposing this via this trait, is there?

consensus/safety-rules/src/tests/safety_rules.rs
```diff
@@ -1,18 +0,0 @@
-// Copyright (c) The Libra Core Contributors
-// SPDX-License-Identifier: Apache-2.0
-
-use crate::{test_utils, tests::suite, SafetyRules, TSafetyRules};
-use consensus_types::common::{Payload, Round};
-use libra_types::validator_signer::ValidatorSigner;
-
-#[test]
-fn test() {
```
davidiw, line 9: why?

consensus/safety-rules/benches/safety_rules.rs
davidiw: PersistentSafetyStorage is a glue wrapper for SafetyRules around SecureStorage. The executor key will be stored in our storage solutions that allow a peer to safely access the public key without having access to the private key. We leverage it for all these type of activities and will see more usage as we expand our authenticated interfaces.

execution/execution-correctness/src/thread.rs
```diff
@@ -22,12 +23,13 @@ pub struct ThreadService {
 }
 
 impl ThreadService {
-    pub fn new(storage_addr: SocketAddr) -> Self {
+    pub fn new(storage_addr: SocketAddr, prikey: Ed25519PrivateKey) -> Self {
```
davidiw, line 26: every implementation of LEC should take in a secure storage where it obtains its private key
if we want to move forward with this PR, we should figure out 1) will be fix that in this PR and 2) if we do not, can we avoid making LEC a requirement while we fix it. Otherwise this will make for a very wonky implementation.

Note, I started from the bottom and am moving up today, so I don't know exactly what was done to get here.

execution/execution-correctness/src/local_client.rs
```diff
@@ -1,57 +0,0 @@
-// Copyright (c) The Libra Core Contributors
```
davidiw, line 1: please avoid leaking this into a massive commit, two commits here just like in safety-rules

execution/execution-correctness/src/execution_correctness_manager.rs
```diff
@@ -2,25 +2,48 @@
...
+use libra_secure_storage::{Storage, Value};
 use libra_vm::LibraVM;
 use std::{
+    convert::TryInto,
     net::SocketAddr,
     sync::{Arc, Mutex},
 };
 use storage_client::StorageClient;
 
+pub fn extract_execution_prikey(config: &mut NodeConfig) -> Ed25519PrivateKey {
```
davidiw, line 26: this diverges from the safety rules approach, the big concern here is that in the real final implementation we'll have
1) key for signing for lsr
2) key for authenticated channel with libra-node
and who knows what else we decide might make sense at some point in time here or whether or not we want to think about rotation, would recommend we just adopt passing in a storage solution instead.

Now if you want to move fast, this seems okay, but we really need to get this documented in our startup process for config/management, which I haven't yet seen in this PR how you do. But documentation goes beyond github docs.

consensus/src/test_utils/mock_state_computer.rs
```diff
@@ -108,24 +127,45 @@ impl StateComputer for MockStateComputer {
...
             vec![],
+        );
+        let vote_proposal = VoteProposal::new(
+            result.extension_proof(),
+            block.clone(),
+            result.epoch_state().clone(),
+        );
+        Ok(StateComputeResultWithSignature::new(
+            result,
+            self.private_key.sign_message(&vote_proposal.hash()),
```
davidiw, line 168: if you made signatures optional, it would make some of these tests easier. could we add a commit to this pr or come up with a plan to get us there?

consensus/src/round_manager_test.rs
```diff
@@ -104,7 +109,11 @@ impl NodeSetup {
                 signer.private_key().clone(),
                 waypoint,
             );
-            let safety_rules_manager = SafetyRulesManager::new_local(author, safety_storage);
+            let safety_rules_manager = SafetyRulesManager::new_local(
+                author,
+                safety_storage,
+                Ed25519PrivateKey::generate_for_testing().public_key(),
```
davidiw, line 115: probably repeating feedback from before, but this /must/ come from safety_storage not as an extra input parameter.

consensus/src/liveness/proposal_generator_test.rs
```diff
@@ -120,66 +117,3 @@ async fn test_old_proposal_generation() {
         .err();
     assert!(proposal_err.is_some());
 }
-
-#[tokio::test]
-async fn test_empty_proposal_after_reconfiguration() {
```
davidiw, line 125: @zekun000, for you to review.

consensus/src/block_storage/block_store_test.rs
```diff
@@ -396,41 +396,3 @@ fn test_need_fetch_for_qc() {
         NeedFetchResult::QCAlreadyExist,
     );
 }
-
-#[test]
-fn test_empty_reconfiguration_suffix() {
```
davidiw, line 401: @zekun000 for you to review

consensus/src/block_storage/block_store.rs
```diff
@@ -294,25 +306,14 @@ impl<T: Payload> BlockStore<T> {
...
-            )
-        } else {
-            // Although NIL blocks don't have payload, we still send a T::default() to compute
-            // because we may inject a block prologue transaction.
-            self.state_computer
-                .compute(&block, parent_block.id())
-                .with_context(|| format!("Execution failure for block {}", block))?
-        };
-        Ok(ExecutedBlock::new(block, state_compute_result))
+        // Although NIL blocks don't have payload, we still send a T::default() to compute
```
davidiw, line 309: *a payload

config/global-constants/src/lib.rs
```diff
@@ -11,6 +11,7 @@
 /// Definitions of global cryptographic keys (e.g., as held in secure storage)
 pub const ASSOCIATION_KEY: &str = "association";
 pub const CONSENSUS_KEY: &str = "consensus";
+pub const EXECUTION_KEY: &str = "execution";
```
davidiw, line 14: let's update any relevant documentation about this as well as @mgorven

consensus/consensus-types/src/vote_proposal.rs
davidiw: this was to Mathieu not you :)

execution/execution-correctness/src/thread.rs
lightmark: I don't quite understand you insist on taking secure storage into LEC service.

consensus/src/test_utils/mock_state_computer.rs
lightmark: I will use the optional field then this file should be intact.

config/global-constants/src/lib.rs
lightmark: What documentation you can think of?

execution/execution-correctness/src/serializer.rs
```diff
@@ -37,15 +40,29 @@ impl SerializerService {
...
+                    )
+                    .map(|mut result| {
+                        if let Some(prikey) = self.prikey.as_ref() {
+                            let vote_proposal = VoteProposal::new(
+                                result.extension_proof(),
+                                block_with_parent_id.0.clone(),
+                                result.epoch_state().clone(),
+                            );
+                            let signature = prikey.sign_message(&vote_proposal.hash());
+                            result.set_signature(signature);
```
davidiw, line 55: The general model should be to encapsulate this rather than amend it.

Specifically, VoteProposal should become SignedVoteProposal or MaybeSignedVoteProposal that way you avoid this nuisance.

execution/executor-types/src/lib.rs
```diff
@@ -151,9 +177,47 @@ impl StateComputeResult {
...
+
+    pub fn set_signature(&mut self, sig: Ed25519Signature) {
+        self.signature = Some(sig);
+    }
+}
+
+/// A structure that wraps the state_compute_result of a block and the signature of the
+/// VoteProposal corresponding to the block.
+#[derive(Debug, PartialEq, Eq, Clone, Serialize, Deserialize)]
+pub struct StateComputeResultWithSignature {
```
davidiw, line 204: why do we have the signature here and in StateComputeResult?

config/management/src/smoke_test.rs
```diff
@@ -126,6 +126,7 @@ fn smoke_test() {
 
         config.consensus.safety_rules.backend =
             secure_backend(helper.path(), &swarm_path, &ns, "safety-rules");
+        config.execution.backend = secure_backend(helper.path(), &swarm_path, &ns, "execution");
```
davidiw, line 129: nice

execution/execution-correctness/src/local.rs
```diff
@@ -11,37 +13,65 @@ use std::{
...
+            .execute_block(id_and_transactions_from_block(&block), parent_block_id)?;
+        if let Some(prikey) = local.prikey.as_ref() {
+            let vote_proposal = VoteProposal::new(
+                result.extension_proof(),
+                block,
+                result.epoch_state().clone(),
+            );
+            let signature = prikey.sign_message(&vote_proposal.hash());
+            result.set_signature(signature);
+        }
```
davidiw, line 73: so why the copy pasta? why not have a common interface like we have in safety rules?

consensus/safety-rules/src/persistent_safety_storage.rs
```diff
@@ -16,32 +18,45 @@ use std::str::FromStr;
...
+        execution_public_key: Option<Ed25519PublicKey>,
         waypoint: Waypoint,
     ) -> Self {
-        Self::initialize_(&mut internal_store, private_key, waypoint)
+        Self::initialize_(&mut internal_store, consensus_private_key, waypoint)
             .expect("Unable to initialize backend storage");
-        Self { internal_store }
+        Self {
+            internal_store,
+            execution_public_key,
```
davidiw, line 50: if we go this path, let's just pass in the execution private key?

consensus/safety-rules/src/persistent_safety_storage.rs
```diff
@@ -61,6 +79,16 @@ impl PersistentSafetyStorage {
...
     }
 
+    pub fn execution_public_key(&self) -> Result<Ed25519PublicKey> {
+        Ok(if let Some(key) = self.execution_public_key.as_ref() {
+            key.clone()
+        } else {
+            self.internal_store
+                .get_public_key(EXECUTION_KEY)
+                .map(|r| r.public_key)?
+        })
```
davidiw, line 89: this is some really unhappy logic to deal with testing.

config/src/config/safety_rules_config.rs
```diff
@@ -8,6 +8,7 @@ use std::{net::SocketAddr, path::PathBuf};
 #[derive(Clone, Debug, Deserialize, PartialEq, Serialize)]
 #[serde(default, deny_unknown_fields)]
 pub struct SafetyRulesConfig {
+    pub verify_vote: bool,
```
davidiw, line 11: naming sucks, maybe verify_execution_signature or verify_vote_proposal_signature

consensus/safety-rules/src/tests/local.rs
```diff
@@ -2,17 +2,33 @@
...
     let storage = test_utils::test_storage(&signer);
-    let safety_rules_manager = SafetyRulesManager::new_local(signer.author(), storage);
+    let safety_rules_manager = SafetyRulesManager::new_local(signer.author(), storage, verify_vote);
     let safety_rules = safety_rules_manager.client();
-    (safety_rules, signer)
+    (
+        safety_rules,
+        signer,
+        if verify_vote {
+            Some(Ed25519PrivateKey::generate_for_testing())
```
davidiw, line 29: I mean for tests... why not just always pass in a key

config/src/config/execution_config.rs
```diff
@@ -19,6 +19,7 @@ const GENESIS_DEFAULT: &str = "genesis.blob";
 pub struct ExecutionConfig {
     #[serde(skip)]
     pub genesis: Option<Transaction>,
+    pub sign_vote: bool,
```
davidiw, line 22: naming could be improved... sign execution results etc

consensus/consensus-types/src/vote_proposal.rs
```diff
@@ -19,6 +22,25 @@ pub struct VoteProposal {
     block: Block,
     /// An optional field containing the next epoch info.
     next_epoch_state: Option<EpochState>,
+    /// The signature of this proposal's hash from Libra Execution Correctness service. It is
+    /// an `Option` because the LEC can be configured to not sign the vote hash.
+    #[serde(skip)]
```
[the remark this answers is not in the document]
lightmark: I think I removed the `skip` as it must be serialized.
I need to sign the hash of vote_proposal.

config/src/config/execution_config.rs
lightmark: but what I signed inside is `vote_proposal.hash()`

consensus/src/block_storage/block_store_test.rs
zekun000: do we have counter-part in LEC test?

lightmark: LEC has no way to know the reconfig logic. I think the test self is sort of hacky since the blocks in test are inserted manually. The blocker is MockStateComputor.

zekun000: I thought we move the hack from block store to LEC? so that LEC now provides the guarantee any block has a parent with reconfiguration has the same execution result.

lightmark: Yes. But the test here definitely belongs to consensus. LEC doesn't know what reconfiguration flow is. The problem is the way how this test runs is hacky two with `TreeInserter`. If you really want these tests let me see whether I can hack into it.

zekun000: LEC can test executing a reconfiguration block and then executing a child block and assert the result is the same?

the empty part belongs to consensus

lightmark: This part is easier. Then I will leave the empty part for you later :)

execution/execution-correctness/src/local.rs
lightmark: What do you mean here? Do you mean why we have `ExecutionCorrectness` trait and `BlockExecutor` trait, both of which look so similar. The reason is the signature of `execute_block` is different. the second paramater is `Block` and `Vec<Transaction>`, respectively.

execution/execution-correctness/src/execution_correctness_manager.rs
```diff
@@ -33,25 +55,45 @@ pub struct ExecutionCorrectnessManager {
...
+                Self::new_thread(storage_address, execution_prikey)
             }
-            ExecutionCorrectnessService::SpawnedProcess(_) => Self::new_spawned_process(config),
-            ExecutionCorrectnessService::Local => Self::new_local(storage_address),
-            ExecutionCorrectnessService::Serializer => Self::new_serializer(storage_address),
-            ExecutionCorrectnessService::Thread => Self::new_thread(storage_address),
+            _ => panic!(
+                "Unimplemented ExecutionCorrectnessService: {:?}",
+                config.execution.service
+            ),
```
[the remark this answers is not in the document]
lightmark: I just followed this. https://github.com/libra/libra/blob/master/consensus/safety-rules/src/safety_rules_manager.rs#L77
ofc it is unreachable.

execution/executor-types/src/lib.rs
lightmark: I think I forget to remove it.

consensus/safety-rules/src/persistent_safety_storage.rs
lightmark: some technical problems. I cannot fetch the private_key here I think.

consensus/safety-rules/src/persistent_safety_storage.rs
lightmark: test_config has an execution keypair and privkey can only be taken once. If I take out private key here, then LEC will get `None` and throw error. Another way is to copy the privkey somehow but I feel it is more hacky.

consensus/safety-rules/src/tests/local.rs
lightmark: What's the difference?

execution/execution-correctness/src/local.rs
davidiw: ugh, I lost context on this comment and don't see why it's here....

execution/execution-correctness/src/execution_correctness_manager.rs
```diff
@@ -2,25 +2,47 @@
...
+            .as_mut()
+            .expect("Missing consensus keypair in test config")
+            .take_private()
+        {
+            storage
+                .import_private_key(EXECUTION_KEY, private_key)
+                .expect("Unable to insert execution key");
+        }
+    }
+    storage.export_private_key(EXECUTION_KEY).ok()
```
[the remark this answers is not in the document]
lightmark: I think I changed this line in the following commits.

consensus/src/block_storage/block_store_test.rs
lightmark: @zekun000 plz take a look on the new test I added.

execution/executor/src/executor_test.rs
```diff
@@ -431,6 +431,23 @@ fn test_executor_execute_and_commit_chunk_local_result_mismatch() {
...
+    let output1 = executor
+        .execute_block((first_block_id, vec![first_txn]), parent_block_id)
+        .unwrap();
+    parent_block_id = first_block_id;
+    let second_block = TestBlock::new(0..10, 10, gen_block_id(2));
+    let output2 = executor
+        .execute_block((second_block.id, second_block.txns), parent_block_id)
+        .unwrap();
+    assert_eq!(output1.root_hash(), output2.root_hash());
+}
```
lightmark, line 449: @zekun000 fyi.

consensus/safety-rules/src/persistent_safety_storage.rs
zekun000: why the execution_public_key is not stored in the internal_store?

consensus/src/block_storage/block_store.rs
```diff
@@ -149,18 +149,23 @@ impl BlockStore {
             root_metadata.accu_hash,
         );
 
+        let result = StateComputeResult::new(
+            root_metadata.accu_hash,
+            root_metadata.frozen_root_hashes,
+            root_metadata.num_leaves, /* num_leaves */
+            vec![],                   /* parent_root_hashes */
+            0,                        /* parent_num_leaves */
+            None,                     /* validators */
```
zekun000, line 158: it's called epoch_state now

consensus/src/block_storage/block_store_test.rs
zekun000: you can keep this test but just delete a5?

consensus/src/liveness/proposal_generator_test.rs
zekun000: why this test failed?

consensus/src/round_manager.rs
```diff
@@ -526,27 +522,19 @@ impl RoundManager {
...
         let vote_proposal = VoteProposal::new(
-            AccumulatorExtensionProof::<TransactionAccumulatorHasher>::new(
-                parent_block.compute_result().frozen_subtree_roots().clone(),
-                parent_block.compute_result().num_leaves(),
-                executed_block
-                    .compute_result()
-                    .transaction_info_hashes()
-                    .clone(),
-            ),
+            executed_block.compute_result().extension_proof(),
```
zekun000, line 526: it looks executed_block can generate the vote_proposal and maybe_signed_vote_proposal directly?
```
fn vote_proposal(&self) -> VoteProposal {
}
```

execution/execution-correctness/src/execution_correctness.rs
```diff
@@ -0,0 +1,30 @@
...
+use consensus_types::block::Block;
+use executor_types::{Error, StateComputeResult};
+use libra_crypto::HashValue;
+use libra_types::{
+    contract_event::ContractEvent, ledger_info::LedgerInfoWithSignatures, transaction::Transaction,
+};
+
+/// Interface for ExecutionCorrectness.
+/// It is basically the same as BlockExecutor except some interfaces will return signature with result.
+pub trait ExecutionCorrectness: Send {
```
zekun000, line 13: this has the same function signatures as BlockExecutor, why create a separate trait?

consensus/src/block_storage/block_store_test.rs
lightmark: a6 is the child of a5...

consensus/src/liveness/proposal_generator_test.rs
lightmark: It inserts reconfiguration block using the hacky test helper function.

zekun000:
get it landed first, we can deal with tests later

execution/execution-correctness/src/execution_correctness.rs
lightmark: Not the same. Take a look on `execute_block`

lightmark:
@davidiw

consensus/src/block_storage/block_store_test.rs
davidiw: status here?

consensus/src/liveness/proposal_generator_test.rs
davidiw: status here?

config/src/config/safety_rules_config.rs
```diff
@@ -8,6 +8,7 @@ use std::{net::SocketAddr, path::PathBuf};
 #[derive(Clone, Debug, Deserialize, PartialEq, Serialize)]
 #[serde(default, deny_unknown_fields)]
 pub struct SafetyRulesConfig {
+    pub verify_vote_proposal_signature: bool,
```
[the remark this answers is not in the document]
lightmark: IIRC, if I put a `bool` after a `struct`, there are some serde issue.

consensus/consensus-types/src/vote_proposal.rs
```diff
@@ -54,3 +58,22 @@ impl Display for VoteProposal {
         write!(f, "VoteProposal[block: {}]", self.block,)
     }
 }
+
+/// Wraps a vote_proposal and its signature.
+#[derive(Clone, Deserialize, Serialize)]
+pub struct MaybeSignedVoteProposal {
```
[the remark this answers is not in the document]
lightmark: `verify` still needs a public_key as an input so maybe the current way is good enough.

lightmark:
> on retrospect, I think the right thing to do here is just clone the private key and insert it into multiple storages... if we are pulling the key out of the config, we are already doing something inherently insecure. Do you want to do that in this PR or the next? I don't want us to move forward beyond the end of this month with this wonky logic.
I agree it would be wise to clone private_key everywhere. But according to the current design and implementation, I don't agree it is reasonable.
1. The api of `Keypair` can allow private_key to be taken once. So there is no way to clone it, even in a hacky way. Because you cannot take it and put it back.
2. Another concern is the keypair api is designed that if LEC took the private_key, it will be gone and LSR will get `None` when try to clone the key. To avoid this, you need LSR initialization is always ahead of LEC. That's not ideal.

For tests, Zekun already said okay.

PR #4119, opened 2020-05-29
[secure-storage] Implement an optional ability to import keys

davidiw:
May the crypto gods forgive me... but moving forward, we will probably
have cluster test and various test networks using vault yet test
networks depend on the ability to generate keys deterministically. So we
introduce a means to "import" keys into various secure storages. Because
this is for testing only, it is optional to implement and users should be
extremely careful using it.

### DIFF
--- a/secure/storage/vault/src/lib.rs
+++ b/secure/storage/vault/src/lib.rs
@@ -3,7 +3,10 @@
 
 #![forbid(unsafe_code)]
 
-use libra_crypto::ed25519::{Ed25519PrivateKey, Ed25519PublicKey, Ed25519Signature};
+use libra_crypto::{
+    ed25519::{Ed25519PrivateKey, Ed25519PublicKey, Ed25519Signature},
+    PrivateKey,
+};
 use serde::{Deserialize, Serialize};
 use serde_json::json;
 use std::{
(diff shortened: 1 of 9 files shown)


secure/storage/vault/src/lib.rs
```diff
@@ -391,6 +406,134 @@ impl Client {
...
+    version_template: String,
+    storage_prefix: String,
+}
+
+#[derive(Debug, Default, Deserialize, PartialEq, Serialize)]
+pub struct KeyBackupInfo {
+    key: Option<String>,
+    hhmac_key: Option<String>,
+    time: String,
+    ec_x: Option<String>,
```
JoshLind, line 522: I'm wondering if it's strictly necessary to define all of these fields when sending the "restore" request? I know "backup" returns them as null, so it probably makes sense to define them, but there's a lot of unused ones here.. especially as we aren't calling "backup" ourselves. Either way, this is probably okay!

davidiw: At some point, I just left well enough alone. I already trimmed a lot of content from this :)

davidiw:
Thanks for the quick review!

PR #4123, opened 2020-05-29
[json-rpc] update TransactionView and get_metadata

christinacelee:
## Summary

- expose txn hash in TransactionView
- add optional version arg for get_metadata to support querying blockchain metadata for past versions, not just the most current one. If no version is provided, default to original behavior of returning current blockchain metadata

## Test Plan

Existing/updated tests


thefallentree:
this broke get_metadata call wihout the param

PR #4127, opened 2020-05-29
[json-rpc] add get_network_status endpoint

christinacelee:
## Summary

Adding `get_network_status` that returns the number of peers the node is connected to, to gauge network health
Refactored some test code accordingly

## Test Plan

Existing/new tests

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2488,6 +2488,7 @@ dependencies = [
  "libra-workspace-hack 0.1.0",
  "libradb 0.1.0",
  "move-core-types 0.1.0",
+ "network 0.1.0",
  "once_cell 1.4.0 (registry+https://github.com/rust-lang/crates.io-index)",
  "proptest 0.9.6 (git+https://github.com/mimoo/proptest.git?branch=fuzzing)",
  "reqwest 0.10.5 (registry+https://github.com/rust-lang/crates.io-index)",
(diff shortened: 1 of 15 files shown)


thefallentree:
why not just do this in get_metadata API ? 

Also, on fullnode isn't there only going to be 1 ?

christinacelee:
> why not just do this in get_metadata API ?
> 
> Also, on fullnode isn't there only going to be 1 ?

Get metadata now supports querying for past blocks, so we can't add network connectivity info there

christinacelee:
Are there risks involved with exposing this info to the internet? Coinbase, for example, supports querying a node for network status to check network health

thefallentree:
I don't think json-rpc is the right approach though, typically healcheck are handled differently, maybe by a /health handler directly, returning an json or just text, we should consider doing that.


or. just expose some prometheus metrics under /metrics, which are all more extensible than this.

davidiw:
Interesting, not sure why this got re-opened.

We actually had the debug interface ported to json rpc but in its own service.

Closing PR, it seems good to me. It might be good in the future to reference partner services that we're replicating to better motivate these type of PRs.

Thanks for humoring us with answers to our questions :).

thefallentree:
I'm not sure what you meant, but if you don't want to discuss anymore what can I say.

PR #4133, opened 2020-05-29
[language][vm] implement additional resource safety checks

vgao1996:
## Summary
This introduces Container::Resource, making it possible to determine the kinds of the values. A few checks were added to prevent resources from being copied or implicitly destroyed.

Note this is an additional safety net in case the verifier is bugged. Our vision is to start safe, and as we gradually gain trust on the verifier running in production, we may consider removing these checks, if performance ends up being an issue.

## Test Plan
cargo test

Note: still a WIP. Existing functional tests are passing but need to write a few unit tests.

### DIFF
--- a/language/move-vm/runtime/src/interpreter.rs
+++ b/language/move-vm/runtime/src/interpreter.rs
@@ -143,6 +143,7 @@ impl Interpreter {
                     .or_else(|err| Err(self.maybe_core_dump(err, &current_frame)))?;
             match exit_code {
                 ExitCode::Return => {
+                    current_frame.locals.check_resources_for_return()?;
                     if let Some(frame) = self.call_stack.pop() {
                         current_frame = frame;
                     } else {
@@ -814,9 +815,10 @@ impl Frame {
                             |acc, v| acc.add(v.size()),
                         );
                         cost_strategy.charge_instr_with_size(Opcodes::PACK, size)?;
+                        let is_resource = resolver.struct_at(*sd_idx).is_resource;
                         interpreter
                             .operand_stack
-                            .push(Value::struct_(Struct::pack(args)))?;
+                            .push(Value::struct_(Struct::pack(args, is_resource)))?;
                     }
                     Bytecode::PackGeneric(si_idx) => {
                         let field_count = resolver.field_instantiation_count(*si_idx);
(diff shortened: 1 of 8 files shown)


language/move-vm/runtime/src/interpreter.rs
```diff
@@ -826,9 +828,25 @@ impl Frame {
...
+                        } else {
+                            let mut is_resource = false;
+                            for ty in struct_def.get_instantiation() {
+                                if resolver.is_resource(&ty.subst(self.ty_args())?)? {
+                                    is_resource = true;
+                                }
+                            }
+                            is_resource
+                        };
+
```
vgao1996, line 846: @dariorussi FYI I decided to not worry about the Mutex right now so we can get the changes to the values in. The multithreading story is overall complicated and I don't think we need to get it sorted out in this PR.

language/move-vm/types/src/values/values_impl.rs
```diff
@@ -476,30 +505,50 @@ impl IndexedRef {
...
+                v1[self.idx] == *v2[other.idx].as_value_ref::<bool>()?
+            }
 
-            (General(v1), Address(v2)) => {
+            (General(v1), Address(v2)) | (Resource(v1), Address(v2)) => {
                 *v1[self.idx].as_value_ref::<AccountAddress>()? == v2[other.idx]
             }
-            (Address(v1), General(v2)) => {
+            (Address(v1), General(v2)) | (Address(v1), Resource(v2)) => {
                 v1[self.idx] == *v2[other.idx].as_value_ref::<AccountAddress>()?
```
vgao1996, line 552: We're getting a bit bloated here, but I'm not too worried as `or_patterns` has landed. Hopefully we'll be able to bear with this until the day `or_patterns` gets stabilized ;)

language/move-vm/types/src/natives/function.rs
```diff
@@ -56,6 +56,8 @@ pub trait NativeContext {
     fn save_event(&mut self, event: ContractEvent) -> VMResult<()>;
     /// Converts types to fet types.
     fn convert_to_fat_types(&self, types: Vec<Type>) -> VMResult<Vec<FatType>>;
+    /// Whether a type is a resource of not.
```
dariorussi, line 59: typo - 'of' -> 'or'

language/move-vm/types/src/natives/function.rs
```diff
@@ -56,6 +56,8 @@ pub trait NativeContext {
...
+    fn is_resource(&self, ty: &Type) -> VMResult<bool>;
```
dariorussi, line 60: I would expect this type to be fully instantiatied always and no need to go through the resolver ever.
I am fine fixing it later but I am wondering if I am missing something or this was mostly to have the API be the most general possible (which may or may not be a good thing)

vgao1996: I added this because it was the most straightforward way for me to determine the kind of a generic type, but let me check if I can instead make the `Type` carry its kind so we don't have to go through the loader again.

PR #4134, opened 2020-05-29
[language] Move all native functions in move-vm/natives

dariorussi:
Isolate all native functions into the same crate.
The vector API will be reworked soon and that would make
this break up even cleaner.

### DIFF
--- a/language/e2e-tests/Cargo.toml
+++ b/language/e2e-tests/Cargo.toml
@@ -21,8 +21,9 @@ libra-state-view = { path = "../../storage/state-view", version = "0.1.0" }
 libra-types = { path = "../../types", version = "0.1.0", features = ["fuzzing"] }
 libra-workspace-hack = { path = "../../common/workspace-hack", version = "0.1.0" }
 move-core-types = { path = "../move-core/types", version = "0.1.0" }
-move-vm-types = { path = "../move-vm/types", version = "0.1.0", features = ["debug_module"] }
+move-vm-natives = { path = "../move-vm/natives", version = "0.1.0", features = ["debug_module"] }
 move-vm-runtime = { path = "../move-vm/runtime", version = "0.1.0", features = ["debug_module"] }
+move-vm-types = { path = "../move-vm/types", version = "0.1.0" }
 transaction-builder = { path = "../transaction-builder", version = "0.1.0", features = ["fuzzing"]}
 vm = { path = "../vm", version = "0.1.0" }
 vm-genesis = { path = "../tools/vm-genesis", version = "0.1.0" }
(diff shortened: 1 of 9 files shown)


language/e2e-tests/Cargo.toml
```diff
@@ -21,8 +21,9 @@ libra-state-view = { path = "../../storage/state-view", version = "0.1.0" }
 libra-types = { path = "../../types", version = "0.1.0", features = ["fuzzing"] }
 libra-workspace-hack = { path = "../../common/workspace-hack", version = "0.1.0" }
 move-core-types = { path = "../move-core/types", version = "0.1.0" }
-move-vm-types = { path = "../move-vm/types", version = "0.1.0", features = ["debug_module"] }
+move-vm-natives = { path = "../move-vm/natives", version = "0.1.0", features = ["debug_module"] }
 move-vm-runtime = { path = "../move-vm/runtime", version = "0.1.0", features = ["debug_module"] }
+move-vm-types = { path = "../move-vm/types", version = "0.1.0", features = ["debug_module"] }
```
dariorussi, line 26: not sure we need to pass this to anybody except natives but it does not seem to create issues so we may leave it as is

language/move-vm/types/src/values/values_impl.rs
```diff
@@ -1395,102 +1445,50 @@ impl IntegerValue {
...
+            StatusCode::UNKNOWN_INVARIANT_VIOLATION_ERROR,
+        )
+        .with_message(format!(
+            "vector elem layout mismatch, expected {:?}, got {:?}",
+            ty, v
+        ))),
+    }
+}
+
+impl VectorRef {
```
dariorussi, line 1487: this api is not super consistent but once @vgao1996 reworks the container and the cast to a vector is should all be much more reasonable.
In fact this API may disappear entirely at that point and an implementation only defined in the natives crate

dariorussi:
this is mostly for @vgao1996 to review and I'll be waiting for some comments from him anyway....

language/e2e-tests/Cargo.toml
vgao1996: If I remember correctly feature flags don't get propagated to dependencies so I think you can remove `features = ["debug_module"]` for move-vm-types. (Also remember to delete it from the [feature] section in cargo.toml.)

language/move-vm/types/src/values/values_impl.rs
vgao1996: Yeah I'll review this while I continue working on the cleanup. Not immediately seeing how this API could disappear with our planned changes but we'll see.

PR #4136, opened 2020-05-29
[Key Manager] Added a docker container to deploy the key manager.

JoshLind:
## Motivation

This PR adds a docker container for the key manager. To avoid creating multiple similar container images, we refactor the safety rules docker image into a "common" image that can run both the key manager and safety rules. 

Yes.

## Test Plan

I've locally built (./build.sh) and run (./run.sh) the docker container. I've also tested both the safety rules invocation script and the key manager script and things seem to start up okay.
## Related PRs

This PR requires us first landing the config-builder for key manager support (https://github.com/libra/libra/pull/4094), before the key manager can be started using docker.


docker/key-manager/run.sh
```diff
@@ -0,0 +1,17 @@
...
+# Copyright (c) The Libra Core Contributors
+# SPDX-License-Identifier: Apache-2.0
+set -ex
+
+image="${1:-libra_key_manager:latest}"
+
+docker network create --subnet 172.18.0.0/24 testnet || true
+
+docker run \
+    -e JSON_RPC_ENDPOINT="http://127.0.0.1:8080" \
```
JoshLind, line 11: @mgorven @davidiw what is the best way to define these? Right now the script just uses random values (e.g., localhost:8080 and a garbage token), but this is obviously not right. Do we have a story for how we'll know these constants at deployment time?

mgorven: The values in here are designed for running locally.

docker/key-manager/run.sh
```diff
@@ -0,0 +1,17 @@
...
+    -e ROTATION_PERIOD_SECS="3600" \
+    -e VAULT_HOST="http://127.0.0.1:8888" \
```
mgorven, line 13: `docker/vault/run.sh` runs vault on port 8200, so that should be used here.

docker/key-manager/run.sh
JoshLind: e.g., the json rpc interface exposed by each validator node can be accessed at localhost (assuming) the KM and node are deployed on the same machine, but the port is "randomized", so we don't really know this until the node config exists... similarly, what are our assumptions for vault?

mgorven: For running docker containers locally we assume fixed port numbers. `run.sh` isn't used anywhere else, Terraform/Kubernetes handle this in that case.

docker/key-manager/Dockerfile
```diff
@@ -0,0 +1,42 @@
+FROM debian:buster AS toolchain
```
davidiw, line 1: @mgorven, I think we should have a release docker container with all binaries. Then we can amend it with the docker-run.sh as appropriate. What are your thoughts?

docker/key-manager/docker-run-dynamic.sh
```diff
@@ -0,0 +1,38 @@
+#!/bin/bash
```
davidiw, line 1: just call this docker-run.sh

docker/key-manager/run.sh
davidiw: Look at docker/mint/run.sh 
let's be consistent like that...

docker/key-manager/Dockerfile
mgorven: I think having one image with both safety-rules and key-manager makes sense. I'm not sure about including validator though, from a security perspective.

JoshLind: @davidiw @mgorven is this something you'd like me to address in this PR, or submit as a separate PR?

mgorven: If we're going to do that (which makes sense to me) it's worth doing now. There's a bunch of other stuff to do for adding a new image which can be avoided that way (creating an ECR repo, updating CI to build it, etc.)

JoshLind: Okay, awesome! Thanks @mgorven, I'll make that change today and update this PR when it's ready 😄

JoshLind:
@mgorven @davidiw I've updated this PR to try and refactor safety rules and the key manager to use a single docker image, and have run scripts for both. 

Despite the previous approval, I think this needs some eyes because there's a few parts I'm a little uncertain about/might need cleaning up. Let me know what you think!

docker/common/run-key-manager.sh
```diff
@@ -0,0 +1,27 @@
...
+# SPDX-License-Identifier: Apache-2.0
+set -e
+
+# Example Usage (arguments that are not specified will be assigned defaults):
+# ./run-key-manager.sh <IMAGE> <JSON_RPC_ENDPOINT> <ROTATION_PERIOD_SECS> <VALIDATOR_ACCOUNT> <VAULT_HOST> <VAULT_TOKEN>
+
+IMAGE="${1:-libra_common:latest}"
+JSON_RPC_ENDPOINT="${2:-http://127.0.0.1:8080}"
+ROTATION_PERIOD_SECS="${3:-3600}"
+VALIDATOR_ACCOUNT="${4:-00000000000000000000000000000000}"
```
JoshLind, line 12: @davidiw, we can remove this in my next PR that will fetch the validator operator account from storage!

docker/common/Dockerfile
```diff
@@ -17,21 +17,28 @@ FROM toolchain AS builder
 COPY . /libra
 
 ENV RUSTFLAGS "-Ctarget-cpu=skylake -Ctarget-feature=+aes,+sse2,+sse4.1,+ssse3"
-RUN cargo build --release -p libra-node -p cli -p config-builder -p safety-rules && cd target/release && rm -r build deps incremental
+RUN cargo build --release -p libra-node -p cli -p config-builder -p libra-key-manager -p safety-rules && cd target/release && rm -r build deps incremental
 RUN strip /libra/target/release/config-builder
+RUN strip /libra/target/release/libra-key-manager
```
mgorven, line 22: Let's not strip key-manager and safety-rules for now, I think more useful stacktraces are worth it for now.

docker/common/docker-run.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
```
mgorven, line 1: Let's keep all the scripts in `safety-rules/` for now; we can tackle renaming the image to something like "tcb" later.

docker/common/docker-run.sh
```diff
@@ -0,0 +1,13 @@
...
+# Copyright (c) The Libra Core Contributors
+# SPDX-License-Identifier: Apache-2.0
+set -e
+
+if [ "${COMMAND_TO_RUN}" = "key-manager" ]; then # Execute the key manager
```
mgorven, line 6: I'd rather use the docker command to select what to run. Also this needs to be backwards compatible for now. So:
* Remove this script
* Make `/safety-rules.sh` the CMD in the dockerfile
* Have `run-key-manager.sh` specify `/key-manager.sh` as the command

JoshLind:
Thanks @mgorven! This is ready for another look when you are 😄

docker/safety-rules/run-key-manager.sh
```diff
@@ -0,0 +1,27 @@
...
+
+docker network create --subnet 172.18.0.0/24 testnet || true
+
+docker run \
+    -e JSON_RPC_ENDPOINT="$JSON_RPC_ENDPOINT" \
+    -e ROTATION_PERIOD_SECS="$ROTATION_PERIOD_SECS" \
+    -e VALIDATOR_ACCOUNT="$VALIDATOR_ACCOUNT" \
+    -e VAULT_HOST="$VAULT_HOST" \
+    -e VAULT_TOKEN="$VAULT_TOKEN" \
+    --ip 172.18.0.2 \
```
mgorven, line 24: Give this a different IP so safety-rules and key-manager can be run at the same time.

docker/safety-rules/Dockerfile
```diff
@@ -17,7 +17,7 @@ FROM toolchain AS builder
 COPY . /libra
 
 ENV RUSTFLAGS "-Ctarget-cpu=skylake -Ctarget-feature=+aes,+sse2,+sse4.1,+ssse3"
-RUN cargo build --release -p libra-node -p cli -p config-builder -p safety-rules && cd target/release && rm -r build deps incremental
+RUN cargo build --release -p libra-node -p cli -p config-builder -p libra-key-manager -p safety-rules && cd target/release && rm -r build deps incremental
```
mgorven, line 20: Copy this exact build command to the other Docker files (this is done for docker caching, so that the cargo build is only done once across all the images).

    docker/client/Dockerfile:RUN cargo build --release -p libra-node -p cli -p config-builder -p safety-rules && cd target/release && rm -r build deps incremental
    docker/init/Dockerfile:RUN cargo build --release -p libra-node -p cli -p config-builder -p safety-rules && cd target/release && rm -r build deps incremental
    docker/mint/Dockerfile:RUN cargo build --release -p libra-node -p cli -p config-builder -p safety-rules && cd target/release && rm -r build deps incremental
    docker/validator/Dockerfile:RUN cargo build --release -p libra-node -p cli -p config-builder -p safety-rules && cd target/release && rm -r build deps incremental

JoshLind: Ooh -- that's a cool trick. Each build is pretty long, so it's cool to know we have that 😄

JoshLind:
Thanks @mgorven for the quick response! PTAL when you get a chance!

PR #4139, opened 2020-05-29
[move source language] Add aliasing to spec contexts

tnowacki:
## Motivation

- Spec functions and schemas added to a Modules set of module members
  - Implies implicit aliasing
  - Cannot conflict with function/struct/constant names
- Added restriction to spec schema names (same as structs/constants)

## Test Plan

- new tests
- cargo test

### DIFF
--- a/language/move-prover/tests/sources/stdlib/modules/libra_account.move
+++ b/language/move-prover/tests/sources/stdlib/modules/libra_account.move
@@ -99,10 +99,10 @@ module LibraAccount {
     }
     spec fun deposit {
         aborts_if to_deposit.value == 0;
-        aborts_if !exists<T>(sender());
+        aborts_if !::exists<T>(sender());
         aborts_if global<T>(sender()).sent_events.counter + 1 > max_u64();
-        aborts_if !exists<T>(payee);
-        aborts_if !exists<Balance<Token>>(payee);
+        aborts_if !::exists<T>(payee);
+        aborts_if !::exists<Balance<Token>>(payee);
         aborts_if global<Balance<Token>>(payee).coin.value + to_deposit.value > max_u64();
         aborts_if global<T>(payee).received_events.counter + 1 > max_u64();
         ensures global<T>(sender()).sent_events.counter == old(global<T>(sender()).sent_events.counter) + 1;
@@ -116,9 +116,9 @@ module LibraAccount {
     }
     spec fun deposit_to_sender {
         aborts_if to_deposit.value == 0;
-        aborts_if !exists<T>(sender());
+        aborts_if !::exists<T>(sender());
         aborts_if global<T>(sender()).sent_events.counter + 1 > max_u64();
-        aborts_if !exists<Balance<Token>>(sender());
+        aborts_if !::exists<Balance<Token>>(sender());
         aborts_if global<Balance<Token>>(sender()).coin.value + to_deposit.value > max_u64();
         aborts_if global<T>(sender()).received_events.counter + 1 > max_u64();
         ensures global<T>(sender()).sent_events.counter == old(global<T>(sender()).sent_events.counter) + 1;
@@ -141,10 +141,10 @@ module LibraAccount {
     }
     spec fun deposit_with_metadata {
         aborts_if to_deposit.value == 0;
-        aborts_if !exists<T>(sender());
+        aborts_if !::exists<T>(sender());
         aborts_if global<T>(sender()).sent_events.counter + 1 > max_u64();
-        aborts_if !exists<T>(payee);
-        aborts_if !exists<Balance<Token>>(payee);
+        aborts_if !::exists<T>(payee);
+        aborts_if !::exists<Balance<Token>>(payee);
         aborts_if global<Balance<Token>>(payee).coin.value + to_deposit.value > max_u64();
         aborts_if global<T>(payee).received_events.counter + 1 > max_u64();
         ensures global<T>(sender()).sent_events.counter == old(global<T>(sender()).sent_events.counter) + 1;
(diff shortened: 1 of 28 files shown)


language/move-prover/tests/sources/stdlib/modules/libra_account.move
```diff
@@ -99,10 +99,10 @@ module LibraAccount {
     }
     spec fun deposit {
         aborts_if to_deposit.value == 0;
-        aborts_if !exists<T>(sender());
+        aborts_if !::exists<T>(sender());
```
wrwg, line 102: Just to confirm that we are on the same page here. As we discussed, we are planning to get rid of this by having a list of builtin function names for spec expressions.

tnowacki:
> We could also perhaps add the list of builtin functions to this PR already, avoiding the need of the changes in libra_account.

Maybe there was some confusion. Adding a list of builtin functions could simplify some work on the move-prover side. But it would not change the behavior in libra_account. 

The only thing that would change the behavior in libra_account is if you could do 
```
use exists; // refers to the builtin exists
....
```

These builtin names are in the top level namespace, and can be shadowed in any inner name space. The same thing works for any other name-spacing layering. For example:
```
address 0x1 {
module M {
  fun foo() { ... }
  fun bar() { ... }
  fun baz() {
    use 0x1::M::bar as foo; // bar as foo shadows Self::foo as foo
    foo(); // resolves to Self::bar
  }
}
}
```
This works the same way in `LibraAccount` in that `LibraAccount::exists` shadows `exists`. You then need some new `use` to then reselect `exists into the current namespace

language/move-prover/tests/sources/stdlib/modules/libra_account.move
tnowacki: This would not change if the handling of builtin names were moved from Naming to Expansion. See longer comment below

wrwg:
> > We could also perhaps add the list of builtin functions to this PR already, avoiding the need of the changes in libra_account.
> 
> Maybe there was some confusion. Adding a list of builtin functions could simplify some work on the move-prover side. But it would not change the behavior in libra_account.
> 
> The only thing that would change the behavior in libra_account is if you could do
> 
> ```
> use exists; // refers to the builtin exists
> ....
> ```
> 
> These builtin names are in the top level namespace, and can be shadowed in any inner name space. The same thing works for any other name-spacing layering. For example:
> 
> ```
> address 0x1 {
> module M {
>   fun foo() { ... }
>   fun bar() { ... }
>   fun baz() {
>     use 0x1::M::bar as foo; // bar as foo shadows Self::foo as foo
>     foo(); // resolves to Self::bar
>   }
> }
> }
> ```
> 
> This works the same way in `LibraAccount` in that `LibraAccount::exists` shadows `exists`. You then need some new `use` to then reselect `exists` into the current namespace

It appears we had a misunderstanding. This is unfortunate but I'm OK to move forward with this.

tnowacki:
@sblackshear @bob-wilson @tzakian 
The sort of sticking point here after talking to @wrwg is whether or not we should keyword the builtins? The consequences of this are, should you be able to shadow the builtins. 

# Allowing shadowing can be confusing

It can be misleading and confusing when reading a module. E.g
```
module Confuse {
  resource struct R { }
  // no relation to real exists, shadows
  public fun exists(a: addr): u64 { 0 }
  // no relation to real move_to, shadows
  public fun move_to(_s: &signer, r: R) {
    R { } = r;
  }
  ...
  fun foo(s: &signer) {
    ::move_to(s, R {})
  }
}
```

If they were keyworded, you could have the wrapper but it would have to be named differently, i.e. 
```
module LibraAccount {
  ...
  public fun exists_(a: address): bool {
    ::exists<Account>(a)
  } 
  public fun has_account(a: address): bool {
    ::exists<Account>(a)
  } 
}
```

# Allowing shadowing can be helpful

All of the builtins for a module are private, so by creating public wrappers, you can "expose" the behavior. But it shadows the function internally. We do this in `LibraAccount`:
```
module LibraAccount {
  ...
  public fun exists(a: address): bool {
    ::exists<Account>(a)
  } 
}
```
More generally you could imagine an all of the builtins (except the borrows) being exposed
```
module MyResource {
    resource struct MyResource { ... }

    public fun move_to(s: &signer, r: MyResource) {
        ::move_to(s, r)
    }

    public fun move_from(a: address): MyResource {
        ::move_from<MyResource>(a)
    }

    public fun exists(a: address): bool {
        ::exists<MyResource>(a)
    }
    ...
}
```
This might be natural in that situation we (@sblackshear  @tzakian) discussed where the module name might match up with the resource, i.e.the module has one key resource

The downside is you might want to add aliasing to help with certain cases, which makes the whole thing more confusing, e.g.

```
module LibraAccount {
  ...
  public fun exists(a: address): bool {
    ::exists<Account>(a)
  } 

  public fun has_Foo_and_Bar(a: address): bool {
    use exists;
    assert(Self::exists<Account>(a), 42);
    exists<Foo>(a) && exists<Bar>
  } 
}
```
--- 

I could see us going either way here. The key things to weigh are (1) How much are we concerned about the confusing aspect? (2) How useful do we think these same-name wrappers being?

bob-wilson:
I don't have a strong opinion on either part of the question, but I'm generally in favor of allowing shadowing of builtins. I see the potential for confusion, but for me, that is just an indication that we need to be careful to avoid using that ability in confusing ways.

How much of this discussion is being driven by the conflicting "exists" builtins for the Move language vs. the prover? I haven't been in the loop on all the discussions, but that's the impression I get. If so, maybe the solution is to rename one or the other (or both). Regardless of this issue, it seems like a bad idea to overload the same builtin name with different meanings in those different contexts.

wrwg:
Two thoughts:

1. I'm wondering why you are making builtins private to a module. This seems very unusual to me. If you do not want that an external module can say `exists<OtherModule::R>(a)`, I would find it much more natural to just make this a special context condition. The error message would be much cleaner. If you want to "export" the ability to check for a module's resource, you should use a different, more abstract name. E.g. `LibraAccount::account_exists_at(addr)`.
Or maybe I'm missing some other reason why you are doing this? Anyway, in most programming languages I know, builtins are either in some special namespace which is implicitly imported into each module (and not automatically re-exported), or they are hardwired in the syntax. 

2. I think shadowing is something which should be avoided in "readable" source code as much as possible if it comes to global symbols. If a global symbol is ambiguous, qualification should be the preferred way to resolve this. I understand that "writing" code is easier with this feature, but in my opinion "reading" Move source code is the more important and frequent use case.

PR #4140, opened 2020-05-29
[network] Add a listener for receiving connnectivity config updates f…

agc-quadra:
…rom statesync

## Motivation

This is an initial step to address #3925.  

The PR does three things.

1.  It introduces a  ConfigurationListener intended to stand between the ConnectivityManager and any source of network updates.

1.  It wires the ConfigurationListener into the **validator** network via network_builder in main_node

1.  It eliminates the link between consensus/epoch_manager and updates to ConnectivityManager in favor of the direct link of the previous step.

yes.

## Test Plan

There is no integration/unit test written yet.  Ideally that should be in place before the code is committed, but given time pressures that may not be viable.  @davidiw @phlip9  for judgment.

smoke_tests pass in CI.

## Related PRs

### DIFF
--- a/libra-node/src/main_node.rs
+++ b/libra-node/src/main_node.rs
@@ -16,11 +16,11 @@ use libra_json_rpc::bootstrap_from_config as bootstrap_rpc;
 use libra_logger::prelude::*;
 use libra_mempool::MEMPOOL_SUBSCRIBED_CONFIGS;
 use libra_metrics::metric_server;
-use network_simple_onchain_discovery::ConfigurationChangeListener;
 use libra_types::{on_chain_config::ON_CHAIN_CONFIG_REGISTRY, waypoint::Waypoint, PeerId};
 use libra_vm::LibraVM;
 use libradb::LibraDB;
 use network::validator_network::network_builder::{AuthenticationMode, NetworkBuilder};
+use network_simple_onchain_discovery::ConfigurationChangeListener;
 use onchain_discovery::{client::OnchainDiscovery, service::OnchainDiscoveryService};
 use state_synchronizer::StateSynchronizer;
 use std::{
@@ -283,19 +283,16 @@ pub fn setup_environment(node_config: &mut NodeConfig) -> LibraHandle {
         );
 
         // Set up to listen for network configuration changes from StateSync.
-        match network_builder.conn_mgr_reqs_tx() {
-            Some(conn_mgr_reqs_tx) => {
-                let (simple_discovery_reconfig_subscription, simple_discovery_reconfig_rx) =
-                    ReconfigSubscription::subscribe(ON_CHAIN_CONFIG_REGISTRY);
-                reconfig_subscriptions.push(simple_discovery_reconfig_subscription);
-                let  network_config_listener = ConfigurationChangeListener::new(
-                    conn_mgr_reqs_tx,
-                    RoleType::Validator,
-                );
-                runtime.handle().spawn(network_config_listener.start(simple_discovery_reconfig_rx));
-            },
-            None => ()
-        }
+        if let Some(conn_mgr_reqs_tx) = network_builder.conn_mgr_reqs_tx() {
+            let (simple_discovery_reconfig_subscription, simple_discovery_reconfig_rx) =
+                ReconfigSubscription::subscribe(ON_CHAIN_CONFIG_REGISTRY);
+            reconfig_subscriptions.push(simple_discovery_reconfig_subscription);
+            let network_config_listener =
+                ConfigurationChangeListener::new(conn_mgr_reqs_tx, RoleType::Validator);
+            runtime
+                .handle()
+                .spawn(network_config_listener.start(simple_discovery_reconfig_rx));
+        };
 
         let (state_sync_sender, state_sync_events) =
             state_synchronizer::network::add_to_network(&mut network_builder);
--- a/libra-node/src/main_node.rs
+++ b/libra-node/src/main_node.rs
@@ -16,6 +16,7 @@ use libra_json_rpc::bootstrap_from_config as bootstrap_rpc;
 use libra_logger::prelude::*;
 use libra_mempool::MEMPOOL_SUBSCRIBED_CONFIGS;
 use libra_metrics::metric_server;
+use network_simple_onchain_discovery::ConfigurationChangeListener;
 use libra_types::{on_chain_config::ON_CHAIN_CONFIG_REGISTRY, waypoint::Waypoint, PeerId};
 use libra_vm::LibraVM;
 use libradb::LibraDB;
@@ -175,6 +176,7 @@ pub fn setup_network(
             .trusted_peers(trusted_peers)
             .seed_peers(seed_peers)
             .connectivity_check_interval_ms(config.connectivity_check_interval_ms)
+            // TODO:  Why is the connectivity manager related to remote_authentication?
             .add_connectivity_manager();
     } else if config.enable_noise {
         let identity_key = config
@@ -279,6 +281,21 @@ pub fn setup_environment(node_config: &mut NodeConfig) -> LibraHandle {
             Arc::clone(&db_rw.reader),
             node_config.base.waypoint.expect("No waypoint in config"),
         );
+
+        // Set up to listen for network configuration changes from StateSync.
+        match network_builder.conn_mgr_reqs_tx() {
+            Some(conn_mgr_reqs_tx) => {
+                let (simple_discovery_reconfig_subscription, simple_discovery_reconfig_rx) =
+                    ReconfigSubscription::subscribe(ON_CHAIN_CONFIG_REGISTRY);
+                reconfig_subscriptions.push(simple_discovery_reconfig_subscription);
+                let  network_config_listener = ConfigurationChangeListener::new(
+                    conn_mgr_reqs_tx,
+                    RoleType::Validator,
+                );
+                runtime.handle().spawn(network_config_listener.start(simple_discovery_reconfig_rx));
+            },
+            None => ()
+        }
 
         let (state_sync_sender, state_sync_events) =
             state_synchronizer::network::add_to_network(&mut network_builder);
--- a/network/simple-onchain-discovery/src/lib.rs
+++ b/network/simple-onchain-discovery/src/lib.rs
@@ -1,0 +1,193 @@
+// Copyright (c) The Libra Core Contributors
+// SPDX-License-Identifier: Apache-2.0
+
+use channel::libra_channel;
+use futures::{sink::SinkExt, StreamExt};
+use libra_config::config::RoleType;
+use libra_logger::prelude::*;
+use libra_metrics::{register_histogram, DurationHistogram};
+use libra_network_address::NetworkAddress;
+use libra_types::on_chain_config::{OnChainConfigPayload, ValidatorSet};
+use network::{
+    common::NetworkPublicKeys,
+    connectivity_manager::{ConnectivityRequest, DiscoverySource},
+};
+use once_cell::sync::Lazy;
+use std::{convert::TryFrom, time::Instant};
+
+/// Histogram of idle time of spent in event processing loop
+pub static EVENT_PROCESSING_LOOP_IDLE_DURATION_S: Lazy<DurationHistogram> = Lazy::new(|| {
+    DurationHistogram::new(
+        register_histogram!(
+            "simple_onchain_discovery_event_processing_loop_idle_duration_s",
+            "Histogram of idle time of spent in event processing loop"
+        )
+        .unwrap(),
+    )
+});
+
+/// Histogram of busy time of spent in event processing loop
+pub static EVENT_PROCESSING_LOOP_BUSY_DURATION_S: Lazy<DurationHistogram> = Lazy::new(|| {
+    DurationHistogram::new(
+        register_histogram!(
+            "simple_onchain_discovery_event_processing_loop_busy_duration_s",
+            "Histogram of busy time of spent in event processing loop"
+        )
+        .unwrap(),
+    )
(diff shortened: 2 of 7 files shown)


agc-quadra:
@davidiw @phlip9 

Putting the listener up as a standalone PR for the moment.   We can either push this as is or wait for the NetworkId consolidation in order to unblock the integration steps.

agc-quadra:
After talking with David off-line, I'm pushing the integration steps to explicitly bind the configuration listener to the validator network and remove the hook that goes from consensus to connectivity manager.

current state is that smoke tests fail.

libra-node/src/main_node.rs
```diff
@@ -298,6 +300,21 @@ pub fn setup_environment(node_config: &mut NodeConfig) -> LibraHandle {
             node_config.base.waypoint.expect("No waypoint in config"),
         );
 
+        // Set up to listen for network configuration changes from StateSync.
+        match network_builder.conn_mgr_reqs_tx() {
```
davidiw, line 304: let's use an expect, because otherwise we're in a really bad state.

davidiw: honestly the fact that we even have a "NetworkBuilder" without a connectivity manager seems ... wrong.

network/simple-onchain-discovery/src/lib.rs
```diff
@@ -0,0 +1,193 @@
...
+
+    // Collect the set of address updates.
+    let address_map = validator_keys
+        .iter()
+        .map(|validator| {
+            (
+                validator.account_address().clone(),
+                vec![
+                    NetworkAddress::try_from(&validator.config().validator_network_address)
+                        .expect("WTF"),
```
davidiw, line 61: let's log an error here and not panic.... someone could accidentally put bad data here, this shouldn't kill validators or other nodes.

network/simple-onchain-discovery/src/lib.rs
```diff
@@ -0,0 +1,193 @@
...
+    // Collect the set of updated addresses.
+    let address_map = full_node_infos
+        .iter()
+        .map(|full_node| {
+            (
+                full_node.account_address().clone(),
+                vec![NetworkAddress::try_from(
+                    &full_node.config().full_node_network_address.clone(),
+                )
+                .expect("WTF")],
```
davidiw, line 106: same comment as above: let's log an error here and not panic.... someone could accidentally put bad data here, this shouldn't kill validators or other nodes.

network/simple-onchain-discovery/src/lib.rs
```diff
@@ -0,0 +1,193 @@
...
+                (
+                    full_node.account_address().clone(),
+                    NetworkPublicKeys {
+                        identity_public_key: full_node
+                            .config()
+                            .full_node_network_identity_public_key
+                            .clone(),
+                        // TODO: this overload of the validator key onto the full_node network is
+                        // confusing at best and misleading at worst.
+                        signing_public_key: full_node
```
davidiw, line 129: I have a PR to delete the signing key :)

libra-node/src/main_node.rs
agc-quadra: > honestly the fact that we even have a "NetworkBuilder" without a connectivity manager seems ... wrong.

I can't say that I disagree, but it is what it is.  Question arises later as to why the connectivity manager is tied to an authenticated network.

agc-quadra: Where exactly do you want the expect heree?

network/simple-onchain-discovery/src/lib.rs
agc-quadra: Done.

network/simple-onchain-discovery/src/lib.rs
agc-quadra: Done.

agc-quadra:
Updated to address code comments, still one or two outsanding questions there.

Currently failing to pass two of the smoke tests non-deterministically due to occasional double logger instantiation:

** One run, locally**

---- smoke_test::test_basic_fault_tolerance stdout ----
Unable to set logger: attempted to set a logger after the logging system was already initialized
INFO 2020-06-01 07:42:52 common/workspace-builder/src/lib.rs:16 Building project binaries
INFO 2020-06-01 07:43:21 common/workspace-builder/src/lib.rs:30 Finished building project binaries
INFO 2020-06-01 07:43:21 testsuite/libra-swarm/src/swarm.rs:304 logs at Temporary(TempPath { path_buf: "/var/folders/vy/j71_b8zs27d_7zp6w9rtdjxr0000gn/T/127c787635efaf0991f8b1f36210512b", persist: false })
thread 'smoke_test::test_basic_fault_tolerance' panicked at 'called `Result::unwrap()` on an `Err` value: IO error: While open directory: /var/folders/vy/j71_b8zs27d_7zp6w9rtdjxr0000gn/T/f514edd4a58397f0e8a08452cd3fa3cd/libradb: Too many open files', testsuite/tests/libratest/smoke_test.rs:54:31
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace

---- smoke_test::smoke_test_multi_node stdout ----
Unable to set logger: attempted to set a logger after the logging system was already initialized
INFO 2020-06-01 07:43:21 testsuite/libra-swarm/src/swarm.rs:304 logs at Temporary(TempPath { path_buf: "/var/folders/vy/j71_b8zs27d_7zp6w9rtdjxr0000gn/T/2037ec751639514e812391cce9af817a", persist: false })
thread 'smoke_test::smoke_test_multi_node' panicked at 'called `Result::unwrap()` on an `Err` value: IO error: While open directory: /var/folders/vy/j71_b8zs27d_7zp6w9rtdjxr0000gn/T/3cce88320c50dda9259f7eaac8185c78/libradb: Too many open files', testsuite/tests/libratest/smoke_test.rs:54:31

** A second run, locally**
---- smoke_test::test_basic_state_synchronization stdout ----
Unable to set logger: attempted to set a logger after the logging system was already initialized
INFO 2020-06-01 07:48:37 testsuite/libra-swarm/src/swarm.rs:304 logs at Temporary(TempPath { path_buf: "/var/folders/vy/j71_b8zs27d_7zp6w9rtdjxr0000gn/T/47d20b38c0f6d39f2e9741f742c42b4b", persist: false })
thread 'smoke_test::test_basic_state_synchronization' panicked at 'called `Result::unwrap()` on an `Err` value: IO error: While open directory: /var/folders/vy/j71_b8zs27d_7zp6w9rtdjxr0000gn/T/c657ead3e6c1bb5636cc72b5a47a1cd7/libradb: Too many open files', testsuite/tests/libratest/smoke_test.rs:54:31
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace

---- smoke_test::test_basic_restartability stdout ----
Unable to set logger: attempted to set a logger after the logging system was already initialized
INFO 2020-06-01 07:48:37 testsuite/libra-swarm/src/swarm.rs:304 logs at Temporary(TempPath { path_buf: "/var/folders/vy/j71_b8zs27d_7zp6w9rtdjxr0000gn/T/290d6298e13a810f5f538cab4b0079f5", persist: false })
thread 'smoke_test::test_basic_restartability' panicked at 'called `Result::unwrap()` on an `Err` value: IO error: While open directory: /var/folders/vy/j71_b8zs27d_7zp6w9rtdjxr0000gn/T/22339f6e5358d46c596d80ba47dd5aaf/libradb: Too many open files', testsuite/tests/libratest/smoke_test.rs:54:31

libra-node/src/main_node.rs
```diff
@@ -298,6 +300,18 @@ pub fn setup_environment(node_config: &mut NodeConfig) -> LibraHandle {
             node_config.base.waypoint.expect("No waypoint in config"),
         );
 
+        // Set up to listen for network configuration changes from StateSync.
+        if let Some(conn_mgr_reqs_tx) = network_builder.conn_mgr_reqs_tx() {
```
davidiw, line 304: This returns an option, we should use an expect to enforce that this is some or panic otherwise:
https://doc.rust-lang.org/std/option/enum.Option.html#method.expect

network/simple-onchain-discovery/src/lib.rs
```diff
@@ -0,0 +1,216 @@
...
+fn extract_full_node_updates(full_node_set: ValidatorSet) -> Vec<ConnectivityRequest> {
+    let full_node_infos = full_node_set.payload().to_vec();
+
+    let mut updates = Vec::new();
+
+    // Collect the set of updated addresses.
+    let address_map = full_node_infos
+        .iter()
+        .filter(|full_node| {
+            match NetworkAddress::try_from(&full_node.config().full_node_network_address) {
```
davidiw, line 113: https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.filter_map

filter_map will let you reduce code and run-time complexity (likely negligible on the latter) but most importanly remove this ugly expect statement below

note, I would probably do something like this:
```
map_filter(|fn|
  NetworkAddress::try_from(fn.config().full_node_network_address)
    .map_err(|e| {
       log_err;
       e
    })
    .ok()
)
```

or

```
let addr =  NetworkAddress::try_from(fn.config().full_node_network_address);
if let Err(err) = addr {
   log_err
}
addr.ok()
```

libra-node/src/main_node.rs
agc-quadra: Why should we panic here?  we only have a connectivity_manager if _enable\_remote\_authentication is enabled (main_node.rs:146), which is not guaranteed to be the case.

network/simple-onchain-discovery/src/lib.rs
```diff
@@ -0,0 +1,216 @@
...
+                Err(e) => {
+                    warn!(
+                        "Cannot parse validator network address from validator info: {}",
+                        e
+                    );
+                    false
+                }
+            }
+        })
+        .map(|validator| {
```
zekun000, line 68: I think you can squash this two closure into one with flat_map https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.flat_map
```
flat_map(|validator| {
  match NetworkAddress::try_from(&validator.config().validator_network_address) {
    Ok(addr) => Some((*validator.account_address(), addr),
    Err(e) => {
       warn!();
       None
    }
  }
}).collect()
```

network/simple-onchain-discovery/src/lib.rs
zekun000: oh of course david will stand up for this

zekun000: I'm used to use flat_map for this but anyway this two closures look ugly
```
flat_map(|node| {
  match NetworkAddress::try_from(&full_node.config().full_node_network_address) {
    Ok(addr) => Some((..)),
    Err(e) => {
       warn!(..);
       None
    }
  }
})
```

network/simple-onchain-discovery/src/lib.rs
```diff
@@ -0,0 +1,216 @@
...
+    /// Processes a received OnChainConfigPayload.  Depending on role (Validator or FullNode), parses
+    /// the appropriate configuration changes and passes it to the ConnectionManager channel.
+    async fn process_payload(&mut self, payload: OnChainConfigPayload) {
+        let node_set: ValidatorSet = payload
+            .get()
+            .expect("failed to get ValidatorSet from payload");
+
+        let updates = match self.role {
+            RoleType::Validator => extract_validator_updates(node_set),
+            RoleType::FullNode => extract_full_node_updates(node_set),
```
zekun000, line 183: these two functions still look like the same except the address it's extracted, probably can be simplified with a lambda just to get the address,
```
extract_updates(validator_set: ValidatorSet, address: Fn(Config) -> NetworkAddress)
```

libra-node/src/main_node.rs
davidiw: we cannot form a network otherwise :)

let's just panic for now and we can deal with the plumbing issues later.

of course one could argue we have bigger issues if someone runs this without auth...

I'll leave it up to you.

network/simple-onchain-discovery/src/lib.rs
davidiw: I think my solution looks better than yours :P -- well the second one at least.

libra-node/src/main_node.rs
agc-quadra: In the essence of minimal changes, I'll defer removing the `enable_remote_authentication` to a follow up.  
1). it's feature creep
2). I lack the context to make that judgement call/appreciate implications thereof
3). 9 hour round trips exacerbate (2)

network/simple-onchain-discovery/src/lib.rs
agc-quadra: address and public_key;  but yes.

libra-node/src/main_node.rs
agc-quadra: Will defer removing enable_remote_authentication for a couple of reasons:

1). scope creep
2). I'm missing context to make judgement calls
3). timezones will make it inefficient for me to do that removal

network/simple-onchain-discovery/src/lib.rs
agc-quadra: done.

network/simple-onchain-discovery/src/lib.rs
agc-quadra: flat_map wins.

agc-quadra:
@zekun000 @davidiw @phlip9 
 Now failing to launch the cluster test.    any insights into how to debug that are greatly appreciated.

davidiw:
Flaky tests for the win @bors-libra r+

davidiw:
Neat, looks like the manual re-run was good enough for bors to land it, I'm a little confused what its trying to do now...

PR #4141, opened 2020-05-29
[x] rewrite test-only lint as a package classifier 

sunshowers:
We now enforce that the list of default members consists of
binary crates, and binary crates only.

A binary crate also includes dylib, cdylib and staticlib.

execution-correctness, which is one of the binaries we care about the most, wasn't listed in default-members.


devtools/x/src/lint/workspace_classify.rs
```diff
@@ -0,0 +1,128 @@
...
+                        .next();
+                }
+                return None;
+            })
+            .next();
+
+        let default_root = default_members.is_root_member(package.id());
+        let default_reachable = default_members.contains(package.id());
+
+        match (binary_kind, default_root, default_reachable) {
```
sunshowers, line 73: Here's the core decision table that determines where things should go.

rexhoffman: good pruning mechanism.  dig it.

Issue #4125, opened 2020-05-29
[Bug][Consensus] deadlock between peers


zekun000:
It's deadlocked with this short circuit 
- https://github.com/libra/libra/blob/master/consensus/src/round_manager.rs#L377

a.is_stale(b) == true can't infer b.is_stale(a) == false, both statement can be true and should be handled, otherwise two peers will think others are stale and keep sending it back sync info.

The two statement can both be true when a has higher round than b but b has higher commit than a.

PR #4142, opened 2020-05-30
[consensus] fix deadlock in sync_up

zekun000:
## Motivation

The short circuit return after help remote causes the deadlock because
the is_stale check is not symmetric, a.is_stale(b) == true and
b.is_stale(a) == true can both hold and result in deadlock this case (peers both think others are stale and bounce sync info back and forth without processing).

This commit removes the return statement which fixed the problem and rename
is_stale to has_newer_certificate (hopefully it's less confusing).

Y

## Test Plan

Added tests.

## Related PRs
https://github.com/libra/libra/issues/4125


zekun000:
yeah, it stops the pre-mainnet, it does need some weird network condition to trigger

PR #4143, opened 2020-05-30
[move prover] warn unused schema when modules are targets and escape …

emmazzz:
…with UNUSED prefix

## Motivation

This PR improves warning experience such that Move Prover only warns about unused schemas defined in target modules and who names don't start with "UNUSED".

Yes.

## Test Plan

Added in the test file a schema whose name starts with UNUSED and gets ignored.

### DIFF
--- a/language/move-prover/spec-lang/src/translate.rs
+++ b/language/move-prover/spec-lang/src/translate.rs
@@ -68,7 +68,7 @@ pub struct Translator<'env> {
     spec_schema_table: BTreeMap<QualifiedSymbol, SpecSchemaEntry>,
     /// A symbol table storing unused schemas, used later to generate warnings. All schemas
     /// are initially in the table and are removed when they are used in expressions.
-    unused_schema_table: BTreeMap<QualifiedSymbol, Loc>,
+    unused_schema_set: BTreeSet<QualifiedSymbol>,
     // A symbol table for structs.
     struct_table: BTreeMap<QualifiedSymbol, StructEntry>,
     /// A reverse mapping from ModuleId/StructId pairs to QualifiedSymbol. This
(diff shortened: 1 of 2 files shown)


emmazzz:
@wrwg

language/move-prover/spec-lang/src/translate.rs
```diff
@@ -2686,6 +2678,24 @@ impl<'env, 'translator> ModuleTranslator<'env, 'translator> {
             std::mem::take(&mut self.instantiation_map),
             std::mem::take(&mut self.spec_block_infos),
         );
+        // Warn about unused schemas.
```
emmazzz, line 2681: I moved it here so that module_env is available.

language/move-prover/spec-lang/src/translate.rs
```diff
@@ -2686,6 +2678,24 @@ impl<'env, 'translator> ModuleTranslator<'env, 'translator> {
...
+        for name in &self.parent.unused_schema_set {
+            let entry = self
+            .parent
+            .spec_schema_table
+            .get(&name)
+            .expect("schema defined");
+            let schema_name = name.display_simple(self.symbol_pool()).to_string();
+            let module_env = self.parent.env.get_module(entry.module_id);
```
wrwg, line 2689: This unfortunately doesn't work, because the translator builds the module_env's incrementally up in the GlobalEnv. That is if you have modules 1..N and you are translating module N, then the environment only contains modules 1..(N-1). The result of translation is to add module N to the environment. (For background, 1..N are organized in topological use order.)

emmazzz: But this is at the end of module translation inside function `populate_env_from_result` and after the module has been added to the global env(on line 2667). So module N should be available in the environment at this point?

language/move-prover/spec-lang/src/translate.rs
wrwg: Ah, SG then!

PR #4144, opened 2020-05-30
update coverage reporting

tzakian:


language/ir-testsuite/tests/expressions/my_address_equality.mvir
```diff
@@ -0,0 +1,25 @@
+main() {
```
huitseeker, line 1: Should this be included in this PR?

tzakian: This PR was a test of the bot. I added this as change so that it would be triggered if it was still active.

PR #4145, opened 2020-05-30
[CI] detect CLI breaking change

sausagee:
## Motivation
We want to find out if there is CLI breaking change after a PR lands in
master. The signal is expected to give the downstream client
applications a heads-up of impending integration breakage. For now, we
don't want to block the PR from landing due to breaking change. However,
this could change when we start stabilization.

The new workflow builds the CLI, connects to testnet and attempts to
execute a few simple transactions. The workflow will run for every PR
landed in master until the first breaking change is detected.

The known limitation is that if the PR doesnt squash its commits before
landing, the workflow will only identify and alert the first commit in
the PR that causes breakage. Also, the workflow does not support bisect
at the moment.

## Test Plan
CI
1. Canary with 4edd44c in https://app.circleci.com/pipelines/github/libra/libra/18135/workflows/672c85aa-b59a-499a-a873-da8cb8020d2a/jobs/102215/steps
We already had a breaking change in trunk since this week's testnet release (4bc10151d) on Wednesday, so the job failed. The failed git rev (the canary itself) is cached.

image

2. Re-run the same canary in https://app.circleci.com/pipelines/github/libra/libra/18135/workflows/8de030f2-6882-4732-aed9-408c5b8082a0/jobs/102232/steps
The workflow hit the cache in an early step and the remaining steps were skipped.  Thus the job passed.

image

3. Canary with cac44df in https://app.circleci.com/pipelines/github/libra/libra/18139/workflows/9d795da3-1660-4a14-ab1f-fb52235caf0d/jobs/102241/steps
The canary force-checked out a51f08157 (the commit right after origin/testnet, 4bc10151d). Since the breaking change is further down, the CLI check passed.

image

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,5 +1,8 @@
 version: 2.1
 
+orbs:
+  slack: circleci/slack@3.3.0
+
 executors:
   build-executor:
     docker:
@@ -91,6 +94,7 @@ commands:
     description: Save cargo package cache for subsequent jobs
     steps:
       - save_cache:
+          name: Save cargo package cache
           key: cargo-package-cache-{{ checksum "Cargo.lock" }}
           # paths are relative to /home/circleci/project/
           paths:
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -101,12 +105,29 @@ commands:
...
           name: Check cargo package cache
           command: |
             ls -all /usr/local/cargo
             du -ssh /usr/local/cargo
+  save_breaking_change_rev:
+    description: Save the breaking change rev since last testnet update.
+    steps:
+      - save_cache:
+          name: Save breaking change rev
+          key: testnet-{{ checksum "testnet_rev" }}
```
metajack, line 120: Do you have to hardcode the paths even though it's in an ENV var in the `check-breaking-change` step?

sausagee: Unfortunately I have to hard coded. tldr - the save_cache/restore_cache cmds don't take random ENV VARs we defined. The cache key is literally the `testnet-${TESTNET_REV}` string.  There is an explanation note in lines 458 and 468 below.

PR #4146, opened 2020-05-30
[network] Remove network signing key

davidiw:
In my attempts to make it possible to place and read the identity key into vault, I explored many different avenues to handling keys. Ultimately what I realized is that this key is a pain and an unnecessary pain. Let me copy my blurb from my first commit and place it below. The TL;DR, is that this is not production code, it will never be production code as it is.

Please take a quick review of this PR, it sprawls across the code base and will be a nightmare to rebase :).

We only have gossip so that we can instantiate networks on the fly --
for test purposes. We had planned to remove gossip, but there will be a
slow movement from gossip to onchain discovery because test networks --
work so well... So we are left with a dilema. Do we maintain the
complexity of the existing gossip network, which impacts our network
configuration substantially -- just for a testing api or do we
substantially simplify the testing framework to not impact the
production environment?

This PR says tests will be tests and compromises towards simplifying our
production deployment. With that in mind, there are no more verified
notes. gossip is "insecure".

What does it mean for gossip to be insecure? Well, some really evil
sneaky peer (on a test network) could overwhelm us with peers that we
don't know about or eliminate our ability to leverage gossip from other
peers. That's not such a big issue:
1) Testnetworks are private to a single organization so passing around
these gossip messages is ... not possible.
2) If we are concerned about some sort of attack on overwhelming us, we
could substnatially improve this -- and we can -- it isn't production
code, so let's not over think this.

By doing this, we can eliminate signing keys from all configuration,
which will make our genesis construction so much simpler too...


davidiw:
It seems that the mirai bot is flaky at this point in time.

testsuite/cli/src/client_proxy.rs
```diff
@@ -568,12 +568,11 @@ impl ClientProxy {
...
-        let network_signing_key = Ed25519PublicKey::from_encoded_string(space_delim_strings[4])?;
-        let network_identity_key = x25519::PublicKey::from_encoded_string(space_delim_strings[5])?;
-        let network_address = NetworkAddress::from_str(space_delim_strings[6])?;
+        let network_identity_key = x25519::PublicKey::from_encoded_string(space_delim_strings[4])?;
+        let network_address = NetworkAddress::from_str(space_delim_strings[5])?;
         let network_address = RawNetworkAddress::try_from(&network_address)?;
-        let fullnode_identity_key = x25519::PublicKey::from_encoded_string(space_delim_strings[7])?;
-        let fullnode_network_address = NetworkAddress::from_str(space_delim_strings[8])?;
+        let fullnode_identity_key = x25519::PublicKey::from_encoded_string(space_delim_strings[6])?;
+        let fullnode_network_address = NetworkAddress::from_str(space_delim_strings[7])?;
```
gregnazario, line 575: Is there a diagram somewhere that says what the order of these are?  Just for future reference could be helpful.

config/src/config/network_config.rs
```diff
@@ -151,7 +149,8 @@ impl NetworkConfig {
             if !network_role.is_validator() && self.enable_remote_authentication {
                 ensure!(
                     self.peer_id == peer_id,
-                    "For full-nodes that use remote authentication, the peer_id must be derived from the identity key.",
+                    "For full-nodes that use remote authentication,
+                    the peer_id must be derived from the identity key.",
```
gregnazario, line 153: Minor: Just in case you didn't realize this will cause a 2 line string of the form:
```
 For full-nodes that use remote authentication,
                    the peer_id must be derived from the identity key.
```
(See: https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=282e9c9418d2e4cd26b75bdb5d84ae46)
It might be a little ugly

mimoo:
MIRAI-bot: cc @junkil-park 

@davidiw I suggest adding more context in the PR description for plebs like me

junkil-park:
You can ignore Move Prover's complaint about "Move prover returns: boogie exited with status None" for now. This seems to be an issue on the Prover side, and we are currently investigating it. Sorry for the inconvenience.

config/src/config/network_config.rs
phlip9: right, should be 

```rust
                    "For full-nodes that use remote authentication, \
                     the peer_id must be derived from the identity key.",
```

phlip9:
beautiful! this simplifies things a lot : )

@bors-libra delegate+

testsuite/cli/src/client_proxy.rs
davidiw: Maybe @zekun000 knows, I don't know if anyone else is using this at this point in time and is unlikely to be useful come V1. At that point, management or something more in that domain is actually the tool to add / remove validators.

davidiw:
Thanks everyone for the super fast reviews :D

bmwill:
I still feel like having this gossip mechanism is needed to mitigating situations where the on-chain data may be unable to be updated so I'm glad you didn't opt to remove it. (This may have been hardened or dealt with in some other ways by this point though).

Having said that I think that dropping the signing key is probably the right move since even if someone flooded you with bogus notes you'd quickly figure out they were bogus when you tried to connect and the noise handshake failed. So honestly its probably not inherently less secure without the signing (given we're a little smart about how we update stored notes)

config/src/config/network_config.rs
huitseeker, line 153: Nit: error message spaceing.

network/src/validator_network/network_builder.rs
```diff
@@ -413,8 +402,6 @@ impl NetworkBuilder {
         let conn_mgr_reqs_tx = self
             .conn_mgr_reqs_tx()
             .expect("ConnectivityManager not enabled");
-        let (signing_private_key, _signing_public_key) =
-            self.signing_keypair.take().expect("Signing keys not set");
```
huitseeker, line 417: Would it be appropriate to at least comment on the future of `add_gossip_discovery(&mut self)` above? maybe a [deprecation](https://doc.rust-lang.org/edition-guide/rust-2018/the-compiler/an-attribute-for-deprecation.html)?

config/src/config/network_config.rs
huitseeker, line 153: Nit: error message spaceing.

network/src/validator_network/network_builder.rs
huitseeker, line 417: Would it be appropriate to at least comment on the future of `add_gossip_discovery(&mut self)` above? maybe a [deprecation](https://doc.rust-lang.org/edition-guide/rust-2018/the-compiler/an-attribute-for-deprecation.html)?

config/src/config/network_config.rs
davidiw: The readme is super out of date, I cannot touch that in this PR 🐑

davidiw:
@bors-libra r=gregnazario

Note: this only breaks changes for validators in testnet.

PR #4147, opened 2020-05-31
[easy] Remove a few panics

huitseeker:
This collapses a potential panic into the `Result` of its enclosing function, keeping the error semantics but turning those into a managed outcome.

By request [[1]](https://github.com/libra/libra/pull/3879#issuecomment-628976319) [[2]](https://github.com/libra/libra/pull/3879#issuecomment-629417720) this does NOT include eight such panics in `language/` that could be so collapsed in the `Error` case of their `Result`:
https://gist.github.com/huitseeker/e6272a1b523e033979c29ebb7d4cf072

### DIFF
--- a/json-rpc/src/methods.rs
+++ b/json-rpc/src/methods.rs
@@ -111,8 +111,8 @@ async fn get_account_state(
     let currency_info = currencies_info(service, request).await?;
     let currencies: Vec<_> = currency_info
         .into_iter()
-        .map(|info| from_currency_code_string(&info.code).unwrap())
-        .collect();
+        .map(|info| from_currency_code_string(&info.code))
+        .collect::<Result<_, _>>()?;
     if let Some(blob) = response {
         let account_state = AccountState::try_from(&blob)?;
         if let Some(account) = account_state.get_account_resource()? {
--- a/network/src/peer_manager/mod.rs
+++ b/network/src/peer_manager/mod.rs
@@ -171,12 +171,10 @@ impl PeerManagerRequestSender {
             res_tx,
             timeout,
         };
-        self.inner
-            .push(
-                (peer_id, protocol),
-                PeerManagerRequest::SendRpc(peer_id, request),
-            )
-            .unwrap();
+        self.inner.push(
+            (peer_id, protocol),
+            PeerManagerRequest::SendRpc(peer_id, request),
+        )?;
         res_rx.await?
     }
 }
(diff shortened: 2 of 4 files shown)


zekun000:
sometimes unwrap is used to indicate unreachable, switching it to error seems changing the behavior?

json-rpc/src/methods.rs
```diff
@@ -111,8 +111,8 @@ async fn get_account_state(
     let currency_info = currencies_info(service, request).await?;
     let currencies: Vec<_> = currency_info
         .into_iter()
-        .map(|info| from_currency_code_string(&info.code).unwrap())
-        .collect();
+        .map(|info| from_currency_code_string(&info.code))
+        .collect::<Result<_, _>>()?;
```
davidiw, line 115: this looks like a good catch here, if somehow there were an errant code, this would result in an attack on this service...

network/src/peer_manager/mod.rs
```diff
@@ -171,12 +171,10 @@ impl PeerManagerRequestSender {
             res_tx,
             timeout,
         };
-        self.inner
-            .push(
-                (peer_id, protocol),
-                PeerManagerRequest::SendRpc(peer_id, request),
-            )
-            .unwrap();
+        self.inner.push(
```
davidiw, line 174: agree, without a comment, it is unclear if this unwrap is intentional, for example, the node could be so overloaded that we want to explicitly crash the service. so this seems okay.

vm-validator/src/mocks/mock_vm_validator.rs
```diff
@@ -41,8 +41,7 @@ impl TransactionValidation for MockVMValidator {
         };
 
         let sender = txn.sender();
-        let account_dne_test_add =
-            AccountAddress::try_from(&[0 as u8; AccountAddress::LENGTH]).unwrap();
+        let account_dne_test_add = AccountAddress::try_from(&[0 as u8; AccountAddress::LENGTH])?;
```
davidiw, line 44: this one... is weird, there are a whole bunch below, but only this one is different, this should really be unified

huitseeker:
> sometimes unwrap is used to indicate unreachable, switching it to error seems changing the behavior?

Sometimes, and it's a poor substitute for the much better `unreachable!()` at that. By default, unwrap is a panic liability. The present PR tries to fix the later.

zekun000:
well, unreachable()! requires substantial more code and is not ergonomic to write.

vm-validator/src/mocks/mock_vm_validator.rs
huitseeker: :facepalm:  cleaned that up

PR #4150, opened 2020-06-01
[core modules] Change transaction fee distribution method

tzakian:
Transaction fees are no longer distributed to each validator, but instead paid to the association.

This is done by adding the collected fees to a preburn resource for transaction fees, the association then needs to send a transaction to burn these coins and can then extract the underlying fiat.


language/libra-vm/src/libra_vm.rs
```diff
@@ -528,21 +524,6 @@ impl LibraVM {
             return Err(VMStatus::new(StatusCode::MALFORMED));
         };
 
-        // Now distribute the txn fees for each currency
```
sblackshear, line 531: <3

language/move-lang/tests/functional/transaction_fee/burn_collected_fees.move
```diff
@@ -0,0 +1,54 @@
...
+//! new-transaction
+//! sender: bob
+//! max-gas: 1000
+//! gas-price: 1
+//! gas-currency: Coin1
+script {
+    fun main() { while (true) {} }
+}
+// check: OUT_OF_GAS
+
```
sblackshear, line 22: Lol, love this scheme for collecting a large amount of fees...

language/stdlib/modules/libra.move
```diff
@@ -288,13 +288,15 @@ module Libra {
...
-    // the `preburn` resource has no pending burn requests.
+    // Can only be invoked by the holder of the `BurnCapability`. Does
+    // nothing if the `preburn` resource has no pending burn requests.
     public fun burn_with_resource_cap<Token>(
         preburn: &mut Preburn<Token>,
         preburn_address: address,
         _capability: &BurnCapability<Token>
     ) acquires CurrencyInfo {
+        // If there aren't any preburn requests don't try to burn any
+        if (Vector::is_empty(&preburn.requests)) return;
```
sblackshear, line 299: Would it be possible for this to be an assert rather than a silent return? My sense is that failing loudly in this case rather than succeeding + doing nothing is easier to debug (and easier to specify for the prover as well).

language/stdlib/modules/transaction_fee.move
```diff
@@ -1,105 +1,87 @@
...
 
-    // Initialize the transaction fee distribution module. We keep track of the last paid block
-    // height in order to ensure that we don't try to pay more than once per-block. We also
-    // encapsulate the withdrawal capability to the transaction fee account so that we can withdraw
-    // the fees from this account from block metadata transactions.
-    public fun initialize_transaction_fees(fee_account: &signer) {
+    resource struct LBRIdent<LBRCoinType> { coin: Libra<LBRCoinType> }
+
+    // Called in genesis. Once for each fiat currency.
+    public fun add_transaction_fee<CoinType>(fee_account: &signer, preburn: Libra::Preburn<CoinType>) {
```
sblackshear, line 20: I find this name a bit confusing; it sounds like it might be paying a fee. Maybe something like `add_currency_type`?

language/stdlib/modules/transaction_fee.move
```diff
@@ -1,105 +1,87 @@
...
-    // We have made a couple design decisions here that are worth noting:
-    //  * We pay out once per-block for now.
-    //    TODO: Once we have a better on-chain representation of
-    //          epochs this should be changed over to be once per-epoch.
-    //  * Sometimes the number of validators does not evenly divide the transaction fees to be
-    //    distributed. In such cases the remainder ("dust") is left in the transaction fees pot and
-    //    these remaining fees will be included in the calculations for the transaction fee
-    //    distribution in the next epoch. This distribution strategy is meant to in part minimize the
-    //    benefit of being the first validator in the validator set.
+    resource struct FeePaymentCapability { }
```
sblackshear, line 11: Could you add a doc comment for each of these resources?

language/stdlib/modules/transaction_fee.move
```diff
@@ -1,105 +1,87 @@
...
-        let distribution_resource = borrow_global<TransactionFees>(0xFEE);
-        let index = 0;
+    // Called for LibraAccount setup
+    public fun grant_fee_payment_capability(libra_account_singleton_addr: &signer): FeePaymentCapability {
+        Transaction::assert(Signer::address_of(libra_account_singleton_addr) == 0xA550C18, 0);
+        FeePaymentCapability {}
+    }
 
-        while (index < num_validators) {
+    public fun pay_fee<CoinType>(fee: Libra<CoinType>, _cap: &FeePaymentCapability)
```
sblackshear, line 37: Wondering: do we need `FeePaymentCapability`? It doesn't seem particularly problematic if someone wants to send funds to the fee pot, and it's not possible without custom transaction scripts anyway.

language/stdlib/modules/transaction_fee.move
```diff
@@ -1,105 +1,87 @@
...
+    acquires CollectedFees, LBRIdent {
+        // For LBR we need to extract the underlying coins from the
+        // on-chain reserve and then push these to the
```
sblackshear, line 40: Need to finish comment?

language/stdlib/modules/transaction_fee.move
```diff
@@ -1,105 +1,87 @@
...
-    // This calculates the amount to be distributed to each validator equally. We do this by calculating
-    // the integer division of the transaction fees collected by the number of validators. In
-    // particular, this means that if the number of validators does not evenly divide the
-    // transaction fees collected, then there will be a remainder that is left in the transaction
-    // fees pot to be distributed later.
-    fun per_validator_distribution_amount(amount_collected: u64, num_validators: u64): u64 {
-        Transaction::assert(num_validators != 0, 0);
-        let validator_payout = amount_collected / num_validators;
-        validator_payout
+    public fun burn_fees<CoinType>(burn_cap: &Libra::BurnCapability<CoinType>)
```
sblackshear, line 77: Hmm, interesting requirements question here. This lets us burn fees in only one currency at a time; I'm not sure if what we want is that or "burn fees in all currencies at once" (we should ask). The latter might actually be easier to implement; it will let us get rid of all the special casing for LBR.

language/stdlib/modules/transaction_fee.move
```diff
@@ -1,105 +1,87 @@
...
-    //    distributed. In such cases the remainder ("dust") is left in the transaction fees pot and
-    //    these remaining fees will be included in the calculations for the transaction fee
-    //    distribution in the next epoch. This distribution strategy is meant to in part minimize the
-    //    benefit of being the first validator in the validator set.
+    resource struct FeePaymentCapability { }
 
-    resource struct TransactionFees {
-        fee_withdrawal_capability: LibraAccount::WithdrawalCapability,
+    resource struct CollectedFees<CoinType> {
+        preburn: Libra::Preburn<CoinType>,
```
sblackshear, line 14: Might be useful to have a getter that allows you to read the quantity of collected fees.

language/transaction-builder/src/lib.rs
```diff
@@ -116,6 +116,15 @@ encode_txn_script! {
           `MintCapability` stored under their account and `preburn_address` has a pending burn request"
 }
 
+encode_txn_script! {
```
sblackshear, line 119: Would you mind adding a basic test for this to `e2e-tests/src/tests/transaction-builder.rs`? I try to write a test for every script there to make sure the types of the Rust `encode_`  function stay in sync with the Move tx script (have  been saved by this several times while refactoring).

language/stdlib/modules/libra.move
tzakian: It's a good question and I was going back and forth. If we don't have this we'll need to have some kind of "num_preburn_requests" function to return the number of preburn requests in a preburn resource (since if no txn fees in a given currency have been collected the burn will abort)

language/stdlib/modules/transaction_fee.move
tzakian: How does `add_transaction_fee_currency` sound?

language/stdlib/modules/transaction_fee.move
tzakian: ah! yes. Will do this.

language/stdlib/modules/transaction_fee.move
tzakian: More than happy to take it away :) I don't think there should be adverse consequences, just that other people can pay into it with custom transaction scripts.

language/stdlib/modules/transaction_fee.move
sblackshear: `transaction_fee` part sounds a bit redundant given the module name, but not picky :)

language/stdlib/modules/transaction_fee.move
```diff
@@ -1,105 +1,87 @@
...
-           }
+            let collected_fee_coin2 = borrow_global_mut<CollectedFees<Coin2>>(0xFEE);
+            Libra::preburn_with_resource(
+                coin2,
+                &mut collected_fee_coin2.preburn,
+                0xFEE
+            );
+        } else {
+            let collected_fee = borrow_global_mut<CollectedFees<CoinType>>(0xFEE);
+            Libra::preburn_with_resource(
```
sblackshear, line 69: Just realized: I think this will add a new preburn request to the `Preburn.requests` for every transaction. That will get big fast, and it will also require one `BurnTxnFees` transaction for reach fee-paying transaction. We probably want to design this differently, but not sure what the best approach is:
1. Pay fees into an DD-like Association account, which can then preburn/burn to liquidate the fees
2. Don't pay fees into the `Preburn` resource; use a dedicated resource + add logic for burning from that instead
3. Something else?

My bias is toward (1) because it seems simplest. It does mean that the Association could choose to send the funds somewhere instead of burning them, but that seems ok.

tzakian: You are so right... this will add a preburn request for each txn...which is definitely not good.

What about this (which I think is 1?):
* `0xFEE` holds coins (or some resource containing coins) for each currency.
* Transaction fees are deposited into these.
* the `0xB1E55ED` account would hold withdrawal capability on `0xFEE`. Since it has a`BurnCapability` for each cointype (except LBR, and we can do the unpacking to handle this)
* The `0xB1E55ED` account would have preburn resources for each currency != LBR.

We can then have the transaction be something like:
```rust
fun main<CoinType>(blessed: &signer) {
    // uses the withdrawal capability underneath
	let coins = TransactionFee::withdraw_fees<CoinType>();
    Libra::preburn_to_sender(coins);
    Libra::burn(Signer::address_of(blessed));
}
```

sblackshear: Love this solution!

language/stdlib/modules/transaction_fee.move
tzakian: This can now be done by just looking at the balance for the currency under `0xFEE`.

language/stdlib/modules/transaction_fee.move
tzakian: This should be implemented now. A couple notes:
* The burning is per currency (since that's what sounded would be most in line with AOS). Not hard to change if we decide against this though.
* The transaction fees are stored in the balance's for the account at `0xFEE`.
* the TC account now holds preburn resources for the fiat currencies. We still need to special case LBR in the script, but it's not nearly as awful as before.

language/stdlib/modules/transaction_fee.move
```diff
@@ -1,105 +1,104 @@
...
-        let index = 0;
-
-        while (index < num_validators) {
+    /// Sets ups the needed transaction fee state for a given `CoinType`
+    /// currency.
+    public fun add_txn_fee_currency<CoinType>(fee_account: &signer, burn_cap: &BurnCapability<CoinType>) {
+        Transaction::assert(Signer::address_of(fee_account) == 0xFEE, 0);
+        LibraAccount::add_currency<CoinType>(fee_account);
+        move_to(fee_account, TransactionFeePreburn<CoinType>{
+            preburn: Libra::new_preburn_with_capability(burn_cap)
```
moeziniaf, line 47: can we maybe get rid of publish_preburn and new_preburn? what's the use of un_approved Preburn resource?

language/stdlib/modules/transaction_fee.move
```diff
@@ -1,105 +1,104 @@
...
+        if (is_lbr<CoinType>()) {
+            let amount = LibraAccount::balance<LBR>(0xFEE);
+            let coins = LibraAccount::withdraw_with_capability<LBR>(
+                &borrow_global<TransactionFeeCollection>(Signer::address_of(blessed_sender)).cap,
+                amount
             );
-           }
+            let (coin1, coin2) = LBR::unpack(coins);
+            preburn_coin<Coin1>(coin1);
+            preburn_coin<Coin2>(coin2)
```
moeziniaf, line 71: cool!

language/stdlib/modules/transaction_fee.move
```diff
@@ -1,105 +1,104 @@
...
-        move_to(
-            fee_account,
-            TransactionFees {
-                fee_withdrawal_capability: LibraAccount::extract_sender_withdrawal_capability(),
-            }
-        );
+    /// The `TransactionFeePreburn` holds a preburn resource for each
+    /// fiat `CoinType` that can be collected as a transaction fee.
+    resource struct TransactionFeePreburn<CoinType> {
+        preburn: Preburn<CoinType>
```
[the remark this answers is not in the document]
tzakian: Good question! This only holds one preburn resource for each coin type. So under the 0xFEE account we'd see two of these published: `TransactionFeePreburn<Coin1>` and `TransactionFeePreburn<Coin2>`.

language/stdlib/modules/transaction_fee.move
tzakian: I believe we had these for testing purposes in testnet, but I think they can be removed now. @sblackshear do you know of any reason to keep these around? I can't think of any remaining reasons.

PR #4151, opened 2020-06-01
[move core modules] payments no longer create accounts

sblackshear:
In  the 6/18 testnet release of Libra, the payment script created a new account if the receiving address did not exist. This behavior does not match the requirements of Libra V1:
- There are now several different account roles (`ParentVASP`, `ChildVASP`, `Validator`, `DesignatedDealer`, ...), so it's not clear what kind of account we should create for the recipient
- The creation of most roles is privileged (e.g., a `ChildVASP` cannot create a `ParentVASP`).

This PR removes the old behavior and (along with it) the `auth_key_prefix` parameter of the payment script. This parameter is only used if the payment is creating a new account.

This may be disruptive to existing clients that were relying on the old creation behavior. Clients that wish to create accounts in testnet should either:
- Create a `ChildVASP` account from an existing `ParentVASP` account using the appropriate [script](https://github.com/libra/libra/blob/master/language/stdlib/transaction_scripts/create_child_vasp_account.move)
- Create a `ParentVASP` account via the [faucet](https://developers.libra.org/docs/my-first-transaction#add-libra-coins-to-alices-and-bobs-accounts)

There are two loose ends for account creation that need to be removed in follow-up PRs:
- The `CreateAccount` script
- The `PeerToPeerScript` (the payment script described above is `PeerToPeerWithMetadata`)

## Motivation

Getting closer to the production configuration of Libra V1.

Y

## Test Plan

Fixed existing tests that rely on this functionality


language/transaction-builder/src/lib.rs
```diff
@@ -134,20 +134,15 @@ encode_txn_script! {
 pub fn encode_transfer_with_metadata_script(
     type_: TypeTag,
     recipient: &AccountAddress,
-    auth_key_prefix: Vec<u8>,
```
tzakian, line 137: Mega nit: now that we don't have the validation of the auth_key_prefix, this can be written using the `encode_txn_script!` macro.

sblackshear: Good call; fixed!

thefallentree:
ok. but PeerToPeerScript should be deleted, it is only ever used in test and same as using metadata script with an empty metadata

testsuite/cluster-test/src/tx_emitter.rs
```diff
@@ -698,6 +689,10 @@ async fn create_new_accounts(
             max_num_accounts_per_batch as usize,
             min(MAX_TXN_BATCH_SIZE, num_new_accounts - i),
         ));
+        // TODO: this no longer creates an account for the nonexisting receipient address.
```
sblackshear, line 692: @andll, could you advise me on the best approach here?

testsuite/cli/src/client_proxy.rs
```diff
@@ -756,7 +752,7 @@ impl ClientProxy {
```
christinacelee, line 752: can we locally test the transfer command on CLI to make sure it still works? https://github.com/libra/libra/blob/master/testsuite/cli/src/transfer_commands.rs#L25-L38 and update potentially outdated CLI outputs 😄

sblackshear: Checked it out and  it still works! For `transfer_commands.rs`, the min/max # of parameters hasn't changed and the `get_params_help` actually didn't mention `auth_key_prefix`, so I think we're good.

christinacelee: thanks for checking! lgtm from client side, just gotta fix the failing tests

andll:
Hmm, TPS went down quite a bit

christinacelee:
also how did this land with test failures :O

andll:
I don't think it landed. Cluster test runs in parallel with other test, but they both need to succeed before land

sblackshear:
> Hmm, TPS went down quite a bit

Does the TPS calculation include the account setup time? The only thing I changed in cluster tests is using the `create_child_vasp` script instead of `peer_to_peer_with_metadata` in  account setup + removing `auth_key_prefix` from the p2p script.

andll:
>Does the TPS calculation include the account setup time? 

It includes account setup time, yeah... We can improve this if we believe account setup now takes much longer

sblackshear:
> It includes account setup time, yeah... We can improve this if we believe account setup now takes much longer

It's not immediately obvious to me this would take longer (the create child VASP script actually does slightly less work than the old create-and-pay-script...), but we can look into it.

PR #3478, opened 2020-04-16
[types] Add NetworkAddress as multiaddr replacement

phlip9:
Libra `NetworkAddress` is a compact, efficient, self-describing and future-proof network address represented as a stack of protocols. Essentially libp2p's [multiaddr](https://multiformats.io/multiaddr/) but using `lcs` to describe the binary format.

One additional motivation is to eliminate our dependency on `parity-multiaddr`, which is apparently a weighty dependency and would have to be included in our very-locked-down safety rules module.

Related to this issue: https://github.com/libra/libra/issues/1158

Issue #3960, opened 2020-05-19
[network] re-thinking network peer id story


phlip9:
### Current State

LibraNet was originally written for the validator network use-case, while other use-cases such as the validator-internal network and Public FullNode network were engineered in afterwards. These historical artefacts are apparent in how we identify peers in LibraNet networks.

All connections across all use-cases are authenticated (either server-only or mutually) and encrypted with NoiseIK. The NoiseIK handshake proves ownership of a network identity keypair (specifically, an `(x25519::PrivateKey, x25519::PublicKey)` keypair).

In the original validator network use-case, connections are identified and dedup'ed by each validator peers' on-chain `AccountAddress`. In a connection handshake, the validator peer's `x25519::PublicKey` is reverse-mapped to the validator's `AccountAddress` via the on-chain `ValidatorSet`, which for our purposes we can abstract as a `HashMap<AccountAddress, Vec<x25519::PublicKey>>`. This fact is important because it allows us to maintain _at most one_ connection per validator and simplify application protocols, since the mental model is just one validator, one account address, one peer id, one connection.

For validator-internal network and public fullnode network use-cases, `PeerId = AccountAddress = Hash(x25519::PublicKey)`, which is awkward as `AccountAddress` is meaningless in these cases. In an ideal world, `PeerId` would just be a `x25519::PublicKey`. The intention of this issue is to explore how we can remove this weird conversion or otherwise rethink network id's as whole in LibraNet.

### Options

1. Do nothing. `PeerId = AccountAddress` everywhere.

2. Make `PeerId` generic, e.g., `NetworkEvents<TPeerId, TMessage>`.

For validator networks, `PeerId = AccountAddress` while in others `PeerId = x25519::PublicKey`. Validator transports would probably have a `.map(|connection| { .. })` that maps negotiated `x25519::PublicKey` to `AccountAddress`. Unfortunately, adding a generic `PeerId` everywhere in `network/` would cause significant API churn and complicate application protocols like state sync and mempool, which need to hold multiple distinct network handles (which would now be different types).

3. `PeerId = x25519::PublicKey` everywhere.

[...]

PR #4163, opened 2020-06-01
[network] monolithic transport

phlip9:
+ As discussed in a [previous PR](https://github.com/libra/libra/pull/3478#issuecomment-617385633)
    it's too much effort to totally refactor our combinators-based transport
    abstraction to properly support parsing `NetworkAddress`es
    incrementally. In addition, we need to better distinguish b/w inbound
    and outbound upgrades (i.e., non-symmetric handshake protocols).

+ Instead, we will just simplify all of our transports into a single
    `LibraNetTransport` in exchange for being a bit more brittle. This
    transport supports a pluggable "base" transport (Tcp or Memory) and then
    handles upgrading connections with Noise and protocol negotiation
    handshake. It also properly parses `NetworkAddress`es and returns a fully
    formatted `NetworkAddress` when listening.

+ Eventually, we'd like to move back to a combinator-based approach.
    However, we just don't have enough time in the short term to iterate on
    the right trait APIs.

+ We add some unit tests (finally!) for the different transport cases so
    we don't accidentally break things (for example, the previous no-noise
    testing transport was actually broken for a long time, since we had no
    code exercising it).

+ TcpTransport and MemoryTransport now go back to expecting no trailing
    protocols in the dialing address.

+ Remove unauthenticated, test-only auth mode and transports.
    We're not using this enough to warrent the extra complexity needed to
    support this use case. The only consumer is a state sync integration
    test which can easily be converted to just using authentication.

+ Connection id generator does not need a mutex.
    We can just use an `AtomicU32` to atomically increment a shared
    connection id counter instead of using a whole shared mutex.


phlip9:
rebase

config/src/config/network_config.rs
```diff
@@ -37,10 +37,13 @@ pub struct NetworkConfig {
...
     pub connectivity_check_interval_ms: u64,
     // Flag to toggle if Noise is used for encryption and authentication.
+    // TODO(philiphayes): remove this flag. no longer useful. all networks must use noise.
     pub enable_noise: bool,
     // If the network uses remote authentication, only trusted peers are allowed to connect.
     // Otherwise, any node can connect. If this flag is set to true, `enable_noise` must
     // also be set to true.
+    // TODO(philiphayes): rename this flag. should reflect `AuthenticationMode` in
+    // `NetworkBuilder`, i.e., `Mutual` vs `ServerOnly` authentication.
     pub enable_remote_authentication: bool,
```
davidiw, line 47: we have to be careful here, because ServerOnly is super ambiguous on who is the server and who is the client.

network/src/protocols/identity.rs
```diff
@@ -1,44 +1,14 @@
...
+//! Protocol used to exchange supported protocol information with a remote.
+
 use crate::protocols::wire::handshake::v1::HandshakeMsg;
 use bytes::BytesMut;
 use futures::io::{AsyncRead, AsyncWrite, AsyncWriteExt};
-use libra_types::PeerId;
 use netcore::framing::{read_u16frame, write_u16frame};
 use std::io;
 
-/// The PeerId exchange protocol.
```
davidiw, line 15: makes sense, the peer id exchange should happen in noise anyway since we authenticate against it.

network/src/transport.rs
```diff
@@ -281,25 +281,6 @@ pub fn build_unauthenticated_memory_noise_transport(
         .boxed()
 }
 
-pub fn build_memory_transport(
```
davidiw, line 284: do we want to remove this? it gives us insights into performance? though I guess it adds tech debt?

state-synchronizer/src/tests/integration_tests.rs
```diff
@@ -194,6 +201,7 @@ impl SynchronizerEnv {
             clients: vec![],
             storage_proxies: vec![],
             signers,
+            network_id_keys,
```
davidiw, line 204: hmm... this is a bit unfortunate... lemme know about the tech debt assumption -- otherwise having unauthed could be super useful for testing.

network/netcore/src/transport/tcp.rs
```diff
@@ -75,27 +75,27 @@ impl Transport for TcpTransport {
     ) -> Result<(Self::Listener, NetworkAddress), Self::Error> {
         let ((ipaddr, port), addr_suffix) =
             parse_ip_tcp(addr.as_slice()).ok_or_else(|| invalid_addr_error(&addr))?;
```
davidiw, line 77: move parse_ip_tcp into network_address.rs / NetworkAddress?

network/netcore/src/transport/tcp.rs
```diff
@@ -75,27 +75,27 @@ impl Transport for TcpTransport {
...
-            actual_addr,
+            listen_addr,
         ))
     }
 
     fn dial(&self, addr: NetworkAddress) -> Result<Self::Outbound, Self::Error> {
         let (addr_string, _addr_suffix) =
             parse_addr_and_port(addr.as_slice()).ok_or_else(|| invalid_addr_error(&addr))?;
+        // TODO(philiphayes): base tcp transport should not allow trailing protocols
         // TODO(philiphayes): use `tokio::net::lookup_host()` then filter the results
```
davidiw, line 99: nice

network/netcore/src/transport/tcp.rs
```diff
@@ -75,27 +75,27 @@ impl Transport for TcpTransport {
...
                 config: self.clone(),
             },
-            actual_addr,
+            listen_addr,
         ))
     }
 
     fn dial(&self, addr: NetworkAddress) -> Result<Self::Outbound, Self::Error> {
         let (addr_string, _addr_suffix) =
             parse_addr_and_port(addr.as_slice()).ok_or_else(|| invalid_addr_error(&addr))?;
```
davidiw, line 97: this should also come from NetworkAddress directly and in that way we can enforce your todo.... since it should ideally return a sock addr with a resolved dns...

network/src/noise_wrapper/handshake.rs
```diff
@@ -87,7 +87,7 @@ impl NoiseWrapper {
         &self,
         socket: TSocket,
         origin: ConnectionOrigin,
-        anti_replay_timestamps: Option<Arc<RwLock<AntiReplayTimestamps>>>,
+        anti_replay_timestamps: Option<&Arc<RwLock<AntiReplayTimestamps>>>,
```
davidiw, line 91: It seems like we should introduce this lock into the NoiseWrapper itself, no?

network/src/transport.rs
```diff
@@ -132,56 +143,56 @@ pub struct Connection<TSocket> {
...
-fn identity_key_to_peer_id(
-    trusted_peers: &RwLock<HashMap<PeerId, NetworkPublicKeys>>,
-    remote_static_key: &[u8],
-) -> Option<PeerId> {
-    for (peer_id, public_keys) in trusted_peers.read().unwrap().iter() {
-        if public_keys.identity_public_key.to_bytes() == remote_static_key {
-            return Some(*peer_id);
-        }
+/// Convert a remote pubkey into a network `PeerId`.
+fn identity_pubkey_to_peer_id(
```
davidiw, line 147: the remote peer should submit a peer id in noise, that should either match their provided public key or be indexed into trusted peers... this will mitigate id confusion which I fear is happening right now in some of our tests

I left a comment here: https://github.com/libra/libra/issues/3960#issuecomment-637250990

and spoke with @mimoo to nail this down...

network/src/transport.rs
```diff
@@ -132,56 +143,56 @@ pub struct Connection<TSocket> {
...
+    if let Some(trusted_peers) = maybe_trusted_peers {
+        // if mutual authentication, try to find peer with this identity pubkey.
+        find_peer_with_pubkey(&trusted_peers.read().unwrap(), remote_pubkey)
+            .ok_or_else(|| io::Error::new(io::ErrorKind::Other, "Not a trusted peer"))
+            .map_err(|err| {
+                security_log(SecurityEvent::InvalidNetworkPeer)
+                    .error("UntrustedPeer")
+                    .data(&trusted_peers)
+                    .data(&remote_pubkey)
+                    .log();
```
davidiw, line 160: security_log is unmaintained, if you want to log this, go for it, but just use the regular logger

@mimoo are we deleting it and just going to depend on the regular structured logger?

network/src/transport.rs
davidiw: for what it is worth this doesn't violate the proposal it just isn't as strict :)

network/src/transport.rs
```diff
@@ -132,56 +143,56 @@ pub struct Connection<TSocket> {
...
-
-    Ok(())
+fn find_peer_with_pubkey(
+    trusted_peers: &HashMap<PeerId, NetworkPublicKeys>,
+    remote_pubkey: &x25519::PublicKey,
+) -> Option<PeerId> {
+    trusted_peers
+        .iter()
+        .find(|(_peer_id, public_keys)| &public_keys.identity_public_key == remote_pubkey)
+        .map(|(peer_id, _)| *peer_id)
```
davidiw, line 185: if we're only using this once, let's put it inline?

network/src/transport.rs
```diff
@@ -207,156 +218,948 @@ pub async fn perform_handshake<T: TSocket>(
...
+            .or_else(|| parse_dns_tcp(protos).map(|x| (&protos[..2], x.1)))
+            .or_else(|| parse_memory(protos).map(|x| (&protos[..1], x.1)))
+            .ok_or_else(|| {
+                io::Error::new(
+                    io::ErrorKind::InvalidInput,
+                    format!(
+                        "Unexpected dialing network address: '{}', expected: \
+                         memory, ip+tcp, or dns+tcp",
+                        addr
+                    ),
```
davidiw, line 401: yeah, this feels weird, we should explicitly know which type we support by the builder we construct... otherwise we will potentially parse_memory in a ip network :O

network/src/transport.rs
```diff
@@ -207,156 +218,948 @@ pub async fn perform_handshake<T: TSocket>(
...
+            supported_protocols.clone(),
+        );
+
+        let dialer_transport = LibraNetTransport::new(
+            LIBRA_TCP_TRANSPORT.clone(),
+            dialer_key,
+            Some(trusted_peers),
+            HANDSHAKE_VERSION,
+            supported_protocols,
+        );
```
davidiw, line 1138: this is the same code over and over again and I suspect there's a lot more copy pasta

perhaps we could shove more of this into setup and pass in the transport or a transport new function pointer to create that type...

network/network-address/src/lib.rs
```diff
@@ -1007,8 +960,8 @@ mod test {
             ((&dns_name, 123), expected_suffix)
         );
 
-        let addr = NetworkAddress::from_str("/dns/example.com/tcp/123/ln-peerid-ex").unwrap();
-        let expected_suffix: &[Protocol] = &[Protocol::PeerIdExchange];
+        let addr = NetworkAddress::from_str("/dns/example.com/tcp/123/memory/44").unwrap();
```
davidiw, line 963: do we have any docs in this code that need to be updated?

network/src/transport.rs
```diff
@@ -60,20 +61,19 @@ impl From<u32> for ConnectionId {
 
 /// Generator of unique ConnectionIds.
 struct ConnectionIdGenerator {
-    next: ConnectionId,
+    ctr: AtomicU32,
 }
 
 impl ConnectionIdGenerator {
-    fn new() -> ConnectionIdGenerator {
+    const fn new() -> ConnectionIdGenerator {
```
mimoo, line 79: how can this be a const fn and why make it a const fn oO?

network/src/transport.rs
```diff
@@ -60,20 +61,19 @@ impl From<u32> for ConnectionId {
...
+            ctr: AtomicU32::new(0),
         }
     }
 
-    fn next(&mut self) -> ConnectionId {
-        let ret = self.next;
-        self.next = ConnectionId(ret.0.wrapping_add(1));
-        ret
+    fn next(&self) -> ConnectionId {
+        let next = self.ctr.fetch_add(1, Ordering::Relaxed);
```
mimoo, line 86: shouldn't this be SeqCst?

network/src/protocols/identity.rs
```diff
@@ -1,44 +1,14 @@
...
-        io::Error::new(
-            io::ErrorKind::InvalidData,
-            format!("Failed to parse identity msg: {}", e),
-        )
-    })?;
-    Ok(remote_peer_id)
-}
-
 /// The Handshake exchange protocol.
 pub async fn exchange_handshake<T>(
```
mimoo, line 13: unrelated to this PR, but IIUC this is only to negotiate supported protocol, which is just a part of the handshake. So I find it weird to call that a handshake :D (and we already have a handshake in noise)

network/src/validator_network/network_builder.rs
```diff
@@ -65,11 +65,6 @@ pub const MAX_CONNECTION_DELAY_MS: u64 = 10 * 60 * 1000 /* 10 minutes */;
 
 #[derive(Debug)]
 pub enum AuthenticationMode {
-    /// Inbound and outbound connections are unauthenticated and unencrypted, using
-    /// a simple PeerId exchange protocol to share each peer's PeerId in-band.
-    /// This mode is obviously insecure and should never be used in production.
-    // TODO(philiphayes): remove? gate with cfg(test)? should never use this in prod.
-    Unauthenticated,
```
mimoo, line 72: 🌶

network/src/validator_network/network_builder.rs
```diff
@@ -81,16 +76,16 @@ pub enum AuthenticationMode {
...
+impl AuthenticationMode {
+    /// Convenience method to retrieve the public key for the auth mode's inner
+    /// network identity key.
+    ///
+    /// Note: this only works because all auth modes are Noise-based.
+    pub fn public_key(&self) -> x25519::PublicKey {
+        match self {
+            AuthenticationMode::ServerOnly(key) | AuthenticationMode::Mutual(key) => {
+                key.public_key()
+            }
```
mimoo, line 87: this match feels hacky, but I guess there's no other way to write this in Rust?
Probably AuthenticationMode should be re-written as a struct: 

```rust
pub struct AuthenticationMode {
  mutual_authentication: bool,
  private_key: x25519::PrivateKey,
}
```

network/src/noise_wrapper/handshake.rs
```diff
@@ -181,7 +181,7 @@ impl NoiseWrapper {
     pub async fn accept<TSocket>(
         &self,
         mut socket: TSocket,
-        anti_replay_timestamps: Option<Arc<RwLock<AntiReplayTimestamps>>>,
+        anti_replay_timestamps: Option<&Arc<RwLock<AntiReplayTimestamps>>>,
```
mimoo, line 200: isn't it weird to borrow an Arc btw?

network/src/noise_wrapper/handshake.rs
```diff
@@ -222,7 +222,7 @@ impl NoiseWrapper {
         }
 
         // if on a mutually authenticated network
-        if let Some(anti_replay_timestamps) = &anti_replay_timestamps {
+        if let Some(anti_replay_timestamps) = anti_replay_timestamps {
```
mimoo, line 241: it feels cleaner to me to have anti_replay_timestamps.clone() instead of borrowing a pointer

network/src/noise_wrapper/handshake.rs
mimoo: once this is pushed I can probably simplify the handshake interface : o

network/src/noise_wrapper/handshake.rs
```diff
@@ -121,7 +122,14 @@ impl NoiseWrapper {
...
 
-    pub async fn dial<TSocket>(
+    /// Perform an outbound protocol upgrade on this connection.
+    ///
+    /// This runs the "client" side of the Noise IK handshake to establish a
+    /// secure Noise stream and exchange static public keys. In mutual auth
+    /// scenarios, we will also include an anti replay attack counter in the
+    /// Noise handshake payload. Currently this counter is always a millisecond-
+    /// granularity unix epoch timestamp.
+    pub async fn upgrade_outbound<TSocket>(
```
mimoo, line 132: (I find dial/accept clearer since this is what really happens)

network/src/noise_wrapper/fuzzing.rs
```diff
@@ -8,7 +8,7 @@
 // This fuzzes the wrappers we have around our Noise library.
 //
 
-use crate::noise_wrapper::NoiseWrapper;
+use crate::noise_wrapper::NoiseUpgrader;
```
mimoo, line 11: if you change the name of the struct, should you change the name of the module as well?

network/src/transport.rs
```diff
@@ -207,193 +218,948 @@ pub async fn perform_handshake<T: TSocket>(
...
+        let mut own_handshake = HandshakeMsg::new();
+        own_handshake.add(SUPPORTED_MESSAGING_PROTOCOL, application_protocols);
+        let identity_pubkey = identity_key.public_key();
+
+        // Only use anti replay protection in mutual-auth scenarios. In theory,
+        // this is applicable everywhere; however, we would need to spend some
+        // time making this more sophisticated so it garbage collects old
+        // timestamps and doesn't use unbounded space. These are not problems in
+        // mutual-auth scenarios because we have a bounded set of trusted peers
+        // that rarely changes.
```
mimoo, line 359: this is not the rational behind not using timestamps in one-way authenticated connections. We don't do it because there is no point for an attacker to replay messages, as they can just create new connections with new public keys (no trusted peers from the server side).

network/src/transport.rs
```diff
@@ -132,56 +143,56 @@ pub struct Connection<TSocket> {
...
-    trusted_peers: &RwLock<HashMap<PeerId, NetworkPublicKeys>>,
-    remote_static_key: &[u8],
-) -> Option<PeerId> {
-    for (peer_id, public_keys) in trusted_peers.read().unwrap().iter() {
-        if public_keys.identity_public_key.to_bytes() == remote_static_key {
-            return Some(*peer_id);
-        }
+/// Convert a remote pubkey into a network `PeerId`.
+fn identity_pubkey_to_peer_id(
+    maybe_trusted_peers: Option<&Arc<RwLock<HashMap<PeerId, NetworkPublicKeys>>>>,
```
mimoo, line 148: why call it `maybe_*` when the `Option` already indicates this?

config/src/config/network_config.rs
phlip9: Yeah, i'm still not sure what's a good name for not-mutual auth. In my mind, it was always "client-only" auth b/c the dialing peer is the only one meaningfully authenticating the server. But you could also read "client-only" as the client is the only one who is authenticated, so I understand the confusion.

Any suggestions for better names?

network/src/transport.rs
```diff
@@ -132,56 +143,56 @@ pub struct Connection<TSocket> {
...
+    remote_pubkey: &x25519::PublicKey,
+) -> io::Result<PeerId> {
+    if let Some(trusted_peers) = maybe_trusted_peers {
+        // if mutual authentication, try to find peer with this identity pubkey.
+        find_peer_with_pubkey(&trusted_peers.read().unwrap(), remote_pubkey)
```
mimoo, line 153: no need to unwrap since we're returning a result

state-synchronizer/src/tests/integration_tests.rs
phlip9: I would actually prefer everyone to use as-close-to-prod-as-possible transports, since they will then exercise more realistic code paths.

network/src/transport.rs
mimoo: this is a good question. Do you have context on why there's a separate security_log? I'll start to look at security logs, intuition was that I could just use the current logging framework that we have with a specific flag to mark it for security

network/src/transport.rs
mimoo: > the remote peer should submit a peer id in noise

you mean send it as part of the handshake?

network/src/noise_wrapper/handshake.rs
davidiw: How would you simplify it?

network/src/transport.rs
davidiw: @gedigi said that he introduced it at a time where there was no logging and we needed logging before launching test net. I think at this point in time, we should focus on having a single logging framework and there should be a component outside that grabs logs that it wants rather than having specific logging infra inside.

davidiw: tl;dr let's delete it

network/src/transport.rs
davidiw: we always unwrap locks, since we have been poisoned otherwise and need to crash the service, there is no recovering from that :/

network/src/transport.rs
```diff
@@ -207,193 +218,948 @@ pub async fn perform_handshake<T: TSocket>(
...
-            }
-        })
-        .and_then(move |(peer_id, socket), addr, origin| async move {
-            perform_handshake(peer_id, socket, addr, origin, &own_handshake).await
-        })
-        .with_timeout(TRANSPORT_TIMEOUT)
-        .boxed()
+/// Check that the parsed `handshake_version` from the `NetworkAddress` is supported.
+/// Return `Err` otherwise.
+fn verify_handshake_version_supported(expected_version: u8, version: u8) -> io::Result<()> {
```
mimoo, line 223: it seems weird to have this as its own function when it's just an equality check. Let's inline it :)

network/src/noise_wrapper/handshake.rs
mimoo: there's a lot of stuff that's unnecessary just because I didn't want to copy/paste it accross all the transport functions. For one we don't need an `upgrade_connection()` function and we should be able to directly call `dial()` and `accept()`

network/src/transport.rs
mimoo: so let's not delete it, and I'll replace these in another PR :)

config/src/config/network_config.rs
mimoo: I think you guys are talking about different things, @davidiw is pointing out that the same code gets called by both clients and servers (but this PR improves things!)

The server-side authentication is correct, but indeed doesn't really describe the situation. Perhaps:

* KnownClient
* UnknownClient

mimoo: in the case of bools, we could have even simpler names

* client_in_trusted_peers
* verify_client

network/src/transport.rs
phlip9: socket bench is just using a `MemoryTransport` directly, so we don't really need these anymore

network/netcore/src/transport/tcp.rs
phlip9: it already is : p

network/netcore/src/transport/tcp.rs
phlip9: I don't think `parse_addr_and_port` belongs in `NetworkAddress`. it's just a temporary, minimum effort way to get dialing working. doing the proper filtering requires a bit more work.

network/src/noise_wrapper/handshake.rs
phlip9: this is what i was thinking about, but left out b/c PR is already huge

```rust
struct NoiseUpgrader {
    config: NoiseConfig,
    anti_replay_timestamps: Option<Arc<RwLock<AntiReplayTimestamps>>>,
    trusted_peers: Option<Arc<RwLock<HashMap<PeerId, NetworkPublicKeys>>>>,
}
```

network/src/transport.rs
mimoo: I had the same dilemma on the noise side, and decided that at this point you can still continue to use the current connection you have and log the fact that you're having issues creating new connections. This way we can continue running while we fix the issue.

network/src/transport.rs
phlip9: yeah, i'm not sure what/who uses security_log. I just preserved the current behaviour... should I remove this?

network/src/transport.rs
phlip9: this is hacky because the `base_transport` doesn't have a `parse_dial_addr` or what not. ideally, you would let `base_transport` like parse out its bit, then parse out your piece, but `Transport` would need to change to support that. I would also have to figure out how the combinators would change to support this, which is explicitly what I wanted to avoid.

network/src/transport.rs
mimoo: again, do not remove it

config/src/config/network_config.rs
davidiw: EnableRemotePeerAuthentication
if you want a boolean
RemotePeerAuthenticated
RemotePeerUnauthenticated
if you want a an enum
and if we want to have all three in one:
MutualAuthenticated
...

network/src/transport.rs
phlip9: managed to clean this up a lot. -300 lines : )

network/network-address/src/lib.rs
phlip9: think i got everything!

network/src/transport.rs
phlip9: this let's you do

```rust
static CONNECTION_ID_GENERATOR: ConnectionIdGenerator = ConnectionIdGenerator::new();
```

without lazy static. it can be a `const fn` b/c `AtomicU32::new` is a `const fn`

network/src/transport.rs
phlip9: we don't care what the ordering is, just that the increment is atomic and therefore returns a unique id (no torn reads/writes etc)

network/src/protocols/identity.rs
phlip9: agreed, handshake is not descriptive enough. it's a protocol+version negotiation handshake.

network/src/noise_wrapper/handshake.rs
phlip9: yeah `&Arc` is usually an anti-pattern imo. i'll add a commit to refactor these into fields

phlip9: done!

network/src/transport.rs
phlip9: If we condense our whole handshake (noise + protocol negotiation) into 1 RTT (which we can actually do today, but we're not rn because it breaks abstractions a bit), replays become much more dangerous, since a successful replay would replace the honest peer's current connection. In this world, anti replay mitigation would help everywhere (unless clients in one-way auth are literally sampling a new static keypair for every connection).

network/src/transport.rs
phlip9: it does, but I think it's confusing when in some places `trusted_peers: Arc<..>` and in others `trusted_peers: Option<Arc<..>>`

network/netcore/src/transport/tcp.rs
davidiw: Oooh, that's interesting... checking apis...
oh two questions about that
1) Why aren't these implemented directly on NetworkAddress
2) Why not import the module and call the functions that way, at least old farts like me wouldn't make such bold and inaccurate statements :)

network/netcore/src/transport/tcp.rs
davidiw: why's that? what's the plan?

network/src/noise_wrapper/handshake.rs
```diff
@@ -60,8 +60,59 @@ impl AntiReplayTimestamps {
...
+    /// actually contained in the current validator set.
+    Mutual {
+        // Only use anti replay protection in mutual-auth scenarios. In theory,
+        // this is applicable everywhere; however, we would need to spend some
+        // time making this more sophisticated so it garbage collects old
+        // timestamps and doesn't use unbounded space. These are not problems in
+        // mutual-auth scenarios because we have a bounded set of trusted peers
+        // that rarely changes.
+        anti_replay_timestamps: RwLock<AntiReplayTimestamps>,
+        trusted_peers: Arc<RwLock<HashMap<PeerId, NetworkPeerInfo>>>,
```
davidiw, line 80: nice

network/src/noise_wrapper/handshake.rs
```diff
@@ -60,8 +60,59 @@ impl AntiReplayTimestamps {
...
+                ..
+            } => Some(&anti_replay_timestamps),
+            HandshakeAuthMode::ServerOnly => None,
+        }
+    }
+
+    fn trusted_peers(&self) -> Option<&RwLock<HashMap<PeerId, NetworkPeerInfo>>> {
+        match &self {
+            HandshakeAuthMode::Mutual { trusted_peers, .. } => Some(&trusted_peers),
+            HandshakeAuthMode::ServerOnly => None,
```
davidiw, line 109: This seems weird since we have a case where the ServerOnly, does the client enable Mutual and the Server enable this? And what if neither authenticates? Both are ServerOnly? We should nail down consistent naming here :)

network/src/noise_wrapper/handshake.rs
```diff
@@ -307,40 +373,67 @@ mod test {
...
+                vec![(client_id, client_keys), (server_id, server_keys)]
+                    .into_iter()
+                    .collect(),
+            ));
+            let client_auth = HandshakeAuthMode::mutual(trusted_peers.clone());
+            let server_auth = HandshakeAuthMode::mutual(trusted_peers);
+            (client_auth, server_auth)
+        } else {
+            (HandshakeAuthMode::ServerOnly, HandshakeAuthMode::ServerOnly)
+        };
```
davidiw, line 395: This is kind of my concern :/

ServerOnly, as you call it, most certainly has a set of trusted peers if it is the client.

network/src/protocols/identity.rs
davidiw: I think we also need to figure out what this really means... might be worth having an adhoc sync next week...
but what does it mean to support a protocol? What if I can send queries on that protocol (i.e., a client) but have no service? Do I express this in the handshake or is the handshake only for server protocols?

My gut feeling is that we have a lot of flexibility here and this is more of how we internally handle this versus what the wire spec enforces.

network/src/validator_network/network_builder.rs
```diff
@@ -401,58 +396,41 @@ impl NetworkBuilder {
...
-
-        // would like to do this in `build_with_transport` but ownership is very
-        // tricky here since the private keys aren't cloneable and we don't have
-        // a safe key container yet...
-        let unbound_listen_addr = self.listen_address.clone();
-        let unbound_len = unbound_listen_addr.len();
-        let unbound_listen_addr =
-            append_libranet_protocols(unbound_listen_addr, &authentication_mode);
+        let (key, maybe_trusted_peers) = match authentication_mode {
+            AuthenticationMode::ServerOnly(key) => (key, None),
```
davidiw, line 407: I'm confused, we are showing server only for the server side

probably starting to sound repetitive here, let's chat tomorrow :)

network/src/transport.rs
davidiw: actually as an API name I do find it kinda repetitive but I could see it being called maybe within a function body but your call

network/src/transport.rs
```diff
@@ -132,56 +143,52 @@ pub struct Connection<TSocket> {
...
+/// Convert a remote pubkey into a network `PeerId`.
+fn identity_pubkey_to_peer_id(
+    maybe_trusted_peers: Option<&Arc<RwLock<HashMap<PeerId, NetworkPublicKeys>>>>,
+    remote_pubkey: &x25519::PublicKey,
+) -> io::Result<PeerId> {
+    if let Some(trusted_peers) = maybe_trusted_peers {
+        // if mutual authentication, try to find peer with this identity pubkey.
+
+        let trusted_peers = trusted_peers.read().unwrap();
+        let maybe_peerid_with_pubkey = trusted_peers
```
davidiw, line 155: this seems like a good maybe candidate

network/src/transport.rs
davidiw: I guess I should have said delete and leave a TODO, my concern with just leaving it is that we don't really use it, so keeping it around just adds noise whereas a TODO is much clearer...

@gedigi @gedigi @gedigi 
I thought we had to say your name 3 times?

network/src/transport.rs
```diff
@@ -207,193 +214,667 @@ pub async fn perform_handshake<T: TSocket>(
...
+
+    #[test]
+    fn test_memory_transport_mutual_auth() {
+        test_transport_success(
+            memory::MemoryTransport,
+            Auth::Mutual,
+            "/memory/0",
+            expect_memory_noise_addr,
+        );
+    }
```
davidiw, line 827: this is super easy to understand, thank you!

network/src/transport.rs
```diff
@@ -207,193 +214,667 @@ pub async fn perform_handshake<T: TSocket>(
...
+
+        let (mut inbounds, listener_addr) = rt.enter(|| {
+            listener_transport
+                .listen_on(listen_addr.parse().unwrap())
+                .unwrap()
+        });
+        expect_formatted_addr(&listener_addr);
+        let supported_protocols_clone = supported_protocols.clone();
+
+        let listener_task = async move {
```
davidiw, line 702: could we put the comment at the top saying exactly what's happening in here, it is a bit long so it helps me re-focus as I read the code, I suspect other people might benefit

very long functions are difficult to parse

network/src/transport.rs
```diff
@@ -207,193 +214,667 @@ pub async fn perform_handshake<T: TSocket>(
...
+
+            // test the socket works
+
+            let msg = write_read_msg(&mut conn.socket, b"foobar").await;
+            assert_eq!(&msg, b"barbaz".as_ref());
+
+            conn.socket.close().await.unwrap();
+        };
+
+        let dialer_task = async move {
```
davidiw, line 730: same here

network/src/transport.rs
phlip9: A poisoned lock means something else panicked while holding the lock, which means we're already in a presumably unrecoverable state (?). I'm not sure there's much value to handling this defensively while we're already crashing. If anything, instead of trying to continue in this sort of weird state, we should probably just crash the whole process and let the monitoring agent or w/e restart it...

of course, erlang has some opinions on this, about crashing actors and recovering them. there's actually an erlang inspired framework/runtime for rust https://docs.rs/bastion/0.3.4/bastion/ which might have some interesting ideas to look into.

network/src/noise_wrapper/handshake.rs
phlip9: why would the client dial someone they don't want to talk to? Noise IK already provides this authentication for the client w/o needing a specific trusted_peers set.

network/netcore/src/transport/tcp.rs
phlip9: because the end state would probably look like

```rust
parse_ip_tcp => tokio::net::ToSocketAddrs => .await  \
                                                      => maybe filter ipv4/ipv6 => try SocketAddrs until one works
parse_dns_tcp => tokio::net::ToSocketAddrs => .await /
```

like this, but filtering ipv4/ipv6
https://docs.rs/tokio/0.2.21/src/tokio/net/tcp/stream.rs.html#97-115

hence parse_addr_and_port is temporary and i want to remove it soon.

phlip9:
rebase

mimoo:
awesome, make sure to address all the important comments :D

@bors-libra delegate+

network/src/transport.rs
mimoo: interesting!

network/src/transport.rs
mimoo: I don't have a lot of understanding for this, but isn't the danger that you start a connection with connection id = X when you wanted X+1 because of some compiler/machine re-ordering?

network/src/noise_wrapper/handshake.rs
mimoo: yeah makes sense actually

network/src/transport.rs
mimoo: ah that's a good point... another way to obtain "key confirmation" here is to wait for a message from the initiator post-handshake. It's a bit ugly but we could in theory add that to the handshake.

This message could be the encryption of an empty message, or even better any negotiation stuff that can happen post-handshake.

network/src/transport.rs
mimoo: agree here ^ it would be useful as a variable name, but I don't think it's idiomatic to have it as an argument name in the function signature

network/src/transport.rs
mimoo: yeah I don't think we should assume that we're in an unrecoverable state. Maybe the debug interface panicked, and why would we care about that, etc.

I really like erlang's approach too, but it's a bit too late to transform our codebase :D we should do that for another implementation though :p

network/src/transport.rs
phlip9: again, we just care that each call to `next` returns a unique id. the ordering relative to other atomics/instructions doesn't matter.

network/src/transport.rs
phlip9: if we continue as we are today (i.e., second RTT w/ protocol negotiation) then replays are not nearly as dangerous and we can afford to skip the engineering effort required to make the protection work in public-facing endpoints. But that's obviously a pretty implicit property that we should definitely document : )

PR #4175, opened 2020-06-01
remove AC and gRPC

bmwill:
This PR completely removes the deprecated AC endpoint as well as all remaining uses of gRPC in the codebase.

DO NOT LAND till after 6/14


thefallentree:
woohoo

christinacelee:
thanks for this! 😄 could we just put this in draft PR mode so no one accidentally lands this before 6/14?

thefallentree:
and we picked 6/14 because ?

bmwill:
@thefallentree 6/14 was the date that @christinacelee gave me IIRC. I would prefer to land this earlier but I think the idea was to wait until the next pre-mainnet/testnet deployment. If everyone can agree on doing it before then, all the better IMO.

christinacelee:
it's based on the blog post announcement about JSON RPC on 5/14 - we're giving 4 weeks as a transition period for developers before we fully remove AC/gRPC 😄 
https://libra.org/en-US/blog/new-json-rpc/

bmwill:
That makes sense :)

This means that we should be able to land this PR after testnet release next week (which is slightly before 6/14) since it won't have a public affect till after a new testnet release is made post 6/14.

thefallentree:
6/14 is an sunday,  and we release on Wednesday, so if we want to have it disappear in 6/17 release then we can safely land this in anytime after 6/10

bmwill:
Got the go-ahead from @sausagee that this should be safe to land today since the release this week has already been "cut".

@christinacelee mind taking a second to do a review?

bmwill:
@bors-libra r=davidiw

The end of an era...

PR #4176, opened 2020-06-01
[cleanup][types] Remove 6 unused, untested functions from account_config

huitseeker:
This PR removes 6 untested, unused functions from `types::account_config`. The value of the PR may reside more in the signaling of the nature of those untested, unused methods rather than actual removal.

(I also have found 9 such unused, untested functions that might be removed in `language/`, 
https://gist.github.com/2f8ddffbbc18df783f1bad0b22375613
which I won't submit for a PR, as requested  [[1]](https://github.com/libra/libra/pull/3879#issuecomment-628976319) [[2]](https://github.com/libra/libra/pull/3879#issuecomment-629417720) )

### DIFF
--- a/types/src/account_config/constants/account.rs
+++ b/types/src/account_config/constants/account.rs
@@ -4,7 +4,7 @@
 use crate::account_config::constants::CORE_CODE_ADDRESS;
 use move_core_types::{
     identifier::{IdentStr, Identifier},
-    language_storage::{ModuleId, StructTag},
+    language_storage::ModuleId,
 };
 use once_cell::sync::Lazy;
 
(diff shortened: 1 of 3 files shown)


davidiw:
Adding @runtian-zhou, as I think he might have context. I'm planning on landing this PR end of day tomorrow (Tuesday).

runtian-zhou:
LGTM! These are no longer in use due to the introduction of `MoveResource` trait.

types/src/account_config/constants/account.rs
```diff
@@ -27,41 +24,10 @@ static SENT_EVENT_NAME: Lazy<Identifier> =
...
-pub fn account_struct_tag() -> StructTag {
-    StructTag {
-        address: CORE_CODE_ADDRESS,
-        module: ACCOUNT_MODULE_IDENTIFIER.clone(),
-        name: ACCOUNT_STRUCT_NAME.to_owned(),
-        type_params: vec![],
-    }
-}
-
-pub fn sent_payment_tag() -> StructTag {
```
thefallentree, line 51: wait, isn't this used in json-rpc crate?

huitseeker: Nope!

PR #4177, opened 2020-06-01
[safety-rules] Handling key verification during epoch changes

zaxguo:
## Motivation

This PR addresses the consensus key verification during epoch changes. Specifically, it verifiers the validator consensus key whenever there is a new epoch, started by start_new_epoch(). It ensures that the key is in the keys of validator set, otherwise it retrieves the appropriate one from the storage. 

Yes

## Test Plan

The test considers the following cases:
1) when the validator's account address is not in the validator set,
2) when the account address is in but the key does not match.
  1. the retrieved key matches ==> the right path
  2. the retrieved key does not match (TODO: address) 

## Related PRs

### DIFF
--- a/consensus/safety-rules/src/error.rs
+++ b/consensus/safety-rules/src/error.rs
@@ -32,8 +32,8 @@ pub enum Error {
     #[error("Invalid QC: {}", {0})]
     InvalidQuorumCertificate(String),
 
-    #[error("validator_verifier is not set, SafetyRules is not initialized")]
-    NotInitialized,
+    #[error("{0} is not set, SafetyRules is not initialized")]
+    NotInitialized(String),
 
     /// This proposal's round is less than round of preferred block.
     /// Returns the id of the preferred block.
--- a/consensus/safety-rules/src/safety_rules.rs
+++ b/consensus/safety-rules/src/safety_rules.rs
@@ -12,8 +12,9 @@ use consensus_types::{
 use libra_crypto::{ed25519::Ed25519Signature, hash::HashValue};
 use libra_logger::debug;
 use libra_types::{
-    block_info::BlockInfo, epoch_change::EpochChangeProof, ledger_info::LedgerInfo,
-    validator_signer::ValidatorSigner, validator_verifier::ValidatorVerifier, waypoint::Waypoint,
+    block_info::BlockInfo, epoch_change::EpochChangeProof, epoch_state::EpochState,
+    ledger_info::LedgerInfo, validator_signer::ValidatorSigner,
+    validator_verifier::ValidatorVerifier, waypoint::Waypoint,
 };
 
 /// SafetyRules is responsible for the safety of the consensus:
@@ -27,7 +28,7 @@ use libra_types::{
 /// set)
 pub struct SafetyRules {
     persistent_storage: PersistentSafetyStorage,
-    validator_signer: ValidatorSigner,
+    validator_signer: Option<ValidatorSigner>,
     validator_verifier: Option<ValidatorVerifier>,
 }
 
@@ -39,12 +40,22 @@ impl SafetyRules {
         let consensus_key = persistent_storage
             .consensus_key()
             .expect("Unable to retrieve consensus private key");
-        let validator_signer = ValidatorSigner::new(author, consensus_key);
+        let validator_signer = Some(ValidatorSigner::new(author, consensus_key));
         Self {
             persistent_storage,
             validator_signer,
             validator_verifier: None,
         }
+    }
+
+    fn signer(&self) -> Result<&ValidatorSigner, Error> {
+        self.validator_signer.as_ref().ok_or(Error::NotInitialized)
+    }
+
+    fn verifier(&self) -> Result<&ValidatorVerifier, Error> {
+        self.validator_verifier
+            .as_ref()
+            .ok_or(Error::NotInitialized)
     }
 
     /// Produces a LedgerInfo that either commits a block based upon the 3-chain commit rule
@@ -72,10 +83,7 @@ impl SafetyRules {
     /// This verifies a QC makes sense in the current context, specifically that this is for the
     /// current epoch and extends from the preffered round.
     fn verify_qc(&self, qc: &QuorumCert) -> Result<(), Error> {
-        let validator_verifier = self
-            .validator_verifier
-            .as_ref()
-            .ok_or(Error::NotInitialized)?;
+        let validator_verifier = self.verifier()?;
 
         qc.verify(validator_verifier)
             .map_err(|e| Error::InvalidQuorumCertificate(e.to_string()))?;
@@ -89,17 +97,52 @@ impl SafetyRules {
         }
     }
 
+    /// This reconciles the key pair of a validator signer with a given validator set
+    /// during epoch changes.
+    /// @TODO Given we cannot panic, we must handle the following two error cases:
+    ///     1. Validator not in the set
+    ///     2. Validator in the set, but no matching key found in storage
+    fn reconcile_key(&mut self, epoch_state: &EpochState) -> Result<(), Error> {
+        let signer = self.signer()?;
+        if let Some(expected_key) = epoch_state.verifier.get_public_key(&signer.author()) {
+            let curr_key = signer.public_key();
+            if curr_key != expected_key {
+                let consensus_key = self
+                    .persistent_storage
+                    .consensus_key_for_version(expected_key.clone())
+                    .expect("Unable to retrieve consensus private key");
+                debug!(
+                    "Reconcile pub key for signer {} [{} -> {}]",
+                    signer.author(),
+                    curr_key,
+                    expected_key
+                );
+                self.validator_signer = Some(ValidatorSigner::new(signer.author(), consensus_key));
+            } else {
+                debug!("Validator key matches the key in validator set.");
+            }
+        } else {
+            debug!("The validator is not in the validator set!");
+        }
+        Ok(())
+    }
+
     /// This sets the current validator verifier and updates the epoch and round information
     /// if this is a new epoch ending ledger info. It also sets the current waypoint to this
     /// LedgerInfo.
     /// @TODO if public key does not match private key in validator set, access persistent storage
     /// to identify new key
     fn start_new_epoch(&mut self, ledger_info: &LedgerInfo) -> Result<(), Error> {
+        debug!("Starting new epoch.");
         let epoch_state = ledger_info
             .next_epoch_state()
             .cloned()
             .ok_or(Error::InvalidLedgerInfo)?;
+
+        self.reconcile_key(&epoch_state)?;
+
         self.validator_verifier = Some(epoch_state.verifier);
+
         let current_epoch = self.persistent_storage.epoch()?;
 
         if current_epoch < epoch_state.epoch {
@@ -130,7 +173,7 @@ impl SafetyRules {
 
     /// This checkes whether the author of one proposal is the validator signer
     fn verify_author(&self, author: Option<Author>) -> Result<(), Error> {
-        let validator_signer_author = &self.validator_signer.author();
+        let validator_signer_author = &self.signer()?.author();
         let author = author
             .ok_or_else(|| Error::InvalidProposal("No author found in the proposal".into()))?;
         if validator_signer_author != &author {
@@ -219,6 +262,7 @@ impl TSafetyRules for SafetyRules {
         self.persistent_storage
             .set_last_voted_round(proposed_block.round())?;
 
+        let validator_signer = self.signer()?;
         Ok(Vote::new(
             VoteData::new(
                 proposed_block.gen_block_info(
@@ -228,9 +272,9 @@ impl TSafetyRules for SafetyRules {
                 ),
                 proposed_block.quorum_cert().certified_block().clone(),
             ),
-            self.validator_signer.author(),
+            validator_signer.author(),
             self.construct_ledger_info(proposed_block),
-            &self.validator_signer,
+            validator_signer,
         ))
     }
 
@@ -265,10 +309,11 @@ impl TSafetyRules for SafetyRules {
             ));
         }
 
+        let validator_signer = self.signer()?;
         COUNTERS.sign_proposal.inc();
         Ok(Block::new_proposal_from_block_data(
             block_data,
-            &self.validator_signer,
+            validator_signer,
         ))
     }
 
@@ -303,7 +348,8 @@ impl TSafetyRules for SafetyRules {
                 .set_last_voted_round(timeout.round())?;
         }
 
-        let signature = timeout.sign(&self.validator_signer);
+        let validator_signer = self.signer()?;
+        let signature = timeout.sign(validator_signer);
         COUNTERS.sign_timeout.inc();
         debug!("Successfully signed timeout message.");
         Ok(signature)
--- a/consensus/safety-rules/src/safety_rules.rs
+++ b/consensus/safety-rules/src/safety_rules.rs
@@ -30,23 +30,19 @@ pub struct SafetyRules {
     persistent_storage: PersistentSafetyStorage,
     validator_signer: Option<ValidatorSigner>,
     validator_verifier: Option<ValidatorVerifier>,
+    author: Author,
 }
 
 impl SafetyRules {
     /// Constructs a new instance of SafetyRules with the given persistent storage and the
     /// consensus private keys
     /// @TODO replace this with an API that takes in a SafetyRulesConfig
-    /// @TODO change the constructor to fit Option<validator_signer> and address the circular
-    /// dependency with reconcile_key() during initializationg
     pub fn new(author: Author, persistent_storage: PersistentSafetyStorage) -> Self {
-        let consensus_key = persistent_storage
-            .consensus_key()
-            .expect("Unable to retrieve consensus private key");
-        let validator_signer = Some(ValidatorSigner::new(author, consensus_key));
         Self {
             persistent_storage,
-            validator_signer,
+            validator_signer: None,
             validator_verifier: None,
+            author,
         }
     }
 
@@ -103,30 +99,28 @@ impl SafetyRules {
 
     /// This reconciles the key pair of a validator signer with a given validator set
     /// during epoch changes.
-    /// @TODO Given we cannot panic, we must handle the following two error cases:
+    /// Current impl of consensus does not expect key reconciliation to panic or fail.
+    /// Instead, we signal debug messages indicating following error causes:
     ///     1. Validator not in the set
     ///     2. Validator in the set, but no matching key found in storage
     fn reconcile_key(&mut self, epoch_state: &EpochState) -> Result<(), Error> {
-        let signer = self.signer()?;
-        if let Some(expected_key) = epoch_state.verifier.get_public_key(&signer.author()) {
-            let curr_key = signer.public_key();
-            if curr_key != expected_key {
+        if let Some(expected_key) = epoch_state.verifier.get_public_key(&self.author) {
+            let curr_key = self.signer().ok().map(|s| s.public_key());
+            if curr_key != Some(expected_key.clone()) {
+                debug!(
+                    "Reconciled pub key for signer {} [{:#?} -> {:#?}]",
+                    self.author, curr_key, expected_key
+                );
                 let consensus_key = self
                     .persistent_storage
-                    .consensus_key_for_version(expected_key.clone())
+                    .consensus_key_for_version(expected_key)
                     .expect("Unable to retrieve consensus private key");
-                debug!(
-                    "Reconciled pub key for signer {} [{} -> {}]",
-                    signer.author(),
-                    curr_key,
-                    expected_key
-                );
-                self.validator_signer = Some(ValidatorSigner::new(signer.author(), consensus_key));
+                self.validator_signer = Some(ValidatorSigner::new(self.author, consensus_key));
             } else {
                 debug!("Validator key matches the key in validator set.");
             }
         } else {
-            debug!("The validator is not in the validator set!");
+            debug!("The validator key is not in store!");
         }
         Ok(())
     }
--- a/consensus/safety-rules/src/safety_rules.rs
+++ b/consensus/safety-rules/src/safety_rules.rs
@@ -36,6 +36,8 @@ impl SafetyRules {
     /// Constructs a new instance of SafetyRules with the given persistent storage and the
     /// consensus private keys
     /// @TODO replace this with an API that takes in a SafetyRulesConfig
+    /// @TODO change the constructor to fit Option<validator_signer> and address the circular
+    /// dependency with reconcile_key() during initializationg
     pub fn new(author: Author, persistent_storage: PersistentSafetyStorage) -> Self {
         let consensus_key = persistent_storage
             .consensus_key()
@@ -49,13 +51,15 @@ impl SafetyRules {
     }
 
     fn signer(&self) -> Result<&ValidatorSigner, Error> {
-        self.validator_signer.as_ref().ok_or(Error::NotInitialized)
+        self.validator_signer
+            .as_ref()
+            .ok_or_else(|| Error::NotInitialized("validator_signer".into()))
     }
 
     fn verifier(&self) -> Result<&ValidatorVerifier, Error> {
         self.validator_verifier
             .as_ref()
-            .ok_or(Error::NotInitialized)
+            .ok_or_else(|| Error::NotInitialized("validator_verifier".into()))
     }
 
     /// Produces a LedgerInfo that either commits a block based upon the 3-chain commit rule
@@ -112,7 +116,7 @@ impl SafetyRules {
                     .consensus_key_for_version(expected_key.clone())
                     .expect("Unable to retrieve consensus private key");
                 debug!(
-                    "Reconcile pub key for signer {} [{} -> {}]",
+                    "Reconciled pub key for signer {} [{} -> {}]",
                     signer.author(),
                     curr_key,
                     expected_key
@@ -130,17 +134,13 @@ impl SafetyRules {
     /// This sets the current validator verifier and updates the epoch and round information
     /// if this is a new epoch ending ledger info. It also sets the current waypoint to this
     /// LedgerInfo.
-    /// @TODO if public key does not match private key in validator set, access persistent storage
-    /// to identify new key
     fn start_new_epoch(&mut self, ledger_info: &LedgerInfo) -> Result<(), Error> {
         debug!("Starting new epoch.");
         let epoch_state = ledger_info
             .next_epoch_state()
             .cloned()
             .ok_or(Error::InvalidLedgerInfo)?;
-
         self.reconcile_key(&epoch_state)?;
-
         self.validator_verifier = Some(epoch_state.verifier);
 
         let current_epoch = self.persistent_storage.epoch()?;
--- a/consensus/safety-rules/src/safety_rules.rs
+++ b/consensus/safety-rules/src/safety_rules.rs
@@ -223,10 +223,9 @@ impl TSafetyRules for SafetyRules {
     /// @TODO verify QC correctness
     fn construct_and_sign_vote(&mut self, vote_proposal: &VoteProposal) -> Result<Vote, Error> {
         debug!("Incoming vote proposal to sign.");
+        self.signer()?;
         let proposed_block = vote_proposal.block();
-
         self.verify_epoch(proposed_block.epoch())?;
-
         let last_voted_round = self.persistent_storage.last_voted_round()?;
         if proposed_block.round() <= last_voted_round {
             debug!(
@@ -264,7 +263,6 @@ impl TSafetyRules for SafetyRules {
 
         self.persistent_storage
             .set_last_voted_round(proposed_block.round())?;
-
         let validator_signer = self.signer()?;
         Ok(Vote::new(
             VoteData::new(
@@ -277,14 +275,16 @@ impl TSafetyRules for SafetyRules {
             ),
             validator_signer.author(),
             self.construct_ledger_info(proposed_block),
-            validator_signer,
+            &validator_signer,
         ))
     }
 
     fn sign_proposal(&mut self, block_data: BlockData) -> Result<Block, Error> {
         debug!("Incoming proposal to sign.");
+        let validator_signer = self.signer()?;
         self.verify_author(block_data.author())?;
         self.verify_epoch(block_data.epoch())?;
+
         let last_voted_round = self.persistent_storage.last_voted_round()?;
         if block_data.round() <= last_voted_round {
             debug!(
@@ -312,11 +312,10 @@ impl TSafetyRules for SafetyRules {
             ));
         }
 
-        let validator_signer = self.signer()?;
         COUNTERS.sign_proposal.inc();
         Ok(Block::new_proposal_from_block_data(
             block_data,
-            validator_signer,
+            &validator_signer,
         ))
     }
 
@@ -329,6 +328,7 @@ impl TSafetyRules for SafetyRules {
         debug!("Incoming timeout message for round {}", timeout.round());
         COUNTERS.requested_sign_timeout.inc();
 
+        self.signer()?;
         self.verify_epoch(timeout.epoch())?;
 
         let preferred_round = self.persistent_storage.preferred_round()?;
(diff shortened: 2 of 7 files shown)


zaxguo:
@davidiw @zekun000 @JoshLind Guys please review this PR. Thanks!

consensus/safety-rules/src/safety_rules.rs
```diff
@@ -273,7 +334,11 @@ impl<T: Payload> TSafetyRules<T> for SafetyRules<T> {
                 .set_last_voted_round(timeout.round())?;
         }
 
-        let signature = timeout.sign(&self.validator_signer);
+        let validator_signer = self
```
davidiw, line 337: probably good to convert this into a function that returns a result
`fn signer(&self) -> Result<&ValidatorSigner, Error>`

consensus/safety-rules/src/safety_rules.rs
```diff
@@ -98,6 +99,37 @@ impl<T: Payload> SafetyRules<T> {
...
+            .validator_signer
+            .as_ref()
+            .ok_or(Error::NotInitialized)?;
+
+        // validator account not in the validator set
+        let public_key = validator_signer.public_key();
+        let author = validator_signer.author();
+        let validator_key = validator_verifier
+            .get_public_key(&author)
+            .ok_or(Error::InvalidAuthor)?;
```
davidiw, line 118: This is confusing, the error implies the author of a proposal doesn't exist within the validator verifier, perhaps we forgot to include this check in sign proposal :O

consensus/safety-rules/src/safety_rules.rs
```diff
@@ -98,6 +99,37 @@ impl<T: Payload> SafetyRules<T> {
...
+
+        // validator public key does not match
+        if public_key != validator_key {
+            let consensus_key = self
+                .persistent_storage
+                .consensus_key()
+                .expect("Unable to retrieve consensus private key");
+            self.validator_signer = Some(ValidatorSigner::new(author, consensus_key));
+            // Maybe we want to signal the safety rule with Error::InvalidPublicKey
+            return Err(Error::InvalidPublicKey);
```
davidiw, line 128: this isn't the right API, we need to request the consensus key given our current public key, look at https://github.com/libra/libra/blob/master/secure/storage/src/crypto_storage.rs#L45

consensus/safety-rules/src/safety_rules.rs
```diff
@@ -108,7 +140,28 @@ impl<T: Payload> SafetyRules<T> {
             .next_epoch_state()
             .cloned()
             .ok_or(Error::InvalidLedgerInfo)?;
+
+        // By definition, the signer key shall be the same as consensus_key
+        // in case we have an unmatched key, retrieve from the storage the appropriate key
+        match self.verify_key(&epoch_state) {
```
davidiw, line 146: probably better to log these where they happen and just have self.verify_key(&epoch_state)?;

consensus/safety-rules/src/error.rs
```diff
@@ -26,6 +26,12 @@ pub enum Error {
     #[error("No next_epoch_state specified in the provided Ledger Info")]
     InvalidLedgerInfo,
 
+    #[error("No matching author in the valdiator set")]
```
JoshLind, line 29: nit: valdiator -> validator

consensus/safety-rules/src/safety_rules.rs
```diff
@@ -108,7 +140,28 @@ impl<T: Payload> SafetyRules<T> {
...
+            // TODO: should we re-initialize the validator??
+            Err(Error::NotInitialized) => {
+                info!("Validator signer uninitialized.");
+            }
+            // TODO: This happens when the validator account is not even in the set
```
JoshLind, line 151: nit: for multi-line comments, it's common practice to use full stops and/or capitalization to mark separation between sentences (new lines can be unclear, for example, this comment and the one on line 145).

consensus/safety-rules/src/safety_rules.rs
JoshLind: As David says, we should call "export_private_key_for_version(..)" using the consensus key name and the public key currently registered in the epoch change. This is because this code may still fetch a consensus private key that does not match the one in the epoch change.

zaxguo: Thanks for the pointer to the right API! I'm also scratching my head in writing the tests for this part -- to test, safety rules must see epoch changes to update() while current tests under safety_rules seem "static": constructing proposal and checking a few properties, which do not update any epoch. Should I go look for other existing tests that triggers epoch changes for safety rules to update?

consensus/safety-rules/src/safety_rules.rs
zaxguo: I also find this tricky and what I do here is I'm trying to catch the possiblity that the signer is not in the validator set. And my understanding is sign_proposal() and epoch changing is two different thing; the signer verification being passed in sign_proposal (i.e. with verify_qc) cannot transfer during epoch changes. Therefore I tried to propagate the error ahd place some TODOs to see how that could be handled (I tried to panic!() but it does not work).

zekun000: the naming is confusing, we have invalid author as one of the verification error. this one means the node itself is not a validator anymore

zekun000: also i prefer to just panic in such a case instead of returning error

consensus/safety-rules/src/safety_rules.rs
```diff
@@ -34,7 +35,7 @@ use std::marker::PhantomData;
 /// set)
 pub struct SafetyRules<T> {
     persistent_storage: PersistentSafetyStorage,
-    validator_signer: ValidatorSigner,
+    validator_signer: Option<ValidatorSigner>,
```
zekun000, line 38: why this has to be Option?

consensus/safety-rules/src/safety_rules.rs
```diff
@@ -98,6 +108,31 @@ impl<T: Payload> SafetyRules<T> {
...
+            .ok_or(Error::InvalidAuthor)?;
+
+        if curr_key != expected_key {
+            let consensus_key = self
+                .persistent_storage
+                .consensus_key_for_version(expected_key)
+                .expect("Unable to retrieve consensus private key");
+            self.validator_signer = Some(ValidatorSigner::new(author, consensus_key));
+            // Maybe we want to signal the safety rule with Error::InvalidPublicKey
+            return Err(Error::InvalidPublicKey);
```
zekun000, line 131: this is not an error right? if it retrieves the right key

consensus/safety-rules/src/safety_rules.rs
zaxguo: I guess @davidiw suggested this change because it then makes the signer type aligned with verifier (i.e. both are Option<>)... But logically this shall never be None?

davidiw: It isn't initialized until we get our first epoch ending li, otherwise how do we know which private key to retrieve from storage.

consensus/safety-rules/src/safety_rules.rs
```diff
@@ -98,6 +106,32 @@ impl<T: Payload> SafetyRules<T> {
...
+            .verifier
+            .get_public_key(&validator_signer.author())
+        {
+            let curr_key = validator_signer.public_key();
+            if curr_key != expected_key {
+                let consensus_key = self
+                    .persistent_storage
+                    .consensus_key_for_version(expected_key)
+                    .expect("Unable to retrieve consensus private key");
+                debug!("Reconcile curr key {} using {}", curr_key, consensus_key);
```
davidiw, line 123: fortunately we elide our private keys, but in general we should never print private keys since logs can lead to compromises of private data :/
probably good enough to say Found public key in validator set and call it that.

consensus/safety-rules/src/safety_rules.rs
```diff
@@ -98,6 +106,32 @@ impl<T: Payload> SafetyRules<T> {
...
     }
 
+    /// This reconciles the key pair of a validator signer with a given validator set
+    /// during epoch changes.
+    fn reconcile_key(&mut self, epoch_state: &EpochState) -> Result<(), Error> {
+        let validator_signer = self.signer()?;
+        if let Some(expected_key) = epoch_state
+            .verifier
+            .get_public_key(&validator_signer.author())
+        {
```
davidiw, line 116: I'd make this two distinct lines to avoid the ugly line breaks

consensus/safety-rules/src/error.rs
```diff
@@ -26,6 +26,12 @@ pub enum Error {
     #[error("No next_epoch_state specified in the provided Ledger Info")]
     InvalidLedgerInfo,
 
+    #[error("No matching author in the validator set")]
+    InvalidAuthor,
+
```
davidiw, line 31: remove

consensus/safety-rules/src/safety_rules.rs
```diff
@@ -98,6 +106,32 @@ impl<T: Payload> SafetyRules<T> {
...
+            let curr_key = validator_signer.public_key();
+            if curr_key != expected_key {
+                let consensus_key = self
+                    .persistent_storage
+                    .consensus_key_for_version(expected_key)
```
davidiw, line 121: probably can remove the other api that gets the consensus key without version?

zaxguo: That one will be used when instantiating a new SafetRule instance, which passes only an author address and a PersistentStorage struct. At the time the verifier is still None hence cannot get the public key for the signer or get a consensus key with version.

consensus/safety-rules/src/safety_rules.rs
```diff
@@ -56,6 +57,16 @@ impl<T: Payload> SafetyRules<T> {
         }
     }
 
+    fn signer(&self) -> Result<&ValidatorSigner, Error> {
+        self.validator_signer.as_ref().ok_or(Error::NotInitialized)
```
davidiw, line 61: what's not initialized, we should probably make this take in a string to specify which field it failed on

consensus/safety-rules/src/safety_rules.rs
davidiw: pointer? I don't think we should be initializing a key / signer outside of this block.

consensus/safety-rules/src/safety_rules.rs
zekun000: then we probably should change the constructor too

consensus/safety-rules/src/error.rs
```diff
@@ -26,6 +26,9 @@ pub enum Error {
     #[error("No next_epoch_state specified in the provided Ledger Info")]
     InvalidLedgerInfo,
 
+    #[error("No matching public key found for the validator")]
+    NoMatchingPublicKey,
```
zekun000, line 30: not used?

consensus/safety-rules/src/tests/suite.rs
```diff
@@ -600,3 +607,144 @@ fn test_sign_proposal_with_early_preferred_round(func: Callback) {
...
+        &a1,
+        Some(&a1),
+        &signer,
+        Some(1),
+        Some(next_epoch_state),
+    );
+    assert!(safety_rules.update(a2.block().quorum_cert()).is_err());
+}
+
+fn test_missing_validator(func: Callback) {
```
davidiw, line 657: I think the test description could be cleaned up to be inclusive of what you're doing...

I think this test is too complicated for what it is attempting to do:
Ensure a proposal from a validator is within the ValidatorSet.

Then I would just get straight to the test that is... passing in a signed proposal from someone not in the validator set. No need to create a new verifier just a new signer and you're good to go.

consensus/safety-rules/src/tests/suite.rs
```diff
@@ -600,3 +607,144 @@ fn test_sign_proposal_with_early_preferred_round(func: Callback) {
...
+    let a2 = test_utils::make_proposal_with_parent_and_overrides(
+        vec![],
+        round + 2,
+        &a1,
+        Some(&a1),
+        &signer,
+        Some(1),
+        Some(next_epoch_state),
+    );
+    assert!(safety_rules.update(a2.block().quorum_cert()).is_err());
```
davidiw, line 654: unwrap_err please

davidiw: also its not really clear to me why we don't just make a1 with the rand signer and eliminate a2? do we need the epoch change?

consensus/safety-rules/src/tests/suite.rs
```diff
@@ -600,3 +607,144 @@ fn test_sign_proposal_with_early_preferred_round(func: Callback) {
...
+    let a2 = test_utils::make_proposal_with_parent_and_overrides(
+        vec![],
+        round + 2,
+        &a1,
+        Some(&a1),
+        &signer,
+        Some(1),
+        Some(next_epoch_state),
+    );
+    safety_rules.update(a2.block().quorum_cert()).unwrap();
```
davidiw, line 733: sweet

davidiw: might be good to call out at each block of code what little snippet you are testing

consensus/safety-rules/src/tests/suite.rs
zaxguo: This is because the whole tests are about key reconciliation during epoch change, requiring the test code to actually trigger "update(...)". 
Making a a legal a1 is necessary to pass the QC verification in the 1st line of update(...), and thus enabling the test code to reach the part I'm trying to test here -- the validator tries to grab a its new consensus key but fails due to its acct is not in the set.

consensus/safety-rules/src/tests/suite.rs
```diff
@@ -600,3 +607,144 @@ fn test_sign_proposal_with_early_preferred_round(func: Callback) {
...
+        Some(1),
+        Some(next_epoch_state),
+    );
+    assert!(safety_rules.update(a2.block().quorum_cert()).is_err());
+}
+
+fn test_missing_validator(func: Callback) {
+    // Testing for an validator missing from the validator set
+    // It does so by updating the safey rule to an epoch state, which does not contain the
+    // current validator. The validator later fails to verify QC with its signer key.
```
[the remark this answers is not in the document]
zaxguo: MB. Text description is wrong. Just fixed.

consensus/safety-rules/src/tests/suite.rs
zaxguo: I've already done the test you mentioned in sign_proposal_with_bad_signer(). The difficulty of testing key reconciliation during epoch changes is two-fold: 1) to emulate epoch changes, because only in this way the code can be exercised, 2) to achieve high coverage, which requires me to carefully construct inputs in various forms -- each fails one stage of test but will fail progressively. Does this explain why my tests seem to more redundant than other tests? :)

consensus/safety-rules/src/safety_rules.rs
davidiw: @zaxguo, it doesn't seem like we have refactored the constructor (new method)

consensus/safety-rules/src/safety_rules.rs
```diff
@@ -27,26 +28,40 @@ use libra_types::{
...
 }
 
 impl SafetyRules {
     /// Constructs a new instance of SafetyRules with the given persistent storage and the
     /// consensus private keys
     /// @TODO replace this with an API that takes in a SafetyRulesConfig
+    /// @TODO change the constructor to fit Option<validator_signer> and address the circular
+    /// dependency with reconcile_key() during initializationg
     pub fn new(author: Author, persistent_storage: PersistentSafetyStorage) -> Self {
         let consensus_key = persistent_storage
```
davidiw, line 42: we don't know the public key at this point in time, we should not instantiate this.

consensus/safety-rules/src/test_utils.rs
```diff
@@ -187,3 +194,14 @@ pub fn test_storage(signer: &ValidatorSigner) -> PersistentSafetyStorage {
     let storage = Box::new(InMemoryStorage::new());
     PersistentSafetyStorage::initialize(storage, signer.private_key().clone(), waypoint)
 }
+
+pub fn rotate_consensus_key(
```
davidiw, line 198: if this isn't used in more than one place, let's put it in the same file where it is used. Or just do it inline since it is only used once.

consensus/safety-rules/src/tests/suite.rs
```diff
@@ -600,3 +607,144 @@ fn test_sign_proposal_with_early_preferred_round(func: Callback) {
...
+
+    let a1 = test_utils::make_proposal_with_qc(round + 1, genesis_qc, &signer);
+    let err = safety_rules.update(a1.block().quorum_cert()).unwrap_err();
+    assert_eq!(err, Error::NotInitialized);
+
+    safety_rules.initialize(&proof).unwrap();
+    assert!(safety_rules.update(a1.block().quorum_cert()).is_ok());
+}
+
+#[allow(dead_code)]
```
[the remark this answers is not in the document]
davidiw: Please respond to this comment rather than resolving it without response. In general, for discussions, you should let the one that opened the discussion to resolve it.

On the other hand, if it is a feedback such as rename this to that or use a filter_map instead of a map + filter then feel free to resolve yourself.

consensus/safety-rules/src/tests/suite.rs
```diff
@@ -600,3 +607,144 @@ fn test_sign_proposal_with_early_preferred_round(func: Callback) {
...
+        Some(&a1),
+        &signer,
+        Some(1),
+        Some(next_epoch_state),
+    );
+    assert!(safety_rules.update(a2.block().quorum_cert()).is_err());
+}
+
+fn test_missing_validator(func: Callback) {
+    // Testing for an validator missing from the validator set
```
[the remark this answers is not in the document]
davidiw: I think you responded to these outside of this box but that takes me and others following along quite a bit more effort to make sense of.

consensus/safety-rules/src/tests/suite.rs
```diff
@@ -600,3 +607,151 @@ fn test_sign_proposal_with_early_preferred_round(func: Callback) {
...
+        None,
+    );
+    let err = safety_rules.update(a3.block().quorum_cert()).unwrap_err();
+    assert_eq!(
+        err,
+        Error::InvalidQuorumCertificate("Fail to verify QuorumCert".into())
+    );
+}
+
+fn test_reconcile_key(_func: Callback) {
```
davidiw, line 706: If you follow the feedback given in this and previous reviews, the recommendation has been that there is "None" set for the current validator signer and hence at initialization we will already be testing reconciliation. I would argue we want just want one positive test that shows we can retrieve the key even if a rotation has happened in storage. And then we rotate to that key in a QC and reconciliation updates to the new key. Thoughts?

consensus/safety-rules/src/tests/suite.rs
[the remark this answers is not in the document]
davidiw: note "an" validator is not grammatically correct.

consensus/safety-rules/src/tests/suite.rs
davidiw: Per the comments given in the PR, we should be going through reconciliation every time we start a clean safety rules and hit initialize.

consensus/safety-rules/src/tests/suite.rs
zaxguo: Sorry about the resolving conversation thing, which costs you more time to realize the changes -- I wasn't aware of the right practice and will do it correctly afterwards. 
I agree with the test method. 
To be more specific, the TODO items are: 
1. refactor the constructor, 
2. reduce test cases into 2 (1 pos + 1 neg).
On them..

consensus/safety-rules/src/tests/suite.rs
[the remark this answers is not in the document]
zaxguo: Sorry about not notifying the changes but resolving the conversation myself..

consensus/safety-rules/src/tests/suite.rs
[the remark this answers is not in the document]
zaxguo: Plan added in comment of the code and is_ok() refactored.

consensus/safety-rules/src/safety_rules.rs
zaxguo: @davidiw @zekun000  Guys, how about the current one? Had to do some dirty hack in reconcile_key make it work with existing tests

consensus/safety-rules/src/safety_rules.rs
```diff
@@ -107,24 +110,29 @@ impl SafetyRules {
     ///     1. Validator not in the set
     ///     2. Validator in the set, but no matching key found in storage
     fn reconcile_key(&mut self, epoch_state: &EpochState) -> Result<(), Error> {
-        let signer = self.signer()?;
+        // fake a signer using the same author in case we have not initialized
+        let tmp_signer = {
+            let mut rng = StdRng::from_seed([8u8; 32]);
+            ValidatorSigner::new(self.author, Ed25519PrivateKey::generate(&mut rng))
+        };
```
zaxguo, line 117: I'm not quite sure if I should do this... This is in case we have not initialized a validator_signer yet, it will generate one with the account address when safety rule is first intialized. Then during intialization, safety rule will reconcile the generated key with the one in persistent storage.. @davidiw

consensus/safety-rules/src/safety_rules.rs
```diff
@@ -89,17 +104,53 @@ impl SafetyRules {
         }
     }
 
+    /// This reconciles the key pair of a validator signer with a given validator set
+    /// during epoch changes.
+    /// @TODO Given we cannot panic, we must handle the following two error cases:
+    ///     1. Validator not in the set
+    ///     2. Validator in the set, but no matching key found in storage
```
davidiw, line 111: this comment is strange, we should never panic, I don't know why we would have this comment here or written like this. Perhaps you want to say we shouldn't make consensus panic, but I think that is actually okay.

consensus/safety-rules/src/safety_rules.rs
davidiw: I think you're trying too hard to conform this to your original code, instead what you likely want is:

```
let curr_key = self.signer.map(|s| s.public_key()); // this is now either None or Some(public_key);
if curr_key != Some(expected_key) {
   // reconcile
} else {
  // matches
}
```

consensus/safety-rules/src/tests/suite.rs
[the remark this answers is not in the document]
davidiw: panic is not the right word

@zekun000, if our key isn't in the list, do we want to keep running or panic? The panic happens in the consensus code, which expects safety rules to update successfully.

@zaxguo, even if we do not want to panic, we can still return an enum on the Ok that implies whether or not we are present in the list.

consensus/safety-rules/src/tests/suite.rs
```diff
@@ -600,3 +607,151 @@ fn test_sign_proposal_with_early_preferred_round(func: Callback) {
...
+}
+
+// Tests for fetching a missing validator key from persistent storage.
+// Ideally under this circumstance, safety rules should panic. However,
+// at current stage of implementation the upper layer (i.e. RoundManager) is
+// unable to handle such failure (actually it is unable ANY failure
+// during key reconciliation!). For more fail cases, refer to debug messages in
+// reconcile_key()
+#[allow(dead_code)]
+fn test_a_missing_validator_key(func: Callback) {
```
davidiw, line 633: again the naming here is awkward... this is explicitly 
```test_key_not_in_store```

File: consensus/safety-rules/src/tests/suite.rs [hunk 27]
```diff
@@ -600,3 +607,151 @@ fn test_sign_proposal_with_early_preferred_round(func: Callback) {
...
+        &a1,
+        Some(&a1),
+        &signer,
+        Some(1),
+        Some(next_epoch_state),
+    );
+    safety_rules.update(a2.block().quorum_cert()).unwrap_err();
+}
+
+fn test_a_missing_validator(func: Callback) {
```
davidiw, line 659: test_not_in_validator_set

consensus/safety-rules/src/safety_rules.rs
```diff
@@ -27,26 +33,38 @@ use libra_types::{
...
-        let consensus_key = persistent_storage
-            .consensus_key()
-            .expect("Unable to retrieve consensus private key");
-        let validator_signer = ValidatorSigner::new(author, consensus_key);
         Self {
             persistent_storage,
-            validator_signer,
+            validator_signer: None,
             validator_verifier: None,
+            author,
```
davidiw, line 52: if you want to do this here, we should follow up with another PR, but this is good for now.

consensus/safety-rules/src/safety_rules.rs
zekun000: It was me suggesting we should just panic if the key not found but then realize we can't

zaxguo: Just fixed the comment here. Does that make sense now?

consensus/safety-rules/src/safety_rules.rs
zaxguo: This is fixed in latest commit

consensus/safety-rules/src/safety_rules.rs
```diff
@@ -89,17 +97,46 @@ impl SafetyRules {
...
+                let consensus_key = self
+                    .persistent_storage
+                    .consensus_key_for_version(expected_key)
+                    .expect("Unable to retrieve consensus private key");
+                self.validator_signer = Some(ValidatorSigner::new(self.author, consensus_key));
+            } else {
+                debug!("Validator key matches the key in validator set.");
+            }
+        } else {
+            debug!("The validator key is not in store!");
```
davidiw, line 123: if it isn't in the store, we should None our signer otherwise it might still be initialized and we might do something bad

consensus/safety-rules/src/safety_rules.rs
```diff
@@ -89,17 +97,46 @@ impl SafetyRules {
...
+    /// Current impl of consensus does not expect key reconciliation to panic or fail.
+    /// Instead, we signal debug messages indicating following error causes:
+    ///     1. Validator not in the set
+    ///     2. Validator in the set, but no matching key found in storage
+    fn reconcile_key(&mut self, epoch_state: &EpochState) -> Result<(), Error> {
+        if let Some(expected_key) = epoch_state.verifier.get_public_key(&self.author) {
+            let curr_key = self.signer().ok().map(|s| s.public_key());
+            if curr_key != Some(expected_key.clone()) {
+                debug!(
+                    "Reconciled pub key for signer {} [{:#?} -> {:#?}]",
```
davidiw, line 111: don't the keys support display?, I think you can just do {} -> {}

consensus/safety-rules/src/test_utils.rs
davidiw: please address

consensus/safety-rules/src/tests/suite.rs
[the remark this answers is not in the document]
davidiw: perhaps we can have a flag in consensus state that we can check to see if this node is in the validator set that way you can write a proper test here.

consensus/safety-rules/src/tests/suite.rs
davidiw: I think the tests are okay, because they test distinct things. I do wonder the reconcile is a little long, but I'm okay with just wrapping up at this point :)

consensus/safety-rules/src/safety_rules.rs
zaxguo: curr_key is of Option; fixed in latest commit

consensus/safety-rules/src/safety_rules.rs
zaxguo: Fixed in the latest commit

consensus/safety-rules/src/tests/suite.rs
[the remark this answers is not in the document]
zaxguo: > perhaps we can have a flag in consensus state that we can check to see if this node is in the validator set that way you can write a proper test here.

This is execellent advice. I adopted it in the latest commit, which greatly simply one test_validator_not_in_set.

consensus/safety-rules/src/safety_rules.rs
```diff
@@ -89,17 +97,47 @@ impl SafetyRules {
...
+            let curr_key = self.signer().ok().map(|s| s.public_key());
+            if curr_key != Some(expected_key.clone()) {
+                debug!(
+                    "Reconciled pub key for signer {} [{:#?} -> {}]",
+                    self.author, curr_key, expected_key
+                );
+                let consensus_key = self
+                    .persistent_storage
+                    .consensus_key_for_version(expected_key)
+                    .expect("Unable to retrieve consensus private key");
```
davidiw, line 117: This isn't production worthy code.

1) Operator starts safety rules and lets it run for a long time
2) Operator over-rotates and now the key for the next round isn't actually in secure storage
3) Operator then fixes issue

The node will always crash when seeing it isn't in the set, even though there's not much that we can do.

Alternatively consider the case where
1) SafetyRules is nice and far into the future
2) There was a corruption of state and we need to start from a time before our current keys but that will lead up to our current keys
3) As we replay the updates, safety rules is going to get a set of keys that they no longer have access to

This should be an error and not a panic, this is recoverable. Please adjust.

cc: @zekun000

consensus/safety-rules/src/tests/safety_rules.rs
```diff
@@ -2,13 +2,48 @@
...
 
 #[test]
 fn test() {
     suite::run_test_suite(safety_rules);
 }
 
+// Tests for fetching a missing validator key from persistent storage.
+#[test]
+#[should_panic(expected = "Unable to retrieve consensus private key")]
+fn test_key_not_in_store() {
```
davidiw, line 18: again should be an error so it can go back into the suite

consensus/safety-rules/src/tests/suite.rs
```diff
@@ -67,6 +70,10 @@ pub fn run_test_suite(func: Callback) {
     test_sign_proposal_with_bad_signer(func);
     test_sign_proposal_with_invalid_qc(func);
     test_sign_proposal_with_early_preferred_round(func);
+    test_uninitialized_signer(func);
+    test_reconcile_key(func);
+    test_validator_not_in_set(func);
+    // test_key_not_in_store(func);
```
davidiw, line 76: ?

consensus/safety-rules/src/safety_rules.rs
davidiw: also to be clear, if we do not find the key, we should set the signer to none.

zaxguo: Fixed. Now it returns an Error::InternalError{"Validator key not found"} instead of just panic()

consensus/safety-rules/src/tests/safety_rules.rs
zaxguo: Fixed

consensus/safety-rules/src/tests/suite.rs
zaxguo: Fixed

zaxguo:
> Two changes for future:
> 
> 1. ConsensusState should just have public fields and not getters, it isn't a shared state, so who really cares
> 2. Author should be retrieved via secure storage (and inserted during test)

Address these two in another PR?

consensus/safety-rules/src/safety_rules.rs
```diff
@@ -264,8 +263,6 @@ impl TSafetyRules for SafetyRules {
 
         self.persistent_storage
             .set_last_voted_round(proposed_block.round())?;
-
-        let validator_signer = self.signer()?;
```
zaxguo: Here is a potential bug where signer is not initialized but consensus state is already updated. Fixed in latest commit.

davidiw: how would we get here? this means consensus never initialized safety rules, no?

zaxguo: Right. If the safety rule is not initialized (e.g. the key is not found) but this still directly called, it will increment the last voted round, which is definitely not right. I spotted this while writing the specs.

davidiw: ahhh good call

zaxguo:
Memo -- 
Two changes for future:

1. ConsensusState should just have public fields and not getters, it isn't a shared state, so who really cares
2. Author should be retrieved via secure storage (and inserted during test)
--
Code refactor to address redundancy

PR #4181, opened 2020-06-02
[network] Allow pulling network identity from secure storage

davidiw:
This PR is pretty straight forward:

* Add a means to specify where the networking key is coming from (config or storage)
* Actually enable pulling the keys from storage
* Enable pulling peer id from storage
* Update management smoke test to ensure proper end-to-end correctness, so awesome
* Remove no noise transports from config
* Make the default networking config valid by having a key / peer id
* Updates the peer id config handling to match our hardening issue: https://github.com/libra/libra/issues/3985

The management smoke tests is the proof of correctness

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2797,6 +2797,7 @@ dependencies = [
  "libra-logger 0.1.0",
  "libra-mempool 0.1.0",
  "libra-metrics 0.1.0",
+ "libra-secure-storage 0.1.0",
  "libra-types 0.1.0",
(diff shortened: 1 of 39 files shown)


davidiw:
Realized that the identity in networking should be both key and id, so I'm updating this. Will push this to draft for now.

davidiw:
@bors-libra r=zekun000

feel free to give comments for follow ups, landing this before I get more breaking issues...

this changes our config files

PR #4186, opened 2020-06-02
[Key Manager] Moved the Key Manager Config out of Node Config.

JoshLind:
## Motivation

At present, the KeyManagerConfig lives inside NodeConfig. This is problematic from a deployment perspective, as the Key Manager is a self-contained entity that should be deployable without requiring irrelevant configuration details (e.g., ValidatorConfig). This PR moves KeyManagerConfig out of NodeConfig and into its own high-level domain. To achieve this, we offer two commits:

1. We define a key_manager_config, remove secure_config and create secure_backend_config for secure storage related configurations.

2. We update all key manager related tests (including the smoke test).

Yes.

## Test Plan

All tests still pass (including the smoke test).

## Related PRs

None.

### DIFF
--- a/config/src/config/key_manager_config.rs
+++ b/config/src/config/key_manager_config.rs
@@ -1,0 +1,64 @@
+// Copyright (c) The Libra Core Contributors
+// SPDX-License-Identifier: Apache-2.0
+
+use crate::config::{LoggerConfig, PersistableConfig, SecureBackend};
+use anyhow::Result;
+use libra_types::account_address::AccountAddress;
+use serde::{Deserialize, Serialize};
+use std::path::{Path, PathBuf};
+
+const DEFAULT_JSON_RPC_ENDPOINT: &str = "https://127.0.0.1:8080";
+
+// Timing related defaults
+const DEFAULT_ROTATION_PERIOD_SECS: u64 = 604_800; // 1 week
+const DEFAULT_SLEEP_PERIOD_SECS: u64 = 600; // 10 minutes
+const DEFAULT_TXN_EXPIRATION_SECS: u64 = 3600; // 1 hour
+
+#[derive(Clone, Debug, Deserialize, PartialEq, Serialize)]
+#[serde(default, deny_unknown_fields)]
+pub struct KeyManagerConfig {
+    pub rotation_period_secs: u64,
+    pub sleep_period_secs: u64,
+    pub txn_expiration_secs: u64,
+
+    pub json_rpc_endpoint: String,
+    pub validator_account: AccountAddress,
+
+    pub logger: LoggerConfig,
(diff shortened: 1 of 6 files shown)


config/src/config/key_manager_config.rs
```diff
@@ -0,0 +1,64 @@
...
+
+#[derive(Clone, Debug, Deserialize, PartialEq, Serialize)]
+#[serde(default, deny_unknown_fields)]
+pub struct KeyManagerConfig {
+    pub rotation_period_secs: u64,
+    pub sleep_period_secs: u64,
+    pub txn_expiration_secs: u64,
+
+    pub json_rpc_endpoint: String,
+    pub validator_account: AccountAddress,
```
davidiw, line 25: this is in secure storage, if not, we need to add it :O

crap more work for me, I think

testsuite/tests/libratest/smoke_test.rs
```diff
@@ -1230,23 +1232,31 @@ fn test_key_manager_consensus_rotation() {
...
+
+    let mut key_manager_config = KeyManagerConfig::default();
+    key_manager_config.json_rpc_endpoint = json_rpc_endpoint.clone();
+    key_manager_config.rotation_period_secs = 10;
+    key_manager_config.sleep_period_secs = 10;
+    key_manager_config.validator_account = node_config.validator_network.clone().unwrap().peer_id;
+    let mut on_disk_storage_config = OnDiskStorageConfig::default();
+    on_disk_storage_config.path =
+        node_config_path.with_file_name(on_disk_storage_config.path.clone());
+    key_manager_config.secure_backend = SecureBackend::OnDiskStorage(on_disk_storage_config);
```
davidiw, line 1249: there's kind of a fun race condition you're playing with here, lol

config/src/config/key_manager_config.rs
JoshLind: Thanks, David! As discussed offline, I'll come back and remove this from the config once your PR lands!

JoshLind:
Thanks, David! I'm going to ship this as-is for now. Regarding your comment about the potential smoke test race condition in the on-disk storage: I know what you mean and this is a part of the reason I've been keeping a close eye on flaky tests. 

For what it's worth: I haven't ever seen a case where LSR and the KM execute this race (i.e., modify the on-disk file concurrently). I think the nature of the test avoids this because LSR doesn't do anything with storage during a rotation. For now, I'll leave this here and keep an eye on it. Once Liwei adds his tests, or this test becomes significantly flaky, I'll remove it. Note: the flakiness behaviour of the test has changed a little since Lera modified things, but nothing significant -- this is more a result of the test taking a while to execute.

@bors-libra r=davidiw

PR #4191, opened 2020-06-02
[network] Remove enable_remote_authentication and enable_noise flags

agc-quadra:
## Motivation

Remove the enable_remote_authentication and enable_noise flags.  Their current uses have them always set to true.

yes

## Test Plan

Don't break existing integration and unit tests.

## Related PRs


agc-quadra:
@davidiw @phlip9 @zekun000 

Following up to #4140 to  remove the use_remote_authentication flag.

Note for @davidiw :  this change is not on its own enough to eliminate the Option type for the connectivity_manager_tx channel in NetworkBuilder.

zekun000:
that's my bad,  I suggested to remove it because I got a sense it's not used

agc-quadra:
> that's my bad, I suggested to remove it because I got a sense it's not used

@zekun000:  @davidiw  actually suggested removing it as well.   I think the core issue here is that the use of the flag is not really clear.   And to complicate matters, neither flag is used in the codebase (?!?)

For my context:  what is the intended meaning of "noise" and "remote_authentication"?   Specifically, what features are disabled when the two flags are not set?  and why does one depend on the other?

davidiw:
Lots of activity in this space, I took out a massive chunk of code in this PR: https://github.com/libra/libra/pull/4181
specifically this commit:
https://github.com/libra/libra/pull/4181/commits/365dbd5ec97817e8750a702220f68c3066c42fde

we have to address changing the default network too, which now requires a key being provided since we cannot expect average users to generate a key on demand...

davidiw:
(Thanks for taking a stab at this, I wouldn't have bothered, but the same deadline I had mentioned for the onchain discovery is also for that work.)

agc-quadra:
> (Thanks for taking a stab at this, I wouldn't have bothered, but the same deadline I had mentioned for the onchain discovery is also for that work.)

It was either an easy way to pay off tech debt (sadly no) or it provides additional insight and understanding (always a positive)

agc-quadra:
> So I think the gist of it is that we want to require noise but still leave remote auth optional. How would a public node connect to a validator full node otherwise? If both peers enable remote auth, this would require the validator full node knows everyone else.

At present remote_auth and noise are both always set to true and never instantiated with false.

Looking forward, if I understand your statement correctly the allowable states are:

 remote_auth = {true, false}
  noise = {true}

If so, I can look into making that apply to the code.

davidiw:
I think that all this code has been updated in one of my latest PRs, but if there are any dangling pieces, yes, a PR to clean that up would be great :)

agc-quadra:
Minimal value add from continuing this PR in teh context of @davidiw changes.

Closing it.

PR #4195, opened 2020-06-02
[move core modules] remove create_account script

sblackshear:
This is another relic of 6/18 Libra that must be removed.

## Test Plan

Refactored existing tests that use the script


sblackshear:
Correct; this is a separate (and infrequently used, except in tests) script.

PR #4198, opened 2020-06-02
[move-prover] Disallow old in aborts_if + fix test

kkmc:
Disallow the use of old in aborts_if. Aborts_if already refers to
the old variables (at the beginning of a procedure). It is
confusing to allow someone to write old(v) and v in
aborts_if while both refer to the same value.

Changes:
    - Disallow old in aborts_if
    - Fix all tests by removing old in aborts_if conditions

## Motivation

Refer to description above.

Yes.

## Test Plan

Existing test suite. Run the following:

```
cargo test
```

## Related PRs

None.

### DIFF
--- a/language/move-prover/spec-lang/src/ast.rs
+++ b/language/move-prover/spec-lang/src/ast.rs
@@ -77,7 +77,7 @@ impl ConditionKind {
     /// Returns true of this condition allows the `old(..)` expression.
     pub fn allows_old(&self) -> bool {
         use ConditionKind::*;
-        matches!(self, Ensures | AbortsIf | InvariantUpdate | VarUpdate(..))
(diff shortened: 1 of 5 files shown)


kkmc:
@DavidLDill @wrwg

kkmc:
Yes. That's the way I set it up and just double checked that it does indeed run the files.



> Looks great. I can trigger merge for you if the PR is ready for it; let me know.
> 
> Just to check: have you installed Z3 and boogie and set the environment variables (see move-prover/scripts)? Only if so, `cargo test` in the move-prover dir will do something useful (it will silently succeed if not set)

Yes. That's the way I set it up and just double checked that it does indeed run the files (fails when I add incorrect assertions to some of the specified files).

PR #4203, opened 2020-06-02
[move-prover] Added loops.move

junkil-park:
- Created `functional/loops.move` and added tests cases for verifying aborts_if specs of the functions with loops
    - Set the "module pragma verify" to be true to start working
    - See the TODOs and FIXMEs in the file

- Specified and verified the postcondition of Vector::index_of
    - Two FIXMEs are added to `functional/verify_vector.move` which are related to the verification of aborts_if

## Motivation

To verify the standard vector module with loops, and add test cases for verifying loops

Yes

## Test Plan

cargo test

### DIFF
--- a/language/move-prover/tests/sources/functional/loops.move
+++ b/language/move-prover/tests/sources/functional/loops.move
@@ -1,0 +1,88 @@
+module VerifyLoops {
+
+    spec module {
+        pragma verify=false;
+    }
+
+    // ----------------------
+    // `aborts_if` statements
+    // ----------------------
+
+    public fun iter10_missing_inc_spec1() {
+        let i = 0;
+        while (i <= 10) { // an infinite loop
+            if (i > 10) abort 10;
+        }
+    }
+    spec fun iter10_missing_inc_spec1 { // Verified. This is expected because Prover checks the partial correctness of this function which contains an infinite loop.
+        aborts_if false;
+        ensures false;
+    }
+
+    public fun iter10_missing_inc_spec2() {
+        let i = 0;
+        while (i <= 10) { // an infinite loop
+            if (i > 10) abort 10;
+        }
+    }
+    spec fun iter10_missing_inc_spec2 { // Verified. This is expected because Prover checks the partial correctness of this function which contains an infinite loop.
+        aborts_if true;
+        ensures false;
+    }
+
+    public fun iter10_no_abort() {
+        let i = 0;
+        while ({
+            spec { assert i <= 11; };
+            (i <= 10)
+        }) {
+            if (i > 10) { spec { assert false; }; abort 10 };
+            i = i + 1;
+        }
+    }
+    spec fun iter10_no_abort { // FIXME: Should be proved.
+        aborts_if false;
+    }
+
+    public fun iter10_no_abort_incorrect() {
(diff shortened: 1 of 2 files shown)


language/move-prover/tests/sources/functional/loops.move
```diff
@@ -0,0 +1,86 @@
...
+    // ----------------------
+
+    public fun iter10_missing_inc_spec1() {
+        let i = 0;
+        while (i <= 10) { // an infinite loop
+            if (i > 10) abort 10;
+        }
+    }
+    spec fun iter10_missing_inc_spec1 { // TODO: What is the expected output for this spec?
+        aborts_if false;
```
shazqadeer, line 18: I think this will verify because aborts_if is a postcondition and Move Prover should be able to prove that control will never reach the procedure exit.

language/move-prover/tests/sources/functional/loops.move
```diff
@@ -0,0 +1,86 @@
...
+    }
+
+    public fun iter10_missing_inc_spec2() {
+        let i = 0;
+        while (i <= 10) { // an infinite loop
+            if (i > 10) abort 10;
+        }
+    }
+    spec fun iter10_missing_inc_spec2 { // TODO: What is the expected output for this spec?
```
shazqadeer, line 27: This should also verify.  It really does not matter what postcondition you write. They will all verify.

language/move-prover/tests/sources/functional/loops.move
```diff
@@ -0,0 +1,86 @@
...
+        aborts_if true;
+    }
+
+    public fun iter10_no_abort() {
+        let i = 0;
+        while ({
+            spec { assert i <= 11; };
+            (i <= 10)
+        }) {
+            if (i > 10) abort 10;
```
shazqadeer, line 37: Are you able to prove "assert false" just before abort 10? That is, make the conditional inside the loop body like this:

if (i > 10) {
    spec { assert false;}; abort 10
}

language/move-prover/tests/sources/functional/loops.move
junkil-park: I see. I noticed that Prover proved both `iter10_missing_inc_spec1` and `iter10_missing_inc_spec2`. I just wondered what output from Prover we desire in these cases.

language/move-prover/tests/sources/functional/loops.move
junkil-park: Got it. Our spec asserts the partial correctness of the functions, so any spec is supposed to be simply true when the function contains an infinite loop does not terminate?

language/move-prover/tests/sources/functional/loops.move
junkil-park: Yes, I just check that the following is verified (Prover finds no errors):
```
    public fun iter10_no_abort() {
        let i = 0;
        while ({
            spec { assert i <= 11; };
            (i <= 10)
        }) {
            if (i > 10) { spec { assert false; }; abort 10 };
            i = i + 1;
        }
    }
    spec fun iter10_no_abort { // FIXME: Should be proved.
        pragma verify=true;
        //aborts_if false;
    }
```

PR #4204, opened 2020-06-02
[cluster test] Use client side latency in cluster version of--emit-tx

snakech:
## Motivation

(Write your answer here.)

## Test Plan

## Related PRs

### DIFF
--- a/testsuite/cluster-test/src/main.rs
+++ b/testsuite/cluster-test/src/main.rs
@@ -24,10 +24,8 @@ use cluster_test::{
     prometheus::Prometheus,
     report::SuiteReport,
     slack::SlackClient,
-    stats,
     suite::ExperimentSuite,
     tx_emitter::{AccountData, EmitJobRequest, EmitThreadParams, TxEmitter, TxStats},
-    util::unix_timestamp_now,
 };
 use futures::{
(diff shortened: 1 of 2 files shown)


testsuite/cluster-test/src/main.rs
```diff
@@ -382,6 +380,13 @@ impl BasicSwarmUtil {
     }
 }
 
+impl std::ops::Deref for ClusterUtil {
```
andll, line 383: Deref is pretty reserved thing for wrapper types, I don't think it makes sense to have it here.
Maybe just refactor `emit_tx` method, instead of having it as an instance method on `BasicSwarmUtil`, you can make it regular function (outside of any class) that takes &Cluster as an argument?

snakech: I tried to make BasicSwarmUtil to be the base class and ClusterUtil to be the child class, which emit_tx can be Polymorphism for both class.

But yes, I can move emit_tx to be a out side normal fn for both of them instead of what I was thinking

andll: Yeah, I don't think we want to next BasicSwarmUtil into ClusterUtil just for that

PR #4205, opened 2020-06-02
[move source language] Cleanup hex and byte strings

tnowacki:
## Motivation

- Moved decoding to expansion
- Parser now keeps the string syntax exact, even if invalid
- Required a split of parser and expansion value representation
- Cleaned up byte string decoding

## Test Plan

- cargo test


language/move-lang/src/expansion/byte_string.rs
```diff
@@ -0,0 +1,140 @@
...
+            }
+            c => {
+                context.error(cur, cur + 2, format!("Invalid escape sequence: '\\{}'", c));
+            }
+        }
+    }
+}
+
+fn push(buffer: &mut Vec<u8>, ch: char) {
+    let mut next = vec![0; ch.len_utf8()];
```
bob-wilson, line 137: Since we decided not to support Unicode escapes in byte strings, these are purely ASCII strings and there is no need to encode the bytes to UTF-8. (And, we should not do that since it is confusing.)

tnowacki: Agreed. But it was the only I could figure out to convert it into u8. 

Let me play around with a few other things. If I can't get it to work, I'll leave it as is

tnowacki: Oops, turns I can just cast it to `u8`. Finding out what things cast into what things is Rust documentation is a little tricky....

PR #4207, opened 2020-06-02
[libra-dev] Add function to create different transaction scripts

sunmilee:
## Motivation

Adding function in libra-dev to create different transaction scripts. Refactored signed transaction function to take in bytes of script object rather than individual parameters. 

## Test Plan

cargo x test -p libra-dev


thefallentree:
looks great, for completeness pls add an test function for libra_ScriptP2P_from() as well.

PR #4208, opened 2020-06-02
[move source language] Improved trailing semi warnings

tnowacki:
## Motivation

- Long standing TODO to improve error messages for unnecessary trailing semicolons
- Explained why they are unnecessary and in some sense "dead code"
- See https://community.libra.org/t/odd-error-when-semi-is-put-after-break-or-continue/2868 for more discussion

## Test Plan

- new tests

### DIFF
--- a/language/move-lang/src/hlir/translate.rs
+++ b/language/move-lang/src/hlir/translate.rs
@@ -283,6 +283,7 @@ fn function_body(
             }
             let (mut locals, used) = context.extract_function_locals();
             let unused = check_unused_locals(context, &mut locals, used);
+            check_trailing_unit(context, &mut body);
             remove_unused_bindings(&unused, &mut body);
             context.signature = None;
             context.has_return_abort = false;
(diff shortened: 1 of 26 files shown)


language/move-lang/tests/move_check/liveness/trailing_semi.move
```diff
@@ -0,0 +1,64 @@
+script {
+    fun main() {
+        loop {
```
bob-wilson, line 3: I don't see a testcase where the divergent expression is an infinite loop (with no break). We should have one of those, right?

language/move-lang/src/hlir/translate.rs
```diff
@@ -1415,6 +1421,101 @@ fn bind_for_short_circuit_sequence(seq: &T::Sequence) -> bool {
...
+                | Some(hcmd!(_, C::Continue))
+                | Some(hcmd!(_, C::Abort(_)))
+                | Some(hcmd!(_, C::Return(_)))
+                | Some(hexp!(_, E::Unreachable))
+        )
+    }
+    macro_rules! invalid_trailing_unit {
+        ($context:ident, $loc:expr, $uloc:expr) => {{
+            let semi_msg = "Unnecessary trailing ';'";
+            let unreachable_msg = "Divergent expression. Any code following the evaluation of \
```
bob-wilson, line 1453: I think we should try to improve this message.

- "Unnecessary" seems like the wrong word. This isn't just a suggestion to simplify the code; we're treating this as invalid. So maybe change it to "Invalid"?
- The unreachable_msg has too much jargon, specifically "divergent" and "dead". Maybe drop "Divergent expression" and replace the whole thing with: "Any code after this expression will not be reached"?
- The info_msg seems pretty good as-is

language/move-lang/tests/move_check/liveness/trailing_semi.move
tnowacki: Good catch! I'll add one

PR #4209, opened 2020-06-02
[network] fix flaky gossip discovery test

phlip9:
In removing signatures from the gossip discovery protocol, the tests
execute much faster. The `get_unix_epoch()` timer is not fine enough and
so the new note ended up with the same epoch counter as the old note.

To fix this issue, we just mock out the timer. There's another,
unmockable timer in the `outbound` test. However, there we just add a
100000 ms buffer, which should be Good Enough^tm


davidiw:
uh oh, conflict found :/

phlip9:
bors pls :d

PR #4211, opened 2020-06-02
[on-chain config] add event subscription for reconfig notif to support writeset txn

christinacelee:
## Motivation

Add ability to subscribe to events for on-chain config notifications from state sync
Mostly to support writeset txns triggering new epoch change event and sending consensus reconfig notification


runtian-zhou:
Can confirm that this PR now works with the writeset by locally rebasing it onto #3887 and the upgrade just works as expected.

PR #4212, opened 2020-06-03
[libra-crypto-derive] fix proc-macro declaration

sunshowers:
`proc_macro` appears to work with the v1 resolver and for `cargo metadata` (and therefore `guppy`), but the v2 resolver requires it to be `proc-macro`.


sunshowers:
This is a really weird bug though. I want to file it upstream.

PR #4218, opened 2020-06-03
[x] allow running "cargo fix" through x

huitseeker:
No provisions for particular arguments.


rexhoffman:
Seems it's busted currently:
failure

davidiw:
@rexhoffman, @sunshowers, @metajack, @bmwill 

Do you want this feature?

davidiw:
It'd be nice if x had tests to ensure the functions worked...

huitseeker:
There are a number of other missing functionalities, such as `build`, too.

rexhoffman:
@davidiw @huitseeker looking at this now, sorry that we've let it fall through the cracks.

rexhoffman:
Hmmm this time around we're failing on what I suspect is not forcing the use of the fuzzing features with the use of all features:

Currently the first compilation target runs this:

"cargo" "fix" "--workspace" "--exclude" "testsuite" "--exclude" "libra-crypto" "--exclude" "libra-node"

And gets this error:

    Checking db-bootstrapper v0.1.0 (/Users/rexhoffman/git/libra/execution/db-bootstrapper)
    Checking libra-storage-inspector v0.1.0 (/Users/rexhoffman/git/libra/storage/inspector)
error[E0432]: unresolved import `storage_client`
  --> storage/storage-service/src/storage_service_test.rs:15:5
   |
15 | use storage_client::StorageClient;
   |     ^^^^^^^^^^^^^^ use of undeclared type or module `storage_client`

    Checking libra-management v0.1.0 (/Users/rexhoffman/git/libra/config/management)
    Checking cluster-test v0.1.0 (/Users/rexhoffman/git/libra/testsuite/cluster-test)
    Checking executor-benchmark v0.1.0 (/Users/rexhoffman/git/libra/execution/executor-benchmark)
error: aborting due to previous error

For more information about this error, try `rustc --explain E0432`.

We need to be running this:

"cargo" "fix" "--workspace"  "--all-features" "--exclude" "testsuite" "--exclude" "libra-crypto" "--exclude" "libra-node"

Should probably try to address feature sets in some way shape or form soon, despite upstream cargo bugs.

@sunshowers can we carve out some time to go over that upstream bug and discuss if we have work arounds for at least a subset of problems?   At least the fuzzing feature?

huitseeker:
@rexhoffman Thanks for the hint!

rexhoffman:
I ran this locally and it looks pretty good.   I suspect we shouldn't be using --all-features everywhere, like I observered when running this locally, but that can be debt for another day, especially if this meets your needs today.

devtools/x/src/fix.rs
```diff
@@ -0,0 +1,27 @@
...
+    #[structopt(name = "ARGS", parse(from_os_str), last = true)]
+    args: Vec<OsString>,
+}
+
+pub fn run(args: Args, xctx: XContext) -> Result<()> {
+    let mut pass_through_args = vec![];
+    pass_through_args.extend(args.args);
+
+    let with_all_targets = check::Args {
+        all_targets: true,
```
rexhoffman, line 21: this is fine for now, but if we wanted to mirror test's behavior it'd likely be a bit safer, anyway, it appears to work well now.

PR #4219, opened 2020-06-03
[cleanup][common][crypto][storage][...] Remove 18 more untested, unused methods

huitseeker:
This PR removes 18 untested, unused methods from various crates (see below). The value of the PR may reside more in the signaling of the nature of those untested, unused methods rather than actual removal.

### DIFF
--- a/common/metrics/src/lib.rs
+++ b/common/metrics/src/lib.rs
@@ -25,15 +25,10 @@ pub use prometheus::{
 
 use anyhow::Result;
 use libra_logger::prelude::*;
-use prometheus::{
-    core::{Collector, Metric},
-    proto::MetricType,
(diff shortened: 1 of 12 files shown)


andll:
I would prefer to keep methods in `common/logger/src/struct_log.rs` - I don't thin having them introduce any problems, and we might need them soon when/if people start more actively switching to structured logging

huitseeker:
@andll OK! What about `common/metrics`?

andll:
I don't have context on counter methods. We can ask @sherry-x too, but if it will turn out that no one knows as well it's probably safe to remove

sherry-x:
I think it should be fine for `common/metrics` cause all the things removed here are some legacy metrics stuff.

PR #4223, opened 2020-06-03
[move core modules] add initial_balance field to child VASP creation …

sblackshear:
…script

This removes a step from child VASP account setup. The creator of a child VASP can specify an `initial_balance` that is transferred from the parent to the child.

## Motivation

Simplifying child VASP account setup

## Test Plan

New E2E test for the script


language/stdlib/transaction_scripts/create_child_vasp_account.move
```diff
@@ -1,16 +1,26 @@
 script {
 use 0x0::LibraAccount;
+
+/// Create a `ChildVASP` account for sender `parent_vasp` at `child_address` with a balance of
+/// `CoinType` and an initial authentication_key `auth_key_prefix | child_address`.
+/// If `add_all_currencies` is true,
+/// If `child_initial_balance` is nonzero,
```
tzakian, line 7: Comment fragments?

sblackshear: Good call, will fix!

PR #4224, opened 2020-06-03
[cluster-test] Improve logging during k8s resource deletion

ankushagarwal:
## Summary

* During retries, we were printing logs which shouldn't've been at the ERROR level as they can be false indication that there is a problem with cluster test
* Handle `resource_api.delete` error logging more carefully - don't log if the resource has already been deleted

## Test Plan

Ran cti

### DIFF
--- a/common/retrier/src/lib.rs
+++ b/common/retrier/src/lib.rs
@@ -3,7 +3,7 @@
 
 #![forbid(unsafe_code)]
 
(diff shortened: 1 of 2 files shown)


rexhoffman:
thanks for the heads up, will have to see what happens when tests fail - but looks like this will consolidate the error messages.

PR #4226, opened 2020-06-03
[deps] use cargo-edit's cargo upgrade command to cleanup project tomls

rexhoffman:
## Motivation

Our Cargo.toml files are inconsistent in how they define dependency versions.   Some use equal values to pin, others define partial versions (which cargo expands), and many define versions less than our lock file versions -- that they are never tested with.

Use cargo-edit's  `cargo upgrade --workspace --to-lockfile` to clean up our tomls.

y

## Test Plan

ci

## Related PRs

none

### DIFF
--- a/common/lcs/Cargo.toml
+++ b/common/lcs/Cargo.toml
@@ -11,11 +11,11 @@ edition = "2018"
 
 [dependencies]
 libra-workspace-hack = { path = "../workspace-hack", version = "0.1.0" }
-thiserror = "1.0"
-serde = { version = "1.0", features = ["derive"] }
+thiserror = "1.0.19"
+serde = { version = "1.0.111", features = ["derive"] }
 
 [dev-dependencies]
-criterion = "=0.3.2"
+criterion = "0.3.2"
 proptest = "0.10.0"
 proptest-derive = "0.2.0"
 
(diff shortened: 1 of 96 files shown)


common/lcs/Cargo.toml
```diff
@@ -11,11 +11,11 @@ edition = "2018"
 
 [dependencies]
 libra-workspace-hack = { path = "../workspace-hack", version = "0.1.0" }
-thiserror = "1.0"
-serde = { version = "1.0", features = ["derive"] }
+thiserror = "1.0.19"
+serde = { version = "1.0.111", features = ["derive"] }
 
 [dev-dependencies]
-criterion = "=0.3.2"
```
rexhoffman, line 18: getting rid of things like this.

network/Cargo.toml
```diff
@@ -44,7 +44,7 @@ stream-ratelimiter = { path = "../common/stream-ratelimiter", version = "0.1.0"
 rand_core = { version = "0.5.1", optional = true }
 
 [dev-dependencies]
-criterion = "=0.3.2"
```
rexhoffman, line 47: also here.

rexhoffman:
After this goes in (if we want it).    Only outdated crates we will have referenced anywhere will be codespan.

PR #3441, opened 2020-04-15
[crypto] renaming NamespaceKeys -> namespace::Keys

mimoo:
This PR applies four changes all related to the same issue: naming in crypto.

1. Use namespaces to differentiate different algorithm keys (`ed25519::Ed25519PublicKey` becomes `ed25519::PublicKey` (see rational below).
2. Prefix specific crypto traits with `T` to avoid name collisions (because people like to import `traits::*` and we can't do `use traits::* as _` in rust yet.
3. Rename private/public key into signing/verifying key (this gives a good indication as to what type of keys we're handling).
4. Rename multi ed25519 signing/verifying "key" into signing/verifying "keys".

The biggest change is obviously the use of namespaces to indicate what types of keys we are using.
The rational is that:

* from the point of view of the ed25519 library, a key is just a key
* from the point of view of the caller, rust provides namespace just for this

PR #4229, opened 2020-06-03
[crypto] x25519::PrivateKey -> x25519::X25519PrivateKey

mimoo:
Currently other types of keys in crypto/ are fully named.
For example ed25519::Ed25519PrivateKey.

Let's keep it consistent.

### DIFF
--- a/config/management/src/smoke_test.rs
+++ b/config/management/src/smoke_test.rs
@@ -84,7 +84,7 @@ fn smoke_test() {
             .export_private_key(libra_global_constants::VALIDATOR_NETWORK_KEY)
             .unwrap();
         let identity_key =
-            x25519::PrivateKey::from_ed25519_private_bytes(&identity_key.to_bytes()).unwrap();
+            x25519::X25519PrivateKey::from_ed25519_private_bytes(&identity_key.to_bytes()).unwrap();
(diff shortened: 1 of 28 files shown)


matbd:
Nice, thanks!

If anyone pushes back, as far as the YAML file is concerned, note that `#[serde(rename="X25519PrivateKey")]` would do it too.

davidiw:
> Nice, thanks!
> 
> If anyone pushes back, as far as the YAML file is concerned, note that `#[serde(rename="X25519PrivateKey")]` would do it too.

@mimoo, can we not do this to the ed25519 keys instead? and go back in the other direction <3

mimoo:
@matbd I'm familiar with a field rename, but you're talking about a type rename by having this annotation before a type definition?

matbd:
> The approach in #3441 seemed to be simpler. It was deemed "blocked by Mathieu's tool" (what of it?), but I understood this to be just a delay. Why the change of tack?

What did not work technically for me in #3441 is trying to infer Rust namespaces automatically. I can work with explicitly qualified public names (e.g. `#[serde(rename = "qualified::Name")]`) but people did not seem to be very enthusiastic about it when I brought up the idea. (Partly because you end up adding ``#[serde(rename = ".."` to every serializable object)

davidiw:
What's the plan here @mimoo?

PR #4238, opened 2020-06-04
[move core modules] eliminating Transaction::sender in validator and …

sblackshear:
…config code

Getting closer to eliminating `Transaction::sender()`. This gets rid of all the uses in `LibraSystem`, `LibraConfig`,  and `ValidatorConfig`.

## Motivation

Deprecating `Transaction::sender()` and `move_to_sender`

Y

## Test Plan

Refactored existing tests, all pass.


language/stdlib/modules/genesis.move
```diff
@@ -74,7 +72,7 @@ module Genesis {
         LibraAccount::create_treasury_compliance_account<LBR::T>(
             association,
             tc_addr,
-            tc_auth_key_prefix,
+            copy dummy_auth_key_prefix,
```
tzakian, line 75: Not sure I quite follow this change -- I can't see where the key is rotated to the correct key. Now that rotate key takes a `signer` maybe do those all at the bottom of this?

language/tools/vm-genesis/src/lib.rs
```diff
@@ -144,35 +144,38 @@ fn create_and_initialize_main_accounts(
...
     context.set_sender(config_address());
     context.exec(
         "LibraAccount",
         "rotate_authentication_key",
         vec![],
-        vec![Value::vector_u8(genesis_auth_key.to_vec())],
+        vec![
+            Value::transaction_argument_signer_reference(config_address()),
+            Value::vector_u8(genesis_auth_key.to_vec()),
+        ],
```
tzakian, line 159: I think this should be able to be done in the main genesis procedure now that `rotate_authentication_key` takes a `signer`.

language/tools/vm-genesis/src/lib.rs
```diff
@@ -144,35 +144,38 @@ fn create_and_initialize_main_accounts(
...
         "LibraAccount",
         "rotate_authentication_key",
         vec![],
-        vec![Value::vector_u8(genesis_auth_key.to_vec())],
+        vec![
+            Value::transaction_argument_signer_reference(
+                account_config::treasury_compliance_account_address(),
+            ),
+            Value::vector_u8(genesis_auth_key.to_vec()),
+        ],
```
tzakian, line 172: Same here?

language/tools/vm-genesis/src/lib.rs
```diff
@@ -144,35 +144,38 @@ fn create_and_initialize_main_accounts(
...
     );
 
-    // TODO: use signer to eliminate this
     context.set_sender(fee_account_address);
     context.exec(
-        GENESIS_MODULE_NAME,
-        "initialize_txn_fee_account",
+        "LibraAccount",
+        "rotate_authentication_key",
```
tzakian, line 178: ditto

language/tools/vm-genesis/src/lib.rs
sblackshear: good call!

language/stdlib/modules/genesis.move
sblackshear: Ah, I probably should have made this change somewhere else. The old code was wrong (but harmlessly). The problem is that (1) all the accounts we create in genesis get rotated to the genesis auth key and (2) even if we wanted a different auth key for the TC account, passing `tc_auth_key_prefix` into the account creation wouldn't work. The problem is  that even if we get to choose `tc_auth_key_prefix`, we won't know the private key that can sign for `auth_key_prefix | 0xb1e55ed`....

PR #4239, opened 2020-06-04
[move-prover] added informal specs for fixedpoint32.move and lbr.move

junkil-park:
- added the informal specs for both modules FixedPoint32 and LBR

## Motivation

To specify the standard library

Yes

## Test Plan

cargo test

### DIFF
--- a/language/stdlib/modules/FixedPoint32.move
+++ b/language/stdlib/modules/FixedPoint32.move
@@ -74,6 +74,95 @@ module FixedPoint32 {
     public fun get_raw_value(num: T): u64 {
         num.value
     }
+
+    // **************** SPECIFICATIONS ****************
+
+    /*
+    This module defines the fixed-point numeric type, the constructors and the operations (mul, div) for that type.
+    It is unlike that this module would have any module invariant (or global property) because it does not declare
+    any resource type whose value can be stored and retained in the global state. However, it is possible that
+    FixedPoint32::T is used in the invariants of other modules.
+
+    An apparent local property to check is that the fixed-point operations (including the constructors) are implemented
+    correctly. However, what this rigorously means is not obvious because it requires the mathematical model of
+    fixed-point operations to check the implementations against. Mathematically, a finite-precision arithmetic operation
+    (like the fixed-point one's) can be defined using the notion of "exact rounding". This means that such an operation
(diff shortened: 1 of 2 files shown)


language/stdlib/modules/fixedpoint32.move
```diff
@@ -74,6 +74,95 @@ module FixedPoint32 {
...
+    * Let FP0 refer to u64 which can be seen as the 64-bit unsigned fixed-point numeric type with 0 factional bits.
+    * round_FP32: R -> FP32 is a function that takes a real number as an input and returns the largest FP32 number which
+      is less than or equal to the input.
+    * round_FP0: R -> FP0 is a function that takes a real number as an input and returns the largest FP0 number which is
+      less than or equal to the input.
+    * Let mul_R and div_R be the multiplication and the division function over real numbers respectively.
+    */
+
+    spec fun multiply_u64 {
+        // aborts_if mul_R(num, multiplier) >= max_u64() + 1; // overflow (note that max_u64() == max_FP0())
```
DavidLDill, line 104: This is a nitpick, but wouldn't > max_u64() be more clear than >= max_u64() + 1?

language/stdlib/modules/fixedpoint32.move
```diff
@@ -74,6 +74,95 @@ module FixedPoint32 {
...
+        // aborts_if 0 < mul_R(num, multiplier) < 1; // underflow (note that 1 is the smallest non-zero FP0);
```
DavidLDill, line 105: Shouldn't the result just be 0 in this case?

language/stdlib/modules/fixedpoint32.move
junkil-park: They are subtly different. (max_u64()+0.1) can be rounded down to max_u64(), but the integer that (max_u64()+1.1) is rounded down to is not u64. So, the latter is definitely an overflow.

language/stdlib/modules/fixedpoint32.move
junkil-park: This is a condition for underflow in general. We may need to abort if it's not a big deal for us. However, it seems that the other function called `create_from_rational` actually cares about the underflow case.

PR #4242, opened 2020-06-04
Refactor config builder

JoshLind:
## Motivation

At present, config builder is a little confusing and inconsistent; it contains a lot of unnecessary code complexity which makes it difficult to reason about with respect to the actual generated configs. This PR takes some steps to simplifying config builder. It does so in 6 commits:

- Commits 1 - 3 refactor key manager config, validator config and full node config, respectively. Here, we avoid using setter methods() where appropriate and simply make the internal fields public. This not only removes a lot of code, but it makes future refactors easier and helps to avoid potential errors: for example, full node config currently has two methods (public() and enable_remote_authentication()) which both modify the same internal field "enable_remote_authentication".. and one of the methods appears to be unused in the code base? Moreover, as public() is undocumented (and arguably, not well named) it's unclear what this method should actually do..

- Commit 4 renames some of the fields in validator config builder: (i) to provide more self clarity; and (ii) to be consistent with the names in validator config. This helps avoid confusion about what config builder is doing.

- Commit 5 renames some of the fields in full node config builder to be consistent with validator config and the rest of the config code.

Note: in commits 4 and 5, this PR does not change the fields/arguments exposed by config-builder/main.rs. This is because changes to the names of these arguments may break external dependencies (e.g., docker containers). If we think it's worth fixing these arguments, this can be done in a future PR.

Yes.

## Test Plan

All existing tests still pass.

## Related PRs

None.

### DIFF
--- a/config/config-builder/src/full_node_config.rs
+++ b/config/config-builder/src/full_node_config.rs
@@ -61,7 +61,7 @@ impl FullNodeConfig {
     }
 
     pub fn validators(&mut self, nodes: usize) -> &mut Self {
-        self.validator_config.nodes = nodes;
+        self.validator_config.num_nodes = nodes;
         self
     }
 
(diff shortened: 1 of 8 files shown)


JoshLind:

PR #4243, opened 2020-06-04
[config] support waypoint from storage

davidiw:
Per title, this is the last PR in this series before we make config
depend on secure storage and clean up this fun little mess.

### DIFF
--- a/secure/storage/src/config.rs
+++ b/secure/storage/src/config.rs
@@ -2,10 +2,10 @@
 // SPDX-License-Identifier: Apache-2.0
 
 use crate::Storage;
-use libra_config::config::{Identity, NetworkConfig};
+use libra_config::config::{Identity, NetworkConfig, WaypointConfig};
 use libra_crypto::x25519;
-use libra_types::PeerId;
-use std::convert::TryInto;
+use libra_types::{waypoint::Waypoint, PeerId};
+use std::{convert::TryInto, str::FromStr};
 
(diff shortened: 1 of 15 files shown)


secure/storage/src/config.rs
```diff
@@ -46,3 +41,21 @@ pub fn peer_id(config: &NetworkConfig) -> PeerId {
...
+                .get(libra_global_constants::WAYPOINT)
+                .expect("Unable to read waypoint")
+                .value
+                .string()
+                .expect("Expected string for waypoint");
+            Some(Waypoint::from_str(&waypoint).expect("Unable to parse waypoint"))
+        }
+        WaypointConfig::None => None,
+    };
+    waypoint.expect("peer id should be present")
```
mgorven, line 60: s/peer id/waypoint/ here?

davidiw: copypasta :D
good call!

state-synchronizer/src/synchronizer.rs
```diff
@@ -60,7 +61,7 @@ impl StateSynchronizer {
             network,
             state_sync_to_mempool_sender,
             config.base.role,
-            config.base.waypoint,
+            Some(waypoint),
```
zekun000, line 64: thanks for fixing it, why not fix all the way?

davidiw: not the right PR for that :D

PR #4245, opened 2020-06-04
[management] Do not enable any discovery for validators

davidiw:
@agclement made it so that we have a dead simple onchain discovery protocol, this ensures that we use it and only it to avoid races.

@mgorven, what are you specifying right now in the network configs for discovery methd?

### DIFF
--- a/config/management/src/smoke_test.rs
+++ b/config/management/src/smoke_test.rs
@@ -70,7 +70,7 @@ fn smoke_test() {
         let mut config = NodeConfig::default();
 
         let mut network = NetworkConfig::default();
-        network.discovery_method = DiscoveryMethod::Onchain;
+        network.discovery_method = DiscoveryMethod::None;
         config.validator_network = Some(network);
 
         let mut network = NetworkConfig::default();


mgorven:
> @mgorven, what are you specifying right now in the network configs for discovery methd?

Nothing at the moment: https://github.com/libra/partners/blob/master/helm/libra-validator/files/configs/validator.toml

davidiw:
> > @mgorven, what are you specifying right now in the network configs for discovery methd?
> 
> Nothing at the moment: https://github.com/libra/partners/blob/master/helm/libra-validator/files/configs/validator.toml

Cool so default which is None.. then you've already been testing this :)

zekun000:
this is confusing, with None we use the simple discovery?

davidiw:
> this is confusing, with None we use the simple discovery?

Yes, in general, we should have onchain reading / discoverability on by default for validators. This is different than the networking service that gossips among other peers which must be explicitly turned on.

This is akin to consensus, by default, automatically receiving onchain updates.

PR #4246, opened 2020-06-04
[OnchainDiscovery] incremental change to make OnchainDiscovery setup …

agc-quadra:
…look more similar to other network using components

## Motivation

This is a collection of refactorings around setting up OnchainDiscovery to simplify main_node and bring more clarity to the building process.

1.  Move setup_onchain_discovery out of main_node.rs and into the OnchainDiscovery crate.
2.  Make the construction of the OnChainDiscoverySender match the general NetworkSender pattern.
3.  Create a onchain_discovery_builder to encapsulate the logic of setup_onchain_discovery

The second change effectively moves maintenance of the connecion_mgr_tx from the OnChaindiscoverySender struct to the client/OnchainDiscovery struct.  This specific change benefits follow up work on clarifying main_node, network_builder, and the overall building process.

yes
## Test Plan

don't break existing integration/unit tests

## Related PRs

### DIFF
--- a/libra-node/src/main_node.rs
+++ b/libra-node/src/main_node.rs
@@ -24,6 +24,7 @@ use network::validator_network::network_builder::{AuthenticationMode, NetworkBui
 use network_simple_onchain_discovery::{
     gen_simple_discovery_reconfig_subscription, ConfigurationChangeListener,
 };
+use onchain_discovery::builder::OnchainDiscoveryBuilder;
 use state_synchronizer::StateSynchronizer;
 use std::{boxed::Box, collections::HashMap, net::ToSocketAddrs, sync::Arc, thread, time::Instant};
 use storage_interface::{DbReader, DbReaderWriter};
@@ -133,14 +134,21 @@ pub fn setup_network(
                 .add_gossip_discovery();
         }
         DiscoveryMethod::Onchain => {
-            onchain_discovery::setup_onchain_discovery(
-                &mut network_builder,
+            let (network_tx, discovery_events) =
+                onchain_discovery::network_interface::add_to_network(&mut network_builder);
+            let onchain_discovery_builder = OnchainDiscoveryBuilder::build(
+                network_builder
+                    .conn_mgr_reqs_tx()
+                    .expect("ConnectivityManager must be installed"),
+                network_tx,
+                discovery_events,
                 peer_id,
                 role,
                 libra_db,
                 waypoint,
                 runtime.handle(),
             );
+            onchain_discovery_builder.start(runtime.handle());
         }
         DiscoveryMethod::None => {}
     }
--- a/libra-node/src/main_node.rs
+++ b/libra-node/src/main_node.rs
@@ -80,10 +80,14 @@ pub fn setup_onchain_discovery(
     waypoint: Waypoint,
     executor: &Handle,
 ) {
-    let (network_tx, peer_mgr_notifs_rx, conn_notifs_rx) =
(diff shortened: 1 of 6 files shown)


agc-quadra:
@davidiw  FYI

agc-quadra:
@davidiw   please take a look

network/onchain-discovery/src/client.rs
```diff
@@ -298,6 +304,13 @@ where
         };
     }
 
+    async fn send_connectivity_request(
+        &mut self,
+        req: ConnectivityRequest,
+    ) -> Result<(), mpsc::SendError> {
+        self.conn_mgr_reqs_tx.send(req).await
+    }
+
```
davidiw, line 313: why not put this inline?

agc-quadra:
> It isn't clear to me what's going on here, we already have access to a network_tx with the ability to send a connectivity request, perhaps we should wait for your other commits to see what's going on here?

This PR primarily serves to move 

> It isn't clear to me what's going on here, we already have access to a network_tx with the ability to send a connectivity request, perhaps we should wait for your other commits to see what's going on here?

Short version is that this PR pulls the connection to the connectivitymanager out of onchaindiscoverysender and into onchaindiscovery (client).   This allows the other beast to be marginally smaller and more manageable.

agc-quadra:
Expanded the scope of the PR to be a bit meatier.  This represents a manageable standalone chunk of a much larger piece of ongoing local work to simplify network_builder.

agc-quadra:
@davidiw this PR can be viewed in the context of #4300;  it collects the set of preliminary changes to onchain-discovery that can be made in isolation from other crates.   The first commit in the PR is the key for unblocking the next stage of the work;  that stage has a very high fan out so getting this set out of the way will keep the next chunks manageable.

network/onchain-discovery/src/builder.rs
```diff
@@ -0,0 +1,83 @@
...
+        Self {
+            onchain_discovery_service,
+            onchain_discovery,
+            started: false,
+        }
+    }
+
+    /// Starts the provided onchain_discovery_service and onchain_discovery.
+    pub fn start(self, executor: &Handle) {
+        assert!(!self.started);
```
davidiw, line 79: start is never set

libra-node/src/main_node.rs
```diff
@@ -184,14 +134,21 @@ pub fn setup_network(
                 .add_gossip_discovery();
         }
         DiscoveryMethod::Onchain => {
-            setup_onchain_discovery(
-                &mut network_builder,
+            let (network_tx, discovery_events) =
+                onchain_discovery::network_interface::add_to_network(&mut network_builder);
+            let onchain_discovery_builder = OnchainDiscoveryBuilder::build(
```
davidiw, line 139: would it make sense to just pass network builder in here? seems like we're just moving around data types...

agc-quadra: Yes, we are 'just' moving around data types.  

The reason that setup_network is a part of main_node right now is explictily because we are passing network_builder into every component as a parameter.   This creates a dependency loop where setup_network -> XXX_component -> Network_builder.  Which blocks setup_network from being part of network_builder.

This is one of the points to look into #4300 for context;  specifically look for

1)  the updated versions of "add_protocol_handler" and the corresponding "get_endpoing_configuration" calls that show up in every component.  
2). the move of network_builder to a separate crate inside networking, invoked *only* by node_builder

agc-quadra:
Merged as part of #4299

PR #4248, opened 2020-06-04
[Move] create preburn resource under DD during DD account creation

moeziniaf:
## Motivation
important files: Libra.move

-DD now has preburn area created upon account creation
-add nonce to create_designated_dealer txn


language/stdlib/modules/libra.move
```diff
@@ -262,6 +262,40 @@ module Libra {
...
 
+    ///////////////////////////////////////////////////////////////////////////
+    // Treasury Compliant specific methods for DDs
+    ///////////////////////////////////////////////////////////////////////////
+
+    // Publish `preburn` under 'account'. Used for bootstrapping designated dealer
+    // as assocation TC account is creating this resource for DD
+    public fun publish_preburn_to_account<Token>(creator: &signer, account: &signer) {
+        assert_account_is_blessed(creator);
+        let preburn = Preburn<Token> { requests: Vector::empty(), is_approved: true };
```
sblackshear, line 273: Unrelated, but I think we can kill the `is_approved` bool now that this is the only way to create a `Preburn` resource...

language/stdlib/modules/libra.move
```diff
@@ -262,6 +262,40 @@ module Libra {
...
+        assert_account_is_blessed(tc_account);
+        let preburn = borrow_global_mut<Preburn<CoinType>>(dd_address);
+        preburn_with_resource(
+            coin,
+            preburn,
+            dd_address,
+        );
+    }
+
+    fun assert_account_is_blessed(sender_account: &signer) {
```
sblackshear, line 292: This feels like it belongs in the `Association` module

language/stdlib/modules/libra.move
```diff
@@ -262,6 +262,40 @@ module Libra {
...
+        assert_account_is_blessed(creator);
+        let preburn = Preburn<Token> { requests: Vector::empty(), is_approved: true };
+        move_to(account, preburn)
+    }
+
+
+    /// Send coin to the preburn holding area for `dd_address`, where it will wait to either be burned
+    /// or returned to the balance of `account`.
+    /// Fails if `dd_address` does not have a published DesignatedDealerPreburn resource
+    fun preburn_coin<CoinType>(tc_account: &signer, dd_address: address, coin: T<CoinType>)
```
sblackshear, line 281: I'm confused about this --isn't it the DD's job to preburn and the TC's job to approve the burn or cancel?

language/stdlib/modules/libra.move
moeziniaf: OK, i think i'll do this in a follow up PR

language/stdlib/modules/LibraAccount.move
```diff
@@ -679,21 +679,32 @@ module LibraAccount {
...
     }
 
+    /// Create Preburn resource under account 'address' - should be called
+    /// in tandem with 'create_designated_dealer' during DD creation
+    public fun create_preburn_for_address<CoinType>(
+        blessed: &signer,
+        address: address,
+    ) {
+        let account_signer = create_signer(address);
+        Libra::publish_preburn_to_account<CoinType>(blessed, &account_signer);
```
sblackshear, line 698: I think we can simplify things a bit by just adding this line inside `create_dd_account`. This will make sure that  the only way to get `Preburn` resource is inside a DD account at the time it is created.

language/stdlib/modules/libra.move
moeziniaf: yes, i was wrong. Changed now

language/stdlib/modules/Libra.move
```diff
@@ -262,6 +262,20 @@ module Libra {
...
 
+    ///////////////////////////////////////////////////////////////////////////
+    // Treasury Compliant specific methods for DDs
+    ///////////////////////////////////////////////////////////////////////////
+
+    // Publish `preburn` under 'account'. Used for bootstrapping designated dealer
+    // as assocation TC account is creating this resource for DD
+    public fun publish_preburn_to_account<Token>(creator: &signer, account: &signer) {
+        Association::assert_account_is_blessed(creator);
+        let preburn = Preburn<Token> { requests: Vector::empty(), is_approved: true };
```
tzakian, line 273: Now that we have `move_to` and this awesome logic, do we need the `is_approved` flag in the preburn resource anymore?

language/stdlib/modules/LibraAccount.move
```diff
@@ -659,32 +659,34 @@ module LibraAccount {
...
         blessed: &signer,
         new_account_address: address,
         auth_key_prefix: vector<u8>,
     ) {
-        DesignatedDealer::assert_account_is_blessed(blessed);
+        Association::assert_account_is_blessed(blessed);
         Transaction::assert(!Libra::is_synthetic_currency<CoinType>(), 202);
         let new_dd_account = create_signer(new_account_address);
+        Event::publish_generator(&new_dd_account);
+        Libra::publish_preburn_to_account<CoinType>(blessed, &new_dd_account);
```
tzakian, line 687: As a clarification, is there a distinct designated dealer account for each currency type? Or could a given designated dealer address hold multiple preburn resources (one for each currency).

moeziniaf: There will be a singular DD per currency. For MVP we don't need to link them together (alla Parent/child VASP)

language/stdlib/modules/Libra.move
moeziniaf: I was going to deprecate it in a followup PR since there are a whole bunch of uses of `new_preburn` is used in a whole bunch of places

language/stdlib/modules/Libra.move
```diff
@@ -262,6 +262,20 @@ module Libra {
         };
     }
 
+    ///////////////////////////////////////////////////////////////////////////
+    // Treasury Compliant specific methods for DDs
```
sblackshear, line 266:
```suggestion
    // Treasury Compliance specific methods for DDs
```

moeziniaf:
> LGTM, very good stuff!

thanks for your help!

language/stdlib/modules/LibraAccount.move
tzakian: Awesome!

PR #3596, opened 2020-04-24
[language][vm] non-recursive bounds checker implementation for SignatureToken

vgao1996:
## Summary
Reimplement the bounds checker for SignatureToken in a non-recursive fashion to avoid overflowing the stack.

## Test Plan
cargo test

## Dependencies
#3572

PR #3572, opened 2020-04-23
[language][vm] non-recursive serializer/deserializer implementation for SignatureToken

vgao1996:
## Summary
`SignatureToken` can be deeply nested, causing the deserializer to overflow its stack. This implements a non-recursive deserializer for `SignatureToken`, using `Vec` for memory management, simulating the program stack.

## Test Plan
cargo test

**Note: Still a WIP. Implementation is complete & tests are passing, but still need to check in the fuzzer artifact and see if it can handle that.**

PR #4250, opened 2020-06-04
[LCS] Use serde_stacker to avoid stack blowups

### LINKED ISSUE #3046: [Bug] Recursion & Stack Overflow in the VM
We need to review recursions in our VM code to ensure that they cannot cause stack overflow under any circumstances. This means either no stack overflow for any given inputs, or explicitly state that certain inputs that could cause stack overflow are prohibited by some internal invariants.

We also need to think about how gas comes to play with this.

Known Cases:
- [ ] During deserialization, a large number of `SignatureToken::Reference(Reference(Reference(...` can cause a stack overflow.

huitseeker:
~~DO NOT MERGE; This is a PoC pending resolution of an [upstream PR](https://github.com/dtolnay/serde-stacker/pull/7).~~

This uses a stack-growth library ([stacker](https://github.com/rust-lang/stacker)) behind a [serde adapter](https://github.com/dtolnay/serde-stacker) to help with stack size in (de)serialization.

The perf cost seems manageable (< 10%):
https://gist.github.com/huitseeker/6cd90ee29d90b7ce0c0cb00f5a263fe1#file-gistfile1-txt-L1-L4

Fixes #3046.

The present PR is offered as a PoC, ~~and uses my personal repo to enact a [small fix on `serde-stacker`](https://github.com/dtolnay/serde-stacker/pull/7). It will be modified to point to the normal crate once a new release of `serde-stacker` comes out.~~

~~This is still WIP, but tests will pass once with https://github.com/dtolnay/serde-stacker/pull/8 merged upstream (they rely on branching on `is_human_readable` in the (de)serialization of `AccountAddress`).~~

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2264,6 +2264,7 @@ dependencies = [
  "proptest 0.10.0 (registry+https://github.com/rust-lang/crates.io-index)",
  "proptest-derive 0.2.0 (registry+https://github.com/rust-lang/crates.io-index)",
  "serde 1.0.112 (registry+https://github.com/rust-lang/crates.io-index)",
+ "serde_stacker 0.1.3 (registry+https://github.com/rust-lang/crates.io-index)",
  "thiserror 1.0.19 (registry+https://github.com/rust-lang/crates.io-index)",
 ]
 
@@ -4000,6 +4001,14 @@ dependencies = [
 ]
 
 [[package]]
+name = "psm"
+version = "0.1.8"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+dependencies = [
+ "cc 1.0.54 (registry+https://github.com/rust-lang/crates.io-index)",
+]
+
+[[package]]
 name = "qstring"
 version = "0.7.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -4814,6 +4823,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "serde_stacker"
+version = "0.1.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+dependencies = [
+ "serde 1.0.112 (registry+https://github.com/rust-lang/crates.io-index)",
+ "stacker 0.1.9 (registry+https://github.com/rust-lang/crates.io-index)",
+]
+
+[[package]]
 name = "serde_urlencoded"
 version = "0.6.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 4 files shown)


mimoo:
cc @vgao1996 and @bmwill who have had to look at similar issues

mimoo:
on serde stacker:

> Be aware that you may need to protect against other recursive operations outside of serialization and deserialization when working with deeply nested data, including, but not limited to, Display and Debug and Drop impls.

vgao1996:
Sorry for the late reply. Finally had some time to think about this. Based on my own experience writing #3572 and #3596, and a failed attempt to make `Drop::drop` non-recursive for `SignatureToken`, I'd say that
- Serialization/Deserialization is just the tip of the iceberg. The real issue here is that any operation on such recursive structures can blow up the stack.
  - All such operations need to be written in a careful manner or otherwise the system is still unsafe. 
  - `Drop::drop` is the real monster, which is tightly coupled with the move semantics and borrow rules of the language, and follow special rules that render a few common fixes impossible.
- serde_stacker does offer an interesting solution that can be adapted to rewrite all recursive operations: dynamically growing the stack via stacker. This can be another viable option in addition to emulating the stack using a Vec. However I'm not sure if it's dangerous or if we feel comfortable having code that manipulates the stack limit in our TCB. @dariorussi what do you think?
- Ultimately I think the most effective way to deal with such data structures is to stop early during or even before deserialization. I think we should terminate once we reach a certain depth, or check before hand if the number of bytes is within limit. (I do realize the former may not be easy if we use serde's implementation.)

huitseeker:
- the intent of this PR wrapping `serde_stacker` around LCS is to armor LCS against stack blowups in general, including serialization, and without limitation to `Drop` or #3046,
- on comprehensiveness, the only really safe implementation of a deeply-nested drop (or deeply-nested serialization) I can think of is manual T.C.O. w/a trampoline, at significant cost in performance,
- it's possible to limit the recursion depth in serde, either by contributing a solution to https://github.com/dtolnay/serde-stacker/issues/1 or by implementing it directly in LCS,

All of the things above are three complementary approaches to get a better handle on stack growth in serialization / drop.

With a somewhat higher priority¹, we'll *also* need to decide if we want to keep serialization as infaillible as it is now (aking to `fn to_bytes(&self) -> Vec<u8>`) or if we want to explicitly make it faillible (homologous to `fn to_bytes(&self) -> Result<Vec<u8>, lcs::Error>`).

I think capping recursion depth implies spending time on the later question, which becomes even more important.

¹: we're on the way to bringing serialization on the critical path for signing with https://github.com/libra/libra/pull/4073 and we'll be using it on user-provided inputs when Move contracts can be provided by 3d parties

dtolnay:
I chatted with @vgao1996 off thread regarding:

> a failed attempt to make `Drop::drop` non-recursive for `SignatureToken`

Consider looking at how https://github.com/dtolnay/miniserde implements non-recursive drop on https://docs.rs/miniserde/0.1.12/miniserde/json/enum.Value.html because it would work for SignatureToken.

davidiw:
seems like we won't be going this direction, feel free to re-open if this is a misunderstanding.

PR #4251, opened 2020-06-04
[language] fix some bugs in bytestring literals and use them in the stdlib

bob-wilson:
There were a few character strings (actually vector<u8>) in the stdlib that were described as having UTF-8 encodings. However, these are all limited to ASCII characters, so we can use the new bytestring literal syntax to specify the values. We want to avoid getting into Unicode complexities, at least for now.

That change exposed a problem with comment markers inside strings. The addition of byte string literals to Move means that we can have comment-starting characters that occur inside a string, e.g., "//". The code to identify and strip comments needs to understand the string syntax so that it can avoid treating those characters as comments when they are used inside strings.

Finally, while fixing that comment problem, I happened to notice an issue with escaped backslashes before the closing quote of a string. The previous code looked for a closing quote that was not immediately preceded by a backslash. But, that doesn't handle the case where there is an escaped backslash before the quote.

## Motivation

Fix bugs and make the stdlib code more readable

Yes

## Test Plan

Added tests for the bytestring bugs and ran them

### DIFF
--- a/language/stdlib/modules/lbr.move
+++ b/language/stdlib/modules/lbr.move
@@ -44,7 +44,7 @@ module LBR {
             true,    // is_synthetic
             1000000, // scaling_factor = 10^6
             1000,    // fractional_part = 10^3
-            x"4C4252" // UTF8-encoded "LBR" as a hex string
+            b"LBR"
         );
         let preburn_cap = Libra::new_preburn_with_capability(&burn_cap);
         let coin1 = ReserveComponent<Coin1::T> {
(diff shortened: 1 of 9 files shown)


language/stdlib/modules/lbr.move
```diff
@@ -44,7 +44,7 @@ module LBR {
             true,    // is_synthetic
             1000000, // scaling_factor = 10^6
             1000,    // fractional_part = 10^3
-            x"4C4252" // UTF8-encoded "LBR" as a hex string
+            b"LBR"
```
tnowacki, line 47: Beautiful!

tnowacki: @tzakian @sblackshear 
Any other spots to clean up?

bob-wilson: I searched for all the hex string literals but I suppose I might have missed something.

tnowacki: No idea, just figured I'd ask since @tzakian had talked about having to go onto a website to get the encoding

Issue #4066, opened 2020-05-26
Add --duration to --health-check


andll:
Cluster test can be run with --health-check in normal mode (see cluster_test/main.rs:run_health_check). With this command cluster test tails event stream of a node and runs the health check

Currently this runs health check in an infinite loop. Instead of this, we want use `--duration` flag that will specify for how many seconds health check should be run (the flag is already there, we just need to reuse it).

In this mode, we also want exit code to indicate whether last healthcheck was healthy - 0 when it was healthy, non-0 when it was not.

To verify how this works - run `./scripts/cti --pr <your PR> --health-check --duration 10` - make sure this runs for 10 seconds after setup is done. 

Note - do not run infinite loop version with `cti` - this will consume one of cluster test clusters for a very long time

PR #4253, opened 2020-06-04
[cluster_test] Add --duration to --health-check

changyonghu:
Closes: #4066

### DIFF
--- a/testsuite/cluster-test/src/main.rs
+++ b/testsuite/cluster-test/src/main.rs
@@ -155,7 +155,8 @@ pub fn main() {
     let mut perf_msg = None;
 
     if args.health_check {
-        runner.run_health_check();
+        let duration = Duration::from_secs(args.duration);
+        exit_on_error(runner.run_health_check(duration));
     } else if args.perf_run {
         perf_msg = Some(runner.perf_run());
     } else if args.cleanup {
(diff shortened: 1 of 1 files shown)


testsuite/cluster-test/src/main.rs
```diff
@@ -738,16 +742,24 @@ impl ClusterTestRunner {
...
             let events = self.logs.recv_all_until_deadline(deadline);
-            let _ignore =
-                self.health_check_runner
-                    .run(&events, &HashSet::new(), PrintFailures::All);
+            let result = self
+                .health_check_runner
+                .run(&events, &HashSet::new(), PrintFailures::All);
+            let now = Instant::now();
+            if now > health_check_deadline {
+                break match result {
```
ankushagarwal, line 758: You can simplify this to `break result.map(|_| ())`

ankushagarwal:
@changyonghu : Were you able to test that this works.

LGTM other than minor comment

changyonghu:
> @changyonghu : Were you able to test that this works.
> 
> LGTM other than minor comment
Yes, I am able to manually test. ./scripts/cti --pr 4253 --health-check --duration 10 runs 10 sec after setup.

ankushagarwal:
@changyonghu you can merge using `@bors-libra r=ankushagarwal` command when you're ready

andll:
@bors-libra r-
Looks like bors read the message in quotes, should probably address comment before merging

testsuite/cluster-test/src/main.rs
```diff
@@ -155,7 +155,11 @@ pub fn main() {
     let mut perf_msg = None;
 
     if args.health_check {
-        runner.run_health_check();
+        let duration = Duration::from_secs(args.duration);
+        if runner.run_health_check(duration).is_err() {
```
andll, line 159: You can use `exit_on_error(runner.run_health_check(duration))`

testsuite/cluster-test/src/main.rs
andll: Also I think return here will be more obvious/appropriate then break

testsuite/cluster-test/src/main.rs
```diff
@@ -155,7 +155,11 @@ pub fn main() {
     let mut perf_msg = None;
 
     if args.health_check {
-        runner.run_health_check();
+        let duration = Duration::from_secs(args.duration);
```
snakech, line 158: @andll @ankushagarwal is accuracy of duration unit by secs enough?

andll: Yes

PR #4258, opened 2020-06-04
[consensus] fix notification to mempool

zekun000:
## Motivation

It's off-by-one because of the appended metadata txn, this commit fixes
it and ensures the lengths are equal.

Y

## Test Plan

It doesn't seem easy to test it, just zip_eq in place

## Related PRs

### DIFF
--- a/consensus/src/txn_manager.rs
+++ b/consensus/src/txn_manager.rs
@@ -7,6 +7,7 @@ use consensus_types::{block::Block, common::Payload};
 use debug_interface::prelude::*;
 use executor_types::StateComputeResult;
 use futures::channel::{mpsc, oneshot};
+use itertools::Itertools;
 use libra_mempool::{
     CommittedTransaction, ConsensusRequest, ConsensusResponse, TransactionExclusion,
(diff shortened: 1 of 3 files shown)


consensus/src/txn_manager.rs
```diff
@@ -69,7 +70,11 @@ impl TxnManager for MempoolProxy {
             Some(txns) => txns,
             None => return Ok(()),
         };
-        for (txn, status) in txns.iter().zip(compute_results.compute_status().iter()) {
+        // skip the block metadata txn result
+        for (txn, status) in txns
+            .iter()
+            .zip_eq(compute_results.compute_status().iter().skip(1))
```
davidiw, line 76: zip_eq to be safe?
is there another way to skip the metadata txn via looking at the txn or the outputs? arbitrarily skipping seems ... easy to introduce bugs later...

zekun000: yeah, not right now, unfortunately the compute result doesn't have any info I can link back to the original txn

PR #4259, opened 2020-06-04
[management] improve the verify framework

davidiw:
Print owner and operator accounts
use the operator account from storage rather than computing it from the key

### DIFF
--- a/config/management/src/lib.rs
+++ b/config/management/src/lib.rs
@@ -399,7 +399,8 @@ pub mod tests {
             .unwrap()
             .split("Key not set")
             .count();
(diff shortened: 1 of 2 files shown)


config/management/src/lib.rs
```diff
@@ -408,7 +409,8 @@ pub mod tests {
             .unwrap()
             .split("Key not set")
             .count();
-        assert_eq!(output, 1); // 0 KeyNotSet results in 1 split
+        // 2 KeyNotSet results in 3 split (the accounts aren't initialized via initialize
```
JoshLind, line 412: nit: close the brackets in the comment :P

PR #4284, opened 2020-06-04
rust: bump to 1.44.0

bmwill:


andll:
@bors-libra r-
Looks like there are more lint problems

bmwill:
Let me fix them and I'll re-queue. idk how i missed those

PR #4286, opened 2020-06-04
[fuzzing] do not change behavior with feature="fuzzing"

mimoo:


sunshowers:
From our Slack conversation:

> will the fuzzers keep working?
> or do they need some changes as well
> I would really hope we can architect our code to plug behavior changes in, whether through generics or something else
> e.g. you can have a generic that decides what to do, set it to reject errors by default (Rust generics can have a default), then specify an alternate option that's only available to test code
> that would be a practical way to do a lot of this

mimoo:
@rexhoffman this is so weird, we're getting this error that is completely unrelated

```
thread 'node-2' panicked at 'itertools: .zip_eq() reached end of one iterator before the other', <::std::macros::panic macros>:2:4
```

cc @zekun000

zekun000:
`zip_eq` doesn't cause the unit tests failed, it's
```
failures:
    noise_wrapper::fuzzing::test_noise_fuzzer
    noise_wrapper::stream::test::interleaved_writes
    noise_wrapper::stream::test::simple_test
    noise_wrapper::stream::test::u16_max_writes
```

mimoo:
@zekun000 it's both it seems : o

https://github.com/libra/libra/pull/4286/checks?check_run_id=740391978#step:6:1456

weirdly these don't count as a failed test

zekun000:
the tests didn't fail, it's just one of the spawned thread panic, it should also be fixed after https://github.com/libra/libra/pull/4290

mimoo:
urg, I should rebase before panicking

PR #4291, opened 2020-06-05
Ensure that all edges from test-only crates turn on the fuzzing feature if available

sunshowers:
We'd keep having CI pass because of feature unification, but breakages if people tried to run tests locally.

There's another thing we need to do, to ensure that the fuzzing feature uniformly gets turned on for regular crates as well. Will do in another PR.

### DIFF
--- a/devtools/x-core/src/workspace_subset.rs
+++ b/devtools/x-core/src/workspace_subset.rs
@@ -149,3 +149,13 @@ pub enum WorkspaceStatus {
     /// This package ID is not a dependency of the workspace subset.
     Absent,
 }
+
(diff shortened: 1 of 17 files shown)


sunshowers:
Ahh, the issue is https://github.com/rust-lang/cargo/issues/6915. Because of how the v1 feature resolver works, I don't think we can get this done until that is fixed.

bmwill:
@sunshowers Does this mean this PR is going to be shelved until that bug is fixed or would you still like a review?

sunshowers:
Yeah, the PR can't land until then. Let's try and get this fixed upstream then revisit once it lands.

PR #4296, opened 2020-06-05
[core modules] Add verification checks to Ed25519 public keys before storing them on-chain

tzakian:
Simple PR that verifies the public keys being stored on-chain. 

Updated the `dual_attestation` test, and moved the test from the functional tests in to it.


language/stdlib/modules/VASP.move
```diff
@@ -64,8 +64,8 @@ module VASP {
         base_url: vector<u8>,
         compliance_public_key: vector<u8>
     ): ParentVASP {
-        // NOTE: Only callable in testnet
-        Transaction::assert(Testnet::is_testnet(), 10041);
+        // We allow invalid compliance public keys in testnet
```
sblackshear, line 67: I'm confused by this comment. Isn't the check firing in testnet and mainnet?

language/stdlib/modules/VASP.move
```diff
@@ -117,6 +117,7 @@ module VASP {
     /// Rotate the compliance public key for `parent_vasp` to `new_key`
     public fun rotate_compliance_public_key(parent_vasp: &mut ParentVASP, new_key: vector<u8>) {
         Transaction::assert(Vector::length(&new_key) == 32, 7004);
```
sblackshear, line 119: You can drop this check because `validate_pubkey` fails if  the length is invalid.

language/stdlib/modules/VASP.move
tzakian: good spot! I wrote that then realized it should be on and forgot to remove.

PR #4297, opened 2020-06-05
[consensus] simplify timestamp guarantee

zekun000:
## Motivation

This commit removes the length wait_if_possible function, instead split the guarantees in two places:

1. round manager guarantees block.timestamp <= round deadline
2. block store guarantees any block inside has timestamp < current timestamp

Y

## Test Plan

Existing tests.

## Related PRs


aching:
Ack, at first glance I'm concerned that removing these counters gets rid of debugging information that will be useful in understanding the issues related to timestamps.

zekun000:
this type of errors belong to logs not counters imo. we never used those counters, not even once, in the mean time we got a panic with one of those counters because it's never tested.

aching:
Sure, we could do some debugging on the fly to aggregate logs to reverse generate these counters, but given this is a complex mechanism, I think counters are appropriate to get a clear signal and alert on it.  

I have looked at it in the past and observed it is non-zero, just didn't take the time to figure out what the issue was.

Personally, I find the logic split harder to understand and also the waiting semantics appear to be different.  For instance, an honest node no longer waits until the previous block's timestamp has passed prior to proposing.  It might even propose a block with a timestamp less than the previous block or one so far in the future that it has no chance of being accepted.

Is there still an issue with a panic on the counters?

zekun000:
> Sure, we could do some debugging on the fly to aggregate logs to reverse generate these counters, but given this is a complex mechanism, I think counters are appropriate to get a clear signal and alert on it.

I don't expect any alerts on this type of counters, and not sure if we'd like to debug even it's non-zero if other metrics are healthy.

>Personally, I find the logic split harder to understand and also the waiting semantics appear to be different. For instance, an honest node no longer waits until the previous block's timestamp has passed prior to proposing. It might even propose a block with a timestamp less than the previous block or one so far in the future that it has no chance of being accepted.

That's the simplicity of it, it doesn't need to worry about timestamp when generate proposals.
Honest node always waits for the block timestamp to past before inserting into the block store so there's no chance proposed timestamp less than its parent.
Honest node also proposes block when the round starts so there's no chance the proposed timestamp exceeds then round deadline.

> Is there still an issue with a panic on the counters?
no, but it shows we have never executed such code path for more than a year.

consensus/src/liveness/proposal_generator.rs
```diff
@@ -135,73 +128,7 @@ impl ProposalGenerator {
...
-                                "Even after waiting for {:?}, parent block timestamp usecs {:?} >= current timestamp usecs {:?}, will not create a proposal for this round",
-                                wait_duration,
-                                hqc.certified_block().timestamp_usecs(),
-                                current_duration_since_epoch);
-                        }
-                    };
-                }
-            }
-        };
+        let block_timestamp = self.time_service.get_current_timestamp();
```
mimoo, line 131: what was the reason for waiting for proposing a block before?

mimoo: (tests?)

consensus/src/block_storage/block_store.rs
```diff
@@ -274,6 +283,13 @@ impl BlockStore {
             return Ok(existing_block);
         }
         let executed_block = self.execute_block(block)?;
+        // ensure local time past the block time
+        let block_time = Duration::from_micros(executed_block.timestamp_usecs());
+        let current = self.time_service.get_current_timestamp();
+        if current < block_time {
+            let wait_duration = block_time - current + Duration::from_millis(1);
+            self.time_service.sleep(wait_duration);
```
mimoo, line 291: are we going to sleep 1hour here if needs be?

zekun000: yes if 1. Quorum agrees the time is 1 hour after 2. The Round deadline is 1 hour after.
we can obviously set up counter and alert for long sleep

consensus/src/liveness/proposal_generator.rs
zekun000: because the timestamp has to be monotonically increasing, and we don't wait when insert block before. so there's a chance local timestamp < proposed block's parent timestamp.

now we ensure local time past every inserted block's timestamp

consensus/src/block_storage/block_store.rs
zekun000: when a node has time screw for 1 hour we probably don't want it to serve clients. this is no different than a down/byzantine node to me.

my principle is always serving **local_state** to clients, and replicate **remote_state** if local_state agrees with remote_state not blindly even if it's signed by quorum, this helps simplify my mental model.

the state here can be either merkle state or timestamp, think about what if a quorum signs a chain, do we execute and see if it matches the local execution or blindly take it and ignore local execution (what if it's a fork chain)? the same applies to timestamp imo.

aching: > if we're optimizing for this, we shouldn't wait when propose too, just propose a timestamp = parent's + 1 can minimize sleep and get proposal accepted.

Well, this would just violate the agreement that no one proposes until their timestamp follows the protocol.  My point was within the constraints of the existing timestamp properties, we should provide committed data to clients ASAP.  VASPs would want the shortest confirmation time as possible for instance.  Given the code already exists and has been running in production for a very long time without issue I'd say it's pretty well tested.

> if we're talking about seconds or more, the operator should get alert about long waiting time and fix it rather suffer it in a weird way (I mean it'd still sleep anyway)

I agree with alerting upon issues (existing metrics can be used for this alerting), but I disagree that it is necessary to unnecessarily introduce latency for clients while the issue is being debugged.  I would assume more often than not, it might be just be a situation where ntp syncing stopped or something is wrong with the hardware (e.g. something else writing to SSD or something).  Seconds matter when the expectation is to have commit latency in the seconds.

zekun000: > within the constraints of the existing timestamp properties, we should provide committed data to clients ASAP

as I said above, I disagree we should serve data that's not consistent with local state even it's signed by quorum.

**consistency weighs more than performance** imo, even full nodes should not serve the data if its local timestamp doesn't agree.

then it's a matter of how aggressively we want to fix it, if we want to minimize the latency more, we have alerts on smaller wait durations.

aching: Maybe it's simpler to think in terms of properties:

The existing properties we have for time are:
1) Time Monotonicity: Time is monotonically increasing in the block chain. (i.e. If H1 < H2, H1.Time < H2.Time).
2) If a block of transactions B is agreed on with timestamp T, then at least f+1 honest replicas think that T is in the past. An honest replica will only vote on a block when its own clock >= timestamp T.

There is another property that you are proposing:
3) If a block of transactions B is committed or has a QC with timestamp T, an honest validator will not serve such a block (to other validators in the case of a QC or to clients in the case of a commit) until its own clock >= timestamp T.

zekun000: that's accurate, I like this summary

zekun000: the 3 property is an extrapolation of the existing property we have:

a node only replicates (and serves) a state if it agrees with it after execution.

mimoo: ^ I like this property way of describing things btw :p that'd be nice to have that in the spec

davidiw: Just glancing at the code the big delta is that instead of ending up in a stale state, which would happen if we were to wait for too long we now just sleep forever?

If my understanding is correct, I think this makes sense, we should have counters that indicate the current time before sleeping and the wake up time or the duration that way we can easily detect these kind of scenarios. It might also be useful to have an explicit state here and also have in prometheus what that state is.

In terms of implementation, with this change, I would argue we should merge secure/time and eliminate the consensus time since they effectively are the same thing now.

zekun000:
Fixed linter and added counters and error messages.

consensus/src/block_storage/block_store.rs
```diff
@@ -274,6 +282,9 @@ impl BlockStore {
             return Ok(existing_block);
         }
         let executed_block = self.execute_block(block)?;
+        // ensure local time past the block time
+        let block_time = Duration::from_micros(executed_block.timestamp_usecs());
+        self.time_service.wait_until(block_time);
```
aching, line 287: We know that the wait_time is bounded by the round duration, however it is not obvious since that check occurs outside of this function.  A naive caller could wait indefinitely.  

It would be helpful to add this assumption to the function comment that the block must have a valid timestamp that is within the current round duration.  If someone ever removes that check, it could open up a validator for infinite waiting from a rogue proposal.

aching: It might also be worth logging if the wait is a lot longer than expected.  For instance if the wait is > 1 minute, something is probably really wrong either with this node's clock or the quorum's clock.

consensus/src/liveness/proposal_generator.rs
aching: If a node accepts a block and then the clock goes backward they might accidentally propose a block with a timestamp < the previous block.  We should ensure that doesn't happen, please add a check and fail to propose due to an issue with the clock running backward.

Given it's not intuitive, let's add a comment here, something like:

"All proposed blocks in a branch are guaranteed to have increasing timestamps since their predecessor block will not be added to the BlockStore until the local time exceeds it."

zekun000:
@aching a lot of  your concern comes from clock goes backwards which I don't think it's the application's problem to deal with.

rather the TimeService should guarantee a monotonically non-decreasing timestamp view like  Instant::now https://doc.rust-lang.org/std/time/struct.Instant.html

aching:
> @aching a lot of your concern comes from clock goes backwards which I don't think it's the application's problem to deal with.
> 
> rather the TimeService should guarantee a monotonically non-decreasing timestamp view like Instant::now https://doc.rust-lang.org/std/time/struct.Instant.html

I thought about that too, but I think it would not be practical to have a monotonically non-decreasing timestamp.  Imagine if your clock gets set to 10 years in the future.  We don't want to have to wait 10 years until this validator can participate again.  The dynamic clock will adjust faster, I think we just need to adjust our properties to reflect the correct guarantee.

zekun000:
@aching I don't  think we should bother with the  property definition this much, it's easy to add assuming  honest nodes' timestamp is monotonically increasing.

If it goes back 10 years , it  certainly falls into byzantine fault category. similarly if the time goes back so that the node proposes a block with timestamp less than parent block.

Those are all the points of why we have a bft algorithm, byzantine fault is not necessarily malicious, a bug in the OS can cause a byzantine fault too e.g. timestamp goes backwards.

consensus/src/round_manager.rs
```diff
@@ -598,8 +535,8 @@ impl RoundManager {
             .get_block(executed_block.parent_id())
             .expect("[RoundManager] Parent block not found after execution");
 
-        self.wait_before_vote_if_needed(block.timestamp_usecs())
-            .await?;
+        // self.wait_before_vote_if_needed(block.timestamp_usecs())
+        //     .await?;
```
aching, line 539: Remove these?

aching:
@zekun000, overall looks good.  Can you please resolve the minor comments I made and then we can commit?

consensus/src/liveness/proposal_generator.rs
```diff
@@ -135,73 +125,7 @@ impl ProposalGenerator {
...
```
aching, line 128:
```suggestion
        // All proposed blocks in a branch are guaranteed to have increasing timestamps
        // since their predecessor block will not be added to the BlockStore until 
        // the local time exceeds it.
        let block_timestamp = self.time_service.get_current_timestamp();
```

Issue #749, opened 2019-08-28
[Bug] consensus stuck after restart and resume state sync


zekun000:
# 🐛 Bug

## To reproduce

** Stack trace/error message **
```
nightly-bench-4 validator-1e5d5a74/validator/d0b54f9991cb48d094245d55d9258844 I0828 08:43:29.287941 140506868741888 consensus/src/chained_bft/persistent_storage.rs:374] Consensus recovery done but additional state synchronization is required.
....
nightly-bench-4 validator-1e5d5a74/validator/d0b54f9991cb48d094245d55d9258844 I0828 08:43:31.161557 140506868741888 network/src/peer_manager/mod.rs:443] Inbound connection with peer ab0d6a54 established
nightly-bench-4 validator-1e5d5a74/validator/d0b54f9991cb48d094245d55d9258844 I0828 08:43:33.834076 140506868741888 network/src/peer_manager/mod.rs:443] Inbound connection with peer 57ff8374 established
nightly-bench-4 validator-1e5d5a74/validator/d0b54f9991cb48d094245d55d9258844 I0828 08:43:34.243086 140506868741888 network/src/connectivity_manager/mod.rs:216] Should be connected to peer: 8deeeaed at addr(s): ********
nightly-bench-4 validator-1e5d5a74/validator/d0b54f9991cb48d094245d55d9258844 I0828 08:43:34.243144 140506868741888 network/src/connectivity_manager/mod.rs:240] Will dial peer: 8deeeaed after 2s
nightly-bench-4 validator-1e5d5a74/validator/d0b54f9991cb48d094245d55d9258844 I0828 08:43:34.524315 140506868741888 network/src/protocols/discovery/mod.rs:350] Received updated note for peer: 8deeeaed from peer: 57ff8374
nightly-bench-4 validator-1e5d5a74/validator/d0b54f9991cb48d094245d55d9258844 I0828 08:43:34.524367 140506868741888 network/src/protocols/discovery/mod.rs:350] Received updated note for peer: 57ff8374 from peer: 57ff8374
nightly-bench-4 validator-1e5d5a74/validator/d0b54f9991cb48d094245d55d9258844 I0828 08:43:34.524390 140506868741888 network/src/protocols/discovery/mod.rs:350] Received updated note for peer: ab0d6a54 from peer: 57ff8374
[...]

PR #4299, opened 2020-06-05
[node] Refactor the network setup code to eliminate redundancies

agc-quadra:
## Motivation

Simplify main_node by eliminating code redundancies and streamlining the steps for constructing a libra node.

This is an isolated baby step along the path to a simpler main_node/network_builder endgoal.

yes
## Test Plan

don't break existing integration/unit tests.

## Related PRs

### DIFF
--- a/libra-node/src/main_node.rs
+++ b/libra-node/src/main_node.rs
@@ -240,7 +240,7 @@ pub fn setup_environment(node_config: &mut NodeConfig) -> LibraHandle {
     let mut network_runtimes = vec![];
     let mut state_sync_network_handles = vec![];
     let mut mempool_network_handles = vec![];
-    let mut validator_network_provider = None;
+    let mut consensus_network_handles = None;
     let mut reconfig_subscriptions = vec![];
 
     let (mempool_reconfig_subscription, mempool_reconfig_events) =
@@ -265,7 +265,7 @@ pub fn setup_environment(node_config: &mut NodeConfig) -> LibraHandle {
     }
 
     let mut validator_count = 0;
-    // Instantiate every network.
+    // Instantiate every network and collect the requisite endpoints for state_sync, mempool, and consensus.
     for (role, network_config) in network_configs {
         // Perform common instantiation steps
         let (runtime, mut network_builder) =
@@ -284,13 +284,14 @@ pub fn setup_environment(node_config: &mut NodeConfig) -> LibraHandle {
         );
         mempool_network_handles.push((peer_id, mempool_sender, mempool_events));
 
-        // Perform actions specific to a particular network type.
         match role {
+            // Perform steps relevant specifically to Validator networks.
             RoleType::Validator => {
                 validator_count += 1;
                 // A valid config is allowed to have exactly one ValidatorNetwork
                 assert_eq!(1, validator_count);
                 // Set up to listen for network configuration changes from StateSync.
+                // TODO:  move this inside network_builder.
                 if let Some(conn_mgr_reqs_tx) = network_builder.conn_mgr_reqs_tx() {
                     let (simple_discovery_reconfig_subscription, simple_discovery_reconfig_rx) =
                         gen_simple_discovery_reconfig_subscription();
@@ -302,18 +303,19 @@ pub fn setup_environment(node_config: &mut NodeConfig) -> LibraHandle {
                         .spawn(network_config_listener.start(simple_discovery_reconfig_rx));
                 };
 
-                // Cache the network information to be started later.
-                validator_network_provider = Some((peer_id, runtime, network_builder));
+                consensus_network_handles = Some(consensus::network_interface::add_to_network(
+                    &mut network_builder,
+                ));
             }
-            RoleType::FullNode => {
-                // Cache the runtime so it does not go out of scope immediately.
-                network_runtimes.push(runtime);
-
-                // Start the network provider.  FullNode networks can operate silently.3
-                let _listen_addr = network_builder.build();
-                debug!("Network started for peer_id: {}", peer_id);
-            }
+            // Currently no FullNode network specific steps.
+            RoleType::FullNode => (),
         }
+
+        // Start the network and cache the runtime so it does not go out of scope.
+        // TODO:  move all 'start' commands to a second phase at the end of setup_environment.  Target is to have one pass to wire the pieces together and a second pass to start processing in an appropriate order.
+        let _listen_addr = network_builder.build();
+        network_runtimes.push(runtime);
+        debug!("Network started for peer_id: {}", peer_id);
     }
 
     // TODO set up on-chain discovery network based on UpstreamConfig.fallback_network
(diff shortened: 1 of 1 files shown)


agc-quadra:
@davidiw  Incremental improvement to main_node

libra-node/src/main_node.rs
```diff
@@ -255,60 +255,68 @@ pub fn setup_environment(node_config: &mut NodeConfig) -> LibraHandle {
...
         let (state_sync_sender, state_sync_events) =
             state_synchronizer::network::add_to_network(&mut network_builder);
         state_sync_network_handles.push((peer_id, state_sync_sender, state_sync_events));
+
+        // Create the endpoints to connect the network to MemPool.
         let (mempool_sender, mempool_events) = libra_mempool::network::add_to_network(
             &mut network_builder,
             node_config.mempool.max_broadcasts_per_peer,
         );
         mempool_network_handles.push((peer_id, mempool_sender, mempool_events));
```
davidiw, line 288: rather than merge these into a single vector of network_configs it might make sense to move this logic to setup_network.

libra-node/src/main_node.rs
```diff
@@ -255,60 +255,68 @@ pub fn setup_environment(node_config: &mut NodeConfig) -> LibraHandle {
...
+                        .handle()
+                        .spawn(network_config_listener.start(simple_discovery_reconfig_rx));
+                };
+
+                consensus_network_handles = Some(consensus::network_interface::add_to_network(
+                    &mut network_builder,
+                ));
+            }
+            // Currently no FullNode network specific steps.
+            RoleType::FullNode => (),
```
davidiw, line 313: then we don't have to have this awkward situation

libra-node/src/main_node.rs
agc-quadra: These should definitely stay outside of 'setup_network' (at least in its current form.

Taking a step back, I'm aiming to apply the following principles:

1. Make building hierarchical
1. main_node is the 'root' builder for the entire system.
1. move setup_network out of the root and into network
1. ensure that a builder explicitly references 'child' components but not 'sibling' components.   consensus and networking are siblings;  healthchecker is a child of networking.

agc-quadra:
> Much appreciation for identifying the common code and unifying it! Perhaps we can organize it in a slightly different approach? Not sure without spending more time thinking about it. Let me know your thoughts.

we will circle back around to this point over time.   I dont think this is the final step, but there are other bigger changes which it unlocks before I want to get into the exact suggestion you're making here.

libra-node/src/main_node.rs
agc-quadra: I actually see this as a feature not a bug for the time being --- similar to not wanting to overthink moving things into setup_network at this point, keeping this specific hook in place while refactoring the builders will be useful (and also makes it very explicit that there is no functionality that is currently full node specific/where that functionality should be configured if it appears)

libra-node/src/main_node.rs
```diff
@@ -255,60 +255,68 @@ pub fn setup_environment(node_config: &mut NodeConfig) -> LibraHandle {
...
     }
 
-    for mut full_node_network in node_config.full_node_networks.iter_mut() {
-        let (runtime, mut network_builder) = setup_network(
-            &mut full_node_network,
-            RoleType::FullNode,
-            Arc::clone(&db_rw.reader),
-            waypoint,
-        );
+    let mut validator_count = 0;
```
davidiw, line 269: this is the type of weird logic that makes me feel this isn't as great :/

libra-node/src/main_node.rs
```diff
@@ -255,60 +255,68 @@ pub fn setup_environment(node_config: &mut NodeConfig) -> LibraHandle {
...
         let (state_sync_sender, state_sync_events) =
             state_synchronizer::network::add_to_network(&mut network_builder);
         state_sync_network_handles.push((peer_id, state_sync_sender, state_sync_events));
+
+        // Create the endpoints to connect the network to MemPool.
         let (mempool_sender, mempool_events) = libra_mempool::network::add_to_network(
             &mut network_builder,
             node_config.mempool.max_broadcasts_per_peer,
         );
         mempool_network_handles.push((peer_id, mempool_sender, mempool_events));
```
davidiw, line 287: maybe we should have add_common_services and put these in there? it seems all we need is network_builder and node_config... network_builder already has the peer_id.

libra-node/src/main_node.rs
davidiw: that's fair, forgive me for nitpicking here, but you know once you finish this, no one will revisit this code except to come back and complain it is too confusing -- without realizing how much more confusing it was :)

libra-node/src/main_node.rs
```diff
@@ -351,22 +359,16 @@ pub fn setup_environment(node_config: &mut NodeConfig) -> LibraHandle {
...
-        // network provider -> consensus -> state synchronizer -> network provider. This deadlock
-        // was observed in GitHub Issue #749. A long term fix might be make
-        // consensus initialization async instead of blocking on state synchronizer.
-        let (consensus_network_sender, consensus_network_events) =
-            consensus::network_interface::add_to_network(&mut network_builder);
-        let _listen_addr = network_builder.build();
-        network_runtimes.push(runtime);
-        debug!("Network started for peer_id: {}", peer_id);
-
+    // Note: We need to start network provider before consensus, because the consensus
```
davidiw, line 362: haha, you don't want to touch this overly verbose statement that explicitly boils down to

Start state sync first to prepare the ledger for consensus and prevent a cyclic dependency that might cause a deadlock. See GitHub Issue #749 [link].

libra-node/src/main_node.rs
agc-quadra: A couple of outcomes come up:

1). I finish the work and the sequence of ~10 PRs needed to get there including documenting the final state.

2).  I don't finish the work and the next person tries to make it cleaner working from a moderately improved state all the while cursing my name.

Either way we're in a better position moving forward.

libra-node/src/main_node.rs
agc-quadra: I'm not touching it.  I'm leaving it in place.   problem is that the network start code all moved up :)

libra-node/src/main_node.rs
agc-quadra: Two different weird logics.  We'll have to choose a poison; i can actually replace this specific one and use the Option<protocol handles> instead to make it just a different kind of the same weird logic that was already htere.

libra-node/src/main_node.rs
agc-quadra: Lets's do that in a follow up.   Avoid scope creep and work on incremental improvements.

libra-node/src/main_node.rs
davidiw: well... that's an unfortunately reasonable argument.

PR #4302, opened 2020-06-05
[stdlib] Create a separate failure epilogue for transactions failed transaction

runtian-zhou:
…o be exectued

## Motivation

Create a separate failed epilogue for transactions that failed to be executed. Currently the code look exactly the same as the regular prologue. We might want to implement some different logic for charging in the future.

Yes


runtian-zhou:
I would also want to take this chance to discuss about the charging rule for failed transaction. Currently we charge the gas used when the execution aborts at failure, this introduced a bunch of logics in the implementation of libra adapter that seems quite weird, as we need to keep track of the unit of gas left during the execution. The logic here may be much simplified if we just charge the max unit of gas at failure. The change will affect two cases:
1. When the transaction aborted in the middle of transaction
2. When the transaction drained out of its fund after the script body execution and has no balance to pay for the gas.

In either case we charge the gas that's been consumed before. The cost for both cases will be higher. Was wondering what's your take @tzakian @sblackshear

language/stdlib/modules/LibraAccount.move
```diff
@@ -943,6 +943,43 @@ module LibraAccount {
         }
     }
 
+    // The epilogue is invoked at the end of transactions when the transaction is aborted during execution or during
+    // regular epilogue.
+    //
+    // It collects gas and bumps the sequence number
+    fun failure_epilogue<Token>(
```
sblackshear, line 950: Could we extract the common logic into a private function (or a pair of private functions) and call it from both success and failure epilogues? E.g.,

```
success_epilogue() {
        let transaction_fee_amount = txn_gas_price * (txn_max_gas_units - gas_units_remaining);
        assert(balance_for(sender_balance) >= transaction_fee_amount)
        epilogue(transaction_fee_amount);
}

failure_epilogue() {
        let transaction_fee_amount = txn_gas_price * (txn_max_gas_units - gas_units_remaining);
        epilogue(transaction_fee_amount);
}
```

language/stdlib/modules/LibraAccount.move
```diff
@@ -943,6 +933,43 @@ module LibraAccount {
...
+        let transaction_fee_amount = txn_gas_price * (txn_max_gas_units - gas_units_remaining);
+        Transaction::assert(
+            balance_for(sender_balance) >= transaction_fee_amount,
+            6
+        );
+        epilogue<Token>(sender, transaction_fee_amount, txn_sequence_number);
+    }
+
+    // The failure_epilogue is invoked at the end of transactions when the transaction is aborted during execution or during
+    // regular epilogue.
```
sblackshear, line 958:
```suggestion
    // `success_epilogue`.
```

language/stdlib/modules/LibraAccount.move
```diff
@@ -945,6 +935,43 @@ module LibraAccount {
...
+        txn_sequence_number: u64,
+        txn_gas_price: u64,
+        txn_max_gas_units: u64,
+        gas_units_remaining: u64
+    ) acquires T, Balance, AccountOperationsCapability {
+        let sender = Signer::address_of(account);
+        // Charge for gas
+        let transaction_fee_amount = txn_gas_price * (txn_max_gas_units - gas_units_remaining);
+
+        epilogue<Token>(sender, transaction_fee_amount, txn_sequence_number);
```
jolestar, line 972: success_epilogue and failure_epilogue main difference are not asserting 'balance >= transaction_fee_amount' ? what happened if user balance is < transaction_fee_amount ? abort at withdraw_from_balance ?

runtian-zhou: That's a great question! The failure epilogue will be executed from clean state. So the assertion should never fail as we've already check the balance at prologue and nothing is done afterwards.

runtian-zhou:
Just run the cluster test myself and got the following output:
`all up : 1384 TPS, 3262 ms latency, 3950 ms p99 latency, no expired txns`
Will rerun the job again

Issue #1200, opened 2019-10-04
[netcore] Add DNS Transport and Multiaddr support


phlip9:
# 🚀 Feature Request

The current supported base multiaddr formats are: [ `/ip4/1.2.3.4`, `/ip6/2620:10d:c097::c0de`, `/memory` ].

Implementing this feature would add support for dialing multiaddrs with the `dns` prefix, i.e., `/dns/example.com`. A DNS Transport would likely wrap an inner transport and just map a `/dns/example.com` prefix to an `/ip4/1.2.3.4` or `/ip6/::1` prefix, which would then be delegated to the inner transport. It's not clear that listening on a DNS Transport makes sense, other than just delegating listening to the inner transport.

There is also a question of how to resolve dns domain names. The simplest approach is to use rust's (and by extension the OS's) built-in resolver via [`std::net::ToSocketAddr`](https://doc.rust-lang.org/std/net/trait.ToSocketAddrs.html). The downside is that `.to_socket_addr()` is a blocking call, which is generally a no-no in an async `Future`. The alternative is to use [`tokio-dns-unofficial`](https://github.com/sbstp/tokio-dns), which adds a small threadpool for resolving dns lookups. Another alternative is to pull in [`trust-dns-resolver`](https://github.com/bluejekyll/trust-dns/tree/master/crates/resolver), which is a clean-room, rust-only, user-space DNS resolver that supports modern DNS-over-HTTPS  (among other things).

I suspect just using `ToSocketAddr` for now is fine.

Another small issue is that rust-libp2p doesn't parse the `/dns` prefix (only `/dns4` and `/dns6` prefixes), despite `/dns` being a "de-facto" standard for years (https://github.com/multiformats/multiaddr/blob/master/protocols.csv). Implementing this feature might have to include forking our multiaddr lib or upstreaming a small diff so we can support `/dns` in the list of protocols.

cc @bothra90 @bmwill 

EDIT(2020/01/16): With #2215 landing, we now support dialing and advertising dns domain names. From looking at tokio, it already handles running the dns lookup on a threadpool (https://docs.rs/tokio/0.2.9/src/tokio/net/addr.rs.html#130). The only remaining work is then supporting the `/dns` prefix and restricting `/dns4` and `/dns6` lookups to ip4 and ip6 respectively.

PR #4310, opened 2020-06-05
[network] filter ip's from DNS resolver according to protocol

phlip9:
`NetworkAddress` allows us to advertise a dns name but restrict a dialer
to only connect over IPv4 or IPv6. This diff actually enforces this
distinction in our tcp dialing logic, so we will never accidentally
connect over IPv4 if the address only allows IPv6 (and vice versa).

For example, dialing `"/dns4/example.com/tcp/8080"` will now only ever
connect over IPv4 and likewise for `"/dns6/example.com/tcp/8080"`.

Of course, if the address advertiser doesn't care (or supports both
protocols), we continue to support `"/dns/example.com/tcp/8080"`, which
does no filtering and will use whatever resolved SocketAddr that works
first.

Closes: #1200

### DIFF
--- a/network/netcore/src/transport/tcp.rs
+++ b/network/netcore/src/transport/tcp.rs
@@ -99,7 +99,8 @@ impl Transport for TcpTransport {
         // ensure addr is well formed to save some work before potentially
         // spawning a dial task that will fail anyway.
         // TODO(philiphayes): base tcp transport should not allow trailing protocols
-        None.or_else(|| parse_ip_tcp(protos).map(|_| ()))
+        parse_ip_tcp(protos)
+            .map(|_| ())
             .or_else(|| parse_dns_tcp(protos).map(|_| ()))
             .ok_or_else(|| invalid_addr_error(&addr))?;
 
--- a/network/netcore/src/transport/tcp.rs
+++ b/network/netcore/src/transport/tcp.rs
@@ -9,16 +9,17 @@ use futures::{
     ready,
     stream::Stream,
 };
-use libra_network_address::{parse_dns_tcp, parse_ip_tcp, NetworkAddress, Protocol};
+use libra_network_address::{parse_dns_tcp, parse_ip_tcp, IpFilter, NetworkAddress};
 use std::{
     convert::TryFrom,
     fmt::Debug,
     io,
+    net::SocketAddr,
     pin::Pin,
     task::{Context, Poll},
     time::Duration,
 };
-use tokio::net::{TcpListener, TcpStream};
+use tokio::net::{lookup_host, TcpListener, TcpStream};
 
 /// Transport to build TCP connections
 #[derive(Debug, Clone, Default)]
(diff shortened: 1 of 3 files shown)


network/netcore/src/transport/tcp.rs
```diff
@@ -93,14 +94,17 @@ impl Transport for TcpTransport {
...
+
+        // ensure addr is well formed to save some work before potentially
+        // spawning a dial task that will fail anyway.
         // TODO(philiphayes): base tcp transport should not allow trailing protocols
-        // TODO(philiphayes): use `tokio::net::lookup_host()` then filter the results
-        // so e.g. `Protocol::Dns4` only returns `SocketAddrV4`s and `Protocol::Dns6`
-        // only returns `SocketAddrV6`.
+        None.or_else(|| parse_ip_tcp(protos).map(|_| ()))
+            .or_else(|| parse_dns_tcp(protos).map(|_| ()))
+            .ok_or_else(|| invalid_addr_error(&addr))?;
```
gregnazario, line 104: I'm not sure what the`None.or_else(` paradigm does.  But I do get that you're parsing for IP/TCP & DNS/TCP to see if the address is valid.

network/netcore/src/transport/tcp.rs
```diff
@@ -93,14 +94,17 @@ impl Transport for TcpTransport {
     }
 
     fn dial(&self, addr: NetworkAddress) -> Result<Self::Outbound, Self::Error> {
-        let (addr_string, _addr_suffix) =
-            parse_addr_and_port(addr.as_slice()).ok_or_else(|| invalid_addr_error(&addr))?;
+        let protos = addr.as_slice();
```
gregnazario, line 97: qq: What does protos stand for at this point?

network/netcore/src/transport/tcp.rs
```diff
@@ -109,6 +113,59 @@ impl Transport for TcpTransport {
...
+            match TcpStream::connect(socketaddr).await {
+                Ok(stream) => return Ok(stream),
+                Err(err) => last_err = Some(err),
+            }
+        }
+
+        Err(last_err.unwrap_or_else(|| {
+            io::Error::new(
+                io::ErrorKind::InvalidInput,
+                "could not resolve to any address",
```
gregnazario, line 154: Should we say what IP Filter we used for debugging purposes?

network/netcore/src/transport/tcp.rs
zekun000: can't it be just 
```
parse_ip_tcp(protos).map(|_| ()).or_else(parse_dns_tcp(protos).map(|_| ()).ok_or_else(..)?;
```

network/netcore/src/transport/tcp.rs
phlip9: good idea!

network/netcore/src/transport/tcp.rs
phlip9: `NetworkAddress` is "just" a list of `Protocol`s (really a `Vec<Protocol>`), so `NetworkAddress::as_slice()` gives us a slice into the list of protocols, hence "protos" for short.

network/netcore/src/transport/tcp.rs
phlip9: the only reason for the `None.or_else` (which is effectively a no-op) is so the parsing lines align. cargo fmt would otherwise align this block like

```rust
parse_ip_tcp(protos)
    .map(|_| ())
    .or_else(|| parse_dns_tcp(protos).map(|_| ()))
    .ok_or_else(|| invalid_addr_error(&addr))?;
```

kind of on-the-fence about this pattern... I think it looks nicer but is also clearly more confusing : (

@zekun000 what would you prefer?

gregnazario: I'd go over straightforward code than looking nicer.  That way the next guy doesn't come in and go, what is this `None.or_else()` and ask the same question.

zekun000: agree, I prefer straightforward and less confusing one too

phlip9: done!

PR #4311, opened 2020-06-05
[language][vm] binary file format cleanup

vgao1996:
## Summary
This implements a few changes to the binary file format, with the goal of getting it finalized. 

List of changes:
- Improved representation for table indices & content blob.
- Self module handle index is more lightweight and no longer occupies a table.
- Main is more lightweight and no longer occupies a table.
- Using uleb128 for branch offsets.
- Using uleb128 for counts & lengths.

Each change is in a separate commit.

## Test Plan
cargo test


dariorussi:
let's talk about this. There is a usage of uleb that I am not clear about, particularly the byte.
I wonder if we went a bit too far

vgao1996:
@dariorussi You made a really good point. Let me explain why I switched to uleb in those places. 

Recall that the uleb128 encoding has two interesting properties: 
1) Its storage size is propotional to the size of the integer, which can drastically improve storage efficiency in certain cases, with decent worst case performance.
  - 0-127: 1 byte
  - 128-16383: 2 bytes
  - 16384-2097511: 3 bytes
2) It's a variable length encoding capable of storing integers of any size.

When I made these changes I had (2) in my mind more than (1) -- having a variable length encoding will allow us to have a consistent encoding for all indices and lengths, whose limits can be raised in a backward compatible way. (Of course (1) still matters, and we'll see it later.) 

To take a closer look at what this PR changes:
1) Signatures (Locals, Function Parameters etc.)
2) Type arguments for Struct Instantiations
3) Jump Instructions
4) Code length

For signatures and type arguments, the numbers stay under 128 99% of the time, so changing from the naive 1-byte encoding to uleb will unlikely introduce a noticable regression. For the rare case where you have a Signature of length 128, the extra overhead of 1B is tiny compared to the 128 Signature Tokens you store...
For code lengths and offsets, a lot of our functions stay under 128 and the complicated ones are still likely to stay under 16384, so this even sounds like a pure win to me.

Summarizing all I said above, uleb128 seems to be a free lunch of flexibility, storage efficiency and consistency, and it's hard for me to come up with a reason to not grab and eat it :)

bob-wilson:
> Summarizing all I said above, uleb128 seems to be a free lunch of flexibility, storage efficiency and consistency, and it's hard for me to come up with a reason to not grab and eat it :)

uleb128 is great for minimizing size, but it makes it harder to efficiently read/write the file format and could impact performance. I suspect that Move files will remain small enough for the foreseeable future that neither the size nor performance will be significant issues, though.

As an example, in many file formats like this, you can put all the counts in fixed-size fields in a header and quickly multiply those by the fixed-size fields to calculate the overall space to allocate for tables and then use fast bulk copy operations to pull the data in from the file (or even mmap directly). Etc. I'm not suggesting that we attempt to rearrange the Move format to optimize like that, or even advocating against uleb128 -- just pointing out the tradeoff here....

dariorussi:
@vgao1996 I understand where you are coming from but I am still unconvinced uleb'ing all things in the format is the right thing to do. @bob-wilson has good reference suggestions and of course those are not all things we can apply to our format (given the lack of alignment rules) however it is something to keep in mind. It is also something that seems reasonable to me, particularly the ability to load the headers without any compression.
Anyway about some specifics here:
- a more backward compatible solution: I am not convinced this is really true. To me it would provide a possibly less intrusive change in your code if your code was well written. What I mean here is that when we save into a byte we have a check in the code that the specific value is `< u8::max` and if we changed that, we would have to change the code and version the VM. In other words once we lock down all the checks, the choice of a given size should be tested for and as such we need to be able to work with old and new versions.
- on the one byte uleb. I don't think the issue is whether that would have a significant impact (which like you I doubt it) but it's really just a logical fact of having a chance to double the size of a byte for no reason. The idea that I need a uleb to read a single byte is a bit of a stretch. Of course in some situation here we are just making the length of an array more standard and that is something to consider.
- on what needs to be compressed vs uncompressed. I think the main issue here is whether we should choose to align things a bit better in the binary format and if so definitely a top level uncompressed makes sense to me. As Bob mentioned you can do easy math and break up the file as needed. That is a nice property no matter perf or other considerations. However right now we are not really accounting for much of a memory map story so we don't have that pressure. On the other end having us using uleb everywhere because in being a variable size encoding may "let you skip defining a size" is a bit of an over engineering that I am not sure it buys us much. 

We are going to talk later about all of this, just wanted to put some references here....

vgao1996:
@bob-wilson you brought up a very good point. I had the same concerns initially, but eventually convinced myself this is not going to be an issue. Let me elaborate my thoughts on these tradeoffs.

Since our table indices are unordered, in order to actually understand something in the content blob, you'll need to parse all the indices anyways. Fixed size indices will allow you to locate and mmap the blob directly, but without the indices you still don't know which bytes are part of which table. Therefore fixed size fields isn't really making a meaningful difference here.

vgao1996:
@dariorussi thanks for the comments. The points on alignment and the ability to read the top level constructs uncompressed are interesting and we'll definitely talk about these. There could be something I missed about these, especially the latter one. 

- Backward compatibility: my understanding is that with the all-uleb approach, you don't need different code for different versions of the deserializer if the two versions only differ in maximum values allowed. You do need to version the limits though.
- 1-byte uleb. Yeah my point was more about having a consistent encoding for all lengths and offsets. I felt it would complicate the spec if we had uleb and 2B/4B little endian at the same time.

Anyways I don't have a strong opinion either way and I'd be eager to learn more about how other languages/virtual machines handle this. (I had a look at webassembly and noticed they use leb128 for all integers, including indices and integer values, which is an interesting reference point in addition to JVM and CLR. I actually wonder if webassembly is closer to what we're doing than general purpose VMs like JVM or CLR.)

PR #4312, opened 2020-06-05
[move-prover] Specified Offer.move.

kkmc:
Wrote functional specifications for all four functions and
two global specifications to restrict creation and removal of
Offer.T resources to the create and remove functions.

Additional notes: In an earlier implementation, I was able to write
a more general property that states that only the sender or the intended
recipient (i.e. Offer<T<Offered>>.for) of an Offer could receive the
offered struct (refer to the properties below). This would allow
for additional functions that change the Offer to be implemented without
changing the specification (e.g. if an update offer function is
implemented).

```
/// **Informally:** If the Offer is removed from an address, then the sender
///                 must be the intended recipient
ensures all(domain<address>(), |addr|
  (old(exists<T<Offered>>(addr)) && !exists<T<Offered>>(addr)) ==>
  (old(sender_is_offer_recipient<Offered>(addr))));
/// **Informally:** An offer under an account is unchanged if it is not the sender's
///                 account or if the sender is not the recipient
ensures all(domain<address>(), |addr|
  (old(exists<T<Offered>>(addr)) && old(!sender_is_offer_recipient<Offered>(addr))) ==>
  (exists<T<Offered>>(addr) && old(global<T<Offered>>(addr)) == global<T<Offered>>(addr)));
/// **Informally:** An offer under an account is unchanged if the intended
///                 recipient does not redeem it
ensures all(domain<address>(), |addr|
  (old(exists<T<Offered>>(addr)) && exists<T<Offered>>(addr)) ==>
  (old(global<T<Offered>>(addr)) == global<T<Offered>>(addr)));
```

However, with the introduction of signers the property is false
without the precondition / assumption that the signer is the sender,

## Motivation

First draft of formal specifications for functional and global properties of the Offer module. 
 
Yes.

## Test Plan

Current test suite. 

To verify the specifications, run the following command at the move-prover directory:

```
cargo run -- --verbose debug ../stdlib/modules/transaction.move ../stdlib/modules/Signer.move ../stdlib/modules/Offer.move
```

## Related PRs

None.

### DIFF
--- a/language/stdlib/modules/Offer.move
+++ b/language/stdlib/modules/Offer.move
@@ -35,6 +35,97 @@ module Offer {
   public fun address_of<Offered>(offer_address: address): address acquires T {
     borrow_global<T<Offered>>(offer_address).for
   }
+
+  // **************** SPECIFICATIONS ****************
+
+  /*
+  This module defines a resource T that is used as a permissioned trading scheme between accounts.
+  It defines two main functions for creating and retrieving a struct offered by some user
+  inside the resource T under the offerer's account.
+
+  Currently, the only other module that depends on this module is LibraConfig, where it's used to
+  pass a capability to an account that allows it to modify a config.
+  */
+
+  /// # Module specification
+
+  spec module {
+    /// Verify all functions in this module
+    pragma verify = true;
+    /// Helper function that returns whether or not the `recipient` is an intended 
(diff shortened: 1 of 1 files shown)


kkmc:
@DavidLDill @wrwg @junkil-park 
Can someone please change the reviewer for me?

Issue #4065, opened 2020-05-26
Support --health-check in swarm mode


andll:
Cluster test can be run with --health-check in normal mode (see `cluster_test/main.rs:run_health_check`). With this command cluster test tails event stream of a node and runs the health check.

We want to be able to run this functionality in a swarm mode - so method similar to `run_health_check` should be added to `BasicSwarmUtil`.
This way we can verify health checks without running AWS cluster.

This requires few changes in various components, although all of those changes are fairly small

1) When running libra swarm(cargo run -p libra-swarm), it prints out command line for cluster test. Currently this command line has `--peers` section with coma separated list of ip:port. Port here is an jsonrpc port. We need to modify this string here and add debug interface port to it, so that for each peer we have `ip:rpc.port:debug_interface.port`

2) In the cluster test we need to add `debug_interface_port` field to `Instance` class. This field should be Optional. In the `DebugPortLogThread::spawn_new`, we need to read this port from the instance(panic if it is not set), instead of hardcoded value `6191`.

3) In swarm mode we need to parse input(see `BasicSwarmUtil::setup`) and if it is in form ip:port:port, use second port to set `debug_interface_port` on the instance. We should still support old format of ip:port, and set debug_interface_port to None if it is not present.

4) In Instance::new_k8s we want to use `NodeConfig::default().port:debug_interface.port` as a value for the debug_interface_port.

5) After this all preparation is done last step is to actually add implementation for --health-check to BasicSwarmUtil, similar to `ClusterTestRunner::run_health_check`

To verify how this works
1) Run libra swarm locally (`cargo run -p libra-swarm`)
2) Run cluster test on started swarm cluster (`cargo run -p cluster-test --swarm --peers <as printed by libra swarm> --health-check`).

PR #4313, opened 2020-06-05
[cluster-test] Support --health-check in swarm mode

changyonghu:
Closes: #4065
Manually test cargo run -p libra-swarm and cargo run -p cluster-test --swarm --peers <as printed by libra swarm> --health-check --duration 10. Also rechecked ./scripts/cti --pr <your PR> --health-check --duration 10 in normal (aws) mode.

### DIFF
--- a/testsuite/cluster-test/src/cluster.rs
+++ b/testsuite/cluster-test/src/cluster.rs
@@ -23,7 +23,7 @@ pub struct Cluster {
 }
 
 impl Cluster {
-    pub fn from_host_port(peers: Vec<(String, u32)>, mint_file: &str) -> Self {
+    pub fn from_host_port(peers: Vec<(String, u32, Option<u32>)>, mint_file: &str) -> Self {
         let instances: Vec<Instance> = peers
             .into_iter()
             .map(|host_port| {
@@ -31,6 +31,7 @@ impl Cluster {
                     format!("{}:{}", &host_port.0, host_port.1), /* short_hash */
                     host_port.0,
                     host_port.1,
+                    host_port.2,
                 )
             })
             .collect();
--- a/testsuite/cluster-test/src/health/debug_interface_log_tail.rs
+++ b/testsuite/cluster-test/src/health/debug_interface_log_tail.rs
@@ -42,7 +42,11 @@ impl DebugPortLogThread {
         for instance in cluster.all_instances() {
             let (started_sender, started_receiver) = mpsc::channel();
             started_receivers.push(started_receiver);
-            let client = NodeDebugClient::new(instance.ip(), 6191);
+            let debug_interface_port = instance
+                .debug_interface_port()
+                .expect("Debug interface port is not found")
+                as u16;
+            let client = NodeDebugClient::new(instance.ip(), debug_interface_port);
             let debug_port_log_thread = DebugPortLogThread {
                 instance: instance.clone(),
                 client,
--- a/testsuite/cluster-test/src/main.rs
+++ b/testsuite/cluster-test/src/main.rs
@@ -111,8 +111,8 @@ pub fn main() {
 
     let args = Args::from_args();
 
-    if args.swarm && !(args.emit_tx || args.diag) {
-        panic!("Can only use --emit-tx or --diag in --swarm mode");
+    if args.swarm && !(args.emit_tx || args.diag || args.health_check) {
+        panic!("Can only use --emit-tx or --diag or --health-check in --swarm mode");
     }
 
     if args.diag {
@@ -148,6 +148,13 @@ pub fn main() {
             ));
             return;
         }
+    } else if args.health_check && args.swarm {
+        let util = BasicSwarmUtil::setup(&args);
+        let logs = DebugPortLogThread::spawn_new(&util.cluster).0;
+        let mut health_check_runner = HealthCheckRunner::new_all(util.cluster);
+        let duration = Duration::from_secs(args.duration);
+        exit_on_error(run_health_check(&logs, &mut health_check_runner, duration));
+        return;
     }
 
     let mut runner = ClusterTestRunner::setup(&args);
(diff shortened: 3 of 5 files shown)


testsuite/cluster-test/src/cluster.rs
```diff
@@ -23,14 +23,15 @@ pub struct Cluster {
 }
 
 impl Cluster {
-    pub fn from_host_port(peers: Vec<(String, u32)>, mint_file: &str) -> Self {
+    pub fn from_host_port(peers: Vec<(String, u32, Option<u16>)>, mint_file: &str) -> Self {
```
andll, line 26: I think either u32 or u16 are fine for port, but within single crate(cluster test in this case) we want to use same type - it does not make sense that first port in this tuple is u32 and second is u16

testsuite/cluster-test/src/health/debug_interface_log_tail.rs
```diff
@@ -42,7 +42,8 @@ impl DebugPortLogThread {
         for instance in cluster.all_instances() {
             let (started_sender, started_receiver) = mpsc::channel();
             started_receivers.push(started_receiver);
-            let client = NodeDebugClient::new(instance.ip(), 6191);
+            let client =
+                NodeDebugClient::new(instance.ip(), instance.debug_interface_port().unwrap());
```
andll, line 46: Can you replace `.unwrap` with `.expect("Some sensible message about debug interface port required for debug log thread")`

testsuite/cluster-test/src/main.rs
```diff
@@ -302,8 +316,40 @@ impl BasicSwarmUtil {
             .iter()
             .map(|peer| parse_host_port(peer).unwrap())
             .collect();
+
+        let cluster = Cluster::from_host_port(parsed_peers, &args.mint_file);
+
+        let log_tail_started = Instant::now();
```
andll, line 322: Can we set up tailer inside `--health-check` action?
We need to make sure we can still run `--swarm --emit-tx` without debug port - we use it when generating load for premainnet, and there is no access to debug port there. With current change we would loose this ability as BasicSwarmUtil here requires debug port to start DebugPortLogThread.

testsuite/cluster-test/src/main.rs
```diff
@@ -302,8 +316,40 @@ impl BasicSwarmUtil {
...
+
         Self {
-            cluster: Cluster::from_host_port(parsed_peers, &args.mint_file),
+            logs,
+            cluster,
+            health_check_runner,
+        }
+    }
+
+    fn run_health_check(&mut self, duration: Duration) -> Result<()> {
```
andll, line 338: Any chance we can organize it similarly to `main.rs/emit_tx` function - have single function outside of any class(instead of duplication in BasicSwarmUtil and Runner)?

testsuite/libra-swarm/src/main.rs
```diff
@@ -103,8 +103,12 @@ fn main() {
...
         .iter()
         .map(|config| {
-            let port = NodeConfig::load(config).unwrap().rpc.address.port();
-            format!("localhost:{}", port)
+            let validator_config = NodeConfig::load(config).unwrap();
+            let port = validator_config.rpc.address.port();
+            let debug_interface_port = validator_config
+                .debug_interface
+                .admission_control_node_debug_port;
+            format!("localhost:{}:{}", port, debug_interface_port)
```
andll, line 111: Can we actually print two separate options for cluster test here?

```
To run transaction generator run: 
(old cmd line without debug port in peers, make sure it works)
To run health check:
(new command line with debug port and --health-check action)

**Comment by andll** [comment]
Looks mostly good, but had some comments

File: testsuite/libra-swarm/src/main.rs (hunk 5 above)
Comment (changyonghu) [reply testsuite/libra-swarm/src/main.rs:128]: Not sure if this is ok. I could not find a way to reuse node_address_list and provide readability.

File: testsuite/cluster-test/src/main.rs [hunk 6]
```diff
@@ -148,6 +148,16 @@ pub fn main() {
             ));
             return;
         }
+    } else if args.health_check && args.swarm {
+        let mut util = BasicSwarmUtil::setup(&args);
+        util.logs = Some(DebugPortLogThread::spawn_new(&util.cluster).0);
```
andll, line 153: Is `BasicSwarmUtil` every reads `logs`?
It feels like you just set up logs here, put it into BasicSwarmUtil and then read here?

changyonghu: I put logs and health_check_runner as member of BasicSwarmUtil. HealthCheckRunner will read logs.

testsuite/libra-swarm/src/main.rs
andll: @changyonghu I think in terms of 'how to make it look good', I would separate 'data extraction' and 'data formatting'.

1) Have a var that would hold `Vec<(rpc_port, debug_port)>`.
2) Separately `.map` this var to either 'localhost:rpc_port` or `localhost:rpc_port:debug_port`

testsuite/cluster-test/src/main.rs
andll: What I am trying to say here, is that you don't seem to need `log` as a field on BasicSwarmUtil?
You can just define `let logs = DebugPortLogThread::spawn_new` here

changyonghu: Got it.

testsuite/cluster-test/src/main.rs
```diff
@@ -148,6 +148,16 @@ pub fn main() {
             ));
             return;
         }
+    } else if args.health_check && args.swarm {
+        let mut util = BasicSwarmUtil::setup(&args);
+        let logs = DebugPortLogThread::spawn_new(&util.cluster).0;
+        let duration = Duration::from_secs(args.duration);
+        exit_on_error(run_health_check(
+            &logs,
+            &mut util.health_check_runner,
```
andll, line 157: Same question for `health_check_runner` - looks like it could be variable, not a field on `BasicSwarmUtil`?

changyonghu: Yes, agree.

PR #4316, opened 2020-06-05
[cluster-test] LSR and Vault integration with cluster-test

ankushagarwal:
## Summary

LSR and Vault integration with cluster-test

* Update set_asg_size so that it can handle cluster-size of more than 100 by using `next_token` 
* Add LSR and Vault pod and service specs
* Update validator spec to take another boolean - `enable_lsr`
* Use single service account - clustertest for all pods
* Add VaultConfig and LSRConfig to InstanceConfig enum
* Take `enable_lsr` as a boolean flag in cluster test main.rs

### DIFF
--- a/docker/build-aws.sh
+++ b/docker/build-aws.sh
@@ -26,7 +26,7 @@ while [[ "$1" =~ ^- ]]; do case $1 in
     BUILD_PROJECTS=(libra-validator libra-cluster-test libra-init libra-mint libra-safety-rules)
     ;;
   --build-all-cti )
-    BUILD_PROJECTS=(libra-validator libra-cluster-test libra-safety-rules)
+    BUILD_PROJECTS=(libra-validator libra-cluster-test libra-init libra-safety-rules)
     ;;
   --build-cluster-test )
     BUILD_PROJECTS=(libra-cluster-test)
--- a/testsuite/cluster-test/src/aws.rs
+++ b/testsuite/cluster-test/src/aws.rs
@@ -41,34 +41,45 @@ pub async fn set_asg_size(
     libra_retrier::retry_async(libra_retrier::fixed_retry_strategy(10_000, 30), || {
         let asc_clone = asc.clone();
         Box::pin(async move {
-            let auto_scaling_group_names_type = AutoScalingGroupNamesType {
-                auto_scaling_group_names: Some(vec![asg_name.to_string()]),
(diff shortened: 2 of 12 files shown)


andll:
This PR is in 'draft' - not ready for review yet?

Also, I am super curious what TPS we get with vault+lsr?

ankushagarwal:
Still testing this out, so its in draft.

Initial results don't seem to be good :( https://toroeng.slack.com/archives/CSZN5D27R/p1591402946160200

ankushagarwal:
@andll : Marked this as ready for review.
Verified that with `enable_lsr` set to false, we get same TPS results.
With `enable_lsr` set, we seem to be seeing a lot of timeouts in consensus - this is something which will need to be debugged.

andll:
Actually, I realized formula was correct.
What is our plan for actually turning on vault+lsr in cluster test?
Do you want to merge this diff to avoid merge conflict and look into problems after that? or..?

ankushagarwal:
> What is our plan for actually turning on vault+lsr in cluster test?

Once we figure out the timeouts issue we can enable it in nightly. What we need to figure out at a later point is whether we want to run all our experiments with LSR+Vault enabled or run LSR+Vault experiments separately from our current benchmark numbers.

> Do you want to merge this diff to avoid merge conflict and look into problems after that? or..?

Yup, the timeouts issue with LSR+Vault can be investigated later on, we should not block this diff from landing as it does not affect the current experiments.

andll:
But how do we know if this code actually works/it's not an issue with cluster test setup?

ankushagarwal:
> But how do we know if this code actually works/it's not an issue with cluster test setup?

All the validators are successfully committing transactions with this setup.

andll:
I see. Ok, let's land it @bors-libra r+

ankushagarwal:
Updated `--build-all-cti` -> `--build-all` since now we need to build `libra_init` image as well

andll:
Wait, so if build-cti is basically what we need for cti, if we need init image, we need to include it in build-cti, not switch to build-all.
Otherwise cti will stop working

@bors-libra r-

andll:
The reason we did this split between build-all and build-cti is because some images(like faucet) are not needed by cluster test

ankushagarwal:
ack. Updated `build-all-cti`

@bors-libra r=andll

PR #4319, opened 2020-06-06
[move-prover] Attempt to verify Vector::reverse.

kkmc:
Initial attempt / ramp up exercise to verify vector reverse's loop invariant.
Verified the reverse function when the argument is not a reference, but unable to verify the loop invariant when the argument has type &mut vector.

When references are used, I can prove the following at every loop iteration:

```
assert all(0..len, |k| old_v<Element>[k] == v[k])
```

Which means that the old array is always equal to the current array.
Also, placing the assertion that the elements are swapped
right after Vector::swap fails when v is a mutable reference (&mut):

```
assert old_v<Element>[front_index] == v[back_index];
```

## Motivation

First attempt at verifying `verify_reverse` / `Vector::reverse` using a loop invariant.

Yes.

## Test Plan

Run the following to verify `verify_reverse` (and other functions in the file):
```
cargo run -- --verbose debug ../stdlib/modules/Vector.move tests/sources/functional/verify_vector.move
```

## Related PRs

None.

### DIFF
--- a/language/move-prover/tests/sources/functional/verify_vector.move
+++ b/language/move-prover/tests/sources/functional/verify_vector.move
@@ -97,22 +97,129 @@ module VerifyVector {
         ensures result[0] == e;
     }
 
+    spec module {
+        /// Ghost variable `old_v` used to store the old value of the array v
+        global old_v<Element>: vector<Element>;
+    }
+
     // Reverses the order of the elements in the vector in place.
     fun verify_reverse<Element>(v: &mut vector<Element>) {
         let len = Vector::length(v);
         if (len == 0) return ();
 
-//        let front_index = 0;
-//        let back_index = len -1;
-//        while (front_index < back_index) {
-//            Vector::swap(v, front_index, back_index);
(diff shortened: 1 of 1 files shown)


kkmc:
@junkil-park @shazqadeer
Can someone change the reviewer for me? Thanks.

shazqadeer:
Was @andll added to the list of reviewers for this PR by mistake?

kkmc:
> Was @andll added to the list of reviewers for this PR by mistake?

I was unable to choose reviewers at the time of the PR, so the bot choose a random one for me.

@junkil-park I just updated the file to have both versions.

junkil-park:
@kkmc, Shaz just submitted a PR (https://github.com/libra/libra/pull/4324) to fix the issue in the mutable parameters. Could you try verifying the original again once it's landed?

junkil-park:
@kkmc, thank you for your work on this!! Since the loop invariants for `Vector::reverse` were landed via https://github.com/libra/libra/pull/4613. I will close this PR.

PR #4325, opened 2020-06-06
[network] error on receiving a noise handshake payload in FN network

mimoo:
In a full node network, or as we call it currently a ServerOnly handshake mode,
we do not send a payload during the handshake.
(In the validator network we do, for anti-replay mitigation.)
Thus, we should check on the responder/server side that the payload was empty,
and return an error otherwise.

### DIFF
--- a/network/src/noise/handshake.rs
+++ b/network/src/noise/handshake.rs
@@ -289,7 +289,8 @@ impl NoiseUpgrader {
             }
         }
 
-        // if mutual auth mode, verify this handshake is not a replay
+        // if on a mutually authenticated network,
+        // the payload should contain a u64 client timestamp
         if let Some(anti_replay_timestamps) = self.auth_mode.anti_replay_timestamps() {
             // check that the payload received as the client timestamp (in seconds)
             if payload.len() != PAYLOAD_SIZE {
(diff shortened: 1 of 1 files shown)


network/src/noise/handshake.rs
```diff
@@ -323,6 +324,12 @@ impl NoiseUpgrader {
 
             // store the timestamp
             anti_replay_timestamps.store_timestamp(their_public_key, client_timestamp);
+        } else if payload.len() != 0 {
+            // TODO: security logging
+            return Err(io::Error::new(
+                io::ErrorKind::InvalidData,
+                "noise: a handshake payload was received in the full-node network: {}",
+            ));
```
davidiw, line 332: I wonder if we already get this logging via the error same as above...

mimoo: yeah, and in any case it'll be cleaner to consolidate the logging on the caller side :o good point

phlip9: 1. probably remove the "{}" if you're not going to format anything : p

2. maybe call this "noise: received an unexpected handshake payload on a public-facing full-node network"?

phlip9:
@bors-libra delegate+

land after fixing small nit : )

network/src/noise/handshake.rs
mimoo: > probably remove the "{}" if you're not going to format anything : p

🤦🏼‍♂️

PR #4328, opened 2020-06-07
[move-prover] Add "assume !$abort_flag" at each label

shazqadeer:
## Motivation

This PR makes the following changes:
- add "assume  !$abort_flag" at each label
- replace "assert !$abort_flag$ at the beginning of a procedure with "assume !$abort_flag"
- enable loops.move test

Yes

## Test Plan

Enabled a test.

### DIFF
--- a/language/move-prover/src/bytecode_translator.rs
+++ b/language/move-prover/src/bytecode_translator.rs
@@ -662,7 +662,7 @@ impl<'env> ModuleTranslator<'env> {
         emitln!(self.writer, "var $saved_m: $Memory;");
 
         emitln!(self.writer, "\n// initialize function execution");
-        emitln!(self.writer, "assert !$abort_flag;");
+        emitln!(self.writer, "assume !$abort_flag;");
         emitln!(self.writer, "$saved_m := $m;");
 
         emitln!(self.writer, "\n// track values of parameters at entry time");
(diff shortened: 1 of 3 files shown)


junkil-park:
My one little concern about having "assume !$abort_flag;" at the beginning of each basic block is that: it could introduce a contradiction when `$abort_flag` is set to be "true", but the procedure does not exit right away due to a bug/mistake. 

Could we also add "assert !$abort_flag;" somewhere properly to make sure the absence of this situation?

shazqadeer:
I had thought about your concern when I was deciding whether to put assert or assume at each label.  I realized that putting an assume is sound because whenever $abort_flag is set to true, control jumps to the end of the function and from then on the entire call stack is unwound.  All this is being done inside the translator from Move into Boogie. So unless there is a bug in this translation scheme, putting an assume should be sound.

wrwg:
I'm concerned that this is just special-casing one particular instance of the loop havoc problem. Isn't there a better and more general solution?

shazqadeer:
Indeed, it is special handling for $abort_flag.  Unfortunately, the only general solution I know is automatic inference of loop invariants.  We are currently not doing that and instead requiring the programmer to supply loop invariants manually.  We could have approached the problem similarly for $abort_flag except that $abort_flag is a variable introduced by the translation.

Issue #3342, opened 2020-04-09
[network] Improve initialization via NetworkBuilder


bothra90:
The `NetworkBuilder` module has seen significant change over time which give it a "batteries included, but optional" nature, since modules such as Discovery and HealthChecker are only added if needed. However, there's still room for some simplifications:
  - arguments (such as `trusted_peers`) that are specific to `discovery` module should be initialized in the `add_discovery` function rather than separately, which leaves room for bugs.
  - Similarly, arguments for health_checker (`ping_interval_ms` and `ping_timeout_ms`) should be sent in `add_connection_monitoring`.

davidiw:
Before jumping into there, we should consider merging NetworkBuilder and NetworkConfig, since the two are basically the same thing.

Rustinante:
Agreed, the `NetworkBuilder` seems to have too many states, refactoring the code into a more functional style probably be better.

Rustinante:
Hi @davidiw and @phlip9, I've created a PR #3881 for this issue to refactor the `NetworkBuilder`. It turns out we need a structure to store the various network senders, and it feels a little weird to put them into the `NetworkConfig`, so I created another struct called `NodeNetwork` to store them, so now we have these two structures, where the first one is the configs responsible to set up the network, while the second one is sort of the result of setting up the network...

Please let me know what you think, and we can refactor this further.

PR #4330, opened 2020-06-07
Let NetworkBuilder Consume NetworkConfig

Rustinante:
## Motivation

Right now both `NetworkBuilder` and `NetworkConfig` have these fields:
- `peer_id`
- `network_id`
- `listen_address`
- `advertised_address`
- `discovery_interval_ms`
- `connectivity_check_interval_ms`

This PR will let `NetworkBuilder` incorporate `NetworkConfig` as a member to avoid the duplication of these fields.

Added `fn clone_without_private_key` to `network_config::Identity` so that `clone_for_template` preserves the `Identity` enum variant for `NetworkConfig`.

Yes

## Test Plan

cargo test

### DIFF
--- a/libra-node/src/main_node.rs
+++ b/libra-node/src/main_node.rs
@@ -17,7 +17,7 @@ use libra_logger::prelude::*;
 use libra_mempool::gen_mempool_reconfig_subscription;
 use libra_metrics::metric_server;
 use libra_secure_storage::config;
-use libra_types::{waypoint::Waypoint, PeerId};
+use libra_types::waypoint::Waypoint;
 use libra_vm::LibraVM;
 use libradb::LibraDB;
 use network::validator_network::network_builder::{AuthenticationMode, NetworkBuilder};
@@ -74,7 +74,6 @@ fn setup_debug_interface(config: &NodeConfig) -> NodeDebugService {
 
 pub fn setup_onchain_discovery(
     network: &mut NetworkBuilder,
-    peer_id: PeerId,
     role: RoleType,
     libra_db: Arc<dyn DbReader>,
     waypoint: Waypoint,
@@ -98,7 +97,7 @@ pub fn setup_onchain_discovery(
         let storage_query_ticker = interval(Duration::from_secs(30)).fuse();
 
         OnchainDiscovery::new(
-            peer_id,
+            network.peer_id,
             role,
             waypoint,
             network_tx,
@@ -114,7 +113,7 @@ pub fn setup_onchain_discovery(
 
 // TODO(abhayb): Move to network crate (similar to consensus).
 pub fn setup_network(
-    config: &mut NetworkConfig,
+    mut config: NetworkConfig,
     role: RoleType,
     libra_db: Arc<dyn DbReader>,
     waypoint: Waypoint,
@@ -126,27 +125,21 @@ pub fn setup_network(
         .build()
         .expect("Failed to start runtime. Won't be able to start networking.");
 
-    let identity_key = config::identity_key(config);
-    let peer_id = config::peer_id(config);
+    let identity_key = config::identity_key(&mut config);
 
-    let mut network_builder = NetworkBuilder::new(
-        runtime.handle().clone(),
-        config.network_id.clone(),
-        peer_id,
-        role,
-        config.listen_address.clone(),
-    );
+    let mut network_builder = NetworkBuilder::new(config, runtime.handle().clone(), role);
     network_builder.add_connection_monitoring();
 
-    if config.enable_remote_authentication {
+    if network_builder.config.enable_remote_authentication {
         // Sanity check seed peer addresses.
-        config
+        network_builder
+            .config
             .seed_peers
             .verify_libranet_addrs()
             .expect("Seed peer addresses must be well-formed");
 
-        let network_peers = config.network_peers.peers.clone();
-        let seed_peers = config.seed_peers.seed_peers.clone();
+        let network_peers = network_builder.config.network_peers.peers.clone();
+        let seed_peers = network_builder.config.seed_peers.seed_peers.clone();
 
         let trusted_peers = if role == RoleType::Validator {
             // for validators, trusted_peers is empty will be populated from consensus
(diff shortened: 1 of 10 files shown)


Rustinante:
Hi @davidiw and @phlip9 , I don't have the permission to add reviewers, could you guys help take a look? Thank you!

davidiw:
This makes sense but I think @agclement is a few steps further ahead, would like his feedback. I'll also send you an email to follow up.

agclement:
> This makes sense but I think @agclement is a few steps further ahead, would like his feedback. I'll also send you an email to follow up.

I agree with the intent of this change and have been contemplating when to tackle it directly.   In the long run, I would like to see network_builder be the unique consumer of a top level config and for network_builder to consume every field, either directly or by delegating the field to an internal component.

I estimate that I'm about a week out from having the builder structure in place where that set of evaluations can be sensibly made;  in the meantime I believe this change is an incremental improvement in that direction.

Rustinante:
So shall we still proceed with this change? I think it's still failing some tests which would require some digging...

Rustinante:
All checks are passing now, please see if this looks okay to land. Thanks!

libra-node/src/main_node.rs
```diff
@@ -255,14 +241,15 @@ pub fn setup_environment(node_config: &mut NodeConfig) -> LibraHandle {
...
+            validator_network,
             RoleType::Validator,
             Arc::clone(&db_rw.reader),
             waypoint,
         );
-        let peer_id = network_builder.peer_id();
+        let peer_id = network_builder.peer_id;
 
         // Set up to listen for network configuration changes from StateSync.
         if let Some(conn_mgr_reqs_tx) = network_builder.conn_mgr_reqs_tx() {
```
agclement, line 255: There is a conflict between this code and #4299 .   This set of changes is also unrelated to the announced intent of the PR to make NetworkBuilder take the NetworkConfig.  

Suggest to role this set of changes back and keep the PR narrowly focused on having NetworkBuilder consume NetworkConfig.  Those changes should not have a blast radius that reaches outside of setup_network.

network/src/validator_network/network_builder.rs
```diff
@@ -125,21 +118,14 @@ pub struct NetworkBuilder {
...
 impl NetworkBuilder {
     /// Return a new NetworkBuilder initialized with default configuration values.
-    pub fn new(
-        executor: Handle,
-        network_id: NetworkId,
-        peer_id: PeerId,
-        role: RoleType,
-        listen_address: NetworkAddress,
-    ) -> NetworkBuilder {
+    pub fn new(network_config: NetworkConfig, executor: Handle, role: RoleType) -> NetworkBuilder {
```
agclement, line 128: This change is going sideways at present.  Intuitively passing `network_config` as a parameter here rather than the individual subfields is an improvement because it reduces the number of parameters to `new(...)`.   But it's pyrrhic:  specifically, we (a) still have an inordinate number of fields in NetworkBuilder, (b) we're using network_config as a dictionary where we are interested in a very small number of the fields of network_config, and now (c) there are very similar fields which are in different places for seemingly arbitrary reasons.

The main pain point this PR is addressing is that there are too many fields in network_builder.  The 'fix' here  reduces the number of fields from way way too many to way too many, and introduces a struct field (`network_config`) with many unused subfields.  Not clear if that's a win.

Is there anyway that we can better structure the fields in `network_builder` so it is clear which component requires them?   Would it make sense for new to create a set of XXX configs, one per internal component, which are directly fed to those components?

network/src/protocols/network/dummy.rs
```diff
@@ -132,13 +135,16 @@ pub fn setup_network() -> DummyNetwork {
...
+    listener_network_config.random_with_peer_id(&mut rng, Some(listener_peer_id));
+    listener_network_config.listen_address = listener_addr;
     // Set up the listener network
     let mut network_builder = NetworkBuilder::new(
+        listener_network_config,
         runtime.handle().clone(),
-        network_id.clone(),
-        listener_peer_id,
         RoleType::Validator,
-        listener_addr,
```
agclement, line 141: Are these the only three fields of `network_config`?  Or are there a multitude of other fields?

If the former, then this collapse makes a lot of sense.  If the latter, then my experience says we're better of peeling the outer layer of the onion away as we go down the rabbit hole;  put another way, if `network_builder` is not using 3/4 of the fields in `network_config`, then network_builder should never see the `network_config` object.

network/src/validator_network/network_builder.rs
```diff
@@ -302,18 +261,19 @@ impl NetworkBuilder {
...
             ConnectivityManager::new(
-                peer_id,
+                self.peer_id,
                 trusted_peers,
                 seed_peers,
-                interval(Duration::from_millis(connectivity_check_interval_ms)).fuse(),
+                interval(Duration::from_millis(
+                    self.config.connectivity_check_interval_ms,
+                ))
+                .fuse(),
```
agclement, line 276: either always use `self.___` below or never.  The middle ground where `self` is used sometimes but not always is  between two worlds.

network/src/validator_network/network_builder.rs
Rustinante: I'm not sure this is sideways. From your description it seems that it's an incremental improvement. I think to totally refactor the NetworkBuilder might be too much for one PR, and would potentially step on your toes even more. Right now even with this small change it seems that it would likely cause some merge conflict with your changes.

network/src/protocols/network/dummy.rs
Rustinante: I'm not sure what you mean by the only fields of `network_config`, since it has more fields than these. These are removed because they are duplicates between `NetworkBuilder` and `NetworkConfig`

network/src/validator_network/network_builder.rs
Rustinante: Sounds good, I'll update those as well

Rustinante:
@agclement I think there are two questions:
1. Whether this PR constitutes an incremental improvement towards #3342 
2. Whether this PR will cause insurmountable problems for you at resolving merge conflicts.

From my understanding 1 is likely to be true. Please let me know if 2 is a problem. If it will make it impossible for you to rebase then I won't land these changes, but will leave them here for you to incorporate as part of your changes as you see fit.

network/src/validator_network/network_builder.rs
agclement: > In #3342 they want to combine `NetworkConfig` and `NetworkBuilder`, and I think this is in the right direction to achieve that.

Can you sketch the next couple of steps to get there?  I know I've had challenges with incremental changes in the past when the reviewers didn't have a sense for the roadmap I was following.

Rustinante: I don't have any roadmap for how to get the `NetworkBuilder` completely refactored...

agclement:
> @agclement I think there are two questions:
> 
> 1. Whether this PR constitutes an incremental improvement towards #3342
> 2. Whether this PR will cause insurmountable problems for you at resolving merge conflicts.
> 
> From my understanding 1 is likely to be true. Please let me know if 2 is a problem. If it will make it impossible for you to rebase then I won't land these changes, but will leave them here for you to incorporate as part of your changes as you see fit.

The only issue I see for (2) is the changes in main_node which seem to be completely orthogonal to the primary point of the PR.

For (1), a sketch of the next steps would help my appreciation that this increment is an improvement.

Rustinante:
It's not possible to this PR without changing the `main_node.rs`
I don't have any roadmap for the next steps. We have to use our common sense to see if these changes actually help with anything. I think they do because reducing the duplication between `NetworkConfig`  and `NetworkBuilder` reduces the chances of having bugs and makes the whole thing a little easier to reason about.

davidiw:
I think this PR makes sense, but there's kind of a collision in this space. To minimize overlap, I'm going to close out this PR for now, let's revisit this a while later.

PR #4339, opened 2020-06-08
[Move] Mint tiers based on time of AOS approval check 

moeziniaf:
Mint flow is based on a preapproval by AOS (which collects and validates proposal for mint) and then at any point in the future, a fiat transfer triggers the actual mint event.

Previously, if we used the current time (of mint flow, triggered by fiat transfer) then AOS would have a hard time calculating how many approvals are needed. This change enforces chain to base check off of AOS' preapproval collection and verificaition.

Add some more tiered mint tests to ensure windowing checks work, and reset properly based on new time model.


andll:
Why are we using approval timestmap and not block timestmap?

dmitri-perelman:
@andll the timestamp here corresponds to the timestamp used by the client (in our case AOS) at the moment it determines the policy to be used for collecting the signatures. We can call this timestamp different things but the most important thing is that it'll be passed by the client as part of the transaction. Otherwise AOS and Libra would consider different time windows for tier calculation.

moeziniaf:
See PR description ^ - AOS needs to know which tier to mint in (and therefore how many signatures to collect). If we use blocktimestamp, AOS will run out of sync with chain and txns may fail often because of incorrect tier

language/transaction-builder/src/lib.rs
```diff
@@ -396,9 +396,10 @@ encode_txn_script! {
 encode_txn_script! {
     name: encode_tiered_mint,
     type_arg: coin_type,
-    args: [nonce: U64, designated_dealer_address: Address, mint_amount: U64, tier_index: U64],
+    args: [nonce: U64, designated_dealer_address: Address, mint_amount: U64, tier_index: U64, approval_timestamp: U64],
     script: TieredMint,
     doc: "Mints 'mint_amount' to 'designated_dealer_address' for 'tier_index' tier.\
+          Tier based on time of AOS approval (approval_timestamp unixtime) of operation.
```
sblackshear, line 402: Libra time is (somewhat oddly) Unix time in microseconds--might be worth mentioning that here and/or double-checking that this is what AOS is doing.

moeziniaf: ah @dmitri-perelman would aos prefer unix_time in seconds? I can make that change here

dmitri-perelman: No preferences, this code is not written yet :). If all the other Libra timestamps are using microseconds, we should probably use microseconds here as well...

andll:
Do we have any doc/spec that describes this tiered mint time usage?
E.g.
1. This is the goal
2. This is what client/aos/hsm/chain checks
3. This is why goal is achieved

moeziniaf:
I don't believe so but I can add it to https://fburl.com/ecdze5ur

tzakian:
One thing with the pre-approval timestamp: don't we also want to check that the provided timestamp is less than the current timestamp on-chain? This way mint tiers can't be artificially "elongated" by saying the pre-approval is in the future.

andll:
I think it would be good to have some description about how those timestmaps allow to achieve the goal - it is not obvious to me for example

dmitri-perelman:
@tzakian great catch. Yes, we should absolutely enforce that the given timestamp is smaller than the current time!

PR #4340, opened 2020-06-08
[cleanup][types, json-rpc, storage, consensus, network] Remove 22 more untested, unused items

huitseeker:
This PR removes ~~09~~ 22 untested, unused methods from `types` and various other modules. The value of the PR may reside more in the signaling of the nature of those untested, unused methods rather than actual removal.


phlip9:
network stuff LGTM!

davidiw:
but yet no one is bold enough to accept

@bors-libra delegate+

PR #4341, opened 2020-06-08
[move prover] specify TestNet.move

emmazzz:
## Motivation

Add global specifications and pre/post conditions to TestNet.move in stdlib

Yes.

## Test Plan

```
cargo run -- ../stdlib/modules/Signer.move ../stdlib/modules/Transaction.move ../stdlib/modules/Testnet.move
```

### DIFF
--- a/language/stdlib/modules/Testnet.move
+++ b/language/stdlib/modules/Testnet.move
@@ -21,5 +21,96 @@ module Testnet {
         Transaction::assert(Signer::address_of(account) == 0xA550C18, 0);
         IsTestnet{} = move_from<IsTestnet>(0xA550C18);
     }
+
+    // **************** SPECIFICATIONS ****************
+
+    /**
+       This module is used by Genesis, LibraAccount and Unhosted as part of
+       the initialization and to check whether the transaction is on the
+       testnet.
+    **/
+
+    /// # Module Specification
+
+    spec module {
+        /// Verify all functions in this module, including private ones.
+        pragma verify = true;
+
+        /// Returns the association root address.
+        define spec_root_address(): address { 0xA550C18 }
+
+        /// Returns true if the testnet has been initialized.
+        /// Initialization can be reversed by function `remove_testnet`.
+        define spec_is_initialized(): bool {
+            exists<IsTestnet>(spec_root_address())
+        }
+    }
+
+    /// ## Management of TestNet
+
+    spec module {
+        /// Returns true if no address has IsTestNet resource.
+        define spec_no_addr_has_testnet(): bool {
+            all(domain<address>(), |addr| !exists<IsTestnet>(addr))
+        }
+
+        /// Returns true if only the root address has an IsTestNet resource.
+        define spec_only_root_addr_has_testnet(): bool {
+            all(domain<address>(), |addr|
+                exists<IsTestnet>(addr)
+                    ==> addr == spec_root_address())
+        }
+    }
+
+    spec schema OnlyRootAddressHasTestNet {
+        /// Base case of the induction before the association address is
+        /// initialized with IsTestnet.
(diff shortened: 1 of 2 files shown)


language/stdlib/modules/Testnet.move
```diff
@@ -21,5 +21,97 @@ module Testnet {
         Transaction::assert(Signer::address_of(account) == 0xA550C18, 0);
         IsTestnet{} = move_from<IsTestnet>(0xA550C18);
     }
+
+    // **************** SPECIFICATIONS ****************
+
+    /*
+       This modules is used by Genesis, LibraAccount and Unhosted as part of
```
DavidLDill, line 28: should be "module" not "modules"

language/stdlib/modules/Testnet.move
```diff
@@ -21,5 +21,97 @@ module Testnet {
...
+    */
+
+    /// # Module Specification
+
+    spec module {
+        /// Verify all functions in this module, including private ones.
+        pragma verify = true;
+
+        /// Returns the association root address.
+        define spec_root_address(): address { 0xA550C18 }
```
DavidLDill, line 40: We should import this from somewhere else.  The code needs to be reorganized to have all these functions in one place, for both Move and specs.  No change needed now.

language/stdlib/modules/Testnet.move
```diff
@@ -21,5 +21,97 @@ module Testnet {
...
+
+        /// Returns true if the testnet has been initialized.
+        define spec_testnet_initialized(): bool {
```
DavidLDill, line 43: I think we should call similar functions "spec_is_initialized" in all modules, consistently.

Please add a comment that this doesn't follow our conventions because it can be reversed by remove_testnet.

language/stdlib/modules/Testnet.move
```diff
@@ -21,5 +21,97 @@ module Testnet {
...
+            exists<IsTestnet>(spec_root_address())
+        }
+    }
+
+    /// ## Management of TestNet
+
+    spec module {
+        /// **Informally:** No address has IsTestNet resource,
```
wrwg, line 51: I find this "informally" convention redundant. Indeed everything not in spec lang is informal. I'd prefer we use just prose here (full English sentences as common in documentations). Also the function definition itself doesn't state any property until it is applied, so the "informal" doc string is misleading.

language/stdlib/modules/Testnet.move
```diff
@@ -21,5 +21,97 @@ module Testnet {
         Transaction::assert(Signer::address_of(account) == 0xA550C18, 0);
         IsTestnet{} = move_from<IsTestnet>(0xA550C18);
     }
+
+    // **************** SPECIFICATIONS ****************
+
+    /*
```
wrwg, line 27: Is this intended to be part of the documentation? Then one should use `/**`.

Also if this is general information, it could as well go to the module, and doesn't belong into the specification section.

language/stdlib/modules/Testnet.move
```diff
@@ -21,5 +21,97 @@ module Testnet {
...
+    }
+
+    spec fun initialize {
+        aborts_if Signer::get_address(account) != spec_root_address();
+        aborts_if spec_exists_testnet(account);
+        ensures spec_exists_testnet(account);
+    }
+
+    spec fun is_testnet {
+        aborts_if false; // why does ading aborts_if true give me "postcondition might not hold"?
```
wrwg, line 107: Please add a TODO if you add comments like this, so we can find such places and remove them from the final source.

I guess that is the question we discussed in the standup. Could you please open an issue for this?

emmazzz:
bors-libra r=wrwg

PR #4344, opened 2020-06-08
[network][noise] send peer id and remote public key as prologue

mimoo:
DRAFT


network/src/noise/handshake.rs
```diff
@@ -124,6 +125,8 @@ impl HandshakeAuthMode {
 
 /// The Noise configuration to be used to perform a protocol upgrade on an underlying socket.
 pub struct NoiseUpgrader {
+    /// The validator's own peer id.
+    peer_id: PeerId,
```
phlip9, line 129: i've started using `self_peer_id` to make it less confusing with a "remote" peer id

network/src/noise/handshake.rs
```diff
@@ -132,13 +135,41 @@ pub struct NoiseUpgrader {
...
+        if let Some(trusted_peers) = self.auth_mode.trusted_peers() {
+            trusted_peers
+                .read()
+                .map_err(|_| {
+                    io::Error::new(
+                        io::ErrorKind::Other,
+                        "noise: unable to read trusted_peers lock",
+                    )
+                })?
+                .iter()
```
phlip9, line 158: maybe assign the read.map_err? to an intermediate variable, this is a pretty long chain

network/src/noise/handshake.rs
```diff
@@ -187,7 +218,7 @@ impl NoiseUpgrader {
         &self,
         mut socket: TSocket,
         remote_public_key: x25519::PublicKey,
-    ) -> io::Result<NoiseStream<TSocket>>
+    ) -> io::Result<(NoiseStream<TSocket>, PeerId)>
```
phlip9, line 221: switch the order here

network/src/noise/handshake.rs
```diff
@@ -251,39 +290,91 @@ impl NoiseUpgrader {
     pub async fn upgrade_inbound<TSocket>(
         &self,
         mut socket: TSocket,
-    ) -> io::Result<NoiseStream<TSocket>>
+    ) -> io::Result<(NoiseStream<TSocket>, PeerId)>
```
phlip9, line 293: same

network/src/noise/handshake.rs
```diff
@@ -236,8 +272,11 @@ impl NoiseUpgrader {
             .finalize_connection(initiator_state, &server_response)
             .map_err(|e| io::Error::new(io::ErrorKind::Other, e))?;
 
+        // fetch their peer id
+        let their_peer_id = self.identity_pubkey_to_peer_id(remote_public_key)?;
```
phlip9, line 276: remote_peer_id

network/src/noise/handshake.rs
mimoo: I've used `my_` and `their_` also, maybe we should make sure we keep it consistent. No strong opinion :)

network/src/noise/handshake.rs
mimoo: why?

network/src/noise/handshake.rs
mimoo: (I removed that function in the second commit)

network/src/validator_network/network_builder.rs
```diff
@@ -414,14 +414,22 @@ impl NetworkBuilder {
...
-        let (key, maybe_trusted_peers) = match authentication_mode {
-            AuthenticationMode::ServerOnly(key) => (key, None),
-            AuthenticationMode::Mutual(key) => (key, Some(self.trusted_peers.clone())),
+        let (key, maybe_trusted_peers, peer_id) = match authentication_mode {
+            AuthenticationMode::ServerOnly(key) => {
+                let public_key = key.public_key();
+                // TODO: should I check that self.peer_id is not set then?
+                let peer_id = PeerId::from_identity_public_key(public_key);
+                (key, None, peer_id)
+            }
```
davidiw, line 423: This isn't accurate and perhaps was part of my hangup in the previous PR. The upstream peers ID may be in a trusted peers from upstream. If I am the client, my peer id is derived this way, but my peer id may not be derived.

language/move-core/types/src/account_address.rs
```diff
@@ -70,6 +70,19 @@ impl AccountAddress {
 
         AccountAddress::try_from(padded_result)
     }
+
+    // Note: This is inconsistent with current types because AccountAddress is derived
+    // from consensus key which is of type Ed25519PublicKey. Since AccountAddress does
+    // not mean anything in a setting without remote authentication, we use the network
+    // public key to generate a peer_id for the peer.
+    // See this issue for potential improvements: https://github.com/libra/libra/issues/3960
+    pub fn from_identity_public_key(identity_public_key: x25519::PublicKey) -> Self {
```
phlip9, line 79: this should be a `From` impl

```rust
impl From<x25519::PublicKey> for AccountAddress {
    fn from(pubkey: x25519::PublicKey) -> Self {
        let mut array = [0u8; Self::LENGTH];
        let pubkey_slice = identity_public_key.as_slice();
        // keep only the last 16 bytes
        array.copy_from_slice(&pubkey_slice[x25519::PUBLIC_KEY_SIZE - Self::LENGTH..]);
        Self(array)
    }
}
```

then we can do `PeerId::from(remote_static_pubkey)`

network/Cargo.toml
```diff
@@ -22,6 +22,7 @@ thiserror = "1.0.19"
 tokio = { version = "0.2.21", features = ["full"] }
 tokio-util = { version = "0.3.1", features = ["codec"] }
 tokio-retry = "0.2.0"
+hex = "0.4.2"
```
phlip9, line 25: this isn't a lint yet (it should be) but i prefer when the crates are sorted

language/move-core/types/src/account_address.rs
mimoo: intuitively, it feels cleaner if this is a custom function as it indicates that it is not doing something straight forward (a hack) 

(as opposed to `From` which usually sound to me like they're just convenience functions)

what do you think?

network/src/noise/handshake.rs
gregnazario: I've also seen `own_peer_id` along with `self_peer_id`.

network/src/validator_network/network_builder.rs
mimoo: what previous PR btw : o ?

not sure I understand your comment. Why would you set a peer_id if you're in the full node network?

network/src/noise/handshake.rs
```diff
@@ -179,11 +187,11 @@ impl NoiseUpgrader {
...
     ///
     /// This runs the "client" side of the Noise IK handshake to establish a
-    /// secure Noise stream and exchange static public keys. In mutual auth
-    /// scenarios, we will also include an anti replay attack counter in the
+    /// secure Noise stream and send its static public key to the server.
+    /// In mutual auth scenarios, we will also include an anti replay attack counter in the
     /// Noise handshake payload. Currently this counter is always a millisecond-
     /// granularity unix epoch timestamp.
-    pub async fn upgrade_outbound<TSocket>(
+    pub async fn upgrade_outbound_with_noise<TSocket>(
```
phlip9, line 194: why rename these? should be clear from the namespace...

network/netcore/src/transport/mod.rs
```diff
@@ -86,7 +87,7 @@ pub trait Transport {
         Self: Sized;
 
     /// Dials the given [`NetworkAddress`], returning a future for a pending outbound connection.
-    fn dial(&self, addr: NetworkAddress) -> Result<Self::Outbound, Self::Error>
+    fn dial(&self, peer_id: PeerId, addr: NetworkAddress) -> Result<Self::Outbound, Self::Error>
```
phlip9, line 90: i'm still not sure how i feel about this, since it doesn't really make sense for base transports. or for example the old peer-id exchange protocol. it really only makes sense for the libranet transport... this is asserting that we must always know the precise peer id for a remote peer for any transport, while in some hypothetical transports this could be exchanged in-band.

I suppose it's fine for now. we'll definitely have to revisit these decisions when/if we look at refactoring the Transport trait to support combinators again.

network/src/noise/handshake.rs
```diff
@@ -205,14 +213,19 @@ impl NoiseUpgrader {
             HandshakeAuthMode::ServerOnly => None,
         };
 
+        // send [self_peer_id | expected_public_key] as prologue
+        let mut prologue = self.self_peer_id.to_vec();
+        prologue.extend_from_slice(remote_public_key.as_slice());
+        socket.write_all(&prologue).await?;
```
phlip9, line 219: i wonder if it makes sense to combine prologue and first_message into a single buffer and send op? we use TCP_NODELAY so the two writes will probably get split...

network/src/noise/handshake.rs
```diff
@@ -248,42 +261,97 @@ impl NoiseUpgrader {
...
+    pub async fn upgrade_inbound_with_noise<TSocket>(
         &self,
         mut socket: TSocket,
-    ) -> io::Result<NoiseStream<TSocket>>
+    ) -> io::Result<(NoiseStream<TSocket>, PeerId)>
     where
         TSocket: AsyncRead + AsyncWrite + Unpin,
     {
+        // receive [remote_peer_id | self_public_key] as prologue
+        let mut prologue = vec![0; PeerId::LENGTH + x25519::PUBLIC_KEY_SIZE];
```
phlip9, line 272: don't think this needs to be a vec, can just be an array?

network/netcore/src/transport/mod.rs
mimoo: I see what you're saying, but I don't think we should build something too general. If we only have situation when we're dialing a specific peer, then we should use that information :o 

worst case we can make it an option in the future?

network/src/noise/handshake.rs
mimoo: reasons:

- we kept confusing these in discussions
- for the specification it's not super clean to have name collisions

in general I think it can be confusing in an implementation as well, but I don't have a strong enough opinion to fight it if you think we should rename it back :p

network/src/noise/handshake.rs
mimoo: good point!

network/src/noise/handshake.rs
mimoo: made the change for outbound as well!

network/src/validator_network/network_builder.rs
mimoo: ok I implemented the correct behavior now :)

```rust
        let (key, maybe_trusted_peers, peer_id) = match authentication_mode {
            // validator-operated full node
            AuthenticationMode::ServerOnly(key) if self.peer_id == PeerId::default() => {
                let public_key = key.public_key();
                let peer_id = PeerId::from_identity_public_key(public_key);
                (key, None, peer_id)
            }
            // full node
            AuthenticationMode::ServerOnly(key) => (key, None, self.peer_id),
            // validator
            AuthenticationMode::Mutual(key) => {
                (key, Some(self.trusted_peers.clone()), self.peer_id)
            }
        };
```

mimoo: cc @phlip9 I'm wondering why we don't just make the distinction between the three types of networks we have?

There's also an argument to make AuthenticationMode hold the peer_id as well.

network/src/noise/handshake.rs
phlip9: we can distinguish with `noise::upgrade_inbound` vs `transport::upgrade_inbound`? the reason why i mention this is because I think we can (in the future) extract this sort of "upgrade" structure into a trait for upgrading existing sockets (different from dial/listen which produce a new socket/stream of sockets). not sure what this would look like, but maybe

```rust
trait Upgrade {
    type Output;

    fn parse_outbound_addr(protos: &[Protocol]) -> Option<(/* addr args? */, &[Protocol])>;

    async fn upgrade_inbound(/* socket? addr args? */) -> Result<Self::Output>;

    async fn upgrade_outbound(/* socket? addr args? */) -> Result<Self::Output>;
}
```

and then NoiseUpgrader would impl `Upgrade` for example

network/netcore/src/transport/mod.rs
phlip9: right, like i said, in the interest of time i'm fine with adding the `PeerId` arg. i think `Transport` trait could definitely use some love, but we'll have to investigate that later : )

network/src/noise/handshake.rs
```diff
@@ -179,44 +187,54 @@ impl NoiseUpgrader {
...
+        // craft 8-byte payload as current timestamp (in milliseconds)
         let payload = {
             let now: u64 = time::SystemTime::now()
                 .duration_since(time::UNIX_EPOCH)
                 .expect("system clock should work")
-                .as_millis() as u64;
+                .as_millis() as u64; // (PAYLOAD_SIZE)
+
             // e.g. [157, 126, 253, 97, 114, 1, 0, 0]
             let now = now.to_le_bytes().to_vec();
```
phlip9, line 215: does not need `to_vec` can just return `to_le_bytes` which should just give us an array

network/src/noise/handshake.rs
```diff
@@ -179,44 +187,54 @@ impl NoiseUpgrader {
...
             Some(now)
```
phlip9, line 216: doesn't need to be Option now, can wrap it in `Some` when you call `initiate_connection`

network/src/noise/handshake.rs
```diff
@@ -245,42 +263,97 @@ impl NoiseUpgrader {
...
         mut socket: TSocket,
-    ) -> io::Result<NoiseStream<TSocket>>
+    ) -> io::Result<(NoiseStream<TSocket>, PeerId)>
     where
         TSocket: AsyncRead + AsyncWrite + Unpin,
     {
+        // receive [remote_peer_id | self_public_key] as prologue
+        let mut prologue = [0; PeerId::LENGTH + x25519::PUBLIC_KEY_SIZE];
+        socket.read_exact(&mut prologue).await?;
+        let (remote_peer_id, self_expected_public_key) = prologue.split_at(PeerId::LENGTH);
```
phlip9, line 276: ahh, same thing here. we should probably just read everything at once?

network/src/noise/handshake.rs
```diff
@@ -179,44 +187,54 @@ impl NoiseUpgrader {
...
         let initiator_state = self
             .noise_config
             .initiate_connection(
                 &mut rng,
-                &[],
+                &prologue,
                 remote_public_key,
                 payload.as_ref().map(|x| &x[..]),
                 &mut first_message,
             )
```
phlip9, line 234: i think something like this should work

```rust
const PROLOGUE_SIZE = PeerId::LENGTH + x25519::PUBLIC_KEY_SIZE;
const CLIENT_MESSAGE_SIZE = PROLOGUE_SIZE + noise::handshake_init_msg_len(PAYLOAD_SIZE);

// ..

let mut client_msg = [0u8; CLIENT_MESSAGE_SIZE];
client_msg[..PeerId::LENGTH].copy_from_slice(self.self_peer_id.as_ref());
client_msg[PeerId::LENGTH..PROLOGUE_SIZE].copy_from_slice(remote_public_key.as_slice());

let (prologue_msg, client_noise_msg) = client_msg.split_at_mut(PROLOGUE_SIZE);

let payload = {
    let now: u64 = time::SystemTime::now()
        .duration_since(time::UNIX_EPOCH)
        .expect("system clock should work")
        .as_millis() as u64; // (PAYLOAD_SIZE)

    // e.g. [157, 126, 253, 97, 114, 1, 0, 0]
    now.to_le_bytes()
};

let initiator_state = self
    .noise_config
    .initiate_connection(
        &mut rng,
        &prologue,
        remote_public_key,
        Some(&payload[..]),
        &mut client_noise_msg,
    )
// ..

socket.write_all(&client_msg).await?;
```

network/src/noise/handshake.rs
mimoo: less system calls I guess

network/src/noise/handshake.rs
mimoo: actually are you sure we needed one write_all since we only flushed at the end?

mimoo: alright done!

phlip9: 1. tcp flush is a no-op anyway : p (https://docs.rs/tokio/0.2.21/src/tokio/net/tcp/stream.rs.html#867-870)

2. like i said, we have TCP_NODELAY turned on. without TCP_NODELAY, the socket will wait for enough data to buffer before sending (according to [Nagle's algorithm](https://en.wikipedia.org/wiki/Nagle's_algorithm)). with TCP_NODELAY it'll just send the write immediately

network/src/noise/handshake.rs
mimoo: ait, renamed back

mimoo:
ping @phlip9 addressed your comments o.o

network/src/validator_network/network_builder.rs
```diff
@@ -414,15 +414,26 @@ impl NetworkBuilder {
             .take()
             .expect("Authentication Mode not set");
 
-        let (key, maybe_trusted_peers) = match authentication_mode {
-            AuthenticationMode::ServerOnly(key) => (key, None),
-            AuthenticationMode::Mutual(key) => (key, Some(self.trusted_peers.clone())),
+        let (key, maybe_trusted_peers, peer_id) = match authentication_mode {
+            // validator-operated full node
+            AuthenticationMode::ServerOnly(key) if self.peer_id == PeerId::default() => {
```
phlip9, line 419: it looks like `peer_id` should be `Option<PeerId>` in the config then? these comments also don't match up with my current mental model for the different node types. hopefully we can clarify this in tomorrow's call.

anyway, I'm fine putting this off. I'm partly blocked on this so let's just land it now?

PR #4353, opened 2020-06-08
[network][connmgr] Improve logging of peer address updates

phlip9:
Recently, while debugging some connectivity issues, it was noticeably
difficult to glean the current state of the discovery info in connectivity
manager.

This diff is a first step in trying to improve network debuggability by
ensuring that we actually log updates to the peer_addresses set (with
some small tweaks so we only log when it actually updates).

### DIFF
--- a/network/src/connectivity_manager/mod.rs
+++ b/network/src/connectivity_manager/mod.rs
@@ -391,6 +391,7 @@ where
                 // Keep track of if any peer's addresses have actually changed, so
                 // we can log without too much spam.
                 let mut have_any_changed = false;
+                let self_peer_id = self.self_peer_id.short_str();
 
                 for (peer_id, addrs) in address_map {
                     // Do not include self_peer_id in the address list for dialing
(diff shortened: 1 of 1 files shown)


network/src/connectivity_manager/mod.rs
```diff
@@ -473,6 +491,54 @@ fn log_dial_result(peer_id: PeerId, addr: NetworkAddress, dial_result: DialResul
     }
 }
 
+impl PeerAddresses {
```
zekun000, line 494: it seems the only need for this struct is to impl Display? I think it's proper to impl Deref for HashMap for this

phlip9: what about `as_inner` / `as_inner_mut`?  i think `Deref` trait is kinda hard to read and seems to break rust-analyzer autocomplete...

phlip9: hmm, actually I'll just use `.0` directly

network/src/connectivity_manager/mod.rs
```diff
@@ -376,21 +388,42 @@ where
...
+                    // to avoid pointless self-dials.
+                    if peer_id == self.self_peer_id {
+                        continue;
+                    }
+
+                    // Update peer's addresses
+                    let curr_addrs = self.peer_addresses.0.entry(peer_id).or_default();
+                    if curr_addrs.update(src, addrs) {
+                        // At least one peer's addresses have actually changed.
+                        have_any_changed = true;
```
davidiw, line 406: maybe we should just print the updated address? we could get a lot of noise by printing out all addresses, thoughts?

gregnazario: Yeah, would be best if we could collect the id -> id changes, and print just those.

phlip9: i think i'll just print both diffs and current state. there's two orthogonal issues: 1. quickly find out what the current state is 2. trace how that state changed over time. with just the diffs, it's super annoying trying to mentally sum the diffs to find the current state.

PR #4357, opened 2020-06-08
[move-prover] More specs for RegisteredCurrencies, LibraConfig, Assoc…

DavidLDill:
…iation

Property that initialization persists was added to Association.move

pragma verify = true for RegisteredCurrencies and LibraConfig.

RegisteredCurrencies is reasonably well-specified (I think).  I have some
concerns about how to make sure `initialize` cannot be called a second time,
overwriting existing registered currencies.

LibraConfig just has a few definitions (with properties to check them) for
use in RegisteredCurrencies.

Re-generated documentation.

## Motivation

Ongoing specification of Move stdlib modules.

yes.

## Test Plan

## Related PRs


language/stdlib/modules/RegisteredCurrencies.move
```diff
@@ -52,33 +58,62 @@ module RegisteredCurrencies {
 
     /// # Module specifications
 
+    /// **Note:** I am worried in this module about the possibility that `initialize` could
+    /// be called a second time (in another module), which would reset the
+    /// currency_codes vector. I've had trouble formulating a property that
+    /// catches that possibility in this module.
```
emmazzz, line 64: It seems that calling `initialize` a second time would lead to abortion, in `LibraConfig::publish_new_config_with_capability` since there is already a config under `config_account`. So `aborts_if is_initialized()` should work even without the newly added assertion in `initialize`?

language/stdlib/modules/LibraConfig.move
```diff
@@ -192,10 +192,10 @@ module LibraConfig {
...
     spec module {
 
-        // Verification is disabled because of a false error in signer, called
-        // from offer.  There are other problems that may be genuine, but we have to
-        // debug the previous first.
-        pragma verify = false;
+        /// Specifications of LibraConfig are very incomplete.  There are just a few
+        /// definitions that are used by RegisteredCurrencies
+
+        pragma verify = true;
```
emmazzz, line 198: I can't comment on line 208 so I'm commenting here. Line 208 should probably be changed to `spec_is_published  is the spec version of is_published<Config>`.

language/stdlib/modules/RegisteredCurrencies.move
DavidLDill: Yes, maybe you're right.  At one time, I thought T was being stored instead of LibraConfig::T<T>, and T is just a struct, not a resource, so I thought it could be overwritten.
Also, I think I tried aborts_if is_initialized() and couldn't verify it.  I'll go back and look at it more carefully.

DavidLDill: I checked this more carefully, and you're right.  I've removed code and specs based on this concern.

PR #4361, opened 2020-06-08
[crypto] Curve25519 dalek dependency update to new fork

moeziniaf:
fix duplicate dependency from fork


moeziniaf:
just ran cargo update after the toml change

davidiw:
Good enough for me

@zekun000 consensus now has depends on smol, cached -> async-std -> smol :/

@bors-libra r+

PR #4364, opened 2020-06-09
[cluster-test] Introduce Instance::{start, stop, util_cmd}

andll:
This diff introduces convenience methods on the Instance.

They only available in k8s right now, and calls to appropriate methods in ClusterSwarmKube.

This allows to have Instance as a self-sufficient entity and not to pass reference to swarm to every place where we need to manipulate instance state

Also removes legacy(and no longer working) Instance:run_cmd/scp and family

### DIFF
--- a/testsuite/cluster-test/src/cluster_swarm/mod.rs
+++ b/testsuite/cluster-test/src/cluster_swarm/mod.rs
@@ -19,7 +19,7 @@ pub trait ClusterSwarm {
     /// Runs the given command in a container against the given Instance
     async fn run(
         &self,
-        instance: &Instance,
+        k8s_node: &str,
         docker_image: &str,
         command: String,
         job_name: &str,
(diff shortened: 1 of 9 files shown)


testsuite/cluster-test/src/cluster_swarm/mod.rs
```diff
@@ -34,7 +34,7 @@ pub trait ClusterSwarm {
     ) -> Result<Instance>;
 
     /// Deletes a node from the ClusterSwarm
-    async fn delete_node(&self, instance: &Instance) -> Result<()>;
```
ankushagarwal, line 37: You just made this opposite change :D

andll: I know)
But now I wanted to get rid of usage of `Instance::k8s_node_name()`

PR #4367, opened 2020-06-09
[move scripts] create Move scripts "ABIs"

matbd:
## Motivation

Refactor the existing list of Move scripts in `transaction-builder` so that we can export the list as a data format.

Eventually, we wish to send the list to clients (somehow) and use for code generation.

## Test Plan

```
cargo test -p transaction-builder
```

### DIFF
--- a/language/move-core/types/src/language_storage.rs
+++ b/language/move-core/types/src/language_storage.rs
@@ -46,7 +46,7 @@ pub struct StructTag {
     pub address: AccountAddress,
     pub module: Identifier,
     pub name: Identifier,
-    // TODO: rename to "type_args"
+    // TODO: rename to "type_args" (or better "ty_args"?)
     pub type_params: Vec<TypeTag>,
 }
 
--- a/types/src/transaction/script.rs
+++ b/types/src/transaction/script.rs
@@ -9,6 +9,7 @@ use std::fmt;
 #[allow(dead_code)]
 pub const SCRIPT_HASH_LENGTH: usize = 32;
 
+/// Call a Move script.
 #[derive(Clone, Hash, Eq, PartialEq, Serialize, Deserialize)]
 pub struct Script {
     code: Vec<u8>,
@@ -48,5 +49,99 @@ impl fmt::Debug for Script {
             .field("code", &hex::encode(&self.code))
             .field("args", &self.args)
             .finish()
+    }
+}
+
+/// How to call a particular Move script (aka. an "ABI").
+#[derive(Clone, Debug, Hash, Eq, PartialEq, Serialize, Deserialize)]
+pub struct ScriptABI {
+    /// The public name of the script.
+    name: String,
+    /// Some text comment.
+    doc: String,
+    /// The `code` value to set in the `Script` object.
+    code: Vec<u8>,
+    /// The names of the type arguments.
+    ty_args: Vec<TypeArgumentABI>,
+    /// The description of regular arguments.
+    args: Vec<ArgumentABI>,
+}
+
+/// The description of a (regular) argument in a script.
+#[derive(Clone, Debug, Hash, Eq, PartialEq, Serialize, Deserialize)]
+pub struct ArgumentABI {
+    /// The name of the argument.
+    name: String,
+    /// The expected type.
+    /// In Move scripts, this does contain generics type parameters.
+    type_tag: TypeTag,
+}
+
+/// The description of a type argument in a script.
+#[derive(Clone, Debug, Hash, Eq, PartialEq, Serialize, Deserialize)]
+pub struct TypeArgumentABI {
+    /// The name of the argument.
+    name: String,
+}
+
+impl ScriptABI {
+    pub fn new(
+        name: String,
+        doc: String,
+        code: Vec<u8>,
+        ty_args: Vec<TypeArgumentABI>,
+        args: Vec<ArgumentABI>,
+    ) -> Self {
+        Self {
+            name,
+            doc,
+            code,
+            ty_args,
+            args,
+        }
+    }
+
+    pub fn name(&self) -> &str {
+        &self.name
+    }
+
+    pub fn doc(&self) -> &str {
+        &self.name
+    }
+
+    pub fn code(&self) -> &[u8] {
+        &self.code
+    }
+
+    pub fn ty_args(&self) -> &[TypeArgumentABI] {
+        &self.ty_args
+    }
+
+    pub fn args(&self) -> &[ArgumentABI] {
+        &self.args
+    }
+}
+
+impl ArgumentABI {
+    pub fn new(name: String, type_tag: TypeTag) -> Self {
+        Self { name, type_tag }
+    }
+
+    pub fn name(&self) -> &str {
+        &self.name
+    }
+
+    pub fn type_tag(&self) -> &TypeTag {
+        &self.type_tag
+    }
+}
+
+impl TypeArgumentABI {
+    pub fn new(name: String) -> Self {
+        Self { name }
+    }
+
+    pub fn name(&self) -> &str {
+        &self.name
     }
 }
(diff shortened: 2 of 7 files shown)


types/src/transaction/script.rs
```diff
@@ -50,3 +51,57 @@ impl fmt::Debug for Script {
...
+    name: String,
+    /// Some text comment.
+    doc: String,
+    /// The `code` value to set in the `Script` object.
+    code: Vec<u8>,
+    /// The names of the type arguments.
+    ty_args_names: Vec<String>,
+    /// The names and expected types of the regular arguments.
+    /// Note: by construction, these types never depend on the type arguments.
+    args_names_and_types: Vec<(String, TypeTag)>,
```
thefallentree, line 68: is there an type-erasure happening here? to construct actual arguments we need to know the actual sub-type of TypeTag, right? 

plus, having this generic cotnainer inmplementation seems an overkill, I would settle to just have macro constrctut an struct and we can change the ScriptAPIView into an enum that reference these script struct manually.

matbd: Let me add some example(s) tomorrow to illustrate what the generated code is going to look like on the client side.

types/src/transaction/script.rs
```diff
@@ -50,3 +51,80 @@ impl fmt::Debug for Script {
...
+
+    pub fn code(&self) -> &[u8] {
+        &self.code
+    }
+
+    pub fn ty_args(&self) -> &[String] {
+        &self.ty_args
+    }
+
+    pub fn argss(&self) -> &[ArgumentABI] {
```
sblackshear, line 113:
```suggestion
    pub fn args(&self) -> &[ArgumentABI] {
```

types/src/transaction/script.rs
```diff
@@ -50,3 +51,80 @@ impl fmt::Debug for Script {
...
+#[derive(Clone, Debug, Hash, Eq, PartialEq, Serialize, Deserialize)]
+pub struct ScriptABI {
+    /// The public name of the script.
+    name: String,
+    /// Some text comment.
+    doc: String,
+    /// The `code` value to set in the `Script` object.
+    code: Vec<u8>,
+    /// The names of the type arguments.
+    ty_args: Vec<String>,
```
sblackshear, line 65: It's probably worth having a `Vec<TypeArgumentABI>` here, where `TypeArgumentABI` is something like

```
pub struct TypeArgumentABI  {
   /// The name of the type argument
   name: String,
   /// The kind of the type argument: All, Copyable, or Resource. Could also use an enum instead of referencing the file_format:Kind
   kind_constraint: file_format::Kind
}
```

sblackshear: Curious to hear what @vgao1996 / @dariorussi think here

types/src/transaction/script.rs
matbd: good catch!

types/src/transaction/script.rs
matbd: Sure, I could do that -- assuming that clients can indeed use the extra information. On this note, we will need to document type checking of type arguments (TypeTag vs Kind) and regular arguments (TransactionArgument vs TypeTag), so that clients know what to do.

p.s. TypeTag vs Kind: are there cases where we expect an infinite number of TypeTags? Otherwise, how about passing an enum of acceptable TypeTag values instead of kind?

vgao1996: Yeah "ty_args: Vec<String>" seems a little bit weird to me and I'd expect "file_format::Kind" here instead. Also I'd use "ty_params" instead of "ty_args" here because this is a declaration. I'm unsure if we could have a `name: String` in `TypeArgumentABI ` because such information is not available in the file format. Do we have such annotations elsewhere?

types/src/transaction/script.rs
```diff
@@ -50,3 +51,80 @@ impl fmt::Debug for Script {
...
+    /// The description of regular arguments.
+    args: Vec<ArgumentABI>,
+}
+
+/// The description of a script argument.
+#[derive(Clone, Debug, Hash, Eq, PartialEq, Serialize, Deserialize)]
+pub struct ArgumentABI {
+    /// The name of the argument.
+    name: String,
```
vgao1996, line 74: Same as above, I'm not sure if the name will always be available.

types/src/transaction/script.rs
```diff
@@ -50,3 +51,80 @@ impl fmt::Debug for Script {
...
+    /// The public name of the script.
+    name: String,
+    /// Some text comment.
+    doc: String,
+    /// The `code` value to set in the `Script` object.
+    code: Vec<u8>,
+    /// The names of the type arguments.
+    ty_args: Vec<String>,
+    /// The description of regular arguments.
+    args: Vec<ArgumentABI>,
```
vgao1996, line 67: `params` instead of `args`?

types/src/transaction/script.rs
matbd: It is available now and we need it on the client side.

types/src/transaction/script.rs
matbd: I agree but this ship has sailed: the other definitions use `args` consistently. I think I even saw a TODO to remove the last `param`

types/src/transaction/script.rs
sblackshear: It's true that we can't scrape these names from bytecode (if we go that route in the future), but seems nice to provide them if we're starting from Rust instead.

vgao1996: That's interesting. I'm not that familiar with the background but I assume this ABI is generated from the Move source files rather than the binary file format?

types/src/transaction/script.rs
vgao1996: Well still I think the right thing to do is to replace all other `args` with `params`. These two terms have clear meanings and are not really interchangeable in the PL context... with that being said I totally understand this is not a high pri so please don't feel obligated to correct it in this PR.

(To clarify, by "other definitions" you mean things outside `language` right? AFAIK all except 1 or 2 definitions in language use the correct term.)

matbd:
I'm going to postpone the JSON RPC endpoint because Sam and I just discussed some alternative ideas that may be more future-proof.

types/src/transaction/script.rs
dariorussi: here I am assuming the `ty_args` refer to the generic type the script is providing so for `main<T>() {...}` it's the `T`. Am I correct?
Under that assumption it's true that the `Kind` is the only information needed by a client, and we can have an enum for that (whether using the `Kind` enum as is or defining a new one).
The name of the generic as in @sblackshear's `TypeArgumentABI` is an option if we wanted to do that.
The `String` really gives you nothing as the only info that is useful is the restrictions over the generic types.

types/src/transaction/script.rs
matbd: Sam convinced me today that these ABIs will likely (1) not be stored onchain (2) be managed by the author of the source code (and made available in the same adhoc way).

dariorussi: @vgao1996: it's hard coded now in the description of the script as in files above

types/src/transaction/script.rs
matbd: Other definitions in `types/src/transactions`.

language/move-core/types/src/language_storage.rs
```diff
@@ -46,7 +46,7 @@ pub struct StructTag {
     pub address: AccountAddress,
     pub module: Identifier,
     pub name: Identifier,
-    // TODO: rename to "type_args"
```
matbd, line 49: @vgao1996

types/src/transaction/script.rs
dariorussi: fwiw, @vgao1996 is referring to what the binary format uses and what we have tried to move the code base to. And by codebase we mean the one in language. 
So in that respect, if the intent was consistency with the `language` part of the code `parameters` is what we went for. 
Of course even parts within `language` are not consistent yet

types/src/transaction/script.rs
matbd: FYI I'm landing the PR with this definition to begin with:
```
pub struct TypeArgumentABI  {
   /// The name of the type argument
   name: String,
}
```

PR #4368, opened 2020-06-09
[storage] Replace the Box<dyn Storage> runtime shim by a static enum

huitseeker:
We use an open trait hierarchy for `CryptoStorage + KVStorage`, which we represent with a(n emulated) trait-alias `Storage`, which carries signing and verification APIs through the `CryptoStorage` component of the trait alias.
We are also keen to be able to instantiate the implementations of this hierarchy through a factory method taking a configuration as an input, speficically a `SecureBackend`, which is a closed sum type (an `enum`). Hence this conversion must return trait objects (specifically `Box<dyn Storage>`), and crucially relies on this dispatch.

This has two consequences:
- since trait objects cannot include generic methods, the sign and verify methods of `CryptoStorage` and all its dependents (`CryptoHash` in particular) would have to take `v: &dyn Cryptohash` arguments instead of `v: CryptoHash`,
- this reliance on dynamic dispatch is slow: I measure about 2000 microseconds spent on this dynamic dispatch, against 100 microseconds for actually verifying a signature afterwards.

The present PR re-establishes static dispatch through a mirror enum, which instantiates a closed sum type over the inhabitants of the `KVStorage + CryptoStorage` hierarchy. That's what the conversion from `SecureBackend` returns. It should offer better performance, and allow `CryptoStorage` to offer sign & verify APis that take a `x: CryptoHash + Storage`.

Details:
- Don't worry about the `BoxedStorage` name, I'll modify it programmatically to `Storage` (o.b.o.), it's meant to help review temproraily.
- The enum and all its contents can thankfully all be generated automatically, which should help with the maintenance of the boilerplate in `libra_secure_storage::storage`. A good starting point to do that is the [enum_dispatch](https://gitlab.com/antonok/enum_dispatch) crate, which lacks two things in order to work for us as is:
  + support for implementation of multiple traits,
  + support for a `Box<dyn T>` parameter for an enum variant, because ...
- The KVStorage + CryptoStorage hierarchy contains a recursive variant (NamespacedStorage), which obviously complicates the enum definition a bit (the conversion from SecureBackend would not know how to return a `BoxedStorage<T>` without knowing `T`), and hence requires me to use a (slow) `dyn Trait` in the `BoxedStorage` destination enum.
- I've removed the bespoke trait alias emulation performed to equate `Storage` with `KVStorage + CryptoStorage` because I expect macro-powered generation `impl`s of an enum closing over a trait hierearchy to be simpler if the traits one must implement on the enum are direct requirements in client code.


secure/storage/src/storage.rs
```diff
@@ -10,68 +14,310 @@ use libra_crypto::{
...
-pub trait Storage: KVStorage + CryptoStorage {}
-
-impl<T> Storage for T where T: KVStorage + CryptoStorage {}
 
-/// This is a hack that allows us to convert from SecureBackend into a useable T: Storage, since
-/// Box<dyn Storage> is not T: Storage.
-pub struct BoxStorage(pub Box<dyn Storage>);
+/// This is a hack that allows us to convert from SecureBackend into a useable
+/// T: Storage. This boilerplate can be 100% generated by a proc macro.
+/// TODO(fga): generate this boilerplate with a proc macro.
```
matbd, line 20: In this precise case, I feel that we could go a long way with regular macros (*).

(*) We may have to go 2nd-order, though:
https://www.reddit.com/r/rust/comments/dopgqi/higher_order_macros/
https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=51143f2e22e1635a0361310aaa21f4ac

secure/storage/src/namespaced_storage.rs
```diff
@@ -32,8 +32,8 @@ impl<T: KVStorage> KVStorage for NamespacedStorage<T> {
     }
 }
 
-impl<T> NamespacedStorage<T> {
-    pub fn new(storage: T, namespace: String) -> Self {
+impl NamespacedStorage {
+    pub fn new(storage: Box<dyn KVStorage>, namespace: String) -> Self {
```
davidiw, line 36: we could perhaps make new<T: KVStorage> and then box internally ... not sure one is better than the other. Thoughts?

secure/storage/src/storage.rs
davidiw: more precisely, enum_dispatch?

secure/storage/src/storage.rs
```diff
@@ -10,68 +14,310 @@ use libra_crypto::{
...
+            }
+            BoxedStorage::InMemoryStorage(_v) => {
+                Err("Tried to convert variant InMemoryStorage to OnDiskStorage")
+            }
+            BoxedStorage::NamespacedStorage(_v) => {
+                Err("Tried to convert variant NamespacedStorage to OnDiskStorage")
+            }
+        }
+    }
+}
```
davidiw, line 152: Are these even being used?

secure/storage/src/storage.rs
```diff
@@ -10,68 +14,310 @@ use libra_crypto::{
...
 
-impl KVStorage for BoxStorage {
+impl KVStorage for BoxedStorage {
+    #[inline]
     fn available(&self) -> Result<(), Error> {
-        self.0.available()
+        match self {
+            BoxedStorage::GitHubStorage(inner) => KVStorage::available(inner),
```
davidiw, line 158: what's the advantage of this format versus
```inner.available()```

secure/storage/src/namespaced_storage.rs
huitseeker: That's impossible: if `NamespacedStorage` takes a generic parameter (i.e. recursion), then so does `BoxedStorage`. Then the `impl From<SecureBackend> for BoxedStorage` becomes an `impl<T> From<SecureBackend> for BoxedStorage<T>`, but returning this `<T>` from an unknown runtime content is not possible / nonsensical.

secure/storage/src/storage.rs
huitseeker: Not today.

secure/storage/src/storage.rs
huitseeker: Just fully-qualified code that's impervious to any wandering method the underlying concrete type (e.g. GithubStorage ...) might implement.

secure/storage/src/storage.rs
matbd: +1 to update the comment

secure/storage/src/storage.rs
davidiw: Personally, I like minimal interfaces, and I thought you did too -- given your previous set of PRs. Could we remove them?

Issue #1863, opened 2019-11-27
[cluster-test] Migrate debug_interface_log_tail to async api


andll:
Currently debug_interface_log_tail spawns dedicated thread for each node.
Instead of doing that we can switch to using async grpc API and do all work in main tokio Runtime

Code pointer: `debug_interface_log_tail.rs`

KiChjang:
I'm interested in trying to solve this, but I am completely dumb on async Rust. I can't seem to figure out which part of the code needs to be rewritten to use `tokio::spawn`... or perhaps we shouldn't even be using that at all and reuse `DebugPortLogThread.client.client()` instead?

andll:
The way I would see it is to make `DebugPortLogThread::run` to be async function, and spawn it from `DebugPortLogThread::spawn_new`.
This `spawn_new` is called from `ClusterTestRunner::setup`, that has tokio runtime - you can just pass it from there to `spawn_new`

mrbuzz:
I'd like to pick up this issue if no one is actively working on it.

PR #2844, opened 2020-03-08
[cluster-test] Migrate debug_interface_log_tail to async api

mrbuzz:
## Motivation
See issue [1863](https://github.com/libra/libra/issues/1863)
Currently DebugPortLogThread spawns a dedicated thread for each node
this change is for switching to the async api and use the tokio runtime
instead ot std::threads

### Have you read the [Contributing Guidelines on pull requests](https://github.com/libra/libra/blob/master/CONTRIBUTING.md#pull-requests)?

Yes

## Test Plan

Just ran the entire test suite and linters, not sure if any other test is needed.

## Related PRs

None

PR #4369, opened 2020-06-09
[cluster-test] Migrate debug_interface_log_tail to async api

changyonghu:
Closes: #1863
Referred #2844 
Manually tested transaction generator run and health check run in swarm mode and ./scripts/cti --pr <PR> --run bench in normal mode.

### DIFF
--- a/common/debug-interface/src/lib.rs
+++ b/common/debug-interface/src/lib.rs
@@ -59,3 +59,55 @@ impl NodeDebugClient {
         Ok(response.json()?)
     }
 }
+
+/// Implement default utility client for AsyncNodeDebugInterface
+pub struct AsyncNodeDebugClient {
+    client: reqwest::Client,
+    addr: String,
+}
+
+impl AsyncNodeDebugClient {
+    /// Create AsyncNodeDebugInterface from a valid socket address.
+    pub fn new<A: AsRef<str>>(client: reqwest::Client, address: A, port: u16) -> Self {
+        let addr = format!("http://{}:{}", address.as_ref(), port);
+
+        Self { client, addr }
+    }
+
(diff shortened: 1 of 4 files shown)


common/debug-interface/src/lib.rs
```diff
@@ -59,3 +59,56 @@ impl NodeDebugClient {
...
+
+/// Implement default utility client for AsyncNodeDebugInterface
+pub struct AsyncNodeDebugClient {
+    client: reqwest::Client,
+    addr: String,
+}
+
+impl AsyncNodeDebugClient {
+    /// Create AsyncNodeDebugInterface from a valid socket address.
+    pub fn new<A: AsRef<str>>(address: A, port: u16) -> Self {
```
andll, line 71: Can we pass reqwest::Client here? It is a good practice to reuse same Client to save resources

changyonghu: I don't quite get it. Where is the same Client from? Like we check if there is any client for this IP address first?

andll: We just want to one Client instance and pass this (cloned) instance into each  AsyncNodeDebugClient , as an example you can look at how `Cluster::from_host_port` shares single Client across all Instance.

[As a note, cloning Client only clones 'handle', but resources are shared]

You don't need to check one client per IP or smth like that.

changyonghu: Can we reuse the http Client in Cluster for this purpose? I don't know what is the http client used for?

andll: Yes, we ca reuse http client from cluster, or create a new one, just don't want to create O(N) http clients

andll:
Great job!

PR #4370, opened 2020-06-09
[cluster-test] Cleanup ClusterSwarm trait

andll:
This diff clean ups duplication methods between ClusterSwarm and Instance.

ClusterSwarm trait no longer have methods to upsert/delete/run methods, as they are replaced with Instance::{start, stop, util_cmd}.

Instead, ClusterSwarm has `spawn_new_instance` and the family.

upsert/delete/run are moved out of ClusterSwarm trait into ClusterSwarmKube.

If we have some other implementation for ClusterSwarm, we'll need to update Instance methods to handle those(we can make InstanceBackend a trait for example, or smth like that).


ankushagarwal:
@bors-libra r+
Yet another merge conflict for me :-/

PR #4371, opened 2020-06-09
[Key Manager] Retrieve operator account from secure storage.

JoshLind:
## Motivation

The Key Manager requires the account address of the validator operator for which it is responsible for rotating keys. At present, this is retrieved using the Key Manager config. However, this information now exists in secure storage (see https://github.com/libra/libra/pull/4181). As such, this PR updates the Key Manager to retrieve the validator operator account from secure storage.

Yes.

## Test Plan

All local tests pass. Moreover, a local run of the Key Manager docker container shows it behaves as expected.

## Related PRs

This PR addresses the TODO set by https://github.com/libra/libra/pull/4181


secure/key-manager/src/lib.rs
```diff
@@ -158,7 +155,8 @@ where
 
     pub fn compare_storage_to_config(&self) -> Result<(), Error> {
         let storage_key = self.storage.get_public_key(CONSENSUS_KEY)?.public_key;
-        let validator_config = self.libra.retrieve_validator_config(self.account)?;
+        let operator_account = self.get_operator_account()?;
```
davidiw, line 158: get versus retrieve :O

secure/key-manager/src/lib.rs
```diff
@@ -288,6 +293,19 @@ where
...
+        match self.storage.get(OPERATOR_ACCOUNT) {
+            Ok(response) => match response.value.string() {
+                Ok(account_address) => match AccountAddress::from_str(&account_address) {
+                    Ok(account_address) => Ok(account_address),
+                    Err(e) => Err(Error::UnknownError(e.to_string())),
+                },
+                Err(e) => Err(Error::MissingAccountAddress(e)),
+            },
+            Err(e) => Err(Error::MissingAccountAddress(e)),
+        }
```
davidiw, line 307: Have you played aorund with map or and_then?

JoshLind: Thanks, David! I managed to simplify it a little (removed one match statement), but can't remove the inner match statement because of the different error type.. it's also unnecessarily ugly because we can't convert directly from anyhow.. but such is life!

PR #4373, opened 2020-06-09
[Secure Storage] Simplify import_private_key() logic.

JoshLind:
## Motivation

Policy support was recently removed from the secure storage backends and the semantics of create vs. set was simplified. As such, when we implement create_key() we no longer need to immediately initialize the previous key version. This version will later be set when the first rotate() is called. This small change simplifies a few tests and avoids having to leak internal secure storage state.

Yes.

## Test Plan

All local tests still pass.

## Related PRs

None.


JoshLind:

PR #4376, opened 2020-06-09
Replace add_to_network(network_builder) with a dedicated call in network_builder.

agc-quadra:
## Motivation

Broad goal at present is to cleanup and organizer main_node.rs and network_builder.rs

This PR is an intermediate step which will allow moving setup_network out of main_node.rs and into network_builder.

A core problem in the current paradigm of XXX::add_to_network(&mut network_builder) is that each component XXX  has an explicit dependency on the network_builder crate, so setup_network cannot be a part of network_builder without creating a circular dependency.

This change sidesteps this problem by taking away the explicit dependency from every component onto network_builder.

yes
## Test Plan

Don't break existing integration and unit tests.

## Related PRs

### DIFF
--- a/consensus/src/network_interface.rs
+++ b/consensus/src/network_interface.rs
@@ -12,6 +12,7 @@ use consensus_types::{
     sync_info::SyncInfo,
     vote_msg::VoteMsg,
 };
+use libra_metrics::IntCounterVec;
 use libra_types::{epoch_change::EpochChangeProof, PeerId};
 use network::{
     error::NetworkError,
@@ -20,7 +21,7 @@ use network::{
         network::{NetworkEvents, NetworkSender, NewNetworkSender},
         rpc::error::RpcError,
     },
-    validator_network::network_builder::{NetworkBuilder, NETWORK_CHANNEL_SIZE},
+    validator_network::network_builder::NETWORK_CHANNEL_SIZE,
     ProtocolId,
 };
 use serde::{Deserialize, Serialize};
@@ -69,19 +70,21 @@ pub struct ConsensusNetworkSender {
     network_sender: NetworkSender<ConsensusMsg>,
 }
 
-/// Create a new Sender that only sends for the `CONSENSUS_DIRECT_SEND_PROTOCOL` and
-/// `CONSENSUS_RPC_PROTOCOL` ProtocolId and a Receiver (Events) that explicitly returns only said
-/// ProtocolId.
-pub fn add_to_network(
-    network: &mut NetworkBuilder,
-) -> (ConsensusNetworkSender, ConsensusNetworkEvents) {
-    network.add_protocol_handler((
+/// Configuration for the network endpoints to support consensus.
+pub fn network_endpoint_config() -> (
+    Vec<ProtocolId>,
+    Vec<ProtocolId>,
+    QueueStyle,
+    usize,
+    Option<&'static IntCounterVec>,
+) {
+    (
         vec![ProtocolId::ConsensusRpc],
         vec![ProtocolId::ConsensusDirectSend],
         QueueStyle::LIFO,
         NETWORK_CHANNEL_SIZE,
         Some(&counters::PENDING_CONSENSUS_NETWORK_EVENTS),
-    ))
+    )
 }
 
 impl NewNetworkSender for ConsensusNetworkSender {
--- a/consensus/src/network_interface.rs
+++ b/consensus/src/network_interface.rs
@@ -17,7 +17,7 @@ use network::{
     error::NetworkError,
     peer_manager::{ConnectionRequestSender, PeerManagerRequestSender},
     protocols::{
-        network::{NetworkEvents, NetworkSender},
+        network::{NetworkEvents, NetworkSender, NewNetworkSender},
         rpc::error::RpcError,
     },
     validator_network::network_builder::{NetworkBuilder, NETWORK_CHANNEL_SIZE},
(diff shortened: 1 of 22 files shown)


agc-quadra:
@davidiw This is a follow on to #4246 which should not be considered until the previous PR is resolved.   Otherwise the scope becomes difficult to review.

agc-quadra:
@davidiw This PR is a coherent milestone along the path towards builder cleanup.   #4384 is the immediate extension.

agc-quadra:
@davidiw 

PTAL and let me know if you feel this PR is scoped where the context is sufficient for review;  if this is too narrow/too much context is missing, then let's consider #4384 instead.

I'll move whichever is more appropriate out of draft.

consensus/src/network_interface.rs
```diff
@@ -69,38 +70,37 @@ pub struct ConsensusNetworkSender {
...
+    Option<&'static IntCounterVec>,
+) {
     (
-        ConsensusNetworkSender::new(network_sender, connection_reqs_tx),
-        ConsensusNetworkEvents::new(network_receiver, connection_notifs_rx),
+        vec![ProtocolId::ConsensusRpc],
+        vec![ProtocolId::ConsensusDirectSend],
+        QueueStyle::LIFO,
+        NETWORK_CHANNEL_SIZE,
+        Some(&counters::PENDING_CONSENSUS_NETWORK_EVENTS),
```
davidiw, line 86: so my vision for this was to actually define a trait with a set of consts or functions if need be... what are your thoughts? I think that gives us a little more flexibility / strictness.

FWIW, this is still better.

network/src/protocols/network/mod.rs
```diff
@@ -95,8 +95,16 @@ pub struct NetworkEvents<TMessage> {
...
 
-impl<TMessage: Message> NetworkEvents<TMessage> {
-    pub fn new(
+/// Trait specifying the signature for `new()` `NetworkEvents`
+pub trait NewNetworkEvents {
+    fn new(
+        peer_mgr_notifs_rx: libra_channel::Receiver<(PeerId, ProtocolId), PeerManagerNotification>,
+        connection_notifs_rx: libra_channel::Receiver<PeerId, ConnectionNotification>,
+    ) -> Self;
+}
```
davidiw, line 104: Its unclear to me what the advantage of this is... maybe we could unify this with the "network_endpoint_config" ?

consensus/src/network_interface.rs
agc-quadra: We could definitely wrap this into a trait definition instead of the method.   

I was partly driven by two things:
1). keep the change limited in scope so it's clear that it is an incremental step
2). breaking the explicit dependency from the various components (consensus, onchaindiscovery, statesync, ....)  on NetworkBuilder.

In hindsight and in the context of #4440 , we can define the trait in `network/protocol/network`.   That wasn't clear at the time.  I'll suggest taking that step once #4384 is merged*.


* #4384 is the immediate follow up to this PR and addresses the key project level structural change.

network/src/protocols/network/mod.rs
agc-quadra: This enables the generic `add_protocol_handler` in network_builder to return the `(Sender, Receiver)` pair rather than the underlying set of channels.

In the old state of affairs, each call to `add_to_network` was converting the 4 underlying channels into Sender/Receiver pair.  The current state of affairs has that this step being done entirely within add_to_protocol, but relies on having the common trait for constructing the specific derived builders.

Putting this into network_endpoing_config would require network_endpoint_config to invoke network_builder to get the channel information;  this is explicitly the dependency link that needs to be broken in order to get setup_network out of main_node.

davidiw: I guess I was thinking something like...

```
NetworkBuilder::add_service<T: NetworkService>(T: NetworkService)
```

Then we can initiate StateSync as a NetworkService, it can in turn have the following API

```
const NetworkService::RPC_PROTOCOLS: vec<PROTOCOL_ID> = ...;
...
NetworkService::initialize(peer_mgr_notifs_rx, connection_notifs_rx)
```

Then all you would do is in main:

```
let state_sync = StateSync::new();
network_builder::add_service(&state_sync);
```

note the above is super bad Rust code, but I hope it explains the idea.

davidiw:
So I didn't take a look at your references to future PRs, so I'll do that to make sense of what's going on here. I think your rebase hell should ideally be minimal -- though there were some recent tweaks recently.

network/src/protocols/network/mod.rs
agc-quadra: That's much more clear. It aligns with where I want to get to;  unfortunately the path between here and there is windy and potentially long.   Either I run through an eternal rebase hell or we go through a number of intermediate states which move the system sideways in order to unblock forward progress.

Key intermediate milestones:

1. Getting `setup_network` into `network_builder`;  this is blocked by the current state of circular dependencies wrapped up in the current `XXX::add_to_network` pattern
1. Applying the pattern (and hopefully NetworkService trait) internally within `setup_network/create_network` for the internal network_components.  This is hard blocked by `setup_network` being outside of `network_builder`
1. Detangling the various construction dependencies; right now `StateSync::new` takes parameters that are produced by calling `network_builder::add_service(...)`.   Progress here is blocked by the poor construction/internals of `setup_network`
1.  (potential) decoupling the configuration/build/start phases of network_builder (and by extension of node_builder, consensus_builder, state_sync_builder, mempool_builder, etc.)

Of the concerns above, the biggest chunk of work to getting to the trait you're asking for is dealing with (2) and (3).   Making cohesive progress on those two steps requires getting setup_network into network_builder.   Also, making that pattern work on the internal components of network_builder is an important scope-limiting proof of concept before getting sucked into the broader system.

#4384 is the immediate follow up to this PR in order to get setup_network out of main_node and into network_builder.

agc-quadra: In terms of the end-state target, I'm expecting main_node, network_builder, (and possibly consensus/state_sync/Mempool) to  look similar to the following four phases:

*1*:  Configure the components.
```
  let (network) = NetworkBuilder::Create(node_config.NetworkConfig);
  let (consensus) = ConsensusBuilder::Create(node_config.ConsensusConfig);
  let (state_sync) = StateBuilder::Create(node_config.StateSyncConfig);
```

*2* Create wires
```
   let consensus_wires = network.add_service(consensus);
   let state_sync_wires = network.add_service(state_sync);
   consensus_wires.append(state_sync.add_service(consensus));
```

*3*:  Build the component objects with the connecting wires.
```
  network.build(incoming_wires);
  consensus.build(consensus_wires);
  state_sync.build(sate_sync_wires);
```

*4*:  start the services
```
   let executor = Runtime::new();
   network.start(&executor);
   stateysnc.start(&executor);
```

For network builder this would translate to:
```
impl NetworkBuilder {
  // PHASE 1
  pub fn create(NetworkConfig) {
      self.peer_manager_config = peer_manager_config_from_network_config();
      self.connectivity_manager_config = peer_manager_config_from_network_config();
      ...
  }

  // PHASE 2
  add_service(NetworkService) -> (SenderChannel, EventChannel) {
      ...
  }

  // PHASE 3
  pub fn build() {
     if self.peer_manager_config.exists() {
        build_peer_manager_from(self.peer_manager_config);
     }
     if self.connectivity_manager_config.exists() {
        build_connectivity_manager_from(self.connectivity_manager_config);
     }
     ...
  }

  // PHASE 4
  pub fn start(&runtime) {
    runtime.spawn(peer_manager);
    runtime.spawn(health_checker);
    runtime.spawn(connectivity_manager);
    ...
  }
}
```

Related to this structure, much of the complexity of network_builder can be traced back to the fact that it is about 90% PeerManager with about 10% pasting things onto that core.

agc-quadra:
> So I didn't take a look at your references to future PRs, so I'll do that to make sense of what's going on here. I think your rebase hell should ideally be minimal -- though there were some recent tweaks recently.

The rebase hell is mostly associated with two points in the immediate follow up #4384:
1.  Move `network_builder` from `network/src/validator_network/` to `network/builder/src/`
2.  Move `setup_network` from `main_node` to `network_builder`

cross file structural changes are tediously complicated.

davidiw:
Closing out as this is included in #4384

PR #4384, opened 2020-06-09
Move setup_network out of main_node

agc-quadra:
## Motivation

Move setup_network out of main_node to clean up main and consolidate the network construction logic.

An important side piece of this change is to move network_builder into its own crate in order to eliminate a dependency loop.

This change sets the groundwork for being able to directly simplify the setup_network logic in a focused and narrow scope.

yes

## Test Plan

Don't break existing unit and integration tests.

## Related PRs

### DIFF
--- a/network/src/constants.rs
+++ b/network/src/constants.rs
@@ -1,0 +1,24 @@
+// Copyright (c) The Libra Core Contributors
+// SPDX-License-Identifier: Apache-2.0
+
+/// A collection of constants and default values for configuring various network components.
+
+// NB: Almost all of these values are educated guesses, and not determined using any empirical
+// data. If you run into a limit and believe that it is unreasonably tight, please submit a PR
+// with your use-case. If you do change a value, please add a comment linking to the PR which
+// advocated the change.
+// TODO:  Each of these should be commented with semantic meaning, intended use place, and justification for the value.
+// TODO:  Better --- these should be encapsulated in configurations somewhere.
+pub const NETWORK_CHANNEL_SIZE: usize = 1024;
+pub const DISCOVERY_INTERVAL_MS: u64 = 1000;
+pub const PING_INTERVAL_MS: u64 = 1000;
+pub const PING_TIMEOUT_MS: u64 = 10_000;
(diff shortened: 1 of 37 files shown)


agc-quadra:
@davidiw More steps along the way towards cleaning up network_builder.  This is a step which follows from #4376

agc-quadra:
@davidiw This extends #4376  with the next stage to move setup_network into network_builder.

25 of the updated files are inherited.

network/src/constants.rs
```diff
@@ -0,0 +1,24 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+/// A collection of constants and default values for configuring various network components.
+
+// NB: Almost all of these values are educated guesses, and not determined using any empirical
+// data. If you run into a limit and believe that it is unreasonably tight, please submit a PR
+// with your use-case. If you do change a value, please add a comment linking to the PR which
+// advocated the change.
+// TODO:  Each of these should be commented with semantic meaning, intended use place, and justification for the value.
+// TODO:  Better --- these should be encapsulated in configurations somewhere.
```
davidiw, line 11: my first thought...

agc-quadra: These constants should absolutely *NOT* be constants and need to be provided through configuration so the values can be operationally changed.  The PRs coming through in the last week to change these for production reasons are red flags.

This is a problem that exists in the code today;  a side effect of this PR is to make the existence of that problem much more visible and explicit.  Fixing it is orthogonal to reworking node setup/XXX_builder and can be done (a) on the side or (b) after the fact;  it should not be inserted as a blocking step to cleaning the builder mechanics.  In general, I believe that the builder refactoring will make it very clear which configuration should hold each of these values;  so I vote for (b).

agc-quadra:
@davidiw the failing test `storage::command_adapter::tests::test_write_and_read` appears to b failing at master....

davidiw:
Please rebase and then I'll merge.

agc-quadra:
> Please rebase and then I'll merge.

rebased.  Waiting for CI to confirm.

davidiw:
@bors-libra r+

let's be optimistic!

PR #4386, opened 2020-06-10
[Key Management] Support configurable structured logging to endpoint.

JoshLind:
## Motivation

This small PR updates the Key Manager to support structured logging to a configurable logging endpoint.

Note: in this PR, we also:
(i) add an error log to the Key Manager when it fails, so that this failure is now obvious in the logs;
(ii) fix the template config argument for the Key Manager script.

Yes.

## Test Plan

- To test logging to file, I simply verified that the logs we were written as expected to the specified location.

- To test udp logging, I spawned a local instance of logstash, and pointed the key manager to it. I then verified that the logs generated by the key manager were forwarded to the logstash instance. Here's an example of the key manager starting up and failing:

{"location":"secure/key-manager/src/lib.rs:133","host":"127.0.0.1","log":"Checking the status of the keys.","pattern":"Checking the status of the keys.","@version":"1","timestamp":"2020-06-09 23:53:49","@timestamp":"2020-06-09T23:53:49.487Z","module":"libra_key_manager"}
{"location":"secure/key-manager/src/lib.rs:131","host":"127.0.0.1","log":"The key manager has been created and is starting execution.","pattern":"The key manager has been created and is starting execution.","@version":"1","timestamp":"2020-06-09 23:53:49","@timestamp":"2020-06-09T23:53:49.487Z","module":"libra_key_manager"}
{"location":"secure/key-manager/src/main.rs:71","host":"127.0.0.1","log":"Creating a libra interface that talks to the JSON RPC endpoint at: \"http://127.0.0.1:8080\".","pattern":"Creating a libra interface that talks to the JSON RPC endpoint at: {:?}.","data":{"json_rpc_endpoint":"\"http://127.0.0.1:8080\""},"@version":"1","timestamp":"2020-06-09 23:53:49","@timestamp":"2020-06-09T23:53:49.472Z","module":"libra_key_manager"}
{"location":"secure/push-metrics/src/pusher.rs:48","host":"127.0.0.1","log":"PUSH_METRICS_ENDPOINT env var is not set. Skipping sending metrics.","pattern":"PUSH_METRICS_ENDPOINT env var is not set. Skipping sending metrics.","@version":"1","timestamp":"2020-06-09 23:53:49","@timestamp":"2020-06-09T23:53:49.472Z","module":"libra_secure_push_metrics::pusher"}
~                                                                                                                             

## Related PRs

None.

### DIFF
--- a/secure/key-manager/src/main.rs
+++ b/secure/key-manager/src/main.rs
@@ -36,6 +36,8 @@ fn main() {
         .is_async(key_manager_config.logger.is_async)
         .level(key_manager_config.logger.level)
         .init();
+    libra_logger::init_struct_log_from_env().expect("Failed to initialize structured logging");
+
     MetricsPusher::new(COUNTERS.clone()).start();
 
     create_and_execute_key_manager(key_manager_config).unwrap_or_else(|e| {
--- a/docker/safety-rules/key-manager.sh
+++ b/docker/safety-rules/key-manager.sh
@@ -3,10 +3,11 @@
 # SPDX-License-Identifier: Apache-2.0
 set -e
 
+# Parse parameters and execute config builder
 declare -a params
-if [ -n "${CFG_BASE_CONFIG}" ]; then # Path to base config
-    echo "${CFG_BASE_CONFIG}" > /opt/libra/etc/base.config.toml
-    params+="--template /opt/libra/etc/base.config.toml "
+if [ -n "${KEY_MANAGER_CONFIG}" ]; then # Path to key manager config
+    echo "${KEY_MANAGER_CONFIG}" > /opt/libra/etc/key_manager.config.toml
+    params+="--template /opt/libra/etc/key_manager.config.toml "
 fi
 if [ -n "${JSON_RPC_ENDPOINT}" ]; then
     params+="--json-rpc-endpoint ${JSON_RPC_ENDPOINT} "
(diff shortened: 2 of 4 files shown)


JoshLind:
@sherry-x one thing that I think is worth mentioning: it seems that on failure, the last error!(..) call that the key manager makes (i.e., the error!(..) call added by this PR) isn't actually logged. It seems that the process fails too early for the log to actually be sent? (I've tested this by adding a sleep and seeing the error call printed out). I'm wondering what the best way to address this will be? Given how other services are deployed, this might actually be a common problem we'll need to fix?

secure/key-manager/src/main.rs
```diff
@@ -36,6 +36,8 @@ fn main() {
         .is_async(key_manager_config.logger.is_async)
         .level(key_manager_config.logger.level)
         .init();
+    libra_logger::init_struct_log_from_env().expect("Failed to initialize structured logging");
```
sherry-x, line 39: should be easy to add support for logging to a local file as well. https://github.com/libra/libra/blob/master/common/logger/src/struct_log.rs#L185
sometimes log to local file is easier to use.

sherry-x:
> @sherry-x one thing that I think is worth mentioning: it seems that on failure, the last error!(..) call that the key manager makes (i.e., the error!(..) call added by this PR) isn't actually logged. It seems that the process fails too early for the log to actually be sent? (I've tested this by adding a sleep and seeing the error call printed out). I'm wondering what the best way to address this will be? Given how other services are deployed, this might actually be a common problem we'll need to fix? Is there a way to force flush the logs when a process shuts down?

when it fails early, does the error message get print out as stderr? If that's the case, stderr should be able to find in docker logs

secure/key-manager/src/main.rs
JoshLind: Thanks, @sherry-x! Done :)

JoshLind:
> when it fails early, does the error message get print out as stderr? If that's the case, stderr should be able to find in docker logs

Thanks, Sherry! So it does get logged to stderr, so I guess one could debug it in the docker logs as you say. I just figured it would be nice for the logs to also correctly get forwarded to the service, but that's likely a problem for another day 😄

JoshLind:
Thanks, all! @bors-libra r=davidiw

PR #4389, opened 2020-06-10
[Move] transaction builder codegen in python (and Rust)

matbd:
## Motivation

* Generate some python code equivalent to what is currently generated by Rust macros in the `transaction-builder` crate.

* For testing purposes, I also generate Rust transaction builders from scratch.

Note that the ABI described by macros in `language/transaction-builder` seems currently out of date compared to the Move code.

## Test Plan

```
# Spit out the python code
cargo run -p transaction-builder-generator

# Verify that the generated Rust code compiles
cargo test -p transaction-builder-generator

# Run test on the python codegen (not supported by CI)
cargo test -p transaction-builder-generator -- --ignored 
```

## Related PRs

#4367

### DIFF
--- a/language/transaction-builder-generator/src/python3.rs
+++ b/language/transaction-builder-generator/src/python3.rs
@@ -1,0 +1,140 @@
+// Copyright (c) The Libra Core Contributors
+// SPDX-License-Identifier: Apache-2.0
+
+use crate::type_not_allowed;
+use libra_types::transaction::{ArgumentABI, ScriptABI, TypeArgumentABI};
+use move_core_types::language_storage::TypeTag;
+
+use std::io::{Result, Write};
+
+pub fn output(out: &mut dyn Write, abis: &[ScriptABI]) -> Result<()> {
+    output_preamble(out)?;
+    for abi in abis {
+        output_builder(out, abi)?;
+    }
+    Ok(())
+}
+
+fn output_preamble(out: &mut dyn Write) -> Result<()> {
+    writeln!(
+        out,
+        r#"# pyre-ignore-all-errors
+import libra_types as libra
+import typing
+import serde_types as st"#
+    )
+}
+
+fn output_builder(out: &mut dyn Write, abi: &ScriptABI) -> Result<()> {
+    writeln!(
+        out,
+        "\ndef encode_{}_script({}) -> libra.Script:",
+        abi.name(),
+        [
+            quote_type_parameters(abi.ty_args()),
+            quote_parameters(abi.args()),
+        ]
+        .concat()
+        .join(", ")
+    )?;
+    writeln!(out, "{}", quote_doc(abi.doc()))?;
+    writeln!(
+        out,
+        r#"    return libra.Script(
+        code={},
+        ty_args=[{}],
+        args=[{}],
+    )"#,
+        quote_code(abi.code()),
+        quote_type_arguments(abi.ty_args()),
+        quote_arguments(abi.args()),
+    )?;
+    Ok(())
+}
+
+fn quote_doc(doc: &str) -> String {
+    let s: Vec<_> = doc.splitn(2, |c| c == '.').collect();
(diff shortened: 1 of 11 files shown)


language/transaction-builder-generator/src/python3.rs
```diff
@@ -0,0 +1,139 @@
...
+
+pub fn output(out: &mut dyn Write, abis: &[ScriptABI]) -> Result<()> {
+    output_preambule(out)?;
+    for abi in abis {
+        output_builder(out, abi)?;
+    }
+    Ok(())
+}
+
+fn output_preambule(out: &mut dyn Write) -> Result<()> {
```
sblackshear, line 17:
```suggestion
fn output_preamble(out: &mut dyn Write) -> Result<()> {
```

language/transaction-builder-generator/src/python3.rs
```diff
@@ -0,0 +1,139 @@
...
+        U8 => "st.uint8".into(),
+        U64 => "st.uint64".into(),
+        U128 => "st.uint128".into(),
+        Address => "libra.AccountAddress".into(),
+        Vector(type_tag) => match type_tag.as_ref() {
+            U8 => "bytes".into(),
+            _ => format!("typing.Sequence[{}]", quote_type(type_tag.as_ref())),
+        },
+
+        Struct(_) | Signer => panic!("not supported"),
```
sblackshear, line 120: Might want this error message to imply that this is banned by Move--the current message sounds like this just might not be feature supported by the codegen.

sblackshear: Same with other "not supported"'s on `Struct/Signer`

language/transaction-builder-generator/src/rust.rs
```diff
@@ -0,0 +1,124 @@
...
+
+pub fn output(out: &mut dyn Write, abis: &[ScriptABI]) -> Result<()> {
+    output_preambule(out)?;
+    for abi in abis {
+        output_builder(out, abi)?;
+    }
+    Ok(())
+}
+
+fn output_preambule(out: &mut dyn Write) -> Result<()> {
```
sblackshear, line 17:
```suggestion
fn output_preamble(out: &mut dyn Write) -> Result<()> {
```

language/transaction-builder-generator/src/python3.rs
matbd: haha that's the right spelling but not in English :)

language/transaction-builder-generator/src/python3.rs
matbd: Sure, I can do that.

For the records, I'm still curious why we allow complex types as type arguments but not complex values as regular arguments.

thefallentree:
cc @ssinghaldev

language/transaction-builder-generator/src/python3.rs
sblackshear: The following invariant is important for Move's security model: A record type `R` declared in module `M` can only be created by a `Pack` instruction inside a procedure of `M`.  If record values can be passed into a transaction script, this invariant doesn't hold.

The use of types is not privileged in the same--any module can declare a procedure or type that uses any type declared in another module (as long it doesn't create a circular dependency).

sblackshear: Said differently: producing a value of a particular type may be privileged, but producing a type of a particular kind never is.

matbd: Thanks for the explanation! Based on what you're saying, nothing prevents us from having arbitrarily nested objects in TransactionArgument as long as they are benign enough (say vectors of vectors). So I understand we should probably plan for it in terms of code generation and LCS stack overflow.

thefallentree:
looks great! I assume the generated file will be checked in and checked similar how yaml spec is checked ?

PR #4403, opened 2020-06-10
[secure storage] Remove BoxedStorage boilerplate

huitseeker:
- use [enum-dispatch](https://crates.io/crates/enum_dispatch) to remove boilerplate in `BoxedStorage`
- rename BoxedStorage -> Storage

This follows on #4386. Please unfold individual commit messages for details.


secure/storage/src/kv_storage.rs
```diff
@@ -22,8 +24,8 @@ pub trait KVStorage: Send + Sync {
 
     /// Resets and clears all data held in the storage engine.
     /// Note: this should only be exposed and used for testing. Resetting the storage engine is not
-    /// something that should be supported in production.
-    #[cfg(any(test, feature = "testing"))]
+    /// something that should be supported in production. In production, this
+    /// operation should be a no-op.
     fn reset_and_clear(&mut self) -> Result<(), Error>;
```
davidiw, line 29: this is not good, I would prefer that if this is implemented in production that it panics.... does enum_dispatch not work effectively for feature flags?

why can we not retain the existing behavior?

secure/storage/src/tests/suite.rs
```diff
@@ -34,16 +35,29 @@ const U64_KEY: &str = "U64_Key";
...
 /// Executes all storage tests on a given storage backend.
-pub fn execute_all_storage_tests(storage: &mut BoxedStorage) {
+pub fn execute_all_storage_tests(storage: &mut Storage) {
     for test in STORAGE_TESTS.iter() {
         test(storage);
         storage.reset_and_clear().unwrap();
     }
 }
 
+/// This test tries to set a key, reset the storage and then retrieve its
```
davidiw, line 45: this is already implicitly tested throughout, but it seems alright to have an explicit test...

secure/storage/src/kv_storage.rs
huitseeker: Well, quite simply, `enum_dispatch` does not support that yet.

huitseeker: I have a PR open to fix it, which should help lighten the present PR: https://gitlab.com/antonok/enum_dispatch/-/merge_requests/15

secure/storage/src/crypto_storage.rs
```diff
@@ -12,6 +13,7 @@ use serde::{Deserialize, Serialize};
 /// keys securely. The API offered here is inspired by the 'Transit Secret Engine' provided by
 /// Vault: a production-ready storage engine for cloud infrastructures (e.g., see
 /// https://www.vaultproject.io/docs/secrets/transit/index.html).
+#[enum_dispatch]
```
JoshLind, line 16: Somewhat orthogonal question, but, have we thought about the security implications of adding the enum_dispatch crate to our dependencies? Secure storage is part of the TCB, and in general we're trying to reduce our dependencies (and avoid adding new ones). Part of me wonders what kinds of rust magic (e.g., unsafe code), the enum_dispatch crate uses to achieve its amazing performance boosts 😄

huitseeker: So, I have audited the crate in question (and it's a proc-macro crate, so it really can't do anything but inject rust code) and it does not generate unsafe code. Whether it contains a use of unsafe is irrelevant, but I nonetheless checked it is not the case for enum_dispatch itself. 

> what kinds of rust magic

It generates the code that I replace, and adds `TryInto<X> for Storage` for every variant that `Storage` closes over. 

The performance gain brought by enum_dispatch is that brought by an enum closing over an open trait hierarchy, whether that enum is written manually or not. Hence that performance gain exists between the dynamic dispatch that existed pre-#4386, and the enum that exists in #4386. That change in performance profile is well understood (dynamic dispatch is costly).

The present PR does not change performance, it removes manual boilerplate and generates the same solution it with a macro.

The trade-off here is not simply whether we take a dependency or not, but whether the risk induced by generating this code using a macro is smaller or greater than letting such boilerplate hinder the maintenability of our code base.

If we wanted, we could also rewrite the boilerplate-generation ourselves (enum_dispatch was written a while ago and could use conveniences brought on by the latest syn/quote practices — also @matbd suggested he could use plain macros to do something similar).

JoshLind: Awesome, thanks for looking into it and for the thorough explanation @huitseeker! This now makes sense to me.

secure/storage/src/kv_storage.rs
huitseeker: No longer a problem with enum_dispatch 0.3.1.

PR #4404, opened 2020-06-10
[secure-storage] Support HTTPS proxy for Github backend

mgorven:
## Motivation

If the `https_proxy` environment variable is set, use it with ureq for
Github requests.

## Test Plan

 Ran upload without `https_proxy` set, failed as before. Set
`https_proxy`, started working.

### DIFF
--- a/secure/storage/github/src/lib.rs
+++ b/secure/storage/github/src/lib.rs
@@ -5,6 +5,7 @@
 
 use serde::{Deserialize, Serialize};
 use serde_json::json;
+use std::env;
 use thiserror::Error;
 
 /// Request timeout for github operations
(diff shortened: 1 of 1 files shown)


secure/storage/github/src/lib.rs
```diff
@@ -173,7 +174,10 @@ impl Client {
             .set("Authorization", &format!("token {}", self.token))
             .set(ACCEPT_HEADER, ACCEPT_VALUE)
             .timeout_connect(TIMEOUT);
-        request
+        match env::var("https_proxy") {
+            Ok(proxy) => request.set_proxy(ureq::Proxy::new(proxy).unwrap()).build(),
```
davidiw, line 178: maybe use an expect here => expect("Unable to parse https_proxy environmental variable")

PR #4408, opened 2020-06-10
put the db-bootstrapper tool into validator docker image

msmouse:
## Motivation

for further usage, when e.g. drilling on operational scenarios.

yes

## Test Plan
built image and saw it in validator-dynamic


docker/validator/Dockerfile
```diff
@@ -17,7 +17,7 @@ FROM toolchain AS builder
 COPY . /libra
 
 ENV RUSTFLAGS "-Ctarget-cpu=skylake -Ctarget-feature=+aes,+sse2,+sse4.1,+ssse3"
-RUN cargo build --release -p libra-node -p cli -p config-builder -p libra-key-manager -p safety-rules && cd target/release && rm -r build deps incremental
+RUN cargo build --release -p libra-node -p cli -p config-builder -p libra-key-manager -p safety-rules -p db-bootstrapper && cd target/release && rm -r build deps incremental
```
mgorven, line 20: You need to copy this exact build command to the other docker files, otherwise the build will get repeated for the different images.

docker/validator/Dockerfile
```diff
@@ -17,7 +17,7 @@ FROM toolchain AS builder
...
 RUN strip /libra/target/release/config-builder
```
mgorven, line 21: Please strip the new binary.

docker/validator/Dockerfile
msmouse: Hmm. but why didn't we issue separate build commands so that different roles doesn't refer to components they don't need?

mgorven: Building everything once is faster than doing separate builds for each image.

PR #4410, opened 2020-06-10
[core modules] Update the staging so we don't overwrite binary if it's the same

tzakian:
This was causing us to constantly rebuild the stdlib Rust binary since the mtime was changing.

Now, as long as the binaries don't change we won't change the staged files. Further, this PR separates out the `stdlib` and `compiled_stdlib` crates. Due to this separation, we shouldn't ever need to recompile Rust to build the stdlib. However, if the compiled binaries changed, then the a build will be triggered for crates that use the `compiled_stdlib`.


sblackshear:
Unrelated question: we only do `include_bytes` to make docker happy, right? And if so, I wonder if we could do it somewhere else where it won't disrupt the build...

tzakian:
So we could do it elsewhere. I would say the best way to do that would be to separate out the "get me the staged stdlib" related functions into a separate library crate, and have that pull in the staged files. I'm happy to do that if you think that would be preferable.

sblackshear:
Seems worth considering later perhaps (since changing the standard library Move code, but not wanting to wait for the Rust binary to recompile is pretty common), but definitely not a blocker for this clear improvement!

PR #4412, opened 2020-06-10
[Move] Support for on chain travel rule limit, updating and retrieving

moeziniaf:
This PR establishes an on chain config for travel rule limits.
It introduces an update travel rule limit script via Blessed account, sets limit at genesis and allows for other changes by account with Config changing privileges


language/stdlib/src/transaction_scripts.rs
```diff
@@ -55,6 +55,7 @@ pub enum StdlibScript {
     RotateConsensusPubkey,
     RotateSharedEd2551PublicKey,
     TieredMint,
+    TravelRuleLimitUpdate,
```
sblackshear, line 58: Nit: verb first (`UpdateTravelRuleLimit`) is the pattern used  in most of  these.

language/stdlib/modules/TravelRuleLimit.move
```diff
@@ -0,0 +1,39 @@
...
+            Transaction::assert(Signer::address_of(account) == LibraConfig::default_config_address(), 1);
+
+            LibraConfig::publish_new_config<TravelRuleLimit>(
+                account,
+                TravelRuleLimit { micro_libra_limit: 1000000 },
+            );
+        }
+
+        // Anyone can fetch the current set limit
+        public fun get_cur_limit(): u64 {
```
sblackshear, line 24: Might be worth including "micro" in the name to avoid the expectation that this is 1000...

language/stdlib/modules/LibraConfig.move
```diff
@@ -59,7 +59,7 @@ module LibraConfig {
         let signer_address = Signer::address_of(account);
         Transaction::assert(
             ::exists<ModifyConfigCapability<Config>>(signer_address)
-             || signer_address == Association::root_address(),
+             || signer_address == Association::root_address() || signer_address == Association::treasury_compliance_account(),
```
sblackshear, line 62: Maybe we should do `Association::is_association(signer_address)` (which covers both root and TC) here?

language/stdlib/modules/LibraAccount.move
```diff
@@ -266,13 +267,10 @@ module LibraAccount {
...
         Transaction::assert(deposit_value > 0, 7);
-
-        // TODO: on-chain config for travel rule limit instead of hardcoded value
-        // TODO: nail down details of limit (specified in LBR? is 1 LBR a milliLibra or microLibra?)
-        let travel_rule_limit = 1000;
+        let travel_rule_limit_microlibra = TravelRuleLimit::get_cur_limit();
         // travel rule only applies for payments over a threshold
-        let above_threshold =
-            Libra::approx_lbr_for_value<Token>(deposit_value) >= travel_rule_limit;
+        let approx_lbr_microlibra_value = Libra::approx_lbr_for_value<Token>(deposit_value) * 1000;
```
sblackshear, line 272: If the value in `TravelRuleLimit` is already  in microlibra (1000000), doesn't that mean you shouldn't multiply by 1000 here?

moeziniaf: ah, is Libra::value already microLibra? / I guess everywhere else too?

ankushagarwal:
What exactly is travel rule threshold enforcing? Is it the max LBR that can be transferred to/from an unhosted wallet?

moeziniaf:
right now the logic is L279 of LibraAccount:
If txn above limit is between two different vasps, it is denied

language/stdlib/modules/LibraAccount.move
sblackshear: Yeah, `Libra<T>::value` is always micro`T`. `approx_lbr_for_value` is just for converting between `T` and `LBR` (cc @tzakian to be sure)

language/stdlib/modules/TravelRuleLimit.move
```diff
@@ -0,0 +1,39 @@
...
+        // Anyone can fetch the current set limit
+        public fun get_cur_microlibra_limit(): u64 {
+            let rule = LibraConfig::get<TravelRuleLimit>();
+            rule.micro_libra_limit
+        }
+
+        public fun set(account: &signer, micro_libra_limit: u64) {
+            Transaction::assert(Signer::address_of(account) == Association::treasury_compliance_account(), 3);
+
+            LibraConfig::set<TravelRuleLimit>(
```
tzakian, line 32: I'm very confused how this is working (assuming the functional test passes)...

`LibraConfig::set` requires that `account` must have a `TravelRuleLimit` config published under  it. However, `initialize` is publishing it under `default_config_address()`, and  `account` here is `0xB1E55ED`. You might need to store the `ModifyConfigCapability` from `publish_new_config_with_capability` in an internal (wrapper) resource, and then use that here to modify the config (using `set_with_capability`).

language/stdlib/modules/LibraAccount.move
tzakian: Sam is correct. The `approx_lbr_for_value` is meant to be a "base units of currency to base units of LBR" (microlibra).

moeziniaf: it seems elsewhere (for example Libra<CoinType> { value } is treated as LBR not microLBR.
As the threshold was hardcoded to 1000 previously, it seems like some tests use that 1000 value when it should have been 1_000_000?

sblackshear:
>  As the threshold was hardcoded to 1000 previously, it seems like some tests use that 1000 value when it should have been 1_000_000?

Yeah, that was my fault--I forgot about microlibra and used 1000 everywhere.

language/stdlib/modules/TravelRuleLimit.move
```diff
@@ -0,0 +1,39 @@
+address 0x0 {
+
+    module TravelRuleLimit {
```
sblackshear, line 3: Also: can we call this `DualAttestationLimit`? This is the term that we use elsewhere, and the travel rule isn't the only use-case for payments with dual attestation.

language/stdlib/modules/DualAttestationLimit.move
```diff
@@ -0,0 +1,50 @@
...
+
+    public fun set_microlibra_limit(tc_account: &signer, micro_lbr_limit: u64) acquires ModifyCapability {
+        Transaction::assert(
+            micro_lbr_limit >= 1000,
+            4
+        );
+        let tc_address = Signer::address_of(tc_account);
+        Transaction::assert(tc_address == Association::treasury_compliance_account(), 3);
+        let modify_cap = &borrow_global<ModifyCapability>(tc_address).cap;
+        // let modify_cap = borrow_global<ModifyCapability<Self::DualAttestationLimit>>(tc_address).cap;
```
sblackshear, line 42: Delete?

language/stdlib/modules/DualAttestationLimit.move
```diff
@@ -0,0 +1,50 @@
...
+    struct DualAttestationLimit {
+        micro_lbr_limit: u64,
+    }
+
+    // A wrapped capability to allow updating of the on-chain config by Blessed
+    resource struct ModifyCapability {
+        cap: LibraConfig::ModifyConfigCapability<Self::DualAttestationLimit>,
+    }
+
+    /// Travel rule limit set during genesis
```
sblackshear, line 18:
```suggestion
    /// Dual attestation limit set during genesis
```

language/stdlib/modules/DualAttestationLimit.move
```diff
@@ -0,0 +1,50 @@
...
+        move_to(tc_account, ModifyCapability { cap })
+    }
+
+    // Anyone can fetch the current set limit
+    public fun get_cur_microlibra_limit(): u64 {
+        let rule = LibraConfig::get<DualAttestationLimit>();
+        rule.micro_lbr_limit
+    }
+
+    public fun set_microlibra_limit(tc_account: &signer, micro_lbr_limit: u64) acquires ModifyCapability {
```
sblackshear, line 34: I think if you publish `LibraConfig::ModifyConfigCapability<Self::DualAttestationLimit>` at the top-level of `tc_account`,  you can:
- delete `ModifyCapability`
- move the logic in this procedure to the `update_travel_rule_limit` script  (which can `move_from` the cap, pass a ref to `LibraConfig::set_with_capability`, and then move it back)

But this can also be done in a follow-up (I'm happy to do it); this looks great and has already gone through a lot of iteration!

moeziniaf: To clarify: publish the capability during genesis/tc account creation in `LibraAccount.move`, using the config_account signer?

sblackshear: Yep!

PR #4414, opened 2020-06-10
[cluster_test] Pre allocate node manually - Flexible config 1/n

andll:
This diff is part of series of changes that aim to move away from libra_init usage in cluster test and generate configs inside cluster test binary. This will allow for higher flexibility for different scenarios and will make cluster test setup closer to production (that is not going to use libra_init).

In this diff we pre-allocate node manually, instead of relying on kubernetes to do so. This will be needed later to populate config on the node before running it.

Also, update job_template to allow utilities to access node data dir.

### DIFF
--- a/testsuite/cluster-test/src/cluster_swarm/cluster_swarm_kube.rs
+++ b/testsuite/cluster-test/src/cluster_swarm/cluster_swarm_kube.rs
@@ -29,7 +29,7 @@ use k8s_openapi::api::batch::v1::Job;
 use kube::api::ListParams;
 use libra_config::config::DEFAULT_JSON_RPC_PORT;
 use reqwest::Client as HttpClient;
-use std::process::Command;
+use std::{collections::HashSet, process::Command};
 
 const DEFAULT_NAMESPACE: &str = "default";
 
@@ -264,6 +264,12 @@ impl ClusterSwarmKube {
         Ok(())
     }
 
+    async fn list_nodes(&self) -> Result<Vec<Node>> {
+        let node_api: Api<Node> = Api::all(self.client.clone());
+        let lp = ListParams::default().labels("nodeType=validators");
+        Ok(node_api.list(&lp).await?.items)
+    }
+
     async fn get_pod_node_and_ip(&self, pod_name: &str) -> Result<(String, String)> {
         libra_retrier::retry_async(libra_retrier::fixed_retry_strategy(10000, 60), || {
             let pod_api: Api<Pod> = Api::namespaced(self.client.clone(), DEFAULT_NAMESPACE);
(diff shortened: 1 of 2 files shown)


testsuite/cluster-test/src/cluster_swarm/job_template.yaml
```diff
@@ -19,11 +19,20 @@ spec:
       containers:
       - name: main
         image: {image}
+        volumeMounts:
```
snakech, line 22: may I ask what is this file used for and when do we need to touch it?

andll: job_template is used when `Instance::util_cmd` is called to run arbitrary command on validator node.
This just adds ability for whatever command runs this util to access validator database and configs

testsuite/cluster-test/src/cluster_swarm/job_template.yaml
```diff
@@ -19,11 +19,20 @@ spec:
...
+          runAsUser: 0 # To get permissions to write to /opt/libra/data
           privileged: true
           capabilities:
             add: ["SYS_ADMIN"]
       restartPolicy: Never
+      volumes:
+        - name: data
+          hostPath:
+            path: /data
+            type: Directory
```
ankushagarwal, line 37: It's probably safer to set type to `DirectoryOrCreate`. Not all hosts will be guaranteed to have the `/data/` directory

testsuite/cluster-test/src/cluster_swarm/cluster_swarm_kube.rs
```diff
@@ -264,6 +264,12 @@ impl ClusterSwarmKube {
         Ok(())
     }
 
+    async fn list_nodes(&self) -> Result<Vec<Node>> {
+        let node_api: Api<Node> = Api::all(self.client.clone());
+        let lp = ListParams::default().labels("nodeType=validators");
```
ankushagarwal, line 269: Can we take the `labels` as a parameter. I imagine we will run different nodeTypes when we start running LSR/Vault

testsuite/cluster-test/src/cluster_swarm/cluster_swarm_kube.rs
```diff
@@ -450,6 +454,32 @@ impl ClusterSwarmKube {
...
+        // Holding lock for read-verfy-write to avoid race conditions on this map
+        let mut node_map = self.node_map.lock().await;
+        if let Some(existed) = node_map.get(instance_config) {
+            return Ok(existed.clone());
+        }
+        let used_nodes: HashSet<_> = node_map.values().collect();
+        for node in nodes {
+            let name = node
+                .metadata
+                .expect("Metadata not found on node")
```
ankushagarwal, line 469: Can we propagate the error instead of panic?

testsuite/cluster-test/src/cluster_swarm/job_template.yaml
andll: Yeah, that's actually good idea

testsuite/cluster-test/src/cluster_swarm/cluster_swarm_kube.rs
andll: Will need to think more about splitting node types later.
Some of the usage need 'all nodes cluster test created nodes' and some need 'nodes of type x'

PR #4418, opened 2020-06-10
[network] decrease MAX_CONNECTION_DELAY_MS

gregnazario:
10 minutes is too long for a maximum backoff, and it causes much longer
recovery times than expected.  It is now changed to 1 minute as a
maximum backoff to ensure that it retries every minute.


andll:
Would it harm if we reduce it even further (say 10 sec?) until we get proper fix for the issue?
Thing is, this number play some role as it increases duration of test and cause flakiness. So unless we think it will destroy prod I would have some small value here to make sure our dev experience does not suffer

gregnazario:
> Would it harm if we reduce it even further (say 10 sec?) until we get proper fix for the issue?
> Thing is, this number play some role as it increases duration of test and cause flakiness. So unless we think it will destroy prod I would have some small value here to make sure our dev experience does not suffer

Yeah, we can reduce it further.  Usually, the only thing with decreasing a backoff like this is either 1. overloading the system with retries, or 2. cost.  And right now I don't think we have too much of a worry for either, so we can trade it off to be less flaky.

andll:
Thanks, let's put some small value then for now?

Issue #3665, opened 2020-04-30
[move-prover][Feature Request] Specification sanity checks


DavidLDill:
# 🚀 Feature Request

Warnings for possible specification problems.

## Motivation

Erroneous specifications that lead to false negatives are a big risk.  We will need to do whatever we can to try to catch them early.

## Pitch

**Describe the solution you'd like**
I will suggest two possible sanity checks for specifications.  I'm sure more will occur to us as we get more experience.  (These are more related to checking "ensures false" but more complicated and perhaps controversial.)

I don't expect us to implement these immediately.  They're really just ideas, but I wanted to get them on the record.

1.  Check that every variable and every field of every struct can change value.  We can do this by trying to prove "ensures x == old(x)" for each such variable x.  We can do this in one pass, possibly while verifying useful conditions.  The difference from normal verification is that report when the condition is SATISFIED, not when it is VIOLATED.  The warning should, of course, be something like "Warning: foo.x never changes"

Example: In a version of libra.move, I discovered that it was burn unexpectedly satisfied some properties.  It turned out that it always aborted because it contained "assert is_authorized" and is_authorized could never be set to true.  It took quite a lot of debugging to figure this out, but could have been caught instantaneously.

2. When you have an expression P ==> Q, check whether P can ever be true.  That can be done by checking whether !P holds at the same point.  This probably should be generalized, since other Boolean operators have similar effects.  Probably should check that Q can be false  in P ==> Q, that P, Q can be false in P||Q, that P,Q can be true in P&&Q, etc.  That's a lot of checks, so we might want to think more about it.  Basically, you want to check that each variable has a possibility of taking a value that is NOT a controlling value for the whole expression.   We could come up with an even more general idea, but maybe start with something simple where people have actually made the error (I made the P ==> Q error in the is_valid example above, because is_valid was P).

**Describe alternatives you've considered**

**Are you willing to open a pull request?** (See [CONTRIBUTING](../../CONTRIBUTING.md))

PR #4420, opened 2020-06-11
[move-prover] Added `aborts_if true;` smoke test.

kkmc:
Added `aborts_if true;` smoke test to the move-prover.
This can be turned on for specific functions by declaring
`pragma smoke_test = true;` inside of a spec fun block.
An error is reported if the function being verified can prove
$abort_flag is always true.

Related to issue #3665.

Note: Had a discussion with David about smoke testing ensures false in Boogie vs at the Move language and decided on the latter / to check for `aborts_if true;` (at the Move level). This is not the same as the soundness check for assert false.

## Motivation

When writing Move programs, it's possible that a combination of implementation and specifications will force the procedure to always abort. In these situations, we want to report an error to the user informing them about this issue.

Yes.

## Test Plan

Existing test suite.

## Related PRs

None.

### DIFF
--- a/language/move-prover/src/boogie_wrapper.rs
+++ b/language/move-prover/src/boogie_wrapper.rs
@@ -6,7 +6,7 @@
 use anyhow::anyhow;
 
 use std::{
-    collections::{BTreeMap, BTreeSet},
+    collections::{BTreeMap, BTreeSet, HashSet},
     fs,
     option::Option::None,
     process::Command,
@@ -22,11 +22,12 @@ use regex::Regex;
 
 use spec_lang::{
     code_writer::CodeWriter,
-    env::{FunId, GlobalEnv, Loc, ModuleId, StructId},
+    env::{FunId, GlobalEnv, Loc, ModuleId, StructId, SMOKE_TEST_PRAGMA, VERIFY_PRAGMA},
+    symbol::Symbol,
     ty::{PrimitiveType, Type},
 };
 
-use crate::cli::Options;
+use crate::cli::{Options, VerificationScope};
 // DEBUG
 // use backtrace::Backtrace;
 use spec_lang::env::NodeId;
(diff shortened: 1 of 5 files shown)


language/move-prover/src/boogie_wrapper.rs
```diff
@@ -138,12 +145,107 @@ impl<'env> BoogieWrapper<'env> {
...
+            // Currently the only type of test is the aborts_if true check
+            let kind = SmokeTestKind::AbortsIfTrue;
+            Some((fname_symbol, kind))
+        } else {
+            None
+        }
+    }
+
+    /// Helper function for adding smoke test errors as Diagnostics.
+    fn add_smoke_test_errors(&self, funs_with_errors: HashSet<(Symbol, SmokeTestKind)>) {
```
wrwg, line 205: I wonder whether we would be matching every function with a simple name in the set of modules with this approach, even if only one of those has the pragma set.

This deserves at least some comment. It's probably ok since most of the time we have only one module here. On the other hand, it is possible to inject precise FunId and ModuleId into models to make this exact, like someone would expect using a pragma on a specific function. 

Please add a TODO and comment that this should be revisited, or correct if I'm wrong that the error identification is ambigious.

kkmc: I agree that we should inject precise FunIds and ModuleIds into the models (that's exactly what I was trying to look for originally). I've placed a comment for now.

language/move-prover/src/boogie_wrapper.rs
```diff
@@ -138,12 +145,107 @@ impl<'env> BoogieWrapper<'env> {
...
+        // Find the source location of the error
+        let (_, loc_opt) = self.get_locations(error.position);
+        let source_loc_opt = self.to_proper_source_location(loc_opt).and_then(|loc| {
+            if error.kind.is_from_verification() {
+                Some(loc)
+            } else {
+                None
+            }
+        });
+        source_loc_opt.as_ref()?;
```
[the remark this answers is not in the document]
kkmc: I just amended my commit. Is this what you meant?

CI told me to use .as_ref()? which was a little strange but also worked. This seems to make much more sense.

PR #4422, opened 2020-06-11
[state-sync] tidying up

mimoo:
1. move the client handle to state-sync to its own file for clarity
2. make set_initialization_listener panic to prevent multiple calls
3. added README to folder

### DIFF
--- a/state-synchronizer/src/coordinator.rs
+++ b/state-synchronizer/src/coordinator.rs
@@ -8,6 +8,7 @@ use crate::{
     executor_proxy::ExecutorProxyTrait,
     network::{StateSynchronizerEvents, StateSynchronizerMsg, StateSynchronizerSender},
     peer_manager::{PeerManager, PeerScoreUpdateType},
+    state_sync_client::{CoordinatorMessage, SyncRequest},
     PeerId, SynchronizerState,
 };
 use anyhow::{bail, ensure, format_err, Result};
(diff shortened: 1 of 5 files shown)


state-synchronizer/src/coordinator.rs
```diff
@@ -866,16 +867,14 @@ impl<T: ExecutorProxyTrait> SyncCoordinator<T> {
...
-                    }
-                    TargetType::TargetLedgerInfo(sync_req.target.clone())
+                Some(sync_req) if sync_req.target.ledger_info().version() <= known_version => {
+                    debug!(
+                        "[state sync] Reached version {}, no need to send more requests",
+                        known_version
+                    );
+                    return Ok(());
                 }
+                Some(sync_req) => TargetType::TargetLedgerInfo(sync_req.target.clone()),
```
JoshLind, line 851: I'm not sure if this is more or less clear than the original.. (but that's my personal preference -- I'm not a big fan of overlapping match guards if they can be otherwise avoided)

christinacelee: personally I find this less clear and more error-prone.. what is wrong with the current way it is written?

mimoo: when using matching you have access to guards for exactly this type of scenarios, which is supposed to make the code clearer. We don't really have guidelines about this so I don't care much, but it's clearer to me.

Maybe some rust experts should think about the place of this pattern in our codebase :) @bmwill @sunshowers

sunshowers: Don't have a strong opinion either way, but this is fairly similar to how match is used in FP languages so it's not immediately objectionable to me.

mimoo: thanks for chiming in! Indeed, if you're unfamiliar with FP this might look weird, at the same time Rust looks weird to anyone unfamiliar with FP :D

PR #4424, opened 2020-06-11
[core modules/vm] Add a module publishing capability to the association root account.

tzakian:
Module publishing will be restricted, however the association still needs to have the ability to publish modules to the chain. 

This grants the Association privilege of `ModulePublishing` to the root association account (and not to the TC account, but that's easy if we want to allow it). 

## Tests
Adds a bunch of e2e tests to make sure that the new rules for publishing are being respected.


language/stdlib/modules/Association.move
```diff
@@ -36,6 +36,11 @@ module Association {
     /// means that the address is an association address.
     struct Association { }
 
+    /// The presence of an instance of an `Association::ModulePublishing`
+    /// privilege at an address means that that address can publish code to
+    /// the chain.
+    struct ModulePublishing { }
```
sblackshear, line 42: Nit: I am trying to use present tense, verb first (e.g. `PublishModule`) names for all capability/script names (since that seems to be the convention we follow most consistently, though not perfectly).

tzakian:
@bors-libra r=sblackshear

come on........

Issue #4378, opened 2020-06-09
[config] Hardening


davidiw:
The following is an non-exhaustive list of areas in which our configs can be improved leading up to V1 to make it less likely for failure and easy to work with.

Management:
- [x] Add set-operator into management (owner chooses operator based upon account identity — name within layout — or account address)
- [x] Explicit hierarchical owner / operator framework
- [x] Modify validator address from operator to owner
- [x] Add insert-waypoint into management
- [x] owner account can optionally be inserted via validator-config
- [x] Support environment proxy for management / ureq / secure storage

Secure config:
- [x] Config should depend on secure storage and we should have accessors for fields that can be in either
- [ ] Unify secure storage backends into a single configuration that maps name -> storage, then use those storage names instead of direct storage configs in data that can be stored in configs

Core config:
- [x] In network, merge gossip fields into enum variant (discovery interval and advertised address)
- [x] TOML -> YAML
- [x] Specific config for safety rules
- [ ] Specific config for execution correctness
- [x] Simplify fields to eliminate unnecessary nesting

Safety Rules:
- [x] Pull author account from secure storage (operator -> owner)
- [x] Support dns for safety rules (use NetworkAddress)

PR #4427, opened 2020-06-11
InsertWaypoint

Rustinante:
Added an `insert-waypoint` command for the `config/management` crate.

`insert-waypoint --local BACKEND [--remote BACKEND] [--waypoint WAYPOINT_VALUE]`
where one any only one of `--remote` and `--waypoint` must be provided.

Issue: #4378 

Yes

## Test Plan
Modified the `smoke_test` in the management crate to test `insert_waypoint`.

### DIFF
--- a/config/management/src/lib.rs
+++ b/config/management/src/lib.rs
@@ -44,6 +44,8 @@ pub enum Command {
     CreateWaypoint(crate::waypoint::CreateWaypoint),
     #[structopt(about = "Retrieves data from a store to produce genesis")]
     Genesis(crate::genesis::Genesis),
+    #[structopt(about = "Insert a waypoint")]
+    InsertWaypoint(crate::waypoint::InsertWaypoint),
     #[structopt(about = "Submits an Ed25519PublicKey for the operator")]
     OperatorKey(crate::key::OperatorKey),
     #[structopt(about = "Submits an Ed25519PublicKey for the owner")]
@@ -61,6 +63,7 @@ pub enum CommandName {
     AssociationKey,
     CreateWaypoint,
     Genesis,
+    InsertWaypoint,
     OperatorKey,
     OwnerKey,
     SetLayout,
@@ -74,6 +77,7 @@ impl From<&Command> for CommandName {
             Command::AssociationKey(_) => CommandName::AssociationKey,
             Command::CreateWaypoint(_) => CommandName::CreateWaypoint,
             Command::Genesis(_) => CommandName::Genesis,
+            Command::InsertWaypoint(_) => CommandName::InsertWaypoint,
             Command::OperatorKey(_) => CommandName::OperatorKey,
             Command::OwnerKey(_) => CommandName::OwnerKey,
             Command::SetLayout(_) => CommandName::SetLayout,
@@ -89,6 +93,7 @@ impl std::fmt::Display for CommandName {
             CommandName::AssociationKey => "association-key",
             CommandName::CreateWaypoint => "create-waypoint",
             CommandName::Genesis => "genesis",
+            CommandName::InsertWaypoint => "insert-waypoint",
             CommandName::OperatorKey => "operator-key",
             CommandName::OwnerKey => "owner-key",
             CommandName::SetLayout => "set-layout",
@@ -105,6 +110,7 @@ impl Command {
             Command::AssociationKey(_) => self.association_key().unwrap().to_string(),
             Command::CreateWaypoint(_) => self.create_waypoint().unwrap().to_string(),
             Command::Genesis(_) => format!("{:?}", self.genesis().unwrap()),
+            Command::InsertWaypoint(_) => self.insert_waypoint().unwrap().to_string(),
             Command::OperatorKey(_) => self.operator_key().unwrap().to_string(),
             Command::OwnerKey(_) => self.owner_key().unwrap().to_string(),
             Command::SetLayout(_) => self.set_layout().unwrap().to_string(),
@@ -141,6 +147,17 @@ impl Command {
         } else {
             Err(Error::UnexpectedCommand(
                 CommandName::Genesis,
+                CommandName::from(&self),
+            ))
+        }
+    }
+
+    pub fn insert_waypoint(self) -> Result<Waypoint, Error> {
+        if let Command::InsertWaypoint(insert_waypoint) = self {
+            insert_waypoint.execute()
+        } else {
+            Err(Error::UnexpectedCommand(
+                CommandName::InsertWaypoint,
                 CommandName::from(&self),
             ))
         }
--- a/config/management/src/smoke_test.rs
+++ b/config/management/src/smoke_test.rs
@@ -1,7 +1,7 @@
 // Copyright (c) The Libra Core Contributors
 // SPDX-License-Identifier: Apache-2.0
 
-use crate::{layout::Layout, storage_helper::StorageHelper};
+use crate::{constants, layout::Layout, storage_helper::StorageHelper};
 use config_builder::{BuildSwarm, SwarmConfig};
 use libra_config::{
     config::{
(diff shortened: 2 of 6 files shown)


config/management/src/lib.rs
```diff
@@ -28,6 +28,7 @@ pub mod constants {
     pub const COMMON_NS: &str = "common";
     pub const LAYOUT: &str = "layout";
     pub const VALIDATOR_CONFIG: &str = "validator_config";
+    pub const WAYPOINT: &str = "waypoint";
```
davidiw, line 31: we already have this in libra_global_constants

config/management/src/smoke_test.rs
```diff
@@ -120,6 +120,23 @@ fn smoke_test() {
     for (i, mut config) in configs.iter_mut().enumerate() {
         let ns = i.to_string();
         let waypoint = helper.create_waypoint(&ns).unwrap();
```
davidiw, line 122: you need to change this, we should create the waypoint outside this loop -- create it in the shared secure storage -- create should not be modifying any validators storage

config/management/src/smoke_test.rs
```diff
@@ -120,6 +120,23 @@ fn smoke_test() {
...
+        // read it and perform insert-waypoint
+        common_storage
+            .set(
+                crate::constants::WAYPOINT,
+                Value::String(waypoint.to_string()),
+            )
+            .unwrap();
+
+        // TODO(Rustinante): figure out how to run the validator
+        if let Some(_validator_config) = &config.validator_network {
```
davidiw, line 134: remove if let some?

config/management/src/smoke_test.rs
```diff
@@ -120,6 +120,23 @@ fn smoke_test() {
...
         let waypoint = helper.create_waypoint(&ns).unwrap();
+
+        // save the waypoint into the common namespace so that validators can
+        // read it and perform insert-waypoint
+        common_storage
+            .set(
+                crate::constants::WAYPOINT,
+                Value::String(waypoint.to_string()),
+            )
+            .unwrap();
```
davidiw, line 131: this is what insert should do, delete

config/management/src/smoke_test.rs
```diff
@@ -120,6 +120,23 @@ fn smoke_test() {
...
+
+        // TODO(Rustinante): figure out how to run the validator
+        if let Some(_validator_config) = &config.validator_network {
+            helper
+                .insert_waypoint(crate::constants::COMMON_NS, Some(&waypoint.to_string()))
```
davidiw, line 136: insert should happen within the current nodes ns and it should be read from the remote secure storage (ns) that was create in the create-waypoint

config/management/src/storage_helper.rs
```diff
@@ -120,6 +120,36 @@ impl StorageHelper {
         command.genesis()
     }
 
+    pub fn insert_waypoint(
+        &self,
+        remote_ns: &str,
+        waypoint_str: Option<&str>,
```
davidiw, line 126: these are mutually exclusive, but for storage helper let's only expose remote_ns

config/management/src/waypoint.rs
```diff
@@ -41,18 +43,86 @@ impl CreateWaypoint {
...
+}
+
+impl InsertWaypoint {
+    pub fn execute(self) -> Result<Waypoint, Error> {
+        if self.waypoint.is_some() && self.secure_backends.remote.is_some() {
+            return Err(Error::CommandArgumentError(
+                "only one of --waypoint and --remote can be provided".to_string(),
+            ));
+        }
+        let mut waypoint_string = self.waypoint.clone();
```
davidiw, line 72: nit new line for readability

davidiw: also why are we cloning this?

config/management/src/waypoint.rs
```diff
@@ -41,18 +43,86 @@ impl CreateWaypoint {
...
+        if let Some(remote) = self.secure_backends.remote {
```
davidiw, line 73:
```
  let waypoint = if let Some .... {
  } else {
    self.waypoint.clone()
  }
```

config/management/src/waypoint.rs
```diff
@@ -41,18 +43,86 @@ impl CreateWaypoint {
...
+                &waypoint,
+                &mut backend,
+                RelativePosition::Local,
+            )?;
+            Ok(waypoint)
+        } else {
+            Err(Error::CommandArgumentError(
+                "please provide either --waypoint or --remote".to_string(),
+            ))
+        }
```
davidiw, line 99: seems weird to have this twice, use an expect / panic if we get here... if we need this

config/management/src/waypoint.rs
```diff
@@ -41,18 +43,86 @@ impl CreateWaypoint {
...
+    #[structopt(flatten)]
+    secure_backends: SecureBackends,
+
+    #[structopt(long)]
+    waypoint: Option<String>,
+}
+
+impl InsertWaypoint {
+    pub fn execute(self) -> Result<Waypoint, Error> {
+        if self.waypoint.is_some() && self.secure_backends.remote.is_some() {
```
davidiw, line 67: recommend you do
```
  let waypoint = if let Some(...) {
  } else if let Some (...) {
  } else {
    return Err...
  }
```

config/management/src/waypoint.rs
```diff
@@ -41,18 +43,86 @@ impl CreateWaypoint {
...
+            ))
+        }
+    }
+
+    fn insert_waypoint_to_backend(
+        waypoint: &Waypoint,
+        backend: &mut BoxedStorage,
+        backend_location: RelativePosition,
+    ) -> Result<(), Error> {
+        use RelativePosition::{Local, Remote};
```
davidiw, line 107: pretty cheap just to use RelativePosition::Local / Remote below

config/management/src/waypoint.rs
Rustinante: Good tip!!

config/management/src/waypoint.rs
Rustinante: fixed by the tip provded!

config/management/src/waypoint.rs
Rustinante: fixed by the tip provided!

config/management/src/storage_helper.rs
Rustinante: sounds good!

Rustinante:
Addressed the review comments! Please let me know if it looks okay. Thanks

config/management/src/smoke_test.rs
```diff
@@ -116,10 +116,15 @@ fn smoke_test() {
...
+    // Save the waypoint into shared secure storage so that validators can perform insert_waypoint
+    let waypoint = helper.create_waypoint(constants::COMMON_NS).unwrap();
+
     // Step 5) Introduce waypoint and genesis into the configs and verify along the way
     for (i, mut config) in configs.iter_mut().enumerate() {
         let ns = i.to_string();
-        let waypoint = helper.create_waypoint(&ns).unwrap();
+
+        helper.insert_waypoint(&ns, constants::COMMON_NS).unwrap();
+
```
davidiw, line 127: no strong benefit for all the added whitespace :P

config/management/src/waypoint.rs
```diff
@@ -40,19 +41,82 @@ impl CreateWaypoint {
...
+}
+
+impl InsertWaypoint {
+    pub fn execute(self) -> Result<Waypoint, Error> {
+        if self.waypoint.is_some() && self.secure_backends.remote.is_some() {
+            return Err(Error::CommandArgumentError(
+                "only one of --waypoint and --remote can be provided".to_string(),
+            ));
         }
+        let waypoint_string = if let Some(waypoint_string) = self.waypoint {
```
davidiw, line 71: nit, have a new line between complex logic

config/management/src/waypoint.rs
```diff
@@ -40,19 +41,82 @@ impl CreateWaypoint {
...
+        } else {
+            return Err(Error::CommandArgumentError(
+                "please provide either --waypoint or --remote".to_string(),
+            ));
+        };
 
+        let waypoint = Waypoint::from_str(&waypoint_string)
+            .map_err(|e| Error::UnexpectedError(e.to_string()))?;
+        let mut local_storage: BoxedStorage = self.secure_backends.local.try_into()?;
+        InsertWaypoint::insert_waypoint_to_backend(
```
davidiw, line 87: nit if you're in a struct just use Self instead of the struct's name.

Rustinante:
Great! I just fixed them. Please also let me know which other task I should pick up next, thank you :-)

PR #4429, opened 2020-06-11
[mempool][easy] inline function used in one place

mimoo:


mempool/src/core_mempool/transaction_store.rs
```diff
@@ -143,27 +160,6 @@ impl TransactionStore {
         OP_COUNTERS.set("txn.priority_index", self.priority_index.size());
     }
 
-    /// checks if Mempool is full
-    /// If it's full, tries to free some space by evicting transactions from ParkingLot
-    /// We only evict on attempt to insert a transaction that would be ready for broadcast upon insertion
-    fn check_if_full(&mut self, txn: &MempoolTransaction, curr_sequence_number: u64) -> bool {
```
davidiw, line 149: would argue the naming isn't great "check_if_full" implies read-only

maybe garbage_collect is a better term?

mimoo:
Please let's not do this, as it makes the code harder to read (cf http://number-none.com/blow/john_carmack_on_inlined_code.html)

christinacelee:
I think this is more a matter of style, I can see arguments going for both ways. We can do this democratically, and if someone also agrees this PR is an improvement, they can approve it and land it

kphfb:
While I completely agree with the premise of improving code readability, I'd actually argue that this makes the code harder to parse.  Anyone doing development these days should be using a modern IDE where they can jump between code.  But clearly named functions that tell what is happening without showing the whole level of functionality is a pretty common pattern and generally fairly commonly taught as the "right" way to do things. But this really boils down to a style preference where there isn't really a right or wrong way to do this and everyone will have their own preference.  So I guess tl;dr, I find this to be harder to read than having the function just be renamed to make it clear what it's doing

mimoo:
Thanks for chiming-in. I'm pretty passionate about this so I'll explain my point of view further:

I think everyone can agree that layers of indirection and abstraction can increase the complexity on the reader side. 
It's of course not always true, good use of abstraction actually decreases the burden for the reader.
But there's a balance. Every time you use abstraction you need to ask yourself: does this make the code harder to read?

What Carmack outlines is a pattern that you'll find in many high-assurance software: **make main loops and critical codepaths as clear as possible**. The amount of lines of code has a direct correlation to the amount of bugs you'll find (that's proven) and it is well-known that complexity is the enemy of security.

Keep in mind that the use of that function (before the PR):

- decreases the burden of a "I want to understand what's going on" reviewer
- but increases the burden of a "I want to make sure everything is done correctly" reviewer

I'm obviously the latter one so this is why I'm still here to argue about this.
Never the less I can also pick my battles, so I'll close this PR for now :)

PR #4432, opened 2020-06-11
[network][noise] adding test for handshake replay

mimoo:


network/src/noise/handshake.rs
```diff
@@ -523,4 +529,46 @@ mod test {
...
+
+        client_session.unwrap();
+        server_session.unwrap();
+
+        // 3. perform the handshake again with the same timestamp, it should fail
+        let (dialer_socket, listener_socket) = MemorySocket::new_pair();
+        let (client_session, server_session) = block_on(join(
+            client.upgrade_outbound(dialer_socket, server_public, bad_timestamp(0)),
+            server.upgrade_inbound(listener_socket),
+        ));
```
phlip9, line 559: just need one more case: test that it also fails with a timestamp in the past, not just the same timestamp.

mimoo: good point!

PR #4433, opened 2020-06-11
[network] support server only authentication

davidiw:
This does several things:
1) Adds a proper smoke test to ensure this works
2) fixes the naming of enable_remote_authentication
3) handles all the quirks of networking to make it work

It then does to ancillary things that are relevant for simplifying configs:
1) Move all gossip config stuff into a GossipConfig
2) Eliminate the NetworkPeerInfo / NetworkPublicKeys in favor of just exposing the x25519::PublicKey

### DIFF
--- a/config/management/src/smoke_test.rs
+++ b/config/management/src/smoke_test.rs
@@ -6,7 +6,7 @@ use config_builder::{BuildSwarm, SwarmConfig};
 use libra_config::{
     config::{
         DiscoveryMethod, Identity, NetworkConfig, NodeConfig, OnDiskStorageConfig, RoleType,
-        SecureBackend, WaypointConfig,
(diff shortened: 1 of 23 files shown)


config/management/src/smoke_test.rs
```diff
@@ -153,8 +158,97 @@ fn smoke_test() {
...
+        config: SwarmConfig::build(&management_builder, &swarm_path).unwrap(),
+    };
+
+    swarm.launch_attempt(RoleType::FullNode, false).unwrap();
+    assert!(check_connectivity(&mut swarm, 1));
+}
+
+fn check_connectivity(swarm: &mut LibraSwarm, expected_peers: i64) -> bool {
+    for node in swarm.nodes.iter_mut() {
+        let mut timedout = true;
```
gregnazario, line 186: Minor: I'd change to `timed_out`

PR #4439, opened 2020-06-11
[libra-trace] Add custom sampling to libra trace

changyonghu:
Using default sampling rate to sample trace event.
Using the leading 6 hex of account address to hash for txn and using the leading 6 hex of block hashvalue to hash. Any suggestion? I found there is an event_edge (txn -> block), if the block sampling rate is not 1:1, then potentially, we would sample some transactions such that we lose the block that the transaction belongs to.

### DIFF
--- a/common/debug-interface/src/lib.rs
+++ b/common/debug-interface/src/lib.rs
@@ -1,8 +1,6 @@
 // Copyright (c) The Libra Core Contributors
 // SPDX-License-Identifier: Apache-2.0
 
-#![forbid(unsafe_code)]
-
 use crate::json_log::JsonLogEntry;
 use anyhow::Result;
 use reqwest::blocking;
@@ -13,8 +11,12 @@ pub mod libra_trace;
 pub mod node_debug_service;
 
 pub mod prelude {
-    pub use crate::{end_trace, event, trace_code_block, trace_edge, trace_event};
+    pub use crate::{
+        end_trace, event, node_sampling_data, trace_code_block, trace_edge, trace_event,
+    };
 }
+
+pub use libra_trace::{is_selected, libra_trace_set, set_libra_trace};
 
 /// Implement default utility client for NodeDebugInterface
 pub struct NodeDebugClient {
--- a/common/debug-interface/src/libra_trace.rs
+++ b/common/debug-interface/src/libra_trace.rs
@@ -2,15 +2,39 @@
 // SPDX-License-Identifier: Apache-2.0
 
 use crate::json_log::JsonLogEntry;
+use anyhow::{bail, ensure, Result};
 use std::time::Instant;
 
 pub const TRACE_EVENT: &str = "trace_event";
 pub const TRACE_EDGE: &str = "trace_edge";
 
+use std::{
+    collections::HashMap,
+    sync::atomic::{AtomicUsize, Ordering},
+};
+
+// This is poor's man AtomicReference from crossbeam
+// It have few unsafe lines, but does not require extra dependency
+// Sampling rate is the form of (nominator, denominator)
+static mut SAMPLING_CONFIG: Option<Sampling> = None;
+static LIBRA_TRACE_STATE: AtomicUsize = AtomicUsize::new(UNINITIALIZED);
+const UNINITIALIZED: usize = 0;
+const INITIALIZING: usize = 1;
+const INITIALIZED: usize = 2;
+
+struct Sampling(HashMap<&'static str, CategorySampling>);
+
+struct CategorySampling {
+    denominator: u64,
+    nominator: u64,
+}
+
 #[macro_export]
 macro_rules! trace_event {
     ($stage:expr, $node:tt) => {
-        trace_event!($stage; {$crate::format_node!($node), module_path!(), Option::<u64>::None})
+        if $crate::is_selected($crate::node_sampling_data!($node)) {
+            trace_event!($stage; {$crate::format_node!($node), module_path!(), Option::<u64>::None})
+        }
     };
     ($stage:expr; {$node:expr, $path:expr, $duration:expr}) => {
         $crate::json_log::send_json_log($crate::json_log::JsonLogEntry::new(
@@ -22,6 +46,23 @@ macro_rules! trace_event {
                "duration": $duration,
             }),
         ));
+    }
+}
+
+#[macro_export]
+macro_rules! node_sampling_data {
+    ({$type:expr, $($rest:tt)*}) => {{
+        use std::hash::Hasher;
+        let mut hasher = std::collections::hash_map::DefaultHasher::new();
+        node_sampling_data!(hasher; $($rest)*);
+        ($type, hasher.finish())}
+    };
+    ($hasher:expr; $p:expr, $($rest:tt)*) => {
+        std::hash::Hash::hash(&$p, &mut $hasher);
+        node_sampling_data!($hasher; $($rest)*);
+    };
+    ($hasher:expr; $p:expr) => {
+        std::hash::Hash::hash(&$p, &mut $hasher);
     }
 }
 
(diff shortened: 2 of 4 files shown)


changyonghu:
> FYI, u should create pr as draft

I forgot to edit the comments, it is ready for review.

common/debug-interface/src/lib.rs
```diff
@@ -1,7 +1,7 @@
 // Copyright (c) The Libra Core Contributors
 // SPDX-License-Identifier: Apache-2.0
 
-#![forbid(unsafe_code)]
+// #![forbid(unsafe_code)]
```
andll, line 4: Please delete it, don't leave commented out code

common/debug-interface/src/libra_trace.rs
```diff
@@ -7,10 +7,25 @@ use std::time::Instant;
 pub const TRACE_EVENT: &str = "trace_event";
 pub const TRACE_EDGE: &str = "trace_edge";
 
+use std::sync::atomic::{AtomicUsize, Ordering};
+
+// This is poor's man AtomicReference from crossbeam
+// It have few unsafe lines, but does not require extra dependency
+// Sampling rate is the form of (nominator, denominator)
+static mut TXN_SAMPLING_RATE: (u64, u64) = (1, 100);
```
andll, line 15: Can we have single static var like `SAMPLING_CONFIG` of type `SamplingConfig`?

common/debug-interface/src/libra_trace.rs
```diff
@@ -296,3 +317,47 @@ fn abbreviate_crate(name: &str) -> &str {
...
+pub fn libra_trace_set() -> bool {
+    LIBRA_TRACE_STATE.load(Ordering::SeqCst) == INITIALIZED
+}
+
+pub fn is_selected(s: String) -> bool {
+    let v = s.split("::").collect::<Vec<&str>>();
+    if v.len() != 2 && v.len() != 3 {
+        panic!("Invalid node format");
+    }
+    if v[0] == "txn" {
```
andll, line 354: I think this should be generalized, without hardcoding "txn" in libra trace code.

You can declare SamplingConfig from config crate as following:
`struct SamplingConfig(HashMap<String, String>)`.
Then in libra_trace you can define separate class
```
struct CategorySampling{denominator: u64, nominator: u64}

struct Sampling(HashMap<&'static str, CategorySampling>)
```
You can have a method that will parse SamplingConfig you get from a file into Sampling, and have your global var to hold `Sampling` instance.
Then you don' need separate handling for txn/block, and if someone adds another type of sampling, they can just add it to config as a string.

common/debug-interface/src/libra_trace.rs
```diff
@@ -79,30 +101,36 @@ impl Drop for TraceBlockGuard {
...
-            $crate::libra_trace::TRACE_EVENT,
-            serde_json::json!({
-               "path": module_path!(),
-               "node": $crate::format_node!($node),
-               "stage": $stage,
-               "end": true,
-            }),
-        ));
+        let node = $crate::format_node!($node);
+        if $crate::is_selected(node.to_string()) {
```
andll, line 105: Looks like what we do here currently is first 'gluing' `{node, xxx}` into string and then in is_selected we are parsing it back.
I think we could do it differently - make a separate macro (similar to `format_node`) that will select first element in $node and then use it in is_selected. We can assume that first element in `$node` is always `&str`. This way even `format_node` code can be inside `if` and we can skip formatting node if it does not need to be sampled.

common/debug-interface/src/libra_trace.rs
```diff
@@ -296,3 +333,60 @@ fn abbreviate_crate(name: &str) -> &str {
...
+    )
+}
+
+/// Checks if libra trace is enabled
+pub fn libra_trace_set() -> bool {
+    LIBRA_TRACE_STATE.load(Ordering::SeqCst) == INITIALIZED
+}
+
+pub fn is_selected(node: (&str, &Vec<u8>)) -> bool {
+    unsafe {
```
andll, line 379: We should verify if LIBRA_TRACE_STATE is initialized before reading `&SAMPLING_CONFIG` it to avoid partial reads

common/debug-interface/src/libra_trace.rs
```diff
@@ -296,3 +333,60 @@ fn abbreviate_crate(name: &str) -> &str {
...
+
+fn parse_sampling_config(config: &HashMap<String, String>) -> Sampling {
+    Sampling(
+        config
+            .iter()
+            .map(|(k, v)| {
+                let k: &'static str = Box::leak(k.clone().into_boxed_str());
+                let v = v.split('/').collect::<Vec<&str>>();
+                let v = CategorySampling {
+                    nominator: v[0].parse::<u64>().unwrap(),
```
andll, line 364: Perhaps `parse_sampling_config` could return anyhow::Result instead of unwrapping here?

common/debug-interface/src/libra_trace.rs
```diff
@@ -296,3 +333,60 @@ fn abbreviate_crate(name: &str) -> &str {
         _ => name,
     }
 }
+
+// This is exact copy of similar function in log crate
+/// Sets libra trace config
+pub fn set_libra_trace(config: &HashMap<String, String>) -> Result<(), ()> {
+    unsafe {
```
andll, line 340: We should probably parse config outside of unsafe block(and return error, change this fn to anyhow::Result instead of Result<(), ()>).

common/debug-interface/src/libra_trace.rs
```diff
@@ -3,14 +3,37 @@
...
+// It have few unsafe lines, but does not require extra dependency
+// Sampling rate is the form of (nominator, denominator)
+static mut SAMPLING_CONFIG: Option<Sampling> = None;
+static LIBRA_TRACE_STATE: AtomicUsize = AtomicUsize::new(UNINITIALIZED);
+const UNINITIALIZED: usize = 0;
+const INITIALIZING: usize = 1;
+const INITIALIZED: usize = 2;
+
+struct Sampling(HashMap<&'static str, CategorySampling>);
+struct CategorySampling {
```
andll, line 26: New line between structs please

common/debug-interface/src/libra_trace.rs
```diff
@@ -296,3 +333,60 @@ fn abbreviate_crate(name: &str) -> &str {
...
+/// Checks if libra trace is enabled
+pub fn libra_trace_set() -> bool {
+    LIBRA_TRACE_STATE.load(Ordering::SeqCst) == INITIALIZED
+}
+
+pub fn is_selected(node: (&str, &Vec<u8>)) -> bool {
+    unsafe {
+        match &SAMPLING_CONFIG {
+            Some(Sampling(sampling)) => {
+                // Take the leading 6 hex of account address or block hashvalue to hash
```
andll, line 382: I don't think we can take first few letters of value and parse them as hex - that's not very universal and seem tailored specifically for txn/blocks.

What we can do instead is this: instead of `parse_node` have macro `node_sampling_data -> (&'static str, u64)`, where `&'static str` will be same as what you currently return (first element) and `u64` will be proper hash of remaining elements

This 'proper hash' can be calculated by initializing `std::collections::hash_map::DefaultHasher` and then calling `.hash(&mut hasher)` on each remaining element of the node.

You can see `format_struct_args` in logger create for an example of how you can recursively iterate through `$rest` in a macro. (example is logger create much more scary, in your case it should be much shorter).

One of the problems in current approach you are taking(aside from it not being universal) is that you calculate transaction sampling based on block id (first element in tuple). That's pretty bad as it would generate big spike of logging for each block that won the lottery

common/debug-interface/src/libra_trace.rs
```diff
@@ -2,15 +2,40 @@
...
+struct CategorySampling {
+    denominator: u64,
+    nominator: u64,
+}
+
 #[macro_export]
 macro_rules! trace_event {
     ($stage:expr, $node:tt) => {
-        trace_event!($stage; {$crate::format_node!($node), module_path!(), Option::<u64>::None})
+        let node = $crate::node_sampling_data!($node);
```
andll, line 35: node_sampling_data called twice?

common/debug-interface/src/libra_trace.rs
```diff
@@ -296,3 +342,58 @@ fn abbreviate_crate(name: &str) -> &str {
...
+    }
+    Ok(Sampling(map))
+}
+
+/// Checks if libra trace is enabled
+pub fn libra_trace_set() -> bool {
+    LIBRA_TRACE_STATE.load(Ordering::SeqCst) == INITIALIZED
+}
+
+pub fn is_selected(node: (&'static str, u64)) -> bool {
```
andll, line 388: Can we check `libra_trace_set` here?
Currently you do it in every call site before is_selected, there are few problems with that
a) Code duplication
b) In the future, someone might forgot to check libra_trace_set before calling is_select, which is unsafe.

common/debug-interface/src/libra_trace.rs
```diff
@@ -296,3 +342,58 @@ fn abbreviate_crate(name: &str) -> &str {
...
+    unsafe {
+        match &SAMPLING_CONFIG {
+            Some(Sampling(sampling)) => {
+                let sampling_rate = sampling.get(node.0).expect("Invalid sampling category");
```
andll, line 392: We should not file if sampling category not in config, we should just return true(assume no sampling and all recorded)

common/debug-interface/src/libra_trace.rs
```diff
@@ -296,3 +342,58 @@ fn abbreviate_crate(name: &str) -> &str {
...
+        },
+        Err(s) => bail!("Failed to parse sampling config: {}", s),
+    }
+}
+
+fn parse_sampling_config(config: &HashMap<String, String>) -> Result<Sampling> {
+    let mut map = HashMap::new();
+    for (k, v) in config {
+        let k: &'static str = Box::leak(k.clone().into_boxed_str());
+        let v = v.split('/').collect::<Vec<&str>>();
```
andll, line 373: Can we check v.length==2?

common/debug-interface/src/libra_trace.rs
changyonghu: Oh forget to delete this one.

common/debug-interface/src/libra_trace.rs
changyonghu: Yes, this makes more sense.

common/debug-interface/src/libra_trace.rs
```diff
@@ -296,3 +341,70 @@ fn abbreviate_crate(name: &str) -> &str {
...
+    Ok(Sampling(map))
+}
+
+/// Checks if libra trace is enabled
+pub fn libra_trace_set() -> bool {
+    LIBRA_TRACE_STATE.load(Ordering::SeqCst) == INITIALIZED
+}
+
+pub fn is_selected(node: (&'static str, u64)) -> bool {
+    if libra_trace_set() {
```
andll, line 393: nit: for situations when you have early termination condition, best practice is to do something like
```
if !trace_set() {
return false
}
...do work....
```

This allows to reduce big code blocks that are hard to read

PR #4440, opened 2020-06-11
Network builder/refactor everything

agc-quadra:
## Motivation

NOT FOR SUBMISSION

This is a draft PR to show a mostly complete end-state for refactoring the network_builder.

If the end-state looks reasonable, then there is a set of smaller PRs that needs to be worked through to get to this state in an incremental and workable fashion.

yes

## Test Plan

## Related PRs

### DIFF
--- a/network/builder/src/builder.rs
+++ b/network/builder/src/builder.rs
@@ -9,7 +9,20 @@
 //! authentication -- a network end-point running with remote authentication enabled will
 //! connect to or accept connections from an end-point running in authenticated mode as
 //! long as the latter is in its trusted peers set.
-use crate::{
+use channel::{self, libra_channel, message_queues::QueueStyle};
+use futures::stream::StreamExt;
+use libra_config::{
+    config::{DiscoveryMethod, NetworkConfig, RoleType, HANDSHAKE_VERSION},
+    network_id::{NetworkContext, NetworkId},
+};
+use libra_crypto::x25519;
+use libra_logger::prelude::*;
+use libra_metrics::IntCounterVec;
+use libra_network_address::NetworkAddress;
+use libra_secure_storage::config;
+use libra_types::{waypoint::Waypoint, PeerId};
+use netcore::transport::{memory, Transport};
+use network::{
     connectivity_manager::{ConnectivityManager, ConnectivityRequest},
     constants, counters,
     peer_manager::{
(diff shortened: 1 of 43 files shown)


agc-quadra:
@davidiw 

This shows the (almost) final target for network_builder

The PR is not submittable in its current form because of some smoke_test breaks resulting from overly big changes on my part.  If the end-state looks reasonable, then I will start to pull isolated changes off for incremental review.

network/builder/src/builder.rs
```diff
@@ -237,8 +235,7 @@ impl NetworkBuilder {
             DiscoveryMethod::Gossip(gossip_config) => {
                 network_builder
                     .advertised_address(gossip_config.advertised_address.clone())
```
davidiw, line 294: can we not do the same with advertised address?

network/src/connectivity_manager/test.rs
```diff
@@ -45,17 +45,21 @@ fn setup_conn_mgr(
...
+    let conn_mgr_config = ConnectivityManagerConfig::new(
+        network_context,
+        Arc::new(RwLock::new(eligible_peers)),
+        seed_peers,
+        ticker_rx,
+        FixedInterval::from_millis(100),
+        300, /* ms */
+    );
+
     let conn_mgr = {
```
davidiw, line 57: what is the purpose of these extra `{` and `}`

network/src/connectivity_manager/mod.rs
```diff
@@ -130,27 +130,57 @@ struct DialState<TBackoff> {
     addr_idx: usize,
 }
 
-impl<TTicker, TBackoff> ConnectivityManager<TTicker, TBackoff>
+/// The configuration fields for ConnectivityManager
```
davidiw, line 133: what's the motivation for splitting this? I actually removed a lot of these "config" structures from consensus because it felt like I was jumping through hoops to instantiate the service(s).

network/src/connectivity_manager/test.rs
agc-quadra: there is none.

network/src/connectivity_manager/mod.rs
agc-quadra: 1.  Hierarchically apply the Builder pattern at every level.  main_node can be viewed as node_builder and needs to glue network/consensus/statesync/mempool together to form the node.  network_builder needs to glue discovery, peermanager, connectivitymanager, ... together to form the network.
1.  Each builder should have a complete configuration which is sufficient to build the underlying component.  This may require some amount of logic (i.e., setup_network should actually be a part of NetworkBuilder, and is logic applied to configuring internal components and wiring them together.
1.  Overwhelming majority of the 'shared' fields in network_builder can be removed/moved into the component specific configs so network_builder is a high level view of the system (which components exist) and is not tied to internal details of those components
1.  A fair number of the hoops and circular dependencies are tied to configuring/building/starting the components all at once.   Having the configs does two things.  First, it allows for isolating configure/build/start into separate passes (helpful for dealing with dependencies).  Second,  it makes components look the same regardless of where they sit in the system hierarchy and helps to enforce abstraction isolation.

Caveat:  explicit config splitting is most useful if the configure/build/start phases are explicitly split from each other.  if they bleed together then the configs simply exacerbate the already existing problem.

PR #4443, opened 2020-06-11
[dep] update crypto libs

rexhoffman:
## Motivation

Update crypto deps.

y

## Test Plan

CI

## Related PRs

https://github.com/libra/libra/pull/4437
https://github.com/libra/libra/pull/4436
https://github.com/libra/libra/pull/4435
https://github.com/libra/libra/pull/4396
https://github.com/libra/libra/pull/4394
https://github.com/libra/libra/pull/4392
https://github.com/libra/libra/pull/4391


crypto/crypto/src/hkdf.rs
```diff
@@ -116,9 +119,9 @@ where
             None => Hmac::<D>::new(&Default::default()),
         };
 
-        hmac.input(ikm);
+        hmac.update(ikm);
 
-        Ok(hmac.result().code().to_vec())
+        Ok(hmac.clone().finalize().into_bytes().to_vec())
```
bmwill, line 124: You probably don't need the clone here

crypto/crypto/src/hkdf.rs
```diff
@@ -138,14 +141,15 @@ where
...
             if let Some(_info) = info {
-                hmac.input(_info);
+                hmac.update(_info);
             }
-            hmac.input(&[blocknum as u8 + 1]);
+            hmac.update(&[blocknum as u8 + 1]);
 
-            let output = hmac.result_reset().code();
+            let output = hmac.clone().finalize().into_bytes();
+            hmac.reset();
```
bmwill, line 152: Instead of cloning and then resetting you should use [`finalize_reset`](https://docs.rs/digest/0.9.0/digest/trait.Digest.html#tymethod.finalize_reset)

testsuite/libra-fuzzer/src/commands.rs
```diff
@@ -39,9 +39,7 @@ pub fn make_corpus(
             }
         };
 
-        // Use the SHA-1 of the result as the file name.
-        sha1.input(&result);
-        let hash = sha1.result_reset();
+        let hash = sha1.clone().finalize();
```
bmwill, line 42: Cloning here also looks suspicious

crypto/crypto/src/hkdf.rs
rexhoffman: yep, mut var. good catch.

crypto/crypto/src/hkdf.rs
```diff
@@ -89,15 +91,16 @@ use thiserror::Error;
 #[derive(Clone, Debug)]
 pub struct Hkdf<D>
 where
-    D: Input + BlockInput + FixedOutput + Reset + Default + Clone,
+    D: Update + BlockInput + FixedOutput + Reset + Default + Clone,
+    D::BlockSize: ArrayLength<u8>,
```
huitseeker, line 95: The more I look at this, the more I think this could leverage the [hkdf](https://github.com/RustCrypto/KDFs/tree/master/hkdf) crate more efficiently, and that not doing so leaves us with the burden of maintening this. Unless @kchalkias comes up with a good reaons why we're doing somethign fundamentally different here, could we open an issue to replace this with hkdf? Possibly label it help-wanted?

testsuite/libra-fuzzer/src/commands.rs
```diff
@@ -39,9 +39,7 @@ pub fn make_corpus(
             }
         };
 
-        // Use the SHA-1 of the result as the file name.
-        sha1.input(&result);
```
huitseeker, line 43: What happened for this to go away? Does the fuzzer no longer need a uinque file name?

rexhoffman: gah, good catch, thank you

crypto/crypto/src/hkdf.rs
kchalkias: There are two major updated to the original crate, this is one is generic for any hash function >=256 bits and does not allow empty messages + a few helper functions.

rexhoffman: @kchalkias - is the pr good to merge?   Would like to merge it before other dependency prs?

PR #4449, opened 2020-06-11
[core modules] Centralize core account addresses.

tzakian:
This is the first step towards #4267. This just centralizes all of the core code addresses in one module. Note that the function naming conventions are intentionally violated here since they will eventually be replaced with constants once we have those.


language/stdlib/modules/Addresses.move
```diff
@@ -0,0 +1,29 @@
...
+
+    public fun TRANSACTION_FEE_ADDRESS(): address {
+        0xFEE
+    }
+
+    public fun DEFAULT_CONFIG_ADDRESS(): address {
+        0xF1A95
+    }
+}
+}
```
sblackshear, line 29: <3

language/stdlib/modules/Addresses.move
```diff
@@ -0,0 +1,29 @@
+address 0x0 {
+module CoreAddresses {
+    public fun ASSOCIATION_ROOT_ADDRESS(): address {
```
sblackshear, line 3: Comments for each of these might be useful

language/stdlib/modules/Addresses.move
```diff
@@ -0,0 +1,29 @@
+address 0x0 {
+module CoreAddresses {
```
sblackshear, line 2: Nit: rename either the file or the module?

tzakian: Woops...forgot to rename the file after renaming the module....good spot!

PR #4450, opened 2020-06-11
[network][connmgr] reset both address and backoff on update

phlip9:
When receiving an updated set of address for a peer, we reset their dial
state (if we're already trying to dial them). As a result, we will begin
our next dial attempt from the first address (which might have changed)
and from a fresh backoff (since the current backoff delay might be maxed
out if we can't reach any of their previous addresses).

This change should help to mitigate some of the issues we've been having
with paritions healing very slowly (2-5 min!) when a peer has some
invalid or unreachable addresses in on-chain discovery.

This change also sets the maximum retry delay to 1 minute (up from 10
seconds prev, down from 10 minutes prev^2). 10 seconds is too aggressive
with this change and spams the logs too much. We already have logic to
reduce the max delay if we're not fully connected.

### DIFF
--- a/network/src/connectivity_manager/mod.rs
+++ b/network/src/connectivity_manager/mod.rs
@@ -44,7 +44,7 @@ use num_variants::NumVariants;
 use std::{
     cmp::min,
     collections::HashMap,
-    fmt,
+    fmt, mem,
     sync::{Arc, RwLock},
     time::{Duration, Instant},
 };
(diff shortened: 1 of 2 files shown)


andll:
Question - do you want to get any data points to actually confirm how this change affected recovery?

As a cluster test advertising - one option for this, for example 10% down experiment and compare graphs between master and this diff to see if recovery time improves?
You could do this by those two commands:
```
# Reference
./scripts/cti --master --run bench -- --percent-nodes-down 10
# This PR
./scripts/cti --pr <PR> --run bench -- --percent-nodes-down 10
```
They print link to dashboard in the end, so in theory those links can be used to see a difference

davidiw:
The alignment of the comment looks weird... is it at 80 chars and the code at 100?

phlip9:
looks like the cluster initially connects much more quickly with this fix 

current master:

Screen Shot 2020-06-11 at 13 25 57

with pr:

Screen Shot 2020-06-11 at 13 27 34

andll:
Thats awesome!
First picture(master) is exactly problem that we seen today with @sherry-x  in premainnet drill

phlip9:
forgot the top level metrics : p

```
before: 10% down : 1285 TPS, 3160 ms latency, 7550 ms p99 latency, no expired txns
after:  10% down : 1333 TPS, 3050 ms latency, 4350 ms p99 latency, no expired txns
```

PR #4452, opened 2020-06-11
[language][vm] remove content length from binary format

vgao1996:
## Summary
Content length can be computed from table indices and therefore does not need to be stored in the binary. This gets it removed.

## Test Plan
`cargo test`


language/vm/src/deserializer.rs
```diff
@@ -178,6 +176,7 @@ fn read_table(cursor: &mut Cursor<&[u8]>) -> BinaryLoaderResult<Table> {
 
 fn read_table_contents(cursor: &mut Cursor<&[u8]>, n: usize) -> BinaryLoaderResult<Vec<u8>> {
     let mut bytes = vec![];
+    // TODO: can we rewrite this in a better way?
```
dariorussi, line 179: do we need this TODO?

vgao1996: No harm to keep it. There must be a way to read the whole blob without making so many IO calls.

PR #4455, opened 2020-06-11
[cluster-test] Add --lsr-backend flag to cluster-test

ankushagarwal:
## Summary

* `--lsr-backend` lets you specify the storage backend used by LSR. Options are in-memory, on-disk, vault
* Add health check to validators to ensure all LSRs are healthy before validators start
* Add health check to LSR to ensure all vaults are healthy before LSRs start 

## Test Plan

Ran against different backends


andll:
@bors-libra delegate+
Looks great, thank you!

PR #4457, opened 2020-06-12
[network] added chain-id to differentiate chains

gregnazario:
The chain-id is to differentiate between different blockchains, and is
enforced at the handshake level similar to network-id.

### DIFF
--- a/config/src/config/mod.rs
+++ b/config/src/config/mod.rs
@@ -43,7 +43,7 @@ pub use safety_rules_config::*;
 mod upstream_config;
 pub use upstream_config::*;
 mod test_config;
-use crate::network_id::NetworkId;
+use crate::{chain_id::ChainId, network_id::NetworkId};
 use libra_types::waypoint::Waypoint;
 pub use test_config::*;
 
@@ -89,6 +89,7 @@ pub struct NodeConfig {
 #[serde(default, deny_unknown_fields)]
 pub struct BaseConfig {
     data_dir: PathBuf,
+    pub chain_id: ChainId,
     pub role: RoleType,
     pub waypoint: WaypointConfig,
 }
@@ -97,6 +98,7 @@ impl Default for BaseConfig {
     fn default() -> BaseConfig {
         BaseConfig {
             data_dir: PathBuf::from("/opt/libra/data/commmon"),
+            chain_id: ChainId::default(),
             role: RoleType::Validator,
             waypoint: WaypointConfig::None,
         }
--- a/libra-node/src/main_node.rs
+++ b/libra-node/src/main_node.rs
@@ -8,6 +8,7 @@ use executor::{db_bootstrapper::bootstrap_db_if_empty, Executor};
 use executor_types::ChunkExecutor;
 use futures::{channel::mpsc::channel, executor::block_on};
 use libra_config::{
+    chain_id::ChainId,
     config::{DiscoveryMethod, NetworkConfig, NodeConfig, RoleType},
     utils::get_genesis_txn,
 };
@@ -62,8 +63,9 @@ fn setup_debug_interface(config: &NodeConfig) -> NodeDebugService {
 
 // TODO(abhayb): Move to network crate (similar to consensus).
 pub fn setup_network(
-    config: &mut NetworkConfig,
+    chain_id: &ChainId,
     role: RoleType,
+    config: &mut NetworkConfig,
     libra_db: Arc<dyn DbReader>,
     waypoint: Waypoint,
 ) -> (Runtime, NetworkBuilder) {
@@ -79,9 +81,10 @@ pub fn setup_network(
 
     let mut network_builder = NetworkBuilder::new(
         runtime.handle().clone(),
+        chain_id.clone(),
         config.network_id.clone(),
-        peer_id,
         role,
+        peer_id,
         config.listen_address.clone(),
     );
     network_builder.add_connection_monitoring();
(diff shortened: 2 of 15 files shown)


config/src/config/network_config.rs
```diff
@@ -2,6 +2,7 @@
 // SPDX-License-Identifier: Apache-2.0
 
 use crate::{
+    chain_id::ChainId,
```
davidiw, line 5: So do we duplicate this across every network config or insert it into the BaseConfig?

network/src/protocols/identity.rs
```diff
@@ -59,10 +59,11 @@ mod tests {
     #[test]
     fn simple_handshake() {
         let network_id = NetworkId::Validator;
-        let (mut outbound, mut inbound) = build_test_connection();
+        let chain_id = ChainId::default();
+        let (outbound, inbound) = build_test_connection();
 
         // Create client and server handshake messages.
-        let mut server_handshake = HandshakeMsg::new(network_id.clone());
+        let mut server_handshake = HandshakeMsg::new(network_id.clone(), chain_id.clone());
```
davidiw, line 66: my mental model is chain >> network
if you agree, let's make that ordering consistent

network/src/protocols/identity.rs
```diff
@@ -83,28 +84,24 @@ mod tests {
...
-            assert_eq!(
-                lcs::to_bytes(&handshake).unwrap(),
-                lcs::to_bytes(&server_handshake_clone).unwrap()
-            );
-        };
+    async fn handshake(
+        remote_msg: &HandshakeMsg,
+        self_msg: &HandshakeMsg,
+        mut socket: MemorySocket,
+    ) {
```
davidiw, line 97: why the split here? if we're not reusing this function, the separation seems a little awkward as it doesn't reduce the the actual primary function's length much.

network/src/protocols/wire/handshake/v1/mod.rs
```diff
@@ -122,6 +124,10 @@ impl HandshakeMsg {
             .insert(messaging_protocol, application_protocols);
     }
 
+    pub fn represents_same_network(&self, other: &HandshakeMsg) -> bool {
```
davidiw, line 127: maybe just verify?

network/src/protocols/wire/handshake/v1/mod.rs
```diff
@@ -63,6 +63,7 @@ pub struct SupportedProtocols(bitvec::BitVec);
 pub struct HandshakeMsg {
     pub supported_protocols: BTreeMap<MessagingProtocolVersion, SupportedProtocols>,
     pub network_id: NetworkId,
+    pub chain_id: ChainId,
```
davidiw, line 66: per messaging earlier, think about the ordering here, cause in terms of evaluation, it would seem we first check universe (chain id), before we check galaxy (network id)... horrible analogy :P

gregnazario: This was because of TOML for the ordering.  Once your PR goes through, we can reorder everything :)

network/src/protocols/identity.rs
gregnazario: Yeah, I think in general we should reorder a lot of these things to some sort of canonical order, and chain_id is definitely > network

config/src/config/network_config.rs
gregnazario: The BaseConfig is probably a better place to put it, either way it has to be trickled down to handshake.  I was considering putting it into the NetworkContext, but realize it probably doesn't have to be logged all that much.

network/src/protocols/identity.rs
gregnazario: CLion was just telling me it was duplicated, and I felt like collapsing it, but I can put it back.  It was actually helpful for me to understand `async` and how a bunch of borrowing around it worked.

davidiw: if we could merge, that would make me happy, but I'll leave that to you.

config/src/config/mod.rs
```diff
@@ -93,6 +93,7 @@ pub struct NodeConfig {
 #[serde(default, deny_unknown_fields)]
 pub struct BaseConfig {
     data_dir: PathBuf,
+    pub chain_id: ChainId,
```
davidiw, line 96: Should we just make this a string? What benefit do we have by making ChainId a struct?

libra-node/src/main_node.rs
```diff
@@ -62,8 +63,9 @@ fn setup_debug_interface(config: &NodeConfig) -> NodeDebugService {
 
 // TODO(abhayb): Move to network crate (similar to consensus).
 pub fn setup_network(
```
aching, line 65: Could we add a info! level logging message with the chain-id?  This could be helpful in debugging situations where some nodes are on different chain-ids and all we have are the logs.

aching: Actually, scratch this, looks like all config is already printed.

config/src/config/mod.rs
gregnazario: The idiom of the newtype struct is to prevent from clashing between different "strings" (https://doc.rust-lang.org/stable/rust-by-example/generics/new_types.html)

It could just be a string, since it's simply an identifier.

network/src/protocols/identity.rs
gregnazario: Do you mean change it back by merge?  Or the changed version I have here?

gregnazario:
I can't seem to find @aching's comment, but the whole config is logged during startup.

https://github.com/libra/libra/blob/master/libra-node/src/main.rs#L36

network/src/protocols/identity.rs
davidiw: the original, i.e., no need for this extra handshake function afaiu.

gregnazario: Changed back to the original.

PR #4458, opened 2020-06-12
[testnet] switch testnet minting from AssocRoot account to TreasuryCompliance account

sblackshear:
Today, minting in testnet and in all tests is performed from the AssocRoot account `0xA550C18`. This is different from the production configuration, where there is a single `MintCapability` given to the TreasuryCompliance account `0xB1E55ED`.

This PR changes tests/testnet to reflect production.
- remove `Libra::grant_mint_capability_to_association`. This was a temporary hack that violates an important invariant: the `MintCapability` for a given currency should be globally unique. This property should now be true.
- Add `treasury_compliance_account` to the client CLI. Minting operations now use this account. It uses the same keypair as AssocRoot for now (but need not)
- Fix all the tests that used `0xA550C18` for minting


sblackshear:
That is a very good point. I will check the faucet code and make sure--ideally it should be as simple as sending faucet txes from `0xB1E55ED` instead of `0xA550C18`.

sblackshear:
CC @christinacelee @thefallentree to make sure this won't break the faucet service.

thefallentree:
looks good to me since you already modified CLI which faucet uses.

PR #4461, opened 2020-06-12
[network][noise] early detection and prevention of self-dials

phlip9:
Imagine the following scenario:

We have two peers Alice (A) and Mallory (M). Alice is an honest peer
advertising her address via e.g. on-chain discovery
`"/../ln-noise-ik/<s_A>/.."`. In an effort to waste Alice's resources,
Mallory attempts to get Alice to inadvertently connect with herself:

First, Mallory rotates his set of advertised addresses to include one
that looks like `"/ip4/127.0.0.1/tcp/6180/ln-noise-ik/<s_A>/.."`. In
other words, Mallory is advertising one of his addresses as a local
loopback with _Alice's_ pubkey. After rotating, the on-chain discovery
information will look like:

```rust
{
    id_A: [ "/../ln-noise-ik/<s_A>/..", .. ],
    id_M: [ "/ip4/127.0.0.1/tcp/6180/ln-noise-ik/<s_A>/..", .. ],
    // ..
}
```

When other peers attempt to dial Mallory, they will (without additional
hardening) try to handshake with themselves. For all other peers
(except Alice), this handshake will fail since they clearly don't own
the `s_A` keypair. However, Alice will end up _successfully_ dialing
herself, since she does own the `s_A` keypair.

Fortunately, there's a simple way to prevent this from happening. We've
already begun sending the dialer's peer id in the handshake prologue. We
can just check that the dialer's peer id != listener's peer id. In the
above scenario, Alice (as the server) realizes that the inbound
connection is from herself (the peer id sent in the prologue is her
own), so she can then immediately drop the connection.

### DIFF
--- a/network/src/noise/fuzzing.rs
+++ b/network/src/noise/fuzzing.rs
@@ -46,10 +46,10 @@ impl ExposingSocket {
         }
     }
     fn new_pair() -> (Self, Self) {
-        let (dialer_socket, listener_socket) = MemorySocket::new_pair();
+        let (memsocket1, memsocket2) = MemorySocket::new_pair();
         (
-            ExposingSocket::new(dialer_socket),
-            ExposingSocket::new(listener_socket),
+            ExposingSocket::new(memsocket1),
+            ExposingSocket::new(memsocket2),
         )
     }
 }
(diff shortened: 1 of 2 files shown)


network/src/noise/handshake.rs
```diff
@@ -481,41 +497,61 @@ mod test {
...
     }
 
     fn test_handshake_success(is_mutual_auth: bool) {
         // perform handshake with two testing peers
         let ((client, client_public), (server, server_public)) = build_peers(is_mutual_auth);
-        let (client, (server, _)) = perform_handshake(client, server, server_public);
 
-        assert_eq!(client.get_remote_static(), server_public);
-        assert_eq!(server.get_remote_static(), client_public);
+        let (client_res, server_res) = perform_handshake(&client, &server, server_public);
```
gregnazario, line 521: I know this was already here, but `server_public_key` might be more clear to the reader.

network/src/noise/handshake.rs
```diff
@@ -298,6 +298,19 @@ impl NoiseUpgrader {
...
+        // this situation could occur either as a result of our own discovery
+        // mis-configuration or a potentially malicious discovery peer advertising
+        // a (loopback ip or mirror proxy) and our public key.
+        if remote_peer_id == self.self_peer_id {
+            // TODO(philiphayes): security logging. someone should investigate
+            // on-chain reconfiguration history to see if someone is misbehaving.
+            return Err(io::Error::new(
+                io::ErrorKind::InvalidData,
+                "noise: detected accidental self-dial: we have the same peer id as the client",
+            ));
```
gregnazario, line 311: I was thinking maybe we want to put `peer_id` in the error, but then I realized that may be added in logging in the future.  Thoughts?

network/src/noise/fuzzing.rs
```diff
@@ -87,35 +87,66 @@ impl AsyncRead for ExposingSocket {
...
+    );
+    let responder = NoiseUpgrader::new(
+        responder_peer_id,
+        responder_key,
+        HandshakeAuthMode::ServerOnly,
+    );
 
     // create exposing socket
-    let (dialer_socket, listener_socket) = ExposingSocket::new_pair();
+    let (initiator_socket, listener_socket) = ExposingSocket::new_pair();
```
gregnazario, line 130: I like `initiator` better than `dialer`

network/src/noise/handshake.rs
phlip9: self_peer_id should get logged w/ the NetworkContext

network/src/noise/fuzzing.rs
phlip9: woops! my find and replace grabbed a bit too much...

network/src/noise/handshake.rs
```diff
@@ -304,6 +304,19 @@ impl NoiseUpgrader {
             )
         })?;
 
+        // prevent accidental self-dials
+        // this situation could occur either as a result of our own discovery
+        // mis-configuration or a potentially malicious discovery peer advertising
+        // a (loopback ip or mirror proxy) and our public key.
+        if remote_peer_id == self.self_peer_id {
```
mimoo, line 311: do you want to update our network-noise spec with this change :) ? should be a one liner

network/src/noise/handshake.rs
```diff
@@ -448,19 +464,19 @@ mod test {
     ) {
         let mut rng = ::rand::rngs::StdRng::from_seed(TEST_SEED);
 
-        let client_private = x25519::PrivateKey::generate(&mut rng);
-        let client_public = client_private.public_key();
+        let client_key = x25519::PrivateKey::generate(&mut rng);
```
mimoo, line 467: pet peeve: to reflect `client_public_key` this should be `client_private_key`

network/src/noise/handshake.rs
```diff
@@ -479,50 +495,29 @@ mod test {
...
-        client: NoiseUpgrader,
-        server: NoiseUpgrader,
+        client: &NoiseUpgrader,
+        server: &NoiseUpgrader,
         server_public_key: x25519::PublicKey,
     ) -> (
-        NoiseStream<MemorySocket>,
-        (NoiseStream<MemorySocket>, PeerId),
+        io::Result<NoiseStream<MemorySocket>>,
+        io::Result<(NoiseStream<MemorySocket>, PeerId)>,
```
mimoo, line 511: why return a result and not unwrap here?

mimoo: ok nevermind

mimoo:
LGTM! @bors-libra delegate+

network/src/noise/handshake.rs
phlip9: what's our naming convention? i personally prefer `key` + `pubkey`. here it's `key` + `public_key`. in other places it's `private` + `public` or `private` + `public_key`...

let's choose one and use it everywhere

mimoo: we're pretty much all over the place on this one. 
`key` is usually used for symmetric algorithms. 

I think private_key and public_key is better than private/public imo

PR #4462, opened 2020-06-12
[core modules] Update core code address

### LINKED ISSUE #4460: Change `CORE_CODE_ADDRESS` to not be the VM reserved address. Add account.
Right now the core code address is at `0x0`. However, if we wish to publish new modules we can't publish them at `0x0` since no account can exist at that address.

The new `CORE_CODE_ADDRESS` is `0x1`

tzakian:
Changes the `CORE_CODE_ADDRESS` from `0x0` to `0x1`. Adds a test to make sure that we can publish modules from `0x1`. 

The move from 0x0 to 0x1 is due to 0x0 being the VM reserved address. This meant that we could not add modules at 0x0 in the future since an account cannot be created at that address. 

Closes #4460


sblackshear:
Looks excellent! Does this change the prover tests too?

tzakian:
Ah! Good spot, I forgot to change those. I'll make a pass though and update those as well :)

Issue #4210, opened 2020-06-02
[network] Consider another persistence format other than TOML


gregnazario:
After spending a lot of time reworking the NetworkConfig, I found that TOML will fail serialization, and the only way I fixed it was moving a field in a struct from the first field to the last field.  This means that the TOML implementation is fragile, and can be problematic in the future as it is extended.

We should look into using a different format (JSON/YAML/...) as the TOML formatter has some rough edges associated with [Enums with extra info](https://stackoverflow.com/questions/57560593/why-do-i-get-an-unsupportedtype-error-when-serializing-to-toml-with-a-manually-i), as well as the [ordering issue](https://github.com/alexcrichton/toml-rs/issues/142) I had, which wasn't solved by adding annotations, or custom building serializers.

phlip9:
Closed by #4465

PR #4465, opened 2020-06-12
[config] Improve config usability

davidiw:
* toml to yaml
  * Rust Toml was difficult to use for nested structures (tags, placing them at the end of the config, and no Variant(T))
  * Toml is super confusing for nested data structures and lists
* remove deprecated admission control config
* replaced config tests that simply validate what should be our default configs for various networks in production -- the tests could be more rigid by ensuring defaults are what we expect -- but let's save that for another time.
* cleaned up network and seed peers to inline them rather than requiring another file -- this cleaned up a lot of logic and was the only reason for a lot of test complexity

My general goal has been to get to the simplest config and then add complexity where there feels like there is friction. I do not envision appending a set of seed peers to be a particularly complex process, so it seemed unnecessarily heavy to leave it out.


config/src/config/mod.rs
```diff
@@ -392,109 +380,9 @@ impl RootPath {
...
-            // random.default.node.config.toml
-            RANDOM_COMPLETE,
-        ]
-        .iter()
-        .map(|path| {
-            NodeConfig::load(PathBuf::from(path)).unwrap_or_else(|_| panic!("Error in {}", path))
-        })
-        .collect::<Vec<_>>();
-    }
+    const PUBLIC_FULL_NODE: &str = "src/config/test_data/public_full_node.config.yaml";
```
mgorven, line 383: Can we drop the useless `.config` in the names since they're changing anyway?

config/src/config/test_data/public_full_node.config.yaml
```diff
@@ -0,0 +1,18 @@
+base:
+    data_dir: "/opt/libra/data/common"
+    role: "full_node"
+    waypoint:
+        type: "from_config"
```
mgorven, line 5: [For later]: This `type` param is ugly for the user, ideally the type would be inferred from the presence of a `waypoint` or `backend` section.

config/src/config/test_data/public_full_node.config.yaml
```diff
@@ -0,0 +1,18 @@
...
+        waypoint:
+            version: 0
+            value: "01234567890ABCDEFFEDCA098765421001234567890ABCDEFFEDCA0987654210"
+
+execution:
+    genesis_file_location: "relative/path/to/genesis"
+
+full_node_networks:
+    - discovery_method:
+        type: "onchain"
```
mgorven, line 15: Can't this just be `discovery_method: "onchain"`?

config/src/config/test_data/validator.config.yaml
```diff
@@ -0,0 +1,57 @@
+base:
+    data_dir: "/opt/libra/data/common"
+    role: "validator"
+    waypoint:
+        type: "from_storage"
```
mgorven, line 5: [For later] Same `type` comment as waypoint.

config/src/config/test_data/validator.config.yaml
```diff
@@ -0,0 +1,57 @@
...
+        backend:
+            type: "vault"
+            server: "https://127.0.0.1:8200"
+            ca_certificate: "/full/path/to/certificate"
+            token:
+                type: "from_disk"
```
mgorven, line 11: [For later] Ideally the user would either just set `token_file: /foo` or `token: s.foo`.

config/src/config/test_data/validator_full_node.config.yaml
```diff
@@ -0,0 +1,38 @@
...
+                  path: "/full/path/to/token"
+      network_id:
+          type: "public"
+    - listen_address: "/ip4/0.0.0.0/tcp/7180"
+      network_id:
+          type: "private"
+          name: "vfn"
+      seed_peers:
+          "c227da54069989f283712e4016704660":
+              - "/ip4/127.0.0.1/tcp/58259/ln-noise-ik/c998dcd54c3daf48e0ad516d94b7be0b0b7a27caa00541f2b2c14b13500df10b/ln-handshake/0"
```
mgorven, line 38: Is this to support multiple addresses per peer?

config/config-builder/src/full_node_config.rs
```diff
@@ -182,7 +177,7 @@ impl FullNodeConfig {
             network.discovery_method = DiscoveryMethod::gossip(self.advertised_address.clone());
             network.mutual_authentication = self.mutual_authentication;
 
-            network_peers.peers.insert(
+            network_peers.insert(
```
gregnazario, line 180: Awesome, that always confused me

config/src/config/network_config.rs
```diff
@@ -41,13 +41,9 @@ pub struct NetworkConfig {
     pub mutual_authentication: bool,
     // Leveraged by mutual_authentication for incoming peers that may not have a well-defined
     // network address.
-    #[serde(skip)]
     pub network_peers: NetworkPeersConfig,
-    pub network_peers_file: PathBuf,
     // Initial set of peers to connect to
-    #[serde(skip)]
     pub seed_peers: SeedPeersConfig,
-    pub seed_peers_file: PathBuf,
```
gregnazario, line 50: Was there a reason the peers / seed_peers were being saved to a different file?

config/src/config/network_config.rs
```diff
@@ -220,19 +166,6 @@ impl SeedPeersConfig {
...
-    #[serde(serialize_with = "utils::serialize_ordered_map")]
-    pub peers: HashMap<PeerId, x25519::PublicKey>,
-}
-
-impl std::fmt::Debug for NetworkPeersConfig {
-    fn fmt(&self, f: &mut std::fmt::Formatter) -> std::fmt::Result {
-        write!(f, "<{} keys>", self.peers.len())
-    }
-}
-
```
gregnazario, line 235: Less confusing sub configs!

config/src/config/secure_backend_config.rs
```diff
@@ -157,18 +157,17 @@ mod tests {
...
-
-[vault.token]
-type = \"from_config\"
-token = \"test\"
+vault:
+    server: \"127.0.0.1:8200\"
+    token:
+        type: \"from_config\"
+        token: \"test\"
         ";
```
gregnazario, line 165: Might be a little more friendly to use the raw string literal: https://doc.rust-lang.org/reference/tokens.html#raw-string-literals

Alternatively, I believe you might be able to `'` single quote them.

config/src/config/network_config.rs
davidiw: When I first started working on configs last year, almost anything that could be in a different file was. It wasn't motivated, it was just done. My take is that we should build the simplest config and improve it later if we need to. But you can see in this PR, removing this aspect removes a lot of code and complexity.

config/src/config/secure_backend_config.rs
davidiw: Oh, will try that out!

config/src/config/test_data/public_full_node.config.yaml
davidiw: I'll see what I can do with these, I suspect there may be some limitations due to the use of serde / dynamic typing.

config/src/config/test_data/validator.config.yaml
davidiw: I think this might be easier in Yaml

config/src/config/test_data/validator.config.yaml
davidiw: I think this might be possible in yaml

config/src/config/test_data/validator_full_node.config.yaml
davidiw: I believe so. The onchain addresses support multiple addresses for when the client rotates... at which point if we just pull down all of them and store them to this seed_peers we would ideally have all of them since it would be unclear which one was the "best"

gregnazario:
resolves #4210

PR #4467, opened 2020-06-12
Support delete operation in jellyfish-merkle

yihuang:
## Motivation

To make the jellyfish merkle library more feature complete, can potentially used for other purposes, like UTxO set.

Yes

## Test Plan

Unit tests are not done yet, if you like this, I'll add the tests.

### DIFF
--- a/storage/jellyfish-merkle/src/lib.rs
+++ b/storage/jellyfish-merkle/src/lib.rs
@@ -236,6 +236,30 @@ where
             let version = first_version + idx as u64;
             blob_set
                 .into_iter()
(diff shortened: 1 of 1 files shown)


davidiw:
Is this a feature we want in our storage?

lightmark:
I'll take a look this week.
@davidiw Currently not a high-pri thing since we don't have a use case in our case, maybe for general purpose it is more useful as it is a good-to-have feature. But I think even we approve this from contributors, it has to be merged after code freeze to avoid any risks.

davidiw:
We are currently not accepting new features into the code base at this point in time. Given @lightmark's feedback, we'll re-open this and explore again in the future. Thanks for your contribution!

PR #4475, opened 2020-06-12
[cluster_test] Convert all code in main.rs to async code

andll:
main.rs was a strange mix of sync and async code.
At this point it actually becoming hard to read and reason about it, plus some functions like run_single_experiment became too complex because of this mix

This diff cleans this up - main function is now async tokio::main and everything else is async


andll:
@bors-libra r=snakech 

@snakech Your pr should introduce new health check instance, why would there be conflicts?

snakech:
> @bors-libra r=snakech
> 
> @snakech Your pr should introduce new health check instance, why would there be conflicts?

I mean the current draft PR has some helper functions here, but they will be removed in formal PR

PR #4479, opened 2020-06-12
[libra-channel] fix race causing channel close to deadlock

phlip9:
We've been experiencing intermittent flaky tests in network related to
simultaneous dial tests. In the simultaneous dial logic, we sometimes
signal an existing `Peer` instance that it should close by dropping its
`libra_channel::Sender<ProtocolId, NetworkRequest>` in the `active_peers`
map. After _much_ investigation, it turns out that dropping the peer
`Sender` handle would sometimes fail to wake up the `Receiver`, who
was responsible for tearing down the `Peer` instance upon observing the
channel shutdown. As a result, the tests would sometimes deadlock,
leading to the observed flakiness in our test suite.

The fundamental issue in `libra-channel` was using the shared state
`Arc`'s `strong_count` as a proxy for the number of live `Sender`s and
the resulting race condition between dropping a `Sender` and waking the
`Receiver`.

```rust
// 1. queue is empty, one Sender, one Receiver
// 2. Sender drop
// 3. Sender acquires lock, wakes Receiver, releases lock.
// 4. Sender gets preempted _before finishing drop_ (so strong_count is
//    not updated yet)
// 5. Receiver wakes up, observes no items in the queue, observes
//    strong_count > 1.
// 6. Receiver registers its waker, and goes back to sleep.
// 7. Sender finally gets run and finishes dropping.
// 8. Deadlock! All Sender's are dropped and Receiver will never receive
//    an external stimulus to wake again.

impl Drop for Sender {
    fn drop(&mut self) {
        // 2.
        let mut shared_state = self.shared_state.lock().unwrap();
        if let Some(waker) = shared_state.waker.take() {
            waker.wake();
        }
        // 3. compiler: shared_state unlock
        // 4. preempted
        // compiler: Arc::drop
        // 7.
    }
}

// ..

impl Stream for Receiver {
    fn poll_next(/* .. */) -> Poll<Option<Self::Item>> {
        // 5.
        let mut shared_state = self.shared_state.lock().unwrap();
        if /* popped item from queue */ {
            // ..
        } else if Arc::strong_count(&self.shared_state) == 1 {
            shared_state.stream_terminated = true;
            Poll::Ready(None)
        } else {
            // 6.
            shared_state.waker = Some(cx.waker().clone());
            Poll::Pending
        }
    }
}
```

To fix this issue, we add an explicit `num_senders` field to keep track
of the number of active `Sender` handles. When `num_senders` reaches 0
during `Sender::drop` we can safely wake the `Receiver`.

Note: unfortunately, using `num_senders` means we must acquire the lock
to increment `num_senders` on `Sender::clone`, which is intended to be a
very cheap operation... If this appears to be a bottleneck, we can try
optimizing it (ideally with the help of a concurrency model checker like
[tokio-rs/loom](https://github.com/tokio-rs/loom)).

### DIFF
--- a/common/channel/src/libra_channel.rs
+++ b/common/channel/src/libra_channel.rs
@@ -24,22 +24,25 @@ use std::{
 };
 
 /// SharedState is a data structure private to this module which is
-/// shared by the sender and receiver.
+/// shared by the `Receiver` and any `Sender`s.
 #[derive(Debug)]
 struct SharedState<K: Eq + Hash + Clone, M> {
-    /// The internal queue of messages in this Channel
+    /// The internal queue of messages in this channel.
     internal_queue: PerKeyQueue<K, (M, Option<oneshot::Sender<ElementStatus<M>>>)>,
-    /// Waker is needed so that the Sender can notify the task executor/scheduler
-    /// that something has been pushed to the internal_queue and it ready for
-    /// consumption by the Receiver and then the executor/scheduler will wake up
-    /// the Receiver task to process the next item.
+    /// The `Receiver` registers its `Waker` in this slot when the queue is empty.
+    /// `Sender`s will try to wake the `Receiver` (if any) when they push a new
+    /// item onto the queue. The last live `Sender` will also wake the `Receiver`
+    /// as it's tearing down so the `Receiver` can gracefully drain and shutdown
+    /// the channel.
     waker: Option<Waker>,
-
-    /// A boolean which tracks whether the receiver has dropped
+    /// The number of active senders. When this value reaches 0, all senders have
+    /// been dropped.
+    num_senders: usize,
+    /// A boolean which tracks whether the receiver has dropped.
     receiver_dropped: bool,
-    /// A boolean which tracks whether the stream has terminated
-    /// A stream is considered terminated when sender has dropped
-    /// and we have drained everything inside our internal queue
+    /// A boolean which tracks whether the stream has terminated. A stream is
+    /// considered terminated when sender has dropped and we have drained everything
+    /// inside our internal queue.
     stream_terminated: bool,
 }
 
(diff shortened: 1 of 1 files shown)


common/channel/src/libra_channel.rs
```diff
@@ -157,9 +173,8 @@ impl<K: Eq + Hash + Clone, M> Stream for Receiver<K, M> {
                 let _err = status_ch.send(ElementStatus::Dequeued);
             }
             Poll::Ready(Some(val))
-        // if the only Arc reference to `shared_state` is 1, which is this receiver,
-        // this must mean all senders have been dropped (and so the stream is terminated)
-        } else if Arc::strong_count(&self.shared_state) == 1 {
+        // all senders have been dropped (and so the stream is terminated)
+        } else if shared_state.num_senders == 0 {
```
gregnazario, line 177: I see, so now we're using the count rather than the count of references.

I'm new to the asynchronus model of Rust, but should we use an atomic type?

Also, I would just for safety, even though I know you have a ton off asserts around it, to say `shared_state.num_senders <= 0`

common/channel/src/libra_channel.rs
```diff
@@ -109,17 +114,27 @@ impl<K: Eq + Hash + Clone, M> Sender<K, M> {
...
 impl<K: Eq + Hash + Clone, M> Drop for Sender<K, M> {
     fn drop(&mut self) {
         let mut shared_state = self.shared_state.lock().unwrap();
-        if let Some(w) = shared_state.waker.take() {
-            w.wake();
+
+        debug_assert!(shared_state.num_senders > 0);
+        shared_state.num_senders -= 1;
+
+        if shared_state.num_senders == 0 {
```
gregnazario, line 134: Same, I know you have the debug_assert, but I would put `<= 0`

common/channel/src/libra_channel.rs
phlip9: 1. `num_senders` is inside an `Arc<Mutex<SharedState>>`, so we can only read/write it if we hold the lock. Rust (mostly) inherits the C++ atomic memory model. We could probably get away with `num_senders: AtomicUsize` outside the mutex (so the shared part would be `Arc<(AtomicUsize, Mutex<SharedState>)>`), then use `Ordering::AcqRel` to increment/decrement the counter before acquiring the lock in `drop`, and `Ordering::Acquire` in the `Receiver` to observe `num_senders`. However, I would only be comfortable adding this if we use a concurrency model checker like [tokio-rs/loom](https://github.com/tokio-rs/loom). Unfortunately, setting up loom in libra is somewhat painful and I don't really have the cycles to spare on it : (

2. `num_senders: usize` so it's already implied that `num_senders < 0` is impossible : )

gregnazario: Oh, duh `usize`

PR #4483, opened 2020-06-12
[libradb] fix get_ledger_info and rename to get_nearest_epoch_change_…

lightmark:
…ledger_info

## Motivation
The current implementation name is confusing.
```
JsonRpcError { code: -32000, message: "Server error: DB corruption: the last ledger info of epoch 1 is missing next epoch info", data: None }
```
libradb::get_ledger_info(v) has a flaw: when v is in the latest epoch it raises the above error

Y

## Test Plan

add new test plan

### DIFF
--- a/state-synchronizer/src/executor_proxy.rs
+++ b/state-synchronizer/src/executor_proxy.rs
@@ -175,7 +175,7 @@ impl ExecutorProxyTrait for ExecutorProxy {
     }
 
     fn get_ledger_info(&self, version: u64) -> Result<LedgerInfoWithSignatures> {
-        let waypoint_li = self.storage.get_ledger_info(version)?;
+        let waypoint_li = self.storage.get_nearest_epoch_change_ledger_info(version)?;
         ensure!(
             waypoint_li.ledger_info().version() == version,
             "Version of Waypoint LI {} is different from requested waypoint version {}",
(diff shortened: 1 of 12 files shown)


state-synchronizer/src/executor_proxy.rs
```diff
@@ -175,7 +175,7 @@ impl ExecutorProxyTrait for ExecutorProxy {
     }
 
     fn get_ledger_info(&self, version: u64) -> Result<LedgerInfoWithSignatures> {
-        let waypoint_li = self.storage.get_ledger_info(version)?;
+        let waypoint_li = self.storage.get_nearest_epoch_change_ledger_info(version)?;
         ensure!(
             waypoint_li.ledger_info().version() == version,
```
msmouse, line 180: Looking at the `ensure!()` here, the `version` should really be named `waypoint_version` or `epoch_change_version` or `epoch_boundary_version`, and it's expected to be a epoch_change ledger version at that version. I don't think we should even introduce `get_nearest_epoch_change_ledger_info()`, rather, how about let's just do a straightforward `.get_ledger_info_get_version() -> Result<Option<LI>>`, and ensure it returns a "Some" here.

lightmark: here it is. But it is also called in json-rpc method, which is another case.

msmouse: The other call site is a misuse, should not consider that. Can we get around it?

lightmark:
Overlap with #4511

PR #4484, opened 2020-06-12
[network] added structured logging for event loops

gregnazario:
As a first attempt at using the StructuredLogging API, I'm adding next
to the already existing trace! logs, a structured log that keeps track
of what's going on in loops in onchain-discovery and
connectivity_manager as a starting place.

### My thoughts about structured logging
1. They help the parsing a lot, and they work great in my testing.  I think its a great addition, and I'm sure it'll help a lot in Kibana!
2. We should have an environment variable that can let us turn them on for local log testing, rather than making a static function call.
3. I think we need guidelines on what level of logging we want in structured logging, and if we want different levels of it such that we don't overfill the logs / can filter out the important things, more than just the `critical` and `warning` levels, but I could be persuaded otherwise.
4. Let's make a centralized place for common terms, and maybe some examples of how to use the paradigm.  I think what I've created here is decent as a first step, but I'm willing if this isn't the best way to do it another way.
5. Do we want common fields hard typed into the structured logging?  For example, `PeerId` should always be a `PeerId` rather than possibly a string, which could cause accidental mistakes by putting the wrong value in.
6. Serialization is nice and simple, but it does cause us to think strongly about what we do and do not want to serialize at any one point in time for logging purposes, at the same time we think about serialization for into a file or across a wire.

### Testing

I added `init_println_struct_log()` to the front of some tests to check it out.  Here are some of the logs from my changes at least for `ConnectivityManager`
```
{"name":"connectivity_manager_loop","module":"network::connectivity_manager","location":"network/src/connectivity_manager/mod.rs:220","timestamp":"2020-06-12 22:09:08","data":{"type":"start","network_context":{"network_id":{"type":"Validator"},"peer_id":"42daa04d56535e51816f1074f40f3f49","role":"validator"}}}
{"name":"connectivity_manager_loop","module":"network::connectivity_manager","location":"network/src/connectivity_manager/mod.rs:220","timestamp":"2020-06-12 22:09:08","data":{"network_context":{"network_id":{"type":"Validator"},"peer_id":"262840ce151585da276e25a315a68529","role":"validator"},"type":"start"}}
{"name":"connectivity_manager_loop","module":"network::connectivity_manager","location":"network/src/connectivity_manager/mod.rs:220","timestamp":"2020-06-12 22:09:08","data":{"network_context":{"network_id":{"type":"Validator"},"peer_id":"e3d468c1f89164029d38845fe7672635","role":"validator"},"type":"start"}}
{"name":"connectivity_manager_loop","module":"network::connectivity_manager","location":"network/src/connectivity_manager/mod.rs:220","timestamp":"2020-06-12 22:09:08","data":{"network_context":{"network_id":{"type":"Validator"},"peer_id":"b65e208fb8c1a31eb2f9261277e3e929","role":"validator"},"type":"start"}}
{"name":"connectivity_manager_loop","module":"network::connectivity_manager","location":"network/src/connectivity_manager/mod.rs:220","timestamp":"2020-06-12 22:09:08","data":{"type":"start","network_context":{"network_id":{"type":"Validator"},"peer_id":"8c45f1749333432fb19304e34b2a86f0","role":"validator"}}}
{"name":"connectivity_manager_loop","module":"network::connectivity_manager","location":"network/src/connectivity_manager/mod.rs:220","timestamp":"2020-06-12 22:09:08","data":{"type":"start","network_context":{"network_id":{"type":"Validator"},"peer_id":"00deb9cd9804c546953f5b13b16106c4","role":"validator"}}}
{"name":"connectivity_manager_loop","module":"network::connectivity_manager","location":"network/src/connectivity_manager/mod.rs:220","timestamp":"2020-06-12 22:09:08","data":{"network_context":{"network_id":{"type":"Validator"},"peer_id":"07850f8e166d97bebdb738cd1e0fdee7","role":"validator"},"type":"start"}}
{"name":"connectivity_manager_loop","module":"network::connectivity_manager","location":"network/src/connectivity_manager/mod.rs:228","timestamp":"2020-06-12 22:09:08","data":{"request":{"UpdateAddresses":["Gossip",{"fe2c04e2c71b68674172ad7261812671":["/ip4/127.0.0.1/tcp/9090"]}]},"event_id":1,"type":"connectivity_request","network_context":{"network_id":{"type":"Validator"},"peer_id":"262840ce151585da276e25a315a68529","role":"validator"}}}
```

### DIFF
--- a/network/src/connectivity_manager/mod.rs
+++ b/network/src/connectivity_manager/mod.rs
@@ -27,7 +27,10 @@
 //! any partitions asap, as we aren't currently gossiping consensus messages or
 //! using a relay protocol.
 
-use crate::peer_manager::{self, conn_notifs_channel, ConnectionRequestSender, PeerManagerError};
+use crate::{
+    logging,
+    peer_manager::{self, conn_notifs_channel, ConnectionRequestSender, PeerManagerError},
+};
 use futures::{
     channel::oneshot,
     future::{BoxFuture, FutureExt},
@@ -35,10 +38,11 @@ use futures::{
 };
 use libra_config::network_id::NetworkContext;
 use libra_crypto::x25519;
-use libra_logger::prelude::*;
+use libra_logger::{prelude::*, StructuredLogEntry};
 use libra_network_address::NetworkAddress;
 use libra_types::PeerId;
 use num_variants::NumVariants;
+use serde::Serialize;
 use std::{
     cmp::min,
     collections::HashMap,
@@ -86,7 +90,7 @@ pub struct ConnectivityManager<TTicker, TBackoff> {
 /// Different sources for peer addresses, ordered by priority (Onchain=highest,
 /// Config=lowest).
 #[repr(u8)]
-#[derive(Copy, Clone, Debug, Eq, PartialEq, Ord, PartialOrd, NumVariants)]
+#[derive(Copy, Clone, Debug, Eq, PartialEq, Ord, PartialOrd, NumVariants, Serialize)]
 pub enum DiscoverySource {
     OnChain,
     Gossip,
@@ -94,13 +98,14 @@ pub enum DiscoverySource {
 }
 
 /// Requests received by the [`ConnectivityManager`] manager actor from upstream modules.
-#[derive(Debug)]
+#[derive(Debug, Serialize)]
 pub enum ConnectivityRequest {
     /// Request to update known addresses of peer with id `PeerId` to given list.
     UpdateAddresses(DiscoverySource, HashMap<PeerId, Vec<NetworkAddress>>),
     /// Update set of nodes eligible to join the network.
     UpdateEligibleNodes(HashMap<PeerId, x25519::PublicKey>),
     /// Gets current size of dial queue. This is useful in tests.
+    #[serde(skip)]
     GetDialQueueSize(oneshot::Sender<usize>),
 }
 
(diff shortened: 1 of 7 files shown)


andll:
This looks pretty cool!
I do like this pattern of declaring strings for logging in a separate logging

andll:
For (5) - I think we can offer this as an optional feature via separate method(like `.field(..)` instead of `.data(..)`) on StructuredLogEntry that will bound type of value to type of the field.

Do you want to look into that or want me to do it?

gregnazario:
> For (5) - I think we can offer this as an optional feature via separate method(like `.field(..)` instead of `.data(..)`) on StructuredLogEntry that will bound type of value to type of the field.
> 
> Do you want to look into that or want me to do it?

I don't think it's high priority to do it right now, but I think it would be helpful to prevent possible mistakes.

andll:
This is actually fairly easy to do, here you go - https://github.com/libra/libra/pull/4488

andll:
Do you want to wait until `.field` lands and incorporate it?

gregnazario:
> Do you want to wait until `.field` lands and incorporate it?

Yeah, that sounds like a good idea.  I'll also remove the `trace!` statements

network/src/connectivity_manager/mod.rs
```diff
@@ -211,29 +216,61 @@ where
...
+                        .data(logging::NETWORK_CONTEXT, &self.network_context)
+                        .data(logging::TYPE, "connectivity_request")
+                        .data(logging::EVENT_ID, self.event_id)
+                        .data(logging::REQUEST, &req)
+                    );
                     trace!("{} Event Id: {}, type: ConnectivityRequest, req: {:?}", self.network_context, self.event_id, req);
                     self.handle_request(req);
                 },
                 notif = self.connection_notifs_rx.select_next_some() => {
+                    send_struct_log!(StructuredLogEntry::new_named(logging::CONNECTIVITY_MANAGER_LOOP)
```
davidiw, line 248: @andll, I feel we could do better here... I know why we ended up with this type of macro, but I think something like:
```
struct_log!(logging::CONNECTIVITY_MANAGER_LOOP)
    .data(...)
    .log();
```
would be a little nicer.

andll: What is `CONNECTIVITY_MANAGER_LOOP` here? Your example is pretty much how it is right now, no?

gregnazario: It's a little different, I think what he's going for is removing the constructor part.
```
 send_struct_log!(StructuredLogEntry::new_named(logging::CONNECTIVITY_MANAGER_LOOP)
  .data(...)
```
to using the macro to create it, similar to the `debug!` or `info!` macros

```
 struct_log!(logging::CONNECTIVITY_MANAGER_LOOP)
  .data(...)
```

davidiw: Yeah, we have a wrapper around the object so that we can add additional fields. My suggestion is to just make the macro contain both the constructor and additional annotation.

andll: Oh. removing constructor will also remove type hints support from IDE for those who use it(like me)

gregnazario: FYI, my IDE really does not like the macro, and the type hinting / autoformatting doesn't work within the macro.  I end up making it outside, and then sticking it in the macro.

andll: So there are two separate proposals
- wrapper object with specific fields. I am not sure it needs any support in logger crate, I just imaging is as declaring some `struct NetworkLogEntry(StructuredLogEntry)` with appropriate methods?
- Changing macro so that it's shorter, e.g. `struct_log!(logging::CONNECTIVITY_MANAGER_LOOP` instead of `struct_log!(logging::CONNECTIVITY_MANAGER_LOOP::new(...)`. I think this will save maybe 3-4 characters on `::new(`, but will also remove auto-complete support from IDE. is it worth it?

gregnazario: I think we should move this discussion to an issue: https://github.com/libra/libra/issues/4516

PR #4488, opened 2020-06-12
[logger] Introduce .field on structured logger

andll:
This will allow to declare specific type of the logging field.

Example of usage:
```
mod logging {
   pub const MY_FIELD:LoggingField<u64> = LoggingField::new("my_field");
}

mod my_code {
   fn my_fn() {
       send_struct_log!(StructuredLogEntry::new(...).field(&logging::MY_FIELD, 0))
   }
}
```


andll:
Fair enough, added this to comment since commit message will eventually be buried under history

@bors-libra r=gregnazario

PR #4491, opened 2020-06-13
[cluster test] Full node health check

snakech:
## Motivation

We call wait_until_all_healthy at the end of experiment, to make sure all nodes recover from whatever happened in experiment.

Added new full node health check to wait for full node to catch up for a validators in this function.

After this new health check is introduced, try re-running --perf-suite and confirm that there is no expired transactions in 3 region simulation

(Write your answer here.)

## Test Plan

## Related PRs

### DIFF
--- a/testsuite/cluster-test/src/experiments/performance_benchmark_three_region_simulation.rs
+++ b/testsuite/cluster-test/src/experiments/performance_benchmark_three_region_simulation.rs
@@ -97,7 +97,7 @@ impl Experiment for PerformanceBenchmarkThreeRegionSimulation {
     }
 
     fn deadline(&self) -> Duration {
-        Duration::from_secs(420)
+        Duration::from_secs(600)
     }
 }
 
--- a/testsuite/cluster-test/src/health/fullnode_check.rs
+++ b/testsuite/cluster-test/src/health/fullnode_check.rs
@@ -1,0 +1,120 @@
+// Copyright (c) The Libra Core Contributors
+// SPDX-License-Identifier: Apache-2.0
+
+#![forbid(unsafe_code)]
+
+use crate::{
+    cluster::Cluster,
+    health::{HealthCheck, HealthCheckContext},
+    prometheus::Prometheus,
+    util::unix_timestamp_now,
+};
+use anyhow::{format_err, Result};
+use once_cell::sync::Lazy;
+use regex::Regex;
+use std::{collections::HashMap, time::Duration};
+
+static VAL_REGEX: Lazy<Regex> = Lazy::new(|| Regex::new(r"val-(\d+)").unwrap());
+static FULLNODE_REGEX: Lazy<Regex> = Lazy::new(|| Regex::new(r"fn-(\d+)").unwrap());
+
+pub struct FullNodeHealthCheck {
+    cluster: Cluster,
+    prometheus: Prometheus,
+}
+
+impl FullNodeHealthCheck {
+    pub fn new(cluster: Cluster, prometheus: Prometheus) -> FullNodeHealthCheck {
+        Self {
+            cluster,
+            prometheus,
+        }
+    }
+
+    pub fn state_sync_committed_version(
+        &self,
+        start: Duration,
+        end: Duration,
+    ) -> Result<HashMap<String, f64>> {
+        libra_retrier::retry(libra_retrier::fixed_retry_strategy(1_000, 5), || {
+            self.prometheus
+                .query_range_max(
+                    "libra_state_sync_committed_version{{}}".to_string(),
+                    &start,
+                    &end,
+                    10, /* step */
+                )
+                .map_err(|e| format_err!("No sync committed version data: {}", e))
+        })
+    }
+}
+
+pub fn convert_index(input: HashMap<String, f64>) -> HashMap<String, Vec<i64>> {
+    let mut res = HashMap::new();
+    for (peer_name, version) in input {
+        if let Some(cap) = VAL_REGEX.captures(peer_name.as_ref()) {
+            // this is index of validator
+            if let Some(cap) = cap.get(1) {
+                let index = cap.as_str();
+                // we record version of validator to be a negative number
+                // to calculate diff with all full nodes are associated
+                res.entry(index.to_string())
+                    .or_insert_with(Vec::new)
+                    .push(-version as i64);
+            }
+        } else if let Some(cap) = FULLNODE_REGEX.captures(peer_name.as_ref()) {
+            // this is index of fullnode
+            if let Some(cap) = cap.get(1) {
+                let index = cap.as_str();
+                res.entry(index.to_string())
+                    .or_insert_with(Vec::new)
+                    .push(version as i64);
+            }
+        } else {
+            panic!(
+                "Failed to parse peer name {} into validator_index",
+                peer_name
+            )
+        }
+    }
+    res
+}
+
+impl HealthCheck for FullNodeHealthCheck {
+    fn verify(&mut self, ctx: &mut HealthCheckContext) {
+        let validator = self.cluster.validator_instances();
+        let buffer = Duration::from_secs(30);
+        let timestamp = unix_timestamp_now() - buffer;
+        let commit_version =
+            match FullNodeHealthCheck::state_sync_committed_version(&self, timestamp, timestamp) {
+                Err(_e) => return,
+                Ok(s) => convert_index(s),
+            };
+        for (index, version) in &commit_version {
+            if version.len() < 2 {
+                panic!("Missed record to compare fullnode and validator version difference, record length: {}",
+                       version.len());
+            }
+            let mut version = version.clone();
+            version.sort();
+            // because validator version is the only negative number, it always be sorted at the first slot
+            for i in 1..version.len() {
+                // we define tolerance of version diff is 20000
+                if (version[i] + version[0]) > 20000_i64 || (version[i] + version[0]) < -20000_i64 {
+                    ctx.report_failure(
+                        validator[index.parse::<usize>().unwrap()]
+                            .peer_name()
+                            .clone(),
+                        format!(
+                            "state sync committed version of fullnode: {} is behind validator: {}",
+                            version[i], -version[0],
+                        ),
+                    );
+                }
+            }
+        }
+    }
+
+    fn name(&self) -> &'static str {
+        "fullnode_check"
+    }
+}
--- a/testsuite/cluster-test/src/health/mod.rs
+++ b/testsuite/cluster-test/src/health/mod.rs
@@ -5,13 +5,15 @@
 
 mod commit_check;
 mod debug_interface_log_tail;
+mod fullnode_check;
 mod liveness_check;
 mod log_tail;
 
-use crate::{cluster::Cluster, util::unix_timestamp_now};
+use crate::{cluster::Cluster, prometheus::Prometheus, util::unix_timestamp_now};
 use anyhow::{bail, Result};
 pub use commit_check::CommitHistoryHealthCheck;
 pub use debug_interface_log_tail::DebugPortLogWorker;
+pub use fullnode_check::FullNodeHealthCheck;
 use itertools::Itertools;
 pub use liveness_check::LivenessHealthCheck;
 pub use log_tail::{LogTail, TraceTail};
@@ -59,7 +61,7 @@ impl fmt::Debug for ValidatorEvent {
 
 pub trait HealthCheck {
     /// Verify specific event
-    fn on_event(&mut self, event: &ValidatorEvent, ctx: &mut HealthCheckContext);
+    fn on_event(&mut self, _event: &ValidatorEvent, _ctx: &mut HealthCheckContext) {}
     /// Periodic verification (happens even if when no events produced)
     fn verify(&mut self, _ctx: &mut HealthCheckContext) {}
     /// Optionally marks validator as failed, requiring waiting for at least one event from it to
@@ -87,15 +89,18 @@ impl HealthCheckRunner {
         }
     }
 
-    pub fn new_all(cluster: Cluster) -> Self {
+    pub fn new_all(cluster: Cluster, prometheus: Option<Prometheus>) -> Self {
+        let mut health_check: Vec<Box<dyn HealthCheck>> = vec![];
         let liveness_health_check = LivenessHealthCheck::new(&cluster);
-        Self::new(
-            cluster,
-            vec![
-                Box::new(CommitHistoryHealthCheck::new()),
-                Box::new(liveness_health_check),
-            ],
-        )
+        health_check.push(Box::new(CommitHistoryHealthCheck::new()));
+        health_check.push(Box::new(liveness_health_check));
+        if let Some(p) = prometheus.as_ref() {
+            health_check.push(Box::new(FullNodeHealthCheck::new(
+                cluster.clone(),
+                p.clone(),
+            )))
+        };
+        Self::new(cluster, health_check)
     }
 
     /// Takes a list of affected_validators. If there are validators which failed
--- a/testsuite/cluster-test/src/prometheus.rs
+++ b/testsuite/cluster-test/src/prometheus.rs
@@ -4,6 +4,7 @@
 #![forbid(unsafe_code)]
 
 use anyhow::{bail, format_err, Result};
+use libra_logger::debug;
 use reqwest::Url;
 use serde::Deserialize;
 use std::{collections::HashMap, time::Duration};
@@ -65,7 +66,7 @@ impl Prometheus {
                 step
             ))
             .expect("Failed to make query_range url");
-
+        debug!("Query url is: {:?}", url);
         let response = self
             .client
             .get(url.clone())
@@ -88,6 +89,7 @@ impl Prometheus {
             ),
         }
     }
+
     pub fn query_range_avg(
         &self,
         query: String,
@@ -100,9 +102,42 @@ impl Prometheus {
             .avg()
             .ok_or_else(|| format_err!("Failed to compute avg"))
     }
+
+    pub fn query_range_max(
+        &self,
+        query: String,
+        start: &Duration,
+        end: &Duration,
+        step: u64,
+    ) -> Result<HashMap<String, f64>> {
+        let response = self.query_range(query, start, end, step)?;
+        response
+            .max()
+            .ok_or_else(|| format_err!("Failed to compute max"))
+    }
 }
 
 impl MatrixResponse {
+    pub fn max(&self) -> Option<HashMap<String, f64>> {
+        let mut max = HashMap::new();
+        for (peer_id, time_series) in &self.inner {
+            let cur_max = time_series
+                .inner
+                .iter()
+                .map(|tuple| tuple.1)
+                // Some time series can return NaN, but f64::max always
+                // chooses the non-NaN argument
+                .fold(f64::NAN, f64::max);
+
+            max.insert(peer_id.clone(), cur_max);
+        }
+        if max.is_empty() {
+            None
+        } else {
+            Some(max)
+        }
+    }
+
     pub fn avg(&self) -> Option<f64> {
         if self.inner.is_empty() {
             return None;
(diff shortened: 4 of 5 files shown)


testsuite/cluster-test/src/experiments/performance_benchmark_three_region_simulation.rs
```diff
@@ -97,7 +97,7 @@ impl Experiment for PerformanceBenchmarkThreeRegionSimulation {
     }
 
     fn deadline(&self) -> Duration {
-        Duration::from_secs(420)
+        Duration::from_secs(600)
```
andll, line 100: Have you tried this experiment with latest master, without increasing deadline?
We had a problem with network discovery that made startup very long, but it is resolved now. Do we still need to bump deadline to 10 minutes(that a lot...)

testsuite/cluster-test/src/health/mod.rs
```diff
@@ -108,6 +113,7 @@ impl HealthCheckRunner {
         events: &[ValidatorEvent],
         affected_validators_set: &HashSet<String>,
         print_failures: PrintFailures,
+        full_node_check: bool,
```
andll, line 116: Why do we have `full_node_check` param in run method?
I understand some setups don't support full node health check, should we just initialize HealthCheckRunner without full node health check in this case?

testsuite/cluster-test/src/health/fullnode_check.rs
```diff
@@ -0,0 +1,144 @@
...
+use std::time::Duration;
+
+const APP_LIBRA_VALIDATOR: &str = "libra-validator";
+const APP_LIBRA_FULLNODE: &str = "libra-fullnode";
+static VAL_REGEX: Lazy<Regex> = Lazy::new(|| Regex::new(r"val-(\d+)").unwrap());
+static FULLNODE_REGEX: Lazy<Regex> = Lazy::new(|| Regex::new(r"fn-(\d+)").unwrap());
+
+pub struct FullNodeHealthCheck {
+    cluster: Cluster,
+    prometheus: Option<Prometheus>,
```
andll, line 25: FullNodeHealthCheck does not work without prometheus at all right? In this case I think we want this to be non-optional, and just skip initializing FullNodeHealthCheck if there is no prometheus

testsuite/cluster-test/src/health/fullnode_check.rs
```diff
@@ -0,0 +1,144 @@
...
+impl FullNodeHealthCheck {
+    pub fn new(cluster: Cluster, prometheus: Option<Prometheus>) -> FullNodeHealthCheck {
+        Self {
+            cluster,
+            prometheus,
+        }
+    }
+
+    pub fn state_sync_committed_version(
+        prometheus: &Prometheus,
```
andll, line 37: This function could just take `&self` and get prometheus from there?

testsuite/cluster-test/src/health/fullnode_check.rs
```diff
@@ -0,0 +1,144 @@
...
+        let fullnode = self.cluster.fullnode_instances();
+        let validator = self.cluster.validator_instances();
+        let prometheus = match self.prometheus.as_ref() {
+            Some(p) => p,
+            None => return,
+        };
+        assert_eq!(
+            fullnode.len(),
+            validator.len(),
+            "fullnode and validator must have same size, fullnode: {}, validator: {}",
```
andll, line 96: Why is this a requirement for this check?
we can have more then one full node per validator(see `k8s_fullnodes_per_validator` param), and check should be able to handle that

testsuite/cluster-test/src/experiments/performance_benchmark_three_region_simulation.rs
snakech: I tired experiment with master but I reckon it was not the latest, I can re-do the experiment to confirm this point. Why I keep this change here is because I found we use 600s for other performance tests, so I assume 600s is a reasonable number for all performance tests. May I ask why we set 420s here which is much shorter than others at the first place?

testsuite/cluster-test/src/health/mod.rs
snakech: I wanna save one query time especially in fn ```run_single_experiment```, and in ```wait_until_all_healthy```, which is using same HealthCheckRunner, also we do call ```HealthCheckRunner``` twice in one experiment, but we also only need one time full node check at end of the experiment. I have thought about this point, instead of add a new member of ```HealthCheckRunner``` with different initializer in ```ClusterTestRunner```, maybe a flag is much easier and flexible?

testsuite/cluster-test/src/health/fullnode_check.rs
snakech: u are correct, this will be more clear instead of option

testsuite/cluster-test/src/health/fullnode_check.rs
snakech: yeah... forgot self here

testsuite/cluster-test/src/health/fullnode_check.rs
snakech: oh, I c, by my experiment I thought full node number should match the validator. Thx for pointed it out

testsuite/cluster-test/src/experiments/performance_benchmark_three_region_simulation.rs
andll: Ok, I see that other set to 10min, you are right. We can keep it here

testsuite/cluster-test/src/health/mod.rs
andll: I don't think querying prometheus is expensive and have this extra flag is a bit strange(HealthCheckRunner otherwise is not aware of health check specifics)

I think we want full node healthcheck everywhere, not just in the end of experiment, same way we have it for other health checks

I think what worries me more right now is that our prometheus seem to be flacky, but that's separate issue

snakech: yes, prometheus is sometimes flacky and returned nothing, so this is why I added protection to go through check if prometheus is flacky.

querying prometheus one time with 30 nodes and 30 validators will return around 5Mb data(by chromn tool) and we also need to process it. I think it will not become critical with current performance test framework, but what if we have more nodes and validators? If this will be a case I think saving times of querying should be there, but if it not a problem, I can remove this flag and run full node check always

andll: So yes, I think we need to run this healthcheck always.

We need to address flaky prometheus as well, obviously more often we query prometheus more often test would fail if it is flaky.

To be honest, maybe given flakiness of prometheus, we can just query node for the metric using debug interface, .e.g. `AsyncNodeDebugClient::get_node_metric`?

snakech: ok, I can make full node health check always at this stage.

I did not read debug interface code I can have a look first. Once we decide make some change of this part, we can discuss again then?

testsuite/cluster-test/src/health/fullnode_check.rs
```diff
@@ -0,0 +1,129 @@
...
+            APP_LIBRA_VALIDATOR,
+        ) {
+            Err(_e) => return,
+            Ok(s) => convert_index(s),
+        };
+        for (node_index, version) in &res_fullnode {
+            if res_validator.contains_key(node_index) {
+                let validator_version = res_validator[node_index];
+                // we define tolerance of version diff is 2000
+                if (version - validator_version).abs() > 20000 as f64 {
```
andll, line 110: Number here is 20k, is it a typo?
Seem like a huge gap?

testsuite/cluster-test/src/health/mod.rs
```diff
@@ -87,15 +90,19 @@ impl HealthCheckRunner {
...
-        Self::new(
-            cluster,
-            vec![
-                Box::new(CommitHistoryHealthCheck::new()),
-                Box::new(liveness_health_check),
-            ],
-        )
+        health_check.push(Box::new(CommitHistoryHealthCheck::new()));
+        health_check.push(Box::new(liveness_health_check));
+        match prometheus.as_ref() {
```
andll, line 98: use `if let` here if you only care about case when prometheus is set

testsuite/cluster-test/src/health/fullnode_check.rs
```diff
@@ -0,0 +1,129 @@
...
+    let mut res = HashMap::new();
+    for (peer_name, version) in input {
+        let peer_name = node_index(peer_name);
+        res.insert(peer_name, version);
+    }
+    res
+}
+
+impl HealthCheck for FullNodeHealthCheck {
+    fn on_event(&mut self, _event: &ValidatorEvent, _ctx: &mut HealthCheckContext) {}
```
andll, line 82: Can you actually change `HealthCheck` interface to have default implementation for `on_event`, so that it does not have to be defined here?

testsuite/cluster-test/src/health/fullnode_check.rs
snakech: oh, 20k is not a typo, comment is typo, it should be 20k

testsuite/cluster-test/src/health/mod.rs
snakech: yeah, lint check noticed this format error, will update

testsuite/cluster-test/src/health/fullnode_check.rs
andll: 20k is a lot, can you give a link to prometheus graph that shows that full node and validator can diverge this much in regular state?
Each version is basically a transaction, 20k transaction means full nodes is 20 seconds behind. That's a lot?

testsuite/cluster-test/src/health/fullnode_check.rs
snakech: yes, this is a good idea

testsuite/cluster-test/src/health/fullnode_check.rs
snakech: based on my collecting datas last week, I do find same case which full node and around 16k difference with validator but test can still go through under that difference

andll: So the question is not whether test can go through, it's mostly what is the minimal difference that covers normal operation.

For example, can you run `--emti-tx` for 10 minutes, and post here link to prometheus for this run? We can analyze version graph during this run and see how far full node is behind, and use this as a number for theshold

snakech: 1717 |  
-- | --
5339 | TRUE
998 |  
500 |  
251 |  
9050 | TRUE
750 |  
0 |  
2287 | TRUE
1001 |  
250 |  
2579 | TRUE
5255 | TRUE
1732 |  
1507 |  
751 |  
11629 | TRUE
1084 |  
6497 | TRUE
1787 |  
1797 |  
466 |  
3745 | TRUE
5496 | TRUE
6858 | TRUE
1546 |  
1787 |  
 there are one set of test difference of normal operation, I can run --emti-tx for more data

testsuite/cluster-test/src/health/fullnode_check.rs
```diff
@@ -0,0 +1,125 @@
...
+        } else if let Some(cap) = FULLNODE_REGEX.captures(peer_name.as_ref()) {
+            // this is index of fullnode
+            if let Some(cap) = cap.get(1) {
+                let index = cap.as_str();
+                if res.contains_key(index) {
+                    res.get_mut(index).unwrap().push(version as i64);
+                } else {
+                    let mut versions = vec![];
+                    versions.push(version as i64);
+                    res.insert(index.to_string(), versions);
```
huitseeker, line 73: Consider using the [`entry` API](http://doc.rust-lang.org/nightly/std/collections/struct.HashMap.html#method.entry) to simplify this edition, as well as the one above.

testsuite/cluster-test/src/health/fullnode_check.rs
```diff
@@ -0,0 +1,125 @@
...
+            version.sort();
+            // because validator version is the only negative number, it always be sorted at the first slot
+            for i in 1..version.len() {
+                // we define tolerance of version diff is 20000
+                if (version[i] + version[0]) > 20000_i64 || (version[i] + version[0]) < -20000_i64 {
+                    ctx.report_failure(
+                        validator[index.parse::<usize>().unwrap()]
+                            .clone()
+                            .peer_name()
+                            .clone(),
```
huitseeker, line 111: I suspect one of these clones may be superfluous.

testsuite/cluster-test/src/prometheus.rs
```diff
@@ -100,9 +102,51 @@ impl Prometheus {
...
+                }
+                if *v > cur_max {
+                    cur_max = *v;
+                }
+            }
+
+            max.insert(peer_id.clone(), cur_max);
+            count += 1;
+        }
+        if count == 0 {
```
huitseeker, line 143: You may want a simple boolean for this guard. Or you could use the size of the max table.

testsuite/cluster-test/src/health/fullnode_check.rs
snakech: aha, you are correct, just mis-added

testsuite/cluster-test/src/prometheus.rs
```diff
@@ -100,9 +102,45 @@ impl Prometheus {
...
+            let cur_max = time_series
+                .inner
+                .iter()
+                .map(|tuple| tuple.1)
+                // Some time series can return NaN, but f64::max always
+                // chooses the non-NaN argument
+                .fold(f64::NAN, f64::max);
+
+            max.insert(peer_id.clone(), cur_max);
+        }
```
huitseeker, line 136: FYI: since `HashMap` implements `FromIterator`, you can also:
```
        let max: HashMap<_, _> = self
            .inner
            .iter()
            .map(|(peer_id, time_series)| {
                (
                    peer_id.clone(),
                    time_series
                        .inner
                        .iter()
                        .map(|tuple| tuple.1)
                        // Some time series can return NaN, but f64::max always
                        // chooses the non-NaN argument
                        .fold(0. / 0., f64::max),
                )
            })
            .collect();
```
but that's just a matter of taste.

testsuite/cluster-test/src/prometheus.rs
```diff
@@ -100,9 +102,45 @@ impl Prometheus {
...
+        let response = self.query_range(query, start, end, step)?;
+        response
+            .max()
+            .ok_or_else(|| format_err!("Failed to compute max"))
+    }
 }
 
 impl MatrixResponse {
+    pub fn max(&self) -> Option<HashMap<String, f64>> {
+        if self.inner.is_empty() {
```
huitseeker, line 122: If `self.inner` is empty, what would the length of the `max` `HashMap` be?

testsuite/cluster-test/src/prometheus.rs
huitseeker, line 136: FYI: since `HashMap` implements `FromIterator`, you can also:
```
        let max: HashMap<_, _> = self
            .inner
            .iter()
            .map(|(peer_id, time_series)| {
                (
                    peer_id.clone(),
                    time_series
                        .inner
                        .iter()
                        .map(|tuple| tuple.1)
                        // Some time series can return NaN, but f64::max always
                        // chooses the non-NaN argument
                        .fold(0. / 0., f64::max),
                )
            })
            .collect();
```
but that's just a matter of taste, and what you have is perfectly suitable.

testsuite/cluster-test/src/health/fullnode_check.rs
```diff
@@ -0,0 +1,120 @@
...
+                )
+                .map_err(|e| format_err!("No sync committed version data: {}", e))
+        })
+    }
+}
+
+pub fn convert_index(input: HashMap<String, f64>) -> HashMap<String, Vec<i64>> {
+    let mut res = HashMap::new();
+    for (peer_name, version) in input {
+        if let Some(cap) = VAL_REGEX.captures(peer_name.as_ref()) {
```
huitseeker, line 54: Purely a nit / matter of taste / fyi, but you could also
```
if let Some(index) = VAL_REGEX.captures(peer_name.as_ref()).and_then(|cap| cap.get(1).map(|cap| cap.as_str())) {
...
```

testsuite/cluster-test/src/prometheus.rs
snakech: I assume if self.inner is empty, it returns None? Do we need to care max HashMap then?

huitseeker: If self.inner is empty, the iteration through its elements is a no-op, the max HashMap is empty, and you return None already.

snakech: Sorry I think I did not get the point before. yes, this check here is redundant I will remove it, we will eventually return None if inner is empty

testsuite/cluster-test/src/health/fullnode_check.rs
```diff
@@ -0,0 +1,120 @@
...
+    res
+}
+
+impl HealthCheck for FullNodeHealthCheck {
+    fn verify(&mut self, ctx: &mut HealthCheckContext) {
+        let validator = self.cluster.validator_instances();
+        let buffer = Duration::from_secs(30);
+        let timestamp = unix_timestamp_now() - buffer;
+        let commit_version =
+            match FullNodeHealthCheck::state_sync_committed_version(&self, timestamp, timestamp) {
```
andll, line 88: self.state_sync_committed_version?

testsuite/cluster-test/src/health/fullnode_check.rs
```diff
@@ -0,0 +1,120 @@
...
+                Err(_e) => return,
+                Ok(s) => convert_index(s),
+            };
+        for (index, version) in &commit_version {
```
andll, line 92: So `commit_version` is not really used after this, so you can just do 
`for (index, mut version) in commit_version`

testsuite/cluster-test/src/health/fullnode_check.rs
```diff
@@ -0,0 +1,120 @@
...
+            if version.len() < 2 {
+                panic!("Missed record to compare fullnode and validator version difference, record length: {}",
+                       version.len());
+            }
+            let mut version = version.clone();
```
andll, line 97: You can then avoid this `.clone`

testsuite/cluster-test/src/health/fullnode_check.rs
```diff
@@ -0,0 +1,120 @@
...
+                    &start,
+                    &end,
+                    10, /* step */
+                )
+                .map_err(|e| format_err!("No sync committed version data: {}", e))
+        })
+    }
+}
+
+pub fn convert_index(input: HashMap<String, f64>) -> HashMap<String, Vec<i64>> {
```
andll, line 51: Returning negative index for validator and then sorting array seem to be very fragile and error prone

I think you can avoid convert_index function all together.

When you get `Map<String, f64>` from prometheus, you can do something like this:
```
for full_node in cluster.fullnode_instances() {
let full_node_data_point = map.get(full_node.peer_name());
let validator_data_point = map.get("val-" + full_node.validator_index()); // well use format! here, but you got the idea
// compare full_node_data_point and validator_data_point
}
```

Doing this way have few benefits over your approach
a) It is simpler
b) It allows to catch problems like if there is no data point for some full node or validator (this should fail health check, as this is not normal situation)

testsuite/cluster-test/src/prometheus.rs
```diff
@@ -100,9 +102,42 @@ impl Prometheus {
...
+    ) -> Result<HashMap<String, f64>> {
+        let response = self.query_range(query, start, end, step)?;
+        response
+            .max()
+            .ok_or_else(|| format_err!("Failed to compute max"))
+    }
 }
 
 impl MatrixResponse {
+    pub fn max(&self) -> Option<HashMap<String, f64>> {
```
andll, line 121: To be honest I am wondering why do you want to return `Option<HashMap<String, f64>> ` instead of just returning `HashMap<String, f64>>`?
If it's empty then return empty hash map

andll:
Btw, given that prometheus is flaky, do you think at this point it can be easier to just query debug client to get metric that we want?
I know it some rework and will throw away some of work, but I am a bit hesitant to merge this, knowing it will increase failure rate

PR #4493, opened 2020-06-13
[xtest] attempt to prevent oom in linking

rexhoffman:
## Motivation

Attempt to prevent linkage errors during compilation in cargo x test.

y

## Test Plan

ci

## Related PRs

Example failure on a 2xlarge+ machine https://app.circleci.com/pipelines/github/libra/libra/20218/workflows/e3d71fb3-35f6-4bd0-b9ee-d6bb65d3249e/jobs/120474/steps

### DIFF
--- a/devtools/x/src/test.rs
+++ b/devtools/x/src/test.rs
@@ -32,6 +32,9 @@ pub struct Args {
     #[structopt(long)]
     /// Do not run tests, only compile the test executables
     no_run: bool,
+    #[structopt(long, short)]
+    /// Number of parallel jobs, defaults to # of CPUs
+    jobs: Option<u16>,
     #[structopt(long, parse(from_os_str))]
     /// Directory to output HTML coverage report to (using grcov)
     html_cov_dir: Option<OsString>,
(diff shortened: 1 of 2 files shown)


devtools/x/src/test.rs
```diff
@@ -76,7 +79,10 @@ pub fn run(mut args: Args, xctx: XContext) -> Result<()> {
     if args.lib {
         direct_args.push(OsString::from("--lib"));
     }
-
+    if args.jobs.is_some() {
```
bmwill, line 82: Use and “if let” here instead of unwrapping below

rexhoffman: @bmwill  Not sure what you're asking here,  

if let jobs = if args.jobs.??? {
}

rexhoffman: I'm just going to land for now, so folks aren't blocked.   But I would like to figure out what I should have done here instead.

rexhoffman: pattern matching is a thing, as is Some(var) = Option(somevar) {....  okay interesting.

PR #4494, opened 2020-06-13
[x] add a command to generate summaries for the LSR, LEC, default members and the full workspace

sunshowers:
Also do a guppy upgrade.

This codifies the LSR and LEC TCBs, and allows generating summaries which can be diffed by `cargo-dep-bot`.


sunshowers:
bleh, the introduction of the old itertools is annoying. I'll see if I can address this upstream.

rexhoffman:
it'd be good to generate summaries in the build, maybe during the linting?  just to prove the commands/and the output is well formed?

sunshowers:
> it'd be good to generate summaries in the build, maybe during the linting? just to prove the commands/and the output is well formed?

Hmmm, I'd expect `cargo-dep-bot` to do that (and error out if it fails).

sunshowers:
> cargo-dep-bot is doing that work, then should the code to execute the command be in x?

Good question! Ideally I'd want it to live in cargo-guppy, but I don't want to do the work to spec this out (in terms of adding configuration etc) at the moment. In that sense x is a bit of a workaround.

PR #4495, opened 2020-06-13
[language][vm] rewrite uleb serialization/deserialization and improve size checking

vgao1996:
## Summary
This introduces a few improvements to the serializer/deserializer:
- Unified handling of uleb numbers. 
- Making it possible to explicitly specify a limit on each integer type, which will be enforced by both the serializer and deserializer. The limits used to be implied by the runtime type of the integer (u8, u16, etc.), which was inflexible and hard to reason about.

## Test Plan
cargo test

### DIFF
--- a/language/vm/src/file_format_common.rs
+++ b/language/vm/src/file_format_common.rs
@@ -31,6 +31,45 @@ impl BinaryConstants {
     pub const TABLE_HEADER_SIZE: u8 = size_of::<u32>() as u8 * 2 + 1;
 }
 
+pub const TABLE_COUNT_MAX: u64 = 255;
+
+pub const TABLE_OFFSET_MAX: u64 = 0xffff_ffff;
+pub const TABLE_SIZE_MAX: u64 = 0xffff_ffff;
+pub const TABLE_CONTENT_SIZE_MAX: u64 = 0xffff_ffff;
+
+pub const TABLE_INDEX_MAX: u64 = 65535;
+pub const SIGNATURE_INDEX_MAX: u64 = TABLE_INDEX_MAX;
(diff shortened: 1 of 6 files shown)


language/vm/src/file_format_common.rs
```diff
@@ -31,6 +31,64 @@ impl BinaryConstants {
...
+
+pub const BYTECODE_COUNT_MAX: u64 = 65535;
+pub const BYTECODE_INDEX_MAX: u64 = BYTECODE_COUNT_MAX - 1;
+
+pub const STRUCT_DEF_INST_COUNT_MAX: u64 = 65535;
+pub const STRUCT_DEF_INST_INDEX_MAX: u64 = STRUCT_DEF_INST_COUNT_MAX - 1;
+
+pub const CONSTANT_COUNT_MAX: u64 = 65535;
+pub const CONSTANT_INDEX_MAX: u64 = CONSTANT_COUNT_MAX - 1;
+
```
vgao1996, line 91: Note: some of the numbers here are probably not correct. I will take another look at this. Still, please let me know if you spot something really wrong.

Also keep in mind that I haven't figured out how we should deal with the maximum number of elements in each individual table. (For example, `MODULE_HANDLE_COUNT_MAX` is not directly referenced in ser/de.) Tables are packed in a blob and we may just want to set a limit on the size of the blob as a whole.

dariorussi:
this seems fine and it may help us to control those value.
I am not sure about all the indexes defined. A table max is a u16::max and I don't think we would ever want special treatment for each table.
However, having said so I am fine with whatever we want to get in and I would say let's get this quickly "in shape" for check in.
We can and will correct max values and the number and kind of values while we finish the spec.

PR #4496, opened 2020-06-13
Eliminated $vmap function on $Vector

murali24adithya:
$Vector type already has a $select_vector function that reads a vector at a single index.
The function $vmap returned an integer map corresponding to the given vector, but was predominantly used to access values at individual indices. It also broke through the abstraction layer on the $Vector type. This change eliminates the $vmap function in favour of $select_vector.

### DIFF
--- a/language/move-prover/src/boogie_helpers.rs
+++ b/language/move-prover/src/boogie_helpers.rs
@@ -193,7 +193,7 @@ fn boogie_well_formed_expr_impl(
         Type::Vector(elem_ty) => {
             conds.push(format!("$Vector_is_well_formed({})", name));
             if !matches!(**elem_ty, Type::TypeParameter(..)) {
-                let nest_value = &format!("$vmap({})[$${}]", name, nest);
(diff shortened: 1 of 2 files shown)


shazqadeer:
@junkil-park : The MIRAI bot is failing.  Thoughts?

junkil-park:
@murali24adithya murali24adithya, could you rebase your PR on the upstream/master and push again? 
I see an error message: "Error: Unexpected prover output (expected none): Move prover returns: boogie exited with: Output { status: ExitStatus(ExitStatus(256)), stdout: "Use /help for available options\n", stderr: "Boogie: Error: unknown switch: -noinfer\n" }"

shazqadeer:
@junkil-park : I think I understand why MIRAI bot is failing.  It is because it runs on the bits pushed by @murali24adithya and not on the merged bits.  I suggest we go ahead and land this PR and I will work with Adithya to make sure he sorts this out for subsequent PRs.

PR #4497, opened 2020-06-13
[config] simplifications

davidiw:
This is a follow up to the previous config PR that looks to simplify as many fields as possible.

This includes:
* NetworkId -- both simplification and removal of extra serialization code
* secure storage token
* discovery method
* waypoint -- both simplification and proper "string" format


config/src/config/mod.rs
```diff
@@ -380,9 +380,9 @@ impl RootPath {
 mod test {
     use super::*;
 
-    const PUBLIC_FULL_NODE: &str = "src/config/test_data/public_full_node.config.yaml";
-    const VALIDATOR: &str = "src/config/test_data/validator.config.yaml";
-    const VALIDATOR_FULL_NODE: &str = "src/config/test_data/validator_full_node.config.yaml";
+    const PUBLIC_FULL_NODE: &str = "src/config/test_data/public_full_node.yaml";
+    const VALIDATOR: &str = "src/config/test_data/validator.yaml";
+    const VALIDATOR_FULL_NODE: &str = "src/config/test_data/validator_full_node.yaml";
```
JoshLind, line 385: Seems like you're just focussing on the tests here, right? Do we have a plan for updating config-builder and the corresponding docker containers to follow the new naming convention? e.g., config-builder/src/main.rs still generates "*.config.toml" files :(

types/src/waypoint.rs
```diff
@@ -144,6 +144,45 @@ impl Ledger2WaypointConverter {
...
+    where
+        D: Deserializer<'de>,
+    {
+        if deserializer.is_human_readable() {
+            let s = <String>::deserialize(deserializer)?;
+            Waypoint::from_str(&s).map_err(D::Error::custom)
+        } else {
+            // In order to preserve the Serde data model and help analysis tools,
+            // make sure to wrap our value in a container with the same name
+            // as the original type.
```
JoshLind, line 158: I'm not sure I have enough context to fully understand this, but, if we're changing the WaypointConfig enum, why not also update any "analysis tools" or dependencies, so we don't need to support this edge case?

config/src/network_id.rs
```diff
@@ -89,76 +79,7 @@ impl fmt::Display for NetworkId {
 impl NetworkId {
     /// Convenience function to specify the VFN network
     pub fn vfn_network() -> NetworkId {
-        NetworkId::private_network("vfn")
-    }
-
-    /// Creates a private network so we don't have to keep track of `NetworkInfo` outside of here.
-    pub fn private_network(name: &str) -> NetworkId {
```
gregnazario, line 96: Finally `toml` -> `yaml` gets rid of this mess

types/src/waypoint.rs
```diff
@@ -144,6 +144,45 @@ impl Ledger2WaypointConverter {
...
+impl Serialize for Waypoint {
+    fn serialize<S>(&self, serializer: S) -> std::result::Result<S::Ok, S::Error>
+    where
+        S: Serializer,
+    {
+        if serializer.is_human_readable() {
+            self.to_string().serialize(serializer)
+        } else {
+            // See comment in deserialize.
+            serializer.serialize_newtype_struct("Waypoint", &(self.version, self.value))
```
gregnazario, line 181: Isn't this a tuple struct?

config/src/config/mod.rs
davidiw: well that is indeed problematic :D but not something we should do in this PR due to the tangled aspect of that code. Perhaps once we move pre-mainnet off config-builder we can follow up. Should make sure we sync with dev-net and testnet usage -- but CT should be self-service.

types/src/waypoint.rs
davidiw: so the question is what's the right representation in a binary format... I don't know, we don't dictate that nor share it anywhere, so I suspect just being explicit about the format and leveraging the underlying style should be good, but you know, I don't want to rock that boat.

types/src/waypoint.rs
davidiw: This might be an LCS specific hack, but I saw other cases, but effectively a tuple is an anonymous struct.

PR #4499, opened 2020-06-13
[move core modules] combine standard minting script and LBR minting s…

sblackshear:
…cript

- Add `LBR::is_lbr` function
- Use it to combine the two minting scripts
- Also allows us to simplify `TransactionFeeCollection` by eliminating its `is_lbr` function


language/stdlib/modules/LBR.move
```diff
@@ -60,6 +60,12 @@ module LBR {
         move_to(association, Reserve { mint_cap, burn_cap, preburn_cap, coin1, coin2 });
     }
 
+    /// Return true if CoinType is LBR::LBR
+    public fun is_lbr<CoinType>(): bool {
+        // TODO: use constant here
```
tzakian, line 65: `currency_code<CoinType>` existing implies `is_currency` so I don't think that check should be needed (but might be good to keep in for the prover?).

Also, might be what the TODO comment is about but instead of comparing `== b"LBR"` couldn't you do:
```
Libra::is_currency<CointType>() && Libra::currency_code<CoinType>() == Libra::currency_code<LBR>()
```

sblackshear: Good call. I was thinking of a constant for the string `x"LBR"`, but the `currency_code()` approach is better.

sblackshear: > currency_code<CoinType> existing implies is_currency so I don't think that check should be needed (but might be good to keep in for the prover?).

I did want to keep the `is_currency` check because I feel like it's confusing when functions like this can abort.

PR #4501, opened 2020-06-14
[core modules] Introduce roles module  and switch over to capabilities for permissions.

tzakian:
This PR introduces a `Roles` module that gives the exact layout of roles, in particular, which ones exist, and who is created by whom.

This also adds a privilege and capability model based on these roles and that is a refinement of `&signer`; privileges are granted to a role via `add_privilege_to_account`, and capabilities can be extracted from an account and passed via `extract_privilege_to_capability`.

The main file to look at is `Roles.move`. The other changes are then adding in capability-passing instead of explicit permissions checks in the code (and the necessary test changes...).

There are a couple hacks that I had to put in order to minimize the number of semantic changes that had to be made to the tests (and are easily spotted...). I'll be making some cleanup PRs this week.

### Key invariants:
* Once an account has a role, that account does not change roles (not quite the case due to testnet right now, but working on getting rid of this ASAP).
* Once a privilege is granted to an account, that privilege is "stuck" at that account.
* A capability `Capability<P>` can only be extracted from an account holding a `Privilege<P>`. The capability can only be returned to the account that declared it[1].
* The privileges granted to a role are a module-level property of the  module in which the privilege is defined i.e. if module `A` declares a privilege `P`, then reasoning about what roles can hold a `P` privilege is a property of module `A` (and not of the `Roles` module!).

[1] Note that we really want this `Capability` type to be a "hot-potato" type that cannot be stored. However due to the rules around returning references, we cannot enforce that a capability cannot be stored today. However, we are careful to not take ownership of these capabilities in the modules (outside of `Genesis`), and only take ownership in the tests/scripts. This is fine since we are in a closed-script, and closed-module world. But, this could be a security issue in an open world since someone could try and grab a capability and store it, and we would need to have "hot-potato" types for `Capability`s in order to prevent this.


language/stdlib/modules/LBR.move
```diff
@@ -37,18 +38,23 @@ module LBR {
     // already be registered in order for this to succeed. The sender must
     // both be the correct address and have the correct permissions. These
     // restrictions are enforced in the Libra::register_currency function.
-    public fun initialize(association: &signer) {
+    public fun initialize(
+        register_currency_capability: &Capability<RegisterNewCurrency>,
+        preburn_creation_capability: &Capability<CreatePreburn>,
+        association: &signer
+    ) {
         assert(Signer::address_of(association) == 0xA550C18, 0);
```
wrwg, line 46: Use CoreAddresses::XXX?

language/stdlib/modules/Roles.move
```diff
@@ -0,0 +1,282 @@
...
+    resource struct CreateValidator {}
+    resource struct CreateSlidingNonce {}
+    resource struct TriggerReconfiguration {}
+    resource struct CreateChildVASP {}
+
+    resource struct AddTier {}
+    resource struct UpdateTier {}
+
+    /*
+    // QUESTION: Do we want these here? Or should they be defined in `Libra`? I'm leaning towards `Libra`.
```
wrwg, line 68: I feel it would be good to have all role and capability related stuff somewhere centralized.

language/stdlib/modules/LibraAccount.move
```diff
@@ -569,73 +611,118 @@ module LibraAccount {
...
+        let role_creation_cap = &borrow_global<AccountOperationsCapability>(
+            CoreAddresses::ASSOCIATION_ROOT_ADDRESS()
+        ).role_creation_cap;
+        let role = Roles::new_role(
+            treasury_compliance_role,
+            role_creation_cap,
+            &new_dd_account,
+            Roles::DESIGNATED_DEALER_ROLE_ID(),
+        );
+        Roles::grant_parent_vasp_privileges(
```
sblackshear, line 634: Intended?

language/stdlib/modules/LibraAccount.move
```diff
@@ -521,43 +540,66 @@ module LibraAccount {
...
-        make_account<Token>(new_account, auth_key_prefix, false, role_id)
+        make_account<Token>(new_account, auth_key_prefix, false, role)
     }
 
     /// Create a treasury/compliance account at `new_account_address` with authentication key
     /// `auth_key_prefix` | `new_account_address`
     public fun create_treasury_compliance_account<Token>(
-        association: &signer,
+        _tc_creation_capability: &Capability<CreateTreasuryCompliance>,
+        sliding_nonce_creation_capability: &Capability<CreateSlidingNonce>,
```
sblackshear, line 585: Interesting: if we only want sliding nonces to exist in freshly created TreasuryCompliance account, should we combine the capabilities (i.e., `publishing_sliding_nonce` asks for a `CreateTreasuryCompliance`)?

language/stdlib/modules/LibraAccount.move
```diff
@@ -569,73 +611,118 @@ module LibraAccount {
     /// `auth_key_prefix` | `new_account_address`, for non synthetic CoinType.
     /// Creates Preburn resource under account 'new_account_address'
     public fun create_designated_dealer<CoinType>(
-        association: &signer,
+        treasury_compliance_role: &RoleId,
+        dd_creation_capability: &Capability<CreateDesignatedDealer>,
+        parent_vasp_creation_capability: &Capability<CreateParentVASP>,
+        preburn_creation_capability: &Capability<CreatePreburn>,
```
sblackshear, line 617: Similar question here: should this be combined with `CreateDesignatedDealer`?

language/stdlib/transaction_scripts/unfreeze_account.move
```diff
@@ -1,11 +1,16 @@
...
 use 0x1::LibraAccount;
 use 0x1::SlidingNonce;
+use 0x1::Roles::{Self, AccountUnfreezing};
 
 /// Script for un-freezing account by authorized initiator
 /// sliding_nonce is a unique nonce for operation, see sliding_nonce.move for details
 fun main(account: &signer, sliding_nonce: u64, to_unfreeze_account: address) {
     SlidingNonce::record_nonce_or_abort(account, sliding_nonce);
-    LibraAccount::unfreeze_account(account,  to_unfreeze_account);
+    let role = LibraAccount::extract_role(account);
```
sblackshear, line 10: After reading through a few of these, I'm wondering if:
- `RoleID` wants to be a top-level resource instead of inside `LibraAccount`
- Instead of extracting/restoring the `RoleID`, `Roles::extract_privileged_to_capability` can take a `&signer`, look at the `RoleID` under `&signer`, and give back the relevant privilege

This design is possible because you have a discipline of always passing around `Capability`'s (never `RoleId`s), which is interesting. Do you think we would ever want to pass a `&RoleID` to something other than account creation?

language/stdlib/modules/LBR.move
```diff
@@ -37,18 +38,23 @@ module LBR {
     // already be registered in order for this to succeed. The sender must
     // both be the correct address and have the correct permissions. These
     // restrictions are enforced in the Libra::register_currency function.
-    public fun initialize(association: &signer) {
+    public fun initialize(
+        register_currency_capability: &Capability<RegisterNewCurrency>,
+        preburn_creation_capability: &Capability<CreatePreburn>,
+        association: &signer
```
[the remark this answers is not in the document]
tzakian: * Good point, I will update :) 
* I was following the convention that capabilities are passed first (and this seems to fit in with the conventions around `&signer`. But happy to switch around if we want to do that, but I think the main thing is that all capabilities are passed first or last (irrespective of if they're used or not))

language/stdlib/modules/DesignatedDealer.move
```diff
@@ -104,7 +106,9 @@ module DesignatedDealer {
     public fun tiered_mint<CoinType>(
         blessed: &signer, amount: u64, addr: address, tier_index: u64
     ): Libra<CoinType> acquires Dealer {
-        Association::assert_account_is_blessed(blessed);
+        // TODO/QUESTION: I don't think we need this check -- it should be checked in
```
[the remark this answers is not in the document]
tzakian: That's what I'm wondering about... there's a bit of a chicken-and-egg problem. Unless we want to have "two levels" for minting:
The `&MintCapability` (which would be defined in `Roles`) allows whoever can get access to it to try and mint. Then the second tier is currency specific (`MintCapability<CointType>` and defined in `Libra`). But this seems overly complicated to me...

language/stdlib/modules/Roles.move
```diff
@@ -0,0 +1,282 @@
...
+    use 0x1::Signer;
+    use 0x1::CoreAddresses;
+
+    /// The roleId contains both the role id and the account address that
+    /// it was created for. This way when privileges are granted to the
+    /// role, we have a way of linking the passed in `account: &signer`
+    /// and `role_id: &RoleId` together in a module-local manner.
+    resource struct RoleId {
+        role_id: u64,
+        owner_address: address,
```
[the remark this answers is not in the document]
tzakian: I was thinking that we specifically don't want that, and to make sure roles are static to the account, and therefore the role <-> account <-> privileges mapping is defined more statically. Particularly, if roles can shift around someone could grant privileges to an account with a role, and then offer that to another account which could use it to grant itself those privileges, and we lose this mapping possibly.

language/stdlib/modules/LibraAccount.move
tzakian: yes, but I was just following the code that was already there, so happy to take it out :)

language/stdlib/modules/LibraAccount.move
tzakian: Good question. I'm fine combining these (because I'm not sure about who should be able to have a sliding nonce). My thinking was to keep the privileges as fine-grained as possible, and let the roles define what set of privileges are allowed. But happy to combine these two :)

language/stdlib/modules/LibraAccount.move
tzakian: Happy to combine, but this was again in the "make the privileges as fine-grained as possible" -- when I look at `publish_preburn_resource` that taking a `CreateDesignatedDealer` capability seems odd, so that's why I split them.

language/stdlib/transaction_scripts/unfreeze_account.move
tzakian: I really like this idea, and it simplifies a lot of things. The other thing is that this also means that `&RoleId` will never be passed as a parameter which makes me very happy... 

This also makes it provable that a role can't change (there's one `move_to`, and no `move_from` from there so the `RoleId` won't need to hold the `owner_address`!)

language/stdlib/modules/LibraAccount.move
sblackshear: I agree with the principle of making the privileges as fine-grained as possible. The one thing I'm wondering is how we should handle privilege creation (i.e., publishing a resource under your account via `publish_preburn_resource` etc). That seems like the one case where it makes sense to take a role as input instead of a privilege.

For example, imagine defining (after launch) a new module that exposes privilege `X`. Privilege `X` should only be available to DD's. Should we:
(a) Do a direct write set to `Roles` to add a `CreateX` privilege + take `Capability<CreateX>` as input in `create_x`?
(b) Write `create_x` to take a `DDRole` (or  similar) as input and leave `Roles` alone?

Said differently: what do you think about a design where all `Roles::Create`* disappear and are replaced with role checks? This actually works nicely in the account creation case (I think): if you are creating a new account with role *R_creator* from role *R_created*, you would:
- Present *R_creator* to something like `Roles::new_role`. It will hand you back a `RoleId` `R_created`
- Pass *R_created* to `LibraAccount::create<relevant_account_type>`.  It can then use *R_created* to publish all the relevant resources under the freshly created account.

tzakian:
I like this idea, however, I also really like the encapsulation of roles, and I think it's a very useful invariant to prove that roles are "stuck" to an account. That being said, I also like this idea.... what about this as a compromise:

Just like we have extraction of a `Privilege` to a `Capability` we would have extraction of `RoleId` to `Capability<TypeForRoleId>`. In essence, every account of a specific `RoleId` would be granted the implicit `TypeForRoleId` privilege. Then in the code this would be passed as a capability:
```rust
public fun publish_child_vasp_credential(
   parent: &signer,
   child: &signer,
   _: &Capability<ParentVASPRole>,
)
```

and

```rust 
public fun publish_new_privilege(vasp: &signer, _: &Capability<ParentVASPRole>);
```

When the new `role_id: u64` is granted to the account this publishes the `RoleId` as today, but it would also publish the correct default `RoleIdType` to the account[1]. This could then be extracted using the same mechanisms that we have now -- `Roles::extract_privilege_to_capability<ParentVASPRole>(parent_vasp_account)` -- and would then be passed just like any other privilege.

This has a couple things that are nice about it:
* Roles are still _somewhat_ opaque -- you can only figure out what your role is.
* We only have one type of privilege floating around -- `Capability`. The role of the account is then just a "meta privilege."
* `RoleId` is still a monomorphic resource.
* This gives us the simplicity that we want. 

What do you think?


[1] Keeping `RoleId` monomorphic here is important for transaction prioritization and "role discovery" in `add_privilege_to_account`.

sblackshear:
That scheme sounds even better to me. My only question about it was answered by your footnote (which makes a very good point).

tzakian:
Awesome! In that case, I will start in on adding solidifying things and changing tests (yay....)

language/stdlib/modules/LBR.move
```diff
@@ -37,18 +38,24 @@ module LBR {
     // already be registered in order for this to succeed. The sender must
     // both be the correct address and have the correct permissions. These
     // restrictions are enforced in the Libra::register_currency function.
-    public fun initialize(association: &signer) {
-        assert(Signer::address_of(association) == 0xA550C18, 0);
+    public fun initialize(
+        association: &signer,
+        register_currency_capability: &Capability<RegisterNewCurrency>,
+        tc_capability: &Capability<TreasuryComplianceRole>,
```
sblackshear, line 44: Naming question: should our convention be `tc_capability` or `tc_role` here?

language/stdlib/modules/LibraConfig.move
```diff
@@ -152,12 +146,10 @@ module LibraConfig {
         move_to(account, Offer::redeem<ModifyConfigCapability<Config>>(account, offer_address))
     }
 
-    public fun reconfigure(account: &signer) acquires Configuration {
+    public fun reconfigure(
+        _: &Capability<AssociationRootRole>,
```
sblackshear, line 150: itsbeautiful

language/stdlib/modules/Roles.move
```diff
@@ -0,0 +1,265 @@
+address 0x1 {
+/// This module describes two things:
+/// 1. The relationship between roles, e.g. Role_A can creates accounts of Role_B
+/// 2. The granting of privileges to an account with a specific role
+/// It is important to note here that this module _does not_ describe the
+/// privileges that a specific role can have. This is a property of each of
+/// the modules that declares a privilege.
+///
+/// It also defines a capability for extracting capabilities from an
```
sblackshear, line 9:
```suggestion
/// It also defines a function for extracting capabilities from an
```

language/stdlib/modules/Roles.move
```diff
@@ -0,0 +1,265 @@
+address 0x1 {
+/// This module describes two things:
+/// 1. The relationship between roles, e.g. Role_A can creates accounts of Role_B
+/// 2. The granting of privileges to an account with a specific role
+/// It is important to note here that this module _does not_ describe the
+/// privileges that a specific role can have. This is a property of each of
+/// the modules that declares a privilege.
```
sblackshear, line 7: Might be useful to add a quick terminology primer somewhere here:
- `Privilege<T>`: label on an account signifying that the account can do `T`
- `Capability<T>`: object representing the authority to do `T`. Requested by procedures of the module that implements `T`

This is a question for later, but I'm wondering if we want a term to distinguish stuck privileges with temporary capabilities from things like `WithdrawCapability` (which are not stuck).

language/stdlib/modules/Roles.move
```diff
@@ -0,0 +1,265 @@
...
+    /// "tainted"). Capabilities can then only be restored to the account
+    /// where they were extracted from.
+    resource struct Capability<Privilege: resource> {
+        owner_address: address,
+    }
+
+    /// The internal representation of of a privilege. We wrap every
+    /// privilege witness resource here to avoid having to write extractors/restorers
+    /// for each privilege, but can instead write this generically.
+    resource struct Privilege<Priv: resource>  {
```
sblackshear, line 62: What do you think about adding an `is_extracted` bool to guarantee that there is at most one `Capability<T>` for a `Privilege<T>` under a particular account at a time?

language/stdlib/modules/Roles.move
```diff
@@ -0,0 +1,265 @@
...
+    resource struct ChildVASPRole {}
+    resource struct UnhostedRole {}
+
+    ///////////////////////////////////////////////////////////////////////////
+    // Privilege Granting
+    ///////////////////////////////////////////////////////////////////////////
+
+    /// The privilege `witness: Priv` is granted to `account` as long as
+    /// `account` has a `role` with `role.role_id == role_id`.
+    /// INVARIANT 1: Once a privilege witness `Priv` has been granted to an
```
sblackshear, line 89: Drop the 1 if this is the only invariant?

language/stdlib/modules/Roles.move
```diff
@@ -0,0 +1,265 @@
...
+    }
+
+    ///////////////////////////////////////////////////////////////////////////
+    // ROLE GRANTING
+    ///////////////////////////////////////////////////////////////////////////
+
+    /// Granted in genesis. So there cannot be any pre-existing privileges
+    /// and roles. This is _not_ called from within LibraAccount -- these
+    /// privileges need to be created before accounts can be made
+    /// (specifically, initialization og currency)
```
sblackshear, line 108:
```suggestion
    /// (specifically, initialization of currency)
```

language/stdlib/modules/Roles.move
```diff
@@ -0,0 +1,265 @@
...
+
+    /// NB: currency-related privileges are defined in the `Libra` module.
+    /// Granted in genesis. So there cannot be any pre-existing privileges
+    /// and roles.
+    /// TODO: Make explicit that this can only be called at genesis
+    public fun grant_treasury_compliance_role(
+        treasury_compliance_account: &signer,
+        _: &Capability<AssociationRootRole>,
+    ) {
+        // Grant the TC role to the treasury_compliance_account
```
sblackshear, line 128: Check `is_genesis` or assert hardcoded address here?

language/stdlib/modules/Roles.move
```diff
@@ -0,0 +1,265 @@
...
+            move_to(new_account, RoleId { role_id })
+        } else if (role_id == UNHOSTED_ROLE_ID()) {
+            // TODO(tzakian): remove unhosted creation/guard so that only
+            // assoc root can create.
+            publish_role_privilege_for_role_id(new_account, role_id);
+            move_to(new_account, RoleId { role_id })
+        } else { // Unknown account role
+            // UNRECOGNIZED_ACCOUNT_ROLE
+            abort 2
+        }
```
sblackshear, line 183: <3

language/stdlib/modules/Roles.move
```diff
@@ -0,0 +1,265 @@
...
+    /// The new module can use a capability for the role `R`;
+    /// `&Capability<R>` to guard the  granting of the privilege in order
+    /// to ensure that only those with appropriate permissions are granted
+    /// the new permission. e.g.
+    /// ```
+    /// public fun publish_new_privilege(account: &signer, _: &Capability<R>) {
+    ///    Roles::add_privilege_to_account(account, Roles::R_ROLE_ID());
+    /// }
+    ///```
+    fun publish_role_privilege_for_role_id(account: &signer, role_id: u64) {
```
sblackshear, line 197: Wondering if this would be simpler for the prover if you combined this with `new_role`--that way, it will only  have to reason about the  branching on `role_id` once.

language/stdlib/modules/LBR.move
tzakian: My thinking was `tc_capability` instead of `tc_role` here, since in my thinking the `TreasuryComplianceRole` is the "default" privilege for the TC role, and since it's wrapped in a capability this is a "capability" fully. But I don't have strong opinions about this so happy to rename if you have a preference for `tc_role` :)

language/stdlib/modules/Roles.move
```diff
@@ -0,0 +1,265 @@
...
+
+    ///////////////////////////////////////////////////////////////////////////
+    // Privilege Granting
+    ///////////////////////////////////////////////////////////////////////////
+
+    /// The privilege `witness: Priv` is granted to `account` as long as
+    /// `account` has a `role` with `role.role_id == role_id`.
+    /// INVARIANT 1: Once a privilege witness `Priv` has been granted to an
+    ///              account it remains at that account.
+    public fun add_privilege_to_account<Priv: resource>(
```
sblackshear, line 91: Wondering: can we make this private and expose wrappers without `role_id` for each role? That would also let us make the `ROLE_ID()` functions private (I think).

language/stdlib/modules/Roles.move
tzakian: Good point on terminology, I'll add it in.

In terms of unstuck capabilities (`WithdrawCapability`, `MintCapability<T>` etc), I think having a good way of distinguishing those is important. I think if we could build a unified model for representing these like this does for roles/"stuck" privileges, that would be very nice indeed...

language/stdlib/modules/Roles.move
tzakian: I'll do both :)

language/stdlib/modules/Roles.move
tzakian: ahhhh I was about to disagree, but I actually kinda like this idea a lot. Will change!

language/stdlib/modules/LBR.move
sblackshear: `tc_capability` makes sense to me; just wondering about your thinking.

PR #4508, opened 2020-06-15
Fix UDP logging

sherry-x:
## Motivation

Do not break the loop on error. this will stop logging thread when encounter any error.

Yes

## Test Plan

Build and run locally

## Related PRs

### DIFF
--- a/common/logger/src/struct_log.rs
+++ b/common/logger/src/struct_log.rs
@@ -377,8 +377,8 @@ impl UDPStructLogThread {
                     continue;
                 }
                 Err(e) => {
+                    // do not break on error, move on to the next log message
                     println!("[Logging] Error while sending data to socket: {}", e);
-                    break;
                 }
             }
         }


common/logger/src/struct_log.rs
```diff
@@ -378,7 +378,6 @@ impl UDPStructLogThread {
                 }
                 Err(e) => {
                     println!("[Logging] Error while sending data to socket: {}", e);
-                    break;
```
davidiw, line 381: curious, wouldn't this just result in the same error message over and over again? like this is a socket error at this point in time, right? or would this be perhaps that the socket is out of buffer? in which case should we do something differently?

sherry-x: @davidiw It will failed to log this current message, but it can move forward the the next log. For example, if the current message is too long and exceed the buffer size, then we would failed to send this current message, but this doesn't necessarily fail the next one.

davidiw: Can we add a note here on why it is unsafe to break, other folks might foolishly "optimize" this in the future.

PR #4511, opened 2020-06-15
[storage] new internal API: get_epoch_change_ledger_info()

msmouse:
## Motivation

in state-sync `get_ledger_info()` (now renamed to `get_epoch_change_ledger_info()` as well) calls `storage.get_ledger_info(version)` with `version` always being at an epoch boundary. This provides a new straightforward API to do what it does today (and make it more efficient.) 

(Write your answer here.)

## Test Plan

## Related PRs

### DIFF
--- a/json-rpc/src/methods.rs
+++ b/json-rpc/src/methods.rs
@@ -131,7 +131,8 @@ async fn get_account_state(
 async fn get_metadata(service: JsonRpcService, request: JsonRpcRequest) -> Result<BlockMetadata> {
     let (version, timestamp) = match serde_json::from_value::<u64>(request.get_param(0)) {
         Ok(version) => {
-            let li = service.db.get_ledger_info(version)?;
+            // TODO: fix once we have a real way to get transaction timestamps
+            let li = service.db.get_epoch_change_ledger_info(version)?;
             (version, li.ledger_info().timestamp_usecs())
         }
         _ => (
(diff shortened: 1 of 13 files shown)


json-rpc/src/methods.rs
```diff
@@ -131,7 +131,8 @@ async fn get_account_state(
 async fn get_metadata(service: JsonRpcService, request: JsonRpcRequest) -> Result<BlockMetadata> {
     let (version, timestamp) = match serde_json::from_value::<u64>(request.get_param(0)) {
         Ok(version) => {
-            let li = service.db.get_ledger_info(version)?;
+            // TODO: fix once we have a real way to get transaction timestamps
```
lightmark, line 134: @christinacelee fyi

storage/libradb/src/ledger_store/mod.rs
```diff
@@ -98,16 +98,35 @@ impl LedgerStore {
     /// [`start_epoch`, `end_epoch`). If there is no more than `limit` results, this function
```
lightmark, line 98: Should we allow search in [start, end] instead of [start, end) since the end epoch may be the latest epoch and it could have an `epoch_change_ledger_info`.

msmouse: https://github.com/libra/libra/blob/38536ae729a6b23a23e41979fb4d9309cd24c995/storage/libradb/src/lib.rs#L588

the call site offsets this by calling `.next_block_epoch()`. 

That can be a separate fix, we need to think it through before touching the interface.

PR #4514, opened 2020-06-15
[move source language] Added constant folding

tnowacki:
- Updated optimization of functions to a loop
- Added constant folding optimization
- Only folds unary ops, binary ops, and casts

## Motivation

- Needed for constants

## Test Plan

- Utilizing existing tests in #4513
- Might add another short, but exhaustive (over operators), test

### DIFF
--- a/language/move-lang/src/cfgir/constant_fold.rs
+++ b/language/move-lang/src/cfgir/constant_fold.rs
@@ -1,0 +1,342 @@
+// Copyright (c) The Libra Core Contributors
+// SPDX-License-Identifier: Apache-2.0
+
+use super::cfg::BlockCFG;
+use crate::{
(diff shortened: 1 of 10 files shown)


tnowacki:
@wrwg @junkil-park Can you take a look at the changed tests for `stackless-bytecode-generator`? Let me know if the update is fine, or if you need the optimizations turned off for these tests

wrwg:
I’m out. Shaz is the guy for this codr

tnowacki:
@shazqadeer Same question. Is this update safe, or should we turn off optimizations for these?

shazqadeer:
The updates are fine (as long at the Boogie/Z3 tests for the Move prover are all working).

PR #4518, opened 2020-06-15
[network] add counter for noise handshake

mimoo:


network/src/counters.rs
```diff
@@ -87,6 +87,15 @@ pub static LIBRA_NETWORK_DIRECT_SEND_BYTES: Lazy<HistogramVec> = Lazy::new(|| {
     .unwrap()
 });
 
+pub static LIBRA_NETWORK_NOISE_HANDSHAKE: Lazy<IntCounterVec> = Lazy::new(|| {
+    register_int_counter_vec!(
+        "libra_network_noise_handshake",
+        "Libra network noise handshake counter",
+        &["role_type", "network_type"]
```
phlip9, line 94: i think we're already using "role_type" and "network_type" for different purposes in other counters?

network/src/noise/handshake.rs
```diff
@@ -221,6 +227,11 @@ impl NoiseUpgrader {
         TSocket: AsyncRead + AsyncWrite + Unpin,
         F: Fn() -> [u8; AntiReplayTimestamps::TIMESTAMP_SIZE],
     {
+        // increment counter for metrics
+        counters::LIBRA_NETWORK_NOISE_HANDSHAKE
+            .with_label_values(&["initiator", self.auth_mode.to_str()])
+            .inc();
```
phlip9, line 233: maybe we should push down the network_id into NoiseUpgrader and include that, then rename the "role_type" label into "connection_origin" x { "inbound", "outbound" }. also rename the "network_type" label to "auth_mode"?

network/src/counters.rs
```diff
@@ -87,6 +87,15 @@ pub static LIBRA_NETWORK_DIRECT_SEND_BYTES: Lazy<HistogramVec> = Lazy::new(|| {
...
+    )
+    .unwrap()
+});
```
phlip9, line 97: i would also add a commit with the updated grafana dashboard displaying the noise handshake counter

network/src/counters.rs
mimoo: I can't re-use the same labels across counters?

network/src/noise/handshake.rs
mimoo: if we have the `network_id` we don't need the `auth_mode` right?

network/src/counters.rs
phlip9: err no, i think it's just confusing to use `role_type := initiator | responder` in one place and `role_type := validator | fullnode` in another. what i'm suggesting is to rename "role_type" and "network_type" for this specific counter since labels with the same name are being used differently in other counters.

network/src/noise/handshake.rs
phlip9: right, i think we can just use `network_id` instead of this `auth_mode`/`network_type`

davidiw:
@mimoo, you have a lot of outstanding PRs :P

mimoo:
sorry bout' the spam! I'll try to get this one in soon

mimoo:
(but as for others, feel free to close and I'll re-open when I get to it)

PR #4521, opened 2020-06-15
Create tiers for DD by default and handle incorrect tier indeces

moeziniaf:


language/stdlib/modules/DesignatedDealer.move
```diff
@@ -89,7 +89,7 @@ module DesignatedDealer {
         // If the tier_index is one past the bounded tiers, minting is unbounded
         let number_of_tiers = Vector::length(tiers);
         let tier_check = &mut false;
-        if (tier_index == number_of_tiers) {
+        if (tier_index >= number_of_tiers) {
```
dmitri-perelman, line 92: Why did you change it to `>=`?

moeziniaf: Currently, this is not an issue due to (1) the assert i added (2) the fact that 4 tiers are created on setup. But previously if the tier_index was greater than the number of tiers then we would enter the else loop and access an invalid index of the tier list (which was causing the NATIVE_FN_ERROR). This >= should prevent this from happening in case (1) or (2) are changed in the future.

language/stdlib/modules/DesignatedDealer.move
```diff
@@ -105,7 +105,8 @@ module DesignatedDealer {
         blessed: &signer, amount: u64, addr: address, tier_index: u64
     ): Libra<CoinType> acquires Dealer {
         Association::assert_account_is_blessed(blessed);
-
+        // INVALID TIER_INDEX (if tier is 4, can mint unlimited)
+        assert(tier_index <= 4, 66);
```
sblackshear, line 109: I would feel better if this check + the `amount` check were in `tiered_mint_` to prevent a different caller from forgetting to check them. This might also let you get rid of the `tier_index >= number_of_tiers` check in `tiered_mint`.

language/stdlib/modules/DesignatedDealer.move
```diff
@@ -41,7 +41,7 @@ module DesignatedDealer {
         let tiers = &mut dealer.tiers;
         let number_of_tiers: u64 = Vector::length(tiers);
         // INVALID_TIER_ADDITION
-        assert(number_of_tiers <= 4, 3);
+        assert(number_of_tiers <= 4, 31);
```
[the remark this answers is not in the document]
moeziniaf: We need to be able to update the tiers though - they will probably be different across DDs and currencies

language/stdlib/modules/DesignatedDealer.move
dmitri-perelman: I definitely agree that we don't want to enter the `else` clause here.
Just thought that there should be a better way to handle it, may be with separate branches explicitly checking `>` and `==` conditions (the current solution is introducing a potential loophole in case the function is not called by `tiered_mint`)

language/stdlib/modules/DesignatedDealer.move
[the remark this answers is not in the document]
sblackshear: Fair, that would be a lot of code duplication. But does `add_tier` need to be public (i.e., can't you initialize the tiers in `publish_designated_dealer_credential`)? That would make it obvious that the tiers are initialized + eliminate some code.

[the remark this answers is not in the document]
moeziniaf: yes - you're right. I think the specs are not fully set it stone yet - it may be that they do only want one/two default tiers and then be able to add more. I might leave it for now while I find out more about exactly what the desired usage is.

moeziniaf:
> I'm not sure about the `add_tier` abstraction: it seems that if we'd need some dynamic update, it could be about changing the thresholds (and not just about adding a new tier).
> However, I agree that it goes beyond the scope of this specific fix.

Agreed. I am not sure if there is a use case for < 4 tiers, but we can find out

PR #4531, opened 2020-06-16
[move-prover] Specifying event

kkmc:
## Motivation

Specify Event.move.

Currently, this file is specified fully at the procedure level, however the main uniqueness property cannot be proved because LCS::serialize does not return the same length byte arrays for inputs that have the same primitive types. This causes collisions for the `fresh_guid` function.

As an example, having requires in the example below enabled me to prove that the fresh_guid function is injective. I will investigate what axioms we need for the serialize builtin function.
```
    public fun fresh_guids(c1: &mut EventHandleGenerator, c2: &mut EventHandleGenerator): (vector<u8>, vector<u8>) {
        (fresh_guid(c1), fresh_guid(c2))
    }
    spec fun fresh_guids {
        requires len(serialized_ehg_addr(c1)) == len(serialized_ehg_addr(c2));
        ensures c1.addr != c2.addr ==> result_1 != result_2;
    }
```

I also ran into issues trying to prove that EventHandleGenerator is only created. I think this is required to prove that the EventHandleGenerator at any address at any program point (transient state) is unique. Please refer to the comments in the specifications. 

Yes.

## Test Plan

Existing testsuite.

## Related PRs

None.

### DIFF
--- a/language/stdlib/modules/Event.move
+++ b/language/stdlib/modules/Event.move
@@ -69,6 +69,186 @@ module Event {
     public fun destroy_handle<T: copyable>(handle: EventHandle<T>) {
         EventHandle<T> { counter: _, guid: _ } = handle;
     }
+
+    // ****************** SPECIFICATIONS *******************
+
+    /**
+        The Event module defines an `EventHandleGenerator` that is used to create
+        unqiue `EventHandle`s based on a strictly monotonically increasing counter.
+        An `EventHandle` counts the number of events emitted to a unique handle
+        for a partilar GUID.
+    */
+
+    /// # Module specifications
+    spec module {
+        /// Helper function that returns whether or not an EventHandleGenerator is
+        /// initilaized at the given address `addr`.
+        define initialized_ehg(addr: address): bool {
+            exists<EventHandleGenerator>(addr)
+        }
+        /// Helper function that returns the EventHandleGenerator at `addr`.
+        define ehg(addr: address): EventHandleGenerator {
+            global<EventHandleGenerator>(addr)
+        }
+        // Helper function that returns the serialized counter of the `EventHandleGenerator` ehg.
+        define serialized_ehg_counter(ehg: EventHandleGenerator): vector<u8> {
+            LCS::serialize(ehg.counter)
+        }
+        // Helper function that returns the serialized address of the `EventHandleGenerator` ehg.
+        define serialized_ehg_addr(ehg: EventHandleGenerator): vector<u8> {
+            LCS::serialize(ehg.addr)
+        }
+    }
+
+    /// ## Creation and Uniqueness of `EventHandleGenerator`s
+
+    spec module {
+        global ehg_destroyed: bool;
+    }
+    spec struct EventHandleGenerator {
+        invariant pack ehg_destroyed = ehg_destroyed;
+        invariant unpack ehg_destroyed = true;
+    }
+    spec schema EventHandleGeneratorNeverDestroyed {
+        /// **Informally:** No EventHandleGenerator should ever be destroyed.
+        invariant module !ehg_destroyed;
+    }
+    spec schema EventHandleGeneratorAtSameAddress {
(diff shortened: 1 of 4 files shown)


language/stdlib/modules/Event.move
```diff
@@ -69,6 +75,189 @@ module Event {
...
         EventHandle<T> { counter: _, guid: _ } = handle;
     }
+
+    // ****************** SPECIFICATIONS *******************
+
+    /// # Module specifications
+    spec module {
+        /// Helper function that returns whether or not an EventHandleGenerator is
+        /// initilaized at the given address `addr`.
+        define initialized_ehg(addr: address): bool {
```
DavidLDill, line 85: I think we should save "initialized" for things that are called during genesis.  So, perhaps ehg_exists or whatever?

language/stdlib/modules/Event.move
```diff
@@ -69,6 +75,189 @@ module Event {
...
+            exists<EventHandleGenerator>(addr)
+        }
+        /// Helper function that returns the EventHandleGenerator at `addr`.
+        define ehg(addr: address): EventHandleGenerator {
```
DavidLDill, line 89: I prefer get_ehg

language/stdlib/modules/Event.move
```diff
@@ -69,6 +75,189 @@ module Event {
...
+    }
+
+    /// ## Management of EventHandleGenerators
+
+    spec module {
+        /// Variable that is true whenever an `EventHandleGenerator` is destroyed.
+        /// This should never to true to preserve uniqueness of EventHandleGenerators.
+        global ehg_destroyed: bool;
+    }
+    spec struct EventHandleGenerator {
```
DavidLDill, line 109: Maybe all we really care about is:
(1) event handle generator at an address persists ( old(exists<EventHandleGenerator>(a) ==> exists<EventHandleGenerator>(a)
(2) count is non-decreasing (actually, it remains constant except for the function that increments it).

At this point, I'm not sure the suggestion is better than pack (but pack seems more imperative).

language/stdlib/modules/Event.move
```diff
@@ -69,6 +75,189 @@ module Event {
...
+        ensures global<EventHandleGenerator>(Signer::get_address(account))
+                    == EventHandleGenerator { counter: 0, addr: Signer::get_address(account) };
+    }
+
+    // Switch documentation context back to module level.
+    spec module {}
+
+    /// ## Uniqueness of EventHandleGenerator and Counter Monotonicity
+
+    spec schema CounterMonotonicallyIncreases {
```
DavidLDill, line 155: There are ehg counters and eh counters.  Clearer to say which counter?

language/stdlib/modules/Event.move
```diff
@@ -69,6 +75,189 @@ module Event {
...
+    }
+    spec schema EventHandleGeneratorNeverDestroyed {
+        /// **Informally:** No EventHandleGenerator should ever be destroyed.
+        invariant module !ehg_destroyed;
+    }
+    spec schema EventHandleGeneratorAtSameAddress {
+        /// **Informally:** An EventHandleGenerator should be located at `addr` and is never moved.
+        invariant module forall addr: address where initialized_ehg(addr): ehg(addr).addr == addr;
+    }
+    spec module {
```
DavidLDill, line 123: It seems obvious that fresh_guid cannot violate this property!  Must investigate...

language/stdlib/modules/Event.move
DavidLDill, line 155: A more precise spec would be that the counter does not change unless new_event_handler is called (then it increases by 1).  However, I don't know if anything depends on this more precise property.

language/stdlib/modules/Event.move
kkmc: > At this point, I'm not sure the suggestion is better than pack (but pack seems more imperative).

What suggestion are you referring to?

PR #4532, opened 2020-06-16
[move-prover] Updated spec doc on aborts_if true.

kkmc:
Updated the spec-lang user documentation to include an experimental section
that contains information about the new smoke test pragma.

## Motivation

Update spec-lang doc on experimental smoke test pragma.

Yes.
## Test Plan

Not relevant.
## Related PRs

None.

### DIFF
--- a/language/move-prover/doc/user/spec-lang.md
+++ b/language/move-prover/doc/user/spec-lang.md
@@ -183,6 +183,20 @@ spec fun increment {
 }
 ```
 
+### Experimental
+
+We are working towards supporting smoke testing to check unsound or imprecise models. To turn this option on, set the smoke\_test pragma to true as shown below. By default, the experimental smoke\_test pragma is set to false.
+
+List of checks smoke\_test pragma runs:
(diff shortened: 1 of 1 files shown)


language/move-prover/doc/user/spec-lang.md
```diff
@@ -183,6 +183,20 @@ spec fun increment {
 }
 ```
 
+### Experimental
+
+We are working towards supporting smoke testing to check unsound or imprecise models. To turn this option on, set the smoke\_test pragma to true as shown below. By default, the experimental smoke\_test pragma is set to false.
+
+List of checks smoke\_test pragma runs:
+- Checks if `aborts_if true;` or `ensures false;` is provable for a function; this usually indicates a bug in the specification or implementation.
```
DavidLDill, line 191: When does it check aborts_if true and when does it check ensures false.  Or are they equivalent?  If equivalent, say "or, equivalently, ensures false" (that, of course, will raise the slightly non-obvious question of why they are equivalent).

kkmc: I believe if `aborts_if true;` holds, this means that you can prove `ensures false;`. But if you can prove `ensures false;`, then it could be that you always abort so that the condition !$abort_flag ==> false holds vacuously (false ==> false). I think it could also be the case that model has contradictory specifications (e.g. through spec block assumptions, invariants, or a requires) or the model is unsound and anything can be proved.

PR #4534, opened 2020-06-16
[move core][DRAFT] Added module to query genesis status.

DavidLDill:
This file implements a state variable to track the status of genesis.
It is in a separate leave module, because it will be used by lots of
other modules to check whether functions are being called during the
genesis process or not. Putting it in Genesis would introduce a ton
of dependencies, but using GenesisState will not.

In the Genesis module, the very first thing `initialize` should
to is to call begin_initialize().  The last thing it should do
is call end_initialized().

The implementation is a state machine with three states: before,
during, and after genesis. `begin_genesis` advances the state
machine from before to during, and `end_genesis` advances it
from during to after. Each of these transitions is irreversible.

Functions in other modules, especially initialization functions, can
call predicates is_during_genesis(): bool and is_after_genesis: bool
return if genesis is underway or complete, respectively.

## Motivation

(Write your answer here.)

## Test Plan

## Related PRs

### DIFF
--- a/language/stdlib/modules/GenesisState.move
+++ b/language/stdlib/modules/GenesisState.move
@@ -1,0 +1,145 @@
+address 0x1 {
+
+/// This file implements a state variable to track the status of genesis.
+/// It is in a separate leave module, because it will be used by lots of
+/// other modules to check whether functions are being called during the
+/// genesis process or not. Putting it in Genesis would introduce a ton
+/// of dependencies, but using GenesisState will not.
+module GenesisState {
+    use 0x1::Signer;
+    use 0x1::CoreAddresses;
+
+    /// A singleton resource that stores the state of the genesis process
+    /// on genesis_address()
+    resource struct GenesisState {
+        genesis_complete: bool,
+    }
+
+    /// **TODO:** May want to move to CoreAddresses, or just use
+    /// ASSOCIATION_ROOT_ADDRESS
+    public fun genesis_address(): address {
+        CoreAddresses::ASSOCIATION_ROOT_ADDRESS()
+    }
+
+    /// Abort if genesis has already begun, to protect against unverified
+    /// code re-starting genesis.
+    public fun begin_genesis(account: &signer) {
+        assert(Signer::address_of(account) == genesis_address(), 1942);
+        move_to(account, GenesisState { genesis_complete: false })
+    }
+
+    public fun end_genesis() acquires GenesisState {
+        let gen_state = borrow_global_mut<GenesisState>(genesis_address());
+        gen_state.genesis_complete = true;
+    }
+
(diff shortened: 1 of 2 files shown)


language/stdlib/modules/GenesisState.move
```diff
@@ -0,0 +1,145 @@
...
+/// other modules to check whether functions are being called during the
+/// genesis process or not. Putting it in Genesis would introduce a ton
+/// of dependencies, but using GenesisState will not.
+module GenesisState {
+    use 0x1::Signer;
+    use 0x1::CoreAddresses;
+
+    /// A singleton resource that stores the state of the genesis process
+    /// on genesis_address()
+    resource struct GenesisState {
```
tzakian, line 14: My initial thought was that this `genesis_complete` field could be able be elided, and that this might make the other functions simpler: 
* `begin_genesis` publishes the resource, 
* `is_during_genesis` checks for the existence of the resource,
* `end_genesis` unpublishes the resource, and 
* `is_after_genesis` checks that the resource doesn't exist. 

The only problem with this is that it looks like doing this might make the `TransBeginToAfter` impossible to  prove? Since  if the resource is unpublished nothing prevents `begin_genesis` from being called again. 

If that's the case, this looks good to me. Only thing might be getting rid of `genesis_address` and moving that in to the `CoreAddresses` module.

sblackshear:
Potentially a dumb question, but do we really need a separate module/separate state for this? I was really hoping we could use `LibraTimestamp::time() == 0` to implement `is_during_genesis()` and  `LibraTimestamp::time() > 0` for `is_after_gensis()`. We can change the code of `Gensis.move` to publish the time resource + set the time to 0 at the very beginning, then set it to  1 (or some other value) at the very end. Is there a reason this won't work?

language/stdlib/modules/GenesisState.move
DavidLDill: "The only problem" was the problem I was trying to solve.  I really didn't want people to be able to re-enter genesis.

I agree about CoreAddresses.

tzakian: Ah ok awesome. 

One other thought I had with this, is that this boolean separates it into three phases (before, during, after). Would it be ok to instead only have two? e.g. (genesis, after genesis)? Then `is_genesis` looks for the non-existence, and `is_after_genesis` looks for the existence. This would have the irreversible property that you want, but there is no distinction between "before" and "during" genesis.

tzakian:
I think Sam has a great point here! I had a quick look, and I think LibraTimeStamp should work for this. I was worried about dependencies, but having a look it only pulls in `CoreAddresses` and `Signer`, so  we don't need to worry about circular includes. This also has the irreversible property to genesis as well.

PR #4537, opened 2020-06-16
[move-modules] Only the Association may trigger reconfiguration

valerini:
This PR makes reconfiguration callable by the Association account only.
Add/remove a validator to/from the set and update_and_reconfigure (to copy consensus configs from validators accounts into the on-chain config) may only be called by the Association.


language/e2e-tests/src/common_transactions.rs
```diff
@@ -82,6 +82,23 @@ pub fn add_validator_txn(
     )
 }
 
+/// Returns a transaction to update validators' configs and reconfigure
+///   (= emit reconfiguration event and change the epoch)
+pub fn update_validators_and_reconfigure_txn(
```
davidiw, line 87: should we just call this "reconfigure"

curious what other folks think about the naming here.

valerini: I agree - the name is too long... Reconfigure event is also emitted on add_validator_txn and remove_validator_txn.

language/stdlib/transaction_scripts/register_validator.move
```diff
@@ -1,5 +1,4 @@
 script {
-    use 0x1::LibraSystem;
```
sblackshear, line 2: I think this is now identical to `rotate_validator_config`--we can probably kill it?

language/stdlib/transaction_scripts/rotate_validator_config.move
```diff
@@ -1,5 +1,4 @@
 script {
-    use 0x1::LibraSystem;
```
sblackshear, line 2: Wondering: do we need both this and `rotate_consensus_pubkey`?

language/e2e-tests/src/common_transactions.rs
sblackshear: +1 to shorter name

language/e2e-tests/src/tests/validator_set_management.rs
```diff
@@ -106,6 +113,17 @@ fn validator_rotate_key() {
 
     let txn = rotate_consensus_pubkey_txn(&validator_account, vec![251; 32], 1);
     let output = executor.execute_transaction(txn);
+    executor.apply_write_set(output.write_set());
+
+    assert_eq!(
+        output.status(),
+        &TransactionStatus::Keep(VMStatus::new(StatusCode::EXECUTED))
+    );
+    println!("{:?}", output);
```
sblackshear, line 122: Delete?

language/stdlib/transaction_scripts/rotate_validator_config.move
valerini: You are right, we don't need the rotate_consensus_pubkey. Nunced it.

zekun000:
lgtm, re the distinguish between validator set change from other on-chain config, we discuss this before.

ValidatorSet is one of the on-chain config, so it's either we have a type aware event or we have a single event for all. personally I think former are better but it also requires more work.

davidiw:
> This looks great to me!
> 
> One loose end worth discussing: changing on-chain configs also triggers reconfiguration https://github.com/libra/libra/blob/master/language/stdlib/modules/LibraConfig.move#L164.
> 
> This doesn't seem like a big deal because only LibraRoot or TreasuryCompliance can change these (I think). But it might be better if we distinguished between reconfiguration for a validator set change and reconfiguration for an on-chain config update...

Without reviewing any code... my personal preference would be that we explicitly differentiate an update from a reconfiguration, but I suspect that this is impossible if we directly mutate an on-chain config state?

config/management/src/validator_config.rs
```diff
@@ -64,19 +64,21 @@ impl ValidatorConfig {
 
         // Step 2) Generate transaction
 
+        // TODO(davidiw): This is currently not supported
+        // let sender = self.owner_address;
+        let sender = account_address::from_public_key(&operator_key);
+
```
davidiw, line 70: @JoshLind now the comment makes sense!!!!!!

config/management/src/validator_config.rs
```diff
@@ -64,19 +64,21 @@ impl ValidatorConfig {
...
         // TODO(philiphayes): remove network identity pubkey field from struct when
         // transition complete
-        let script = transaction_builder::encode_register_validator_script(
+        let script = transaction_builder::encode_set_validator_config_script(
+            sender,
```
davidiw, line 74: this will be owner account, whereas the actual sender would be the operator

execution/executor/tests/storage_integration_test.rs
```diff
@@ -33,7 +33,8 @@ use std::convert::TryFrom;
 use storage_interface::DbReaderWriter;
 use transaction_builder::{
     encode_block_prologue_script, encode_mint_script, encode_publishing_option_script,
-    encode_rotate_consensus_pubkey_script, encode_transfer_with_metadata_script,
+    encode_reconfigure_script, encode_set_validator_config_script,
```
davidiw, line 36: as a big to do, we need to move these out of storage_integration_tests and create a proper LibraMoveE2E testsuite in libra/testsuite/libra_move_interface or something

config/management/src/validator_config.rs
JoshLind: @davidiw woooo! 🍾

types/src/validator_config.rs
```diff
@@ -13,7 +13,6 @@ use serde::{Deserialize, Serialize};
 pub struct ValidatorConfigResource {
     pub validator_config: Option<ValidatorConfig>,
     pub delegated_account: Option<AccountAddress>,
-    pub is_certified: bool,
```
davidiw, line 16: oh wow, so we're getting rid of all this too :/

I guess simplicity is best...

language/stdlib/modules/LibraAccount.move
```diff
@@ -914,8 +910,19 @@ module LibraAccount {
         make_account<Token>(new_account, auth_key_prefix, false)
     }
 
-    ///////////////////////////////////////////////////////////////////////////
-    // End of the proof of concept code
-    ///////////////////////////////////////////////////////////////////////////
+    public fun create_validator_operator_account<Token>(
+        creator_account: &signer,
```
tzakian, line 914: I believe only the association root can create a validator operator. If that's the case, could you have this take a `_: &Capability<AssociationRootRole>` (similar to `create_parent_vasp_account`)?

language/stdlib/modules/LibraSystem.move
```diff
@@ -98,12 +94,11 @@ module LibraSystem {
...
     public fun remove_validator(
-        operator: &signer,
+        assoc: &signer,
         account_address: address
     ) acquires CapabilityHolder {
-        // Validator's operator can remove its certified validator from the validator set
-        assert(Signer::address_of(operator) ==
-                            ValidatorConfig::get_operator(account_address), 22);
+        // Only the Association may remove validator from the validator set
+        assert(Signer::address_of(assoc) == CoreAddresses::ASSOCIATION_ROOT_ADDRESS(), 1002);
```
tzakian, line 101: Same here as in `add_validator`

language/stdlib/modules/LibraSystem.move
```diff
@@ -63,23 +63,19 @@ module LibraSystem {
...
+        assoc: &signer,
         account_address: address
     ) acquires CapabilityHolder {
-        // Validator's operator can add its certified validator to the validator set
-        assert(
-            Signer::address_of(operator) == ValidatorConfig::get_operator(account_address),
-            22
-        );
+        // Only the Association may add a validator to the validator set
+        assert(Signer::address_of(assoc) == CoreAddresses::ASSOCIATION_ROOT_ADDRESS(), 1002);
```
tzakian, line 75: Instead of doing a check on the `assoc` address, could you instead pass in a `_: &Capability<AssociationRootRole>` here? e.g.
```rust
public fun add_validator(
   _: &Capability<AssociationRootRole>,
   account_address: address
) ... {
// Can remove line 75, no checks are needed since the capability was passed in.
....
}
```

language/stdlib/modules/LibraSystem.move
```diff
@@ -116,19 +111,13 @@ module LibraSystem {
...
+    // For all of the validators the information from ValidatorConfig will
     // get copied into the ValidatorSet.
-    // Invalid or decertified validators will get removed from the Validator Set.
+    // Invalid validators will get removed from the Validator Set.
     // NewEpochEvent event will be fired.
-    public fun update_and_reconfigure(account: &signer) acquires CapabilityHolder {
-        assert(is_authorized_to_reconfigure_(account), 22);
+    public fun update_and_reconfigure(assoc: &signer) acquires CapabilityHolder {
+        // Only the Association may trigger the reconfiguration
+        assert(Signer::address_of(assoc) == CoreAddresses::ASSOCIATION_ROOT_ADDRESS(), 1002);
```
tzakian, line 120: Similar comment here as in `add_validator` and `remove_validator`:

```suggestion
    public fun update_and_reconfigure(_: &Capability<AssociationRootRole>) acquires CapabilityHolder {
  
```

types/src/validator_config.rs
valerini: We had the is_certified flag before in order to give the Association a way to remove validator from the set - it could set the is_certified to false. Now, the add/remove validator are callable by the Association only, so it seems there is no good reason to keep this flag. For additional certification/decertification the freeze/unfreeze methods on the account could be used.

But if you think that this flag is still useful, we can keep it - please let me know!

secure/key-manager/src/lib.rs
```diff
@@ -203,14 +204,34 @@ where
...
         let new_key = self.storage.rotate_key(CONSENSUS_KEY)?;
         info!("Successfully rotated the consensus key in secure storage.");
         COUNTERS.completed_consensus_key_rotations.inc();
-        self.submit_key_rotation_transaction(new_key)
+        let result = self.submit_key_rotation_transaction(new_key);
+        self.submit_reconfiguration_transaction()?;
+        result
+    }
+
+    pub fn submit_reconfiguration_transaction(&self) -> Result<(), Error> {
```
JoshLind, line 220: @davidiw, @valerini how do we feel about adding the "submit_reconfiguration_transaction" logic to the key manager, here? My feeling is that this is a little out of place and breaks the abstractions (given that a key manager only operates on each validator node and won't have access to the association key in production). 

The way that the key manager operates at the moment is that is assumes an external force is responsible for performing reconfigurations. Perhaps this should happen somewhere else? e.g., the management tool?

execution/executor/tests/storage_integration_test.rs
```diff
@@ -115,19 +116,34 @@ fn test_reconfiguration() {
     // Create a dummy block prologue transaction that will bump the timer.
     let txn2 = encode_block_prologue_script(gen_block_metadata(1, validator_account));
 
-    // rotate the validator's connsensus pubkey to trigger a reconfiguration
+    // rotate the validator's connsensus pubkey
```
JoshLind, line 119: nit: connsensus 😄

secure/key-manager/src/lib.rs
davidiw: yeah, this isn't good, we'll need to address this, I guess I'll leave it up to you to decide in this PR or another...

valerini: Indeed, this needs to be moved out! I can put this function in the test.rs. Let me try to do it in the next PR as this PR is becoming too big.

valerini:
> One request @valerini, please create an issue where you had to hack to make this work and add assignees as best you can, so we can track the areas which are needed more attention.

Great idea, just created: https://github.com/libra/libra/issues/4676

PR #4546, opened 2020-06-16
[move-prover] Working around a restriction in type inference of spec lang

### LINKED ISSUE #4541: [Bug] Type inference does not work for elements in a vector
I added the following code to RecoveryAddress.move:

```
spec module {
    define spec_holds_key_for(
        caps: vector<LibraAccount::KeyRotationCapability>,
        addr: address): bool
    {
        exists cap in caps: cap.account_address == addr
    }
}
```

This is giving me the following error:
```
error: type `?5` cannot be resolved as a struct

     ┌── ../stdlib/modules/RecoveryAddress.move:123:33 ───
     │
 123 │             exists cap in caps: cap.account_address == addr
     │                                 ^^^^^^^^^^^^^^^^^^^
     │
```

It seems that the prover fails to infer that `cap` is a `LibraAccount::KeyRotationCapability` struct.

wrwg:
Type inference in spec lang does not work in some instances where overloading and lambdas are involved (lambdas are generated implicitly from quantifiers). This PR excludes lambda expressions from checking during overload resolution, and checks them after the overload has been identified, providing a proper expected type which helps inference. This is a restriction to the overload resolution model which, however, in practice will not hurt as we will turn off overloading anyway for user functions.

The PR also makes `all`, `any`, and `domain` internal functions not longer reachable by users as discussed with Todd.

## Motivation

Closes #4541.

Yes

## Test Plan

New test

## Related PRs

NA

### DIFF
--- a/language/move-lang/src/parser/syntax.rs
+++ b/language/move-lang/src/parser/syntax.rs
@@ -1079,7 +1079,7 @@ fn parse_quant_cont<'input>(is_forall: bool, tokens: &mut Lexer<'input>) -> Resu
         // Built `domain<ty>()` expression.
         tokens.advance()?;
         let ty = parse_type(tokens)?;
-        make_builtin_call(ty.loc, "domain", Some(vec![ty]), vec![])
+        make_builtin_call(ty.loc, "$spec_domain", Some(vec![ty]), vec![])
     } else {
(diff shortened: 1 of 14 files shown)


wrwg:
@tnowacki FYI

PR #4547, opened 2020-06-16
[core modules] Add `ToLBRExchangeRateUpdateEvent` to the `Libra` module

tzakian:
Adds an event for when the `to_lbr_exchange_rate` for a currency is updated.


types/src/account_config/events/exchange_rate_update.rs
```diff
@@ -0,0 +1,38 @@
...
+
+use crate::account_config::LIBRA_MODULE_NAME;
+use anyhow::Result;
+use move_core_types::{
+    identifier::{IdentStr, Identifier},
+    move_resource::MoveResource,
+};
+use serde::{Deserialize, Serialize};
+
+/// Struct that represents a MintEvent.
```
sblackshear, line 12:
```suggestion
/// Struct that represents a ToLBRExchangeRateUpdateEvent.
```

types/src/account_config/events/exchange_rate_update.rs
```diff
@@ -0,0 +1,38 @@
...
+    new_to_lbr_exchange_rate: u64,
+}
+
+impl ToLBRExchangeRateUpdateEvent {
+    /// Exchange rates are represented as 32|32 fixed-point numbers
+    pub fn new_to_lbr_exchange_rate(&self) -> f32 {
+        (self.new_to_lbr_exchange_rate as f32) / 2f32.powf(32f32)
+    }
+
+    /// Return the code for the currency that was minted
```
sblackshear, line 25:
```suggestion
    /// Return the code for the currency that was updated
```

language/stdlib/modules/Libra.move
```diff
@@ -551,7 +564,8 @@ module Libra {
             mint_events: Event::new_event_handle<MintEvent>(account),
             burn_events: Event::new_event_handle<BurnEvent>(account),
             preburn_events: Event::new_event_handle<PreburnEvent>(account),
-            cancel_burn_events: Event::new_event_handle<CancelBurnEvent>(account)
+            cancel_burn_events: Event::new_event_handle<CancelBurnEvent>(account),
+            exchange_rate_update_events: Event::new_event_handle<ToLBRExchangeRateUpdateEvent>(account)
```
sblackshear, line 568: Hopefully our tests will catch this, but just wanted to double-check that this isn't changing any important event keys for handles created later in genesis...

language/stdlib/modules/Libra.move
```diff
@@ -630,6 +644,14 @@ module Libra {
         assert_assoc_and_currency<FromCoinType>(account);
         let currency_info = borrow_global_mut<CurrencyInfo<FromCoinType>>(CoreAddresses::CURRENCY_INFO_ADDRESS());
         currency_info.to_lbr_exchange_rate = lbr_exchange_rate;
+        Event::emit_event(
+            &mut currency_info.exchange_rate_update_events,
+            ToLBRExchangeRateUpdateEvent {
+                currency_code: *&currency_info.currency_code,
```
sblackshear, line 650: @tnowacki: does it work to write `copy currency_info.currency_code` here (and if it does, is there a reason to prefer one of `copy` and `*&`?

types/src/account_config/events/exchange_rate_update.rs
```diff
@@ -0,0 +1,38 @@
...
+#[derive(Debug, Serialize, Deserialize)]
+pub struct ToLBRExchangeRateUpdateEvent {
+    currency_code: Identifier,
+    new_to_lbr_exchange_rate: u64,
+}
+
+impl ToLBRExchangeRateUpdateEvent {
+    /// Exchange rates are represented as 32|32 fixed-point numbers
+    pub fn new_to_lbr_exchange_rate(&self) -> f32 {
+        (self.new_to_lbr_exchange_rate as f32) / 2f32.powf(32f32)
```
sblackshear, line 22: This requires more explanation (at least for me...)

language/stdlib/modules/Libra.move
tzakian: I actually wound up writing `copy currency_info.currency_code` here at first, but the parser was unhappy with that. I didn't try `copy &currency_info.currency_code`, but I'm guessing the compiler wants to see ` &` in front of a field access to know that it's grabbing a reference to that field.

language/stdlib/modules/Libra.move
tzakian: Our tests do indeed catch this :) (see the change to `block_metadata.rs`).

types/src/account_config/events/exchange_rate_update.rs
tzakian: ahhh yes.... I got this from here:
https://medium.com/incredible-coder/converting-fixed-point-to-floating-point-format-and-vice-versa-6cbc0e32544e#:~:text=Converting%20from%20fixed%2Dpoint%20to%20floating%2Dpoint&text=Divide%20the%20number%20by%202,to%20the%20power%20of%20n).

and also tested it (valid up-to rounding errors): 
https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=2456fe91c9c3034233555f4aa4fe7e08

n4ss:
That looks great!

PR #4550, opened 2020-06-16
[language_e2e_test] Provide function execution functionality in FakeExecutor

runtian-zhou:
…xecutor

## Motivation

This functionality is requested by @sblackshear. We want the ability to execute functions in move so as to better simulate the account creation process, capability management, etc.

Yes

### DIFF
--- a/language/e2e-tests/src/executor.rs
+++ b/language/e2e-tests/src/executor.rs
@@ -14,7 +14,8 @@ use libra_crypto::HashValue;
 use libra_state_view::StateView;
 use libra_types::{
     access_path::AccessPath,
-    account_config::{AccountResource, BalanceResource},
+    account_address::AccountAddress,
+    account_config::{AccountResource, BalanceResource, CORE_CODE_ADDRESS},
     block_metadata::{new_block_event_key, BlockMetadata, NewBlockEvent},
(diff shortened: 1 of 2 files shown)


language/e2e-tests/src/executor.rs
```diff
@@ -256,4 +266,41 @@ impl FakeExecutor {
...
+        module_name: &str,
+        function_name: &str,
+        type_params: Vec<TypeTag>,
+        args: Vec<Value>,
+        sender: AccountAddress,
+    ) {
+        let write_set = {
+            let cost_table = zero_cost_schedule();
+            let mut cost_strategy = CostStrategy::system(&cost_table, GasUnits::new(100_000_000));
+            let vm = MoveVM::new();
```
sblackshear, line 289: Is creating a fresh VM for every account creation expensive? I suppose we could add this to `self` if so.

runtian-zhou: That is the main question in my mind as well! The main concern is if we were to introduce the writeset functionality in those tests we would need to think about the VM dropping strategy. That's why I did it in a fresh way. Plus the current transaction execution api creates a new MoveVM instance at every invocation. I'd love to add this to self as well if that's more efficient. Thoughts? cc @dariorussi

PR #4552, opened 2020-06-17
[cluster-test] Reduce get_pod_node_and_ip to 120 seconds

andll:
We had insanely large retry limit on get_pod_node_and_ip, this cause fail slow and creates backlog when something is wrong.
This allows to fail faster


andll:
@bors-libra r=changyonghu p=100

Boosting priority as this might help draining land queue

PR #4553, opened 2020-06-17
[LBT] increase workflow timeout to account for queue

sausagee:
## Motivation
Increased timeout to 40 minutes, which is sufficient to include build
and test with retries and wait time in queue.

## Test Plan
LBT in CI


andll:
@bors-libra r+

Thank you!

Issue #4512, opened 2020-06-15
Stream libra trace into structured log


andll:
Currently libra trace entries are sent to local storage only. We want to send them to structured logging using `send_struct_log!` macro. We can probably use single name for all libra trace events, like `libra_trace` and keep libra trace objects inside single `.data` field.

In terms of configurability we want it to be enabled when ran in production, and disabled in tests

PR #4559, opened 2020-06-17
[libra-trace] Stream libra trace into structured log

changyonghu:
Closes: #4512
I checked Kibana dashboard. It looks good to me. 
Also about the struct_log, I use "libra_trace" as name, and in .data field, I use "trace_event" or "trace_edge" as the key, not sure if I should use "libra_trace" here.


andll:
@bors-libra r+

Great!

PR #4561, opened 2020-06-17
[language][vm] remove interpreter functionality for move_to_sender an…

sblackshear:
…d Transaction::sender

All uses of these deprecated bytecodes have been removed from the standard library and the movelang  tests. The only remaining uses are in the Move prover. The prover doesn't actually run the code, so deprecating the interpreter support is a good commitment device that will prevent us from adding more uses of these bytecodes.

In time (hopefully very  soon), we'll refactor the prover tests and eliminate these bytecodes altogether.

### DIFF
--- a/language/move-vm/runtime/src/interpreter.rs
+++ b/language/move-vm/runtime/src/interpreter.rs
@@ -57,8 +57,6 @@ pub(crate) struct Interpreter {
     operand_stack: Stack,
     /// The stack of active functions.
     call_stack: CallStack,
-    /// The sender of the transaction. This can materialize as an argument to a script.
-    sender: AccountAddress,
 }
 
 impl Interpreter {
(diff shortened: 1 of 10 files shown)


language/move-vm/runtime/src/interpreter.rs
```diff
@@ -1007,10 +990,8 @@ impl Frame {
                             .push(Value::bool(!lhs.equals(&rhs)?))?;
                     }
                     Bytecode::GetTxnSenderAddress => {
-                        cost_strategy.charge_instr(Opcodes::GET_TXN_SENDER)?;
-                        interpreter
-                            .operand_stack
-                            .push(Value::address(interpreter.sender))?;
+                        // GetTxnSenderAddress is deprecated and will be removed soon
+                        return Err(VMStatus::new(StatusCode::UNKNOWN_INVARIANT_VIOLATION_ERROR));
```
vgao1996, line 994: Could you add `.with_message(...)` to indicate what triggered this error? It will help save a lot of time in case someone actually hits this case and need to figure out the origin of the error.

language/move-vm/runtime/src/interpreter.rs
```diff
@@ -1087,27 +1068,13 @@ impl Frame {
...
-                            resolver,
-                            data_store,
-                            interpreter.sender,
-                            *sd_idx,
-                            Interpreter::move_to_sender,
-                        )?;
-                        cost_strategy.charge_instr_with_size(Opcodes::MOVE_TO_SENDER, size)?;
+                    Bytecode::MoveToSender(_) => {
+                        // MoveToSender is deprecated and will be removed soon
+                        return Err(VMStatus::new(StatusCode::UNKNOWN_INVARIANT_VIOLATION_ERROR));
```
vgao1996, line 1073: Same here.

language/move-vm/runtime/src/interpreter.rs
```diff
@@ -1087,27 +1068,13 @@ impl Frame {
...
-                            interpreter.sender,
-                            *si_idx,
-                            self,
-                            Interpreter::move_to_sender,
-                        )?;
-                        cost_strategy
-                            .charge_instr_with_size(Opcodes::MOVE_TO_SENDER_GENERIC, size)?;
+                    Bytecode::MoveToSenderGeneric(_) => {
+                        // MoveToSender is deprecated and will be removed soon
+                        return Err(VMStatus::new(StatusCode::UNKNOWN_INVARIANT_VIOLATION_ERROR));
```
vgao1996, line 1077: Same here.

language/move-vm/runtime/src/interpreter.rs
sblackshear: Good call--will  push the comment into the `with_message`

PR #4563, opened 2020-06-17
[config] Make test cases default and use them for smoke tests

davidiw:
This makes the test configs defaults and uses them in the management smoke test. This makes a nice framework for ensuring correctness of our configs and harder to break production until we actually embody these configs in more smoke tests / cluster test.

### DIFF
--- a/config/management/src/smoke_test.rs
+++ b/config/management/src/smoke_test.rs
@@ -3,12 +3,9 @@
 
 use crate::{constants, layout::Layout, storage_helper::StorageHelper};
 use config_builder::{BuildSwarm, SwarmConfig};
-use libra_config::{
-    config::{
-        DiscoveryMethod, Identity, NetworkConfig, NodeConfig, OnDiskStorageConfig, RoleType,
-        SecureBackend, SeedPeersConfig, WaypointConfig, HANDSHAKE_VERSION,
-    },
-    network_id::NetworkId,
+use libra_config::config::{
+    Identity, NodeConfig, OnDiskStorageConfig, RoleType, SecureBackend, SeedPeersConfig,
+    WaypointConfig, HANDSHAKE_VERSION,
 };
 use libra_crypto::ed25519::Ed25519PrivateKey;
 use libra_secure_storage::{CryptoStorage, KVStorage, Value};
@@ -70,16 +67,7 @@ fn smoke_test() {
         let operator_key = helper.operator_key(&ns, &ns_shared).unwrap();
 
         let validator_account = account_address::from_public_key(&operator_key);
-        let mut config = NodeConfig::default();
-
-        let mut network = NetworkConfig::network_with_id(NetworkId::Validator);
-        network.discovery_method = DiscoveryMethod::None;
-        network.mutual_authentication = true;
-        config.validator_network = Some(network);
-
-        let mut network = NetworkConfig::network_with_id(NetworkId::Public);
-        network.discovery_method = DiscoveryMethod::None;
-        config.full_node_networks = vec![network];
+        let mut config = NodeConfig::default_for_validator();
         config.randomize_ports();
 
         let validator_network = config.validator_network.as_mut().unwrap();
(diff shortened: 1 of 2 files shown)


davidiw:
@aching / @christinacelee 

just a heads up into where we currently are in terms of setting up our production network in terms of testability...

config/src/config/mod.rs
```diff
@@ -320,6 +320,27 @@ impl NodeConfig {
         self.set_data_dir(test.temp_dir().unwrap().to_path_buf());
         self.test = Some(test);
     }
+
+    #[cfg(any(test, feature = "fuzzing"))]
+    pub fn default_for_public_full_node() -> Self {
+        let contents = std::include_str!("test_data/public_full_node.yaml");
+        let path = "default_for_public_full_node";
```
JoshLind, line 327: any reason we don't just inline path into the error message?

config/src/config/mod.rs
```diff
@@ -409,12 +426,8 @@ mod test {
...
-        let _ = vec![PUBLIC_FULL_NODE, VALIDATOR, VALIDATOR_FULL_NODE]
-            .iter()
-            .map(|path| {
-                NodeConfig::load_config(PathBuf::from(path))
-                    .unwrap_or_else(|e| panic!("Error in {}: {}", path, e))
-            })
-            .collect::<Vec<_>>();
+        NodeConfig::default_for_public_full_node();
+        NodeConfig::default_for_validator();
+        NodeConfig::default_for_validator_full_node();
```
gregnazario, line 431: I guess this just tests that they load?

config/management/src/smoke_test.rs
```diff
@@ -70,16 +67,7 @@ fn smoke_test() {
...
-
-        let mut network = NetworkConfig::network_with_id(NetworkId::Validator);
-        network.discovery_method = DiscoveryMethod::None;
-        network.mutual_authentication = true;
-        config.validator_network = Some(network);
-
-        let mut network = NetworkConfig::network_with_id(NetworkId::Public);
-        network.discovery_method = DiscoveryMethod::None;
-        config.full_node_networks = vec![network];
+        let mut config = NodeConfig::default_for_validator();
```
gregnazario, line 70: Woo!  so much cleaner

config/management/src/smoke_test.rs
```diff
@@ -139,11 +127,11 @@ fn smoke_test() {
             config.base.waypoint = WaypointConfig::FromStorage(backend);
         }
         config.execution.genesis = Some(genesis.clone());
+        config.execution.genesis_file_location = PathBuf::from("");
```
gregnazario, line 130: Should we make a file for this to load later too?

config/management/src/smoke_test.rs
```diff
@@ -202,38 +190,23 @@ fn check_connectivity(swarm: &mut LibraSwarm, expected_peers: i64) -> bool {
...
-    full_node_config.full_node_networks.push(pfn);
-    validator_config.full_node_networks.push(internal_fn);
+    // The FN's first network is the external, public network, it needs to swap listen addresses
+    // with the validator's VFN and to copy it's key access:
+    let pfn = &mut full_node_config.full_node_networks[0];
+    let v_vfn = &mut validator_config.full_node_networks[0];
+    pfn.identity = v_vfn.identity.clone();
+    let temp_listen = v_vfn.listen_address.clone();
+    v_vfn.listen_address = pfn.listen_address.clone();
+    pfn.listen_address = temp_listen;
```
gregnazario, line 205: Is it possible to just set them correctly in the first place rather than swap them?

config/src/config/mod.rs
davidiw: yeah, let's just verify they load correctly...

config/management/src/smoke_test.rs
davidiw: well I guess the issue is the templates contain some examples, so I'm half considering making actual defaults that contain no examples... or instead to make these functions return really default configs...

config/management/src/smoke_test.rs
davidiw: we can probably move toward that direction, but we'd have to change the current genesis structure and start a full node with each libra-node... maybe a future PR?

config/src/config/mod.rs
davidiw: wanted to avoid the weird nested new line that would come as a result :P

PR #4573, opened 2020-06-17
[cluster test] Always wipe data on startup 

andll:
We lost delete_data on startup in refactoring.
This works fine most of times, but if cluster did not scale down from previous run, it keeps data and it messes up the run

Also, this adds some additional logging to node allocation - we log aws instance id and private ip of allocated node, this all is combined into `KubeNode` structure


andll:
@bors-libra r+ p=100
Boosting priority as this fixes flackiness of cluster test

PR #4579, opened 2020-06-17
[language] all modules published to 0x1 under closed publishing

sblackshear:
- The Libra and Move VM both require that modules must be published under the address of the transaction sender (*)
- In V1, only the holder of the PublishModule capability can publish new modules.
- This capability is held by the LibraRoot account 0xA550C18
- All existing  modules are published under 0x1

This PR relaxes the (*) restriction in the Libra VM only to allow us to publish modules under 0x1 via transactions sent from 0xA550C18. If the on-chain publishing option is set to closed, the module will be published under 0x1. If the on-chain publishing option is set to open, the module will be published under the sender's address.


sblackshear:
> YESSSSSSSS. This will make things so much easier!
> 
> It's probably a good idea to remove the account at `0x1` that's created in genesis, and update the module publishing tests in e2e that test it as well in this PR.

Good call, will do both!

jolestar:
if Module publish is open, the stdlib new module will be at 0xA550C18?

sblackshear:
The stdlib modules will be at 0x1. 

If the on-chain publishing option is set to closed, the modules published after genesis will be under 0x1. If the on-chain publishing option is set to open, modules published after genesis will be under the sender's address.

PR #4581, opened 2020-06-17
[network] peer manager structured logging

gregnazario:
This structured logging, similar to connectivity manager and discovery
to keep track of events that come through the actor loop.

### More structured logging thoughts
1. Generics prevent from having constant `field` types, and can either then create the field type at runtime, or just use a constant string (I opted for the second)
2. Serialization in component types causes a chain of serialization changes when adding a higher level object.  This is great for serialization, but maybe some parts we want to serialize differently for logs rather than real usage.
3. Generics' types must also implement Serialize even if they aren't used (I opted to ignore a field due to it, because it became complicated)

### DIFF
--- a/network/src/protocols/rpc/mod.rs
+++ b/network/src/protocols/rpc/mod.rs
@@ -65,6 +65,7 @@ use futures::{
 };
 use libra_logger::prelude::*;
 use libra_types::PeerId;
+use serde::Serialize;
 use std::{collections::HashMap, fmt::Debug, time::Duration};
 
 pub mod error;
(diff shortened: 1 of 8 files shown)


network/src/protocols/rpc/mod.rs
```diff
@@ -102,18 +103,20 @@ pub struct InboundRpcRequest {
 }
 
 /// A wrapper struct for an outbound rpc request and its associated context.
-#[derive(Debug)]
+#[derive(Debug, Serialize)]
```
davidiw, line 106: do we benefit much with this one :P

gregnazario: It gives us the protocol, and the duration, so I guess not really.

I think this is harder to determine what should and shouldn't be serialized

PR #4582, opened 2020-06-17
libra_config secure_storage_config

Rustinante:
## Motivation

Moved the `identity_key`, `peer_id`, `waypoint` functions from libra secure storage into `libra_config` as methods for the corresponding structs.

Yes

## Test Plan

cargo xtest


config/management/src/smoke_test.rs
```diff
@@ -235,10 +236,10 @@ fn attach_validator_full_node(
     );
 
     let v_vfn_network_address = v_vfn.listen_address.clone();
-    let v_vfn_pub_key = libra_secure_storage::config::identity_key(v_vfn).public_key();
+    let v_vfn_pub_key = identity_key(v_vfn).public_key();
```
davidiw: its not very rusty to import functions, we instead import modules / crates, so if you really wanted to "optimize" the right approach would have been config::identity_key

but that's not very clear so I left in the full path.

davidiw:
I /think/ those errors are flaky, let's say what homu says

@bors-libra r+

Rustinante:
Thanks!! By the way, I think there's still a little bit of a code duplication between `Identity::peer_id_from_config` and the `peer_id()` method of `NetworkConfig`, which handles the variant of `Identity::FromStorage` and panics when on `Identity::None`. I'm not sure if `Identity::peer_id_from_config` should be updated with the new implementation.

davidiw:
> Thanks!! By the way, I think there's still a little bit of a code duplication between `Identity::peer_id_from_config` and the `peer_id()` method of `NetworkConfig`, which handles the variant of `Identity::FromStorage` and panics when on `Identity::None`. I'm not sure if `Identity::peer_id_from_config` should be updated with the new implementation.

I think you could clean up the old APIs in another commit, we absolutely do not need a lot of that fancy logic any more... it was primarily there to limit how much you would need to import -- and some of it was legacy due to the pain that toml introduced.

Rustinante:
Okay, I just pushed another commit to address that. Please let me know if this is what we wanted. Thanks

davidiw:
> Okay, I just pushed another commit to address that. Please let me know if this is what we wanted. Thanks

In the future, as an external contributor, your life would be made easier by making a new PR and letting this PR land :P

davidiw:
And now that I look at this, it seems like this indeed will not work due to limitations in the PPC build environment because for some reason... our ppc build depends on configs :O

I'll need to see what's that about...

davidiw:
https://github.com/libra/libra/pull/4607

once this lands, ideally your PR should be able to land

Rustinante:
I see, thanks a lot for the fix!!

config/src/config/network_config.rs
```diff
@@ -260,7 +248,17 @@ impl Identity {
     pub fn peer_id_from_config(&self) -> Option<PeerId> {
```
davidiw, line 248: oh, I actually meant we should delete this snippet.

My recommendation is if you can remove this latest commit, we can land this pr... otherwise eliminate this function.

Rustinante: Okay, I just removed that function. Right now there's an `.expect("identity key should be present")` call in the `peer_id()` method, I'm not sure if there'd be any use cases where a return type of `Option<PeerId>` would be more desirable.

Rustinante:
Just rebased the commits, the build is working now!

config/src/config/network_config.rs
davidiw: Definitely something to contemplate! I think we'll need to do some improvements in the network code base to catch up.

PR #4583, opened 2020-06-17
[cluster test] Allow to restart instance with different tag

andll:
* kube swarm now maps node by pod_name, not by InstanceConfig. This allows to change instance config and restart instance on same node
* Introducing Instance::instance_config to grab instance config of this instance
* Introducing InstanceConfig::replace_tag to update tag on InstanceConfig


testsuite/cluster-test/src/instance.rs
```diff
@@ -80,6 +80,42 @@ struct K8sInstanceInfo {
     kube: ClusterSwarmKube,
 }
 
+impl InstanceConfig {
+    pub fn replace_tag(&mut self, new_tag: String) -> Result<()> {
+        match self {
```
ankushagarwal, line 85: since self is already &mut, I don't think you need `ref mut` in the match handles

andll: Yep, you are right

PR #4584, opened 2020-06-17
[cluster-test] Low TPS latency experiment

snakech:
## Motivation

We currently have an experiment that stress tests node and measures both latency and TPS in stress case.

We want a separate experiment that will submit low TPS(~50 txn/sec) and measure latency in this low tps case.

This should be part of new_pre_release and new_perf_suite as defined in suite.rs.
As a result there should be a new line in changelog showing latency in a low tps case.

(Write your answer here.)

## Test Plan

## Related PRs

### DIFF
--- a/testsuite/cluster-test/src/experiments/performance_benchmark.rs
+++ b/testsuite/cluster-test/src/experiments/performance_benchmark.rs
@@ -38,6 +38,8 @@ pub struct PerformanceBenchmarkParams {
     help = "Duration of an experiment in seconds"
     )]
     pub duration: u64,
+    #[structopt(long, help = "Set fixed tps during perf experiment")]
+    pub tps: Option<u64>,
 }
 
 pub struct PerformanceBenchmark {
@@ -47,6 +49,7 @@ pub struct PerformanceBenchmark {
     percent_nodes_down: usize,
     duration: Duration,
     trace: bool,
+    tps: Option<u64>,
 }
 
 pub const DEFAULT_BENCH_DURATION: u64 = 120;
@@ -57,6 +60,16 @@ impl PerformanceBenchmarkParams {
             percent_nodes_down,
             duration: DEFAULT_BENCH_DURATION,
             trace: false,
+            tps: None,
+        }
+    }
+
+    pub fn new_fixed_tps(percent_nodes_down: usize, fixed_tps: u64) -> Self {
+        Self {
+            percent_nodes_down,
+            duration: DEFAULT_BENCH_DURATION,
+            trace: false,
+            tps: Some(fixed_tps),
         }
     }
 }
@@ -85,6 +98,7 @@ impl ExperimentParam for PerformanceBenchmarkParams {
             percent_nodes_down: self.percent_nodes_down,
             duration: Duration::from_secs(self.duration),
             trace: self.trace,
+            tps: self.tps,
         }
     }
 }
@@ -100,16 +114,14 @@ impl Experiment for PerformanceBenchmark {
         try_join_all(futures).await?;
         let buffer = Duration::from_secs(60);
         let window = self.duration + buffer * 2;
-        let emit_job_request = if context.emit_to_validator {
-            EmitJobRequest::for_instances(
-                self.up_validators.clone(),
-                context.global_emit_job_request,
-            )
+        let instances = if context.emit_to_validator {
+            self.up_validators.clone()
         } else {
-            EmitJobRequest::for_instances(
-                self.up_fullnodes.clone(),
-                context.global_emit_job_request,
-            )
+            self.up_fullnodes.clone()
+        };
+        let emit_job_request = match self.tps {
+            Some(tps) => EmitJobRequest::fixed_tps(instances, tps),
+            None => EmitJobRequest::for_instances(instances, context.global_emit_job_request),
         };
         let emit_txn = context.tx_emitter.emit_txn_for(window, emit_job_request);
         let trace_tail = &context.trace_tail;
@@ -208,7 +220,9 @@ impl Experiment for PerformanceBenchmark {
 
 impl Display for PerformanceBenchmark {
     fn fmt(&self, f: &mut Formatter<'_>) -> Result<(), Error> {
-        if self.percent_nodes_down == 0 {
+        if let Some(tps) = self.tps {
+            write!(f, "fixed tps {}", tps)
+        } else if self.percent_nodes_down == 0 {
             write!(f, "all up")
         } else {
             write!(f, "{}% down", self.percent_nodes_down)
--- a/testsuite/cluster-test/src/tx_emitter.rs
+++ b/testsuite/cluster-test/src/tx_emitter.rs
@@ -124,6 +124,28 @@ impl EmitJobRequest {
             },
         }
     }
+
+    pub fn fixed_tps_params(instance_count: usize, tps: u64) -> (usize, u64) {
+        if tps < 1 {
+            panic!("Target tps {} can not less than 1", tps)
+        }
+        let num_workers = tps as usize / instance_count + 1;
+        let wait_time = (instance_count * num_workers * 1000_usize / tps as usize) as u64;
+        (num_workers, wait_time)
+    }
+
+    pub fn fixed_tps(instances: Vec<Instance>, tps: u64) -> Self {
+        let (num_workers, wait_time) = EmitJobRequest::fixed_tps_params(instances.len(), tps);
+        Self {
+            instances,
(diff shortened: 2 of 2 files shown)


testsuite/cluster-test/src/experiments/performance_benchmark.rs
```diff
@@ -38,6 +38,12 @@ pub struct PerformanceBenchmarkParams {
     help = "Duration of an experiment in seconds"
     )]
     pub duration: u64,
+    #[structopt(
+        long,
+        default_value = "0",
+        help = "Set fixed tps during perf experiment"
+    )]
+    pub tps: u64,
```
andll, line 46: Ideally you want to avoid magic values. Right now this magic value is 0, meaning that 'don't use this field.
Instead, you can have this field as `Option<u64>`

testsuite/cluster-test/src/suite.rs
```diff
@@ -40,6 +40,9 @@ impl ExperimentSuite {
         experiments.push(Box::new(
             PerformanceBenchmarkParams::new_nodes_down(10).build(cluster),
         ));
+        experiments.push(Box::new(
+            PerformanceBenchmarkParams::new_fixed_tps(0,50).build(cluster),
```
andll, line 44: So the TPS we would care about in this experiment is something lower, like maybe 5 TPS

testsuite/cluster-test/src/suite.rs
```diff
@@ -57,6 +60,9 @@ impl ExperimentSuite {
...
```
andll, line 64: Same here

testsuite/cluster-test/src/tx_emitter.rs
```diff
@@ -124,6 +124,26 @@ impl EmitJobRequest {
...
+
+    pub fn fixed_tps(instances: Vec<Instance>, tps: u64) -> Self {
+        let mut worker_num: usize = 10;
+        while instances.len() * worker_num > tps as usize {
+            if worker_num == 1 {
+                break;
+            }
+            worker_num -= 1;
+        }
+        let wait_time = (instances.len() * worker_num * 1000_usize / tps as usize) as u64;
```
andll, line 136: Can you run this with `--run perf -- --tps 5` and provide link to dashboard and output?

testsuite/cluster-test/src/experiments/performance_benchmark.rs
```diff
@@ -47,6 +52,7 @@ pub struct PerformanceBenchmark {
     percent_nodes_down: usize,
     duration: Duration,
     trace: bool,
+    tps: u64,
```
andll, line 55: This field should be option too, by same logic?)

testsuite/cluster-test/src/experiments/performance_benchmark.rs
```diff
@@ -208,7 +231,9 @@ impl Experiment for PerformanceBenchmark {
 
 impl Display for PerformanceBenchmark {
     fn fmt(&self, f: &mut Formatter<'_>) -> Result<(), Error> {
-        if self.percent_nodes_down == 0 {
+        if self.tps != 0 {
+            write!(f, "low tps {}", self.tps)
```
andll, line 235: Can probably replace 'low' -> 'fixed'

testsuite/cluster-test/src/experiments/performance_benchmark.rs
snakech: I think this filed is constructed by PerformanceBenchmarkParams, we set it is optional in PerformanceBenchmarkParams and we may use 0 to be the default. I can change it to option as well

andll: Yeah, this should be option for the same reason - avoiding 'magical' values

testsuite/cluster-test/src/experiments/performance_benchmark.rs
```diff
@@ -101,15 +118,25 @@ impl Experiment for PerformanceBenchmark {
         let buffer = Duration::from_secs(60);
         let window = self.duration + buffer * 2;
         let emit_job_request = if context.emit_to_validator {
-            EmitJobRequest::for_instances(
```
andll, line 104: Quick nit - you can avoid too much code duplication by splitting this nested if+matcg into to:
```
let instances = if context.emit_to_validator {self.up_validators.clone()} else {...};
match self.tps {...}
```

testsuite/cluster-test/src/experiments/performance_benchmark.rs
```diff
@@ -208,7 +220,10 @@ impl Experiment for PerformanceBenchmark {
 
 impl Display for PerformanceBenchmark {
     fn fmt(&self, f: &mut Formatter<'_>) -> Result<(), Error> {
-        if self.percent_nodes_down == 0 {
+        let tps = self.tps.unwrap_or(0);
+        if tps != 0 {
```
andll, line 224:
```
if let Some(tps) = self.tps {...
```

testsuite/cluster-test/src/tx_emitter.rs
```diff
@@ -124,6 +124,26 @@ impl EmitJobRequest {
             },
         }
     }
+
+    pub fn fixed_tps(instances: Vec<Instance>, tps: u64) -> Self {
+        let mut worker_num: usize = 10;
+        while instances.len() * worker_num > tps as usize {
```
andll, line 130: I have a strong feeling that this loop can be replaced with a division operator

testsuite/cluster-test/src/tx_emitter.rs
```diff
@@ -124,6 +124,26 @@ impl EmitJobRequest {
             },
         }
     }
+
+    pub fn fixed_tps(instances: Vec<Instance>, tps: u64) -> Self {
+        let mut worker_num: usize = 10;
```
andll, line 129: worker_num -> num_workers

testsuite/cluster-test/src/tx_emitter.rs
```diff
@@ -377,6 +397,8 @@ impl SubmissionWorker {
                 let now = Instant::now();
                 if wait_util > now {
                     time::delay_for(wait_util - now).await;
+                    // we do not include wait_time as latency
+                    tx_offset_time += (wait_util - now).as_millis() as u64;
```
andll, line 401: Should we not make this change for now? (since we see that it does not really fix the problem)

testsuite/cluster-test/src/tx_emitter.rs
andll, line 136: Do you mind extracting out function like
fn fixed_tps_params(instance_count: usize, tps: u64) -> (wait_time, num_workers)

And adding few unit tests on this fn demonstrating how it behaves?

testsuite/cluster-test/src/tx_emitter.rs
```diff
@@ -124,6 +124,32 @@ impl EmitJobRequest {
             },
         }
     }
+
+    pub fn fixed_tps_params(instance_count: usize, tps: u64) -> (usize, u64) {
+        let tps = if tps < 1 { 1 } else { tps };
```
andll, line 129: I don't think we need this check?
`--tps 0` is clearly a error, we can just panic there

testsuite/cluster-test/src/tx_emitter.rs
```diff
@@ -124,6 +124,32 @@ impl EmitJobRequest {
...
+        let mut num_workers: usize = 10;
+        while instance_count * num_workers > tps as usize {
```
andll, line 131: I think there was some misunderstanding with my comment, i did not mint to `num_workers /= 2` every loop
I think this entire loop seems like an attempt to divide `tps / (instance_count * num_workers)` (possibly some +-1 here and there)

testsuite/cluster-test/src/tx_emitter.rs
```diff
@@ -814,3 +840,27 @@ impl fmt::Display for TxStatsRate {
...
+    pub fn test_fixed_tps_params() {
+        let inst_num = 30;
+        let target_tps = 10;
+        let (num_workers, wait_time) = EmitJobRequest::fixed_tps_params(inst_num, target_tps);
+        assert_eq!(num_workers, 1usize);
+        assert_eq!(wait_time, 3000u64);
+    }
+
+    #[test]
+    pub fn test_fixed_tps_params_with_invalid_tps() {
```
andll, line 858: I don't think we need to test this

snakech:
> Nice!

thx for reviewing~!

PR #4591, opened 2020-06-17
[cluster test] Full node health check

snakech:
## Motivation

We call wait_until_all_healthy at the end of experiment, to make sure all nodes recover from whatever happened in experiment.

Added new full node health check to wait for full node to catch up for a validators in this function.

After this new health check is introduced, try re-running --perf-suite and confirmed that there is no expired transactions in 3 region simulation

## Test Plan

## Related PRs

### DIFF
--- a/testsuite/cluster-test/src/health/fullnode_check.rs
+++ b/testsuite/cluster-test/src/health/fullnode_check.rs
@@ -1,0 +1,83 @@
+// Copyright (c) The Libra Core Contributors
+// SPDX-License-Identifier: Apache-2.0
+
+#![forbid(unsafe_code)]
+
+use crate::{
+    cluster::Cluster,
+    health::{HealthCheck, HealthCheckContext},
+    instance::Instance,
+};
+use async_trait::async_trait;
+use futures::future::join_all;
+use once_cell::sync::Lazy;
+use std::{collections::HashMap, env};
+
+pub static THRESHOLD: Lazy<i64> = Lazy::new(|| {
+    if let Ok(v) = env::var("FULL_NODE_HEALTH_THRESHOLD") {
+        v.parse()
+            .expect("Failed to parse FULL_NODE_HEALTH_THRESHOLD")
+    } else {
+        20000_i64
+    }
+});
+
+pub struct FullNodeHealthCheck {
+    cluster: Cluster,
+}
+
+impl FullNodeHealthCheck {
+    pub fn new(cluster: Cluster) -> FullNodeHealthCheck {
+        Self { cluster }
+    }
+}
+
+async fn get_version(instance: &Instance) -> (&Instance, i64) {
+    let res = instance
+        .debug_interface_client()
+        .get_node_metric("libra_state_sync_committed_version{}")
+        .await;
+    let content = match res {
+        Ok(res) => res.unwrap_or_default(),
+        _ => 0i64,
+    };
+    (instance, content)
+}
+
+#[async_trait]
+impl HealthCheck for FullNodeHealthCheck {
+    async fn verify(&mut self, ctx: &mut HealthCheckContext) {
+        let validators = self.cluster.validator_instances();
+        let fullnodes = self.cluster.fullnode_instances();
+
+        let futures = validators.iter().map(get_version);
+        let val_latest_versions = join_all(futures).await;
+        let val_latest_versions: HashMap<String, i64> = val_latest_versions
+            .into_iter()
+            .map(|(instance, version)| (instance.validator_index(), version))
+            .collect();
+
+        let futures = fullnodes.iter().map(get_version);
+        let fullnode_latest_versions = join_all(futures).await;
+
+        for (fullnode, fullnode_version) in fullnode_latest_versions {
+            let index = fullnode.validator_index();
+            let val_version = val_latest_versions.get(&index).unwrap();
+            if val_version - fullnode_version > *THRESHOLD {
+                ctx.report_failure(
+                    format!("val-{}", index),
+                    format!(
+                        "fullnode {} state sync committed version: {} is behind validator: {}",
+                        fullnode.peer_name(),
+                        fullnode_version,
+                        val_version,
+                    ),
+                );
+            }
+        }
+    }
+
+    fn name(&self) -> &'static str {
+        "fullnode_check"
+    }
+}
--- a/testsuite/cluster-test/src/instance.rs
+++ b/testsuite/cluster-test/src/instance.rs
@@ -5,6 +5,7 @@
 
 use crate::cluster_swarm::cluster_swarm_kube::ClusterSwarmKube;
 use anyhow::{format_err, Result};
+use debug_interface::AsyncNodeDebugClient;
 use libra_config::config::NodeConfig;
 use libra_json_rpc_client::{JsonRpcAsyncClient, JsonRpcBatch};
 use once_cell::sync::Lazy;
@@ -307,6 +308,15 @@ impl Instance {
         } else {
             Ok(())
         }
+    }
+
+    pub fn debug_interface_client(&self) -> AsyncNodeDebugClient {
+        AsyncNodeDebugClient::new(
+            self.http_client.clone(),
+            self.ip(),
+            self.debug_interface_port
+                .expect("debug_interface_port is not known on this instance") as u16,
+        )
     }
 }
 
--- a/testsuite/cluster-test/src/experiments/performance_benchmark_three_region_simulation.rs
+++ b/testsuite/cluster-test/src/experiments/performance_benchmark_three_region_simulation.rs
@@ -98,7 +98,7 @@ impl Experiment for PerformanceBenchmarkThreeRegionSimulation {
     }
 
     fn deadline(&self) -> Duration {
-        Duration::from_secs(420)
+        Duration::from_secs(600)
     }
 }
 
--- a/testsuite/cluster-test/src/health/debug_interface_log_tail.rs
+++ b/testsuite/cluster-test/src/health/debug_interface_log_tail.rs
@@ -40,16 +40,10 @@ impl DebugPortLogWorker {
         let pending_messages = Arc::new(AtomicI64::new(0));
         let (trace_sender, trace_receiver) = mpsc::channel();
         let trace_enabled = Arc::new(AtomicBool::new(false));
-        let http_client = reqwest::Client::new();
         for instance in cluster.all_instances() {
             let (started_sender, started_receiver) = mpsc::channel();
             started_receivers.push(started_receiver);
-            let debug_interface_port = instance
-                .debug_interface_port()
-                .expect("Debug interface port is not found")
-                as u16;
-            let client =
-                AsyncNodeDebugClient::new(http_client.clone(), instance.ip(), debug_interface_port);
+            let client = instance.debug_interface_client();
             let debug_port_log_worker = DebugPortLogWorker {
                 instance: instance.clone(),
                 client,
--- a/testsuite/cluster-test/src/health/mod.rs
+++ b/testsuite/cluster-test/src/health/mod.rs
@@ -5,6 +5,7 @@
 
 mod commit_check;
 mod debug_interface_log_tail;
+mod fullnode_check;
 mod liveness_check;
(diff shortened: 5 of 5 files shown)


testsuite/cluster-test/src/instance.rs
```diff
@@ -271,6 +272,14 @@ impl Instance {
             )
             .await
     }
+
+    pub fn debug_interface_client(&self) -> AsyncNodeDebugClient {
+        AsyncNodeDebugClient::new(
+            self.http_client.clone(),
+            self.ip(),
+            self.debug_interface_port.unwrap_or_default() as u16,
```
andll, line 280: So returning default(which is `0`) here does not make much sense - debug client certainly won't be able to connect to port 0.
More reasonable would be to just crash here: `self.debug_interface_port.expect("debug_interface_port is not known on this instance")`

testsuite/cluster-test/src/instance.rs
```diff
@@ -271,6 +272,14 @@ impl Instance {
             )
             .await
     }
+
+    pub fn debug_interface_client(&self) -> AsyncNodeDebugClient {
```
andll, line 276: There is similar piece of code in `DebugPortLogWorker::new`:
```
let client = AsyncNodeDebugClient::new(http_client.clone(), instance.ip(), debug_interface_port);
```
Do you mind replacing it with your method there, e.g. replace this line with 
```
let client = instance.debug_interface_client();
```
To avoid code duplication.

testsuite/cluster-test/src/health/fullnode_check.rs
```diff
@@ -0,0 +1,97 @@
...
+            let content = match res {
+                Ok(res) => res.unwrap_or_default(),
+                _ => 0i64,
+            };
+            let index = node_index(v.peer_name().clone());
+            if !val_latest_version.contains_key(&index) {
+                continue;
+            }
+            let val_version = val_latest_version[&index];
+            if content - val_version > 20000_i64 || content - val_version < -20000_i64 {
```
andll, line 78: We should probably only have one check - normally full node version is always below validator version, we don't need special check for opposite case

testsuite/cluster-test/src/health/fullnode_check.rs
```diff
@@ -0,0 +1,97 @@
...
+    cluster: Cluster,
+}
+
+impl FullNodeHealthCheck {
+    pub fn new(cluster: Cluster) -> FullNodeHealthCheck {
+        Self { cluster }
+    }
+}
+
+pub fn node_index(peer_name: String) -> String {
```
andll, line 28: You can just use Instance::validator_index?

testsuite/cluster-test/src/health/fullnode_check.rs
```diff
@@ -0,0 +1,97 @@
...
+    )
+}
+
+#[async_trait]
+impl HealthCheck for FullNodeHealthCheck {
+    async fn verify(&mut self, ctx: &mut HealthCheckContext) {
+        let validator = self.cluster.validator_instances();
+        let fullnode = self.cluster.fullnode_instances();
+        let mut val_latest_version = HashMap::new();
+        for v in validator {
```
andll, line 51: One note - it would be much more effective to query different validators in parallel.
For example, this look could be replaced with something like this:
```rust
 // This is helper function to make code below simpler
async fn get_version(instance: &Instance) -> (String, i64) {
   (instance.vaidator_index(), instance.debug_interface_client(). get_node_metric("libra_state_sync_committed_version{}").await.unwrap_or(0))
}
// ....
// Then this code can be used here to query validators in parallel:
let futures = self.cluster.validator_instances().iter().map(get_version);
// Note that join_all invokes futures in parallel
let val_latest_version = join_all(futures)?; // this wll be Vec<(String, i64)> - see return type of our get_version
let val_latest_version: HashMap<_> = val_latest_version.into_iter().collect(); // this converts vec<(key, value)> into map;
```

testsuite/cluster-test/src/health/fullnode_check.rs
```diff
@@ -0,0 +1,97 @@
...
+                .await;
+            let version = match res {
+                Ok(res) => res.unwrap_or(0i64),
+                _ => 0i64,
+            };
+            if version != 0 {
+                val_latest_version.insert(node_index(v.peer_name().to_string()), version);
+            }
+        }
+        for v in fullnode {
```
andll, line 64: Similarly to above(without converting to map indeed), you can make query all full nodes in parallel and make list of tuples of `(full node, version)`.

After you got results, you can just iterate over this list to do the check - this step don't need to be parallelized

Goal is to parallelize network calls to `get_node_metric`, everything else can be done sequentially

testsuite/cluster-test/src/health/fullnode_check.rs
```diff
@@ -0,0 +1,97 @@
...
+            let res = v
+                .debug_interface_client()
+                .get_node_metric("libra_state_sync_committed_version{}")
+                .await;
+            let content = match res {
+                Ok(res) => res.unwrap_or_default(),
+                _ => 0i64,
+            };
+            let index = node_index(v.peer_name().clone());
+            if !val_latest_version.contains_key(&index) {
```
andll, line 74: You want to be strict here - if there is no validator version, then it's not good and should fail healthcheck (this is actually quite unexpected, I am not sure how this can even be the case)

testsuite/cluster-test/src/health/fullnode_check.rs
andll, line 78: Do you mind making this `20000_i64` into constant loaded from env variable?
E.g. declare it on top level as:
```
pub static THRESHOLD: Lazy<i64> = Lazy::new(|| {
if let Ok(v) = env::var("FULL_NODE_HEALTH_THRESHOLD") {
v.parse().expect("Failed to parse FULL_NODE_HEALTH_THRESHOLD");
} else {20_000}
});
```

This way we'll be able to try different values for the thershold without rebuilding, e.g. something like this:
 `./cti -E FULL_NODE_HEALTH_THRESHOLD=10000 <...>`

testsuite/cluster-test/src/health/fullnode_check.rs
snakech: I did manual test, and in 10% down test, the ```libra_state_sync_committed_version``` returned back from debug client, the values are 0 for the nodes are down.

snakech:
> Looks much shorter this time!
> 
> Left few comments - most importantly, we do want to make all network calls in parallel, we have quite a few nodes there, and querying them one-by-one can quickly become a bottleneck

good to know the way can call async fn by each instance in an array parallel.

testsuite/cluster-test/src/health/fullnode_check.rs
```diff
@@ -0,0 +1,84 @@
...
+async fn get_version(instance: &Instance) -> (String, i64) {
+    let res = instance
+        .debug_interface_client()
+        .get_node_metric("libra_state_sync_committed_version{}")
+        .await;
+    let content = match res {
+        Ok(res) => res.unwrap_or_default(),
+        _ => 0i64,
+    };
+    (instance.peer_name().clone(), content)
```
andll, line 44: I don't like that we had to change `Instance::validator_index` function to parse string returned by here.

I think, you can just return instance here and make return type `(&Instance, i64)`

testsuite/cluster-test/src/health/fullnode_check.rs
```diff
@@ -0,0 +1,84 @@
...
+
+#[async_trait]
+impl HealthCheck for FullNodeHealthCheck {
+    async fn verify(&mut self, ctx: &mut HealthCheckContext) {
+        let validator = self.cluster.validator_instances();
+        let fullnode = self.cluster.fullnode_instances();
+
+        let futures = validator.iter().map(get_version);
+        let val_latest_version = join_all(futures).await;
+        let val_latest_version: HashMap<String, i64> = val_latest_version.into_iter().collect();
```
andll, line 55: When `get_version` returns `&Instance`, you can update this line to smth like
```
let val_latest_version: HashMap<String, i64> = val_latest_version.into_iter()
   .map(|(instance, version)| (instance.validator_index(), version)).collect();
```

testsuite/cluster-test/src/health/fullnode_check.rs
```diff
@@ -0,0 +1,84 @@
...
+
+        let futures = fullnode.iter().map(get_version);
+        let fullnode_latest_version = join_all(futures).await;
+        let fullnode_latest_version: HashMap<String, i64> =
+            fullnode_latest_version.into_iter().collect();
+
+        for (peer_name, fullnode_version) in fullnode_latest_version {
+            let index = validator_index(&peer_name);
+            let val_version = val_latest_version[&format!("val-{}", index)];
```
andll, line 64: After this you going to have `instance` instead of `peer_name` here, and can use `let val_version = val_latest_version[instance.validator_index()]`

andll:
Btw, you probably noticed too - but looks like you accidentally added merge commit to this PR - you'll need to remove it and rebase off master, instead of merge

testsuite/cluster-test/src/health/fullnode_check.rs
snakech: my actual problem is we may have multi-fullnode with same index map to same validator, we have to verify all fullnode status with that one validator, but return Map<Instance, i64> is a good way to solve it

testsuite/cluster-test/src/health/fullnode_check.rs
```diff
@@ -0,0 +1,86 @@
...
+async fn get_version(instance: &Instance) -> (&Instance, i64) {
+    let res = instance
+        .debug_interface_client()
+        .get_node_metric("libra_state_sync_committed_version{}")
+        .await;
+    let content = match res {
+        Ok(res) => res.unwrap_or_default(),
+        _ => 0i64,
+    };
+    (&instance, content)
```
andll, line 45: `&` here should not be needed

testsuite/cluster-test/src/health/fullnode_check.rs
```diff
@@ -0,0 +1,86 @@
...
+            .into_iter()
+            .map(|(instance, version)| (instance.validator_index(), version))
+            .collect();
+
+        let futures = fullnode.iter().map(get_version);
+        let fullnode_latest_version = join_all(futures).await;
+
+        for (fullnode, fullnode_version) in fullnode_latest_version {
+            let index = fullnode.validator_index();
+            let val_version = val_latest_version[&index];
```
andll, line 66: Btw, I think you want to use some method like `get` here, I think `[]` might just panic if key is not found?
We want to avoid panics in code like this

testsuite/cluster-test/src/health/fullnode_check.rs
```diff
@@ -0,0 +1,86 @@
...
+            if val_version - fullnode_version > *THRESHOLD {
+                ctx.report_failure(
+                    validator[index.parse::<usize>().unwrap()]
```
andll, line 69: I think here something like `format!("val-{}", index)` will be more appropriate. I am worried about assuming that `validator` array has same order as pod index.

Also, do you mind renaming `validator`->`validators`, `fullnode`->`fullnodes` and other places in this code like val_latest_version, val_latest_version

andll:
Before merging, do you mind running `--perf-suite` to confirm it works (and also that it does not spam issues).

Additionally, if you don't mind - with new method, can you double check(using `-E FULL_NODE_HEALTH_THRESHOLD=` that smaller values, like 5000 does not work? If they do, let's lower it

Maybe previous issue was related to the prometheus lagging, rather then version difference

snakech:
> Before merging, do you mind running `--perf-suite` to confirm it works (and also that it does not spam issues in 3-region test like it did before).
> 
> Additionally, if you don't mind - with new method, can you double check(using `-E FULL_NODE_HEALTH_THRESHOLD=` that smaller values, like 5000 does not work? If they do, let's lower it. Note that `-E` should be one of the first arguments in `cti`
> 
> Maybe previous issue was related to the prometheus lagging, rather then version difference

I tried 5000 and 10000 but seems still failed, merge first and I will keep trying

PR #4592, opened 2020-06-17
[cluster test] Remove get_pod_node_and_ip

andll:
This call is now obsolete - we used it to get node name and ip, but currently we switched to pre-allocation of node, and we get all this information from KubeNode.

We had an assert just to confirm that information we retrieved with old method is same as new, and now we can remove assert and method all together


andll:
Going to look at the failure after we resolve the other issue

PR #4594, opened 2020-06-18
[move-prover] Constrained serialized address size.

kkmc:
Serialized representation of addresses are constant sized. Added a quantifer to constrain
all $Address typed values to have the serialized representations of the same vector length in the prelude. I didn't notice any slowdowns from this addition.

Also added a test case to check that serialize and to_bytes functions do
indeed satisfy this property.

## Motivation

The Event module relies on LCS::to_bytes and (the specification version of the function LCS::serialize) to return unique GUIDs. This is indeed the case if serialized addresses return the same sized vectors of u8. Confirmed with @sblackshear that this is the case.

As a side note, it looks like integer typed values of the same bit width also have the serialized representations. This is not included in the PR.
 
Yes.

## Test Plan

Existing testsuite along with the additional test: language/move-prover/tests/sources/functional/address_serialization_constant_size.move.

## Related PRs

#4531

### DIFF
--- a/language/move-prover/tests/sources/functional/address_serialization_constant_size.exp
+++ b/language/move-prover/tests/sources/functional/address_serialization_constant_size.exp
@@ -1,0 +1,15 @@
+Move prover returns: exiting with boogie verification errors
+error:  A postcondition might not hold on this return path.
+
+    ┌── tests/sources/functional/address_serialization_constant_size.move:19:9 ───
+    │
+ 19 │         ensures len(LCS::serialize(mv1)) == len(LCS::serialize(mv2));
+    │         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+    │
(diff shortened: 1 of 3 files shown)


language/move-prover/tests/sources/functional/address_serialization_constant_size.exp
```diff
@@ -0,0 +1,15 @@
+Move prover returns: exiting with boogie verification errors
+error:  A postcondition might not hold on this return path.
+
+    ┌── tests/sources/functional/address_serialization_constant_size.move:17:9 ───
+    │
+ 17 │         ensures len(LCS::serialize(u1)) == len(LCS::serialize(u2));
```
DavidLDill, line 6: If a function is SUPPOSED to get an error, suffix the name with _incorrect so people don't get worried when they see it.

The u64 example might be confusing, because (I think) u64's actually SHOULD always serialize to the same length. You just didn't include that fact in the model, so they fail here as expected.  Perhaps you should use a different type (e.g., a vector<u8>?) which really could serialize to different lengths.

kkmc: Yes, good points.

PR #4595, opened 2020-06-18
More specs for RecoveryAddress and generate docs for modules

emmazzz:
## Motivation

This PR adds more specifications to `RecoveryAddress.move`. I commented out some specs that would cause the prover to hang. Some spec functions are added to `VASP` and `LibraAccount` to help prove specs in `RecoveryAddress`.

Yes.

## Test Plan

cargo test


language/stdlib/modules/doc/LibraAccount.md
```diff
@@ -2260,3 +2261,40 @@ also be added. This account will be a child of
...
+
+<a name="0x1_LibraAccount_Specification"></a>
+
+## Specification
+
+
+Returns field
+<code>key_rotation_capability</code> of the
+LibraAccount under
+<code>addr</code>.
```
mimoo, line 2273: you don't need to mix html with markdown this way (https://daringfireball.net/projects/markdown/syntax)

wrwg: You are welcome to rewrite docgen as long as you preserve its functionality. There are endless constraints you need to consider. In this example, {backquote} can't be used because we are bolding move keywords and adding links to definitions, which doesn't work inside markdown code blocks. 

It's generally OK to use html in markdown and is common practice, as you find documented in the countless markdown specs on the web.

mimoo: not sure what docgen is, I thought this was an issue with a lack of markdown knowledge :)

it is fine to mix html and markdown, but heavy use of html instead of markdown elements reduces the benefits of markdown.

wrwg: This file is autogenerated from the Move sources. I'll remember to put a disclaimer in the output to avoid such confusion.

language/stdlib/modules/LibraAccount.move
```diff
@@ -891,5 +891,25 @@ module LibraAccount {
...
     // End of the proof of concept code
     ///////////////////////////////////////////////////////////////////////////
+
+    // ****************** SPECIFICATIONS *******************
+
+    spec module {
+        /// Returns field `key_rotation_capability` of the
+        /// LibraAccount under `addr`.
+        define spec_get_key_rotation_cap(addr: address):
+            Option<KeyRotationCapability>
```
wrwg, line 901: Put on same line? Also consistent use of `{`. Next function has it on same line. I think this is the preferred style.

language/stdlib/modules/LibraAccount.move
```diff
@@ -891,5 +891,25 @@ module LibraAccount {
...
+        {
+            global<LibraAccount>(addr).key_rotation_capability
+        }
+
+        /// Returns true if the LibraAccount at `addr` holds
+        /// `KeyRotationCapability` for itself.
+        define spec_holds_own_key_rotation_cap(addr: address): bool {
+            Option::is_some_spec(spec_get_key_rotation_cap(addr))
```
wrwg, line 909: Consistent of use of spec_ or _spec

language/stdlib/modules/RecoveryAddress.move
```diff
@@ -56,14 +59,29 @@ module RecoveryAddress {
...
+            // spec {
+            //     assert forall j in 0..i: caps[j].account_address != to_recover;
+            // };
             let cap = Vector::borrow(caps, i);
             if (LibraAccount::key_rotation_capability_address(cap) == &to_recover) {
                 LibraAccount::rotate_authentication_key(cap, new_key);
                 return
             };
             i = i + 1
         };
```
wrwg, line 80: Why is the spec block commented out? What means the `// (i < len)`? Please add a TODO if this is draft. Imagine someone wants to read this code and does not care much about spec.

We need to have a convention to have those loop invariant stuff. It makes code pretty unreadable as of now. @shazqadeer

emmazzz: Oops I forgot to delete those lines. 
And yes I agree we should rethink how we write loop invariants. Right now it's hard to find the actual termination condition among all the spec code.

PR #4596, opened 2020-06-18
[libra-node] Remove duplicate setup_panic_handler

wqfish:
I think we are calling `setup_panic_handler` twice. We already call this function at the beginning of `main`:

https://github.com/libra/libra/blob/f4982909c6e0b04752c391dfe25c3eb093164873/libra-node/src/main.rs#L37

so there is no need to call it again in `setup_environment`?


wqfish:
@davidiw I was setting up panic handler in another project and trying to copy some code from here when I noticed it.

davidiw:
> @davidiw I was setting up panic handler in another project and trying to copy some code from here when I noticed it.

I was going to say... I might start asking you to audit more of the code if you can just find these random ... snippets :)

PR #4597, opened 2020-06-18
[cluster test] Honor cooldown during downscaling

andll:
This should help to prevent spurious 'pod not found' issue we've seen
It should also allow to reduce run time of cluster test in 'peak hours', as instances would not downscale


ankushagarwal:
I'm not sure if this is what we want. With this flag, the scale down or the scale up action can be ignored if we are within the cooldown period.

andll:
Yeah, looks like cooldown does not work as expected and just returns
```
An error occurred (ScalingActivityInProgress) when calling the SetDesiredCapacity operation: Scaling activity 0685cb16-898e-25fe-d43c-fc9f97233857 is in progress and blocks this action
```

PR #4599, opened 2020-06-18
[management] Add basic json-rpc client

davidiw:
To better support the ecosystem, it makes sense to consolidate
management related components into management. This means we'll need the
ability to at least submit and craft transactions. This right now is
intended to be used for emergency situations, where the admin will craft
the transaction on the fly and then submit it. In the future, this
framework will enable generating and submitting any transaction relevant
to the entities the system supports.

Why not re-use the existing cli / client? For this tool, we really don't
need something so heavy. This is intended to be used in a trusted
environment and quick to use. The other tool has a lot of different
features and would require substantial effort to integrate into this
framework. That would also require two tools for handling this work.

Perhaps one day we can explore a unification, but not now, not now.

### DIFF
--- a/config/management/src/lib.rs
+++ b/config/management/src/lib.rs
@@ -5,6 +5,7 @@
 
 mod error;
 mod genesis;
+mod json_rpc;
 mod key;
 mod layout;
 mod secure_backend;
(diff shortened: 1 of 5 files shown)


config/management/src/lib.rs
```diff
@@ -113,7 +124,13 @@ impl Command {
             Command::InsertWaypoint(_) => self.insert_waypoint().unwrap().to_string(),
             Command::OperatorKey(_) => self.operator_key().unwrap().to_string(),
             Command::OwnerKey(_) => self.owner_key().unwrap().to_string(),
+            Command::ReadAccountState(_) => format!("{:?}", self.read_account_state().unwrap()),
             Command::SetLayout(_) => self.set_layout().unwrap().to_string(),
+            Command::SubmitTransaction(_) => self
+                .submit_transaction()
```
JoshLind, line 130: Do you think it makes sense to have "submit_transaction()" return the Transaction that was submitted? This might also be useful for debugging/testing (i.e., the transaction is assumed to be in file, in lcs encoded format). Printing out the transaction that was submitted might provide the user with some assurances about what was submitted? It would also avoid having to define an arbitrary "success" message 😄

davidiw: Hmm... seems an interesting proposal... we actually need to distinguish verbosity here... something to think about..

Yes, I did test this lol on pre-mainnet -- the one thing missing is the ability to query the outcome of a transaction

PR #4600, opened 2020-06-18
[Move ABIs] generate ABI files and use them for good

matbd:
##  Motivation

* Create a new tool `abigen` inspired by `docgen` in the Move prover.
* Use it in `language/stdlib` to generate ABI files (at first in JSON).
* (optional) Use the ABI files as source of truth in compiled-stdlib (this makes doc and signatures visible -- also enables e2e test to test naming consistency, which happened to reveal a bug in `update_unhosted_wallet_limits`)
* Modify the tool `transaction-builder-generator` to use the ABI files on disk (instead of curated list in `transaction-builder`) for code generation of transaction builders (e.g in python)
* Refactor `transaction-builder` to use generated code instead of macros (this means we can remove the manually curated list of scripts there). For now, I use aliases to avoid breaking users.
* Fix a bug in stdlib generation where we never removed any files!! (I ran the tool next and removed many of them)
* (optional) Switch ABI files from JSON to LCS at the very end (we trust LCS more and it's faster to parse)

Future work:
* Remove compatibility aliases to make naming more consistent. (Also, probably rename `peer_to_peer_with_metadata` into something more meaningful everywhere.)
* One possible drawback of removing the curated list is that the alphabetical ordering of script builders looks a bit random in Rust docs. It would be nice to have a way to group generated transaction builders by categories (or anything to make `cargo doc -p transaction-builder` nicer really).
* Also, the move (parser/prover) framework doesn't remove the whitespace `/// <- Here` so I had to normalize whitespaces (thus removing any manual formatting).
* I think we should really stop singling out particular Move scripts in the codebase (e.g. in JSON RPC) and just have people use "official" transaction builders. In this respect, we may wish to discourage people from using `compiled-stdlib` directly?

## Test Plan

included unit tests


sblackshear:
> Refactor transaction-builder to use generated code instead of macros (this means we can remove the manually curated list of scripts there)

I'm been waiting for this day for so long!

sblackshear:
> By the way do we prefer transfer_with_metadata or peer_to_peer_with_metadata

I very much prefer `transfer_with_metadata` or even just `transfer` or `payment`, but have been wary of changing the name because (I think) JSON-rpc does some script lookups by name. But fully support changing it; pretty much anything is better than the current name!

sblackshear:
> I think we should really stop singling out particular Move scripts in the codebase (e.g. in JSON RPC) and just have people use "official" transaction builders. In this respect, we may wish to discourage people from using compiled-stdlib directly?

Ah, I think this would address the problem I was mentioning. Strongly support!

matbd:
> > By the way do we prefer transfer_with_metadata or peer_to_peer_with_metadata
> 
> I very much prefer `transfer_with_metadata` or even just `transfer` or `payment`, but have been wary of changing the name because (I think) JSON-rpc does some script lookups by name. But fully support changing it; pretty much anything is better than the current name!

Now that I know all the differences, I'm thinking I should temporarily keep the old names alive (by manually adding entries) to make this one PR non-breaking

sblackshear:
That doesn't sound like a bad idea

thefallentree:
We should make an enum , and add an script to enum function in transaction builder, and use  enums in jsonrpc view.rs

We should keep strong name in view.rs , since that is part of jsonrpc api

Yucong Sun, Libra PE

matbd:
> We should make an enum , and add an script to enum function in transaction builder, and use enums in jsonrpc view.rs We should keep strong name in view.rs , since that is part of jsonrpc api Yucong Sun, Libra PE

@thefallentree It sounds like what you are describing is creating an AST on top of the "transaction builders" then having an interpretation function.

Concretely, we provide this:
```
fn encode_a_script(u64) -> Script;
fn encode_b_script(u64) -> Script;
```

and it sounds like you would like this:
```
enum RPCCall {
    CallA(u64),
    CallB(u32),
}

fn make_script(rpc: &RPC) -> Script { ... }
```

If `make_script` could live server-side, it could be interesting to see `RPCCall` as data traveling in a JSON RPC API, but unfortunately, because of crypto signatures, what really travels is the `Script` object.

So while we cannot really prevent a user to create such an enum RPCCall on the client side, for us, it would only be a burden for us on the server-side. In particular, any change in the available Move scripts (recall that this set is managed onchain) would require a restart of the JSON-RPC server.

thefallentree:
> > We should make an enum , and add an script to enum function in transaction builder, and use enums in jsonrpc view.rs We should keep strong name in view.rs , since that is part of jsonrpc api Yucong Sun, Libra PE
> 
> @thefallentree It sounds like what you are describing is creating an AST on top of the "transaction builders" then having an interpretation function.
> 
> Concretely, we provide this:
> 
> ```
> fn encode_a_script(u64) -> Script;
> fn encode_b_script(u64) -> Script;
> ```
> 
> and it sounds like you would like this:
> 
> ```
> enum RPCCall {
>     CallA(u64),
>     CallB(u32),
> }
> 
> fn make_script(rpc: &RPC) -> Script { ... }
> ```
> 
> If `make_script` could live server-side, it could be interesting to see `RPCCall` as data traveling in a JSON RPC API, but unfortunately, because of crypto signatures, what really travels is the `Script` object.
> 
> So while we cannot really prevent a user from creating such an enum `RPCCall` on the client side, it does not simplify anything. In general, for us, making the JSON RPC API exposes each of the Move scripts individually is essentially only a burden on the server-side. In particular, in this setup, any change in the available Move scripts (recall that this set is managed onchain) requires a restart of the JSON-RPC server.

actually, not this complicated, I really meant that we should add an enum in transaction-builder package  , give each script an enum name,  that way we can avoid relying on internal names for these transactions.

matbd:
> actually, not this complicated, I really meant that we should add an enum in transaction-builder package , give each script an enum name, that way we can avoid relying on internal names for these transactions.

We have such an enum in `compiled-stdlib` but I'm not sure what it buys us -- once we have proper "transaction builders" in place. In fact, I would suggest that we avoid using `compiled-stdlib` from outside Move.

Are you worried about unintentional breaking changes due to changes in Move names?
Note that the result of code generation is checked on disk, so breaking changes are visible (if not already unit-tested).

wrwg:
> [move-prover] Move Prover test (`language/move-prover/cargo test`) on commit [b9acfdd](https://github.com/libra/libra/commit/b9acfdd3238e881588990562586aa5e859167cac) did **fail**. See [details](https://github.com/mimoo/mirai-bot/runs/786460479?check_suite_focus=true#step:7:1).

I think you where hit by one of this race conditions stemming from that we do not have set up move prover tests in CI but use this bot here. This bot runs on the commit not the merge with head. Its therefore possible that two PRs pass the bot when merged together don't longer.

To fix, please see ./move-prover/scripts installation instructions if you haven't running this yet, and also set env vars there. Then run `UPBL=1 cargo test -p move-prover address_serialization`

thefallentree:
can you add an .gitattributes file to treat ABI file as text?

matbd:
> can you add an .gitattributes file to treat ABI file as text?
> 
> if they are yml files we should just call it .abi.yml, what is the reason to pick LCS vs json/yaml?

I don't think YAML will work. JSON works now but these data-structures are part of a larger set of definitions which use LCS. So, LCS just feels like a more natural choice to me. Also, it turns out that an `.abi` file must include the bytes of the corresponding `.mv` file, so JSON is a bit verbose (and possibly slower if used as source of truth in `compiled-stdlib`).

I'm thinking that these ABI files may not be widely distributed in the future anyway because script authors can just make generated transaction builders directly available in various languages.

matbd:
> > [move-prover] Move Prover test (`language/move-prover/cargo test`) on commit [b9acfdd](https://github.com/libra/libra/commit/b9acfdd3238e881588990562586aa5e859167cac) did **fail**. See [details](https://github.com/mimoo/mirai-bot/runs/786460479?check_suite_focus=true#step:7:1).
> 
> I think you where hit by one of this race conditions stemming from that we do not have set up move prover tests in CI but use this bot here. This bot runs on the commit not the merge with head. Its therefore possible that two PRs pass the bot when merged together don't longer.
> 
> To fix, please see ./move-prover/scripts installation instructions if you haven't running this yet, and also set env vars there. Then run `UPBL=1 cargo test -p move-prover address_serialization`

The command `cargo test -p move-prover address_serialization` actually succeeds on my laptop. Let see if this happens again.

wrwg:
> > > [move-prover] Move Prover test (`language/move-prover/cargo test`) on commit [b9acfdd](https://github.com/libra/libra/commit/b9acfdd3238e881588990562586aa5e859167cac) did **fail**. See [details](https://github.com/mimoo/mirai-bot/runs/786460479?check_suite_focus=true#step:7:1).
> > 
> > 
> > I think you where hit by one of this race conditions stemming from that we do not have set up move prover tests in CI but use this bot here. This bot runs on the commit not the merge with head. Its therefore possible that two PRs pass the bot when merged together don't longer.
> > To fix, please see ./move-prover/scripts installation instructions if you haven't running this yet, and also set env vars there. Then run `UPBL=1 cargo test -p move-prover address_serialization`
> 
> The command `cargo test -p move-prover address_serialization` actually succeeds on my laptop. Let's see if this happens again.

If you do not have set up the prover tools and configured env variables to point to them, the prover tests will pass. But they will fail in the MIRAI bot.

thefallentree:
> > can you add an .gitattributes file to treat ABI file as text?
> > if they are yml files we should just call it .abi.yml, what is the reason to pick LCS vs json/yaml?
> 
> I don't think YAML will work. JSON works now but these data-structures are part of a larger set of definitions which use LCS. So, LCS just feels like a more natural choice to me. Also, it turns out that an `.abi` file must include the bytes of the corresponding `.mv` file, so JSON is a bit verbose (and possibly slower if used as source of truth in `compiled-stdlib`).
> 
> I'm thinking that these ABI files may not be widely distributed in the future anyway because script authors can just make generated transaction builders directly available in various languages.

hmm, I see what you mean, so what is the point of checking them in?  If we want to use it in unitest to detect the change it seems better to include generated rust code and that will show changes better.

matbd:
> If you do not have set up the prover tools and configured env variables to point to them, the prover tests will pass. But they will fail in the MIRAI bot.

Gotcha, thanks! I added a fix at the end of this PR.

matbd:
> hmm, I see what you mean, so what is the point of checking them in? If we want to use it in unitest to detect the change it seems better to include generated rust code and that will show changes better.

So technically, the "transaction builders" can be generated in three languages Python, Rust, and "Local" (naming TBD). "Local" means Rust within the Libra repo (as opposed to dependency-free Rust using Serde codegen for Libra types). Currently, "Local" must be checked in because it is used in JSON RPC server, unit tested in e2e tests, used as a Rust Libra client, etc. I feel this is a rather special case though (and possibly only temporarily ok while we develop Move scripts in the same repo as Libra Core and ship them within the genesis block).

thefallentree:
> > hmm, I see what you mean, so what is the point of checking them in? If we want to use it in unitest to detect the change it seems better to include generated rust code and that will show changes better.
> 
> So technically, the "transaction builders" can be generated in three languages Python, Rust, and "Local" (naming TBD). "Local" means Rust within the Libra repo (as opposed to dependency-free Rust using Serde codegen for Libra types). Currently, "Local" must be checked in because it is used in JSON RPC server, unit tested in e2e tests, used as a Rust Libra client, etc. I feel this is a rather special case though (and possibly only temporarily ok while we develop Move scripts in the same repo as Libra Core and ship them within the genesis block).

could we migrate the various part that uses "local"  to use the new "rust" generated code? is that an big migration effort? (looks like we would just need an transaction-builder-builder and an  build.rs in transaction-builder for this?)

matbd:
> could we migrate the various part that uses "local" to use the new "rust" generated code? is that an big migration effort? (looks like we would just need an transaction-builder-builder and an build.rs in transaction-builder for this?)

Definitely not in this PR because many operations would be missing (e.g. crypto).

matbd:
Yay, I just created an (unsigned) P2P transaction in pure python on top of this PR!

https://github.com/novifinancial/libra-client-dev/pull/99

language/move-prover/abigen/Cargo.toml
```diff
@@ -0,0 +1,31 @@
+[package]
```
tzakian, line 1: Not sure if there are crate restrictions at play here (I can't see any?), but it seems a bit odd to have `abigen` under move-prover. Could this instead be a crate in the top level `language`, or possibly a crate under `stdlib`?

language/move-prover/abigen/src/abigen.rs
```diff
@@ -0,0 +1,189 @@
...
+    /// Options.
+    options: &'env AbigenOptions,
+    /// Input definitions.
+    env: &'env GlobalEnv,
+    /// Map from file name to generated script ABI (if any).
+    output: BTreeMap<String, ScriptABI>,
+}
+
+impl<'env> Abigen<'env> {
+    /// Creates a new documentation generator.
```
tzakian, line 47: Update/remove comment?

language/move-prover/abigen/src/abigen.rs
```diff
@@ -0,0 +1,189 @@
...
+        std::mem::take(&mut self.output)
+            .into_iter()
+            .map(|(path, abi)| {
+                let content = lcs::to_bytes(&abi).expect("ABI serialization should not fail");
+                (path, content)
+            })
+            .collect()
+    }
+
+    /// Generates ABIs for all script modules in the environment (exclusing the dependency set).
```
tzakian, line 68:
```suggestion
    /// Generates ABIs for all script modules in the environment (excluding the dependency set).
```

language/move-prover/abigen/src/abigen.rs
```diff
@@ -0,0 +1,189 @@
...
+            None => bail!("A script module should define a function."),
+        };
+        let name = symbol_pool.string(func.get_name()).to_string();
+        let doc = func.get_doc().to_string();
+        let code = self.load_compiled_bytes(&module_env)?.to_vec();
+        let ty_args = func
+            .get_named_type_parameters()
+            .iter()
+            .map(|ty_param| {
+                TypeArgumentABI::new(symbol_pool.string(ty_param.0).to_string().to_lowercase())
```
tzakian, line 108: I don't think this is a major issue, but if we have two types `Cointype` and `CoinType` they'd both be mapped to `cointype` in this transformation, while also being viewed as different type parameters in Move. I doubt we'd encounter it normally, but wondering if going to snake case (converting upper to `_lower`) might be better? (e.g. the two examples would be converted to `coin_type` and `cointype`)

language/move-prover/abigen/src/abigen.rs
```diff
@@ -0,0 +1,190 @@
...
+        };
+        Ok(Some(tag))
+    }
+
+    fn get_type_tag(&self, ty: &ty::Type) -> anyhow::Result<TypeTag> {
+        if let Some(tag) = self.get_type_tag_skipping_references(ty)? {
+            return Ok(tag);
+        }
+        bail!(
+            "References such as {:?} are only not allowed in the list of parameters.",
```
tzakian, line 185: Nit: Typo? I don't quite follow the meaning of "only not allowed".

language/move-prover/src/cli.rs
```diff
@@ -50,6 +51,8 @@ pub struct Options {
     pub verbosity_level: LevelFilter,
     /// Whether to run the documentation generator instead of the prover.
     pub run_docgen: bool,
+    /// Whether to run the ABI generator instead of the prover.
+    pub run_abigen: bool,
```
tzakian, line 55: Hmm this is making me think more and more that this should be a part of the `stdlib` crate -- we'll probably want to specify there whether or not a new set of ABIs should be generated as opposed to in the prover.

language/move-prover/abigen/Cargo.toml
matbd: I followed the example of `docgen` (which is also called by stdlib). We may want to figure out a better place for both crates.

tzakian: Agreed. I don't view this placement as a blocker for this PR -- better to get it in and being used and then we can move it around if need be.

language/move-prover/abigen/src/abigen.rs
matbd: I agree snake case would be better. Let me see if I can find a library for this.

matbd: For name collisions, I need to think about it more. I don't know if we can realistically achieve a truly non-failing (aka "monkey proof") code gen.

language/move-prover/abigen/src/abigen.rs
matbd: yes, it should be "only allowed"

language/move-prover/src/cli.rs
matbd: You guys convinced me to do it like this in the thread :)
More seriously, abigen and docgen are quite similar (both called by the stdlib tool, both generate files) so it seems to me that there is broader discussion to have separately.

language/stdlib/compiled/src/transaction_scripts.rs
```diff
@@ -263,4 +267,12 @@ mod test {
             }
         );
     }
+
+    #[test]
+    fn test_names() {
+        // Make sure that the names listed here matches the function names in the code.
+        for script in StdlibScript::all() {
+            assert_eq!(script.name(), script.abi().name());
```
tzakian, line 275: <3

language/transaction-builder-generator/src/rust.rs
```diff
@@ -159,3 +162,49 @@ fn make_transaction_argument(type_tag: &TypeTag, name: &str) -> String {
...
+        let mut cargo = std::fs::File::create(&dir_path.join("Cargo.toml"))?;
+        write!(
+            cargo,
+            r#"[package]
+name = "{}"
+version = "0.1.0"
+edition = "2018"
+
+[dependencies]
+serde = {{ version = "1.0", features = ["derive"] }}
```
tzakian, line 199: Not at all crucial for this PR, but might be worth thinking about eventually: If there's a good way to handle dependency bumps for external dependencies without having to manually change the generated code here.

language/move-prover/abigen/src/abigen.rs
tzakian: Agreed on the "monkey proof" code gen. I trust your judgement on this though, so I'm interested to see what you come up with! :)

language/move-prover/src/cli.rs
tzakian: ahhh sorry about that! I absolutely agree that this is a broader discussion in terms of crate placement, and the placement of this is absolutely _not_ a blocker for this PR.

language/transaction-builder-generator/src/rust.rs
matbd: Yes, that's a good point. I need to track this somewhere.

PR #4603, opened 2020-06-18
[move-prover] Generalize smoke test feature & custom error messages.

wrwg:
This PR generalizes the smoke test feature and re-implements the existing aborts_if_true smoke test as a transformation.

- There is a new processor `stackless-bytecode-generator/src/test_instrumenter.rs`. Moving forward, we should be able to add more tests of this kind to this file without needing to change much anywhere else. This includes tests which rewrite specs, add specs, or rewrite bytecode.
- The feature is based on some new information about spec conditions added to the environment. The ConditionInfo type allows to specify the error message to print if a condition fails and whether a condition is  negative. A negative condition is expected to fail for the test to pass. The boogie wrapper deals with reporting errors if negative conditions succeed.
- The new ConditionInfo has been also used to streamline error messages for regular condition failures (e.g. if an aborts_if fails, it not longer prints "a post condition might not hold", but a more concrete error). This lead to lots of updates on baseline files.

## Motivation

Spec testing (Kevin's intern project)

Yes

## Test Plan

Existing tests

## Related PRs

NA


language/move-prover/spec-lang/src/env.rs
```diff
@@ -70,7 +71,7 @@ pub const ABORTS_IF_IS_STRICT_PRAGMA: &str = "aborts_if_is_strict";
 pub const REQUIRES_IF_ABORTS: &str = "requires_if_aborts";
 
 /// Pragma indicating that the function will run smoke tests
-pub const SMOKE_TEST_PRAGMA: &str = "smoke_test";
+pub const ALWAYS_ABORTS_TEST_PRAGMA: &str = "always_aborts_test";
```
DavidLDill, line 74: Eventually, we should come up with a better name than "smoke test"

wrwg: This PR does this in a way, the pragma is now called "always_aborts_test". The code calls the thing now "test instrumentation".

PR #4606, opened 2020-06-18
[cluster_test] Fix 3-region simulation

andll:
Root cause is having `sudo` in network effects cmd line.

Relied on land blocking push to catch the issue, but it does not test 3-region simulation, so the problem slipped in master

There was also a typo in `wait_job_completion` so failed job was treated as `still in progress`

### DIFF
--- a/testsuite/cluster-test/src/main.rs
+++ b/testsuite/cluster-test/src/main.rs
@@ -763,7 +763,7 @@ impl ClusterTestRunner {
 
     async fn wait_until_all_healthy(&mut self) -> Result<()> {
         info!("Waiting for all validators to be healthy");
-        let wait_deadline = Instant::now() + Duration::from_secs(20 * 60);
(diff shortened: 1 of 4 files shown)


testsuite/cluster-test/src/main.rs
```diff
@@ -763,7 +763,7 @@ impl ClusterTestRunner {
 
     async fn wait_until_all_healthy(&mut self) -> Result<()> {
         info!("Waiting for all validators to be healthy");
-        let wait_deadline = Instant::now() + Duration::from_secs(20 * 60);
+        let wait_deadline = Instant::now() + Duration::from_secs(5 * 60);
```
snakech, line 766: just concern the shorten duration is enough for node recovery?

andll: Yeah, from all practical purposes recovery happens much faster than 5 minutes.

andll:
@bors-libra r=snakech p=100
Fixes nightly build

PR #4611, opened 2020-06-18
[Management] Add set-operator command.

JoshLind:
## Motivation

This PR adds support for the "set-operator" command to the management tool. "set-operator" allows a validator owner to set the operator responsible for operating the validator node. 

In this PR, we provide an MVP solution, with this flow:
1. Each owner will run "set-operator" specifying the name of the operator (as defined in the shared repo -- e.g., namespace), and the shared repo backend.
2. The management tool will then take the operator name, fetch the corresponding operator public key (as published in the shared repo) and upload the operator public key to the owner namespace.
3. Each operator will then run "validator-config" to generate a config for the owner validator and upload this config to the owner namespace.
3. When creating genesis, the management tool will go through each owner namespace, fetch the corresponding operator public keys and validator configs published in the owner namespace, and use these to generate genesis.

This PR provides several commits:
1.  A simple commit that adds missing availability() checks to storage.
2. A commit that supports the set-operator() command and updates the smoke test to handle the separation between the operator and owner namespaces. To achieve this, some parts of the smoke test were refactored.
3. A refactor commit that refactors the way we create storage instances from secure backend configs. This avoids us missing availability() checks as was highlighted by the need for commit 1 above.

While this PR provides an MVP solution, there's a number of things still left to improve/fix, e.g.,:
- Verifying signatures over assignments (i.e., verify that each owner has indeed assigned the operator written to their namespace -- perhaps using a signature or transaction).
- Verify the owner address specified in the validator config is signed and correct.
- Provide on-chain support for the assignment of operators to owners (e.g., after genesis has been created and a new operator is assigned)

There are also a few refactors that need to be performed to clean up the management tool/code. These will follow in another PR.

Yes.

## Test Plan

All local tests pass, including the newly modified smoke test.

## Related PRs

None. But this relates to the first item in the issue: https://github.com/libra/libra/issues/4378

### DIFF
--- a/config/management/src/genesis.rs
+++ b/config/management/src/genesis.rs
@@ -55,7 +55,11 @@ impl Genesis {
         association_config
             .parameters
             .insert("namespace".into(), layout.association[0].clone());
+
         let association: Storage = association_config.try_into()?;
+        association
+            .available()
+            .map_err(|e| Error::RemoteStorageUnavailable(e.to_string()))?;
 
         let association_key = association
             .get(ASSOCIATION_KEY)
@@ -72,7 +76,11 @@ impl Genesis {
         common_config
             .parameters
             .insert("namespace".into(), constants::COMMON_NS.into());
+
         let common: Storage = common_config.try_into()?;
+        common
+            .available()
+            .map_err(|e| Error::RemoteStorageUnavailable(e.to_string()))?;
 
         let layout = common
             .get(constants::LAYOUT)
@@ -90,7 +98,11 @@ impl Genesis {
             validator_config
                 .parameters
                 .insert("namespace".into(), operator.into());
+
             let validator: Storage = validator_config.try_into()?;
+            validator
+                .available()
+                .map_err(|e| Error::RemoteStorageUnavailable(e.to_string()))?;
 
             let key = validator
                 .get(OPERATOR_KEY)
--- a/config/management/src/genesis.rs
+++ b/config/management/src/genesis.rs
@@ -91,41 +91,43 @@ impl Genesis {
     }
 
     /// Produces a set of ValidatorRegistration from the remote storage.
+    /// TODO(joshlind): verify that owner account address specified by the validator config matches
+    /// the owner account address that was uploaded to the remote storage. Also verify that the
+    /// operator selection has been signed by the owner (somehow...)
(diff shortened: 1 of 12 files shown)


config/management/src/genesis.rs
```diff
@@ -91,41 +91,43 @@ impl Genesis {
     }
 
     /// Produces a set of ValidatorRegistration from the remote storage.
+    /// TODO(joshlind): verify that owner account address specified by the validator config matches
+    /// the owner account address that was uploaded to the remote storage. Also verify that the
+    /// operator selection has been signed by the owner (somehow...)
```
davidiw, line 91: eeeh, let's not add this verify statement on operator selection, we'll evaluate if it is practical or not

config/management/src/genesis.rs
```diff
@@ -91,41 +91,43 @@ impl Genesis {
     }
 
     /// Produces a set of ValidatorRegistration from the remote storage.
+    /// TODO(joshlind): verify that owner account address specified by the validator config matches
```
davidiw, line 89: in the future we'll use name resolution so not a big deal, right?

config/management/src/genesis.rs
```diff
@@ -91,41 +91,43 @@ impl Genesis {
...
+                .and_then(|v| v.value.transaction())
                 .map_err(|e| {
                     Error::RemoteStorageReadError(constants::VALIDATOR_CONFIG, e.to_string())
-                })?
-                .value;
-            let txn = txn.transaction().unwrap();
-            let txn = txn.as_signed_user_txn().unwrap().payload();
-            let txn = if let TransactionPayload::Script(script) = txn {
-                script.clone()
+                })?;
```
davidiw, line 112: this isn't quite right...

I'll have to look at more of the code, but I'm not reading what I expected...

create owner account
create operator account
owner selects operator
operator sets validator config

instead what I'm seeing is we have the same setup as before just owners addresses instead of operators addresses?

davidiw: I think you could refactor this a little into the following commits:
* Create validator accounts for owners
* Set the validator operator (note the cli portion here is fine)
* Use the same txn for the operator to set the owner config
* Remove the operator config and port everyone to use the owner account instead

davidiw:
I think you're refactoring this PR into other PRs? Feel free to reopen...

PR #4612, opened 2020-06-18
[cluster test] Convert HealthCheck implementation fn verify to Async

snakech:
since prometheus is flaky, we decided not rely on query metic from prometheus, instead of querying debug client to get metric with AsyncNodeDebugClient, we need to convert implementation of fn verify to Async

## Motivation

(Write your answer here.)

## Test Plan

## Related PRs

### DIFF
--- a/testsuite/cluster-test/src/health/mod.rs
+++ b/testsuite/cluster-test/src/health/mod.rs
@@ -10,6 +10,7 @@ mod log_tail;
 
 use crate::{cluster::Cluster, util::unix_timestamp_now};
 use anyhow::{bail, Result};
+use async_trait::async_trait;
 pub use commit_check::CommitHistoryHealthCheck;
 pub use debug_interface_log_tail::DebugPortLogWorker;
 use itertools::Itertools;
@@ -57,11 +58,12 @@ impl fmt::Debug for ValidatorEvent {
     }
 }
 
-pub trait HealthCheck {
+#[async_trait]
+pub trait HealthCheck: Send {
     /// Verify specific event
     fn on_event(&mut self, event: &ValidatorEvent, ctx: &mut HealthCheckContext);
     /// Periodic verification (happens even if when no events produced)
-    fn verify(&mut self, _ctx: &mut HealthCheckContext) {}
+    async fn verify(&mut self, _ctx: &mut HealthCheckContext);
     /// Optionally marks validator as failed, requiring waiting for at least one event from it to
     /// mark it as healthy again
     fn invalidate(&mut self, _validator: &str) {}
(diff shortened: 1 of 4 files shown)


andll:
@bors-libra r+

Thank you!

testsuite/cluster-test/src/health/mod.rs
```diff
@@ -57,11 +58,12 @@ impl fmt::Debug for ValidatorEvent {
     }
 }
 
+#[async_trait]
 pub trait HealthCheck {
     /// Verify specific event
     fn on_event(&mut self, event: &ValidatorEvent, ctx: &mut HealthCheckContext);
     /// Periodic verification (happens even if when no events produced)
-    fn verify(&mut self, _ctx: &mut HealthCheckContext) {}
+    async fn verify(&mut self, _ctx: &mut HealthCheckContext);
```
snakech, line 66: I changed verify back to declaration is because of the known warning of Rust if we use default implementation:
```
warning: the trait `health::HealthCheck` cannot be made into an object
  --> testsuite/cluster-test/src/health/mod.rs:66:14
   |
62 | pub trait HealthCheck {
   |           ----------- this trait cannot be made into an object...
...
66 |     async fn verify(&mut self, _ctx: &mut HealthCheckContext) {}
   |              ^^^^^^ ...because method `verify` references the `Self` type in its `where` clause
```
Refer to:
[https://github.com/rust-lang/rust/issues/51443](url)
[https://doc.rust-lang.org/error-index.html](url) Chapter -> Method references the Self type in its parameters or return type

andll: ok :)

testsuite/cluster-test/src/health/mod.rs
```diff
@@ -74,12 +76,12 @@ pub trait HealthCheck {
 
 pub struct HealthCheckRunner {
     cluster: Cluster,
-    health_checks: Vec<Box<dyn HealthCheck>>,
+    health_checks: Vec<Box<dyn HealthCheck + Send>>,
```
andll, line 79: Oh, one note

You can just make `HealthCheck: Send` in declaration of the HealthCheck, rather then adding `+ Send` everywhere

andll:
@bors-libra r-

Giving chance to fix the comment, feel free to use `r=andll` after that

testsuite/cluster-test/src/health/mod.rs
snakech: ok, I will update in the next PR

andll: You can update it here and use `r=andll` after to merge it

Issue #4355, opened 2020-06-08
[Bug] [move-prover] Type-checking and using len(v) in the specification language.


kkmc:
# 🐛 Bug

In verifying the vector reverse function of the Vector module, we encountered issues in using len(v) inside of the specification expressions.

Referring to len(v) outside of a quantifier range (e.g. first argument to all and any) raises an error.

Also, a local argument whose name is len may conflict with len(v) inside the Boogie models. Refer to https://github.com/libra/libra/pull/4319/commits/3fcb7b44b54924214cb1bb50ad4701515e6fe70f#diff-d85da9b8b2c51c1651760f69f1eff8b4R151.

## To reproduce

Removing the comment for line 135 in tests/sources/functional/verify_vector.move (https://github.com/libra/libra/pull/4319/commits/3fcb7b44b54924214cb1bb50ad4701515e6fe70f#diff-d85da9b8b2c51c1651760f69f1eff8b4R135
) and keep the comments for every other line in the spec block to reveal the error below.

Now comment out line 135 and remove the comment for line 138  (https://github.com/libra/libra/pull/4319/commits/3fcb7b44b54924214cb1bb50ad4701515e6fe70f#diff-d85da9b8b2c51c1651760f69f1eff8b4R138), the type checker will no longer raise this error (which is not consistent with the previous error).

**Stack trace/error message**
```
error: expected `|vector<#0>|?1` but found `u64` in expression

     ┌── tests/sources/functional/verify_vector.move:123:32 ───
     │
 123 │                 assert (len == len(v));
     │                                ^^^^^^
     │
```

## Expected Behavior

Expected no errors when using a local variable `len` (issue with variable renaming?) and using len(v) anywhere inside of specification expressions (not just the quantifier range).

wrwg:
This is by design. To access a shadowed global name, one can use `::len(x)`

PR #4613, opened 2020-06-18
[move-prover] added support for adding trusted loop invariants

shazqadeer:
## Motivation

This PR adds trusted loop invariants into the generated Boogie code.  Some of the tests that had been disabled earlier because of lack of invariants have been enabled again.  I also added annotations for Vector::reverse, work that was earlier done by @kkmc, and uncommented some of the postconditions on Vector::remove that can now be verified.

Yes.

## Test Plan

Existing tests and couple more tests in verify_vector.move.

### DIFF
--- a/language/move-prover/tests/sources/functional/verify_vector.move
+++ b/language/move-prover/tests/sources/functional/verify_vector.move
@@ -97,22 +97,43 @@ module VerifyVector {
         ensures result[0] == e;
     }
 
+    spec module {
+        /// Ghost variable `old_v` used to store the old value of the array v.
+        /// TODO: by allowing old(.) to appear in assume/assert expressions,
+        /// we can eliminate the need for old_v.
+        global old_v<Element>: vector<Element>;
+    }
+
     // Reverses the order of the elements in the vector in place.
     fun verify_reverse<Element>(v: &mut vector<Element>) {
-        let len = Vector::length(v);
-        if (len == 0) return ();
+        let vlen = Vector::length(v);
+        if (vlen == 0) return ();
 
-//        let front_index = 0;
-//        let back_index = len -1;
-//        while (front_index < back_index) {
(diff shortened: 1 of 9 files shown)


language/move-prover/tests/sources/functional/verify_vector.move
```diff
@@ -97,22 +97,43 @@ module VerifyVector {
...
     // Reverses the order of the elements in the vector in place.
     fun verify_reverse<Element>(v: &mut vector<Element>) {
-        let len = Vector::length(v);
-        if (len == 0) return ();
+        let vlen = Vector::length(v);
+        if (vlen == 0) return ();
+
+        spec {
+            /// Initialize the old vector `old_v` ghost variable
+            assume forall k in 0..vlen: old_v<Element>[k] == v[k];
```
kkmc, line 114: Can probably replace it with `assume old_v<Element> == v`, although not important if we implement some version of old(.).

language/move-prover/tests/sources/functional/verify_vector.move
```diff
@@ -231,21 +252,41 @@ module VerifyVector {
 
     // Remove the `i`th element E of the vector, shifting all subsequent elements
     // It is O(n) and preserves ordering
-    fun verify_remove<Element>(v: &mut vector<Element>, i: u64): Element {
-        let len = Vector::length(v);
+    fun verify_remove<Element>(v: &mut vector<Element>, j: u64): Element {
+        let vlen = Vector::length(v);
```
junkil-park, line 256: Just a note for later. We may need to adjust the name changes (i.e., "len" -> "vlen", and "i" -> "j") later when we add these specs into the actual stdlib vector file (`language/stdlib/modules/vector.move`).

kkmc: There is a related bug on using len #4355 .

PR #4619, opened 2020-06-19
[cluster_test] Unify startup / tear down sequence for runner commands

andll:
This diff unifies startup / teardown sequence for commands that requires cluster test runner.

Previously we had things like wait_until_all_healthy, cleanup and teardown in various commands, but there was no uniform pattern there.

Some commands(like --perf-run) for example would spin up cluster, but don't delete kube pods, but would downscale cluster.

Practical outcome:

* This allows to run certain commands, like `--emit-tx`, otherwise they just fail because they start before cluster becomes healthy.

* Cleanup / tear down cluster if some steps cause error. For example, if cluster is not healthy we previously did not clean it up on some commands


andll:
Tested that `--run`, `--emit-tx` and `--perf-run` works

PR #4624, opened 2020-06-19
[move-prover] Optionally reasoning with ValueArray using sequence theory

murali24adithya:
## Motivation

There have been problems observed in dealing with finite arrays. This PR adds a flag vector_using_sequences in the backend options to reduce the expressions involving the ValueArray type to sequence theory in SMT rather than array theory. The motivation is to use the sequence theory in Z3 and other SMT solvers to compare with the current reasoning scheme, and perhaps as an alternative reasoning mechanism.

Yes

## Test Plan

To test the suite as usual without the sequence theory:
 cargo test 
To use sequence theory on a particular file:
cargo run -- <path/to/file> --config-str "backend.vector_using_sequences = true"
To test the entire suite with sequence theory:
In file move-prover/src/cli.rs, change the default value of the flag which can be found under `impl Default for BackendOptions' to true. Then run cargo test as usual.

Note the following:
(i) There is a boogie crash phenomenon unrelated to this PR that shows up during normal cargo test on functional/module_invariants.move due to unknown reasons.
(ii) Running cargo test with vector_using_sequences set to true does not terminate, and gives failure for several test cases. A few of them are likely reported as failures only because of the difference in expected output, despite being verified appropriately. This has not been addressed in this PR.

### DIFF
--- a/language/move-prover/src/boogie_wrapper.rs
+++ b/language/move-prover/src/boogie_wrapper.rs
@@ -674,6 +674,13 @@ fn make_position(line_str: &str, col_str: &str) -> Location {
 // -----------------------------------------------
 // # Boogie Model Analysis
 
+/// Represents whether the Vector type is implemented at the SMT level using integer maps or sequences
+#[derive(Debug, PartialEq, Eq)]
+pub enum ValueArrayRep {
+    ValueArrayIsMap,
+    ValueArrayIsSeq,
+}
+
 /// Represents a boogie model.
 #[derive(Debug)]
 pub struct Model {
@@ -681,6 +688,7 @@ pub struct Model {
     tracked_locals: BTreeMap<Loc, Vec<(LocalDescriptor, ModelValue)>>,
     tracked_aborts: BTreeMap<(String, LineIndex), AbortDescriptor>,
     tracked_exps: BTreeMap<ExpDescriptor, Vec<ModelValue>>,
+    value_array_rep: ValueArrayRep,
 }
 
 impl Model {
@@ -768,11 +776,17 @@ impl Model {
                 //    );
                 // }
                 // END DEBUG
+                let value_array_rep = if wrapper.options.backend.vector_using_sequences {
+                    ValueArrayRep::ValueArrayIsSeq
+                } else {
+                    ValueArrayRep::ValueArrayIsMap
+                };
                 let model = Model {
                     vars,
                     tracked_locals,
                     tracked_aborts,
                     tracked_exps,
+                    value_array_rep,
                 };
                 Ok(model)
             })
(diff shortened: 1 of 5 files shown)


language/move-prover/src/boogie_wrapper.rs
```diff
@@ -993,45 +1007,87 @@ impl ModelValue {
...
+    ///      else -> (Integer 0)
+    ///    }
+    /// ```
+    /// If the value array is represented by a sequence instead, there are no indirections.
+    /// It has the form
+    /// ```(seq.++ (seq.unit (Integer 0)) (seq.unit (Integer 1)))```
+    /// or
+    /// ```(as seq.empty (Seq T@$Value))```
+    /// depending on whether it is an empty or nonempty sequence, respectively.
+    // In this case the sequence representation does not explicitly denote a constructor like ValueArray(..),
```
shazqadeer, line 1027: Inconsistent number of /.

language/move-prover/src/prelude.bpl
```diff
@@ -131,16 +131,111 @@ function {:inline} $IsValidNum(v: $Value): bool {
...
+        var prefix := $ValueArrayExtract(a, 0, i); (
+        var suffix := $ValueArrayExtract(a, i+1, $LenValueArray(a)-i-1);
+        $ConcatValueArray(prefix, suffix)
+    ))
+}
+
+function {{backend.func_inline}} $ReverseValueArray(a: $ValueArray): $ValueArray;
+axiom (forall a: $ValueArray :: $LenValueArray($ReverseValueArray(a)) == $LenValueArray(a));
+axiom (forall a: $ValueArray :: (forall i: int :: $ValueArrayExtract(a, i, 1) == $ValueArrayExtract(a, $LenValueArray(a)-1-i, 1)));
+// The second axiom works outside the legal range because in z3 seq.extract outside the legal range degenerates to the empty sequence
```
wrwg, line 175: Put a comment line before, not after the thing you are documenting.

language/move-prover/src/prelude.bpl
```diff
@@ -131,16 +131,111 @@ function {:inline} $IsValidNum(v: $Value): bool {
...
+// Not using seq.nth because of weird behaviour outside the legal range
```
wrwg, line 176: Because you commenting practice is not consistent it is unclear what definition you are referring too. In most software languages the convention is:

[empty line]
// comment about the thing
definition of the thing

language/move-prover/src/prelude.bpl
```diff
@@ -131,16 +131,111 @@ function {:inline} $IsValidNum(v: $Value): bool {
...
+
+function {{backend.func_inline}} $SliceValueArray(a: $ValueArray, i: int, j: int): $ValueArray { // return the sliced vector of a for the range [i, j)
```
wrwg, line 178: Move comment before function. Comment must be a legal (syntax, grammar) English sentence.

language/move-prover/src/prelude.bpl
```diff
@@ -131,16 +131,111 @@ function {:inline} $IsValidNum(v: $Value): bool {
...
+        var swap_last_part := $ConcatValueArray($ValueArrayExtract(a,lower,1), end);
+        $ConcatValueArray(swap_first_part, $ConcatValueArray(middle, swap_last_part))
+    )))))))
+}
+
+function {:inline} $IsEmpty(a: $ValueArray): bool {
+    $LenValueArray(a) == 0
+}
+
+//// All invalid elements of array are DefaultValue. This is useful in specialized
```
wrwg, line 215: What does '////` mean? It is confusing.

language/move-prover/src/boogie_wrapper.rs
murali24adithya: I did that because I didn't want to include the last two lines in the description of the function. Maybe I could remove those lines entirely.

language/move-prover/src/bytecode_translator.rs
```diff
@@ -264,30 +264,50 @@ impl<'env> ModuleTranslator<'env> {
...
+                let field_param =
+                    &format!("{}", field_env.get_name().display(struct_env.symbol_pool()));
+                let type_check = boogie_well_formed_check(
+                    self.module_env.env,
+                    field_param,
+                    &field_env.get_type(),
+                    WellFormedMode::Default,
+                );
+                emit!(self.writer, &type_check);
+                ctor_expr = format!("$ExtendValueArray({},{})", ctor_expr, field_param);
```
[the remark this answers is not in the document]
murali24adithya: Maybe this is a separate PR, but if extend is removed how would I handle this construction in this file? Should I collect all the elements and first and then build the sequence myself? Because if everything is inlined that's what the code currently does. Extend is concat applied to the original sequence and a unit constructor, which is how I'd end up constructing a sequence.

PR #4625, opened 2020-06-19
[safety-rules] Improve safety-rules debuggability

davidiw:
Lots of commits that make it easier to identify what is going on in safety rules
* add error logging locally -- before it was only on the libra-node side
* test storage availability at startup -- so we don't let the process run if it can't at least check storage state
* eliminate the use of networking config in safety rules -- now just rely on OPERATOR_ACCOUNT (OWNER_ACCOUNT once @JoshLind catches up)
* Fix an issue in configs where we accidentally left out the deny unknown fields for serde
* Also fixed the vault safety rules test... really need to get vault into ci/cd

### DIFF
--- a/consensus/safety-rules/src/safety_rules_manager.rs
+++ b/consensus/safety-rules/src/safety_rules_manager.rs
@@ -11,7 +11,6 @@ use crate::{
     thread::ThreadService,
     SafetyRules, TSafetyRules,
 };
-use consensus_types::common::Author;
 use libra_config::config::{NodeConfig, SafetyRulesService};
 use libra_secure_storage::{KVStorage, Storage};
 use std::{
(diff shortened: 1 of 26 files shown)


davidiw:
@mgorven this should help out a little bit in the future...

consensus/safety-rules/src/safety_rules_manager.rs
```diff
@@ -11,27 +11,29 @@ use crate::{
...
-            .as_ref()
-            .expect("Missing validator network"),
-    );
-
+pub fn storage(config: &mut NodeConfig) -> PersistentSafetyStorage {
     let backend = &config.consensus.safety_rules.backend;
     let internal_storage: Storage = backend.try_into().expect("Unable to initialize storage");
+    internal_storage
+        .available()
+        .expect("Storage isn ot available");
```
mgorven, line 27: Typo

secure/storage/vault/src/lib.rs
```diff
@@ -198,6 +198,7 @@ impl Client {
     /// Read a key/value pair from a given secret store.
     pub fn read_secret(&self, secret: &str, key: &str) -> Result<ReadResponse<String>, Error> {
         let request = ureq::get(&format!("{}/v1/secret/data/{}", self.host, secret));
+        println!("{} {}", secret, key);
```
mgorven, line 201: Leftover debugging?

PR #4626, opened 2020-06-19
Refactor NetworkBuilder for clarity and modularity

agc-quadra:
## Motivation

WIP

This is a tracking PR on efforts to refactor/cleanup network-builder.

This PR is a holistic effort to better organise network_builder and everything under its purview.  The legacy state of network_builder (and main_node) reflects the initial efforts to get the system to work and to get the functionalities in place and working together.  Those initial efforts succeeded in achieving those goals, but because it grew organically led to a chaotic state of affairs which was occasionally difficult to follow and adjust.

This PR attempts to address those problems in multiple ways.

1. Move construction towards a three phase process:
    (i).  Configure components and construct the requisite wires to connect them.
    (ii).  Build the components and connect the wires
    (iii).  Start the components
   Decomposing the building process in this fashion isolates goals temporally and makes it easier to resolve functional and start-order dependencies.
1.  Encapsulate component construction behind component specific builders for clear isolation of concerns.  Strong encapsulation/isolation makes it easier to reason about individual components in isolation while simultaneously making it easier to see how they connect to each other.  Additionally, strong encapsulation makes it more clear (a) how configurations impact the constructed system and (b) what shape the configurations should take.

This PR achieves the following goals:

1.  It demonstrates a clearer structure for the Networking components.
    (i). it gets `setup_network` out of main_node and into network_builder
    (ii). it encapsulates network_builder logic associated with internal components tightly into the codebases of those components.
1.  It provides a template/structure for a similar principled refactoring of the other components (consensus, mempool, statesync)
1.  In combination with the other refactorings, it facilitates a main_node implementation which is systematic and relatively easy to extend.  When the approach is carried through to the end, main_node could look approximately as follows

```
// create
consensus = ConsensusBuilder::create(ConsensusConfig)
mempool = MempoolBuilder::create(MempoolConfig);
network = NetworkBuilder::create(NetworkConfig);

// wire
network.attach(consensus);
network.attach(mempool);
mempool.attach(consensus);

// build
network.build();
consensus.build();
mempool.build();

// start
network.start();
mempool.start();
consensus.start();

```

Limitations

1.  This PR does introduce structural boiler plate to the code.  My experience has taught me that the encapsulation and regular structure of this boiler plate is useful;  nonetheless it must be acknowledged.
1.  This PR (in current form) addresses the *structural* changes;  there are a multitude of ways in which specific pieces of code can be improved locally.
1. This PR is an isolated milestone that addresses a concern within Networking.   Before the benefits can be applied to main_node, similar PRs for the other top level components (consensus, statesync, mempool, ...) will be needed.

yes

## Test Plan

Not for submission

## Related PRs


network/builder/src/builder.rs
```diff
@@ -371,21 +376,39 @@ impl NetworkBuilder {
...
-        self.executor.spawn(health_checker.start());
-        debug!("{} Started health checker", self.network_context);
+
+        self.health_checker_cfg = Some(health_checker_config);
+
+        self.build_connection_monitoring()
+    }
+
+    fn build_connection_monitoring(&mut self) -> &mut Self {
+        if let Some(health_checker_config) = self.health_checker_cfg.take() {
```
davidiw: why do we have this indirection?

davidiw: Haha, "Outdated" do we fix this in a later PR?

network/builder/src/builder.rs
```diff
@@ -302,25 +306,42 @@ impl NetworkBuilder {
...
-            )
-        });
-        self.executor.spawn(conn_mgr.start());
+
+        let config = ConnectivityManagerBuilderConfig::new(
+            self.network_context().clone(),
+            self.trusted_peers.clone(),
+            self.seed_peers.clone(),
+            self.connectivity_check_interval_ms,
+            2, // Hard-coded constant
```
davidiw, line 276: either don't expose this or make it a const somewhere

agc-quadra: Actually, It should go to a config somewhere.    
But yes.

(too many magic constants)

network/builder/src/builder.rs
agc-quadra: Looking forward $k$ steps, the goal is to have the following in network builder:

```
pub fn build(&mut self) {
  health_checker_builder.build();
  connectivity_monitor_builder.build();
  peer_manager_builder.build();
  ...
}

pub fn start(&mut self, executor) {
  health_checker.start(executor);
  connectivity_monitor.start(executor);
 ...
}
```
and main_node to have something like:

```
setup_environment() {
  for network in network_config {
     network_builders.push(network_builder::create(network));
  }
  consensus_builder = consensus.create(consensus_config)
  ....

  network_builders.build();
  consensus_builder.build();
  state_sync_builder.build();

  ...

  state_sync_builder.start();
  network_builders.start();
   consensus_builder.start();


}

```

The idea being to decouple configure/build/start and isolate race conditions as much as possible.

agc-quadra: > why do we have this indirection?

If you're asking about the config / object gap.... that distinction will go away shortly (within the next ~3 commits)

network/builder/src/builder.rs
```diff
@@ -183,6 +183,99 @@ impl NetworkBuilder {
...
+                        network_builder
+                            .conn_mgr_reqs_tx()
+                            .expect("ConnectivityManager must be installed"),
+                        network_tx,
+                        discovery_events,
+                        network_builder.network_context().clone(),
+                        libra_db,
+                        waypoint,
+                        runtime.handle(),
+                    );
```
davidiw: in a future PR, let's make this the simple on chain discovery... and if it is set and / or seed_peers is set and we're not in mutual_authentication, we want to add connectivity manager.

network/builder/src/builder.rs
```diff
@@ -183,6 +183,99 @@ impl NetworkBuilder {
         }
     }
 
+    pub fn create(
```
davidiw, line 115: nit create versus new, is there a common usage of create in the codebase?

agc-quadra: Intent:

`new` for blindly creating the struct

`create` for applying non-trivial logic and processing from the provided inputs.   Should typically invoke `new` with the processed information.

network_builder needs `create`.  The other builders could probably get away with `new`, but this leads to a consistency question.

network/builder/src/builder.rs
```diff
@@ -184,96 +179,97 @@ impl NetworkBuilder {
...
-            let peer_id = config.peer_id();
-
-            let mut network_builder = NetworkBuilder::new(
-                runtime.handle().clone(),
-                chain_id.clone(),
-                config.network_id.clone(),
-                role,
-                peer_id,
-                config.listen_address.clone(),
+        chain_id: &ChainId,
```
davidiw, line 116: it looks like you had some tabbing issue in this PR and it isn't clear what's changed, is this something you could fix?

network/builder/src/builder.rs
```diff
@@ -496,6 +485,49 @@ impl NetworkBuilder {
...
+    fn build_onchain_discovery(&mut self) -> &mut Self {
+        if let Some(config) = self.onchain_discovery_cfg.take() {
+            let onchain_discovery_builder = OnchainDiscoveryBuilder::build(config, &self.executor);
+            self.onchain_discovery = Some(onchain_discovery_builder);
+            debug!("Built OnchainDiscovery");
+        }
+        self.start_onchain_discovery()
+    }
+
+    fn start_onchain_discovery(&mut self) -> &mut Self {
```
davidiw: prefer inline, this pattern doesn't make sense. I think I saw this earlier... so apologies if you responded.

agc-quadra: this is one of the important discussion points and a primary reason to make the PR visible sooner than later.

network/builder/src/builder.rs
agc-quadra: > Haha, "Outdated" do we fix this in a later PR?

I have no idea how "outdated" shows up. :)

Maybe it gets fixed in a later commit that's in the PR????

network/builder/src/builder.rs
```diff
@@ -108,12 +112,16 @@ pub struct NetworkBuilder {
...
+    configuration_change_listener_cfg: Option<ConfigurationChangeListenerConfig>,
+    configuration_change_listener: Option<ConfigurationChangeListener>,
     discovery_cfg: Option<DiscoveryBuilderConfig>,
     discovery: Option<DiscoveryService>,
     health_checker_cfg: Option<HealthCheckerBuilderConfig>,
     health_checker: Option<HealthCheckerService>,
     onchain_discovery_cfg: Option<OnchainDiscoveryBuilderConfig>,
     onchain_discovery: Option<OnchainDiscoveryBuilder>,
+
+    reconfig_subscriptions: Vec<ReconfigSubscription>,
```
davidiw, line 70: why not put this in ConfigurationChangeListener

network/builder/src/builder.rs
```diff
@@ -295,58 +239,13 @@ impl NetworkBuilder {
...
-    }
-
     pub fn add_connection_event_listener(&mut self) -> conn_notifs_channel::Receiver {
-        let (tx, rx) = conn_notifs_channel::new();
-        self.connection_event_handlers.push(tx);
-        rx
+        match &mut self.peer_manager_builder {
+            Some(builder) => builder.add_connection_event_listener(),
+            None => panic!(
+                "Cannot add a conneciton event listener if PeerManagerBuilder does not exist"
```
davidiw, line 237: sp

network/builder/src/builder.rs
```diff
@@ -295,58 +239,13 @@ impl NetworkBuilder {
...
-            connection_notifs_rx,
-        )
-    }
-
     pub fn add_connection_event_listener(&mut self) -> conn_notifs_channel::Receiver {
-        let (tx, rx) = conn_notifs_channel::new();
-        self.connection_event_handlers.push(tx);
-        rx
+        match &mut self.peer_manager_builder {
+            Some(builder) => builder.add_connection_event_listener(),
```
davidiw, line 235: peer manager isn't optional. also I think it is better to use .as_ref().expect() instead.

network/builder/src/builder.rs
```diff
@@ -369,6 +268,12 @@ impl NetworkBuilder {
         );
         self.conn_mgr_reqs_tx = Some(conn_mgr_reqs_tx);
         let pm_conn_mgr_notifs_rx = self.add_connection_event_listener();
+        let connection_reqs_tx = match &self.peer_manager_builder {
```
davidiw, line 263: again, same here, the logic is a little confusing since these are /expected/ to be here.

network/builder/src/builder.rs
```diff
@@ -291,24 +287,24 @@ impl NetworkBuilder {
...
             connectivity_builder.build(&self.executor);
-            if let Some(config) = self.configuration_change_listener_cfg.take() {
-                let configuration_change_listener =
-                    network_simple_onchain_discovery::build_configuration_change_listener(config);
-                self.configuration_change_listener = Some(configuration_change_listener);
+            if let Some(ref mut change_listener_builder) =
+                self.configuration_change_listener_builder
+            {
+                change_listener_builder.build();
             }
```
davidiw: have some fun with rust:
`self.configuration_change_listener_builder.map(|c| c.build());`

network/builder/src/builder.rs
```diff
@@ -492,13 +490,19 @@ impl NetworkBuilder {
...
+                queue_preference,
+                max_queue_size_per_peer,
+                counter,
+            ): (
+                Vec<ProtocolId>,
+                Vec<ProtocolId>,
+                QueueStyle,
+                usize,
+                Option<&'static IntCounterVec>,
+            ),
```
davidiw: probably best to make a type here?

network/builder/src/builder.rs
agc-quadra: The reconfig subscriptions are needed (later on) by StateSync, so network_builder must provide an accessing interface.    We can implement that either by invoking all internal components that produce subscriptions or by pro-actively gathering the subscriptions as they are generated.

I chose the latter, but there is no fundamental reason to not choose the former.

network/builder/src/builder.rs
agc-quadra: Yeah.  These options are really bad hacks to work around (a) needing to set values when the struct is created and (b) the apparent need to drop the object into the tokio.runtime.

There is a slight question (not quite there with the code to make it obvious) if there is benefit to (a) treating all of the builders the same way for symmetry or (b) being smart with special casing.

The latter is more concise and does fewer unnecessary things;  the former is more flexible and imposes fewer handcuffs moving forward.

network/builder/src/builder.rs
agc-quadra: yes.  Probably, maybe.   That can be done in an orthgonal PR.

network/builder/src/builder.rs
agc-quadra: An implied corollary is removing onchain discovery?

davidiw: Removing, not quite, we likely need to postpone to after V1. I imagine it will be quite useful to our ecosystem for light clients -- which we currently have yet to implement :).

network/builder/src/builder.rs
davidiw: outdated means you have a massive stack!

network/builder/src/builder.rs
davidiw: I see, given that we have these other builders and it is only used once, should we push it into the builders themselves?

network/builder/src/builder.rs
davidiw: Well, considering the panic below, I think it seems perfectly fine to do the following:
`self.peer_manager_builder.as_ref().expect("...");`

agc-quadra:
> Gonna ask @phlip9 and @gregnazario for a quick look over of this.

Thanks in advance for taking a look.   @davidiw may be optimistic with respect to 'quick'

> 
> My concern is we're taking in a lot of code +800. The reason for this is that we glued things together in the past. Now we're working toward better encapsulation and separation of service creation and service starting. In fact many of the new lines of code are just from re-organizing into new files and the headers associated with that.

Very good summary of the tradeoffs.   I believe it to be a worthwhile tradeoff, but it is important that others see it that way as wel.

> 
> @agclement, it might be useful to place at the description of this PR your mindset to help drive the conversation.

PR description updated.

network/builder/src/builder.rs
agc-quadra: > have some fun with rust:
> `self.configuration_change_listener_builder.map(|c| c.build());`

that makes my head hurt.

network/builder/src/builder.rs
agc-quadra: I lean towards keeping this specifically at the network_builder level because it is somehow 'specific' to StateSync.  So it is good for network_builder to consolidate things together since it can, in principle, come from multiple internal components.

There are other things, such as conn_mgr_reqs_tx, which should remain with the "internal" component because multiple external components may access it.

That is actually the principled distinction that I draw:

- For wires that are many-external to one-internal, the wire should live internally.
- For wires that are one-external to **-internal, the wires should leave as close to the external boundary as possible.

network/builder/src/builder.rs
agc-quadra: This would be good as a type and the function itself would be very good as a `pub trait NetworkClientService` or something along those lines;  a significant difficulty on that front is `onchain_discovery` (and `simple_onchain_discovery`) being separate crates within networking.

This specific piece would be significantly simpler/cleaner if we moved `onchain_discovery` properly into the networking crate (at which point we could move `network_builder` back into the networking crate as well)

network/builder/src/builder.rs
agc-quadra: > Haha, "Outdated" do we fix this in a later PR?

I think outdated shows up when I rebase and then force push.

agc-quadra: > outdated means you have a massive stack!

22 commits and counting.

network/builder/src/builder.rs
agc-quadra: Clarify:  this is one of the important low-level discussion points.    I am prioritizing the top level structure at present and can do lower level cleanups once the shape of the end goal has converged.

agc-quadra:
> I think the direction is a lot less complicated / confusing for what we got right now, so that's great!

Woohoo!

> 
> My quick thoughts:
> 
> 1. There's now a ton more configs, but it seems like they're simpler.  Should we put in the RustDocs for all the parameters, or leave those to just be defined in the actual objects.

I look to you guys for guidance there.  I would lean towards RustDocs, but don't know

> 2. I see you're going forward with the `create(init args)` -> `build(executor)` -> `start`. I see we have `if let ..., do` should we add errors in case an object is tried to be started / built without the prerequisite steps within the `Builders` rather than in the downstream code?

I believe these checks should be (in some variant) everywhere.

In `foo_builder` ({network, health_checker, peer_manager} \in foo), the following steps are appropriate:

- `build` should be called once
- `build_inner_component` should invoke `inner_component_builder.build()` only if inner_component exists

There is a very real question if we want graceful failures or believe that this is an appropriate context to panic.   Given that builders are invoked on startup, I tend to think panic is appropriate.  This is an exception to my normal belief.

> 3. Will the builders now handle all of the `channel` wiring?

The patterns that I've identified are as follows:

-  Network_builder (and eventually main_node) are responsible for wiring components together
-  The source component 'owns' all incoming wires.  i.e., peermanager owns/creates all of the wires used by other to connect to peermanager
- the "grouping" component (i.e. network_builder) holds the wires that multiple internal components can use to connect to a single external component (i.e., reconfiguration_subscriptions that connect to statesync)

The intent is that main_node creates the top level components and gets the wires to connect them from the top level builders;  network_builder creates the internal components and wires them together;  network_builder also provides external facing wires (typically by fetching them from the appropriate internal component).

So I think the answer to your question is 'yes', but with some specific caveats.

> 4. Maybe we should only do one of the builders at a time in a PR, it might make it easier to rebase and get in?

hehehheehe.   Much much easier;  but going one commit at a time is very incremental and doesn't offer a good picture of the end-to-end view.   My goal in this (way too big) PR is to get the end-to-end picture in place and accessible so that direction can be evaluated for a go/no-go on the direction.  (seems to be leaning towards 'go')

There's a second question of how we want to move it forward.  I can either pull off the individual commits into single PRs one at a time, which will give time/space/scope to get all local changes correct or I can send the larger end-to-end structural change through and do targeted local cleanup PRs afterwards.

The former path will be 'cleaner' but take longer while the latter will get in faster but leave some cleanup to be done afterwards.

I can go either way.  My aesthetic instincts prefer the series of tiny PRs;  my pragmatic mind wants to get the skeletal change in place so there are fewer dependent balls in the air.

libra-node/src/main_node.rs
```diff
@@ -181,12 +168,36 @@ pub fn setup_environment(node_config: &mut NodeConfig) -> LibraHandle {
...
+    for network_builder in &mut network_builders {
         // Start the network and cache the runtime so it does not go out of scope.
-        // TODO:  move all 'start' commands to a second phase at the end of setup_environment.  Target is to have one pass to wire the pieces together and a second pass to start processing in an appropriate order.
-        let _listen_addr = network_builder.build();
+        let runtime = Builder::new()
+            .thread_name("network-")
+            .threaded_scheduler()
+            .enable_all()
+            .build()
+            .expect("Failed to start runtime. Won't be able to start networking.");
```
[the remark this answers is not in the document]
agc-quadra: Legacy code.  I've been focusing on the structure and intentionally avoiding polluting my thoughts with 'details' for how it works or what it is doing. :)

Put another way:  "say no to scope creep!"

libra-node/src/main_node.rs
```diff
@@ -128,15 +127,16 @@ pub fn setup_environment(node_config: &mut NodeConfig) -> LibraHandle {
 
     // Instantiate every network and collect the requisite endpoints for state_sync, mempool, and consensus.
     for (role, network_config) in network_configs {
+        let peer_id = network_config.peer_id();
+
         // Perform common instantiation steps
-        let (runtime, mut network_builder) = network_builder::builder::setup_network(
+        let mut network_builder = NetworkBuilder::create(
```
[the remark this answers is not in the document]
agc-quadra: I'm attempting to enforce a meaningful (and somewhat arbitrary) distinction between `::new` which should blindly create a struct and `::create`/`::init` which may apply additional logic and processing before invoking `::new` internally.

There are likely additional passes needed before that religion is dogmatically followed in the code; but it is a best-pratices religion that I see as being quite valuable.

For `::init` versus `::create` I am completely and totally agnostic between the two.   Unless you feel very strongly that `::init` is a better way to go, (and possibly even then) I'd prefer to keep `::create` for now and change to `::init` in a focused follow-up PR.

libra-node/src/main_node.rs
```diff
@@ -181,12 +168,36 @@ pub fn setup_environment(node_config: &mut NodeConfig) -> LibraHandle {
...
+        let runtime = Builder::new()
+            .thread_name("network-")
+            .threaded_scheduler()
+            .enable_all()
+            .build()
+            .expect("Failed to start runtime. Won't be able to start networking.");
+
+        // Note that due to idiosyncracies of internal networking components, the runtime handle
+        // must be provided before they can be built.
+        network_builder.build(runtime.handle().clone());
```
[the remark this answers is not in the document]
agc-quadra: See previous comment.  `::new` should be dumb.

libra-node/src/main_node.rs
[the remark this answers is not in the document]
gregnazario: I have no real preference as well.  Just stick with what you got, but as I've gone further through the code I understood the create flow.

gregnazario:
> > I think the direction is a lot less complicated / confusing for what we got right now, so that's great!
> 
> Woohoo!
> 
I'd ask @phlip9 to take a quick look and make sure that the ideology of the new method is fine to him as well.  Personally, I haven't been here very long, so I'm much more inclined for change.


> > My quick thoughts:
> > 
> > 1. There's now a ton more configs, but it seems like they're simpler.  Should we put in the RustDocs for all the parameters, or leave those to just be defined in the actual objects.
> 
> I look to you guys for guidance there. I would lean towards RustDocs, but don't know.
> 
> At this point, I would also lean towards doing a follow up PR for full documentation purposes. But that introduces a dangerous liveness problem.
> 
> > 1. I see you're going forward with the `create(init args)` -> `build(executor)` -> `start`. I see we have `if let ..., do` should we add errors in case an object is tried to be started / built without the prerequisite steps within the `Builders` rather than in the downstream code?
> 
> I believe these checks should be (in some variant) everywhere.
> 
> In `foo_builder` ({network, health_checker, peer_manager} \in foo), the following steps are appropriate:
> 
> * `build` should be called once
> * `build_inner_component` should invoke `inner_component_builder.build()` only if inner_component exists
> 
> There is a very real question if we want graceful failures or believe that this is an appropriate context to panic. Given that builders are invoked on startup, I tend to think panic is appropriate. This is an exception to my normal belief.
> 
I think `panic` is the right way to go, let's not be too graceful if it won't work without it.

> > 1. Will the builders now handle all of the `channel` wiring?
> 
> The patterns that I've identified are as follows:
> 
> * Network_builder (and eventually main_node) are responsible for wiring components together
> * The source component 'owns' all incoming wires.  i.e., peermanager owns/creates all of the wires used by other to connect to peermanager
> * the "grouping" component (i.e. network_builder) holds the wires that multiple internal components can use to connect to a single external component (i.e., reconfiguration_subscriptions that connect to statesync)
> 
> The intent is that main_node creates the top level components and gets the wires to connect them from the top level builders; network_builder creates the internal components and wires them together; network_builder also provides external facing wires (typically by fetching them from the appropriate internal component).
> 
> So I think the answer to your question is 'yes', but with some specific caveats.
> 
> > 1. Maybe we should only do one of the builders at a time in a PR, it might make it easier to rebase and get in?
> 
> hehehheehe. Much much easier; but going one commit at a time is very incremental and doesn't offer a good picture of the end-to-end view. My goal in this (way too big) PR is to get the end-to-end picture in place and accessible so that direction can be evaluated for a go/no-go on the direction. (seems to be leaning towards 'go')
> 
> There's a second question of how we want to move it forward. I can either pull off the individual commits into single PRs one at a time, which will give time/space/scope to get all local changes correct or I can send the larger end-to-end structural change through and do targeted local cleanup PRs afterwards. Note that the incremental sequence of single commits will necessarily take detours through somewhat arbitrary intermediate states that are, in isolation, not obvious.
> 
> The former path will be 'cleaner' but take longer while the latter will get in faster but leave some cleanup to be done afterwards.
> 
> I can go either way. My aesthetic instincts prefer the series of tiny PRs; my pragmatic mind wants to get the skeletal change in place so there are fewer dependent balls in the air.

I'm inclined if you do make a change like this to do it one builder at a time, or at least some smaller pieces.  It would be preferred if it was all at once, but you risk 2 things:
1. No one wants to read a really long PR, so it may get stale quickly and require lots of rebases
2. It has a large surface area, so it's really easy to have to also require a bunch of changes

It may also be possible you can sneak in pieces while people aren't working with the builders directly.

agc-quadra:
> I'm inclined if you do make a change like this to do it one builder at a time, or at least some smaller pieces. It would be preferred if it was all at once, but you risk 2 things:
> 
> 1. No one wants to read a really long PR, so it may get stale quickly and require lots of rebases
> 2. It has a large surface area, so it's really easy to have to also require a bunch of changes
> 
> It may also be possible you can sneak in pieces while people aren't working with the builders directly.

In principle I agree with you.  In practice for this specific changeset (and with the added complexity of 9 timezones), the incremental path is likely to take 3-4 weeks.   That's the reality as I see it right now.

Unless y'all come back otherwise, tomorrow I'll go down the path of trying to pull individual builders out and see where that leads us.   In practice I'm probably going to need to keep this branch up to date so I have a good "final" image in place, but that will be an issue for me to stay on top of.

gregnazario:
> > I'm inclined if you do make a change like this to do it one builder at a time, or at least some smaller pieces. It would be preferred if it was all at once, but you risk 2 things:
> > 
> > 1. No one wants to read a really long PR, so it may get stale quickly and require lots of rebases
> > 2. It has a large surface area, so it's really easy to have to also require a bunch of changes
> > 
> > It may also be possible you can sneak in pieces while people aren't working with the builders directly.
> 
> In principle I agree with you. In practice for this specific changeset (and with the added complexity of 9 timezones), the incremental path is likely to take 3-4 weeks. That's the reality as I see it right now.
> 
> Unless y'all come back otherwise, tomorrow I'll go down the path of trying to pull individual builders out and see where that leads us. In practice I'm probably going to need to keep this branch up to date so I have a good "final" image in place, but that will be an issue for me to stay on top of.
> 
> ps. ask @davidiw about some earlier conversations on PR size :)

Poke me anytime and I'll take a look at the PR.

If I'm at my computer I'll respond even if it's beginning of the next day.  I'll try to keep it timely :)

agc-quadra:
> > > I'm inclined if you do make a change like this to do it one builder at a time, or at least some smaller pieces. It would be preferred if it was all at once, but you risk 2 things:
> > > 
> > > 1. No one wants to read a really long PR, so it may get stale quickly and require lots of rebases
> > > 2. It has a large surface area, so it's really easy to have to also require a bunch of changes
> > > 
> > > It may also be possible you can sneak in pieces while people aren't working with the builders directly.
> > 
> > 
> > In principle I agree with you. In practice for this specific changeset (and with the added complexity of 9 timezones), the incremental path is likely to take 3-4 weeks. That's the reality as I see it right now.
> > Unless y'all come back otherwise, tomorrow I'll go down the path of trying to pull individual builders out and see where that leads us. In practice I'm probably going to need to keep this branch up to date so I have a good "final" image in place, but that will be an issue for me to stay on top of.
> > ps. ask @davidiw about some earlier conversations on PR size :)
> 
> Poke me anytime and I'll take a look at the PR.
> 
> If I'm at my computer I'll respond even if it's beginning of the next day. I'll try to keep it timely :)

Poke for #4734 --- it's a structural rename/move of setup_network into create;  it blocks being able to do the individual builder changes in isolation.

davidiw:
Looks like this is long since deprecated? Feel free to re-open.

PR #4628, opened 2020-06-19
[move source language] Inline blocks optimization

tnowacki:
## Motivation

- Optimized JumpIf(true|false) to Jump
- Inlined blocks that have a single Jump targeting them
- Needed for constant folding (boolean operator constants)

## Test Plan

- cargo test
- cannot explicitly test without constants

### DIFF
--- a/language/move-lang/src/cfgir/inline_blocks.rs
+++ b/language/move-lang/src/cfgir/inline_blocks.rs
@@ -1,0 +1,89 @@
+// Copyright (c) The Libra Core Contributors
+// SPDX-License-Identifier: Apache-2.0
+
+use super::cfg::{BlockCFG, CFG};
+use crate::hlir::ast::{BasicBlocks, Command_, Label};
+use std::collections::{BTreeMap, BTreeSet};
+
+/// returns true if anything changed
+pub fn optimize(cfg: &mut BlockCFG) -> bool {
+    let changed = optimize_(cfg.start_block(), cfg.blocks_mut());
+    if changed {
+        let dead_blocks = cfg.recompute();
+        assert!(dead_blocks.is_empty())
+    }
+    changed
+}
+
+fn optimize_(start: Label, blocks: &mut BasicBlocks) -> bool {
+    let single_target_labels = find_single_target_labels(start, &blocks);
+    inline_single_target_blocks(&single_target_labels, blocks)
+}
+
+fn find_single_target_labels(start: Label, blocks: &BasicBlocks) -> BTreeSet<Label> {
+    use Command_ as C;
+    let mut counts = BTreeMap::new();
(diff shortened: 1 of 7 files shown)


language/move-lang/src/cfgir/inline_blocks.rs
```diff
@@ -0,0 +1,90 @@
...
+        let mut block = match working_blocks.remove(&cur) {
+            None => {
+                next = labels.next();
+                continue;
+            }
+            Some(b) => b,
+        };
+
+        match block.back().unwrap() {
+            // check for single target forward jump
```
bob-wilson, line 69: Maybe checking for forward jumps is safe -- I'm not sure -- but it seems like the wrong condition to check. It seems to me like inlining should be fine for a backward jump as long as there are no other jumps to that label and if there is no fall-through into that block. You're already checking for the single-jump condition. How about changing to check for fall-through instead of forward jumps?

tnowacki: There is are no fall thoughs in the cfgir. And the block ordering might not be contiguous, so you cannot check for `cur.0 + 1 == target.0` 

The point of this check is to ensure that there are no loops. The backward jumps are *not* safe if there are loops. And currently the *only* back jumps we emit are loops

That being said. I believe that in the current system, if it is not the start block, loop heads always serve as the target for 2 or more jumps. So I will switch the check for not being the start

bob-wilson: Ah, OK, I didn't realize that the cfgir has no fallthroughs. Maybe I'm being dense, but what's the problem with backward jumps with loops?

tnowacki: Imagine this scenario 

```
label Start:
    JumpIf(copy(check),  End, Next)

Comment (bob-wilson) [reply language/move-lang/src/cfgir/inline_blocks.rs]: In that case, there would be a fall-through into InfiniteLoopHead that would block the inlining. (Either that or a jump into the body of the loop, but you're going to ban that with the pending control flow checks.)

Comment (bob-wilson) [reply language/move-lang/src/cfgir/inline_blocks.rs]: Sorry, not a fall-through, but another jump to the loop head (which is what you said above). It sounds like we're in agreement on that.

File: language/move-lang/src/cfgir/inline_blocks.rs [hunk 2]
```diff
@@ -0,0 +1,89 @@
...
+        let mut block = match working_blocks.remove(&cur) {
+            None => {
+                next = labels.next();
+                continue;
+            }
+            Some(b) => b,
+        };
+
+        match block.back().unwrap() {
+            // Do not need to worry about inifinitely unwrapping loops as loop heads will always
```
bob-wilson, line 72: Typo: "infinitely"

PR #4629, opened 2020-06-19
[cluster-test] Introduce Instance::exec

andll:
For some use cases we need to run command on the container
This diff introduces method `Instance::exec` and `--exec` command for a runner to try it out

This diff also adds `--cfg name=value,name=value` argument that allows to set command override

One way to check that override is working is something like this:

```
 ./scripts/cti --pr 4629 --cfg prune_window=5000 --exec 'val-1:cat /opt/libra/etc/node.config.toml'
```

### DIFF
--- a/scripts/cti
+++ b/scripts/cti
@@ -43,7 +43,7 @@ shuffle() {
 
 join_args() {
   retval_join_args=""
-  for var in $*
+  for var in "$@"
   do
     retval_join_args="${retval_join_args}, \"${var}\""
   done
@@ -217,7 +217,7 @@ pod_name="cluster-test-$(whoami)-$(date +%s)"
 pod_name=${pod_name/_/-} #underscore not allowed in pod name
 specfile=$(mktemp)
 echo "Pod Spec : ${specfile}"
-join_args $*
+join_args "$@"
 ENV="$ENV AWS_ROLE_SESSION_NAME=AWS_ROLE_SESSION_NAME"
 join_env_vars $ENV
 DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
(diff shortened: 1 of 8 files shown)


testsuite/cluster-test/src/instance.rs
```diff
@@ -271,6 +272,42 @@ impl Instance {
...
             .await
     }
+
+    /// Unlike util_cmd, exec runs command inside the container
+    pub async fn exec(&self, command: &str) -> Result<()> {
+        let child = Command::new("kubectl")
+            .arg("exec")
+            .arg(&self.peer_name)
+            .arg("--container")
+            .arg("main")
```
ankushagarwal, line 282: Can you update the LSR container to be named `main` as well?

scripts/cti
```diff
@@ -217,15 +217,15 @@ pod_name="cluster-test-$(whoami)-$(date +%s)"
...
+join_args "$@"
 ENV="$ENV AWS_ROLE_SESSION_NAME=AWS_ROLE_SESSION_NAME"
 join_env_vars $ENV
 DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
 sed -e "s/{pod_name}/${pod_name}/g" \
     -e "s/{image_tag}/${TAG}/g" \
     -e "s/{cluster_test_image_tag}/${CLUSTER_TEST_TAG}/g" \
     -e "s^{env_variables}^${retval_join_env_vars}^g" \
-    -e "s/{extra_args}/${retval_join_args}/g" \
+    -e "s+{extra_args}+${retval_join_args}+g" \
```
ankushagarwal, line 228: I imagine this must've been a fun thing to debug :)

testsuite/cluster-test/src/instance.rs
andll: done

andll:
Added few more commits

PR #4631, opened 2020-06-19
[core modules] Unify error code representation

tzakian:
This PR unifies the error code representation within the core modules.

The general proposal for error representation:
* There are a set of shared "core error" codes defined in `0x1::CoreErrors`. These are in the range `[0,255]` (`CORE_ERR_RANGE = 255`).
* Each module declares a `MODULE_ERROR_BASE`  in ranges of 1000 (e.g. `[1000, 2000)`.
* Errors specific to that module are declared in
`(MODULE_ERROR_BASE + CORE_ERR_RANGE, MODULE_ERROR_BASE + 1000)`
* Core error `E` raised from a module with `MODULE_BASE_ERROR` `M` will be represented as `M + E` 

This PR is a draft at the moment, but once we agree on the error representation I'll update the VM and tests.


language/stdlib/modules/SlidingNonce.move
```diff
@@ -31,20 +37,20 @@ module SlidingNonce {
     /// Tries to record this nonce in the account.
     /// Returns 0 if a nonce was recorded and non-0 otherwise
     /// Reasons for nonce to be rejected:
-    /// * code 10001: This nonce is too old and impossible to ensure whether it's duplicated or not
-    /// * code 10002: This nonce is too far in the future - this is not allowed to protect against nonce exhaustion
-    /// * code 10003: This nonce was already recorded previously
+    /// * NONCE_TOO_OLD: This nonce is too old and impossible to ensure whether it's duplicated or not
+    /// * NONCE_TOO_NEW: This nonce is too far in the future - this is not allowed to protect against nonce exhaustion
+    /// * NONCE_ALREADY_RECORDED: This nonce was already recorded previously
```
moeziniaf, line 42: I know this is for documentation but it feels weird to describe the Errors where they are used, and not where they are defined.

sblackshear:
Love the approach!

Thoughts on how we can make it easier to select non-conflicting `MODULE_ERROR_BASE`s? E.g., maybe the code that builds the stdlib can check uniqueness

language/stdlib/modules/SlidingNonce.move
tzakian: Good point :) I will move up to the point of definition.

tzakian:
> We will look back at the old days and laugh :) I don't know what conversations have been had around this topic but I'm assuming there's some reason why we prefer error codes to custom struct defined errors (also custom errors need func->u64 defining in any case)?

The main reason we're thinking of moving away from custom errors is the ease of debugging and writing -- we would still have custom errors, just this formalizes the representation of them so that we don't need to worry about the same error code coming from different modules. Not sure if I answered your question though, so let me know if I didn't...  

> It's awesome to have shared CoreErrors (0-255) for searchability, and for ease of finding the throwing module - do we have a plan if we exceed 255 CoreErrors?

It's a good question. My thinking was that if we get to the point where there are more than 255 core errors, we need to look at pruning/combining some of these errors (and whether some of them should be moved to a module-level error code). I'm happy to bump the range on this if we think this will be a real limitation though!

> Also, if the debugger shows an error code it might take two steps to find the error (first look up Core error name and use the error base number to find the corresponding module name) but I guess we'll all know the Core error by heart very soon :)

I think you're right that this might make things a bit more work at first but, that being said, I think this will overall be better than the current setting where you have to constantly wonder if your error is stepping on some other module's error. Additionally, due to the range restrictions on core errors it should be easy to determine if it's a "core error" or a non-core error, and in which module.

tnowacki:
I don't really like this. But we will see based on these questions.

What is the purpose of `MODULE_ERROR_BASE`? Why are errors between modules not inherently distinct? 
If they are inherently distinct, why do we need these sectioned off regions for codes?
If they aren't inherently distinct, I think this is failing of the language. I know on several occasions we have talked with @dariorussi about improving this at the VM level, that it would include the module with the error code. Do we not do this?

tnowacki:
> The main reason we're thinking of moving away from custom errors is the ease of debugging and writing -- we would still have custom errors, just this formalizes the representation of them so that we don't need to worry about the same error code coming from different modules

I think we should worry about that deeply. Especially for the system going forward for 10+ years of new code being added. From a language level, we should separate the codes based on the module they were thrown from

tnowacki:
Going further, I think from the VM level we should give back the: Module, Function, Code Offset, Error type + Code 

This PR seems to be assuming we will never have that additional information

tzakian:
> What is the purpose of MODULE_ERROR_BASE? Why are errors between modules not inherently distinct?
> If they are inherently distinct, why do we need these sectioned off regions for codes?
If they aren't inherently distinct, I think this is failing of the language. I know on several occasions we have talked with @dariorussi about improving this at the VM level, that it would include the module with the error code. Do we not do this?

I understand what you're saying about errors being unique from the VM, and I believe this is something that Dario is working on. However, unique `VMStatus`'s thrown from the VM don't solve all of the issues. In particular, we have access to these on-node (i.e. for local debugging etc.) but these are too bulky to save with every committed transaction[1]. So we need some more paired-down version that can uniquely identify and point to the position where an error occured--an error code. However, we can't make sure that the error codes between modules are distinct without having a well-defined method for determining the allowable set of error codes that can be thrown within a given module. This is the reason for `MODULE_ERROR_BASE`; instead of having to worry about this on a per-`assert`/`abort` level, you only have to worry about uniqueness at the module level. 

> I think we should worry about that deeply. Especially for the system going forward for 10+ years of new code being added. From a language level, we should separate the codes based on the module they were thrown from

But...that's what this is doing. It's defining a convention for how to separate the error codes by the module they were thrown from. Additionally, relying on the uniqueness of errors in tests by using mod+fun+code offset means that any tests written against these identifiers could be exceptionally brittle--non-semantic changes in the modules themselves (addition of code in the function, changing of mod/fun name) as well as changes to the compiler could cause these to change and tests to break. The number of changes that would cause a test matching against a unique error code to fail are generally fewer and would generally represent an actual semantic--as opposed to syntactic--change in the code.

> Going further, I think from the VM level we should give back the: Module, Function, Code Offset, Error type + Code

Totally agreed, and I think this level of detail is excellent for debugging. However, I don't think it's necessarily what we want to use for testing, or for storing the result of a transaction.


[1] In a conversation with Dario just now, we discovered that only the major status is saved, but I am looking in to seeing if this can be both major + minor status.

tnowacki:
I skipped a lot of steps in my reasoning above and I think it has led to some confusion 

TL;DR The `M+E` approach has the right intuition. But we need `M::E` if we want error codes to be discernible without retracing execution 

**Assumption: We want error codes to be meaningful and discernible** 

> Totally agreed, and I think this level of detail is excellent for debugging. However, I don't think it's necessarily what we want to use for testing, or for storing the result of a transaction.

Based on this information, I'm going to assume we want error codes to cary some sort of meaning (give an indication of why the error occurred), and that we want error codes to be discernible (cannot confuse an error code for another error code)

**Assumption: We will have user specified Move programs in 10+ years**

This means that some user will be writing modules/scripts that interface with the standard library and Libra framework code. And I'm assuming we want these error codes to carry that same meaning in this context.

**Argument: The current M+E proposal is not a safe solution**

- `abort` is a normal instruction that can be used in any script/module. There are no control on `abort` like there are for other instructions (`move_to` and the like)
- `abort` takes a normal u64 specified by the programmer 
  - (switching this to a constant makes things mildly better but not really, you can encode the dynamic u64 behavior with a constant `abort` using a big switch/case)
- If we have custom code in 10+ years, any programmer can specify any error code they want in their module or script
- This means that error codes are no longer discernible since you don't know if it was a "real" version of that code. 
  - For example, a user could do `abort OUT_OF_GAS` wherever they please
- Since they are no longer discernible, you cannot rely on the codes for anything useful (access control, rate limiting, etc). Thus they lose any meaning in the system itself. They could still carry some meaning for debugging purposes, but it is a "weak" meaning, other information would be needed for high debugging assurance 
- Therefore, we have arrived at a possible world where just having `M+E` does not provide the guarantees we want 
  - I would argue that the current implementation of `abort` means this `M+E` proposal can provide *no* guarantees. So even if I haven't properly represented the guarantees that you want, I posit that any other guarantee would fail under the same test. 

- As a followup, I know one intent of this `M+E` proposal is to make discernible *stored* error codes. But from the argument above, I think that you lose any of that discernibility with custom code. 
  - As such, you would need to analyze the transaction itself to figure out which error code it *actually* was
  - And thus, going through all this trouble to build this `M+E` suite of constants isn't worth the effort

**Counter proposal: Give module information alongside the the error code**

Much like we have done with resources and event: If you want behavior tied to a module, you have to make it private to the module.

In resources and events, we (in some sense) achieve this by attaching the module id (address + name) to the type associated with that resource/event. 

I think this `M+E` proposal had the right idea of trying to do the same thing with error codes! But as I argued above, the current mechanism is just too brittle.

So as a counter proposal I would give is `M::E`, that is we store `(Option<(Module Address, Module Name)>, Error Code)` . 
Details:
- The VM could provide the full information of Module > Function > code offset > error code, but we would not store all of that information
- It needs to be an option due to scripts 
- Just storing `(Option<(Module Address, Module Name)>, Error Code)` is the minimal amount of information needed to provide assurance that the error code is really the error code you think it has
- This has the benefit of simplifying the code. The module can just enumerate its codes `0, 1, 2, 3, ....`
- This makes adding new modules easier and less brittle
- This prevents user defined error codes from interfering with the core frame work
- If we are in a tight space crunch we could maybe switch error codes to `u8`?
  - If that is not enough to help, we could hash the `(Module Address, Module Name)` pair
  - Still need to make sure we have an option for scripts

tzakian:
> I skipped a lot of steps in my reasoning above and I think it has led to some confusion
> TL;DR The M+E approach has the right intuition. But we need M::E if we want error codes to be discernible without retracing execution

Totally agreed in that case. If we can have the `M` automatically guaranteed for us in the errors, I will happily abandon this PR.

> Assumption: We want error codes to be meaningful and discernible
> Totally agreed, and I think this level of detail is excellent for debugging. However, I don't think it's necessarily what we want to use for testing, or for storing the result of a transaction.
> Based on this information, I'm going to assume we want error codes to cary some sort of meaning (give an indication of why the error occurred), and that we want error codes to be discernible (cannot confuse an error code for another error code)

Agreed. Although the semantic meaning of abort error codes is, strictly speaking, non-semantic outside of the prologue.

> Assumption: We will have user specified Move programs in 10+ years
>This means that some user will be writing modules/scripts that interface with the standard library and > Libra framework code. And I'm assuming we want these error codes to carry that same meaning in this context.

> Argument: The current M+E proposal is not a safe solution

> abort is a normal instruction that can be used in any script/module. There are no control on abort like there are for other instructions (move_to and the like)

>abort takes a normal u64 specified by the programmer

>(switching this to a constant makes things mildly better but not really, you can encode the dynamic u64 behavior with a constant abort using a big switch/case)
If we have custom code in 10+ years, any programmer can specify any error code they want in their module or script

>This means that error codes are no longer discernible since you don't know if it was a "real" version of that code.

Exactly, and this is exactly the case that we have today. This proposal was trying to add some structure to this to make it easier. It was not trying to be the end-all for this issue, but as a temporary measure until we get the auto-generated `M` to the `E`.

>For example, a user could do abort OUT_OF_GAS wherever they please
>Since they are no longer discernible, you cannot rely on the codes for anything useful (access control, rate limiting, etc). Thus they lose any meaning in the system itself. They could still carry some meaning for debugging purposes, but it is a "weak" meaning, other information would be needed for high debugging assurance

Major status codes (e.g. `ABORTED`, `INVARIANT_VIOLATION`, `OUT_OF_GAS`) are major error codes, have a semantic meaning in the system, and are not able to be expressed from Move. The sub-status codes (which is what the code in abort translates to) don't have any semantic meaning outside of the prologue. These would only be "weak" in the sense of being useful to determine where something went wrong. 


> Therefore, we have arrived at a possible world where just having M+E does not provide the guarantees we want

> I would argue that the current implementation of abort means this M+E proposal can provide no guarantees. So even if I haven't properly represented the guarantees that you want, I posit that any other guarantee would fail under the same test.
> As a followup, I know one intent of this M+E proposal is to make discernible stored error codes. But from the argument above, I think that you lose any of that discernibility with custom code.

Agreed. But I think that's the core thing here. This proposal wasn't trying to setup a system so that we are guaranteed to have a unique error code, but have a programming convention to give us that as a temporary measure until the module-level `M` is given to us automatically by the VM.

> I think this M+E proposal had the right idea of trying to do the same thing with error codes! But as I argued above, the current mechanism is just too brittle.

Agreed that it is brittle, but it is _far_ more robust than the current system today where we hope to not step on other error codes. This is getting at my point of it being a temporary solution. 

Now to your proposal:

> So as a counter proposal I would give is M::E, that is we store (Option<(Module Address, Module Name)>, Error Code) .

I like this idea, I think it pinpoints issues, and makes them easy to find. However, this increases the size of the transaction status considerably, and this is stored for every transaction. The two counter-proposals I have to limit the amount of bloat from this additional information are:
1. The stored error is `(<hash of address::module>, <error number>)`. Script would be represented by a `0` hash value. Agreed about your comment on hashing, but it's pretty easy to build a lookup table of hash to `address::module`. 
2. The stored error is represented as an enum:
```rust
enum TransactionStatus {
    Executed,
    Error(Option<(address, module)>, u64) 
}
```

 
All-in-all I am less convinced of my proposed approach now, and if we can go for the M::E approach where M is some autogenerated unique ID of the module, then I'm happy :)

jolestar:
Can the module's error code be automatically assigned at deployment time?  Then the module developer only needs to be concerned with the error code assignment of the current module, it begin with 1.

PR #4634, opened 2020-06-20
[network] added limit on outgoing dialed peers

gregnazario:
Limiting the number of outgoing dialed peers on full nodes to limit the
amount of work incoming from full nodes.  This includes limiting on
active dials to ensure that no more than 3 connections are made at one
time.

Ideally, in the future we can limit by network type, once we've guaranteed that network IDs are properly set on the networks in the configs.

### DIFF
--- a/network/src/connectivity_manager/mod.rs
+++ b/network/src/connectivity_manager/mod.rs
@@ -42,6 +42,10 @@ use libra_logger::{prelude::*, StructuredLogEntry};
 use libra_network_address::NetworkAddress;
 use libra_types::PeerId;
 use num_variants::NumVariants;
+use rand::{
+    prelude::{SeedableRng, SmallRng},
+    seq::SliceRandom,
+};
 use serde::Serialize;
 use std::{
     cmp::min,
@@ -85,6 +89,10 @@ pub struct ConnectivityManager<TTicker, TBackoff> {
     /// A local counter incremented on receiving an incoming message. Printing this in debugging
     /// allows for easy debugging.
     event_id: u32,
+    /// A way to limit the number of connected peers by outgoing dials.
+    connection_limit: Option<usize>,
+    /// Random for shuffling which peers will be dialed
+    rng: SmallRng,
 }
 
 /// Different sources for peer addresses, ordered by priority (Onchain=highest,
@@ -110,11 +118,12 @@ pub enum ConnectivityRequest {
 }
 
 /// The set of `NetworkAddress`'s for all peers.
+#[derive(Serialize)]
 struct PeerAddresses(HashMap<PeerId, Addresses>);
 
 /// A set of `NetworkAddress`'s for a single peer, bucketed by DiscoverySource in
 /// priority order.
-#[derive(Clone, Default)]
+#[derive(Clone, Default, Serialize)]
 struct Addresses([Vec<NetworkAddress>; DiscoverySource::NUM_VARIANTS]);
 
 #[derive(Debug)]
@@ -151,6 +160,7 @@ where
         requests_rx: channel::Receiver<ConnectivityRequest>,
         backoff_strategy: TBackoff,
         max_delay_ms: u64,
+        connection_limit: Option<usize>,
     ) -> Self {
         {
             // Reconcile the keysets eligible is only used to allow us to dial the remote peer
@@ -201,6 +211,8 @@ where
             backoff_strategy,
             max_delay_ms,
             event_id: 0,
+            connection_limit,
+            rng: SmallRng::from_entropy(),
         }
     }
 
@@ -338,13 +350,26 @@ where
             })
             .collect();
 
+        // Limit the number of dialed connections from a Full Node
+        // This does not limit the number of incoming connections
+        // It enforces that a full node cannot have more outgoing connections than `connection_limit`
+        // including in flight dials.
+        let to_connect_size = if let Some(conn_limit) = self.connection_limit {
+            min(
+                conn_limit - self.connected.len() - self.dial_queue.len(),
+                to_connect.len(),
+            )
+        } else {
+            to_connect.len()
+        };
+
         // The initial dial state; it has zero dial delay and uses the first
         // address.
         let init_dial_state = DialState::new(self.backoff_strategy.clone());
 
-        for (p, addrs) in to_connect.into_iter() {
+        for (p, addrs) in to_connect.choose_multiple(&mut self.rng, to_connect_size) {
             let mut connction_reqs_tx = self.connection_reqs_tx.clone();
-            let peer_id = *p;
+            let peer_id = **p;
             let dial_state = self
                 .dial_states
                 .entry(peer_id)
--- a/network/src/connectivity_manager/test.rs
+++ b/network/src/connectivity_manager/test.rs
@@ -10,13 +10,14 @@ use channel::{libra_channel, message_queues::QueueStyle};
 use core::str::FromStr;
 use futures::SinkExt;
 use libra_config::{config::RoleType, network_id::NetworkId};
-use libra_crypto::{test_utils::TEST_SEED, x25519, Uniform};
+use libra_crypto::{x25519, x25519::PublicKey, Uniform};
 use libra_logger::info;
 use libra_network_address::NetworkAddress;
-use rand::{rngs::StdRng, SeedableRng};
 use std::{io, num::NonZeroUsize};
 use tokio::runtime::Runtime;
 use tokio_retry::strategy::FixedInterval;
+
+const MAX_TEST_CONNECTIONS: usize = 3;
 
 fn setup_conn_mgr(
     rt: &mut Runtime,
@@ -30,20 +31,34 @@ fn setup_conn_mgr(
 ) {
     let network_context =
         NetworkContext::new(NetworkId::Validator, RoleType::Validator, PeerId::random());
+
+    let eligible_peers = eligible_peers
+        .into_iter()
+        .map(|peer_id| {
+            let identity_public_key = x25519::PrivateKey::generate_for_testing().public_key();
+            (peer_id, identity_public_key)
+        })
+        .collect::<HashMap<_, _>>();
+
+    setup_conn_mgr_with_context(network_context, rt, eligible_peers, seed_peers)
+}
+
+fn setup_conn_mgr_with_context(
+    network_context: NetworkContext,
+    rt: &mut Runtime,
+    eligible_peers: HashMap<PeerId, PublicKey>,
+    seed_peers: HashMap<PeerId, Vec<NetworkAddress>>,
+) -> (
+    libra_channel::Receiver<PeerId, ConnectionRequest>,
+    conn_notifs_channel::Sender,
+    channel::Sender<ConnectivityRequest>,
+    channel::Sender<()>,
+) {
     let (connection_reqs_tx, connection_reqs_rx) =
         libra_channel::new(QueueStyle::FIFO, NonZeroUsize::new(1).unwrap(), None);
     let (connection_notifs_tx, connection_notifs_rx) = conn_notifs_channel::new();
     let (conn_mgr_reqs_tx, conn_mgr_reqs_rx) = channel::new_test(0);
     let (ticker_tx, ticker_rx) = channel::new_test(0);
-    let mut rng = StdRng::from_seed(TEST_SEED);
-
-    let eligible_peers = eligible_peers
-        .into_iter()
-        .map(|peer_id| {
-            let identity_public_key = x25519::PrivateKey::generate(&mut rng).public_key();
-            (peer_id, identity_public_key)
-        })
-        .collect::<HashMap<_, _>>();
 
     let conn_mgr = {
         ConnectivityManager::new(
@@ -56,6 +71,7 @@ fn setup_conn_mgr(
             conn_mgr_reqs_rx,
             FixedInterval::from_millis(100),
             300, /* ms */
+            Some(MAX_TEST_CONNECTIONS),
         )
     };
     rt.spawn(conn_mgr.start());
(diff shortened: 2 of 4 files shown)


network/src/connectivity_manager/mod.rs
```diff
@@ -338,6 +340,25 @@ where
...
+        // If there's nothing to dial, let's just skip the rest of the function
+        if to_connect_size == 0 {
+            return;
+        }
+
+        let to_connect = &to_connect[0..to_connect_size];
+
         // We tune max delay depending on the number of peers to which we're not connected. This
         // ensures that if we're disconnected from a large fraction of peers, we keep the retry
         // window smaller.
```
davidiw, line 364: I wonder if this was ever evaluated... it seems highly unlikely that we need this any more... @phlip9 any thoughts if we could delete it? Especially with the recent change that you added. If we cannot connect to a peer, being more aggressive won't make it better...

network/src/connectivity_manager/mod.rs
```diff
@@ -338,6 +340,25 @@ where
...
+        // This does not limit the number of incoming connections
+        // It enforces that a full node cannot have more outgoing connections than `MAX_FULLNODE_CONNECTIONS`
+        // including in flight dials.
+        let to_connect_size = match self.network_context.role() {
+            RoleType::FullNode => min(
+                to_connect.len(),
+                MAX_FULLNODE_CONNECTIONS - self.connected.len() - self.dial_queue.len(),
+            ),
+            _ => to_connect.len(),
+        };
```
davidiw, line 353: this feels loong, I wonder if an if let else would only be 1 line less, I guess :/

gregnazario: Ah, it should be an `if let`

network/src/connectivity_manager/mod.rs
gregnazario: I really wanted to delete it.

Having that kind of strange variability in the timeoff backing causes:
1. operational confusion because you don't know really what's going on without a full picture
2. like you said being more aggressive won't help, especially in a lot of cases you could have some sort of retry storm which might cause things to be worse.

network/src/connectivity_manager/mod.rs
gregnazario: :( I actually doesn't get shorter with an if else, because of the else clause.

network/src/connectivity_manager/mod.rs
gregnazario: Moved to here: https://github.com/libra/libra/issues/4651

network/src/connectivity_manager/mod.rs
```diff
@@ -55,6 +55,8 @@ use tokio::time;
 #[cfg(test)]
 mod test;
 
+const MAX_FULLNODE_CONNECTIONS: usize = 3;
+
```
phlip9, line 59: let's make this a more generic parameter, so you can pass in `Option<usize>` to `ConnectivityManager`, where `None` is no limit and `Some(limit)` will limit the number of connections to `limit`.

network/src/connectivity_manager/mod.rs
```diff
@@ -338,6 +340,26 @@ where
             })
             .collect();
 
+        // Limit the number of dialed connections from a Full Node
```
phlip9, line 343: i don't think this will work, since we'll probably keep trying to dial the same 3 peers even if we can't connect to them.

we should shuffle `to_connect` before slicing (can use https://docs.rs/rand/0.7.3/rand/seq/trait.SliceRandom.html#tymethod.shuffle or https://docs.rs/rand/0.7.3/rand/seq/trait.SliceRandom.html#tymethod.partial_shuffle if you're feeling fancy i guess)

you can setup a SmallRng like how protocol/discovery does it

network/src/connectivity_manager/mod.rs
phlip9: yeah, i think this PR will probably break this and effectively set the backoff super low.

network/src/connectivity_manager/mod.rs
davidiw: Good call! We absolutely do not want to be dialing the same 3 over and over again or across the entire network.

network/src/connectivity_manager/mod.rs
```diff
@@ -338,6 +340,26 @@ where
...
+        // It enforces that a full node cannot have more outgoing connections than `MAX_FULLNODE_CONNECTIONS`
+        // including in flight dials.
+        let to_connect_size = if let RoleType::FullNode = self.network_context.role() {
+            min(
+                to_connect.len(),
+                MAX_FULLNODE_CONNECTIONS - self.connected.len() - self.dial_queue.len(),
+            )
+        } else {
+            to_connect.len()
+        };
```
davidiw, line 354: yeah, not entirely surprised on the size not decreasing :/ but at least now it is more syntactically correct if else is better than a single match else

network/src/connectivity_manager/mod.rs
davidiw: I guess if you do that... then the logic below can change...
```
    let to_connect = min(limit.or_else(to_connect.len()), to_connect.len());
```

gregnazario: > I guess if you do that... then the logic below can change...
> 
> ```
>     let to_connect = min(limit.or_else(to_connect.len()), to_connect.len());
> ```

Sadly no, because then that just limits to "3 dials started at one time" rather than 3 connections total.  Have to add in the calculation including inflight connections and dials.

zekun000: this makes more sense to me, I don't like distinguish by role

gregnazario: It's still random, and there is still a limit by role, but it's controlled outside of the connectivity manager, and shouldn't be exposed there.

network/src/connectivity_manager/mod.rs
gregnazario: It now does slice random, https://docs.rs/rand/0.6.5/rand/seq/trait.SliceRandom.html#tymethod.choose_multiple no need to shuffle

network/src/connectivity_manager/mod.rs
```diff
@@ -105,16 +113,16 @@ pub enum ConnectivityRequest {
     /// Update set of nodes eligible to join the network.
     UpdateEligibleNodes(HashMap<PeerId, x25519::PublicKey>),
     /// Gets current size of dial queue. This is useful in tests.
-    #[serde(skip)]
-    GetDialQueueSize(oneshot::Sender<usize>),
+    GetDialQueueSize(#[serde(skip)] oneshot::Sender<usize>),
```
davidiw, line 116: What's up yo?

network/src/connectivity_manager/mod.rs
```diff
@@ -285,20 +296,27 @@ where
...
-        for p in stale_connections.into_iter() {
-            info!(
-                "{} Should no longer be connected to peer: {}",
-                self.network_context,
-                p.short_str()
-            );
-            // Close existing connection.
-            if let Err(e) = self.connection_reqs_tx.disconnect_peer(p).await {
+
+        if !stale_connections.is_empty() {
```
davidiw, line 300: should we instead check this and return early? excessive nesting makes me sad

network/src/connectivity_manager/mod.rs
```diff
@@ -316,8 +334,14 @@ where
             .filter(|peer_id| !eligible.contains_key(peer_id))
             .cloned()
             .collect();
-        for p in stale_dials.into_iter() {
-            self.dial_queue.remove(&p);
+
+        if !stale_dials.is_empty() {
+            send_struct_log!(StructuredLogEntry::new_named("cancel_stale_dials")
```
davidiw, line 339: same as above
but ... why are we adding this in this commit :O

network/src/connectivity_manager/mod.rs
```diff
@@ -338,6 +362,30 @@ where
...
+        // If there's nothing to dial, let's just skip the rest of the function
+        if to_connect_size == 0 {
+            return;
+        }
+        to_connect.choose_multiple(&mut self.rng, to_connect_size);
+        let to_connect = &to_connect[0..to_connect_size];
+
+        send_struct_log!(StructuredLogEntry::new_named("dial_eligible_peers")
+            .field(&logging::NETWORK_CONTEXT, &self.network_context)
+            .data("new_peers", &to_connect));
```
davidiw, line 387: probably put this before the return statement

network/src/connectivity_manager/mod.rs
```diff
@@ -338,6 +362,30 @@ where
...
+
```
davidiw, line 388: promise me you'll send a pr to delete this blob below <3

network/src/connectivity_manager/mod.rs
```diff
@@ -338,6 +362,30 @@ where
...
+            )
+        } else {
+            to_connect.len()
+        };
+
+        // If there's nothing to dial, let's just skip the rest of the function
+        if to_connect_size == 0 {
+            return;
+        }
+        to_connect.choose_multiple(&mut self.rng, to_connect_size);
```
davidiw, line 382: shouldn't we move this line to the line below in 406

`  for (p, addrs) in to_connect(choose_multiple(&mut self.rng, to_connect_size).iter()`

I think right now, you're effectively randomizing and then throwing away, since the choose_multiple doesn't mutate to_connect.

network/src/connectivity_manager/test.rs
```diff
@@ -1101,3 +1157,75 @@ fn multiple_addrs_shrinking() {
...
+        conn_mgr_reqs_tx
+            .send(ConnectivityRequest::UpdateAddresses(
+                DiscoverySource::Gossip,
+                [(new_peer_id, vec![new_peer_addr])]
+                    .iter()
+                    .cloned()
+                    .collect(),
+            ))
+            .await
+            .unwrap();
```
davidiw, line 1216: why go through this headache and not just have them passed in via the new api / seed_peers?

network/src/connectivity_manager/mod.rs
gregnazario: Oh, `serde` was doing this weird thing complaining that oneshot wasn't serializable that wasn't happening to me before.  When I moved it to just be on the `sender` it stopped complaining, which is strange because if I skipped the whole field, why would it care about the subfield.

This happened in the context of structured logging.  Let me check if it's still happening

network/src/connectivity_manager/mod.rs
gregnazario: Yeah, can do.

network/src/connectivity_manager/mod.rs
gregnazario: I put this after the eligible peers because we aren't dialing any peers.  But, we can put it before and have the `0` show up.

network/src/connectivity_manager/mod.rs
gregnazario: oh, whoops, I thought it said `let to_connect = to_connect.choose_multiple`

network/src/connectivity_manager/mod.rs
davidiw: yeah, otherwise when we log, we wonder ... what /might/ be happening lol

network/src/connectivity_manager/mod.rs
gregnazario: Oh, I'll cut it out :/. I spent a bunch of time chugging through adding this structured logging through my hunt after the seg fault that kept happening during the test.

network/src/connectivity_manager/mod.rs
gregnazario: I will send a separate PR for it!

network/src/connectivity_manager/test.rs
gregnazario: True, I could pass in MAX_CONNS + 1 into the seed peers, and there should only be MAX_CONNS.

I wanted to test a 'realistic' example, of having the max connections set up, and then a new connection shows up

network/src/connectivity_manager/mod.rs
zekun000: oh we have seg fault?

gregnazario: I was getting seg faults during testing either by the structured logging, or by me seemingly dereferencing something in the logic through the channels.  Turning on structured logging made it worse :)

network/src/connectivity_manager/test.rs
gregnazario: Yeah, moved to just seed peers to simplify it.

network/src/connectivity_manager/mod.rs
zekun000: rust claims it'd never seg fault without unsafe code

network/src/connectivity_manager/mod.rs
gregnazario: Here it is!https://github.com/libra/libra/pull/4684

network/src/connectivity_manager/mod.rs
```diff
@@ -338,6 +350,32 @@ where
...
+            .data("new_peers", &to_connect));
+
+        // If there's nothing to dial, let's just skip the rest of the function
+        if to_connect_size == 0 {
+            return;
+        }
+
+        let to_connect: Vec<_> = to_connect
+            .choose_multiple(&mut self.rng, to_connect_size)
+            .collect();
```
phlip9, line 377: if #4684 lands first, you can remove the conditional `if to_connect_size == 0`, since it will just iterate over an empty vec

network/src/connectivity_manager/mod.rs
```diff
@@ -355,9 +393,9 @@ where
         // address.
         let init_dial_state = DialState::new(self.backoff_strategy.clone());
 
-        for (p, addrs) in to_connect.into_iter() {
+        for (p, addrs) in to_connect.iter() {
```
phlip9, line 396: into_iter()

network/src/connectivity_manager/mod.rs
davidiw: Yeah, I would argue that this would be cleaner to just benefit from the fact that this produces an iter and so you could either eliminate the collect or move the function inline into line 345/382 below
no need for the early return

gregnazario: I've gone ahead and put it in the iter.  I've removed the conditional, and the structured log, as it on longer applies to what is actually being connected to.

PR #4638, opened 2020-06-21
[core modules] Remove balances from root, TC, and validator accounts

### LINKED ISSUE #4545: Stdlib cleanup: remove balances from LibraRoot, TC, validator, and validator operator accounts
These accounts should not be able to send/receive money. We can adjust account creation to avoid adding balance resources and (possibly) add a role-based check to `add_currency` to prevent them from adding balances later.

tzakian:
This PR removes the balances from the root, TC, and validator account.

`make_account` no longer creates balances, but this is done in a separate `add_currencies_to_account` function in `LibraAccount`. Additionally, blances are only checked in the prologue/epilogue now if the transaction fee is greater than 0.

Closes #4545


language/stdlib/modules/LibraAccount.move
```diff
@@ -836,15 +849,17 @@ module LibraAccount {
...
 
         // Charge for gas
         let transaction_fee_amount = txn_gas_price * (txn_max_gas_units - gas_units_remaining);
-        assert(
-            balance_for(sender_balance) >= transaction_fee_amount,
-            6
-        );
+        if (transaction_fee_amount > 0)  {
+            // Load the transaction sender's balance resource only if the txn fee is non-zero
+            let sender_balance = borrow_global_mut<Balance<Token>>(sender);
```
sblackshear, line 857: This is currently guaranteed to succeed because of the check in the prologue (right)?

If we ever add functionality for removing a `Balance` resource with value 0 (which seems reasonable enough), I think this will no longer be guaranteed to succeed. It would be nice to rewrite this in a way that ensures the epilogue will fail with 6 even if someone:
- Sends a tx with a nonzero fee in LBR
- Transfers all of their LBR elsewhere
- Deletes their (now-zero) `Balance<LBR>` resource

E.g.,

```
let sender_balance = if (exists<Balance<Token>>(sender)) {
    balance(sender)
} else {
    0
};
assert(balance_for(sender_balance) >= transaction_fee_amount, 6);
```

tzakian: ahhh good point! I'll make that update :) (and check with ops/AOS to make sure it looks good with them).

PR #4640, opened 2020-06-21
[language] Clean up Move VM API

dariorussi:
This removes `cache_module()` from the runtime and VM API which is
a good thing. That entry point had a usage that was testing and
convenience only.
We can replace it everywhere with proper changes to the data
view and with a publish/execute model.
The tricky usage is, as expected, in the genesis generator
but the change is very contained and could be argued the
right thing.

### DIFF
--- a/language/benchmarks/src/move_vm.rs
+++ b/language/benchmarks/src/move_vm.rs
@@ -2,7 +2,6 @@
 // SPDX-License-Identifier: Apache-2.0
 
 use anyhow::Result;
-use bytecode_verifier::VerifiedModule;
 use criterion::Criterion;
 use libra_state_view::StateView;
 use libra_types::{access_path::AccessPath, account_address::AccountAddress};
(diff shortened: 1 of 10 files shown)


language/tools/vm-genesis/src/genesis_context.rs
```diff
@@ -218,7 +231,14 @@ impl<'txn> DataStore for GenesisDataCache<'txn> {
         self.data_store.move_resource_from(ap, ty)
     }
 
+    // Check first if a module reference is in the modules provided at initialization,
+    // otherwise check in the "live" data store.
+    // Normally the only code requested is in the `code_map`, it was initialized
+    // at startup.
     fn load_module(&self, module: &ModuleId) -> VMResult<Vec<u8>> {
```
dariorussi, line 238: this is not the most consistent usage of the data cache.
`load_module` is the only API changed, `exists_module` and `publish_module` ignore the `code_map` entirely.
Given usage of the other API and execution in genesis this is all good, but something to keep in mind.
I don't think is worth to sweat over the risks here, but genesis usage of the VM could be reviewed a bit more.
Right now (and unrelated to this change, it was always there) code is published directly in the data store which means it's not verified by the VM (as a publish call would do). We verify the module upfront but I don't know....
so bottom line, this seems good to me
@runtian-zhou @tzakian @sblackshear let me know if you have thoughts or concerns

language/move-vm/runtime/src/move_vm.rs
```diff
@@ -28,6 +27,7 @@ impl MoveVM {
         function_name: &IdentStr,
         ty_args: Vec<TypeTag>,
         args: Vec<Value>,
+        _sender: AccountAddress,
```
tzakian, line 30: Any reason not to remove the arg?

language/tools/vm-genesis/src/genesis_context.rs
tzakian: I agree with your sentiment that the genesis usage is fine, but that it might be worthwhile taking a look at it and seeing if anything can be tightened up there.

language/benchmarks/src/move_vm.rs
```diff
@@ -16,42 +15,55 @@ use move_lang::{compiled_unit::CompiledUnit, shared::Address};
...
 use std::path::PathBuf;
+use vm::CompiledModule;
 
 /// Entry point for the bench, provide a function name to invoke in Module Bench in bench.move.
 pub fn bench(c: &mut Criterion, fun: &str) {
-    let module = compile_module();
     let move_vm = MoveVM::new();
-    execute(c, &move_vm, module, fun);
+
+    // perform all operations for sender with all 1s
```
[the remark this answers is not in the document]
dariorussi: lol, fair, there is a problem in the benchmark with the compiler (I believe) and I can get things to run only with address 0x0, but I initially had all 1s and it did not work :)
I'll remove the comment and look at some point at the issue but given is the benchmark and I am the only one running it, I figured I'll do it at my leisure...

language/move-vm/runtime/src/move_vm.rs
dariorussi: good question, I am not sure how this whole sender story will end up. Point being that not having the sender everywhere is odd. You can imagine that the semantic is that the client is always responsible for the sender and only when the API takes a sender that is materialized. You could imagine prologue and epilogue or other system API may want a sender. 
So I added for consistency and to review soon with the team just so we think about it and come to a closure

language/benchmarks/src/move_vm.rs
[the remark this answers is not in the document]
dariorussi: oh and let me point out that the benchmark works just fine like this

language/libra-vm/src/libra_vm.rs
```diff
@@ -347,13 +347,19 @@ impl LibraVM {
                     // Module publishing is currently restricted to the Association, so we choose to
                     // publish all modules under the address 0x0...1 for convenience.
                     // This will change to the sender's address once module publishing becomes open.
+                    // REVIEW: should we check that the address of the Module is in fact
+                    // `CORE_CODE_ADDRESS`?
                     let module_address = if self.on_chain_config()?.publishing_option.is_open() {
                         txn_data.sender()
                     } else {
                         account_config::CORE_CODE_ADDRESS
                     };
```
vgao1996, line 356: IMO we don't have to, as we set the sender address to CORE if open publishing is not enabled, which will make the transaction fail as it wasn't signed by CORE.

However I found this a little bit confusing... Shall we stop tweaking the sender address and instead just reject a transaction directly?

dariorussi: I think here we are past the verification step you mention, isn't it?

dariorussi: @runtian-zhou comments?

dariorussi: actually we are performing that after so @vgao1996 is correct that this is not vital at all.
But it's indeed odd we are changing a transaction in the hope it fails that way

PR #4641, opened 2020-06-21
Verify QC and Vote Proposal Block Signature

Rustinante:
## Motivation

- Call `validate_signatures` and `verify_qc` inside `construct_and_sign_vote`.
- Refactored snippets into `verify_qc_proposed_round`.

Yes

## Test Plan
cargo xtest

### DIFF
--- a/consensus/safety-rules/src/safety_rules.rs
+++ b/consensus/safety-rules/src/safety_rules.rs
@@ -75,20 +75,31 @@ impl SafetyRules {
     }
 
     /// This verifies a QC makes sense in the current context, specifically that this is for the
-    /// current epoch and extends from the preffered round.
-    fn verify_qc(&self, qc: &QuorumCert) -> Result<(), Error> {
+    /// current epoch and extends from the preferred round.
+    fn verify_qc(&self, qc: &QuorumCert, start_from_parent: bool) -> Result<(), Error> {
         let validator_verifier = self.verifier()?;
 
         qc.verify(validator_verifier)
             .map_err(|e| Error::InvalidQuorumCertificate(e.to_string()))?;
 
-        if qc.parent_block().round() < self.persistent_storage.preferred_round()? {
-            Err(Error::InvalidQuorumCertificate(
-                "Preferred round too early".into(),
-            ))
+        if start_from_parent {
+            if qc.parent_block().round() < self.persistent_storage.preferred_round()? {
+                return Err(Error::InvalidQuorumCertificate(
+                    "Parent round too old".into(),
+                ));
+            }
         } else {
-            Ok(())
+            let preferred_round = self.persistent_storage.preferred_round()?;
+            if qc.certified_block().round() < preferred_round {
+                debug!(
+                    "Vote proposal certified round is lower than preferred round, {} < {}",
+                    qc.certified_block().round(),
+                    preferred_round,
+                );
+                return Err(Error::ProposalRoundLowerThanPreferredBlock { preferred_round });
+            }
         }
+        Ok(())
     }
 
     /// This reconciles the key pair of a validator signer with a given validator set
@@ -204,7 +215,7 @@ impl TSafetyRules for SafetyRules {
     /// Verify the QC is correct and up to date, if it is either set the preferred round or start a
     /// new epoch.
     fn update(&mut self, qc: &QuorumCert) -> Result<(), Error> {
-        self.verify_qc(qc)?;
+        self.verify_qc(qc, true)?;
         if qc.ends_epoch() {
             self.start_new_epoch(qc.ledger_info().ledger_info())
         } else {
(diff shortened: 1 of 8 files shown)


consensus/safety-rules/src/safety_rules.rs
```diff
@@ -233,16 +245,11 @@ impl TSafetyRules for SafetyRules {
...
-            debug!(
-                "Vote proposal certified round is lower than preferred round, {} < {}",
-                proposed_block.quorum_cert().certified_block().round(),
-                preferred_round,
-            );
-            return Err(Error::ProposalRoundLowerThenPreferredBlock { preferred_round });
-        }
+        let qc = proposed_block.quorum_cert();
+        self.verify_qc(qc)?;
+        self.verify_qc_proposed_round(qc)?;
```
davidiw, line 252: I don't think there should be any issue from merging these together. In fact, we're planning on merging update into construct_and_sign_vote and sign_proposal to simplify the API. cc: @zekun000 

if you wanna add a commit to this stick to remove update and adjust those two calls, that would be great.

consensus/safety-rules/src/safety_rules.rs
```diff
@@ -233,16 +245,11 @@ impl TSafetyRules for SafetyRules {
                 last_voted_round: self.persistent_storage.last_voted_round()?,
             });
         }
+        proposed_block.validate_signatures(self.verifier()?)?;
```
davidiw, line 248: we could also add to this stack a commit that renames validate_signatures to validate_signature, I don't know why this was ever named to be plural.

consensus/safety-rules/src/safety_rules.rs
Rustinante: Yes, and I'm actually a little confused about the second check. In the first check we made sure that the parent's round is at least the preferred round, and in the second check it checks that the proposed round is at least the preferred round. But wouldn't the proposed round be 1 larger than the parent round?

Rustinante:
Sure! And I think some tests are failing on the newly added check_qc call... I'm trying to figure out why by reading the `State Machine Replication in the Libra Blockchain` document...

davidiw:
If you look at the code, you'll see update for b3 has unwrap_err, so it is already expecting that the QC update will fail... in fact, this seems a little strange that we would sign this at all...

@zekun000, it might be worth thinking through what type of attack might have happened here, if any at all... I suspect this is more likely just calling out some poorly formed chain, e.g., something that shouldn't have happened naturally. but it is a little late, so I might be rushing to a conclusion.

consensus/safety-rules/src/safety_rules.rs
davidiw: So that makes it pretty easy to merge right?

* QuorumCert::verify should ensure that the parent round < certified round, which it currently does not
* SafetyRules::verify_qc should ensure that parent round >= preferred round (which it currently does)

zekun000: `SafetyRules::verify_qc` should only verify the signatures if we want to use it both in `update` and `vote` because
- `update` requires `parent_round > preferred_round` - this is 2-chain
- `vote` requires `certified_round > preferred_round` - this is 1-chain

davidiw:
Sorry, @zekun000 was far more excited to work on this than I anticipated :O.

So what's left is to simply verify the signature on the proposal. We did internally discuss today adding more counters to monitor the state of the various values in consensus state. There are already some examples in the code and it would be super useful if you could do that. Alternatively the perf work is also pretty important!

davidiw:
> If you look at the code, you'll see update for b3 has unwrap_err, so it is already expecting that the QC update will fail... in fact, this seems a little strange that we would sign this at all...
> 
> @zekun000, it might be worth thinking through what type of attack might have happened here, if any at all... I suspect this is more likely just calling out some poorly formed chain, e.g., something that shouldn't have happened naturally. but it is a little late, so I might be rushing to a conclusion.

Also following up here... this could potentially happen either by a proposer being left in the dark about the current state of the chain or the last proposer not broadcasting out their proposal to everyone. The implications of modifying behavior could impact liveness. Ultimately we are reducing the depth of our checks to focus on what the actual safety requirements are versus more restrictive that might come at the cost of actually impacting liveness.

Rustinante:
I see, so what should we do about the failing test? Should we simply remove the `vote = safety_rules.construct_and_sign_vote(&b3).unwrap();`?

consensus/safety-rules/src/safety_rules.rs
Rustinante: Which signature is `SafetyRules::verify_qc` verifying?

Rustinante:
Should this throw an error? 
https://github.com/libra/libra/blob/master/consensus/safety-rules/src/safety_rules.rs#L91

Rustinante:
Okay I've create a PR just for verifying the signature #4680

davidiw:
closing this out since the work was done in two separate PRs.

PR #4642, opened 2020-06-21
[config] Improve debuggability and other stuff

davidiw:
This is a cross between cleaning up legacy code and inserting more useful error messages.

here's a quote from one of the commits:

anyhow is great to throw something together, but without diligence it
becomes pretty useless except for error propagation. Specifically, it is
super easy to just add `?` or `map_err(|e| e.into())` everywhere and not
really annotate errors with enough information. Secondly, even if
properly annotated, the error messages will not be uniform, so what
might be an IO error in one context could be listed as a access error
elsewhere but still fundamentally the same thing. Again, anyhow supports
intelligent usage, but it only comes with discipline and there's no way
to enforce it after leaving a crate. The better solution for strictness
is to use a dedicated error type as done in this PR.

The big win of this PR is that when something bad happens, it should be
very clear where that bad thing is happening... no longer will we get
unclear errors at least from within the config folder.

Also Rust is shortly going to enable backtraces in errors, that will
make it even better... oh happy day...


config/src/config/consensus_config.rs
```diff
@@ -19,14 +19,14 @@ pub struct ConsensusConfig {
 impl Default for ConsensusConfig {
     fn default() -> ConsensusConfig {
         ConsensusConfig {
+            contiguous_rounds: 2,
```
mimoo, line 22: btw do you know why we have this as part of the consensus config o_X ?

config/src/config/consensus_config.rs
```diff
@@ -8,8 +8,8 @@ use std::path::PathBuf;
 #[derive(Clone, Debug, Deserialize, PartialEq, Serialize)]
 #[serde(default, deny_unknown_fields)]
 pub struct ConsensusConfig {
-    pub max_block_size: u64,
     pub contiguous_rounds: u32,
+    pub max_block_size: u64,
```
mimoo, line 12: IIUC by swapping these, but not swapping these in config files that we already have, we now have configs that will load as:

* contiguous_rounds = previous_max_block_size
* max_block_size = previous_contiguous_round

and this is in part facilitated by the fact that u32 and u64 are (de)serialized the same way.

That's probably the only danger there is in refactoring the config module.

config/Cargo.toml
```diff
@@ -10,7 +10,6 @@ publish = false
 edition = "2018"
 
 [dependencies]
-anyhow = "1.0.31"
```
mimoo, line 13: bold

mimoo: but yeah IIUC a library should use thiserror

mimoo:
question for you: would it make sense to actually convert all errors to crashes in config? My (perhaps flawed) understanding is that we always end up not going on if there's an error returned. In this case, we could greatly simplify error handling in this crate by just unwrapping everything.

That is, unless we load configs dynamically while running a node to detect config changes :)

mimoo:
@bors-libra delegate+

you now have the power to merge, but with with great power comes great responsibility.

config/src/config/consensus_config.rs
davidiw: pre-dates me, I think there is some change in the guarantees by shrinking this or increasing it, but without thinking very hard.... @zekun000 / @aching do you want to eliminate this variable?

config/src/config/consensus_config.rs
davidiw: that's not the case, all the configs we have are named (yaml) versus lcs. If we did use lcs or some other binary / anonymous field-based config, you are 100% right.

config/Cargo.toml
davidiw: yeah, not sure if you read the commit or description, but anyhow is really awesome for prototyping and not falling back on expect / panic / unwrap, but then you end up in a world of hurt because it is too easy to use and hard to get right.

davidiw:
> question for you: would it make sense to actually convert all errors to crashes in config? My (perhaps flawed) understanding is that we always end up not going on if there's an error returned. In this case, we could greatly simplify error handling in this crate by just unwrapping everything (and we could also print a good error message).
> 
> That is, unless we load configs dynamically while running a node to detect config changes :)

I went back and forth on this for a while. When I first joined, this was all expect / unwrap driven and I converted to Results. I still think that's the right behavior, because this is a library and not the main function to an application. Fwiw, every application .expects to load a node config and blows up upon failure.

Once errors have backtraces, I think this is coming by end of August, the user experience will be identical.

From a programmer perspective, beyond the fact that this is a library, errors also help unify and enable better logging. If we use panic, the only way to get this to logging is via the panic handler, which itself may have issue, I don't know what we get at that point in time.

davidiw:
@mimoo thanks for the review!

config/src/config/consensus_config.rs
mimoo: oh my brain did fart indeed

config/src/config/consensus_config.rs
aching: It's the number of contiguous proposal rounds for a rotating proposer.  E.g. proposer X does rounds Y and Y+1, proposer X+1 does rounds Y+2 and Y+3.  It isn't used in the current LeaderReputation mode, but something we might want to in the future.

davidiw: Oh, in that case, it should be part of the proposer type or completely eliminated (as it isn't being used :))

Issue #4635, opened 2020-06-20
[Feature Request][consensus] error handling between consensus and safety rules


zekun000:
we assumed safety rules are infallible, the error handling is basically `?` operator. now it's time to distinguish recoverable/unrecoverable errors and handle them correspondingly.

PR #4655, opened 2020-06-22
[safety-rules] eliminate update

zekun000:
## Motivation

The preferred round update changes to construct_and_sign_vote in this
commit.

Y
## Test Plan

Existing tests.

## Related PRs
https://github.com/libra/libra/issues/4635

### DIFF
--- a/consensus/safety-rules/src/safety_rules.rs
+++ b/consensus/safety-rules/src/safety_rules.rs
@@ -13,9 +13,9 @@ use libra_crypto::{ed25519::Ed25519Signature, hash::HashValue};
 use libra_logger::debug;
 use libra_types::{
     block_info::BlockInfo, epoch_change::EpochChangeProof, epoch_state::EpochState,
-    ledger_info::LedgerInfo, validator_signer::ValidatorSigner,
-    validator_verifier::ValidatorVerifier, waypoint::Waypoint,
+    ledger_info::LedgerInfo, validator_signer::ValidatorSigner, waypoint::Waypoint,
 };
+use std::cmp::Ordering;
 
 /// SafetyRules is responsible for the safety of the consensus:
 /// 1) voting rules
@@ -26,7 +26,7 @@ use libra_types::{
 pub struct SafetyRules {
     persistent_storage: PersistentSafetyStorage,
     validator_signer: Option<ValidatorSigner>,
-    validator_verifier: Option<ValidatorVerifier>,
+    epoch_state: Option<EpochState>,
 }
 
 impl SafetyRules {
@@ -36,7 +36,7 @@ impl SafetyRules {
         Self {
             persistent_storage,
             validator_signer: None,
-            validator_verifier: None,
+            epoch_state: None,
         }
     }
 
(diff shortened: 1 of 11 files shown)


consensus/safety-rules/src/safety_rules.rs
```diff
@@ -195,32 +190,37 @@ impl TSafetyRules for SafetyRules {
...
+        }
+        if proposed_block.quorum_cert().certified_block().round() < preferred_round {
+            debug!(
+                "Vote proposal certified round is lower than preferred round, {} < {}",
+                proposed_block.quorum_cert().certified_block().round(),
+                preferred_round,
+            );
+            return Err(Error::ProposalRoundLowerThenPreferredBlock { preferred_round });
+        }
+        self.signer()?;
```
davidiw, line 222: Seems reasonable, since we are now using this as an update method, but one could argue, this isn't simpler :P

consensus/safety-rules/src/safety_rules.rs
```diff
@@ -195,32 +190,37 @@ impl TSafetyRules for SafetyRules {
...
+                .set_preferred_round(proposed_block.quorum_cert().parent_block().round())?;
+        }
+        if proposed_block.quorum_cert().certified_block().round() < preferred_round {
+            debug!(
+                "Vote proposal certified round is lower than preferred round, {} < {}",
+                proposed_block.quorum_cert().certified_block().round(),
+                preferred_round,
+            );
+            return Err(Error::ProposalRoundLowerThenPreferredBlock { preferred_round });
+        }
```
davidiw, line 221: Recommend we do error checking first. It would be pretty weird that we update our preferred round and then error our because the certified round is older :O

Also as I mentioned in the other pr, this should just go into verify_qc replacing what you have there.

davidiw: also could we add some white space between the if statements?

consensus/safety-rules/src/safety_rules.rs
zekun000: I probably can split this into smaller functions

consensus/safety-rules/src/safety_rules.rs
davidiw: one last ask, could we add a log statement for the case where we find an older parent block then preferred round? it is an unexpected and uncommon event, so it would be good to note that we're in such a state if we have issues getting blocks committed.

consensus/safety-rules/src/safety_rules.rs
```diff
@@ -46,12 +49,75 @@ impl SafetyRules {
...
+            return Err(Error::OldProposal {
+                proposal_round: proposed_block.round(),
+                last_voted_round: self.persistent_storage.last_voted_round()?,
+            });
+        }
+        Ok(())
+    }
+
+    /// Second voting rule
+    fn preferred_round_check(&mut self, quorum_cert: &QuorumCert) -> Result<(), Error> {
```
davidiw, line 76: verify_and_update or check_and_update

davidiw: should also update sign_timeout

zekun000: there's no qc in sign_timeout

davidiw: sorry comment was intended to go to the last voted round

PR #4657, opened 2020-06-22
[core modules/types] Move empty script binary from stdlib to types/test_helpers

### LINKED ISSUE #4620: [Stdlib cleanup] Remove `empty_script` from the stdlib, and move in to the e2e tests.

tzakian:
The empty script wasn't being used outside of the test helpers in `libra-types` so this moves the script to there.

One unfortunate aspect is if the binary format changes, this binary will need to be changed. But, the file format should be stable at this point, so not expecting this to be an issue.

closes #4620


dariorussi:
check out the lint error obviously :)

matbd:
> what is the empty script for?

exactly :)

tzakian:
Empty script is only used for testing purposes so it shouldn't be included in the `transaction_scripts` (and therefore the allowed set of transactions...).

PR #4660, opened 2020-06-22
[types/vm] Move `vm_errors` to `move-core-types`. Remove `vm` dep on `libra-types`. 

tzakian:
Moves the `vm_errors` to the `move-core-types` crate, and then re-imports this back in to `libra-types` (just like `AccountAddress`). This also removes the dependency between the VM and libra-types. 

All tests should pass

This is stacked on #4656 so only look at the top commit (tagged `types/vm`)


language/vm/src/errors.rs
```diff
@@ -2,23 +2,11 @@
...
 
 use crate::IndexKind;
-use libra_types::{
+use move_core_types::{
     account_address::AccountAddress,
-    transaction::TransactionStatus,
     vm_error::{StatusCode, VMStatus},
 };
 
-// We may want to eventually move this into the VM runtime since it is a semantic decision that
```
dariorussi, line 11: do we need the prologue errors here, they hardly are a VM thing.... 
it would be good to remove this file entirely

tzakian: Good point. I think these should be moved. I'll move these to `vm_errors` if that sounds good to you?

dariorussi: I think it's more what it dons good to you ;)
this file for me was always odd and it never had anything VM strictly speaking. `Location` will be removed soon and you can stash it anywhere.
Another PR fine too to me. 
Thanks!

tzakian: I'll put it up in another PR and scatter the functions and whatnot around appropriately :)

PR #4664, opened 2020-06-22
[build] Refactor common build cmd line into signle shell script

andll:
We had same long build string and RUSTFLAGS scattered across many Dockerfiles.

This makes update of docker files comlicated and increases chance of an error(for example, skipping some rust flag or not building some binary)

It is also important that this script is same across all dockerfiles, as otherwise layer caching is not working effectively

tested: ./docker/validator/build.sh locally, land blocking will test other builds


docker/build-common.sh
```diff
@@ -0,0 +1,17 @@
+#!/bin/bash
```
mimoo, line 1: since this is not necessarily a docker thing, I suggest having this somewhere else (for example `scripts/`) and renaming to something that indicate this is for building in release mode (`release-build.sh`)

rexhoffman: I think there's value in that.   At the very least we could run this in circle (w/o docker) to verify the builds in prs, or should volume mounts speed it up enough w/ docker - should know today.

andll:
@sausagee note that `libra-management` is a different build command line(only builds libra-management and nothing else), and everything else does not build libra-management. I don't mind adding it, but just wanted to make sure that we understand that this was different from other docker files

davidiw:
> @sausagee note that `libra-management` is a different build command line(only builds libra-management and nothing else), and everything else does not build libra-management. I don't mind adding it, but just wanted to make sure that we understand that this was different from other docker files

I'm okay with this. @mgorven just in case...

docker/build-common.sh
andll: @mimoo this kind of _is_ a docker thing - this script generalizes build cmd line across all docker files.
I don't want to argue too much on where to put the file though -  I don't think calling it from docker or scripts makes a difference

mimoo: no biggie, I was just thinking of people who want to build a release and run it without docker : o

andll:
Looks like compilation in Circle with this change takes longer (while aws-build takes same time)... I'll look, but let me know if you have idea why this could be

andll:
Ok, so I think any change to docker file is going to fail `build_docker` stage in circle.
It works normally, because `should_build_docker.sh` terminates it early on regular PRs, but otherwise looks like something regressed in docker build in circle, so attempt to actually invoke this step fails.

I don't think this is related to change in this PR.

Question it begs is - do we still need `build_docker` stage in Circle? We check the build anyway in build-aws during land blocking test

@sausagee / @rexhoffman ?

sausagee:
The docker build in CI failed due to the build step exceeds the no output timeout (10 min).  It was unrelated to this PR.  The linker doesn't generate  intermediates to stdout/stderr, so CI host interpreted it as stuck.

Chatted offline with @andll, it's fine to remove the check in Circle and move it to LBT.

andll:
Docker build has lint stage, which I think we should keep - renamed stage build_docker->lint_docker and removed build step from there

@bors-libra r=sausagee

Issue #4666, opened 2020-06-22
[Feature Request] Boogie model for cryptographic functions


emmazzz:
A lot of stdlib modules depends on cryptographic modules like `Signature`. We need to think about how we want to model the functions inside including `ed25519_validate_pubkey`, `ed25519_verify`, `ed25519_threshold_verify`.

PR #4667, opened 2020-06-23
[move prover] add specs to VASP

emmazzz:
## Motivation

This PR adds specifications to `VASP.move`. Verification is turned off because we need to yet implement `Signature`(issue #4666 ). I have also added assertions asserting that a VASP cannot be a ChildVASP and ParentVASP at the same time.

Yes.

## Test Plan

cargo test


language/stdlib/modules/VASP.move
```diff
@@ -174,6 +180,138 @@ module VASP {
...
+
+    spec module {
+        /// > TODO(emmazzz): verification is turned off because `Signature` module
+        /// > has not been specified. See issue #4666.
+        pragma verify = false;
+    }
+
+    spec module {
+        /// Returns true if `addr` is a VASP.
+        define spec_is_vasp(addr: address): bool {
```
DavidLDill, line 195: This could use spec_is_parent_vasp and spec_is_child_vasp, to reduce redundancy (suppose someone changes the test in spec_is_parent_vasp?)

language/stdlib/modules/VASP.move
```diff
@@ -174,6 +180,138 @@ module VASP {
...
+        apply NumChildrenRemainsSame to * except publish_child_vasp_credential;
+    }
+
+    /// ## Specifications for individual functions
+    spec schema AbortsIfNotVASP {
+        addr: address;
+        aborts_if !spec_is_vasp(addr);
+    }
+
+    spec module {
```
DavidLDill, line 259: This is an interesting style.  I think Wolfgang used "include" in spec fun <whatever>  to reduce repetition.  This is kind of like the transpose of that (specs -> functions, not functions -> specs).  I don't know which is better.

PR #4668, opened 2020-06-23
[network] NetworkContext in NewPeer

davidiw:


network/builder/src/builder.rs
```diff
@@ -170,8 +170,8 @@ impl NetworkBuilder {
         }
     }
 
-    pub fn network_context(&self) -> &NetworkContext {
-        &self.network_context
+    pub fn network_context(&self) -> Arc<NetworkContext> {
+        self.network_context.clone()
```
gregnazario, line 174: Hehe, I remember I was asked specifically to change this to how it was.  Makes sense this way

davidiw:
> Looks good, how do the log lines end up as a result of putting the Arc around it? Or I guess some of these are files that don't have network context in the logs. And will it serialize correctly for structured logs?

Likely one of two ways, either [deref](https://doc.rust-lang.org/beta/std/ops/trait.Deref.html) or more likely as a pass through via display / debug: https://doc.rust-lang.org/std/sync/struct.Arc.html#impl-Display

PR #4670, opened 2020-06-23
[libra-trace] read libra trace from elastic search

changyonghu:
Read libra trace from elastic search instead of debug_interface. Use --use-logs-for-trace in cluster test to print traces in console. Currently not all events are seen in elastic search, use --use-logs-for-trace may not capture an entire trace point.

### DIFF
--- a/common/debug-interface/src/trace.rs
+++ b/common/debug-interface/src/trace.rs
@@ -1,0 +1,125 @@
+// Copyright (c) The Libra Core Contributors
+// SPDX-License-Identifier: Apache-2.0
+
+#![forbid(unsafe_code)]
+
+use crate::json_log::JsonLogEntry;
+use anyhow::Result;
+use chrono::{DateTime, Duration, Utc};
+use reqwest::Client;
+use serde::Deserialize;
+use serde_json::json;
+
+const TRACE_EVENT: &str = "trace_event";
+const TRACE_EDGE: &str = "trace_edge";
+const PAGING_SIZE: usize = 10000;
+
+#[derive(Deserialize)]
+struct Response {
+    #[serde(rename = "_scroll_id")]
+    scroll_id: String,
+    hits: Hits,
+}
+
+#[derive(Deserialize)]
+struct Hits {
+    hits: Vec<Hit>,
+}
+
+#[derive(Deserialize)]
+struct Hit {
+    #[serde(rename = "_source")]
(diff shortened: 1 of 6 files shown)


common/debug-interface/src/trace.rs
```diff
@@ -0,0 +1,106 @@
...
+                    }
+                }
+            }
+            ))
+            .send()
+            .await?;
+
+        let response: Value = response.json().await?;
+
+        let mut id = response["_scroll_id"].clone();
```
andll, line 60: I think this will panic if response does not have _scroll_id.
You might want to use some method that returns Option (like .get) and return error if _scroll_id is not found, instead of panic

common/debug-interface/src/trace.rs
```diff
@@ -0,0 +1,106 @@
...
+            let hits = response["hits"]["hits"].as_array().unwrap().len();
+            if hits < PAGING_SIZE {
+                break;
+            }
+        }
+        Ok(v)
+    }
+}
+
+fn parse_response(v: &mut Vec<(String, JsonLogEntry)>, response: &Value) -> Result<()> {
```
andll, line 88: You need to change this function so it does not panic if something is wrong with response and return error instead.

One interesting option here is to use serde to actually deserialize response into some struct. For example, you can define
```
#[derive(Deserialize)]
struct Response {
hits: Hits
}
#[derive(Deserialize)]
struct Hits {
hits: Vec<Hit>
}
#[derive(Deserialize)]
struct Hit {
#[serde(rename="_source")]
source: Source;
}
#[derive(Deserialize)]
struct Source {
#[serde(rename="kubernetes.pod_name")]
pod_name: String,
data: serde_json::Value,
}
```

And after this you can use deserialization from json to verify response structure in one line

andll:
Overall looks good, only concern I have right now is that we don't want to panic if we see some unexpected response from ES

common/debug-interface/src/trace.rs
changyonghu: Square brackets can also be used to index into a value in a more concise way. This returns Value::Null in cases where get would have returned None. response["_scroll_id"] will return Value::Null if not found I think. But we can make it return error.

andll: Ok, if you confirmed that `[]` does not panic, but return value::Null then we are good

Issue #2338, opened 2020-01-28
[Client] Include block metadata on querying TXNs


thefallentree:
# 🚀 Feature Request

Provide additional block metadata information  with proof when client query for TXNs . 

The current useful information includes:  timestamp and version .

## Motivation

Right now, when using UpdateToLatestLedger(GetTransactionsRequest()),  the client is able to retreive an list of transactions , however,  there are no version and timestamp metadata associated with the transaction. To infer these information, client needs to rely on the fact that UserTranscation Always follows after an BlockMetaData Txn (prologue) and have to scan the list of transactions so it could use the information carried in that TXN to infer when and what version is for an given transaction. 

This is not ideal, and not secure, and not provable. 

## Pitch

Storage should return additional proofs for that transaction that includes block metadata information for each transaction, which should also be signed by validators with verifiable signatures.

specifically, adding more information and extend this proto

https://github.com/libra/libra/blob/9d1104f20a19efc83838acbfc396738ab0b61035/types/src/proto/transaction.proto#L88

to carry  metadata information for each transaction with individual proofs.

thefallentree:
Hi guys, any thoughts on this?

msmouse:
@thefallentree  Thank or bringing this up.
1. the version is included in `TransactionWithProof` and is verifiable, see `TransactionWithProof::verify_user_txn`
2. IIRC to prove and verify the timestamp, we planned to rely on an event stream emitted with each of the blockmeta type transactions to "chain" the blocks with consecutive sequence numbers. @sblackshear  is that already being emitted?

sblackshear:
That is not being emitted, but we can easily do that. I will add this.

msmouse:
I'm on back on this topic and realize it's probably worth discussion to put `height` back to the `BlockMetadata` transaction (then we can prove timestamps based on transactions only, instead of relying on events.) So please hold on committing #2389.

And I see `height` is tracked by `Resource BlockMetadata` today in Move (althought there's a `// TODO: Should we keep the height?` one line above), I'm not sure if it complicates things by adding the same info to the transaction.

Thoughts @sblackshear @zekun000  @dmitri-perelman?

msmouse:
Consensus side discussion wrt this: #2616

msmouse:
Now that #2955 has implemented `NewBlockEvent` which carries the timestamp, we can build the functionality around it.

[...]

PR #4673, opened 2020-06-23
Make get_metadata working on non-metadata txns

thefallentree:
In this implementation, we get the block timestamp by looking for immediate preceding blockmetadata txn in storage, and use the timestamp in it.  This relies on the fact that blockmetadata txn always happens before the user transaction, which is an very strong guarantee. 

Also, since this is only used in json-rpc non-verifiable api, no need to bother include proofs.

Ran an local swarm ,  and submit an mint transaction at version 590

```
Transaction at version 589: TransactionView {
    version: 589,
    transaction: BlockMetadata {
        timestamp_usecs: 1592894528634709,
    },
    hash: "a36921a7",
    events: [],
    vm_status: EXECUTED,
    gas_used: 600,
}
Transaction at version 590: TransactionView {
    version: 590,
    transaction: UserTransaction {
        sender: "0000000000000000000000000b1e55ed",
        signature_scheme: "Scheme::Ed25519",
        signature: "eba20fc4a3175a78fd6b305f13b20d279a04f6218e0b658e98dec0bdcb39e5317a3804d9ed3004381beb7a888d7447e864a5abd5e5c0a09a2bc22a3a46440505",
        public_key: "340db4a117a835557b11c6f41962dc118df9b84d4ef78757614d77f4265ca210",
        sequence_number: 0,
        max_gas_amount: 1000000,
        gas_unit_price: 0,
        gas_currency: "LBR",
        expiration_time: 1592894628,
        script_hash: "a121d36975ed76634aafc2adf18cbd5c230b35e22953e72d2032e8855d1af4da",
        script: Unknown,
    },
    hash: "980054f8",
    events: [],
    vm_status: EXECUTED,
    gas_used: 555684,
}
Transaction at version 591: TransactionView {
    version: 591,
    transaction: BlockMetadata {
        timestamp_usecs: 1592894529032495,
    },
    hash: "8aa172cb",
    events: [],
    vm_status: EXECUTED,
    gas_used: 600,
}

```

verified version 590 shows same timestamp as 589
```
fallentree-mbp:libra fallentree$ curl -X POST -H "Content-Type: application/json" --data '{"jsonrpc":"2.0","method":"get_metadata","params":[589],"id":1}' http://localhost:52009
{"id":1,"jsonrpc":"2.0","result":{"timestamp":1592894528634709,"version":589}}fallentree-mbp:linrpc":"2.0","method":"get_metadata","params":[590],"id":1}' http://localhost:52009
{"id":1,"jsonrpc":"2.0","result":{"timestamp":1592894528634709,"version":590}}fallentree-mbp:linrpc":"2.0","method":"get_metadata","params":[591],"id":1}' http://localhost:52009
{"id":1,"jsonrpc":"2.0","result":{"timestamp":1592894529032495,"version":591}}fallentree-mbp:libra fallentree$ 

```

### DIFF
--- a/storage/libradb/src/lib.rs
+++ b/storage/libradb/src/lib.rs
@@ -46,7 +46,7 @@ use crate::{
     system_store::SystemStore,
     transaction_store::TransactionStore,
 };
-use anyhow::{ensure, Result};
+use anyhow::{ensure, format_err, Result};
 use itertools::{izip, zip_eq};
 use jellyfish_merkle::{restore::JellyfishMerkleRestore, TreeReader, TreeWriter};
 use libra_crypto::hash::{CryptoHash, HashValue, SPARSE_MERKLE_PLACEHOLDER_HASH};
@@ -67,8 +67,8 @@ use libra_types::{
         SparseMerkleRangeProof, TransactionListProof,
     },
     transaction::{
-        TransactionInfo, TransactionListWithProof, TransactionToCommit, TransactionWithProof,
-        Version, PRE_GENESIS_VERSION,
+        Transaction, TransactionInfo, TransactionListWithProof, TransactionToCommit,
+        TransactionWithProof, Version, PRE_GENESIS_VERSION,
     },
 };
 use once_cell::sync::Lazy;
@@ -672,6 +672,26 @@ impl DbReader for LibraDB {
         };
 
         Ok(tree_state)
+    }
+
+    fn get_block_timestamp(&self, version: u64) -> Result<u64> {
+        // Maximum TPS from benchmark is around 1000.
+        const MAX_VERSIONS_TO_SEARCH: usize = 1000 * 3;
+
+        let lower_bound = std::cmp::max(0, version - MAX_VERSIONS_TO_SEARCH as u64);
+
+        // rely on the fact that blockmetadata txn always precedes user txn.
+        for i in (lower_bound..=version).rev() {
+            if let Transaction::BlockMetadata(t) = self.transaction_store.get_transaction(i)? {
+                return Ok(t.into_inner()?.1);
+            } else if i == 0 {
+                // genesis timestamp is 0
+                return Ok(0);
+            }
+        }
+        Err(format_err!(
+            "Unable to find preceding blockmetadata transaction"
+        ))
     }
 }
 
--- a/storage/storage-client/src/lib.rs
+++ b/storage/storage-client/src/lib.rs
@@ -173,6 +173,10 @@ impl DbReader for StorageClient {
     fn get_epoch_ending_ledger_info(&self, _: u64) -> Result<LedgerInfoWithSignatures> {
         unimplemented!()
     }
+
+    fn get_block_timestamp(&self, _version: u64) -> Result<u64> {
+        unimplemented!()
+    }
 }
 
 impl DbWriter for StorageClient {
--- a/json-rpc/src/methods.rs
+++ b/json-rpc/src/methods.rs
@@ -129,19 +129,16 @@ async fn get_account_state(
 /// returning the current blockchain metadata
 /// Can be used to verify that target Full Node is up-to-date
 async fn get_metadata(service: JsonRpcService, request: JsonRpcRequest) -> Result<BlockMetadata> {
-    let (version, timestamp) = match serde_json::from_value::<u64>(request.get_param(0)) {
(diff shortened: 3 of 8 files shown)


storage/libradb/src/lib.rs
```diff
@@ -673,6 +673,17 @@ impl DbReader for LibraDB {
...
+    fn get_block_timestamp(&self, version: u64) -> Result<u64> {
+        // rely on the fact that blockmetadata txn always precedes user txn.
+        for i in (1..=version).rev() {
+            if let Transaction::BlockMetadata(t) = self.transaction_store.get_transaction(i)? {
+                return Ok(t.into_inner()?.1);
+            }
+        }
+        // genesis timestamp is 0
+        Ok(0)
+    }
```
msmouse, line 686: @christinacelee  what's the use case for this API? Will it be called much? Is linear search like this acceptable performance wise?

msmouse: (I can try to expose the .prev() interface on the iterator though, which will eliminate most of the random IOs here.)

metajack: .prev() still makes it a linear search, just with a much smaller space to traverse. Do we not have an index of closest metadata block?

msmouse: I think the iterator loads in blocks, and .prev() won't issue another IO until the loaded block of bytes are exhausted. 

The index needed is described here https://github.com/libra/libra/issues/2338#issuecomment-606211862, we may implement it if use case requires it.

christinacelee: this API will probably used most for querying for latest version to check how up-to-date your FN is on sync status, querying past versions will probably be called even less. But since this is a publicly open endpoint (not just something we're using internally), I don't think we should settle for this solution

msmouse: We agreed offline that linear search via a proper db iterator is probably good considering size of the RocksDB data block, the size of a avg transaction and number of transactions in a libra block. Let's land this and (I'll) upgrade it by exposing `.prev()` on the iter.

storage/libradb/src/lib.rs
```diff
@@ -673,6 +673,17 @@ impl DbReader for LibraDB {
 
         Ok(tree_state)
     }
+
+    fn get_block_timestamp(&self, version: u64) -> Result<u64> {
+        // rely on the fact that blockmetadata txn always precedes user txn.
+        for i in (1..=version).rev() {
```
msmouse, line 679: Let's give it a bound by introducing a `MAX_VERSIONS_TO_SEARCH_FOR_BLOCK_META`, or similar.

storage/storage-client/src/lib.rs
```diff
@@ -173,6 +173,10 @@ impl DbReader for StorageClient {
     fn get_epoch_ending_ledger_info(&self, _: u64) -> Result<LedgerInfoWithSignatures> {
         unimplemented!()
     }
+
+    fn get_block_timestamp(&self, _version: u64) -> Result<u64> {
```
msmouse, line 177: what else in the block metadata we are potentially exposing? like the `round`, the `proposer`, etc? @metajack , any ideas?

I think we'd better expose them all at the storage level and probably trim the size down at the json-rpc level.

christinacelee: at least for JSON RPC, currently there are no plans to expose other block info (e.g. `round`, `proposer`) in block metadata, but other use cases (not just JSON RPC) might come up in the future?

storage/libradb/src/lib.rs
thefallentree: sounds good!

storage/libradb/src/lib.rs
thefallentree: done.

storage/storage-client/src/lib.rs
thefallentree: I know what you are thinking, you want to implement an get_preceeding_blockmetadata_txn  or even get_blockmetadata_txn_version()  ?  Right now we don't really have that need

thefallentree:
PTAL, I hard coded an 3000 search limit, which should be plenty, since maximum TPS we get is ~1000

msmouse:
storage part looks good to me (I will add proper testing later in my own diff.)

thefallentree:
@metajack ping?

PR #4674, opened 2020-06-23
[move-prover] Partial specs for Roles.move and test for bug

DavidLDill:
There are partial specifications for Roles.move.

In the process of checking those specifications, I discovered a
limitation of the Prover that results in false errors.  The
false error can occur when a module defines a resource and a public
function that takes that resource as an argument.  External code
(another module or script) cannot call the function unless it has an
instance of the resource, but the defining module may prevent
instances of the resource from ever leaving the module.

The prover needs to know whether resources can escape the module
to avoid reporting this error.

There is a new file in tests/sources/regressions/Escape.move that
illustrates this issue, with explanatory comments.

## Motivation

(Write your answer here.)

## Test Plan

## Related PRs

### DIFF
--- a/language/stdlib/modules/Roles.move
+++ b/language/stdlib/modules/Roles.move
@@ -50,8 +50,6 @@ module Roles {
 
     /// The roleId contains the role id for the account. This is only moved
     /// to an account as a top-level resource, and is otherwise immovable.
-    /// INVARIANT: Once an account at address `A` is granted a role `R` it
-    ///            will remain an account with role `R` for all time.
     resource struct RoleId {
         role_id: u64,
     }
@@ -63,7 +61,7 @@ module Roles {
     /// Privileges are extracted in to capabilities. Capabilities hold /
     /// the account address that they were extracted from (i.e. tagged or
     /// "tainted"). Capabilities can then only be restored to the account
-    /// where they were extracted from.
+    /// from which they were extracted.
     resource struct Capability<Privilege: resource> {
         owner_address: address,
     }
@@ -318,7 +316,7 @@ module Roles {
     /// only be granted if you have the correct role. When a capability
     /// leaves the module we tag it with the account where it was held. You
     /// can only put the capability back if the `account` address you are
-    /// storing it back under and the `owner_address` if the incoming capability agree.
+    /// storing it back under and the `owner_address` of the incoming capability agree.
     /// INVARIANT: Once a privilege witness is created and stored under
     /// a Privilege<PrivWitness> resource at an address A there are only two states:
     /// 1. The resource Privilege<PrivWitness> is stored at A;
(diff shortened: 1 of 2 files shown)


language/stdlib/modules/Roles.move
```diff
@@ -354,5 +354,101 @@ module Roles {
...
+    /// Each address has a role-specific privilege corresponding iff that address has
+    /// an RoleId and the role_id of the account matches the privilege.
+    ///
+    spec schema AssociationRootRoleMatchesRoleId {
+        // precondition to grant_root_association
+        // every address has assoc_root role_id iff if it has the privilege
+        // adding one more should preserve it.
+        // TODO BUG (dd): The prover finds many false errors for the following because
+        // it thinks many add_privilege_* functions can put AssociationRootRole on
+        // addresses that don't have the association root RootId.  This is due to
```
wrwg, line 415: I'm wondering why we don't use the usual inductive approach here, which says exactly what you state above (namely that AssociationRootRole can only be at address that have RootId role). Because it's a module invariant, it is assumed on entry and does not need to be proven, but just constraints our global states. On exit we have to prove it still holds (with the usual exception of functions which change it).

emmazzz: Consider some other module calling say `add_privilege_to_account_treasury_compliance_role<Privilege<AssociationRootRole>>(account, assoc_priv)`, where `account` is indeed a treasury compliance account.

This is entirely legal right now. What we need to add here is the restriction that the type parameter `Priv` for function `add_privilege_to_account_treasury_compliance_role` cannot be an indoor thing like `Privilege<AssociationRootRole>`.

DavidLDill: I think that's the approach I'm trying to use.

The problem is that the proof fails because add_privilege_designated_dealer appears to violate the property because add_privilege_designated_dealer can add an arbitrary privilege to an address that has the designated_dealer role.  It thinks that arbitrary privilege can be the AssociationRootRole.  In reality, the arbitrary privilege CANNOT be AssociationRootRole because no instance of that resource is accessible to a module that can call add_privilege_designated_dealer.  I think.

language/stdlib/modules/Roles.move
```diff
@@ -354,5 +354,101 @@ module Roles {
...
+        define SPEC_DESIGNATED_DEALER_ROLE_ID(): u64 { 2 }
+        define SPEC_VALIDATOR_ROLE_ID(): u64 { 3 }
+        define SPEC_VALIDATOR_OPERATOR_ROLE_ID(): u64 { 4 }
+        define SPEC_PARENT_VASP_ROLE_ID(): u64 { 5 }
+        define SPEC_CHILD_VASP_ROLE_ID(): u64 { 6 }
+        define SPEC_UNHOSTED_ROLE_ID(): u64 { 7 }
+
+    }
+
+    /// Property stated informally at definition of RoleId resource
```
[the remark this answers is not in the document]
DavidLDill: Tim put the informal spec at the point of definition.  We can definitely move it.  I was just collecting opinions about what the convention should be.

language/stdlib/modules/Roles.move
wrwg: Well that can be done. We do have the `type<T>()` expression, so you can say such things. I guess it would be a precondition (because there is no matching abort in the program?). Something like:

```
spec add_privilege_to_account_treasury_compliance_role {
   requires type<Priv>() != type<AssociationRootRole>();
}
```

language/stdlib/modules/Roles.move
```diff
@@ -50,7 +50,7 @@ module Roles {
 
     /// The roleId contains the role id for the account. This is only moved
     /// to an account as a top-level resource, and is otherwise immovable.
-    /// INVARIANT: Once an account at address `A` is granted a role `R` it
+    /// SPEC RoleIdPersists: Once an account at address `A` is granted a role `R` it
```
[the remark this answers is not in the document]
DavidLDill: I don't need to use SPEC.  It was just a placeholder until I figured out the right style.

language/stdlib/modules/Roles.move
emmazzz: Right I think adding this pre condition should(?) solve the issue. It seems that the solution I had in mind was to essentially automate the process by identifying which types `Priv` can't be using some analysis and adding these `requires` automatically.

wrwg: Perhaps the core of the problem is that code actually does not abort on calls like this, and henceforth allows real dangerous operations. If Move cannot abort on concrete type instances for generic parameters, we should ask the language team to add this as a native function. It should be easy, there could be a native function `types_equal<T1, T2>`.

@sblackshear @tzakian

DavidLDill: I tried to explain better in the comments in the code.

DavidLDill: Missed some discussion.  I can add the precondition that the arg is not of type AssociationRootRole. 

The prover will catch violations of this in calling modules if we verify them.  I've been trying to avoid assuming that we have verified all calling modules.  When we have an open system, we won't necessarily have access to all calling code.

The Move team should not have to put an assert in there to ensure it, since it's true.  The information exists for us to prove, just looking at the Roles module, that add_privilege_to.... cannot have an AssociationRootRole as an argument.

As a work-around, a precondition might be fair game, though.

PR #4675, opened 2020-06-23
[management] For now, allow arbitrary scripts on pre-mainnet

davidiw:
As per discussion, we plan on allowing arbitrary transaction scripts, since we don't know what we'll need or might not have enough time to build it all.

### DIFF
--- a/config/management/src/genesis.rs
+++ b/config/management/src/genesis.rs
@@ -30,7 +30,7 @@ impl Genesis {
         let genesis = vm_genesis::encode_genesis_transaction_with_validator(
             association_key,
             &validators,
-            None,
+            Some(libra_types::on_chain_config::VMPublishingOption::CustomScripts),
         );
 
         if let Some(path) = self.path {


config/management/src/genesis.rs
```diff
@@ -30,7 +30,7 @@ impl Genesis {
         let genesis = vm_genesis::encode_genesis_transaction_with_validator(
             association_key,
             &validators,
-            None,
+            Some(libra_types::on_chain_config::VMPublishingOption::Open),
```
sblackshear, line 33:
```suggestion
            Some(libra_types::on_chain_config::VMPublishingOption::CustomScripts),
```

`Open` allows both custom modules and custom scripts. The suggested config is more constrained if we don't need/want custom modules on premainnet.

davidiw: Oh, I didn't land this yet... I'm okay with either, but I don't want to explore modules, if we have to do that, I'm inclined to restart :P

so more constrained it is! Thanks for catching this!

zekun000: is it a big problem to allow modules? https://github.com/libra/libra/pull/4686 migrates the smoke test to use this and one of the test is for custom module and scripts.

I don't know how hard it is to make this configurable, but Open is a simple fix :)

davidiw:
If you want to wait until tomorrow, you can land a quick hack and fix it sometime in the following two weeks

PR #4681, opened 2020-06-23
[cluster-test] Dont wait for completion on scale down during cluster teardown

ankushagarwal:
## Summary

* During teardown, set wait_to_completion = false for scale down
* During cluster-test init, try to scale down again, but with wait_to_completion = true
* This allows the ASG to scale down in the background when cluster-test is not running and hence reduce overall cluster test run duration

### DIFF
--- a/testsuite/cluster-test/src/aws.rs
+++ b/testsuite/cluster-test/src/aws.rs
@@ -14,7 +14,7 @@ use rusoto_sts::WebIdentityProvider;
 /// set_asg_size sets the size of the given autoscaling group
 #[allow(clippy::collapsible_if)]
 pub async fn set_asg_size(
-    min_desired_capacity: i64,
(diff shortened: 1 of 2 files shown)


snakech:
May I ask the reason we wanna scale down as well during cluster-test init?

ankushagarwal:
@snakech The previous scale down might not have completed since we set `wait_for_completion=false`. Hence we are calling  it during init with `wait_for_completion=true` to ensure that it completes successfully before calling scale_up.

snakech:
> @snakech The previous scale down might not have completed since we set `wait_for_completion=false`. Hence we are calling it during init with `wait_for_completion=true` to ensure that it completes successfully before calling scale_up.

Make sense, thx for explanation @ankushagarwal

PR #4682, opened 2020-06-23
[testnet] Implement a writeset transaction that resets the auth key for blessed account

runtian-zhou:
…or blessed account.

## Motivation

DO NOT Commit this PR. This is only for demonstration purpose only.

Implement a writeset transaction that resets the auth key for blessed account. In the current setup, if an association address is rotated to an invalid key there's no way for us to recover it other than sending a writeset transaction that overrides the existing resource. This is the very first "hot" fix for the testnet.

Yes

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -624,6 +624,7 @@ dependencies = [
  "libra-types 0.1.0",
  "libra-wallet 0.1.0",
  "libra-workspace-hack 0.1.0",
+ "move-core-types 0.1.0",
  "num-traits 0.2.12 (registry+https://github.com/rust-lang/crates.io-index)",
  "proptest 0.10.0 (registry+https://github.com/rust-lang/crates.io-index)",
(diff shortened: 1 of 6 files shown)


runtian-zhou:
There are two choices here: we can either rewrite the resource by modifying the field directly and serialize the data back, which is the approach I took in this draft, or we can run a fake VM instance to invoke some existing procedures and generates the writeset from the VM. In general the second approach seems to be more desirable as it will be a much cleaner solution. However the current structure doesn't support such functionality. The main reason is that the current LibraClient doesn't implement `StateView`, which is the main api VM uses to fetch on chain state. We might want to consider implementing that in the future to get a more systematic way of generating writeset transactions.

runtian-zhou:
Another interesting point raised by @tzakian is that we can have a `forge_signer` function that's only invokable by `association_root`. This will provide a pseudo "superpower" that allows association to fix most of the on-chain states.

davidiw:

PR #4686, opened 2020-06-23
[management] migrate swarm to use the new builder

zekun000:
## Motivation

1. migrate the libra swarm to use management instead of config-builder
2. support public fullnode in smoke test.

(Write your answer here.)

## Test Plan

smoke test

## Related PRs


config/src/config/test_data/validator.yaml
```diff
@@ -50,3 +51,7 @@ validator_network:
                 from_disk: "/full/path/to/token"
     mutual_authentication: true
     network_id: "validator"
+
+upstream:
+    networks:
+        - validator
```
christinacelee, line 57: this should be empty for a validator (by default `NetworkId::Validator` is upstream)

config/management/src/smoke_test.rs
```diff
@@ -98,79 +110,219 @@ fn smoke_test() {
...
-    };
+#[test]
+fn smoke_test() {
+    LibraNode::prepare();
+    let mut swarm = ManagementBuilder::new(5).create_swarm();
+    swarm.launch_attempt(RoleType::Validator, false).unwrap();
 
-    swarm.launch_attempt(RoleType::FullNode, false).unwrap();
-    assert!(check_connectivity(&mut swarm, 1));
+    // let helper = StorageHelper::new();
```
JoshLind: nit: is there any benefit in leaving this here if it just gets removed in the next commit? Might be cleaner to just delete this now, no?

zekun000: oh no, it's just for me to easily copy the code around lol

config/management/src/smoke_test.rs
```diff
@@ -12,59 +12,73 @@ use libra_secure_storage::{CryptoStorage, KVStorage, Value};
...
+    }
+
+    fn create_association(&self) {
+        self.storage_helper.initialize(ASSOCIATION.into());
+        self.storage_helper
+            .association_key(ASSOCIATION, ASSOCIATION_SHARED)
+            .unwrap();
+    }
+
+    fn initialize_validator_config(&self, index: usize) -> NodeConfig {
```
JoshLind: One of the disadvantages of this refactoring is that it becomes more difficult to identity the distinct steps that need to happen to bootstrap the network and produce genesis (e.g., the steps outlined in the smoke test as comments). Perhaps we could try and keep some of the steps by: (i) adding a comment to the functions "initialize_validator_config" and "finish_validator_config", to explain what each functions does; and/or (ii) renaming the functions?

For example, the method "initialize_validator_config" internally uploads the validator operator key to the shared repo, creates a local NodeConfig for the validator, and then uploads the validator config to the shared repo. Likewise, "finish_validator_config" internally constructs genesis, creates and inserts a waypoint, verifies genesis and then returns a NodeConfig for the validator.

Note: for me, I think the confusing comes from the fact that the term "validator config" is overloaded; here it can mean either: (i) a configuration file, such as NodeConfig; (ii) the name of the command executed by each operator to upload a ValidatorConfig transaction to the shared repo; or (iii) the ValidatorConfig transaction itself.

config/management/src/smoke_test.rs
```diff
@@ -75,21 +89,19 @@ fn smoke_test() {
...
             libra_global_constants::OPERATOR_ACCOUNT.into(),
-            secure_backend(helper.path(), &swarm_path, &ns, "full_node"),
+            self.secure_backend(&ns, "full_node"),
         );
 
-        configs.push(config);
-
-        helper.operator_key(&ns, &ns_shared).unwrap();
-        helper
+        self.storage_helper.operator_key(&ns, &ns_shared).unwrap();
```
JoshLind: I think this might be a bug in the existing code. There shouldn't be any need to call "self.storage_helper.operator_key(..)" twice (see line 81 above)? You should be able to remove this (e.g., tested this locally on master and it seemed to work).

testsuite/tests/libratest/smoke_test.rs
```diff
@@ -454,10 +462,7 @@ fn test_basic_restartability() {
     let peer_to_restart = 0;
     // restart node
     env.validator_swarm.kill_node(peer_to_restart);
-    assert!(env
-        .validator_swarm
-        .add_node(peer_to_restart, RoleType::Validator, false)
-        .is_ok());
+    assert!(env.validator_swarm.add_node(peer_to_restart, false).is_ok());
```
JoshLind, line 465: @davidiw, did you want to move away from the pattern "assert!(x.is_ok())", and instead do unwrap/expect? I know when these asserts fail, it can be difficult to debug

testsuite/libra-swarm/src/swarm.rs
```diff
@@ -307,14 +315,26 @@ impl LibraSwarm {
...
-    pub fn launch_attempt(
-        &mut self,
-        role: RoleType,
-        disable_logging: bool,
-    ) -> Result<(), SwarmLaunchFailure> {
+    pub fn launch(&mut self) {
+        let num_attempts = 5;
+        for _ in 0..num_attempts {
+            match self.launch_attempt(false) {
+                Ok(_) => {
```
JoshLind, line 326: super nit: any reason we can't compact this if each action is only a singe line? Seems like a lot of curly braces :P

config/management/src/smoke_test.rs
zekun000: I think I followed the steps in the original tests for each function i.e. every function corresponds to one step in the `build_swarm`

config/config-builder/src/swarm_config.rs
```diff
@@ -41,7 +41,7 @@ impl SwarmConfig {
         Ok(SwarmConfig {
             config_files,
             faucet_key_path,
-            waypoint: configs[0].base.waypoint.waypoint_from_config().unwrap(),
+            waypoint: configs[0].base.waypoint.waypoint(),
```
davidiw, line 44: yes!

config/management/src/config_builder.rs
```diff
@@ -0,0 +1,270 @@
...
+        genesis_path.create_as_file().unwrap();
+        let genesis = self.storage_helper.genesis(genesis_path.path()).unwrap();
+        self.storage_helper
+            .insert_waypoint(&ns, constants::COMMON_NS)
+            .unwrap();
+        let output = self
+            .storage_helper
+            .verify_genesis(&ns, genesis_path.path())
+            .unwrap();
+        assert_eq!(output.split("match").count(), 5);
```
davidiw, line 132: we have 4 outputs, so when we split we should expect 5 values, e.g., splitting "hello,you" at "," would get "hello" "you"

config/management/src/config_builder.rs
```diff
@@ -0,0 +1,270 @@
...
+    }
+}
+
+#[derive(Debug)]
+pub enum FullnodeType {
+    ValidatorFullnode,
+    PublicFullnode(usize),
+}
+
+pub struct FullnodeBuilder {
```
davidiw, line 173: why not make a new file?

config/management/src/config_builder.rs
```diff
@@ -0,0 +1,270 @@
...
+    ) -> Self {
+        Self {
+            validator_config_path,
+            faucet_key_path,
+            template,
+            build_type,
+        }
+    }
+
+    fn attach_validator_full_node(&self, validator_config: &mut NodeConfig) -> NodeConfig {
```
davidiw, line 195: I see, so we keep my amazing logic :D

config/management/src/storage_helper.rs
```diff
@@ -19,6 +19,7 @@ pub struct StorageHelper {
     temppath: libra_temppath::TempPath,
 }
 
+#[allow(dead_code)]
```
davidiw, line 22: why keep things around, what is dead?

testsuite/tests/libratest/smoke_test.rs
```diff
@@ -146,15 +121,20 @@ impl TestEnvironment {
         .unwrap()
     }
 
-    fn get_validator_ac_client(
+    fn get_client(
```
davidiw, line 124: thank you

testsuite/tests/libratest/smoke_test.rs
davidiw: Yeah, why not use unwrap?

config/management/src/storage_helper.rs
davidiw: I actually want to resolve this one before landing ... maybe just delete what we don't use? Unless you did something really sneaky and I didn't notice it :P

config/management/src/config_builder.rs
zekun000: the magic 5 is the same as num_validators you use and I was confused

testsuite/tests/libratest/smoke_test.rs
zekun000: it's just copy-paste, I don't even look at the code lol

PR #4688, opened 2020-06-23
[cluster-test] Make wait_healthy timeouts more nuanced

andll:
1) Default startup timeout is reduced back to 5 minutes - this is how long normal start takes, and increasing it just wastes more time when startup fails

2) wait_healthy after the experiment now takes deadline from experiment deadline. This should be sufficient for 10% down experiment, and if not, we will increase timeout of this specific experiment, but not everywhere


snakech:
Do we need to extend deadline for 10% down since it is already addressed timeout?

andll:
Good point.

PR #4690, opened 2020-06-23
[Key Manager] Delegate operator signatures to vault, and avoid issuing reconfiguration transactions.

JoshLind:
## Motivation

This PR addresses two issues that currently prevent the key manager from being deployment in the testing environments: (i) the key manager should not export the operator key directly, but rather, have vault sign transactions using the key; and (ii) the key manager should not be responsible for issuing reconfiguration transactions, as it does not have access to the association keys.

These are each addressed in separate commits:

1. First, we delegate signing of the rotation transaction from the key manager to vault. This avoids having to export the operator key to the key manager.

2. Second, we move the reconfiguration transaction creation and submission to the key manager tests.

Note: I am currently planning a refactor commit that cleans up the key manager code. This is more a hot-fix style PR to avoid @mgorven being blocked.

Yes.

## Test Plan

All tests still pass locally.

## Related PRs

None.

### DIFF
--- a/secure/key-manager/src/lib.rs
+++ b/secure/key-manager/src/lib.rs
@@ -22,6 +22,7 @@
 use crate::{counters::COUNTERS, libra_interface::LibraInterface};
 use libra_crypto::{
     ed25519::{Ed25519PrivateKey, Ed25519PublicKey},
+    hash::CryptoHash,
     x25519, PrivateKey,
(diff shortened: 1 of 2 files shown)


JoshLind:
Thanks, Michael! @bors-libra r=mgorven

PR #4699, opened 2020-06-24
[config] simplify UpstreamConfig

christinacelee:
## Motivation

Simplify existing `UpstreamConfig` s.t.:
- upstream config now only defines the upstream network of a node
- validators do not need to explicitly set its validator network as upstream
This PR also updates mempool and state sync uses the actual NetworkId to identify networks, instead of its local peer ID within that network. 

## Test Plan

Existing/refactored tests

### DIFF
--- a/config/config-builder/src/full_node_config.rs
+++ b/config/config-builder/src/full_node_config.rs
@@ -4,9 +4,7 @@
 use crate::{BuildSwarm, Error, ValidatorConfig};
 use anyhow::{ensure, Result};
 use libra_config::{
-    config::{
-        DiscoveryMethod, NetworkPeersConfig, NodeConfig, PeerNetworkId, RoleType, UpstreamConfig,
-    },
+    config::{DiscoveryMethod, NetworkPeersConfig, NodeConfig, RoleType},
     generator,
     network_id::NetworkId,
     utils,
@@ -15,7 +13,7 @@ use libra_crypto::ed25519::Ed25519PrivateKey;
 use libra_network_address::NetworkAddress;
 use libra_types::transaction::Transaction;
 use rand::{rngs::StdRng, SeedableRng};
-use std::{collections::HashSet, str::FromStr};
+use std::str::FromStr;
 
 pub struct FullNodeConfig {
     pub advertised_address: NetworkAddress,
@@ -195,8 +193,7 @@ impl FullNodeConfig {
             .ok_or(Error::MissingFullNodeNetwork)?;
         let seed_peers =
             generator::build_seed_peers(&validator_full_node_network, self.bootstrap.clone());
-        let upstream_peer_id = validator_full_node_network.peer_id();
-        for config in configs.iter_mut() {
+        for (idx, config) in configs.iter_mut().enumerate() {
             let network = config
                 .full_node_networks
                 .last_mut()
@@ -204,12 +201,9 @@ impl FullNodeConfig {
             network.network_id = NetworkId::Public;
             network.network_peers = network_peers.clone();
             network.seed_peers = seed_peers.clone();
-            let mut upstream = UpstreamConfig::default();
-            let peer_id = network.peer_id();
-            upstream.upstream_peers = vec![PeerNetworkId(peer_id, upstream_peer_id)]
-                .into_iter()
-                .collect::<HashSet<_>>();
-            config.upstream = upstream;
+            if idx < actual_nodes - 1 {
+                config.upstream.networks.push(network.network_id.clone());
+            }
         }
 
         Ok((configs, faucet_key))
@@ -262,14 +256,10 @@ mod test {
         FullNodeConfig::new()
             .extend_validator(&mut validator_config)
             .unwrap();
-        let val_fn = &validator_config.full_node_networks[0];
 
         let fnc = FullNodeConfig::new().build().unwrap();
-        let fn_network_id = fnc.full_node_networks[0].peer_id();
-        assert!(fnc
-            .upstream
-            .upstream_peers
-            .contains(&PeerNetworkId(fn_network_id, val_fn.peer_id())));
+        let fn_network_id = &fnc.full_node_networks[0].network_id;
+        assert!(fnc.upstream.networks.contains(fn_network_id));
     }
 
     #[test]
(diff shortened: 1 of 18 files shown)


config/config-builder/src/full_node_config.rs
```diff
@@ -195,21 +193,19 @@ impl FullNodeConfig {
...
             network.network_id = NetworkId::Public;
             network.network_peers = network_peers.clone();
             network.seed_peers = seed_peers.clone();
-            let mut upstream = UpstreamConfig::default();
-            let peer_id = network.peer_id();
-            upstream.upstream_peers = vec![PeerNetworkId(peer_id, upstream_peer_id)]
-                .into_iter()
-                .collect::<HashSet<_>>();
-            config.upstream = upstream;
+            if idx < actual_nodes - 1 {
```
davidiw, line 204: oh... I hate config-builder

config/src/config/upstream_config.rs
```diff
@@ -1,52 +1,60 @@
...
 // SPDX-License-Identifier: Apache-2.0
 
+use crate::network_id::NetworkId;
 use libra_types::PeerId;
 use serde::{Deserialize, Serialize};
-use std::collections::HashSet;
 
 /// In general, a network ID is a PeerId that this node uses to uniquely identify a network it belongs to.
 /// This is equivalent to the `peer_id` field in the NetworkConfig of this NodeConfig
 pub type UpstreamNetworkId = PeerId;
```
davidiw, line 10: should we eliminate this?

config/src/config/upstream_config.rs
```diff
@@ -1,52 +1,60 @@
...
+    // specified here, in this order
+    pub networks: Vec<NetworkId>,
 }
 
 impl UpstreamConfig {
     /// Determines whether a node `peer_id` in network `network_id` is an upstream peer of a node with this NodeConfig.
-    pub fn is_upstream_peer(&self, peer: PeerNetworkId) -> bool {
-        self.is_primary_upstream_peer(peer) || self.fallback_networks.contains(&peer.network_id())
+    pub fn is_upstream_peer(&self, peer: &PeerNetworkId) -> bool {
+        let network_id = peer.network_id();
```
davidiw, line 30: this feels a little weird, should we pass in just network id? is_upstream_network?

config/src/config/upstream_config.rs
```diff
@@ -1,52 +1,60 @@
...
+        network_id == NetworkId::Validator || self.networks.contains(&network_id)
     }
 
-    pub fn is_primary_upstream_peer(&self, peer: PeerNetworkId) -> bool {
-        self.primary_networks.contains(&peer.network_id()) || self.upstream_peers.contains(&peer)
+    pub fn is_primary_upstream_peer(&self, peer: &PeerNetworkId) -> bool {
+        let network_id = peer.network_id();
+        network_id == NetworkId::Validator || self.networks.get(0) == Some(&network_id)
```
davidiw, line 36: maybe we could combine these APIs and return Option<u32> None implies not upstream, 0 is primary, 1 is secondary etc

config/src/config/upstream_config.rs
```diff
@@ -1,52 +1,60 @@
...
     }
 }
 
-#[derive(Copy, Clone, Debug, Deserialize, Eq, Hash, PartialEq, Serialize)]
+#[derive(Clone, Debug, Deserialize, Eq, Hash, PartialEq, Serialize)]
 /// Identifier of a node, represented as (network_id, peer_id)
-pub struct PeerNetworkId(pub UpstreamNetworkId, pub PeerId);
+pub struct PeerNetworkId(pub NetworkId, pub PeerId);
```
davidiw, line 42: feels like we don't need this any more?

state-synchronizer/src/peer_manager.rs
```diff
@@ -166,7 +166,7 @@ impl PeerManager {
     fn get_active_upstream_peers(&self) -> Vec<(&PeerNetworkId, &PeerInfo)> {
```
davidiw, line 166: I see, maybe we should file an issue to fix the PeerNetworkId thing...

config/src/config/upstream_config.rs
zekun000: I was banging my head on this, and learned UpstreamNetworkId is not a NetworkId but is a PeerId and as a single node it can have several different peer ids in different networks. amazing

christinacelee: so `PeerNetworkId` is used by mempool/state sync to identify a peer across all the networks it is involved in, since last time I checked there can be duplicate peer IDs across different networks..

davidiw: I see, I guess I had envisioned the model slightly differently, but I am giving feedback without digging into the code very well. I do reckon that some of this logic should probably be outside of the config space, but at the same time,  I think we should divide the needs and the wants.

config/src/config/upstream_config.rs
```diff
@@ -1,52 +1,59 @@
...
     }
 
     pub fn random() -> Self {
-        Self(UpstreamNetworkId::random(), PeerId::random())
+        Self(NetworkId::default(), PeerId::random())
+    }
+
+    pub fn random_validator() -> Self {
+        Self(NetworkId::Validator, PeerId::random())
     }
```
davidiw, line 58: if these are test code, we should put them behind said flags

```
```

mempool/src/shared_mempool/peer_manager.rs
```diff
@@ -56,13 +56,13 @@ impl PeerManager {
     }
 
     // returns true if `peer` is discovered for first time, else false
-    pub fn add_peer(&self, peer: PeerNetworkId) -> bool {
+    pub fn add_peer(&self, peer: Arc<PeerNetworkId>) -> bool {
```
davidiw, line 59: probably don't need Arc here and would argue for simplicity implementing Copy on NetworkId and NetworkContext is probably better at this point in time.

christinacelee: yeah, it's unfortunate that NetworkId has a String so implementing Copy is kind of a pain :/

PR #4700, opened 2020-06-24
[compat] add CLI tool for managing backwards compatibility testing workflows

rustielin:
## Motivation

Adds a CLI tool at `devtools/compat/run_compat.sh` for administration and usage of long-running Libra Network instances for mixed-version validator/backwards compatibility testing.

The main functionality allows us to reserve a running ECS cluster and perform upgrade a subset of the validators. Currently, we can either test a certain revision against the latest testnet revision, or manually specify two revisions. Each upgrade occurs on half of the network at once, so as to make backwards compatibility necessary for QC.

Some TODO:
* command to restore back to testnet version after test run
* testing with real testnet data on test run (ROAR on testnet via snapshot copy from prod) https://github.com/novifinancial/libra-ops/issues/601
* switch to persistent storage instead of relying on state-sync from fullnodes
* Github Actions workflow, non-blocking for now
* healthchecks for various layers (e.g. network, consensus, VM)

Yes

## Test Plan

Spin up `compat-0` cluster using Terraform. Our pool of available clusters are named `compat-[0-9]+`.

Let's grab `master_65e3b04` and check its compatibility with the latest testnet version. (Certain output is abbreviated for sanity)

```
$ ./run_compat.sh test --tag master_65e3b04 testnet
...
------------ Acquired compat-0 ------------
...
Network is currently running the following versions:
        compat-0-validator-0=master_d4ad450
        compat-0-validator-1=master_d4ad450
        compat-0-validator-2=master_01cf0ac
        compat-0-validator-3=master_01cf0ac
        compat-0-fullnode-0=master_d4ad450
        compat-0-fullnode-1=master_d4ad450


------------ Starting TESTNET compat test ------------

Upgrade path:
        vault (force)   ==> latest
        validators 0, 1 ==> master_65e3b04
        validators 2, 3 ==> master_1fd1e5a
        fullnodes 0, 1  ==> master_1fd1e5a

Diagnosing node health...
Cluster is unhealthy, patching entire network to TESTNET_TAG = master_1fd1e5a and exiting...
...
Please retry when the network is healthy
```

Output of the above command will indicate which cluster it was able to lock, or to retry if it times out. From the above, we were able to lock the cluster named `compat-0` by acquiring its corresponding lock (in s3). Looks like `compat-0` was not originally running testnet version, perhaps from a previous test, so the first run upgrades all fullnodes and validators to testnet version, which is `master_1fd1e5a` at the moment. This allows the network to reach a steady state first before further upgrades.

Then, subsequent runs in the same test can explicitly specify `compat-0` (bypass cluster selection since we still hold the lock) to upgrade the other half to `master_65e3b04`. 

```
$ ./run_compat.sh test --workspace compat-0 --tag master_65e3b04 testnet
...
Network is currently running the following versions:
        compat-0-validator-0=master_1fd1e5a
        compat-0-validator-1=master_1fd1e5a
        compat-0-validator-2=master_1fd1e5a
        compat-0-validator-3=master_1fd1e5a
        compat-0-fullnode-0=master_1fd1e5a
        compat-0-fullnode-1=master_1fd1e5a

------------ Starting TESTNET compat test ------------

Upgrade path:
        vault (force)   ==> latest
        validators 0, 1 ==> master_65e3b04
        validators 2, 3 ==> master_1fd1e5a
        fullnodes 0, 1  ==> master_1fd1e5a

Diagnosing node health...
Cluster is healthy!
...
Done!
```

We can also manually specify which revisions we want to check. `master_d4ad450` and `master_01cf0ac` are adjacent so there's a good chance they're compatible... let's check:

```
$  ./run_compat.sh test --tag master_01cf0ac  manual --alt-tag master_d4ad450
...
Network is currently running the following versions:
        compat-0-validator-0=master_65e3b04
        compat-0-validator-1=master_65e3b04
        compat-0-validator-2=master_1fd1e5a
        compat-0-validator-3=master_1fd1e5a
        compat-0-fullnode-0=master_1fd1e5a
        compat-0-fullnode-1=master_1fd1e5a


------------ Starting MANUAL compat test ------------

Upgrade path:
        vault (force)   ==> latest
        validators 0, 1 ==> master_d4ad450
        validators 2, 3 ==> master_01cf0ac
        fullnodes 0, 1  ==> master_d4ad450
...
```

Now, run an arbitrary workload against it!

```
$ cargo run -p cli -- --url https://client.compat-0.aws.hlw3truzy4ls.com/ --waypoint <waypoint>
...
> account create
> account mint 0 100 LBR
> query balance 0
```

If another process tries to get `compat-0` while we still hold the lock, they will spin until we finish testing and release the lock, or time out. Locks have default retention time of 3 hrs for liveness:

```
$ ./run_compat.sh test --tag master_01cf0ac  manual --alt-tag master_d4ad450
...
Lock compat-lock-0 expires in 2:53:57.154780 or upon job complete
Attempt 0: All clusters have jobs running on them. Retrying in 10 secs.
Lock compat-lock-0 expires in 2:53:46.418962 or upon job complete
Attempt 1: All clusters have jobs running on them. Retrying in 10 secs.
```

To clean up, unlock the cluster using `admin` subcommand: `./run_compat.sh admin unlock compat-lock-0`. We can also use `./run_compat.sh admin reset compat-0` to reset the cluster to its initial steady state.

## Related PRs

https://github.com/novifinancial/libra-ops/pull/587 for discussion on retention locks in s3


devtools/compat/pyhelpers/git.py
```diff
@@ -0,0 +1,33 @@
+#!/usr/bin/env python3
+# Copyright (c) The Libra Core Contributors
+# SPDX-License-Identifier: Apache-2.0
+from .cli import execute_cmd_with_text_output
+
+# length of git hash prefix
+SHORT_LEN = 7
+TAG_PREFIX = "master_"
```
sausagee, line 8: Heads up - this image may not exist in ECR.

ATM we don't build the full set of images for every commit. 
- LBT builds validator and safety-rules at the tip of the auto branch when a PR tries to land, and they use different tags (https://github.com/libra/libra/blob/ba042341101cfec720f6ec46992bf0bb3c99bbb1/.github/workflows/land-blocking.yml#L38) to distinguish images in trunk.   
- Continuous push builds every 3 hrs from the tip of then master. Only that commit has the full set of images and with the master_xxx tag.

devtools/compat/.gitignore
```diff
@@ -0,0 +1,4 @@
+__pycache__
```
sausagee, line 1: I just looked at `devtools/`. It was originally used for the internal `cargo x` tooling in Rust.  The `compat` workflow is currently in py and would cause confusion to some.  Would you consider moving compat to `testsuite` dir? I apologize for failing to spot this earlier.

devtools/compat/compat.vars
```diff
@@ -0,0 +1,5 @@
+NUM_VALIDATORS=4
```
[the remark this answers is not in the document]
rustielin: Agreed. We could increase the number of validators/fullnodes if there's a need, but I'll just leave that as a comment for future discussion

devtools/compat/.gitignore
rustielin: will do

devtools/compat/compat.py
```diff
@@ -0,0 +1,291 @@
...
+    all_tags = [
+        curr_tag_v0,
+        curr_tag_v1,
+        curr_tag_v2,
+        curr_tag_v3,
+        curr_tag_f0,
+        curr_tag_f1,
+    ]
+    if not all([tag == ALT_TEST_TAG for tag in all_tags]):
+        unhealthy = True
```
sausagee, line 236: I think the node health actually refers to if the ECS task is using the intended image tag.  I think we should add other checks, such as containers in the node are actually up. 

It looks like what we have here is setting up the skeleton. I supposed storage, consensus, client checks would be added later to determine backward compatibility breakage?

rustielin: Yea, I should change up the terminology. We already have some stuff for checking container up for devnet right? 

And yea the idea was that initial run of this tool will lock/bring up the necessary network for test. Then other tests (don't exist yet) can do their query/submit txn/ etc to the network to determine breakages. I imagine pylibra would be a good way to test this. Plus reading metrics from prometheus

devtools/compat/pyhelpers/git.py
rustielin: Ah right, thanks for catching that. I'll only need safety-rules and validator images for now, so I'll borrow the LBT `land_*` ones.

rustielin: Also doesn't look like we hold onto those images for too long. Shouldn't be too big of an issue for now but might limit option to test for example: HEAD vs HEAD~15. I've added a basic check for if the image exists before launching the test.

testsuite/compat/compat.py
```diff
@@ -0,0 +1,301 @@
...
+    WORKSPACE = select_compat_cluster(ECS_POOL_SIZE, TEST_TAG)
+    # fails if no workspace
+
+def_map = get_task_defs(WORKSPACE)
+
+# get the task definition families as they are the def_map keys
+v0_fam = f"{WORKSPACE}-validator-0"
+v1_fam = f"{WORKSPACE}-validator-1"
+v2_fam = f"{WORKSPACE}-validator-2"
+v3_fam = f"{WORKSPACE}-validator-3"
```
mgorven, line 147: Ideally this should support arbitrary sized deployments (even if the size needs to be provided to this script).

rustielin: Agreed @ Python AWS API. I'll hold off on it for now though, until this MVP proves useful enough. Then we can decide if we want to do that or make the better decision of just porting it to Rust :)

rustielin: > Ideally this should support arbitrary sized deployments (even if the size needs to be provided to this script).

So currently the way the upgrades happen for the network are very specific to the network topology of 4 validators and 2 fullnodes. The cluster needs to be first created by Terraform before compat can grab it.

I imagine this tool can also be used to test certain types of upgrades on network with various configurations. They won't be truly "arbitrary" though, as the network would have to be predefined and created via Terraform -- then compat can take care of the rest, applying revision A to N/3 + 1 validators, and B to the rest, etc.

mgorven: I'm fine if the script actually only works with N=4, I think the code is just nicer if it does `for i in range(num_validators)` instead of copy/pasta lines.

rustielin: Gotcha. I originally thought it would be easier to understand if the lines were more explicit, but I've updated it

testsuite/compat/compat_helpers.py
```diff
@@ -0,0 +1,183 @@
...
+            ],
+            err=f"could not get task definition for {fam}",
+        )
+        key = task_def.get("taskDefinition").get("family")
+        ret[key] = task_def.get("taskDefinition")
+        # put the tags separately
+        tags = task_def.get("tags")
+        ret[key]["tags"] = tags if tags else []
+
+    # ecs_tasks = execute_cmd_with_json_output(
```
sausagee, line 56: Clean up if we don't need it anymore.

devtools/compat/compat.py
sausagee: alrighty, land this as is and add more tests later.

testsuite/compat/compat_helpers.py
rustielin: whoops well that's embarrassing

rustielin:
Thanks for the feedback @sausagee and @mgorven! I'll land this as is and make revisions and tests where necessary.

I'm in the process of preparing a report for how to run tests against compat + some other proposals. I'll send to you guys and devinfra folks for review on slack

PR #4702, opened 2020-06-24
[transaction scripts] Add missing transaction scripts for validator set management

valerini:
This PR adds two missing transactions scripts.

With this PR there are now 7 scripts, they cover all of the public methods of modules ValidatorConfig and LibraSystem
(except for ValidatorConfig::remove_operator method which is likely to be removed from the module):

1) `create_validator_account` (called with AssociationRootRole capability)
2) `create_validator_operator_account` (called with AssociationRootRole capability)
3) `add_validator.move` (called with AssociationRootRole capability)
4) `remove_validator.move` (called with AssociationRootRole capability)
5) `reconfigure.move` (called with AssociationRootRole capability): aggregate update of all validator configs in the set
6) `set_validator_operator.move` (called by validator)
7) `set_validator_config.move` (called by validator or, if set, by validator's operator)


language/transaction-builder/src/generated.rs
```diff
@@ -325,6 +325,37 @@ pub fn encode_create_validator_account_script(
...
 
+/// Create a validator operator account at `new_validator_address` with `auth_key_prefix`.
+pub fn encode_create_validator_operator_account_script(
+    token: TypeTag,
+    new_account_address: AccountAddress,
+    auth_key_prefix: Vec<u8>,
+) -> Script {
+    Script::new(
+        vec![
+            161, 28, 235, 11, 1, 0, 7, 1, 0, 4, 2, 4, 9, 3, 13, 17, 4, 30, 4, 5, 34, 45, 7, 79,
```
JoshLind, line 336: n00b question: how are these generated?

PR #4706, opened 2020-06-24
testing strong commit

danielxiangzl:
implement strong vote with marker

add marker info to the QC

implemented the computation of the number of endorsers

## Motivation

(Write your answer here.)

## Test Plan

## Related PRs

### DIFF
--- a/consensus/consensus-types/src/block_data.rs
+++ b/consensus/consensus-types/src/block_data.rs
@@ -148,6 +148,7 @@ where
             LedgerInfoWithSignatures::new(
                 LedgerInfo::new(ancestor, HashValue::zero()),
                 BTreeMap::new(),
+                BTreeMap::new(),
             ),
         );
 
--- a/consensus/consensus-types/src/timeout_certificate.rs
+++ b/consensus/consensus-types/src/timeout_certificate.rs
@@ -17,6 +17,7 @@ use std::{collections::BTreeMap, fmt};
 pub struct TimeoutCertificate {
     timeout: Timeout,
     signatures: BTreeMap<Author, Ed25519Signature>,
+    markers: BTreeMap<Author, Round>,
 }
 
 impl fmt::Display for TimeoutCertificate {
@@ -36,6 +37,7 @@ impl TimeoutCertificate {
         Self {
             timeout,
             signatures: BTreeMap::new(),
+            markers: BTreeMap::new(),
         }
     }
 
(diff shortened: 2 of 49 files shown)


consensus/consensus-types/src/block_data.rs
```diff
@@ -148,6 +148,7 @@ where
             LedgerInfoWithSignatures::new(
                 LedgerInfo::new(ancestor, HashValue::zero()),
                 BTreeMap::new(),
+                BTreeMap::new(),
```
dahliamalkhi, line 151: I am not sure we need to aggregate markers in the QC. 

It seems we may want to aggregate endorsements on a target block, coming from votes on potentially different blocks, whose markers endorse the target.

consensus/consensus-types/src/timeout_certificate.rs
```diff
@@ -17,6 +17,7 @@ use std::{collections::BTreeMap, fmt};
 pub struct TimeoutCertificate {
     timeout: Timeout,
     signatures: BTreeMap<Author, Ed25519Signature>,
+    markers: BTreeMap<Author, Round>,
```
dahliamalkhi, line 20: really puzzled: why do we need markers on timeout certificates?

consensus/consensus-types/src/vote.rs
```diff
@@ -27,17 +28,20 @@ pub struct Vote {
     signature: Ed25519Signature,
     /// The round signatures can be aggregated into a timeout certificate if present.
     timeout_signature: Option<Ed25519Signature>,
+    /// The marker equals the largest round number of any conflicting block that the validator voted for
+    marker: Round,
```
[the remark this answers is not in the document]
danielxiangzl: Agreed.

consensus/consensus-types/src/block_data.rs
danielxiangzl: The current LibraBFT implementation does not use threshold sig. Instead, QC stores 2f+1 signatures in the BTreeMap struct. I am doing similar things for the marker, but the current implementation is not secure (markers are not signed).

consensus/consensus-types/src/timeout_certificate.rs
danielxiangzl: Yes, you are right..we don't need markers on TC. Earlier I thought a TC can serve as a QC as well, but it is not the case in the current implementation.

davidiw:
Curious if this is intended to be a feature added to the Libra code base or an exploratory project? If the latter, I would recommend doing this as a PR in your own repo. We've done the similar with our exploration of SGX: https://github.com/zaxguo/libra/pull/1

danielxzl:
> Curious if this is intended to be a feature added to the Libra code base or an exploratory project? If the latter, I would recommend doing this as a PR in your own repo. We've done the similar with our exploration of SGX: [zaxguo#1](https://github.com/zaxguo/libra/pull/1)

Hi, this is currently a research project and hopefully can be added to the codebase in the future :) The PR is mainly for the purpose of cluster test and measurements, I would be happy to close this and reopen it only when doing the measurement.

PR #4712, opened 2020-06-24
[libra framework] Add expiration to VASPs.

tzakian:
Adds expiration to VASPs, and if either VASP is expired in an above-threshold deposit aborts the transaction.

#### Testing
Adds a new test to make sure that an expired VASP aborts the transaction with the expected error code.

Added a `TransactionBuilder` to the e2e tests to make it easier to build transactions. Eventual goal is to replace all the helper functions in `Account` that create signed transactions with this.


language/stdlib/modules/LibraTransactionTimeout.move
```diff
@@ -25,7 +25,7 @@ module LibraTransactionTimeout {
 
   public fun is_valid_transaction_timestamp(timestamp: u64): bool acquires TTL {
     // Reject timestamp greater than u64::MAX / 1_000_000;
-    if(timestamp > 9223372036854) {
+    if(timestamp > 18446744073709) {
```
tzakian, line 28: @runtian-zhou I wound up changing this since it was set to `std::i64::MAX/1_000_000` instead of `std::u64::MAX`, let me know if this change looks OK.

language/stdlib/modules/LibraAccount.move
```diff
@@ -218,6 +218,8 @@ module LibraAccount {
             // travel rule does not apply for intra-VASP transactions
             VASP::parent_address(sender) != VASP::parent_address(payee)
         ) {
+            // If either of the VASPs are expired, then abort
+            assert(!VASP::is_expired(sender) && !VASP::is_expired(payee), 9000); // TODO: proper error code
```
sblackshear, line 222: Should we also prohibit transfers under the dual attestation threshold (i.e., move this before the `if`)?

language/stdlib/modules/VASP.move
```diff
@@ -164,6 +164,15 @@ module VASP {
         *&borrow_global<ParentVASP>(parent_address(addr)).num_children
     }
 
+    /// Return whether the VASP at `addr` is expired based upon the parent
+    /// VASP's `expiration_date` field.
+    public fun is_expired(account_address: address): bool
```
sblackshear, line 169: Should we make this return `false` for any non-VASP account, or does an abort make sense?

language/e2e-tests/src/tests/transaction_builder.rs
```diff
@@ -623,6 +624,49 @@ fn dual_attestation_payment() {
...
     }
+    {
+        // Transaction >= 1_000_000 threshold goes through signature verification with valid signature, passes
+        // Do the offline protocol: generate a payment id, sign with the receiver's private key, include
+        // in transaction from sender's account
+        let ref_id = lcs::to_bytes(&9999u64).unwrap();
+        // choose an amount above the dual attestation threshold
+        let payment_amount = 1_000_000u64;
+        // UTF8-encoded string "@@$$LIBRA_ATTEST$$@@" without length prefix
+        let mut domain_separator = vec![
```
bob-wilson, line 635: Is there some reason that this can't be written as a byte string literal? i.e., b"@@$$LIBRA_ATTEST$$@@"

tzakian: Don't think so, was just copying and pasting a different test. I'll update :)

language/stdlib/modules/LibraAccount.move
tzakian: Mmm I was going back and forth on it. My thinking was that by having it under the if, the "fail state" of a VASP is a UHW. I'll go for the more conservative approach at this point (move it above), and then if we want to relax it later it should be easier.

language/stdlib/modules/VASP.move
tzakian: Good question! My thinking was abort since other functions in the module will abort if they're called on a non-VASP account. I think it will make it easier to have it return false if we move the check for VASPs above the if. I'll update :)

tzakian:
After offline discussions we decided that VASP expiration was not needed for V1 and that this logic was best left till later if we need it. Closing since it's no longer needed.

PR #4714, opened 2020-06-24
[executor] fuzzing for execute_and_commit_chunk

mimoo:
This PR adds a fuzzer for ChunkExecutor.
This code is eventually called by state-sync to execute a chunk of transaction and verify a ledger info proof.

### DIFF
--- a/testsuite/libra-fuzzer/src/fuzz_targets/execute_and_commit_chunk.rs
+++ b/testsuite/libra-fuzzer/src/fuzz_targets/execute_and_commit_chunk.rs
@@ -1,0 +1,31 @@
+// Copyright (c) The Libra Core Contributors
+// SPDX-License-Identifier: Apache-2.0
+
+use crate::FuzzTargetImpl;
+use executor::fuzzing::fuzz;
+use libra_proptest_helpers::ValueGenerator;
+use rand::RngCore;
+
+#[derive(Clone, Debug, Default)]
+pub struct ExecuteAndCommitChunk;
+
+impl FuzzTargetImpl for ExecuteAndCommitChunk {
+    fn name(&self) -> &'static str {
+        module_name!()
+    }
+
+    fn description(&self) -> &'static str {
+        "state-sync > executor::execute_and_commit_chunk"
+    }
+
+    fn generate(&self, _idx: usize, _gen: &mut ValueGenerator) -> Option<Vec<u8>> {
+        let mut output = vec![0u8; 4096];
+        let mut rng = rand::thread_rng();
+        rng.fill_bytes(&mut output);
+        Some(output)
+    }
+
+    fn fuzz(&self, data: &[u8]) {
+        fuzz(data);
+    }
+}
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2520,6 +2520,8 @@ dependencies = [
  "consensus 0.1.0",
  "consensus-types 0.1.0",
  "datatest-stable 0.1.0",
+ "executor 0.1.0",
+ "executor-types 0.1.0",
  "hex 0.4.2 (registry+https://github.com/rust-lang/crates.io-index)",
  "language-e2e-tests 0.1.0",
  "libra-canonical-serialization 0.1.0",
@@ -2539,6 +2541,7 @@ dependencies = [
  "sha-1 0.9.1 (registry+https://github.com/rust-lang/crates.io-index)",
  "state-synchronizer 0.1.0",
  "stats_alloc 0.1.8 (registry+https://github.com/rust-lang/crates.io-index)",
+ "storage-interface 0.1.0",
  "structopt 0.3.15 (registry+https://github.com/rust-lang/crates.io-index)",
  "vm 0.1.0",
 ]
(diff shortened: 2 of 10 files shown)


mimoo:
please review @lightmark @wqfish :)

testsuite/libra-fuzzer/src/fuzz_targets/execute_and_commit_chunk.rs
```diff
@@ -0,0 +1,70 @@
...
+
+        // verified_target_li
+        let strategy = LedgerInfoWithSignatures::arbitrary();
+        let strategy_tree = match strategy.new_tree(&mut runner) {
+            Ok(x) => x,
+            Err(_) => return,
+        };
+        let verified_target_li = strategy_tree.current();
+
+        //
```
lightmark, line 67: ?

testsuite/libra-fuzzer/src/fuzz_targets/execute_and_commit_chunk.rs
```diff
@@ -0,0 +1,70 @@
...
+        // txn_list_with_proof
+        let strategy = TransactionListWithProof::arbitrary();
+        let strategy_tree = match strategy.new_tree(&mut runner) {
+            Ok(x) => x,
+            Err(_) => return,
+        };
+        let txn_list_with_proof = strategy_tree.current();
+
+        // verified_target_li
+        let strategy = LedgerInfoWithSignatures::arbitrary();
```
lightmark, line 60: how do you make sure the LI is consistent with txn_list?

mimoo: right now it isn't, because it doesn't have to be IIRC (state sync doesn't verify it).

but the fuzzer is probably not going to get passed `verify_chunk()` because of that.

We can fix this to always have a consistent LI, or we can do it in another fuzzer to make sure we cover inputs that don't have consistent LI. What do you think?

Also how do we make it so that the LI is consistent...

lightmark: I just realized it is a draft. I think we verified a lot of stuff before execution the chunk.
`txn_list` should be valid with proof and it should be consistent with the `root_hash` carried by the LI passed in. Basically you have to generate all the txns backed by a LI and break them into smaller chunks with less txns in each chunk.

sausagee:
This has been sitting around for a month.  Can we give it a second look?

testsuite/libra-fuzzer/src/fuzz_targets/execute_and_commit_chunk.rs
lightmark: Can you remove this line?

testsuite/libra-fuzzer/src/fuzz_targets/execute_and_commit_chunk.rs
lightmark: @sausagee How to test this fuzzing function? According to what I said, I don't think it can pass through the verification process.

mimoo: check the README in libra-fuzzer, tl;dr is:

```
cargo run -p libra-fuzzer --bin libra-fuzzer generate execute_and_commit_chunk
cargo run -p libra-fuzzer --bin libra-fuzzer fuzz execute_and_commit_chunk
```

mimoo: state-sync does the following checks ([according to this](https://github.com/libra/libra/blob/master/state-synchronizer/src/coordinator.rs#L925)):

* `target_epoch_and_version = (target.ledger_info().epoch(), target.ledger_info().version())`
* 
   ```
   local_epoch_and_version = (
      self.local_state.highest_local_li.ledger_info().epoch(),
      self.local_state.highest_local_li.ledger_info().version(),
   );
   ```
* ensure `target_epoch_and_version >= local_epoch_and_version`

mimoo: for the first test, the code should just return an error anyway instead of crashing so we should be able to fuzz it without change of the proptest strategy: https://github.com/libra/libra/blob/master/execution/executor/src/lib.rs#L184

mimoo: made a PR for this: https://github.com/libra/libra/pull/5454

mimoo: for the second test we can maybe bootstrap a local epoch = 0 & version = 0

mimoo:
note that this will find a crash if this doesn't land before tomorrow :P https://github.com/libra/libra/pull/5454

lightmark:
> note that this will find a crash if this doesn't land before tomorrow :P #5454

I don't think #5454 makes sense. If this PR crashes, does that mean the fuzzing function doesn't create valid chunk object?

PR #4716, opened 2020-06-24
[language][vm] data cache clean up

vgao1996:
## Summary
WIP draft for the planned data cache clean up.

## Test Plan
cargo test

## Current Progress
- [x] Move VM
- [x] Hooking up Libra VM
- [x] Fix Tests & Other Build Errors
- [x] Benchmarking
  - [x] Implement Additional Caching if Necessary
- [x] Review: should `RmoteCache` return bytes or deserialized values?
- [x] Session API (or other ways to hide the transaction data cache)
- [x] ~~Fix bug: excessive deletion ops~~ Will put up a separate PR
- [x] Lint
- [ ] Cleanup

Note: Although this PR doesn't get `FatType` & `FatStructType` removed, we're pretty close. `FatType` is still used by a few clients that can be fixed individually.


language/libra-vm/src/data_cache.rs
```diff
@@ -75,8 +82,38 @@ impl<'a> StateViewCache<'a> {
 }
 
 impl<'block> RemoteCache for StateViewCache<'block> {
-    fn get(&self, access_path: &AccessPath) -> VMResult<Option<Vec<u8>>> {
-        StateViewCache::get(self, access_path)
+    fn get_module(&self, module_id: &ModuleId) -> VMResult<Option<CompiledModule>> {
```
dariorussi, line 85: I am not sure I like this more than what it was but I am fine changing it and going out either way.
The thing I like about returning the `Vector<u8>` is that the VM materializes things. The VM is in charge of everything and the only thing in are blobs.
I understand the tension with serialization and access path, and I am not sure we are in a position to make a final call there. In that respect anything works. 
But I invite you to quickly evaluate if this change adds much

language/move-vm/types/src/data_store.rs
```diff
@@ -18,39 +20,36 @@ pub trait DataStore {
     // ---
 
     /// Publish a resource.
-    fn publish_resource(
-        &mut self,
-        ap: &AccessPath,
-        g: (FatStructType, GlobalValue),
-    ) -> VMResult<()>;
+    fn publish_resource(&mut self, addr: AccountAddress, ty: Type, gv: GlobalValue)
```
dariorussi, line 23: I love all those changes

language/move-vm/runtime/src/move_vm.rs
```diff
@@ -21,6 +21,10 @@ impl MoveVM {
         }
     }
 
+    pub fn loader(&self) -> &Loader {
```
dariorussi, line 24: this feels really really odd....

language/move-vm/runtime/src/interpreter.rs
```diff
@@ -262,78 +258,14 @@ impl Interpreter {
         self.binop(|lhs, rhs| Ok(Value::bool(f(lhs, rhs)?)))
     }
 
-    /// Entry point for all global store operations (effectively opcodes).
```
dariorussi, line 265: all the interpreter changes are a wonderful thing to see :)

language/move-vm/runtime/src/data_cache.rs
```diff
@@ -67,149 +90,164 @@ impl<'txn> TransactionDataCache<'txn> {
...
-            Some(bytes) => Ok(bytes.clone()),
-            None => {
-                let ap = AccessPath::from(module);
-                self.data_cache.get(&ap).and_then(|data| {
-                    data.ok_or_else(|| {
-                        VMStatus::new(StatusCode::LINKER_ERROR)
-                            .with_message(format!("Cannot find {:?} in data cache", module))
-                    })
-                })
+    // REVIEW: The txn data cache isn't really caching the modules read from storage. Is this really desired?
```
dariorussi, line 210: I think that is an ok thing to do given the code cache will. Effectively `load_module` should only be called by the loader

language/move-vm/runtime/src/data_cache.rs
```diff
@@ -44,21 +58,30 @@ impl ConfigStorage for &dyn RemoteCache {
...
-pub struct TransactionDataCache<'txn> {
-    data_map: BTreeMap<AccessPath, Option<(FatStructType, GlobalValue)>>,
-    module_map: BTreeMap<ModuleId, Vec<u8>>,
-    event_data: Vec<ContractEvent>,
-    data_cache: &'txn dyn RemoteCache,
+pub struct TransactionDataCache<'r, 'l, C> {
+    data_cache: &'r C,
+    loader: &'l Loader,
+    account_map: BTreeMap<AccountAddress, AccountDataCache>,
+    event_data: Vec<(Vec<u8>, u64, Type, Value)>,
```
dariorussi, line 65: I understand the intent here but I am not really sure that removing `ContractEvent` makes things better

language/libra-vm/src/lib.rs
```diff
@@ -150,3 +155,9 @@ pub trait VMExecutor: Send {
         state_view: &dyn StateView,
     ) -> Result<Vec<TransactionOutput>, VMStatus>;
 }
+
+/// Get the AccessPath to a resource stored under `address` with type name `tag`
+fn create_access_path(address: AccountAddress, tag: StructTag) -> AccessPath {
```
dariorussi, line 160: is the caching of the resource_tag not around anymore?
that seems pretty vital to me but I may have not seen it

vgao1996: Not implemented yet. I have that planned but will do some benchmarks without caching to help us confirm the necessity of it.

language/move-vm/runtime/src/data_cache.rs
vgao1996: `ContractEvent` is now a Libra concept. What I did was essentially mirroring the concept in Move. It might make sense just to promote `ContractEvent` into Move instead. Let's think about whether that's reasonable.

language/move-vm/runtime/src/move_vm.rs
vgao1996: Yes this is indeed a temporary solution -- I'm planning to improve it in a subsequent diff. (My intuition is that I'll either implement the `Session` API I talked about previously or something similar. In other words, the client (Libra VM) really shouldn't be creating a `TransactionDataCache` on its own without resorting to the MoveVM instance.)

language/libra-vm/src/lib.rs
vgao1996: So you can imagine there are type additional caches I may implement:
1. type => type tag, type layout
  - This is in the Move VM.
2. type tag => AccessPath
  - This is in the Libra VM.

vgao1996:
@dariorussi `FatType` will be gone eventually. There's some work involved to convert its largest existing users (prop tests, fuzzing and debug) to the lighter alternatives (TypeTag, MoveTypeLayout) so I'd like to make those separate PRs. 

For a short period of time from now on, the two systems will co-exist and we'll have to rely on a few conversion functions to not break existing features.

language/move-vm/runtime/src/data_cache.rs
vgao1996: Another alternative I discussed with @runtian-zhou was to define 
```
struct LibraContractEvent<T> {
    guid: vector<u8>,
    seq_num: u64,
    msg: T,
}
```
in Move so we can have a much simpler event model in the Move VM. However this does require us to take the Value apart in the Libra VM and construct a ContractEvent there.

language/move-vm/runtime/src/move_vm.rs
dariorussi: my take is that it would be good if we did the runtime work (which is here already), keep caches change at a minimum and move the AP/FatType cache from the loader, elsewhere. But I would keep the loader reference inside the VM and maybe no try to expose much for now

language/move-vm/runtime/src/data_cache.rs
dariorussi: I understand the intent. 
However my point is that there is no concept of event today in the VM, it's not clear what that would be, it relates to native functions and more.
So conceptually this sounds good but the truth is that we are just deconstructing a `ContractEvent` and making a full signature for that. If `ContractEvent` changes we change the signature and we have not made any separation real.
We just insisted on the logical separation, which is not there today at different levels.
Once we solve the whole picture it's even possible the client of the VM has a way to plug native functions and all of that disappears from the VM code and the first thing you do is to make that back to a `ContractEvent`.

Again though, I am fine either way, let's not just focus on what would be right, if that "right" is not really reachable with the timeline we have

language/libra-vm/src/lib.rs
dariorussi: I think the cache may have decent implications to be honest and try to get to that asap.
We had to add it back because of perf related reasons so I don't see how it's not going to matter any longer. It was not just an optimization for the sake of it.
So let's get that info asap or we may hit surprises.
I hoped we could remove caches from the loader that are not related to runtime. Of course the loader would have to be used for resolution of certain type information.
Don't know we could go either way I guess.

Let's add the caches asap though

dariorussi:
> @dariorussi `FatType` will be gone eventually. There's some work involved to convert its existing users (prop tests, fuzzing and debug) to the lighter alternatives (TypeTag, MoveTypeLayout) so I'd like to make those separate PRs.
> 
> For a short period of time from now on, the two systems will co-exist and we'll have to rely on a few conversion functions to not break existing features.

I guess I am fine with that. I guess I am not clear where the caches are going to be and as mentioned I feel that is a key part of this.
Again the whole refactor of the data cache is ok but we could move all of that to another PR

language/libra-vm/src/lib.rs
vgao1996: Ah my bad I didn't make this clear. I'll definitely add the cache implementation to this PR before it lands. It's just that I wanna run benchmarks without them first so I can compare the two. I absolutely trust your judgement on the importance of the caches, just saying some first hand experiments/data will help me understand this even better.

language/move-vm/runtime/src/data_cache.rs
vgao1996: Yeah I think what you said is an accurate summarization of the current situation and I agree we don't have an event story for the Move VM yet. 

I'm also fine reverting this tuple to `ContractEvent`. I'll take another look during the final clean up of this PR and see what would be the cleanest option at that time.

language/move-vm/runtime/src/move_vm.rs
vgao1996: I should be able to keep the loader reference inside the VM with a simple change. (Adding a `fn new_txn_data_cache(&self) -> TransactionDataCache` to `MoveVM`)

jolestar:
How about the native function? Does this PR also decouple the native function?

vgao1996:
@jolestar No, the goal of this PR is just to clean up the storage interfaces between the Move VM and the outer environment  (the Libra framework in this case). 

Just to be clear we do envision a clear separation between the Libra & Move layers, and consequently, a Move VM that can be more easily plugged into other blockchains or database frameworks. However there's still some work to do before we reach that state.

Native functions are particularly tricky in the sense that some of them are "more privileged" than the others, requiring at least one of the following special capabilities:
  - Control over the underlying value representation
  - Adding a new primitive type
  - Semantics not expressible in terms of values. (e.g. the Debug module)

These special capabilities naturally couples native functions with the specific vm implementation we have right now, making it hard for us to make a direct switch to a parametric or dynamic native function registration/dispatch system. Our previous patch on this has made some progress, splitting out these special capabilities from the Move <-> Rust FFI boundaries, offering them as interfaces via the `NativeContext`, but it is still unclear what the next steps are. Let us know if you have any ideas! We really think this is an important issue in the long term and highly welcome community inputs on this!

jolestar:
@vgao1996  Thanks. This PR is a big step to reach our's goal.

I'm trying to use MoveVM in another blockchain and have encountered some requirements :

1. Adding a new primitive type, such as u256. It seems that support for primitive basic types also requires the compiler's support.
2. is Move's data persistence bound to the LCS? If yes, it seems we should include the LCS as part of the MoveVM as well.
3. Some special Native functions like debugging, event, vector, signer, which seems difficult to decouple, should be part of MoveVM. Is it possible to consider providing a mini Move stdlib? Libra and other Blockchain systems can be extended on top of this stdlib.
4. Other native methods, such as hash, signature, and so on, if the In the future, when the Move language is more powerful (for binary operations), these Native functions can be implemented directly with Move. Currently, there is only a need to provide a mechanism to register the Native Function at runtime.

vgao1996:
@jolestar 
1. Yes unless you roll your own compiler or write bytecode programs manually. 
2. Strictly speaking they don't have to, and this PR is a move in that direction. However keep in mind there's no guarantee I can give at the moment about when the total separation will happen. We're really talking about some long term goals here.
  - Note there is an `LCS` module in the stdlib right now, but technically speaking it's part of the "Libra stdlib" (see 3) so no need to worry about that.
3. You made a good point. Right now the stdlib is a mixed bag of Libra & Move stuff and we should get them separated. I'll keep track of this and push for it when the opportunity comes. 
4. Theoretically this can happen for some native functions. Although it won't be on our roadmap anytime soon and I'm not sure if we'll ever have the incentive to do so. If we want to beat or even trail native code by an acceptable margin in terms of performance, we'll have to rewrite the VM completely, potentially with a JIT compiler, which doesn't seem to be an effective use of our time. We'll likely focus on higher level features (ADT for example) for the foreseeable future, which can bring more immediate benefits to blockchain applications.

vgao1996:
@dariorussi I've added the caches and the new implementation has almost identical running time as the old one. It turned out crypto hashes for the type tags were **not** the cause of the slowdown. It was the recursive type lookups being super expensive. 

BTW I profiled `cycle_peer_to_peer_multi_block` using valgrind and generated a breakdown of the instruction usage of functions:
<script src="https://gist.github.com/vgao1996/585e7bfafb4a4e233bf52f4a918b449d.js"></script>

language/libra-vm/src/libra_vm.rs
```diff
@@ -332,115 +354,118 @@ impl LibraVM {
...
-                gas_schedule,
-                failed_gas_left,
-                txn_data,
-                remote_cache,
-                account_currency_symbol,
-            )
-        })
+
+        let mut cost_strategy = CostStrategy::system(gas_schedule, cost_strategy.remaining_gas());
+        unwrap_or_fail!(self.run_success_epilogue(
```
dariorussi, line 419: do we really need the macros here? just curious...

vgao1996: Hmm... Just to clarify, are you questioning the existence of the macro itself, or certain use cases here?

language/e2e-tests/src/data_store.rs
```diff
@@ -113,8 +119,51 @@ impl StateView for FakeDataStore {
...
 
 // This is used by the `process_transaction` API.
-impl RemoteCache for FakeDataStore {
+/*impl RemoteCache for FakeDataStore {
     fn get(&self, access_path: &AccessPath) -> VMResult<Option<Vec<u8>>> {
         Ok(StateView::get(self, access_path).expect("it should not error"))
     }
+}*/
+
+impl RemoteCache for FakeDataStore {
```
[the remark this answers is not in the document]
vgao1996: Yes, the current implementation is a bit messy. Alternatively, maybe instead of implementing `RemoteCache` for `FakeDataStore` directly, we should just wrap it inside a `RemoteStorage`.

Anyways it's a bit hard for me to tell exactly what the details should be right now, as I'm still finalizing the `RemoteCache` API and may take a step back, making `get_module` and `get_resource` return bytes instead. I'll review the whole story later.

language/libra-vm/src/libra_vm.rs
```diff
@@ -332,115 +354,112 @@ impl LibraVM {
         payload: VerifiedTransactionPayload,
         account_currency_symbol: &IdentStr,
     ) -> TransactionOutput {
-        let gas_schedule = match self.get_gas_schedule() {
+        let gas_schedule = match gas_schedule!(self) {
```
runtian-zhou, line 357: Why do we need to change the macro over here?

language/libra-vm/src/libra_vm.rs
```diff
@@ -332,115 +354,112 @@ impl LibraVM {
...
-                        self.move_vm.execute_script(
-                            s,
-                            ty_args,
-                            args,
-                            txn_data.sender(),
-                            &mut data_store,
-                            &mut cost_strategy,
-                        )
-                    });
+                unwrap_or_fail!(cost_strategy.charge_intrinsic_gas(txn_data.transaction_size()));
```
runtian-zhou, line 396: Love the cleanup here!

language/move-core/types/src/value.rs
```diff
@@ -40,6 +40,23 @@ pub enum MoveTypeLayout {
     Signer,
 }
 
+#[derive(Debug, Clone)]
+pub enum MoveKindInfo {
+    Base(bool),
```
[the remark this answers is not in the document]
vgao1996: It's used by the VM. When deserializing a struct, we need this information to tell whether the struct or its fields are resources or not.

language/move-vm/runtime/src/data_cache.rs
vgao1996: Let me move `ContractEvent` into move core types and get rid of the tuple.

language/move-core/types/src/value.rs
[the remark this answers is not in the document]
vgao1996: I'd actually argue that the enum variants here are used to describe the recursive structure of the data, not the resource-ness of it, which is the job of the bool:
- `Base` is when you reach a leaf location, hence the base case
- `Vector` and `Struct` are the recursive cases

If we want to merge these two discriminators, we'll end up having 3 * 2 = 6 variants. (It's the cartesian product)
```
pub enum MoveKindInfo {
    CopyableBase,
    ResourceBase,
    CopyableVec(Vec<MoveKindInfo>),
    ResourceVec(Vec<MoveKindInfo>),
    CopyableStruct(Vec<MoveKindInfo>),
    ResourceStruct(Vec<MoveKindInfo>),
}
```

I do understand your concern about readability. How about replacing the bool with an enum with named variants?

vgao1996:
@dariorussi @runtian-zhou I just updated the PR and added a new `Session` API. This keeps the `Loader` and `TransactionDataCache` private to the move vm. Make sure to refresh to page so you're not staring at the old code!

language/move-core/types/src/value.rs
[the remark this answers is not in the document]
runtian-zhou: That seems a bit better.  What't ur take on this @dariorussi ?

> it's technically not used in the VM 

What I meant by this is that the VM doesn't need to distinguish between resource & value in its implementation, but it's a nice extra safety feature we can enforce in the implementation of VM.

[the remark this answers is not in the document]
vgao1996: Fully agreed and my understanding is that this extra safety feature will stay for v1.

language/libra-vm/src/libra_vm.rs
vgao1996: Good question. TLDR: to make the borrow checker happy. This all comes down to a limitation that borrow checking stops at function boundaries. When you have a struct with multiple fields, it is perfectly fine to borrow the fields in a disjoint manner as the borrow checker is smart enough to track each field separately. However such information is not available across function boundaries. If you call a function that takes a reference to the struct, the borrow checker has to assume the whole struct is borrowed, even if the function only uses some of the fields.
```
struct Foo {
    counter: usize,
    v: Vec<usize>,
}

fn bar(v: &Vec<usize>) {
    println!("{:?}", v);
}

impl Foo {
    fn f(&mut self) {
        let v = &self.v;
        self.counter += 1; // good
        bar(v);
    }
    
    fn g(&mut self) {
        let v = &self.v;
        self.inc_counter(); // bad
        bar(v);
    }
    
    fn inc_counter(&mut self) {
        self.counter += 1;
    }
}
```

With that being said, I'll take another look at this and see if the macro is necessarily or I can just write the code inlined.

vgao1996:
@runtian-zhou I'm more inclined to NOT promote `ContractEvent` into a Move type right now. The reason being `ContractEvent` is versioned and I'm really unsure if this versioning story aligns with the Move side of things. I feel like we need to spend a little bit more time to really figure this out. Thoughts?

language/libra-vm/src/libra_vm.rs
dariorussi: I am questioning the usage of the macro altogether yeah :)

dariorussi:
> @runtian-zhou I'm more inclined to NOT promote `ContractEvent` into a Move type right now. The reason being `ContractEvent` is versioned and I'm really unsure if this versioning story aligns with the Move side of things. I feel like we need to spend a little bit more time to really figure this out. Thoughts?

personally I think using `ContractEvent` is the cleanest way to do it. But I also think it's not a big deal one way or the other.
So overall let's not struggle too much with whatever we have.
Once we make a proper story for native functions all of that is likely to be a moot point

language/libra-vm/src/libra_vm.rs
vgao1996: It's used to emulate a "?" operator with custom behavior. I also had a discussion with @tnowacki as he also ran into this issue. It seems to me this is the most sensible solution that exists in stable rust for now:
-  `?` cannot be overridden  
- `try_blocks` is only available on nightly
- Combinators are hard to understand and cannot handle mutability
  - IMO it's much easier to explicitly sequence the side effects in this case

vgao1996:
@dariorussi Right now `ContractEvent` is tightly coupled with certain libra concepts so it's hard to move it into `move_core_types`. The usual trick of leaving a `pub use` in the original place wouldn't work as we derive `LCSCryptoHash` and `CryptoHasher` for it and it seems to be a bad idea to add those dependencies to `move_core_types`. 

Therefore if I were to use `ContractEvent` I would have to add back a dependency from `move_vm_runtime` to `libra_types`, which kinda irks me as we only have a handful Libra dependencies in `move_vm_runtime` right now:
1) `HashValue` for script cache
2) `CORE_CODE_ADDRESS` for native functions
3) The logger
Maybe it's just my personal taste but I think a translation layer in this case is a bit simpler and cleaner in a conceptual sense.

BTW `RemoteCache` now returns bytes. Might still make some minor cleanups but you can consider this PR ready for a final review.

language/move-vm/runtime/src/data_cache.rs
```diff
@@ -67,149 +85,166 @@ impl<'txn> TransactionDataCache<'txn> {
...
-        g: (FatStructType, GlobalValue),
-    ) -> VMResult<()> {
-        self.data_map.insert(ap.clone(), Some(g));
+impl<'r, 'l, C: RemoteCache> DataStore for TransactionDataCache<'r, 'l, C> {
+    fn publish_resource(&mut self, addr: AccountAddress, ty: Type, g: GlobalValue) -> VMResult<()> {
+        let account_cache = Self::get_mut_or_insert_with(&mut self.account_map, &addr, || {
+            (addr, AccountDataCache::new())
+        });
+
+        // TODO: shall we check if resource already exists here?
```
dariorussi, line 184: don't we check in `data_operation.rs`?

language/move-vm/runtime/src/data_cache.rs
```diff
@@ -67,149 +85,166 @@ impl<'txn> TransactionDataCache<'txn> {
...
-            None => {
-                let ap = AccessPath::from(module);
-                self.data_cache.get(&ap).and_then(|data| {
-                    data.ok_or_else(|| {
-                        VMStatus::new(StatusCode::LINKER_ERROR)
-                            .with_message(format!("Cannot find {:?} in data cache", module))
-                    })
-                })
+    // REVIEW: The txn data cache isn't really caching the modules read from storage. Is this really desired?
+    fn load_module(&self, module_id: &ModuleId) -> VMResult<CompiledModule> {
```
dariorussi, line 209: I guess this is fine for now but we'll review later.
I liked the idea that all the operation to deserialize and verify are inside the loader and I would love if deserialization was there too.
We'll see if that is doable once this land and we'll go from there

language/move-vm/runtime/src/data_cache.rs
```diff
@@ -1,33 +1,42 @@
...
 /// Can be used to define a "fake" implementation of the remote cache.
 pub trait RemoteCache {
-    fn get(&self, access_path: &AccessPath) -> VMResult<Option<Vec<u8>>>;
+    fn get_module(&self, module_id: &ModuleId) -> VMResult<Option<Vec<u8>>>;
+    fn get_resource(&self, address: &AccountAddress, tag: &TypeTag) -> VMResult<Option<Vec<u8>>>;
+}
+
+pub struct AccountDataCache {
+    data_map: BTreeMap<Type, Option<GlobalValue>>,
+    module_map: BTreeMap<Identifier, (Vec<u8>, CompiledModule)>,
```
dariorussi, line 31: as per comment below, I am not sure `CompiledModule` is proper here, but I am not clear yet if removing it is an issue

language/move-vm/runtime/src/loader.rs
```diff
@@ -107,7 +105,7 @@ impl ScriptCache {
 // It holds all Modules, Types and Functions loaded.
 // Types and Functions are pushed globally to the ModuleCache.
 // A ModuleCache is accessed under lock.
-struct ModuleCache {
+pub(crate) struct ModuleCache {
```
dariorussi, line 108: why is this going out? that sounds a tad scary....

language/move-vm/runtime/src/loader.rs
```diff
@@ -307,22 +305,51 @@ impl ModuleCache {
...
+        Self {
+            struct_tag: None,
+            struct_layout: None,
+            kind_info: None,
+        }
+    }
+}
+
+pub(crate) struct TypeCache {
+    structs: BTreeMap<usize, BTreeMap<Vec<Type>, StructInfo>>,
```
dariorussi, line 325: not sure we need the BTreeMap but I guess that is fine and the `usize` key is probably just perfect for it

language/move-vm/runtime/src/loader.rs
```diff
@@ -618,13 +645,7 @@ impl Loader {
         data_store: &mut dyn DataStore,
     ) -> VMResult<CompiledModule> {
         let module = match data_store.load_module(id) {
```
dariorussi, line 647: yeah that is the part I was questioning with the comment I had. I would love for this code to stay here rather than relying on the data cache to pass a valid `CompiledModule`

language/move-vm/runtime/src/loader.rs
```diff
@@ -1442,6 +1378,193 @@ impl Loader {
         })
     }
 
+    fn struct_gidx_to_type_tag(&self, gidx: usize, ty_args: &[Type]) -> VMResult<StructTag> {
```
dariorussi, line 1381: I wonder if all this code and generally the code about the type cache could be moved elsewhere (different file) and leave the loader a bit more about loading.
We don't want it in this PR and it's more of a question if w could do it

language/move-vm/runtime/src/move_vm.rs
```diff
@@ -21,53 +24,74 @@ impl MoveVM {
...
+    }
+}
+
+impl Default for MoveVM {
+    fn default() -> Self {
+        Self::new()
+    }
+}
+
+pub struct Session<'r, 'l, R> {
```
dariorussi, line 41: I am not clear about the `Session` story but I think it's good enough for now and we can go out that way!

language/move-vm/runtime/src/runtime.rs
```diff
@@ -29,6 +29,10 @@ impl VMRuntime {
         }
     }
 
+    pub(crate) fn loader(&self) -> &Loader {
```
dariorussi, line 32: wasn't this supposed to go?

language/tools/vm-genesis/src/genesis_context.rs
```diff
@@ -5,125 +5,10 @@
 
 use anyhow::Result;
 use libra_state_view::StateView;
-use libra_types::{
```
dariorussi, line 8: @runtian-zhou @tzakian not sure who added this but let me flag it just in case :)

language/move-vm/runtime/src/data_cache.rs
```diff
@@ -44,21 +53,30 @@ impl ConfigStorage for &dyn RemoteCache {
...
+    )>,
+    pub modules: Vec<(ModuleId, Vec<u8>)>,
+    pub events: Vec<(Vec<u8>, u64, TypeTag, MoveTypeLayout, Value)>,
+}
+
+impl<'r, 'l, C: RemoteCache> TransactionDataCache<'r, 'l, C> {
     /// Create a `TransactionDataCache` with a `RemoteCache` that provides access to data
     /// not updated in the transaction.
-    pub fn new(data_cache: &'txn dyn RemoteCache) -> Self {
+    pub(crate) fn new(data_cache: &'r C, loader: &'l Loader) -> Self {
```
dariorussi, line 75: this makes me think that a cache separate from the loader would be better

language/move-vm/runtime/src/loader.rs
vgao1996: It doesn't have to. It's very likely I changed it for some reason that no longer exists. Will revert it before landing. Good catch BTW!

language/move-vm/runtime/src/loader.rs
vgao1996: I didn't have a strong reason leaning towards any of these two, so tho choice was somewhat arbitrary. Let me switch to HashMap so it aligns with the other caches.

language/move-vm/runtime/src/runtime.rs
vgao1996: It's used to create a `Session`, but yeah I can change it to a `pub(crate) fn new_session` so to not expose the `Loader`.

language/move-vm/runtime/src/data_cache.rs
vgao1996: That is an interesting idea. I'd actually think about this within a bigger context: how do we view the `Type`s stored in the `TransactionDataCache`? There are two possible approaches I can see here:
1) `TransactionDataCache` must know the meanings of a `Type`, which means it must contain a reference to the loader. 
2) `TransactionDataCache` treats `Type`s as black boxes and only use them as keys. My guess is that this will make the `TransactionDataCache` more light-weighted as we'll move some logic to the callee positions (`Session` for example). 

Anyways one advantage we have right now is that `TransactionDataCache` is a private to the crate so it should be much easier for us to change it in the future.

dariorussi: I think 2) is what we should do. Basically that data cache treats Types as keys and if the layout was not loaded yet the loader is accessed to get the info to create a layout.
But anyway that is obviously for another

language/move-vm/runtime/src/loader.rs
vgao1996: I kinda get what you mean here. I think this also depends on how we view the `Type` similar to the case for `TransactionDataCache`. Definitely something for us to think about.

language/move-vm/runtime/src/data_cache.rs
vgao1996: Good point. I'll think about this and put up a follow up PR if necessary.

vgao1996:
@dariorussi Thanks for the comments! I'll land this PR in its current state. There are still some issues around the interactions between the `TransactionDataCache` and I'll into those later and put up separate PRs if necessary.

PR #4719, opened 2020-06-24
[language] Consolidate loading\verification in the runtime

dariorussi:
This PR makes few changes bundled together for convenience.
* All verification passes are with the same model, a static function called `verify`, 
so `VerificationCheck::verify(&CompiledModule) -> VMResult<()>`. This is a bunch of mechanical changes.
* `VerifiedModule` usage is removed from a bunch of places. 
The runtime, some compiler stuff, and functional tests should have a very limited use of VerifiedModule.
* The VM runtime uses the passes directly and logic for verification and loading is unified

### DIFF
--- a/language/bytecode-verifier/bytecode-verifier-tests/src/unit_tests/constants_tests.rs
+++ b/language/bytecode-verifier/bytecode-verifier-tests/src/unit_tests/constants_tests.rs
@@ -8,7 +8,7 @@ use vm::file_format::{empty_module, CompiledModule, Constant, SignatureToken};
 proptest! {
     #[test]
     fn valid_generated(module in CompiledModule::valid_strategy(20)) {
-        prop_assert!(ConstantsChecker::new(&module).verify().is_ok());
+        prop_assert!(ConstantsChecker::verify(&module).is_ok());
     }
 }
 
@@ -38,7 +38,7 @@ fn valid_primitives() {
         },
     ];
     let module = module_mut.freeze().unwrap();
-    assert!(ConstantsChecker::new(&module).verify().is_ok());
+    assert!(ConstantsChecker::verify(&module).is_ok());
 }
 
 #[test]
@@ -142,7 +142,7 @@ fn valid_vectors() {
         },
     ];
     let module = module_mut.freeze().unwrap();
-    assert!(ConstantsChecker::new(&module).verify().is_ok());
+    assert!(ConstantsChecker::verify(&module).is_ok());
 }
 
 #[test]
@@ -207,8 +207,7 @@ fn error(type_: SignatureToken, data: Vec<u8>, code: StatusCode) {
     let mut module_mut = empty_module();
     module_mut.constant_pool = vec![Constant { type_, data }];
     assert!(
-        ConstantsChecker::new(&module_mut.freeze().unwrap())
-            .verify()
+        ConstantsChecker::verify(&module_mut.freeze().unwrap())
             .unwrap_err()
             .major_status
             == code
--- a/language/bytecode-verifier/src/lib.rs
+++ b/language/bytecode-verifier/src/lib.rs
@@ -38,3 +38,52 @@ pub use verifier::{
     batch_verify_modules, verify_main_signature, verify_module_dependencies,
     verify_script_dependencies, VerifiedModule, VerifiedScript,
 };
+use vm::{
+    access::ModuleAccess,
+    file_format::{Kind, SignatureToken},
+    CompiledModule,
+};
+
+//
+// Helpers functions for signatures
+//
+
+pub(crate) fn kind(module: &CompiledModule, ty: &SignatureToken, constraints: &[Kind]) -> Kind {
+    use SignatureToken::*;
+
+    match ty {
+        // The primitive types & references have kind unrestricted.
+        Bool | U8 | U64 | U128 | Address | Reference(_) | MutableReference(_) => Kind::Copyable,
+        Signer => Kind::Resource,
+        TypeParameter(idx) => constraints[*idx as usize],
+        Vector(ty) => kind(module, ty, constraints),
+        Struct(idx) => {
+            let sh = module.struct_handle_at(*idx);
+            if sh.is_nominal_resource {
+                Kind::Resource
+            } else {
+                Kind::Copyable
+            }
+        }
+        StructInstantiation(idx, type_args) => {
+            let sh = module.struct_handle_at(*idx);
+            if sh.is_nominal_resource {
+                return Kind::Resource;
+            }
+            // Gather the kinds of the type actuals.
+            let kinds = type_args
+                .iter()
+                .map(|ty| kind(module, ty, constraints))
+                .collect::<Vec<_>>();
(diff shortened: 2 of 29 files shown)


dariorussi:
I have not removed `VerifiedModule` entirely and I am looking for feedback on whether you want me to do it or not (in another PR obviously). I am even fine as things are after this PR with a possible minor cleanup coming.
I know people feel differently about it so please comment. @tnowacki, @vgao1996 take a look. I think @tzakian liked to have `VerifiedModule` in some context (I am sure @sunshowers has feelings about it too). Everybody, really, feel free to comment.
The VM is not using `VerifiedModule` and that part I feel pretty strong about, I can clarify if you want me to and let me know.
The loader and the consolidation of verification and loading needs a bit more cleanup but I think this PR is a good improvement over what we have and in a decent shape.
Some inline comment may help when reviewing...

language/move-vm/types/src/values/values_impl.rs
```diff
@@ -245,15 +245,23 @@ impl ValueImpl {
 }
 
 impl Value {
-    pub fn is_valid_script_arg(&self, sig: &SignatureToken) -> bool {
+    // REVIEW: this allows primitive (including `Sender`) and vectors of them. Vectors can only
+    // be one level (a different organization of `ValueImpl` could give us that property in an
+    // efficient way). Conceptually if we serialize the value and deserialize according to
+    // the argument type we would have accomplished a proper check
+    pub fn is_valid_arg(&self, sig: &SignatureToken) -> bool {
```
dariorussi, line 252: this is an interpreter entry point that needs some work/love from @vgao1996.
In principle here we would like to accept anything as long as it matches the type but that may be a difficult story.
Also changes in `ValueImpl` may affect this so @vgao1996 please take a look and think about it

language/move-vm/types/src/values/values_impl.rs
```diff
@@ -263,6 +271,33 @@ impl Value {
                 (SignatureToken::Signer, Container::Resource(v)) => {
                     v.len() == 1 && matches!(&v[0], ValueImpl::Address(_))
                 }
+                _ => true,
+            },
+            _ => false,
+        }
+    }
+
+    pub fn is_constant_or_sender_ref(&self) -> bool {
```
dariorussi, line 280: this is just a check that the type is a constant or a sender reference. In any case this is the check from the Move Adapter that an argument is from a limited set of types

language/move-vm/runtime/src/runtime.rs
```diff
@@ -145,21 +142,13 @@ impl VMRuntime {
...
-                signature.len(),
-                args.len()
-            )),
-        );
-    }
-    for (tok, val) in signature.0.iter().zip(args) {
-        if !val.is_valid_script_arg(tok) {
+/// Check that the transaction arguments are acceptable by the VM.
+/// Constants are the only arguments allowed.
+fn check_args(args: &[Value]) -> VMResult<()> {
```
dariorussi, line 147: at this level we only check that the arguments are restricted.
The interpreter when setting up the initial frame verifies arguments for both scripts and functions

language/move-vm/types/src/values/values_impl.rs
vgao1996: Yeah as you mentioned offline, we should come up with a function checking if a `Value` matches its type here. I'll take over this part once I'm back to work on the values. (It should be easier with the cleanup in the data cache.)

language/bytecode-verifier/bytecode-verifier-tests/src/unit_tests/constants_tests.rs
```diff
@@ -8,7 +8,7 @@ use vm::file_format::{empty_module, CompiledModule, Constant, SignatureToken};
 proptest! {
     #[test]
     fn valid_generated(module in CompiledModule::valid_strategy(20)) {
-        prop_assert!(ConstantsChecker::new(&module).verify().is_ok());
+        prop_assert!(ConstantsChecker::verify(&module).is_ok());
```
tnowacki, line 11: Hot

language/bytecode-verifier/src/lib.rs
```diff
@@ -38,3 +38,52 @@ pub use verifier::{
...
+    access::ModuleAccess,
+    file_format::{Kind, SignatureToken},
+    CompiledModule,
+};
+
+//
+// Helpers functions for signatures
+//
+
+pub(crate) fn kind(module: &CompiledModule, ty: &SignatureToken, constraints: &[Kind]) -> Kind {
```
tnowacki, line 51: A weird spot for this, but significantly less weird than the old spot

language/move-vm/types/src/values/values_impl.rs
tnowacki: nit, `s/sender/signer/` ?

tnowacki:
Maybe rename the PR since VerifiedModule still exists? Its just gone from the VM (I think)

language/move-vm/types/src/values/values_impl.rs
```diff
@@ -245,15 +245,23 @@ impl ValueImpl {
 }
 
 impl Value {
-    pub fn is_valid_script_arg(&self, sig: &SignatureToken) -> bool {
+    // REVIEW: this allows primitive (including `Sender`) and vectors of them. Vectors can only
```
tnowacki, line 248: s/Sender/Signer

sunshowers:
I don't have much to add other than that I introduced `VerifiedModule` specifically because we had code paths which were accidentally not verifying modules, and that I thought leveraging the type system to assert a verification invariant was a good idea (it found bugs!). I'm worried about us regressing in the future, given that maintaining correctness over time is far more important than simply ensuring correctness today.

dariorussi:
> I don't have much to add other than that I introduced `VerifiedModule` specifically because we had code paths which were accidentally not verifying modules, and that I thought leveraging the type system to assert a verification invariant was a good idea (it found bugs!). I'm worried about us regressing in the future, given that maintaining correctness over time is far more important than simply ensuring correctness today.

I would agree that the way the system was architected before the `VerifiedModule` had more of a role.
From the VM perspective now, there is a single code path to code loading in the system and I think it's important to make sure the runtime path is well defined and the only source of truth for what verification for the runtime means (that is obvious but I think it's important to point out).
We could stick with the `VerifiedModule` but it's not adding much value given it is not something that is passed around. The model is more that we verify, and we transform while loading. In that sense a type transformation only adds a layer (logically and while coding, in terms of code generation of course `VerifiedModule` was mostly a no-op) that it is making things less obvious.

Also for verification I do believe a "library" of steps/functions is more appropriate. I think clients can (maybe they should) use a type transformation in their component if that was helpful (e.g. the instance travels around in API). But I am not clear a universal concept of `VerifiedModule` is helpful. 

A latest talk with the team was suggesting we could still have a VerifiedModule::verify(&CompiledModule) which performs a set of "canonical" steps for verification but, for me at least, having the runtime owning its own logic "in place" is still the right thing.

Not sure if you have thoughts about this

PR #4720, opened 2020-06-24
[cluster-test] Implement a cluster test to test validator update behavior

runtian-zhou:
## Motivation

Implement a cluster test to test networks behavior when we have multiple versions of validator software running together. This test is built upon the hypothetical feature in #4416 where we would like to reject a transaction with gas equals to a magic number.

Note that you would need to rebase #4416 first when trying to run this test, otherwise other breaking changes in the system might cause the test to fail.

Yes

## Test Plan

Commands to run the test:
1. Rebase and build the image for #4416 by: `./docker/build-aws.sh --build-all --version pull/4416 --addl_tags version_test_2`
2. Run the cluster test:
`./scripts/cti --pr 4720 --run versioning_testing -- --updated-image-tag version_test_2`

### DIFF
--- a/testsuite/cluster-test/src/experiments/versioning_test.rs
+++ b/testsuite/cluster-test/src/experiments/versioning_test.rs
@@ -1,0 +1,274 @@
+// Copyright (c) The Libra Core Contributors
+// SPDX-License-Identifier: Apache-2.0
+
+#![forbid(unsafe_code)]
+
+use std::{collections::HashSet, fmt, time::Duration};
+
+use rand::Rng;
+
+use crate::{
+    cluster::Cluster,
+    experiments::{Context, Experiment, ExperimentParam},
+    instance,
+    instance::Instance,
+    tx_emitter::{execute_and_wait_transactions, EmitJobRequest},
+};
+use async_trait::async_trait;
+use futures::future::try_join_all;
+use libra_logger::prelude::*;
+use libra_types::{
+    account_config::{lbr_type_tag, LBR_NAME},
+    on_chain_config::LibraVersion,
+    transaction::{helpers::create_user_txn, TransactionPayload},
+};
+use structopt::StructOpt;
+use tokio::time;
+use transaction_builder::{
+    encode_transfer_with_metadata_script, encode_update_libra_version_script,
+};
+
+#[derive(StructOpt, Debug)]
+pub struct ValidatorVersioningParams {
+    #[structopt(long, default_value = "10", help = "Number of nodes to reboot")]
+    pub count: usize,
+    #[structopt(long, help = "Image tag of newer validator software")]
+    pub updated_image_tag: String,
+}
+
+pub struct ValidatorVersioning {
+    first_batch: Vec<Instance>,
+    second_batch: Vec<Instance>,
+    full_nodes: Vec<Instance>,
+    updated_image_tag: String,
+}
+
+impl ExperimentParam for ValidatorVersioningParams {
+    type E = ValidatorVersioning;
+    fn build(self, cluster: &Cluster) -> Self::E {
+        if self.count > cluster.validator_instances().len() {
+            panic!(
+                "Can not reboot {} validators in cluster with {} instances",
+                self.count,
+                cluster.validator_instances().len()
+            );
+        }
+        let mut instances = Vec::with_capacity(self.count);
+        let mut all_instances = cluster.validator_instances().to_vec();
+        let mut rnd = rand::thread_rng();
+        for _i in 0..self.count {
+            let instance = all_instances.remove(rnd.gen_range(0, all_instances.len()));
+            instances.push(instance);
+        }
+
+        Self::E {
+            first_batch: instances,
+            second_batch: all_instances,
+            full_nodes: cluster.fullnode_instances().to_vec(),
+            updated_image_tag: self.updated_image_tag,
+        }
+    }
+}
+
+#[async_trait]
+impl Experiment for ValidatorVersioning {
+    fn affected_validators(&self) -> HashSet<String> {
+        instance::instancelist_to_set(&self.first_batch)
+            .union(&instance::instancelist_to_set(&self.second_batch))
+            .cloned()
+            .collect()
+    }
+
+    async fn run(&mut self, context: &mut Context<'_>) -> anyhow::Result<()> {
+        // Mint a number of accounts
+        context
+            .tx_emitter
+            .mint_accounts(
+                &EmitJobRequest::for_instances(
+                    context.cluster.validator_instances().to_vec(),
+                    context.global_emit_job_request,
+                ),
+                150,
+            )
+            .await?;
+
+        info!("1. Changing the images for the instances in the first batch");
+        let futures: Vec<_> = self.first_batch.iter().map(Instance::stop).collect();
+        try_join_all(futures).await?;
+        let futures: Vec<_> = self
+            .first_batch
+            .iter()
+            .map(|instance| {
+                let mut newer_config = instance.instance_config().clone();
+                newer_config
+                    .replace_tag(self.updated_image_tag.clone())
+                    .unwrap();
+                context
+                    .cluster_swarm
+                    .spawn_new_instance(newer_config, false)
+            })
+            .collect();
+        try_join_all(futures).await?;
+        time::delay_for(Duration::from_secs(20)).await;
+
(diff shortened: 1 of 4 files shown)


testsuite/cluster-test/src/experiments/versioning_test.rs
```diff
@@ -0,0 +1,264 @@
...
+        // Change the images for the instances in the first batch
+        let futures: Vec<_> = self.first_batch.iter().map(Instance::stop).collect();
+        try_join_all(futures).await?;
+        let futures: Vec<_> = self
+            .first_batch
+            .iter()
+            .map(|_| {
+                context
+                    .cluster_swarm
+                    .spawn_new_instance(self.updated_image_config.clone(), false)
```
andll, line 125: Note InstanceConfig contain things like validator index, etc.

So you would ned to update config for each instance individually, otherwise you are just starting copy of the same instance multiple times

ankushagarwal:
First external experiment contribution to cluster-test! 🎉

andll:
Btw, I assume this diff is still WIP? @runtian-zhou can you ping me when you test it and think it's ready for review?

runtian-zhou:
> Btw, I assume this diff is still WIP? @runtian-zhou can you ping me when you test it and think it's ready for review?

This is indeed a WIP right now! I'll update the description once it is ready!

dariorussi:
this is pretty cool and very happy to see it. Looking forward to see this test running. Thank you @runtian-zhou and @andll

testsuite/cluster-test/src/experiments/versioning_test.rs
```diff
@@ -0,0 +1,262 @@
...
+                .spawn_new_instance(newer_config, false)
+        })
+        .collect();
+    let instances = try_join_all(futures).await?;
+
+    // 3. Wait for the instances to recover.
+    let futures: Vec<_> = instances
+        .iter()
+        .map(|instance| instance.wait_json_rpc(deadline))
+        .collect();
```
andll, line 104: Similarly, I think you don't need collect here

testsuite/cluster-test/src/experiments/versioning_test.rs
```diff
@@ -0,0 +1,262 @@
...
+
+        let txn_payload = TransactionPayload::Script(encode_transfer_with_metadata_script(
+            lbr_type_tag(),
+            account_2.address,
+            1,
+            vec![],
+            vec![],
+        ));
+
+        let txn1 = create_user_txn(
```
andll, line 148: This call (as well as `.sequence_number+=1`) seem to be repeating in this fn - since this is fairly large call, do you think it make sense to make separate function for it?

testsuite/cluster-test/src/experiments/versioning_test.rs
```diff
@@ -0,0 +1,262 @@
...
+            0,
+            LBR_NAME.to_owned(),
+            10,
+        )
+        .expect("Failed to create signed transaction");
+
+        full_node_client
+            .submit_transaction(txn3.clone())
+            .await
+            .expect("Transaction should pass the full node mempool");
```
andll, line 223: Can you use `.map_err(|e|format_err!("Transaction should pass the full node mempool: {}", e)?` instead of expect?

testsuite/cluster-test/src/experiments/versioning_test.rs
```diff
@@ -0,0 +1,262 @@
...
+            &account_1.key_pair,
+            txn_payload,
+            account_1.address,
+            account_1.sequence_number,
+            123456,
+            0,
+            LBR_NAME.to_owned(),
+            10,
+        )
+        .expect("Failed to create signed transaction");
```
andll, line 218: Can you use `.map_err(|e|format_err!("<message>: {}", e)?` instead of expect?

testsuite/cluster-test/src/experiments/versioning_test.rs
```diff
@@ -0,0 +1,262 @@
...
+
+        full_node_client
+            .submit_transaction(txn3.clone())
+            .await
+            .expect("Transaction should pass the full node mempool");
+
+        execute_and_wait_transactions(&mut full_node_client, &mut account_1, vec![txn3.clone()])
+            .await
+            .unwrap_err();
```
andll, line 227: Can you use `?` instead of panic?

testsuite/cluster-test/src/experiments/versioning_test.rs
```diff
@@ -0,0 +1,262 @@
...
+            .json_rpc_client();
+        updated_full_node
+            .submit_transaction(txn3)
+            .await
+            .unwrap_err();
+        Ok(())
+    }
+
+    fn deadline(&self) -> Duration {
+        Duration::from_secs(20 * 60)
```
andll, line 246: How long does this experiment take usually?
Good guideline for timeout is 2*<typical_execution_time>

testsuite/cluster-test/src/tx_emitter.rs
```diff
@@ -156,6 +156,10 @@ impl TxEmitter {
         }
     }
 
+    pub fn take_account(&mut self) -> AccountData {
+        self.accounts.remove(0)
```
andll, line 160: nit: `self.accounts.remove(self.accounts.len()-1)` is faster :)

testsuite/cluster-test/src/experiments/versioning_test.rs
```diff
@@ -0,0 +1,263 @@
...
+        info!("7. Change the images for the full nodes");
+        update_batch_instance(context, &self.full_nodes, self.updated_image_tag.clone()).await?;
+
+        info!("8. Send a transaction to make sure it gets dropped by the full node mempool.");
+
+        let updated_full_node = context
+            .cluster
+            .random_full_node_instance()
+            .json_rpc_client();
+        if updated_full_node.submit_transaction(txn3).await.is_ok() {
```
andll, line 238: I think `if let Err(e)` pattern will work better here, because it will allow to pop error cause on top, e.g.
```
return Err(format_err!(
                "Transaction should not be accepted by the full node.: {}", e
            ));
```

This was proven to be important when debugging problem (not propagating root cause makes it hard to debug what was wrong)

testsuite/cluster-test/src/experiments/versioning_test.rs
```diff
@@ -0,0 +1,263 @@
...
+            .submit_transaction(txn3.clone())
+            .await
+            .map_err(|e| format_err!("Transaction should pass the full node mempool: {}", e))?;
+
+        if execute_and_wait_transactions(&mut full_node_client, &mut account_1, vec![txn3.clone()])
+            .await
+            .is_ok()
+        {
+            return Err(format_err!(
+                "Transaction should not be committed by the validators"
```
andll, line 225: Need to propagate root cause(see comment below)

testsuite/cluster-test/src/experiments/versioning_test.rs
andll: Did you think this can be refactored to avoid code duplication below?

testsuite/cluster-test/src/experiments/versioning_test.rs
```diff
@@ -0,0 +1,263 @@
...
+    }
+}
+
+// Reboot `updated_instance` with newer image tag
+async fn update_batch_instance(
+    context: &mut Context<'_>,
+    updated_instance: &[Instance],
+    updated_tag: String,
+) -> anyhow::Result<()> {
+    let deadline = Instant::now() + Duration::new(60, 0);
```
andll, line 78: suggestion: `Duration::from_sec(60);`

testsuite/cluster-test/src/experiments/versioning_test.rs
runtian-zhou: We can't use this pattern here :( . The main reason is that we are expecting this call to fail. Thus any error over there is actually desirable in this case.

testsuite/cluster-test/src/experiments/versioning_test.rs
runtian-zhou: Same here as well.

testsuite/cluster-test/src/experiments/versioning_test.rs
andll: Right, got it!

PR #4724, opened 2020-06-24
[config] fix local file endings to match contents

gregnazario:
Several files are using "toml" as the file ending but are actually using
json or yaml.  Made it a little confusing surprise every time the files are opened!

### DIFF
--- a/config/config-builder/src/main.rs
+++ b/config/config-builder/src/main.rs
@@ -9,8 +9,8 @@ use libra_network_address::NetworkAddress;
 use std::{convert::TryInto, fs, fs::File, io::Write, net::SocketAddr, path::PathBuf};
 use structopt::StructOpt;
 
-const KEY_MANAGER_CONFIG: &str = "key_manager.config.toml";
-const NODE_CONFIG: &str = "node.config.toml";
+const KEY_MANAGER_CONFIG: &str = "key_manager.yaml";
+const NODE_CONFIG: &str = "node.yaml";
 
 #[allow(clippy::large_enum_variant)]
 #[derive(Debug, StructOpt)]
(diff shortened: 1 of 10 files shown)


config/src/config/secure_backend_config.rs
```diff
@@ -104,7 +104,7 @@ impl Default for OnDiskStorageConfig {
     fn default() -> Self {
         Self {
             namespace: None,
-            path: PathBuf::from("secure_storage.toml"),
+            path: PathBuf::from("secure_storage.json"),
```
davidiw, line 107: this is fine!

config/config-builder/src/main.rs
```diff
@@ -9,8 +9,8 @@ use libra_network_address::NetworkAddress;
 use std::{convert::TryInto, fs, fs::File, io::Write, net::SocketAddr, path::PathBuf};
 use structopt::StructOpt;
 
-const KEY_MANAGER_CONFIG: &str = "key_manager.config.toml";
-const NODE_CONFIG: &str = "node.config.toml";
+const KEY_MANAGER_CONFIG: &str = "key_manager.config.yaml";
```
davidiw, line 12: we'll need to update the docker containers to load the right file.

rexhoffman:
Recommend a codesearch in the libra/novifinancial org for the old file names -- should be easy to eyeball from there as a good last minute check.

gregnazario:
> Recommend a codesearch in the libra/novifinancial orgs for the old file names -- should be easy to eyeball from there as a good last minute check.

I went ahead and did the searches.

Looks like the website and one tool use these files.  I could update them as well.

PR #4725, opened 2020-06-24
[cluster-test] twin node test

snakech:
## Motivation

We want to run twin node for a validator and confirm that this does not affect liveness/TPS.

(Write your answer here.)

## Test Plan

## Related PRs

### DIFF
--- a/testsuite/cluster-test/src/cluster_builder.rs
+++ b/testsuite/cluster-test/src/cluster_builder.rs
@@ -7,7 +7,8 @@ use crate::{
     cluster_swarm::{cluster_swarm_kube::ClusterSwarmKube, ClusterSwarm},
     instance::{
         ApplicationConfig::{Fullnode, Validator, Vault, LSR},
-        FullnodeConfig, Instance, InstanceConfig, LSRConfig, ValidatorConfig, VaultConfig,
+        FullnodeConfig, Instance, InstanceConfig, LSRConfig, ValidatorConfig, ValidatorGroup,
+        VaultConfig,
     },
 };
 use anyhow::{format_err, Result};
@@ -137,7 +138,7 @@ impl ClusterBuilder {
                         let vault_config = VaultConfig {};
                         self.cluster_swarm.spawn_new_instance(
                             InstanceConfig {
-                                validator_group: i,
+                                validator_group: ValidatorGroup::new_for_index(i),
                                 application_config: Vault(vault_config),
                             },
                             delete_data,
@@ -155,7 +156,7 @@ impl ClusterBuilder {
                     };
                     self.cluster_swarm.spawn_new_instance(
                         InstanceConfig {
-                            validator_group: i,
+                            validator_group: ValidatorGroup::new_for_index(i),
                             application_config: LSR(lsr_config),
                         },
                         delete_data,
@@ -174,7 +175,7 @@ impl ClusterBuilder {
             };
             self.cluster_swarm.spawn_new_instance(
                 InstanceConfig {
-                    validator_group: i,
+                    validator_group: ValidatorGroup::new_for_index(i),
                     application_config: Validator(validator_config),
                 },
                 delete_data,
@@ -205,7 +206,7 @@ impl ClusterBuilder {
                 };
                 self.cluster_swarm.spawn_new_instance(
                     InstanceConfig {
-                        validator_group: validator_index,
+                        validator_group: ValidatorGroup::new_for_index(validator_index),
                         application_config: Fullnode(fullnode_config),
                     },
                     delete_data,
--- a/testsuite/cluster-test/src/cluster_swarm/cluster_swarm_kube.rs
+++ b/testsuite/cluster-test/src/cluster_swarm/cluster_swarm_kube.rs
@@ -142,6 +142,7 @@ impl ClusterSwarmKube {
     fn validator_spec(
         &self,
         index: u32,
+        pod_name: &str,
         num_validators: u32,
         num_fullnodes: u32,
         enable_lsr: bool,
@@ -159,6 +160,7 @@ impl ClusterSwarmKube {
         let pod_yaml = format!(
             include_str!("validator_spec_template.yaml"),
             index = index,
+            pod_name = pod_name,
             num_validators = num_validators,
             num_fullnodes = num_fullnodes,
             enable_lsr = enable_lsr,
@@ -456,7 +458,8 @@ impl ClusterSwarmKube {
         let (p, s): (Pod, Service) = match &instance_config.application_config {
             Validator(validator_config) => (
                 self.validator_spec(
-                    instance_config.validator_group,
+                    instance_config.validator_group.index,
+                    pod_name.as_str(),
                     validator_config.num_validators,
                     validator_config.num_fullnodes,
                     validator_config.enable_lsr,
@@ -465,24 +468,26 @@ impl ClusterSwarmKube {
                     &validator_config.config_overrides.iter().join(","),
                     delete_data,
                 )?,
-                self.service_spec(instance_config.pod_name()),
+                self.service_spec(pod_name.clone()),
             ),
             Fullnode(fullnode_config) => (
                 self.fullnode_spec(
                     fullnode_config.fullnode_index,
                     fullnode_config.num_fullnodes_per_validator,
-                    instance_config.validator_group,
+                    instance_config.validator_group.index_only(),
                     fullnode_config.num_validators,
                     &node.name,
                     &fullnode_config.image_tag,
                     &fullnode_config.config_overrides.iter().join(","),
                     delete_data,
                 )?,
-                self.service_spec(instance_config.pod_name()),
+                self.service_spec(pod_name.clone()),
             ),
-            Vault(_vault_config) => self.vault_spec(instance_config.validator_group, &node.name)?,
+            Vault(_vault_config) => {
+                self.vault_spec(instance_config.validator_group.index_only(), &node.name)?
+            }
             LSR(lsr_config) => self.lsr_spec(
-                instance_config.validator_group,
+                instance_config.validator_group.index_only(),
                 lsr_config.num_validators,
                 &node.name,
                 &lsr_config.image_tag,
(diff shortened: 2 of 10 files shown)


testsuite/cluster-test/src/cluster_builder.rs
```diff
@@ -137,7 +138,10 @@ impl ClusterBuilder {
                         let vault_config = VaultConfig {};
                         self.cluster_swarm.spawn_new_instance(
                             InstanceConfig {
-                                validator_group: i,
+                                validator_group: ValidatorGroup {
```
andll, line 141: Can you make `ValidatorGroup::new_for_index(u64)` and call it here and in other similar places?

testsuite/cluster-test/src/cluster_swarm/cluster_swarm_kube.rs
```diff
@@ -456,7 +458,12 @@ impl ClusterSwarmKube {
         let (p, s): (Pod, Service) = match &instance_config.application_config {
             Validator(validator_config) => (
                 self.validator_spec(
-                    instance_config.validator_group,
+                    instance_config.validator_group.index as u32,
+                    match instance_config.validator_group.twin_index {
```
andll, line 462: This should call into `InstanceConfig::pod_name` which should be updated to handle twin index.
This `format!("val-{}")` part is fragile, we want to make sure there is only single place in codebase that does this formatting(InstanceConfig::pod_name)

testsuite/cluster-test/src/cluster_swarm/cluster_swarm_kube.rs
```diff
@@ -456,7 +458,12 @@ impl ClusterSwarmKube {
...
+                        None => format!("val-{}", instance_config.validator_group.index),
+                        _ => format!("val-{}-twin", instance_config.validator_group.index),
```
andll, line 464: Also, since twin index is a number, it should be part of this too? Can be smth like `val-{}-twin-{}`

testsuite/cluster-test/src/cluster_swarm/cluster_swarm_kube.rs
```diff
@@ -471,7 +478,7 @@ impl ClusterSwarmKube {
                 self.fullnode_spec(
                     fullnode_config.fullnode_index,
                     fullnode_config.num_fullnodes_per_validator,
-                    instance_config.validator_group,
+                    instance_config.validator_group.index as u32,
```
andll, line 481: Problem with this part is that currently twin index can be set anywhere, but if node is not validator it is simply ignored. Ideally we want to fail(panic) in this case to let programmer know that they used twin index on instance type that does not support it

Can you replace this with `.index_only()->u32` method, and in this method panic if twin index is set.
Use it here and below in all places where twin index is not supported.

testsuite/cluster-test/src/experiments/mod.rs
```diff
@@ -129,6 +131,7 @@ pub fn get_experiment(name: &str, args: &[String], cluster: &Cluster) -> Box<dyn
         "reboot_random_validators",
         f::<RebootRandomValidatorsParams>(),
     );
+    known_experiments.insert("twin_validator", f::<TwinValidatorsParams>());
```
andll, line 134: Let's just use `twin` as a short name - shorter the better(experiment struct name can stay the same)

testsuite/cluster-test/src/experiments/twin_validator.rs
```diff
@@ -0,0 +1,152 @@
...
+                .await?;
+            new_instances.push(new_inst.clone());
+        }
+        let instances = self.instances.clone();
+        let emit_job_request =
+            EmitJobRequest::for_instances(instances, context.global_emit_job_request);
+        let emit_txn = context.tx_emitter.emit_txn_for(window, emit_job_request);
+        let trace_tail = &context.trace_tail;
+        let trace_delay = buffer;
+        let trace = false;
```
andll, line 78: You don't need trace in this experiment, let's remove it to not make code too complex.

testsuite/cluster-test/src/instance.rs
```diff
@@ -18,9 +18,15 @@ use std::{
 };
 use tokio::{process::Command, time};
 
+#[derive(Debug, Clone)]
+pub struct ValidatorGroup {
+    pub index: u64,
```
andll, line 23: Should we have index: u32 here, same way `validator_group` was before? Seems like it will help to avoid a lot of `as u32` in code?

testsuite/cluster-test/src/experiments/twin_validator.rs
```diff
@@ -0,0 +1,152 @@
...
+        info!(
+            "Tx status: txn {}, avg latency {}",
+            stats.committed as u64, avg_latency_client
+        );
+        let submitted_txn = stats.submitted;
+        let expired_txn = stats.expired;
+        let futures: Vec<_> = new_instances.iter().map(|ic| ic.stop()).collect();
+        try_join_all(futures).await?;
+        time::delay_for(Duration::from_secs(10)).await;
+        context
```
andll, line 111: A lot of code here related to metric reporting seem to be just a copy paste from perf experiment.
That's not great, should we add a method in `SuiteReport` like `report_txn_stats(stats: TxStats, window)`?

testsuite/cluster-test/src/instance.rs
```diff
@@ -207,7 +240,7 @@ impl Instance {
 
     pub fn validator_group(&self) -> u32 {
         let backend = self.k8s_backend();
-        backend.instance_config.validator_group
+        backend.instance_config.validator_group.index as u32
```
andll, line 243: Let's just return ValidatorGroup in this function?

testsuite/cluster-test/src/experiments/twin_validator.rs
```diff
@@ -0,0 +1,102 @@
...
+    }
+
+    fn deadline(&self) -> Duration {
+        Duration::from_secs(20 * 60)
+    }
+}
+
+impl fmt::Display for TwinValidators {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        write!(f, "twin validator")
```
andll, line 100: Can you include specific peer_name of validators that are being twined?

testsuite/cluster-test/src/experiments/twin_validator.rs
```diff
@@ -0,0 +1,102 @@
...
+impl Experiment for TwinValidators {
+    fn affected_validators(&self) -> HashSet<String> {
+        instance::instancelist_to_set(&self.twin_validators)
+    }
+
+    async fn run(&mut self, context: &mut Context<'_>) -> anyhow::Result<()> {
+        let buffer = Duration::from_secs(60);
+        let window = Duration::from_secs(240);
+        let mut new_instances = vec![];
+        for i in 0..self.twin_validators.len() {
```
andll, line 62: With current way experiment is written there is one problem - we do not know if at some point spawning twin will just completely break (e.g. for example won't launch twin).

To avoid this I propose following:
* Stop original validator before starting twin
* After starting twin, do new_inst.wait_for_json_rpc to confirm it has started
* Start original validator (without waiting for jsonrpc)

testsuite/cluster-test/src/experiments/twin_validator.rs
```diff
@@ -0,0 +1,102 @@
...
+    }
+}
+
+#[async_trait]
+impl Experiment for TwinValidators {
+    fn affected_validators(&self) -> HashSet<String> {
+        instance::instancelist_to_set(&self.twin_validators)
+    }
+
+    async fn run(&mut self, context: &mut Context<'_>) -> anyhow::Result<()> {
```
andll, line 58: Can you add `info!` with some major milestones of experiment?
Like 
- Stopping node X
- Starting node X
- Starting txn generation

testsuite/cluster-test/src/experiments/twin_validator.rs
snakech: good to know this way to check ```do new_inst.wait_for_json_rpc to confirm it has started```

testsuite/cluster-test/src/cluster_swarm/cluster_swarm_kube.rs
```diff
@@ -501,7 +506,7 @@ impl ClusterSwarmKube {
                         .ok_or_else(|| { format_err!("name not found for pod {}", pod_name) })?
                 );
             }
-            Err(e) => bail!("Failed to create pod {} : {}", pod_name, e),
+            Err(e) => bail!("Failed to create pod {} : {}", pod_name.clone(), e),
```
andll, line 509: nit: I don't think you need clone here

testsuite/cluster-test/src/experiments/performance_benchmark.rs
```diff
@@ -96,7 +96,7 @@ impl ExperimentParam for PerformanceBenchmarkParams {
             .filter_map(|val| {
                 all_fullnode_instances
                     .iter()
-                    .find(|x| val.validator_group() == x.validator_group())
+                    .find(|x| val.validator_group().index == x.validator_group().index)
```
andll, line 99: Should we just compare entire group, not just index?

testsuite/cluster-test/src/experiments/twin_validator.rs
```diff
@@ -0,0 +1,125 @@
...
+        let buffer = Duration::from_secs(60);
+        let window = Duration::from_secs(240);
+        let mut new_instances = vec![];
+        for i in 0..self.twin_validators.len() {
```
andll, line 66: Seems like this could be for each loop?
`for validator in & self.twin_validators`?

testsuite/cluster-test/src/experiments/twin_validator.rs
```diff
@@ -0,0 +1,125 @@
...
+                .cluster_swarm
+                .spawn_new_instance(new_twin_config, true)
+                .await?;
+            info!("Waiting for twin node to be up: {}", new_inst);
+            new_inst
+                .wait_json_rpc(Instant::now() + Duration::from_secs(120))
+                .await?;
+            info!("Twin node {} is up", new_inst);
+            info!("Restarting origin validator {}", self.twin_validators[i]);
+            self.twin_validators[i].start(true).await?;
```
andll, line 86: We probably don't need to wipe db, can use false

andll:
Nice

@bors-libra r+

Next thing is to think of inclusion it into nightly?

PR #4735, opened 2020-06-25
[safety-rules] A few improvements around testing and benchmarks

davidiw:
This addresses benchmark inadequacies, cleans up some ugliness in tests, and fixes a performance issue between our vault client and vault: connection churn was hurting performance substantially. This brings us to within 85% of Vault performance. Each commit has a pretty well written description that I haven't surfaced in this space.

### DIFF
--- a/consensus/safety-rules/benches/safety_rules.rs
+++ b/consensus/safety-rules/benches/safety_rules.rs
@@ -1,14 +1,13 @@
 // Copyright (c) The Libra Core Contributors
 // SPDX-License-Identifier: Apache-2.0
 
-use consensus_types::block::{block_test_utils, block_test_utils::random_payload, Block};
+use consensus_types::block::block_test_utils;
 use criterion::{black_box, criterion_group, criterion_main, Criterion};
-use libra_config::config::{OnDiskStorageConfig, SecureBackend};
 use libra_crypto::{ed25519::Ed25519PrivateKey, Uniform};
 use libra_secure_storage::{InMemoryStorage, OnDiskStorage, Storage};
 use libra_types::validator_signer::ValidatorSigner;
 use safety_rules::{
-    process_client_wrapper::ProcessClientWrapper, test_utils, PersistentSafetyStorage,
+    test_utils, PersistentSafetyStorage,
     SafetyRulesManager, TSafetyRules,
 };
 use tempfile::NamedTempFile;
@@ -16,11 +15,12 @@ use tempfile::NamedTempFile;
 /// Execute an in order series of blocks (0 <- 1 <- 2 <- 3 and commit 0 and continue to rotate
 /// left, appending new blocks on the right, committing the left most block
 fn lsr(mut safety_rules: Box<dyn TSafetyRules>, signer: ValidatorSigner, n: u64) {
-    let data = random_payload(2048);
+    let data = block_test_utils::random_payload(1);
 
-    let genesis_block = Block::make_genesis_block();
-    let genesis_qc = block_test_utils::certificate_for_genesis();
-    let mut round = genesis_block.round();
+    let (proof, genesis_qc) = test_utils::make_genesis(&signer);
+    safety_rules.initialize(&proof).unwrap();
+
+    let mut round = genesis_qc.certified_block().round();
 
     round += 1;
     let mut b0 = test_utils::make_proposal_with_qc(round, genesis_qc, &signer, None);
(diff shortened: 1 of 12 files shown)


aching:
Very nice win.  I look forward to reviewing this PR!  Can you please provide a more detailed explanation of the performance issue and how this PR addresses it?

secure/storage/vault/src/lib.rs
```diff
@@ -78,6 +78,7 @@ impl From<serde_json::Error> for Error {
 /// token, but policies can be amended afterward. So you cannot add new policies to a token, but
 /// you can increase the tokens abilities by modifying the underlying policies.
 pub struct Client {
+    agent: ureq::Agent,
```
zekun000, line 81: lol

davidiw:
> Very nice win. I look forward to reviewing this PR! Can you please provide a more detailed explanation of the performance issue and how this PR addresses it?

@aching, this commit: https://github.com/libra/libra/pull/4735/commits/300b9ba75f51bdab38cd2c2167cece1c1d10d687

specifically the issue was related to Vault not handling requests from distinct TCP connections and eventually getting overwhelmed and responding very slowly. In practice, we had a 10 second timeout that would effectively cause libra-node to crash due to an expect in libra-node. This would allow the node to recover and is why we see up and downs on performance in our deployments.

the fix is clarified in the commit itself, which is effectively to leverage connection pooling so we don't hammer Vault with new connections. inside of that, ureq doesn't seem to reuse connections if you don't read the response from a request, so we had to explicitly read the response on a put secret request. ureq uses the notion of an agent to encapsulate connection pooling.

secure/storage/vault/src/lib.rs
JoshLind: So it looks like this was the only issue? The fact that we were having to open a new connection for each request?

secure/storage/vault/src/lib.rs
```diff
@@ -366,20 +397,26 @@ impl Client {
...
         let resp = self
             .upgrade_request(request)
             .send_json(json!({ "data": { key: value } }));
         match resp.status() {
-            200 => Ok(()),
+            200 => {
+                // explicitly read contents so the socket can be re-used
+                resp.into_string()?;
+                Ok(())
+            }
```
davidiw, line 411: @aching , @JoshLind , @zekun000 this is the other case... we'd still be creating a new connection each proposal since the writes weren't flushing the buffer... this actually feels like a weird issue on the ureq side, and I'm going to file an issue.

JoshLind: Wow, yeah, this seems odd to me..

zekun000: it probably should impl Drop for this

davidiw: https://github.com/algesten/ureq/issues/94

consensus/safety-rules/benches/safety_rules.rs
```diff
@@ -114,14 +114,51 @@ fn thread(n: u64) {
...
+
 pub fn benchmark(c: &mut Criterion) {
     let count = 100;
+    let duration_secs = 5;
+    let samples = 10;
+
+    let storage = VaultStorage::new(VAULT_HOST.to_string(), VAULT_TOKEN.to_string(), None, None);
+
+    let enable_vault = if storage.available().is_err() {
+        println!("Vault is not availble");
```
aching, line 144:
```suggestion
        println!("Vault is not available at {}", VAULT_HOST.to_string());
```

secure/storage/vault/src/lib.rs
aching: Curious why this isn't necessary for explicitly reading contents for other responses - such as rotate_key, sign_ed25519, etc.?

davidiw: Well, I wouldn't say it isn't necessary, these aren't used in heavily used loops so the connection churning is extremely unlikely. Though as good hygiene, I'll add it!

consensus/safety-rules/benches/safety_rules.rs
aching: Since you are anyway making changes, maybe one more slight change.

println!("Vault is not available at {}, so Vault will not be benchmarked", VAULT_HOST.to_string());

secure/storage/vault/src/lib.rs
```diff
@@ -217,20 +238,32 @@ impl Client {
...
             404 => Err(Error::NotFound("transit/".into(), name.into())),
             _ => Err(resp.into()),
         }
     }
 
     pub fn delete_key(&self, name: &str) -> Result<(), Error> {
-        let request = ureq::post(&format!("{}/v1/transit/keys/{}/config", self.host, name));
+        let request = self
+            .agent
+            .post(&format!("{}/v1/transit/keys/{}/config", self.host, name));
```
aching, line 266: Sorry for the bad line, but does 271 also need to clear the buffer if it's okay?

secure/storage/vault/src/lib.rs
```diff
@@ -217,20 +238,32 @@ impl Client {
...
+            200 => {
+                // Explicitly clear buffer so the stream can be re-used.
+                resp.into_string()?;
+                Ok(())
+            }
+            204 => {
+                // Explicitly clear buffer so the stream can be re-used.
+                resp.into_string()?;
+                Ok(())
+            }
```
aching, line 257:
```suggestion
            200 | 204 => {
                // Explicitly clear buffer so the stream can be re-used.
                resp.into_string()?;
                Ok(())
            }
```

secure/storage/vault/src/lib.rs
davidiw: yes... so hacky... but yes

davidiw:
Thanks for the reviews!!!

@bors-libra r=aching

secure/storage/vault/src/lib.rs
zekun000: it's probably better to have a wrapper around and make it less hacky

aching:
@davidiw, @jsha's suggestion seems like a good one to clean up the code.

davidiw:
Agreed, I didn't want to do a refactor before landing the code as is since we had to make a cut that evening.

PR #4737, opened 2020-06-25
[build] Refactor build.sh to same pattern

andll:
- Adds retry to every build.sh file - previously some of them had it, and some did not
- Each build.sh now looks the same - it basically calls into libra-build.sh with appropriate docker file name
- Remove build-in-container code from cluster-test - this was not used


.circleci/config.yml
```diff
@@ -356,16 +356,6 @@ jobs:
           command: |
             RUST_BACKTRACE=1 $CI_TIMEOUT \
             ./scripts/run_quarantined.sh -c <your package here> -r 3 -f
-  validate-cluster-test-dockerfile:
```
sausagee, line 359: Thanks for reworking the Cluster Test Dockerfile workflow and retiring this check in CI.

docker/circleci/build.sh
```diff
@@ -5,9 +5,4 @@ set -e
```
sausagee, line 5: IIRC this circleci image was originally used for bundling sccache in the base container in CircleCI. This image is no longer used in the CI workflow due to speed tradeoff not working out in our favor, ie it takes more time to pull the custom image from our ECR than build without sccatch on CircleCI's default container image.  

I think we can clean up and remove this image.

docker/libra-build.sh
```diff
@@ -0,0 +1,35 @@
...
+shift 2
+
+DIR="$( cd "$( dirname "$0" )" && pwd )"
+
+PROXY=""
+if [ "$https_proxy" ]; then
+    PROXY=" --build-arg https_proxy=$https_proxy --build-arg http_proxy=$http_proxy"
+fi
+
+for _ in seq 1 30; do
```
sausagee, line 23: Is it possible to get detailed docker build failure status? If we can do that, I think we can retry based on specific build failure.  For example, it's worth retrying if the build failed due to flaky network or external infra. If the build failed due to compilation issue in Rust and 3rd party tooling, it is probably better to stop right away.

sausagee: Heads up - the build step in LBT also has retry logic but for different failure mechanism (https://github.com/libra/libra/blob/ba042341101cfec720f6ec46992bf0bb3c99bbb1/.github/workflows/land-blocking.yml#L35-L36)

You might want to take a look and adjust the retry limit as needed.

andll: I am not sure how to separate build failure between retriable and non-retriable. If we have ideas we can do it in next PR.

Maybe I should just delete retry from docker - to be completely honest I am not sure what does it help with(crates.io issues? How often are they? are they solved within retry period?)

LBT need separate retry regardless of docker builds, because most common failure there happens before even getting to docker compilation

docker/circleci/build.sh
andll: We can clean it up on separate PR?) Why mix all the changes in one?

andll:
For now I just reduced retry count, just to make sure build does not hang forever if there is, for example, compilation error

sausagee:
@bors-libra delegate+

Chatted with Andrey offline.  The circleci image will be addressed in a separate PR.

PR #4739, opened 2020-06-25
[libra framework] rename AssocRoot -> LibraRoot

sblackshear:
The Root role represents the combined authority of the Libra Association and Libra Payment Networks, so "AssocRoot" is a misnomer. Switching to the preferred term "LibraRoot".


sblackshear:
> In terms of follow ups on this PR, I'm thinking we'll want to go through and do a rename of association ~> libra as well, but that's going to be painful..thoughts?

Ugh, good point... I think I would prefer association -> root just because Libra is so overloaded.

sblackshear:
Related: I've been meaning to try [Comby](https://comby.dev/), might be interesting for us to attempt the assoc refactoring using that. It looks awesome and is created/maintained by Rijnard, a CMU PL PhD and former FB intern.

PR #4741, opened 2020-06-25
[tf] Enable Vault Transit engine and use Vault by default

mgorven:
We're using Vault in pre-mainnet now, we should be using it in testnets
as well.


mgorven:
> I wonder if we can get a couple more commands in there!

:grin: 

> does tf not play kindly with line breaks or ?

I'm just lazy :)

PR #4754, opened 2020-06-25
[storage] update get_block_timestamp with db reverse iter

msmouse:
## Motivation

improve performance.

yes

## Test Plan
unit test

### DIFF
--- a/storage/libradb/src/transaction_store/mod.rs
+++ b/storage/libradb/src/transaction_store/mod.rs
@@ -77,6 +77,9 @@ impl TransactionStore {
         // Maximum TPS from benchmark is around 1000.
         const MAX_VERSIONS_TO_SEARCH: usize = 1000 * 3;
 
+        // Linear search via `DB::rev_iter()` here, NOT expecting performance hit, due to the fact
+        // that the iterator caches data block and that there are limited number of transactions in
+        // each block.
         let mut iter = self.db.rev_iter::<TransactionSchema>(Default::default())?;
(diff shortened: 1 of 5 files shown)


metajack:
Great work. It might be worth adding a comment to get_block_metadata about the performance characteristics due to how the I/O is performed. Ie, that it doesn't cause more i/o since the blocks since the last metadata block are cached along with the metadata block.

@bors-libra delegate+

storage/libradb/src/transaction_store/mod.rs
```diff
@@ -68,6 +69,31 @@ impl TransactionStore {
...
+    pub fn get_block_metadata(&self, version: Version) -> Result<Option<(Version, BlockMetadata)>> {
+        // Maximum TPS from benchmark is around 1000.
+        const MAX_VERSIONS_TO_SEARCH: usize = 1000 * 3;
+
+        // Linear search via `DB::rev_iter()` here, NOT expecting performance hit, due to the fact
+        // that the iterator caches data block and that there are limited number of transactions in
+        // each block.
+        let mut iter = self.db.rev_iter::<TransactionSchema>(Default::default())?;
+        iter.seek(&version)?;
+        for res in iter.take(MAX_VERSIONS_TO_SEARCH) {
```
thefallentree, line 85: I was hoping you would use some magic here like 
```
.position(|x| return let Transaction:BlockMetatdata(t) = x).map(|x|..)
```

but TBH I don't know exactly how to write infunctional style

PR #4757, opened 2020-06-25
[libra framework] remove minting of LBR

sblackshear:
We currently have scripts for "minting" LBR by minting Coin1 and Coin2, then immediately creating LBR from it. In the production configuration of Libra, this should never happen:
- A single transaction should not mint both Coin1 and Coin2
- A single transaction should not both mint and create LBR
- LBR should be created from existing coins, not freshly minted ones

This PR eliminates the LBR minting functionality in the mint script as well as the LBR minting functionality in LibraAccount and LBR.


sblackshear:
This is currently RFC; I will almost certainly need to make some changes in `client` and `cluster-test` (looking into that now)

language/stdlib/modules/LibraAccount.move
```diff
@@ -289,18 +289,6 @@ module LibraAccount {
...
-    // Libra::MintCapability<Coin2>
-    public fun mint_lbr_to_address(
-        account: &signer,
-        payee: address,
-        amount: u64
-    ) acquires LibraAccount, Balance, AccountOperationsCapability {
-        // Mint and deposit the coin
-        deposit(account, payee, LBR::mint(account, amount));
-    }
-
```
tzakian, line 303: <3

PR #4762, opened 2020-06-25
[libra framework] remove Testnet module

sblackshear:
This module was a temporary measure to allow different behavior in testnet and mainnnet. We no longer need it and will implement behavior differences via sending transactions or using a different genesis rather than using this module. This will allow us to ensure that all deployments of Libra are running the same Move code paths.

 This has a few consequences:
    1. All P2P transactions on testnet are now subject to dual attestation for transfers over the threshold (1000 LBR, or 1000 * 10^6 microLibra)
    2. Because of (1), I discovered that the dual attestation threshold is not set correctly. I changed the code to use 1000 * Libra::scaling_factor<LBR>
    3. The account creation tests in account universe rely on the existence of an account type T that can create other accounts of type T. After this commit, no such account exists in our system. I disabled these tests, but hopefully we can resurrect them once we add unhosted wallets to the system.


language/stdlib/modules/LibraAccount.move
```diff
@@ -209,11 +208,7 @@ module LibraAccount {
         let above_threshold = approx_lbr_microlibra_value >= travel_rule_limit_microlibra;
         // travel rule only applies if the sender and recipient are both VASPs
         let both_vasps = VASP::is_vasp(sender) && VASP::is_vasp(payee);
-        // Don't check the travel rule if we're on testnet and sender
-        // doesn't specify a metadata signature
-        let is_testnet_transfer = Testnet::is_testnet() && Vector::is_empty(&metadata_signature);
```
sblackshear, line 214: This is the one significant effect of this change: transactions > 1k Libra on testnet will now be subject to dual attestation.

@tzakian: this also means that once we go to implementing minting via transfers from a loaded DD account + address https://github.com/libra/libra/issues/4277, the max mint size will coincide with the dual attestation threshold.

tzakian: This sounds good to me, and makes sense :)

language/stdlib/modules/LibraAccount.move
```diff
@@ -618,21 +612,20 @@ module LibraAccount {
...
 
-    // For now, only the association root can
-    // > TODO(tzakian): For now we make it so that anyone can create an unhosted
-    // account. This needs to be updated
+    // For now, only the association root can create an unhosted account, and it will choose not to
+    // on mainnet
+    // > TODO(tzakian): eventually, anyone will be able to create an unhosted wallet accunt
     public fun create_unhosted_account<Token>(
         creator_account: &signer,
+        _: &Capability<LibraRootRole>,
```
tzakian, line 624: <3

andll:
@bors-libra retry

Not sure what happened there, see some discussion in #network

PR #4770, opened 2020-06-26
[move-prover] Partial global specifications of Roles.move

DavidLDill:
There is a verified specification that, for each address with a
RoleId, the role id matches the privilege stored at that same address.
This works around a Move Prover limitation by defining a pre-condtion
that the "witness" arg to the public add_privilege_to_* functions
cannot be one of the role privileges defined in this module.  This
solution is not entirely satisfactory because it is not modular --
correctness depends on verifying that the precondition is met in every
module that uses Roles, which we probably don't want to assume in a
system with third-party Move programs.

In addition, there is a specification that, once a Role is assigned,
it cannot change.

There is also a preliminary attempt to specify that the "is_extracted"
flag for a a privilege is set if and only if there exists a capability
for that privilege.  I'm not sure it's correct, and the verifier gets
stuck on it, so 3 out of 4 properties are commented out for now.

## Motivation

(Write your answer here.)

## Test Plan

## Related PRs

### DIFF
--- a/language/stdlib/modules/Roles.move
+++ b/language/stdlib/modules/Roles.move
@@ -357,7 +357,6 @@ module Roles {
 
     /// >**Note:** Just started, only a few specs.
     ///
-    /// ## Role persistence
 
     spec module {
         pragma verify = true;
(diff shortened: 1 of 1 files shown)


language/stdlib/modules/Roles.move
```diff
@@ -381,61 +380,238 @@ module Roles {
...
-    /// **Informally:** Every address that has the treasury compliance
-    /// role ID also has a treasury compliance privilege.
-    ///
-    /// > TODO (dd): Need to add the converse, but that will have the
-    /// same problem as AssociationRootRoleMatchesRoleId due to prover
-    /// limitation.
     spec schema TreasuryComplianceRoleMatchesRoleId {
+        invariant module forall addr: address where exists<Privilege<TreasuryComplianceRole>>(addr):
+             spec_has_role_id(addr)
+                && (spec_get_role_id(addr) == SPEC_TREASURY_COMPLIANCE_ROLE_ID());
```
wrwg, line 456: nit: no need to add parenthesis around `... && .. == ..`

language/stdlib/modules/Roles.move
```diff
@@ -381,61 +380,238 @@ module Roles {
...
+
         invariant module forall addr: address where spec_has_role_id(addr):
             (spec_get_role_id(addr) == SPEC_TREASURY_COMPLIANCE_ROLE_ID())
-             ==> exists<Privilege<TreasuryComplianceRole>>(addr);
+                ==> exists<Privilege<TreasuryComplianceRole>>(addr);
```
wrwg, line 460: Same here. Redundant parentheses are considered a style violation in some disciplines, because they are confusing and lead to that other people "learn" from this.

language/stdlib/modules/Roles.move
```diff
@@ -381,61 +380,238 @@ module Roles {
...
+/// >TODO: This has not been proved yet and may be wrong, or need further
+/// strengthening.  Or, it may just be too hard for the prover.
+/// The capability returned from extract_privilege_to_capability is free-floating,
+/// so we have to quantify over members of the type, not over addresses or something.
+
+    spec schema IsExtractedIffCapabilityExists {
+        /// **Informally:** No capabilities for a privilege witness at a given
+        /// address exist until the privilege exists.
+        /// > NOTE: This verifies by itself when applied to everything.
+        invariant forall priv: type, addr: address where !exists<Privilege<priv>>(addr):
```
wrwg, line 568: invariant module?

language/stdlib/modules/Roles.move
```diff
@@ -381,61 +380,238 @@ module Roles {
...
+            !(exists c: Capability<priv>: c.owner_address == addr);
```
wrwg, line 569: I think this doesn't make sense to me. You can indeed construct a `c` for which this does not hold. The particular function which does the extract should have this as a post condition. Why make this even a global spec? 

Also you don't need to quantify priv, you can use it as a type parameter to the schema, because only generic functions can even talk about Privilege<priv>. 

I'm fine we submit this for now as an experiment but I think this should be done differently.

DavidLDill: 1.  !exists ... says there does not exist a capability c for that privilege and address, which is what I wanted (I think).
2. I think it needs to be an invariant, because it's going to be true for an arbitrarily long time until someone registers the privilege at the address.  It's forall time/ forall state, so we need to show that all function calls preserve it.

I *think* it's necessary to prove the other (commented out) properties here, but I was working fast and late at night.

language/stdlib/modules/Roles.move
emmazzz, line 568: It's probably more clear to state the contrapositive here:
`forall priv: type, c: Capability<priv>: exists<Privilege<priv>>(c.owner_address)`.
This way we are directly saying that capabilities must come from accounts with the corresponding privileges.

PR #4771, opened 2020-06-26
[swiss-knife] Public release of swiss-knife

ankushagarwal:
## Summary

`swiss-knife` can be used to generate and serialize (LCS) raw and signed Libra transactions for a supported set of move scripts (See the `MoveScriptParams` enum for a list of supported MoveScripts). This binary is intended for use both by humans (interactive use) and to be called by automation (programs in any language which support json).

`swiss-knife` expects json input from stdin (wherever applicable) and writes json output to stdout. The output json object will always have two fields: `error_message` and `data`. If the operation succeeds, then `data` is set, otherwise `error_message` is set. Only one of them will be set and the other one will be an empty string.

Typical invocation looks like `swiss-knife operation_name < operation_specific_input.json`. This writes the output json to stdout. Run `cargo run --package swiss-knife -- --help` to print all the supported operations.

The sample_inputs folder contains a list of sample json inputs for various operations


thefallentree:
I think it would be great if we name it as swissknife or something , obviously it can do more than just transaction generation

ankushagarwal:
I like the idea of swissknife 🙂 
Will rename it

PR #4772, opened 2020-06-26
[libra framework] Remove account creation from minting

### LINKED ISSUE #4758: Remove account creation on minting
Right now if the account being minted to doesn't exist, then the account is created at the time of mint creation. This doesn't follow mainnet semantics

tzakian:
Before, the mint script would create an account if it didn't already exist. However, this is not the semantics of mainnet. 

This PR removes this from the script, and instead moves it in to the client proxy that's used for testing. 

Closes #4758

### Testing
Updates existing tests work with this new semantics around mint. Changed the expected result of `mint_to_new_account` to be the expected failure code instead of success.


sblackshear:
This looks great to me!

The one thing I'm wondering (and this can be a topic for a future PR if it's too painful here) is if the script for creating a parent VASP should continue to use dummy values for the `compliance_public_key` and `base_url`. In our current VASP integration [guide](https://docs.google.com/document/d/1od5v_gIVcvX3E-Rym6sH5Z7-mDyd47A2b_64PpR8ruo/edit#heading=h.prkabi2uo463), the "create account" instructions say to send LPN your account public key, default currency, and a VASP name (see step 4). We could amend the instructions to say "send public key + base URL" as well, but that is extra info and introduces the risk of confusing the compliance public key and account public key....

@davidiw since this also ties into "VASPs in genesis Y/N?"

execution/executor-benchmark/src/lib.rs
```diff
@@ -98,10 +101,48 @@ impl TransactionGenerator {
...
+                    self.genesis_key.public_key(),
+                    encode_create_parent_vasp_account(
+                        lbr_type_tag(),
+                        account.address,
+                        account.auth_key_prefix(),
+                        b"testnet".to_vec(),
+                        b"https://libra.org".to_vec(),
+                        // A bogus (but valid ed25519) compliance public key
+                        vec![
+                            0xb7, 0xa3, 0xc1, 0x2d, 0xc0, 0xc8, 0xc7, 0x48, 0xab, 0x07, 0x52, 0x5b,
```
runtian-zhou, line 128: How are these bytes generated? Maybe include some comments for regenerating such bytes?

tzakian: These are just random valid bytes that I found on Google. It doesn't need to be anything special, just a valid ed25519 public key is all.

tzakian:
@sblackshear I don't believe we use dummy values anymore in the parent vasp creation script? We only had the dummy values in the `create_testnet_account` (which is now gone! 🎉 ) but, I might be misunderstanding?

sblackshear:
@tzakian: I was suggesting that we should perhaps continue to use dummy values (because that is the process we explain to VASPs in our guide).

tzakian:
ahhhh I see. If that's the case, I'll add a `create_testing_account_script` back that passes the right args through.

testsuite/cli/src/client_proxy.rs
```diff
@@ -453,12 +453,43 @@ impl ClientProxy {
...
         ensure!(num_coins > 0, "Invalid number of coins to mint.");
 
+        if self.assoc_root_account.is_some() {
+            let script = transaction_builder::encode_create_testing_account_script(
+                type_tag_for_currency_code(currency_code.clone()),
+                receiver,
+                receiver_auth_key.prefix().to_vec(),
+                false, /* add all currencies */
+            );
+            // If the account is local, then create the account now.
```
sblackshear, line 463:
```suggestion
            // If the receiver account is local, create it now.
```

sblackshear: Also, might make sense to put this after line 470?

testsuite/cli/src/client_proxy.rs
```diff
@@ -453,12 +453,43 @@ impl ClientProxy {
...
+            {
+                let status = &self.accounts.get(pos).unwrap().status;
+                if &AccountStatus::Local == status {
+                    // This needs to be blocking since the mint can't happen until it completes
+                    self.association_transaction_with_local_assoc_root_account(
+                        TransactionPayload::Script(script),
+                        true,
+                    )?;
+                    self.accounts.get_mut(pos).unwrap().status = AccountStatus::Persisted;
+                }
```
sblackshear, line 477:
```suggestion
                } // else, account has already been created--do nothing
```

testsuite/cli/src/client_proxy.rs
```diff
@@ -453,12 +453,43 @@ impl ClientProxy {
...
+                receiver_auth_key.prefix().to_vec(),
+                false, /* add all currencies */
+            );
+            // If the account is local, then create the account now.
+            if let Some(pos) = self
+                .accounts
+                .iter()
+                .position(|account_data| account_data.address == receiver)
+            {
+                let status = &self.accounts.get(pos).unwrap().status;
```
sblackshear, line 469: Is it possible to do `get_mut` here and avoid the second unwrap at 476?

testsuite/cluster-test/src/tx_emitter.rs
```diff
@@ -688,7 +676,7 @@ fn gen_mint_txn_requests(
             gen_mint_txn_request(
                 sending_account,
                 &account.address,
-                account.auth_key_prefix(),
+                //account.auth_key_prefix(),
```
sblackshear, line 679: Delete?

testsuite/cli/src/client_proxy.rs
tzakian: Sadly it's not :( the Rust borrowchecker complains (rightly) in this case that `self` is mutably borrowed on line 472 to send the transaction

thefallentree:
how does this work?  Does testnet faucet still able to create an account?

tzakian:
Yep! Testnet faucet will first try to create the account, and then mint to it.

PR #4773, opened 2020-06-26
[LEC] block_tree recovery

lightmark:
## Motivation

This PR adds the feature to recovery block_tree on either side of block_store or LEC if the other crashes and restarts.

Y

## Test Plan

e2e tests added for both cases.

### DIFF
--- a/consensus/src/block_storage/block_store.rs
+++ b/consensus/src/block_storage/block_store.rs
@@ -16,8 +16,7 @@ use consensus_types::{
     timeout_certificate::TimeoutCertificate,
 };
 use debug_interface::prelude::*;
-use executor_types::StateComputeResult;
-
+use executor_types::{Error, StateComputeResult};
 use libra_crypto::HashValue;
 use libra_logger::prelude::*;
 #[cfg(any(test, feature = "fuzzing"))]
@@ -33,6 +32,10 @@ use termion::color::*;
 #[cfg(test)]
 #[path = "block_store_test.rs"]
 mod block_store_test;
+
+#[cfg(test)]
+#[path = "block_store_and_lec_recovery_test.rs"]
+mod block_store_and_lec_recovery_test;
 
 #[path = "sync_manager.rs"]
 pub mod sync_manager;
@@ -290,7 +293,27 @@ impl BlockStore {
         if let Some(existing_block) = self.get_block(block.id()) {
             return Ok(existing_block);
         }
-        let executed_block = self.execute_block(block)?;
+        ensure!(
+            self.inner.read().unwrap().root().round() < block.round(),
+            "Block with old round"
+        );
+
+        let executed_block = match self.execute_block(block.clone()) {
+            Ok(res) => Ok(res),
+            Err(Error::BlockNotFound(parent_block_id)) => {
+                // recover the block tree in executor
+                let blocks_to_reexecute = self
+                    .path_from_root(parent_block_id)
+                    .unwrap_or_else(Vec::new);
+
+                for block in blocks_to_reexecute {
+                    self.execute_block(block.block().clone())?;
+                }
+                self.execute_block(block)
+            }
+            err => err,
+        }?;
+
         // ensure local time past the block time
         let block_time = Duration::from_micros(executed_block.timestamp_usecs());
         self.time_service.wait_until(block_time);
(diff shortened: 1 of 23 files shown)


consensus/src/block_storage/block_store_and_lec_recovery_test.rs
```diff
@@ -0,0 +1,256 @@
...
+    let server_port = utils::get_available_port();
+    let server_address = SocketAddr::new(IpAddr::V4(Ipv4Addr::LOCALHOST), server_port);
+
+    // Start LEC service as a process.
+    config.execution.service =
+        ExecutionCorrectnessService::SpawnedProcess(RemoteExecutionService { server_address });
+
+    // Store the config
+    let mut config_path = TempPath::new();
+    config_path.persist();
```
zekun000, line 159: why persist?

consensus/src/block_storage/block_store.rs
```diff
@@ -290,7 +293,37 @@ impl BlockStore {
...
+                loop {
+                    if parent_block_id == root_id {
+                        blocks.reverse();
+                        break;
+                    }
+                    let parent_block = self.get_block(parent_block_id).ok_or_else(|| {
+                        format_err!("Block with missing parent {}", block.parent_id())
+                    })?;
+                    blocks.push(parent_block.block().clone());
+                    parent_block_id = parent_block.parent_id();
```
zekun000, line 316: there's a function `path_from_root`

consensus/src/block_storage/block_store_and_lec_recovery_test.rs
lightmark: Otherwise it give me some random serde error if I use the same config.

zekun000:
also `speculation_cache.add_block` seems ignoring existing blocks, it should at least check if the results are the same. imagine non-deterministic execution, the original behavior is for consensus to crash and restart then recompute. but if LEC doesn't crash and silently ignore after consensus tries to add it again, consensus would never get recomputed result and crash-looping.

consensus/src/block_storage/block_store_and_lec_recovery_test.rs
zekun000: sounds weird

consensus/src/block_storage/block_store.rs
```diff
@@ -290,7 +293,27 @@ impl BlockStore {
...
+                let blocks_to_reexecute = self
+                    .path_from_root(parent_block_id)
+                    .unwrap_or_else(Vec::new);
+
+                for block in blocks_to_reexecute {
+                    self.execute_block(block.block().clone())?;
+                }
+                self.execute_block(block)
+            }
+            Err(err) => Err(err),
```
zekun000, line 314: i think this can be simplified => 
```
  Err(...) => ...,
  r @ _ => r,
```

consensus/src/block_storage/block_store_and_lec_recovery_test.rs
```diff
@@ -0,0 +1,255 @@
...
+    // Restart block_store
+    {
+        config = NodeConfig::load(&config_path).unwrap();
+        let (coordinator_sender, _coordinator_receiver) = mpsc::unbounded();
+        let state_computer = Arc::new(ExecutionProxy::new(
+            execution_correctness_manager.client(),
+            Arc::new(StateSyncClient::new(coordinator_sender)),
+        ));
+        let (initial_data, qc) = get_initial_data_and_qc(&*db);
+        let mut inserter = TreeInserter::new_with_store(
```
zekun000, line 214: make a helper function?

execution/execution-correctness/src/remote_service.rs
```diff
@@ -64,8 +64,14 @@ impl RemoteClient {
 impl TSerializerClient for RemoteClient {
     fn request(&mut self, input: ExecutionCorrectnessInput) -> Result<Vec<u8>, Error> {
         let input_message = lcs::to_bytes(&input)?;
-        self.network_client.write(&input_message)?;
-        let result = self.network_client.read()?;
+        let result = loop {
+            self.network_client.write(&input_message)?;
+            match self.network_client.read() {
+                Ok(res) => break res,
+                Err(NetError::RemoteStreamClosed) => (),
```
zekun000, line 71: why?

state-synchronizer/src/lib.rs
```diff
@@ -13,7 +13,7 @@ pub use synchronizer::{StateSyncClient, StateSynchronizer};
 
 mod chunk_request;
 mod chunk_response;
-mod coordinator;
+pub mod coordinator;
```
zekun000, line 16: so ugly

execution/execution-correctness/src/remote_service.rs
lightmark: seamless reconnection

state-synchronizer/src/lib.rs
lightmark: I can do little unless someone refactor this part to be a trait.

consensus/src/block_storage/block_store_and_lec_recovery_test.rs
lightmark: Which line

zekun000: these 10 lines below

PR #4774, opened 2020-06-26
[transaction] Make expiration time explicit in RawTransaction and helpers

ankushagarwal:
## Summary

Variable `expiration_time` is open to interpretation. In one place(`RawTransaction struct`), it represents epoch timestamp and in another place (`fn create_unsigned_txn`), it refers to the "duration in seconds from current time"

Renaming variables in both these places to clarify its usage

### DIFF
--- a/types/src/transaction/mod.rs
+++ b/types/src/transaction/mod.rs
@@ -73,7 +73,8 @@ pub struct RawTransaction {
 
     gas_currency_code: String,
 
-    // Expiration time for this transaction.  If storage is queried and
+    // Expiration timestamp for this transaction. timestamp is represented
+    // as duration in seconds from Unix Epoch. If storage is queried and
     // the time returned is greater than or equal to this time and this
     // transaction has not been included, you can be certain that it will
     // never be included.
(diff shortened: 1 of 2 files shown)


runtian-zhou:
Do we want to rename the variables in `LibraAccount::prologue` and `LibraTransactionTimeout::is_valid_transaction_timestamp` in stdlib?

ankushagarwal:
I have never touched move code, so I am not comfortable doing that. We land this and you can follow up with the appropriate changes to move?

ankushagarwal:
@runtian-zhou Gentle ping for this PR.

runtian-zhou:
oops sorry for missing the message. Renaming parts looks good to me!

types/src/transaction/mod.rs
```diff
@@ -71,15 +71,16 @@ pub struct RawTransaction {
...
+    // as duration in seconds from Unix Epoch. If storage is queried and
     // the time returned is greater than or equal to this time and this
     // transaction has not been included, you can be certain that it will
     // never be included.
     // A transaction that doesn't expire is represented by a very large value like
     // u64::max_value().
     #[serde(serialize_with = "serialize_duration")]
     #[serde(deserialize_with = "deserialize_duration")]
-    expiration_time: Duration,
+    expiration_timestamp: Duration,
```
sausagee, line 83: Is this a breaking change (and add the breaking label)?

runtian-zhou: I don't think so. It's just a naming change that shouldn't affect lcs result.

PR #4777, opened 2020-06-26
[cluster-test] Add timeout to cti and check for InvalidImageName state

ankushagarwal:
## Summary

* If there is a bug in cluster-test binary, we might never exit the process and the cluster will get blocked
  * Future: Add custom timeout from cti (single experiment will require small timeout, suite will require larger timeout)
* If one of the pods goes into an `InvalidImageName` state, we ignore it and that cluster gets blocked because the pod is still in `pending` phase. Fix this so that we delete this pod and fail

## Testing

TODO. Will test before landing

### DIFF
--- a/scripts/cluster_test_pod_template.yaml
+++ b/scripts/cluster_test_pod_template.yaml
@@ -18,7 +18,7 @@ spec:
     image: 853397791086.dkr.ecr.us-west-2.amazonaws.com/libra_cluster_test:{cluster_test_image_tag}
     imagePullPolicy: Always
     env: [{env_variables}]
-    command: [cluster-test, --deploy={image_tag} {extra_args}]
+    command: [timeout, "{timeout_secs}", cluster-test, --deploy={image_tag} {extra_args}]
   affinity:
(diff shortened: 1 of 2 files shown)


snakech:
thx for fixing this

andll:
We can probably reduce timeout to 2 hr - i don't see use cases for now for longer runs

ankushagarwal:
> We can probably reduce timeout to 2 hr - i don't see use cases for now for longer runs

Will do. For some reason, I thought Circle runs took way longer.

andll:
Ideally would be great if can introduce way to override timeout (e.g. --timeout xxx on cti) so that for use cases that need longer run, we can set it in cmd line

ankushagarwal:
I thought of doing that later, but I can do it as part of this diff. Have default timeout of 30 mins and let cti override it to something else.

ankushagarwal:
Tested  `~/git/libra/scripts/cti -E RUST_LOG=debug --timeout-secs 100 --tag master --run bench` exits after 100 seconds

PR #4778, opened 2020-06-26
[json-rpc][easy] by default, we should only listen on localhost

mimoo:
We should not expose this interface over the internet by default!

### DIFF
--- a/config/src/config/rpc_config.rs
+++ b/config/src/config/rpc_config.rs
@@ -16,7 +16,7 @@ pub const DEFAULT_JSON_RPC_PORT: u16 = 8080;
 impl Default for RpcConfig {
     fn default() -> RpcConfig {
         RpcConfig {
-            address: format!("0.0.0.0:{}", DEFAULT_JSON_RPC_PORT)
+            address: format!("127.0.0.1:{}", DEFAULT_JSON_RPC_PORT)
                 .parse()
                 .unwrap(),
         }


mimoo:
cc @mgorven

mgorven:
> We should not expose this interface over the internet by default!

Why? In what deployment situations is this default behaviour problematic?

mimoo:
> Why? In what deployment situations is this default behaviour problematic?

if I run a full node on my server, I don't want anyone to be able to query my server to access my full node. For now we don't manage secrets in this API, but you can imagine someone using this to DDoS your services.

mgorven:
> if I run a full node on my server, I don't want anyone to be able to query my server to access my full node. For now we don't manage secrets in this API, but you can imagine someone using this to DDoS your services.

If you're running a public full node you're probably doing so exactly so that you can access the JSON-RPC interface, and you likely don't want only localhost access. Almost all pieces of server software listen on 0.0.0.0 by default, and the operator is responsible for ensuring appropriate access with firewalls etc. I don't see how this is different.

mimoo:
> If you're running a public full node you're probably doing so exactly so that you can access the JSON-RPC interface, and you likely don't want only localhost access. Almost all pieces of server software listen on 0.0.0.0 by default, and the operator is responsible for ensuring appropriate access with firewalls etc. I don't see how this is different.

I would say the complete reverse, most people running full nodes have no intention of enabling public access to their nodes. This is quite a surprising default (and a vulnerability in my book).

aching:
+1
I'd like to voice my support for a default of not running JSON-RPC by default.  Principle of least privilege, IMO, would indicate starting with as little service area as possible.

Many many other blockchains don't run the RPC server by default (e.g. https://geth.ethereum.org/docs/rpc/server and https://ma.ttias.be/enable-the-rpc-json-api-with-password-authentication-in-bitcoin-core/).

aching:
If it were up to me, I'd remove the default completely and force the administrator to specify the network endpoint (e.g. 0.0.0.0:8080) in their config file explicitly.  If it wasn't specified, don't run the JSON-RPC server.

PR #4783, opened 2020-06-26
Revert "[network] migrate to simple discovery"

andll:
Looks like validators are not connecting in cluster test (see #network), and since this is only networking PR that was merged while we had land-blocking cluster test disabled, this is top suspect

Still confirming if reverting this PR make things work again, but I think possibility is strong


davidiw:
oh gosh, we probably shouldn't have pushed the smoke test refactor PR in with land blocking off...

andll:
@bors-libra r+ p=1000
Confirmed that cluster test passes with this revert

I think my learning for future is that I will try to fix cluster test(even if it means blocking pipeline for short duration), otherwise it's just too dangerous. Good thing we did not have merge conflicts on revert (although, we are not merged yet :D )

PR #4785, opened 2020-06-26
[vault] unify usage of ureq to clean up stream on request completion

davidiw:
Following up on the previous PR to make this a bit more uniform and deeper understanding of the ureq error codes.

### DIFF
--- a/secure/storage/vault/src/lib.rs
+++ b/secure/storage/vault/src/lib.rs
@@ -53,7 +53,17 @@ impl From<std::io::Error> for Error {
 
 impl From<ureq::Response> for Error {
     fn from(resp: ureq::Response) -> Self {
-        Error::HttpError(resp.status(), resp.status_line().into())
+        if let Some(e) = resp.synthetic_error() {
+            // Local error
+            Error::InternalError(e.to_string())
+        } else {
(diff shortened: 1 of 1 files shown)


secure/storage/vault/src/lib.rs
```diff
@@ -130,7 +140,11 @@ impl Client {
                 Ok(policies.policies)
             }
             // There are no policies.
-            404 => Ok(vec![]),
+            404 => {
+                // Explicitly clear buffer so the stream can be re-used.
```
JoshLind, line 144: If it were me, I'd define a helper function (e.g., "clear_response_buffer(..)") that contains the comment and the resp.into_string()? call. That way we don't need to duplicate the comment everywhere. 

Of course, the best approach would probably be to wrap ureq somehow to do this automatically for us (that is, assuming we won't get an upstream fix).. You know me and duplication :P

davidiw: Gonna sit on it for now... I would like to see this issue fixed upstream... or to explore creating some sort of droppable wrapper.

davidiw: Thanks for the review!!

PR #4786, opened 2020-06-26
Revert the revert https://github.com/libra/libra/pull/4783

zekun000:
## Motivation

Fix the gossip/discovery issue.

(Write your answer here.)

## Test Plan

## Related PRs


zekun000:
confirmed cluster test works

testsuite/tests/libratest/smoke_test.rs
```diff
@@ -1249,12 +1220,12 @@ fn test_malformed_script() {
 #[ignore]
 fn test_key_manager_consensus_rotation() {
     // Create and launch a local validator swarm of 2 nodes.
-    let mut swarm = TestEnvironment::new(2);
-    swarm.launch_swarm(RoleType::Validator);
+    let mut env = TestEnvironment::new(2);
+    env.validator_swarm.launch();
```
gregnazario, line 1224: FYI: I know this test says ignore, but it now will completely not work :) @JoshLind Either you or me will have to come back and visit this test later.

testsuite/tests/libratest/smoke_test.rs
```diff
@@ -146,15 +121,20 @@ impl TestEnvironment {
         .unwrap()
     }
 
-    fn get_validator_ac_client(
+    fn get_client(
```
gregnazario, line 124: As someone who just learned about the smoke tests, I appreciate getting rid of a lot of the extra words!

testsuite/libra-swarm/src/swarm.rs
```diff
@@ -363,13 +348,17 @@ impl LibraSwarm {
             .unwrap();
             self.nodes.insert(node_id, node);
         }
+        let expected_peer = match self.role {
```
gregnazario, line 351: *`expected_peers`

libra-node/src/main_node.rs
```diff
@@ -160,19 +160,20 @@ pub fn setup_environment(node_config: &mut NodeConfig) -> LibraHandle {
...
                 }
-
-                // Set up to listen for network configuration changes from StateSync.
-                // TODO:  move this inside network_builder.
-                if let Some(conn_mgr_reqs_tx) = network_builder.conn_mgr_reqs_tx() {
+                // HACK: gossip relies on on-chain discovery for the eligible peers update.
+                if network_config.discovery_method != DiscoveryMethod::Onchain {
+                    let conn_mgr_reqs_tx = network_builder
+                        .conn_mgr_reqs_tx()
+                        .expect("ConnectivityManager must be installed for on-chain discovery");
```
gregnazario, line 167: Maybe a more appropriate comment would be:
Gossip protocol requires p2p networks which needs ConnectivityManager.  

`expect("ConnectivityManager is necessary for non-on-chain discovery")`

config/src/config/test_data/validator_full_node.yaml
```diff
@@ -32,3 +32,7 @@ full_node_networks:
       seed_peers:
           "c227da54069989f283712e4016704660":
               - "/ip4/127.0.0.1/tcp/58259/ln-noise-ik/c998dcd54c3daf48e0ad516d94b7be0b0b7a27caa00541f2b2c14b13500df10b/ln-handshake/0"
+
+upstream:
+    networks:
+        - private: vfn
```
christinacelee, line 38: is this public network the one advertised on-chain or the one that will be used to fallover to other on-chain-advertised VFN's? how are we differentiating b/w these two public networks (because one of them has to be defined in the upstream config)

christinacelee, line 38: is the public network defined here the one used to failover to other on-chain discovered network, or the one advertised on-chain? how are we differentiating b/w the two public networks? because one of them has to be defined in the upstream config (the one where you failover to another publicly advertised VFN)

zekun000: the public network defined here `is` the one that's advertised on-chain

christinacelee: and this will be the same network used to for this VFN to fallback to other publicly advertised VFN?
then we have a situation where the same network identity is used by this VFN to communicate with downstream nodes and the upstream failover VFN, and I don't think that'll work for upstream config?

davidiw: Ahh... hmmm... this is interesting... and now I see a flaw in how we got here... we actually can end up in a strange split-case where we want 2 public networks... one for incoming clients and one for connecting to other VFNs.

I think the solution is to actually support an optional classifier for the public network that lets us distinguish these use cases... so that the on wire NetworkId is the same, but inside the code it allows for two distinct networking stacks. So effectively, we could imagine Public(bool) where bool defaults to false and true would denote this redundancy.

christinacelee: not familiar with networking code, but I assume it's possible to differentiate b/w which peers are incoming and outgoing dials? we would have to map that to true/false and propagate that to application layer (mempool/state sync), maybe in network context? so an incoming public peer would come with NetworkId::Public(true) or something like this

davidiw: that might work as well! we already get NetworkId in the mempool/state sync via the NewPeer event, right? So we'd just be adding another network. AFAIU, right?

zekun000: lol, that's the ugliness of using p2p model for upstream/downstream, time to consider client/server model?

zekun000: distinguish inbound/outbound connections has to work in validator network too i.e. a validator should treat both inbound/outbound connection as upstream while a full node should not

christinacelee: at least in our current code, validator network already has its own NetworkId::Validator that's always treated upstream, so we don't have to worry about distinguishing inbound/outbound for that (validator will just ignore that concept)

> time to consider client/server model?
if only we had more time.. T_T

for now.. I guess I can push a PR for adding this true/false variations of NetworkId::Public, set default value as placeholder and pipe that out from NewPeer event to mempool/state sync, and someone on networking side can actually make sure the correct variant of NetworkId::Public gets passed to NewPeer? we'll basically meet each other halfway (again)

or if there is another way to organize these two networking stacks, I am also open to other options, this was just what came to my head immediately

PR #4790, opened 2020-06-27
[safety-rules] config improvements

davidiw:
* ToSocketAddrs for NetworkAddress (confirmed with @phlip9 on the approach)
* Leverage NetworkAddress for SafetyRules so that it can have access to DNS.
* SafetyRules specific config

### DIFF
--- a/config/config-builder/src/main.rs
+++ b/config/config-builder/src/main.rs
@@ -4,7 +4,7 @@
 #![forbid(unsafe_code)]
 
 use config_builder::{FullNodeConfig, KeyManagerConfig, ValidatorConfig};
-use libra_config::config::{KeyManagerConfig as KMConfig, NodeConfig};
+use libra_config::config::{KeyManagerConfig as KMConfig, NodeConfig, PersistableConfig};
 use libra_network_address::NetworkAddress;
 use std::{convert::TryInto, fs, fs::File, io::Write, net::SocketAddr, path::PathBuf};
 use structopt::StructOpt;
@@ -329,9 +329,14 @@ fn build_safety_rules(args: SafetyRulesArgs) {
     }
 
     let config_builder = safety_rules_common(&args.validator_common);
-    let mut node_config = config_builder.build().expect("ConfigBuilder failed");
-    node_config.set_data_dir(args.validator_common.data_dir);
-    save_node_config(node_config, &args.validator_common.output_dir);
+    let node_config = config_builder.build().expect("ConfigBuilder failed");
+    let mut safety_rules_config = node_config.consensus.safety_rules;
+    safety_rules_config.set_data_dir(args.validator_common.data_dir);
+    let output_dir = &args.validator_common.output_dir;
+    fs::create_dir_all(output_dir).expect("Unable to create output directory");
+    safety_rules_config
+        .save_config(output_dir.join(NODE_CONFIG))
+        .expect("Unable to save config");
 }
 
 fn build_validator(args: ValidatorArgs) {
--- a/config/config-builder/src/validator_config.rs
+++ b/config/config-builder/src/validator_config.rs
@@ -187,8 +187,9 @@ impl ValidatorConfig {
     fn build_safety_rules(&self, config: &mut NodeConfig) -> Result<()> {
         let safety_rules_config = &mut config.consensus.safety_rules;
         if let Some(server_address) = self.safety_rules_addr {
-            safety_rules_config.service =
-                SafetyRulesService::Process(RemoteService { server_address })
+            safety_rules_config.service = SafetyRulesService::Process(RemoteService {
+                server_address: server_address.into(),
+            })
         }
 
         if let Some(backend) = &self.safety_rules_backend {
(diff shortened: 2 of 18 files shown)


config/config-builder/src/main.rs
```diff
@@ -329,9 +329,14 @@ fn build_safety_rules(args: SafetyRulesArgs) {
...
-    let mut node_config = config_builder.build().expect("ConfigBuilder failed");
-    node_config.set_data_dir(args.validator_common.data_dir);
-    save_node_config(node_config, &args.validator_common.output_dir);
+    let node_config = config_builder.build().expect("ConfigBuilder failed");
+    let mut safety_rules_config = node_config.consensus.safety_rules;
+    safety_rules_config.set_data_dir(args.validator_common.data_dir);
+    let output_dir = &args.validator_common.output_dir;
+    fs::create_dir_all(output_dir).expect("Unable to create output directory");
+    safety_rules_config
+        .save_config(output_dir.join(NODE_CONFIG))
```
JoshLind, line 338: We don't want a separate config for safety rules and a config name specific to safety rules? e.g., the key manager does this?

config/config-builder/src/validator_config.rs
```diff
@@ -154,22 +154,26 @@ impl ValidatorConfig {
                 .and_then(|config| config.publishing_option.clone()),
         );
 
-        let waypoint = if self.build_waypoint {
+        let (waypoint, maybe_waypoint) = if self.build_waypoint {
```
JoshLind, line 157: nit: maybe "maybe_waypoint" -> "test_waypoint"?

config/src/config/mod.rs
```diff
@@ -327,8 +326,14 @@ impl NodeConfig {
 
             let validator_network = self.validator_network.as_mut().unwrap();
             validator_network.random_with_peer_id(rng, Some(peer_id));
-            test.random_consensus_key(rng);
+            // We want to produce this key twice
+            let mut cloned_rng = rng.clone();
```
JoshLind, line 330: nit: I'd probably inline cloned_rng. I think it's somewhat obvious that we want to produce keys based on (potentially seeded) rngs

config/src/config/mod.rs
```diff
@@ -454,5 +459,9 @@ mod test {
         NodeConfig::default_for_public_full_node();
         NodeConfig::default_for_validator();
         NodeConfig::default_for_validator_full_node();
+
+        let contents = std::include_str!("test_data/safety_rules.yaml");
```
JoshLind, line 463: nit: if we give safety rules it's own config name constant, this string can be derived from that name (see my comment above)

davidiw:
> Thanks, David! Except a few nits, it mostly looks good. I guess my big question is why you decided to keep the safety rules config a part of the node config? You don't think it's worth moving it into its own separate config in this PR? Seems like if we want to do it, now might be the time?

We cannot, Consensus can start with a SafetyRules internally or externally. We could have two distinct and overlapping configs, but I am not convinced it makes a lot of sense at this point, since the configuration for a client and server are largely symmetric.

config/src/config/safety_rules_config.rs
```diff
@@ -52,5 +56,15 @@ pub enum SafetyRulesService {
...
 pub struct RemoteService {
-    pub server_address: SocketAddr,
+    pub server_address: NetworkAddress,
+}
+
+impl RemoteService {
+    pub fn server_address(&self) -> SocketAddr {
+        self.server_address
+            .to_socket_addrs()
+            .expect("server_address invalid")
```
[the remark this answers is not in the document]
davidiw: You'll also get the internal error message and stack trace. So for the first, it would be something to the extent that it wasn't a valid representation, whereas the next one would be that there no known hosts listening at that service. i.e., dns doesn't resolve ... maybe .. without knowing exactly why, I didn't want to be overly prescriptive.

config/config-builder/src/main.rs
davidiw: eventually but this would require changes across docker containers as well, can follow up in another PR / commit.

config/config-builder/src/validator_config.rs
davidiw: well in all fairness all of these are test values, maybe_waypoint implies that it might or might not be there, it is commonly used in other parts of our code base.

config/src/config/mod.rs
davidiw: That would result in some slightly weird code unless we wanted to change the structure of the following code. I, personally, found this more readable and explicit. But perhaps you're suggesting something else?

davidiw: I mean the ideal state is we would have clone available -- but we do not lol

config/src/config/mod.rs
davidiw: well, those config constants are only defined in config-builder, no? not that it doesn't make sense to probably move everything to libra-global-constants.

davidiw:
> Thanks, David! I guess the biggest question in my mind was why we couldn't separate safety rules out into its own config. Even in the "integrated" deployment model, where safety rules is started and run by consensus internally, one could still make the code fetch both the NodeConfig and the SafetyRulesConfig and pass those around, no? Either way, I don't want to be the blocker on this!

I see, I guess that might make sense longer term, maybe, but the routing of the data is extra work for ... possibly limited gain.

davidiw:
Thanks for the review!!!

@bors-libra r=joshlind

PR #4791, opened 2020-06-27
[libra framework] Make minting on testnet from a DD

tzakian:
Minting is now only done to a DD in the module code.  Minting on testnet is now just a P2P transaction with the sender being a DD account with a very large balance. 

This is stacked on #4772 so only look at the top commit for this PR.


language/tools/genesis-viewer/src/main.rs
```diff
@@ -208,3 +214,34 @@ fn print_resources(ws: &WriteSet) {
         println!("RawData: {:?}", v);
     }
 }
+
+fn print_account_states(ws: &WriteSet) {
+    let mut accounts: BTreeMap<AccountAddress, Vec<(AccessPath, Vec<u8>)>> = BTreeMap::new();
+    for (k, v) in ws {
+        match v {
+            WriteOp::Deletion => println!("found WriteOp::Deletion in WriteSet"),
```
sblackshear, line 222: Maybe even an assert for this one (if we don't check it elsewhere...)

language/tools/vm-genesis/src/lib.rs
```diff
@@ -93,6 +94,9 @@ pub fn encode_genesis_change_set(
     create_and_initialize_validators(&mut genesis_context, &validators);
     reconfigure(&mut genesis_context);
 
+    // XXX/TODO: for testnet only
```
sblackshear, line 97: My main question about this PR was how we are going to implement this. Let's sit down with @davidiw next week 
 and talk about how to manage the common and distinct genesis blobs for all the deployments?

language/tools/vm-genesis/src/lib.rs
```diff
@@ -239,7 +303,8 @@ fn verify_genesis_write_set(events: &[ContractEvent]) {
     // (1) The genesis tx should emit 1 event: a NewEpochEvent.
     assert_eq!(
         events.len(),
-        1,
+        //1,
```
sblackshear, line 306: Delete?

language/stdlib/transaction_scripts/testnet_mint.move
```diff
@@ -0,0 +1,13 @@
+script {
+use 0x1::LibraAccount;
+use 0x1::Signer;
+
+/// Send `amount` coins to `payee`.
+fun testnet_mint<Token>(payer: &signer, payee: address, amount: u64) {
```
sblackshear, line 6: Should this enforce a max on `amount`?  I think for now, any `amount` smaller than the DD's balance will succeed.

Things will get somewhat better after we address https://github.com/libra/libra/issues/4277: any `amount` over the  TR threshold will fail. But even then we might want to fail with a nice error code here + say what the error code means in the script documentation.

tzakian: I'll add some enforcement in this script, and then once we add in TR for DDs we can remove :)

language/tools/vm-genesis/src/lib.rs
tzakian: It's a good question :) The genesis additions are separable, so I'll land this now to avoid rebase hell, and then once we have a good separation strategy I'll change it.

PR #4792, opened 2020-06-27
[network] Add a ConnectivityBuilder to encapsulte create/build/start

agc-quadra:
Blocked on moving the configuration change listener.

## Motivation

Make a modular ConnectivityManagerBuilder to encapsulate the create/build/start phases associated with instantiating ConnectivityManager

Incremental step towards modularizing network-builder according to #4626

yes

## Test Plan

dont break existing tests

## Related PRs

### DIFF
--- a/network/src/connectivity_manager/builder.rs
+++ b/network/src/connectivity_manager/builder.rs
@@ -1,0 +1,132 @@
+// Copyright (c) The Libra Core Contributors
+// SPDX-License-Identifier: Apache-2.0
+
+use crate::{
+    connectivity_manager::{ConnectivityManager, ConnectivityRequest},
+    counters,
+    peer_manager::{conn_notifs_channel, ConnectionRequestSender},
+};
+use channel::{self};
+use futures::stream::StreamExt;
+use futures_util::stream::Fuse;
+use libra_config::network_id::NetworkContext;
+use libra_crypto::x25519;
+use libra_network_address::NetworkAddress;
+use libra_types::PeerId;
+use std::{
+    collections::HashMap,
+    sync::{Arc, RwLock},
+    time::Duration,
+};
+use tokio::{
+    runtime::Handle,
+    time::{interval, Interval},
+};
+use tokio_retry::strategy::ExponentialBackoff;
+
+pub type ConnectivityManagerService = ConnectivityManager<Fuse<Interval>, ExponentialBackoff>;
+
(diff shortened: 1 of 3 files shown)


agc-quadra:
@davidiw

davidiw:
Might be worth exploring some templatization here -- if that's even possible...

@bors-libra r+

agc-quadra:
@davidiw lint failure caused by a set of imports becoming unused in conjunction with concurrent PR.

agc-quadra:
@gregnazario

network/src/connectivity_manager/builder.rs
```diff
@@ -0,0 +1,132 @@
...
+        });
+    }
+
+    pub fn start(&mut self, executor: &Handle) {
+        assert_eq!(self.state, State::BUILT);
+        self.state = State::STARTED;
+        let conn_mgr = self
+            .connectivity_manager
+            .take()
+            .expect("ServiceMust be present");
```
gregnazario, line 129: nit: `Service must be present` :)

network/src/connectivity_manager/builder.rs
```diff
@@ -0,0 +1,132 @@
...
+    requests_rx: channel::Receiver<ConnectivityRequest>,
+    connection_limit: Option<usize>,
+}
+
+#[derive(Debug, PartialEq, PartialOrd)]
+enum State {
+    CREATED,
+    BUILT,
+    STARTED,
+}
```
gregnazario, line 48: It may help to share enums, and create a trait that each of these implement at least for start/build.

agc-quadra: I thought about that.  and it is probably right, but there is a very complicated bikeshed to be had about *where* that trait lives.

let's not bite that problem off quite yet.

agc-quadra:
> Let's see if we can come back on another pass after these to share at least a `Builder` trait. The trait would be paired with an `enum` for the internal state machine, and have `build` and `start` functions. That'll cut down on duplicated code and help a little enforcing the paradigm.

I think this is a good idea in principle, but the question becomes *where* that trait should be defined.

And that is a *HARD* problem to solve.

network/src/connectivity_manager/builder.rs
gregnazario: Yeah, right now it's fine.  Let's just think about it in the future.

PR #4796, opened 2020-06-27
[tf] Increase sleep before initialising Vault

mgorven:
Sometimes Vault isn't ready and initialisation fails.


mgorven:
> I guess if the issue persists we'll want to do some sort of while loop here

Yeah, although currently if Vault is already initialised then this is expected to fail.

> I wonder if we should consider using the vault binary directly.

What do you mean?

PR #4800, opened 2020-06-28
[stdlib][move-prover] Libra spec and more.

wrwg:
This finishes most local specs in `Libra.move` and does multiple other things:

- Changes in related specs (RegisteredCurrencies, LibraConfig).
- Requiring Libra::initialize and RegisteredCurrencies::initialize to be only callable if `LibraTimestamp::is_genesis` is true. I think we should apply this pattern to all other `initialize` functions, and then specify properties based on it. This required a change in `move-lang/functional_tests` to pretend we are still in genesis for passing.
- Experimenting with a different spec convention: putting function specs directly after the function definition. It appears incredible clumsy to write function specs at the end of a file. For global specs, this seems still be the best place.
- Mocking out Events. Events cause serious verification issues because of the GUID construction via LCS. In general, it seems unnecessary to include Event verification when other modules are verified, because it is a "write-only" feature for Move. The existing specs in Event have been commented out for now, and a TODO has been added to add a feature to the prover to have both the Event module verified standalone but mocked out elsewhere.
- New pragmas have been added for better dealing with aborts_if. Specifically the `pragma assume_no_abort_from_here = true` can be used to mark a function that it should be assumed to not abort when it is called from elsewhere. This is to deal with useless abort specs of underlying details (here, timestamps in LibraConfig) without the sledge hammer of making all abort specs partial. A TODO has been added to make this feature more general, e.g. to be able to distinguish between codes for programmer errors (should not be propagated for specs) and input errors (should be propagated). This finally makes aborts_if actually feasible.
- The old stdlib copy of the prover has been removed, also in docgen and stackless. This finally removes any dependencies from `sender()`. Some modules which still contain specs which have not been ported to the new lib are kept, but not connected to any tests.

## Motivation

(Write your answer here.)

## Test Plan

## Related PRs


wrwg:
@bob-wilson @tnowacki once this PR landed, we should have removed all dependencies from `sender()` in prover tests.

language/move-prover/doc/user/spec-lang.md
```diff
@@ -626,3 +626,5 @@ verification success.
 | `aborts_if_is_partial` | Allows a function to abort [under non-specified conditions](#abortsif-condition).
 | `aborts_if_is_strict`  | Disallows a function to abort even if no conditions are specified.
 | `requires_if_aborts`   | Makes a requires condition mandatory to hold even in cases where the function is specified to abort.
+| `addition_overflow_unchecked` | Makes addition of large integers (`u64` and `u128`) unchecked, avoiding the need to specify `aborts_if` for those.
+| `assume_no_abort_from_here` | Assumes that this function, if called from elsewhere, does not abort.
```
emmazzz, line 630: I want to make sure I understand what `assume_no_abort_from_here` does. Is this mainly for functions like `mint` which just calls `mint_with_capability` and does nothing else? So when this pragma is on, we don't need to copy over the `aborts_if` conditions from `mint_with_capability` to `mint`?

If that's the case, having this pragma and the `addition_overflow_unchecked` pragma can save us some spec writing time. However, I'm not sure if that's a great idea in the long run as users might just ignore overflows and pay less attention to aborts?

wrwg: This is *only* to be used for functions which cause aborts based on internal errors. Check out in this PR LibraConfig::reconfigure_ for the only place where `assume_no_abort_from_here` is used. This function is not raising an error based on input parameters but on a weird internal condition.

The `addition_overflow_unchecked` is *only* to be used where say you increment a u64 by 1 or so. This happened in `Event` before I mocked it out, so currently it is not actually used. I actually first wanted to restrict this pragma so one of the parameters to add needs to be a constant, but this wasn't that easy, so its now perhaps a bit to general.

Without this, people see the need to use the pragma `aborts_if_is_partial`. This appeared absolutely deadly to me. As soon as you add this one (which I additionally did), you have no idea whether your function aborts or not. These pragmas are there to have better control on how you want to relax abort checking. We may likely want to further refine this, but it's a start.

DavidLDill: I've been thinking about the overflow problem.  At some future point, do you think it would be worthwhile to check some basic conditions (1) counter starts at a small value, and (2) no function causes it to increase by more than 1.  If those conditions are met, then a u64 cannot realistically overflow.

I was worrying about a case where someone makes a code change that accidentally assigns a large value to a counter or accidentally allows it to increment by a large amount, suddenly making an overflow possible.  Then, the more detailed spec would catch the change.

language/move-prover/src/prelude.bpl
```diff
@@ -802,6 +802,12 @@ procedure {:inline 1} $AddU64(src1: $Value, src2: $Value) returns (dst: $Value)
     dst := $Integer(i#$Integer(src1) + i#$Integer(src2));
 }
 
+procedure {:inline 1} $AddU64_unchecked(src1: $Value, src2: $Value) returns (dst: $Value)
```
DavidLDill, line 805: I suppose we could reduce the amount of code by having a boolean flag about whether to check or not?

language/stdlib/modules/LibraConfig.move
```diff
@@ -165,6 +165,18 @@ module LibraConfig {
 
        emit_reconfiguration_event();
     }
+    spec fun reconfigure_ {
+        /// Consider only states for verification where this function does not abort
```
DavidLDill, line 169: Maybe there shouldn't be asserts for internal program errors.  I don't understand the code for reconfigure_, so that didn't clarify things for me.

language/move-prover/src/prelude.bpl
wrwg: Then we have more code which the prover need to evaluate (I don't think Boogie does constant folding).

language/stdlib/modules/LibraConfig.move
wrwg: I want to open a discussion about this after this PR. We should really partition error codes into internal errors (which don't need an aborts_if spec) and others (which are caused by user input). That is also common in languages like Java and C#, where there are checked and unchecked exceptions.

PR #4802, opened 2020-06-28
[safety-rules] Clean up as specs are written

davidiw:
This is an aesthetics only PR
* Simplify error types and formatting
* Merge together single use functions
* Update comments
* Reorder to be alphabetical where sensible

### DIFF
--- a/consensus/safety-rules/src/error.rs
+++ b/consensus/safety-rules/src/error.rs
@@ -1,77 +1,41 @@
 // Copyright (c) The Libra Core Contributors
 // SPDX-License-Identifier: Apache-2.0
 
-use consensus_types::common::Round;
 use serde::{Deserialize, Serialize};
 use thiserror::Error;
 
 #[derive(Debug, Deserialize, Error, PartialEq, Serialize)]
 /// Different reasons for proposal rejection
 pub enum Error {
-    #[error("Timeout round, {0}, is incompatible with last votedx round, {1}")]
-    BadTimeoutLastVotedRound(u64, u64),
-
-    #[error("Timeout round, {0}, is incompatible with preferred round, {1}")]
-    BadTimeoutPreferredRound(u64, u64),
-
     #[error("Provided epoch, {0}, does not match expected epoch, {1}")]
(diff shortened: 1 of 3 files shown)


consensus/safety-rules/src/safety_rules.rs
```diff
@@ -308,44 +285,38 @@ impl TSafetyRules for SafetyRules {
...
             validator_signer.author(),
             self.construct_ledger_info(proposed_block),
-            &validator_signer,
+            validator_signer,
         ))
     }
 
     fn sign_proposal(&mut self, block_data: BlockData) -> Result<Block, Error> {
         debug!("Incoming proposal to sign.");
+        self.signer()?;
```
JoshLind, line 306: nit: I'd avoid doing the check twice, e.g.,

let validator_signer = self.signer()?;
...

Ok(Block...
    block_data,
    validator_signer
)

consensus/safety-rules/src/error.rs
```diff
@@ -1,77 +1,41 @@
...
-    #[error("Internal error: {:?}", error)]
-    InternalError { error: String },
-
-    #[error("Invalid proposal: {}", {0})]
-    InvalidProposal(String),
-
-    #[error("Unable to verify that the new tree extneds the parent: {:?}", error)]
-    InvalidAccumulatorExtension { error: String },
-
+    BadEpoch(u64, u64),
```
JoshLind, line 11: nit: "Bad" is often an unhelpful adjective as it can mean many different things.. Perhaps make these error names a little more descriptive/specific? e.g., (using the description text for each error), you could refactor:

BadEpoch -> UnexpectedEpoch
BadLastVotedRound -> IncompatibleLastVotedRound
BadPreferredRound -> IncompatiblePreferredRound

consensus/safety-rules/src/safety_rules.rs
davidiw: Rust, references, and mutability :/

That was how I had it originally until the compiler informed me how great it is :P

PR #4806, opened 2020-06-28
[libra framework][move-prover] Refactor Roles.move and clients

DavidLDill:
[libra framework][move-prover] Refactor Roles.move and clients

This is a refactor of Roles.move and modules that use it to make
formal verification of access control policies less difficult, by
basing permissions on role_ids of addresses, instead of Privilege and
Capability objects.

The previous approach of publishing Privilege<Priv> objects and
extacting capabilities from them created several problems for the Move
Prover. The most fundamental of these is that the Move Prover cannot
prove a property unless it sees all of the code that could invalidate
that property. As currently designed, the available code is the module
containing the property and all modules in its dependencies.
The advantage of this approach is that properties that are so proved
will continue to hold even in an open blockchain with modules and
transactions that we do not trust.

More pragmatically, there were a number of other problems, ranging
from tricky verification problems that we haven't yet solved to
limitations of the verifier that we don't have time to fix.  The
refactor should eliminate the need to prove many properties and will
make others much simpler to prove.

Before the refactor, a Move function that is restricted to certain signers
has an argument of type &Capability<Priv>, which can only be generated
if there is a signer of the transaction with a published Privilege<Priv>.

The refactor replaces that &Capability argument with an &Signer
argument, and calls assert(has_..._privilege(account), <abort code>)
to abort the function if the signer does not have the required
permissions.  As it happens, has_..._privilege is always a property of
the role of the address as defined by the RoleId in Roles.move, so we
don't actually need to have or publish Privilege<Priv> objects.  To
achieve obvious, visible correspondence to our spreadsheet of
"governance roles and privileges", these functions have names
corresponding to privileges from the spreadsheet.  Several identical
functions have the same body, because some roles have multiple
privileges. (In the future, perhaps we can auto-generate verification
properties from something like that spreadsheet.)

This approach retains the flexibility of changing the addresses that
have a given role, because the RoleId implementation is unchanged.
The correspondence between roles and privileges can also be changed
easily by changing the appropriate has_..._privilege functions. Right
now, they all just check for a particular role ID, but there would be
no problem changing them to allow multiple alternative roles, or to
check for the presence of a published resource of some type, or lots
of other policies.
[libra framework][move-prover] Refactor Roles.move and clients

This is a refactor of Roles.move and modules that use it to make
formal verification of access control policies less difficult, by
basing permissions on role_ids of addresses, instead of Privilege and
Capability objects.

The previous approach of publishing Privilege<Priv> objects and
extracting capabilities from them created several problems for the Move
Prover. The most fundamental of these is that the Move Prover cannot
prove a property unless it sees all of the code that could invalidate
that property. As currently designed, the available code is the module
containing the property and all modules in its dependencies.
The advantage of this approach is that properties that are so proved
will continue to hold even in an open blockchain with modules and
transactions that we do not trust.

More pragmatically, there were a number of other problems, ranging
from tricky verification problems that we haven't yet solved to
limitations of the verifier that we don't have time to fix.  The
refactor should eliminate the need to prove many properties and will
make others much simpler to prove.

Before the refactor, a Move function that is restricted to certain signers
has an argument of type &Capability<Priv>, which can only be generated
if there is a signer of the transaction with a published Privilege<Priv>.

The refactor replaces that &Capability argument with an &Signer
argument, and calls assert(has_..._privilege(account), <abort code>)
to abort the function if the signer does not have the required
permissions.  As it happens, has_..._privilege is always a property of
the role of the address as defined by the RoleId in Roles.move, so we
don't actually need to have or publish Privilege<Priv> objects.  To
achieve obvious, visible correspondence to our spreadsheet of
"governance roles and privileges", these functions have names
corresponding to privileges from the spreadsheet.  Several identical
functions have the same body, because some roles have multiple
privileges. (In the future, perhaps we can auto-generate verification
properties from something like that spreadsheet.)

This approach retains the flexibility of changing the addresses that
have a given role, because the RoleId implementation is unchanged.
The correspondence between roles and privileges can also be changed
easily by changing the appropriate has_..._privilege functions. Right
now, they all just check for a particular role ID, but there would be
no problem changing them to allow multiple alternative roles, or to
check for the presence of a published resource of some type, or lots
of other policies.

## Motivation

(Write your answer here.)

## Test Plan

## Related PRs


language/move-prover/src/prelude.bpl
```diff
@@ -827,12 +827,138 @@ procedure {:inline 1} $Sub(src1: $Value, src2: $Value) returns (dst: $Value)
 // to src2 == any integer Value from 0..127.
 // Left them out for brevity
 function $power_of_2(power: $Value): int {
-    (var p := i#$Integer(power);
-     if p == 32 then 4294967296
-     else if p == 64 then 18446744073709551616
-     // Value is undefined, otherwise.
-     else -1
-     )
+   (var p := i#$Integer(power);
```
wrwg, line 830: This looks super dangerous for non-termination problems because this can lead to exponential explosion via case splitting of a very high degree. We can try this out for now but I'm pretty sure we run into issues.

language/stdlib/modules/Libra.move
```diff
@@ -572,14 +572,18 @@ module Libra {
...
+        tc_account: &signer,
         to_lbr_exchange_rate: FixedPoint32,
         is_synthetic: bool,
         scaling_factor: u64,
         fractional_part: u64,
         currency_code: vector<u8>,
     ): (MintCapability<CoinType>, BurnCapability<CoinType>)
     acquires CurrencyRegistrationCapability {
+        // Privilege check.  TODO (dd): Error code.
+        assert(has_register_new_currency_privilege(tc_account), 99);
```
wrwg, line 586: Unrelated to this PR, but we really need to give those errors like `99` names and then divide them into categories. It seems currently they are fully random. This also allows us to do a better way to deal with aborts_if specs, because errors of a certain category (internal errors) we can ignore.

@sblackshear @tzakian

language/stdlib/modules/Roles.move
```diff
@@ -295,6 +295,48 @@ module Roles {
...
+    /// Naming conventions: Many of the "has_*_privilege" functions do have the same body
+    /// because the spreadsheet grants all such privileges to addresses (usually a single
+    /// address) with that role. In effect, having the privilege is equivalent to having the
+    /// role, but the function names document the specific privilege involved.  Also, modules
+    /// that use these functions as a privilege check can hide specific roles, so that a change
+    /// in the privilege/role relationship can be implemented by changing Roles and not the
+    /// module that uses it.
+    public fun has_treasury_compliance_role(tc_account: &signer): bool acquires RoleId {
+       let tc_address = Signer::address_of(tc_account);
+            exists<RoleId>(tc_address)
```
wrwg, line 315: nit: please do not indent after let. The expression on the next line after let has the same indentation as the let. The style you are using is super uncommon.

language/stdlib/modules/Libra.move
```diff
@@ -572,14 +572,18 @@ module Libra {
...
+        // _: &Capability<RegisterNewCurrency>,
+        tc_account: &signer,
         to_lbr_exchange_rate: FixedPoint32,
         is_synthetic: bool,
         scaling_factor: u64,
         fractional_part: u64,
         currency_code: vector<u8>,
     ): (MintCapability<CoinType>, BurnCapability<CoinType>)
     acquires CurrencyRegistrationCapability {
+        // Privilege check.  TODO (dd): Error code.
```
wrwg, line 585: I think we should also introduce `assert_has_register_new_currency_privilege` and then have the error code produced for this there once and for all.

language/stdlib/modules/Roles.move
```diff
@@ -295,6 +295,48 @@ module Roles {
...
+    // DD refactor: privilege-checking functions for roles
+    /// has_*_role and has_*_privilege functions are derived exactly from the
+    /// "Libra Governance roles & permissions" spreadsheet
+    /// https://docs.google.com/spreadsheets/d/13pL_0JCdNkGTMemVYKLvGifiF-svP_fW_pT6CjxFMCA/edit#gid=0
+    /// MintCurrency, BurnCurrency, PreburnCurrency, RotateAuthenticationKey, WithdrawalCapability are
+    /// omitted because they all involve real capabilities. They were also not handled in the
+    /// Roles module before refactoring.
+    ///
+    /// Naming conventions: Many of the "has_*_privilege" functions do have the same body
+    /// because the spreadsheet grants all such privileges to addresses (usually a single
```
wrwg, line 307: This is totally ok.

language/stdlib/modules/VASP.move
```diff
@@ -290,6 +327,11 @@ module VASP {
     }
 
     spec fun publish_child_vasp_credential {
+        pragma aborts_if_is_partial = true;
+        /// >TODO (dd): I'm having a problem importing this from Roles
```
wrwg, line 331: That comment seems now obsolete

language/move-prover/src/prelude.bpl
DavidLDill: Yeah, it has nothing to do with the Role refactoring, and I'll just take it out. I had some problem I don't understand where I suddenly started getting errors from SlidingNonce, and I was trying to fix them.  I agree that it's dangerous.  It may be the only way to implement variable shift, but maybe we need to find another solution to SlidingNonce so we don't need it.

language/stdlib/modules/Libra.move
DavidLDill: Yeah, no more thinking up error codes sounds nice.

language/stdlib/modules/Libra.move
DavidLDill: Tim's working on that, I believe.

language/stdlib/modules/Roles.move
DavidLDill: Mistake. Fixed.

language/stdlib/modules/Libra.move
DavidLDill: Ok.  I think I'll forge ahead with the current approach, but plan to change it after I've finished refactoring.  This sounds like a change that would be easy to do with sed.

language/stdlib/modules/LBR.move
```diff
@@ -67,15 +70,20 @@ module LBR {
     /// restrictions are enforced in the `Libra::register_currency` function, but also enforced here.
     public fun initialize(
         association: &signer,
-        register_currency_capability: &Capability<RegisterNewCurrency>,
+        // DD: refactor
```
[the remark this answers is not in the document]
DavidLDill: That's a good point.  I think verifier can also guarantee that we haven't missed any calls, and our verification strategy is based on the idea that people will add or change functions.  But defense in depth might be appropriate.

language/stdlib/modules/VASP.move
```diff
@@ -93,8 +97,16 @@ module VASP {
...
         parent: &signer,
         child: &signer,
-        _: &Capability<ParentVASPRole>,
+        // DD: refactor
+        // _: &Capability<ParentVASPRole>,
     ) acquires ParentVASP {
+        // DD: The spreadsheet does not have a "privilege" for creating
+        // child VASPs. All logic in the code is based on the parent VASP role.
+        // DD: Since it checks for a ParentVASP property, anyway, checking
+        // for role might be a bit redundant (need invariant that only
```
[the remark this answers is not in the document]
DavidLDill: Maybe.  I wasn't 100% sure of what everything meant, so I left the old test.  Optimization is not as important as security at this point.

language/stdlib/modules/LibraConfig.move
```diff
@@ -25,16 +29,22 @@ module LibraConfig {
...
-    /// Will fail if the account is not association root
-    public fun grant_privileges(account: &signer) {
-        Roles::add_privilege_to_account_association_root_role(account, CreateOnChainConfig{});
+    // Will fail if the account is not association root
+    // DD: refactor
+    // public fun grant_privileges(account: &signer) {
+    //     Roles::add_privilege_to_account_association_root_role(account, CreateOnChainConfig{});
+    // }
+
+    public fun assert_has_on_chain_config_privilege(lr_account: &signer) {
```
tzakian, line 38: Not sure if this should be in the module or the `Roles` module. I would have preference for `Roles` since it's in the spreadsheet, but I'm also fine with keeping it here if you would prefer.

language/stdlib/modules/Roles.move
```diff
@@ -285,6 +288,76 @@ module Roles {
...
+    /// address) with that role. In effect, having the privilege is equivalent to having the
+    /// role, but the function names document the specific privilege involved.  Also, modules
+    /// that use these functions as a privilege check can hide specific roles, so that a change
+    /// in the privilege/role relationship can be implemented by changing Roles and not the
+    /// module that uses it.
+
+    public fun assert_has_role(account: &signer, role_id: u64) acquires RoleId {
```
sblackshear, line 307: As mentioned/explained earlier, please do not add `assert_` functions like this--the `assert` needs to be at the call site so we can see what was happening when the privilege check fails.

wrwg: I was the one suggesting to introduce this, and I did already something similar in submitted code (`LibraTimestamp::assert_is_genesis`). This is ugly. The reason why we are doing this is to ensure error codes are consistently used. The error code story is kind of a big mess right now. I don't quite understand why Move VM wasn't designed from the beginning to give you a stack trace on abort, it is a very basic feature.

If we roll this back, we should also do other places, and think about at least making error codes constants throughout the place (module `ErrorCodes`).

sblackshear: Totally agreed on rolling it back in other places + making all error codes constants. This is planned (https://github.com/orgs/libra/projects/7), but things are indeed a mess right now (we were only recently unblocked by source lang constants). Just don't want to make the problem worse before we fix it :). 

>  I don't quite understand why Move VM wasn't designed from the beginning to give you a stack trace on abort, it is a very basic feature.

This is something that we have (though it's not enabled automatically), but the problem is that we can't use this on-chain. The `vm_status` part of a transaction output needs to be constant size and fairly small. Today, you only get `SUCCESS` or `ABORTED` (not even the abort error code!). @tnowacki and @tzakian are working on adding at least the module name and abort error code to mitigate this.

wrwg: If you have this but can't use on-chain I'd say this is a PITA for DevOps. If you have an unexpected failure in production, you want to be able to identify its source and mitigate as fast as possible. This should strictly override concerns about resource usage (and has in production environments I've seen before).

@dariorussi

sblackshear: This is a nuanced topic that I probably shouldn't have brought up on in a GH comment thread, but let me explain a few more of the nuances since we're here:

- There are two ways to get a stack trace and other metadata from an aborted execution: (1) in the VM logs (2) the `TransactionInfo` stored directly on the blockchain. We have very good support for (1); the logs are thorough and there is tooling for accessing, scraping, and indexing these logs. This is the way DevOps folks running nodes should/will get at VM information (and totally agreed that stack traces and other data is essential there.
- (2) is how clients will get data about aborted execution (since they don't have access to the logs of the validator node). There are definite tradeoffs on having more vs less information here. Resource usage is one concern, but not the only or primary one. Another concern is maintaining a small TCB and spec footprint. Note that the `vm_status` part of an executed transaction is stored on the blockchain, so changes there affect the Merkle root hash. If the status is more complex (e.g., a stack trace), the code for producing that stack trace is in the TCB and the format of that stack trace is part of the Libra spec. 
- The reasoning behind including only a status code and module name name in (2) is that it strikes a reasonable balance between client friendliness and TCB/resource usage/other concerns. A client gets decent info by reading the blockchain, and it can (if necessary) get the full Libra logs by syncing to the latest blockchain state and replaying the transaction execution locally.

language/stdlib/modules/Libra.move
DavidLDill: I changed it so that each function has assert(has_..._privilege, <code>) in each function call, instead of the previous assert_has_...._privilege with one common error code.  Error codes are unique but not sensibly assigned (Tim said he'll reassign them).

language/stdlib/modules/LBR.move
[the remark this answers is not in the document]
DavidLDill: I have in mind a systematic approach to verifying that we've checked all the right privileges.  Maybe I'll  go back and check these higher-level calls later with asserts after more consideration.

language/stdlib/modules/LibraConfig.move
DavidLDill: I put these kinds of functions in the modules that used to define the privileges.  Theory:  If someone wants to change it, that person is more likely to be making other changes in the defining module, not Roles.  Also, a new modules can define it's function there instead of changing Roles, also.

But it's easy to move these functions, and putting them all in Roles, which would have the advantage of making them easier to compare with the spreadsheeet.

language/stdlib/modules/Roles.move
DavidLDill: I'm using individual asserts now, based on an earlier comment.  It's not too hard to change back if we want to.

language/stdlib/modules/LibraConfig.move
DavidLDill: Actually, everything is in Roles. That was a *duplicate* function, in LibraConfig by mistake (fixed).

language/stdlib/modules/Roles.move
```diff
@@ -282,65 +174,75 @@ module Roles {
...
         move_to(new_account, RoleId { role_id: UNHOSTED_ROLE_ID() });
-        move_to(new_account, Privilege<UnhostedRole>{ witness: UnhostedRole{}, is_extracted: false })
     }
 
-    ///////////////////////////////////////////////////////////////////////////
-    // Capability Extraction from Privileges, and Restoration to Privileges
-    ///////////////////////////////////////////////////////////////////////////
+    ///  ## privilege-checking functions for roles ##
+    ///
+    ///>TODO (dd) remove spreadsheet link unless we're going to publish it.
```
sblackshear, line 181: Do this before landing perhaps? I do think we'll eventually have a CSV of the spreadsheet in the public Roles LIP

language/stdlib/modules/SlidingNonce.move
```diff
@@ -1,9 +1,9 @@
 address 0x1 {
 module SlidingNonce {
     use 0x1::Signer;
-    use 0x1::Roles::{Self, Capability};
+    use 0x1::Roles::{has_libra_root_role};
 
-    resource struct CreateSlidingNonce {}
+    // resource struct CreateSlidingNonce {}
```
sblackshear, line 6: Delete?

language/stdlib/modules/SlidingNonce.move
```diff
@@ -18,16 +18,20 @@ module SlidingNonce {
 
     /// Grants the `CreateSlidingNonce` privilege to the calling `account`.
     /// Aborts if the calling account does not have the association root role.
-    public fun grant_privileges(account: &signer) {
-        Roles::add_privilege_to_account_association_root_role(account, CreateSlidingNonce{});
-    }
+    // public fun grant_privileges(account: &signer) {
```
sblackshear, line 21: Delete?

language/stdlib/modules/SlidingNonce.move
```diff
@@ -18,16 +18,20 @@ module SlidingNonce {
...
 
     /// Calls try_record_nonce and aborts transaction if returned code is non-0
     public fun record_nonce_or_abort(account: &signer, seq_nonce: u64) acquires SlidingNonce {
         let code = try_record_nonce(account, seq_nonce);
         assert(code == 0, code);
     }
 
+
+    public fun has_create_sliding_nonce_privilege(lr_account: &signer): bool {
+        has_libra_root_role(lr_account)
```
sblackshear, line 32: Nit: The style we use use elsewhere in the stdlib is important just the module + using the module prefix at every call site (i.e., `Roles::has_libra_root_role`). That makes it easier to see where functions are defined.

language/stdlib/modules/ValidatorConfig.move
```diff
@@ -43,12 +41,17 @@ module ValidatorConfig {
...
+        lr_account: &signer,
+        ) {
+        // TODO: abort code
+        assert(has_libra_root_role(lr_account), 919425);
         move_to(account, ValidatorConfig {
             config: Option::none(),
             operator_account: Option::none(),
         });
-        Roles::add_privilege_to_account_validator_role(account, UpdateValidatorConfig{})
+        // Roles::add_privilege_to_account_validator_role(account, UpdateValidatorConfig{})
```
sblackshear, line 54: Delete?

language/stdlib/transaction_scripts/create_validator_account.move
```diff
@@ -1,16 +1,16 @@
...
-    use 0x1::Roles::{Self, LibraRootRole};
 
     /// Create a validator account at `new_validator_address` with `auth_key_prefix`.
-    fun create_validator_account(creator: &signer, new_account_address: address, auth_key_prefix: vector<u8>) {
-        let assoc_root_role = Roles::extract_privilege_to_capability<LibraRootRole>(creator);
+    fun create_validator_account(
+        creator: &signer,
+        new_account_address: address,
+        auth_key_prefix: vector<u8>,
+        ) {
```
sblackshear, line 9: Nit: indentation

language/stdlib/transaction_scripts/freeze_account.move
```diff
@@ -1,14 +1,11 @@
 script {
-use 0x1::LibraAccount::{Self, AccountFreezing};
+use 0x1::LibraAccount::{Self};
```
sblackshear, line 2:
```suggestion
use 0x1::LibraAccount;
```

language/stdlib/transaction_scripts/update_minting_ability.move
```diff
@@ -1,11 +1,11 @@
...
 
 /// Allows--true--or disallows--false--minting of `currency` based upon `allow_minting`.
-fun update_minting_ability<Currency>(account: &signer, allow_minting: bool) {
-    let tc_capability = Roles::extract_privilege_to_capability<TreasuryComplianceRole>(account);
-    Libra::update_minting_ability<Currency>(&tc_capability, allow_minting);
-    Roles::restore_capability_to_privilege(account, tc_capability);
+fun update_minting_ability<Currency>(
+    tc_account: &signer,
+    allow_minting: bool
+    ) {
```
sblackshear, line 8: Nit: indentation

language/stdlib/transaction_scripts/update_travel_rule_limit.move
```diff
@@ -1,14 +1,10 @@
 script {
-    use 0x1::Signer;
     use 0x1::SlidingNonce;
-    use 0x1::DualAttestationLimit::{Self, UpdateDualAttestationThreshold};
-    use 0x1::Roles;
+    use 0x1::DualAttestationLimit::{Self};
```
sblackshear, line 3:
```suggestion
    use 0x1::DualAttestationLimit;
```

language/stdlib/modules/DualAttestationLimit.move
```diff
@@ -3,10 +3,10 @@ address 0x1 {
 module DualAttestationLimit {
     use 0x1::LBR::LBR;
     use 0x1::Libra;
-    use 0x1::LibraConfig::{Self, CreateOnChainConfig};
+    use 0x1::LibraConfig::{Self};
```
sblackshear, line 6:
```suggestion
    use 0x1::LibraConfig;
```

language/stdlib/modules/Genesis.move
```diff
@@ -10,24 +10,23 @@ module Genesis {
     use 0x1::DualAttestationLimit;
     use 0x1::Event;
     use 0x1::LBR;
-    use 0x1::Libra::{Self, RegisterNewCurrency};
+    use 0x1::Libra::{Self};
```
sblackshear, line 13:
```suggestion
    use 0x1::Libra;
```

language/stdlib/modules/Genesis.move
```diff
@@ -10,24 +10,23 @@ module Genesis {
...
     use 0x1::LibraAccount;
     use 0x1::LibraBlock;
-    use 0x1::LibraConfig::{Self, CreateOnChainConfig};
+    use 0x1::LibraConfig::{Self};
```
sblackshear, line 16:
```suggestion
    use 0x1::LibraConfig;
```

language/stdlib/modules/Genesis.move
```diff
@@ -10,24 +10,23 @@ module Genesis {
...
     use 0x1::LibraSystem;
     use 0x1::LibraTimestamp;
     use 0x1::LibraTransactionTimeout;
     use 0x1::LibraVersion;
     use 0x1::LibraWriteSetManager;
     use 0x1::Signer;
     use 0x1::TransactionFee;
-    use 0x1::Roles::{Self, LibraRootRole, TreasuryComplianceRole};
-    use 0x1::SlidingNonce::{Self, CreateSlidingNonce};
+    use 0x1::Roles::{Self};
```
sblackshear, line 24:
```suggestion
    use 0x1::Roles;
```

language/stdlib/modules/LBR.move
```diff
@@ -16,8 +16,9 @@ module LBR {
     use 0x1::Coin2::Coin2;
     use 0x1::CoreAddresses;
     use 0x1::FixedPoint32::{Self, FixedPoint32};
-    use 0x1::Libra::{Self, Libra,  RegisterNewCurrency};
-    use 0x1::Roles::{Capability, TreasuryComplianceRole};
+    use 0x1::Libra::{Self, Libra,
+    // RegisterNewCurrency
```
sblackshear, line 20: Delete?

language/stdlib/modules/Libra.move
```diff
@@ -189,38 +186,38 @@ module Libra {
     /// the calling account as long as it has the correct role (TC).
     /// Aborts if `account` does not have a `RoleId` that corresponds with
     /// the treacury compliance role.
-    public fun grant_privileges(account: &signer) {
-        Roles::add_privilege_to_account_treasury_compliance_role(account, RegisterNewCurrency{});
-    }
+    // public fun grant_privileges(account: &signer) {
```
sblackshear, line 189: Delete?

language/stdlib/modules/Libra.move
```diff
@@ -663,14 +663,17 @@ module Libra {
     /// `MintCapability<CoinType>` and `BurnCapability<CoinType>` resources.
     public fun register_currency<CoinType>(
         account: &signer,
-        _: &Capability<RegisterNewCurrency>,
+        // DD refactor
```
sblackshear, line 666: Delete?

language/stdlib/modules/LibraAccount.move
```diff
@@ -23,11 +23,10 @@ module LibraAccount {
     use 0x1::Libra::{Self, Libra};
     use 0x1::Option::{Self, Option};
     use 0x1::DualAttestationLimit;
-    use 0x1::Roles::{Self, Capability, LibraRootRole, ParentVASPRole, TreasuryComplianceRole};
-    use 0x1::SlidingNonce::CreateSlidingNonce;
+    use 0x1::Roles::{Self, has_libra_root_role, has_treasury_compliance_role};
 
-    resource struct AccountFreezing {}
-    resource struct AccountUnfreezing {}
+    // resource struct AccountFreezing {}
```
sblackshear, line 28: Delete?

language/stdlib/modules/LibraAccount.move
```diff
@@ -812,26 +810,28 @@ module LibraAccount {
...
-        _: &Capability<LibraRootRole>,
         new_account_address: address,
         auth_key_prefix: vector<u8>,
     ) {
+        // TODO: abort code
+        assert(has_libra_root_role(creator_account), 919413);
         let new_account = create_signer(new_account_address);
         Event::publish_generator(&new_account);
-        Roles::new_validator_operator_role(creator_account, &new_account);
+//        Roles::new_validator_operator_role(creator_account, &new_account);
```
sblackshear, line 834: Delete?

language/stdlib/modules/LibraConfig.move
```diff
@@ -5,9 +5,10 @@ module LibraConfig {
     use 0x1::LibraTimestamp;
     use 0x1::Signer;
     use 0x1::Offer;
-    use 0x1::Roles::{Self, Capability, LibraRootRole};
+    // TODO: Can't use aliasing because spec functions won't import
+    use 0x1::Roles;
 
-    resource struct CreateOnChainConfig {}
+    // resource struct CreateOnChainConfig {}
```
sblackshear, line 11: Delete?

language/stdlib/modules/LibraConfig.move
```diff
@@ -25,16 +26,14 @@ module LibraConfig {
     // Accounts with this privilege can modify config of type TypeName under default_address
     resource struct ModifyConfigCapability<TypeName> {}
 
-    /// Will fail if the account is not association root
-    public fun grant_privileges(account: &signer) {
-        Roles::add_privilege_to_account_association_root_role(account, CreateOnChainConfig{});
-    }
+    // Will fail if the account is not association root
```
sblackshear, line 29: Delete?

language/stdlib/modules/LibraVMConfig.move
```diff
@@ -1,8 +1,7 @@
 address 0x1 {
 
 module LibraVMConfig {
-    use 0x1::LibraConfig::{Self, CreateOnChainConfig};
-    use 0x1::Roles::Capability;
+    use 0x1::LibraConfig::{Self};
```
sblackshear, line 4:
```suggestion
    use 0x1::LibraConfig;
```

language/stdlib/modules/RegisteredCurrencies.move
```diff
@@ -3,11 +3,10 @@ address 0x1 {
 /// Module managing the registered currencies in the Libra framework.
 module RegisteredCurrencies {
     use 0x1::CoreAddresses;
-    use 0x1::LibraConfig::{Self, CreateOnChainConfig};
+    use 0x1::LibraConfig::{Self};
```
sblackshear, line 6:
```suggestion
    use 0x1::LibraConfig;
```

language/move-prover/docgen/tests/sources/LibraTest.move
```diff
@@ -670,14 +663,17 @@ module LibraTest {
     /// `MintCapability<CoinType>` and `BurnCapability<CoinType>` resources.
     public fun register_currency<CoinType>(
         account: &signer,
-        _: &Capability<RegisterNewCurrency>,
+        // DD refactor
```
sblackshear, line 666: Delete?

PR #4812, opened 2020-06-29
[network] test rotating networking key

gregnazario:
The network keys must be rotated manually at this point.  This adds a
smoke test that ensures the process is well formed and that when a key
is rotated, the node can reconnect to the rest of the nodes.

Ideally, this would use the JSON RPC call [`get_account_transaction`](https://github.com/libra/libra/blob/master/json-rpc/src/methods.rs#L202) to verify the transaction was propagated.  However, that added to either having messy code around manual client building, or extending the JSON RPC client in this PR, so I kept it less complicated and just ensured that the nodes have retrieved the new key first.  It's not ideal but works to verify that the key can be rotated.

### DIFF
--- a/network/network-address/src/lib.rs
+++ b/network/network-address/src/lib.rs
@@ -335,6 +335,22 @@ impl NetworkAddress {
         })
     }
 
+    /// A function to rotate public keys for `NoiseIK` protocols
+    pub fn rotate_noise_public_key(
+        &mut self,
+        to_replace: &x25519::PublicKey,
+        new_public_key: &x25519::PublicKey,
+    ) {
+        for protocol in self.0.iter_mut() {
+            // Replace the public key in any Noise protocols that match the key
+            if let Protocol::NoiseIK(public_key) = protocol {
+                if public_key == to_replace {
+                    *protocol = Protocol::NoiseIK(*new_public_key);
+                }
+            }
+        }
+    }
+
     #[cfg(any(test, feature = "fuzzing"))]
     pub fn mock() -> Self {
         NetworkAddress::new(vec![Protocol::Memory(1234)])
--- a/testsuite/tests/libratest/smoke_test.rs
+++ b/testsuite/tests/libratest/smoke_test.rs
@@ -3,29 +3,44 @@
 
 use cli::client_proxy::ClientProxy;
 use debug_interface::{libra_trace, NodeDebugClient};
-use libra_config::config::{KeyManagerConfig, NodeConfig, OnDiskStorageConfig, SecureBackend};
-use libra_crypto::{ed25519::Ed25519PrivateKey, hash::CryptoHash, PrivateKey, SigningKey, Uniform};
-use libra_global_constants::{CONSENSUS_KEY, OPERATOR_ACCOUNT, OPERATOR_KEY};
+use libra_config::config::{
+    Identity, KeyManagerConfig, NodeConfig, OnDiskStorageConfig, SecureBackend,
+};
+use libra_crypto::{
+    ed25519::Ed25519PrivateKey, hash::CryptoHash, PrivateKey, SigningKey, Uniform,
+    ValidCryptoMaterial,
+};
+use libra_global_constants::{
+    CONSENSUS_KEY, OPERATOR_ACCOUNT, OPERATOR_KEY, VALIDATOR_NETWORK_KEY,
+};
 use libra_json_rpc::views::{ScriptView, TransactionDataView};
-use libra_key_manager::libra_interface::{JsonRpcLibraInterface, LibraInterface};
-use libra_management::config_builder::FullnodeType;
+use libra_key_manager::{
+    self,
+    libra_interface::{JsonRpcLibraInterface, LibraInterface},
+};
+use libra_management::{config_builder::FullnodeType, constants};
+use libra_network_address::{NetworkAddress, RawNetworkAddress};
 use libra_secure_storage::{CryptoStorage, KVStorage, Storage, Value};
+use libra_secure_time::{RealTimeService, TimeService};
 use libra_swarm::swarm::{LibraNode, LibraSwarm};
 use libra_temppath::TempPath;
 use libra_types::{
     account_address::AccountAddress,
     account_config::{association_address, testnet_dd_account_address, COIN1_NAME},
     ledger_info::LedgerInfo,
-    transaction::authenticator::AuthenticationKey,
+    transaction::{
+        authenticator::AuthenticationKey, RawTransaction, Script, SignedTransaction, Transaction,
+    },
+    validator_config::ValidatorConfig,
     waypoint::Waypoint,
 };
 use num_traits::cast::FromPrimitive;
 use rust_decimal::Decimal;
 use std::{
     collections::BTreeMap,
-    convert::TryInto,
+    convert::{TryFrom, TryInto},
     fs,
-    io::{Result, Write},
+    io::{self, Write},
     path::{Path, PathBuf},
     process::{Command, Stdio},
     str::FromStr,
@@ -174,7 +189,7 @@ impl TestEnvironment {
     }
 }
 
-fn copy_file_with_sender_address(file_path: &Path, sender: AccountAddress) -> Result<PathBuf> {
+fn copy_file_with_sender_address(file_path: &Path, sender: AccountAddress) -> io::Result<PathBuf> {
     let tmp_source_path = TempPath::new().as_ref().with_extension("move");
     let mut tmp_source_file = std::fs::File::create(tmp_source_path.clone())?;
     let mut code = fs::read_to_string(file_path)?;
@@ -235,7 +250,7 @@ fn test_smoke_script(mut client_proxy: ClientProxy) {
         .unwrap();
     assert!(compare_balances(
         vec![(10.0, "Coin1".to_string())],
-        client_proxy.get_balances(&["b", "0"]).unwrap()
+        client_proxy.get_balances(&["b", "0"]).unwrap(),
     ));
     client_proxy.create_next_account(false).unwrap();
     client_proxy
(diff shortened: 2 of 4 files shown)


testsuite/tests/libratest/smoke_test.rs
```diff
@@ -1312,3 +1330,190 @@ fn test_key_manager_consensus_rotation() {
...
+        private_key: operator_key,
+    };
+
+    // Setup an RPC endpoint so we can talk to the node
+    let validator_account = node_config.validator_network.as_ref().unwrap().peer_id();
+    let libra = get_libra_interface(&node_config);
+
+    // Rotate key in storage
+    storage.rotate_key(VALIDATOR_NETWORK_KEY).unwrap();
+    let new_network_key = { node_config.validator_network.unwrap().identity_key() };
```
phlip9, line 1456: nit: do you need these braces?

phlip9, line 1456: it's not clear to me if this is a different key? wouldn't this just be the old key? how is `storage.rotate_key()` modifying this config handle?

maybe add an assertion that it's actually a different key : p

testsuite/tests/libratest/smoke_test.rs
```diff
@@ -1312,3 +1330,190 @@ fn test_key_manager_consensus_rotation() {
...
+    // This is submitting a new reconfig on a different validator using the association key
+    {
+        let node_config_path = swarm.validator_swarm.config.config_files.get(1).unwrap();
+        let node_config = NodeConfig::load(&node_config_path).unwrap();
+        let libra = get_libra_interface(&node_config);
+        submit_new_reconfig(&libra, &time_service, &association_info).unwrap();
+    }
+
+    // Restart validator At this point, the `add_node` call ensures connectivity to all nodes
+    swarm.validator_swarm.kill_node(0);
```
phlip9, line 1499: we probably need to wait for all nodes to observe the reconfig before restarting node 0. at least until validators use other VFNs as backup upstream sync sources

testsuite/tests/libratest/smoke_test.rs
phlip9: edit: nevermind, i see that `identity_key()` queries secure storage

gregnazario: I can check if I can remove them.  I spent a lot of time fighting with mutable vs immutable references, and some of these are necessary.

gregnazario: This is a perfect example of how our function / variable mixing is very confusing along with the storage / config lines.  We should come back at some point to see how we can improve upon this, but basically to get out a key in the right form without converting it manually, is to first rotate the key, then retrieve it from the config.

testsuite/tests/libratest/smoke_test.rs
gregnazario: I've gone ahead and basically mirrored what I did for checking that the network key had been updated, but for reconfig.  Also, did some combining so that there's not repeated work.

JoshLind: Interesting.. I was going to ask what happens if node 0 dies between the time a new validator config is submitted to the blockchain and a reconfiguration occurs? The mental model I had is that a node's network key is only "live" once a reconfiguration occurs, right? So, I'm guessing the issue here would be that node 0 won't be able to reconnect to anyone until the reconfiguration tells all online nodes that the node address has changed?

Do we have plans to address this? e.g., to either get node 0 to retrieve its old VALIDATOR_NETWORK_KEY from storage to create new connections. Or will we simply assume that node 0 won't participate in the validator network temporarily, and instead it'll use the full nodes to stay up to sync until the reconfig occurs, after which point it'll be able to reconnect?

gregnazario: Yes, as I learned from my failed tests previously.  The node that's rotated it's key cannot reconnect if it tries to reconnect for whatever reason before a reconfigure occurs.

Perhaps we could bind a VALIDATOR_NETWORK_KEY change locally to a reconfiguration number, that way a validator once it's given a "validator config" update, the next reconfiguration number will be applied to the versioning in the storage.

I don't think we have current plans to address this.  The idea here is this is an MVP of key rotation, and in the future we'll work towards a more robust and automated network key rotation system.

testsuite/tests/libratest/smoke_test.rs
```diff
@@ -235,7 +250,7 @@ fn test_smoke_script(mut client_proxy: ClientProxy) {
         .unwrap();
     assert!(compare_balances(
         vec![(10.0, "Coin1".to_string())],
-        client_proxy.get_balances(&["b", "0"]).unwrap()
+        client_proxy.get_balances(&["b", "0"]).unwrap(),
```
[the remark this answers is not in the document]
gregnazario: `cargo x fmt` did this automagically!  Otherwise, I wouldn't have sat around to add so many commas

testsuite/tests/libratest/smoke_test.rs
```diff
@@ -1312,3 +1327,185 @@ fn test_key_manager_consensus_rotation() {
         );
     }
 }
+
+struct AccountContext {
+    public_key: Ed25519PublicKey,
```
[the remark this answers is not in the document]
gregnazario: True, because it was for testing, I went through and did it this way, but it's perfectly fine to not copy around a second key and derive it from the private.

testsuite/tests/libratest/smoke_test.rs
```diff
@@ -1312,3 +1327,185 @@ fn test_key_manager_consensus_rotation() {
...
+    private_key: Ed25519PrivateKey,
+    address: AccountAddress,
+}
+
+/// Submits a transaction, and returns appropriate account and sequence number for verification
+fn submit_new_transaction<L: LibraInterface, T: TimeService>(
+    libra: &L,
```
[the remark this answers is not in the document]
gregnazario: Yeah, I don't think they necessarily needed to be generic, mostly just stole it from KeyManager

testsuite/tests/libratest/smoke_test.rs
zekun000: I roughly remember the node will try to use both keys for negotiation?

phlip9: i think the right approach is that the validator should always run a fallback/reliability network to state sync from other vfns. if they ever get partitioned/crash and can't observe the reconfig, they can always sync from the public-facing vfns.

testsuite/tests/libratest/smoke_test.rs
```diff
@@ -1312,3 +1327,185 @@ fn test_key_manager_consensus_rotation() {
...
+
+    // Rotate key in storage
+    storage.rotate_key(VALIDATOR_NETWORK_KEY).unwrap();
+    let mut validator_network = node_config.clone().validator_network.unwrap();
+    let new_network_key = validator_network.identity_key();
+
+    // Submit a transaction telling that the validator key rotated
+    let time_service = RealTimeService::new();
+    let validator_config = libra.retrieve_validator_config(validator_account).unwrap();
+    let validator_config = ValidatorConfig {
```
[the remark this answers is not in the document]
gregnazario: Nope, reduced down some

network/network-address/src/lib.rs
```diff
@@ -335,6 +335,28 @@ impl NetworkAddress {
...
+            if let Protocol::NoiseIK(public_key) = protocol {
+                if public_key == to_replace {
+                    new_protocols.insert(i, Protocol::NoiseIK(*new_public_key));
+                    continue;
+                }
+            }
+            new_protocols.insert(i, protocol.clone());
+        }
+        NetworkAddress::new(new_protocols)
+    }
```
phlip9, line 358: as mentioned in chat, we can probably golf this a bit : p

```rust
    pub fn rotate_pubkeys(&mut self, to_replace: &x25519::PublicKey, new_pubkey: x25519::PublicKey) {
        for proto in self.0.iter_mut() {
            match proto {
                Protocol::NoiseIK(pubkey) if pubkey == to_replace => *pubkey = new_pubkey,
                _ => {}
            }
        }
    }

**Comment by phlip9** [comment]
awesome, lgtm!

@bors-libra delegate+

File: testsuite/tests/libratest/smoke_test.rs [hunk 8]
```diff
@@ -1312,3 +1327,194 @@ fn test_key_manager_consensus_rotation() {
...
+        consensus_key.to_bytes().to_vec(),
+        validator_network_key.to_bytes(),
+        validator_network_address.into(),
+        full_node_network_key.to_bytes(),
+        full_node_network_address.into(),
+    );
+    submit_new_transaction(libra, time_service, account_context, script)
+}
+
+fn submit_new_reconfig(
```
zekun000, line 1386: those functions probably should be associated methods of JsonRpcLibraInterface

testsuite/tests/libratest/smoke_test.rs
```diff
@@ -1312,3 +1327,194 @@ fn test_key_manager_consensus_rotation() {
...
+    let script = Script::new(
+        libra_transaction_scripts::RECONFIGURE_TXN.clone(),
+        vec![],
+        vec![],
+    );
+    submit_new_transaction(libra, time_service, account_context, script)
+}
+
+/// Helper function to build libra interfaces for smoke test
+fn get_libra_interface(node_config: &NodeConfig) -> JsonRpcLibraInterface {
```
zekun000, line 1400: this probably should be part of the TestEnvironment or LibraSwarm, given an index return the Interface

gregnazario: Thanks for the suggestions!  I think these are great followups for the helper functions of where they can be leveraged to help further development, once I get done with the next thing, I'll see if I can circle back and poke a bit at them.  But, it's not really high priority right now.

zekun000: it's just nit, no worries

PR #4814, opened 2020-06-29
aggressive prune_window in cluster-test

msmouse:
## Motivation

Setting to 50k versions, so that state pruning kicks in after about 1 min under full load.
 

Yes

## Test Plan

scripts/cti --pr 4814  --run bench

http://grafana.ct-2-k8s-testnet.aws.hlw3truzy4ls.com/d/2Q0p7AVWk/storage?orgId=1&from=1593454192129&to=1593454577844

## Related PRs

### DIFF
--- a/testsuite/cluster-test/src/cluster_builder.rs
+++ b/testsuite/cluster-test/src/cluster_builder.rs
@@ -23,7 +23,7 @@ pub struct ClusterBuilderParams {
     #[structopt(long, default_value = "1")]
     pub fullnodes_per_validator: u32,
     #[structopt(long, use_delimiter = true, default_value = "")]
-    pub cfg: Vec<String>,
+    cfg: Vec<String>,
     #[structopt(long, parse(try_from_str), default_value = "30")]
     pub num_validators: u32,
     #[structopt(long)]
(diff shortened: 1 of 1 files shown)


andll:
Qq - what was TPS of the bench with this change?
Did it change or is it the same?

msmouse:
@andll  I got "all up : 1288 TPS, 3502 ms latency, 4100 ms p99 latency, no expired txns" -- a little lower than 1294 I last saw in #changlog

andll:
One more q - is it right that reduce prune window on all the validators? What is real prod scenario - are all validators going to have 'spike of pruning' at the same time? Or is it going to be balanced across validators? (In second case, should we only enable shorten pruning on one validator in this test?)

msmouse:
They all prune continuously at the same time once the configured window is full.

andll:
Ok, lets merge it as-is then
@bors-libra r+

In general I think good practice is to stagger those things so that maintenance operations don't happen synchronously, but this is probably not relevant to this change

msmouse:
Right, this is supposed to happen continuously to NOT create peaks in the first place. What you said applies to the coming state snapshot backup better. Thanks!

PR #4817, opened 2020-06-29
[Libra types] use serde_bytes to optimize Vec<u8> + format linter

matbd:
## Motivation

* Many languages have a native notion of bytes that is more efficient to (de)serialize than `Vec<u8>`. Let's use it consistently in our formats.

* Add linter to detect regressions in this respect as well as other issues.

## Test Plan

CI

### DIFF
--- a/language/move-core/types/src/transaction_argument.rs
+++ b/language/move-core/types/src/transaction_argument.rs
@@ -11,7 +11,7 @@ pub enum TransactionArgument {
     U64(u64),
     U128(u128),
     Address(AccountAddress),
-    U8Vector(Vec<u8>),
+    U8Vector(#[serde(with = "serde_bytes")] Vec<u8>),
     Bool(bool),
 }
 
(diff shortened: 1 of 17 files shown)


language/move-core/types/src/transaction_argument.rs
```diff
@@ -11,7 +11,7 @@ pub enum TransactionArgument {
     U64(u64),
     U128(u128),
     Address(AccountAddress),
-    U8Vector(Vec<u8>),
+    U8Vector(#[serde(with = "serde_bytes")] Vec<u8>),
```
thefallentree, line 14: should we rename this to BYTES as well?

matbd: I think it would make a lot of sense but I cannot speak for Move specialists (@vgao1996 @sblackshear @dariorussi what do you think?)

matbd: note: I'm going to land this in the meantime, but we can always iterate in a next PR.

matbd:
@bob-wilson r=thefallentree

bob-wilson:
I think you probably meant to tag bors, not bob-wilson, but since you got my attention, it would be nice to hold off to give a few more people a chance to look at this.

matbd:
OMG what did just happen :)

matbd:
OK I should listened to Bob because I forgot to plug in the linter :)

PR #4819, opened 2020-06-29
[move-prover] Enabling modular verification.

wrwg:
This was easy after the change to eliminate redundant verification of postconditions.

This feature works as follows. With

```
spec fun foo {
    pragma opaque = true;
    ...
}
```

... a function will be verified standalone, but its body will not longer be inlined at call sites. However, the specified pre/post/invariant conditions will still be enforced at the call site.

This can be combined with turning verification off:

```
spec fun foo {
    pragma opaque = true;
    pragma verify = false;
    ...
}
```

With this, pre and post conditions are used at call sites, but the function itself will not be verified. This can be used to deal with notorious diffcult problems like loops or to temporarily work around Z3 termination problems.

As always, the pragmas can also be used at module level to effect all functions in a given module.

## Motivation

Modular verification.

Yes

## Test Plan

Added test

## Related PRs

NA

### DIFF
--- a/language/move-prover/tests/sources/functional/opaque.move
+++ b/language/move-prover/tests/sources/functional/opaque.move
@@ -1,0 +1,23 @@
+module TestOpaque {
+
+    spec module {
+        pragma verify = true;
+    }
+
+    // Function which has a wrong post condition, so verification fails.
+    fun opaque_incorrect(): u64 {
+        1
+    }
+    spec fun opaque_incorrect {
+        pragma opaque = true;
+        ensures result == 2;
+    }
+
+    fun opaque_caller(): u64 {
+        opaque_incorrect()
+    }
+    spec fun opaque_caller {
+        // because we only use the post condition but not the definition, this should verify
+        ensures result == 2;
+    }
+}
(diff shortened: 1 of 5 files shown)


language/move-prover/tests/sources/functional/opaque.move
```diff
@@ -0,0 +1,23 @@
...
+    spec fun opaque_incorrect {
+        pragma opaque = true;
+        ensures result == 2;
+    }
+
+    fun opaque_caller(): u64 {
+        opaque_incorrect()
+    }
+    spec fun opaque_caller {
+        // because we only use the post condition but not the definition, this should verify
```
junkil-park, line 20: Perhaps, this should be the opposite? Like:
"we only use the definition but not the post-condition of `opaque_incorrect()`?"

wrwg: Nope it means just this. Here, where we call this function, we only use its post condition. Therefore this verifies OK (even though the definition returns 1).

junkil-park: Ah, I now understand it. Thanks.

Issue #4308, opened 2020-06-05
Stdlib cleanup: Replace hardcoded error codes with sensible constants [P2]


sblackshear:
The use of error codes in the standard library is not very  disciplined. Once we have constants in the source language, we can apporiately named constants for each error code.

PR #4822, opened 2020-06-29
[libra-types] Change major status to VMStatus in prep for future

tzakian:
This lays the external groundwork for the changes to the VMStatus that @tnowacki is working on at the moment and plumbs this additional information in to the stored structures for the result of a transaction to allow off-node clients to debug transactions that failed to execute. 

The plan for the new `VMStatus` is
```rust
pub struct VMStatus {
   major_status: StatusCode,
   info: Option<(Option<ModuleId>, u64)>,
}
```
* `major_status` is as before
* `info` is `None` if there is no further information provided (e.g., for an `OUT_OF_GAS` error). In the case of an `ABORTED` major status, this will be a `Some`, and contain the
  - The `ModuleId` for the module that aborted. This is `None` if it was a script that aborted.
  - The status code that it aborted with. 

This PR plumbs this in to the structures that are used for storage, and the externally visible interfaces so that we can easily hook this up once the changes to the VM are complete. 

This is part of #4308

### DIFF
--- a/json-rpc/types/src/views.rs
+++ b/json-rpc/types/src/views.rs
@@ -15,7 +15,7 @@ use libra_types::{
     ledger_info::LedgerInfoWithSignatures,
     proof::{AccountStateProof, AccumulatorConsistencyProof},
     transaction::{Transaction, TransactionArgument, TransactionPayload},
-    vm_status::StatusCode,
+    vm_status::{StatusCode, VMStatus},
 };
 use move_core_types::{
     identifier::Identifier,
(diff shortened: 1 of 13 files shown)


lightmark:
lgtm

json-rpc/types/src/views.rs
```diff
@@ -313,13 +313,40 @@ impl From<&Vec<u8>> for BytesView {
...
+}
+
 #[derive(Clone, Debug, Deserialize, Serialize, PartialEq)]
 pub struct TransactionView {
     pub version: u64,
     pub transaction: TransactionDataView,
     pub hash: String,
     pub events: Vec<EventView>,
-    pub vm_status: StatusCode,
+    pub vm_status: VMStatusView,
```
thefallentree, line 349: nit: If we just add more items here instead of introducing an new structure, we will make it an non-breaking change.

json-rpc/types/src/views.rs
```diff
@@ -313,13 +313,40 @@ impl From<&Vec<u8>> for BytesView {
     }
 }
 
+#[derive(Clone, Debug, Deserialize, Serialize, PartialEq)]
+pub struct VMStatusView {
+    pub major_status: StatusCode,
```
thefallentree, line 318: the _status  suffix here seems to be not necessary

json-rpc/types/src/views.rs
tzakian: Good point! I would be ok with breaking these out if it would be a significant issue, but I do have a preference for not breaking this out since I think these should be viewed as logically connected.

json-rpc/types/src/views.rs
```diff
@@ -313,13 +313,37 @@ impl From<&Vec<u8>> for BytesView {
     }
 }
 
+#[derive(Clone, Debug, Deserialize, Serialize, PartialEq)]
+pub struct VMStatusSourceContextView {
```
tnowacki, line 317: What does this do?

tzakian: If there is a source-level error (e.g., `abort 5`) this is what will return that information out (`context: 0x1::Coin1, error: 4`). It is separated out from the major code since this does not need to exist always and is only for "source-derived" errors.

tnowacki: Why is it just a string then? And why is it just one code and not two? Major and minor?

tzakian: Because this is being sent out over the wire to clients (e.g. someone `Curl`ing it). We could send hex bytes for these things if we wanted to....but I don't see the use for that really. For on-chain data structures, we do store this as a moduleid etc.

In terms of major/minor. The major is held in the `vm_status` view for the `TransactionView`. This struct is meant as a context for errors that have further information to provide.

tzakian:
These changes were incorporated into #5071 so this PR is no longer needed.

PR #4828, opened 2020-06-30
[generate-format] add linter for Serde formats (take 2)

matbd:
## Motivation

* Activate the Serde format linter for good (I meant to do it in https://github.com/libra/libra/pull/4817).
* Fix the last occurrence of `SEQ[U8]` in bitvec (this makes (de)serialization slightly faster and changes nothing otherwise for Rust).

## Test Plan

CI

### DIFF
--- a/testsuite/generate-format/tests/detect_format_change.rs
+++ b/testsuite/generate-format/tests/detect_format_change.rs
@@ -20,8 +20,16 @@ fn analyze_serde_formats() {
             assert_registry_has_not_changed(&corpus.to_string(), path, registry.clone(), expected);
         }
 
-        // Test that the definitions in all corpus are unique.
+        // Test that the definitions in all corpus are unique and pass the linter.
         for (key, value) in registry {
+            assert_eq!(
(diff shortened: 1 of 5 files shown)


testsuite/generate-format/tests/detect_format_change.rs
```diff
@@ -20,8 +20,16 @@ fn analyze_serde_formats() {
...
-        // Test that the definitions in all corpus are unique.
+        // Test that the definitions in all corpus are unique and pass the linter.
         for (key, value) in registry {
+            assert_eq!(
+                generate_format::lint_lcs_format(&value),
+                Ok(()),
+                "In corpus {}: lint error while analyzing {}",
+                corpus.to_string(),
+                key
+            );
```
gregnazario, line 31: Will this tell us what the lint error was when it failed?  It looks like it'll tell you where the problem is, but not what the problem is.

matbd: Yes it will, as part of the explanation why `Err(err) != Ok(())`.

PR #4829, opened 2020-06-30
[cluster-test] Introduce cti --build-local for local build

andll:
Small incremental changes are faster to build locally, rather then using aws.

Local build takes about 2 minutes, while AWS build takes 9 minutes.
It also does not require creating pull request

This change also disables LTO - this reduces TPS by roughly 5%, however LTO takes insane amount of build time, so this seem like a reasonable trade off.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -163,7 +163,6 @@ default-members = [
 
 [profile.release]
 debug = true
-lto = 'thin'
 
 [profile.bench]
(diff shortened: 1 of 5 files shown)


snakech:
thx for doing this

thefallentree:
what, why was lto disabled ? 

5% trade-off in release seem very reasonable.

PR #4830, opened 2020-06-30
[Libra Doc tool] initial version

matbd:
## Motivation

Make it easy to keep specification in sync w.r.t. Serde formats.

The Rust printing logics is provided by `serde-generate` but it should be no problem tweaking and adding options there if needed:
https://github.com/facebookincubator/serde-reflection/blob/master/serde-generate/src/rust.rs#L39

## Test Plan

```
cargo run -p libra-documentation-tool testsuite/generate-format/tests/staged/consensus.yaml --update-libra-specs-dir ../libra-specs/common
```

### DIFF
--- a/common/libradoc/Cargo.toml
+++ b/common/libradoc/Cargo.toml
@@ -1,0 +1,32 @@
+[package]
+name = "libra-documentation-tool"
+version = "0.1.0"
+authors = ["Libra Association <opensource@libra.org>"]
+description = "Libra transaction-builder"
+repository = "https://github.com/libra/libra"
+homepage = "https://libra.org"
+license = "Apache-2.0"
+edition = "2018"
+
+[dependencies]
+serde_yaml = "0.8.13"
+serde-generate = "0.5.1"
+serde-reflection = "0.3.0"
+anyhow = "1.0.31"
+regex = "1.3.9"
+structopt = "0.3.15"
+once_cell = "1.4.0"
+
+libra-workspace-hack = { path = "../../common/workspace-hack", version = "0.1.0" }
+
+[dev-dependencies]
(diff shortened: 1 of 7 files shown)


common/libradoc/Cargo.toml
```diff
@@ -0,0 +1,32 @@
+[package]
```
mimoo, line 1: nit: since this is not really a prod dependency, would it make sense to move this to a different folder? (e.g. tooling/, specs/, testsuite/)

mimoo:
@bors-libra delegate+

Thanks for working on that!

common/libradoc/Cargo.toml
matbd: I don't mind as long as such directory exists already. Note that I see some proptest helpers and workspace hack in `common/`.

mimoo: I wish the distinction between prod stuff and dev stuff was clearer FWIW

common/libradoc/tests/update.rs
```diff
@@ -0,0 +1,74 @@
...
+}
+
+#[derive(Deserialize)]
+struct MyStruct(u64);
+
+#[test]
+fn test_doctool() {
+    let mut tracer = Tracer::new(TracerConfig::default());
+    let samples = Samples::new();
+    tracer.trace_type::<MyEnum>(&samples).unwrap();
```
davidiw, line 26: that's kind of interesting in that the enum is used here but allow dead code is required.

matbd: Let me double-check.

matbd: So the linter complains that we never read the value in the field `a` of MyEnum::Struct (which is true).

PR #4833, opened 2020-06-30
[move prover] update a few more tests to use "signer" instead of "sender"

bob-wilson:
Most of these seem straightforward. I could not get the trace200527.move test to continue reproducing the regression that it was intended to demonstrate.

## Motivation

Removing support for the obsolete "sender" features

Yes

## Test Plan

Updated tests and confirmed that they pass.


bob-wilson:
@wrwg I added a "TODO" comment to trace200527.move as you suggested. Does this look OK then?

PR #507, opened 2019-08-09
[ci] Enable sccache to speed up build times

metajack:
This installs and uses sccache for builds. The resulting cache file is stored
between CircleCI runs using CicleCI's caching system. While not perfectly
cacheable yet, this saves about 30% of build time.

See blog post about this at https://medium.com/@edouard.oger/rust-caching-on-circleci-using-sccache-c996344f0115

## Motivation

Builds times can always be faster.

### Have you read the [Contributing Guidelines on pull requests](https://github.com/libra/libra/blob/master/CONTRIBUTING.md#pull-requests)?

yes

## Test Plan

Current build times are ~15-20m on CI. This should reduce builds to ~10-15m. You can check in the [CircleCI jobs page](https://circleci.com/gh/libra)

## Related PRs

none

PR #4839, opened 2020-06-30
[CI] use sccache

sausagee:
## Motivation
Use sccache to speed up builds. 

Summary of sccache implementation
- stored via CircleCI's builtin cache 
- indexed by build type (dev, release) and date
- rebuilt in the first PR of the day
- 1.5 min overhead to install sccache 
- 0.5 min overhead to retrieve cache
- 1.5 min overhead to save cache

Overall Gains
- 9.5 min -> 6.5 min in build-dev
- 8 min -> 4.5 min in build-release

## Test Plan
CI 
image

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -88,6 +88,43 @@ commands:
       - run:
           name: Install nightly toolchain for features not in beta/stable
           command: rustup install nightly
+  install_sccache:
+    steps:
+      - run:
+          name: Install scccache
+          command: |
+            cargo install sccache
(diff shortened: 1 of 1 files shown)


huitseeker:
I believe @metajack had interesting experiments re:sccache in #507, #619, #908

metajack:
This looks identical to my initial version, which ended up being a wash, however, Young has a lot more knowledge of CircleCI due to 9 months of other improvement work, so it might work now. Also, we should test sccache on GHA to see if it works there.

rexhoffman:
Those gains look worth it to me.   If we get our act together with features we may need to revisit this (less to compile may mean time savings is negated).

PR #4847, opened 2020-06-30
[network] derive public keys from network addresses

phlip9:
+ all onchain discovery sources now derive pubkey sets from addrs (also temporarily include the explicit pubkey field, but will remove that soon).

+ support multiple public keys for each peer. trusted peers is now modeled as a `HashMap<PeerId, HashSet<x25519::PublicKey>>` (instead of a `HashMap<PeerId, x25519::PublicKey>`).

+ support multiple discovery sources for deriving trusted peers set. connmgr will union pubkey sets across discovery sources.

+ rename `network_peers` to `seed_pubkeys` in config


network/builder/src/builder.rs
```diff
@@ -379,10 +370,26 @@ impl NetworkBuilder {
...
         };
 
+        // union pubkeys from addrs with pubkeys directly in config
+        let addr_pubkeys_iter = seed_addrs.iter().map(|(peer_id, addrs)| {
+            let pubkey_set: HashSet<_> = addrs
+                .iter()
+                .filter_map(NetworkAddress::find_noise_proto)
+                .collect();
+            (*peer_id, pubkey_set)
+        });
```
gregnazario, line 380: Should you `filter_map` the outside?  Maybe there's no noise protocol in one.

network/builder/src/builder.rs
```diff
@@ -90,8 +90,9 @@ pub struct NetworkBuilder {
     // TODO(philiphayes): better support multiple listening addrs
     listen_address: NetworkAddress,
     advertised_address: Option<NetworkAddress>,
-    seed_peers: HashMap<PeerId, Vec<NetworkAddress>>,
-    trusted_peers: Arc<RwLock<HashMap<PeerId, x25519::PublicKey>>>,
+    trusted_peers: Arc<RwLock<HashMap<PeerId, HashSet<x25519::PublicKey>>>>,
```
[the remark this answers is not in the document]
mimoo: I'd say it's the correct data structure to use in any case (it'd be annoying to implement a set on top of a vec). Also you're supposed to only talk to a fixed number of validators (in the hundreds) so it's not that much.

network/builder/src/builder.rs
```diff
@@ -278,6 +278,15 @@ impl NetworkBuilder {
         &mut self,
         trusted_peers: HashMap<PeerId, x25519::PublicKey>,
     ) -> &mut Self {
+        // TODO(philiphayes): remove
```
mimoo: why the TODO?

network/src/connectivity_manager/mod.rs
```diff
@@ -163,16 +163,16 @@ where
         max_delay_ms: u64,
         connection_limit: Option<usize>,
     ) -> Self {
+        // TODO(philiphayes): replace this with better approach
         {
             // Reconcile the keysets eligible is only used to allow us to dial the remote peer
```
mimoo: I don't understand this comment :D

network/src/noise/handshake.rs
```diff
@@ -348,13 +348,13 @@ impl NoiseUpgrader {
                 })?
                 .get(&remote_peer_id)
             {
-                Some(key) => {
-                    if key != &remote_public_key {
+                Some(remote_pubkey_set) => {
```
mimoo, line 351: what about outbound stuff :D?

config/src/config/network_config.rs
```diff
@@ -29,7 +29,7 @@ use std::{
 // might need to extract into a separate network_constants crate or something.
 pub const HANDSHAKE_VERSION: u8 = 0;
 
-pub type NetworkPeersConfig = HashMap<PeerId, x25519::PublicKey>;
+pub type SeedPubkeySetsConfig = HashMap<PeerId, HashSet<x25519::PublicKey>>;
```
davidiw, line 32: I don't think the nomenclature is quite correct in either case...
in practice this will only be used when the remote peers are /not/ on chain discovery, mutual auth, on a unidirectional network, something like a Private(V)FN. I don't imagine we'll have a good case for multiple keys. I would kinda prefer we leave this well enough alone and not modify it, since there really isn't a clear V1 use case that we need to support with it.

davidiw: looking below, I think the term "trusted_peers" and TrustedPeers or TrustedPeersSet is reasonable!

network/simple-onchain-discovery/src/lib.rs
```diff
@@ -96,9 +101,22 @@ fn extract_updates(role: RoleType, node_set: ValidatorSet) -> Vec<ConnectivityRe
...
+                let pubkey = public_key(role, node.config());
+                // parse out pubkey from address
+                let pubkey_set: HashSet<_> = network_address(role, node.config())
+                    .ok()
+                    .iter()
+                    .filter_map(NetworkAddress::find_noise_proto)
+                    // for now, also add the explicit pubkey field
+                    .chain(iter::once(pubkey))
+                    .collect();
+                (*node.account_address(), pubkey_set)
```
davidiw, line 118: what's the value in having UpdateEligibleNodes at this point? Can't this be derived from UpdateAddresses?

network/src/connectivity_manager/builder.rs
```diff
@@ -29,8 +28,9 @@ pub type ConnectivityManagerService = ConnectivityManager<Fuse<Interval>, Expone
 /// The configuration fields for ConnectivityManager
 struct ConnectivityManagerBuilderConfig {
     network_context: Arc<NetworkContext>,
-    eligible: Arc<RwLock<HashMap<PeerId, x25519::PublicKey>>>,
-    seed_peers: HashMap<PeerId, Vec<NetworkAddress>>,
+    eligible: Arc<RwLock<HashMap<PeerId, HashSet<x25519::PublicKey>>>>,
+    seed_addrs: HashMap<PeerId, Vec<NetworkAddress>>,
+    seed_pubkey_sets: HashMap<PeerId, HashSet<x25519::PublicKey>>,
```
davidiw, line 33: what's the incentive to maintain both sets? We can pretty cheaply get the key from the address, no?

config/src/config/network_config.rs
```diff
@@ -110,19 +110,11 @@ impl NetworkConfig {
...
                 .ok_or_else(|| Error::InvariantViolation("No local IP".to_string()))?;
         }
-
-        if network_role.is_validator() {
-            crate::config::invariant(
-                self.network_peers.is_empty(),
-                "Validators should not define network_peers".into(),
-            )?;
-        }
-
```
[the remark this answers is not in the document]
phlip9: it wasn't really an invariant, just that it was previously pointless to set the trusted peers as a validator, since consensus would immediately clobber it with an `UpdateEligibleNodes`. with this diff, we can support multiple sources for trusted peers, so it's fine to add pubkeys from config.

network/builder/src/builder.rs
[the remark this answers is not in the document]
phlip9: definitely agree a Vec would be more efficient backing since each set is very small. in fact, we could probably get away with [tinyvec](https://docs.rs/tinyvec/0.3.3/tinyvec/) to remove an extra indirection. we're just missing some convenient wrappers to emulate set-like api, which i would probably tackle later, when we have more time to tackle optimizations.

network/builder/src/builder.rs
phlip9: it's correctly handled already. 

if no pubkeys from addr, then pubkey_set is empty HashSet. then, `.extend(pubkey_set)` will just do nothing.

connectivity manager will also just ignore the whole peer entry if no pubkeys from either addrs or config.

network/simple-onchain-discovery/src/lib.rs
phlip9: i've gone back and forth on this a bit, but it's mostly because (1) i don't want gossip discovery polluting the trusted peers set and (2) i feel it's cleaner if connectivity manager is unaware of network address structure.

network/src/connectivity_manager/builder.rs
phlip9: for potential client/server model in the future, i want to avoid conflating pubkeys from addresses we can dial and pubkeys we'll accept for inbound connections. for example, maybe a server will only accept connections from some set of pubkeys, but doesn't have any addresses that it can dial (since it's a server).

network/src/noise/handshake.rs
phlip9: hmm, we never dial addrs of peers not in the trusted peers set, so the only point would be to more quickly reject an outbound connection to a peer that's been removed between when we started the dial and when the handshake completes (as opposed to waiting a second or two for connectivity manager to clean up stale connections).

network/builder/src/builder.rs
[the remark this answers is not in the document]
gregnazario: Yeah, I was just posing the question.  A set is the right way to go definitely.

network/src/noise/handshake.rs
mimoo: oh yeah...

network/src/connectivity_manager/builder.rs
zekun000: probably can use better naming to represent two sets like allowed inbound/outbound? but it also looks over optimization for now

network/src/connectivity_manager/mod.rs
```diff
@@ -596,6 +684,26 @@ fn log_dial_result(
     }
 }
 
+/////////////////////
+// DiscoverySource //
+/////////////////////
+
+impl DiscoverySource {
+    fn as_usize(self) -> usize {
```
zekun000, line 671: fwiw, I just use `discriminant`

network/builder/src/builder.rs
```diff
@@ -90,8 +90,9 @@ pub struct NetworkBuilder {
     // TODO(philiphayes): better support multiple listening addrs
     listen_address: NetworkAddress,
     advertised_address: Option<NetworkAddress>,
-    seed_peers: HashMap<PeerId, Vec<NetworkAddress>>,
-    trusted_peers: Arc<RwLock<HashMap<PeerId, x25519::PublicKey>>>,
+    trusted_peers: Arc<RwLock<HashMap<PeerId, HashSet<x25519::PublicKey>>>>,
+    seed_addrs: HashMap<PeerId, Vec<NetworkAddress>>,
+    seed_pubkey_sets: HashMap<PeerId, HashSet<x25519::PublicKey>>,
```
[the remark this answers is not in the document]
phlip9: right now i'm mirroring the internal structure. if/when i merge the UpdateAddresses / UpdateEligibleNodes i'll probably mirror whatever the internal structure looks like after that (which will probably look like `(Vec<NetworkAddress>, HashSet<x25519::PublicKey>)`)

network/builder/src/builder.rs
```diff
@@ -203,34 +205,21 @@ impl NetworkBuilder {
...
-            info!(
-                "network setup: role: {}, seed_peers: {:?}, trusted_peers: {:?}",
-                role, seed_peers, trusted_peers,
-            );
-
             network_builder
                 .authentication_mode(AuthenticationMode::Mutual(identity_key))
-                .trusted_peers(trusted_peers)
-                .seed_peers(seed_peers)
+                .seed_addrs(config.seed_peers.clone())
```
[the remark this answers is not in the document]
phlip9: renamed seed_peers -> seed_addrs and pubkey_set -> pubkeys everywhere!

network/src/connectivity_manager/mod.rs
phlip9: hmm, i'm not sure how to use discriminant to index into the array : d

network/builder/src/builder.rs
```diff
@@ -153,53 +154,37 @@ impl NetworkBuilder {
...
-            }
+        // Sanity check seed addresses.
+        config
+            .verify_seed_addrs()
+            .expect("Seed addresses must be well-formed");
+
+        // Don't turn on connectivity manager if we're a public facing vfn for example.
+        // TODO(philiphayes): could probably use a better way to specify these cases
+        if config.mutual_authentication
+            || config.discovery_method == DiscoveryMethod::Onchain
```
davidiw, line 176: curious why this is config.discovery != DiscoveryMethod::None -- but this is probably just as good...

config/src/config/network_config.rs
```diff
@@ -29,8 +29,8 @@ use std::{
 // might need to extract into a separate network_constants crate or something.
 pub const HANDSHAKE_VERSION: u8 = 0;
 
-pub type NetworkPeersConfig = HashMap<PeerId, x25519::PublicKey>;
-pub type SeedPeersConfig = HashMap<PeerId, Vec<NetworkAddress>>;
+pub type SeedPubkeysConfig = HashMap<PeerId, HashSet<x25519::PublicKey>>;
```
davidiw, line 32: the term "config" seems redundant for this and below... it is just SeedPub(lic)Keys and SeedAddr(esse)s

network/builder/src/builder.rs
phlip9: yeah this is a pretty weird condition imo, i just preserved the behaviour that was already there. thinking a bit more, i think the only cases where we don't want connmgr are server-only use cases. that this is hard to specify probably means we could be more explicit in our configs

config/src/config/network_config.rs
phlip9: agreed  : )

network/src/connectivity_manager/mod.rs
zekun000: ah, I was using it to index a hashmap

PR #4848, opened 2020-06-30
[libra framework] Remove `PublishModule` from privileges

tzakian:
With the changes to Roles that David is working on, this will break the module publishing story that we have. This just moves that resource to no longer be a privilege so that the future changes won't break it.

This PR also removes some unused system module names.


tzakian:
This resource is still used by the VM to determine if the account can publish a module. We could remove it if we have module publishing based on a role check, but this then means that other accounts without that specific role won't be able to publish modules without a VM upgrade (and that every account with that role can publish).

PR #3499, opened 2020-04-18
[Bytecode verifier] Add (unused) structural control flow checks

tnowacki:
## Landing plans 

Currently, the checks described below are *not* hooked up to the code unit verifier. This means they are not run on any code that normally passes through the bytecode verifier. 

The plan is to land this PR and continue testing in a parallel implementation

## Motivation

Long term, we want to get the bytecode verifier to a place where the cost can be calculated easily, and up front. Currently, the big thorn preventing the cost from being represented by an easy formula is the abstract interpreter. The traversal of blocks is ill-defined, and the number of iterations needed to reach a fix point is unknown.

A long term solution would be to require all "back edges" in the control flow graph to have their abstract state annotated. This would be a massive undertaking to implement and likely isn't necessary to support the current core modules.

An interim solution is to require that all "back edges" cannot change the abstract state. This means that when checking a loop, it cannot result in any changes to the pre-state at the beginning of the loop. In less technical terms, loops cannot add new information the the check being performed.

## Open Issues

The reason I think this PR is important to get up ASAP is for these open questions

Critical:

* Does an 'interim solution' implementation require changes to the bytecode file format? (do we need new instructions/labels?)
* Does an 'interim solution' require changes to the core modules?

Less Critical:

* Do we ever need the 'long term' solution of annotated back edges?

## An 'Interim Solution' Implementation

* Step 1) Enforce a "structured" control flow on the bytecode. This allows for the identification of back edges.
* Step 2) Utilize the structure to require that "back edges" do not change the state

This PR provides Step 1 of 2.

## Specifics of this PR

* This PR adds a new verification step, `ControlFlowVerifier`, that checks certain properties about jumps inside of loops
* These restrictions go above and beyond what is strictly necessary for the 'Interim Solution' just to make our story very simple
* The rules that it enforces are as follows:
  * All "continues" (back jumps in a loop) are only to the current loop
  * All "breaks" (forward, loop-exiting jumps) go to the "end" of the loop
  * All forward jumps do not enter into the middle of a loop
* I think that only this first rule "All 'continues' (back jumps in a loop) are only to the current loop" is necessary for the back edge restriction on the CFG
* But the other rules are nice in that they keep loops incredibly sane, which gives us more flexibility for changes (not just in the bytecode verifier) in the future. 

### Initial Data Gathered

* The first step is to find all of the "loop heads".
* We consider any `CodeOffset` to be a loop head if there is a "back jump" to that offset
  * A jump where the instruction index is less than or equal to the target offset
[...]

PR #4853, opened 2020-06-30
[bytecode verifier] Enable control flow verifier

tnowacki:
- Enabled control flow checks in the bytecode verifier
- The control flow verifier has been running over the stdlib for months with no hiccups. Should be safe to turn on
- See #3499 for more details

## Test Plan

- cargo test


tnowacki:
Blocked by #4860

Issue #4854, opened 2020-06-30
[Bug] Equality assertion fails for vector element reference


emmazzz:
Running the following code gives me move prover error:
```
module TestVector {
    use 0x1::Vector;
    struct T {
        x: u64,
    }

    public fun update_ith_elem(v: &mut vector<T>, i: u64): bool {
        let size = Vector::length(v);
        if (i >= size) {
            return false
        };
        let elem = Vector::borrow_mut(v, i);
        let int_ref = &mut elem.x;

        *int_ref = 42;
        spec {
            assert int_ref == v[i].x;
        };
        true
    }
}
```

Error looks like this:
```
error:  This assertion might not hold.

    ┌── tests/sources/regression/vector_200630.move:17:13 ───
    │
 17 │             assert int_ref == v[i].x;
    │             ^^^^^^^^^^^^^^^^^^^^^^^^^
    │
    =     at tests/sources/regression/vector_200630.move:7:5: update_ith_elem
    =         v =
    =           vector{
    =             (size): 21240,
    =             21239: TestVector.T{x = 0},
    =             default: undef},
    =         i = 21239,
    =         elem = &TestVector.T{x = 0},
    =         int_ref = <? List([Literal("$Boolean"), Literal("false")])>,
    =         int_ref = &0,
    =         size = 21240

```
Why does it say `int_ref` is a boolean in the error message. Is there a bug in the type inference or in the memory model?

DavidLDill:
I'm guessing that the update doesn't "happen" until the end of the function call, and the assert in a spec block is testing it before then.  (@shazqadeer is the expert).

emmazzz:
The following post condition also doesn't work:
```
spec fun update_ith_elem {
        ensures (i < len(v)) ==> v[i].x == 42;
}
```

emmazzz:
> I'm guessing that the update doesn't "happen" until the end of the function call, and the assert in a spec block is testing it before then. (@shazqadeer is the expert).

Yes. That's likely what is happening here. Any way we can make the post condition in my last comment work?

PR #4856, opened 2020-07-01
[move prover] partial specs for LibraSystem and fix assumes in loop

emmazzz:
## Motivation

This PR adds partial specs to LibraSystem and fixes a bug related to loops.
Most of the functions have been specified and verified. The rest is pending issue #4872. 
This PR also resolves issue #4854.

Yes.

## Test Plan

cargo test


language/stdlib/modules/LibraSystem.move
```diff
@@ -182,14 +191,27 @@ module LibraSystem {
         };
 
         let i = 0;
-        while (i < size) {
+        while ({
+            spec {
+                assert i <= size + 1;
+                assert forall j in 0..i: validators[j].addr != addr;
+            };
+            (i <= size)
```
tzakian, line 199: Shouldn't this be `<` instead of `<= size`?

PR #4857, opened 2020-07-01
DRAFT! [Management] Support owner and operator separation in genesis!

JoshLind:
*This a draft PR. It is not ready for reviews.

This PR takes a crack at separating out owners from operators in the genesis process.

To separate owners from operators, this PR offers several commits:
1. It updates the management genesis process to create validator owner and validator operator accounts on-chain.
2. It updates the management tool to allow validator owners to execute the set_operator() operation, thus selecting an operator on-chain.
3. It provides a temporary deterministic hashing function for owner and operator account names. This hashing function will be changed/improved in the future, once we've decided on the best approach. Right now it's treated as a black-box to illustrate how account addresses can be derived from names.
4. It updates the "validator-config" command in the management tool to allow operators to specify owners using their names, rather than account addresses.
5. It updates the set-operator() and validator-config() management commands to determine the sender of each transaction based upon the entity running the command. Right now, we do this by parsing the remote backend namespace provided to the tool, but a better approach would be first have operators and/or owners run an explicit command which tells the tool which entity is using it (e.g., set-tool-executor())
6. It attempts to update and fix the smoke tests that are based on the genesis management code as well as all other tests that are impacted by these changes.

### DIFF
--- a/config/management/src/config_builder.rs
+++ b/config/management/src/config_builder.rs
@@ -113,19 +113,9 @@ impl<T: AsRef<Path>> ValidatorBuilder<T> {
 
         let validator_network = config.validator_network.as_mut().unwrap();
         let validator_network_address = validator_network.listen_address.clone();
-        validator_network.identity = Identity::from_storage(
-            libra_global_constants::VALIDATOR_NETWORK_KEY.into(),
-            libra_global_constants::OPERATOR_ACCOUNT.into(),
-            self.secure_backend(&local_ns, "validator"),
-        );
 
         let fullnode_network = &mut config.full_node_networks[0];
         let fullnode_network_address = fullnode_network.listen_address.clone();
-        fullnode_network.identity = Identity::from_storage(
-            libra_global_constants::FULLNODE_NETWORK_KEY.into(),
-            libra_global_constants::OPERATOR_ACCOUNT.into(),
-            self.secure_backend(&local_ns, "full_node"),
-        );
 
         self.storage_helper
             .validator_config(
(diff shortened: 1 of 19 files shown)


config/management/src/config_builder.rs
```diff
@@ -92,24 +116,24 @@ impl<T: AsRef<Path>> ValidatorBuilder<T> {
...
             libra_global_constants::OPERATOR_ACCOUNT.into(),
-            self.secure_backend(&ns, "validator"),
+            self.secure_backend(&local_ns, "validator"),
         );
 
         let fullnode_network = &mut config.full_node_networks[0];
         let fullnode_network_address = fullnode_network.listen_address.clone();
         fullnode_network.identity = Identity::from_storage(
             libra_global_constants::FULLNODE_NETWORK_KEY.into(),
             libra_global_constants::OPERATOR_ACCOUNT.into(),
```
davidiw, line 126: OWNER_ACCOUNT?

config/management/src/config_builder.rs
```diff
@@ -92,24 +116,24 @@ impl<T: AsRef<Path>> ValidatorBuilder<T> {
         validator_network.identity = Identity::from_storage(
             libra_global_constants::VALIDATOR_NETWORK_KEY.into(),
             libra_global_constants::OPERATOR_ACCOUNT.into(),
```
davidiw, line 118: OWNER_ACCOUNT?

language/tools/vm-genesis/src/lib.rs
```diff
@@ -239,23 +235,44 @@ fn create_and_initialize_testnet_minting(
...
 
-        let auth_key = AuthenticationKey::ed25519(&account_key);
-        let account = auth_key.derived_address();
-        let create_script = transaction_builder::encode_create_validator_account(
-            account,
-            auth_key.prefix().to_vec(),
+        // TODO(joshlind): use the owner_name to derive a unique and deterministic account address
+        // For now, we use a random auth key and account below:
+        let owner_auth_key = AuthenticationKey::random();
+        let owner_account = AccountAddress::random();
```
davidiw: how would you even know how to use this account if it is random? this commit probably won't pass smoke tests.

JoshLind: This is fixed in subsequent commits. Trying to tackle one thing at a time 😄

JoshLind:
Closing this as it's now being handled by a different PR.

PR #4858, opened 2020-07-01
[logging] convert security logging to structured logging

mimoo:
This commit converts all the security logs to use the new structured logging framework.
In addition, it creates a new severity level to segregate security logs.


common/logger/src/security.rs
```diff
@@ -0,0 +1,96 @@
...
+/// helper function to create a security log
+pub fn security_log(name: &'static str) -> StructuredLogEntry {
+    StructuredLogEntry::new_named(&name)
+        // set the category to security
+        .security()
+    // set the error description
+}
+
+/// Security events that are possible
+pub mod security_events {
```
mimoo, line 30: all of these events are copy/pasted from the security-log crate. I'll have a pass at them in another PR.

common/logger/src/struct_log.rs
```diff
@@ -38,30 +38,40 @@ const INITIALIZED: usize = 2;
...
 const SEVERITY_CRITICAL: usize = 1;
 const SEVERITY_WARNING: usize = 2;
+const SEVERITY_SECURITY: usize = 3;
 
 #[derive(Default, Serialize)]
 pub struct StructuredLogEntry {
+    /// log message
     #[serde(skip_serializing_if = "Option::is_none")]
     log: Option<String>,
+    /// description of the log
```
mimoo, line 48: I'm going to need help for some of these fields that I'm not sure I understand =)

common/logger/src/struct_log.rs
```diff
@@ -38,30 +38,40 @@ const INITIALIZED: usize = 2;
 // severity level - lower is worse
 const SEVERITY_CRITICAL: usize = 1;
 const SEVERITY_WARNING: usize = 2;
+const SEVERITY_SECURITY: usize = 3;
```
mimoo, line 41: I added security logs as a severity level after discussion with @andll

andll: // severity level - lower is worse
To keep this ordering, should we rearrange those?
SEVERITY_SECURITY -> 1
SEVERITY_CRITICAL -> 2
SEVERITY_WARNING -> 3

common/logger/src/struct_log.rs
andll: This is a pattern used in info! and similar macro. This is not field otherwise

common/logger/src/struct_log.rs
```diff
@@ -38,30 +38,40 @@ const INITIALIZED: usize = 2;
...
 
 #[derive(Default, Serialize)]
 pub struct StructuredLogEntry {
+    /// log message
     #[serde(skip_serializing_if = "Option::is_none")]
     log: Option<String>,
+    /// description of the log
     #[serde(skip_serializing_if = "Option::is_none")]
     pattern: Option<&'static str>,
+    /// name of the logger emitting the log
```
andll, line 51: name of the event

common/logger/src/struct_log.rs
```diff
@@ -38,30 +38,40 @@ const INITIALIZED: usize = 2;
...
+    /// filename + line (e.g. consensus/src/round_manager.rs:678)
     #[serde(skip_serializing_if = "Option::is_none")]
     location: Option<&'static str>,
+    /// git revision
     #[serde(skip_serializing_if = "Option::is_none")]
     git_rev: Option<&'static str>,
+    /// time of the log
     #[serde(skip_serializing_if = "Option::is_none")]
     timestamp: Option<String>,
+    /// warning or critical
```
andll, line 66: or security :D

consensus/src/round_manager.rs
```diff
@@ -318,10 +317,9 @@ impl RoundManager {
             sync_info
                 .verify(&self.epoch_state().verifier)
                 .map_err(|e| {
-                    security_log(SecurityEvent::InvalidSyncInfoMsg)
-                        .error(&e)
-                        .data(&sync_info)
-                        .log();
+                    send_struct_log!(security_log(security_events::INVALID_SYNC_INFO_MSG)
+                        .data("sync_info", &sync_info)
```
andll, line 321: Btw, if you _want_ you can consider using fields.

Example declaration:
https://github.com/libra/libra/blob/51e6e69c8a8b68e0087fb43c93ab89ad85aa0769/network/src/logging.rs#L25
Usage:
https://github.com/libra/libra/blob/ebecc430e7f78d0968fdcf188555cf3b27b79cfc/network/onchain-discovery/src/client.rs#L131

andll:
Overall this looks good from logging crate perspective, I added some optional comments

@bors-libra delegate+

consensus/src/round_manager.rs
mimoo: I was wondering about that, this is specifically when the same log is used in different places right?

common/logger/src/security.rs
```diff
@@ -0,0 +1,94 @@
...
+    /// Mempool received an invalid network event
+    pub const INVALID_NETWORK_EVENT_MP: &str = "INVALID_NETWORK_EVENT_MP";
+
+    // Consensus
+    // ---------
+
+    /// Consensus received a transaction with an invalid signature
+    pub const INVALID_TRANSACTION_CONSENSUS: &str = "InvalidTransactionConsensus";
+
+    /// Consensus received an invalid vote
```
davidiw, line 50: most of these constants are self-descriptive and if not, they probably should be, no? I don't see the value of the additional comments

mimoo: copy/pasted for now. I'll think about it when I go over each of them.

mimoo:
damn bors...

PR #4860, opened 2020-07-01
[move source language] Fix block inlining

tnowacki:
- Fixed labels for block inlining

## Motivation

- Fix block inlining for #4853 

## Test Plan

- new test (tested with #4853)

### DIFF
--- a/language/move-lang/src/cfgir/inline_blocks.rs
+++ b/language/move-lang/src/cfgir/inline_blocks.rs
@@ -2,7 +2,10 @@
 // SPDX-License-Identifier: Apache-2.0
 
 use super::cfg::{BlockCFG, CFG};
-use crate::hlir::ast::{BasicBlocks, Command_, Label};
+use crate::{
+    cfgir::ast::remap_labels,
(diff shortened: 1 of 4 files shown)


language/move-lang/src/cfgir/inline_blocks.rs
```diff
@@ -85,5 +89,37 @@ fn inline_single_target_blocks(
...
+///   s.t. A is the only one jumping to B
+///
+/// After:
+///   B: block_a; block_b
+fn remap_to_last_target(
+    mut remapping: BTreeMap<Label, Label>,
+    start: Label,
+    blocks: &mut BasicBlocks,
+) {
+    remapping.remove(&start);
```
bob-wilson, line 114: It might be worth a comment here to mention why the start block doesn't need to be relabeled.

tnowacki: It's more that it *can't* be relabeled due to limitations on the current cfg api. I'll add a comment

PR #4861, opened 2020-07-01
[mempool] FN failover

christinacelee:
## Motivation

- update full node failover mechanism for mempool based on new networking paradigm
- remove outdated k-policy

## Test Plan

New/existing tests


mempool/src/shared_mempool/peer_manager.rs
```diff
@@ -63,15 +71,33 @@ impl PeerManager {
...
-                .entry(peer)
-                .or_insert(PeerSyncState {
-                    timeline_id: 0,
-                    is_alive: true,
-                    broadcast_info: BroadcastInfo::new(),
-                })
-                .is_alive = true;
+            if peer.network_id() == NetworkId::Validator {
+                // For a validator network, resume broadcasting from previous state
+                // we can afford to be not re-broadcast here since the transaction is already in a validator
```
kphfb, line 76: *we can afford to not re-broadcast

mempool/src/tests/shared_mempool_test.rs
```diff
@@ -726,36 +726,33 @@ fn test_state_sync_events_committed_txns() {
     assert_eq!(timeline.get(0).unwrap().1, kept_txn);
 }
 
-// primary_peers = k, fallbacks > 0
 #[test]
-fn test_k_policy_broadcast_no_fallback() {
-    // all nodes
+fn test_fn_failover() {
```
kphfb, line 730: Should we add a test of what happens if there are no upstream or peers available?

mempool/src/shared_mempool/peer_manager.rs
```diff
@@ -84,6 +110,67 @@ impl PeerManager {
         {
             state.is_alive = false;
         }
+        self.update_failover();
+    }
+
+    fn update_failover(&self) {
```
kphfb, line 116: Think it would be good to add a few more comments to this function.  At least a high level comment above it that describes what the overall purpose of the function is and maybe a couple on the loops to describe what they are doing to make it more easily readable

PR #4866, opened 2020-07-01
Builder cleanup

agc-quadra:
## Motivation

Incorporate the network-builder structure into main_node and apply a number of cleanups and simplifications to network_builder that are enabled by having the distinct builders for the various sub-components.  

yes
## Test Plan

don't break existing tests

## Related PRs

### DIFF
--- a/libra-node/src/main_node.rs
+++ b/libra-node/src/main_node.rs
@@ -165,7 +165,7 @@ pub fn setup_environment(node_config: &mut NodeConfig) -> LibraHandle {
         // Start the network and cache the runtime so it does not go out of scope.
         // TODO:  move all 'start' commands to a second phase at the end of setup_environment.  Target is to have one pass to wire the pieces together and a second pass to start processing in an appropriate order.
         let peer_id = network_builder.peer_id();
-        let _listen_addr = network_builder.build();
+        network_builder.build();
         network_runtimes.push(runtime);
         debug!("Network started for peer_id: {}", peer_id);
     }
--- a/libra-node/src/main_node.rs
+++ b/libra-node/src/main_node.rs
@@ -122,6 +122,8 @@ pub fn setup_environment(node_config: &mut NodeConfig) -> LibraHandle {
         network_configs.push((RoleType::Validator, network_config));
     }
 
+    let mut network_builders = Vec::new();
+
     // Instantiate every network and collect the requisite endpoints for state_sync, mempool, and consensus.
     for (role, network_config) in network_configs {
         // Perform common instantiation steps
@@ -162,20 +164,33 @@ pub fn setup_environment(node_config: &mut NodeConfig) -> LibraHandle {
 
         reconfig_subscriptions.append(network_builder.reconfig_subscriptions());
 
-        // Start the network and cache the runtime so it does not go out of scope.
-        // TODO:  move all 'start' commands to a second phase at the end of setup_environment.  Target is to have one pass to wire the pieces together and a second pass to start processing in an appropriate order.
+        network_builders.push(network_builder);
+    }
+
+    // Build the configured networks.
+    for network_builder in &mut network_builders {
+        debug!(
+            "Creating runtime for {:?}",
+            network_builder.network_context()
+        );
         let runtime = Builder::new()
             .thread_name("network-")
             .threaded_scheduler()
             .enable_all()
             .build()
             .expect("Failed to start runtime. Won't be able to start networking.");
-        network_builder.build(runtime.handle().clone()).start();
+        network_builder.build(runtime.handle().clone());
         network_runtimes.push(runtime);
         debug!(
-            "Network started for peer_id: {:?}",
+            "Network Built for network_context: {:?}",
             network_builder.network_context()
         );
+    }
+
+    // Start the configured networks.
+    // TODO:  Collect all component starts at the end of this function
+    for network_builder in &mut network_builders {
+        network_builder.start();
     }
 
     // TODO set up on-chain discovery network based on UpstreamConfig.fallback_network
--- a/libra-node/src/main_node.rs
+++ b/libra-node/src/main_node.rs
@@ -170,7 +170,7 @@ pub fn setup_environment(node_config: &mut NodeConfig) -> LibraHandle {
             .enable_all()
             .build()
             .expect("Failed to start runtime. Won't be able to start networking.");
-        network_builder.build(runtime.handle().clone());
+        network_builder.build(runtime.handle().clone()).start();
         network_runtimes.push(runtime);
         debug!(
             "Network started for peer_id: {:?}",
--- a/libra-node/src/main_node.rs
+++ b/libra-node/src/main_node.rs
@@ -22,7 +22,7 @@ use state_synchronizer::StateSynchronizer;
 use std::{boxed::Box, net::ToSocketAddrs, sync::Arc, thread, time::Instant};
 use storage_interface::DbReaderWriter;
 use storage_service::start_storage_service_with_db;
-use tokio::runtime::Runtime;
+use tokio::runtime::{Builder, Runtime};
 
 const AC_SMP_CHANNEL_BUFFER_SIZE: usize = 1_024;
 const INTRA_NODE_CHANNEL_BUFFER_SIZE: usize = 1;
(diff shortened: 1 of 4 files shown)


agc-quadra:
@gregnazario @davidiw 

This is the 'end' for the core network-builder cleanup.  There are a couple of directions that I can go from here:

1.  address the network_endpoint_config() trait/struct/lack thereof (@davidiw )
2.  attempt to address the builder trait (@gregnazario )
3.  work towards a builder for the other top level components (consensus, statesync, mempool) in order to get to structural consistency across the codebase and reap the full rewards of splitting create/build/start in main_node
4.  Push the "TODO:  encode this in a config" items into a config and out of the default constants values.
5.  Finish the removal of onchain discovery --- it is currently being kept alive by a test
6.  Work on updating state-synchronizer/integration_test and network/builder/dummy to use the network_builder::create interface (and implicitly a networkconfig) rather than manually configuring the networkbuilder.

My instincts say that 4/5/6 are the easiest pieces to tackle, that 1/2 depend to a certain extent on 3, and that 3 is most likely to induce significant bikeshed.

Your thoughts/priorities/interests are much appreciated.

davidiw:
Will try to review this -- much -- later today.

libra-node/src/main_node.rs
```diff
@@ -162,12 +164,33 @@ pub fn setup_environment(node_config: &mut NodeConfig) -> LibraHandle {
...
+            .threaded_scheduler()
+            .enable_all()
+            .build()
+            .expect("Failed to start runtime. Won't be able to start networking.");
+        network_builder.build(runtime.handle().clone());
         network_runtimes.push(runtime);
-        debug!("Network started for peer_id: {}", peer_id);
+        debug!(
+            "Network Built for network_context: {:?}",
+            network_builder.network_context()
```
davidiw, line 186: seems weird we never implemented Display for NetworkContext....

libra-node/src/main_node.rs
```diff
@@ -162,12 +164,33 @@ pub fn setup_environment(node_config: &mut NodeConfig) -> LibraHandle {
...
+        );
+    }
+
+    // Start the configured networks.
+    // TODO:  Collect all component starts at the end of this function
+    for network_builder in &mut network_builders {
```
davidiw, line 192: any reason to not just merge these two? the networks don't seem to have any overlap.. and that way we can unify the log into built -> start, right now we have no long for starting... so maybe something weird happened between at starting

network/builder/src/builder.rs
```diff
@@ -43,75 +43,65 @@ use std::{
...
-        executor: Handle,
         chain_id: ChainId,
-        network_id: NetworkId,
-        role: RoleType,
-        peer_id: PeerId,
+        trusted_peers: Arc<RwLock<HashMap<PeerId, x25519::PublicKey>>>,
+        network_context: Arc<NetworkContext>,
         listen_address: NetworkAddress,
         authentication_mode: AuthenticationMode,
     ) -> NetworkBuilder {
```
davidiw, line 84: nit -> Self when inside a an impl for a struct

network/builder/src/builder.rs
```diff
@@ -144,19 +128,32 @@ impl NetworkBuilder {
...
             config.network_id.clone(),
             role,
             peer_id,
+        ));
+
+        let trusted_peers = if config.mutual_authentication && role == RoleType::FullNode {
+            Arc::new(RwLock::new(config.network_peers.clone()))
+        } else {
+            Arc::new(RwLock::new(HashMap::new()))
+        };
```
davidiw, line 141: I guess this looks a little out of date....

frankly, our support for mutual authentication except for the validator network is not tested or documented... I am out until next Thursday, flagging @phlip9 / @gregnazario to discuss internally if we want to keep this or kill this, since it is actually causing us to push around "tech debt"

network/builder/src/builder.rs
```diff
@@ -166,42 +163,39 @@ impl NetworkBuilder {
...
-                network_builder
-                    .seed_peers(seed_peers)
-                    .add_connectivity_manager();
-            }
-        }
+        // Setup the ConnectivityManager.  Do not setup a ConnectivityManager iff
+        //   1.  no mutual authentication
+        //   2.  DiscoveryMethod::Gossip
+        //   3.  no seed_peers
+        // TODO:  Why not add ConnectivityManager always?
```
davidiw, line 175: tech debt, it will kill connections if they aren't in the trusted peers test... which makes no sense for server-only auth...

state-synchronizer/src/tests/integration_tests.rs
```diff
@@ -296,24 +297,39 @@ impl SynchronizerEnv {
...
+                    trusted_peers,
+                    constants::MAX_FULLNODE_CONNECTIONS,
+                    constants::MAX_CONNECTION_DELAY_MS,
+                    constants::CONNECTIVITY_CHECK_INTERNAL_MS,
+                    constants::NETWORK_CHANNEL_SIZE,
+                )
+                .add_gossip_discovery(
+                    GossipConfig {
+                        advertised_address: addr,
+                        discovery_interval_ms: constants::DISCOVERY_INTERVAL_MS,
```
davidiw, line 324: this kinda interesting that they aren't uniform... why reveal the config outside the internals of the builder?

libra-node/src/main_node.rs
agc-quadra: Other components lack support for separating create/build/start.   It is technically meaningless to separate the pieces for network and not the others.

Stylistically, performing the (meaningless) separation now at least points to the direction and may encourage people to think in that way moving forward).

network/builder/src/builder.rs
agc-quadra: done.

state-synchronizer/src/tests/integration_tests.rs
agc-quadra: +1

One of the big outstanding TODOS is to make add_gossip *private* and for this test (and for dummy.rs) to use a NodeConfig and ::create to instantiate the builder rather than reaching inside to do it manually.

(those manual reach-ins are actually pretty bad, and encourage some rather unfortunate practices and corner case support)

network/builder/src/builder.rs
agc-quadra: This and the following both expose some unfortunate tech debt pieces that I'm continuing to push around.

At least the tech debt is more visible now.

network/builder/src/builder.rs
agc-quadra: This seems very .... bad to me.  I think we should fix that behavior so connectivitymanager can be added always.

agc-quadra:
Thanks for the comments.

And for picking out the now glaringly obvious tech debt sore thumbs   At least they're glaringly obvious now.

The behavior of server_only authentication is the biggest sore thumb from where I sit.   It's pretty inconceivable to me that seed_peers being empty or not controls whether connectivitymanager is added.

network/builder/src/builder.rs
gregnazario: I think it comes more to the problem of the general networking model we're using at this point, where we expect it to be a p2p network, and that the connectivity manager only connects to trusted versions of those.

I think we need to look forward into refactoring this to allow for that situation, but it's not made yet for this.

agc-quadra: I would expect that ConnectivityManager is (a) reconnecting to the approved peers and (b) only accepting lincoming connections from approved peers.

I would hope that the difference in the scenario you hint at is simply that the list of approvied peers changes.  But I dont have enough context on the intent.

It's also extraordinarily odd to me that the existence of seed_peers is the deciding factor as to whether or not to instantiate the connectivitymanager.

network/builder/src/dummy.rs
```diff
@@ -136,13 +141,16 @@ pub fn setup_network() -> DummyNetwork {
 
     let authentication_mode = AuthenticationMode::Mutual(listener_identity_private_key);
 
+    let network_context = Arc::new(NetworkContext::new(
+        network_id.clone(),
+        RoleType::Validator,
+        listener_peer_id,
+    ));
     // Set up the listener network
```
davidiw, line 150: comment should be above the assignment.

agc-quadra: Put a space between the preceding lines and the comment.  I actually did intend to associate it with the builder and not the network_context

agc-quadra:
@davidiw  One more round of conflict resolution.

PR #4867, opened 2020-07-01
[consensus] ignore stale messages instead of reporting errors

zekun000:
## Motivation

It's observed ~20% error rate because every round there's f stale votes
because QC already formed, it's better not to report errors on those messages.

Y

## Test Plan

Existing tests.

## Related PRs

### DIFF
--- a/consensus/src/round_manager.rs
+++ b/consensus/src/round_manager.rs
@@ -16,7 +16,7 @@ use crate::{
     state_replication::{StateComputer, TxnManager},
     util::time_service::duration_since_epoch,
 };
-use anyhow::{ensure, Context, Result};
+use anyhow::{bail, ensure, Context, Result};
 use consensus_types::{
     block::Block,
     block_retrieval::{BlockRetrievalResponse, BlockRetrievalStatus},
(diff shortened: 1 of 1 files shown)


consensus/src/round_manager.rs
```diff
@@ -539,17 +543,21 @@ impl RoundManager {
...
-        self.ensure_round_and_sync_up(
-            vote_msg.vote().vote_data().proposed().round(),
-            vote_msg.sync_info(),
-            vote_msg.vote().author(),
-            true,
-        )
-        .await
-        .context("[RoundManager] Stop processing vote")?;
-        self.process_vote(vote_msg.vote())
+        if self
```
aching, line 546: I agree that reporting errors for the f votes that occur after 2f+1 doesn't make sense.  

We shouldn't expect this for proposals though right?  I.e. For debugging, would it be useful to keep some error message that we received an invalid stale proposal?

zekun000: it looks similar to me, if one node is stale it'd broadcast stale proposals. it's just not as expected as the f votes

aching: I'm a little worried about losing hints about misbehaving nodes to remove these expected error messages - more so in the proposal case instead of the voting case (since its expected).  

I.e. How would we determine that a node is accidentally flooding others with bad proposals or votes?

zekun000: a misbehaving node can flood others with the exact votes that we ignore too, I don't think the added value of detecting stale messages worth specializing. if things goes wrong, there's enough logging to debug.

aching: Which logs are you referring to specifically for detecting floods?

zekun000:
updated with specializing proposals

consensus/src/round_manager.rs
```diff
@@ -344,21 +355,18 @@ impl RoundManager {
...
-    ) -> anyhow::Result<()> {
-        ensure!(
-            message_round >= self.round_state.current_round(),
-            "round {} is stale than local {}",
-            message_round,
-            self.round_state.current_round()
-        );
+    ) -> anyhow::Result<bool> {
+        if message_round < self.round_state.current_round() {
+            return Ok(false);
```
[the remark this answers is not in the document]
zekun000: it's contextual error before, do  you mean turn the anyhow error into thiserror? I'm too lazy for that, although it seems useful in aggregating and reporting

PR #4868, opened 2020-07-01
[tf] Specify chain_id

mgorven:
Set `chain_id` to "ecs-[workspace]-[4 hex chars]". Add base templates to
pass to config-builder in order to do this. The random ID will change if
the workspace is destroyed, or can be manually changed with `terraform
taint random_id.chain_id`.

## Test Plan

Deployed to my workspace.

### DIFF
--- a/terraform/validators.tf
+++ b/terraform/validators.tf
@@ -138,16 +138,16 @@ data "template_file" "user_data" {
 }
 
 locals {
-  image_repo                 = var.image_repo
-  image_version              = substr(var.image_tag, 0, 6) == "sha256" ? "@${var.image_tag}" : ":${var.image_tag}"
-  override_image_versions    = [for img in var.override_image_tags : substr(img, 0, 6) == "sha256" ? "@${img}" : ":${img}"]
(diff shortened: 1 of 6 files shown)


terraform/validators.tf
```diff
@@ -202,44 +202,57 @@ locals {
   validator_command      = var.log_to_file || var.enable_logstash ? jsonencode(["bash", "-c", "/docker-run-dynamic.sh >> ${var.log_path} 2>&1"]) : ""
   aws_elasticsearch_host = var.enable_logstash ? join(",", aws_elasticsearch_domain.logging.*.endpoint) : ""
   logstash_config        = "input { file { path => '${var.structlog_path}'\\n codec => 'json'\\n}}\\n filter {  json {  \\nsource => 'message'\\n}}\\n output {  amazon_es { \\nhosts => ['https://${local.aws_elasticsearch_host}']\\nregion => 'us-west-2'\\nindex => 'validator-logs-%%{+YYYY.MM.dd}'\\n}}"
+  chain_id               = "tf-testnet-${terraform.workspace}"
```
gregnazario, line 205: I'm assuming this only applies to testnet?  I don't really have a lot of context on how the terraform configurations are set up

mgorven: Anything deployed with these Terraform configs. The persistent ones are prod and devnet.

PR #4869, opened 2020-07-01
[json-rpc] add sender/recipient (resp.) to SentPaymentEvent/ReceivedPaymentEvent (resp.)

sblackshear:
If you have an `EventHandle` for a `SentPayment` or `ReceivedPaymentEvent`, you know the sender/recipient, but it's not obvious how to find it: extract from the `EventKey`. This PR does this extraction and adds it to the json-rpc event payload for the convenience of the client.

### DIFF
--- a/json-rpc/types/src/views.rs
+++ b/json-rpc/types/src/views.rs
@@ -136,12 +136,14 @@ pub enum EventDataView {
     ReceivedPayment {
         amount: AmountView,
         sender: BytesView,
+        receiver: BytesView,
         metadata: BytesView,
     },
     #[serde(rename = "sentpayment")]
     SentPayment {
         amount: AmountView,
         receiver: BytesView,
+        sender: BytesView,
         metadata: BytesView,
     },
     #[serde(rename = "upgrade")]
@@ -171,6 +173,7 @@ impl From<(u64, ContractEvent)> for EventView {
                 Ok(EventDataView::ReceivedPayment {
                     amount: amount_view,
                     sender: BytesView::from(received_event.sender().as_ref()),
+                    receiver: BytesView::from(&event.key().get_creator_address().to_vec()),
                     metadata: BytesView::from(received_event.metadata()),
                 })
             } else {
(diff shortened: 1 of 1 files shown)


thefallentree:
we need to include the original key string as well, since the API for looking up events require full string to query events.

I'm not sure why does rosetta needs to extract creator address from event_key, but I think an more natural api is to just include more field in the response to give what the client needs . (and more flat the better, less breaking change)

thefallentree:
To be extra clear, what I suggest is,   add an burn_account_address = "xxxxxx" addtion to burn_events_key ="xxx" directly to  get_currencys api, instead of modifying this. 

obviously, more redundant fields,  but the cost is minimum,  this will also have better chance of not-breaking everything if we decide to move the events to some other account in the future.

sblackshear:
Good call, fixed as recommended

thefallentree:
If this is about adding sender/receiver field to these events now , could you fix the commit the message and PR message? 

cc @bob-wilson  (I suggested this to him back in Nov/Dec 2019, glad to see we finally agreed to do it :-D)

bob-wilson:
From what I can see, this is not touching the Events at all. It's just adding some fields in the json-rpc info to make it easier to find the sender/receiver. If that's what you were suggesting last fall, then I guess we just didn't understand.

But, I agree that the title and description of this PR should be updated before it lands.

thefallentree:
> From what I can see, this is not touching the Events at all. It's just adding some fields in the json-rpc info to make it easier to find the sender/receiver. If that's what you were suggesting last fall, then I guess we just didn't understand.
> 
> But, I agree that the title and description of this PR should be updated before it lands.

well, at the time JSONRPC doesn't even exists, we were using ContractEvent struct,   for space saving purpose (which I am still skeptical about), it seems now we choose to extract that information from event key, not an actual field in the struct.

But regardless, the need is to get the sender address.

thefallentree:
@sblackshear  please also update https://github.com/libra/libra/blob/master/json-rpc/json-rpc-spec.md

sblackshear:
@thefallentree: is this generated from something, or should I manually edit the spec file?

PR #4874, opened 2020-07-01
[cluster_test] Introduce named suites

andll:
* Introducing `--suite <NAME>` command to run named suite
* They are still hardcoded in `suite.rs` for now
* `--run-ci-suite` and `--perf-run` are deprecated and replaced with `--suite perf` and `--suite pre_release`
* New suite `land_blocking` is introduced

### DIFF
--- a/testsuite/cluster-test/src/suite.rs
+++ b/testsuite/cluster-test/src/suite.rs
@@ -12,13 +12,14 @@ use crate::{
         RecoveryTimeParams,
     },
 };
+use anyhow::{format_err, Result};
 
 pub struct ExperimentSuite {
     pub experiments: Vec<Box<dyn Experiment>>,
 }
 
 impl ExperimentSuite {
-    pub fn new_pre_release(cluster: &Cluster) -> Self {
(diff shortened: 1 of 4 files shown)


testsuite/cluster-test/src/suite.rs
```diff
@@ -68,4 +69,21 @@ impl ExperimentSuite {
...
+    fn new_land_blocking_suite(cluster: &Cluster) -> Self {
+        let mut experiments: Vec<Box<dyn Experiment>> = vec![];
+        experiments.push(Box::new(
+            PerformanceBenchmarkParams::new_nodes_down(0).build(cluster),
+        ));
+        Self { experiments }
+    }
+
+    pub fn new_by_name(cluster: &Cluster, name: &str) -> Result<Self> {
+        match name {
```
snakech, line 82: maybe make suit name as const?

andll: My personal judgement here is to make strings const if they are used in multiple places (which is not the case here)

PR #4875, opened 2020-07-02
[network] consolidated usage of ConnectionNotif

gregnazario:
ConnectionNotification was used in multiple places, and had been copied
many times around, which made it hard to add a field to it.  I've now
consolidated many of the places in tests and elsewhere that had repeated
behavior, to make it easier to add fields.

### DIFF
--- a/network/src/connectivity_manager/test.rs
+++ b/network/src/connectivity_manager/test.rs
@@ -109,6 +109,31 @@ async fn get_dial_queue_size(conn_mgr_reqs_tx: &mut channel::Sender<Connectivity
     queue_size_rx.await.unwrap()
 }
 
+async fn send_new_peer_await_delivery(
+    connection_notifs_tx: &mut conn_notifs_channel::Sender,
+    peer_id: PeerId,
+    notif_peer_id: PeerId,
+    address: NetworkAddress,
+) {
+    let notif = peer_manager::ConnectionNotification::NewPeer(
+        notif_peer_id,
+        address,
+        NetworkContext::mock(),
+    );
+    send_notification_await_delivery(connection_notifs_tx, peer_id, notif).await;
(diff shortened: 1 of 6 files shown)


network/src/connectivity_manager/test.rs
```diff
@@ -301,14 +318,12 @@ fn connect_to_seeds_on_startup() {
 
         // We expect the peer which changed its address to also disconnect.
         info!("Sending lost peer notification for seed peer at old address");
-        send_notification_await_delivery(
+        send_lost_peer_await_delivery(
```
davidiw, line 321: I wonder if the "await_delivery" is kind of implicit in the fact that this is an awaitable / async?

network/src/peer_manager/conn_notifs_channel.rs
```diff
@@ -28,84 +28,54 @@ mod test {
...
 
+    fn send_new_peer(sender: &mut Sender, peer_id: PeerId) {
+        sender
+            .push(
+                peer_id,
+                ConnectionNotification::NewPeer(
+                    peer_id,
+                    NetworkAddress::mock(),
+                    NetworkContext::mock(),
+                ),
```
davidiw, line 39: if you were to move this to another line `let cn = ...` followed by the sender.push, I bet this would be a lot more compact, same in a few other places.

gregnazario: I'll see if I can do that!

Basically this stuff was copied so many places this should help for later

network/src/connectivity_manager/test.rs
gregnazario: Yeah, I wasn't a fan of the "await_delivery", but I kept it the way it was

PR #4877, opened 2020-07-02
[libra framework] Add account limits checks to VASP accounts

### LINKED ISSUE #4823: Account limits for VASPs
VASPs need to have account limits
* The parent VASP account holds a limits resource for the union of all currencies held by its child accounts.
* All txns  with the child accounts/parent account are counted in this. Intra-VASP transfers are not counted.

tzakian:
VASPs need to track account limits (inflow + outflow) in a time window, and allow setting of maximum balances. This is set per currency.

This PR implements the logic for this. In particular:
* Intra-vasp transfers are not counted towards inflow/outflow/max balance for any currency.
* VASP->any  we check and update the sending account's account limits
* any->VASP we check and update the receiving account's account limits

Even if a VASP is "unrestricted" we still track the balance, this is because we cannot really retroactively assign the balance for all of the VASP's accounts. Max holding is not reset with the time window.

## Implementation details
* Every VASP has a `LimitsDefintion<Currency>` and `Window<Currency>` for each currency it or any of its children can hold.
* A child account _cannot_ add a currency to its account without the parent VASP first publishing a `LimitsDefintion` and `Window` for that currency. In particular this means that the parent VASP account will hold a `LimitsDefintion/Window` for each currency any its children can hold.
* Only the parent VASP can publish limits/windows for a currency. We start out a parent VASP account with all of the available currencies. If one is added later, the parent account will need to add a limits/window for it before any of its accounts can have a balance in that currency.

closes #4823


moeziniaf:
We have our sprint ending tonight, I'll take a look once that's over :)

language/stdlib/modules/AccountLimits.move
```diff
@@ -79,89 +96,72 @@ module AccountLimits {
...
-        borrow_global_mut<LimitsDefinition>(limits_addr).is_certified = true;
-    }
-
-    // Decertify the limits_definition published under the account at
-    // `limits_addr`.
-    public fun decertify_limits_definition(_: &Capability<TreasuryComplianceRole>, limits_addr: address)
-    acquires LimitsDefinition {
-        borrow_global_mut<LimitsDefinition>(limits_addr).is_certified = false;
+        let limits_def = borrow_global_mut<LimitsDefinition<CoinType>>(limit_address);
+        if (new_max_inflow > 0) limits_def.max_inflow = new_max_inflow;
```
sblackshear, line 162: Nit: can we use curly braces when the conditional is used as an imperative statement? I thought our style guidelines omitting the curly braces when conditonals were used as expressions, but not otherwise.

language/stdlib/modules/AccountLimits.move
```diff
@@ -185,63 +186,74 @@ module AccountLimits {
...
     fun can_receive<CoinType>(
         amount: u64,
-        receiving: &mut Window,
+        receiving: &mut Window<CoinType>,
     ): bool acquires LimitsDefinition {
-        let limits_definition = borrow_global_mut<LimitsDefinition>(receiving.limits_definition);
-        // If the limits ares unrestricted then no more work needs to be done
-        if (is_unrestricted(limits_definition)) return true;
+        let limits_definition = borrow_global_mut<LimitsDefinition<CoinType>>(receiving.limit_address);
+        // If the limits ares unrestricted then only track the balance
```
sblackshear, line 192:
```suggestion
        // If the limits are unrestricted then only track the balance
```

sblackshear: Could also be useful to explain why we need to track the balance in this case (is it because balance holding fees are tracked + assessed independently of account limits)?

language/stdlib/modules/AccountLimits.move
```diff
@@ -185,63 +186,74 @@ module AccountLimits {
...
-        sending: &mut Window,
+        sending: &mut Window<CoinType>,
     ): bool acquires LimitsDefinition {
-        let limits_definition = borrow_global_mut<LimitsDefinition>(sending.limits_definition);
+        let limits_definition = borrow_global_mut<LimitsDefinition<CoinType>>(sending.limit_address);
         // If the limits are unrestricted then no more work is required
-        if (is_unrestricted(limits_definition)) return true;
+        if (is_unrestricted(limits_definition)) {
+            sending.tracked_balance = if (amount >= sending.tracked_balance) 0
+                                      else sending.tracked_balance  - amount;
```
sblackshear, line 222:
```suggestion
                                      else sending.tracked_balance - amount;
```

language/stdlib/modules/AccountLimits.move
```diff
@@ -24,36 +22,55 @@ module AccountLimits {
...
         tracked_balance: u64,
         // address storing the LimitsDefinition resource that governs this window
-        limits_definition: address,
+        limit_address: address,
     }
 
+    const NOT_GENESIS: u64 = 0;
+    const INVALID_INITIALIZATION_ADDRESS: u64 = 1;
+    /// 24 hours in microseconds
+    const TIME_WINDOW: u64 = 86400000000;
```
sblackshear, line 54: Nit: I think calling this `ONE_DAY` will make the code slightly more readable (I was looking at a function that uses this and thought it might be U64_MAX because of the name).

language/stdlib/modules/LibraAccount.move
```diff
@@ -157,6 +158,14 @@ module LibraAccount {
         );
     }
 
+    fun should_track_limits(scrutinee: address, other_account: address): bool {
```
sblackshear, line 161: I think it might be clearer for if we avoid introducing the concept of "scrutinee", which is context-dependent and thus potentially confusing for callers. If we just use `payer` and `payee` here, that will make the body of this function a bit bigger, but it will be easier for callers and will map very directly to the clear description you lay out in the PR summary:

```
VASP->any we check and update the sending account's account limits
any->VASP we check and update the receiving account's account limits
```

language/stdlib/modules/AccountLimits.move
```diff
@@ -24,36 +22,55 @@ module AccountLimits {
...
+        // The inflow during this time window
+        window_outflow: u64,
         // The balance that this account has held during this time period.
         tracked_balance: u64,
         // address storing the LimitsDefinition resource that governs this window
-        limits_definition: address,
+        limit_address: address,
     }
 
+    const NOT_GENESIS: u64 = 0;
```
wrwg, line 51: Let's move these two right away in a module `Errors` or similar? Then we can incrementally extend it.

language/stdlib/modules/AccountLimits.move
```diff
@@ -24,36 +22,55 @@ module AccountLimits {
...
+    const INVALID_INITIALIZATION_ADDRESS: u64 = 1;
+    /// 24 hours in microseconds
+    const TIME_WINDOW: u64 = 86400000000;
+    const U64_MAX: u64 = 18446744073709551615;
```
wrwg, line 55: Probably U64_MAX too into something like `Limits` or `NumericLimits`? We can then incrementally replace occurrences of the literal.

language/stdlib/modules/LibraAccount.move
```diff
@@ -562,6 +574,13 @@ module LibraAccount {
         );
         Event::publish_generator(&new_account);
         add_currencies_for_account<Token>(&new_account, add_all_currencies);
+        // VASPs start off by default with the full range of currencies. If a
```
sblackshear, line 577: Yeesh, this parent/child limits problem is a nasty one...

Does this mean: VASPs start off with `LimitsDefinition`s for every currency even if they don't have a `Balance` resource for that currency? That feels a bit surprising to me; it would be nice if we can ensure that "balance exists in parent or child <=> limits exists in parent". I think we can do this by gating these calls with `add_all_currencies` (but not sure).

language/stdlib/modules/VASP.move
```diff
@@ -37,6 +40,17 @@ module VASP {
     /// A resource that represents a child account of the parent VASP account at `parent_vasp_addr`
     resource struct ChildVASP { parent_vasp_addr: address }
 
+    /// A singleton resource allowing this module to publish limits definitions and accounting windows
+    resource struct VASPOperationsResource { limits_cap: CallingCapability }
+
+    const NOT_A_REGISTERED_CURRENCY: u64 = 0;
+
+    public fun initialize(association: &signer, cap: &Capability<LibraRootRole>) {
```
sblackshear, line 48: Add `is_genesis` check?

language/stdlib/modules/LibraAccount.move
```diff
@@ -562,6 +574,13 @@ module LibraAccount {
...
+        // new currency is added to the system they will need to call
+        // `VASP::try_allow_currency<NewCurrency>(parent_account)` in order to
+        // allow any of their child VASP accounts to hold a balance of `NewCurrency`.
+        VASP::try_allow_currency<LBR>(&new_account);
```
sblackshear, line 581: Could assert the return value of `try_allow_currency` for extra assurance? Otherwise, these checks could be erroneously moved before `publish_parent_vasp_credential` and nothing would crash (I think).

language/stdlib/modules/AccountLimits.move
```diff
@@ -24,36 +22,55 @@ module AccountLimits {
...
+
     // Grant a capability to call this module. This does not necessarily
     // need to be a unique capability.
     public fun grant_calling_capability(_: &Capability<LibraRootRole>): CallingCapability {
+        assert(LibraTimestamp::is_genesis(), NOT_GENESIS);
         CallingCapability{}
     }
 
+    public fun initialize(account: &signer, _: &Capability<TreasuryComplianceRole>) {
```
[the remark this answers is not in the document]
tzakian: Good point. Let me take a look this and see if I can do something here :)

language/stdlib/modules/AccountLimits.move
tzakian: good point! I'll update.

language/stdlib/modules/AccountLimits.move
tzakian: Will do. It's more so that we can track the balance of the entire entity. If we didn't track this from the start, there would be no easy way to impose a balance limit later on since we wouldn't know (on chain) how much is held across all of the entity's accounts.

language/stdlib/modules/LibraAccount.move
tzakian: Yea this is the one nasty part. 

We could have balance exists in VASP <=> limits in parent VASP. However, this means that if the parent VASP doesn't already have a published limits for a currency none of its children can have a balance in those currencies until it publishes it. e.g.
```
txn1: create_parent_vasp<Coin1>(...., add_all_currencies = false); ~> EXECUTED
txn2: create_child_vasp<Coin1>(parent,..., add_all_currencies = false); ~> EXECUTED
txn3: create_child_vasp<Coin2>(parent, ..., add_call_currencies = fasle); ~> ABORTED (since parent doesn't have limits for Coin2)
```

I'll take a look at this some more, hopefully this can be made cleaner without breaking things...

language/stdlib/modules/LibraAccount.move
tzakian: ah! good point. I think I can remove these if we say that a parent vasp needs to explicitly allow a currency before it can be used.

tzakian:
@wrwg sounds good. I'll try and get this in before lunch if possible

language/stdlib/modules/AccountLimits.move
tzakian: Sadly the source language doesn't support public constants yet so I have to keep these local for now :'(

language/stdlib/modules/LibraAccount.move
sblackshear: I guess one could view "child can't have a balance in T until parent approves it" as a feature to prevent a child from (e.g.) accidentally violating capital controls... But this is nasty enough that there doesn't seem like a right answer, happy  with any approach you choose.

language/stdlib/modules/AccountLimits.move
```diff
@@ -1,261 +1,264 @@
...
-        let limits_definition = borrow_global_mut<LimitsDefinition>(receiving.limits_definition);
-        // If the limits ares unrestricted then no more work needs to be done
-        if (is_unrestricted(limits_definition)) return true;
+        let limits_definition = borrow_global_mut<LimitsDefinition<CoinType>>(receiving.limit_address);
+        // If the limits are unrestricted then only track the balance. We track
+        // the balance even if the limits are "unlimited" since otherwise there
+        // is no way to know the combined balance of an entity on-chain, this
+        // would then preclude setting balance holding limits on an entity at a future date.
+        if (is_unrestricted(limits_definition)) {
+            /*receiving.tracked_balance = receiving.tracked_balance + amount;*/
```
tzakian, line 197: This and the other line in `can_withdraw` are causing a serious perf regression; on the order of about 230TPS. Not sure what we should do about it... any changing of the parent VASP blob is going to cause this issue.

moeziniaf: How did you get a regression breakdown by method?

PR #4878, opened 2020-07-02
[logging] security logging for invalid transactions received from peer

mimoo:
This commit adds a security log to detect invalid transactions that were forwarded by other peers (via mempool)

### DIFF
--- a/common/logger/src/security.rs
+++ b/common/logger/src/security.rs
@@ -26,16 +26,10 @@ pub fn security_log(name: &'static str) -> StructuredLogEntry {
 
 /// Security events that are possible
 pub mod security_events {
-    // JSON-RPC server
-    // ---------------
-
-    /// Admission Control received a transaction with an invalid signature
-    pub const INVALID_TRANSACTION_AC: &str = "InvalidTransactionAC";
-
     // Mempool
     // -------
 
-    /// Mempool received a transaction with an invalid signature
+    /// Mempool received a transaction from another peer with an invalid signature
     pub const INVALID_TRANSACTION_MP: &str = "InvalidTransactionMP";
 
     /// Mempool received an invalid network event
--- a/mempool/src/shared_mempool/tasks.rs
+++ b/mempool/src/shared_mempool/tasks.rs
@@ -236,8 +236,29 @@ pub(crate) async fn process_transaction_broadcast<V>(
 ) where
     V: TransactionValidation,
 {
-    let results = process_incoming_transactions(&smp, transactions, timeline_state).await;
+    // process transactions and log the result
+    let results = process_incoming_transactions(&smp, transactions.clone(), timeline_state).await;
     log_txn_process_results(&results, Some(peer.peer_id()));
+
+    // we assume that the results contains a vec of the same length of transactions
+    debug_assert_eq!(results.len(), transactions.len());
+
+    // security log all the failed transactions
+    let failed_transactions = results
+        .iter()
+        // create iterator of tuple (VM status, transaction)
+        .map(|(_, maybe_vm_status)| maybe_vm_status)
+        .zip(transactions.iter())
+        // VM status will be set if the transaction failed
+        .filter(|(maybe_vm_status, _)| maybe_vm_status.is_some());
+
+    for (maybe_vm_status, failed_transaction) in failed_transactions {
+        send_struct_log!(security_log(security_events::INVALID_TRANSACTION_MP)
+            .data("failed_transaction", &failed_transaction)
+            .data("vm_status", &maybe_vm_status)
+            .data("from_peer", &peer));
+    }
+
     // send back ACK
     let ack_response = gen_ack_response(request_id, results);
     let mut network_sender = smp


common/logger/src/security.rs
```diff
@@ -26,16 +26,10 @@ pub fn security_log(name: &'static str) -> StructuredLogEntry {
 
 /// Security events that are possible
 pub mod security_events {
-    // JSON-RPC server
-    // ---------------
-
-    /// Admission Control received a transaction with an invalid signature
-    pub const INVALID_TRANSACTION_AC: &str = "InvalidTransactionAC";
```
mimoo, line 33: I don't think we should log this, as we're probably going to receive a ton of invalid transaction via the json-rpc interface

mempool/src/shared_mempool/tasks.rs
```diff
@@ -238,6 +238,16 @@ pub(crate) async fn process_transaction_broadcast<V>(
 {
     let results = process_incoming_transactions(&smp, transactions, timeline_state).await;
     log_txn_process_results(&results, Some(peer.peer_id()));
+
+    // security log
+    for (_mempool_status, maybe_vm_status) in &results {
+        if let Some((transaction, _vm_status)) = maybe_vm_status {
+            send_struct_log!(security_log(security_events::INVALID_TRANSACTION_MP)
+                .data("transaction", &transaction)
+                .data("from_peer", &peer));
```
mimoo, line 247: logging the transaction and the peer who sent it to us

mempool/src/shared_mempool/tasks.rs
```diff
@@ -360,7 +372,7 @@ where
                     Some(validation_status) => {
                         statuses.push((
                             MempoolStatus::new(MempoolStatusCode::VmError),
-                            Some(validation_status.clone()),
+                            Some((transaction.clone(), validation_status.clone())),
```
mimoo, line 375: This option is only set when a transaction couldn't get executed.

common/logger/src/security.rs
```diff
@@ -26,16 +26,10 @@ pub fn security_log(name: &'static str) -> StructuredLogEntry {
...
-
     // Mempool
     // -------
 
-    /// Mempool received a transaction with an invalid signature
+    /// Mempool received a transaction from another peer with an invalid signature
```
mimoo, line 32: This log is probably going to be noisier if we are not in sync (or another validator is not in sync) with the latest state. Nonetheless an increase in the rate should give us a good indication that something is off.

mempool/src/shared_mempool/tasks.rs
davidiw: why don't we just keep track of the transactions and then zip the results together? It seems unfortunate to require that we return the transaction when we already have it... and then we have to modify a bunch of other APIs too.

mimoo: I did start doing this, but two things that went in my head when I made this change that made me go this way:

- I wanted to avoid using functions that can crash if the length, for any reason, ends up differing (after a refactor for example).
- I thought it would be useful to return the transaction along with a failure

That being said, it is only used in one place for now. Wdyt?

json-rpc/src/tests/unit_tests.rs
```diff
@@ -183,9 +183,15 @@ fn test_transaction_submission() {
...
-                (MempoolStatus::new(MempoolStatusCode::VmError), vm_status)
+            let vm_status = validator
+                .validate_transaction(txn.clone())
+                .unwrap()
+                .status();
+            let result = if let Some(vm_status) = vm_status {
+                (
+                    MempoolStatus::new(MempoolStatusCode::VmError),
+                    Some((txn, vm_status)),
+                )
```
davidiw, line 194: well that's not entirely true because now we modify every single API else where... in fact, I would argue that we should /expect/ an equal number of transactions going in as statuses coming out... and if that fails, we likely have a pretty nasty bug. So hardening around potential issues in the future seems ... unintuitive since that would mask programming bugs.

mimoo: good point, I fixed this and it looks much simpler now :) and no unwrap

mempool/src/shared_mempool/tasks.rs
```diff
@@ -236,8 +236,28 @@ pub(crate) async fn process_transaction_broadcast<V>(
...
+    let results = process_incoming_transactions(&smp, transactions.clone(), timeline_state).await;
     log_txn_process_results(&results, Some(peer.peer_id()));
+
+    // we assume that the results contains a vec of the same length of transactions
+    debug_assert_eq!(results.len(), transactions.len());
+
+    // security log all the failed transactions
+    let failed_transactions = results
+        .iter()
+        // create (MempoolStatus, transaction) iterator
```
JoshLind, line 249: nit: This comment confused me initially. I'd maybe update this comment to either: (i) change MempoolStatus to SubmissionStatus (as that is the type returned by process_incoming_transactions); or (ii) I'd be more explicit and just say:
// create ((MempoolStatus, Option<VMStatus>), transaction) iterator. It

mempool/src/shared_mempool/tasks.rs
```diff
@@ -236,8 +236,28 @@ pub(crate) async fn process_transaction_broadcast<V>(
 ) where
     V: TransactionValidation,
 {
-    let results = process_incoming_transactions(&smp, transactions, timeline_state).await;
+    // process transactions and log the result
+    let results = process_incoming_transactions(&smp, transactions.clone(), timeline_state).await;
     log_txn_process_results(&results, Some(peer.peer_id()));
+
+    // we assume that the results contains a vec of the same length of transactions
+    debug_assert_eq!(results.len(), transactions.len());
```
JoshLind, line 244: Yeah, I think if this fails at runtime, we have a bigger issue to worry about :)

davidiw:
Thanks!

Issue #4807, opened 2020-06-28
[tcb / consensus] Hardening


davidiw:
This issue covers the remainder of tasks necessary for completion of SafetyRules

- [x] LEC issues should not cause consensus to panic
- [x] Logging of all events, both positive and negative
- [x] Moving consensus_state logging into safety rules
- [x] Network errors do not affect consensus
- [ ] Consensus handling a key is missing
- [ ] Consensus handling not being involved in an epoch
- [x] Consensus handling a request to perform an outdated action (construct_and_sign_vote, sign_timeout, sign_proposal)
- [x] Consensus handling an unconfigured SafetyRules in the middle of an epoch / Consensus safely restarts an epoch if SafetyRules goes offline
- [x] Consensus losing a vote and re-requesting for the same round
- [x] The waypoint is more than 100 epochs behind the current state (currently the framework only retrieves 100
- [x] SafetyRules idempotent behavior to initialize with the same value without restart
- [x] Metrics for all public persistent store data and failures
- [x] Logging error where appropriate
- [x] Implement a write-through cache for persistent storage
- [ ] Remove explicit dependency on anyhow from interfaces
- [x] SafetyRules config
- [x] SafetyRules supports dns

## Discussion

**Issue**) Consensus handling a key is missing
**David**) Ideally, SafetyRules would keep checking and logging errors so the admin would respond. Consensus would fail operations but could still participate in storing and committing blocks.

**Issue**) Consensus handling not being involved in an epoch
**David**) Should consensus disable itself and rely on state sync? Should it exhibit the same behavior as if the key is missing? Does it even bother calling into SafetyRules? I'd say if we want to enter consensus the behavior should be the same, but SafetyRules only logs debugging info but still errors.

**Issue**) Consensus handling a request to perform an outdated action (construct_and_sign_vote, sign_timeout, sign_proposal)
**David**) Simply log and wait for the next event.

**Issue**) Consensus losing a vote and re-requesting for the same round
 **David**) Opinion, this is an optimization to increase liveness and avoid cases where consensus is stuck because only `2f + 1` nodes remain and `1` has a failure of this nature. This does not seem like a V1 critical area and can be added without breaking changes.

**Issue**) The waypoint is more than 100 epochs behind the current state (currently the framework only retrieves 100,
**David**) This should be optional, the waypoint should probably be updated in the case it is this far behind.

davidiw:
[...]

PR #4881, opened 2020-07-02
[network] Write-Through Caches for KVStorage

Rustinante:
## Motivation

- Implemented a write-through cache `CachedStorage`, which caches the methods of any structs implementing `KVStorage`, and which forwards the calls of any structs implementing `CryptoStorage`.
- Added the `CachedStorageVariant` enum as the cached equivalent of `Storage`.
- Changed the storage representation of `PersistentSafetyStorage` to a `StorageVariant` enum to facilitate `CachedStorage` conversion.

Issue: #4807 

### TLDR
The serializer and thread benchmarks received about 10% speedup, while the vault benchmark received a speedup of almost 2X!!

### Details

```
SafetyRules/InMemory    time:   [40.328 ms 40.691 ms 41.095 ms]                                
                        change: [+0.0246% +1.3950% +2.7212%] (p = 0.07 > 0.05)
                        No change in performance detected.
Found 2 outliers among 10 measurements (20.00%)
  2 (20.00%) high mild
Benchmarking SafetyRules/OnDisk: Warming up for 3.0000 s
Warning: Unable to complete 10 samples in 5.0s. You may wish to increase target time to 9.7sor enable flat sampling.
SafetyRules/OnDisk      time:   [245.75 ms 404.89 ms 537.65 ms]                             
                        change: [-0.4897% +46.000% +106.64%] (p = 0.12 > 0.05)
                        No change in performance detected.
Found 1 outliers among 10 measurements (10.00%)
  1 (10.00%) high mild
SafetyRules/Serializer  time:   [182.23 ms 187.17 ms 193.11 ms]                                 
                        change: [-11.787% -8.9537% -5.9244%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 2 outliers among 10 measurements (20.00%)
  1 (10.00%) high mild
  1 (10.00%) high severe
SafetyRules/Thread      time:   [186.72 ms 189.96 ms 194.74 ms]                             
                        change: [-18.203% -13.498% -9.0997%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 1 outliers among 10 measurements (10.00%)
  1 (10.00%) high severe
Benchmarking SafetyRules/Vault: Warming up for 3.0000 s
Warning: Unable to complete 10 samples in 5.0s. You may wish to increase target time to 6.8s.
SafetyRules/Vault       time:   [652.60 ms 672.96 ms 691.13 ms]                            
                        change: [-47.991% -45.612% -43.150%] (p = 0.00 < 0.05)
                        Performance has improved.
```

Yes

## Test Plan

cargo xtest

cc: @davidiw


consensus/safety-rules/src/persistent_safety_storage.rs
```diff
@@ -80,35 +85,56 @@ impl PersistentSafetyStorage {
...
-            .export_private_key_for_version(CONSENSUS_KEY, version)
-            .map_err(|e| e.into())
+        match &self.internal_store {
+            StorageVariant::Storage(storage) => storage
+                .export_private_key_for_version(CONSENSUS_KEY, version)
+                .map_err(|e| e.into()),
+            StorageVariant::CachedStorage(_) => {
+                Err(Error::InternalError("Cannot operate on keys with CachedStorage".into()).into())
+            }
+        }
```
davidiw, line 123: if we're going with this style, the cached layer would just pass any CryptoKV calls to the underlying storage

consensus/safety-rules/src/persistent_safety_storage.rs
```diff
@@ -80,35 +85,56 @@ impl PersistentSafetyStorage {
...
-            .get_public_key(EXECUTION_KEY)
-            .map(|r| r.public_key)?)
+        match &self.internal_store {
+            StorageVariant::Storage(storage) => Ok(storage
+                .get_public_key(EXECUTION_KEY)
+                .map(|r| r.public_key)?),
+            StorageVariant::CachedStorage(_) => {
+                Err(Error::InternalError("Cannot operate on keys with CachedStorage".into()).into())
+            }
+        }
```
davidiw, line 134: same

consensus/safety-rules/src/persistent_safety_storage.rs
```diff
@@ -117,7 +143,7 @@ impl PersistentSafetyStorage {
         Ok(())
     }
 
-    pub fn last_voted_round(&self) -> Result<Round> {
+    pub fn last_voted_round(&mut self) -> Result<Round> {
```
davidiw, line 146: why does this become mutable?

davidiw: I would prefer this retains its non-mutable interface by either using locks or refcell in the cache.

Rustinante: It seems that we can't use a RefCell since KVStorage requires the `Sync` trait... I think locks might be possible

Rustinante:
I tried to run the x bench but for vault I get
```
Vault (http://localhost:8200) is not availble, experiment will not be launched
```...

**Comment by davidiw** [comment]
Awesome! You can run a local vault instance either by downloading vault for your OS and running it in dev mode or using docker and running docker/vault/run.sh -- will review code shortly.

File: secure/storage/src/storage.rs [hunk 4]
```diff
@@ -25,3 +25,34 @@ pub enum Storage {
...
+    InMemoryStorage(CachedKVStorage<InMemoryStorage>),
+    NamespacedStorage(CachedKVStorage<NamespacedStorage>),
+    OnDiskStorage(CachedKVStorage<OnDiskStorage>),
+}
+
+#[enum_dispatch(KVStorage, CryptoStorage)]
+pub enum StorageVariant {
+    Storage(Storage),
+    CachedStorage(CachedStorage),
+}
```
davidiw, line 58: Considering that we are only caching the KV side, could we do something like how the NamespacedStorage works? That way we don't end up with this additional layer of indirection? I would also probably consider just reusing the the concept with how that works rather than introducing "into_cached", unless we want to add a new feature "into_namespaced" which would be kinda hard.

Other than that, this looks good!

As future work, we should consider that each Storage interface is just a client to a backend and make them not Send + Sync and then we can get rid of the locks on the cache and instead make them within the storage itself -- if necessary.

Rustinante: I see, so are we thinking about replacing the original `Storage` enum with a struct with a Boxed storage inside, like
```
pub enum Storage {
    storage: Box<dyn KVStorage + CryptoStorage>
}
```
so that `storage` can be either a normal `GitHubStorage`, `VaultStorage`, etc., or their cached equivalent?
For the `IntoCached` trait, I added it only for ergonomics, and can be removed without any problems, in which case we would just call `CachedStorage::new` in its place :) Please let me know if that's what you have in mind. Thanks!!

davidiw: Yeah, let's keep CachedStorage to be similar to NamespacedStorage and then we should be good to go with this PR, right?

Rustinante: Okay, I renamed a few things and used `Box`. Please let me know if they look okay, thanks!

Rustinante: Actually should we also cache the public and private keys? I'm not sure if they'll be modified by other processes, which would result in a stale cache..

secure/storage/src/storage.rs
```diff
@@ -21,3 +21,9 @@ pub enum Storage {
     NamespacedStorage(NamespacedStorage),
     OnDiskStorage(OnDiskStorage),
 }
+
+#[enum_dispatch(KVStorage, CryptoStorage)]
+pub enum StorageVariant {
+    Storage(Storage),
+    CachedStorage(CachedStorage<Storage>),
```
davidiw, line 28: Is there a reason for not including CachedStorage in the enum for Storage? much like NamespacedStorage?


Other than that, this is looking good.

Rustinante: I didn't think of that! Okay so I just updated it and removed `StorageVariant`, thanks!

secure/storage/src/storage.rs
davidiw: yeah, those are modifiable by other processes, so we do not want to cache them. That's why I had wondered if we should have just made a cache specific to LSR by adding the layer to PersistentSafetyStorage. But this is good too!

davidiw:
going to run our perf tool on this and report back.

secure/storage/src/storage.rs
Rustinante: Hmm if that's the case then I'd have to make sure the CachedStorage doesn't cache the public and private keys, cuz right now they're caching them...

davidiw: oh, that might explain why I cannot run the perf test... there's no logging enabled right now for safety rules, I'm trying to fix that in another PR :)

Rustinante: Haha, okay I just changed the cache so that it doesn't cache the public/private keys.. Please let me know if it works now, thanks!

Rustinante:
I'm actually not too familiar with how the persistent storage will be called, but I'm wondering if it's possible for multiple threads access the storage and cause a deadlock? Additionally, whenever I try to run the `tests::spawned_process::test ` for the first time after a build, it won't terminate, but it does terminate for the subsequent runs..

consensus/safety-rules/src/persistent_safety_storage.rs
```diff
@@ -47,52 +49,65 @@ impl PersistentSafetyStorage {
...
     ) -> Result<()> {
-        internal_store.import_private_key(CONSENSUS_KEY, consensus_private_key)?;
-        internal_store.import_private_key(EXECUTION_KEY, execution_private_key)?;
-        internal_store.set(EPOCH, Value::U64(1))?;
-        internal_store.set(LAST_VOTED_ROUND, Value::U64(0))?;
-        internal_store.set(OWNER_ACCOUNT, Value::String(author.to_string()))?;
-        internal_store.set(PREFERRED_ROUND, Value::U64(0))?;
-        internal_store.set(WAYPOINT, Value::String(waypoint.to_string()))?;
-        internal_store.set(
+        storage.import_private_key(CONSENSUS_KEY, consensus_private_key)?;
```
davidiw, line 76: the reason for the naming was because it distinguishes it from the fact that this is called PersistentSafetyStore

davidiw:
> I'm actually not too familiar with how the persistent storage will be called, but I'm wondering if it's possible for multiple threads to access the storage and cause a deadlock? Additionally, whenever I try to run the `tests::spawned_process::test ` for the first time after a build, it won't terminate, but it does terminate for the subsequent runs..

no the application is single threaded at that layer

there are only assumptions that the consensus key could change the rest should be only modified by safety rules, hence why we can cache :)

Rustinante:
Great! I just renamed the variables. Any luck on running the perf tools?

davidiw:
There was about a 2% perf gain. Here's a look at the Safety Rules latency (after / before):

Screen Shot 2020-07-19 at 10 26 46 AM

so definitely an improvement but not significant enough...

I think merging the write fields (last_vote and last_voted_round) will be big as well eliminating the extra data that comes with the vote proposal that isn't critically necessary for voting. There's someone looking at that second problem, where as we discussed the first before.

Let's land this as is and decide if we want caching to be default later.

@bors-libra r+

Thanks!!

PR #4883, opened 2020-07-02
[state sync] handle upstream peers in different networks and add fail…

christinacelee:
…over logic

## Summary

Before this PR, in state sync, upstream peers in different networks were treated equally in terms of requesting syncs. 
This PR adds network preference logic for selecting the upstream peer to request syncs from:
- select the peer from the highest-pref'ed network (defined by UpstreamConfig) with any live peers
- selecting a peer from a set of active peers in the same network will be based on peer score

Notes:
- before, each call to `PeerManager::pick_peer` was iterating through the entire peer set. This PR makes peer-picking more efficient by removing that extra iteration by persisting `eligible_peers` in `PeerManager` , which is only updated whenever a peer is added/lost or when a meaningful score change happens. 

## Test Plan

new/existing tests

### DIFF
--- a/state-synchronizer/src/peer_manager.rs
+++ b/state-synchronizer/src/peer_manager.rs
@@ -2,6 +2,7 @@
 // SPDX-License-Identifier: Apache-2.0
 
 use crate::counters;
+use itertools::Itertools;
 use libra_config::config::{PeerNetworkId, UpstreamConfig};
 use libra_logger::prelude::*;
 use rand::{
@@ -61,19 +62,22 @@ pub enum PeerScoreUpdateType {
 }
 
 pub struct PeerManager {
+    // list of peers that are eligible for this node to send sync requests to
+    eligible_peers: Vec<PeerNetworkId>,
     peers: HashMap<PeerNetworkId, PeerInfo>,
     requests: BTreeMap<u64, ChunkRequestInfo>,
-    weighted_index: Option<WeightedIndex<f64>>,
     upstream_config: UpstreamConfig,
+    weighted_index: Option<WeightedIndex<f64>>,
 }
 
 impl PeerManager {
     pub fn new(upstream_config: UpstreamConfig) -> Self {
         Self {
+            eligible_peers: vec![],
             peers: HashMap::new(),
             requests: BTreeMap::new(),
-            weighted_index: None,
             upstream_config,
+            weighted_index: None,
         }
     }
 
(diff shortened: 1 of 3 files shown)


state-synchronizer/src/peer_manager.rs
```diff
@@ -122,52 +126,80 @@ impl PeerManager {
...
+            self.upstream_config
+                .networks
+                .iter()
+                .find_map(|network| active_peers_by_network.get(network))
+                .unwrap_or(&vec![])
+                .to_vec()
+        } else {
+            // no failover option
+            // all upstream peers belong to the same network
+            self.peers
```
zekun000, line 198: do you need this else branch? isn't it the same as above?

christinacelee: for validators, its `NetworkId::Validator` is not explicitly specified in the upstream config, so the above branch would give no active peers

state-synchronizer/src/peer_manager.rs
```diff
@@ -122,52 +126,80 @@ impl PeerManager {
...
+                .iter()
+                .filter(|(_peer, peer_info)| peer_info.is_alive)
+                .map(|(peer, peer_info)| (peer.network_id(), (peer, peer_info)))
+                .into_group_map();
+
+            // find the first network with any live peers
+            self.upstream_config
+                .networks
+                .iter()
+                .find_map(|network| active_peers_by_network.get(network))
```
zekun000, line 192: it looks to me the ranking follows the order of the positions, do we want to align with `get_upstream_preference`?

e.g. if upstream_config.network is [fullnode 1, fullnode 2, validator], they return different result.

state-synchronizer/src/peer_manager.rs
zekun000: hmm, when there's failover, `validator` would be specified in the upstream config as first element?

state-synchronizer/src/peer_manager.rs
christinacelee: it's part of the config req that the `networks` in the upstream config are already ordered by preference

state-synchronizer/src/peer_manager.rs
christinacelee: I don't think there ever is a use case where a validator will ever failover? and a FN would never list the validator entwork as an upstream config

zekun000: can you give examples how the upstream config looks like as validator, FN, and FN with failover?

christinacelee: validator: empty []
VFN w/o failover: [NetworkId::Private("vfn")] 
VFN w/ failover: [NetworkId::Private("vfn"), NetworkId::Public]
public FN: [NetworkId::Public]

PR #4884, opened 2020-07-02
[cluster-test] Create a test for compatibility between commits

runtian-zhou:
## Motivation

Create a backward compatibility test between two commits in cluster test. The test is done by following:
1. Build an image of master as a comparison base, and intialize a network with all validators running at master branch.
2. Generate some traffic to make sure network is healthy.
3. Replace partial node with a newer build (e.g: PR trying to merge in).
4. Generate some traffic to make sure the computation result still converges.
5. Repeat 3&4 again untill all nodes are updated to the newer build.

This is not a full compatibility test as it only touches features that is generated through the tx_emitter, which should only include payment transaction. To better test the system we might need to pull in the system in the real network in the future. This is only the first step to start this compatibility testing process.

Yes

## Test Plan

I ran two examples:
1. Checked two compatible changes:
```
./scripts/cti --pr 4884 --validator-tag land_3ebaffed --run compatibility_test  -- --updated-image-tag land_6016a2ba
```
And observe the test is executed successfully.
2. Checked two imcompaitble changes:
```
./scripts/cti  --pr 4884  --validator-tag land_9994cbc0 --run compatibility_test  -- --updated-image-tag land_57b4cb41
```
In such case the replaced node will fail to boot up and cause `wait_for_json_rpc` to fail.

### DIFF
--- a/testsuite/cluster-test/src/experiments/compatibility_test.rs
+++ b/testsuite/cluster-test/src/experiments/compatibility_test.rs
@@ -56,6 +56,7 @@ pub async fn update_batch_instance(
 
     // Add a timeout to have wait for validators back to healthy mode.
     // TODO: Replace this with a blocking health check.
+    info!("Wait for the instance to sync up with peers");
     time::delay_for(Duration::from_secs(20)).await;
     Ok(())
 }
@@ -168,7 +169,7 @@ impl Experiment for CompatibilityTest {
     }
 
     fn deadline(&self) -> Duration {
-        Duration::from_secs(20 * 60)
+        Duration::from_secs(16 * 60)
     }
 }
 
--- a/testsuite/cluster-test/src/experiments/compatibility_test.rs
+++ b/testsuite/cluster-test/src/experiments/compatibility_test.rs
@@ -1,0 +1,142 @@
+// Copyright (c) The Libra Core Contributors
+// SPDX-License-Identifier: Apache-2.0
+
+#![forbid(unsafe_code)]
+
+use crate::{
+    cluster::Cluster,
+    experiments::{update_batch_instance, Context, Experiment, ExperimentParam},
+    instance,
+    instance::Instance,
+    tx_emitter::EmitJobRequest,
+};
+use async_trait::async_trait;
+use libra_logger::prelude::*;
+use std::{collections::HashSet, fmt, iter::once, time::Duration};
+use structopt::StructOpt;
+
+#[derive(StructOpt, Debug)]
+pub struct ComaptiblityTestParams {
+    #[structopt(
+        long,
+        default_value = "15",
+        help = "Number of nodes to update in the first batch"
+    )]
+    pub count: usize,
+    #[structopt(long, help = "Image tag of newer validator software")]
+    pub updated_image_tag: String,
+}
+
+pub struct CompatibilityTest {
+    first_node: Instance,
+    first_batch: Vec<Instance>,
+    second_batch: Vec<Instance>,
+    full_nodes: Vec<Instance>,
+    updated_image_tag: String,
+}
+
+impl ExperimentParam for ComaptiblityTestParams {
+    type E = CompatibilityTest;
+    fn build(self, cluster: &Cluster) -> Self::E {
+        if self.count > cluster.validator_instances().len() || self.count == 0 {
+            panic!(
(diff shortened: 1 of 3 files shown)


testsuite/cluster-test/src/experiments/mod.rs
```diff
@@ -131,7 +140,46 @@ pub fn get_experiment(name: &str, args: &[String], cluster: &Cluster) -> Box<dyn
...
     known_experiments.insert("generate_cpu_flamegraph", f::<CpuFlamegraphParams>());
     known_experiments.insert("versioning_testing", f::<ValidatorVersioningParams>());
+    known_experiments.insert("compatibility_test", f::<ComaptiblityTestParams>());
 
     let builder = known_experiments.get(name).expect("Experiment not found");
     builder(args, cluster)
 }
+
+/// Reboot `updated_instance` with newer image tag
+pub async fn update_batch_instance(
```
andll, line 150: I know it is draft PR but have some early ask - I would keep mod.rs for interfaces, can we just move it to CompatibilityTest file and import in versioning test?

sblackshear:
Can you add some more detail on what "backward compatibility" means in the summary?

testsuite/cluster-test/src/experiments/compatibility_test.rs
```diff
@@ -0,0 +1,187 @@
...
+    info!("Wait for the instances to recover.");
+    let futures: Vec<_> = instances
+        .iter()
+        .map(|instance| instance.wait_json_rpc(deadline))
+        .collect();
+    try_join_all(futures).await?;
+
+    // Add a timeout to have wait for validators back to healthy mode.
+    // TODO: Replace this with a blocking health check.
+    time::delay_for(Duration::from_secs(20)).await;
```
andll, line 59: Do we want an `info` before this delay, to know that their jsonrpc are recovered and we are just sleeping now?

testsuite/cluster-test/src/experiments/compatibility_test.rs
```diff
@@ -0,0 +1,187 @@
...
+        update_batch_instance(context, &self.full_nodes, self.updated_image_tag.clone()).await?;
+        context
+            .tx_emitter
+            .emit_txn_for(job_duration, fullnode_txn_job)
+            .await?;
+        Ok(())
+    }
+
+    fn deadline(&self) -> Duration {
+        Duration::from_secs(20 * 60)
```
andll, line 171: How long does it normally take? Guideline is to put here 2 * normal_experiment_duration.
Too large values aren't great as sometimes it may have experiment running when it simply hanged

zekun000:
not only the fullnode, afaiu, LSR is a separate instance that needs update too. we probably should consider some sort of grouping of the instances.

andll:
Sure.
Right now LSR is not running in separate instance, but full nodes are.
When we have full support of LSR in separate instance and have it in nightly/land blocking, we will update compatibility test - most likely grouping API will be part of introducing better support for LSR

testsuite/cluster-test/src/experiments/compatibility_test.rs
runtian-zhou: Each restart will take 2min at most. There were four restarts. So I would imagine the execution time would be 4*2 min + txn generation. Given such calculation 20min seems fine? We might want to reduce it to 16min just to constraint the time consumption a little bit.

andll: Make sense!

testsuite/cluster-test/src/experiments/compatibility_test.rs
```diff
@@ -0,0 +1,187 @@
...
+
+        info!("3. Changing images for the rest of the nodes");
+        update_batch_instance(context, &self.second_batch, self.updated_image_tag.clone()).await?;
+        context
+            .tx_emitter
+            .emit_txn_for(job_duration, validator_txn_job)
+            .await?;
+
+        info!("4. Changing images for full nodes");
+        update_batch_instance(context, &self.full_nodes, self.updated_image_tag.clone()).await?;
```
runtian-zhou, line 162: Full nodes are upgraded here already right?

andll: Right, I think I missed it!

PR #4886, opened 2020-07-02
[vm] New error framework

tnowacki:
- VM now has its own internal error structure
- VMStatus is now used starting at the Libra layer

## Motivation

- We need gather location information up to the VM status. this is the first step to do this

## Test Plan

- cargo test

### DIFF
--- a/language/move-core/types/src/vm_status.rs
+++ b/language/move-core/types/src/vm_status.rs
@@ -3,6 +3,7 @@
 
 #![allow(clippy::unit_arg)]
 
+use crate::account_address::AccountAddress;
 use anyhow::Result;
 #[cfg(any(test, feature = "fuzzing"))]
 use proptest::prelude::*;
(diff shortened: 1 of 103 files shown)


language/vm/src/errors.rs
```diff
@@ -32,98 +32,299 @@ pub const EINSUFFICIENT_PRIVILEGES: u64 = 11; // user lacks the credentials to d
...
+    Script,
+    Module(ModuleId),
+}
+
+#[derive(Debug)]
+pub struct VMError {
+    pub major_status: StatusCode,
+    pub sub_status: Option<u64>,
+    pub message: Option<String>,
+    pub location: Location,
```
tzakian, line 51: Wondering what the location of an `EXECUTED` major_status would be. Would it be Script?

tnowacki: It is most likely Unknown

tnowacki: Maybe "Undefined" would be better than Unknown. Thoughts?

language/move-vm/runtime/src/interpreter.rs
```diff
@@ -117,10 +117,12 @@ impl Interpreter {
         ty_args: Vec<Type>,
         args: Vec<Value>,
     ) -> VMResult<()> {
-        verify_args(function.parameters(), &ty_args, &args)?;
+        verify_args(function.parameters(), &ty_args, &args).map_err(|e| self.set_location(e))?;
```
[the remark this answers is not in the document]
tnowacki: I can't because of natives :(

language/move-core/types/src/vm_status.rs
```diff
@@ -106,86 +107,83 @@ impl std::error::Error for VMStatus {}
 
 impl VMStatus {
     /// Create a new VM status with major status `major_status`.
-    pub fn new(major_status: StatusCode) -> Self {
+    pub fn new(major_status: StatusCode, sub_status: Option<u64>, message: Option<String>) -> Self {
```
tzakian, line 110: Could `message` be an `Option<Location>` instead of a String? Either that or a separate more structured type:
```
enum SourceContext {
    Script
    Module(ModuleId)
}
```
Reason being we want this type to be pretty well structured since it will be a part of consensus (and therefore they all have to agree on the exact status), and also stored in the transaction accumulator.

Additionally, if the `message` is `None` we should also set the `sub_status` to `None` I think. We only want to put out source-generated errors otherwise the versioning story for `sub_status` becomes more complex since it is then versioned by the VM and modules+txn scripts as opposed to just the modules+txn scripts.


Edit: on second thought, it might be easier if I make this change as part of the PR that adds VMStatus to the rest of the system. If that makes more sense to you I'm happy to make these changes in that PR.

language/vm/src/errors.rs
```diff
@@ -1,126 +1,319 @@
...
-    VMStatus::new(err).with_message(msg)
+pub fn verification_error(status: StatusCode, kind: IndexKind, idx: TableIndex) -> PartialVMError {
+    PartialVMError::new(status).at_index(kind, idx)
 }
 
-pub fn append_err_info(status: VMStatus, kind: IndexKind, idx: usize) -> VMStatus {
-    let msg = format!("at index {} while indexing {}", idx, kind);
-    status.append_message_with_separator(' ', msg)
-}
+// pub fn append_err_info(error: PartialVMError, kind: IndexKind, idx: usize) -> PartialVMError {
```
tzakian, line 311: Remove?

language/vm/src/errors.rs
```diff
@@ -1,126 +1,319 @@
...
+//     let msg = format!("at index {} while indexing {}", idx, kind);
+//     error.append_message_with_separator(' ', msg)
+// }
 
-pub fn err_at_offset(status: StatusCode, offset: usize) -> VMStatus {
-    let msg = format!("At offset {}", offset);
-    VMStatus::new(status).with_message(msg)
-}
+// pub fn err_at_offset(status: StatusCode, offset: usize) -> PartialVMError {
```
tzakian, line 316: Remove?

language/move-vm/runtime/src/interpreter.rs
```diff
@@ -210,6 +230,28 @@ impl Interpreter {
         function: Arc<Function>,
         ty_args: Vec<Type>,
     ) -> VMResult<()> {
+        // TODO Remove once native functions push a frame on the stack
```
dariorussi, line 233: I am not sure native functions will ever push a frame on the stack.... :)

language/move-vm/runtime/src/interpreter.rs
```diff
@@ -210,6 +230,28 @@ impl Interpreter {
...
+        let module_id = match function.module_id() {
```
dariorussi, line 234: should this go inside the `map_err`? I am not clear what is doing up here....

language/move-vm/runtime/src/interpreter.rs
```diff
@@ -614,6 +674,11 @@ impl CallStack {
     fn pop(&mut self) -> Option<Frame> {
         self.0.pop()
     }
+
+    fn current_location(&self) -> Location {
```
dariorussi, line 678: I find this odd, but I assume it was the only way to do it and we can review later.
In other words it seems strange to ask the call stack for the location

language/vm/src/errors.rs
dariorussi: hmmm, not sure what I would use here... I don't like either, though they are both fine.
Maybe `NoLocation`? :)

language/move-core/types/src/vm_status.rs
tnowacki: Let's talk after this lands about next steps for the errors

language/move-vm/runtime/src/interpreter.rs
tnowacki: I had some issues with Arc. I think I just need to clone to fix this

language/move-vm/runtime/src/interpreter.rs
tnowacki: Why? The call stack is by its nature what defines the location. In order to know what module/script I am in, I need to know what function I am in. The call stack is the source of truth for that information

language/vm/src/errors.rs
tnowacki: I want something that really describes this thing comes from out in the ether, not inside a specific Move context. So I think NoLocation is good, but Undefined better captures this ethereal feeling. 

I feel like there is a better word out there, but I cannot find it

language/move-vm/runtime/src/interpreter.rs
dariorussi: I did not see the details of the PR and maybe there are places where that matters.
I guess I was coming from a model where we are either inside a frame, and so unwinding would lead you there, or not and you don't need it, or that is what it seemed to me
Again I'll have to play with it when I merge ;) so I'll understand things better

language/vm/src/errors.rs
dariorussi: after reading your summary in the slack channel I see what you are after.
`Internal`?

tnowacki: Maybe... Sometimes it sort feels like its tied to a module, but it's not clear exactly whats going on. I'll ship it for now, and we should all get a better feeling

PR #4888, opened 2020-07-02
[validator-set] Make update_and_reconfigure preserve the members of the set

valerini:
## Motivation

As per extensive discussions, it's been decided that the update_and_reconfigure should not implicitly remove any validators, or otherwise change the members of the set. Addition and removal of validators should be done via explicit calls to LibraSystem module's methods.

Yes


language/stdlib/modules/LibraSystem.move
```diff
@@ -106,36 +108,24 @@ module LibraSystem {
...
-            // if the validator is invalid, remove it from the set
-            let validator_address = Vector::borrow(validators, i).addr;
-            if (ValidatorConfig::is_valid(validator_address)) {
-                let validator_info_update = update_ith_validator_info_(validators, i);
-                configs_changed = configs_changed || validator_info_update;
-            } else {
-                _  = Vector::swap_remove(validators, i);
-                configs_changed = true;
-            }
+        let i = 0;
```
emmazzz, line 124: Looks great! This change will greatly reduce the amount of work I have to do 😄

PR #4890, opened 2020-07-02
[language] fix disassembler crash when source map is not available

bob-wilson:
This changes the source map's dummy_function_map to add generated names ("Arg0", "Arg1", etc.) for function parameters. This is similar to what that function is already doing for generic type parameters. Without this, the disassembler crashes for even trivial code when a compiler-generated source map is not available.

## Motivation

Fix crashes

Yes

## Test Plan

I compiled a simple test program without generating a source map and verified that the disassembler no longer crashes. At some point, we should figure out how to set up some automated testing for the disassembler.

### DIFF
--- a/language/compiler/bytecode-source-map/src/source_map.rs
+++ b/language/compiler/bytecode-source-map/src/source_map.rs
@@ -255,6 +255,13 @@ impl<Location: Clone + Eq> FunctionSourceMap<Location> {
             self.add_type_parameter((name, default_loc.clone()))
         }
 
+        // Generate names for each parameter
+        let params = module.signature_at(function_handle.parameters);
+        for i in 0..params.0.len() {
+            let name = format!("Arg{}", i);
+            self.add_parameter_mapping((name, default_loc.clone()))
+        }
+
         if let Some(code) = &function_def.code {
             let locals = module.signature_at(code.locals);
             for i in 0..locals.0.len() {


language/compiler/bytecode-source-map/src/source_map.rs
```diff
@@ -255,6 +255,13 @@ impl<Location: Clone + Eq> FunctionSourceMap<Location> {
             self.add_type_parameter((name, default_loc.clone()))
         }
 
+        // Generate names for each parameter
+        let params = module.signature_at(function_handle.parameters);
+        for i in 0..params.0.len() {
+            let name = format!("Arg{}", i);
```
tzakian, line 261: Maybe lowercase `arg` instead here?

bob-wilson: Not sure. I was going for consistency with the generated type parameter names, which are "Ty0", "Ty1", etc. WDYT?

tzakian: Hmm good point. I was thinking lowercase since arguments should be lowercase, but I'm fine keeping this as-is as well.

bob-wilson: It's easy to change later if people prefer lowercase.... I'll stick with this for now

PR #4891, opened 2020-07-02
[JSON RPC] Refactor set_response_error()

JoshLind:
## Motivation

This small PR refactors some of the error handling code in the JSON RPC server. More specifically, it refactors counter labels into constants and provides a helper method to set the JSON RPC response error (to minimize code duplication).

Yes.

## Test Plan

All local tests still pass.

## Related PRs

None.

### DIFF
--- a/json-rpc/src/runtime.rs
+++ b/json-rpc/src/runtime.rs
@@ -19,6 +19,14 @@ use warp::{
     Filter,
 };
 
+// Counter labels for runtime metrics
+const LABEL_FAIL: &str = "fail";
+const LABEL_INVALID_FORMAT: &str = "invalid_format";
+const LABEL_INVALID_METHOD: &str = "invalid_method";
+const LABEL_INVALID_PARAMS: &str = "invalid_params";
+const LABEL_MISSING_METHOD: &str = "method_not_found";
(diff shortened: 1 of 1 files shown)


json-rpc/src/runtime.rs
```diff
@@ -19,6 +19,14 @@ use warp::{
     Filter,
 };
 
+// Counter labels for runtime metrics
+const LABEL_FAIL: &str = "fail";
```
mimoo, line 23: maybe a more descriptive error :P ?

mimoo:
I'm surprised that this actually adds more lines than it removes, but it makes it easier to read the function so it's a win in my book

json-rpc/src/runtime.rs
JoshLind: I'm gonna punt on this for now. I'm not sure how much ecosystem support there already is for the JSON RPC metrics, so will refrain from modifying the counter name. I agree though.. it's not great!

JoshLind:
Thanks, David! 😄 @bors-libra r=mimoo

> I'm surprised that this actually adds more lines than it removes, but it makes it easier to read the function so it's a win in my book

Yeah.. I think it's because I added constants 😄

PR #4897, opened 2020-07-02
[libra-trace] sample trace_event when trace_guard is dropped.

changyonghu:
trace_event during the drop of trace_guard should also be sampled.

### DIFF
--- a/common/debug-interface/src/libra_trace.rs
+++ b/common/debug-interface/src/libra_trace.rs
@@ -80,21 +80,29 @@ macro_rules! send_logs {
 #[macro_export]
 macro_rules! trace_code_block {
     ($stage:expr, $node:tt) => {
-        let trace_guard = $crate::libra_trace::TraceBlockGuard::new_entered(
-            concat!($stage, "::done"),
-            $crate::format_node!($node),
-            module_path!(),
-        );
-        trace_event!($stage, $node);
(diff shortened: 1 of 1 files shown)


common/debug-interface/src/libra_trace.rs
```diff
@@ -80,21 +80,27 @@ macro_rules! send_logs {
...
     ($stage:expr, $node:tt) => {
-        let trace_guard = $crate::libra_trace::TraceBlockGuard::new_entered(
-            concat!($stage, "::done"),
-            $crate::format_node!($node),
-            module_path!(),
-        );
-        trace_event!($stage, $node);
+        if $crate::is_selected($crate::node_sampling_data!($node)) {
+            let node = $crate::format_node!($node);
+            let trace_guard = $crate::libra_trace::TraceBlockGuard::new_entered(
```
andll, line 85: I suspect doing it this way will just drop guard right away, instead of keeping it for duration of calling call block

You probably want to do smth like:
```
let trace_guard = if $crate::is_selected($crate::node_sampling_data... {
...
Some(trace_guard)
else {
None
}
```

changyonghu: I got your point. Yes, the guard will drop after if block. For the other one, we push trace_guard into a vector, so I think it's fine within if block.

PR #4905, opened 2020-07-04
[language] Verifier cleanup

dariorussi:
Here we complete the major clean up of the verifier. 
There is a bit more to do to make everything more consistent but that 
is not for this PR which is already massive.
VerifiedModule/Script are completely removed, given that was the general feeling in the team.
*View are removed from the verifier in favor of ad-hoc structures needed for verification.
CompiledScript is verified without converting to module. That implies that each verification pass
has 2 entry points, one for `CompiledModule`, the other for `CompiledScript`.

The common API is in `binary_views.rs` which provides a common access to the binary
and a unified function view.

We also made dependencies check a pass like the others, which cleaned up things quite a bit


language/bytecode-verifier/src/binary_views.rs
```diff
@@ -0,0 +1,273 @@
...
+        StructHandleIndex,
+    },
+    CompiledModule,
+};
+
+// A `BinaryIndexedView` provides table indexed access for both `CompiledModule` and
+// `CompiledScript`.
+// Operations that are not allowed for `CompiledScript` return an error.
+// A typical use of a `BinaryIndexedView` is while resolving indexes in bytecodes.
+pub(crate) enum BinaryIndexedView<'a> {
```
tnowacki, line 29: Hooray enums!

I feel we lean on traits way way too much when we want "interface" or "object" like behavior. Enums behave just like parent/child classes, and it feels like the right application here

language/bytecode-verifier/src/binary_views.rs
```diff
@@ -0,0 +1,273 @@
...
+
+const EMPTY_SIGNATURE: &Signature = &Signature(vec![]);
+
+// A `FunctionView` holds all the information needed by the verifier for a
+// `FunctionDefinition` and its `FunctionHandle` in a single view.
+// A control flow graph is built for a function when the `FunctionView` is
+// created.
+// A `FunctionView` is created for all module functions except native functions.
+// It is also created for a script.
+pub struct FunctionView<'a> {
```
tnowacki, line 208: not pub crate?

language/bytecode-verifier/src/binary_views.rs
```diff
@@ -0,0 +1,273 @@
...
+                Err(VMStatus::new(StatusCode::INVALID_OPERATION_IN_SCRIPT))
+            }
+        }
+    }
+
+    // Return the `Kind` of a `SignatureToken` given a context.
+    // A `TypeParameter` used the kind of the `constraints`.
+    // `StructInstantiation` have to "merge" the kinds of the instantiation with that
+    // of the generic type.
+    pub(crate) fn kind(&self, ty: &SignatureToken, constraints: &[Kind]) -> Kind {
```
tnowacki, line 152: Has this code finally found a forever home...? I feel like this has been moved in everyone of your past few PRs I've reviewed :p

language/bytecode-verifier/src/check_duplication.rs
```diff
@@ -22,133 +27,177 @@ pub struct DuplicationChecker<'a> {
 }
 
 impl<'a> DuplicationChecker<'a> {
-    pub fn verify(module: &'a CompiledModule) -> VMResult<()> {
+    pub fn verify_module(module: &'a CompiledModule) -> VMResult<()> {
+        Self::check_identifiers(module.identifiers())?;
```
tnowacki, line 31: I usually like pattern matching so I don't forget to check anything. It doesn't work here obviously, but its something we could look into in the future. e.g.

```
let CompiledModuleTables { identifiers, module, constant_pool, signatures, module_handles, struct_handles, function_handles, function_instantiations } = module.view_tables();
...
```
It doesn't buy much, and might not be needed here. But I use a pattern like this in the source language compiler in translations, and it tends to really help me not forget things

language/bytecode-verifier/src/code_unit_verifier.rs
```diff
@@ -5,50 +5,84 @@
...
+            name_def_map.insert(fh.name, FunctionDefinitionIndex(idx as u16));
+        }
+        // verify
+        let code_unit_verifier = CodeUnitVerifier {
+            resolver,
+            function_view,
+            name_def_map,
+        };
+        code_unit_verifier.verify_common()?;
+        AcquiresVerifier::verify(module, function_definition)
```
tnowacki, line 78: While we are refactoring, the `AcquiresVerifier` should really be moved out into `verifier.rs` I was a noob when I added it, ands its definitely in the wrong spot

language/bytecode-verifier/src/binary_views.rs
dariorussi: LOL - it's so true!!!! 
I am actually liking this location, thoughts?

language/bytecode-verifier/src/code_unit_verifier.rs
dariorussi: I can do that

dariorussi: ok I'll do this in another PR because I want to check better if combine with something else

language/bytecode-verifier/src/binary_views.rs
tnowacki: I think its better than anywhere else I've seen so far

vgao1996: This also seems good to me. You'll always need an environment if you want to query the kind of a type so it makes sense to me to associate this with `BinaryIndexedView` (the env).

PR #4908, opened 2020-07-05
[move-prover] Added spec for Coin1

DavidLDill:
I wanted to see how far I could get with proving that only the
treasury compliance account can mint Coin1.

Unfortunately, I was able to prove it when it's not true! The problem
is a generic function in another module that can invalidate the invariant.

In the process, I changed spec_has_..._role functions in Roles.move to
take addresses instead of signers as arguments, because it seems to be
more appropriate in specifications.  The corresponding Move functions
all take signers as arguments, which has not been a problem, so far.

## Motivation

(Write your answer here.)

## Test Plan

## Related PRs

### DIFF
--- a/language/stdlib/modules/Coin1.move
+++ b/language/stdlib/modules/Coin1.move
@@ -3,6 +3,7 @@ address 0x1 {
 module Coin1 {
     use 0x1::Libra;
     use 0x1::FixedPoint32;
+    use 0x1::Roles;
 
     struct Coin1 { }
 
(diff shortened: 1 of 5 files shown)


language/stdlib/modules/Coin1.move
```diff
@@ -24,6 +25,39 @@ module Coin1 {
         Libra::publish_mint_capability<Coin1>(tc_account, coin1_mint_cap, tc_account);
         Libra::publish_burn_capability<Coin1>(tc_account, coin1_burn_cap, tc_account);
     }
+
+// **************** SPECIFICATIONS  ****************
+
+/// Only TreasuryComplianceRole may mint Coin1
+
+/// BUG: This property verifies, but it does not necessarily hold.
+/// Libra.move has a function `remove_mint_capability<CoinType>` that removes
```
sblackshear, line 34: I think there's no problem with getting rid of `remove_mint_capability` now that we don't use it in genesis anymore. We may need it later if treasury compliance needs the flexibility to mint the same coin from multiple accounts, but that is hypothetical. I would vote to eliminate!

CC @tzakian

wrwg:
Is this still up for review?

language/stdlib/modules/Coin1.move
tzakian: Ditto! Happy to see this function go. If we need to add it back in at a later date it's a code update, so not the worst thing in the world.

PR #4909, opened 2020-07-06
[crypto] add tiny-keccak x RustCrypto compat and remove RustCrypto sha3 dep

phlip9:
This diff adds a thin wrapper struct around `tiny_keccak::Sha3::v256` so that we can use tiny_keccak's SHA3-256 implementation in RustCrypto-compatible APIs (such as libra's HKDF implementation). It also removes the RustCrypto `sha3` crate as a dependency, since it is no longer needed.


mimoo:
I'm a bit late, but why not the other way around? Why not use sha-3 instead of tiny-keccak?

kchalkias:
@mimoo see benchmarks at the end of their readme doc (comparing against rust-crypto's sha3)
https://github.com/debris/tiny-keccak

mimoo:
interesting! Looks pretty similar to me, I find tiny-keccak pretty clean personally

PR #4916, opened 2020-07-06
[move framework] delete ed25519_threshold_verify

sblackshear:
This was unused and occupied a fair amount of code in signature.rs. Killing it also eliminates the VM dependency on the `bit_vec` crate.


vgao1996:
LGTM from the VM standpoint.

huitseeker:
Thanks folks, this is really making my life easier!

PR #4917, opened 2020-07-06
[state-synchronizer] remove Option qualifier from coordinator.waypoint

agc-quadra:
SyncCoordinator previously had an Option<Waypoint> field in order to support integration_test.rs;  this necessitated the addition of code paths that were test only and introduced unnecessary Option check code in production code paths.

This change (a) removes the option qualifier and (b) updates the tests to use default "0" values for waypoints.

## Motivation

SyncCoordinator has a legacy `Option<Waypoint>` field;  this change removes the Option qualifier.

In production paths, there is always a valid Waypoint provided.  The `None` values for SyncCoordinator were used only in tests and necessitated error checking logic that would never be exercised in production (but was fundamental to tests passing).

Here we remove Option and update the tests to provide Waypoint::default() instead of None.

yes

## Test Plan

Don't break existing tests, specifically state-synchronizer/src/tests/integration_test.rs

## Related PRs


agc-quadra:
@davidiw  poking you for triage/redirection to domain expert.

gregnazario:
> @davidiw @gregnazario poking you two for triage/redirection to domain experts.

David is out, and I'm unsure about waypoint (still kinda new), but I'll poke around and see.

As far as I can tell, it seems like what you're doing is a good idea, but I'm unsure if that will cause the configs to break / other side effects

agc-quadra:
> > @davidiw @gregnazario poking you two for triage/redirection to domain experts.
> 
> David is out, and I'm unsure about waypoint (still kinda new), but I'll poke around and see.
> 
> As far as I can tell, it seems like what you're doing is a good idea, but I'm unsure if that will cause the configs to break / other side effects

Yeah.  I suspect folks more intimately involved in that part of the code base will have a clearer perspective on the traps I'm undoubtably stepping into.

gregnazario:
@mgorven is there any situation that waypoint isn't set in our configs?  This change seems like it should clean up the mess around accessing an `Option<Waypoint>` arg, but it just matters whether we deliberately ever not provide a waypoint in configs.

davidiw:
I think we should consider relabeling Waypoint::default to Waypoint::mock, since there is no "default" Waypoint.

@bors-libra r+

mgorven:
> @mgorven is there any situation that waypoint isn't set in our configs? This change seems like it should clean up the mess around accessing an `Option<Waypoint>` arg, but it just matters whether we deliberately ever not provide a waypoint in configs.

We always set waypoints in pre-mainnet. As long as config-builder sets them there shouldn't be situations where they are not.

PR #4918, opened 2020-07-06
Add specs to ValidatorConfig.move

emmazzz:
## Motivation

This PR adds global and local specifications to ValidatorConfig.move.

Yes.

## Test Plan

cargo test

### DIFF
--- a/language/stdlib/modules/ValidatorConfig.move
+++ b/language/stdlib/modules/ValidatorConfig.move
@@ -10,9 +10,9 @@ module ValidatorConfig {
 
     struct Config {
         consensus_pubkey: vector<u8>,
-        // TODO(philiphayes): restructure
-        //   3) remove validator_network_identity_pubkey
-        //   4) remove full_node_network_identity_pubkey
+        /// TODO(philiphayes): restructure
+        ///   3) remove validator_network_identity_pubkey
+        ///   4) remove full_node_network_identity_pubkey
         validator_network_identity_pubkey: vector<u8>,
         validator_network_address: vector<u8>,
         full_node_network_identity_pubkey: vector<u8>,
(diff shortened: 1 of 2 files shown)


language/stdlib/modules/ValidatorConfig.move
```diff
@@ -149,5 +149,102 @@ module ValidatorConfig {
...
+        define spec_exists_config(addr: address): bool {
+            exists<ValidatorConfig>(addr)
+        }
+
+        /// Returns true if addr has an operator account.
+        define spec_has_operator(addr: address): bool {
+            Option::spec_is_some(global<ValidatorConfig>(addr).operator_account)
+        }
+
+        /// Returns the operator accounr of a validator if it has one,
```
wrwg, line 170: accounr

language/stdlib/modules/ValidatorConfig.move
```diff
@@ -149,5 +149,102 @@ module ValidatorConfig {
...
+            old(spec_is_valid(validator)) ==> spec_is_valid(validator);
+    }
+
+    spec module {
+        apply ValidatorStaysValid to *;
+    }
+
+    /// # Specifications for individual functions
+
+    spec fun publish {
```
wrwg, line 209: I would suggest to move the function specs right after the function. As well as any helper functions related to a given function `f` (like `define spec_f`). This makes it so much easier to maintain the code. It's already done in various places.

If you do so, you should use the following ordering:

```
fun foo(...): ... {
}
spec fun foo {
   // Having this first will give doc generator the context where to associate the following specs with
   ...
}
spec schema FooS {
   // Any schemas belonging to this function
}
...
spec module {
   // Any spec functions related to this function
   define spec_foo(...): ... { ... }
}
...
```

```

PR #4919, opened 2020-07-06
[backup] run in cluster-test

msmouse:
## Motivation

To evaluate performance hit imposed by background backup operation.

yes
## Test Plan

cti --pr 4919 --run bench -- --backup

One of the nodes indeed kept backingi up the state snapshot:

image

Performance was not affected (kept at ~1070 tps)

### DIFF
--- a/testsuite/cluster-test/src/experiments/performance_benchmark.rs
+++ b/testsuite/cluster-test/src/experiments/performance_benchmark.rs
@@ -10,11 +10,12 @@ use crate::{
     tx_emitter::EmitJobRequest,
     util::unix_timestamp_now,
 };
-use anyhow::Result;
+use anyhow::{anyhow, Result};
 use async_trait::async_trait;
 use debug_interface::trace::LibraTraceClient;
 use futures::{future::try_join_all, join};
 use libra_logger::{info, warn};
+use rand::{rngs::ThreadRng, seq::SliceRandom};
 use serde_json::Value;
 use std::{
     collections::HashSet,
@@ -22,6 +23,7 @@ use std::{
     time::Duration,
 };
 use structopt::StructOpt;
+use tokio::task::JoinHandle;
 
 #[derive(StructOpt, Debug)]
 pub struct PerformanceBenchmarkParams {
@@ -46,6 +48,11 @@ pub struct PerformanceBenchmarkParams {
     pub duration: u64,
     #[structopt(long, help = "Set fixed tps during perf experiment")]
     pub tps: Option<u64>,
+    #[structopt(
+        long,
+        help = "Whether benchmark should pick one node to run DB backup."
+    )]
+    pub backup: bool,
 }
 
 pub struct PerformanceBenchmark {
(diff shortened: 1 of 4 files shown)


testsuite/cluster-test/src/instance.rs
```diff
@@ -283,6 +284,9 @@ impl Instance {
             .arg("-c")
             .arg(command)
             .kill_on_drop(true)
+            .stdout(Stdio::null())
+            .stderr(Stdio::null())
+            .stdin(Stdio::null())
```
msmouse, line 289: @andll  Otherwise it pollutes output of the `cti` tool.

testsuite/cluster-test/src/experiments/performance_benchmark.rs
```diff
@@ -218,6 +232,39 @@ impl Experiment for PerformanceBenchmark {
...
+            .clone();
+
+        const COMMAND: &str = "while true; do \
+            /opt/libra/bin/db-backup one-shot backup \
+            --max-chunk-size 1073741824 --backup-service-port 7777 \
+            --state-version $(/opt/libra/bin/db-backup one-shot query  --backup-service-port 7777 --latest-version | sed -n 's/latest-version: //p') \
+            local-fs --dir $(mktemp -d -t libra_backup_XXXXXXXX); \
+            done";
+
+        Ok(Some(tokio::spawn(async move {
```
andll, line 255: We seem not to get any signal in test on whether backup failed.
So if it start failing in nightly or in commit check, we will never know.

We can, for example, peek join handle in before dropping it to see if async fn completed

testsuite/cluster-test/src/instance.rs
andll: For some use cases we actually want to see stdout/err.
I wonder if we can make add boolean argument(like tee_output) to exec function and set it to false in backup, but keep it in other use case?

testsuite/cluster-test/src/experiments/performance_benchmark.rs
msmouse: @andll  even if it keeps looping, there can still be issues. I would output backup throughput from the backup service going forward and monitor that.

testsuite/cluster-test/src/instance.rs
msmouse: k. sure.

testsuite/cluster-test/src/experiments/performance_benchmark.rs
andll: I see. Do you mean watching for prometheus metric, or something else?

andll: (You can also just pull metrics endpoint from the instance.debug_client())

andll:
@bors-libra delegate+

This PR looks good. Before we can include it into suites, we need to find a way to monitor backup

msmouse:
@bors-libra r=andll

Thanks @andll  I'll land this first to get the metrics for our dashboard. Will follow up with monitoring backup throughput.

PR #4924, opened 2020-07-06
[libra framework] comments and specs for add_currency_to_account

sblackshear:
- Adding thorough comments and a prover spec for `add_currency_to_account`.
- Adding aborts with meaningful error codes for `ADD_EXISTING_CURRENCY` and `NOT_A_CURRENCY`
- Attempted to fully specify abort behavior, but ran into some problems (see inline comments). Went with a partial spec for now.


language/stdlib/transaction_scripts/add_currency_to_account.move
```diff
@@ -1,9 +1,29 @@
...
 use 0x1::LibraAccount;
+// Prover deps:
+use 0x1::Libra;
+use 0x1::Signer;
 
-/// Add the currency identified by the type `currency` to the sending accounts.
-/// Aborts if the account already holds a balance fo `currency` type.
+/// Add a `Currency` balance to `account`, which will enable `account` to send and receive
+/// `Libra<Currency>`.
+/// Aborts with NOT_A_CURRENCY if `Currency` is not an accepted currency type in the Libra system
```
tzakian, line 9: Nit: might want to make these bullet points to make the markdown docs prettier.

language/stdlib/transaction_scripts/add_currency_to_account.move
```diff
@@ -1,9 +1,29 @@
...
+    pragma verify = true;
+    pragma aborts_if_is_partial = true;
+    // `Currency` must be valid
+    aborts_if !Libra::spec_is_currency<Currency>();
+    // Can't have existing balance in `Currency`
+    aborts_if exists<LibraAccount::Balance<Currency>>(Signer::spec_address_of(account));
+    // TODO(shb): Prover claims that there's no abort in this situation. I suspect we need to
+    // rewrite VASP::try_allow_currency to be more amenable to verification
+    // If child VASP, parent needs `AccountLimits<Currency>`
+    //aborts_if VASP::spec_is_child_vasp(Signer::spec_address_of(account)) && !exists<AccountLimits::Window<Currency>>(VASP::spec_parent_address(Signer::spec_address_of(account)));
```
tzakian, line 27: Have you tried having this look for the `LimitsDefintion<Currency>` as opposed to the `Window`? The only reason I ask is the `try_allow_currency` looks at this resource as opposed the `Window` resource in the `if`.

language/stdlib/transaction_scripts/add_currency_to_account.move
```diff
@@ -1,9 +1,29 @@
...
+/// Aborts with `LibraAccount::ADD_EXISTING_CURRENCY` if the account already holds a balance in
+/// `Currency`.
+/// Aborts with `LibraAccount::PARENT_VASP_CURRENCY_LIMITS_DNE` if `account` is a `ChildVASP` whose
+/// parent does not have an `AccountLimits<Currency>` resource.
 fun add_currency_to_account<Currency>(account: &signer) {
     LibraAccount::add_currency<Currency>(account);
 }
+spec fun add_currency_to_account {
+    pragma verify = true;
+    pragma aborts_if_is_partial = true;
```
wrwg, line 19: I understand why you'd use this, but just for clarification, goal should be eventually to avoid this. We want that all aborts are covered, unless they are programmer errors. Otherwise the aborts_if specs have diminished value. To distinguish programmer from user errors, we would like to see a partitioning of the error space at some point in the future. Also would like to have the convention that any implicit aborts should be considered a programmer error. E.g.:

```
assert(!exists<R>(a), Errrors::RESOURCE_EXISTS);
move_to(a, R{...}); /// This would produce Errors::UNEXPECTED_RESOURCE_EXISTS or similar
```

language/stdlib/transaction_scripts/add_currency_to_account.move
sblackshear: I tried that and also tried rewriting the code to only publish a `Window` under the parent VASP (hence our messenger convo this morning :)) and then check for that here, but I still couldn't get it to work... The counterexample I got at that point had a state where an `ChildVASP` resource  at address *A* had a `parent_address` that also pointed to *A*. I figured it would be hard to teach the prover why that couldn't happen, so decided to punt for later...

language/stdlib/transaction_scripts/add_currency_to_account.move
sblackshear: Agreed with both conventions! It hurt me to use `aborts_if_partial` because I was very interested to see what other error cases I might be missing...

PR #4925, opened 2020-07-06
[network] ConnectionOrigin added to NewPeer events

gregnazario:
In order to differentiate between incoming new peers and outgoing new
peers, this has been added.  It helps distinguish between the types of
connections we need to have limits on.  It also helps us determine whether a peer is upstream or downstream for mempool.

Note: This doesn't adjust mempool / state-sync.  I spent some time trying to do it, and then decided to limit the scope down and follow up on the next PR.

Also, this doesn't address the limits from ConnectivityManager, and the limits would still be shared at this point which will also be followed up on.


christinacelee:
lgtm from application layer consuming this new flag, will let networking side take a look

phlip9:
cool, lgtm!

@bors-libra r+

PR #4926, opened 2020-07-07
[libra framework] per-module & sequential error codes

tzakian:
This PR moves to per-module sequential error codes defined in local constants.

Tests needed to be updated to look for the new error codes.

Also updated the transaction timeout to be `u64::max / 1_000_000` instead of `i64::max / 1_000_000`. 

Right now `LibraSystem` will time out in the prover if run as part of the testsuite. However, if you run the test separately it will pass.


tnowacki:
Do we want to wait until the new VM error stuff is in to land this?

tzakian:
I would prefer to land this ASAP since this will accumulate merge conflicts very quickly.

Issue #4206, opened 2020-06-02
[consensus] simplify PendingVote


zekun000:
PendingVote was designed to support votes in any round, and as we move to the RoundState world, PendingVote is per round instance.

We could simplify how it works by assuming all the votes/timeout inside is for the same round now.

mimoo:
this would mean moving PendingVotes to the internal state of a RoundState right? Right now it is in block store

zekun000:
@mimoo what revision are you looking at? it's part of the RoundState since long time ago

mimoo:
Ah indeed! I got tricked seeing that the file is in `block_storage/`! There's a number of things that can be simplified here then :)

mimoo:
So here's what I propose:

- [ ] make the logic work with the new assumption that we are at a specific/fixed round https://github.com/libra/libra/pull/4927
- [ ] simplify the logic to make it easy to match it to a spec (by inlining functions inside of `insert_vote()` https://github.com/libra/libra/pull/4927
- [ ] pass epoch+round as arguments to insert_vote()
- [ ] move out of `block_storage/`

PR #4927, opened 2020-07-07
[consensus] simplifying pending vote logic

mimoo:
Now that a pending vote store is linked to a specific RoundState,
it cannot contain votes that are for a different (epoch, round) tuple.

Thus, we can greatly simplify the logic.

Notably:

* I do not replace an equivocating vote with the new one anymore (simpler logic for something that had no proven benefits).
* I get rid of a panic when `validator_verifier.check_voting_power()` returns an error we don't know of.
* I inline every functions into `insert_vote()` to make it easier to follow along the specification.

the second commit:

* moves pending_votes out of block_store (it doesn't have anything to do with blocks)
* moves all the pending_vote_tests into pending_votes (there's only two tests)


consensus/consensus-types/src/vote.rs
```diff
@@ -62,6 +62,16 @@ impl Vote {
         }
     }
 
+    /// Returns true if the two votes are equivalent. This does not take into account the timeout signature.
+    pub fn equivalent_to(&self, other: &Self) -> bool {
+        let mut equivalent = true;
```
zekun000, line 67: I think you can just compare ledger_info.hash, `verify` already guarantees other fields hash matches consensus_data_hash inside

consensus/src/block_storage/pending_votes.rs
```diff
@@ -1,12 +1,15 @@
 // Copyright (c) The Libra Core Contributors
 // SPDX-License-Identifier: Apache-2.0
 
+//! PendingVotes store pending votes observed for a fixed epoch and round.
```
zekun000, line 4: use `///` instead

mimoo:
cool! I'll start working on making tests pass, and move this out of block_storage as per https://github.com/libra/libra/issues/4206#event-3521192785

consensus/consensus-types/src/vote.rs
mimoo: OK I see, the only field that can still differ is the signature but this needs to be tolerated because of malleability

Screen Shot 2020-07-07 at 8 33 23 PM

mimoo:
> you probably want to split into smaller functions

100 LOC with comments, I don't think it's that bad : o
imo it makes it easier to read along the specification.

consensus/src/block_storage/pending_votes.rs
mimoo: why? That's a module-level doc : o

zekun000: hmm, isn't this documenting the module?

zekun000: oh i see, https://learning-rust.github.io/docs/a5.comments_and_documenting_the_code.html TIL

consensus/src/pending_votes.rs
```diff
@@ -0,0 +1,365 @@
...
+    /// There might be some issues adding a vote
+    ErrorAddingVote(VerifyError),
+    /// The vote is not for the current round.
+    UnexpectedRound(u64, u64),
+}
+
+/// A PendingVotes structure keep track of votes
+pub struct PendingVotes {
+    /// Maps LedgerInfo digest to associated signatures (contained in a partial LedgerInfoWithSignatures).
+    /// This might keep multiple LedgerInfos (for the current round) to tolerate non-determinism in execution.
```
zekun000, line 49: this is not to tolerate non-determinism, a round can have a good proposal and a NIL proposal

mimoo: yeah actually execution should always be deterministic oO (I just copied that comment), IIUC it is possible that if the proposer creates different proposals, we end up receiving timeout votes that carry different proposals as well right? In this case we would save them under different li_digest_to_votes

zekun000: > then we can end up receiving timeout votes that carry different proposals as well right?

yup

mimoo: ok I'll modify the comment according to this!

mimoo: which makes me think, we should log the receiving of different non-NIL proposals for the same round

PR #4928, opened 2020-07-07
[LCS] Add a limit to recursion depth

matbd:
## Motivation

Last batch of LCS improvements:
* Prevent stack overflows by setting a limit on the depth of LCS data.
* (nit) Set MAX_SEQUENCE_LENGTH = 1 << 31 - 1 instead of 1<< 31 to accomodate Java arrays.

The notion of "depth" suggested in this PR consists in counting named boundaries i.e. struct and enum nodes (but not tuple nodes for instance). This is meant to minimize code changes and computation while still preventing unbounded stack allocations. (Indeed, recursive formats only happens through recursive structs and enums.)

The current code costs some extra memory for serialization but (at least on my mac) when I increase the limit I still get a stack overflow first for deserialization. By the way, the limit is around 900 for my mac with the default stack settings and using single lists as a benchmark.

As a follow up, we should probably design a test to create the most stack-expensive Libra transaction authorized by the specs and allow validators to adjust their stack sizes.

## Test Plan

CI

### DIFF
--- a/common/lcs/src/de.rs
+++ b/common/lcs/src/de.rs
@@ -38,7 +38,7 @@ pub fn from_bytes<'a, T>(bytes: &'a [u8]) -> Result<T>
 where
     T: Deserialize<'a>,
 {
-    let mut deserializer = Deserializer::new(bytes);
+    let mut deserializer = Deserializer::new(bytes, crate::MAX_CONTAINER_DEPTH);
     let t = T::deserialize(&mut deserializer)?;
     deserializer.end().map(move |_| t)
 }
@@ -48,7 +48,7 @@ pub fn from_bytes_seed<'a, T>(seed: T, bytes: &'a [u8]) -> Result<T::Value>
 where
     T: DeserializeSeed<'a>,
 {
-    let mut deserializer = Deserializer::new(bytes);
+    let mut deserializer = Deserializer::new(bytes, crate::MAX_CONTAINER_DEPTH);
     let t = seed.deserialize(&mut deserializer)?;
     deserializer.end().map(move |_| t)
 }
@@ -56,13 +56,17 @@ where
 /// Deserialization implementation for LCS
 struct Deserializer<'de> {
     input: &'de [u8],
+    max_remaining_depth: usize,
 }
 
 impl<'de> Deserializer<'de> {
     /// Creates a new `Deserializer` which will be deserializing the provided
     /// input.
-    fn new(input: &'de [u8]) -> Self {
-        Deserializer { input }
+    fn new(input: &'de [u8], max_remaining_depth: usize) -> Self {
+        Deserializer {
+            input,
+            max_remaining_depth,
+        }
     }
 
     /// The `Deserializer::end` method should be called after a type has been
@@ -173,6 +177,18 @@ impl<'de> Deserializer<'de> {
     fn parse_string(&mut self) -> Result<&'de str> {
         let slice = self.parse_bytes()?;
         std::str::from_utf8(slice).map_err(|_| Error::Utf8)
+    }
+
+    fn enter_named_container(&mut self, name: &'static str) -> Result<()> {
+        if self.max_remaining_depth == 0 {
+            return Err(Error::ExceededContainerDepthLimit(name));
+        }
+        self.max_remaining_depth -= 1;
+        Ok(())
+    }
+
+    fn leave_named_container(&mut self) {
+        self.max_remaining_depth += 1;
     }
 }
 
@@ -333,18 +349,24 @@ impl<'de, 'a> de::Deserializer<'de> for &'a mut Deserializer<'de> {
         visitor.visit_unit()
     }
 
-    fn deserialize_unit_struct<V>(self, _name: &'static str, visitor: V) -> Result<V::Value>
+    fn deserialize_unit_struct<V>(self, name: &'static str, visitor: V) -> Result<V::Value>
     where
         V: Visitor<'de>,
     {
-        self.deserialize_unit(visitor)
+        self.enter_named_container(name)?;
+        let r = self.deserialize_unit(visitor);
+        self.leave_named_container();
+        r
     }
 
-    fn deserialize_newtype_struct<V>(self, _name: &'static str, visitor: V) -> Result<V::Value>
+    fn deserialize_newtype_struct<V>(self, name: &'static str, visitor: V) -> Result<V::Value>
     where
         V: Visitor<'de>,
     {
-        visitor.visit_newtype_struct(self)
+        self.enter_named_container(name)?;
+        let r = visitor.visit_newtype_struct(&mut *self);
+        self.leave_named_container();
+        r
     }
 
     fn deserialize_seq<V>(mut self, visitor: V) -> Result<V::Value>
@@ -364,14 +386,17 @@ impl<'de, 'a> de::Deserializer<'de> for &'a mut Deserializer<'de> {
 
     fn deserialize_tuple_struct<V>(
         mut self,
-        _name: &'static str,
+        name: &'static str,
         len: usize,
         visitor: V,
     ) -> Result<V::Value>
     where
         V: Visitor<'de>,
     {
-        visitor.visit_seq(SeqDeserializer::new(&mut self, len))
+        self.enter_named_container(name)?;
+        let r = visitor.visit_seq(SeqDeserializer::new(&mut self, len));
+        self.leave_named_container();
+        r
     }
 
     fn deserialize_map<V>(mut self, visitor: V) -> Result<V::Value>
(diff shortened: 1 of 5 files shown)


bob-wilson:
@vgao1996 We should make sure the depth checks you're planning at the Move level are consistent with this.

vgao1996:
@bob-wilson sounds good. I'll likely set a limit below the LCS limit for both types and values.

matbd:
> @bob-wilson sounds good. I'll likely set a limit below the LCS limit for both types and values.

@vgao1996 For a simple (but not super efficient) solution, you may call `lcs::serialized_size` to verify that a newly created object is "LCS-compliant".

common/lcs/src/de.rs
```diff
@@ -48,21 +48,25 @@ pub fn from_bytes_seed<'a, T>(seed: T, bytes: &'a [u8]) -> Result<T::Value>
...
     input: &'de [u8],
+    max_remaining_depth: usize,
 }
 
 impl<'de> Deserializer<'de> {
     /// Creates a new `Deserializer` which will be deserializing the provided
     /// input.
-    fn new(input: &'de [u8]) -> Self {
-        Deserializer { input }
+    fn new(input: &'de [u8], max_remaining_depth: usize) -> Self {
```
huitseeker, line 65: I suspect you want to check that max_remaining_depth > 0.
Since the last argument will be difficult to parametrize for newcomers (or people who don't read docs) *and* this is safety-sensitive, would it make sense to have it be an `Into<Option<usize>>`? That way uncertain caller could pass `None`? Alternatively, would it be wise to implement `Default`?
I really want to avoid people putting a Very Large Number there because they don't know what a good value looks like. And I'm nearly rempted to recommend comparing `max_remaining_depth` with `MAX_OBJECT_DEPTH`, and issuing a warning when the former is bigger.
Same remark for `Serializer` ofc.

common/lcs/src/de.rs
```diff
@@ -174,6 +178,18 @@ impl<'de> Deserializer<'de> {
         let slice = self.parse_bytes()?;
         std::str::from_utf8(slice).map_err(|_| Error::Utf8)
     }
+
+    fn enter_named_container(&mut self, name: &'static str) -> Result<()> {
+        if self.max_remaining_depth == 0 {
```
huitseeker, line 183: <= 0 ?

common/lcs/src/error.rs
```diff
@@ -13,8 +13,10 @@ pub enum Error {
     Eof,
     #[error("I/O error: {0}")]
     Io(String),
-    #[error("exceeded max sequence length")]
+    #[error("exceeded max sequence length: {0}")]
     ExceededMaxLen(usize),
+    #[error("exceeded max object depth while entering: {0}")]
+    ExceededObjectDepthLimit(&'static str),
```
huitseeker, line 19: "object" depth? O_o What are those "objects" you speak of? :p

common/lcs/src/error.rs
```diff
@@ -13,8 +13,10 @@ pub enum Error {
...
     #[error("expected boolean")]
```
huitseeker, line 20: What I'm getting at is this is a recursion limit (i.e. an implementation constraint) rather than a statement on a predefined notion of structural depth that would exist in Rust independently of LCS  / Serde details.
How about "recursion depth" wherever you use "object depth", especially in the initial doc comment of lib.rs?

common/lcs/src/lib.rs
```diff
@@ -50,6 +50,24 @@
...
 //!
+//! ## Recursion and Depth of LCS Data
+//!
+//! Recursive data-structures (e.g. trees) are allowed. However, because of the possibility of stack
+//! overflow during (de)serialization, the *depth* of any valid LCS data cannot exceed the value `MAX_OBJECT_DEPTH`.
+//! Formally, we define *depth* as the number of named containers ---that is,
+//! in Rust, the number of struct and enum boundaries--- that are crossed during (de)serialization.
+//!
+//! This definition aims to minimize the number of operations while ensuring that
+//! (de)serialization of a known LCS format cannot cause arbitrary stack allocations.
```
huitseeker, line 61: "arbitrary large"?

common/lcs/src/ser.rs
```diff
@@ -91,22 +91,26 @@ where
...
 }
 
 impl<'a, W> Serializer<'a, W>
 where
     W: std::io::Write,
 {
     /// Creates a new `Serializer` which will emit LCS.
-    fn new(output: &'a mut W) -> Self {
-        Self { output }
+    fn new(output: &'a mut W, max_remaining_depth: usize) -> Self {
```
huitseeker, line 109: I suspect you want to check that max_remaining_depth > 0.

common/lcs/src/ser.rs
```diff
@@ -132,6 +136,14 @@ where
         }
         self.output_u32_as_uleb128(len as u32)
     }
+
+    fn enter_named_container(&mut self, name: &'static str) -> Result<()> {
+        if self.max_remaining_depth == 0 {
```
huitseeker, line 141: <= 0 ?

common/lcs/src/ser.rs
```diff
@@ -132,6 +136,14 @@ where
...
+            return Err(Error::ExceededObjectDepthLimit(name));
+        }
+        self.max_remaining_depth -= 1;
```
huitseeker, line 144: Could one initialize `max_remaining_depth` with a large negative number, and make this underflow to stg large on first iteration? Do you want to use `checked_sub`?

common/lcs/src/ser.rs
```diff
@@ -239,7 +251,8 @@ where
         Ok(())
     }
 
-    fn serialize_unit_struct(self, _name: &'static str) -> Result<()> {
+    fn serialize_unit_struct(mut self, name: &'static str) -> Result<()> {
+        self.enter_named_container(name)?;
         self.serialize_unit()
```
huitseeker, line 256: I'm slightly worried about the maintenability of not having a `leave_named_container` in the recursive calls in `Serializer` (e.g. here). While I suspect that just manifests all your calls occur in terminal position (so that you only pop a stack frame when you're done), I'm wondering what would happen if that was to change (e.g. because you restructure seq / tuple serialization). Wouldn't you have to refactor every "named" method here, potentially leading to a bug if one of them is forgotten?
It's possible I may just be overthinking this.

common/lcs/src/de.rs
matbd: @huitseeker These constructors are not `pub`.

common/lcs/src/de.rs
matbd: not sure why that would be better

common/lcs/src/error.rs
matbd: Good question. I'm fine removing "object" everywhere.

common/lcs/src/lib.rs
matbd: sure

common/lcs/src/ser.rs
matbd: This function is not `pub`. Also, `0` could actually make sense for testing.

common/lcs/src/ser.rs
matbd: `max_remaining_depth ` is a `usize` so there are no negative values.

`checked_sub` is a good idea but it returns an option so I was not sure about if the generated assembly was as good. (Let me check)

common/lcs/src/ser.rs
matbd: So, initially I tried adding a call `self.leave_named_container()` at the end and the borrow checker made it clear that it makes no sense to modify something after use :)

To make this PR, I searched for `name:` and changed the (5) callbacks. It definitely seems like the right approach to me.

common/lcs/src/ser.rs
huitseeker: You're correct ... in debug mode. 
https://play.rust-lang.org/?version=stable&mode=release&edition=2018&gist=4543f51666cebea902bd639690440012

huitseeker: :facepalm: Indeed, you're playing within ℤ/18446744073709551615ℤ and with decrements of size 1 you could only wrap to a larger number from 0, which is explicitly excluded, my bad.

huitseeker: :facepalm: Indeed, you're playing within ℤ/(18446744073709551615ℤ) and with decrements of size 1 you could only wrap to a larger number from 0, which is explicitly excluded, my bad.

common/lcs/src/de.rs
huitseeker: :facepalm:

common/lcs/src/ser.rs
huitseeker: :facepalm:

common/lcs/src/de.rs
huitseeker: Do we expect every validator that would ever run LCS serialization to have/need the same stack size limits ? 

I suspect not.

If not, do we plan to get all validators on the same recursion depth limit ? If so, would that be the lowest common denominator among those validators, and would we be ok with the implication on max transaction/block/what-have-you size over a long period of time? How would we handle an upgrade?

I suspect we'll rather allow distinct validators to run with different values for this parameter (under controlled conditions). What I'm getting at is this constructor will probably not remain private forever.

matbd: In this PR, the depth limit is part of the definition of LCS, so it is the same everywhere.

matbd: I don't see how to do otherwise by the way, since we want to deterministic execution etc.

matbd: Although in practice, it's probably fine to make this value larger in rare occasions.

common/lcs/src/error.rs
matbd: fixed

common/lcs/src/lib.rs
matbd: fixed

PR #4929, opened 2020-07-07
[state sync] optimistically choose target LI from pending LI store

christinacelee:
## Motivation

When selecting a target LI from the pending LI store to form a FN chunk request, the state sync coordinator was choosing the target LI based on the state after verifying the LI, writing to and syncing with storage. To optimize for perf, this PR changes the behavior of crafting FN chunk requests s.t. the chunk request is formed before `validate_and_store_chunk` happens under the assumption that the chunk response was processed successfully. 

## Test Plan

Existing tests (especially `test_sync_pending_ledger_infos`)

### DIFF
--- a/state-synchronizer/src/coordinator.rs
+++ b/state-synchronizer/src/coordinator.rs
@@ -31,8 +31,8 @@ use libra_types::{
 };
 use network::protocols::network::Event;
 use std::{
-    collections::{BTreeMap, HashMap},
-    ops::Bound::Included,
+    collections::{BTreeMap, Bound, HashMap},
+    ops::Bound::{Excluded, Included},
     time::{Duration, SystemTime, UNIX_EPOCH},
 };
 use tokio::time::{interval, timeout};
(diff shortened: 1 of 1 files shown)


zekun000:
does this really affect perf? a storage sync is supposed to be very cheap

christinacelee:
I can try running cluster test, but I assumed that was the whole motivation of it for the pre-existing optimistic approach? @dmitri-perelman can maybe give insight?

zekun000:
from what I remember, the original optimistic fetching is to send the request **before** `validate_and_store_chunk` which makes sense since it parallels the execution and network latency.

christinacelee:
ah i mixed up storage read/write - will update description/code accordingly

dmitri-perelman:
Haha, I don't recognize this code anymore. Indeed, we used to have the optimistic fetching feature of sending the next chunk request before the processing of the received chunk. It was our "poor man pipelining". I don't think we've run any perf benchmark to check whether it's actually making things faster.

state-synchronizer/src/coordinator.rs
```diff
@@ -102,31 +102,58 @@ impl PendingLedgerInfos {
...
+            // first, try to find LI with max version that will fit in a single chunk
+            let li_within_single_chunk = self
+                .pending_li_queue
+                .range((Excluded(sync_version), Included(sync_version + chunk_limit)))
                 .rev()
                 .next()
-                .map(|(_version, ledger_info)| ledger_info.clone())
+                .map(|(_version, ledger_info)| ledger_info.clone());
+
+            if li_within_single_chunk.is_some() {
```
zekun000, line 140: you can use `or_else` here, but the logic in this function is so fancy. I'd rather just maintain a FIFO queue we request for.

PR #4933, opened 2020-07-07
Removing concurrent preburns

moeziniaf:
##  Motivation

There are some vulnerabilities with our current 'preburn request vector' on chain vs how AOS handles burns which could lead to unexpected behaviour like the wrong Coin being burned. Preburn area should not be concurrent, only handle one preburn request at a time.
Move prover/spec changes also made here to reflect preburn area change (still some left over TODOs in BurnAbortsIf spec)

## Test Plan

functional test preburn_burn.move tests no concurrent preburns and preburn request limits


language/stdlib/modules/Libra.move
```diff
@@ -480,38 +482,38 @@ module Libra {
...
     public fun cancel_burn_with_capability<CoinType>(
         preburn_address: address,
         _capability: &BurnCapability<CoinType>
     ): Libra<CoinType> acquires CurrencyInfo, Preburn {
-        // destroy the coin at the head of the preburn queue
+        // destroy the coin in the preburn area
         let preburn = borrow_global_mut<Preburn<CoinType>>(preburn_address);
-        let coin = Vector::remove(&mut preburn.requests, 0);
+        let coin = withdraw_all<CoinType>(&mut preburn.request);
+        deposit(&mut preburn.request, zero<CoinType>());
```
sblackshear, line 516: Delete? This is depositing a coin with value zero into a coin with value zero :).

language/stdlib/modules/Libra.move
```diff
@@ -445,21 +446,22 @@ module Libra {
...
+    /// `CoinType` is empty (error code 7).
     public fun burn_with_resource_cap<CoinType>(
         preburn: &mut Preburn<CoinType>,
         preburn_address: address,
         _capability: &BurnCapability<CoinType>
     ) acquires CurrencyInfo {
         let currency_code = currency_code<CoinType>();
-        // destroy the coin at the head of the preburn queue
-        let Libra { value } = Vector::remove(&mut preburn.requests, 0);
+        // Throw if no coin present in preburn area
```
sblackshear, line 461:
```suggestion
        // Abort if no coin present in preburn area
```

language/move-lang/functional-tests/tests/libra/preburn_burn.move
```diff
@@ -55,12 +76,30 @@ use 0x1::Coin1::Coin1;
...
 // check: BurnEvent
 // check: EXECUTED
+
+// Preburn allowed but larger than balance
+//! new-transaction
+//! sender: dd
+//! gas-currency: Coin1
+script {
+    use 0x1::Coin1::Coin1;
+    // use 0x1::Libra;
```
sblackshear, line 95: Delete?

language/move-lang/functional-tests/tests/transaction_fee/burn_collected_fees.move
```diff
@@ -34,6 +36,5 @@ fun burn_txn_fees<CoinType>(blessed_account: &signer) {
     TransactionFee::burn_fees<CoinType>(blessed_account);
 }
 }
-// check: PreburnEvent
-// check: BurnEvent
-// check: EXECUTED
+
+// check: ABORTED
```
sblackshear, line 40: Check error code also?

language/stdlib/modules/Libra.move
```diff
@@ -171,11 +170,10 @@ module Libra {
     /// contains this address has the authority to initiate a burn request. A burn request can be
     /// resolved by the holder of a `BurnCapability` by either (1) burning the funds, or (2)
     /// returning the funds to the account that initiated the burn request.
-    /// including multiple burn requests from the same account. However, burn requests
-    /// (and cancellations) from the same account must be resolved in FIFO order.
+    /// Concurrent preburn requests are not allowed, only one request can be handled at any time.
     resource struct Preburn<CoinType> {
-        /// The queue of pending burn requests
-        requests: vector<Libra<CoinType>>,
+        /// A single pending burn amount
```
sblackshear, line 175: Say that there's no request pending if the value is 0?

language/stdlib/modules/Libra.move
```diff
@@ -171,11 +170,10 @@ module Libra {
...
+        request: Libra<CoinType>,
```
sblackshear, line 176: We might want to call this something like `to_burn` now?

wrwg:
This was submitted with verification failures. Please DO NOT do this. If the MIRAI bot fails, the reason has to be investigated, and if there is not easy solution, the minimal solution is to inform the prover team and turn verification off for the according failures.

moeziniaf:
Note: ran the prover tests locally and they all pass, so I went for it. Discussed with @DavidLDill offline, I put the relevant TODOs in order to add back relevant specs later

wrwg:
Have you setup the prover tools following the instructions in language/move-prover/scripts and after that configured the environment variables as described in the README there? Because if you don't all your tests will silently pass. The system is setup to skip tests if tools are not configured instead of failing. This is to create no burden for dev platform folks which aren't working on libra framework. This is from the old days of the prover where things weren't very stable yet; we are now ready to change this, but had higher priorities with the looming deadlines.

In any case, if you submit a PR, and the MIRAI bot complains on it, tests are failing. It does not say anything if tests pass. So the rule of thumb is if there is a MIRAI bot failure *after* your last commit, something is wrong.

@DavidLDill

PR #4935, opened 2020-07-07
[client] adding two standard field on all JSONRPC response

thefallentree:
These two field is needed to assist stateful client to detect stale response from server side.

### DIFF
--- a/json-rpc/src/runtime.rs
+++ b/json-rpc/src/runtime.rs
@@ -8,6 +8,9 @@ use crate::{
 };
 use futures::future::join_all;
 use libra_config::config::{NodeConfig, RoleType};
+use libra_json_rpc_types::views::{
+    JSONRPC_LIBRA_LEDGER_TIMESTAMPUSECS, JSONRPC_LIBRA_LEDGER_VERSION,
+};
 use libra_mempool::MempoolClientSender;
 use libra_types::ledger_info::LedgerInfoWithSignatures;
 use serde_json::{map::Map, Value};
@@ -87,7 +90,8 @@ async fn rpc_endpoint(
     let ledger_info = service
         .get_latest_ledger_info()
         .map_err(|_| reject::custom(DatabaseError))?;
-    if let Value::Array(requests) = data {
+
+    let resp = Ok(if let Value::Array(requests) = data {
         // batch API call
         let futures = requests.into_iter().map(|req| {
             rpc_request_handler(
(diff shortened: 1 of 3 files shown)


json-rpc/types/src/views.rs
```diff
@@ -26,6 +26,9 @@ use serde::{Deserialize, Serialize};
 use std::{collections::BTreeMap, convert::TryFrom};
 use transaction_builder::get_transaction_name;
 
+pub const JSONRPC_LIBRA_VERSION: &str = "libra-version";
```
andll, line 29: Should we call it something like 'ledger_version' / 'ledger_timestampused'?
When I first seen libra-version I thought it means version of the node software

thefallentree: yeah we can do that, but google json style guide actually advocate for camel case naming, do you think we should do snake_case or camelCase ?

most of the other api name is snake_case, I think we should do it here too

andll: Oh, I am ok with either snake or camel, I just meant to call prefix 'ledger', not 'libra', because libra-version sounds more like 'version of libra software'. So I think 'ledger-version', better represents what this field contains, then 'libra-version'

thefallentree: yeah I got that, but think we still want libra_ prefix, as an namespace  to avoid potential collision , so I'm proposaling libra_ledger_version  and libra_ledger_timestampusecs instead.

andll: That sounds good to me

thefallentree:
@xli @andll  can you guys take another look?

json-rpc/src/runtime.rs
```diff
@@ -116,10 +122,20 @@ async fn rpc_request_handler(
...
     response.insert("jsonrpc".to_string(), Value::String("2.0".to_string()));
     response.insert("id".to_string(), Value::Null);
+    response.insert(
+        JSONRPC_LIBRA_LEDGER_VERSION.to_string(),
+        Value::Number(version.into()),
+    );
+    response.insert(
+        JSONRPC_LIBRA_LEDGER_TIMESTAMPUSECS.to_string(),
+        Value::Number(timestamp.into()),
+    );
```
xli, line 138: 2 concerns:
1. We found the version and timestamp before execute, this is confusing as response, because I thought it is version and timestamp after method is executed.
2. this does not match jsonrpc spec standard, which means, it's possible a standard jsonrpc client may not be able to parse these data out, if it only returns result / error from response.

A nit concern is get_metadata also response these 2 values, the response will be odd.

xli: resolved offline

aching:
@thefallentree and @xli, the spec needs to be updated as well.

thefallentree:
@aching yeah, will send additional PR for that.

PR #4936, opened 2020-07-07
[network] Simplify structured logging for networking

gregnazario:
This removes a lot of the extra code that was being copied around
between structured logging lines, and will help decrease the amount of
code and work needed to add more.

I modeled this after @mimoo 's security log changes

### DIFF
--- a/common/debug-interface/src/libra_trace.rs
+++ b/common/debug-interface/src/libra_trace.rs
@@ -71,6 +71,7 @@ macro_rules! send_logs {
         let log_entry = $crate::json_log::JsonLogEntry::new($name, $json);
         $crate::json_log::send_json_log(log_entry.clone());
         libra_logger::send_struct_log!(libra_logger::StructuredLogEntry::new_named(
+            $crate::libra_trace::LIBRA_TRACE,
             $crate::libra_trace::LIBRA_TRACE
         )
         .data($name, log_entry));
--- a/common/logger/src/security.rs
+++ b/common/logger/src/security.rs
@@ -18,9 +18,9 @@ use crate::StructuredLogEntry;
 
 /// helper function to create a security log
 pub fn security_log(name: &'static str) -> StructuredLogEntry {
-    StructuredLogEntry::new_named(&name)
-        // set the category to security
-        .security()
+    StructuredLogEntry::new_named("security", &name)
+        // set the level to critical
+        .critical()
     // set the error description
 }
 
--- a/network/src/logging.rs
+++ b/network/src/logging.rs
@@ -1,34 +1,62 @@
 // Copyright (c) The Libra Core Contributors
 // SPDX-License-Identifier: Apache-2.0
 
-use crate::{
-    peer_manager::{ConnectionNotification, ConnectionRequest, PeerManagerRequest},
-    ConnectivityRequest,
-};
+//!
+//! This module is to contain all networking logging information.
+//!
+//! ```
+//! use libra_config::network_id::NetworkContext;
+//! use libra_logger::prelude::*;
+//! use network::logging::*;
+//!
+//! send_struct_log!(
+//!   network_log(network_events::CONNECTIVITY_MANAGER_LOOP, &NetworkContext::mock())
+//!     .data(network_events::TYPE, network_events::START)
+//!     .field(network_events::EVENT_ID, &5)
+//! );
+//! ```
+//!
+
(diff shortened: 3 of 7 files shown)


network/src/logging.rs
```diff
@@ -1,34 +1,62 @@
...
+
 use libra_config::network_id::NetworkContext;
-use libra_logger::LoggingField;
-use libra_types::PeerId;
+use libra_logger::StructuredLogEntry;
+
+/// A helper function to cut down on a bunch of repeated network struct log code
+pub fn network_log(label: &'static str, network_context: &NetworkContext) -> StructuredLogEntry {
+    StructuredLogEntry::new_named(label)
+        .data("category", "network")
```
andll, line 25: This is interesting. I originally thought we won't need category at all and can just use auto-identified crate name.
But at the same time, I guess single 'logical' category can span across multiple crates, so perhaps category is needed.

In this case, I would modify StructuredLogEntry and add `category:Option<&'static str>` and then change new_named to `new_named(category: &'static str, name: &'static str)`, e.g. basically always require category on named log entries.

Do you mind including this small refactoring in your pr(I can also do it myself, if you prefer me to do it)

mimoo: so we should probably move security to a category as well : o

gregnazario: Sure I'll take a look into it.

My thoughts on how this would work changed over time.  I was confused why there was a "name" field, until I realized that we want to probably label groupings of logs.  Then I realized, that as a whole we should probably categorize them.

We could use the crate name, but I just figured this would make it simple in something like Kibana to filter "category"="network" or "name"="connectivity_manager_loop".

Thoughts?

gregnazario: Yeah, sounds like it I'll go throw in security as well with a category.

andll: Yeah, I think we should just make a category field required.
After this we should also remove 'security' as a severity level.
Instead, for security logs we should have category=security and severity=critical

common/debug-interface/src/libra_trace.rs
```diff
@@ -71,7 +71,8 @@ macro_rules! send_logs {
         let log_entry = $crate::json_log::JsonLogEntry::new($name, $json);
         $crate::json_log::send_json_log(log_entry.clone());
         libra_logger::send_struct_log!(libra_logger::StructuredLogEntry::new_named(
-            $crate::libra_trace::LIBRA_TRACE
+            $crate::libra_trace::LIBRA_TRACE,
```
andll, line 74: This should probably be `($crate::libra_trace::LIBRA_TRACE, $crate::libra_trace::LIBRA_TRACE)`, those are actually machine queries right now, so we should not change name on this entry

andll:
@bors-libra delegate+
One nit on libra-trace change, but other then that looks great, thanks for doing this @gregnazario !

common/debug-interface/src/libra_trace.rs
gregnazario: Yeah I stared at this trying to figure out if that duplicate would be better, seemed weird to me but let's do it.

andll:
@bors-libra r=andll
^ normally you would use this to merge after delegate

gregnazario:
> @bors-libra r=andll
> ^ normally you would use this to merge after delegate

woops

common/logger/src/security.rs
```diff
@@ -18,9 +18,9 @@ use crate::StructuredLogEntry;
 
 /// helper function to create a security log
 pub fn security_log(name: &'static str) -> StructuredLogEntry {
-    StructuredLogEntry::new_named(&name)
-        // set the category to security
-        .security()
+    StructuredLogEntry::new_named("security", &name)
```
mimoo, line 21: I was wondering if it wouldn't be better to have a file listing all the categories as constants

gregnazario: Yeah, that's I think a next step.  It's only being used in 2 places right now, so we can easily add it / centralize it later.

gregnazario:
> 💔 Test failed - [checks-circle_commit_workflow](https://circleci.com/workflow-run/2095c251-1ce6-4540-a98e-75bee5d2ed76?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-checks-link)

Fun story, `lint` and `xclippy` don't catch compilation problems in the comments, but `x test` does!

PR #4937, opened 2020-07-07
[libra framework] Separate freezing out into separate module, add freezing to VASPs

tzakian:
This pulls account freezing out into its own module. The idea being that this module provides the "atomic semantics" of what it means for an account to be frozen. Other modules (e.g., VASP) can then use these semantics to provide a more semantically rich version of "freezing" e.g., if the parent account is frozen all of the child accounts are frozen, but if the child account is frozen then only that account is viewed as frozen.

Additionally, the libra root account, and TC account cannot be frozen. This is ensured in the `AccountFreezing` module

Won't land until all spec changes have passed.


json-rpc/types/src/views.rs
```diff
@@ -70,7 +70,6 @@ pub struct AccountView {
     pub received_events_key: BytesView,
     pub delegated_key_rotation_capability: bool,
     pub delegated_withdrawal_capability: bool,
-    pub is_frozen: bool,
```
tzakian, line 73: This is a breaking change to the json-rpc endpoint.

thefallentree:
instead of removing is_frozen attribute, can you add logic to implement that in the view ? Otherwise we would need an new API for it again.

tzakian:
Shouldn't be an issue to implement in the view. I'll update.

language/stdlib/modules/AccountFreezing.move
```diff
@@ -0,0 +1,115 @@
...
+    public fun freeze_account(
+        account: &signer,
+        frozen_address: address,
+    )
+    acquires FreezingBit, FreezeEventsHolder {
+        assert(Roles::has_treasury_compliance_role(account), ENOT_ABLE_TO_FREEZE);
+        let initiator_address = Signer::address_of(account);
+        // The libra root account and TC cannot be frozen
+        assert(frozen_address != CoreAddresses::LIBRA_ROOT_ADDRESS(), ECANNOT_FREEZE_LIBRA_ROOT);
+        assert(frozen_address != CoreAddresses::TREASURY_COMPLIANCE_ADDRESS(), ECANNOT_FREEZE_TC);
```
sblackshear, line 68: Do we want to check an address or a role here (thinking of a future where there are multiple TC addresses...). Same question for the previous line, though perhaps less important given that there is supposed to be only one LR account...

language/stdlib/modules/VASP.move
```diff
@@ -260,6 +261,15 @@ module VASP {
         }
     }
 
+    /// A VASP account is frozen if itself is frozen, or if its parent account is frozen.
+    public fun is_frozen(addr: address): bool
```
sblackshear, line 265: Nice!!!!

language/stdlib/modules/LibraAccount.move
```diff
@@ -970,7 +887,8 @@ module LibraAccount {
         // Verify that the transaction sender's account exists
         assert(exists_at(transaction_sender), EPROLOGUE_ACCOUNT_DNE);
 
-        assert(!account_is_frozen(transaction_sender), EPROLOGUE_ACCOUNT_FROZEN);
+        assert(!AccountFreezing::account_is_frozen(transaction_sender), EPROLOGUE_ACCOUNT_FROZEN);
+        assert(!VASP::is_frozen(transaction_sender), EPROLOGUE_ACCOUNT_FROZEN);
```
sblackshear, line 891: Might be worth adding a comment explaining why there's a separate check for VASPs.

language/stdlib/modules/AccountFreezing.move
tzakian: Good question! I would like a role check here, however all of the role check functions take a `&signer` whereas we only have an `address` at this point. So we could update the `has_role_*` functions to take an address argument instead of a `&signer`, but doing that does make those functions a bit more "loose" in a sense (since you no longer have the direct connection to authority over the account), but I don't think it would be too bad.

moeziniaf:
> Looks great to me from the Move perspective! The only thing I'm wondering is whether moving the freeze/unfreeze events to a new module is going to create churn for AOS (@moeziniaf?)

Thanks! I believe we're just tailing the chain for Preburn events right now so doesn't seem to be an issue from our end

language/stdlib/modules/VASP.move
```diff
@@ -260,6 +261,15 @@ module VASP {
...
+    acquires ChildVASP {
+        is_vasp(addr) && (
+            AccountFreezing::account_is_frozen(parent_address(addr)) ||
+            AccountFreezing::account_is_frozen(addr)
```
moeziniaf, line 269: 👍

PR #4938, opened 2020-07-07
[libra-trace] Add seperate binary for libra trace.

changyonghu:
Create separate binary to retrieve libra trace from elastic search.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2925,6 +2925,19 @@ dependencies = [
 ]
 
 [[package]]
+name = "libra-trace"
+version = "0.1.0"
+dependencies = [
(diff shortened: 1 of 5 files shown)


changyonghu:
Now I just call trace client from debug_interface. I will refactor some part of debug_interface into monitoring in separate PR.

common/monitoring/Cargo.toml
```diff
@@ -0,0 +1,20 @@
+[package]
+name = "monitoring"
```
andll, line 2: Let's call it libra-trace to be specific? It is not clear whether this will become a general monitoring crate, or there will be some other layout in future

changyonghu: Yes, I will refactor them in a separate PR.

PR #4941, opened 2020-07-07
[CI] cache sccache binary

sausagee:
## Motivation
Cache sccache binary to trim another 1.5 minutes from total workflow time.

sccache size is increased to 4GB to accomodate the dev build which generates larger cache volumne.

The first workflow of the day will miss the cache.  The cache miss costs ~3 minutes due to installing the sccache binary and is absorbed in the first job, `prefetch-crates`.  The subsequent jobs in the workflow can avoid installing sccache.

image

Subsequent workflows will hit the cache and can enjoy the saving of sccache install time ~1.5 mintues. Plus extra saving due to increased sccache size.   

image

## Test Plan
CI

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -90,14 +90,26 @@ commands:
           command: rustup install nightly
   install_sccache:
     steps:
+      - restore_cache:
+          name: Restore sccache binary
+          key: sccache-0.2.13
       - run:
           name: Install scccache
           command: |
-            cargo install sccache
-            echo 'export SCCACHE_CACHE_SIZE=3G' >> $BASH_ENV
+            if ! [ -e /usr/local/cargo/bin/sccache ]; then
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -140,18 +144,19 @@ commands:
     steps:
       - save_cache:
           name: Save cargo package cache
-          key: cargo-package-cache-{{ checksum "Cargo.lock" }}
+          key: cargo-package-cash-{{ checksum "Cargo.lock" }}
```
huitseeker, line 147: Are you sure?

sausagee: lol, definitely not. The PR is in draft, and I need to test it in CI in-place.  So I deliberately changed it to cause a cache miss.  Might as well name it $_$ :)

sausagee:
SG. One more iteration to put sccache in its own cache area.

PR #4945, opened 2020-07-08
[move framework] add add_all_currencies flag to DD account creation

sblackshear:
Enable creation of multi-currency DD accounts using an `add_all_currencies` flag in the relevant tx script. This is identical to the approach used in the creation of parent VASP accounts.


ankushagarwal:
Thanks @sblackshear!

PR #4949, opened 2020-07-08
[language] loader cleanup

dariorussi:
Refactor code a bit to organize it better for readability.
Added some commments.
Cleaned up `Loader` and `Resolver` API.
There is more coming but this is a good change

### DIFF
--- a/language/move-vm/runtime/src/loader.rs
+++ b/language/move-vm/runtime/src/loader.rs
@@ -334,6 +334,10 @@ impl TypeCache {
     }
 }
 
+//
+// Loader
+//
+
 // A Loader is responsible to load scripts and modules and holds the cache of all loaded
 // entities. Each cache is protected by a `Mutex`. Operation in the Loader must be thread safe
 // (operating on values on the stack) and when cache needs updating the mutex must be taken.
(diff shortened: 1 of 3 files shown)


dariorussi:
this is not a big change but moving code around is reporting a much bigger number of line changed that it really is

language/move-vm/runtime/src/loader.rs
```diff
@@ -386,10 +371,105 @@ impl Loader {
...
+    // Script verification steps.
+    // See `verify_module()` for module verification steps.
+    fn verify_script(&self, script: &CompiledScript) -> VMResult<()> {
+        DuplicationChecker::verify_script(&script)?;
+        SignatureChecker::verify_script(&script)?;
+        InstructionConsistency::verify_script(&script)?;
+        constants::verify_script(&script)?;
+        CodeUnitVerifier::verify_script(&script)?;
+        verify_main_signature(&script)
+    }
```
tnowacki, line 422: This looks like the same code in verifier.rs? Why the duplication?

dariorussi: I want the runtime to have its own concept of verification and I do not want to use a common function.
Both `verify_module` and `verify_script` are implemented by the runtime.
Maybe I could point that out in the comment but it really seem too much information.
Bottom line I do not want the runtime to share anything with anybody when it comes to core logic.
Of course that has a lot of not so nice side effects together with a lot of very nice, and truth is, I don't know what's right.
But I really do not want to share for now... I want to be really selfish here... ;) 
thoughts?

dariorussi:
there is one more PR coming for the loader and built on top of this. 
If people think is better I am happy to merge it here.
Also I am fine to wait going in if people have bad merges, please let me know. Happy to wait.
I am cleaning up the loader on top of this and it's all internal API so I don't have to go in right away if I am disruptive to others

vgao1996:
@dariorussi Sorry tried to land the wrong PR... Will get the flow canceled

vgao1996:
Still I'm surprised bors-libra permitted this...

dariorussi:
> Still I'm surprised bors-libra permitted this...

this is hilarious!!!!

language/move-vm/runtime/src/loader.rs
```diff
@@ -690,32 +742,69 @@ impl<'a> Resolver<'a> {
...
             BinaryType::Script(script) => script.function_instantiation_at(idx.0),
+        };
+        let mut instantiation = vec![];
+        for ty in &func_inst.instantiation {
+            instantiation.push(ty.subst(type_params)?);
         }
+        Ok(instantiation)
+    }
+
+    pub(crate) fn generic_function_arity(
```
vgao1996, line 792: I found this name a bit confusing. Upon first glance, "arity" could mean a few things to me:
1. The total number of parameters, including type parameters and value parameters
2. The number of value parameters.
3. The number of type parameters.
and it was only clear until I read the body of the function.

I'd personally prefer a more straightforward name including sth like "num_type_params".

dariorussi: went for `type_params_count`

dariorussi:
some feedback

PR #4950, opened 2020-07-08
[network][breaking][1/2] add encrypted network address

phlip9:
Encrypting the on-chain network addresses is purely a defense-in-depth mitigation to minimize attack surface and reduce DDoS attacks on the validators by restricting the visibility of their public-facing network addresses only to other validators.

These encrypted network addresses are intended to be stored on-chain under each validator's advertised network addresses in their `ValidatorConfig`s. All validators share the secret `shared_val_netaddr_key`, though each validator's addresses are encrypted using a per-validator `derived_key`.

We use per-validator `derived_key`s to limit the blast radius of nonce reuse to each validator, i.e., a validator that accidentally reuses a nonce will only leak information about their network addresses or `derived_key`.

```txt
derived_key := HKDF-SHA3-256::extract_and_expand(
    salt=HKDF_SALT,
    ikm=shared_val_netaddr_key,
    info=account_address,
    output_length=32,
)
```

A raw network address, `addr`, is then encrypted using AES-256-GCM.

```txt
enc_addr := AES-256-GCM::encrypt(
    key=derived_key,
    nonce=seq_num || addr_idx,
    ad=key_version,
    message=addr,
)
```

Note: the last 16 bytes of `enc_addr` contains the authentication tag.

In order to reduce the probability of nonce reuse, validators should use the sequence number of the rotation transaction in the `seq_num` field.

The `EncNetworkAddress` struct also contains a `key_version` field, which identifies the specific `shared_val_netaddr_key` used to encrypt/decrypt the `EncNetworkAddress`.

Note: to help split this feature into smaller diffs, we push out the encrypted network addresses on-chain but only using a constant test key. Subsequent diffs will also add the corresponding key to the key manager, load it on startup, and use it in the places where the test key is currently used.

### DIFF
--- a/config/management/src/validator_config.rs
+++ b/config/management/src/validator_config.rs
@@ -76,14 +76,13 @@ impl ValidatorConfig {
                     validator_address, e
                 ))
             })?;
-        let enc_validator_address = raw_validator_address
-            .encrypt(&root_key, key_version, &sender, sequence_number, addr_idx)
-            .map_err(|e| {
-                Error::UnexpectedError(format!(
-                    "error encrypting validator address: \"{}\", error: {}",
-                    validator_address, e
-                ))
-            })?;
+        let enc_validator_address = raw_validator_address.encrypt(
+            &root_key,
+            key_version,
+            &sender,
+            sequence_number,
+            addr_idx,
+        );
         let raw_enc_validator_address = RawEncNetworkAddress::try_from(&enc_validator_address)
             .map_err(|e| {
                 Error::UnexpectedError(format!(
--- a/config/management/src/validator_config.rs
+++ b/config/management/src/validator_config.rs
@@ -12,7 +12,10 @@ use libra_crypto::{ed25519::Ed25519PublicKey, x25519, ValidCryptoMaterial};
 use libra_global_constants::{
     CONSENSUS_KEY, FULLNODE_NETWORK_KEY, OPERATOR_KEY, VALIDATOR_NETWORK_KEY,
 };
-use libra_network_address::{NetworkAddress, RawNetworkAddress};
+use libra_network_address::{
+    encrypted::{RawEncNetworkAddress, TEST_ROOT_KEY, TEST_ROOT_KEY_VERSION},
+    NetworkAddress, RawNetworkAddress,
+};
 use libra_secure_storage::{CryptoStorage, KVStorage, Storage, Value};
 use libra_secure_time::{RealTimeService, TimeService};
 use libra_types::{
@@ -45,6 +48,20 @@ impl ValidatorConfig {
         let validator_network_key = x25519_from_storage(VALIDATOR_NETWORK_KEY, &local_storage)?;
         let operator_key = ed25519_from_storage(OPERATOR_KEY, &local_storage)?;
 
+        // TODO(davidiw): In genesis this is irrelevant -- afterward we need to obtain the
+        // current sequence number by querying the blockchain.
+        let sequence_number = 0;
+
+        // TODO(davidiw): This is currently not supported
+        // let sender = self.owner_address;
+        let sender = account_address::from_public_key(&operator_key);
+
+        // TODO(philiphayes): actually get the real keys from storage
+        let root_key = TEST_ROOT_KEY;
+        let key_version = TEST_ROOT_KEY_VERSION;
+        // only supports one address for now
+        let addr_idx = 0;
+
         // append ln-noise-ik and ln-handshake protocols to base network addresses
 
         let validator_address = self
@@ -53,6 +70,11 @@ impl ValidatorConfig {
             .append_prod_protos(validator_network_key, HANDSHAKE_VERSION);
         let raw_validator_address = RawNetworkAddress::try_from(&validator_address)
             .map_err(|e| Error::UnexpectedError(format!("(raw_validator_address) {}", e)))?;
+        let enc_validator_address = raw_validator_address
+            .encrypt(&root_key, key_version, &sender, sequence_number, addr_idx)
+            .map_err(|e| Error::UnexpectedError(format!("(enc_validator_address) {}", e)))?;
+        let raw_enc_validator_address = RawEncNetworkAddress::try_from(&enc_validator_address)
+            .map_err(|e| Error::UnexpectedError(format!("(raw_enc_validator_address) {}", e)))?;
 
         let fullnode_address = self
             .fullnode_address
@@ -63,24 +85,16 @@ impl ValidatorConfig {
 
         // Step 2) Generate transaction
 
-        // TODO(davidiw): This is currently not supported
-        // let sender = self.owner_address;
-        let sender = account_address::from_public_key(&operator_key);
-
         // TODO(philiphayes): remove network identity pubkey field from struct when
         // transition complete
         let script = transaction_builder::encode_set_validator_config_script(
             sender,
             consensus_key.to_bytes().to_vec(),
             validator_network_key.to_bytes(),
-            raw_validator_address.into(),
+            raw_enc_validator_address.into(),
             fullnode_network_key.to_bytes(),
             raw_fullnode_address.into(),
         );
-
-        // TODO(davidiw): In genesis this is irrelevant -- afterward we need to obtain the
-        // current sequence number by querying the blockchain.
-        let sequence_number = 0;
         let expiration_time = RealTimeService::new().now() + constants::TXN_EXPIRATION_SECS;
         let raw_transaction = RawTransaction::new_script(
             sender,
--- a/config/management/src/validator_config.rs
+++ b/config/management/src/validator_config.rs
@@ -13,7 +13,9 @@ use libra_global_constants::{
     CONSENSUS_KEY, FULLNODE_NETWORK_KEY, OPERATOR_KEY, VALIDATOR_NETWORK_KEY,
 };
 use libra_network_address::{
-    encrypted::{RawEncNetworkAddress, TEST_ROOT_KEY, TEST_ROOT_KEY_VERSION},
+    encrypted::{
+        RawEncNetworkAddress, TEST_SHARED_VAL_NETADDR_KEY, TEST_SHARED_VAL_NETADDR_KEY_VERSION,
+    },
     NetworkAddress, RawNetworkAddress,
 };
 use libra_secure_storage::{CryptoStorage, KVStorage, Storage, Value};
@@ -56,9 +58,6 @@ impl ValidatorConfig {
         // let sender = self.owner_address;
         let sender = account_address::from_public_key(&operator_key);
 
-        // TODO(philiphayes): actually get the real keys from storage
-        let root_key = TEST_ROOT_KEY;
-        let key_version = TEST_ROOT_KEY_VERSION;
         // only supports one address for now
         let addr_idx = 0;
 
@@ -77,8 +76,8 @@ impl ValidatorConfig {
                 ))
             })?;
         let enc_validator_address = raw_validator_address.encrypt(
-            &root_key,
-            key_version,
+            &TEST_SHARED_VAL_NETADDR_KEY,
+            TEST_SHARED_VAL_NETADDR_KEY_VERSION,
             &sender,
             sequence_number,
             addr_idx,
--- a/config/management/src/validator_config.rs
+++ b/config/management/src/validator_config.rs
@@ -63,25 +63,45 @@ impl ValidatorConfig {
         let addr_idx = 0;
 
         // append ln-noise-ik and ln-handshake protocols to base network addresses
+        // and encrypt the validator address.
 
         let validator_address = self
             .validator_address
             .clone()
             .append_prod_protos(validator_network_key, HANDSHAKE_VERSION);
-        let raw_validator_address = RawNetworkAddress::try_from(&validator_address)
-            .map_err(|e| Error::UnexpectedError(format!("(raw_validator_address) {}", e)))?;
+        let raw_validator_address =
+            RawNetworkAddress::try_from(&validator_address).map_err(|e| {
+                Error::UnexpectedError(format!(
+                    "error serializing validator address: \"{}\", error: {}",
+                    validator_address, e
+                ))
+            })?;
         let enc_validator_address = raw_validator_address
             .encrypt(&root_key, key_version, &sender, sequence_number, addr_idx)
-            .map_err(|e| Error::UnexpectedError(format!("(enc_validator_address) {}", e)))?;
+            .map_err(|e| {
+                Error::UnexpectedError(format!(
+                    "error encrypting validator address: \"{}\", error: {}",
+                    validator_address, e
+                ))
+            })?;
         let raw_enc_validator_address = RawEncNetworkAddress::try_from(&enc_validator_address)
-            .map_err(|e| Error::UnexpectedError(format!("(raw_enc_validator_address) {}", e)))?;
+            .map_err(|e| {
+                Error::UnexpectedError(format!(
+                    "error serializing encrypted validator address: {:?}, error: {}",
+                    enc_validator_address, e
+                ))
+            })?;
 
         let fullnode_address = self
             .fullnode_address
             .clone()
             .append_prod_protos(fullnode_network_key, HANDSHAKE_VERSION);
-        let raw_fullnode_address = RawNetworkAddress::try_from(&fullnode_address)
-            .map_err(|e| Error::UnexpectedError(format!("(raw_fullnode_address) {}", e)))?;
+        let raw_fullnode_address = RawNetworkAddress::try_from(&fullnode_address).map_err(|e| {
+            Error::UnexpectedError(format!(
+                "error serializing fullnode address: \"{}\", error: {}",
+                fullnode_address, e
+            ))
+        })?;
 
         // Step 2) Generate transaction
 
--- a/network/network-address/src/encrypted.rs
+++ b/network/network-address/src/encrypted.rs
@@ -147,18 +147,9 @@ pub struct EncNetworkAddress {
     enc_addr: Vec<u8>,
 }
 
-/// Possible errors when parsing a human-readable [`NetworkAddress`].
 #[derive(Error, Debug)]
-pub enum Error {
-    #[error("error encrypting network address")]
-    EncryptError,
-
-    #[error("error decrypting network address")]
-    DecryptError,
-
-    #[error("ciphertext is too small to even contain the auth tag: length: {0}")]
-    CiphertextTooSmall(usize),
-}
+#[error("error decrypting network address")]
+pub struct DecryptError;
 
 //////////////////////////
 // RawEncNetworkAddress //
@@ -208,6 +199,9 @@ impl Arbitrary for RawEncNetworkAddress {
 ///////////////////////
 
 impl EncNetworkAddress {
+    /// ### Panics
+    ///
+    /// encrypt will panic if `addr` length > 64 GiB.
     pub fn encrypt(
         addr: RawNetworkAddress,
         root_key: &RootKey,
@@ -215,7 +209,7 @@ impl EncNetworkAddress {
         account: &AccountAddress,
         seq_num: u64,
         addr_idx: u32,
-    ) -> Result<Self, Error> {
+    ) -> Self {
         // unpack the RawNetworkAddress into its base Vec<u8>
         let mut addr_vec: Vec<u8> = addr.into();
 
@@ -236,18 +230,20 @@ impl EncNetworkAddress {
         let ad_slice = &ad_buf[..];
 
         // encrypt the raw network address in-place
+        // note: this can technically panic if the serialized network address
+        //       length is > 64 GiB
         let auth_tag = aead
             .encrypt_in_place_detached(nonce_slice, ad_slice, &mut addr_vec)
-            .map_err(|_| Error::EncryptError)?;
+            .expect("addr.len() must be <= 64 GiB");
 
         // append the authentication tag
         addr_vec.extend_from_slice(auth_tag.as_slice());
 
-        Ok(Self {
+        Self {
             root_key_version,
             seq_num,
             enc_addr: addr_vec,
-        })
+        }
     }
 
     pub fn decrypt(
@@ -255,7 +251,7 @@ impl EncNetworkAddress {
         root_key: &RootKey,
         account: &AccountAddress,
         addr_idx: u32,
-    ) -> Result<RawNetworkAddress, Error> {
+    ) -> Result<RawNetworkAddress, DecryptError> {
         let root_key_version = self.root_key_version;
         let seq_num = self.seq_num;
         let mut enc_addr = self.enc_addr;
@@ -263,7 +259,7 @@ impl EncNetworkAddress {
         // ciphertext is too small to even contain the authentication tag, so it
         // must be invalid.
         if enc_addr.len() < AES_GCM_TAG_LEN {
-            return Err(Error::CiphertextTooSmall(enc_addr.len()));
+            return Err(DecryptError);
         }
 
         let account_key = Self::derive_account_key(root_key, account);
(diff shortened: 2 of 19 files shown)


config/management/src/validator_config.rs
```diff
@@ -53,6 +70,11 @@ impl ValidatorConfig {
             .append_prod_protos(validator_network_key, HANDSHAKE_VERSION);
         let raw_validator_address = RawNetworkAddress::try_from(&validator_address)
             .map_err(|e| Error::UnexpectedError(format!("(raw_validator_address) {}", e)))?;
+        let enc_validator_address = raw_validator_address
+            .encrypt(&root_key, key_version, &sender, sequence_number, addr_idx)
+            .map_err(|e| Error::UnexpectedError(format!("(enc_validator_address) {}", e)))?;
+        let raw_enc_validator_address = RawEncNetworkAddress::try_from(&enc_validator_address)
```
gregnazario, line 76: Should we come up with a new name for these?

I found it confusing earlier with the:
* NetworkAddress (Uses Protocols)
* RawNetworkAddress (Uses encoded protocols)

And we've now added:
* RawEncNetworkAddress (Uses Encrypted, encoded protocols) but, for me looking at it with no knowledge might think it's just encoded.

It might not be for right now, but we should think of a way to simplify it or rename it, I don't have a good suggestion off the top of my head right now.

config/management/src/validator_config.rs
```diff
@@ -53,6 +70,11 @@ impl ValidatorConfig {
...
+            .map_err(|e| Error::UnexpectedError(format!("(raw_enc_validator_address) {}", e)))?;
```
gregnazario, line 77: Nit: "Unable to convert validator address" or something more descriptive?

gregnazario, line 77: I know this seems to be the Rust way to do it, but could we attach the error handling of a poorly formed NetworkAddress conversion to the converting function.  e.g. use RawEncNetworkAddress::from(&enc_validator_address), and put the mapping of error there to cut down on places we may reuse it.

config/management/src/validator_config.rs
phlip9: hmm, maybe "Raw" is the wrong prefix? would it be less confusing if it was like `SerNetworkAddress` (i.e., serialized network address)? personally, i think `EncryptedNetworkAddress` is getting a bit too long : ( maybe we shorten `NetworkAddress` to `NetAddr`?

one of the reasons for the thin `RawNetworkAddress`/`RawEncNetworkAddress` wrappers is to ensure we can't accidentally deserialize as the wrong thing.

for example, the on-chain representation of a `validator_network_address` in each `ValidatorConfig` is just a `vector<u8>` in Move, but we represent it as a `RawEncNetworkAddress` in rust so it's not possible (or at least harder) to accidentally deserialize it as something that's not an `EncNetworkAddress`.

in my mind the types sort of encode the transformations

NetworkAddress <--> RawNetworkAddress <--> EncNetworkAddress <--> RawEncNetworkAddress

gregnazario: Will we always know ahead of time if it's encrypted or not?

Maybe `Bin` instead of `Raw` to point out it's binary, but I don't think that part is really important.  

I guess my nit here is EncNetworkAddress isn't really an Encrypted NetworkAddress, but instead an encrypted RawNetworkAddress.

network/network-address/src/encrypted.rs
```diff
@@ -0,0 +1,482 @@
...
+use move_core_types::account_address::AccountAddress;
+use once_cell::sync::Lazy;
+#[cfg(any(test, feature = "fuzzing"))]
+use proptest::prelude::*;
+use serde::{Deserialize, Serialize};
+use std::{convert::TryFrom, mem};
+use thiserror::Error;
+
+/// The length in bytes of the AES-256-GCM authentication tag.
+pub const AES_GCM_TAG_LEN: usize = 16;
```
mimoo, line 23: btw I also set this in crypto/noise IIRC, maybe worth setting this in one place in crypto

phlip9: yeah i remember that, just thought it was clearer to have it nearby the actual code that uses it

network/network-address/src/encrypted.rs
```diff
@@ -0,0 +1,482 @@
...
+
+/// We salt the HKDF for deriving the account keys to provide application
+/// separation.
+///
+/// Note: modifying this salt is a backwards-incompatible protocol change.
+///
+/// For readers, this hash value is equal to the following hex string:
+/// `"dfc8ffcc7f62ea4e5b9bc41ee7969b44275419ebaad1db27d2a191b6d1db6d13"`
+pub static HKDF_SALT: Lazy<HashValue> =
+    Lazy::new(|| HashValue::sha3_256_of(b"LIBRA_ENCRYPTED_NETWORK_ADDRESS_SALT"));
```
mimoo, line 54: why hash the string?

network/network-address/src/encrypted.rs
```diff
@@ -0,0 +1,482 @@
...
+
+/// A serialized `EncNetworkAddress`.
+#[derive(Clone, Debug, Eq, PartialEq, Deserialize, Serialize)]
+pub struct RawEncNetworkAddress(#[serde(with = "serde_bytes")] Vec<u8>);
```
mimoo, line 58: why specify serde_bytes?

network/network-address/src/encrypted.rs
phlip9: iirc the HKDF salt should be the size of the hash output? also "nothing up my sleeve" constants and all that

network/network-address/src/encrypted.rs
```diff
@@ -0,0 +1,482 @@
...
+///
+/// Encrypting the on-chain network addresses is purely a defense-in-depth
+/// mitigation to minimize attack surface and reduce DDoS attacks on the validators
+/// by restricting the visibility of their public-facing network addresses only
+/// to other validators.
+///
+/// These encrypted network addresses are intended to be stored on-chain under
+/// each validator's advertised network addresses in their `ValidatorConfig`s.
+/// All validators share the secret `root_key`, though each validator's addresses
+/// are encrypted using a per-validator derived `account_key`.
```
mimoo, line 72: because we already have account keys in Libra, I suggest choosing a different name :P (I know it tripped me when we were talking about it)

network/network-address/src/encrypted.rs
phlip9: it makes the serialization/deserialization of bytes much more efficient. without the annotation, it will literally ser/de a single byte at a time, whereas with the annotation it will just memcpy the whole chunk in one go

network/network-address/src/encrypted.rs
phlip9: agghh naming is hard :'(

any suggestions for a better name? all i can think of is like "derived_key" or something

network/network-address/src/encrypted.rs
gregnazario: You could just put the string you commented directly in :)

network/network-address/src/encrypted.rs
```diff
@@ -0,0 +1,482 @@
...
+        root_key_version: RootKeyVersion,
+        account: &AccountAddress,
+        seq_num: u64,
+        addr_idx: u32,
+    ) -> Result<Self, Error> {
+        // unpack the RawNetworkAddress into its base Vec<u8>
+        let mut addr_vec: Vec<u8> = addr.into();
+
+        let account_key = Self::derive_account_key(root_key, account)?;
+        let aead = Aes256Gcm::new(GenericArray::from_slice(&account_key));
```
gregnazario, line 220: `aead`?  Maybe `cipher`?

network/network-address/src/encrypted.rs
```diff
@@ -0,0 +1,482 @@
...
+    seq_num: u64,
+    #[serde(with = "serde_bytes")]
+    enc_addr: Vec<u8>,
+}
+
+/// Possible errors when parsing a human-readable [`NetworkAddress`].
+#[derive(Error, Debug)]
+pub enum Error {
+    #[error("error encrypting network address")]
+    EncryptError,
```
mimoo, line 148: why would an EncryptError ever happen?

network/network-address/src/encrypted.rs
```diff
@@ -0,0 +1,482 @@
...
+
+    #[error("error decrypting network address")]
+    DecryptError,
+
+    #[error("ciphertext is too small to even contain the auth tag: length: {0}")]
+    CiphertextTooSmall(usize),
```
mimoo, line 154: shouldn't that just be a DecryptError? I don't think it's that useful info

network/network-address/src/encrypted.rs
```diff
@@ -0,0 +1,482 @@
...
+        let mut addr_vec: Vec<u8> = addr.into();
+
+        let account_key = Self::derive_account_key(root_key, account)?;
+        let aead = Aes256Gcm::new(GenericArray::from_slice(&account_key));
+
+        // nonce := seq_num || addr_idx; sizeof(nonce) == 12
+        // ex: seq_num = 0x1234, addr_idx = 0x04
+        //     ==> nonce_slice == &[0, 0, 0, 0, 0, 0, 0x12, 0x34, 0, 0, 0, 0x4]
+        let nonce = ((seq_num as u128) << 32) | (addr_idx as u128);
+        let nonce_buf = (nonce as u128).to_be_bytes();
```
gregnazario, line 226: This should already be a u128 nonce right?

network/network-address/src/encrypted.rs
```diff
@@ -0,0 +1,482 @@
...
+    EncryptError,
+
+    #[error("error decrypting network address")]
+    DecryptError,
+
+    #[error("ciphertext is too small to even contain the auth tag: length: {0}")]
+    CiphertextTooSmall(usize),
+
+    #[error("error deriving account key: {0}")]
+    DeriveAccountKeyError(#[from] HkdfError),
```
mimoo, line 157: can this really happen as well?

network/network-address/src/encrypted.rs
```diff
@@ -0,0 +1,482 @@
...
+/// identifies the specific `root_key` used to encrypt/decrypt the
+/// `EncNetworkAddress`.
+///
+/// TODO(philiphayes): expand this section
+#[derive(Clone, Debug, Eq, PartialEq, Deserialize, Serialize)]
+pub struct EncNetworkAddress {
+    root_key_version: RootKeyVersion,
+    seq_num: u64,
+    #[serde(with = "serde_bytes")]
+    enc_addr: Vec<u8>,
```
mimoo, line 141: I'm happier with full names like `encrypted_address`, but up to you

network/network-address/src/encrypted.rs
```diff
@@ -0,0 +1,482 @@
...
+    ) -> Result<Self, Error> {
+        // unpack the RawNetworkAddress into its base Vec<u8>
+        let mut addr_vec: Vec<u8> = addr.into();
+
+        let account_key = Self::derive_account_key(root_key, account)?;
+        let aead = Aes256Gcm::new(GenericArray::from_slice(&account_key));
+
+        // nonce := seq_num || addr_idx; sizeof(nonce) == 12
+        // ex: seq_num = 0x1234, addr_idx = 0x04
+        //     ==> nonce_slice == &[0, 0, 0, 0, 0, 0, 0x12, 0x34, 0, 0, 0, 0x4]
```
gregnazario, line 224: I think this is a little confusing of a comment, it's similar to below, but I guess what I'm seeing is:

each hex number is a 8 bit number.  So, if it's a u128, it should have 4 more 0's at the beginning.

network/network-address/src/encrypted.rs
```diff
@@ -0,0 +1,482 @@
...
+        let nonce = ((seq_num as u128) << 32) | (addr_idx as u128);
+        let nonce_buf = (nonce as u128).to_be_bytes();
+        let nonce_slice = &nonce_buf[mem::size_of::<u128>() - AES_GCM_NONCE_LEN..];
+        let nonce_slice = GenericArray::from_slice(nonce_slice);
+
+        // the root_key_version is in-the-clear, so we include it in the integrity check
+        // using the "additonal data"
```
gregnazario, line 231: *additional

network/network-address/src/encrypted.rs
```diff
@@ -0,0 +1,482 @@
...
+        let account_key = Self::derive_account_key(root_key, account)?;
+        let aead = Aes256Gcm::new(GenericArray::from_slice(&account_key));
+
+        // nonce := seq_num || addr_idx; sizeof(nonce) == 12
+        // ex: seq_num = 0x1234, addr_idx = 0x04
+        //     ==> nonce_slice == &[0, 0, 0, 0, 0, 0, 0x12, 0x34, 0, 0, 0, 0x4]
+        let nonce = ((seq_num as u128) << 32) | (addr_idx as u128);
+        let nonce_buf = (nonce as u128).to_be_bytes();
+        let nonce_slice = &nonce_buf[mem::size_of::<u128>() - AES_GCM_NONCE_LEN..];
+        let nonce_slice = GenericArray::from_slice(nonce_slice);
```
gregnazario, line 275: Perhaps put this in a function, since it's the same between encrypt / decrypt?

network/network-address/src/encrypted.rs
```diff
@@ -0,0 +1,482 @@
...
+        // unpack the RawNetworkAddress into its base Vec<u8>
+        let mut addr_vec: Vec<u8> = addr.into();
+
+        let account_key = Self::derive_account_key(root_key, account)?;
+        let aead = Aes256Gcm::new(GenericArray::from_slice(&account_key));
+
+        // nonce := seq_num || addr_idx; sizeof(nonce) == 12
+        // ex: seq_num = 0x1234, addr_idx = 0x04
+        //     ==> nonce_slice == &[0, 0, 0, 0, 0, 0, 0x12, 0x34, 0, 0, 0, 0x4]
+        let nonce = ((seq_num as u128) << 32) | (addr_idx as u128);
```
mimoo, line 225: I feel like it's less dangerous (and cleaner) to just convert to bytes, then concatenate

network/network-address/src/encrypted.rs
```diff
@@ -0,0 +1,482 @@
...
+impl TryFrom<&RawEncNetworkAddress> for EncNetworkAddress {
+    type Error = lcs::Error;
+
+    fn try_from(value: &RawEncNetworkAddress) -> Result<Self, lcs::Error> {
+        let enc_addr: EncNetworkAddress = lcs::from_bytes(value.as_ref())?;
+        Ok(enc_addr)
+    }
+}
+
+#[cfg(any(test, feature = "fuzzing"))]
```
mimoo, line 324: maybe move that down under the test comment

network/network-address/src/encrypted.rs
```diff
@@ -0,0 +1,482 @@
...
+    // Ensure that modifying the ciphertext or associated data causes a decryption
+    // error.
+    #[test]
+    fn expect_decryption_failures() {
+        let root_key = TEST_ROOT_KEY;
+        let root_key_version = TEST_ROOT_KEY_VERSION;
+        let account = AccountAddress::ZERO;
+        let seq_num = 0x4589;
+        let addr_idx = 123;
+        let addr = NetworkAddress::from_str("/ip4/127.0.0.1/tcp/1234").unwrap();
```
gregnazario, line 380: nit: `NetworkAddress::mock()`

network/network-address/src/encrypted.rs
```diff
@@ -0,0 +1,482 @@
...
+        let seq_num = self.seq_num;
+        let mut enc_addr = self.enc_addr;
+
+        // ciphertext is too small to even contain the authentication tag, so it
+        // must be invalid.
+        if enc_addr.len() < AES_GCM_TAG_LEN {
+            return Err(Error::CiphertextTooSmall(enc_addr.len()));
+        }
+
+        let account_key = Self::derive_account_key(root_key, account)?;
```
mimoo, line 266: IIUC this will never fail, but I can agree that it's better to just bubble up the error if we're already returning a Result anyway

network/network-address/src/encrypted.rs
```diff
@@ -0,0 +1,482 @@
...
+        let aead = Aes256Gcm::new(GenericArray::from_slice(&account_key));
+
+        // nonce := seq_num || addr_idx; sizeof(nonce) == 12
+        // ex: seq_num = 0x1234, addr_idx = 0x04
+        //     ==> nonce_slice == &[0, 0, 0, 0, 0, 0, 0x12, 0x34, 0, 0, 0, 0x4]
+        let nonce = ((seq_num as u128) << 32) | (addr_idx as u128);
+        let nonce_buf = (nonce as u128).to_be_bytes();
+        let nonce_slice = &nonce_buf[mem::size_of::<u128>() - AES_GCM_NONCE_LEN..];
```
mimoo, line 274: same comment as above

network/network-address/src/encrypted.rs
```diff
@@ -0,0 +1,482 @@
...
+        account: &AccountAddress,
+        addr_idx: u32,
+    ) -> Result<RawNetworkAddress, Error> {
+        let root_key_version = self.root_key_version;
+        let seq_num = self.seq_num;
+        let mut enc_addr = self.enc_addr;
+
+        // ciphertext is too small to even contain the authentication tag, so it
+        // must be invalid.
+        if enc_addr.len() < AES_GCM_TAG_LEN {
```
mimoo, line 262: IIUC we actually know the exact size of the payload no? If so we could check that it's the correct size instead

network/simple-onchain-discovery/src/lib.rs
```diff
@@ -45,95 +50,136 @@ pub static EVENT_PROCESSING_LOOP_BUSY_DURATION_S: Lazy<DurationHistogram> = Lazy
...
 
     // Collect the set of address updates.
-    let address_map = node_list
+    let new_peer_addrs: HashMap<_, _> = node_set
+        .into_iter()
+        .map(|info| {
+            let peer_id = *info.account_address();
+            let config = info.into_config();
+            // only one field currently, though this will change soon
+            let addr_idx = 0;
```
gregnazario, line 114: Is this `addr_idx` per peer?

network/simple-onchain-discovery/src/lib.rs
```diff
@@ -45,95 +50,136 @@ pub static EVENT_PROCESSING_LOOP_BUSY_DURATION_S: Lazy<DurationHistogram> = Lazy
...
+                    decrypt_validator_netaddr(&root_key_map, &peer_id, addr_idx, raw_enc_addr)
+                }
+                RoleType::FullNode => Ok(config.full_node_network_address),
+            };
+
+            let addr_res = raw_addr_res.and_then(|raw_addr| {
+                let addr = NetworkAddress::try_from(&raw_addr)
+                    .map_err(anyhow::Error::new)
+                    .with_context(|| {
+                        format_err!("error deserializing network address: {:?}", raw_addr)
```
gregnazario, line 128: It seems like a lot of nesting in one map function.  Do you think it's possible / worth it to break it up?

network/network-address/src/encrypted.rs
gregnazario: `derived_key` is probably good for this case.  It's descriptive and short.  The fact that it's derived from the AccountAddress is just an implementation detail.

config/management/src/validator_config.rs
phlip9: > Will we always know ahead of time if it's encrypted or not?

yes, the `validator_network_address` field will only contain serialized encrypted network addresses (hoping to have time to rename these on-chain things a bit : p)

> I guess my nit here is EncNetworkAddress isn't really an Encrypted NetworkAddress, but instead an encrypted RawNetworkAddress.

I thought about this as well, but "RawEncRawNetworkAddress" looks pretty wild. it's definitely a more consistent naming scheme though

gregnazario: Let's just revisit the naming for those after this then

network/network-address/src/encrypted.rs
phlip9: i found this weird as well. aes-gcm will fail if the plaintext length is > 2^36 bytes, but that should probably never happen

https://github.com/RustCrypto/AEADs/blob/master/aes-gcm/src/lib.rs#L234

```rust
    fn encrypt_in_place_detached(/* .. */) -> Result {
        if buffer.len() as u64 > P_MAX || associated_data.len() as u64 > A_MAX {
            return Err(Error);
        }
        // ..
    }
```

should we just unwrap here and call it a day? i also don't believe the HKDF should ever fail (since we always use valid output length. root key and account are always valid sizes too, since they're always 32 bytes), so we could unwrap that too.

network/network-address/src/encrypted.rs
phlip9: mentioned this above, but this should never fail. let's unwrap and call it a day?

network/network-address/src/encrypted.rs
phlip9: but muh efficiency : d

network/network-address/src/encrypted.rs
phlip9: AES-256-GCM expects a 92-bit nonce. of course, rust doesn't have a u92, so i'm "projecting" the nonce into 128-bit space (16-byte buffer) and then truncating it into a 12-byte buffer.

network/simple-onchain-discovery/src/lib.rs
phlip9: yeah, this is in preparation for

```rust
struct ValidatorConfig {
    // ..
    validator_network_addresses: Vec<Vec<u8>>,
}
```

i.e., there are multiple advertised addresses per validator (and full node)

so the addr_idx is the index of the address in this list of addresses

JoshLind:
I haven't dug too deeply into the PR, but a couple high-level comments (especially after our offline sync):

- We need to settle on the right names for these keys (especially as there's already a lot of keys being handled by each validator). Note that once we've settled, we should probably include the keys name in the global constants crate (e.g., https://github.com/libra/libra/blob/67f1e5ef17b79889ebd5793a8ca7e9345eda9841/config/global-constants/src/lib.rs#L17). My feeling is root_key, account_key are too overloaded. Perhaps something like "SHARED_NET_KEY" and "DERIVED_NET_KEY"?

- We should also sync with @davidiw soon to discuss how we envision getting the shared network keys into each validator secure storage instance, and what modifications we need to make to the genesis flow to get this right.

- One thing I would avoid right now is over-complicating the design. I think the decision to assume a single "SHARED_NET_KEY" that cannot be rotated yet is best for now. Especially as this is a defence-in-depth approach.

network/network-address/src/encrypted.rs
mimoo: I don't think it has to be (unless the specific HKDF library forbids it). The salt is just here to differentiate your instance of HKDF.

network/network-address/src/encrypted.rs
mimoo: (or even associated ;))

network/network-address/src/encrypted.rs
mimoo: my 2 cents: imo it's not worth adding another layer of abstraction for something straight forward like that

state-synchronizer/src/tests/integration_tests.rs
```diff
@@ -168,13 +174,25 @@ impl SynchronizerEnv {
         // The voting power of peer 0 is enough to generate an LI that passes validation.
         for (idx, signer) in signers.iter().enumerate() {
             let voting_power = if idx == 0 { 1000 } else { 1 };
-            let addr: NetworkAddress = "/memory/0".parse().unwrap();
+            let addr = NetworkAddress::from_str("/memory/0").unwrap();
+            let raw_addr = RawNetworkAddress::try_from(&addr).unwrap();
+            let shared_val_netaddr_key = TEST_SHARED_VAL_NETADDR_KEY;
+            let key_version = TEST_SHARED_VAL_NETADDR_KEY_VERSION;
+            let enc_addr = raw_addr.clone().encrypt(
+                &shared_val_netaddr_key,
```
gregnazario, line 182: Nit: I don't think there's a lot of value declaring a constant a variable, and then immediately using the variable.

PR #4960, opened 2020-07-08
[libra framework] Update account limits to use centralized account li…

### LINKED ISSUE #4940: Stdlib cleanup: make LimitsDefinition for VASPs global rather than VASP-specific
This will de-duplicate LimitsDefinitions, make them work the same way for VASPs and UHW, and make VASP accounts smaller.

tzakian:
Previously every parent VASP account had to have a limits definition for every currency that it can hold. Now it only needs to hold a window for that currency in the parent account. 

* Only unrestricted limits definitions can be published now. The limits can then be restricted via an `update_account_limit_definition` transaction. 
* Anyone can publish an unrestricted limits definition for `C`, but no one will point to it. One can only be published as long as a limits definition for `C` is not already published under that address.
* Only the association, or a caller with access to a `&CallingCapability` can point a window at a specific limit address,
* Only the association can change the fields inside of a limits definition.

In this setting, if a VASP needs to be limited, the association can either point the VASP's windows at pre-existing limits definitions that they wish to use, or they will need to publish limits definitions and point the windows at those, along with setting the limits in the newly-published limits definitions.

Closes #4940


language/stdlib/modules/AccountLimits.move
```diff
@@ -165,26 +156,38 @@ module AccountLimits {
...
-    /// the entity across all of its accounts will need to be set by the association.
-    public fun set_current_holdings<CoinType>(
+    /// Update either the `tracked_balance` or `limit_address` fields of the
+    /// `Window<CoinType>` stored under `window_address`.
+    /// * Since we don't track balances of accounts before they are limited, once
+    ///   they do become limited the approximate balance in `CointType` held by
+    ///   the entity across all of its accounts will need to be set by the association.
+    ///   if `aggregate_balance` is set to zero the field is not updated.
+    /// * If we wish to point this window to a new limits definition, we can do
+    ///   so by passing a `new_limit_address` to this function. If the
```
moeziniaf, line 178: Are we guaranteeing that new_limit_address has a LimitsDefinition resource via genesis?
Why is VM_RESERVED_ADDRESS specifically used as a means of bypassing setting the new_limit_address?

language/stdlib/modules/LibraAccount.move
```diff
@@ -150,12 +150,10 @@ module LibraAccount {
         assert(LibraTimestamp::is_genesis(), ENOT_GENESIS);
         // Operational constraint, not a privilege constraint.
         assert(Signer::address_of(lr_account) == CoreAddresses::LIBRA_ROOT_ADDRESS(), EINVALID_SINGLETON_ADDRESS);
-        let limits_cap = AccountLimits::grant_calling_capability(lr_account);
-        AccountLimits::initialize(lr_account, &limits_cap);
         move_to(
             lr_account,
             AccountOperationsCapability {
-                limits_cap,
+                limits_cap: AccountLimits::grant_calling_capability(lr_account),
```
moeziniaf, line 156: maybe this happened for a different reason, but why is it called 'calling capability'? is it used for other caps?

language/stdlib/modules/AccountLimits.move
tzakian: Good questions. 
* You are not guaranteed that the limits definition resource is published in genesis. You can see an example of publishing a limits definition after genesis in the updates to the `account_limits` test. The association will need to make sure that the limits definition resource lives at the address before pointing the window to it. Windows are by default pointed to global limits definitions when they are created, so they will remain consistent. I'll add a check to make sure a limits definition live at the `new_limit_address` address.
* We use VM_RESERVED_ADDRESS since no one can send transactions from that address except for the VM, in particular, no account can be created there. Because of this, short of the VM creating a limits definition at its address (highly highly unlikely) this address will never hold a limits definition resource so it's unlikely you'll want to set a new_limit_address to point at the VM's reserved address.

language/stdlib/modules/LibraAccount.move
tzakian: Just used for account limits. It's probably a good idea to update the name; I'll make that change in this PR.

language/stdlib/modules/AccountLimits.move
```diff
@@ -5,13 +5,13 @@ module AccountLimits {
...
     use 0x1::LBR::LBR;
-    use 0x1::Roles::{has_libra_root_role, has_treasury_compliance_role};
+    use 0x1::Roles;
     use 0x1::CoreAddresses;
     use 0x1::Signer;
 
     /// An operations capability that restricts callers of this module since
     /// the operations can mutate account states.
-    resource struct CallingCapability { }
+    resource struct AccountLimitMutationCapability { }
```
sblackshear, line 14: Nice

language/stdlib/transaction_scripts/publish_account_limit_definition.move
```diff
@@ -0,0 +1,9 @@
+script {
+use 0x1::AccountLimits;
+
+/// Publishes an unrestricted `LimitsDefintion<CoinType>` under `account`.
+/// Will if a resource with the same type already exists under `account`. No windows will point to this limit at the
```
sblackshear, line 5: Finish comment?

language/stdlib/modules/AccountLimits.move
```diff
@@ -165,26 +157,41 @@ module AccountLimits {
...
+        if (new_time_period > 0) { limits_def.time_period = new_time_period };
     }
 
-    /// Since we don't track balances of accounts before they are limited, once
-    /// they do become limited the approximate balance in `CointType` held by
-    /// the entity across all of its accounts will need to be set by the association.
-    public fun set_current_holdings<CoinType>(
+    /// Update either the `tracked_balance` or `limit_address` fields of the
+    /// `Window<CoinType>` stored under `window_address`.
+    /// * Since we don't track balances of accounts before they are limited, once
```
sblackshear, line 174: This is a really nasty problem, I didn't fully understand it in the diff that encountered the performance problem while trying to do this the right way. No suggestions for improvement at the moment, but thinking...

language/stdlib/modules/AccountLimits.move
```diff
@@ -165,26 +157,41 @@ module AccountLimits {
...
         window_address: address,
         aggregate_balance: u64,
+        new_limit_address: address,
     ) acquires Window {
-        assert(has_treasury_compliance_role(tc_account), ENOT_TREASURY_COMPLIANCE);
-        borrow_global_mut<Window<CoinType>>(window_address).tracked_balance = aggregate_balance;
+        assert(Roles::has_treasury_compliance_role(tc_account), ENOT_TREASURY_COMPLIANCE);
+        let window = borrow_global_mut<Window<CoinType>>(window_address);
+        if (aggregate_balance != 0)  { window.tracked_balance = aggregate_balance };
+        if (new_limit_address != CoreAddresses::VM_RESERVED_ADDRESS()) {
```
sblackshear, line 191: I would vote to omit this this check given that (a) it seems to suggest the possibility that 0x0 can hold a resource (hence Frederic's question!), but it's an important Libra framework invariant that it cannot (hopefully the prover can eventually help us with this) and (b) the next line will already fail with a reasonable error message if this doesn't exist.

language/stdlib/modules/AccountLimits.move
tzakian: It really is quite nasty. The only way I can think is to have the windows be held in child VASP accounts, but then that doesn't solve the issue of collecting the balances for all the child accounts...I think it's definitely something worth brainstorming about at some point.

PR #4961, opened 2020-07-08
[logging] adding security logging to state-sync

mimoo:


common/logger/src/security.rs
```diff
@@ -62,11 +62,11 @@ pub mod security_events {
     /// A block being committed or executed is invalid
     pub const INVALID_BLOCK: &str = "InvalidBlock";
 
-    // Executor
-    // --------
+    // State-Sync
+    // ----------
 
-    /// Executor received an invalid transactions chunk
-    pub const INVALID_CHUNK_EXECUTOR: &str = "InvalidChunkExecutor";
```
mimoo, line 69: I don't think it makes sense to log anything security in executor, as we don't know who sent us the malicious payload there. Better to log things on the client side (state-sync or consensus)

gregnazario: I'm wondering why is an invalid chunk a security log?

mimoo: I should probably document this better, security logs are for detection of malicious behaviors by other validators. This pretty much tells us that a validator full node has sent us a chunk of transaction that is incorrect (this should not happen in practice unless someone is trying to mess with us).

PR #4967, opened 2020-07-08
[logging] security logging for invalid consensus messages

mimoo:


consensus/src/epoch_manager.rs
```diff
@@ -374,10 +374,23 @@ impl EpochManager {
         peer_id: AccountAddress,
         consensus_msg: ConsensusMsg,
     ) -> anyhow::Result<()> {
-        if let Some(event) = self.process_epoch(peer_id, consensus_msg).await? {
+        // we can't verify signatures from a different epoch
+        let unverified_event = self.process_epoch(peer_id, consensus_msg).await?;
```
zekun000, line 378: why you create this var? you probably want
```
if let Some(unverified_event) = self.process_epoch
```

mimoo: That was to make the step (and the comment) stand out.
I find it easier to read while matching the specification like this, wdyt?

PR #4969, opened 2020-07-08
[types] add ChainID to user txn

christinacelee:
###  Motivation

Adding ChainId to a user txn, to prevent txns potentially being replayed across networks (chains). The ChainID is represented as a u8

In addition to the regular ChainId, this PR also adds the `ReservedChain` type to specify chain IDs that are reserved. This enum type is primarily used to improve human readability of chain ID in config files and CLI (is used as an adaptor from string to the actual u8 representation of the chain ID). 

### Test plan
Checked that CLI and config can consume both human-readable and numeric input of chain ID


christinacelee:
some of the changes are uncharted territory for me personally, especially KeyManager/secure/config things, so please take a look and make sure I'm not doing something obviously clowny! @JoshLind

network/src/transport.rs
```diff
@@ -589,7 +589,7 @@ mod test {
             listener_key,
             trusted_peers.clone(),
             HANDSHAKE_VERSION,
-            chain_id.clone(),
+            chain_id,
```
gregnazario, line 592: A nice side effect of it now being a number.

config/src/config/key_manager_config.rs
```diff
@@ -32,6 +34,7 @@ impl Default for KeyManagerConfig {
             secure_backend: SecureBackend::InMemoryStorage,
             sleep_period_secs: DEFAULT_SLEEP_PERIOD_SECS,
             txn_expiration_secs: DEFAULT_TXN_EXPIRATION_SECS,
+            chain_id: ChainId::mainnet(),
```
gregnazario, line 37: I'd be concerned with a default that connects to the mainnet chain.  I'll let @JoshLind make the call on that one though.

types/src/chain_id.rs
```diff
@@ -0,0 +1,52 @@
...
+
+impl Default for ChainId {
+    fn default() -> Self {
+        ChainId::new(1)
+    }
+}
+
+impl Display for ChainId {
+    fn fmt(&self, f: &mut Formatter) -> std::fmt::Result {
+        write!(f, "ChainId {}", self.0)
```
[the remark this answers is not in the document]
christinacelee: I thought `ChainId` might make things easier to debug, others can weigh in too

types/src/chain_id.rs
```diff
@@ -0,0 +1,52 @@
...
+// SPDX-License-Identifier: Apache-2.0
+use anyhow::{Error, Result};
+use serde::{Deserialize, Serialize};
+use std::{
+    fmt::{Display, Formatter},
+    str::FromStr,
+};
+
+#[derive(Clone, Copy, Debug, Deserialize, Eq, Hash, PartialEq, Serialize)]
+pub struct ChainId(pub u64);
```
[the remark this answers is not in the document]
zekun000: enum is better than a u64 imo

sblackshear:
@zekun000 I think we should definitely read it from genesis, otherwise validators with different configs will disagree...

zekun000:
sounds like a good on chain config value :)

types/src/chain_id.rs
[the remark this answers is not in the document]
mimoo: agree, I think we should use an enum here

types/src/chain_id.rs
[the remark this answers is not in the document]
mimoo: if it's an enum we'll get the chainId name directly :)

types/src/chain_id.rs
```diff
@@ -0,0 +1,52 @@
...
+        ChainId(chain_id)
+    }
+
+    pub fn mainnet() -> Self {
+        ChainId::new(MAINNET_CHAIN_ID)
+    }
+
+    #[cfg(any(test, feature = "fuzzing"))]
+    pub fn mock() -> Self {
+        ChainId::new(MOCK_CHAIN_ID)
```
mimoo, line 50: I feel like we tend to use "mock" for components, and "testing" for values

types/src/chain_id.rs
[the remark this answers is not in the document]
mimoo: actually I'm thinking that we might want to go further, it'd be good if we can just have "mainnet" or "testnet" as chainid in CLIs or config files : o wdyt?

mimoo:
Good job on the PR! This looks good to me (besides the ChainId `u64` -> `Enum`), as this is touching a lot of stuff I feel like someone else should approve this :D

> not for this pr, but it's worth discussing whether we want to read the chain id from genesis upon start to eliminate two source of truth (config + genesis)

I would prefer that we have it in both the config AND the genesis, this way if my config says "testnet" but I start with the "mainnet" genesis then it won't work. A good way to avoid configuration mistakes and be explicit about what we're doing.

sblackshear:
> Good job on the PR! This looks good to me (besides the ChainId u64 -> Enum), as this is touching a lot of stuff I feel like someone else than me should approve this :D

I also prefer enums to plain numbers in general, but Move does not have enums... So I would vote for sticking with the u64 to avoid Rust -> Move conversion surprises.

> I would prefer that we have it in both the config AND the genesis, this way if my config says "testnet" but I start with the "mainnet" genesis then it won't work. A good way to avoid configuration mistakes and be explicit about what we're doing.

Makes sense, but at what point in the code should we compare genesis to the config and fail?

davidiw:
Thanks for pushing this PR out! It has been a loooong time coming :D!!!! 

How did we land on using a u64 to represent the ChainId versus the string earlier? My guess is so that it could be incorporated into the transaction at low cost?

Couple of discussion thoughts / topics:
* How do we distinguish against the internal / external / third-party chains?
* How do we ensure that admins properly update this value in their config?

One idea we had floating around was to make the ChainId deterministic computation of shared data plus some fixed shared counter. E.g., "ledger version:ledger hash:counter", but that would be a lot of fields and in a string, so not that great... Have we considered a hash?

Also one concern about this being within an "on-chain config" is -- would that make it mutable? Is that a property we want to have? Is it possible to have immutable values in on-chain configs?

sblackshear:
> Also one concern about this being within an "on-chain config" is -- would that make it mutable? Is that a property we want to have? Is it possible to have immutable values in on-chain configs?

Agreed that we probably don't want this to be an on-chain config because it's immutable. It should just be an ordinary constant written at genesis.

zekun000:
I'm a little confused about immutability, if the mainnet forks do we want to change the chain id e.g. a fork enabling more risk smart contracts? from what I heard from @aching we do, or we simply change it by writeset.

aching:
I think immutable is right.  Supporting easy forks doesn't seem to be a use case.  Writeset seems reasonable to me.

aching:
> How did we land on using a u64 to represent the ChainId versus the string earlier? My guess is so that it could be incorporated into the transaction at low cost?

u64, u32, enum, etc. become a single byte if mainnet is 0.

> 
> Couple of discussion thoughts / topics:
> 
> * How do we distinguish against the internal / external / third-party chains?

I think the only idea is LA will define a set of chains (i.e. mainnet = 0).  Any other chain can pick any id. Perhaps if we want LA could also define ranges for private chains, testnet chains, etc.

> * How do we ensure that admins properly update this value in their config?
> 
> One idea we had floating around was to make the ChainId deterministic computation of shared data plus some fixed shared counter. E.g., "ledger version:ledger hash:counter", but that would be a lot of fields and in a string, so not that great... Have we considered a hash?
> 

This is about a tradeoff of debuggability of safety vs size.  Hash is bigger than a byte. =)

> Also one concern about this being within an "on-chain config" is -- would that make it mutable? Is that a property we want to have? Is it possible to have immutable values in on-chain configs?

Agree on immutable.

types/src/chain_id.rs
```diff
@@ -1,18 +1,52 @@
...
+};
 
-#[derive(Clone, Debug, Deserialize, Eq, Hash, PartialEq, Serialize)]
-pub struct ChainId(pub String);
+#[derive(Clone, Copy, Debug, Deserialize, Eq, Hash, PartialEq, Serialize)]
+pub struct ChainId(pub u64);
+
+const MAINNET_CHAIN_ID: u64 = 0;
+#[cfg(any(test, feature = "fuzzing"))]
+pub const MOCK_CHAIN_ID: u64 = 1;
```
xli, line 15: should we just have TESTNET_CHAIN_ID here and for all tests? mock gives impression it is generated number. 
also, since 0 is mainnet, 1 wil be testnet?

config/src/config/key_manager_config.rs
JoshLind: I don't have strong feelings, but I think this should probably match whatever default we're using in the actual validator/node configs.

config/management/src/validator_config.rs
```diff
@@ -33,6 +34,8 @@ pub struct ValidatorConfig {
     fullnode_address: NetworkAddress,
     #[structopt(flatten)]
     backends: SecureBackends,
+    #[structopt(long)]
+    chain_id: ChainId,
```
JoshLind, line 38: I don't think this is quite right, as it's essentially setting a chainID on a single transaction used during the genesis process (this transaction is submitted by each validator operator). Instead, my feeling is that we'll probably want to have the association determine the ID and set it in genesis once, before bootstrapping the network. However, from reading the comments it seems like there's still a lot of on-going discussions about how this should actually work..

types/src/chain_id.rs
```diff
@@ -0,0 +1,52 @@
...
+
+impl Default for ChainId {
+    fn default() -> Self {
+        ChainId::new(1)
+    }
+}
```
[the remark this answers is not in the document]
JoshLind: +1 (I'm a big fan of just using constants where applicable to remove ambiguity about what ID is actually being picked)

config/management/src/validator_config.rs
JoshLind: (If this is something that the association is going to decide at genesis, we may want to setup a sync with @davidiw, to figure out the best place to do this for the management tool)

thefallentree:
please also add changes to json-rpc/types/view.rs to show chainID in JSONRPC

thefallentree:
as well as updating "json-rpc-spec.md"

thefallentree:
LGTM on jsonrpc side.

types/src/chain_id.rs
```diff
@@ -0,0 +1,120 @@
...
+/// (e.g. MAINNET has numeric chain ID 0, PREMAINNET has chain ID 1, etc)
+#[repr(u8)]
+#[derive(Copy, Clone, Debug)]
+pub enum ReservedChain {
+    MAINNET,
+    PREMAINNET,
+    TESTNET,
+    DEVNET,
+    TESTING,
+}
```
aching, line 23: When it comes to reserved Libra networks, I agree MAINNET = 0.

However, PREMAINNET, TESTNET, DEVNET, TESTING, are non-reserved in my opinion.  They could evolve over time.  One thing we could do if we want to keep a single enum is something like below - renaming to NamedChain and then clearly laying out what is reserved and what is not:

```suggestion
pub enum NamedChain {
    /// MAINNET is the Libra mainnet production chain and is reserved for 0.
    MAINNET,
    // The below chains are non-reserved, non-production and may change over time.  They are listed for convenience here.
    PREMAINNET,
    TESTNET,
    DEVNET,
    TESTING,
}
```

types/src/transaction/mod.rs
```diff
@@ -80,6 +81,9 @@ pub struct RawTransaction {
     #[serde(serialize_with = "serialize_duration")]
     #[serde(deserialize_with = "deserialize_duration")]
     expiration_time: Duration,
+
+    // chain ID of the network this transaction is intended for
```
aching, line 85:
```suggestion
    // chain ID of the Libra network this transaction is intended for
```

config/src/config/key_manager_config.rs
```diff
@@ -21,6 +22,8 @@ pub struct KeyManagerConfig {
     pub secure_backend: SecureBackend,
     pub sleep_period_secs: u64,
     pub txn_expiration_secs: u64,
+    #[serde(deserialize_with = "chain_id::deserialize_config_chain_id")]
+    pub chain_id: ChainId,
```
christinacelee, line 26: chain ID needs to be added to key manager config too @mgorven @sausagee

christinacelee: @runtian-zhou

PR #4972, opened 2020-07-09
[libra framework] Make Designated Dealer <-> VASP transactions subject to dual attestation

sblackshear:
This is a series of commits to make DD -> VASP and VASP -> DD transfers subject to dual attestation checks.

First commit: extract the data required for the off-chain dual attestation protocol from `ParentVASP` and put it in its own module/resource (`DualAttestation::Credential`). (cc @emmazzz)
Second commit: move all logic related to dual attestation out of `LibraAccount` and into `DualAttesation`. `LibraAccount` now just calls: `DualAttestation::assert_payment_ok` (cc @wrwg)
Third commit: change the dual attestation logic to check for a DD sender or recipient (cc @dmitri-perelman, @moeziniaf, @shazqadeer)

I have to add some tests and fix some prover errors on the third commit, but this should be mostly ready to go.


language/stdlib/modules/DualAttestation.move
```diff
@@ -0,0 +1,308 @@
...
+    use 0x1::Libra;
+    use 0x1::LibraTimestamp;
+    use 0x1::Roles;
+    use 0x1::Signature;
+    use 0x1::Signer;
+    use 0x1::VASP;
+    use 0x1::Vector;
+
+    /// This resource holds an entity's globally unique name and all of the metadata it needs to
+    /// participate if off-chain protocols.
```
tzakian, line 16:
```suggestion
    /// participate in off-chain protocols.
```

language/stdlib/modules/DualAttestation.move
```diff
@@ -0,0 +1,308 @@
...
+        human_name: vector<u8>,
+        /// The base_url holds the URL to be used for off-chain communication. This contains the
+        /// entire URL (e.g. https://...). Mutable.
+        base_url: vector<u8>,
+        /// 32 byte Ed25519 public key whose counterpart must be used to sign
+        /// (1) the payment metadata for on-chain transactions that require dual attestation (e.g.,
+        ///     transactions subject to the travel rule)
+        /// (2) information exchanged in the off-chain protocols (e.g., KYC info in the travel rule
+        ///     protocol)
+        /// Note that this is different than `authentication_key` used in LibraAccount::T, which is
```
tzakian, line 28:
```suggestion
        /// Note that this is different than `authentication_key` used in `LibraAccount::LibraAccount`, which is
```

language/stdlib/modules/DualAttestation.move
```diff
@@ -0,0 +1,308 @@
...
+        // Expire the parent credential.
+        credential.expiration_date = 0;
+    }
+    spec fun decertify {
+        aborts_if false;
+        ensures credential.expiration_date == 0;
+    }
+
+    /// A year in microseconds
+    fun cert_lifetime(): u64 {
```
tzakian, line 159: Pull into constant now that we have those?

shazqadeer:
I see two temporary files output.bpl and output.bpl.log being checked in as part of this PR.  Is that an oversight?

sblackshear:
Good call; git mistake. I will delete.

language/stdlib/modules/DualAttestation.move
```diff
@@ -0,0 +1,308 @@
...
+    /// (1) dual attestation is not required for this payment, or
+    /// (2) dual attestation is required, and `metadata_signature` can be verified on the message
+    ///     `metadata` | `payer` | `amount` | `DOMAIN_SEPARATOR` using the `compliance_public_key`
+    ///     published in `payee`'s `Credential` resource
+    /// It aborts with an appropriate error code if dual attestation is required, but one or more of
+    /// the conditions in (2) is not met.
+    public fun assert_payment_ok<Currency>(
+        payer: address,
+        payee: address,
+        value_microlibra: u64,
```
tzakian, line 298: Maybe `value` instead of `value_microlibra`? If I understand correctly this value is in `Currency`.

language/tools/vm-genesis/src/lib.rs
```diff
@@ -194,6 +194,9 @@ fn create_and_initialize_testnet_minting(
         0,
         account_config::testnet_dd_account_address(),
         genesis_auth_key.prefix().to_vec(),
+        b"moneybags".to_vec(), // name
```
tzakian, line 197: 😆

sblackshear:
Note: ran the prover tests locally and they all pass. Gonna go for it!

@bors-libra r=tzakian

moeziniaf:
This is an awesome PR. Love the refactoring to bring out all the Credentials and, and the `DualAttestation` module is super pristine 💯 . I didn't realize we had the testnet DD minter being created already, that's awesome. 
Is there any way to test a successful dual attested p2p payment works with the right keys and sigs so  `assert_signature_is_valid` is true?

sblackshear:
> Is there any way to test a successful dual attested p2p payment works with the right keys and sigs so assert_signature_is_valid is true?

This can be done today on testnet between VASPs, but this PR won't hit testnet for a bit. If you spin up a local network and create test DD/VASP accounts, some variant of [these](https://github.com/libra/libra/tree/master/client/libra-dev#dual-attestationtravel-rule-protocol) instructions should work.

moeziniaf:
What is obstructing us from having language/e2e_tests/ for this?

sblackshear:
We have these tests; check out the [`dd_dual_attestation_payments`](https://github.com/libra/libra/blob/master/language/e2e-tests/src/tests/transaction_builder.rs#L666) test added by this PR.

moeziniaf:
But those are asserting `BAD_METADATA_SIGNATURE_ERROR_CODE`. Do we have any tests which execute p2p txns with correct signatures and dual attestation data?

sblackshear:
The `dual_attesation_payment` test does this with VASP accounts. I did not duplicate it for DD's because the dual attestation logic run is identical to VASPs; only the role is different. But it wouldn't be a bad idea to add this. I will do so.

PR #4964, opened 2020-07-08
[libra_vm] Refactor libra_vm implementation

runtian-zhou:
## Motivation

Refactored libra_vm into three major components.
1. `LibraVMImpl`, which defines how move functions are being invoked by the execution/validation process.
2. `LibraVM`, which implements the `VMExecutor` api. 
3. `LibraVMValidator`, which implements thr `VMValidator` api.

I got rid of the AccessPathCache for now in the LibraVM due to the borrow_checker issue and we will need to look into that a bit further. The major change in the execution flow is that I got rid of the `VerifiedPayload` and created separate flow for executing the script and executing the module, and a common entry for a failed transaction cleanup.

I intentionally leave `LibraVM` name unchanged just to avoid major renaming in a number of code locations. This can be done in a separate PR.

There were a number of direct move function calls in `LibraVM` such as calling into block_prologue and execute_script body. We might want to get rid of those as well.

### Have you read the [Contributing Guidelines on pull requests](https://github.com/libra/libra/blob/master/CONTRIBUTING.md#pull-requests)?

Yes

## Test Plan

`cargo test`

PR #4974, opened 2020-07-09
[language] combine VM counters for per-transaction gas usage

bob-wilson:
This combines the TXN_EXECUTION_GAS_USAGE and TXN_TOTAL_GAS_USAGE histogram counters into a single TXN_GAS_USAGE counter. This fixes a problem where system transactions (e.g., block prologues) were included in the "total" count, but since they are not charged gas, they pull down the per-transaction average, especially on testnet where there are currently far more system transactions than user transactions. Moreover, for user transactions, we do not charge gas for the epilogues so there was no difference between the two counters we had before. The new counter is only used for user transactions.

Besides fixing up some comment typos, this also cleans up the handling of gas for block prologue transactions. Some of the comments there were no longer relevant, and in addition to using the zero_cost_schedule, that code should use the "system" CostStrategy which does not charge gas.

Yes

## Test Plan

Ran the existing tests

### DIFF
--- a/language/libra-vm/src/counters.rs
+++ b/language/libra-vm/src/counters.rs
@@ -85,18 +85,10 @@ pub static TXN_EPILOGUE_SECONDS: Lazy<Histogram> = Lazy::new(|| {
     .unwrap()
 });
 
-pub static TXN_EXECUTION_GAS_USAGE: Lazy<Histogram> = Lazy::new(|| {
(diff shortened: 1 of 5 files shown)


bob-wilson:
OK, thanks. I'm planning to wait to land this after https://github.com/libra/libra/pull/4964

bob-wilson:
@dariorussi Can you take another look at this? Things shifted around since the code changed underneath me

PR #4978, opened 2020-07-09
[libra framework] Make DD tiers multi-currency compatible

### LINKED ISSUE #4973: Sdtlib feature: make DD tiers per-currency
Currently there is a single tier for all currencies, which won't fly given that DD accounts are now multi-currency.

tzakian:
Parametrizes DD tiers by currency.

Need to add some tests for this still but want to make sure these changes look good from the AOS side first. 

Closes #4973


language/stdlib/transaction_scripts/create_designated_dealer.move
```diff
@@ -19,9 +21,20 @@ fun create_designated_dealer<Currency>(
         tc_account, addr, auth_key_prefix, add_all_currencies
     );
     // Create default tiers for newly created DD
-    DesignatedDealer::add_tier(tc_account, addr, 500000);
-    DesignatedDealer::add_tier(tc_account, addr, 5000000);
-    DesignatedDealer::add_tier(tc_account, addr, 50000000);
-    DesignatedDealer::add_tier(tc_account, addr, 500000000);
+    if (add_all_currencies) {
```
sblackshear, line 24: Could/should we move the tier addition inside `DesignatedDealer`? I  think that would make it easier to establish module invariants about the existence of tiers. This would also let us abstract `add_tier` into a function and save some code duplication.

language/stdlib/modules/DesignatedDealer.move
```diff
@@ -34,22 +41,42 @@ module DesignatedDealer {
...
+        tc_account: &signer,
+        add_all_currencies: bool,
+    ) {
+        assert(Roles::has_treasury_compliance_role(tc_account), EACCOUNT_NOT_TREASURY_COMPLIANCE);
+        move_to(dd, Dealer { mint_event_handle: Event::new_event_handle<ReceivedMintEvent>(dd) });
+        if (add_all_currencies) {
+            Libra::publish_preburn_to_account<Coin1>(dd, tc_account);
+            Libra::publish_preburn_to_account<Coin2>(dd, tc_account);
+            publish_designated_dealer_tier<Coin1>(dd, tc_account);
+            publish_designated_dealer_tier<Coin2>(dd, tc_account)
```
moeziniaf, line 62: looks good

language/stdlib/transaction_scripts/create_designated_dealer.move
moeziniaf: I agree, also since we have to add specific bounds when adding tiers, it makes sense to keep these constants (which will probably change) in `DesignatedDealer`

tzakian: Sounds good! I'll move this to the DD module

shazqadeer: @tzakian : I have a PR in progress that makes changes to DesignatedDealer.move.  If you try to touch it too, we will have a hard time landing.  I can make this change myself.  Is it ok for you to land this PR without making changes to Designated Dealer?

language/stdlib/modules/DesignatedDealer.move
```diff
@@ -5,29 +5,41 @@ module DesignatedDealer {
...
+    /// currencies it can hold. All `ReceivedMintEvent` events for all
+    /// currencies will be emitted on `mint_event_handle`.
     resource struct Dealer {
+        /// Handle for mint events
+        mint_event_handle: Event::EventHandle<ReceivedMintEvent>,
+    }
+
+    /// The `TierInfo` resource holds the information needed to track which
+    /// tier a mint to a DD needs to be in.
+    resource struct TierInfo<CoinType> {
```
sblackshear, line 22: Wonder if it would make sense for us to put the `Preburn` resource here in the future?

language/stdlib/modules/DesignatedDealer.move
```diff
@@ -5,29 +5,41 @@ module DesignatedDealer {
...
     }
     // Preburn published at top level in Libra.move
 
-    // Message for mint events
+    /// Message for mint events
     struct ReceivedMintEvent {
-        // The address that receives the mint
+        /// The currency minted
+        currency_code: vector<u8>,
+        /// The address that receives the mint
```
sblackshear, line 36:
```suggestion
        /// The address that received the mint
```

PR #4979, opened 2020-07-09
[consensus] removing MIRAI assumption that is false

mimoo:
This assumption is false (as was found by fuzzing).


hermanventer:
Is there a test case that we can add?

mimoo:
> Is there a test case that we can add? 

So this function is a getter and so it is just use to access a field of that struct. I don't think a post condition makes sense here. I also don't think we should assume that at any point this would be true, intuitively it seems like it should be true passed a certain point in consensus (the rounds can only be incremented one by one, so they won't reach high values like `u64::MAX`, but I'm not sure what we are trying to prove with this.

hermanventer:
I would like to understand exactly how the fuzzing test failed. I.e. did it construct an invalid block chain? If so, how? That seems like a problem in and of itself.

Just removing a check that establishes that the block chain is in an invalid state does not make the problem go away.

mimoo:
My guess is that it just created a proposal for a block at round `u64::MAX`, which is something that is totally possible.
At some point consensus does make sure that the round is consistent with what it has in store, but this comes later (see the `ensure_round` step in the [consensus spec](https://github.com/novifinancial/libra-specs/blob/master/consensus/spec.md#proposalmsg-1)). 
In any case, to do these checks, it needs to access the round field which will trigger this `debug_assert` (in debug mode).

consensus/consensus-types/src/block_data.rs
```diff
@@ -98,11 +98,6 @@ impl BlockData {
     }
 
     pub fn round(&self) -> Round {
-        // Round numbers:
-        // - are reset to 0 periodically.
-        // - do not exceed std::u64::MAX - 2 per the 3 chain safety rule
-        // (ConsensusState::commit_rule_for_certified_block)
-        assumed_postcondition!(self.round < std::u64::MAX - 1);
```
hermanventer, line 105: The assumption should instead be checked here: https://github.com/libra/libra/blob/1860e532734ec181b50124391cc67ce44bce6a48/consensus/safety-rules/src/safety_rules.rs#L107

PR #4980, opened 2020-07-09
[Management] Split Validator owners and operators in the codebase.

JoshLind:
Disclaimer: This PR isn't actually that scary once you see what's going on 😄. While I would have much preferred to split this PR into several, the strong dependencies between the commits meant that a single PR was cleaner than trying to stage things in a fancy way to prevent breaking the code after each individual PR... As such, here it is.

## Motivation

This PR separates out validator owners from validator operators in the Libra codebase. To achieve this, the PR offers the following commits:

1. First, we add the set-operator() command to the genesis tool. This allows validator owners to specify their operator using the operator name. The operator name is then looked up in shared storage (to retrieve the operator public key) and then the tool generates a signed set_operator transaction to upload to the shared storage. This commit adds a test for this functionality.

2. Next, we update the existing validator-config() command in the genesis tool to allow operators to specify the names of the owners using the owner name (instead of account address). The owner name is then looked up in the shared storage (to retrieve the owner public key) and this is used to generate the set_validator_config transaction uploaded to the shared storage. This commit also adds two tests for this functionality, as well as updates the end_to_end() test.

3. Next, we offer some small refactors to the genesis functions. Nothing changes semantically.

4. Next, we add an owner_keypair to the test configurations. This allows future commits to update tests to use the owner_keypair where appropriate.

5. 🥩 Next, we update the genesis() and verify() commands in the management tool and fix config-builder (responsible for generating configs for tests, e.g., smoke tests). This is the meat of the PR and is responsible for changing genesis to create accounts for owners and operators, set operators, upload validator configs and add each validator to the validator set. At the end of this commit, some smoke tests pass. The ones that fail need to be updated to use the validator owner keypair/address as added in commit 4. ('cc @zekun000 😄)

--- From here, each commit is simply fixing different failing tests and components to use the new owner keys! 😆  ---

6. Here, we update safety rules to use the owner account instead of the operator account (where appropriate). This fixes more smoke tests.

7. Here, we update the executor tests to use the owner account and keypair (where appropriate). This fixes the executor tests.

8. Here, we update the key manager component to use the owner account (where appropriate). We also update the key manage unit tests.

9. Here, we update the various move testing infrastructure to use the appropriate keys for testing. We also updated various individual tests, where behaviour is now different. For example, by setting operators in genesis, validator owners cannot update their own configs. ('cc @valerini, this is what I was blocked on yesterday 😄)

10. Here, we update some state sync tests to use the correct keys.

11. Finally, we update two specific smoke tests: (i) the network key rotation test; and (ii) the key manager key rotation test. Also did some tiny simplifications ('cc @gregnazario 😄).

Everything now passes! 🤭 

Yes.

## Test Plan

On each commit, I verified that all related tests pass locally. I also ensured that at the final commit, all tests pass locally.

## Related PRs

None.

### DIFF
--- a/config/management/src/config_builder.rs
+++ b/config/management/src/config_builder.rs
@@ -10,7 +10,6 @@ use libra_config::config::{
 use libra_crypto::ed25519::Ed25519PrivateKey;
 use libra_secure_storage::{CryptoStorage, KVStorage, Value};
 use libra_temppath::TempPath;
-use libra_types::account_address;
 use std::path::{Path, PathBuf};
 
 const SHARED: &str = "_shared";
@@ -81,9 +80,6 @@ impl<T: AsRef<Path>> ValidatorBuilder<T> {
         let ns_shared = ns.clone() + SHARED;
         self.storage_helper.initialize(ns.clone());
 
-        let operator_key = self.storage_helper.operator_key(&ns, &ns_shared).unwrap();
-
-        let validator_account = account_address::from_public_key(&operator_key);
         let mut config = self.template.clone_for_template();
         config.randomize_ports();
 
@@ -103,9 +99,12 @@ impl<T: AsRef<Path>> ValidatorBuilder<T> {
             self.secure_backend(&ns, "full_node"),
         );
 
+        // TODO(joshlind): FIX ME: owner_name is currently unused!
+        let owner_name = format!("owner_{}", index);
+
         self.storage_helper
             .validator_config(
-                validator_account,
+                &owner_name,
                 validator_network_address,
                 fullnode_network_address,
                 self.template.base.chain_id,
--- a/config/management/src/config_builder.rs
+++ b/config/management/src/config_builder.rs
@@ -12,9 +12,12 @@ use libra_secure_storage::{CryptoStorage, KVStorage, Value};
 use libra_temppath::TempPath;
 use std::path::{Path, PathBuf};
 
-const SHARED: &str = "_shared";
-const ASSOCIATION: &str = "association";
-const ASSOCIATION_SHARED: &str = "association_shared";
+const ASSOCIATION_NS: &str = "association";
+const ASSOCIATION_SHARED_NS: &str = "association_shared";
+const OPERATOR_NS: &str = "_operator";
+const OPERATOR_SHARED_NS: &str = "_operator_shared";
+const OWNER_NS: &str = "_owner";
+const OWNER_SHARED_NS: &str = "_owner_shared";
 
 pub struct ValidatorBuilder<T: AsRef<Path>> {
     storage_helper: StorageHelper,
@@ -50,9 +53,12 @@ impl<T: AsRef<Path>> ValidatorBuilder<T> {
     /// Association uploads the validator layout to shared storage.
     fn create_layout(&self) {
         let mut layout = Layout::default();
-        layout.association = vec![ASSOCIATION_SHARED.into()];
+        layout.association = vec![ASSOCIATION_SHARED_NS.into()];
+        layout.owners = (0..self.num_validators)
+            .map(|i| (i.to_string() + OWNER_SHARED_NS))
+            .collect();
         layout.operators = (0..self.num_validators)
-            .map(|v| (v.to_string() + SHARED))
+            .map(|i| (i.to_string() + OPERATOR_SHARED_NS))
             .collect();
 
         let mut common_storage = self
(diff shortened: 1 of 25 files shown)


config/management/src/config_builder.rs
```diff
@@ -103,9 +99,12 @@ impl<T: AsRef<Path>> ValidatorBuilder<T> {
             self.secure_backend(&ns, "full_node"),
         );
 
+        // TODO(joshlind): FIX ME: owner_name is currently unused!
+        let owner_name = format!("owner_{}", index);
+
```
JoshLind, line 129: Note: this is addressed in a later commit! One thing at a a time.. :)

testsuite/tests/libratest/smoke_test.rs
```diff
@@ -1245,6 +1245,7 @@ fn test_malformed_script() {
 
 #[test]
 #[ignore]
+// TODO(joshlind): FIX ME!
```
JoshLind: Note: this is fixed in a later commit!

config/management/src/validator_operator.rs
```diff
@@ -0,0 +1,119 @@
...
+                        Error::RemoteStorageWriteError(constants::VALIDATOR_OPERATOR, e.to_string())
+                    })?;
+            }
+        };
+
+        Ok(set_operator_tx)
+    }
+
+    /// Creates and returns a signed set-operator transaction to be uploaded to the remote storage.
+    fn create_set_operator_transaction(&self, script: Script) -> Result<Transaction, Error> {
```
davidiw, line 61: due to "reasons", let's avoid making this a transaction and instead just upload the name of the operator. If you want, we can put this under a test flag or comment it out... because I can imagine post genesis, this will be more of a utility app.  or come up with some alternative...

config/management/src/lib.rs
```diff
@@ -404,27 +431,68 @@ pub mod tests {
...
+                    TransactionArgument::Address(account_address) => {
+                        assert_eq!(&owner_account, account_address);
+                    }
+                    _ => panic!(
+                        "Found an invalid argument type for the validator-config transaction script!"
+                    ),
+                };
+            }
+            _ => panic!("Invalid validator-config transaction payload found!"),
+        };
```
davidiw: this feels a little heavy considering the other tests we have... the original purpose of this test was just to ensure that the functions CLI interaction remained consistent.

config/management/src/lib.rs
```diff
@@ -487,6 +555,67 @@ pub mod tests {
         };
     }
 
+    #[test]
```
davidiw, line 558: this test and the above seem somewhat duplicate.

config/management/src/validator_config.rs
```diff
@@ -37,81 +35,145 @@ pub struct ValidatorConfig {
...
+        // Save the owner account in local storage for deployment later on
+        local_storage
+            .set(OWNER_ACCOUNT, Value::String(owner_account.to_string()))
+            .map_err(|e| {
+                Error::LocalStorageWriteError(OWNER_ACCOUNT, e.to_string())
+            })?;
+
+        // Upload the validator config to shared storage
+        match self.backends.remote {
+            None => return Err(Error::RemoteStorageMissing),
```
davidiw, line 69: why are we making this mandatory? it is entirely possible that I want to perform a manual rotation and this is the tool to do it, we just have to improve the API to export the transaction for submission.

config/management/src/validator_config.rs
```diff
@@ -37,81 +35,145 @@ pub struct ValidatorConfig {
...
-                .set(constants::VALIDATOR_CONFIG, txn)
-                .map_err(|e| {
-                    Error::RemoteStorageWriteError(constants::VALIDATOR_CONFIG, e.to_string())
-                })?;
-        }
+        Ok(Transaction::UserTransaction(signed_txn))
+    }
 
-        Ok(txn)
+    /// Retrieves the owner key from the remote storage using the owner name given by
```
davidiw, line 159: ahh good call

config/management/src/genesis.rs
```diff
@@ -51,9 +52,9 @@ impl Genesis {
         Ok(genesis)
     }
 
-    /// Retrieves association key from the remote storage. Note, at this point in time, genesis
+    /// Retrieves the association key from the remote storage. Note, at this point in time, genesis
     /// only supports a single association key.
-    pub fn association(&self, layout: &Layout) -> Result<Ed25519PublicKey, Error> {
+    pub fn association_key(&self, layout: &Layout) -> Result<Ed25519PublicKey, Error> {
```
gregnazario, line 57: Should we reconcile the terms `faucet_key`, `root_key`, and `association_key` into one consistent term?

config/management/src/config_builder.rs
```diff
@@ -146,21 +185,38 @@ impl<T: AsRef<Path>> BuildSwarm for ValidatorBuilder<T> {
...
+        for index in 0..self.num_validators {
+            self.set_validator_operator(index);
+        }
+
+        // Upload the validator configs for each node
         let mut configs: Vec<_> = (0..self.num_validators)
             .map(|i| self.initialize_validator_config(i))
             .collect();
-        self.storage_helper
+
```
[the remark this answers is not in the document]
JoshLind: Sure, done, thanks!

config/management/src/config_builder.rs
```diff
@@ -146,21 +185,38 @@ impl<T: AsRef<Path>> BuildSwarm for ValidatorBuilder<T> {
...
+        let _ = self
+            .storage_helper
             .create_waypoint(constants::COMMON_NS)
             .unwrap();
         for (i, config) in configs.iter_mut().enumerate() {
             self.finish_validator_config(i, config);
         }
-        Ok((configs, faucet_key))
+
+        Ok((configs, association_key))
```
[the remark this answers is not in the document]
JoshLind: Yes, there's a lot of places in the code that are inconsistently using faucet/root/association key. Let's handle this refactor in a different PR.

config/management/src/validator_operator.rs
JoshLind: Done, thanks! For now, we just upload the name, and then later use that name to set the operator account in genesis.

config/management/src/lib.rs
JoshLind: This test is a subset of the end-to-end test. It was useful for testing the 2nd commit, but I've removed it now.

config/management/src/validator_config.rs
JoshLind: As discussed, we can remove this requirement later on (once we update this tool to support post-genesis operations). For now though, we'll leave it in to be explicit.

config/management/src/lib.rs
JoshLind: Interesting.. I kinda liked being able to assert the outcome of the given CLI command (i.e., ensure that it's doing what we expect on the backend. It definitely made this PR easier to reason about/verify correctness. If you would prefer not to test this, I can just truncate the backend assertions.

config/management/src/genesis.rs
JoshLind: Agreed. As mentioned above, we'll delegate this to a follow up PR.

JoshLind:
Thanks @davidiw! It's ready for another look, when you are!

JoshLind:

Issue #4963, opened 2020-07-08
[Bug] Validator to Safety-Rules networking is extremely brittle


mgorven:
# 🐛 Bug

We deployed a bugfix to pre-mainnet which involved restarting the validator pod but not the safety-rules pod. 7 nodes got into a state where the validator was stuck communicating with safety-rules. Operators had to manually restart both the validator and safety-rules pods to get them going again.

The validator logs show `Connection established to upstream 10.100.106.73:6185`, but not `RoundManager started`. The safety-rules logs had no messages since the validator restarted.

I believe that safety-rules thought the connection from the previous validator process was still alive, and it only handles one connection at a time. The new validator process established a TCP connection (because this is handled by the kernel on the remote side), but when it tried to communicate with safety-rules there was no response. There are two main problems here:

* safety-rules doesn't handle new connections if it thinks there is an existing one. It should accept new connections, and probably terminate any old ones if we only want it handling one at a time.
* The validator didn't time out waiting for safety-rules. It should have timeouts for all operations, logging and reconnecting.

This communication channel has been the likely cause of numerous issues in the past, we need to take a thorough look at the connection handling and ensure that it is robust.

mgorven:
cc @sherry-x @gregnazario

mgorven:
S204936

gregnazario:
Okay, so this sounds like a combined safety-rules/networking problem of multiple things.
1. Connections are being bottlenecked / blocked by previous connections in safety rules
2. Needs more timeouts

David I know doesn't get back until tomorrow, how urgent / bad was it?

sherry-x:
@gregnazario currently we resolved this issue on Pre-mainnet with a restart for both validator and safety-rule pods, so it's mitigated for now. this can wait until david come back

davidiw:
Thanks for the issue. Couple of notes:
* In most cases, I would have expected peers to cleanly close, any idea what happened to the original validator node? In my local tests, I turned off a validator node and connected a new one and it worked without issue. I would like to be able to repro this.
* In the cases where it doesn't cleanly close, we should expect TCP connections to eventually timeout (https://unix.stackexchange.com/questions/519501/default-tcp-keepalive-settings) however in the circumstances where the issue was being exhibited, the default is unacceptable.

I think moving forward, the easiest solution is to use a 30 second read timeout on an accepted socket. If the read doesn't come in within the expected period of time, tear down the connection and start listening again.

[...]

PR #4982, opened 2020-07-09
[secure-net] add timeouts

davidiw:
This improve reliability on services that do not gracefully cleanup sockets by enforcing a 30 second blocking period before trying again.

* LSR this is a no-op
* Consensus will treat all LSR errors as debug and wait for next op
(this is a todo), for now, it will result in a panic in many cases, but
that's better than requiring an admin to manually restart.
* LEC this is a no-op
* Consensus has a lot of different entry points, some of them are part
of bootup and should likely panic and others are loggable errors.

In the case of panics, these seem like optimizations to improve recovery
of the node. If we hit a panic due to a faulty service, this behavior is
better than manual restarting from an admin. At the same time the
outcome can be improved if we can gracefully handle timeouts / retries.

See https://github.com/libra/libra/issues/4963
I believe this should address the concerns, additional logging is up to the users of this framework to log errors: https://github.com/libra/libra/blob/master/consensus/safety-rules/src/remote_service.rs#L34

### DIFF
--- a/secure/net/src/lib.rs
+++ b/secure/net/src/lib.rs
@@ -23,6 +23,12 @@ use std::{
 };
 use thiserror::Error;
 
+/// Read, Write, Connect timeout in seconds.
+#[cfg(not(test))]
+const TIMEOUT: u64 = 30;
+#[cfg(test)]
+const TIMEOUT: u64 = 5;
+
 #[derive(Debug, Error)]
 pub enum Error {
     #[error("Already called shutdown")]
@@ -82,14 +88,15 @@ impl NetworkClient {
 
     fn server(&mut self) -> Result<&mut NetworkStream, Error> {
         if self.stream.is_none() {
+            let timeout = std::time::Duration::from_secs(TIMEOUT);
             debug!("Attempting to connect to upstream {}", self.server);
-            let mut stream = TcpStream::connect(self.server);
+            let mut stream = TcpStream::connect_timeout(&self.server, timeout);
 
             let sleeptime = time::Duration::from_millis(100);
             while let Err(e) = stream {
                 debug!("Failed to connect to upstream {} {:?}", self.server, e);
                 thread::sleep(sleeptime);
-                stream = TcpStream::connect(self.server);
+                stream = TcpStream::connect_timeout(&self.server, timeout);
             }
 
             let stream = stream?;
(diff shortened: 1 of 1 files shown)


secure/net/src/lib.rs
```diff
@@ -351,4 +363,57 @@ mod test {
...
+        client.write(&data1).unwrap();
+        let result1 = server.read().unwrap();
+        assert_eq!(data1, result1);
+
+        // Timedout
+        server.read().unwrap_err();
+
+        // New client, success, note the previous client connection is still active, the server is
+        // actively letting it go due to lack of activity.
+        let mut client2 = NetworkClient::new(server_addr);
```
gregnazario, line 386: Let's hope it doesn't somehow become a flaky test with a 5 second timeout.

davidiw: I originally had an ignore, but opted against it, if 5 seconds leads to flakiness, I'll feel sad! and we can just add ignore where appropriate.

davidiw:
> is this the issue happening in pre-mainnet? I saw `Connection established to upstream` printed in validators but seems LSR didn't accept the connection for whatever reason.

That's the suspicion (see the tagged issue), I'd love to see a netstat on the safety rules side that shows that we had two connections in established mode.

The alternative of course is to have the health checks cause the service to restart but ...

gregnazario:
> > is this the issue happening in pre-mainnet? I saw `Connection established to upstream` printed in validators but seems LSR didn't accept the connection for whatever reason.
> 
> That's the suspicion (see the tagged issue), I'd love to see a netstat on the safety rules side that shows that we had two connections in established mode.
> 
> The alternative of course is to have the health checks cause the service to restart but ...

Yeah, having health checks cause a service to restart can be pretty dangerous without some tuning to make sure it isn't restarting all the time.

secure/net/src/lib.rs
```diff
@@ -351,4 +363,57 @@ mod test {
...
+        let result1 = server.read().unwrap();
+        assert_eq!(data1, result1);
+
+        // Timedout, it is hard to simulate a client receiving a write timeout
+        client.read().unwrap_err();
+
+        // Clean up old Server listener but keep the stream online. Start a new server, which will
+        // be the one the client now connects to.
+        server.listener = None;
+        let mut server2 = NetworkServer::new(server_addr);
```
zekun000, line 412: does it work with the same client/server after timeout?

davidiw: It should, will expand the test.

davidiw:
Forgot to land for branch cut...

@bors-libra r+

PR #4988, opened 2020-07-09
[libra framework] Remove entity-based freezing.

### LINKED ISSUE #4986: Make withdrawal capabilities respect account freezing.
Right now coins can be withdrawn from a frozen account if the withdrawal capability has previously been extracted. We need to prevent this from happening.

tzakian:
This PR removes per-entity-based freezing, and we now only have per-account freezing. 

If an account is frozen:
* Money can't be withdrawn from the account.
* Money can't be sent to the account.
* Transactions can't be sent from the account.

Closes #4986


junkil-park:
> [move-prover] Move Prover test (`language/move-prover/cargo test`) on commit [a9bda9e](https://github.com/libra/libra/commit/a9bda9e5965f87da3d09606018b1baf1aab0181f) did **fail**. See [details](https://github.com/mimoo/mirai-bot/runs/856061435?check_suite_focus=true#step:7:1).

Does this happen after rebasing? Perhaps, @emmazzz knows how to fix the precondition violation error in `Libra.move`?

junkil-park:
Although it's unrelated to this PR, it seems that the comments right above `LibraAccount::make_account` does not match well.

tzakian:
Working on fixing the precondition failures in `LibraAccount` right now. IIRC the failures in Libra.move are from a different PR.

sblackshear:
If https://github.com/libra/libra/pull/4972 manages to land, maybe try rebasing--managed to get all prover tests passing there.

language/stdlib/modules/AccountFreezing.move
```diff
@@ -105,11 +105,15 @@ module AccountFreezing {
     /// Returns if the account at `addr` is frozen.
     public fun account_is_frozen(addr: address): bool
     acquires FreezingBit {
-        borrow_global<FreezingBit>(addr).is_frozen
+        exists<FreezingBit>(addr) && borrow_global<FreezingBit>(addr).is_frozen
      }
```
sblackshear, line 110: I believe you need to add 
```
spec fun account_is_frozen {
   pragma verify = true;
   ensures spec_account_is_frozen(addr);
}
```
to verify the spec fun against the impl (is that right and/or is there a more concise way @junkil-park)?

junkil-park: Yes, that's right. `pragma verify = true;` can be omitted because verification is already turned on at the module level (i.e., "pragma verify" in "spec module"). Also, "aborts_if false;" could be added.

PR #4990, opened 2020-07-09
[vm] NativeResult can only abort

tnowacki:
- NativeResult for native functions can now only give back an abort code
- The native functions can technically still error, but hopefully that is more discouraged

## Motivation

- Bring natives inline with other functions and give them only the ability to "abort"

## Test Plan

- cargo test


tnowacki:
> I assume they should only do this if some internal invariants are broken, right?

Thats the intent yeah, but there is nothing enforcing that intent. The current ones that would fail are things like events, which might fail to write

tnowacki:
Fixed tests

PR #4991, opened 2020-07-10
[tcb] add -> checked_add

mimoo:
Following a discussion with Herman,
this will wrap around in release mode and might lead to subtle attacks[1].
We never know, so let's just crash instead if we can.

[1]: https://huonw.github.io/blog/2016/04/myths-and-legends-about-integer-overflow-in-rust/#myth-overflow-is-unspecified

### DIFF
--- a/consensus/safety-rules/src/safety_rules.rs
+++ b/consensus/safety-rules/src/safety_rules.rs
@@ -99,20 +99,33 @@ impl SafetyRules {
     /// 1) B0 <- B1 <- B2 <--
     /// 2) round(B0) + 1 = round(B1), and
     /// 3) round(B1) + 1 = round(B2).
-    pub fn construct_ledger_info(&self, proposed_block: &Block) -> LedgerInfo {
+    pub fn construct_ledger_info(&self, proposed_block: &Block) -> Result<LedgerInfo, Error> {
+        // get rounds of the three blocks
         let block2 = proposed_block.round();
         let block1 = proposed_block.quorum_cert().certified_block().round();
         let block0 = proposed_block.quorum_cert().parent_block().round();
 
-        let commit = block0 + 1 == block1 && block1 + 1 == block2;
-        if commit {
+        // checked additions
+        let block0plus1 =
+            u64::checked_add(block0, 1).ok_or(Error::IncorrectRounds(block0, block1, block2))?;
+        let block1plus1 =
+            u64::checked_add(block1, 1).ok_or(Error::IncorrectRounds(block0, block1, block2))?;
+
+        // verify 3-chain rule
+        let commit = block0plus1 == block1 && block1plus1 == block2;
+
+        // create a ledger info
+        let ledger_info = if commit {
             LedgerInfo::new(
                 proposed_block.quorum_cert().parent_block().clone(),
                 HashValue::zero(),
             )
         } else {
             LedgerInfo::new(BlockInfo::empty(), HashValue::zero())
-        }
+        };
+
+        // return
+        Ok(ledger_info)
     }
 
     /// Second voting rule
@@ -309,7 +322,7 @@ impl TSafetyRules for SafetyRules {
         let vote = Vote::new(
             vote_data,
             validator_signer.author(),
-            self.construct_ledger_info(proposed_block),
+            self.construct_ledger_info(proposed_block)?,
             validator_signer,
         );
         self.persistent_storage.set_last_vote(Some(vote.clone()))?;
--- a/consensus/safety-rules/src/safety_rules.rs
+++ b/consensus/safety-rules/src/safety_rules.rs
@@ -100,19 +100,14 @@ impl SafetyRules {
     /// 2) round(B0) + 1 = round(B1), and
     /// 3) round(B1) + 1 = round(B2).
     pub fn construct_ledger_info(&self, proposed_block: &Block) -> Result<LedgerInfo, Error> {
-        // get rounds of the three blocks
         let block2 = proposed_block.round();
         let block1 = proposed_block.quorum_cert().certified_block().round();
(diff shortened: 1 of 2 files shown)


consensus/safety-rules/src/safety_rules.rs
```diff
@@ -104,7 +104,8 @@ impl SafetyRules {
         let block1 = proposed_block.quorum_cert().certified_block().round();
         let block0 = proposed_block.quorum_cert().parent_block().round();
 
-        let commit = block0 + 1 == block1 && block1 + 1 == block2;
+        let commit = u64::checked_add(block0, 1).unwrap() == block1
+            && u64::checked_add(block1, 1).unwrap() == block2;
```
davidiw, line 108: considering unwrap will cause a panic, I think the correct behavior would be to return a result here and fail safely.

hermanventer:
I think this is OK because certified and parent blocks will never be able to get to the max value because they can only exist in the block chain if there have been subsequent rounds. So I don't expect this to ever fail. But doing the check here makes it more likely that future mistakes will be caught before making it into production.

Perhaps add the gist of the above in a comment to justify why the checked_add is not expected to ever fail.

consensus/safety-rules/src/safety_rules.rs
hermanventer: I assume that such a failure will get logged somewhere, of course.

mimoo: done!

mimoo:
We actually can in theory fail here as we accept this directly from consensus and a malicious consensus could send us wrong rounds. That being said the fact that the TCB does not protect against liveness attacks is probably documented in the specification. (It'd still be nice to have a README or some [module-level doc](https://libra.github.io/libra/safety_rules/index.html) *wink wink*)

davidiw:
Thanks for addressing this, this is important work.

mimoo:
the wellformedness check you linked to actually doesn't allow to verify that "grandparent < parent < proposal < max u64"

but I think we can totally add such a check in order to thwart subtle bugs.
Hell we can check that whatever blocks we see in proposals or votes always have versions, rounds, and epochs lower than `u64::MAX - X` for X = 5, 10, or whatever number.

Wdyt?

davidiw:
Yeah, I think that's the right way to go, that way we isolate Errors cleanly and don't end up add Results or unwraps and checks in random places.

mimoo:
zekun disagrees :P

zekun000:
as herman and david said, a valid block will never be able to get to the max value because they can only exist in the block chain if there have been subsequent rounds.

we should focus on correct validity check rather than adding redundant defensive checks, so definitely not in consensus, I'll let @davidiw dictates tcb.

hermanventer:
There is a valid point here: input validation should be performed at most once and this should be done in the right place. Validation checks should not crash the program, obviously. Once validation has been done and we have reason to believe that a certain condition is no longer possible it is, however, actually OK to redundantly check it defensively. You can do this too often and too much of a good thing is no longer a good thing, but it can be a good thing. The heuristic for it being a good thing is when it adds documentation and clarity. When reading code locally, it is good to have reminders about the relevant invariant conditions. This makes the code easier to read in parts and easier to debug when a subtle change breaks the invariant in some cases.

From a debugging point of view it is best if the program stops immediately if a supposedly redundant check finds that the invariant does not actually hold. This does, however, provide a convenient attack vector for an attacker that has discovered a subtle bug in our code. So, it probably makes more sense to fail the transaction gracefully and carry on. In the next year or so, it might make more sense to statically check, using MIRAI, that invariants always hold. At that point, the runtime checks can become compile time only assertions.

Right now I think this PR, as it stands, makes enough sense to merge.

consensus/safety-rules/src/safety_rules.rs
```diff
@@ -99,20 +99,30 @@ impl SafetyRules {
...
             LedgerInfo::new(
                 proposed_block.quorum_cert().parent_block().clone(),
                 HashValue::zero(),
             )
         } else {
             LedgerInfo::new(BlockInfo::empty(), HashValue::zero())
-        }
+        };
+
+        // return
```
davidiw, line 124: I think this is quite clear to any rustian, remove?

consensus/safety-rules/src/safety_rules.rs
```diff
@@ -99,20 +99,30 @@ impl SafetyRules {
     /// 1) B0 <- B1 <- B2 <--
     /// 2) round(B0) + 1 = round(B1), and
     /// 3) round(B1) + 1 = round(B2).
-    pub fn construct_ledger_info(&self, proposed_block: &Block) -> LedgerInfo {
+    pub fn construct_ledger_info(&self, proposed_block: &Block) -> Result<LedgerInfo, Error> {
+        // get rounds of the three blocks
```
davidiw, line 103: not sure the comments add value since we have the doc block above?

consensus/safety-rules/src/safety_rules.rs
```diff
@@ -99,20 +99,30 @@ impl SafetyRules {
...
         let block2 = proposed_block.round();
         let block1 = proposed_block.quorum_cert().certified_block().round();
         let block0 = proposed_block.quorum_cert().parent_block().round();
 
-        let commit = block0 + 1 == block1 && block1 + 1 == block2;
-        if commit {
+        // verify 3-chain rule
+        fn next_round(round: u64) -> Result<u64, Error> {
```
davidiw, line 109: why not use a lambda `let increment = |round| u64::check_add(round, 1).ok_or(Error::IncorrectRound(round))`

mimoo: I find it less clear personally but I can change it

PR #4992, opened 2020-07-10
[move-prover][libra-framework] True modular verification.

wrwg:
This PR deals with the various long running verification problems by changing our compilation scheme, and realizing true modular verification. As a result, most of the disabled verification problems in the framework could be re-enabled (around two dozen functions had been disabled because of timeout). Some are still remaining which need further investigation (3 or 4 or such).

The main insights underlying this PR are the followings:

- Only the function currently being verified should provide assumptions to the VC context (VC = Verification Condition, essential the smtlib file we give to z3 for each function). As experiments showed, if we saturate the VC context with assumptions not relevant for the VC at hand, Z3 goes nuts, trying to endlessly instantiate quantifiers which are irrelevant for the VC. Sometimes this can be fixed by changing seeds for Z3, but sometimes not.
- Module invariants are private to the module. When a module function is called from another module, we do not need to assume the module invariant because we also do not need to assert anything about it.
- If we have have non-opaque function which is not subject of the current VC being called and inlined from the VC function, it is more efficient to not "freely" assume or assert any pre/post conditions, but instead let the code speak for itself. Otherwise we again pollute the VC context with useless assumptions. In fact, the top-level VC function might constitute a much simpiler VC problem than the called function.

This has been worked into the compilation schema as follows:

- We now have 6 (yes 6) different entry points into a function, each one serving a different purpose in verification:
  - Definition: just the definition of the function, no pre/post conditions. Is being inlined by the others below.
  - DirectInter: a direct call from the VC function to this function which is in a different module. For such calls, we only assert the explicit pre-conditions, but not those coming from the module invariants or schema apply. We only emit an `ensures` for direct calls if they are opaque.
  - DirectIntra: a direct call from the VC functions to this function which is in the same module. For such calls, we assert explicit preconditions as well as the ones injected from module invariants. Ensures is treated as above.
  - Indirect: an entrypoint for calls not from the VC function, but from functions called by the VC function. No preconditions are generated for those, because we verified them elsewhere. Ensures are only generated if the function is opaque.
  - VerificationDefinition: the definition we use for verification. The difference to `Definition` is that we use the DirectXXX stubs to call other functions from this one, in contrast to Indirect.
  - Verification: finally the top-level verification entry point. This imposes all pre and post conditions and inlines VerificationDefinition.

There is a way to override the omission of injected predicates from schema apply or module invariant, so that they are exported to the caller context. For example, to inject publically visible aborts spec into multiple functions (as used in VASP.move) one can write:

```
schema CommonAborts {
    aborts_if [export] P;
}
apply CommonAborts to f, g, h;
```

*Note*: I had to disable verification for `Libra.move` as it was broken at head. A fix for this does not belong into this PR (it looks complicated). However, before I rebased, Libra passed, so the failures should be orthogonal to this PR.

## Motivation

Fixing timeouts.

Yes

## Test Plan

Existing tests.

## Related PRs

NA


language/move-prover/spec-lang/src/env.rs
```diff
@@ -84,6 +88,14 @@ pub const ADDITION_OVERFLOW_UNCHECKED_PRAGMA: &str = "addition_overflow_unchecke
 /// Pragma indicating that aborts from this function shall be ignored.
 pub const ASSUME_NO_ABORT_FROM_HERE_PRAGMA: &str = "assume_no_abort_from_here";
 
+/// Internal property attached to conditions if they are injected via an apply or a module
+/// invariant.
+pub const CONDITION_INJECTED_PROP: &str = "$injected";
```
shazqadeer, line 93: I suppose this is how you are distinguishing between explicitly-written preconditions and postconditions and ones injected from module invariants.

language/move-prover/spec-lang/src/env.rs
```diff
@@ -857,6 +869,24 @@ impl GlobalEnv {
             .iter()
             .for_each(|(l, i)| f(l, i))
     }
+
+    /// Returns true if the boolean property is true.
+    pub fn is_property_true(&self, properties: &PropertyBag, name: &str) -> Option<bool> {
```
shazqadeer, line 874: Is this "property" something like the attribute mechanism in Boogie?  What things can have properties attached to them?

language/move-prover/spec-lang/src/env.rs
wrwg: Yes

language/move-prover/spec-lang/src/env.rs
wrwg: Pretty much the same. Currently it can appear in form of a `pragma` at functions and in conditions (`<cond> [prop] <predicate>`)

PR #4993, opened 2020-07-10
[language][vm] implement recursion depth checks for SignatureToken

vgao1996:
## Summary
This sets a maximum depth for `SignatureToken`, enforced in both the serializer and deserializer.

## Test Plan
cargo test, new tests added

### DIFF
--- a/language/vm/src/deserializer.rs
+++ b/language/vm/src/deserializer.rs
@@ -819,7 +819,9 @@ fn load_signature_tokens(cursor: &mut Cursor<&[u8]>) -> BinaryLoaderResult<Vec<S
 }
 
 /// Deserializes a `SignatureToken`.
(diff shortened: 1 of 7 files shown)


language/vm/src/deserializer.rs
```diff
@@ -819,7 +819,9 @@ fn load_signature_tokens(cursor: &mut Cursor<&[u8]>) -> BinaryLoaderResult<Vec<S
 }
 
 /// Deserializes a `SignatureToken`.
-fn load_signature_token(cursor: &mut Cursor<&[u8]>) -> BinaryLoaderResult<SignatureToken> {
+pub(crate) fn load_signature_token(
```
vgao1996, line 822: I really don't want to expose this function (and a bunch of other stuff), but I don't see how otherwise I'll be able to test it in a reasonable way. Please let me know if you have a better solution.

PR #4994, opened 2020-07-10
[libra framework] Update minting of LBR to take the exact amount of coins

tzakian:
`LBR::create` no longer returns the remaining coins, and instead the precise amount for each coin according to the new `calculate_component_amounts_for_lbr` function in the `LBR`  must be passed in. 

This also means that we can remove the redeposit of coins in `LibraAccount::staple_lbr`. 

Won't land until verification checks pass.


language/stdlib/modules/LBR.move
```diff
@@ -102,6 +105,17 @@ module LBR {
             Libra::currency_code<CoinType>() == Libra::currency_code<LBR>()
     }
 
+    /// We take the truncated multiplication + 1 (not ceiling!) to withdraw for each currency that makes up the `LBR`.
+    /// We do this to ensure that the reserve is always positive. We could do this with other more complex methods such as
+    /// bankers rounding, but this adds considerable arithmetic complexity.
```
sblackshear, line 110:
```suggestion
    /// banker's rounding, but this adds considerable arithmetic complexity.
```

tzakian: Good questions:
* I don't think this will clean those up since those use `LibraAccount::staple_lbr`. But I may be missing something?
* We can make `LBR::swap_into` private (well, we can remove it). I'll do that in this PR.

tzakian:
@bors-libra r=sblackshear

Checked mirai bot to make sure the most run of the prover passed.

PR #4995, opened 2020-07-10
[consensus] add an Epoch type (to mimic the Round type)

mimoo:


consensus/consensus-types/src/common.rs
```diff
@@ -3,12 +3,14 @@
 
 use libra_types::{account_address::AccountAddress, transaction::SignedTransaction};
 
-/// The round of a block is a consensus-internal counter, which starts with 0 and increases
-/// monotonically. It is used for the protocol safety and liveness (please see the detailed
-/// protocol description).
+/// Consensus starts at epoch 0 and increments it periodically.
```
davidiw, line 6: increments at reconfiguration events

mimoo:
One way or the other works. But I thought we should be symmetric yeah. 

I was actually thinking that we could potentially enforce checked arithmetic if we use a wrapper type at some point. I'm experimenting with checks like these that are making me a bit antsy: https://github.com/libra/libra/pull/4997/files#diff-7838dea25f2d82e521edb77e6f370ecbR47

davidiw:

PR #4996, opened 2020-07-10
[CI][easy] tune sccache

sausagee:
## Motivation
Like conventional CPU cache design, larger cache size reduces miss. However, in our CI setup, when the sccache size reaches certain point, scale appears to tip - the hit (time spent in searching and returning) could outweigh the miss (time spent compiling as is). 

## Test Plan
Testing in CI.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -105,9 +105,10 @@ commands:
             if ! [ -e /usr/local/cargo/bin/sccache ]; then
               $CARGO $CARGOFLAGS install sccache --version=0.2.13
             else
+              sccache --version
               echo 'sccache binary is found. Skipping install.'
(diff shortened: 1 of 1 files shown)


rexhoffman:
@sausagee should we merge this?

sausagee:
I am not convinced by the gains so far.  Want to do one more iteration.

sausagee:
After @metajack's v2 feature resolver landed, the cache size required also dropped nearly half from >4G to ~2G. Further optimization looks like diminishing return. 

https://app.circleci.com/pipelines/github/libra/libra/24831/workflows/b7826b3d-929a-4371-8b08-aa16462b239f/jobs/159581
image

PR #4997, opened 2020-07-10
[logging] security logging for consensus votes

mimoo:
This adds a way to detect votes that display byzantine behavior.
If two non-NIL block end up voting for two different consensus_data_hash at the same round,
then it means that either:

- the proposer sent different blocks to different voters
- voters executed the same block differently

to detect this, we simply add a cache in pending_votes to save non-NIL votes by consensus_data_hash.

### DIFF
--- a/consensus/consensus-types/src/vote_data.rs
+++ b/consensus/consensus-types/src/vote_data.rs
@@ -40,6 +40,37 @@ impl VoteData {
         &self.proposed
     }
 
+    /// Returns true if the vote is on a proposal block
+    pub fn block_type_proposal(&self) -> bool {
+        if self.proposed.timestamp_usecs() == self.parent.timestamp_usecs() {
+            // NIL or genesis block
+            return false;
+        }
+        // proposals always have a block metadata transaction
+        self.proposed.version() >= self.parent.version().saturating_add(1)
+    }
+
+    /// Returns true if the vote is on a NIL block
+    pub fn block_type_nil(&self) -> bool {
+        if self.proposed.timestamp_usecs() != self.parent.timestamp_usecs() {
+            // proposal block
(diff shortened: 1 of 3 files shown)


consensus/consensus-types/src/vote_data.rs
```diff
@@ -40,6 +40,12 @@ impl VoteData {
         &self.proposed
     }
 
+    /// Returns true if the vote is on a NIL block
+    pub fn is_nil(&self) -> bool {
+        self.proposed.timestamp_usecs() == self.parent.timestamp_usecs()
```
mimoo, line 45: so this is wrong, another way I can check this is `proposed.version = parent.version + 1`. That should work all the time IIUC (cc @zekun000)

zekun000: no, a nil block can be reconfiguration suffix too which have the same version

mimoo: so I would have to filter any block that has more than one version added to it for this PR to work... I feel like I'm missing something about block metadata though, I'll dig a bit more.

consensus/consensus-types/src/vote_data.rs
```diff
@@ -40,6 +40,14 @@ impl VoteData {
         &self.proposed
     }
 
+    /// Returns true if the vote is on a block of type proposal
+    /// This assumes that the QC of parent was verified
+    /// This also doesn't find proposals that have 0 transactions
+    pub fn block_type_proposal(&self) -> bool {
+        self.proposed.version() > self.parent.version() + 1
```
mimoo, line 47: I should use `saturating_add()` here

davidiw:

PR #4998, opened 2020-07-10
[libra framework] add doc comments for vector, option, lcs modules

sblackshear:
Tweaking the comments that were already there and making them into doc comments.

### DIFF
--- a/language/stdlib/modules/Vector.move
+++ b/language/stdlib/modules/Vector.move
@@ -1,41 +1,47 @@
 address 0x1 {
 
-// A variable-sized container that can hold both unrestricted types and resources.
+/// A variable-sized container that can hold both unrestricted types and resources.
 module Vector {
 
     const EINDEX_OUT_OF_BOUNDS: u64 = 0;
 
+    /// Create an empty vector.
     native public fun empty<Element>(): vector<Element>;
 
-    // Return the length of the vector.
+    /// Return the length of the vector.
     native public fun length<Element>(v: &vector<Element>): u64;
 
-    // Acquire an immutable reference to the ith element of the vector.
+    /// Acquire an immutable reference to the `i`th element of the vector `v`.
+    /// Aborts if `i` is out of bounds.
     native public fun borrow<Element>(v: &vector<Element>, i: u64): &Element;
 
-    // Add an element to the end of the vector.
+    /// Add element `e` to the end of the vector `v`.
     native public fun push_back<Element>(v: &mut vector<Element>, e: Element);
 
-    // Get mutable reference to the ith element in the vector, abort if out of bound.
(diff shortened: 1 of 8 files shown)


language/stdlib/modules/Vector.move
```diff
@@ -1,41 +1,45 @@
...
 
+    /// Create an empty vector.
     native public fun empty<Element>(): vector<Element>;
 
-    // Return the length of the vector.
+    /// Return the length of the vector.
     native public fun length<Element>(v: &vector<Element>): u64;
 
-    // Acquire an immutable reference to the ith element of the vector.
+    /// Acquire an immutable reference to the `i`th element of the vector `v`.
```
tzakian, line 14: Nit maybe add the same "aborts if i is out of bounds" comment here as in `borrow_mut`

language/stdlib/modules/Vector.move
```diff
@@ -85,8 +89,8 @@ module Vector {
     }
 
 
-    // Remove the `i`th element E of the vector, shifting all subsequent elements
-    // It is O(n) and preserves ordering
+    /// Remove the `i`th element of the vector `v`, shifting all subsequent elements.
+    /// This is O(n) and preserves ordering of elements in the vector.
```
tzakian, line 93: Nit: Add "aborts if i  is out of bounds"

language/stdlib/modules/Vector.move
```diff
@@ -97,8 +101,8 @@ module Vector {
         pop_back(v)
     }
 
-    // and then popping it off
-    // It is O(1), but does not preserve ordering
+    /// Swap the `i`th element of the vector `v` with the last element and then pop the vector.
+    /// This is O(1), but does not preserve ordering of elements in the vector.
```
tzakian, line 105: Nit: add "aborts if i is out of bounds"

language/stdlib/modules/Vector.move
```diff
@@ -72,8 +76,8 @@ module Vector {
         false
     }
 
-    // Return (true, i) if `e` is in the vector `v` at index `i`.
-    // Otherwise returns (false, 0).
+    /// Return `(true, i)` if `e` is in the vector `v` at index `i`.
+    /// Otherwise, returns `(false, 0)`.
     public fun index_of<Element>(v: &vector<Element>, e: &Element): (bool, u64) {
```
tzakian, line 81: A hold-over from the non-option days...I'll file an issue to change this to an option.

sblackshear: SGTM, although the fact that `Option` is implemented using `Vector` may make this tricky...

tzakian: oh....you're right...nevermind! I keep forgetting that our options aren't really options.

PR #4999, opened 2020-07-10
[consensus] better well-formedness check for VoteData

mimoo:
Instead of checking that (from the spec):

- ensure parent.timestamp_usecs <= proposed.timestamp_usecs
- ensure parent.version <= proposed.version

we should check stricter assumptions based on if the vote is a NIL vote (i.e. on a NIL block)

### DIFF
--- a/consensus/consensus-types/src/vote_data.rs
+++ b/consensus/consensus-types/src/vote_data.rs
@@ -1,6 +1,7 @@
 // Copyright (c) The Libra Core Contributors
 // SPDX-License-Identifier: Apache-2.0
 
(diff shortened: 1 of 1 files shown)


mimoo:
Ah right, so I think we can check if it's a NIL block by just checking that there is only one transaction (the blockmetadata). Does the reconfiguration suffix have that one as well?

zekun000:
no, but reconfiguration suffix can be NIL block too

mimoo:
need to think more about that. We can theoretically check that version = version + 1 but we risk wrapping around

PR #5001, opened 2020-07-10
[JSON-RPC] Expose preburn balance in AccountView

moeziniaf:
EDITED: 
Preburn balance is like LibraAccount balance. As discussed below, for now we will expose the preburn balances in the `AccountRoleView::DesignatedDealer`. There it is represented as a `Vec<AmountView>` since we have multi currency DD accounts now and DDs need to know their currency/preburn_amount breakdown.
I have `PreburnResource` decoupled from account balance in case the two resource types diverge in the future.

Test:
```curl http://localhost:57425 -X POST --data '{"jsonrpc":"2.0","method":"get_account","params":["0000000000000000000000000000000000000000000000dd"],"id":1}’```
 
on testnet's 0xdd account yields

```......."role":{"designated_dealer":{"base_url":"https://libra.org","expiration_time":18446744073709551615,"human_name":"moneybags","preburn_balances":[{"amount":0,"currency":"Coin1"},{"amount":0,"currency":"Coin2"}]}},".....```

The 0xdd can only hold non-LBR currencies, so the relevant preburn balances are shown.

### CODE REVIEW

**Comment by runtian-zhou** [comment]
To test this I think you should run `cargo run -p libra_swarm` and it should spawn up a json-rpc server locally, say `http://localhost:60783`. You can then run 
```
curl http://localhost:60783 -X POST -H "Content-Type: application/json" --data '{"jsonrpc":"2.0","method":"get_account_state","params":["c1fda0ec67c1b87bfb9e883e2080e530"],"id":1}'
```

And check if the response is as expected.

**Comment by runtian-zhou** [comment]
Personally I like the idea of putting this balance in another method as well, because not all accounts have this `Preburn` resource.

**Comment by tzakian** [comment]
One other option would be to put this under the `AccountRoleView` for `DesignatedDealer`'s. Since those are the only accounts that really will have any preburns. In terms of testing, you should be able to run queries against 0xdd which is the wealthy designated dealer we use for minting on testnet.

**Comment by dmitri-perelman** [comment]
The idea of extending different role views to keep the specific per-role information seems both intuitive and simple to implement.

**Comment by xli** [comment]
> actually after discussion, we should not add it to AccountView, instead I propose we create an new method called aos_get_preburn_balances() instead.
> 
> and this method should have built in account to look, not relying on the user to supply it

client side does not have the AOS account address? otherwise, for a given AOS account, we return it's preburn balances seems right to me.

@moeziniaf it would be great if we could get test in

**Comment by moeziniaf** [comment]
@tzakian how can I trigger a preburn for 0xdd on testnet while also querying the preburn balance with my local changes? Doing so locally was impossible because we don't have the seed to initialize the blessed account to send a valid preburn txn.

**Comment by thefallentree** [comment]
> The idea of extending different role views to keep the specific per-role information seems both intuitive and simple to implement.

as currently designed in the code level, maybe, but that could change again since it is entirely up to Move to structure that data. 

By creating an new method/new view , we will make the API independent on however Move is structuring it, which make the interface more stable

**Comment by tzakian** [comment]
@moeziniaf good point. I don't think there is a way of triggering a preburn sadly. You would only be able to see preburn balance of 0 :(

**Comment by dmitri-perelman** [comment]
It seems that there are several opinions and no consensus here. I'd like us to not block the DDs: should we setup a meeting?

**Comment by dmitri-perelman** [comment]
Looked at it again: we already have the `AccountRoleView::DesignatedDealer` for exposing information for dual attestation (similarly to the way we're exposing it with VASPs). Seems like adding a pre-burn balance to there is absolutely no-brainer.

**Comment by thefallentree** [comment]
> Looked at it again: we already have the `AccountRoleView::DesignatedDealer` for exposing information for dual attestation (similarly to the way we're exposing it with VASPs). Seems like adding a pre-burn balance to there is absolutely no-brainer.

I still strongly advocate for create an special JSONRPC method for this specific use case, but I'm okay with adding it to AccountRoleView for now to unblock.

**Comment by moeziniaf** [comment]
Added to `AccountRoleView` for now. Please see summary for updated details.

File: types/src/account_config/resources/preburn_balance.rs [hunk 1]
```diff
@@ -0,0 +1,55 @@
...
+impl PreburnResource {
+    pub fn new(coin: u64) -> Self {
+        Self { coin }
+    }
+
+    pub fn coin(&self) -> u64 {
+        self.coin
+    }
+
+    // TODO/XXX: remove this once the MoveResource trait allows type arguments to `struct_tag`.
```
dmitri-perelman, line 32: Not sure I understand the code below: who is the best person to review it?

types/src/account_config/resources/role.rs
```diff
@@ -12,7 +14,11 @@ pub enum AccountRole {
         credential: Credential,
     },
     ChildVASP(ChildVASP),
-    DesignatedDealer(Credential), // TODO: add DesignatedDealer resource as well
+    DesignatedDealer {
+        dd_credential: Credential,
+        preburn_balances: BTreeMap<Identifier, PreburnResource>,
+        // TODO DesignatedDealer resource (tiered mint events)
```
dmitri-perelman, line 20: I guess we need just the key for the events.

types/src/account_state.rs
```diff
@@ -50,6 +50,23 @@ impl AccountState {
...
+        currency_codes
+            .iter()
+            .filter_map(|currency_code| {
+                let currency_type_tag = type_tag_for_currency_code(currency_code.to_owned());
+                // TODO: update this to use PreburnResource::resource_path once that takes type
+                // parameters
+                self.get_resource(&PreburnResource::access_path_for(currency_type_tag))
+                    .transpose()
+                    .map(|preburn_balance| preburn_balance.map(|b| (currency_code.to_owned(), b)))
+            })
```
dmitri-perelman, line 66: nice!

PR #5002, opened 2020-07-10
[logging] removing and adding network security logs

mimoo:
removing one legacy security log, and adding a security log for failed inbound noise connections

### DIFF
--- a/network/src/transport.rs
+++ b/network/src/transport.rs
@@ -219,7 +219,13 @@ async fn upgrade_inbound<T: TSocket>(
     let socket = fut_socket.await?;
 
     // try authenticating via noise handshake
-    let (socket, peer_id) = ctxt.noise.upgrade_inbound(socket).await?;
+    let (socket, peer_id) = ctxt.noise.upgrade_inbound(socket).await.map_err(|err| {
+        // security logging
+        send_struct_log!(
+            security_log(security_events::INVALID_NETWORK_PEER).data("error", format!("{}", err))
+        );
+        err
+    })?;
     let remote_pubkey = socket.get_remote_static();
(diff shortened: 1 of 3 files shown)


network/src/transport.rs
```diff
@@ -219,7 +219,13 @@ async fn upgrade_inbound<T: TSocket>(
     let socket = fut_socket.await?;
 
     // try authenticating via noise handshake
-    let (socket, peer_id) = ctxt.noise.upgrade_inbound(socket).await?;
+    let (socket, peer_id) = ctxt.noise.upgrade_inbound(socket).await.map_err(|err| {
+        // security logging
+        send_struct_log!(
+            security_log(security_events::INVALID_NETWORK_PEER).data("error", format!("{}", err))
```
gregnazario, line 225: So the `err` here I'm assuming doesn't serialize.  This was actually one of my concerns with the serialization that we use.  Maybe we should create another function to just use `display` instead of `serialize` for cases like this. @andll

andll: Yes, we can add another method like data_display for that :)
I think for most case json serialization is preferred(as it allows to filter in ELK)
But for errors and such we probably will want another method instead of doing this format!.

gregnazario: I'll go ahead and make a PR for it then.

mimoo: nice! can you add me to that one when it's done?

gregnazario: I'll wait for this one to be pushed, and update the instance you're using

phlip9:
i think we can also add the `addr` as context (to see where the inbound is dialing from).

PR #5003, opened 2020-07-10
[move framework] limit number of child VASP accounts created

sblackshear:
This is to keep the storage requirements reasonable for MVP. It is easy to relax this later if we have an urgent need for >256 accounts or if the storage requirements of validators change. Happy with values other than 256 too; I tried to pick a value that was large enough for the child VASP account use-cases that makes sense for Libra (e.g., multiple accounts for easier accounting or to avoid seq # contention, but *NOT* accounts per-customer or per-transaction).

I could have used an on-chain config for this instead of a constant, but that seemed like overkill. I think on-chain configs make sense for values:
- that we need to update frequently (e.g., LBR exchange rates)
- that we might need to update urgently (e.g., Libra version)
- that are sensitive in the sense that they change the client-observable behavior in a significant way (e.g., gas schedule).
- that clients and/or > 1 module want to read

If a value doesn't check any of these boxes, I think we should use a constant instead. A constant is much more lightweight in terms of code (don't need a new module!), AOS process (don't need a new approval flow for updating the constant!), and verifiability.


msmouse:
Forgive my ignorance, but just to confirm, this constant itself is on chain, so VM updates it's expectation at the version we change it, and when we replay old txns the expectation remains the old value, right?

sblackshear:
@msmouse: correct. Module are just ordinary on-chain data, and replaying txes always reads the modules present at that version.

PR #5004, opened 2020-07-10
[types] fix bug in Rust struct for NewEpochEvent

sblackshear:
Struct name here was `NewBlockEvent`, which doesn't look right.

### DIFF
--- a/types/src/account_config/events/new_epoch.rs
+++ b/types/src/account_config/events/new_epoch.rs
@@ -28,5 +28,5 @@ impl NewEpochEvent {
 
 impl MoveResource for NewEpochEvent {
     const MODULE_NAME: &'static str = "LibraConfig";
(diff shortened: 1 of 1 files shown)


sblackshear:
I had the same question... do you know @runtian-zhou? 

My hypothesis is that consensus subscribes based on the event key, which wouldn't be affected by this bug. But anyone trying to read `NewEpochEvent`'s via [json-rpc](https://github.com/libra/libra/blob/master/json-rpc/types/src/views.rs#L272) definitely would be...

runtian-zhou:
> My hypothesis is that consensus subscribes based on the event key

That is indeed correct. The reason why the struct name isn't causing an error is because it is an event. The struct name is only useful when it is a resource where we need the name to derive the accesspath.

PR #5006, opened 2020-07-10
[storage] set default prune window

msmouse:
## Motivation

On test net our full nodes exhausted disk in roughly a week's time.
Setting a default prune window to prevent that. (40Mil is >4 days
worth of history assuming we have 100 tps on avg.)

Yes
## Test Plan

build

### DIFF
--- a/config/src/config/storage_config.rs
+++ b/config/src/config/storage_config.rs
@@ -26,7 +26,8 @@ impl Default for StorageConfig {
             backup_service_port: 7777,
             dir: PathBuf::from("libradb/db"),
             grpc_max_receive_len: Some(100_000_000),
-            prune_window: None,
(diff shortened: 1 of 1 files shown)


config/src/config/storage_config.rs
```diff
@@ -26,7 +26,8 @@ impl Default for StorageConfig {
             backup_service_port: 7777,
             dir: PathBuf::from("libradb/db"),
             grpc_max_receive_len: Some(100_000_000),
-            prune_window: None,
+            // At 100 tps on avg, we keep 4~5 days of history.
```
sausagee, line 29: nit - add a note to indicate it is for fullnode.

davidiw: this will affect everything though, is this the default behavior that we want for all nodes?

PR #5010, opened 2020-07-10
[move framework] Removing DualAttestationLimit from on-chain configs

sblackshear:
All updates of on-chain configs require reconfiguration and are slightly more heavyweight as a result. There's no need to trigger reconfiguration when the dual attestation limit changes, so just make it a field in an ordinary singleton resource.

- Merge `DualAttestationLimit` module into `DualAttestation`
- Putting all logic related to dual attestation into a single place
- Renaming `travel_rule` to `dual_attestation` everywhere
- Always using "limit" instead of threshold
- Some cleanup of unused resources, error codes. The main change is removing the lower bound check on the dual attestation limit, which is neither a requirement nor desirable.


language/stdlib/modules/DualAttestation.move
```diff
@@ -336,11 +353,65 @@ module DualAttestation {
...
+        }
+
+        /// Mirrors `Self::get_cur_microlibra_limit`.
+        define spec_get_cur_microlibra_limit(): u64 {
+            global<Limit>(CoreAddresses::SPEC_LIBRA_ROOT_ADDRESS()).micro_lbr_limit
+        }
+
+        /// After genesis, the Limit is always published.
+        invariant !LibraTimestamp::spec_is_genesis() ==> spec_is_published();
+    }
```
moeziniaf, line 414: are you writing specs inline now too? :)

PR #5011, opened 2020-07-10
[logging] add data_display for logging errors

gregnazario:
data_display now allows for using the display functionality rather than
serialization for logging data.

### DIFF
--- a/common/logger/src/struct_log.rs
+++ b/common/logger/src/struct_log.rs
@@ -8,6 +8,7 @@ use serde_json::Value;
 use std::{
     collections::HashMap,
     env,
+    fmt::Display,
     fs::{File, OpenOptions},
     io,
     io::Write as IoWrite,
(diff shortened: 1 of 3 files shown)


common/logger/src/struct_log.rs
```diff
@@ -112,6 +112,11 @@ impl StructuredLogEntry {
         self
     }
 
+    /// Used for errors and other types that don't serialize well
+    pub fn data_display<D: std::fmt::Display>(self, key: &'static str, value: D) -> Self {
```
andll, line 116: nit: Usually we prefer to import things, rather then use fully qualified names

network/src/transport.rs
```diff
@@ -221,9 +222,9 @@ async fn upgrade_inbound<T: TSocket>(
     // try authenticating via noise handshake
     let (socket, peer_id) = ctxt.noise.upgrade_inbound(socket).await.map_err(|err| {
         // security logging
-        send_struct_log!(
-            security_log(security_events::INVALID_NETWORK_PEER).data("error", format!("{}", err))
-        );
+        send_struct_log!(security_log(security_events::INVALID_NETWORK_PEER)
+            .data_display("error", &err)
+            .field(network_events::NETWORK_ADDRESS, &addr));
```
mimoo, line 227: that's much better thanks!

PR #5012, opened 2020-07-10
[network] added request size limits rpc & direct send

gregnazario:
This puts an upper limit on the request sizes to partially prevent
malicious actors from sending extremely large requests.  This is
enforced both on input and output at the protocol level to ensure that
we also don't send anything over our new limits.

I also thought about putting the incoming request limit at the transport level, so that we don't pass large messages all the way to the protocol level.  Thoughts?

In addition, this could be configurable, but I figured since this is just a high level number, I hardcoded it for now.

### DIFF
--- a/config/src/config/test_data/public_full_node.yaml
+++ b/config/src/config/test_data/public_full_node.yaml
@@ -12,6 +12,7 @@ full_node_networks:
     - discovery_method: "onchain"
       listen_address: "/ip4/0.0.0.0/tcp/6180"
       network_id: "public"
+      max_frame_size: 8388608 # 8 MiB
 
 upstream:
     networks:
--- a/network/src/peer/mod.rs
+++ b/network/src/peer/mod.rs
@@ -79,6 +79,9 @@ pub struct Peer<TSocket> {
     direct_send_notifs_tx: channel::Sender<PeerNotification>,
     /// Flag to indicate if the actor is being shut down.
     state: State,
+    /// The maximum size of an inbound or outbound request frame
+    /// Currently, requests are only a single frame
+    max_frame_size: usize,
 }
 
 impl<TSocket> Peer<TSocket>
@@ -92,6 +95,7 @@ where
         peer_notifs_tx: channel::Sender<PeerNotification>,
         rpc_notifs_tx: channel::Sender<PeerNotification>,
         direct_send_notifs_tx: channel::Sender<PeerNotification>,
+        max_frame_size: usize,
     ) -> Self {
         let Connection {
             metadata: connection_metadata,
@@ -106,6 +110,7 @@ where
             rpc_notifs_tx,
             direct_send_notifs_tx,
             state: State::Connected,
+            max_frame_size,
         }
     }
 
@@ -119,16 +124,24 @@ where
             "Starting Peer actor for peer: {:?}",
             self_peer_id.short_str()
         );
+
         // Split the connection into a ReadHalf and a WriteHalf.
         let (reader, writer) = tokio::io::split(IoCompat::new(self.connection.take().unwrap()));
+        let mut codec_builder = LengthDelimitedCodec::builder();
+        codec_builder
+            .max_frame_length(self.max_frame_size)
+            .length_field_length(4)
+            .big_endian();
         // Convert ReadHalf to Stream of length-delimited messages.
-        let reader = FramedRead::new(reader, LengthDelimitedCodec::new()).fuse();
+
+        let reader = FramedRead::new(reader, codec_builder.new_codec()).fuse();
         // Create a rate-limited stream of inbound messages.
         let mut reader = reader
             .ratelimit(MESSAGE_RATE_LIMIT_WINDOW, MESSAGE_RATE_LIMIT_COUNT)
             .fuse();
+
         // Convert WriteHalf to Sink of length-delimited messages.
-        let writer = FramedWrite::new(writer, LengthDelimitedCodec::new());
+        let writer = FramedWrite::new(writer, codec_builder.new_codec());
         // Start writer "process" as a separate task. We receive two handles to communicate with
         // the task:
         // `write_reqs_tx`: Instruction to send a NetworkMessage on the wire.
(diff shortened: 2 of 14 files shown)


phlip9:
there's actually already an 8 MiB limit for both send/recv on the whole `NetworkMessage` in `peer/mod.rs` but this should be made more explicit (and needs to go in the spec). we should also push the limits up the stack so application modules never send messages that are too large.

```rust
        // Convert WriteHalf to Sink of length-delimited messages.
        let writer = FramedWrite::new(writer, LengthDelimitedCodec::new());
```

inside [`LengthDelimitedCodec`](https://github.com/tokio-rs/tokio/blob/master/tokio-util/src/codec/length_delimited.rs#L655)

```rust
    pub fn new() -> Builder {
        Builder {
            // Default max frame length of 8MB
            max_frame_len: 8 * 1_024 * 1_024,

            // Default byte length of 4
            length_field_len: 4,

            // Default to the header field being at the start of the header.
            length_field_offset: 0,

            length_adjustment: 0,

            // Total number of bytes to skip before reading the payload, if not set,
            // `length_field_len + length_field_offset`
            num_skip: None,

            // Default to reading the length field in network (big) endian.
            length_field_is_big_endian: true,
        }
    }
```

gregnazario:
> there's actually already an 8 MiB limit for both send/recv on the whole `NetworkMessage` in `peer/mod.rs` but this should be made more explicit (and needs to go in the spec). we should also push the limits up the stack so application modules never send messages that are too large.
> 
> ```rust
>         // Convert WriteHalf to Sink of length-delimited messages.
>         let writer = FramedWrite::new(writer, LengthDelimitedCodec::new());
> ```
> 
> inside [`LengthDelimitedCodec`](https://github.com/tokio-rs/tokio/blob/master/tokio-util/src/codec/length_delimited.rs#L655)
> 
> ```rust
>     pub fn new() -> Builder {
>         Builder {
>             // Default max frame length of 8MB
>             max_frame_len: 8 * 1_024 * 1_024,
> 
>             // Default byte length of 4
>             length_field_len: 4,
> 
>             // Default to the header field being at the start of the header.
>             length_field_offset: 0,
> 
>             length_adjustment: 0,
> 
>             // Total number of bytes to skip before reading the payload, if not set,
>             // `length_field_len + length_field_offset`
>             num_skip: None,
> 
>             // Default to reading the length field in network (big) endian.
>             length_field_is_big_endian: true,
>         }
>     }
> ```

Yeah, I found this and was discussing this outside of here.

First step here is making that configurable just in general as requested.  But, it's good to know we already have an 8MB maximum message size until we do chunking to make our messages larger than a single frame.

phlip9:
> First step here is making that configurable just in general as requested. But, it's good to know we already have an 8MB maximum message size until we do chunking to make our messages larger than a single frame.

i don't quite understand your comment : p 

so just to be safe, this is the max size per `NetworkMessage`. so for example, a single consensus Proposal can never be larger than 8 MiB minus the overhead for enum variant, protocol id, and priority

gregnazario:
> > First step here is making that configurable just in general as requested. But, it's good to know we already have an 8MB maximum message size until we do chunking to make our messages larger than a single frame.
> 
> i don't quite understand your comment : p
> 
> so just to be safe, this is the max size per `NetworkMessage`. so for example, a single consensus Proposal can never be larger than 8 MiB minus the overhead for enum variant, protocol id, and priority

tl;dr There was a lot of conversation off this thread, and everything in this PR is being thrown away.

* The MVP is to make that frame limit configurable.
* After that is to move closer to the application layer about rejecting too large amounts of data before it even gets into the deeper networking layers.

As I've learned through this process, the limits I put in were on the wrong layer, and I'll move them around some.

phlip9:
yeah i agree this is probably not the right layer. in my mind, all application modules need hardening to explicitly generate messages under a given limit. for example, if mempool is pushing out a tx batch it needs to ensure its tx batch is < 8 MiB (or w/e). mempool currently has a maximum `shared_mempool_batch_size = 100`, but you might have a two 5 MiB txs that push us over the limit, so maybe we just need a maximum `tx_size <= ~81 KiB`. of course, what if we have big emergency writeset tx...

network/src/peer/mod.rs
```diff
@@ -119,16 +124,23 @@ where
...
         // Create a rate-limited stream of inbound messages.
         let mut reader = reader
             .ratelimit(MESSAGE_RATE_LIMIT_WINDOW, MESSAGE_RATE_LIMIT_COUNT)
             .fuse();
+
         // Convert WriteHalf to Sink of length-delimited messages.
-        let writer = FramedWrite::new(writer, LengthDelimitedCodec::new());
+        let mut write_codec = LengthDelimitedCodec::new();
+        write_codec.set_max_frame_length(self.max_frame_size);
+        let writer = FramedWrite::new(writer, write_codec);
```
phlip9, line 143: let's be explicit about all the relevant fields

```rust
let codec_builder = LengthDelimitedCodec::builder()
    .big_endian() // length prefix is big-endian encoded
    .length_field_length(4) // u32 length prefix
    .max_frame_length(8 * 1024 * 1024) // 8 MiB maximum frame length
    .new();

let reader = codec_builder.new_read(reader).fuse();
let writer = codec_builder.new_write(writer);
```

config/src/config/test_data/public_full_node.yaml
```diff
@@ -12,6 +12,7 @@ full_node_networks:
     - discovery_method: "onchain"
       listen_address: "/ip4/0.0.0.0/tcp/6180"
       network_id: "public"
+      max_frame_size: 8388608 # 8 MiB
```
mimoo, line 15: why use MiB and not MB btw :P ?

gregnazario: The original limit in the code is 8 MiB, so I didn't want to change it.

phlip9: 8 MiB == 8388608 bytes
8 MB == 8000000 bytes

: p

mimoo: yeah, that's why 8 MB is better imo :P

PR #5013, opened 2020-07-10
[libra-trace] Add counters for libra-trace and struct_log.

changyonghu:
Add counters for libra-trace and struct_log.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2658,6 +2658,7 @@ dependencies = [
  "libra-workspace-hack 0.1.0",
  "log 0.4.8 (registry+https://github.com/rust-lang/crates.io-index)",
  "once_cell 1.4.0 (registry+https://github.com/rust-lang/crates.io-index)",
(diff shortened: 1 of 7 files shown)


metajack:
Why didn't this use libra-metrics? I tried to isolate the prometheus dependency there.

andll:
@davidiw what is the solution you propose? How do we add counter that works for both other libra crates _and_ TCB?

@metajack it seems libra-metrics references prometheus crate, so using it is not going to help?

davidiw:
Well this won't even be used by TCB, you create a counter that is a no-op in TCB land.

I suspect you probably want some sort of ability to register a counter to be used by logging and we'd need a common interface for that. That way the TCB can push the metric and prometheus + hyper can share it.

andll:
We probably need to figure out this 'metrics done differently' in TCB.

There probably will be other components in libra shared between TCB and non-TCB. Currently if they would want to export counters, what would they do?

davidiw:
sure, but this seems like a pretty special metric, do we imagine more cases where we'll want to log at a common layer?

PR #5015, opened 2020-07-10
[json-rpc] fix transaction hash string: should be full length

xli:
## Motivation

4 bytes version was returned as transaction hash, change it to be full hash hex value.

Yes

## Test Plan

Unit test assertion was updated and added string length assertion.

## Related PRs

### DIFF
--- a/json-rpc/src/tests/unit_tests.rs
+++ b/json-rpc/src/tests/unit_tests.rs
@@ -378,7 +378,7 @@ fn test_get_transactions() {
             let version = base_version + i as u64;
             assert_eq!(view.version, version);
             let (tx, status) = &mock_db.all_txns[version as usize];
-            assert_eq!(view.hash, tx.hash().to_string());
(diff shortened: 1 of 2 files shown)


json-rpc/src/tests/unit_tests.rs
```diff
@@ -378,7 +378,8 @@ fn test_get_transactions() {
             let version = base_version + i as u64;
             assert_eq!(view.version, version);
             let (tx, status) = &mock_db.all_txns[version as usize];
-            assert_eq!(view.hash, tx.hash().to_string());
+            assert_eq!(view.hash.len(), 64);
```
thefallentree, line 381: I think this is unnecessary now, due to the second part.  

we should really get rid of to_string() api on that hash() , at least rename it to to_short_string() or something. otherwise there might be a lot of places were misused.

xli: deleted hash len assertion

PR #5016, opened 2020-07-10
[vm] Reworked VMStatus

tnowacki:
- Switched to enum
- Move abort is now the only one that can carry a "sub status"
- Move aborts can cary a location

- TODOs with setting the location in e2e tests 

## Motivation

- Remove sub status from anything other than move abort

## Test Plan

- ran em

### DIFF
--- a/language/move-core/types/src/vm_status.rs
+++ b/language/move-core/types/src/vm_status.rs
@@ -3,7 +3,7 @@
 
 #![allow(clippy::unit_arg)]
 
-use crate::account_address::AccountAddress;
+use crate::language_storage::ModuleId;
 use anyhow::Result;
 #[cfg(any(test, feature = "fuzzing"))]
 use proptest::prelude::*;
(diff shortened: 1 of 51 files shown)


language/move-core/types/src/vm_status.rs
```diff
@@ -42,22 +42,36 @@ pub static EXECUTION_STATUS_MIN_CODE: u64 = 4000;
...
+    /// The VM status corresponding to an EXECUTED status code
+    Executed,
 
-    /// The optional sub status. Used e.g. for things such as the abort code, or the arithmetic
-    /// error type for an ARITHMETIC_ERROR major status.
-    pub sub_status: Option<u64>,
+    /// Indicates an error from the VM, e.g. OUT_OF_GAS, INVALID_AUTH_KEY, RET_TYPE_MISMATCH_ERROR
+    /// etc.
+    /// The code will neither EXECUTED nor ABORTED
+    Error(StatusCode),
```
dariorussi, line 59: so changing `StatusCode` to a more generic one we'll think about, correct?
To be honest I'd be even fine with `Error` and nothing else, like `Executed`

language/move-core/types/src/vm_status.rs
```diff
@@ -152,33 +188,28 @@ pub const EASSERT_ERROR: u64 = 42; // catch-all error code for assert failures
...
-            sub_status => {
-                return VMStatus::new(err.major_status, sub_status, err.message);
-            }
-        };
-        VMStatus::new(new_major_status, None, message)
-    } else {
-        err
+pub fn convert_prologue_runtime_error(status: VMStatus) -> VMStatus {
+    match status {
+        // TODO look at location? or remove this function entirely
```
dariorussi, line 193: I don't know, this looks reasonable to me. If those are the only prologue errors that is great.
It also seems they don't need location or anything, just the code should be enough.
If we are not there yet, you'd hope we'll get there at some point, no?

language/move-core/types/src/vm_status.rs
tzakian: Sadly collapsing to just `Error` won't work; the status code is used to determine whether the transaction should be included in the transaction accumulator as well.

tnowacki: Could we maybe condense the amount of codes though? Basically: All Status Codes relating to "libra errors" + a few (<10) codes for "VM errors"

language/move-core/types/src/vm_status.rs
tnowacki: My thought is you don't need this function at all. You have the location + the abort code, thats MORE granular and trustable than this current implementation. It feels like an un needed layer of complexity now that we have more information

language/move-core/types/src/vm_status.rs
dariorussi: yeah I'd like that, I'd like the strictly needed for the chain to function

language/move-core/types/src/vm_status.rs
dariorussi: but I think those status code are for the client as validation errors.
I guess you are saying the the client can filter properly but it seems to me this is the client.
maybe we'll briefly chat at some point but I don't think there is much to change for now

PR #5018, opened 2020-07-11
Include txn_hash in sign-transaction-using-ed25519 output

ankushagarwal:
## Summary

This `txn_hash` can be used to verify that this particular transaction was already included in the chain.


client/swiss-knife/src/main.rs
```diff
@@ -248,6 +252,7 @@ fn generate_signed_txn(request: GenerateSignedTxnRequest) -> GenerateSignedTxnRe
         })
         .unwrap();
     let signed_txn = SignedTransaction::new(raw_txn, public_key, signature);
+    let txn_hash = CryptoHash::hash(&Transaction::UserTransaction(signed_txn.clone())).to_hex();
```
dmitri-perelman, line 255: Just to verify something: is it the hash of a signed transaction, or the hash of a raw transaction that is included in the TransactionView in JsonRPC?

ankushagarwal: Neither. It is the hash of the Transaction object https://github.com/libra/libra/blob/a31a1607729f95dfe8f9d1eb94d61b8f38559af8/testsuite/generate-format/tests/staged/libra.yaml which is included in respinses in JSON RPC.

ankushagarwal: [Here](https://github.com/libra/libra/blob/a31a1607729f95dfe8f9d1eb94d61b8f38559af8/json-rpc/types/src/views.rs#L320) is the field in JSON RPC which this hash corresponds to

ankushagarwal: Also as an FYI, this field shows only a 4-byte prefix of the entire 32-byte hash today. This PR fixes it https://github.com/libra/libra/pull/5015

dmitri-perelman: Oh, nice! makes sense.

PR #5023, opened 2020-07-11
[language][vm] implement recursion depth checks for runtime types and data layouts

vgao1996:
## Summary
This adds a depth check to all vm operations that could create new types or data layouts recursively. An error will be thrown if the maximum depth is reached. This should make it impossible for an attacker to cause a stack-overflow in the Move VM, given that the stack size is reasonably big. 

## Test Plan
`cargo test`

### DIFF
--- a/language/ir-testsuite/tests/recursion/runtime_layout_deeply_nested.mvir
+++ b/language/ir-testsuite/tests/recursion/runtime_layout_deeply_nested.mvir
@@ -1,0 +1,88 @@
+module M {
+    import 0x1.Signer;
+
+    resource Box<T> { x: T }
+    resource Box3<T> { x: Self.Box<Self.Box<T>> }
+    resource Box7<T> { x: Self.Box3<Self.Box3<T>> }
+    resource Box15<T> { x: Self.Box7<Self.Box7<T>> }
+    resource Box31<T> { x: Self.Box15<Self.Box15<T>> }
+    resource Box63<T> { x: Self.Box31<Self.Box31<T>> }
+    resource Box127<T> { x: Self.Box63<Self.Box63<T>> }
+    resource Box255<T> { x: Self.Box127<Self.Box127<T>> }
+
+    public box3<T>(x: T): Self.Box3<T> {
+        return Box3<T> { x: Box<Self.Box<T>> { x: Box<T> { x: move(x) } } };
+    }
+
+    public box7<T>(x: T): Self.Box7<T> {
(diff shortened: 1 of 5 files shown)


vgao1996:
Currently the test for data layout is not effective. This is because the system only allows a maximum of 2,000,000 gas units for a single transaction and the test would need somewhere around 6,500,000 units. I raised the limit locally to 8,000,000 and it worked, but I feel like I should not check this in. @tzakian @dariorussi thoughts?

language/ir-testsuite/tests/recursion/runtime_layout_deeply_nested.mvir
```diff
@@ -0,0 +1,88 @@
...
+
+//! new-transaction
+import {{default}}.M;
+
+main(account: &signer) {
+    M.publish_257(move(account));
+    return;
+}
+// check: OUT_OF_GAS
+// TODO: should be VM_MAX_VALUE_DEPTH_REACHED
```
vgao1996, line 88: Currently the test for deeply nested data layouts is not effective. This is because the system only allows a maximum of 2,000,000 gas units for a single transaction and the test would need about 6,500,000 units. I raised the limit locally to 8,000,000 and it worked, but I feel like I should raise the limit because of this reason. @tzakian @dariorussi thoughts?

language/move-vm/types/src/loaded_data/runtime_types.rs
```diff
@@ -35,41 +37,62 @@ pub enum Type {
...
+                            ty_args.len(),
+                            idx
+                        )),
+                ),
+            },
+            1,
+        )
+    }
+
+    pub fn clone_type(&self) -> PartialVMResult<Type> {
```
dariorussi, line 93: what is this? and why we need it?

language/ir-testsuite/tests/recursion/runtime_layout_deeply_nested.mvir
dariorussi: I think we need to review all of those limits so let's not worry too much about it

language/move-vm/types/src/loaded_data/runtime_types.rs
vgao1996: I got rid of the "Clone" impl for `Type` and implemented a custom version of `clone` using subst. I thought it would be nice to still expose such an interface. But yeah it's unused right now and I can get it removed.

PR #5024, opened 2020-07-11
Safety rules counters

davidiw:
* Improve gauge for secure-counters, checked with @sherry-x that we only need the latest value as this matches pull metrics as well
* Moved safety rules counters into safety rules as this reduces an unnecessary round trip that can also fail ... just to publish metrics to prometheus :/
* updated prometheus template, but I'm not 100% sure it is correct as CT doesn't have push metrics enabled for LSR... though it should be corrected based upon the query that @sherry-x shared with me before

### DIFF
--- a/consensus/safety-rules/src/counters.rs
+++ b/consensus/safety-rules/src/counters.rs
@@ -10,15 +10,29 @@ use std::sync::Arc;
 define_counters![
     "libra_safety_rules",
     (
-        requested_sign_timeout: Counter,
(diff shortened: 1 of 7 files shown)


davidiw:
Well there was no observable performance change, which seems really interesting!!! I'm curious if the same is true on pre-mainnet. Perhaps not, which would likely imply that it is the write path that's causing us perf issues. Oh well, this is better from a reliability standpoint :)

@bors-libra r=ankushagrawal

PR #5029, opened 2020-07-11
[move-prover] Verified AccountFreezing

junkil-park:
- Added/verified the local specs for initialize and create
- Added/verified some global specs.
- TODO: need to decide whether Validator, ValidatorOperator and DesginatedDealer are freezable or not.

## Motivation

To verify AccountFreezing

Yes

## Test Plan

cargo test

### DIFF
--- a/language/stdlib/modules/AccountFreezing.move
+++ b/language/stdlib/modules/AccountFreezing.move
@@ -50,9 +50,19 @@ module AccountFreezing {
             unfreeze_event_handle: Event::new_event_handle(lr_account),
         });
     }
+    spec fun initialize {
+        aborts_if !LibraTimestamp::spec_is_genesis();
+        aborts_if Signer::spec_address_of(lr_account) != CoreAddresses::SPEC_LIBRA_ROOT_ADDRESS();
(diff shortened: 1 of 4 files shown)


language/stdlib/modules/AccountFreezing.move
```diff
@@ -50,10 +50,21 @@ module AccountFreezing {
...
+        ensures exists<FreezeEventsHolder>(Signer::spec_address_of(lr_account));
+    }
 
     public fun create(account: &signer) {
         move_to(account, FreezingBit { is_frozen: false })
     }
+    spec fun create {
+        aborts_if exists<FreezingBit>(Signer::spec_address_of(account));
+        ensures exists<FreezingBit>(Signer::spec_address_of(account));
+        ensures global<FreezingBit>(Signer::spec_address_of(account)).is_frozen == false;
```
tzakian, line 66: Does `ensures !spec_account_is_frozen(Signer::spec_address_of(account))` work here? That would be cleaner if it does.

junkil-park: Good point! It's just done with an additional spec fun `spec_account_is_not_frozen`. `spec_account_is_not_frozen` is not equivalent to `!spec_account_is_frozen`.

PR #5032, opened 2020-07-12
[language] Update computational gas costs to reflect VM updates

tzakian:
This PR updates the computational gas prices for instructions to reflect the real-world costs of executing those instructions. 

# Updates to costing
We now do a post-processing of the writeset before the epilogue to determine the number of account blobs changed by the execution of the transaction. We then charge for these writes to storage at that point. This in particular means that we can run out of gas after "executing" a transaction. However, this does not violate "charge before performing the work" since we haven't realized the write costs yet when we charge. 

# Scaling of gas units
We internally scale gas units in the VM by `gas_units_scaling_factor`. This allows finer-grained costing of instructions while not requiring floating point numbers. 

# Testing
All tests pass. I removed the "failed_tx_uses_more_gas" test in the ir-testsuite since this was failing, and it was testing a now false property; we don't charge more for transactions that fail, we charge as much as it cost up to the point of failure in the transaction.

Updated gas costs:
* *P2P*: 172 gas units
* *Rotate auth key*: 43


dariorussi:
reading your comment I assume this PR is not ready yet to go in, correct? (if it was marked draft would be better)
Can we talk tomorrow about this? I may put something on your calendar

andll:
Can we estimate what is expected p2p transaction cost after this?
It looks like right now its about 160K, which probably needs to be dialed down at least 1000x to be usable, does this PR address this?

dariorussi:
> Can we estimate what is expected p2p transaction cost after this?
> It looks like right now its about 160K units, which probably needs to be dialed down at least 1000x to be usable, does this PR address this?

no idea, but I love the question. @tzakian can we use this PR to lock down all things cost related for v1?

vgao1996:
@tzakian Just FYI it should be pretty straightforward to add a `fn is_global` to vm references if that helps with gas calculation. Let me know if you need help here!

language/move-vm/runtime/src/data_cache.rs
```diff
@@ -135,6 +135,10 @@ impl<'r, 'l, R: RemoteCache> TransactionDataCache<'r, 'l, R> {
         })
     }
 
+    pub(crate) fn mutated_accounts(&self) -> Vec<AccountAddress> {
```
dariorussi, line 138: not sure about this cloning of keys just to find the count but I guess that is fine....

andll:
>P2P: 172 gas units

Looks like we did reduce it 1000x, this is so cool!!

language/move-vm/runtime/src/data_cache.rs
tzakian: Good point. I can just as well take the count of these keys which doesn't require cloning. I'll update.

runtian-zhou:
LibraVM/config side seems good to me!

bob-wilson:
The multiplier looks good. It's suspicious to me that such a nice round number of 1,000 is the right value to scale the gas costs into the desired range, but it does look like the resulting numbers are in the right ballpark at least.

PR #5033, opened 2020-07-12
[consensus] handle safety-rules restart

davidiw:
Before this, a SafetyRules restart would cause consensus to sit around
being useless except for following along until the next epoch, because
it was uninitialized.

Thanks @zekun000 for making MetricsSafetyRules

Addresses unconfigured safety rules in the middle of an epoch

### DIFF
--- a/consensus/src/round_manager_test.rs
+++ b/consensus/src/round_manager_test.rs
@@ -13,7 +13,7 @@ use crate::{
     network::{IncomingBlockRetrievalRequest, NetworkSender},
     network_interface::{ConsensusMsg, ConsensusNetworkEvents, ConsensusNetworkSender},
     network_tests::{NetworkPlayground, TwinId},
-    persistent_liveness_storage::{PersistentLivenessStorage, RecoveryData},
+    persistent_liveness_storage::RecoveryData,
     round_manager::RoundManager,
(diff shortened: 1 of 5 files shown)


consensus/src/round_manager_test.rs
```diff
@@ -835,3 +835,55 @@ fn sync_on_partial_newer_sync_info() {
...
+            let round = vote_msg.vote().vote_data().proposed().round();
+            node.round_manager
+                .process_local_timeout(round)
+                .await
+                .unwrap();
+            let vote_msg = node.next_vote().await;
+
+            // sign proposal
+            reset_safety_rules(&mut node);
+            node.round_manager.process_vote_msg(vote_msg).await.unwrap();
```
zekun000, line 886: does process_vote_msg invoke safety rules?

davidiw: since there's only a single node, process_vote -> qc accumulation -> proposal generation -> sign_proposal

zekun000: can we unwrap next_proposal() to make it more explicit?

davidiw: that just reads it from the channel, but happy to do so :)

PR #5036, opened 2020-07-13
[move framework][move-prover] Changed spec_has_*_role to take addresses.

DavidLDill:
Changed Roles::spec_has_*_role and Roles::spec_has_*_privilege functions
to consistently take addresses instead of signers as arguments.  This
is more convenient for some specifications.

## Motivation

(Write your answer here.)

## Test Plan

## Related PRs


sblackshear:
> I also think that such spec functions taking `addr` as an argument would be very convenient because global properties often quantify over addresses rather than accounts. However, one little concern is that `Roles::spec_has_role` becomes inconsistent to `Roles::has_role` which takes an `account` as an argument. How about defining additional spec functions `Roles::spec_addr_has_role` or `Roles::spec_has_role_addr`?

Changing `Roles::has_role` to accept an address also makes sense to me.

DavidLDill:
I changed spec_has...role to spec_has...role_addr everywhere.

I haven't needed a Move version.  It's good to keep the Move has...role functions checking signers, because the main (only?) purpose is to check authorization of the signer to perform some action.

PR #5047, opened 2020-07-13
Transaction metadata typing

n4ss:
## Motivation

This PR is implementing transaction metadata types to allow easy parsing and introspection into the metadata, whether the transaction is using regular subaddressing or is subject to travel rule.

Metadata types are added as part of `libra-types::transaction`, they are also part of the auto-generated yaml file that defines libra types so that they can be accessible through different languages.

Yes

## Test Plan

```
$> cargo x test -p generate-format

     Running target/debug/deps/generate_format-cf5c23909f982dc5

running 0 tests

test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out

     Running target/debug/deps/detect_format_change-fb743365d0b74934

running 2 tests
test test_we_can_detect_changes_in_yaml ... ok
test analyze_serde_formats ... ok

test result: ok. 2 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out

     Running target/debug/deps/linter-81889fefa012c8ee

running 1 test
test test_lint_lcs_format ... ok

test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out

   Doc-tests generate-format

running 0 tests

test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out
```

cc @sblackshear  @matbd @LuZhang-Lou


types/src/transaction/mod.rs
```diff
@@ -36,11 +36,13 @@ use std::{
 pub mod authenticator;
 mod change_set;
 pub mod helpers;
+mod metadata;
 mod module;
 mod script;
 mod transaction_argument;
 
 pub use change_set::ChangeSet;
+pub use metadata::*;
```
matbd, line 45: Since we are adding a layer of specifications that is relatively independent, I would make the `metadata` module visible as such with `pub mod metadata` instead `pub use metadata::*`. It should also be cleaner in `cargo doc` (see previous comment)

types/src/transaction/metadata.yaml
```diff
@@ -0,0 +1,50 @@
+---
```
matbd, line 1: We don't need this file any more, do we?

If you want, we can also generate such a file next to `libra.yaml`. The problem here is that consistency between `metadata.rs` and `metadata.yaml` is not guaranteed by CI.

types/src/transaction/metadata.rs
```diff
@@ -0,0 +1,49 @@
+// Copyright (c) The Libra Core Contributors
+// SPDX-License-Identifier: Apache-2.0
+
+use serde::{Deserialize, Serialize};
```
matbd, line 4: Maybe add some module-level explanations here with `//! something ...` and use doc comments `///` below for each struct/enum.

You can check the result with `cargo doc -p libra-types --open`

testsuite/generate-format/src/libra.rs
```diff
@@ -55,6 +55,10 @@ pub fn get_registry() -> Result<Registry> {
     // 2. Trace the main entry point(s) + every enum separately.
     tracer.trace_type::<contract_event::ContractEvent>(&samples)?;
     tracer.trace_type::<language_storage::TypeTag>(&samples)?;
+    tracer.trace_type::<transaction::MetadataType>(&samples)?;
```
matbd, line 58: OK so technically you need to list `GeneralMetadata`, `MetadataType ` and `TravelRuleMetadata` and the rest will follow (in a more future-proof way than the current list).

types/src/transaction/metadata.rs
matbd: Btw. are these metadata only used by the Move script `PeerToPeerWithMetadata`? Maybe add cross references from the doc? https://github.com/libra/libra/blob/master/language/stdlib/transaction_scripts/peer_to_peer_with_metadata.move#L5

types/src/transaction/metadata.yaml
n4ss: Woops yep this is actually a legacy, removing that

types/src/transaction/metadata.rs
```diff
@@ -0,0 +1,49 @@
...
+#[derive(Clone, Debug, Eq, PartialEq, Serialize, Deserialize)]
+pub struct GeneralMetadataV0 {
+    // Subaddress to which the funds are being sent
+    #[serde(with = "serde_bytes")]
+    to_subaddress: Option<Vec<u8>>,
+    // Subaddress from which the funds are being sent
+    #[serde(with = "serde_bytes")]
+    from_subaddress: Option<Vec<u8>>,
+    // Event sequence number of referenced payment
+    referenced_event: Option<u64>,
```
LuZhang-Lou, line 29: let's add more description here for `referenced_event`, this is from the quip:
>  * referenced_event: *In the case where funds must be returned, referenced_event refers to the event sequence number of the sender’s original sent payment event.  Since refunds are just another form of p2p transfer, the referenced event field allows a refunded payment to refer back to the original payment.

types/src/transaction/metadata.rs
```diff
@@ -1,8 +1,14 @@
 // Copyright (c) The Libra Core Contributors
 // SPDX-License-Identifier: Apache-2.0
 
+//! This file implements Libra transaction metadata types to allow
```
matbd, line 4: @sblackshear @n4ss Is the concept of "metadata" specific to the peer-to-peer script, or meant to be shared between several Move scripts?

In the former case, we may want to choose names that leave space for the next move script with a different kind of metadata, say concretely, replacing "Metadata" with wither "TransferMetadata" or "PeerToPeerMetadata".

language/stdlib/transaction_scripts/peer_to_peer_with_metadata.move
```diff
@@ -5,6 +5,7 @@ use 0x1::LibraAccount;
 /// associated metadata `metadata` and (optional) `signature` on the metadata, amount, and
 /// sender address. The `metadata` and `signature` parameters are only required if
 /// `amount` >= 1_000_000 micro LBR and the sender and recipient of the funds are two distinct VASPs.
+/// Standardized `metadata` LCS format can be found in `libra_types::transaction::metadata::MetadataType`.
```
matbd, line 8: @n4ss Thanks! Ideally, now, you would run `(cd language/stdlib && cargo run --release)` to upgrade the doc in generated files. (This is going to take a while to compile, sorry :-|)

n4ss: Doing that now!

types/src/transaction/metadata.rs
n4ss: Afaik, metadata might be relevant for more use cases but only PeerToPeer for now

types/src/transaction/metadata.rs
```diff
@@ -25,23 +34,30 @@ pub struct GeneralMetadataV0 {
...
+/// subject to travel rule
 #[derive(Clone, Debug, Eq, PartialEq, Serialize, Deserialize)]
 pub enum TravelRuleMetadata {
     TravelRuleMetadataVersion0(TravelRuleMetadataV0),
 }
 
+/// Transaction metadata format for transactions subject to travel rule
 #[derive(Clone, Debug, Eq, PartialEq, Serialize, Deserialize)]
 pub struct TravelRuleMetadataV0 {
     // Off-chain reference_id.  Used when off-chain APIs are used.
```
matbd, line 55: I suspect you need `///` here too for this to show up in docs.

types/src/transaction/metadata.rs
```diff
@@ -0,0 +1,65 @@
...
+
+use serde::{Deserialize, Serialize};
+
+/// List of all supported metadata types
+#[derive(Clone, Debug, Eq, PartialEq, Serialize, Deserialize)]
+pub enum MetadataType {
+    Undefined,
+    GeneralMetadataType(GeneralMetadata),
+    TravelRuleMetadataType(TravelRuleMetadata),
+    UnstructuredStringMetadataType(UnstructuredStringMetadata),
```
thefallentree, line 17: Can we just have StringMetadataType(String) and BytesMetadataType(Vec[u8])  cause this is actually two different stuff.

n4ss: I'll just put UnstructuredBytesMetadataType, since this doesn't conform to subaddressing or travel rule, it's very likely that the sender (and hopefully the receiver) know how to interpret those bytes in a custom manner - so no need to come up with a string format until we have a need for it (Memos should probably be explicited at the structure level for example).

types/src/transaction/metadata.rs
n4ss: Fixed

types/src/transaction/metadata.rs
```diff
@@ -0,0 +1,66 @@
...
+//! This file implements Libra transaction metadata types to allow
+//! easy parsing and introspection into metadata, whether the transaction
+//! is using regular subaddressing, is subject to travel rule or corresponds
+//! to an on-chain payment refund.
+
+use serde::{Deserialize, Serialize};
+
+/// List of all supported metadata types
+#[derive(Clone, Debug, Eq, PartialEq, Serialize, Deserialize)]
+pub enum MetadataType {
```
matbd, line 13: I would suggest to remove the `-Type` suffix here (seems like a typo?) and below (re-using names between variants and containers is well supported).

n4ss: Hmm I don't think that's a typo, in general you type opaque data structures with:
1] type
2] struct details

But given that rust supports storing struct details in each variant, I'm fine with it.

n4ss: Actually the ones below are typos

n4ss:
-Fixed the conflicts from the rebase-

PR #5048, opened 2020-07-13
[cluster-test] add lbt compat test suite

rustielin:
## Motivation

Write a test suite for compatibility test (https://github.com/libra/libra/pull/4884). One change is that we launch cluster-test with the old image tag first, and then perform upgrades on it. The way we can do that is specify older tag with `-T` option, and then specify the tag you want to test with `-E TEST_TAG=<tag here>`. The network will start with 100% older tag, then 50% old 50% new, then end 100% new tag, after which we can perform the normal performance benchmark. 

The eventual goal is to be able to start with testnet versions, then perform incremental upgrade to PR version. However, compatibility with testnet version is not likely at this point, and we can just enable it later down the line after 7/15.

Yes

## Test Plan

Canary in CI (https://github.com/libra/libra/pull/5049), and also manual test:

Running manually, to upgrade `land_97c2442a` to `dev_rustielin_pull_5048`, which compares images built off of the previous commit to the commit in this PR.

```
./scripts/cti -T land_97c2442a --cluster-test-tag dev_rustielin_pull_5048 \
    -E RUST_LOG=debug -E BATCH_SIZE=20 -E TEST_TAG=dev_rustielin_pull_5048 \
    --report report.json --suite land_blocking_compat
```

## Related PRs

### DIFF
--- a/testsuite/cluster-test/src/suite.rs
+++ b/testsuite/cluster-test/src/suite.rs
@@ -7,9 +7,9 @@ use std::{cmp::min, env};
 use crate::{
     cluster::Cluster,
     experiments::{
-        CpuFlamegraphParams, Experiment, ExperimentParam, PerformanceBenchmarkParams,
-        PerformanceBenchmarkThreeRegionSimulationParams, RebootRandomValidatorsParams,
-        RecoveryTimeParams, TwinValidatorsParams,
+        CompatiblityTestParams, CpuFlamegraphParams, Experiment, ExperimentParam,
+        PerformanceBenchmarkParams, PerformanceBenchmarkThreeRegionSimulationParams,
+        RebootRandomValidatorsParams, RecoveryTimeParams, TwinValidatorsParams,
     },
 };
 use anyhow::{format_err, Result};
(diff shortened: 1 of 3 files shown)


testsuite/cluster-test/src/suite.rs
```diff
@@ -79,11 +79,27 @@ impl ExperimentSuite {
         Self { experiments }
     }
 
+    fn new_land_blocking_compat_suite(cluster: &Cluster) -> Self {
+        let count: usize = env::var("BATCH_SIZE").unwrap().parse().unwrap();
```
andll, line 83: 1) You probably can use some default value if BATCH_SIZE is not set instead of failing - I would suggest `cluster.validator_instances().len() / 2`.

2) `.unwrap` is usually undesired, because it is hard to debug(with few exceptions). Best practice here is probably changing this fn to return `Result<Self>` and returning error(via `bail!("Clear error message")`, for example) if parsing fails.

testsuite/cluster-test/src/suite.rs
```diff
@@ -79,11 +79,27 @@ impl ExperimentSuite {
...
+        let updated_image_tag = env::var("TEST_TAG").unwrap();
```
andll, line 84: Can we use some better then TEST_TAG name? UPDATE_TO_TAG?

testsuite/cluster-test/src/suite.rs
```diff
@@ -7,9 +7,9 @@ use std::{cmp::min, env};
 use crate::{
     cluster::Cluster,
     experiments::{
-        CpuFlamegraphParams, Experiment, ExperimentParam, PerformanceBenchmarkParams,
-        PerformanceBenchmarkThreeRegionSimulationParams, RebootRandomValidatorsParams,
-        RecoveryTimeParams, TwinValidatorsParams,
+        ComaptiblityTestParams, CpuFlamegraphParams, Experiment, ExperimentParam,
```
andll, line 10: Looks like there is type in `ComaptiblityTestParams` :(

rustielin: LOL nice catch, I'll fix that 😅

testsuite/cluster-test/src/suite.rs
```diff
@@ -79,11 +79,31 @@ impl ExperimentSuite {
         Self { experiments }
     }
 
+    fn new_land_blocking_compat_suite(cluster: &Cluster) -> Result<Self> {
+        let count: usize = match env::var("BATCH_SIZE") {
+            Ok(val) => val.parse().expect(&format!("Invalid BATCH_SIZE {}", val)),
```
andll, line 84: So .expect here will just panic, since you returning Result, you can just return proper error instead:

`val.parse().map_err(|e| format_err!("Failed to parse BATCH_SIZE {}: {}", val, e)?`

testsuite/cluster-test/src/suite.rs
```diff
@@ -79,11 +79,31 @@ impl ExperimentSuite {
...
+            Err(_) => cluster.validator_instances().len() / 2,
+        };
+        let updated_image_tag =
+            env::var("UPDATE_TO_TAG").expect("Expected environment variable UPDATE_TO_TAG");
```
andll, line 88: Similarly:
`env::var("UPDATE_TO_TAG").map_err(|_| format_err!("Expected environment variable UPDATE_TO_TAG")?`

testsuite/cluster-test/src/suite.rs
rustielin: noted!

PR #5050, opened 2020-07-13
[move framework] cleanup of account limits and balance holding

sblackshear:
- Restricting the set of entities that can hold balances by defining a new Roles function + using it in add_currency
- Combine adding currency to ParentVASP and publishing account limits to ParentVASP. This simplifies the client-facing API by eliminating the allow_currency transaction script
- Split VASP::try_accept_currency into separate functions for checking whether limits exists + publishing them, which should simplify verification of the add_currency script


language/stdlib/modules/Roles.move
```diff
@@ -259,6 +259,23 @@ module Roles {
          has_libra_root_role(account)
     }
 
+    /// Return true if `addr` is allowed to receive and send `Libra<T>` for any T
+    public fun can_hold_balance(account: &signer): bool acquires RoleId {
```
sblackshear, line 263: @junkil-park: This is to address the issue with the "balances" column in the spreadsheet that you pointed out.

junkil-park: LGTM!

PR #5051, opened 2020-07-13
[libra_vm] Get rid of PublishModule capability

runtian-zhou:
## Motivation

Remove the `PublishModule` capability to remove the hard coded pattern to determine if a module is publishable.  The capability functionality is moved to `VMPublishingOption` instead.

Yes


types/src/on_chain_config/vm_config.rs
```diff
@@ -14,27 +17,67 @@ use serde::{Deserialize, Serialize};
...
 /// publishing are mutually exclusive options.
 #[derive(Clone, Debug, Deserialize, PartialEq, Serialize)]
-pub enum VMPublishingOption {
+pub struct VMPublishingOption {
+    pub script_option: ScriptPublishingOption,
+    pub module_option: ModulePublishingOption,
+}
+
+#[derive(Clone, Debug, Deserialize, PartialEq, Serialize)]
+pub enum ScriptPublishingOption {
```
tzakian, line 26: Do we want a `LimitedSender` option here similar to module publishing options? This way certain senders could send custom transactions (e.g., TC)

runtian-zhou: So `Locked` will contain a `whitelist` for script hashes and a whitelist for senders right?

tzakian: That's my thinking. Either that or if anyone who can publish a module can also send arbitrary transactions. What are your thoughts @dariorussi @sblackshear?

dariorussi: not sure if we do custom transactions via write set.
Don't know it's an interesting question but I'd love to close on it for now.
Thoughts?

PR #5059, opened 2020-07-13
[validator-set] Allow validators to update their individual configs

valerini:
This PR allows operators of individual validators to update their configs in the validator set directly and trigger the reconfiguration.
A new function got added to LibraSystem module:
`update_config_and_reconfigure(operator_account: &signer, validator_address: address)`
A new transaction script `set_validator_config_and_reconfigure.move` sets the config locally and updates it in the set triggering the reconfiguration and emitting a NewEpochEvent.
The bulk-updates functionality is removed (i.e. the function `LibraSystem::update_and_reconfigure()` with no parameters), and the transaction script `reconfigure.move` is removed as well.

Next PR will introduce the rate limits to the call `update_config_and_reconfigure`.

## Motivation

Bulk updates were optimizing the system by inducing less frequent ad-hoc updates to the consensus keys, but rather encouraging regularly scheduled key rotations. This optimization though poses a risk, first because a lot of keys might be updating all at once and second because the admin of a operating node might not be online to watch the node successfully do the key rotation as there can be a significant time gap between scheduling the rotation and actually executing it. So this PR removes bulk updates and reintroduces individual config rotations.

Yes

## Test Plan

cargo xtest


language/stdlib/modules/LibraConfig.move
```diff
@@ -73,6 +73,33 @@ module LibraConfig {
         reconfigure_();
     }
 
+    // Set a config item to a new value and trigger a reconfiguration.
+    public fun set_with_capability<Config: copyable>(
```
sblackshear, line 77: We might want this name to shout that it triggers reconfiguration to avoid surprises (`set_and_reconfigure`?)

language/stdlib/modules/LibraConfig.move
```diff
@@ -73,6 +73,33 @@ module LibraConfig {
...
+        config_account: &signer,
+        payload: Config,
+    ): ModifyConfigCapability<Config> {
+        assert(Roles::has_on_chain_config_privilege(config_account), ENO_CONFIG_PRIVILEGE);
+        assert(Signer::address_of(config_account) == CoreAddresses::LIBRA_ROOT_ADDRESS(), EINVALID_SINGLETON_ADDRESS);
+        move_to(config_account, LibraConfig { payload });
+        // We don't trigger reconfiguration here, instead we'll wait for all validators update the binary
+        // to register this config into ON_CHAIN_CONFIG_REGISTRY then send another transaction to change
+        // the value which triggers the reconfiguration.
+        return ModifyConfigCapability<Config> {}
```
sblackshear, line 100: Nit: `return` not needed.

language/stdlib/modules/LibraConfig.move
```diff
@@ -73,6 +73,33 @@ module LibraConfig {
...
+    // Publish a new config item. The caller will use the returned ModifyConfigCapability to specify the access control
+    // policy for who can modify the config.
+    public fun publish_new_config_with_capability<Config: copyable>(
+        config_account: &signer,
+        payload: Config,
+    ): ModifyConfigCapability<Config> {
+        assert(Roles::has_on_chain_config_privilege(config_account), ENO_CONFIG_PRIVILEGE);
+        assert(Signer::address_of(config_account) == CoreAddresses::LIBRA_ROOT_ADDRESS(), EINVALID_SINGLETON_ADDRESS);
+        move_to(config_account, LibraConfig { payload });
+        // We don't trigger reconfiguration here, instead we'll wait for all validators update the binary
```
sblackshear, line 97: This  comment seems out of place given that not all `LibraConfig`s are validator-related.

language/stdlib/modules/LibraConfig.move
```diff
@@ -73,6 +73,33 @@ module LibraConfig {
...
+        reconfigure_();
+    }
+
+    // Publish a new config item. The caller will use the returned ModifyConfigCapability to specify the access control
+    // policy for who can modify the config.
+    public fun publish_new_config_with_capability<Config: copyable>(
+        config_account: &signer,
+        payload: Config,
+    ): ModifyConfigCapability<Config> {
+        assert(Roles::has_on_chain_config_privilege(config_account), ENO_CONFIG_PRIVILEGE);
```
sblackshear, line 94: Add `        assert(LibraTimestamp::is_genesis(), ENOT_GENESIS);`?

language/e2e-tests/src/common_transactions.rs
```diff
@@ -260,6 +246,38 @@ pub fn set_validator_config_txn(
     )
 }
 
+/// Returns a transaction to set config for a candidate validator
+pub fn set_validator_config_and_reconfigure_txn(
```
zekun000, line 250: is the set_validator_config txn only used in genesis build after this?

language/stdlib/modules/LibraConfig.move
```diff
@@ -73,6 +73,33 @@ module LibraConfig {
...
+        let addr = CoreAddresses::LIBRA_ROOT_ADDRESS();
+        assert(exists<LibraConfig<Config>>(addr), ECONFIG_DOES_NOT_EXIST);
+        let config = borrow_global_mut<LibraConfig<Config>>(addr);
+        config.payload = payload;
+        reconfigure_();
+    }
+
+    // Publish a new config item. The caller will use the returned ModifyConfigCapability to specify the access control
+    // policy for who can modify the config.
+    public fun publish_new_config_with_capability<Config: copyable>(
```
zekun000, line 90: the name is confusing, it's not publish with capability, it's publish and return the capability, this probably can be consolidated with `publish_new_config`

language/e2e-tests/src/common_transactions.rs
davidiw: should we be using the transaction builder? (see below)

language/transaction-builder/src/generated.rs
```diff
@@ -668,6 +653,39 @@ pub fn encode_set_validator_config_script(
     )
 }
 
+/// Set validator's config and updates the config in the validator set. NewEpochEvent is
+/// emitted.
+pub fn encode_set_validator_config_and_reconfigure_script(
```
davidiw, line 658: oh yeah, let's use this in the e2e test above

language/stdlib/modules/LibraConfig.move
valerini: There is another `set` function that also triggers the reconfiguration, and I didn't touch it...
I'll rename this one to `set_with_capability_and_reconfigure` then...

language/e2e-tests/src/common_transactions.rs
valerini: @zekun000 , not just in genesis, but any time we want to add a new validator: the validator will set its config locally first (without reconfig) before being added to the validator set.

language/stdlib/modules/LibraConfig.move
valerini: Actually, question here: is it possible we'll want to publish more configs in the future, i.e. post-genesis?

language/stdlib/modules/LibraConfig.move
valerini: Great point!!!
I'll rename this to `publish_new_config_and_get_capability` and make the `publish_new_config` call it to avoid code duplication.

language/e2e-tests/src/common_transactions.rs
zekun000: ah, can we rename that to something like `register_validator_config` and make it clear it publishes new config instead of changing it.

language/stdlib/modules/LibraConfig.move
sblackshear: I think so, but the way we've be handling situations like this elsewhere is to use the most constrained configuration now (easier for verification) with the understanding that we can relax things later. As long as we don't need to modify a resource layout, it's fairly easy to make changes.

language/e2e-tests/src/common_transactions.rs
valerini: Sure, that's a good idea!

language/stdlib/modules/LibraConfig.move
valerini: I see, I was trying to unify the code of `publish_new_config` and `publish_new_config_with_capability`, do you think for both it makes sense to include the assert that we are calling from genesis? The `publish_new_config` is called for LibraVersion, RegisteredCurrencies and LibraVM, and `publish_new_config_with_capability` only called for LibraSystem.

language/e2e-tests/src/common_transactions.rs
valerini: This PR becomes a little crowded, I'll put a TODO and rename in the next PR as all the accompanying rust functions will also have to be renamed as well...

language/stdlib/modules/LibraConfig.move
```diff
@@ -72,19 +72,41 @@ module LibraConfig {
...
+        let config = borrow_global_mut<LibraConfig<Config>>(addr);
+        config.payload = payload;
+        reconfigure_();
+    }
+
+    // Publish a new config item.
+    // The caller will use the returned ModifyConfigCapability to specify the access control
+    // policy for who can modify the config.
+    // Does not trigger a reconfiguration.
+    public fun publish_new_config_and_get_capability<Config: copyable>(
```
zekun000, line 91: sorry I probably miss something, looking at this again, the original code should work as is? what's the motivation of creating these two methods?

the config account holds the `ModifyConfigCapability` and is able to call `set::<LibraSystem>` directly, no?

zekun000: ah nvm I see, it enables the operator account to borrow the capability

valerini:
Second attempt, resolving the merge conflict :) @bors-libra r=sblackshear

PR #5064, opened 2020-07-14
introduce proc macro - inject_error

zekun000:
## Motivation

This commit introduces a proc macro inject-error which can help test the robustness of the system.

The feature is gated by a feature flag and not enabled in cargo test i.e. #[cfg(all(not(test), feature="inject-error"))]

Y

## Test Plan

test binary

## Related PRs

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -11,6 +11,7 @@ members = [
     "common/datatest-stable",
     "common/debug-interface",
     "common/futures-semaphore",
+    "common/inject-error",
     "common/lcs",
     "common/libradoc",
     "common/logger",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2055,6 +2055,17 @@ dependencies = [
 ]
 
 [[package]]
+name = "inject-error"
+version = "0.1.0"
+dependencies = [
+ "anyhow 1.0.31 (registry+https://github.com/rust-lang/crates.io-index)",
+ "libra-workspace-hack 0.1.0",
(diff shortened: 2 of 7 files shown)


andll:
My two worries here are those:
- We are going to generate failure in _any_ unit test, it can contribute to higher degree of flakiness. Especially if some people later won't know about this feature
- We _might_ have a problem of test feature activating in release mode(this have happened in past - do we have any active guard against it @metajack / @sunshowers ?). If this happen that would be pretty sad

zekun000:
@andll for first no, I'm not looking for injecting it in unit tests where it tests expected behavior. it aims for smoke test/cluster test at integration level.
for second, I don't have a good idea where to gate it, what I have in mind is config + env var given smoke test/cluster test is built without features.

andll:
For first - looking at the code it seems it just sets fixed chance of failure when compiled with test feature? Do you mean there will be some additional work to adjust that with config?

Also note that cluster test runs release binaries, since it verifies docker images before they shipped to prod.

zekun000:
@andll yeah, it's just prototype, one way is to gate with an env var. so it can be `if enable && rand < chance` and removed the cfg(test), we probably can further customize the chance with some config value

mimoo:
why not gate it behind a feature flag?

Cargo.toml
```diff
@@ -11,6 +11,7 @@ members = [
     "common/datatest-stable",
     "common/debug-interface",
     "common/futures-semaphore",
+    "common/inject_error",
```
metajack, line 14: Please use `inject-error`.

PR #5065, opened 2020-07-14
[network] added seed peer generation from on chain

gregnazario:
This builds a seed peers config file in YAML, from querying the on chain
validator set.  This allows us to easily get seed peers to provide for
onboarding.

Tested locally against my docker setup:
```
gnazario-mbp:libra gnazario$ ./target/debug/seed-peer-generator generate-seed-peers -o ~/ -p 0000000000000000000000000A550C18 -e http://localhost:8080
Response[status: 200, status_text: OK]
gnazario-mbp:libra gnazario$ cat ~/seed_peers.yaml
---
5942d356f114089d4a46f2f3b0b15b52:
  - /ip4/0.0.0.0/tcp/6180/ln-noise-ik/bc9e1e478585fa19da7e09de7031d1836357969f9807f58c954a4b8c84ed9a77/ln-handshake/0/ln-noise-ik/bc9e1e478585fa19da7e09de7031d1836357969f9807f58c954a4b8c84ed9a77/ln-handshake/0
c5efcfba9f950ba45077bc13668a1a9e:
  - /ip4/0.0.0.0/tcp/6180/ln-noise-ik/e54dc8abcecca702e217a2f04ae4e156e6cdd729d59ab46e86bebe0ec2b5ed03/ln-handshake/0/ln-noise-ik/e54dc8abcecca702e217a2f04ae4e156e6cdd729d59ab46e86bebe0ec2b5ed03/ln-handshake/0
147bbe917424895927838437371fc66e:
  - /ip4/0.0.0.0/tcp/6180/ln-noise-ik/dcff2eacefeafb696665a291e562222fbb0832e48c1e5f2a1a661b26a3beb86e/ln-handshake/0/ln-noise-ik/dcff2eacefeafb696665a291e562222fbb0832e48c1e5f2a1a661b26a3beb86e/ln-handshake/0
5551761fba0e6216e466938067d8acaf:
  - /ip4/0.0.0.0/tcp/6180/ln-noise-ik/0dff7b4b8ca9df32200f6b8c0eaa18b35acb1e7f04085e8dc4cf69464daeb034/ln-handshake/0/ln-noise-ik/0dff7b4b8ca9df32200f6b8c0eaa18b35acb1e7f04085e8dc4cf69464daeb034/ln-handshake/0
8473e0a89638af8c33915a9da303ae9a:
  - /ip4/0.0.0.0/tcp/6180/ln-noise-ik/e50a3febb8a6d94c3acadb99cb6d3870f662930ca63f45370c4238192a5f6e54/ln-handshake/0/ln-noise-ik/e50a3febb8a6d94c3acadb99cb6d3870f662930ca63f45370c4238192a5f6e54/ln-handshake/0
```


config/seed-peer-generator/src/main.rs
```diff
@@ -0,0 +1,81 @@
...
+struct GenerateArgs {
+    #[structopt(short = "o", long, parse(from_os_str))]
+    /// The output directory
+    output_dir: PathBuf,
+    #[structopt(short = "e", long)]
+    /// JSON RPC endpoint
+    endpoint: String,
+    #[structopt(short = "a", long)]
+    /// Specify the AccountAddress to retrieve peers for
+    account_address: AccountAddress,
```
davidiw, line 35: This is safe to assume, no need to request it

config/seed-peer-generator/src/main.rs
```diff
@@ -0,0 +1,81 @@
...
+
+// TODO: Use the definition from network?
+pub type SeedPeersConfig = HashMap<PeerId, Vec<NetworkAddress>>;
+
+#[allow(clippy::large_enum_variant)]
+#[derive(Debug, StructOpt)]
+#[structopt(about = "Tool to generate configs from chain")]
+enum Args {
+    #[structopt(about = "Generate seed peers config")]
+    GenerateSeedPeers(GenerateArgs),
```
davidiw, line 22: Why not inline args below? Are we doing to use this for other things? The only other operation I could imagine is to update an existing config, but I don't know if we need this unflattened layout.

config/seed-peer-generator/src/main.rs
```diff
@@ -0,0 +1,81 @@
...
+        Args::GenerateSeedPeers(args) => generate_seed_peers_config(args)
+    };
+}
+
+fn generate_seed_peers_config(args: GenerateArgs) {
+    let mut seed_peers_config = SeedPeersConfig::new();
+    let validator_set = get_validator_set(args.endpoint, args.peer_id).unwrap();
+
+    // Add the validators to the config!
+    if let Some(validator_set) = validator_set {
```
davidiw, line 51: let validator_set = validator_set.expect("No ValidatorSet found");

config/seed-peer-generator/src/main.rs
```diff
@@ -0,0 +1,81 @@
...
+    seed_peers_config
+        .save_config(args.output_dir.join("seed_peers.yaml"))
+        .expect("Unable to save seed peers config");
+}
+
+/// Retrieve validator set from the JSON-RPC endpoint
+fn get_validator_set(endpoint: String, peer_id: PeerId) -> anyhow::Result<Option<ValidatorSet>> {
+    let json_rpc = JsonRpcClient::new(endpoint);
+    let account_state = json_rpc.get_account_state(peer_id, None)?;
+    Ok(account_state.get_validator_set()?)
```
davidiw, line 70: just for fun, you could also use `.map_err(|e| e.into())`
I don't know which one is better...

config/seed-peer-generator/src/main.rs
```diff
@@ -0,0 +1,81 @@
...
+fn generate_seed_peers_config(args: GenerateArgs) {
+    let mut seed_peers_config = SeedPeersConfig::new();
+    let validator_set = get_validator_set(args.endpoint, args.peer_id).unwrap();
+
+    // Add the validators to the config!
+    if let Some(validator_set) = validator_set {
+        for validator in validator_set.payload().iter() {
+            let (peer_id, seed_addrs) = to_seed_peer(validator);
+            seed_peers_config.insert(peer_id, seed_addrs);
+        }
```
davidiw, line 55: Maybe we should get rid of this SeedPeersConfig and use a type statement instead.. ahh but not here :)

config/seed-peer-generator/src/main.rs
gregnazario: Ah, I just was doing something quick and dirty by copying the pattern made by ConfigGenerator.  I'm sure I can flatten it.

I wonder if we should just put it in the `ConfigGenerator`

config/seed-peer-generator/src/main.rs
gregnazario: I'm not sure what you mean.  SeedPeersConfig is just a type named of a specific map.

config/seed-peer-generator/src/main.rs
```diff
@@ -0,0 +1,64 @@
...
+    let args = Args::from_args();
+    let mut seed_peers_config = SeedPeersConfig::new();
+    let validator_set = get_validator_set(args.endpoint, libra_root_address()).unwrap();
+
+    // Add the validators to the config!
+    let validator_set = validator_set.expect("No validator set found.");
+    for validator in validator_set.payload().iter() {
+        let (peer_id, seed_addrs) = to_seed_peer(validator);
+        seed_peers_config.insert(peer_id, seed_addrs);
+    }
```
davidiw, line 41: oh in that case, we could consider using map...
```
let seed_peers = validator_set.payload().iter().map(|v| to_seed_peer(v)).collect::<SeedPeersConfig>();
```

PR #5066, opened 2020-07-14
[language] adjust VM counters for transaction counts

bob-wilson:
- Add a new counter for system transactions. This should help clarify
  the discrepancy with transaction counts from other parts of the system,
  which are counting all kinds of transactions. Testnet right now has
  very few user transactions.
- Rename the existing TRANSACTIONS_EXECUTED counter to be specific to
  user transactions.
- Rename the TRANSACTIONS_VERIFIED counter to TRANSACTIONS_VALIDATED.
- Drop the discarded transaction graph from the VM dashboard. At least
  for now, this does not seem to be very useful, since transactions are
  rarely discarded during execution. If there is a problem in the system
  that causes many transactions to be discarded, we should be able to
  observe that as a drop in the number of successful transactions.
- For the same reason, change the dashboard graph of validated
  transactions to show only the successful count.

Yes

## Test Plan

Ran existing tests

### DIFF
--- a/language/libra-vm/src/counters.rs
+++ b/language/libra-vm/src/counters.rs
@@ -2,29 +2,38 @@
 // SPDX-License-Identifier: Apache-2.0
 
 use libra_metrics::{
-    register_histogram, register_int_counter_vec, register_int_gauge, Histogram, IntCounterVec,
-    IntGauge,
+    register_histogram, register_int_counter, register_int_counter_vec, register_int_gauge,
+    Histogram, IntCounter, IntCounterVec, IntGauge,
 };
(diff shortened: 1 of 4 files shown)


bob-wilson:
> One clarification, when you say "drop in the number of successful transactions" what does it mean exactly?

I mean that if there is a problem in the system that somehow causes a spike in the number of discarded transactions, there should be a corresponding drop in the USER_TRANSACTIONS_EXECUTED counter that shows the number of successfully executed transactions. Of course that assumes a roughly steady rate of incoming transaction requests, so I'm not sure what will be needed in practice. I'm going to follow up and ask around about that.

PR #5071, opened 2020-07-14
[vm] Filter StatusCodes away from TransactionOutput

tnowacki:
- Changes the information kept in TransactionStatus
- Kept transactions have a new KeptVMStatus. Runtime errors have detailed information, but no error codes. All other errors are bucketed
- Discarded transactions have a StatusCode

## Motivation

- We want to have move aborts visible in committed transactions 
- We want to make sure versioning is easy and we don't have too much VM error information in committed transactions

## TODOs

- Some errors are missing execution location information
- Cleanup `known_locations`
- Fix the double `check` needed for aborts and some other functional test issues 

## Test Plan

- ran em

### DIFF
--- a/language/move-core/types/src/vm_status.rs
+++ b/language/move-core/types/src/vm_status.rs
@@ -60,7 +60,34 @@ pub enum VMStatus {
 
     /// Indicates an `abort` from inside Move code. Contains the location of the abort and the code
     MoveAbort(AbortLocation, /* code */ u64),
+
+    /// Indicates an failure from inside Move code, where the VM could not continue exection, e.g.
+    /// dividing by zero or a missing resource
+    ExecutionFailure {
+        status_code: StatusCode,
+        location: AbortLocation,
+        function: u16,
+        code_offset: u16,
+    },
 }
+
+#[derive(Clone, PartialEq, Eq, PartialOrd, Ord, Hash, Serialize, Deserialize)]
+#[cfg_attr(any(test, feature = "fuzzing"), derive(Arbitrary))]
+#[cfg_attr(any(test, feature = "fuzzing"), proptest(no_params))]
+pub enum KeptVMStatus {
+    Executed,
+    OutOfGas,
+    MoveAbort(AbortLocation, /* code */ u64),
+    ExecutionFailure {
+        location: AbortLocation,
+        function: u16,
+        code_offset: u16,
+    },
+    VerificationError,
+    DeserializationError,
+}
+
+pub type DiscardedVMStatus = StatusCode;
 
 /// An `AbortLocation` specifies where a Move program `abort` occurred, either in a function in
 /// a module, or in a script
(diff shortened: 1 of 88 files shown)


language/move-core/types/src/vm_status.rs
```diff
@@ -60,6 +60,31 @@ pub enum VMStatus {
...
+    Executed,
+    OutOfGas,
+    MoveAbort(AbortLocation, /* code */ u64),
+    ExecutionFailure {
+        location: AbortLocation,
+        function: u16,
+        code_offset: u16,
+    },
+    VerificationError,
+    DeserializationError,
```
sblackshear, line 87: @tzakian/@runtian-zhou: does this only happen when we fail to deserialize the script or its arguments? If so, might be slightly more intuitive to call it something like `MalformedScript` or `ScriptDeserializationError` to indicate that we parsed the "outer" transaction fine (hence why the tx was included/gas was charged), but not the script.

dariorussi: that part is not well defined now and we should make sure it's properly handled.
That applies to verification errors as well. Both a deserialization and a verification errors on published modules would be disastrous, whereas in scripts or a module to publish perfectly expected.
I am not sure about the `DeserializationError` it seems too specific to me, but I am fine waiting the end of the PR and make a call later

tzakian: I think this should only happen when deserializing a user-submitted blob so I think `MalformedScript` wouldn't be bad, but @runtian-zhou  or @vgao1996  would probably know better than me. 

I also agree with Dario that any deserialization of on-chain data that fails is much more serious than a failure to deserialize user-submitted data  -- so should fall in the  invariant violation category.

tnowacki: It's currently a StatusType that is marked as keep https://github.com/libra/libra/blob/15cc989a7ec611a07d860bf9a1e16add40c755a4/types/src/transaction/mod.rs#L582

These errors mostly come from that spot. Do we want new buckets? Are there things we are keeping we should discard?

tnowacki: >  so should fall in the invariant violation category.

Invariant violations are discarded. Deserialization errors are not

tzakian: I think we are keeping things that we should discard, and we should adjust the buckets a bit. In particular:
* I think we should keep deserialization errors that arise when we deserialize any data that is submitted by the user (KEPT).
* I think any errors that arise when deserializing on-chain data (code or values) should be an invariant violation (DISCARDED).

dariorussi: yes that is correct, and I can look into all of this once this is in if it helps

tnowacki: > yes that is correct, and I can look into all of this once this is in if it helps

Sounds like a plan. Let's keep the behavior as is, and switch it after this lands. A ton going on right now. 
I can do it too, I'm just afraid of this PR being too big to handle

tzakian: Agreed. Also once this lands I'll do the plumbing of it in to the transactiontocommit and json-rpc side of things as well

language/move-core/types/src/vm_status.rs
```diff
@@ -113,6 +213,30 @@ impl fmt::Display for VMStatus {
...
+            RecordedStatus::MoveAbort(location, code) => {
+                write!(f, "ABORTED with code {} in {}", code, location)
+            }
+            RecordedStatus::ExecutionFailure {
+                location,
+                function,
+                code_offset,
+            } => write!(
+                f,
+                "EXECUTION_FAILURE at bytecode offset {} in function index {} in {}",
```
dariorussi, line 233: I would even format this in a more compact way, like
`location::idx[offset]`
if that makes sense. It feels that info is too technical and reading it in words just sounds funny.
But just to say, anything works

language/move-core/types/src/vm_status.rs
```diff
@@ -123,96 +247,65 @@ impl fmt::Debug for VMStatus {
...
     }
 }
 
-impl fmt::Debug for AbortLocation {
+impl fmt::Debug for RecordedStatus {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
         match self {
-            AbortLocation::Script => write!(f, "Script"),
-            AbortLocation::Module(id) => write!(f, "{}", id),
+            RecordedStatus::Executed => write!(f, "EXECUTED"),
```
dariorussi, line 269: those Display/Debug all look pretty cool to me

language/move-core/types/src/vm_status.rs
```diff
@@ -536,6 +702,21 @@ impl StatusCode {
 
         StatusType::Unknown
     }
+
+    /// Should only be used in debug asserts. These status codes are missing some data
+    /// checked in debug asserts
+    pub fn should_skip_checks_for_todo(self) -> bool {
```
tnowacki, line 708: @dariorussi I don't know how to set this up properly. But some checks don't have their bytecode offest/`pc` set during execution. No time to figure this out just now, so I need this todo check splattered in front of some debug asserts.

language/move-lang/functional-tests/tests/libra_timestamp/basic.move
```diff
@@ -7,6 +7,8 @@ script {
         LibraTimestamp::initialize(account);
     }
 }
+// TODO(status_migration) remove duplicate check
```
tnowacki, line 10: @vgao1996 Lots of weirdness going on here. Maybe I did something wrong with how I implemented adding the VMStatus to the function tests? (in addition to the transaction output)

language/move-core/types/src/vm_status.rs
dariorussi: it looks all beautiful to me....

language/move-core/types/src/vm_status.rs
tnowacki: I feel it feels weird since its not the name of the function, its the offset.

language/move-core/types/src/vm_status.rs
```diff
@@ -157,14 +134,59 @@ impl VMStatus {
...
+                    // Any unknown error should be discarded
+                    StatusType::Unknown => Err(code),
+                    // Any error that is a validation status (i.e. an error arising from the prologue)
+                    // causes the transaction to not be included.
+                    StatusType::Validation => Err(code),
+                    // If the VM encountered an invalid internal state, we should discard the transaction.
+                    StatusType::InvariantViolation => Err(code),
+                    // A transaction that publishes code that cannot be verified will be charged.
+                    StatusType::Verification => Ok(KeptVMStatus::VerificationError),
+                    // Even if we are unable to decode the transaction, there should be a charge made to
```
sblackshear, line 174:
```suggestion
                    // If we are able to decode the`SignedTransaction`, but failed to decode `SingedTransaction.raw_transaction.payload` (i.e., the transaction script), there should be a charge made to
```

thefallentree:
is it still true the only success status is "executed" ?  all the other status is side-effect free?

aching:
This appears to be a breaking change for clients right?  Is so, we should add the "breaking" label.

tnowacki:
> is it still true the only success status is "executed" ? all the other status is side-effect free?

That's correct. No real change other than the amount of information being recorded

PR #5076, opened 2020-07-14
[move adapter] add verification of chain_id

sblackshear:
This is the Move side of https://github.com/libra/libra/pull/4969.

- add `ChainId` resource published in genesis
- Extend transaction prologue with `chain_id` argument and check it against the value in the `ChainId` resource
- Read `chain_id` from the transaction in the Libra adapter


config/management/src/genesis.rs
```diff
@@ -33,7 +33,12 @@ impl Genesis {
         let genesis = vm_genesis::encode_genesis_transaction_with_validator(
             association_key,
             &validators,
+<<<<<<< HEAD
```
dariorussi, line 36: this is certainly an interesting change... ;)

sblackshear: I have my reasons...

tzakian:
Overall LGTM

config/config-builder/src/validator_config.rs
```diff
@@ -152,6 +152,7 @@ impl ValidatorConfig {
                 .test
                 .as_ref()
                 .and_then(|config| config.publishing_option.clone()),
+            1, // TODO: chain_id constant
```
sblackshear, line 155: @davidiw @JoshLind, what should I use here?

config/management/src/genesis.rs
```diff
@@ -34,6 +34,7 @@ impl Genesis {
             association_key,
             &validators,
             Some(libra_types::on_chain_config::VMPublishingOption::open()),
+            1, // TODO: read chain_id from Genesis struct
```
sblackshear, line 37: Same question here

language/move-core/types/src/vm_status.rs
```diff
@@ -205,6 +206,7 @@ pub fn convert_prologue_runtime_error(status: VMStatus) -> VMStatus {
                 // Can't pay for transaction gas deposit/fee
                 ECANT_PAY_GAS_DEPOSIT => StatusCode::INSUFFICIENT_BALANCE_FOR_TRANSACTION_FEE,
                 ETRANSACTION_EXPIRED => StatusCode::TRANSACTION_EXPIRED,
+                EBAD_CHAIN_ID => StatusCode::BAD_CHAIN_ID,
```
sblackshear, line 209: @tnowacki: Apologies in advance for this if it causes conflicts...

config/management/src/genesis.rs
```diff
@@ -33,7 +33,8 @@ impl Genesis {
         let genesis = vm_genesis::encode_genesis_transaction_with_validator(
             association_key,
             &validators,
-            Some(libra_types::on_chain_config::VMPublishingOption::open()),
+            Some(libra_types::on_chain_config::VMPublishingOption::open()), // TODO: intended?
+            ChainId::test(), // TODO: where should we read this from?
```
sblackshear, line 37: @JoshLind @davidiw: I am happy to fix this if it's easy or leave the TODO if not... Also, wondering  about the line  above.

davidiw: I think we'll want to add this to the layout. Feel free to leave it as a todo and we can jump on it, since the layout has other tweaks that need to be made.

sblackshear:
This should be good to go now; rebased on #4969 + added a test.

language/e2e-tests/src/tests/verify_txn.rs
```diff
@@ -442,6 +443,26 @@ pub fn test_publish_from_libra_root() {
...
+    let sender = AccountData::new(900_000, 0);
+    executor.add_account_data(&sender);
+    let private_key = Ed25519PrivateKey::generate_for_testing();
+    let txn = transaction_test_helpers::get_test_txn_with_chain_id(
+        *sender.address(),
+        0,
+        &private_key,
+        private_key.public_key(),
+        // all tests use TESTING != 253 for chain_id
+        ChainId::new(253),
```
tzakian, line 458: Nit: might be slightly fragile. Could you do something like `ChainId::new(ChainId::test() + 1)`?

language/libra-vm/src/transaction_metadata.rs
```diff
@@ -79,6 +86,7 @@ impl Default for TransactionMetadata {
             gas_unit_price: GasPrice::new(0),
             transaction_size: AbstractMemorySize::new(0),
             expiration_time: Duration::new(0, 0),
+            chain_id: ChainId::test(),
```
tzakian, line 89: I think this may cause issues on mainnet without an update to `process_block_prologue` in LibraVM. In particular, it currently uses `TransactionMetatdata::default` when spinning up a block prologue. So we need to make sure that the chain_id field is set correctly there. I think it's going to need to read the `ChainId` resource from chain for that though. @runtian-zhou @dariorussi  what are your thoughts on this?

sblackshear: I hate the `Default` trait...

sblackshear: Took a look and it seems that the `process_block_prologue` code path doesn't read the `chain_id` from `txn_data` and doesn't need to. If BlockMetadata were a user-submitted transaction, we would definitely need to check the chain ID there to prevent replays. But since BlockMetadata txes are only created/executed inside the VM, I think we are safe from replays.

sblackshear: It looks like `process_block_prologue` actually shouldn't need `txn_data` at all; perhaps Runtian's refactor will fix that.

tzakian: ah! I see. Totally makes sense, and this doesn't go through prologue etc. Good call.

PR #5079, opened 2020-07-14
[libra_vm] Implement WriteSet scripts

runtian-zhou:
## Motivation

Implement the WriteSet Script feature that allows association to send transaction scripts that get executed as writeset transactions. With this feature, the association can send a transaction that behaves as if it is another address.

Yes

## Test Plan

Working on examples to demonstrate the process.

### DIFF
--- a/language/e2e-tests/src/account.rs
+++ b/language/e2e-tests/src/account.rs
@@ -17,7 +17,7 @@ use libra_types::{
     event::EventHandle,
     transaction::{
         authenticator::AuthenticationKey, ChangeSet, Module, RawTransaction, Script,
-        SignedTransaction, TransactionArgument, TransactionPayload,
+        SignedTransaction, TransactionArgument, TransactionPayload, WriteSetPayload,
     },
     write_set::{WriteOp, WriteSet, WriteSetMut},
 };
@@ -209,9 +209,19 @@ impl Account {
         gas_currency_code: String,
     ) -> RawTransaction {
         match payload {
-            TransactionPayload::WriteSet(writeset) => {
+            TransactionPayload::WriteSet(WriteSetPayload::Direct(writeset)) => {
                 RawTransaction::new_change_set(address, sequence_number, writeset, ChainId::test())
             }
+            TransactionPayload::WriteSet(WriteSetPayload::Script {
+                execute_as: signer,
+                script,
+            }) => RawTransaction::new_writeset_script(
+                address,
+                sequence_number,
+                script,
+                signer,
+                ChainId::test(),
+            ),
             TransactionPayload::Module(module) => RawTransaction::new_module(
                 address,
                 sequence_number,
@@ -430,7 +440,7 @@ impl TransactionBuilder {
     }
 
     pub fn write_set(mut self, w: ChangeSet) -> Self {
-        self.program = Some(TransactionPayload::WriteSet(w));
+        self.program = Some(TransactionPayload::WriteSet(WriteSetPayload::Direct(w)));
         self
     }
 
--- a/language/libra-vm/src/libra_transaction_executor.rs
+++ b/language/libra-vm/src/libra_transaction_executor.rs
@@ -10,7 +10,7 @@ use crate::{
     },
     system_module_names::*,
     transaction_metadata::TransactionMetadata,
-    VMExecutor,
+    txn_effects_to_writeset_and_events, VMExecutor,
 };
 use libra_logger::prelude::*;
 use libra_state_view::StateView;
@@ -20,7 +20,7 @@ use libra_types::{
     block_metadata::BlockMetadata,
     transaction::{
         ChangeSet, Module, Script, SignatureCheckedTransaction, Transaction, TransactionArgument,
-        TransactionOutput, TransactionPayload, TransactionStatus,
+        TransactionOutput, TransactionPayload, TransactionStatus, WriteSetPayload,
     },
     vm_status::{StatusCode, VMStatus},
     write_set::{WriteSet, WriteSetMut},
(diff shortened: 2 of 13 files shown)


language/libra-vm/src/libra_transaction_executor.rs
```diff
@@ -386,9 +375,40 @@ impl LibraVM {
             )
             .map_err(|e| e.into_vm_status())?;
 
+        let change_set = match txn.payload() {
```
dariorussi, line 378: I wonder if we could work by "de-structuring" the transaction rather than working by ref and making a copy of something we are about to throw away. 
But maybe the API sadly does not allow that?

bob-wilson:
We had discussed this being merely a different flavor of writeset transactions. The proposed implementation introduces a new TransactionPayload::AdminScript, but how hard would it be to put an enum inside of TransactionPayload::WriteSet? I think that might be important to avoid confusion about this (if it doesn't mess with the implementation too much).

runtian-zhou:
> We had discussed this being merely a different flavor of writeset transactions. The proposed implementation introduces a new TransactionPayload::AdminScript, but how hard would it be to put an enum inside of TransactionPayload::WriteSet? I think that might be important to avoid confusion about this (if it doesn't mess with the implementation too much).

The intension at the moment is that we don't want this change to be a client-breaking change. That's why it is using a different enum. (In fact, I believe that is the major pattern that we will see once the mainnet is deployed). On the other hand, the implementation actually reused the exact same logic as `WriteSet` transaction so I would consider that it is yet another form of writeset txn.

dariorussi:
> > We had discussed this being merely a different flavor of writeset transactions. The proposed implementation introduces a new TransactionPayload::AdminScript, but how hard would it be to put an enum inside of TransactionPayload::WriteSet? I think that might be important to avoid confusion about this (if it doesn't mess with the implementation too much).
> 
> The intension at the moment is that we don't want this change to be a client-breaking change. That's why it is using a different enum. (In fact, I believe that is the major pattern that we will see once the mainnet is deployed). On the other hand, the implementation actually reused the exact same logic as `WriteSet` transaction so I would consider that it is yet another form of writeset txn.

understood, though if the change was to create an enum within the `WriteSet` we would have accomplished the same thing right? (apology for the names)

```
pub enum TransactionPayload {
    WriteSet(AdminOp),
    ...
}

pub enum AdminOp {
  Data(ChangeSet),
  Code(Script),
}

```
even wondering if WriteSet should become an AdminOperation and have code and data. Don't know...
I think @sblackshear @tzakian may clarify a bit more. @gc3 may also have a say here....

bob-wilson:
If the code isn't much worse, I suggest going with something like what Dario suggested. It makes it clear that these new scripts are a safer form of writesets without introducing some new entity that needs to be understood. If we can do the client-breaking part ASAP (even if the whole feature isn't done), then it should be fine.

runtian-zhou:
Just updated the PR as @dariorussi suggested

language/e2e-tests/src/account.rs
```diff
@@ -208,9 +208,12 @@ impl Account {
         gas_currency_code: String,
     ) -> RawTransaction {
         match payload {
-            TransactionPayload::WriteSet(writeset) => {
+            TransactionPayload::WriteSet(WriteSetPayload::Direct(writeset)) => {
```
gc3, line 211: recording for posterity that I think we should change the name from AdminScript to WritesetScript to reinforce that these are subsets of the existing writeset functionality.

language/e2e-tests/src/account.rs
```diff
@@ -208,9 +208,12 @@ impl Account {
...
                 RawTransaction::new_change_set(address, sequence_number, writeset)
             }
+            TransactionPayload::WriteSet(WriteSetPayload::AdminScript { signer, script }) => {
```
gc3, line 214: recording for posterity that I think we should change the name from AdminScript to WritesetScript to reinforce that these are subsets of the existing writeset functionality.

types/src/transaction/mod.rs
```diff
@@ -297,13 +309,24 @@ impl RawTransaction {
...
     /// A transaction that publishes code.
     Module(Module),
 }
 
+#[derive(Clone, Debug, Hash, Eq, PartialEq, Serialize, Deserialize)]
+pub enum WriteSetPayload {
+    /// Directly passing in the write set.
+    Direct(ChangeSet),
+    /// Generate the writeset by running a script.
+    WriteSetScript {
```
sblackshear, line 324: Nit: it's a bit asymmetric to have one called `WriteSetPayload::Direct` and one called `WriteSetPayload::WriteSetScript`. Maybe just `WriteSetPayload::Script`?

types/src/transaction/mod.rs
```diff
@@ -297,13 +309,24 @@ impl RawTransaction {
...
+        signer: AccountAddress,
```
sblackshear, line 325: Nits: 
- Perhaps rename to something like `execute_as` or `to_imitate` to convey that this isn't an ordinary use of `signer`?
- Could you amend the doc comment for `WriteSetScript` to explain what execution does with this address?

runtian-zhou:
- Renamed `WriteSetScript` to `Script`
- Renamed `signer` to `execute_as`
- Added the comment for `execute_as` field

bob-wilson:
I edited the PR title and description to match the new name

dariorussi:
let us know when the PR is up for review. I think people are happy about this at this point.
Do we have anything still open?

dariorussi:
> let us know when the PR is up for review. I think people are happy about this at this point.
> Do we have anything still open?

and should we push it before tomorrow?

bob-wilson:
@runtian-zhou Can you edit the git commit message to update the name before you land this?

PR #5080, opened 2020-07-14
[cluster-test] Added vault and LSR instances into Cluster

snakech:
## Motivation

Added vault and LSR instances into Cluster

(Write your answer here.)

## Test Plan

## Related PRs

### DIFF
--- a/testsuite/cluster-test/src/cluster_builder.rs
+++ b/testsuite/cluster-test/src/cluster_builder.rs
@@ -12,10 +12,7 @@ use crate::{
     },
 };
 use anyhow::{format_err, Result};
-use futures::{
-    future::{join, try_join_all},
-    try_join,
-};
+use futures::{future::try_join_all, try_join};
 use libra_logger::info;
 use structopt::StructOpt;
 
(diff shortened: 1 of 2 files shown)


testsuite/cluster-test/src/cluster_builder.rs
```diff
@@ -129,8 +126,9 @@ impl ClusterBuilder {
         image_tag: &str,
         config_overrides: &[String],
         delete_data: bool,
-    ) -> Result<Vec<Instance>> {
+    ) -> Result<(Vec<Instance>, (Vec<Instance>, Vec<Instance>))> {
```
andll, line 129: Is there benefit of using double nested tuple like `(vec, (vec, vec))`, instead of `(vec, vec, vec)`?

andll:
Can you rename commit message so it is more clear on specific change? 'lsr/vault followup supports' is more of a status of overall vault work, in a few weeks reading this commit message will give no clue on what have happened there.

andll:
cc @davidiw @JoshLind

testsuite/cluster-test/src/cluster_builder.rs
snakech: My thought is logically combine vault and LSR together in a nested tuple. I can flat them if you think that is redundant

andll: Yeah, I would rather flatten them

PR #5081, opened 2020-07-14
[libra framework] collapse update_compliance_public_key and update_ba…

sblackshear:
…se_url into a single tx script

- This allows atomic updates of the endpoint and public key used for the dual attestation APIs. This is important if you want to change both at the same time (as one might expect in practice).
- This script will be used by both VASPs and DDs
- This shrinks the public-facing script API by a bit, which always helps


sunmilee:
This breaks https://github.com/libra/libra/blob/master/client/libra-dev/src/transaction.rs#L246-L328, the two functions will have to be consolidated, let me know if you run into issues merging

sblackshear:
> This breaks https://github.com/libra/libra/blob/master/client/libra-dev/src/transaction.rs#L246-L328, the two functions will have to be consolidated, let me know if you run into issues merging

Thanks, I will fix them! But just to double-check: is this change an improvement from the client perspective?

sunmilee:
Do rotating base URL and compliance key  always need to happen together? If so this is a simplification/improvement. What if clients want to do only  one or is that not allowed

sblackshear:
> Do rotating base URL and compliance key always need to happen together? If so this is a simplification/improvement. 

They don't always need to, but I would imagine that they often would (e.g., when you spin up a new endpoint, it's probably easier to create a new private key than it is to safely share the old one). 

> What if clients want to do only one or is that not allowed

If they want to do only one (e.g., `compliance_public_key` only), they can read the old value of `base_url` from on-chain and pass the old base url + new key to the script. We could also use a convention that empty url or empty key means "don't update", but that seemed slightly more complex to me.

PR #5083, opened 2020-07-14
[network] add noise specification

mimoo:
Adding the network specification for noise


mimoo:
and get the rest of the network specs here @phlip9  ;)

phlip9:
bors github API call failed 😭 

@bors-libra retry

PR #5084, opened 2020-07-14
[network] fix conn_limit underflow error

gregnazario:
Connection limit was underflowing on an unsigned number causing panics.
This should replace it without causing underflow.

### DIFF
--- a/network/src/connectivity_manager/mod.rs
+++ b/network/src/connectivity_manager/mod.rs
@@ -348,7 +348,7 @@ where
         // including in flight dials.
         let to_connect_size = if let Some(conn_limit) = self.connection_limit {
             min(
-                conn_limit - self.connected.len() - self.dial_queue.len(),
+                conn_limit.saturating_sub(self.connected.len() + self.dial_queue.len()),
                 to_connect.len(),
             )
         } else {


phlip9:
perfect!

@bors-libra r+

network/src/connectivity_manager/mod.rs
```diff
@@ -347,10 +347,12 @@ where
...
-            min(
-                conn_limit - self.connected.len() - self.dial_queue.len(),
-                to_connect.len(),
-            )
+            let current = self.connected.len() + self.dial_queue.len();
+            if current < conn_limit {
+                min(conn_limit - current, to_connect.len())
+            } else {
+                0
+            }
```
davidiw, line 355: https://doc.rust-lang.org/std/primitive.u32.html
this can be written as
```
conn_limit.checked_sub(self.connected.len()).and_then(|v| v.checked_sub(to_connect.len())).unwrap_or(0);
```
or you can do the add too

davidiw:
@bors-libra r-

hint on slightly more readable code / rusty code

davidiw:
feel free to land as is, just wanted to give another approach.

gregnazario:
> feel free to land as is, just wanted to give another approach.

I figured there'd be a more "rusty" way to do it

network/src/connectivity_manager/mod.rs
gregnazario: `saturating_sub` does `checked_sub().unwrap_or(0)`

network/src/connectivity_manager/mod.rs
```diff
@@ -348,7 +348,7 @@ where
         // including in flight dials.
         let to_connect_size = if let Some(conn_limit) = self.connection_limit {
             min(
-                conn_limit - self.connected.len() - self.dial_queue.len(),
+                conn_limit.saturating_sub(self.connected.len() + self.dial_queue.len()),
```
mimoo, line 351: oh I didn't know we could do this! This is going to simplify my stuff :D

PR #5085, opened 2020-07-14
[consensus][spec] cleaning blockdata checks

mimoo:


specifications/consensus/consensus.md
```diff
@@ -48,10 +48,13 @@ Fields:
...
-Verification
+Verification:
 
-* if `signature` is set, ensure `block_data.block_type == BlockType::Proposal {author, ..}`
-* `ed25519_verify(signature, block_data.hash(), OnChainConfig.validators().public_key(author))`
+* if this block is received through a proposal, the `signature` field MUST be set:
+    - ensure that the block is a legitimate proposal (`block_data.block_type == BlockType::Proposal {author, ..}`)
+    - verify the block's signature by calling `verify(signature, block_data.hash(), OnChainConfig.validators().public_key(author))`
+* if this block is received via a BlockRetrievalResponse, it is possible that the signature is not set. In this case verify that the block is a NIL block (`block_data.block_type == BlockType::NilBlock`)
+* verify `block_data` according to [BlockData](#blockdata)
```
mimoo, line 57: we were missing this check

specifications/consensus/consensus.md
```diff
@@ -88,15 +91,12 @@ Fields
...
 
 * ensure that if `BlockData` is received from the network that `block_type`, is not genesis. Genesis is only constructed from end-epoch LedgerInfo locally and is never sent to other validators.
-* Assume that `parent_block` is the `BlockInfo` of `quorum_cert.vote_data.proposed`.
-
-  * ensure `round` is greater than `parent_block.round`
-  * ensure `epoch` is equal to `parent_block.epoch`
-  * if `parent_block.next_epoch_state` is set, ensure `block_type == BlockType::NIL` or the payload in the `BlockType::Proposal` is empty
-  * if a block is a nil block, or if `parent_block.has_reconfiguration`, ensure the block and the parent block have the same timestamp. otherwise ensure `timestamp_usecs` > `parent_block.timestamp_usecs`
-
-* ensure that if `quorum_cert.commit_info().next_epoch_state()` is set, a proposal should not be generated if the epoch ends, it should transition to next epoch.
```
mimoo, line 98: I moved this to `process_proposal_msg`

mimoo: looking more closely at `BlockRetrievalResponse` it looks like we verify that this is not set as well. So it looks like we want to just rephrase this to ensure that `quorum_cert.commit_info().next_epoch_state()` is not set.

but this contradits my understanding of reconfig, that says that there is a last block that carries the commit_info().next_epoch_state()  (see https://github.com/novifinancial/libra-specs/blob/master/consensus/reconfig.png at round 4)

mimoo: ok another issue with the spec is that we don't specify where/when `next_epoch_state` is set (in QC and ledger_info.commit).

Another thing that I think needs to be specified somewhere (not sure if consensus is the correct place though): blockmetadata txn

specifications/consensus/consensus.md
```diff
@@ -1114,6 +1114,8 @@ fn process_proposal_msg(&mut self, proposal_msg: ProposalMsg) -> Result<()> {
     let ProposalMsg {proposal, sync_info} = proposal_msg;
     self.ensure_epoch(proposal.epoch())?;
     self.ensure_round(proposal.round(), sync_info)?;
+    proposal_msg.verify()?;
```
mimoo, line 1117: in reality we check the epoch, then we verify using `validator_verifier`, then we check the round/sync_info

davidiw:

PR #5092, opened 2020-07-14
[libra-framework][move-prover] Some more specs around AccountLimits + prover fixes

wrwg:
This PR further refines specs for LibraAccount and AccountLimits. The major functions `AccountLimits::update_deposit_limits/update_withdrawal_limit` had previously been mocked out, and are now fully specified. `LibraAccount::deposit/withdraw_from_balance` based on this have also been specified and verified. Functions on top of those (`pay_from` et. al) time out during verification, though. There are also intrinsic problems in specifying such functions because of the intermediate state problem which makes it very hard to specify pre/post conditions (e.g. a withdraw followed by a deposit, or two deposits in sequence). This is far from done, and needs likely additional language support, and optimization of the VC to make it better feasible for Z3.

This PR also introduces a few changes in the prover:

- A bug in schema inclusion dependency analysis has been fixed.
- A bug in type inference in schema expressions has been fixed.
- The prover does not longer generate by default `output.bpl[.log]` files, as people continue to check those in. One needs to provide the `-g` flag to have those files.
- Abstract and concrete modifiers for conditions have been introduced as they were needed at some point, though they aren't any longer. But we may need those in the future.

## Motivation

Specs

Yes

## Test Plan

Prover tests.

## Related PRs

NA


language/stdlib/modules/LibraAccount.move
```diff
@@ -151,7 +151,7 @@ module LibraAccount {
         if (is_withdrawal) {
             VASP::is_vasp(payer) && (!VASP::is_vasp(payee) || !VASP::is_same_vasp(payer, payee))
         } else {
-            VASP::is_vasp(payee) && (!VASP::is_vasp(payee) || !VASP::is_same_vasp(payer, payee))
+            VASP::is_vasp(payee) && (!VASP::is_vasp(payer) || !VASP::is_same_vasp(payee, payer))
```
wrwg, line 154: @sblackshear @tzakian I think this was a bug in the code. Discovered while trying to verify things.

sblackshear: Nice!!

PR #5094, opened 2020-07-15
[consensus] minor improvement

zekun000:
## Motivation

1. Prevent large block retrieval requests.
2. Reduce the timeout exponent so that maximum round is 3s instead of
11s.

(Write your answer here.)

## Test Plan

## Related PRs

### DIFF
--- a/consensus/consensus-types/src/block_retrieval.rs
+++ b/consensus/consensus-types/src/block_retrieval.rs
@@ -8,6 +8,8 @@ use libra_types::validator_verifier::ValidatorVerifier;
 use serde::{Deserialize, Serialize};
 use std::fmt;
 
+pub const MAX_BLOCKS_PER_REQUEST: u64 = 10;
+
 /// RPC to get a chain of block of the given length starting from the given block id.
 #[derive(Serialize, Deserialize, Clone, Debug, PartialEq, Eq)]
 pub struct BlockRetrievalRequest {
--- a/consensus/src/epoch_manager.rs
+++ b/consensus/src/epoch_manager.rs
@@ -134,7 +134,7 @@ impl EpochManager {
         // Timeout goes from initial_timeout to initial_timeout*11 in 6 steps
         let time_interval = Box::new(ExponentialTimeInterval::new(
             Duration::from_millis(self.config.round_initial_timeout_ms),
-            1.5,
+            1.2,
             6,
         ));
         RoundState::new(time_interval, time_service, timeout_sender)
--- a/consensus/src/network.rs
+++ b/consensus/src/network.rs
@@ -9,7 +9,7 @@ use anyhow::{anyhow, ensure};
 use bytes::Bytes;
 use channel::{self, libra_channel, message_queues::QueueStyle};
 use consensus_types::{
-    block_retrieval::{BlockRetrievalRequest, BlockRetrievalResponse},
+    block_retrieval::{BlockRetrievalRequest, BlockRetrievalResponse, MAX_BLOCKS_PER_REQUEST},
     common::Author,
     proposal_msg::ProposalMsg,
     sync_info::SyncInfo,
@@ -263,6 +263,13 @@ impl NetworkTask {
                 Event::RpcRequest((peer_id, msg, callback)) => match msg {
                     ConsensusMsg::BlockRetrievalRequest(request) => {
                         debug!("Received block retrieval request {}", request);
+                        if request.num_blocks() > MAX_BLOCKS_PER_REQUEST {
+                            warn!(
+                                "Ignore block retrieval with too many blocks: {}",
+                                request.num_blocks()
+                            );
+                            continue;
+                        }
                         let req_with_callback = IncomingBlockRetrievalRequest {
                             req: *request,
                             response_sender: callback,
--- a/consensus/src/liveness/round_state.rs
+++ b/consensus/src/liveness/round_state.rs
@@ -80,7 +80,6 @@ pub struct ExponentialTimeInterval {
     max_exponent: usize,
 }
 
-#[allow(dead_code)]
 impl ExponentialTimeInterval {
     #[cfg(any(test, feature = "fuzzing"))]
     pub fn fixed(duration: Duration) -> Self {
--- a/specifications/consensus/consensus.md
+++ b/specifications/consensus/consensus.md
@@ -913,6 +913,10 @@ fn new(sync_info: SyncInfo) -> RoundState {
 * enter new round with quorum/timeout certificates
 * calculate timeout value as exponential of unhappy round gap
 * start timer to signal a local timeout of `round` after `timeout`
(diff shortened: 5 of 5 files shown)


consensus/consensus-types/src/block_retrieval.rs
```diff
@@ -8,6 +8,8 @@ use libra_types::validator_verifier::ValidatorVerifier;
 use serde::{Deserialize, Serialize};
 use std::fmt;
 
+pub const MAX_BLOCKS_PER_REQUEST: u64 = 10;
```
davidiw, line 11: do we want to add this to a config? do we have any metrics on average / max blocks per request on any network?

consensus/src/network.rs
```diff
@@ -263,6 +263,13 @@ impl NetworkTask {
                 Event::RpcRequest((peer_id, msg, callback)) => match msg {
                     ConsensusMsg::BlockRetrievalRequest(request) => {
                         debug!("Received block retrieval request {}", request);
+                        if request.num_blocks() > MAX_BLOCKS_PER_REQUEST {
```
davidiw, line 266: should we use min(request.num_blocks(), MAX_BLOCKS_PER_REQUEST);

davidiw: I'd be concerned if we bump this number up and someone is running an old service, that old service would just reject everyone....

zekun000: we currently only use two fix numbers 1 and 3

consensus/src/epoch_manager.rs
```diff
@@ -134,7 +134,7 @@ impl EpochManager {
         // Timeout goes from initial_timeout to initial_timeout*11 in 6 steps
         let time_interval = Box::new(ExponentialTimeInterval::new(
             Duration::from_millis(self.config.round_initial_timeout_ms),
-            1.5,
+            1.2,
```
aching, line 137: This round timeout should be part of the consensus specification btw.  https://github.com/libra/libra/blob/master/specifications/consensus/consensus.md

aching: How are we deciding that the upper bound should be 3 seconds btw?  For instance if the network latency did slow to 11, we would still retain quorum and be able to remove high latency nodes rather than rely on a genesis transaction to get healthy again.  Personally I prefer the current upper bound.

davidiw:
@bors-libra delegate+
seems like there's some interesting client code here... probably some work to be done later

consensus/consensus-types/src/block_retrieval.rs
aching: Also curious how to come up with this number?  400 bytes * 1000 * 10 = 4MB?  Seems reasonable.

consensus/src/epoch_manager.rs
zekun000: a reasonable network latency number imo. when it gets up to 11 sec to get a quorum (instead of something broken), it's probably worth some manual intervention.

zekun000: why 11 sec is bad? it hurts latency badly when some proposers have problems. there's a lot of data of how mis-configured nodes hurt the performance because the timeout hits the upper bound and every round takes 11 sec to timeout.

consensus/consensus-types/src/block_retrieval.rs
zekun000: random small number :)

consensus/src/epoch_manager.rs
zekun000: I don't think round timeout is part of the spec, it's a local config value. the system still works if all nodes have different timeout value as long as it's not small and they eventually timeout.

zekun000:
update spec with suggested value

consensus/src/epoch_manager.rs
aching: Agreed that anytime that latency gets too high, someone should look into it, that would be true for both 3 and 11 seconds.  I just would like to have more tools available for fixing the issue rather than relying on genesis transactions.

The explicit timeout backoff doesn't need to be part of the spec, but it could be to make it easier for implementors.  Even if the exact formula isn't defined, we should at least define backoff guidance to ensure liveness is preserved.  Still as an implementor, I would prefer a precise formula and it will result in more uniform behavior.

zekun000: I added suggested timeout to the spec, for the timeout value, it's trade-off between more tools and worse performance. imo there's very little chance things are broken with 3 sec but not with 11 sec.

aching: Thanks for adding it to the spec.  

In a recent SEV, timeout of hundreds of ms of delay to safety rules were causing issues.  I'd like more than 3 seconds of buffer before we lose the ability for on-chain consensus to resolve issues.  I'd be open to increasing the max timeout btw to 20-30 seconds, but was okay with 11 seconds.

davidiw: aaaahhhh, this is not true... there was a completely unrelated issue that caused a SEV... someone (some :troll:) just mistakingly pointed out the LSR latency :). LSR latency is > 100 ms across all CT I've looked at.

zekun000: 3s = 11s = forever imo.  200ms is 10x less than 3s

davidiw: I think there's a fun philosophical argument to be made here... if you have effective scale up / scale down, your maximum just has to be bounded within reason. We'd only hit that maximum if things are really bad. At which point, we're really saying let's reduce our cost and just be patient. On the flip side, one could argue, that the upper bound could also be the worst-case we expect, something like 5*stddev + avg or your p99.99~ in a healthy network. I guess there just the tradeoff of how expensive is noise. In our case, I think lower timeouts help us recover slightly faster. How fast is consensus now between proposals on average and what's the stddev?

aching: > it hurts latency badly when some proposers have problems. there's a lot of data of how mis-configured nodes hurt the performance because the timeout hits the upper bound and every round takes 11 sec to timeout.

This only occurs in the phase where we restart the blockchain generally in a testing environment.  It should not be a common occurrence in mainnet since testing is done ahead of time.  I think the extra upper bound for consensus is a valuable tool rather than relying on genesis transactions as the only tool to resolve a situation where round latencies exceed 3 seconds.  Sometimes network latencies could spike to seconds (https://blog.cloudflare.com/the-story-of-one-latency-spike/) and a buffer to keep an on-chain reconfiguration tool functional under unexpected circumstances would help resolve issues more quickly and smoothly.

zekun000: I still don't get it, why the only tooling is genesis transactions if the upper bound is 3s? if we believe relaxing the round timeout can fix the issue we can easily override the base interval to 5s which results in a 15s upper bound.

PR #5098, opened 2020-07-15
[docker] update for chain ID

christinacelee:
## Motivation

Updating docker-related files to include chain ID:
- CLI command
- Docker mint

### DIFF
--- a/docker/mint/docker-run.sh
+++ b/docker/mint/docker-run.sh
@@ -6,11 +6,6 @@ set -ex
 export RUST_BACKTRACE=full
 
 declare -a params
-if [ -z "${CHAIN_ID}" ]; then
-  echo "Docker run failed to start, please set required env var CHAIN_ID"
-  exit 1
-fi
-
 if [ -n "${CFG_SEED}" ]; then
     params+="-s $CFG_SEED "
 fi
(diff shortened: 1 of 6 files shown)


christinacelee:
we should discuss/agree how to set the environment variables for devnet and testnet
devnet should be "DEVNET" or 3, testnet should be "TESTNET" or 2

sausagee:
I'd say DEVNET/TESTNET over the numeric. It is easier to understand when looking at the terraform config. 

BTW, I should have given you a heads up before pushing to your branch.

christinacelee:
oh I meant where in the prod workflow we want to set the env var - rn in this PR for example, it's being set in the Dockerfile to "TESTING", which won't work for dev/testnet. I just want to confirm that each *net release will correctly set the env var

sherry-x:
It's set here https://github.com/libra/libra/blob/master/terraform/templates/faucet.json#L11
in terraform we can make a variable and set different value accordingly

docker/mint/docker-run.sh
```diff
@@ -6,6 +6,11 @@ set -ex
 export RUST_BACKTRACE=full
 
 declare -a params
+if [ -z "${CHAIN_ID}" ]; then
+  echo "Docker run failed to start, please set required env var CHAIN_ID"
+  exit 1
```
sherry-x, line 11: Do you need to add this parameter in the command when running the config-builder for faucet? see like 22-24

sherry-x: If not needed, I think we don't have to set the env in this script

PR #5099, opened 2020-07-15
[language] Fix deserializer bug

dariorussi:
Fresh from the fuzzer...
Basically there was a way to trick the table offsets and make the
deserializer crash


dariorussi:
T70008241 for more details, going to add some tests

dariorussi:
test will come at a later time because I am tired and I am not sure about more things so... that is it....

PR #5101, opened 2020-07-15
[gha] select cluster-test suite based on breaking

### LINKED ISSUE #5172: [gha][cluster-test] add compatibility test to LBT
Detect breaking flag in commit message and determine which cluster-test experiment suite to run, reporting accordingly to PR comments.

rustielin:
## Motivation

Resolves #5172 

Select whether to run `land_blocking` or `land_blocking_compat` experiment suite for cluster-test in LBT based on whether the PR commit message has `[breaking]`. Having it in the commit message will not only inform LBT which suite to run, but also as an indicator of which commits are breaking without relying on Github PR labels:

```
git log $(git log origin/testnet --format=format:"%h" -1)..HEAD --pretty=oneline | grep '\[breaking\]''
```

Configures `land_blocking_compat` to run compatibility test against the previous commit in `master`.

Yes

## Test Plan

Canary in GHA:
* with `[breaking]` flag: https://github.com/libra/libra/runs/892241240?check_suite_focus=true

## Related PRs

Depends on https://github.com/libra/libra/pull/5150, which introduces prettier report text for LBT results.

### DIFF
--- a/.github/workflows/land-blocking.yml
+++ b/.github/workflows/land-blocking.yml
@@ -19,6 +19,7 @@ jobs:
         run: |
           echo "::set-env name=LIBRA_GIT_REV::$(git rev-parse --short=8 HEAD)"
           echo "::set-env name=TEST_TAG::land_$(git rev-parse --short=8 HEAD)"
+          echo "::set-env name=PREV_TAG::land_$(git rev-parse --short=8 origin/master)"
       - name: Check kill switch
         id: check_ks
         run: |
@@ -49,6 +50,22 @@ jobs:
           done
           echo "Build failed after retries"
           exit 1
+      - name: Determine which cluster-test suite to run
+        if: steps.check_ks.outputs.should_run == 'true'
+        run: |
+          set +e
+          commit_message=$(git log -1 --pretty=oneline)
+          echo $commit_message | grep '\[breaking\]'
+          if ${{ secrets.KILL_SWITCH_LAND_BLOCKING_COMPAT }}; then
+            echo "Compat killswitch activated! Will run land_blocking suite"
+            echo "::set-env name=TEST_COMPAT::0"
+          elif [ $? -eq 0 ]; then
+            echo "Breaking change detected! Will run land_blocking suite"
+            echo "::set-env name=TEST_COMPAT::0"
+          else
+            echo "Will run land_blocking_compat suite"
+            echo "::set-env name=TEST_COMPAT::1"
+          fi
       - name: Launch cluster test
         if: steps.check_ks.outputs.should_run == 'true'
         # NOTE Remember to update PR comment payload if cti cmd is updated.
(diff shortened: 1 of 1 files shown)


andll:
Few comments about code:
- You want to analyze all commits between origin/master and HEAD - PR can contain more then one commit
- I would suggest using smth like `pretty=oneline` to only search for breaking tag in commit header. I suggest this because this way we can see breaking changes in the `#changelog`
- Ideally can we have some knob that will disable compat test without making commit, similar how we now have killswitch to disable land blocking test?


And a few meta comments on the procedure:
- We should make a post announcing this change _before_ we land this diff. Ideally give it at least a day to gather feedback
- There might be some confusion about 'breaking' changes, as there are different types of breaking changes. In general, I would say there are server-breaking and client-breaking changes. Current compat test focuses on server-breaking changes, and many people are actually interested in client-breaking changes. Should we have separate labels for those?

andll:
Actually, on client vs server breaking I am not so sure. Maybe it will just overcomplicate things and we can just leave it as 'breaking', just highlight that client compatibility check is not perfect at the moment and there will be more work to do in this area

rustielin:
Thanks for the feedback @andll ! few comments:

> You want to analyze all commits between origin/master and HEAD - PR can contain more then one commit

I might be wrong, but as far as I'm aware, Bors will squash the PR commits for us in `auto`, so just checking the previous commit is fine.

> Should we have separate labels for those? [client and server breaking changes]

I think we can just leave it as `[breaking]` for now, and rely on the report text to see what actually broke. I imagine in the future we'll have a separate experiment that will use various clients to test against the cluster, and we can report errors there.

andll:
Bor's won't necessarily squash commits at one - I think it depends on some settings, but there are certainly PRs that push more then one commit to master

I think keeping single `[breaking]` label is ok for now

rustielin:
> Ideally can we have some knob that will disable compat test without making commit, similar how we now have killswitch to disable land blocking test?

I think we can just reuse the existing LBT killswitch, unless you mean to still run `land_blocking` suite and not `land_blocking_compat`? User can still tag their commit as breaking and try to land their PR again.

andll:
Yeah, I meant separate kill switch that will just make all PRs to use land_blocking, regardless of commit message

We have barely tried compat test - I think activating it without ability to quickly disable is a bit dangerous. And disabling entire land blocking test when we discover problem in compat is also far from ideal, as we've seen in the past, disabling land blocking test quickly leads to bad things getting into master

sausagee:
Why do we need to create this dependency based on the `[breaking]` tag in the commit message?  Shouldn't we run the compat test by default?

.github/workflows/land-blocking.yml
```diff
@@ -57,11 +74,22 @@ jobs:
...
+          if [ $TEST_COMPAT -eq 1 ]; then
+            ./scripts/cti \
+              --tag ${PREV_TAG} \
+              --cluster-test-tag ${TEST_TAG} \
+              -E RUST_LOG=debug \
+              -E BATCH_SIZE=15 \
+              -E UPDATE_TO_TAG=${TEST_TAG} \
+              --report report.json \
+              --suite land_blocking_compat
+          else
```
sausagee, line 86: @andll I thought we want to run land-blocking on all PRs

rustielin:
@sausagee compat is run by default, but can be avoided by using `[breaking]` tag if breaking changes *must* be made. Otherwise, compat test is run by default, and its suite `land_blocking_compat` is essentially the `land_blocking` suite but with compat test before running the benchmark

.github/workflows/land-blocking.yml
andll: Yes, so as @rustielin mentioned, we want to run some cluster test on all PRs

By default, we are running land blocking + compat test, unless user explicitly specifies that the commit is a breaking change, in which case we are falling back to regular becnhmark without compatibility.

Reason for this is that we _are_ going to have breaking changes, so we need to account for this in the our land hook

rustielin:
Updated with new `DISABLE_FN_UPGRADE=1` from https://github.com/libra/libra/pull/5264

andll:
PR looks good

I wonder if there is (easy) way to define repro CMD once, instead of having two pieces of code, once where we execute it, and second time when printing it

PR #5104, opened 2020-07-15
[transaction builder generator] add generation of Java transaction builders

matbd:
## Motivation

* Support Java clients
* Add a code example in all supported languages
* Test that the code examples output the expected serialization of a demo transfer script in Libra.
* Update [README](https://github.com/libra/libra/tree/master/language/transaction-builder-generator) with quick start instructions

## Test Plan

```
cargo x test -p transaction-builder-generator
# We don't have Java in Libra's CI.
cargo x test -p transaction-builder-generator -- --ignored
```

## Related PRs

https://github.com/facebookincubator/serde-reflection/pull/27

### DIFF
--- a/language/transaction-builder-generator/src/java.rs
+++ b/language/transaction-builder-generator/src/java.rs
@@ -1,0 +1,192 @@
+// Copyright (c) The Libra Core Contributors
+// SPDX-License-Identifier: Apache-2.0
+
+use crate::common::type_not_allowed;
+use libra_types::transaction::{ArgumentABI, ScriptABI, TypeArgumentABI};
+use move_core_types::language_storage::TypeTag;
+
+use std::{
+    io::{Result, Write},
+    path::PathBuf,
+};
+
+/// Output transaction builders in Java for the given ABIs.
+pub fn output(
+    out: &mut dyn Write,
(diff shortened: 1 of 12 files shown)


language/transaction-builder-generator/src/java.rs
```diff
@@ -0,0 +1,191 @@
...
+fn output_preamble(out: &mut dyn Write, package: Option<&str>) -> Result<()> {
+    if let Some(name) = package {
+        writeln!(out, "package {};\n", name)?;
+    }
+    writeln!(
+        out,
+        r#"
+import java.math.BigInteger;
+import java.util.Arrays;
+import libra.types.AccountAddress;
```
thefallentree, line 38: I suggest we use package org.libra.something  here

matbd: `org.libra.types` and `org.libra.lcs` make sense but then it gets tricky with `bincode` and `serde`. I need to think about it.

matbd: Yes, I think I can work with `org.libra` and `com.facebook`.

matbd: @thefallentree addressed

PR #5106, opened 2020-07-15
[management] Improve secure config / storage 

davidiw:
The secure configs were really a mess and so was storage... this isn't exactly how I want it, but at least now the naming is accurate, the errors are minimized and unified and it is much easier to specified if you want an optional or non-optional storage.

### DIFF
--- a/config/management/src/error.rs
+++ b/config/management/src/error.rs
@@ -14,22 +14,14 @@ pub enum Error {
     BackendParsingError(String),
     #[error("Invalid arguments: {0}")]
     CommandArgumentError(String),
-    #[error("Local storage unavailable, please check your configuration: {0}")]
-    LocalStorageUnavailable(String),
-    #[error("Failed to read, {0}, from local storage: {1}")]
-    LocalStorageReadError(&'static str, String),
-    #[error("Failed to sign {0} with {1} using local storage: {2}")]
-    LocalStorageSigningError(&'static str, &'static str, String),
-    #[error("Failed to write, {0}, to local storage: {1}")]
-    LocalStorageWriteError(&'static str, String),
-    #[error("No remote storage was specified, but one is required!")]
-    RemoteStorageMissing,
-    #[error("Failed to read, {0}, from remote storage: {1}")]
-    RemoteStorageReadError(&'static str, String),
-    #[error("Failed to write, {0}, to remote storage: {1}")]
-    RemoteStorageWriteError(&'static str, String),
-    #[error("Remote storage unavailable, please check your configuration: {0}")]
-    RemoteStorageUnavailable(String),
+    #[error("{0} storage unavailable, please check your configuration: {1}")]
+    StorageUnavailable(&'static str, String),
+    #[error("Failed to read, {1}, from {0} storage: {2}")]
+    StorageReadError(&'static str, &'static str, String),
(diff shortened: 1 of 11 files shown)


config/management/src/genesis.rs
```diff
@@ -19,7 +21,7 @@ use vm_genesis::ValidatorRegistration;
 #[derive(Debug, StructOpt)]
 pub struct Genesis {
     #[structopt(flatten)]
-    pub backend: SingleBackend,
+    pub backend: SharedBackend,
```
JoshLind, line 19: nit: I'd consider renaming "backend" -> "shared_backend" to be consistent and explicit? (e.g., here, in verify() and layout()?)

config/management/src/secure_backend.rs
```diff
@@ -162,6 +163,55 @@ impl TryInto<Storage> for SecureBackend {
...
+    InMemory: "backend=memory"
+    OnDisk: "backend=disk;path=LOCAL_PATH"
+                "#)
+            )]
+            pub $field_name: $struct_type,
+        }
+    }
+}
+
+secure_backend!(
```
JoshLind, line 188: nit: I guess the only downside of using field names matching the struct name is that we have things like "self.validator_backend.validator_backend" in the code. Perhaps consider making all field names just be secure_backend so they're consistent?

config/management/src/error.rs
```diff
@@ -14,20 +14,14 @@ pub enum Error {
...
-    #[error("Remote storage unavailable, please check your configuration: {0}")]
-    RemoteStorageUnavailable(String),
+    #[error("{0} storage unavailable, please check your configuration: {1}")]
+    StorageUnavailable(&'static str, String),
+    #[error("Failed to read, {1}, from {0} storage: {2}")]
+    StorageReadError(&'static str, &'static str, String),
+    #[error("Failed to sign {1} with {2} using {0} storage: {2}")]
+    StorageSigningError(&'static str, &'static str, &'static str, String),
+    #[error("Failed to write, {1}, to {0} storage: {2}")]
+    StorageWriteError(&'static str, &'static str, String),
```
JoshLind, line 24: tiny nit: it seems odd to me that we're using {1}, then {0}, then {2}? Perhaps we can change the call sites so that we can just using the arguments in order? e.g., Error::StorageReadError(key_name, storage_name, e.to_string()) still makes sense?

config/management/src/secure_backend.rs
```diff
@@ -17,12 +17,6 @@ pub const GITHUB: &str = "github";
 pub const MEMORY: &str = "memory";
 pub const VAULT: &str = "vault";
 
-#[derive(Copy, Clone, Debug)]
-pub enum StorageLocation {
```
JoshLind, line 20: Nice! Always thought this was a little odd..

config/management/src/genesis.rs
davidiw: there's only one backend, so it seemed excessive, I'll see if Michael notices or complains ;)

config/management/src/error.rs
davidiw: I personally wanted the types passed in to be indicative of their status... the first one tells you whom or what is the source, the next one the actor or retrieved piece, and the 3rd is the extra info. I agree it is weird. This is where using named types would be better but also ... uglier... Rust :D

PR #5112, opened 2020-07-15
Set mint tiers to appropriate fiat values using scaling factor

moeziniaf:
This should be in line with AOS tiers now, as they were off by scaling_factor (10^6).

Tests: functional testing of tiered_mint with new tiers

### DIFF
--- a/language/stdlib/modules/DesignatedDealer.move
+++ b/language/stdlib/modules/DesignatedDealer.move
@@ -54,8 +54,9 @@ module DesignatedDealer {
     /// The maximum number of tiers allowed
     const MAX_NUM_TIERS: u64 = 4;
 
-    /// Default amounts for tiers when a DD is created
-    const TIER_0_DEFAULT: u64 = 500000;
+    /// Default FIAT amounts for tiers when a DD is created
+    /// These get scaled by coin specific scaling factor on tier setting
+    const TIER_0_DEFAULT: u64 = 500000; // ex $500,000
     const TIER_1_DEFAULT: u64 = 5000000;
     const TIER_2_DEFAULT: u64 = 50000000;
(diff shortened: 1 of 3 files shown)


language/stdlib/modules/DesignatedDealer.move
```diff
@@ -100,10 +101,12 @@ module DesignatedDealer {
             window_inflow: 0,
             tiers: Vector::empty(),
         });
-        add_tier<CoinType>(tc_account, dd_addr, TIER_0_DEFAULT);
-        add_tier<CoinType>(tc_account, dd_addr, TIER_1_DEFAULT);
-        add_tier<CoinType>(tc_account, dd_addr, TIER_2_DEFAULT);
-        add_tier<CoinType>(tc_account, dd_addr, TIER_3_DEFAULT);
+        // Add tier amounts in base_units of CoinType
+        let coin_scaling_factor = Libra::scaling_factor<CoinType>();
+        add_tier<CoinType>(tc_account, dd_addr, TIER_0_DEFAULT * coin_scaling_factor);
```
sblackshear, line 106: This is safe because the scaling factor is immutable, right?

tzakian: Yep! Good callout, scaling factor is an immutable field in currency info.

moeziniaf: 👍 . There's no way to make data in move always immutable is there?

tzakian: In the current version of the stdlib there is no way to mutate this field, so we have immutability via the module's public functions. But, the module could be updated to mutate it, or a writeset could happen. However both of these are incredibly unlikely, and we would have to factor this if mutation is performed/allowed at a latter date.

PR #5113, opened 2020-07-15
[language][vm] move Rc<RefCell<>> into Container

vgao1996:
## Summary
This moves the `Rc<RefCell<>>`s into Container, allowing variant tags to be read without borrowing the cells. This also results in a more natural shared data ownership.

## Test Plan
cargo test

### DIFF
--- a/language/ir-testsuite/tests/effects/move_from.mvir
+++ b/language/ir-testsuite/tests/effects/move_from.mvir
@@ -1,0 +1,52 @@
+module M {
+    resource R { x: bool }
+
+    public publish(s: &signer) {
+        move_to<R>(move(s), R { x: false });
+        return;
+    }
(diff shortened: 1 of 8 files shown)


dariorussi:
@sblackshear, @emmazzz this is blocked by the problem Victor talked about in the slack channel

sblackshear:
@vgao1996: did you add a test that will catch the "forgot to mark the resource as dirty when performing vector ops on a vector it contains" problem? I think we probably shouldn't go forward with this until we're protected against that + related problems...

vgao1996:
@sblackshear I'll add a few tests for this. Thanks for pointing this out!

vgao1996:
@sblackshear I've added the tests.

PR #5114, opened 2020-07-15
[crypto][test-only] Improve the malleability tests

huitseeker:
https://github.com/libra/libra/pull/5057 (I) was confused when it (I) stated:
> This check is discussed in [Brendel at al. 5.4 and provides message binding, along with protection against substitution attacks2020]

This present test rather makes clear:
- that ed25519_dalek performs a cofactorless verification equation check, both in its `verify` and `verify_strict` forms,
- that ed25519_dalek, both in `verify` and `verify_strict` forms is still "transformable"¹, absent a prohibitively expensive `.is_torsion_free()` check on the `R` / `A` elements

So when Brendel provides guarantees through a check in section 5.4, they speak of the following check:
> Specifically, Ed25519-LibS ensures that received elements are canonically encoded and have order L

i.e.
https://github.com/jedisct1/libsodium/blob/d1d777532fecc42a2e04f7bca3ec43f4481921c2/src/libsodium/crypto_core/ed25519/core_ed25519.c#L13-L26

which this PR should make clear is not what's applied here (so we correspondingly do not get the guarantees).

¹: in a [transform type 6](http://www.bolet.org/~pornin/2005-acns-pornin+stern.pdf) limited sense.

### DIFF
--- a/crypto/crypto/src/unit_tests/ed25519_test.rs
+++ b/crypto/crypto/src/unit_tests/ed25519_test.rs
@@ -18,19 +18,176 @@ use crate::{
 
 use core::{
     convert::TryFrom,
-    ops::{Index, IndexMut, Neg},
+    ops::{Add, Index, IndexMut, Mul, Neg},
 };
 
+use digest::Digest;
 use libra_crypto_derive::{CryptoHasher, LCSCryptoHash};
 use proptest::{collection::vec, prelude::*};
 use serde::{Deserialize, Serialize};
+use sha2::Sha512;
 
 #[derive(CryptoHasher, LCSCryptoHash, Serialize, Deserialize)]
 struct CryptoHashable(pub usize);
 
+// Takes a point in eight_torsion and finds its order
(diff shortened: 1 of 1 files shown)


huitseeker:
@davidiw The purpose of those tests is to check a different contract boundary. 
- Ed25519 as originally designed and as specified is in fact not one, but a range of algorithms,
- Dalek as written implements a subset of those algorithms,
- We implement an even narrower subset of these algorithms.

The purpose of having those tests in our code base is to alert us in case the choices of Dalek vary (they're also a complement to our spec, and educational).

kchalkias:
I would also point out that nothing prevents us from using another crate on the background or a dalek branch (ie formally verified or optimised). So having these tests is actually a 2nd layer of defence, among the other benefits @huitseeker mentioned.

huitseeker:
Days since we caught a change of semantics in ed25519-dalek : 0
https://github.com/libra/libra/pull/5178

PR #5119, opened 2020-07-15
[libra framework] cleanup in RecoveryAddress

sblackshear:
- Fix vector index OOB bug introduced by https://github.com/libra/libra/pull/4595
- Add some new abort codes
- Add functional tests that exercise the abort codes
- Take KeyRotationCapability as input instead of extracting it
- Add comments + document abort behavior in relevant scripts


language/stdlib/modules/RecoveryAddress.move
```diff
@@ -77,33 +79,33 @@ module RecoveryAddress {
             i = i + 1
         };
         spec {
-            assert i == len + 1;
+            // TODO(shb): Prover thinks this doesn't hold. Is it because of the return in the loop?
+            // How is tis spec block interpreted?
+            //assert i == len;
```
emmazzz, line 84: I think you also need to change line 68 to `assert i <= len;`. Prover relies on the loop invariants to prove line 84.

sblackshear: Thanks, that did it!

sblackshear:
Credit to @vgao1996 for noticing the indexing problem!

PR #5126, opened 2020-07-15
Change to the new cargo feature resolver

metajack:
This PR changes Libra to use the new Cargo feature resolver. The new resolver does not unify features across targets, normal and dev dependencies, or host and target dependencies. This means we can remove all the cruft we introduced to de-unify features with the old resolver and express the true dependencies. For example, after this PR lands, `cargo build -p foo` will work for any package and the same is true for `cargo test` and other commands.

This PR has several parts:

1. Adds in the true dependencies. This mostly includes adding dev-dependencies that use the "fuzzing" feature.
2. Changes to use nightly cargo (the version is pinned in `x.toml`) to access the new resolver before it is officially stabilized. We still use stable Rust defined in `rust-toolchain` for building code.
3. Reworks `cargo x` to remove all the exclusions and run_separately/together logic since none of that is needed any more. This will improve build/test parallelism since we can now batch all packages if we want.

### DIFF
--- a/consensus/src/block_storage/block_store.rs
+++ b/consensus/src/block_storage/block_store.rs
@@ -19,8 +19,6 @@ use executor_types::{Error, StateComputeResult};
 use libra_crypto::HashValue;
 use libra_logger::prelude::*;
 use libra_trace::prelude::*;
-#[cfg(any(test, feature = "fuzzing"))]
-use libra_types::epoch_state::EpochState;
 use libra_types::{ledger_info::LedgerInfoWithSignatures, transaction::TransactionStatus};
 use std::{
     collections::vec_deque::VecDeque,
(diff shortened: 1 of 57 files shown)


metajack:
I'll get this rebased in the morning.

@bors-libra p=-1

consensus/src/block_storage/block_store.rs
```diff
@@ -478,26 +476,4 @@ impl BlockStore {
...
-            Some(EpochState::empty()),
-            compute_result.compute_status().clone(),
-            compute_result.transaction_info_hashes().clone(),
-        );
-        Ok(self
-            .inner
-            .write()
-            .unwrap()
-            .insert_block(ExecutedBlock::new(executed_block.block().clone(), result))?)
-    }
```
huitseeker, line 502: I suspect a rebase would eliminate those changes to consensus?

metajack: No, that code is unused but because it was tied up in the mess of features, the linter never saw it. Now that the dependencies are more correctly expressed, I fixed the resulting linting errors.

metajack:
I updated CONTRIBUTING already for how to express dependencies. I was not planning to have devs install nightly toolchains and do cargo magic, though that will happen automatically if you use `cargo x`. This will mean devs will overbuild, as the will be building release code with fuzzing enabled due to the old resolver being enabled.

metajack:
@sausagee @rexhoffman We need to ensure all production builds use nightly cargo. Is there anywhere else in our release pipeline that I need to change?

rexhoffman:
Cargo lint seems to want the features defined in a feature section....

metajack:
@huitseeker I added in the expanded macro as you suggested. Is it possilbe to just imrpove the macro to avoid this? I see the expansion of serde stuff contains a lot more isolation than our macros do, so maybe we can just copy that?

PR #5127, opened 2020-07-15
[docker] propagate ChainId change to mint

sausagee:
## Motivation
Update mint docker build to be in sync after fbd735248

## Test Plan
Build mint locally

### DIFF
--- a/docker/mint/run.sh
+++ b/docker/mint/run.sh
@@ -11,7 +11,8 @@ set -ex
 IMAGE="${1:-libra_mint:latest}"
 AC_HOST="${2:-172.18.0.10}"
 AC_PORT="${3:-8080}"
-LOG_LEVEL="${4:-info}"
+CHAIN_ID="${4:-TESTING}"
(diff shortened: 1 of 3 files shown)


sausagee:
Need one more PR for the terraform stuff.

docker/mint/run.sh
```diff
@@ -11,14 +11,16 @@ set -ex
 IMAGE="${1:-libra_mint:latest}"
 AC_HOST="${2:-172.18.0.10}"
 AC_PORT="${3:-8080}"
-LOG_LEVEL="${4:-info}"
+CHAIN_ID="${4:-TESTING}"
```
christinacelee, line 14: so this will be different for devnet and testnet, fwiu this script is used for both.
for devnet, should be "DEVNET" or 3, for testnet, should be "TESTNET" or 2

sausagee: Didn't see #5098. I will close this one.  Let's move the conversation there.

PR #5128, opened 2020-07-15
[consensus][rfc] Forcing BlockInfo creation to come from a Block or Execut…

mimoo:
just toying with an idea as:

* execution of blocks is not specified anywhere
* it is hard to trace how different BlockInfo are created (genesis ones, NIL ones, from `ExecutedBlock`s for votes)

### DIFF
--- a/types/src/block_info.rs
+++ b/types/src/block_info.rs
@@ -43,6 +43,8 @@ pub struct BlockInfo {
 }
 
 impl BlockInfo {
+    /// Only use for tests to avoid unclear creation of BlockInfo
+    #[cfg(test)]
     pub fn new(
         epoch: u64,
         round: Round,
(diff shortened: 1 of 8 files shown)


types/src/block_info.rs
```diff
@@ -63,6 +65,52 @@ impl BlockInfo {
...
+            ledger_info.transaction_accumulator_hash(),
+            ledger_info.version(),
+            ledger_info.timestamp_usecs(),
+            next_epoch_state: None,
+        }
+    }
+
+    /// Creates a BlockInfo from a Block structure. 
+    pub fn from_block(
+        block: &Block,
```
davidiw, line 83: the entire intent was block info is public to the client but the block is not, much like block metadata so this has issues :/

types/src/block_info.rs
```diff
@@ -63,6 +65,52 @@ impl BlockInfo {
...
+            executed_state_id,
+            version,
+            block.timestamp_usecs(),
+            next_epoch_state,
+        }
+    }
+
+    /// Creates a BlockInfo from an ExecutedBlock structure. 
+    /// Used to vote on a proposal's block.
+    pub fn from_executed_block(executed_block: &ExecutedBlock) -> Self {
```
davidiw, line 101: same here

types/src/block_info.rs
mimoo: the point of this is that you HAVE to create a BlockInfo from a block, so make this change explicit instead of allowing someone to create one with new() and make a mistake

davidiw:

PR #5130, opened 2020-07-15
[libra framework] Add comments to `create_child_vasp_account`. Add partial specs

tzakian:
Adds comments to the `create_child_vasp_account` along with some partial specs for it.


language/stdlib/modules/LibraAccount.move
```diff
@@ -123,6 +123,7 @@ module LibraAccount {
     /// Attempting to send funds in (e.g.) LBR to an account that exists, but does not have a
     /// Balance<LBR> resource
     const EPAYEE_CANT_ACCEPT_CURRENCY_TYPE: u64 = 18;
+    const EPAYER_DOESNT_HOLD_CURRENCY: u64 = 19;
```
sblackshear, line 126: Nice!

language/stdlib/transaction_scripts/create_child_vasp_account.move
```diff
@@ -29,4 +46,18 @@ fun create_child_vasp_account<CoinType>(
         LibraAccount::restore_withdraw_capability(vasp_withdrawal_cap);
     };
 }
+spec fun create_child_vasp_account {
```
sblackshear, line 49: Add ensures (if it's not too complex)?

tzakian: 😢  Even the simplest of ensures seems to make the prove unhappy. I think once we have more specs in LibraAccount we'll be able to add some.

PR #5133, opened 2020-07-16
[tf] add ChainId

sausagee:
## Motivation
Propagate ChainId change to terraform config.

## Test Plan
Testing on own cluster.


sausagee:
Added b6d5f6c to correct a typo missed in #5098.

PR #5135, opened 2020-07-16
[deps] dependabot config switch to live (from nightly) and security updates only

rexhoffman:
## Motivation

As we are now in a mode where the master branch is to be treated as a release branch we should now disable updating dependency on the master branch unless they are security fixes.   Which will now be opened as soon as they are available rather than being batched nightly.

It may be a tad premature as we still have several months before go live.   I'd be okay with waiting until we actually cut the branch but want the call able to made now.

Y

## Test Plan

Dependabot's handy-dandy config validator:  https://dependabot.com/docs/config-file/validator/

## Related PRs

None

### DIFF
--- a/.dependabot/config.yml
+++ b/.dependabot/config.yml
@@ -2,12 +2,14 @@ version: 1
 update_configs:
   - package_manager: "rust:cargo"
     directory: "/"
-    update_schedule: "daily"
+    update_schedule: "live"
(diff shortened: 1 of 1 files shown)


rexhoffman:
@sausagee yes dependabot supports multiple branches, I'll configure it for security updates only on release branches when the time comes.

PR #5136, opened 2020-07-16
[json-rpc] rename json-rpc method get_account_state to get_account

xli:
## Motivation

get_account_state returns a struct named AccountView on the server side, it contains account resources data as field values. The name account_state is widely used as internal data structure AccountState, which is a btree map.
For non-verifying API, we will never return the btree map AccountState, but for verifying client API (will be added in future, or existing experimental method get_account_state_with_proof), we will return the btree map AccountState.
Using the name get_account_state to return data structure that is not the btree map AccountState is confusing, especially when we publish verifying client APIs.

This change renames get_account_state method name to get_account to reflect common meaning of Account struct, it also aligns with the return value struct AccountView on server side.

Also, this change will keep cli query account_state command as if we change it, need to update all published blogs/docs which are out of this codebase.

Yes

## Test Plan

1. Unit test
2. Manual local test:
* cargo run -p libra-swarm -- -s -n 4
* account create
* account mintb 0 2 LBR false
* query balance 0
* query account_state 0

## Related PRs

### DIFF
--- a/json-rpc/src/tests/unit_tests.rs
+++ b/json-rpc/src/tests/unit_tests.rs
@@ -164,13 +164,15 @@ fn test_json_rpc_protocol() {
     assert_eq!(fetch_error(resp), -32601);
 
     // invalid arguments
-    let request = serde_json::json!({"jsonrpc": "2.0", "method": "get_account_state", "params": [1, 2], "id": 1});
+    let request =
+        serde_json::json!({"jsonrpc": "2.0", "method": "get_account", "params": [1, 2], "id": 1});
     let resp = client.post(&url).json(&request).send().unwrap();
     assert_eq!(resp.status(), 200);
     assert_eq!(fetch_error(resp), -32000);
 
     // Response includes two mandatory field, regardless of errors
-    let request = serde_json::json!({"jsonrpc": "2.0", "method": "get_account_state", "params": [1, 2], "id": 1});
+    let request =
+        serde_json::json!({"jsonrpc": "2.0", "method": "get_account", "params": [1, 2], "id": 1});
     let resp = client.post(&url).json(&request).send().unwrap();
     assert_eq!(resp.status(), 200);
     let data: JsonMap = resp.json().unwrap();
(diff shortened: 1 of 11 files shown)


json-rpc/src/tests/unit_tests.rs
```diff
@@ -240,15 +242,15 @@ fn test_transaction_submission() {
 
 // TODO: Once account configs are published in the mock DB this test can be turned back on
```
sunmilee, line 243: Can this be uncommented?

christinacelee: this test requires mocking the on-chain currency resources, which is a much larger task.. for now we can just manually curl

christinacelee:
can we also test the new endpoint with a local curl? (so curling against a local Libra swarm)

xli:
> can we also test the new endpoint with a local curl? (so curling against a local Libra swarm)

did the test with local libra swarm

xli:
Local test console log (trimmed most of unrelated output):
```

Connected to validator at: http://localhost:54135, latest version = 5, timestamp = 2020-07-16 01:57:51.933954 UTC
usage: <command> <args>

......

libra% a create
>> Creating/retrieving next account from wallet
INFO 2020-07-16 01:58:36 testsuite/cli/src/libra_client.rs:268 Verified version change to: 205
Created/retrieved account #0 address d0935c7f74e6ac1d2976b425ed84ab05

libra% a mb 0 2 LBR false
>> Minting coins
waiting for 0000000000000000000000000a550c18 with sequence number 2

......

Finished minting!

libra% q b 0
Balance is: 2.000000LBR
libra% q as 0
>> Getting latest account state
Latest account state is:
 Account: (
    d0935c7f74e6ac1d2976b425ed84ab05,
    Some(
        AuthenticationKey(......),
    ),
)
 State: Some(
    AccountView {
        balances: [
            AmountView {
                amount: 2000000,
                currency: "LBR",
            },
        ],
        sequence_number: 0,
        ......
    },
)
 Blockchain Version: 564

```

PR #5138, opened 2020-07-16
[language] Restriction on arguments passed to the VM

dariorussi:
We are restricting the runtime to take only constants as arguments.
A reference to the Signer too, actually but that is destined to
end at some brighter future

### DIFF
--- a/language/bytecode-verifier/src/verifier.rs
+++ b/language/bytecode-verifier/src/verifier.rs
@@ -56,31 +56,10 @@ fn verify_main_signature_impl(script: &CompiledScript) -> PartialVMResult<()> {
     fn is_valid_arg_type(idx: usize, arg_type: &SignatureToken) -> bool {
         use SignatureToken as S;
         match arg_type {
-            S::Bool | S::U8 | S::U64 | S::U128 | S::Address => true,
-            S::Vector(inner) => match &**inner {
-                S::U8 => true,
-                S::Bool
-                | S::U64
-                | S::U128
-                | S::Address
-                | S::Signer
-                | S::Struct(_)
-                | S::Vector(_)
-                | S::StructInstantiation(_, _)
-                | S::Reference(_)
-                | S::MutableReference(_)
-                | S::TypeParameter(_) => false,
-            },
-
             // &signer is a type that can only be populated by the Move VM. And its value is filled
             // based on the sender of the transaction
             S::Reference(inner) => idx == 0 && matches!(&**inner, S::Signer),
-
-            S::Signer
-            | S::Struct(_)
-            | S::StructInstantiation(_, _)
-            | S::MutableReference(_)
-            | S::TypeParameter(_) => false,
+            _ => arg_type.is_constant(),
         }
     }
 
--- a/language/move-vm/runtime/src/interpreter.rs
+++ b/language/move-vm/runtime/src/interpreter.rs
@@ -1097,9 +1097,8 @@ impl Frame {
 }
 
 // Verify the the type of the arguments in input from the outside is restricted (`is_valid_arg()`)
-// TODO: we need to check the instantiation
-// TODO: specify the values allowed, we could try to match everything, this function should have
-// minimal or non existent policy (besides any type/value match) but it's tricky
+// and it honors the signature of the function invoked.
+// TODO: we need to check the instantiation, once we expose signatures with generic argument
 fn verify_args(signature: &Signature, _ty_args: &[Type], args: &[Value]) -> PartialVMResult<()> {
     if signature.len() != args.len() {
         return Err(
--- a/language/move-vm/runtime/src/unit_tests/arguments.rs
+++ b/language/move-vm/runtime/src/unit_tests/arguments.rs
@@ -1,0 +1,439 @@
+// Copyright (c) The Libra Core Contributors
+// SPDX-License-Identifier: Apache-2.0
+
+use crate::{data_cache::RemoteCache, move_vm::MoveVM};
+use libra_types::vm_status::StatusCode;
+use move_core_types::{
+    account_address::AccountAddress,
+    gas_schedule::{GasAlgebra, GasUnits},
+    identifier::Identifier,
+    language_storage::{ModuleId, TypeTag},
+};
+use move_vm_types::{
+    gas_schedule::{zero_cost_schedule, CostStrategy},
+    values::Value,
+};
+use vm::{
+    errors::{PartialVMResult, VMResult},
+    file_format::{
+        AddressIdentifierIndex, Bytecode, CodeUnit, CompiledScriptMut, FunctionHandle,
+        FunctionHandleIndex, IdentifierIndex, ModuleHandle, ModuleHandleIndex, Signature,
+        SignatureIndex, SignatureToken, StructHandle, StructHandleIndex,
+    },
+};
+
+fn make_script(signature: Signature) -> Vec<u8> {
+    let mut blob = vec![];
+    CompiledScriptMut {
+        module_handles: vec![],
+        struct_handles: vec![],
+        function_handles: vec![],
+
+        function_instantiations: vec![],
+
+        signatures: vec![Signature(vec![]), signature],
+
+        identifiers: vec![],
+        address_identifiers: vec![],
+        constant_pool: vec![],
+
+        type_parameters: vec![],
+        parameters: SignatureIndex(1),
+        code: CodeUnit {
+            locals: SignatureIndex(0),
+            code: vec![Bytecode::Ret],
+        },
+    }
+    .serialize(&mut blob)
+    .expect("script must serialize");
+    blob
+}
+
+fn make_script_with_imports(signature: Signature) -> Vec<u8> {
+    let mut blob = vec![];
+    CompiledScriptMut {
+        module_handles: vec![ModuleHandle {
+            address: AddressIdentifierIndex(0),
+            name: IdentifierIndex(0),
+        }],
+        struct_handles: vec![StructHandle {
+            module: ModuleHandleIndex(0),
+            name: IdentifierIndex(1),
+            is_nominal_resource: false,
+            type_parameters: vec![],
+        }],
+        function_handles: vec![],
+
+        function_instantiations: vec![],
+
+        signatures: vec![Signature(vec![]), signature],
+
+        identifiers: vec![
+            Identifier::new("one").unwrap(),
+            Identifier::new("two").unwrap(),
+        ],
+        address_identifiers: vec![AccountAddress::random()],
+        constant_pool: vec![],
+
+        type_parameters: vec![],
+        parameters: SignatureIndex(1),
+        code: CodeUnit {
+            locals: SignatureIndex(0),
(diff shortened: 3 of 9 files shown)


dariorussi:
there are more tests coming but this should be it

language/bytecode-verifier/src/verifier.rs
```diff
@@ -58,18 +58,9 @@ fn verify_main_signature_impl(script: &CompiledScript) -> PartialVMResult<()> {
...
-                | S::Address
-                | S::Signer
-                | S::Struct(_)
-                | S::Vector(_)
-                | S::StructInstantiation(_, _)
-                | S::Reference(_)
-                | S::MutableReference(_)
-                | S::TypeParameter(_) => false,
+                S::Bool | S::U8 | S::U64 | S::U128 | S::Address => true,
+                S::Vector(vec_type) => is_valid_arg_type(idx, vec_type),
```
sblackshear, line 62: Would this accept `vector<&signer>`, `vector<vector<&signer>>`, etc  (which we don't want)? Looks like it might, but I can't tell... Might  be worth a test.

language/move-vm/runtime/src/unit_tests/arguments.rs
```diff
@@ -0,0 +1,119 @@
...
+    )
+    .expect("good vector<address> failed");
+
+    // bad vector of address
+    let script = make_script_with_sig(Signature(vec![SignatureToken::Vector(Box::new(
+        SignatureToken::Address,
+    ))]));
+    call_script(script, vec![Value::vector_bool(vec![true])]).expect_err("bad vector<address>");
+
+    // bad vector of vector
```
sblackshear, line 113: Is "good vector of vector" one of the tests coming later?

language/bytecode-verifier/src/verifier.rs
dariorussi: absolutely correct! nice!

language/move-vm/runtime/src/unit_tests/arguments.rs
dariorussi: yeah there is more test coming as I said, but I need to be able to build those and the runtime is not that nice towards that, so I stopped.
To be clear, thanks for the approval but this was not going in until tomorrow when I get time to add all the tests.
Having said so, the verifier mistake was there so that is great!

dariorussi:
more tests and fixes

language/bytecode-verifier/src/verifier.rs
```diff
@@ -56,31 +56,23 @@ fn verify_main_signature_impl(script: &CompiledScript) -> PartialVMResult<()> {
...
-            S::Signer
-            | S::Struct(_)
-            | S::StructInstantiation(_, _)
-            | S::MutableReference(_)
-            | S::TypeParameter(_) => false,
+    fn is_constant_arg_type(idx: usize, arg_type: &SignatureToken) -> bool {
+        use SignatureToken as S;
+        match arg_type {
+            S::Bool | S::U8 | S::U64 | S::U128 | S::Address => true,
+            S::Vector(inner) => match &**inner {
```
bob-wilson, line 70: Two comments:
- I don't see anything using the idx argument. Can you drop that?
- Can't you write the inner match as the recursive call? E.G., S::Vector(inner) => is_constant_arg_type(&**inner)

language/move-vm/runtime/src/interpreter.rs
```diff
@@ -1095,9 +1095,8 @@ impl Frame {
 }
 
 // Verify the the type of the arguments in input from the outside is restricted (`is_valid_arg()`)
-// TODO: we need to check the instantiation
-// TODO: specify the values allowed, we could try to match everything, this function should have
-// minimal or non existent policy (besides any type/value match) but it's tricky
+// and it honors the siganture of the function invoked.
```
bob-wilson, line 1098: typo: "signature"

language/move-vm/types/src/values/values_impl.rs
```diff
@@ -302,33 +286,104 @@ impl Value {
...
+
+    // Walks a `SignatureToken` to verify that represents a constant
+    fn is_signature_constant(sig: &SignatureToken) -> bool {
+        match sig {
+            SignatureToken::U8
+            | SignatureToken::U64
+            | SignatureToken::U128
+            | SignatureToken::Bool
+            | SignatureToken::Address => true,
+            SignatureToken::Vector(ty) => match &**ty {
```
bob-wilson, line 348: Same thing here: Recurse earlier so you don't have to repeat the match

language/bytecode-verifier/src/verifier.rs
vgao1996: Hmm isn't the script already verified at the moment? `vector<&...>` isn't really a structurally valid type.

language/bytecode-verifier/src/verifier.rs
```diff
@@ -56,31 +56,23 @@ fn verify_main_signature_impl(script: &CompiledScript) -> PartialVMResult<()> {
...
+                S::Bool | S::U8 | S::U64 | S::U128 | S::Address => true,
+                S::Vector(vec_type) => is_constant_arg_type(idx, vec_type),
+                _ => false,
+            },
+            _ => false,
```
tnowacki, line 75: Nit full pattern match. Otherwise adding new things to constants is going to be confusing

language/bytecode-verifier/src/verifier.rs
tnowacki: Agreed with @bob-wilson. I think you can just do a simple recursive call

language/bytecode-verifier/src/verifier.rs
```diff
@@ -56,31 +56,23 @@ fn verify_main_signature_impl(script: &CompiledScript) -> PartialVMResult<()> {
...
+            _ => is_constant_arg_type(idx, arg_type),
+        }
+    }
 
-            S::Signer
-            | S::Struct(_)
-            | S::StructInstantiation(_, _)
-            | S::MutableReference(_)
-            | S::TypeParameter(_) => false,
+    fn is_constant_arg_type(idx: usize, arg_type: &SignatureToken) -> bool {
```
tnowacki, line 66: We have this logic already in constants.rs. Should we reuse that?

language/move-vm/types/src/values/values_impl.rs
```diff
@@ -302,33 +286,104 @@ impl Value {
...
+                SignatureToken::Bool
+                | SignatureToken::U8
+                | SignatureToken::U64
+                | SignatureToken::U128
+                | SignatureToken::Address => true,
+                SignatureToken::Vector(inner) => Self::is_signature_constant(&*inner),
+                _ => false,
+            },
+            _ => false,
```
tnowacki, line 357: nit exhaustive match

language/move-vm/types/src/values/values_impl.rs
tnowacki: +1

language/move-vm/types/src/values/values_impl.rs
```diff
@@ -302,33 +286,104 @@ impl Value {
...
+                        }
+                    }
+                    true
                 }
-            }
-            ValueImpl::ContainerRef(ContainerRef::Local(Container::StructR(r))) => {
-                let v = &*r.borrow();
-                v.len() == 1 && matches!(&v[0], ValueImpl::Address(_))
-            }
+                _ => false,
```
tnowacki, line 385: Nit full pattern match.

language/move-vm/types/src/values/values_impl.rs
```diff
@@ -302,33 +286,104 @@ impl Value {
...
+                    true
+                }
+                _ => false,
+            },
+            _ => false,
+        }
+    }
+
+    // Walks a `SignatureToken` to verify that represents a constant
+    fn is_signature_constant(sig: &SignatureToken) -> bool {
```
tnowacki, line 341: This logic is now duplicated in 3 places. We should just have 1 in my opinion. In my view, all of these functions *need* to line up properly

Thoughts?

language/bytecode-verifier/src/verifier.rs
dariorussi: yes true! and I read what @sblackshear said as vector<signer> which would have been a problem.
So ignore the reference and that is what happened ;)

language/bytecode-verifier/src/verifier.rs
dariorussi: I'll check it, thanks!

language/move-vm/types/src/values/values_impl.rs
dariorussi: let me check

language/bytecode-verifier/src/verifier.rs
dariorussi: though maybe that was ok too.... I'll look again as I am trying to reconcile all the functions related to constants

dariorussi:
ok take another look if you can...

language/bytecode-verifier/src/verifier.rs
dariorussi: still have to do that and we'll look soon. Whether I like the idea or not I don't know yet :)

dariorussi:
if we take out the test and comments is around the same amount of code that was there before (~70LOC).
It should be better than before, so take a look and let me know

language/move-vm/types/src/values/values_impl.rs
dariorussi: I'll try and see if I like it :)

dariorussi:
added one more test. If I don't get much feedback soon I'll go in later today.
Let me know...

language/move-vm/types/src/values/values_impl.rs
dariorussi: I think it should be ok now, thanks for the feedback, I was going all over the show....

language/vm/src/file_format.rs
```diff
@@ -739,6 +739,23 @@ impl SignatureToken {
         matches!(self, MutableReference(_))
     }
 
+    /// Returns true if the `SignatureToken` can represent a constant (as in representable in
+    /// the constants table).
+    pub fn is_constant(&self) -> bool {
```
tnowacki, line 744: I find the wording a bit confusing but not a big deal. maybe `is_valid_for_constant` `valid_constant_type`? 

Just that its not about the type itself being a "constant", which is a term that comes up some times in type systems

language/move-vm/runtime/src/unit_tests/arguments.rs
```diff
@@ -0,0 +1,439 @@
...
+            AccountAddress::random(),
+            AccountAddress::random(),
+            AccountAddress::random(),
+            AccountAddress::random(),
+        ])],
+    )
+    .expect("good vector<address> failed");
+    // wrong vector
+    call_script(script.clone(), vec![Value::vector_bool(vec![true])])
+        .expect_err("bad vector<address>");
```
tnowacki, line 410: Should we be matching on the status here for these? or do we not care as long as it failed?

PR #5139, opened 2020-07-16
Update swiss-knife examples with ChainId

ankushagarwal:
## Summary

Update swiss-knife examples with ChainId


client/swiss-knife/sample_inputs/generate_raw_txn_peer_to_peer_transfer.json
```diff
@@ -5,6 +5,7 @@
     "max_gas_amount": 1000000,
     "gas_unit_price": 0,
     "gas_currency_code": "LBR",
+    "chain_id": "TESTING",
```
christinacelee, line 8: just fyi this also works with "4" (numeric representation of chain ID as a string), maybe it'll be helpful for one of the examples to use this format to make users aware of this representation too (bc most 3rd party chains actually won't have the human-readable string repr)? (completely your call!)

ankushagarwal: Didn't update the json example, but added a note in the readme about the alternative numeric representation for chain_id

PR #5140, opened 2020-07-16
[secure storage] bug in vault use of signing with a version

davidiw:
smh :/


davidiw:
Actually this was detected by unit tests, I don't know how it ... made it this long, probably should look into it so we can harden our process. We should also reach out to @rexhoffman and @sausagee to get vault integration tests into land blocking.

@bors-libra r=joshlind

PR #5143, opened 2020-07-16
Secure net config timeout

agclement:
## Motivation

Plumb secure/net timeouts through to the relevant configs.

yes

## Test Plan

Don't break existing tests.

## Related PRs

### DIFF
--- a/config/src/config/execution_config.rs
+++ b/config/src/config/execution_config.rs
@@ -22,6 +22,7 @@ pub struct ExecutionConfig {
     pub genesis_file_location: PathBuf,
     pub service: ExecutionCorrectnessService,
     pub backend: SecureBackend,
+    pub network_timeout_ms: u64,
 }
 
 impl std::fmt::Debug for ExecutionConfig {
@@ -54,6 +55,8 @@ impl Default for ExecutionConfig {
             service: ExecutionCorrectnessService::Thread,
             backend: SecureBackend::InMemoryStorage,
             sign_vote_proposal: true,
+            // Default value of 30 seconds for the network timeout.
+            network_timeout_ms: 30_000,
         }
     }
 }
--- a/config/src/config/safety_rules_config.rs
+++ b/config/src/config/safety_rules_config.rs
@@ -23,6 +23,8 @@ pub struct SafetyRulesConfig {
     pub service: SafetyRulesService,
     pub test: Option<SafetyRulesTestConfig>,
     pub verify_vote_proposal_signature: bool,
+    // Read/Write/Connect networking operation timeout in milliseconds.
+    pub network_timeout_ms: u64,
 }
 
 impl Default for SafetyRulesConfig {
@@ -33,6 +35,8 @@ impl Default for SafetyRulesConfig {
             service: SafetyRulesService::Thread,
             test: None,
             verify_vote_proposal_signature: true,
+            // Default value of 30seconds for a timeout
+            network_timeout_ms: 30_000,
         }
     }
 }
--- a/consensus/safety-rules/src/process.rs
+++ b/consensus/safety-rules/src/process.rs
@@ -31,6 +31,7 @@ impl Process {
                 server_addr,
                 storage,
                 verify_vote_proposal_signature,
+                network_timeout: config.network_timeout_ms,
             }),
         }
     }
@@ -41,6 +42,7 @@ impl Process {
             data.storage,
             data.server_addr,
             data.verify_vote_proposal_signature,
+            data.network_timeout,
         );
     }
 }
@@ -49,20 +51,30 @@ struct ProcessData {
     server_addr: SocketAddr,
     storage: PersistentSafetyStorage,
     verify_vote_proposal_signature: bool,
+    // Timeout in Seconds for network operations
+    network_timeout: u64,
 }
 
 pub struct ProcessService {
     server_addr: SocketAddr,
+    network_timeout_ms: u64,
 }
 
 impl ProcessService {
-    pub fn new(server_addr: SocketAddr) -> Self {
-        Self { server_addr }
+    pub fn new(server_addr: SocketAddr, network_timeout: u64) -> Self {
+        Self {
+            server_addr,
+            network_timeout_ms: network_timeout,
+        }
     }
 }
 
 impl RemoteService for ProcessService {
     fn server_address(&self) -> SocketAddr {
         self.server_addr
+    }
+
+    fn network_timeout_ms(&self) -> u64 {
+        self.network_timeout_ms
     }
 }
--- a/consensus/safety-rules/src/safety_rules_manager.rs
+++ b/consensus/safety-rules/src/safety_rules_manager.rs
@@ -67,7 +67,9 @@ pub struct SafetyRulesManager {
 impl SafetyRulesManager {
     pub fn new(config: &SafetyRulesConfig) -> Self {
         match &config.service {
-            SafetyRulesService::Process(conf) => return Self::new_process(conf.server_address()),
+            SafetyRulesService::Process(conf) => {
+                return Self::new_process(conf.server_address(), config.network_timeout_ms)
+            }
             SafetyRulesService::SpawnedProcess(_) => return Self::new_spawned_process(config),
             _ => (),
         };
@@ -79,7 +81,11 @@ impl SafetyRulesManager {
             SafetyRulesService::Serializer => {
                 Self::new_serializer(storage, verify_vote_proposal_signature)
             }
-            SafetyRulesService::Thread => Self::new_thread(storage, verify_vote_proposal_signature),
+            SafetyRulesService::Thread => Self::new_thread(
+                storage,
+                verify_vote_proposal_signature,
+                config.network_timeout_ms,
+            ),
             _ => panic!("Unimplemented SafetyRulesService: {:?}", config.service),
         }
     }
@@ -94,8 +100,8 @@ impl SafetyRulesManager {
         }
     }
 
-    pub fn new_process(server_addr: SocketAddr) -> Self {
-        let process_service = ProcessService::new(server_addr);
+    pub fn new_process(server_addr: SocketAddr, timeout_ms: u64) -> Self {
+        let process_service = ProcessService::new(server_addr, timeout_ms);
         Self {
             internal_safety_rules: SafetyRulesWrapper::Process(process_service),
         }
@@ -124,8 +130,9 @@ impl SafetyRulesManager {
     pub fn new_thread(
         storage: PersistentSafetyStorage,
         verify_vote_proposal_signature: bool,
+        timeout_ms: u64,
     ) -> Self {
-        let thread = ThreadService::new(storage, verify_vote_proposal_signature);
+        let thread = ThreadService::new(storage, verify_vote_proposal_signature, timeout_ms);
         Self {
             internal_safety_rules: SafetyRulesWrapper::Thread(thread),
         }
--- a/consensus/safety-rules/src/tests/networking.rs
+++ b/consensus/safety-rules/src/tests/networking.rs
@@ -8,7 +8,9 @@ use libra_types::validator_signer::ValidatorSigner;
 fn test_reconnect() {
     let signer = ValidatorSigner::from_int(0);
     let storage = test_utils::test_storage(&signer);
-    let safety_rules_manager = SafetyRulesManager::new_thread(storage, false);
+    // test value for network timeout, in milliseconds.
+    let network_timeout = 5_000;
+    let safety_rules_manager = SafetyRulesManager::new_thread(storage, false, network_timeout);
 
     // Verify that after a client has disconnected a new client will connect and resume operations
     let state0 = safety_rules_manager.client().consensus_state().unwrap();
--- a/secure/net/src/lib.rs
+++ b/secure/net/src/lib.rs
@@ -25,9 +25,9 @@ use thiserror::Error;
 
 /// Read, Write, Connect timeout in seconds.
 #[cfg(not(test))]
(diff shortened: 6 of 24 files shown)


agclement:
@davidiw timeouts for secure/net all the way out to the invoking configs.

config/src/config/safety_rules_config.rs
```diff
@@ -23,6 +23,8 @@ pub struct SafetyRulesConfig {
     pub service: SafetyRulesService,
     pub test: Option<SafetyRulesTestConfig>,
     pub verify_vote_proposal_signature: bool,
+    // Read/Write/Connect networking operation timeout in seconds.
+    pub timeout: u64,
```
davidiw, line 27: network_timeout?
also should we just add this to the remote service since it is only really required for that?

consensus/safety-rules/src/process.rs
```diff
@@ -32,6 +34,7 @@ impl Process {
                 storage,
                 verify_vote_proposal_signature,
             }),
+            timeout: config.timeout,
```
davidiw, line 37: why not make this part of ProcessData too? This creates an interesting asymmetry.

consensus/safety-rules/src/safety_rules_manager.rs
```diff
@@ -80,7 +80,9 @@ pub struct SafetyRulesManager {
 impl SafetyRulesManager {
     pub fn new(config: &mut SafetyRulesConfig) -> Self {
         match &config.service {
-            SafetyRulesService::Process(conf) => return Self::new_process(conf.server_address()),
+            SafetyRulesService::Process(conf) => {
+                return Self::new_process(conf.server_address(), config.timeout)
+            }
```
davidiw, line 85: we could potentially just pass in the config if it were in remote service

consensus/safety-rules/src/safety_rules_manager.rs
```diff
@@ -92,7 +94,9 @@ impl SafetyRulesManager {
             SafetyRulesService::Serializer => {
                 Self::new_serializer(storage, verify_vote_proposal_signature)
             }
-            SafetyRulesService::Thread => Self::new_thread(storage, verify_vote_proposal_signature),
+            SafetyRulesService::Thread => {
+                Self::new_thread(storage, verify_vote_proposal_signature, config.timeout)
```
davidiw, line 98: I'd argue if thread ever timed out, we'd have some serious issues :P, but I can see one argument for moving the value out of RemoteService or making thread have a RemoteService that is just ignores except for the timeout.

secure/net/src/lib.rs
```diff
@@ -25,9 +25,9 @@ use thiserror::Error;
 
 /// Read, Write, Connect timeout in seconds.
 #[cfg(not(test))]
-const TIMEOUT: u64 = 30;
+pub const TIMEOUT: u64 = 30;
 #[cfg(test)]
-const TIMEOUT: u64 = 5;
+pub const TIMEOUT: u64 = 5;
```
davidiw, line 30: actually, now with this, we should probably move this value into the the test module below

storage/storage-service/src/storage_service_test.rs
```diff
@@ -20,11 +20,13 @@ fn start_test_storage_with_client() -> (JoinHandle<()>, libra_temppath::TempPath
 
     let server_port = utils::get_available_port();
     config.storage.address = SocketAddr::new(IpAddr::V4(Ipv4Addr::LOCALHOST), server_port);
+    // Test timeout of 5 secons
```
davidiw, line 23: seconds

config/src/config/safety_rules_config.rs
agclement: RemoteService is a trait implemented by various instantiations of the SafetyRules object. 

We are able to add the pub network_timeout() -> u64 function to remote service, but not the field itself.

consensus/safety-rules/src/process.rs
agclement: ProcessData is the result of invoking execute;  the invocation of execute relies on network_timeout to produce the ProcessData

consensus/safety-rules/src/safety_rules_manager.rs
agclement: RemoteService is a trait and not a struct which complicates your suggestion significantly.

consensus/safety-rules/src/safety_rules_manager.rs
agclement: I don't disagree.  Unfortunately, `Thread` invokes `RemoteService::execute` which is a static function and not a struct function.  And `RemoteService::execute` requires a timeout because it is also used by the potentially network-able SafetyRules.

This... behaviour... is a more or less direct result of the c-style static functions that are very dominant in all of the non-network crates as far as I can tell.

secure/net/src/lib.rs
agclement: Simply removing it.

agclement:
@davidiw Thanks for the feedback and comments I've done my best to address them and discovered that I hadn't quite finished pushing the network_timeout all the way into execution config.

Structurally my belief is that we're parsing the various config objects at the wrong place and passing the raw values (i.e., execution.network_timeout) around too frequently.  This is stemming from two issues:

1). an overuse of c style static functions which limit our ability to re-use `self.` values
2). Not employing some specific disciplines/patterns with configs, namely that 
     a.  configs are always moved OR cloned into the receiving struct,
     b.  Upon receipt of a config, the struct either passes the config through without peaking inside XOR consumes all fields of the config either directly or by passing them on to child structs.
3).  Defining the configs in a central/global location rather than locally with the consuming struct;  the global config can then be the root of the tree that specifies the various child configs.

config/src/config/safety_rules_config.rs
davidiw: I was referring to the https://github.com/libra/libra/blob/master/config/src/config/safety_rules_config.rs#L68

consensus/safety-rules/src/process.rs
davidiw: I don't follow, the exact same code path is used by data.server_addr. It does look pretty awkward for the data for NetworkService to come from two different sources.

consensus/safety-rules/src/safety_rules_manager.rs
davidiw: Overloaded terminology :)

davidiw:
You'll probably be happy with this PR: https://github.com/libra/libra/pull/5185

My only real issue is the ProcessData <=> network_timeout, the asymmetry is worth beating on.

consensus/safety-rules/src/process.rs
agclement: Ahh, now that I see the RemoteService(Config) in addition to the trait RemoteService, the comment is much more clear.

I agree with that awkward asymmetry, but it rears its head again when Thread invokes RemoteService::execute and does not have access to a RemoteService(Config).

The bigger issue blocking doing this in a clean fashion lies in RemoteService::execute being applied to structs which do not implement the RemoteService trait.

static functions considered evil. ;)

config/src/config/safety_rules_config.rs
agclement: Ahh, now that I see the RemoteService(Config) in addition to the trait RemoteService, the comment is much more clear.

I agree with that awkward asymmetry, but it rears its head again when Thread invokes RemoteService::execute and does not have access to a RemoteService(Config).

The bigger issue blocking doing this in a clean fashion lies in RemoteService::execute being applied to structs which do not implement the RemoteService trait.

static functions considered evil. ;)

agclement:
> You'll probably be happy with this PR: #5185

(party time!)

> 
> My only real issue is the ProcessData <=> network_timeout, the asymmetry is worth beating on.

Dug into it again and you're right.  I was reading a much higher connection between Process and ProcesService into the code than what actually exists, and I didn't properly grok that Process is used *only* in https://github.com/libra/libra/blob/master/consensus/safety-rules/src/main.rs

consensus/safety-rules/src/process.rs
```diff
@@ -49,20 +51,29 @@ struct ProcessData {
...
+            network_timeout,
+        }
     }
 }
 
 impl RemoteService for ProcessService {
     fn server_address(&self) -> SocketAddr {
         self.server_addr
     }
+    fn network_timeout(&self) -> u64 {
```
davidiw, line 76: I wish this was a lint..... new line.

davidiw:
@bors-libra r+

cc: @mgorven configurable timeouts for safety rules / lec

consensus/safety-rules/src/process.rs
agclement: It ought to be if it's not.  (and it's not locally).

BTW, I observe two things right now:

1). `cargo xclippy` seems to have been removed?
2) `cargo xtest` seems to have been removed?
3). I can no longer run circles integration tests (which may explain not hitting the ought to exist lint rule,  if it does in fact exist)

agclement:
> Thanks!
> 
> We also need timeouts for vault :O and preferably configurable. I suspect that one is a bit easier since we can shove that into the Vault config

will do that next.

config/src/config/execution_config.rs
```diff
@@ -54,6 +55,8 @@ impl Default for ExecutionConfig {
             service: ExecutionCorrectnessService::Thread,
             backend: SecureBackend::InMemoryStorage,
             sign_vote_proposal: true,
+            // Default value of 30 seconds for the network timeout.
+            network_timeout: 30,
```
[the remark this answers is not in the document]
mgorven: Yup, I realise this timeout covers multiple aspects and that's fine for now. I'd just like to avoid the case where we later add a `timeout_connect` option in ms, and then we have some timeouts in s and some in ms.

agclement:
@davidiw Time to try the re-approve button press again.

config/src/config/execution_config.rs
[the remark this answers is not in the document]
agclement: @mgorven I'll make sure to get your thoughts on adding timeout to vault.  Better to get it close the first time from the handles that you'll need.

davidiw:
For future PRs, let's try to merge these type of commits in: [secure] Address Reviewer Comments
rather than having them visible.

config/src/config/execution_config.rs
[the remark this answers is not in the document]
davidiw: This value seems problematic, shouldn't it be 30_000?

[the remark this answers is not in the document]
agclement: Caught it.  Fixed all the other instances of 30 except the one that Michael actually commented on.

agclement:
> @davidiw Time to try the re-approve button press again.

And again.

consensus/safety-rules/src/process.rs
gregnazario: As of a couple days ago, these have been changed.

You should be able to do  `cargo xlint`, `cargo clippy` instead as well as `cargo test`, but you'll probably have to do `cargo test -p <package> --all-features` or enable the testing features other ways.

Clippy might not be quite as smart without the test features enabled.  Hopefully, these will be fixed in the future to work again :)

rexhoffman:
So @agclement the reason we aren't seeing circleci checks here is supposed because you enabled circleci on your fork
https://support.circleci.com/hc/en-us/articles/360008097173-Why-aren-t-pull-requests-triggering-jobs-on-my-organization-

agclement:
> So @agclement the reason we aren't seeing circleci checks here is supposedly because you enabled circleci on your fork
> https://support.circleci.com/hc/en-us/articles/360008097173-Why-aren-t-pull-requests-triggering-jobs-on-my-organization-
> 
> You'll have to disable circle ci, close this pr, and open a new one.

Thanks @rexhoffman for getting to the bottom of that.

For posterity --- on the circleci 'projects' page I clicked on "unfollow project" `libra` and that appears to have unblocked circleci.     It was not necessary to close/reopen the PR.

I'm not exactly sure where in the "setup circleci" process I ended up adding libra as a project to be followed, but I apparently clicked a yes when I should have clicked a no.

consensus/safety-rules/src/tests/networking.rs
```diff
@@ -8,7 +8,9 @@ use libra_types::validator_signer::ValidatorSigner;
 fn test_reconnect() {
     let signer = ValidatorSigner::from_int(0);
     let storage = test_utils::test_storage(&signer);
-    let safety_rules_manager = SafetyRulesManager::new_thread(storage, false);
+    // test value for network timeout, in milliseconds.
+    let network_timeout = 5_000;
+    let safety_rules_manager = SafetyRulesManager::new_thread(storage, false, network_timeout);
```
gregnazario, line 13: Nit: If this is testing the timeout 1 second might be better to keep the tests rolling.

gregnazario:
@agclement If you could squash `[secure] Address Reviewer Comments` please :)

agclement:
> @agclement If you could squash `[secure] Address Reviewer Comments` please :)

@gregnazario done

secure/net/src/lib.rs
```diff
@@ -46,13 +40,16 @@ pub enum Error {
 pub struct NetworkClient {
     server: SocketAddr,
     stream: Option<NetworkStream>,
+    /// Read, Write, Connect timeout in milliseconds.
+    timeout: u64,
```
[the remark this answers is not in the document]
agclement: Done

consensus/safety-rules/src/tests/networking.rs
agclement: Legacy test timeout value was 5 seconds.  I'll leave tuning to follow ups.

PR #5144, opened 2020-07-16
[safety-rules] allow re-initialization of storage

davidiw:
In general, we wanted to avoid re-initialization of storage because it
shouldn't happen in integration and unit tests. However, cluster tests
do expect to re-initialize and we could even imagine cases where some
integration and smoke tests would simply re-initialize as well --
obviously we have worked around that issue to get the testability that
we have, but ... let's just get this working so we can ensure we have a
robust solution on our hands!

This should let this test pass: https://github.com/libra/libra/pull/5132/

### DIFF
--- a/consensus/safety-rules/src/persistent_safety_storage.rs
+++ b/consensus/safety-rules/src/persistent_safety_storage.rs
@@ -71,7 +71,17 @@ impl PersistentSafetyStorage {
         execution_private_key: Ed25519PrivateKey,
         waypoint: Waypoint,
     ) -> Result<()> {
-        internal_store.import_private_key(CONSENSUS_KEY, consensus_private_key)?;
+        let result = internal_store.import_private_key(CONSENSUS_KEY, consensus_private_key);
+        // Attempting to re-initialize existing storage. This can happen in environments like
+        // cluster test. Rather than be rigid here, leave it up to the developer to detect
+        // inconsistencies or why they did not reset storage between rounds. Do not repeat the
+        // checks again below, because it is just too strange to have a partially configured
+        // storage.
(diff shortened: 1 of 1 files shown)


consensus/safety-rules/src/persistent_safety_storage.rs
```diff
@@ -71,7 +71,16 @@ impl PersistentSafetyStorage {
         execution_private_key: Ed25519PrivateKey,
         waypoint: Waypoint,
     ) -> Result<()> {
-        internal_store.import_private_key(CONSENSUS_KEY, consensus_private_key)?;
+        let result = internal_store.import_private_key(CONSENSUS_KEY, consensus_private_key);
+        // Attempting to re-initialized existing storage. This can happen in environments like
```
JoshLind, line 75: nit: "re-initialized" -> "re-initialize"

consensus/safety-rules/src/persistent_safety_storage.rs
```diff
@@ -71,7 +71,16 @@ impl PersistentSafetyStorage {
...
+        // cluster test. Rather than be rigid here, leave it up to the developer to detect
+        // inconsistencies or why they did not reset storage between rounds. Do not repeat the
+        // checks again below, because it is just too strange to have a partially configured
+        // storage.
+        if let Err(libra_secure_storage::Error::KeyAlreadyExists(_)) = result {
+            return Ok(())
```
JoshLind, line 81: nit: I wonder if it's worth logging that initialization failed because the data already exists? At least then we won't get completely silent failures (e.g., if someone assumes this method completed all the way to the end)

PR #5145, opened 2020-07-16
[config] Remove max_peers_size from example configs

gregnazario:
The config wasn't setting a specific default, so if it was missing from
the config, it would likely set to the usize default which is 0.


gregnazario:
> Do we want to make the value a const?

Yeah, let's do that.

PR #5146, opened 2020-07-16
[compiler] add constant mapping to source map

emmazzz:
## Motivation

This PR adds a mapping from constant pool indices to constant names in `SourceMap`. In a later PR, we will use the source map to support constants in move prover.

Yes.

## Test Plan

cargo test

### DIFF
--- a/language/compiler/bytecode-source-map/src/source_map.rs
+++ b/language/compiler/bytecode-source-map/src/source_map.rs
@@ -4,15 +4,15 @@
 use anyhow::{format_err, Result};
 use libra_types::account_address::AccountAddress;
 use move_core_types::identifier::Identifier;
-use move_ir_types::ast::{ModuleName, NopLabel, QualifiedModuleIdent};
+use move_ir_types::ast::{ConstantName, ModuleName, NopLabel, QualifiedModuleIdent};
 use serde::{Deserialize, Serialize};
 use std::{collections::BTreeMap, ops::Bound};
 use vm::{
     access::*,
     file_format::{
-        CodeOffset, CompiledModule, CompiledScript, FunctionDefinition, FunctionDefinitionIndex,
-        LocalIndex, MemberCount, ModuleHandleIndex, StructDefinition, StructDefinitionIndex,
-        TableIndex,
+        CodeOffset, CompiledModule, CompiledScript, ConstantPoolIndex, FunctionDefinition,
+        FunctionDefinitionIndex, LocalIndex, MemberCount, ModuleHandleIndex, StructDefinition,
+        StructDefinitionIndex, TableIndex,
     },
 };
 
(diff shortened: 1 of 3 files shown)


emmazzz:
Submitting this PR so @tnowacki can take a look before I make too many mistakes.

language/compiler/bytecode-source-map/src/source_map.rs
```diff
@@ -449,6 +453,20 @@ impl<Location: Clone + Eq> SourceMap<Location> {
...
+        const_idx: ConstantPoolIndex,
+        name: ConstantName,
+    ) -> Result<()> {
+        self.constant_map
+            .entry(const_idx.0)
+            .or_insert_with(Vec::new);
+        self.constant_map.entry(const_idx.0).and_modify(|v| {
+            v.push(name);
+        });
+        Ok(())
```
tnowacki, line 467: You can chain this. So
```
self.constant_map
    .entry(const_idx.0)
    .or_insert_with(Vec::new)
    .push(name)     
```
should work

language/compiler/bytecode-source-map/src/source_map.rs
```diff
@@ -449,6 +453,20 @@ impl<Location: Clone + Eq> SourceMap<Location> {
                 )) })
     }
 
+    pub fn add_const_mapping(
+        &mut self,
+        const_idx: ConstantPoolIndex,
+        name: ConstantName,
+    ) -> Result<()> {
```
tnowacki, line 460: @tzakian Thoughts on this being a `Result<()>`?

It currently can't fail, so it feels weird having a result here. But, it might be nice to have for API parity?

language/compiler/bytecode-source-map/src/source_map.rs
```diff
@@ -549,6 +567,13 @@ impl<Location: Clone + Eq> SourceMap<Location> {
                 .dummy_struct_map(&module, &struct_def, default_loc.clone())?;
         }
 
+        for const_idx in 0..module.constant_pool().len() {
+            empty_source_map.add_const_mapping(
+                ConstantPoolIndex(const_idx as TableIndex),
+                ConstantName::new(format!("CONST{}", const_idx)),
+            )?;
+        }
```
tnowacki, line 575: @tzakian thoughts? I'm not familiar with this area

language/compiler/bytecode-source-map/src/source_map.rs
tzakian: I think not having a result where we don't need one is probably preferable. But not a strong opinion.

language/compiler/bytecode-source-map/src/source_map.rs
tzakian: This makes sense to me. Since there isn't a source map there isn't really a way of getting the name of the constant. This is similar to what is done for type arguments/function arguments if a source map doesn't exist.

language/compiler/bytecode-source-map/src/source_map.rs
```diff
@@ -71,6 +71,9 @@ pub struct SourceMap<Location: Clone + Eq> {
 
     // A mapping of FunctionDefinitionIndex to the soure map for that function.
     function_map: BTreeMap<TableIndex, FunctionSourceMap<Location>>,
+
+    // A mapping of ConstantPoolIndex to a vector of constant names.
+    pub constant_map: BTreeMap<TableIndex, Vec<ConstantName>>,
```
tzakian, line 76: Probably a stupid question, but this seems to imply that  a constant at a table index can have multiple names, which doesn't seem right. Shouldn't it be something like `BTreeMap<TableIndex, ConstantName>`?

emmazzz: There can indeed be multiple names. If we have two constants with the same value, they will be mapped to the same index in the constant pool.

PR #5147, opened 2020-07-16
Adding /v1 for JSONRPC endpoint.

thefallentree:
Create an /v1 endpoint ,  but for now still allows user to use "/"

in the future PR we will start to redirect "/" to latest "/vX" .


xli:
do we need also update testsuites/cli?

Issue #5063, opened 2020-07-14
[cluster-test] [Node launcher 1-3/n] Separate nodes and instances


andll:
There going to be more issues in the thread. Goal of all of this is to change how we launch libra nodes in cluster test to allow higher degree of flexibility when writing tests and unblocking some scenarios that are not currently possible.

Part 1-3 of this journey will be removing usage of libra-init from validator/full node specs.

### Seed IP bash magic
Currently when Instance runs, it initialize it's internal state by [running libra-init in the init container](https://github.com/libra/libra/blob/master/testsuite/cluster-test/src/cluster_swarm/validator_spec_template.yaml#L19).

This code contains bit of bash magic, that limits what we can do and hard to maintain.

We want to reduce amount of bash magic there :)

Specifically, init container does following:

- Configures fluentbit. *This part we can actually leave as is for now.*
- Removes data dir when `delete_data` is set
- Populates `safety_rules_addr` from kubernetes
- Populates `seed_peer_ip` from kuberenetes

The last 3 pieces is what we want to replace. Instead of searching for those values in Kubernetes, we want to pre-allocate nodes and populate those values right away in the cluster test.

Instead of figuring out IP addresses inside the container, we want to pre-allocate all the nodes in `ClusterBuilder`, and include ip addresses to the containers configuration before launching them.

The way this can be done:

* Make `ClusterSwarmKube::allocate_node` public. Currently node allocation is done internally in ClusterSwarmKube, but we want to change it.
* In `ClusterBuilder`, we should pre-allocate nodes for all containers using public method above.
* Add appropriate fields(safety_rules_addr/seed_peer_ip) to the ValidatorConfig/FullNodeConfig. Note that seed_peer_ip has different meaning for validators and full nodes - for validators this is ip of pod 0 for validator, while for full node this is IP address of appropriate validator
* In ClusterSwarmKube::validator_spec/fullnode_spec we should use those parameters in formatting spec template
* In the validator|fullnode_spec_template files, we need to remove kubectl magic and usage of `/opt/libra/data/seed_peer_ip`, etc files, and use supplied parameters instead.

### Delete data
After addressing seed IP problem, we need to look at the delete_data(in a separate PR):

* delete_data parameter should be removed from template and code, like `Instance::start` and `spawn_new_instance`. Default should be false(this is also most common I believe)
[...]

PR #5148, opened 2020-07-16
[cluster-test][Node launcher] remove seed_peer_ip and safety_rules_addr bash magic and instead pre-allocate nodes for validator and fullnode set.

changyonghu:
remove searching seed_peer_ip through kubectl, instead pre-allocate node for each pod and config seed_peer_ip and safety_rules_addr in cluster_builder.
## Related Issue: #5063

### DIFF
--- a/testsuite/cluster-test/src/cluster_builder.rs
+++ b/testsuite/cluster-test/src/cluster_builder.rs
@@ -6,6 +6,7 @@ use crate::{
     cluster::Cluster,
     cluster_swarm::{cluster_swarm_kube::ClusterSwarmKube, ClusterSwarm},
     instance::{
+        fullnode_pod_name, lsr_pod_name, validator_pod_name, vault_pod_name,
         ApplicationConfig::{Fullnode, Validator, Vault, LSR},
         FullnodeConfig, Instance, InstanceConfig, LSRConfig, ValidatorConfig, ValidatorGroup,
         VaultConfig,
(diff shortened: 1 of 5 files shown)


changyonghu:
> It looks like we are doing two step process, when we first allocate nodes with seed_peer_ip=None, and then update seed_peer_ip.
> 
> Does it have to be that way?
> We create vault-lsr first, then create validators, then full nodes.
> 
> It seems at every stage we have all the data we need right away - when validators are created, safety_rules_addr and seed_peer_ip are already known, and when full nodes are created, we already allocated nodes for validators, so their IPs are known too?

Oh yes, good point to notice the dependent creation.

zekun000:
can you elaborate more in the summary why this is better/required?

testsuite/cluster-test/src/cluster_builder.rs
```diff
@@ -163,13 +176,27 @@ impl ClusterBuilder {
                 .collect();
             lsrs.append(&mut lsr_instances);
         }
+
+        let validator_nodes = try_join_all((0..num_validators).map(|i| async move {
+            let pod_name = format!("val-{}", i);
```
andll, line 181: Do you mind moving this format! to Instance as some static method (like `Instance::validator_pod_name(index)`), and modify Instance::pod_name so it calls to this method. Basically, I want to make sure that `format!("val-{}")` happens exactly once(at least in rust code) so that if we change it, there is only one place to update

andll:
@zekun000 idea here is to remove bunch of config-builder related bash magic from the validator spec, which is one of the things needed to unblock migration to management tools.

testsuite/cluster-test/src/cluster_builder.rs
changyonghu: sure.

PR #5152, opened 2020-07-16
[tools] split genesis tool from libra-management

gregnazario:
In order to have a shared libary between multiple CLI tools, I've moved
the genesis tool to a separate crate.  This will let us have common code
between the tools.  Note: there are some rust files (StorageHelper / ConfigBuilder) that are tied to the
genesis tool used in testing, that may not necessarily belong there as
the first step in this split.

The main pieces here are:
1. Pull all that code out of the `lib.rs` for `Command` and put it in it's own crate with the CLI tool
2. Reconcile all the `use` moves


config/management/genesis/Cargo.toml
```diff
@@ -0,0 +1,42 @@
+[package]
+name = "libra-genesis-tool"
+version = "0.1.0"
+authors = ["Libra Association <opensource@libra.org>"]
+description = "Libra Management is a tool used to manage the configuration of a Libra Node"
```
JoshLind, line 5: nit: perhaps update this description to be about the genesis tool, as it's the genesis module, no?

config/management/genesis/Cargo.toml
```diff
@@ -0,0 +1,42 @@
...
+libra-secure-time = { path = "../../../secure/time", version = "0.1.0" }
+libra-types = { path = "../../../types", version = "0.1.0" }
+libra-workspace-hack = { path = "../../../common/workspace-hack", version = "0.1.0" }
+libra-temppath = { path = "../../../common/temppath", version = "0.1.0" }
+libra-vm = { path = "../../../language/libra-vm", version = "0.1.0" }
+libradb = { path = "../../../storage/libradb", version = "0.1.0" }
+storage-interface = { path = "../../../storage/storage-interface", version = "0.1.0" }
+transaction-builder = { path = "../../../language/transaction-builder", version = "0.1.0" }
+vm-genesis = { path = "../../../language/tools/vm-genesis", version = "0.1.0" }
+config-builder = { path = "../../config-builder", version = "0.1.0" }
```
JoshLind, line 38: nit: generally we do our best to keep all imports in alphabetical order :)

config/management/genesis/src/command.rs
```diff
@@ -0,0 +1,525 @@
...
+            _ => Err(self.unexpected_command(CommandName::Verify)),
+        }
+    }
+
+    fn unexpected_command(self, expected: CommandName) -> Error {
+        Error::UnexpectedCommand(expected.to_string(), CommandName::from(&self).to_string())
+    }
+}
+
+/// These tests depends on running Vault, which can be done by using the provided docker run script
```
JoshLind, line 203: nit: do you mind removing these comments? I think this is out-of-date as the tests now use onDiskStorage. 'cc @davidiw 👀

config/management/genesis/src/storage_helper.rs
```diff
@@ -198,7 +201,11 @@ impl StorageHelper {
     }
 
     #[cfg(test)]
-    pub fn set_layout(&self, path: &str, namespace: &str) -> Result<crate::layout::Layout, Error> {
+    pub fn set_layout(
+        &self,
+        path: &str,
+        namespace: &str,
+    ) -> Result<libra_management::layout::Layout, Error> {
```
JoshLind, line 208: nit: if we're going to just import libra_management things directly (e.g., libra_management::secure_backend::DISK), we could just import layout too :)

gregnazario: Yeah, my IDE didn't like it before but now it does 🤷🏻‍♂️

config/management/genesis/src/command.rs
gregnazario: I won't remove them right now but we can in the future.

PR #5154, opened 2020-07-16
[smoke test] vfn failover

christinacelee:
## Summary

This PR adds a smoke test for a VFN being able to failover to its fallback on-chain VFN peers if its upstream validator goes down, and all the mempool/state sync/config changes made to make this happen:
- incorporate `ConnectionOrigin` info into determining whether a peer is upstream in mempool/state sync
- config changes: add another public network to VFN s.t. it has one public network reserved for upstream connections (on-chain), and another public network for downstream connections
- add `NodeNetworkId`, an identifier unique across all networks a node belongs to. We need this because currently NetworkId is no longer unique (the VFN now has two public networks) within a node
- change mempool failover peer picking logic to broadcasting to all peers in the failover network peer in failover mode. this change is originally made to handle the situation where mempool chooses to broadcast itself only, but need to learn more details about upgrade scenario to see whether this change should be permanent. Will put out a follow-up PR accordingly if this is the case
- updated state sync timeout config value from 30 seconds to 10 seconds

## Test Plan

new smoke test, refactored/existing tests

### DIFF
--- a/config/src/config/mod.rs
+++ b/config/src/config/mod.rs
@@ -257,13 +257,13 @@ impl NodeConfig {
         for network in &mut config.full_node_networks {
             network.load(RoleType::FullNode)?;
 
-            // Validate that a network isn't repeated
-            let network_id = network.network_id.clone();
+            // Check a validator network is not included in a list of full-node networks
+            let network_id = &network.network_id;
             invariant(
-                !network_ids.contains(&network_id),
-                format!("network_id {:?} was repeated", network_id),
+                !matches!(network_id, NetworkId::Validator),
+                "Included a validator network in full_node_networks".into(),
             )?;
-            network_ids.insert(network_id);
+            network_ids.insert(network_id.clone());
         }
         config.set_data_dir(config.data_dir().clone());
         Ok(config)
--- a/config/src/config/test_data/validator_full_node.yaml
+++ b/config/src/config/test_data/validator_full_node.yaml
@@ -26,14 +26,17 @@ full_node_networks:
               token:
                   from_disk: "/full/path/to/token"
       network_id: "public"
-      max_frame_size: 8388608 # 8 MiB
     - listen_address: "/ip4/0.0.0.0/tcp/7180"
       network_id:
           private: "vfn"
       seed_addrs:
           "c227da54069989f283712e4016704660":
               - "/ip4/127.0.0.1/tcp/58259/ln-noise-ik/c998dcd54c3daf48e0ad516d94b7be0b0b7a27caa00541f2b2c14b13500df10b/ln-handshake/0"
+    - discovery_method: "onchain"
+      listen_address: "/ip4/0.0.0.0/tcp/8180"
+      network_id: "public"
 
 upstream:
     networks:
         - private: vfn
+        - public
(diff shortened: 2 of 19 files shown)


config/src/config/mod.rs
```diff
@@ -259,10 +259,13 @@ impl NodeConfig {
...
-                !network_ids.contains(&network_id),
-                format!("network_id {:?} was repeated", network_id),
-            )?;
+            // Allow multiple public networks for a node (may change in the future)
+            if network_id != NetworkId::Public {
+                invariant(
+                    !network_ids.contains(&network_id),
+                    format!("network_id {:?} was repeated", network_id),
+                )?;
+            }
```
davidiw, line 268: let's just delete this whole thing... sanity checking be damned

davidiw: but we shouldn't see the validator network here, that's for sure.

config/src/config/test_data/validator_full_node.yaml
```diff
@@ -26,14 +26,17 @@ full_node_networks:
...
           "c227da54069989f283712e4016704660":
               - "/ip4/127.0.0.1/tcp/58259/ln-noise-ik/c998dcd54c3daf48e0ad516d94b7be0b0b7a27caa00541f2b2c14b13500df10b/ln-handshake/0"
+    - discovery_method: "onchain"
+      listen_address: "/ip4/0.0.0.0/tcp/8180"
+      network_id: "public"
 
 upstream:
     networks:
         - private: vfn
+        - public
```
davidiw, line 42: cc: @mgorven for config changes

testsuite/tests/libratest/smoke_test.rs
```diff
@@ -1217,6 +1217,149 @@ fn test_client_waypoints() {
     assert!(client_with_bad_waypoint.test_trusted_connection().is_err());
 }
 
+#[test]
+fn test_vfn_failover() {
```
davidiw, line 1221: beautiful... but scary long

config/src/config/test_data/validator_full_node.yaml
mgorven: What does this config change do? Why is it needed?

christinacelee: tl;dr: this adds a public network for a node to failover to for state sync/mempool if the first private vfn network goes down

all networks listed here are 'upstream' networks, which means a node will request state sync from / broadcast txns to this peers in this network. The networks are ordered in preference, so adding the public network allows a node to be able to fallback to it if the first (=highest preferred) network is unavailable

mgorven: So this means that a VFN will try to sync state from whatever random public full nodes are connected to it? Is it expected that running a public fullnode potentially means providing outgoing state sync bandwidth to VFNs?

christinacelee: public network here is going to be the validators advertised on-chain (which is their VFN), so the VFN will in reality fallback to another VFN

mgorven: Oh, there are two `public` networks defined in this config file, that doesn't seem right. The network with other VFNs should be named differently.

zekun000: this is due to our **beautiful** network model, I don't think we can change the name since one of the public actually connects to another public (in a different node) (they handshake the network id) lol

PR #5157, opened 2020-07-16
[move-prover] Separate resource memory by type.

wrwg:
This PR implements separation of resource memory by type instead of using one global memory variable. This makes life for Z3 simpler because

- Automatic modifies analysis is more precise and only those memories are marked as modified which are actually touched on a per-type basis.
- Less aliasing has to be considered which likely reduces case splitting.

As a result, all timeout problems in existing modules could be resolved (except one related to loops). Note that this doesn't necessarily mean yet we have more verified, because many of the functions timing out have no specification conditions yet, but where just calling other functions which have. The overall performance improvement is hard to measure, but some problems which timed out now pass in seconds, and LibraAccount verification has become faster by about 40%.

The separation of memory is possible because:

- Move does not allow for `borrow_global<T>` where T is a type parameter. The outermost type in borrow must be statically known. This allows us to statically know the memory being queried.
- The new memory model made it particular easy to implement this, because WriteRef on an arbitrary reference does not commit to the global state yet. Rather commit happens only on the root, where we know the affected memory statically.

This PR also cleans up some stuff, removing Sender and transaction support, and fixing a few more loose ends.

## Motivation

Prover performance.

Yes

## Test Plan

Existing tests pass.

## Related PRs

NA


language/move-prover/spec-lang/src/ty.rs
```diff
@@ -131,6 +131,15 @@ impl Type {
         }
     }
 
+    /// Require this to be a struct and extract it contents.
```
emmazzz, line 134: should be `its contents`

language/move-prover/src/bytecode_translator.rs
```diff
@@ -926,8 +894,19 @@ impl<'env> ModuleTranslator<'env> {
             WriteBack(_, dest, src) => {
                 use BorrowNode::*;
                 match dest {
-                    GlobalRoot(_) => {
-                        emitln!(self.writer, "call $WritebackToGlobal({});", str_local(*src));
+                    GlobalRoot(struct_decl) => {
+                        let memory = boogie_resorce_memory_name(
```
shazqadeer, line 898: resorce is misspelled.

language/move-prover/src/bytecode_translator.rs
```diff
@@ -966,20 +945,19 @@ impl<'env> ModuleTranslator<'env> {
...
-                    emitln!(self.writer, "assume $ExistsTxnSenderAccount($m, $txn);");
                     let targets = &ctx.loop_targets[label];
                     for idx in 0..func_target.get_local_count() {
                         if let Some(ref_proxy_idx) = func_target.get_ref_proxy_index(idx) {
                             if targets.contains(ref_proxy_idx) {
                                 let ref_proxy_var_name = str_local(*ref_proxy_idx);
                                 let proxy_idx = func_target.get_proxy_index(idx).unwrap();
                                 emitln!(self.writer,
-                            "assume l#$Reference({}) == $Local({}) && p#$Reference({}) == $EmptyPath;",
+                            "assume l#$Mutation({}) == $Local({}) && p#$Mutation({}) == $EmptyPath;",
```
shazqadeer, line 960: I was wondering about the reason for this renaming Reference -> Mutation.  Perhaps you wanted to indicate that this  is reference is really a mutable reference.  Did you consider MutableReference?

language/move-prover/src/spec_translator.rs
```diff
@@ -1475,8 +1462,14 @@ impl<'env> SpecTranslator<'env> {
     fn translate_resource_access(&self, node_id: NodeId, args: &[Exp]) {
         self.trace_value(node_id, TraceItem::Exp, || {
             let rty = &self.module_env().get_node_instantiation(node_id)[0];
-            let type_value = self.translate_type(rty);
-            emit!(self.writer, "$ResourceValue($m, {}, ", type_value);
+            let (mid, sid, targs) = rty.require_struct();
```
shazqadeer, line 1465: Just trying to wrap my head around what is going on.  It looks like spec functions can access global memory.  Since the memory is split now, the parts that they access has to be passed in as explicit parameters.  Therefore, the lookup translation will also change.  Is my understanding correct?

language/stdlib/modules/DesignatedDealer.move
```diff
@@ -247,7 +247,7 @@ module DesignatedDealer {
     }
 
     spec module {
-        pragma verify = false;
+        pragma verify = true;
```
shazqadeer, line 250: Aha!  Did you have to make Libra::mint opaque to get the verification to succeed?

language/move-prover/src/prelude.bpl
```diff
@@ -72,7 +72,7 @@ function {:constructor} $IntegerType() : $TypeValue;
 function {:constructor} $AddressType() : $TypeValue;
 function {:constructor} $StrType() : $TypeValue;
 function {:constructor} $VectorType(t: $TypeValue) : $TypeValue;
-function {:constructor} $StructType(name: $TypeName, ps: $TypeValueArray, ts: $TypeValueArray) : $TypeValue;
+function {:constructor} $StructType(name: $TypeName, ts: $TypeValueArray) : $TypeValue;
```
shazqadeer, line 75: I remember you explained ps and ts to me once.  But I can't recall any more.  Since we only have ts, I assume this is simply the type arguments of the generic type which matches my intuition which is great.

language/move-prover/src/prelude.bpl
```diff
@@ -517,27 +518,42 @@ function {:inline} $InRange(r: $Value, i: int): bool {
...
+// are single threaded in Move, we can keep them together and treat them as a value
+// during mutation until the point they are stored back to their original location.
+type {:datatype} $Mutation;
+function {:constructor} $Mutation(l: $Location, p: $Path, v: $Value): $Mutation;
+const $DefaultMutation: $Mutation;
+
+// Representation of memory for a given type. The maps take the content of a Global location.
 type {:datatype} $Memory;
-function {:constructor} $Memory(domain: [$Location]bool, contents: [$Location]$Value): $Memory;
+function {:constructor} $Memory(domain: [$TypeValueArray, int]bool, contents: [$TypeValueArray, int]$Value): $Memory;
```
shazqadeer, line 544: We should remember to ask SMT experts whether there is a cost to multi-dimensional arrays.

language/move-prover/src/prelude.bpl
```diff
@@ -546,156 +562,129 @@ procedure {:inline 1} $InitVerification() {
...
 // Instructions
 
-procedure {:inline 1} $Exists(address: $Value, t: $TypeValue) returns (dst: $Value)
-{{backend.type_requires}} is#$Address(address);
-{
-    dst := $ResourceExists($m, t, address);
-}
-
-procedure {:inline 1} $MoveToRaw(ta: $TypeValue, a: int, v: $Value)
+procedure {:inline 1} $MoveToRaw(m: $Memory, ta: $TypeValueArray, a: int, v: $Value) returns (m': $Memory)
```
shazqadeer, line 595: Since all of global memory is being treated functionally (as inout parameter essentially), do we not even need global declarations ($m) any more?

language/move-prover/spec-lang/src/ty.rs
wrwg: Done.

language/move-prover/src/bytecode_translator.rs
wrwg: Done.

language/move-prover/src/bytecode_translator.rs
wrwg: If find both Reference and MutRef confusing, because the thing carries the value which is mutated. Under this view, Mutation made more sense to me (after playing with other options a while)

language/move-prover/src/spec_translator.rs
wrwg: Yes, all individual memories which are used by the helper function as well as the functions called by it need to be passed as individual parameters. Fortunately, a similar mechanism already existed for spec (ghost) variables, so that one just needed to be extended.

language/stdlib/modules/DesignatedDealer.move
wrwg: I did not change anything else.

wrwg:
> One thing I couldn't get out of examining the PR is how we are handling type quantification. My understanding was that the user could write formulas quantifying over all types. Could the translation of that feature be affected by the memory splitting?

This does not cause a problem, because in resource access, the outest type in `global<R>` always needs to be concrete. (Move restriction). We can still write `forall t: type: exists<R<t>>(addr)`. Already before, we could not write `forall t: type: exists<t>(addr)`.

language/move-prover/src/prelude.bpl
wrwg: We should remove ts in the near future. It are the types of the fields, but we don't need them if we do not do dynamic type assumptions.

wrwg: The other array are the field types. We would need it if we do dynamic type assumption, we can add it back later once we need this.

language/move-prover/src/prelude.bpl
wrwg: It's removed. We do not have a single state variable in the prelude, $txn is also gone.

language/move-prover/src/prelude.bpl
wrwg: ... where we should add: compared to a single variant datatype with two fields (tuple).

PR #5159, opened 2020-07-16
Add Error enum to JSON rpc client

ZhenjieRuan:
## Motivation

Add error enum to json rpc client to distinguish between retriable and non-retriable error.

Yes

## Test Plan
cargo test

### DIFF
--- a/client/json-rpc/src/client.rs
+++ b/client/json-rpc/src/client.rs
@@ -115,6 +115,44 @@ impl JsonRpcBatch {
     }
 }
 
+#[derive(Debug)]
+pub enum JsonRpcAsyncClientError {
+    ClientError(reqwest::Error),
+    InvalidServerResponse(String),
+}
+
+impl JsonRpcAsyncClientError {
+    pub fn is_retriable(&self) -> bool {
+        match self {
+            JsonRpcAsyncClientError::ClientError(e) => {
+                // TODO: Add following condition when is_request is added to reqwest stable release
+                // // Error sending request
+                //  if e.is_request() {
+                //      return true;
+                //  }
+                if let Some(status) = e.status() {
+                    // Returned status code indicates a server error
+                    return status.is_server_error();
+                }
+                false
+            }
+            JsonRpcAsyncClientError::InvalidServerResponse(_) => false,
+        }
+    }
+}
+
+impl std::convert::From<JsonRpcAsyncClientError> for anyhow::Error {
+    fn from(e: JsonRpcAsyncClientError) -> Self {
+        anyhow::Error::msg(e.to_string())
+    }
+}
+
+impl std::fmt::Display for JsonRpcAsyncClientError {
+    fn fmt(&self, f: &mut std::fmt::Formatter) -> std::fmt::Result {
+        write!(f, "JsonRpcAsyncClientError {}", self)
+    }
+}
+
 #[derive(Clone)]
 pub struct JsonRpcAsyncClient {
     address: String,
@@ -143,7 +181,7 @@ impl JsonRpcAsyncClient {
     pub async fn get_accounts(
         &self,
         accounts: &[AccountAddress],
-    ) -> Result<Vec<Option<AccountView>>> {
+    ) -> Result<Vec<Option<AccountView>>, JsonRpcAsyncClientError> {
         let mut batch = JsonRpcBatch::new();
         for account in accounts {
             batch.add_get_account_request(*account);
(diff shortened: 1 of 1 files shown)


client/json-rpc/src/client.rs
```diff
@@ -115,6 +115,33 @@ impl JsonRpcBatch {
     }
 }
 
+#[derive(Debug)]
+pub enum JsonRpcAsyncClientError {
+    SubmissionError(reqwest::Error),
```
thefallentree, line 120: can we name this ClientError  / InvalidServerResponse to match what the python library is doing?


cc @ssinghaldev  to provide more context around this

dmitri-perelman: Tbh we're less interested about Client vs. Server dichotomy, and more about whether to retry or not. For example, we don't want to retry in case a given transaction cannot be properly serialized (but it's not a `InvalidServerResponse`). In that sense, Execution vs Submission could be a nice way to partition things. 
No strong opinions about naming though. We could also just add more enum options to be more specific about the error types.

client/json-rpc/src/client.rs
```diff
@@ -143,54 +170,76 @@ impl JsonRpcAsyncClient {
...
+                "Http error code {}",
+                resp.status()
+            )));
+        }
 
-        let responses: Vec<Value> = resp.json().await?;
+        let responses: Vec<Value> = resp
+            .json()
+            .await
+            .map_err(|e| JsonRpcAsyncClientError::ExecutionError(e.to_string()))?;
```
dmitri-perelman, line 240: When does this function return an error? In case it's about connectivity issues, it shouldn't be ExecutionError.

client/json-rpc/src/client.rs
```diff
@@ -143,54 +170,76 @@ impl JsonRpcAsyncClient {
...
             .send()
-            .await?;
+            .await
+            .map_err(JsonRpcAsyncClientError::SubmissionError)?;
 
-        ensure!(
-            resp.status() == 200,
-            format!("Http error code {}", resp.status())
-        );
+        if resp.status() != 200 {
```
dmitri-perelman, line 230: I suspect that the specific type of http error could determine whether to retry or not, but I don't know the details. Should it be a TODO?

client/json-rpc/src/client.rs
dmitri-perelman: After reading the code I'd like to propose a more detailed enum with more values, e.g., `WrongInput`, `HttpError` (with http error code), `ConnectivityError`. The client could then determine the retry policy for themselves.

client/json-rpc/src/client.rs
ZhenjieRuan: This means we can't de-serialize the json from response, connection error will be thrown when sending the request

client/json-rpc/src/client.rs
ZhenjieRuan: If the return status is not in 200 range, it will already error out when we are sending the request.

ZhenjieRuan:
@dmitri-perelman I think the `reqwest::Error` already includes all three types of error you mentioned already. I'm not against having a more detailed enum though, because one risk we are taking here is that we are making ourselves depend on specifics of reqwest here, may make it hard to replace in the future

client/json-rpc/src/client.rs
ZhenjieRuan: @dmitri-perelman I think the reqwest::Error already includes all three types of error you mentioned already. I'm not against having a more detailed enum though, because one risk we are taking here is that we are making ourselves depend on specifics of reqwest here, may make it hard to replace in the future

ssinghaldev: > can we name this ClientError / InvalidServerResponse to match what the python library is doing?
> 
> cc @ssinghaldev to provide more context around this

Just for adding info on how we are doing it in python client. Errors are ValueError(WrongInput), RequestException(ConnectionError), InvalidServerResponse(deserializing & response is not according to json-rpc spec), JsonRpcError(which is when the rpc returns "error" instead of result). These errors are from json-rpc client I have written. But the actual API (which uses json-rpc underneath) throws ValueError(re-raises basically), packages RequestException, InvalidServerResponse, JsonRpcError  into ClientError & throws that. For now, the idea is ValueError represents - u can't retry with same options but with ClientError u can. Ofccourse, it can be more fine grained. But for now, we have this. 

The idea of repackaging is basically @ZhenjieRuan mentioned that we do not wish to depend/pass on Library error(in my context, it is request & underlying json-rpc client which I have written)

client/json-rpc/src/client.rs
dmitri-perelman: Huh, so you suspect this check is kind of redundant?

client/json-rpc/src/client.rs
dmitri-perelman: Sure, no objection for renaming.

client/json-rpc/src/client.rs
```diff
@@ -143,54 +181,76 @@ impl JsonRpcAsyncClient {
...
         let exec_results = self.execute(batch).await?;
         let mut results = vec![];
         for exec_result in exec_results {
-            let exec_result = exec_result?;
+            let exec_result = exec_result
+                .map_err(|e| JsonRpcAsyncClientError::InvalidServerResponse(e.to_string()))?;
             if let JsonRpcResponse::AccountResponse(r) = exec_result {
                 results.push(r);
             } else {
                 panic!("Unexpected response for get_accounts {:?}", exec_result)
```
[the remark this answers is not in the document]
thefallentree: agreed

client/json-rpc/src/client.rs
```diff
@@ -143,54 +181,76 @@ impl JsonRpcAsyncClient {
...
-    pub async fn submit_transaction(&self, txn: SignedTransaction) -> Result<()> {
+    pub async fn submit_transaction(
+        &self,
+        txn: SignedTransaction,
+    ) -> Result<(), JsonRpcAsyncClientError> {
         let mut batch = JsonRpcBatch::new();
-        batch.add_submit_request(txn)?;
+        batch
+            .add_submit_request(txn)
+            .map_err(|e| JsonRpcAsyncClientError::InvalidServerResponse(e.to_string()))?;
```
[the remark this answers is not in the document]
thefallentree: +1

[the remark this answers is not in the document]
thefallentree: This should be an InvalidArgument   , but mostly just no need for map_err

PR #5161, opened 2020-07-17
[management] make attach_validator_full_node logic less dependent on …

christinacelee:
…FN network config ordering

## Motivation

Currently `FullnodeBuilder::attach_validator_full_node` relies too much on the ordering of the networks in the config, which makes swarm-building error prone. This PR reduces that risk, especially if the config yaml files change

## Test Plan

Existing tests

### DIFF
--- a/config/management/src/config_builder.rs
+++ b/config/management/src/config_builder.rs
@@ -3,9 +3,12 @@
 
 use crate::{constants, layout::Layout, storage_helper::StorageHelper};
 use config_builder::BuildSwarm;
-use libra_config::config::{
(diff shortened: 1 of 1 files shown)


config/management/src/config_builder.rs
```diff
@@ -272,7 +281,17 @@ impl FullnodeBuilder {
         let mut seed_addrs = SeedAddresses::default();
         seed_addrs.insert(v_vfn_id, vec![v_vfn_network_address]);
 
-        let fn_vfn = &mut full_node_config.full_node_networks[1];
+        let fn_vfn = &mut full_node_config
+            .full_node_networks
+            .iter_mut()
+            .find(|n| {
+                if let NetworkId::Private(_) = n.network_id {
```
zekun000, line 288: you can use matches!

PR #5162, opened 2020-07-17
[test-cli] add info command to replace cli output debug log

xli:
## Motivation

When use test cli, we output verbose information / log about client trusted state and epoch change, these logs becomes noise for new user to try out cli.

This change turns off these log by setting logger level to warn when cli option verbose is not set to true, added a simple info command for checking client trusted state and ledger info.

## Test Plan

Manual test:

```

Wallet recovered and the first 0 child accounts were derived
Connected to validator at: http://localhost:65113, latest version = 5, timestamp = 2020-07-17 00:21:06.402935 UTC
usage: <command> <args>

Use the following commands:

......
info | i
	Print cli config and client internal information
......

Please, input commands:

libra% a c
>> Creating/retrieving next account from wallet
Created/retrieved account #0 address 40f54565890f8b3adf546e6958a050f7
libra% a mb 0 111 LBR
>> Minting coins
waiting for 0000000000000000000000000a550c18 with sequence number 2
......................................................................................................
transaction executed!
no events emitted
waiting for 000000000000000000000000000000dd with sequence number 1
..........................................................................................
transaction executed!
Finished minting!

libra% i
ChainID: ChainId 4
Trusted State: TrustedState {
    verified_state: Waypoint {
        version: 5,
        value: .........
    },
    verifier: EpochState {
        epoch: 1,
        verifier: ValidatorVerifier {
              .........
        },
    },
}
LedgerInfo: Some(
    V0(
        LedgerInfoWithV0 {
            ledger_info: LedgerInfo {
                commit_info: BlockInfo {
                      ..........
                },
                consensus_data_hash: .....
            },
            signatures: {},
        },
    ),
)
libra%

```

### DIFF
--- a/testsuite/cli/src/client_proxy.rs
+++ b/testsuite/cli/src/client_proxy.rs
@@ -739,6 +739,7 @@ impl ClientProxy {
                 .get_txn_by_acc_seq(account, sequence_number - 1, true)
             {
                 Ok(Some(txn_view)) => {
+                    println!();
                     if txn_view.vm_status == VMStatusView::Executed {
                         println!("transaction executed!");
                         if txn_view.events.is_empty() {
@@ -753,6 +754,7 @@ impl ClientProxy {
                     }
                 }
                 Err(e) => {
+                    println!();
                     println!("Response with error: {:?}", e);
                 }
                 _ => {
--- a/testsuite/cli/src/libra_client.rs
+++ b/testsuite/cli/src/libra_client.rs
@@ -279,6 +279,11 @@ impl LibraClient {
         self.latest_epoch_change_li.as_ref()
     }
 
+    /// Latest trusted state
+    pub(crate) fn trusted_state(&self) -> TrustedState {
+        self.trusted_state.clone()
+    }
+
     /// Get transaction from validator by account and sequence number.
     pub fn get_txn_by_acc_seq(
         &mut self,
--- a/testsuite/cli/src/commands.rs
+++ b/testsuite/cli/src/commands.rs
@@ -3,7 +3,7 @@
 
 use crate::{
     account_commands::AccountCommand, client_proxy::ClientProxy, dev_commands::DevCommand,
-    query_commands::QueryCommand, transfer_commands::TransferCommand,
+    info_commands::InfoCommand, query_commands::QueryCommand, transfer_commands::TransferCommand,
 };
 use anyhow::Error;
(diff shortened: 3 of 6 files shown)


testsuite/cli/src/libra_client.rs
```diff
@@ -252,21 +251,11 @@ impl LibraClient {
...
-                        .ledger_info()
-                        .next_epoch_state()
-                        .expect("no validator set in epoch change ledger info"),
-                );
                 // Update client state
                 self.trusted_state = new_state;
                 self.latest_epoch_change_li = Some(latest_epoch_change_li.clone());
             }
             TrustedStateChange::Version { new_state } => {
-                if self.trusted_state.latest_version() < new_state.latest_version() {
```
christinacelee, line 267: I can see how this can be considered too 'noisy', but for me personally, I found this actually gives a good sanity check on a node's progress during a blocking call to a client request 😮

zekun000:
I mostly rely on this log in smoke test to see if we're making progress, maybe we can provide a param in constructor for those messages?

xli:
> I mostly rely on this log in smoke test to see if we're making progress, maybe we can provide a param in constructor for those messages?

could you point me to the smoke test rely on this log? All tests passed after removed, so I don't think there is smoke test rely on it.
Or it is more of for checking log when the test failed?

testsuite/cli/src/client_proxy.rs
```diff
@@ -753,6 +754,7 @@ impl ClientProxy {
                     }
                 }
                 Err(e) => {
+                    println!("");
```
xli, line 757: end the ..... line

testsuite/cli/src/client_proxy.rs
```diff
@@ -739,6 +739,7 @@ impl ClientProxy {
                 .get_txn_by_acc_seq(account, sequence_number - 1, true)
             {
                 Ok(Some(txn_view)) => {
+                    println!();
```
christinacelee, line 742: maybe it'll be better to add the extra line after the actual "..." output here? https://github.com/libra/libra/blob/master/testsuite/cli/src/client_proxy.rs#L724-L759

xli: Add this is because we have following message "Transaction executed", without this, it looks like this:

```
..................................transaction executed!
```
After added this println:
```
..................................
transaction executed!
```

PR #5166, opened 2020-07-17
denylist -> blocklist

aching:
After some discussion, aligning on blocklist rather than denylist was preferred.


aching:
@bors-libra retry

Filed #5170 and #5171 for flaking unittests

PR #5167, opened 2020-07-17
[Prometheus] add storage api latency graphs

lightmark:
## Motivation

add api latency graphs to grafana

Y

## Test Plan

n/a

### DIFF
--- a/terraform/templates/dashboards/storage.json
+++ b/terraform/templates/dashboards/storage.json
@@ -15,23 +15,38 @@
   "editable": true,
   "gnetId": null,
   "graphTooltip": 0,
-  "id": 17,
-  "iteration": 1565807898889,
+  "iteration": 1595032058505,
   "links": [],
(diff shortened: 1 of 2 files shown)


msmouse:
* 1000000 is us, not ms.

remove it and do this:

image

msmouse:
create a "api_name" variable, like this:

image

image

and define widget like this:

image

image
image

and new widget will show up / disappear automatically when we add / remove new apis.

terraform/templates/dashboards/storage.json
```diff
@@ -2304,5 +3577,5 @@
   "timezone": "",
   "title": "Storage",
   "uid": "2Q0p7AVWk",
-  "version": 2
+  "version": 1
```
msmouse, line 3580: interesting it gets lower.. but probalbly doesn't matter

PR #5175, opened 2020-07-17
[swarm] enable-inject-error in smoke tests

zekun000:
## Motivation

This commit enables enable-inject-error in smoke tests with 10% chances
errors in a few places in consensus.

Y

## Test Plan

smoke tests, revealed https://github.com/libra/libra/issues/5180

## Related PRs

### DIFF
--- a/testsuite/libra-swarm/src/swarm.rs
+++ b/testsuite/libra-swarm/src/swarm.rs
@@ -20,8 +20,6 @@ use std::{
 };
 use thiserror::Error;
 
-const LIBRA_NODE_BIN: &str = "libra-node";
-
 pub struct LibraNode {
     node: Child,
     node_id: String,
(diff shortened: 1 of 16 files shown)


testsuite/libra-swarm/src/swarm.rs
```diff
@@ -57,7 +55,7 @@ impl LibraNode {
     /// substantially longer time than the networking ports are reserved. Calling prior to
     /// reserving those ports will reduce the liklihood of issues.
     pub fn prepare() {
-        Command::new(workspace_builder::get_bin(LIBRA_NODE_BIN));
+        Command::new(workspace_builder::get_libra_node_with_inject_error());
```
aching, line 58: A few thoughts:

1)  The default is 10%, it might be nice to environment variable to be able to override this to avoid having to recompile.

2) I wonder if it might make sense to run swarm with the injection turned on or off for tests rather than a 5% failure rate all the time (maybe an environment override or something).  When turned off the swarm tests should be fast and have a low timeout.  For the injection parts turned on perhaps boost the failure rate to something much higher, say 20% and see what happens and have a higher timeout.  

3) If we choose to just inject random errors in the swarm, it would be good to probably add some documentation so when test failures occur, it's pretty obvious that they were injected.  This is only for swarm tests and for others running the binary.  Maybe we could at least add an initialization message, "This binary is running with error injection in testing"?  That would also help with the scenario where someone accidentally tries to deploy an injected binary with error injection.

PR #5177, opened 2020-07-17
[crypto] Initial draft of the crypto specification

huitseeker:
Adds description for sign / verify / multi-sign


mimoo:
awesome work!

@bors-libra delegate+

PR #5179, opened 2020-07-17
[executor] reset txns to empty in reconfig suffix

lightmark:
## Motivation

A reconfiguration suffix block may have a metadata transaction but inherit the execution result from the parent block.

Y

## Test Plan

later, we will have injector-error to cause the crash.


aching:
@lightmark, could you please add a unittest for this particular scenario?

PR #5181, opened 2020-07-17
[management][2/2] use real shared_val_netaddr_key in genesis + config-builder

phlip9:
+ `NetworkConfig` can now load the `shared_val_netaddr_key` either from secure storage or directly from the config.

+ In the management tooling, operators specify the `shared_val_netaddr_key` in the `validator-config` command as an argument. This key is then used to encrypt the validator network address before is storing it in the local secure storage.

+ Add a wrapper struct for `netaddr::Key` and impl the relevant `libra_crypto::traits`.

+ Hide the `libra_network_address::encrypted` module and just re-export the important items.

+ Update the libra cli `register_validator` command. Though, it appears to be unused as it was broken previously...


config/global-constants/src/lib.rs
```diff
@@ -18,6 +18,7 @@ pub const OPERATOR_KEY: &str = "operator";
 pub const OWNER_ACCOUNT: &str = "owner_account";
 pub const OWNER_KEY: &str = "owner";
 pub const VALIDATOR_NETWORK_KEY: &str = "validator_network";
+pub const SHARED_VAL_NETADDR_KEY: &str = "shared_val_netaddr_key";
```
davidiw, line 21: oh god that is an ugly undecipherable string :P
what happened to validator_network_address_encryption_key
super long, but very descriptive

we could even add a shared in their...

maybe leveraging both of ours:
shared_validator_network_address_key
(it isn't an encryption key...)

config/src/config/network_config.rs
```diff
@@ -153,6 +160,45 @@ impl NetworkConfig {
...
+            SharedValNetAddrKeys::None => HashMap::new(),
+        };
+
+        assert!(
+            !keys.is_empty(),
+            "shared_val_netaddr_keys should not be empty for validators"
+        );
+        keys
+    }
+
```
davidiw, line 201: I wonder if this is a bit too future proof. Maybe for now we just want a single key?

config/src/config/network_config.rs
```diff
@@ -296,3 +342,16 @@ pub struct IdentityFromStorage {
...
     pub peer_id_name: String,
 }
+
+#[cfg_attr(any(test, feature = "fuzzing"), derive(Clone, PartialEq))]
+#[derive(Debug, Deserialize, Serialize)]
+#[serde(rename_all = "snake_case", tag = "type")]
+pub enum SharedValNetAddrKeys {
+    FromConfig {
+        keys: HashMap<netaddr::KeyVersion, netaddr::Key>,
+    },
```
davidiw, line 352: for example, this is only used for testing, making this robust suggests it isn't just for testing.

network/network-address/src/encrypted.rs
```diff
@@ -147,6 +159,83 @@ pub struct EncNetworkAddress {
 #[error("error decrypting network address")]
 pub struct DecryptError;
 
+/////////
```
davidiw, line 162: this feels like we should be building a generic AESKey or something in crypto

testsuite/cli/src/client_proxy.rs
```diff
@@ -663,47 +660,59 @@ impl ClientProxy {
...
 
         // parse args
-        let (address, _) = self.get_account_address_from_parameter(space_delim_strings[1])?;
-        let private_key = Ed25519PrivateKey::from_encoded_string(space_delim_strings[2])?;
-        let consensus_public_key = Ed25519PublicKey::from_encoded_string(space_delim_strings[3])?;
-        let network_identity_key = x25519::PublicKey::from_encoded_string(space_delim_strings[4])?;
-        let network_address = NetworkAddress::from_str(space_delim_strings[5])?;
-        let raw_network_address = RawNetworkAddress::try_from(&network_address)?;
-        let fullnode_identity_key = x25519::PublicKey::from_encoded_string(space_delim_strings[6])?;
+        let (account, _) = self.get_account_address_from_parameter(space_delim_strings[1])?;
```
davidiw, line 668: I cannot wait to delete all this code

config/global-constants/src/lib.rs
mgorven: That long isn't great either. How about "network_address_key"?

phlip9: i thought we already discussed naming in the previous PR? i feel "validator_network_address_encryption_key" and "shared_validator_network_address_key" are both too long and "network_address_key" is too ambiguous. hence, "shared_val_netaddr_key" : p

network/network-address/src/encrypted.rs
gregnazario: I agree, either `AESKey` or `SymmetricKey`.  Lower down it becomes a little more confusing because you're now using `netaddr::Key` to be specific to the key.  Naming things is hard :/

config/global-constants/src/lib.rs
mgorven: I'm fine with "shared_val_netaddr_key" BTW.

network/network-address/src/encrypted.rs
phlip9: technically this key is not directly used for AES : p

we probably want a more general + consistent approach to storing secrets. for example, not only this key but also the network identity private key(s) should be allocated on the heap, mlock'ed, with guard pages around it https://libsodium.gitbook.io/doc/memory_management

we could probably make a generic `SecretBox<T>` that does this

phlip9:
rebase

config/src/config/network_config.rs
phlip9: in my mind, it's just the secure storage model that's behind and will catch up after v1. though, it doesn't really matter to me either way.

config/src/config/network_config.rs
phlip9: personally, i don't like it when the test model diverges from the prod model. of course, prod is currently only using one key, so maybe that's a moot point : )

config/management/genesis/src/config_builder.rs
```diff
@@ -135,6 +136,7 @@ impl<T: AsRef<Path>> ValidatorBuilder<T> {
                 &(index.to_string() + OWNER_SHARED_NS),
                 validator_network_address,
                 fullnode_network_address,
+                TEST_SHARED_VAL_NETADDR_KEY,
```
davidiw, line 139: this should be in the owner storage not passed to the owner storage.

config/management/src/validator_config.rs
```diff
@@ -87,6 +86,20 @@ impl ValidatorConfig {
                 )
             })?;
 
+        // Import the shared_val_netaddr_key to local storage
```
davidiw, line 89: this is being set by terraform, we can also have an explicit command to set it, but we should only be reading it

config/src/config/network_config.rs
```diff
@@ -136,6 +142,45 @@ impl NetworkConfig {
...
+                    .expect("Unable to read shared_val_netaddr_keys")
+                    .value
+                    .bytes()
+                    .expect("Expected bytes for shared_val_netaddr_keys");
+                let key = netaddr::Key::try_from(key_bytes.as_slice())
+                    .expect("Expected shared_val_netaddr_keys bytes to be well-formatted");
+
+                // TODO(philiphayes): add rotation + multiple keys support. for
+                // now just assume a constant version.
+                let key_version = SHARED_VAL_NETADDR_KEY_VERSION;
```
davidiw, line 168: maybe we should be storing a tuple in storage (key, version)

PR #5185, opened 2020-07-17
[config] make key access easier for testing cases

davidiw:
We had overcomplicated our configs for testing purposes only. This allows us to make NodeConfigs non-mutable and cloneable without impacting the security of our actual production keys.

### DIFF
--- a/config/config-builder/src/full_node_config.rs
+++ b/config/config-builder/src/full_node_config.rs
@@ -180,10 +180,7 @@ impl FullNodeConfig {
             network.discovery_method = DiscoveryMethod::gossip(self.advertised_address.clone());
             network.mutual_authentication = self.mutual_authentication;
 
-            let pubkey = network
-                .identity
-                .public_key_from_config()
-                .ok_or(Error::MissingNetworkKeyPairs)?;
+            let pubkey = network.identity_key().public_key();
             let pubkey_set: HashSet<_> = [pubkey].iter().copied().collect();
             seed_pubkeys.insert(network.peer_id(), pubkey_set);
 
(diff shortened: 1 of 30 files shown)


davidiw:
> Why is the mutbility of configs problematic? What I mean by this is that it's a very restricted form of mutability, strictly following the [`mem::replace` pattern](https://github.com/rust-unofficial/patterns/blob/master/idioms/mem-replace.md). I.e. you either have a config with a private key or one that has `None`.
> 

I think we should have a principled approach in dealing with configs. They /should not/ be mutable and they should be copyable. This pattern is very problematic in that we allow test-only code to dictate the utility / simplicity of other components.

> If you indeed have (future) reasons to need cloneable private keys in production, why does this PR not remove the `Options` in `SafetyRulesConfig`, which are there exclusively to support the `mem::replace` pattern?
> 

Are you referring to the **optional** test config: https://github.com/libra/libra/blob/master/config/src/config/safety_rules_config.rs#L24

> We have several tools for simplifying the cloning patterns of private keys.
> Have you considered making testconfig accessible under `#[cfg(test)]` only? The new resolver has made the use of features less cumbersome, and under that flag, private keys are cloneable.
> 

As long as we rely on config builder to test production code (note **code** not **environment**), we're stuck in this ugliness.

> Have you thought of documenting the requirement, reinforced by this PR, that we assume validators run with a key manager in all cases? There's more work on bringing our memory use for confidential keys to state-of-the-art coming down the 'pike, and that may now be obsoleted, if we've taken a decision to rely on out-of-memory separation?

The name of each config wherein keys live (except networking) is called *TestConfig. I think that's pretty intuitive, but if there aren't comments, we can easily amend them.

For what it is worth, the keys from storage are cloneable because there's no sensible way to ensure that a key is only ever retrieved once and that may very well not be the right behavior either.

mimoo:
I can agree that configs should not be mutable, but clonable.
I think the only keys that we need to make sure don't get cloned everywhere are really network keys, and them being passed in a config file is not great in any case. (The others are in the TCB and it should be pretty easy to check that there.)

cc @phlip9 @gregnazario I think we should look into a way for full nodes to store and load network keys securely

@bors-libra delegate+

PR #5187, opened 2020-07-17
[types] Mainnet ChainId 0 -> 1 to handle accidental initialization.

aching:
## Motivation

Users might accidentally initialize the ChainId field to 0, hence reserving ChainId 0 for accidental initialization and starting mainnet at 1.


aching:
#5189 takes care of this.

PR #5191, opened 2020-07-18
[Key Manager] Add ChainID to ConfigBuilder and update docker containers.

JoshLind:
## Motivation

Chain ids were recently added to transactions to prevent replaying transactions across different blockchains (https://github.com/libra/libra/pull/4969). As part of this change, we need to update the config builder support for the key manager, as well as the docker containers, to allow configurable chain ids. This PR provides two commits:

1. First, we add chain id support to the config builder test for the key manager (as well as refactor the ordering of the chain id in the config).
2. Second, we update the key manager docker containers to pass the chain id to the key manager and use consistent "CFG_*" naming where appropriate.

Yes.

## Test Plan

All tests still pass locally. Moreover, running the key manager locally using the docker container seemed to work without any issues.

## Related PRs

https://github.com/libra/libra/pull/4969

### DIFF
--- a/config/config-builder/src/key_manager_config.rs
+++ b/config/config-builder/src/key_manager_config.rs
@@ -6,11 +6,11 @@ use libra_config::config::{KeyManagerConfig as KMConfig, SecureBackend, Token, V
 use libra_types::chain_id::ChainId;
 
 pub struct KeyManagerConfig {
-    pub chain_id: ChainId,
     pub rotation_period_secs: Option<u64>,
     pub sleep_period_secs: Option<u64>,
     pub txn_expiration_secs: Option<u64>,
 
+    pub chain_id: ChainId,
     pub json_rpc_endpoint: String,
 
     pub vault_host: String,
@@ -24,10 +24,10 @@ impl Default for KeyManagerConfig {
     fn default() -> Self {
         let template = KMConfig::default();
         Self {
-            chain_id: ChainId::test(),
             rotation_period_secs: None,
             sleep_period_secs: None,
             txn_expiration_secs: None,
+            chain_id: ChainId::test(),
             json_rpc_endpoint: template.json_rpc_endpoint.clone(),
             vault_host: "127.0.0.1:8200".to_string(),
             vault_namespace: None,
@@ -73,12 +73,14 @@ mod test {
 
     #[test]
     fn verify_generation() {
+        let chain_id = ChainId::test();
         let json_rpc_endpoint = "http://127.12.12.12:7873";
         let rotation_period_secs = 100;
         let vault_host = "182.0.0.1:8080";
         let vault_token = "root_token";
 
         let mut key_manager_config = KeyManagerConfig::new();
+        key_manager_config.chain_id = chain_id;
         key_manager_config.json_rpc_endpoint = json_rpc_endpoint.into();
         key_manager_config.rotation_period_secs = Some(rotation_period_secs);
         key_manager_config.vault_host = vault_host.into();
@@ -86,6 +88,7 @@ mod test {
 
         let key_manager_config = key_manager_config.build().unwrap();
 
+        assert_eq!(chain_id, key_manager_config.chain_id);
         assert_eq!(json_rpc_endpoint, key_manager_config.json_rpc_endpoint);
         assert_eq!(
             rotation_period_secs,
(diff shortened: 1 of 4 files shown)


docker/safety-rules/key-manager.sh
```diff
@@ -9,6 +9,9 @@ if [ -n "${KEY_MANAGER_CONFIG}" ]; then # Path to key manager config
     echo "${KEY_MANAGER_CONFIG}" > /opt/libra/etc/key_manager.yaml
     params+="--template /opt/libra/etc/key_manager.yaml "
 fi
+if [ -n "${CHAIN_ID}" ]; then
```
davidiw, line 12: interesting, why aren't these CFG_?
I guess because Michael hasn't gotten to them, but if you take a look at the rest they are all CFG_...

JoshLind:
@davidiw, rebased, please take another look when you get a chance 😄

docker/safety-rules/key-manager.sh
JoshLind: Done, thanks!

config/config-builder/src/key_manager_config.rs
```diff
@@ -6,11 +6,11 @@ use libra_config::config::{KeyManagerConfig as KMConfig, SecureBackend, Token, V
 use libra_types::chain_id::ChainId;
 
 pub struct KeyManagerConfig {
-    pub chain_id: ChainId,
     pub rotation_period_secs: Option<u64>,
     pub sleep_period_secs: Option<u64>,
     pub txn_expiration_secs: Option<u64>,
 
+    pub chain_id: ChainId,
```
davidiw, line 13: why? :O

config/config-builder/src/key_manager_config.rs
```diff
@@ -24,10 +24,10 @@ impl Default for KeyManagerConfig {
     fn default() -> Self {
         let template = KMConfig::default();
         Self {
-            chain_id: ChainId::test(),
             rotation_period_secs: None,
             sleep_period_secs: None,
             txn_expiration_secs: None,
+            chain_id: ChainId::test(),
```
davidiw, line 30: the ordering here is a bit interesting... why not everything in alpha?

config/config-builder/src/key_manager_config.rs
```diff
@@ -73,19 +73,22 @@ mod test {
...
         let vault_token = "root_token";
 
         let mut key_manager_config = KeyManagerConfig::new();
+        key_manager_config.chain_id = chain_id;
         key_manager_config.json_rpc_endpoint = json_rpc_endpoint.into();
         key_manager_config.rotation_period_secs = Some(rotation_period_secs);
         key_manager_config.vault_host = vault_host.into();
         key_manager_config.vault_token = vault_token.into();
 
         let key_manager_config = key_manager_config.build().unwrap();
```
davidiw, line 89: the naming here is weird, maybe you should call the first one builder and this one is the actual config. I had to double check what we're doing here :P

docker/safety-rules/key-manager.sh
```diff
@@ -12,26 +12,26 @@ fi
 if [ -n "${CFG_CHAIN_ID}" ]; then
         params+="--chain-id ${CFG_CHAIN_ID} "
 fi
-if [ -n "${JSON_RPC_ENDPOINT}" ]; then
-    params+="--json-rpc-endpoint ${JSON_RPC_ENDPOINT} "
+if [ -n "${CFG_JSON_RPC_ENDPOINT}" ]; then
```
davidiw, line 15: are these defined anywhere else? do we need to update some k8s or terraform?

config/config-builder/src/key_manager_config.rs
JoshLind: This ordering matches the order of the field definitions (see above).

config/config-builder/src/key_manager_config.rs
JoshLind: The chain id has more in common with the json rpc interface (i.e., it helps to specify which blockchain we're pointing to). The previous grouping with the time periods seemed very odd to me! Although this is a clarification that I don't feel super strongly about..

docker/safety-rules/key-manager.sh
JoshLind: 'cc @mgorven. Any impact to these files? I can find one reference but not sure if it's breaking.

mgorven: k8s doesn't use these entrypoint files. The keymanager config file will need to be updated: https://github.com/libra/partners/blob/master/validator/helm/files/configs/keymanager.yaml

JoshLind: Awesome, thanks! I'll update that repo too

JoshLind:

PR #5195, opened 2020-07-18
[config] Add chain-id to config-builder

davidiw:
Remove templates for config-builder based nodes, that's an anti-pattern
Add chain-id (CFG_CHAIN_ID) to all docker containers
Add chain-id to all relevant config(-builders)


config/config-builder/src/main.rs
```diff
@@ -51,6 +52,9 @@ enum FullNodeCommand {
 
 #[derive(Debug, StructOpt)]
 struct FullNodeArgs {
+    #[structopt(long)]
+    /// Defines which chain version libra will use
+    chain_id: Option<u8>,
```
zekun000, line 57: why not use the string representation?

terraform/templates/validator.yaml
```diff
@@ -1,5 +0,0 @@
-base:
-  chain_id: "${chain_id}"
-
-validator_network:
-  mutual_authentication: true
```
zekun000, line 5: is this the default for validator network?

terraform/fullnodes.tf
```diff
@@ -38,14 +38,6 @@ resource "aws_instance" "fullnode" {
 
 }
 
-data "template_file" "fullnode_config" {
-  template = file("templates/fullnode.yaml")
-
-  vars = {
```
zekun000, line 44: delete the base config too? also in the validator/fullnodes.json

config/config-builder/src/main.rs
davidiw: because then we have to use a fixed subset, which is antithetical to the entire goal of using the chain id for networks....

terraform/templates/validator.yaml
davidiw: https://github.com/libra/libra/blob/master/config/src/generator.rs#L38

terraform/fullnodes.tf
davidiw: that defines some properties, I think we keep that, I'm gonna use this push result to let me know what I'm doing wrong :p

config/config-builder/src/main.rs
christinacelee: fwiw the numeric value as a string also works (e.g. "2" and "TESTNET" both work)

davidiw: Let's just get this working for now and we can screw around with the semantics later!

terraform/templates/fullnode.json
```diff
@@ -20,7 +20,7 @@
             {"sourceVolume": "libra-data", "containerPath": "/opt/libra/data"}
         ],
         "environment": [
-            {"name": "CFG_BASE_CONFIG", "value": ${cfg_base_config}},
+            {"name": "CFG_CHAIN_ID", "value": ${cfg_chainIid}},
```
christinacelee, line 23: typo for `cfg_chain_id`!

terraform/templates/validator.json
```diff
@@ -36,7 +36,7 @@
             {"name": "CFG_FULLNODE_SEED", "value": "${cfg_fullnode_seed}"},
             {"name": "CFG_NUM_FULLNODES", "value": "${cfg_num_fullnodes}"},
         %{ endif }
-            {"name": "CFG_BASE_CONFIG", "value": ${cfg_base_config}},
+            {"name": "CFG_CHAIN_ID", "value": ${cfg_chainIid}},
```
christinacelee, line 39: typo here too

davidiw: oh my god....

docker/mint/docker-run.sh
```diff
@@ -6,10 +6,12 @@ set -ex
 export RUST_BACKTRACE=full
 
 declare -a params
+if [ -n "${CFG_CHAIN_ID}" ]; then
```
sausagee, line 9: That's it, the final piece!  It explained why we saw different waypoints last night.

types/src/chain_id.rs
```diff
@@ -21,8 +21,8 @@ pub enum NamedChain {
     MAINNET = 1,
     // The Libra chain below are non-reserved, non-production, and may change over time.  They are listed for convenience here.
     TESTNET,
-    DEVNET,
-    TESTING,
+    DEVNET = 5,
+    TESTING = 6,
```
ankushagarwal, line 25: @davidiw : Why was there a breaking change to chain_id numbers? Can we use the previous numbers? This breaks AOS and partners whom we gave the chain id of `4` for `TESTING`

PR #5203, opened 2020-07-20
[language] Fix error in `remaining_gas` where we were rounding up

### LINKED ISSUE #5169: [Bug]No charge gas for some script execute error
# 🐛 Bug

Gas should also be charged when a user fails to execute a script due to, for example, passing the wrong parameter.

## To reproduce

**Code snippet to reproduce**
Run follow test at e2e_tests

```rust
# Your code goes here
# Please make sure it does not require any external dependencies

#[test]
fn test_invalid_argument_txn_gas_charge() {
    let mut fake_executor = FakeExecutor::from_genesis_file();
    let sender = AccountData::new(1_000_000, 0);
    fake_executor.add_account_data(&sender);

    let mut txn_data = TransactionMetadata::default();
    txn_data.sender = *sender.address();
    txn_data.max_gas_amount = GasUnits::new(100_000);
    txn_data.gas_unit_price = GasPrice::new(1);

    // get a SignedTransaction
    let txn = sender.account().create_signed_txn_with_args(
        StdlibScript::PeerToPeerWithMetadata
            .compiled_bytes()
            .into_vec(),
        vec![lbr_type_tag()],
        //miss all argument.
        vec![],
        0,
        gas_costs::TXN_RESERVED, // this is a default for gas
        0,                       // this is a default for gas
        LBR_NAME.to_owned(),
    );

    let output = fake_executor.execute_transaction(txn);
    assert!(!output.status().is_discarded());
    assert!(output.gas_used() > 0);
}

```

expect output.gas_used > 0, but fail.

**Stack trace/error message**
```
// Paste the output here

assertion failed: output.gas_used() > 0
thread 'tests::failed_transaction_tests::test_invalid_argument_txn_gas_charge' panicked at 'assertion failed: output.gas_used() > 0', language/e2e-tests/src/tests/failed_transaction_tests.rs:94:5
stack backtrace:

```

## System information

**Please complete the following information:**
- 
- 
- 

latest master

tzakian:
Remaining gas is subtracted from the original gas price. We want to round up
the gas used which is `max_gas_amount - remaining` so we actually want to round down `remaining` instead of
rounding it up.

Closes #5169

## Tests
Added test in the bug report to make sure this is fixed.

### DIFF
--- a/language/e2e-tests/src/tests/verify_txn.rs
+++ b/language/e2e-tests/src/tests/verify_txn.rs
@@ -6,7 +6,7 @@ use crate::{
     assert_prologue_disparity, assert_prologue_parity,
     compile::compile_module_with_address,
     executor::FakeExecutor,
-    transaction_status_eq,
+    gas_costs, transaction_status_eq,
 };
 use compiled_stdlib::transaction_scripts::StdlibScript;
 use compiler::Compiler;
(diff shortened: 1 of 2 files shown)


tzakian:
Good spot! This was introduced in #5032 due to the fact that we were rounding up in `remaining_gas` from the internal gas units when in fact this number should be rounded down/truncated. #5203 fixes this issue.

PR #5204, opened 2020-07-20
[JSON-RPC] Add tiered/received mint event to DD account role view

moeziniaf:
## Motivation

Right now, there are MintEvents which are used for global mints. We would like to give a keyed indexed to DDs with the specific ability to track tiered mint events which deposit to their accounts.
Here, we expose this event key in the DD Role view to allow DDs and AOS to tail these events.
Additionally, we expose these events on the TransactionView in the get_transactions API

## Test Plan

get_transactions RPC

https://github.com/libra/libra/pull/5001 also edits the DD account view role

```curl http://localhost:60359 -X POST -H "Content-Type: application/json" --data '{"jsonrpc":"2.0","method":"get_account","params":["000000000000000000000000000000dd"],"id":1}'```

```..."role":{"designated_dealer":{"base_url":"https://libra.org","compliance_key":"93ea047fc05d19fa589593cdeaec2a0392a6bbe927f55eb2876c3349ddf0a752","expiration_time":18446744073709551615,"human_name":"moneybags","received_mints_events_key":"0000000000000000000000000000000000000000000000dd"}},... ```

### CODE REVIEW

File: json-rpc/types/src/views.rs [hunk 1]
```diff
@@ -64,6 +65,7 @@ pub enum AccountRoleView {
         base_url: String,
         expiration_time: u64,
         compliance_key: BytesView,
+        received_mints_key: BytesView,
```
thefallentree, line 68: received_mints_events_key

moeziniaf:
RFR!

json-rpc/types/src/views.rs
```diff
@@ -66,6 +67,7 @@ pub enum AccountRoleView {
         expiration_time: u64,
         compliance_key: BytesView,
         preburn_balances: Vec<AmountView>,
+        received_mints_events_key: BytesView,
```
tzakian, line 70: Nit: 
```suggestion
        received_mint_event_key: BytesView,
```
That or `received_mint_events_key` whichever makes more sense to you.

PR #5210, opened 2020-07-20
[cluster-test][node launcher] remove delete_data parameter.

changyonghu:
Related Issue: #5063

### DIFF
--- a/testsuite/cluster-test/src/cluster_swarm/cluster_swarm_kube.rs
+++ b/testsuite/cluster-test/src/cluster_swarm/cluster_swarm_kube.rs
@@ -153,7 +153,6 @@ impl ClusterSwarmKube {
         seed_peer_ip: &str,
         safety_rules_addr: &str,
         cfg_overrides: &str,
-        delete_data: bool,
     ) -> Result<Pod> {
         let cfg_fullnode_seed = if num_fullnodes > 0 {
             CFG_FULLNODE_SEED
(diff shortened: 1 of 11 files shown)


testsuite/cluster-test/src/cluster_swarm/cluster_swarm_kube.rs
```diff
@@ -642,16 +632,38 @@ impl ClusterSwarmKube {
...
-        &self,
-        instance_config: InstanceConfig,
-        delete_data: bool,
-    ) -> Result<Instance> {
-        self.upsert_node(instance_config, delete_data).await
+    async fn spawn_new_instance(&self, instance_config: InstanceConfig) -> Result<Instance> {
+        self.upsert_node(instance_config).await
+    }
+
+    async fn clean_data(&self, instance_config: &InstanceConfig) -> Result<()> {
```
andll, line 659: I think it would be better to change this method to just accept node, instead of instance.
In conjunction with your previous PR this also should be easy to call, since you will have a node in cluster builder

changyonghu: Yes, that is what I plan to do once the previous pr is approved.

changyonghu:
> Code looks good, but please confirm that delete_data actually deletes data:
> 
> * Run recovery experiment on master and on this PR
> * Observe state sync version counter, to confirm it dips to low values during recovery experiments on this PR. Use same experiment on master as a guideline on what this dip looks like

Confirmed.
Test Plan: 

> Code looks good, but please confirm that delete_data actually deletes data:
> 
> * Run recovery experiment on master and on this PR
> * Observe state sync version counter, to confirm it dips to low values during recovery experiments on this PR. Use same experiment on master as a guideline on what this dip looks like

Confirmed. 
Test Plan:
Add time delay during recovery_time experiment. Evidence:  /opt/libra/data/event.log decreased during restarting and time record of /opt/libra/data/fluent-bit.conf is updated during restarting.

PR #5213, opened 2020-07-20
Add libra_chain_id for all JSONRPC Response

thefallentree:
This will allow client to detect it has connect to the right fullnode.

### DIFF
--- a/json-rpc/src/runtime.rs
+++ b/json-rpc/src/runtime.rs
@@ -9,10 +9,10 @@ use crate::{
 use futures::future::join_all;
 use libra_config::config::{NodeConfig, RoleType};
 use libra_json_rpc_types::views::{
-    JSONRPC_LIBRA_LEDGER_TIMESTAMPUSECS, JSONRPC_LIBRA_LEDGER_VERSION,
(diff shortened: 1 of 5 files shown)


json-rpc/src/runtime.rs
```diff
@@ -93,6 +102,7 @@ async fn rpc_endpoint(
     data: Value,
     service: JsonRpcService,
     registry: Arc<RpcRegistry>,
+    chain_id: ChainId,
```
christinacelee, line 105: can we store chain_id in JsonRpcService instead?

thefallentree: yeah, good idea, fixed.

json-rpc/src/tests/unit_tests.rs
```diff
@@ -189,6 +189,7 @@ fn test_json_rpc_protocol() {
     let data: JsonMap = resp.json().unwrap();
     assert!(data.get(JSONRPC_LIBRA_LEDGER_VERSION).is_some());
     assert!(data.get(JSONRPC_LIBRA_LEDGER_TIMESTAMPUSECS).is_some());
+    assert!(data.get(JSONRPC_LIBRA_CHAIN_ID).is_some());
```
christinacelee, line 192: can we go further and check that chain_id == ChainId::test?

thefallentree: done.

PR #5216, opened 2020-07-20
[move prover] add constant support in specs

emmazzz:
## Motivation

This PR added support for constants in specifications so now users can use constants defined in the module in specs as well.

I also changed the source map for constants from `BTreeMap<TableIndex, Vec<ConstantName>>` to  `BTreeMap<ConstantName, TableIndex>`. This way it's easier to retrieve the value of a constant by its name.
 
Yes.

## Test Plan

Added a test case `consts.move`.

### DIFF
--- a/language/move-prover/spec-lang/src/env.rs
+++ b/language/move-prover/spec-lang/src/env.rs
@@ -200,6 +200,10 @@ pub struct FunId(Symbol);
 #[derive(Debug, PartialEq, Eq, PartialOrd, Ord, Hash, Clone, Copy)]
 pub struct SchemaId(Symbol);
 
+/// Identifier for a constant.
+#[derive(Debug, PartialEq, Eq, PartialOrd, Ord, Hash, Clone, Copy)]
+pub struct ConstId(Symbol);
+
 /// Identifier for a specification function, relative to module.
 #[derive(Debug, PartialEq, Eq, PartialOrd, Ord, Hash, Clone, Copy)]
 pub struct SpecFunId(RawIndex);
@@ -255,6 +259,16 @@ impl StructId {
 }
 
 impl FieldId {
+    pub fn new(sym: Symbol) -> Self {
+        Self(sym)
+    }
+
+    pub fn symbol(self) -> Symbol {
+        self.0
+    }
+}
+
+impl ConstId {
     pub fn new(sym: Symbol) -> Self {
         Self(sym)
     }
--- a/language/move-prover/spec-lang/src/translate.rs
+++ b/language/move-prover/spec-lang/src/translate.rs
@@ -19,7 +19,8 @@ use num::{BigUint, FromPrimitive, Num};
 use regex::Regex;
 
 use bytecode_source_map::source_map::SourceMap;
-use move_ir_types::location::Spanned;
+use move_core_types::value::MoveValue;
+use move_ir_types::{ast::ConstantName, location::Spanned};
 use move_lang::{
     compiled_unit::{FunctionInfo, SpecInfo},
     expansion::ast as EA,
@@ -30,7 +31,7 @@ use move_lang::{
 };
 use vm::{
     access::ModuleAccess,
-    file_format::{FunctionDefinitionIndex, StructDefinitionIndex},
+    file_format::{Constant, FunctionDefinitionIndex, StructDefinitionIndex},
     views::{FunctionHandleView, StructHandleView},
     CompiledModule,
 };
@@ -42,9 +43,9 @@ use crate::{
         SpecVarDecl, Value,
     },
     env::{
-        FieldId, FunId, FunctionData, GlobalEnv, Loc, ModuleId, MoveIrLoc, NodeId, QualifiedId,
-        SchemaId, SpecFunId, SpecVarId, StructData, StructId, TypeConstraint, TypeParameter,
-        CONDITION_GLOBAL_PROP, CONDITION_INJECTED_PROP, SCRIPT_BYTECODE_FUN_NAME,
+        ConstId, FieldId, FunId, FunctionData, GlobalEnv, Loc, ModuleEnv, ModuleId, MoveIrLoc,
+        NodeId, QualifiedId, SchemaId, SpecFunId, SpecVarId, StructData, StructId, TypeConstraint,
+        TypeParameter, CONDITION_GLOBAL_PROP, CONDITION_INJECTED_PROP, SCRIPT_BYTECODE_FUN_NAME,
     },
     project_1st, project_2nd,
     symbol::{Symbol, SymbolPool},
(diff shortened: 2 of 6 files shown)


language/move-prover/spec-lang/src/env.rs
```diff
@@ -599,13 +613,15 @@ impl GlobalEnv {
     /// Adds a new module to the environment. StructData and FunctionData need to be provided
     /// in definition index order. See `create_function_data` and `create_struct_data` for how
     /// to create them.
+    #[allow(clippy::too_many_arguments)]
```
emmazzz, line 616: Apparently clippy doesn't allow functions to have more than 13 arguments.

language/move-prover/spec-lang/src/env.rs
```diff
@@ -1940,3 +1976,58 @@ impl<'env> FunctionEnv<'env> {
...
+        self.data.type_.clone()
+    }
+
+    pub fn get_value(&self) -> Value {
+        let const_ = self.module_env.get_constant(self.data.idx);
+        let move_value = self.module_env.get_constant_value(const_);
+        Self::translate_from_move_value(&move_value)
+    }
+
+    fn translate_from_move_value(value: &MoveValue) -> Value {
```
emmazzz, line 2012: I'm not sure if here is the best place to put this function.

language/move-prover/spec-lang/src/env.rs
```diff
@@ -1940,3 +1976,58 @@ impl<'env> FunctionEnv<'env> {
...
+#[derive(Debug, Clone, PartialEq)]
+pub struct ConstantData {
+    pub loc: Loc,
+    pub name: Symbol,
+    pub type_: Type,
+    pub idx: ConstantPoolIndex,
+}
+
+#[derive(Debug, Clone)]
+pub struct ConstantEnv<'env> {
```
wrwg, line 1989: I don't think you actually need a ConstantEnv. The XEnv idiom is used if there is something to lookup about, like e.g. it's a struct, and its fields. You can just have

```
pub fn get_constant(&self, id: ConstantId) -> &ConstantDecl
pub fn get_constants(&self) -> &[ConstantDecl]
```

language/move-prover/spec-lang/src/env.rs
wrwg: I would put the ast::Value already in the ConstantDecl, then do the translation in `translate.rs`. There should be already functions inside this module to map MoveValue to ast::Value.

language/move-prover/spec-lang/src/translate.rs
```diff
@@ -140,6 +142,14 @@ struct FunEntry {
     result_type: Type,
 }
 
+#[derive(Debug, Clone)]
+struct ConstEntry {
+    loc: Loc,
+    module_id: ModuleId,
+    const_id: ConstId,
+    signature: Type,
```
wrwg, line 150: Do not call types 'signature', we don't use that naming in this code. Call it `ty: Type`.

Also I'm wondering why we don't inline constants already here in translate.rs, and never add them to the GlobalEnv. This would be much easier:

1. Attach the translated ast::Value to ConstEntry
2. Populate the ConstEntry during decl_ana.
3. During def_ana, when expressions are translated, if you come along a name which resolves as a ConstEntry, just inline its value.

There is little merit to us to have the ConstDecls available in subsequent phases. This would simplify this implementation significantly.

language/move-prover/spec-lang/src/env.rs
```diff
@@ -1069,6 +1089,22 @@ impl<'env> ModuleEnv<'env> {
             })
     }
 
+    /// Returns iterator over ConstantEnvs in this module.
+    pub fn get_constants(&'env self) -> impl Iterator<Item = ConstantEnv<'env>> {
+        self.clone().into_constants()
+    }
+
+    /// Returns iterator over FunctionEnvs in this module.
```
wrwg, line 1097: comment c&p

language/move-prover/spec-lang/src/env.rs
emmazzz: I don't actually see such function in `translate.rs`. `translate.rs` deals with values of type `EA::Value` and translates them to `ast::Value`.

language/move-prover/spec-lang/src/translate.rs
emmazzz: Yeah we could do what you suggested. I thought maybe it would be nice to still have the names of the constants at Boogie level. Since you think there's little merit to it, I'll change it.

wrwg: Note, just a suggestion.

language/move-prover/spec-lang/src/env.rs
wrwg: There is ExprTranslator::translate_value.

emmazzz: `ExpTranslator:: translate_value` translates `EA::Value`, not `MoveValue`

language/move-prover/spec-lang/src/translate.rs
emmazzz: Right I think your suggestion makes sense. There's no need to carry constant name all the way down.

language/move-prover/spec-lang/src/translate.rs
```diff
@@ -900,18 +934,57 @@ impl<'env, 'translator> ModuleTranslator<'env, 'translator> {
...
 
+    fn decl_ana_const(
+        &mut self,
+        name: &PA::ConstantName,
+        def: &EA::Constant,
+        compiled_module: &CompiledModule,
+        source_map: &SourceMap<MoveIrLoc>,
+    ) {
+        let qsym = self.qualified_by_module_from_name(&name.0);
+        let name = self.symbol_pool().make(&name.0.value);
```
wrwg, line 965: nit: let name = qsym.symbol;

language/move-prover/spec-lang/src/translate.rs
```diff
@@ -4105,6 +4186,28 @@ impl<'env, 'translator, 'module_translator> ExpTranslator<'env, 'translator, 'mo
...
+            MoveValue::U128(n) => Value::Number(BigUint::from_u128(*n).unwrap()),
+            MoveValue::Bool(b) => Value::Bool(*b),
+            MoveValue::Address(a) => Value::Address(ModuleEnv::addr_to_big_uint(a)),
+            MoveValue::Signer(a) => Value::Address(ModuleEnv::addr_to_big_uint(a)),
+            MoveValue::Vector(vs) => {
+                let b = vs
+                    .iter()
+                    .map(|v| match v {
+                        MoveValue::U8(n) => *n,
+                        _ => unimplemented!("Not yet supported constant vector value: {:?}", v),
```
wrwg, line 4203: Can this happen in reality? In that case we should do here `env.error` and not crash.

BTW, a good example if to report an error while I nearly never use static methods. As soon as you need more context for your functions you are stuck with static methods. If this would have a &self, you can add this kind of thing without any hassle.

language/move-prover/src/spec_translator.rs
```diff
@@ -1109,7 +1109,7 @@ impl<'env> SpecTranslator<'env> {
         match exp {
             Exp::Value(node_id, val) => {
                 self.set_writer_location(*node_id);
-                self.translate_value(*node_id, val);
+                self.translate_value(val);
```
wrwg, line 1112: ... I also pass information to functions even if its not needed right now. Say translate_value would need to report an error, then it needs node_id to get the location where to report the error...

PR #5217, opened 2020-07-20
[ci] Enable LTO in release builds

### LINKED ISSUE #4987: [Bug] Release/CI builds don't have LTO
# 🐛 Bug

#4829 disabled LTO by default. We need to keep this default for developer ergonomics but ensure that release builds and CI builds include LTO. This can be done with Cargo's profile dependencies (https://github.com/rust-lang/rfcs/blob/master/text/2282-profile-dependencies.md) which is now stabilized and shipping as of Rust 1.43.

metajack:
LTO was disabled by default in #4829. This PR overrides the LTO settings in CI for the release profile. This will keep local developers building quickly without LTO while still testing real release conditions in CI.

Fixes #4987

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -53,7 +53,6 @@ commands:
             echo 'export CI_TIMEOUT="timeout 40m"' >> $BASH_ENV
             export RUST_NIGHTLY=$(cat cargo-toolchain)
             echo 'export RUST_NIGHTLY='${RUST_NIGHTLY} >> $BASH_ENV
-
   install_deps:
     steps:
       - run:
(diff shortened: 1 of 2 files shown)


andll:
One note - we were using thin lto(lto=thin, instead of true) previously, this seemed like a good choice between TPS and compile speed, although I don't have exact numbers

thefallentree:
> One note - we were using thin lto(lto=thin, instead of true) previously, this seemed like a good choice between TPS and compile speed, although I don't have exact numbers

thin was chosen mainly to enable fast development cycles, but for CI it doesn't matter , and fat LTO gives more performance back.

thefallentree:
the error looks like an OOM issue.

andll:
To the best of my knowledge, this PR does not affect performance of prod at all, this is just about setting up LTO in unit tests

It is also not true that 'performance does not matter for CI', because CI blocks diffs landing. We do consider land time to be important, for example in cluster test land blocking, I don't understand why do we have different view here

metajack:
In the interest of streamlining CI, I moved the flag to build-common.sh which will enable LTO for all release binaries, and removed all our CI workflow that builds unused release stuff. LBT will still catch build failures under --release with LTO on.

metajack:
20% to TPS it seems. Yay :)

cc @andll

PR #5219, opened 2020-07-20
[management] rename storage helper commands to match reality

gregnazario:
Realistically, the arg parser doesn't care what the name of the binary
is in any of the commands, but it does need one.  This will update it to
look like the real thing.

This is really just a cosmetic change.


gregnazario:
> thanks!
> 
> we'll need to revisit the readmes too
> 
> and then poke Michael

I have the READMEs in one commit for all three crates once I push the other PR.

Issue #5117, opened 2020-07-15
[validator-set] Remaining work items. Updated.


valerini:
This issue replaces the old one [Issue 4676](https://github.com/libra/libra/issues/4676).

1) [ ] Move key rotation capability for validator owner to recovery address (to decide if this is a separate transaction script: `create_validator_account_with_delegated_key_rotation`?)

2) [ ] Make `ValidatorConfig::set_operator` method callable by LibraRoot with a possibility to reclaim the ability to call it by the validator owner when it starts to custody its own key (straightforward approach is to have a boolean flag under the owner account indicating who can call the set_operator function - whether this is a validator owner or the LibraRoot).

3) [ ] Add human name to the `ValidatorConfig` module and to transaction scripts (to decide if we then need a ValidatorOperatorConfig module to have a place to hold the name).

4) [ ] Add rate limits to function `LibraSystem::set_validator_config_and_reconfigure(..)` to make sure a faulty operator is not abusing the system (probably at most once in 5 min).

5) [ ] Rename functions and matching scripts: `LibraSystem::add_validator` -> `LibraSystem::add_validator_and_reconfigure`; `LibraSystem::remove_validator -> LibraSystem::remove_validator_and_reconfigure`; `ValidatorConfig::set_validator_config -> ValidatorConfig::register_validator_config`.

PR #5224, opened 2020-07-21
[validator-set] Add permissions to LibraRoot

valerini:
Draft, not ready for review yet: currently discussing whether we want to do it with this mechanisms or with admin scripts (those scripts can be run sent libra root but be executed as if they were signed by another account - these are effectively write-set transactions).

This PR resolves to sub-cases of the [Issue #5117](https://github.com/libra/libra/issues/5117):
1) it makes set_operator callable by both the libra_root and the validator owner. Once all of the owners migrate to custodying their own keys, we can remove the libra_root from being able to call the set_operator function.
2) Make libra_root be able to rotate the validator owner's authentication key in order to allow the AOS to initially create validator owners' accounts with dummy auth keys:
    * the RecoveryAddress resource is instantiated in LibraRoot account at genesis,
    * the `LibraAccount::create_validator_account_and_extract_key_rotation_cap` function is added to create an account and immediately extract the key rotation capability and return it back to the caller; the create_validator_account script calls this function and places the returned capability to the RecoveryAddress of LibraRoot.

The solution is not very elegant from the move side. Any suggestion on how to improve it are very welcome!
The account should still have an auth key (it can be dummy), but it leaves the possibility for someone to crack into that key and take advantage of the account. It is especially important to create the key in a way that makes sure no secret key is known for it. It also seems beneficial to have different dummy auth keys for different validators, otherwise cracking into one auth-key will make it possible to break the rest of them.

Yes

## Test Plan

cargo xtest


sblackshear:
>  It is especially important to create the key in a way that makes sure no secret key is known for it. It also seems beneficial to have different dummy auth keys for different validators, otherwise cracking into one auth-key will make it possible to break the rest of them.

I think we can addresses this by using the @JoshLind scheme to generate the validator addrs, but then use 0x0...0 for the auth key prefix of each validator?

PR #5227, opened 2020-07-21
[ci][circleci] fix rustup command

mimoo:


mimoo:
that's a fix : o not a readability change

@bors-libra r=rexhoffman

rexhoffman:
@mimoo with both rust versions installed it works both ways, without the nightly version it fails both ways.   Tested on my local box.

mimoo:
ah... nevermind then :/

PR #5228, opened 2020-07-21
[management] refactor validator_config

davidiw:
This merges shared logic more effectively in
management::validator_config and unifies logic within
operational::validator_config

also owner name / account is not expected in json rpc, so that was
removed

There might be some flakiness on tests, but the code is pretty much done.

### DIFF
--- a/config/management/genesis/src/validator_config.rs
+++ b/config/management/genesis/src/validator_config.rs
@@ -1,123 +1,60 @@
 // Copyright (c) The Libra Core Contributors
 // SPDX-License-Identifier: Apache-2.0
 
-use libra_config::config::HANDSHAKE_VERSION;
-use libra_global_constants::OWNER_ACCOUNT;
+use libra_global_constants::{OWNER_ACCOUNT, OWNER_KEY};
 use libra_management::{
-    constants,
-    error::Error,
-    secure_backend::{SharedBackend, ValidatorBackend},
-    storage::StorageWrapper,
-    validator_config::{
-        create_transaction, create_validator_config_script, encode_address,
-        encode_validator_address, fetch_keys_from_storage, owner_from_key,
-    },
+    constants, error::Error, secure_backend::SharedBackend, storage::StorageWrapper,
 };
-use libra_network_address::{
-    encrypted::{TEST_SHARED_VAL_NETADDR_KEY, TEST_SHARED_VAL_NETADDR_KEY_VERSION},
-    NetworkAddress,
-};
+use libra_network_address::NetworkAddress;
 use libra_secure_storage::Value;
-use libra_types::{chain_id::ChainId, transaction::Transaction};
+use libra_types::{account_address, transaction::Transaction};
 use structopt::StructOpt;
 
-// TODO(davidiw) add operator_address, since that will eventually be the identity producing this.
 #[derive(Debug, StructOpt)]
 pub struct ValidatorConfig {
     #[structopt(long)]
     owner_name: String,
+    #[structopt(flatten)]
+    validator_config: libra_management::validator_config::ValidatorConfig,
     #[structopt(long)]
     validator_address: NetworkAddress,
     #[structopt(long)]
     fullnode_address: NetworkAddress,
-    #[structopt(long)]
-    chain_id: ChainId,
-    #[structopt(flatten)]
-    validator_backend: ValidatorBackend,
     #[structopt(flatten)]
     shared_backend: SharedBackend,
 }
 
 impl ValidatorConfig {
     pub fn execute(self) -> Result<Transaction, Error> {
-        let storage_name = self.validator_backend.name();
-        let sequence_number = 0u64;
-        // Only supports one address for now
-        let idx_addr = 0u32;
-
-        // Fetch the owner key from remote storage using the owner_name and derive an address
-        let owner_account = owner_from_key(
-            storage_name,
-            &self.shared_backend.shared_backend,
+        // Retrieve and set owner account
+        let owner_shared_storage = StorageWrapper::new_with_namespace(
+            self.shared_backend.name(),
             self.owner_name,
+            &self.shared_backend.shared_backend,
         )?;
+        let owner_key = owner_shared_storage.ed25519_key(OWNER_KEY)?;
+        let owner_account = account_address::from_public_key(&owner_key);
 
-        // Fetch keys from storage
-        let (consensus_key, fullnode_network_key, validator_network_key) =
-            fetch_keys_from_storage(storage_name, &self.validator_backend.validator_backend)?;
-
-        // Append ln-noise-ik and ln-handshake protocols to base network addresses
-        // and encrypt the validator address.
-        let validator_address = self
-            .validator_address
-            .append_prod_protos(validator_network_key, HANDSHAKE_VERSION);
-        let raw_enc_validator_address = encode_validator_address(
-            validator_address,
-            &owner_account,
-            sequence_number,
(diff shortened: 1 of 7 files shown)


config/management/genesis/src/validator_config.rs
```diff
@@ -1,123 +1,60 @@
...
-        let mut validator_storage =
-            StorageWrapper::new(storage_name, &self.validator_backend.validator_backend)?;
-        validator_storage.set(
-            constants::VALIDATOR_CONFIG,
-            Value::Transaction(validator_config_tx.clone()),
+        let txn = self.validator_config.build_transaction(
+            0,
+            self.fullnode_address,
+            self.validator_address,
+            false,
```
gregnazario, line 48: This is sometimes contentious in languages, but I really wish that you could put here `reconfiguration=false`.  But, I don't think rust supports that

config/management/src/validator_config.rs
```diff
@@ -1,202 +1,127 @@
...
+        let validator_network_key = storage.x25519_public_from_private(VALIDATOR_NETWORK_KEY)?;
+
+        // Build Validator address including protocols and encryption
+        // Append ln-noise-ik and ln-handshake protocols to base network addresses
+        // and encrypt the validator address.
+        let validator_address =
+            validator_address.append_prod_protos(validator_network_key, HANDSHAKE_VERSION);
+        let raw_validator_address = encode_address(validator_address)?;
+        // Only supports one address for now
+        let addr_idx = 0u32;
```
gregnazario, line 59: Nit: I realized this later, I think we can just put 0 and it'll figure out the size :). I just forgot to fix it

config/management/src/validator_config.rs
```diff
@@ -1,202 +1,127 @@
...
+        let key = TEST_SHARED_VAL_NETADDR_KEY;
+        let version = TEST_SHARED_VAL_NETADDR_KEY_VERSION;
+        let enc_validator_address = raw_validator_address.encrypt(
+            &key,
+            version,
+            &owner_account,
+            sequence_number + 1,
```
gregnazario, line 66: ngl, this took me a while to figure out.  If only we could make this function later (current_sequence_number) instead of the sequence number we expect it to be.  But, that might cause a whole other layer of issues

config/management/src/validator_config.rs
```diff
@@ -1,202 +1,127 @@
...
+        );
+        let raw_enc_validator_address = RawEncNetworkAddress::try_from(&enc_validator_address)
+            .map_err(|e| {
+                Error::UnexpectedError(format!(
+                    "error serializing encrypted address: '{:?}', error: {}",
+                    enc_validator_address, e
+                ))
+            })?;
+
+        // Build Fullndoe address including protocols
```
gregnazario, line 77: *Fullnode

config/management/src/validator_config.rs
```diff
@@ -1,202 +1,127 @@
...
+        let fullnode_address =
+            fullnode_address.append_prod_protos(fullnode_network_key, HANDSHAKE_VERSION);
+        let raw_fullnode_address = encode_address(fullnode_address)?;
+
+        // Generate the validator config script
+        // TODO(philiphayes): remove network identity pubkey field from struct
```
gregnazario, line 83: What's this all about btw?  I copied it, then you copied it.

config/management/src/validator_config.rs
```diff
@@ -1,202 +1,127 @@
...
+        let consensus_key = storage.ed25519_public_from_private(CONSENSUS_KEY)?;
+        let fullnode_network_key = storage.x25519_public_from_private(FULLNODE_NETWORK_KEY)?;
+        let validator_network_key = storage.x25519_public_from_private(VALIDATOR_NETWORK_KEY)?;
+
+        // Build Validator address including protocols and encryption
+        // Append ln-noise-ik and ln-handshake protocols to base network addresses
+        // and encrypt the validator address.
+        let validator_address =
+            validator_address.append_prod_protos(validator_network_key, HANDSHAKE_VERSION);
+        let raw_validator_address = encode_address(validator_address)?;
```
gregnazario, line 57: Will this work instead of the `rotate_noise_public_key`?

config/management/genesis/src/validator_config.rs
davidiw: no, no it doesn't, if you are really anal, you can make enums :)

config/management/operational/src/validator_config.rs
```diff
@@ -1,173 +1,144 @@
...
+            | Protocol::Dns6(_)
+            | Protocol::Ip4(_)
+            | Protocol::Ip6(_)
+            | Protocol::Memory(_)
+            | Protocol::Tcp(_) => true,
+            _ => false,
+        })
+        .cloned()
+        .collect();
+    Ok(NetworkAddress::new(protocols))
```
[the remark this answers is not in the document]
davidiw: I kinda feel like all of this code belongs under network, but that feels like another PR so we don't end up modifying too many crates here. Since we seem to agree, I'll follow up in another commit / PR.

config/management/src/storage.rs
```diff
@@ -50,6 +51,14 @@ impl StorageWrapper {
             .map_err(|e| Error::StorageReadError(self.storage_name, name, e.to_string()))
     }
 
+    pub fn account_address(&self, name: &'static str) -> Result<AccountAddress, Error> {
+        let value = self
+            .value(name)?
+            .string()
+            .map_err(|e| Error::StorageReadError(self.storage_name, name, e.to_string()))?;
```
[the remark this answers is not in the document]
davidiw: woops

config/management/src/validator_config.rs
davidiw: ahh, yeah, I wondered!

config/management/src/validator_config.rs
davidiw: yup, remember the smoke test still passes :)

config/management/src/validator_config.rs
davidiw: ahh, yeah, we should add a comment on why we select this magic number... :D

config/management/src/validator_config.rs
davidiw: ha

config/management/src/validator_config.rs
davidiw: the *network_keys below need to be removed from various configs, scripts, and smart contracts, I pinged @phlip9 to try to get it out in this cut, since this is a breaking change.

network/network-address/src/lib.rs
```diff
@@ -270,7 +270,7 @@ impl Arbitrary for RawNetworkAddress {
 ////////////////////
 
 impl NetworkAddress {
-    fn new(protocols: Vec<Protocol>) -> Self {
+    pub fn new(protocols: Vec<Protocol>) -> Self {
```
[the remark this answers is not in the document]
phlip9: ahh that's because it's trying to maintain the invariant that `protocols` is always non-empty. that's why `fn new` is private and only `TryFrom<Vec<Protocol>>` is exposed, which will return an `EmptyError` if the input is empty.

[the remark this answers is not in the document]
phlip9: damn, i thought i documented this... but it seems i missed it :sob:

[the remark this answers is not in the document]
gregnazario: Interesting, we could enforce the invariant here on construction?

PR #5229, opened 2020-07-21
[secure] add a vault network_timeout_ms

agclement:
## Motivation

add a configurable timeout to Vault.

yes

## Test Plan

Don't break unit tests

## Related PRs

### DIFF
--- a/config/config-builder/src/key_manager_config.rs
+++ b/config/config-builder/src/key_manager_config.rs
@@ -5,6 +5,8 @@ use anyhow::Result;
 use libra_config::config::{KeyManagerConfig as KMConfig, SecureBackend, Token, VaultConfig};
 use libra_types::chain_id::ChainId;
 
+const DEFAULT_NETWORK_TIMEOUT_MS: u64 = 30_000;
+
 pub struct KeyManagerConfig {
     pub rotation_period_secs: Option<u64>,
     pub sleep_period_secs: Option<u64>,
@@ -15,6 +17,7 @@ pub struct KeyManagerConfig {
 
     pub vault_host: String,
     pub vault_namespace: Option<String>,
+    pub vault_network_timeout_ms: Option<u64>,
     pub vault_token: String,
 
     pub template: KMConfig,
@@ -31,6 +34,7 @@ impl Default for KeyManagerConfig {
             json_rpc_endpoint: template.json_rpc_endpoint.clone(),
             vault_host: "127.0.0.1:8200".to_string(),
             vault_namespace: None,
+            vault_network_timeout_ms: Some(DEFAULT_NETWORK_TIMEOUT_MS),
             vault_token: "root_token".to_string(),
             template,
         }
@@ -52,6 +56,9 @@ impl KeyManagerConfig {
             server: self.vault_host.clone(),
             token: Token::FromConfig(self.vault_token.clone()),
             renew_ttl_secs: None,
+            network_timeout_ms: self
+                .vault_network_timeout_ms
+                .unwrap_or(DEFAULT_NETWORK_TIMEOUT_MS),
         });
 
         if let Some(rotation_period_secs) = &self.rotation_period_secs {
--- a/config/config-builder/src/validator_config.rs
+++ b/config/config-builder/src/validator_config.rs
@@ -41,6 +41,7 @@ pub struct ValidatorConfig {
     pub safety_rules_token: Option<String>,
     pub seed: [u8; 32],
     pub template: NodeConfig,
+    pub vault_network_timeout_ms: u64,
 }
 
 impl Default for ValidatorConfig {
@@ -61,6 +62,7 @@ impl Default for ValidatorConfig {
             safety_rules_token: None,
             seed: DEFAULT_SEED,
             template: NodeConfig::default(),
+            vault_network_timeout_ms: 30_000,
         }
     }
 }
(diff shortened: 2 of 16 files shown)


config/management/src/secure_backend.rs
```diff
@@ -132,6 +132,8 @@ impl TryInto<config::SecureBackend> for SecureBackend {
                     server,
                     ca_certificate: certificate,
                     token: Token::FromDisk(PathBuf::from(token)),
+                    // TODO(davidiw):  parse the network_timeout_ms from the parameter list.
+                    network_timeout_ms: 30_000,
```
agclement, line 136: @davidiw It's not clear to me how this field should go "up" from here.  A follow on commit/PR is needed to get this to the point of editable by an operator.

davidiw: I don't think we need a todo here, I cannot imagine tweaks, but we always support adding optional parameters if we really want that.

config/src/config/secure_backend_config.rs
```diff
@@ -45,6 +45,8 @@ pub struct VaultConfig {
     pub server: String,
     /// The authorization token for accessing secrets
     pub token: Token,
+    /// Network Timeout in milliseconds.
+    pub network_timeout_ms: u64,
```
davidiw, line 49: nit alpha ordeing

agclement:
@davidiw alpha nit fixed.   TODO eliminated;  it accomplished the primary purpose of determining further changes not needed :)

config/config-builder/src/key_manager_config.rs
```diff
@@ -51,6 +53,7 @@ impl KeyManagerConfig {
             namespace: self.vault_namespace.clone(),
             server: self.vault_host.clone(),
             token: Token::FromConfig(self.vault_token.clone()),
+            network_timeout_ms: self.vault_network_timeout_ms,
```
davidiw, line 56: maybe there's something quirky happening here but this isn't in alpha ordering

config/config-builder/src/validator_config.rs
```diff
@@ -41,6 +41,7 @@ pub struct ValidatorConfig {
     pub safety_rules_token: Option<String>,
     pub seed: [u8; 32],
     pub template: NodeConfig,
+    pub vault_network_timeout_ms: u64,
```
davidiw, line 44: I think we can just use the default, no need to expose this... how did we do the default for safety rules?

config/src/config/secure_backend_config.rs
```diff
@@ -211,6 +216,7 @@ vault:
                 server: "127.0.0.1:8200".to_string(),
                 ca_certificate: None,
                 token: Token::FromDisk(PathBuf::from("/token")),
+                network_timeout_ms: 5_000,
```
davidiw, line 219: same here

config/src/config/secure_backend_config.rs
```diff
@@ -195,6 +199,7 @@ vault:
     server: "127.0.0.1:8200"
     token:
         from_config: "test"
+    network_timeout_ms: 5000
```
davidiw, line 202: here

config/src/config/secure_backend_config.rs
```diff
@@ -219,6 +225,7 @@ vault:
     server: "127.0.0.1:8200"
     token:
         from_disk: "/token"
+    network_timeout_ms: 5000
```
davidiw, line 228: here

config/src/config/secure_backend_config.rs
agclement: The default is 30_000;  the 5_000 that shows up here is suitable for test usage only.

config/src/config/secure_backend_config.rs
agclement: Relevant if we're parsing the config from text;  either the field is present in vault_config in which case we should expose the value or not.

config/config-builder/src/validator_config.rs
agclement: We should use the default.  The question is if we want to hardcode the default value on line 226 or hardcode the default value on line 65.  I prefer hardcoding that value in `default()`, but leave that as a judgement call for you.

agclement:
Applied one nit;  mild pushback on the other two.  One more round of discussion and I think we'll be done ...

davidiw:
Ugh, looks like I dropped the ball here... will possibly commandeer unless you want to bump this again?

agclement:
@davidiw  hopefully this resolves conflicts and passes tests.

config/config-builder/src/key_manager_config.rs
```diff
@@ -15,6 +15,7 @@ pub struct KeyManagerConfig {
 
     pub vault_host: String,
     pub vault_namespace: Option<String>,
+    pub vault_network_timeout_ms: u64,
```
davidiw, line 18: seems that this either needs to be optional or defined everywhere and one of the config doesn't have it defined... probably best to make this optional and then an unwrap_or(DEFAULT_VAULT_NETWORK_TIMEOUT_MS)

agclement:
@davidiw problems solved!

config/src/config/secure_backend_config.rs
```diff
@@ -45,6 +45,8 @@ pub struct VaultConfig {
     /// renewed up to that many seconds more. If this is not specified, the lease will not
     /// automatically be renewed.
     pub renew_ttl_secs: Option<u32>,
+    /// Network Timeout in milliseconds.
+    pub network_timeout_ms: u64,
```
davidiw, line 49: I think making this non-optional is dangerous as it breaks existing configs that are outside the scope of this code base :/. I'll try to commandeer. Thanks for pushing on this!

davidiw:
we have something in place now, closing out old PR

PR #5230, opened 2020-07-21
[crypto] Check both backend features are not enabled simultaneously

huitseeker:
This would typically happen with somebody compiling with `--features
"vanilla"` rather than `--features "vanilla" --no-default-features`.

This is a follow-up to #4970, which made a feature control how we load dalek dependencies, and #5126 and following, which enable the new resolver and therefore help features have real mutually-exclusive semantics.


mimoo:
isn't this going not to work due to our use of `--all-features` in scripts?

huitseeker:
It's my understanding #5126 has obviated the need for `--all-features` and why I added @metajack in review. Note this passed CI.

PR #5231, opened 2020-07-21
[language] Clean up of transaction argument tests

dariorussi:
Made them more explicit and with some comment.

Also changed the name of `SignatureToken::is_constant` to `SigantureToken::is_valid_for_constant`


dariorussi:
I wanted to add some tricky ones with bogus values but it's not easy to build a bogus value. So I left the test as they are with the minor corrections reported.
Considering this is only called internally from the adapter it seems ok.
Let me know if you spot tests you'd like to see added

tnowacki:
> Let me know if you spot tests you'd like to see added

Nothing really jumped to mind. Maybe some crazy nested vector? Otherwise should be good

PR #5232, opened 2020-07-21
[config] propagate chain id in to configs

davidiw:


thefallentree:
wow, this is not set?  then what is the default value? why didn't it fail to start  at all?

docker/mint/server.py
```diff
@@ -68,7 +68,7 @@ def send_transaction():
             "a m {} {} {} use_base_units".format(auth_key, amount, currency_code))
         application.client.expect("Mint request submitted", timeout=2)
 
-        application.client.sendline("a la")
```
thefallentree, line 71: can we verify what the correct behavior is for faucet?  I think this use to return "next seq" , but it is returning "current seq" now , which changed faucet behavior

PR #5235, opened 2020-07-21
[json-rpc] make get_metadata parameter version optional

xli:
## Motivation

JSON-RPC method get_metadata supports optional parameter version, but requires client sends params: [null] in the request body.

1. This diff improves the API to accept: params:[] as request body when client does not send version param.
2. As we don't support optional params before, this change also changed how we validate request params, and response an InvalidArgument data when client sends params less than required params size or more than required + optional params size.
3. When client sends invalid arguments, we should not response server error, changed to return invalid params error (code: -32602) with context information as error data.

Note: the optional solution in this diff is very limited, we probably should improve it later by accepting named arguments e.g. {fieldName: fieldValue}. But for get_metadata should be good enough.

## Test Plan

1. Unit tests are updated.
2. Local manual test, error message when send more params: 
```
curl -X POST -H "Content-Type: application/json" --data '{"jsonrpc":"2.0","method":"get_metadata","params":[2,3],"id":1}' http://localhost:60565

{"error":{"code":-32602,"data":{"InvalidArguments":{"given":2,"optional":1,"required":0}},"message":"Invalid params"},"id":1,"jsonrpc":"2.0","libra_chain_id":4,"libra_ledger_timestampusec":1595362466550086,"libra_ledger_version":215}
```

### DIFF
--- a/json-rpc/src/methods.rs
+++ b/json-rpc/src/methods.rs
@@ -3,7 +3,7 @@
 
 //! Module contains RPC method handlers for Full Node JSON-RPC interface
 use crate::{
-    errors::JsonRpcError,
+    errors::{ErrorData, InvalidArguments, JsonRpcError},
     views::{
         AccountStateWithProofView, AccountView, BlockMetadata, CurrencyInfoView, EventView,
         StateProofView, TransactionView,
@@ -76,10 +76,19 @@ pub(crate) struct JsonRpcRequest {
 }
 
 impl JsonRpcRequest {
-    /// Returns the request parameter at the given index. Note: this method should not panic with
-    /// an out of bounds error, as the number of request parameters has already been checked.
+    /// Returns the request parameter at the given index.
+    /// Returns Null if given index is out of bounds.
     fn get_param(&self, index: usize) -> Value {
-        self.params[index].clone()
+        self.get_param_with_default(index, Value::Null)
+    }
+
+    /// Returns the request parameter at the given index.
+    /// Returns default Value if given index is out of bounds.
+    fn get_param_with_default(&self, index: usize, default: Value) -> Value {
+        if self.params.len() > index {
+            return self.params[index].clone();
+        }
+        default
     }
 
     fn version(&self) -> u64 {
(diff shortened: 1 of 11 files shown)


json-rpc/src/methods.rs
```diff
@@ -356,29 +365,32 @@ async fn get_network_status(service: JsonRpcService, _request: JsonRpcRequest) -
...
 /// To register new RPC method, add it via `register_rpc_method!` macros call
 /// Note that RPC method name will equal to name of function
+#[allow(unused_comparisons)]
 pub(crate) fn build_registry() -> RpcRegistry {
     let mut registry = RpcRegistry::new();
-    register_rpc_method!(registry, "submit", submit, 1);
-    register_rpc_method!(registry, "get_metadata", get_metadata, 1);
-    register_rpc_method!(registry, "get_account", get_account, 1);
-    register_rpc_method!(registry, "get_transactions", get_transactions, 3);
+    register_rpc_method!(registry, "submit", submit, 1, 0);
```
thefallentree, line 371: I feel it make it more readable to design it as  total params & optional params  instead.   otherwise it is really hard to see how many arguments 

    register_rpc_method!(registry, "submit", submit, 1, 1);
    register_rpc_method!(registry, "get_metadata", get_metadata, 1, 0);
    get_transactions", get_transactions, 3, 3)

xli: To me, required arguments size and optional arguments size are metadata of the method, hence prefer to set them directly.
It probably does not matter as long as we pass 2 numbers, someone will feel confusing.
So you always need to read register_rpc_method macro to understand what they are.

What I'd like to add later is something like: Argument{"fieldName": {"Desc": "...", "required": true/false, "type": ....}}
Then we can generate our document from this definition, or we should generate register code from a definition that can be used to generate document.

thefallentree: ok, if we could generate these from an rust Request struct with serde-json tags that would be awesome (we can even set "rename" through serde tags)

PR #5238, opened 2020-07-22
[types/mempool] Update roles so that they are ranked in mempool

tzakian:
Right now we classify all transactions as either "governance" or "non-governance" where governance transactions are prioritized over all non-governance transactions. However, within governance there is no prioritization but we should, in particular LibraRoot and TC transactions should be prioritized over all other transactions including other governance-related transactions. 

This PR introduces this ranking for governance transactions.

This should not break client behavior when interacting with the system.

### DIFF
--- a/language/libra-vm/src/libra_transaction_validator.rs
+++ b/language/libra-vm/src/libra_transaction_validator.rs
@@ -11,7 +11,8 @@ use libra_types::{
     account_config::{self, RoleId},
     on_chain_config::{LibraVersion, VMConfig},
     transaction::{
-        SignatureCheckedTransaction, SignedTransaction, TransactionPayload, VMValidatorResult,
+        GovernanceRole, SignatureCheckedTransaction, SignedTransaction, TransactionPayload,
+        VMValidatorResult,
     },
     vm_status::{convert_prologue_runtime_error, StatusCode, VMStatus},
 };
@@ -22,10 +23,6 @@ use move_core_types::{
 };
 
 use move_vm_types::gas_schedule::CostStrategy;
-
-/// Any transation sent from an account with a role id below this cutoff will be priorited over
-/// other transactions.
-const PRIORITIZED_TRANSACTION_ROLE_CUTOFF: u64 = 5;
 
 #[derive(Clone)]
 pub struct LibraVMValidator(LibraVMImpl);
@@ -107,7 +104,7 @@ impl VMValidator for LibraVMValidator {
                     return VMValidatorResult::new(
                         Some(StatusCode::INVALID_GAS_SPECIFIER),
                         gas_price,
-                        false,
+                        GovernanceRole::NonGovernanceRole,
                     )
                 }
             };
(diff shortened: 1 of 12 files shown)


types/src/transaction/mod.rs
```diff
@@ -592,16 +592,52 @@ impl From<VMStatus> for TransactionStatus {
...
+    }
+}
+
 /// The result of running the transaction through the VM validator.
 #[derive(Clone, Debug, Eq, PartialEq)]
 pub struct VMValidatorResult {
     status: Option<DiscardedVMStatus>,
     score: u64,
-    is_governance_txn: bool,
+    governance_role: Option<GovernanceRole>,
```
sblackshear, line 632: Thoughts on: 
- Having the `GovernanceRole` enum include all the roles (or one catch-all `NonGovernance` role)
- Getting rid of the `Option` here? 

I think that would let us simplify the comparison of `priority` in the mempool code and make it easier to test that `None` < `Some(GovernanceRole)`.

language/libra-vm/src/libra_transaction_validator.rs
```diff
@@ -153,19 +150,22 @@ impl VMValidator for LibraVMValidator {
...
             res.map(|s| s.status_code()),
             normalized_gas_price,
-            is_prioritized_txn,
+            is_governance_txn,
         )
     }
 }
 
-fn is_prioritized_txn(sender: AccountAddress, remote_cache: &StateViewCache) -> bool {
+fn is_governance_transaction(
```
sblackshear, line 158: Nit: perhaps a different name like `get_role`?

mempool/src/core_mempool/index.rs
```diff
@@ -84,7 +84,12 @@ impl PartialOrd for OrderedQueueKey {
 
 impl Ord for OrderedQueueKey {
     fn cmp(&self, other: &OrderedQueueKey) -> Ordering {
-        match self.is_governance_txn.cmp(&other.is_governance_txn) {
+        match self
+            .governance_role
+            .as_ref()
+            .map(|x| x.priority())
+            .cmp(&other.governance_role.as_ref().map(|x| x.priority()))
```
sblackshear, line 91: Will this do the right thing when one of the `governance_role`'s is `None`?

types/src/transaction/mod.rs
tzakian: I like this idea a lot more than the option. I'll update and add a `NonGovernanceRole` variant.

mempool/src/core_mempool/index.rs
tzakian: It should. But I think it's another reason to get rid of the option :)

types/src/transaction/mod.rs
```diff
@@ -592,16 +592,54 @@ impl From<VMStatus> for TransactionStatus {
...
+            0 => LibraRoot,
+            1 => TreasuryCompliance,
+            2 => DesignatedDealer,
+            3 => Validator,
+            4 => ValidatorOperator,
+            _ => NonGovernanceRole,
+        }
+    }
+
+    pub fn priority(&self) -> u64 {
```
christinacelee, line 618: nit: maybe a comment that the larger the priority value, the more greatly it is prioritized?

mempool/src/core_mempool/index.rs
```diff
@@ -53,7 +53,7 @@ impl PriorityIndex {
             expiration_time: txn.expiration_time,
             address: txn.get_sender(),
             sequence_number: txn.get_sequence_number(),
-            is_governance_txn: txn.is_governance_txn,
+            governance_role: txn.governance_role.clone(),
```
christinacelee, line 56: nit: maybe just derive `Copy` for GovernanceRole to remove cloning?

types/src/transaction/mod.rs
```diff
@@ -592,16 +592,54 @@ impl From<VMStatus> for TransactionStatus {
...
+    LibraRoot,
+    TreasuryCompliance,
+    Validator,
+    ValidatorOperator,
+    DesignatedDealer,
+    NonGovernanceRole,
+}
+
+impl GovernanceRole {
+    pub fn from_role_id(role_id: u64) -> Self {
```
christinacelee, line 606: will the numeric role_id mappings be documented/maintained somewhere?

tzakian: Good question! yes. These are documented in the Move modules published on-chain (specifically Roles.move).

PR #5242, opened 2020-07-22
[cli] Improving documentation and printing for a CLI commands

sblackshear:
- Emphasize that account create is a local operation with no on-chain effect
- Explain that minting transfers funds from a faucet address and that it may create accounts. Add printing when accounts are created.
- Small clarifications in add_currency

### DIFF
--- a/testsuite/cli/src/account_commands.rs
+++ b/testsuite/cli/src/account_commands.rs
@@ -18,7 +18,7 @@ impl Command for AccountCommand {
     }
     fn execute(&self, client: &mut ClientProxy, params: &[&str]) {
         let commands: Vec<Box<dyn Command>> = vec![
-            Box::new(AccountCommandCreate {}),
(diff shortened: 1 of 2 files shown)


xli:
This is great!

Do you know why we don't create onchain account in this account create command? it seems odd to me we don't do it in this command when we need it anyway for anyone to start test mint and other commands.

sblackshear:
> Do you know why we don't create onchain account in this account create command? it seems odd to me we don't do it in this command when we need it anyway for anyone to start test mint and other commands.

Good question--it seems odd to me too :). I think (but am not sure) that these reasons are historical and would strongly support a change. 

One small gotcha is that `account create` would now need to accept a currency type for the created account (that, or we would need to pick a default).

sblackshear:
Cool--if everyone agrees on `account create`, maybe I'll go ahead with this and create a separate PR for that (debatably breaking) change.

PR #5247, opened 2020-07-22
[docker] Use thin LTO for release artifacts

metajack:
Switching to thin LTO to see what performance impact it has relative to full LTO.


metajack:
thin LTO reduces TPS about 5-6%, so it doesn't seem worth the build time to enable full LTO

andll:
@bors-libra r=andll p=1000

I think currently thick lto makes build-aws fail spuriously, so boosting priority for this PR to land faster (other land blocking requests can be stuck otherwise)

Issue #5182, opened 2020-07-17
[Bug] Validator not trying to connect to one peer


mgorven:
# 🐛 Bug

Pre-mainnet currently has 23 nodes online, so validators should have 22 connected peers. Some validators have fewer than this however, and are not trying to connect to the peers they are disconnected from.

    08:36:52.819 [Validator,validator,1c28a02d] Dial future: dialing peer: 5f553934, at address: /dns4/<redacted>.elb.eu-west-1.amazonaws.com/tcp/6180/ln-noise-ik/b4e70d8af20d41e668c7a2666228a40610d6336d9c55029ce81596707c29bd57/ln-handshake/0, after delay: 2.000000468s
    08:36:52.819 [Validator,validator,1c28a02d] Create dial future: peer: 5f553934, at address: /dns4/<redacted>.elb.eu-west-1.amazonaws.com/tcp/6180/ln-noise-ik/b4e70d8af20d41e668c7a2666228a40610d6336d9c55029ce81596707c29bd57/ln-handshake/0, after delay: 2s
    08:36:54.083 [Validator,validator,1c28a02d] Dialing peer: 5f553934, at addr: /dns4/<redacted>.elb.eu-west-1.amazonaws.com/tcp/6180/ln-noise-ik/b4e70d8af20d41e668c7a2666228a40610d6336d9c55029ce81596707c29bd57/ln-handshake/0
    08:36:54.357 [Validator,validator,1c28a02d] Peer '5f553934' successfully dialed at '/dns4/<redacted>.elb.eu-west-1.amazonaws.com/tcp/6180/ln-noise-ik/b4e70d8af20d41e668c7a2666228a40610d6336d9c55029ce81596707c29bd57/ln-handshake/0'
    08:36:54.360 [Validator,validator,1c28a02d] Successfully connected to peer: 5f553934 at address: /dns4/<redacted>.elb.eu-west-1.amazonaws.com/tcp/6180/ln-noise-ik/b4e70d8af20d41e668c7a2666228a40610d6336d9c55029ce81596707c29bd57/ln-handshake/0
    08:36:54.679 [Validator,validator,1c28a02d] Connection from 5f553934 at /ip4/192.168.132.40/tcp/38266 successfully upgraded
    08:36:54.688 [Validator,validator,1c28a02d] Closing existing connection with Peer 5f553934 to mitigate simultaneous dial
[...]

PR #5250, opened 2020-07-22
[network] Remove address check on LostPeer notifs

gregnazario:
During LostPeer notifications, the address may mismatch what is in
storage.  This removes that check and will disconnect the peer if it
doesn't match, and we'll let us later redial or otherwise handle the
connection.

Related to: https://github.com/libra/libra/issues/5182

### DIFF
--- a/network/src/connectivity_manager/mod.rs
+++ b/network/src/connectivity_manager/mod.rs
@@ -116,6 +116,9 @@ pub enum ConnectivityRequest {
     UpdateAddresses(DiscoverySource, HashMap<PeerId, Vec<NetworkAddress>>),
     /// Update set of nodes eligible to join the network.
     UpdateEligibleNodes(DiscoverySource, HashMap<PeerId, HashSet<x25519::PublicKey>>),
+    /// Gets current size of connected peers. This is useful in tests.
+    #[serde(skip)]
+    GetConnectedSize(oneshot::Sender<usize>),
     /// Gets current size of dial queue. This is useful in tests.
     #[serde(skip)]
     GetDialQueueSize(oneshot::Sender<usize>),
@@ -468,6 +471,9 @@ where
             ConnectivityRequest::GetDialQueueSize(sender) => {
                 sender.send(self.dial_queue.len()).unwrap();
             }
+            ConnectivityRequest::GetConnectedSize(sender) => {
+                sender.send(self.connected.len()).unwrap();
+            }
         }
     }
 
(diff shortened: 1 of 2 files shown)


network/src/connectivity_manager/mod.rs
```diff
@@ -118,6 +118,9 @@ pub enum ConnectivityRequest {
     UpdateEligibleNodes(DiscoverySource, HashMap<PeerId, HashSet<x25519::PublicKey>>),
     /// Gets current size of dial queue. This is useful in tests.
     #[serde(skip)]
+    GetConnectedSize(oneshot::Sender<usize>),
```
phlip9, line 121: nit: comment is copy-pasted : p

network/src/connectivity_manager/mod.rs
```diff
@@ -597,19 +603,16 @@ where
...
-                        debug!(
-                            "{} Ignoring stale lost peer event for peer: {}, addr: {}",
-                            self.network_context,
-                            peer_id.short_str(),
-                            addr
-                        );
-                    }
+                if self.connected.contains_key(&peer_id) {
+                    // Remove node from connected peers list.
+                    self.connected.remove(&peer_id);
```
davidiw, line 608: do we want to log this as well?

gregnazario: Luckily now we have the incoming / outgoing messages from the events.

I'll add a structured log for every event received

network/src/connectivity_manager/mod.rs
phlip9, line 121: man, i wish we could refactor these actors so we didn't need these test-only messages...

gregnazario: I actually tried to refactor it, and then gave up to do this instead

gregnazario: :(

network/src/connectivity_manager/mod.rs
gregnazario: Oh, turns out we already log the entire event in the structured logs event loop, with the internals, which would tell the directions and all.

davidiw: Agreed! But that doesn't show up per our previous conversation. Once that works, I think we can probably start eliminating a bunch of logs. I don't know if there's any value from having both structured and text logs. We eliminated text logs from the TCB unless there was something that didn't fit into the model for structured loggings.

PR #5253, opened 2020-07-22
[state sync] update counters/dashboards

christinacelee:
## Summary

Added/updated counters and dashboards for state sync

## Test Plan

Checked grafana dashboard renders properly in cluster test run
Art:
screencapture-grafana-ct-0-k8s-testnet-aws-hlw3truzy4ls-d-r-u0dWtZkd-state-sync-2020-07-23-17_02_21

### DIFF
--- a/state-synchronizer/src/coordinator.rs
+++ b/state-synchronizer/src/coordinator.rs
@@ -230,13 +230,21 @@ impl<T: ExecutorProxyTrait> SyncCoordinator<T> {
                 msg = self.client_events.select_next_some() => {
                     match msg {
                         CoordinatorMessage::Request(request) => {
+                            let timer = counters::PROCESS_MSG_LATENCY
+                                .with_label_values(&[counters::SYNC_MSG_LABEL, counters::CONSENSUS_SENDER_LABEL])
+                                .start_timer();
                             if let Err(e) = self.request_sync(*request) {
                                 error!("[state sync] request sync fail: {}", e);
                             }
                         }
                         CoordinatorMessage::Commit(txns, events, callback) => {
-                            if let Err(e) = self.process_commit(txns, Some(callback)).await {
-                                error!("[state sync] process commit fail: {}", e);
+                            {
+                                let timer = counters::PROCESS_MSG_LATENCY
+                                    .with_label_values(&[counters::COMMIT_MSG_LABEL, counters::CONSENSUS_SENDER_LABEL])
+                                    .start_timer();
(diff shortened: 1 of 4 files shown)


state-synchronizer/src/coordinator.rs
```diff
@@ -230,14 +230,21 @@ impl<T: ExecutorProxyTrait> SyncCoordinator<T> {
...
                             }
                         }
                         CoordinatorMessage::Commit(txns, events, callback) => {
+                            let timer = counters::PROCESS_MSG_LATENCY
+                                .with_label_values(&[counters::COMMIT_MSG_LABEL, counters::CONSENSUS_SENDER_LABEL])
+                                .start_timer();
                             if let Err(e) = self.process_commit(txns, Some(callback)).await {
                                 error!("[state sync] process commit fail: {}", e);
                             }
+                            timer.stop_and_record();
```
zekun000, line 247: do you need this? isn't it RAII?

state-synchronizer/src/executor_proxy.rs
```diff
@@ -146,11 +146,14 @@ impl ExecutorProxyTrait for ExecutorProxy {
         intermediate_end_of_epoch_li: Option<LedgerInfoWithSignatures>,
         _synced_trees: &mut ExecutedTrees,
     ) -> Result<()> {
+        // track chunk execution time
+        let execute_start_time = Instant::now();
```
zekun000, line 150: timer doesn't work?

state-synchronizer/src/coordinator.rs
christinacelee: i excluded the on-chain config publishing time intentionally, but can include it, i have no strong opinions

state-synchronizer/src/executor_proxy.rs
christinacelee: this is `DurationHistogram`, which is just a wrapper around `Histogram`but doesn't have start_timer support

state-synchronizer/src/coordinator.rs
zekun000: oh you can wrap the code block instead, which makes it nicer (imo)
```
{
   let _timer = ...;
   work()
}
```

state-synchronizer/src/executor_proxy.rs
christinacelee: actually i can just change this to `Histogram`, I don't see any real advantage to `DurationHistogram`

PR #5254, opened 2020-07-22
[ci] add dependency verification action

metajack:
This adds a new GitHub Actions workflow to report on dependency changes to specific subsets of Libra. It currently tests the safety rules TCB and the execution correctness TCB as well as all the validator release binaries.

It will comment and label PRs that make such changes, as well as notifying @davidiw when TCB deps are changed.

This effectively replaces cargo-dep-bot.


metajack:
Note that I will update the in-tree sumarries to the latest  before I land this. Otherwise it will trigger on every PR.

Also, I need to add support for removing comments and labels if the PR changes to not have dep changes.

metajack:
Nice, comments and labels were deleted by the action. This is ready to go!

.github/CODEOWNERS
```diff
@@ -5,3 +5,6 @@
 /client/libra-dev/ @thefallentree @xli
 # Subscribe to LCS format changes
 /testsuite/generate-format/tests/staged/libra.yaml @thefallentree @matbd @xli
+# Subscribe to TCB dependency summary changes
+/summaries/summary-lsr.toml @davidiw
+/summaries/summary-lec.toml @davidiw
```
bmwill, line 10: Do we want an owner for the release summary?

summaries/summary-lec.toml
```diff
@@ -0,0 +1,1595 @@
+# Summary for Libra subset 'lec'. @generated by x.
+# To regenerate, run 'cargo x generate-summaries'.
```
bmwill, line 2: Actually thinking about this more, we probably can't merge this yet until we can block landing  if these summary files are out of date. The last thing we want is the action reporting differences when a PR doesn't actually include any (but a previously merged PR into master had differences but didn't update these files)

bmwill: So we probably just need to add a step in CI that just runs `cargo x generate-summaries` and ensures that no files are modified after that.

metajack: I'll add something to fail the PR if the sumamries don't match.

.github/CODEOWNERS
metajack: I was only going to enforce the TCB with the extra review as a first step. We can enforce the release binaries later if this goes well.

metajack:
Ok, so the desired behavior should be to pass/fail based on summary differences, but post comment/label should always happen with the delta between the merge base and the new calculated summary. I'll fix that up.

metajack:
Ok, new behavior added.

PR #5256, opened 2020-07-22
[language] Add ability to call stdlib scripts from functional tests

tzakian:
Allows us to call the precompiled stdlib scripts from the functional tests.

### DIFF
--- a/language/functional-tests/src/evaluator.rs
+++ b/language/functional-tests/src/evaluator.rs
@@ -7,7 +7,7 @@ use crate::{
     errors::*,
 };
 use bytecode_verifier::DependencyChecker;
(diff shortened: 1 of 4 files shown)


language/functional-tests/src/evaluator.rs
```diff
@@ -40,6 +42,18 @@ use vm::{
     views::ModuleView,
 };
 
+static PRECOMPLIED_TXN_SCRIPTS: Lazy<HashMap<String, CompiledScript>> = Lazy::new(|| {
```
sblackshear, line 45:
```suggestion
static PRECOMPILED_TXN_SCRIPTS: Lazy<HashMap<String, CompiledScript>> = Lazy::new(|| {
```

Issue #5082, opened 2020-07-14
[Tools] Build an operator tool to do common tasks


gregnazario:
### Problem
Operators need a tool to do common tasks like rotate keys, set values, and lookup values.

### Approach
Let's make a shared tool library that has all the basics needed for reading/writing information to JSON RPC, and local/remote storage.  Then we can have common built work for different tools and save the work of reimplementing the same building blocks over and over again.

* Commands should be simple, as much information that can be removed from input args and pulled from data sources should be used.
* Commands should output any information that the task run creates that would be useful to users.
* Consider using a config for any common information that would be used among runs

### Subtasks
- [x] Take the current existing `management` tool, and rename it `genesis` (https://github.com/libra/libra/pull/5152)
- [x] Move common tooling work into a tooling library to be shared between both (https://github.com/libra/libra/pull/5186)
- [x] Make a new tool named `libra-operator-tool` for operators and other users not during genesis (https://github.com/libra/libra/pull/5186)
- [x] JsonRpc entrypoint, transaction status check, and print account
- [ ] Implement Phase 1 functions for the next testing iteration (Ask @sherry-x for which)
- [x] Implement Phase 2 functions
- [ ] Change smoke test to use these functions to ensure that management tool works correctly

### Phase 1 functions
- [ ] Create validator operator
- [ ] Create validator
- [ ] Set operator
- [ ] Print operator state (e.g., similar to validator config)
- [x] Add validator
- [x] Remove validator
- [x] Set validator config (https://github.com/libra/libra/pull/5186)
- [x] Operational tool - read a key from storage

### Phase 2 functions
- [x] Update network address either or both of validator(s) or fullnode(s)
- [x] Refresh validator config
- [x] Rotate validator network key (https://github.com/libra/libra/pull/5186)
- [x] Rotate fullnode network key (https://github.com/libra/libra/pull/5257)
- [x] Rotate operator key
- [x] Rotate consensus key

gregnazario:
@davidiw We can brainstorm here any other info

gregnazario:
We should also add NetworkKey rotation to this list

davidiw:
Ahh, yes, there's a bit of interesting work to be thought about here! We likely want something that can trigger the rotation. Like do we want a single function to do the rotation and submit the validator-config transaction?

gregnazario:
[...]

PR #5257, opened 2020-07-22
[management] added rotate-fullnode-network-key in op tooling

gregnazario:
This adds the functionality to rotate the fullnode network key
in the op tool.  The caveat is that there is currently no testing for
connecting through a full node.  It does consolidate network key
rotation common logic between the two rotation functions.

TBD: We need a way to test this functionality, by first having a way to connect through the fullnodes

https://github.com/libra/libra/issues/5082


JoshLind:
Actually, let's land this ASAP to avoid conflicts with the consensus key @bors-libra r+

PR #5260, opened 2020-07-23
[Management Operations] Support consensus key rotations

JoshLind:
## Motivation

This PR adds support for rotating the consensus key in the management operations tool. It does so in two commits:

1. Adding a new command that rotates the consensus key by refactoring some of the network key rotation logic.
2. Adding a new smoke test for this functionality as well as refactoring some of the duplication between the consensus key rotation test and the validator network key rotation test.

Note: there's probably quite a few refactors and cleanups we can do to both the smoke tests and the operator tool, but this work won't be done in this PR. Instead, let's get this in first, and then can refactor later 😄  

Yes.

## Test Plan

The new smoke test passes locally.

## Related PRs

None, but this does relate to the issue: https://github.com/libra/libra/issues/5082

### DIFF
--- a/testsuite/tests/libratest/smoke_test.rs
+++ b/testsuite/tests/libratest/smoke_test.rs
@@ -3,7 +3,7 @@
 
 use cli::client_proxy::ClientProxy;
 use debug_interface::NodeDebugClient;
-use libra_config::config::{Identity, KeyManagerConfig, NodeConfig};
+use libra_config::config::{Identity, KeyManagerConfig, NodeConfig, SecureBackend};
 use libra_crypto::{ed25519::Ed25519PrivateKey, PrivateKey, SigningKey, Uniform};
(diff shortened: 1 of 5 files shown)


testsuite/tests/libratest/smoke_test.rs
```diff
@@ -1473,11 +1474,30 @@ fn wait_for_all_nodes(
...
+fn load_backend_storage(node_config: &&NodeConfig) -> SecureBackend {
+    if let Identity::FromStorage(storage_identity) =
+        &node_config.validator_network.as_ref().unwrap().identity
+    {
+        storage_identity.backend.clone()
+    } else {
+        panic!("Couldn't load identity from storage");
+    }
+}
+
```
gregnazario, line 1498: Nice!

JoshLind:
Thanks, Greg! 😄 @bors-libra r=gregnazario

PR #5261, opened 2020-07-23
[move-prover] Spec block level lets and lambdas.

wrwg:
This PR introduces let in spec blocks. Goal is to reduce boilerplate in large specifcations.

Example usage:

```
fun some(dd_addr: address) { ... }
spec fun some {
    let dealer = global<TierInfo<CoinType>>(dd_addr);
    ensures old(dealer.window_start) <= dealer.window_start;
    aborts_if is_invalid(dealer);
}
```

One can also use lambdas with lets:

```
spec fun some {
    let dealer_of = |dd_addr| global<TierInfo<CoinType>>(dd_addr);
    ensures old(dealer_of(dd_addr).window_start) <= dealer_of(dd_addr).window_start;
    ...
}
```

The lambdas are a side-effect of the implementation of lets, which are implemented by lifting them to global specification functions. Notice that those lambdas can only be called; we still can't pass function values as parameters to other functions.

Lets and lambdas do full capture of context, including transparent usage of `old(...)` and `result_i` symbols. For examples see the new test source `let.move`.

This PR also does a few other things:
- fixes a bug with lets in expressions. Let's in expressions are different from the spec block lets, and are still translated to Boogie.
- Enables syntactic sugar `spec define f(..) { .. }` as shortcut for `spec module { define f(..) { .. } }`
- Disallows shadowing of locals. There are bugs around shadowing both in the prover and in Boogie. Better turn this off for now and produce an good error. It appears none of the existing framework specs is affected by this.

## Motivation

Reduce boilerplate in specifications.

Yes

## Test Plan

New test.

## Related PRs

NA

### DIFF
--- a/language/move-prover/tests/sources/functional/let.move
+++ b/language/move-prover/tests/sources/functional/let.move
@@ -1,0 +1,71 @@
+module TestLet {
+
+    spec module {
+        pragma verify = true;
+    }
+
+    resource struct R {
+        x: u64
+    }
+
+    // Spec Block Let
+    // ==============
+
+    fun spec_let_with_result(a: u64, b: u64): (u64, u64) {
+        (a + 1 + b, a + b)
+    }
+    spec fun spec_let_with_result {
+        let x = a + 1;
+        let y = x + b;
+        let r2 = result_1 - 1;
+        ensures result_1 == y;
+        ensures result_2 == r2;
+    }
+
+    fun spec_let_with_old(a: &mut u64, b: &mut u64) {
+        let saved_a = *a;
+        *a = *a + 1 + *b;
+        *b = saved_a + *b;
+    }
+    spec fun spec_let_with_old {
+       let x = old(a) + 1;
+       let y = x + old(b);
+       let r2 = a - 1;
+       ensures a == y;
+       ensures b == r2;
+    }
+
+    fun spec_let_with_lambda(a: u64, b: u64): (u64, u64) {
+        (a + 1 + b, a + b)
+    }
+    spec fun spec_let_with_lambda {
+        let add_to_a = |n| a + n;
+        let add_to_b = |n| b + n;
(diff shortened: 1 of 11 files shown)


language/move-prover/tests/sources/functional/let.move
```diff
@@ -0,0 +1,71 @@
...
+    }
+
+    // Spec Block Let
+    // ==============
+
+    fun spec_let_with_result(a: u64, b: u64): (u64, u64) {
+        (a + 1 + b, a + b)
+    }
+    spec fun spec_let_with_result {
+        let x = a + 1;
```
shazqadeer, line 18: The semantics appears to be macro definition and substitution.  I am wondering if the use of word "let" for this feature will lead to confusion.  As you explained in the PR description, perhaps the right way to think of this feature as a let is that we are let-binding a lambda which is evaluated later when used in a function.  Is there a way to somehow indicate this in the syntax?

shazqadeer: Perhaps use some keyword other than let, e.g., let-fun?

shazqadeer: Yet another option might be to use let keyword but require the lambda notation always  even if there are no explicit parameters, e.g.,

let y = || x + b;

wrwg: It's actually not substitution but polymorphic let (in Hindley-Milner sense -- ML language). Yes, they are lazily evaluated, but they are not substituted but replaced by functions which are called as needed. I feel since the prover doesn't really makes any difference between "strict" and "lazy" evaluation, we do not need to complicate the mental model by additional keywords and syntax. I agree we should document this lazy aspect once we write the language definition, just so people aren't shy of defining lots of lets even if they are only used inside of conditional expressions. 

For the "let-fun" and the "let y = ||...", I would argue similar.

PR #5267, opened 2020-07-23
[logging] add a maximum size to structured logging

gregnazario:
###  Commit message
The autoconverted structured logging lines are causing there to be too
much data that's getting cut off.  This limits the amount of data in a
log line to ensure delivery, adding an error message to know if we need
to shrink the size of the line.

### Why make this change?
Logs like [this](https://github.com/libra/libra/blob/2019189392119bc2f4c9f6ae36f00049770eaf57/consensus/src/round_manager.rs#L314-L319)
```
Peer {} has stale state {}, send it back {}
```
have way too much data and cause there to be fragmented data in the batches.  As a result, we need to shorten them so that they break an entire batch of log lines with `_jsonparsefailure`

### DIFF
--- a/common/logger/src/struct_log.rs
+++ b/common/logger/src/struct_log.rs
@@ -40,6 +40,11 @@ const INITIALIZED: usize = 2;
 const SEVERITY_CRITICAL: usize = 1;
 const SEVERITY_WARNING: usize = 2;
 
+// Size configurations
+const MAX_LOG_LINE_SIZE: usize = 4096; // 4KiB
+const LOG_INFO_OFFSET: usize = 256;
+const WRITE_CHANNEL_SIZE: usize = 1024;
+
 #[derive(Default, Serialize)]
 pub struct StructuredLogEntry {
     /// log message set by macros like info!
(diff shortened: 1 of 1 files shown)


andll:
1) I would put value to something more reasonable, like at least 4k.
I think we discussed with @davidiw that we will need to switch to TCP, so packet size does not apply, we just need to make sure we have something reasonable

2) Can we generalize it, so that both file and UDP cats size same way? It would not be cool if some logs works ok in cluster test, but does not work in prod because of this

common/logger/src/struct_log.rs
```diff
@@ -90,6 +95,56 @@ impl StructuredLogEntry {
...
+                MAX_LOG_LINE_SIZE
+            );
+
+            let mut entry = self.clone_without_data();
+            if let Some(mut log) = entry.log {
+                log.truncate(MAX_LOG_LINE_SIZE - LOG_INFO_OFFSET);
+                // Leave 128 bytes for all the other info
+                entry.log = Some(log);
+            }
+            entry = entry.data(
```
andll, line 138: Do you also want to add information to data about what was original length? As a separate field so that it could be analyzed

common/logger/src/struct_log.rs
```diff
@@ -90,6 +95,56 @@ impl StructuredLogEntry {
         ret
     }
 
+    pub fn clone_without_data(&self) -> Self {
```
andll, line 98: Looks like this could be private?

common/logger/src/struct_log.rs
```diff
@@ -90,6 +95,56 @@ impl StructuredLogEntry {
...
+
+    fn to_json(&self) -> Result<Value, serde_json::Error> {
+        let value = serde_json::to_value(self);
+        if let Err(e) = &value {
+            log::error!("[Logging] Failed to serialize struct log entry: {}", e);
+        }
+        value
+    }
+
+    pub fn to_json_string(&self) -> Result<String, serde_json::Error> {
```
andll, line 121: This can be private too, to keep API clean?

common/logger/src/struct_log.rs
```diff
@@ -90,6 +95,56 @@ impl StructuredLogEntry {
...
+        let json = self.to_json()?;
+
+        // If the message is too long, let's drop the
```
andll, line 124: Looks like end of this comment is also dropped :)

PR #5270, opened 2020-07-23
[consensus] cut down on logging of consensus objects

gregnazario:
The debug output was being used in structured logging, which blew up to
a much larger size due to the many internal items.  This uses the
display implementation for the debug information.

Note: I'm not sure all places these are used, but I need someone else to check whether these objects can have their debug info ignored.


consensus/consensus-types/src/sync_info.rs
```diff
@@ -18,6 +18,13 @@ pub struct SyncInfo {
     highest_timeout_cert: Option<TimeoutCertificate>,
 }
 
+// this is required by structured log
+impl std::fmt::Debug for SyncInfo {
+    fn fmt(&self, f: &mut std::fmt::Formatter) -> std::fmt::Result {
```
davidiw, line 23: super nit: std::fmt::Formatter is already in scope and you could trivially make std::fmt::Debug into scope.

PR #5276, opened 2020-07-23
[network] reduced structured logging overhead

gregnazario:
Most items should now implement display and use display rather than
debug.  Debug is overridden with display so that logs don't expand
drastically.  Additionally, serialization is cut down for objects in
structured logging that didn't add value.

I've gone through all networking logs that were non-structured and implemented display / debug for a ton of objects to ensure that we have an actual display method and don't have to use debug.  There are a few cases that I didn't clean up, but it's much less debug usage now.

In addition, cleaned up whenever we used `{:?}` and the value was already implemented as display.

### DIFF
--- a/types/src/chain_id.rs
+++ b/types/src/chain_id.rs
@@ -1,7 +1,7 @@
 // Copyright (c) The Libra Core Contributors
 // SPDX-License-Identifier: Apache-2.0
 use anyhow::{format_err, Error, Result};
-use serde::{de::Visitor, Deserialize, Deserializer, Serialize};
+use serde::{de::Visitor, export::fmt::Debug, Deserialize, Deserializer, Serialize};
 use std::{
     convert::TryFrom,
     fmt::{Display, Formatter},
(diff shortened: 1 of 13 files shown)


types/src/chain_id.rs
```diff
@@ -86,10 +96,35 @@ where
...
+impl Display for NamedChain {
+    fn fmt(&self, f: &mut Formatter) -> std::fmt::Result {
+        write!(
+            f,
+            "{}",
+            match self {
+                NamedChain::DEVNET => "Devnet",
+                NamedChain::TESTNET => "Testnet",
+                NamedChain::MAINNET => "Mainnet",
+                NamedChain::TESTING => "Testing",
```
davidiw, line 125: Can you humor me and test out the config-builder faucet command with a chain-id Testnet versus TESTNET, because we're using TESTNET internally.

gregnazario: Ah, yes this is problematic on my side, it's all caps in the other direction.  will fix

PR #5277, opened 2020-07-24
[state-sync] cleaned up state-sync structured logging

gregnazario:
Replaced most debug logging with display logging accordingly, and
attempted to reduce the likelihood of spewing out large amounts of data
into structured logging.

It's really up to those who are familiar with it to tell me whether the LedgerInfo, which I've mostly stopped logging because of its extremely large overhead is required to be logged and whether these new amounts are the right level.

### DIFF
--- a/config/src/config/upstream_config.rs
+++ b/config/src/config/upstream_config.rs
@@ -4,6 +4,7 @@
 use crate::network_id::{NetworkId, NodeNetworkId};
 use libra_types::PeerId;
 use serde::{Deserialize, Serialize};
+use std::fmt;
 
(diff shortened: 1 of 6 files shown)


zekun000:
I think log the Display format is good enough, no one will look at the signatures in the debug format.

gregnazario:
> I think log the Display format is good enough, no one will look at the signatures in the debug format.

Here's the problem.  The structured log converter takes every field that's in the `debug!("{},{}",a,b)` and runs the debug printer on it.  Which is why in consensus it was spewing large amounts of data in structured logging, even though it had a display handler.

tl;dr We have a display handler, if the debug doesn't provide us more useful info just override it with the display.

PR #5278, opened 2020-07-24
[mempool] add fuzzing for processing incoming txns

christinacelee:
## Motivation

Adding fuzzing for mempool's `process_incoming_transactions` endpoint

## Test Plan

New/existing tests


mempool/src/tests/fuzzing.rs
```diff
@@ -0,0 +1,68 @@
...
+        network_senders: HashMap::new(),
+        db: Arc::new(mock_db),
+        validator: vm_validator,
+        peer_manager: Arc::new(PeerManager::new(config.upstream)),
+        subscribers: vec![],
+    };
+
+    // generate fuzz data
+    let txns: Vec<SignedTransaction> = lcs::from_bytes(data)
+        .expect("[fuzzing] failed to convert bytes to transactions submitted to mempool");
```
mimoo, line 60: So I don't think this will work, a lot of the time the data will not be deserializable and this will panic.

I would recommend using the fuzzer to generate a `Vec<SignedTransaction>` via proptest instead. You have an example on how to do that here: https://github.com/libra/libra/blob/master/testsuite/libra-fuzzer/src/fuzz_targets/storage_save_blocks.rs#L32

mempool/src/tests/fuzzing.rs
```diff
@@ -0,0 +1,68 @@
...
+pub fn generate_corpus(gen: &mut ValueGenerator) -> Vec<u8> {
+    // generate 100 random SignedTransactions via proptest
+    // 100 is the max number of transactions in a submission batch
+    let mut txns = vec![];
+    for _ in 0..100 {
+        let txn = gen.generate(proptest::arbitrary::any::<SignedTransaction>());
+        txns.push(txn);
+    }
+    lcs::to_bytes(&txns)
+        .expect("[fuzzing] failed to convert mempool submitted transactions to bytes")
```
mimoo, line 39: maybe change the message to "failed to generate corpus" or something?

mempool/src/tests/fuzzing.rs
```diff
@@ -0,0 +1,68 @@
...
+    };
+
+    // generate fuzz data
+    let txns: Vec<SignedTransaction> = lcs::from_bytes(data)
+        .expect("[fuzzing] failed to convert bytes to transactions submitted to mempool");
+
+    // randomly select timeline state
+    let possible_timeline_states = vec![TimelineState::NotReady, TimelineState::NonQualified];
+    let timeline_state = possible_timeline_states
+        .choose(&mut rand::thread_rng())
```
mimoo, line 65: in general it's best to avoid RNGs in fuzzers because it is not deterministic (imagine trying to re-run it with an input you know will fail, but then doesn't because the other timeline_state gets called) and it slows down the run.

Here I would suggest instead using `prop_oneof![]` as part of a proptest strategy

mimoo: also what is this `choose()` function oO?

mimoo: maybe something like this would work?

```rust

fn mempool_something_strategy() -> impl Strategy<(Vec<SignedTransaction>, TimelineState)> {
  (
    proptest::collection::vec(any::<SignedTransaction>(), 100),
    prop_oneof![
      Just(TimelineState::NotReady), 
      Just(TimelineState::NonQualified),
    ],
  )
}
```

mempool/src/tests/fuzzing.rs
```diff
@@ -0,0 +1,61 @@
...
+    sync::{Arc, Mutex, RwLock},
+};
+use storage_interface::mock::MockDbReader;
+use vm_validator::mocks::mock_vm_validator::MockVMValidator;
+
+/// Proptest strategy for fuzzing input to mempool incoming transactions endpoint
+pub fn mempool_incoming_transactions_strategy(
+) -> impl Strategy<Value = (Vec<SignedTransaction>, TimelineState)> {
+    (
+        proptest::collection::vec(proptest::arbitrary::any::<SignedTransaction>(), 100),
```
mimoo, line 25: actually I think you should give a range as second argument to `vec` here so that it creates a vector of arbitrary size in that range, I think `0..100` would work? I haven't thought much about 100 and if it's too high :D maybe @sunshowers has an opinion on this

mimoo:
nice job! Just left you a comment on the range

@bors-libra delegate+

PR #5279, opened 2020-07-24
[network] add incoming and outgoing connection metrics

gregnazario:
In order to keep track of the incoming and outgoing connections on the transport connection level, we can now keep track of whether a peer connection is incoming or outgoing in metrics.  That'll let us know if a node can connect to others, or others can connect to the node.

### DIFF
--- a/network/src/peer_manager/mod.rs
+++ b/network/src/peer_manager/mod.rs
@@ -352,6 +352,28 @@ where
         }
     }
 
+    pub fn update_connected_peers_metrics(&self) {
+        let total = self.active_peers.len();
+        let inbound = self
+            .active_peers
+            .iter()
+            .filter(|(_, (metadata, _))| metadata.origin() == ConnectionOrigin::Inbound)
+            .count();
+        let outbound = total.saturating_sub(inbound);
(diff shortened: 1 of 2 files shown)


network/src/peer_manager/mod.rs
```diff
@@ -352,6 +352,28 @@ where
...
+            .with_label_values(&[role, "connected"])
+            .set(total as i64);
+
+        counters::LIBRA_NETWORK_PEERS
+            .with_label_values(&[role, "inbound"])
+            .set(inbound as i64);
+
+        counters::LIBRA_NETWORK_PEERS
+            .with_label_values(&[role, "outbound"])
+            .set(outbound as i64);
```
gregnazario, line 374: For some extremely odd reason, adding another dimension causes the node not to start up.  In addition, renaming the "state" field as "direction" or any other value also causes the node not to start up.  This became apparent in the `test_vfn_failover` e2e test, but seemingly not in any other tests?

gregnazario: I'll push this for now so we have a metric, and I'll loop back with @ankushagarwal about if he has any insight into why this doesn't work.

PR #5280, opened 2020-07-24
[faucet] return expected DD next sequence number

sausagee:
## Summary
Fix a sequence number issue introduced in #5232, where the faucet does not wait before returning the DD's sequence number, which could be stale.  With this commit, the faucet will return to the client the expected next sequence number.

## Test
Tested on own cluster.

### DIFF
--- a/docker/mint/server.py
+++ b/docker/mint/server.py
@@ -64,15 +64,20 @@ def send_transaction():
 
     try:
         create_client()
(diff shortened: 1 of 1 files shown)


thefallentree:
in theory there could be some race condition here, between querying AS and submit transaction, the seq number could have changed.  But I think we can pretty safely ignore it here.

PR #5281, opened 2020-07-24
[json-rpc] add json-rpc test "test_get_account" back

### LINKED ISSUE #5245: [json-rpc] fix json-rpc unit test test_get_account
test is commented after we added on-chain currencies support at server side, mockdb does not have genesis information.
We need a better solution to test get_account and add the test back

xli:
Fix #5245

1. added genesis data to json-rpc mockdb
2. exposed process_write_set in executor for processing genesis change set and get account state blob for setting up mockdb.
3. moved get registered currencies info to account_state
4. fixed test suite setup account state blob missing freezing_bit problem.

### DIFF
--- a/json-rpc/Cargo.toml
+++ b/json-rpc/Cargo.toml
@@ -49,6 +49,12 @@ libra-proptest-helpers = { path = "../common/proptest-helpers" }
 libra-temppath = { path = "../common/temppath", version = "0.1.0" }
 libra-types = { path = "../types", version = "0.1.0", features = ["fuzzing"] }
 vm-validator = { path = "../vm-validator", version = "0.1.0" }
+compiled-stdlib = { path = "../language/stdlib/compiled",  version = "0.1.0" }
+vm-genesis = { path = "../language/tools/vm-genesis", version = "0.1.0" }
+executor = { path = "../execution/executor", version = "0.1.0" }
+executor-types = { path = "../execution/executor-types", version = "0.1.0" }
+scratchpad = { path = "../storage/scratchpad", version = "0.1.0" }
+move-vm-types = { path = "../language/move-vm/types", version = "0.1.0" }
 
 [features]
 fuzzing = ["proptest", "libra-mempool/fuzzing", "libra-proptest-helpers", "libra-temppath", "libradb/fuzzing", "reqwest"]
--- a/json-rpc/src/methods.rs
+++ b/json-rpc/src/methods.rs
@@ -18,19 +18,17 @@ use libra_mempool::MempoolClientSender;
 use libra_trace::prelude::*;
 use libra_types::{
     account_address::AccountAddress,
-    account_config::{from_currency_code_string, CurrencyInfoResource},
+    account_config::{from_currency_code_string, libra_root_address},
     account_state::AccountState,
     chain_id::ChainId,
     event::EventKey,
     ledger_info::LedgerInfoWithSignatures,
     mempool_status::MempoolStatusCode,
-    move_resource::MoveStorage,
-    on_chain_config::{OnChainConfig, RegisteredCurrencies},
     transaction::SignedTransaction,
 };
 use network::counters;
 use serde_json::Value;
-use std::{collections::HashMap, convert::TryFrom, ops::Deref, pin::Pin, str::FromStr, sync::Arc};
+use std::{collections::HashMap, convert::TryFrom, pin::Pin, str::FromStr, sync::Arc};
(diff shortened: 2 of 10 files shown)


json-rpc/src/tests/genesis.rs
```diff
@@ -0,0 +1,37 @@
...
+use executor_types::ProofReader;
+use libra_types::{
+    account_address::AccountAddress, account_state_blob::AccountStateBlob, transaction::Transaction,
+};
+//use once_cell::sync::Lazy;
+use scratchpad::SparseMerkleTree;
+use std::{collections::HashMap, sync::Arc};
+use vm_genesis::generate_genesis_change_set_for_testing;
+
+// pub static GENESIS_STATE: Lazy<Map<AccountAddress, AccountStateBlob>> =
```
thefallentree, line 15: delete this?

json-rpc/src/tests/genesis.rs
```diff
@@ -0,0 +1,37 @@
+// Copyright (c) The Libra Core Contributors
+// SPDX-License-Identifier: Apache-2.0
+
+use compiled_stdlib::StdLibOptions;
+use executor::process_write_set;
+use executor_types::ProofReader;
+use libra_types::{
+    account_address::AccountAddress, account_state_blob::AccountStateBlob, transaction::Transaction,
+};
+//use once_cell::sync::Lazy;
```
thefallentree, line 10: delete?

json-rpc/Cargo.toml
```diff
@@ -35,8 +35,14 @@ libra-types = { path = "../types", version = "0.1.0" }
...
 libra-workspace-hack = { path = "../common/workspace-hack", version = "0.1.0" }
 move-core-types = { path = "../language/move-core/types", version = "0.1.0" }
+move-vm-types = { path = "../language/move-vm/types", version = "0.1.0" }
 network = { path = "../network", version = "0.1.0" }
 storage-interface = { path = "../storage/storage-interface", version = "0.1.0" }
+compiled-stdlib = { path = "../language/stdlib/compiled",  version = "0.1.0" }
+vm-genesis = { path = "../language/tools/vm-genesis", version = "0.1.0" }
+executor = { path = "../execution/executor", version = "0.1.0" }
+executor-types = { path = "../execution/executor-types", version = "0.1.0" }
+scratchpad = { path = "../storage/scratchpad", version = "0.1.0" }
```
thefallentree, line 45: all these should be "optional" deps, since it is only used in test, right?  but it is kind of concerning that we are importing so many stuff in, all these should probably be moved intoto libradb package in the future.

json-rpc/src/methods.rs
```diff
@@ -125,32 +123,40 @@ async fn get_account(
...
-    let response = service
+    let account_state_blob = service
         .db
         .get_account_state_with_proof_by_version(account_address, request.version())?
         .0;
+    if account_state_blob.is_none() {
+        return Ok(None);
+    }
+
+    let account_state = AccountState::try_from(&account_state_blob.unwrap())?;
```
thefallentree, line 134: .unwrap() would make a lot of people nervous  (it panics) , I suggest we rewrite into an if-else or. an match Some()=> , None=>  that doesn't need unwrap()

thefallentree: you can also write this way

let accoutn_state = account_state_blob.map(|x|  AccountState::try_from(x))? ;

json-rpc/Cargo.toml
xli: we have to keep them for now, because the tests dir is under src, so build need them.
I will have next diff to move tests out of src, then update the dependencies.

xli: my bad, missing cfg(test) for genesis file, fixed now.

json-rpc/src/methods.rs
```diff
@@ -294,28 +302,21 @@ async fn currencies_info(
...
-    let currencies = RegisteredCurrencies::from_bytes(&raw_data[0])?;
-    let access_paths: Vec<_> = currencies
-        .currency_codes()
-        .iter()
-        .map(|code| CurrencyInfoResource::resource_path_for(code.clone()))
-        .collect();
-
-    let mut currencies = vec![];
-    for raw_data in service
+    let (blob, _) = service
```
thefallentree, line 305: nit: this could be shortned as

```
if let Some(blob), _ = service ....  {
 .... 
}
```

and you don't need the second match

thefallentree:
looks good, minor nit

PR #5282, opened 2020-07-24
[state sync] add fuzzing for state sync network msgs

christinacelee:
## Summary

Add fuzzing for state sync's `process_one_message` endpoint, which accepts `StateSynchronizerMsg`

## Test Plan

New/existing tests


testsuite/libra-fuzzer/src/fuzz_targets/state_sync_msg.rs
```diff
@@ -0,0 +1,27 @@
...
+    fn description(&self) -> &'static str {
+        "State sync network message"
+    }
+
+    fn generate(&self, _idx: usize, gen: &mut ValueGenerator) -> Option<Vec<u8>> {
+        Some(generate_corpus(gen))
+    }
+
+    fn fuzz(&self, data: &[u8]) {
+        fuzzer(data.to_vec())
```
mimoo, line 25: why convert to a vec here?

state-synchronizer/src/tests/fuzzing.rs
```diff
@@ -0,0 +1,226 @@
...
+    let data = generate_corpus(&mut gen);
+    fuzzer(data);
+}
+
+pub fn generate_corpus(gen: &mut ValueGenerator) -> Vec<u8> {
+    let msg = gen.generate(state_sync_msg_strategy());
+    lcs::to_bytes(&msg).expect("[fuzzing] failed to generate corpus for state sync msg")
+}
+
+pub fn fuzzer(data: Vec<u8>) {
```
mimoo, line 53: if you change this to &[u8] it should work

state-synchronizer/src/tests/fuzzing.rs
```diff
@@ -0,0 +1,226 @@
...
+    let msg = match lcs::from_bytes::<StateSynchronizerMsg>(&data) {
```
mimoo, line 54: why not use the proptest with the bytes like with the mempool fuzzer?

The problem with this is that it'll more often than not be an early return, whereas a proptest generated value from the data will always works

state-synchronizer/src/tests/fuzzing.rs
christinacelee: with &[u8] Rust complains about lifetime later when I pass it into the `block_on` call

state-synchronizer/src/tests/fuzzing.rs
christinacelee: but in the fuzzer I'm calling the same method right? `state_sync_msg_strategy()`? or is there another fuzzing workflow?

christinacelee: oh i don't LOL, I thought i did..

state-synchronizer/src/tests/fuzzing.rs
```diff
@@ -0,0 +1,226 @@
...
+        Ok(msg) => msg,
+        Err(_) => return,
+    };
+
+    // start up coordinator
+    let (_coordinator_sender, coordinator_receiver) = mpsc::unbounded();
+    let (mempool_sender, _mempool_receiver) = mpsc::channel(1_024);
```
mimoo, line 61: so two thoughts:

- maybe using these large buffers or unbounded channels will slow down the fuzzer
- since the receiver is not used, there's the possibility of the fuzzer blocking (I'm guessing that's why the first one is unbounded)

I think a better approach would be to have fake channels where sending is just a no-op. That might be tricky to do though since a sender doesn't implement a sending trait... food for thoughts

state-synchronizer/src/tests/fuzzing.rs
```diff
@@ -0,0 +1,226 @@
...
+    prop_oneof![get_chunk_request_strategy(), get_chunk_response_strategy()]
+}
+
+fn get_chunk_request_strategy() -> impl Strategy<Value = StateSynchronizerMsg> {
+    (any::<GetChunkRequest>()).prop_map(|get_chunk_request| {
+        StateSynchronizerMsg::GetChunkRequest(Box::new(get_chunk_request))
+    })
+}
+
+impl Arbitrary for GetChunkRequest {
```
mimoo, line 112: should this implementation live next to where the type is defined?

state-synchronizer/src/tests/fuzzing.rs
```diff
@@ -0,0 +1,226 @@
...
+    type Strategy = BoxedStrategy<Self>;
+}
+
+impl Arbitrary for TargetType {
+    type Parameters = ();
+    fn arbitrary_with(_args: ()) -> Self::Strategy {
+        prop_oneof![
+            target_ledger_info_strategy(),
+            highest_available_strategy(),
+            waypoint_strategy()
```
mimoo, line 136: you could inline some of these when they're simple

state-synchronizer/src/tests/fuzzing.rs
christinacelee: this throws a compile error because the fuzzing/testing dependency are tightly correlated, so I have to keep all the fuzzing dependencies together. That's probably also the reason why we have a separate `proptest_types` under `types`

state-synchronizer/src/tests/fuzzing.rs
```diff
@@ -0,0 +1,226 @@
...
+        Waypoint::default(),
+        config.state_sync,
+        config.upstream,
+        MockExecutorProxy::new(SynchronizerEnvHelper::default_handler(), storage_proxy),
+        initial_state,
+    );
+    let mut rt = tokio::runtime::Runtime::new().unwrap();
+    rt.block_on(async move {
+        coordinator
+            .process_one_message(PeerNetworkId(node_network_id, PeerId::random()), msg)
```
mimoo, line 97: avoid non-determinism, I suggest just creating a lazy all zeros peerid

```rust
pub static PEER_ID: Lazy<PeerId> =
    Lazy::new(|| PeedId::new([0u8; PeerId::LENGTH]));
```

state-synchronizer/src/tests/fuzzing.rs
mimoo: I think you copied the consensus fuzzer which is sorta out of date now :p the flow you used in the mempool fuzzer is the better one

state-synchronizer/src/tests/fuzzing.rs
```diff
@@ -0,0 +1,226 @@
...
+#[test]
+fn test_state_sync_fuzzer() {
+    let mut gen = ValueGenerator::new();
+    let data = generate_corpus(&mut gen);
+    fuzzer(data);
+}
+
+pub fn generate_corpus(gen: &mut ValueGenerator) -> Vec<u8> {
+    let msg = gen.generate(state_sync_msg_strategy());
+    lcs::to_bytes(&msg).expect("[fuzzing] failed to generate corpus for state sync msg")
```
mimoo, line 50: I would replace this with the way you do it in the mempool fuzzer as well

state-synchronizer/src/tests/fuzzing.rs
christinacelee: SyncCoordinator requires a very specific type of channel, so I don't think I can write a mocked channel I can switch it out for without actually changing the SyncCoordinator implementation...

mimoo: yeah I ran into this problem elsewhere. I really don't understand why a `Sender` channel wouldn't implement a `TSender` trait : c

I think I've seen workarounds where people use a certain type depending on the flags. For a channel it would look like this perhaps:

```rust
#[cfg(not(feature = "fuzzing"))]
type Sender = mpsc::Sender;
#[cfg(feature = "fuzzing"))]
type Sender = fakeSender;
```

a bit ugly though :P

mimoo:
Left you some comments, but the direction looks good to me. Nice job!

@bors-libra delegate+

state-synchronizer/src/tests/fuzzing.rs
christinacelee: I'll explore and put out a follow-up PR if I come up with something good.. but re: the receiver not being used, that channel state sync actually sets a timeout on, so it won't hang forever, so the fuzzer at least won't be hanging for that

PR #5283, opened 2020-07-24
Add sliding nonce resource to LR

moeziniaf:
## Motivation

We need SR in LR for AOS transactions.

Yes

## Test Plan

Existing SR tests.

## Other

I ran cargo run which has included some other markdown changes, which should be fine.


valerini:
Nice!

sblackshear:
Would you mind adding a test for this? A tx that checks `exists<SlidingNonce>(0xa550c18)` should be enough.

tzakian:
@bors-libra r- 

Seems as though cancel doesn't work :(

moeziniaf:
> Would you mind adding a test for this? A tx that checks `exists<SlidingNonce>(0xa550c18)` would be good. @tzakian has recently made it possible to call tx scripts from the functional tests, so we can also test that some of the scripts that require a sliding nonce can now be sent from LR.

Covered in https://github.com/libra/libra/pull/5284 and https://github.com/libra/libra/pull/5321

PR #5284, opened 2020-07-24
Add human name, sliding nonce to Validator owner, operator accounts

moeziniaf:
We need these fields for owner and operator accounts.
Validator operators seemingly don't have any specific resources published to them - I don't have context here so will leave this as a TODO for @valerini


language/e2e-tests/src/tests/validator_set_management.rs
```diff
@@ -181,8 +187,10 @@ fn validator_set_operator_set_key_reconfigure() {
         libra_root_account
             .transaction()
             .script(encode_create_validator_account_script(
+                0,
```
valerini, line 190: This transaction with the previous one has a zero nonce? Is it allowed, is not failing as a "replay"?

language/tools/vm-genesis/src/lib.rs
```diff
@@ -388,8 +388,10 @@ fn create_and_initialize_owners_operators(
         let owner_auth_key = AuthenticationKey::ed25519(&owner_key);
         let owner_account = account_address::from_public_key(owner_key);
         let create_owner_script = transaction_builder::encode_create_validator_account_script(
+            0,
             owner_account,
             owner_auth_key.prefix().to_vec(),
+            vec![],
```
valerini, line 394: It would be super nice to try to put some readable name in here, like b"validator_0", b"validator_1", b"validator_2" if it's not too hard (I was trying to do this part, but may should've rolled out it like you did it would've been faster :) )

And same below for operators: b"operator_0", b"operator_1", ...

language/stdlib/modules/LibraAccount.move
```diff
@@ -883,18 +883,20 @@ module LibraAccount {
...
-        ValidatorConfig::publish(&new_account, creator_account);
+        ValidatorConfig::publish(&new_account, creator_account, human_name);
         make_account(new_account, auth_key_prefix)
     }
 
     public fun create_validator_operator_account(
         creator_account: &signer,
         new_account_address: address,
         auth_key_prefix: vector<u8>,
+        _human_name: vector<u8>,
```
valerini, line 899: I am not sure what is the best way to add names: we can add another module `ValidatorOperatorConfig` with a human name in it, but I think a better solution is to add a universal module `NamedAccount` and make both validator accounts and validator operators accounts instantiate it. @sblackshear, what do you think?

language/e2e-tests/src/tests/validator_set_management.rs
moeziniaf: So 0 is actually a special nonce which is always accepted :) (look at SlidingNonce module) - so we can just 0 for tests.

language/stdlib/modules/LibraAccount.move
sblackshear: Hmm, good question. I do like `NamedAccount`. The problem is that we already have `DualAttestation::Credential`, which has human-readable names for VASPs/DDs. We could take the names out of there and use `NamedAccount` instead (and I think that would be better), but we can't because that would be a client-breaking change...

I think given that constraint, the most consistent way is probably to have the names live in the role-specific config? Maybe `ValidatorOperatorConfig` could be part of `ValidatorConfig` instead of a new module, and we could add some other useful info to it. E.g., we could add the number of validators that each operator is currently responsible for; that would let us do some on-chain enforcement of the "an operator shouldn't run > 1/3 nodes" rule.

language/tools/vm-genesis/src/lib.rs
sblackshear: Agree with this and would like to see it here! Ensuring that the `human_name` of each validator/operator is unique seems important to avoid confusion.

sblackshear:
Please make sure the prover tests pass before landing. The #move-prover channel can help if the errors aren't clear.

PR #5308, opened 2020-07-24
[crypto][test-only] Turns comments on type safety in cross_test into tests

huitseeker:
- turns the claims on things being impossible to write in `cross_test` into actual compilation tests (see individual commit message). These tests matter even if we don't use several schemes, as they show a trait implementer can't "fool" us on expected types.
- two comment and annotation-only lints

### DIFF
--- a/crypto/crypto/src/unit_tests/compilation/cross_test.rs
+++ b/crypto/crypto/src/unit_tests/compilation/cross_test.rs
@@ -1,0 +1,29 @@
+// Copyright (c) The Libra Core Contributors
+// SPDX-License-Identifier: Apache-2.0
+
+use libra_crypto::{
+    ed25519::{Ed25519PrivateKey, Ed25519PublicKey},
+    multi_ed25519::{MultiEd25519PrivateKey, MultiEd25519PublicKey},
+    test_utils::KeyPair,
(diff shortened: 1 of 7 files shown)


crypto/crypto/src/unit_tests/compilation/cross_test.rs
```diff
@@ -0,0 +1,29 @@
...
+use rand::{prelude::ThreadRng, thread_rng};
+use serde::{Deserialize, Serialize};
+
+#[derive(CryptoHasher, LCSCryptoHash, Serialize, Deserialize)]
+struct TestTypedSemantics(String);
+
+fn main() {
+    let mut csprng: ThreadRng = thread_rng();
+    let ed25519_keypair: KeyPair<Ed25519PrivateKey, Ed25519PublicKey> =
+        KeyPair::generate(&mut csprng);
```
kchalkias, line 20: I would slightly prefer a fixed KeyPair here, but I'm fine either way.

huitseeker: The `ThreadRng` used instead of an `OsRng` significantly decreases the cost of producing this.

PR #5309, opened 2020-07-24
[logging] moved structured logging to use TCP instead of UDP

gregnazario:
### Background
In order to decrease the likelihood of dropped messages, we're moving  to
TCP instead of UDP to send messages.  This will ensure that the
successfully sent messages actually make it, but we will still be
dropping messages in the event of overloads or errors.

For appropriate config changes:
https://github.com/libra/partners/pull/239

### Testing
I thoroughly tested this with ClusterTest.  TCP works!  And the messages do in fact make it to the cluster for querying.  It seems like we may be dropping data on the ES side still due to improperly formed JSON blobs.

I added metrics to track whether we dropped any logs through sending (we don't), but it turns out we get much less logs than we submit in Elasticsearch.  Which can be grabbed with a query like below:
```
GET _count
{
  "query": {
    "range": {
      "@timestamp": {
        "gte": "2020-07-28T18:19:00Z",
        "lt": "2020-07-28T18:26:06Z"
      }
    }
  }
}
```

I suspect we still have work to figure out why we're getting partial logs, but it's hard to tell since our submission is fluent-bit in cluster test and logstash elsewhere.

### DIFF
--- a/common/logger/src/struct_log.rs
+++ b/common/logger/src/struct_log.rs
@@ -1,6 +1,10 @@
 // Copyright (c) The Libra Core Contributors
 // SPDX-License-Identifier: Apache-2.0
-
+use crate::counters::{
+    PROCESSED_STRUCT_LOG_COUNT, SENT_STRUCT_LOG_COUNT, STRUCT_LOG_CONNECT_ERROR_COUNT,
+    STRUCT_LOG_PARSE_ERROR_COUNT, STRUCT_LOG_QUEUE_ERROR_COUNT, STRUCT_LOG_SEND_ERROR_COUNT,
+    STRUCT_LOG_TCP_CONNECT_COUNT,
+};
 use chrono::Utc;
 use once_cell::sync::Lazy;
 use serde::Serialize;
@@ -419,6 +423,7 @@ impl TCPStructLog {
 impl StructLogSink for TCPStructLog {
     fn send(&self, entry: StructuredLogEntry) {
         if let Err(e) = self.sender.try_send(entry) {
+            STRUCT_LOG_QUEUE_ERROR_COUNT.inc();
             // Use log crate macro to avoid generation of structured log in this case
             // Otherwise we will have infinite loop
             log::error!("[Logging] Failed to send structured log: {}", e);
@@ -440,14 +445,19 @@ impl TCPStructLogThread {
     // Continually iterate over requests unless writing fails.
     fn process_requests(&self, stream: &mut TcpStream) {
         for entry in &self.receiver {
+            PROCESSED_STRUCT_LOG_COUNT.inc();
             // Parse string, skipping over anything that can't be parsed
             let json_string = match entry.to_json_string() {
                 Ok(json_string) => json_string,
-                Err(_) => continue,
+                Err(_) => {
+                    STRUCT_LOG_PARSE_ERROR_COUNT.inc();
+                    continue;
+                }
             };
 
             // If we fail to write, exit out and create a new stream
             if let Err(e) = stream.write(json_string.as_bytes()) {
+                STRUCT_LOG_SEND_ERROR_COUNT.inc();
                 println!(
                     "[Logging] Error while sending data to logstash({}): {}",
                     self.address, e
@@ -460,6 +470,8 @@ impl TCPStructLogThread {
 
                 // Start over stream
                 return;
+            } else {
+                SENT_STRUCT_LOG_COUNT.inc()
             }
         }
     }
@@ -479,6 +491,7 @@ impl TCPStructLogThread {
     pub fn run(self) {
         loop {
             let mut maybe_stream = self.connect();
+            STRUCT_LOG_TCP_CONNECT_COUNT.inc();
 
             // This is to ensure that we do actually connect before sending requests
             // If the request process loop ends, the stream is broken.  Reset and create a new one.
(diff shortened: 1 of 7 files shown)


common/logger/src/struct_log.rs
```diff
@@ -426,25 +421,34 @@ impl StructLogSink for UDPStructLog {
...
             };
-            if let Err(e) = self
-                .socket
-                .send_to(json_string.as_bytes(), self.udp_address.clone())
-            {
-                // do not break on error, move on to the next log message
-                println!("[Logging] Error while sending data to socket: {}", e);
+
+            // TODO: we should be able to share TcpStreams over messages
+            let mut stream = match TcpStream::connect(&self.address) {
```
davidiw, line 439: A few things:

* Add a connection timeout
* Add a write timeout
* Implement connection pooling / reuse -- my experience with vault suggests that the OSes don't like churning through sockets...

gregnazario: They're all coming next!

andll:
1) Did you notice any effects of switching to TCP logging works for cluster test? @changyonghu added some counters, so we can compare how many logs we submit vs how many logs do we receive on ELK side - does switch to TCP affect those numbers?

2) Purely in terms of readability of commit messages, I would flip lines on this commit:
```
[cluster-test] make cluster test more like production …
Use TCP for logging rather than a file
```
E.g. what does commit do goes in headline -  `Use TCP for logging rather than a file`, and _why_ we doing it, can go into description(`make cluster test more like production`)

gregnazario:
> 1. Did you notice any effects of switching to TCP logging works for cluster test? @changyonghu added some counters, so we can compare how many logs we submit vs how many logs do we receive on ELK side - does switch to TCP affect those numbers?
> 2. Purely in terms of readability of commit messages, I would flip lines on this commit:
> 
> ```
> [cluster-test] make cluster test more like production …
> Use TCP for logging rather than a file
> ```
> 
> E.g. what does commit do goes in headline - `Use TCP for logging rather than a file`, and _why_ we doing it, can go into description(`make cluster test more like production`)

I'm using this PR right now for just running on cluster test (See draft PR).  The commit messages are temporary for now, as normally, I would put more descriptive commit messages and descriptions.

I messed up the configuration, and I haven't had TCP work yet

gregnazario:
> 1. Did you notice any effects of switching to TCP logging works for cluster test? @changyonghu added some counters, so we can compare how many logs we submit vs how many logs do we receive on ELK side - does switch to TCP affect those numbers?
> 2. Purely in terms of readability of commit messages, I would flip lines on this commit:
> 
> ```
> [cluster-test] make cluster test more like production …
> Use TCP for logging rather than a file
> ```
> 
> E.g. what does commit do goes in headline - `Use TCP for logging rather than a file`, and _why_ we doing it, can go into description(`make cluster test more like production`)

1. This is problematic, as @changyonghu as notified me that Fluentbit retries and doesn't dedupe logs together.  A clear example is this:
```
Timestamp                               k8s node  category  name
Jul 27, 2020 @ 15:37:33.060	fn-26-0	 logger	 connected
Jul 27, 2020 @ 15:37:33.060	fn-26-0	 logger	 connected
Jul 27, 2020 @ 15:37:33.060	fn-26-0	 logger	 connected
```

As much as I think our code can run through so fast to write a message 3 times in a millisecond, I suspect it's being sent 3 times and accepted by ES.  I think this is what points to that we really should use Kibana and keep it closest to prod, so we know whether these things are happening.

2. I went ahead and did some commit cleanup :)

gregnazario:
@andll @davidiw if you guys could take a look, I've successfully run it all on cluster test, and now have metrics and all for submission of logs.  In the event the logging goes out, we'll at least know why it's going out now.

I also can send you the latest run of the cluster test logs, I just didn't think it needed to be posted here.

common/logger/src/struct_log.rs
```diff
@@ -426,25 +426,84 @@ impl StructLogSink for UDPStructLog {
...
+                    self.address, e
+                );
+
+                // Attempt to clear the buffer and send whatever we can
+                if let Err(e) = stream.flush() {
+                    println!("[Logging] Failed to flush rest of buffer: {}", e);
+                }
+
+                // Start over stream
+                return;
```
davidiw: we'll lose the log here, no?

davidiw: should we be sending this to our other logger?

common/logger/src/struct_log.rs
```diff
@@ -281,13 +284,16 @@ pub fn struct_logger_set() -> bool {
...
 /// Can only be called once
 pub fn init_struct_log_from_env() -> Result<(), InitLoggerError> {
     if let Ok(file) = env::var("STRUCT_LOG_FILE") {
         init_file_struct_log(file)
-    } else if let Ok(udp_address) = env::var("STRUCT_LOG_UDP_ADDR") {
-        init_udp_struct_log(udp_address)
+    } else if let Ok(address) = env::var("STRUCT_LOG_TCP_ADDR") {
+        init_tcp_struct_log(address)
+    } else if let Ok(address) = env::var("STRUCT_LOG_UDP_ADDR") {
+        // Remove once all usages of STRUCT_LOG_UDP_ADDR are transferred over
```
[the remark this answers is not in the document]
gregnazario: This whole "logging", with no logger is tricky.  So, I added metrics to take its place since the revision you looked at.  I did a grep of all of our files for TCP_ADDR vs UDP_ADDR, and I think we should be fine, between the two PRs

common/logger/src/struct_log.rs
gregnazario: It now sends metrics that says it crashed :)

common/logger/src/struct_log.rs
```diff
@@ -396,55 +406,122 @@ impl FileStructLogThread {
...
-    udp_address: String,
+    address: SocketAddr,
 }
 
-impl UDPStructLogThread {
-    pub fn run(self) {
-        for entry in self.receiver {
+impl TCPStructLogThread {
+    fn new(receiver: Receiver<StructuredLogEntry>, address: &str) -> TCPStructLogThread {
+        let address = SocketAddr::from_str(address).unwrap();
```
mgorven, line 441: Does this support DNS hostnames? I'm getting this error when updating k8s to include this change:
```ERROR 2020-07-29 17:11:16 common/crash-handler/src/lib.rs:38 details = '''panicked at 'called `Result::unwrap()` on an `Err` value: AddrParseError(())', common/logger/src/struct_log.rs:441:23'''```

gregnazario: It does not :(  I've updated here with resolving DNS https://github.com/libra/libra/pull/5363

PR #5310, opened 2020-07-24
[Config] Update test config data to use "owner_account" for network identities.

JoshLind:
## Motivation

This PR updates the test configs to use the "owner_account" instead of the "operator_account" for network identities. To prevent this from causing issues again, I've also modified the genesis config builder to leverage the default network identities provided to it (through the given config). As a result, if the test configs contain invalid network identities, the smoke tests will now break. 

Note: if the test configs are going to be used as a source of truth, we should probably go through them to verify that all values are indeed actually correct/representative of values that make sense. It wouldn't surprise me if we have other "inconsistencies".

Yes.

## Test Plan

By modifying the genesis config builder to leverage the default identities provided to it, I was able to break the smoke tests. I then fixed the smoke tests by updating the test config data to use the "owner_account" instead of the "operator_account".

## Related PRs

None.

### DIFF
--- a/config/management/genesis/src/config_builder.rs
+++ b/config/management/genesis/src/config_builder.rs
@@ -141,14 +141,17 @@ impl<T: AsRef<Path>> ValidatorBuilder<T> {
             )
             .unwrap();
 
+        let validator_identity = validator_network.identity_from_storage();
         validator_network.identity = Identity::from_storage(
-            libra_global_constants::VALIDATOR_NETWORK_KEY.into(),
-            libra_global_constants::OWNER_ACCOUNT.into(),
+            validator_identity.key_name,
+            validator_identity.peer_id_name,
             self.secure_backend(&local_ns, "validator"),
         );
+
+        let fullnode_identity = fullnode_network.identity_from_storage();
         fullnode_network.identity = Identity::from_storage(
-            libra_global_constants::FULLNODE_NETWORK_KEY.into(),
-            libra_global_constants::OWNER_ACCOUNT.into(),
+            fullnode_identity.key_name,
+            fullnode_identity.peer_id_name,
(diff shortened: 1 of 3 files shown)


config/management/genesis/src/config_builder.rs
```diff
@@ -141,16 +141,27 @@ impl<T: AsRef<Path>> ValidatorBuilder<T> {
...
+            validator_network.identity = Identity::from_storage(
+                identity.key_name,
+                identity.peer_id_name,
+                self.secure_backend(&local_ns, "validator"),
+            );
+        } else {
+            panic!(
+                "Invalid identity found for the validator network! Expected a storage identity."
+            );
+        };
```
gregnazario, line 154: Okay, so I guess there will always be an identity then?

config/management/genesis/src/config_builder.rs
```diff
@@ -141,16 +141,27 @@ impl<T: AsRef<Path>> ValidatorBuilder<T> {
...
-            libra_global_constants::VALIDATOR_NETWORK_KEY.into(),
-            libra_global_constants::OWNER_ACCOUNT.into(),
-            self.secure_backend(&local_ns, "validator"),
-        );
-        fullnode_network.identity = Identity::from_storage(
-            libra_global_constants::FULLNODE_NETWORK_KEY.into(),
-            libra_global_constants::OWNER_ACCOUNT.into(),
-            self.secure_backend(&local_ns, "full_node"),
-        );
+        if let Identity::FromStorage(identity) = validator_network.identity.clone() {
```
davidiw, line 144: why not just add a function called identity_from_storage in the NetworkConfig that does this so you don't end up with copypasta?

JoshLind:
Thanks, all! @bors-libra r=davidiw

PR #5311, opened 2020-07-25
[logging] Removed debug output from ExecutedBlock

gregnazario:
ExecutedBlock debug output was too long for structured logs.  This
makes the functionality the same as Display, which is the same as Block.


zekun000:
@lightmark ^^ want to improve the state compute print?

davidiw:
Perhaps the better choice is to remove cases where we print the debug output?

gregnazario:
> Perhaps the better choice is to remove cases where we print the debug output?

The structured logs converter prints the debug output of every object passed in.  The other option is to put `to_string()` on every log statement

PR #5313, opened 2020-07-25
[cluster-test][node launcher] move fluentbit configuration into spawn_new_instance

changyonghu:
move fluentbit configuration into spawn_new_instance

## Motivation

(Write your answer here.)

## Test Plan

## Related PRs


testsuite/cluster-test/src/cluster_swarm/cluster_swarm_kube.rs
```diff
@@ -660,6 +679,53 @@ impl ClusterSwarmKube {
...
+                key: format!("data/{}/{}/fluent-bit/fluent-bit.conf", run_id, pod_name),
+                body: Some(fluentbit_config.into_bytes().into()),
+                ..Default::default()
+            })
+            .await?;
+
+        let dir = "/opt/libra/data/fluent-bit/";
+        self.util_cmd(format!("mkdir -p {}", dir), node, "mkdir-fluentbit")
+            .await?;
+        self.util_cmd(
```
andll, line 716: 1) For the `run_id`, can we use same value across all calls in the same run? I would propose this:
- In cti, there is pod_name(that sets name of cluster test pod, this is unique across al cluster test runs) and $WORKSPACE. You can pass env var(with -E), like RUN_ID to the cluster test container by combining this two, e.g. RUN_ID=$WORKSPACE-$pod_name.
- Then in cluster test code you can read run id with env::var

2) Since we are going to reuse this later for genesis, etc, do you mind adding method to KubeSwarm, like `put_file(path, content)`
And have this method doing put_object and then s3 cp. object name for s3 can be derived from host path(e.g. if we put file to `/opt/libra/data/fluent-bit/xxx`, you can derive s3 path as `data/{run}/{pod}/opt/libra/data/fluent-bit/xxx`)

changyonghu:
> Great work!
> 
> I assume you confirmed that logs are still working after this change (both for safety rules and validator/fullnode)

Hmm, I am running CT with this change. I will check everything works before merging.

testsuite/cluster-test/src/cluster_swarm/cluster_swarm_kube.rs
```diff
@@ -660,6 +679,65 @@ impl ClusterSwarmKube {
...
+        pod_name: &str,
+        path: &str,
+        content: Vec<u8>,
+    ) -> Result<()> {
+        let bucket = "toro-cluster-test-flamegraphs";
+        let run_id = env::var("RUN_ID").expect("RUN_ID is not set.");
+        self.s3_client
+            .put_object(PutObjectRequest {
+                bucket: bucket.to_string(),
+                key: format!("data/{}/{}/{}", run_id, pod_name, path),
```
andll, line 695: One more ask - do you mind updating path for flamegraphs, to use similar name like `/flamegraphs/{run_id}-libra-node-perf.svg`
E.g. (a) move it to flamegraphs section, and (b) user run-id instead of random id

changyonghu: Sure.

changyonghu: https://toro-cluster-test-flamegraphs.s3-us-west-2.amazonaws.com/flamegraphs/ct-0-cluster-test-changyonghu-1596236068-libra-node-perf.svg

PR #5314, opened 2020-07-25
[management] improve validate-transaction

davidiw:
This will give us a rich response upon execution, no response if it has
yet to execute, and an error otherwise. Versus the previous step that
merely informed you that the transaction had been executed.

This is quite ugly in that we do not have the LCS encoded transaction
retrieval yet, but it will do for now.

### DIFF
--- a/config/management/operational/src/validate_transaction.rs
+++ b/config/management/operational/src/validate_transaction.rs
@@ -2,6 +2,7 @@
 // SPDX-License-Identifier: Apache-2.0
 
 use libra_management::{error::Error, json_rpc::JsonRpcClientWrapper};
+use libra_secure_json_rpc::VMStatusView;
 use libra_types::account_address::AccountAddress;
 use structopt::StructOpt;
 
@@ -17,9 +18,8 @@ pub struct ValidateTransaction {
 
 /// Returns `true` if we've passed by the expected sequence number
 impl ValidateTransaction {
-    pub fn execute(self) -> Result<bool, Error> {
-        let client = JsonRpcClientWrapper::new(self.host);
-        let sequence_number = client.sequence_number(self.account_address)?;
-        Ok(sequence_number >= self.sequence_number)
+    pub fn execute(self) -> Result<Option<VMStatusView>, Error> {
+        JsonRpcClientWrapper::new(self.host)
+            .transaction_status(self.account_address, self.sequence_number)
     }
 }
--- a/secure/json-rpc/Cargo.toml
+++ b/secure/json-rpc/Cargo.toml
@@ -11,6 +11,7 @@ edition = "2018"
 
 [dependencies]
 hex = "0.4.2"
+serde = { version = "1.0.114", features = ["derive"], default-features = false }
 serde_json = "1.0.56"
 thiserror = "1.0.20"
 ureq = { version = "1.3.0", features = ["json"] }
@@ -19,7 +20,6 @@ lcs = { path = "../../common/lcs", version = "0.1.0", package = "libra-canonical
 libra-crypto = { path = "../../crypto/crypto", version = "0.1.0" }
 libra-types = { path = "../../types", version = "0.1.0" }
 libra-workspace-hack = { path = "../../common/workspace-hack", version = "0.1.0" }
-serde = { version = "1.0.114", features = ["derive"], default-features = false }
 
 [dev-dependencies]
 anyhow = "1.0.31"
--- a/secure/json-rpc/src/lib.rs
+++ b/secure/json-rpc/src/lib.rs
@@ -148,6 +148,34 @@ impl JsonRpcClient {
         }
     }
 
+    /// Retrieves the status of a transaction on a given account.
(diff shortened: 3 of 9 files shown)


secure/json-rpc/Cargo.toml
```diff
@@ -11,15 +11,16 @@ edition = "2018"
...
 [dependencies]
 hex = "0.4.2"
+serde = { version = "1.0.114", features = ["derive"], default-features = false }
 serde_json = "1.0.56"
 thiserror = "1.0.20"
 ureq = { version = "1.3.0", features = ["json"] }
 
 lcs = { path = "../../common/lcs", version = "0.1.0", package = "libra-canonical-serialization" }
 libra-crypto = { path = "../../crypto/crypto", version = "0.1.0" }
+libra-json-rpc-types = { path = "../../json-rpc/types", version = "0.1.0"}
```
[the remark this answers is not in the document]
davidiw: This should have been removed, but those types are already there, what it also brings in is the VM :/

testsuite/tests/libratest/smoke_test.rs
```diff
@@ -1462,27 +1462,26 @@ fn get_libra_interface(node_config: &NodeConfig) -> JsonRpcLibraInterface {
...
-fn verify_transaction_executed(op_tool: OperationalTool, transaction_context: TransactionContext) {
-    op_tool
-        .validate_transaction(
-            transaction_context.address,
-            transaction_context.sequence_number,
-        )
-        .unwrap();
+    while !call() {
+        sleep(Duration::from_secs(1));
+    }
```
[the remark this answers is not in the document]
davidiw: Sure

davidiw:
@JoshLind and @matbd, this PR now only contains changes for management, I've moved the transaction-builder to another PR. There really was no good reason to bundle them. I've also taken care of your concerns, Josh. Please review again when you get a chance.

The other PR: https://github.com/libra/libra/pull/5319

testsuite/tests/libratest/smoke_test.rs
```diff
@@ -1578,18 +1548,12 @@ fn test_network_key_rotation() {
...
         op_tool.rotate_validator_network_key(&backend).unwrap();
-
-    // Ensure all nodes have been reconfigured
-    wait_for_all_nodes(&libra_interfaces, |libra| {
-        libra.last_reconfiguration().unwrap() > last_reconfig
-    });
-
-    verify_transaction_executed(op_tool, transaction_context);
+    let mut client = swarm.get_validator_client(0, None);
+    client.wait_for_transaction(txn_ctx.address, txn_ctx.sequence_number + 1).unwrap();
```
gregnazario, line 1556: Woo, love to see shrunk tests

secure/json-rpc/src/lib.rs
```diff
@@ -148,6 +148,34 @@ impl JsonRpcClient {
...
+        let params = vec![
+            Value::String(account.to_string()),
+            json!(sequence_number),
+            json!(false),
+        ];
+        let response = self.execute_request(method, params);
+
+        match response.status() {
+            200 => {
+                let response = response.into_string()?;
```
gregnazario, line 167: Nit: Could probably move `response.into_string()?` before the match.

config/management/operational/src/validate_transaction.rs
```diff
@@ -17,9 +18,8 @@ pub struct ValidateTransaction {
...
 /// Returns `true` if we've passed by the expected sequence number
 impl ValidateTransaction {
-    pub fn execute(self) -> Result<bool, Error> {
-        let client = JsonRpcClientWrapper::new(self.host);
-        let sequence_number = client.sequence_number(self.account_address)?;
-        Ok(sequence_number >= self.sequence_number)
+    pub fn execute(self) -> Result<Option<VMStatusView>, Error> {
+        JsonRpcClientWrapper::new(self.host)
+            .transaction_status(self.account_address, self.sequence_number)
     }
```
gregnazario, line 24: So this isn't really `ValidateTransaction` anymore, as much as `TransactionStatus`

davidiw: or check transaction so that it implies an action / verb instead of a state or noun

davidiw: I'm kinda inclined that validate is okay :P

secure/json-rpc/src/lib.rs
davidiw: you could if into_string didn't consume response leaving no status to be read :/

testsuite/tests/libratest/smoke_test.rs
davidiw: This is kind of a bummer, in the ideal world, the code wouldn't have taken this interesting yet divergent path. The existing cli tool isn't very modular, so we cannot just take arbitrary pieces of it :/. I do look forward to the point in time where we can embrace async in the TCB code, because the overhead of maintaining both sync and async isn't fun or good.

PR #5315, opened 2020-07-25
[Operational Management] Add support for verifying validator config and validator infos on chain.

JoshLind:
## Motivation

This PR adds support for two new commands to the operational management tool, validator-config(..) and validator-set(..). These commands print out the current validator config for a given account address, as well as the validator set infos for a given account address, respectively. This PR offers 4 commits:

1. Add the new validator-config() command to the operational management tool.
2. Update the key rotation smoke tests to use the operational management tool to verify the validator config states on-chain after a key rotation.
3. Add the new validator-info() command to the operational management tool.
4. Update the key rotation smoke tests to use the operational management tool to verify the validator set info for the specific account address has been updated after the rotation.

Note: I'm planning on doing a refactor PR in the future around the smoke tests and some of the management tooling, but for now, let's move towards feature complete on this.

Yes.

## Test Plan

The updated smoke tests pass. Moreover, running the tool manually displays the on-chain validator config and validator infos.

## Related PRs

None.

### DIFF
--- a/config/management/operational/src/test_helper.rs
+++ b/config/management/operational/src/test_helper.rs
@@ -7,7 +7,9 @@ use libra_crypto::{ed25519::Ed25519PublicKey, x25519};
 use libra_management::{error::Error, secure_backend::DISK, TransactionContext};
 use libra_network_address::NetworkAddress;
 use libra_secure_json_rpc::VMStatusView;
-use libra_types::{account_address::AccountAddress, chain_id::ChainId};
+use libra_types::{
+    account_address::AccountAddress, chain_id::ChainId, validator_config::ValidatorConfig,
+};
 use structopt::StructOpt;
 
 const TOOL_NAME: &str = "libra-operational-tool";
@@ -116,6 +118,25 @@ impl OperationalTool {
 
         let command = Command::from_iter(args.split_whitespace());
         command.validate_transaction()
+    }
+
+    pub fn validator_config(
+        &self,
+        account_address: AccountAddress,
+    ) -> Result<ValidatorConfig, Error> {
+        let args = format!(
+            "
+                {command}
+                --host {host}
+                --account-address {account_address}
+        ",
+            command = command(TOOL_NAME, CommandName::ValidatorConfig),
+            host = self.host,
+            account_address = account_address,
+        );
+
+        let command = Command::from_iter(args.split_whitespace());
+        command.validator_config()
     }
 }
 
--- a/config/management/operational/src/test_helper.rs
+++ b/config/management/operational/src/test_helper.rs
@@ -9,6 +9,7 @@ use libra_network_address::NetworkAddress;
 use libra_secure_json_rpc::VMStatusView;
 use libra_types::{
     account_address::AccountAddress, chain_id::ChainId, validator_config::ValidatorConfig,
+    validator_info::ValidatorInfo,
 };
 use structopt::StructOpt;
 
@@ -137,6 +138,25 @@ impl OperationalTool {
 
         let command = Command::from_iter(args.split_whitespace());
         command.validator_config()
+    }
+
+    pub fn validator_set(
+        &self,
+        account_address: AccountAddress,
+    ) -> Result<Vec<ValidatorInfo>, Error> {
+        let args = format!(
+            "
+                {command}
+                --host {host}
+                --account-address {account_address}
+        ",
+            command = command(TOOL_NAME, CommandName::ValidatorSet),
+            host = self.host,
+            account_address = account_address,
+        );
+
+        let command = Command::from_iter(args.split_whitespace());
+        command.validator_set()
     }
 }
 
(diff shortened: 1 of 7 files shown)


config/management/operational/src/test_helper.rs
```diff
@@ -6,7 +6,10 @@ use libra_config::config;
 use libra_crypto::{ed25519::Ed25519PublicKey, x25519};
 use libra_management::{error::Error, secure_backend::DISK, TransactionContext};
 use libra_network_address::NetworkAddress;
-use libra_types::{account_address::AccountAddress, chain_id::ChainId};
+use libra_types::{
```
davidiw, line 9: we need to make this closer to storage_helper in the genesis tooling...

config/management/src/json_rpc.rs
```diff
@@ -48,6 +49,48 @@ impl JsonRpcClientWrapper {
         .ok_or_else(|| Error::JsonRpcReadError("validator-config", "not present".to_string()))
     }
 
+    /// This method returns all validator infos currently registered in the validator set of the
+    /// Libra blockchain. If account is specified, only a single validator info is returned: the
+    /// one that matches the given account.
+    pub fn validator_set(
```
davidiw, line 55: if we only use this logic once, maybe it is better to figure out how to get it into the actual place where it is being invoked instead of having these tiny command files above?

config/management/operational/src/validator_config.rs
```diff
@@ -184,3 +185,18 @@ fn decode_address(raw_address: RawNetworkAddress) -> Result<NetworkAddress, Erro
         .collect::<Vec<_>>();
     Ok(NetworkAddress::try_from(protocols).unwrap())
 }
+
+#[derive(Debug, StructOpt)]
+pub struct ValidatorConfig {
+    #[structopt(long, help = "JSON-RPC Endpoint (e.g. http://localhost:8080")]
+    host: String,
```
davidiw, line 192: I think we should rename this to json_server, sinceh ost is kinda ambiguous. That's what I'm naming it in the config and will update all places after I make my hopeful progress.

config/management/operational/src/test_helper.rs
```diff
@@ -116,6 +119,44 @@ impl OperationalTool {
...
+
+    pub fn validator_set(
+        &self,
+        account_address: AccountAddress,
+    ) -> Result<Vec<ValidatorInfo>, Error> {
+        let args = format!(
+            "
+                {command}
+                --host {host}
+                --account-address {account_address}
```
davidiw, line 150: @zekun000, let's use consistent naming across the tools, I think you call this account_to_add?

config/management/src/json_rpc.rs
```diff
@@ -48,6 +49,48 @@ impl JsonRpcClientWrapper {
...
+                        ));
+                    }
+                    Ok(validator_infos)
+                }
+                None => Err(Error::JsonRpcReadError(
+                    "validator-set",
+                    "not present".to_string(),
+                )),
+            },
+            Err(e) => Err(Error::JsonRpcReadError("validator-set", e.to_string())),
```
gregnazario, line 90: You can chain match clauses, which might make it cleaner / less tabbing
```
match validator_set {
  Ok(Some(validator_set)) => { ... },
  Ok(None) => {...},
  Err(e) => {...},
}
```

Tested with rust playground:
https://gist.github.com/rust-play/388c798f77702ad287aee82d23229d1f

config/management/operational/src/validator_set.rs
```diff
@@ -0,0 +1,21 @@
...
+use structopt::StructOpt;
+
+#[derive(Debug, StructOpt)]
+pub struct ValidatorSet {
+    #[structopt(long, help = "JSON-RPC Endpoint (e.g. http://localhost:8080")]
+    host: String,
+    #[structopt(long, help = "AccountAddress to retrieve the validator set info")]
+    account_address: Option<AccountAddress>,
+}
+
```
gregnazario, line 15: We'll have to come up with a better organizational structure than the tiny files (my fault) at some point.  Maybe by some sort of grouping.

config/management/src/json_rpc.rs
JoshLind: Ooh, nice! Thanks 😄

config/management/src/json_rpc.rs
JoshLind: Yeah, we should work on this in a cleanup PR.

config/management/operational/src/test_helper.rs
JoshLind: Okay. Will leave this for a follow up PR.

config/management/operational/src/validator_config.rs
JoshLind: Okay, I'll update it here and assume that you'll catch the rest of the places 😄

JoshLind:
Thanks, all! 😄 @bors-libra r=davidiw

PR #5316, opened 2020-07-25
[faucet] propagate CLI update to server

sausagee:
## Motivation
CLI was updated in #5242, but the faucet server was updated and broke.  Propagate the CLI changes in 5120a9047bd to faucet server.

image

## Test Plan
Tested on own cluster

### DIFF
--- a/docker/mint/server.py
+++ b/docker/mint/server.py
@@ -66,7 +66,7 @@ def send_transaction():
         create_client()
         application.client.sendline(
             "a m {} {} {} use_base_units".format(auth_key, amount, currency_code))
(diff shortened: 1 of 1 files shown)


sausagee:
> I'm surprised that we have logic that needs to match a hardcoded string in account_commands.py to setup_scripts/docker/mint/server.py otherwise it breaks faucet... I'm curious why we have this set up this way?

From what I can tell, there is no DD on testnet, so faucet provides a proxy for that and send mint requests to local CLI client that holds the mint key. 
cc @davidiw

PR #5317, opened 2020-07-25
[management] cleanup

davidiw:
This moves files into their respective crates
Eliminates commands that don't make sense for different services (genesis doesn't insert waypoints or use json-rpc, operational will insert waypoints)
Fixes some alpha ordering

Basically trying to clean this up to maybe make it easier to integrate the config work...


config/management/src/secure_backend.rs
```diff
@@ -199,20 +199,6 @@ secure_backend!(
     "shared information"
 );
 
-secure_backend!(
-    OptionalValidatorBackend,
-    validator_backend,
-    Option<SecureBackend>,
-    "validator configuration"
-);
```
gregnazario, line 207: Woo,no more optional backends

config/management/genesis/src/command.rs
```diff
@@ -8,61 +8,53 @@ use structopt::StructOpt;
 #[derive(Debug, StructOpt)]
 #[structopt(about = "Tool used for genesis")]
 pub enum Command {
-    #[structopt(about = "Submits an Ed25519PublicKey for the libra root")]
-    LibraRootKey(libra_management::key::LibraRootKey),
-    #[structopt(about = "Create a waypoint and optionally place it in a store")]
-    CreateWaypoint(libra_management::waypoint::CreateWaypoint),
+    #[structopt(about = "Create a waypoint and place it in a store")]
```
gregnazario, line 11: We use `storage` in operational, and `store` in `genesis`.  We should probably do another pass to keep things similar.

config/management/genesis/src/waypoint.rs
```diff
@@ -0,0 +1,67 @@
...
+        let waypoint = CreateWaypoint {
+            shared_backend: self.shared_backend,
+        }
+        .execute()?;
+        let storage_name = self.validator_backend.name();
+        let mut validator_storage =
+            StorageWrapper::new(storage_name, &self.validator_backend.validator_backend)?;
+        validator_storage.set(WAYPOINT, Value::String(waypoint.to_string()))?;
+        Ok(waypoint)
+    }
```
gregnazario, line 66: I would say we could share this between "CreateAndInsert" and "Insert".  But, it's like 4 lines and probably not worth it.

config/management/genesis/src/storage_helper.rs
```diff
@@ -126,17 +100,21 @@ impl StorageHelper {
...
                 libra-genesis-tool
-                insert-waypoint
+                libra-root-key
                 --validator-backend backend={backend};\
                     path={path};\
                     namespace={validator_ns}
                 --shared-backend backend={backend};\
                     path={path};\
-                    namespace={shared_ns}
+                    namespace={shared_ns}\
```
gregnazario, line 117: Any particular reason for the `\`

config/management/genesis/src/command.rs
davidiw: yeah, I agree.

config/management/genesis/src/waypoint.rs
davidiw: hahah that and insert ends up in operational, so we'd have to add it to a library in management. FWIW, we did lose a insert call in config-builder.rs, so it probably makes sense to have a dedicated config-builder crate that can depend on either library.

config/management/genesis/src/storage_helper.rs
davidiw: that's weird... I know for a fact I did not insert that, might be some rebase issue.

PR #5318, opened 2020-07-25
[move-prover] Assumed/Asserted aborts and error codes.

### LINKED ISSUE #4906: [Feature Request] Improvements to aborts_if
# 🚀 Feature Request

Improvements to reduce chance of missed errors from and increase convenience of aborts_if

## Motivation

A missed error  from out-of-bounds vector indexing (Emma found this problem when e2e testing failed)

## Pitch

I propose that we designate certain abort conditions as "impermissible aborts", which are ALWAYS reported as errors by the Move Prover.  Obvious candidates are vector out-of-bounds indices and divide by 0.  On the other hand, overflow conditions and existence of published resources with move_from, move_to, and borrows_global(_mut) would be "permissible aborts", which are errors UNLESS there is an aborts_if condition that accounts for them.

pragma aborts_if_is_partial would permissible aborts but NOT impermissible aborts

 We have debated this issue several times, starting in Summer of 2019, and we decided not to do this because Move programming conventions permitted some abort conditions that were not protected by asserts.  We decided to wait until the programming conventions had stabilized and we had more experience.  We now have that experience, and can see which aborts should be impermissible by inspecting the aborts_ifs that are specified (or implementing them and running "cargo test").

I think this is easy to implement, and should be high priority AFTER 7/15

**Describe alternatives you've considered**

**Are you willing to open a pull request?** (See [CONTRIBUTING](../../CONTRIBUTING.md))

### LINKED ISSUE #5266: [Feature Request] Error codes in aborts_ifs
Some times multiple aborts might happen under the same condition. See issue #5124 for an example where abort can happen because of index out of bound and actual user coded aborts under the same aborting condition. The bug was missed because I only had the second abort in mind while writing the specs but the specs also covered the first abort and thus hiding the otherwise obvious bug.

If we have something like `aborts_with_error ERROR_CODE if P`, we will be able to detect cases where the program aborts under the same condition but with **different** error code.

wrwg:
This PR adds new features around aborts specification, as outlined in the design doc:

- `aborts_if P with code` allows to specify an error code for this aborts condition. If `P` is true and the function aborts, it must abort with `code` otherwise a special verification error is produced. The error message contains detail information about expected and provided error code.  For examples, see [here](https://github.com/wrwg/libra/blob/errors/language/move-prover/tests/sources/functional/aborts_if_with_code.move).
- `aborts_with code1, ..., codeN` allows to specify an aborts code without any condition.
- `aborts_if [assert] P` and `aborts_if [assume] P` allows to specify that the error condition should either become a precondition which needs to be verified to not happen, or an assumption. For examples, see [here](https://github.com/wrwg/libra/blob/errors/language/move-prover/tests/sources/functional/aborts_if_assume_assert.move).

Together with this PR comes a refactoring in spec_translator.rs on how conditions are translated for the different entry points of a function. The new approach uses a more systematic and declarative way how to do this which is easier to understand and modify using the new `ConditionDistribution` data type.

Closes #5266 
Closes #4906

## Motivation

More scalable treatment of aborts.

Yes

## Test Plan

New tests.

## Related PRs

NA

### DIFF
--- a/language/move-prover/spec-lang/src/env.rs
+++ b/language/move-prover/spec-lang/src/env.rs
@@ -112,6 +112,16 @@ pub const CONDITION_ABSTRACT_PROP: &str = "abstract";
 /// Opposite to the abstract property.
 pub const CONDITION_CONCRETE_PROP: &str = "concrete";
 
+/// Property which indicates that an aborts_if should be assumed.
+/// For callers of a function with such an aborts_if, the negation of the condition becomes
+/// an assumption.
+pub const CONDITION_ABORT_ASSUME_PROP: &str = "assume";
+
+/// Property which indicates that an aborts_if should be asserted.
+/// For callers of a function with such an aborts_if, the negation of the condition becomes
+/// an assertion.
+pub const CONDITION_ABORT_ASSERT_PROP: &str = "assert";
+
 /// Pragma which indicates that the functions aborts and ensure conditions shall be exported
 /// to the verification context even if the implementation of the function is inlined.
 pub const EXPORT_ENSURES_PRAGMA: &str = "export_ensures";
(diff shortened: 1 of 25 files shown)


language/move-prover/spec-lang/src/env.rs
```diff
@@ -112,6 +112,16 @@ pub const CONDITION_ABSTRACT_PROP: &str = "abstract";
 /// Opposite to the abstract property.
 pub const CONDITION_CONCRETE_PROP: &str = "concrete";
 
+/// Property which indicates that an aborts_if should be assumed.
+/// For callers of a function with such an aborts_if, the negation of the condition becomes
+/// an assumption./is_st
```
emmazzz, line 117: delete `/is_st`?

wrwg:
This PR has been updated to implement the semantics as discussed yesterday w/ @shazqadeer , @emmazzz , and @DavidLDill. Moreover, the `aborts_with codes` clause has been added as suggested by @sblackshear .

language/move-prover/spec-lang/src/env.rs
wrwg: Done

PR #5319, opened 2020-07-25
Transaction builder

davidiw:
* Split transaction-builder generated code from the richer library so we can minimize dependencies and ...
* Use transaction-builder-generated instead of transaction-scripts allowing us to delete that legacy crate


matbd:
LGTM but remove `language/transaction-builder-generated/src/.lib.rs.swp`

@sblackshear Do you guys prefer `language/transaction-builder-generated` or `language/transaction-builder/generated` for the new crate directory?

Make sure to run `(cd language/stdlib && cargo run --release)` because afaik it is tested in CI yet.

language/transaction-builder-generated/Cargo.toml
```diff
@@ -1,15 +1,17 @@
 [package]
-name = "libra-transaction-scripts"
+name = "transaction-builder-generated"
 version = "0.1.0"
-edition = "2018"
 authors = ["Libra Association <opensource@libra.org>"]
-description = "Libra transaction scripts"
+description = "Libra transaction-builder"
```
matbd, line 5: "Generated transaction builders"?

Actually it would be nice to have `stdlib` in the Rust namespace to be future-proof. (In the future, there could be several such libraries for different bundles of scripts.)

matbd: To be concrete, `language/transaction-builder/generated/src/stdlib.rs` could define the module `stdlib` of crate `transaction-builder-generated`, maybe?

sblackshear:
> @sblackshear Do you guys prefer `language/transaction-builder-generated` or `language/transaction-builder/generated` for the new crate directory?

I think the latter, but either is fine by me

matbd:
> > @sblackshear Do you guys prefer `language/transaction-builder-generated` or `language/transaction-builder/generated` for the new crate directory?
> 
> I think the latter, but either is fine by me

Ok. I could also move `language/transaction-builder-generator` into `language/transaction-builder/generator` later on.

davidiw:
I'm happy to take care of all these things in this PR, no big deal, will try to resolve and land this by EOD.

Thanks folks!

language/stdlib/src/lib.rs
```diff
@@ -40,7 +40,7 @@ pub const COMPILED_TRANSACTION_SCRIPTS_ABI_DIR: &str = "compiled/transaction_scr
 
 /// Where to write generated transaction builders.
 pub const TRANSACTION_BUILDERS_GENERATED_SOURCE_PATH: &str =
-    "../transaction-builder/src/generated.rs";
+    "../transaction-builder-generated/src/stdlib.rs";
```
matbd, line 43: oops

PR #5320, opened 2020-07-25
[json-rpc] fix AccountView#role and TransactionView#vm_status serialization bug

xli:
## Motivation

according to https://github.com/libra/libra/blob/master/json-rpc/json-rpc-spec.md#account---type,
the account#role should always be an object with a field type to indicate what is role type.

For TransactionView#vm_status, doc https://github.com/libra/libra/blob/master/json-rpc/json-rpc-spec.md#transaction---type does not mention type information, but when the type is MoveAbort or ExecutionFailure, the serialization result will be an object with several fields, and other values are string, hence it should behave similar.

## Test Plan

unit test

### DIFF
--- a/json-rpc/types/src/views.rs
+++ b/json-rpc/types/src/views.rs
@@ -47,6 +47,7 @@ impl AmountView {
 }
 
 #[derive(Clone, Serialize, Deserialize, Debug, PartialEq)]
+#[serde(tag = "type")]
 pub enum AccountRoleView {
     #[serde(rename = "unknown")]
     Unknown,
@@ -349,6 +350,7 @@ impl From<&Vec<u8>> for BytesView {
 }
 
 #[derive(Clone, Debug, Deserialize, Serialize, PartialEq)]
+#[serde(tag = "type")]
 pub enum VMStatusView {
     #[serde(rename = "executed")]
     Executed,
--- a/json-rpc/types/src/views.rs
+++ b/json-rpc/types/src/views.rs
@@ -47,29 +47,51 @@ impl AmountView {
 }
 
 #[derive(Clone, Serialize, Deserialize, Debug, PartialEq)]
-#[serde(tag = "type")]
-pub enum AccountRoleView {
-    #[serde(rename = "unknown")]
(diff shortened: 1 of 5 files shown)


xli:
I'll try to work on a no breaking change version, hold on review

xli:
2 commits:
1. first commit is a breaking change: direct change response data as what described in document.
2. second commit is a versioning change: add a <field>_v2 with documented data structure, keep <field> data unchanged to be backward compatible.

I'd prefer to not make a breaking change, hence versioning is better to me. The implementation probably can be improved, current one duplicates the enum definition for adding serde(tag = "type"

json-rpc/types/src/views.rs
```diff
@@ -47,28 +47,51 @@ impl AmountView {
...
-        preburn_balances: Vec<AmountView>,
-        received_mint_events_key: BytesView,
-    },
+    ChildVasp(ChildVASP),
+    ParentVasp(ParentVASP),
+    DesignatedDealer(DesignatedDealer),
+}
+
+#[derive(Clone, Serialize, Deserialize, Debug, PartialEq)]
+#[serde(rename_all = "snake_case", tag = "type")]
```
thefallentree, line 83: instead of relying on the automatic tokenization here , I think we should follow convention in the rest of the code, manually tag all enum

json-rpc/types/src/views.rs
```diff
@@ -81,6 +104,7 @@ pub struct AccountView {
     pub delegated_key_rotation_capability: bool,
     pub delegated_withdrawal_capability: bool,
     pub is_frozen: bool,
+    #[serde(flatten)]
```
thefallentree, line 107: what does this do? this is not obvious from code

thefallentree:
it is unfortunate that we forgot "tag=type" on the new Views's review, but I don't think we should be adding field level  _v1 and _v2 now, we should just start collecting these changes to "/v2 " endpoint , and in an undecided future date making the default endpoint to /v2 instead.

xli:
let's fix it in v2 endpoint

PR #5321, opened 2020-07-26
Move prover follow up and SlidingNonce tests

moeziniaf:
Follow up to https://github.com/libra/libra/pull/5284/files 

- Adds a txn check to ensure SlidingNonce present for LR.
- Adds a test to verify nonce aborts on repeated txn from LR.
- Fixes move-prover spec changes for human_name under ValidatorConfig


language/stdlib/modules/SlidingNonce.move
```diff
@@ -72,6 +72,10 @@ module SlidingNonce {
         move_to(account, SlidingNonce {  min_nonce: 0, nonce_mask: 0 });
     }
 
+    public fun holds_sliding_nonce(addr: address) {
```
valerini, line 75: The name looks like the function should return a boolean.
May be it can be called something like `assert_holds_sliding_nonce`?

language/stdlib/modules/ValidatorConfig.move
```diff
@@ -231,6 +231,17 @@ module ValidatorConfig {
         ensures result == spec_get_operator(addr);
     }
 
+    /// Return the human-readable name for the validator account.
+    /// Aborts if `addr` does not have a `ValidatorConfig` resource.
+    public fun human_name(addr: address): vector<u8> acquires ValidatorConfig {
```
valerini, line 236: May be we can call it `get_human_name` to be consistent with the rest of the getter functions?

language/e2e-tests/src/tests/validator_set_management.rs
```diff
@@ -63,6 +63,24 @@ fn validator_add() {
         .events()
         .iter()
         .any(|e| e.key() == &new_epoch_event_key()));
+
```
sblackshear, line 66: I think this would be easier to read/maintain as a functional test instead of an E2E test (now that we have functional test support for scripts). Check out https://github.com/libra/libra/tree/master/language/move-lang/functional-tests/tests/stdlib_scripts for examples

language/stdlib/modules/SlidingNonce.move
```diff
@@ -72,6 +72,10 @@ module SlidingNonce {
         move_to(account, SlidingNonce {  min_nonce: 0, nonce_mask: 0 });
     }
 
+    public fun assert_holds_sliding_nonce(addr: address) {
```
sblackshear, line 75: I think it would be better to define:
```
public fun has_nonce(addr: address): bool {
  exists<SlidingNonce>(addr)
}
```
? This was
- other modules can check existence without triggering an abort
- we won't need to define an error code that is only used in tests

moeziniaf:
Added functional e2e test, and has_nonce method

language/stdlib/modules/SlidingNonce.move
moeziniaf: Done

language/e2e-tests/src/tests/validator_set_management.rs
moeziniaf: Done

moeziniaf:
Bumping for review

davidiw:
@sblackshear  / @valerini thoughts on this PR?

language/e2e-tests/src/tests/validator_set_management.rs
davidiw: I think Sam was proposing that we not keep this. Personally I don't really mind either way. I think we should be comfortable deleting "duplicate" tests if we're doing a refactor.

davidiw (approved):

davidiw:
Considering this is now 1.5 month olds, gonna close this out.

PR #5323, opened 2020-07-27
[move prover] add support for calling pure functions from specs

emmazzz:
## Motivation

This PR enables users of the move prover to call pure move functions. See the new test case `pure_function_call.move` for examples of allowed functions. I have changed calls to `CoreAddresses` spec functions to pure move functions in `Libra` module.

In the future, we need to implement a more general algorithm to check if some function is pure or not.

Yes.

## Test Plan

Added two test cases.


language/move-prover/spec-lang/src/translate.rs
```diff
@@ -613,6 +616,28 @@ impl<'env> Translator<'env> {
                     result_type: param_t.clone(),
                 },
             );
+            add_builtin(
+                self,
+                self.builtin_fun_symbol("borrow_global"),
```
wrwg, line 621: We can go with this for now, but what is a problem is that users can now also call these functions in their spec funs. They shouldn't be able to do so, because this leads to confusion. Better would be we would be mapping calls to those to global. But as said, for not postponing the PR, we can do it. Please add a TODO.

language/move-prover/spec-lang/src/translate.rs
```diff
@@ -1033,16 +1058,46 @@ impl<'env, 'translator> ModuleTranslator<'env, 'translator> {
...
+            &loc,
+            qsym,
+            self.module_id,
+            spec_fun_id,
+            type_params.iter().map(|(_, ty)| ty.clone()).collect(),
+            params.iter().map(|(_, ty)| ty.clone()).collect(),
+            result_type.clone(),
+        );
+
+        // Add $ to the name so the spec version does not name clash with the move version.
```
wrwg, line 1085: Why is it needed? Aren't those functions disjoint by definition? For the SpecFunEntry, we don't seem to need it.

language/move-prover/spec-lang/src/translate.rs
```diff
@@ -3086,6 +3188,10 @@ pub struct ExpTranslator<'env, 'translator, 'module_translator> {
     outer_context_scopes: usize,
     /// A boolean indicating whether we are translating a let expression
     in_let: bool,
+    /// A flag to indicate whether we are translating expressions in a spec fun.
+    translating_spec_fun: bool,
+    /// A flag to indicate whether have been generated so far.
```
wrwg, line 3193: generated errors

language/move-prover/spec-lang/src/translate.rs
```diff
@@ -3086,6 +3188,10 @@ pub struct ExpTranslator<'env, 'translator, 'module_translator> {
     outer_context_scopes: usize,
     /// A boolean indicating whether we are translating a let expression
     in_let: bool,
+    /// A flag to indicate whether we are translating expressions in a spec fun.
+    translating_spec_fun: bool,
```
wrwg, line 3192: More descriptive, like `translating_fun_as_spec_fun`, and default false?

language/move-prover/spec-lang/src/translate.rs
```diff
@@ -3129,6 +3238,10 @@ impl<'env, 'translator, 'module_translator> ExpTranslator<'env, 'translator, 'mo
         self
     }
 
+    fn flip_translation_mode(&mut self) {
```
wrwg, line 3241: More descriptive, like `translate_fun_as_spec_fun`?

language/move-prover/spec-lang/src/translate.rs
```diff
@@ -4382,6 +4507,43 @@ impl<'env, 'translator, 'module_translator> ExpTranslator<'env, 'translator, 'mo
                 // Construct result.
                 let id = self.new_node_id_with_type_loc(&ty, loc);
                 self.set_instantiation(id, instantiation);
+
+                if let Operation::Function(module_id, spec_fun_id) = cand.oper {
+                    if self.translating_spec_fun {
+                        // Record the usage of function in specs, used later
+                        // in spec translator;
```
wrwg, line 4514: nit: ; -> .

language/move-prover/spec-lang/src/translate.rs
```diff
@@ -4382,6 +4507,43 @@ impl<'env, 'translator, 'module_translator> ExpTranslator<'env, 'translator, 'mo
                 // Construct result.
                 let id = self.new_node_id_with_type_loc(&ty, loc);
                 self.set_instantiation(id, instantiation);
+
+                if let Operation::Function(module_id, spec_fun_id) = cand.oper {
+                    if self.translating_spec_fun {
+                        // Record the usage of function in specs, used later
```
wrwg, line 4513: It's probably a good idea to make clear in such comments what functions you are talking about. Perhaps 'regular function' or 'Move function' for the one, 'spec fun' for the other. Here I'm not sure whether you are tracking spec funs or move funcs (and in the later case, the condition would be wrong.)

language/move-prover/spec-lang/src/translate.rs
```diff
@@ -4382,6 +4507,43 @@ impl<'env, 'translator, 'module_translator> ExpTranslator<'env, 'translator, 'mo
...
+                    if let Some(entry) = self.parent.parent.fun_table.get(&qsym) {
+                        if self.translating_spec_fun && !entry.is_pure {
+                            let display = self.display_call_target(module, name);
+                            let notes = vec![format!(
+                                "impure function `{}`",
+                                self.display_call_cand(module, name, cand),
+                            )];
+                            self.parent.parent.env.error_with_notes(
+                                loc,
+                                &format!("calling impure function `{}` is disallowed", display),
```
wrwg, line 4540: 'is disallowed' -> 'not allowed'

language/move-prover/spec-lang/src/translate.rs
```diff
@@ -1331,6 +1392,46 @@ impl<'env, 'translator> ModuleTranslator<'env, 'translator> {
...
+            for (n, ty) in &type_params {
+                et.define_type_param(&loc, *n, ty.clone());
+            }
+            et.enter_scope();
+            for (n, ty) in &params {
+                et.define_local(&loc, *n, ty.clone(), None);
+            }
+            let translated = et.translate_seq(&loc, &seq, &result_type);
+            et.finalize_types();
+            // If no errors were generated, then the function is considered pure.
```
wrwg, line 1423: I'm wondering what happens if we do have a move function which is otherwise pure call a move function which is not. Do we catch this? This might be a bit more difficult to implement, so I'm fine we would punt on this and add a TODO.

emmazzz: Right we don't detect this right now. This is why I think we need to implement a more general algorithm later. Implementing it can be quite nontrivial.

language/move-prover/spec-lang/src/translate.rs
emmazzz: There will be two versions of pure functions generated, a boogie **function** for the spec version and a boogie **procedure** for the regular version. The function and the procedure need to have different names, so I add `$` to the function's name.

PR #5328, opened 2020-07-27
[validator-set] Add human_names and nonces

valerini:
In this PR:
* add `ValidatorOperatorConfig` resource that contains human_name of the operator;
* add the operator's name to the `set_validator_operator` transaction script and check that it matches the human name under operator's address;
* add human name and nonce to `add_validator` and `remove_validator` transaction scripts,
* rename the transaction scripts for better clarity: `set_validator_config` -> `register_validator_config` (to be called when the validator is not in the validator set, does not try to reconfigure the system), `add_validator` -> `add_validator_and_reconfigure`, `remove_validator` -> `remove_validator_and_reconfigure`.

## Motivation

Human name will prevent accidental use of an incorrect address in the transaction scripts and serve as a second level of protection.

Yes

## Test Plan

cargo xtest


language/stdlib/transaction_scripts/add_validator.move
```diff
@@ -1,12 +1,21 @@
...
     /// TODO(valerini): rename to add_validator_and_reconfigure?
-    fun add_validator(lr_account: &signer, validator_address: address) {
+    fun add_validator(
+        lr_account: &signer,
+        sliding_nonce: u64,
+        validator_name: vector<u8>,
+        validator_address: address
+    ) {
+        SlidingNonce::record_nonce_or_abort(lr_account, sliding_nonce);
+        assert(ValidatorConfig::get_human_name(validator_address) == validator_name, 0);
```
sblackshear, line 18: Wondering: what is this check protecting us from? Shouldn't the caller check offline that `validator_address` is indeed a validator account, has the expected name, is not already in the validator set, and so on? 

Similar question for remove/set_operator

language/stdlib/transaction_scripts/add_validator.move
```diff
@@ -1,12 +1,21 @@
 script {
     use 0x1::LibraSystem;
+    use 0x1::SlidingNonce;
+    use 0x1::ValidatorConfig;
 
     /// Add `new_validator` to the validator set.
     /// Fails if the `new_validator` address is already in the validator set
     /// or does not have a `ValidatorConfig` resource stored at the address.
     /// Emits a NewEpochEvent.
     /// TODO(valerini): rename to add_validator_and_reconfigure?
```
sblackshear, line 10: Might be a good opportunity to do this since we're making a breaking change to the script (same with scripts below).

language/stdlib/transaction_scripts/add_validator.move
valerini: Good question. This requirement comes from people building AOS, I think @dmitri-perelman will better explain this.
But as far as I understand, the approvers for LibraRoot operations (which are humans) will be approving a new validator by name (not by address) for convenience. We then don't want to trust the HSMs in AOS to re-translate approved validator's name into the address, we want to delegate this check to the chain, the HSM will then simply resign the action (= adding/removing the validator).

dmitri-perelman: Yes, @valerini is right, that was the idea.
We thought it'd be nice if the approvers would operate with human names if possible. In this case the flow would be like this:
* AOS sends the (name, address) pair the phone
* The app shows just the human name on the screen
* The signature generated on the phone covers the (name, address) pair, which is then used for generating the add_validator transaction by HSM.

In case AOS is compromised, the name might not match the address, hence the blockchain could verify this match.

sblackshear: That makes sense, thanks for the explanation! It's very cool that we can double-check this and get some protection even in the case where AOS is compromised.

config/management/genesis/src/genesis.rs
```diff
@@ -120,7 +120,7 @@ impl Genesis {
 
             let operator_account = self.fetch_operator_account(operator_name)?;
             let set_operator_script =
-                transaction_builder::encode_set_validator_operator_script(operator_account);
+                transaction_builder::encode_set_validator_operator_script(vec![], operator_account);
```
zekun000, line 123: why set it empty? there's a operator_name available above

PR #5330, opened 2020-07-27
[crypto] Simplify our Hkdf implementation

### LINKED ISSUE #4473: Simplify crypto::hkdf to the difference with `hmac`
There are two major updates to the original crate, ours is still generic but only for any hash function >=256 bits and does not allow empty messages + a few helper functions.

_Originally posted by @kchalkias in https://github.com/libra/libra/pull/4443_

Our hkdf module is very similar to the crate from rust-crypto. @kchalkias' answer above seems to mean this can be used through a facade over hkdf, with a constraint over the `ArrayLength` argument of the `Digest`, and limiting this module to the helper functions.

huitseeker:
Our implementation of HKDF is the same as the one in the hkdf crate modulo a few edge case decisions and signatures.
This reuses the HKDF crate while respecting our idiosyncracies, lightening our maintenance load.

Closes #4473

### DIFF
--- a/crypto/crypto/src/hkdf.rs
+++ b/crypto/crypto/src/hkdf.rs
@@ -76,12 +76,11 @@
 //! ```
 
 use digest::{
-    generic_array::{self, ArrayLength, GenericArray},
+    generic_array::{self, ArrayLength},
     BlockInput, FixedOutput, Reset, Update,
 };
 
-use generic_array::typenum::{IsGreaterOrEqual, True, Unsigned, U32};
-use hmac::{Hmac, Mac, NewMac};
+use generic_array::typenum::{IsGreaterOrEqual, True, U32};
 
 use std::marker::PhantomData;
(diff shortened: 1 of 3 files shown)


crypto/crypto/src/hkdf.rs
```diff
@@ -111,46 +110,25 @@ where
...
-            okm_block.copy_from_slice(&output[..okm_block.len()]);
-
-            prev = Some(output);
-        }
 
+        let hkdf =
+            hkdf::Hkdf::<D>::from_prk(prk).map_err(|_| HkdfError::WrongPseudorandomKeyError)?;
+        let mut okm = vec![0u8; length];
+        hkdf.expand(info.unwrap_or_else(|| &[]), &mut okm)
+            // length > D::OutputSize::to_usize() * 255
```
kchalkias, line 130: +1

PR #5335, opened 2020-07-27
[management] config

davidiw:
Add a config for management purposes. Lots of goodies here:
* new config file
* eliminate direct dependence on secure storage
* cleanup some code here and there
* unify everything to leverage the config

### DIFF
--- a/config/management/genesis/src/validator_operator.rs
+++ b/config/management/genesis/src/validator_operator.rs
@@ -3,65 +3,40 @@
 
 use libra_global_constants::OPERATOR_KEY;
 use libra_management::{
-    constants,
-    error::Error,
-    secure_backend::{SharedBackend, ValidatorBackend},
+    config::ConfigPath, constants, error::Error, secure_backend::SharedBackend,
 };
-use libra_secure_storage::{KVStorage, Value};
+use libra_secure_storage::Value;
 use structopt::StructOpt;
 
 #[derive(Debug, StructOpt)]
 pub struct ValidatorOperator {
+    #[structopt(flatten)]
+    config: ConfigPath,
     #[structopt(long)]
     operator_name: String,
-    #[structopt(flatten)]
-    validator_backend: ValidatorBackend,
     #[structopt(flatten)]
     shared_backend: SharedBackend,
 }
 
 impl ValidatorOperator {
     pub fn execute(self) -> Result<String, Error> {
-        // Get the operator name and verify it exists in the remote storage
-        let operator_name = self.get_and_verify_operator_name()?;
+        let config = self
+            .config
+            .load()?
+            .override_shared_backend(&self.shared_backend.shared_backend)?;
+        let operator_name = self.operator_name;
+
+        // Verify the operator exists in the shared storage
+        let operator_storage = config.shared_backend_with_namespace(operator_name.clone());
+        let _ = operator_storage.ed25519_key(OPERATOR_KEY)?;
 
         // Upload the operator name to shared storage
-        let storage_name = self.shared_backend.name();
-        let mut shared_storage = self
-            .shared_backend
-            .shared_backend
-            .create_storage(storage_name)?;
-
-        shared_storage
-            .set(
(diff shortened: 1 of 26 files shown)


config/management/src/validator_config.rs
```diff
@@ -38,10 +40,13 @@ impl ValidatorConfig {
...
     ) -> Result<Transaction, Error> {
-        let mut storage = StorageWrapper::new(
-            self.validator_backend.name(),
-            &self.validator_backend.validator_backend,
-        )?;
+        let config = self
+            .config
+            .load()?
+            .override_chain_id(self.chain_id)
+            .override_validator_backend(&self.validator_backend.validator_backend)?;
```
gregnazario, line 47: Should we do some smart checks and say if you have the ChainId with the config path to reject it?

config/management/src/storage.rs
```diff
@@ -13,37 +13,11 @@ use std::str::FromStr;
 
 /// A helper to handle common error handling and functionality for tooling
 pub struct StorageWrapper {
-    storage_name: &'static str,
-    storage: Storage,
+    pub(crate) storage_name: &'static str,
```
gregnazario, line 16: Interesting, that's a new one for me. It's like I guess a Java `protected`.  We should probably be doing that lots of places in our code.

config/management/genesis/src/verify.rs
```diff
@@ -147,45 +147,35 @@ fn compare_genesis(
...
     let expected_validator_key = validator_config.validator_network_identity_public_key;
     write_assert(
         buffer,
         VALIDATOR_NETWORK_KEY,
         actual_validator_key == expected_validator_key,
     );
 
-    let actual_fullnode_key = x25519_from_storage(FULLNODE_NETWORK_KEY, storage)
-        .map_err(|e| Error::StorageReadError(storage_name, FULLNODE_NETWORK_KEY, e))?;
+    let actual_fullnode_key = storage.x25519_public_from_private(FULLNODE_NETWORK_KEY)?;
```
gregnazario, line 178: Yay! less error handling, this was in my list of refactors I was going to get back to :)

config/management/genesis/src/validator_operator.rs
```diff
@@ -3,65 +3,40 @@
...
-                Error::StorageWriteError(storage_name, constants::VALIDATOR_OPERATOR, e.to_string())
-            })?;
+        // Upload the operator name to shared storage
+        let mut shared_storage = config.shared_backend();
+        shared_storage.set(
+            constants::VALIDATOR_OPERATOR,
+            Value::String(operator_name.clone()),
+        )?;
 
         Ok(operator_name)
```
gregnazario, line 40: We don't need to return the `operator_name` as it was passed by the caller

config/management/src/validator_config.rs
davidiw: I had thought about it, but I'm kinda okay with overrides. If folks wanna do weird stuff, it makes tool writing really a pain.

config/management/src/storage.rs
davidiw: ahh, well looking back now, I could have probably done this in such a way to use the new function.

config/management/genesis/src/validator_operator.rs
davidiw: I think the better behavior is to just return () and print success or something, but another PR :D

config/management/operational/src/governance.rs
```diff
@@ -14,76 +17,99 @@ use libra_types::{
 use structopt::StructOpt;
 
 #[derive(Debug, StructOpt)]
-pub struct AddValidator {
-    #[structopt(long, help = "JSON-RPC Endpoint (e.g. http://localhost:8080")]
-    host: String,
-    #[structopt(long)]
-    chain_id: ChainId,
+struct RootValidatorOperation {
```
davidiw, line 20: @zekun000, I unified your fields to match the rest of the code.

PR #5337, opened 2020-07-28
[Operational Management] Add command to rotate operator key.

JoshLind:
## Motivation

This PR adds the "rotate-operator-key" command to the operational management CLI tool. It also fixes a bug relating to the validator config commands (which currently derive the sequence number from the OWNER_ACCOUNT, when in fact that sequence number should come from the OPERATOR_ACCOUNT). 

This PR offers 4 commits:

1. First, we add the "rotate-operator-key" command to the operational management tool.
2. Next, we add a smoke test to rotate the operator key using the newly added command. For this test, we rotate the operator key, validate the transaction was executed correctly, and then attempt to rotate the consensus key (to ensure the operator key was indeed rotated).
3. Next, we fix the bug with the current set-validator command (which now allows the smoke test added in commit 2 above to work).
4. Finally, we offer a small refactor to the smoke tests.

Yes.

## Test Plan

All tests pass locally, including the newly added smoke test.

## Related PRs

None.

### DIFF
--- a/config/management/operational/src/operator_key.rs
+++ b/config/management/operational/src/operator_key.rs
@@ -1,0 +1,68 @@
+// Copyright (c) The Libra Core Contributors
+// SPDX-License-Identifier: Apache-2.0
+
+use crate::{json_rpc::JsonRpcClientWrapper, validator_config::RotateKey, TransactionContext};
+use libra_crypto::ed25519::Ed25519PublicKey;
+use libra_global_constants::{OPERATOR_ACCOUNT, OPERATOR_KEY};
+use libra_management::{constants, error::Error, storage::StorageWrapper};
+use libra_secure_time::{RealTimeService, TimeService};
+use libra_types::transaction::{authenticator::AuthenticationKey, RawTransaction, Transaction};
+use structopt::StructOpt;
+
+#[derive(Debug, StructOpt)]
+pub struct RotateOperatorKey {
+    #[structopt(flatten)]
+    rotate_key: RotateKey,
+}
+
+impl RotateOperatorKey {
+    pub fn execute(self) -> Result<(TransactionContext, Ed25519PublicKey), Error> {
+        // Fetch the operator account from storage
+        let mut storage = StorageWrapper::new(
+            &self.rotate_key.validator_config.validator_backend.name(),
+            &self
+                .rotate_key
+                .validator_config
+                .validator_backend
+                .validator_backend,
+        )?;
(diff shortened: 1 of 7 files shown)


config/management/operational/src/validator_config.rs
```diff
@@ -37,10 +37,11 @@ impl SetValidatorConfig {
             self.validator_config.validator_backend.name(),
             &self.validator_config.validator_backend.validator_backend,
         )?;
+        let operator_account = storage.account_address(OPERATOR_ACCOUNT)?;
         let owner_account = storage.account_address(OWNER_ACCOUNT)?;
 
         let client = JsonRpcClientWrapper::new(self.host);
-        let sequence_number = client.sequence_number(owner_account)?;
+        let sequence_number = client.sequence_number(operator_account)?;
```
gregnazario, line 44: Curious, what's the difference between these sequence numbers?

config/management/operational/src/operator_key.rs
```diff
@@ -0,0 +1,68 @@
...
+        let rotate_key_txn = RawTransaction::new_script(
+            operator_account,
+            sequence_number,
+            rotate_key_script,
+            constants::MAX_GAS_AMOUNT,
+            constants::GAS_UNIT_PRICE,
+            constants::GAS_CURRENCY_CODE.to_owned(),
+            RealTimeService::new().now() + constants::TXN_EXPIRATION_SECS,
+            self.rotate_key.validator_config.chain_id,
+        );
```
gregnazario, line 52: We should probably break this out into a function so we don't copy all the constants and everything between multiple places.  I know it's used elsewhere.

JoshLind: Yeah, I noticed it too when going through this. We should definitely take a look at our code and do some big refactors. If it wasn't for the latency requirements, I'd expand this PR to do a few of these things.. for now let's land and then follow up 😄

config/management/operational/src/validator_config.rs
JoshLind: From what I understand, sequence numbers are monotonically increasing numbers associated with accounts. Each transaction submitted and signed by an account must include this sequence number, to prevent against replay attacks. Given that the set_validator_config transaction is submitted by the operator account, we should read the sequence number of the operator_account (the owner account is a separate account 😄)

JoshLind:
Thanks, Greg! 😄 @bors-libra r=gregnazario

PR #5338, opened 2020-07-28
[consensus] add option in config to stop making progress

zekun000:
## Motivation

Add an option for consensus to only sync to last known committed state
and broadcast its own state but not vote on anything.

This helps sync every nodes to the same version under extreme situation
which needs operators to manually apply a transaction to fix the ledger
state.

Y

## Test Plan

Added smoke test.

## Related PRs

### DIFF
--- a/config/src/config/consensus_config.rs
+++ b/config/src/config/consensus_config.rs
@@ -15,6 +15,7 @@ pub struct ConsensusConfig {
     pub round_initial_timeout_ms: u64,
     pub proposer_type: ConsensusProposerType,
     pub safety_rules: SafetyRulesConfig,
+    pub stop_consensus: bool,
 }
 
 impl Default for ConsensusConfig {
@@ -29,6 +30,7 @@ impl Default for ConsensusConfig {
                 inactive_weights: 1,
             }),
             safety_rules: SafetyRulesConfig::default(),
+            stop_consensus: false,
         }
     }
 }
--- a/consensus/src/round_manager.rs
+++ b/consensus/src/round_manager.rs
@@ -141,7 +141,7 @@ impl RecoveryManager {
         self.sync_up(&sync_info, author).await
     }
 
-    async fn sync_up(&mut self, sync_info: &SyncInfo, peer: Author) -> Result<RecoveryData> {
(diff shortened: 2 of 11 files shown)


config/src/config/consensus_config.rs
```diff
@@ -15,6 +15,7 @@ pub struct ConsensusConfig {
     pub round_initial_timeout_ms: u64,
     pub proposer_type: ConsensusProposerType,
     pub safety_rules: SafetyRulesConfig,
+    pub stop_consensus: bool,
```
aching, line 18: I think we could probably have a detailed comment on when to use this option - given it is for extreme scenarios.  Essentially an operator wants to be available for syncing state but not commit any new transactions.  This is useful when validators coordinate on the latest version to apply a manual transaction or allow the network to synchronize the existing state without advancing it.

Also, stop_consensus is a bit ambiguous IMO.  Perhaps we could call it something like - sync_only or something?

consensus/src/round_manager.rs
```diff
@@ -245,9 +248,10 @@ impl RoundManager {
             .proposer_election
             .is_valid_proposer(self.proposal_generator.author(), new_round_event.round)
         {
-            let proposal_msg = self.generate_proposal(new_round_event).await?;
+            let proposal_msg =
```
aching, line 251: Should we even broadcast a proposal if consensus should not make progress?  It seems like it would be faster to halt progress by avoiding proposing as well.

config/src/config/consensus_config.rs
zekun000: `sync_only` sounds a better name

consensus/src/round_manager.rs
zekun000: doesn't matter much imo, we only broadcast once in a round

aching: I'm not concerned about the extra traffic, I'm just saying that if the goal is to stop consensus from progressing as fast as possible, getting rid of proposals in addition to votes will slow down the commit rate - rounds when the sync_only validator is leader will never get a QC.

PR #5339, opened 2020-07-28
[draft] finalize validator config contract

phlip9:
+ Finalize ValidatorConfig contract

before:

```
struct Config {
    consensus_pubkey: vector<u8>,
    validator_network_identity_pubkey: vector<u8>,
    validator_network_address: vector<u8>,
    full_node_network_identity_pubkey: vector<u8>,
    full_node_network_address: vector<u8>,
}
```

after:

```
struct Config {
    consensus_pubkey: vector<u8>,
    validator_network_addresses: vector<vector<u8>>,
    full_node_network_addresses: vector<vector<u8>>,
}
```

+ Support bulk serialize/encrypt and bulk deserialize/decrypt of network addresses with more complete error handling.

+ Convert management and operational tooling to support multiple network addresses.

+ Add HACKY support for `vector<vector<u8>>` transaction arguments with a specific `U8VectorVector` variant. I don't know if we should land this part... we need to decide whether we want general `vector<T>` tx arg support...

+ Note: this PR is a draft / blocked for now until we decide how to proceed with the transaction arguments. Another hacky workaround could be storing the whole list of addresses as a serialize `vector<u8>`, but that also feels pretty unclean.


language/stdlib/modules/ValidatorConfig.move
```diff
@@ -10,13 +10,8 @@ module ValidatorConfig {
...
     struct Config {
         consensus_pubkey: vector<u8>,
-        /// TODO(philiphayes): restructure
-        ///   3) remove validator_network_identity_pubkey
-        ///   4) remove full_node_network_identity_pubkey
-        validator_network_identity_pubkey: vector<u8>,
-        validator_network_address: vector<u8>,
-        full_node_network_identity_pubkey: vector<u8>,
-        full_node_network_address: vector<u8>,
+        validator_network_addresses: vector<vector<u8>>,
```
sblackshear, line 13: To your point from slack: if these are fixed length or have a known max length, you could do:

```
struct NetworkAddress { addr: vector<u8> }
public fun create(addr: vector<u8>): NetworkAddress { // add length check }
```

This is somewhat similar to newtypes in that it won't change the data layout of `Config` (i.e., you won't need to change your code in `libra/network`), although it does have the (very small) runtime overhead of creating a `NetworkAddress` struct, doing an extra borrow to get at `addr`, etc.

sblackshear, line 13: Comments for what these are?

language/move-core/types/src/transaction_argument.rs
```diff
@@ -13,6 +13,8 @@ pub enum TransactionArgument {
     Address(AccountAddress),
     U8Vector(#[serde(with = "serde_bytes")] Vec<u8>),
     Bool(bool),
+    // HACK
+    U8VectorVector(Vec<Vec<u8>>),
```
sblackshear, line 17: CC @vgao1996 / @runtian-zhou for this part

config/management/operational/src/validator_config.rs
```diff
@@ -38,34 +42,47 @@ impl SetValidatorConfig {
...
+        ))
+        .collect();
+
+        // Retrieve the current validator / fullnode addresses and update accordingly
+        let validator_addresses = if !validator_addresses.is_empty() {
+            validator_addresses
         } else {
-            decode_validator_address(
-                validator_config.validator_network_address,
+            decode_validator_addresses(
```
davidiw, line 63: it might make sense to give a handle to secure storage rather than passing in the encryption keys.

config/management/src/verify.rs
```diff
@@ -175,22 +186,61 @@ fn compare_genesis(
...
+    let raw_fn_addrs = validator_config.full_node_network_addresses;
+    let fn_addrs = deserialize_addresses(&validator_account, &raw_fn_addrs)
+        .collect::<Result<Vec<_>, _>>()
+        .map_err(|err| {
+            Error::UnexpectedError(format!("Unable to read fullnode network address: {}", err))
+        })?;
+    let fn_pubkeys: HashSet<_> = fn_addrs
+        .iter()
+        .filter_map(NetworkAddress::find_noise_proto)
+        .collect();
```
davidiw, line 224: there seems to be a lot of code and it feels redundant, I wonder if we need to think about how to make this a more effective library perhaps only within the management code base.

vgao1996:
Haven't had a detailed look but t definitely seems fine to me to add `U8VectorVector as the first step`, if that unblocks you. Also feel free to rename `U8Vector` to `VectorU8` and `U8VectorVector` to `VectorVectorU8` if that makes sense. I don't know why I flipped the order initially just for `TransactionArgument` and now it reads weird to me.

language/move-lang/src/typing/translate.rs
```diff
@@ -112,6 +112,7 @@ fn check_primitive_script_arg(context: &mut Context, mloc: Loc, idx: usize, ty:
         Type_::bool(loc),
         Type_::address(loc),
         Type_::vector(loc, Type_::u8(loc)),
+        Type_::vector(loc, Type_::vector(loc, Type_::u8(loc))),
```
dariorussi, line 115: not sure which version you are on but I thought @tnowacki already did all the work in the move copiler.
Todd?

language/move-core/types/src/transaction_argument.rs
dariorussi: @matbd what is the `serde` specification on the `U8Vector` and does this need one for `U8VectorVector`?

language/move-lang/src/typing/translate.rs
dariorussi, line 115: I thought @tnowacki already fixed this. Are you on latest master? Todd?

language/stdlib/modules/ValidatorConfig.move
dariorussi: alternatively there is a single entry point for setting that array so a check with abort there seems as good to me. Actually better than a side type, but that is just me...

language/move-lang/src/typing/translate.rs
tnowacki: Yeah, I already did this. This change should be removed

tnowacki: Fixed on master

phlip9: yep, @tnowacki fixed on master. i need to cleanup and rebase this PR 😅

language/move-core/types/src/transaction_argument.rs
matbd: Yes, LCS/Serde support arbitrarily nested vectors. We will just need to add the proper translation code for `U8VectorVector`.

matbd: @sblackshear @dariorussi To update the code generation, we need a final decision on this data structure. Special casing `U8Vector` (aka Bytes) is not entirely illogical but perhaps we'd rather avoid `U8VectorVector`. Also, because of `TypeTag`, we have recursive data structures in Libra types already.

language/transaction-builder-generator/src/rust.rs
```diff
@@ -138,11 +138,14 @@ fn quote_type(type_tag: &TypeTag) -> String {
         U64 => "u64".into(),
         U128 => "u128".into(),
         Address => "AccountAddress".into(),
-        Vector(type_tag) => match type_tag.as_ref() {
+        Vector(type_tag2) => match type_tag2.as_ref() {
             U8 => "Vec<u8>".into(),
+            Vector(type_tag3) => match type_tag3.as_ref() {
+                U8 => "Vec<Vec<u8>>".into(),
```
matbd, line 144: LGTM!

testsuite/cli/src/client_proxy.rs
```diff
@@ -673,43 +673,53 @@ impl ClientProxy {
...
+        let validator_network_addresses = [validator_network_address];
+        // let fullnode_identity_key = x25519::PublicKey::from_encoded_string(space_delim_strings[6])?;
         let fullnode_network_address = NetworkAddress::from_str(space_delim_strings[7])?;
-        let raw_fullnode_network_address = RawNetworkAddress::try_from(&fullnode_network_address)?;
+        let fullnode_network_addresses = [fullnode_network_address];
 
         let mut sender = Self::get_account_data_from_address(
             &mut self.client,
-            address,
+            account,
```
gregnazario, line 688: It's naming confusion like this, that makes me wish that we just called it `Account` instead of `AccountAddress`

testsuite/cli/src/client_proxy.rs
```diff
@@ -673,43 +673,53 @@ impl ClientProxy {
...
         let private_key = Ed25519PrivateKey::from_encoded_string(space_delim_strings[2])?;
         let consensus_public_key = Ed25519PublicKey::from_encoded_string(space_delim_strings[3])?;
-        let network_identity_key = x25519::PublicKey::from_encoded_string(space_delim_strings[4])?;
-        let network_address = NetworkAddress::from_str(space_delim_strings[5])?;
-        let raw_network_address = RawNetworkAddress::try_from(&network_address)?;
-        let fullnode_identity_key = x25519::PublicKey::from_encoded_string(space_delim_strings[6])?;
+        // let network_identity_key = x25519::PublicKey::from_encoded_string(space_delim_strings[4])?;
+        let validator_network_address = NetworkAddress::from_str(space_delim_strings[5])?;
+        let validator_network_addresses = [validator_network_address];
+        // let fullnode_identity_key = x25519::PublicKey::from_encoded_string(space_delim_strings[6])?;
```
gregnazario, line 682: Let's remove the commented lines :)

network/simple-onchain-discovery/src/lib.rs
```diff
@@ -61,106 +59,63 @@ pub fn gen_simple_discovery_reconfig_subscription(
...
-                        raw_enc_addr,
+                        &shared_val_netaddr_key_map,
+                        &config.validator_network_addresses,
                     )
+                    .filter_map(|maybe_addr| {
+                        maybe_addr
+                            .map_err(|err| {
+                                warn!("Failed to read validator network address: {}", err)
+                            })
+                            .ok()
```
gregnazario, line 89: Does this map the errors, and then just check if there wasn't an error?  I think you can remove the map_err

network/simple-onchain-discovery/src/lib.rs
```diff
@@ -61,106 +59,63 @@ pub fn gen_simple_discovery_reconfig_subscription(
...
+                RoleType::FullNode => {
+                    // deserialize full node network addresses.
+                    // log and skip those that fail for whatever reason.
+                    deserialize_addresses(&peer_id, &config.full_node_network_addresses)
+                        .filter_map(|maybe_addr| {
+                            maybe_addr
+                                .map_err(|err| {
+                                    warn!("Failed to read fullnode network address: {}", err)
+                                })
+                                .ok()
```
gregnazario, line 102: Same

config/seed-peer-generator/src/main.rs
```diff
@@ -66,11 +56,18 @@ fn get_validator_set(endpoint: String, peer_id: PeerId) -> anyhow::Result<Option
...
-        NetworkAddress::try_from(&validator_info.config().full_node_network_address)?;
-    Ok((peer_id, vec![network_addr]))
+    let raw_addrs = &validator_info.config().full_node_network_addresses;
+
+    // deserialize the raw network addresses. log and skip ones that fail to deserialize
+    let addrs: Vec<NetworkAddress> = deserialize_addresses(&peer_id, raw_addrs)
+        .filter_map(|maybe_addr| {
+            maybe_addr
+                .map_err(|err| warn!("Failed to deserialize fullnode network address: {}", err))
+                .ok()
```
gregnazario, line 68: Same about `map_err`, then `.ok()`

config/management/src/verify.rs
```diff
@@ -175,22 +186,61 @@ fn compare_genesis(
...
+
+    // deserialize full node network address
+
+    let raw_fn_addrs = validator_config.full_node_network_addresses;
+    let fn_addrs = deserialize_addresses(&validator_account, &raw_fn_addrs)
+        .collect::<Result<Vec<_>, _>>()
+        .map_err(|err| {
+            Error::UnexpectedError(format!("Unable to read fullnode network address: {}", err))
+        })?;
```
gregnazario, line 220: Maybe we could pass in an error conversion to deserialize_addresses?  Seems like we need to map the error on every deserialize_addresses function call.

I'll leave that up to you though, it might make it more complicated.

language/move-core/types/src/transaction_argument.rs
dariorussi: I'd be fine with us supporting constants in `TransactionArgument`s.
I agree with you that we could get away with a limited set of args if that was important.
I also agree with you that the recursion is fixable or we can work around it.
If we go with the more generic approach (all constants) we would not have to change `TransactionArgument`s for a long time (if not ever) and that has a big appeal.
If we want to limit `TransactionArgument`s then someone on the client side should make the call on what are the foreseeable arguments we want.

language/move-lang/functional-tests/tests/validator_set/rotate_and_reconfigure.move
```diff
@@ -45,7 +46,7 @@ script {
         // alice rotates bob's public key
         ValidatorConfig::set_config(account, {{bob}},
```
sblackshear, line 47: Would you adding some tests that call `set_config` with a nonempty vector containing invalid network addresses?

PR #5341, opened 2020-07-28
[ci] Compile the benchmarks

huitseeker:
Some of our benchmarks are not compiled regularly in CI and their code
is hence at risk of bit-rotting. This fixes it.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -417,6 +417,16 @@ jobs:
             RUST_BACKTRACE=1 $CARGO $CARGOFLAGS test \
               --features='vanilla' \
               --no-default-features
+  compile-benches:
+    executor: unittest-executor
+    description: Compile (but don't run) the benchmarks, to insulate against bit rot
+    steps:
+      - build_setup
+      - run:
+          name: Run bench --no-run
+          command: |
+            RUST_BACKTRACE=1 $CARGO $CARGOFLAGS x bench \
+              --no-run
   run-flaky-unit-test:
     executor: test-executor
     description: Run a list of known flaky tests.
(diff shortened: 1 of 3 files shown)


.circleci/config.yml
```diff
@@ -402,6 +402,17 @@ jobs:
             RUST_BACKTRACE=1 $CARGO $CARGOFLAGS test \
               --features='vanilla' \
               --no-default-features
+  compile-benches:
```
sausagee, line 405: @huitseeker this defines the job, but it will not run in CI.  To make it run, it needs to be part of the commit-verify workflow https://github.com/libra/libra/blob/96774332c59edf678a8813cca83996aba584417c/.circleci/config.yml#L634-L635

huitseeker: Thanks!

.circleci/config.yml
```diff
@@ -402,6 +402,17 @@ jobs:
...
+    executor: audit-executor
+    description: Compile (but don't run) the benchmarks, to insulate against bit rot
+    steps:
+      - build_setup
+      - restore_cargo_package_cache
+      - run:
+          name: Run bench --no-run
+          command: |
+            RUST_BACKTRACE=1 $CARGO $CARGOFLAGS x bench \
+              --no-run
```
sausagee, line 415: One more nit - we need `-- --no-run`

huitseeker: yep :facepalm:

huitseeker: Actually, we need pass-through arguments on x's `bench` command, it seems. :tired_face:

bmwill:
@huitseeker Hopefully https://github.com/libra/libra/pull/5420 can fix the compilation error. While testing I noticed though that this job you've added ends up OOMing because the builder size is too small. Once that PR lands you should rebase and try to either up the executor used or try to add a `--jobs` field to the `x bench` command to reduce memory pressure while linking.

bmwill:
@huitseeker Just migrated to a new implementation of bors so the commands are a little different now.
@bors-libra help

PR #5345, opened 2020-07-28
[fuzzing] use trait defaults to reduce repeated code

christinacelee:
## Motivation

Reduce some repeated code across fuzz target impl's by making the most out of trait default impls

### DIFF
--- a/testsuite/libra-fuzzer/src/lib.rs
+++ b/testsuite/libra-fuzzer/src/lib.rs
@@ -2,6 +2,7 @@
 // SPDX-License-Identifier: Apache-2.0
 
 use libra_proptest_helpers::ValueGenerator;
+use rand::RngCore;
 use std::{fmt, ops::Deref, str::FromStr};
 
 pub mod commands;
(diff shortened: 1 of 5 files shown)


testsuite/libra-fuzzer/src/lib.rs
```diff
@@ -12,7 +13,12 @@ pub mod fuzz_targets;
 /// Implementation for a particular target of a fuzz operation.
 pub trait FuzzTargetImpl: Sync + Send + fmt::Debug {
     /// The name of the fuzz target.
-    fn name(&self) -> &'static str;
+    fn name(&self) -> &'static str {
+        module_path!()
+            .rsplit("::")
+            .next()
+            .expect("module path must have at least one component")
+    }
```
bmwill, line 21: You may want to double check this but this may not be what you want. [`module_path`](https://doc.rust-lang.org/stable/std/macro.module_path.html) expands based on the context of where the macro physically is located so unless this default impl is overridden, I believe all impls will have module_path expanded to the root module.

bmwill: Here's an example in the [playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=ef151b3e3647417c81f675c6fc760a09)

mimoo:
a bit late to the party, but this is not a good idea for two reasons:

1. not all fuzzers are structured fuzzers
2. this is a placeholder that will have to be customized once this https://github.com/AltSysrq/proptest/pull/202 lands

PR #5347, opened 2020-07-28
[transaction builder generation] Add Script decoding capabilities in Rust

matbd:
## Motivation

The exiting `encode_XX_script` functions provide a convenient and type-safe way to build `Script` object from individual arguments. 
This PR provides the other direction in Rust: converting a `Script` value into a structured representation `ScriptCall`.

## Test Plan

unit tests + see the new `generated.rs`


language/transaction-builder-generator/Cargo.toml
```diff
@@ -10,10 +10,11 @@ edition = "2018"
 
 [dependencies]
 anyhow = "1.0.31"
+heck = "0.3.1"
```
sblackshear, line 13: Annoying question, but: is it possible for us to hand-implement the functionality we're using from this to avoid adding the dep?

language/transaction-builder-generator/src/common.rs
```diff
@@ -13,12 +17,58 @@ pub(crate) fn type_not_allowed(type_tag: &TypeTag) -> ! {
...
+    use TypeTag::*;
+    match type_tag {
+        Bool => Format::Bool,
+        U8 => Format::U8,
+        U64 => Format::U64,
+        U128 => Format::U128,
+        Address => Format::TypeName("AccountAddress".into()),
+        Vector(type_tag) => match type_tag.as_ref() {
+            U8 => Format::Bytes,
+            _ => type_not_allowed(type_tag),
```
sblackshear, line 33: We may have to extend this to handle the upcoming changes to `TransactionArgument` (but a TODO placeholder seems fine for now).

language/transaction-builder/src/generated.rs
```diff
@@ -5,94 +5,697 @@
...
+/// Structured representation of a call into a known Move script.
+/// ```ignore
+/// impl ScriptCall {
+///     pub fn encode(self) -> Script { .. }
+///     pub fn decode(&Script) -> Option<ScriptCall> { .. }
+/// }
+/// ```
+#[derive(Serialize, Deserialize, Clone, Debug, PartialEq, PartialOrd)]
+pub enum ScriptCall {
+    /// Add a `Currency` balance to `account`, which will enable `account` to send and receive
```
sblackshear, line 31: So awesome that the comment formatting is preserved!

language/transaction-builder-generator/src/common.rs
matbd: https://github.com/libra/libra/pull/5350 will make sure that we catch missing cases (at least in `rust.rs`) before landing a script that uses them.

language/transaction-builder/src/generated.rs
matbd: The auto-wrap that I added earlier was not a great idea.

language/transaction-builder-generator/Cargo.toml
matbd: @sblackshear @metajack @bmwill `abigen` and `serde-generate` also need this. A minimal version would have to support at least snake case, camel case, and shouty snake case for plain ASCII characters. I suppose it could be developed within the `serde-reflection` repository (and perhaps even provided as codegen helper functions in `serde-generate`).

PR #5350, opened 2020-07-28
[Move stdlib] add test to ensure that generated files are up to date

matbd:
## Motivation

Run the stdlib tool in CI and use the git state to make sure all generated files are up to date.

Using git is not ideal but at least we err on the safe side.

## Test Plan

CI

### DIFF
--- a/language/transaction-builder/generator/src/rust.rs
+++ b/language/transaction-builder/generator/src/rust.rs
@@ -52,7 +52,7 @@ fn output_preamble(out: &mut dyn Write, local_types: bool) -> Result<()> {
 
 // This file was generated. Do not modify!
 //
-// To re-generate this code, run: `(cd language/stdlib && cargo run --release)`
+// To re-generate this code, run: `cargo run --release -p stdlib`
 "#
         )?;
     }
--- a/language/stdlib/src/main.rs
+++ b/language/stdlib/src/main.rs
@@ -52,6 +52,14 @@ fn main() {
     let no_script_builder = matches.is_present("no-script-builder");
     let no_compiler = matches.is_present("no-compiler");
 
+    // Make sure that the current directory is `language/stdlib` from now on.
+    let exec_path = std::env::args().next().expect("path of the executable");
+    let base_path = std::path::Path::new(&exec_path)
+        .parent()
+        .unwrap()
+        .join("../../language/stdlib");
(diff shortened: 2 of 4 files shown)


bmwill:
What's the intent of this test? We already have a check in CI for this so idk how much more benefit this would give us

matbd:
> What's the intent of this test? We already have a check in CI for this so idk how much more benefit this would give us

@bmwill Cool, do you have a link? It must be somehow recent (or partial?) because in the past I noticed that running the stdlib tool would sometimes generate more changes than my PR. (Also, we were not necessary deleting obsolete generated files, see #4600)

matbd:
Also, https://github.com/libra/libra/pull/5319 just landed with a typo breaking the stdlib tool.

bmwill:
This script is run as a part of CI: https://github.com/libra/libra/blob/master/scripts/changed-files.sh

Though I now understand how this diff is different. This actually runs that binary and verifies theres no changed files.

matbd:
> This script is run as a part of CI: https://github.com/libra/libra/blob/master/scripts/changed-files.sh
> 
> Though I now understand how this diff is different. This actually runs that binary and verifies theres no changed files.

Oh I see. I'm totally going to steal the `--porcelain` trick.

There is definitely some overlapping but as you said, the new script runs the tool (and it is not guaranteed to be run at the end of CI).

jolestar:
How about using build.rs for auto compile stdlib at the build stage?

matbd:
> How about using build.rs for auto compile stdlib at the build stage?

We want to make it easy to use the generated objects, including from outside the repo.

language/transaction-builder/generator/src/rust.rs
```diff
@@ -52,7 +52,7 @@ fn output_preamble(out: &mut dyn Write, local_types: bool) -> Result<()> {
 
 // This file was generated. Do not modify!
 //
-// To re-generate this code, run: `(cd language/stdlib && cargo run --release)`
```
phlip9, line 55: does stdlib tool now work outside of language/stdlib? it breaks for me otherwise (though my branch is several days off master)

matbd: I thought I tried but let me try again (and change the cwd in the test command)

matbd: problem solved

PR #5356, opened 2020-07-29
[management] add key extraction

davidiw:
verified by generating a storage using the genesis initialization
and then running extract public and private

### DIFF
--- a/config/management/operational/src/command.rs
+++ b/config/management/operational/src/command.rs
@@ -11,8 +11,16 @@ use structopt::StructOpt;
 #[derive(Debug, StructOpt)]
 #[structopt(about = "Tool used for Operators")]
 pub enum Command {
+    #[structopt(about = "Remove a validator from ValidatorSet")]
+    AddValidator(crate::governance::AddValidator),
+    #[structopt(about = "Extract a private key from the validator storage")]
+    ExtractPrivateKey(crate::keys::ExtractPrivateKey),
+    #[structopt(about = "Extract a public key from the validator storage")]
+    ExtractPublicKey(crate::keys::ExtractPublicKey),
     #[structopt(about = "Set the waypoint in the validator storage")]
     InsertWaypoint(crate::waypoint::InsertWaypoint),
+    #[structopt(about = "Remove a validator from ValidatorSet")]
+    RemoveValidator(crate::governance::RemoveValidator),
     #[structopt(about = "Rotates the consensus key for a validator")]
     RotateConsensusKey(crate::validator_config::RotateConsensusKey),
     #[structopt(about = "Rotates a full node network key")]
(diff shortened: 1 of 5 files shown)


config/management/operational/src/command.rs
```diff
@@ -90,9 +100,11 @@ impl std::fmt::Display for CommandName {
 impl Command {
     pub fn execute(self) -> String {
         match self {
-            Command::InsertWaypoint(cmd) => {
-                format!("{:?}", cmd.execute().map(|()| "success").unwrap())
-            }
+            Command::AddValidator(cmd) => format!("{:?}", cmd.execute().unwrap()),
+            Command::InsertWaypoint(cmd) => cmd.execute().map(|()| "success".into()).unwrap(),
+            Command::ExtractPrivateKey(cmd) => cmd.execute().map(|()| "success".into()).unwrap(),
```
gregnazario, line 105: If we're extracting PrivateKeys, do we force HTTPS on the json_endpoint?

config/management/operational/src/command.rs
```diff
@@ -90,9 +100,11 @@ impl std::fmt::Display for CommandName {
...
+            Command::ExtractPublicKey(cmd) => cmd.execute().map(|()| "success".into()).unwrap(),
```
gregnazario, line 106: Nit: Maybe make a small function to handle the success conversion, so we don't have to copy "success" around.

config/management/operational/src/keys.rs
```diff
@@ -0,0 +1,70 @@
...
+impl ExtractPublicKey {
+    pub fn execute(self) -> Result<(), Error> {
+        let config = self
+            .extract_key
+            .config
+            .load()?
+            .override_validator_backend(&self.extract_key.validator_backend.validator_backend)?;
+
+        let storage = config.validator_backend();
+        let key_name = Box::leak(self.extract_key.key_name.into_boxed_str());
```
gregnazario, line 38: Curious why do we leak it?

config/management/operational/src/keys.rs
```diff
@@ -0,0 +1,70 @@
...
+        let key = storage.ed25519_private(key_name)?;
+        save_key(&key, key_name, &self.extract_key.key_file)
+    }
```
gregnazario, line 62: Nit: Probably "too" over-optimizing, but you could just make this "ExtractKey" and pass in a function pointer to retrieve the key.

config/management/operational/src/command.rs
davidiw: Great question, I hope no one uses this and it is blocked by the tokens that folks use. Maybe we could add an extra parameter --i-know-what-i-am-doing :)

config/management/operational/src/command.rs
davidiw: Hmm... I agree there's something awkward here, but we use these so we can actually get values for tests that leverage these operations too :/

config/management/operational/src/keys.rs
davidiw: hahahaha, rust hackery, I should have commented, but this effectively let's us make it a &'static str from a String... Since this is a CLI app, the String value should have no issues living long enough.

config/management/operational/src/keys.rs
davidiw: I thought about this too and said... I'll save myself the extra 10 minutes and give an opportunity for @gregnazario to shine ;)

PR #5362, opened 2020-07-29
[client] Remove unstable/obsolete API from libra-dev

thefallentree:
These apis are obsoleted and we should remove them.


xli:
could you separate delete with adding new api? it is easier to accept adding new one, but I don't think we should delete old one

thefallentree:
> could you separate delete with adding new api? it is easier to accept adding new one, but I don't think we should delete old one

the old ones are not in good shape either, and it is actually not used anymore, we should not keep them unless we have an good design

PR #5363, opened 2020-07-29
[logging] TCP structured logging DNS support

gregnazario:
The previous iteration was assuming that the input value was an IP
address and not a DNS name.  Now, it'll resolve the DNS name to a list
of addresses, and randomly choose one to connect to.

### Testing
Cluster test ran successful, using a DNS name of `localhost`

### DIFF
--- a/common/logger/src/struct_log.rs
+++ b/common/logger/src/struct_log.rs
@@ -17,7 +17,7 @@ use std::{
     io,
     io::Write as IoWrite,
     marker::PhantomData,
-    net::{SocketAddr, TcpStream},
+    net::{TcpStream, ToSocketAddrs},
     str::FromStr,
     sync::{
         atomic::{AtomicUsize, Ordering},
@@ -414,7 +414,7 @@ struct TCPStructLog {
 impl TCPStructLog {
     pub fn start_new(address: String) -> io::Result<Self> {
         let (sender, receiver) = mpsc::sync_channel(WRITE_CHANNEL_SIZE);
-        let sink_thread = TCPStructLogThread::new(receiver, &address);
+        let sink_thread = TCPStructLogThread::new(receiver, address);
         thread::spawn(move || sink_thread.run());
         Ok(Self { sender })
     }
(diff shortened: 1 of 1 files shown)


common/logger/src/struct_log.rs
```diff
@@ -476,8 +481,17 @@ impl TCPStructLogThread {
         }
     }
 
-    pub fn connect(&self) -> io::Result<TcpStream> {
-        TcpStream::connect_timeout(&self.address, Duration::from_millis(CONNECTION_TIMEOUT_MS))
+    pub fn connect(&mut self) -> io::Result<TcpStream> {
+        // resolve addresses to handle DNS names
+        let addrs: Vec<SocketAddr> = self.address.to_socket_addrs()?.collect();
+        if let Some(addr) = addrs.choose(&mut self.rand) {
+            TcpStream::connect_timeout(addr, Duration::from_millis(CONNECTION_TIMEOUT_MS))
```
davidiw, line 488: I would recommend just iterating through the list, in general, my experience has been that this list will contain an ipv6 and an ipv4 address and you're going to get a connection refused if the app only binds to one or the either.

common/logger/src/struct_log.rs
```diff
@@ -476,8 +475,20 @@ impl TCPStructLogThread {
...
+            if let Ok(stream) =
+                TcpStream::connect_timeout(&addr, Duration::from_millis(CONNECTION_TIMEOUT_MS))
+            {
+                return Ok(stream);
+            }
+        }
+
+        Err(io::Error::new(
+            io::ErrorKind::Other,
+            format!("Unable to resolve and connect to {}", self.address),
```
mgorven, line 490: Will this show the actual connection error?

davidiw: ahhh good, call, we should probably take in the result and check if it is okay, if it is not, we can return the last error

gregnazario: No, it doesn't show the actual connection error.

As something David mentioned before that one could be IPv6 and one could be Ipv4, and we could run through a bunch, I probably should merge the errors together

mgorven: Just getting the last error is fine, in k8s the address will be v4 only.

gregnazario: Fixed to retrun the last error.

PR #5364, opened 2020-07-29
[smoke test] fix flaky vfn failover test

christinacelee:
## Motivation

Fix flaky vfn_failover test
client proxy on a new node was panicking, so make sure that node is fully caught up on txns from another node before client proxy attempts any actions


christinacelee:
currently unfortunately for on-chain discovery, it's possible that a VFN discovers and connects to itself as a failover upstream peer... so if this happens and a timeout occurs, the VFN should penalize itself to reduce the chance of picking itself again. 
but seems like that didn't fix the test's flakiness.. I think the smoke test was timing out on submitting too many txns so I tried reducing the number, let's see if flakiness goes away with that

zekun000:
> a VFN discovers and connects to itself as a failover upstream peer

really?

christinacelee:
> really?
yes... networking team is supposed to fix this

PR #5365, opened 2020-07-29
[ci] Add hyperjump actions

metajack:
See https://github.com/libra/github-hyperjump for details. These allow
proxying benign capabilities triggered from PRs originating in forked repos to
work around GitHub Actions limitations.

They are currently unused, but I am working on a dependency summary action that makes use of this for commenting and labeling PRs.


bmwill:
Could you point me to the relevant parts of this PR? Its a bit difficult to see given its >55k lines of code 😄

metajack:
`.github/actions/*/README.md` for an overview. The real code is in `.github/actions/*/src/index.js` and is quite small. Ignore everything in `.github/actions/*/dist`. The action configs are in the `.yml` files.

metajack:
Note that we bundle the `dist/*` files so that the actions can run directly from the repo, but this requires that they be bundled with all their npm dependencies into a single JS file. See https://docs.github.com/en/actions/creating-actions/creating-a-javascript-action

.github/workflows/hyperjump-comment.yml
```diff
@@ -0,0 +1,21 @@
...
+
+jobs:
+  comment:
+    runs-on: ubuntu-latest
+    name: (hyperjump) comment
+    steps:
+      - name: checkout
+        uses: actions/checkout@v2
+      - name: comment
+        uses: ./.github/actions/hyperjump-comment
```
sausagee, line 15: sweeeet! It looks like we can do the same in LBT route its comment through `hyperjump`

.github/actions/comment/.gitignore
```diff
@@ -0,0 +1 @@
+node_modules
```
sausagee, line 1: nit - need new line

.github/actions/labels/src/index.js
```diff
@@ -0,0 +1,44 @@
+const core = require("@actions/core");
+const github = require("@actions/github");
+const got = require("got");
+const process = require("process");
+
+const hyperjump_url = "http://github.aws.hlw3truzy4ls.com:6080/hyperjump/jump";
```
bmwill, line 6: My only question is: does this technically allow anyone to use these actions and piggyback on top of the service running at "http://github.aws.hlw3truzy4ls.com:6080/hyperjump/jump" for their own projects?

metajack: Anyone can hit the hyperjump/jump url and request a jump to an arbitrary repo. The repository dispatch api request will fail unless the GitHub app is installed and.authorized for that repo. So repos have to opt-in to hyperjumps. However, any opted-in repo can receive hyperjumps from anywhere, so we limit the actions a hyperjump can do to actions any public user could normally perform, or that we don't mind them performing (like commenting and labeling PRs).

devtools/x/src/lint/whitespace.rs
```diff
@@ -80,9 +81,22 @@ impl ContentLinter for TrailingWhitespace {
...
+    };
+
+
+    // extension based opt outs
+    match file.extension() {
+        Some("exp") => return RunStatus::Skipped(SkipReason::UnsupportedExtension(file.extension())),
+        _ => (),
     }
+
+    RunStatus::Executed
```
bmwill, line 101: Looks good! At some point (not in this PR of course) we should move these so they're configurable in the x.toml file but this is sufficient for now.

PR #5375, opened 2020-07-30
[libra framework] Derive `Arbitrary` on `ScriptCall`. Use to generate a transaction fuzzer

tzakian:
With the recent changes to the transaction builder generator, we can now derive `Arbitrary` on the generated `ScriptCall` enum. This then means that we get a free way of fuzzing with type-correct transactions for all allowlisted transactions. 

Right now the fuzzing doesn't do anything too complex, it just makes sure that a type correct transaction:
1. Isn't marked as discared (i.e., only Abort, ExecutionFailure, and VerificationErrors).
   - A TODO here is to get rid of the `ExecutionFailure` class. But that's for a future PR.
2. Doesn't panic the VM. 

The way we add the proptest derives in the generated rust code is a bit...hacky, but I don't think there's any easy way of changing it without updating serde-reflection and the way the transaction builder generator calls in to that. Would love any suggestions on how to possibly clean this up.

### DIFF
--- a/language/e2e-tests/src/tests/transaction_fuzzer.rs
+++ b/language/e2e-tests/src/tests/transaction_fuzzer.rs
@@ -1,0 +1,86 @@
+// Copyright (c) The Libra Core Contributors
+// SPDX-License-Identifier: Apache-2.0
+
+use crate::{
+    account::{self, Account},
+    executor::FakeExecutor,
+    keygen::KeyGen,
+};
+use libra_types::account_config;
+use proptest::{collection::vec, prelude::*};
+use transaction_builder::encode_create_parent_vasp_account_script;
+use transaction_builder_generated::stdlib::ScriptCall;
+
+proptest! {
+    #![proptest_config(ProptestConfig::with_cases(16))]
+    #[test]
+    fn fuzz_scripts_genesis_state(
+        txns in vec(any::<ScriptCall>(), 0..100),
+    ) {
+        let executor = FakeExecutor::from_genesis_file();
+        let mut accounts = vec![];
+        accounts.push((Account::new_libra_root(), 1));
+        accounts.push((Account::new_blessed_tc(), 0));
+        let num_accounts = accounts.len();
+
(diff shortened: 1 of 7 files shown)


language/stdlib/src/lib.rs
```diff
@@ -165,14 +165,34 @@ fn build_abi(output_path: &str, sources: &[String], dep_path: &str, compiled_scr
     move_prover::run_move_prover_errors_to_stderr(options).unwrap();
 }
 
+fn post_process_add_proptest_derivations(buffer: &[u8]) {
```
matbd, line 168: Can you add a TODO pointing to https://github.com/facebookincubator/serde-reflection/issues/32 ?

language/e2e-tests/src/tests/transaction_fuzzer.rs
```diff
@@ -0,0 +1,31 @@
...
+        let executor = FakeExecutor::from_genesis_file();
+        let mut accounts = vec![];
+        accounts.push((Account::new_libra_root(), 1));
+        accounts.push((Account::new_blessed_tc(), 0));
+        let num_accounts = accounts.len();
+
+        for (i, txn) in txns.into_iter().enumerate() {
+            let script = txn.encode();
+            let (account, account_sequence_number) = &accounts[i % num_accounts];
+            let output = executor.execute_transaction(
```
sblackshear, line 23: Should we apply the `output.write_set` (or perhaps do this in a different proptest)? I think as it stands, the test is executing a bunch of random txes against the genesis state, which is very useful. But we will see even more behaviors if we can execute txes against non-genesis states as well.

tzakian: I think that's a really good idea! I'll add it as another proptest :)

language/stdlib/src/lib.rs
tzakian: Will do!

tzakian:
Nope! I think it's good to go. I'll be landing shortly.

PR #5377, opened 2020-07-30
[metrics] Add health check endpoint, setup metrics server earlier

mgorven:
* Start the metrics server earlier so we can collect metrics even if
  startup takes a long time.
* Add `/-/healthy` endpoint to use for healthchecks. For now it always
  returns 200.


libra-node/src/main_node.rs
```diff
@@ -59,6 +59,15 @@ fn setup_debug_interface(config: &NodeConfig) -> NodeDebugService {
 }
 
 pub fn setup_environment(node_config: &NodeConfig) -> LibraHandle {
+    let metrics_port = node_config.debug_interface.metrics_server_port;
```
davidiw, line 62: pretty ugly blob... should really be in a function or something...

PR #5378, opened 2020-07-30
[management] pretty print op tool output

gregnazario:
The libra-operational-tool output was using Debug output from rust which
was non-standard and very verbose.  It wasn't very user friendly, so now
we output with JSON, and we decrypt the validator addresses for
readability.


config/management/operational/src/command.rs
```diff
@@ -100,19 +103,44 @@ impl std::fmt::Display for CommandName {
...
+            Ok(_) => "success".into(),
+            Err(err) => err.to_string(),
+        }
+    }
+
+    /// For pretty printing outputs in JSON
+    fn pretty_print<T: Serialize>(result: Result<T, Error>) -> String {
+        match result {
+            Ok(value) => serde_json::to_string_pretty(&value).unwrap(),
+            Err(err) => err.to_string(),
```
davidiw, line 143: cc: @JoshLind, I really feel like this PR was meant for you....

gregnazario:
> I know you shared this internally, but it would be good to share outputs of these type of things in the PR / commit.

Done, I added an output.

I think this will be an iterative process, because the cluster at some point yesterday started rejecting my requests, and I'll try troubleshooting again today to make sure everything is okay

PR #5383, opened 2020-07-30
[cli] Implement a command to increment LibraVersion

runtian-zhou:
## Motivation

Add a command to simulate the version increment transaction.

Yes

## Test Plan

I've already tested the cli in #4416. The plan is that we will separate that PR into two and commit the upgrade change in the next release in order to build and test it.

### DIFF
--- a/testsuite/cli/src/dev_commands.rs
+++ b/testsuite/cli/src/dev_commands.rs
@@ -29,6 +29,7 @@ impl Command for DevCommand {
             Box::new(DevCommandRemoveValidator {}),
             Box::new(DevCommandGenWaypoint {}),
             Box::new(DevCommandRegisterValidator {}),
+            Box::new(DevCommandChangeLibraVersion {}),
         ];
         subcommand_execute(&params[0], commands, client, &params[1..]);
     }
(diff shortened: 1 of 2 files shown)


testsuite/cli/src/dev_commands.rs
```diff
@@ -175,6 +176,33 @@ impl Command for DevCommandDisableCustomScript {
...
 
+pub struct DevCommandChangeLibraVersion {}
+
+impl Command for DevCommandChangeLibraVersion {
+    fn get_aliases(&self) -> Vec<&'static str> {
+        vec!["change_libra_version"]
+    }
+
+    fn get_params_help(&self) -> &'static str {
+        ""
```
xli, line 187: may put a params message here, by reading the diff, I'm not even sure what is 2 params, one is version seems clear, what is next one? I understand it is inside change_libra_version func, but, some help message will help a lot for user.

sherry-x:
Oh no, did we not landed this last week? This won't be included in the new deployment for Pre-mainnet.
You will have to build this tool use this commit and make sure it's compatible with Premainnet

PR #5385, opened 2020-07-30
[language][vm] rework GlobalValue to fix the deletion bug

vgao1996:
## Summary
Currently if you publish then immediately unpublish a resource in the same transaction, the Move VM issues an extraneous deletion op, trying to get this resource removed from storage, where it doesn't exist. This PR gets it fixed by implementing a new model of `GlobalValue` that correctly tracks the states. As a bonus, it also removes a bunch of boilerplate code and makes certain operations slightly more efficient. 

## Test Plan
cargo test, new tests will be added


language/move-vm/types/src/values/values_impl.rs
```diff
@@ -191,11 +191,21 @@ pub struct Struct {
...
+    },
+    Cached {
+        fields: Rc<RefCell<Vec<ValueImpl>>>,
+        status: Rc<RefCell<GlobalDataStatus>>,
+    },
+    Deleted,
 }
 
+#[derive(Debug)]
+pub struct GlobalValue(GlobalValueImpl);
```
vgao1996, line 207: The `GlobalValue` is no longer an owned thing being passed between the VM and the data cache. It's more like a storage slot right now. Shall we change its name to something more descriptive? I thought about "GlobalSlot" but that sounds weird and unclear by its own...

language/move-vm/types/src/values/values_impl.rs
```diff
@@ -191,11 +191,21 @@ pub struct Struct {
...
+enum GlobalValueImpl {
+    None,
+    Fresh {
+        fields: Rc<RefCell<Vec<ValueImpl>>>,
+    },
+    Cached {
+        fields: Rc<RefCell<Vec<ValueImpl>>>,
+        status: Rc<RefCell<GlobalDataStatus>>,
+    },
+    Deleted,
```
vgao1996, line 203: Reviewers: please pay attention to the states encoded by this enum and GlobalDataStatus, and the transition rules via the operations below. These are the core and tricky part of this PR.

language/move-vm/types/src/values/value_tests.rs
```diff
@@ -110,84 +110,13 @@ fn struct_borrow_nested() -> PartialVMResult<()> {
...
-    }
-
-    assert!(gv.is_dirty()?);
-
-    {
-        let r1: StructRef = gv.borrow_global()?.value_as()?;
-        let r2: StructRef = r1.borrow_field(0)?.value_as()?;
-        let r3: Reference = r2.borrow_field(0)?.value_as()?;
-        assert!(r3.read_ref()?.equals(&Value::u64(0))?);
-    }
```
vgao1996, line 190: These already got covered by an IR test I added last time so I'm deleting them.

vgao1996:
@dariorussi @runtian-zhou Just added the `crit!` and comments. The PR is now 100% ready.

andll:
Does this PR change genesis?

vgao1996:
Simple answer: no. But let me double check.

andll:
So cluster test failed compat check - is it possible that this PR changed something else in VM behaviour? For example same txn producing different result with old and new code?

vgao1996:
Yes, the PR gets rid of some extraneous deletion ops.

vgao1996:
Can confirm this does not change the genesis, which was consistent with our previous understanding -- the genesis doesn't contain deletion ops.

andll:
So if VM logic changes, and VM might produce different hash, then it is also a breaking change.

So if we believe this is the case, then we need to follow breaking change protocol here - file a sev, add `[breaking]` prefix to commit message and make a wp post

andll:
So basically what I am trying to understand is this:

- Cluster test thinks the change is breaking.

Do we agree? Yes -> follow breaking change procedure. No => We can retry with land again and see if ct will accept this time. If so, then its bug in ct

vgao1996:
> So if VM logic changes, and VM might produce different hash, then it is also a breaking change.
So just to double check, the logic here is that while both the new and old writesets could be handled by the storage backend, varying VM behaviors could cause validators to disagree in consensus and therefore making this a breaking change, right?

> So if we believe this is the case, then we need to follow breaking change protocol here - file a sev, add [breaking] prefix to commit message and make a wp post
Will do.

andll:
> varying VM behaviors could cause validators to disagree in consensus and therefore making this a breaking change

If this change could force VM to produce different results before and after, then yes, this is a breaking change

vgao1996:
Cool I'll follow the breaking change protocol.

vgao1996:
Thanks for pointing this out!

runtian-zhou:
Comments for breaking change: The testsuite will deploy a network with all nodes running on the old validator software (i.e: master), and gradually replace them with the newer software (master with this commit) and see if consensus can be remained. It seems to me that the code change here doesn't seem to be a breaking change to me as the deletion op will be a nil operation from the executor standpoint, so it should not cause the network to diverge. The fact that it passes the scenario where a half nodes is running on the old version and a half node running on a new version indicates that as well. Thus I would more inclined to this is due to some transient error in the cluster test and let's see if reruning the test gives us a different answer.

andll:
/land

^ Running this command to create a new image rebased of latest master with https://github.com/libra/libra/pull/5473

language/move-vm/types/src/values/values_impl.rs
```diff
@@ -2178,57 +2184,136 @@ impl Struct {
...
-    pub fn new(v: Value) -> PartialVMResult<Self> {
-        match v.0 {
-            ValueImpl::Container(Container::StructR(r)) => {
-                // TODO: check strong count?
-                Ok(Self {
-                    status: Rc::new(RefCell::new(GlobalDataStatus::Clean)),
-                    fields: r,
-                })
+impl GlobalValueImpl {
+    fn cached(val: ValueImpl, status: GlobalDataStatus) -> PartialVMResult<Self> {
```
sblackshear, line 2188: The only caller of this passes in `status: GlobalDataStatus::Clean`. Maybe we should simplify this API accordingly (+ to match the old code)?

language/move-vm/types/src/values/values_impl.rs
```diff
@@ -178,24 +178,30 @@ pub struct Struct {
...
+enum GlobalValueImpl {
+    /// No resource resides in this slot or in storage.
+    None,
+    /// A resource has been published to this slot and it did not previously exist in storage.
+    Fresh { fields: Rc<RefCell<Vec<ValueImpl>>> },
+    /// A resource resides in this slot and also in storage. The status flag indicates whether
+    /// it has potentially been altered.
+    Cached {
+        fields: Rc<RefCell<Vec<ValueImpl>>>,
+        status: Rc<RefCell<GlobalDataStatus>>,
```
sblackshear, line 194: Is this ever mutated after a `GlobalValueImpl` is created? As far as I can tell, it is set to `Clean` upon creation and never changed.

language/move-vm/runtime/src/data_operations.rs
```diff
@@ -1,115 +0,0 @@
...
-        Ok(None) => true,
-        Ok(Some(_)) => false,
-        Err(e) => match e.major_status() {
-            StatusCode::MISSING_DATA => true,
-            _ => return Err(e),
-        },
-    };
-    if can_write {
-        let new_root = GlobalValue::new(Value::struct_(resource))?;
-        new_root.mark_dirty()?;
```
sblackshear, line 43: This code is deleted, but I can't find the equivalent piece of new code that updates the `status`

vgao1996:
@sblackshear Search for "mark_dirty" in `values_impl.rs`. We have also have a bunch of tests ensuring these mutations result in the desired effects.

language/move-vm/runtime/src/data_operations.rs
vgao1996: It doesn't because when you move a resource, the state changes either from `Cached` to `Deleted` or from Fresh to None. If you re-publish the resource, it'll change from `Deleted` to `Cached { Dirty }` or from `None` to `Fresh`

language/move-vm/types/src/values/values_impl.rs
```diff
@@ -2178,57 +2184,136 @@ impl Struct {
...
+
+    fn move_to(&mut self, val: ValueImpl) -> PartialVMResult<()> {
+        match self {
+            Self::Fresh { .. } | Self::Cached { .. } => {
+                return Err(PartialVMError::new(
+                    StatusCode::CANNOT_WRITE_EXISTING_RESOURCE,
+                ))
+            }
+            Self::None => *self = Self::fresh(val)?,
+            Self::Deleted => *self = Self::cached(val, GlobalDataStatus::Dirty)?,
```
sblackshear, line 2242: @vgao1996: I guess my question is: if you replace this with `Self::fresh { val }` and delete the `status` field altogether, what would go wrong?

language/move-vm/types/src/values/values_impl.rs
vgao1996: `move_to` calls this with `Dirty` on line 2242.

language/move-vm/types/src/values/values_impl.rs
vgao1996: Yes, see my other comments.

language/move-vm/types/src/values/values_impl.rs
sblackshear: Ok, I think I see the answer :). Sorry, just ramping up and trying to understand this...

vgao1996: `Deleted` will never become `Fresh`. Only the following conversions are allowed through move_to/move_from: `Deleted` <=> `Cached`,  `None <=> Fresh` and mutability through `borrow_global_mut` only allow `Cached { Clean } => Cached { Dirty }`.

vgao1996:
@sblackshear just to double check, does this PR look correct to you?

sblackshear:
> @sblackshear just to double check, does this PR look correct to you?

Yep, LGTM now that I understand how the cache status works!

PR #5386, opened 2020-07-30
[gha] check breaking tag for all commits in PR

rustielin:
## Motivation

Land-blocking workflow fails to check `[breaking]` tag in all commits in a PR. This makes git log check all commit messages between master HEAD and current branch HEAD.

Yes

## Test Plan

auto branch is rebased on master, so git revs will always calculate the right range.

## Related PRs

### DIFF
--- a/.github/workflows/land-blocking.yml
+++ b/.github/workflows/land-blocking.yml
@@ -18,6 +18,7 @@ jobs:
       - name: Setup env
         run: |
           echo "::set-env name=LIBRA_GIT_REV::$(git rev-parse --short=8 HEAD)"
+          echo "::set-env name=MASTER_GIT_REV::$(git rev-parse --short=8 origin/master)"
           echo "::set-env name=TEST_TAG::land_$(git rev-parse --short=8 HEAD)"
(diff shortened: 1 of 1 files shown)


.github/workflows/land-blocking.yml
```diff
@@ -54,7 +55,7 @@ jobs:
         if: steps.check_ks.outputs.should_run == 'true'
         run: |
           set +e
-          commit_message=$(git log -1 --pretty=oneline)
+          commit_message=$(git log --pretty=oneline $MASTER_GIT_REV..$LIBRA_GIT_REV)
```
andll, line 58: I don't think you _need_ to rev-parse, you can probably do smth like `origin/master..$LIBRA_GIT_REV`, but up to you

rustielin: Leaving like this for now, but thanks for pointing that out. I think one thing in the future is we might want to check against `origin/testnet` for client-compatibility. e.g. because pylibra and other clients will only be known compatible with latest testnet version.

PR #5390, opened 2020-07-30
[json-rpc] limit batch request size to 20 by default

xli:
## Motivation

Limit batch size to avoid client sends too many requests and cause server OOM or other problems.

## Test Plan

Unit tests

### DIFF
--- a/client/json-rpc/src/client.rs
+++ b/client/json-rpc/src/client.rs
@@ -124,6 +124,7 @@ pub enum JsonRpcAsyncClientError {
     ClientError(reqwest::Error),
     InvalidArgument(String),
     InvalidServerResponse(String),
+    JsonRpcError(JsonRpcError),
 }
 
 impl JsonRpcAsyncClientError {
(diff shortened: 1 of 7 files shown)


client/json-rpc/src/client.rs
```diff
@@ -261,12 +262,29 @@ impl JsonRpcAsyncClient {
...
         }
 
-        let responses: Vec<Value> = resp
+        let json: Value = resp
             .json()
             .await
             .map_err(|e| JsonRpcAsyncClientError::InvalidServerResponse(e.to_string()))?;
-        process_batch_response(batch, responses)
-            .map_err(|e| JsonRpcAsyncClientError::InvalidServerResponse(e.to_string()))
+        match json {
```
thefallentree, line 269: I think you can write it like this

```
if let Value::Array(response) = json {
  process_batch_response(batch, responses)
                .map_err(|e| JsonRpcAsyncClientError::InvalidServerResponse(e.to_string()))
} else if let Value::Object(response) = json {
   ...
} 
  Err(JsonRpcAsyncClientError::InvalidServerResponse(
                    "expect batch response array instead of object".to_string(),
                ))

```

PR #5391, opened 2020-07-30
[gha] be more explicit with breaking tag check

rustielin:
## Motivation

Fix of a fix (https://github.com/libra/libra/pull/5386)...
Exit code was getting overwritten when we import the killswitch in conditional, so capture exit code of grep explicitly. (When killswitch is `false`, that still makes bash evaluate it to 1 since it's a keyword)

(Write your answer here.)

## Test Plan

## Related PRs


rustielin:
> that's how bash magic scares me away

Hey it's scary for me too, but it's a power I must yield 😤
(I hope this works)

andll:
I think a bit scary part here is that we don't seem to have an easy way to canary it

We probably need to refactor this, so that land blocking hook just calls some shell script file, and rework it in a way where we can try it locally

Otherwise, without testing, we are bound to have those problems

PR #5395, opened 2020-07-30
[vm] Bolstered prologue/epilogue VM status special case

tnowacki:
- Added debug assert of unexpected errors
- Any unexpected error is now discarded as it should be treated as an invariant violation

- We might want to add logging here? unsure

## Motivation

- Prevent charging when there is an unexpected error in the prologue/epilogue that would otherwise be charged (verification/serialization/etc)

## Test Plan

- Not really possible to test the failure case as it should never happen...
- cargo test

### DIFF
--- a/language/libra-vm/src/errors.rs
+++ b/language/libra-vm/src/errors.rs
@@ -1,0 +1,147 @@
+// Copyright (c) The Libra Core Contributors
+// SPDX-License-Identifier: Apache-2.0
+
+use libra_logger::prelude::*;
+use move_core_types::vm_status::{known_locations, StatusCode, VMStatus};
+
+/// Error codes that can be emitted by the prologue. These have special significance to the VM when
+/// they are raised during the prologue.
+/// These errors are only expected from the `known_locations::account_module_abort()` module
+/// The prologue should not emit any other error codes or fail for any reason, doing so will result
+/// in the VM throwing an invariant violation
+pub const EACCOUNT_FROZEN: u64 = 0; // sending account is frozen
+pub const EBAD_ACCOUNT_AUTHENTICATION_KEY: u64 = 1; // auth key in transaction is invalid
+pub const ESEQUENCE_NUMBER_TOO_OLD: u64 = 2; // transaction sequence number is too old
+pub const ESEQUENCE_NUMBER_TOO_NEW: u64 = 3; // transaction sequence number is too new
+pub const EACCOUNT_DOES_NOT_EXIST: u64 = 4; // transaction sender's account does not exist
+pub const EINSUFFICIENT_BALANCE: u64 = 5; // insufficient balance (to pay for gas deposit)
+pub const ETRANSACTION_EXPIRED: u64 = 6; // transaction expiration time exceeds block time.
+pub const EBAD_CHAIN_ID: u64 = 7; // chain_id in transaction doesn't match the one on-chain
+
+// invalid sender (not libra root) for write set
+pub const EINVALID_WRITESET_SENDER: u64 = 33;
+// writeset prologue sequence nubmer is too new
+pub const EWS_PROLOGUE_SEQUENCE_NUMBER_TOO_NEW: u64 = 11;
+
+/// Converts particular Move abort codes to specific validation error codes for the prologue
+/// Any non-abort non-execution code is considered an invariant violation, specifically
+/// `UNEXPECTED_ERROR_FROM_KNOWN_MOVE_FUNCTION`
+pub fn convert_normal_prologue_error(status: VMStatus) -> VMStatus {
+    match status {
+        VMStatus::Executed => VMStatus::Executed,
+        VMStatus::MoveAbort(location, code) => {
+            if location != known_locations::account_module_abort() {
+                crit!(
+                    "[libra_vm] Unexpected prologue move abort: {:?}::{:?}",
+                    location,
+                    code
+                );
+                return VMStatus::Error(StatusCode::UNEXPECTED_ERROR_FROM_KNOWN_MOVE_FUNCTION);
+            }
+
+            let new_major_status = match code {
+                EACCOUNT_FROZEN => StatusCode::SENDING_ACCOUNT_FROZEN,
+                // Invalid authentication key
+                EBAD_ACCOUNT_AUTHENTICATION_KEY => StatusCode::INVALID_AUTH_KEY,
+                // Sequence number too old
+                ESEQUENCE_NUMBER_TOO_OLD => StatusCode::SEQUENCE_NUMBER_TOO_OLD,
+                // Sequence number too new
+                ESEQUENCE_NUMBER_TOO_NEW => StatusCode::SEQUENCE_NUMBER_TOO_NEW,
+                // Sequence number too new
+                EACCOUNT_DOES_NOT_EXIST => StatusCode::SENDING_ACCOUNT_DOES_NOT_EXIST,
+                // Can't pay for transaction gas deposit/fee
+                EINSUFFICIENT_BALANCE => StatusCode::INSUFFICIENT_BALANCE_FOR_TRANSACTION_FEE,
+                ETRANSACTION_EXPIRED => StatusCode::TRANSACTION_EXPIRED,
+                EBAD_CHAIN_ID => StatusCode::BAD_CHAIN_ID,
+                code => {
+                    crit!(
+                        "[libra_vm] Unexpected prologue move abort: {:?}::{:?}",
+                        location,
+                        code
+                    );
+                    return VMStatus::Error(StatusCode::UNEXPECTED_ERROR_FROM_KNOWN_MOVE_FUNCTION);
+                }
+            };
+            VMStatus::Error(new_major_status)
+        }
+        status @ VMStatus::ExecutionFailure { .. } | status @ VMStatus::Error(_) => {
+            crit!("[libra_vm] Unexpected prologue error: {:?}", status);
+            VMStatus::Error(StatusCode::UNEXPECTED_ERROR_FROM_KNOWN_MOVE_FUNCTION)
+        }
+    }
+}
+
+/// Checks for only move aborts or successful execution.
+/// Any other errors are mapped to the invariant violation
+/// `UNEXPECTED_ERROR_FROM_KNOWN_MOVE_FUNCTION`
+pub fn convert_normal_success_epilogue_error(status: VMStatus) -> VMStatus {
+    match status {
+        VMStatus::MoveAbort(location, code @ EACCOUNT_FROZEN)
+        | VMStatus::MoveAbort(location, code @ EINSUFFICIENT_BALANCE) => {
+            if location != known_locations::account_module_abort() {
+                crit!(
+                    "[libra_vm] Unexpected success epilogue move abort: {:?}::{:?}",
+                    location,
+                    code
+                );
+                return VMStatus::Error(StatusCode::UNEXPECTED_ERROR_FROM_KNOWN_MOVE_FUNCTION);
+            }
+            VMStatus::MoveAbort(location, code)
+        }
+
+        status @ VMStatus::Executed => status,
+
+        status => {
+            crit!("[libra_vm] Unexpected success epilogue error: {:?}", status);
+            VMStatus::Error(StatusCode::UNEXPECTED_ERROR_FROM_KNOWN_MOVE_FUNCTION)
+        }
+    }
+}
+
+/// Converts move aborts or execution failures to `REJECTED_WRITE_SET`
+/// Any other errors are mapped to the invariant violation
+/// `UNEXPECTED_ERROR_FROM_KNOWN_MOVE_FUNCTION`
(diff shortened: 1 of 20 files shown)


language/move-core/types/src/vm_status.rs
```diff
@@ -235,9 +235,26 @@ pub fn convert_prologue_runtime_error(status: VMStatus) -> VMStatus {
             };
             VMStatus::Error(new_major_status)
         }
-        status @ VMStatus::ExecutionFailure { .. }
-        | status @ VMStatus::Error(_)
+        status @ VMStatus::ExecutionFailure { .. } | status @ VMStatus::Executed => status,
+        VMStatus::Error(code) => {
+            debug_assert!(false, "Unexpected prologue/epilogue error: {:?}", code);
```
dariorussi, line 240: we should log critical here. We can do that at another time if needed. Just wanted to get consensus on that

language/move-core/types/src/vm_status.rs
```diff
@@ -235,9 +235,26 @@ pub fn convert_prologue_runtime_error(status: VMStatus) -> VMStatus {
             };
             VMStatus::Error(new_major_status)
         }
-        status @ VMStatus::ExecutionFailure { .. }
-        | status @ VMStatus::Error(_)
+        status @ VMStatus::ExecutionFailure { .. } | status @ VMStatus::Executed => status,
```
dariorussi, line 238: are we sending `ExecutionFailure` back? sorry I am getting confused here...

language/move-core/types/src/vm_status.rs
```diff
@@ -235,9 +235,26 @@ pub fn convert_prologue_runtime_error(status: VMStatus) -> VMStatus {
...
+/// Checks particular Move abort codes to specific validation error codes for the epilogue
+/// Any non-abort non-execution code is considered an invariant violation, specifically
+/// `UNEXPECTED_ERROR_FROM_EPILOGUE`
+pub fn convert_epilogue_error(status: VMStatus) -> VMStatus {
+    match status {
+        status @ VMStatus::MoveAbort(_, _)
+        | status @ VMStatus::ExecutionFailure { .. }
         | status @ VMStatus::Executed => status,
+        VMStatus::Error(code) => {
+            debug_assert!(false, "Unexpected prologue/epilogue error: {:?}", code);
```
dariorussi, line 255: logging critical here but this looks good!

language/move-core/types/src/vm_status.rs
tnowacki: ExecutionFailure and Executed are passed through. 
- Executed is obvious as to why (we want to be able to run the prologue in a success case)
- ExecutionFailure should be impossible given the definition, but I want to prevent any writeset nonsense from borking the VM here (i.e. the VM vomiting after the writeset caused the code to execution failure in some case)

language/move-core/types/src/vm_status.rs
tnowacki: Let me do this in a separate PR. There is a bunch of logging I want to add to this file

language/move-core/types/src/vm_status.rs
dariorussi: fine on `Executed` of course.
I don't understand `ExecutionFailure`. This is `prologue` handling, or is it? An error here should never charge the user, aren't we going to charge here if we get an `ExecutionFailure` during prologue?

language/move-core/types/src/vm_status.rs
```diff
@@ -741,12 +760,9 @@ impl StatusCode {
     /// checked in debug asserts
     pub fn should_skip_checks_for_todo(self) -> bool {
         fn todo_needs_execution_trace_information() -> std::collections::HashSet<StatusCode> {
-            vec![
-                StatusCode::VM_MAX_TYPE_DEPTH_REACHED,
-                StatusCode::VM_MAX_VALUE_DEPTH_REACHED,
-            ]
-            .into_iter()
-            .collect()
+            vec![StatusCode::VM_MAX_VALUE_DEPTH_REACHED]
```
bob-wilson, line 763: I don't understand this part. Why did you remove VM_MAX_TYPE_DEPTH_REACHED here? It doesn't seem to have anything to do with the rest of this PR anyway

vgao1996: I guess Todd and I talked about this. So one thing special about `VM_MAX_VALUE_DEPTH_REACHED` is that we don't throw this error immediately when a value is constructed via `Pack`, but only when the value gets converted info effects, before committing to storage, at which time the location info is already lost.

language/move-core/types/src/vm_status.rs
```diff
@@ -235,9 +235,89 @@ pub fn convert_prologue_runtime_error(status: VMStatus) -> VMStatus {
...
+            debug_assert!(false, "Unexpected prologue error: {:?}", code);
+            VMStatus::Error(StatusCode::UNEXPECTED_ERROR_FROM_KNOWN_MOVE_FUNCTION)
+        }
+    }
+}
+
+/// Checks for only move aborts, execution failures, and successful execution.
+/// Any other errors are mapped to the invariant violation
+/// `UNEXPECTED_ERROR_FROM_KNOWN_MOVE_FUNCTION`
+pub fn convert_normal_epilogue_error(status: VMStatus) -> VMStatus {
```
jolestar, line 249: It feels like these convert functions don't belong to Move VM Status and can be migrated over to LibraVm.

language/move-core/types/src/vm_status.rs
tnowacki: Yes. 

My worry here is that we fail on some arithmetic or something, and then by not charging, we are somewhat vulnerable to an attack. 

Additionally, I worry that some change in the prologue/epilogue (via a writeset) would *want* an execution error because of arithmetic or a missing resource or something similar. But then the VM would need to change to accommodate that writeset. 

@sblackshear @tzakian Any thoughts here? How likely is it that the prologue/epilogue could have a runtime error (div by zero, missing resource, etc)?

language/move-core/types/src/vm_status.rs
tnowacki: It's here because I suck. 

#5354 was supposed to fix it, and this commit was stacked on top of that one. I can make a separate PR if we really want

bob-wilson: I'll leave it up to you. It seems like it should be separate but I'm not sure if it's worth the hassle.

language/move-core/types/src/vm_status.rs
tnowacki: Logging was a lot easier once I moved the errors. So its up on the latest revision :)

language/move-core/types/src/vm_status.rs
tnowacki: Thanks for the suggestion! Cleaned things up a bit and made logging easier to boot!

language/move-core/types/src/vm_status.rs
tzakian: The likelihood of the prologue running into an internal error like this should be none, and should be treated in a similar manner to an invariant violation in the VM (i.e., no charge). Especially for the prologue no charge should ever happen if it fails since at that point we don't "know" whether the sending account actually owns the account...

For the epilogue, I would likewise classify execution failures arising from here as an invariant violation. I would likewise push to not charge in such instances. But, I could be convinced otherwise here if anyone feels strongly a particular way. 

In general we've taken a pretty strict view of the prologue/epilogue of "it should _never_ abort with anything other than the precise set of abort conditions that are expected" anything else should be treated as an invariant violation since we are essentially straying in to undefined transaction behavior at that point.

sblackshear: I agree with Tim. I think anything that doesn't fall into the expected behavior of the system should be a discard (in both prologue and epilogue).

tnowacki: Thanks for the input!

- I'll mark execution failures for the prologue as invariant violations
- For the epilogue, I will split it into success/failure ones. The success one will need to be more flexible to be able to kick over to the failure one

language/libra-vm/src/errors.rs
```diff
@@ -0,0 +1,140 @@
...
+            crit!("[libra_vm] Unexpected prologue error: {:?}", code);
+            VMStatus::Error(StatusCode::UNEXPECTED_ERROR_FROM_KNOWN_MOVE_FUNCTION)
+        }
+    }
+}
+
+/// Checks for only move aborts, execution failures, and successful execution.
+/// Any other errors are mapped to the invariant violation
+/// `UNEXPECTED_ERROR_FROM_KNOWN_MOVE_FUNCTION`
+pub fn convert_normal_epilogue_error(status: VMStatus) -> VMStatus {
```
sblackshear: I think the only expected results here should be `MoveAbort(INSUFFICIENT_FUNDS_FOR_GAS)` and `Executed`. Can we make the other cases invariant violations?

language/libra-vm/src/errors.rs
```diff
@@ -0,0 +1,140 @@
...
+            crit!("[libra_vm] Unexpected epilogue error: {:?}", code);
+            VMStatus::Error(StatusCode::UNEXPECTED_ERROR_FROM_KNOWN_MOVE_FUNCTION)
+        }
+    }
+}
+
+/// Converts move aborts and execution failures to `REJECTED_WRITE_SET`
+/// Any other errors are mapped to the invariant violation
+/// `UNEXPECTED_ERROR_FROM_KNOWN_MOVE_FUNCTION`
+pub fn convert_write_set_prologue_error(status: VMStatus) -> VMStatus {
```
sblackshear, line 90: @runtian-zhou will know for sure, but I think the only expected results here are the following Move aborts:

- `INVALID_WRITESET_SENDER`
- `EPROLOGUE_SEQUENCE_NUMBER_TOO_OLD`
- `E_WSPROLOGUE_SEQUENCE_NUMBER_TOO_NEW` (question for Runtian: do we need both of these)?
- `EPROLOGUE_INVALID_ACCOUNT_AUTH_KEY`

Everything else should be an invariant violation.

language/move-vm/runtime/src/session.rs
```diff
@@ -19,23 +20,26 @@ pub struct Session<'r, 'l, R> {
...
-    ) -> VMResult<()> {
-        self.runtime.execute_function(
-            module,
-            function_name,
-            ty_args,
-            args,
-            &mut self.data_cache,
-            cost_strategy,
-        )
+        error_specializer: F,
```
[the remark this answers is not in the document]
tnowacki: Correct, it just passes it through. 

I switched to this model because the conversions were ALL over the place. Some did it right after the function, some did it more upstream. 

This was the only way I could keep it sane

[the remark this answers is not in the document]
dariorussi: yes I have no issue with this as is. There is stuff to review in terms of integration with the Move VM so this does not break or make things. And it's a good centralization of error handling

language/libra-vm/src/errors.rs
```diff
@@ -0,0 +1,153 @@
...
+/// Any errors are mapped to the invariant violation
+/// `UNEXPECTED_ERROR_FROM_KNOWN_MOVE_FUNCTION`
+pub fn convert_block_prologue_error(status: VMStatus) -> VMStatus {
+    match status {
+        VMStatus::Executed => VMStatus::Executed,
+
+        status => {
+            crit!("[libra_vm] Unexpected block prologue error: {:?}", status);
+            VMStatus::Error(StatusCode::UNEXPECTED_ERROR_FROM_KNOWN_MOVE_FUNCTION)
+        }
```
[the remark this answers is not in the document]
dariorussi: I'd be fine with that

language/libra-vm/src/errors.rs
```diff
@@ -0,0 +1,147 @@
...
+        | VMStatus::MoveAbort(location, code @ EINSUFFICIENT_BALANCE) => {
+            if location != known_locations::account_module_abort() {
+                crit!(
+                    "[libra_vm] Unexpected success epilogue move abort: {:?}::{:?}",
+                    location,
+                    code
+                );
+                return VMStatus::Error(StatusCode::UNEXPECTED_ERROR_FROM_KNOWN_MOVE_FUNCTION);
+            }
+            VMStatus::MoveAbort(location, code)
```
bob-wilson, line 89: Should these 2 MoveAbort codes be translated to VMStatus::Error with the corresponding StatusCode::SENDING_ACCOUNT_FROZEN and StatusCode::INSUFFICIENT_BALANCE_FOR_TRANSACTION_FEE codes? This is the only case where prologue/epilogue aborts are returned as VMStatus::MoveAbort, so if this is really intentional, it would be good to have a comment here explaining why this is different than the other places.

tnowacki: Not really sure. Would be a good question for maybe @sblackshear @runtian-zhou or @tzakian 
My guess is that the other ones are converted so its easy on the validator (we could get rid of the conversion if we really wanted). If the success epilogue aborts, we don't really care why, we just want to then call the failure epilogue

PR #5398, opened 2020-07-30
[types] remove unnecessary rounding from exchange_rate function

bob-wilson:
The result of this function is a floating-point value so it doesn't make sense to round it based on a certain number of decimal digits, when the floating-point representation cannot exactly represent those decimal values anyway. Also, the comment here said that it was rounding up but the code was actually doing round-to-nearest. Regardless, there should not be any reason to round this result.

Yes

## Test Plan

Ran existing tests.

### DIFF
--- a/language/e2e-tests/src/tests/transaction_fuzzer.rs
+++ b/language/e2e-tests/src/tests/transaction_fuzzer.rs
@@ -36,6 +36,7 @@ proptest! {
     }
 
     #[test]
+    #[ignore]
     fn fuzz_scripts(
         txns in vec(any::<ScriptCall>(), 0..100),
     ) {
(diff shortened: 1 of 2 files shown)


bob-wilson:
The tests failed with:
---- tests::transaction_fuzzer::fuzz_scripts stdout ----
proptest: Aborting shrinking after the PROPTEST_MAX_SHRINK_ITERS environment variable or ProptestConfig.max_shrink_iters iterations (set 64 to a large(r) value to shrink more; current configuration: 64 iterations)

That doesn't seem to have anything to do with this change.

bob-wilson:
It failed again with the same error:

thread 'tests::transaction_fuzzer::fuzz_scripts' panicked at 'Test failed: assertion failed: !output.status().is_discarded() at language/e2e-tests/src/tests/transaction_fuzzer.rs:81; minimal failing input: txns = [ ...very long input... ]

I'm starting to wonder if this is somehow related to my change, but I can't figure out how. @tzakian any ideas?

bob-wilson:
@tzakian is looking into the flaky transaction fuzzing. In the meantime, he suggested that I add "[ignore]" to that test, which I have done. Let's try this again....

PR #5399, opened 2020-07-31
[network] move directional connections to different metric

gregnazario:
In order to not conflict with the current metrics, a different metric is
    kept for per peer and directional metrics all in one.

An example dashboard:
http://grafana.ct-2-k8s-testnet.aws.hlw3truzy4ls.com/d/bvXz6aVMk/validator-network-connections?orgId=1

### DIFF
--- a/network/src/peer_manager/mod.rs
+++ b/network/src/peer_manager/mod.rs
@@ -365,13 +365,16 @@ where
             .with_label_values(&[role, "connected"])
             .set(total as i64);
 
-        counters::LIBRA_NETWORK_PEERS
-            .with_label_values(&[role, "inbound"])
-            .set(inbound as i64);
-
-        counters::LIBRA_NETWORK_PEERS
-            .with_label_values(&[role, "outbound"])
-            .set(outbound as i64);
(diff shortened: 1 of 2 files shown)


network/src/peer_manager/mod.rs
```diff
@@ -365,13 +365,16 @@ where
             .with_label_values(&[role, "connected"])
             .set(total as i64);
 
-        counters::LIBRA_NETWORK_PEERS
-            .with_label_values(&[role, "inbound"])
-            .set(inbound as i64);
-
-        counters::LIBRA_NETWORK_PEERS
-            .with_label_values(&[role, "outbound"])
-            .set(outbound as i64);
```
davidiw, line 374: is this counter still being used? if not lets delete it

 we'll need to update dashboards.

davidiw: It is probably better to either rebase this onto the premainnet branch or to get it landed in master

gregnazario: The original counter is the one that's use for all the upstream peer information, and on dashboards.

PR #5400, opened 2020-07-31
[json-rpc] read page size limit from config instead of hardcode

xli:
## Motivation

In case the default page size is too big / small, full node should be able to adjust it in configuration.

## Test Plan

Unit test

### DIFF
--- a/json-rpc/src/methods.rs
+++ b/json-rpc/src/methods.rs
@@ -3,7 +3,7 @@
 
 //! Module contains RPC method handlers for Full Node JSON-RPC interface
 use crate::{
-    errors::{ErrorData, InvalidArguments, JsonRpcError},
+    errors::{ErrorData, ExceedSizeLimit, InvalidArguments, JsonRpcError},
     views::{
         AccountStateWithProofView, AccountView, BlockMetadata, CurrencyInfoView, EventView,
         StateProofView, TransactionView,
@@ -37,7 +37,8 @@ pub(crate) struct JsonRpcService {
     mempool_sender: MempoolClientSender,
     role: RoleType,
     chain_id: ChainId,
-    pub batch_size_limit: u16,
+    batch_size_limit: u16,
+    page_size_limit: u16,
 }
 
 impl JsonRpcService {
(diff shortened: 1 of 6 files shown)


json-rpc/src/runtime.rs
```diff
@@ -113,30 +122,26 @@ async fn rpc_endpoint(
...
-            });
-            let mut resp = Map::new();
-            resp.insert(
-                "error".to_string(),
-                JsonRpcError::invalid_request_with_data(Some(data)).serialize(),
-            );
+        match service.validate_batch_size_limit(requests.len()) {
+            Ok(_) => {
+                // batch API call
+                let futures = requests.into_iter().map(|req| {
```
christinacelee, line 128: I think we should be a bit more careful about concurrent write operations - if we have a batch request for multiple sequential txn submissions, the submissions might fail depending on what order they're executed

json-rpc/src/methods.rs
```diff
@@ -64,6 +67,28 @@ impl JsonRpcService {
     pub fn chain_id(&self) -> ChainId {
         self.chain_id
     }
+
+    pub fn validate_batch_size_limit(&self, size: usize) -> Result<(), JsonRpcError> {
+        self.validate_size_limit("batch size", self.batch_size_limit, size)
+    }
+
+    pub fn validate_page_size_limit(&self, size: usize) -> Result<(), JsonRpcError> {
+        self.validate_size_limit("page size", self.page_size_limit, size)
```
christinacelee, line 76: nit: should we have enums to represent the different size limits?

christinacelee: and my preference for the size limit names is snake_case, but your call

json-rpc/src/runtime.rs
```diff
@@ -113,30 +122,26 @@ async fn rpc_endpoint(
...
+                    rpc_request_handler(
+                        req,
+                        service.clone(),
+                        Arc::clone(&registry),
+                        ledger_info.clone(),
+                    )
+                });
+                let responses = join_all(futures).await;
```
christinacelee, line 136: fwiw if any of the futures fail, `join_all` as a whole will fail, so we should be clear on whether what JSON RPC's policy is on batched responses (should it return an error for an entire batch if any of the requests are bad?)

json-rpc/src/methods.rs
xli: It's a message to client (in the struct field "name"). So keep it simple and human readable format.

PR #5401, opened 2020-07-31
[libra-node] use chain ID from genesis, not config

christinacelee:
## Motivation

Preparation for removing chain ID from config 
In node initialization, read and use chain ID from genesis, not config

## Test Plan

Existing tests

### DIFF
--- a/libra-node/src/main_node.rs
+++ b/libra-node/src/main_node.rs
@@ -16,11 +16,14 @@ use libra_json_rpc::bootstrap_from_config as bootstrap_rpc;
 use libra_logger::prelude::*;
 use libra_mempool::gen_mempool_reconfig_subscription;
 use libra_metrics::metric_server;
+use libra_types::{
+    account_config::libra_root_address, account_state::AccountState, chain_id::ChainId,
+};
 use libra_vm::LibraVM;
 use libradb::LibraDB;
 use network_builder::builder::NetworkBuilder;
 use state_synchronizer::StateSynchronizer;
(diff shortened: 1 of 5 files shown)


libra-node/src/main_node.rs
```diff
@@ -38,6 +42,21 @@ pub struct LibraHandle {
...
+        .map(|blob| {
+            AccountState::try_from(&blob).and_then(|state| {
+                Ok(state
+                    .get_chain_id_resource()?
+                    .ok_or_else(|| format_err!("ChainIdResource does not exist"))?
+                    .chain_id())
+            })
+        })
+        .ok_or_else(|| format_err!("Failed to fetch ChainIdResource"))?
+}
```
davidiw, line 58: I think we can use expects here.

libra-node/src/main_node.rs
```diff
@@ -105,6 +124,10 @@ pub fn setup_environment(node_config: &NodeConfig) -> LibraHandle {
         "ChunkExecutor setup in {} ms",
         instant.elapsed().as_millis()
     );
+    let chain_id = match fetch_chain_id(&db_rw) {
+        Ok(chain_id) => chain_id,
+        Err(e) => panic!("ChainId missing in node initialization: {:?}", e),
+    };
```
davidiw, line 130: using expect above, simplifies this.

christinacelee:
> why not remove chainid from the config

I'll do this in a separate PR, it's too scary until config-builder is fully deprecated everywhere

PR #5404, opened 2020-07-31
[management] pretty print op tool

gregnazario:
Ensure these two pretty print commits get into premainnet
https://github.com/libra/libra/pull/5378
https://github.com/libra/libra/pull/5392


davidiw:
Manually merged. Thanks!

PR #5405, opened 2020-07-31
[breaking] Add nonce to LR called scripts

moeziniaf:
- Add sliding nonce to CreateParentVasp, UpdateLibraVersion
- expose LibraVersion in AccountState
Test: update_libra_version stdlib_script test


types/src/account_state.rs
```diff
@@ -128,6 +130,10 @@ impl AccountState {
         self.get_resource(&ValidatorSet::CONFIG_ID.access_path().path)
     }
 
+    pub fn get_libra_version(&self) -> Result<Option<LibraVersion>> {
+        self.get_resource(&LibraVersion::CONFIG_ID.access_path().path)
```
dmitri-perelman, line 134: 👍

PR #5410, opened 2020-07-31
[dkr] Don't strip libra-node and safety-rules

mgorven:
## Motivation

For now we still want somewhat useful stacktraces in prod.

## Test Plan

Built the validator image.


mgorven:
@bor-libra r=sausagee

PR #5414, opened 2020-07-31
[cluster-test] enable gas price into cluster test, which allow to setup non-zero gas price

snakech:
## Motivation

enable configurable non zero gap price in cluster test
default gas_price = 1, use ```-E GAS_UNIT_PRICE=1``` to set non-zero gas price

(Write your answer here.)

## Test Plan

tested with gas price = 1 with both swarm and cti:
```
cargo run -p cluster-test -- --mint-file "/var/folders/qd/tg50mr0d7k75qvztr_8yk2km0000gn/T/c81ae98e57fdc3fd55f563d37c7eb3fa/mint.key" --swarm --peers "localhost:49536" --emit-tx --workers-per-ac 1
```
```
./scripts/cti -E GAS_UNIT_PRICE=1 --pr 5414 --run bench
```

## Related PRs

### DIFF
--- a/testsuite/cluster-test/src/tx_emitter.rs
+++ b/testsuite/cluster-test/src/tx_emitter.rs
@@ -5,7 +5,7 @@
 
 use crate::{atomic_histogram::*, cluster::Cluster, instance::Instance};
 use std::{
-    fmt, slice,
+    env, fmt, slice,
     sync::Arc,
     time::{Duration, Instant},
 };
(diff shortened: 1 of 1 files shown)


testsuite/cluster-test/src/tx_emitter.rs
```diff
@@ -107,6 +108,14 @@ pub struct EmitJobRequest {
     pub thread_params: EmitThreadParams,
 }
 
+pub static GAS_UNIT_PRICE: Lazy<u64> = Lazy::new(|| {
+    if let Ok(v) = env::var("GAS_UNIT_PRICE") {
+        v.parse().expect("Failed to parse GAS_UNIT_PRICE")
+    } else {
+        0_u64
```
andll, line 115: Can we set default to `1`?

snakech: observed 20% to 30% perf regression if we default gas_price = 1, just set it back to 0 and we can change this in future. Good news is I confirmed with platform team which this perf regression is expected

PR #5416, opened 2020-07-31
[move prover] improvements to pure move function call feature

emmazzz:
## Motivation

This PR made a few improvements to the pure function call feature in Move Prover:
- Pure functions which call other pure functions are allowed and won’t result in boogie compilation error anymore.
- Pure-looking functions which call impure functions are now detected as impure and using them in specs are not allowed. For example:
```rust
fn pure_looking_fun(): u64 {
     impure_fun() + 1 // not allowed
 }
```
- Functions in the following format are allowed and the `assert` statements are ignored.
 ```rust
 fn pure_fun(): some_type {
     assert(…); // ignored
     assert(…); // ignored
     pure_expression // allowed
 }
 ```

Yes.

## Test Plan

Added test cases.

### DIFF
--- a/language/move-prover/spec-lang/src/ast.rs
+++ b/language/move-prover/spec-lang/src/ast.rs
@@ -387,6 +387,9 @@ pub enum Operation {
     MaxU8,
     MaxU64,
     MaxU128,
+
+    // Operation with no effect
+    NoOp,
 }
 
 #[derive(Debug, Clone, PartialEq)]
--- a/language/move-prover/spec-lang/src/env.rs
+++ b/language/move-prover/spec-lang/src/env.rs
@@ -421,7 +421,7 @@ pub struct GlobalEnv {
     /// A map from global memories to global invariants which refer to them.
     global_invariants_for_memory: BTreeMap<QualifiedId<StructId>, BTreeSet<GlobalId>>,
     /// A set containing spec functions which are called/used in specs.
-    used_spec_funs: BTreeSet<QualifiedId<SpecFunId>>,
+    pub used_spec_funs: BTreeSet<QualifiedId<SpecFunId>>,
 }
 
 /// Information about a verification condition stored in the environment.
@@ -695,9 +695,10 @@ impl GlobalEnv {
             .unwrap_or_else(Default::default)
     }
 
-    /// Adds a spec function to used_spec_funs set.
-    pub fn add_used_spec_fun(&mut self, module_id: ModuleId, spec_fun_id: SpecFunId) {
-        self.used_spec_funs.insert(module_id.qualified(spec_fun_id));
+    /// Returns true if a spec fun is used in specs.
+    pub fn is_spec_fun_used(&self, module_id: ModuleId, spec_fun_id: SpecFunId) -> bool {
+        self.used_spec_funs
+            .contains(&module_id.qualified(spec_fun_id))
     }
 
     /// Adds a new module to the environment. StructData and FunctionData need to be provided
--- a/language/move-prover/spec-lang/src/translate.rs
+++ b/language/move-prover/spec-lang/src/translate.rs
@@ -83,6 +83,8 @@ pub struct Translator<'env> {
     fun_table: BTreeMap<QualifiedSymbol, FunEntry>,
     /// A symbol table for constants.
     const_table: BTreeMap<QualifiedSymbol, ConstEntry>,
+    /// A call graph mapping callers to callees that are move functions.
+    move_fun_call_graph: BTreeMap<QualifiedId<SpecFunId>, BTreeSet<QualifiedId<SpecFunId>>>,
 }
 
 /// A declaration of a specification function or operator in the translator state.
@@ -168,6 +170,7 @@ impl<'env> Translator<'env> {
             reverse_struct_table: BTreeMap::new(),
             fun_table: BTreeMap::new(),
             const_table: BTreeMap::new(),
+            move_fun_call_graph: BTreeMap::new(),
         };
         translator.declare_builtins();
         translator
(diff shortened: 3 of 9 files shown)


language/move-prover/spec-lang/src/translate.rs
```diff
@@ -4025,6 +4091,12 @@ impl<'env, 'translator, 'module_translator> ExpTranslator<'env, 'translator, 'mo
         }
         // Next treat this as a call to a global function.
         let (module_name, name) = self.parent.module_access_to_parts(maccess);
+
+        // Ignore assert statement.
```
emmazzz, line 4095: Do we want to allow `abort` as well?

language/move-prover/spec-lang/src/ast.rs
```diff
@@ -248,6 +248,7 @@ pub enum Exp {
     Lambda(NodeId, Vec<LocalVarDecl>, Box<Exp>),
     Block(NodeId, Vec<LocalVarDecl>, Box<Exp>),
     IfElse(NodeId, Box<Exp>, Box<Exp>, Box<Exp>),
+    Empty(NodeId),
```
wrwg, line 251: Instead of introducing a new expression node, could you introduce instead Operation:::Empty and represent this as Call(node_id, Empty, [])? This is the current design philosophy here. The number of Exp variants should be kept to the minimum. This would also prevent the need to extend various matches here and there.

language/move-prover/spec-lang/src/env.rs
```diff
@@ -691,7 +694,41 @@ impl GlobalEnv {
...
+            .or_insert_with(BTreeSet::new)
+            .insert(callee_mid.qualified(callee_fid));
+    }
+
+    /// Runs DFS to propagate the usage of move functions from callers
+    /// to callees on the call graph.
+    pub fn propagate_move_fun_usage(&mut self, qid: QualifiedId<SpecFunId>) {
+        if let Some(neighbors) = self.call_graph.get(&qid) {
+            neighbors.clone().iter().for_each(|n| {
+                if self.used_spec_funs.insert(*n) {
```
wrwg, line 727: Add a comment here that via the if here, you deal with recursion.

wrwg: Also, you only do the callgraph for move functions, right? Should then lokely call the field move_fun_call_graph.

language/move-prover/spec-lang/src/env.rs
```diff
@@ -416,6 +416,8 @@ pub struct GlobalEnv {
     global_invariants_for_memory: BTreeMap<QualifiedId<StructId>, BTreeSet<GlobalId>>,
     /// A set containing spec functions which are called/used in specs.
     used_spec_funs: BTreeSet<QualifiedId<SpecFunId>>,
+    /// A call graph mapping callers to callees.
+    call_graph: BTreeMap<QualifiedId<SpecFunId>, BTreeSet<QualifiedId<SpecFunId>>>,
```
wrwg, line 420: Does this need to be in the GlobalEnv, can it be in the translator instead?

language/move-prover/spec-lang/src/translate.rs
```diff
@@ -1423,7 +1440,7 @@ impl<'env, 'translator> ModuleTranslator<'env, 'translator> {
 impl<'env, 'translator> ModuleTranslator<'env, 'translator> {
     /// Definition analysis for move functions.
     /// If the function is pure, we translate its body.
-    fn def_ana_fun(&mut self, name: &PA::FunctionName, body: &EA::FunctionBody) {
+    fn def_ana_fun(&mut self, name: &PA::FunctionName, body: &EA::FunctionBody, fun_idx: usize) {
```
wrwg, line 1443: The fun_idx parameter is confusing. Below you use SpecFunId::new(fun_idx), but at the caller side you use the definition index of the Move function. How is that possible? Wouldn't that clash with spec fun ids created from real spec funs? Whatever is going on here, please use more descriptive parameter names.

language/move-prover/spec-lang/src/translate.rs
```diff
@@ -1446,21 +1463,63 @@ impl<'env, 'translator> ModuleTranslator<'env, 'translator> {
...
         }
         self.spec_fun_index += 1;
     }
+
+    /// Propagate the impurity of move functions from callees to callers so
+    /// that we can detect pure-looking move functions which calls impure
+    /// move functions.
+    fn propagate_function_impurity(
+        &mut self,
+        mut visited: &mut BTreeMap<usize, bool>,
```
wrwg, line 1487: Instead of using usize, using the id type (whatever it is, SpecFunId or FunId) would help to make the code better understandable.

language/move-prover/spec-lang/src/translate.rs
```diff
@@ -4025,6 +4087,12 @@ impl<'env, 'translator, 'module_translator> ExpTranslator<'env, 'translator, 'mo
         }
         // Next treat this as a call to a global function.
         let (module_name, name) = self.parent.module_access_to_parts(maccess);
+
+        // Ignore assert statement.
+        if name == self.parent.parent.assert_symbol() {
```
wrwg, line 4092: Guess we need to check for the qualified name which I think move_lang has created after expansion, otherwise every function the user calls assert shadowing the builtin assert will be also ignored. Not sure though how move_lang represents those calls.

language/move-prover/spec-lang/src/translate.rs
wrwg: Lets stay with assert for now.

language/move-prover/spec-lang/src/env.rs
emmazzz: This call graph also includes all spec functions(both actual spec fun and move-fun-converted spec fun) although we currently don't use the information of whether an actual spec fun is used in later translation. I will change the function name or the field name to be more consistent with each other.

language/move-prover/spec-lang/src/env.rs
emmazzz: Yes translator is a better place to put it.

language/move-prover/spec-lang/src/translate.rs
emmazzz: Will use use more descriptive parameter names.

Because the `decl_ana` for move functions comes before spec functions, the `SpecDecl` for move functions will be pushed into `ModuleTranslator.spec_funs` before the `SpecDecl` for spec functions. Therefore, the definition index of the Move function is the same as the spec fun index for the spec version of that Move function, and it won't clash with real spec funs.

language/move-prover/spec-lang/src/translate.rs
emmazzz: Move lang doesn't seem to allow users to shadow the builtin assert, based on my shallow experiment.

PR #5417, opened 2020-07-31
[gha][lbt] report flaky when report missing

rustielin:
Report failures on runs like this: https://github.com/libra/libra/runs/926203494, which fail with no report generated.


andll:
Btw, did we test if this works?
I am really surprised that the `ret=1` would fix the issue(as I am surprised that cti returned wrong error code)

Also, for future debugging, do we want to print cti exit code right after it completes?

PR #5420, opened 2020-07-31
ci: pin rustc for all invocations of cargo

bmwill:
Pin the toolchain used for all bare invocations of rustc. Since we're executing a nightly version of cargo directly in CI, rustup isn't involved initially, meaning that our `rust-toolchain` file is not respected when building third-party crates.


metajack:
That close button is so close to "Comment" :)

PR #5422, opened 2020-07-31
Update some comments on public APIs and types

bob-wilson:
This cleans up and clarifies some comments that I came across while reviewing the libra-vm validation code.

## Motivation

Improving documentation in the code

Yes

## Test Plan

Ran existing tests.

### DIFF
--- a/types/src/transaction/mod.rs
+++ b/types/src/transaction/mod.rs
@@ -55,33 +55,39 @@ pub const PRE_GENESIS_VERSION: Version = u64::max_value();
 
 pub const MAX_TRANSACTION_SIZE_IN_BYTES: usize = 4096;
 
-/// RawTransaction is the portion of a transaction that a client signs
+/// RawTransaction is the portion of a transaction that a client signs.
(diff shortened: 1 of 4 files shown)


types/src/transaction/mod.rs
```diff
@@ -633,8 +642,17 @@ impl GovernanceRole {
...
 pub struct VMValidatorResult {
+    /// Result of the validation: `None` if the transaction was successfully validated
+    /// or `Some(DiscardedVMStatus)` if the transaction should be discarded.
     status: Option<DiscardedVMStatus>,
+
+    /// Score for ranking the transaction priority (e.g., based on the gas price).
+    /// Only used when the status is `None`. Higher values indicate a higher priority.
     score: u64,
+
+    /// The account role for the transaction sender, so that the certain
```
dariorussi, line 653: remove "the"? "so that certain..."

bob-wilson:
> Overall I think we should stay away from Rust comments `///` in favor of normal comments `//` but I see this was more to make things more or less consistent....

See the "Code documentation" section of documentation/coding_guidelines.md: "Any public fields, functions, and methods should be documented with Rustdoc". I'm not sure we're following that style very well, but we should at least work toward that direction.

PR #5423, opened 2020-07-31
[breaking][vm] Add sender signer to `AdminScript`s

tzakian:
This PR adds an additional signer for the sender of the transaction to
admin script arguments (see the updated admin script transaction test for an example of what this looks like). This updates the bytecode verifier to allow multiple signers, but requires that all signer arguments are a prefix of the `main()` arguments.


language/bytecode-verifier/src/verifier.rs
```diff
@@ -53,23 +53,25 @@ pub fn verify_main_signature(script: &CompiledScript) -> VMResult<()> {
...
-        use SignatureToken as S;
-        match arg_type {
-            // &signer is a type that can only be populated by the Move VM. And its value is filled
-            // based on the sender of the transaction
-            S::Reference(inner) => idx == 0 && matches!(&**inner, S::Signer),
-            _ => arg_type.is_valid_for_constant(),
-        }
+    use SignatureToken as S;
+    let arguments = &script.signature_at(script.as_inner().parameters).0;
+    // Remove the prefix of signer types
```
sblackshear, line 58: Nit: comment slightly confusing because this isn't actually removing anything. Maybe something like "check that all `&signer` formals occur before non-`&signer` formals"?

language/libra-vm/src/libra_transaction_executor.rs
```diff
@@ -307,23 +307,26 @@ impl LibraVM {
         &self,
         remote_cache: &StateViewCache<'_>,
         writeset_payload: &WriteSetPayload,
+        txn_data: Option<&TransactionMetadata>,
```
sblackshear, line 310: Could this be `sender:  Option<AccountAddress>` (since that's the only thing we read from `txn_data`)?

moeziniaf:
Will we need change the args of all scripts we want to call as admin scripts? For example, I imagine `set_validator_operator` can now take an `lr_account: &signer` and do the sliding nonce check. However what about `rotate_authentication_key_with_nonce` which we want to use both as a regular script and admin script?

tzakian:
> Will we need change the args of all scripts we want to call as admin scripts? For example, I imagine set_validator_operator can now take an lr_account: &signer and do the sliding nonce check. However what about rotate_authentication_key_with_nonce which we want to use both as a regular script and admin script?

@moeziniaf Good question. For `rotate_authentication_key_with_nonce` you would need two separate versions of the script, one to be used as an admin script, and the other to be used  normally (but, the admin script should not be allowlisted, so should not be included in the stdlib) 

@dariorussi I'm not sure if I understand what you're saying, could you possibly explain a bit more? This approach seems like the natural way of doing this to me, and I'm not sure how the logic in rust could do the right thing without passing it over to the Move side of things.

language/bytecode-verifier/src/verifier.rs
```diff
@@ -53,23 +53,25 @@ pub fn verify_main_signature(script: &CompiledScript) -> VMResult<()> {
...
+    let args_checkout = arguments
```
tnowacki, line 59: Maybe a more "propositional" sort of name here? `let all_args_have_valid_type` or something

language/ir-testsuite/tests/signer/double_signer.mvir
```diff
@@ -1,14 +1,14 @@
 main(s: &signer, s2: &signer) {
     return;
 }
-// check: INVALID_MAIN_FUNCTION_SIGNATURE
+// check: TYPE_MISMATCH
```
tnowacki, line 4: I'm confused. This is the wrong error message. How did this get changed?

tzakian: I believe this is the correct error message: before two-signer txns were disallowed, however with this change two-signer transactions are allowed as long as they are a prefix. So this is no longer an invalid main signature, but a type error since two signers aren't passed through.

tnowacki: that is a really large change to be doing right now no? How are we putting in more than one signer right now? I didn't see anywhere that did that

tzakian: It's not a terribly large change I would argue since this is only allowed for admin scripts (and passes both the sender and "as account" signers so has well-defined behavior). You can see this in the flow for the `AdminScript` I'll comment on this PR and call that part out.

language/libra-vm/src/libra_transaction_executor.rs
```diff
@@ -307,23 +307,26 @@ impl LibraVM {
...
         Ok(match writeset_payload {
             WriteSetPayload::Direct(change_set) => change_set.clone(),
-            WriteSetPayload::Script {
-                script,
-                execute_as: signer,
-            } => {
+            WriteSetPayload::Script { script, execute_as } => {
                 let mut tmp_session = self.0.new_session(remote_cache);
+                let mut args = convert_txn_args(script.args());
+                args.insert(0, Value::transaction_argument_signer_reference(*execute_as));
```
tzakian, line 320: @tnowacki This is where the other signer is added. It then follows the normal transaction flow that will add the sender's `&signer` to the arguments.

tnowacki: I don't think we should be doing this at this point if we can make do with what we have. 

Changing the rules for the bytecode verifier and how we feed in signer information into an admin script is a pretty major change. Especially since the bytecode verifier now accepts all sorts of different scripts it didn't before (granted there is no way to actually invoke them, but it is still a large change)

I'll book some time to discuss this

language/move-vm/runtime/src/runtime.rs
```diff
@@ -106,11 +106,21 @@ impl VMRuntime {
...
 
-        // build the arguments list for the main and check the arguments are of restricted types
-        let first_param_opt = main.parameters().0.get(0);
-        if first_param_opt.map_or(false, |sig| is_signer_reference(sig)) {
-            args.insert(0, Value::transaction_argument_signer_reference(sender))
+        // Build the arguments list for the main and check the arguments are of restricted types.
+        // Signers are built up from left-to-right. Not all signer arguments passed in may be used (e.g., if
+        // the script doesn't take a &signer, or if it isn't an admin script with one &signer).
+        let parameters = &main.parameters().0;
+        for (i, signer) in signers.into_iter().enumerate() {
```
dariorussi, line 113: do we want to impose some correlation between signers passed in and expected?
I suppose we cannot because of the fact not all script take a signer, but I wonder....

language/move-vm/runtime/src/runtime.rs
```diff
@@ -90,7 +90,7 @@ impl VMRuntime {
         script: Vec<u8>,
         ty_args: Vec<TypeTag>,
         mut args: Vec<Value>,
-        sender: AccountAddress,
+        signers: Vec<Value>,
```
dariorussi, line 93: this could be a Vector<AccountAddress> right?

language/libra-vm/src/libra_transaction_executor.rs
```diff
@@ -170,7 +171,9 @@ impl LibraVM {
                     script.code().to_vec(),
                     script.ty_args().to_vec(),
                     convert_txn_args(script.args()),
-                    txn_data.sender(),
+                    vec![Value::transaction_argument_signer_reference(
```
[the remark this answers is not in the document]
tzakian: Should be updated now: we only pass a `Vec<AccountAddress` in now and no references are passed in.

language/move-vm/runtime/src/runtime.rs
tzakian: It's a good question, but I think not imposing a correlation is fine for a couple reasons:
* Historically we've allowed this as you mentioned, but also
* Using a prefix of the signers is at most as powerful as using the full list of signers, so from a privilege standpoint, not using all of the signers is fine. Essentially the VM is saying "these are all the account privileges you have" and the script is saying "ok I'll only take a couple" here which seems sensible to me.

language/move-vm/runtime/src/runtime.rs
tzakian: yes! Updated the PR to make this change.

language/libra-vm/src/libra_transaction_executor.rs
```diff
@@ -307,23 +308,26 @@ impl LibraVM {
...
-                script,
-                execute_as: signer,
-            } => {
+            WriteSetPayload::Script { script, execute_as } => {
                 let mut tmp_session = self.0.new_session(remote_cache);
+                let args = convert_txn_args(script.args());
+                let mut senders = vec![*execute_as];
+                if let Some(sender) = txn_sender {
+                    senders.push(sender)
+                }
```
tnowacki, line 324: To each their own here
But I'd probably do 
```
let senders = match txn_sender {
  Some(sender) => vec![*execute_as, sender],
  None => vec![*execute_as]
}
```

language/move-lang/src/typing/translate.rs
```diff
@@ -119,7 +124,7 @@ fn check_primitive_script_arg(context: &mut Context, mloc: Loc, idx: usize, ty:
         result.is_ok()
     };
     if is_signer_ref {
-        if idx == 0 {
+        if !*seen_non_signer {
```
tnowacki, line 127: test for this?

language/move-vm/runtime/src/runtime.rs
```diff
@@ -106,11 +106,21 @@ impl VMRuntime {
...
         let (main, type_params) = self.loader.load_script(&script, &ty_args, data_store)?;
 
-        // build the arguments list for the main and check the arguments are of restricted types
-        let first_param_opt = main.parameters().0.get(0);
-        if first_param_opt.map_or(false, |sig| is_signer_reference(sig)) {
-            args.insert(0, Value::transaction_argument_signer_reference(sender))
+        // Build the arguments list for the main and check the arguments are of restricted types.
+        // Signers are built up from left-to-right. Not all signer arguments passed in may be used (e.g., if
+        // the script doesn't take a &signer, or if it isn't an admin script with one &signer).
+        let parameters = &main.parameters().0;
```
tnowacki, line 112: This is... a bit peculiar. It feels brittle in how it would work out. 

Would it be possible to take all or nothing here? Either no signers, or you need a signer for every sender

language/move-lang/src/typing/translate.rs
tzakian: The previous tests are testing for this (unless I'm reading them wrong). Happy to add a couple more.

language/move-vm/runtime/src/runtime.rs
tzakian: We could take all or nothing. One reason I preferred this is because this allows reusing "normal" scripts in an admin script context. Happy to go either way, but I do have a _slight_ preference for this "prefix" semantics.

tnowacki: Lets see how other people feel
@dariorussi @sblackshear any thoughts?

tzakian:
Unless anyone has strong objections, I would like to try and get this PR in today since it is blocking work on the AOS side of things. Happy to land some followup PRs to cleanup as we see fit later on.

tnowacki:
I'd prefer to se the all or nothing approach than this go in (since all or nothing is more conservative). We could relax it later if we wanted?

Thoughts?

tzakian:
I'm fine with moving to the all-or-nothing approach. I'll start making those changes now.

language/move-lang/src/typing/translate.rs
tnowacki: Didn't see the move tests being updated (maybe the push for that happened after I commented? unsure)
Regardless, the tests look fine

tnowacki: Looks like we need to update the error message

language/move-lang/tests/move_check/translated_ir_tests/signer/triple_signer.exp
```diff
@@ -0,0 +1,33 @@
+error: 
+
+    ┌── tests/move_check/translated_ir_tests/signer/triple_signer.move:13:9 ───
+    │
+ 13 │     fun t1(_s: &signer, _s2: &signer, u: u64, _s3: &signer) {
+    │         ^^ Invalid parameter for script function 't1'
+    ·
+ 13 │     fun t1(_s: &signer, _s2: &signer, u: u64, _s3: &signer) {
+    │                                                    ------- '&signer' must be the first argument to a script
```
tnowacki, line 9: Let's update this error message

language/e2e-tests/src/tests/admin_script.rs
```diff
@@ -77,3 +78,54 @@ fn admin_script_rotate_key() {
...
+
+    let privkey = Ed25519PrivateKey::generate_for_testing();
+    let pubkey = privkey.public_key();
+    let new_key_hash = AuthenticationKey::ed25519(&pubkey).to_vec();
+
+    let script_body = {
+        let code = "
+    import 0x1.LibraAccount;
+
+    main(account: &signer, auth_key_prefix: vector<u8>) {
```
moeziniaf, line 96: why does this take a single signer (and above double signer)?

language/stdlib/transaction_scripts/remove_validator_and_reconfigure_admin.move
```diff
@@ -0,0 +1,21 @@
+script {
+    use 0x1::LibraSystem;
+    use 0x1::SlidingNonce;
+    use 0x1::ValidatorConfig;
+
+    /// Removes a validator from the validator set.
+    /// Fails if the validator_address is not in the validator set.
+    /// Emits a NewEpochEvent.
+    /// Can only be called as an administrative script to trigger waypoints.
+    fun remove_validator_and_reconfigure_admin(
```
moeziniaf, line 10: cc @kchalkias didn't realize this operation would be admin script but this is what you'll need to use

language/stdlib/transaction_scripts/remove_validator_and_reconfigure_admin.move
```diff
@@ -0,0 +1,21 @@
...
+        _: &signer,
```
moeziniaf, line 11: why isn't LR the first param here?

language/e2e-tests/src/tests/admin_script.rs
tzakian: This is a test to make sure that it fails if it only uses one signer when it should take two.

language/stdlib/transaction_scripts/remove_validator_and_reconfigure_admin.move
moeziniaf, line 10: why does this have to be an admin script at all? why can't we just assert the sender is LR? will other accounts be able to remove validators in the future?

language/stdlib/transaction_scripts/remove_validator_and_reconfigure_admin.move
tzakian: Both of these signers are the LR account: in the case that the admin script is sent as a waypoint this is the script that is used. There is also a non-admin version of this script as well that only takes the single `lr_account` signer (`remove_validator_and_reconfigure`).

tzakian:
Shouldn't be an issue to add the rotate_with_nonce as a non-admin script. I'll update now.

language/move-vm/runtime/src/runtime.rs
```diff
@@ -106,11 +106,21 @@ impl VMRuntime {
...
 
-        // build the arguments list for the main and check the arguments are of restricted types
-        let first_param_opt = main.parameters().0.get(0);
-        if first_param_opt.map_or(false, |sig| is_signer_reference(sig)) {
-            args.insert(0, Value::transaction_argument_signer_reference(sender))
+        // Build the arguments list for the main and check the arguments are of restricted types.
+        // Signers are built up from left-to-right. Either all signer arguments are used, or no
+        // signer arguments can be be used by a script.
+        let parameters = &main.parameters().0;
+        if parameters.get(0).map_or(false, is_signer_reference) {
```
tnowacki, line 113: nit. maybe bind this to a variable for readability. Just so its easy to see this is checking if a signer argument exists

language/libra-vm/src/libra_transaction_executor.rs
```diff
@@ -308,23 +309,26 @@ impl LibraVM {
...
         let gas_schedule = zero_cost_schedule();
         let mut cost_strategy = CostStrategy::system(&gas_schedule, GasUnits::new(0));
 
         Ok(match writeset_payload {
             WriteSetPayload::Direct(change_set) => change_set.clone(),
-            WriteSetPayload::Script {
-                script,
-                execute_as: signer,
-            } => {
+            WriteSetPayload::Script { script, execute_as } => {
```
tnowacki, line 319: Maybe we could make `execute_as` an option in another PR? Would that alleviate some of the pain of the all or nothing signers?

language/libra-vm/src/libra_transaction_executor.rs
```diff
@@ -308,23 +309,26 @@ impl LibraVM {
...
                 let mut tmp_session = self.0.new_session(remote_cache);
+                let args = convert_txn_args(script.args());
+                let senders = match txn_sender {
+                    Some(sender) => vec![*execute_as, sender],
+                    None => vec![*execute_as, *execute_as],
```
tnowacki, line 324: why do we need to double up here?

tnowacki: i.e. why not just `vec![*execute_as]`?

tzakian: I went for the double-up since it seems odd that some admin scripts would take one signer, and some would take two...but I like this idea that you had. Let me try it out :)

language/libra-vm/src/libra_transaction_executor.rs
tzakian: I don't think execute_as should be an option: if it were `None` it would be basically the same as a normal transaction sent from the Libra Root account.

tnowacki: The more I think about it, I think an option will be the right thing. Before this PR, this was the only way to set the signer. Now that we don't always need it, it feels a bit silly

tzakian: I don't think an option here is needed/good. If Libra Root sends an admin script with `execute_as` set to `None` how is that different from Libra Root sending a normal (user) transaction? 

There are cases where an admin script is used, but where there is no sender, in particular for waypoints in the system. So the sender being an option makes sense, but the `execute_as` being an option doesn't no?

tnowacki: Ah yes, that makes sense. Maybe cleaning up the `sender`s below will help this crystalize for me

language/move-lang/tests/move_check/translated_ir_tests/signer/triple_signer.exp
```diff
@@ -0,0 +1,33 @@
+error: 
+
+    ┌── tests/move_check/translated_ir_tests/signer/triple_signer.move:13:9 ───
+    │
+ 13 │     fun t1(_s: &signer, _s2: &signer, u: u64, _s3: &signer) {
+    │         ^^ Invalid parameter for script function 't1'
+    ·
+ 13 │     fun t1(_s: &signer, _s2: &signer, u: u64, _s3: &signer) {
+    │                                                    ------- '&signer's must be a prefix of the arguments to a script
```
tnowacki, line 9: Might be worth explaining this more. Feels maybe a tad compilery
`"'&signer's must be a prefix of the script arguments--they must come before any non-signer types"`

language/stdlib/transaction_scripts/remove_validator_and_reconfigure_admin.move
tnowacki: This was cleaned up by the single `execute_as` above?

tzakian: yep! This is because we were no longer passing in the doubled-up "execute_as" sender through (since this transaction was being sent by the VM for waypoints).

language/move-vm/runtime/src/runtime.rs
sblackshear: All-or-nothing seems like a good discipline to me (but we can always fix it once we allow "true" multisigner txes without forging).

PR #5425, opened 2020-07-31
[cluster-test] Initialize Vaults and genesis using management tools

changyonghu:
Initialize Vault keys and generate genesis.blob in cluster test pod and temporarily deliver them to each validator pod, /opt/libra/etc/genesis2.blob

### DIFF
--- a/testsuite/cluster-test/Cargo.toml
+++ b/testsuite/cluster-test/Cargo.toml
@@ -28,6 +28,7 @@ rusoto_autoscaling = { version = "0.45.0", default-features = false, features = 
 rusoto_sts = { version = "0.45.0", default-features = false, features = ["rustls"] }
 rusoto_s3 = { version = "0.45.0", default-features = false, features = ["rustls"] }
 chrono = "0.4.13"
+toml = { version = "0.5.6", default-features = false }
 
 debug-interface = { path = "../../common/debug-interface", version = "0.1.0"}
 libra-json-rpc-client = { path = "../../client/json-rpc", version = "0.1.0"}
@@ -38,7 +39,13 @@ config-builder = { path = "../../config/config-builder", version = "0.1.0" }
 generate-key = { path = "../../config/generate-key", version = "0.1.0" }
 libra-crypto = { path = "../../crypto/crypto", version = "0.1.0" }
 libra-config = { path = "../../config", version = "0.1.0" }
+libra-genesis-tool = { path = "../../config/management/genesis", version = "0.1.0", features = ["testing"] }
+libra-global-constants = { path = "../../config/global-constants", version = "0.1.0" }
 libra-logger = { path = "../../common/logger", version = "0.1.0" }
+libra-management = { path = "../../config/management", version = "0.1.0", features = ["testing"] }
+libra-network-address = { path = "../../network/network-address", version = "0.1.0" }
+libra-operational-tool = {path = "../../config/management/operational", version = "0.1.0", features = ["testing"] }
+libra-secure-storage = { path = "../../secure/storage", version = "0.1.0", features = ["testing"] }
 libra-trace = {path = "../../common/trace", version = "0.1.0"}
 libra-types = { path = "../../types", version = "0.1.0", features = ["fuzzing"] }
 libra-workspace-hack = { path = "../../common/workspace-hack", version = "0.1.0" }
--- a/testsuite/cluster-test/src/cluster_builder.rs
+++ b/testsuite/cluster-test/src/cluster_builder.rs
@@ -4,7 +4,11 @@
 use crate::{
     aws,
     cluster::Cluster,
-    cluster_swarm::{cluster_swarm_kube::ClusterSwarmKube, ClusterSwarm},
+    cluster_swarm::{
+        cluster_swarm_kube::{ClusterSwarmKube, KubeNode},
+        ClusterSwarm,
+    },
+    genesis_helper::GenesisHelper,
     instance::{
         fullnode_pod_name, lsr_pod_name, validator_pod_name, vault_pod_name,
         ApplicationConfig::{Fullnode, Validator, Vault, LSR},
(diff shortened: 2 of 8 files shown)


testsuite/cluster-test/src/cluster_builder.rs
```diff
@@ -191,13 +217,45 @@ impl ClusterBuilder {
...
+
+        let fullnode_nodes = try_join_all((0..num_validators).flat_map(move |validator_index| {
+            (0..num_fullnodes_per_validator).map(move |fullnode_index| async move {
+                let pod_name = fullnode_pod_name(validator_index, fullnode_index);
+                self.cluster_swarm.allocate_node(&pod_name).await
+            })
+        }))
+        .await?;
+
+        if !vault_nodes.is_empty() {
```
andll, line 241: Do you mind adding log before/after this code block, like 'generating genesis with management tool', 'done generating genesis'

testsuite/cluster-test/src/cluster_builder.rs
```diff
@@ -272,11 +323,174 @@ impl ClusterBuilder {
...
+                None,
+                None,
+            ));
+            if validator_index == 0 {
+                vault_storage
+                    .create_key(format!("{}__{}", LIBRA_ROOT_NAME, LIBRA_ROOT_KEY).as_str())?;
+            }
+            vault_storage.create_key(
+                format!("{}__{}", validator_pod_name(validator_index), OWNER_KEY).as_str(),
+            )?;
```
andll, line 346: Do you mind adding `.map_err` with description of failed place in all places in this function where you use `?`

This seem to be useful pattern when debugging what is going wrong

testsuite/cluster-test/src/cluster_builder.rs
```diff
@@ -272,11 +323,174 @@ impl ClusterBuilder {
...
+            )
+            .await?;
+
+        for (i, node) in vault_nodes.iter().enumerate() {
+            genesis_helper
+                .owner_key(
+                    VAULT_BACKEND,
+                    format!("http://{}:{}", node.internal_ip, VAULT_PORT).as_str(),
+                    token_path,
+                    validator_pod_name(i as u32).as_str(),
```
snakech, line 413: u can define a pod name and use it here and there

testsuite/cluster-test/src/cluster_builder.rs
```diff
@@ -272,11 +323,174 @@ impl ClusterBuilder {
...
+                .set_operator(
+                    validator_pod_name(i as u32).as_str(),
+                    validator_pod_name(i as u32).as_str(),
+                )
+                .await?;
+        }
+        genesis_helper
+            .genesis(ChainId::new(1), Path::new(GENESIS_PATH))
+            .await?;
+        for (i, node) in vault_nodes.iter().enumerate() {
```
snakech, line 455: do we iterate in same loop on purpose?

changyonghu: Yes, genesis_help.genesis() should be called in between these two loops.

testsuite/cluster-test/src/cluster_builder.rs
```diff
@@ -272,11 +322,174 @@ impl ClusterBuilder {
...
+    async fn initialize_vault(&self, validator_index: u32, vault_node: &KubeNode) -> Result<()> {
+        let addr = vault_node.internal_ip.clone();
+        tokio::task::spawn_blocking(move || {
+            let mut vault_storage = Box::new(VaultStorage::new(
+                format!("http://{}:{}", addr, VAULT_PORT),
+                VAULT_TOKEN.to_string(),
+                None,
+                None,
+            ));
+            if validator_index == 0 {
```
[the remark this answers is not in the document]
changyonghu: Good point. Save a lot lines.

testsuite/cluster-test/src/cluster_swarm/cluster_swarm_kube.rs
```diff
@@ -751,6 +719,38 @@ impl ClusterSwarm for ClusterSwarmKube {
...
+        &self,
+        node: &str,
+        pod_name: &str,
+        path: &str,
+        content: Vec<u8>,
+    ) -> Result<()> {
+        let bucket = "toro-cluster-test-flamegraphs";
+        let run_id = env::var("RUN_ID").expect("RUN_ID is not set.");
+        self.s3_client
+            .put_object(PutObjectRequest {
```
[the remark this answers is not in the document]
changyonghu: Rebased.

testsuite/cluster-test/src/cluster_swarm/cluster_swarm_kube.rs
```diff
@@ -760,6 +719,47 @@ impl ClusterSwarm for ClusterSwarmKube {
...
+        &self,
+        node: &str,
+        pod_name: &str,
+        path: &str,
+        content: Vec<u8>,
+    ) -> Result<()> {
+        let bucket = "toro-cluster-test-flamegraphs";
+        let run_id = env::var("RUN_ID").expect("RUN_ID is not set.");
+        libra_retrier::retry_async(libra_retrier::fixed_retry_strategy(5000, 15), || {
+            let run_id = &run_id;
```
snakech, line 733: u do not need a local variable for run_id

testsuite/cluster-test/Cargo.toml
```diff
@@ -38,6 +39,12 @@ config-builder = { path = "../../config/config-builder", version = "0.1.0" }
 generate-key = { path = "../../config/generate-key", version = "0.1.0" }
 libra-crypto = { path = "../../crypto/crypto", version = "0.1.0" }
 libra-config = { path = "../../config", version = "0.1.0" }
+libra-genesis-tool = { path = "../../config/management/genesis", version = "0.1.0", features = ["testing"] }
```
snakech, line 42: do you mind make this by alphabetical order?

testsuite/cluster-test/src/cluster_swarm/cluster_swarm_kube.rs
changyonghu: What do you mean? I think this is intentional to make &run_id to move.

snakech:
btw, this pr is still a draft, could you change it back?

testsuite/cluster-test/src/cluster_swarm/cluster_swarm_kube.rs
snakech: u can either use clone()? but just styling preference, i think its ok.

testsuite/cluster-test/src/cluster_swarm/cluster_swarm_kube.rs
```diff
@@ -760,6 +719,47 @@ impl ClusterSwarm for ClusterSwarmKube {
...
     }
+
+    async fn put_file(
+        &self,
+        node: &str,
+        pod_name: &str,
+        path: &str,
+        content: Vec<u8>,
+    ) -> Result<()> {
+        let bucket = "toro-cluster-test-flamegraphs";
```
snakech, line 730: I found this bucket name has been used somewhere else, I think u may can create a followup task/PR to make this point into config or global scope, which can be shared and easy access

changyonghu: Yes, will do. We also need to change the name.

snakech:
LGTM, please address the comments

snakech:
do you have any test plans?

changyonghu:
> do you have any test plans?

./scripts/cti -E RUST_LOG=debug --pr 5425 --run bench --enable-lsr --lsr-backend vault

changyonghu:
> This is a bit unusual, these are Rust APIs, but you're using the CLI parser to access the commands. This is very fragile and not a very good pattern because it will fail at runtime and not compile time. We should update this to directly call into the APIs.
> 
> @andll @snakech

Gotta. I will look at how smoke test works and refactor this PR. Thanks for pointing out.

snakech:
by then, please revert this change first I reckon

PR #5428, opened 2020-08-01
[CI] build all top level crates

sausagee:
## Motivation
Add a check to build individual top level crates to catch hidden dep error.

## Test Plan
CI
Canary 1eaa713 caught issues in
- `executor-test-helpers`  https://app.circleci.com/pipelines/github/libra/libra/26359/workflows/6c22884e-10d8-4f54-b35d-d13d92c1243f/jobs/170986/parallel-runs/2?filterBy=ALL
- `transaction-builder-generated` https://app.circleci.com/pipelines/github/libra/libra/26359/workflows/6c22884e-10d8-4f54-b35d-d13d92c1243f/jobs/170986/parallel-runs/3?filterBy=ALL

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -339,6 +339,34 @@ jobs:
             RUST_BACKTRACE=1 $CARGO $CARGOFLAGS build -j 16 -p transaction-builder -p move-vm-types --target powerpc-unknown-linux-gnu
       - save_sccache
       - build_teardown
+  build-individual-crates:
+    executor: build-executor
+    parallelism: 7
+    description: Build all top level individual crates
+    steps:
+      - build_setup
+      - restore_cargo_package_cache
+      - install_sccache
+      - restore_sccache
+      - run:
+          name: Determine crate targets for this container.
+          command: |
+            for x in $(find . -name Cargo.toml); do dirname $x >> /tmp/crates; done
+            echo "Found $(wc -l /tmp/crates) crates"
+            cat /tmp/crates | circleci tests split > /tmp/crates_to_build
+            echo -e "This container will build these crates\n$(cat /tmp/crates_to_build)"
+      - run:
+          name: Build assigned crates in this container.
+          command: |
+            set -x
+            _pwd=$PWD
+            for crate in $(cat /tmp/crates_to_build) ; do
+              cd $_pwd/$crate
+              RUST_BACKTRACE=1 $CARGO $CARGOFLAGS build
+              RUST_BACKTRACE=1 $CARGO $CARGOFLAGS test --no-run
+            done
+      - save_sccache
+      - build_teardown
   run-e2e-test:
     executor: build-executor
     parallelism: 2
@@ -706,6 +734,17 @@ workflows:
                 - /^premainnet-0.[\d|.]+$/
                 - /^v[\d|.]+-release$/
       - build-dev:
+          requires:
+            - prefetch-crates
+          filters:
+            branches:
+              ignore:
+                - gh-pages
+                - master
+                - /^testflow[\d|.]+$/
+                - /^premainnet-[\d|.]+$/
+                - /^v[\d|.]+-release$/
+      - build-individual-crates:
           requires:
             - prefetch-crates
           filters:


bmwill:
Note that the issue that we had today wouldn't have been caught by this. We'd need to build all test targets individually as well :(

sausagee:
Ouch... this is going to rack up lots of minutes per PR. It will certainly make someone watching the meter very happy. We will target determinator to scale.  

Let me do another iteration to see how much resource it takes.

sausagee:
> Ouch... this is going to rack up lots of minutes per PR. It will certainly make someone watching the meter very happy. We will target determinator to scale.
> 
> Let me do another iteration to see how much resource it takes.

It turns to be an expensive proposition - ~25min build time over 7 parallel containers.

.circleci/config.yml
```diff
@@ -716,6 +744,17 @@ workflows:
                 - /^testflow[\d|.]+$/
                 - /^premainnet-[\d|.]+$/
                 - /^v[\d|.]+-release$/
+      - build-individual-crates:
+          requires:
+            - prefetch-crates
+          filters:
+            branches:
+              ignore:
+                - gh-pages
```
rexhoffman, line 753: unfortunately this "gh-pages" doesn't matter.   Since this file will never be in gh-pages branch the config will never be honored.   The one time I tried putting a minimal ignore-this-branch circle ci config in gh-pages it was immediately overwritten.   So line 753 will do no good.  It will also do no harm aside from cause a little confusion.

.circleci/config.yml
```diff
@@ -339,6 +339,34 @@ jobs:
...
+    description: Build all top level individual crates
+    steps:
+      - build_setup
+      - restore_cargo_package_cache
+      - install_sccache
+      - restore_sccache
+      - run:
+          name: Determine crate targets for this container.
+          command: |
+            for x in $(find . -name Cargo.toml); do dirname $x >> /tmp/crates; done
```
rexhoffman, line 354: we really do need an xbuild that can build all top level members for us, as a one shot or separately.   Something like cargo xbuild --no-feature-unification  p1 p2 p3. Where each one would be built independently.   I'll add this to the xbuild card.

.circleci/config.yml
```diff
@@ -339,6 +339,34 @@ jobs:
...
+            echo -e "This container will build these crates\n$(cat /tmp/crates_to_build)"
+      - run:
+          name: Build assigned crates in this container.
+          command: |
+            set -x
+            _pwd=$PWD
+            for crate in $(cat /tmp/crates_to_build) ; do
+              cd $_pwd/$crate
+              RUST_BACKTRACE=1 $CARGO $CARGOFLAGS build
+              RUST_BACKTRACE=1 $CARGO $CARGOFLAGS test --no-run
```
rexhoffman, line 366: looks like xtest supports no run, if you use it here it's a good entry point for global behavior change on tests.

rexhoffman:
This type of individual test/build is done daily by the code coverage run btw.   And failed crates (test or compilation) are printed as a list in the code-coverage slack channel daily.   I occasionally clean them up.   That may be good enough?

.circleci/config.yml
sausagee: I'd wager a target determinator - through the dependency graph (such as guppy), it can determine which target(s) need to re-build.  

Maybe that's the next big thing to tackle? cc @metajack

.circleci/config.yml
sausagee: Will update to xtest.

sausagee:
> This type of individual test/build is done daily by the code coverage run btw. And failed crates (test or compilation) are printed as a list in the code-coverage slack channel daily. I occasionally clean them up. That may be good enough?
> 
> Only handles the test, not the executables, though.

Yep, I noticed that too. I was thinking to myself if this tradeoff (doing this per PR) is the right call ATM.  Or we should wait for things like target determinator.  I want to hear what folks think of the tradeoff before closing this PR.

metajack:
I think a nightly run is fine, as long as someone is filing bugs as a result, or we can file the issues or notify slack automatically so they don't get ignored.

Target determinator will probably reduce build times enough we could do this per-PR with minimal cost.

PR #5429, opened 2020-08-01
[rfc][language][vm] implement Drop for public value types

vgao1996:
## Summary
Currently we don't have a good way to prevent one from writing a Move program creating a deeply nested value, which makes recursive operations on them susceptible to stack overflows. There are two recursive operations as of now:
1. Serialization/Deserialization -- already protected by the limit on layout depth
2. Compiler generated Drop

This PR suggests implementing a safe Drop for all public value types, making it impossible for clients to accidentally run into this issue.

## Test Plan
TBD


language/ir-testsuite/tests/effects/publish_then_unpublish.mvir
```diff
@@ -0,0 +1,27 @@
+module M {
```
sblackshear, line 1: Can these be written as source lang tests? I think we should avoid adding new IR tests unless we can't avoid it...

language/move-vm/types/src/data_store.rs
```diff
@@ -19,27 +19,12 @@ pub trait DataStore {
...
-    /// Get a reference to a resource.
-    fn borrow_resource(
-        &mut self,
-        addr: AccountAddress,
-        ty: &Type,
-    ) -> PartialVMResult<Option<&GlobalValue>>;
-
-    /// Transfer ownership of a resource to the VM.
-    fn move_resource_from(
+    /// TBA
```
sblackshear, line 22: Comment needed?

language/ir-testsuite/tests/effects/publish_then_unpublish.mvir
vgao1996: Other tests of the same class are written in IR so I'd decided to keep it that way in this PR. How about I put up another PR to migrate them all at once?

language/move-vm/types/src/data_store.rs
vgao1996: Ah good catch. Will fix it in the base commit.

davidiw:
Closing out old PR, feel free to re-open.

PR #5430, opened 2020-08-01
[state sync] add support for multicasting chunk request

christinacelee:
## Motivation

Add  multicasting functionality in state sync
Currently there is the option to failover to the next available upstream network if there are no active peers in the primary upstream network, i.e. based on the liveness of peers in a network
This PR extends the failover functionality based on no progress made by requesting chunks from a given network. (This is possible if all peers in the network are behind, possibly due to trying to state sync and catch up itself) 
If no progress is made by a certain configurable timeout duration, state sync next sends the same  chunk request to more networks, if available. This "multicasting" continues until a primary upstream peer gives a successful chunk response first, and afterwards state sync resumes only requesting chunks from the primary network

Also refactored peer manager into request manager to handle logic for sending chunk requests

## Test Plan

New integration test


zekun000:
not related to the change itself, but the logic is getting more and more complicated and exceeds my ability to reason about. I hope we can have a more organized structure both for reviewing and testing.

for example, the peer manager can evolve to handle all the logic of how/who to send the requests, and maybe other actors and let coordinator only does coordination.

state-synchronizer/src/request_manager.rs
```diff
@@ -0,0 +1,436 @@
...
+            // if chunk from a primary upstream is successful, stop multicasting the request to failover networks
+            self.multicast_level = MIN_UPSTREAM_NETWORK_CT;
+        }
+
+        // update score
+        self.update_score(peer, PeerScoreUpdateType::Success);
+    }
+
+    // penalize peer's score for giving chunk with starting version that doesn't match local synced version
+    pub fn process_bad_chunk_start(
```
zekun000, line 308: this name is .. confusing to me

PR #5431, opened 2020-08-01
[Management Operations] Support failure recovery mechanisms for key rotations.

JoshLind:
## Motivation

This PR adds support to the management operations tool to allow the tool to handle key rotation failures (e.g., during previous rotation attempts). To do this, whenever an operator tries to rotate a key (e.g., the operator key or the consensus key), the tool first checks that the on-chain key state matches the state in storage. If so, a rotation is performed. If not, the tool simply regenerates a rotation transaction to update the state on chain.

To achieve this, this PR offers 3 commits:

1. First, we add support to handle key rotation failures relating to validator config based keys. This commit also updates the relevant smoke tests to check this.
2. Next, we expose a new method to secure storage: get_public_key_previous_version(), which returns the previous version of any crypto key. For this, we add an appropriate test to storage.
3. Finally, we add support to handle operator key rotation failures, and add an appropriate smoke test.

Yes.

## Test Plan

All local tests pass, including the newly added smoke tests and storage tests. Note, for the storage tests, I also ran the vault tests to make sure that things have been implemented correctly on vault.

## Related PRs

None.

### DIFF
--- a/config/management/operational/src/operator_key.rs
+++ b/config/management/operational/src/operator_key.rs
@@ -7,6 +7,7 @@ use libra_global_constants::{OPERATOR_ACCOUNT, OPERATOR_KEY};
 use libra_management::{constants, error::Error};
 use libra_secure_time::{RealTimeService, TimeService};
 use libra_types::transaction::{authenticator::AuthenticationKey, RawTransaction, Transaction};
+use std::convert::TryFrom;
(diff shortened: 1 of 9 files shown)


JoshLind:
Awesome! Thanks, David! 😄 @bors-libra r=davidiw

PR #5433, opened 2020-08-01
[cli] Accept full URL in --faucet param

mgorven:
The new k8s testnet exposes the faucet over HTTPS, but the CLI currently
enforces HTTP. Change the `-f` param to accept a full URL so HTTPS can
be specified. If this isn't provided use the current behaviour of
converting with `s/client/faucet/` and forcing HTTP.

Test Plan:

Against devnet:

    target/debug/cli -u https://client.dev.aws/ --waypoint 0:3061d4079c277b98c612e210b055c088264e5ffe17c36a616d5dc0f891a19fac --chain-id DEVNET

Against k8s testnet:

    target/debug/cli -u https://mgorvennet.k8s.aws/ -f https://mgorvennet.k8s.aws/mint --waypoint_url https://mgorvennet.k8s.aws/waypoint.txt --chain-id 13


testsuite/cli/src/client_proxy.rs
```diff
@@ -172,12 +172,15 @@ impl ClientProxy {
             Some(dd_account_data)
         };
 
-        let faucet_server = match faucet_server {
-            Some(server) => server,
-            None => url
-                .host_str()
-                .ok_or_else(|| format_err!("Missing host in URL"))?
-                .replace("client", "faucet"),
+        let faucet_url = match faucet_url {
```
thefallentree, line 175:
```
facuet_url.map_or( Url::parse(format!(...)),  |x| Url::parse(x)).ok_or_else (|| fomrat_err!(""))?; 
```

mgorven: Couldn't figure this out exactly, did something slightly different with `unwrap_or()`.

PR #5440, opened 2020-08-03
[gha][lbt] fix repro command extra newlines

rustielin:
## Motivation

Repro cmd from LBT gave extra newlines, e.g. https://github.com/libra/libra/pull/5385#issuecomment-668154024
This is because javascript template literals preserve newlines. So now, save command string, `eval` it, and then use that env var to print the repro cmd in PR comments. NOTE: Because it's a bit tricky to have multi-line strings in env vars, just writing it to a single line for now.

After this PR, this...

      ./scripts/cti --tag land_c22fd5ce --cluster-test-tag land_ffd2140a \ 

          -E BATCH_SIZE=15 -E UPDATE_TO_TAG=land_ffd2140a -E DISABLE_FN_UPGRADE=1 \ 

          --suite land_blocking_compat

turns into...

```
./scripts/cti --tag land_c22fd5ce --cluster-test-tag land_ffd2140a -E BATCH_SIZE=15 -E UPDATE_TO_TAG=land_ffd2140a -E DISABLE_FN_UPGRADE=1 --suite land_blocking_compat
```

(Write your answer here.)

## Test Plan

## Related PRs

### DIFF
--- a/.github/workflows/land-blocking.yml
+++ b/.github/workflows/land-blocking.yml
@@ -76,24 +76,16 @@ jobs:
           date
           export CTI_OUTPUT_LOG=$(mktemp)
           echo "::set-env name=CTI_OUTPUT_LOG::$CTI_OUTPUT_LOG"
+          cmd=""
           if [ $TEST_COMPAT -eq 1 ]; then
-            ./scripts/cti \
-              --tag ${PREV_TAG} \
-              --cluster-test-tag ${TEST_TAG} \
-              -E RUST_LOG=debug \
-              -E BATCH_SIZE=15 \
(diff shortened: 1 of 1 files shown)


andll:
Should we fix the root cause of the problem?

Right now command then runs cluster test and repro cmd are created in different places.
Is there way we can just define something like CMD="xxx" in one place, and then use `eval $CMD` when we want to call it, and print `$CMD` in repro cmd?

rustielin:
I couldn't figure out a prettier way to do this.. Repro cmd will just be a single line. But this should be more consistent now that we're just defining a cmd string, and using `eval` on it.

PR #5444, opened 2020-08-03
[state sync] fix target and local version comparison

christinacelee:
Target LI should be rejected if it is also equal to local version


christinacelee:
> does this break anything you're testing?

no, will add test

zekun000:
bors can even react with emoji? exceed expectation!

bmwill:
> bors can even react with emoji? exceed expectation!

@zekun000 I might save this for PSC next half 😉

Issue #3713, opened 2020-05-04
[Bug] "cargo run -p libra-swarm -- -s" Unable to build all workspace binaries.


AntoineMkr:
# 🐛 Bug

After installing Libra without any problems, I wanted to launch a local node to play with.
But I encountered that error:
**Stack trace/error message**
```
thread 'main' panicked at '
    Unable to build all workspace binaries. Cannot continue running tests.

    Try running 'cargo build --all --bins --exclude cluster-test' yourself.
', <::std::macros::panic macros>:2:4
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
logs located at "/tmp/7e2ae50ceb175cc22182c4863eeecf7a"
```
So I entered `cargo build --all --bins --exclude cluster-test`
That command made my computer very laggy, and it took a while to achieve. The output is:
```
# Compiling 

error: linking with `cc` failed: exit code: 1
  = note: "cc" "-Wl,--as-needed" "-Wl,-z,noexecstack" "-m64" "-L" 

# followed by all the deps pathes like 
"/home/USERNAME/.rustup/toolchains/1.43.0-x86_64-unknown-linux-gnu/lib/rustlib/x86_64-unknown-linux-gnu/lib" 
"/home/USERNAME/libra/target/debug/deps/investigate-73900e6e17357af2.1joi3ianzc265m37.rcgu.o" "/home/USERNAME/libra/target/debug/deps/investigate-73900e6e17357af2.1nz7oau12r5bgl01.rcgu.o" 
# and so on... It does a lot of pathes

"-Wl,-Bdynamic" "-lstdc++" "-lutil" "-ldl" "-lutil" "-ldl" "-lrt" "-lpthread" "-lgcc_s" "-lc" "-lm" "-lrt" "-lpthread" "-lutil" "-lutil"
  = note: collect2: fatal error: ld terminated with signal 9 [Processus arrêté]
          compilation terminated.
error: aborting due to previous error

error: could not compile `libra-fuzzer`.

To learn more, run the command again with --verbose.
warning: build failed, waiting for other jobs to finish...
# tries to compile the next packages, returns the same error 
error: aborting due to previous error

error: could not compile `libra-fuzz`.
To learn more, run the command again with --verbose.
warning: build failed, waiting for other jobs to finish...
error: build failed
```
## To reproduce

**Code snippet to reproduce**
```rust
# I followed the tutorial
git clone https://github.com/libra/libra.git && cd libra
git checkout testnet
[...]

Issue #1118, opened 2019-09-30
[libra-swarm] enable libra-swarm to be usable outside of the repository


bmwill:
Right now the library and binary `libra-swarm` is only able to be launch inside of the libra repository. The reason for this is that it currently relies on being able to invoke cargo to ensure that all binaries are built before launching a number of `libra-node` processes. It might be nice to instead allow `libra-swarm` to be runnable outside of a repository if the user has `cargo install`ed both `libra-swarm` as well as the `libra-node` binary.

PR #5448, opened 2020-08-03
swarm: don't build workspace binaries 

bmwill:
Change `libra-swarm` to take in a path to the `libra-node` binary instead of assuming that we're inside of the repository and invoking `cargo build` to build the `libra-node` binary.

This does a couple of things:
* Pushes the building of all workspace binaries up making it the responsibility of the smoke_tests
* Enables `libra-swarm` to be invoked outside of the repository. This makes it much easier for folks to do a `cargo install` of both `libra-swarm` and `libra-node` which can then be run independent of a libra repository.

Fixes: #1118
Fixes: #3713


testsuite/libra-swarm/src/swarm.rs
```diff
@@ -294,12 +286,11 @@ impl LibraSwarm {
     }
 
     pub fn configure_validator_swarm(
+        libra_node_bin_path: &Path,
         num_nodes: usize,
         config_dir: Option<String>,
         template: Option<NodeConfig>,
     ) -> Result<LibraSwarm> {
-        LibraNode::prepare();
-
```
davidiw, line 302: By eliminating prepare, we increase the likelihood that our ephemeral ports are no longer reserved between the time the config has been generated, compilation, and then execution of the nodes. Thoughts?

testsuite/libra-swarm/src/main.rs
```diff
@@ -74,7 +87,7 @@ fn main() {
     println!("To run the Libra CLI client in a separate process and connect to the validator nodes you just spawned, use this command:");
 
     println!(
-        "\tcargo run --bin cli -- -u {} -m {:?} --waypoint {} --chain-id {:?}",
+        "\tcli -u {} -m {:?} --waypoint {} --chain-id {:?}",
```
davidiw, line 90: Could you elaborate on this and the other changes that remove "cargo run". I'm guessing because you imagine that these would be installed?

testsuite/libra-swarm/src/swarm.rs
bmwill: Actually, remove this and pushing to ensure that the binaries are build before calling into the swarm lib should make it less error prone to run into this specific issue

testsuite/libra-swarm/src/main.rs
bmwill: The idea should be that swarm is a binary/library that can be used to orchestrate local instances of a network. Theres been a ton of interest in the past for allowing for this to be run outside of the repo and because of that you can't tell people to run cargo (this does bring up the point that our binaries have terrible names 😄 )

bmwill:
The point of this PR is to eliminate workspace-builder completely from the swarm library and binary. If not, then the goal of being able to run outside of the libra repo (with potentially installed libra-node binary) cannot be achieved. I'm thinking that if we really want to continue to have the ability to build workspace-binaries and then launch libra swarm that we could build an `cargo xswarm` entry point to do that.

testsuite/tests/libratest/smoke_test.rs
```diff
@@ -60,8 +60,13 @@ impl TestEnvironment {
         let mut template = NodeConfig::default_for_validator();
         template.state_sync.chunk_limit = chunk_limit;
 
-        let validator_swarm =
-            LibraSwarm::configure_validator_swarm(num_validators, None, Some(template)).unwrap();
+        let validator_swarm = LibraSwarm::configure_validator_swarm(
+            &workspace_builder::get_libra_node_with_inject_error(),
```
davidiw, line 64: seems I misread something, so we generate the builder here way before generating the configs

testsuite/libra-swarm/src/main.rs
davidiw: I'm a little curious about this one... I wish we had metrics on what people are using and building :). I've been able to reason about most of my concerns. Could we update search for and update the readmes for swarm? I think there's some on the website.

testsuite/tests/libratest/smoke_test.rs
bmwill: Yeah so before `prepare` was called first thing when we enter `configure_validaotr_swarm`, now we're forcing users to provide us with a path to the binary so `prepare` becomes redundant because the binaries are already all built.

testsuite/libra-swarm/src/main.rs
bmwill: Yeah...thinking about this a little more maybe there's a bit more that should be done before this lands. I'm thinking maybe at the very least providing a wrapper around `libra-swarm` that has the exact same behavior we have today. Let me see what I can do in the next couple of days and I'll have you take 1 more look at it then

bmwill:
/land

Double checked with @davidiw offline that this was still good to land

PR #5453, opened 2020-08-03
[network] log error when failing to dial peers

gregnazario:
Added more detail for logging errors so that we can keep track of what
actually fails when dialing peers.

We know that this error message is actually showing up, so let's make it be useful with more detail, as was feedback from the PEs.

### DIFF
--- a/network/src/peer_manager/mod.rs
+++ b/network/src/peer_manager/mod.rs
@@ -985,10 +985,11 @@ where
             }
             Err(error) => {
                 error!(
-                    "{} Error dialing Peer {} at {}",
+                    "{} Error dialing Peer {} at {}: {}",
                     self.network_context,
                     peer_id.short_str(),
-                    addr
+                    addr,
+                    error
                 );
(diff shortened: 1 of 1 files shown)


network/src/peer_manager/mod.rs
```diff
@@ -985,10 +985,11 @@ where
             }
             Err(error) => {
                 error!(
-                    "{} Error dialing Peer {} at {}",
+                    "{} Error dialing Peer {} at {}: {}",
                     self.network_context,
                     peer_id.short_str(),
-                    addr
+                    addr,
+                    error.to_string()
```
phlip9, line 992: i don't think we need the `.to_string()`, since `.to_string()` just uses the Display impl anyway, so we can remove it

can also avoid an extra allocation : p

gregnazario: putting an error in the log, will make it both do "display" and "debug" in the structured logs :(

PR #5454, opened 2020-08-03
[executor] replace panic with Err

mimoo:


lightmark:
What's wrong with `expect`? `first_version` must exist.

mimoo:
why crash if we can return an error? Here's my reasoning:

This might not be a problem now, but someone could make a change to the code, or add a new way to call this function. Right now we make the assumption that `first_version` must exist, but since we return a `Result` type we do not need to make that assumption. Usually the less assumption we make and the more robust the code will be : o (it's hard to keep track of many assumptions and ensure they remain true as we change the code)

execution/executor/src/lib.rs
```diff
@@ -179,10 +179,18 @@ where
...
-        let first_txn_version = txn_list_with_proof
-            .first_transaction_version
-            .expect("first_transaction_version should exist.")
-            as Version;
+        let first_txn_version = match txn_list_with_proof.first_transaction_version {
+            Some(tx) => tx as Version,
+            None => {
+                send_struct_log!(StructuredLogEntry::new_named(
+                    "MUST_FIX",
+                    "first_transaction_version should exist."
```
mimoo, line 187: @andll can I use space in a name?

andll: You can, in sense that it would not fail
But I think ideally I would prefer not to, just because it looks strange. Name is not a message

cc @gregnazario

lightmark:
My question is:
We already verified the chunk is valid in step 1. So here it cannot fail, So I don't understand why we need to change to `Err`.

execution/executor/src/lib.rs
gregnazario: So, what you want here is a critical message, with a message.

tbh, you should just use text logging.  I'm looking into ways to do what you want, which is have a log message, but use the structured log API directly.

```
crit!("first_transaction_version should exist")
```

andll: I am not sure we should recommend people using text api. If anything should probably converge everything on one logging macro? (struct log?)

> Especially since there isn't any extra info being added

What if tomorrow they decide to add some field for debugging?

gregnazario: I'm assuming because this is `MUST_FIX` that it wouldn't be here for long, so I was just proposing a simple solution.

If you want @mimoo you could do the following in the time being:
```
StructuredLogEntry::new_named(
  "MUST_FIX",
  "assertion")
.data("details", "first_transaction_version should exist.")
```

That fits the current paradigm, but I recognize that it isn't great and I'm thinking of ways to improve it.

PR #5455, opened 2020-08-03
[breaking][libra framework] allow opt-in dual attestation check for DD sender/r…

sblackshear:
…ecipients

Requirements have changed and we no longer need to require dual attestation for DD's.

This will also unblock cluster-test, which currently has some issues with amount-restricted minting because of DD dual attestation. Replacing uses of `encode_testnet_mint` in cluster-test with `encode_peer_to_peer_with_metadata_script` should allow big mints.


language/e2e-tests/src/tests/transaction_builder.rs
```diff
@@ -624,14 +565,47 @@ fn dual_attestation_payment() {
...
 
+fn create_dual_attestation_payment(
+    sender_address: AccountAddress,
+    receiver_address: AccountAddress,
+    amount: u64,
+    coin_type: TypeTag,
+    ref_id: Vec<u8>,
+    receiver_compliance_private_key: &Ed25519PrivateKey,
+) -> Script {
+    // UTF8-encoded string "@@$$LIBRA_ATTEST$$@@" without length prefix
```
bob-wilson, line 576: Please change this comment to say "ASCII-encoded". (I thought I had removed all the references to UTF-8 earlier, but I must have missed some or they were added back.) Or, better yet, maybe you could write the domain_separator as a byte string literal? e.g., b"@@$$LIBRA_ATTEST$$@@".to_vec()
Then there would be no need for the comment at all.

language/stdlib/modules/DualAttestation.move
```diff
@@ -201,23 +200,9 @@ module DualAttestation {
...
-        let is_payee_vasp = VASP::is_vasp(payee);
-        let is_payer_dd = DesignatedDealer::exists_at(payer);
-        let is_payee_dd = DesignatedDealer::exists_at(payee);
-        let is_inter_vasp =
-            is_payer_vasp && is_payee_vasp &&
-            VASP::parent_address(payer) != VASP::parent_address(payee);
-        is_inter_vasp || // (1) inter-VASP
-            (is_payer_dd && is_payee_dd) || // (2) inter-DD
-            (is_payer_vasp && is_payee_dd) || // (3) VASP -> DD
-            (is_payer_dd && is_payee_vasp) // (4) DD -> VASP
```
tzakian, line 220: <3

testsuite/cluster-test/src/tx_emitter.rs
```diff
@@ -615,10 +615,12 @@ fn gen_submit_transaction_request(
 fn gen_mint_request(faucet_account: &mut AccountData, num_coins: u64) -> SignedTransaction {
     let receiver = faucet_account.address;
     gen_submit_transaction_request(
-        transaction_builder::encode_testnet_mint_script(
+        transaction_builder::encode_peer_to_peer_with_metadata_script(
             account_config::coin1_tag(),
             receiver,
             num_coins,
```
sblackshear, line 621: @snakech: once this lands, you should be able to change `num_coins` to any amount.

ankushagarwal:
Can you update the comment section of `PeerToPeerWithMetadata` in `generated/src/stdlib.rs#L222`

It currently says:
```
    /// The `metadata` and `metadata_signature` parameters are only required if `amount` >=
    /// `DualAttestation::get_cur_microlibra_limit` LBR and `payer` and `payee` are distinct entities
    /// (e.g., different VASPs, or a VASP and a DesignatedDealer).

```

This is no longer true after this PR

PR #5457, opened 2020-08-04
[state sync] make highest_li optional in ProgressiveLI and verify pen…

christinacelee:
…ding LIs

## Motivation

Before, the later LI's being added to pending LI store were not being verified. This PR ensures that only verified LIs are added to the pending LI store. 
Also update `ResponseLedgerInfo::ProgressiveLedgerInfo` to be better verifiable by adding requirement that `highest_li`, if provided, must be the highest LI for the request's known epoch. This is so that whatever `highest_li` in the response is verifiable by the request upon receiving the response. If `highest_li` is in a future epoch, the requester can't verify the `highest_li` until waiting to reach the later epoch

## Test Plan

Existing/updated tests

### DIFF
--- a/state-synchronizer/src/chunk_response.rs
+++ b/state-synchronizer/src/chunk_response.rs
@@ -20,8 +20,9 @@ pub enum ResponseLedgerInfo {
     ProgressiveLedgerInfo {
         // LedgerInfo that the corresponding GetChunkResponse is built relative to.
         target_li: LedgerInfoWithSignatures,
-        // LedgerInfo that is the highest LI of the responder
-        highest_li: LedgerInfoWithSignatures,
+        // LedgerInfo for a version later than that of `target_li`
+        // If `None`, this is the same as `target_li`
+        highest_li: Option<LedgerInfoWithSignatures>,
     },
     /// During the initial catchup upon startup the chunks carry LedgerInfo that is verified
     /// using the local waypoint.
@@ -96,7 +97,7 @@ impl fmt::Display for GetChunkResponse {
             } => format!(
                 "[progressive LI: target LI {}, highest LI {}]",
                 target_li.ledger_info(),
-                highest_li.ledger_info()
+                highest_li.as_ref().unwrap_or(target_li).ledger_info(),
             ),
             ResponseLedgerInfo::LedgerInfoForWaypoint {
                 waypoint_li,
--- a/state-synchronizer/src/coordinator.rs
+++ b/state-synchronizer/src/coordinator.rs
@@ -90,6 +90,7 @@ impl PendingLedgerInfos {
         }
     }
 
+    /// Adds `new_li` to the queue of pending LI's
     fn add_li(&mut self, new_li: LedgerInfoWithSignatures) {
         // update pending_ledgers if new LI is ahead of target LI (in terms of version)
         let target_version = self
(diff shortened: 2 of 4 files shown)


state-synchronizer/src/chunk_response.rs
```diff
@@ -20,8 +20,11 @@ pub enum ResponseLedgerInfo {
     ProgressiveLedgerInfo {
         // LedgerInfo that the corresponding GetChunkResponse is built relative to.
         target_li: LedgerInfoWithSignatures,
-        // LedgerInfo that is the highest LI of the responder
-        highest_li: LedgerInfoWithSignatures,
+        // LedgerInfo that is the highest LI available on the responder
+        // for the epoch of `target_li`
+        // If the responder is in a later epoch compared to the request's `target_li`, `target_li` will be an end-of-epoch LI,
+        // and `highest_li` will be empty
+        highest_li: Option<LedgerInfoWithSignatures>,
```
zekun000, line 27: why not make this the end-of-epoch LI instead of None?

christinacelee: if the corresponding request for this response is for a past epoch, the `target_li` will already be the end-of-epoch LI, so it would be redundant to specify it here. nothing wrong about repeating that end-of-epoch LI here, just wanted to make this message compact

zekun000: well it's already possible for the target_li to be the same as the highest_li if that's the highest in remote? I don't like special case, I'd expect it's either both cases the highest_li is None or redundant

state-synchronizer/src/coordinator.rs
```diff
@@ -635,6 +634,18 @@ impl<T: ExecutorProxyTrait> SyncCoordinator<T> {
             return Ok(());
         }
 
+        // If the request's epoch is in the past, `target_li` will be set to the end-of-epoch LI for that epoch
+        let target_li =
+            self.choose_response_li(request.known_version, request.current_epoch, target_li)?;
+        // Only populate highest_li field if it is different from target_li
+        let highest_li = if target_li.ledger_info().version() < local_version
```
zekun000, line 641: this is where I was confused, isn't this just, or we don't impl Eq for LI?
```
  let highest_li = self.local_state.highest_local_li.clone();
  let highest_li = Some(highest_li).filter(|li| li != target_li);
```

state-synchronizer/src/coordinator.rs
```diff
@@ -777,20 +788,25 @@ impl<T: ExecutorProxyTrait> SyncCoordinator<T> {
...
             } => {
-                ensure!(
-                    target_li.ledger_info().version() <= highest_li.ledger_info().version(),
-                    "Progressive ledger info received target LI {} higher than highest LI {}",
-                    target_li,
-                    highest_li
-                );
-                self.pending_ledger_infos.add_li(highest_li);
-                self.process_response_with_verifiable_li(txn_list_with_proof, target_li)
+                let highest_li = match highest_li {
```
zekun000, line 797: slightly prefer 
```
let highest_li = highest_li.unwrap_or(target_li);
ensure!(...);
```

state-synchronizer/src/coordinator.rs
christinacelee: if `target_li` is an end-of-epoch for a past epoch, this won't work - highest_li should be none, but this will return `Some`

zekun000: ah I'm too bad at reviewing

PR #5459, opened 2020-08-04
[db-bootstrapper] support commit genesis with waypoint at startup

zekun000:
## Motivation

This commit supports loading genesis and commit it if it matches the
waypoint specified so that operators can avoid manually invoking
db-bootstrapper but just update the genesis blob/waypoint.

If the waypoint and genesis blob are set and local version + 1 ==
waypoint.version, the db-bootstrapper will try to commit it.

If it's intended to sync to the waypoint instead, set genesis_file_location to
empty.

Y

## Test Plan

Changed the smoke test to reflect the new flow.

## Related PRs


execution/executor/src/db_bootstrapper.rs
```diff
@@ -24,19 +24,38 @@ use move_core_types::move_resource::MoveResource;
...
+) -> Result<Waypoint> {
     let tree_state = db.reader.get_latest_tree_state()?;
-    if !tree_state.is_empty() {
-        return Ok(None);
+
+    let committer = calculate_genesis::<V>(db, tree_state, genesis_txn)?;
+    Ok(committer.waypoint)
+}
+
+pub fn bootstrap_genesis<V: VMExecutor>(
```
msmouse, line 37: I'd call it `maybe_bootstrap_x()` or `apply_genesis_if_not_yet`, etc. And document the return value.

execution/executor/tests/db_bootstrapper_test.rs
```diff
@@ -81,9 +80,7 @@ fn test_empty_db() {
         .unwrap();
 
     // `bootstrap_db_if_empty()` does nothing on non-empty DB.
```
msmouse, line 82: update comment

execution/executor/tests/db_bootstrapper_test.rs
```diff
@@ -276,10 +274,6 @@ fn test_pre_genesis() {
     let (db, db_rw) = DbReaderWriter::wrap(LibraDB::new_for_test(&tmp_dir));
     restore_state_to_db(&db, accounts_backup, proof, root_hash, PRE_GENESIS_VERSION);
 
-    // DB is not empty, `bootstrap_db_if_empty()` won't apply default genesis txn.
-    assert!(bootstrap_db_if_empty::<LibraVM>(&db_rw, &genesis_txn)
-        .unwrap()
-        .is_none());
```
msmouse, line 282: test your bootstrap_genesis() called with old waypoint returns error on this?

secure/key-manager/src/tests.rs
```diff
@@ -350,8 +350,15 @@ fn setup_node_using_test_mocks() -> Node<MockLibraInterface> {
...
-        .expect("Failed to execute genesis");
+    let waypoint =
+        db_bootstrapper::generate_waypoint::<LibraVM>(&db_rw, get_genesis_txn(config).unwrap())
+            .expect("Failed to execute genesis");
+    db_bootstrapper::bootstrap_genesis::<LibraVM>(
+        &db_rw,
+        get_genesis_txn(config).unwrap(),
+        waypoint,
+    )
+    .unwrap();
```
msmouse, line 361: looks like we need a helper to use in tests

zekun000: I'll try to make a test only function, the cfg(feature=testing) part makes me lazy..

execution/executor/tests/db_bootstrapper_test.rs
zekun000: this is actually not an old waypoint, it's a wrong waypoint, I can assert it returns an error.
both waypoint are at version 0 but with different state.

msmouse: yes that's what I meant.

secure/key-manager/src/tests.rs
msmouse: I need a emoji here. oh this one looks proper:

😝😝😝😝😝😝😝😝😝😝

PR #5461, opened 2020-08-04
[network] add jitter to dials

gregnazario:
We schedule the dials in the future, but we schedule them all one after
another.  Therefore, when they all start up at the same time they
conflict with each other.  Adding jitter should allow them to run in
parallel and be less likely of stepping on top of each other.


andll:
/land

This failure in land blocking is KP

PR #5462, opened 2020-08-04
[json-rpc] better error handling: return invalid param message for param errors

xli:
1. created standalone invalid error test.
2. for converting param into internal type errors, return invalid param error with specific type information instead of service internal error with unclear error message.
3. created some methods response validation test, direct test respond json instead of using JsonRpcClient to parse out, this helps remove dependency on JsonRpcClient and avoid breaking change (all changes should be backward compatible, use same class deserialize the json will not fail the test)
4. removed InvalidArguments & ExceedSizeLimit type in ErrorData, use simple error message instead, these are added recently (after 07/15 release), and they are JsonRpc error data, so removing them should be fine (not a breaking change)

## Test Plan

Unit Test

### DIFF
--- a/json-rpc/src/methods.rs
+++ b/json-rpc/src/methods.rs
@@ -21,7 +21,6 @@ use libra_types::{
     account_config::{from_currency_code_string, libra_root_address, AccountResource},
     account_state::AccountState,
     chain_id::ChainId,
-    event::EventKey,
     ledger_info::LedgerInfoWithSignatures,
     mempool_status::MempoolStatusCode,
     transaction::SignedTransaction,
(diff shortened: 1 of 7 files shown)


json-rpc/src/methods.rs
```diff
@@ -120,12 +121,60 @@ impl JsonRpcRequest {
...
+            JsonRpcError::invalid_param(index, name, type_info)
+        })?)
+    }
 }
 
 /// Submits transaction to full node
 async fn submit(mut service: JsonRpcService, request: JsonRpcRequest) -> Result<()> {
-    let txn_payload: String = serde_json::from_value(request.get_param(0))?;
-    let transaction: SignedTransaction = lcs::from_bytes(&hex::decode(txn_payload)?)?;
+    let transaction: SignedTransaction = request.parse_submit_signed_transaction(0, "data")?;
```
thefallentree, line 176: no need for `:SignedTransaction` here

types/src/event.rs
```diff
@@ -131,6 +131,15 @@ impl TryFrom<&[u8]> for EventKey {
     }
 }
 
+impl TryFrom<String> for EventKey {
```
thefallentree, line 134: Do we actually need this anywhere else? It seems we could easily do it in parse_* function() instead exposing this everywhere else, which might be confusing to other code, since String could also just hold bytes, in our case, it is  a specific format.

json-rpc/src/methods.rs
xli: updated added try_parse_from generic type instead, so keep this

types/src/event.rs
xli: It is not used anywhere yet. I think adding TryFrom<String> is convenient for the EventKey, 
This impl assumes the String is hex-encoded string, which is most likely case when we serialize bytes into string.
AccountAddress has similar func.

I added try_parse_from to combining this TryFrom<String> with deserialize for parsing EventKey, SignedTransaction and AccountAddress

thefallentree: an better way is to create an enum::EventKeyParam that parse and returns EvenKey instead

PR #5463, opened 2020-08-04
[writeset-me][Libra Framework][Move Prover] Revised Move Error Codes

wrwg:
## Breaking Change

*Justification*: Previously, abort error codes in the Libra Framework where fully local to modules and did not share a common error code space. For example, in one module, the error condition that a function needs to have a signer which has the libra root role might have had code 1, but in another module code 2. This made it impossible to write transitive specifications about error codes. Also reasoning about script failures in application code would need to reference implementation details like from which module the error stems.

This change introduces a new error model where we combine module-local codes with stable, global error categories. Those categories resemble generic error situations like often found in OS or Web APIs, like INVALID_STATE, INVALID_ARGUMENT, REQUIRES_PRIVILEGE, etc.

*Breaking nature*: This change is *only* breaking for code which depends on concrete *Move abort codes* from transaction execution, of which there are likely few, if any. The VM status codes are not affected by this change. The change only updates Move code, specs and related tests, but in a compatible way which should not require to reset the chain.

## Details

This PR introduces a new approach to error handling throughout the framework, as previously discussed with @sblackshear, @tzakian, and others.

The general idea of the new approach is to divide errors into a part for the *category* (first 8 bits) and a part for the *reason*. The former is fairly general, small enumeration of codes. The later is a code local to a module which describes the background of the error. Examples:

- `Errors::not_published(ERESOURCE)` creates an error indicating that a specific resource in a module is not published when accessed.
- `Errors::requires_role(ETREASURY_COMPLIANCE)` creates an error indicating the calling account needs to have role of tresury compliance.

The capital constants are typically local to a module, and are interpreted in context of the module where the abort happened, as reported by the VM.

On specification level we write:

- `aborts_if P with Errors:NOT_PUBLISHED` specifies that a function aborts with the given error category.

Note that we currently cannot specify reasons (and perhaps we don't want to). Implementation is not trivial because we need to make the originating module explicit.

In order to keep error reporting consistent, this PR also pulls generic error conditions (like genesis status, presence of core addresses, and roles) into their matching modules. While this may reduce the fidelity of reason-based error reporting a bit, those conditions always act on the current signer(s), of which there are only a few, and require not much context. This leads to a significant increase in uniformity and brevity of code and specs, which (I hope) make it more readable.

As many of the aborts_if specifications had to be rewritten anyway for this change, I also cleaned them up and adjusted for our newest spec language features. Also, many missing specifications have been filled in while adapting the error codes.

Yes

## Test Plan

Existing tests.

## Related PRs

NA


language/stdlib/modules/AccountLimits.move
```diff
@@ -1,6 +1,7 @@
 address 0x1 {
 
 module AccountLimits {
+	use 0x1::Errors;
```
sblackshear, line 4: Nit: indentation

language/stdlib/modules/Authenticator.move
```diff
@@ -4,6 +4,7 @@
 
 address 0x1 {
 module Authenticator {
+	use 0x1::Errors;
```
sblackshear, line 7: Nit: indentation

language/stdlib/modules/Errors.move
```diff
@@ -0,0 +1,75 @@
...
+    /// A function to create an error from from a category and a reason.
+    fun make(category: u8, reason: u64): u64 {
+        (category as u64) + (reason << 8)
+    }
+
+    /// The system is in a state where the performed operation is not allowed. Example: call to a function only allowed
+    /// in genesis.
+    const INVALID_STATE: u8 = 1;
+
+    /// The signer of a transaction has not the expected address for this operation. Example: a call to a function which
```
sblackshear, line 27:
```suggestion
    /// The signer of a transaction does not have the expected address for this operation. Example: a call to a function which
```

language/stdlib/modules/Errors.move
```diff
@@ -0,0 +1,75 @@
...
+    /// requires a signer which is the Libra root.
```
sblackshear, line 28: This might be a confusing example because Libra Root is a role (and thus the reader might wonder "why not use `REQUIRES_ROLE` here?"). Maybe something like "a function that publishes a singleton resource under a particular address" could work.

language/stdlib/modules/DesignatedDealer.move
```diff
@@ -1,6 +1,7 @@
 address 0x1 {
 module DesignatedDealer {
-    use 0x1::Libra::{Self, Libra};
+	use 0x1::Errors;
```
sblackshear, line 3: Nit: indentation

language/stdlib/modules/LibraTimestamp.move
```diff
@@ -23,40 +23,93 @@ module LibraTimestamp {
...
 
-    const EINVALID_SINGLETON_ADDRESS: u64 = 0;
-    const ETIME_NOT_INITIALIZED: u64 = 1;
-    const ENOT_VM: u64 = 2;
-    const EINVALID_TIMESTAMP: u64 = 3;
-    const EGENESIS_ONLY: u64 = 4;
+    const ENOT_GENESIS: u64 = 0;
+    const ENOT_OPERATING: u64 = 1;
+    const ETIMER_RESOURCE: u64 = 2;
+    const ESTATE_MARKER_RESOURCE: u64 = 3;
```
sblackshear, line 29: Unused?

language/stdlib/modules/AccountFreezing.move
```diff
@@ -32,37 +33,41 @@ module AccountFreezing {
         unfrozen_address: address,
     }
 
-    const ENOT_GENESIS: u64 = 0;
-    const EINVALID_SINGLETON_ADDRESS: u64 = 1;
-    const ENOT_ABLE_TO_FREEZE: u64 = 2;
+	const EFREEZE_EVENTS_HOLDER: u64 = 1;
```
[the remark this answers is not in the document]
wrwg: Done.

language/stdlib/modules/AccountLimits.move
wrwg: Done

language/stdlib/modules/Authenticator.move
wrwg: Done

language/stdlib/modules/DesignatedDealer.move
wrwg: Done

language/stdlib/modules/Errors.move
wrwg: Done.

language/stdlib/modules/Errors.move
wrwg: Done

language/stdlib/modules/Errors.move
```diff
@@ -0,0 +1,75 @@
...
+    const REQUIRES_ADDRESS: u8 = 2;
+
+    /// The signer of a transaction has not the expected  role for this operation. Example: a call to a function which
+    /// requires the signer to have the role of treasury compliance.
+    const REQUIRES_ROLE: u8 = 3;
+
+    /// The signer of a transaction does not have a required capability.
+    const REQUIRES_CAPABILITY: u8 = 4;
```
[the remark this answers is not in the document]
wrwg: Done, but I find the use of privilege very confusing, because it largely overlaps with roles after DDs refactoring. There is essentially a 1:1 mapping between roles and "the permanent right to do something". We may need to revisit this later.

language/stdlib/modules/Errors.move
```diff
@@ -0,0 +1,75 @@
...
+
+    /// A resource is required but not published. Example: access to non-existing account limits resource.
+    const NOT_PUBLISHED: u8 = 5;
+
+    /// A resource is attempted to publish which is already published. Example: calling an initialization function
```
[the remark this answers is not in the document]
wrwg: Done

language/stdlib/modules/Errors.move
```diff
@@ -0,0 +1,75 @@
...
+    /// An argument provided to an operation is invalid. Example: a signing key has the wrong format.
+    const INVALID_ARGUMENT: u8 = 7;
+
+    /// A limit on an amount, e.g. a currency, is exceeded. Example: withdrawal of money after account limits window
+    /// is exhausted.
+    const LIMIT_EXCEEDED: u8 = 8;
+
+    /// An operation is unavailable in the current state of the system. It may later become available. Example:
+    /// A capability which has been already extracted is attempted to extract again.
+    const UNAVAILABLE: u8 = 9;
```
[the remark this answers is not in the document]
wrwg: It's a standard error in network protocols and Web APIs. The difference to the ones you are mentioning is that it is temporary, and retry (using backoff strategy or similar) may resolve it. However, there was no case where we could use it, so I nuked it.

language/stdlib/modules/Errors.move
```diff
@@ -0,0 +1,75 @@
...
+
+    /// An internal error (bug) has occurred.
```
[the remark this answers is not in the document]
wrwg: Actually he very may will see it (and asked to send a bug report). We don't use it right now, but I keep it perhaps we need it.

language/stdlib/modules/LibraTimestamp.move
wrwg: Done (after push)

wrwg:
This PR is now ready for review.

tnowacki:
Are we changing anything for the Move functions known by the VM? (prologue/epilogues, bump_sequence_number, etc)

I was surprised to see no changes in `language/libra-vm/src/errors.rs`. Fine if there don't need to be any, just curious :)

language/stdlib/modules/RegisteredCurrencies.move
```diff
@@ -15,34 +15,23 @@ module RegisteredCurrencies {
         currency_codes: vector<vector<u8>>,
     }
 
-    const ENOT_GENESIS: u64 = 0;
     const EINVALID_SINGLETON_ADDRESS: u64 = 1;
```
tzakian, line 18: nit: shift these numbers down by one to make sequential starting at 0

language/stdlib/modules/VASP.move
```diff
@@ -21,22 +22,23 @@ module VASP {
...
 
     /// Maximum number of child accounts that can be created by a single ParentVASP
     const MAX_CHILD_ACCOUNTS: u64 = 256;
 
     public fun initialize(lr_account: &signer) {
-        assert(LibraTimestamp::is_genesis(), ENOT_GENESIS);
-        assert(Roles::has_libra_root_role(lr_account), ENOT_LIBRA_ROOT);
-        assert(Signer::address_of(lr_account) == CoreAddresses::LIBRA_ROOT_ADDRESS(), EINVALID_SINGLETON_ADDRESS);
+        LibraTimestamp::assert_genesis();
+        Roles::assert_libra_root(lr_account);
```
tzakian, line 37: Missing `CoreAddresses::assert_libra_root(lr_account);` here?

tzakian:
@tnowacki 

> Are we changing anything for the Move functions known by the VM? (prologue/epilogues, bump_sequence_number, etc)
I was surprised to see no changes in language/libra-vm/src/errors.rs. Fine if there don't need to be any, just curious :)

This PR left the error code numbers for the prologue and writeset prologue untouched so that it doesn't effect the VM which seems like the right thing to do since the VM doesn't care about error category etc.

tnowacki:
> This PR left the error code numbers for the prologue and writeset prologue untouched so that it doesn't effect the VM which seems like the right thing to do since the VM doesn't care about error category etc.

Okie doke. Not a big deal for me either way. 

If you need/want it for consistency, I'd say go for it. The breaking change is already happening (and we will have more) so it might not be too bad add it on. 
But if y'all don't care, then no need to worry :)

wrwg:
> > This PR left the error code numbers for the prologue and writeset prologue untouched so that it doesn't effect the VM which seems like the right thing to do since the VM doesn't care about error category etc.
> 
> Okie doke. Not a big deal for me either way.
> 
> If you need/want it for consistency, I'd say go for it. The breaking change is already happening (and we will have more) so it might not be too bad add it on.
> But if y'all don't care, then no need to worry :)

For this PR, I wanted to avoid any changes to the VM code, so it can be rolled out as writeset-me only (I only later learned that it actually means it wouldn't be "breaking" in the full sense). But I'm fully on the same page that we want to change this special codes somehow eventually, probably mapping them into some reserved code space. In another PR.

language/stdlib/modules/LibraAccount.move
```diff
@@ -245,15 +272,16 @@ module LibraAccount {
         metadata: vector<u8>,
         metadata_signature: vector<u8>
     ) acquires LibraAccount, Balance, AccountOperationsCapability {
-        assert(!AccountFreezing::account_is_frozen(payee), EACCOUNT_FROZEN);
+        LibraTimestamp::assert_operating();
+        assert(!AccountFreezing::account_is_frozen(payee), Errors::invalid_state(EACCOUNT_FROZEN));
```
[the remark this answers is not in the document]
wrwg: Done. However, in case of the prologue, we still need to do it literally.

language/stdlib/modules/LibraAccount.move
```diff
@@ -364,7 +394,8 @@ module LibraAccount {
         balance: &mut Balance<Token>,
         amount: u64
     ): Libra<Token> acquires AccountOperationsCapability {
-        assert(!AccountFreezing::account_is_frozen(payer), EACCOUNT_FROZEN);
+        LibraTimestamp::assert_operating();
+        assert(!AccountFreezing::account_is_frozen(payer), Errors::invalid_state(EACCOUNT_FROZEN));
```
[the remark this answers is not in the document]
wrwg: Done

language/stdlib/modules/LibraAccount.move
```diff
@@ -635,6 +693,7 @@ module LibraAccount {
         compliance_public_key: vector<u8>,
         add_all_currencies: bool,
     ) {
+        // TODO: restrictions on creator_account?
```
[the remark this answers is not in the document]
wrwg: Ok, punting on it then.

language/stdlib/modules/LibraAccount.move
```diff
@@ -726,12 +787,12 @@ module LibraAccount {
     /// Add a balance of `Token` type to the sending account
     public fun add_currency<Token>(account: &signer) {
         // aborts if `Token` is not a currency type in the system
-        assert(Libra::is_currency<Token>(), ENOT_A_CURRENCY);
+        assert(Libra::is_currency<Token>(), Errors::invalid_argument(ENOT_A_CURRENCY));
```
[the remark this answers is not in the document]
wrwg: Done.

language/stdlib/modules/LibraAccount.move
```diff
@@ -726,12 +787,12 @@ module LibraAccount {
...
         // Check that an account with this role is allowed to hold funds
-        assert(Roles::can_hold_balance(account), EROLE_CANT_STORE_BALANCE);
+        assert(Roles::can_hold_balance(account), Errors::invalid_argument(EROLE_CANT_STORE_BALANCE));
```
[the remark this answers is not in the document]
wrwg: It's also a matter of where we report the error. Reporting it here gives better diagnostics.

language/stdlib/modules/LibraConfig.move
```diff
@@ -24,21 +24,22 @@ module LibraConfig {
...
 
-    const ENOT_GENESIS: u64 = 0;
+    const ECONFIGURATION: u64 = 0;
+    const ELIBRA_CONFIG: u64 = 1;
+    const EMODIFY_CAPABILITY: u64 = 4;
     const ENOT_LIBRA_ROOT: u64 = 1;
     const EINVALID_SINGLETON_ADDRESS: u64 = 2;
     const ECONFIG_DOES_NOT_EXIST: u64 = 3;
-    const EMODIFY_CAPABILITY_NOT_HELD: u64 = 4;
     const EINVALID_BLOCK_TIME: u64 = 5;
```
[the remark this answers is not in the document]
wrwg: Done.

language/stdlib/modules/LibraConfig.move
```diff
@@ -92,30 +110,36 @@ module LibraConfig {
         config_account: &signer,
         payload: Config,
     ): ModifyConfigCapability<Config> {
-        assert(LibraTimestamp::is_genesis(), ENOT_GENESIS);
-        assert(Roles::has_libra_root_role(config_account), ENOT_LIBRA_ROOT);
-        assert(Signer::address_of(config_account) == CoreAddresses::LIBRA_ROOT_ADDRESS(), EINVALID_SINGLETON_ADDRESS);
+        LibraTimestamp::assert_genesis();
+        Roles::assert_libra_root(config_account);
```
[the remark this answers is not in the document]
wrwg: We actually do this together with the role check for singleton address roles. If we generalize at some point, we can relax this directly in the roles module

language/stdlib/modules/LibraSystem.move
```diff
@@ -33,13 +34,20 @@ module LibraSystem {
...
+            exists<CapabilityHolder>(CoreAddresses::LIBRA_ROOT_ADDRESS());
+    }
+
+    const ECAPABILITY_HOLDER: u64 = 0;
+    const EINVALID_PROSPECTIVE_VALIDATOR: u64 = 1;
+    const EALREADY_A_VALIDATOR: u64 = 2;
+    const ENOT_AN_ACTIVE_VALIDATOR: u64 = 3;
+    const EINVALID_TRANSACTION_SENDER: u64 = 4;
+    const EVALIDATOR_INDEX: u64 = 5;
     const ENO_VALIDATOR_OPERATOR_ROLE: u64 = 7;
```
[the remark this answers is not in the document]
wrwg: Done

language/stdlib/modules/LibraTimestamp.move
```diff
@@ -23,40 +23,90 @@ module LibraTimestamp {
...
 
-    const EINVALID_SINGLETON_ADDRESS: u64 = 0;
-    const ETIME_NOT_INITIALIZED: u64 = 1;
-    const ENOT_VM: u64 = 2;
-    const EINVALID_TIMESTAMP: u64 = 3;
-    const EGENESIS_ONLY: u64 = 4;
+    const ENOT_GENESIS: u64 = 0;
+    const ENOT_OPERATING: u64 = 1;
+    const ETIMER_RESOURCE: u64 = 2;
+    const ETIMESTAMP: u64 = 4;
```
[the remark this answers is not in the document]
wrwg: Done

language/stdlib/modules/LibraVersion.move
```diff
@@ -2,36 +2,41 @@ address 0x1 {
...
         LibraConfig::publish_new_config<LibraVersion>(
             lr_account,
             LibraVersion { major: 1 },
         );
     }
 
     public fun set(account: &signer, major: u64) {
+        LibraTimestamp::assert_operating();
+
+        // TODO: is this restricted to be called from libra root?
```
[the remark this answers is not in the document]
wrwg: I leave the TODO for this PR.

language/stdlib/modules/RegisteredCurrencies.move
wrwg: Done

language/stdlib/modules/VASP.move
wrwg: Again, happens in Roles now.

PR #5464, opened 2020-08-04
[fuzz] enable language_transaction_execution

sausagee:
## Motivation
Add the existing fuzzing target to the list so CI can pick it up.

## Test Plan
CI


sausagee:
@mimoo we switched to the new Bors 2.0.  The old delegate+/r+ won't work anymore.  You need to accept the PR.  We can then land with `/land`

PR #5471, opened 2020-08-04
[noise] change noisy noise log critical -> warning

mimoo:


network/src/transport.rs
```diff
@@ -246,6 +246,7 @@ async fn upgrade_inbound<T: TSocket>(
     let (socket, peer_id) = ctxt.noise.upgrade_inbound(socket).await.map_err(|err| {
         // security logging
         send_struct_log!(security_log(security_events::INVALID_NETWORK_PEER)
+            .warning()
             .data_display("error", &err)
             .field(network_events::NETWORK_ADDRESS, &addr));
```
gregnazario, line 251: We should probably make the security logs not default to `critical` instead.  But, I think this is fine for now.

mimoo: they're all criticals though, for a few exceptions : o

PR #5473, opened 2020-08-04
[logging] Keep important fields from being lost

gregnazario:
Due to the limit on the number of bytes in a log line, we are losing
important data due to some really large fields.  This will keep some of
the important error data around for debugging.  This also raises the max
size of a log line.

Realistically, this does not solve the problem of large logs, or keeping information that some logs may find useful or not useful.  But, as a temporary fix, we will at least keep all errors.

### DIFF
--- a/common/logger/src/struct_log.rs
+++ b/common/logger/src/struct_log.rs
@@ -46,11 +46,14 @@ const SEVERITY_CRITICAL: usize = 1;
 const SEVERITY_WARNING: usize = 2;
 
 // Size configurations
-const MAX_LOG_LINE_SIZE: usize = 4096; // 4KiB
+const MAX_LOG_LINE_SIZE: usize = 10240; // 10KiB
 const LOG_INFO_OFFSET: usize = 256;
 const WRITE_CHANNEL_SIZE: usize = 1024;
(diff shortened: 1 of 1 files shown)


andll:
The unit test on this PR failed - we can try to land it with bors, but i am not sure if failure was transient

gregnazario:
> The unit test on this PR failed - we can try to land it with bors, but i am not sure if failure was transient

```
thread 'block_storage::block_store::block_store_and_lec_recovery_test::test_executor_restart' panicked at '
    Unable to build all workspace binaries. Cannot continue running tests.
```
This doesn't seem like it's due to the PR.

common/logger/src/struct_log.rs
```diff
@@ -66,7 +69,7 @@ pub struct StructuredLogEntry {
     /// name of the event
     #[serde(skip_serializing_if = "Option::is_none")]
     name: Option<&'static str>,
-    /// rust module (e.g. consensus::round_manager)
+    /// rust module (e.g. consensus::round_manager):156
```
mimoo, line 72: ?

gregnazario: definitely an accident

PR #5469, opened 2020-08-04
[libra framework] fix specs in p2p script and beyond

emmazzz:
## Motivation

This PR 
- Fixes specs and turns verify back on for `peer_to_peer_with_metadata` transaction script.
- Adds some global invariants in `LibraAccount` needed by the transaction scripts `peer_to_peer_with_metadata` and `rotate_authentication_key`.

### Have you read the [Contributing Guidelines on pull requests](https://github.com/libra/libra/blob/master/CONTRIBUTING.md#pull-requests)?

Yes.

## Test Plan

cargo test

PR #5474, opened 2020-08-04
[breaking][libra framework] decouple AccountLimits and Balance

sblackshear:
- Currently, a ParentVASP needs to have an AccountLimits<CoinType> resource before allowing a child to publish a balance in CoinType. This is confusing and not needed given that AccountLimits are U64_MAX at launch.
- This PR removes the need to do this. Instead, LibraRoot can send an AdminScript to add AccountLimits to any VASP on which limits need to be imposed.
- This simplifies the flows for creating new child accounts and adding new currencies (see smaller specs/fewer comments and error codes). I also was able to remove the partial aborts_if for the add_currency script.
- Also eliminated the tx scripts for updating account limits, since this can now be done via the admin script that installs them.

### Breaking changes
- This is a breaking change in the sense that it changes the bytecode of a few modules and the transaction script allowlist.
- This is *not* a breaking change for clients, although the `add_currency_to_account` script and `create_child_vasp` script will now succeed in some cases that would previously have failed (e.g., creating a child VASP with `add_all_currencies=true` from a parent that only has a balance in LBR. This convenience is the primary motivation for the change.


andll:
Note that for breaking change you need to prefix commit headline as `[breaking]` now, otherwise it will be rejected by land blocking test

language/stdlib/modules/VASP.move
```diff
@@ -154,6 +146,9 @@ module VASP {
                 global<ChildVASP>(addr).parent_vasp_addr
             }
         }
+        define spec_has_account_limits<Token>(addr: address): bool {
```
sblackshear, line 149: @emmazzz: I had to define a spec function for this because the prover wasn't happy with the purity of `has_account_limits`. Do you know why? Is it because it's not happy with the purity of `parent_address`?

emmazzz: That's right. `parent_address` has an `abort` in it, which is currently not supported.

emmazzz:
> [move-prover] Move Prover test (`language/move-prover/cargo test`) on commit [d47cc50](https://github.com/libra/libra/commit/d47cc50ccbdaa7ea67658ed192ee0b5c2a9b239d) did **fail**. See [details](https://github.com/mimoo/mirai-bot/runs/955621228?check_suite_focus=true#step:7:1).

I've seen this failure too after my PR(#5469 ) got merged, and I can't reproduce it locally. Can you reproduce it @sblackshear?

sblackshear:
I did do something 

> > [move-prover] Move Prover test (`language/move-prover/cargo test`) on commit [d47cc50](https://github.com/libra/libra/commit/d47cc50ccbdaa7ea67658ed192ee0b5c2a9b239d) did **fail**. See [details](https://github.com/mimoo/mirai-bot/runs/955621228?check_suite_focus=true#step:7:1).
> 
> I've seen this failure too after my PR(#5469 ) got merged, and I can't reproduce it locally. Can you reproduce it @sblackshear?

I did accidentally break `peer_to_peer_with_metadata` in d47cc50, but fixed it in 028d647 (works locally as well).

language/stdlib/modules/VASP.move
sblackshear: Is there a conceptual blocker to treating funs with `abort`s as pure, or just an implementation one? I would guess the latter since @wrwg told me that spec functions can be partial, but wanted to make sure I understand.

moeziniaf:
> Love red diffs! This looks good to me. cc @moeziniaf for visibility about the removal of the two scripts in case it has AOS implications.

Thanks - which admin script will be used to add AccountLimits after this PR?

sblackshear:
> > Love red diffs! This looks good to me. cc @moeziniaf for visibility about the removal of the two scripts in case it has AOS implications.
> 
> Thanks - which admin script will be used to add AccountLimits after this PR?

We do not have a script for this yet, but it should be as simple as calling `AccountLimits::publish_new_window`.

moeziniaf:
> > > Love red diffs! This looks good to me. cc @moeziniaf for visibility about the removal of the two scripts in case it has AOS implications.
> > 
> > 
> > Thanks - which admin script will be used to add AccountLimits after this PR?
> 
> We do not have a script for this yet, but it should be as simple as calling `AccountLimits::publish_new_window`.

Is there some limit address which holds default limits? I'm guessing there's a default and after MVP we will want to be able to create/modify new limits (but for now just publish and point an account window to the default limits address?)

PR #5475, opened 2020-08-04
[cluster-test] Increase delete_resource backoff timeout

andll:
Increasing backoff for k8s resources deletion 150sec -> 300sec. This fails when deleting large clusters, like ones with 100 nodes

### DIFF
--- a/testsuite/cluster-test/src/cluster_swarm/cluster_swarm_kube.rs
+++ b/testsuite/cluster-test/src/cluster_swarm/cluster_swarm_kube.rs
@@ -292,7 +292,7 @@ impl ClusterSwarmKube {
     {
         debug!("Deleting {} {}", T::KIND, name);
         let resource_api: Api<T> = Api::namespaced(self.client.clone(), DEFAULT_NAMESPACE);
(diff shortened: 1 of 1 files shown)


snakech:
> /land

we use land instead of r+ now?

andll:
Yes :

changyonghu:
@andll I think this is because we add hundreds jobs (mkdir/put_file) recently. I saw delete job failed error yesterday when I ran CT, so this fix still have chances to fail. I would like to propose that since we are using S3 instead of kubectl cp, we can just upload files to S3 in CT, removing all util_cmd(mkdir) and util_cmd(aws cp), later in init_container, we can create dir and download files from S3. What do you think?

andll:
@changyonghu I think I have alternative proposal - it looks like jobs are left in cluster after they complete, which is probably wrong anyway. Perhaps better way is to just delete them in `wait_job_completion` ones they complete?

The reason why I think this is better, is because we want to have util_cmd as a first class citizen in cluster test, there will be future usages of it too. So I would rather fix it so it does not cause issues, rather then use some workaround

What do you think?

changyonghu:
Yes, if this issue got fixed, I think the current way is fine.

PR #5476, opened 2020-08-04
[logging] add all log levels to structured logging

gregnazario:
Text logs are automatically converted to structured logs, and we didn't
log the levels.  This now logs the log levels for all logs, and adds the
ability to have more fine grained levels for structured logs in general.

### DIFF
--- a/common/logger/src/lib.rs
+++ b/common/logger/src/lib.rs
@@ -137,13 +137,28 @@ pub use log::Level;
 pub use text_log::{Logger, CHANNEL_SIZE, DEFAULT_TARGET};
 pub mod counters;
 
+use serde::Serialize;
+
+// severity level - lower is worse
+// TODO: We can remove this if we get rid of critical
+// TODO: should we serialize it into text?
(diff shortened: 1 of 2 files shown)


andll:
There is some inconsistency here in terms of naming - the field on log entry is called 'severity', but some other things are called level(like LogLevel enum and method to set level)
We should probably converge on single name

gregnazario:
> There is some inconsistency here in terms of naming - the field on log entry is called 'severity', but some other things are called level(like LogLevel enum and method to set level)
> We should probably converge on single name

Yeah, I wanted to move it just to `level` as that's what's fairly common among logging.  Just need to talk to @sherry-x and make sure we don't have anything setup as a log scan on severity right now

gregnazario:
> There is some inconsistency here in terms of naming - the field on log entry is called 'severity', but some other things are called level(like LogLevel enum and method to set level)
> We should probably converge on single name

Discussed with Sherry.  Will followup here https://github.com/libra/libra/issues/5484 after this has been deployed

PR #5477, opened 2020-08-04
[network] decrease logging size for large log messages

gregnazario:
The log messages printing out the full status of addresses and eligible
peers was significantly longer due to being duplicated in structured
logs.  By making it a structured log it reduces that duplication, and
allows us to use strucutred logs for this information.

### DIFF
--- a/network/src/connectivity_manager/mod.rs
+++ b/network/src/connectivity_manager/mod.rs
@@ -28,7 +28,7 @@
 //! using a relay protocol.
 
 use crate::{
-    logging::*,
+    logging::{network_events::DISCOVERY_SOURCE, *},
     peer_manager::{self, conn_notifs_channel, ConnectionRequestSender, PeerManagerError},
 };
 use futures::{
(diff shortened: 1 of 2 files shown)


network/src/connectivity_manager/mod.rs
```diff
@@ -544,10 +544,9 @@ where
         // Only log the total state if anything has actually changed.
         if have_any_changed {
             let peer_addrs = &self.peer_addrs;
-            info!(
-                "{} current addresses: update src: {}, all peer addresses: {}",
-                self.network_context, src, peer_addrs,
-            );
+            send_struct_log!(network_log("PeerAddressesUpdate", &self.network_context)
+                .field(DISCOVERY_SOURCE, &src)
+                .data("PeerAddresses", &peer_addrs));
```
phlip9, line 549: nit: should these keys be snake cased?

network/src/connectivity_manager/mod.rs
```diff
@@ -595,11 +594,9 @@ where
             // to generate the new eligible peers set.
             let new_eligible = self.peer_pubkeys.union_all();
 
-            let peer_pubkeys = &self.peer_pubkeys;
-            info!(
-                "{} current pubkeys: update src: {}, all peer pubkeys: {}, new eligible set: {:?}",
-                self.network_context, src, peer_pubkeys, new_eligible,
-            );
+            send_struct_log!(network_log("EligiblePeersUpdate", &self.network_context)
+                .field(DISCOVERY_SOURCE, &src)
```
phlip9, line 598: i'm guessing we're using `.field` when we use a key in multiple places?

network/src/connectivity_manager/mod.rs
gregnazario: Yeah, let's snake case everything.

network/src/connectivity_manager/mod.rs
gregnazario: Yeah, it was the thought that you could make a typed field, so that you don't accidentally put the wrong type in.

PR #5478, opened 2020-08-04
[config] Allow waypoint to be read from a file

mgorven:
To get public fullnodes running in the k8s testnet I need to get the
waypoint from a configmap at runtime. I could do some hacky runtime
interpolation of the config file, but it seems nicer to just support
reading the waypoint from a file.

Test Plan: Tested in testnet.


mgorven:
> so both the genesis and the waypoint will be coded in the config map?

For public fullnodes yes. For validators and validator fullnodes the waypoint will still be retrieved from Vault.

PR #5479, opened 2020-08-05
safety-rules: don't build all workspace binaries

bmwill:
Using workspace-builder inside of a unit-test (and outside of the main smoke tests) isn't the most ideal thing to do. It would be better to instead rely on cargo ensuring that it builds the binary that is under test. This is done by converting the safety-rules tests to be integration tests.


rexhoffman:
Will the binaries be built with fuzzing/testing features -- I suspect so?   Is this okay in our tests?

davidiw:
Thanks for initiating the conversation here and showing me how to setup integration tests. This was a really great move! Closing this out as we landed on a slightly different approach.

Issue #5134, opened 2020-07-16
[client tooling] Tool to sync local state to Libra state at a certain height, execute a transaction, view the state


sblackshear:

sblackshear:
This would be useful for local testing by clients and Libra administrators. It is also related to https://github.com/libra/libra/issues/4803: supporting read-only Move calls.

zekun000:
cc: @msmouse this sounds like backup&restore (sync to a certain height) + db bootstrapper (execute a txn and view the state)

msmouse:
The tools already exists if we are talking about a DB that's behind the target version -- you just recover additional transactions in replay mode on top of the existing db, and use the db-bootstrapper. 

There's no tooling to go back in history though.

bob-wilson:
@runtian-zhou and I discussed this today and he is going to follow up to figure out how we can make it easier to put these tools together

bob-wilson:
CC @runtian-zhou 
The transaction-replay tool is what we needed here. This is done now.

PR #5481, opened 2020-08-05
[transaction-replay] Build a tool to replay transactions stored onchain

runtian-zhou:
## Motivation

Create a binary to replay transactions stored inside of a LibraDB and to annotate on chain resources for debugging purposes. Currently it supports following four commands:

1. Annotate move resources stored in an account at given block height(version).
```
cargo run -- --db <path-to-db> annotate-account <address> <version>
```

2. Rerun transactions from version X to Y.
```
cargo run -- --db <path-to-db> replay-transactions <begin_version> <limit>
```

3. Rerun n latest transations
```
cargo run -- --db <path-to-db> replay-recent-transactions <number>
```

4. Rerun account X's nth transaction
```
cargo run -- --db <path-to-db> replay-transaction-by-sequence-number <account_address> <sequence_number>
```

Note that currently this tool requires reading from db directly to get the transactions due to the lack of getting transaction api for json-rpc. In the long run we would be able to run this tool on our own machine by connecting the debug interface to the json-rpc.

Yes

## Test Plan

Tested all four commands over the db created by libra-swarm. I specifically tried to create some reconfiguration transactions to see if the configs are being refreshed correctly.

## Related PRs

#5134


language/tools/transaction-replay/src/transaction_debugger_interface.rs
```diff
@@ -0,0 +1,112 @@
...
+    account_state::AccountState,
+    account_state_blob::AccountStateBlob,
+    transaction::{Transaction, Version},
+};
+use libradb::LibraDB;
+use std::{convert::TryFrom, path::Path, sync::Arc};
+use storage_interface::DbReader;
+
+pub trait StorageDebuggerInterface {
+    fn get_account_state_by_version(
```
runtian-zhou, line 18: I defined the API needed to be provided from the system here. Currently it is implemented with a read-only Local DB instance. In the future we would like this to be done via some rpc services to be able to run this tool outside of production environment.

language/tools/transaction-replay/src/transaction_debugger_interface.rs
```diff
@@ -0,0 +1,112 @@
...
+    }
+}
+
+pub struct DebuggerStateView<'a, S: StorageDebuggerInterface> {
+    db: &'a S,
+    version: Version,
+}
+
+impl<'a, S: StorageDebuggerInterface> DebuggerStateView<'a, S> {
+    pub fn new(db: &'a S, version: Version) -> Self {
```
runtian-zhou, line 46: The version here is off by one from the real version due to the genesis reason.

sblackshear:
> Annotate move resources stored in an account at given block height(version).

What does this mean?

thefallentree:
I feels like we could just make this part of the storage-inspector , but don't know enough of the use case to see if that worth the effort.

For get_transcations_with_proof json-rpc- api yeah we were going to add it, but it is not on high priority list,  also we don't really have an concept of "debug" interface of jsonrpc api yet, it is just one endpoint

language/tools/transaction-replay/src/transaction_debugger_interface.rs
```diff
@@ -0,0 +1,112 @@
...
+
+impl StorageDebuggerInterface for LocalDBDebugger {
+    fn get_account_state_by_version(
+        &self,
+        account: AccountAddress,
+        version: Version,
+    ) -> Result<Option<AccountStateBlob>> {
+        Ok(self
+            .0
+            .get_account_state_with_proof_by_version(account, version)?
```
sblackshear, line 86: I assume this will fail if the local DB doesn't contain the state at `version` Could we create the state by replaying transactions up to `version` instead of failing here?

language/tools/transaction-replay/src/transaction_debugger_interface.rs
sblackshear: Even if json-rpc or a similar service doesn't support fetching the state at *every* version (which would be a lot of data!), perhaps someone could provide a lighter API for fetching the state at (e.g.) every epoch change or every waypoint?

language/tools/transaction-replay/src/lib.rs
```diff
@@ -0,0 +1,118 @@
...
+    ) -> Result<Vec<TransactionOutput>> {
+        let state_view = DebuggerStateView::new(&self.debugger, version);
+        LibraVM::execute_block(txns, &state_view)
+            .map_err(|err| format_err!("Unexpected VM Error: {:?}", err))
+    }
+
+    pub fn execute_past_transactions(
+        &self,
+        mut begin: Version,
+        mut limit: u64,
```
sblackshear, line 39: What value should clients provide if they want to say "replay everything from `begin` to the latest known version" (probably a common use-case)?

runtian-zhou:
> > Annotate move resources stored in an account at given block height(version).
> 
> What does this mean?

It means that using the resource-viewer to annotate the resources stored in an account's blob at particular height.

language/tools/transaction-replay/src/transaction_debugger_interface.rs
runtian-zhou: I'm not sure how this restore part is going to work. IIRC the LibraDB will do the trim every month so this tool should be able to allow us to look into transactions that happened in the past month. @msmouse do you have any clues on how this could happen in production?

language/tools/transaction-replay/src/lib.rs
runtian-zhou: "client" as in end point users like VASP operators?

runtian-zhou:
IMHO cli serves for a different purpose. But in general, this binary, genesis-viewer and storage-inspector could live in the same binary ideally as they all serve for the same debugging in prod purposes.

sblackshear:
> > > Annotate move resources stored in an account at given block height(version).
> > 
> > 
> > What does this mean?
> 
> It means that using the resource-viewer to annotate the resources stored in an account's blob at particular height.

Ah, got it. "Annotate" somewhat is confusing to me because it implies a write of some kind. Perhaps "decode" or similar could work?

runtian-zhou:
Huge thanks to @thefallentree 's exellent work of adding transaction raw bytes to the json rpc. With this piece of data I can build the entire replay tooling based on json rpc. I also implemented a smoke_test to hook the tool up with the libra-swarm to make sure things don't go off in the future.

language/tools/transaction-replay/src/lib.rs
sblackshear: Yes (or anyone else using this tool).

language/tools/transaction-replay/src/transaction_debugger_interface.rs
msmouse: > I assume this will fail if the local DB doesn't contain the state at version Could we create the state by replaying transactions up to version instead of failing here?

We can't replay backwards, so when a state is pruned, we need to use use a backup and restore to the specific version to an empty folder.

> IIRC the LibraDB will do the trim every month

no, it trims continuously according to this config (keeping 40 Million versions by default, that's about 350GB of disc space): 
https://github.com/libra/libra/blob/650983ad5bb3098b75d2aede6ad47ca6c977cc68/config/src/config/storage_config.rs#L34-L36

runtian-zhou: Does that mean that json-rpc can only respond to queries within that time window?

msmouse: Yes, otherwise error.

PR #5482, opened 2020-08-05
[network][noise] adding a fuzzer for the post-handshake part

mimoo:


testsuite/libra-fuzzer/src/fuzz_targets/network_noise_stream.rs
```diff
@@ -0,0 +1,27 @@
...
+    fn name(&self) -> &'static str {
+        module_name!()
+    }
+
+    fn description(&self) -> &'static str {
+        "Network Noise crate stream"
+    }
+
+    fn generate(&self, _idx: usize, gen: &mut ValueGenerator) -> Option<Vec<u8>> {
+        Some(generate_corpus(gen))
```
phlip9, line 21: shouldn't the corpus be a valid message frame after handshake and not a handshake message?

mimoo: it shouldn't matter because the fuzzer will never generate a message that you can decrypt successfully

PR #5486, opened 2020-08-05
[config] Log important config values on startup

gregnazario:
We aren't logging the config because we print out the values before the
logger is set up.  Now, we should be able to add all values after
logging is set up that we care about.  The first value that we should
log is the chain id.

### DIFF
--- a/libra-node/src/main.rs
+++ b/libra-node/src/main.rs
@@ -4,7 +4,7 @@
 #![forbid(unsafe_code)]
 
 use libra_config::config::NodeConfig;
-use libra_logger::prelude::*;
+use libra_logger::{prelude::*, LogLevel};
 use libra_types::PeerId;
 use std::{
     path::PathBuf,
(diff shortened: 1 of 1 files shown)


gregnazario:
> LGTM
> 
> do we have any plan to log the whole config? or do we want to just pick out the useful pieces individually?

The whole config is pretty large.  I think we can look at that next, but the most important piece was the chain id.

aching:
Glad to at least get the chain id, but why not log the whole config btw?  Seems pretty useful.

gregnazario:
> Glad to at least get the chain id, but why not log the whole config btw? Seems pretty useful.

I figured it would be a massive log, and currently since we're dealing with some weird fallout from the logs, shorter is better at the moment.  But, I could log the entire thing

gregnazario:
@aching @phlip9 Turned out to not be that big, so I've changed to just printing out the whole config (or at least what will serialize)

http://kibana.ct-1-k8s-testnet.aws.hlw3truzy4ls.com/app/kibana#/doc/9d3f1cd0-bcab-11ea-9157-7553013abcd5/kubernetes_cluster-2020.08.07?id=1cb819a6-5fa1-89f6-1b07-602d4a107595&_g=(filters:!(),time:(from:'2020-08-07T22:27:38Z',to:'2020-08-07T22:38:43Z'))

PR #5488, opened 2020-08-05
[cluster-test] add spawn_job and kill_job for ClusterSwarmKube

rustielin:
## Motivation

This allows us to spawn long-running jobs like faucet, execute
depedent jobs such as for CLI commands, and kill it later on.

Adds a skeleton for client_compatibility_test, which will include
spawning a faucet as a k8s job, running CLI jobs to mint against
that, and killing the faucet job at the end. For now, it's just a dummy
experiment that kills a job that runs forever.

Note: added an additional param `killed` to `wait_job_completion` to handle cases where we expect the job to be missing. This handles a race condition where jobs are fully deleted by the time we wait for completion.

Yes

## Test Plan

Run client_compatibility_test, which is a dummy experiment for now that spins up a job that runs forever, waits a while, then kills it:

```
DEBUG 2020-08-06 01:29:53 testsuite/cluster-test/src/cluster_swarm/cluster_swarm_kube.rs:462 Starting job run-forever for node ip-192-168-204-177.us-west-2.compute.internal
INFO 2020-08-06 01:29:53 testsuite/cluster-test/src/experiments/client_compatibility_test.rs:48 Wait for the other jobs to finish
Deleting Job: Some(JobStatus { active: Some(1), completion_time: None, conditions: None, failed: None, start_time: Some(Time(2020-08-06T01:29:53Z)), succeeded: None })
DEBUG 2020-08-06 01:30:13 common/retrier/src/lib.rs:43 job in still in progress run-forever-ku5vyby7ak. Retrying in 5 seconds..
INFO 2020-08-06 01:30:18 testsuite/cluster-test/src/main.rs:640 Experiment finished, waiting until all affected validators recover
```

Also run the standard land_blocking_compat experiment suite to make sure I haven't broken anything:

```
Compatibility test results for land_4f212d33 ==> dev_rustielin_pull_5488 (PR)
1. All instances running land_4f212d33, generating some traffic on network
2. First validator land_4f212d33 ==> dev_rustielin_pull_5488, to validate storage
3. First batch validators (14) land_4f212d33 ==> dev_rustielin_pull_5488, to test consensus
4. Second batch validators (15) land_4f212d33 ==> dev_rustielin_pull_5488, to upgrade rest of the validators
5. Reset all nodes ==> dev_rustielin_pull_5488, to finish the network upgrade
all up : 1148 TPS, 3941 ms latency, 4650 ms p99 latency, no expired txns
```

## Related PRs

### DIFF
--- a/testsuite/cluster-test/src/cluster_builder.rs
+++ b/testsuite/cluster-test/src/cluster_builder.rs
@@ -78,7 +78,7 @@ impl ClusterBuilderParams {
 
 pub struct ClusterBuilder {
     pub current_tag: String,
-    cluster_swarm: ClusterSwarmKube,
+    pub cluster_swarm: ClusterSwarmKube,
 }
 
 impl ClusterBuilder {
--- a/testsuite/cluster-test/src/cluster_swarm/cluster_swarm_kube.rs
+++ b/testsuite/cluster-test/src/cluster_swarm/cluster_swarm_kube.rs
@@ -11,7 +11,7 @@ use async_trait::async_trait;
 use futures::{future::try_join_all, lock::Mutex};
 use k8s_openapi::api::core::v1::{ConfigMap, Node, Pod, Service};
 use kube::{
-    api::{Api, PostParams},
+    api::{Api, DeleteParams, PostParams},
     client::Client,
     Config,
 };
@@ -218,7 +218,42 @@ impl ClusterSwarmKube {
             .map_err(|e| format_err!("serde_json::from_value failed: {}", e))
     }
 
-    async fn wait_job_completion(&self, job_name: &str, back_off_limit: u32) -> Result<bool> {
+    fn job_spec(
+        &self,
+        k8s_node: &str,
+        docker_image: &str,
+        command: &str,
+        job_name: &str,
+        back_off_limit: u32,
+    ) -> Result<(Job, String)> {
+        let suffix = thread_rng()
+            .sample_iter(&Alphanumeric)
+            .take(10)
+            .collect::<String>()
+            .to_ascii_lowercase();
+        let job_full_name = format!("{}-{}", job_name, suffix);
+        let job_yaml = format!(
+            include_str!("job_template.yaml"),
+            name = &job_full_name,
+            label = job_name,
+            image = docker_image,
+            node_name = k8s_node,
+            command = command,
+            back_off_limit = back_off_limit,
+        );
+        let job_spec: serde_yaml::Value = serde_yaml::from_str(&job_yaml)?;
+        let job_spec = serde_json::value::to_value(job_spec)?;
+        let job_spec = serde_json::from_value(job_spec)
+            .map_err(|e| format_err!("serde_json::from_value failed: {}", e))?;
+        Ok((job_spec, job_full_name))
+    }
+
+    async fn wait_job_completion(
+        &self,
+        job_name: &str,
+        back_off_limit: u32,
+        killed: bool,
+    ) -> Result<bool> {
         libra_retrier::retry_async(libra_retrier::fixed_retry_strategy(5000, 20), || {
             let job_api: Api<Job> = Api::namespaced(self.client.clone(), DEFAULT_NAMESPACE);
             let job_name = job_name.to_string();
@@ -241,14 +276,45 @@ impl ClusterSwarmKube {
                         }
                         bail!("job in still in progress {}", job_name)
                     }
-                    Err(e) => bail!("job_api.get failed for job {} : {:?}", job_name, e),
+                    Err(e) => {
+                        if killed {
+                            info!("Job {} has been killed already", job_name);
+                            return Ok(true);
+                        }
+                        bail!("job_api.get failed for job {} : {:?}", job_name, e)
+                    }
                 }
             })
         })
         .await
     }
 
-    async fn run_jobs(&self, jobs: Vec<Job>, back_off_limit: u32) -> Result<()> {
+    pub async fn kill_job(&self, job_full_name: &str) -> Result<()> {
+        let dp = DeleteParams::default();
+        let job_api: Api<Job> = Api::namespaced(self.client.clone(), DEFAULT_NAMESPACE);
+        job_api
+            .delete(job_full_name, &dp)
+            .await?
+            .map_left(|o| println!("Deleting Job: {:?}", o.status))
+            .map_right(|s| println!("Deleted Job: {:?}", s));
+        let back_off_limit = 0;
+        // the job night have been deleted already, so we do not handle error
+        match self
+            .wait_job_completion(job_full_name, back_off_limit, true)
+            .await
+        {
+            Ok(_) => info!("Killing job {} returned job_status.success", job_full_name),
+            Err(error) => info!(
+                "Killing job {} returned job_status.failed: {}",
+                job_full_name, error
+            ),
+        }
+
+        Ok(())
+    }
+
+    // just ensures jobs are started, but does not wait on completion
+    async fn start_jobs(&self, jobs: Vec<Job>) -> Result<Vec<String>> {
         let pp = PostParams::default();
         let job_api: Api<Job> = Api::namespaced(self.client.clone(), DEFAULT_NAMESPACE);
         let create_jobs_futures = jobs.iter().map(|job| job_api.create(&pp, job));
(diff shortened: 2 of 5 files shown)


testsuite/cluster-test/src/cluster_swarm/cluster_swarm_kube.rs
```diff
@@ -387,6 +434,21 @@ impl ClusterSwarmKube {
         Ok(workspace.clone())
     }
 
+    pub async fn spawn_job(
```
andll, line 437: Can you make `Instance::spawn_job` that will call into this function, and use it from client compat test?
In this case, you don't need to make `K8sInstanceInfo `, `K8sInstanceInfo::k8s_node` and `Instance::k8s_backend` public.

testsuite/cluster-test/src/experiments/client_compatibility_test.rs
```diff
@@ -0,0 +1,72 @@
...
+
+impl ExperimentParam for ClientCompatiblityTestParams {
+    type E = ClientCompatibilityTest;
+    fn build(self, _cluster: &Cluster) -> Self::E {
+        Self::E {}
+    }
+}
+
+#[async_trait]
+impl Experiment for ClientCompatibilityTest {
```
andll, line 29: Can you make some comment here that test itself does not do anything right now (some kind of TODO)

testsuite/cluster-test/src/cluster_swarm/cluster_swarm_kube.rs
```diff
@@ -248,7 +278,19 @@ impl ClusterSwarmKube {
...
-    async fn run_jobs(&self, jobs: Vec<Job>, back_off_limit: u32) -> Result<()> {
+    pub async fn kill_job(&self, job_full_name: &str) -> Result<()> {
+        let dp = DeleteParams::default();
+        let job_api: Api<Job> = Api::namespaced(self.client.clone(), DEFAULT_NAMESPACE);
+        job_api
+            .delete(job_full_name, &dp)
+            .await?
+            .map_left(|o| println!("Deleting Job: {:?}", o.status))
+            .map_right(|s| println!("Deleted Job: {:?}", s));
+        Ok(())
```
andll, line 289: Do you want to also call into `wait_job_completion` here to wait until job is actually deleted?

testsuite/cluster-test/src/experiments/client_compatibility_test.rs
```diff
@@ -0,0 +1,72 @@
...
+    fn affected_validators(&self) -> HashSet<String> {
+        HashSet::new()
+    }
+    async fn run(&mut self, context: &mut Context<'_>) -> anyhow::Result<()> {
```
andll, line 33: new line between functions :)

testsuite/cluster-test/src/cluster_swarm/cluster_swarm_kube.rs
rustielin: I figured it might make sense to leave that up to the user to decide, as per my note above. Jobs are already cleaned up during cluster-test runner teardown, so might make sense to batch all the deleting there? What do you think

andll: I think waiting is probably a safer bet.
I understand your argument about saving time, but I think most people aren't going to read to much into it, and will just assume that kill have job killed when succeeded. So waiting will probably provide a safer api this way.

rustielin: Got it, updated. Had to add a new param to `wait_job_completion` to handle a race if job doesn't exist by the time we start to wait for it.

testsuite/cluster-test/src/cluster_builder.rs
```diff
@@ -49,7 +49,7 @@ impl ClusterBuilderParams {
 
 pub struct ClusterBuilder {
     pub current_tag: String,
-    cluster_swarm: ClusterSwarmKube,
+    pub cluster_swarm: ClusterSwarmKube,
```
andll, line 52: One more interface nit - instead of making this one public, do you mind just adding `kill_job` to the ClusterSwarm interface, and then call it through context.cluster_swarm?

testsuite/cluster-test/src/cluster_swarm/cluster_swarm_kube.rs
```diff
@@ -241,14 +276,36 @@ impl ClusterSwarmKube {
...
+        let dp = DeleteParams::default();
+        let job_api: Api<Job> = Api::namespaced(self.client.clone(), DEFAULT_NAMESPACE);
+        job_api
+            .delete(job_full_name, &dp)
+            .await?
+            .map_left(|o| println!("Deleting Job: {:?}", o.status))
+            .map_right(|s| println!("Deleted Job: {:?}", s));
+        let back_off_limit = 0;
+        // the job night have been deleted already, so we do not handle error
+        let _ = self
```
andll, line 301: I like the idea of supplying `killed` arg to not return error in wait_job_completion if we ourselves triggered kill

However I am a bit surprised we still need to ignore error even after having this arg.

I think we should not ignore error here - there is legitimate case for it, for example if job is not terminated for whatever reason [it is always good to surface some strange cases, instead of sweeping them under the rug, even if in theory you can argue that we don't care if job is deleted or not]

I think you might need to add another `if killed` to part where you handle status code, e.g. do not return error in `if let Some(failed) = job_status.failed` when `killed` is set

testsuite/cluster-test/src/cluster_builder.rs
rustielin: How would we do `kill_job` for ClusterSwarm (not kubernetes)? To me, there's no clear mapping and *_job makes sense for only ClusterSwarmKube

andll: We probably will refactor it a bit, so that spawning job returns some handle, not a string. In this case job without kubernetes probably going to be a process, and handle will contain PID

testsuite/cluster-test/src/cluster_swarm/cluster_swarm_kube.rs
rustielin: I'll play around with your suggestion. My initial thought was that we're not guaranteed a good job_status code when we terminate a job, since that seems like it depends on the nature of the job we run.

testsuite/cluster-test/src/cluster_builder.rs
rustielin: Can we push this out to later PR? I'd like to get this out soon so I can unblock client compat testing. I've created an issue for this https://github.com/libra/libra/issues/5507

andll: sure.
Just to make sure - we don't need to refactor string->handle right now.
I only want to move kill_job to interface

testsuite/cluster-test/src/cluster_swarm/cluster_swarm_kube.rs
rustielin: Updated to just print more logs. I'll still let it accept any job_status, but might be useful for debug later on.

```
DEBUG 2020-08-06 22:36:49 testsuite/cluster-test/src/cluster_swarm/cluster_swarm_kube.rs:471 Starting job run-forever for node ip-192-168-227-150.us-west-2.compute.internal
INFO 2020-08-06 22:36:49 testsuite/cluster-test/src/experiments/client_compatibility_test.rs:48 Wait for the other jobs to finish
Deleting Job: Some(JobStatus { active: Some(1), completion_time: None, conditions: None, failed: None, start_time: Some(Time(2020-08-06T22:36:49Z)), succeeded: None })
DEBUG 2020-08-06 22:37:09 common/retrier/src/lib.rs:43 job in still in progress run-forever-u6eywbvbyx. Retrying in 5 seconds..
INFO 2020-08-06 22:37:14 testsuite/cluster-test/src/cluster_swarm/cluster_swarm_kube.rs:281 Job run-forever-u6eywbvbyx has been killed already
INFO 2020-08-06 22:37:14 testsuite/cluster-test/src/cluster_swarm/cluster_swarm_kube.rs:306 Killing job run-forever-u6eywbvbyx returned job_status.success
INFO 2020-08-06 22:37:14 testsuite/cluster-test/src/main.rs:640 Experiment finished, waiting until all affected validators recover
```

PR #5490, opened 2020-08-05
[storage] Fix a lint error due to missing test attribute.

changyonghu:
Since nobody is fixing this.


andll:
Thanks @changyonghu 

@changyonghu @metajack do we want to catch those problems in CI, so that we don't have to manually fix unrelated files?

snakech:
> @bors-libra /land

...

metajack:
Bors understood the original command and queued the build. It will land the code when this PR is up in the queue.

metajack:
> > @bors-libra /land
> 
> ...

Don't prefix your command with `@bors-libra`. The new system does not require that.

metajack:
> Bors understood the original command and queued the build. It will land the code when this PR is up in the queue.

FYI, you can tell it did this because of the rocket ship emoji bors put on the original `/land` command.

snakech:
> Bors understood the original command and queued the build. It will land the code when this PR is up in the queue.

ahaha, Acked, Bors is running reinforcement learning

snakech:
> > Bors understood the original command and queued the build. It will land the code when this PR is up in the queue.
> 
> FYI, you can tell it did this because of the rocket ship emoji bors put on the original `/land` command.

thx for the new feature, just replaced a new laptop without slack during shipment, lost some context about bors change

changyonghu:
@metajack Seems ValidatorSet is auto imported, so we don't need to import twice.

PR #5491, opened 2020-08-05
[vm] Make unexpected verification errors invariant violations

tnowacki:
- Mapped errors coming out of verifying modules in the loader as internal invariant violations

How it works:
- `load_module` is the only spot we verify modules coming off chain
  - Modules to be published are verified in the runtimes `publish_module` via the loaders `verify_module`
- Any error coming out of `deserialize_and_verify_module` for `load_module` are marked as invariant violations 
- `deserialize_and_verify_module` was made private to `load_module` to prevent any shenanigans 
- There are two versions of `load_module` to specify whether or not a missing module is an error or not
  - When first publishing a module, or loading a script, a missing module needs to be considered a user error. It's possible its an internal invariant violation, but it's really impossible to know
  - When loading dependencies, the transitive modules *must* exist, or we have unverifiable modules on chain 

Ideally we would want to mark *exactly* what spots we expect verifier errors to come from (with all other spots being automatically invariant violations). But due to the recursive nature of the loader, particularly with `load_type` this isn't easily doable. 

Eventually, we might not even verify modules when loading them, so this whole thing becomes moot. This approach feels more inline with the current verification being an extra assertion

## Motivation

- Without doing this, we could charge users for somehow having unverifiable code published on chain

## Test Plan

- Ran tests


language/move-vm/runtime/src/loader.rs
```diff
@@ -711,14 +664,98 @@ impl Loader {
...
+                return Err(err.expect_no_verification_errors());
+            }
+        };
+
+        let module = deserialize_and_verify_module(self, bytes, data_store)
+            .map_err(|e| e.expect_no_verification_errors())?;
         self.module_cache.lock().unwrap().insert(id.clone(), module)
     }
 
+    fn load_module_possible_missing_modules(
```
dariorussi, line 723: I cannot stop laughing at those function names...

language/move-vm/runtime/src/loader.rs
```diff
@@ -711,14 +664,98 @@ impl Loader {
...
+        }
+
         if let Some(module) = self.module_cache.lock().unwrap().module_at(id) {
             return Ok(module);
         }
-        let module = self.deserialize_and_verify_module(id, data_store)?;
+
+        let bytes = match data_store.load_module(id) {
+            Ok(bytes) => bytes,
+            Err(err) if possible_missing_modules => return Err(err),
```
dariorussi, line 711: it's pretty funny how unreadable I find this pattern... :)

language/move-vm/runtime/src/loader.rs
tnowacki: oh I know lol
I couldn't think of anything succinct, so I tried as much as possible to make it *extra* clear what was going on. In doing so, it might actually be more confusing than something shorter.

tnowacki:
> I'd be inclined to get it in and we'll work on some tests and keep thinking of it.... thoughts?

Luckily, we already have one test (a non verifiable module is published, and then accessed via a script). I was going to add two more for the module case and one or two for the `load_type` case

language/move-vm/runtime/src/loader.rs
```diff
@@ -563,7 +562,28 @@ impl Loader {
     // This step performs all verification steps to load the module without loading it.
     // The module is not added to the code cache. It is simply published to the data cache.
     // See `verify_script()` for script verification steps.
-    pub(crate) fn verify_module(&self, module: &CompiledModule) -> VMResult<()> {
+    pub(crate) fn verify_module_verify_no_missing_dependencies(
```
dariorussi, line 565: c'mon this is obviously trolling....

tnowacki:
old habits

PR #5492, opened 2020-08-05
[backup] backup controllers: add random suffix to backup names

msmouse:
## Motivation

so when one fails in the middle, the retry won't use the same name.
it may leave behind some garbage, but the coordinator will be more robust.

Y

## Test Plan

existing coverage

and examine temp dir during the test runs:

```shellsession
$ ll
total 0
drwxr-xr-x   4 aldenhu  staff  128 Aug  5 16:27 metadata
drwxr-xr-x  22 aldenhu  staff  704 Aug  5 16:27 state_ver_18.4bd5
drwxr-xr-x  23 aldenhu  staff  736 Aug  5 16:27 transaction_14-.c20a
$ cat metadata/*
{"StateSnapshotBackup":{"version":18,"manifest":"/var/folders/hb/d4b0grts37307v_s4h4z0xwr0000gn/T/78c65908ac1d4e6ed7113b49de9d1716/state_ver_18.4bd5/state.manifest"}}
{"TransactionBackup":{"first_version":14,"last_version":23,"manifest":"/var/folders/hb/d4b0grts37307v_s4h4z0xwr0000gn/T/78c65908ac1d4e6ed7113b49de9d1716/transaction_14-.c20a/transaction.manifest"}}
```

### DIFF
--- a/storage/backup/backup-cli/src/backup_types/epoch_ending/backup.rs
+++ b/storage/backup/backup-cli/src/backup_types/epoch_ending/backup.rs
@@ -7,7 +7,7 @@ use crate::{
     storage::{BackupHandleRef, BackupStorage, FileHandle, ShellSafeName},
     utils::{
         backup_service_client::BackupServiceClient, read_record_bytes::ReadRecordBytes,
-        should_cut_chunk, GlobalBackupOpt,
+        should_cut_chunk, storage_ext::BackupStorageExt, GlobalBackupOpt,
     },
 };
 use anyhow::{anyhow, ensure, Result};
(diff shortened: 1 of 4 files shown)


lightmark:
don't understand the motivation.

msmouse:
@lightmark 
1. on retry if the backend can't do overwriting, using same name may lead to failure
2. even if overwriting is supported everywhere, it may leave behind stale fileS instead of a stale dir when you change settinigs across different runs (like chunk size); and it's harder to debug when it crashes repeatedly because you don't know if they crash at the same point.

PR #5495, opened 2020-08-06
[management] cleanup operational tool repeated code

gregnazario:


config/management/src/transaction.rs
```diff
@@ -0,0 +1,29 @@
...
+use crate::constants;
+use libra_secure_time::{RealTimeService, TimeService};
+use libra_types::{
+    account_address::AccountAddress,
+    chain_id::ChainId,
+    transaction::{RawTransaction, Script},
+};
+
+/// Builds a transaction based on all of our
+pub fn transaction(
```
JoshLind, line 13: Maybe "build_transaction" or "build_raw_transaction"?

config/management/genesis/src/command.rs
```diff
@@ -98,73 +98,47 @@ impl Command {
...
+            CommandName::CreateAndInsertWaypoint
+        )
     }
 
     pub fn create_waypoint(self) -> Result<Waypoint, Error> {
-        match self {
-            Command::CreateWaypoint(cmd) => cmd.execute(),
-            _ => Err(self.unexpected_command(CommandName::CreateWaypoint)),
-        }
+        execute_command!(self, Command::CreateWaypoint, CommandName::CreateWaypoint)
```
zekun000, line 109: do you need both arguments? there's a `impl From<&Command> for CommandName`

gregnazario: We actually do need both arguments.  The CommandName is the expected command

gregnazario: I've gone ahead and renamed it accordingly though to be less confusing.

zekun000: the second and third argument are always the command and its corresponding name, no?

gregnazario: Yeah, but the problem comes more with the way that macros are defined.

In this case the `path` `Command::CreateWaypoint` refers to the path of the type, but not an instantiation of the object, which would include the `Command::CreateWaypoint(cmd)`, where the `cmd` needs to be defined.  So, sadly `CommandName::from(Command::CreateWaypoint)` doesn't work, because it's not an actual instantiation of the enum. :(

gregnazario: If we find a better way we can come back and fix it later :)

PR #5500, opened 2020-08-06
[cluster-test] Enable LSR/Vault by default in cluster test

andll:
This will enable lsr by default, when used with cti or in land blocking test

Please accept but do not land, some final testing is pending


changyonghu:
My change yesterday will call function in bootstrap_db and print quite a lot info! logs. Should we change info! to debug!?

andll:
@changyonghu yes, I think any O(num_of_nodes) output ideally should be at debug level, while info only for high level process

changyonghu:
> @changyonghu yes, I think any O(num_of_nodes) output ideally should be at debug level, while info only for high level process

Ok. I will look at it and see which info! are called too much.

PR #5501, opened 2020-08-06
[cluster-test] Use validator config template to generate node config.

changyonghu:
## Test Plan
./scripts/cti --pr 5501 --run bench
./scripts/cti --pr 5501 --run bench --enable-lsr false
./scripts/cti --pr 5501 --run bench --fullnodes-per-validator 0

### DIFF
--- a/docker/validator-dynamic/docker-run-dynamic.sh
+++ b/docker/validator-dynamic/docker-run-dynamic.sh
@@ -42,7 +42,7 @@ fi
     ${params[@]}
 
 
-if [ -n "${CFG_FULLNODE_SEED}" ]; then # We have a full node seed, add fullnode network
+if [ -n "${CFG_FULLNODE_SEED}" -a "${CFG_ENABLE_MGMT_TOOL}" = false ]; then # We have a full node seed, add fullnode network
 	declare -a fullnode_params
 	    fullnode_params+="-s ${CFG_FULLNODE_SEED} "
 	    fullnode_params+="-a /ip4/${CFG_LISTEN_ADDR}/tcp/6181 "
--- a/testsuite/cluster-test/src/cluster_swarm/cluster_swarm_kube.rs
+++ b/testsuite/cluster-test/src/cluster_swarm/cluster_swarm_kube.rs
@@ -27,13 +27,14 @@ use crate::instance::{
 use itertools::Itertools;
 use k8s_openapi::api::batch::v1::Job;
 use kube::api::ListParams;
-use libra_config::config::DEFAULT_JSON_RPC_PORT;
+use libra_config::config::{
+    NodeConfig, PersistableConfig, SafetyRulesConfig, DEFAULT_JSON_RPC_PORT,
+};
 use reqwest::Client as HttpClient;
-use std::{collections::HashSet, convert::TryFrom, process::Command};
-
 use rusoto_core::Region;
 use rusoto_s3::{PutObjectRequest, S3Client, S3};
 use rusoto_sts::WebIdentityProvider;
+use std::{collections::HashSet, convert::TryFrom, process::Command};
 
 const DEFAULT_NAMESPACE: &str = "default";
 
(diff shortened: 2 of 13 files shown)


testsuite/cluster-test/src/cluster_swarm/cluster_swarm_kube.rs
```diff
@@ -525,19 +530,34 @@ impl ClusterSwarmKube {
...
+            pod_name, node.name
+        );
         match &instance_config.application_config {
-            Validator(_) => {
-                self.config_fluentbit("validator-events", pod_name.as_str(), &node.name)
-                    .await?
+            Validator(validator_config) => {
+                self.config_fluentbit("validator-events", &pod_name, &node.name)
+                    .await?;
+                if validator_config.enable_mgmt_tool {
```
snakech, line 541: if it does not have order have to after config_fluentbit, u may can move this if condition outside. Otherwise I think its good.

testsuite/cluster-test/src/cluster_swarm/cluster_swarm_kube.rs
```diff
@@ -525,19 +530,34 @@ impl ClusterSwarmKube {
...
+                    self.generate_config(&instance_config, &pod_name, &node.name)
+                        .await?;
+                }
             }
             LSR(_) => {
-                self.config_fluentbit("safety-rules-events", pod_name.as_str(), &node.name)
-                    .await?
+                self.config_fluentbit("safety-rules-events", &pod_name, &node.name)
+                    .await?;
+                self.generate_config(&instance_config, &pod_name, &node.name)
```
snakech, line 559: for LSR we always generate_config no matter ```enable_mgmt_tool```?

testsuite/cluster-test/src/cluster_swarm/cluster_swarm_kube.rs
changyonghu: This if is only used to gate the case LSR/Vault is enabled or not. So I think it should be deleted.

testsuite/cluster-test/src/cluster_swarm/cluster_swarm_kube.rs
changyonghu: enable_mgmt_tool is used to gate LSR/Vault. In the case of no LSR, this arm is not called.

docker/validator-dynamic/docker-run-dynamic.sh
```diff
@@ -42,7 +42,7 @@ fi
     ${params[@]}
 
 
-if [ -n "${CFG_FULLNODE_SEED}" ]; then # We have a full node seed, add fullnode network
+if [ -n "${CFG_FULLNODE_SEED}" -a "${CFG_ENABLE_MGMT_TOOL}" = false ]; then # We have a full node seed, add fullnode network
```
davidiw, line 45: was this tested? cc: @snakech

andll: I was not there when this was merged, but FWIW this file is no longer used by cluster test, so if it breaks something, feel free to just revert the change

snakech: I confirmed with @changyonghu he did the test. And since we do not use config builder in CT anymore, I guess its ok to remove this flag

changyonghu: @davidiw  CT uses this file before, since @andll already removed config_builder in CT, it's ok to remove this flag.

changyonghu: I used this flag to ensure both new version (with mgmt tool) and old version works.

snakech: > @davidiw CT uses this file before, since @andll already removed config_builder in CT, it's ok to remove this flag.

gosh, how do you pop up here? for after sale service I assume aha

davidiw: Thanks for following up. The problem with this check is that it breaks compatibility with tools that are unaware of the change. The safer approach would be to just check if it is set or not. Now if setting it is mandatory, the alternative would be = false or not set. See https://github.com/libra/libra/pull/5793/files

changyonghu: Good to know. Not very familiar with bash. I thought it was set to false in default.

davidiw: Bash is a very painful experience... :D

PR #5503, opened 2020-08-06
[vm] Rewrite the PublishOption to become a native move resource

runtian-zhou:
## Motivation

Rewrite the logic for `VMPublishingOption` to become a native move resource:

Previously `VMPublishingOption` is stored as a lcs serialized byte on chain. This is really error prone and brings trouble for maintainence purpose as it's really hard to specify what gets modified over there from move land. This PR rewrote the logics over there to make it become a pure move resource.

Yes

## Test Plan

Modified the flow for existing tests. Add an extra e2e test to test the append script hash flow.


testsuite/tests/libratest/smoke_test.rs
```diff
@@ -1141,15 +1141,6 @@ fn test_e2e_modify_publishing_option() {
         .disable_custom_script(&["disable_custom_script"], true)
         .unwrap();
 
-    // mint another 10 coins after restart
```
runtian-zhou, line 1144: This is a pretty unfortunate consequence of this PR: If a network tries to flip from custom script execution to closed allowlist, no scripts could be executed at all since there were nothing in the allowlist.

bob-wilson:
Is this really necessary now? We're trying to stabilize prior to launching the v1 release, and at this point, merely being error prone doesn't seem like a good enough reason to justify a breaking change like this. Is there any reason we can't wait and do this post-v1?

jolestar:
How about moving the publishing_option and allow_scripts check logic to Move prologue?

runtian-zhou:
> Is this really necessary now? We're trying to stabilize prior to launching the v1 release, and at this point, merely being error prone doesn't seem like a good enough reason to justify a breaking change like this. Is there any reason we can't wait and do this post-v1?

The main issue here is that this change won't be backward compatible later. So we either need to make the call now or we would make that as it is until we replace our LibraVMConfig module with LibraVMConfigV2 module.

language/stdlib/modules/LibraVMConfig.move
```diff
@@ -3,15 +3,30 @@ address 0x1 {
...
         gas_schedule: GasSchedule,
     }
 
+    struct PublishingOption {
+        // Only script hashes in the following list can be executed by the network.
+        script_allow_list: vector<vector<u8>>,
+        // VM can execute arbitrary script provided by the user if this flag is set to true.
+        is_open_script: bool,
+        // Transactions with sender in this list won't abide to the allow list rules above.
+        sender_allow_list: vector<address>,
```
sblackshear, line 25: I think this probably should be a role-based check in the transaction prologue instead. i.e.: 
- kill this field
- check if the `&signer` is LR in the prologue, do allowlist check if not

language/stdlib/modules/LibraVMConfig.move
```diff
@@ -3,15 +3,30 @@ address 0x1 {
...
-        publishing_option: vector<u8>,
+        publishing_option: PublishingOption,
         gas_schedule: GasSchedule,
     }
 
+    struct PublishingOption {
+        // Only script hashes in the following list can be executed by the network.
+        script_allow_list: vector<vector<u8>>,
+        // VM can execute arbitrary script provided by the user if this flag is set to true.
+        is_open_script: bool,
```
sblackshear, line 23: What did you think of @tzakian's suggestion to represent this with an empty `script_allow_list`?

language/stdlib/modules/LibraVMConfig.move
```diff
@@ -3,15 +3,30 @@ address 0x1 {
...
+        // Transactions with sender in this list won't abide to the allow list rules above.
+        sender_allow_list: vector<address>,
+        // Anyone can publish new module if this flag is set to true.
+        is_open_module: bool,
```
sblackshear, line 27: Nit: I find this name confusing without "publishing" in it; sounds like "openness" might be a quality of existing modules. Perhaps `module_publishing_allowed` or similar?

language/stdlib/modules/LibraVMConfig.move
```diff
@@ -99,12 +124,33 @@ module LibraVMConfig {
         pragma aborts_if_is_partial = true; // TODO: added for a module property. Remove this once the "aborts_if" spec is completely specified.
     }
 
-    public fun set_publishing_option(account: &signer, publishing_option: vector<u8>) {
+    // Add `new_hash` to the list of script hashes that is allowed to be executed by the network.
+    public fun append_script_allow_list(account: &signer, new_hash: vector<u8>) {
```
sblackshear, line 128: Nit: `add_script? `append` sounds like we are adding another list of scripts to the existing allowlist.

language/stdlib/modules/LibraVMConfig.move
```diff
@@ -3,15 +3,30 @@ address 0x1 {
 module LibraVMConfig {
     use 0x1::LibraConfig;
     use 0x1::Roles;
+    use 0x1::Vector;
+
+    const SCRIPT_HASH_LENGTH: u64 = 32;
+    const ESCRIPT_HASH_LENGTH_NOT_MATCH: u64 = 0;
```
sblackshear, line 9:
```suggestion
    const EINVALID_SCRIPT_HASH: u64 = 0;
```

language/stdlib/modules/LibraVMConfig.move
```diff
@@ -99,12 +124,33 @@ module LibraVMConfig {
         pragma aborts_if_is_partial = true; // TODO: added for a module property. Remove this once the "aborts_if" spec is completely specified.
     }
 
-    public fun set_publishing_option(account: &signer, publishing_option: vector<u8>) {
+    // Add `new_hash` to the list of script hashes that is allowed to be executed by the network.
+    public fun append_script_allow_list(account: &signer, new_hash: vector<u8>) {
+        assert(Vector::length(&new_hash) == SCRIPT_HASH_LENGTH, ESCRIPT_HASH_LENGTH_NOT_MATCH);
+
+        let current_config = LibraConfig::get<LibraVMConfig>();
+        if(Vector::contains(&current_config.publishing_option.script_allow_list, &new_hash)) {
```
sblackshear, line 132:
```suggestion
        if (Vector::contains(&current_config.publishing_option.script_allow_list, &new_hash)) {
```

language/move-lang/functional-tests/tests/on_chain_config/change_script_allowlist.move
```diff
@@ -21,13 +21,12 @@ module FooConfig {
...
 //! sender: libraroot
-//! args: x"0100010000000000000000000000000a550c18",
 // Step 2: Change option to CustomScript
 script {
 use 0x1::LibraVMConfig;
 
-fun main(config: &signer, args: vector<u8>) {
-    LibraVMConfig::set_publishing_option(config, args)
+fun main(config: &signer) {
+    LibraVMConfig::set_open_module(config, false)
```
sblackshear, line 29: Intended? This doesn't match the comment on line 24

language/stdlib/compiled/src/transaction_scripts.rs
```diff
@@ -49,6 +49,7 @@ pub enum StdlibScript {
     RotateDualAttestationInfo,
     RotateSharedEd2551PublicKey,
     UpdateAccountLimitWindowInfo,
+    SetOpenScript,
```
sblackshear, line 52: We can probably omit a script for this, since it's something we'll do only once. I think a good rule of thumb is: if it's a LR-only operation that we will only do once, no need for a script.

sblackshear:
> Is this really necessary now? We're trying to stabilize prior to launching the v1 release, and at this point, merely being error prone doesn't seem like a good enough reason to justify a breaking change like this. Is there any reason we can't wait and do this post-v1?

@bob-wilson: the primary motivation is simplifying the AOS flow for adding a new tx script to the allowlist. Today, that operation is quite hard because AOS must write code to download the uninterpreted `PublishingOptions` blob from on-chain, parse it in Rust, figure out which part is the script allowlist, add a new hash to that, reserialize it, and then create a transaction that resets the entire blob.

After this change, the AOS flow is just: take the new script hash and pass it as an argument to the relevant tx script (no need to read on-chain state or understand how to parse `PublishingOptions`.

language/stdlib/compiled/src/transaction_scripts.rs
runtian-zhou: The problem is that we have a number of locations in the codebase that uses this script. If we don't use this approach then we need to have those users to import move compiler to compile the scripts, which I'm not sure how easy would that be.

language/move-lang/functional-tests/tests/on_chain_config/change_script_allowlist.move
runtian-zhou: That's intended, cuz the functional test first set the publishing option to be open script and we would like to then disable module publishing. Lemme update the comment

language/stdlib/modules/LibraVMConfig.move
moeziniaf: +1 on this name changing 🥇

language/stdlib/modules/LibraVMConfig.move
runtian-zhou: I don't hink that's doable at the moment. The reason is that the logic should be:
If sender in list, we don't perform the hash allow list check.

Th logic should live entirely in either rust land or in move land. If we would like to do it in the prologue, we would need to pass other info like if this txn is trying to publish a module/script hash to the prologue

language/stdlib/modules/LibraVMConfig.move
runtian-zhou: That would actually simplify the settings IMHO as long as this seems fine to you. I'm a bit hesitated just I view it as a wrapper for Option<Vec<HashValue>>, so I'm more inclined to use a boolean field. On the other hand I'm ok with the suggestion.

runtian-zhou: Looking at the code it might be a bit trickier to change. The reason is that we have a smoke test suite to test the behavior from changing the network from arbitrary script execution to a limited mode. I think the bigger issue here is to come up with the state machine between aribitrary script execution and limited execution mode. Thought? @tzakian

language/stdlib/compiled/src/transaction_scripts.rs
sblackshear: Ah, I see that. Maybe we could export a constant for the script binary (similar to what we do [here](https://github.com/libra/libra/blob/master/language/e2e-tests/src/common_transactions.rs#L17)?). I know this is somewhat of a pain, but I am very hesitant to clutter the public allowlist/API docs/generated script bindings with test-only scripts. I have been on a personal crusade to make `StdlibScript` as small as possible by refactoring to delete other test-only scripts...

tzakian: Would adding `pub`'s to these constants be workable for this? https://github.com/libra/libra/blob/master/language/transaction-builder/generated/src/stdlib.rs#L1744?

One thing we will want to do after this PR is separate out "admin" scripts that actually need/should be on the allowlist out, and remove those that no longer need to be there.

sblackshear: > One thing we will want to do after this PR is separate out "admin" scripts that actually need/should be on the allowlist out, and remove those that no longer need to be there.

I think this will address a lot of my concerns (e.g., it's perfect for `create_parent_vasp`). But I'm still not sure if we want the `set_open_script` in the admin script bindings because it's a special operation that we'll only do once at a point far in the future. Thoughts on the "will only do it once" -> "don't clutter the bindings with it" rule of thumb?

language/stdlib/modules/LibraVMConfig.move
tzakian: Wouldn't changing from arbitrary to limited mode just be adding a number of transactions to the allowlist? 

The way I view the state diagram is something like:
* STATE 1: open scripts
* STATE 2: only allowed scripts

We would then have the following transitions.

* In STATE 1:
   - Add script to allowlist ~> TRANSITION(STATE 2)
* in STATE 2:
   - Add script to allowlist ~> TRANSITION(STATE 2)
   - Remove script to allowlist ~> if EMPTY(allowlist) TRANSITION(STATE 1) else TRANSITION(STATE 2)

language/stdlib/compiled/src/transaction_scripts.rs
tzakian: I agree that `set_open_script` doesn't need to be included in the admin scripts. The proposal I have in my head right now is something like:
```
transaction_scripts/
  |-> public/ # This contains public scripts that need to be in the allowlist (e.g., create_parent_vasp)
  |-> testing/ # Contains scripts used for testing only (both admin/non-admin)
```
Only the `public` scripts would be included in the API/external bindings. Exactly how we want to expose the `testing` scripts is a question that I'm still thinking through, but I think we have a number of decent options on it. 

The specific LR, TC, and admin scripts needed by AOS would then be separated out from this, and we should see about moving the building of those scripts to the AOS side of things if possible since that will be needed in the future I'm sure, would declutter the API, and hopefully make things simpler :) 

cc @moeziniaf to get his thoughts on this idea.

sblackshear: The `testing` idea seems great. It addresses all of my concerns and is also much nicer than the "constant that holds a binary" solution.

moeziniaf: Tim and I discussed offline exposing transaction_scripts in a more comprehensible way. I'm not sure exactly which scripts would fall under 'testing', but clearly any/all scripts which are for single client use (AOS) should be decoupled from the rest. (The rest could be broken down by Role of invocator).

IMO it doesn't even need to be 'only build these scripts from AOS',  since there is an argument to be made that the semantics of building move scripts and the modules they call should be tied to the language/ dir, but at least a logical separation between public/ vs aos/ scripts should be made.

Having a public API for flipping a boolean (once) seems like unnecessary clutter to me, but the question rides on the frequency of these one-off (non writeset) scripts and how we want to handle them. It seems that it could be useful to have these scripts saved somewhere (even on AOS) and then have an 'execute admin script X'  (not necessarily on behalf of another account?) where script X is operation data.

moeziniaf:
I would love to try and have this landed today as we have an AOS deadline tomorrow for all governance operations

language/e2e-tests/src/tests/on_chain_configs.rs
```diff
@@ -122,3 +126,70 @@ fn updated_limit_allows_txn() {
...
+        lr.transaction()
+            .script(encode_add_to_script_allow_list_script(
+                HashValue::sha3_256_of(random_script.as_ref()).to_vec(),
+                0,
+            ))
+            .sequence_number(1)
+            .sign(),
+    );
+
+    // Regular accounts cannot send arbitrary txn to the network.
```
tzakian, line 180: nit: update comment

language/stdlib/modules/LibraTransactionPublishingOption.move
```diff
@@ -0,0 +1,92 @@
...
+        module_publishing_allowed: bool,
+    }
+
+    public fun initialize(
+        lr_account: &signer,
+        script_allow_list: vector<vector<u8>>,
+        sender_allow_list: vector<address>,
+        module_publishing_allowed: bool,
+    ) {
+        assert(LibraTimestamp::is_genesis(), ENOT_GENESIS);
```
tzakian, line 34: Please change to `LibraTimestamp::assert_genesis()` now that the error refactor has landed. 

Possibly add checks on the LR account to make sure it's the right account as well (`Roles::assert_libra_root(lr_account)`). But I think this is enforced in `publish_new_config` so not totally necessary.

language/stdlib/modules/LibraTransactionPublishingOption.move
```diff
@@ -0,0 +1,92 @@
...
+    // Check if a sender can publish a module
+    public fun is_module_allowed(sender: &address): bool {
+        let publish_option = LibraConfig::get<LibraTransactionPublishingOption>();
+
+        publish_option.module_publishing_allowed || Vector::contains(&publish_option.sender_allow_list, sender)
+    }
+
+    // Add `new_hash` to the list of script hashes that is allowed to be executed by the network.
+    public fun add_to_script_allow_list(lr_account: &signer, new_hash: vector<u8>) {
+        // TODO: who is allowed to do this?
```
tzakian, line 62: The only option that makes sense here is LibraRoot (and I believe this is even being enforced in `LibraConfig::set`). So I would say add a `Roles::assert_libra_root(lr_account)` here.

language/stdlib/modules/LibraTransactionPublishingOption.move
```diff
@@ -0,0 +1,92 @@
...
+
+        let publish_option = LibraConfig::get<LibraTransactionPublishingOption>();
+        if (Vector::contains(&publish_option.script_allow_list, &new_hash)) {
+            return
+        };
+        Vector::push_back(&mut publish_option.script_allow_list, new_hash);
+
+        LibraConfig::set<LibraTransactionPublishingOption>(lr_account, publish_option);
+    }
+
```
tzakian, line 74: Do we want another function that will add a sender to the allowed senders list as well?

I doubt it will be used too often so a script for it isn't needed, but having the plumbing to allow it here seems like a good idea.

types/src/on_chain_config/vm_publishing_option.rs
```diff
@@ -0,0 +1,102 @@
...
+}
+
+impl OnChainConfig for VMPublishingOption {
+    const IDENTIFIER: &'static str = "LibraTransactionPublishingOption";
+
+    fn deserialize_into_config(bytes: &[u8]) -> Result<Self> {
+        let raw_publishing_option =
+            lcs::from_bytes::<VMPublishingOptionInner>(&bytes).map_err(|e| {
+                format_err!(
+                    "Failed first round of deserialization for VMConfigInner: {}",
```
tzakian, line 84:
```suggestion
                    "Failed first round of deserialization for VMPublishingOptionInner: {}",
```

language/stdlib/modules/LibraTransactionPublishingOption.move
```diff
@@ -0,0 +1,92 @@
...
+        let publish_option = LibraConfig::get<LibraTransactionPublishingOption>();
+
+        publish_option.module_publishing_allowed || Vector::contains(&publish_option.sender_allow_list, sender)
+    }
+
+    // Add `new_hash` to the list of script hashes that is allowed to be executed by the network.
+    public fun add_to_script_allow_list(lr_account: &signer, new_hash: vector<u8>) {
+        // TODO: who is allowed to do this?
+
+        assert(Vector::length(&new_hash) == SCRIPT_HASH_LENGTH, EINVALID_SCRIPT_HASH);
```
tzakian, line 64: After Wolfgang's error PR that just landed, could you update this to add the correct error category to it (see `Error.move`)?

language/stdlib/modules/LibraTransactionPublishingOption.move
```diff
@@ -0,0 +1,92 @@
...
+            }
+        );
+    }
+
+    // Check if sender can execute script with `hash`
+    public fun is_script_allowed(hash: &vector<u8>, sender: &address): bool {
+        let publish_option = LibraConfig::get<LibraTransactionPublishingOption>();
+
+        Vector::is_empty(&publish_option.script_allow_list)
+            || Vector::contains(&publish_option.script_allow_list, hash)
```
sblackshear, line 49:
```suggestion
            || Roles::has_libra_root_role(account)
```

language/stdlib/modules/LibraTransactionPublishingOption.move
```diff
@@ -0,0 +1,92 @@
...
+            || Vector::contains(&publish_option.sender_allow_list, sender)
+    }
+
+    // Check if a sender can publish a module
+    public fun is_module_allowed(sender: &address): bool {
+        let publish_option = LibraConfig::get<LibraTransactionPublishingOption>();
+
+        publish_option.module_publishing_allowed || Vector::contains(&publish_option.sender_allow_list, sender)
```
sblackshear, line 57:
```suggestion
        publish_option.module_publishing_allowed || Roles::has_libra_root_role(account)
```

language/stdlib/modules/LibraTransactionPublishingOption.move
```diff
@@ -0,0 +1,92 @@
...
+    public fun is_script_allowed(hash: &vector<u8>, sender: &address): bool {
+        let publish_option = LibraConfig::get<LibraTransactionPublishingOption>();
+
+        Vector::is_empty(&publish_option.script_allow_list)
+            || Vector::contains(&publish_option.script_allow_list, hash)
+            || Vector::contains(&publish_option.sender_allow_list, sender)
+    }
+
+    // Check if a sender can publish a module
+    public fun is_module_allowed(sender: &address): bool {
```
sblackshear, line 54:
```suggestion
    public fun is_module_allowed(account: &signer): bool {
```

language/stdlib/modules/LibraTransactionPublishingOption.move
sblackshear, line 62:
```suggestion
        assert(Roles::has_libra_root_role(lr_account), ENOT_LIBRA_ROOT)
```

language/stdlib/modules/LibraTransactionPublishingOption.move
```diff
@@ -0,0 +1,92 @@
...
+
+    // Add `new_hash` to the list of script hashes that is allowed to be executed by the network.
+    public fun add_to_script_allow_list(lr_account: &signer, new_hash: vector<u8>) {
+        // TODO: who is allowed to do this?
+
+        assert(Vector::length(&new_hash) == SCRIPT_HASH_LENGTH, EINVALID_SCRIPT_HASH);
+
+        let publish_option = LibraConfig::get<LibraTransactionPublishingOption>();
+        if (Vector::contains(&publish_option.script_allow_list, &new_hash)) {
+            return
```
sblackshear, line 68:
```suggestion
            abort EALLOWLIST_ALREADY_CONTAINS_SCRIPT
```

language/stdlib/modules/LibraAccount.move
```diff
@@ -857,10 +867,27 @@ module LibraAccount {
         txn_gas_price: u64,
         txn_max_gas_units: u64,
         txn_expiration_time: u64,
-        chain_id: u8
+        chain_id: u8,
+        transaction_type: u8,
```
sblackshear, line 871: Do we need this? It seems like we could get away with something like:

```
if (script_hash.is_empty() {
  // module tx logic
} else {
  // script tx logic
}
```

That will let us get rid of the `EPROLOGUE_UNEXPECTED_WRITESET` error code and the `transaction_type` parameter (both here and in the Rust `TransactionMetadata`).

language/stdlib/modules/LibraTransactionPublishingOption.move
sblackshear: This is a tricky one. I'm of two minds about this.

- In the long term, I think functions like this should take `&ModifyConfigCapability<Config>` instead of doing a role check
- In the short term, I think doing an explicit role check in all of the functions in this module that call `LibraConfig::set`probably makes sense for consistency. It's redundant, but this is the pattern we use elsewhere...

language/stdlib/modules/LibraTransactionPublishingOption.move
```diff
@@ -0,0 +1,92 @@
...
+    /// 1. No module publishing, only allowlisted scripts are allowed.
+    /// 2. No module publishing, custom scripts are allowed.
+    /// 3. Both module publishing and custom scripts are allowed.
+    /// We represent these as the following resource.
+    struct LibraTransactionPublishingOption {
+        // Only script hashes in the following list can be executed by the network. If the vector is empty, no
+        // limitation would be enforced.
+        script_allow_list: vector<vector<u8>>,
+        // Transactions with sender in this list won't abide to the allow list rules above.
+        sender_allow_list: vector<address>,
```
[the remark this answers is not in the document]
tzakian: Good point on using roles instead of address for this. Seems like a much more reliable mechanism than address so I'm all for this.

language/stdlib/modules/LibraAccount.move
runtian-zhou: My concern is if we add more types to TransactionPayload in the future. Having an explicit tag here allow us to have different prologue logic for each payload type in the future. What do you think?

language/stdlib/modules/LibraTransactionPublishingOption.move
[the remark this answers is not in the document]
junkil-park: @sblackshear, @tzakian , what "permission" in LIP-2 is related to this? Is it "PublishModule"?

PR #5504, opened 2020-08-06
[execution][logger] Change one log from info! to debug!.

changyonghu:
This log prints out a lot in cluster test running.

### DIFF
--- a/execution/executor/src/db_bootstrapper.rs
+++ b/execution/executor/src/db_bootstrapper.rs
@@ -155,7 +155,7 @@ pub fn calculate_genesis<V: VMExecutor>(
     );
 
     let committer = GenesisCommitter::new(executor, ledger_info_with_sigs)?;
-    info!(
+    debug!(
         "Genesis calculated: ledger_info_with_sigs {:?}, waypoint {:?}",
         committer.ledger_info_with_sigs, committer.waypoint,
     );


andll:
Is this really printed freaquently in cluster test?
I would imagine genesis only created once?

changyonghu:
> Is this really printed freaquently in cluster test?
> I would imagine genesis only created once?

I see. Let me check which function calls this.

changyonghu:
> > Is this really printed freaquently in cluster test?
> > I would imagine genesis only created once?
> 
> I see. Let me check which function calls this.

Ok. create-and-insert-waypoint will call generate_waypoint, which in turn calls calculate_genesis. I think this info! actually means waypoint calculated.

andll:
Is `create-and-insert-waypoint` something we have to call per-validator/node?

changyonghu:
> Is `create-and-insert-waypoint` something we have to call per-validator/node?

https://github.com/libra/partners/blob/master/testnet/testnet/templates/genesis.yaml#L88
In testnet, yes.

changyonghu:
> > Is `create-and-insert-waypoint` something we have to call per-validator/node?
> 
> https://github.com/libra/partners/blob/master/testnet/testnet/templates/genesis.yaml#L88
> In testnet, yes.

Yes, this is a bit confusing. Maybe we only need to call once. We don't need the waypoint.txt file?

andll:
Waypoint is same across all validators(its just description of the state of blockchain)

We might want to have `waypoint.txt` at some point on cluster test, but I am not sure why we need to create it more then once

changyonghu:
> Waypoint is same across all validators(its just description of the state of blockchain)
> 
> We might want to have `waypoint.txt` at some point on cluster test, but I am not sure why we need to create it more then once

Let me check with @mgorven

PR #5512, opened 2020-08-07
[move-prover] Prover changes and fix resulting from work on error model.

wrwg:
This PR pulls out some bug fixes and feature additions from the evolving errors PR. This is also so we do not have those changes in the PR which need to be marked as breaking.

- fixed a crash in error location handling of abort code verification failures.
- enabled lets in schemas. This was omitted previously, indeed lets also work in schemas.
- fixed a crash with aborts_if with a code in presence of schema expressions.
- printing error codes now in hex format, for better decoding.
- added a cli flag to report or not report warnings (like unused schemas). The default is now false until we come up with a better solution for schemas exported to other modules.
- Restricted the heuristic to find usages of a source to *only* consider `use` declarations. Previously, it was also looking for fully qualified names, but it has no  way to see whether they are in comments or not without some parsing. So previously, this was over approximated leading to ugly effects (e.g. compile a leaf module but
  get errors for parents). **NOTE**: we assume that the framework uses proper `use` clauses, and not absolute qualification.
- implemented shl/shr in spec expressions. This is for being able to specify FixedPoint32.
- implemented category-only code matching by truncating the relevant code to 8 bits.

## Motivation

New error model.

Yes

## Test Plan

Existing tests.

## Related PRs

#5463

### DIFF
--- a/language/move-prover/spec-lang/src/env.rs
+++ b/language/move-prover/spec-lang/src/env.rs
@@ -183,8 +183,10 @@ impl Loc {
         let mut start = loc.span.start();
         let mut end = loc.span.end();
         for l in locs.iter().skip(1) {
-            start = std::cmp::min(start, l.span().start());
-            end = std::cmp::max(end, l.span().end());
+            if l.file_id() == loc.file_id() {
+                start = std::cmp::min(start, l.span().start());
+                end = std::cmp::max(end, l.span().end());
+            }
         }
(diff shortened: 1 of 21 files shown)


wrwg:
@tzakian mostly to sign off this: "NOTE: we assume that the framework uses proper use clauses, and not absolute qualification."

PR #5514, opened 2020-08-07
[json-rpc] fix transaction type doc

xli:


andll:
/land


Sorry for the noise :)

PR #5516, opened 2020-08-07
[libra framework][move prover] enable calling of Move native functions from specs

emmazzz:
## Motivation

This PR
- Defines the spec version of most applicable Move public native functions in `prelude.bpl`,
- Enables the direct and indirect calling of aforementioned Move public native functions from specifications,
- Modifies specs to take advantage of the new feature.

Yes.

## Test Plan

Added a test case.

### DIFF
--- a/language/move-prover/spec-lang/src/translate.rs
+++ b/language/move-prover/spec-lang/src/translate.rs
@@ -1149,7 +1149,7 @@ impl<'env, 'translator> ModuleTranslator<'env, 'translator> {
 
         // Add $ to the name so the spec version does not name clash with the move version.
         let name = self.symbol_pool().make(&format!("${}", name.0.value));
-        let fun_decl = SpecFunDecl {
+        let mut fun_decl = SpecFunDecl {
             loc,
             name,
             type_params,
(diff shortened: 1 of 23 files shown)


language/move-prover/spec-lang/src/translate.rs
```diff
@@ -1160,8 +1160,12 @@ impl<'env, 'translator> ModuleTranslator<'env, 'translator> {
             used_memory: BTreeSet::new(),
             uninterpreted: false,
             is_move_fun: true,
+            is_native: false,
             body: None,
         };
+        if let EA::FunctionBody_::Native = def.body.value {
```
wrwg, line 1166: nit: `fun_decl.is_native = matches!(def.body.value, EA::FunctionBody_::Native)` is nicer. Or you use this right away in the constructor above and then do not need the `mut`.

language/move-prover/spec-lang/src/translate.rs
```diff
@@ -1536,8 +1553,14 @@ impl<'env, 'translator> ModuleTranslator<'env, 'translator> {
         let body = if self.spec_funs[spec_fun_idx].body.is_some() {
             std::mem::replace(&mut self.spec_funs[spec_fun_idx].body, None).unwrap()
         } else {
-            // No function body: the move function is already impure so no need to compute.
-            return false;
+            // If the function is native, consider it pure.
+            // Otherwise the function is non-native, its body cannot be parsed
+            // so we consider it impure.
+            // TODO(emmazzz) right now all the native move functions are considered pure.
```
wrwg, line 1559: Doesn't that break already now with some of the vector functions which take `&mut` parameters? We really want to avoid that user would learn this via crashes/boogie errors. Perhaps you can check the signature of the Move function whether it contains any `&mut`?

emmazzz: Right that's a good point. We should disallow them at least for now.

PR #5522, opened 2020-08-07
[logging] ensure TCP structured logging writes all data

gregnazario:
Previous writing system would write X number of bytes, and not guarantee
that the whole message is written.  This now guarantees that the whole
message is written by using the write_all function.

https://doc.rust-lang.org/std/io/trait.Write.html#method.write_all

### DIFF
--- a/common/logger/src/struct_log.rs
+++ b/common/logger/src/struct_log.rs
@@ -476,17 +476,13 @@ impl TCPStructLogThread {
             };
 
             // If we fail to write, exit out and create a new stream
-            if let Err(e) = stream.write(json_string.as_bytes()) {
+            if let Err(e) = stream.write_all(json_string.as_bytes()) {
                 STRUCT_LOG_SEND_ERROR_COUNT.inc();
-                println!(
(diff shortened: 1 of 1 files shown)


common/logger/src/struct_log.rs
```diff
@@ -476,18 +476,14 @@ impl TCPStructLogThread {
             };
 
             // If we fail to write, exit out and create a new stream
-            if let Err(e) = stream.write(json_string.as_bytes()) {
+            if let Err(e) = stream.write_all(json_string.as_bytes()) {
                 STRUCT_LOG_SEND_ERROR_COUNT.inc();
-                println!(
+                log::error!(
```
mgorven, line 481: I think this was intentional, so errors during logging wouldn't create a positive feedback loop

gregnazario: I wrote it that way, but I changed my mind and I'd rather it make the positive feedback loop, and write to text logs (maybe struct logs), than just have println! logs.

PR #5525, opened 2020-08-07
[mirai] add tag annotations to libra-crypto crate

stonebuddha:
## Motivation

This commit adds documentation to the code and helps reduce the false positives produced by MIRAI so that newly introduced problems would be easier to find. Some of the annotations make use of [tag analysis](https://github.com/facebookexperimental/MIRAI/blob/master/documentation/TagAnalysis.md) to track validity of public keys at compile time. (Note that #3846 explained the importance of public key validation.)

Yes.

## Test Plan

cargo xtest libra-crypto

### DIFF
--- a/crypto/crypto/src/ed25519.rs
+++ b/crypto/crypto/src/ed25519.rs
@@ -41,6 +41,7 @@ use crate::{
 use anyhow::{anyhow, Result};
 use core::convert::TryFrom;
 use libra_crypto_derive::{DeserializeKey, SerializeKey, SilentDebug, SilentDisplay};
+use mirai_annotations::*;
 use serde::Serialize;
 use std::{cmp::Ordering, fmt};
 
@@ -75,6 +76,11 @@ impl Clone for Ed25519PrivateKey {
 /// An Ed25519 public key
 #[derive(DeserializeKey, Clone, SerializeKey)]
 pub struct Ed25519PublicKey(ed25519_dalek::PublicKey);
+
+#[cfg(mirai)]
+use crate::tags::ValidatedPublicKeyTag;
+#[cfg(not(mirai))]
+struct ValidatedPublicKeyTag {}
 
 /// An Ed25519 signature
 #[derive(DeserializeKey, Clone, SerializeKey)]
@@ -378,7 +384,9 @@ impl TryFrom<&[u8]> for Ed25519PublicKey {
         // Unfortunately, tuple struct `PublicKey` is private so we cannot
         // Ok(Ed25519PublicKey(ed25519_dalek::PublicKey(compressed, point)))
         // and we have to again invoke deserialization.
-        Ed25519PublicKey::from_bytes_unchecked(bytes)
+        let public_key = Ed25519PublicKey::from_bytes_unchecked(bytes)?;
+        add_tag!(&public_key, ValidatedPublicKeyTag); // This key has gone through validity checks.
+        Ok(public_key)
     }
 }
 
@@ -410,6 +418,8 @@ impl Signature for Ed25519Signature {
         message: &T,
         public_key: &Ed25519PublicKey,
     ) -> Result<()> {
+        // Public keys should be validated to be safe against small subgroup attacks, etc.
+        precondition!(has_tag!(public_key, ValidatedPublicKeyTag));
         let mut bytes = <T::Hasher as CryptoHasher>::seed().to_vec();
         lcs::serialize_into(&mut bytes, &message)
             .map_err(|_| CryptoMaterialError::SerializationError)?;
@@ -420,6 +430,8 @@ impl Signature for Ed25519Signature {
     /// Outside of this crate, this particular function should only be used for native signature
     /// verification in move
     fn verify_arbitrary_msg(&self, message: &[u8], public_key: &Ed25519PublicKey) -> Result<()> {
+        // Public keys should be validated to be safe against small subgroup attacks, etc.
+        precondition!(has_tag!(public_key, ValidatedPublicKeyTag));
         Ed25519Signature::check_malleability(&self.to_bytes())?;
 
         public_key
--- a/crypto/crypto/src/lib.rs
+++ b/crypto/crypto/src/lib.rs
@@ -3,6 +3,8 @@
 
 #![forbid(unsafe_code)]
 #![deny(missing_docs)]
+//! This feature gets turned on only if libra-crypto is compiled via MIRAI in a nightly build.
+#![cfg_attr(mirai, allow(incomplete_features), feature(const_generics))]
 
 //! A library supplying various cryptographic primitives
 pub mod compat;
@@ -18,6 +20,9 @@ pub mod x25519;
 
 #[cfg(test)]
 mod unit_tests;
+
+#[cfg(mirai)]
+mod tags;
 
 pub use self::traits::*;
 pub use hash::HashValue;
@@ -49,3 +54,9 @@ compile_error!(
     "at most one dalek arithmetic backend cargo feature should be enabled! \
      please enable one of: fiat, vanilla"
 );
+
+// MIRAI's tag analysis makes use of the incomplete const_generics feature, so the module
+// containing the definitions of MIRAI tag types should not get compiled in a release build.
+// The code below fails a build of the crate if mirai is on but debug_assertions is not.
+#[cfg(all(mirai, not(debug_assertions)))]
+compile_error!("MIRAI can only be used to compile the crate in a debug build!");
--- a/crypto/crypto/src/multi_ed25519.rs
+++ b/crypto/crypto/src/multi_ed25519.rs
@@ -19,6 +19,7 @@ use crate::{
 use anyhow::{anyhow, Result};
 use core::convert::TryFrom;
 use libra_crypto_derive::{DeserializeKey, SerializeKey, SilentDebug, SilentDisplay};
+use mirai_annotations::*;
 use rand::Rng;
 use serde::Serialize;
 use std::{convert::TryInto, fmt};
@@ -42,6 +43,11 @@ pub struct MultiEd25519PublicKey {
     public_keys: Vec<Ed25519PublicKey>,
     threshold: u8,
 }
+
+#[cfg(mirai)]
+use crate::tags::ValidatedPublicKeyTag;
+#[cfg(not(mirai))]
+struct ValidatedPublicKeyTag {}
 
 /// Vector of the multi-key signatures along with a 32bit [u8; 4] bitmap required to map signatures
 /// with their corresponding public keys.
@@ -299,9 +305,13 @@ impl TryFrom<&[u8]> for MultiEd25519PublicKey {
             .chunks_exact(ED25519_PUBLIC_KEY_LENGTH)
             .map(Ed25519PublicKey::try_from)
             .collect();
-        public_keys.map(|public_keys| MultiEd25519PublicKey {
-            public_keys,
-            threshold,
+        public_keys.map(|public_keys| {
+            let public_key = MultiEd25519PublicKey {
+                public_keys,
+                threshold,
+            };
+            add_tag!(&public_key, ValidatedPublicKeyTag);
+            public_key
         })
     }
 }
@@ -474,6 +484,8 @@ impl Signature for MultiEd25519Signature {
         message: &T,
         public_key: &MultiEd25519PublicKey,
     ) -> Result<()> {
+        // Public keys should be validated to be safe against small subgroup attacks, etc.
+        precondition!(has_tag!(public_key, ValidatedPublicKeyTag));
         let mut bytes = <T as CryptoHash>::Hasher::seed().to_vec();
         lcs::serialize_into(&mut bytes, &message)
             .map_err(|_| CryptoMaterialError::SerializationError)?;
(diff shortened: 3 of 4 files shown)


stonebuddha:
@hermanventer @huitseeker @mimoo This commit could be a first step to apply tag analysis.

crypto/crypto/src/ed25519.rs
```diff
@@ -410,6 +431,7 @@ impl Signature for Ed25519Signature {
         message: &T,
         public_key: &Ed25519PublicKey,
     ) -> Result<()> {
+        precondition!(has_tag!(public_key, ValidPublicKeyTag)); // Public keys should be valid.
```
hermanventer, line 434: The comment does not add any value beyond what the annotation already provides.

crypto/crypto/src/noise.rs
```diff
@@ -645,8 +645,8 @@ impl NoiseSession {
             .encrypt_in_place_detached(nonce, b"", message)
             .map_err(|_| NoiseError::Encrypt)?;
 
-        // increment nonce
-        self.write_nonce += 1;
+        // increment nonce, which won't exceed the maximum u64 value in practice
```
hermanventer, line 648: This is a claim without evidence. Why?

mimoo: While this is true (because the nonce is initialized to 0 and never reaches `u64::max`), if it ever becomes false then this change will introduce a bug. I would propose another change: since this function returns a `Result`, let's use a `checked_add` and return an error if the addition wraps around.

crypto/crypto/src/ed25519.rs
mimoo: I for one appreciate these kind of comments : o (I also wouldn't mind the comment being on its own line)

crypto/crypto/src/ed25519.rs
```diff
@@ -378,7 +397,9 @@ impl TryFrom<&[u8]> for Ed25519PublicKey {
         // Unfortunately, tuple struct `PublicKey` is private so we cannot
         // Ok(Ed25519PublicKey(ed25519_dalek::PublicKey(compressed, point)))
         // and we have to again invoke deserialization.
-        Ed25519PublicKey::from_bytes_unchecked(bytes)
+        let key = Ed25519PublicKey::from_bytes_unchecked(bytes)?;
+        add_tag!(&key, ValidPublicKeyTag); // This key has gone through validity checks.
```
mimoo, line 401: Why not using a string here instead of defining a type `ValidPublicKeyTag`?

crypto/crypto/src/ed25519.rs
```diff
@@ -76,6 +77,24 @@ impl Clone for Ed25519PrivateKey {
...
+struct ValidPublicKeyTag {}
+
+/// A generic tag type intended to only be used by ValidPublicKeyTag
+#[cfg(feature = "const_generics")]
+struct ValidPublicKeyTagKind<const MASK: TagPropagationSet> {}
+
+/// The propagation set of ValidPublicKeyTag. An empty propagation is used to make sure that
+/// ValidPublicKeyTag can only be explicitly attached to public keys.
+#[cfg(feature = "const_generics")]
+const VALID_PUBLIC_KEY_TAG_MASK: TagPropagationSet = tag_propagation_set!();
```
mimoo, line 96: not sure I understand what this really does :D

huitseeker: It prevents propagating the tag (which presence marks a checked public key) through e.g. equality. 
Though in all rigor, cloning and equality should propagate `ValidPublicKeyTag`.
 
https://github.com/facebookexperimental/MIRAI/blob/master/documentation/TagAnalysis.md#declaring-tags

crypto/crypto/src/ed25519.rs
```diff
@@ -76,6 +77,24 @@ impl Clone for Ed25519PrivateKey {
 #[derive(DeserializeKey, Clone, SerializeKey)]
 pub struct Ed25519PublicKey(ed25519_dalek::PublicKey);
 
+/// A private tag type that tracks if a public key is checked to protect against invalid point
```
huitseeker, line 80: Would it be envisionable to put those in their own module? A [`mod`](https://doc.rust-lang.org/reference/items/modules.html) declaration would go a long way towards unifying declarations with this unified functionality.

crypto/crypto/src/hash.rs
```diff
@@ -248,6 +248,7 @@ impl HashValue {
 
     /// Returns the `index`-th nibble.
     pub fn get_nibble(&self, index: usize) -> Nibble {
+        precondition!(index < HashValue::LENGTH);
```
huitseeker, line 251: Can you put this one in a separate (first) commit, please? That way we can manage tag analysis and bound checks independently (mentally, in history, in reverts, ...).

crypto/crypto/src/lib.rs
```diff
@@ -3,6 +3,7 @@
 
 #![forbid(unsafe_code)]
 #![deny(missing_docs)]
+#![cfg_attr(mirai, allow(incomplete_features), feature(const_generics))]
```
huitseeker, line 6: This is bound to scare people off: `const_generics` is something we do not want anywhere near production releases, and we really want to make it really easy to check that fact.

Here's one way I think this could be achieved: 
- having a *crate* feature in the crypto crate, 
- making it non-default, and 
- making it the condition for compilation of the distinct module that 
    + introduces the tag types, 
    + and therein turns on `allow(incomplete_features)` and `#![feature(const_generics)]`.

Then ideally the `has_tag` preconditions refer to the fully-qualified name for this module.

crypto/crypto/src/ed25519.rs
```diff
@@ -76,6 +77,24 @@ impl Clone for Ed25519PrivateKey {
...
+/// attacks, small subgroup attacks, and typos. This tag type is only used at compilation time.
+/// This type should remain private.
+#[cfg(feature = "const_generics")]
+type ValidPublicKeyTag = ValidPublicKeyTagKind<VALID_PUBLIC_KEY_TAG_MASK>;
+
+#[cfg(not(feature = "const_generics"))]
+struct ValidPublicKeyTag {}
```
huitseeker, line 87: That's an OK name, but I'd prefer something that grammatically conveys the "has-happened-before" concept, hence `ValidatedPublicKeyTag`

crypto/crypto/src/noise.rs
stonebuddha: As suggested by @huitseeker, in order to keep this PR tag-specific, I've separated new bound checks from this PR to another PR #5540.

crypto/crypto/src/ed25519.rs
stonebuddha: The current tagging mechanism requires explicitly defined tag types. These types can be parameterized by a propagation set, which specifies the set of operations (e.g., bit-xor, subcomponents) that can propagate the tag.

I'll make the tag definition more modular.

stonebuddha:
A quick reply to the first two questions in https://github.com/libra/libra/pull/5525#pullrequestreview-464472325:

> * I expect tags on trait methods are nowhere near checkable yet, is that correct?

If you mean that we annotate trait methods with contracts like "the result must have a validated tag," I think currently a possible way is to use the [design by contracts crate](https://gitlab.com/karroffel/contracts) with the `contract_trait` attribute.

> * is it possible to run cross-crate checks yet?

Yes. I'm also trying to annotate and analyze some other crates that make use of libra-crypto, such as safety-rules.

crypto/crypto/src/lib.rs
stonebuddha: It seems that `#![feature(const_generics)]` cannot be used at module level. I've tried to move the definitions of tag types to a separate tag crate, but I'm not sure how to make the tag crate only accessible to the libra-crypto crate.

huitseeker: Have you tried conditional compilation of the feature gate?

`#cfg_attr(all(feature = "mirai-analysis-on", [debug_assertions), feature(link_llvm_intrinsics))]`

crypto/crypto/src/ed25519.rs
mimoo: Ah I see, makes sense! From a quick glance I don't see a cloning here though: https://github.com/facebookexperimental/MIRAI/blob/8f0596f02b557c3e357767ee73c437d4a9bf8c5c/annotations/src/lib.rs#L33

crypto/crypto/src/lib.rs
mimoo: what's wrong with const_generics?

hermanventer: Nothing much, other than that is not stable. We are not using any strange features and the standard libraries are using constant generics as well. I expect that something close enough like this will become part of the stable language at some point.

huitseeker: `const_generics` is not only unstable, it's [incomplete](https://doc.rust-lang.org/nightly/nightly-rustc/rustc_feature/constant.INCOMPLETE_FEATURES.html). It's nowhere near stabilization. 

I'd urge you to consider the risk of allowing anybody to depend on the activation of such a feature (which, as it stands in the current state of the PR, is crate-wide).  In fact I'm going to formally request changes to reflect that.

bmwill: Given we build and run all of libra using a stable compiler, we can have some solace in the fact that if that feature happens to get turned on somehow, that compilation should fail.

hermanventer: Just to help me understand the issue better: With the code as it is in this PR, can the stable compiler compile Libra?

If so, what is wrong with the code as it is in this PR?

bmwill: Given the current [CI failures](https://circleci.com/gh/libra/libra/174477?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-checks-link) it seems like there's probably some more work to be done, specifically to ensure that our CI does not try to build/test with anything that enables this feature.

hermanventer: Indeed, the only way code that depends on this feature should ever be compiled is when the mirai flag is set, which is something the CI should never do.

stonebuddha: The current CI failure comes from the fact that tag type definitions are moved to a separate crate in the latest commit.

Originally, the tag types were defined inside the libra-crypto crate, so there was no CI failure.

hermanventer: It seems to me that doing things in the original way makes sense. I would be in favor of reverting to it.

stonebuddha: Reverted back to the original way, but moved tag definitions to a separate tags module, which gets compiled only if both mirai and debug_assestions are on.

Is is still necessary to have a crate feature like `mirai-tag-analysis`, rather than use the mirai identifier directly?

crypto/crypto/src/ed25519.rs
stonebuddha: It's a bit unfortunate that cloning is not yet handled properly (see this [issue](https://github.com/facebookexperimental/MIRAI/issues/611)), because essentially tags are stored as a model field of a struct, which is not transferred by cloning. I'm investigating this problem.

mimoo: equal means copy?

stonebuddha: Equal means equality checks. From the perspective of semantics, copying generates duplicates of a value, so it is assumed to always propagate tags.

stonebuddha: Cloning now tracks tag/model fields properly (see this [PR](https://github.com/facebookexperimental/MIRAI/pull/613)).

stonebuddha:
A reply to the last two questions in https://github.com/libra/libra/pull/5525#pullrequestreview-464472325:

> * while I have you here, how about adding tag checks to each element of the public key iterator argument  of `batch_verify`?

Currently MIRAI does not support quantification in preconditions (e.g., "all the keys in the input have the Validated tag").
We could instead use `verify!` macros, but they would probably flag many false positives.

> * if we're being super ambitious, those checks straightforwardly extend to MultiEd25519PublicKey, and both modules could share the same set of tag declarations (see my suggestion to open a module for those in PR comments)

Added annotations to MultiEd25519PublicKey.

Regarding https://github.com/libra/libra/pull/5525#discussion_r468105758:

Added conditional compilation to guard the usage of const_generics by both mirai and debug_assertions.
Is it still necessary to add an extra crate feature like `mirai-tag-analysis` as a third condition?

crypto/crypto/src/lib.rs
huitseeker: - the use of the `mirai` feature is fine,
- after having played around with the code, the current submodule for tags is great,
- I'd still love a comment above the `cfg_attr(mirai, allow(incomplete_features), ...` which states something like:
```
//! This feature gets turned on only if libra-crypto is compiled via MIRAI in a nightly build.
```

crypto/crypto/src/lib.rs
```diff
@@ -49,3 +54,9 @@ compile_error!(
     "at most one dalek arithmetic backend cargo feature should be enabled! \
      please enable one of: fiat, vanilla"
 );
+
+// MIRAI's tag analysis makes use of the incomplete const_generics feature, so the module
+// containing the definitions of MIRAI tag types should not get compiled in a release build.
+// The code below fails a build of the crate if mirai is on but debug_assertions is not.
+#[cfg(all(mirai, not(debug_assertions)))]
+compile_error!("MIRAI can only be used to compile the crate in a debug build!");
```
mimoo, line 62: is it really the right place for this assertion?

stonebuddha: It would be better if the assertion is implemented universally for the project. Any suggestions?

PR #5526, opened 2020-08-08
[fuzzing] downgrading version of libfuzzer-sys for lionhead

mimoo:


rexhoffman:
Huh weird that this is needed.   If it is needed, do we know why lionhead needs 0.3.2, and not 0.3.3?

Also deps we don't want to upgrade should be codified like this:   https://github.com/libra/libra/blob/master/.dependabot/config.yml#L7-L9

rexhoffman:
Read your chat thread this morning, merging this, but will make a pr to ensure dependabot would leave the dep alone.

bor r+

PR #5527, opened 2020-08-10
[breaking] set validator operator admin script

moeziniaf:
## Motivation

LR will set validator operators for owners at launch. We will use an admin script to execute this setting on behalf of the validator (and use a nonce to prevent replay).

## Test Plan

validator_set_operator_set_key_reconfigure

## Related PRs

see #5423 for rewriting how admin scripts work.
This is breaking since it changes the stdlib (an additional script at genesis)


language/stdlib/transaction_scripts/set_validator_operator_with_nonce_admin.move
```diff
@@ -0,0 +1,13 @@
+script {
+use 0x1::SlidingNonce;
+use 0x1::ValidatorConfig;
+use 0x1::ValidatorOperatorConfig;
+/// Set validator operator as 'operator_account' of validator owner 'account' (via Admin Script).
+/// operator_name should match expected from operator account. This script also
+/// takes `sliding_nonce`, as a unique nonce for this operation. See sliding_nonce.move for details.
+fun set_validator_operator_with_nonce_admin(lr_account: &signer, account: &signer, sliding_nonce: u64, operator_name: vector<u8>, operator_account: address) {
```
tzakian, line 8: nit: Line is long, so could you break up on to multiple lines?

language/stdlib/transaction_scripts/set_validator_operator_with_nonce_admin.move
```diff
@@ -0,0 +1,13 @@
+script {
+use 0x1::SlidingNonce;
+use 0x1::ValidatorConfig;
+use 0x1::ValidatorOperatorConfig;
+/// Set validator operator as 'operator_account' of validator owner 'account' (via Admin Script).
+/// operator_name should match expected from operator account. This script also
```
tzakian, line 6:
```suggestion
/// `operator_name` should match expected from operator account. This script also
```

language/stdlib/transaction_scripts/set_validator_operator_with_nonce_admin.move
```diff
@@ -0,0 +1,13 @@
...
+/// takes `sliding_nonce`, as a unique nonce for this operation. See sliding_nonce.move for details.
```
tzakian, line 7:
```suggestion
/// takes `sliding_nonce`, as a unique nonce for this operation. See `SlidingNonce.move` for details.
```

tzakian:
Hmm that's odd not sure why bors is commenting...cc @bmwill for this ^ in case it's a bug (or, is it a feature?).

bmwill:
oh yuck 😞 This means that the parsing for commands is a little too simple. It just looks at lines that start with `/` and since you have some that do its triggering

andll:
Just to reiterate on what happened here with cluster test:

We previously had an agreement that PR should have 'breaking' label, when it contains breaking change.

However, as we migrated to having automatic detection for breaking changes, this convention has changed - we now need **commit**, not pull request, to have `[breaking]` prefix, in order to indicate to LBT hook that it should run cluster test without compatibility test.

What happened in this PR is that LBT did not see `[breaking]` prefix on commit message, and therefore it was running compatibility test, which indeed failed because this PR changes genesis

PR #5529, opened 2020-08-10
[bytecode tools] Add normalized modules and compatibility checking

sblackshear:
This PR turns some team conversations we've have about module/resource linking/layout APIs and into some prototype code that can be used in tooling.
    
    - Define normalized representations of Module, Type, Struct, and Field structs + functions for creating each from its CompiledModule counterpart
    - Define check_compatibility(old: &Module, new: &Module) -> Compatibility, which identifies linking and struct layout incompatibilities. In essence, a linking incompatibility can be fixed with a WriteSet that updates dependent modules (but no changes to on-chain data). A layout incompatibility requires a data migration of structs published on-chain to the new format (but no changes to on-chain code).
    - Partial definition of can_update(old: &Module, new: &Module) -> bool + open question about preventing cyclic dependencies
    - Hooked up the compatibility checking to the stdlib build and manually tested the a bunch of change scenarios like the following (all work). Will figure out how to turn them into automated tests if we like this approach
   
```
    // Expected: linking incompatible
    struct S<T> { b: bool }
    struct S<T, T2> { b: bool }
    
    // Expected: linking incompatible
    struct S<T1, T2> { f: T1 }
    struct S<T1, T2> { f: T2 }
    
    // Expected: linking incompatible
    struct S { b: bool }
    // deleted S
    
    // Expected: layout incompatible
    struct S { b: bool }
    struct S { b: bool, b2: bool }
    
    // Expected: layout incompatible
    struct S { b: bool, b2: bool }
    struct S { b: bool }
    
    // Expected: linking incompatible
    public fun foo(x: u64): u64 { x }
    public fun foo(x: u64, y: u64): u64 { x + y }
    
    // Expected: compatible
    fun foo(x: u64): u64 { x }
    fun foo(x: u64, y: u64): u64 { x + y }
    
    // Expected: compatible
    public fun f(): u64 { 2 }
    public fun f(): u64 { 3 }
```

### DIFF
--- a/language/vm/src/normalized.rs
+++ b/language/vm/src/normalized.rs
@@ -1,0 +1,258 @@
+// Copyright (c) The Libra Core Contributors
+// SPDX-License-Identifier: Apache-2.0
+
+use crate::{
+    access::ModuleAccess,
+    file_format::{
+        CompiledModule, FieldDefinition, FunctionHandle, Kind, SignatureToken, StructDefinition,
+        StructFieldInformation, TypeParameterIndex,
+    },
+};
+use move_core_types::{
+    account_address::AccountAddress,
+    identifier::Identifier,
+    language_storage::{StructTag, TypeTag},
+};
+
+/// Defines normalized representations of Move types, fields, kinds, structs, functions, and
+/// modules. These representations are useful in situations that require require comparing
+/// functions, resources, and types across modules. This arises in linking, compatibility checks
+/// (e.g., "is it safe to deploy this new module without updating its dependents and/or restarting
+/// genesis?"), defining schemas for resources stored on-chain, and (possibly in the future)
+/// allowing module updates transactions.
+
+/// A normalized version of `SignatureToken`, a type expression appearing in struct or function
+/// declarations. Unlike `SignatureToken`s, `normalized::Type`s from different modules can safely be
+/// compared.
+#[derive(Clone, Debug, Eq, PartialEq)]
+pub enum Type {
+    Bool,
+    U8,
+    U64,
+    U128,
+    Address,
+    Signer,
+    Struct {
+        address: AccountAddress,
+        module: Identifier,
+        name: Identifier,
+        type_arguments: Vec<Type>,
+    },
+    Vector(Box<Type>),
+    TypeParameter(TypeParameterIndex),
+    Reference(Box<Type>),
+    MutableReference(Box<Type>),
+}
+
+/// Normalized version of a `FieldDefinition`. The `name` is included even though it is
+/// metadata that it is ignored by the VM. The reason: names are important to clients. We would
+/// want a change from `Account { bal: u64, seq: u64 }` to `Account { seq: u64, bal: u64 }` to be
+/// marked as incompatible. Not safe to compare without an enclosing `Struct`.
+#[derive(Clone, Debug, Eq, PartialEq)]
+pub struct Field {
+    pub name: Identifier,
+    pub type_: Type,
+}
+
+/// Normalized version of a `StructDefinition`. Not safe to compare without an associated
+/// `ModuleId` or `Module`.
+#[derive(Clone, Debug, Eq, PartialEq)]
+pub struct Struct {
+    pub name: Identifier,
+    pub kind: Kind,
+    pub type_parameters: Vec<Kind>,
+    pub fields: Vec<Field>,
+}
+
+/// Normalized version of a `FunctionDefinition`. Not safe to compare without an associated
+/// `ModuleId` or `Module`.
+#[derive(Clone, Debug, Eq, PartialEq)]
+pub struct FunctionSignature {
+    pub name: Identifier,
+    pub type_parameters: Vec<Kind>,
+    pub formals: Vec<Type>,
+    pub ret: Vec<Type>,
+}
+
+/// Normalized version of a `CompiledModule`: its address, name, struct declarations, and public
+/// function declarations.
+#[derive(Clone, Debug, Eq, PartialEq)]
+pub struct Module {
+    pub address: AccountAddress,
+    pub name: Identifier,
+    pub structs: Vec<Struct>,
+    pub public_functions: Vec<FunctionSignature>,
+}
+
+impl Module {
+    /// Extract a normalized module from a `CompiledModule`. The module `m` should be verified.
+    /// Nothing will break here if that is not the case, but there is little point in computing a
+    /// normalized representation of a module that won't verify (since it can't be published).
+    pub fn new(m: &CompiledModule) -> Self {
+        let structs = m.struct_defs().iter().map(|d| Struct::new(m, d)).collect();
+        let public_functions = m
+            .function_defs()
+            .iter()
+            .filter_map(|f| {
+                if f.is_public {
+                    Some(FunctionSignature::new(m, m.function_handle_at(f.function)))
+                } else {
+                    None
+                }
+            })
+            .collect();
+        Self {
+            address: *m.address(),
+            name: m.name().to_owned(),
+            structs,
+            public_functions,
+        }
+    }
+}
+
+impl Type {
+    /// Create a normalized `Type` for `SignatureToken` `s` in module `m`.
+    pub fn new(m: &CompiledModule, s: &SignatureToken) -> Self {
+        use SignatureToken::*;
+        match s {
+            Struct(shi) => {
+                let handle = m.struct_handle_at(*shi);
+                assert!(handle.type_parameters.is_empty(), "A struct with N type parameters should be encoded as StructModuleInstantiation with type_arguments = [TypeParameter(1), ..., TypeParameter(N)]");
+                Type::Struct {
+                    address: *m.address(),
+                    module: m.name().to_owned(),
+                    name: m.identifier_at(handle.name).to_owned(),
+                    type_arguments: Vec::new(),
+                }
+            }
+            StructInstantiation(shi, type_actuals) => Type::Struct {
+                address: *m.address(),
+                module: m.name().to_owned(),
+                name: m.identifier_at(m.struct_handle_at(*shi).name).to_owned(),
+                type_arguments: type_actuals.iter().map(|t| Type::new(m, t)).collect(),
+            },
+            Bool => Type::Bool,
+            U8 => Type::U8,
+            U64 => Type::U64,
+            U128 => Type::U128,
+            Address => Type::Address,
+            Signer => Type::Signer,
+            Vector(t) => Type::Vector(Box::new(Type::new(m, t))),
+            TypeParameter(i) => Type::TypeParameter(*i),
+            Reference(t) => Type::Reference(Box::new(Type::new(m, t))),
+            MutableReference(t) => Type::MutableReference(Box::new(Type::new(m, t))),
+        }
+    }
+
+    /// Return true if `self` is a closed type with no free type variables
+    pub fn is_closed(&self) -> bool {
+        use Type::*;
+        match self {
+            TypeParameter(_) => false,
+            Bool => true,
+            U8 => true,
+            U64 => true,
+            U128 => true,
+            Address => true,
+            Signer => true,
+            Struct { type_arguments, .. } => type_arguments.iter().all(|t| t.is_closed()),
+            Vector(t) | Reference(t) | MutableReference(t) => t.is_closed(),
+        }
+    }
+
+    pub fn into_type_tag(self) -> Option<TypeTag> {
+        use Type::*;
+        Some(if self.is_closed() {
+            match self {
+                Reference(_) | MutableReference(_) => return None,
+                Bool => TypeTag::Bool,
+                U8 => TypeTag::U8,
+                U64 => TypeTag::U64,
+                U128 => TypeTag::U128,
+                Address => TypeTag::Address,
+                Signer => TypeTag::Signer,
+                Vector(t) => TypeTag::Vector(Box::new(
+                    t.into_type_tag()
+                        .expect("Invariant violation: vector type argument contains reference"),
+                )),
+                Struct {
+                    address,
+                    module,
+                    name,
+                    type_arguments,
+                } => TypeTag::Struct(StructTag {
+                    address,
+                    module,
+                    name,
+                    type_params: type_arguments
+                        .into_iter()
+                        .map(|t| {
+                            t.into_type_tag().expect(
+                                "Invariant violation: struct type argument contains reference",
+                            )
+                        })
+                        .collect(),
+                }),
+                TypeParameter(_) => unreachable!(),
+            }
+        } else {
+            return None;
+        })
+    }
+}
+
+impl Field {
+    /// Create a `Field` for `FieldDefinition` `f` in module `m`.
+    pub fn new(m: &CompiledModule, f: &FieldDefinition) -> Self {
+        Field {
+            name: m.identifier_at(f.name).to_owned(),
+            type_: Type::new(m, &f.signature.0),
+        }
+    }
+}
+
+impl Struct {
+    /// Create a `Struct` for `StructDefinition` `def` in module `m`. Panics if `def` is a
+    /// a native struct definition.
+    pub fn new(m: &CompiledModule, def: &StructDefinition) -> Self {
+        let handle = m.struct_handle_at(def.struct_handle);
+        let fields = match &def.field_information {
+            StructFieldInformation::Native => panic!("Can't extract  for native struct"),
+            StructFieldInformation::Declared(fields) => {
+                fields.iter().map(|f| Field::new(m, f)).collect()
+            }
+        };
+        Struct {
+            name: m.identifier_at(handle.name).to_owned(),
+            kind: if handle.is_nominal_resource {
+                Kind::Resource
+            } else {
+                Kind::Copyable
+            },
+            type_parameters: handle.type_parameters.clone(),
+            fields,
+        }
+    }
+}
+
+impl FunctionSignature {
+    /// Create a `FunctionSignature` for `FunctionHandle` `f` in module `m`.
+    pub fn new(m: &CompiledModule, f: &FunctionHandle) -> Self {
+        FunctionSignature {
+            name: m.identifier_at(f.name).to_owned(),
+            type_parameters: f.type_parameters.clone(),
+            formals: m
+                .signature_at(f.parameters)
+                .0
+                .iter()
+                .map(|s| Type::new(m, s))
+                .collect(),
+            ret: m
+                .signature_at(f.return_)
+                .0
+                .iter()
+                .map(|s| Type::new(m, s))
+                .collect(),
+        }
+    }
+}
--- a/language/stdlib/src/compatibility.rs
+++ b/language/stdlib/src/compatibility.rs
@@ -1,0 +1,120 @@
+// Copyright (c) The Libra Core Contributors
+// SPDX-License-Identifier: Apache-2.0
+
+/// The result of a linking and layoutcompatibility check. Here is what the different combinations
+/// mean:
+/// `{ struct: true, struct_layout: true }`: fully backward compatible
+/// `{ struct_and_function_linking: true, struct_layout: false }`: Dependent modules that reference functions or types in this module may not link. However, fixing, recompiling, and redeploying all dependent modules will work--no data migration needed.
+/// `{ type_and_function_linking: true, struct_layout: false }`: Attempting to read structs published by this module will now fail at runtime. However, dependent modules will continue to link. Requires data migration, but no changes to dependent modules.
+/// `{ type_and_function_linking: false, struct_layout: false }`: Everything is broken. Need both a data migration and changes to dependent modules.
+pub struct Compatibility {
+    /// If false, dependent modules that reference functions or structs in this module may not link
+    pub struct_and_function_linking: bool,
+    /// If false, attempting to read structs previously published by this module will fail at runtime
+    pub struct_layout: bool,
+}
+
+impl Compatibility {
+    /// Return true if the two module s compared in the compatiblity check are both linking and
+    /// layout compatible.
+    pub fn is_fully_compatible(&self) -> bool {
+        self.struct_and_function_linking && self.struct_layout
+    }
+
+    /// Return compatibility assessment for `new_module` relative to old module `old_module`
+    pub fn check(old_module: &Module, new_module: &Module) -> Compatibility {
+        let mut struct_and_function_linking = true;
+        let mut struct_layout = true;
+
+        // module's name and address are unchanged
+        if old_module.address != new_module.address || old_module.name != new_module.name {
+            struct_and_function_linking = false;
+        }
+
+        // old module's structs are a subset of the new module's structs
+        for old_struct in &old_module.structs {
+            match new_module
+                .structs
+                .iter()
+                .find(|s| s.name == old_struct.name)
+            {
+                Some(new_struct) => {
+                    if new_struct.kind != old_struct.kind
+                        || new_struct.type_parameters != old_struct.type_parameters
+                    {
+                        // Declared kind and/or type parameters changed. Existing modules that depend on this struct will fail to link with the new version of the module
+                        struct_and_function_linking = false;
+                        // This does not change the struct layout, but it may leave some published
+                        // values "orphaned". For example: if
+                        // `resource struct S<T: copyable> { t : T }` is changed to
+                        // `resource struct S<T: resource> { t : T}`, code can no longer access
+                        // published values of type (e.g.) `S<u64>`.
+                    }
+                    if new_struct.fields != old_struct.fields {
+                        // Fields changed. Code in this module will fail at runtime if it tries to
+                        // read a previously published struct value
+                        // TODO: this is a stricter definition than required. We could in principle
+                        // choose to label the following as compatible
+                        // (1) changing the name (but not position or type) of a field. The VM does
+                        //     not care about the name of a field (it's purely informational), but
+                        //     clients presumably do.
+                        // (2) changing the type of a field to a different, but layout and kind
+                        //     compatible type. E.g. `struct S { b: bool }` to `struct S { b: B }`
+                        //     where B is struct B { some_name: bool }. For a more realistic example
+                        //     , you might want to change `public_key: vector<u8>` to
+                        //     `struct Ed25519PublicKey { bytes: vector<u8> }`.
+                        //     TODO: does this affect clients? I
+                        //     think not--the serialization of the same data with these two types
+                        //     will be the same.
+                        struct_layout = false
+                    }
+                }
+                None => {
+                    // Struct not present in new . Existing modules that depend on this struct will fail to link with the new version of the module.
+                    struct_and_function_linking = false;
+                    // Note: we intentionally do *not* label this a layout compatibility violation.
+                    // Existing modules can still successfully read previously published values of
+                    // this struct `Parent::T`. That is, code like the function `foo` in
+                    // ```
+                    // struct S { t: Parent::T }
+                    // public fun foo(a: addr): S { move_from<S>(addr) }
+                    // ```
+                    // in module `Child` will continue to run without error. But values of type
+                    // `Parent::T` in `Child` are now "orphaned" in the sense that `Parent` no
+                    // longer exposes any API for reading/writing them.
+                }
+            }
+        }
+
+        // old module's public functions are a subset of the new module's public functions
+        for function in &old_module.public_functions {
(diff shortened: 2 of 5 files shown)


language/vm/src/abi.rs
```diff
@@ -0,0 +1,302 @@
...
+}
+
+impl TypeABI {
+    /// Create a `TypeABI` for `SignatureToken` `s` in module `m`.
+    pub fn new(m: &CompiledModule, s: &SignatureToken) -> Self {
+        use SignatureToken::*;
+        match s {
+            Struct(shi) => {
+                let handle = m.struct_handle_at(*shi);
+                assert!(handle.type_parameters.is_empty(), "A struct with N type parameters should be encoded as StructModuleInstantiation with type_arguments = [TypeParameter(1), ..., TypeParameter(N)]");
```
sblackshear, line 226: @dariorussi / @vgao1996: double-checking that this is true

dariorussi: yes it's true (if that is what you mean) that the handle on a `Struct` has to be a non generic type and a handle on a `StructInstantiation` must be a generic type

language/vm/src/abi.rs
```diff
@@ -0,0 +1,302 @@
...
+#[derive(Clone, Debug, Eq, PartialEq)]
+pub struct FieldABI {
+    name: Identifier,
+    type_: TypeABI,
+}
+
+/// ABI for a struct declaration. Not safe to compare without an associated `ModuleId` or `ModuleABI`.
+#[derive(Clone, Debug, Eq, PartialEq)]
+pub struct StructABI {
+    name: Identifier,
```
matbd, line 51: Make the fields `pub`?

Note: I've come to think that hiding fields under getters in plain data objects (without particular internal invariants) is not so useful and in fact slightly detrimental in Rust. The builder pattern is arguably a lifesaver in Java where interior mutability is the default but that's not the case in Rust.

language/vm/src/abi.rs
```diff
@@ -0,0 +1,302 @@
...
+use move_core_types::{identifier::Identifier, language_storage::ModuleId};
+
+/// Defines the Application Binary Interface (ABI) for Move modules, the `check_compatibility`
+/// function for checking ABI compatibility of two modules, and the `can_update` function for
+/// determining whether a new version of a module can safely update the old one.
+
+/// A normalized version of `SignatureToken`, a type expression appearing in struct or function
+/// declarations. Unlike `SignatureToken`s, `TypeABI`s from different modules can safely be compared.
+#[derive(Clone, Debug, Eq, PartialEq)]
+pub enum TypeABI {
```
tnowacki, line 20: I'm not really a big fan of exposing all of these public structs/enums--especially without a clear goal. (perhaps there is a goal I'm missing)

To me, the only thing we need to expose is `pub fn can_update(old_module: &CompiledModule, new_module: &CompiledModule)`

This also makes writing tools over this a bit easier, since we can change the underlying representations without a ton of hassle

sblackshear: > It should be fairly straightforward to convert an instance of ModuleABI into an LCS struct that @matbd can use to generate Rust wrappers, clients can use to decode a resource published on-chain (e.g. @runtian-zhou's resource-viewer), etc. For both of these use-cases, we might want to have a DecoratedModuleABI that augments a bytecode ModuleABI with comments, parameter names, and other info that might be useful to a client or code generator.

I think it is useful for the *ABI enums to be public in order to support this ^. E.g., a client might read a resource blob of type `M::T` and then want to extract the ABI for `M`, read the ABI for `T` from there, and then use it to deserialize the blob into an LCS representation. 

> This also makes writing tools over this a bit easier, since we can change the underlying representations without a ton of hassle

Good point. I think a good goal for these enums is to make the underlying representation of each *ABI generic enough that we never have to change it unless the bytecode format adds a backward-compatible, ABI-relevant construct (e.g., new type, public constants). Are there enums or fields here that look too brittle to support this goal? `Identifier` and `ModuleId` jump out to me as bad choices in light of that, but the others look ok to me.

sblackshear: `TypeABI` and `FunctionABI` are also useful when you need to compare normalized representations of types or functions from different modules (e.g., during linking). I think this [code](https://github.com/libra/libra/blob/master/language/bytecode-verifier/src/dependencies.rs#L275) could basically be simplified to `TypeABI::new(t1) == `TypeABI::new(t2)`.

matbd: Perhaps we should precise what "exposing" means here. Changing the (subset of) ABI interfaces that are used for code generation will force maintainers of Libra clients to (fix or more likely pull the latest version of) the codegen tools. For now, it's only us maintaining Libra clients but in the future it might be a larger community.

sblackshear: >  Changing the (subset of) ABI interfaces that are used for code generation will force maintainers of Libra clients to (fix or more likely pull the latest version of) the codegen tools. 

Agreed that we should try to nail these down! The way I'm thinking of it is first defining the "raw" bytecode ABI (this PR), and then defining a "decorated ABI" that can wrap the bytecode API with metadata extracted from the source that's useful for clients (e.g., parameter names, type parameter names, comments). A Libra client should be able to codegen from a decorated ABI, and we can provide a simple utility for creating a dummy decorated ABI from a raw one.

I think of these (excellent!) YAML files you mention as defining the decorated ABI, and our task now is to figure out how to related the raw and the decorated. Some quick thoughts from skimming the YAML:

- `ArgumentABI` and `TypeArgumentABI: `name` shouldn't be part of the raw ABI (because it's not in the bytecode). But totally makes sense to include both in the decorated ABI
- The raw form of`TypeArgumentABI` should include a `KindABI`
- `ScriptABI` is a very interesting case.
  -- `name`/`doc` don't need to be part of the raw ABI (but could/should be part of the decorated ABI).
  -- In principle, the raw `ScriptABI` could be identical to the raw `FunctionABI`. The current `ScriptABI` is more restrictive in the arguments that it allows (which is reasonable, but not requred).
  -- Whether or not `code` should be part of the raw ABI is the hardest question. If we have a policy of only extending the script allowlist (i.e., never removing old script hashes), a change to `code` is not actually a breaking change (only a change to the args or type args is breaking). But I'm not sure if we want to commit ourselves to that, so treating changes to `code` as part of the raw ABI seems reasonable for no. Perhaps we could define the raw ScriptABI as raw FunctionABI + code?

dariorussi:
not a big fan of the `*ABI` name but I guess it depends on how those could evolve.
I think there may be quite a bit we can do from code inspection. Resources may not be published ever and struct may be private to a module making upgrade possible, I think.

Also dependencies are not clear to me. Not sure what we have and what we want to verify there

vgao1996:
In general I'm somewhat troubled by our use of the name "ABI" here. I'm following this definition: https://en.wikipedia.org/wiki/Application_binary_interface and when I see this abbreviation I tend to think about something like the `.o` files for C/C++. There's years of complaints about C++ not having a stable ABI and that means the object file generated by one compiler may necessarily link correctly by another linker/compiler. IMO this is what ABI compatibility is about. Rust currently doesn't even guarantee ABI compatibility between two versions of the same compiler.

The close analogy of ABI I naturally think about in Move is the file format itself as it defines exactly what everything means, in terms of bytes, hence the "binary" in "application binary interface". Like it's not about the content of a particular module at all, but rather the structure and encoding that govern all modules. 

I feel like we really need to distinguish ABI compatibility and API compatibility:
- ABI Compatibility: a compiler/interpreter being or not being able to use a certain version of the binary format. An ABI incompatible change would be, a new file format being no longer accepted by the current VM or a new VM no longer supporting and old version of the file format.
- API Compatibility: changes in some Move (bytecode) program making it still possible or no longer possible to compile/link with downstream client code (also in Move). I feel like API compatibility is really what we're trying to define here.

sblackshear:
> In general I'm somewhat troubled by our use of the name "ABI" here. I'm following this definition: https://en.wikipedia.org/wiki/Application_binary_interface and when I see this abbreviation I tend to think about something like the `.o` files for C/C++. There's years of complaints about C++ not having a stable ABI and that means the object file generated by one compiler may not necessarily link correctly with another linker/compiler. IMO this is what ABI compatibility is about. Another example is that Rust currently doesn't even guarantee ABI compatibility between two versions of the same compiler.
> 
> The close analogy of ABI I naturally think about in Move is the file format itself as it defines exactly what everything means, in terms of bytes, hence the "binary" in "application binary interface". Like it's not about the content of a particular module at all, but rather the structure and encoding that govern all modules.
> 
> I feel like we really need to distinguish ABI compatibility and API compatibility:
> 
> * ABI Compatibility: a compiler/interpreter being or not being able to use a certain version of the binary format. An ABI incompatible change would be, a new file format being no longer accepted by the current VM or a new VM no longer supporting an old version of the file format.
> * API Compatibility: changes in some Move (bytecode) program making it still possible or no longer possible to compile/link with downstream client code (also in Move). I feel like API compatibility is really what we're trying to define here. It also makes a lot sense to say you generate FFIs (foreign function interfaces) from the APIs -- just that the resulting endpoint is in another language.

Good point! It would be more appropriate to call these something like `*LinkingAPI`. The only one that I am not sure about is `StructABI`, which is serving double duty as a struct linking API (everything except `fields`) and a "struct API template" (`fields`). Not sure what to call the second part... `StructLayoutAPI`?

dariorussi:
> > In general I'm somewhat troubled by our use of the name "ABI" here. I'm following this definition: https://en.wikipedia.org/wiki/Application_binary_interface and when I see this abbreviation I tend to think about something like the `.o` files for C/C++. There's years of complaints about C++ not having a stable ABI and that means the object file generated by one compiler may not necessarily link correctly with another linker/compiler. IMO this is what ABI compatibility is about. Another example is that Rust currently doesn't even guarantee ABI compatibility between two versions of the same compiler.
> > The close analogy of ABI I naturally think about in Move is the file format itself as it defines exactly what everything means, in terms of bytes, hence the "binary" in "application binary interface". Like it's not about the content of a particular module at all, but rather the structure and encoding that govern all modules.
> > I feel like we really need to distinguish ABI compatibility and API compatibility:
> > 
> > * ABI Compatibility: a compiler/interpreter being or not being able to use a certain version of the binary format. An ABI incompatible change would be, a new file format being no longer accepted by the current VM or a new VM no longer supporting an old version of the file format.
> > * API Compatibility: changes in some Move (bytecode) program making it still possible or no longer possible to compile/link with downstream client code (also in Move). I feel like API compatibility is really what we're trying to define here. It also makes a lot sense to say you generate FFIs (foreign function interfaces) from the APIs -- just that the resulting endpoint is in another language.
> 
> Good point! It would be more appropriate to call these something like `*LinkingAPI`. The only one that I am not sure about is `StructABI`, which is serving double duty as a struct linking API (everything except `fields`) and a "struct API template" (`fields`). Not sure what to call the second part... `StructLayoutAPI`?

I think those should be called: `Module`, `Struct`, `Function`, `Field`, `Type` and whatever else is the typical name for a type system. The module/package defines what they are for. I find any other name intrusive. And that is my view of course ;) 
What those structs are used for is the definition of the API, which may be linking or something else. So there the name could help.
At the end we have (and want) a very canonical representation of parts of our type system (for lack of a better word).

sblackshear:
> I think there may be quite a bit we can do from code inspection. Resources may not be published ever and struct may be private to a module making upgrade possible.

Agreed--in the current setup where we control everything, we can write a much fancier static analyzer that determines whether an incompatible change actually causes problems in dependencies.

> Also dependencies are not clear to me. Not sure what we have and what we want to verify there

Not sure what this means--is this referring to cyclic dependencies?

dariorussi:
> > I think there may be quite a bit we can do from code inspection. Resources may not be published ever and struct may be private to a module making upgrade possible.
> 
> Agreed--in the current setup where we control all the code, we can write a much fancier static analyzer that determines whether an incompatible change actually causes problems in dependent modules/scripts.
> 
> > Also dependencies are not clear to me. Not sure what we have and what we want to verify there
> 
> Not sure what this means--is this referring to cyclic dependencies?

no cyclic (but also that somewhat), I mean that dependencies can alter the result of a yes/no response.
Let me put it this way (and this PR may do it already or be ready to) I think this should take a stdlib previous and new, and perform cross module analysis and evaluate the answer wrt all is known via code and structure.
We are doing already something here and the canonical representation is what we want. 
Basically this should not only look for changes in a given module, but also evaluating how those affect the overall stdlib (or whatever set of modules is provided).

Also I do not think that is a minor or later step. I think that is the only step that matters for now and possibly for quite some time. 
Single module upgrade is only partially interesting and related anyway to what upgrade means. I can imagine a system where specific entities have permission to change some code (applications, libraries, etc.) and the range of changes they can allow can vary....

dariorussi:
> > I think there may be quite a bit we can do from code inspection. Resources may not be published ever and struct may be private to a module making upgrade possible.
> 
> Agreed--in the current setup where we control all the code, we can write a much fancier static analyzer that determines whether an incompatible change actually causes problems in dependent modules/scripts.
> 
Not sure about the fancy static analyzer, but yeah a whole program analyzer is what - I think - we want.
Effectively things that should be "easy" to do:
- identify all storage related types (all global opcodes) and check compatibility for those
- checking that scripts do not change hash, and mark the set of functions called from scripts (basically marking the true public interface)
- identify changes from old to new, in the context of a single module, but as related to the entire code

I think that would be a much better starting point that the single module story we are trying to do here. Also, it would reuse a lot of what's here anyway

tnowacki:
```
// Expected: linking incompatible
fun foo(x: u64): u64 { x }
fun foo(x: u64, y: u64): u64 { x + y }
```
I think you wanted `public` on this one?

```
// Expected: linking compatible
public fun f(): u64 { 2 }
// deleted f
```
`s/compatible/incompatible/`?

```
// Expected: linking incompatible
public fun f<T: copyable>() {}
public fun f<T>() {}
```
Why is this linking incompatible?

language/vm/src/abi.rs
tnowacki: > Are there enums or fields here that look too brittle to support this goal? Identifier and ModuleId jump out to me as bad choices in light of that, but the others look ok to me.

>  I think this code could basically be simplified to `TypeABI::new(t1) == TypeABI::new(t2)`

I'm actually generally terrified of code like that. The "compatibility" question in some sense is a typing question. It looks really like classical inheritance overrides. 

```
// Expected: linking incompatible
public fun f<T: copyable>() {}
public fun f<T>() {}
```

Taking this example again, updating/"overriding" something to make it more general should be fine as it doesn't break any existing end points. And these sort of checks don't work under `==`

My worry for exposing these sorts of structs is that we lose the ability to tightly define "compatibility" as you might be doing checks with the structs/enums directly, instead of going through the normal API


--- 

I also just generally dislike APIs that look like `MyStruct::new(a, b).foo(c, d)` instead of just `my_foo(a, b, c, d)`. I think it tends to feel really leaky in terms of complexity. This distaste is likely influencing my initial reactions here :P

language/vm/src/abi.rs
```diff
@@ -0,0 +1,399 @@
...
+                .collect(),
+            ret: m
+                .signature_at(f.return_)
+                .0
+                .iter()
+                .map(|s| TypeABI::new(m, s))
+                .collect(),
+        }
+    }
+}
```
vgao1996, line 378: The way the constructors are exposed looks very similar to the old `views` we had and that resulted in a large number of clients using the file format and the views at the same time, which was fragile and took long to clean up. With that being said this could be fine... I'm not entirely sure if the two cases are comparable, just something for us to think about.

language/vm/src/abi.rs
vgao1996: I'd echo on Todd's dislike on public constructors. The way these constructors are exposed looks very similar to the old `views` we had and that resulted in a large number of clients using the file format and the views at the same time, which was fragile and took long to clean up. With that being said this may be fine... I'm not entirely sure if the two cases are comparable as views contain references and these don't.

language/vm/src/abi.rs
```diff
@@ -0,0 +1,399 @@
...
+    /// Return true if the two module ABIs compared in the compatiblity check are both linking and
+    /// layout compatible.
+    pub fn is_fully_compatible(&self) -> bool {
+        self.struct_and_function_linking && self.struct_layout
+    }
+}
+
+impl ModuleABI {
+    /// Return compatibility assessment for `new_module` relative to old module `self`
+    pub fn check_compatibility(&self, new_module: &ModuleABI) -> Compatibility {
```
vgao1996, line 111: I assume the preconditions or assumptions on the verification status are similar to what's described in the doc? To me that's also a crucial part of this story.

language/vm/src/abi.rs
```diff
@@ -0,0 +1,399 @@
...
+                    if new_struct.fields != old_struct.fields {
+                        // Fields changed. Code in this module will fail at runtime if it tries to
+                        // read a previously published struct value
+                        // TODO: this is a stricter definition than required. We could in principle
+                        // choose to label the following as compatible
+                        // (1) changing the name (but not position or type) of a field. The VM does
+                        //     not care about the name of a field (it's purely informational), but
+                        //     clients presumably do.
+                        // (2) changing the type of a field to a different, but layout and kind
+                        //     compatible type. E.g. `struct S { b: bool }` to `struct S { b: B }`
```
vgao1996, line 148: I don't think this should ever be allowed. Assume you already have a resource of this type stored on-chain. Changing the type will effectively transmute the data.

language/vm/src/abi.rs
```diff
@@ -0,0 +1,399 @@
...
+/// `{ struct: true, struct_layout: true }`: fully backward compatible
+/// `{ struct_and_function_linking: true, struct_layout: false }`: Dependent modules that reference functions or types in this module may not link. However, fixing, recompiling, and redeploying all dependent modules will work--no data migration needed.
+/// `{ type_and_function_linking: true, struct_layout: false }`: Attempting to read structs published by this module will now fail at runtime. However, dependent modules will continue to link. Requires data migration, but no changes to dependent modules.
+/// `{ type_and_function_linking: false, struct_layout: false }`: Everything is broken. Need both a data migration and changes to dependent modules.
+pub struct Compatibility {
+    /// If false, dependent modules that reference functions or structs in this module may not link
+    pub struct_and_function_linking: bool,
+    /// If false, attempting to read structs previously published by this module will fail at runtime
+    pub struct_layout: bool,
+}
```
vgao1996, line 99: I kinda get the desire of trying to be a little bit more specific on the exact kinds of incompatibility found, but I feel like this distinction is adding a lot more complexity to this story and I'm not sure if any client will  need to consume this information rather than a single bool. I also wonder if an `enum CompatibilityError` model will work better if we want to be specific.

language/vm/src/abi.rs
tnowacki: For visibility, some updates from the meeting:

This particular generics case is not a valid new version/override because the linker expects exact equality when comparing function signatures. That is, when a the bytecode verifier checks a function handle against the actual function definition, it expects them to match *exactly*. So there is no wiggle room here for public functions. 
---
Nonetheless, I would still have hesitations of exposing `==`, but significantly less so if we break this up into tools and passes as @dariorussi described

language/vm/src/abi.rs
sblackshear: > I don't think this should ever be allowed

I think I agree with you on this philosophically. But it's worth asking whether there are some cases where this is (a) safe from a layout perspective and (b) acceptable from a client perspective. For example: we had a recent bug where `public_key: vector<u8>` allowed an empty vector. It might be possible/desirable for us to allow `public_key: Ed25519PublicKey` where `struct Ed25519PublicKey { bytes: vector<u8> }` and the non-emptiness of `bytes` is enforced at both construction and read time.

To be clear: I think "no" is probably the right policy here, but perhaps worth thinking about the technical feasibility.

language/vm/src/abi.rs
sblackshear: If you look at `stdlib/main.rs`, it is consuming the separate bools. Basically, we care about each case individually because it tells us what we need to do from a deployment perspective (redeploy code, or redeploy genesis)?

Agreed that if we were using this only for single-module upgrade, a single bool is sufficient/preferable. The `can_update: bool` API is intended for this use-case.

language/vm/src/abi.rs
```diff
@@ -0,0 +1,399 @@
...
+    pub type_: TypeABI,
+}
+
+/// ABI for a kind constraint on a generic type parameter
+#[derive(Clone, Debug, Eq, PartialEq)]
+pub enum KindABI {
+    All,
+    Resource,
+    Copyable,
+}
```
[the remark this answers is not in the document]
sblackshear: Agreed that this looks a bit silly because this is identical to `file_format::Kind`. I did this because I wanted the normalized types to be completely independent from the types defined in file_format.rs. The discipline I'm going for is that these normalized types are at a higher level of abstraction than the file format and can tolerate non-semantic changes in the more low-level bytecode file format.

sblackshear:
Sorry for taking so long to reply to all of these great comments! Here is what I did:
- Rename abi.rs to normalized.rs and get rid of all uses of ABI. I left normalized in `vm/src` because I think these normalized types are broadly useful--I think it would be great to use them (or a variant) in the bytecode verifier linking phase and module update transactions (once we allow them).
- I pulled out the compatibility checking code into its own file and put it in `stdlib` where it is used.

I didn't address the "public fields/public constructors" concerns because it didn't seem like we had agreement on what was desirable there. But don't have strong opinions and am happy to make changes to that (now or in a follow-up!).

Let me know if there are other blocking concerns that I missed.

language/vm/src/abi.rs
vgao1996: That's an interesting and legit use case. My take is that if we're making an offline tool then we can go a little bit stricter as the user makes the final decision. However for the long run I'm still thinking about the possibility of having something like this running on-chain, providing some protection for writeset txns. Maybe in that case we could disallow this behavior by default but allow the user to override it explicitly?

sblackshear:
> is this ready to go in? I would be fine accepting and hacking on it myself once in.
> Overall can you explain why not putting all of this in tools?
> I am not clear about having the "abstraction" in the `vm` folder, I would think that too would be perfect in `tools`. Thoughts?

I think it is ready. I wanted to have it in `vm` because (1) I think the bytecode verifier and (when it eventually exists) module update code in the VM should use these normalized types, and the vm shouldn't depend on `tools` and (2) this feels conceptually similar to `access.rs` to me, which lives in `vm` and is easy to discover as a result (whereas I wouldn't immediately know to go looking for this in tools). But don't feel strongly and am happy to move it either before or after this PR.

vgao1996:
@sblackshear How about creating another top level directory for this "privileged tool" if we want to use it on-chain. Just look at the bytecode verifier for example. The normalized types can stay there for now and if there's need we could promote them to move core types.

@dariorussi BTW I feel like the name "vm" is really obsolete right now and should probably be renamed to something like `move-vm-binary`.

sblackshear:
> @sblackshear How about creating another top level directory for this "privileged tool" if we want to use it on-chain. Just look at the bytecode verifier for example. The normalized types can stay there for now and if there's need we could promote them to move core types.

Do we need a new top-level directory for a single file?

vgao1996:
Then I personally think move core types is a better place. While the normalized types are currently derived from the binary format, they don't necessarily have to. Similar to `TypeTag`, they are just general Move concepts that could in theory be created by another source.

sblackshear:
> Then I personally think move core types is a better place. While the normalized types are currently derived from the binary format, they don't necessarily have to. Similar to `TypeTag`, they are just general Move concepts that could in theory be created by another source.

Conceptually, that makes sense to me. But the constructors for the normalized types depend on `src/file_format.rs`, which itself depends on `move-core-types`. That cyclic dependency seems unfortunate to me, which is why I like `vm` as a home.

vgao1996:
So `vm` definitely needs to depend on `move-core-types` and IMO the normalized types themselves are abstract concepts that belong to `move-core-types`. We also have some functions that gets some concrete stuff (binary) and extract/reconstruct abstract concepts (norm. types). So it seems to me that the conceptually correct way to model this is to put the type definitions in `move-core-types` yet leaving the constructors in `vm`?

sblackshear:
> So `vm` definitely needs to depend on `move-core-types` and IMO the normalized types themselves are abstract concepts that belong to `move-core-types`. We also have some functions that gets some concrete stuff (binary) and extract/reconstruct abstract concepts (norm. types). So it seems to me that the conceptually correct way to model this is to put the type definitions in `move-core-types` yet leaving the constructors in `vm` or a new crate that depends on both `move-core-types` and `vm`?

That is fine by me in the long term. But there is a question to answer first that I'd prefer not to solve in this PR: how do we implement the normalized constructors separately from the definitions (a trait? random utility functions in the VM? something else)? I think once we actually have some code that uses these types and start wanting to construct them in other ways (e.g., from a TypeTag or from prover env types), the answer will be more clear. But I don't want to over-abstract or over-design before we understand this.

language/vm/src/normalized.rs
```diff
@@ -0,0 +1,288 @@
...
+/// want a change from `Account { bal: u64, seq: u64 }` to `Account { seq: u64, bal: u64 }` to be
+/// marked as incompatible. Not safe to compare without an enclosing `Struct`.
+#[derive(Clone, Debug, Eq, PartialEq)]
+pub struct Field {
+    pub name: Identifier,
+    pub type_: Type,
+}
+
+/// Normalized version of a kind constraint on a generic type parameter. At present, this is
+/// identical to file_format::Kind, though this may change in the future.
```
tnowacki, line 58: Then let's change it in the future? :) 

I think `pub use file_format::Kind` would be just as effective in keeping things separate and avoiding duplication (until we need the split)

language/vm/src/normalized.rs
```diff
@@ -0,0 +1,288 @@
...
+}
+
+/// Normalized version of a `CompiledModule`: its address, name, struct declarations, and public
+/// function declarations.
+#[derive(Clone, Debug, Eq, PartialEq)]
+pub struct Module {
+    pub address: AccountAddress,
+    pub name: Identifier,
+    pub structs: Vec<Struct>,
+    pub public_functions: Vec<FunctionSignature>,
```
tnowacki, line 93: I know right now we only need the function signatures, but why not just include all function information here? 

If we do this, we could maybe in the future rewrite the bytecode verifier to work over this "normalized" version rather than having to deal with the soup of indexes in the file format

language/vm/src/normalized.rs
```diff
@@ -0,0 +1,288 @@
...
+
+    /// Construct a `Kind` from the `is_nominal_resource` bool used in the bytecode file format's
+    /// `StructDefinition`
+    pub fn from_bool(is_nominal_resource: bool) -> Self {
+        if is_nominal_resource {
+            Kind::Resource
+        } else {
+            Kind::Copyable
+        }
+    }
```
tnowacki, line 287: My thoughts is to remove this as it feels misleading, but I could be wrong. 

The point I'm trying to make is that `is_nominal_resource` implies `Resource` but `Resource` does not imply `is_nominal_resource`

language/vm/src/normalized.rs
sblackshear: What other normalized function information does the bytecode verifier need? I would expect it to only need normalized function signatures and struct handles in the linking phase and for every other check to be intra-module (and thus not require a normalized representation).

language/vm/src/normalized.rs
sblackshear: Fair point; doesn't seem good to have this as a public API. I will make it private or inline it in the one place that it's used.

language/vm/src/normalized.rs
tnowacki: I don't see this code ever living somewhere it cannot refer to the `file_format`, so I don't really follow the issue

language/vm/src/normalized.rs
tnowacki: I was imagining something a lot bigger than what is currently here, which is really a complete AST-style transformation over the file format, so we'd basically inline all of the indexes with their target (whether it be via copying or by reference or some other memory indirection magic)

This was going back to what @dariorussi was saying in the meeting of having a one-stop-shop for writing tools over the file_format. But perhaps that is outside the scope of this PR

language/vm/src/normalized.rs
sblackshear: Fixed this.

language/vm/src/normalized.rs
sblackshear: I think this is very useful to have, but definitely not the goal of this PR. @emmazzz's stackless bytecode (or a variant) of it seems like what you want here--it normalizes away indexes and the operand stack and some other nasty stuff too.

PR #5531, opened 2020-08-10
[docker] Return wget to libra init docker

andll:


snakech:
may I ask the reason we removed wget before?

andll:
unit test failure (

mgorven:
Sorry, didn't realise this was used at runtime. I replaced it with `busybox wget`.

PR #5536, opened 2020-08-10
[libra framework] Add role checks for Validator & ValidatorOperator

tzakian:
Previously we weren't performing role checks when adding validator/validator operator metadata to an account. This adds these checks.

This PR also cleans up the tests so that:
* A validator role resource is published under a validator account in the functional tests
* Removes code for unhosted that stuck around during removal of it.


language/stdlib/modules/LibraSystem.move
```diff
@@ -158,6 +159,11 @@ module LibraSystem {
         operator_account: &signer,
         validator_address: address,
     ) acquires CapabilityHolder {
+        assert(
+            Roles::has_validator_operator_role(operator_account) ||
+            Roles::has_validator_role(operator_account),
```
sblackshear, line 164: I think this is always supposed to be an operator rather than the validator itself, but I could be wrong. But @valerini will know for sure!

tzakian: I think Lera is out this week, cc @davidiw who might know as well?

davidiw: Yeah, we don't envision the case for a validator to be able to reconfigure since only the validator operator should be able to modify the validator config (outside of setting the operator)

tzakian: Awesome! I'll update :)

PR #5538, opened 2020-08-10
[json-rpc] add content-length limit

xli:
Add content-length limit to json-rpc server.

### DIFF
--- a/json-rpc/src/tests/unit_tests.rs
+++ b/json-rpc/src/tests/unit_tests.rs
@@ -15,7 +15,7 @@ use futures::{
     },
     StreamExt,
 };
-use libra_config::utils;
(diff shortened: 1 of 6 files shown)


json-rpc/src/tests/unit_tests.rs
```diff
@@ -1318,3 +1326,11 @@ fn execute_batch_and_get_first_response(
         .remove(0)
         .unwrap()
 }
+
+fn gen_string(len: usize) -> String {
```
thefallentree, line 1330: https://docs.rs/rand/0.7.3/rand/trait.Rng.html#method.fill maybe faster? I don't actually know

xli: not sure, but prefered Alphanumeric string, all random may produce double quote, which may break json format. so will just keep this version

PR #5539, opened 2020-08-10
[move-prover] Test instrumenter wrapper per check

kkmc:
## Motivation

#3665 (https://github.com/libra/libra/issues/3665)

Initial implementation for 3 specification checks:
(1) Constant global resource field checker
(2) Constant predicate subexpression checker
(3) WriteRef checker
(4) Underspecified return value checker (this subsumes the WriteRef checker).

Brief summary of what each does:
(1) Injects an ensures statement checking if the value of a field at a particular instance at the top level of the global store is equal to its old value, which means it didn't change for that function. An error is flagged if all functions do not raise an error.
(2) Recursively traverse the subexpressions of a precondition and postcondition and assert the subexpression to find if it's always true or always false. Raise an error if any pass. This ignores the full expressions because these should always be true (although it will raise an error if you write `ensures p && q` instead of `ensures p; ensures q`.
(3) Iterates over all WriteRef instructions and adds a specification check for each with the source value replaced with an unconstrained temporary value. 
(4) Iterates over all return values in the boogie procedure and rewrites the code to write an unconstrained value to these return values. For structs, it iterates over the top level field (one specification check for each field). For primitives and vectors, it adds one specification check.

Other changes:
1. Added SpecCheck which is a pair of conditions to check and optionally a code that is mutated. By default the original code is used. 
2. Adds constructors in the spec-lang/src/ast.rs to build ast nodes.
3. Adds an annotation field to Loc to differentiate error reporting for same location. This enables specification checks to generate different warnings for code at the same location.

Yes.

## Test Plan

Existing testsuite.

## Related PRs

None.


language/move-prover/spec-lang/src/ast.rs
```diff
@@ -60,6 +65,8 @@ pub enum ConditionKind {
     Ensures,
     Requires,
     RequiresModule,
+    RequiresSpecCheck,
```
wrwg, line 68: Do we really need a new condition kind for this? This will leak all over the place where we are dealing with condition kinds. If it's needed what do these actually mean? Is it general or specific for one of those tests? In the later case, it wouldn't scale to introduce a new kind for each of those checks.

language/move-prover/spec-lang/src/ast.rs
```diff
@@ -135,6 +142,8 @@ impl std::fmt::Display for ConditionKind {
             Ensures => write!(f, "ensures"),
             Requires => write!(f, "requires"),
             RequiresModule => write!(f, "requires module"),
+            RequiresSpecCheck => write!(f, "assume"),
```
wrwg, line 145: This doesn't seem to be assume/assert. As this is used in error message, please use a more specific name.

language/move-prover/spec-lang/src/ast.rs
```diff
@@ -621,3 +638,74 @@ impl<'a> fmt::Display for ExpDisplay<'a> {
         Ok(())
     }
 }
+
+// =================================================================================================
+// # AST Expression Helpers
```
wrwg, line 643: Please move this into a new module ast_helpers.

Also you like to have a type `ExpBuilder` here, then wrap the module_env inside, and make all those functions have a `&self` parameter.

language/move-prover/spec-lang/src/ast.rs
```diff
@@ -621,3 +638,74 @@ impl<'a> fmt::Display for ExpDisplay<'a> {
...
+
+impl Exp {
+    fn new_node(module_env: &ModuleEnv<'_>, typ: Type) -> NodeId {
+        module_env.new_node(None, typ)
```
wrwg, line 647: Each node also needs a location. If you absolutely don't have one, there is a GlobalEnv::unknown_loc

language/move-prover/spec-lang/src/ast.rs
```diff
@@ -621,3 +638,74 @@ impl<'a> fmt::Display for ExpDisplay<'a> {
...
+    pub fn make_value_number(module_env: &ModuleEnv<'_>, i: BigInt) -> Exp {
+        let node_id = Self::new_node(module_env, NUM_TYPE.clone());
+        Exp::Value(node_id, Value::Number(i))
+    }
+    pub fn make_value_bool(module_env: &ModuleEnv<'_>, value: bool) -> Exp {
+        let node_id = Self::new_node(module_env, BOOL_TYPE.clone());
+        Exp::Value(node_id, Value::Bool(value))
+    }
+    // LocalVar
+    pub fn make_localvar(module_env: &ModuleEnv<'_>, name: &str, typ: Type) -> Exp {
```
wrwg, line 663: We use `ty` as a general convention.

language/move-prover/spec-lang/src/env.rs
```diff
@@ -140,11 +153,16 @@ pub const EXPORT_ENSURES_PRAGMA: &str = "export_ensures";
 pub struct Loc {
     file_id: FileId,
     span: Span,
+    annotation: Option<String>
```
wrwg, line 156: Oops? Loc is used all over the place as map keys and elsehwere. It should have no extra field. Why is this needed? It doesn't seem to make logical sense to add data to this very basic type.

language/move-prover/spec-lang/src/env.rs
```diff
@@ -1432,9 +1451,11 @@ impl<'env> ModuleEnv<'env> {
     }
 
     /// Allocates a new node id and assigns location and type to it.
-    pub fn new_node(&self, loc: Loc, ty: Type) -> NodeId {
+    pub fn new_node(&self, loc_opt: Option<Loc>, ty: Type) -> NodeId {
```
wrwg, line 1454: This can't be an option. Code assumes that the location already exists.

language/move-prover/spec-lang/src/env.rs
```diff
@@ -1443,10 +1464,27 @@ impl<'env> ModuleEnv<'env> {
...
         self.data
             .instantiation_map
+            .borrow()
             .get(&node_id)
             .cloned()
             .unwrap_or_else(Vec::new)
     }
+
+    /// Add type parameter associated with the given node.
+    pub fn set_node_instantiation(&self, node_id: &NodeId, typ: Type) {
```
wrwg, line 1474: node_id: NodeId. ty: Type

language/move-prover/spec-lang/src/env.rs
```diff
@@ -1443,10 +1464,27 @@ impl<'env> ModuleEnv<'env> {
...
+        let new_types = match self.data.instantiation_map.borrow().get(node_id) {
+            Some(types) => {
```
wrwg, line 1476: This function has a misleading name. It does not set the instantiation, but adds a type to it. Perhaps `extend_node_instantiation`?

language/move-prover/spec-lang/src/env.rs
```diff
@@ -1649,6 +1687,15 @@ impl<'env> StructEnv<'env> {
     pub fn get_spec(&'env self) -> &'env Spec {
         &self.data.spec
     }
+
+    /// Returns the type of the struct
+    pub fn get_type(&self) -> Type {
```
wrwg, line 1692: nit: Type::Struct(..., (0..self.get_type_parameters().len()).iter().map(|i|  Type::TypeParameter(i)).collect_vec())

language/move-prover/spec-lang/src/env.rs
```diff
@@ -1884,6 +1931,27 @@ impl<'env> FunctionEnv<'env> {
         default()
     }
 
+    /// Returns true iff there is a specification check pragma turned on
+    pub fn is_pragma_spec_check(&self) -> bool {
+        let env = self.module_env.env;
+        if let Some(b) = env.is_property_true(&self.module_env.get_spec().properties, ALWAYS_ABORTS_TEST_PRAGMA) {
```
wrwg, line 1937:
```
let props = &self.module_env.get_spec().properties;
env.is_property_true(props, P1).unwrap_or(false) ||
env.is_property_true(props, P2).unwrap_or(false) ||
...
```

language/move-prover/spec-lang/src/env.rs
```diff
@@ -1906,6 +1974,37 @@ impl<'env> FunctionEnv<'env> {
             .any(|Parameter(_, ty)| ty.is_mutable_reference())
     }
 
+    /// Returns the types of mutable reference function arugments
```
wrwg, line 1977: arugments

language/move-prover/spec-lang/src/env.rs
```diff
@@ -1906,6 +1974,37 @@ impl<'env> FunctionEnv<'env> {
...
+            .map(|Parameter(_, ty)| ty.clone())
+            .collect_vec()
+    }
+
+    /// Returns the number of mutable references
+    pub fn get_mut_ref_count(&self) -> usize {
+        self.get_mut_ref_types().len()
+    }
+
+    /// Returns the types of the return values of the procedure.
```
wrwg, line 1991: s/procedure/function

language/move-prover/spec-lang/src/env.rs
```diff
@@ -1906,6 +1974,37 @@ impl<'env> FunctionEnv<'env> {
             .any(|Parameter(_, ty)| ty.is_mutable_reference())
     }
 
+    /// Returns the types of mutable reference function arugments
+    pub fn get_mut_ref_types(&self) -> Vec<Type> {
```
wrwg, line 1978: get_mut_ref_parameter_types

language/move-prover/spec-lang/src/env.rs
```diff
@@ -1906,6 +1974,37 @@ impl<'env> FunctionEnv<'env> {
...
+
+    /// Returns the number of mutable references
+    pub fn get_mut_ref_count(&self) -> usize {
+        self.get_mut_ref_types().len()
+    }
+
+    /// Returns the types of the return values of the procedure.
+    /// This includes the values returned by the function and the
+    /// mutable reference arguments.
+    pub fn get_proc_return_types(&self) -> Vec<Type> {
```
wrwg, line 1994: So this set of functions looks like it is simulating the effect of the memory model transformation. As such it does not belong into FunctionEnv but in FunctionTarget. In fact I don't remember how Shaz has been doing this, but really the FunctionTarget::get_parameters and FunctionTarget::get_return_type already should do the right thing here. @shazqadeer for confirmation how to deal with this.

language/move-prover/src/bytecode_translator.rs
```diff
@@ -541,6 +545,30 @@ impl<'env> ModuleTranslator<'env> {
             emitln!(self.writer);
         }
 
+        // If the function has been annotated with rewritten specifications
+        // (or code), then generate the corresponding specification check functions
+        if func_target.data.rewritten_spec.len() > 0 {
```
wrwg, line 550: Have you run clippy (or Ci)? It will tell you to use !is_empty here.

language/move-prover/src/bytecode_translator.rs
```diff
@@ -767,7 +922,11 @@ impl<'env> ModuleTranslator<'env> {
     /// This generates boogie code for everything after the function signature
     /// The function body is only generated for the `FunctionEntryPoint::Definition`
     /// version of the function.
-    fn generate_inline_function_body(&self, func_target: &FunctionTarget<'_>) {
+    fn generate_inline_function_body(
+        &self,
+        func_target: &FunctionTarget<'_>,
+        rewritten_bytecode: &Option<Vec<Bytecode>>,
```
wrwg, line 928: I don't understand why we need a bytecode parameter here. The idea of the transformation pipeline is that the rewritten code is what you get from function_target. This seems to be not matching the current design.

language/move-prover/src/bytecode_translator.rs
```diff
@@ -600,7 +645,8 @@ impl<'env> ModuleTranslator<'env> {
             | FunctionEntryPoint::VerificationDefinition
             | FunctionEntryPoint::DirectIntraModule
             | FunctionEntryPoint::Indirect
-            | FunctionEntryPoint::DirectInterModule => "{:inline 1} ".to_string(),
+            | FunctionEntryPoint::DirectInterModule
+            | FunctionEntryPoint::Mutated(_) => "{:inline 1} ".to_string(),
```
wrwg, line 649: Why do we need a new entry point kind? This is strange and will lead to confusion, because those entry points have a rather specific semantics. Similar as you extended ConditionKind, this will also leak over the place and increase the complexity of the code model.

language/move-prover/src/bytecode_translator.rs
```diff
@@ -759,6 +805,115 @@ impl<'env> ModuleTranslator<'env> {
...
+
+        // Find the specification check assertion
+        // Only one is allowed at a time because Boogie is not
+        // gauranteed to return all errors in a procedure
+        let spec_check_opt = conds.iter().find(|cond| match cond.kind {
+            ConditionKind::RequiresSpecCheckAssert | ConditionKind::Ensures => true,
+            _ => false,
+        });
+        let spec_check = spec_check_opt
+            .expect("There should be at least one assertion for the specification check.");
```
wrwg, line 836: Is this an error the user can create? Then we must not crash but produce an error. If it is not, how it is prevented that this will fire?

language/move-prover/src/bytecode_translator.rs
```diff
@@ -759,6 +805,115 @@ impl<'env> ModuleTranslator<'env> {
...
             )
         }
+
+        self.writer.unindent();
+        emitln!(self.writer, "}");
+        emitln!(self.writer);
+    }
+
+    /// Spec check body
+    fn generate_function_spec_check_body(
```
wrwg, line 815: This is a bunch of new code with some c&p from existing code. Why do we need this extra handling?

The idea would have been that rewritten functions are entirely dealt with in the transformation pipeline, and no or only marginal changes would be needed in the boogie generator.

language/move-prover/src/bytecode_translator.rs
```diff
@@ -810,7 +969,11 @@ impl<'env> ModuleTranslator<'env> {
         emitln!(self.writer, "\n// bytecode translation starts here");
 
         // Generate bytecode
-        let code = func_target.get_bytecode();
+        let code = if let Some(bytecode) = rewritten_bytecode {
+            &bytecode
```
wrwg, line 973: See above. We should not need to carry rewritten bytecode around, it should be in the FunctionTarget as result of the transformation pipeline.

language/move-prover/src/prelude.bpl
```diff
@@ -1329,3 +1329,10 @@ procedure {:inline 1} $Event_publish_generator(account: $Value) {
 procedure {:inline 1} $Event_emit_event(t: $TypeValue, handler: $Value, msg: $Value) returns (res: $Value) {
     res := handler;
 }
+
+// ==================================================================================
+// Specification checking constants
+
+// An address used for the constant expression specification check
+const $sc_addr: $Value;
+axiom (is#$Address($sc_addr));
```
wrwg, line 1338: nit: braces not needed. Redundant braces are confusing in code, because future users think they are necessary for some reason. They are forbidden my multiple style guides.

language/move-prover/src/spec_translator.rs
```diff
@@ -98,18 +98,23 @@ pub enum FunctionEntryPoint {
     Indirect,
     /// Variant used for verification.
     Verification,
+    /// Mutated inlined variant without pre/post conditions. Used by
+    /// specification checks. Holds `rewritten_code_index` from `Spec`.
+    /// To distinguish between different mutated functions.
+    Mutated(usize),
```
wrwg, line 104: Do you have different mutated functions at one point? I thought to have seen there is only one.

In any case, I think we should not extend this enumeration. We should just rewrite the function and use existing entry point types.

language/move-prover/src/spec_translator.rs
```diff
@@ -972,6 +977,51 @@ impl<'env> SpecTranslator<'env> {
         self.global_env()
             .set_condition_info(loc.clone(), tag, ConditionInfo::for_message(message));
     }
+
+    /// Assumes preconditions for function. This is used for the top-level verification
+    /// entry point of a function.
+    pub fn assume_preconditions(&self) {
```
wrwg, line 983: There is a new type ConditionDistribution which handles this. It might have been introduced after you have written this. This current treatment is not consistent with how we deal with conditions.

language/move-prover/src/spec_translator.rs
```diff
@@ -972,6 +977,51 @@ impl<'env> SpecTranslator<'env> {
...
+                self.writer.set_location(&cond.loc);
+                emit!(self.writer, "assume b#$Boolean(");
+                self.translate_exp(&cond.exp);
+                emit!(self.writer, ");")
+            });
+            emitln!(self.writer);
+        }
+    }
+
+  //pub fn assume_cond(&self, cond: &Condition) {
```
wrwg, line 1006: What's this?

language/move-prover/src/spec_translator.rs
```diff
@@ -1951,6 +2001,9 @@ impl<'env> SpecTranslator<'env> {
         emit!(self.writer, "$Boolean(i#$Integer(");
         self.translate_exp(&args[0]);
         emit!(self.writer, ") {} i#$Integer(", boogie_op);
+        if args.len() == 1 {
+            panic!("fialed at");
```
wrwg, line 2005: Oops. Can this be caused by user input? What does it mean?

davidiw:
Closing out old PR, feel free to re-open.

PR #5540, opened 2020-08-10
[mirai] add bound-checking annotations to libra-crypto crate

stonebuddha:
## Motivation

This commit adds documentation about bound checks to the code and helps reduce the false positives produced by MIRAI so that newly introduced problems would be easier to find.

Yes.

## Test Plan

cargo xtest libra-crypto

### DIFF
--- a/crypto/crypto/src/hash.rs
+++ b/crypto/crypto/src/hash.rs
@@ -248,6 +248,7 @@ impl HashValue {
 
     /// Returns the `index`-th nibble.
     pub fn get_nibble(&self, index: usize) -> Nibble {
+        precondition!(index < HashValue::LENGTH);
         Nibble::from(if index % 2 == 0 {
             self[index / 2] >> 4
         } else {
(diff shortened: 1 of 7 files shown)


crypto/crypto/src/hash.rs
```diff
@@ -248,6 +248,7 @@ impl HashValue {
 
     /// Returns the `index`-th nibble.
     pub fn get_nibble(&self, index: usize) -> Nibble {
+        precondition!(index < HashValue::LENGTH);
```
mimoo, line 251: 👍

crypto/crypto/src/noise.rs
```diff
@@ -646,7 +650,10 @@ impl NoiseSession {
             .map_err(|_| NoiseError::Encrypt)?;
 
         // increment nonce
-        self.write_nonce += 1;
+        self.write_nonce = self
+            .write_nonce
+            .checked_add(1)
```
mimoo, line 655: 👍

PR #5541, opened 2020-08-10
[gha][lbt] add check for previous image tags

rustielin:
## Motivation

LBT compat test assumes that the latest commit in `master` has an image tag associated with it, and uses that to test backwards compatibility. Some commits skip the image build process. This PR adds a check for that, and if it can't find an image tag for the previous commit, just keeps going back in git history until it can find one.

Yes

## Test Plan

Separated this functionality out to separate script, so we can test with:

```
▶ ./.github/actions/land-blocking/find-lbt-images.sh 
libra_validator:land_e47a6961 found
libra_cluster_test:land_e47a6961 found
libra_init:land_e47a6961 found
libra_safety_rules:land_e47a6961 found
All images found for tag land_e47a6961
land_e47a6961
```

## Related PRs

### DIFF
--- a/.github/actions/land-blocking/find-lbt-images.sh
+++ b/.github/actions/land-blocking/find-lbt-images.sh
@@ -1,0 +1,37 @@
+#!/bin/bash
+# Copyright (c) The Libra Core Contributors
+# SPDX-License-Identifier: Apache-2.0
+
+# adapted from --build-all-cti option from libra/docker/build-aws.sh, which is how
+# land blocking test builds images before running cluster-test
+REPOS=(libra_validator libra_cluster_test libra_init libra_safety_rules)
+# the number of commits backwards we want to look
+END=50
+
+image_tag_exists() {
+    if [ -z $1 ]; then
+        echo "Missing image tag"
+        exit 1
+    fi
+    image_tag=$1
+    for (( i=0; i < ${#REPOS[@]}; i++ )); do
+        repo=${REPOS[$i]}
+        aws ecr describe-images --repository-name $repo --image-ids imageTag=$image_tag
+        if [ $? -eq 0 ]; then
+            echo "$repo:$image_tag found"
+        else
+            echo "$repo:$image_tag not found"
+            return 1
+        fi
+    done
+    echo "All images found for tag $image_tag"
+    retval_image_tag=$image_tag
+    return 0
+}
+
+for i in $(seq 0 $END); do
+    test_tag=land_$(git rev-parse --short=8 origin/master~$i)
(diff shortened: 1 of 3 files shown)


rustielin:
@bmwill how can we use the new bors canary to test this?

.github/actions/land-blocking/find-lbt-images.sh
```diff
@@ -0,0 +1,37 @@
+#!/bin/bash
+# Copyright (c) The Libra Core Contributors
+# SPDX-License-Identifier: Apache-2.0
+
+# adapted from --build-all-cti option from libra/docker/build-aws.sh, which is how
+# land blocking test builds images before running cluster-test
+REPOS=(libra_validator libra_cluster_test libra_init libra_safety_rules)
+# the number of commits backwards we want to look
+END=5
```
andll, line 9: We probably want this to be at least ~50.
Reason is that imagine we disabled LBT for some time, and then re-enabled it - it might need to look a bit deep to find proper commit (note that some PRs have multiple commits, so single PR can easily add 5 commits)

.github/actions/land-blocking/find-lbt-images.sh
```diff
@@ -0,0 +1,37 @@
+#!/bin/bash
+# Copyright (c) The Libra Core Contributors
+# SPDX-License-Identifier: Apache-2.0
+
+# adapted from --build-all-cti option from libra/docker/build-aws.sh, which is how
+# land blocking test builds images before running cluster-test
+REPOS=(libra_validator libra_cluster_test libra_init libra_safety_rules)
```
andll, line 7: Any way we could generalize this into one place somehow?
Maybe make `cluster_test_images.sh` and include it both here and in other build-aws?

rustielin: That might get a bit messy, since the list in `build-aws.sh` actually specifies the codebuild project name, which can be different than the actual repo name. I think we can just be mindful about keeping them in sync (I'll add a note in both scripts). I don't expect these images to change too often

andll: Ok :)
Let's make a note then.

andll:
/canary

Enabled compat test killswitch

rustielin:
/canary

stopped sending aws cli output to /dev/null so we can see the output

rustielin:
/canary

GHA does not like multiline strings, so writing to file and tail'ing that instead

andll:
Looks like LBT picked up tag correctly, its just flaky unit test

rustielin:
> Looks like LBT picked up tag correctly, its just flaky unit test

Yea works as intended. I'll land it

PR #5542, opened 2020-08-10
[storage] a few more metrics and dashboard update

msmouse:
## Motivation
1. backup progress gauges.
2. separate timers for commits from save_transactions() and the state pruner.
3. utilize the above and other counters we've been adding and not reflected on the dashboard yet.
    1. set fill=0 in most places - it's much clearer and revealing anomalies way easier.
    2. use irate whenever possible - playing with it I found at 15s interval `irate()` is not really too jumpy, and in cluster-tests we can really use more sensitive graphs than `rate()` can provide.

Y

## Test Plan

cluster test and import new dashboard.

image

image

image

image

image

image

image

image

image

image

image

image

### DIFF
--- a/storage/libradb/src/metrics.rs
+++ b/storage/libradb/src/metrics.rs
@@ -82,3 +82,37 @@ pub static LIBRA_STORAGE_API_LATENCY_SECONDS: Lazy<HistogramVec> = Lazy::new(|| 
     )
     .unwrap()
 });
+
+// Backup progress gauges:
+
+pub(crate) static BACKUP_EPOCH_ENDING_EPOCH: Lazy<IntGauge> = Lazy::new(|| {
+    register_int_gauge!(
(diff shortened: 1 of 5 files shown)


storage/libradb/src/metrics.rs
```diff
@@ -82,3 +82,49 @@ pub static LIBRA_STORAGE_API_LATENCY_SECONDS: Lazy<HistogramVec> = Lazy::new(||
     )
     .unwrap()
 });
+
+pub static LIBRA_STORAGE_OTHER_TIMERS_SECONDS: Lazy<HistogramVec> = Lazy::new(|| {
```
lightmark, line 86: other timers?

msmouse: Updated the description, hopefully it's easy to understand now.

PR #5544, opened 2020-08-10
[json-rpc] Add chain id to BlockMetadata

metajack:
Chain id is in the metadata of the response, but this is inaccessible in the current Rust client API. This adds the chain id to BlockMetadata.

### DIFF
--- a/json-rpc/docs/type_metadata.md
+++ b/json-rpc/docs/type_metadata.md
@@ -5,7 +5,7 @@
 |-----------|----------------|-----------------------------------------------|
 | version   | unsigned int64 | The latest block (ledger) version             |
 | timestamp | unsigned int64 | The latest block (ledger) timestamp           |
-
+| chain_id  | unsigned int8  | Chain ID of the Libra network                 |
 
 ### Example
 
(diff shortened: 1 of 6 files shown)


json-rpc/docs/type_metadata.md
```diff
@@ -5,7 +5,7 @@
 |-----------|----------------|-----------------------------------------------|
 | version   | unsigned int64 | The latest block (ledger) version             |
 | timestamp | unsigned int64 | The latest block (ledger) timestamp           |
-
+| chain_id  | unsigned int8  | Chain ID of the Libra network                 |
```
thefallentree, line 8: should include it in the example below.

metajack: done, missed this

xli:
could you also add one simple change to API-CHANGELOG.md

metajack:
Unit test fixed and API-CHANGELOG.md updated.

/land

thefallentree:
looking back, we should probably fix the rust client instead of adding this here. oh-well

PR #5545, opened 2020-08-10
[state sync] reject responses for unknown requests

christinacelee:
## Motivation

Reject state sync responses for requests not sent by a node. This better protects against noisy, malicious upstream peers. This also rejects timed-out responses, but with a reasonable timeout duration this should be okay 

## Test Plan

Unit test

### DIFF
--- a/state-synchronizer/src/coordinator.rs
+++ b/state-synchronizer/src/coordinator.rs
@@ -764,6 +764,17 @@ impl<T: ExecutorProxyTrait> SyncCoordinator<T> {
                     format_err!("[state sync] Empty chunk from {:?}", peer)
                 })?;
 
+        // check whether this response corresponds to a sent request
+        // only accept/process responses for requests we sent
+        // this might be a response for a timed-out request, and penalization of such responses
+        // happens in the `RequestManager::check_timeout` calls in the periodic `check_progress` calls
+        if !self
+            .request_manager
+            .is_sent_request(chunk_start_version, peer)
+        {
+            bail!("[state sync] received chunk response for unknown request");
+        }
+
         if chunk_start_version != known_version + 1 {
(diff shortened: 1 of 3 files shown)


state-synchronizer/src/coordinator.rs
```diff
@@ -764,6 +764,17 @@ impl<T: ExecutorProxyTrait> SyncCoordinator<T> {
                     format_err!("[state sync] Empty chunk from {:?}", peer)
                 })?;
 
+        // check whether this response corresponds to a sent request
+        // only accept/process responses for requests we sent
+        // this might be a response for a timed-out request, and penalization of such responses
+        // happens in the `RequestManager::check_timeout` calls in the periodic `check_progress` calls
+        if !self
+            .request_manager
+            .is_sent_request(chunk_start_version, peer)
```
zekun000, line 773: it seems redundant to the chunk_version_mismatch check? if we didn't send the request the version can't match? if the request is valid why not take the response even it times out?

at least can we consolidate this with process_chunk_version_mismatch? or even unifies all the verification of the GetChunkResponse

christinacelee: yeah I agree probabilistically there is a big overlap b/w chunk version mismatch and an unrequested chunk - it's very unlikely an unrequested chunk's version matches the local version. I guess I was trying to address the very unlikely case where the unrequested chunk's version does match the local version. 

Another issue is differentiating b/w timed-out chunk response and completely unrequested chunk response, we have to do more bookkeeping to keep track of past responses, and at that point the complexity is not worth.

The main concern I wanted to address was protecting from bad spammy messages. I'm also starting to think the chunk mismatch is good enough protection, so maybe we'll leave it as is?

PR #5548, opened 2020-08-11
[state sync] fix flaky multicast test

christinacelee:
## Motivation

Fix flaky `test_multicast_failover` test
It was flaky because the `wait_for_version` calls take too long and were triggering multicast mode when the test didn't expect it
Also extended multicast timeout duration in `test_fn_failover` because multicast was also being unintentionally triggered due to long test duration, especially in running the full test suite (parallelized test runs add to time)

## Test Plan

Local flake seems to be gone

### DIFF
--- a/state-synchronizer/src/request_manager.rs
+++ b/state-synchronizer/src/request_manager.rs
@@ -356,6 +356,12 @@ impl RequestManager {
             .map(|req_info| req_info.first_request_time)
     }
 
+    pub fn get_multicast_start_time(&self, version: u64) -> Option<SystemTime> {
+        self.requests
+            .get(&version)
+            .map(|req_info| req_info.multicast_start_time)
+    }
+
     /// Removes requests for all versions before `version` (inclusive) if they are older than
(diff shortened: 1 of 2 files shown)


state-synchronizer/src/request_manager.rs
```diff
@@ -403,7 +409,8 @@ impl RequestManager {
         }
 
         // increment multicast level if this request is also multicast-timed-out
-        if Self::is_timeout(last_request_time, self.multicast_timeout) {
+        let multicast_start_time = self.get_multicast_start_time(version).unwrap_or(UNIX_EPOCH);
```
zekun000, line 412: multicast_start_time is always <= last_request_time, so this would trigger the timeout more often?

christinacelee: multicast timeout and regular timeout are tracked differently, regular timeout is computed earlier in the code, and they each have different timeout durations. This is just checking for multicast timeout and updating state accordingly

PR #5550, opened 2020-08-11
[Fuzzing] Added fuzz tests to the secure JSON RPC client.

JoshLind:
## Motivation

This PR adds fuzz tests to the secure JSON RPC client (one of the core components relied on by the key manager). More specifically, this PR tests the client's handling of JSON RPC responses from a JSON RPC endpoint (i.e., responses sent over the network). This makes the client an ideal candidate for an adversary to target (e.g., by manipulating data/responses sent to the client). 

To achieve this, the PR offers a small single commit that:
1) Moves the handling of JSON RPC responses into functions (so they can be called directly by the fuzzer).
2) Extracts a "test_helpers" module to allow unit tests and fuzz tests to share useful testing functions (e.g., for generating various data structures)
3) Implements the fuzz targets for each of the 3 API calls the JSON RPC client currently supports. For this, we add support for generating a template response (corpus) used by the fuzzer.

Yes.

## Test Plan

1) I first ran all unit tests locally to ensure the fuzzing paths were being executed.
2) I next ran the command 'cargo run --bin libra-fuzzer list' to ensure that the new fuzzing targets were shown.
3) I followed by generating a corpus locally, and running the fuzzing targets locally (e.g., using 'RUSTC_BOOTSTRAP=1 cargo run --bin libra-fuzzer fuzz <TARGET> -- --release --debug-assertions'). I let it run for about 20 minutes before killing it. No errors found thus far for each of the fuzzers.

## Related PRs

None.

### DIFF
--- a/secure/json-rpc/Cargo.toml
+++ b/secure/json-rpc/Cargo.toml
@@ -35,4 +35,4 @@ storage-interface = { path = "../../storage/storage-interface", version = "0.1.0
 vm-validator = { path = "../../vm-validator", version = "0.1.0" }
 
 [features]
-fuzzing = ["libra-json-rpc/fuzzing", "libra-types/fuzzing"]
+fuzzing = []
--- a/secure/json-rpc/src/lib.rs
+++ b/secure/json-rpc/src/lib.rs
@@ -83,21 +83,7 @@ impl JsonRpcClient {
         )?))];
         let response = self.execute_request(method, params);
 
-        match response.status() {
-            200 => {
-                let response = &response.into_string()?;
-                if let Ok(failure_response) =
-                    serde_json::from_str::<JSONRpcFailureResponse>(response)
-                {
-                    Err(Error::InternalRPCError(format!("{:?}", failure_response)))
-                } else {
-                    let _submit_response =
-                        serde_json::from_str::<SubmitTransactionResponse>(response)?;
-                    Ok(())
-                }
-            }
-            _ => Err(Error::RPCFailure(response.into_string()?)),
-        }
+        process_submit_transaction_response(response)
     }
 
     /// Returns the associated AccountState for a specific account at a given version height.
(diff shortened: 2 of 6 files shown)


JoshLind:
Note: this PR isn't ready to be reviewed yet. It requires some changes to the fuzz testing infrastructure. Will update when it's ready 😄

secure/json-rpc/Cargo.toml
```diff
@@ -35,4 +35,4 @@ storage-interface = { path = "../../storage/storage-interface", version = "0.1.0
 vm-validator = { path = "../../vm-validator", version = "0.1.0" }
 
 [features]
-fuzzing = ["libra-json-rpc/fuzzing", "libra-types/fuzzing"]
+fuzzing = []
```
JoshLind, line 38: These deps were unused 😄

secure/json-rpc/src/lib.rs
```diff
@@ -737,3 +683,182 @@ mod test {
...
+            result: Some(TransactionView {
+                vm_status: VMStatusView::Executed,
+            }),
+        };
+        serde_json::to_vec::<TransactionViewResponse>(&response).unwrap()
+    }
+
+    // This function fuzzes a submit_transaction() response.
+    pub fn fuzz_submit_transaction_response(data: &[u8]) {
+        if let Ok(submit_tx_response) = String::from_utf8(data.to_vec()) {
```
JoshLind, line 765: @mimoo, the only downside to this approach is that the 'data' input from the fuzzer must be converted to a string to be inserted into a Response, otherwise the fuzzer fails early. I'm considering making more finer grained tests, but hopefully this is an okay starting point..

mimoo: yeah, you're going to have a LOT of inputs that are just going to get dropped here. This would be fine if the real code really used `String::from_utf8()`, as it could be something we would want to fuzz, but it looks like this isn't the case: https://docs.rs/ureq/1.3.0/src/ureq/response.rs.html#368-384 so you might want to really fuzz this logic instead.

JoshLind:
Thanks @mimoo! As discussed, I'll upgrade these fuzzers to use the proptest format in the new PR!

PR #5552, opened 2020-08-11
[backup] fix backup throughput report in cluster test

msmouse:
## Motivation

Just realized the reported bandwidth is an average across all backup related endpoints. Should be sum of them instead.
Also the final number is an average of 10 sample points in a short time window (5 mins), using `rate` will make the samples overlap much. Using `irate` instead here.

Also, report -1 on error, to make it more obvious.

Y

## Test Plan

cluster test


andll:
Do we also want to throw in some function that will convert multidigt Bytes/sec to something like '1.3 Mbytes/sec'?

PR #5557, opened 2020-08-11
[sccache][CI] use checksum of toolchains in cache key

sausagee:
## Motivation
Since last week we were hit by supposedly rare sccache checksum issue in libfuzzer-sys. We had to disable sccache to reduce churns. 

Upon closer look, I found that CircleCI offers cache corruption check but up to 500MB (https://circleci.com/docs/2.0/caching/#cache-size), while we are saving/restoring somewhere close to 2GB.  
image

At the same time, I realized the cached sccache _binary_ was complied with different rust and cargo toolchains than those of sccache _compilation cache content_. The two didn't use the same key for indexing either.

    Ouch! x2

This PR attempts to address both issues to re-enable sccache.  It has 3 parts

Part 1

    Both the binary and the compilation cache use checksum of the same
    set of toolchain version files in the index key. The compilation
    adds the date code in addition to the key to keep the cache fresh for
    the day. After development stablize, the date code can be removed.

Part 2

    CircleCI supports cache corruption check upto 500MB. We will downsize
    the sccache size to 2GB and split into 4 chunks. The downsizeing is fine
    as our current sccache usage does not exceed 2GB.

Part 3

    Re-enable sccache in workflow.

## Test Plan
CI

Canary example https://app.circleci.com/pipelines/github/libra/libra/27005/workflows/9cf93f08-f46c-42d2-8a37-e8267d3b9916/jobs/176288
image

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -102,7 +102,7 @@ commands:
             else
               echo 'sccache binary is found. Skipping install.'
             fi
-            echo 'export SCCACHE_CACHE_SIZE=4G' >> $BASH_ENV
+            echo 'export SCCACHE_CACHE_SIZE=2G' >> $BASH_ENV
             echo 'export RUSTC_WRAPPER=sccache' >> $BASH_ENV
             echo 'export CC="sccache cc"' >> $BASH_ENV
             echo 'export CXX="sccache c++"' >> $BASH_ENV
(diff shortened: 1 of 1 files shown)


metajack:
Note that sccache does not need to be compiled with the same toolchain. sccache binary needs only to be changed when we upgrade the sccache version.

rexhoffman:
Ahh the split is the internal structure of sccache.
  
So if a checksum fails, just the one cache gets blown away, or only the part that failed.
We could possible have our own checksumming that would be portable, no?

Get the sha/md5 of the full dir ourselves and store it in a smaller cache, and check against our own checksum? -- If the check fails, blow away the cache.   Or a checksum of each file, and kill only failed files.

Just looking for possible improvements.

sausagee:
> This looks good.
> 
> I am slightly concerned that the clever way you split up the sccache buckets is not guaranteed to be <500MB, but regardless I think this should be an improvement as at least the ones that are smaller will be corruption checked.
> 
> /land

That's right, there is no guarantee.  I glanced over the LruDiskCache in sccache and didn't see enforcement to evenly distribute the size among the chunks.  I will monitor it this week after landing.

sausagee:
> Ahh the split is the internal structure of sccache.
> 
> So if a checksum fails, just the one cache gets blown away, or only the part that failed.
> We could possible have our own checksumming that would be portable, no?
> 
> Get the sha/md5 of the full dir ourselves and store it in a smaller cache, and check against our own checksum? -- If the check fails, blow away the cache. Or a checksum of each file, and kill only failed files.
> 
> Just looking for possible improvements.

I think there is lots of opportunity inside the sccache internals, such as mixing S3 and local cache on disk. We can bring this up at the roadmap planning tomorrow.

PR #5562, opened 2020-08-11
[gha][lbt] increase LBT timeout to 50min

rustielin:
After recent changes to cluster-test, LSR and Vault are enabled by
default. This has caused LBT to consistently either time out, or run
very close to the existing 40 min timeout. Increasing this to 50 for
now. Land this only when bors has been reconfigured (EOD).

e.g. https://github.com/libra/libra/runs/969410138 and https://github.com/libra/libra/runs/972408423 both experience long running time.

### DIFF
--- a/.github/workflows/land-blocking.yml
+++ b/.github/workflows/land-blocking.yml
@@ -11,9 +11,9 @@ jobs:
   build-and-run-cluster-test:
     name: Build images and run cluster test
     runs-on: self-hosted
-    # NOTE the total time should cover build (~10 min) and test (~10 min).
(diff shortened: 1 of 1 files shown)


rustielin:
> Have we thought about splitting the testcases in Cluster Test and run them in parallel? Assuming we can allocate more pods.

yea good idea, we can probably do this the next time we make a major change to cluster-test. If we split by perf benchmark and compat test, we can probably save ~10 min per run.

rustielin:
/land

Bors has been redeployed with new timeout

PR #5567, opened 2020-08-11
[Fuzzing] Derive fuzz target names from implementation and clean up boiler plate.

JoshLind:
## Motivation

At present, fuzz target names are derived from the module in which they are declared. This means that each fuzz target must be declared in its own file. As a result, a lot of files are produced, many with names that are context specific and not easy to understand. Instead, it would be ideal if multiple fuzz targets for a single component (e.g., network, executor, etc.) could be grouped together and defined in the same file. This would allow us to better understand what targets exist, and which components are lacking.

This PR achieves this by deriving the names of fuzz targets from the struct implementation, rather than the module (file) name. To do this, the PR offers several commits:

1. We derive the names of fuzz targets from the struct implementation.
2. We group all network noise related fuzz targets into a single file (network_noise.rs).
3. We group all storage related fuzz targets into a single file (storage.rs).
4. We group transaction related fuzz targets into a single file (transaction.rs), as well as rename existing files to be according to the component/high-level entity that is being tested.
5. We alphabetize the fuzzing target list and fix a typo in the README.

Another benefit of this PR is that we produce less boilerplate: the name of the fuzz target can be derived by default, which means fuzz test authors don't need to manually call: "module_name!" in the implementations. See commit 1.

Yes.

## Test Plan

Running `cargo run --bin libra-fuzzer list` produces the new list of target names:                                                              
  * CompiledModuleTarget        VM CompiledModule (custom deserializer)
  * ConsensusProposal           Consensus proposal messages
  * ExecuteAndCommitChunk       state-sync > executor::execute_and_commit_chunk
  * JsonRpcSubmitTransactionRequest    JSON RPC submit transaction request
  * LanguageTransactionExecution    Language execute randomly generated transactions
  * MempoolIncomingTransactions    Transactions submitted to mempool
  * NetworkNoiseInitiator       Network Noise crate initiator side
  * NetworkNoiseResponder       Network Noise crate responder side
  * NetworkNoiseStream          Network Noise crate stream
  * RpcInboundRequest           P2P Network Inbound RPC Request
  * SignedTransactionTarget     SignedTransaction (LCS deserializer)
  * StateSyncMsg                State sync network message
  * StorageSchemaDecode         Storage schemas do not panic on corrupted bytes.
  * ValueTarget                 VM values + types (custom deserializer)

Moreover, generating the corpus and running a few fuzzers locally seems to work fine.

## Related PRs

This PR (https://github.com/libra/libra/pull/5550) is currently blocked on this change as it adds several new fuzzers for the secure json rpc client in the same module.


testsuite/libra-fuzzer/src/lib.rs
```diff
@@ -13,7 +13,13 @@ pub mod fuzz_targets;
...
 pub trait FuzzTargetImpl: Sync + Send + fmt::Debug {
     /// The name of the fuzz target.
-    fn name(&self) -> &'static str;
+    /// By default, we use the struct name, however, implementations may prefer to override this.
+    fn name(&self) -> &'static str {
+        std::any::type_name::<Self>()
+            .rsplit("::")
+            .next()
+            .expect("Implementation struct name must have at least one component")
+    }
```
sunshowers, line 22: This makes sense.

sunshowers:
Ah you're going to have to update the locations for the checked in artifacts :)

testsuite/libra-fuzzer/src/lib.rs
```diff
@@ -13,7 +13,13 @@ pub mod fuzz_targets;
 /// Implementation for a particular target of a fuzz operation.
 pub trait FuzzTargetImpl: Sync + Send + fmt::Debug {
     /// The name of the fuzz target.
-    fn name(&self) -> &'static str;
+    /// By default, we use the struct name, however, implementations may prefer to override this.
+    fn name(&self) -> &'static str {
+        std::any::type_name::<Self>()
+            .rsplit("::")
+            .next()
```
mimoo, line 20: don't you want last() ?

JoshLind: rsplit does a reverse split 😃 https://doc.rust-lang.org/std/primitive.str.html#method.rsplit

JoshLind:
Awesome, thanks @sunshowers! 😄 Hopefully I got them all... /land

PR #5571, opened 2020-08-12
[vm] More logging

tnowacki:
- Logged the function for `expect_only_successful_execution`
- Logged for unverified modules loaded on chain

## Motivation

- logging is good

## Test Plan

- cargo test
- not sure how we normally test logging (if at all)

### DIFF
--- a/language/libra-vm/src/errors.rs
+++ b/language/libra-vm/src/errors.rs
@@ -110,7 +110,7 @@ pub fn convert_write_set_prologue_error(status: VMStatus) -> VMStatus {
         | VMStatus::MoveAbort(location, code @ EBAD_ACCOUNT_AUTHENTICATION_KEY) => {
             if location != known_locations::write_set_manager_module_abort() {
                 crit!(
-                    "[libra_vm] Unexpected write set prologue move abort: {:?}::{:?}",
+                    "[libra_vm] Unexpected write set prologue Move abort: {:?}::{:?}",
                     location,
                     code
                 );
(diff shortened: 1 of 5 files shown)


dariorussi:
@metajack @sunshowers we are adding `libra-logger` to the` move-vm`. That adds a dependency.
I remember Rain kept `libra-logger` out of the `move-vm` in the past, but I don't remember the reason. I think it was for clean dependencies, and this should be ok, but it would be good if you could confirm

language/libra-vm/src/errors.rs
```diff
@@ -132,16 +132,21 @@ pub fn convert_write_set_prologue_error(status: VMStatus) -> VMStatus {
...
 
-        status => {
-            crit!(
-                "[libra_vm] Unexpected error from known move function. Error: {:?}",
-                status
-            );
-            VMStatus::Error(StatusCode::UNEXPECTED_ERROR_FROM_KNOWN_MOVE_FUNCTION)
+            status => {
+                crit!(
+                    "[libra_vm] Unexpected error from known move function, '{}'. Error: {:?}",
```
bob-wilson, line 144: Nit (not new to this change): "move" should start with an uppercase letter, i.e., "...known Move function"

language/vm/src/errors.rs
```diff
@@ -107,7 +108,11 @@ impl VMError {
     pub fn expect_no_verification_errors(self) -> Self {
         match self.status_type() {
             status_type @ StatusType::Deserialization | status_type @ StatusType::Verification => {
-                let message = format!("Unexpected verifier/deserialization error! This likely means there is code stored on chain is unverifiable!\nPrevious error {:?}", &self);
+                let message = format!(
+                    "Unexpected verifier/deserialization error! This likely means there is code \
```
bob-wilson, line 112: Grammar (not new to this change): that last sentence has two verbs. I think it needs a "that": e.g., "...code stored on chain *that* is unverifiable!"

sunshowers:
Yeah it wasn't obvious whether libra-logger should have been a dependency of the move runtime or not. I think eventually when we split up the Move VM entirely, we'll want to probably have logging/metrics interfaces.

vgao1996:
@dariorussi I think it's fine we depend on libra-logger for now. One day if we do achieve a total Libra/Move separation, I'd assume us defining a logger trait for the VM which will get implemented for libra-logger or some adapters, but that a story for the far future...

tnowacki:
- transaction-builder depends on the stdlib crate, which depends on the VM crate
- Having the logging in `vm/src/errors.rs` caused `transaction-builder` to not build on PowerPC32, which it needs to for various reasons 
- I moved the logging into vm/runtime which fixed it

PR #5573, opened 2020-08-12
[libra framework] Add CLI tool to explain Move abort errors

tzakian:
This PR adds the creation of an error mapping of `module_id, abort_code` to the constant name, and any doc comments on that constant in the standard library. It does this by hooking in to the prover in a similar manner to the ABI and document generators.

The error mapping and error descriptions are generated from the source constants and doc comments whenever the stdlib is recompiled.

A couple things:
* The `Errors` module is specified as a special module that declares the error categories (is user configurable)
* Error constants must all start with an uppercase `E`. 
* If we find duplicate constant values for errors in a module (or error categories) then we will exit and report an error during generation of the error map.

## Example:

If execute a transaction and you get a `move_abort` back:
```
"vm_status": {
  "move_abort": {
     "abort_code": 263,
       "location": "00000000000000000000000000000001::LibraTransactionPublishingOption"
  }
}
```

you can use this tool to explain what that error means:

```bash
$ cargo run -p move-explain -- -l 00000000000000000000000000000001::LibraTransactionPublishingOption -a 263
Category:
  Name: INVALID_ARGUMENT
  Description:  An argument provided to an operation is invalid. Example: a signing key has the wrong format.
Reason:
  Name: EALLOWLIST_ALREADY_CONTAINS_SCRIPT
  Description:  The script hash already exists in the allowlist
```


sblackshear:
> Right now the creation of the mapping is a bit hacky using regex's to pull out the constants. In the future we'll probably want to make this generated in a similar manner to the ABIs, however we don't have access to constants at that point right now, and it looked like it could be a fair bit of work to plug this in to the prover at the moment due to this.

Agreed that this is the right way to populate these (but I think probably as part of docgen, not as part of ABI generation). It would be really nice for docgen to emit a single markdown file with all of the error codes and documentation in addition to this json-rpc explainer.

tzakian:
I actually wound up adding constants to the document generator in #5591, so once that's in I will update this PR to use that to generate the error mapping. Good point on the single file for errors too, it shouldn't be hard to generate that along with it :)

language/move-prover/errmapgen/src/errmapgen.rs
```diff
@@ -0,0 +1,239 @@
...
+    pub fn add_module_error(
+        &mut self,
+        module_id: ModuleId,
+        abort_code: u64,
+        description: ErrorDescription,
+    ) -> Result<()> {
+        let module_error_map = self
+            .module_error_maps
+            .entry(module_id.clone())
+            .or_insert_with(BTreeMap::new);
```
wrwg, line 105: or_default()?

language/move-prover/src/cli.rs
```diff
@@ -344,6 +351,12 @@ impl Options {
                     .help("run the ABI generator instead of the prover. \
                     Generated ABIs will be written into the directory `./abi` unless configured otherwise via toml"),
             )
+            .arg(
+                Arg::with_name("errmapgen")
+                    .long("errmapgen")
```
wrwg, line 356: One more like this and I guess we need to make a real cli out of this, like `mc errmapgen <options`...

riyazdf:
this is great, thanks!

PR #5578, opened 2020-08-12
[Move codegen] support TransactionArgument::Vector

matbd:
## Motivation

* Add definitions and (tentatively) add parsing support for TransactionArgument::{Vector, U8Vector, U64Vector, U128Vector, BoolVector, AddressVector} in Move
* Support the new TransactionArgument::*Vector in the codegen of each language

The converter into value is possibly too simplistic at the moment. We could easily attempt to do some (syntax-oriented) type-inference/type-checking but I suspect that this adds nothing.

If people like the PR, happy to add more tests on the Move side.

## Test Plan

test included

### DIFF
--- a/language/libra-vm/src/libra_transaction_executor.rs
+++ b/language/libra-vm/src/libra_transaction_executor.rs
@@ -20,8 +20,8 @@ use libra_types::{
     account_config,
     block_metadata::BlockMetadata,
     transaction::{
-        ChangeSet, Module, Script, SignatureCheckedTransaction, Transaction, TransactionArgument,
-        TransactionOutput, TransactionPayload, TransactionStatus, WriteSetPayload,
+        ChangeSet, Module, Script, SignatureCheckedTransaction, Transaction, TransactionOutput,
+        TransactionPayload, TransactionStatus, WriteSetPayload,
     },
     vm_status::{KeptVMStatus, StatusCode, VMStatus},
     write_set::{WriteSet, WriteSetMut},
@@ -162,6 +162,11 @@ impl LibraVM {
 
         // Run the execution logic
         {
+            let args = script
+                .args()
+                .iter()
+                .map(Value::try_from)
+                .collect::<Result<Vec<_>, _>>()?;
             cost_strategy.enable_metering();
             cost_strategy
                 .charge_intrinsic_gas(txn_data.transaction_size())
@@ -170,7 +175,7 @@ impl LibraVM {
                 .execute_script(
                     script.code().to_vec(),
                     script.ty_args().to_vec(),
-                    convert_txn_args(script.args()),
+                    args,
                     vec![txn_data.sender()],
                     cost_strategy,
                 )
@@ -316,7 +321,12 @@ impl LibraVM {
             WriteSetPayload::Direct(change_set) => change_set.clone(),
             WriteSetPayload::Script { script, execute_as } => {
                 let mut tmp_session = self.0.new_session(remote_cache);
-                let args = convert_txn_args(script.args());
+                let args = script
+                    .args()
+                    .iter()
+                    .map(Value::try_from)
+                    .collect::<Result<Vec<_>, _>>()
+                    .map_err(Err)?;
                 let senders = match txn_sender {
                     None => vec![*execute_as],
                     Some(sender) => vec![*execute_as, sender],
@@ -710,20 +720,6 @@ pub(crate) fn discard_error_output(err: StatusCode) -> TransactionOutput {
         0,
         TransactionStatus::Discard(err),
     )
-}
-
-/// Convert the transaction arguments into Move values.
-fn convert_txn_args(args: &[TransactionArgument]) -> Vec<Value> {
-    args.iter()
-        .map(|arg| match arg {
-            TransactionArgument::U8(i) => Value::u8(*i),
-            TransactionArgument::U64(i) => Value::u64(*i),
-            TransactionArgument::U128(i) => Value::u128(*i),
-            TransactionArgument::Address(a) => Value::address(*a),
-            TransactionArgument::Bool(b) => Value::bool(*b),
-            TransactionArgument::U8Vector(v) => Value::vector_u8(v.clone()),
-        })
-        .collect()
 }
 
 impl AsRef<LibraVMImpl> for LibraVM {
--- a/language/move-core/types/src/transaction_argument.rs
+++ b/language/move-core/types/src/transaction_argument.rs
@@ -13,6 +13,15 @@ pub enum TransactionArgument {
     Address(AccountAddress),
     U8Vector(#[serde(with = "serde_bytes")] Vec<u8>),
     Bool(bool),
+
+    U64Vector(Vec<u64>),
+    U128Vector(Vec<u128>),
+    BoolVector(Vec<bool>),
+    AddressVector(Vec<AccountAddress>),
+
+    /// Generic vector for the remaining cases.
+    /// Note that the specialized vectors above (U8Vector, U64Vector, etc) must be used when applicable.
+    Vector(Vec<TransactionArgument>),
 }
 
 impl fmt::Debug for TransactionArgument {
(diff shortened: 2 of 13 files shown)


language/libra-vm/src/libra_transaction_executor.rs
```diff
@@ -724,6 +724,7 @@ fn convert_txn_args(args: &[TransactionArgument]) -> Vec<Value> {
             TransactionArgument::Address(a) => Value::address(*a),
             TransactionArgument::Bool(b) => Value::bool(*b),
             TransactionArgument::U8Vector(v) => Value::vector_u8(v.clone()),
```
vgao1996, line 726: Just to double check, what do we plan to do with `TransactionArgument::U8Vector` once `TransactionArgument::Vector` gets fully supported? Shall we remove it eventually?

matbd: We should keep it because it is much more efficient that a vector a tagged bytes

matbd:
@dariorussi I was not going to land this right now but if the process could be sped up a little bit it would save me some nasty rebase in my own stack of PRs.

vgao1996:
@dariorussi Could you elaborate why `TransactionArgument` must be self-descriptive? By self-descriptive I mean you can deduce the type directly from it. Also isn't the type already being passed around with this? (Vec<TransactionArgument> + Script, which contains the signature). The VM will check this against the signatures specified by the script anyways.

dariorussi:
> @dariorussi I was not going to land this right now (as this creates a way to crash the VM from outside) but if the process could be sped up a little bit it would save me some nasty rebase in my own stack of PRs.

I know, sorry :)

dariorussi:
> @dariorussi Could you elaborate why `TransactionArgument` must be self-descriptive? By self-descriptive I mean you can deduce the type directly from it. Also isn't the type already being passed around with this? (Vec + Script, which contains the signature). The VM will check this against the signatures specified by the script anyways.

there are few problems here:
1- the adapter is doing the `TransactionArgument` -> `Value` conversion. The adapter does not have the `CompiledScript` it has a `Vec<u8>`.
2- by self-descriptive I mean that at the very least you have to be able to convert to some type. The VM will do the check later as you said and reject any value that does not match the signature (the VM obviously has the signature). With this representation I don't know how to describe an empty vector. And there are other representations where at least you an go in with more expectations on what to find

so that is what I am thinking about and talk a bit at the meeting today

language/move-core/types/src/transaction_argument.rs
```diff
@@ -13,6 +13,7 @@ pub enum TransactionArgument {
     Address(AccountAddress),
     U8Vector(#[serde(with = "serde_bytes")] Vec<u8>),
     Bool(bool),
+    Vector(Vec<TransactionArgument>),
```
phlip9, line 16: it seems like this would allow heterogenous vectors, like `[U8(4), U64(123), Address(..), ..]`, which seems wrong i guess.

also, do we want to disallow the generic `Vector(U8)`? it definitely seems a bit weird to have multiple encodings of the same thing.

do we need any hardening against deeply nested vectors?

dariorussi: yeah this would allow a `Struct` effectively but we think that may be inevitable and error'ed on conversation

dariorussi: > do we need any hardening against deeply nested vectors?
we should be protected at few level, in las and in the VM, but certainly something to double check when we have the final solution in place

matbd: > It seems like this would allow heterogenous vectors, like `[U8(4), U64(123), Address(..), ..]`, which seems wrong i guess.

This is equivalent to sending garbage LCS bytes. (Calling a Move script is a RPC protocol using LCS/TransactionArgument as transport.)

> also, do we want to disallow the generic `Vector(U8)`? it definitely seems a bit weird to have multiple encodings of the same thing.

LCS itself has two types actually (as well as the move vm). This is a bit annoying but it is worth it : many languages have a special type for byte-strings which is not the same as Vec<u8>. The latter is typically less efficient (and this is visible in this PR as well).

Anyways, it is up to Move converters to disallow it.

> do we need any hardening against deeply nested vectors?

LCS already has a limit in place (although I haven't finished updated the client-side runtimes).

matbd:
> > @dariorussi Could you elaborate why `TransactionArgument` must be self-descriptive? By self-descriptive I mean you can deduce the type directly from it. Also isn't the type already being passed around with this? (Vec + Script, which contains the signature). The VM will check this against the signatures specified by the script anyways.
> 
> there are few problems here:
> 1- the adapter is doing the `TransactionArgument` -> `Value` conversion. The adapter does not have the `CompiledScript` it has a `Vec<u8>`.
> 2- by self-descriptive I mean that at the very least you have to be able to convert to some type. The VM will do the check later as you said and reject any value that does not match the signature (the VM obviously has the signature). With this representation I don't know how to describe an empty vector. And there are other representations where at least you can go in with more expectations on what to find
> 
> so that is what I am thinking about and talk a bit at the meeting today

Although the idea of type-carrying vectors is not a major problem (except network waste in high dimensional vectors) in a world where everything is code-generated, I hope we can avoid this user-visible complication if it is not strictly needed.

matbd:
I just looked at the Move code and it looks that the change is doable without adding an extra argument to the 2 functions `convert_txn_args`.

Specifically, there is a function `Value::constant_vector_generic` that looks like what we need. This function takes an extra parameter `inner: &SignatureToken` but uses it only for a quick consistency check (move vm containers do not seem to carry types) so I'm assuming that we just want to avoid creating values that break local invariants of a "constant vector container".

Therefore, we can just look at the transaction arguments in a `TransactionArgument::Vector` and recursively figure out if they look like a homogenous vector (of any type, possible unknown for an empty vector).

matbd:
> I just looked at the Move code and it looks that the change is doable without adding an extra argument to the 2 functions `convert_txn_args`.
> 
> Specifically, there is a function `Value::constant_vector_generic` that looks like what we need. This function takes an extra parameter `inner: &SignatureToken` but uses it only for a quick consistency check (move vm containers do not seem to carry types) so I'm assuming that we just want to avoid creating values that break local invariants of a "constant vector container".
> 
> Therefore, we can just look at the transaction arguments in a `TransactionArgument::Vector` and recursively figure out if they look like a homogenous vector (of any type, possible unknown for an empty vector).

I just talked to @dariorussi who pointed out we must be careful to build empty vectors with the right constructor (out of the many ones available internally: u8vector, address_vector, ...) With this in mind, I'm going to try to fill the gaps in this PR and see how it goes.

language/libra-vm/src/libra_transaction_executor.rs
```diff
@@ -724,6 +724,7 @@ fn convert_txn_args(args: &[TransactionArgument]) -> Vec<Value> {
...
+            TransactionArgument::Vector(_) => panic!("not implemented"),
```
mimoo, line 727: `unimplemented!()` is for that : o also isn't this adding a potential crash oO? (no idea, haven't looked further)

thefallentree: yes, pls don't add panic() in the VM layer :-D

matbd: this is a POC (see other comments)

language/move-core/types/src/transaction_argument.rs
matbd: I carried on with the idea and explicitly do not attempt to type-check arguments during conversion to (tentative) move values.

language/move-vm/types/src/values/values_impl.rs
```diff
@@ -9,7 +9,9 @@ use move_core_types::{
         words_in, AbstractMemorySize, GasAlgebra, GasCarrier, GasUnits, CONST_SIZE, REFERENCE_SIZE,
         STRUCT_SIZE,
     },
+    transaction_argument::TransactionArgument,
```
dariorussi, line 12: it was good for `TransactionArgument` not to be included by VM stuff... so overall I would love for this to be out of here

matbd: Technically this crate already depends on `move-core-types`. This conversion code is a new way to construct values.

matbd:
Putting this PR for review again.

dariorussi:
it's possible this feature is not required any longer (@davidiw for details). If that was the case how do people feel about not checking in anything at all?

davidiw:
We have moved forward by keeping an opaque blob in a vector of u8s: https://github.com/libra/libra/pull/5731 the other pr containing a vec<vec<u8>> has been closed.

matbd:
> @matbd let's wait to get this in, as it is not needed and it may not be needed ever

I don't know what changed in our use case, but it seems pretty clear that the need for richer input vectors in Move scripts is going to come back in the future.

dariorussi:
> I don't know what changed in our use case, but it seems pretty clear that the need for richer input vectors in Move scripts is going to come back in the future.

strictly speaking we have removed the need of a `vector<vector<u8>>` (see @davidiw comment). 
So right now we do not need this work and `TransactionArgument`s are fully typed. Everything is good.
If we want to push for a future, we should talk about LCS encoding and not go through this intermediate step.
I understand that is throwing this work away but if that is what we need, it seems a good thing, no?

matbd:
Let's pause this for a while

PR #5580, opened 2020-08-12
[move-prover] Modifies checking

shazqadeer:
##  Motivation

This PR implements modifies checking.  As a side effect, the handling of opaque functions is made sound with respect to aborts and modifications.  Some stdlib code that was verifying earlier (due to the unsoundness) did not verify as a result of the change.  I fixed some of these problems by adding specifications.  This PR also generalizes the usage analysis engine to track not only used memory but also modified memory.  This generalization is exploited in the implementation of modifies checking.

The new specification provided by this PR is the following annotation on Move functions:

modifies global_access_expr, ...;

Here global_access_expr is something like global<T>(address_expr) or global<T<U>>(address_expr).  The usage analysis computes the set of all top-level resource types (T as opposed to T<U>) that may be modified by a function transitively.  The specification above is used to refine this inferred modification footprint by supplying a list of global access expressions that circumscribe the modification by providing type parameters and addresses.  If function f calls function g and f refines the modification for resource T, then either g should not modify the memory for resource T at all or f should provide a refined footprint for  T.  This condition is checked by Move Prover.

The Move Prover views modifies specifications as permissions granted to a function to modify memory.  This viewpoint informs a methodology for these specifications that is constructive rather than declarative.   Specifically, the modifies specification for a function f compiles to a set of preconditions of f that must be discharged by a caller and a set of free (as opposed to checked) postconditions of f that are exploited by the caller.  This is in contrast to traditional modifies checking that simply compiles modifies specifications to checked (as opposed to free) postconditions.

To understand the methodology behind modifies specifications more concretely, let us consider a non-parameterized resource type T and a function f that calls a function g such that both modify the memory associated with T at some addresses.  The modifies specification of f can be viewed as a finite set of addresses (say, S_f) obtained by evaluating the global access expressions in the specification of f in the entry state of f.  S_g is similar.  When f calls g, S_g is evaluated and it is checked that S_g is contained in S_f via a compiled precondition of g.  Notice here that S_f is evaluated when f is entered but S_g is evaluated when g is entered.  To enable reasoning about g as an opaque function, a free postcondition is added to g. Suppose S_g = {e1, e2, ...}.  Then the added postcondition is as follows:

ensures let v1 = old(e1), v2 = old(v2),  ... in Mem_T == old(Mem_T)[v1 := Mem_T(v1)][v2 := Mem_T(v2)]...;

This postcondition uses the theory of arrays to indicate that Mem_T differs from old(Mem_T) at most at one of v1, v2, ....

Note that Move Prover also adds modifies clauses for every resource memory modified by g, as indicated by the usage analysis, to the Boogie procedure corresponding to g.  This is required for soundness.  The postcondition above is simply constraining the "havoc" introduced in the modeled semantics as a result of that modifies clause.

Yes

## Test Plan

New tests

## Related PRs

Node

### DIFF
--- a/language/move-prover/doc/user/spec-lang.md
+++ b/language/move-prover/doc/user/spec-lang.md
@@ -13,12 +13,13 @@ an overview of the design and usage of the Spec language. The reader is expected
 - [Spec Blocks](#spec-blocks)
 - [Pragmas](#pragmas)
 - [Helper Functions](#helper-functions)
+- [Assume and Assert Conditions](#assume-and-assert-conditions)
 - [AbortsIf Condition](#abortsif-condition)
 - [SucceedsIf Condition](#succeedsif-condition)
 - [Requires Condition](#requires-condition)
 - [Requires Module Condition](#requires-module-condition)
 - [Ensures Condition](#ensures-condition)
-- [Assume and Assert Conditions](#assume-and-assert-conditions)
+- [Modifies Condition](#modifies-condition)
 - [Invariant Condition on Functions](#invariant-condition-on-functions)
 - [Invariant Condition on Modules](#invariant-condition-on-modules)
 - [Invariant Condition on Structs](#invariant-condition-on-structs)
@@ -212,6 +213,63 @@ spec module {
 Is seen in the example, they can be generic. Helper functions can also access global state as done in the
 example.
 
+## Assume and Assert Conditions
+
+These conditions can only appear in spec blocks inside function bodies.
+A spec block can occur anywhere an ordinary Move statement block can occur.
+Here is an example:
+```
+fun simple1(x: u64, y: u64) {
+    let z;
+    y = x;
+    z = x + y;
+    spec {
+        assert x == y;
+        assert z == 2*x;
+    }
+}
+```
+An assert statement inside a spec block indicates a condition that must hold when control reaches that block.
+If the condition does not hold, an error is reported by the Move Prover.
+An assume statement, on the other hand, blocks executions violating the condition in the statement.
+The function `simple2` shown below is verified by the Move Prover.
+However, if the first spec block containing the assume statement is removed, Move Prover will
+show a violating to the assert statement in the second spec block.
+```
+fun simple2(x: u64, y: u64) {
+    let z: u64;
+    spec {
+        assume x > y;
+    };
+    z = x + y;
+    spec {
+        assert z > 2*y;
+    }
+}
+```
+An assert statement can also encode a loop invariant if it is placed at a loop head, as in the following example.
+```
+fun simple3(n: u64) {
+    let x = 0
+    loop {
+        spec {
+            assert x <= n;
+        };
+        if (x < n) {
+            x = x + 1
+        } else {
+            break
+        }
+    };
+    spec {
+        assert x == n;
+    }
+}
+```
+A loop invariant may comprise both assert and assume statements.
+The assume statements will be assumed at each entry into the loop while the assert statements will be
+checked at each entry into the loop.
+
 ## AbortsIf Condition
 
 The `aborts_if` condition is a spec block member which can appear only in a function context. It specifies conditions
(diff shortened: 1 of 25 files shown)


language/move-prover/tests/sources/functional/ModifiesErrorTest.move
```diff
@@ -0,0 +1,102 @@
+address 0x0 {
```
DavidLDill, line 1: It's very nice when you can see at a glance whether an error in an .exp file is intended.  We've done this in the past by adding _incorrect to function names.  It seems like the function name doesn't appear in the error messages above, though (should they?).  If you add a comment after each error line // incorrect, maybe that will make it more clear that the error is intentional?

language/move-prover/doc/user/spec-lang.md
```diff
@@ -212,6 +213,63 @@ spec module {
 Is seen in the example, they can be generic. Helper functions can also access global state as done in the
 example.
 
+## Assume and Assert Conditions
```
shazqadeer, line 216: I moved the section about in-body spec blocks before function-level spec blocks because my explanation for modifies was using assertions.

language/move-prover/tests/sources/functional/ModifiesErrorTest.move
shazqadeer: I have added a suffix "_incorrect" to names of functions on which some error is reported.

shazqadeer:
Modifies documentation has been added.

language/move-prover/src/bytecode_translator.rs
```diff
@@ -1420,6 +1491,26 @@ impl<'env> ModuleTranslator<'env> {
                         spec_translator.emit_on_update_global_invariant_assumes(memory);
                         spec_translator.save_memory_for_update_invariants(memory);
                         let memory_name = boogie_resource_memory_name(self.module_env.env, memory);
+                        if func_target.get_modify_targets_for_type(&memory).is_some() {
```
wrwg, line 1494: This looks pretty much the same as above ==> pull in a helper function?

language/move-prover/src/bytecode_translator.rs
```diff
@@ -1443,6 +1534,26 @@ impl<'env> ModuleTranslator<'env> {
                         let spec_translator = self.new_spec_translator_for_module();
                         spec_translator.save_memory_for_update_invariants(memory);
                         let memory_name = boogie_resource_memory_name(self.module_env.env, memory);
+                        if func_target.get_modify_targets_for_type(&memory).is_some() {
```
wrwg, line 1537: ... and again.

language/move-prover/src/prelude.bpl
```diff
@@ -539,7 +539,7 @@ function {:constructor} $Local(i: int): $Location;
 function {:constructor} $Param(i: int): $Location;
 
 
-// A mutable reference which also carries its current value. Since mutual references
+// A mutable reference which also carries its current value. Since mutable references
```
wrwg, line 542: Haha! Thank you.

language/move-prover/src/spec_translator.rs
```diff
@@ -618,6 +620,27 @@ impl<'env> SpecTranslator<'env> {
             distribution.add(self, for_entry_point, cond);
         }
 
+        // emit preconditions for modifies
+        for (ty, targets) in func_target.get_modify_targets() {
```
wrwg, line 624: So do we really need to emit these for indirect calls?

Consider f calls g calls h. g and h have modifies clauses. We already proven that M_g subsumes M_h. We do not need to prove it again when we verify f.

This would call for adding these conditions to the ConditionDistribution in the regular way. However, we would need to also generate different function bodies for the indirect and the direct call, where the code only in the direct case need to make the assertions on memory update. I thought doing  something similar for global invariants.

I'm fine we would do this (if my argument is correct) in a future PR, but then please add a TODO.

language/move-prover/tests/sources/functional/ModifiesTest.move
```diff
@@ -0,0 +1,83 @@
+address 0x0 {
```
wrwg, line 1: You can put them both into one file. Via the `incorrect` convention, this is at least how we do it normally in this testsuite.

language/move-prover/tests/sources/functional/ModifiesTest.move
```diff
@@ -0,0 +1,83 @@
...
+        mutate_at_test(addr2)
+    }
+    spec fun mutate_S_test {
+        requires addr1 != addr2;
+        ensures global<A::S>(addr2) == old(global<A::S>(addr2));
+        ensures global<A::S>(addr1).x == 2;
+        ensures global<T>(addr2).x == 2;
+        modifies global<A::S>(addr1), global<T>(addr2);
+    }
+}
```
wrwg, line 82: Could you add a small test with a modifies in a schema? I think it should work, but we should double check. 

And a question. At this point, the modifies would not work in schema expressions like below:

```
fun f(x: u64, a: address) { ... }
spec fun f { include x > 0 ==> ModifiesResource; } 
spec schema ModifiesResource {
   a: address;
   modifies global<Resource>(a);
}
```

We translate schema expressions by reducing them to simple conditions. For example, if there would be an `aborts_if P` in the schema, we reduce the schema and generate an `aborts_if x > 0 && P`. In order to make this work for modifies we need conditional modifies, like `modifies P ==> global<R>(a)`. Would this be hard to add?

language/stdlib/modules/DualAttestation.move
```diff
@@ -251,6 +251,7 @@ module DualAttestation {
         /// messages which fail verification and which do not.
         pragma opaque = true, verify = false;
         ensures result == spec_dual_attestation_message(payer, metadata, deposit_value);
+        aborts_if false;
```
wrwg, line 254: I probably would revert most or all of the spec changes, and just sync to the errors PR (which is now submitted). You may have an easier time reverting before rebase. As modifies is nowhere used in the library, thinks should just work.

wrwg:
> Here global_access_expr is something like global(address_expr) or global<T>(address_expr). 

I was confused by the global without type parameters. After seeing the code, I think it is not possible.

PR #5584, opened 2020-08-12
[fuzzing] fix version of libfuzzer in Cargo.lock

mimoo:


.dependabot/config.yml
```diff
@@ -7,9 +7,11 @@ update_configs:
...
           dependency_name: "codespan*"
+      - match:
+          dependency_name: "libfuzzer-sys*"
     allowed_updates:
       - match:
           update_type: "security"
     #allowed_updates:
-      #- match:
-          #dependency_type: "direct"
+    #- match:
```
rexhoffman, line 16: <shakes fist="@formatter"/>

mimoo: D:

PR #5586, opened 2020-08-12
[logging] Add ability to filter by log levels

gregnazario:
Log levels for structured logs were not being filtered automatically.
Only conversion logs were being filtered by log level.  This now
provides the ability to set `STRUCT_LOG_LEVEL` to filter by level, as
well as provides the `sl_*!` macros as a convenience for setting log
levels.

Previously `STRUCT_LOG_LEVEL` filtered logs on conversion, and not on all submission of structured logs.

### DIFF
--- a/common/logger/src/lib.rs
+++ b/common/logger/src/lib.rs
@@ -119,7 +119,8 @@ pub mod prelude {
     pub use crate::{
         crit, debug, error, event, info,
         security::{security_events, security_log},
-        send_struct_log, trace, warn, StructuredLogEntry,
+        send_struct_log, sl_debug, sl_error, sl_info, sl_trace, sl_warn, trace, warn,
+        StructuredLogEntry,
     };
 }
 pub mod json_log;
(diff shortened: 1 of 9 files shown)


common/logger/src/lib.rs
```diff
@@ -138,11 +139,14 @@ pub use text_log::{Logger, CHANNEL_SIZE, DEFAULT_TARGET};
...
 use serde::Serialize;
+use std::str::FromStr;
+
+static LOG_LEVEL_NAMES: [&str; 7] = ["OFF", "CRITICAL", "ERROR", "WARN", "INFO", "DEBUG", "TRACE"];
 
 // severity level - lower is worse
-// TODO: We can remove this if we get rid of critical
-// TODO: should we serialize it into text?
-#[derive(Serialize, Clone, Copy, Debug)]
+// TODO: Consider removing Critical, and using RustLog's level instead, as setting the log level
```
davidiw: let's get rid of crit... I don't want to be a special snowflake!

gregnazario: Done

PR #5589, opened 2020-08-12
[safety-rules] test the binary

davidiw:
This removes the nonsense that was SpawnedProcess and replaces it with a test to ensure that our binary loads -- which was the only purpose of SpawnedProcess.

As a follow up, we should consider making a test-utils crate for consensus with the hope that we can eliminate some of the testing flags in our code base.


bmwill:
I can't leave a comment on it, but it looks like this adds an empty file `consensus/safety-rules/tests/stupid_test.rs`?

consensus/safety-rules/tests/binary.rs
```diff
@@ -0,0 +1,42 @@
...
+
+const BINARY: &str = env!("CARGO_BIN_EXE_safety-rules");
+
+#[test]
+fn test_consensus_state() {
+    let mut config = NodeConfig::random().consensus.safety_rules;
+    let test_config = config.test.as_mut().unwrap();
+    let private_key = test_config.consensus_key.as_ref().unwrap().private_key();
+    let signer = ValidatorSigner::new(test_config.author, private_key);
+    let waypoint = test_utils::validator_signers_to_waypoint(&[&signer]);
```
bmwill, line 20: Is there no way to get this information without this particular test_util? If so then you could drop the requirement to build with the testing feature flag

davidiw:
> I can't leave a comment on it, but it looks like this adds an empty file `consensus/safety-rules/tests/stupid_test.rs`?

This was part of my 45 minute exploration into why my integration test wouldn't compile :D

consensus/safety-rules/tests/binary.rs
davidiw: I mean we could generate these somehow else by copypastaing stuff around, but these values are required to start the binary and keep it running.

davidiw:
> you know if you cleanup LSR you have to cleanup LEC too right?

begrudgingly

consensus/safety-rules/tests/binary.rs
bmwill: I'm specifically talking about this single function here, `test_utils::validator_signers_to_waypoint`. But yeah i understand.

bmwill:
I mean I had a diff out to fix both, but David decided he wanted to do it in a better way :D

PR #5590, opened 2020-08-12
[lec] test the binary

davidiw:
This removes the nonsense that was SpawnedProcess and replaces it with a test to ensure that our binary loads -- which was the only purpose of SpawnedProcess.


execution/execution-correctness/tests/binary.rs
```diff
@@ -0,0 +1,37 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+use execution_correctness::ExecutionCorrectnessManager;
+use libra_config::{
+    config::{ExecutionCorrectnessService, PersistableConfig, RemoteExecutionService},
+    utils,
+};
+use std::net::{IpAddr, Ipv4Addr, SocketAddr};
+
+const BINARY: &str = env!("CARGO_BIN_EXE_execution-correctness");
```
lightmark, line 11: What's this guy?

davidiw: magic: https://doc.rust-lang.org/cargo/reference/cargo-targets.html#integration-tests

bmwill: When you run an integration test, cargo will make sure that binary targets are produced so that you can execute them for testing the cli.  Cargo test's itself using this mechanism.

lightmark: TIL. Thanks @davidiw @bmwill

bmwill: There's some caveats to it, (the binary is only produced for integration tests for *this* crate) but its nice to use when you can use it. And it eliminates the nastiness of workspace-builder

davidiw:
sigh sigh....


/land

PR #5596, opened 2020-08-13
[breaking][Libra Framework] DesignatedDealer and ParentVASP account creation no…

sblackshear:
… longer accepts base_url and compliance_public_key

- Previously, creating a DesignatedDealer or ParentVASP account required specifying a (dummy, but valid) `compliance_public_key` and `base_url`. The idea is that the account owners will rotate to meaningful values after creation.
- However, this creates a problem in the case that the private key for the dummy `compliance_public_key` is known: an adversary (including AOS, the creator of the dummy key) can use this to spoof dual attestation approvals.
- Thus, this PR removes the dummy `compliance_public_key` and `base_url` at account creation altogether. Instead, these are filled with empty values that make dual attestation fail with a new `EPAYEE_COMPLIANCE_KEY_NOT_SET` error code. This change is in `DualAttestation.move` and is the only interesting part of the PR
- Once a DD or ParentVASP account has been created, the owner can rotate these fields to meaningful values (as before).

Concretely, what this does is:
- Remove the `compliance_public_key` and `base_url` parameters from the `create_parent_vasp` and `create_designated_dealer` scripts (breaking change). 
- Remove the `create_testing_account` script, which is now identical to `create_parent_vasp`
- Refactor the tests and prover specs that used the old scripts + functions


n4ss:
Awesome, that looks good on our end too!

Just a quick question: is `EPAYEE_COMPLIANCE_KEY_NOT_SET` only triggered for TR-subject transactions or all of them? Eg. can we send funds <$1000 without a compliance key

dmitri-perelman:
Btw, if we're already landing a breaking change, could we also change the DD creation to be sent by LR (instead of Blessed account)? IIUC this is the current requirement...

sblackshear:
> Awesome, that looks good on our end too!
> 
> Just a quick question: is `EPAYEE_COMPLIANCE_KEY_NOT_SET` only triggered for TR-subject transactions or all of them? Eg. can we send funds <$1000 without a compliance key

Only on TR-subject transactions; sending <$1000 before setting up a compliance key will work just fine.

sblackshear:
> Btw, if we're already landing a breaking change, could we also change the DD creation to be sent by LR (instead of Blessed account)? IIUC this is the current requirement...

No problem, will fix!

sblackshear:
> > Btw, if we're already landing a breaking change, could we also change the DD creation to be sent by LR (instead of Blessed account)? IIUC this is the current requirement...
> 
> No problem, will fix!

I started doing this and it is is actually more complex than expected/raises some issues that make me question the requirement. I think we should consider this in a separate PR to give these issues appropriate consideration.

dmitri-perelman:
> I started doing this and it is is actually more complex than expected/raises some issues that make me question the requirement. I think we should consider this in a separate PR to give these issues appropriate consideration.

Absolutely! I'm totally fine with both VASP and DD creation txns coming from the Blessed account. The LR requirements on this one is coming from your doc :)

PR #5598, opened 2020-08-13
[secure-storage] Add the ability for vault to auto-renew leases

davidiw:
If a renew_ttl is provided, the system will renew the lease immediately
upon first use and then renew it at the next half life. There are now no
direct uses of client within the code, so it is well enforced.

The only tokens that can be renewed are created tokens, so the default /
test root token cannot and will cause an error if a renew_ttl is provided.

### DIFF
--- a/config/src/config/secure_backend_config.rs
+++ b/config/src/config/secure_backend_config.rs
@@ -41,6 +41,10 @@ pub struct VaultConfig {
     /// a secret, S, without a namespace would be available in secret/data/S, with a namespace, N, it
     /// would be in secret/data/N/S.
     pub namespace: Option<String>,
+    /// Vault leverages leases on many tokens, specify this to automatically have your lease
+    /// renewed up to that many seconds more. If this is not specified, the lease will not
+    /// automatically be renewed.
+    pub renew_ttl_secs: Option<u32>,
     /// Vault's URL, note: only HTTP is currently supported.
     pub server: String,
     /// The authorization token for accessing secrets
@@ -165,6 +169,7 @@ impl From<&SecureBackend> for Storage {
                     .ca_certificate
                     .as_ref()
                     .map(|_| config.ca_certificate().unwrap()),
+                config.renew_ttl_secs,
             )),
         }
     }
@@ -187,6 +192,7 @@ mod tests {
                 server: "127.0.0.1:8200".to_string(),
                 ca_certificate: None,
                 token: Token::FromConfig("test".to_string()),
+                renew_ttl_secs: None,
             },
         };
 
@@ -211,6 +217,7 @@ vault:
                 server: "127.0.0.1:8200".to_string(),
                 ca_certificate: None,
                 token: Token::FromDisk(PathBuf::from("/token")),
+                renew_ttl_secs: None,
             },
         };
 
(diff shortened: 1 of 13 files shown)


mgorven:
Awesome stuff.

So the validator config references Vault three times (waypoint, execution key, network key). Does this result in three independent clients all of which will renew the (same) token at startup?

What if instead of renewing at startup, it queries the token metadata (https://www.vaultproject.io/api-docs/auth/token#lookup-a-token-self) to find the current TTL and calculates the renewal time from that?

davidiw:
> Awesome stuff.
> 
> So the validator config references Vault three times (waypoint, execution key, network key). Does this result in three independent clients all of which will renew the (same) token at startup?
> 
> What if instead of renewing at startup, it queries the token metadata (https://www.vaultproject.io/api-docs/auth/token#lookup-a-token-self) to find the current TTL and calculates the renewal time from that?

Hahaha, your insight is correct :D. Though arguably it shouldn't matter except look a little weird from a log perspective.

I had thought about leveraging a "tell me my current lease time" approach, but then we would still effectively have all the code that is there now with the addition of a read in the new function. Of course, this would eliminate a write on first use. I'm not super inclined to pivot but happy to be persuaded. At steady state, I can't see either approach having performance implications.

mgorven:
This should be okay for now.

secure/storage/src/vault.rs
```diff
@@ -25,6 +27,8 @@ const LIBRA_DEFAULT: &str = "libra_default";
 pub struct VaultStorage {
     pub client: Client,
```
JoshLind, line 28: Given the new client() accessor, can we make this private?

secure/storage/src/vault.rs
```diff
@@ -33,43 +37,65 @@ impl VaultStorage {
...
         Self {
             client: Client::new(host, token, certificate),
             namespace,
+            renew_ttl,
+            next_renewal: AtomicU64::new(0),
         }
     }
 
+    // Made into an accessor so we can get auto-renewal
+    fn client(&self) -> &Client {
```
JoshLind, line 51: Given the new client() accessor, can we make the client field private?

config/src/config/secure_backend_config.rs
```diff
@@ -41,6 +41,10 @@ pub struct VaultConfig {
     /// a secret, S, without a namespace would be available in secret/data/S, with a namespace, N, it
     /// would be in secret/data/N/S.
     pub namespace: Option<String>,
+    /// Vault leverages leases on many tokens, specify this to automatically have your lease
+    /// renewed up to that many seconds more. If this is not specified, the lease will not
+    /// automatically be renewed.
+    pub renew_ttl: Option<u32>,
```
JoshLind, line 47: nit: any chance we could rename this something like "renew_ttl_secs"? Just to be explicit that it is in seconds? It seems like this comment is the only thing that makes this obvious?

davidiw:
> Thanks, David! LGTM. Assuming Michael doesn't have any other concerns, the only question I have here is about testing: we could perhaps add a unit test to check that this works? For example, you could use the the token lookup API to retrieve the expiration time[1], renew the lease, lookup the token again and ensure that the time has been incremented?
> 
> [1] https://www.vaultproject.io/api-docs/auth/token#lookup-a-token-accessor

I had that originally but that exposes client... also if it didn't extend the lease, it would error on the web request.

secure/storage/src/vault.rs
davidiw: I think it should be private anyway and so should this function

PR #5600, opened 2020-08-13
[transaction builder generator] Support decoding of Move script in python

matbd:
## Motivation

* Allow decoding Move scripts in python (after Rust and Java).
* Various cleanups + minimize generated for Move decoders in Rust and Java

## Test Plan

example included

## Related PRs

depends on https://github.com/facebookincubator/serde-reflection/pull/41


language/transaction-builder/generator/examples/python3/stdlib_demo.py
```diff
@@ -12,12 +12,12 @@
 def make_address(content: bytes) -> libra.AccountAddress:
     assert len(content) == 16
     # pyre-fixme
-    return libra.AccountAddress(tuple(st.uint8(x) for x in content))
+    return libra.AccountAddress(value=tuple(st.uint8(x) for x in content))
 
 
 def main() -> None:
     token = libra.TypeTag__Struct(
```
thefallentree, line 19: feature request: can we create an top level type instead of exposing TypeTag for libra token?

so user can just call stdlib.encode_libra_token(xxxx)

matbd: Yes, I would love that to be generated as well. I put this in the "Move resources" project which should be my next big thing after we add Golang.

PR #5602, opened 2020-08-13
[vault] enable vault tests by default but guard by availability check

davidiw:
Once vault is available on CI, we can ensure that vault tests run as
part of land blocking :D


consensus/safety-rules/src/tests/vault.rs
```diff
@@ -6,22 +6,38 @@ use libra_crypto::{ed25519::Ed25519PrivateKey, Uniform};
...
 #[test]
 fn test() {
+    let storage = Storage::from(VaultStorage::new(
+        HOST.to_string(),
+        ROOT_TOKEN.to_string(),
+        None,
+        None,
+    ));
+    if let Err(e) = storage.available() {
+        println!("Vault not available: {}", e.to_string());
```
JoshLind, line 24: Do you know if this will actually show up? I was under the impression that tests only dump output on a failure, which won't be the case here?

davidiw: Yeah, I whined about the flaky test last night, I was told that it was actually disabled in master :D

println => nocapture -- sadly... I couldn't find anything in the spec that would make it obvious that the test did NOT run. I think we could explore an environmental variable that would return an error if vault didn't run.

@rexhoffman, what do you think?

LIBRA_REQUIRE_VAULT_TESTS and if set fail if available is err?

bmwill: IIRC This is what the language folks have been doing with their move prover tests. They have an environment variable that they set to "run" the tests. Same thing should be able to be applied here and if that envvar is set (which we would do in CI) and vault can't run, just panic to explicitly fail the test.

We probably want to think about the longer-term viability of this and if this is a maintainable solution but this would work in the short term.

davidiw:
@rexhoffman / @bmwill this follows a model that doesn't rely on CI as much. Please take a look. I verified it on my dev box and let's see what CI says... My dev box has vault, CI does not (yet).

PR #5603, opened 2020-08-13
[network] eliminate unreachable that is very reachable

davidiw:
We'll need to follow up:

* Heartbeat service should leverage Ping/Pong and not another Rpc service
* Should send Errors back if a remote peer is unable to process an incoming request
* Having a ProtoclId in the ErrorCode::NotSupported is ... well pretty limiting since it would make future protocols unparseable ...

### DIFF
--- a/network/src/peer/mod.rs
+++ b/network/src/peer/mod.rs
@@ -318,19 +318,11 @@ where
             oneshot::Sender<Result<(), PeerManagerError>>,
         )>,
     ) -> Result<(), PeerManagerError> {
-        trace!("Received message from Peer {}", self.peer_id().short_str(),);
+        trace!("Received message from Peer {}", self.peer_id().short_str());
         // Read inbound message from stream.
         let message = message.freeze();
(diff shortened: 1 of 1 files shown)


network/src/peer/mod.rs
```diff
@@ -343,17 +335,32 @@ where
...
                 let pong = NetworkMessage::Pong(nonce);
                 let (ack_tx, _) = oneshot::channel();
                 // Resond to a ping right away.
                 write_reqs_tx.send((pong, ack_tx)).await?;
-                Ok(())
             }
-            _ => unreachable!("Unhandled"),
-        }
+            NetworkMessage::Pong(_nonce) => {
+                // TODO(davidiw) forward to a pong pub/sub and move HealthChecker to leverage this
```
gregnazario, line 353: Let's start putting TODOs with issues associated with them!

davidiw: I plan on taking care of this one rather quickly :P but I agree

PR #5605, opened 2020-08-13
[mempool] refactor garbage collection to be sub-quadratic runtime

christinacelee:
## Motivation

The current implementation of garbage-collection in mempool has runtime O(n^2) - for each txn to be removed, we iterate through all the subsequent txns for that given account to try and park them, and worst-case it is possible that all the txns that were parked may be txns we need to remove anyway - this is inefficient. This PR refactors such that the runtime is sub-quadratic O(n log(n)): O(n log n) from sorting the gc-ready txns, and O(n) from parking all txns for an account with n txns

## Test Plan

Existing tests

### DIFF
--- a/mempool/src/core_mempool/transaction_store.rs
+++ b/mempool/src/core_mempool/transaction_store.rs
@@ -410,15 +410,28 @@ impl TransactionStore {
         };
         OP_COUNTERS.inc(index_name);
 
-        for key in index.gc(now) {
+        let mut gc_txns = index.gc(now);
+        // sort the expired txns by order of sequence number per account
+        gc_txns.sort_by_key(|key| (key.address, key.sequence_number));
+        let mut gc_iter = gc_txns.iter().peekable();
+
+        while let Some(key) = gc_iter.next() {
             if let Some(txns) = self.transactions.get_mut(&key.address) {
-                // mark all following transactions as non-ready
-                for (_, t) in txns.range((Bound::Excluded(key.sequence_number), Bound::Unbounded)) {
+                let park_range_start = Bound::Excluded(key.sequence_number);
+                let park_range_end = gc_iter
+                    .peek()
+                    .filter(|next_key| key.address == next_key.address)
+                    .map_or(Bound::Unbounded, |next_key| {
+                        Bound::Excluded(next_key.sequence_number)
+                    });
+                // mark all following txns as non-ready, i.e. park them
+                for (_, t) in txns.range((park_range_start, park_range_end)) {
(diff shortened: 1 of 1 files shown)


mempool/src/core_mempool/transaction_store.rs
```diff
@@ -410,15 +410,28 @@ impl TransactionStore {
...
+                let park_range_start = Bound::Excluded(key.sequence_number);
+                let park_range_end = gc_iter
+                    .peek()
+                    .filter(|next_key| key.address == next_key.address)
+                    .map_or(Bound::Unbounded, |next_key| {
+                        Bound::Excluded(next_key.sequence_number)
+                    });
+                // mark all following txns as non-ready, i.e. park them
+                for (_, t) in txns.range((park_range_start, park_range_end)) {
                     self.parking_lot_index.insert(&t);
```
kphfb, line 429: I could be misreading your code, but from what I can see, it seems like you are sorting by account,sequence for the expired transactions.  So let's say I have 100 transactions in my queue and txns numbered 1,2,7,9. are expired.  There are also some other people's transactions that are expired, but now gc_iter will be:
other_people_txns, my_1, my_2, my_7, my_9...

So now we hit the while loop and grab my_1.  We grab all of my transactions in mempool which would be 1->100.  We now say park_range_start is index(my_1.sequence)+1 and park_range_end in this case will be index(100) and we filtered out all of the sequence numbers of ones that were in the expired list of sequence numbers.  And we then park all of the others.  

But the problem is that let's say 11->100 aren't expired.  So we now park those all.  We then loop to my_2 and again park 11->100.  And again at my_3, etc.  So we repeat the parking still for all of 11->100 in the case of every expired transaction for this account.

christinacelee: > We now say park_range_start is index(my_1.sequence)+1 and park_range_end in this case will be index(100)

I think `park_range_end` in this case will actually be `Bound::Excluded(2)` because we'll check the next element in gc_iter, which is my_2, and set the `park_range_end` to its seq number, 2. So when the while loop is at my_1, we'll park Bound::Excluded(1) to Bound::Excluded(2), which is nothing
and when we're in the next while loop iteration at my_2, we'll park Bound::Excluded(3) to Bound::Excluded(7), etc. So each txn will be parked at most once I think?

kphfb: Ah ok, I see what you're doing now.  Yeah, that makes sense

sausagee:
LBT/CT is still having issue. I just pulled its plug.  Try to /land again.

PR #5606, opened 2020-08-13
[mempool] pop random account's latest txn from parking lot

christinacelee:
## Motivation

Popping from parking lot should be popping the last txn for a random account, but because parking lot index is ordered by `TxnPointer`, the account chosen to pop from is deterministically the last account lexicographically. This PR makes the account selection actually random. 

## Test Plan

Existing tests

### DIFF
--- a/mempool/src/core_mempool/index.rs
+++ b/mempool/src/core_mempool/index.rs
@@ -3,10 +3,12 @@
 
 /// This module provides various indexes used by Mempool
 use crate::core_mempool::transaction::{MempoolTransaction, TimelineState};
+use libra_logger::prelude::*;
 use libra_types::{account_address::AccountAddress, transaction::GovernanceRole};
+use rand::seq::SliceRandom;
 use std::{
     cmp::Ordering,
-    collections::{btree_set::Iter, BTreeMap, BTreeSet},
+    collections::{btree_set::Iter, BTreeMap, BTreeSet, HashMap},
     iter::Rev,
     ops::Bound,
     time::Duration,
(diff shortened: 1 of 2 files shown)


mempool/src/core_mempool/index.rs
```diff
@@ -256,33 +257,55 @@ impl TimelineIndex {
...
+        }
     }
 
     /// returns random "non-ready" transaction (with highest sequence number for that account)
     pub(crate) fn pop(&mut self) -> Option<TxnPointer> {
-        self.data.iter().rev().next().cloned()
+        let mut rng = rand::thread_rng();
+        self.data
+            .iter()
+            .choose(&mut rng)
```
kphfb, line 303: Hmmm, from what I can tell this is O(n) complexity.  And if we have a lot of addresses w/ values in the hashmap, I wonder if we should instead use something like https://crates.io/crates/indexmap

Thoughts?

christinacelee:
updated s.t. insert, remove, and pop are all O(1)

mempool/src/core_mempool/index.rs
```diff
@@ -256,33 +257,81 @@ impl TimelineIndex {
...
+                    if let Some((swapped_account, _)) = self.data.get(index) {
+                        self.account_indices.insert(*swapped_account, index);
+                    }
+                }
+            }
+        }
     }
 
     /// returns random "non-ready" transaction (with highest sequence number for that account)
     pub(crate) fn pop(&mut self) -> Option<TxnPointer> {
```
kphfb, line 326: When is this function actually used?  It doesn't seem to actually pop the element off, it just returns one random element?

mempool/src/core_mempool/index.rs
```diff
@@ -256,33 +257,81 @@ impl TimelineIndex {
...
     pub(crate) fn insert(&mut self, txn: &MempoolTransaction) {
-        self.data.insert(TxnPointer::from(txn));
+        let sender = &txn.txn.sender();
+        let sequence_number = txn.txn.sequence_number();
+        let is_new_entry = match self.account_indices.get(sender) {
+            Some(index) => {
+                if let Some((_account, seq_nums)) = self.data.get_mut(*index) {
+                    seq_nums.insert(sequence_number)
+                } else {
+                    unreachable!("parking lot invariant violated: for account {}, account index exists but missing entry in data", sender);
```
kphfb, line 286: This line makes me nervous.  Can we just log here and not crash?  I know it shouldn't happen, but just in case someone figures out an exploit, it seems dangerous to crash

mempool/src/core_mempool/index.rs
```diff
@@ -256,33 +257,81 @@ impl TimelineIndex {
...
+                }
+            }
+            None => {
+                let seq_nums = [sequence_number].iter().cloned().collect::<BTreeSet<_>>();
+                self.data.push((*sender, seq_nums));
+                self.account_indices.insert(*sender, self.data.len() - 1);
```
kphfb, line 292: Sorry, been too long since I've written code :/, self.account_indices.get(sender) is concurrency safe?  i.e. it locks access to the element?  Otherwise, you could reach the unreachable!() if you have two concurrent inserts

mempool/src/core_mempool/index.rs
```diff
@@ -256,33 +257,81 @@ impl TimelineIndex {
...
+
+                // maintain DS invariant
+                if txns.is_empty() {
+                    // remove account with no more txns
+                    self.data.swap_remove(index);
+                    self.account_indices.remove(sender);
+
+                    // update DS for account that was swapped in `swap_remove`
+                    if let Some((swapped_account, _)) = self.data.get(index) {
+                        self.account_indices.insert(*swapped_account, index);
```
kphfb, line 318: Doesn't this introduce a possible concurrency bug where this thread is swapping an element but the account_indices map hasn't been updated yet and another thread tries to access that element?

mempool/src/core_mempool/index.rs
christinacelee: yeah, it just returns a random element that is eligible for popping, and the actual removal happens after the function call (only place is when we're trying to evict a txn to make space for an incoming txn if mempool is already full). the name is misleading (this method is a getter not a mutator), I can rename it to something like `get_poppable`?

mempool/src/core_mempool/index.rs
christinacelee: I think all of core mempool is wrapped up in a mutex when accessed, so we don't have to worry about making the internal workings of core mempool concurrency-safe (in fact I'm p sure currently no aspect of core mempool's internal DS's are individually concurrency-safe)

mempool/src/core_mempool/index.rs
kphfb: Yeah let's rename it to that

PR #5610, opened 2020-08-13
[network] Reduce network log frequency

gregnazario:
Decreasing duplicate logs where sometimes we'd log an entry message and
then immediately different paths which acted like a trace log.  In
addition, remove usage of the crit! macro.


gregnazario:
> error / crit seem strange for exiting a loop when we should expect that upon normal termination, no?

Yeah, it is strange.  I guess the thought is it that it would be an error if it unexpectedly happened.  Except if it happens on purpose, obviously it's not an error.

Issue #5597, opened 2020-08-13
[logging] Reduce commonly overly logged logs


gregnazario:
### Overview
After doing some experiments with setting the network address to an invalid network address, I found that logs started being dropped due to high log volume.  The main culprits being consensus and networking.  There was up to a 30% drop in metrics as a result of a burst of large amounts of logs.
http://logging.premainnet-test.aws.hlw3truzy4ls.com/_plugin/kibana/goto/c1179343aa35fc9422be1276dba5d603

### Tasks
- [x] Reduce EpochManager logs
- [x] Reduce PeerManager logs

#### Reduce PeerManager Logs
These logs were overly verbose, and were originally at a `trace` level.  This has been corrected, and structured log leveling implemented here (https://github.com/libra/libra/pull/5586)

gregnazario:
Peer manager reducing logs https://github.com/libra/libra/pull/5610

PR #5614, opened 2020-08-13
[consensus] return Err with successful timeout

zekun000:
## Motivation

This commit makes `process_local_timeout` returns Err when the messages
are broadcasted.

It can help reduce the log amount without losing visibility. After this
change, we only log round state when errors occur.

Y

## Test Plan

## Related PRs
https://github.com/libra/libra/issues/5597

### DIFF
--- a/consensus/src/epoch_manager.rs
+++ b/consensus/src/epoch_manager.rs
@@ -502,7 +502,7 @@ impl EpochManager {
             self.start_processor(payload).await;
         }
         loop {
-            if let Err(e) = monitor!(
+            let result = monitor!(
                 "main_loop",
                 select! {
                     payload = reconfig_events.select_next_some() => {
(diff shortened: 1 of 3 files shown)


consensus/src/round_manager.rs
```diff
@@ -399,19 +399,20 @@ impl RoundManager {
     /// 1) In case a validator has previously voted in this round, it repeats the same vote and sign
     /// a timeout.
     /// 2) Otherwise vote for a NIL block and sign a timeout.
+    /// Note this function returns Err even if messages are broadcasted successfully because timeout
+    /// is considered as error. It only returns Ok(()) when the timeout is stale.
```
gregnazario, line 403: Okay so timeouts are now Err, and everything downstream handles that?

consensus/src/epoch_manager.rs
```diff
@@ -519,12 +519,18 @@ impl EpochManager {
...
+                p.round_state().to_string()
+            } else {
+                "RoundState: None".into()
+            };
+            match result {
+                Ok(_) => trace!("{}", round_state),
+                Err(e) => {
+                    counters::ERROR_COUNT.inc();
+                    error!("{:?}, {}", e, round_state);
+                }
```
gregnazario, line 533: Nice, hopefully this won't log so much.  Was this being hit so much just because a lot of messages were coming through?  Or was it a combination with this and the Round manager differences

consensus/src/round_manager.rs
zekun000: it's a high level function, the only downstream is that main loop

consensus/src/epoch_manager.rs
zekun000: I haven't figured out why wrong dns would trigger the log spamming, it can only come from many more messages. probably need to dig further after this change.

PR #5617, opened 2020-08-13
[safety-rules, consensus, ...] Remove a few panics

huitseeker:
There are rumors that people don't like panics, specially in key areas such as the TCB.

Just to map out the feasibility of tooling that would make those panics harder to introduce, this PR aims at starting a conversation by removing a bunch of panics, through folding them in the `Error` case of their enclosing `Result`.

Kind reviewer:
If you see one of your favorite panics listed in this PR, now's the time to defend it. But more importantly, if you can say a word on how *that* panic represents a legitimate use case you'd want in high-assurance code, that would be even more helpful.

### DIFF
--- a/consensus/src/metrics_safety_rules.rs
+++ b/consensus/src/metrics_safety_rules.rs
@@ -32,7 +32,9 @@ impl MetricsSafetyRules {
         let proofs = self
             .storage
             .retrieve_epoch_change_proof(sr_waypoint.version())
-            .expect("Unable to retrieve Waypoint state from Storage");
+            .map_err(|_| {
+                Error::InternalError("Unable to retrieve Waypoint state from storage".into())
+            })?;
         self.initialize(&proofs)
     }
 }
--- a/consensus/src/metrics_safety_rules.rs
+++ b/consensus/src/metrics_safety_rules.rs
@@ -32,8 +32,11 @@ impl MetricsSafetyRules {
         let proofs = self
             .storage
             .retrieve_epoch_change_proof(sr_waypoint.version())
-            .map_err(|_| {
-                Error::InternalError("Unable to retrieve Waypoint state from storage".into())
+            .map_err(|e| {
+                Error::InternalError(format!(
+                    "Unable to retrieve Waypoint state from storage, encountered Error:{}",
+                    e
+                ))
             })?;
         self.initialize(&proofs)
     }
--- a/language/move-lang/src/shared/mod.rs
+++ b/language/move-lang/src/shared/mod.rs
@@ -43,7 +43,7 @@ impl Address {
         }
 
         let mut result = hex::decode(hex_string.as_str())
-            .map_err(|_| "hex string does not decode properly".to_owned())?;
+            .map_err(|e| format!("hex string {} fails to decode with Error {}", hex_string, e))?;
         let len = result.len();
         if len < ADDRESS_LENGTH {
             result.reverse();
--- a/language/move-lang/src/shared/mod.rs
+++ b/language/move-lang/src/shared/mod.rs
@@ -42,7 +42,8 @@ impl Address {
             hex_string.insert(0, '0');
         }
 
-        let mut result = hex::decode(hex_string.as_str()).unwrap();
+        let mut result = hex::decode(hex_string.as_str())
+            .map_err(|_| "hex string does not decode properly".to_owned())?;
         let len = result.len();
         if len < ADDRESS_LENGTH {
             result.reverse();
--- a/language/move-prover/stackless-bytecode-generator/src/borrow_analysis.rs
+++ b/language/move-prover/stackless-bytecode-generator/src/borrow_analysis.rs
@@ -488,7 +488,10 @@ impl<'a> BorrowAnalysis<'a> {
         let livevar_annotation_at = self
             .livevar_annotation
             .get_live_var_info_at(code_offset)
-            .unwrap();
+            .ok_or_else(|| PackError {
+                code_offset,
+                indices,
+            })?;
         for idx in livevar_annotation_at
             .before
             .difference(&livevar_annotation_at.after)
--- a/language/move-prover/stackless-bytecode-generator/src/packref_analysis.rs
+++ b/language/move-prover/stackless-bytecode-generator/src/packref_analysis.rs
@@ -94,7 +94,7 @@ impl<'a> CopyAnalysis<'a> {
         let after_refs = &self
             .borrow_annotation
             .get_borrow_info_at(code_offset)
-            .unwrap()
+            .ok_or_else(|| ())?
             .after
             .all_refs();
         post.copies = post.copies.intersection(after_refs).cloned().collect();
--- a/language/move-prover/stackless-bytecode-generator/src/packref_analysis.rs
+++ b/language/move-prover/stackless-bytecode-generator/src/packref_analysis.rs
@@ -94,7 +94,7 @@ impl<'a> CopyAnalysis<'a> {
         let after_refs = &self
             .borrow_annotation
             .get_borrow_info_at(code_offset)
-            .ok_or_else(|| ())?
+            .unwrap()
             .after
             .all_refs();
         post.copies = post.copies.intersection(after_refs).cloned().collect();
--- a/secure/storage/src/github.rs
+++ b/secure/storage/src/github.rs
@@ -34,8 +34,11 @@ impl KVStorage for GitHubStorage {
     fn get<T: DeserializeOwned>(&self, key: &str) -> Result<GetResponse<T>, Error> {
         let data = self.client.get_file(key)?;
         let data = base64::decode(&data)?;
-        let data = std::str::from_utf8(&data).map_err(|_| {
-            Error::InternalError("Unparseable data returned from Github KV Storage".into())
+        let data = std::str::from_utf8(&data).map_err(|e| {
+            Error::InternalError(format!(
+                "Unparseable data: {:?}\n returned from Github KV Storage, met Error:{}",
+                data, e
+            ))
         })?;
         serde_json::from_str(&data).map_err(|e| e.into())
     }
--- a/secure/storage/src/github.rs
+++ b/secure/storage/src/github.rs
@@ -34,7 +34,9 @@ impl KVStorage for GitHubStorage {
     fn get<T: DeserializeOwned>(&self, key: &str) -> Result<GetResponse<T>, Error> {
         let data = self.client.get_file(key)?;
         let data = base64::decode(&data)?;
-        let data = std::str::from_utf8(&data).unwrap();
+        let data = std::str::from_utf8(&data).map_err(|_| {
+            Error::InternalError("Unparseable data returned from Github KV Storage".into())
+        })?;
         serde_json::from_str(&data).map_err(|e| e.into())
     }
 
--- a/secure/storage/src/vault.rs
+++ b/secure/storage/src/vault.rs
@@ -329,9 +329,9 @@ impl CryptoStorage for VaultStorage {
     ) -> Result<Ed25519Signature, Error> {
         let name = self.crypto_name(name);
         let mut bytes = <T::Hasher as libra_crypto::hash::CryptoHasher>::seed().to_vec();
-        lcs::serialize_into(&mut bytes, &message)
-            .map_err(|_| libra_crypto::traits::CryptoMaterialError::SerializationError)
-            .expect("Serialization of signable material should not fail.");
+        lcs::serialize_into(&mut bytes, &message).map_err(|_| {
+            Error::InternalError("Serialization of signable material should not fail.".into())
+        })?;
         Ok(self.client().sign_ed25519(&name, &bytes, None)?)
     }
 
@@ -344,9 +344,9 @@ impl CryptoStorage for VaultStorage {
         let name = self.crypto_name(name);
         let vers = self.key_version(&name, &version)?;
         let mut bytes = <T::Hasher as libra_crypto::hash::CryptoHasher>::seed().to_vec();
-        lcs::serialize_into(&mut bytes, &message)
-            .map_err(|_| libra_crypto::traits::CryptoMaterialError::SerializationError)
-            .expect("Serialization of signable material should not fail.");
+        lcs::serialize_into(&mut bytes, &message).map_err(|_| {
+            Error::InternalError("Serialization of signable material should not fail.".into())
+        })?;
         Ok(self.client().sign_ed25519(&name, &bytes, Some(vers))?)
     }
 }
--- a/secure/storage/src/vault.rs
+++ b/secure/storage/src/vault.rs
@@ -329,8 +329,11 @@ impl CryptoStorage for VaultStorage {
     ) -> Result<Ed25519Signature, Error> {
         let name = self.crypto_name(name);
         let mut bytes = <T::Hasher as libra_crypto::hash::CryptoHasher>::seed().to_vec();
-        lcs::serialize_into(&mut bytes, &message).map_err(|_| {
-            Error::InternalError("Serialization of signable material should not fail.".into())
+        lcs::serialize_into(&mut bytes, &message).map_err(|e| {
+            Error::InternalError(format!(
+                "Serialization of signable material should not fail, yet returned Error:{}",
+                e
+            ))
         })?;
         Ok(self.client().sign_ed25519(&name, &bytes, None)?)
     }
@@ -344,8 +347,11 @@ impl CryptoStorage for VaultStorage {
         let name = self.crypto_name(name);
         let vers = self.key_version(&name, &version)?;
         let mut bytes = <T::Hasher as libra_crypto::hash::CryptoHasher>::seed().to_vec();
-        lcs::serialize_into(&mut bytes, &message).map_err(|_| {
-            Error::InternalError("Serialization of signable material should not fail.".into())
+        lcs::serialize_into(&mut bytes, &message).map_err(|e| {
+            Error::InternalError(format!(
+                "Serialization of signable material should not fail, yet returned Error:{}",
+                e
+            ))
         })?;
         Ok(self.client().sign_ed25519(&name, &bytes, Some(vers))?)
     }
--- a/storage/inspector/src/main.rs
+++ b/storage/inspector/src/main.rs
@@ -3,7 +3,7 @@
 
 #![forbid(unsafe_code)]
 
-use anyhow::{anyhow, Result};
+use anyhow::Result;
 use libra_logger::info;
 use libradb::LibraDB;
 use std::path::PathBuf;
@@ -44,7 +44,7 @@ enum Command {
 fn print_head(db: &LibraDB) -> Result<()> {
     let si = db
         .get_startup_info()
-        .map_err(|_| anyhow!("Can't get startup info."))?
+        .expect("Can't get startup info")
         .expect("StartupInfo is empty, database is empty.");
 
     let version = si.latest_ledger_info.ledger_info().version();
--- a/storage/inspector/src/main.rs
+++ b/storage/inspector/src/main.rs
@@ -3,7 +3,7 @@
 
 #![forbid(unsafe_code)]
 
-use anyhow::Result;
+use anyhow::{anyhow, Result};
 use libra_logger::info;
 use libradb::LibraDB;
 use std::path::PathBuf;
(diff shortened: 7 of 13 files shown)


JoshLind:
Overall this makes sense to me. Going forward, it would be nice if we had a linter to prevent such cases (e.g., where a panic is possible in a function that explicitly returns a Result<..>).

consensus/src/test_utils/mock_state_computer.rs
```diff
@@ -101,7 +100,7 @@ impl StateComputer for MockStateComputer {
             .commit_to_storage(commit.ledger_info().clone());
         self.commit_callback
             .unbounded_send(commit)
-            .expect("Fail to notify about sync");
+            .ok_or_else(|| anyhow!("Failed to notify about sync"))?;
```
davidiw: nit, this is a test utils, probably expect is more desirable?

consensus/src/metrics_safety_rules.rs
```diff
@@ -32,7 +32,9 @@ impl MetricsSafetyRules {
         let proofs = self
             .storage
             .retrieve_epoch_change_proof(sr_waypoint.version())
-            .expect("Unable to retrieve Waypoint state from Storage");
+            .map_err(|_| {
+                Error::InternalError("Unable to retrieve Waypoint state from storage".into())
+            })?;
```
davidiw, line 37: this is interesting, do we want consensus (libra-node) to crash if it cannot sync LSR -- cc: @zekun000

language/move-lang/src/shared/mod.rs
```diff
@@ -42,7 +42,8 @@ impl Address {
             hex_string.insert(0, '0');
         }
 
-        let mut result = hex::decode(hex_string.as_str()).unwrap();
+        let mut result = hex::decode(hex_string.as_str())
+            .map_err(|_| "hex string does not decode properly".to_owned())?;
```
davidiw, line 46: thanks!

secure/storage/src/vault.rs
```diff
@@ -47,7 +47,9 @@ impl VaultStorage {
             if secret.ends_with('/') {
                 self.reset_kv(&secret)?;
             } else {
-                self.client.delete_secret(&format!("{}{}", path, secret))?;
+                self.client
+                    .delete_secret(&format!("{}{}", path, secret))
+                    .map_err(|_| Error::InternalError("Unable to initialize storage".into()))?;
```
davidiw, line 52: I don't follow?

secure/storage/src/vault.rs
```diff
@@ -304,9 +306,9 @@ impl CryptoStorage for VaultStorage {
     ) -> Result<Ed25519Signature, Error> {
         let name = self.crypto_name(name);
         let mut bytes = <T::Hasher as libra_crypto::hash::CryptoHasher>::seed().to_vec();
-        lcs::serialize_into(&mut bytes, &message)
-            .map_err(|_| libra_crypto::traits::CryptoMaterialError::SerializationError)
-            .expect("Serialization of signable material should not fail.");
```
davidiw, line 309: I think this is an expect, why would it fail? If it fails, I want the system to tell me and not make further progress.

secure/storage/src/vault.rs
```diff
@@ -319,9 +321,9 @@ impl CryptoStorage for VaultStorage {
         let name = self.crypto_name(name);
         let vers = self.key_version(&name, &version)?;
         let mut bytes = <T::Hasher as libra_crypto::hash::CryptoHasher>::seed().to_vec();
-        lcs::serialize_into(&mut bytes, &message)
-            .map_err(|_| libra_crypto::traits::CryptoMaterialError::SerializationError)
-            .expect("Serialization of signable material should not fail.");
```
davidiw, line 324: Same

storage/inspector/src/main.rs
```diff
@@ -43,8 +43,7 @@ enum Command {
 /// Print out latest information stored in the DB.
 fn print_head(db: &LibraDB) -> Result<()> {
     let si = db
-        .get_startup_info()
-        .expect("Can't get startup info")
```
davidiw, line 47: This might result in an unclear error.

testsuite/cluster-test/src/aws.rs
```diff
@@ -36,7 +36,7 @@ pub async fn set_asg_size(
     };
     let credentials_provider = WebIdentityProvider::from_k8s_env();
 
-    let dispatcher = rusoto_core::HttpClient::new().expect("failed to create request dispatcher");
+    let dispatcher = rusoto_core::HttpClient::new()?;
```
davidiw, line 39: This might result in an unclear error.

testsuite/cluster-test/src/experiments/cpu_flamegraph.rs
```diff
@@ -66,7 +66,7 @@ impl Experiment for CpuFlamegraph {
             .tx_emitter
             .emit_txn_for(tx_emitter_duration, emit_job_request)
             .boxed();
-        let run_id = env::var("RUN_ID").expect("RUN_ID is not set");
+        let run_id = env::var("RUN_ID")?;
```
davidiw, line 69: This might result in an unclear error.

testsuite/cluster-test/src/github.rs
```diff
@@ -40,7 +40,7 @@ impl GitHub {
     /// Paging is not implemented yet
     pub fn get_commits(&self, repo: &str, sha: &str) -> Result<Vec<CommitInfo>> {
         let url = format!("https://api.github.com/repos/{}/commits?sha={}", repo, sha);
-        let url: Url = url.parse().expect("Failed to parse github url");
+        let url: Url = url.parse()?;
```
davidiw, line 43: This might result in an unclear error.

testsuite/cluster-test/src/prometheus.rs
```diff
@@ -55,17 +55,13 @@ impl Prometheus {
...
-        let url = self
-            .url
-            .join(&format!(
-                "api/v1/query_range?query={}&start={}&end={}&step={}",
-                query,
-                start.as_secs(),
-                end.as_secs(),
-                step
-            ))
-            .expect("Failed to make query_range url");
```
davidiw, line 67: This might result in an unclear error.

huitseeker:
> Overall this makes sense to me. Going forward, it would be nice if we had a linter to prevent such cases (e.g., where a panic is possible in a function that explicitly returns a Result<..>).

Yeah, creating it is the idea, but feedback on specifics helps scope out the least intrusive way to make it happen.

secure/storage/src/vault.rs
huitseeker: :facepalm:

consensus/src/test_utils/mock_state_computer.rs
huitseeker: :+1:

mimoo:
> Overall this makes sense to me. Going forward, it would be nice if we had a linter to prevent such cases (e.g., where a panic is possible in a function that explicitly returns a Result<..>).

we're working on it :)

secure/storage/src/vault.rs
mimoo: why don't you log a MUST_FIX instead so that you can continue to perform the other operations that still work? Also maybe you will retry with a different message and it'll work. No reason to panic if we can work on the fix while this continues to run : o

storage/inspector/src/main.rs
mimoo: suggest adding a `map_err()` then : o

secure/storage/src/vault.rs
huitseeker: It is my understanding that:
- we do not have, either in signing or secure storage, comprehensive pre-execution checks of recursion depths, in the way that the language / VM does (#4993 #5023)¹
- yet signables are very much under the same serialization limit as everybody else (#4928), which makes at least one plausible reason where serialization could fail,
- I don't know for sure if we have any guarantees that at this point of call the contents message is controlled by the current process (rather than external input),
- it's my understanding this panic will at least irrecoverably blow up the secure storage service, making the node fail and become byzantine.

This seems to look — and please correct me if I'm wrong — like a textbook vulnerability: any attacker able to make generate a value that will fail serialization knocks down the node.

On the other hand, could you remind me of the issues that would be created in not managing to sign this particular message, and trying to move on to the next one nonetheless? 

¹: are they planned? have I missed something? /cc @matbd

consensus/src/metrics_safety_rules.rs
hermanventer: There are several considerations here and I'd like to weigh in on some of them. First of all `expect`. This is not used consistently, which makes it hard to read. Much of what I've seen to date, suggests that people use it as "it is a programming mistake if this happens, but I pretty much guarantee that it wont". It can also mean "this might happen, but if it does things are hosed and the best we can do is to crash here".

The proposed change introduces yet another meaning: "This might happen and if it does, the caller can decide what to do about it." (This is not unreasonable, if an invariant is found to be violated, a program can carry on if there is a reasonable way to repair the program state so that the invariant is restored. Perhaps the caller can do this, but in this case I think David is saying, he is not so sure.)

I don't like it when things are ambiguous like this. It makes the code hard to read and it makes it difficult for an outsider to flag this as a potential error since it refers to considerations that are either undocumented, or documented elsewhere in a way that requires a substantial up front cost before you can properly read the code.

My recommendation is the make things as explicit as possible. 

- If it is the case that the caller can conceivably repair the program state, then return the error.
- If this can only happen in a situation where the process should stop immediately, then do `.unwrap_or_else(|| unrecoverable!(".."))`
- If this should not happen at all (i.e. it is a programming mistake if it happens) then explicitly do a `checked_assume!(proofs.is_some()); // justification for the assumption`

davidiw: @fga, I think the easy way to support doing this is to state that the only use of perform_initialize itself has an expect to resolve the result and therefor this ehis expect is duplicitous. At the same time, however, you are losing the internal error string, why was it unable, was that message meaningful. We are losing some context.

language/move-lang/src/shared/mod.rs
davidiw: could we add more context, what was the string? what was the internal error?

language/move-prover/stackless-bytecode-generator/src/packref_analysis.rs
```diff
@@ -94,7 +94,7 @@ impl<'a> CopyAnalysis<'a> {
         let after_refs = &self
             .borrow_annotation
             .get_borrow_info_at(code_offset)
-            .unwrap()
```
davidiw, line 97: we're just throwing away the error now, is this safe?

secure/storage/src/github.rs
```diff
@@ -33,7 +33,9 @@ impl KVStorage for GitHubStorage {
     fn get(&self, key: &str) -> Result<GetResponse, Error> {
         let data = self.client.get_file(key)?;
         let data = base64::decode(&data)?;
-        let data = std::str::from_utf8(&data).unwrap();
+        let data = std::str::from_utf8(&data).map_err(|_| {
+            Error::InternalError("Unparseable data returned from Github KV Storage".into())
```
davidiw, line 37: why is it unparseable? what was the data?

secure/storage/src/vault.rs
davidiw: what is the internal error? lost context.

secure/storage/src/vault.rs
davidiw: lost context

storage/inspector/src/main.rs
```diff
@@ -44,7 +44,7 @@ enum Command {
 fn print_head(db: &LibraDB) -> Result<()> {
     let si = db
         .get_startup_info()
-        .expect("Can't get startup info")
+        .map_err(|_| anyhow!("Can't get startup info."))?
```
davidiw, line 47: what happens if we return an error here? this is also a utility app, I'm not sure I see the point in this modificatoin.

testsuite/cli/libra-wallet/src/key_factory.rs
```diff
@@ -161,8 +161,7 @@ impl KeyFactory {
 
         let hkdf_expand = Hkdf::<Sha3_256>::expand(&self.main(), Some(&info), 32)?;
         let sk = Ed25519PrivateKey::try_from(hkdf_expand.as_slice())
-            .expect("Unable to convert into private key");
-
+            .map_err(|_| anyhow!("Unable to convert into private key"))?;
```
davidiw, line 164: lost context ...

testsuite/cli/libra-wallet/src/mnemonic.rs
```diff
@@ -77,7 +77,9 @@ impl Mnemonic {
         bit_writer.write_buffer();
 
         // This will never fail as we've already checked the word-list is not empty.
-        let (checksum, entropy) = bit_writer.bytes.split_last().unwrap();
+        let (checksum, entropy) = bit_writer.bytes.split_last().ok_or_else(|| {
+            WalletError::LibraWalletGeneric("Bit writer producing inconsistent byte strings".into())
```
davidiw, line 81: what happens if we hit this? is it really safe to continue on in the libra-wallet? how do we handle this upstream?

testsuite/cluster-test/src/aws.rs
```diff
@@ -36,7 +36,8 @@ pub async fn set_asg_size(
     };
     let credentials_provider = WebIdentityProvider::from_k8s_env();
 
-    let dispatcher = rusoto_core::HttpClient::new().expect("failed to create request dispatcher");
+    let dispatcher = rusoto_core::HttpClient::new()
+        .map_err(|_| anyhow!("Failed to create request dispatcher"))?;
```
davidiw, line 40: maybe this allows an early exit rather than wasting time retrying?

testsuite/cluster-test/src/experiments/cpu_flamegraph.rs
```diff
@@ -66,7 +66,7 @@ impl Experiment for CpuFlamegraph {
             .tx_emitter
             .emit_txn_for(tx_emitter_duration, emit_job_request)
             .boxed();
-        let run_id = env::var("RUN_ID").expect("RUN_ID is not set");
```
davidiw, line 69: what does this obtain? what happens if we error here instead of expect?

testsuite/cluster-test/src/github.rs
```diff
@@ -40,7 +40,9 @@ impl GitHub {
     /// Paging is not implemented yet
     pub fn get_commits(&self, repo: &str, sha: &str) -> Result<Vec<CommitInfo>> {
         let url = format!("https://api.github.com/repos/{}/commits?sha={}", repo, sha);
-        let url: Url = url.parse().expect("Failed to parse github url");
+        let url: Url = url
+            .parse()
+            .map_err(|_| anyhow!("Failed to parse github url"))?;
```
davidiw, line 45: why not print the url?

testsuite/cluster-test/src/prometheus.rs
davidiw: how is this handled upstream? does the error improve or degrade the ux?

consensus/src/metrics_safety_rules.rs
hermanventer: If you want to maintain the behavior of expect, then terminate the execution in this context by all means. I would recommend, however, that instead of expect, you either unwrap_or_else or do a checked_assume. That makes it clear whether you expect the unwrap to succeed or whether you expect to to sometimes fail. Making this distinction clear to the reader (which most likely is your future self) will be a great help to them.

testsuite/cluster-test/src/experiments/cpu_flamegraph.rs
huitseeker: We run the other experiments in the `Vec<Box<dyn Experiment>>` list rather than stopping *everything* at this one.

testsuite/cluster-test/src/aws.rs
huitseeker: All three calls to `set_asg_size` are `map_err`-ed or `unwrap_or_else`-ed, which makes me think @andll did not want them to panic out of a failed tls negotiation. There's no retry AFAICT.

huitseeker:
> Using errors almost certainly eliminates the backtrace

For many Error types (including anyhow which is used in many of these examples):

If you want panics and errors to both have backtraces, set RUST_BACKTRACE=1;
If you want only errors to have backtraces, set RUST_LIB_BACKTRACE=1;
If you want only panics to have backtraces, set RUST_BACKTRACE=1 and RUST_LIB_BACKTRACE=0.

The tracking issue for having this in the std lib is https://github.com/rust-lang/rust/issues/53487

davidiw:
> > Using errors almost certainly eliminates the backtrace
> 
> For many Error types which use `std::backtrace` (including anyhow which is used in many of these examples and throughout the code base):
> 
> If you want panics and errors to both have backtraces, set RUST_BACKTRACE=1;
> If you want only errors to have backtraces, set RUST_LIB_BACKTRACE=1;
> If you want only panics to have backtraces, set RUST_BACKTRACE=1 and RUST_LIB_BACKTRACE=0.
> 
> The tracking issue for having this in the std lib is [rust-lang/rust#53487](https://github.com/rust-lang/rust/issues/53487)

Errors do not have backtraces unless they are compiled with the nightly compiler :/ https://doc.rust-lang.org/std/error/trait.Error.html#method.backtrace

so the only way to get them is via expect statements or rolling your own Error type that has backtrace -- something I wish we had done much earlier.

consensus/src/metrics_safety_rules.rs
```diff
@@ -31,9 +31,14 @@ impl MetricsSafetyRules {
...
-            .expect("Unable to retrieve Waypoint state from Storage");
-        self.initialize(&proofs)
+            .retrieve_epoch_change_proof(sr_waypoint.version());
+        if let Err(e) = proofs {
+            panic!(
+                "Unable to retrieve Waypoint state from storage, encountered Error:{}",
+                e
+            )
+        }
+        self.initialize(&proofs.unwrap())
```
davidiw, line 41: For this one, I think just returning an InternalError(format!(....)) would suffice. I don't think this is particularly more readable than expect. what do we gain by this indirection?

testsuite/cli/libra-wallet/src/mnemonic.rs
```diff
@@ -77,7 +77,10 @@ impl Mnemonic {
         bit_writer.write_buffer();
 
         // This will never fail as we've already checked the word-list is not empty.
-        let (checksum, entropy) = bit_writer.bytes.split_last().unwrap();
+        let (checksum, entropy) = bit_writer
+            .bytes
+            .split_last()
+            .unwrap_or_else(|| unreachable!());
```
davidiw, line 83: I know @hermanventer really likes this, but -- honestly -- it feels like an anti-pattern that we avoid facets that the language provides... I think expect is okay in "production" code ... unwrap maybe we could agree is bad.

testsuite/cluster-test/src/experiments/cpu_flamegraph.rs
davidiw: yeah, so this seems bad, since if the run_id isn't set here, why would it be set elsewhere? cc: @andll / @snakech 

I'd prefer these folks to weigh in...

consensus/src/metrics_safety_rules.rs
davidiw: I'm not sure I follow. I would think unwrap is maybe unclear, but an expect statement that doesn't state why this is expected is a fail. At the same time, when using locks, folks use unwrap religiously. It seems a shame to add a lot of boiler plate code where ideally where we already have unwrap / expect :/

testsuite/cli/libra-wallet/src/mnemonic.rs
huitseeker: Here, this is just a contraction of:
```
match bit_writer.bytes.split_last() {
  Some(tuple) => tuple,
  None => unreachable!()
}
```

except it equips the compiler with the proper kind of information to perform unreachability analysis or use [hardware traps](https://github.com/rust-lang/rust/pull/45920)

language/move-prover/stackless-bytecode-generator/src/borrow_analysis.rs
```diff
@@ -488,7 +488,10 @@ impl<'a> BorrowAnalysis<'a> {
         let livevar_annotation_at = self
             .livevar_annotation
             .get_live_var_info_at(code_offset)
-            .unwrap();
+            .ok_or_else(|| PackError {
+                code_offset,
+                indices,
+            })?;
```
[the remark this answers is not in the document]
huitseeker: The move prover runs under a very special context, so I think the answer is no. The violation of `live_var` invariants would lead to a failure of borrow analysis, which would be investigated on the spot anyway.

/cc @wrwg

PR #5619, opened 2020-08-13
[logging] add unique ids to every log request

gregnazario:
As a step towards preventing duplication of log messages, I'm adding a
unique ID to every log message.  At first, this will allow us to tell if
something is duplicated.  Afterwards, we should be able to integrate
with logstash to use these IDs as the database unique IDs.

### DIFF
--- a/common/logger/src/struct_log.rs
+++ b/common/logger/src/struct_log.rs
@@ -7,6 +7,7 @@ use crate::counters::{
 };
 use chrono::{SecondsFormat, Utc};
 use once_cell::sync::Lazy;
+use rand::RngCore;
 use serde::Serialize;
 use serde_json::Value;
 use std::{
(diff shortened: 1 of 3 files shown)


common/logger/src/struct_log.rs
```diff
@@ -53,6 +54,8 @@ static FIELDS_TO_KEEP: &[&str] = &["error"];
 
 #[derive(Default, Serialize)]
 pub struct StructuredLogEntry {
+    /// Unique Id representing this message in Elasticsearch
+    id: String,
```
davidiw, line 58: how expensive is uuid generation?

gregnazario: It's a `thread_rng` for 16 bytes.  I've basically copied this and removed the extra versioning and information that isn't really needed from the unique ID for this use case.

bmwill: One small improvement would be to avoid the extra heap allocation and make this a fixed length array of bytes (if you know that this will always be 16 bytes for example), but that's probably splitting hairs.

PR #5620, opened 2020-08-14
[storage] DB::open_as_secondary()

msmouse:
## Motivation

So we can run read only tools while the node is still running.

Y
## Test Plan

unit test.

### DIFF
--- a/storage/libradb/src/lib.rs
+++ b/storage/libradb/src/lib.rs
@@ -75,7 +75,7 @@ use libra_types::{
     },
 };
 use once_cell::sync::Lazy;
-use schemadb::{DB, DEFAULT_CF_NAME};
+use schemadb::{ColumnFamilyName, DB, DEFAULT_CF_NAME};
 use std::{iter::Iterator, path::Path, sync::Arc, time::Instant};
 use storage_interface::{DbReader, DbWriter, Order, StartupInfo, TreeState};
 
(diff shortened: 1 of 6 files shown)


storage/libradb/src/lib.rs
```diff
@@ -109,12 +109,8 @@ pub struct LibraDB {
 }
 
 impl LibraDB {
-    pub fn open<P: AsRef<Path> + Clone>(
-        db_root_path: P,
-        readonly: bool,
-        prune_window: Option<u64>,
-    ) -> Result<Self> {
-        let column_families = vec![
+    fn column_families() -> Vec<ColumnFamilyName> {
```
zekun000, line 112: const fn?

storage/libradb/src/lib.rs
```diff
@@ -127,32 +123,67 @@ impl LibraDB {
...
+            state_store: Arc::new(StateStore::new(Arc::clone(&db))),
+            transaction_store: Arc::new(TransactionStore::new(Arc::clone(&db))),
+            system_store: SystemStore::new(Arc::clone(&db)),
+            pruner: prune_window.map(|n| Pruner::new(Arc::clone(&db), n)),
+        }
+    }
+
+    pub fn open<P: AsRef<Path> + Clone>(
+        db_root_path: P,
+        readonly: bool,
```
thefallentree, line 145: should we make readonly always use "open as secondary ?"

storage/libradb/src/lib.rs
msmouse: It's allocating a `Vec` on the heap.. so can't be const..

storage/libradb/src/lib.rs
msmouse: We'd need to update call sites to pass in the secondary path. (Using a tempdir is not always the right thing to do.)

msmouse:
@zekun000  走好不送

/land

PR #5621, opened 2020-08-14
[libra-vm] Add simple debugging interface to the VM

tzakian:
This PR adds a basic debugging interface to the VM. It was designed to be minimally invasive, and hook in where we already hook in for coverage gathering. Any transaction can be debugged using this, and it is turned on by setting the `MOVE_VM_STEP` env variable. In release builds this code is not included. 

The debugging interface is interactive and provides a basic set of commands. These are:
* `stack` which prints the current call stack and the contents of the frames on the call stack (see below for an example)
* `step` which performs one step of computation (e.g., advances the PC by 1)
* `continue` which runs the computation either until a breakpoint is hit, or execution terminates
* `breakpoint <instruction>` sets a breakpoint for a prefix of an instruction e.g., `Abort`, or `LdConst`. Can also be set on a fully-qualified function name (e.g. `0x00000000000000000000000000000001::LibraAccount::deposit`).
* `delete <instruction>` removes the `<instruction>` breakpoint
* `breakpoints` prints the current set of breakpoints that are set

## Example 1

Bob just tried to send Alice some money, but the transaction aborted. Bob wants to inspect the state of the transaction at the point the abort happened. Bob can use the debugger to see this:
```
$ MOVE_VM_STEP=1 cargo run -- -u http://localhost:55400 replay-transaction-by-sequence-number 0xB0B 0
Connection Succeeded
Starting epoch execution at 941, 1 transactions remaining
function >> 0x00000000000000000000000000000001::LibraAccount::prologue
instruction >> MoveLoc(0)
program counter >> 0
> breakpoint Abort
> continue
Breakpoint Abort hit with instruction Abort
function >> 0x00000000000000000000000000000001::LibraAccount::deposit
instruction >> Abort
program counter >> 16
> stack
Call Stack:
    [0] main<00000000::LBR::LBR>

        Code:
            [6] MoveLoc(3)
            [7] MoveLoc(4)
            [8] CallGeneric(0)
          > [9] MoveLoc(5)
            [10] Call(2)
            [11] Ret

        Locals:
            [0] -
            [1] 1cbbf5bf2bf60e0e12896e87030fa9cd
            [2] 0
            [3] -
            [4] -
            [5] { 6a8b8c5a7e7b5a1284b2b129b05a9c2c }


    [1] 00000000000000000000000000000001::LibraAccount::pay_from<00000000::LBR::LBR>

        Code:
            [9] MoveLoc(3)
            [10] MoveLoc(4)
            [11] CallGeneric(9)
          > [12] Ret

        Locals:
            [0] -
            [1] 1cbbf5bf2bf60e0e12896e87030fa9cd
            [2] 0
            [3] -
            [4] -


Operand Stack:
    [0] 519

Current frame: 0x00000000000000000000000000000001::LibraAccount::deposit

        Code:
            [0] Call(11)
            [1] CopyLoc(1)
            [2] Call(26)
            [3] ImmBorrowLoc(2)
            [4] CallGeneric(7)
            [5] StLoc(5)
            [6] CopyLoc(5)
            [7] LdU64(0)
            [8] Gt
            [9] LdConst(4)
            [10] Call(3)
            [11] StLoc(7)
            [12] StLoc(6)
            [13] MoveLoc(6)
            [14] BrTrue(17)
            [15] MoveLoc(7)
          > [16] Abort
            [17] CopyLoc(1)
            [18] Call(86)
            [19] LdConst(11)
            [20] Call(6)
            [21] StLoc(9)
            [22] StLoc(8)
            [23] MoveLoc(8)
            [24] BrTrue(27)
            [25] MoveLoc(9)
            [26] Abort
            [27] CopyLoc(1)
            [28] ExistsGeneric(StructDefInstantiationIndex(0))
            [29] LdConst(10)
            [30] Call(3)
            [31] StLoc(11)
            [32] StLoc(10)
            [33] MoveLoc(10)
            [34] BrTrue(37)
            [35] MoveLoc(11)
            [36] Abort
            [37] CopyLoc(0)
            [38] CopyLoc(1)
            [39] CopyLoc(5)
            [40] CopyLoc(3)
            [41] MoveLoc(4)
            [42] CallGeneric(14)
            [43] CopyLoc(0)
            [44] CopyLoc(1)
            [45] LdFalse
            [46] CallGeneric(15)
            [47] BrTrue(49)
            [48] Branch(64)
            [49] CopyLoc(5)
            [50] CopyLoc(1)
            [51] Call(47)
            [52] Call(7)
            [53] ImmBorrowGlobal(StructDefinitionIndex(0))
            [54] ImmBorrowField(FieldHandleIndex(5))
            [55] CallGeneric(16)
            [56] LdConst(5)
            [57] Call(5)
            [58] StLoc(13)
            [59] StLoc(12)
            [60] MoveLoc(12)
            [61] BrTrue(64)
            [62] MoveLoc(13)
            [63] Abort
            [64] CopyLoc(1)
            [65] MutBorrowGlobalGeneric(StructDefInstantiationIndex(0))
            [66] MutBorrowFieldGeneric(FieldInstantiationIndex(0))
            [67] MoveLoc(2)
            [68] CallGeneric(17)
            [69] CopyLoc(1)
            [70] MutBorrowGlobal(StructDefinitionIndex(3))
            [71] MutBorrowField(FieldHandleIndex(6))
            [72] CopyLoc(5)
            [73] CallGeneric(18)
            [74] CopyLoc(0)
            [75] MoveLoc(3)
            [76] Pack(4)
            [77] CallGeneric(19)
            [78] Ret
        Locals:
            [0] 6a8b8c5a7e7b5a1284b2b129b05a9c2c
            [1] 1cbbf5bf2bf60e0e12896e87030fa9cd
            [2] { 0 }
            [3] []
            [4] []
            [5] 0
            [6] -
            [7] -
            [8] -
            [9] -
            [10] -
            [11] -
            [12] -
            [13] -

>
```

## Example 2

Lets say you want to inspect the state of things right before this `abort` is hit in the following script, and then take a look at what happens after it:

```rust
script {
fun main() {
    let x = 1;
    let y = 2:
    let _ = x + y;
    abort 10;
}
}
```

```
$ ct dummy.move
function >> 0x00000000000000000000000000000001::Genesis::initialize
instruction >> LdConst(0)
program counter >> 0
> breakpoint Abort
> continue
Breakpoint Abort hit with instruction Abort
function >> Script::main
instruction >> Abort
program counter >> 3
> stack
Call Stack:
Operand Stack:
    [0] 10

Current frame: Script::main

        Code:
            [0] LdU64(3)
            [1] Pop
            [2] LdU64(10)
          > [3] Abort
        Locals:
            [0] -
            [1] -

> breakpoints
[0] Abort
> step
function >> 0x00000000000000000000000000000001::LibraAccount::prologue
instruction >> MoveLoc(0)
program counter >> 0
> stack
Call Stack:
Operand Stack:

Current frame: 0x00000000000000000000000000000001::LibraAccount::prologue

        Code:
          > [0] MoveLoc(0)
            [1] Call(0)
            [2] StLoc(26)
            [3] Call(25)
.......
> delete Abort
> breakpoint Call
> continue
Breakpoint Call hit with instruction Call(0)
function >> 0x00000000000000000000000000000001::LibraAccount::prologue
instruction >> Call(0)
program counter >> 1
> stack
Call Stack:
Operand Stack:
    [0] (&) { a4a46d1b1421502568a4a6ac326d7250 }

Current frame: 0x00000000000000000000000000000001::LibraAccount::prologue

        Code:
            [0] MoveLoc(0)
          > [1] Call(0)
            [2] StLoc(26)
```

### DIFF
--- a/language/move-vm/runtime/src/debug.rs
+++ b/language/move-vm/runtime/src/debug.rs
@@ -1,0 +1,194 @@
+// Copyright (c) The Libra Core Contributors
+// SPDX-License-Identifier: Apache-2.0
+
+use crate::{
+    interpreter::Interpreter,
+    loader::{Function, Loader},
+};
+use move_vm_types::values::{self, Locals};
+use std::{
+    collections::BTreeSet,
+    io::{self, Write},
+    str::FromStr,
+};
+use vm::file_format::Bytecode;
+
+#[derive(Debug)]
+enum DebugCommand {
+    PrintStack,
+    Step,
+    Continue,
+    Breakpoint(String),
+    DeleteBreakpoint(String),
+    PrintBreakpoints,
+}
+
+impl DebugCommand {
+    pub fn debug_string(&self) -> &str {
+        match self {
+            Self::PrintStack => "stack",
+            Self::Step => "step",
+            Self::Continue => "continue",
+            Self::Breakpoint(_) => "breakpoint ",
+            Self::DeleteBreakpoint(_) => "delete ",
+            Self::PrintBreakpoints => "breakpoints",
+        }
+    }
+
+    pub fn commands() -> Vec<DebugCommand> {
+        vec![
+            Self::PrintStack,
+            Self::Step,
+            Self::Continue,
+            Self::Breakpoint("".to_string()),
+            Self::DeleteBreakpoint("".to_string()),
+            Self::PrintBreakpoints,
+        ]
+    }
+}
+
+impl FromStr for DebugCommand {
+    type Err = String;
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        use DebugCommand::*;
+        let s = s.trim();
+        if s.starts_with(PrintStack.debug_string()) {
+            return Ok(PrintStack);
+        }
+        if s.starts_with(Step.debug_string()) {
+            return Ok(Step);
+        }
+        if s.starts_with(Continue.debug_string()) {
+            return Ok(Continue);
+        }
+        if let Some(breakpoint) = s.strip_prefix(Breakpoint("".to_owned()).debug_string()) {
+            return Ok(Breakpoint(breakpoint.to_owned()));
+        }
+        if let Some(breakpoint) = s.strip_prefix(DeleteBreakpoint("".to_owned()).debug_string()) {
+            return Ok(DeleteBreakpoint(breakpoint.to_owned()));
+        }
+        if s.starts_with(PrintBreakpoints.debug_string()) {
+            return Ok(PrintBreakpoints);
+        }
+        Err(format!(
+            "Unrecognized command: {}\nAvailable commands: {}",
+            s,
+            Self::commands()
+                .iter()
+                .map(|command| command.debug_string())
+                .collect::<Vec<_>>()
+                .join(", ")
+        ))
+    }
+}
+
+#[derive(Debug)]
+pub(crate) struct DebugContext {
+    breakpoints: BTreeSet<String>,
+    should_take_input: bool,
+}
+
+impl DebugContext {
+    pub(crate) fn new() -> Self {
+        Self {
+            breakpoints: BTreeSet::new(),
+            should_take_input: true,
+        }
+    }
+
+    pub(crate) fn debug_loop(
+        &mut self,
+        function_desc: &Function,
+        locals: &Locals,
+        pc: u16,
+        instr: &Bytecode,
+        resolver: &Loader,
+        interp: &Interpreter,
+    ) {
+        let instr_string = format!("{:?}", instr);
+        let function_string = function_desc.pretty_string();
+        let breakpoint_hit = self.breakpoints.contains(&function_string)
+            || self
+                .breakpoints
+                .iter()
+                .any(|bp| instr_string[..].starts_with(bp.as_str()));
+
+        if self.should_take_input || breakpoint_hit {
+            self.should_take_input = true;
+            if breakpoint_hit {
+                let bp_match = self
+                    .breakpoints
+                    .iter()
+                    .find(|bp| instr_string.starts_with(bp.as_str()))
+                    .unwrap()
+                    .clone();
+                println!(
+                    "Breakpoint {} hit with instruction {}",
+                    bp_match, instr_string
+                );
+            }
+            println!(
+                "function >> {}\ninstruction >> {:?}\nprogram counter >> {}",
+                function_string, instr, pc
+            );
+            loop {
+                print!("> ");
+                std::io::stdout().flush().unwrap();
+                let mut input = String::new();
+                match io::stdin().read_line(&mut input) {
+                    Ok(_) => match input.parse::<DebugCommand>() {
+                        Err(err) => println!("{}", err),
+                        Ok(command) => match command {
+                            DebugCommand::Step => {
+                                self.should_take_input = true;
+                                break;
+                            }
+                            DebugCommand::Continue => {
+                                self.should_take_input = false;
+                                break;
+                            }
+                            DebugCommand::Breakpoint(breakpoint) => {
+                                self.breakpoints.insert(breakpoint.to_string());
+                            }
+                            DebugCommand::DeleteBreakpoint(breakpoint) => {
+                                self.breakpoints.remove(&breakpoint);
+                            }
+                            DebugCommand::PrintBreakpoints => self
+                                .breakpoints
+                                .iter()
+                                .enumerate()
+                                .for_each(|(i, bp)| println!("[{}] {}", i, bp)),
+                            DebugCommand::PrintStack => {
+                                let mut s = String::new();
+                                interp.debug_print_stack_trace(&mut s, resolver).unwrap();
+                                println!("{}", s);
+                                println!("Current frame: {}\n", function_string);
+                                let code = function_desc.code();
+                                println!("        Code:");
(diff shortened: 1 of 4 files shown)


runtian-zhou:
Do we need to have debug_assertions feature to be on for `libra-transaction-replay`?

runtian-zhou:
Also what's the difference between `MOVE_VM_STEP ` and `MOVE_VM_TRACE `?

language/move-vm/runtime/src/tracing.rs
```diff
@@ -1,6 +1,14 @@
 // Copyright (c) The Libra Core Contributors
 // SPDX-License-Identifier: Apache-2.0
 
+#[cfg(debug_assertions)]
```
runtian-zhou, line 4: Should we make `debug_assertions` a feature for `move-vm-runtime` and `libra-vm` and make sure it is off by default so that they won't be activated by default?

tzakian:
> Also what's the difference between MOVE_VM_STEP and MOVE_VM_TRACE ?

Good question! 
* `MOVE_VM_TRACE` gathers traces through the VM without halting execution, and is used to gather coverage information for Move.
* `MOVE_VM_STEP` halts execution and drops you in to this debug interface to inspect the stack etc. It does not gather execution traces/coverage-related information.

language/move-vm/runtime/src/tracing.rs
tzakian: Not sure if I quite follow so I'd be interested in hearing more about your thoughts here. 

My thinking at least was that `debug_assertions` are automatically turned off in release builds, and these being around for debug builds are fine since it might just slow things down a very little bit (one extra Rust function call per instruction).

language/move-vm/runtime/src/debug.rs
```diff
@@ -0,0 +1,194 @@
...
+                                self.breakpoints.remove(&breakpoint);
+                            }
+                            DebugCommand::PrintBreakpoints => self
+                                .breakpoints
+                                .iter()
+                                .enumerate()
+                                .for_each(|(i, bp)| println!("[{}] {}", i, bp)),
+                            DebugCommand::PrintStack => {
+                                let mut s = String::new();
+                                let _ = interp.debug_print_stack_trace(&mut s, resolver);
```
vgao1996, line 163: So here we're calling a function with side effects and silently discarding its result. A little bit disturbed by this but I understand you can't simply propagate the error up. Would it be better we `unwrap` here? I assume it's fine for it to crash in case something goes wrong in `debug_print_stack_trace` as it only happens in a non-prod build.

language/move-vm/runtime/src/debug.rs
```diff
@@ -0,0 +1,194 @@
...
+            }
+            println!(
+                "function >> {}\ninstruction >> {:?}\nprogram counter >> {}",
+                function_string, instr, pc
+            );
+            loop {
+                print!("> ");
+                std::io::stdout().flush().unwrap();
+                let mut input = String::new();
+                match io::stdin().read_line(&mut input) {
```
vgao1996, line 138: Interesting approach! So we're effectively instrumenting each instruction with a call into the debugging loop, which will then block on stdin for user input. I think this would be a great and very usable at least for now. Wondering if we would prefer offering a set of debugging/stepping interfaces from the VM instead in the future.

@dariorussi I'd love to hear your opinions on this as a short-term solution.

language/move-vm/runtime/src/debug.rs
tzakian: Good spot, I'll add that `unwrap` in!

language/move-vm/runtime/src/tracing.rs
runtian-zhou: That is indeed what's happening right now: release build of VM binary will not include those features whereas debug build will contain those tracing and debugging logic. That's the reason why `transaction-replay` get this feature without any changes made over there, which is really cool. What I'm suggestingis that instead of using `debug_assertions` flag, we can make those become a feature of `move_vm_runtime` that is switched automatically of in the release build, and we can choose if we need to enable it in the debug/test phase.

tzakian: I guess my main question here that I'm still wondering about is what adding the additional flag buys us. Having this switched on in debug builds should have essentially no impact, but, I worry that adding the additional flag adds additional complexity across the codebase since other crates that pull in the VM will needing to include this feature flag  to then pass down to the VM etc.  What are your thoughts on this @dariorussi and @vgao1996?

dariorussi: this does not seem a critical aspect for now, or at least not in my opinion.
The different flag could allow us to build a release with VM debug capability if needed, so that would be the win for me. Not that we could not do it otherwise but creates a clear "feature" to depend on. 
However I am not clear what we need now, so this should be ok.
Thoughts?

language/move-vm/runtime/src/debug.rs
dariorussi: this seems fine to me and it's a way to have some minimal debugging ability so it's all good.... :)

tzakian:
It sounds like this is good to land, so I'll do that now. 

I think moving both `tracing` and `debug` out in to their own crate would be nice, but I think there's some work there to avoid circular includes.

If we want to circle back on the feature flags, and/or move `tracing` and `debug` to a separate module happy to do that as well in a followup PR :) 

/land

PR #5622, opened 2020-08-14
[network] [breaking] Remove ping / pong

davidiw:
## Breaking Nature
The breaking change is limited to LibraNet communication:
* NetworkMessage::ErrorCode definitions changed
* NetworkMessage::RpcRequest format changed
* NetworkMessage::Ping/Pong removed

## Justification
* Error definitions were ineffective and practically unusable -- such as reporting the network version in a pre-negotiated channel, returning full length messages as errors, and being incomplete on which message type was being used
* Ping/Pong was unused and duplicate of a high level protocol -- there was no value in keeping it around if we never used it properly
* Ordering of the fields within RpcRequest required additional parsing to retrieve the protocol which is used for error handling

## Details
This PR does a few things:
* Improves the error messages to be more useful:
  * ParsingError: the messaging version isn't relevant since this was already exchanged as part of the handshake, sending the entire message back is bad due to DOS attacks. This returns the first two bytes which typically are the message type and protocol type.
  * NotSupported: returning back the protocol type isn't that great since it doesn't help us route it to the correct handler, we actually need to first send it to the message type handler, hence we made this a bit more robust
* Removes ping / pong as we never got around to migrating the healthcheck service to use this, it was an unimplemented component of the API
* Adds an error response on unparseable messages -- which was primarily done to help an integration test
* Reorder fields within the RpcRequest so that the first 2 bytes define message type and then protocol type.

It is breaks the LibraNet protocol.

### DIFF
--- a/network/src/peer/mod.rs
+++ b/network/src/peer/mod.rs
@@ -6,7 +6,7 @@
 use crate::{
     counters,
     peer_manager::PeerManagerError,
-    protocols::wire::messaging::v1::NetworkMessage,
+    protocols::wire::messaging::v1::{ErrorCode, NetworkMessage},
     transport,
     transport::{Connection, ConnectionMetadata},
     ProtocolId,
@@ -321,7 +321,19 @@ where
         trace!("Received message from Peer {}", self.peer_id().short_str());
         // Read inbound message from stream.
         let message = message.freeze();
-        let message: NetworkMessage = lcs::from_bytes(&message)?;
+        let message = match lcs::from_bytes(&message) {
+            Ok(message) => message,
+            Err(err) => {
+                // Don't bother returning errors for tiny messages
+                if message.len() >= 2 {
+                    let error = ErrorCode::parsing_error(message[0], message[1]);
+                    let message = NetworkMessage::Error(error);
+                    let (ack_tx, _) = oneshot::channel();
+                    write_reqs_tx.send((message, ack_tx)).await?;
+                }
+                return Err(err.into());
+            }
+        };
         match message {
             NetworkMessage::DirectSendMsg(_) => {
                 let notif = PeerNotification::NewMessage(message);
--- a/network/src/peer/mod.rs
+++ b/network/src/peer/mod.rs
@@ -355,15 +355,6 @@ where
                     error,
                 );
             }
-            NetworkMessage::Ping(nonce) => {
-                let pong = NetworkMessage::Pong(nonce);
-                let (ack_tx, _) = oneshot::channel();
-                // Resond to a ping right away.
-                write_reqs_tx.send((pong, ack_tx)).await?;
-            }
-            NetworkMessage::Pong(_nonce) => {
-                // TODO(davidiw) forward to a pong pub/sub and move HealthChecker to leverage this
-            }
             NetworkMessage::RpcRequest(_) | NetworkMessage::RpcResponse(_) => {
                 let notif = PeerNotification::NewMessage(message);
                 self.rpc_notifs_tx.send(notif).await.map_err(|err| {
--- a/network/src/protocols/wire/messaging/v1/mod.rs
+++ b/network/src/protocols/wire/messaging/v1/mod.rs
@@ -2,16 +2,15 @@
 // SPDX-License-Identifier: Apache-2.0
 
 //! This module defines the structs transported during the network messaging protocol v1.
-//! These should serialize as per [link](TODO: Add ref).
+//! These should serialize as per the [specification](https://github.com/libra/libra/blob/master/specifications/network/messaging-v1.md)
 
-use crate::protocols::wire::handshake::v1::{MessagingProtocolVersion, ProtocolId};
+use crate::protocols::wire::handshake::v1::ProtocolId;
 use serde::{Deserialize, Serialize};
 
 #[cfg(test)]
 mod test;
 
-/// Message variants that are sent on the wire.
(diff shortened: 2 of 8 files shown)


network/src/peer/mod.rs
```diff
@@ -321,7 +321,19 @@ where
         trace!("Received message from Peer {}", self.peer_id().short_str());
         // Read inbound message from stream.
         let message = message.freeze();
-        let message: NetworkMessage = lcs::from_bytes(&message)?;
+        let message = match lcs::from_bytes(&message) {
+            Ok(message) => message,
+            Err(err) => {
+                // Don't bother returning errors for tiny messages
+                if message.len() >= 2 {
```
zekun000, line 328: is this 2 bytes?

network/src/protocols/wire/messaging/v1/mod.rs
```diff
@@ -46,10 +58,10 @@ pub type Priority = u8;
 
 #[derive(Clone, Debug, PartialEq, Deserialize, Serialize)]
 pub struct RpcRequest {
-    /// RequestId for the RPC Request.
-    pub request_id: RequestId,
     /// `protocol_id` is a variant of the ProtocolId enum.
     pub protocol_id: ProtocolId,
+    /// RequestId for the RPC Request.
+    pub request_id: RequestId,
```
zekun000, line 64: want to piggy back this breaking change too? lol

network/src/protocols/wire/messaging/v1/mod.rs
```diff
@@ -2,41 +2,53 @@
...
-    ParsingError(MessagingProtocolVersion, Box<NetworkMessage>),
+    ParsingError(ParsingErrorType),
     /// A message was received for a protocol that is not supported over this connection.
-    NotSupported(ProtocolId),
+    NotSupported(NotSupportedType),
 }
 
-/// Nonces used by Ping and Pong message types.
+impl ErrorCode {
+    pub fn parsing_error(message: u8, protocol: u8) -> Self {
```
zekun000, line 31: use ProtocolId instead of u8?

network/src/protocols/wire/messaging/v1/mod.rs
davidiw: woops, meant to call this out :D

network/src/peer/mod.rs
davidiw: Yeah, if you notice, this is some of my lowest quality code... I only want to return meaningful error messages :P

network/src/protocols/wire/messaging/v1/mod.rs
davidiw: for the parsing error, the message and protocol type may not be within our enum... consider this the "safe" version of the stricter unsupported protocol.

aching:
Curious on @phlip9 's thoughts as he worked on the health checking and ping/pong ideas.

network/src/protocols/wire/messaging/v1/mod.rs
aching: Is the reason for this that you can send out the ProtocolId related from reading the first bytes off the wire as we expect them to be more common than request_id related errors?

network/src/peer/mod.rs
aching: What is the specification for returning errors?  Optional but recommended?  Only if the message is bigger than 2 bytes?  I took a quick look at the network specs but couldn't find it.

network/src/protocols/wire/messaging/v1/mod.rs
aching: You answered this proactively in the updated summary.  Thanks!

network/src/peer/mod.rs
davidiw: Great question, considering the amount of effort it would take to make it mandatory, I think we can say say it is optional assuming that it is parseable. For non-parseable errors, we only return errors for those of 2 bytes or more otherwise they aren't well-formed enough to provide any meaningful response.

davidiw:
Updated the general verbiage on the specification on what must be handled and what it optional. I think we'll still likely need an update on the spec on language, but at least the implementation / code do not diverge.

specifications/network/messaging-v1.md
```diff
@@ -112,7 +100,9 @@ Pending inbound and outbound messages MAY be reordered and dropped according to
 
 ## Errors
 
-Errors are sent as messages of type `NetworkMessage::Error`, with the `ErrorCode` indicating the type of error. For example, if an `RpcRequest` is received for a `ProtocolId` that was not advertised to a node, we send an error message with the code `ErrorCode::NotSupported(ProtocolId)`.
+Errors are sent as messages of type `NetworkMessage::Error`, with the `ErrorCode` indicating the type of error. For example, if an `RpcRequest` is received for a `ProtocolId` that was not advertised to a node, we send an error message with the code `ErrorCode::NotSupported(1, ProtocolId)`, where 1 represents the index for RpcRequest's in NetworkMessage.
+
+Responding to errors is not requried. A message must be of at least length 2 in order to trigger an error response, otherwise an error would have insufficient data to be meaningful.
```
aching, line 105:
```suggestion
Responding to errors is not requred. A message must be of at least length 2 in order to trigger an error response, otherwise an error would have insufficient data to be meaningful.
```

davidiw: :smh: need to use ispell....

Issue #5612, opened 2020-08-13
[logging] Update security logs with leveled logs


gregnazario:
### Overview
Let's migrate send_struct_log! to use leveled macros, and remove `crit!` logs

### Tasks
- [ ] Remove any reference to `.warning()` or `.critical()`
- [ ] Change `send_struct_log!` to `sl_info!`, `sl_error!`, etc... accordingly similar to the text log macros
- [ ] replace any `crit!` with `error!`

gregnazario:
@mimoo  :) if you could take some time to appropriately fix these logs, that would be great!

I'm tracking all of these in the logging project, to ensure I know how things are working out.

mimoo:
wait, no more send_struct_log!? what's the rationale behind that : o ? Also why no more critical?

gregnazario:
In some macro magic, in order to decrease the amount of logs being logged into the structured logs, moving to a simple macro system that people are familiar with `warn!` or `info!` to define the levels at the time it's written rather than having an optional level system.  I also considered just putting it in the constructor, but I decided that this was simpler from a usability standpoint.  My goal here is to make it such that we can move all text logs to struct logs at some point.  The complete picture of that isn't quite complete.

Critical had a bunch of TODOs around it because they weren't supported by the underlying text log system.  We had moved from a system that had `critical` to one that didn't, so I consolidated it to be the same.  If there's some particular reason we need `critical` over `error` plus filters then let me know.  But, how I see it we should be able to make an alarm on `security` and `error` for your use case.

mimoo:
https://github.com/libra/libra/pull/5625

PR #5625, opened 2020-08-14
[logging][easy] move security logs to new structure logging macro

mimoo:
closes https://github.com/libra/libra/issues/5612

### DIFF
--- a/consensus/src/epoch_manager.rs
+++ b/consensus/src/epoch_manager.rs
@@ -396,14 +396,13 @@ impl EpochManager {
                 .clone()
                 .verify(&self.epoch_state().verifier)
                 .context("[EpochManager] Verify event")
-                .map_err(|err|
-                        // security log + return the error
-                        {send_struct_log!(security_log(security_events::CONSENSUS_INVALID_MESSAGE)
-                            .data("from_peer", &peer_id)
-                            .data_display("error", &err)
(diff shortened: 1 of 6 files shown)


consensus/src/epoch_manager.rs
```diff
@@ -398,7 +398,7 @@ impl EpochManager {
                 .context("[EpochManager] Verify event")
                 .map_err(|err|
                         // security log + return the error
-                        {send_struct_log!(security_log(security_events::CONSENSUS_INVALID_MESSAGE)
+                        {sl_error!(security_log(security_events::CONSENSUS_INVALID_MESSAGE)
                             .data("from_peer", &peer_id)
                             .data_display("error", &err)
                             .data("event", &unverified_event)
```
gregnazario, line 404: Tbh, this formatting seems weird, but it was already there.  Was this the autoformatter?

mimoo: I made it nicer, that's one problem with rustfmt: it seems to allow different formatting to be valid : ( (unlike gofmt)

gregnazario: Rust format doesn't seem to format nicely under some very specific nesting conditions.  I notice it doesn't like macros and we sometimes use macros for large blocks

PR #5627, opened 2020-08-14
[logging] Allow text logs to be replaced with struct logs

gregnazario:
### Overview
This approach opens the `log` field previously only used by text logs
that were converted to struct logs, by struct logs.  In that way, we can
use a category and name rather than a pattern to find log lines
associated in ELK.

The main purpose of this is to have text logs have explicit indexing
guidelines, on how and what we want to index.  The automatic conversion
of text logs was easy, but caused a lot of confusion for anonymous field
names e.g. `_0` and `_1`, and it forced one type of output `debug`.  This
also kills some unused fields, and cleans up the interface to make a
clear difference between mutref and non-mutref functions.

Ideally, this will allow us to have one logging system, using the structured log API interface.

#### Why is there a network log change in here?
* I used it as an example

#### But, what about the current text logs?
1. We can keep the current text logs if we desire, but I want to remove the automatic indexing that causes data inconsistencies and confusion.  Ideally, we could move these to structured logs.
2. The plan is to convince the stakeholders that we need to write the Structured logs to disk before we write to Logstash.  Some reasons:
* Logstash can handle resubmission of logs, backoff, timeouts, and all the things I couldn't properly handle with the direct TCP submission, without a lot of work
* We get the added benefit of having disk logs for whatever else, archiving, etc.  Even if they're thrown away immediately, it means we can tail the logs on a running host
* Logging to disk is familiar and is most commonly used

### Followup actions
* The docs are a little out of date for structured logging, I want to come back and fix that in a separate commit.
* Cut issues to owners to migrate their logs to structured logs
* Work with the PEs to see if we can connect a shared storage between pods to write the logs to Elasticsearch via Logstash.

### DIFF
--- a/common/logger/src/struct_log.rs
+++ b/common/logger/src/struct_log.rs
@@ -71,9 +71,6 @@ pub struct StructuredLogEntry {
     /// filename + line (e.g. consensus/src/round_manager.rs:678)
     #[serde(skip_serializing_if = "Option::is_none")]
     location: Option<&'static str>,
-    /// git revision
-    #[serde(skip_serializing_if = "Option::is_none")]
-    git_rev: Option<&'static str>,
     /// time of the log
     #[serde(skip_serializing_if = "Option::is_none")]
     timestamp: Option<String>,
@@ -117,7 +114,6 @@ impl StructuredLogEntry {
             name: self.name,
             module: self.module,
             location: self.location,
-            git_rev: self.git_rev,
             timestamp: self.timestamp.clone(),
             level: self.level,
             data: HashMap::new(),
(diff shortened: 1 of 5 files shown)


common/logger/src/struct_log.rs
```diff
@@ -176,65 +172,61 @@ impl StructuredLogEntry {
...
+    /// Sets the context log line used for text logs.  This is useful for migration of text logs
+    pub fn log(mut self, log: String) -> Self {
         self.log = Some(log);
         self
     }
 
     #[doc(hidden)] // set from macro
-    pub fn pattern(&mut self, pattern: &'static str) -> &mut Self {
-        self.pattern = Some(pattern);
+    pub fn add_log(&mut self, log: String) -> &mut Self {
```
davidiw, line 205: add_* seems like an anti-pattern much like set_ or get_

gregnazario: Yeah, it's an anti-pattern, but the issue was the confusion between the `&mut Self` vs `Self` returns.  Rather than rework the entire macro system, I went with this to differentiate between the two.

gregnazario:
> beside the add_ and the removal of the GIT_REV it looks good. Why did we get rid of GIT_REV?

`GIT_REV` isn't used, because we never set the environment variable.  It's added via logstash right now.

PR #5631, opened 2020-08-14
[language] reorganize testing infrastructures

vgao1996:
## Summary
This creates a new directory `language/testing` and a moves a few crates into it:
- `language/functional-tests` -> `language/testing/functional-tests`
- `language/tools/test-generation` -> `language/testing/test-generation`
- `language/tools/utils` -> `language/testing/module-generation` (with minor changes to its internal module/directory structure)
- `language/e2e-tests` -> `language/testing/e2e-tests` & `language/e2e-testsuite`
  - `e2e-tests` contains all the library code and `e2e-testsuite` contains the actual tests.

The goal is to create a home for testing related libraries instead of having them scattered throughout language.

## Test Plan
cargo test


sblackshear:
One separation I would like to see as we reorg the tests is VM/bytecode verifier/source language tests vs Libra Framework tests. I think ideally, these tests should live in different directories and the VM etc tests should not depend on Libra Framework modules.

vgao1996:
@sblackshear That will happen, at a later time though. My plan consists of building one or more pipelines that directly invokes the Move VM instead of the Libra Adapter.

tzakian:
Overall I'm very very _very_ happy to see this move. The only question I have is the naming of the new directory. Perhaps something like `testing-infra` or `testing-infrastructure`, or a different name that makes it unambiguous that code under it should only hold testing libraries/infra and not tests themselves.

vgao1996:
@tzakian will rename to "testing-infra".

PR #5632, opened 2020-08-14
[metrics] update push metrics to leverage prometheus counters

davidiw:
These already found their way into the logger. We don't have bandwidth
to update all the code to provide the labeled counters for TCBs and also
fix the use of logging counters, so might as well adapt.


davidiw:
Verified in cluster test!

PR #5639, opened 2020-08-17
[Fuzzing] Fix a bug in the secure JSON RPC client: SecureJsonRpcGetAccountTransaction fuzz target

JoshLind:
## Motivation

Fuzz targets for the secure JSON RPC client were recently added (https://github.com/libra/libra/pull/5550), which included 3 different fuzz targets. However, looking at the fuzzing code coverage, I can see that one of the targets is not being covered at all. This PR fixes the bug, by calling the correct method in the fuzz target.

Yes.

## Test Plan

All tests still pass locally. Moreover, once this lands, I'll be to able to verify the fuzzing execution path by checking the code coverage.

## Related PRs

None.


JoshLind:
Thanks, David! 😄

PR #5640, opened 2020-08-17
[fuzzing] Improving structure fuzzing corpus generation

mimoo:
This uses the changes we've made to proptest to more accurately generate corpus when doing structured fuzzing.

The gist is that we can now create a random value with proptest, and THEN dump the bytes that were read from the RNG and used by proptest to create that value.
These bytes can then be fed back to proptest when we fuzz to re-create the same value. This is more accurate than what we do currently: we just generate random bytes and hope that this creates a valid value.

The PR also adds two helpers that one can use to quickly implement structured fuzzing:

* `corpus_from_strategy(strategy) -> Vec<u8>` creates a random value using a strategy and dumps the bytes used to create the value, useful for corpus generation.
* `fuzz_data_to_value(data, strategy) -> T` creates a type from the fuzzer input (data) and some strategy

### DIFF
--- a/testsuite/libra-fuzzer/src/fuzz_targets/transaction.rs
+++ b/testsuite/libra-fuzzer/src/fuzz_targets/transaction.rs
@@ -1,13 +1,14 @@
 // Copyright (c) The Libra Core Contributors
 // SPDX-License-Identifier: Apache-2.0
 
-use crate::FuzzTargetImpl;
+use crate::{corpus_from_strategy, fuzz_data_to_value, FuzzTargetImpl};
 use language_e2e_tests::account_universe::{
-    all_transactions_strategy, log_balance_strategy, run_and_assert_universe, AccountUniverseGen,
+    all_transactions_strategy, log_balance_strategy, run_and_assert_universe, AUTransactionGen,
+    AccountUniverseGen,
 };
 use libra_proptest_helpers::ValueGenerator;
 use libra_types::transaction::SignedTransaction;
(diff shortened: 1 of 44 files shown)


mimoo:
@runtian-zhou can you make sure my refactor is correct o.o ?

mimoo:
@christinacelee FYI this is what it looks like to generate now : o

testsuite/libra-fuzzer/src/fuzz_targets/transaction.rs
```diff
@@ -17,24 +18,25 @@ impl FuzzTargetImpl for LanguageTransactionExecution {
...
 
+    fn fuzz(&self, data: &[u8]) {
+        let (universe, txns) = fuzz_data_to_value(data, run_and_assert_universe_input());
         run_and_assert_universe(universe, txns).unwrap();
     }
 }
 
+prop_compose! {
+    fn run_and_assert_universe_input()(
+        universe in AccountUniverseGen::strategy(2..20, log_balance_strategy(10_000_000)),
```
runtian-zhou, line 33: Seems right to me!

sausagee:
> The gist is that we can now create a random value with proptest, and THEN dump the bytes that were read from the RNG and used by proptest to create that value.
These bytes can then be fed back to proptest when we fuzz to re-create the same value. This is more accurate than what we do currently: we just generate random bytes and hope that this creates a valid value.

@mimoo this sounds like a good note to put in the README.md

PR #5643, opened 2020-08-17
[x] re-introduce check for overlay features, update for v2 resolver

sunshowers:
The Cargo v2 resolver can avoid unifying features for dev dependencies, so the
check can be relaxed a little.

This reverts commit 9a484bde and makes a minor change to it: the callback for the `resolve_with_fn` method is now conditional on `!link.dev_only()`.

### DIFF
--- a/devtools/x/src/lint/guppy.rs
+++ b/devtools/x/src/lint/guppy.rs
@@ -3,11 +3,12 @@
 
 //! Project and package linters that run queries on guppy.
 
-use crate::config::{BannedDepsConfig, EnforcedAttributesConfig};
-use guppy::Version;
+use crate::config::{BannedDepsConfig, EnforcedAttributesConfig, OverlayConfig};
+use guppy::{graph::feature::FeatureFilterFn, Version};
 use std::{
     collections::{BTreeMap, HashMap},
     ffi::OsStr,
+    iter,
 };
 use x_core::WorkspaceStatus;
 use x_lint::prelude::*;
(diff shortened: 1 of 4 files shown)


metajack:
Can you summarize what changed other than the revert? It's not clear from the diff which parts are new.

sunshowers:
> Can you summarize what changed other than the revert? It's not clear from the diff which parts are new.

Ah sorry, just updated the summary. The line is at `// We now use the v2 resolver, so dev-only links can be skipped.`

devtools/x/src/lint/guppy.rs
```diff
@@ -308,3 +309,108 @@ impl ProjectLinter for DirectDepDups {
...
+            .feature_graph()
+            .query_packages(&package_query, filter);
+
+        let mut overlays: Vec<(Option<&str>, &str, Option<&str>)> = vec![];
+
+        feature_query.resolve_with_fn(|_, link| {
+            // We now use the v2 resolver, so dev-only links can be skipped.
+            if !link.dev_only() {
+                // Consider the dependency even if it's dev-only since the v1 resolver unifies these.
+                // TODO: might be able to relax this for the v2 resolver.
```
metajack, line 377: This should probably get removed now :)

metajack:
Did someone not check in a serde-generate update? Why is that showing up here.

metajack:
Hmm, i wonder if it is because of the rebase. I'll have to check what the detected merge base sha is after bors pushes and see if I'm calculating that wrong.

PR #5648, opened 2020-08-17
[cluster-test] Limit kubernetes commands concurrency

andll:
We have a lot of backoff in cluster test due to having too many kubernetes commands issues concurrently on tear down stage. This diff limits concurrency of those futures to max 16 in parallel


snakech:
is it ```Experiment deadline reached``` because of we put limitation of semaphore which may slower the process?

PR #5651, opened 2020-08-17
[Fuzzing] Update Secure JSON RPC Client fuzzers to use proptest

JoshLind:
## Motivation

This PR updates the existing Secure JSON RPC Client fuzzers to use proptest, thus allowing us to perform structured fuzzing and hopefully achieve better coverage. 

Currently, the fuzzers used by the Secure JSON RPC Client take the approach of generating a ureq::Response template, converting the template into a json string, allowing the fuzzer to modify the json string, and then turning the modified string back into a ureq::Response. This is not ideal, as the modified string produced by the fuzzer may not always convert back into a ureq::Response successfully (causing the fuzzer to prematurely exit). To avoid this, we instead use proptest to generate an arbitrary ureq::Response, with different fields/internals directly. This avoids the need to serialize using json (which may fail), and thus will hopefully help us to achieve better coverage.

Yes.

## Test Plan

All tests still pass locally.

## Related PRs

None.

### DIFF
--- a/secure/json-rpc/Cargo.toml
+++ b/secure/json-rpc/Cargo.toml
@@ -11,6 +11,7 @@ edition = "2018"
 
 [dependencies]
 hex = "0.4.2"
+proptest = { version = "0.10.1", optional = true }
 serde = { version = "1.0.114", features = ["derive"], default-features = false }
 serde_json = "1.0.57"
 thiserror = "1.0.20"
@@ -18,15 +19,18 @@ ureq = { version = "1.3.0", features = ["json"] }
 
 lcs = { path = "../../common/lcs", version = "0.1.0", package = "libra-canonical-serialization" }
 libra-crypto = { path = "../../crypto/crypto", version = "0.1.0" }
+libra-proptest-helpers = { path = "../../common/proptest-helpers", version = "0.1.0", optional = true }
 libra-types = { path = "../../types", version = "0.1.0" }
 libra-workspace-hack = { path = "../../common/workspace-hack", version = "0.1.0" }
 
 [dev-dependencies]
 anyhow = "1.0.32"
 futures = "0.3.5"
+proptest = "0.10.1"
 tokio = { version = "0.2.22", features = ["full"] }
 
 libra-config = { path = "../../config", version = "0.1.0" }
+libra-proptest-helpers = { path = "../../common/proptest-helpers", version = "0.1.0"}
 libra-json-rpc = { path = "../../json-rpc", version = "0.1.0", features = ["fuzzing"] }
 libra-temppath = { path = "../../common/temppath", version = "0.1.0" }
 libra-types = { path = "../../types", version = "0.1.0", features = ["fuzzing"] }
@@ -35,4 +39,4 @@ storage-interface = { path = "../../storage/storage-interface", version = "0.1.0
 vm-validator = { path = "../../vm-validator", version = "0.1.0" }
 
 [features]
-fuzzing = []
+fuzzing = ["proptest", "libra-proptest-helpers"]
--- a/secure/json-rpc/src/lib.rs
+++ b/secure/json-rpc/src/lib.rs
@@ -687,105 +687,165 @@ mod test {
 #[cfg(any(test, feature = "fuzzing"))]
 /// This module provides all the functionality necessary to test the secure JSON RPC client using
 /// fuzzing.
-/// Note: the unit tests are to ensure that the various fuzzers are maintained (i.e., not broken
-/// at some time in the future and only discovered when a fuzz test fails).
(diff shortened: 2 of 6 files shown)


secure/json-rpc/Cargo.toml
```diff
@@ -11,22 +11,26 @@ edition = "2018"
 
 [dependencies]
 hex = "0.4.2"
+proptest = { version = "0.10.0", optional = true }
```
mimoo, line 14: version 0.10.1 !

secure/json-rpc/src/lib.rs
```diff
@@ -687,105 +687,178 @@ mod test {
...
+                jsonrpc,
+                result: Some(transaction_view),
+            };
+            let response_body =
+                serde_json::to_string::<TransactionViewResponse>(&response_body).unwrap();
+            Response::new(status, &status_text, &response_body)
         }
     }
+
+    pub fn arb_account_state_response() -> impl Strategy<Value = Response> {
```
mimoo, line 798: why are you re-wrapping these functions?

secure/json-rpc/src/lib.rs
```diff
@@ -687,105 +687,178 @@ mod test {
...
+    pub fn arb_submit_transaction_response() -> impl Strategy<Value = Response> {
+        arb_submit_transaction_response_impl()
+    }
+
+    pub fn arb_transaction_status_response() -> impl Strategy<Value = Response> {
+        arb_transaction_status_response_impl()
+    }
+
+    // This function generates an arbitrary serde_json::Value.
+    fn arb_json_value() -> impl Strategy<Value = Value> {
```
mimoo, line 811: so if I understand correctly you're trying to respond with garbage but valid json objects to submit?

secure/json-rpc/src/lib.rs
```diff
@@ -687,105 +687,178 @@ mod test {
...
+            ),
+            Just(VMStatusView::VerificationError),
+            Just(VMStatusView::DeserializationError),
+            Just(VMStatusView::PublishingFailure),
+        ]
+    }
 }
 
 #[cfg(any(test, feature = "fuzzing"))]
+#[allow(dead_code)]
```
mimoo, line 861: why :D?

testsuite/libra-fuzzer/src/fuzz_targets/secure_json_rpc_client.rs
```diff
@@ -19,12 +27,14 @@ impl FuzzTargetImpl for SecureJsonRpcSubmitTransaction {
         "Secure JSON RPC submit_transaction() response"
     }
 
-    fn generate(&self, _idx: usize, _gen: &mut ValueGenerator) -> Option<Vec<u8>> {
```
mimoo, line 22: you'll need a generate function for all of these, check my previous PR on how to write that : o

testsuite/libra-fuzzer/src/fuzz_targets/secure_json_rpc_client.rs
```diff
@@ -19,12 +27,14 @@ impl FuzzTargetImpl for SecureJsonRpcSubmitTransaction {
...
-
     fn fuzz(&self, data: &[u8]) {
-        fuzz_submit_transaction_response(data);
+        let runner = create_test_runner(data);
+        let strategy = arb_submit_transaction_response();
+
+        let input = get_fuzzer_response(runner, strategy);
+        if let Some(input) = input {
+            let _ = process_submit_transaction_response(input);
+        }
```
mimoo, line 37: you should be able to use the new helper instead of this code, see my previous PR as well : o

secure/json-rpc/src/lib.rs
JoshLind: Yes exactly. The client logic doesn't actually ever use the response json result, but we may as well test what happens when one is returned :)

secure/json-rpc/src/lib.rs
JoshLind: Aah, I saw this pattern used elsewhere (before you refactored things today), so figured there must be a reason! Removed them 😄

secure/json-rpc/src/lib.rs
JoshLind: Test helpers is now only used by test code, so it was either #[allow(dead_code)], or remove compilation for the fuzzing feature. I left it in because we might need this support for more complex fuzzing tests in the future, but for now I've updated it to remove compilation (if we need it in the future, we can add it back!). Looks cleaner!

JoshLind:
Thanks, @mimoo! Ready for another look. Note: I've collapsed the 3 commits into 1 because the changes are much smaller with your refactor 😄

PR #5654, opened 2020-08-17
[fuzzing] remove unused default generate function

mimoo:


mimoo:
thanks! https://github.com/libra/libra/pull/5640 needs to land first : c

PR #5655, opened 2020-08-17
[network][noise] returning EOF when receiving a frame of length 0u16

mimoo:
In our integration of noise, all noise messages are prefixed with a 16-bit length.
Previously, receiving a prefix of 0u16 would make us reset our read_state to ReadState::Init,
which would in turn prompt us to try to read another frame length.

This might seem fine, but it is a weird behavior if you're trying to read a long stream of 0s,
as this read will never end.
I believe this is the only edge-case where a read can never end
(as noise messages are limited in size due to the 16-bit length constraint),
and so I propose that we EOF if we ever receive a frame of length 0.
I don't see any reason why a peer would ever do that to us, why would they ;_;

### DIFF
--- a/network/src/noise/stream.rs
+++ b/network/src/noise/stream.rs
@@ -112,7 +112,8 @@ where
                         Ok(Some(frame_len)) => {
                             // Empty Frame
                             if frame_len == 0 {
-                                self.read_state = ReadState::Init;
+                                // 0-length messages are not expected
+                                self.read_state = ReadState::Eof(Err(()));
                             } else {
                                 self.read_state = ReadState::ReadFrame {
                                     frame_len,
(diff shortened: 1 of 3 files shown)


network/src/noise/stream.rs
```diff
@@ -112,7 +112,7 @@ where
                         Ok(Some(frame_len)) => {
                             // Empty Frame
                             if frame_len == 0 {
-                                self.read_state = ReadState::Init;
```
gregnazario, line 115: Yeah, this seemed like it wasn't a good idea to not just end for an "empty frame" anyways.  Can you write a unit test for this?

mimoo: sure thing!

mimoo: done, also I changed 

```
self.read_state = ReadState::Eof(Ok(()));
```

to

```
self.read_state = ReadState::Eof(Err(()));
```

which I think is more sensical

mimoo:
damn it, looks like another test is failing now

gregnazario:
> damn it, looks like another test is failing now

This failed for me three times this morning.  I think it's a flaky test

mimoo:
ah, let's try again then

mimoo:
looks like it worked :D

PR #5656, opened 2020-08-17
[LBT] Remove DISABLE_FN_UPGRADE from compat LBT

andll:
DISABLE_FN_UPGRADE was a temporary measure for compat test due to limitations of config builder based setup.

This PR removes DISABLE_FN_UPGRADE as at this point it actually hurts us


andll:
@rustielin there is actually some problem with calling setup_cluster twice, we'll need to dig more for future use cases, but for compat test we just don't need it

PR #5661, opened 2020-08-18
[logging] added tests around logging and levels

gregnazario:
### Overview
There were previously no tests around logging, so hopefully these new
tests will prevent accidental mistakes, as well as illustrate how
logging can be used throughout the codebase.

In addition, microseconds seems to be a better choice for the timestamp.

### DIFF
--- a/common/logger/src/tests/struct_log.rs
+++ b/common/logger/src/tests/struct_log.rs
@@ -1,0 +1,163 @@
+// Copyright (c) The Libra Core Contributors
+// SPDX-License-Identifier: Apache-2.0
+
+use crate::{
+    prelude::*, set_struct_logger, struct_log::InitLoggerError, LoggingField, StructLogSink,
+    StructuredLogEntry,
+};
+use chrono::{DateTime, Utc};
+use serde_json::{Map, Value};
+use std::{
+    io,
+    sync::mpsc::{self, Receiver, SyncSender},
+};
+
+struct StreamStructLog {
+    sender: SyncSender<StructuredLogEntry>,
+}
+
+impl StreamStructLog {
+    pub fn start_new() -> io::Result<(Self, Receiver<StructuredLogEntry>)> {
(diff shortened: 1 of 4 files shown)


common/logger/src/tests/struct_log.rs
```diff
@@ -0,0 +1,161 @@
...
+    let (logger, receiver) = StreamStructLog::start_new().map_err(InitLoggerError::IoError)?;
+    let logger = Box::leak(Box::new(logger));
+    set_struct_logger(logger).map(|_| receiver)
+}
+
+fn as_object(value: Value) -> Map<String, Value> {
+    value.as_object().unwrap().clone()
+}
+
+fn val(value: &Map<String, Value>, key: &str) -> Value {
```
zekun000, line 46: you probably want to have a `impl Map<String, Value>` as helper functions

common/logger/src/tests/struct_log.rs
```diff
@@ -0,0 +1,161 @@
...
+    value.get(key).unwrap().clone()
+}
+
+fn string(value: &Map<String, Value>, key: &str) -> String {
+    String::from(val(value, key).as_str().unwrap())
+}
+
+fn json_map(receiver: &Receiver<StructuredLogEntry>) -> Map<String, Value> {
```
zekun000, line 54: ditto, use impl and self

common/logger/src/tests/struct_log.rs
gregnazario: Can you `impl Map<String, Value>` outside of the Map's crate?

gregnazario: nvm, just made a trait for it

zekun000: I think you can? I only remember you can't impl for a trait outside the crate

gregnazario: I had to create a trait for it, as you can't impl for the struct outside of its crate.

PR #5664, opened 2020-08-18
[Move prover] Added specs to LibraConfig

DavidLDill:
Added many local specs to LibraConfig and LibraSystem.  Made more functions opaque.

I changed the parameter name config_account to lr_account, since it is
required to be Libra Root in all cases.

Changed specs caused some issues in RegisteredCurrencies, which I fixed.

In Libra system: There are more detailed specifications of
add_validator, remove_validator, and update_config_and_reconfigure. I
made set_validator_sets opaque in LibraSystem.  This change seems to
allow remove_validator to complete in reasonable time, but
add_validator and update_config_and_reconfigure still time out. There
is a bug marked that should lead to a violation of the postcondition I
added to add_validator, so that may be part of the problem.

There are three "TODO (dd) BUG" comments in LibraConfig.move.

1. Errors from imported schemas are (sometimes?) reported in the
module that defines the scheme, not where the error occurs.

2. result == Foo {} (where Foo is a resource with no fields) fails
because the spec language thinks it has an empty value array, while
the "pack" bytecode instruction gives it a dummy field (I don't know
why).

3. There is a panic on some specifications that have wrong type
parameters, instead of a more graceful error message.

There is one bug reported in multiple places in LibraSystem.move:
"old" sometimes gets lost when using "let" and "spec" defines.
This bug can result in false negatives.

Issues have been field for all the above bugs.

Thank you for sending a PR. We appreciate you spending time to help improve the Libra project.

The project is undergoing daily changes. Pull Requests will be reviewed and responded to as time permits.
-->

## Motivation

(Write your answer here.)

## Test Plan

## Related PRs


language/stdlib/modules/LibraConfig.move
```diff
@@ -100,38 +121,94 @@ module LibraConfig {
...
+        pragma opaque = true;
+        // TODO (dd) BUG: the erroneous line caused a panic.
+        //   modifies global<Config>(CoreAddresses::LIBRA_ROOT_ADDRESS());
+        
+        // Everything has to be published at LIBRA_ROOT_ADDRESS because
+        // publish_new_config_and_get_capability aborts otherwise.
+        modifies global<LibraConfig<Config>>(CoreAddresses::LIBRA_ROOT_ADDRESS());
+        modifies global<ModifyConfigCapability<Config>>(CoreAddresses::LIBRA_ROOT_ADDRESS());
+        include PublishNewConfigAbortsIf<Config>;
+        // TODO (dd): the erroneous ensures below caused a panic.  Here is part of the stack trace:
```
wrwg, line 188: Interesting. That is worth an issue (every crash should). Feel free to keep it short and just point to here. You can also replace TODO(dd) with TODO(#<issue number>) if you open those issues before you merge. This may have the lowest boilerplate, as the actual description is easily discoverable by search.

language/stdlib/modules/LibraConfig.move
```diff
@@ -180,6 +257,23 @@ module LibraConfig {
         );
     }
 
+    spec fun emit_reconfiguration_event {
+        pragma addition_overflow_unchecked = true;
+        // pragma opaque = true;
+        // TODO (dd): Should this warn if modifies are incomplete?
+        //    Is it dangerous to make this opaque with incomplete modifies?
```
wrwg, line 264: This is a good question. 

Currently we treat modifies like aborts_if -- if we do not provide one, anything is allowed. This allows us to incrementally add modifies (and aborts_if).

Perhaps when a function is declared *opaque*, we should choose as a default that both modifies and aborts_if are strict. 

I don't know whether this is possible for modifies, it is definitely for aborts_if. @shazqadeer @emmazzz

language/stdlib/modules/LibraConfig.move
```diff
@@ -215,22 +309,10 @@ module LibraConfig {
...
-        include Roles::AbortsIfNotLibraRoot{account: config_account};
-        aborts_if spec_is_published<Config>();
-        aborts_if exists<ModifyConfigCapability<Config>>(Signer::spec_address_of(config_account));
-    }
-    spec schema PublishNewConfigEnsures<Config> {
-        config_account: signer;
-         ensures spec_is_published<Config>();
-         ensures exists<ModifyConfigCapability<Config>>(Signer::spec_address_of(config_account));
-    }
+    // DD: spec fun publish_new_config {
```
wrwg, line 312: This can be deleted now.

language/stdlib/modules/LibraSystem.move
```diff
@@ -90,12 +90,28 @@ module LibraSystem {
     // This copies the vector of validators into the LibraConfig's resource
     // under ValidatorSet address
     fun set_validator_set(value: LibraSystem) acquires CapabilityHolder {
+        assert(
+            exists<CapabilityHolder>(CoreAddresses::LIBRA_ROOT_ADDRESS()),
```
wrwg, line 94: Is this function called from genesis (initialize)? If not you can just add `LibratimeStamp::assert_operating()` and the according schema to the spec. By the invariant, it follows then that the capability exists, and you don't need this assert.

PR #5665, opened 2020-08-18
Parameterize build, remove redundant step

rexhoffman:
## Motivation

Remove redundant build step.
Simplify/Standardize config with parameters.
Currently push new docker.io/libra/build_environment:circleci-latest whenever a file that would modify the image is committed and lands in master
Will need a second commit to update the build to the new tag once this lands.

Yes

## Test Plan

CI, Folding the build artifacts like origami.

## Related PRs

None


.circleci/config.yml
```diff
@@ -1037,60 +893,43 @@ workflows:
...
               ignore:
-                - auto
                 - /^test-[\d|.]+$/
                 - /^release-[\d|.]+$/
                 - master
-      - dev_setup-test-centos:
-          requires:
-            - dev-setup-changes
+      - docker-ci-build:
+          base-image: centos
```
sausagee, line 912: Why are we building these variants?

.circleci/config.yml
```diff
@@ -653,165 +660,20 @@ jobs:
...
-              echo no relevant files have changed - exit
-              circleci step halt
-            fi
-            set -x
-            #building with content trust tends to break in circleci
-            export DOCKER_CONTENT_TRUST=0
-            docker build -f docker/ci/Dockerfile -t libra/build_environment:latest .
             #signs the pushed image
-            docker push --disable-content-trust=false libra/build_environment:latest
+            docker push --disable-content-trust=false libra/build_environment:<< parameters.base-image >>-latest
```
sausagee, line 676: Need to think about DockerHub image retention policy soon :)

.circleci/config.yml
rexhoffman: to test the dev_setup.sh that claims to support centos.  now we know.

.circleci/config.yml
rexhoffman: I'm aware.   This only builds on changes to the relevant files.

rexhoffman: I'm not done with the nightly run yet that will be triggered on master to handle the cleanups.   It's a bit of a pia, and got side tracked with lcov stuff today.

PR #5666, opened 2020-08-18
 [cluster-test] Remove config-builder usage

andll:
- Switch to libra image without config builder
- Remove enable_mgmt_tool config
- Merge full node and validator specs into one, since they are the same


docker/validator/docker-run.sh
```diff
@@ -4,9 +4,9 @@
 set -ex
 
 cd /opt/libra/etc
-echo "$NODE_CONFIG" > node.yaml
-echo "$SEED_PEERS" > seed_peers.yaml
-echo "$NETWORK_KEYPAIRS" > network_keypairs.yaml
-echo "$CONSENSUS_KEYPAIR" > consensus_keypair.yaml
-echo "$FULLNODE_KEYPAIRS" > fullnode_keypairs.yaml
+[ -n "${DEPLOY_ENV}" ] && echo "$NODE_CONFIG" > node.yaml
```
rustielin, line 7: What is DEPLOY_ENV used for? Can't find reference in the existing codebase

testsuite/cluster-test/src/cluster_swarm/cluster_swarm_kube.rs
```diff
@@ -153,76 +141,45 @@ impl ClusterSwarmKube {
...
-            cfg_seed = CFG_SEED,
-            cfg_seed_peer_ip = seed_peer_ip,
-            cfg_fullnode_seed = CFG_FULLNODE_SEED,
         );
         let pod_spec: serde_yaml::Value = serde_yaml::from_str(&pod_yaml)?;
         let pod_spec = serde_json::value::to_value(pod_spec)?;
         serde_json::from_value(pod_spec)
             .map_err(|e| format_err!("serde_json::from_value failed: {}", e))
     }
+    //
```
rustielin, line 163: Why keep this as comment?

docker/validator/docker-run.sh
andll: Oh, good eye. This is indeed a bug

testsuite/cluster-test/src/cluster_swarm/libra_node_spec_template.yaml
```diff
@@ -37,19 +24,21 @@ spec:
     - mountPath: /opt/libra/data
       name: data
   - name: debug
-    image: 853397791086.dkr.ecr.us-west-2.amazonaws.com/libra_validator:{image_tag}
+    image: 853397791086.dkr.ecr.us-west-2.amazonaws.com/libra_e2e:{image_tag}
```
rustielin, line 27: I think you will have to update `docker/build-aws.sh`'s `--build-all-cti` list of projects to include the `libra_e2e` image. According to our ECR, that repo hasn't been pushed to in more than a year

rustielin: er nevermind

rustielin: This image gets built in the validator CodeBuild spec right?

andll: So actually libra_e2e is build and pushed as part of build and push of libra_validator

rustielin: Got it!

andll:
/land

Yep, tested indeed :)

andll:
/land

Adding minor commit with storage config

andll:
Looks like my small last commit broke it, i'll make a fix

docker/validator/docker-run.sh
```diff
@@ -4,9 +4,9 @@
...
-echo "$NODE_CONFIG" > node.yaml
-echo "$SEED_PEERS" > seed_peers.yaml
-echo "$NETWORK_KEYPAIRS" > network_keypairs.yaml
-echo "$CONSENSUS_KEYPAIR" > consensus_keypair.yaml
-echo "$FULLNODE_KEYPAIRS" > fullnode_keypairs.yaml
+[ -n "${NODE_CONFIG}" ] && echo "$NODE_CONFIG" > node.yaml
+[ -n "${SEED_PEERS}" ] && echo "$SEED_PEERS" > seed_peers.yaml
+[ -n "${NETWORK_KEYPAIRS}" ] && echo "$NETWORK_KEYPAIRS" > network_keypairs.yaml
+[ -n "${CONSENSUS_KEYPAIR}" ] && echo "$CONSENSUS_KEYPAIR" > consensus_keypair.yaml
+[ -n "${FULLNODE_KEYPAIRS}" ] && echo "$FULLNODE_KEYPAIRS" > fullnode_keypairs.yaml
```
davidiw, line 11: these don't exist any more

andll:
Ok, I finally realized what is wrong with this PR

It has 'breaking' change in a sense that I updated `docker-run.sh`. This docker-run is compatible for purpose of production, but it is not compatible for cluster test (original one overwrites cfg with empty file when config environment var is not set)

Adding breaking label here to skip compat test, otherwise this PR is good to go. This incompatibility should not affect production in any way

/land

PR #5667, opened 2020-08-18
[docker] add option to toggle build of non-release binaries

rustielin:
This PR adds an optional argument to `docker/build-common.sh` that can
be used to skip the build of test/non-release binaries.
`docker/build-common.sh true` does a full build. Changes have been
made to the appropriate Dockerfiles that require test binaries
(e.g. faucet, client, and cluster-test). This should hopefully
shave off some time in LBT build phase.

There will still need to be some optimization on cluster-test side (e.g. cluster clean up routine.)

### DIFF
--- a/docker/build-common.sh
+++ b/docker/build-common.sh
@@ -14,6 +14,9 @@ export CARGO=$(rustup which --toolchain $RUST_NIGHTLY cargo)
 export CARGOFLAGS=$(cat cargo-flags)
 export CARGO_PROFILE_RELEASE_LTO=thin # override lto setting to turn on thin-LTO for release builds
 
+# skip building non-release binaries to shave off time
+BUILD_TEST=$1
+
 # Build release binaries
 ${CARGO} ${CARGOFLAGS} build --release \
          -p libra-genesis-tool \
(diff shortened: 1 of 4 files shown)


andll:
I am worried about this change
- It probably not going to reduce LBT time, since all builds are done in parallel and unless all of them are shorter, there won't be noticeable effect
- However, in the future it will mean that there won't be single docker layer between cluster test and other builds, that might affect our ability to cache things(depending on how build system will evolve)

rustielin:
@andll Not too sure about the caching point (cc @rexhoffman). On the parallel build point, now the build time will be bottlenecked by only the build of cluster-test, rather than the max of all builds. It probably won't (consistently) shave off too much time, but it's something small we can do now, as long as there are no unforeseen side effects in the build system cache like you mention.

mgorven:
For Circle builds this will perform the main build twice (once for `build-common.sh` and again for `build-common.sh test`). For this to work `build-common.sh test` needs to skip the main build completely.

andll:
Note that even if `build-common.sh test` skips main build, it will still do pretty much 90% of same work(which is compiling same rust crates etc). Only last mile of some dependencies that are different and linking with LTO are different in those two steps

rustielin:
Ah gotcha. Thanks for the clarification @andll and @mgorven. Seems like there are some other things we can do to speed up build, like sccache. I'll close this for now while we converge on some better options.

PR #5670, opened 2020-08-19
[Fuzzing] Add fuzz targets to the vault client of secure storage.

JoshLind:
## Motivation

This PR refactors the vault client to better allow fuzz targets to be added to client response handling logic. Moreover, it adds 3 new fuzz targets as a means to show how these fuzzers can be implemented. Further fuzz targets for the remaining processing logic will be added in a following PR (I wanted to avoid making this PR too large).

The PR offers 5 commits:
1. Refactor the vault client response handling logic by moving handling logic into dedicated functions that can be fuzzed.
2. Remove handling functions that contain duplicate logic.
3. Add a fuzzing target for process_generic_response().
4. Add a fuzzing target for process_policy_list_response().
5. Add a fuzzing target for process_secret_data_response().

While the fuzzing targets added by this PR are somewhat simple, they highlight the approach being taken. More complex targets will be added once this PR lands.

Yes.

## Test Plan

All existing and newly added tests pass locally. Moreover, running the fuzzers locally appears to work. Once this PR lands, we'll further be able to verify the kind of coverage we're seeing in our code coverage reports.

## Related PRs

None.


testsuite/libra-fuzzer/src/fuzz_targets/secure_storage_vault.rs
```diff
@@ -0,0 +1,72 @@
...
+        "Secure storage vault: process_secret_read_response()"
+    }
+
+    fn generate(&self, _idx: usize, _gen: &mut ValueGenerator) -> Option<Vec<u8>> {
+        Some(corpus_from_strategy(arb_secret_read_response()))
+    }
+
+    fn fuzz(&self, data: &[u8]) {
+        let secret = any::<String>();
+        let key = any::<String>();
```
mimoo, line 68: do you know how this works?so I'm a bit confused on what's happening here, `any` is supposed to generate a strategy, but `process_secret_read_response()` take two `&str` not strategies oO

secure/storage/vault/src/lib.rs
```diff
@@ -513,6 +358,230 @@ impl Client {
     }
 }
 
+/// Processes a generic response returned by a vault request. This function simply just checks
+/// that the response was not an error and calls response.into_string() to clear the ureq stream.
+pub fn process_generic_response(resp: Response) -> Result<(), Error> {
+    if resp.ok() {
```
mimoo, line 364: according to the doc, this checks

> Whether the response status is: 200 <= status <= 299

do we want to check that the status is 200 exactly instead here?

secure/storage/vault/src/lib.rs
```diff
@@ -724,7 +793,7 @@ struct ExportKey {
 /// }
 #[derive(Debug, Deserialize, PartialEq, Serialize)]
 struct ListPoliciesResponse {
-    policies: Vec<String>,
+    pub policies: Vec<String>,
```
mimoo, line 796: why setting all these fields to pub?

mimoo:
good job :) left you some comments!

secure/storage/vault/src/lib.rs
JoshLind: This is a good question for @davidiw, but my preference right now is not to change the logic in this PR. If we want to make a change to logic, I'd suggest doing it in another PR to keep things simple 😄

secure/storage/vault/src/lib.rs
JoshLind: Aah, thanks David! This was a legacy thing from an old approach I tried last week! Fixed 😄

testsuite/libra-fuzzer/src/fuzz_targets/secure_storage_vault.rs
JoshLind: Hah! 😕 I spent some time trying to figure this out, and I was a bit confused too.. I'm surprised the compiler didn't kick up a fuss.. anyway, to avoid ambiguity, I've updated this to use the input data to generate the two values (similar to the response). Thanks, David!

JoshLind:
Thanks, @mimoo! Ready for another look when you are 😄

testsuite/libra-fuzzer/src/fuzz_targets/secure_storage_vault.rs
mimoo: and you could run the fuzzer fine?
I think the tests might have been able to run fine because they were inside a proptest! (maybe)

testsuite/libra-fuzzer/src/fuzz_targets/secure_storage_vault.rs
```diff
@@ -0,0 +1,73 @@
...
+    }
+
+    fn generate(&self, _idx: usize, _gen: &mut ValueGenerator) -> Option<Vec<u8>> {
+        Some(corpus_from_strategy(arb_secret_read_response()))
+    }
+
+    fn fuzz(&self, data: &[u8]) {
+        let secret = fuzz_data_to_value(data, any::<String>());
+        let key = fuzz_data_to_value(data, any::<String>());
+        let response = fuzz_data_to_value(data, arb_secret_read_response());
```
mimoo, line 69: OK you actually cannot do this : o and maybe I should refactor fuzz_data_to_value to make it clearer. The reason why is that you're going to use the same data to produce secret, key, and response. Instead you want to use different parts of data to do this. I would suggest doing something like this:

```
let (secret, key, response) = fuzz_data_to_value(data, some_strategy());
```

secure/storage/vault/src/lib.rs
mimoo: yeah I was mostly commenting on the code I was reading, and this tripped me a bit

testsuite/libra-fuzzer/src/fuzz_targets/secure_storage_vault.rs
JoshLind: Yeah, it ran! Super weird! 😕

testsuite/libra-fuzzer/src/fuzz_targets/secure_storage_vault.rs
mimoo: does secret and key have to be fuzzed? In other words can they be fixed?

I'm not sure how proptest generates String, but I'm guessing it might have some size limitation (cc @sunshowers). I would recommend inverting the order of the strategy so that the lengths of secret & key don't affect the fuzzing of response (if that makes sense):

```
let (response, secret, key) = fuzz_data_to_value(data, some_strategy());
```

JoshLind: Aah, gotcha! Fixed, thanks! 😄 

1. I think it's better if secret and key are fuzzed, for completeness sake.. Just to be clear though (for my understanding), technically, there is nothing stopping someone from taking the approach of calling 'fuzz_data_to_value()' multiple times, right?.. It's more just that it's not ideal a single change in the input data should change all 3 values.. rather we'd like a single change in the the input data to only change 1 of the values at a time.

2. Regarding the ordering, that's an open question to me, too. My feeling is that if there is some limit to the input data, then this could be a problem. Although, I'm interested to know what happens in these cases? Would it just repeat the input data as the source of randomness? Anyway, I've re-ordered them as suggested! 😄

mimoo: 2 -> yeah you're right, you can call it multiple times but that's probably not what you want to do

mimoo: 3 -> if proptest runs out of random bytes it'll use 0s to fill in the rest

secure/storage/vault/src/lib.rs
```diff
@@ -513,6 +358,230 @@ impl Client {
     }
 }
 
+/// Processes a generic response returned by a vault request. This function simply just checks
+/// that the response was not an error and calls response.into_string() to clear the ureq stream.
+pub fn process_generic_response(resp: Response) -> Result<(), Error> {
```
davidiw, line 363: these kinda make me sad as it adds a bit more layers... perhaps instead we should make these methods on the structs themselves:

```
  from_response(resp: Response) -> Result<Self, Error> {
    ...
  }
```
thoughts?

davidiw:
apologies for commenting so late in the process...

secure/storage/vault/src/lib.rs
JoshLind: No worries! We'll take this offline and respond as appropriate in a follow up PR 😄

PR #5671, opened 2020-08-19
Set up cache file as a daily latch for code coverage and fresh circleci build images

rexhoffman:
## Motivation

Set up a latches to allow the first master build on any given day to:

1. generate and push code coverage.
2. generate a new circleci base image.  libra/build_environment:circleci-latest

Yes

## Test Plan

All the CIs

## Related PRs

None

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -8,27 +8,27 @@ orbs:
 executors:
   build-executor:
     docker:
-      - image: docker.io/libra/build_environment:latest
+      - image: docker.io/libra/build_environment:circleci-latest
     resource_class: 2xlarge
   unittest-executor:
     docker:
-      - image: docker.io/libra/build_environment:latest
+      - image: docker.io/libra/build_environment:circleci-latest
     resource_class: 2xlarge+
   test-executor:
     docker:
-      - image: docker.io/libra/build_environment:latest
+      - image: docker.io/libra/build_environment:circleci-latest
     resource_class: xlarge
   premainnet-cluster-test-executor:
     docker:
-      - image: docker.io/libra/build_environment:latest
+      - image: docker.io/libra/build_environment:circleci-latest
     resource_class: xlarge
   audit-executor:
     docker:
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -770,22 +815,32 @@ jobs:
...
+            #Push lcov
+            aws s3 cp --recursive ${CODECOV_OUTPUT}/lcovhtml "s3://${PREFIX}/lcov-unit-coverage/${SUFFIX}/";
+            aws s3 cp --recursive ${CODECOV_OUTPUT}/lcovhtml "s3://${PREFIX}/lcov-unit-coverage/latest/";
+            echo "lcov available in s3 https://${PREFIX}/lcov-unit-coverage/${SUFFIX}/index.html" >> ${MESSAGE_PAYLOAD_FILE}
       - send_message:
           payload_file: "${MESSAGE_PAYLOAD_FILE}"
           build_url: "${CIRCLE_BUILD_URL}#tests/containers/${CIRCLE_NODE_INDEX}"
           webhook: "${WEBHOOK_COVERAGE_INFO}"
+      - run:
+          name: Update cache key file
```
metajack, line 836: Will this conflict with the other cache check? Whichever runs first will update the cache key and then the other will not run when it should.

rexhoffman: No, different cache keys means a different cache will be pulled in.   I could change the file names if you find this confusing.   Of course if might be more confusing the other way different files different cache keys not visible togehter.

.circleci/config.yml
```diff
@@ -893,12 +948,24 @@ workflows:
               only:
                 - /^test-[\d|.]+$/
                 - /^release-[\d|.]+$/
+      - code_coverage:
+          requires:
+            - prefetch-crates
```
sausagee, line 953: This is not needed for the following reasons:
- the `code_coverage` job does not restore the pre-fetched crates from cache yet
- after the restore step is added to the job above, the cache entry was created by prior jobs (during PR or auto branch during land) that touched the cache key.

PR #5672, opened 2020-08-19
[libra framework] fix all TODOs in LibraBlock

sblackshear:
- Address TODO about making block prologue private. This changes the module bytecode, but is not a client-breaking change and need not be deployed to testnet.
- Remove TODO about triggering reconfiguration in block prologue. Reconfiguration should only be triggerable by LibraRoot or a validator operator
- Remove TODO about block height. We have decided to keep it.


language/stdlib/modules/LibraBlock.move
```diff
@@ -90,13 +86,12 @@ module LibraBlock {
...
-                round: round,
-                proposer: proposer,
-                previous_block_votes: previous_block_votes,
+                round,
+                proposer,
+                previous_block_votes,
                 time_microseconds: timestamp,
             }
         );
-        // TODO(valerini): call regular reconfiguration here LibraSystem2::update_all_validator_info()
```
sblackshear, line 99: @valerini: Just double-checking that this is an outdated comment left over from when we used to trigger reconfiguration automatically at block boundaries.

davidiw: she may be offline for a bit, if memory serves me correct, this was intended to be used for batching within a block.

bob-wilson: If she's not around, maybe file a GitHub issue about looking into this and then go ahead with the change. It's not like removing a comment is going to break anything....

sblackshear: I remember the same as David, but I will double-check with Lera when she gets back.

PR #5674, opened 2020-08-19
[breaking][libra framework] Remove LibraTransactionTimeout module

sblackshear:
The only purpose of `LibraTransactionTimeout` was to an enforce upper and lower bounds on transaction expiration time. The lower bound check is very simple to do and the upper bound check is no longer a requirement, so we can kill this module.

The PR deletes `LibraTransactionTimeout` and moves the lower bound check into the transaction prologue. To keep the check simple, it adds a `LibraTimestamp::now_seconds` function that can be directly compared against the transaction expiration time (which is specified in seconds).

This is a breaking change in the sense that it removes the `LibraTransactionTimeout` module and orphans the `TTL` resource, but it is not observable to clients and should be deployable via a WriteSet.


riyazdf:
noted, thanks!

n4ss:
Looks good!

language/stdlib/modules/LibraTimestamp.move
```diff
@@ -151,7 +153,7 @@ module LibraTimestamp {
     }
 
 
-    /// Gets the timestamp representing `now` in microseconds.
+    /// Gets the current time in microseconds.
     public fun now_microseconds(): u64 acquires CurrentTimeMicroseconds {
```
tzakian, line 157: More of a meta question here: now that we have `now_seconds` do we want to keep `now_microseconds` around as part f the public API since we don't need that fine of a granularity on time?

sblackshear: Great question! I don't see `now_microseconds` used anywhere else except for `AccountLimits`, and I would be very much in favor of replacing that usage with `now_seconds` and hiding `now_microseconds`. Let me try to tackle this before landing.

sblackshear: Starting looking at this and it feels like a big enough change that it should happen in a separate PR. But definitely worth doing!

PR #5680, opened 2020-08-19
[vm] Fix native function errors

tnowacki:
- Some native functions (most notably LCS) threw system-level errors instead of aborting.
- Changed the NativeContext to only have Options to prevent accidentally propagating errors
- Fixed spot in LCS
- Fixed code that needlessly was wrapped in a Result

## Motivation

- Make sure native functions are behaving correctly

## Test Plan

- new test
- cargo test

### DIFF
--- a/language/move-vm/types/src/natives/function.rs
+++ b/language/move-vm/types/src/natives/function.rs
@@ -24,6 +24,9 @@ use move_core_types::{
 use std::fmt::Write;
 use vm::errors::PartialVMResult;
 
+pub use move_core_types::vm_status::StatusCode;
+pub use vm::errors::PartialVMError;
+
 /// `NativeContext` - Native function context.
 ///
 /// This is the API, the "privileges", a native function is given.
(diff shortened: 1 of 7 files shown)


language/move-vm/types/src/natives/function.rs
```diff
@@ -102,6 +114,24 @@ pub fn native_gas(table: &CostTable, key: NativeCostIndex, size: usize) -> GasUn
 #[macro_export]
 macro_rules! pop_arg {
     ($arguments:ident, $t:ty) => {{
-        $arguments.pop_back().unwrap().value_as::<$t>()?
+        match $arguments.pop_back().map(|v| v.value_as::<$t>()) {
+            None => {
+                return NativeResult::invariant_violation(vm::errors::PartialVMError::new(
```
tnowacki, line 119: @tzakian How do I set this up properly so that it always refers to the correct crate?

language/move-vm/natives/src/signature.rs
```diff
@@ -33,14 +32,14 @@ pub fn native_ed25519_publickey_validation(
     // This deserialization performs point-on-curve and small subgroup checks
     let valid = ed25519::Ed25519PublicKey::try_from(&key_bytes[..]).is_ok();
     let return_values = vec![Value::bool(valid)];
-    Ok(NativeResult::ok(cost, return_values))
+    NativeResult::ok(cost, return_values)
 }
 
 pub fn native_ed25519_signature_verification(
```
sblackshear, line 38: Ooh, this is a nasty one. I think we can/should make this one abort-free; I will put up a separate PR for that.

tnowacki: I'm not sure if I follow. This function cannot abort

tnowacki:
@dariorussi A lot of this was me being too clumsy in tryin to enforce than any error is an invariant violation. 
It is really bizarre to allow native functions to throw anything else, so I still want that debug_assert. But I'll move it around. (Unless of course you think native functions should be able throw other things, which is weird, but we could discuss it)

In this case, `type_to_type_layout` can throw `VM_MAX_VALUE_DEPTH_REACHED` on really large values. @vgao1996 and I talked and we want to somehow move that error out of `ExecutionFailure` and into it's own bucket, something like `UnableToProduceChangeSet` or something. In order to do this, I switched `type_to_type_layout` to an `Option` to make aborting the easiest thing for the function (and any future users) to do

language/move-vm/natives/src/signature.rs
tnowacki: Was reading a comment for the wrong spot :) Glad to see the fix on the other PR!

PR #5683, opened 2020-08-19
[CI] attempt to have multiple non-overlapping flows

rexhoffman:
## Motivation

The circleci commit-workflow could trigger on multiple branches/pushes and cause github to get confused by the check performed by circleci.   Potentially pointing a pr at the wrong build from another branch.

Create multiple workflows that prevent collisions in circleci github checks.   If it works the way I expect.

Yes

## Test Plan

So much CI

## Related PRs

None.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -845,111 +845,12 @@ jobs:
           webhook: "${WEBHOOK_COVERAGE_INFO}"
 
 workflows:
-  commit-workflow:
+  ######################################################################################################################
+  # Once a day when a change lands on master, rebuild the CI base image and code coverage.
+  # Will build the CI base image when relevant files are changed vs on master, on either master or test branches.
+  ######################################################################################################################
+  latched-publish-workflow:
     jobs:
-      - lint-docker:
-          filters:
-            branches:
-              ignore:
-                - master
-                - /^test-[\d|.]+$/
-                - /^release-[\d|.]+$/
-      - terraform:
-          filters:
-            branches:
-              ignore:
-                - master
-                - /^test-[\d|.]+$/
-                - /^release-[\d|.]+$/
-      - prefetch-crates
-      - lint:
-          requires:
-            - prefetch-crates
-          filters:
-            branches:
-              ignore:
-                - gh-pages
-                - /^test-[\d|.]+$/
-                - /^release-[\d|.]+$/
-      - build-dev:
-          requires:
-            - prefetch-crates
-          filters:
-            branches:
-              ignore:
-                - master
-                - /^test-[\d|.]+$/
-                - /^release-[\d|.]+$/
-      - run-e2e-test:
-          requires:
-            - prefetch-crates
-          filters:
-            branches:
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -937,69 +903,129 @@ workflows:
           filters:
             branches:
               only: master
+
+  bors-auto-workflow:
```
bmwill, line 907: Bors is explicitly looking for a workflow titled `commit-workflow` so if we land this as is no PR will be able to land since the check bors is looking for wont ever happen. We can change this, it would just require re-deploying bors with a config change.

.circleci/config.yml
```diff
@@ -937,69 +903,129 @@ workflows:
...
           base-image: arch
           filters:
             branches:
-              ignore:
-                - /^test-[\d|.]+$/
-                - /^release-[\d|.]+$/
-                - master
+              only: auto
+
+  pull-request-workflow:
```
bmwill, line 990: Do we need a separate "pull-request" and "bors" workflow? Might be easier to understand if these were the same, simply with the addition of the additional docker build/publish on the auto branch?

.circleci/config.yml
rexhoffman: No problem, changing.   I thought bors was looking at all git checks on the pr?

.circleci/config.yml
rexhoffman: Name collision prevention.   Separate checks will be easier to understand what roll they actually play.   I do wish I could group the jobs for reuse though.

rexhoffman:
K doing the right thing for /canary
https://app.circleci.com/pipelines/github/libra/libra/27885/workflows/fe3c5fc0-2e1d-4c4e-8b17-d76e760eb5bf

.circleci/config.yml
```diff
@@ -845,161 +845,202 @@ jobs:
...
+      - prefetch-crates:
           filters:
             branches:
-              only:
-                - /^test-[\d|.]+$/
-                - master
+              only: /^pull\/.+$/
+      - lint:
+          requires:
+            - prefetch-crates
```
sausagee, line 1003: Curious why we don't have the branch filters here.  Does that mean this job will trigger if and only if `prefetch-crates` triggers and runs?

.circleci/config.yml
```diff
@@ -845,161 +845,202 @@ jobs:
...
+            branches:
+              only:
+                - auto
+                - canary
+      - lint:
+          requires:
+            - prefetch-crates
       - build-dev:
           requires:
             - prefetch-crates
```
sausagee, line 941: Similarly, crious why we don't need to the branch filter here.

rexhoffman: the requires acts as the filter.... since prefetch crates had a filter, the linter doesn't need it.

sausagee: TIL - that's cool feature.

rexhoffman:
Now how would I cancel a canary?

/help

PR #5684, opened 2020-08-19
[gha][lbt] add suggestion when report empty

rustielin:
Some runs of LBT have legitimate errors, but the report makes it sounds like it's a `cti` error with not being able to parse the report. e.g. https://github.com/libra/libra/pull/5666#issuecomment-676805350 among others

### DIFF
--- a/.github/workflows/land-blocking.yml
+++ b/.github/workflows/land-blocking.yml
@@ -212,7 +212,7 @@ jobs:
                   console.error("Failed to check infra error in CT output log.\n", err);
                 }
               } else {
-                body = "Cluster Test failed - test report processing failed.";
+                body = "Cluster Test runner failed.";
                 console.error(err);
               }
               body += " See https://github.com/libra/libra/actions/runs/${{github.run_id}}";


andll:
So this is a bit misleading, in a sense that `no report` can have more then just one reasonable explanation. For example it can be just panic in any other place in cluster test

To fix this we probably just need to report error when cluster builder fails to setup cluster. I can take a look at it (or, if you want , feel free to do it)

rustielin:
> So this is a bit misleading, in a sense that no report can have more then just one reasonable explanation. For example it can be just panic in any other place in cluster test

Yea that's true. I see this as just a suggestion/guess at what's going on under the hood, so we can get more devs to actually look at the cluster-test logs rather than reporting that reporting is broken. I agree that we should probably have a way to report error outside of experiments though. That might get a bit messy depending on how I do it -- I'll let you take a look at that :)

andll:
Sure :)
https://github.com/libra/libra/pull/5685

rustielin:
@andll Changed message to be more generic after discussion in https://github.com/libra/libra/pull/5685#issuecomment-676821826

PR #5685, opened 2020-08-19
[cluster-test] Report error when cluster setup fails

andll:
This is needed so user get better feedback in land blocking bot comment


andll:
I think for other errors for now we can repurpose your PR, just change message there to something like 'cluster test runner has failed'

/land

PR #5686, opened 2020-08-20
[logging] Guarantee fully written logs are kept

gregnazario:
### Overview
The previous iteration used a much smaller buffer for logs, as well as
kept an unmaintained size of the queue.  This now gives a bounded size
for the queue.  In addition, we were dropping logs when we couldn't
connect to the TCP endpoint.  Now, we will keep retrying to connect
without dropping logs.

### Where we still drop logs?
We still will drop logs if there's a disconnection in the middle of writing a log.  Because the logstash side isn't yet handling duplicate logs with the id, I went on a conservative approach, rather than resubmitting the log.

### Testing
Cluster test successfully writes logs.  It's not the greatest of tests because it doesn't test large amounts of pressure on the logging system.

### DIFF
--- a/common/logger/src/struct_log.rs
+++ b/common/logger/src/struct_log.rs
@@ -16,7 +16,7 @@ use std::{
     fmt::Display,
     fs::{File, OpenOptions},
     io,
-    io::Write as IoWrite,
+    io::Write,
     marker::PhantomData,
     net::{TcpStream, ToSocketAddrs},
     str::FromStr,
@@ -45,7 +45,7 @@ const INITIALIZED: usize = 2;
 // Size configurations
 const MAX_LOG_LINE_SIZE: usize = 10240; // 10KiB
 const LOG_INFO_OFFSET: usize = 256;
-const WRITE_CHANNEL_SIZE: usize = 1024;
+const WRITE_CHANNEL_SIZE: usize = 10000; // MAX_LOG_LINE_SIZE * 10000 = max buffer size
 const WRITE_TIMEOUT_MS: u64 = 2000;
 const CONNECTION_TIMEOUT_MS: u64 = 5000;
 
@@ -426,13 +426,13 @@ impl FileStructLogThread {
 
 /// Sink that streams all structured logs to an address through TCP protocol
 struct TCPStructLog {
-    sender: SyncSender<StructuredLogEntry>,
+    sender: SyncSender<String>,
 }
 
 impl TCPStructLog {
-    pub fn start_new(address: String) -> io::Result<Self> {
+    pub fn start_new(endpoint: String) -> io::Result<Self> {
         let (sender, receiver) = mpsc::sync_channel(WRITE_CHANNEL_SIZE);
-        let sink_thread = TCPStructLogThread::new(receiver, address);
+        let sink_thread = TCPStructLogThread::new(receiver, endpoint);
         thread::spawn(move || sink_thread.run());
         Ok(Self { sender })
     }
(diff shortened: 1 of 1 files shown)


common/logger/src/struct_log.rs
```diff
@@ -426,108 +426,122 @@ impl FileStructLogThread {
...
 
-    /// Writes a log line into json_lines logstash format, which has a newline at the end
-    fn write_log_line(stream: &mut TcpStream, message: String) -> io::Result<()> {
-        let message = message + "\n";
-        stream.write_all(message.as_bytes())
+    /// Ensure that we get a connection, no matter how long it takes
+    /// This will block until there is a connection
+    fn refresh_connection(&mut self) {
+        loop {
+            // TODO: Backoff?
```
davidiw, line 518: even sleeping for a second?

gregnazario:
> Logic is a little interesting in that we block on the last message until we get connected. It might be good to document the behavior of the TcpWriter in a doc block.

The other option is to have two queues, a resubmission queue, and a sending queue.  Then we could possibly have two threads writing so we always get the latest logs, as well as not drop logs.  This didn't happen that often, but when we did disconnect, the old logic would just straight up drop the logs.

> Seems this is mostly a refactor with one additional condition, we block logging until we have a connection. 

It is mostly a refactor with one additional condition.  It was refactored this way so that:
1. It's logically simpler and less confusing
2. It allows for usage with other logging frameworks such as tracing if that is a path we go down

This is a quick fix in general, the reason we now parse the JSON before writing to the Sender rather than after is to keep the storage to a maximum fixed size.  Before, we'd have messages that could be very large in the sender.  

>If we don't get a connection, we'll keep around our oldest logs until we do. Interesting. Is there another delta here? Besides 1024 -> 10K? Do we have data to suggest a proper upperbound?

For the proper upperbound, I can't tell you what the concurrent writes are.  Our metrics are at the smallest minutely metrics.  At worst, we log from code ~2200 logs per minute, and at best we write to TCP ~800 logs per minute.  That gives us roughly 7 minutes of backlog, at that highest rate.  It seems to me we probably need another writer thread.

We do mostly write at ~1000 logs per minute.

Either way, it looks like we have roughly a 15% write failure on the logs in general :(

http://prometheus.premainnet.testnet.libra.org:9091/d/4Na_dsIMk/logging?orgId=1&from=now-12h&to=now

common/logger/src/struct_log.rs
```diff
@@ -16,7 +16,7 @@ use std::{
     fmt::Display,
     fs::{File, OpenOptions},
     io,
-    io::Write as IoWrite,
+    io::{Write as IoWrite, Write},
```
bmwill, line 19: Why does `Write` need to be brought in twice?

common/logger/src/struct_log.rs
```diff
@@ -426,108 +426,122 @@ impl FileStructLogThread {
...
 impl TCPStructLog {
-    pub fn start_new(address: String) -> io::Result<Self> {
+    pub fn start_new(endpoint: String) -> io::Result<Self> {
         let (sender, receiver) = mpsc::sync_channel(WRITE_CHANNEL_SIZE);
-        let sink_thread = TCPStructLogThread::new(receiver, address);
+        let sink_thread = TCPStructLogThread::new(receiver, endpoint);
         thread::spawn(move || sink_thread.run());
         Ok(Self { sender })
     }
 }
```
bmwill, line 439: This may be a naive question but is there a particular reason why we can't write directly to stdout/stderr and instead just have stdout/stderr itself somehow redirected to where ever it needs to go? In other words make this problem, not our software's issue but some other piece of infra's problem? Maybe this sort of goes back to writing to a file instead...

common/logger/src/struct_log.rs
gregnazario: hmm woops

common/logger/src/struct_log.rs
gregnazario: That does go back to the writing to a file instead.  The TCP log writing is only written specifically to ship the logs to a different container, which is an infrastructure problem.

We could write to stdout, and somehow take kubernetes logs and spew them into elasticsearch with either Logstash or Vector.

davidiw:
> > Logic is a little interesting in that we block on the last message until we get connected. It might be good to document the behavior of the TcpWriter in a doc block.
> 
> The other option is to have two queues, a resubmission queue, and a sending queue. Then we could possibly have two threads writing so we always get the latest logs, as well as not drop logs. This didn't happen that often, but when we did disconnect, the old logic would just straight up drop the logs.
> 
> > Seems this is mostly a refactor with one additional condition, we block logging until we have a connection.
> 
> It is mostly a refactor with one additional condition. It was refactored this way so that:
> 
> 1. It's logically simpler and less confusing
> 2. It allows for usage with other logging frameworks such as tracing if that is a path we go down
> 
> This is a quick fix in general, the reason we now parse the JSON before writing to the Sender rather than after is to keep the storage to a maximum fixed size. Before, we'd have messages that could be very large in the sender.
> 
> > If we don't get a connection, we'll keep around our oldest logs until we do. Interesting. Is there another delta here? Besides 1024 -> 10K? Do we have data to suggest a proper upperbound?
> 
> For the proper upperbound, I can't tell you what the concurrent writes are. Our metrics are at the smallest minutely metrics. At worst, we log from code ~2200 logs per minute, and at best we write to TCP ~800 logs per minute. That gives us roughly 7 minutes of backlog, at that highest rate. It seems to me we probably need another writer thread.
> 
> We do mostly write at ~1000 logs per minute. But our average TCP log write is ~200 or less logs per minute
> 
> Either way, it looks like we have roughly a 15% write failure on the logs in general :(
> 
> http://prometheus.premainnet.testnet.libra.org:9091/d/4Na_dsIMk/logging?orgId=1&from=now-12h&to=now

I would like to believe that we aren't bottlenecked on local processing and that a single thread should be able to easily saturate the logging network thread. At any rate, my only goal was to comment on the potential benefit of adding a doc block that describes the intention of the TcpWriter so other folks don't think they spot a bug when it is actually part of the implementation.

davidiw:
Looks good!

gregnazario:
> > > Logic is a little interesting in that we block on the last message until we get connected. It might be good to document the behavior of the TcpWriter in a doc block.
> > 
> > 
> > The other option is to have two queues, a resubmission queue, and a sending queue. Then we could possibly have two threads writing so we always get the latest logs, as well as not drop logs. This didn't happen that often, but when we did disconnect, the old logic would just straight up drop the logs.
> > > Seems this is mostly a refactor with one additional condition, we block logging until we have a connection.
> > 
> > 
> > It is mostly a refactor with one additional condition. It was refactored this way so that:
> > 
> > 1. It's logically simpler and less confusing
> > 2. It allows for usage with other logging frameworks such as tracing if that is a path we go down
> > 
> > This is a quick fix in general, the reason we now parse the JSON before writing to the Sender rather than after is to keep the storage to a maximum fixed size. Before, we'd have messages that could be very large in the sender.
> > > If we don't get a connection, we'll keep around our oldest logs until we do. Interesting. Is there another delta here? Besides 1024 -> 10K? Do we have data to suggest a proper upperbound?
> > 
> > 
> > For the proper upperbound, I can't tell you what the concurrent writes are. Our metrics are at the smallest minutely metrics. At worst, we log from code ~2200 logs per minute, and at best we write to TCP ~800 logs per minute. That gives us roughly 7 minutes of backlog, at that highest rate. It seems to me we probably need another writer thread.
> > We do mostly write at ~1000 logs per minute. But our average TCP log write is ~200 or less logs per minute
> > Either way, it looks like we have roughly a 15% write failure on the logs in general :(
> > http://prometheus.premainnet.testnet.libra.org:9091/d/4Na_dsIMk/logging?orgId=1&from=now-12h&to=now
> 
> I would like to believe that we aren't bottlenecked on local processing and that a single thread should be able to easily saturate the logging network thread. At any rate, my only goal was to comment on the potential benefit of adding a doc block that describes the intention of the TcpWriter so other folks don't think they spot a bug when it is actually part of the implementation.

I hope this helps significantly and it won't be too much of a problem.  Been taking a 2 pronged approach of reducing unnecessary log noise, as well as increasing the capacity here

PR #5682, opened 2020-08-19
[Move transaction builders] add support for Go

matbd:
## Motivation

Support Move encoders and decoders in Go

## Test Plan

included unit test

## Related PRs

https://github.com/facebookincubator/serde-reflection/issues/37

PR #5687, opened 2020-08-20
[serde generate](nit) remove useless `&` in demo code in Go

matbd:
## Motivation

Remove leftover `&` after https://github.com/facebookincubator/serde-reflection/commit/89c5f4eb5b59ec5d028475771475bd54b786f6f2

## Test Plan

```
cargo x test -p transaction-builder-generator golang -- --ignored
```

## Related PRs

#5682

### DIFF
--- a/language/transaction-builder/generator/examples/golang/stdlib_demo.go
+++ b/language/transaction-builder/generator/examples/golang/stdlib_demo.go
@@ -11,7 +11,7 @@ import (
 )
 
 func main() {
-	token := &libra.TypeTag__Struct{
+	token := libra.TypeTag__Struct{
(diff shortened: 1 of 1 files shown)


language/transaction-builder/generator/examples/golang/stdlib_demo.go
```diff
@@ -11,7 +11,7 @@ import (
 )
 
 func main() {
-	token := &libra.TypeTag__Struct{
+	token := libra.TypeTag__Struct{
```
xli, line 14: as libra.TypeTag is an interface, we should pass in as ptr instead of value, so I suggest keep original &xxx version

PR #5689, opened 2020-08-20
[network] Random hardening

davidiw:
* Add observability into noise handshake
* Remove unnecessary nesting in peer_manager
* Remove unnecessary update to peer id in peer_manager

### DIFF
--- a/config/src/config/network_config.rs
+++ b/config/src/config/network_config.rs
@@ -150,17 +150,11 @@ impl NetworkConfig {
             Identity::None => {
                 let mut rng = StdRng::from_seed(OsRng.gen());
                 let key = x25519::PrivateKey::generate(&mut rng);
-                let peer_id = AuthenticationKey::try_from(key.public_key().as_slice())
-                    .unwrap()
-                    .derived_address();
+                let peer_id = PeerId::from_identity_public_key(key.public_key());
                 self.identity = Identity::from_config(key, peer_id);
             }
             Identity::FromConfig(config) => {
-                let pubkey = config.key.public_key();
-                let peer_id = AuthenticationKey::try_from(pubkey.as_slice())
-                    .unwrap()
-                    .derived_address();
-
+                let peer_id = PeerId::from_identity_public_key(config.key.public_key());
                 if config.peer_id == PeerId::ZERO {
                     config.peer_id = peer_id;
                 }
--- a/config/src/network_id.rs
+++ b/config/src/network_id.rs
@@ -55,9 +55,18 @@ impl NetworkContext {
     }
 
     #[cfg(any(test, feature = "testing", feature = "fuzzing"))]
+    pub fn mock_with_peer_id(peer_id: PeerId) -> std::sync::Arc<Self> {
+        std::sync::Arc::new(Self {
+            network_id: NetworkId::Validator,
+            role: RoleType::Validator,
+            peer_id,
+        })
+    }
+
+    #[cfg(any(test, feature = "testing", feature = "fuzzing"))]
     pub fn mock() -> std::sync::Arc<Self> {
         std::sync::Arc::new(Self {
-            network_id: NetworkId::default(),
+            network_id: NetworkId::Validator,
             role: RoleType::Validator,
             peer_id: PeerId::random(),
         })
--- a/network/src/noise/handshake.rs
+++ b/network/src/noise/handshake.rs
@@ -10,14 +10,20 @@
 //!
 //! [stream]: network::noise::stream
 
-use crate::noise::stream::NoiseStream;
+use crate::{
+    logging::{network_events, network_log},
+    noise::stream::NoiseStream,
+};
 use futures::io::{AsyncRead, AsyncReadExt, AsyncWrite, AsyncWriteExt};
+use libra_config::network_id::NetworkContext;
 use libra_crypto::{noise, x25519};
+use libra_logger::sl_debug;
 use libra_types::PeerId;
 use netcore::transport::ConnectionOrigin;
 use std::{
     collections::{HashMap, HashSet},
     convert::TryFrom as _,
+    fmt::Debug,
     io,
     sync::{Arc, RwLock},
     time,
@@ -135,8 +141,8 @@ impl HandshakeAuthMode {
 
 /// The Noise configuration to be used to perform a protocol upgrade on an underlying socket.
 pub struct NoiseUpgrader {
-    /// The validator's own peer id.
-    self_peer_id: PeerId,
+    /// The validator's network context
+    network_context: Arc<NetworkContext>,
     /// Config for executing Noise handshakes. Includes our static private key.
     noise_config: noise::NoiseConfig,
     /// Handshake authentication can be either mutual or server-only authentication.
@@ -145,9 +151,13 @@ pub struct NoiseUpgrader {
 
 impl NoiseUpgrader {
     /// Create a new NoiseConfig with the provided keypair and authentication mode.
-    pub fn new(peer_id: PeerId, key: x25519::PrivateKey, auth_mode: HandshakeAuthMode) -> Self {
+    pub fn new(
+        network_context: Arc<NetworkContext>,
+        key: x25519::PrivateKey,
+        auth_mode: HandshakeAuthMode,
+    ) -> Self {
         Self {
-            self_peer_id: peer_id,
+            network_context,
             noise_config: noise::NoiseConfig::new(key),
             auth_mode,
         }
@@ -165,7 +175,7 @@ impl NoiseUpgrader {
         remote_public_key: Option<x25519::PublicKey>,
     ) -> io::Result<(x25519::PublicKey, NoiseStream<TSocket>)>
     where
-        TSocket: AsyncRead + AsyncWrite + Unpin,
+        TSocket: AsyncRead + AsyncWrite + Debug + Unpin,
     {
         // perform the noise handshake
         let socket = match origin {
@@ -218,14 +228,14 @@ impl NoiseUpgrader {
         time_provider: F,
     ) -> io::Result<NoiseStream<TSocket>>
     where
-        TSocket: AsyncRead + AsyncWrite + Unpin,
+        TSocket: AsyncRead + AsyncWrite + Debug + Unpin,
         F: Fn() -> [u8; AntiReplayTimestamps::TIMESTAMP_SIZE],
     {
         // buffer to hold prologue + first noise handshake message
         let mut client_message = [0; Self::CLIENT_MESSAGE_SIZE];
 
         // craft prologue = self_peer_id | expected_public_key
-        client_message[..PeerId::LENGTH].copy_from_slice(self.self_peer_id.as_ref());
+        client_message[..PeerId::LENGTH].copy_from_slice(self.network_context.peer_id().as_ref());
         client_message[PeerId::LENGTH..Self::PROLOGUE_SIZE]
             .copy_from_slice(remote_public_key.as_slice());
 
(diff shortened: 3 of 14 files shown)


network/src/noise/handshake.rs
```diff
@@ -279,12 +316,21 @@ impl NoiseUpgrader {
...
     {
         // buffer to contain the client first message
         let mut client_message = [0; Self::CLIENT_MESSAGE_SIZE];
 
         // receive the prologue + first noise handshake message
+        sl_info!(
+            network_log(network_events::NOISE_UPGRADE, &self.network_context)
+                .data("origin", "server")
+                .data("step", "reading_data")
+                .log(format!(
```
davidiw, line 329: should log take in a parameterized string like format?

network/src/noise/handshake.rs
```diff
@@ -145,9 +151,13 @@ pub struct NoiseUpgrader {
 
 impl NoiseUpgrader {
     /// Create a new NoiseConfig with the provided keypair and authentication mode.
-    pub fn new(peer_id: PeerId, key: x25519::PrivateKey, auth_mode: HandshakeAuthMode) -> Self {
+    pub fn new(
+        network_context: Arc<NetworkContext>,
+        key: x25519::PrivateKey,
+        auth_mode: HandshakeAuthMode,
```
mimoo, line 157: shouldn't this be part of NetworkContext?

network/src/noise/handshake.rs
```diff
@@ -248,15 +258,42 @@ impl NoiseUpgrader {
             .map_err(|e| io::Error::new(io::ErrorKind::Other, e))?;
 
         // send the first handshake message
+        sl_info!(
+            network_log(network_events::NOISE_UPGRADE, &self.network_context)
+                .data("origin", "client")
```
mimoo, line 263: origin -> role ?

network/src/noise/handshake.rs
```diff
@@ -248,15 +258,42 @@ impl NoiseUpgrader {
...
+                .data("step", "writing_data")
+                .log(format!(
+                    "{} noise client writing_data: {}",
+                    self.network_context, remote_public_key
```
mimoo, line 267: why not client_message?

network/src/noise/handshake.rs
```diff
@@ -248,15 +258,42 @@ impl NoiseUpgrader {
...
         socket.flush().await?;
 
         // receive the server's response (<- e, ee, se)
+        sl_info!(
+            network_log(network_events::NOISE_UPGRADE, &self.network_context)
+                .data("origin", "client")
+                .data("step", "reading_data")
+                .log(format!(
+                    "{} noise client reading_data: {}",
+                    self.network_context, remote_public_key
```
mimoo, line 280: same here, why not log the actual server_response after reading it? remote_public_key here is the expected remote public key not the one read

network/src/noise/handshake.rs
```diff
@@ -248,15 +258,42 @@ impl NoiseUpgrader {
...
 
         // parse the server's response
         // TODO: security logging here? (mimoo)
+        sl_info!(
+            network_log(network_events::NOISE_UPGRADE, &self.network_context)
+                .data("origin", "client")
+                .data("step", "finalizing")
+                .log(format!(
+                    "{} noise client finalize: {}",
+                    self.network_context, remote_public_key
```
mimoo, line 294: same

network/src/noise/handshake.rs
mimoo: why log though? Why not `data_display()`?

config/src/config/network_config.rs
```diff
@@ -150,17 +150,11 @@ impl NetworkConfig {
             Identity::None => {
                 let mut rng = StdRng::from_seed(OsRng.gen());
                 let key = x25519::PrivateKey::generate(&mut rng);
-                let peer_id = AuthenticationKey::try_from(key.public_key().as_slice())
-                    .unwrap()
-                    .derived_address();
+                let peer_id = PeerId::from_identity_public_key(key.public_key());
```
mimoo, line 153: 👍

config/src/network_id.rs
```diff
@@ -54,6 +54,15 @@ impl NetworkContext {
         self.role
     }
 
+    #[cfg(any(test, feature = "testing", feature = "fuzzing"))]
+    pub fn mock_with_peer_id(peer_id: PeerId) -> std::sync::Arc<Self> {
```
mimoo, line 58: I know it probably doesn't matter but I suggest something more specific:

```
pub fn validator_for_testing(peer_id: PeerId) -> Self {
  Self {
    network_id: NetworkId::Validator,
    ...
  }
}

```

network/src/peer_manager/builder.rs
```diff
@@ -273,35 +273,22 @@ impl PeerManagerBuilder {
...
         let protos = transport_context.supported_protocols();
         let chain_id = transport_context.chain_id;
-        let network_id = self.network_context.network_id().clone();
-        let peer_id = self.network_context.peer_id();
-
-        let (key, maybe_trusted_peers, peer_id) = match transport_context.authentication_mode {
-            // validator-operated full node
-            AuthenticationMode::ServerOnly(key) if peer_id == PeerId::ZERO => {
-                let public_key = key.public_key();
-                let peer_id = PeerId::from_identity_public_key(public_key);
```
mimoo, line 283: aren't you missing this edge case now?

network/src/peer_manager/mod.rs
```diff
@@ -517,7 +517,8 @@ where
                         );
                     }
                 } else {
-                    self.dial_peer(requested_peer_id, addr, response_tx).await;
+                    let request = TransportRequest::DialPeer(requested_peer_id, addr, response_tx);
+                    self.transport_reqs_tx.send(request).await.unwrap();
```
mimoo, line 521: inling code ❤️

network/src/peer_manager/builder.rs
mimoo: the peer LGTM except for this part that seems a bit tricky to me. I don't think we really enforce that your peer id is correctly derived from your pubkey in network where you're not authenticated

network/src/noise/handshake.rs
gregnazario: Yeah for simplicity getting out the door, I didn't.  It is annoying that you have to do `format!` on each one.  Those format strings are mostly handled most easily by macros.

`log` is the replacement for a context text log line like you would use in `info!("Some info: {}", info)`. In this example it is `Some info: info`.  The objective was to provide a way to migrate all logs to structured logs from text logs.

network/src/noise/handshake.rs
```diff
@@ -431,9 +477,27 @@ impl NoiseUpgrader {
             .map_err(|e| io::Error::new(io::ErrorKind::Other, e))?;
 
         // send the response
+        sl_info!(
+            network_log(network_events::NOISE_UPGRADE, &self.network_context)
+                .data("origin", "server")
```
gregnazario, line 482: Do you want to reuse the `ConnectionOrigin::Inbound/Outbound` for this log?

config/src/network_id.rs
davidiw: Hmm... see the mock below, of course apparently I wrote that one too... what better way to have a mock that is ... completely bogus.

network/src/noise/handshake.rs
davidiw: auth_mode? I think a lot could be part of network context... worth reviewing at some point in time what other values we may potentially want... for example, I wondered why chain id wasn't part of it

network/src/noise/handshake.rs
```diff
@@ -248,15 +258,42 @@ impl NoiseUpgrader {
             .map_err(|e| io::Error::new(io::ErrorKind::Other, e))?;
 
         // send the first handshake message
+        sl_info!(
```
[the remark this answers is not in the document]
davidiw: in steady state, this traffic should be really, really low, so I believe info is the correct setting. I compared this to the logging level for other connection establishment logs. I think as long as we have messages along the path, they should be all the same. If we want the enter and exit to be one level higher, that might make sense. In that case, we should modify a bunch in connectivity_manager to be debug and these as well... first stab here though

network/src/noise/handshake.rs
davidiw: I imagine neither are particularly parseable but at remote_public_key is likely more so.

network/src/noise/handshake.rs
davidiw: this is before the server response, we want to make sure we're getting this far.

network/src/noise/handshake.rs
davidiw: this could be the response, but the response isn't in any reasonable format... I believe additional poking in this space could result in higher quality logging messages, but this should be sufficient to make sense of whether or not there are issues in establishing connections.

network/src/noise/handshake.rs
davidiw: @mimoo, @gregnazario 's second paragraph is a response to your question.

network/src/noise/handshake.rs
davidiw: well, I did, but I didn't want to pipe in too much data, I think I'll go with role as @mimoo suggested.

network/src/peer_manager/builder.rs
davidiw: there is really no enforcement, if you look, we only do this if peer_id is zero, if someone tweaks their peer_id, there's not much we can do... the configs already make it impossible to accidentally have a peer_id zero. This is some legacy code when we were trying to figure out should we enforce this in networking or in the configs. I opt for configs since no one should be directly poking at this code.

network/src/peer_manager/mod.rs
davidiw: a lot of opportunity in here :P

I would bet that the above line never occurs, except in random conditions since the check is /the same/ as connection manager.... sigh...

network/src/noise/handshake.rs
mimoo: isn't greg pointing out that you shouldn't use log :D ?

network/src/noise/handshake.rs
mimoo: yeah the auth_mode

mimoo: I'll make a note of that : o compiling some notes about network

network/src/noise/handshake.rs
mimoo: my thinking was: if we want to use this for debugging we might want to print out the actual message we send, not just part of it

network/src/noise/handshake.rs
mimoo: I would suggest modifying this log then to reflect that, right now it sounds like you're reading that key, when that's not what's happening : o maybe "waiting to read server response"

network/src/noise/handshake.rs
davidiw: no, the point is that if you want your log to show up in text logs you need to do both... do I want a second line that is just "info!(...)" that would then route to structured log and then have two structured logs or use the model in the code that will ensure that text and structured are logged only once.

network/src/peer_manager/builder.rs
mimoo: sounds good!

network/src/noise/handshake.rs
davidiw: we are in the process of reading the data :P I think they both mean the same thing... if we wanted to be after the event, we could say "server response: ..."

network/src/noise/handshake.rs
davidiw: sure, but the entirety of the data is actually in this log line... if we want to go the route of the actual message, we should make these proper lcs serializable structs that can then be pretty printed.

PR #5693, opened 2020-08-20
Remove duplicate code for aggregating vote power

jnaulty:
The aggregation of total vote power was implemented slightly differently
in two areas of the `validator_verifier.rs` file.

This commit refactors those implementations into a single function and
adds tests for the case in which a validator increases their voting
power to a value large enough to cause an integer overflow.

## Motivation

Simplify code that aggregates total voting power in `validator_verifier.rs`

Yes

## Test Plan

`cd types; cargo test -- --nocapture test_very_unequal_vote_quorum_validators`

### DIFF
--- a/types/src/validator_verifier.rs
+++ b/types/src/validator_verifier.rs
@@ -8,7 +8,6 @@ use libra_crypto::{
     hash::CryptoHash,
     Signature, VerifyingKey,
 };
-use mirai_annotations::*;
 #[cfg(any(test, feature = "fuzzing"))]
 use proptest_derive::Arbitrary;
 use serde::{Deserialize, Serialize};
@@ -82,10 +81,7 @@ impl ValidatorVerifier {
     pub fn new(
         address_to_validator_info: BTreeMap<AccountAddress, ValidatorConsensusInfo>,
     ) -> Self {
-        let total_voting_power = address_to_validator_info
-            .values()
-            .map(|x| x.voting_power)
-            .sum();
+        let total_voting_power = sum_voting_power(&address_to_validator_info);
         let quorum_voting_power = if address_to_validator_info.is_empty() {
             0
         } else {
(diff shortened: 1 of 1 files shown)


types/src/validator_verifier.rs
```diff
@@ -268,6 +261,17 @@ impl ValidatorVerifier {
     }
 }
 
+/// Returns sum of voting power from Map of validator account addresses, validator consensus info
+pub fn sum_voting_power(
```
mimoo, line 265: remove the `pub` here

types/src/validator_verifier.rs
```diff
@@ -103,11 +100,7 @@ impl ValidatorVerifier {
         address_to_validator_info: BTreeMap<AccountAddress, ValidatorConsensusInfo>,
         quorum_voting_power: u64,
     ) -> Result<Self> {
```
mimoo, line 102: I'd argue that since this is a test we should remove the `Result<>` here

types/src/validator_verifier.rs
```diff
@@ -268,6 +261,17 @@ impl ValidatorVerifier {
     }
 }
 
+/// Returns sum of voting power from Map of validator account addresses, validator consensus info
+pub fn sum_voting_power(
+    address_to_validator_info: &BTreeMap<AccountAddress, ValidatorConsensusInfo>,
+) -> u64 {
+    address_to_validator_info.values().fold(0, |sum, x| {
+        // The voting power of any node is assumed to be small relative to u64::max_value()
+        assume!(sum <= u64::max_value() - x.voting_power);
```
mimoo, line 270: cc @hermanventer 

this assume is not great imo, I think it's fine to crash if we do overwrap, I would change the addition instead here:

```rust
sum.checked_add(x.voting_power).expect("sum of all voting power is greater than u64::max")
```

hermanventer: It is only appropriate to make an assumption when there are compelling (and documented) reasons to believe that the assumption can never be violated. In this case, the caller can supply an argument that might overflow. Hopefully no current caller does so, but a future change might result in such a caller, so a precondition would be better than an assumption.

MIRAI has some heuristics that matter here: If you unwrap and Option with expect, it acts like an assumption. If you use unwrap, it becomes a proof obligation and MIRAI should issue a diagnostic if it cannot prove that the unwrap will never fail.

hermanventer: It is a heuristic to reduce false positives. In this case it is probably better to use unwrap_or_else(|| unrecoverable!("...")).

mimoo: it's starting to be a mouthful :P I'm also not sure I see the point of unrecoverable! since panic, unwrap, except are all functions aimed at handling unrecoverable errors.

hermanventer: You wan't to make it clear that an error is expected but unrecoverable, hence the advice to use unrecoverable! rather than just panic!. Looking at this again, it is perhaps a bad example of a non-recoverable error. This is much more of precondition and when it fails it is a programming mistake.

PR #5697, opened 2020-08-20
[logging] Update logging dashboard to have rate based metrics

gregnazario:
### Overview
Saving the dashboard used in premainnet http://prometheus.premainnet.testnet.libra.org:9091/d/4Na_dsIMk/logging?editview=dashboard_json&orgId=1 as it has become very useful for debugging issues.


terraform/templates/dashboards/logging.json
```diff
@@ -239,6 +1364,6 @@
   },
   "timezone": "",
   "title": "Logging",
-  "uid": "JMhAZhVGk",
-  "version": 3
-}
+  "uid": "4Na_dsIMk",
+  "version": 4
+}
```
bmwill, line 1369: CI might complain about the missing newline here.

gregnazario: it 100% will.  good catch on it.

gregnazario:
> lgtm.
> 
> Quick question: Where do the dashboards live for pre-mainnet, since it seems you're moving a dashboard from there to here?


TBH, I don't actually know how the dashboards are put through to premainnet.  Might need to ask @sherry-x 

In addition, these dashboards break on cluster test, due to differences in the post added dimensions (as I just learned)

PR #5701, opened 2020-08-20
[gha][lbt] don't run LBT compat on cherry-pick and update slack log

rustielin:
## Overview

* Updates how we fetch PR details (number, ref) and reorders some initial steps in the LBT workflow
* Fetch PR base ref for use in compat test. For now, only run compat test if the PR targets `master`
* Slack message shows a dump of the cluster-test report to be more representative of the actual failure since tailing the logs usually doesn't show anything too meaningful
* Bonus: [new button in Slack](https://toroeng.slack.com/archives/C0150QMKLTX/p1598998963002400) to "Visit PR", so we don't have to hunt down the PR number from the job 

## Test

A bunch of canary on this PR and also failure cases on https://github.com/libra/libra/pull/5700

## Note

EDIT:

This PR makes it such that compat test in LBT will only run if the base of the PR is `master`. To get compat to work for other branches, some fixes need to be made to the workflow to ensure that the right images exist to use in test.

### DIFF
--- a/.github/workflows/land-blocking.yml
+++ b/.github/workflows/land-blocking.yml
@@ -16,11 +16,59 @@ jobs:
     timeout-minutes: 50
     steps:
       - uses: actions/checkout@v1
+      - name: Get PR number
+        id: pr-num
+        uses: actions/github-script@v3
+        with:
+          github-token: ${{secrets.GITHUB_TOKEN}}
+          result-encoding: string
+          script: |
+            // Find the number of the pull request that trigggers this push
+            let pr_num = 0;
+            let commit_message = context.payload.head_commit.message;
+            let re = /.*[^]Closes:\s\#(\d+)$/;
+            if (re.test(commit_message)) {
+              let match = re.exec(commit_message);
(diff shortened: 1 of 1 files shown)


andll:
I agree that there is value in running compat test on release branch and we can't assume that 'compat test runs elsewhere so there is no need to worry'.
We can probably land this and then work on activating compat test?
Leaving it to @sausagee to approve since I did not look in detail into the LBT code

rustielin:
@sausagee @andll agreed, thanks for the feedback! I've added the ability for LBT to build two sets of images if it is unable to find the right images to use for compat test (https://github.com/libra/libra/pull/5878 which is stacked on this PR) LBT + compat should now work for cherry-picking into release branches.

sausagee:
> LBT + compat should now work for cherry-picking into release branches.

The PR title reads "[gha][lbt] don't run LBT on cherry-pick and update slack log #5701" ATM, update it to match the intent?

sausagee (approved):
Looks good other the PR title.

PR #5703, opened 2020-08-20
[cluster-test] Increase default cti timeout

andll:
In some rare cases LBT needs to do expensive cleanup from previous job, in this case we need to allow more time for it to complete

### DIFF
--- a/scripts/cti
+++ b/scripts/cti
@@ -13,8 +13,8 @@ ENV=""
 REPORT=""
 LOCAL_BUILD=""
 EXIT_CODE=0
-# Default timeout is 30 mins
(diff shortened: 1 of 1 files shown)


andll:
/priority high
/land

Landing with high priority, because there might be cascade effect where breaking jobs leave running cluster, and next job will fail because it clean up for previous job will take too much time

snakech:
> /priority high
> /land
> 
> Landing with high priority, because there might be cascade effect where breaking jobs leave running cluster, and next job will fail because it clean up for previous job will take too much time

Do you think my pr to cleanup jobs immediately after job done can make this case better?

PR #5704, opened 2020-08-20
[cluster-test] Use exponential retry strategy in kubernetes operations

andll:
This allows for fast operations to complete faster, as initial backoff timeout is small, but for longer operations it just reaches limit(same as fixed interval before).

### DIFF
--- a/common/retrier/src/lib.rs
+++ b/common/retrier/src/lib.rs
@@ -4,7 +4,7 @@
 #![forbid(unsafe_code)]
 
 use libra_logger::debug;
-use std::{future::Future, pin::Pin, thread, time::Duration};
+use std::{cmp::min, future::Future, pin::Pin, thread, time::Duration};
 
 /// Given an operation retries it successfully sleeping everytime it fails
 /// If the operation succeeds before the iterator runs out, it returns success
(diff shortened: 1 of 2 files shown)


common/retrier/src/lib.rs
```diff
@@ -54,11 +55,25 @@ pub fn fixed_retry_strategy(delay_ms: u64, tries: usize) -> impl Iterator<Item =
     FixedDelay::new(delay_ms).take(tries)
 }
 
+pub fn exp_retry_strategy(
```
snakech, line 58: share we re-order these fns to make exp_retry and fixed_retry are grouped by each other?

andll: I am not sure why we prefer this order over some other :)

snakech: hmm, I thought it will be easier to read to group struct + impl + Iterator together but its definitely only personal preference.

andll:
/land

This roughly cats 1 minute from LBT time. This step takes now 1.5 minutes, and was taking 2.5 minutes before this change:

```
INFO 2020-08-20 22:37:54 testsuite/cluster-test/src/cluster_builder.rs:275 Done generating genesis.
INFO 2020-08-20 22:39:32 testsuite/cluster-test/src/cluster_builder.rs:137 Deployed 30 validators and 30 fns
```

common/retrier/src/lib.rs
andll: There are different styles in rust, there is no really guideline for libra as far as I am aware

Arbitrary choice I made here is to group blocks by type, e.g. struct delcaration first, then implementations, then trait implementations

snakech: this makes sense too, I will keep this style consistency in the future changes

PR #5705, opened 2020-08-20
[logging] Use elasticsearch compatible timestamp

gregnazario:
So that Elasticsearch picks up the time of the actual log, we will use
the `@timestamp` field.  This field is overwritten by fluent bit in
logstash compatibility mode, so it has been turned off.


gregnazario:
This doesn't work because fluent bit adds a second `@timestamp`, which makes it invalid JSON...

Issue #5715, opened 2020-08-21
[logging] Remove crit! from codebase


gregnazario:
### Overview
`crit!` provides critical logs, but are being logged as `ERROR` because they don't fit into our logging scheme.  We should remove it from the codebase in favor of `error!`

Remaining remove `crit!` from codebase

```
libra $ grep -r crit! . | sed 's/:.*$//g' | uniq -c | sort -nr
   8 ./language/libra-vm/src/errors.rs
   3 ./language/move-vm/runtime/src/loader.rs
   3 ./language/move-vm/runtime/src/interpreter.rs
   2 ./storage/libradb/src/pruner/mod.rs
   2 ./mempool/src/shared_mempool/coordinator.rs
   1 ./storage/libradb/src/system_store/mod.rs
   1 ./language/libra-vm/src/data_cache.rs
   1 ./execution/executor/src/lib.rs
   1 ./common/crash-handler/src/lib.rs
```

PR #5717, opened 2020-08-21
[logging] Remove crit! in favor of error!

gregnazario:
### Overview
crit! is now deprecated, and is being removed.  Error was the level
everything was being logged at anyways, so it's a simple enough change.

I've also changed some variable names so that the structured logs pick them up as field names

Resolves https://github.com/libra/libra/issues/5715


gregnazario:
I've gone ahead and decreased the amount of changes now that the `name=value` works in logging.

gregnazario:
Strange, I rebased and the test didn't fail locally, so maybe that's why

PR #5719, opened 2020-08-21
[breaking][bugfix][vm] Charge gas for emit_event

tzakian:
This PR has a couple bugfixes related to gas:

1. Emit event wasn't included in both the set of native cost indices, and was having zero gas charged for it (thanks @tnowacki for spotting this!)
2. The default constants in the VM didn't match up with the constants on the Move side that are being published.
3. The number of native functions in the the VM didn't include `emit_event`.
4. Added charging to `FreezeRef`
5. Added charging based on the number of function arguments to `Call`
6. Added code to make sure the size was always non-zero when charging based on size (for both natives and instructions)
7. Added a base charge to an `exists` check if the data isn't found.
8. Update the charging for debug natives to no longer have a zero gas cost
9. Add `debug_assert`s to make sure we are never charging zero gas outside of "system" cost strategies.

This PR also updates some old comments relating to gas both on the Move side and the in the VM.

This PR is breaking since this PR will change the gas usage for transactions that emit events.


dariorussi:
given the breaking nature of this PR, would it make sense to review `Exists` given it charges 0 for non existent resources? also would this be a good time to review whether there are other cases where we charge 0 and we should not? 
`FreezeRef` is an instruction that does not charge at all, and maybe it should at least charge the instruction cost?
Not sure we have other places where we can have a 0 charging...

dariorussi:
@vgao1996, @runtian-zhou any reference to something gas related that you noticed should or could be changed/corrected?

tzakian:
I think that's a good idea @dariorussi, I'll update this with all of those changes as well.

vgao1996:
@dariorussi In addition to what you mentioned, I think we should review whether it is appropriate to charge 0 for the native functions in the Debug module. Although those functions don't do anything meaningful in prod, there's still a cost calling the dispatcher or simply stepping through them in the interpreter.

tzakian:
I'm not sure if I quite follow on the Debug natives: these should only be callable locally, and the modules for calling them and there should be no way of calling them in prod correct? 

Because of that I'm not sure if/when we'd ever hit this linking/calling overhead for these in prod? But I'm not as familiar with this area, so I'm happy to put in a low non-zero gas cost just to be on the safe side if we want.

vgao1996:
Right now the Debug module is included in the prod genesis, but the actual implementation is guarded by a feature flag that's only enabled in e2e tests. This means it's basically a no-op if you call it in prod, but still you can call it and the VM still has to do some minimal work on it.

We made it this way due to these considerations:
1. We kinda don't want these to affect gas usage, which might help make debugging easier and thus the 0 cost.
2. Having the Debug module included means we don't have to manage two versions of genesis, and it also comes with the benefit of allowing modules to be published and used without the Debug statements being explicitly stripped out.

However I'd say both are debatable and there's certainly room for adjustments here. I'd personally welcome discussions on alternative approaches like not registering this native function at all in prod. Anyways this is one of the things we don't fully understand and we'll likely know it better as we go...

language/move-vm/runtime/src/interpreter.rs
```diff
@@ -351,7 +357,13 @@ impl Interpreter {
...
     ) -> PartialVMResult<AbstractMemorySize<GasCarrier>> {
         let gv = Self::load_resource(data_store, addr, ty)?;
-        let mem_size = gv.size();
+        let mem_size = gv.size().map(|mem_size| {
+            if mem_size == 0 {
+                MIN_EXISTS_DATA_SIZE.get()
+            } else {
+                mem_size
+            }
+        });
```
tzakian, line 366: cc @vgao1996 you've put a lot of good thought in here, so want to get your take on this to see if you think it sounds good? Would you prefer/have a different suggestion for this instead of having a base charge?

language/stdlib/modules/LibraVMConfig.move
```diff
@@ -40,19 +40,23 @@ module LibraVMConfig {
         /// The cost per-byte written to storage.
         global_memory_per_byte_write_cost: u64,
 
-        /// We charge one unit of gas per-byte for the first 600 bytes
+        /// The flat minimum amount of gas required for any transaction.
+        /// Charged at the start of execution.
         min_transaction_gas_units: u64,
 
-        /// Any transaction over this size will be charged `INTRINSIC_GAS_PER_BYTE` per byte
+        /// Any transaction over this size will be charged `intrinsic_gas_per_byte` units of gas per byte.
```
bob-wilson, line 47: This comment (also in gas_schedule.rs) still isn't quite right. The new comment for `intrinsic_gas_per_byte` correctly explains how the cutoff works. I don't think it makes sense to repeat the formula in both comments. Can you find a way to say it once? Maybe: "Any transaction over this size will be charged an additional gas fee."?

language/stdlib/modules/LibraVMConfig.move
```diff
@@ -40,19 +40,23 @@ module LibraVMConfig {
...
         large_transaction_cutoff: u64,
 
-        /// The units of gas that should be charged per byte for every transaction.
+        /// The units of gas that to be charged per byte over the `large_transaction_cutoff` in addition to
+        /// `min_transaction_gas_units` for transactions whose size exceeds `large_transaction_cutoff`.
         instrinsic_gas_per_byte: u64,
```
bob-wilson, line 52: I just noticed that this field is misspelled. It should be "intrinsic" not "instrinsic".

tzakian: good spot!

language/move-vm/types/src/gas_schedule.rs
```diff
@@ -93,6 +93,8 @@ impl<'a> CostStrategy<'a> {
         opcode: Opcodes,
         size: AbstractMemorySize<GasCarrier>,
     ) -> PartialVMResult<()> {
+        // Make sure that the size is always non-zero
+        let size = size.map(|x| std::cmp::max(1, x));
```
tnowacki, line 97: Perhaps add a debug assert here if its 0?

language/move-vm/types/src/gas_schedule.rs
```diff
@@ -93,6 +93,8 @@ impl<'a> CostStrategy<'a> {
         opcode: Opcodes,
         size: AbstractMemorySize<GasCarrier>,
     ) -> PartialVMResult<()> {
+        // Make sure that the size is always non-zero
```
tnowacki, line 96: In `deduct_gas` itself, I think we should add some logging/debug_assert for if `amount` is zero. Thoughts?

tzakian: Sadly this will not work as we use the `zero_cost_schedule` in places (genesis, writeset prologue/epilogue, writeset execution, and block prologue) and that will violate that the number of gas units will always be non-zero

tnowacki: maybe we could make 2 API entry points? or would that be too tricky?

tnowacki: Maybe at the very least for `charge_instr`?

language/move-vm/runtime/src/interpreter.rs
vgao1996: Shouldn't we just fix `GlobalValue/GlobalValueImpl::size`?

language/move-vm/types/src/gas_schedule.rs
tzakian: I think making two API entrypoints doesn't work really here: it's running normal Move code through the interpreter but with a zero cost schedule (i.e., the interpreter loop is parametrized in-part by the gas schedule), so in order to do this we would need to trait-ify the gas schedule and have a zero cost implementation of this for that trait which seems overly-complex.

tnowacki: Let's schedule some time to talk about this. I'm a bit confused about the limitations after reading the code

language/move-vm/runtime/src/interpreter.rs
tzakian: I would be fine with doing that. My only reason for not changing it here as opposed to there is that in this context it is a specific case where an "empty value" should have a non-zero size, whereas in `GlobablValue/GlobalValueImpl::size` this is the size of the value in a context-independent setting, so returning a non-zero size for the "empty value" seems incorrect/contradictory. Happy to do it either way, just let me know which you prefer.

tzakian:
@vgao1996 

> Right now the Debug module is included in the prod genesis, but the actual implementation is guarded by a feature flag that's only enabled in e2e tests. This means it's basically a no-op if you call it in prod, but still you can call it and the VM still has to do some minimal work on it.

> We kinda don't want these to affect gas usage, which might help make debugging easier and thus the 0 cost.
Having the Debug module included means we don't have to manage two versions of genesis, and it also comes with the benefit of allowing modules to be published and used without the Debug statements being explicitly stripped out.
However I'd say both are debatable and there's certainly room for adjustments here. I'd personally welcome discussions on alternative approaches like not registering this native function at all in prod. Anyways this is one of the things we don't fully understand and we'll likely know it better as we go...

Ahh I see. In that case I think adding a non-zero gas price makes sense for this until we decide whether or not to remove everything related to debugging in prod. Additionally a zero gas price for this would violate the invariants that are now being checked when charging gas (that the gas amount is non-zero if metering isn't turned off). Does this sound reasonable to you?

If we have time I think it might be worthwhile discussing the debug module/genesis at tomorrow's meeting (or possibly we can just chat if you'd prefer) since I think the inclusion/exclusion of the linking code for the debug module is an interesting question where there isn't a totally clear answer and I'd really like to learn more about it.

language/move-vm/runtime/src/interpreter.rs
vgao1996: That's an interesting point. Although I'd say the empty value isn't really empty -- the new `GlobalValue` is more like a slot, and the empty state still holds a flag indicating whether the value is fresh or deleted. Anyways you certainly understand the gas semantics better than I so I'll trust your judgement on this.

vgao1996:
> adding a non-zero gas price makes sense for this
Sounds reasonable to me.

Yeah feel free to chat with me or bering it up in the meeting!

language/move-vm/runtime/src/interpreter.rs
tzakian: ahh very good point, I hadn't realized that about `GlobalValue`. I think I'll go with your recommendation here and move this in to the `size` impl for `GlobalValue/GlobalValueImpl` since that will much cleaner as well.

PR #5720, opened 2020-08-21
[network][handshake protocol] refactoring and implementing test/fuzzing

mimoo:
This PR does a number of non-logical but architectural refactoring:

* `transports.rs` had a lot of tests, so moving them out in their own file (and removing a TODO from @phlip9)
* move `HandshakeMsg` out of `UpgradeContext`. The rationale is that `UpgradeContext` should be clear and provide information about upgrading, where `HandshakeMsg` is an struct that is relevant only in the wire logic. We're losing an optimization that does not matter much (as we're not going to have too many connection establishment) and that we could regain elsewhere: if we want to optimize this flow we should have a `HandshakeContext` instead that: 1) can be stored in `NetworkContext` and 2) drives the handshake protocol and 3) caches the **serialized** `HandshakeMsg`.
* adding a test to the handshake protocol to make sure that chain id mismatch fails

In addition, it adds two fuzzers for the libranet handshake protocol:

* one that fuzzes the exchange of HandshakeMsg and the negotiation
* one that fuzzes only the negotiation to skip the LCS deserialization and other stuff

To do this it also implements proptest strategies on a few types

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2401,6 +2401,7 @@ dependencies = [
  "libra-proptest-helpers 0.1.0",
  "libra-workspace-hack 0.1.0",
  "proptest 0.10.1 (registry+https://github.com/rust-lang/crates.io-index)",
+ "proptest-derive 0.2.0 (registry+https://github.com/rust-lang/crates.io-index)",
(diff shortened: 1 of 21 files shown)


mimoo:
see for a better committed version: https://github.com/libra/libra/pull/5776/commits

PR #5722, opened 2020-08-21
[logging] Retry to send failed sent logs

gregnazario:
### Overview
Logs that were being cut off in the middle of sending would not be
retried.  Due to not having deduping of messages before, this was
skipped, but since each message has a unique ID, we can double send a
message and have no issues.

Shouldn't go out really without this change: https://github.com/novifinancial/libra-ops/pull/704

### DIFF
--- a/common/logger/src/struct_log.rs
+++ b/common/logger/src/struct_log.rs
@@ -38,7 +38,6 @@ static NOP: NopStructLog = NopStructLog {};
 static mut STRUCT_LOGGER: &'static dyn StructLogSink = &NOP;
 static STRUCT_LOGGER_STATE: AtomicUsize = AtomicUsize::new(UNINITIALIZED);
 const UNINITIALIZED: usize = 0;
-
 const INITIALIZING: usize = 1;
 const INITIALIZED: usize = 2;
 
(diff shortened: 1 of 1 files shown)


common/logger/src/struct_log.rs
```diff
@@ -38,7 +38,6 @@ static NOP: NopStructLog = NopStructLog {};
 static mut STRUCT_LOGGER: &'static dyn StructLogSink = &NOP;
 static STRUCT_LOGGER_STATE: AtomicUsize = AtomicUsize::new(UNINITIALIZED);
 const UNINITIALIZED: usize = 0;
```
davidiw, line 40: why not use an enum?

common/logger/src/struct_log.rs
```diff
@@ -472,9 +472,19 @@ impl TCPStructLogThread {
     /// Writes a log line into json_lines logstash format, which has a newline at the end
     fn write_log_line(stream: &mut TcpWriter, message: String) -> io::Result<()> {
         let message = message + "\n";
+        let bytes = message.as_bytes();
+
+        // Attempt to write the log up to NUM_SEND_RETRIES + 1, and then drop it
```
davidiw, line 477: might be good to state that write_all will acquire a new TcpStream on write error.

common/logger/src/struct_log.rs
gregnazario: Future opportunity :D

common/logger/src/struct_log.rs
gregnazario: True, might be good to point out.  TcpWriter was supposed to be an abstraction so you don't have to worry about it, but i'll mention it here.

common/logger/src/struct_log.rs
bmwill: Specifically because this is state that needs to fit inside an AtomicUsize, based on my reading of the code....

But yeah Greg and I are going to come up with a plan to cleanup lots of this over the coming weeks.

gregnazario: Ah, right I believe this code is [shamelessly copied from Rust Log](https://github.com/rust-lang/log/blob/b77ccefd7e09fa8e6811b7b02e519e566de4837f/src/lib.rs#L300-L311)

PR #5727, opened 2020-08-22
[backup] transaction restore: not try to write LedgerInfo from the pr…

msmouse:
…oof to DB.

## Motivation

It was a mistake to write down the LedgerInfo from the chunk proof, because the version of it is most likely larger than the last transaction in the chunk, corrupting the DB (Db sees the version on the latest LedgerInfo as the "committed version", while we don't have every transaction until then.)

Y
## Test Plan

UT


msmouse:
> didn't have test coverage in the first place.
✝

PR #5730, opened 2020-08-22
[vault] Improve serialisation of Vault secrets

mgorven:
## Motivation

* Change `Value` to use externally tagged serialisation which looks like
  `{"<type>": "<value>"}` instead of `{"type": "<type>", "value": "<value>"}`
* Use the serialised `Value` directly as the payload to Vault instead of
  wrapping within a key.
* Change the `read_secret()/write_secret()` interface to pass a
  `serde_json::Value` since we now require JSON here.

This results in the data in Vault looking like this:

    Key       Value
    ---       -----
    string    0:b407744436fb8b972953e78b2c94279a4380028d830c5cfa686f8c1b1a93130d

    Key            Value
    ---            -----
    safety_data    map[epoch:1 last_vote:<nil> last_voted_round:38177 preferred_round:38175]

This makes it easier for operators to inspect the data in Vault, and the initialisation data in Terraform simpler.

## Test Plan

Deployed to testnet.

### DIFF
--- a/secure/storage/src/value.rs
+++ b/secure/storage/src/value.rs
@@ -11,7 +11,7 @@ use libra_types::transaction::Transaction;
 use serde::{Deserialize, Serialize};
 
 #[derive(Debug, Deserialize, PartialEq, Serialize)]
-#[serde(content = "value", rename_all = "snake_case", tag = "type")]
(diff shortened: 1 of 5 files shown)


mgorven:
#5734 addresses this in a more general way.

PR #5731, opened 2020-08-22
[network] [breaking] enable onchain and encryption and vector of addresses

davidiw:
## Breaking Nature

* Add a new field to validator secure storage validator_network_address_keys => bytes for tracking validator address encryption keys
* change the actual data for on chain addresses to be an lcs serialized vector of (encrypted) raw network address
* change the libra types validator config to return those bytes instead of the richer type
* rename full_node to fullnode in move modules and scripts to be consistent with naming
* rename *_network_address to *_network_addresses since they contain more than one address
* fixed a typo fullnodes to fullnode in a move script
* remove explicit identity keys from the libra types and onchain validator config
* removed the broken add_validator function from our cli tool

## Justification

The networking component was required for upgradeability and important for removing unused and potentially dangerous fields (identity_keys) from move.

The refactoring on the naming makes the logic / api easier to reason about especially since it is an opaque set of bytes.

## Details

* Add a utility to leverage storage to decrypt and encrypt network addresses using the keys therein
* Leverage utility in genesis tool and simple-on-chain encryption
* Remove legacy fields from validator info
* Make validator info fullnode and validator addresses vectors instead of a single one -- this only affects interpretation of the move values and not move itself
* Rename fields to be plural and make full_node consistently fullnode
* add_validator in cli is broken and has been broken... it isn't tested and can be brought back if someone wants to test and fix it

## Note

* For review simplicity, I have not stacked this on https://github.com/libra/libra/pull/5734 the but I have already rebased it and intend to land it after: https://github.com/davidiw/libra/tree/network-enc_storage
* There are a lot of quirks in handling multiple addresses, this PR doesn't really address much of that from the operational side as all the tooling expects 1 address and users themselves aren't likely ready to maintain multiple addresses


language/stdlib/modules/ValidatorConfig.move
```diff
@@ -13,13 +13,8 @@ module ValidatorConfig {
...
     struct Config {
         consensus_pubkey: vector<u8>,
-        /// TODO(philiphayes): restructure
-        ///   3) remove validator_network_identity_pubkey
-        ///   4) remove full_node_network_identity_pubkey
-        validator_network_identity_pubkey: vector<u8>,
-        validator_network_address: vector<u8>,
-        full_node_network_identity_pubkey: vector<u8>,
-        full_node_network_address: vector<u8>,
+        validator_network_addresses: vector<u8>,
```
sblackshear, line 16: Would you mind adding some Move tests that set `validator_network_addresses` and `fullnode_network_addresses` to non-empty garbage that won't parse as a list of (encrypted or non-encrypted) network addresses? Thinking about that nasty crash we had [here](https://github.com/libra/libra/blob/master/execution/executor/src/lib.rs#L353) and want to make sure we cover that case...

config/management/network-address-encryption/src/lib.rs
```diff
@@ -0,0 +1,184 @@
...
+        network_address: &NetworkAddress,
+        account: AccountAddress,
+        seq_num: u64,
+    ) -> Result<Vec<u8>, Error> {
+        let keys = self.read()?;
+        let key = keys
+            .keys
+            .get(&keys.current)
+            .ok_or_else(|| Error::VersionNotFound(keys.current))?;
+        let raw_addr = std::convert::TryFrom::<&NetworkAddress>::try_from(network_address)?;
```
zekun000: into

config/src/config/test_data/validator.yaml
```diff
@@ -49,3 +49,9 @@ validator_network:
                 from_disk: "/full/path/to/token"
     mutual_authentication: true
     max_frame_size: 8388608 # 8 MiB
+    network_address_key_backend:
```
zekun000, line 52: there's 4 identical backend in this file, should we consolidate them in a single backend config per validator?

config/management/network-address-encryption/src/lib.rs
```diff
@@ -89,30 +89,39 @@ impl Encryptor {
...
-        // TODO(davidiw) soon we'll take in a set, for now we assume there's only one addr
-        let encrypted_key =
-            EncNetworkAddress::encrypt(raw_addr, key, keys.current, &account, seq_num, 0);
-        lcs::to_bytes(&encrypted_key).map_err(|e| e.into())
+        let mut enc_addrs = Vec::new();
+        for (idx, addr) in network_addresses.iter().enumerate() {
+            let raw_addr = std::convert::TryFrom::<&NetworkAddress>::try_from(addr)?;
+            let idxu32 = idx as u32;
+            let enc_addr =
+                EncNetworkAddress::encrypt(raw_addr, key, keys.current, &account, seq_num, idxu32);
```
zekun000, line 97: why not idx as u32

config/management/operational/src/validator_config.rs
```diff
@@ -266,17 +261,18 @@ impl DecryptedValidatorConfig {
...
-            .decrypt(config.validator_network_address.as_ref(), account_address)
+        let validator_network_addresses = encryptor
+            .decrypt(&config.validator_network_addresses, account_address)
             .map_err(|e| Error::NetworkAddressDecodeError(e.to_string()))?;
 
         Ok(DecryptedValidatorConfig {
             consensus_public_key: config.consensus_public_key.clone(),
-            full_node_network_address,
-            validator_network_address,
+            full_node_network_address: full_node_network_addresses[0].clone(),
```
zekun000: why not the vector?

config/management/src/validator_config.rs
```diff
@@ -95,8 +95,8 @@ impl ValidatorConfig {
         let validator_config_script = transaction_callback(
             owner_account,
             consensus_key.to_bytes().to_vec(),
-            raw_enc_validator_address,
-            raw_fullnode_address.into(),
+            validator_addresses,
+            lcs::to_bytes(&vec![raw_fullnode_address]).unwrap(),
```
zekun000, line 99: s/unwrap/map_err ?

types/src/validator_config.rs
```diff
@@ -35,20 +36,42 @@ impl MoveResource for ValidatorOperatorConfigResource {
 #[cfg_attr(any(test, feature = "fuzzing"), derive(Arbitrary))]
 pub struct ValidatorConfig {
     pub consensus_public_key: Ed25519PublicKey,
-    pub validator_network_address: RawEncNetworkAddress,
-    pub full_node_network_address: RawNetworkAddress,
+    /// This is an lcs serialized Vec<RawEncNetworkAddress>
+    pub validator_network_addresses: Vec<u8>,
```
zekun000, line 40: why not use Vec\<RawEncNetworkAddress\> directly?

types/src/validator_config.rs
```diff
@@ -35,20 +36,42 @@ impl MoveResource for ValidatorOperatorConfigResource {
...
-            full_node_network_address,
+            validator_network_addresses,
+            full_node_network_addresses,
         }
     }
+
+    /// Returns a filtered list of correct NetworkAddresses within the full_node_network_addresses
+    pub fn full_node_network_addresses<'a>(
+        &self,
+        err_callback: Option<Box<dyn Fn(String) + 'a>>,
```
zekun000, line 61: the callback seems anti-pattern to me, why not just map_err in the caller? (also why not deserialize to Vec\<RawNetworkAddress\> directly)

config/management/network-address-encryption/src/lib.rs
```diff
@@ -0,0 +1,184 @@
...
+        let data = lcs::to_bytes(keys)?;
+        self.storage
+            .set(VALIDATOR_NETWORK_ADDRESS_KEYS, Value::Bytes(data))
+            .map_err(|e| e.into())
+    }
+}
+
+#[derive(Debug, Deserialize, Serialize)]
+struct ValidatorKeys {
+    keys: HashMap<KeyVersion, Key>,
```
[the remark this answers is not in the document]
davidiw: @phlip9 and I discussed this quite a bit.... and the reality is you just said the right word... "Vault". That's a specific implementation, we have an abstraction -- which is just value. We would need to ensure that the implementation supports inserting the tuple (version, value), which is probably not very common.

config/management/network-address-encryption/src/lib.rs
davidiw: interesting that this is outdated: O, but I avoided pulling in the use std::... :P

config/src/config/test_data/validator.yaml
davidiw: There's a note somewhere to explore that... the problem is the routing of that information since we would ideally like the NetworkConfig struct to be independent of other Config structs.

config/management/network-address-encryption/src/lib.rs
davidiw: it becomes > 100 characters resulting in a lot of new lines

config/management/operational/src/validator_config.rs
davidiw: I should probably litter the tooling with note or todos, but it doesn't support multiple addresses and is definitely outside of scope for this version.

config/management/src/validator_config.rs
davidiw: lcs to bytes... this is pretty unfallible

types/src/validator_config.rs
davidiw: what if I want to skip some because they aren't well formed? how do I note that?

zekun000: we probably should validate all on-chain resources are well-formed, so this can be deserialized directly without bothering

language/stdlib/modules/ValidatorConfig.move
davidiw: Spoke offline, as this is just a raw move value, the test is redundant. @zekun000 had a proposal to make this richer -- if we do -- such tests are critical.

At the same time, I think we should investigate a thoroughness check across all move resource / on chain configs in Rust and ensure that nothing nasty happens upon deserialization and / or that the input and output are symmetric.

types/src/validator_config.rs
davidiw: that would be beautiful ... I'll improving this interface in a future commit -- as that would not be breaking.

davidiw:
as  this has been approved, I have now stacked this on the storage PR and will rebase again after that lands ... and then land this one.

davidiw:
let's see if CT likes this... if not, I'm going to have to take care of an earlier than expected refactoring of genesis building :/ :/

davidiw:
yay, it all worked and then flaky...

PR #5732, opened 2020-08-22
[Move Prover] Optimizations for global invariants/removal of some timouts

wrwg:
This PR optimizes global invariants by skipping verification on updates if the memory used by the invariant is not part of the VC (i.e. is not from any modules on the command line, but only from dependencies).

This brings back some advantages of module invariants. Consider the case of a module invariant in M which only refers to memory of module M. This invariant is only included in verification if M itself is verified. This is sound because of data isolation in Move, as only code in M can violate this invariant. Notice that the invariant is still assumed outside of M, to express constraints over its memory.

Now consider the generalized case of a module invariant which touches memory from M1 and M2. It is included in verification if either M1 is verified, M2 is verified, or both. 

Notice that the problem of modular verification and global invariants does not go away with this approach: if M1 uses M2 and imposes new constraints on memory of M2, we would still need to verify M1 and M2 together (multiple module arguments to MVP). However, the case is rare and does not currently happen in the framework AFAICT. It is generally safe to use another module's memory in the premise of an implication. For instance, we frequently have invariants of the form `LibraTimestamp::is_operating() => P`. This invariant does not impose new constraints on `LibraTimestamp` because it is referred to in a premise. 

The new compilation schema removes, however, once instance of where verification happened by 'accident' before. Suppose M1 adds constraints to M2 and calls a function M2::f. Now we verify M1 standalone, without M2. Before this PR, if this function updates by luck memory in M2, the additional constraints from M1 are checked. Not so luckily if M::f does not update memory. This source of inconsistency is removed with this PR; if M1 uses M2 and imposes new constraints on M2, those need to be always verified together.

In addition, this PR renames the property `[on_update]` into `[isolated]` to better express its intention. An invariant marked as isolated does not contribute to the verification context of other conditions. It is only assumed if the underlying memory is actually updated. It is assumed before the update so that the prover can use the pre-state to maintain the invariant for the post-state.

This leads to some speed improvements and removal of a number of timeouts. There have been also functions forgotten to reactivate for verification in an older PR, specifically in LibraAccount. This is fixed with this PR as well.

## Motivation

Performance, Semantics of global invariants

Yes

## Test Plan

Existing tests

## Related PRs

NA

### DIFF
--- a/language/move-prover/src/cli.rs
+++ b/language/move-prover/src/cli.rs
@@ -114,8 +114,13 @@ pub struct ProverOptions {
     pub verify_scope: VerificationScope,
     /// [deprecated] Whether to emit global axiom that resources are well-formed.
     pub resource_wellformed_axiom: bool,
-    /// Whether to assume wellformedness when elements are read from memory.
+    /// Whether to assume wellformedness when elements are read from memory, instead of on
+    /// function entry.
     pub assume_wellformed_on_access: bool,
+    /// Whether to assume a global invariant when the related memory
+    /// is accessed, instead of on function entry. This is currently known to be slower
+    /// if one than off, so off by default.
+    pub assume_invariant_on_access: bool,
     /// Whether to automatically debug trace values of specification expression leafs.
     pub debug_trace: bool,
     /// Report warnings. This is not on by default. We may turn it on if the warnings
@@ -136,6 +141,7 @@ impl Default for ProverOptions {
             assume_wellformed_on_access: false,
             debug_trace: false,
             report_warnings: false,
+            assume_invariant_on_access: false,
         }
     }
 }
@@ -209,7 +215,7 @@ impl Default for BackendOptions {
             func_inline: "{:inline}".to_owned(),
             serialize_bound: 4,
             vector_using_sequences: false,
-            random_seed: 0,
+            random_seed: 1,
             proc_cores: 1,
             vc_timeout: 40,
             keep_artifacts: false,
--- a/language/move-prover/src/spec_translator.rs
+++ b/language/move-prover/src/spec_translator.rs
@@ -33,8 +33,8 @@ use spec_lang::{
         ConditionInfo, ConditionTag, GlobalEnv, GlobalId, QualifiedId, SpecVarId,
         ABORTS_IF_IS_PARTIAL_PRAGMA, ABORTS_IF_IS_STRICT_PRAGMA, CONDITION_ABORT_ASSERT_PROP,
         CONDITION_ABORT_ASSUME_PROP, CONDITION_ABSTRACT_PROP, CONDITION_CONCRETE_PROP,
-        CONDITION_EXPORT_PROP, CONDITION_INJECTED_PROP, CONDITION_ON_UPDATE_PROP,
-        EXPORT_ENSURES_PRAGMA, OPAQUE_PRAGMA, REQUIRES_IF_ABORTS,
+        CONDITION_EXPORT_PROP, CONDITION_INJECTED_PROP, CONDITION_ISOLATED, EXPORT_ENSURES_PRAGMA,
+        OPAQUE_PRAGMA, REQUIRES_IF_ABORTS,
     },
     symbol::Symbol,
     ty::TypeDisplayContext,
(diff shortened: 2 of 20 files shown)


language/move-prover/src/spec_translator.rs
```diff
@@ -1426,28 +1431,110 @@ impl<'env> SpecTranslator<'env> {
...
+                    self.global_env()
+                        .is_property_true(&inv.properties, CONDITION_ISOLATED)
+                })
+                .unwrap_or(false)
+        };
+        let is_connected = |id: &GlobalId| !is_isolated(id);
+
+        // A predicate which determines whether this invariant depends on any memory which
+        // is part of the verification problem, that is touches a memory which is declared by
+        // a module which is verified.
```
shazqadeer, line 1481: Does "a module which is verified" refer to the modules passed as arguments on the command line to Move Prover?

language/move-prover/src/spec_translator.rs
```diff
@@ -1426,28 +1431,110 @@ impl<'env> SpecTranslator<'env> {
...
+                .map(|inv| {
+                    inv.mem_usage
+                        .iter()
+                        .any(|used| !self.global_env().get_module(used.module_id).is_dependency())
+                })
+                .unwrap_or(true)
+        };
+
+        // Option which determines whether invariants should be assumed at access instead of
+        // on function entry. If invariants or assumed on access they my be included many times.
```
shazqadeer, line 1494: typo, spelling, etc.

language/move-prover/src/spec_translator.rs
```diff
@@ -1426,28 +1431,110 @@ impl<'env> SpecTranslator<'env> {
...
+        // On the other hand, they may be localized to the VC sub-problem at hand.
```
shazqadeer, line 1495: I didn't understand the advantage that is being captured by "they may be localized to the VC sub-problem at hand".  You could either expand on the explanation here (or we could discuss 1-1).

language/move-prover/src/cli.rs
```diff
@@ -209,7 +215,7 @@ impl Default for BackendOptions {
             func_inline: "{:inline}".to_owned(),
             serialize_bound: 4,
             vector_using_sequences: false,
-            random_seed: 0,
+            random_seed: 1,
```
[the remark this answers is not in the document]
wrwg: I want to remove any suspicion that a value of 0 is treated by Z3 as "not set so I generate a seed". We have some funny effects around the default seeding of 0 still leading to lots of non-determinism.

language/move-prover/src/spec_translator.rs
wrwg: Yes

language/move-prover/src/spec_translator.rs
wrwg: Fixing it.

language/move-prover/src/spec_translator.rs
wrwg: The assumption only happens at write, not at read. Without this (introduced 2-3 weeks ago) many things wont terminate.

language/move-prover/src/cli.rs
```diff
@@ -114,8 +114,13 @@ pub struct ProverOptions {
     pub verify_scope: VerificationScope,
     /// [deprecated] Whether to emit global axiom that resources are well-formed.
     pub resource_wellformed_axiom: bool,
-    /// Whether to assume wellformedness when elements are read from memory.
+    /// Whether to assume wellformedness when elements are read from memory, instead of on
+    /// function entry.
     pub assume_wellformed_on_access: bool,
+    /// Whether to assume a global invariant when the related memory
+    /// is accessed, instead of on function entry. This is currently known to be slower
+    /// if one than off, so off by default.
```
shazqadeer, line 122: Some typo in the comment here.

language/move-prover/src/spec_translator.rs
shazqadeer: The name of this new option is assume_invariant_on_access but your comment above indicates that it applies to write accesses but not read accesses.  Might it be better to rename the option?

wrwg: Sorry, misread your comment. This is actually about something else.

This PR brings back an option to assume invariants "lazily" as we did a while ago, instead of eagerly at function entry. I was curious what is actually faster. Postponing invariant assumptions until they are needed may avoid having useless quantifiers which confuse Z3. However, benchmarking did not show an advantage, in fact a slight disadvantage. So this option is off by default, but the code should stay so we can play with this.

PR #5733, opened 2020-08-22
[mempool] rebroadcast unACK'ed txn broadcasts

christinacelee:
## Motivation

Originally, mempool relied on TCP for reliable broadcast delivery/rebroadcasting dropped broadcasts. However, experiments showed that it is possible that some mempool broadcasts get dropped within the node (e.g. due to mempool libra-channel overflow), so we cannot rely on TCP to rebroadcast those internally-dropped messages.

This PR addresses this reliability-related gap in mempool's broadcast strategy by adding active checks for unACK'ed broadcasts and rebroadcasting them if a broadcast does not receive an ACK back for a configurable amount of time. 

Supplementary changes
- made some simplifications to the mempool broadcasting strategy for improved debuggability: now ensure that there is no timeline range overlap b/w broadcasted batches, and we rebroadcast entire batches, not parts of batches. This batch consistency is simpler to maintain and easier to debug. 
- some refactoring of peer manager for better code organization/readability

## Test Plan 

New/updated rebroadcasting tests, existing tests
From locally simulating internally dropped mempool broadcasts on libra-swarm, checked they eventually get rebroadcasted as expected


mempool/src/core_mempool/transaction_store.rs
```diff
@@ -343,27 +341,15 @@ impl TransactionStore {
...
-                    }
-                }
-                None
+    pub(crate) fn timeline_range(&mut self, start_id: u64, end_id: u64) -> Vec<SignedTransaction> {
+        self.timeline_index
+            .timeline_range(start_id, end_id)
+            .iter()
+            .filter_map(|(account, sequence_number)| {
+                self.transactions
+                    .get_mut(account)
```
kphfb, line 350: Why does it need mut?

mempool/src/shared_mempool/coordinator.rs
```diff
@@ -152,7 +152,7 @@ pub(crate) async fn coordinator<V>(
                                             .await;
                                     }
                                     MempoolSyncMsg::BroadcastTransactionsResponse{request_id, retry_txns, backoff} => {
-                                        let ack_timestamp = Instant::now();
+                                        let ack_timestamp = SystemTime::now();
```
kphfb, line 155: I'm curious why this was instant before?

mempool/src/shared_mempool/tasks.rs
```diff
@@ -48,9 +48,11 @@ pub(crate) fn execute_broadcast<V>(
 ) where
     V: TransactionValidation,
 {
-    let next_broadcast_backoff = broadcast_single_peer(peer.clone(), backoff, smp);
+    let peer_manager = &smp.peer_manager.clone();
+    peer_manager.execute_broadcast(peer.clone(), backoff, smp);
 
-    let interval_ms = if next_broadcast_backoff {
+    let schedule_backoff = peer_manager.is_backoff_mode(&peer);
```
kphfb, line 54: Where is this defined?  I"m trying to understand what exactly it's doing but I don't see it in github

mempool/src/core_mempool/transaction_store.rs
christinacelee: yeah this doesn't need to be mut, will update

mempool/src/shared_mempool/coordinator.rs
christinacelee: no strong reason, I think it was because this was mostly used to calculate the duration b/w broadcasts, which is what Instant is useful for. But I think SystemTime might be better since it is more meaningful as a standalone value too and may be more helpful for debugging in the future... lmk if you have other/strong opinions on this!

mempool/src/shared_mempool/peer_manager.rs
```diff
@@ -124,6 +181,161 @@ impl PeerManager {
         self.update_failover();
     }
 
+    /// Returns whether backpressure mode is on for broadcasting to this peer
+    pub fn is_backoff_mode(&self, peer: &PeerNetworkId) -> bool {
```
christinacelee, line 185: @kphfb `is_backoff_mode` is here

mempool/src/shared_mempool/tasks.rs
christinacelee: https://github.com/libra/libra/pull/5733/files#r491995108

mempool/src/core_mempool/transaction_store.rs
```diff
@@ -343,27 +341,15 @@ impl TransactionStore {
...
-                        return Some((timeline_id, txn.txn.clone()));
-                    }
-                }
-                None
+    pub(crate) fn timeline_range(&mut self, start_id: u64, end_id: u64) -> Vec<SignedTransaction> {
+        self.timeline_index
+            .timeline_range(start_id, end_id)
+            .iter()
+            .filter_map(|(account, sequence_number)| {
+                self.transactions
```
kphfb, line 349: Is it possible that there is a race and the transaction is gone from self.transactions here?

kphfb (commented):
It generally looks fine to me, but it's a pretty large diff so I'd prefer to have another set of eyes on it.  Can @davidiw review this also?

mempool/src/shared_mempool/peer_manager.rs
```diff
@@ -124,6 +181,161 @@ impl PeerManager {
...
+        ) {
+            error!(
+                "[shared mempool] error broadcasting transactions to peer {:?}: {}",
+                peer, e
+            );
+        } else {
+            let peer_id = &peer.peer_id().to_string();
+            counters::SHARED_MEMPOOL_BROADCAST_LATENCY
+                .with_label_values(&[peer_id])
+                .observe(start_time.elapsed().as_secs_f64());
```
davidiw, line 316: have you considered using... SHARED_MEMPOOL_BROADCAST_LATENCY::start_timer function on this instead? Also it would seem all this timer is doing is evaluating how quickly you can acquire a lock and generate a message to send, is that correct?

mempool/src/shared_mempool/peer_manager.rs
```diff
@@ -124,6 +181,161 @@ impl PeerManager {
...
+            state.timeline_id = std::cmp::max(state.timeline_id, batch_id.1);
+            // turn off backoff mode after every broadcast
+            state.broadcast_info.backoff_mode = false;
+            state.broadcast_info.sent_batches.insert(
+                batch_id,
+                BatchInfo {
+                    send_time,
+                    ack_status: AckStatus::Pending,
+                },
+            );
```
davidiw, line 334: considering we're still holding the peer lock, should we move the latency metric to after this? or even at the end of notify_subscribers? On the other hand, if we just put it at the beginning of the function using the start_timer, it would actually cover this entire function.

mempool/src/shared_mempool/peer_manager.rs
```diff
@@ -124,6 +181,161 @@ impl PeerManager {
...
+            .lock()
+            .expect("failed to acquire core mempool lock");
+
+        // Sync peer's pending broadcasts with latest mempool state
+        // A pending broadcast might become empty if the corresponding txns were committed through
+        // another peer, so don't track broadcasts for committed txns
+        state.broadcast_info.sent_batches = state
+            .broadcast_info
+            .sent_batches
+            .clone()
```
davidiw, line 245: why are we cloning if we are overwriting?

mempool/src/shared_mempool/peer_manager.rs
```diff
@@ -124,6 +181,161 @@ impl PeerManager {
...
+            .into_iter()
+            .filter(|(id, _batch)| !mempool.timeline_range(id.0, id.1).is_empty())
+            .collect();
```
davidiw, line 248: if you need a clone, I believe you can do a cloned before or after the collect to avoid copy all the batch info

mempool/src/shared_mempool/peer_manager.rs
```diff
@@ -124,6 +181,161 @@ impl PeerManager {
...
+        let rebroadcast = state
+            .broadcast_info
+            .sent_batches
+            .iter()
+            .find(|(_id, batch)| {
+                batch
+                    .send_time
+                    .checked_add(Duration::from_millis(
+                        self.mempool_config.shared_mempool_ack_timeout_ms,
+                    ))
```
davidiw, line 262: why are we doing check_add? if we overflow are we even in a good place any more? probably fine to either ignore the "potential" for an overflow or to just expect here...

davidiw: cc: @mimoo

mempool/src/shared_mempool/peer_manager.rs
```diff
@@ -124,6 +181,161 @@ impl PeerManager {
...
+                    .map_or(false, |deadline| {
+                        SystemTime::now().duration_since(deadline).is_ok()
+                    })
+                    || matches!(batch.ack_status, AckStatus::Retry)
```
davidiw, line 266: it seems a little strange to have two checks here... can we merge them together somehow?

davidiw: guess not really... after reading more code

mempool/src/shared_mempool/peer_manager.rs
```diff
@@ -191,98 +403,74 @@ impl PeerManager {
...
+        if retry_txns.is_empty() {
+            // remove broadcast from pending
+            sync_state.broadcast_info.sent_batches.remove(&batch_id);
+        } else {
+            sync_state
+                .broadcast_info
+                .sent_batches
+                .entry(batch_id)
+                .and_modify(|batch| batch.ack_status = AckStatus::Retry);
+        }
```
davidiw, line 466: rather than have a general batches state with two fields, I would recommend two Batches sets one called waiting and one called retry

mempool/src/shared_mempool/peer_manager.rs
```diff
@@ -26,48 +38,93 @@ pub(crate) type PeerInfo = HashMap<PeerNetworkId, PeerSyncState>;
...
+    fn create_request_id(&self) -> String {
+        format!("{}_{}", self.0, self.1)
+    }
+}
+
+/// Txn broadcast-related info for a given remote peer
+#[derive(Clone)]
+struct BroadcastInfo {
+    // sent broadcasts that have not yet received a success ACK
+    pub sent_batches: BTreeMap<BatchId, BatchInfo>,
```
davidiw, line 79: here, you can eliminate a BatchesInfo and have a BTreeMap<BatchId, SystemTime> for sent and BTreeSet<BatchId> for retry (or even Vecs -- I bet you the Vecs are possibly faster :P).

davidiw (commented):
Overall the PR makes sense. But ideally, I'd like to see a little more refinement.

In terms of the PR layout, your three commits really are all only viable together. Hence it should probably be one commit. Now with that being said, I think this PR should not be one commit as it is:

* A commit for refactoring the BatchId interface
* Another commit for improving the read_timeline interface
* Another PR for refactoring BatchInfo
* Final PR for handling the new retry approach

By the way, do we have other areas in state sync or mempool that do their own serialization? If so, we should really clean that up and I'm happy to help!

mempool/src/shared_mempool/tasks.rs
```diff
@@ -585,10 +458,3 @@ pub(crate) async fn process_config_update<V>(
         .restart(config_update)
         .expect("failed to restart VM validator");
 }
-
-/// creates uniques request id for the batch in the format "{start_id}_{end_id}"
-/// where start is an id in timeline index  that is lower than the first txn in a batch
-/// and end equals to timeline ID of last transaction in a batch
-fn create_request_id(start_timeline_id: u64, end_timeline_id: u64) -> String {
-    format!("{}_{}", start_timeline_id, end_timeline_id)
-}
```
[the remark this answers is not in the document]
christinacelee: so per the spec, the request_id can be any String, it's just this particular implementation chooses to store the timeline ID range into the request_id. This assumption only applies to the broadcasts this node crafts and (correctly) does not make this assumption about incoming broadcasts from other peers - it only sends the ACK back with the whatever request_id String value it was provided. The spec requirement of the request_id is set to a generic string (vs some more structured format like u64, u64) because it should be up to the implementor to choose how to manage their broadcasts, and the concept of timeline ID is not part of the spec, so I hesitate to expose that into the spec.. lmk your thoughts

mempool/src/shared_mempool/peer_manager.rs
christinacelee: we need the extra clone because this whole DS is behind a mutable reference

mempool/src/shared_mempool/peer_manager.rs
christinacelee: i tried this before and just now again, and this doesn't work.. i remember being surprised by this too.

mempool/src/shared_mempool/peer_manager.rs
christinacelee: I have this counter down here instead of using start_timer because I only wanted to record latency of actually executed broadcasts. and yes, this latency metrics is to measure the time to acquire a lock and execute to the network send. If this is too trivial I can remove it, I just have it here as a sanity check :P

christinacelee:
moving to #6331

PR #5734, opened 2020-08-23
[breaking][secure-storage] Remove the explicit Value type

davidiw:
## Breaking Nature

Changes the internal secure storage format for all secure storage services (TCB / Configuration)

## Justification

Every storage place kept a record of type information, which wasn't particularly useful adding both operational and programmability challenges (note the amount of lines of code deleted). We had wanted to do this for a while and Michael even has a PR that does this for Vault alone: https://github.com/libra/libra/pull/5730

## Details

This made using storage a lot more painful that it ought to be as new types need to be manually added when the serializing interface knows how to seamlessly move between types.

In order to expedite this work, I've removed the caching layer and added a little bit of complexity into the namespaced storage -- which shouldn't really matter since it isn't actually used in any production environments.

safety_data    map[epoch:1 last_vote:<nil> last_voted_round:0 preferred_round:0]
waypoint    0:9481f5566b6ff1841e37473cdf58758e8ad54e32ca9819792e951f172fcf40b5
validator_network_address_keys    map[current:0 keys:map[0:KoaIX1FUoOQR+Chu9Tav9rqTYve7PaF4LJ5c8OgYyjk=]]

verified github works: https://github.com/davidiw/secure-store-test/commits/master


zekun000:
finally, this change was suggested when you introduced this type lol

mgorven:
Awesome, that serialisation is much nicer. If you use `serde_json::Value` (like in #5730) to pass the value into the Vault client you can get rid of the double JSON encoding. (I'm happy to tackle that in a separate diff though.)

davidiw:
> Awesome, that serialisation is much nicer. If you use `serde_json::Value` (like in #5730) to pass the value into the Vault client you can get rid of the double JSON encoding. (I'm happy to tackle that in a separate diff though.)

great advice, as you experienced, the fuzzer makes its interesting, I've updated the PR and the encoded values

JoshLind:
Nice to see this finally cleaned up 😄 Looks good from my side.

secure/storage/src/namespaced_storage.rs
```diff
@@ -1,42 +1,44 @@
...
+use std::sync::{Arc, RwLock};
 
 /// This provides a light wrapper around KV storages to support a namespace. That namespace is
 /// effectively prefixing all keys with then namespace value and "/" so a namespace of foo and a
 /// key of bar becomes "foo/bar". Without a namespace, the key would just be "bar". This matches
 /// how this library implements namespaces for Vault.
 pub struct NamespacedStorage {
     namespace: String,
-    inner: Box<dyn KVStorage>,
+    inner: Arc<RwLock<Storage>>,
```
zekun000, line 14: why

config/management/genesis/src/storage_helper.rs
```diff
@@ -58,12 +56,9 @@ impl StorageHelper {
...
         // Initialize all other data in storage
         storage
-            .set(
-                SAFETY_DATA,
-                Value::SafetyData(SafetyData::new(0, 0, 0, None)),
-            )
+            .set(SAFETY_DATA, SafetyData::new(0, 0, 0, None))
             .unwrap();
-        storage.set(WAYPOINT, Value::String("".into())).unwrap();
+        storage.set(WAYPOINT, Waypoint::default()).unwrap();
```
gregnazario, line 61: That was weird `"".into()`

secure/storage/src/namespaced_storage.rs
davidiw: Oh you know, I think I can use a Box<Storage>... I think I made this to convoluted, good call out.

testsuite/tests/libratest/smoke_test.rs
```diff
@@ -1996,11 +1986,10 @@ fn test_replay_tooling() {
...
-    Regex::new(r"Got waypoint: (\d+:\w+)")
+fn parse_waypoint(db_bootstrapper_output: &str) -> Waypoint {
+    let waypoint = Regex::new(r"Got waypoint: (\d+:\w+)")
         .unwrap()
         .captures(db_bootstrapper_output)
-        .ok_or_else(|| anyhow!("Failed to parse db-bootstrapper output."))
-        .unwrap()[1]
-        .into()
+        .ok_or_else(|| anyhow!("Failed to parse db-bootstrapper output."));
+    Waypoint::from_str(waypoint.unwrap()[1].into()).unwrap()
```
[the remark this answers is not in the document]
davidiw: I think there are better ways to do this... but ...

secure/storage/src/namespaced_storage.rs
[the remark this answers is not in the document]
davidiw: no, it was a bad pattern mixing too many things in my head at the same time :P

config/management/genesis/src/storage_helper.rs
davidiw: to_string <=> into... not sure which one is better but the latter has fewer characters :D

PR #5735, opened 2020-08-23
[move-prover] fixed specs in Errors

shazqadeer:
## Motivation

There appears to be either a bug in what "cargo test" does or my understanding of what it does.  Over the weekend, I introduced a bug in Errors.move which is not caught by 

cargo test test_runner_stdlib::modules/Errors.move

I happened to run 

cargo run -- ../../../../stdlib/modules/Errors.move -d ../../../../stdlib/modules/ --verbose debug

today and discovered verification failures.  This PR fixes those bugs.

My inability to detect the problems I had introduced in Errors.move led me to a wrong  conclusion about impact of the use of arithmetic modulus in computing error codes.  I erroneously came to the conclusion that the use of modulus is harmless.  I came to the opposite conclusion (mod indeed creates timeouts) once I fixed the bugs in Errors.move and repeated my experiments.  This PR also reverts the change I made over the weekend and eliminates the use of modulus (once again).

Yes.

## Test Plan

Existing tests

### DIFF
--- a/language/stdlib/modules/Errors.move
+++ b/language/stdlib/modules/Errors.move
@@ -14,6 +14,7 @@ address 0x1 {
 ///     framework evolves. TODO(wrwg): determine what kind of stability guarantees we give about reasons/
 ///     associated module.
 module Errors {
+    spec module { pragma verify; }
 
     /// A function to create an error from from a category and a reason.
     fun make(category: u8, reason: u64): u64 {
@@ -21,8 +22,9 @@ module Errors {
     }
     spec fun make {
         pragma opaque = true;
-        aborts_if false;
-        ensures result == category + (reason << 8);
+        ensures [concrete] result == category + (reason << 8);
+        aborts_if [abstract] false;
+        ensures [abstract] result == category;
     }
 
     /// The system is in a state where the performed operation is not allowed. Example: call to a function only allowed
--- a/language/move-prover/src/bytecode_translator.rs
+++ b/language/move-prover/src/bytecode_translator.rs
@@ -1811,7 +1811,7 @@ impl<'env> ModuleTranslator<'env> {
                 );
                 emitln!(
                     self.writer,
-                    "$abort_code := i#$Integer({}) mod 256;",
+                    "$abort_code := i#$Integer({});",
                     str_local(*src)
                 );
                 emitln!(self.writer, "goto Abort;")
--- a/language/stdlib/modules/doc/Errors.md
+++ b/language/stdlib/modules/doc/Errors.md
@@ -430,6 +430,12 @@ A function to create an error from from a category and a reason.
 ## Specification
 
 
+
+<pre><code>pragma verify;
+</code></pre>
+
+
+
 <a name="0x1_Errors_Specification_make"></a>
 
 ### Function `make`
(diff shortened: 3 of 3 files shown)


language/stdlib/modules/Errors.move
```diff
@@ -14,15 +14,17 @@ address 0x1 {
...
 
     /// A function to create an error from from a category and a reason.
     fun make(category: u8, reason: u64): u64 {
         (category as u64) + (reason << 8)
     }
     spec fun make {
         pragma opaque = true;
-        aborts_if false;
-        ensures result == category + (reason << 8);
+        aborts_if category + (reason << 8) > max_u64();
```
wrwg, line 25: Since the abstract version cannot abort, I would suggest to add `aborts_if [abstract] false` and mark this as concrete. Gives on thing less to reason about for the prover.

I'm not sure 100% though whether abstract/concrete works correctly for aborts_if, not remember have tested it. In case you run into trouble with this, perhaps just a TODO and verify = false. That this is light weight for callers is the most important matter.

language/stdlib/modules/Errors.move
```diff
@@ -63,70 +65,70 @@ module Errors {
     public fun invalid_state(reason: u64): u64 { make(INVALID_STATE, reason) }
     spec fun invalid_state {
         pragma opaque = true;
-        aborts_if false;
+        aborts_if INVALID_STATE + (reason << 8) > max_u64();
```
wrwg, line 68: Same here. All those should be aborts_if false for the caller. Otherwise the prover has to show all the time this does not happen. (It doesn't because of constants passed to those functions, but prover has to show this.)

shazqadeer:
I am chuckling at your comments.  I was puzzling about exactly the same things about an hour ago.  I didn't know what specific "unsound" thing to do so I just went with trying to prove the aborts specification hoping that the prover will be able to prove those at the call sites.  Interestingly, I didn't see any timeouts.  I had compiled a list of runtimes for most of the modules yesterday when I was doing my performance investigation.  I see a modest  slowdown for LibraAccount (45s vs 57s solver  time) but no timeouts.  I didn't notice any slowdown for the others.  In light of this data, what is your recommendation?

language/stdlib/modules/Errors.move
wrwg: BTW, why is this even needed? The expression can never overflow (shift does not overflow at least in Move. Perhaps there is a bug in our implementation of it.)

wrwg:
> I am chuckling at your comments. I was puzzling about exactly the same things about an hour ago. I didn't know what specific "unsound" thing to do so I just went with trying to prove the aborts specification hoping that the prover will be able to prove those at the call sites. Interestingly, I didn't see any timeouts. I had compiled a list of runtimes for most of the modules yesterday when I was doing my performance investigation. I see a modest slowdown for LibraAccount (45s vs 57s solver time) but no timeouts. I didn't notice any slowdown for the others. In light of this data, what is your recommendation?

I would just remove any potential overhead for this (even though you don't see a big difference now), making sure the opaque version of those functions is simply identity of the category and never aborts. We can followup on the shift bug later.

language/stdlib/modules/Errors.move
shazqadeer: If left shift is implemented simply as multiplication with a constant without any guard, I suppose it could lead to a number larger than max_u64.

PR #5740, opened 2020-08-24
[Fuzzing]  Add fuzzing target for handle_message() in Libra Safety Rules (LSR).

JoshLind:
## Motivation

This PR adds a fuzz target for the handle_message() method in LSR: the main network entry point into safety rules. 

To achieve this, the PR offers a single commit that alphabetizes the libra-fuzzer inputs, and another single commit that:
1. Adds a new fuzzing module that generates input for the handle_message() method using arb_handle_message_input().
2. Adds new methods for creating arbitrary consensus types and inputs (e.g., the SafetyRulesInput enum for handle_message()).
3. Exposes several new constructor methods for producing arbitrary consensus types for fuzzing. 

Note: 70% this PR is dedicated to just creating arbitrary consensus types using proptest. Once this PR lands, we can verify the coverage of LSR via the dashboard and see how to improve the consensus types generated by proptest. For example, right now, signatures are created from arbitrary bytes, which is highly unlikely to produce a valid signature over data, which will cause the fuzzer not to be able to reach certain paths. We can improve this coverage in a follow up PR and instead use this PR as a starting point.

Yes.

## Test Plan

All local tests pass. Moreover, running the fuzzer locally seems to work.

## Related PRs

None.


testsuite/libra-fuzzer/src/fuzz_targets/safety_rules.rs
```diff
@@ -0,0 +1,36 @@
...
+        Some(corpus_from_strategy(arb_handle_message_input()))
+    }
+
+    fn fuzz(&self, data: &[u8]) {
+        // Generate the fuzzer input
+        let (safety_rules_input, verify_vote_proposal_sig, signer_byte) =
+            fuzz_data_to_value(data, arb_handle_message_input());
+
+        // Create a safety rules serializer test instance for fuzzing
+        let mut serializer_service =
```
JoshLind, line 28: @mimoo, one of the disadvantages of fuzzing a method (as opposed to a function), is that we need to create a new   object before running each fuzz iteration. 
1. I'm guessing there's no easy way around this without completely refactoring all the code (which may not be possible to remove dependencies)?
2. Alternatively, do you know if it's possible to create a single object that is used by all fuzzing iterations? Although, the disadvantages of this approach is that each fuzz iteration may cause the object to enter a state that that prevents us from exploring all states/paths.

consensus/safety-rules/src/test_utils.rs
```diff
@@ -224,3 +226,12 @@ pub fn test_storage(signer: &ValidatorSigner) -> PersistentSafetyStorage {
         waypoint,
     )
 }
+
+pub fn test_serializer(signer_byte: u8, verify_vote_proposal_signature: bool) -> SerializerService {
+    let signer = ValidatorSigner::from_int(signer_byte);
```
mimoo, line 231: I suggest not fuzzing this part and always using the same signer

mimoo: it'll also be faster (if you use once_cell::lazy for that)

consensus/safety-rules/src/fuzzing_utils.rs
```diff
@@ -0,0 +1,301 @@
...
+        SafetyRulesInput::SignTimeout(Box::new(Timeout::new(epoch, round)))
+    }
+}
+
+// This generates an arbitrary and optional Ed25519Signature.
+prop_compose! {
+    pub fn arb_ed25519_signature(
+    )(
+        include_signature in any::<bool>(),
+        signature_bytes in prop::collection::vec(any::<u8>(), 0..Ed25519Signature::LENGTH),
```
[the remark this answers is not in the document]
JoshLind: Yeah, I think so, but I wanted to land this first to get a feeling for coverage and then figure out how to improve things. My concern is that for "good" coverage the signature won't just have to be valid, but it'll need to be using the expected key, as well as sign the associated data, so it'll make this code more complex. I was hoping to land a V0 of this fuzzer first and then iterate if it's blocked by these things. Does that makes sense?

testsuite/libra-fuzzer/src/fuzz_targets/safety_rules.rs
JoshLind: Yeah, true, but if I'm re-initializing it before each run, it's effectively the same logical behaviour, right? In that case, I'll just leave the object creation here and avoid exposing a global.

consensus/safety-rules/src/test_utils.rs
JoshLind: Sure, thanks -- I was originally thinking of fuzzing these values to increase the coverage of possible Serializer states, but I've just kept it simple for now and created these things with constants (e.g., the signer uses the int 0). We can see if we need to improve this later, once we look at the coverage

consensus/consensus-types/src/block_data.rs
```diff
@@ -142,6 +142,24 @@ impl BlockData {
         BlockData::new_genesis(ledger_info.timestamp_usecs(), genesis_quorum_cert)
     }
 
+    #[cfg(any(test, feature = "fuzzing"))]
+    // This method should only used by tests and fuzzers to produce arbitrary BlockData types.
+    pub fn new_block_data(
```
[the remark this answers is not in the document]
JoshLind: Gone for "new_for_tests" as suggested by @mimoo.

JoshLind:
Thanks, all! Ready for another look.

> will this generate full correct or nearly fully correct messages? Should we?

It depends how you define "correct".. It should generate correctly typed messages, so they'll enter the handle_message() function, pass deserialization and go to the corresponding hander. However, when we get to semantically correct (e.g., signatures, expected values of block data, etc.), it depends on what values are generated. My feeling is we'll need at least one other PR to improve coverage by helping to guide the arb_handle_message() function in the right direction 😄

testsuite/libra-fuzzer/src/fuzz_targets/safety_rules.rs
mimoo: depending on the situation sometimes it's faster to re-initialize an already created object : o

consensus/safety-rules/src/fuzzing_utils.rs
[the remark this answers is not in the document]
mimoo: yeah this is a problem with fuzzing, you either have to add valid signatures (takes time) or bypass signature verification (which sometimes require a refactor, or worse some code logic to disable signatures that can scare people)

consensus/safety-rules/src/lib.rs
```diff
@@ -24,7 +24,10 @@ pub use crate::{
     t_safety_rules::TSafetyRules,
 };
 
-#[cfg(any(test, feature = "testing"))]
+#[cfg(any(test, feature = "fuzzing"))]
+pub mod fuzzing_utils;
+
+#[cfg(any(test, feature = "fuzzing", feature = "testing"))]
 #[path = "test_utils.rs"]
```
mimoo, line 31: is that path necessary oO?

JoshLind: Good question... I'm not sure why it's there.. I'll play with it in a separate PR and if not, just remove it 😄

JoshLind:
Thanks, David!

PR #5745, opened 2020-08-24
[crypto, test-only] Point decompression checks + test output

huitseeker:
Strengthens the crypto tests a tad more.


crypto/crypto/src/unit_tests/ed25519_test.rs
```diff
@@ -366,6 +379,7 @@ proptest! {
         prop_assert!(deserialized.verify(&hashable, &keypair.public_key).is_ok());
     }
 
+
```
kchalkias, line 382: \-

PR #5752, opened 2020-08-25
logger: fix bug preventing assigning names to text log fields

bmwill:


common/logger/src/struct_log.rs
```diff
@@ -52,7 +52,7 @@ const CONNECTION_TIMEOUT_MS: u64 = 5000;
 // Fields to keep when over size
 static FIELDS_TO_KEEP: &[&str] = &["error"];
 
-#[derive(Default, Serialize)]
+#[derive(Debug, Default, Serialize)]
```
gregnazario, line 55: Why implement Debug for StructuredLogEntry?

bmwill: No reason not to, that and I used it to help me debug while writing this patch

PR #5755, opened 2020-08-25
[json-rpc] fix fuzzing test

xli:
Fixed existing json-rpc fuzzing test. Tested local with continuous running test 15 hours.
This test should cover rpc_endpoint method in json-rpc.

### DIFF
--- a/json-rpc/src/tests/mod.rs
+++ b/json-rpc/src/tests/mod.rs
@@ -8,3 +8,5 @@ mod unit_tests;
 mod utils;
 
 pub use utils::test_bootstrap;
+#[cfg(any(test, feature = "fuzzing"))]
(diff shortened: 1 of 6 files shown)


json-rpc/src/tests/mod.rs
```diff
@@ -8,3 +8,5 @@ mod unit_tests;
 mod utils;
 
 pub use utils::test_bootstrap;
+#[cfg(any(test, feature = "fuzzing"))]
+pub use utils::MockLibraDB;
```
mimoo, line 12: pub(crate) no?

mimoo: I guess it's test stuff, doesn't matter much

PR #5761, opened 2020-08-25
[logging] All logs with text sent to text logs

gregnazario:
### Overview
In order to allow us to debug live, all logs will now print the text portion to stdout.  In addition, renamed the `log` field to `text` to remove confusion.

This is currently going to have conflicts with my other two changes out to remove `send_struct_log!` and `crit!`, but I'll work them out as they go :)

I believe I will need to reformat the logs to be more useful with level and line information, will do as I test.

### Testing
TBD

### DIFF
--- a/common/logger/src/lib.rs
+++ b/common/logger/src/lib.rs
@@ -163,7 +163,6 @@ mod tests;
 macro_rules! crit {
     ($($arg:tt)+) => ({
         $crate::text_to_struct_log!($crate::log::Level::Error, $($arg)+);
-        $crate::log::error!(target: $crate::DEFAULT_TARGET, $($arg)+);
     })
 }
 
@@ -172,7 +171,6 @@ macro_rules! crit {
 macro_rules! debug {
     ($($arg:tt)+) => ({
         $crate::text_to_struct_log!($crate::log::Level::Debug, $($arg)+);
-        $crate::log::debug!(target: $crate::DEFAULT_TARGET, $($arg)+);
     })
 }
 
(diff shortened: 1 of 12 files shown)


common/logger/src/lib.rs
```diff
@@ -30,7 +30,7 @@
 //! use serde_json::Value;
 //!
 //! pub struct StructuredLogEntry {
-//!     log: Option<String>,
+//!     text: Option<String>,
```
bmwill, line 33: If we're going about changing this, maybe `message` is a little more descriptive than text?

gregnazario: Yeah, message is probably better.  Let me do some git magic to fix it up

common/logger/src/lib.rs
```diff
@@ -292,15 +286,15 @@ macro_rules! location {
...
         $entry.add_pattern($fmt);
     };
     ($entry:ident, $fmt:expr,) => {
-        $entry.add_log(format!($fmt));
+        $entry.add_message(format!($fmt));
         $entry.add_pattern($fmt);
     };
     ($entry:ident, $fmt:expr, $($arg:tt)+) => {
-        $entry.add_log(format!($fmt, $($arg)+));
+        $entry.add_message(format!($fmt, $($arg)+));
```
davidiw, line 297: can we get rid of these add_*? and just have *

gregnazario:
> is this intentionally in draft?

It's still in draft, because I wanted to test it on a cluster test to ensure I can see text logs are printed out.  A lot had been changing on this PR

common/logger/src/lib.rs
gregnazario: Yeah, it's a little confusing because of the way the macros are built recursively, but it is possible.

gregnazario:
Closing in favor of https://github.com/libra/libra/pull/5875

PR #5766, opened 2020-08-25
[network] Remove Raw(Enc)NetworkAddress

davidiw:
* Merge RawNetworkAddress encoding into NetworkAddress, so that we can
still handle incompatible versions in a Vec<Serialize Network Addresses>
* Eliminate RawEncNetworkAddress since versionining should both be
uniform and the format is so simple as to not benefit from versioning.

Removing because it came up often enough that it was confusing people
why we had this layer of abstraction... and writing the addresses is an
atomic operation, if the validator operator messed up one, then they
probably messed up the rest.

this is not actually a breaking change...

### DIFF
--- a/network/network-address/src/encrypted.rs
+++ b/network/network-address/src/encrypted.rs
@@ -1,7 +1,7 @@
 // Copyright (c) The Libra Core Contributors
 // SPDX-License-Identifier: Apache-2.0
 
-use crate::RawNetworkAddress;
+use crate::{NetworkAddress, ParseError};
 use aes_gcm::{
     aead::{generic_array::GenericArray, AeadInPlace, NewAead},
     Aes256Gcm,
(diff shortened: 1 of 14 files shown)


network/network-address/src/encrypted.rs
```diff
@@ -291,7 +242,7 @@ impl EncNetworkAddress {
         // remove the auth tag suffix, leaving just the decrypted network address
         enc_addr.truncate(auth_tag_offset);
 
-        Ok(RawNetworkAddress::new(enc_addr))
+        Ok(lcs::from_bytes(&enc_addr).unwrap())
```
zekun000, line 245: if we assume lcs is infallible we probably should make it not return Result or have a wrapper not return Result but just unwrap inside instead of unwrap at every callsites

davidiw: Humm... good call, this could be fallible actually :/

davidiw:
> Finally no middle conversion layer

middle out compression.

PR #5769, opened 2020-08-25
[consensus] better error handling of execution failure

zekun000:
## Motivation

This commit extends the notify to mempool to support failed block
execution.

This can help address issues when a txn successfully executed but causes
internal error in the execution e.g. rotate to invalid validator network addresses.

Before this commit, such a txn can cause all exeuction fail until the
txn is expired in mempool.

After this commit, such a txn should be dropped immediately after the
first failure.

Y

## Test Plan

will add one when we have real types back in ValidatorConfig, currently it can't fail

## Related PRs

### DIFF
--- a/consensus/src/state_replication.rs
+++ b/consensus/src/state_replication.rs
@@ -17,7 +17,11 @@ pub trait TxnManager: Send + Sync {
 
     /// Notifies TxnManager about the executed result of the block,
     /// which includes the specifics of what transactions succeeded and failed.
-    async fn notify(&self, block: &Block, compute_result: &StateComputeResult) -> Result<()>;
+    async fn notify(
+        &self,
+        block: &Block,
+        executed_result: Result<&StateComputeResult>,
+    ) -> Result<()>;
 
     /// Helper to trace transactions after block is generated
     fn trace_transactions(&self, _block: &Block) {}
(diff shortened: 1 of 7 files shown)


consensus/src/state_replication.rs
```diff
@@ -17,7 +17,11 @@ pub trait TxnManager: Send + Sync {
 
     /// Notifies TxnManager about the executed result of the block,
     /// which includes the specifics of what transactions succeeded and failed.
-    async fn notify(&self, block: &Block, compute_result: &StateComputeResult) -> Result<()>;
+    async fn notify(
+        &self,
+        block: &Block,
+        executed_result: Result<&StateComputeResult>,
```
davidiw, line 23: why don't we embed the result in the StateComputeResult? this feels a little redundant... will read more ... but figured easier to ask a naive question.

davidiw: Hmm.... @zekun000

zekun000: StateComputeResult represents the successful executed result, I don't understand how do we embed the Result into that? that means we need another struct StateComputeSuccessfulResult to represent the existing struct, don't see what's the benefit.

zekun000:
turns out it's not possible for this to happen after we change network address to Vec<u8> in ValidatorConfig, closing

PR #5776, opened 2020-08-25
[network][handshake protocol] refactoring and implementing test/fuzzing

mimoo:
This PR does a number of non-logical but architectural refactoring:

* `transports.rs` had a lot of tests, so moving them out in their own file (and removing a TODO from @phlip9)
* move `HandshakeMsg` out of `UpgradeContext`. The rationale is that `UpgradeContext` should be clear and provide information about upgrading, where `HandshakeMsg` is an struct that is relevant only in the wire logic. We're losing an optimization that does not matter much (as we're not going to have too many connection establishment) and that we could regain elsewhere: if we want to optimize this flow we should have a `HandshakeContext` instead that: 1) can be stored in `NetworkContext` and 2) drives the handshake protocol and 3) caches the **serialized** `HandshakeMsg`.
* adding a test to the handshake protocol to make sure that chain id mismatch fails

In addition, it adds two fuzzers for the libranet handshake protocol:

* one that fuzzes the exchange of HandshakeMsg and the negotiation
* one that fuzzes only the negotiation to skip the LCS deserialization and other stuff

To do this it also implements proptest strategies on a few types


network/src/transport.rs
```diff
@@ -524,405 +524,3 @@ where
         Ok((listener, listen_addr))
     }
 }
-
```
davidiw, line 527: 1) This is an anti-pattern, I think the only time to move tests to their own file is when they are truly integration tests or there is some sort of common testing scheme across multiple implementations
2) This library really needs to be refactored. Transports shouldn't be this complicated.

network/src/transport.rs
```diff
@@ -118,22 +118,6 @@ impl ConnectionMetadata {
             application_protocols,
         }
     }
-
```
davidiw, line 121: this is a good commit

network/src/protocols/wire/handshake/v1/mod.rs
```diff
@@ -14,10 +14,15 @@ use libra_config::network_id::NetworkId;
...
 use std::{collections::BTreeMap, convert::TryInto, fmt, iter::Iterator};
+use thiserror::Error;
 
 #[cfg(test)]
 mod test;
 
+//
+// ProtocolId
+//
+
```
davidiw, line 28: I never understood the purpose of these types of comments. As a vim developer I search for enum ProtocolId ... if we need these, we should just separate them into different files.

network/src/protocols/wire/handshake/v1/mod.rs
```diff
@@ -56,34 +61,49 @@ impl fmt::Display for ProtocolId {
     }
 }
 
+//
+// SupportedProtocols
```
davidiw, line 68: same

network/src/protocols/wire/handshake/v1/mod.rs
```diff
@@ -56,34 +61,49 @@ impl fmt::Display for ProtocolId {
...
+impl SupportedProtocols {
+    /// Returns a new SupportedProtocols struct that is an intersection.
+    fn intersection(self, other: SupportedProtocols) -> SupportedProtocols {
+        SupportedProtocols(self.0 & other.0)
     }
 }
 
+//
+// MessageProtocolVersion
+//
```
davidiw, line 109: same

network/src/transport.rs
```diff
@@ -301,21 +319,33 @@ where
...
+        let network_id = network_context.network_id().clone();
+        let upgrade_context = UpgradeContext {
+            noise: NoiseUpgrader::new(network_context, identity_key, auth_mode),
+            handshake_version,
+            supported_protocols,
+            chain_id,
+            network_id,
+        };
+
+        //
```
davidiw, line 350: ??

network/src/transport.rs
```diff
@@ -301,21 +319,33 @@ where
...
+        let mut supported_protocols = BTreeMap::new();
+        supported_protocols.insert(SUPPORTED_MESSAGING_PROTOCOL, application_protocols);
+
+        // get network public key
         let identity_pubkey = identity_key.public_key();
 
+        // auth mode
+        // TODO(mimoo): should we build this based on the networkid, and not on trusted peers
         let auth_mode = match trusted_peers.as_ref() {
             Some(trusted_peers) => HandshakeAuthMode::mutual(trusted_peers.clone()),
```
davidiw, line 336: yuck, very yuck.... this should be the same setting as we specify further upstream... we explicitly state the direction of verification in the config...

network/src/transport.rs
```diff
@@ -301,21 +319,33 @@ where
         chain_id: ChainId,
         application_protocols: SupportedProtocols,
     ) -> Self {
-        let mut own_handshake = HandshakeMsg::new(chain_id, network_context.network_id().clone());
-        own_handshake.add(SUPPORTED_MESSAGING_PROTOCOL, application_protocols);
+        // build supported protocols
+        let mut supported_protocols = BTreeMap::new();
+        supported_protocols.insert(SUPPORTED_MESSAGING_PROTOCOL, application_protocols);
+
+        // get network public key
```
davidiw, line 330: not useful comment

network/src/transport.rs
```diff
@@ -301,21 +319,33 @@ where
         chain_id: ChainId,
         application_protocols: SupportedProtocols,
     ) -> Self {
-        let mut own_handshake = HandshakeMsg::new(chain_id, network_context.network_id().clone());
-        own_handshake.add(SUPPORTED_MESSAGING_PROTOCOL, application_protocols);
+        // build supported protocols
+        let mut supported_protocols = BTreeMap::new();
+        supported_protocols.insert(SUPPORTED_MESSAGING_PROTOCOL, application_protocols);
```
davidiw, line 328: why aren't we just building the handshake once at the beginning of time and pass in clones later? or why are we only partially building it and then finishing it later? seems like we move an excessive amount of data ... motivation?

network/src/transport.rs
```diff
@@ -301,21 +319,33 @@ where
...
         let identity_pubkey = identity_key.public_key();
 
+        // auth mode
+        // TODO(mimoo): should we build this based on the networkid, and not on trusted peers
         let auth_mode = match trusted_peers.as_ref() {
             Some(trusted_peers) => HandshakeAuthMode::mutual(trusted_peers.clone()),
             None => HandshakeAuthMode::ServerOnly,
         };
 
+        // create upgrade context
```
davidiw, line 340: please avoid these type of comments they're the first thing people delete later

network/src/protocols/wire/handshake/v1/mod.rs
mimoo: It's not for the actual struct or enum or whatever, but to group related structs, functions, implementations, etc. together. Before this refactor ProtocolId implementations where scattered all around the file and it was quite the challenge to read it : o right now it should be smoother for first-time comer or even for us to find things. The comments are really here to separate these things and force you to place new implementations where they belong :)

network/src/transport.rs
mimoo: so what I say in the TODO right?

network/src/transport.rs
mimoo: that's what I point out in my PR description :) I think we can further optimize this by serializing it once and caching the serialized handshake message. But that's for a further PR as this is getting out of hand already.

network/src/transport.rs
mimoo: but why? I actually really appreciate it when people leave these comments, it makes the code much more readable.

network/src/transport.rs
mimoo: (separation of return value)

network/src/transport.rs
mimoo: > This is an anti-pattern, I think the only time to move tests to their own file is when they are truly integration tests or there is some sort of common testing scheme across multiple implementations

That's definitely not a rule we follow in the codebase, and from what I've seen it's highly dependent on the crate. For example crypto moves ALL tests into test files. Having said that, I think it's enough to follow common sense on when to do this: when tests grow large, and your file is already large, it's a good idea to move tests out. It's all about readability.

network/src/protocols/identity.rs
```diff
@@ -74,13 +75,23 @@ mod tests {
...
         );
-        let mut client_handshake = HandshakeMsg::new(chain_id, network_id);
-        client_handshake.add(
+        let server_handshake = HandshakeMsg {
+            chain_id,
+            network_id: network_id.clone(),
+            supported_protocols,
+        };
+        let mut supported_protocols = BTreeMap::new();
+        supported_protocols.insert(
```
gregnazario, line 84: So, this changes the paradigm of adding supported protocols.  I didn't like the name `add` for the previous function, but I think `add_protocol` might be more apt.

The issue here with this change in paradigm, is now supported protocols need to be added ahead of time.

I do like having to put them in ahead of time, but what does this buy to be changed right now?

network/src/protocols/wire/handshake/v1/mod.rs
gregnazario: I agree, cut the comments. We should stick to rust doc comments and not add these sectional comments.

Personally, if the file is so big that you need these "sectional" comments, there should be separate files for these components.  However, I don't want you to move the enums and impl right now.

network/src/transport.rs
gregnazario: I've made this mistake before about moving to this anti-pattern.

Let's *not* move the tests because this causes a couple things:
1. I can't tell if you changed the tests
2. Later it shows you've made all the tests, and you lose all history on the changes (at least by doing git blame)

network/src/transport.rs
gregnazario: ??  No empty comments please.  Whitespace can do the separation itself.

network/src/transport.rs
```diff
@@ -317,21 +323,33 @@ where
         chain_id: ChainId,
         application_protocols: SupportedProtocols,
     ) -> Self {
-        let mut own_handshake = HandshakeMsg::new(chain_id, network_context.network_id().clone());
-        own_handshake.add(SUPPORTED_MESSAGING_PROTOCOL, application_protocols);
+        // build supported protocols
```
gregnazario, line 326: So, about these comments.  I do appreciate helpful comments, but this may be a little too verbose  on about every 2 lines.  I once lost 20 points on a coding assignment in college from these kinds of comments.

If you want to add comments like this, they should be a level higher.  However, since this whole block basically builds an upgrade_context, it doesn't really need any comments on its own.

I'm a fan of named functions that do pieces, like `build_upgrade_context()`, but that's a personal preference that I don't think aligns with the current code base and gets a little flack especially as an anti-pattern.

network/src/transport.rs
```diff
@@ -317,21 +323,33 @@ where
...
 
+        // create upgrade context
+        let network_id = network_context.network_id().clone();
+        let upgrade_context = UpgradeContext {
+            noise: NoiseUpgrader::new(network_context, identity_key, auth_mode),
+            handshake_version,
+            supported_protocols,
+            chain_id,
+            network_id,
+        };
```
gregnazario, line 348: If the `NoiseUpgrader` has the`network_context` why pass the `network_id` to the `UpgradeContext`?

network/src/transport.rs
```diff
@@ -270,16 +246,46 @@ async fn upgrade_outbound<T: TSocket>(
...
-    let socket = ctxt
+    let mut socket = ctxt
         .noise
         .upgrade_outbound(socket, remote_pubkey, AntiReplayTimestamps::now)
         .await?;
 
     // sanity check: Noise IK should always guarantee this is true
     debug_assert_eq!(remote_pubkey, socket.get_remote_static());
 
+    // create HandshakeMsg
```
gregnazario, line 257: This would also be excessive commenting

network/src/transport.rs
gregnazario: Yeah, it is good to get rid of this anti-pattern.  It's something I think we need to do as a whole sweep of the networking code

network/src/transport.rs
```diff
@@ -226,7 +167,9 @@ where
 struct UpgradeContext {
     noise: NoiseUpgrader,
     handshake_version: u8,
-    own_handshake: HandshakeMsg,
+    supported_protocols: BTreeMap<MessagingProtocolVersion, SupportedProtocols>,
```
gregnazario, line 170: Is this a BTreeMap over a HashMap for size performance?

network/src/protocols/wire/handshake/v1/mod.rs
```diff
@@ -110,87 +135,105 @@ impl fmt::Display for MessagingProtocolVersion {
...
         other: &HandshakeMsg,
-    ) -> Option<(MessagingProtocolVersion, SupportedProtocols)> {
-        // First, find the highest MessagingProtocolVersion supported by both nodes.
+    ) -> Result<(MessagingProtocolVersion, SupportedProtocols), HandshakeError> {
+        // verify that both peers are on the same chain
+        if self.chain_id != other.chain_id {
+            return Err(HandshakeError::InvalidChainId(other.chain_id));
+        }
+
+        // verify that both peers are on the same type of network
```
gregnazario, line 190: I think these are self explanatory without the comments

network/src/protocols/wire/handshake/v1/mod.rs
```diff
@@ -110,87 +135,105 @@ impl fmt::Display for MessagingProtocolVersion {
...
+                    return Ok((
                         *k_inner,
                         protocols_self.clone().intersection(protocols_other.clone()),
                     ));
                 }
                 _ => {}
             }
         }
-        None
+
```
gregnazario, line 219: Why inline it all?

network/src/transport.rs
mimoo: 1. I did not change the tests
2. that sounds a bit too much, this means you can't refactor anything or touch any code that was written by someone else?

network/src/protocols/wire/handshake/v1/mod.rs
mimoo: I don't agree with that, you don't split a file that is disorganized, you organize it. If it's too large sure, but that's not the problem here. This is what other crates like crypto do as well, and it greatly helps readability. It'd be good if we encouraged PRs that improve readability imo, instead of sticking to a block of code that someone needs to reverse engineer to understand. I think long-term the handshake can be simplified and optimized and this is a step in the right direction

network/src/transport.rs
mimoo: Sure, that's my style but I can see how people would not like it

config/Cargo.toml
```diff
@@ -11,13 +11,14 @@ edition = "2018"
 
 [dependencies]
 get_if_addrs = { version = "0.5.3", default-features = false }
+log = { version = "0.4.11", features = ["serde"] }
 mirai-annotations = "1.9.1"
 rand = "0.7.3"
 serde = { version = "1.0.114", features = ["rc"], default-features = false }
 serde_yaml = "0.8.13"
-log = { version = "0.4.11", features = ["serde"] }
```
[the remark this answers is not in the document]
mimoo: I see these sort of changes all the times in PR and I think it slowly improves the codebase, perhaps let's focus on the actual content of the PR?

network/src/protocols/identity.rs
mimoo: This is a non-issue for now as this is what we do, so it's simpler. We can potentially add a method on UpgradeContext but I don't see the use, I also don't think HandshakeMsg should be part of UpgradeContext in its current form (if we want to optimize we need to cache the serialized HandshakeMsg instead).

network/src/transport.rs
mimoo: > I once lost 20 points on a coding assignment in college from these kinds of comments

I wouldn't sweat whatever happened with a bad teacher :P

mimoo: > I'm a fan of named functions that do pieces

I argued against this, we should inline code as much as possible and document the behavior of a function, sure if you understand the function it's useless but it's all about readability and making sure that people reviewing the code can quickly understand the different steps we take

network/src/transport.rs
mimoo: this is used by the handshake protocol

network/src/transport.rs
mimoo: no idea, IMO we should have just used a HashMap :P

network/src/protocols/wire/handshake/v1/mod.rs
mimoo: because there's no reason to have that in a separate function, it just adds layers of indirection when this function is clearer with the inlining

network/src/protocols/wire/handshake/v1/mod.rs
mimoo: I like these as it reads like a specification, or can be matched with a specification more easily...

mimoo:
> There's some good stuff here, but I also think there's too much going on in this PR. There's large pieces of code that were moved and it's really hard for me to tell what's changed and what was just moved.

> Can you just have a commit to implement fuzzing and put that through in the PR? Is there something in the refactor that's necessary?

I cut the PR into manageable sized commits, besides the refactor of the handshake protocol which I spent considerable time understanding on how to improve and re-organize in a clearer way. Why waste this work? We should try to encourage folks to contribute and help improve parts of the codebase, asking someone to get rid of work is not a good way to get more folks to help.

network/src/transport.rs
gregnazario: Ah, I finally figured out how to view by commit.  Yeah, now I see there were no changes.

There's an overarching discussion about what our future test breakdown should be.  I've had people argue with me when I separate the files, put it at the end of a file, and it doesn't even match our coding guidelines, which is funny this is the first one that actually matches it!

gregnazario:
> > There's some good stuff here, but I also think there's too much going on in this PR. There's large pieces of code that were moved and it's really hard for me to tell what's changed and what was just moved.
> 
> > Can you just have a commit to implement fuzzing and put that through in the PR? Is there something in the refactor that's necessary?
> 
> I cut the PR into manageable sized commits, besides the refactor of the handshake protocol which I spent considerable time understanding on how to improve and re-organize in a clearer way. Why waste this work? We should try to encourage folks to contribute and help improve parts of the codebase, asking someone to get rid of work is not a good way to get more folks to help.

Sorry about this as I just learned how to view by commit in the PR tool, stupid me.  Now, all the commits make sense.

network/src/transport.rs
gregnazario: :( yeah, okay then let's leave it.

network/src/transport.rs
gregnazario: @davidiw This is now sadly the one of the few files that now actually follows the coding guidelines https://developers.libra.org/docs/community/coding-guidelines#testing

network/src/protocols/wire/handshake/v1/mod.rs
gregnazario: Yeah, I was just wondering.  I think this definitely made it cleaner, especially since the function is only used once.

config/Cargo.toml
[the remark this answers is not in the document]
gregnazario: Yeah, just was wondering.  We all do this all over the place

network/src/transport.rs
mimoo: I know Brandon and josh talked about that, imo we should ere on the side of readability: whenever your file gets too large, and you have many tests  -> move your tests. It's all about readability for people who review code on ramp up on the codebase.

gregnazario: I'm with you on that.  We've got some really long files in the codebase

network/src/protocols/wire/handshake/v1/mod.rs
```diff
@@ -110,87 +135,105 @@ impl fmt::Display for MessagingProtocolVersion {
...
+        // no intersection found
+        Err(HandshakeError::NoCommonProtocols)
+    }
+}
+
+impl fmt::Debug for HandshakeMsg {
+    fn fmt(&self, f: &mut Formatter<'_>) -> fmt::Result {
+        write!(f, "{}", self)
+    }
+}
```
gregnazario, line 229: Note for myself, really should probably make a procedural macro for this, I've written this like 20 times.

network/src/fuzzing.rs
```diff
@@ -0,0 +1,107 @@
...
+    // fake socket to read the other peer's serialized HandshakeMsg from
+    let mut fake_socket = ReadOnlyTestSocketVec::new(data);
+    fake_socket.set_trailing();
+
+    // fuzz the network exchange of HandshakeMsg first
+    let _ = block_on(async move {
+        if let Ok(remote_handshake_msg) = exchange_handshake(self_handshake, &mut fake_socket).await
+        {
+            // then perform the negotiation
+            let _ = self_handshake.perform_handshake(&remote_handshake_msg);
```
gregnazario, line 44: I really wish there was a better way in rust to throw away the return values.  I guess we don't really care what the response is?

network/src/protocols/wire/handshake/v1/mod.rs
mimoo: I think there's a SilentDebug macro somewhere

mimoo: not perfect, but could be copy/pasted: https://github.com/libra/libra/blob/cdbf57301aaee34d6b0102918a87826669e5917d/crypto/crypto-derive/src/lib.rs#L113

mimoo: oh wait, this is just to use display as debug?

network/src/fuzzing.rs
mimoo: we could in theory test some properties here, for example that the intersection only includes things from both handshake message, but that might be a bit too much :p

PR #5777, opened 2020-08-25
[transaction-replay] Implement a transaction bisection tool

runtian-zhou:
## Motivation

Implemented a bisection tool to investigate on chain state by using a move function as a predicate.

Example:
- To find when an account is created, use:
```
cargo run --bin libra-transaction-replay -- --db <path-to-db> bisect-transaction ~/libra/language/tools/transaction-replay/examples/account_exists.move <account-to-query> <begin-version> <end-version> 
```

The bisection tool utilizes binary search to execute a user provided script and find the first version that causes script execution to abort.

Yes

## Test Plan

Added a unit test for binary search and an e2e test to smoke test with libra-swarm.


testsuite/tests/libratest/smoke_test.rs
```diff
@@ -1987,6 +1987,21 @@ fn test_replay_tooling() {
...
+        .bisect_transactions_by_script(script_path.to_str().unwrap(), account, 0, txn.version)
+        .unwrap()
+        .unwrap();
+
+    let account_creation_txn = client_proxy
+        .get_committed_txn_by_acc_seq(&["txn_acc_seq", "0000000000000000000000000a550c18", "1", "false"])
+        .unwrap()
+        .unwrap();
+
+    assert_eq!(account_creation_txn.version + 1, bisect_result);
```
runtian-zhou, line 2004: There is an off-by-one because the bisect tool return you the first version when the script fails, which means the previous version caused the account to be created.

bob-wilson:
This is awesome! I'm so glad to see this. It should be really useful for debugging. I already have a feature request: it would be nice to have an option to test a full transaction (instead of just a snippet of Move code), e.g., so that we can test for problems reported by the transaction prologue. That can come later, though (not in this PR)

language/tools/transaction-replay/src/lib.rs
```diff
@@ -2,22 +2,34 @@
...
 use move_vm_runtime::{move_vm::MoveVM, session::Session};
+use move_vm_types::gas_schedule::{zero_cost_schedule, CostStrategy};
 use resource_viewer::{AnnotatedAccountStateBlob, MoveValueAnnotator};
 use std::{convert::TryFrom, path::Path};
+use stdlib::stdlib_files;
 use vm::errors::VMResult;
 
 pub use crate::transaction_debugger_interface::{DebuggerStateView, StorageDebuggerInterface};
 
+pub const STD_LIB_DIR: &str = "../../stdlib/modules";
```
bob-wilson, line 28: Is this used anywhere? I don't see anything

PR #5778, opened 2020-08-25
[vm] Fix tracking for VM_MAX_VALUE_DEPTH_REACHED

tnowacki:
- Made type layout generation eager in data store for events and resources
- The eagerness moves the error from change set generation to the instruction that causes the change, thus giving location information

## Motivation

- Making VM_MAX_VALUE_DEPTH_REACHED fit into the execution failure bucket 

## Test Plan

- New test for events

### DIFF
--- a/language/move-lang/functional-tests/tests/natives/lcs_large_value.move
+++ b/language/move-lang/functional-tests/tests/natives/lcs_large_value.move
@@ -84,4 +84,4 @@ script {
         M::encode_257();
     }
 }
-// check: ABORTED { code: 453
+// check: "ABORTED { code: 453, location: 00000000000000000000000000000001::LCS }"
--- a/language/move-core/types/src/vm_status.rs
+++ b/language/move-core/types/src/vm_status.rs
@@ -177,14 +177,11 @@ impl VMStatus {
                     // to decoding, running the prologue etc.
                     StatusType::Deserialization => Ok(KeptVMStatus::DeserializationError),
                     // Any error encountered during the execution of the transaction will charge gas.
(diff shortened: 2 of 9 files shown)


tnowacki:
This is "technically" a breaking change, as the offsets for execution failures for VM_MAX_VALUE_DEPTH_REACHED will be different now. But this error will never be reached in any of the code we currently have. 

How do we want to approach this?

bob-wilson:
> How do we want to approach this?

I'm generally inclined to be conservative about what we consider to be breaking changes, but in this case, it really seems like overkill. Maybe someone else will think of a case where it could matter, but I can't come up with anything. I'd be OK with not treating this as breaking if no one objects.

language/move-lang/functional-tests/tests/natives/lcs_large_value.move
```diff
@@ -84,4 +84,4 @@ script {
         M::encode_257();
     }
 }
-// check: ABORTED { code: 453
+// check: ABORTED { code: 453, location: 00000000000000000000000000000001::LCS }
```
vgao1996, line 87: Just a side note:
```
// check: ABORTED { code: 453, location: 00000000000000000000000000000001::LCS }
// gets desugared into 
// check: ABORTED
// check: {
// check: 453,
// check: location:
// check: 00000000000000000000000000000001::LCS
// check: }
```

To be clear, this was a terrible design choice I made at the time and the whole checker is likely going away soon. For now I'd suggest quoting the whole thing.
```// check: "ABORTED { code: 453, location: 00000000000000000000000000000001::LCS }"```

PR #5782, opened 2020-08-26
[fuzzer][executor] create fuzzer for block execution and committing

lightmark:
## Motivation

executor fuzzing targets 100%

Y

### DIFF
--- a/execution/executor/src/fuzzing.rs
+++ b/execution/executor/src/fuzzing.rs
@@ -3,8 +3,8 @@
 
 use crate::Executor;
 use anyhow::Result;
-use executor_types::ChunkExecutor;
-use libra_crypto::HashValue;
+use executor_types::{BlockExecutor, ChunkExecutor};
+use libra_crypto::{hash::SPARSE_MERKLE_PLACEHOLDER_HASH, HashValue};
 use libra_state_view::StateView;
 use libra_types::{
     account_address::AccountAddress,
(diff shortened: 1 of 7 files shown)


testsuite/libra-fuzzer/src/fuzz_targets.rs
```diff
@@ -28,6 +28,7 @@ static ALL_TARGETS: Lazy<BTreeMap<&'static str, Box<dyn FuzzTargetImpl>>> = Lazy
         Box::new(consensus::ConsensusProposal::default()),
         // Executor
         Box::new(executor::ExecuteAndCommitChunk::default()),
+        Box::new(executor::ExecuteAndCommitBlocks::default()),
```
JoshLind, line 31: nit: alphabetical by name (i.e., move this new target above ExecuteAndCommitChunk)

testsuite/libra-fuzzer/src/fuzz_targets/executor.rs
```diff
@@ -31,3 +54,9 @@ prop_compose! {
         (txn_list_with_proof, verified_target_li)
     }
 }
+
+prop_compose! {
+    fn execute_and_commit_blocks_input()(blocks in vec((any::<HashValue>(), vec(any::<Transaction>(), 0..10)), 0..10), li_with_sigs in any::<LedgerInfoWithSignatures>()) -> (Vec<(HashValue, Vec<Transaction>)>, LedgerInfoWithSignatures) {
```
JoshLind, line 59: Can you manually fix the argument indentation for this? It seems like cargo fmt doesn't handle stuff inside prop_compose.. e.g.,

fn execute_and_commit_blocks_input()(
    blocks in vec(...),
    li_with_sigs in any...,
) -> (....) {

execution/executor/src/lib.rs
```diff
@@ -726,17 +726,19 @@ impl<V: VMExecutor> BlockExecutor for Executor<V> {
...
 
-        if version + 1 < num_persistent_txns {
-            return Err(Error::InternalError {
-                error: format!(
-                    "Try to commit stale transactions with the last version as {}",
-                    version
-                ),
-            });
-        }
+        if num_persistent_txns > 0 {
```
JoshLind, line 729: Why the change in logic? Old bug?

testsuite/libra-fuzzer/src/fuzz_targets/executor.rs
lightmark: yes, it ignores everything in macro.

execution/executor/src/lib.rs
lightmark: fuzzer reports a case that the addition in old code can overflow.

execution/executor/src/lib.rs
```diff
@@ -735,7 +736,7 @@ impl<V: VMExecutor> BlockExecutor for Executor<V> {
             });
         }
 
-        if version + 1 == num_persistent_txns {
+        if version.saturating_add(1) == num_persistent_txns {
```
mimoo, line 739: As you're in a function that return a result, it is better to use `checked_add` rather than `saturating_add`, and return an error if `checked_add` returns `None`. I don't think there's a risk in using `saturating_add` here, but we can never be too sure.

execution/executor/src/fuzzing.rs
```diff
@@ -32,10 +32,29 @@ pub fn fuzz(
...
 
-    //
     let _events = executor.execute_and_commit_chunk(txn_list_with_proof, verified_target_li, None);
 }
 
+pub fn fuzz_execute_and_commit_blocks(
+    blocks: Vec<(HashValue, Vec<Transaction>)>,
+    ledger_info_with_sigs: LedgerInfoWithSignatures,
+) {
+    // setup fake db
```
JoshLind, line 42: nit: you could probably save some lines by defining a helper `create_test_executor()`, which is called by both fuzzing methods.

execution/executor/src/lib.rs
```diff
@@ -743,7 +743,11 @@ impl<V: VMExecutor> BlockExecutor for Executor<V> {
         let version = ledger_info_with_sigs.ledger_info().version();
         let num_persistent_txns = self.cache.synced_trees().txn_accumulator().num_leaves();
 
-        if version + 1 < num_persistent_txns {
+        if version
```
JoshLind, line 746: nit: part of me wonders if it'll be cleaner to create a helper for this (i.e., to avoid having to create a new format error with "version + 1 overflows", twice)?

mimoo: that's what I did here: https://github.com/libra/libra/pull/4991/files#diff-f557c4f487f2988c1ebb753153774837R108

execution/executor/src/fuzzing.rs
lightmark: sg

PR #5784, opened 2020-08-26
[RFC][language][vm] take serialized values as arguments

vgao1996:
## Summary
This is an attempt to make the Move VM take serialized values (bytes) as arguments. Early draft/experiment. Not intended to be checked in.


bob-wilson:
We decided not to pursue this for right now, correct?

sblackshear:
I think it would be great to try this again on the `dev` branch once we have one

vgao1996:
@sblackshear Oh yeah definitely! Same for the change from `TransactionEffects` to `ChangeSet` I tried to push. (The two make the VM take and return serialized values respectively.)

PR #5785, opened 2020-08-26
logger: add initial sampling types

bmwill:


common/logger/src/sample.rs
```diff
@@ -0,0 +1,185 @@
...
+use std::{
+    sync::atomic::{AtomicU64, Ordering},
+    time::{Duration, SystemTime},
+};
+
+#[derive(Debug)]
+pub enum SampleRate {
+    Duration(Duration),
+    Frequency(u64),
+    Always,
```
gregnazario, line 15: I would add some comments around this.  Mostly around Frequency e.g. `Frequency(0)` or `Frequency(1)` is the same as `Always`.  `Frequency(2)` is every other message.  We should reject `Frequency(0)` or `Frequency(1)` and have users input `Always`, though I'm not sure we need an `Always`.

common/logger/src/sample.rs
```diff
@@ -0,0 +1,185 @@
...
+        previous_count == 0
+    }
+
+    fn sample_duration(rate: &Duration, last_sample: &AtomicU64) -> bool {
+        let rate = rate.as_secs();
+        // Seconds since Unix Epoch
+        let now = SystemTime::now()
+            .duration_since(SystemTime::UNIX_EPOCH)
+            .expect("SystemTime before UNIX EPOCH!")
+            .as_secs();
```
gregnazario, line 60: So, only by second granularity for sampling makes sense, let's enforce it in the input `Duration` and document it.  Also, I'd like to see if we make sampling `Duration` to restrict a duration of 0 seconds.

common/logger/src/sample.rs
```diff
@@ -0,0 +1,185 @@
...
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn frequency() {
+        // Frequency
+        let sampling = Sampling::new(SampleRate::Frequency(10));
+        let mut v = Vec::new();
+        for i in 0..=25 {
```
gregnazario, line 97: nit: this is the only test that uses `0..=##` vs the others use `0..##`.  Though, I just learned about this now!

davidiw:
I think it might be worth establishing a weighted average for sampling along with a counter on some distinct fields That way we know how many times it wanted to be logged versus how many times it was logged. And we can also say log at most once an hour instead of 1 / x samples. Food for thought.

common/logger/src/sample.rs
bmwill: I agree that we should reject such inputs, unfortunately we probably won't be able to do so until we can do just a bit more in const functions.

PR #5788, opened 2020-08-26
[cli] Derive faucet URL for k8s testnet

mgorven:
Change the faucet URL derivation logic to work with the new k8s testnet.
The faucet is hosted at `/mint` of the same endpoint.

### DIFF
--- a/testsuite/cli/src/client_proxy.rs
+++ b/testsuite/cli/src/client_proxy.rs
@@ -182,11 +182,12 @@ impl ClientProxy {
             Some(dd_account_data)
         };
 
-        let faucet_url = Url::parse(&faucet_url.unwrap_or(format!(
-            "http://{}",
-            url.host_str().unwrap().replace("client", "faucet").as_str()
-        )))
-        .expect("Invalid faucet URL");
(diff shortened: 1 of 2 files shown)


testsuite/cli/src/client_proxy.rs
```diff
@@ -182,11 +182,12 @@ impl ClientProxy {
...
 
-        let faucet_url = Url::parse(&faucet_url.unwrap_or(format!(
-            "http://{}",
-            url.host_str().unwrap().replace("client", "faucet").as_str()
-        )))
-        .expect("Invalid faucet URL");
+        let faucet_url = if let Some(faucet_url) = &faucet_url {
+            Url::parse(faucet_url).expect("Invalid faucet URL specified")
+        } else {
+            url.join("/mint")
```
xli, line 188: this will end with /v1/mint right?

mgorven: It's a URL join, so because it starts with `/` it sets the absolute path. "https://testnet.libra.org/v1" will become "https://testnet.libra.org/mint".

xli (approved):

PR #5792, opened 2020-08-27
cherry picking: [backup] transaction restore: not try to write LedgerInfo from the pr…

msmouse:
…oof to DB.

It was a mistake to write down the LedgerInfo from the chunk proof, because the version of it is most likely larger than the last transaction in the chunk, corrupting the DB (Db sees the version on the latest LedgerInfo as the "committed version", while we don't have every transaction until then.)

Closes: #5727


bmwill:
next time you need to cherry-pick you should try out the `/cherry-pick` command that bors has

msmouse:
@bmwill How can I debug this build failure though...

bmwill:
If there's issues with the cherry-pick then you'll end up having to do the cherry-pick yourself unfortunately :/

PR #5794, opened 2020-08-27
[config] update demo fullnode config

davidiw:
working on documenting how someone might connect an FN to our network...

### DIFF
--- a/config/src/config/test_data/public_full_node.yaml
+++ b/config/src/config/test_data/public_full_node.yaml
@@ -1,17 +1,34 @@
 base:
+    # Update this value to the location you want Libra to store its database
     data_dir: "/opt/libra/data"
     role: "full_node"
     waypoint:
+        # Update this value to that which Libra publicly provides. Please regard the directions
+        # below on how to safely manage your genesis_file_location with respect to the waypoint.
         from_config: "0:01234567890ABCDEFFEDCA098765421001234567890ABCDEFFEDCA0987654210"
 
 execution:
-    genesis_file_location: "relative/path/to/genesis"
+    # Update this to the location to where the genesis.blob is stored, prefer fullpaths
+    # Note, this must be paired with a waypoint. If you update your waypoint without a
+    # corresponding genesis, the file location should be an empty path.
+    genesis_file_location: "/full/path/to/genesis"
+
 
 full_node_networks:
     - discovery_method: "onchain"
-      listen_address: "/ip4/0.0.0.0/tcp/6180"
+      # The network must have a listen address to specify protocols. This runs it locally to
+      # prevent remote, incoming connections.
+      listen_address: "/ip4/127.0.0.1/tcp/6180"
       network_id: "public"
 
+
+rpc:
+    # This specifies your JSON-RPC endpoint. This runs locally to prevent remote queries, setting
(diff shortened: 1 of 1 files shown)


config/src/config/test_data/public_full_node.yaml
```diff
@@ -1,18 +1,33 @@
 base:
+    # Update this value if you would like to connect to a different network
     chain_id: "TESTING"
+    # Update this value to the location you want Libra to store its database
     data_dir: "/opt/libra/data"
     role: "full_node"
     waypoint:
+        # Update this value to that which Libra publicly provides
```
mimoo, line 8: this is great! suggestion: remove "update this value" since this is a configuration and any value can in theory be updated :p I think replacing with an explanation of what the field is might be better? Unless you're doing this specifically for the drills/ceremony

config/src/config/test_data/public_full_node.yaml
```diff
@@ -1,18 +1,33 @@
...
+    # Update this to the location to where the genesis.blob is stored, prefer fullpaths
+    genesis_file_location: "/full/path/to/genesis"
+
 
 full_node_networks:
     - discovery_method: "onchain"
-      listen_address: "/ip4/0.0.0.0/tcp/6180"
+      # The network must have a listen address to specify protocols. This runs it locally to
+      # prevent remote, incoming connections.
+      listen_address: "/ip4/127.0.0.1/tcp/6180"
```
mimoo, line 20: 👍

config/src/config/test_data/public_full_node.yaml
```diff
@@ -1,18 +1,33 @@
...
+    # Update this value to the location you want Libra to store its database
     data_dir: "/opt/libra/data"
     role: "full_node"
     waypoint:
+        # Update this value to that which Libra publicly provides
         from_config: "0:01234567890ABCDEFFEDCA098765421001234567890ABCDEFFEDCA0987654210"
 
 execution:
-    genesis_file_location: "relative/path/to/genesis"
+    # Update this to the location to where the genesis.blob is stored, prefer fullpaths
```
zekun000, line 12: we should also mention if they want to catch up a waypoint without new genesis, set it to ""

davidiw: that's confusing... shouldn't I catch up regardless?

zekun000: how do you differentiate a case when the genesis is set and is supposed to be applied vs catch up?
imagine if the local version is N, waypoint is N+1, and the genesis is set but not update.

davidiw: I imagine some software engineering... can we not detect this in the format for genesis? That it is expected to be applied at Version N and therefore we should have a waypoint of version  > N? We should also be able to detect if a genesis has been applied in the past. If it has been applied, skip it, if it hasn't and the waypoint is bad, we should crash.

zekun000: > We should also be able to detect if a genesis has been applied in the past

This is quite expensive afaiu, also there's no guarantee such a genesis won't be applied again (in theory).

The format of the genesis doesn't tell whether we expect to apply it or not because it's just a transaction unless it's explicitly specified somewhere (either by empty path or specify in the waypoint config (apply or catch up)).

PR #5795, opened 2020-08-27
[config] remove chain_id from config

davidiw:
We can rely on test chain_id in many cases, so ... let's do that. For
the rest of the cases, chain_id is only on the chain, this makes it
clear what we're doing with chain id... or at least less confusing.

### DIFF
--- a/config/config-builder/src/full_node_config.rs
+++ b/config/config-builder/src/full_node_config.rs
@@ -169,7 +169,6 @@ impl FullNodeConfig {
             } else {
                 validator_config.execution.genesis.clone()
             };
-            config.base.chain_id = validator_config.base.chain_id;
             config.base.waypoint = validator_config.base.waypoint.clone();
 
             let network = config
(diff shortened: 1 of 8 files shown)


davidiw:
> so config builder always build with ChainId::test()? I remember before the prod testnet didn't use that one

No, config-builder works as expected? genesis config builder is only used for smoke tests now.

zekun000:
> No, config-builder works as expected? genesis config builder is only used for smoke tests now.

you remove the template there which is how the different chain id being populated iirc.

davidiw:
> > No, config-builder works as expected? genesis config builder is only used for smoke tests now.
> 
> you remove the template there which is how the different chain id being populated iirc.

No, config-builder needs it from the CLI. Passing it via the template won't work even in the current code. The template was only used by the genesis config builder and that's not safe anyway -- so if we want it for that code base, we need to deal with it later.

PR #5796, opened 2020-08-27
log: add support for schema-driven structured logging

bmwill:


consensus/safety-rules/src/logging.rs
```diff
@@ -1,14 +1,105 @@
...
+// TODO implement a proc macro which can generate all of the boilerplate for this struct
+pub struct SafetyLogSchema {
+    name: LogEntry,
+    event: LogEvent,
+    round: Option<Round>,
+    preferred_round: Option<u64>,
+    last_voted_round: Option<u64>,
+    epoch: Option<u64>,
+    error: Option<String>,
+    waypoint: Option<Waypoint>,
```
bmwill, line 18: This is the key part to look at. We can have folks define schemas for their logs and then we can generate the boiler plate below, including all of the setters and the "new" function

bmwill: @davidiw Ideally the "error" field would actually hold a `&'a Error` instead of a string, but I ran into some difficulties with some of the lifetimes that it introduced in some "callback" code in safety_rules.rs

consensus/safety-rules/src/logging.rs
```diff
@@ -1,14 +1,40 @@
...
-use libra_logger::StructuredLogEntry;
+use consensus_types::common::Round;
+use libra_logger::Schema;
+use libra_types::waypoint::Waypoint;
+use serde::Serialize;
 
-pub fn safety_log(entry: LogEntry, event: LogEvent) -> StructuredLogEntry {
-    StructuredLogEntry::new_named("safety_rules", entry.as_str())
-        .data(LogField::Event.as_str(), event.as_str())
+#[derive(Schema)]
```
bmwill, line 9: Added a proc-macro to get rid of the boilerplate.

common/logger/src/lib.rs
```diff
@@ -141,6 +141,7 @@ pub mod prelude {
 }
 pub mod json_log;
 
+mod kv;
```
davidiw, line 166: this whole entry point (crate / module import / re-export) is a mess, care to try to clean it up in one of your future commits / PRs?

common/logger/src/kv.rs
```diff
@@ -0,0 +1,77 @@
...
+        Self(s)
+    }
+
+    pub fn as_str(&self) -> &'static str {
+        self.0
+    }
+}
+
+#[derive(Clone)]
+pub enum Value<'v> {
```
davidiw, line 23: What's the advantage of having a Value with a lifetime versus extracting out the String?

common/logger/src/struct_log.rs
```diff
@@ -86,7 +86,7 @@ pub struct StructuredLogEntry {
 
 impl StructuredLogEntry {
     /// Base implementation for creating a log
-    fn new_unnamed() -> Self {
+    pub fn new_unnamed() -> Self {
```
davidiw, line 89: intentional, this doesn't seem used in this commit.

common/logger/tests/derive.rs
```diff
@@ -0,0 +1,59 @@
+// Copyright (c) The Libra Core Contributors
+// SPDX-License-Identifier: Apache-2.0
+
+use libra_log_derive::Schema;
```
davidiw, line 4: Might it make more sense to make these cargo doc / examples with some text around them on what they're doing?

common/logger/derive/src/lib.rs
```diff
@@ -0,0 +1,179 @@
...
+    let from_debug = quote! { ::libra_logger::Value::from_debug };
+    let key_new = quote! { ::libra_logger::Key::new };
+    let visits = fields.iter().map(|f| {
+        let ident = f.ident.as_ref().unwrap();
+        let ident_str = ident.to_string();
+
+        let from_fn = match f.value_type {
+            Some(ValueType::Display) => &from_display,
+            Some(ValueType::Debug) => &from_debug,
+            Some(ValueType::Serde) | None => &from_serde,
```
davidiw, line 77: is this documented somewhere that the default is from serde?

consensus/safety-rules/src/logging.rs
```diff
@@ -1,14 +1,42 @@
...
+pub struct SafetyLogSchema {
+    name: LogEntry,
+    event: LogEvent,
+    round: Option<Round>,
+    preferred_round: Option<u64>,
+    last_voted_round: Option<u64>,
+    epoch: Option<u64>,
+    #[schema(display)]
+    error: Option<Error>,
```
davidiw, line 19: :O

consensus/safety-rules/src/safety_rules.rs
```diff
@@ -445,19 +452,20 @@ impl TSafetyRules for SafetyRules {
...
-            sl_info!(log_cb(logging::safety_log(log_entry, LogEvent::Success)));
+            sl_info!(log_cb(SafetyLogSchema::new(log_entry, LogEvent::Success)).into_struct_log());
             counters::increment_query(log_entry.as_str(), "success");
             v
         })
         .map_err(|err| {
-            sl_error!(log_cb(logging::safety_log(log_entry, LogEvent::Error))
-                .data(LogField::Message.as_str(), &err));
+            sl_error!(log_cb(SafetyLogSchema::new(log_entry, LogEvent::Error))
+                .error(err.clone())
```
davidiw, line 467: ahh, hahaha

common/logger/src/struct_log.rs
gregnazario: Is there a reason we need the non-categorized version?

common/logger/src/struct_log.rs
```diff
@@ -215,6 +237,14 @@ impl StructuredLogEntry {
         self
     }
 
+    #[doc(hidden)] // set from macro
+    pub fn add_category(&mut self, target: &'static str) -> &mut Self {
+        if self.category.is_none() {
+            self.category = Some(target);
+        }
+        self
+    }
```
gregnazario, line 246: This is slightly different behavior than the other ones.  Is there a reason for that it can't be overridden?

common/logger/src/lib.rs
```diff
@@ -263,6 +289,7 @@ macro_rules! sl_level {
     ($level:expr, $entry:expr) => {
         if $crate::struct_log_enabled!($level) {
             let mut entry = $entry;
+            entry.add_category(module_path!().split("::").next().unwrap());
```
gregnazario, line 292: I was going back and forth on the module, vs category.

This might be good though, the main reason for the category was for something that might be in a different module, but be a different category e.g. security logs.

common/logger/src/lib.rs
```diff
@@ -263,6 +289,7 @@ macro_rules! sl_level {
...
             entry.add_module(module_path!());
```
gregnazario, line 293: If possible, just get the module path once?
```
let module = module_path!();
entry.add_category(...)
entry.add_module(...)
```

common/logger/derive/src/lib.rs
```diff
@@ -0,0 +1,179 @@
...
+                        } else {
+                            panic!("unsupported schema attribute");
+                        };
+
+                        match path.segments.first().unwrap().ident.to_string().as_ref() {
+                            "debug" => return Some(ValueType::Debug),
+                            "display" => return Some(ValueType::Display),
+                            "serde" => return Some(ValueType::Serde),
+                            _ => panic!("unsupported schema attribute"),
+                        }
```
gregnazario, line 171: nit: `return match ...`

common/logger/derive/src/lib.rs
```diff
@@ -0,0 +1,179 @@
...
+                // Only handle schema attrs
+                if segment.ident == "schema" {
+                    for meta in &nested {
+                        let path = if let NestedMeta::Meta(Meta::Path(path)) = meta {
+                            path
+                        } else {
+                            panic!("unsupported schema attribute");
+                        };
+
+                        match path.segments.first().unwrap().ident.to_string().as_ref() {
```
gregnazario, line 166: Maybe we want to case insensitive the match?

common/logger/derive/src/lib.rs
```diff
@@ -0,0 +1,179 @@
...
+    });
+
+    // Build the output, possibly using quasi-quotation
+    let expanded = quote! {
+        impl #impl_generics #name #ty_generics #where_clause {
+            #(#setters)*
+
+            pub fn into_struct_log(self) -> ::libra_logger::StructuredLogEntry {
+                ::libra_logger::StructuredLogEntry::new_unnamed().schema(self)
+            }
```
gregnazario, line 99: I guess we can't implement `into()` directly?

consensus/safety-rules/src/logging.rs
```diff
@@ -1,14 +1,42 @@
...
+use serde::Serialize;
 
-pub fn safety_log(entry: LogEntry, event: LogEvent) -> StructuredLogEntry {
-    StructuredLogEntry::new_named("safety_rules", entry.as_str())
-        .data(LogField::Event.as_str(), event.as_str())
+#[derive(Schema)]
+pub struct SafetyLogSchema {
+    name: LogEntry,
+    event: LogEvent,
+    round: Option<Round>,
```
gregnazario, line 14: Will `Option` fields not be serialized?  I might have missed that.

common/logger/derive/src/lib.rs
bmwill: I'll make sure to document that in the example I added to the docs

common/logger/derive/src/lib.rs
bmwill: I mean we probably can, but I figured we'd just be explicit about it right now. My goal is to eliminate this in the future so folks can just provide the schema and the macro can do the magic conversion itself.

common/logger/derive/src/lib.rs
bmwill: I don't know if that's something that most proc-macros do, we can take a look at serde and see but i'm not overly concerned

common/logger/src/kv.rs
bmwill: I'm trying to have it be a little flexible for things we might want to do in the backend. This way you're not forced to do an allocation unless the backend really wants to, which is a win

common/logger/src/lib.rs
bmwill: Yep that's on my list of things to do, I also want to split this into a crate to be used in libraries (for instrumentation) and a crate which actually implements the logger for binaries. This should allow us to make the dep footprint a little more palatable too

common/logger/src/lib.rs
bmwill: This is the big question I had about all of this. I don't have enough context to know what these "categories" were used for, in most cases it looked like it was just the module name so it just seemed to be extra boilerplate people needed to type. If we can nail down exactly what this is intended to be we can very easily just tack this extra info on to the Schema trait as well

common/logger/src/struct_log.rs
bmwill: If you noticed, this is so that the schema proc-macro can easily create one of these Types and the category is just set by the macro. I'm very open to improvements on this front though

common/logger/src/struct_log.rs
bmwill: I didn't let it overwrite because otherwise the macro would overwrite 100% of these to be just the module-name. So since I didn't have context on its use I just tried to be extra careful.

common/logger/src/tests/struct_log.rs
```diff
@@ -2,16 +2,34 @@
...
+#[derive(serde::Serialize)]
+#[serde(rename_all = "snake_case")]
+enum Enum {
+    FooBar,
+}
+
+struct TestSchema<'a> {
+    foo: usize,
+    bar: &'a Enum,
+}
```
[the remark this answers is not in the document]
bmwill: So the proc-macro can't be tested in a unit test (for reasons I'm happy to go into), but it is being tested in an integration test.

common/logger/tests/derive.rs
bmwill: Good point, I'll add some docs to the tests

consensus/safety-rules/src/logging.rs
bmwill: Yes (I guess more docs are in order). The idea is that, non-optionals will always be logged since they are required, but fields marked as optional will only be logged if it is `Some(T)`

common/logger/src/struct_log.rs
gregnazario: Yeah, this whole thing can be worked around later.

I was planning on removing this "unnamed" vs "named" and just make one.

consensus/safety-rules/src/logging.rs
bmwill: The proc-macro does in fact support lifetimes and I would have preferred to make this `Option<&'a Error>` except @davidiw did some nuts stuff with closures and I was struggling with getting lifetimes to work with the closures so I punted :) 

If you notice some of the tests do test that lifetimes work

consensus/safety-rules/src/safety_rules.rs
bmwill: @gregnazario We don't have to clone things, this is a unique situation just due to the structure of this piece of code. As I mentioned above the proc-macro supports taking things by reference, I just punted on getting the lifetimes to work with these closures.

consensus/safety-rules/src/safety_rules.rs
```diff
@@ -464,7 +459,7 @@ where
         })
         .map_err(|err| {
             sl_error!(log_cb(SafetyLogSchema::new(log_entry, LogEvent::Error))
-                .error(err.clone())
+                .error(&err) //.clone())
```
davidiw, line 462: nit

PR #5798, opened 2020-08-27
[breaking][vm] Merge KeptVMStatus for Verifier/Deserialization errors. 

tnowacki:
- Merge Verifier/Deserialization errors
- Allows room for future error changes without breaking changes
  - We might still have to update the VM or other systems, but historical transactions would still be valid
- This change is breaking only on paper, and should not affect any clients currently. That is because no current transactions on chain should be giving Verifier or Deserialization errors

## Motivation

- Doing this now allows us more flexibility do deal with issues going forward. 
  - If we need to recategorize errors in the VM for some reason, that change would no longer be breaking
  - If we need to update the VM to add some new type of error (distinct from Execution, Verifier, or Deserialization errors), that change would be breaking in the sense that we have to update the system, but it would be backwards compatible 

## Test Plan

- cargo test

### DIFF
--- a/json-rpc/types/src/views.rs
+++ b/json-rpc/types/src/views.rs
@@ -411,12 +411,8 @@ pub enum VMStatusView {
         function_index: u16,
         code_offset: u16,
     },
-    #[serde(rename = "verification_error")]
-    VerificationError,
-    #[serde(rename = "deserialization_error")]
-    DeserializationError,
-    #[serde(rename = "publishing_failure")]
-    PublishingFailure,
+    #[serde(rename = "miscellaneous_error")]
+    MiscellaneousError,
 }
 
 impl From<&KeptVMStatus> for VMStatusView {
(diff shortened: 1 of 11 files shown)


json-rpc/types/src/views.rs
```diff
@@ -371,12 +371,8 @@ pub enum VMStatusView {
...
         code_offset: u16,
     },
-    #[serde(rename = "verification_error")]
-    VerificationError,
-    #[serde(rename = "deserialization_error")]
-    DeserializationError,
-    #[serde(rename = "publishing_failure")]
-    PublishingFailure,
+    #[serde(rename = "uncategorized_program_error")]
+    UncategorizedProgramError,
```
xli, line 375: please update doc https://github.com/libra/libra/blob/master/json-rpc/docs/type_transaction.md#verification_error

and add change to https://github.com/libra/libra/blob/master/json-rpc/API-CHANGELOG.md

tnowacki:
Thanks @xli for the heads up on the docs! Let me know if I missed anything (or if you have feedback on the documentation provided)

tnowacki:
Switched to `MiscellaneousError`

PR #5799, opened 2020-08-27
[Fuzzing] Add fuzz targets for the safety rules message handlers.

JoshLind:
## Motivation

To help achieve better fuzzing coverage for safety rules, this PR adds fuzz targets for each of the safety rules message handling methods. Moreover, it refactors the existing fuzzing target for handle_message() to avoid code duplication and present a cleaner abstraction.

To achieve this, the PR offers seven commits:
1. First, we refactor the existing fuzz target for handle_message() to avoid code duplication.
2. Next, we add a fuzz target for the safety rules initialize() method.
3. We add a fuzz target for the safety rules construct_and_sign_vote() method.
4. We add a fuzz target for the safety rules sign_proposal() method.
5. We add a fuzz target for the safety rules sign_timeout() method.
6. Use consistent naming for corpus generation functions to reduce ambiguity.
7. Updating the fuzzing targets to use the public interface methods of safety rules and merged the fuzzing utils.

Yes.

## Test Plan

All tests pass locally. Moreover, running the fuzzers locally seems to work.

## Related PRs

None.

### DIFF
--- a/consensus/safety-rules/src/lib.rs
+++ b/consensus/safety-rules/src/lib.rs
@@ -27,8 +27,10 @@ pub use crate::{
 #[cfg(any(test, feature = "fuzzing"))]
 pub mod fuzzing_utils;
 
-#[cfg(any(test, feature = "fuzzing", feature = "testing"))]
-#[path = "test_utils.rs"]
+#[cfg(any(test, feature = "fuzzing"))]
+pub use crate::serializer::fuzzing::fuzz_handle_message;
+
+#[cfg(any(test, feature = "testing"))]
 pub mod test_utils;
 
 #[cfg(test)]
--- a/consensus/safety-rules/src/lib.rs
+++ b/consensus/safety-rules/src/lib.rs
@@ -28,7 +28,10 @@ pub use crate::{
 pub mod fuzzing_utils;
 
 #[cfg(any(test, feature = "fuzzing"))]
-pub use crate::{safety_rules::fuzzing::fuzz_initialize, serializer::fuzzing::fuzz_handle_message};
+pub use crate::{
+    safety_rules::fuzzing::{fuzz_construct_and_sign_vote, fuzz_initialize},
+    serializer::fuzzing::fuzz_handle_message,
+};
 
 #[cfg(any(test, feature = "testing"))]
 pub mod test_utils;
--- a/consensus/safety-rules/src/lib.rs
+++ b/consensus/safety-rules/src/lib.rs
@@ -28,7 +28,7 @@ pub use crate::{
 pub mod fuzzing_utils;
 
 #[cfg(any(test, feature = "fuzzing"))]
-pub use crate::serializer::fuzzing::fuzz_handle_message;
+pub use crate::{safety_rules::fuzzing::fuzz_initialize, serializer::fuzzing::fuzz_handle_message};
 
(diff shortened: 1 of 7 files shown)


consensus/safety-rules/src/lib.rs
```diff
@@ -27,8 +27,15 @@ pub use crate::{
 #[cfg(any(test, feature = "fuzzing"))]
 pub mod fuzzing_utils;
 
-#[cfg(any(test, feature = "fuzzing", feature = "testing"))]
-#[path = "test_utils.rs"]
+#[cfg(any(test, feature = "fuzzing"))]
+pub use crate::{
+    safety_rules::fuzzing::{
```
mimoo, line 32: can't you just `pub use crate::safety_rules::fuzzing` ?

consensus/safety-rules/src/safety_rules.rs
```diff
@@ -458,3 +458,38 @@ where
...
+pub mod fuzzing {
+    use crate::{error::Error, test_utils};
+    use consensus_types::{
+        block::Block, block_data::BlockData, timeout::Timeout, vote::Vote,
+        vote_proposal::MaybeSignedVoteProposal,
+    };
+    use libra_crypto::ed25519::Ed25519Signature;
+    use libra_types::epoch_change::EpochChangeProof;
+
+    pub fn fuzz_initialize(proof: EpochChangeProof) -> Result<(), Error> {
```
mimoo, line 474: perhaps refactor these functions to not return anything

consensus/safety-rules/src/safety_rules.rs
```diff
@@ -458,3 +458,38 @@ where
             err
         })
 }
+
+#[cfg(any(test, feature = "fuzzing"))]
+/// This module allows the fuzzer to fuzz the guarded_* request handlers used by safety rules to
+/// handle message requests.
+pub mod fuzzing {
```
davidiw, line 465: Why can't this go into fuzz_utils or fuzzing.rs. This doesn't seem like the right place for it.

consensus/safety-rules/src/serializer.rs
```diff
@@ -118,3 +118,24 @@ impl TSerializerClient for LocalService {
...
     }
 }
+
+#[cfg(any(test, feature = "fuzzing"))]
+/// This module allows the fuzzer to fuzz the handle_message() method exposed by the serializer
+/// service.
+pub mod fuzzing {
+    use crate::{serializer::SafetyRulesInput, test_utils, Error};
+
+    pub fn fuzz_handle_message(safety_rules_input: SafetyRulesInput) -> Result<Vec<u8>, Error> {
```
davidiw, line 128: Same, I'm not a big fan of seeing fuzz code like this :/

consensus/safety-rules/src/safety_rules.rs
JoshLind: As discussed offline, this was not avoid exposing the guarded_*_() methods. For now, I've updated PR to call the public interface methods and take the 20% performance hit for each fuzzing iteration!

consensus/safety-rules/src/serializer.rs
JoshLind: Merged into fuzzing_utils.rs. See above!

PR #5800, opened 2020-08-27
[fuzzing][storage] fix shift overflow bug

lightmark:
## Motivation

When we decode, the input maybe not valid.
thread '<unnamed>' panicked at 'attempt to shift left with overflow', types/src/proof/position/mod.rs:66:27

### DIFF
--- a/storage/libradb/src/schema/transaction_accumulator/mod.rs
+++ b/storage/libradb/src/schema/transaction_accumulator/mod.rs
@@ -35,9 +35,8 @@ impl KeyCodec<TransactionAccumulatorSchema> for Position {
 
     fn decode_key(data: &[u8]) -> Result<Self> {
         ensure_slice_len_eq(data, size_of::<u64>())?;
-        Ok(Position::from_postorder_index(
-            (&data[..]).read_u64::<BigEndian>()?,
-        ))
+        let index = (&data[..]).read_u64::<BigEndian>()?;
+        Ok(Position::from_postorder_index(index)?)
     }
 }
 
--- a/types/src/proof/position/mod.rs
+++ b/types/src/proof/position/mod.rs
@@ -24,6 +24,7 @@
 //! Note3: The leaf index starting from left-most leaf, starts from 0
 
 use crate::proof::definition::{LeafCount, MAX_ACCUMULATOR_LEAVES, MAX_ACCUMULATOR_PROOF_DEPTH};
+use anyhow::{ensure, Result};
 use mirai_annotations::*;
 use std::fmt;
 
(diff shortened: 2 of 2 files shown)


storage/libradb/src/schema/mod.rs
```diff
@@ -54,6 +54,11 @@ fn ensure_slice_len_gt(data: &[u8], len: usize) -> Result<()> {
     Ok(())
 }
 
+fn ensure_accumulator_index_valid(index: u64) -> Result<()> {
+    ensure!(index < 1u64 << 63, "node index {} is invalid (greater than 2^63)", index);
```
msmouse, line 58: I think in theory the biggest index is `2^64 -2` or `(2^63 - 1) * 2`, as in: 

``` text
   6
 2   5
0 1 3 4
```

storage/libradb/src/schema/transaction_accumulator/mod.rs
```diff
@@ -35,9 +37,9 @@ impl KeyCodec<TransactionAccumulatorSchema> for Position {
 
     fn decode_key(data: &[u8]) -> Result<Self> {
         ensure_slice_len_eq(data, size_of::<u64>())?;
-        Ok(Position::from_postorder_index(
-            (&data[..]).read_u64::<BigEndian>()?,
-        ))
+        let key = (&data[..]).read_u64::<BigEndian>()?;
+        ensure_accumulator_index_valid(key)?;
```
mimoo, line 41: why not inline these checks?

storage/libradb/src/schema/mod.rs
```diff
@@ -54,6 +54,11 @@ fn ensure_slice_len_gt(data: &[u8], len: usize) -> Result<()> {
     Ok(())
 }
 
+fn ensure_accumulator_index_valid(index: u64) -> Result<()> {
```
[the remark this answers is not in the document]
lightmark: The problem is maybe functions in `Position` accept `u64` so if we enforced the value limit, it has to be everywhere.

storage/libradb/src/schema/transaction_accumulator/mod.rs
```diff
@@ -35,9 +35,8 @@ impl KeyCodec<TransactionAccumulatorSchema> for Position {
 
     fn decode_key(data: &[u8]) -> Result<Self> {
         ensure_slice_len_eq(data, size_of::<u64>())?;
-        Ok(Position::from_postorder_index(
-            (&data[..]).read_u64::<BigEndian>()?,
-        ))
+        let index = (&data[..]).read_u64::<BigEndian>()?;
+        Ok(Position::from_postorder_index(index)?)
```
msmouse, line 39:
```suggestion
        Position::from_postorder_index(index)
```

types/src/proof/position/mod.rs
```diff
@@ -75,8 +76,13 @@ impl Position {
         self.0
     }
 
-    pub fn from_postorder_index(index: u64) -> Self {
-        Position(postorder_to_inorder(index))
+    pub fn from_postorder_index(index: u64) -> Result<Self> {
+        ensure!(
+            index < (!0u64 >> 2),
```
msmouse, line 81: The math still seems wrong....

probably `index < (!0u64 -1)`?

wqfish: Maybe `u64::MAX` is a bit easier to read than `!0u64`...

PR #5801, opened 2020-08-27
Move rust transaction-builder (generated code) into client/

thefallentree:
In the future, this might be moved to a different repo as part of the rust client.

no functionality changes.


davidiw:
I don't think this would ever be moved out of our code since we have multiple dependencies on it. I don't agree with this PR or the rationale for its move.

thefallentree:
> I don't think this would ever be moved out of libra code since we have multiple dependencies on it. I don't agree with this PR or the rationale for its move.

Which part do you not agree? This PR is only moving generated rust code into client/ directory to make it clear it is part of the client interface, and we can better track it.

I know there is controversy around separating out rust client to another repo, that's up to debate, but doesn't affect this move.

davidiw:
> > I don't think this would ever be moved out of libra code since we have multiple dependencies on it. I don't agree with this PR or the rationale for its move.
> 
> Which part do you not agree? This PR is only moving generated rust code into client/ directory to make it clear it is part of the client interface, and we can better track it.
> 
> I know there is controversy around separating out rust client to another repo, that's up to debate, but doesn't affect this move.

Tracking wasn't mentioned in the PR. That's an interesting topic... I don't have enough insights into this... can we add you as a code owner for the former path? What is it you want to track?

thefallentree:
> have enough insights into this... can we add you as a code owner for the former path? What is it you want to track?

codeowners are for notifications, yeah, can be added anywhere, but by moving this part to client/ directory, the plan is to have some sort of change policy that we can enforce differently than the rest of the codebase.  this generated code is an API that need to be stable to make sure user depending on it will not be broken just because we rename some functions.

matbd:
I believe this code is also used "server-side" (@davidiw would know). No strong opinion otherwise.

thefallentree:
> Regarding the name "client", I believe this code is also used server-side (@davidiw would know). No strong opinion otherwise.

yes, but more importantly, it is a client interface, so it needs to be more stable than other server interfaces.

PR #5806, opened 2020-08-27
[dependabot] use v2 dependabot, and set tags for docker images we build.

rexhoffman:
## Motivation

Use newer version of dependabot with gha/docker/cargo support
Pin docker images with sha256 digests.

Yes
## Test Plan
CI
## Related PRs
None


rexhoffman:
also pinned the docker image to last month's release so we get a pr to update it.
gha check will also result in an update.

PR #5807, opened 2020-08-28
[Network] Avoid using unreachable!() to prevent DOS attacks.

JoshLind:
## Motivation

This PR updates the networking code to avoid using the "unreachable!" macro when something unexpected happens. The issue when using this macro is that the process will panic, causing a shutdown of the node -- thus exposing an DOS attack vector.

Unfortunately, it seems that at least one (maybe more..) of these "unreachable!" statements can be triggered externally. Looking forward we should discuss wider preventions of these kinds of things in entire codebase. This attack vector is not just specific to the networking code.

Yes.

## Test Plan

All tests still pass locally.

## Related PRs

None.

### DIFF
--- a/network/src/interface/mod.rs
+++ b/network/src/interface/mod.rs
@@ -304,7 +304,10 @@ where
                 }
             }
             _ => {
-                unreachable!("Unexpected notification received from Peer actor");
+                warn!(
+                    "Unexpected notification received from Peer actor: {:?}",
+                    notif
+                );
             }
         }
     }
--- a/network/src/peer_manager/mod.rs
+++ b/network/src/peer_manager/mod.rs
@@ -762,9 +762,9 @@ where
                         );
                     }
                 } else {
-                    unreachable!(
-                        "{} Received network event for unregistered protocol",
-                        network_context
+                    debug!(
+                        "{} Received network message for unregistered protocol. Message: {:?}",
+                        network_context, msg,
                     );
                 }
             }
@@ -783,9 +783,9 @@ where
                         );
                     }
                 } else {
-                    unreachable!(
-                        "{} Received network event for unregistered protocol",
-                        network_context
+                    debug!(
+                        "{} Received network rpc request for unregistered protocol. RPC: {:?}",
+                        network_context, rpc_req,
                     );
                 }
             }
--- a/network/src/protocols/direct_send/mod.rs
+++ b/network/src/protocols/direct_send/mod.rs
@@ -163,7 +163,7 @@ impl DirectSend {
                     error!("Unexpected message from peer actor: {:?}", message);
                 }
             }
-            _ => unreachable!("Unexpected PeerNotification"),
+            _ => warn!("Unexpected PeerNotification: {:?}", notif),
         }
     }
 


network/src/interface/mod.rs
```diff
@@ -304,7 +304,7 @@ where
                 }
             }
             _ => {
-                unreachable!("Unexpected notification received from Peer actor");
+                warn!("Unexpected notification received from Peer actor");
             }
```
gregnazario, line 308: Hehe... yeah that's not great.

Can you put the `notif` in the message?

```
warn!("Unexpected notification received from Peer actor", notification=notif);
```

network/src/peer_manager/mod.rs
```diff
@@ -762,7 +762,7 @@ where
                         );
                     }
                 } else {
-                    unreachable!(
+                    warn!(
                         "{} Received network event for unregistered protocol",
                         network_context
                     );
```
gregnazario, line 768: Similarly with adding the offending types, it should help with debugging later :)

network/src/peer_manager/mod.rs
```diff
@@ -762,7 +762,7 @@ where
                         );
                     }
                 } else {
-                    unreachable!(
+                    warn!(
```
davidiw, line 765: I don't even know if this is worthy of a log level of anything more than debug... Considering that this could be a means for an attacker to spam our fullnodes and burn through precious logging traffic.

network/src/interface/mod.rs
```diff
@@ -304,7 +304,7 @@ where
                 }
             }
             _ => {
-                unreachable!("Unexpected notification received from Peer actor");
+                warn!("Unexpected notification received from Peer actor");
```
davidiw, line 307: should we log the notif if it supports debuggable?

network/src/protocols/direct_send/mod.rs
```diff
@@ -163,7 +163,9 @@ impl DirectSend {
                     error!("Unexpected message from peer actor: {:?}", message);
                 }
             }
-            _ => unreachable!("Unexpected PeerNotification"),
+            PeerNotification::PeerDisconnected(metadata, reason) => {
```
davidiw, line 166: why not just log the notif if it supports debuggable?

network/src/peer_manager/mod.rs
JoshLind: Good point! Done, thanks 😄

JoshLind:
Thanks, all! 😄 

> are they really unreachable or not? it's still worth keeping real ones

@zekun000, I can confirm that the ones in peer_manager are reachable. It doesn't seem like direct_send is reachable (unless there exists a bug internally -- I haven't verified all internal callers). Haven't 100% confirmed/exploited interface, but from what I can tell it seems reachable.

Regarding "keeping real ones", I think this is worth discussion. The issue I see here is that with all of these cases, the logic checks some state of a network request. If the request is malformed or unexpected, what benefit is there to crashing? My mental model is that of mimicking a web server/request handler -- if a request is bad, drop it and continue. In general, I can see that there will be cases where internal state is so bad that some critical assumption has been broken, hence a crash makes sense. But I don't think that applies here.

I'm going to land (to wrap this up), but happy to revert the direct_send ones if you see a reason why it's better to crash than to continue in this case. This issue is definitely worth discussion.

JoshLind:
Wow.. things are super flaky today 😕

network/src/interface/mod.rs
```diff
@@ -304,7 +304,10 @@ where
                 }
             }
             _ => {
-                unreachable!("Unexpected notification received from Peer actor");
+                warn!(
+                    "Unexpected notification received from Peer actor: {:?}",
+                    notif
+                );
```
mimoo, line 310: can you actually use a `MUST_FIX` structured log? https://github.com/libra/libra/blob/b4061ca3443ed0c512c677a5a9ef755578b83d15/execution/executor/src/lib.rs#L189

mimoo: and I think error is better also (cc @gregnazario )

mimoo: (same for others)

mimoo: oh I realize that I'm too late : c

JoshLind: @mimoo, no worries! If you feel strongly about the logging type and the MUST_FIX, feel free to create a PR that handles it and send it across to us. I went for the logging types recommended by @davidiw -- noting the log spamming issue with higher levels. Others might have other thoughts?

mimoo: so I think we need to create some guidance and tooling around these type of errors. My idea was that if we can create some detection rules on MUST_FIX logs, then we'll notice them faster in the sea of logging that we'll have to monitor.

mimoo: also if it's a MUST_FIX, we should worry about that rather than the spamming (unless the spamming leads to a DoS ofc)

gregnazario: IMO, if it's a MUST_FIX, shouldn't we just fix it?

If we're worried about having an "unreachable" macro that logs when something unexpected happens, but not panic, we could build that into the logging system.

mimoo: yeah it would be cool to have a must_fix!() macro that just emits a slog of category "MUST_FIX"

PR #5808, opened 2020-08-28
[genesis][breaking] support distinct keys for treasury root libra root in genesis

davidiw:
## Breaking nature

The root genesis transaction's format will change. This will break previous genesis transactions.

## Justification

LibraRoot and TreasuryRoot should not have the same key at genesis time, hence this PR allows the accounts to be created with distinct keys.

## Details

This is the first PR that causes breaking changes. The rest will follow after, since it is more nuanced tooling changes.


language/stdlib/modules/Genesis.move
```diff
@@ -28,9 +28,10 @@ module Genesis {
 
     fun initialize(
         lr_account: &signer,
+        lr_auth_key: vector<u8>,
```
tzakian, line 31: nit: we try to have all the signers as the first argument, so could you move this to be right next to the `tc_auth_key`?

davidiw:
Next step is to add support for this in the genesis tooling, but that's not a breaking change as we've defined breaking changes.

PR #5814, opened 2020-08-28
[Secure Storage] Remove "unimplemented!()" as default implementation for CryptoStorage.

JoshLind:
## Motivation

This PR removes the "unimplemented!()" macro from the default implementation of "import_private_key" for CryptoStorage. Right now it seems that all CryptoStorage implementations use either their own implementation of "import_private_key" (e.g., vault), or the default implementation provided by CryptoKVStorage. As a result, we don't need this default implementation. Moreover, we probably want to avoid providing a default that is not implemented, and instead force the compiler to tell the user that they should provide an implementation for this.

Yes.

## Test Plan

All test still pass locally.

## Related PRs

None.

### DIFF
--- a/secure/storage/src/crypto_storage.rs
+++ b/secure/storage/src/crypto_storage.rs
@@ -21,9 +21,7 @@ pub trait CryptoStorage {
     /// not used correctly. As this is purely a testing API, there is no defined behavior for
     /// importing a key for a given name if that name already exists.  It only exists to allow
     /// Libra to be run in test environments where a set of deterministic keys must be generated.
-    fn import_private_key(&mut self, _name: &str, _key: Ed25519PrivateKey) -> Result<(), Error> {
-        unimplemented!();
-    }
+    fn import_private_key(&mut self, _name: &str, _key: Ed25519PrivateKey) -> Result<(), Error>;
 
     /// Returns the Ed25519 private key stored at 'name' and identified by 'version', which is the
     /// corresponding public key. This may fail even if the 'named' key exists but the version is


JoshLind:
Thanks, David! 😄

davidiw:
I think we should think about this a bit more... there are APIs that are unsafe... like importing and exporting private keys. In fact, many enclaves do not allow for it. Placing unimplemented at the highest levels makes it really clear that this should not be assumed to exist and is dangerous.

JoshLind:
> I think we should think about this a bit more... there are APIs that are unsafe... like importing and exporting private keys. In fact, many enclaves do not allow for it. Placing unimplemented at the highest levels makes it really clear that this should not be assumed to exist and is dangerous.

Thanks, David. That's a fair point. Although, I'm wondering if there is a better way to communicate/enforce/handle this. For example, if the API is meant to be testing/fuzzing only, we should wrap it in compilation flags. If the API is considered to be that unsafe/dangerous, we should not expose it all and refactor the calling code to work around the limitation. If however, we really need to expose the method for some really weird edge case, we should make it even more explicit, e.g., wrap the method in a "dangerous" feature, or at the very least add a comment. Without something that says "We're doing this because we don't want you to use this method", it's difficult for the reader to know why we're using unimplemented.

Note, in this case, all storage implementations already provide an implementation for this method, so the only benefit we get by using "unimplemented" at the trait is to inform *new* storage implementations.. but, again, without the comment I'm not sure the message comes across. Thoughts?

PR #5817, opened 2020-08-28
[fuzzing] downgrade version of libfuzzer for lionhead to work

mimoo:


rexhoffman:
Ah, my bad, thanks @mimoo

rexhoffman:
@mimoo this won't work, the lec summary needs updated.

PR #5824, opened 2020-08-29
[storage] fix warning log in get_account_state_with_proof

msmouse:
## Motivation
1. version <= ledger_version should be an ensure!()
2. clarify warning message under the `if ledger_version > commited_version` condition

Y
## Test Plan
existing coverage

## Related PRs

blaming #5797

### DIFF
--- a/storage/libradb/src/lib.rs
+++ b/storage/libradb/src/lib.rs
@@ -687,18 +687,20 @@ impl DbReader for LibraDB {
             .with_label_values(&["get_account_state_with_proof"])
             .start_timer();
 
-        if version <= ledger_version {
-            warn!(
-                "The queried version {} should be equal to or older than ledger version {}.",
-                version, ledger_version
+        ensure!(
+            version <= ledger_version,
+            "The queried version {} should be equal to or older than ledger version {}.",
+            version,
+            ledger_version
+        );
+        {
+            let latest_version = self.get_latest_version()?;
+            ensure!(
+                ledger_version <= latest_version,
+                "ledger_version specified {} is greater than committed version {}.",
+                ledger_version,
+                latest_version
             );
-        }
-        let latest_version = self.get_latest_version()?;
-        if ledger_version <= latest_version {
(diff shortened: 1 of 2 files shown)


storage/libradb/src/lib.rs
```diff
@@ -687,18 +687,20 @@ impl DbReader for LibraDB {
...
+        ensure!(
+            version <= ledger_version,
+            "The queried version {} should be equal to or older than ledger version {}.",
+            version,
+            ledger_version
+        );
+        {
+            let latest_version = self.get_latest_version()?;
+            if ledger_version > latest_version {
+                warn!(
```
davidiw, line 699: would be good to start leveraging structured logs in storage

would this result in a client breaking change? I guess kinda... but perhaps they were using it in ... poor taste

also I would caution that if this interface is exposed to clients, we probably do not want warn logs, because it can be used to dos our logging backend :/

storage/libradb/src/lib.rs
```diff
@@ -687,18 +687,20 @@ impl DbReader for LibraDB {
...
-                "The ledger version {} is greater than the latest version currently in ledger: {}",
-                ledger_version, latest_version
-            )
+        ensure!(
+            version <= ledger_version,
+            "The queried version {} should be equal to or older than ledger version {}.",
+            version,
+            ledger_version
+        );
+        {
```
davidiw, line 696: curious why you nested / scoped this.

msmouse: `latest_version` is for this warning purpose only, I'm trying to prevent it being (mis)used in the logic below.

storage/libradb/src/lib.rs
msmouse: That's a great argument. Hold on.

msmouse: Realized there's already a "cheating" API used internally to ignore the ledger_version check. will squash and update the commit info  before landing.

PR #5830, opened 2020-08-29
[fix] fix shift error when fuzzing accumulator schema

lightmark:
fix https://github.com/libra/libra/pull/5800

### DIFF
--- a/types/src/proof/position/mod.rs
+++ b/types/src/proof/position/mod.rs
@@ -61,10 +61,10 @@ impl Position {
 
     /// pos count start from 0 on each level
     pub fn from_level_and_pos(level: u32, pos: u64) -> Self {
-        precondition!(level < 63);
+        precondition!(level < 64);
         assume!(1u64 << level > 0); // bitwise and integer operations don't mix.
         let level_one_bits = (1u64 << level) - 1;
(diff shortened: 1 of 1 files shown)


types/src/proof/position/mod.rs
```diff
@@ -61,10 +61,10 @@ impl Position {
 
     /// pos count start from 0 on each level
     pub fn from_level_and_pos(level: u32, pos: u64) -> Self {
-        precondition!(level < 63);
+        precondition!(level < 64);
         assume!(1u64 << level > 0); // bitwise and integer operations don't mix.
         let level_one_bits = (1u64 << level) - 1;
-        let shifted_pos = pos << (level + 1);
+        let shifted_pos = pos.wrapping_shl(level + 1);
```
msmouse, line 67: Could you elaborate the reason for this change? I don't see it making any difference for any valid input; if it's invalid input, I think panicking is better than swallowing the error?

lightmark: if level == 63, the normal << would throw overflow error. but level 63 is valid.

msmouse: K. It'll make more sense if we panic or error out on a pos too large for the level, but your argument holds.

PR #5841, opened 2020-08-31
[Secure Storage] Fix slicing error identified by fuzzing target in vault client.

JoshLind:
## Motivation

This PR fixes a slicing error identified by our fuzzer for "process_transit_export_response()" in the vault client. 

Previously, the vault client code was simply taking a given response and truncating the first 32 bytes from it to produce a key. However, if less than 32 bytes were given to the client, it would panic with a slicing error. Instead, this PR updates the code to first check that the length is sufficient, before slicing. If not, an error is returned. 

In this PR, we also add the previously failing input to the libra-fuzzer artifacts, to ensure these failures don't occur again.

Yes.

## Test Plan

After this failure was identified, I added the failing input to the set of artifacts and re-ran the artifact tests to confirm the code was still failing. I then implemented the fix, re-ran the tests and found that the vault client no longer failed.

## Related PRs

None


secure/storage/vault/src/lib.rs
```diff
@@ -502,6 +500,20 @@ pub fn process_transit_export_response(
...
+fn get_ed25519_private_key_from_composite(composite_key: &str) -> Result<Ed25519PrivateKey, Error> {
+    let composite_key = base64::decode(composite_key)?;
+    if composite_key.len() < ED25519_PRIVATE_KEY_LENGTH {
+        return Err(Error::InternalError(
+            "Insufficient key length returned by vault export key request".into(),
+        ));
+    }
+    Ok(Ed25519PrivateKey::try_from(
+        &composite_key[..ED25519_PRIVATE_KEY_LENGTH],
+    )?)
```
mimoo, line 514: This looks clearer to me:

```rust
let composite_key = base64::decode(composite_key);
if let Some(composite_key) = composite_key.get(0.. ED25519_PRIVATE_KEY_LENGTH) {
  Ed25519PrivateKey::try_from(&composite_key)
} else {
 Err(...)
}
```

mimoo: or even better imo, refactor the original function to avoid the duplicate code:

```rust
if resp.not_ok() {
  return Err(resp.into());
}

let export_key: ExportKeyResponse = serde_json::from_str(&resp.into_string()?)?;
let key = if Some(version) = version {
    let key = export_key.data.keys.iter().find(|(k, _v)| **k == version);
    let (_, key) = key.ok_or_else(|| Error::NotFound("transit".into(), name.into()))?;
    key
} else if let Some(key) = export_key.data.keys.values().last() {
    key
} else {
    return Err(...);
}

let composite_key = base64::decode(composite_key)?;
if let Some(composite_key) = composite_key.get(0..ED25519_PRIVATE_KEY_LENGTH) {
  Ed25519PrivateKey::try_from(&composite_key)
} else {
  Err(...)
}
```

JoshLind:
Thanks, @mimoo! I've refactored the method to avoid having the duplication and the function. Regarding the suggested change to remove the `?` from the return statement, e.g.,
 
`
Ok(Ed25519PrivateKey::try_from(
        &composite_key[..ED25519_PRIVATE_KEY_LENGTH],
    )?)
`

this is not possible because try_from returns a `Result<Ed25519PrivateKey, CryptoMaterialError>`, while our function returns a `Result<Ed25519PrivateKey, Error>`. Hence we need the `?`, to do the automatic conversion.. It's a little ugly but better than doing a manual error conversion.

PR #5844, opened 2020-08-31
[logging] add unexpected macro for unexpected behavior

gregnazario:
Rather than panic with a unreachable! this provides a macro to log
unexpected behavior and continue running.

This seemed to come up in some of the other PRs around the discussion around `unreachable!`. This provides a macro to have an `unreachable!` like feeling, without panics.

### DIFF
--- a/common/logger/src/lib.rs
+++ b/common/logger/src/lib.rs
@@ -136,7 +136,8 @@ pub mod prelude {
     pub use crate::{
         debug, error, event, info,
         security::{security_events, security_log},
-        sl_debug, sl_error, sl_info, sl_level, sl_trace, sl_warn, trace, warn, StructuredLogEntry,
+        sl_debug, sl_error, sl_info, sl_level, sl_trace, sl_warn, trace, unexpected, warn,
+        StructuredLogEntry,
     };
 }
 pub mod json_log;
@@ -206,6 +207,16 @@ macro_rules! warn {
 macro_rules! struct_log_enabled {
     ($level:expr) => {
         $crate::struct_logger_enabled($level)
+    };
+}
+
+/// For logging unexpected behavior rather than panicking with `unreachable!`
+#[macro_export]
+macro_rules! unexpected {
+    ($($arg:tt)+) => {
+        let mut entry = $crate::StructuredLogEntry::new_named("unexpected","assertion");
+        $crate::format_struct_args_and_pattern!(entry, $($arg)+);
+        $crate::sl_error!(entry)
     };
 }
 
--- a/execution/executor/src/lib.rs
+++ b/execution/executor/src/lib.rs
@@ -186,8 +186,7 @@ where
         let first_txn_version = match txn_list_with_proof.first_transaction_version {
             Some(tx) => tx as Version,
             None => {
-                sl_error!(StructuredLogEntry::new_named("MUST_FIX", "assertion")
-                    .data("details", "first_transaction_version should exist."));
+                unexpected!("first_transaction_version should exist.");
                 return Err(anyhow!("first_transaction_version should exist."));
             }
         };
--- a/network/src/interface/mod.rs
+++ b/network/src/interface/mod.rs
@@ -304,7 +304,7 @@ where
                 }
             }
             _ => {
-                warn!(
+                unexpected!(
                     "Unexpected notification received from Peer actor: {:?}",
                     notif
                 );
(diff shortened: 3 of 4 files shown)


common/logger/src/lib.rs
```diff
@@ -209,6 +210,17 @@ macro_rules! struct_log_enabled {
     };
 }
 
+/// For logging unexpected behavior rather than panicking with `unreachable!`
+#[macro_export]
+macro_rules! unexpected {
+    ($($arg:tt)+) => {
+        let mut entry = $crate::StructuredLogEntry::new_named("unexpected","assertion");
+        entry = entry.data("MUST_FIX", true);
```
mimoo, line 218: I'd say either call the category MUST_FIX or let's roll with unexpected instead : o but I don't think we need to have an additional data field

execution/executor/src/lib.rs
```diff
@@ -186,8 +186,7 @@ where
         let first_txn_version = match txn_list_with_proof.first_transaction_version {
             Some(tx) => tx as Version,
             None => {
-                sl_error!(StructuredLogEntry::new_named("MUST_FIX", "assertion")
-                    .data("details", "first_transaction_version should exist."));
+                unexpected!("first_transaction_version should exist.");
```
mimoo, line 189: 👍

common/logger/src/lib.rs
gregnazario: Okay, then will roll with unexpected.

gregnazario:
> That's cool! How would we go about creating detection rules for these?

You can make a detection rule based on the category in Kibana to query for the category, and raise alarm when it occurs.

network/src/interface/mod.rs
```diff
@@ -304,7 +304,7 @@ where
                 }
             }
             _ => {
-                warn!(
+                unexpected!(
```
bmwill, line 307: I think that this is probably not better than using a warn in this case. My main concern which the addition of new macros like this is, what does unexpected do? most folks know what panic and friends do as well as log and friends but since unexpected is new its hard to tell without looking at the macro definition itself to determine what's happening.

bmwill:
I think before merging this it would probably be worth having a discussion with @JoshLind and @davidiw so that we can have a unified plan on how we want to address this problem of dealing with panics. That might include this macro, but given how hot a topic this has been lately its probably worth everyone being on the same page before moving forward.

davidiw:
My gut feeling is that if we're adding something that isn't intuitive or somewhat duplicitous we should have some sort of proposal and get feedback from folks before moving forward.... and I've been considering trying to write down my ideas in this space, but they are all too messy. Unfortunately for you, there's a PR here begging for me to share a little bit :P

* What does unexpected mean? When do I use it? What are the consequences?
* Who instigated the unexpected event? What internally is the right behavior? If it is a client interface and the calls are cheap, do I really want this becoming a > debug log level, wherein an attacker can waste precious cpu? Should it be a counter that is cheap and intuitive.
* If it is a counter, how do I distinguish between the potentially dozes of different unexpected calls?
* If I'm running this in tests, can this be a panic with a **sexy** backtrace so I can figure out why I even got there
* Can we ensure that CI wouldn't land commits that violate unexpected?

With that being said, I am super happy that we're all thinking about these problems :).

gregnazario:
> My gut feeling is that if we're adding something that isn't intuitive or somewhat duplicitous we should have some sort of proposal and get feedback from folks before moving forward.... and I've been considering trying to write down my ideas in this space, but they are all too messy. Unfortunately for you, there's a PR here begging for me to share a little bit :P
> 
> * What does unexpected mean? When do I use it? What are the consequences?
> * Who instigated the unexpected event? What internally is the right behavior? If it is a client interface and the calls are cheap, do I really want this becoming a > debug log level, wherein an attacker can waste precious cpu? Should it be a counter that is cheap and intuitive.
> * If it is a counter, how do I distinguish between the potentially dozes of different unexpected calls?
> * If I'm running this in tests, can this be a panic with a **sexy** backtrace so I can figure out why I even got there
> * Can we ensure that CI wouldn't land commits that violate unexpected?
> 
> With that being said, I am super happy that we're all thinking about these problems :).

No problem here.  Just was trying to consolidate these weird things we're doing now.  I'll let you guys discuss.

gregnazario:
Closing this, as it won't be pushed

PR #5849, opened 2020-08-31
[breaking] Update JSON-RPC service configuration field name and default address

xli:
1. Switch from opt-in to opt-out strategy in this regard. Enable JSON-RPC only on 127.0.0.1 and allow users to switch to 0.0.0.0 if they desire.
2. Renamed config name from "rpc" to "json_rpc"

--------------

This PR breaks on canary test because of renaming config field from rpc to json_rpc, after landed, canary test should be fine.

### DIFF
--- a/config/src/config/json_rpc_config.rs
+++ b/config/src/config/json_rpc_config.rs
@@ -7,22 +7,23 @@ use std::net::SocketAddr;
 
 #[derive(Clone, Debug, Deserialize, PartialEq, Serialize)]
 #[serde(default, deny_unknown_fields)]
-pub struct RpcConfig {
+pub struct JsonRpcConfig {
     pub address: SocketAddr,
     pub batch_size_limit: u16,
     pub page_size_limit: u16,
     pub content_length_limit: usize,
 }
 
+pub const DEFAULT_JSON_RPC_ADDRESS: &str = "127.0.0.1";
 pub const DEFAULT_JSON_RPC_PORT: u16 = 8080;
 pub const DEFAULT_BATCH_SIZE_LIMIT: u16 = 20;
 pub const DEFAULT_PAGE_SIZE_LIMIT: u16 = 1000;
 pub const DEFAULT_CONTENT_LENGTH_LIMIT: usize = 32 * 1024; // 32kb
 
-impl Default for RpcConfig {
-    fn default() -> RpcConfig {
-        RpcConfig {
-            address: format!("0.0.0.0:{}", DEFAULT_JSON_RPC_PORT)
+impl Default for JsonRpcConfig {
+    fn default() -> JsonRpcConfig {
+        JsonRpcConfig {
+            address: format!("{}:{}", DEFAULT_JSON_RPC_ADDRESS, DEFAULT_JSON_RPC_PORT)
                 .parse()
                 .unwrap(),
             batch_size_limit: DEFAULT_BATCH_SIZE_LIMIT,
@@ -32,7 +33,7 @@ impl Default for RpcConfig {
     }
 }
 
-impl RpcConfig {
+impl JsonRpcConfig {
     pub fn randomize_ports(&mut self) {
         self.address.set_port(utils::get_available_port());
     }
--- a/config/src/config/mod.rs
+++ b/config/src/config/mod.rs
@@ -31,8 +31,8 @@ mod mempool_config;
 pub use mempool_config::*;
 mod network_config;
 pub use network_config::*;
-mod rpc_config;
-pub use rpc_config::*;
+mod json_rpc_config;
+pub use json_rpc_config::*;
 mod secure_backend_config;
 pub use secure_backend_config::*;
 mod state_sync_config;
@@ -73,7 +73,7 @@ pub struct NodeConfig {
     #[serde(default)]
     pub mempool: MempoolConfig,
     #[serde(default)]
-    pub rpc: RpcConfig,
+    pub json_rpc: JsonRpcConfig,
     #[serde(default)]
     pub state_sync: StateSyncConfig,
     #[serde(default)]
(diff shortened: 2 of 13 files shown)


ankushagarwal:
How do we opt-in to run on 0.0.0.0?

mimoo:
I think this needs to be configurable, since this'll most likely run in docker containers on our side : o

xli:
Address is configable, what I change is default value.

mimoo:
just want to make sure this is not going to break anything :D @davidiw do you know if anything uses the default value here when it should specify 0.0.0.0 instead?

davidiw:
Please work with @mgorven to enable this, specifically, you'll need to go through the deployment configs and make sure it is configured to the current default.

mimoo:
@davidiw we should allow configuration to disable this interface as well, especially if you're a validator : o do we want to create an issue for this?

davidiw:
While on the surface this looks like a great idea but actually isn't as valuable:

1) It only provides denial of service attacks against folks that somehow are capable enough to deploy a service on a public ip but not capable enough to know how to restrict access. Mind you it isn't really that easy to get a ipv4 address on a machine these days.
2) This service does not give sensitive access to clients, so the worst case scenario is dos
3) There is operational risk to this change. Configuration changes are a nightmare and the test coverage for them can be sparse. Internally we have 4 or more repositories that leverage these configs, who knows what the values are across those. Then we have a community that might already be leveraging these defaults and be utterly confused why their service no longer works.

mimoo:
Just as a counterpoint:

> It only provides denial of service attacks against folks that somehow are capable enough to deploy a service on a public ip but not capable enough to know how to restrict access. Mind you it isn't really that easy to get a ipv4 address on a machine these days.

It sounds like you're saying that we made it to complex for anyone to run this without great expertise in how to properly configure a host to run the node, I don't think this is necessarily true (we are only talking about a PFN here) and I hope that we will work towards not making it true (e.g. if we let anyone sync directly to the latest state via a waypoint).

> This service does not give sensitive access to clients, so the worst case scenario is dos

Imagine that you're running a small wallet iOS app, cater to some users, and someone figures out that you let your JSON-RPC endpoint unprotected: they just DoS'ed your app.

> There is operational risk to this change. Configuration changes are a nightmare and the test coverage for them can be sparse. Internally we have 4 or more repositories that leverage these configs, who knows what the values are across those. Then we have a community that might already be leveraging these defaults and be utterly confused why their service no longer works.

I'd argue that there is even more operational risk to not apply this change, as it'll get harder and harder and harder to apply it as more people use these unsafe defaults. We're creating security debt.

aching:
> While on the surface this looks like a great idea but actually isn't as valuable:
> 
> 1. It only provides denial of service attacks against folks that somehow are capable enough to deploy a service on a public ip but not capable enough to know how to restrict access. Mind you it isn't really that easy to get a ipv4 address on a machine these days.
> 2. This service does not give sensitive access to clients, so the worst case scenario is dos
> 3. There is operational risk to this change. Configuration changes are a nightmare and the test coverage for them can be sparse. Internally we have 4 or more repositories that leverage these configs, who knows what the values are across those. Then we have a community that might already be leveraging these defaults and be utterly confused why their service no longer works.

Just to make sure I am understanding correctly...1 & 2 are good things but not absolutely necessary at this time in your opinion.  Regarding 3, I agree that those who have used these setting could be surprised in the change, but this argument gets worse as time progresses as the community develops, so if we intend to make this change, it's better to do now rather than later.

I'll add 4.  Today, the risk exposed is DoS, tomorrow, it could be more depending on how the API changes, so that's something to keep in mind.  In general, we'll want to audit all externally run interfaces that are defaults more carefully than those that are opt-in.

davidiw:
> Imagine that you're running a small wallet iOS app, cater to some users, and someone figures out that you let your JSON-RPC endpoint unprotected: they just DoS'ed your app.

You're running our FN on an iPhone? That sounds like a recipe for disaster <3

mimoo:
that is not what I said xD

davidiw (commented):
This is beautiful!

davidiw (approved):
/land

let's do it!

dmitri-perelman:
Could you guys help me with proper new configuration here?
AOS is using the docker files for running the local validator nodes.
I used to run a validator like this:
```
docker pull libra/validator:master_c8192377
docker run -d --name validator --entrypoint 'bash' -p 8081:8080 docker.io/libra/validator:master_c8192377 -c '/opt/libra/bin/libra-node --test --config /opt/libra/var &>/dev/null'
```
and be able to json rpc to it on port 8081, but not anymore after this PR. 
What should we change?

davidiw:
> Could you guys help me with proper new configuration here?
> AOS is using the docker files for running the local validator nodes.
> I used to run a validator like this:
> 
> ```
> docker pull libra/validator:master_c8192377
> docker run -d --name validator --entrypoint 'bash' -p 8081:8080 docker.io/libra/validator:master_c8192377 -c '/opt/libra/bin/libra-node --test --config /opt/libra/var &>/dev/null'
> ```
> 
> and be able to json rpc to it on port 8081, but not anymore after this PR.
> What should we change?

Ayyyaaaaa, this is quite the dilemma ...
So clearly we'll need to backport a fix to 1.0 branch cut...

In the mean time, my reading of this: https://stackoverflow.com/questions/30109037/how-can-i-forward-localhost-port-on-my-container-to-localhost-on-my-host

suggests you can do --net=host but that will set it to port 8080 on the localhost

the proper fix is to tweak the test node code so that it exposes the service on 0.0.0.0 before this breaking change...  I'll send out a PR for master within the next 12 hours, but it will be a bit longer until it hits 1.0.

What is your timeline? If I need to move faster, let me know.

Deeply sorry for the inconvenience here :(

dmitri-perelman:
JFYI, PR https://github.com/libra/libra/pull/6542 has fixed the issue.
Thanks David!
(we'll work on cherry-picking next week)

PR #5850, opened 2020-08-31
some cleanups to the coding guidelines

davidiw:


documentation/coding_guidelines.md
```diff
@@ -207,7 +207,9 @@ Generics allow dynamic behavior (similar to [`trait`](https://doc.rust-lang.org/
 
 Excluding test code, set field visibility to private as much as possible. Private fields allow constructors to enforce internal invariants. Implement getters for data that consumers may need, but avoid setters unless a mutable state is necessary.
 
-Public fields are most appropriate for [`struct`](https://doc.rust-lang.org/book/ch05-00-structs.html) types in the C spirit: compound, passive data structures without internal invariants.  Naming suggestions follow the guidance [here](https://rust-lang-nursery.github.io/api-guidelines/naming.html#getter-names-follow-rust-convention-c-getter) as shown below.
+Getters/setters should be avoided for [`struct`](https://doc.rust-lang.org/book/ch05-00-structs.html) types in the C spirit: compound, passive data structures without internal invariants. Adding them only increases the complexity and number of lines of code without improving the developer experience.
```
bmwill, line 210: I sort of just want to cut this paragraph, I don't think it ads much value imo

documentation/coding_guidelines.md
```diff
@@ -254,7 +254,28 @@ We currently use [log](https://docs.rs/log/) for logging.
...
+#[cfg(test)]
+mod tests {
+  #test
+  fn verify_magic_number() {
+    assert_eq!(Foo::magic_number(), 42);
+  }
+}
+```
+
+It is also acceptable to place test files into the same directory as `mod.rs` or `lib.rs` so long as their file names end in `_test.rs`. `#[cfg(test)]` should guard tests from release builds. For example, if in a crate there is a db module, the expected directory structure is as follows:
```
bmwill, line 278: We should remove this paragraph and discourage this type of structure.

davidiw: Fair enough, we don't need the struct, the intent was to show code and tests co-located.

documentation/coding_guidelines.md
davidiw: Unfortunately I see a lot of folks writing get_ functions -- maybe 897 in the code base :O

documentation/coding_guidelines.md
```diff
@@ -167,9 +167,7 @@ We recommend that you use `//` and `///` comments rather than block comments `/*
 
 ### Cloning
 
-If `x` is reference counted, prefer [`Arc::clone(x)`](https://doc.rust-lang.org/std/sync/struct.Arc.html) over `x.clone()`. [`Arc::clone(x)`](https://doc.rust-lang.org/std/sync/struct.Arc.html) explicitly indicates that we are cloning `x`. This avoids confusion about whether we are performing an expensive clone of a `struct`, `enum`, other types, or just a cheap reference copy.
-
-Also, if you are passing around [`Arc<T>`](https://doc.rust-lang.org/std/sync/struct.Arc.html) types, consider using a newtype wrapper:
+If you are passing around [`Arc<T>`](https://doc.rust-lang.org/std/sync/struct.Arc.html) types, consider using a newtype wrapper:
```
mimoo, line 170: why btw?

documentation/coding_guidelines.md
```diff
@@ -199,6 +197,8 @@ Error handling suggestions follow the [Rust book guidance](https://doc.rust-lang
 * `assert!()` - This macro is kept in both debug/release and should be used to protect invariants of the system as necessary
 * `unreachable!()` - This macro will panic on code that should not be reached (violating an invariant) and can be used where appropriate.
 
+In production (non-test) code, outside of lock management,  all unrecoverable errors should be cleanly documented describing why said event is unrecoverable. For example, if the system is now in a bad state, state what that state is and the motivation for why a crash / restart is more expedient than resolving it within a running system, and what if any steps an operator would need to take to resolve the issue.
```
mimoo, line 200: "state what that state" :P

mimoo: 👍  (although a "crash is more expedient" can lead someone to argue that a crash is valid, when it could be loop exploited)

documentation/coding_guidelines.md
davidiw: This is legacy guidelines, I would argue it is good practice to wrap arc + locks into a new type if it improves utility of the code.

documentation/coding_guidelines.md
mimoo: maybe explain that (non-)mutability should be enforced via the absence or presence of the `mut` keyword.

documentation/coding_guidelines.md
davidiw: I think there's a longer conversation to be had here, the intent was to start surfacing a little more discipline when we use the above macros. I think the broader question is do we allow these macros? Do we have our own set of macros? I want to avoid that for now and give a path to those conversations.

documentation/coding_guidelines.md
mimoo: I suggest advising to use this pattern if the test module becomes too big.

davidiw: actually @bmwill comment isn't clear, is it on my crappy Foo structure or on line 278. In general, I am not a big fan of line 278. It probably means that your code is too long and complex or that you're not writing a unit test.

documentation/coding_guidelines.md
davidiw: You mean on the data structure itself? I'll think about how to shove that in, but I can see this being the reason for even having this blurb.

mimoo: on a variable referring to that data structure

aching: Curious why https://rust-lang.github.io/api-guidelines/naming.html#getter-names-follow-rust-convention-c-getter is no longer a best practice?

documentation/coding_guidelines.md
aching: Any reason to change preferring Arc::clone(x) vs x.clone()?

zekun000: too many chars to type (that's my reason)

aching: So the extra 5 chars is not worth the clarity? :). Anything else?

documentation/coding_guidelines.md
bmwill: That is still best practice, and you do want to have getters/setters in order to maintain invariants. This is why I asked to remove it (after the change). Sorry I wasn't exactly clear

documentation/coding_guidelines.md
bmwill: specifically line 278, test modules should be submodules of the code they are trying to test, keeping the tests close to the source. We should not encourage placing tests is `_test.rs` files at the root of a crate....Though now that I'm reading it again its not exactly saying what I thought it was (which maybe means it wasn't clear enough?) lol

documentation/coding_guidelines.md
```diff
@@ -254,7 +254,28 @@ We currently use [log](https://docs.rs/log/) for logging.
 
 *Unit tests*
 
-Ideally, all code should be unit tested.  Unit test files should be in the same directory as `mod.rs` and their file names should end in `_test.rs`.  A module to be tested should have the test modules annotated with `#[cfg(test)]`.  For example, if in a crate there is a db module, the expected directory structure is as follows:
+Ideally, all code should be unit tested.  Unit tests should ideally be in the same file as the code it is testing though in a distinct module, using the following syntax:
```
JoshLind, line 257: nit: Maybe change:
.. it is testing though ..
-> 
... it is testing (though in a distinct module), using..

documentation/coding_guidelines.md
JoshLind: I agree with @bmwill that this isn't super clear. Why not just link to the best practices guide instead of trying to define this ourselves? https://doc.rust-lang.org/book/ch11-03-test-organization.html

Or is it because we want to deviate from the guide with exceptional cases?

documentation/coding_guidelines.md
davidiw: I don't think the use of Arc::clone is systemic through the code base and there are no lints for it. If we have a lint, then I'm all game, but as it is, it is often ignored. Also I don't really know why this should really matter? We should prefer references if we can use references, except maybe on copyable types. If we have to clone, we have to clone.

documentation/coding_guidelines.md
mimoo: I don't think it's that simple though, in some cases I see structs with getters and setters for all fields, what's the point of that versus making all fields `pub`?

davidiw: Are you responding to the thread or the commit? I have updated the text.

documentation/coding_guidelines.md
aching: I think in general, there doesn't appear to be a huge gain one way or another but rather we should aim for consistency in the codebase as a principle - even if there isn't a lint rule.  I'm okay either way as long as we pick one way. =)

documentation/coding_guidelines.md
mimoo: thread

documentation/coding_guidelines.md
davidiw: I'm happy to run a sed script to remove them. But I think what folks want to do is really up to them. Are we going to say eliminate the use of Type::clone() from the code base? It is syntactically correct rust. Similarly should we insist that you cannot use a trait directly with the type as an input but rather must run the trait via an instance of a struct that implements that trait (generalizing the Arc argument to all structs)?

documentation/coding_guidelines.md
bmwill: Yes! Let just link to the best practices :)

documentation/coding_guidelines.md
aching: I don't see a strong argument for those broader suggestions.  I thought the existing text was a reasonable justification for preferring Arc::clone(). It's also used quite a bit in our current codebase (100+ times).

davidiw: I'm willing to bet there are more instances of instance::clone() where instance is of type Arc<T> then there are of Arc::clone(instance)... I've added quite a lot myself. A guideline that isn't easy to follow or enforce and has little ramifications seems superfluous and distracts from guidelines we really care about. I'd be perfectly fine with striking this whole section since there are examples of folks passing around Arc<T> instead of T { innner: Arc<S> }. I digress.

aching: I agree with a guideline that is hard to follow is not very useful.  As I mentioned above, enforcement is a nice to have but not required for a guideline to be useful.

Since it is a bit tricky to figure out when to Arc::clone() vs clone() in downstream cases as well as @zekun000's argument to reduce chars, I'm fine with this change.

PR #5851, opened 2020-08-31
[logging] remove unique structured log id

gregnazario:
The log ids are now handled in logstash, and don't need to be generated
by the program.  It's based off a consistent hash of the data in the
log, and will handle duplicates.


gregnazario:
> Looks like you just need to fix 1 test and this should be good to land

Already ahead of you

PR #5853, opened 2020-08-31
[docker] push base images to ecr, verify shas.

rexhoffman:
## Motivation

The purpose of this pr has changed.    Too much refactoring/linting/lint enablement/base image work went in to this pr, and seperating that from blocking on ecr push (ran in to an interesting roadblock that will require some work).

Instead land all the filtering and refactoring first, and the base image next.

---------------
Novi ecr, used in landblocking tests that build docker images in AWS codebuild, require base images from dockerhub that occasionally hit API limits.

Prevent this, by pushing the base dockerhub debian:buster images we use to novi ecr in the auto branch.

Also, demands that libra base images are tagged with sha256 values as well, and breaks pr/auto builds if they are not.

Purposefully miss a few files in the initial pr to see failures.

We use a hodgepodge of debian:buster and debian:buster-slim for prod images.... why?

Tested failing the build here:
https://app.circleci.com/pipelines/github/libra/libra/28914/workflows/485f3900-186f-4b3f-be2c-92bd6ddb3757/jobs/194687

Yes

## Test Plan

clever use of CI

## Related PRs

None


.circleci/config.yml
```diff
@@ -743,20 +763,40 @@ jobs:
...
               echo Targeting branch $TARGET_BRANCH will not publish docker images.
+              circleci step halt
             fi
+      - setup_docker_signing
+      - setup_aws
+      - fail_if_debian_buster_docker_base_images_are_not_tagged_with_sha256
+      - run:
+          name: Push base images to ecr if needed for cluster test.
+          command: |
+            # We are using debian buster for our base images.   Get all debain tags in a unique sorted list.
```
sausagee, line 776: nit - typo `debain` --> `debian`

rexhoffman:
@sausagee 

This pr now has a generic mechanism to detect when to build based on file changes in a branch (right now targeting master) using cache files with the git hash of the last successful run for comparison.

If an overwrite of a branch's history occurs, as can happen on test branches, that is covered as well, and a build will occur.

The use of conditional when clauses also simplifies the logic in the steps used to determine if a job should halt.

Finally no longer prebuilding docker images, to allow for the removal of secrets from auto/canary branches.

@mgorven I'll need permissions updated on our circleci user to allow it to push debian images to novi aws ecr.

Note: I am not certain if I'm pushing to the right place/name for code build to pick them up.   Guidance appreciated.
@andll or whoever set up codebuild.

rexhoffman:
At this point I'm waiting to get perms for the aws image push and ensure the pull works in codebuild.

I do realize though I could build a reusable script after this....   if PR git rev compare against merge head of merge target, if not a commit pr compare against last cached file githash.    This would always let us detect changed files.   CIRCLE has all the needed info for use to look up this info.   I could build a simple script that would provide all the info + files changed in a simple script if desired.   Right now our use cases are met though.

.circleci/config.yml
```diff
@@ -168,6 +173,74 @@ commands:
...
+      file of relevant git files.
+    parameters:
+      cache_key_part:
+        description: part of the cache key name used by circle in the job.
+        type: string
+      file_of_relevant_git_files:
+        description: file to be provided to these steps, of relevant git files.
+        type: string
+    steps:
+      - restore_cache:
```
sausagee, line 213: The detect-and-determine solution (where it compares HEAD against origin) in the previous commit in this PR was more elegant and portable. Perhaps I didn't see the trade-off with using the circleci built-in cache management.

rexhoffman: there are two implementations that I pulled apart -- on for updates to master/release branchs. which this is --  there is no remote origin to compare to in that instance -- you must use a cache key as the build wont know which commits on the branch are new.   The other mechanism only works on prs.   I'm currently writing a shell script to merge the approaches, decoupled from circle.

mgorven:
> We use a hodgepodge of debian:buster and debian:buster-slim for prod images.... why?

The `-slim` image smaller, but lacks some very basic Linux tools. This is fine for images used as a command (`client`, `tools`) but not as much for long running images (`validator` etc.).

PR #5855, opened 2020-09-01
[libra framework] specify mint_lbr and some functions in LBR, LibraAccount

emmazzz:
## Motivation

This PR adds documentation and specifications for `mint_lbr` transaction script, as well as specifications for underlying modules such as `LibraAccount` and `LBR`. A lot of the functions in these modules are made `opaque`.

Yes.

## Test Plan

cargo test


language/stdlib/transaction_scripts/mint_lbr.move
```diff
@@ -1,11 +1,73 @@
 script {
 use 0x1::LibraAccount;
+// prover deps:
+use 0x1::Signer;
+use 0x1::Option;
 
 /// Mint `amount_lbr` LBR from the sending account's constituent coins and deposits the
 /// resulting LBR into the sending account.
+///
+/// # Events
```
shazqadeer, line 10: This documentation seems the same as that in stdlib.rs.  Is one automatically generated from the other?

language/stdlib/modules/AccountLimits.move
```diff
@@ -311,7 +316,7 @@ module AccountLimits {
         amount: u64,
         receiving: &mut Window<CoinType>,
     ): bool acquires LimitsDefinition {
-        let limits_definition = borrow_global_mut<LimitsDefinition<CoinType>>(receiving.limit_address);
+        let limits_definition = borrow_global<LimitsDefinition<CoinType>>(receiving.limit_address);
```
shazqadeer, line 319: This is good!  @tzakian and I also discovered in some part of the code that a mutable borrow was being taken when an immutable would suffice.

language/stdlib/modules/AccountLimits.move
```diff
@@ -354,6 +354,19 @@ module AccountLimits {
         aborts_if spec_window_reset(receiving).window_inflow + amount > max_u64();
```
shazqadeer, line 354: Is it possible to rewrite the code so that overflow cannot happen?  I managed to do it for DesignatedDealer.  It may be helpful to look over there for coding patterns.

shazqadeer: Also, we are trying to always have an abort code with aborts_if specification.

language/stdlib/transaction_scripts/mint_lbr.move
```diff
@@ -1,11 +1,73 @@
...
+/// are the components amounts of `amount_lbr` LBR, and
+/// `ReceivedPaymentEvent { amount_lbr, currency_code = LBR, address_of(account), metadata = x"" }`
+/// on `account`'s `LibraAccount::received_events` handle.
+///
+/// # Abort Conditions
+/// The following abort conditions have been formally specified and verified. See spec schema
+/// `LibraAccount::StapleLBRAbortsIf` for the formal specifications.
+///
+/// ## Aborts Caused by Invalid Account State
+/// * Aborts with `LibraAccount::EINSUFFICIENT_BALANCE` if `amount_coin1` is greater than sending
```
wrwg, line 24: I think we should document the category here (as the spec does) and leave the error reason to someone who actually wants to debug and uses @tzakian debug tool. The error reason may be brittle and is not part of the contract in my book. But definitely a difficult decision, so if you want to keep this, perhaps add a TODO that we need to settle the approach here.

language/stdlib/transaction_scripts/mint_lbr.move
```diff
@@ -1,11 +1,73 @@
...
+// prover deps:
+use 0x1::Signer;
+use 0x1::Option;
 
 /// Mint `amount_lbr` LBR from the sending account's constituent coins and deposits the
 /// resulting LBR into the sending account.
+///
+/// # Events
+/// When this script executes without aborting, it emits three events:
+/// `SentPaymentEvent { amount_coin1, currency_code = Coin1, address_of(account), metadata = x"" }`
```
wrwg, line 12: ... and ...

language/stdlib/modules/AccountLimits.move
wrwg: This is potentially a candidate for a compiler/linter warning/error. @tnowacki

language/stdlib/modules/AccountLimits.move
wrwg: I'm actually not sure whether this is a good idea, because it has the tendency to hide programming errors. If things are out of bound, instead of a specific error, you may get some strange other error. 

In general, if this is going to happen, because it is a significant code change (I know because I've written those aborts here a while ago), I would not mix it into this PR. Perhaps in DesignatedDealer it was a moderate change, but here it will be quite a thing.

language/stdlib/modules/LBR.move
```diff
@@ -150,6 +150,7 @@ module LBR {
     public fun calculate_component_amounts_for_lbr(amount_lbr: u64): (u64, u64)
     acquires Reserve {
         LibraTimestamp::assert_operating();
+        assert(exists<Reserve>(CoreAddresses::LIBRA_ROOT_ADDRESS()), Errors::not_published(ERESERVE));
```
wrwg, line 153: Is there perhaps a global invariant missing that Reserve exists after genesis? Because you already assert_operating, this may become obsolete. General picture is that if EXECUTION_FAILURE can provenly not happen, we also do not need to catch it. As long as we do not use partial aborts spec, we should always catch if this does not happen.

language/stdlib/modules/LibraAccount.move
```diff
@@ -1127,7 +1263,7 @@ module LibraAccount {
     spec module {
         /// The LibraAccount under addr holds either no withdraw capability
         /// (withdraw cap has been delegated) or the withdraw capability for addr itself.
-        invariant [global, isolated] forall addr1: address where exists_at(addr1):
+        invariant [global] forall addr1: address where exists_at(addr1):
```
wrwg, line 1266: Curious: did you need to remove isolated because some other proof depends on this?

language/stdlib/transaction_scripts/mint_lbr.move
emmazzz: The comments in stdlib.rs are auto generated from here(mint_lbr.move).

language/stdlib/modules/LibraAccount.move
emmazzz: Yes in transaction scripts like this one where we extract the withdraw capability from the sending account and uses that withdraw capability, we need the property that the capability indeed belongs to the sending account.

language/stdlib/modules/LBR.move
emmazzz: You are right. There is already a global invariant there and we don't need this assertion anymore.

language/stdlib/modules/LibraAccount.move
```diff
@@ -27,7 +27,8 @@ module LibraAccount {
     use 0x1::Libra::{Self, Libra};
     use 0x1::Option::{Self, Option};
     use 0x1::Roles;
-
+    // Prover deps:
+    use 0x1::FixedPoint32;
```
tnowacki, line 31: You can put the use at the beginning of the schema if you want

tnowacki: (I think)

PR #5859, opened 2020-09-01
[Common] Add support for a new AnnotatedError type to provide context (e.g., backtraces) to errors

JoshLind:
## Motivation

This PR adds a new AnnotatedError type: a wrapper around std::error::Error. AnnotatedError provides an additional struct in which to hold auxiliary information about the context of an error object, e.g., where the error originated from in the source code, such as the error backtrace. This is useful for debugging, e.g., when an error is thrown or displayed on the console but the error message isn't helpful enough to identify exactly what went wrong. Moreover, this is also useful when updating `panic!` related code to instead use errors and not lose context (e.g., see the conversation in this issue: https://github.com/libra/libra/pull/5617)

This PR offers three commits:
1. First, we create the AnnotedError type.
2. Second, we update the KeyManager to use the AnnotatedError type as a way to show how it works in practice.
3. Third, we update the summaries in the code (as required by land blocking tests)

See the section below for how the AnnotatedError type is formatted when printed.

Yes.

## Test Plan

All local tests pass. Moreover, forcing a test to return a failure successfully prints out the backtrace of the AnnotatedError type (i.e., where the type was created, and thus the error failure point). This helps to identify what went wrong. See the example below. Frame 7 shows the function that the error was created in and the exact line number (this is without using any backtrace env variables!):

```
---- tests::test_key_manager_init_and_basic_rotation stdout ----
thread 'tests::test_key_manager_init_and_basic_rotation' panicked at 'called `Result::unwrap()` on an `Err` value: Error: ConfigStorageKeyMismatch(Ed25519PublicKey(92057b752b05eec0e64f6f1c23b0071e649fc9511ac41e2bc62acba0e525ef51), Ed25519PublicKey(92057b752b05eec0e64f6f1c23b0071e649fc9511ac41e2bc62acba0e525ef51)),
Backtrace:
    0: backtrace::backtrace::libunwind::trace
             at /Users/joshlind/.cargo/registry/src/github.com-1ecc6299db9ec823/backtrace-0.3.50/src/backtrace/libunwind.rs:95
      backtrace::backtrace::trace_unsynchronized
             at /Users/joshlind/.cargo/registry/src/github.com-1ecc6299db9ec823/backtrace-0.3.50/src/backtrace/mod.rs:66
   1: backtrace::backtrace::trace
             at /Users/joshlind/.cargo/registry/src/github.com-1ecc6299db9ec823/backtrace-0.3.50/src/backtrace/mod.rs:53
   2: backtrace::capture::Backtrace::create
             at /Users/joshlind/.cargo/registry/src/github.com-1ecc6299db9ec823/backtrace-0.3.50/src/capture.rs:164
   3: backtrace::capture::Backtrace::new
             at /Users/joshlind/.cargo/registry/src/github.com-1ecc6299db9ec823/backtrace-0.3.50/src/capture.rs:128
   4: libra_annotated_error::AnnotatedError<T>::new
             at /Users/joshlind/Facebook/Libra/Code/fork/libra/common/annotated-error/src/lib.rs:24
   5: <libra_annotated_error::AnnotatedError<T> as core::convert::From<T>>::from
             at /Users/joshlind/Facebook/Libra/Code/fork/libra/common/annotated-error/src/lib.rs:53
   6: <T as core::convert::Into<U>>::into
             at /Users/joshlind/.rustup/toolchains/1.45.2-x86_64-apple-darwin/lib/rustlib/src/rust/src/libcore/convert/mod.rs:559
   7: libra_key_manager::KeyManager<LI,S,T>::compare_storage_to_config
             at src/lib.rs:193
   8: libra_key_manager::tests::verify_init_and_basic_rotation
             at src/tests.rs:538
   9: libra_key_manager::tests::test_key_manager_init_and_basic_rotation
             at src/tests.rs:529
  10: libra_key_manager::tests::test_key_manager_init_and_basic_rotation::{{closure}}
             at src/tests.rs:526
  11: core::ops::function::FnOnce::call_once
             at /Users/joshlind/.rustup/toolchains/1.45.2-x86_64-apple-darwin/lib/rustlib/src/rust/src/libcore/ops/function.rs:232
...
```

The previous output (without AnnotatedError) just displays the snippet below. The debuggability of this seems to just come down to how clear the error message is. In many cases, it's not super helpful without knowing exactly what's going on and digging through the test.

```
failures:

---- tests::test_key_manager_init_and_basic_rotation stdout ----
thread 'tests::test_key_manager_init_and_basic_rotation' panicked at 'called `Result::unwrap()` on an `Err` value: ConfigStorageKeyMismatch(Ed25519PublicKey(92057b752b05eec0e64f6f1c23b0071e649fc9511ac41e2bc62acba0e525ef51), Ed25519PublicKey(92057b752b05eec0e64f6f1c23b0071e649fc9511ac41e2bc62acba0e525ef51)), secure/key-manager/src/tests.rs:538:5
```

## Related PRs

None.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2769,6 +2769,7 @@ dependencies = [
  "executor-test-helpers 0.1.0",
  "executor-types 0.1.0",
  "futures 0.3.5 (registry+https://github.com/rust-lang/crates.io-index)",
+ "libra-annotated-error 0.1.0",
  "libra-config 0.1.0",
  "libra-crypto 0.1.0",
  "libra-global-constants 0.1.0",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2419,6 +2419,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "libra-annotated-error"
+version = "0.1.0"
+dependencies = [
+ "backtrace 0.3.50 (registry+https://github.com/rust-lang/crates.io-index)",
+ "libra-workspace-hack 0.1.0",
+ "thiserror 1.0.20 (registry+https://github.com/rust-lang/crates.io-index)",
+]
+
+[[package]]
 name = "libra-bitvec"
 version = "0.1.0"
 dependencies = [
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -5,6 +5,7 @@ members = [
     "client/libra-dev",
     "client/swiss-knife",
     "client/transaction-builder",
+    "common/annotated-error",
     "common/bitvec",
     "common/bounded-executor",
     "common/channel",
--- a/common/annotated-error/Cargo.toml
+++ b/common/annotated-error/Cargo.toml
@@ -1,0 +1,18 @@
+[package]
+name = "libra-annotated-error"
+version = "0.1.0"
+authors = ["Libra Association <opensource@libra.org>"]
+description = "Libra annotated error type"
+repository = "https://github.com/libra/libra"
+homepage = "https://libra.org"
+license = "Apache-2.0"
+publish = false
+edition = "2018"
+
+[dependencies]
+backtrace = "0.3"
+
+libra-workspace-hack = { path = "../../common/workspace-hack", version = "0.1.0" }
+
+[dev-dependencies]
+thiserror = "1.0.20"
--- a/common/annotated-error/src/lib.rs
+++ b/common/annotated-error/src/lib.rs
@@ -1,0 +1,95 @@
+// Copyright (c) The Libra Core Contributors
+// SPDX-License-Identifier: Apache-2.0
+
+#![forbid(unsafe_code)]
+
+use backtrace::Backtrace;
+use std::{error::Error, fmt};
+
+/// The AnnotatedError struct is a wrapper around std::error:Error. AnnotatedError provides an
+/// additional struct in which to hold auxiliary information about the context of an error object,
+/// e.g., where the error originated from in the source code, such as the error backtrace. This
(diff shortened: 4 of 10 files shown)


JoshLind:
@mimoo, interested to see what you think about this. Note:
1. I'm currently considering renaming "AnnotatedError" to "ErrorWithContext", or something else (if there are better suggestions).
2. I'm also considering improving the bracktrace output by only displaying the frames from the new() call, so that it'll allow developers to see the location of the error more quickly.

Overall, this is an MVP, but it seems it might be useful to incorporate into the TCB.

JoshLind:
Thanks, David!

> One other approach I'm curious about is forking thiserror to use the backtrace crate. Did you look at that? 

I thought about it, but there are some downsides:
- Having to maintain a forked version of a dependency and deal with upstream updates gets messy quickly.
- In order to get the benefits of AnnotatedError, we'd require all users to use our fork (which might not be ideal in some circumstances -- e.g., if callers prefer to define their own std::error::Error implementation, or prefer not to use thiserror in general).

As thiserror provides a std::error::Error implementation, it seemed like the best thing would be to wrap std::error::Error directly. That way AnnotatedError could be used with any std::error::Error implementation (which is as general as possible).

> Also how do I get my T::error? Should you have a to_inner and error function?

Do you mean how do I get the inner error object? Right now we just make the internal error field public to avoid getters. The benefit of this is that it's easy to just call `annotated_error.error`. If you think we should expose a getter method instead of having a public field, we can do that, but I don't feel too strongly about the choice. Does that make sense?

PR #5869, opened 2020-09-01
[network] Refactor current network metrics

gregnazario:
### Overview
* Rename metrics that didn't fit the naming scheme
* Update descriptions to be concise, but helpful
* Added some helpful documentation at the top and remove inlined but inconsistent comments

### DIFF
--- a/network/src/counters.rs
+++ b/network/src/counters.rs
@@ -23,7 +23,7 @@ pub const SENT_LABEL: &str = "sent";
 pub static LIBRA_NETWORK_PEERS: Lazy<IntGaugeVec> = Lazy::new(|| {
     register_int_gauge_vec!(
         "libra_network_peers",
-        "Libra network peers counter",
+        "Number of peers, and their associated state",
         &["role_type", "state"]
     )
     .unwrap()
@@ -32,7 +32,7 @@ pub static LIBRA_NETWORK_PEERS: Lazy<IntGaugeVec> = Lazy::new(|| {
 pub static LIBRA_CONNECTIONS: Lazy<IntGaugeVec> = Lazy::new(|| {
     register_int_gauge_vec!(
         "libra_connections",
-        "Libra connections counter",
+        "Number of current connections and their direction",
         &["role_type", "network_id", "peer_id", "direction"]
     )
     .unwrap()
@@ -65,7 +65,7 @@ pub static LIBRA_NETWORK_DISCOVERY_NOTES: Lazy<IntGaugeVec> = Lazy::new(|| {
 pub static LIBRA_NETWORK_RPC_MESSAGES: Lazy<IntCounterVec> = Lazy::new(|| {
     register_int_counter_vec!(
         "libra_network_rpc_messages",
-        "Libra network rpc messages counter",
+        "Number of RPC messages",
         &["type", "state"]
     )
     .unwrap()
@@ -74,7 +74,7 @@ pub static LIBRA_NETWORK_RPC_MESSAGES: Lazy<IntCounterVec> = Lazy::new(|| {
 pub static LIBRA_NETWORK_RPC_BYTES: Lazy<HistogramVec> = Lazy::new(|| {
     register_histogram_vec!(
         "libra_network_rpc_bytes",
-        "Libra network rpc bytes histogram",
+        "Number of RPC bytes transferred",
         &["type", "state"]
     )
     .unwrap()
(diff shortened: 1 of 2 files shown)


network/src/counters.rs
```diff
@@ -1,6 +1,43 @@
 // Copyright (c) The Libra Core Contributors
 // SPDX-License-Identifier: Apache-2.0
 
+//! # Metrics
```
davidiw, line 4: This is great but seems like something we should either have in our metrics crate or just point to the original source documentation. Being in network means other counter crates won't benefit.

network/src/counters.rs
```diff
@@ -74,7 +102,7 @@ pub static LIBRA_NETWORK_DISCOVERY_NOTES: Lazy<IntGaugeVec> = Lazy::new(|| {
 pub static LIBRA_NETWORK_RPC_MESSAGES: Lazy<IntCounterVec> = Lazy::new(|| {
```
davidiw, line 102: Do we want this? It might make more sense to make a single counter for Rpc Request, Rpc Response, and Direct Send

We can also merge this with the bytes below by an additional field ... metric (bytes versus count)

It would also be great to have a per peer counter for each of these (another field)

network/src/counters.rs
```diff
@@ -92,7 +120,7 @@ pub static LIBRA_NETWORK_RPC_BYTES: Lazy<HistogramVec> = Lazy::new(|| {
 pub static LIBRA_NETWORK_RPC_LATENCY: Lazy<HistogramVec> = Lazy::new(|| {
     register_histogram_vec!(
         "libra_network_rpc_latency_seconds",
-        "Libra network rpc latency histogram",
+        "RPC request latency in seconds",
```
davidiw, line 123: what is a type? that seems redundant with protocol_id

network/src/counters.rs
```diff
@@ -121,7 +149,7 @@ pub static LIBRA_NETWORK_DIRECT_SEND_BYTES: Lazy<HistogramVec> = Lazy::new(|| {
 pub static PENDING_NETWORK_NOTIFICATIONS: Lazy<IntCounterVec> = Lazy::new(|| {
     register_int_counter_vec!(
         "libra_network_pending_network_notifications",
-        "Counters(queued,dequeued,dropped) related to pending inbound network notifications",
+        "State of pending inbound network notifications",
```
davidiw, line 152: I do wonder how meaningful this really is since we're talking about 15 seconds between metric pulls...

network/src/counters.rs
```diff
@@ -140,17 +168,18 @@ pub static PENDING_NETWORK_REQUESTS: Lazy<IntCounterVec> = Lazy::new(|| {
 /// Counter of pending network events to Health Checker.
 pub static PENDING_HEALTH_CHECKER_NETWORK_EVENTS: Lazy<IntCounterVec> = Lazy::new(|| {
     register_int_counter_vec!(
-        "pending_health_checker_network_events",
-        "Counters(queued,dequeued,dropped) related to pending network notifications to HealthChecker",
+        "libra_network_pending_health_check_events",
```
davidiw, line 171: same here... it might be more informative to learn about missed / acked health check events or more specifically when was the last time I request a message and when was the last time I heard from them.

network/src/counters.rs
gregnazario: I agree.  I haven't spent a lot of time looking into common metrics infrastructure, in comparison to how we do logging, so I want a place to put this, just don't know where yet

network/src/counters.rs
gregnazario: So, I did a look into this to combine the RPC & DirectSend messages.  The problem is we have another dimension for RPC (the RPC command), so it becomes awfully complex to have it combined in that way.

We can totally add more dimensions to the counters, this was more of a "naming" and "documentation" change to prevent the explosion of changes in this PR.  Those can follow :)

network/src/counters.rs
gregnazario: The dimensions are very confusing to me too.  That I think has to deal with refactoring the metrics in general (anything reading from the metric has to also get the name updated)

network/src/counters.rs
gregnazario: Yeah and there's only 1 minute granularity through Grafana.

I think this is also not a valid description now that I look at it, it's not the state of pending inbound network notifications, but the number of network notifications that were processed at different states.

network/src/counters.rs
gregnazario: Found it!  will update PR.

network/src/counters.rs
gregnazario: I'll come by on the next pass and update the metrics in general.  I just wanted to tie updating the descriptions to the documentation updates.

davidiw (approved):

gregnazario:
This is just an initial run at some stuff, will be making more updates!
/land

PR #5873, opened 2020-09-01
[storage][dashboard] add restore status panel

lightmark:
## Motivation

add a panel track restore status.

image


sherry-x:
can you attach some screenshots so we can see what it looks like?

lightmark:
@sherry-x Just notice that. looks fine?

lightmark:
just two panels don't justify a separate dashboard- -

PR #5876, opened 2020-09-02
[breaking][vm][move stdlib] Use Errors module everywhere. Merge epilogues. Fix bugs.

tnowacki:
- Errors inside of `LibraAccount` were no longer unique due to some edge cases
- Switching to use `Errors` everywhere will make things unique
- Removes limits from epilogue payments
- Fixes bug where the prologue could hit an invariant violation by not checking for the token to exist
- Also lets us handle withdrawal limits error that was previously not covered
- Switched to a single `epilogue` function 

Followup
- Perhaps case on a `reason` of `0` in `Errors`
- Perhaps check for a reason being the proper size (not too large, but maybe we already check that in `<<`, I forget)

## Motivation

- Errors need to be unique for proper casing in the VM
- There was an edge case where a certain error was not considered inside the VM (if the account hit the limits in the epilogue)

## Test Plan

- Tests for the bugs fixed


tnowacki:
There are still TODOs for tests, but I'm marking this as ready for review to get some more eyes on it

wrwg (approved):
This looks good from my viewpoint. Great to see it, I didn't dare to do this when Errors module was introduced.

language/stdlib/modules/LibraAccount.move
```diff
@@ -1051,126 +1060,105 @@ module LibraAccount {
...
 
         // Verify that the transaction sender's account exists
-        assert(exists_at(transaction_sender), PROLOGUE_EACCOUNT_DNE);
+        assert(exists_at(transaction_sender), Errors::invalid_argument(PROLOGUE_EACCOUNT_DNE));
 
         // We check whether this account is frozen, if it is no transaction can be sent from it.
         assert(
             !AccountFreezing::account_is_frozen(transaction_sender),
-            PROLOGUE_EACCOUNT_FROZEN
+            Errors::invalid_state(PROLOGUE_EACCOUNT_FROZEN)
```
runtian-zhou, line 1071: Why froizen account is of `invalid_state`?

runtian-zhou (commented):
This seems to be a BIG breaking change to me. Do we have a plan of when this thing should be merged into the next release cycle?

language/libra-vm/src/system_module_names.rs
```diff
@@ -33,9 +33,9 @@ pub static WRITESET_PROLOGUE_NAME: Lazy<Identifier> =
 pub static WRITESET_EPILOGUE_NAME: Lazy<Identifier> =
     Lazy::new(|| Identifier::new("epilogue").unwrap());
 pub static SUCCESS_EPILOGUE_NAME: Lazy<Identifier> =
-    Lazy::new(|| Identifier::new("success_epilogue").unwrap());
```
bob-wilson, line 36: It seems like it would be better to just merge these two.

language/libra-vm/src/errors.rs
```diff
@@ -78,19 +96,30 @@ pub fn convert_normal_prologue_error(status: VMStatus) -> VMStatus {
 /// `UNEXPECTED_ERROR_FROM_KNOWN_MOVE_FUNCTION`
 pub fn convert_normal_success_epilogue_error(status: VMStatus) -> VMStatus {
```
bob-wilson, line 97: I think we should drop the "success" from the name of this function: "convert_normal_epilogue_error"

language/libra-vm/src/errors.rs
```diff
@@ -78,19 +96,30 @@ pub fn convert_normal_prologue_error(status: VMStatus) -> VMStatus {
...
-        VMStatus::MoveAbort(location, code @ EACCOUNT_FROZEN)
-        | VMStatus::MoveAbort(location, code @ EINSUFFICIENT_BALANCE) => {
-            if location != known_locations::account_module_abort() {
+        VMStatus::Executed => VMStatus::Executed,
+        VMStatus::MoveAbort(location, code)
+            if location != known_locations::account_module_abort() =>
+        {
+            let (category, reason) = error_split(code);
+            error!(
+                "[libra_vm] Unexpected success epilogue Move abort: {:?}::{:?} \
```
bob-wilson, line 105: ...and also drop the "success" in this message

language/libra-vm/src/errors.rs
```diff
@@ -78,19 +96,30 @@ pub fn convert_normal_prologue_error(status: VMStatus) -> VMStatus {
...
+                (Category: {:?} Reason: {:?})",
+                location, code, category, reason
+            );
+            VMStatus::Error(StatusCode::UNEXPECTED_ERROR_FROM_KNOWN_MOVE_FUNCTION)
+        }
+
+        VMStatus::MoveAbort(location, code) => match error_split(code) {
+            (LIMIT_EXCEEDED, ECANT_PAY_GAS_DEPOSIT) => VMStatus::MoveAbort(location, code),
+            (category, reason) => {
```
bob-wilson, line 114: It looks like you dropped handling EACCOUNT_FROZEN here. Was that intentional?

language/stdlib/modules/LibraAccount.move
```diff
@@ -1051,126 +1060,105 @@ module LibraAccount {
...
             LibraTimestamp::now_seconds() < txn_expiration_time_seconds,
-            PROLOGUE_ETRANSACTION_EXPIRED
+            Errors::invalid_argument(PROLOGUE_ETRANSACTION_EXPIRED)
         );
     }
 
-    /// Collects gas and bumps the sequence number for executing a transaction
+    /// Collects gas and bumps the sequence number for executing a transaction.
+    /// The epilogue is invoked at the end of the transaction.
+    /// If the exection of the epilogue fails, it is re-invoked with different arguments guaranteed
```
bob-wilson, line 1115: I hesitate a bit at the word "guaranteed". Maybe hedge a bit and phrase it as "should never"?

bob-wilson (approved):

language/libra-vm/src/errors.rs
tnowacki: The flows are still separate in the VM, i.e. there still is a success and failure epilogue control flow. Each one of those expects different errors, so I'll keep the name same for the error handling

language/libra-vm/src/errors.rs
tnowacki: see above

language/libra-vm/src/errors.rs
tnowacki: I'm continuously frustrated by how little test coverage we have.

I think its a mistake, I'll take a look

language/stdlib/modules/LibraAccount.move
tnowacki: What else would it be? /shrug

The only one that seems close is `REQUIRES_ROLE`, but that doesn't seem right. State seems like the best fit given the description in `Errors.move`

Maybe @tzakian or @wrwg could chime in

language/libra-vm/src/errors.rs
tnowacki: It looks like someone removed it from the account code but didn't remove it here. Testing might have only fixed this in someone might have realized they needed to clean up the VM, but otherwise its okay I suppose..

language/stdlib/modules/LibraAccount.move
tzakian: I think `INVALID_STATE` makes sense here (the account is in an invalid state to send a transaction).

tzakian (approved):
This looks good to me. Thanks so much for taking this on!

PR #5877, opened 2020-09-02
[language][functional tests] support exp files

vgao1996:
## Summary
This allows exp files to be used in functional tests. If a `.exp` file is detected, the test drive will ignore all the checks and compare the log with the content of that file instead. The environment variable `UPDATE_BASELINE` can be used to update the exp files automatically.

The greater vision is to retire the checker and fully migrate to this new model. More config options are planned in order to allow parts of the log to be disabled/enabled, reducing the noise-to-signal ratio.

## Test Plan
cargo test


language/testing-infra/functional-tests/src/testsuite.rs
```diff
@@ -79,167 +93,227 @@ pub fn functional_tests<TComp: Compiler>(
...
-        )?;
-    } else {
-        for (id, entry) in res.text.iter().enumerate() {
-            if print_stage(entry) {
-                writeln!(output, "    [{}] {}", id, entry)?;
+    let mut exp_file_path = PathBuf::from(path);
+    exp_file_path.set_extension("exp");
+
+    if exp_file_path.exists() {
+        // REVIEW: "\n" is probably not very portable.
```
bob-wilson, line 128: How hard would it be to split this up into separate functions (e.g., one for handling .exp files and the other for embedded checks)? I think that might help make this more readable and maintainable, if it isn't too hard to split up that way.

bob-wilson: Actually, that same comment applies later in this function, too. It's just such a long sequence of code that it's hard to quickly see the overall logic.

bob-wilson (commented):
Do you have a sample of what the output looks like in different cases (not necessarily in the PR but just as a comment here)? I'm curious but too lazy to build the code. Not a big deal if you don't have that handy....

My personal preference would be to omit the ignore_checks_when_exp_file_present.mvir test. The same logic is getting tested in the corresponding functional-tests file, and I'd like to see us move away from IR tests instead of adding more of them.

language/testing-infra/functional-tests/src/testsuite.rs
vgao1996: It shouldn't be hard -- a few macros needs to capture some bindings but it's not that troublesome. Let me try.

vgao1996:
Screen Shot 2020-09-02 at 6 05 16 PM
It outputs nothing if the two are identical.

I get where you are coming from regarding IR vs Move tests. However I'd still suggest we keep `ignore_checks_when_exp_file_present.mvir`. The reason is that this particular test is one of the special ones ensuring the testing infrastructure is functioning as intended. It is not really testing the respective Move/IR pipelines themselves. Even though we all agree on the vision to migrate the high level tests to Move lang (many has already been migrated) and cleanup the IR tests, that will not happen overnight. I think as long as the IR testing pipeline remains as a client of this feature (exp files), even though it's shared with the Move one, we'd better make sure it works in both code paths.

vgao1996:
Interesting... Just realized our tests are not deterministic: a different write set is created each time. I thought I got this issue fixed long time ago? Let me check if there's anything random in this process.

vgao1996:
@bob-wilson I split them into functions. It actually turned out better than I initially anticipated.

bob-wilson (approved):
Thanks for splitting up those functions. I like it much better like this!

language/ir-testsuite/tests/misc/ignore_checks_when_exp_file_present.exp
```diff
@@ -0,0 +1,9 @@
+[0] Transaction(0)
+[1] Stage(Compiler)
+[2] Output(CompilerLog("Script(\nImports()\nDependency(\n)\nConstants(\n)\nMain(() (return ();\n\n)))"))
+[3] Output(CompiledScript(CompiledScript(CompiledScriptMut { module_handles: [], struct_handles: [], function_handles: [], function_instantiations: [], signatures: [Signature([])], identifiers: [], address_identifiers: [], constant_pool: [], type_parameters: [], parameters: SignatureIndex(0), code: CodeUnit { locals: SignatureIndex(0), code: [Ret] } })))
+[4] Stage(Verifier)
+[5] Stage(Serializer)
+[6] Stage(Runtime)
+[7] Output(TransactionOutput(TransactionOutput { write_set: WriteSet(WriteSetMut { write_set: [(AccessPath { address: a4a46d1b1421502568a4a6ac326d7250, path: 0185aa1f79e25ca70b8f54f8a0b673259fc19b1af35acb98885d99e160600387c9 }, Value(20975741f6b37b591141e96e0c60a57ce8a4a46d1b1421502568a4a6ac326d725001a4a46d1b1421502568a4a6ac326d725001a4a46d1b1421502568a4a6ac326d72500000000000000000180000000000000000a4a46d1b1421502568a4a6ac326d72500000000000000000180000000000000000a4a46d1b1421502568a4a6ac326d72500100000000000000))] }), events: [], gas_used: 16, status: Keep(EXECUTED) }))
```
tnowacki, line 8: By default, I think we should just output:
```
Keep(EXECUTED)
```
I wouldn't really want any of the rest of this information as it makes it much harder to discern what is an interesting change or not.

Thoughts?

tnowacki: To be clear, I understand part of this is to start small and expand over time. But I feel that we should start with the least amount of information in the exp file as possible, and add opt-ins as needed

vgao1996: Should be doable. This was what I was planning to do in a subsequent PR anyways.

tnowacki (changes requested):
Lets go out with the next planned PR integrated in this one to keep things a bit simpler

Hopefully that should work for you, @vgao1996. If not, ping me :)

vgao1996:
@tnowacki yeah sounds good. I'm fine with a big PR as long as you can wait a day or two.

vgao1996:
@tnowacki I've implemented the features you requested. Highlights:
- The log will only contain the transaction status by default.
- Config options to include additional information. 
  - `//! show-gas`, `//! show-writeset`, `//! show-events`
- The reduction of the log is only effective when using exp files. It's kinda tricky to modify the existing checker based flow and I'd prefer not to go down that rabbit hole right now, knowing that they'll be migrated to a new system soon.

language/ir-testsuite/tests/misc/ignore_checks_when_exp_file_present.exp
```diff
@@ -0,0 +1,9 @@
+[0] Transaction(0)
+[1] Stage(Compiler)
+[2] Output(CompilerLog("Script(\nImports()\nDependency(\n)\nConstants(\n)\nMain(() (return ();\n\n)))"))
+[3] Output(CompiledScript(CompiledScript(CompiledScriptMut { module_handles: [], struct_handles: [], function_handles: [], function_instantiations: [], signatures: [Signature([])], identifiers: [], address_identifiers: [], constant_pool: [], type_parameters: [], parameters: SignatureIndex(0), code: CodeUnit { locals: SignatureIndex(0), code: [Ret] } })))
+[4] Stage(Verifier)
+[5] Stage(Serializer)
+[6] Stage(Runtime)
+[7] Transaction Status: Keep(EXECUTED)
+[8] Status(Success)
```
tnowacki, line 9: Sorry, I should have been more explicit in my last review. I don't think we should output anything here but step 7. We can opt into the rest. 

Ideally, the exp file for this test would look like. 

```
Transaction 0 (Starts at line 1): 

Transaction Status: Keep(EXECUTED)
```

language/testing-infra/functional-tests/src/evaluator.rs
```diff
@@ -170,6 +179,13 @@ impl EvaluationLog {
         res
     }
 
+    pub fn to_text_for_matching(&self) -> Vec<String> {
```
tnowacki, line 182: I think you are going to run into a lot of issues if you let the log define how it gets printed. I think this could be simpler if you extract all of this logic out and just put it into `check_or_update_expected_output`

language/testing-infra/functional-tests/src/testsuite.rs
```diff
@@ -65,31 +80,87 @@ fn print_stage(haystack: &str) -> bool {
...
+    log: &EvaluationLog,
 ) -> datatest_stable::Result<()> {
-    let input = read_to_string(path)?;
+    // REVIEW: "\n" is probably not very portable.
+    let text = log
+        .to_text_for_matching()
+        .into_iter()
+        .enumerate()
+        .map(|(idx, entry)| format!("[{}] {}", idx, entry))
+        .join("\n");
```
tnowacki, line 116: You want an extra \n here at the end. I imagine eventually some linter is going to yell for not having a newline at the end of the file

language/testing-infra/functional-tests/src/testsuite.rs
```diff
@@ -65,31 +80,87 @@ fn print_stage(haystack: &str) -> bool {
...
+fn check_or_update_expected_output(
+    bufwtr: &BufferWriter,
+    output: &mut Buffer,
+    term_width: usize,
+    test_file_path: &Path,
+    exp_file_path: &Path,
+    log: &EvaluationLog,
 ) -> datatest_stable::Result<()> {
-    let input = read_to_string(path)?;
+    // REVIEW: "\n" is probably not very portable.
```
tnowacki, line 110: why not?

vgao1996: Well the usual "\r" vs "\r\n" vs "\n" thing. Mac uses "\r" but understands "\n". Windows uses "\r\n" and sometimes only understands this. Linux sometimes only understands "\n". Just saying there's a potential risk of the result of the comparison being platform dependent and we should be aware of it. Ideally I should let the library handle this... Hmm maybe I can? Let me see if I can use a string buffer and then `wrtieln!` to it.

language/testing-infra/functional-tests/src/evaluator.rs
vgao1996: In general I totally agree with you and to be clear I've already run into some of these issues. I think the real problem here is that the checker based flow requires a log with different properties than the exp file based one. Trying to make them work together is a losing battle to start with. I'd be happy to put in some minimum effort to make it usable as I've already done, but would prefer avoiding going deeper in this rabbit hole and focus on building the new stuff. The new one will simply have a stream you can `write!` into (kinda like the stdout).

language/ir-testsuite/tests/misc/ignore_checks_when_exp_file_present.exp
tnowacki: Normally, I wouldn't care, but the compiler output stuff is just so crazy, that I'm worried it will hurt the ability to quickly approve PRs

tnowacki: As another note, thoughts about by default also printing out the VM Error code? (not the status code)

vgao1996: I've done both.

vgao1996:
@tnowacki I've made the changes you requested. Let me know if this looks good now!

language/testing-infra/functional-tests/src/testsuite.rs
tnowacki: I don't think this really matters. We only have `\n` throughout our rust codebase, I think its fine to carry that over to move and exp files

language/ir-testsuite/tests/misc/ignore_checks_when_exp_file_present.exp
```diff
@@ -0,0 +1,3 @@
+[0], Transaction(0)
+[1], Move VM Status: EXECUTED
```
tnowacki, line 2: Looks good! It would be cool to get the VM Error full dump in there some day, but not needed for this PR

tnowacki (approved):
I think we could definitely improve the printout for transactions

Something like
```
Transaction(0) (Starts on line 0):
Move VM Status: EXECUTED
Transaction Status: Keep(EXECUTED)


Transaction(1) (Starts on line 42):
Move VM Status: EXECUTED
Transaction Status: Keep(EXECUTED)


Transaction(3) (Starts on line 142):
Move VM Status: EXECUTED
Transaction Status: Keep(EXECUTED)
```
is going to be a lot easier to read than
```
[0], Transaction(0)
[1], Move VM Status: EXECUTED
[2], Transaction Status: Keep(EXECUTED)
[3], Transaction(1)
[4], Move VM Status: EXECUTED
[5], Transaction Status: Keep(EXECUTED)
[6], Transaction(2)
[7], Move VM Status: EXECUTED
[8], Transaction Status: Keep(EXECUTED)
```

Nonetheless, this as it is already VASTLY superior to what we have today with `check` (in most cases, obviously `check` wins in some spots). But all things in good time :) 

Let's get it in and iterate on it. Plus as we start using it, I'm sure other people will have more feedback

language/ir-testsuite/tests/misc/ignore_checks_when_exp_file_present.exp
vgao1996: It already does that! I've added a transaction with an `ARITHMETIC_ERROR ` to the test and it prints out `[7], Move VM Status: EXECUTION_FAILURE { status_code: ARITHMETIC_ERROR, location: Script, function_definition: 0, code_offset: 3 }`. (Ignore the `,` after `[7]`. I'll get it removed before landing.)` Counterintuitively, `EXECUTE` is a complete status as there's nothing wrong to report. I had the same doubt when implementing this because it just looked too short to be complete.

tnowacki: That is the VM Status though, Not the VMError, these are distinct things. The VM Error does not propagate past the Move VM (i.e. it is captured in the Libra VM/adaptor)

vgao1996: Ah I see the VMError will be able to provided unfiltered and more granulated information. Seems like a problem to be solved by the Move-only testing flow I'm building now.

tnowacki: It will also be helpful for testing the bytecode verifier, as you get the offsets for the errors (again probably a separate flow there)

PR #5878, opened 2020-09-02
[gha][lbt] build images if not found by compat

rustielin:
Stacked on 9df83f6 (from https://github.com/libra/libra/pull/5701), only review a813bb4

## Overview 

`find-lbt-images.sh` now only looks for recent images until the first
breaking change has been found, since past that, compat is bound to
fail (compat test will be run across versions that are known to be broken). 
If a recent image is not found, report to Slack, but also build
the expected image from `BASE_REF` and continue with the workflow.

This allows land-blocking compat test to be run on non-master branches,
such as for verifying commits cherry-picked into release branches.
Appropriate changes have been made in scripts to account for `BASE_REF` other
than `master`

Bonus: consolidate codebuild stuff into a separate script, since in the worst case we will
need to build two sets of images, which will happen in parallel.

## Test

Canary in https://github.com/libra/libra/pull/5902. See the test plan there. Basically runs through test scenario of making a PR (e.g. for cherrypick) against a non-master branch, and building two sets of images in parallel.

### DIFF
--- a/.github/workflows/land-blocking.yml
+++ b/.github/workflows/land-blocking.yml
@@ -77,30 +77,12 @@ jobs:
           else
             echo "::set-output name=should_run::true";
           fi;
-      - name: Build, tag and push images
-        if: steps.check_ks.outputs.should_run == 'true'
-        run: |
-          set +e
-          date
-          RETRYABLE_EXIT_CODE=2
-          for ((i = 0; i < 3; i++)); do
-            echo "Build attempt $i"
-            docker/build-aws.sh --build-all-cti --version $HEAD_GIT_REV --addl_tags canary,${TEST_TAG}
-            return_code=$?
-            if [[ $return_code -eq 0 ]]; then
-              echo "Build successful"
-              exit 0
-            fi
-            if [[ $return_code -ne ${RETRYABLE_EXIT_CODE} ]]; then
-              echo "Build failed"
-              exit 1
-            fi
-            echo "Retrying build"
-          done
-          echo "Build failed after retries"
-          exit 1
       - name: Determine which cluster-test suite to run
         if: steps.check_ks.outputs.should_run == 'true'
+        # Runs land_blocking if KS activated or [breaking] in commit message heading.
+        # Runs land_blocking_compat otherwise. Compat needs to find an image with PREV_TAG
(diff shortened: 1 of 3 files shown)


.github/workflows/land-blocking.yml
```diff
@@ -89,13 +116,35 @@ jobs:
...
+          VERSION=$HEAD_GIT_REV ADDL_TAG=$TEST_TAG .github/actions/land-blocking/cti-codebuild.sh &> codebuild.log &
+          build_pid=$!
+          if [ $BUILD_PREV -eq 1 ]; then
+            compat_prev_tag=land_$BASE_GIT_REV
+            echo "Starting codebuild for $compat_prev_tag"
+            echo "::set-env name=PREV_TAG::$compat_prev_tag"
+            VERSION=$BASE_GIT_REV ADDL_TAG=$compat_prev_tag .github/actions/land-blocking/cti-codebuild.sh &> codebuild-prev.log &
+            prev_build_pid=$!
+          fi
+          wait $build_pid
```
sausagee, line 140: clever :)

.github/workflows/land-blocking.yml
```diff
@@ -89,13 +116,35 @@ jobs:
...
               curl -X POST -H 'Content-type: application/json' -d @/tmp/payload ${{ secrets.WEBHOOK_FLAKY_LAND_BLOCKING_CT }}
-              exit 1
+              exit 0
             else
               compat_prev_tag=$(tail -1 lbt_images_output.txt)
               echo "Using previous image tag $compat_prev_tag"
               echo "::set-env name=PREV_TAG::$compat_prev_tag"
             fi
           fi
+          echo "::set-env name=BUILD_PREV::0"
```
sausagee, line 126: This goes hand-in-hand with lines 99 and 109. It took me a while to line up the if/else to see what we are doing here. 

Can we add a comment here or beginning of this step to indicate this step will set the condition to build the images from the $BASE_GIT_REV iff they are not found and iff compat is enabled (ie, not being KS)?

sausagee (approved):
Looks good. 

Please address add the comment to the "Determine which cluster-test suite to run" step. 

Remember to canary before land.

PR #5881, opened 2020-09-02
[RFC][libra-framework] Refactored prologue out of LibraAccount

runtian-zhou:
## Motivation

Moved all transaction related logic outside of LibraAccount and refactored prologue and epilogue logic into a single module.

For the LibraAccount module there will be two public apis: `check_sender` and `charge_transaction`, which is the LibraAccount related logic that is extracted out from the old prologue and epilogue.

Justification:
1. All transaction prologue and epilogues will now be grouped into one module. Now Block Prologue is the only function that LibraVM invokes that is outside of this module.
2. Cleaned up the bump_sequence_number ugliness in the writeset epilogue.
3. Move all Libra transaction related logic outside of LibraAccount module.
4. Maximize the code shared across different prologues and epilogues.

I haven't fixed the code in libra_vm invokation so tests are guarenteed to fail. Just want to know how people feel like about this rework. Will fixup tests if people are happy about this.

Yes

## Test Plan

Haven't tested yet.


language/stdlib/modules/LibraTransaction.move
```diff
@@ -0,0 +1,213 @@
+address 0x1 {
+
+module LibraTransaction {
```
runtian-zhou, line 3: This is the newly created module for transaction prologues and epilogues (what used to be `LibraWriteSetManager`).

language/stdlib/modules/LibraAccount.move
```diff
@@ -935,28 +863,32 @@ module LibraAccount {
...
-            PROLOGUE_ETRANSACTION_EXPIRED
-        );
     }
 
     /// Collects gas and bumps the sequence number for executing a transaction
-    fun epilogue<Token>(
-        sender: address,
+    /// This function can only be called during the epilogue as any invokation during the execution body will cause
+    /// the sequence number to be poisoned.
+    public fun charge_transaction<Token>(
```
runtian-zhou, line 871: Note that although this function is public, it is not invokable from regular user scripts as it will cause the sequence number to be incremented and thus fail the assertion at line 885.

jolestar:
It is possible to support a new type of visibility in Move, similar to java's protected or rust's pub(crate), where the methods that use this visibility can only be called by the modules at the same address.

runtian-zhou:
#

> It is possible to support a new type of visibility in Move, similar to java's protected or rust's pub(crate), where the methods that use this visibility can only be called by the modules at the same address.

I'd really love for support like that. But IMHO we don't have a clear path towards it anytime soon as supporting such feature will actually need to change the bytecode format + bytecode verifier, which seems to be a bit heavy right now.

bob-wilson:
I really like this change, but I have reservations about doing this kind of refactoring so long after our feature-complete date. What about if we put all the prologue/epilogue code into LibraAccount (where most of it is already) instead of creating a new module? It wouldn't be as nicely factored but it seems like it would achieve most of the benefits with far less churn in the code.

runtian-zhou:
> I really like this change, but I have reservations about doing this kind of refactoring so long after our feature-complete date. What about if we put all the prologue/epilogue code into LibraAccount (where most of it is already) instead of creating a new module? It wouldn't be as nicely factored but it seems like it would achieve most of the benefits with far less churn in the code.

I think the major concern is that we need to expose `charge_transaction` even if we keep the current structure of code. The only gain that buys us is that the LibraVM will remain unchanged.

bob-wilson:
> I think the major concern is that we need to expose `charge_transaction` even if we keep the current structure of code. The only gain that buys us is that the LibraVM will remain unchanged.

Why? Wouldn't all the callers be in the LibraAccount module so that `charge_transaction` could be private?

language/stdlib/modules/LibraTransaction.move
```diff
@@ -0,0 +1,213 @@
...
+
+    fun writeset_prologue(
+        account: &signer,
+        writeset_sequence_number: u64,
+        writeset_public_key: vector<u8>,
+    ) {
+        assert(sender == CoreAddresses::LIBRA_ROOT_ADDRESS(), PROLOGUE_EINVALID_WRITESET_SENDER);
+
+        // Currency code don't matter here as it won't be charged anyway.
+        LibraAccount::check_sender<LBR::LBR>(account, 0, writeset_sequence_number, writeset_public_key);
```
sblackshear, line 154: Should we call `prologue_common` here instead of this to resolve the TODO below?

language/stdlib/modules/LibraTransaction.move
```diff
@@ -0,0 +1,213 @@
...
+    }
+
+    fun writeset_prologue(
+        account: &signer,
+        writeset_sequence_number: u64,
+        writeset_public_key: vector<u8>,
+    ) {
+        assert(sender == CoreAddresses::LIBRA_ROOT_ADDRESS(), PROLOGUE_EINVALID_WRITESET_SENDER);
+
+        // Currency code don't matter here as it won't be charged anyway.
```
sblackshear, line 153: This + the 0 gas price feels a bit icky and is making me wonder whether we want separate `LibraAccount::gas_prologue` and `LibraAccount::replay_prevention_prologue`s (or similar...)

language/stdlib/modules/LibraTransaction.move
```diff
@@ -0,0 +1,213 @@
...
+        txn_gas_price: u64,
+        txn_max_gas_units: u64,
+        gas_units_remaining: u64
+    ) {
+        let sender = Signer::address_of(account);
+
+        // Charge for gas
+        assert(txn_max_gas_units >= gas_units_remaining, Errors::invalid_argument(EGAS));
+        let gas_used = txn_max_gas_units - gas_units_remaining;
+        assert((txn_gas_price as u128) * (gas_used as u128) <= MAX_U64, Errors::limit_exceeded(EGAS));
```
sblackshear, line 171: Maybe do `assert(txn_gas_price <= MAX_U64 / gas_used)` to avoid the `u128` conversions?

language/stdlib/modules/LibraTransaction.move
```diff
@@ -0,0 +1,213 @@
...
+    fun success_epilogue<Token>(
+        account: &signer,
+        txn_sequence_number: u64,
+        txn_gas_price: u64,
+        txn_max_gas_units: u64,
+        gas_units_remaining: u64
+    ) {
+        let sender = Signer::address_of(account);
+
+        // Charge for gas
```
sblackshear, line 168: I find this comment confusing given that the actual charging happens in `LibraAccount`. This part is just sanity-checking the gas price and gas unit values.

language/stdlib/modules/LibraTransaction.move
```diff
@@ -0,0 +1,213 @@
...
+        let gas_used = txn_max_gas_units - gas_units_remaining;
+        assert((txn_gas_price as u128) * (gas_used as u128) <= MAX_U64, Errors::limit_exceeded(EGAS));
+        let transaction_fee_amount = txn_gas_price * gas_used;
+
+        LibraAccount::charge_transaction<Token>(sender, transaction_fee_amount, txn_sequence_number);
+    }
+
+    /// The failure_epilogue is invoked at the end of transactions when the transaction is aborted during execution or
+    /// during `success_epilogue`.
+    fun failure_epilogue<Token>(
```
sblackshear, line 179: Agreed with you that these are now the same and we can kill one of them.

tnowacki:
Also very worried about making a change like this so late. 

> Why? Wouldn't all the callers be in the LibraAccount module so that charge_transaction could be private?

Do we have an answer on this? I think that if we don't need to expose new public APIs, we shouldn't do it. Even if it means more code duplication. 
In short, I don't really yet buy many of the motivations outlined in the summary, but could be convinced otherwise


Additionally, we should really re-evaluate this after https://github.com/libra/libra/pull/5876. There would only need to be a single epilogue and might help us look at things differently

language/stdlib/modules/LibraTransaction.move
```diff
@@ -0,0 +1,213 @@
...
+        txn_sequence_number: u64,
+        txn_public_key: vector<u8>,
+        txn_gas_price: u64,
+        txn_max_gas_units: u64,
+        txn_expiration_time: u64,
+        chain_id: u8,
+    ) {
+        assert(
+            LibraTransactionPublishingOption::is_module_allowed(sender),
+            PROLOGUE_EMODULE_NOT_ALLOWED
```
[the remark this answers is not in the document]
tnowacki: https://github.com/libra/libra/pull/5876

I have a PR up to use Errors everywhere. We should probably get that in first

PR #5882, opened 2020-09-02
[dkr] support failpoint build for libra-node

zekun000:
## Motivation

build scripts will enable the feature flag for libra-node when the
ENABLE_FAILPOINTS env is set to 1.

Y

## Test Plan

build

## Related PRs

### DIFF
--- a/docker/build-common.sh
+++ b/docker/build-common.sh
@@ -26,6 +26,12 @@ ${CARGO} ${CARGOFLAGS} build --release \
          -p backup-cli \
          "$@"
 
+# Build and overwrite the libra-node binary with feature failpoints if $ENABLE_FAILPOINTS is configured
+if [ "$ENABLE_FAILPOINTS" = "1" ]; then
+  echo "Building libra-node with failpoints feature"
+  (cd libra-node && ${CARGO} ${CARGOFLAGS} build --release --features failpoints "$@")
+fi
+
 # These non-release binaries are built separately to avoid feature unification issues
 ${CARGO} ${CARGOFLAGS} build --release \
          -p cluster-test \
--- a/docker/libra-build.sh
+++ b/docker/libra-build.sh
@@ -14,7 +14,7 @@ RESTORE='\001\033[0m\002'
 BLUE='\001\033[01;34m\002'
 DOCKERFILE_BUILD_RE='^RUN ./docker/build-common.sh$'
 
-if [ -z "$DOCKERFILE" ] || [ -z "$TAG" ]
+if [ -z "$DOCKERFILE" ] || [ -z "$TAGS" ]
 then
   echo "Usage libra-build.sh <Docker_file> <Local_tag> <Other_args>"
 fi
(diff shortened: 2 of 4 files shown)


docker/libra-build.sh
```diff
@@ -4,17 +4,20 @@
...
-  TAGS="--tag $2 --tag $LIBRA_BUILD_TAG"
+  TAGS="--tag $2$TAG_SUFFIX --tag $LIBRA_BUILD_TAG$TAG_SUFFIX"
 fi
 
 RESTORE='\001\033[0m\002'
 BLUE='\001\033[01;34m\002'
 DOCKERFILE_BUILD_RE='^RUN ./docker/build-common.sh$'
 
-if [ -z "$DOCKERFILE" ] || [ -z "$TAG" ]
+if [ -z "$DOCKERFILE" ] || [ -z "$TAGS" ]
```
davidiw, line 20: how is this currently passing? TAG must be defined somewhere, no? and then what the hell is $TAGS doing? Would recommend applying the suffix on old line 12 without adding an additional variable...


```
if [ -n "$ENABLE_FAILPOINTS" ]; then
  TAGS="$TAGS""_failpoints"
fi

File: docker/libra-build.sh [hunk 2]
```diff
@@ -4,17 +4,20 @@
 set -e
 
 DOCKERFILE=$1
+if [ -n "ENABLE_FAILPOINTS" ]; then
```
davidiw, line 7: missing a $ before E

docker/build-common.sh
```diff
@@ -26,6 +26,12 @@ ${CARGO} ${CARGOFLAGS} build --release \
          -p backup-cli \
          "$@"
 
+# Build and overwrite the libra-node binary with feature failpoints if $ENABLE_FAILPOINTS is configured
+if [ "$ENABLE_FAILPOINTS" = "1" ]; then
+  echo "Building libra-node with failpoints feature"
+  (cd libra-node && ${CARGO} ${CARGOFLAGS} build --features failpoints "$@")
```
andll, line 32: (1) In step above we use `-p libra-node` instead of `cd`. I think its best to do things uniformly

(2) This line is missing `--release` flag, this is actually a big difference for libra-node. Probably need to introduce a flag like `${COMMON_CARGO_BUILD_ARGS}`, put `--release` there and use it here, below when building ct, and in line above.

zekun000: (1) doesn't work, we can't enable feature flag at the workspace root.
(2) good catch

zekun000:
okay it finally works in cti with ENABLE_FAILPOINTS=1 locally. I can see "Failpoints is enabled" in the logs

andll:
Just to be on paranoid side lets also confirm that without setting this env var failpoints are not enabled?

zekun000:
> Just to be on paranoid side lets also confirm that without setting this env var failpoints are not enabled?

yeah i tested this too

andll (approved):

PR #5883, opened 2020-09-02
[management] improve all tools

davidiw:
Composite PR that covers a whole host of outstanding issues:

* Add ability to read keys from files so libra / treasury root can submit their keys from disk
* Expose the two key support in genesis encoding
* Add support for a distinct treasury key for generating genesis
* Simplify (or improve) the output from the genesis tool
* Cleanup a lingering todo in building genesis on disallowing any script to be run

cc: @ankushagarwal / @dmitri-perelman from AOS -- API supports LCS or hex encoded keys
cc: @mgorven / @sherry-x for OPS -- this requires that we add an additional flag to generating genesis for testnets namely treasury-root-key which is the same interface as libra-root-key, let's sync on this since we will definitely need an update to the ops repo


davidiw:
Might take that last todo off if I get an accept before I can fix it... cc: @zekun000 ;)

config/management/genesis/src/command.rs
```diff
@@ -290,7 +322,8 @@ pub mod tests {
         let layout_text = "\
             operators = [\"alice\", \"bob\"]\n\
             owners = [\"carol\"]\n\
-            libra_root = [\"dave\"]\n\
+            libra_root = \"dave\"\n\
+            treasury_root = \"other_dave\"\n\
```
gregnazario, line 326: Interesting, so treasury root is the same account, just a different key?

config/management/genesis/src/genesis.rs
```diff
@@ -41,15 +41,16 @@ impl Genesis {
...
+        let treasury_root_key = self.treasury_root_key(&layout)?;
         let operator_assignments = self.operator_assignments(&layout)?;
         let operator_registrations = self.operator_registrations(&layout)?;
 
         let genesis = vm_genesis::encode_genesis_transaction(
             libra_root_key,
+            treasury_root_key,
             &operator_assignments,
             &operator_registrations,
-            // TODO: swap back by 8/15
```
gregnazario, line 51: Hehe, that has come and passed.

language/tools/vm-genesis/src/lib.rs
```diff
@@ -519,6 +525,7 @@ pub fn generate_genesis_change_set_for_testing(stdlib_options: StdLibOptions) ->
     let swarm = libra_config::generator::validator_swarm_for_testing(10);
 
     encode_genesis_change_set(
+        &GENESIS_KEYPAIR.1,
         &GENESIS_KEYPAIR.1,
```
gregnazario, line 529: Is the treasury and libra root always the same?

config/management/src/secure_backend.rs
```diff
@@ -152,13 +153,21 @@ pair: "k0=v0;k1=v1;...".  The current supported formats are:
     }
 }
 
+// "ignore" allows this to be used elsewhere to provide for greater richness when required
```
gregnazario, line 156: I'm not quite sure what the greater richness is?

config/management/genesis/src/storage_helper.rs
```diff
@@ -47,6 +47,11 @@ impl StorageHelper {
 
         // Initialize all keys in storage
         storage.create_key(LIBRA_ROOT_KEY).unwrap();
+        // TODO(davidiw) use distinct keys in tests for treasury and libra root keys
```
gregnazario, line 50: Ah yeah, I saw this everywhere and was very confused.

config/management/genesis/src/genesis.rs
davidiw: of course, all todos with a date always get cleaned up lifetimes later... and of course @zekun000 is going to force me to make this into another PR :P

language/tools/vm-genesis/src/lib.rs
davidiw: in our tests, unfortunately... if I have time, I'll update smoketest so that it can use two keys that way we get more confidence before next week... but I'll probably delay on this until later...

config/management/src/secure_backend.rs
davidiw: look at keys.rs, this allows us to pull from disk directly without having to make a fake storage

config/global-constants/src/lib.rs
```diff
@@ -12,7 +12,8 @@
 pub const CONSENSUS_KEY: &str = "consensus";
 pub const EXECUTION_KEY: &str = "execution";
 pub const FULLNODE_NETWORK_KEY: &str = "fullnode_network";
-pub const LIBRA_ROOT_KEY: &str = "root";
+pub const LIBRA_ROOT_KEY: &str = "libra_root";
+pub const TREASURY_ROOT_KEY: &str = "treasury_root";
```
ankushagarwal, line 16: Can we call this TREASURY_**COMPLIANCE**_KEY and similar name elsewhere?

This account is referred to as the TreasuryCompliance account everywhere else : eg: https://github.com/libra/libra/blob/master/language/stdlib/modules/Roles.move

config/management/genesis/src/command.rs
ankushagarwal: the treasury account is a different account from the libra root account

ankushagarwal:
> Add ability to read keys from files so libra / treasury root can submit their keys from disk

Can you update the [readme](https://github.com/libra/libra/blob/master/config/management/genesis/README.md) with an example?

config/management/src/secure_backend.rs
JoshLind: I'm still confused by this comment 😕

config/management/src/secure_backend.rs
```diff
@@ -127,13 +127,14 @@ impl TryInto<config::SecureBackend> for SecureBackend {
...
 macro_rules! secure_backend {
-    ($struct_name:ident, $field_name:ident, $purpose:literal) => {
+    ($struct_name:ident, $field_name:ident, $purpose:literal, $required:literal) => {
         #[derive(Clone, Debug, StructOpt)]
         pub struct $struct_name {
             #[structopt(long,
                 help = concat!("Backend for ", $purpose),
-                required_unless("config"),
+                required_unless_one(&["config", $required]),
                 long_help = concat!("Backend for ", $purpose, r#"
```
gregnazario, line 138: You could make this support without the `"ignore"` field by making a second macro path for 

```
($struct_name:ident, $field_name:ident, $purpose:literal) => {
```

config/management/src/secure_backend.rs
davidiw: hidden within the macro itself now

davidiw:
> > Add ability to read keys from files so libra / treasury root can submit their keys from disk
> 
> Can you update the [readme](https://github.com/libra/libra/blob/master/config/management/genesis/README.md) with an example?

I'll swing around once these land to cover that!

PR #5885, opened 2020-09-02
[ci] Remove checked in summaries

metajack:
This removes the checked in summaries. The CI workflow is also updated not to ensure these are kept up to date.

These were checked in to provide a trigger (via `CODEOWNERS`) for bors to latch onto in order to require extra review. However having these generated files in the repo causes problems for other workflows like dependabot, so we've decided to achieve the same goal by an another means.

Since bors was not currently enforcing `CODEOWNERS` review this has no real change to our CI aside from no longer notifying `CODEOWNERS` on the summary changes. Labeling/commenting of dependency changes will still happen.


.github/CODEOWNERS
```diff
@@ -7,6 +7,3 @@
 /client/ @thefallentree @xli
 # Subscribe to LCS format changes
 /testsuite/generate-format/tests/staged/libra.yaml @thefallentree @matbd @xli
-# Subscribe to TCB dependency summary changes
-/summaries/summary-lsr.toml @davidiw
```
davidiw, line 11: oh, this explains a lot :P

I agree that without /requiring/ a review from a codeowner this can appear to be noise :/

PR #5887, opened 2020-09-02
[genesis] actually push names into validators and operators

davidiw:
* Existing genesis creation does not introduce names, that is fixed
* Existing genesis requires owner keys but we cannot expect that.
* Improve operational tooling around validator configs to include names
* Use names to support add / remove validators
* Remove cli stuff around add / remove validators as it is broken

Traditionally, account addresses are derived from authentication keys, which in turn are derived from the public key. If we do not have a public key, we must do it some way else. So we derive a consistent address from the owner name. We then set the authentication key from either the Some(key) or set it to 0; 32

cc: @ankushagarwal , @dmitri-perelman , @huitseeker , @kchalkias  we indeed can set auth account to all 0s! Thanks for your time and thoughts :)

### DIFF
--- a/config/management/operational/src/governance.rs
+++ b/config/management/operational/src/governance.rs
@@ -40,6 +40,9 @@ impl AddValidator {
 
         // Verify that this is a configured validator
         client.validator_config(self.input.account_address)?;
+        let name = client
+            .validator_config(self.input.account_address)?
+            .human_name;
 
         // Prepare a transaction to add them to the ValidatorSet
         let seq_num = client.sequence_number(libra_root_address())?;
@@ -49,7 +52,7 @@ impl AddValidator {
             seq_num,
             transaction_builder::encode_add_validator_and_reconfigure_script(
                 seq_num,
-                vec![],
+                name,
                 self.input.account_address,
             ),
         );
@@ -73,6 +76,9 @@ impl RemoveValidator {
 
         // Verify that this is a validator within the set
         client.validator_set(Some(self.input.account_address))?;
+        let name = client
+            .validator_config(self.input.account_address)?
+            .human_name;
 
         // Prepare a transaction to remove them from the ValidatorSet
         let seq_num = client.sequence_number(libra_root_address())?;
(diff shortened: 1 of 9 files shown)


config/management/operational/src/governance.rs
```diff
@@ -40,6 +40,9 @@ impl AddValidator {
 
         // Verify that this is a configured validator
         client.validator_config(self.input.account_address)?;
+        let name = client
+            .validator_config(self.input.account_address)?
+            .human_name;
```
JoshLind, line 45: Is there any reason you can't merge the two 'client.validator_config(..)?' calls? e.g., just remove the previous line 42 which calls the validator_config() method but does nothing with the return value?

testsuite/tests/libratest/smoke_test.rs
```diff
@@ -1212,15 +1212,23 @@ fn test_client_waypoints() {
...
 
     // Start next epoch
-    let peer_id = env
-        .get_validator(0)
-        .unwrap()
-        .validator_peer_id()
-        .unwrap()
-        .to_string();
+
+    // This ugly blob is to remove a validator, we can do better...
```
JoshLind, line 1216: Indeed.. the smoke tests in general could use some work 😉

JoshLind (approved):
Thanks, @davidiw! 😄 Other than my comments, it looks good to me. Happy to also review the next PR with the missing commit!

config/management/operational/src/governance.rs
davidiw: woops

config/management/operational/src/json_rpc.rs
```diff
@@ -41,13 +41,14 @@ impl JsonRpcClientWrapper {
             .map_err(|e| Error::JsonRpcReadError("account-state", e.to_string()))
     }
 
-    pub fn validator_config(&self, account: AccountAddress) -> Result<ValidatorConfig, Error> {
+    pub fn validator_config(
```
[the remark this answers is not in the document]
davidiw: I don't think we really distinguish the two in common vernacular ... let me check how many uses there are and figure it out.

davidiw:
@JoshLind , thanks! Hopefully the last pr is almost done.. it is ... nuanced..

PR #5889, opened 2020-09-03
Adding address to AccountView

thefallentree:
to assist client in situation where response handler  doesn't have access to request object


xli (approved):

thefallentree:
@rustielin  any idea why this would be a compat failure?  adding a field to jsonrpc should not be :-D

Issue #5675, opened 2020-08-19
[Feature Request] MPSC-like Framework for Move Prover 


emmazzz:
# 🚀 Feature Request

## Motivation

Due to the butterfly effect we have observed, using different random seeds and ordering quantifier declarations differently might cause huge differences in the performance of the prover. Thus we could try to improve the performance of the prover by creating a framework which allows us to run multiple instances of the solver task simultaneously and returns as soon as the fastest instance finishes. This is similar to MPSC(multiple producer, single consumer) since multiple worker threads will work on the same task(with some small twists) and report to one consumer(the Move Prover).

## Pitch

**Describe the solution you'd like**

It will probably be the easiest if we implement the aforementioned framework in Rust. 

We can define a `trait ProverTaskRunner` with methods `init` and `run`.
```rust
pub trait ProverTaskRunner {
    type TaskResult;
    type TaskId;

    /// Initialize the task runner given that there will be num_instances number of instances.
    fn init(&mut self, num_instances: usize) -> Vec[TaskId];

    /// Run the task with task_id. This function will be called from one the the threads.
    fn run(&mut self, task_id: TaskId) -> TaskResult;
}
```  

The framework then can call the methods implemented by `ProverTaskRunner`s.
```rust
pub trait ProverTaskRunningFramework: ProverTaskRunner {
    fn run_tasks(&mut self, num_instances: usize) -> Result<TaskResult> {
        let (sender, receiver) = channel(); // creates a channel for communication        
        let task_ids = self.init(num_instances);
        for task_id in task_ids {
            let sender_n = sender.clone();
            thread::spawn(move ||{
                let res = self.run(task_id);
                sender_n.send(res);
            });
        } 
        loop until times out {
            let res = receiver.recv();
            if !res.is_err() {
                return Ok(res);
            }
        }
        return Error;
    }
}
```

**Describe alternatives you've considered**

Alternatively, this framework can be implemented on Boogie or Z3 level.

PR #5890, opened 2020-09-03
[move prover] implement async framework running multiple boogie instances

emmazzz:
This PR implements a framework that runs multiple prover task instances(Boogie commands) concurrently and returns the result of the fastest running instance. The framework is general in that the `run_tasks` function takes in a `task` parameter of any type that implements the newly created `ProverTask` trait. 

In this PR, we use the framework to run Boogie with different random seeds by passing in a task of type `RunBoogieWithSeeds` which implements `ProverTask` trait. When running on complicated formulas(especially those with quantifiers), SMT solvers can suffer from the so-called butterfly effect, where minor changes such as using different random seed cause significant instabilities in verification times. Thus by running multiple instances of Boogie with different random seeds, we can potentially alleviate this instability. See `move-prover/src/prover-task-runner.rs` for more details.

To use this feature, run the prover with flag `--num-instances` followed by the number of instances to run, such as 
`--num-instances 5`. The default value is 1. 

Addresses issue #5675.

### DIFF
--- a/language/move-prover/src/prover_task_runner.rs
+++ b/language/move-prover/src/prover_task_runner.rs
@@ -1,0 +1,154 @@
+// Copyright (c) The Libra Core Contributors
+// SPDX-License-Identifier: Apache-2.0
+
+//! Prover task runner that runs multiple instances of the prover task and returns
+//! as soon as the fastest instance finishes.
+
+use crate::cli::Options;
+use async_trait::async_trait;
+use futures::{future::FutureExt, pin_mut, select};
+use rand::Rng;
+use std::{
+    process::Output,
+    sync::mpsc::{channel, Sender},
+};
+use tokio::{
+    process::Command,
+    runtime::Runtime,
+    sync::{broadcast, broadcast::Receiver},
+};
+
+#[derive(Debug, Clone)]
+enum BroadcastMsg {
+    Stop,
+}
+
+#[async_trait]
+pub trait ProverTask {
+    type TaskResult: Send + 'static;
+    type TaskId: Send + Copy + 'static;
+
+    /// Initialize the task runner given the number of instances.
+    fn init(&mut self, num_instances: usize) -> Vec<Self::TaskId>;
+
+    /// Run the task with task_id. This function will be called from one of the worker threads.
+    async fn run(&mut self, task_id: Self::TaskId) -> Self::TaskResult;
+}
+
+pub struct ProverTaskRunner();
+
+impl ProverTaskRunner {
+    /// Run `num_instances` instances of the prover `task` and returns the task id
+    /// as well as the result of the fastest running instance.
+    pub fn run_tasks<T>(mut task: T, num_instances: usize) -> (T::TaskId, T::TaskResult)
+    where
+        T: ProverTask + Clone + Send + 'static,
+    {
+        let rt = Runtime::new().unwrap();
+
+        // Create channels for communication.
+        let (worker_tx, master_rx) = channel();
+        let (master_tx, _): (
+            tokio::sync::broadcast::Sender<BroadcastMsg>,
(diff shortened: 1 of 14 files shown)


language/move-prover/src/prover_task_runner.rs
```diff
@@ -0,0 +1,162 @@
...
+}
+
+pub struct ProverTaskRunner();
+
+impl ProverTaskRunner {
+    /// Run `num_instances` instances of the prover `task` and returns the task id
+    /// as well as the result of the fastest running instance.
+    pub fn run_tasks<T>(mut task: T, num_instances: usize) -> (T::TaskId, T::TaskResult)
+    where
+        Self: Sync,
```
shazqadeer, line 46: Random Rust question: Why is "Self: Sync" needed here?  There is no self parameter to the function run_tasks.

language/move-prover/src/prover_task_runner.rs
```diff
@@ -0,0 +1,162 @@
...
+    sync::{broadcast, broadcast::Receiver},
+};
+
+#[derive(Debug, Clone)]
+enum BroadcastMsg {
+    Stop,
+}
+
+#[async_trait]
+pub trait ProverTask {
```
shazqadeer, line 28: I just realized that this trait is closer to a ProverTaskFactory rather than a single ProverTask.

language/move-prover/src/prover_task_runner.rs
```diff
@@ -0,0 +1,162 @@
...
+                    send_n,
+                    worker_rx,
+                ));
+            });
+        }
+        // Listens until one of the workers finishes.
+        loop {
+            let res = master_rx.recv();
+            if let Ok((task_id, result)) = res {
+                // Result received from one worker. Broad cast to other workers
```
shazqadeer, line 78: Broad cast

language/move-prover/src/prover_task_runner.rs
```diff
@@ -0,0 +1,162 @@
...
+    async fn run_task_until_cancelled<T>(
+        mut task: T,
+        task_id: T::TaskId,
+        tx: Sender<(T::TaskId, T::TaskResult)>,
+        rx: Receiver<BroadcastMsg>,
+    ) where
+        Self: Sync,
+        T: ProverTask,
+    {
+        let boogie_fut = task.run(task_id).fuse();
```
shazqadeer, line 99: what does fut stand for?

emmazzz: future

language/move-prover/src/prover_task_runner.rs
emmazzz: That's a good point. It's not needed here.

wrwg (approved):

language/move-prover/src/prover_task_runner.rs
```diff
@@ -0,0 +1,160 @@
...
+
+        // Initialize the prover tasks.
+        let task_ids = task.init(num_instances);
+        for task_id in task_ids {
+            let send_n = worker_tx.clone();
+            let worker_rx = master_tx.subscribe();
+            let cloned_task = task.clone();
+            // Spawn a task worker for each task_id.
+            rt.spawn(async move {
+                task::block_on(Self::run_task_until_cancelled(
```
bmwill, line 65: You probably want to call `.await` on this instead of using `task::block_on` from the `async-std` crate. In fact if thats the only use of the `async-std` crate you might want to drop it as a dependency.

bob-wilson:
It would be nice to have a comment somewhere to explain the motivation for this scheme. It isn't at all obvious that Boogie's execution time is sensitive to a random seed, and that it can sometimes hit particularly bad points where it takes much longer to run.

emmazzz:
\land

PR #5891, opened 2020-09-03
[ErrorWithContext] Create a new ErrorWithContext type and add it to the operational tool.

JoshLind:
## Motivation

This PR adds a new ErrorWithContext type: a wrapper around std::error::Error that provides an additional struct in which to hold auxiliary information about the context of an error object, e.g., where the error originated from in the source code, such as the error backtrace. This is useful for:
- Debugging, e.g., when an error is thrown or displayed on the console but the error message isn't helpful enough to identify exactly what went wrong. So instead of having to grep for the failure message in the code and manually figure out what happened, we can see the backtrace! 😄 
- Updating `panic!` related code to use errors instead of forcing panics and not lose context (e.g., see the conversation in this issue regarding how removing panics prevents the backtrace from being displayed: https://github.com/libra/libra/pull/5617). ErrorWithContext provides the best of both worlds between panic and error handling.

This PR offers three commits:
1. First, we create the ErrorWithContext type and place it in common.
2. Second, we update the Operational tool to use the ErrorWithContext type as a way to show how it works in practice.
3. Third, we update the summaries in the code (as required by land blocking tests)

See the section below for how the ErrorWithContext type is formatted when printed.

Yes.

## Test Plan

When using the operational tool and encountering an error,  the tool successfully prints out the backtrace of the Error (i.e., where the error was created, and thus the error failure point!). See the example below (this is without using any backtrace env variables!).

Without ErrorWithContext:

```
joshlind@joshlind-mbp debug % ./libra-operational-tool account-resource --json-server <JSON SERVER> --account-address 00000000000000000000000000000000
{
  "Error": "Failed to read 'account-state' from JSON-RPC: Data does not exist. Missing data: AccountState"
}
```

Above, we can see that the debuggability of the error just comes down to how clear the error message is. In many cases, it's not super helpful without knowing exactly what's going on and digging through the code... Now with ErrorWithContext:

```
joshlind@joshlind-mbp debug % ./libra-operational-tool account-resource --json-server <JSON SERVER> --account-address 00000000000000000000000000000000
{
  "Error": "Failed to read 'account-state' from JSON-RPC: Data does not exist. Missing data: AccountState"
}
Backtrace:
    0: libra_error_with_context::ErrorWithContext<T>::new
             at /Users/joshlind/Facebook/Libra/Code/fork/libra/common/error-with-context/src/lib.rs:28
   1: <libra_error_with_context::ErrorWithContext<T> as core::convert::From<T>>::from
             at /Users/joshlind/Facebook/Libra/Code/fork/libra/common/error-with-context/src/lib.rs:87
   2: <T as core::convert::Into<U>>::into
             at /Users/joshlind/.rustup/toolchains/1.45.2-x86_64-apple-darwin/lib/rustlib/src/rust/src/libcore/convert/mod.rs:559
   3: libra_operational_tool::json_rpc::JsonRpcClientWrapper::account_state::{{closure}}
             at /Users/joshlind/Facebook/Libra/Code/fork/libra/config/management/operational/src/json_rpc.rs:41
   4: core::result::Result<T,E>::map_err
             at /Users/joshlind/.rustup/toolchains/1.45.2-x86_64-apple-darwin/lib/rustlib/src/rust/src/libcore/result.rs:612
   5: libra_operational_tool::json_rpc::JsonRpcClientWrapper::account_state
             at /Users/joshlind/Facebook/Libra/Code/fork/libra/config/management/operational/src/json_rpc.rs:39
   6: libra_operational_tool::json_rpc::JsonRpcClientWrapper::account_resource
             at /Users/joshlind/Facebook/Libra/Code/fork/libra/config/management/operational/src/json_rpc.rs:102
   7: libra_operational_tool::account_resource::AccountResource::execute
             at /Users/joshlind/Facebook/Libra/Code/fork/libra/config/management/operational/src/account_resource.rs:37
   8: libra_operational_tool::command::Command::execute
             at /Users/joshlind/Facebook/Libra/Code/fork/libra/config/management/operational/src/command.rs:120
   9: libra_operational_tool::main
             at /Users/joshlind/Facebook/Libra/Code/fork/libra/config/management/operational/src/main.rs:9
  10: std::rt::lang_start::{{closure}}
             at /Users/joshlind/.rustup/toolchains/1.45.2-x86_64-apple-darwin/lib/rustlib/src/rust/src/libstd/rt.rs:67
  11: std::rt::lang_start_internal::{{closure}}
             at /rustc/d3fb005a39e62501b8b0b356166e515ae24e2e54/src/libstd/rt.rs:52
      std::panicking::try::do_call
             at /rustc/d3fb005a39e62501b8b0b356166e515ae24e2e54/src/libstd/panicking.rs:297
      std::panicking::try
             at /rustc/d3fb005a39e62501b8b0b356166e515ae24e2e54/src/libstd/panicking.rs:274
      std::panic::catch_unwind
             at /rustc/d3fb005a39e62501b8b0b356166e515ae24e2e54/src/libstd/panic.rs:394
      std::rt::lang_start_internal
             at /rustc/d3fb005a39e62501b8b0b356166e515ae24e2e54/src/libstd/rt.rs:51
  12: std::rt::lang_start
             at /Users/joshlind/.rustup/toolchains/1.45.2-x86_64-apple-darwin/lib/rustlib/src/rust/src/libstd/rt.rs:67
  13: _main
```

Frame 3 shows you exactly where the error was created and what failed. 😄 

## Related PRs

If we're happy to land this, I'll follow up with another PR that does the same for the genesis tooling.


common/error-with-context/src/lib.rs
```diff
@@ -0,0 +1,130 @@
...
+/// The string displayed in the backtrace frame that shows where the ErrorWithContext object was
+/// initialized. We use this for truncating the backtrace to display only the most helpful backtrace
+/// frames.
+const CREATION_FRAME_STRING: &str = "ErrorWithContext<T>::new";
+
+impl<T: Error> ErrorWithContext<T> {
+    /// Creates and returns an ErrorWithContext struct.
+    pub fn new(error: T) -> ErrorWithContext<T> {
+        // Grab the current backtrace
+        let mut backtrace = Backtrace::new();
```
davidiw, line 28: you probably want a macro here, since this will include this new function.

common/error-with-context/src/lib.rs
```diff
@@ -0,0 +1,130 @@
...
+            let mut creation_frame_index = 0;
+            for index in 0..backtrace_frames.len() {
+                if let Some(frame) = backtrace_frames.get(index) {
+                    let frame_string = format!("{:?}", frame.symbols());
+                    if frame_string.contains(CREATION_FRAME_STRING) {
+                        creation_frame_index = index;
+                    }
+                }
+            }
+            backtrace = backtrace_frames[creation_frame_index..].to_vec().into();
```
davidiw, line 44: then you can eliminate this weirdness

common/error-with-context/src/lib.rs
```diff
@@ -0,0 +1,130 @@
...
+        format!(
+            "{}\n{}",
+            self.format_error_no_context(),
+            self.format_context_no_error()
+        )
+    }
+
+    /// Returns a string that displays the error without a context (i.e., just displays the error).
+    pub fn format_error_no_context(&self) -> String {
+        format!("Error:\n {:?}", self.error)
```
davidiw, line 61: maybe error.to_string()?

common/error-with-context/src/lib.rs
```diff
@@ -0,0 +1,130 @@
...
+                }
+            }
+            backtrace = backtrace_frames[creation_frame_index..].to_vec().into();
+        }
+
+        ErrorWithContext { error, backtrace }
+    }
+
+    /// Returns a string that displays the error and the context.
+    pub fn format_error_and_context(&self) -> String {
```
davidiw, line 51: why not call it backtrace instead of context?

common/error-with-context/src/lib.rs
```diff
@@ -0,0 +1,130 @@
...
+    /// Returns a string that displays just the error context (not the error).
+    pub fn format_context_no_error(&self) -> String {
+        format!("Backtrace:\n {:?}", self.backtrace)
+    }
+}
+
+/// Debug shows the ErrorWithContext by displaying the error as well as the underlying context.
+impl<T: Error> fmt::Debug for ErrorWithContext<T> {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(f, "{}", self.format_error_and_context(),)
```
davidiw, line 73: nit weird comma

davidiw: in general I wouldn't give a backtrace unless it was explicitly asked for.

common/error-with-context/src/lib.rs
davidiw: also even if you don't want a macro, I wouldn't worry about the extra stuff, though I'd be curious to see what you're clipping here...

testsuite/cluster-test/src/genesis_helper.rs
```diff
@@ -326,8 +326,12 @@ impl GenesisHelper {
         );
 
         let command = OperationalCommand::from_iter(args.split_whitespace());
-        spawn_blocking(|| command.extract_private_key())
-            .await
-            .expect("tokio spawn_blocking runtime error")
+        spawn_blocking(|| {
+            command
+                .extract_private_key()
+                .map_err(|annotated_error| annotated_error.error)
```
davidiw, line 332: make it small again... map_err(|e| e.error) ... though wy does this matter? I don't know why we need this :O

bmwill:
I'd recommend looking at eyre or anyhow instead of reimplementing behavior that these libraries already handle so well

JoshLind:
> I'd recommend looking at eyre or anyhow instead of reimplementing behavior that these libraries already handle so well

Thanks @bmwill. I wasn't aware of eyre! Although, from what I can tell:
- anyhow requires nightly and also for you to set the various RUST_BACKTRACE* env variables.
- eyre, also requires you setting the various RUST_BACKTRACE* env variables (https://docs.rs/eyre/0.6.0/eyre/).

I guess the issue is whether or not we think it's a good assumption for operational tool users to be required to set these variables, or if we should do it programmatically.. Looking at it now though, it might be better to adopt one of these and just require the env variables to be set. I'd rather avoid re-inventing the wheel and having to manage this debt going forward. 

@davidiw, I'll just close this PR and can revisit this idea sometime in the future. It would be nice to have backtraces for the operational tool (but then again, it would be nice to have a lot of things 😄). As @bmwill pointed out, this probably isn't the best approach in which to do it. Note: we'll also need to see how bad these deps are using 'dep-hell', before adopting them. The backtrace one I prototyped in this PR wasn't that bad.

PR #5896, opened 2020-09-03
log: more logging cleanup

bmwill:
This includes a number of additional cleanups within our logging framework:
* Removes the old structured logging facade and now we forbid unsafe in the logging crate
* Use a new `LogEntry` type in the backend instead of using `StructuredLogEntry` as a middleman
* Remove the unneeded `into_struct_log` method form generated Schemas
* Reworked some tests
* removed a bunch of, now, unused fields and methods in the `StructuredLogEntry` type


common/logger/src/struct_log.rs
```diff
@@ -118,123 +62,11 @@ impl StructuredLogEntry {
...
-                )
-                .data("OriginalLength", json_string.len());
-
-            // Keep important values around
-            // TODO: We should work on decreasing log sizes, as this isn't sustainable
-            for field in FIELDS_TO_KEEP {
-                if let Some(value) = self.data.get(field) {
-                    entry = entry.json_data(field, value.clone());
-                }
-            }
```
bmwill, line 173: I removed this (and didn't add back in similar logic) since we are no longer automatically converting all format message arguments into structured logs. If we start noticing an issue we can rethink this though

gregnazario: As much as I really want to get rid of this, it's not just that we're automatically converting all format message arguments into structured logs, but sometimes just what we're putting in the log.

http://logging.premainnet.testnet.libra.org/_plugin/kibana/goto/8426e1aab82c15b22e09a41b2a2adf3b
https://github.com/libra/libra/blob/master/state-synchronizer/src/coordinator.rs#L334-L338

I think we probably will need something that at least keeps track of repeated messages that are that large.  We shouldn't have to cut logs, but also we need to protect ourselves from starving out the rest of the logs.  

On the other hand, we do need an example of this log that seems to keep growing in size about bad chunks.  It was 2kb and now it's 4kb.  So, we can remove it for now.

Can you make a followup issue to investigate DoS and starvation protection in the logging system around large logs?  I want to protect ourselves from accidentally shooting ourselves in the foot because of logs.

common/logger/src/libra_logger.rs
```diff
@@ -148,30 +198,20 @@ impl LibraLogger {
     }
 }
 
-impl crate::logger::Logger for LibraLogger {
```
gregnazario, line 155: I guess you imported some here

common/logger/src/libra_logger.rs
```diff
@@ -250,19 +290,14 @@ impl Writer for StderrWriter {
...
-        write!(w, "{} ", level)?;
-    }
-
+    write!(w, "{} ", entry.metadata.level())?;
     write!(w, "{} ", entry.timestamp)?;
 
-    if let Some(location) = &entry.location {
-        write!(w, "{} ", location)?;
-    }
+    write!(w, "{}:{} ", entry.metadata.file(), entry.metadata.line())?;
```
gregnazario, line 300: Any reason you don't write these all in one format string?

Also, I see there some advantages to splitting the file and the line, aka being able to search by file.  However, being able to search by line is less helpful.  I think there is the disadvantage that, here we format it all into the familiar `file:line` format, but we'd have to do extra wrangling to see that information easily in Kibana.

We can still prefix search for the file name in Kibana, so I'm not sure if this split is necessary.

common/logger/src/struct_log.rs
```diff
@@ -1,76 +1,32 @@
...
-const MAX_LOG_LINE_SIZE: usize = 10240; // 10KiB
-const LOG_INFO_OFFSET: usize = 256;
 const WRITE_TIMEOUT_MS: u64 = 2000;
 const CONNECTION_TIMEOUT_MS: u64 = 5000;
 
-// Fields to keep when over size
-static FIELDS_TO_KEEP: &[&str] = &["error"];
-
 #[derive(Debug, Serialize)]
 pub struct StructuredLogEntry {
```
gregnazario, line 20: What's this used for at this point?

common/logger/src/struct_log.rs
davidiw: > I removed this (and didn't add back in similar logic) since we are no longer automatically converting all format message arguments into structured logs. If we start noticing an issue we can rethink this though

Did we remove this? Are regular logs going to structured logging still? I ask because that's the only thing that goes to kibana.

gregnazario (commented):
I think we should consider putting line and number back together into one field :/

Also I noticed all your commits do `logger: message`, but our guidelines say `[logger] message`

common/logger/src/struct_log.rs
bmwill: @gregnazario https://github.com/libra/libra/issues/5901 for that.

So just to be clear: Would you like me to put this back in or are we good for now?

bmwill: @davidiw Yes the series I landed yesterday unifies our logging systems. We no longer have "text" or "structured logs" Its all just logs and they all get sent to Kibana.

common/logger/src/lib.rs
```diff
@@ -151,6 +151,8 @@
 //! 3) `FileStructLog` prints structured logs into a file. This logger separate thread for writing files asynchronously.
 //! 4) `TCPStructLog` sends structured logs to a TCP endpoint. This logger separate thread for sending logs asynchronously.
 
+#![forbid(unsafe_code)]
```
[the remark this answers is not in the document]
bmwill: This only needs to go at the root of a crate, not in each file. We do have this in more of our crates but for some reason we choose to use unsafe when setting up the original structured logging system.

common/logger/src/libra_logger.rs
```diff
@@ -102,11 +102,7 @@ impl LibraLoggerBuilder {
             })
         };
 
-        crate::logger::set_global_logger(logger.clone());
-        let logger = Box::leak(Box::new(logger));
-        if let Err(e) = crate::struct_log::set_struct_logger(logger) {
-            eprintln!("Unable to setup structured logger: {:?}", e);
-        }
+        crate::logger::set_global_logger(logger);
```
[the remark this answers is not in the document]
bmwill: Not particularly. There's definitely some more room for improvement in how things are organized

common/logger/src/struct_log/tests.rs
```diff
@@ -64,21 +60,11 @@ impl crate::logger::Logger for StreamStructLog {
...
-    }
-}
-
-fn set_test_struct_logger() -> Result<Receiver<StructuredLogEntry>, InitLoggerError> {
+fn set_test_struct_logger() -> Receiver<StructuredLogEntry> {
     let (logger, receiver) = StreamStructLog::start_new();
     let logger = Arc::new(logger);
     crate::logger::set_global_logger(logger);
-    Ok(receiver)
+    receiver
```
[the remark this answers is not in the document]
bmwill: I still need to solve that problem. David's PR last night changes from panicking to just ignoring if its called multiple times. I'm planning on addressing that in a future series. 1 problem at a time :)

common/logger/src/struct_log.rs
bmwill: I didn't want to go through the effort yet to remove all uses from the codebase. So right now its used everywhere, where people are creating structured logs without defining their own schemas

common/logger/src/libra_logger.rs
bmwill: The big reason was that the old StrucuturedLogEntry didn't have these fields required so you had  to unwrap them to print them. Now theres not that big of a reason to.

From a structure pov, we should keep them separate, as they are different pieces of data and i don't know if its really that much more difficult to look at a `line` field to find the line number. If you feel strongly enough though, we can add it back

bmwill:
`logger: message` for a commit summary is shorter to type and is the format I use everywhere else so I sometimes type that by way of muscle memory. Easy to fix though

gregnazario (approved):

PR #5899, opened 2020-09-03
[language][e2e tests] make event handle generation deterministic

vgao1996:
## Summary
Event handle generation in e2e tests is nondeterministic, which is not desired for a lot of tests. This makes it use the same deterministic logic used in the Move stdlib. 

In the long run, we should get rid of the rust representation of LibraAccount in e2e tests and instead use Move scripts/function invocations to initialize the accounts.

## Test Plan
Rebase https://github.com/libra/libra/pull/5877 on top of this, and the failing tests should pass.


tzakian (approved):
Good catch!

PR #5903, opened 2020-09-03
[genesis] make validator_owner keys optional

davidiw:
we anticipate one or more validator_owners not directly managing their
validator at launch. Rather than include a bogus key, we intend to make
this explicit in genesis.

### DIFF
--- a/config/src/config/mod.rs
+++ b/config/src/config/mod.rs
@@ -279,29 +279,22 @@ impl NodeConfig {
 
     pub fn random() -> Self {
         let mut rng = StdRng::from_seed([0u8; 32]);
-        Self::random_with_rng(&mut rng)
+        Self::random_with_template(0, &NodeConfig::default(), &mut rng)
     }
 
-    pub fn random_with_rng(rng: &mut StdRng) -> Self {
-        let mut config = NodeConfig::default();
-        config.random_internal(rng);
-        config
-    }
-
-    pub fn random_with_template(template: &Self, rng: &mut StdRng) -> Self {
+    pub fn random_with_template(idx: u32, template: &Self, rng: &mut StdRng) -> Self {
         let mut config = template.clone();
-        config.random_internal(rng);
+        config.random_internal(idx, rng);
         config
     }
 
-    fn random_internal(&mut self, rng: &mut StdRng) {
+    fn random_internal(&mut self, idx: u32, rng: &mut StdRng) {
         let mut test = TestConfig::new_with_temp_dir();
 
         if self.base.role == RoleType::Validator {
             test.random_account_key(rng);
-            let peer_id = libra_types::account_address::from_public_key(
-                &test.owner_key.as_ref().unwrap().public_key(),
-            );
+            let peer_id =
+                crate::utils::validator_owner_account_from_name(idx.to_string().as_bytes());
 
             if self.validator_network.is_none() {
                 let network_config = NetworkConfig::network_with_id(NetworkId::Validator);
--- a/config/src/utils.rs
+++ b/config/src/utils.rs
@@ -4,8 +4,23 @@
 use crate::config::NodeConfig;
 use get_if_addrs::get_if_addrs;
 use libra_network_address::{NetworkAddress, Protocol};
-use libra_types::transaction::Transaction;
+use libra_types::{
+    account_address::AccountAddress,
+    transaction::{authenticator::AuthenticationKey, Transaction},
+};
 use std::net::{TcpListener, TcpStream};
+
+pub fn default_validator_owner_auth_key_from_name(name: &[u8]) -> AuthenticationKey {
+    let salt = "validator_owner::";
+    let mut name_in_bytes = salt.as_bytes().to_vec();
+    name_in_bytes.extend_from_slice(name);
+    let hash = libra_crypto::HashValue::sha3_256_of(&name_in_bytes);
+    AuthenticationKey::new(*hash.as_ref())
+}
+
+pub fn validator_owner_account_from_name(name: &[u8]) -> AccountAddress {
+    default_validator_owner_auth_key_from_name(name).derived_address()
+}
 
 /// Return an ephemeral, available port. On unix systems, the port returned will be in the
 /// TIME_WAIT state ensuring that the OS won't hand out this port for some grace period.
--- a/language/testing-infra/e2e-tests/src/account.rs
+++ b/language/testing-infra/e2e-tests/src/account.rs
@@ -85,6 +85,22 @@ impl Account {
         }
     }
 
+    /// Creates a new account with the given addr and key pair
+    ///
+    /// Like with [`Account::new`], the account returned by this constructor is a purely logical
+    /// entity.
+    pub fn new_validator(
+        addr: AccountAddress,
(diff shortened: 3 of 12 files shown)


davidiw:
Will rebase once the first 3 commits land, but this is good for review!

execution/executor-test-helpers/src/integration_test_impl.rs
```diff
@@ -243,6 +243,10 @@ pub fn test_execution_with_storage_impl() -> Arc<LibraDB> {
         db.reader.get_state_proof(0).unwrap();
     let mut trusted_state =
         TrustedState::from(config.base.waypoint.waypoint_from_config().unwrap());
+    println!(
```
davidiw, line 246: need to delete

language/testing-infra/e2e-tests/src/account.rs
```diff
@@ -85,6 +85,22 @@ impl Account {
...
+        addr: AccountAddress,
+        privkey: Ed25519PrivateKey,
+        pubkey: Ed25519PublicKey,
+    ) -> Self {
+        Account {
+            addr,
+            privkey,
+            pubkey,
+        }
+    }
```
gregnazario, line 102: Don't we assume that the account is derived from the pubkey a couple places?  Because I know the `with_keypair()` does the same as this but derive the account.  Oh but I see you're trying to change it.

language/tools/vm-genesis/src/lib.rs
```diff
@@ -396,16 +401,30 @@ fn create_and_initialize_owners_operators(
     let libra_root_address = account_config::libra_root_address();
 
     // Create accounts for each validator owner
-    for (owner_key, owner_name, _) in operator_assignments {
-        let owner_auth_key = AuthenticationKey::ed25519(&owner_key);
-        let owner_account = account_address::from_public_key(owner_key);
+    for (owner_key, owner_name, _op_assignment) in operator_assignments {
```
JoshLind, line 404: I think restructuring this might make it a little clearer as to why we need to do the hack of rotate_authentication_key() after we've just created the account. From what I understand, we only have to rotate in the case that we don't have a valid auth key, because the create_validator_account() call will fail with the invalid auth key, right? If so, can we make this more obvious? Right now it's not super clear.. I guess the issue is that we need to use a dummy auth key on the creation, first? Perhaps adding a little more duplication will make it clearer where the hack is? e.g.,

```rust
let owner_address = libra_config::utils::validator_owner_account_from_name(owner_name);
if let Some(owner_key) = owner_key {
    let auth_key = AuthenticationKey::ed25519(owner_key).to_vec();
    // Create owner script and call exec()
} else {
    let dummy_auth_key = AuthenticationKey::random();
    // Create owner script and call exec()
    // Rotate using the ZERO_AUTH_KEY
}
```

If you'd prefer to keep your structure, you could just add a comment as to what's going on here. 

Also, does it make sense to use the dummy_auth_key instead of staged_owner_auth_key anyway? This will help to keep the default_validator_owner_auth_key_from_name(..) method private?

language/tools/vm-genesis/src/lib.rs
```diff
@@ -556,18 +575,22 @@ pub fn generate_genesis_type_mapping() -> BTreeMap<Vec<u8>, StructTag> {
 
 /// Generates an artificial set of OperatorAssignments using the given node configurations for
 /// testing.
-pub fn operator_assignments(node_configs: &[NodeConfig]) -> Vec<OperatorRegistration> {
```
JoshLind, line 559: Aah.. annoying that OperatorRegistration and OperatorAssignment had the same types.. otherwise the compiler would have told us that this was wrong 😥

JoshLind (approved):
Thanks, David! Other than the comments, it LGTM 😄

config/src/utils.rs
```diff
@@ -4,9 +4,24 @@
...
+};
 use std::net::{TcpListener, TcpStream};
 
+pub fn default_validator_owner_auth_key_from_name(name: &[u8]) -> AuthenticationKey {
+    let salt = "validator_owner::";
+    let mut name_in_bytes = salt.as_bytes().to_vec();
+    name_in_bytes.extend_from_slice(name);
+    let hash = libra_crypto::HashValue::sha3_256_of(&name_in_bytes);
+    AuthenticationKey::new(*hash.as_ref())
+}
```
gregnazario, line 19: I'm assuming this isn't permanent, but just for genesis?  Because otherwise, I could just derive the same key if I know the name right?

gregnazario (approved):

config/src/config/mod.rs
```diff
@@ -279,29 +279,22 @@ impl NodeConfig {
 
     pub fn random() -> Self {
         let mut rng = StdRng::from_seed([0u8; 32]);
-        Self::random_with_rng(&mut rng)
+        Self::random_with_template(0, &NodeConfig::default(), &mut rng)
```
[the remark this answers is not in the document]
davidiw: they were already the same id, since the rng seed is fixed.

config/src/config/mod.rs
```diff
@@ -279,29 +279,22 @@ impl NodeConfig {
...
+    fn random_internal(&mut self, idx: u32, rng: &mut StdRng) {
         let mut test = TestConfig::new_with_temp_dir();
 
         if self.base.role == RoleType::Validator {
             test.random_account_key(rng);
-            let peer_id = libra_types::account_address::from_public_key(
-                &test.owner_key.as_ref().unwrap().public_key(),
-            );
+            let peer_id =
+                crate::utils::validator_owner_account_from_name(idx.to_string().as_bytes());
```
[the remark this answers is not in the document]
davidiw: move doesn't have a string :/

you cannot have errors going from string -> u8 but you can for u8 -> string, so this seemed the most reliable :/

config/src/utils.rs
davidiw: I'm gonna reach out to AOS, as this wasn't hard, maybe we require all validators and validator operators to have well defined addresses. I had a proposal for this earlier that got bumped due to concerns of complexity...

language/tools/vm-genesis/src/lib.rs
davidiw: that doesn't work.

we only pass in a portion of the auth key and the other portion is the address... so the initial authkey is really useless..

davidiw: I guess I'll try to clear up some of the confusion around account creation in move here, even though this is not the best place.

JoshLind: When you say "that doesn't work", what are you referring to? Given that the initial authkey is useless, at the very least you could generate a random one, right? So that way there's no need to make "default_validator_owner_auth_key_from_name" public and that'll be a little bit cleaner?

Don't feel too pressured to fix the move side. My comment was more about providing context/an explanation to the reader about why the hack exists, rather than fixing the hack

davidiw: yeah, I gave a comment blurb up above now. You do make an interesting point.... but then we need to have something that generates half an auth key... hacks hack everywhere hacks

PR #5904, opened 2020-09-03
[language] use StructTag instead of TypeTag in TransactionEffect

sblackshear:
As a result, TransactionEffect is more strongly typed and easier to work with.


dariorussi (approved):
Looks good! I don't remember but I assume `StatusCode::INTERNAL_TYPE_ERROR` ends up being an invariant violation, right? so not a user error but a system error

sblackshear:
Yep, `StatusCode::INTERNAL_TYPE_ERROR` is an internal invariant violation.

PR #5905, opened 2020-09-04
logging: improve ergonomics of emitting a security event

bmwill:
Improve the ergonomics of emitting a security event by making all events an enum and implementing Schema on the event so that it can easily be logged with other, typed, structured data.

### DIFF
--- a/common/logger/src/security.rs
+++ b/common/logger/src/security.rs
@@ -6,81 +6,80 @@
 //! logs to detect malicious behavior from other validators.
 //!
 //! ```
-//! use libra_logger::prelude::*;
+//! use libra_logger::{error, SecurityEvent};
 //!
 //! error!(
-//!   security_log(security_events::INVALID_RETRIEVED_BLOCK)
-//!     .data("some_data", "the data")
+//!     SecurityEvent::InvalidRetrievedBlock,
+//!     "some_key" = "some data",
 //! );
 //! ```
 //!
 
-use crate::StructuredLogEntry;
+use crate::{Key, Schema, Value, Visitor};
+use serde::{Deserialize, Serialize};
 
-/// helper function to create a security log
-pub fn security_log(name: &'static str) -> StructuredLogEntry {
-    StructuredLogEntry::new_named("security", &name)
-}
-
-/// Security events that are possible
-pub mod security_events {
+#[derive(Clone, Copy, Debug, Serialize, Deserialize)]
+#[serde(rename_all = "snake_case")]
+pub enum SecurityEvent {
+    //
     // Mempool
-    // -------
-
+    //
     /// Mempool received a transaction from another peer with an invalid signature
-    pub const INVALID_TRANSACTION_MP: &str = "InvalidTransactionMP";
+    InvalidTransactionMempool,
 
     /// Mempool received an invalid network event
-    pub const INVALID_NETWORK_EVENT_MP: &str = "INVALID_NETWORK_EVENT_MP";
+    InvalidNetworkEventMempool,
 
     // Consensus
     // ---------
-
     /// Consensus received an invalid message (not well-formed or incorrect signature)
-    pub const CONSENSUS_INVALID_MESSAGE: &str = "ConsensusInvalidMessage";
+    ConsensusInvalidMessage,
 
     /// Consensus received an equivocating vote
-    pub const CONSENSUS_EQUIVOCATING_VOTE: &str = "ConsensusEquivocatingVote";
+    ConsensusEquivocatingVote,
 
     /// Consensus received an invalid proposal
-    pub const INVALID_CONSENSUS_PROPOSAL: &str = "InvalidConsensusProposal";
+    InvalidConsensusProposal,
 
     /// Consensus received an invalid vote
-    pub const INVALID_CONSENSUS_VOTE: &str = "InvalidConsensusVote";
+    InvalidConsensusVote,
 
     /// Consensus received an invalid new round message
-    pub const INVALID_CONSENSUS_ROUND: &str = "InvalidConsensusRound";
+    InvalidConsensusRound,
 
     /// Consensus received an invalid sync info message
-    pub const INVALID_SYNC_INFO_MSG: &str = "InvalidSyncInfoMsg";
+    InvalidSyncInfoMsg,
 
     /// A received block is invalid
-    pub const INVALID_RETRIEVED_BLOCK: &str = "InvalidRetrievedBlock";
+    InvalidRetrievedBlock,
 
     /// A block being committed or executed is invalid
-    pub const INVALID_BLOCK: &str = "InvalidBlock";
+    InvalidBlock,
 
     // State-Sync
     // ----------
-
     /// Invalid chunk of transactions received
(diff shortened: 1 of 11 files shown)


common/logger/src/security.rs
```diff
@@ -6,81 +6,80 @@
...
     // Network
     // -------
-
     /// Network identified an invalid peer
-    pub const INVALID_NETWORK_PEER: &str = "InvalidNetworkPeer";
+    InvalidNetworkPeer,
 
     /// Network couldn't negotiate
-    pub const INVALID_NETWORK_HANDSHAKE_MSG: &str = "InvalidNetworkHandshakeMsg";
+    InvalidNetworkHandshakeMsg,
```
davidiw, line 78: we should be careful what log level we use here, since it could be used to dos our node

mempool/src/shared_mempool/coordinator.rs
```diff
@@ -158,18 +158,20 @@ pub(crate) async fn coordinator<V>(
                                 };
                             }
                             Event::RpcRequest((peer_id, msg, res_tx)) => {
-                                error!(security_log(security_events::INVALID_NETWORK_EVENT_MP)
-                                    .data("message", &msg)
-                                    .data("peer_id", &peer_id)
+                                error!(
```
davidiw, line 161: I think this logging level is probably dangerous, couldn't a public fn send me a bunch of invalid messages causing me to overwhelm my logging infra? Is this interesting?

mempool/src/shared_mempool/coordinator.rs
```diff
@@ -158,18 +158,20 @@ pub(crate) async fn coordinator<V>(
...
                                 );
                                 debug_assert!(false, "Unexpected network event rpc request");
                             }
                         }
                     },
                     Err(e) => {
-                        error!(security_log(security_events::INVALID_NETWORK_EVENT_MP)
-                            .data_display("error", &e)
-                    );
+                        error!(
```
davidiw, line 171: same as above

mempool/src/shared_mempool/tasks.rs
```diff
@@ -270,10 +270,12 @@ pub(crate) async fn process_transaction_broadcast<V>(
         .filter(|(maybe_vm_status, _)| maybe_vm_status.is_some());
 
     for (maybe_vm_status, failed_transaction) in failed_transactions {
-        error!(security_log(security_events::INVALID_TRANSACTION_MP)
-            .data("failed_transaction", &failed_transaction)
-            .data("vm_status", &maybe_vm_status)
-            .data("from_peer", &peer));
+        error!(
```
davidiw, line 273: same as above

network/src/protocols/health_checker/mod.rs
```diff
@@ -197,24 +197,31 @@ where
                             match msg {
                             HealthCheckerMsg::Ping(ping) => self.handle_ping_request(peer_id, ping, res_tx),
                             _ => {
-                                error!(security_log(security_events::INVALID_HEALTHCHECKER_MSG)
-                                    .data("error", "Unexpected rpc message")
-                                    .data("message", &msg)
-                                    .data("peer_id", &peer_id)
+                                error!(
```
davidiw, line 200: same as above

network/src/protocols/health_checker/mod.rs
```diff
@@ -197,24 +197,31 @@ where
...
+                                        .data("peer_id", &peer_id)
                                 );
                             },
                             };
                         }
                         Ok(Event::Message(msg)) => {
-                            error!(security_log(security_events::INVALID_NETWORK_EVENT_HC)
-                                .data("error", "Unexpected network event")
-                                .data("event_message", &msg)
+                            error!(
```
davidiw, line 211: same

network/src/protocols/health_checker/mod.rs
```diff
@@ -304,12 +311,15 @@ where
                             }
                         });
                 } else {
-                    error!(security_log(security_events::INVALID_HEALTHCHECKER_MSG)
-                        .data("error", "Pong nonce doesn't match our challenge Ping nonce")
-                        .data("req_nonce", &req_nonce)
-                        .data("peer_id", &peer_id)
-                        .data("pong", pong.0)
-                        .data("round", round));
+                    error!(
```
davidiw, line 314: same

network/src/transport.rs
```diff
@@ -188,9 +188,12 @@ async fn upgrade_inbound<T: TSocket>(
     // try authenticating via noise handshake
     let (mut socket, remote_peer_id) = ctxt.noise.upgrade_inbound(socket).await.map_err(|err| {
         // security logging
-        warn!(security_log(security_events::INVALID_NETWORK_PEER)
-            .data_display("error", &err)
-            .field(network_events::NETWORK_ADDRESS, &addr));
+        warn!(
```
davidiw, line 191: same as other cases... I think when this is open to the public, we either need an aggressive sampler or counters.

network/src/transport.rs
```diff
@@ -208,10 +211,13 @@ async fn upgrade_inbound<T: TSocket>(
     let (messaging_protocol, application_protocols) = handshake_msg
         .perform_handshake(&remote_handshake)
         .map_err(|err| {
-            warn!(security_log(security_events::INVALID_NETWORK_HANDSHAKE_MSG)
-                .data_display("error", &err)
-                .data("origin", "inbound")
-                .data("remote_peer", remote_peer_id));
+            warn!(
```
davidiw, line 214: same

state-synchronizer/src/coordinator.rs
```diff
@@ -332,10 +332,13 @@ impl<T: ExecutorProxyTrait> SyncCoordinator<T> {
                     .await
                 {
                     // security log
-                    error!(security_log(security_events::STATE_SYNC_INVALID_CHUNK)
-                        .data("from_peer", &peer)
-                        .data_display("error", &err)
-                        .data("chunk", &response));
+                    error!(
```
davidiw, line 335: this is good!

davidiw (approved):
Your component of the PR looks great @bmwill , @mimoo the intent here is great, but I am increasingly concerned that logging could be abused to negatively impact the quality of the network :/

common/logger/src/security.rs
mimoo: noob question: why would a log level facilitate a DoS?

mempool/src/shared_mempool/coordinator.rs
mimoo: yes potentially, I didn't really think of attacks through logs, but we've been talking about the state sync log that can log pretty large messages and it seems to be true for a lot of these. The problem if we don't log the message is that we have no proofs that the other validator is misbehaving. As these are enabled not just on validators we could just enable most of these logs only if we are a validator (but on both the validator network and the full node network). Does that make sense?

network/src/transport.rs
mimoo: or we might as well disable it if this is the public network

mimoo:
cc @AnomalRoil so you can see the changes : o

PR #5906, opened 2020-09-04
key-manager: define and use a typed log schema

bmwill:
This also fixes the issue where all log lines were reported coming from a single location in source by removing and `log` function and using the log macros explicitly.

### DIFF
--- a/secure/key-manager/src/lib.rs
+++ b/secure/key-manager/src/lib.rs
@@ -21,7 +21,7 @@
 
 use crate::{
     libra_interface::LibraInterface,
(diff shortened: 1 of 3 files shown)


davidiw (approved):
Looks good!

JoshLind:
Thanks, @bmwill! I wasn't aware that we were logging source lines. Thanks for fixing that!

bmwill:
/land

Yeah the log macros expand and capture the module, file and line number of the log line. Makes things easier to debug later ;)

PR #5907, opened 2020-09-04
executor: define and use a typed log schema

bmwill:


execution/executor/src/lib.rs
```diff
@@ -186,8 +186,10 @@ where
         let first_txn_version = match txn_list_with_proof.first_transaction_version {
             Some(tx) => tx as Version,
             None => {
-                error!(StructuredLogEntry::new_named("MUST_FIX", "assertion")
-                    .data("details", "first_transaction_version should exist."));
+                error!(
+                    category = "MUST_FIX",
+                    "first_transaction_version should exist"
+                );
                 return Err(anyhow!("first_transaction_version should exist."));
```
davidiw, line 193: I'm kinda surprised this isn't an expect, since we cannot recover, can we? @lightmark

execution/executor/src/logging.rs
```diff
@@ -1,27 +1,53 @@
...
-    StructuredLogEntry::new_named("executor", entry.as_str())
+#[derive(Schema)]
+pub struct LogSchema {
+    name: LogEntry,
+    block_id: Option<HashValue>,
+    root_block_id: Option<HashValue>,
+    original_reconfiguration_block_id: Option<HashValue>,
+    num: Option<u64>,
+    local_synced_version: Option<u64>,
+    first_version_in_request: Option<Option<u64>>,
```
davidiw, line 16: Umm? Why not just Option :)

davidiw (approved):
Looks good!

execution/executor/src/logging.rs
bmwill: Idk that's just the type that was being logged lol!

davidiw: why not just use an option here though?

execution/executor/src/lib.rs
lightmark: It's usually the issue of input so executor shouldn't crash. cc@mimoo

execution/executor/src/logging.rs
lightmark: Option<Option<_>> makes little sense but keeps consistency.

execution/executor/src/lib.rs
davidiw: If it doesn't crash, then we have a useless node... does the error propagate upward to an expect statement?

PR #5909, opened 2020-09-04
[Cluster Test] Some refactors and improvements for the cluster swarm code.

JoshLind:
## Motivation

This PR provides several small refactors and improvements for the cluster swarm code (to help first timers better understand the code). Namely, the PR:

1. Moves all the template files to a "template" directory (to match configs and fluent-bit files)
2. Avoids duplication in service spec and pod spec creation logic.
3. Extracts commonly used strings, file paths and binary paths to constants to make it easier to reason about and identify important components in cluster swarm. 

Note: although the PR offers quite a few commits, you could probably just look at all commits in a single review as the changes are pretty small once flattened.

Also, if you don't feel the refactors are that useful, feel free to close the PR 😄 

Yes.

## Test Plan

All tests seem to pass locally.

## Related PRs

None.


andll:
I have mixed feeling about this one

- I like commits that replace inline strings with constants 👍 
- I don't fully understand motivation for replacing inlined `include_str` with macro like `FULLNODE_CONFIG`. Can you explain benefit behind this?

JoshLind:
Thanks, @andll! 😄

Regarding replacing "include_str(STRING_LITERAL)" with the macro, I was originally hoping to change STRING_LITERAL to be constants, so that we could easily see what template/config files were being loaded at compile time (rather than having to look throughout the file manually.. which is tedious.) Also, it helps prevent bugs, e.g., the JOB_TEMPLATE file is used twice, so it would be nice to have only a single definition of the template file name. Of course, the issue I faced is that we can't use `const` here because we use a STRING LITERAL so it doesn't work with the "include_str" call. 

So it seems like I had two options:
1. Use a macro to help define these as constants equivalents (e.g., a macro name that maps to the file string).
2. Change the "include_str!" macro to something else that loads the file at runtime (rather than compile time). 

I was unsure about taking option 2, as I'm not sure if the assumption can be made that the files are available at runtime. Hence why I took approach 1. Does that make sense? 

If you feel that the macros are overkill, I can refactor the PR to avoid doing this. I also have mixed feelings about having to use a macro for this, so I don't feel super strongly.. 😕

JoshLind:
@andll, any response on this? If not, I'm happy to close the PR (as it's only a refactor). But, in general, I think it leaves the code in a better position than it currently is.

andll:
@JoshLind I think my gut feeling is that macro replacing include str is overkill, something like includ_str!(file_name_macro!()) is probably more obvious

All other refactorings in this PR are very helpful, thank you!

JoshLind:
Thanks @andll 😄 

> I think my gut feeling is that macro replacing include str is overkill, something like includ_str!(file_name_macro!()) is probably more obvious.

Sure, sounds good. I've done this. The only downside is of course that the macro definitions take up a few more lines when formatted.

andll (approved):
Thank you @JoshLind !

JoshLind:
Thanks, @andll 😄 

/land

PR #5910, opened 2020-09-04
[framework] refactored LibraTimestamp, LibraConfig, and LibraSystem

shazqadeer:
## Motivation

This PR simplifies the initialization specification so that is_genesis()/is_operating() are respectively absence/presence of LibraTimestamp::CurrentTimeMicroseconds.  This leads to elimination of some predicates and ripple effects in modules that depend on LibraTimestamp.  

This  PR also does a more thorough job of specifying LibraConfig and LibraSystem.  Several functions for whom verification was turned off are being verified now.

Major code changes:

LibraTimestamp:
- eliminated resource TimeHasStarted and function initialize in LibraTimestamp
- function set_time_has_started takes on the role of initialize and Genesis::initialize calls this function
- deleted function reset_time_has_started_for_test in LibraTimestamp that was introduced only for testing

LibraConfig:
- renamed emit_reconfiguration_event to emit_genesis_reconfiguration_event and make it a standalone function reserved only for  calling from configuration setup
- refactored reconfigure_ so it does not call emit_genesis_reconfiguration_event

LibraSystem:
- renamed get_validators in LibraSystem to get_libra_system_config

Yes

## Test Plan

more specifications (verified by Move Prover) and more tests


wrwg (commented):
I do not understand why we have removed the resource which allows us to distinguish genesis from operating state and instead map it to the timestamp presence. This restricts options how to define genesis in the future, as timestamp initialization must now always happen as the very last step. I think it is much cleaner and general to have a dedicated resource to express this (that is why I introduced this resource in the first place a while ago, after discussion with Sam). We also have removed the option to do functional tests of genesis code with this PR.

I had a long discussion with Tim yesterday about this change. Result is I'm OK with moving forward with it, but next time please coordinate this better with the original author (me!). We also need to followup on bringing the option back to test genesis code, and Tim said he would stay on top of this.

We have higher priorities than doing such refactorings, so I'm not asking to undo anything here, but I leave review to others which have worked with you on this one.

tzakian:
> I do not understand why we have removed the resource which allows us to distinguish genesis from operating state and instead map it to the timestamp presence. This restricts options how to define genesis in the future, as timestamp initialization must now always happen as the very last step. I think it is much cleaner and general to have a dedicated resource to express this (that is why I introduced this resource in the first place a while ago, after discussion with Sam). 

The main reason I was ok with/suggested this change was because I didn't see anywhere where this distinction was being used, and/or where it could be useful. If we decide in the review of this PR that doing this was the wrong direction and the distinction really is useful, I'm fine with us changing things back--and with the upcoming changes around initialization/configuration/operation in genesis, the reason for having this may become clear/important.

> We also have removed the option to do functional tests of genesis code with this PR.
>... We also need to followup on bringing the option back to test genesis code, and Tim said he would stay on top of this.

To address your concerns about losing tests, after our discussion yesterday I added a number of tests to the Rust testing framework for Move that exercise the genesis code paths (and a couple more) whose tests were removed from the functional tests. I will keep track of adding this feature in to the functional tests, but I do think it's worth mentioning that even after this change, we do still have ways to test genesis code starting from the empty state. So while we may have lost some expressive power in the functional tests, we haven't lost any expressive power in terms of the behaviors/paths that we can test with the testing infrastructure for Move.

shazqadeer:
@valerini, @sblackshear, @DavidLDill: Please take a look at this PR.  It has been hanging around for a while.  There are new specs in the PR that I would to be able to land soon.

DavidLDill (approved):

sblackshear (approved):
LGTM

PR #5913, opened 2020-09-04
[language][tools] Move CLI

sblackshear:
This is a tool for experimenting with Move without validators, a blockchain, or transactions. Persistent data is stored on-disk in a directory structure that mimics the Move memory model. See the included [README](https://github.com/libra/libra/blob/791b580154478f3fe10cc25fad33b7e6617f6bfd/language/tools/move-cli/README.md) for a much more comprehensive guide. Running `move help` also gives a reasonable summary of the functionality:

```
Move 0.1.0
CLI frontend for Move compiler and VM

USAGE:
    move-cli [FLAGS] [OPTIONS] <SUBCOMMAND>

FLAGS:
    -h, --help             Prints help information
        --no-use-stdlib    If false, compile `move_data` against the Move stdlib and include the stdlib bytecodes in
                           `move run`
    -V, --version          Prints version information

OPTIONS:
        --data <data>    Directory storing Move resources and module bytecodes produced by script execution [default:
                         move_data]
        --lib <lib>      Directory storing Move source files that will be compiled and loaded into the VM before script
                         execution [default: move_lib]

SUBCOMMANDS:
    clean      Delete all modules and resources stored on disk under `move_data`. Does *not* delete anything in
               `move_lib`
    compile    Compile the modules in `move_data` and the given script (if any)
    help       Prints this message or the help of the given subcommand(s)
    run        Compile/run a Move script that reads/writes resources stored on disk in `move_data`. This command
               compiles each each module stored in `move_lib` and loads it into the VM before running the script
    view       View Move resources and modules stored on disk
```


language/stdlib/src/lib.rs
```diff
@@ -196,7 +259,7 @@ pub fn build_packed_types_map() {
     path.push(PACKED_TYPES_FILENAME);
     path.set_extension(PACKED_TYPES_EXTENSION);
     options.output_path = path.to_str().unwrap().to_string();
-    let mut sources = stdlib_files();
+    let mut sources = stdlib_source_files();
```
tzakian, line 262: Shouldn't this be the source files for the LF and not the standard library?

sblackshear: Good catch!!

language/tools/move-cli/README.md
```diff
@@ -0,0 +1,197 @@
...
+
+The `--signers 0xf` part indicates which account address(es) have signed off on the script. Omitting `--signers` or passing multiple signers to this single-`signer` script will trigger a type error.
+
+TODO: show script compilation error
+
+TODO: showcase basic arithmetic, control-flow constructs, `Signer` module, `Vector` and `Option` modules.
+
+## Adding new modules via `move_lib`
+
+By default, the CLI compiles and includes the `Compare`, `Debug`, `Errors`, `Option`, `Signer`, and `Vector` modules from the Move standard library. New modules can be added to the `move_lib` directory. The `move run` command will compile and publish each module source file in `move_lib` before running the given script.
```
JoelMarcey, line 40: Where does the `move_lib` directory live? I assume anywhere you are creating modules, and not necessarily in the main `libra` repo?

sblackshear: Correct; `move_lib` doesn't have to be under `libra`. By default, the CLI looks in the move_lib` subdirectory of wherever you run the binary, and you can point it to a different directory with the `--lib` flag.

language/tools/move-cli/src/main.rs
```diff
@@ -0,0 +1,400 @@
...
+
+    let vm = MoveVM::new();
+    // TODO: use nonzero schedule and pick reasonable max default gas price
+    let cost_schedule = gas_schedule::zero_cost_schedule();
+    let mut cost_strategy = gas_schedule::CostStrategy::system(&cost_schedule, GasUnits::new(0));
+    let mut signers = signers
+        .iter()
+        .map(|s| AccountAddress::from_hex_literal(&s))
+        .collect::<Result<Vec<AccountAddress>>>()?;
+    signers.reverse(); // TODO: figure out why VM reads this right-to-left
```
sblackshear, line 177: CC @vgao1996 in case the answer to this is known

language/tools/move-cli/src/main.rs
```diff
@@ -0,0 +1,400 @@
...
+                    _ => format!("{} error", status_code.status_type()),
+                };
+            // TODO: map to source code location
+            let location_explanation = match location {
+                AbortLocation::Module(id) => {
+                    format!("{}::{}", id, state.resolve_function(&id, function)?)
+                }
+                AbortLocation::Script => "script".to_string(),
+            };
+            // TODO: code offset is 1-indexed, but disassembler instruction numbering starts at zero
```
sblackshear, line 334: @tzakian: would this be difficult to fix?

language/tools/move-cli/src/main.rs
vgao1996:
```
// VMRuntime::execute_script
            senders.into_iter().for_each(|addr| {
                args.insert(0, Value::transaction_argument_signer_reference(addr))
            });
```
Seems like a bug in the current implementation

tnowacki:
Wanted to quickly jump in and say this looks awesome! And it's something we should have built a long time ago. 
I'm still catching up with some things after being out for a week, but I should have some time tomorrow to look

language/tools/move-cli/src/main.rs
sblackshear: Hmm, wondering how the `AdminScript` tests work given this... @tzakian might know?

tzakian: In the Libra VM when we're populating the signers for the transaction they're populated "from left to right" so
```
senders == vec![execute_as_signer, libra_root_signer]
```

And is going through this and pushing those on at the front, so you'll get:

```
vec![libra_root_signer, execute_as_signer, <rest of args>]
```

which is what we expect (LR as the first signer). 

We could build up the signers a different way if we wanted to, but I'm not sure f that buys us anything?

language/tools/move-cli/src/main.rs
tzakian: I don't think it would be that hard to fix. I think the disassembler instruction numbering is just done using an `enumerate()` so should be pretty simple to change.

sblackshear: Sweet, thanks! I will try adding a +1 in the appropriate place and see what happens...

language/tools/move-cli/src/main.rs
sblackshear: I think the thing that would be nice to buy us is that the caller of `execute_script` can pass `vec![signer1, signer2]` for a script like `fun main(signer1: &signer, signer2: &signer)` instead of `vec![signer2, signer1]`. "Pass the signers in the reverse order that the script expects them" seems like a weird API, particularly given that we ask the for the args and type args in non-reverse order...

I would propose to do this by swapping 
```
senders.into_iter().for_each(|addr| {
  args.insert(0, Value::transaction_argument_signer_reference(addr))
});
``` 

for something like

```
let mut signers_and_args = senders.into_iter().map(|addr| Value::transaction_argument_signer_reference(addr)).collect();
signers_and_args.append(&mut args);
...
```

language/stdlib/src/lib.rs
```diff
@@ -59,7 +59,65 @@ pub fn filter_move_files(dir_iter: impl Iterator<Item = PathBuf>) -> impl Iterat
...
+    dir_iter.flat_map(|path| {
+        if path.extension()?.to_str()? == COMPILED_EXTENSION {
+            Some(path)
+        } else {
+            None
+        }
+    })
+}
+
+fn stdlib_modules() -> Vec<&'static str> {
```
tnowacki, line 74: could we break this out into a separate PR?

language/tools/move-cli/src/lib.rs
```diff
@@ -0,0 +1,293 @@
...
+        }
+        path.push(StructID(tag).to_string());
+        let lcs = resource
+            .simple_serialize(&layout)
+            .ok_or_else(|| anyhow!("Failed to serialize resource"))?;
+        let mut f = fs::File::create(path)?;
+        Ok(f.write_all(&lcs)?)
+    }
+
+    /// Save `module` on disk under the path `module.address()`/`module.name()`
```
tnowacki, line 203: I love how the "directory" structure of modules and resources are now literally directories

language/tools/move-cli/src/main.rs
```diff
@@ -0,0 +1,440 @@
...
+#[derive(StructOpt)]
+#[structopt(name = "Move", about = "CLI frontend for Move compiler and VM")]
+struct Move {
+    /// Directory storing Move resources and module bytecodes produced by script execution.
+    #[structopt(name = "data", long = "data", default_value = "move_data")]
+    move_data: String,
+    /// Directory storing Move source files that will be compiled and loaded into the VM before
+    /// script execution.
+    #[structopt(name = "lib", long = "lib", default_value = "move_lib")]
+    move_lib: String,
```
tnowacki, line 35: we might want to rename this? 

The compiler uses the terminology of source files / dependency files. Maybe keep that

language/tools/move-cli/src/main.rs
```diff
@@ -0,0 +1,440 @@
...
+
+#[derive(StructOpt)]
+#[structopt(name = "Move", about = "CLI frontend for Move compiler and VM")]
+struct Move {
+    /// Directory storing Move resources and module bytecodes produced by script execution.
+    #[structopt(name = "data", long = "data", default_value = "move_data")]
+    move_data: String,
+    /// Directory storing Move source files that will be compiled and loaded into the VM before
+    /// script execution.
+    #[structopt(name = "lib", long = "lib", default_value = "move_lib")]
```
tnowacki, line 34: We should try to keep these args the same across the various tools. Tagging in @bob-wilson since I think he looked at that in the past

language/tools/move-cli/src/main.rs
```diff
@@ -0,0 +1,440 @@
...
+    move_lib: String,
+    /// If false, compile `move_data` against the Move stdlib and include the stdlib bytecodes in
+    /// `move run`.
+    #[structopt(name = "no-use-stdlib", long = "no-use-stdlib")]
+    no_use_stdlib: bool,
+    // Command to be run.
+    #[structopt(subcommand)]
+    cmd: Command,
```
tnowacki, line 42: Didn't really realize you could nest structopt like this, very cool

language/tools/move-cli/src/main.rs
```diff
@@ -0,0 +1,440 @@
...
+/// Create a directory at ./`dir_name` if one does not already exist
+fn maybe_create_dir(dir_name: &str) -> Result<&Path> {
+    let dir = Path::new(dir_name);
+    if !dir.exists() {
+        fs::create_dir(dir)?
+    }
+    Ok(dir)
+}
+
+/// Compile the user modules in `move_lib_directory` and the script in `script_file`
```
tnowacki, line 103: It seems like this is an all at once thing for modules? Am I understanding that correctly?

I have some ideas on how to improve this. And talking to Victor earlier about testing, I got some ideas of what to do to improve both of these use cases (hopefully) 

But that discussion is not needed for this PR

language/tools/move-cli/src/main.rs
tnowacki: I don't think the Libra VM should be doing signers *at all* which might be part of the problem

language/tools/move-cli/src/main.rs
```diff
@@ -0,0 +1,440 @@
...
+        println!("Committed changes.")
+    } else if !effects_opt.map_or(false, |effects| effects.resources.is_empty()) {
+        println!("Discarding changes; re-run with --commit if you would like to keep them.")
+    }
+
+    Ok(())
+}
+
+/// Explain an execution error
+fn explain_error(error: VMError, state: &OnDiskStateView) -> Result<()> {
```
tnowacki, line 311: woah....

Some neat stuff. We should probably move this to VMStatus maybe? (bit tricky with the OnDiskStateView as is now) but might be useful for other tools

language/stdlib/src/lib.rs
sblackshear: Good call; I'll just use the whole Libra Framework as the default `move_lib` for now (which might make sense anyway)

language/tools/move-cli/src/main.rs
sblackshear: `move_src` then?

language/tools/move-cli/src/main.rs
sblackshear: Yes, that's correct. It's super naive today--it compiles and publishes everything in `move_lib` even if the bytecodes have already been published to `move_data`.

language/tools/move-cli/src/main.rs
sblackshear: Agreed that putting this somewhere more generally usable would be good! Not sure about `VMStatus` for the reason you mention + that it would be nice to show source context as part of this too. 

FWIW most of this is building on @tzakian's awesome work with the error map.

sblackshear:
> I don't think the Libra VM should be doing signers at all which might be part of the problem

Agreed, I think accepting a list of addresses and then building signers in the Move VM probably makes sense...

language/tools/move-cli/src/main.rs
tzakian: Far point! I think the non-reverse order argument makes a good case for why this should be changed. I would be more than happy to see the change go in, so feel free to change in this PR (it shouldn't be a large change) or I can put up a separate PR if that's preferable.

sblackshear: I am too chicken to change it in this PR, but am happy to put up a separate one.

language/tools/move-cli/src/main.rs
tnowacki: Unsure, but I think we should try to match the source language compiler/prover as much as possible

tnowacki (approved):
I don't see any reason to stop this from going in now. We can iterate on it once it's in. Very excited to see this!

tnowacki:
Very small note. I'm not sure how much I like the general name "CLI" as this is mostly a local mock environment tool. But maybe overtime we could make it into a one stop tool for all sorts of move things, not really sure

sblackshear:
My hope is indeed to go for the one-stop shop approach. This is already partially subsuming the CLI frontends for  the compiler, `resource-viewer`, `disassembler`, and `explain-error`, and could go further in this (or not, depending on what makes sense...).

PR #5914, opened 2020-09-04
logger: improve documentation and output structured data to stderr

bmwill:
This will fix the issue where if no message is included in a log, nothing interesting will be output.

For now we're just outputting the structured data formatted as Json but in the future we may want to better format it (ie unquoting key names, etc)


davidiw (approved):

common/logger/src/lib.rs
```diff
@@ -1,80 +1,40 @@
...
+//! info!(key1 = value1);
 //!
-//! info!(StructuredLogEntry::new_named("TransactionEvents", "Committed")
-//!    .data("block", &block)
-//!    .data_display("author", &author)
-//!    .field(&logging_field, &string)
-//!    .message(format!("Committed block: {:?} by {}", block, author))
+//! // You can even set multiple key/value pairs and a format message
+//! let value2 = false;
+//! info!(key1 = value1, key2 = value2, "hello {}", world);
```
gregnazario, line 27: I would mention that anything after the format message will not be caught by the structured log key pairs.

common/logger/src/lib.rs
```diff
@@ -87,69 +47,22 @@
 //!     #[schema(display)]
 //!     c: Option<&'a str>,
 //! }
```
gregnazario, line 49: Maybe mention debug as well.

gregnazario (approved):

common/logger/src/lib.rs
bmwill: It is mentioned, just github didn't do a good job showing it :)

PR #5915, opened 2020-09-04
[network] Moved ConnectivityManager to structured logging

gregnazario:
### Overview
Using the new structured logging schema, built up some schemas to
replace all logs in ConnectivityManager.  There were a couple logs added, and reworded.  In addition, some dead code was removed from the logging file.

Tracing logs in the main execution loop were also moved so it was easier to read :)

### DIFF
--- a/network/src/connectivity_manager/mod.rs
+++ b/network/src/connectivity_manager/mod.rs
@@ -28,7 +28,7 @@
 //! using a relay protocol.
 
 use crate::{
-    logging::{network_events::DISCOVERY_SOURCE, *},
+    logging::NetworkSchema,
     peer_manager::{self, conn_notifs_channel, ConnectionRequestSender, PeerManagerError},
 };
 use futures::{
@@ -52,9 +52,9 @@ use std::{
     collections::{HashMap, HashSet},
     fmt, mem,
     sync::{Arc, RwLock},
-    time::{Duration, Instant},
+    time::Duration,
 };
-use tokio::time;
+use tokio::{time, time::Instant};
 use tokio_retry::strategy::jitter;
 
 pub mod builder;
@@ -209,7 +209,10 @@ where
             eligible
         );
 
-        info!("{} connectivity manager init", network_context);
+        info!(
+            NetworkSchema::new(&network_context),
+            "{} Initialized connectivity manager", network_context
+        );
 
         let mut connmgr = Self {
             network_context,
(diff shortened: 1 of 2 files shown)


network/src/logging.rs
```diff
@@ -10,37 +10,60 @@
...
+
+#[derive(Schema)]
+pub struct RemotePeerWithAddressSchema<'a> {
+    pub network_context: &'a NetworkContext,
+    #[schema(display)]
+    pub remote_peer: &'a PeerId,
+    #[schema(display)]
+    pub network_address: &'a NetworkAddress,
+}
+
```
bmwill, line 54: This is probably personal preference, but it might make things a little easier to maintain if you have a single Schema for the whole network, defining all of the field names/types you want and use the generated setters like is done in the Saftey-rules crate. E.g:

```
struct LogSchema<'a> {
    network_context: &'a NetworkContext,
    remote_peer: Option<&'a PeerId>,
   network_address: Option<&'a NetworkAddress>,
   . . .
}
```

network/src/connectivity_manager/mod.rs
```diff
@@ -52,9 +52,9 @@ use std::{
     collections::{HashMap, HashSet},
     fmt, mem,
     sync::{Arc, RwLock},
-    time::{Duration, Instant},
+    time::Duration,
 };
-use tokio::time;
+use tokio::{time, time::Instant};
```
bmwill, line 57: Is there a difference between these Instant's?

network/src/connectivity_manager/mod.rs
```diff
@@ -209,7 +209,10 @@ where
             eligible
         );
 
-        info!("{} connectivity manager init", network_context);
+        info!(
+            NetworkSchema::new(&network_context),
+            "{} Initialized connectivity manager", network_context
```
bmwill, line 214: You probably don't need to also include the network context in the formatting message, unless it adds value to the message itself, since the context will be included as structured data.

network/src/logging.rs
```diff
@@ -10,37 +10,64 @@
...
+#[derive(Schema)]
+pub struct NetworkSchema<'a> {
+    #[schema(display)]
+    pub discovery_source: Option<&'a DiscoverySource>,
+    pub error: Option<&'a str>,
+    #[schema(display)]
+    pub network_address: Option<&'a NetworkAddress>,
+    pub network_context: &'a NetworkContext,
+    #[schema(display)]
+    pub remote_peer: Option<&'a PeerId>,
```
bmwill, line 44: Nit: These fields don't really need to be `pub`.
Nit: Assuming there's just 1 type of error serialized in networking code you could take a reference to that explicit error type instead of to a `str`.

network/src/connectivity_manager/mod.rs
```diff
@@ -343,7 +341,14 @@ where
...
             .cloned()
             .collect();
+
         for p in stale_dials.into_iter() {
+            debug!(
+                NetworkSchema::new(&self.network_context).remote_peer(&p),
+                "{} Cancelling stale dial {}",
+                self.network_context,
+                p.short_str()
+            );
```
bmwill, line 351: I don't have context on what "network_context" display's as, but this is a good instance where it's valuable to provide a fmt message with a short peer_id, and include the full peer id as structured data.

bmwill (approved):
This looks good! It'll be nice to finish converting out the rest of networking to use the logging schema.

Just a couple nits, the main one is really a question for clarification: are you intentionally including the "network context" data in every formatted message as well as structured data?

network/src/connectivity_manager/mod.rs
gregnazario: Looking at the documentation, not really.  But, we only use this for logging time until the delay, so I figured I'd clean it up rather than convert types.

network/src/connectivity_manager/mod.rs
gregnazario: I had this before you had added the structured data.  It can be removed.  It's mostly for readability of logs that would show up like:

```
[peer_id, role, network_id] Init connectiv manager
[peer_id, role, network_id] DId A
[peer_id, role, network_id] did B
[peer_id, role, network_id] did C
```

network/src/logging.rs
gregnazario: Yeah, this was before I was using the setters.  I'm changing it back to a String.  There is not 1 type of error serialized in the networking code, so I just made it a String.

network/src/connectivity_manager/mod.rs
gregnazario: network context displays as `[network_id, role, peer_id]`.

network/src/connectivity_manager/mod.rs
bmwill: Ah gotcha, I can see that being useful either way so I'll leave that up to your discretion

network/src/logging.rs
```diff
@@ -8,39 +8,69 @@
...
+//! use libra_network_address::NetworkAddress;
+//! use libra_types::PeerId;
 //!
 //! info!(
-//!   network_log(network_events::CONNECTIVITY_MANAGER_LOOP, &NetworkContext::mock())
-//!     .data(network_events::TYPE, network_events::START)
-//!     .field(network_events::EVENT_ID, &5)
+//!   NetworkSchema::new(&NetworkContext::mock())
+//!     .remote_peer(PeerId::random())
+//!     .network_address(&Some(NetworkAddress::mock())),
```
bmwill, line 19:
```suggestion
//!     .remote_peer(&PeerId::random())
//!     .network_address(&NetworkAddress::mock()),
```

Although you might want to double check that you won't end up with those being dropped too early

gregnazario:
> 💔 Test Failed - [commit-workflow](https://circleci.com/workflow-run/cef5491f-4891-42cc-a9f4-26f34a78b7c3?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-checks-link)

This failed for some docker linting issue, so I've rebased and trying again

rexhoffman:
Canceled the /land as I need to get a fix for lint-docker in for this to (maybe) based on the history land.

rexhoffman:
/land

and back in the queue

Issue #5916, opened 2020-09-04
[Bug] genesis-tools exits 0 on errors


mgorven:
# 🐛 Bug

## To reproduce

    $ target/debug/libra-genesis-tool owner-key --validator-backend 'backend=vault;server=https://localhost:8201;ca_certificate=/home/mgorven/libra-partners/validator/gcp/default-vault.ca;token=vault.gcp;namespace=libra' --shared-backend 'backend=disk;path=/tmp/tmp.vVm7yMg31J;namespace=gcp'; echo $?
    Operation unsuccessful: Failed to read 'owner' from validator storage: Permission denied
    0

## Expected Behavior

Tool should exit with non-zero exit code on errors.

## System information

- 0102f6f89

gregnazario:
I think that's fair, all the exit codes are the same right now

davidiw:
wow, we addressed this quickly... genesis has a pr landing to address this...

PR #5918, opened 2020-09-04
[management] Add exit codes for operational tool

gregnazario:
### Overview 
If the tool has an error, the return code will now be -1.  We can
optimize later for having return codes more specific to errors.

Take a look at this, and then we can apply to the genesis tool.  Related to #5916 

### Testing

```
$  cargo run -p libra-operational-tool validator-set --config  ~/other/libra/gen-config.yaml
{
  "Error": "Failed to read 'account-state' from JSON-RPC: JSON RPC call failed with response: Connection Failed: connection timed out\n"
}
$ echo $?
1

```

### DIFF
--- a/config/management/operational/src/command.rs
+++ b/config/management/operational/src/command.rs
@@ -112,7 +112,7 @@ impl std::fmt::Display for CommandName {
 }
 
 impl Command {
-    pub fn execute(self) -> String {
+    pub fn execute(self) -> Result<String, Error> {
         match self {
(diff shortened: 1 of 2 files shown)


mgorven:
Can we rather use code 1? Negative codes are usually used to indicate signal kills.

gregnazario:
> Can we rather use code 1? Negative codes are usually used to indicate signal kills.

Done.

mgorven (approved):

PR #5921, opened 2020-09-04
[libra framework] fix comment for rotate_authentication_key_with_recovery_address script

bob-wilson:
The account whose key is rotated is specified by the to_recover argument, which might be different than the sender account.

Yes

## Test Plan

N/A (only changes a comment)


sblackshear (approved):
Good catch, thanks Bob!

PR #5922, opened 2020-09-05
[libra framework][cleanup] Remove VASPOperationsResource as it's no longer used

tzakian:
This removes the `VASPOperationsResource` that was being used to create account limits when a currency was added for a VASP. However, that's no longer needed, so this resource was is no longer being used for anything.

This also means that the `VASP` module no longer has any initialization to it! 

This PR will change genesis (since the `VASPOperationsResource` will not be published, but it is not a breaking change since there are no consumers of this resource.

This also removes an old `RegisterNewCurrency` resource in the Libra module since that was no longer being used or created anywhere.


sblackshear (approved):
Good spot!

PR #5924, opened 2020-09-05
[Smoke Tests] Create smoke-tests crate under testsuite and partially refactor the layout.

JoshLind:
## Motivation

This PR offers a small set of refactors for the smoke tests. The PR doesn't offer anything magical (it's still a big mess), but it takes some starting steps. In the commits, we:

1. Create a dedicated smoke-tests crate inside testsuite. It doesn't seem to make much sense having 'tests/libratest/smoke_tests.rs' inside the testsuite folder, when it is shared by many things that are orthogonal to it (e.g., cluster test and fuzzing?)
2. Move the actual smoke tests into src and get rid of the empty tests/main.rs file and src/lib.rs files. This brings us one step closer to being able to split out tests from test environment/framework logic.
3. Clean up the Cargo.toml file and unused dependencies.
4. Update the smoke tests to reference the new dev-module folder.
5. Update the Circle CI script to invoke the smoke tests using the new package name.

There's still a lot that should be done. But one PR at a time.. 😅 

Yes.

## Test Plan

Running all tests locally seems to work! I guess we'll see how CI does when it comments on this PR! 😄 

## Related PRs

None.


bmwill:
I haven't looked at the contents of this PR, just the cover letter, but there's good reasons why the smokes tests were under a tests directory and not under src. The crate is only meant for testing and shouldn't have anything that another crate can depend upon. As for why the test suite directory contains a bunch of other stuff, I'm not exactly sure when or why that started happening and I agree that should be cleaned up.

davidiw:
> As for why the test suite directory contains a bunch of other stuff, I'm not exactly sure when or why that started happening and I agree that should be cleaned up.

Refactors really need to get buy-in via coding guidelines.... I agree...

On this PR, it might be helpful to explain a medium to long term perspective here. I suspect your goal is to make these tests more flexible so that you have test definitions as a library?

JoshLind:
Thanks @bmwill 😄 

> Could you share more insight into point 2? I don't quite understand the reason why you want it to be removed from our testing framework.

Overall, the point of making this change was so that we could start to separate test infrastructure from the tests themselves. Granted, this doesn't necessarily mean that things should all be moved across out of `tests`, but my thinking was that it was nicer than having to maintain empty files in the mean time (e.g., tests/main.rs, and src/lib.rs)  . Moreover, given that the code is all compiled using #[cfg(tests)], and listed in dev-only crate dependencies (similar to cluster test), this shouldn't affect production dependencies. I guess the only thing that changes here is that these tests no longer run under "integration tests" in Rust right now, but I'm not sure that's much of a big deal given how these tests are actually run... I don't feel super strongly here though.. so if you'd prefer this stuff all live in a dedicated `tests` directory, I can update the PR to maintain that structure. Thoughts?

Thanks @davidiw,

> On this PR, it might be helpful to explain a medium to long term perspective here. I suspect your goal is to make these tests more flexible so that you have test definitions as a library?

Exactly. I'm just moving them across for now so it'll be less hassle to separate out testing infrastructure from tests themselves. But I guess there's a bunch of different ways in which to make this happen.

JoshLind:
Any thoughts here folks? @bmwill @davidiw ?

davidiw (approved):
I'm in favor of at least making it clear the structural relationship which has been super confusing since testsuite is now a collection of things rather than just smoke tests

JoshLind:
Thanks @davidiw! That makes sense -- we can improve the clarity going forward.

/land

PR #5927, opened 2020-09-07
[move-prover] Copy Propagation and Transformation Pipeline.

wrwg:
This PR refactors the copy propagation algorithm in `reaching_defs_analysis.rs` and applies it to all bytecode. Previously references where skipped for copy propagation because of some issues which are now fixed.

The PR also removes the use of the `$tmp` variable in the generated bytecode, which is obsolete since the new memory model.

Overall goal was to make code more readable in preparation for understanding and changing the pack/unpack semantics. As a side effect, code may have become some 20% smaller, as well speed seem to have increased by 10-15%.Unfortunately, the butterfly effect required to turn verification off for `staple_lbr`.

*Changes in the algorithm*: In the new version, elimination of dead assignments is left to the `livevar_analysis.rs`, which did already some of this work. The new version also uses the reaching definitions at each program point instead of the exit points; the later would be only correct if we had single-assignment semantics.

*Changes in pipepline tools*: For debugging, a few changes where made: (a) there is now a flag `--dump-bytecode` which causes the prover to dump the intermediate bytecode for each step in the pipeline (b) the format of the bytecode output has been (attempted) to improve (c) some assertions have been added to borrow analysis for debugging. Moreover, the `pipeline.run()` algorithm has been refined that it actually again works breadth first, so processors of functions can access the result of the most recent step in the pipeline for other functions (not used because we do not yet do inter-function analysis and transformation, but might be handy later).

## Motivation

Towards refining pack/unpack semantics (sum of coins etc.)

Yes

## Test Plan

Existing and new tests.

## Related PRs

NA

### DIFF
--- a/language/move-prover/spec-lang/src/env.rs
+++ b/language/move-prover/spec-lang/src/env.rs
@@ -2256,14 +2256,27 @@ impl<'env> FunctionEnv<'env> {
             .get_function_source_map(self.data.def_idx)
         {
             if let Some((ident, _)) = fmap.get_parameter_or_local_name(idx as u64) {
-                // The Move compiler produces temporay names of the form `<foo>%#<num>`.
-                // Ignore those names and use the idx-based repr instead.
-                if !ident.contains("%#") {
-                    return self.module_env.env.symbol_pool.make(ident.as_str());
-                }
+                // The Move compiler produces temporary names of the form `<foo>%#<num>`,
+                // where <num> seems to be generated non-deterministically.
+                // Substitute this by a deterministic name which the backend accepts.
+                let clean_ident = if ident.contains("%#") {
+                    format!("tmp#${}", idx)
+                } else {
+                    ident
+                };
+                return self.module_env.env.symbol_pool.make(clean_ident.as_str());
             }
         }
         self.module_env.env.symbol_pool.make(&format!("$t{}", idx))
+    }
+
+    /// Returns true if the index is for a temporary, not user declared local.
+    pub fn is_temporary(&self, idx: usize) -> bool {
+        if idx >= self.get_local_count() {
+            return true;
+        }
+        let name = self.get_local_name(idx);
+        self.symbol_pool().string(name).contains("tmp#$")
     }
 
     /// Gets the number of proper locals of this function. Those are locals which are declared
--- a/language/move-prover/stackless-bytecode-generator/src/livevar_analysis.rs
+++ b/language/move-prover/stackless-bytecode-generator/src/livevar_analysis.rs
@@ -1,6 +1,9 @@
 // Copyright (c) The Libra Core Contributors
 // SPDX-License-Identifier: Apache-2.0
 
+// Live variable analysis with subsequent dead assignment elimination and
+// computation of new Destroy instructions.
+
 use crate::{
     dataflow_analysis::{
         AbstractDomain, DataflowAnalysis, JoinResult, StateMap, TransferFunctions,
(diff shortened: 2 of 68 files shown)


language/move-prover/spec-lang/src/env.rs
```diff
@@ -2256,16 +2256,29 @@ impl<'env> FunctionEnv<'env> {
             .get_function_source_map(self.data.def_idx)
         {
             if let Some((ident, _)) = fmap.get_parameter_or_local_name(idx as u64) {
-                // The Move compiler produces temporay names of the form `<foo>%#<num>`.
-                // Ignore those names and use the idx-based repr instead.
-                if !ident.contains("%#") {
-                    return self.module_env.env.symbol_pool.make(ident.as_str());
-                }
+                // The Move compiler produces temporary names of the form `<foo>%#<num>`,
+                // where <num> seems to be generated non-deterministically.
```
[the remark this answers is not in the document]
tnowacki: I was just fairly lazy in generating the new temps as it didn't matter at the time. 

I do map them back to the original name for the spec exp blocks. Do we need to do the same here? or does it not matter too much?

language/move-prover/stackless-bytecode-generator/src/livevar_analysis.rs
```diff
@@ -83,9 +108,66 @@ impl LiveVarAnalysisProcessor {
             &code,
             &cfg,
         );
-        let mut annotations = analyzer.post_process(&cfg, &code, state_map);
-        let new_bytecode = analyzer.possibly_transform_code(&mut annotations, code);
-        (new_bytecode, annotations)
+        analyzer.post_process(&cfg, &code, state_map)
+    }
+
+    fn eliminate_unused_vars(
```
shazqadeer, line 114: I am unable to understand what this function is doing.  What does it have to do with removing redundant assignment?

wrwg: Well, you have a lot of unused variables after the dead code elimination, and those are simply eliminated here. Look at the baselines to see the effect.

language/move-prover/spec-lang/src/env.rs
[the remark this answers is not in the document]
wrwg: We only need to remap those because we need deterministic output for baseline files. Baseline files can contain references to those names for negative prover tests.

language/move-prover/stackless-bytecode-generator/src/livevar_analysis.rs
shazqadeer: I am not sure which baseline you are referring to.  There are diffs in tons of files.  Could you point me to a specific test output file whose diff is particularly revealing?

language/move-prover/spec-lang/src/env.rs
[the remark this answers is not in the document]
tnowacki: The more I think about it, I'm surprised this isn't deterministic already. Do you have repro I can look at?

language/move-prover/stackless-bytecode-generator/src/livevar_analysis.rs
wrwg: Look at the livevars baselines. These are the tests for the code you are commenting on. Also, isn't it clear that after dead code elimination there will be a lot of unused locals and it makes sense to eliminate them?

language/move-prover/spec-lang/src/env.rs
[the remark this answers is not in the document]
wrwg: The best I can imagine is in a clean client, uncomment this special treatment here and then run the prover tests multiple times. You'll see over time baseline failures because the number here as changed.

In any case, this is not related to this PR (this treatment was there since long, I only changed a bit of the logic unrelated to what we are discussing).

language/move-prover/stackless-bytecode-generator/src/livevar_analysis.rs
shazqadeer: When I was reading the implementation of eliminate_unused_vars last night, I couldn't understand it.  But today I do.

shazqadeer (approved):

wrwg:
@shazqadeer FYI, I did remove the propagation of the aliases for the destination of assign as discussed offline, and all tests pass unmodified as before. So it was truly redundant.

shazqadeer:
I discovered some other redundant code last night in writeback analysis.  I will remove that separately.

PR #5928, opened 2020-09-08
[move-prover] Avoiding unnecessary writebacks.

wrwg:
This PR extends the borrow analysis to track which references actually have been written to and therefore need to be written back when the reference dies.

Before ths PR, if we have the following Move code:

```
fun f(r: &mut T) {
    let x = r.x;
    *x = 1;
}
```

The generated bytecode would be something like:
```
   x := borrow_field(r, x)
   LocalRoot(r) <- .. // r dies, but unnecessary write back because r has not been written to
   WriteRef(x, 1);.
   LocalRoot(r) <- ..
```

This PR removes this redundancy, which appeared to be quite common in the code generated from the framework.

There seems to be a small performance benefit coming out of this PR, but I have not systematically measured it. More importantly perhaps, the generated code is cleaner, and it is step for me to understand borrow and write back analysis better.

*Note*: This PR is based on unmerged #5927 which is separately reviewed; please focus on the 2nd of the two commits.

## Motivation

Towards refining pack/unpack.

Yes

## Test Plan

Existing tests

## Related PRs

#5927

### DIFF
--- a/language/move-prover/stackless-bytecode-generator/src/borrow_analysis.rs
+++ b/language/move-prover/stackless-bytecode-generator/src/borrow_analysis.rs
@@ -30,6 +30,7 @@ pub struct BorrowInfo {
     pub live_refs: BTreeSet<TempIndex>,
     pub borrowed_by: BTreeMap<BorrowNode, BTreeSet<BorrowNode>>,
     pub borrows_from: BTreeMap<BorrowNode, BTreeSet<BorrowNode>>,
+    pub dirty_nodes: BTreeSet<BorrowNode>,
 }
 
 impl BorrowInfo {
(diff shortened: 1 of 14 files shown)


language/move-prover/stackless-bytecode-generator/src/borrow_analysis.rs
```diff
@@ -561,6 +590,34 @@ impl<'a> BorrowAnalysis<'a> {
         Ok(post)
     }
 
+    fn mark_dirty(&self, state: &mut BorrowState, ref_id: RefID) {
+        if let Some(node) = self.ref_id_to_borrow_node.get(&ref_id).cloned() {
+            // We could stop if the node is already in the set, but the graph can change as we
```
shazqadeer, line 595: I did not understand the comment.  What is the activity that can cause the "graph can change"?  If graph is referring to the borrow graph, then this graph does not appear to be changing while mark_dirty is executing.  I suspect you might be referring to changes to the graph during the dataflow analysis.

You are doing a DFS on an acyclic graph without maintaining a visited set.  Although not needed for termination, a visited  set might be more efficient.

shazqadeer (approved):

language/move-prover/stackless-bytecode-generator/src/borrow_analysis.rs
wrwg: I extended the comment to make it clearer. The "activity" are multiple assignments to the same variable. I don't think it is worth the effort to add a visited set, because I expect borrow graphs have the size of just a few nodes.

PR #5931, opened 2020-09-08
Fix dep verifier

metajack:
This should fix the spurious issues from the dependency verifier. It turned out that GitHub's checkout action doesn't checkout the PR but a new ref that is the PR merged into master. This broke the assumptions of the verifier and caused much havoc.


bmwill (approved):
The justification for why the default is merge commit is just plain wrong...

thanks for the fix!

PR #5933, opened 2020-09-08
[faucet] new faucet service: replicates existing faucet behavior

xli:
New faucet service, put it under client as it is one kind of full node client application.
This diff is focusing on replicating current faucet service feature, improvement will be added later.

see https://github.com/libra/libra/issues/5943 for more details.

## Test Plan

* manual test with libra-swarm
* unit test.


bmwill:
Old faucet was written in python right? Seeing this PR literally made my day. Out of curiosity where does the old faucet code live?

client/faucet/TODO
```diff
@@ -0,0 +1,4 @@
+* better error message when request params is invalid
```
bmwill, line 1: Might be better to have these as Github issues rather than in a TODO file in the codebase

client/faucet/src/main.rs
```diff
@@ -0,0 +1,339 @@
...
+        Ok(body) => Ok(Box::new(body)),
+        Err(err) => Err(warp::reject::custom(ServerInternalError(err.to_string()))),
+    }
+}
+
+struct OptFmt<T>(Option<T>);
+
+impl<T: fmt::Display> fmt::Display for OptFmt<T> {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        if let Some(ref t) = self.0 {
```
bmwill, line 107: if you don't want to write `Some(ref t) = self.0` you can also write `Some(t) = &self.0`

client/faucet/src/main.rs
```diff
@@ -0,0 +1,339 @@
...
+                    )
+                    .as_str(),
+                )
+                .reply(&filter)
+                .await
+        });
+        assert_eq!(resp.body(), "2389"); // 2388+1
+    }
+
+    #[test]
```
bmwill, line 198: You can use the `#[tokio::test]` attribute to make test functions "async" so that you don't have to create a runtime for each test.

bmwill (commented):
Just some cursory comments, I'll let someone more familiar with faucet to look at the actual functionality.

client/faucet/src/main.rs
```diff
@@ -0,0 +1,339 @@
...
+
+    info!("[faucet]: running on: {}", address);
+    warp::serve(routes(service)).run(address).await;
+}
+
+fn routes(
+    service: std::sync::Arc<mint::Service>,
+) -> warp::filters::BoxedFilter<(impl warp::Reply,)> {
+    // POST /mint?amount=25&auth_key=xxx&currency_code=XXX
+    let mint = warp::path("mint")
```
mgorven, line 69: Can you handle minting at `/` as well, old clients expect that path.

client/faucet/src/main.rs
```diff
@@ -0,0 +1,339 @@
...
+        args.server_url.as_str(),
+    );
+    let service = std::sync::Arc::new(mint::Service::new(
+        args.server_url,
+        args.chain_id,
+        args.mint_key_file_path,
+    ));
+
+    info!("[faucet]: running on: {}", address);
+    warp::serve(routes(service)).run(address).await;
```
mgorven, line 62: Are mint requests serialised? Trying to mint in parallel is going to cause problems with sequence numbers AFAIK.

xli: This version tries to keep it simple, parallel call will have one success and another one fail.
What I plan to do is later to add an option to return submitted transaction bytes, so that client can wait and know result of the transaction, and handle retry.

Handling parallel minting will need manage sequence number in memory (assuming we don't want to add db for it), and recover the seq number when in memory one out of sync (if a transaction lost, the seq number may not be used as expected).
It is very complex if we wanted to handle it correctly in async world.

Another way to handle is using blocking thread, every request need to be in a queue, after submitted transaction, we need to wait for transaction executed, and we may need to wait 30 seconds (the transaction expiration duration), long blocking wait is not what we wanted too.

Hence I choose this fire and forget version, let client retry.

xli:
> Old faucet was written in python right? Seeing this PR literally made my day. Out of curiosity where does the old faucet code live?

it's docker/mint/server.py

client/faucet/src/main.rs
xli: you are right, I forgot this old path.

client/faucet/src/main.rs
mgorven: Considering that clients currently have to retry because mint requests are failing, is this going to be an improvement for the user experience? The main reason we created the faucet service in the first place was because of the serialisation requirement. (Personally I don't really care BTW, I just want people to stop complaining about faucet :-) )

client/faucet/src/main.rs
xli: thanks for the hint, look better now

client/faucet/TODO
xli: good idea

client/faucet/src/main.rs
xli: I added a todo here: https://github.com/libra/libra/issues/5943
Will give a try later.

To be clear, it's not that I don't think it is not necessary, it is just too hard to do so. 
I need find a robust 
Handling in memory sequence number  which may end with dead locking Faucet service or using invalid sequence number, and end with Faucet service out of service.

About recent problem, it's not retry causes issue, it is Faucet service does not work with retry.

client/faucet/src/mint.rs
```diff
@@ -0,0 +1,150 @@
...
+impl MintParams {
+    fn create_parent_vasp_account_script(
+        &self,
+        seq: u64,
+    ) -> libra_types::transaction::TransactionPayload {
+        let receiver = self.auth_key.derived_address();
+        libra_types::transaction::TransactionPayload::Script(
+            transaction_builder_generated::stdlib::encode_create_parent_vasp_account_script(
+                type_tag_for_currency_code(self.currency_code.to_owned()),
+                0, // sliding nonce
```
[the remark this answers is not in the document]
xli: don't know, this is from existing cli mint coins logic, didn't find document about sliding nonce

client/faucet/src/mint.rs
```diff
@@ -0,0 +1,150 @@
...
+        seq: u64,
+    ) -> Result<libra_types::transaction::SignedTransaction> {
+        libra_types::transaction::helpers::create_user_txn(
+            self,
+            payload,
+            sender,
+            seq,
+            1_000_000,
+            0,
+            LBR_NAME.to_owned(),
```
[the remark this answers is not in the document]
xli: it is because price unit is zero, so it does not matter what currency we use.

client/faucet/src/mint.rs
```diff
@@ -0,0 +1,150 @@
...
+
+impl MintParams {
+    fn create_parent_vasp_account_script(
+        &self,
+        seq: u64,
+    ) -> libra_types::transaction::TransactionPayload {
+        let receiver = self.auth_key.derived_address();
+        libra_types::transaction::TransactionPayload::Script(
+            transaction_builder_generated::stdlib::encode_create_parent_vasp_account_script(
+                type_tag_for_currency_code(self.currency_code.to_owned()),
```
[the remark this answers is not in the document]
xli: yes, because after created the account, we need transfer coins of the currency to the account. but the script does require a currency anyway.

client/faucet/src/mint.rs
```diff
@@ -0,0 +1,150 @@
...
+
+    fn p2p_script(&self) -> libra_types::transaction::TransactionPayload {
+        let receiver = self.auth_key.derived_address();
+        libra_types::transaction::TransactionPayload::Script(
+            transaction_builder_generated::stdlib::encode_peer_to_peer_with_metadata_script(
+                type_tag_for_currency_code(self.currency_code.to_owned()),
+                receiver,
+                self.amount,
+                vec![],
+                vec![],
```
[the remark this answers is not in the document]
xli: metadata & metadata signature.

sunmilee:
Where does the creation of the DD account happen?

xli:
> Where does the creation of the DD account happen?

wont be added in this diff, aiming for fix existing problem first before we add new feature to support other use cases.

sunmilee:
> > Where does the creation of the DD account happen?
> 
> added into todo list #5943.
> 
> We need focus on fixing existing problem first before we add new feature to support other use cases.

Oh I was asking about "Accepts auth key, amount and currency code for creating online testing account with coins transferred from a **dd account**" <- I  wasn't sure what you meant here. Are you saying faucet is funding accounts through a token DD  account?

xli:
> > > Where does the creation of the DD account happen?
> > 
> > 
> > added into todo list #5943.
> > We need focus on fixing existing problem first before we add new feature to support other use cases.
> 
> Oh I was asking about "Accepts auth key, amount and currency code for creating online testing account with coins transferred from a **dd account**" <- I wasn't sure what you meant here. Are you saying faucet is funding accounts through a token DD account?

Yes, didn't realize this is not well known, as we even discussed to have a rich dd account shared for getting coins in Testnet. 
So, in current faucet with cli implementation, mint coins does 2 things:

1. create parent vasp by treasure compliance account
2. transfer coins from dd account (addres: 0xdd)

Hence you don't see creating dd account. The 0xdd is created by genesis with max coins available for all 3 testnet currencies.

client/faucet/src/mint.rs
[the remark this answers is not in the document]
davidiw: nuance of governance integration to support concurrent transaction generation.

sunmilee (approved):

PR #5934, opened 2020-09-08
[network] refactor most network counters to include context and avoid allocs

phlip9:
Related issue:  #4012

> It's painful to view prometheus metrics or node logs for network when multiple network instances are running, e.g., a validator is connected to the validator swarm via one instance and their fullnode swarm via another instance. The issue is that network metrics don't have an extra label/dimension for the network type to ensure that the values aren't accidentally combining. For example, the Rpc "cancelled requests" metric doesn't have a label for the network type, so it accidentally counts the total number of cancelled requests across all network instances without any way to refine the query to a specific network instance.

+ Make some optimizations so we (almost) never allocate temporary label strings when setting metrics. The exception here is when the network type is `Private(String)`, in which case we only have to allocate once.

+ Push down `NetworkContext` to `Peer`, `Rpc`, and `DirectSend` layer and feed into counter updates.

+ Add new `fn counter_name(network_context: &NetworkContext, label1: &str, ..) -> CounterType {}` pattern to simplify counter usage.

+ Removed random erroneous direct send counter updates in `interface/mod.rs`

+ Try to use `protocol_id` over `protocol` more consistently.

### DIFF
--- a/config/src/network_id.rs
+++ b/config/src/network_id.rs
@@ -124,17 +124,13 @@ impl Default for NetworkId {
 
 impl fmt::Debug for NetworkId {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
-        write!(f, "{}", self)
+        f.write_str(self.as_str())
     }
 }
 
 impl fmt::Display for NetworkId {
-    fn fmt(&self, formatter: &mut fmt::Formatter) -> fmt::Result {
-        match self {
-            NetworkId::Validator => write!(formatter, "Validator"),
-            NetworkId::Public => write!(formatter, "Public"),
-            NetworkId::Private(info) => write!(formatter, "Private({})", info),
-        }
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        f.write_str(self.as_str())
     }
 }
 
@@ -142,6 +138,20 @@ impl NetworkId {
     /// Convenience function to specify the VFN network
     pub fn vfn_network() -> NetworkId {
         NetworkId::Private("vfn".to_string())
+    }
+
+    pub fn as_str(&self) -> &str {
+        match self {
+            NetworkId::Validator => "Validator",
+            NetworkId::Public => "Public",
+            // We used to return "Private({info})" here; however, it's important
+            // to get the network id str without temp allocs, as this is in the
+            // metrics/logging hot path. In theory, someone could set their
+            // network id as `Private("Validator")`, which would result in
+            // confusing metrics/logging output for them, but would not affect
+            // correctness.
+            NetworkId::Private(info) => info.as_ref(),
+        }
     }
 }
 
--- a/config/src/network_id.rs
+++ b/config/src/network_id.rs
@@ -13,6 +13,10 @@ pub struct NetworkContext {
     network_id: NetworkId,
     role: RoleType,
     peer_id: PeerId,
+
+    /// Cache rendering the short PeerId String
+    #[serde(skip)]
+    peer_id_short_str: String,
 }
 
 impl fmt::Debug for NetworkContext {
@@ -26,9 +30,7 @@ impl fmt::Display for NetworkContext {
         write!(
             formatter,
             "[{},{},{}]",
-            self.network_id,
-            self.role,
-            self.peer_id.short_str()
+            self.network_id, self.role, self.peer_id_short_str,
         )
     }
 }
@@ -39,6 +41,7 @@ impl NetworkContext {
             network_id,
             role,
             peer_id,
+            peer_id_short_str: peer_id.short_str(),
         }
     }
 
@@ -46,30 +49,34 @@ impl NetworkContext {
         &self.network_id
     }
 
-    pub fn peer_id(&self) -> PeerId {
-        self.peer_id
-    }
-
     pub fn role(&self) -> RoleType {
         self.role
     }
 
+    pub fn peer_id(&self) -> PeerId {
+        self.peer_id
+    }
+
+    pub fn peer_id_short_str(&self) -> &str {
+        self.peer_id_short_str.as_str()
+    }
+
     #[cfg(any(test, feature = "testing", feature = "fuzzing"))]
     pub fn mock_with_peer_id(peer_id: PeerId) -> std::sync::Arc<Self> {
-        std::sync::Arc::new(Self {
-            network_id: NetworkId::Validator,
-            role: RoleType::Validator,
+        std::sync::Arc::new(Self::new(
+            NetworkId::Validator,
+            RoleType::Validator,
             peer_id,
-        })
+        ))
     }
 
     #[cfg(any(test, feature = "testing", feature = "fuzzing"))]
     pub fn mock() -> std::sync::Arc<Self> {
-        std::sync::Arc::new(Self {
-            network_id: NetworkId::Validator,
-            role: RoleType::Validator,
-            peer_id: PeerId::random(),
-        })
+        std::sync::Arc::new(Self::new(
+            NetworkId::Validator,
+            RoleType::Validator,
+            PeerId::random(),
+        ))
     }
 }
 
--- a/network/src/counters.rs
+++ b/network/src/counters.rs
@@ -136,19 +136,40 @@ pub static LIBRA_NETWORK_DIRECT_SEND_MESSAGES: Lazy<IntCounterVec> = Lazy::new(|
     register_int_counter_vec!(
         "libra_network_direct_send_messages",
         "Number of direct send messages",
-        &["state"]
+        &["role_type", "network_id", "peer_id", "state"]
     )
     .unwrap()
 });
+
+pub fn direct_send_messages(
+    network_context: &NetworkContext,
+    state_label: &'static str,
+) -> IntCounter {
+    LIBRA_NETWORK_DIRECT_SEND_MESSAGES.with_label_values(&[
+        network_context.role().as_str(),
+        network_context.network_id().as_str(),
+        network_context.peer_id_short_str(),
+        state_label,
+    ])
+}
 
 pub static LIBRA_NETWORK_DIRECT_SEND_BYTES: Lazy<HistogramVec> = Lazy::new(|| {
     register_histogram_vec!(
         "libra_network_direct_send_bytes",
         "Number of direct send bytes transferred",
-        &["state"]
+        &["role_type", "network_id", "peer_id", "state"]
     )
     .unwrap()
 });
+
+pub fn direct_send_bytes(network_context: &NetworkContext, state_label: &'static str) -> Histogram {
+    LIBRA_NETWORK_DIRECT_SEND_BYTES.with_label_values(&[
+        network_context.role().as_str(),
+        network_context.network_id().as_str(),
+        network_context.peer_id_short_str(),
+        state_label,
+    ])
+}
 
 /// Counters(queued,dequeued,dropped) related to inbound network notifications for RPCs and
 /// DirectSends.
--- a/network/src/counters.rs
+++ b/network/src/counters.rs
@@ -79,11 +79,8 @@ pub fn rpc_messages(
     ])
 }
 
-// TODO(philiphayes): the default histogram buckets don't make sense for measuring
-// message sizes; we need to set these explicitly to something reasonable...
-
-pub static LIBRA_NETWORK_RPC_BYTES: Lazy<HistogramVec> = Lazy::new(|| {
-    register_histogram_vec!(
+pub static LIBRA_NETWORK_RPC_BYTES: Lazy<IntCounterVec> = Lazy::new(|| {
+    register_int_counter_vec!(
         "libra_network_rpc_bytes",
         "Number of RPC bytes transferred",
         &["role_type", "network_id", "peer_id", "type", "state"]
@@ -95,7 +92,7 @@ pub fn rpc_bytes(
     network_context: &NetworkContext,
     type_label: &'static str,
     state_label: &'static str,
-) -> Histogram {
+) -> IntCounter {
     LIBRA_NETWORK_RPC_BYTES.with_label_values(&[
         network_context.role().as_str(),
         network_context.network_id().as_str(),
@@ -147,8 +144,8 @@ pub fn direct_send_messages(
     ])
 }
 
-pub static LIBRA_NETWORK_DIRECT_SEND_BYTES: Lazy<HistogramVec> = Lazy::new(|| {
-    register_histogram_vec!(
+pub static LIBRA_NETWORK_DIRECT_SEND_BYTES: Lazy<IntCounterVec> = Lazy::new(|| {
+    register_int_counter_vec!(
         "libra_network_direct_send_bytes",
         "Number of direct send bytes transferred",
         &["role_type", "network_id", "peer_id", "state"]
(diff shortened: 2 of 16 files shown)


andll:
Note that each combination of labels on the counter is basically creating a new counter. We need a good estimate on the upper bound of how many counters does this create, to make sure we don't blow amount of counters. given that functions in this PR takes a lot of parameters, I would suspect there might easily be too many counters

Little bit of history - last time we did some change like this we killed prometheus :)

phlip9:
definitely gotta keep label cardinality in mind.

for reference, we have four big counters here (rpc msgs/bytes and direct send msgs/bytes)

```
{ big counters } <= { rpc msgs + rpc bytes + ds msgs + ds bytes } x { network contexts } x { protocol ids } x { types } x { states }
|{ big counters }| <= 4 x 4 x 6 x 2 x 5
                   <= 960
```

although, now that I think about it, the bytes counters are histograms, so that's probably more... 

do you have any intuition for ~ order of magnitude counters prometheus can handle? i could easily remove `protocol_id` label and reduce # counters by factor of 6.

andll:
I don't have intuition on how much it can handle(especially long term).
cc @sherry-x / @mgorven 

I think good analysis would be comparing how many counters we had before this pr / after. If it grows 10x its probably a concern :)

mgorven:
1k counters just for this is quite a lot. What are the `protocol_id` and `state` values?

config/src/network_id.rs
```diff
@@ -143,6 +139,17 @@ impl NetworkId {
...
     }
+
+    /// Return the NetworkId as a copy-on-write str slice. Certain variants are
+    /// constant, so we can just return a `&'static str` instead of allocating
+    /// an extra temporary string.
+    pub fn as_cow_str(&self) -> Cow<'static, str> {
+        match self {
+            NetworkId::Validator => Cow::Borrowed("Validator"),
+            NetworkId::Public => Cow::Borrowed("Public"),
+            NetworkId::Private(info) => Cow::Owned(format!("Private({})", info)),
```
gregnazario, line 157: Realistically, we could probably just print the `info` and ignore the Private part

network/src/interface/mod.rs
```diff
@@ -105,6 +107,7 @@ where
         );
         let peer_handle = PeerHandle::new(peer_id, peer_reqs_tx);
         let peer = Peer::new(
+            Arc::clone(&network_context),
```
gregnazario, line 110: What's the difference between `Arc::clone(obj)` and `obj.clone()`

config/src/network_id.rs
```diff
@@ -13,6 +13,10 @@ pub struct NetworkContext {
     network_id: NetworkId,
     role: RoleType,
     peer_id: PeerId,
+
+    /// Cache rendering the short PeerId String
+    #[serde(skip)]
+    peer_id_short_str: String,
```
gregnazario, line 19: Personally, this always gets confusing to me when it's named `str` but it's a `String`

network/src/counters.rs
```diff
@@ -33,7 +34,7 @@ pub static LIBRA_CONNECTIONS: Lazy<IntGaugeVec> = Lazy::new(|| {
     register_int_gauge_vec!(
         "libra_connections",
         "Number of current connections and their direction",
-        &["role_type", "network_id", "peer_id", "direction"]
+        &["role_type", "network_id", "self_peer_id", "direction"]
```
gregnazario, line 37: Why rename this?

network/src/counters.rs
```diff
@@ -66,29 +67,91 @@ pub static LIBRA_NETWORK_RPC_MESSAGES: Lazy<IntCounterVec> = Lazy::new(|| {
...
+        network_context.network_id().as_cow_str().as_ref(),
+        network_context.peer_id_short_str(),
+        protocol_id.as_str(),
+        type_label,
+        state_label,
+    ])
+}
+
+// TODO(philiphayes): the default histogram buckets don't make sense for measuring
+// message sizes; we need to set these explicitly to something reasonable...
```
gregnazario, line 93: Here's a good callout.

Right now you can only look at 1 minute buckets, and we should really be looking at them in aggregate.  Do these make sense as a metric?  Does knowing how large a message in aggregate tell us anything?  Otherwise, structured logging might be a better place for per request metrics like this.

gregnazario (commented):
There's some good stuff in here, and having the `network_context` pushed through, will make the rest of my log cleanup easier!

I think we should think about which logs are helpful in aggregate, and if they aren't helpful in aggregate we can either:
A) Move them to structured logs
B) Delete them

In addition, we may need to remove the changes you made to expand the number of metrics.  After talking to @ankushagarwal earlier, the cardinality of an individual field should be < 1000, ideally 100 or less.  Which, means we need to be careful when we're posting incoming metrics by something like PeerId, because the metrics system could be overloaded either accidentally or on purpose.

As for the number of fields multiples for the number of metrics, we need to make some sort of test for that, on the total number of metrics.  I think we can actually cut out the `role` field, of each one, because that's added on *again* by the metrics system outside of this portion.

config/src/network_id.rs
phlip9: agreed, we could easily make this just print the inner `info`. would also make this more efficient

i guess someone could name their private network id "Validator", but that would be their fault haha

network/src/interface/mod.rs
phlip9: They're the exact same behaviour. At some point we had a coding guideline to use `Arc::clone(&my_arc)` instead of `my_arc.clone()` so we could easily grep all the `Arc::clone`'s we were doing (for perf debugging or something). Though, I can't seem to find this written down in the repo anywhere...

@bmwill: do you remember this or am I just hallucinating? is this still a thing?

config/src/network_id.rs
phlip9: for sure, i can change the field to `peer_id_short_string` and keep the acessor as `peer_id_short_str() -> &str`

gregnazario: Oh, I wouldn't rename it.  I was just pointing it out, because `peer_id.short_str()` exists.

network/src/counters.rs
phlip9: i'm finding it confusing distinguishing b/w `peer_id` where it's your own id and `peer_id` where it's actually a remote peer id. i was also briefly considering adding a `remote_peer_id` label, but removed it for obvious reasons (label explosion)

network/src/interface/mod.rs
bmwill: We actually ripped that out of the coding guidelines as of last week because everyone hated typing the extra characters :)

config/src/network_id.rs
bmwill: Maybe you can avoid the Alloc altogether and create a wrapping type `struct ShortPeerId<'a>(&'a PeerId)` which just has a display impl that does the short printing

network/src/counters.rs
phlip9: i wrote this down here because the default histogram buckets are targeted towards measuring "service latency" (https://docs.rs/prometheus/0.10.0/src/prometheus/histogram.rs.html#25) and so they look like 

```rust
pub const DEFAULT_BUCKETS: &[f64; 11] = &[
    0.005, 0.01, 0.025, 0.05, 0.1, 0.25, 0.5, 1.0, 2.5, 5.0, 10.0,
];
```

which doesn't make sense for measuring request sizes, since 99% will end up in the last bucket

as for how useful measuring the rpc message sizes is... I've never had any particular reason to look at them yet, but it might be useful for seeing bandwidth usage by network type and direction. maybe we just put a blanket metric on the actual socket r/w and remove the per-protocol counters?

phlip9:
removed `protocol_id` label

andll:
One ask - can you setup your own cluster and check this change before landing, to make sure prometheus is ok with whatever counters we created in this change?

network/src/counters.rs
gregnazario: I've been tackling this in the logging, by leaving the `peer_id` from `network_context` named `peer_id` and naming the external ones `remote_peer_id`.  Seemed a little straightforward to me not to worry about `self` for itself.

andll:
Or alternatively - perhaps we can add some safeguard to metrics endpoint that will limit number of counters exposed by the node and complain if this number is exceeded? @bmwill / @gregnazario what do you think?

bmwill:
That's an interesting thought. I think that in the short term that may not be feasible but in the longer term we might be able to do that.

andll:
@bmwill curious, why do you think it is no feasible short term? Can we just intercept `encode_metrics` in the metrics server and limit number of metrics in response, and do some panic or log error if this is above?

gregnazario:
> @bmwill curious, why do you think it is no feasible short term? Can we just intercept `encode_metrics` in the metrics server and limit number of metrics in response, and do some panic or log error if this is above?

How would we determine which metrics to keep and which ones to drop?

andll:
We can drop arbitrary metrics and issue log cricial. Dropping any metrics is strictly better then killing prometheus and loosing all metrics

bmwill:
As Greg identifying which metrics to drop is the issue. I think that intercepting at the `encode_metrics` level is too late, if we want to solve the issue we'd need to fix the issue well before that point IMO.

andll:
So what is proposed way to verify that PRs like this do not affect prometheus?

phlip9:
I think my original upper bounds were pretty loose, so here's tightening my original (rough) # counters upper bounds, without the protocol_id:

```
|rpc msgs| <= |networks| * |{request, response}| * |{sent, received, failed, canceled, declined}|
           <= 3 * 2 * 5
           <= 30

|rpc bytes| <= |networks| * |{request, response}| * |{sent, received}|
            <= 3 * 2 * 2
            <= 12

|rpc latency| <= |networks| * |{request}| * |{sent}|
              <= 3

|ds msgs| <= |networks| * |{sent, recevied, failed}|
          <= 3 * 3
          <= 9

|ds bytes| <= |networks| * |{sent, received}|
           <= 3 * 2
           <= 6

|counters| <= |rpc msgs| + |rpc bytes| + |rpc latency| + |ds msgs| + |ds bytes|
           <= 30 + 12 + 3 + 9 + 6
           <= 60
```

So at most 60 counters vs ~20 before this PR

network/src/peer/mod.rs
```diff
@@ -186,7 +191,7 @@ where
                                 Some(Err(err)) => {
                                     warn!("Failure in reading messages from socket from peer: {}. Error: {:?}",
                                         self_peer_id.short_str(), err);
-                                    self.close_connection( DisconnectReason::ConnectionLost).await;
+                                    self.close_connection(DisconnectReason::ConnectionLost).await;
```
gregnazario, line 194: Always weird some things are just not picked up by the formatter.

gregnazario (approved):
Let's roll with it.  Two things though:

1. Let's try to run ClusterTest to make sure that prometheus doesn't fall over
2. Update the Dashboards with the new metrics if you can, or we can come back and adjust it when we adjust the rest of the dashboards.

andll:
I am sorry, but we can't run cluster test 'to make sure that prometheus doesn't fall over'. Prometheus is shared resource in cluster test, who is going to repair it after it fail?

We need failsafe mechanism to prevent this from happening, not trying to break cluster test

gregnazario:
> I am sorry, but we can't run cluster test 'to make sure that prometheus doesn't fall over'. Prometheus is shared resource in cluster test, who is going to repair it after it fail?
> 
> We need failsafe mechanism to prevent this from happening, not trying to break cluster test

I agree with you we need a failsafe mechanism to prevent this from happening, and we don't want to break cluster test.  I'm looking for solutions in which we can unblock Philip from getting this change through right now.

I am currently looking into a way to better help us decide how we can handle when we have "too many" metrics.

How I see it there are a couple ways:
1. Drop arbitrary metrics if you have too many based on some arbitrary limit and alarm.  This is problematic, because through a deployment you could lose the important metrics that the blockchain oncall uses.  So this is a no-go, especially with how long our current deployments take.
2. Prioritize metrics, and drop less important metrics and alarm.  This is better, because it means we can keep on running, while losing the new metrics that were just added.  But, how do we know what to prioritize?  And what if we're adding important metrics?  How do we scale prometheus after that?
3. Have some sort of analysis tool that can tell us before we deploy if we have too many counters.  I'm suggesting this with cluster test for this instance, and we don't want to use it permanently.  But even if we had a static analysis tool to just count the number of counters, and reject them, what happens when we hit the limit?  Do we stop adding metrics?

So, we need to probably as well look at how we scale prometheus for when that time comes.  Especially since we're asking everyone to add metrics.

config/src/network_id.rs
phlip9: I actually wrote something like that a long time ago but forgot to merge it in... I'll open a follow up PR with it

network/src/counters.rs
phlip9: ok, seems good to me. just gotta be consistent with our convention

config/src/network_id.rs
davidiw: What motivated this change? This feels like a premature optimization.

config/src/network_id.rs
davidiw: Same as above. We may not make a copy here, but we make a ton more copies to route this to the eventual logs.

network/src/counters.rs
davidiw: No skin in the game, but we ought to be consistent across logging and metrics. Perhaps we can use a common const or enum value?

network/src/counters.rs
davidiw: If we're going to log message sizes, I'd say the most important is peer > message type (consensus, mempool) > protocol type (request, reply, send). In general, I don't think we'll get much out of protocol type. Could remove that to reduce the amount of metrics being created.

davidiw: also this might be fine to just be a counter, not sure we get much out of a histogram... I think in ODS we could take something like the following

(value at time t1 - value at time t0) / (t0 - t1) and do that at second or 60 second intervals to make some pretty graphs, not sure what prometheus / grafana offer.

network/src/counters.rs
```diff
@@ -66,47 +60,111 @@ pub static LIBRA_NETWORK_RPC_MESSAGES: Lazy<IntCounterVec> = Lazy::new(|| {
...
     register_int_counter_vec!(
         "libra_network_direct_send_messages",
         "Number of direct send messages",
-        &["state"]
+        &["role_type", "network_id", "self_peer_id", "state"]
     )
     .unwrap()
 });
 
+pub fn direct_send_messages(
```
davidiw, line 138: again would recommend we just aggregate this ignoring the protocol type and focus on the message type.... though it isn't quite clear to me what we gain with this metric.

davidiw (commented):
This is super exciting, but I'd like to have a quick meeting on what metrics we want and the space that we have.

With respect to the number of metrics, this is the first I've heard of a limit, I heard indirectly from @ankushagarwal the limit was at the per dimension level and it was no more than 1000 distinct active labels. Now we're saying we don't want 1000 distinct labels across all metrics (or effectively 1000 distinct metrics). I'd love to nail this down and get it written in our coding guidelines or something.

network/src/counters.rs
phlip9: yep, I removed the protocol type. I would be fine setting the `rpc_bytes` and `direct_send_bytes` to just plain counters

network/src/counters.rs
```diff
@@ -79,11 +79,8 @@ pub fn rpc_messages(
     ])
 }
 
-// TODO(philiphayes): the default histogram buckets don't make sense for measuring
-// message sizes; we need to set these explicitly to something reasonable...
-
-pub static LIBRA_NETWORK_RPC_BYTES: Lazy<HistogramVec> = Lazy::new(|| {
-    register_histogram_vec!(
+pub static LIBRA_NETWORK_RPC_BYTES: Lazy<IntCounterVec> = Lazy::new(|| {
+    register_int_counter_vec!(
```
gregnazario, line 83: Nice

phlip9:
ran on CT no problem. did some overhauling of network dashboards too

/land

phlip9:
looks like dashboards got moved out of libra/libra, so putting those in a separate PR in the appropriate repo

/land

PR #5935, opened 2020-09-08
[cli] fix cli logger

xli:
ensure init once.

## Test Plan

* manual test with libra-swarm

### DIFF
--- a/testsuite/cli/src/main.rs
+++ b/testsuite/cli/src/main.rs
@@ -79,16 +79,14 @@ struct Args {
 }
 
 fn main() {
-    ::libra_logger::Logger::new().init();
-    crash_handler::setup_panic_handler();
     let args = Args::from_args();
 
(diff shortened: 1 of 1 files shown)


testsuite/cli/src/main.rs
```diff
@@ -79,16 +79,14 @@ struct Args {
...
-    crash_handler::setup_panic_handler();
     let args = Args::from_args();
 
     let mut logger = ::libra_logger::Logger::new();
-
     if !args.verbose {
         logger.level(::libra_logger::Level::Warn);
     }
     logger.init();
+    crash_handler::setup_panic_handler();
```
xli, line 89: this needs logger setup, move to after init

bmwill: Hmm I seemed to have missed this when I tried to fix it last week. Thanks for the catch!

bmwill (approved):
/land

PR #5939, opened 2020-09-08
[libra framework] Add documentation for transaction scripts

tzakian:
Initial documentation for all transaction scripts. Minimal code changes.

Updated the document generator to:
* Handle constants properly
* Not put `<code>...</code>` on a newline as this was breaking code formatting in tables and enumerations. Seems to be working fine. 

Every transaction script is documented following a set template of:

---
```
/// # Summary
/// <summary here>
/// # Technical Description
/// <more detailed/technical description here>
/// # Parameters
/// | Name | Type | Description |
/// |------|------|-------------|
/// <table of parameters, their type, and brief description>
///
/// # Common Abort Conditions
/// | Error Category | Error Reason | Description |
/// |----------------|--------------|-------------|
/// <table of errors>
///
/// # Related Scripts
/// <any scripts that can be thought of as being in a similar category>
```
---


language/stdlib/transaction_scripts/add_recovery_rotation_capability.move
```diff
@@ -2,9 +2,25 @@ script {
...
 
-/// Add the `KeyRotationCapability` for `to_recover_account` to the `RecoveryAddress` resource under `recovery_address`.
-///
-/// ## Aborts
+/// # Sender Preconditions
+/// The sending account (`to_recover_account`) must not have previously given away its key
+/// rotation capability, and must be a VASP account. The account at `recovery_address`
+/// must also be a VASP account belonging to the same VASP as the `to_recover_account`.
+/// Additionally the account at `recovery_address` must have already initalized itself as
+/// a recovery account address using the `create_recovery_address` transaction script.
```
sblackshear, line 10: Is it possible to make this hyperlink to the `create_recovery_address` script? I see that the doc generator does this automatically for types/functions/modules; not sure if there is or easily could be a way to do it with scripts. CC @wrwg

language/stdlib/transaction_scripts/add_currency_to_account.move
```diff
@@ -5,10 +5,19 @@ use 0x1::Libra;
 use 0x1::Roles;
 use 0x1::Signer;
 
-/// Add a `Currency` balance to `account`, which will enable `account` to send and receive
-/// `Libra<Currency>`.
-/// Aborts with NOT_A_CURRENCY if `Currency` is not an accepted currency type in the Libra system
-/// Aborts with `LibraAccount::ADD_EXISTING_CURRENCY` if the account already holds a balance in
+/// # Sender Preconditions
```
sblackshear, line 8: Nit: do we want to say "Signer Preconditions" to make this schema continue to work when we want to describe scripts with multiple signers?

language/stdlib/transaction_scripts/add_to_script_allow_list.move
```diff
@@ -2,6 +2,28 @@ script {
 use 0x1::LibraTransactionPublishingOption;
 use 0x1::SlidingNonce;
 
+/// # Sender Preconditions
+/// The sending account (`lr_account`) must be the Libra Root account. The script allow
+/// list must not already hold the script `hash` being added.
+/// # Accounts Affected
+/// The Libra Root account is affected with normal transaction effects (sequence number,
```
sblackshear, line 9: Do we want to say this, since it's true of every tx?

language/stdlib/transaction_scripts/add_to_script_allow_list.move
```diff
@@ -2,6 +2,28 @@ script {
...
+/// nonce). Additionally the on-chan config
```
sblackshear, line 10:
```suggestion
/// nonce). Additionally the on-chain config
```

language/stdlib/transaction_scripts/add_to_script_allow_list.move
```diff
@@ -2,6 +2,28 @@ script {
...
+/// `LibraTransactionPublishingOption::LibraTransactionPublishingOption`'s
+///`script_allow_list` field is updated to contain the new script `hash`.
```
sblackshear, line 12:
```suggestion
/// `script_allow_list` field is updated to contain the new script `hash`.
```

language/stdlib/transaction_scripts/add_currency_to_account.move
```diff
@@ -5,10 +5,19 @@ use 0x1::Libra;
...
+/// The sending account must have a role that allows holding balances (VASP or DD). The
+/// currency being added must be a registerd on-chain currency.
```
sblackshear, line 10:
```suggestion
/// currency being added must be a registered on-chain currency.
```

language/stdlib/transaction_scripts/add_currency_to_account.move
```diff
@@ -5,10 +5,19 @@ use 0x1::Libra;
...
+/// # Accounts Affected
+/// Only the sending account is updated. Transaction fees may be collected.
+/// # Accounts Created
+/// None
+/// # Description
+/// Adds a `Currency` balance to the sending `account`. This will enable `account` to
+/// send, receive, and hold coins in that currency (`Libra<Currency>`).
```
sblackshear, line 17:
```suggestion
/// send, receive, and hold coins in `Libra::Libra<Currency>`.
```

I think the extra `Libra::`makes the doc generator to link to the correct resource.

language/stdlib/transaction_scripts/add_currency_to_account.move
```diff
@@ -5,10 +5,19 @@ use 0x1::Libra;
...
+/// # Abort Conditions
+/// * Aborts with NOT_A_CURRENCY if `Currency` is not an accepted currency type in the Libra system
+/// * Aborts with `LibraAccount::ADD_EXISTING_CURRENCY` if the account already holds a balance in
```
sblackshear, line 20: It looks like the doc generator does not currently hyperlink to constants. Would it be hard to fix this?

language/stdlib/transaction_scripts/add_to_script_allow_list.move
```diff
@@ -2,6 +2,28 @@ script {
...
+/// # Accounts Created
+/// None
+/// # Description
+/// Adds the script hash `hash` to the script allow list. After this transaction is
+/// committed, transactions with this `hash` may be sent to the network.
```
sblackshear, line 17: I wonder if we should say "script executes successfully" or choose some other terminology instead of "transaction is committed". "is committed" seems both too vague (e.g., it might have committed and aborted) and too specific (e.g., two dependent transactions can be in the same block rather than waiting for the first to be committed).

I'm ok with any terminology, but think that we should be careful/consistent here...

language/stdlib/transaction_scripts/add_validator_and_reconfigure.move
```diff
@@ -1,20 +1,45 @@
...
-        validator_name: vector<u8>,
-        validator_address: address
-    ) {
-        SlidingNonce::record_nonce_or_abort(lr_account, sliding_nonce);
-        assert(ValidatorConfig::get_human_name(validator_address) == validator_name, 0);
-        LibraSystem::add_validator(lr_account, validator_address);
-    }
+/// # Sender Preconditions
+/// The sending account (`lr_account`) must be the Libra Root account. The
+/// `sliding_nonce` must be a valid nonce for the Libra Root account. The account at
```
sblackshear, line 8: We mention the nonce in the preconditions + abort conditions here, but only in the abort conditions for `add_to_script_allow_list`. Not sure what is preferable, but we should probably be consistent.

sblackshear (commented):
As they are currently written, "sender preconditions" and "abort conditions" are sometimes redundant. I see that the intention is for the first to be high-level and the second to be detailed, but they often end up looking similar. I especially felt this with `add_validator_and_reconfigure.move`. 

I'm wondering if it might sense to
- regularize the meaning of `Sender preconditions` to "stuff you gotta do before running this script"  (e.g., perhaps it could list only the script(s) that must be run beforehand?) 
- combine the two sections?

language/stdlib/transaction_scripts/cancel_burn.move
```diff
@@ -1,8 +1,35 @@
...
+/// The transaction must be sent by an account (`account`) with a `Libra::BurnCapability<CoinType>`
+/// published under it. The account at `preburn_address` must have a preburn resource for
+/// `CoinType` published under it, and its value must be nonzero. The transaction removes
+/// the entire balance held in the preburn resource, and returns it back to the account's
+/// balance (`LibraAccount::Balance<CoinType>`) at `preburn_address`. Due to this, the account at
+/// `preburn_address` must already have a balance in the `CoinType` currency published
+/// before this script is called otherwise the transaction will fail.
+///
+/// # Accounts Affected
+/// The sending account will have normal transaction-related effects. The account at
```
sblackshear, line 14:
```suggestion
/// The account at
```

language/stdlib/transaction_scripts/cancel_burn.move
```diff
@@ -1,8 +1,35 @@
...
+/// `preburn_address` will have the entirety of its preburn balance removed, and deposited
```
sblackshear, line 15:
```suggestion
/// `preburn_address` will have the entirety of its `Preburn<CoinType>` balance removed, and deposited
```

language/stdlib/transaction_scripts/cancel_burn.move
```diff
@@ -1,8 +1,35 @@
...
+/// into the balance of `CoinType` currency that it holds.
```
sblackshear, line 16:
```suggestion
/// into the `Balance<CoinType>` currency that it holds.
```

language/stdlib/transaction_scripts/cancel_burn.move
```diff
@@ -1,8 +1,35 @@
...
+///
+/// # Accounts Created
+/// None
+///
+/// # Description
+/// Cancel and return all coins held in the preburn area under the `preburn_address` and
```
sblackshear, line 22:
```suggestion
/// Cancel and return all coins held in the `Preburn<CoinType>` resource under the `preburn_address` and
```

language/stdlib/transaction_scripts/cancel_burn.move
```diff
@@ -1,8 +1,35 @@
...
+/// return the funds to the `preburn_address` account's balance.
```
sblackshear, line 23:
```suggestion
/// return the funds to the `preburn_address` account's `Balance<CoinType>`.
```

language/stdlib/transaction_scripts/create_child_vasp_account.move
```diff
@@ -6,14 +6,36 @@ use 0x1::VASP;
...
 
-/// Create a `ChildVASP` account for sender `parent_vasp` at `child_address` with a balance of
-/// `child_initial_balance` in `CoinType` and an initial authentication_key
+/// # Sender Preconditions
+/// The sending account must be a valid Parent VASP account and `CoinType` must be a
+/// registered on-chain currency. An account must not already exist at the `child_address`
+/// address.
+///
+/// # Accounts Affected
+/// The sending account will have normal transaction-related effects. The transaction will
```
sblackshear, line 15:
```suggestion
/// The transaction will
```

language/stdlib/transaction_scripts/create_child_vasp_account.move
```diff
@@ -6,14 +6,36 @@ use 0x1::VASP;
...
+/// create a new account with a Child VASP role at the specified `child_address` with at
+/// least a balance in `CoinType` (and possibly others if the `add_all_currencies` argument
+/// is `true`).
+///
+/// The creating Parent VASP's account will have the number of children that it has
```
sblackshear, line 20: Link to the field?

language/stdlib/transaction_scripts/create_child_vasp_account.move
```diff
@@ -6,14 +6,36 @@ use 0x1::VASP;
...
+/// incremented.
+///
+/// # Accounts Created
+/// A new account with a Child VASP role and belonging to the Parent VASP account that
+/// sent this transaction (`parent_vasp`) is created at `child_address`.
+///
+/// # Description
+/// Creates a `ChildVASP` account for the sender `parent_vasp` at `child_address` with a balance of
+/// `child_initial_balance` in `CoinType` and an initial authentication_key of
```
sblackshear, line 29:
```suggestion
/// `child_initial_balance` in `CoinType` and an initial authentication key of
```

sblackshear (approved):
Thank you for doing this, these are great!! Lots of comments but nothing major, feel free to ignore if too annoying.

language/stdlib/transaction_scripts/add_currency_to_account.move
```diff
@@ -5,10 +5,19 @@ use 0x1::Libra;
...
 
-/// Add a `Currency` balance to `account`, which will enable `account` to send and receive
-/// `Libra<Currency>`.
-/// Aborts with NOT_A_CURRENCY if `Currency` is not an accepted currency type in the Libra system
-/// Aborts with `LibraAccount::ADD_EXISTING_CURRENCY` if the account already holds a balance in
+/// # Sender Preconditions
+/// The sending account must have a role that allows holding balances (VASP or DD). The
+/// currency being added must be a registerd on-chain currency.
+/// # Accounts Affected
+/// Only the sending account is updated. Transaction fees may be collected.
```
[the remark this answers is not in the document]
tzakian: yea... I need to go back and fix! I am now describing these as "normal transaction-related effects" if that sounds good? Then we can define what those mean in a doc somewhere if need be

language/stdlib/transaction_scripts/add_recovery_rotation_capability.move
tzakian: That would be awesome! Also just as a note to myself: I am realizing that constants aren't linking properly, so I will need to go in and fixing that at some point this week.

language/stdlib/transaction_scripts/add_currency_to_account.move
tzakian: Signer preconditions I think makes sense, but  really like "script lifecycle" so will probably change to that to see if we like it if that sounds good

language/stdlib/transaction_scripts/add_to_script_allow_list.move
tzakian: Not entirely sure,  think it's something that we should ask tomorrow morning on the call, it would be nice if we didn't!

language/stdlib/transaction_scripts/add_currency_to_account.move
tzakian: Don't think it should be, and they should work.... I must have messed up something when I added them so I'll add it to my todo list (maybe fixup in this PR if it's something super simple)

language/stdlib/transaction_scripts/add_to_script_allow_list.move
tzakian: Agreed on terminology. I think this is another thing to bring up tomorrow morning as well possibly. 

I'm leaning towards "is executed successfully and committed" since that should be unambiguous. And I think there's a distinction to be made (that may be too technical, so fine not making it) between "executed successfully" and "executed successfully and committed" especially around reconfigurations. In my thinking the former is saying "dependencies within the block are ok" and the latter is saying "dependencies on this within the block aren't"?

language/stdlib/transaction_scripts/add_validator_and_reconfigure.move
tzakian: ahh good catch. I'll update `add_to_script_allow_list` to mention this as well.

language/stdlib/transaction_scripts/burn.move
```diff
@@ -2,10 +2,40 @@ script {
 use 0x1::Libra;
 use 0x1::SlidingNonce;
 
-/// Permanently destroy the `Token`s stored in the oldest burn request under the `Preburn` resource.
-/// This will only succeed if `account` has a `MintCapability<Token>`, a `Preburn<Token>` resource
-/// exists under `preburn_address`, and there is a pending burn request.
-/// sliding_nonce is a unique nonce for operation, see sliding_nonce.move for details
+/// # Sender Preconditions
+/// The sending account `account` must have a burn capability for the given currency
+/// (`Libra::BurnCapability<Currency>`) published at the top-level under it. The account
```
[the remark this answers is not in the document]
tzakian: Good call out, I think not specifically mentioning "at the the top-level" makes sense. I'll go through and update

language/stdlib/transaction_scripts/add_currency_to_account.move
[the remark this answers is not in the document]
sblackshear: I guess my main issue with this is that if the doc for an individual transaction says that it has "normal transaction-related effects", it makes me think that there might be some transactions that do not... I would personally opt for omitting this in the individual tx docs and explaining it in the single "index doc" that has a table linking to all the types.

language/move-prover/docgen/src/docgen.rs
```diff
@@ -1000,41 +993,43 @@ impl<'env> Docgen<'env> {
...
-                None
-            }
-        };
+        let try_func_struct_or_const =
+            |module: &ModuleEnv<'_>, name: Symbol, is_qualified: bool| {
+                // Below we only resolve a simple name to a function if it is followed by a ( or <.
+                // Otherwise we get too many false positives where names are resolved to functions
+                // but are actually fields.
+                if module.find_struct(name).is_some()
+                    || module.find_named_constant(name).is_some()
```
wrwg, line 1002: Update comment above.

wrwg (approved):
Looks great. Let's get this in and iterate as needed.

PR #5940, opened 2020-09-08
[dkr] Add more duct tape to faucet

mgorven:
* Increase mint timeout to 10s. This was initially set when minting was
  completely non-blocking, it now blocks and has to deal with fullnode
  sync delay.
* Don't kill CLI after a successful mint.

I can do serial mint requests with no errors on my testnet with these
changes.

### DIFF
--- a/docker/mint/server.py
+++ b/docker/mint/server.py
@@ -66,18 +66,16 @@ def send_transaction():
     try:
         create_client()
         application.client.sendline("q as 000000000000000000000000000000dd")
-        application.client.expect(r"sequence_number: ([0-9]+)", timeout=1)
+        application.client.expect(r"sequence_number: ([0-9]+)", timeout=2)
         if application.client.match:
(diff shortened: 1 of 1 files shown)


sunmilee (approved):

docker/mint/server.py
```diff
@@ -66,18 +66,16 @@ def send_transaction():
     try:
         create_client()
         application.client.sendline("q as 000000000000000000000000000000dd")
-        application.client.expect(r"sequence_number: ([0-9]+)", timeout=1)
+        application.client.expect(r"sequence_number: ([0-9]+)", timeout=2)
         if application.client.match:
             next_dd_seq = int(application.client.match.groups()[0]) + 1
         else:
-            return 'DD sequence number not found', 400
+            raise Exception('DD sequence number not found')
```
sausagee, line 73: Probably a good call to raise and kill the client in this case.

sausagee (approved):

Issue #5917, opened 2020-09-04
[Bug][move compiler] infinite loop + unused signer produces non-verifiable code


sblackshear:
The following script compiles, but produces code that makes the bytecode verifier unhappy:

```
script {
fun main(_account: &signer) {
   loop { }
}
}
```

I see the message

```
thread 'main' panicked at 'assertion failed: self.is_canonical() && other.is_canonical()', language/bytecode-verifier/src/reference_safety/abstract_state.rs:599:9
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

Not sure whether this is a problem in the compiler or in the verifier.

tnowacki:
Definitely a bug in the bytecode verifier (it should never fail...)

tnowacki:
#5941

PR #5941, opened 2020-09-08
[bug fix][bytecode verifier] Fixed canonical initial state

tnowacki:
- Initial state of the reference safety verifier was not canonical
- Causes issues if the offset 0 is a loop head

## Motivation

#5917 

## Test Plan

- new test (verified the error was hit and that the fix works)


sblackshear (approved):
thanks for the quick fix!

shazqadeer (approved):
Thanks for the quick fix.  It is great that you added the assertion in the constructor of the abstract state.  If that assertion was already in place, existing tests would likely have caught the error.

PR #5945, opened 2020-09-09
[ci] reenable hadolint, add canary support to calculating changed files.

rexhoffman:
## Motivation

Test canary and treat prs like they were auto, with target branches, etc.

Y

## Test Plan

CI

## Related PRs

None

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -586,11 +586,13 @@ jobs:
       - run:
           name: Lint DockerFile changes
           command: |
-            #eval `.circleci/get_pr_info.sh -b`
-            #if  [[ -n "$BASE_GITHASH" ]] && [[ -n "$CHANGED_FILE_OUTPUTFILE" ]] && [[ "wc -l < $CHANGED_FILE_OUTPUTFILE" != 0 ]]; then
-            #  CHANGED_DOCKER_FILES=$(cat "$CHANGED_FILE_OUTPUTFILE" | grep Dockerfile | tr '\n' ' ')
-            #  hadolint -c .lintrules/hadolint.yaml $CHANGED_DOCKER_FILES || true
-            #fi
+            eval $(.circleci/get_pr_info.sh -b)
+            if  [[ -n "$BASE_GITHASH" ]] && [[ -n "$CHANGED_FILE_OUTPUTFILE" ]]; then
+              CHANGED_DOCKER_FILE_LIST=$( (grep Dockerfile $CHANGED_FILE_OUTPUTFILE || true) | tr '\n' ' ' )
+              if [[ -n "$CHANGED_DOCKER_FILE_LIST" ]]; then
+                hadolint -c .lintrules/hadolint.yaml "$CHANGED_DOCKER_FILE_LIST" || true
+              fi
+            fi
       - fail_if_debian_buster_docker_base_images_are_not_tagged_with_sha256
   check-breaking-change:
     executor: audit-executor
--- a/.circleci/get_pr_info.sh
+++ b/.circleci/get_pr_info.sh
@@ -89,9 +89,9 @@ if [[ -n "$USERSET_PREVIOUS_GITHASH" ]] && git merge-base --is-ancestor "$USERSE
 fi
 
 #If we can find the PR number it overrides any user input git hash.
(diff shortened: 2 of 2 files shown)


.circleci/config.yml
```diff
@@ -598,10 +598,14 @@ jobs:
...
           command: |
-            eval `.circleci/get_pr_info.sh -b`
-            CHANGED_DOCKER_FILES=`cat "$CHANGED_FILE_OUTPUTFILE" | grep Dockerfile | tr '\n' ' '`
-            if [[ ! -z "$CHANGED_DOCKER_FILES" ]]; then
-              hadolint -c .lintrules/hadolint.yaml $CHANGED_DOCKER_FILES || true
+            target_info=$(.circleci/get_pr_info.sh -b)
+            echo "$target_info"
+            eval "$target_info"
+            if  [[ -n "$BASE_GITHASH" ]] && [[ -n "$CHANGED_FILE_OUTPUTFILE" ]]; then
+              CHANGED_DOCKER_FILE_LIST=$( grep Dockerfile $CHANGED_FILE_OUTPUTFILE  || true | tr '\n' ' ' || true )
```
rexhoffman, line 605: apparently grep will exit 1 if it doesn't find matches....  found this to be very weird, and couldn't repro on my mac, but this works.

docker/mint/Dockerfile
```diff
@@ -1,6 +1,4 @@
-FROM debian:buster-20200803@sha256:a44ab0cca6cd9411032d180bc396f19bc98f71972d2398d50460145cab81c5ab AS debian-base
-
-FROM debian-base AS toolchain
+FROM debian:buster-20200803@sha256:a44ab0cca6cd9411032d180bc396f19bc98f71972d2398d50460145cab81c5ab AS toolchain
```
rexhoffman, line 1: doing this to make the first 20 or so lines of docker files identical across files.   Thought my change might have been adding time but it was a change to the validator docker file.   Still want the lines identical so I can add a test.

rexhoffman:
the canary had a unit test failure due to port conflicts, reran and it was fine.

.circleci/config.yml
```diff
@@ -598,11 +598,15 @@ jobs:
...
           name: Lint DockerFile changes
           command: |
-            #eval `.circleci/get_pr_info.sh -b`
-            #if  [[ -n "$BASE_GITHASH" ]] && [[ -n "$CHANGED_FILE_OUTPUTFILE" ]] && [[ "wc -l < $CHANGED_FILE_OUTPUTFILE" != 0 ]]; then
-            #  CHANGED_DOCKER_FILES=$(cat "$CHANGED_FILE_OUTPUTFILE" | grep Dockerfile | tr '\n' ' ')
-            #  hadolint -c .lintrules/hadolint.yaml $CHANGED_DOCKER_FILES || true
-            #fi
+            target_info=$(.circleci/get_pr_info.sh -b)
+            echo "$target_info"
+            eval "$target_info"
```
sausagee, line 603: Maybe try `set -ex` with the original `eval $(blah)` so that we can what the env vars it tired to set and avoid this ad-hoc?

.circleci/config.yml
sausagee: TBH, piping true to another cmd then or-ing true feels weird.  

If `tr` fails, $CHANGED_DOCKER_FILE_LIST would become `"true"`.  I think the hadolint below will fail.

docker/mint/Dockerfile
sausagee: I'd suggest move the mint's Dockerfile change to a separate commit or PR.

sausagee (commented):
Confused by the title and summary - is this a PR to re-enable Hadolint?  Or is it a canary for testing re-enabling Hadolint?

.circleci/config.yml
rexhoffman: true isn't getting piped -- looks like || is taking precedence, will add parens to make it clear.    the second || true probably isn't needed

rexhoffman: also it's grep that is failing if it doesn't find anything,  like so... https://stackoverflow.com/questions/42251386/linux-grep-return-code

sausagee (approved):
LGTM

PR #5947, opened 2020-09-09
Push debian base images to ecr

rexhoffman:
## Motivation

This handles the push of debian base images to ecr.  

Yes

## Test Plan

CI

## Related PRs

None

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -752,6 +752,37 @@ jobs:
                 command: |
                   scripts/dockerhub_prune.sh -u "${DOCKERHUB_USERNAME}"" -p ${DOCKERHUB_PASSWORD}" -x
 
+  docker-update-base-images:
+    executor: vm-medium-executor
+    steps:
+      - checkout
+      - run:
+          name: Define relevant files
+          command: |
+            grep --include=\*Dockerfile -r '.' -e 'FROM debian:' | sed 's/:.*//g' | sed 's/\.\///g' | sort | uniq > /tmp/ci_docker_files
+      - halt_if_no_relevant_files_changed_since_last_evaluation:
+          cache_key_part: docker_base_image
+          file_of_relevant_git_files: /tmp/ci_docker_files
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -764,6 +764,38 @@ jobs:
...
+    steps:
+      - checkout
+      - run:
+          name: Define relevant files
+          command: |
+            grep --include=\*Dockerfile -r '.' -e 'FROM debian:' | sed 's/:.*//g' | sed 's/\.\///g' | sort | uniq > /tmp/ci_docker_files
+      - halt_if_no_relevant_files_changed_since_last_evaluation:
+          cache_key_part: docker_base_image
+          file_of_relevant_git_files: /tmp/ci_docker_files
+      - setup_docker_signing
```
sausagee, line 778: We can skip this step, as we are not DCT signing below.

.circleci/config.yml
```diff
@@ -925,6 +957,13 @@ workflows:
               only:
                 - /^test-[\d|.]+$/
                 - /^release-[\d|.]+$/
+      - docker-update-base-images:
+          context: docker
+          filters:
+            branches:
+              only:
+                - /^test-[\d|.]+$/
+                - /^release-[\d|.]+$/
```
sausagee, line 966: Should we also run this as nightly scheduler and on branches (test-*, release-* and master)?  When we change the base image in master/trunk, I supposed we want it cached in ECR too.

sausagee (commented):
The workflow looks mostly fine.  Will stamp after comments are resolved.

.circleci/config.yml
rexhoffman: moving to master so it works with your pr.    it will only build on change to docker files that build on top of debian.   Without any way to reliably know what already been processed in the past, figured this was the least error prone approach and should still happen less frequently than one build a day.

sausagee (approved):

PR #5949, opened 2020-09-09
[language] Fix dependency order in stdlib

dariorussi:
Make the order of dependencies in the `BTreeMap` correct.
`1_xx` comes after `10_xxx`, etc., and thus the dependencies order
was not honored.


dariorussi:
the `BTreeMap` returned by `build_stdlib()` is used in a `stdlib::main()` which I am not sure I know (who is using it and for what).
I think it should be fine and a CI/test run should tell us, but hope someone can tell me if this is causing issues.
@tzakian, @sblackshear, @matbd?

language/stdlib/src/lib.rs
```diff
@@ -90,7 +90,7 @@ pub fn build_stdlib() -> BTreeMap<String, CompiledModule> {
                     .expect("stdlib module dependency failed to verify");
                 // Tag each module with its index in the module dependency order. Needed for
                 // when they are deserialized and verified later on.
-                modules.insert(format!("{}_{}", i, name), module);
+                modules.insert(format!("{:0>3}_{}", i, name), module);
```
sblackshear, line 93: Do we still need these numerical prefixes @tzakian?

dariorussi: this is also breaking `test_that_generated_file_are_up_to_date_in_git` @matbd

dariorussi: and that's because all file names are different.... I suppose I should just add all the new file name in `stdlib/compiled/stdlib` right?

tzakian: Sadly we do still need the numerical prefixes since the module verification needs to be done in dependency order. 

I'm not sure if this PR is needed though. In particular if you look at how this prefix is used in sorting we don't use it as a string, but parse this numerical prefix as a u64 and sort based on that, see `language/compiled/src/lib.rs` lines 51-55 for where this is done. 

If we do want to keep this change to the numeric prefix, the `sort_by` key code referenced above should be able to be removed.

dariorussi: well the problem is that this API is called to fetch the module to publish (e.g. genesis) and the order is wrong.
Now I don't want to go into why it worked and how a different PR that I am doing unveiled the issue, but the bottom line is that if you get just the modules out of the `BTreeMap`, and you try to publish them on a fresh instance of a VM, you **will** fail because dependency order is bogus. 
As an example, if you look under `stdlib/compiled/stdlib/*` `10_AccountLimits.mv` comes before `2_Errors.mv` and if you loop through the `build_stdlib().values()` publishing (which is the only thing you can do) you **will** fail.
Moreover if you look at the comment you will find out that is not what the code was doing.
Hope I am making some sense here.....

dariorussi: > In particular if you look at how this prefix is used in sorting we don't use it as a string, but parse this numerical prefix as a u64 and sort based on that, see language/compiled/src/lib.rs lines 51-55 for where this is done.

that is not what all code is doing but it is a good reference (thanks) and we could require all callers to do the sorting that way first (though it seems a bit of a stretch to me, but...)

dariorussi: so thinking some more :) why do we have the map in the first place?

sblackshear: It might be useful to provide an API for topologically sorting a `Vec<CompiledModule>` based on dependencies instead of encoding the order in the file names. That way, we know how to sort/verify a random bundle of modules regardless of how they were saved.

I ran into this problem + another one in writing the CLI code. The second problem is that if you have the name of a module in hand and want to find its bytecode on disk, the numerical prefixes force you to iterate through the directory instead of just building a path...

dariorussi: > the numerical prefixes force you to iterate through the directory instead of just building a path...
that is funny....

tnowacki: What if you have more than 1000 modules! :p 
(for real this is fine for now, but it would nice to automagically pad as needed)

tnowacki: > It might be useful to provide an API for topologically sorting a Vec<CompiledModule> based on dependencies instead of encoding the order in the file names.

I originally encoded them in file names because I really didn't anticipate people calling this directly in Rust code. 

As far as providing the sorting, we could do `(BTreeMap<String, usize>, Vec<CompiledModule>)` if that makes things easier?

dariorussi: I think we are going with this for now and then we'll revisit as needed. 
I also think that we could (and probably should) have 2 API one that returns just a vector of `CompiledModule` (in the order returned by the compiler) and one that gives you the file name too. For what is worth the one returning the file name as well could be as simple as a pair `(CompiledModule, String)`

dariorussi: also for @tnowacki this was the reason for the failures I was seeing. Basically the test framework was using the `FRESH_MOVELANG_STDLIB` which returned a vector of modules with an order that was not "dependencies based"

dariorussi:
to summarize this fixes the problem that `FRESH_MOVELANG_STDLIB` was not returning the modules in dependency order.
@tzakian it turns out we cannot remove your ordering based on name because the directory iterator does not return the modules in alphabetic order. So we need to keep that solution.
@sblackshear I have not done the split of the string in the compatibility code because it did not seem necessary and it would be a bit of code to make sure it was solid (not based on too much knowledge of the structure of that directory and the file names). Please let me know if you want me to do that work anyway

sblackshear:
> @sblackshear I have not done the split of the string in the compatibility code because it did not seem necessary and it would be a bit of code to make sure it was solid (not based on too much knowledge of the structure of that directory and the file names). Please let me know if you want me to do that work anyway

No need to do that here, I can tackle it later if it's needed. Thanks!

dariorussi:
https://github.com/libra/libra/pull/5960 (currently a draft and with a copy of these changes in) depends on this.
So if someone could review and send feedback or accept that would be priceless....

@matbd do you see problems with the file renaming in this PR?

sblackshear (approved):

language/stdlib/compiled/src/lib.rs
```diff
@@ -36,23 +36,19 @@ pub const ERROR_DESCRIPTIONS: &[u8] =
 // specified either by the MOVE_NO_USE_COMPILED env var, or by passing the "StdLibOptions::Fresh"
 // option to `stdlib_modules`.
 static COMPILED_MOVELANG_STDLIB: Lazy<Vec<CompiledModule>> = Lazy::new(|| {
-    let mut modules: Vec<(String, CompiledModule)> = COMPILED_STDLIB_DIR
+    let mut modules: Vec<(&str, CompiledModule)> = COMPILED_STDLIB_DIR
```
tzakian, line 39: Could you collect this into a `BTreeMap` and then no sorting is needed later on? i.e.
```rust
let modules: BTreeMap<&str, CompiledModule> = <same code you have here>;
for (_, module) in modules.into_iter() { .... }
...
```

tzakian (approved):

vgao1996:
It seems like #5972 also depends on this. But I wonder how was it possible for the move functional tests able to run fine with this bug?

dariorussi:
> It seems like #5972 also depends on this. But I wonder how was it possible for the move functional tests able to run fine with this bug?

it's because usually the modules are published already in the VM and so they were all available.
Bottom line, if the modules are published in store (whether a fake or not) things are fine. If the modules are already published in the VM (like the genesis does) things will work too. So generally things used to work but a PR I was doing broke things because of the problem I mentioned (hoping it made sense)

vgao1996:
Ah that makes sense... So basically what happened was in my old implementation for genesis generation, the loader was pre-populated with all the stdlib modules (which were injected into storage directly), and then I used the same loader to publish the modules. I should have invalidated the loader and start a new one since I'm effectively removing some modules from the environment. My new implementation didn't have this problem in the first place so it got exposed to this issue.

vgao1996:
So that makes me wonder: is there a way we can ensure the Move VM instance always gets invalidated whenever code in storage changes?

PR #5950, opened 2020-09-09
[move-prover] Added more specs for the capabilities in Libra Framework

junkil-park:
- Simplified the function signature of `Libra::publish_burn_capability`
  - Removed the redundant argument (e.g., account) from the argument list

- Added more governance-related specs regarding capabilities such as MintCap, BurnCap, Preburn, KeyRotationCap, and WithdrawCap

## Motivation

To add more specs for the capabilities in Libra Framework

Yes

## Test Plan

cargo test


language/stdlib/modules/DualAttestation.move
```diff
@@ -130,8 +130,11 @@ module DualAttestation {
...
         let sender = Signer::spec_address_of(account);
-        include LibraTimestamp::AbortsIfNoTime;
+
+        /// Must abort if the account does not have the resource Credential [B25].
         include AbortsIfNoCredential{addr: sender};
+
+        include LibraTimestamp::AbortsIfNoTime;
         ensures global<Credential>(sender).base_url == new_url;
 
         /// The sender can only rotates its own base url [B25].
```
DavidLDill, line 140: rotates --> rotate (let's fix typos in comments when convenient).

language/stdlib/modules/Libra.move
```diff
@@ -1158,73 +1126,177 @@ module Libra {
...
     spec module {
         /// Only mint functions can increase the total amount of currency [B11].
         apply TotalValueNotIncrease<CoinType> to *<CoinType>
             except mint<CoinType>, mint_with_capability<CoinType>;
+
+        /// In order to successfully call `mint` and `mint_with_capability`, MintCapability is
+        /// required. MintCapability must be only granted to a TreasuryCompliance account [B11].
+        /// Only `register_SCS_currency` creates MintCapability, which must abort if the account
+        /// does not have the TreasuryCompliance role [B17].
+        apply PreserveMintCapAbsence<CoinType> to *<CoinType> except register_SCS_currency<CoinType>;
```
DavidLDill, line 1158: I think there's a soundness issue here, unless something has changed recently.  Suppose some other module defines a function that calls register_currency, gets a capability, and publishes it somewhere.  Then, this spec could be violated by a function that PreserveMintAbsence is not applied to.

With global invariants, we can put all such modules on the command line (Coin1, Coin2, LBR? in this case).  But I don't think that works with "apply".

language/stdlib/modules/Libra.move
```diff
@@ -1158,73 +1126,177 @@ module Libra {
...
+
+        /// The permission "MintCurrency" is unique per currency [C11].
+        /// At most one address has a mint capability for SCS CoinType
+        invariant [global, isolated]
+            forall coin_type: type where spec_is_SCS_currency<coin_type>():
+                forall addr1: address, addr2: address
+                     where exists<MintCapability<coin_type>>(addr1) && exists<MintCapability<coin_type>>(addr2):
+                          addr1 == addr2;
+        /// If an address has a mint capability, it is an SCS currency.
+        invariant [global]
```
DavidLDill, line 1179: Need to include modules that have this as a dependency to make sure this holds?  See the comment above about "register_currency"

language/stdlib/modules/Libra.move
shazqadeer: A lot of these schema-based postconditions seem to be specifications which are (more nicely and efficiently, in my opinion) expressed using two-state invariants (specified using the attribute "update").  The guard in the two-state invariant would contain the access-control predicate saying something like "signer has this or that role".  Is there a reason why we prefer these apply-style postconditions rather than two-state invariants?  Could it be the case that "signer has this/that role" is not expressible in our spec language?  

@wrwg, @sblackshear .

junkil-park: > Could it be the case that "signer has this/that role" is not expressible in our spec language?
@shazqadeer, yes, `account: &signer` is a function parameter, and I believe that our spec lang currently does not support using function parameters in invariants. Also, some Move functions does not have such an "account" parameter, and some other Move functions have one with different names (e.g., lr_account, tc_account).

sblackshear: I share David and Shaz's concerns about this `apply` + `Preserve`-based specification methodology. I don't think this is expressing the property of interest.

DavidLDill (approved):

language/stdlib/modules/DualAttestation.move
junkil-park: Done.

language/stdlib/modules/Libra.move
junkil-park: I added a TODO for this in the code. It seems that we don't have a quick solution for this currently.

language/stdlib/modules/DualAttestation.move
```diff
@@ -130,8 +130,11 @@ module DualAttestation {
     }
     spec fun rotate_base_url {
         let sender = Signer::spec_address_of(account);
-        include LibraTimestamp::AbortsIfNoTime;
+
+        /// Must abort if the account does not have the resource Credential [B25].
```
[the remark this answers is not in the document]
junkil-park: Yes, the cell numbers should be gone eventually, but I'd like to keep them until we find a way to utilize them in the document.

language/stdlib/modules/Libra.move
junkil-park: The same as above.

PR #5952, opened 2020-09-09
[Secure Storage] Use filesystem lock to protect concurrent accesses to on-disk storage from different processes.

JoshLind:
## Motivation

This PR updates the on-disk storage backend to use filesystem locks to allow concurrent accesses to the same on-disk storage backend file (from multiple different processes). It does this using the named-lock crate. Moreover, it updates the key manager smoke tests and adds a new smoke test to allow multiple consensus key rotations (which couldn't previously be done because in the old deployment model LSR and the key manager have different backend files!). 

This PR offers three commits to allow smoke tests to use a single secure storage file:
1. Update the on-disk storage to use lock files for concurrency (around read and write).
2. Clean up the key manager smoke test.
3. Refactor the config builder to use only a single secure storage file and create a smoke test to perform multiple consensus key rotations to ensure that the shared backend file works.

Yes.

## Test Plan

All local tests pass, including smoke tests.

## Related PRs

None.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3110,6 +3110,7 @@ dependencies = [
  "libra-types 0.1.0",
  "libra-vault-client 0.1.0",
  "libra-workspace-hack 0.1.0",
+ "named-lock 0.1.1 (registry+https://github.com/rust-lang/crates.io-index)",
  "rand 0.7.3 (registry+https://github.com/rust-lang/crates.io-index)",
  "serde 1.0.115 (registry+https://github.com/rust-lang/crates.io-index)",
  "serde_json 1.0.57 (registry+https://github.com/rust-lang/crates.io-index)",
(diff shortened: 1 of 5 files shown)


JoshLind:
@davidiw, this is ready for review 😄

davidiw:
I'm a little uncomfortable with this PR because it adds dependencies to the TCB for a test only purpose. It also doesn't seem to simplify any tests. Perhaps we could better motivate this? Add it as a wrapper type like we did with namespace but make it only accessible in testing or ... something else?

JoshLind:
Thanks, David! I agree.. the benefits of this PR are more what's happening in the background:
1. It means we no longer have to have separate storage files for each component in smoke tests. This requirement caused a bunch of annoying problems when writing genesis related code and tests.. It also makes debugging smoke tests easier as we can view the state of storage at a failure.
2. It allows us to write smoke tests that cross components (e.g., the key manager can rotate the consensus key several times and LSR can continue to operate). With the current model, the key manager can't do this. It doesn't have safe access to the LSR backend.

Regarding adding dependencies, I agree that it's not great. Let me take a look tomorrow to see if there's any easy way I can improve this..

JoshLind:
Thanks, @davidiw! 😄 As discussed offline, I'll close this out as it should hopefully not be required once we move to a better execution model for tests. Overall it's good to see that this works though!

PR #5954, opened 2020-09-09
[fuzzing] remove dead code

mimoo:


testsuite/libra-fuzzer/src/fuzz_targets/network_noise.rs
```diff
@@ -1,59 +0,0 @@
...
-    fn description(&self) -> &'static str {
-        "Network Noise crate stream"
-    }
-
-    fn generate(&self, _idx: usize, gen: &mut ValueGenerator) -> Option<Vec<u8>> {
-        Some(generate_corpus(gen))
-    }
-
-    fn fuzz(&self, data: &[u8]) {
-        fuzz_post_handshake(data);
```
JoshLind, line 57: Does it make sense to remove these helper functions? e.g., this is defined in the network/src/noise/fuzzing.rs.. Or are you thinking we're going to re-use these at some point in the future?

JoshLind (approved):
Thanks @mimoo! LGTM.

To clarify: what's the plan for networking noise fuzzing?

testsuite/libra-fuzzer/src/fuzz_targets/network_noise.rs
mimoo: the noise fuzz stuff has been moved to network.rs :) so it's still here, just forgot to get rid of that file when I moved stuff last time

PR #5958, opened 2020-09-09
[vm] add a comment about running the failure epilogue

bob-wilson:
## Motivation

Explain the role of the failure epilogue. This was not obvious to me and I only figured it out from looking at some tests that showed the desired behavior.

Yes

## Test Plan

N/A -- Only changes a comment

### DIFF
--- a/language/libra-vm/src/libra_transaction_executor.rs
+++ b/language/libra-vm/src/libra_transaction_executor.rs
@@ -89,6 +89,12 @@ impl LibraVM {
         let mut session = self.0.new_session(remote_cache);
         match TransactionStatus::from(error_code.clone()) {
             TransactionStatus::Keep(status) => {
+                // The transaction should be charged for gas, so run the epilogue to do that.
+                // This is running in a new session that drops any side effects from the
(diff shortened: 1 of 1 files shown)


language/libra-vm/src/libra_transaction_executor.rs
```diff
@@ -89,6 +89,11 @@ impl LibraVM {
         let mut session = self.0.new_session(remote_cache);
         match TransactionStatus::from(error_code.clone()) {
             TransactionStatus::Keep(status) => {
+                // The transaction should be charged for gas, so run the epilogue to do that.
+                // This is running in a new session that drops any side effects from the
+                // attempted transaction, so even if the previous failure occurred while
```
sblackshear, line 94:
```suggestion
                // attempted transaction (e.g., spending funds that were needed to pay for gas), so even if the previous failure occurred while
```

sblackshear (approved):
Good call!

bob-wilson:
I see that Todd added a similar comment to the Move code for the epilogue itself in #5876 but it doesn't hurt to explain in both places.

PR #5960, opened 2020-09-10
[language] Detect and error on circular dependencies

dariorussi:
Make the loader check for circular dependencies while loading.
That seems important given modules are only pushed through WriteSets
and thus never going through the verifier. If a circular dependency
was introduced, we would enter an infinite loop.
So it's a protection against an unlikely but horrible scenario.
It also fixes a problem with publishing that was loading and caching
all dependencies except the published module, though it was intended
to load nothing. Now publishing a module only verifies the dependencies
of the published module and nothing else, it also does not load any module. 
That hopefully simplifies the code a bit


dariorussi:
this PR only adds 22 lines to the loader which are related to separating the publishing verification path from the rest. In that sense it actually removes some code and it may make things slightly simpler.
Having said so, it's a change.
I would venture that it is justified by the fact the a circular dependency may halt the system real bad, though it would save us only if that was not in the common path. We would be messed up anyway otherwise (difference being an error rather than a stack overflow, and who knows what is better).
This also does not load dependencies when publishing which is correct, and good mostly for testing given there is no publishing for now. I don't think it makes sense to split to 2 issues. We either commit both or none.
It may be worth to point out we run with the previous code for the past month and not "forever", so in a sense this is not changing ancient and working behavior.

I am fine not committing this and let it go, if people feel uncomfortable with the fix.
So please speak up and express an opinion....

language/move-vm/runtime/src/unit_tests/loader_circular.rs
```diff
@@ -0,0 +1,352 @@
...
+    let modules = compile_file(&[0; 16]);
+    adapter.publish_modules(modules, "C", "A");
+    assert_eq!(
+        adapter
+            .call_function(
+                &ModuleId::new(WORKING_ACCOUNT, Identifier::new("A").unwrap()),
+                Identifier::new("entry_a").unwrap().as_ident_str(),
+            )
+            .unwrap_err(),
+        VMStatus::Error(StatusCode::UNEXPECTED_VERIFIER_ERROR),
```
dariorussi, line 199: this is reporting a `CIRCULAR_DEPENDENCY` error, but it's obfuscated by the need for `error_specializer` in `execute_function`. I suppose we can rewrite those with `execute_script` which a bit lame but maybe I should...

language/move-vm/runtime/src/loader.rs
```diff
@@ -561,29 +563,60 @@ impl Loader {
     // A module to be published must be loadable.
     // This step performs all verification steps to load the module without loading it.
     // The module is not added to the code cache. It is simply published to the data cache.
-    // See `verify_script()` for script verification steps.
-    pub(crate) fn verify_module_verify_no_missing_dependencies(
+    // (See `verify_script()` for script verification steps(
```
bob-wilson, line 566: Seems like this comment should end with a close paren

language/move-vm/runtime/src/unit_tests/loader_circular.rs
bob-wilson: I don't follow.... The execute_function calls in loader_tests.rs are using an identity function for error_specializer. The "obfuscation" seems to be coming from expect_no_verification_errors() in move-vm/runtime/src/loader.rs. What are you suggesting to avoid that?

bob-wilson (commented):
I'm not very familiar with the loader code, so I'm not going to approve this. Anyway, we should get more people to look at this, since it is a non-trivial change coming late in the schedule. I left a few comments. Overall it seems like this is probably worth doing, but if anyone has concerns about risk from the code change, we could get by without this for now because we don't have open publishing yet.

language/move-vm/runtime/src/loader.rs
dariorussi: parser people....

language/move-vm/runtime/src/unit_tests/loader_circular.rs
dariorussi: I suppose strictly speaking you are right and my description was not exact.
I guess `execute_function` is returning a `VMStatus` whereas the other functions are returning a `VMError` which has all the info.
So strictly speaking is not the `error_specializer` that is causing that, but having that (the `error_specializer`) is causing us to convert at that level (`into_vm_status` - https://github.com/libra/libra/blob/master/language/move-vm/runtime/src/session.rs#L42) and not being able to see the true error.
Or at least that is my understanding, let me know, if that is not the case (though whether that is the case or not is rather immaterial at this point)

dariorussi:
> I'm not very familiar with the loader code, so I'm not going to approve this. Anyway, we should get more people to look at this, since it is a non-trivial change coming late in the schedule. I left a few comments. Overall it seems like this is probably worth doing, but if anyone has concerns about risk from the code change, we could get by without this for now because we don't have open publishing yet.

to be clear, the concern for me is more related to the fact we mostly publish code via WriteSet and do not run any verification over that. But having said so, I am with you overall....

language/move-vm/runtime/src/unit_tests/loader_circular.rs
bob-wilson: Ah, OK, I get it now. Thanks for the explanation

dariorussi:
@tnowacki you'd be a good candidate to look at this. I'd appreciate if you could

language/move-core/types/src/vm_status.rs
```diff
@@ -564,6 +564,7 @@ pub enum StatusCode {
     // The sender is trying to publish a module named `M`, but the sender's account already
     // contains a module with this name.
     DUPLICATE_MODULE_NAME = 1095,
+    CIRCULAR_DEPENDENCY = 1096,
```
tnowacki, line 567: This should be an invariant violation error maybe? It's definitely not a verifier error thats for sure. 

The only way this can occur currently is with a writeset. Ideally long term, the writeset that causes this would be rejected. We might be able to do that today if we make this an invariant violation error (or something) and then try to load the module after the writeset is complete

language/move-vm/runtime/src/loader.rs
```diff
@@ -734,55 +768,44 @@ impl Loader {
...
-        &self,
-        id: &ModuleId,
-        data_store: &mut impl DataStore,
-    ) -> VMResult<Arc<Module>> {
-        self.load_module(id, data_store, false)
-    }
-
-    // Returns a verifier error if the module does not exist
-    fn load_dependencies_verify_no_missing_dependencies(
+    fn load_dependencies(
```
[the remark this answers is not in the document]
dariorussi: yeah I know why you did it, but one of the functions was not needed any longer and this was a better way to go.
I am not too sure about the name being any helpful in general,  particularly for functions inside a module. Or at least they did not help me at all.
Also I was supposed to mark with a comment the flag passes in and did not do that, you are right that would have helped.
And overall I may have introduced a bug in this refactor so, there... maybe it was not as straightforward as I thought...

language/move-vm/runtime/src/loader.rs
```diff
@@ -537,7 +537,9 @@ impl Loader {
         ty_args: &[TypeTag],
         data_store: &mut impl DataStore,
     ) -> VMResult<(Arc<Function>, Vec<Type>)> {
-        self.load_module_expect_no_missing_dependencies(module_id, data_store)?;
+        let mut visited_deps = HashSet::new();
+        visited_deps.insert(module_id.clone());
+        self.load_module(module_id, data_store, &mut visited_deps, false)?;
```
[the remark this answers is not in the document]
dariorussi: I don't understand what moving `visited_deps` to `self` would do, besides the fact I would never do that because it would be promoting something that must live on the stack to global scope.
We would also have to have an API in the caller to clean up the `visited_deps` which amounts to the same problem.
I think this is a caller responsibility and it should be clearly in your face

language/move-core/types/src/vm_status.rs
dariorussi: not sure about this. I hear you and the difficulties but don't know.
I think circular dependencies are definitely a verification step, whether that lives in the verifier directory is a different and orthogonal problem. But certainly something worth to think about.
The reason all of this seemed reasonable to me is because we keep talking and thinking about updating module via publishing and in that case we need this check and it has nothing to do with `WriteSet`.
But it's certainly true `WriteSet` is the only way to change code now...
so don't know.... I guess it's something to think about...

dariorussi:
I am going to abandon this PR given the feedback and the speed is moving at.
Let me know if anybody thinks otherwise, though I am not sure you'd be able to convince me to put it in.
Thanks and sorry for the churn!

bob-wilson:
Close it?

PR #5961, opened 2020-09-10
[json-rpc] standard jsonrpc response object

xli:
Small step to create new json rps response for client, current client does not handle response body chain id, version and timestamp, we need decode them in client for chain id and server side ledger version and timestamp.

This diff adds new JsonRpcResponse to json rpc types, and use the type for server response.

Existing unit test should cover the response data validation.

### DIFF
--- a/json-rpc/types/src/response.rs
+++ b/json-rpc/types/src/response.rs
@@ -9,3 +9,53 @@ pub struct JsonRpcErrorResponse {
     pub jsonrpc: String,
     pub error: JsonRpcError,
 }
+
+#[derive(Clone, Serialize, Deserialize, Debug)]
+pub struct JsonRpcResponse {
+    pub libra_chain_id: u8,
+    pub libra_ledger_version: u64,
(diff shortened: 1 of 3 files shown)


json-rpc/types/src/response.rs
```diff
@@ -9,3 +9,53 @@ pub struct JsonRpcErrorResponse {
...
+    pub fn new(
+        chain_id: libra_types::chain_id::ChainId,
+        libra_ledger_version: u64,
+        libra_ledger_timestampusec: u64,
+    ) -> Self {
+        JsonRpcResponse {
+            libra_chain_id: chain_id.id(),
+            libra_ledger_version,
+            libra_ledger_timestampusec,
+            jsonrpc: "2.0".to_string(),
```
sunmilee, line 38: couldn't this also be passed in?

sunmilee (approved):

json-rpc/types/src/response.rs
xli: never going to be changed.

PR #5951, opened 2020-09-09
[CI] enable DockerHub pub on master

sausagee:
## Motivation
devnet will cut over to pull images from DockerHub and start testing them.

## Test Plan
CI

PR #5962, opened 2020-09-10
[GHA][2/x] migrate post-land integration test

sausagee:
## Motivation
We are moving the integratoin workflow, aka continuous-push, from
Novi's other repo to Libra, for closer integration with DockerHub and
visibility.

The integration test workflow connects to the DockerHub build and
publish pipeline in CI (#5951)

## Test Plan
1. There is no new commit in master. --> It should result a no-op.
Canary with 6e93070 in https://github.com/libra/libra/runs/1094560577?check_suite_focus=true
image

2. There is a commit in master and images pushed to ECR. Cluster Test is triggered but failed. --> It should post alert.
Canary with  3d167c6 in https://github.com/libra/libra/runs/1094643124?check_suite_focus=true
The canary has an override to use images built by Novi's CI pipeline instead of DockerHub. Once #5951 is landed, it will tap into Libra's CI pipeline. 
image
Verified the push alert is received.

3. Same as setup in Test 2 above, except Cluster Test runs to completion. --> It should retag the images. 
Canary with  be81c2d in https://github.com/libra/libra/runs/1098908257?check_suite_focus=true

The canary has an override fake the HEAD to 593805e and another override to retag images with `sausage` instead of `master`. 
image

Once canary completes, verified images are retagged in ECR.

## Deployment plan
Once landed, this GHA based workflow will be shadowing the current Continuous Push for a day, with at least 3 successful e2e pushes.  After that, it will replace the current one.

### DIFF
--- a/.github/workflows/integration-test.yml
+++ b/.github/workflows/integration-test.yml
@@ -1,0 +1,128 @@
+---
+name: Integration Test
+
+on:
+  push:
+    branches:
+      - master
+      - canary
+  schedule:
+    - cron:  '15 */4 * * *'
+
+
+jobs:
+  run-cluster-test:
+    name: Run the pre-release suite of Cluster Test
+    runs-on: self-hosted
+    # The pre-release suite run time varies 1~1.5 hr.
+    timeout-minutes: 120
+    steps:
+      - uses: actions/checkout@v1
+      - name: setup_env
+        run: |
+          HEAD_GIT_REV=$(git rev-parse --short=8 HEAD)
+          echo "::set-env name=HEAD_GIT_REV::$HEAD_GIT_REV"
+      - name: find_last_tested_commit
+        # Find the commit hash of the last tested tag in ECR
+        env:
+          AWS_REGION: us-west-2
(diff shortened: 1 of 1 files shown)


.github/workflows/integration-test.yml
```diff
@@ -0,0 +1,128 @@
...
+            if [[ -z $(aws ecr describe-images --repository-name $ecr_repo \
+              --region $AWS_REGION \
+              --image-ids=imageTag=$new_tag) ]]; then
+              echo "Failed to re-tag $ecr_repo:$old_tag with $new_tag"
+              exit 123
+            fi
+          }
+
+          for repo in $REPOS; do
+            retag_image $repo "master_$HEAD_GIT_REV" "master"
```
sausagee, line 127: Going to change this to `shadow` so that the GHA workflow can shadow the current workflow.

rexhoffman (approved):

sausagee:
Will land this along with https://github.com/libra/libra/pull/5951 tomorrow morning MPK time.

PR #5963, opened 2020-09-10
[docker] Add initial documentation / framework for public full node support

davidiw:
Checkpointing this work as we make progress on opening up our full node access.
This docker compose and node config has been verified to work on a swarm
instances. Will update it once we have testnet publicly available.

### DIFF
--- a/config/src/config/mod.rs
+++ b/config/src/config/mod.rs
@@ -434,6 +434,11 @@ mod test {
         NodeConfig::default_for_validator();
         NodeConfig::default_for_validator_full_node();
 
+        let docker_public_full_node =
+            std::include_str!("../../../docker/compose/public_full_node/public_full_node.yaml");
+        // Only verify it is in the correct format as the values cannot be loaded for this config
+        NodeConfig::parse(docker_public_full_node).unwrap();
+
         let contents = std::include_str!("test_data/safety_rules.yaml");
         SafetyRulesConfig::parse(&contents)
             .unwrap_or_else(|e| panic!("Error in safety_rules.yaml: {}", e));
--- a/config/src/config/test_data/public_full_node.yaml
+++ b/config/src/config/test_data/public_full_node.yaml
@@ -13,14 +13,12 @@ execution:
     # corresponding genesis, the file location should be an empty path.
     genesis_file_location: "/full/path/to/genesis"
 
-
 full_node_networks:
     - discovery_method: "onchain"
       # The network must have a listen address to specify protocols. This runs it locally to
       # prevent remote, incoming connections.
       listen_address: "/ip4/127.0.0.1/tcp/6180"
       network_id: "public"
-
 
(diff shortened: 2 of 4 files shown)


docker/compose/public_full_node/docker-compose.yml
```diff
@@ -0,0 +1,44 @@
...
+#   * Directions on the correct image
+#   * Place to obtain waypoints and genesis.blob
+#   * Any modifications / specifying seed_addrs
+#
+# Additional information:
+# * If you use this compose for different Libra Networks, you will need remove the db volume first.
+version: "3.8"
+services:
+    fullnode:
+        image: libra/validator:test-2_8be5f4fa
```
davidiw, line 23: Probably should add more comments around here, if someone downloads an up to date master, they should be able to just pull the default or master. If they modify it, they'll need to use the docker file. Will save that for later.

docker/compose/public_full_node/docker-compose.yml
```diff
@@ -0,0 +1,48 @@
+# This compose file defines a Public Fullnode docker compose wrapper around libra-node. Currently
+# this can only connect to a local swarm instance, but work is in progress to enable more use
+# cases.
+#
+# In order to use, place a copy of the proper genesis.blob and waypoint.txt in this directory.
```
rexhoffman, line 5: Can these be discovered from a running network, and enforced if set?   Not for this pr, just a general observation.

docker/compose/public_full_node/docker-compose.yml
rexhoffman: Hmm test-2 is just a throw away release I built when changing/testing circleci configs/scripts.   I wouldn't use it here as it will be deleted in 2 weeks.    Perhaps use a release?

rexhoffman: A wrapping script could discover the version of the network, provided there is a discovery mechanism.

davidiw: That's fair. I am oppose to adding scripts here. Scripts only run on Linux and Mac. It doesn't eliminate the need for documentation and only adds more since we'll have to document the scripts usage as well. If we're going to introduce scripts, we should eliminate docker compose and consider docker altogether.

With that being said, if we end up with something so CI can test it, sure, but that will remain undocumented and not a supported tool.

docker/compose/public_full_node/docker-compose.yml
davidiw: If we had engineered this concept in, unfortunately, we stumbled into this too late :/

davidiw:
Are there any issues that folks see with this? :D

davidiw:
@christinacelee, mind taking a look and stamping it if it seems sensible enough?

bmwill (approved):

PR #5964, opened 2020-09-10
[fuzzing] adding two fuzzers to find incorrect deserialization

mimoo:
This commit adds two fuzzers to ensure we can't deserialize different bytestring to the same value:

* MutatedSignedTransaction. The fuzzer will mutate a fixed SignedTransaction (generated via a fixed-seed) and ensure that it never deserializes to the same fixed SignedTransaction.
* TwoSignedTransactions. The fuzzer will mutate two copies of the same serialized SignedTransaction, and ensure that it never finds two different mutations that deserializes to the same SignedTransaction.
 This commit does not belong to any branch on this repository.

### DIFF
--- a/testsuite/libra-fuzzer/src/fuzz_targets/transaction.rs
+++ b/testsuite/libra-fuzzer/src/fuzz_targets/transaction.rs
@@ -8,7 +8,13 @@ use language_e2e_tests::account_universe::{
 };
 use libra_proptest_helpers::ValueGenerator;
 use libra_types::transaction::SignedTransaction;
-use proptest::{collection::vec, prelude::*};
+use once_cell::sync::Lazy;
+use proptest::{
+    collection::vec,
+    prelude::*,
+    strategy::{Strategy, ValueTree},
+    test_runner::{self, RngAlgorithm, TestRunner},
+};
 
 #[derive(Clone, Debug, Default)]
 pub struct LanguageTransactionExecution;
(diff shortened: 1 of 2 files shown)


matbd:
No test plan? :(
In particular, it would be interesting to know if you can catch manually-added bugs in this way.

mimoo:
> No test plan? :(

besides that the two fuzzers work not really : o

> In particular, it would be interesting to know if you can catch manually-added bugs in this way.

how would you manually add a bug though? In your example "If lcs is modified e.g. to accept >1u8 values as a True boolean, does the fuzzer find it?" there is no bool in `SignedTransaction` so we wouldn't find that (since we're using that struct only here). But if there was I would expect the fuzzer to find it pretty quickly : o

testsuite/libra-fuzzer/src/fuzz_targets/transaction.rs
```diff
@@ -54,3 +60,82 @@ impl FuzzTargetImpl for SignedTransactionTarget {
...
+static SERIALIZED_SIGNED_TXN: Lazy<Vec<u8>> =
+    Lazy::new(|| lcs::to_bytes(&SIGNED_TXN.clone()).expect("serialization should work"));
+
+impl FuzzTargetImpl for MutatedSignedTransaction {
+    fn description(&self) -> &'static str {
+        "SignedTransaction (LCS serialized -> mutation -> deserializer)"
+    }
+
+    fn generate(&self, _idx: usize, _gen: &mut ValueGenerator) -> Option<Vec<u8>> {
+        println!("{:?}", SIGNED_TXN);
```
davidiw, line 86: nit

testsuite/libra-fuzzer/src/fuzz_targets/transaction.rs
```diff
@@ -54,3 +60,82 @@ impl FuzzTargetImpl for SignedTransactionTarget {
...
+        // always return the same serialized signed transaction
```
davidiw, line 87: this comment isn't that useful since it says what the code is doing in more words, so maybe we could state why we always returning the same signed transaction?

testsuite/libra-fuzzer/src/fuzz_targets/transaction.rs
```diff
@@ -54,3 +60,82 @@ impl FuzzTargetImpl for SignedTransactionTarget {
...
+        Some(SERIALIZED_SIGNED_TXN.clone())
+    }
+
+    fn fuzz(&self, data: &[u8]) {
+        // only accept different serialization
```
davidiw, line 92: same as above, comment matches exactly what the code does but says so in more words but doesn't give the why

testsuite/libra-fuzzer/src/fuzz_targets/transaction.rs
```diff
@@ -54,3 +60,82 @@ impl FuzzTargetImpl for SignedTransactionTarget {
...
+        if data == SERIALIZED_SIGNED_TXN.as_slice() {
+            return;
+        }
+
+        if let Ok(signed_txn) = lcs::from_bytes::<SignedTransaction>(&data) {
+            // make sure that a different serialization doesn't deserialize to the same txn
```
davidiw, line 98: you should consider adding a /why/ as the last parameter of assert_ne, I think it takes in a &str

testsuite/libra-fuzzer/src/fuzz_targets/transaction.rs
```diff
@@ -54,3 +60,82 @@ impl FuzzTargetImpl for SignedTransactionTarget {
...
+        }
+    }
+}
+
+#[derive(Clone, Debug, Default)]
+pub struct TwoSignedTransactions;
+
+impl FuzzTargetImpl for TwoSignedTransactions {
+    fn description(&self) -> &'static str {
+        "Two SignedTransaction's"
```
davidiw, line 109: nit 's there's no possession

davidiw (approved):
can you make the comments more about why, it isn't super useful to see comments that repeat what very easy to read code is already saying.

also it would be super useful to understand what these are fuzzing but there's no general description in this file about what we're fuzzing or why.

Issue #5543, opened 2020-08-10
[libra-channel] garbage collect disconnected peers from channels


phlip9:
[`libra-channel`](https://github.com/libra/libra/blob/master/common/channel/src/libra_channel.rs) never removes keys (`(PeerId, ProtocolId)` in this case) from its `PerKeyQueue`. this works fine for the validator network, where we have a bounded set of peers that rarely changes; however, this does not work for servicing public clients, where we can have large and frequent connection churn.

1. the simplest approach would be to completely remove a key from the `PerKeyQueue` if we pop a message and the queue is empty. however, this would lead to thrashing as we would frequently add+allocate then remove+deallocate the queue.

2. periodically garbage collect on a timer, i.e.,`Receiver::poll_next()` internally polls a timer and removes empty queues on trigger. similar thrashing issue, just less frequent. adds some tail latency from locking the `PerKeyQueue` while iterating through the keys.

3. hook into LostPeer lifecycle directly. add a `remove_key` API to `Sender` (also `Receiver`?). this approach seems the most efficient, but introduces a bit of boilerplate and error-prone ceremony to application protocols, who might forget to clear lost peers. also seems pretty racy; hard to get right.

option (2) seems like a good tradeoff and is completely transparent in the current API.

@gregnazario @davidiw any thoughts?

davidiw:
Hmm.... could we make an explicit callback or trait that implements a GC? And then we can implement different GCs for where ever they are used? When used for networking, the GC could be to simply kill the queue when the peer has disconnected. So you'd have a struct that implements a trait for doing just that. I don't think we use Libra-channel otherwise, but this would keep the code in libra-channel still crate agnostic.

I think your proposal #2 could also be another implementation within this space.

Do we have an estimate of what it costs for each peer across the system? Like the null state space for a single (fullnode) node across all the queues in the system.

gregnazario:
Yeah, I agree.  Let's look at we can make a trait that implements GC, because we could have multiple GC strategies, such as leaving a threshold on the number of queues we let sit and then provide an eviction policy, such as least recently used.  It would be hard to know without real traffic what would be best, because if the messages from a peer come at a time > the poll timer, it's possible we keep creating the queue over and over.

#2 seems like the path to go forward, but generalizing for expansion could be helpful.

[...]

PR #5968, opened 2020-09-10
[libra-channel] add basic GC strat for removing empty per-key-queues

phlip9:
See issue: #5543

This change adds some small logic to remove empty per-key-queues every `POPS_PER_GC := 50` successful dequeue operations.

libra-channel never removes keys from its PerKeyQueue (without this logic). This works fine for the validator network, where we have a bounded set of peers that almost never changes; however, this does not work for servicing public clients, where we can have large and frequent connection churn.

Periodically removing these empty queues prevents us from causing an effective memory leak when we have lots of transient peers in e.g. the public-facing vfn use-case.

This GC strategy could probably be more sophisticated, though it seems to work well enough in some basic stress tests / micro benches.

Using the added `many_keys_stress_test.rs`, it seems like GC'ing every 50 dequeues was the best across a few different tests. Interestingly, having the GC actually seems to be faster than not by ~1.75x

Additionally, we also add a small change to only allocate a small initial queue for a new key. Previously, we allocated a queue with all `max_queue_size_per_key` entries; however, this breaks down when we have lots of transient peers. For example, many of our queues have a max capacity of 1024. To handle a single rpc from a transient peer, we would end up allocating ~ 96 b * 1024 ~ 64 Kib per queue (of which there are several on path when handling an rpc).

Running `cargo x --release -p many-keys-stress-test` before the change:

Instruments - Before

As you can see, memory is growing unbounded (until the bench ends).

In contrast, running after the change:

Instruments - After

the resident set is fairly constant at ~5 MiB. Note some numbers are a bit different b/w these two examples. For example, the "after" run sends 10m messages instead of 2m in the "before" run. The message size was also slightly higher (I realized libra-channel adds 18 bytes overhead to each message).

### DIFF
--- a/common/channel/many-keys-stress-test/src/main.rs
+++ b/common/channel/many-keys-stress-test/src/main.rs
@@ -1,0 +1,105 @@
+// Copyright (c) The Libra Core Contributors
+// SPDX-License-Identifier: Apache-2.0
+
+use channel::{libra_channel, message_queues::QueueStyle};
+use futures::{executor::block_on, stream::StreamExt};
+use std::{
+    io::{Cursor, Write},
+    num::NonZeroUsize,
+    sync::atomic::{AtomicBool, AtomicUsize, Ordering},
+    thread,
+    time::Duration,
+};
+use structopt::StructOpt;
+
+/// A small benchmark/stress test that sends `num_msgs` for each `num_keys`. The
+/// default arguments simulate many transient keys that just push a single message
+/// and then never more. Without garbage collecting empty per-key-queues, the
+/// program will eventually OOM.
+#[derive(Debug, StructOpt)]
+pub struct Args {
+    #[structopt(default_value = "2000000")]
+    num_keys: usize,
+
+    #[structopt(default_value = "1")]
+    num_msgs: usize,
+
+    #[structopt(default_value = "10")]
+    max_queue_size: usize,
+}
+
+fn main() {
+    let args = Args::from_args();
+    run(args);
+}
+
+pub fn run(args: Args) {
+    // Simulates an AccountAddress/PeerId
+    const KEY_SIZE_BYTES: usize = 16;
+
+    // Simulates a (PeerManagerRequest, Option<Arc<_>>)
+    const MSG_SIZE_BYTES: usize = 96;
+
+    static NUM_PUSH: AtomicUsize = AtomicUsize::new(0);
+    static NUM_POP: AtomicUsize = AtomicUsize::new(0);
+    static IS_DONE: AtomicBool = AtomicBool::new(false);
+
+    let (mut sender, mut receiver) = libra_channel::new::<[u8; KEY_SIZE_BYTES], [u8; MSG_SIZE_BYTES]>(
+        QueueStyle::FIFO,
+        NonZeroUsize::new(args.max_queue_size).unwrap(),
+        None,
+    );
+
+    let sender_thread = thread::spawn(move || {
+        for idx in 0..args.num_keys {
+            let mut key = [0u8; KEY_SIZE_BYTES];
+            let mut cursor = Cursor::new(&mut key[..]);
+            cursor.write_all(&idx.to_le_bytes()).unwrap();
+
+            for msg_idx in 0..args.num_msgs {
+                let mut msg = [0u8; MSG_SIZE_BYTES];
+                let mut cursor = Cursor::new(&mut msg[..]);
+                cursor.write_all(&msg_idx.to_le_bytes()).unwrap();
+
+                sender.push(key, msg).unwrap();
+            }
+
+            NUM_PUSH.fetch_add(1, Ordering::Relaxed);
+        }
+    });
+
+    let logger_thread = thread::spawn(move || {
+        while !IS_DONE.load(Ordering::Relaxed) {
+            println!(
+                "NUM_PUSH: {}, NUM_POP: {}",
+                NUM_PUSH.load(Ordering::Relaxed),
+                NUM_POP.load(Ordering::Relaxed),
+            );
+            thread::sleep(Duration::from_secs(1));
+        }
+    });
+
+    // just drain messages
+    let receiver_task = async move {
+        while receiver.next().await.is_some() {
+            NUM_POP.fetch_add(1, Ordering::Relaxed);
+        }
+    };
+
+    block_on(receiver_task);
+    sender_thread.join().unwrap();
+
+    IS_DONE.store(true, Ordering::Relaxed);
+
+    logger_thread.join().unwrap();
+}
+
+#[test]
+fn test_many_keys_stress_test() {
+    let args = Args {
+        num_keys: 100,
+        num_msgs: 1,
+        max_queue_size: 10,
+    };
+    run(args);
+}
--- a/common/channel/src/message_queues.rs
+++ b/common/channel/src/message_queues.rs
@@ -9,9 +9,13 @@ use std::{
     num::NonZeroUsize,
 };
 
+/// Remove empty per-key-queues every `POPS_PER_GC` dequeue operations.
+const POPS_PER_GC: u32 = 50;
+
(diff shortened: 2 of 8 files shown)


crypto/crypto/src/noise.rs
```diff
@@ -121,7 +121,7 @@ pub const fn handshake_resp_msg_len(payload_len: usize) -> usize {
 
 /// This implementation relies on the fact that the hash function used has a 256-bit output
 #[rustfmt::skip]
-const _: [(); 0 - !{ const ASSERT: bool = HashValue::LENGTH == 32; ASSERT } as usize] = [];
+const _: [(); 32] = [(); HashValue::LENGTH];
```
gregnazario, line 124: I'm not sure how this ended up in this PR, but as far as I see this is just an assertion that the hash value length is always 32?  I will say that the result is definitely easier to understand :D

common/channel/many-keys-stress-test/src/main.rs
```diff
@@ -0,0 +1,105 @@
...
+    let args = Args::from_args();
+    run(args);
+}
+
+pub fn run(args: Args) {
+    // Simulates an AccountAddress/PeerId
+    const KEY_SIZE: usize = 16; // bytes
+
+    // Simulates a (PeerManagerRequest, Option<Arc<_>>)
+    const MSG_SIZE: usize = 96; // bytes
```
gregnazario, line 41: Personally, I find it useful to put sizing like `bytes` in the name of the constant, then you won't need the comment, and it'll be clear where you use it

common/channel/many-keys-stress-test/src/main.rs
```diff
@@ -0,0 +1,105 @@
...
+
+#[test]
+fn test_many_keys_stress_test() {
+    let args = Args {
+        num_keys: 1000,
+        num_msgs: 5,
+        max_queue_size: 10,
+    };
+    run(args);
+}
```
gregnazario, line 105: Just wondering, how long does this take to run?

common/channel/src/message_queues.rs
```diff
@@ -100,17 +107,25 @@ impl<K: Eq + Hash + Clone, T> PerKeyQueue<K, T> {
...
+            .or_insert_with(|| VecDeque::with_capacity(1));
+
         // Add the key to our round-robin queue if it's not already there
         if key_message_queue.is_empty() {
             self.round_robin_queue.push_back(key);
         }
+
         // Push the message to the actual key message queue
-        if key_message_queue.len() == max_queue_size {
+        if key_message_queue.len() == self.max_queue_size.get() {
```
gregnazario, line 128: Is it possible to get 2 messages at the same time, such that we pass the max queue size?  Is this code ever run in two threads at once?

I would make it `key_message_queue.len() >= self.max_queue_size.get()` just in case it's somehow used in multiple threads and we push two messages.

common/channel/src/message_queues.rs
```diff
@@ -139,18 +154,51 @@ impl<K: Eq + Hash + Clone, T> PerKeyQueue<K, T> {
...
+            // e.g. the public-facing vfn use-case.
+            //
+            // This GC strategy could probably be more sophisticated, though it
+            // seems to work well in some basic stress tests / micro benches.
+            //
+            // See: common/channel/src/bin/many_keys_stress_test.rs
+            //
+            // For more context, see: https://github.com/libra/libra/issues/5543
+            self.num_popped_since_gc += 1;
+            if self.num_popped_since_gc == POPS_PER_GC {
```
gregnazario, line 188: Same thing about the `self.num_popped_since_gc >= POPS_PER_GC`

common/channel/src/message_queues.rs
```diff
@@ -139,18 +154,51 @@ impl<K: Eq + Hash + Clone, T> PerKeyQueue<K, T> {
...
+            }
         }
+
         message
     }
 
+    /// Garbage collect any empty per-key-queues.
+    fn remove_empty_queues(&mut self) {
+        self.per_key_queue.retain(|_key, queue| !queue.is_empty());
+    }
```
gregnazario, line 200: I'm not fully up to date with how this works, but I saw earlier you add the key to the round robin queue, should you remove the key from the round robin queue?

common/channel/src/message_queues.rs
```diff
@@ -100,17 +107,25 @@ impl<K: Eq + Hash + Clone, T> PerKeyQueue<K, T> {
...
+            // however, this breaks down when we have lots of transient peers.
+            // For example, many of our queues have a max capacity of 1024. To
+            // handle a single rpc from a transient peer, we would end up
+            // allocating ~ 96 b * 1024 ~ 64 Kib per queue.
+            .or_insert_with(|| VecDeque::with_capacity(1));
+
         // Add the key to our round-robin queue if it's not already there
         if key_message_queue.is_empty() {
             self.round_robin_queue.push_back(key);
         }
```
gregnazario, line 125: I think there might be another race condition here if multi-threaded, where we could add a key to the round_robin_queue twice, which probably isn't a big deal.

gregnazario (commented):
This is great!  And the test with the graphs helped me definitely see the improvement.

I mostly had questions around threading especially since it's a common queue, even though only networking uses it.

common/channel/src/message_queues.rs
```diff
@@ -139,18 +154,51 @@ impl<K: Eq + Hash + Clone, T> PerKeyQueue<K, T> {
...
+
         if message.is_some() {
             if let Some(c) = self.counters.as_ref() {
                 c.with_label_values(&["dequeued"]).inc();
             }
+
+            // Remove empty per-key-queues every `POPS_PER_GC` successful dequeue
+            // operations.
+            //
```
gregnazario, line 170: We'll probably want GC based on time or something else if we find this to be expensive, but this is good for now!

crypto/crypto/src/noise.rs
phlip9: ahh, this slipped in, though i'm probably going to leave it : )

i was looking at common sizes of various network structs (that we send around in libra-channels). for example

```
size_of<(PeerId, ProtocolId)>() == 17 bytes
size_of<PeerManagerRequest>() == 88 bytes
size_of<PeerManagerNotification>() == 72 bytes
size_of<(PeerManagerRequest, Option<oneshot::Sender<ElementStatus<PeerManagerRequest>>>)>() == 96 bytes
```

common/channel/many-keys-stress-test/src/main.rs
phlip9: it only takes like 25 ms or so, even in debug mode

common/channel/src/message_queues.rs
phlip9: ahh, `PerKeyQueue` is always used inside a mutex. the rust compiler also prevents this kind of future refactoring error from happening b/c the fn signature takes a `&mut self` which implies exclusive access : )

phlip9: although I agree, making the condition `>=` would be safer

common/channel/src/message_queues.rs
phlip9: iirc, key ∈ round_robin_queue <==> !per_key_queue[key].is_empty()

so all empty queues will not have their keys in the round_robin_queue

common/channel/src/message_queues.rs
phlip9: my initial intuition was to make `POPS_PER_GC` much larger (like 1000+), but this seemed to be slower for whatever reason (of course, this is just in a small micro benchmark, so take everything with several grains of salt)

gregnazario: Weird, we'll probably later have to test it on cluster test or something with a larger scale

crypto/crypto/src/noise.rs
gregnazario: If you can, put it in another commit.

gregnazario (approved):

PR #5969, opened 2020-09-10
[language][vm] let Move VM handle resource serialization

vgao1996:
## Summary
Currently the Move VM emits values and layouts as side effects and let the Libra VM perform serialization. This is less than ideal as it makes the effects API more complicated and exposes `Value`, which is a representation specific to our Move VM implementation. This PR makes the Move VM return bytes (serialized resources) directly. The replacement of `TransactionEffects` by `ChangeSet` (a new type, not the one in Libra) will also allow us to build tools that handle writesets with ease, benefitting testing infra and genesis generation.

## Test Plan
cargo test

### DIFF
--- a/language/move-core/types/src/effects.rs
+++ b/language/move-core/types/src/effects.rs
@@ -1,0 +1,30 @@
+// Copyright (c) The Libra Core Contributors
+// SPDX-License-Identifier: Apache-2.0
+
+use crate::{
+    account_address::AccountAddress,
+    identifier::Identifier,
+    language_storage::{StructTag, TypeTag},
+};
+use std::collections::btree_map::BTreeMap;
+
+#[derive(Debug, Clone)]
+pub enum Op {
+    Write(Vec<u8>),
+    Deletion,
+}
+
+/// A collection of changes to modules and resources under a Move account.
+#[derive(Debug, Clone)]
+pub struct AccountChangeSet {
+    pub modules: BTreeMap<Identifier, Op>,
+    pub resources: BTreeMap<StructTag, Op>,
+}
+
+/// A collection of changes to a Move state.
+#[derive(Debug, Clone)]
+pub struct ChangeSet {
+    pub accounts: BTreeMap<AccountAddress, AccountChangeSet>,
+}
+
+pub type Event = (Vec<u8>, u64, TypeTag, Vec<u8>);
--- a/language/move-vm/runtime/src/data_cache.rs
+++ b/language/move-vm/runtime/src/data_cache.rs
@@ -6,6 +6,8 @@ use crate::loader::Loader;
 use libra_logger::prelude::*;
 use move_core_types::{
     account_address::AccountAddress,
+    effects::{AccountChangeSet, ChangeSet, Event, Op},
+    identifier::Identifier,
     language_storage::{ModuleId, StructTag, TypeTag},
     value::MoveTypeLayout,
     vm_status::StatusCode,
@@ -15,7 +17,7 @@ use move_vm_types::{
     loaded_data::runtime_types::Type,
     values::{GlobalValue, GlobalValueEffect, Value},
 };
-use std::collections::btree_map::BTreeMap;
+use std::collections::BTreeMap;
 use vm::errors::*;
 
 /// Trait for the Move VM to abstract `StateView` operations.
@@ -32,7 +34,7 @@ pub trait RemoteCache {
 
 pub struct AccountDataCache {
     data_map: BTreeMap<Type, (MoveTypeLayout, GlobalValue)>,
-    module_map: BTreeMap<ModuleId, Vec<u8>>,
+    module_map: BTreeMap<Identifier, Vec<u8>>,
 }
 
 impl AccountDataCache {
@@ -64,15 +66,6 @@ pub(crate) struct TransactionDataCache<'r, 'l, R> {
     event_data: Vec<(Vec<u8>, u64, Type, MoveTypeLayout, Value)>,
 }
 
-pub struct TransactionEffects {
-    pub resources: Vec<(
-        AccountAddress,
-        Vec<(StructTag, Option<(MoveTypeLayout, Value)>)>,
-    )>,
-    pub modules: Vec<(ModuleId, Vec<u8>)>,
-    pub events: Vec<(Vec<u8>, u64, TypeTag, MoveTypeLayout, Value)>,
-}
-
 impl<'r, 'l, R: RemoteCache> TransactionDataCache<'r, 'l, R> {
     /// Create a `TransactionDataCache` with a `RemoteCache` that provides access to data
     /// not updated in the transaction.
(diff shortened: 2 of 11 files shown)


language/move-core/types/src/effects.rs
```diff
@@ -0,0 +1,46 @@
...
+
+/// A collection of changes to modules and resources under a Move account.
+#[derive(Clone)]
+pub struct AccountChangeSet {
+    pub modules: BTreeMap<Identifier, Option<Vec<u8>>>,
+    pub resources: BTreeMap<StructTag, Option<Vec<u8>>>,
+}
+
+impl AccountChangeSet {
+    pub fn squash(&mut self, other: Self) -> Result<()> {
```
sblackshear, line 17: Given that this + the other `squash` is only used in `vm_genesis`, I wonder if we could implement it there instead? I have some reservations about exposing this as a utility function because the behavior when `self` deletes a module/resource and `other` adds it (or vice versa) is potentially surprising... It would be easy to make the mistake of thinking that this is commutative.

language/move-vm/runtime/src/data_cache.rs
```diff
@@ -89,54 +82,57 @@ impl<'r, 'l, R: RemoteCache> TransactionDataCache<'r, 'l, R> {
...
-            );
+
+            account_changesets.insert(addr, AccountChangeSet { modules, resources });
         }
 
         let mut events = vec![];
         for (guid, seq_num, ty, ty_layout, val) in self.event_data {
             let ty_tag = self.loader.type_to_type_tag(&ty)?;
-            events.push((guid, seq_num, ty_tag, ty_layout, val))
+            // REVIEW: This error doesn't look right.
```
sblackshear, line 123: Agreed; this should be something that results in a discard. Perhaps `INTERNAL_TYPE_ERROR`? I think that's the only reason this can fail--`ty_layout` doesn't match `val`.

language/move-core/types/src/effects.rs
```diff
@@ -0,0 +1,46 @@
...
+    identifier::Identifier,
+    language_storage::{StructTag, TypeTag},
+};
+use anyhow::Result;
+use std::collections::btree_map::{self, BTreeMap};
+
+/// A collection of changes to modules and resources under a Move account.
+#[derive(Clone)]
+pub struct AccountChangeSet {
+    pub modules: BTreeMap<Identifier, Option<Vec<u8>>>,
```
sblackshear, line 12: It might be slightly clearer to use a `Write(Vec<u8>), Delete` enum here instead of `Option`.

language/move-core/types/src/effects.rs
vgao1996: Sounds like a good idea in general. I already ended up making this function stricter in a subsequent PR (coming soon) due to similar concerns. Let me try to finalize that one and see what exactly is the best way to handle `squash`.

language/move-vm/runtime/src/data_cache.rs
vgao1996: Should this be classified as a breaking change? I mean we're switching to a different bucket here, but this should be dead code if the VM isn't bugged...

language/move-core/types/src/effects.rs
vgao1996: Just realized I don't immediately need `squash` for the incoming PR (#5972). What do you think about just deleting it for now? I can add it back once I start working on the new testing libraries that actually need it and by that time I should understand the desired behavior much better than now.

bob-wilson:
As I mentioned in #5972 (which I hadn't realized was building on top of this), I'm concerned that this kind of change may not be worth the risk so close to a release, and well past our deadline for cleanup/refactoring changes. We are really only supposed to be fixing bugs right now. Is there an important reason why we should make an exception in this case?

vgao1996:
Copying what I wrote from #5972, but to rephrase, my arguments would be that
1. This is low risk -- it's not really changing the semantics in a meaningful way. Most changes are mechanical.
2. Not doing this will require adding additional conversion logic, which carries their own burden.
3. The effects API proposed offers a sane way to reason about effects produced by the Move VM, which will make the new testing infra, future and existing tools safer. In fact, by simplifying the genesis creation logic using this, I've already realized our current genesis generation tool is not invalidating the Move VM when it should. (See my comments in #5949)

language/move-vm/runtime/src/data_cache.rs
tnowacki: I wouldn't worry about this case being a breaking change or not. But should it be a VALUE_SERIALIZATION_ERROR? that matches whats done for the rest of the changeset

tnowacki: A bigger question is should these two serialization errors be user errors or invariant violations? Maybe FAILED_TO_SERIALIZE_XXX or something?

tnowacki:
I'm also a bit concerned about any changes like this right now. 

I think the best part about this PR is us looking a bit more closely at the errors we produce in these use cases (one of them DEFINITELY shouldn't be a runtime error, we missed that)

I could go either way on some of the locking down of Value into the VM. It doesn't feel like its a big issue if we miss it for this release (and we can make a much bigger, more comprehensive push next release)

language/move-vm/runtime/src/data_cache.rs
vgao1996: I would vote for invariant violation -- both the value and the layout are generated by the Move VM itself and there's no user action that could cause this to fail.

jolestar:
👍， so the LCS should be part of Move VM standard?

language/move-vm/runtime/src/data_cache.rs
sblackshear: I definitely think this should be an invariant violation, but don't have a strong opinion about the specific error code.

vgao1996:
@jolestar Not in the long term, but for the current version it's something we chose to live with. We are evaluating ideas like making the Move VM parametric over serialization/deserialization, or abstracting out the core of the VM into another layer that is agnostic of the storage format. However similar to what we talked about last time, it will take some time before any of those lands, and this "backward" change is to make our life easier during the transient period.

However even with that being said, I don't think serialization of Move values is that tightly coupled to LCS -- all the rules are expressed in the form of serde constructs, and serde does allow you to map those to various output formats, like JSON, YAML etc., and LCS is just the one particular custom binary format we're using. It should be fairly easy for you to implement your own format as long as you are able to work with serde.

language/tools/vm-genesis/src/lib.rs
```diff
@@ -90,14 +91,20 @@ pub fn encode_genesis_transaction(
...
 }
 
-fn merge_txn_effects(
-    mut effects_1: TransactionEffects,
-    effects_2: TransactionEffects,
-) -> TransactionEffects {
-    effects_1.resources.extend(effects_2.resources);
-    effects_1.modules.extend(effects_2.modules);
-    effects_1.events.extend(effects_2.events);
-    effects_1
```
vgao1996, line 100: Just realized this (the old implementation of merge_txn_effects) could also become a time bomb in the future: if the two sets of effects we're merging overlap, we'll issue more than one commands for some resources. The DB won't necessarily choke because of this, but still it's kinda nasty and similar to the extraneous deletion bug we had...

vgao1996: In other words, one property that differentiates `ChangeSet` from `TransactionEffects` is that it is guaranteed by the type system that all entries are unique. I think the reason why I used `Vec` instead of `BTreeMap` for `TransactionEffects` was that I was seeing it as a final result that didn't need to be recomputed/modified. This may be true for the normal transaction flows, but now we can see there's certainly need for manipulating these effects in the realm of writeset generation, in a way similar to how you rebase and edit git commits. I think this is also something we should consider...

PR #5971, opened 2020-09-10
[move-prover] Refactor memory implementation

wrwg:
This PR refactors the memory implementation:

- Uses its own simple representation for borrow graph. There were a few bugs in the combination of the Move graph and the need for us to track edges also after references die. It appeared easier to use our own representation as our usage of the graph is rather basic.
- Moving mutref elimination before borrow analysis. This leads to code reduction because we do not need to special case converted parameters.
- Moving writeback optimization into a dedicated final optimization phase.
- Improving libs for processing pipelines, factoring common code out.
- Fixing boogie generation to assume/assert data invariants at the right points. The old logic was based on &mut ref, but the memory model trafo has those removed for parameters.

All tests pass, and the previously failing tests in `mut_ref_accross_modules.move` now work again (as before the new memory model). These changes have not yet applied to reactivate Libra sum-of-coins, which will happen in a separate PR.

Fixing packref/unpackref.

Yes

## Test Plan

Existing tests (reactivated some old ones)

## Related PRs

NA


shazqadeer (approved):
Approving based on our conversation.  Trust you will land only when appropriate.

PR #5972, opened 2020-09-10
[draft][language][tools] add Move genesis builder

vgao1996:
## Summary
This adds a small library "genesis-builder" that helps simplify genesis generation. This will make it easier for us to build different geneses, which is required for future testing needs.

## Test Plan
cargo test

## Dependencies
This is based on #5969


bob-wilson:
From the description of this PR, I was expecting to see some new code for this small library, but instead I see that it is touching a lot of code in critical parts of the VM. The changes look straightforward but as we get closer to a release, I have to ask if that is necessary. There's always a non-zero risk of bugs resulting from code churn. It looks like most of the changes are related to the use of TransactionEffects. Would it be possible to implement this without all those changes?

vgao1996:
@bob-wilson the part you are describing is #5969, which this PR depends on. Technically it is possible to keep `TransactionEffects` intact, but then I'll still need to create something identical to the current `ChangeSet` and implement conversions between these two, which will add quite some complexities and new logic to the point where I just don't think it's worth it. I'm aware of the risk of changing anything in the VM right now, but given that `ChangeSet` is a drop-in replacement of `TransactionEffects` and that the change is more about making code cleaner and more organized rather than altering the underlying semantics in a meaningful way,I feel fairly comfortable about it.

bob-wilson:
> @bob-wilson the part you are describing is #5969, which this PR depends on. Technically it is possible to keep `TransactionEffects` intact, but then I'll still need to create something identical to the current `ChangeSet` and implement conversions between these two, which will add quite some complexities and new logic to the point where I just don't think it's worth it. I'm aware of the risk of changing anything in the VM right now, but given that `ChangeSet` is a drop-in replacement of `TransactionEffects` and that the change is more about making code cleaner and more organized rather than altering the underlying semantics in a meaningful way,I feel fairly comfortable about it.

I see. I had missed the fact that there were 2 commits here. I'll comment in the other PR....

PR #5976, opened 2020-09-10
[ci] Add new hyperjump capability to request reviews

metajack:
This adds a new hyperjump called request-review which can request reviews from individuals or teams. This will be used by the dependency verification action to request review from @libra/dep-audit.


rexhoffman (approved):
Hopefully the hyperjump token has minimal permissions... Just triage I'd wager.   Based on conversations, I'd wager that's the case.   Also.... so much generated code.   Soon this repo will be more generated javascript than rust.

metajack:
Yes, the user libra-action has triage permissions only, and the token only repo scope.

PR #5982, opened 2020-09-10
[network] rename some peer_ids appropriately to remote_peer_id

gregnazario:
### Overview
I went over a few peer_ids to rename them because i found that the peer actor was using `self_peer_id` improperly, or confusingly.

Should make a run through the rest of the peer_ids.


phlip9 (approved):
LGTM

PR #5983, opened 2020-09-10
[fuzz] types/proof fuzzers

lightmark:
## Motivation

fuzz types/proofs that have `verify` method and `Arbitrary` trait.


testsuite/libra-fuzzer/src/fuzz_targets.rs
```diff
@@ -13,6 +13,8 @@ mod json_rpc_service;
 mod mempool;
 mod move_vm;
 mod network;
+mod network_noise;
```
mimoo, line 16: remove this

mimoo (approved):
woop woop!

PR #5984, opened 2020-09-10
[Safety Rules] Maintain an in-memory cached SafetyData to avoid unnecessary reads from storage.

JoshLind:
## Motivation

This PR adds support for an in-memory cached SafetyData for PersistentSafetyStorage. By caching SafetyData in-memory we avoid having to make unnecessary reads to storage (e.g., vault), improving safety rules performance. Whenever SafetyData is written by safety_rules, the cache as well as storage is updated. 

The PR offers two commits to achieve this:
1. Provide some random code clean ups to PersistentSafetyStorage.
2. Add support for the cached SafetyData as well as a SafetyRulesConfig variable to wrap the feature.

Using the benchmark in safety rules we get the following (rough) performance improvements:
- InMemory: No improvement.
- OnDisk: 8% improvement.
- Vault: 33% improvement.

Yes.

## Test Plan

All tests pass locally.

## Related PRs

None.

### DIFF
--- a/consensus/safety-rules/src/persistent_safety_storage.rs
+++ b/consensus/safety-rules/src/persistent_safety_storage.rs
@@ -16,7 +16,14 @@ use libra_types::waypoint::Waypoint;
 /// SafetyRules needs an abstract storage interface to act as a common utility for storing
 /// persistent data to local disk, cloud, secrets managers, or even memory (for tests)
 /// Any set function is expected to sync to the remote system before returning.
+///
+/// Note: cached_safety_data is a local in-memory copy of SafetyData. As SafetyData should
+/// only ever be used by safety rules, we maintain an in-memory copy to avoid issuing reads
+/// to the internal storage if the SafetyData hasn't changed. On writes, we update the
+/// cache and internal storage.
 pub struct PersistentSafetyStorage {
+    enable_cached_safety_data: bool,
+    cached_safety_data: Option<SafetyData>,
     internal_store: Storage,
 }
 
@@ -29,20 +36,28 @@ impl PersistentSafetyStorage {
         consensus_private_key: Ed25519PrivateKey,
         execution_private_key: Ed25519PrivateKey,
         waypoint: Waypoint,
+        enable_cached_safety_data: bool,
     ) -> Self {
+        let safety_data = SafetyData::new(1, 0, 0, None);
         Self::initialize_(
             &mut internal_store,
+            safety_data.clone(),
             author,
             consensus_private_key,
             execution_private_key,
             waypoint,
         )
         .expect("Unable to initialize backend storage");
-        Self { internal_store }
+        Self {
+            enable_cached_safety_data,
+            cached_safety_data: Some(safety_data),
+            internal_store,
+        }
     }
 
     fn initialize_(
         internal_store: &mut Storage,
+        safety_data: SafetyData,
         author: Author,
         consensus_private_key: Ed25519PrivateKey,
         execution_private_key: Ed25519PrivateKey,
@@ -60,7 +75,7 @@ impl PersistentSafetyStorage {
         }
 
         internal_store.import_private_key(EXECUTION_KEY, execution_private_key)?;
-        internal_store.set(SAFETY_DATA, SafetyData::new(1, 0, 0, None))?;
+        internal_store.set(SAFETY_DATA, safety_data)?;
         internal_store.set(OWNER_ACCOUNT, author)?;
         internal_store.set(WAYPOINT, waypoint)?;
         Ok(())
@@ -68,8 +83,12 @@ impl PersistentSafetyStorage {
 
     /// Use this to instantiate a PersistentStorage with an existing data store. This is intended
     /// for constructed environments.
-    pub fn new(internal_store: Storage) -> Self {
-        Self { internal_store }
+    pub fn new(internal_store: Storage, enable_cached_safety_data: bool) -> Self {
+        Self {
+            enable_cached_safety_data,
+            cached_safety_data: None,
+            internal_store,
+        }
     }
 
     pub fn author(&self) -> Result<Author, Error> {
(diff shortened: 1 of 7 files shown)


davidiw (commented):
Can we make this configurable? I'm just a little concerned that adding this in so late... maybe something weird happens.

Other than that looks good.

JoshLind:
Thanks, @davidiw! 😄 

> Can we make this configurable? I'm just a little concerned that adding this in so late... maybe something weird happens.

What kind of configuration are you thinking -- env variable, compiler flag, cli argument? While I'd normally agree with you that making changes this late isn't great, my only concerns with making this configurable are:
1. The logic here is very simple -- so simple that making it configurable actually comes off as more complex than the change itself 😢. Although not a deal breaker, it leads into:
2. Even if this is made configurable, to be sure we're testing it enough, it should be enabled by default. My concern then becomes that if there is an error, what we'd do is disable the flag and rollback the specific commit (commit 2). However, if it's not configurable, and there is an error, we'd just roll-back the specific commit. I guess the only advantage to configurability then is that if we discover a bug in a prod environment (where it's not feasible to change the binary immediately) we'd just switch the flag. But, again, I don't know how well we support configurability now anyway (e.g., it may require a reboot or a recompile currently...). So it seems like we don't currently have great support for configurable features.
 
Overall, I think we're still in the ball park of integrating this in without a flag. If we were closer to "D-Day", I'd have no hesitations. Thoughts? You may still feel differently..

davidiw:
If we find a bug in prod, we cannot easily roll to another version, but we can modify the config value in the node.config really easily to disable it. Then you would restart the pod and the configuration takes effect.

consensus/safety-rules/src/persistent_safety_storage.rs
```diff
@@ -109,14 +113,17 @@ impl PersistentSafetyStorage {
     }
 
     pub fn safety_data(&self) -> Result<SafetyData, Error> {
-        Ok(self.internal_store.get(SAFETY_DATA).map(|v| v.value)?)
+        Ok(self.cached_safety_data.clone())
```
mgorven, line 116: This doesn't seem to ever read the safety_data from Vault. I think the cache should be `Option<SafetyData>`, and this function must read from Vault if it is `none`.

consensus/safety-rules/src/persistent_safety_storage.rs
```diff
@@ -109,14 +113,17 @@ impl PersistentSafetyStorage {
...
     }
 
     pub fn set_safety_data(&mut self, data: SafetyData) -> Result<(), Error> {
         counters::set_state("epoch", data.epoch as i64);
         counters::set_state("last_voted_round", data.last_voted_round as i64);
         counters::set_state("preferred_round", data.preferred_round as i64);
-        self.internal_store.set(SAFETY_DATA, data)?;
+
+        self.internal_store.set(SAFETY_DATA, data.clone())?;
```
mgorven, line 124: Once we have check-and-set with Vault, this needs to invalidate the cache if the write fails so that it gets re-read the next time.

mgorven:
I think we first need Vault check-and-set (https://github.com/libra/libra/pull/6007) before we can safely add caching.

mgorven (changes requested):

consensus/safety-rules/src/persistent_safety_storage.rs
davidiw: I think we should be clear what the right behavior is... in general, if safety rules ever hits an invariant such as cache failing, we should let the admin know even if we automatically recover. this could be because bad things happened...

consensus/safety-rules/src/persistent_safety_storage.rs
```diff
@@ -76,16 +72,24 @@ impl PersistentSafetyStorage {
...
 
+    fn new_safety_data() -> SafetyData {
+        SafetyData::new(1, 0, 0, None)
+    }
+
     /// Use this to instantiate a PersistentStorage with an existing data store. This is intended
     /// for constructed environments.
     pub fn new(internal_store: Storage) -> Self {
-        Self { internal_store }
+        let safety_data = Self::new_safety_data();
```
davidiw, line 88: Great call @mgorven, this is really unsafe :P, this should indeed be an option.

consensus/safety-rules/src/persistent_safety_storage.rs
mgorven: Currently this will happen on upgrades with a new genesis (because the vault-init TF resets the safety_data in Vault). In production it should never happen though. In my mind the model is that whatever is in Vault is canonical, but I can see a model where the LSR process is also canonical and if its view of the world doesn't match what's in Vault it just stops processing until the operator intervenes.

consensus/safety-rules/src/persistent_safety_storage.rs
JoshLind: Aah, thanks for pointing this out @mgorven! I wasn't aware we had a use case where we could load the SafetyData from storage upon startup, but in hindsight it's clear we need to support for that failure-recovery cases 🤦. I've updated the code so that we only use the cache if the flag is enabled and the cache is set.

consensus/safety-rules/src/persistent_safety_storage.rs
JoshLind: What I've done for now is throw an error if the write fails, which as I understand it, LSR will log but continue execution (similar to any write error as it stands today). I'm not sure we want to panic (we try to avoid doing that in LSR as a matter of principle), so I've handled this case as any other storage failure-related case: if the write fails (for any reason), invalidate the local cache and log the error.

JoshLind:
Thanks @mgorven, @davidiw! 😄 

I've wrapped this in a SafetyRulesConfig flag "enable_cached_safety_data" which is defaulted to true. If the flag is disabled, reads will always hit storage, otherwise we use the cache. I'm not sure if there's an easy way for us to ensure that this flag is tied to the check-and-set flag (although there is no fundamental reason for needing it to be case). I guess having two flags is ideal though because it gives us more flexibility (e.g., if we find a bug in one feature and not the other).

mgorven (approved):
Awesome stuff

consensus/safety-rules/src/persistent_safety_storage.rs
mgorven: Yup, this is the right behaviour.

mgorven: If the write fails, it will fail the current vote though right? (i.e. it won't return the vote to the validator)

JoshLind: Yeah, it'll behave the same as any other failure that's thrown during execution (i.e., exit early and return the error to the caller -- which should be the safety rules client in a remote safety rules deployment).

JoshLind:
Thanks, all! 😄 

/land

PR #5985, opened 2020-09-10
[specifications] add common data structure spec + fix links

mimoo:


JoshLind (approved):
LGTM! Thanks @mimoo!

/land

PR #5988, opened 2020-09-11
[Safety Rules] Don't load safety data before verifying signer.

JoshLind:
## Motivation

This PR provides some tiny refactors to safety rules, namely:

- Be sure to verify "self.signer()?" before loading safety data, as it might require a storage access (which could be expensive depending on how storage is implemented).

Yes.

## Test Plan

All tests pass locally.

## Related PRs

None.

### DIFF
--- a/consensus/safety-rules/src/safety_rules.rs
+++ b/consensus/safety-rules/src/safety_rules.rs
@@ -304,7 +304,6 @@ impl SafetyRules {
     ) -> Result<Vote, Error> {
         // Exit early if we cannot sign
         self.signer()?;
-        let mut safety_data = self.persistent_storage.safety_data()?;
 
         let (vote_proposal, execution_signature) = (
             &maybe_signed_vote_proposal.vote_proposal,
@@ -319,6 +318,7 @@ impl SafetyRules {
         }
 
         let proposed_block = vote_proposal.block();
+        let mut safety_data = self.persistent_storage.safety_data()?;
         self.verify_epoch(proposed_block.epoch(), &safety_data)?;
         self.verify_qc(proposed_block.quorum_cert())?;
         proposed_block
@@ -339,15 +339,14 @@ impl SafetyRules {
             &mut safety_data,
         )?;
 
-        let vote_data = self.extension_check(vote_proposal)?;
-
         let validator_signer = self.signer()?;
         let vote = Vote::new(
-            vote_data,
+            self.extension_check(vote_proposal)?,
             validator_signer.author(),
             self.construct_ledger_info(proposed_block)?,
             validator_signer,
         );
+
         safety_data.last_vote = Some(vote.clone());
         self.persistent_storage.set_safety_data(safety_data)?;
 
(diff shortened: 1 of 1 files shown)


davidiw (commented):
As you can see, we are hitting mutability issue. I think this code has been refactored a few times, but originally the validator signer was at the beginning and at the end, but then there was desire to only call it once.

JoshLind:
Aah, thanks @davidiw! I actually fixed this last night, but apparently forgot to push 😄. Just avoided the issue with a clone on the first call.

consensus/safety-rules/src/safety_rules.rs
```diff
@@ -352,30 +350,34 @@ impl SafetyRules {
...
         Ok(Block::new_proposal_from_block_data(
             block_data,
-            self.signer()?,
+            &validator_signer,
         ))
     }
 
     fn guarded_sign_timeout(&mut self, timeout: &Timeout) -> Result<Ed25519Signature, Error> {
-        self.signer()?;
+        let validator_signer = self.signer()?.clone();
```
davidiw, line 379: I'm confused at how this can work since signer shouldn't be cloneable. Can you try building without testing / fuzzing enabled.

zekun000: I think it clones the reference

JoshLind: @davidiw, yeah, without looking that closely it clones the reference as self.signer()? returns an `&ValidatorSigner`. Thinking about it, I've yet to decide how bad this is.. 😺

davidiw: I don't understand what that means.

davidiw: A clone on a reference should return a copy of the value afaiu

JoshLind: @davidiw, hmm, that's not what I see. For example, see here (https://www.reddit.com/r/rust/comments/d3wo9i/cloning_a_reference/) and the link to the rust code where '&' implements Clone() (https://doc.rust-lang.org/std/primitive.reference.html#trait-implementations

JoshLind: Either way, this is still broken. Give me some time to think about the best way to do what I'd like to and I'll get back to you both 💩

davidiw: Maybe looking at the implementation will make it clearer?

https://doc.rust-lang.org/src/core/clone.rs.html#225-230

`&T::clone() -> T`

JoshLind: Hah! Thanks @davidiw! 😃 Who would have thought that reddit wasn't the best place to get an authoritative opinion on rust...? 🤦 Anyway, I've just reverted the breaking nit picks. Overall the changes I'd have to make to get this to work are not worth it. Even as it is, this PR is a bit nit picky :P

davidiw: I have no comment on reddit or your nitpicking

davidiw (approved):
and then I scratch my head and wonder... what exactly is the purpose of this PR but I appreciate and support your exploration!

JoshLind:
Thanks, @davidiw!

> and then I scratch my head and wonder... what exactly is the purpose of this PR but I appreciate and support your exploration!

Granted, the impact of this is much less given that we couldn't clean up self.signer(), but there is benefit in not forcing a potentially expensive vault data fetch if the signer is invalid and we should just terminate early anyway.. 😜

PR #5998, opened 2020-09-11
Convert remaining network logs to structured logs

gregnazario:
### Overview
This converts the remaining logs to structured logs.

Key takeaways:
* We can take advantage of ConnectionMetadata, to connect logs to a specific instance of a connection
* We need to review levels of logs and strategies holistically after this e.g. not `error!` when a loop is terminated on purpose.
* A couple logs can't currently easily copy the network context in its current form e.g. loop of moves within a move.

### DIFF
--- a/network/src/peer/mod.rs
+++ b/network/src/peer/mod.rs
@@ -5,6 +5,7 @@
 //! and opening substreams as well as negotiating particular protocols on those substreams.
 use crate::{
     counters,
+    logging::NetworkSchema,
     peer_manager::PeerManagerError,
     protocols::wire::messaging::v1::{ErrorCode, NetworkMessage},
     transport,
@@ -78,7 +79,7 @@ enum State {
 
 pub struct Peer<TSocket> {
     /// The network instance this Peer actor is running under.
-    _network_context: Arc<NetworkContext>,
+    network_context: Arc<NetworkContext>,
     /// A handle to a tokio executor.
     executor: Handle,
     /// Connection specific information.
@@ -119,7 +120,7 @@ where
             socket,
         } = connection;
         Self {
-            _network_context: network_context,
+            network_context,
             executor,
             connection_metadata,
             connection: Some(socket),
@@ -139,7 +140,10 @@ where
     pub async fn start(mut self) {
         let remote_peer_id = self.remote_peer_id();
         info!(
-            "Starting Peer actor for peer: {:?}",
+            NetworkSchema::new(&self.network_context)
+                .connection_metadata(&self.connection_metadata),
+            "{} Starting Peer actor for peer: {:?}",
+            self.network_context,
             remote_peer_id.short_str()
         );
 
@@ -164,8 +168,12 @@ where
         // the task:
         // `write_reqs_tx`: Instruction to send a NetworkMessage on the wire.
         // `close_tx`: Instruction to close the underlying connection.
-        let (write_reqs_tx, close_tx) =
-            Self::start_writer_task(&self.executor, remote_peer_id, writer);
+        let (write_reqs_tx, close_tx) = Self::start_writer_task(
+            &self.executor,
+            self.connection_metadata.clone(),
+            self.network_context.clone(),
+            writer,
+        );
         // Start main Peer event loop.
         loop {
             match self.state {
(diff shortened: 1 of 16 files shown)


network/src/peer/mod.rs
```diff
@@ -409,26 +498,36 @@ where
...
+    connection_metadata: ConnectionMetadata,
+    network_context: Arc<NetworkContext>,
     sender: channel::Sender<PeerRequest>,
 }
 
 impl Clone for PeerHandle {
     fn clone(&self) -> Self {
         Self {
-            peer_id: self.peer_id,
+            network_context: self.network_context.clone(),
```
phlip9, line 509: i wonder if we can just `#[derive(Clone)]` on `PeerHandle`?

network/src/peer/mod.rs
```diff
@@ -139,7 +140,10 @@ where
     pub async fn start(mut self) {
         let self_peer_id = self.peer_id();
```
phlip9, line 141: Let's rename this to `remote_peer_id`. I think `self_peer_id` is super confusing here

network/src/transport.rs
```diff
@@ -120,6 +120,17 @@ impl ConnectionMetadata {
             application_protocols,
         }
     }
+
+    pub fn mock(peer_id: PeerId) -> ConnectionMetadata {
```
phlip9, line 125: should be able to `#[cfg(any(test, feature = "testing", feature = "fuzzing"))]` here

network/src/peer/mod.rs
```diff
@@ -409,26 +498,36 @@ where
...
         }
     }
 }
 
 impl PeerHandle {
-    pub fn new(peer_id: PeerId, sender: channel::Sender<PeerRequest>) -> Self {
-        Self { peer_id, sender }
+    pub fn new(
+        network_context: Arc<NetworkContext>,
+        connection_metadata: ConnectionMetadata,
```
phlip9, line 519: Maybe we should `Arc<ConnectionMetadata>`... since we have the same struct copied in 4 places for each peer that connects. It also contains a `NetworkAddress`, which can be ~140 bytes since lots of validators use their long LB dns names. I want to make sure we keep our memory usage in check so we don't like alloc 1 MiB per peer and then suddenly only support 1000 connections before OOM lol

network/src/peer/mod.rs
```diff
@@ -184,18 +192,37 @@ where
                             match maybe_message {
                                 Some(Ok(message)) =>  {
                                     if let Err(err) = self.handle_inbound_message(message, write_reqs_tx.clone()).await {
-                                        warn!("Error in handling inbound message from peer: {}. Error: {:?}",
-                                            self_peer_id.short_str(), err);
+                                        warn!(
+                                            NetworkSchema::new(&self.network_context)
+                                                .connection_metadata(&self.connection_metadata)
```
phlip9, line 197: I see that `.connection_metadata` includes the `network_address` in each log, I wonder if that might be too much overhead? Remember that rendered `network_address` can be as long as

```
/dns4/aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa.elb.us-east-9.amazonaws.com/tcp/9999/ln-noise-ik/aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa/ln-handshake/0
```

I'm not too worried about these error logs, since they're probably very infrequent; however, the "Starting Peer actor", "Peer actor shutdown", and "Closed connection with Peer" will happen for every connection. Of course, if this is no prob. for the logging backend, then just ignore this 😆

network/src/peer/mod.rs
gregnazario: See: https://github.com/libra/libra/pull/5982

network/src/peer/mod.rs
gregnazario: Yeah, I think it's a tradeoff we can make.

Can probably log just the PeerId, I might have gone a little overboard with this.

phlip9: I think it's fine to also log the connection id and connection origin, just adding the addr is probably too much though

davidiw (approved):
Love the PR. My biggest concern is that we probably need to make it really easy to disable logging / log pushing from VFNs. cc: @sherry-x

network/src/peer/mod.rs
gregnazario: Let's open an issue for it, and do in a separate PR

network/src/protocols/health_checker/mod.rs
```diff
@@ -244,7 +252,13 @@ where
                                     self.ping_timeout.clone()));
                         }
                         None => {
-                            debug!("{} No connected peer to ping round: {}", self.network_context, self.round);
+                            debug!(
+                                NetworkSchema::new(&self.network_context),
+                                round = self.round,
+                                "{} No connected peer to ping round: {}",
```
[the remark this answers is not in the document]
gregnazario: I wasn't really worried because it's a `debug` level line.  It was more of, let me get all of the logs for a specific network instance, and put them all together.

gregnazario:
> Love the PR. My biggest concern is that we probably need to make it really easy to disable logging / log pushing from VFNs. cc: @sherry-x

I guess your concern is about the explosion of logs from the PeerIds?  I agree that we'll need a way to prevent logging from VFNs.  Let's create an issue for it and follow up

network/src/peer/mod.rs
gregnazario: Done, added a separate function to allow logging network address sometimes.

PR #6000, opened 2020-09-11
[metrics] Gather metrics of gathering metrics

gregnazario:
### Overview
To not currently limit the number of metrics, we will just add metrics for gathering data about the metrics.  This should let us make informed decisions, especially since there doesn't seem to be any current scaling issues.

### Testing to figure out how logs are used
I wrote a test to determine how two counters would work in storage. and how they're printed as a string.  The metrics poll rate determines the fidelity of gauges as I've learned from this.
```
fn test() {
    LIBRA_NETWORK_PEERS
        .with_label_values(&["A", "connected"])
        .set(1 as i64);

    LIBRA_NETWORK_PEERS
        .with_label_values(&["A", "connected"])
        .set(20 as i64);

    LIBRA_NETWORK_PEERS
        .with_label_values(&["A", "not-connected"])
        .set(2 as i64);

    LIBRA_NETWORK_PEERS
        .with_label_values(&["B", "connected"])
        .set(3 as i64);
    let all_metrics = prometheus::gather();
    for family in all_metrics {
        println!("METRIC[{}]:", family.get_name());
        let metrics = family.get_metric();
        for metric in metrics {
            println!("{:?}", metric);
        }
        println!("--\n");
    }
}
```

```
METRIC[INT_COUNTER]:
Metric { label: [], gauge: Gauge { value: 0.0 }, counter: Counter { value: 12.0 }, summary: Summary { sample_count: 0, sample_sum: 0.0, quantile: [] }, untyped: Untyped { value: 0.0 }, histogram: Histogram { sample_count: 0, sample_sum: 0.0, bucket: [] }, timestamp_ms: 0 }
--

METRIC[libra_network_peers]:
Metric { label: [LabelPair { name: "role_type", value: "A" }, LabelPair { name: "state", value: "connected" }], gauge: Gauge { value: 20.0 }, counter: Counter { value: 0.0 }, summary: Summary { sample_count: 0, sample_sum: 0.0, quantile: [] }, untyped: Untyped { value: 0.0 }, histogram: Histogram { sample_count: 0, sample_sum: 0.0, bucket: [] }, timestamp_ms: 0 }
Metric { label: [LabelPair { name: "role_type", value: "A" }, LabelPair { name: "state", value: "not-connected" }], gauge: Gauge { value: 2.0 }, counter: Counter { value: 0.0 }, summary: Summary { sample_count: 0, sample_sum: 0.0, quantile: [] }, untyped: Untyped { value: 0.0 }, histogram: Histogram { sample_count: 0, sample_sum: 0.0, bucket: [] }, timestamp_ms: 0 }
Metric { label: [LabelPair { name: "role_type", value: "B" }, LabelPair { name: "state", value: "connected" }], gauge: Gauge { value: 3.0 }, counter: Counter { value: 0.0 }, summary: Summary { sample_count: 0, sample_sum: 0.0, quantile: [] }, untyped: Untyped { value: 0.0 }, histogram: Histogram { sample_count: 0, sample_sum: 0.0, bucket: [] }, timestamp_ms: 0 }
--
```

```
# HELP INT_COUNTER An integer counter
# TYPE INT_COUNTER counter
INT_COUNTER 12
# HELP libra_network_peers Number of peers, and their associated state
# TYPE libra_network_peers gauge
libra_network_peers{role_type="A",state="connected"} 20
libra_network_peers{role_type="A",state="not-connected"} 2
libra_network_peers{role_type="B",state="connected"} 3
```

### DIFF
--- a/common/metrics/src/lib.rs
+++ b/common/metrics/src/lib.rs
@@ -67,6 +67,7 @@ pub use prometheus::{
 
 use anyhow::Result;
 use libra_logger::prelude::*;
+use once_cell::sync::Lazy;
 use prometheus::{proto::MetricType, Encoder, TextEncoder};
 use std::{
     collections::HashMap,
@@ -75,6 +76,15 @@ use std::{
     path::Path,
     thread, time,
 };
+
+pub static NUM_METRICS: Lazy<IntCounterVec> = Lazy::new(|| {
+    register_int_counter_vec!(
+        "libra_metrics",
+        "Number of metrics in certain states",
+        &["type"]
+    )
+    .unwrap()
+});
 
 fn get_metrics_file<P: AsRef<Path>>(dir_path: &P, file_name: &str) -> File {
     create_dir_all(dir_path).expect("Create metrics dir failed");
@@ -90,8 +100,39 @@ fn get_metrics_file<P: AsRef<Path>>(dir_path: &P, file_name: &str) -> File {
         .expect("Open metrics file failed")
 }
 
+pub fn gather_metrics() -> Vec<prometheus::proto::MetricFamily> {
+    let metric_families = prometheus::gather();
+    let mut total: i64 = 0;
+    let mut families_over_1000: i64 = 0;
+
+    // Take metrics of metric gathering so we know possible overhead of this process
+    for metric_family in &metric_families {
+        let family_count = metric_family.get_metric().len();
+        if family_count > 1000 {
+            families_over_1000 = families_over_1000.saturating_add(1);
+            let name = metric_family.get_name();
+            warn!(
+                count = family_count,
+                metric_family = name,
+                "Metric Family '{}' over 1000 dimensions '{}'",
+                name,
+                family_count
+            );
+        }
+        total = total.saturating_add(family_count as i64);
+    }
+
+    // These metrics will be reported on the next pull, rather than create a new family
+    NUM_METRICS.with_label_values(&["total"]).inc_by(total);
+    NUM_METRICS
+        .with_label_values(&["families_over_1000"])
+        .inc_by(families_over_1000);
+
+    metric_families
+}
+
 fn get_all_metrics_as_serialized_string() -> Result<Vec<u8>> {
-    let all_metrics = prometheus::gather();
+    let all_metrics = gather_metrics();
 
     let encoder = TextEncoder::new();
     let mut buffer = Vec::new();
@@ -102,7 +143,7 @@ fn get_all_metrics_as_serialized_string() -> Result<Vec<u8>> {
 pub fn get_all_metrics() -> HashMap<String, String> {
     // TODO: use an existing metric encoder (same as used by
     // prometheus/metric-server)
-    let all_metric_families = prometheus::gather();
+    let all_metric_families = gather_metrics();
     let mut all_metrics = HashMap::new();
     for metric_family in all_metric_families {
         let values: Vec<_> = match metric_family.get_field_type() {
--- a/common/metrics/src/metric_server.rs
+++ b/common/metrics/src/metric_server.rs
@@ -2,7 +2,8 @@
 // SPDX-License-Identifier: Apache-2.0
 
 use crate::{
-    json_encoder::JsonEncoder, json_metrics::get_json_metrics, public_metrics::PUBLIC_METRICS,
+    gather_metrics, json_encoder::JsonEncoder, json_metrics::get_json_metrics,
+    public_metrics::PUBLIC_METRICS, NUM_METRICS,
 };
 use futures::future;
 use hyper::{
@@ -18,12 +19,16 @@ use std::{
 use tokio::runtime;
 
 fn encode_metrics(encoder: impl Encoder, whitelist: &'static [&'static str]) -> Vec<u8> {
-    let mut metric_families = prometheus::gather();
+    let mut metric_families = gather_metrics();
     if !whitelist.is_empty() {
         metric_families = whitelist_metrics(metric_families, whitelist);
     }
     let mut buffer = vec![];
     encoder.encode(&metric_families, &mut buffer).unwrap();
+
+    NUM_METRICS
+        .with_label_values(&["total_bytes"])
+        .inc_by(buffer.len() as i64);
     buffer
 }
 


common/metrics/src/lib.rs
```diff
@@ -90,8 +92,28 @@ fn get_metrics_file<P: AsRef<Path>>(dir_path: &P, file_name: &str) -> File {
         .expect("Open metrics file failed")
 }
 
+pub fn gather_metrics() -> Vec<prometheus::proto::MetricFamily> {
```
ankushagarwal, line 95: Should we create a metric to track the total number of counters?

gregnazario: We can do that.  The metric will just magically disappear however, when we start dropping metrics :D

ankushagarwal:
> Limit the cardinality in any one counter, which means we should be able to limit the number of counters statically with an upper bound

I think that this is a solid option instead of the current approach of dropping metrics arbitrarily. Having a reasonable limit on the max number of timeseries a single family can export (eg: libra_network_peers) like 100 should be enough to ensure that we don't overload prometheus.

gregnazario:
> > Limit the cardinality in any one counter, which means we should be able to limit the number of counters statically with an upper bound
> 
> I think that this is a solid option instead of the current approach of dropping metrics arbitrarily. Having a reasonable limit on the max number of timeseries a single family can export (eg: libra_network_peers) like 100 should be enough to ensure that we don't overload prometheus.

I think the only problem is, some of our metrics may already be over that limit.  Realistically, I think we need to determine a more realistic upper bound.  I think 1000 counters shouldn't overload a system built for metrics.

common/metrics/src/lib.rs
gregnazario: Done, added a metric for the number seen, and the number kept.  And purposely keep those metrics.

common/metrics/src/lib.rs
```diff
@@ -90,8 +104,44 @@ fn get_metrics_file<P: AsRef<Path>>(dir_path: &P, file_name: &str) -> File {
...
+    let metric_families = prometheus::gather();
+    let mut kept_families: Vec<_> = Vec::new();
+    let mut total: usize = 0;
+    let mut kept_count: usize = 0;
+    for metric_family in metric_families {
+        let family_count = metric_family.get_metric().len();
+
```
ankushagarwal, line 114: How about something like this so that we can identify the problematic families from the logs:

```
        if family_count > 100 {
            error!(
                "Metrics: Metric Family {} is trying to export too many metrics: {}. Please investigate this as one of the labels might have a very high cardinality.",
                metric_family.get_name(),
                family_count,
            );
        }

```

davidiw:
> > Limit the cardinality in any one counter, which means we should be able to limit the number of counters statically with an upper bound
> 
> I think that this is a solid option instead of the current approach of dropping metrics arbitrarily. Having a reasonable limit on the max number of timeseries a single family can export (eg: libra_network_peers) like 100 should be enough to ensure that we don't overload prometheus.

I think no matter what we need maximum, but then we can do smarter elimination of data points but that can be done in another PR.

common/metrics/src/lib.rs
```diff
@@ -90,8 +104,44 @@ fn get_metrics_file<P: AsRef<Path>>(dir_path: &P, file_name: &str) -> File {
...
+    }
+
+    if kept_count < total {
+        error!(
+            "Metrics: Too many counters >{}.  Kept: {}, Dropped: {}",
+            MAX_COUNTERS,
+            kept_count,
+            total - kept_count
+        );
+    }
```
davidiw, line 131: can we separate this logic and have a test around it?

common/metrics/src/lib.rs
```diff
@@ -90,8 +104,44 @@ fn get_metrics_file<P: AsRef<Path>>(dir_path: &P, file_name: &str) -> File {
...
+
+    // These metrics will be reported on the next pull
+    NUM_METRICS
+        .with_label_values(&["kept"])
+        .set(kept_count as i64);
+
+    NUM_METRICS.with_label_values(&["total"]).set(total as i64);
```
davidiw, line 138: do we want to return all families for now and swap it over at some point in the near future? or do we just wanna go for it?

common/metrics/src/lib.rs
```diff
@@ -76,6 +77,19 @@ use std::{
...
+pub static NUM_METRICS: Lazy<IntGaugeVec> = Lazy::new(|| {
+    register_int_gauge_vec!(
+        NUM_METRICS_NAME,
+        "Number of metrics in certain states",
+        &["state"]
+    )
+    .unwrap()
+});
+
+const MAX_COUNTERS: usize = 1000;
```
davidiw, line 90: let's make this configurable, e.g., in the config/logging / metrics?

davidiw (commented):
good stuff man, good stuff

common/metrics/src/lib.rs
gregnazario: I just did some CT analysis, to see how we're doing.

We're at 800 currently with CT.  So, we can either limit at 1000 now, or let us build up our metrics from the different teams, and then switch.

Given we have 20 nodes we, only have room for... 10 more counters, without dimensions according to this estimation.  It seems to me we need to do some cutting.

common/metrics/src/lib.rs
gregnazario: Sure

common/metrics/src/lib.rs
gregnazario: I agree, this is useful, except that if we're using this metric and it goes from < 100 to > 100 it'll just disappear.

andll:
What's status of this PR? Is it pending review/or handling review comments/or?

gregnazario:
> What's status of this PR? Is it pending review/or handling review comments/or?

Been busy working on operations for networking, will come back to this afterwards and I need to handle some review comments

bmwill (approved):
Looks good, You might have a conflict with https://github.com/libra/libra/pull/6210 though

gregnazario:
> Looks good, You might have a conflict with #6210 though

I just have to beat it in then :)

/land

PR #6003, opened 2020-09-12
[network] handle storage read errors more gracefully during network address decryption

davidiw:
* If we cannot access storage, but have an existing set of keys, decrypt with them
* If we do not, raise the error higher and ultimately bail

### DIFF
--- a/config/management/network-address-encryption/src/lib.rs
+++ b/config/management/network-address-encryption/src/lib.rs
@@ -14,7 +14,7 @@ use libra_network_address::{
 use libra_secure_storage::{Error as StorageError, KVStorage, Storage};
 use move_core_types::account_address::AccountAddress;
 use serde::{Deserialize, Serialize};
-use std::collections::HashMap;
+use std::{collections::HashMap, sync::RwLock};
 use thiserror::Error;
 
 #[derive(Debug, Error)]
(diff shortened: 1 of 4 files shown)


config/management/network-address-encryption/src/lib.rs
```diff
@@ -126,10 +130,27 @@ impl Encryptor {
...
+                "Unable to read {} from storage: {}",
+                VALIDATOR_NETWORK_ADDRESS_KEYS,
+                err
+            );
+        }
+
+        match self.cached_keys.read().unwrap().as_ref() {
+            Some(keys) => Ok(keys.clone()),
+            None => result,
+        }
```
huitseeker, line 153:
```suggestion
        match result {
            OK(keys) => {
                *self.cached_keys.write().unwrap() = Some(keys.clone());
            }
            Err(err) => libra_logger::error!(
                "Unable to read {} from storage: {}",
                VALIDATOR_NETWORK_ADDRESS_KEYS,
                err
            ),
        }

        self.cached_keys.read().unwrap().map_or(result, |x| x.clone())
```

zekun000 (approved):

davidiw:
local test passed.... the logs seem useless, trying again...

/land

PR #6004, opened 2020-09-12
[Draft]

JoshLind:
This is draft PR. Please do not review.

## Motivation

This PR updates Safety Rules to avoid signing votes using the consensus key directly, and instead make all requests to the storage backend. It's a little rough (there's probably a few better ways to make these changes), but it's useful as a prototype for performance measurements to see how bad things will be.

Yes.

## Test Plan

All local smoke tests seem to pass... (some unit tests fail, but this seems to be an orthogonal issue related to error handling)

## Related PRs

None.

### DIFF
--- a/consensus/consensus-types/src/block.rs
+++ b/consensus/consensus-types/src/block.rs
@@ -207,6 +207,19 @@ impl Block {
         }
     }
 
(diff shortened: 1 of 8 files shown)


JoshLind:
Closing this out for now. We can return to it when the time is right.

PR #6006, opened 2020-09-13
Vault signing

mgorven:


davidiw:
@JoshLind / @mgorven do you want to bring this back to life?

JoshLind:
@davidiw, happy to do that. I think I had an old PR that I closed out. I'll bring it back up and see if we can get this going for the next cut 😄

davidiw:
landed in another PR

PR #6007, opened 2020-09-13
[vault] Perform writes with check-and-set

mgorven:
Record the latest secret version and specify when making writes. This
ensures that the secret data hasn't changed since it was read. This
doesn't enable any caching yet which leverages the CAS to avoid reads.

* If the secret version isn't known (because it hasn't been read or
  written yet) the write is just done without CAS.
* If the write fails because the version doesn't match, the caller is
  responsible for performing another read to update the version. LSR
  currently always reads before writing so this should be fine.
* There is no limit on the number of secrets whose versions are tracked;
  I'm assuming that the number of secrets stays small.

### DIFF
--- a/secure/storage/src/vault.rs
+++ b/secure/storage/src/vault.rs
@@ -12,7 +12,13 @@ use libra_crypto::{
 use libra_secure_time::{RealTimeService, TimeService};
 use libra_vault_client::{self as vault, Client};
 use serde::{de::DeserializeOwned, Serialize};
-use std::sync::atomic::{AtomicU64, Ordering};
+use std::{
+    collections::HashMap,
+    sync::{
+        atomic::{AtomicU64, Ordering},
+        RwLock,
+    },
+};
 
 const LIBRA_DEFAULT: &str = "libra_default";
 
@@ -28,6 +34,8 @@ pub struct VaultStorage {
     namespace: Option<String>,
     renew_ttl_secs: Option<u32>,
     next_renewal: AtomicU64,
+    use_cas: bool,
+    secret_versions: RwLock<HashMap<String, u32>>,
 }
 
 impl VaultStorage {
@@ -37,12 +45,15 @@ impl VaultStorage {
         namespace: Option<String>,
         certificate: Option<String>,
         renew_ttl_secs: Option<u32>,
+        use_cas: bool,
     ) -> Self {
         Self {
             client: Client::new(host, token, certificate),
             namespace,
             renew_ttl_secs,
             next_renewal: AtomicU64::new(0),
+            use_cas,
+            secret_versions: RwLock::new(HashMap::new()),
         }
     }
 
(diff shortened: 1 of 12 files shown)


secure/storage/src/vault.rs
```diff
@@ -223,18 +231,32 @@ impl KVStorage for VaultStorage {
...
+            .unwrap()
+            .insert(key.to_string(), resp.version);
         Ok(GetResponse { last_update, value })
     }
 
     fn set<T: Serialize>(&mut self, key: &str, value: T) -> Result<(), Error> {
         let secret = self.secret_name(key);
-        self.client()
-            .write_secret(&secret, key, &serde_json::to_value(&value)?)?;
+        let version = match self.secret_versions.read().unwrap().get(&key.to_string()) {
```
davidiw, line 243: &key.to_string :O
that seems a little weird...
yeah, you can just use the key:
https://github.com/libra/libra/blob/master/secure/storage/src/in_memory.rs#L46

secure/storage/src/vault.rs
```diff
@@ -223,18 +231,32 @@ impl KVStorage for VaultStorage {
...
+            Some(version) => Some(*version),
```
davidiw, line 244: I guess you want to make a copy here?
```self.secret_versions.read().unwrap().get(key).map(|v| *v)```

**Review by davidiw — COMMENTED** [review]
Wow... nice pr! My only concern is that we're entering into an unknown here so late. What message is spit out from vault if CAS is bad?

Should we make this configurable? It seems like we could add a config set option and if it is set to be done, we can either enable / disable updating the HashMap?

File: secure/storage/vault/src/lib.rs [hunk 3]
```diff
@@ -325,15 +325,30 @@ impl Client {
...
         let request = self
             .agent
             .put(&format!("{}/v1/secret/data/{}", self.host, secret));
-        let resp = self
-            .upgrade_request(request)
-            .send_json(json!({ "data": { key: value } }));
+        let resp = self.upgrade_request(request).send_json(payload);
 
-        process_generic_response(resp)
+        if resp.ok() {
```
JoshLind, line 346: nit: the only downside to this is that we should probably add a fuzzer for the response handling here. But overall it's not a big deal -- I can address it once we land.

mgorven:
> Wow... nice pr! My only concern is that we're entering into an unknown here so late.

Yeah there is some risk. I feel that there are significant benefits though (performance improvement from safely caching the safety_data, and pretty solid protection from twins) and that the code is small enough that we should be able to iron out any problems. If we make it configurable then worst case we disable it (if we add caching it should be dependent on this being enabled).

> What message is spit out from vault if CAS is bad?

    HTTP 400 Bad Request
    {"errors":["check-and-set parameter did not match the current version"]}

> Should we make this configurable? It seems like we could add a config set option and if it is set to be done, we can either enable / disable updating the HashMap?

An option in `VaultConfig`?

davidiw:
Yeah: https://github.com/libra/libra/blob/master/config/src/config/secure_backend_config.rs#L37

mgorven:
* Add config option to disable CAS
* Add tests for CAS

secure/storage/src/vault.rs
```diff
@@ -223,18 +234,37 @@ impl KVStorage for VaultStorage {
...
 
     fn set<T: Serialize>(&mut self, key: &str, value: T) -> Result<(), Error> {
         let secret = self.secret_name(key);
-        self.client()
-            .write_secret(&secret, key, &serde_json::to_value(&value)?)?;
+        let version = if self.use_cas {
+            self.secret_versions.read().unwrap().get(key).copied()
+        } else {
+            None
+        };
```
davidiw, line 252: I think you keep this and can get rid of the self.use_cas every where else

secure/storage/src/vault.rs
```diff
@@ -223,18 +234,37 @@ impl KVStorage for VaultStorage {
...
+        let new_version =
+            self.client()
+                .write_secret(&secret, key, &serde_json::to_value(&value)?, version)?;
+        if self.use_cas {
```
davidiw, line 256: remove if statement

secure/storage/src/vault.rs
```diff
@@ -223,18 +234,37 @@ impl KVStorage for VaultStorage {
         let resp = self.client().read_secret(&secret, key)?;
         let last_update = DateTime::parse_from_rfc3339(&resp.creation_time)?.timestamp() as u64;
         let value: T = serde_json::from_value(resp.value)?;
+        if self.use_cas {
```
davidiw, line 237: remove if statement

secure/storage/vault/src/lib.rs
davidiw: no more fuzzers here, they look ugly and make the code unmaintainable

davidiw (approved):
You've put this much work into it, so I will trust you to flip the switch and do what you like here :)

PR #6012, opened 2020-09-14
[vm] accept signers passed to execute_script in order rather than rev…

sblackshear:
…erse order

Previously, execute_script accepted args and type args in order, but signers in reverse order. This PR fixes that oddity.


dariorussi (approved):
this looks definitely reasonable to me but I'd appreciate if you could let @vgao1996 comment before submitting. Thanks!

dariorussi:
meanwhile please fix the problem with listing and whatever else is there, if anything else

vgao1996 (approved):
LGTM

PR #6013, opened 2020-09-14
[move-cli] datatest testing harness + some useful tests

sblackshear:
Introducing a basic datatest testing framework for the CLI. It expects directories that contain an `args.txt` file with arguments that the `move` binary understands (one set of arguments per line). The testing framework runs the commands, compares the result to the expected output, and runs `move clean`.

I added tests for each of the major subcommands (compile, run, view) and their arguments as well as tests for error explanations (both runtime and compile-time) and one fairly meaty Libra smoke test.

In the process, I found a few minor bugs/oddities that I fixed:
- Error explanation sometimes failed to find the relevant module when `commit` is false due to a bug in `resolve_function` (now fixed)
- The CLI create `empty move_src` and `move_data` directories in some cases. I fixed the former, but the latter is still a problem.
- Introduced some constants in lib.rs and moved constants from main.rs there
- Updated the README


dariorussi:
anybody looking at CLI care to comment on this?

bob-wilson:
For the record, this was closed accidentally and replaced by https://github.com/libra/libra/pull/6120

Issue #5791, opened 2020-08-27
missing test coverage for validator


bob-wilson:
We're missing test cases for a few things in the Move adapter validation checks:

- INVALID_GAS_SPECIFIER
- MISSING_DATA (from normalize_gas_price)
- CURRENCY_INFO_DOES_NOT_EXIST
- GAS_UNIT_PRICE_BELOW_MIN_BOUND (there is a proptest for this and coverage data shows it is covered, but we should have a specific test for this in verify_txn.rs)

There are also a few things that are tested but not where I would expect:
- SENDING_ACCOUNT_FROZEN: tested in tests/transaction_builder.rs
- TRANSACTION_EXPIRED: tested in feature_transaction_expiration.mvir

bob-wilson:
Dario, do you have thoughts on who should add tests for these?

bob-wilson:
I have discovered more tests inside move-lang/functional-tests/tests:

- SENDING_ACCOUNT_FROZEN: libra_account/freezing.move
- TRANSACTION_EXPIRED: translated_ir_tests/block/expired_transaction.move

I'd still like to reorganize those to put them all in one place, but at least the tests already exist.

bob-wilson:
...and more tests! vm-validator/src/unit_tests/vm_validator_test.rs is already an attempt to put tests for all validation errors in one place. This is good but it's still missing a few things:

- SEQUENCE_NUMBER_TOO_OLD (comment in the code explains this is due to a limitation of the testing infra)
- SEQUENCE_NUMBER_TOO_NEW (only relevant for execution, not validation)
- INSUFFICIENT_BALANCE_FOR_TRANSACTION_FEE
- GAS_UNIT_PRICE_BELOW_MIN_BOUND (commented out because the min bound is zero)
- BAD_CHAIN_ID

PR #6018, opened 2020-09-14
[language] add tests for txn validation, with fixes for gas currency

bob-wilson:
Add some missing tests for transaction validation, in both the unit tests for the vm-validator and in the e2e verify_txn.rs tests, which also check for consistency between validation & execution.

This includes some minor changes to validation for the gas currency specifier. The execution phase was not checking the gas currency specifier directly. A bad currency specifier would still cause a failure when running the prologue, but this led to a discrepancy with the validation phase. Besides sharing the same check in both phases, this changes to use the CURRENCY_INFO_DOES_NOT_EXIST status code for the case where the CurrencyInfo resource blob cannot be loaded. That
had previously been reported with a MISSING_DATA status, but aside from being less specific, using that status with VMStatus::Error leads to an assertion failure because it is only intended to be used for a runtime execution error.

Closes: #5791

Yes

## Test Plan

Ran the new tests along with existing ones.

### DIFF
--- a/language/libra-vm/src/libra_transaction_executor.rs
+++ b/language/libra-vm/src/libra_transaction_executor.rs
@@ -6,7 +6,7 @@ use crate::{
     data_cache::StateViewCache,
     errors::expect_only_successful_execution,
     libra_vm::{
-        charge_global_write_gas_usage, get_transaction_output,
+        charge_global_write_gas_usage, get_currency_info, get_transaction_output,
         txn_effects_to_writeset_and_events_cached, LibraVMImpl, LibraVMInternals,
     },
     system_module_names::*,
@@ -263,6 +263,11 @@ impl LibraVM {
             account_config::from_currency_code_string(txn.gas_currency_code())
                 .map_err(|_| VMStatus::Error(StatusCode::INVALID_GAS_SPECIFIER))
         );
+        // Check that the gas currency is valid. It is not used directly here but the
+        // results should be consistent with the checks performed for validation.
+        if let Err(err) = get_currency_info(&account_currency_symbol, remote_cache) {
+            return discard_error_vm_status(err);
+        }
         let result = match txn.payload() {
             TransactionPayload::Script(s) => self.execute_script(
                 remote_cache,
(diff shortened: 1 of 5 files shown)


dariorussi (approved):
this looks great to me. Thanks!

language/libra-vm/src/libra_transaction_executor.rs
```diff
@@ -263,6 +263,11 @@ impl LibraVM {
             account_config::from_currency_code_string(txn.gas_currency_code())
                 .map_err(|_| VMStatus::Error(StatusCode::INVALID_GAS_SPECIFIER))
         );
+        // Check that the gas currency is valid. It is not used directly here but the
+        // results should be consistent with the checks performed for validation.
+        if let Err(err) = get_currency_info(&account_currency_symbol, remote_cache) {
+            return discard_error_vm_status(err);
```
dariorussi, line 269: actually I am confused by this now... I checked but not clear and I figured I'd ask which is the easiest/best thing.
Is this going to discard the transaction? I believe we should charge the user for those mistakes, is this what we are doing?

dariorussi (commented):
I actually approved but not sure anymore, please answer this question first...
thanks!

language/libra-vm/src/libra_transaction_executor.rs
bob-wilson: Yes, this will discard the transaction. The prologue is always run without gas metering so no gas cost has been incurred at this point. This also matches what we have currently written in the VM specification: at the beginning of the execution phase, transactions are revalidated and discarded if they are invalid. The issue here was that this one validation check was missing when running validation from the execution phase. (Remember my earlier comment that it would be good to share the same implementation? 😉)

dariorussi: yeah I figured that was adding the missing one, I just find it strange we are not charging for those kind of errors. In fact I think we should.
But I guess it's all good, we can change it at any time and we are reasonably protected by the fact validation would discard those...
anyway, still accepted, so feel free to go in at any time

bob-wilson: FWIW, you can't charge for this specific error because the gas currency is invalid. More generally, in response to your comment about "validation would discard those", this *is* validation. The first step of the execution phase is to re-run validation. Except for sequence number changes, the results should always be the same as during the validation phase, so I wanted to make this change to keep that consistency. I can't think of any way you could hit this in practice.

PR #6019, opened 2020-09-14
[libra framework] specify tiered_mint

emmazzz:
## Motivation

This PR
- specifies `tiered_mint` transaction script and function in `LibraAccount`
- mocks out `try_record_nonce` function in `SlidingNonce` module

Yes


junkil-park (commented):
Mocking out `try_record_nonce` could be done using the `opaque` pragma as follows:
```
spec fun try_record_nonce {
    pragma verify = false;
    pragma opaque = true;
    ensures result == spec_try_record_nonce(account, seq_nonce);
}
spec define spec_try_record_nonce(account: signer, seq_nonce: u64): u64;
```

I wonder if there is a reason why the `intrinsic` way is preferred here?

emmazzz:
> Mocking out `try_record_nonce` could be done using the `opaque` pragma as follows:
> 
> ```
> spec fun try_record_nonce {
>     pragma verify = false;
>     pragma opaque = true;
>     ensures result == spec_try_record_nonce(account, seq_nonce);
> }
> spec define spec_try_record_nonce(account: signer, seq_nonce: u64): u64;
> ```
> 
> I wonder if there is a reason why the `intrinsic` way is preferred here?

Yes you are right. I'm not sure what is the preferred way here. I followed what was done for `Event` and `Vector`.

junkil-park (approved):

PR #6023, opened 2020-09-14
network: add per-peer connected metrics

bmwill:
This adds per-peer connected metrics so that we can easily see which peers a validator is connected to.


network/src/counters.rs
```diff
@@ -47,6 +48,23 @@ pub fn connections(network_context: &NetworkContext, origin: ConnectionOrigin) -
...
+    register_int_gauge_vec!(
+        "libra_network_peer_connected",
+        "Indicates if we are connected to a particular peer",
+        &["peer_id"]
+    )
+    .unwrap()
+});
+
+pub fn peer_connected(network_context: &NetworkContext, peer_id: &PeerId, v: i64) {
+    if network_context.role().is_validator() {
```
davidiw, line 61: excellent!

davidiw (approved):
Curious if / how did you confirm.

PR #6026, opened 2020-09-14
[trusted state] fix the ratchet when epoch_proof has more

zekun000:
## Motivation

The ratchet would fail when the latest_li is in the future epoch and epoch_change_proof doesn't forward that far and indicates more=true.

This commit fixes the issue by fallback to the epoch_change_li when the epoch_change_proof doesn't result in the same epoch as latest_li.

Y

## Test Plan

Added test.

## Related PRs

### DIFF
--- a/types/src/trusted_state.rs
+++ b/types/src/trusted_state.rs
@@ -7,7 +7,7 @@ use crate::{
     transaction::Version,
     waypoint::Waypoint,
 };
-use anyhow::{ensure, format_err, Result};
+use anyhow::{bail, ensure, format_err, Result};
 use std::{convert::TryFrom, sync::Arc};
 
 /// `TrustedState` keeps track of our latest trusted state, including the latest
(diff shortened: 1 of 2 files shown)


types/src/trusted_state.rs
```diff
@@ -94,17 +94,27 @@ impl TrustedState {
...
+                    latest_li.ledger_info()
+                } else {
+                    ensure!(
+                        latest_li == epoch_change_li
+                            || latest_li.ledger_info().epoch() > new_epoch
+                                && epoch_change_proof.more,
+                        "Invalid epoch in latest ledger info",
+                    );
+                    epoch_change_li.ledger_info()
+                });
```
davidiw, line 114: this isn't really readable.... the level of nesting here is :trollface:

davidiw: after reading this a dozen times I can see the logic is correct, but we should really break it up with a clear comment that you're moving into a partial state as you attempt to obtain a new set of li.

There are actually three states here...

the new li doesn't move us into a new epoch, okay, let' s just use it
the new li is in the new epoch, so consume the epoch change li
the new li is beyond the new epoch but there are more epochs to come, so consume the epoch change li
then we have the one error case, which is new li is beyond the the new epoch but we don't have more epochs coming, error

can we write this as a if else if else if else  that returns the appropriate li?

davidiw:
amazing find, now do something with this in safety rules :P

types/src/trusted_state.rs
phlip9: It looks like the key change here is `latest_li.ledger_info().epoch() > new_epoch && epoch_change_proof.more`, i.e., there was a gap and you need more epoch change li's to catch up, so just ratchet as far as you can. Actually, this makes me wonder why we even have `more`? Could just be an implicit property like "there is a gap" ==> "more".

Reading through how `libra-client` uses this also reminded me how we're not checking other proofs lol. 😭

zekun000: agree `more` is more like a hint. I don't know what's our story of verifying-api overall.

phlip9 (approved):

PR #6031, opened 2020-09-15
[Safety Rules] Force a panic if storage returns an Error::PermissionDenied error.

JoshLind:
## Motivation

Access tokens for secure storage can expire or be revoked. In order to ensure that operators are aware of this, we want components to fail fast when this occurs, so that tokens can be refreshed. This PR updates safety rules to panic when an Error::PermissionDenied is thrown by storage which indicates an expired, revoked or incorrect token. 

To achieve this, the PR offers 2 commits:
1.  Refactor some of the strings in vault tests and add a new test to verify that revoked/invalid tokens return Error::PermissionDenied. This requires adding support for revoking tokens from the vault client; this feature is only available for tests.
2. Add a panic to any place in safety rules that receives an Error::PermissionDenied error from storage. 

Yes.

## Test Plan

All tests pass locally, including the newly added one.

## Related PRs

None.

### DIFF
--- a/consensus/safety-rules/src/error.rs
+++ b/consensus/safety-rules/src/error.rs
@@ -51,6 +51,18 @@ impl From<libra_secure_net::Error> for Error {
 
 impl From<libra_secure_storage::Error> for Error {
     fn from(error: libra_secure_storage::Error) -> Self {
-        Self::SecureStorageError(error.to_string())
+        // If a storage error is thrown that indicates a permission failure, we
+        // want to panic immediately to alert an operator that something has gone
+        // wrong. For example, this error is thrown when a storage (e.g., vault)
+        // token has expired, so it makes sense to fail fast and require a token
+        // renewal!
+        if libra_secure_storage::Error::PermissionDenied == error {
+            panic!(
+                "A permission error was thrown: {:?}. Maybe the storage token needs to be renewed?",
+                error
+            );
+        } else {
+            Self::SecureStorageError(error.to_string())
+        }
     }
 }
--- a/secure/storage/src/tests/vault.rs
+++ b/secure/storage/src/tests/vault.rs
@@ -9,10 +9,27 @@ use crate::{
 use libra_crypto::{test_utils::TestLibraCrypto, Signature};
 use libra_vault_client::dev::{self, ROOT_TOKEN};
 
-/// VaultStorage test constants
+/// VaultStorage namespace constants
 const VAULT_NAMESPACE_1: &str = "namespace_1";
 const VAULT_NAMESPACE_2: &str = "namespace_2";
 const VAULT_NAMESPACE_3: &str = "namespace_3";
+
+/// VaultStorage KV data names
+const ANYONE: &str = "anyone";
+const ROOT: &str = "root";
+const PARTIAL: &str = "partial";
+const FULL: &str = "full";
+
+/// VaultStorage Transit key names
+const CRYPTO_KEY: &str = "crypto_key";
+
+/// VaultStorage role names
+const EXPORTER: &str = "exporter";
+const NOONE: &str = "noone";
+const READER: &str = "reader";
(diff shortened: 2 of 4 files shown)


secure/storage/vault/src/lib.rs
```diff
@@ -188,6 +188,21 @@ impl Client {
         process_token_renew_response(resp)
     }
 
+    pub fn revoke_token_self(&self) -> Result<(), Error> {
```
zekun000, line 191: test only?

zekun000 (approved):

consensus/safety-rules/src/error.rs
```diff
@@ -51,6 +51,18 @@ impl From<libra_secure_net::Error> for Error {
 
 impl From<libra_secure_storage::Error> for Error {
     fn from(error: libra_secure_storage::Error) -> Self {
-        Self::SecureStorageError(error.to_string())
+        // If a storage error is thrown that indicates a permission failure, we
+        // want to panic immediately to alert an operator that something has gone
+        // wrong. For example, this error is thrown when a storage (e.g., vault)
+        // token has expired, so it makes sense to fail fast and require a token
+        // renewal!
+        if let libra_secure_storage::Error::PermissionDenied = error {
```
davidiw, line 59: why isn't this an equivalence check? the if let here is kinda weird?

secure/storage/src/tests/vault.rs
```diff
@@ -9,11 +9,28 @@ use crate::{
...
 use libra_vault_client::dev::{self, ROOT_TOKEN};
 
-/// VaultStorage test constants
+/// VaultStorage namespace constants
 const VAULT_NAMESPACE_1: &str = "namespace_1";
 const VAULT_NAMESPACE_2: &str = "namespace_2";
 const VAULT_NAMESPACE_3: &str = "namespace_3";
 
+/// VaultStorage KV data names
+const ANYONE: &str = "anyone";
```
davidiw, line 18: why? these are only used in one test ... global constants should only be used when you want reuse. I don't want reuse :P

secure/storage/src/tests/vault.rs
```diff
@@ -100,146 +113,170 @@ fn test_vault_key_value_policies() {
...
+    let signature = signer_store.sign(CRYPTO_KEY, &message).unwrap();
     signature.verify(&message, &pubkey).unwrap_err();
     signature.verify(&message, &new_pubkey).unwrap();
 }
+
+fn test_vault_tokens() {
+    let mut storage = create_vault_with_namespace(None);
+
+    let partial = Policy::new(vec![Permission::new(
+        Identity::User(WRITER.into()),
```
davidiw, line 258: I would just make a distinct identity here...

secure/storage/vault/src/lib.rs
davidiw: nah, this is just a library, so no need

davidiw (commented):
looks good with two nits

consensus/safety-rules/src/error.rs
JoshLind: Aah, that works too 😄

secure/storage/src/tests/vault.rs
JoshLind: I can move the constants to be scoped only for the test, but either way, a constant is better than having repeated strings throughout the code 😄 Although, I also re-used PARTIAL for the token tests so it is kind of global.. I'll leave as is for now so I can land but if you have a better suggestion, I'll happily follow up in a PR!

secure/storage/src/tests/vault.rs
JoshLind: That works too, but I wanted to follow the pattern used by the other tests.. :/

JoshLind:
Thanks, all! 😄 
/land

PR #6032, opened 2020-09-15
[genesis] updates

davidiw:
This improves the genesis tooling and documentation so that we have a clean canonical write up the e2e flow for genesis.

I found two quirks while doing this:
layout is expected to be in common but the command let it be placed anywhere
we should use insert-waypoint instead of create-and-insert-waypoint (cc: @mgorven, this will impact your tooling)

### DIFF
--- a/config/management/genesis/README.md
+++ b/config/management/genesis/README.md
@@ -165,9 +165,10 @@ cargo run -p libra-genesis-tool -- \
 ```
 * Similarly, the association should publish a genesis waypoint, and the OP should insert it into their storage (using the management tool):
 ```
-cargo run -p libra-management-tool -- \
+cargo run -p libra-genesist-tool -- \
     insert-waypoint \
-    --config config_file.yaml
+    --config config_file.yaml \
+    --waypoint WAYPOINT
 ```
 * Perform a verify that ensures the local store maps to Genesis and Genesis maps
   to the waypoint:
--- a/config/management/genesis/README.md
+++ b/config/management/genesis/README.md
@@ -1,8 +1,9 @@
 # Libra Config Manager
 
-The `libra-genesis-tool` provides a tool for the genesis ceremony of the Libra blockchain. The functionality of the tool is dictated by the organization of nodes within the system:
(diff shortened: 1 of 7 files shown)


config/management/genesis/src/layout.rs
```diff
@@ -66,7 +66,7 @@ impl SetLayout {
             .config
             .load()?
             .override_shared_backend(&self.backend.shared_backend)?;
-        let mut storage = config.shared_backend();
+        let mut storage = config.shared_backend_with_namespace(constants::COMMON_NS.to_string());
```
mgorven, line 69: Will this cause an error if `namespace` is still specified for the backend (but ignored)?

davidiw: No, in fact, if folks use the config, the namespace will always be specified. It is because of the config, that I made this in this format instead of the "hoping" a user realizes this goes in common. This makes writing symmetric with reading (genesis).

The more I read your comment, it feels like you ask two disjoint questions:
* Will it cause an error: No
* Will it be ignored: Yes

mgorven (approved):

config/management/genesis/README.md
```diff
@@ -121,70 +129,58 @@ where each field maps to a role as described in this document.
...
+    --waypoint WAYPOINT
 ```
 * Perform a verify that ensures the local store maps to Genesis and Genesis maps
-  to the waypoint. (TBD)
+  to the waypoint:
+```
+cargo run -p libra-genesis-tool -- \
+    verify \
+    --config config_file.yaml \
+    --genesis_path PATH_TO_GENESIS
```
gregnazario, line 179: One thing that the operators pointed out earlier, is with runbooks to use variables e.g. `$PATH_TO_GENESIS` so you can just set the variable and copy and paste.  But, I think that's really personal preference at this point.

gregnazario (approved):

config/management/genesis/README.md
davidiw: interesting, I wasn't present for that, I'm not used to see that model, but I'll slap another commit on top since it doesn't really change the meaning!

davidiw:
@mimoo umm...

mimoo:
fail :D

PR #6034, opened 2020-09-15
[management] support the remainder of governance tasks

davidiw:
The purpose was to stress test the code and identify any missing pieces.

Features added:
* set-validator-operator
* create-validator
* create-validator-operator

Issues:
* set-validator-config fails if the validator is out of the set -- fixed by checking to see if they are in the set

Verified via some ugliness in libra-swarm:
* start
* grab one of the storage files
* make a config (see bottom, make sure it points to json server and proper storage)
* hack code to print out public keys for owner and operator keys in current namespace
* create a validator account and manually set `libra_root/owner` in local storage
* create an operator account and manually set `libra_root/operator` in local storage
* run set-validator-operator
* run set-validator-config
* run add-validator
* confirm validator-set (note this also kills libra since there are 2 validators but only 1 present)

```
chain_id: 4
json_server: "http://127.0.0.1:51928"
shared_backend:
    type: "on_disk_storage"
    path: "/Users/davidiw/libra/config/management/operational/storage"
    namespace: "libra_root"
validator_backend:
    type: "on_disk_storage"
    path: "/Users/davidiw/libra/config/management/operational/storage"
    namespace: "libra_root"
```

@mgorven, we'll be using this along with the genesis tooling to get things setup...


config/management/operational/src/governance.rs
```diff
@@ -2,11 +2,104 @@
...
+impl CreateValidator {
+    pub fn execute(self) -> Result<(TransactionContext, AccountAddress), Error> {
+        let config = self.input.config()?;
+        let key = self.input.key()?;
+        let client = JsonRpcClientWrapper::new(config.json_server.clone());
+
+        let seq_num = client.sequence_number(libra_root_address())?;
+        let auth_key = AuthenticationKey::ed25519(&key);
+        let account_address = auth_key.derived_address();
+        let script = transaction_builder::encode_create_validator_account_script(
```
JoshLind, line 64: This flow diverges from the one in genesis (where we also allow auth keys to be none in the case that the owner doesn't provide a key..). Do we want to support this functionality too here, or just leave it as is and use a throw away key?

config/management/operational/src/governance.rs
```diff
@@ -2,11 +2,104 @@
...
+}
+
+#[derive(Debug, StructOpt)]
+pub struct CreateValidatorOperator {
+    #[structopt(flatten)]
+    input: CreateAccount,
+}
+
+impl CreateValidatorOperator {
+    pub fn execute(self) -> Result<(TransactionContext, AccountAddress), Error> {
```
JoshLind, line 83: nit: it seems the only difference between this execute() call and the one for the CreateValidator struct is the script type (encode_create_validator_operator_account_script vs encode_create_validator_account_script). Ideally we should refactor into a common helper (e.g., on CreateAccount) to avoid duplication?

config/management/operational/src/governance.rs
```diff
@@ -80,21 +164,27 @@ impl RemoveValidator {
...
+
+fn build_and_submit_libra_root_transaction(
+    config: &Config,
+    seq_num: u64,
+    script: Script,
+) -> Result<TransactionContext, Error> {
+    let txn = build_raw_transaction(config.chain_id, libra_root_address(), seq_num, script);
+
+    let mut storage = config.validator_backend();
+    let signed_txn = storage.sign(LIBRA_ROOT_KEY, "remove-validator", txn)?;
```
JoshLind, line 186: nit: the script name "remove-validator" needs to be fixed.. Perhaps pass it directly into this function from the caller?

config/management/operational/src/owner.rs
```diff
@@ -0,0 +1,56 @@
...
+            config.chain_id,
+            owner_address,
+            client.sequence_number(owner_address)?,
+            transaction_builder::encode_set_validator_operator_script(
+                self.name.as_bytes().to_vec(),
+                self.account_address,
+            ),
+        );
+
+        println!(
```
JoshLind, line 44: nit: do we still want to keep these in here?

config/management/operational/src/validator_config.rs
```diff
@@ -55,32 +55,45 @@ impl SetValidatorConfig {
...
+            "/ip4/127.0.0.1/tcp/6180".parse().unwrap()
         };
 
         let fullnode_address = if let Some(fullnode_address) = self.fullnode_address {
             fullnode_address
+        } else if let Some(vc) = &validator_config {
+            strip_address(&vc.fullnode_network_address)
         } else {
-            strip_address(&validator_config.fullnode_network_address)
+            "/ip4/127.0.0.1/tcp/6180".parse().unwrap()
```
gregnazario, line 89: So, what happens when we do these loopback addresses?  For example, another peer tries to connect to this peer, and finds out that it's addresses are localhost, it would connect to itself, and then realize the peerId doesn't match?

config/management/operational/src/validator_config.rs
```diff
@@ -55,32 +55,45 @@ impl SetValidatorConfig {
...
+            None
+        };
 
         let validator_address = if let Some(validator_address) = self.validator_address {
             validator_address
+        } else if let Some(vc) = &validator_config {
+            strip_address(&vc.validator_network_address)
         } else {
-            strip_address(&validator_config.validator_network_address)
+            "/ip4/127.0.0.1/tcp/6180".parse().unwrap()
```
JoshLind, line 81: Can you document why we want these to be the defaults? Or is this related to the debugging you did locally?

config/management/operational/src/governance.rs
gregnazario, line 186: Do you want to sign it with "remove-validator" when you also add a validator?

JoshLind (approved):
Thanks @davidiw. 😄  I'll approve this now so that you can move fast for the drills if need be, but please see my comments.

config/management/operational/src/owner.rs
davidiw: hahahaha oops

config/management/operational/src/governance.rs
davidiw: oh yeah, I saw this and forgot to comeback .... problem of working too late at night...

config/management/operational/src/governance.rs
davidiw: fair enough

config/management/operational/src/governance.rs
davidiw: The problem is what is exposed via the public interfaces versus the admin scripts. This is the best we can do with public interfaces. It most certainly does not match what genesis will do whereas other systems will. We can follow up later to explore admin scripts to generate the account, but definitely not critical for e2e tests.

config/management/operational/src/governance.rs
davidiw: that's just a message for errors, I need to pass in a string here...

config/management/operational/src/validator_config.rs
davidiw: the better behavior is probably to raise an error... I'll update...

config/management/operational/src/validator_config.rs
davidiw: Did you and @gregnazario work together on reviewing this, rofl... I'll raise an error.

config/management/operational/src/governance.rs
JoshLind: Sure, that makes sense! Just wanted to bring it up to make sure we're aware of the divergence 😄

Issue #5708, opened 2020-08-21
LibraWriteSetManager::prologue needs to check ChainId and expiration time


bob-wilson:
The transaction prologue function for WriteSets is not currently checking the transaction's chain_id. We should fix that.

bob-wilson:
@runtian-zhou said he would look into this.

bob-wilson:
For consistency (at least), the WriteSet prologue should also check the transaction's expiration time.

bob-wilson:
Note that e2e-testsuite/src/tests/write_set.rs should have tests for both of these to check that they result in validation failures with a REJECTED_WRITE_SET status.

bob-wilson:
This was done in #6042

PR #5997, opened 2020-09-11
[move framework][move-prover] Some specs from LIP-6, refactoring of initialization in Genesis

DavidLDill:
Changes to a bunch of .move files related to LibraAccount and LibraSystem.

Added specification that every account has a role in LibraAccount.
Refactored Genesis and LibraAccount some to make this easier to prove,
and to clean up initialization.  Specifically, roles for libra root
and treasury compliance were added in Genesis separately from the
account creation functions.  So, I moved role creation into the account creation functions. Genesis was also creating the libra root and treasury compliance accounts directly, so I moved those into LibraAccount::initialize, which simplifies the Genesis code and eliminates some dependencies.

In various places, I renamed accounts to reflect their roles, if
known.  I also added some assertion checks in SlidingNonce, which was
getting errors from unchecked aborts.

## Motivation

(Write your motivation for proposed changes here.)

### Have you read the [Contributing Guidelines on pull requests](https://github.com/libra/libra/blob/master/CONTRIBUTING.md#pull-requests)?

(Write your answer here.)

## Test Plan

(Share your test plan here. If you changed code, please provide us with clear instructions for verifying that your changes work.)

## Related PRs

(If this PR adds or changes functionality, please take some time to update the docs at https://github.com/libra/website, and link to your PR here.)

PR #6042, opened 2020-09-15
[breaking][libra-framework]  Refactor writeset prologue and epilogue

runtian-zhou:
## Motivation

Refactored writeset prologue and epilogue by merging the LibraWriteSetManager module into LibraAccount. This will make the prologue/epilogue code more modular. Also added checks for chain_id and expiration time for WriteSet Transaction. 

This change is a breaking change.

Yes

## Test Plan

`cargo test`

## Related PRs

This is a less aggressive version of #5881


language/libra-vm/src/libra_transaction_executor.rs
```diff
@@ -443,22 +443,6 @@ impl LibraVM {
             return Ok((e, discard_error_output(StatusCode::REJECTED_WRITE_SET)));
         };
 
-        // Bump the sequence number of sender.
```
runtian-zhou, line 446: By refactoring the epilogue we can get rid of this piece of ugly code.

types/src/block_metadata.rs
```diff
@@ -69,7 +69,7 @@ impl BlockMetadata {
 }
 
 pub fn new_block_event_key() -> EventKey {
-    EventKey::new_from_address(&libra_root_address(), 20)
+    EventKey::new_from_address(&libra_root_address(), 21)
```
runtian-zhou, line 72: This is an unforunate consequnce of this change as we changed the order of how event handles are generated. Do we want to keep the same magic value?

language/stdlib/modules/LibraAccount.move
```diff
@@ -1161,20 +1225,29 @@ module LibraAccount {
...
+        writeset_sequence_number: u64
+    ) acquires LibraWriteSetManager, LibraAccount, Balance {
+        let t_ref = borrow_global_mut<LibraWriteSetManager>(CoreAddresses::LIBRA_ROOT_ADDRESS());
+
+        Event::emit_event<Self::UpgradeEvent>(
+            &mut t_ref.upgrade_events,
+            UpgradeEvent { writeset_payload },
+        );
+        // Currency code don't matter here as it won't be charged anyway.
+        epilogue<LBR::LBR>(lr_account, writeset_sequence_number, 0, 0, 0);
```
runtian-zhou, line 1242: The `0` uglyness still resides. Do we want to make clean separation of incrementing sequence number only? IMHO having multiple function will introduce extra borrow_global. Do we want to trade that for the code cleaness?

types/src/block_metadata.rs
runtian-zhou: Update: I got rid of this change by reorganizing the genesis intialization code, as this will also change the event key for a number of other system events (all mint related events).

dariorussi:
@bob-wilson @sblackshear @tzakian @tnowacki you all have looked at and talked about this quite a bit. 
If this has to go in, it would be great to get it in asap.
So if you could comment that would be great!
thanks

language/stdlib/modules/LibraAccount.move
```diff
@@ -1102,6 +1139,29 @@ module LibraAccount {
         )
     }
 
+    /// The prologue for WriteSet transaction
+    fun writeset_prologue(
+        sender: &signer,
+        writeset_sequence_number: u64,
```
bob-wilson, line 1145: Can we name these "txn_sequence_number" and "txn_public_key" for consistency with the other prologue functions? The account has a single sequence number shared for all kinds of transactions, right?

language/stdlib/modules/LibraAccount.move
```diff
@@ -1102,6 +1139,29 @@ module LibraAccount {
...
+        writeset_public_key: vector<u8>,
+        txn_expiration_time: u64,
+        chain_id: u8,
+    ) acquires LibraAccount, Balance {
+        assert(Roles::has_libra_root_role(sender), Errors::invalid_argument(PROLOGUE_INVALID_WRITESET_SENDER));
```
bob-wilson, line 1150: The previous version of this code had an assert for "sender == CoreAddresses::LIBRA_ROOT_ADDRESS()", as well as the has_libra_root_role assert here. Did you remove that intentionally?

language/stdlib/modules/LibraAccount.move
```diff
@@ -1121,6 +1181,28 @@ module LibraAccount {
...
+        check_sender<Token>(transaction_sender, max_transaction_fee, txn_sequence_number, txn_public_key);
+
+        assert(
+            LibraTimestamp::now_seconds() < txn_expiration_time_seconds,
+            Errors::invalid_argument(PROLOGUE_ETRANSACTION_EXPIRED)
+        );
+    }
+
+    /// Check if sender is able to send a transaction
+    public fun check_sender<Token>(
```
bob-wilson, line 1200: I don't see any reason to introduce this "check_sender" function and rearrange the various "prologue_common" checks as you've done here. It seems like this is just an artifact of the previous version of the code, and if so, I think you should leave out those changes.

language/stdlib/modules/LibraAccount.move
```diff
@@ -188,6 +201,25 @@ module LibraAccount {
         );
     }
 
+    public fun initialize_writeset_manager(
```
bob-wilson, line 204: The previous version of this code had a spec. Can that be pulled over from the old module as well?

language/stdlib/modules/LibraAccount.move
```diff
@@ -1217,20 +1289,29 @@ module LibraAccount {
             TransactionFee::pay_fee(Libra::withdraw(coin, transaction_fee_amount))
         }
     }
+
+    /// Epilogue for WriteSet trasnaction
+    fun writeset_epilogue(
+        lr_account: &signer,
+        writeset_payload: vector<u8>,
+        writeset_sequence_number: u64
```
bob-wilson, line 1297: I think this should be named "txn_sequence_number" for consistency

language/stdlib/modules/LibraAccount.move
bob-wilson: I think it's fine like this. Keep it simple....

language/stdlib/modules/LibraAccount.move
```diff
@@ -1102,6 +1139,29 @@ module LibraAccount {
         )
     }
 
+    /// The prologue for WriteSet transaction
+    fun writeset_prologue(
```
bob-wilson, line 1143: The code in the old module had a (very simple) spec for the prologue function. Should that be pulled over here as well?

bob-wilson (changes requested):

language/stdlib/modules/LibraAccount.move
runtian-zhou: Yes this is intentional. IIUC `sender == CoreAddresses::LIBRA_ROOT_ADDRESS()` is a hard coded version of `has_libra_root_role`. Am I doing this right @tzakian?

language/stdlib/modules/LibraAccount.move
runtian-zhou: My thought is that exposing this function would be easier for us to refactor in the future, as it rips off the account related part. We can get rid of this function for sure as well.

runtian-zhou:
Added back the specs from the old module. Renamed writeset_sequence_number to transaction_sequence_number

language/stdlib/modules/LibraAccount.move
tzakian: has_libra_root_role does not have an address-specific check (only assert_libra_root) IIRC. So I would add the address check here as well.

language/stdlib/modules/LibraAccount.move
```diff
@@ -1217,20 +1289,29 @@ module LibraAccount {
...
         }
     }
+
+    /// Epilogue for WriteSet trasnaction
+    fun writeset_epilogue(
+        lr_account: &signer,
+        writeset_payload: vector<u8>,
+        writeset_sequence_number: u64
+    ) acquires LibraWriteSetManager, LibraAccount, Balance {
+        let t_ref = borrow_global_mut<LibraWriteSetManager>(CoreAddresses::LIBRA_ROOT_ADDRESS());
```
tzakian, line 1306: nit: better variable name

language/stdlib/modules/LibraAccount.move
```diff
@@ -1217,20 +1289,29 @@ module LibraAccount {
...
+
+        Event::emit_event<Self::UpgradeEvent>(
```
tzakian, line 1308: Nit: no `Self::`

tzakian (commented):
Overall this looks good to me. A couple nits here and there.

The main thing I think we should consider here is combining the two singleton resources in this module (`LibraWriteSetManager` and `AccountOperationsCapability`). I would be in favor of combining them unless there are reasons that I'm not aware of.

tzakian:
cc @shazqadeer in case you're working on the prologue

language/stdlib/modules/LibraAccount.move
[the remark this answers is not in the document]
runtian-zhou: Yesss! There's a slight gotcha here: the order of how event handles are generated are depending on the order of calls in  the genesis module. If we merge those functions, the event key for a number of functions will be modified. The most important one of them is probably the EventKey for new_epoch_event will be off by one, which will be a "bigger" change. Some of the json-rpc tests also hard coded those event keys. That's why I keep the two resource separate and have two init functions so that the order of creating event handles could be perserved.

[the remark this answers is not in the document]
bob-wilson: Avoiding churn in the EventKeys seems worth it to have 2 init functions. My $0.02

language/stdlib/modules/LibraAccount.move
```diff
@@ -79,10 +80,18 @@ module LibraAccount {
         limits_cap: AccountLimitMutationCapability,
     }
 
+    /// A resource for the number of WriteSets that has been committed on chain.
```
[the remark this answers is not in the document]
bob-wilson: Certainly the comment should be fixed up, but at this point, I'd say that it would be best to minimize the change and keep it as a separate resource, just from the perspective of reducing risk and churn in the code. I could be convinced otherwise if you really think it is worth doing....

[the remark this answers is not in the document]
tzakian: Good question about churn, and I would agree if this weren't already a breaking change...

I guess my opinion here is that since it is a breaking change already that we should combine these since combining them is no more "breaking" than moving the LibraWriteSetManager resource which is being done in this PR. This will need to be a genesis reset either way.

language/stdlib/modules/LibraAccount.move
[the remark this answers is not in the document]
tzakian: There is already a breaking PR that changes the initialization sequence of events and changes these event keys (see Davids SEV post in libra core). It might be worthwhile to try and coordinate this with that PR, and then we could combine these and have only one break of the event keys?

language/stdlib/modules/LibraAccount.move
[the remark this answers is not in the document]
bob-wilson: I defer to your expertise but please be cautious....

runtian-zhou:
Talked to @tzakian over the genesis initialization seuqnece issue. It seems to us that we are hitting similar issue as #5997. So if both PR can land in the same release cycle then we should aim for merging the two resources/init function.

language/stdlib/modules/LibraAccount.move
```diff
@@ -1121,6 +1198,14 @@ module LibraAccount {
         // Check that the chain ID stored on-chain matches the chain ID specified by the transaction
         assert(ChainId::get() == chain_id, Errors::invalid_argument(PROLOGUE_EBAD_CHAIN_ID));
 
+        // Check that the gas calculation will not overflow
```
tnowacki, line 1201: Why was this check moved?

runtian-zhou: Actually I shouldn't move them. This was coming from my previous version of refactor where I had the gas related function outside of `prologue_common`

language/stdlib/modules/LibraAccount.move
```diff
@@ -1217,20 +1297,30 @@ module LibraAccount {
...
+        txn_sequence_number: u64
+    ) acquires LibraWriteSetManager, LibraAccount, Balance {
+        let writeset_events_ref = borrow_global_mut<LibraWriteSetManager>(CoreAddresses::LIBRA_ROOT_ADDRESS());
+
+        Event::emit_event<UpgradeEvent>(
+            &mut writeset_events_ref.upgrade_events,
+            UpgradeEvent { writeset_payload },
+        );
+        // Currency code don't matter here as it won't be charged anyway.
+        epilogue<LBR::LBR>(lr_account, txn_sequence_number, 0, 0, 0);
```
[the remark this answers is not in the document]
runtian-zhou: I'm a bit torn over here. On one hand I love this idea of having separate functions we can call to get the logics we want. On the other hand, I'm trying to minimize the code change and keep the code structure as it was (and try to merge it ASAP as soon as #5997 lands to leverage the fact that it is touching the genesis init order). I'm also wasn't sure how it might affect the prover side work. Do you have any strong preference over here? cc @tzakian @sblackshear @shazqadeer  for thoughts as well.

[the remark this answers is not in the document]
tnowacki: I feel like splitting it up might make the prover's life easier? beats me though
If its seems like work to convince a human that the checks aren't happening, I feel like it might be cumbersome to convey this in the spec as well, but I could be wrong

runtian-zhou:
Interestingly, when bundled this PR with #5997, it reverted the change to json-rpc tests as the orders are still preserved. However there's still changes done to new_epoch_event and new_block_event which still made this PR a "hard" breaking change.

language/stdlib/modules/LibraAccount.move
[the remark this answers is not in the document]
tzakian: I don't think splitting it would make the prover's life easier (but, it's one of those things where it's hard to determine exactly what will make it harder/easier until you start proving it). I think for now in the interest of getting the breaking change in for this release cycle I would like to hold off on this, and if we do decide to refactor it out we can later on and that shouldn't be a breaking change.

bob-wilson:
@tzakian It sounds like you're OK with the current version, is that right? I'm pretty happy with it now, and if we're going with the initialization change that you wanted, we should land this ASAP so it goes along with #5997

tnowacki:
@tzakian @bob-wilson Is the current version with what is checked in now? In that case I'm happy with that too. 

It looks like there were two checks added to the writeset prologue and the VM? Maybe we should just do those two changes as a smaller PR and get them in ASAP

bob-wilson:
> @tzakian @bob-wilson Is the current version with what is checked in now? In that case I'm happy with that too.

I'm not sure what you're asking. I was referring to the current contents of this PR. I think it's ready to land.

> 
> It looks like there were two checks added to the writeset prologue and the VM? Maybe we should just do those two changes as a smaller PR and get them in ASAP

Yes, this fixes #5708 by adding checks for the ChainID and txn expiration. I'd much prefer to just land this one. Otherwise, if this doesn't go in along with #5997, I think we should go back to a separate initialization function to preserve the event keys, and Tim really didn't like like that.

tzakian:
> @tzakian It sounds like you're OK with the current version, is that right? I'm pretty happy with it now, and if we're going with the initialization change that you wanted, we should land this ASAP so it goes along with #5997

Yep, I'm good with this going in as-is. Agreed on getting this in ASAP (need to land before tomorrow morning!)

bob-wilson (approved):

PR #6037, opened 2020-09-15
[vault] Switch to native-tls for ureq

mgorven:
`rustls` doesn't support verifying IP addresses in certificates, which
prevents Libra connecting directly to Vault in GCP and Azure. Switch to
`native_tls` to make this possible.

Test Plan: Ran `libra-genesis-tool owner-key` from Vault to Github.
Specified Vault address as "localhost" and "127.0.0.1".

PR #6043, opened 2020-09-15
[ci] Don't run dependency verifier workflow on bors-libra events

metajack:
@davidiw noticed that the dependency verifier action had a weird behavior in https://github.com/libra/libra/pull/6037. This happened because right before bors lands a PR it rebases and force pushes the new commit to the pull request. This re-triggers the dependency verifier flow, which is a little odd.

This change prevents running the workflow if the action was triggered by bors-libra.

### DIFF
--- a/.github/workflows/dep-summaries.yml
+++ b/.github/workflows/dep-summaries.yml
@@ -16,26 +16,31 @@ jobs:
     name: Annotate PRs with dependency summary changes
     steps:
       - name: checkout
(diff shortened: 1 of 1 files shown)


metajack:
There doesn't seem to be a simpler way to do this. GHA's do not support an early exit unfortunately, so I have to do the conditional test in every step.

bmwill (approved):
/land

PR #6047, opened 2020-09-16
[cluster-test][reconfiguration] expect epoch and add report text

zekun000:
## Motivation

1. assert the epoch number
2. emit some transactions
3. fix the fullnode problem which fails the health check, the fix also improves the latency of the experiment
4. there's a lot expired transactions which I suspect related to mempool retry

Y

## Test Plan

Experiment Result: Reconfiguration[val-0]: total epoch: 101 : 24 TPS, 46869 ms latency, 51150 ms p99 latency, (!) expired 9539 out of 14790 txns

## Related PRs

### DIFF
--- a/libra-node/src/main_node.rs
+++ b/libra-node/src/main_node.rs
@@ -149,7 +149,9 @@ pub fn setup_environment(node_config: &NodeConfig, logger: Option<Arc<Logger>>) 
     // consensus has to subscribe to ALL on-chain configs
     let (consensus_reconfig_subscription, consensus_reconfig_events) =
         gen_consensus_reconfig_subscription();
-    reconfig_subscriptions.push(consensus_reconfig_subscription);
+    if node_config.base.role.is_validator() {
+        reconfig_subscriptions.push(consensus_reconfig_subscription);
+    }
 
     // Gather all network configs into a single vector.
     // TODO:  consider explicitly encoding the role in the NetworkConfig
--- a/testsuite/cluster-test/src/cluster_swarm/templates/libra_node_spec_template.yaml
+++ b/testsuite/cluster-test/src/cluster_swarm/templates/libra_node_spec_template.yaml
@@ -56,11 +56,9 @@ spec:
       name: config
     env:
     - name: RUST_LOG
-      value: "warn"
+      value: "debug"
     - name: STRUCT_LOG_TCP_ADDR
       value: "127.0.0.1:5044"
-    - name: STRUCT_LOG_LEVEL
-      value: "debug"
     - name: RUST_BACKTRACE
       value: "1"
     - name: MY_POD_IP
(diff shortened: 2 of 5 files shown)


snakech:
the expire tx I suspect is because of fullnode does not catch up validator? We do have a fullnode health check but if node is added to affected list it will bypass that.

zekun000:
okay I think I understand why the transactions expired, I think mempool prioritizes reconfiguration txn, and it leads to no other txn can be committed in between.

and 100 epoch takes ~300s to commit so most txns expired in mempool.

with 10 epochs, it's able to commit all of them. 

Experiment Result: Reconfiguration[val-0]: total epoch: 10 : 179 TPS, 27173 ms latency, 28350 ms p99 latency, no expired txns

libra-node/src/main_node.rs
```diff
@@ -149,7 +149,9 @@ pub fn setup_environment(node_config: &NodeConfig, logger: Option<Arc<Logger>>)
     // consensus has to subscribe to ALL on-chain configs
     let (consensus_reconfig_subscription, consensus_reconfig_events) =
         gen_consensus_reconfig_subscription();
-    reconfig_subscriptions.push(consensus_reconfig_subscription);
+    if node_config.base.role.is_validator() {
+        reconfig_subscriptions.push(consensus_reconfig_subscription);
+    }
```
davidiw, line 154: fun

testsuite/cluster-test/src/cluster_swarm/templates/libra_node_spec_template.yaml
```diff
@@ -56,7 +56,7 @@ spec:
       name: config
     env:
     - name: RUST_LOG
-      value: "warn"
```
davidiw, line 59: do we want to commit this? why isn't this info at least :O

davidiw (approved):

testsuite/cluster-test/src/cluster_swarm/templates/libra_node_spec_template.yaml
andll: We should commit debug log level
It was the case before but was configured via `STRUCT_LOG_LEVEL`. Then we changed logging to only look at RUST_LOG, but did not update this spec.

Since we are here, @zekun000 do you mind also removing STRUCT_LOG_LEVEL two lines below?

andll (approved):

testsuite/cluster-test/src/experiments/reconfiguration_test.rs
```diff
@@ -141,24 +190,37 @@ impl Experiment for Reconfiguration {
...
     }
 }
 
 impl fmt::Display for Reconfiguration {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
         write!(
             f,
-            "Affected peer: {}, pod_name: {}, total epoch: {}",
-            self.affected_peer_id, self.affected_pod_name, self.count
+            "Reconfiguration[{}]: total epoch: {}",
```
andll, line 222: I think we probably don't need this `[{self.affected_pod_name}]`. It will appear in changelog, and probably will just pollute it.

PR #6050, opened 2020-09-16
[cluster test] add framework for load test experiment

christinacelee:
## Summary

Adding basic framework for per-component (mempool, state sync) load test in cluster test. This should support the following experiments/testing either in isolation or in various combinations:
- load test mempool
- load test state sync
- submit client txns (via tx-emitter)

There are many improvements that should be made:
- collect more data about load test (e.g. throughput, latency of traffic per component)
- right now for mempool, we submit empty broadcasts - txns should actually be populated, either 'garbage' txns or legitimate txns (perhaps using similar txn management logic like in tx-emitter)
- lots of code duplication for state sync/mempool load testing - we can probably address this using Traits, but for efficiency I'll leave this for future exploration
- possibly improve concurrency model of concurrent load-testing of >1 components
- measure stats of submitted traffic to client endpoints too (via tx-emitter), like throughput


testsuite/cluster-test/src/cluster_builder.rs
```diff
@@ -521,7 +521,7 @@ impl ClusterBuilder {
                         format!("/ip4/{}/tcp/{}", validator_nodes[i].internal_ip, 6180).as_str(),
                     )
                     .expect("Failed to parse network address"),
-                    NetworkAddress::from_str(format!("/ip4/{}/tcp/{}", fullnode_ip, 6180).as_str())
```
snakech, line 524: May I ask why do we change the port number?

snakech: I found there is a public port 6182 in fullnode yaml, is this the reason we wanna make them consistency?

snakech:
Does this PR still under progress as draft or ready to review?

testsuite/cluster-test/src/cluster_builder.rs
christinacelee: it's not only about consistency, this config must be set correctly for the `StubbedNode` to properly connect to LibraNet via fullnode in the public network (otherwise it will try to connect to the validator's address... which won't work)

snakech: I assume it's 3am now in Korean?..:)

testsuite/cluster-test/src/experiments/load_test.rs
```diff
@@ -0,0 +1,327 @@
...
+
+        Self {
+            network_runtime,
+            mempool_handle,
+            state_sync_handle,
+        }
+    }
+}
+
+async fn mempool_load_test(
```
snakech, line 244: I think if we'd better leave load experiment as a framework for load testing. And all these tasks for different components can be grouped into a mod (like health check mod and effects mod)to allow end users have an ability to construct their own load test.

But since now we only have ur commit to mempool and state sync. we can merge this pr and do refactory later when we have more components involve

testsuite/cluster-test/src/experiments/load_test.rs
```diff
@@ -0,0 +1,327 @@
...
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        write!(
+            f,
+            "Load test components: mempool: {}, state sync: {}, emit txns: {}",
+            self.mempool, self.state_sync, self.emit_txn,
+        )
+    }
+}
+
+// An actor that can participate in LibraNet
```
snakech, line 156: nit: I think we can move this struct and impl to a separate file under /cluster-test/ or any reasonable place

snakech (approved):
Overall LGTM. Could you please add your test plan, the command to run CT with new load experiment, and paste the test output/stats

christinacelee:
I don't print out any stats because I'm leaving general stats collection/publishing as a TODO for the future, this PR just ensures the experiment as a whole runs e2e.
Command: 
`./scripts/cti -T <image_tag> --run load_test -- [--mempool] [--state-sync] [--emit-tx] --duration <duration_in_secs>`

PR #6054, opened 2020-09-16
[gha][lbt] append a happy message when LBT passes

rustielin:
If LBT passes, add a line that says:

:tada: Land-blocking test passed! :ok_hand:

### DIFF
--- a/.github/workflows/land-blocking.yml
+++ b/.github/workflows/land-blocking.yml
@@ -296,6 +296,9 @@ jobs:
             }
             // Post test result on original pull request
             try {
+              if (!should_fail) {
+                body += "\n :tada: Land-blocking cluster test passed! :ok_hand:"
+              }
               await github.issues.createComment(
                   {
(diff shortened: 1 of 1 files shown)


.github/workflows/land-blocking.yml
```diff
@@ -296,6 +296,9 @@ jobs:
             }
             // Post test result on original pull request
             try {
+              if (!should_fail) {
+                body += "\n :tada: Land-blocking test passed! :ok_hand:"
```
andll, line 300: Should we change it to 'Cluster test passed'?

'Land blocking' is a very confusing name - unit tests are also land blocking

rustielin: It's in reference to the name of the GHA workflow, which is "Land-blocking Test". Actually now that you mention it it's a bit confusing... I'll change it to cluster-test

andll (approved):

PR #6055, opened 2020-09-16
[language][genesis] move genesis builder (alternative approach)

vgao1996:
A "less intrusive" approach to add the move effects and genesis builder libraries to be compared with #5969 and #5972. Currently not intended to be checked in.


libra-action:
```
target packages:
  A genesis-builder 0.1.0 (workspace, path 'language/tools/genesis-builder')
    * features: [none]
  A move-vm-txn-effect-converter 0.1.0 (workspace, path 'language/tools/move-vm-txn-effect-converter')
    * features: [none]
```

libra-action:
```
target packages:
  A genesis-builder 0.1.0 (workspace, path 'language/tools/genesis-builder')
    * features: [none]
  A move-vm-txn-effect-converter 0.1.0 (workspace, path 'language/tools/move-vm-txn-effect-converter')
    * features: [none]
```

language/tools/vm-genesis/src/lib.rs
```diff
@@ -90,14 +97,48 @@ pub fn encode_genesis_transaction(
...
 
-fn merge_txn_effects(
-    mut effects_1: TransactionEffects,
-    effects_2: TransactionEffects,
-) -> TransactionEffects {
-    effects_1.resources.extend(effects_2.resources);
-    effects_1.modules.extend(effects_2.modules);
-    effects_1.events.extend(effects_2.events);
-    effects_1
+fn convert_changeset_and_events(
```
vgao1996, line 100: This is the ugly part of this approach. I essentially had to make a copy of `txn_effects_to_writeset_and_events`.

vgao1996:
After comparing this with #5972, I think the old implementation is still a better approach and the benefits overweighs the risks. The deal breaker for me is that by trying to avoid changing the actual transaction execution flows at all, we create a division in what gets run in tests and what gets run in prod. I think this is very dangerous in the sense that it creates an illusion that the critical flows are adequately tested but in fact they are not, defeating the purpose of testing. (Genesis building, or more generally, effect/changeset manipulation will be a big part of future tests and tools.)

Also note that the original implementation itself is a reduction of complexity in terms of what the Libra layer needs to reason about, which from my opinion, could help make the system safer. Remember the wrong error code I discovered? It was easier to think about the invariants it needs to maintain if it doesn't gets leaked into an outer environment. 

Again I'm aware of risky nature of such changes at this time, but I think in this particular case, letting the changes through will actually allow us to cover them with more tests and offset the risks. If something goes wrong we should be able to notice them immediately.

@bob-wilson @tnowacki Let me know what you think!

bob-wilson:
I'm not sure about the relative risk of the two approaches. Can you elaborate on why we need to do either one right now? You mentioned reducing complexity and making it easier to write tests. Could those wait until after we branch for v1?

dariorussi:
I really question the premises here. Both the one here and the one in #5972 and related.
I am not sure `TransactionEffects` belongs to `move-vm` so overall I don't think we should invest in any of this until we have talked and sorted out what we want to do moving forward. And that is not something we can do now.
And that is just one of the issues I have with the overall change.
So to me, the point is not how risky this is, the point is continuing to invest in something that I am not sure it belongs to the VM. And something that we are better looking at when we review how the Move VM can and should be embedded

vgao1996:
@bob-wilson  Sorry for the late reply. Been reviewing these drafts and re-evaluating my own reasoning of the problems in the past few days. I think now I have a much clear grasp of the intuitions I had:
1. We need a separate Move testing flow, that is one without the Libra genesis, and with a set of transactions that's potentially different or trimmed down. There's also no prologue or epilogue or anything. It's very unfortunate that when we almost separated the Move VM from Libra, we didn't separate the tests.
2. This testing flow will need to use a minimal Move genesis, Imagine one with only `Vector` and a few other modules. On the other hand, there's also need for custom geneses and generating writeset transactions. This prompted the creation of the genesis builder library. Naturally, this library will need to return something that encodes the side effects produced by the Move VM. It can't be the Libra `WriteSet` as it involves `AccessPath`. `move_core_types::effects` was an attempt to address this issue. The resulting genesis builder library is easier to use and could be safer than the existing process in `vm_genesis`. We could even implement additional checks ensuring no resources published refer to a temporary module.
3. Now with the new genesis builder, it's natural to switch the existing code to it and IMO this is what made folks uncomfortable. There are two parts involved here:
  1. Make `vm_genesis` use the new genesis builder
  2. Make the move vm use `ChangeSet`

My original implementation did both and this PR does only 1. For comparison see my analysis above.

To answer your question on whether some of the infrastructure work on testing can be pushed back. My view on this is that it all comes down to the expected return based on how you allocate time between writing tests and improving the infra. In my opinion, the existing infra does have some serious (albeit not fatal) limitations, and while you can keep pushing it to its limit and focus solely on writing tests, I don't think it will give us the highest level of confidence for launch. I still think there's value in implementing some of the most needed improvements. Although I'm aware I'll need to be careful not to go too far in that direction.

@dariorussi  Regarding `TransactionEffects`, I totally agree with you that it doesn't seem right. It was a random thing I introduced during my VM refactoring and since I was viewing it as a temporary and final medium to pass back the information to the Libra layer, I made a few choices that are obviously dubious now. My current understanding is that there's a general need for effect manipulation on the Move level and `ChangeSet` was my attempt to address that. That means, I also don't think `TransactionEffects` belongs to the Move VM and instead I made `ChangeSet` a general Move concept that resides in `move_core_types`, but I agree this is just my view and we should definitely talk more about the future design of a more embeddable Move VM. 

Now with all being said, the more practical problem at our hands is how we should proceed now. Without the new effects API, I'll have to do exactly what you want to avoid: building more stuff on top of `TransactionEffects`. Meanwhile I get that the solution I proposed comes at a bad time and we don't have enough time or energy to discuss it (and the broader embedded VM issue) as thoroughly as we want. The way I see it is that `ChangeSet` is strictly better than `TransactionEffects` so I'd still hope we could get it in and improve/replace it later, but I'm also open to other solutions that could unblock my work on the Move testing flow.

vgao1996:
@bob-wilson @tnowacki I talked to @dariorussi and we agreed that now is not the best time to make changes to the critical paths, for the mental burden it places on everyone in the team -- even if you don't work on it directly, you'll still have to change the mental models and intuitions you've built up. Therefore I'm dropping this and the other two original PRs.

However it did came clear to us that having a separate Move testing flow that doesn't involve Libra is the right story to have for the VM and other components/tools on the Move layer. Therefore I'm salvage the non-critical parts of these PRs and putting them together into a new proposal #6186.

PR #6056, opened 2020-09-16
[short-hex-str] Add ShortHexStr stack-alloc'd hex-formatted str

phlip9:
+ Use it in `HashValue::short_str` and `AccountAddress::short_str` instead of returning `String` with `hex::encode`.

+ About 3-4x faster in some micro benchmarks (~50 ns/iter vs ~170 ns/iter).

+ Include a tiny hex encode routine that actually gets inlined and unrolled by the compiler.

+ We format into a stack-allocated `[u8; 8]` array so we can easily get a `&str` ref out. Other alternatives like formatting in `Display` from a borrow don't allow this without running through the fmt machinery + alloc'ing a temp.

+ Useful to avoid unnecessary allocs when e.g. bumping metrics or emitting logs that use a short `PeerId`/`AccountAddress`.

### DIFF
--- a/common/short-hex-str/src/lib.rs
+++ b/common/short-hex-str/src/lib.rs
@@ -1,0 +1,135 @@
+// Copyright (c) The Libra Core Contributors
+// SPDX-License-Identifier: Apache-2.0
+
+use mirai_annotations::debug_checked_precondition;
+use serde::{Serialize, Serializer};
+use std::{fmt, str};
+use thiserror::Error;
+
+/// An efficient container for formatting a byte slice as a hex-formatted string,
+/// stored on the stack.
+///
+/// Using `ShortHexStr` instead of `hex::encode` is about 3-4x faster on a recent
+/// MBP 2019 (~48 ns/iter vs ~170 ns/iter) in an artifical micro benchmark.
+#[derive(Ord, PartialOrd, Eq, PartialEq, Hash, Clone, Copy)]
+pub struct ShortHexStr([u8; ShortHexStr::LENGTH]);
+
+#[derive(Error, Debug)]
+#[error("Input bytes are too short")]
+pub struct InputTooShortError;
+
+impl ShortHexStr {
+    pub const SOURCE_LENGTH: usize = 4;
+    pub const LENGTH: usize = 2 * ShortHexStr::SOURCE_LENGTH;
+
+    /// Format a new `ShortHexStr` from a byte slice.
+    ///
+    /// Returns `Err(InputTooShortError)` if the input byte slice length is less
+    /// than `SOURCE_LENGTH` bytes.
+    pub fn try_from_bytes(src_bytes: &[u8]) -> Result<ShortHexStr, InputTooShortError> {
+        if src_bytes.len() >= ShortHexStr::SOURCE_LENGTH {
+            let src_short_bytes = &src_bytes[0..ShortHexStr::SOURCE_LENGTH];
+            let mut dest_bytes = [0u8; ShortHexStr::LENGTH];
+
+            // We include a tiny hex encode here instead of using the `hex` crate's
+            // `encode_to_slice`, since the compiler seems unable to inline across
+            // the crate boundary.
+            hex_encode(&src_short_bytes, &mut dest_bytes);
+            Ok(Self(dest_bytes))
+        } else {
+            Err(InputTooShortError)
+        }
+    }
+
+    pub fn as_str(&self) -> &str {
+        // We could also do str::from_utf8_unchecked here to avoid the unnecessary
+        // runtime check. Shaves ~6-7 ns/iter in a micro bench but the unsafe is
+        // probably not worth the hassle.
+        str::from_utf8(&self.0).expect(
+            "This can never fail since &self.0 will only ever contain the \
+             following characters: '0123456789abcdef', which are all valid \
+             ASCII characters and therefore all valid UTF-8",
+        )
+    }
+}
+
+impl fmt::Debug for ShortHexStr {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.write_str(self.as_str())
+    }
+}
+
+impl fmt::Display for ShortHexStr {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.write_str(self.as_str())
+    }
+}
+
+impl Serialize for ShortHexStr {
+    fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
+    where
+        S: Serializer,
+    {
+        serializer.serialize_str(self.as_str())
+    }
+}
+
+/// Maps a nibble to its corresponding hex-formatted ASCII character.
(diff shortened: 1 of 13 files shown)


libra-action:
```
target packages:
  A short-hex-str 0.1.0 (workspace, path 'common/short-hex-str')
    * features: [none]
```

libra-action:
```
target packages:
  A short-hex-str 0.1.0 (workspace, path 'common/short-hex-str')
    * features: [none]
```

libra-action:
```
target packages:
  A short-hex-str 0.1.0 (workspace, path 'common/short-hex-str')
    * features: [none]
```

crypto/crypto/src/hash.rs
```diff
@@ -256,9 +256,10 @@ impl HashValue {
         })
     }
 
-    /// Returns first SHORT_STRING_LENGTH bytes as String in hex
-    pub fn short_str(&self) -> String {
-        hex::encode(&self.hash[..SHORT_STRING_LENGTH])
+    /// Returns first 4 bytes as hex-formatted string
+    pub fn short_str(&self) -> ShortHexStr {
+        // This will never fail since HashValue::LENGTH >= 4
+        ShortHexStr::from_bytes_unchecked(&self.hash)
```
mimoo, line 262: what's the point for these short hex strings? It's only for debugging?

common/short-hex-str/src/lib.rs
```diff
@@ -0,0 +1,149 @@
...
+
+impl ShortHexStr {
+    pub const SOURCE_LENGTH: usize = 4;
+    pub const LENGTH: usize = 2 * ShortHexStr::SOURCE_LENGTH;
+
+    /// Format a new `ShortHexStr` from a byte slice.
+    ///
+    /// Returns `Err(InputTooShortError)` if the input byte slice length is less
+    /// than `SOURCE_LENGTH` bytes.
+    pub fn from_bytes(src_bytes: &[u8]) -> Result<ShortHexStr, InputTooShortError> {
```
mimoo, line 29: why not in a `TryFrom` implementation?

common/short-hex-str/src/lib.rs
```diff
@@ -0,0 +1,149 @@
...
+            Ok(ShortHexStr::from_bytes_unchecked(src_bytes))
+        } else {
+            Err(InputTooShortError)
+        }
+    }
+
+    /// Format a new `ShortHexStr` from a byte slice.
+    ///
+    /// Panics if the input byte slice length is less than `SOURCE_LENGTH` bytes.
+    pub fn from_bytes_unchecked(src_bytes: &[u8]) -> Self {
```
mimoo, line 40: instead of doing this, why not just return `0x0` if the source is too small?

common/short-hex-str/src/lib.rs
```diff
@@ -0,0 +1,149 @@
...
+        debug_checked_precondition!(src_bytes.len() >= ShortHexStr::SOURCE_LENGTH);
+
+        let src_short_bytes = &src_bytes[0..ShortHexStr::SOURCE_LENGTH];
+        let mut dest_bytes = [0u8; ShortHexStr::LENGTH];
+
+        // We include a tiny hex encode here instead of using the `hex` crate's
+        // `encode_to_slice`, since the compiler seems unable to inline across
+        // the crate boundary.
+        hex_encode(&src_short_bytes, &mut dest_bytes);
+        Self(dest_bytes)
```
mimoo, line 50: why store this in bytes when we're always going to use this as a string? (unless I'm missing smthg)

crypto/crypto/src/hash.rs
phlip9: Comment probably needs to be updated; we use the short version for most (all?) logs and metrics that contain a peer id or account address. The full 32 chars "b50dd697e24abe597e718ae7f68246dc" is kind of unwieldy compared to just "b50dd697". Makes it a lot easier for humans to read imo.

common/short-hex-str/src/lib.rs
phlip9: So `&str` is basically a `&[u8]` with an additional invariant that it must only contain valid UTF-8. Unfortunately, there's no way other way to represent an owned, stack-allocated string than a `[u8; N]` : (. The usual `String` type is always heap allocated; it's more-or-less a `Box<[u8]>` with some extra invariants.

So we have to store a `[u8; N]` and then cast its `&[u8]` to a `&str`. In `as_str()` you can see the `str::from_utf8(&self.0)` which just checks that the slice contains all valid UTF-8 and then casts the `&[u8]` to a `&str`. If we wanted to be extra (shave off 6-7 ns overhead lol) we could use `unsafe { str::from_utf8_unchecked(&self.0) }`, which literally just casts the refs with no check. Either way it's safe since the hex_encode will only ever emit valid UTF-8.

common/short-hex-str/src/lib.rs
phlip9: Is that really better? I feel like I would rather panic if someone misuses the API. For both cases `HashValue::short_str` and `AccountAddress::short_str` they will never ever panic b/c they have fixed size length >= 4. If someone else uses this, then the comment and fn name `_unchecked` usually make it pretty clear that the caller needs to maintain some invariant before calling.

bmwill, line 40: This is called _unchecked which implies that this should be marked as unsafe.

common/short-hex-str/src/lib.rs
bmwill: This is actually clearer IMO. Its fine to have a TryFrom impl too but this is much more explicit.

common/short-hex-str/src/lib.rs
phlip9: I think you're right. Changed to only `try_from_bytes(&[u8]) -> Result` and added `const_assert` to callers `HashValue::short_str` and `AccountAddress::short_str`.

libra-action:
```
target packages:
  A short-hex-str 0.1.0 (workspace, path 'common/short-hex-str')
    * features: [none]
```

libra-action:
```
target packages:
  A short-hex-str 0.1.0 (workspace, path 'common/short-hex-str')
    * features: [none]
```

libra-action:
```
target packages:
  A short-hex-str 0.1.0 (workspace, path 'common/short-hex-str')
    * features: [none]
```

common/short-hex-str/src/lib.rs
mimoo: the only thing I'm scared of is that we add something that can be triggered in prod to crash : o
what about returning 0x0 only in prod, and crashing in debug mode?

common/short-hex-str/src/lib.rs
mimoo: ah I see, makes sense!

common/short-hex-str/src/lib.rs
phlip9: fyi i just removed this method, now there's just `try_from_bytes`. the callers (`HashValue::short_str` and `AccountAddress::short_str`) not include a const_assert that checks it's valid to call

bmwill (approved):
Thanks for making a few tweaks. I think this is good to go now and should make things a little more performant!

PR #6058, opened 2020-09-16
[network] Print out connected peers state every minute

gregnazario:
### Overview

In order to have periodic state about which peers are connected, I've
added a sampling once a minute to print out the full state of the
network.

This is a test of the `sample` logic, and I know that we now have metrics for this information, but the whole picture as one may be useful.

### DIFF
--- a/common/logger/src/lib.rs
+++ b/common/logger/src/lib.rs
@@ -72,7 +72,12 @@
 
 pub mod prelude {
     pub use crate::{
-        debug, error, event, info, libra_logger::FileWriter, security::SecurityEvent, trace, warn,
+        debug, error, event, info,
+        libra_logger::FileWriter,
+        sample,
+        sample::{SampleRate, Sampling},
+        security::SecurityEvent,
+        trace, warn,
     };
 }
 pub mod json_log;
--- a/network/src/peer_manager/mod.rs
+++ b/network/src/peer_manager/mod.rs
@@ -371,6 +371,25 @@ where
             .set(outbound as i64);
     }
 
+    fn sample_connected_peers(&self) {
+        // Sample final state at most once a minute, ensuring consistent ordering
+        sample!(
+            SampleRate::Duration(Duration::from_secs(60)),
+            info!(
+                NetworkSchema::new(&self.network_context),
+                peers = (&self.active_peers)
+                    .iter()
+                    .map(|(_, (conn_metadata, _))| (
+                        conn_metadata.remote_peer_id,
+                        conn_metadata.origin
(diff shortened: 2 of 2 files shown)


network/src/connectivity_manager/mod.rs
```diff
@@ -493,6 +506,16 @@ where
         // Dial peers which are eligible but are neither connected nor queued for dialing in the
         // future.
         self.dial_eligible_peers(pending_dials).await;
+
+        // Sample final state once a minute
+        sample!(
+            SampleRate::Duration(Duration::from_secs(60)),
```
davidiw, line 512: should this be in peer manager instead? or even a dedicated unit that just listens to all incoming events and sets up a dedicated 60 second callback?

common/logger/src/lib.rs
```diff
@@ -72,7 +72,10 @@
 
 pub mod prelude {
     pub use crate::{
-        debug, error, event, info, security::SecurityEvent, trace, warn, StructuredLogEntry,
+        debug, error, event, info, sample,
+        sample::{SampleRate, Sampling},
+        security::SecurityEvent,
+        trace, warn, Schema, StructuredLogEntry,
```
bmwill, line 78: You're probably going to hit some conflicts with this line being edited in a couple different PRs

network/src/connectivity_manager/mod.rs
```diff
@@ -493,6 +506,16 @@ where
         // Dial peers which are eligible but are neither connected nor queued for dialing in the
         // future.
         self.dial_eligible_peers(pending_dials).await;
+
+        // Sample final state once a minute
```
bmwill, line 510: This comment really should read "at most once a minute" because technically if this line of code isn't executed but every 5 minutes then this will only log once every 5 minutes. I hope that made sense.

bmwill (approved):

network/src/connectivity_manager/mod.rs
gregnazario: I put it here, so we wouldn't have to deal with logging it on a non-permissioned network.  But, maybe we want that?

gregnazario: Done.

common/logger/src/lib.rs
gregnazario: Yeah, it's expected, but i'll deal with it as it goes.

network/src/connectivity_manager/mod.rs
gregnazario: Fixed

PR #6060, opened 2020-09-16
[logging] Remove structured log entry

gregnazario:
### Overview
Removes the previous StructuredLogEntry, moves Security logs, tracing, and a config log all to the new logging system!  Figured I'd finish the cleanup of the logs now to remove this mode.

I've been trying to keep naming of certain types consistent, and if they would conflict to have a different name.

### DIFF
--- a/consensus/src/epoch_manager.rs
+++ b/consensus/src/epoch_manager.rs
@@ -399,10 +399,9 @@ impl EpochManager {
                 .map_err(|err| {
                     error!(
                         SecurityEvent::ConsensusInvalidMessage,
-                        StructuredLogEntry::default()
-                            .data("from_peer", &peer_id)
-                            .data_display("error", &err)
-                            .data("event", &unverified_event)
+                        remote_peer = peer_id,
+                        error = err.to_string(),
+                        unverified_event = unverified_event
                     );
                     err
                 })?;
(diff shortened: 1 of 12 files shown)


consensus/src/epoch_manager.rs
```diff
@@ -399,10 +399,9 @@ impl EpochManager {
                 .map_err(|err| {
                     error!(
                         SecurityEvent::ConsensusInvalidMessage,
-                        StructuredLogEntry::default()
-                            .data("from_peer", &peer_id)
-                            .data_display("error", &err)
-                            .data("event", &unverified_event)
+                        remote_peer = peer_id,
+                        error = err.to_string(),
```
bmwill, line 403: I suppose I could have done this before :D We could add some support for the macros to select if you want to do debug, serde, or display based on a sigil

bmwill (approved):
This is awesome! :D

consensus/src/epoch_manager.rs
gregnazario: Yeah, I think we can do that in the future, like tracing, I've had to handle this in a couple places.  But, right now just been churning through and getting it done

bmwill: 100% agreed

PR #6061, opened 2020-09-16
Make it easy to run a simple testnet

davidiw:
This PR introduces several commits that make it easy (easier?) to run a single test network:

* First, make the genesis config-build deterministic, this is a common request and just makes it easier to work with the tooling. It also makes it easier to support resumption of an existing single test node network.
* Second, add a slight bit of libra-swarm code to libra-node to support a single validator network
* Next up, was to tweak some other components to allow for using config defaults and specifying the config path for this test layer... this let's us have well defined behavior within docker
* Finally, we introduce a docker container that bundles this all together... of course the current image it points to won't work until it gets built with the first three commits, but a dev can build the docker container locally in the mean time.

I'll try to figure out how to make a compose that unifies starting a faucet in addition... but no promises.


libra-action:
```
target packages:
  M libra-genesis-tool 0.1.0 (initial, path 'config/management/genesis')
    * added features: testing
```

libra-action:
```
target packages:
  M libra-genesis-tool 0.1.0 (initial, path 'config/management/genesis')
    * added features: testing
```

ankushagarwal:
Thank you ❤️

Haven't read the code. Any instructions on how I can run this locally?

dmitri-perelman:
Wow! Thank you!

davidiw:
The libra-node takes in a --test and prints out all the yummies that you need to connect to it

If you want to use the docker-compose, I'll add a little more comment into that file, but right now we would need to build it from source.

I'll add a readme.md to the compose directory, but effectively you need docker and docker compose installed (the latter is installed by default on mac and windows). Open up your favorite cli, cd into the appropriate compose directory, and run `docker-compose up`

libra-action:
```
target packages:
  M libra-genesis-tool 0.1.0 (initial, path 'config/management/genesis')
    * added features: testing
```

libra-action:
```
target packages:
  M libra-genesis-tool 0.1.0 (initial, path 'config/management/genesis')
    * added features: testing
```

common/logger/src/lib.rs
```diff
@@ -72,7 +72,8 @@
 
 pub mod prelude {
     pub use crate::{
-        debug, error, event, info, security::SecurityEvent, trace, warn, StructuredLogEntry,
+        debug, error, event, info, libra_logger::FileWriter, security::SecurityEvent, trace, warn,
```
gregnazario, line 75: Nit: I don't think we need to provide the FileWriter to all uses of the prelude.  But, nbd

common/logger/src/libra_logger.rs
```diff
@@ -304,6 +304,38 @@ impl Writer for StderrWriter {
...
+    /// Write to file
+    fn write(&self, log: String) {
+        let result = self
+            .log_file
+            .write()
+            .unwrap()
+            .write_all(format!("{}\n", log).as_bytes());
+        if let Err(err) = result {
+            eprintln!("Unable to write to log file: {}", err.to_string());
+        }
```
gregnazario, line 335: If only the writing on TCP could be this simple!

libra-node/src/main_node.rs
```diff
@@ -97,10 +97,15 @@ pub fn setup_environment(node_config: &NodeConfig, logger: Option<Arc<Logger>>)
...
         .thread_name(|index| format!("rayon-global-{}", index))
-        .build_global()
-        .expect("Building rayon global thread pool should work.");
+        .build_global();
+    // If it was already initialized, move along. There is no access to the internal error.
+    if let Err(err) = &result {
+        if err.to_string() != "The global thread pool has already been initialized." {
+            result.expect("Building rayon global thread pool should work.");
+        }
+    }
```
gregnazario, line 108: It's always sad when we have to do something like this :(. But it's effective

gregnazario (approved):
I guess ConfigBuilder never dies :)

libra-node/src/main_node.rs
davidiw: I actually want to delete this in another PR, I don't think this has any value :P, maybe I can convince you or @bmwill to stamp that. There's no value in specifying a name for Rayon afaiu. That comment above doesn't really sell it to me.

common/logger/src/libra_logger.rs
davidiw: hahaha

common/logger/src/lib.rs
davidiw: I'd have to do some wonky stuff to export it. I think the right approach is to probably move the writers to their own file and allow that to be exported.

davidiw:
> I guess ConfigBuilder never dies :)

This is "our" config-builder based upon genesis. I have one more PR coming that will make it easy to add Faucet.

At which point!!! I am deleting the hell out of config-builder <3 (config/config-builder)

Thanks for the timely review!

common/logger/src/libra_logger.rs
```diff
@@ -304,6 +304,38 @@ impl Writer for StderrWriter {
...
+}
+
+impl Writer for FileWriter {
+    /// Write to file
+    fn write(&self, log: String) {
+        let result = self
+            .log_file
+            .write()
+            .unwrap()
+            .write_all(format!("{}\n", log).as_bytes());
```
bmwill, line 332:
```suggestion
        let result = self
            .log_file
            .write()
            .unwrap()
            .write_all(log.as_bytes());
```

or

```suggestion
        let result = writeln!(self.log_file.write().unwrap(), "{}", log);
```

To avoid the allocation

common/logger/src/libra_logger.rs
```diff
@@ -304,6 +304,38 @@ impl Writer for StderrWriter {
     }
 }
 
+/// A struct for writing logs to a file
+pub struct FileWriter {
```
bmwill, line 308: What's the rationale for writing to a file and not just to stderr?

config/src/config/mod.rs
```diff
@@ -325,21 +325,18 @@ impl NodeConfig {
         self.test = Some(test);
     }
 
-    #[cfg(any(test, feature = "fuzzing"))]
```
bmwill, line 328: Can we remove these all?! :D

libra-node/src/main_node.rs
bmwill: Why does this need to change?

docker/compose/validator-testnet/.gitattributes
```diff
@@ -0,0 +1,3 @@
+genesis.blob -whitespace
+libra_root_key -whitespace
+waypoint.txt -whitespace
```
bmwill, line 3: Given the other PR i put out you should remove these I think

bmwill: Also why do we need to checkin empty files?

common/logger/src/libra_logger.rs
davidiw: I figured I was missing something...

common/logger/src/libra_logger.rs
davidiw: ahh, I put a comment in the main node code, but I'm trying to eliminate more documentation and just making things easy. when we run the test mode, it prints out all the useful information to get connected to it... I don't think most users are aware that what is visible on a terminal can be either err or out. so ... let's just put it into a file and reduce educational burden.

config/src/config/mod.rs
davidiw: I'm now leaning toward the opinion that excessive feature exclusion is not really worth the headache....

common/logger/src/libra_logger.rs
bmwill: Honestly the interface as it is right now should be tweaked to avoid an allocation but that's not your problem (its mine)

docker/compose/validator-testnet/.gitattributes
davidiw: the way the bind mount for docker works is that if these do not exist, the container won't start. the user would then need to create them... I'm probably being too nice :P

bmwill: I mean if these don't exist maybe you don't want docker to start? idk :D

common/logger/src/libra_logger.rs
bmwill: fair enough

docker/compose/validator-testnet/.gitattributes
davidiw: 🤡

libra-node/src/main_node.rs
davidiw: it is implicitly initialized during genesis creation

davidiw:
/land

I hope this goes through!!!!

libra-action:
```
target packages:
  M libra-genesis-tool 0.1.0 (initial, path 'config/management/genesis')
    * added features: testing
```

libra-action:
```
target packages:
  M libra-genesis-tool 0.1.0 (initial, path 'config/management/genesis')
    * added features: testing
```

libra-action:
```
target packages:
  M libra-genesis-tool 0.1.0 (initial, path 'config/management/genesis')
    * added features: testing
```

PR #6064, opened 2020-09-16
[network] Add NetworkContext to onchain discovery & RPC

gregnazario:
### Overview
Basically, while going through all the logging, I found that on chain discovery had configuration changes created for a specific network instance, but wasn't logging the `NetworkContext` for it.  So, I've added it to keep it in sync with the rest of the network logs.


bmwill (approved):
lgtm But maybe wait a bit so someone from network to double check

davidiw (approved):
gogogogo

PR #6065, opened 2020-09-17
[libra framework] Update comments in LBR module, use `create_from_raw`

tzakian:
Updates LBR docs to be clearer around representation of ratios for the LBR.

Also updates the coin ratios to use `FixedPoint32::create_from_raw` instead of `create_from_rational`.


language/stdlib/modules/LBR.move
```diff
@@ -117,11 +118,13 @@ module LBR {
         AccountLimits::publish_unrestricted_limits<LBR>(lr_account);
         let preburn_cap = Libra::create_preburn<LBR>(tc_account);
         let coin1 = ReserveComponent<Coin1> {
-            ratio: FixedPoint32::create_from_rational(1, 2),
+            // 1/2
```
bob-wilson, line 121: It might be helpful to expand that comment to something like: "2^31 = 1/2"
Personally, I would also be inclined to position the comment on the same line right after the number, but that part is up to you.

bob-wilson (approved):

PR #6066, opened 2020-09-17
logger: make structured data output deterministic

bmwill:


JoshLind (approved):
Thanks @bmwill! LGTM. 😄

If the need ultimately arises, I guess we could also write some form of HashMap/BTreeMap wrapper that provides sorting/display by the order of insertion. Although I think this will be good enough for now!

davidiw (commented):
BTreeMaps everywhere!!!

bmwill:
/land

We wouldn't need to implement anything ourselves since there's already an [`IndexMap`](https://docs.rs/indexmap/1.6.0/indexmap/) crate we could use.

PR #6068, opened 2020-09-17
[cleanup] cleanup deprecated counters in storage and execution

lightmark:
## Motivation

migrated to new counters. These mortals can leave the world.


davidiw (approved):
oh, I think this would have alleviated some of the headache with generating a waypoint and genesis in the same pass that config-builder experienced :P

/land

lightmark:
@davidiw why this helps.

PR #6069, opened 2020-09-17
[executor][metrics] add error counter

lightmark:
## Motivation

add a counter for error only.

### DIFF
--- a/execution/executor/src/lib.rs
+++ b/execution/executor/src/lib.rs
@@ -19,9 +19,10 @@ pub mod db_bootstrapper;
 use crate::{
     logging::{LogEntry, LogSchema},
     metrics::{
-        LIBRA_EXECUTOR_COMMIT_BLOCKS_SECONDS, LIBRA_EXECUTOR_EXECUTE_BLOCK_SECONDS,
-        LIBRA_EXECUTOR_SAVE_TRANSACTIONS_SECONDS, LIBRA_EXECUTOR_TRANSACTIONS_SAVED,
-        LIBRA_EXECUTOR_VM_EXECUTE_BLOCK_SECONDS, LIBRA_EXECUTOR_VM_EXECUTE_CHUNK_SECONDS,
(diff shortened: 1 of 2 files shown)


execution/executor/src/lib.rs
```diff
@@ -19,9 +19,10 @@ pub mod db_bootstrapper;
 use crate::{
     logging::{LogEntry, LogSchema},
     metrics::{
-        LIBRA_EXECUTOR_COMMIT_BLOCKS_SECONDS, LIBRA_EXECUTOR_EXECUTE_BLOCK_SECONDS,
-        LIBRA_EXECUTOR_SAVE_TRANSACTIONS_SECONDS, LIBRA_EXECUTOR_TRANSACTIONS_SAVED,
-        LIBRA_EXECUTOR_VM_EXECUTE_BLOCK_SECONDS, LIBRA_EXECUTOR_VM_EXECUTE_CHUNK_SECONDS,
+        LIBRA_EXECUTOR_COMMIT_BLOCKS_SECONDS, LIBRA_EXECUTOR_ERRORS,
+        LIBRA_EXECUTOR_EXECUTE_BLOCK_SECONDS, LIBRA_EXECUTOR_SAVE_TRANSACTIONS_SECONDS,
+        LIBRA_EXECUTOR_TRANSACTIONS_SAVED, LIBRA_EXECUTOR_VM_EXECUTE_BLOCK_SECONDS,
+        LIBRA_EXECUTOR_VM_EXECUTE_CHUNK_SECONDS,
```
davidiw, line 25: you should really consider making some of these vector metrics with labels

davidiw (approved):

execution/executor/src/lib.rs
lightmark: these are still manageable. labels may reduce it from 7 to 4 requiring update to dashboards. If it keeps mounting I would do that.

PR #6070, opened 2020-09-17
[Key Manager] Improve errors and logging/metrics

JoshLind:
## Motivation

This PR updates the errors and logging/metrics for the key manager after seeing it running for a period of time today. It offers two commits:
1. First, we improve the error outputs and fix the error string for Error::MissingAccountAddress (the string should report the correct account being looked up).
2. Second, we add logging and metrics to help differentiate between different cases when the key manager reports that no action (Action::NoAction) is required. This is required to tell the difference between when a rotation transaction has not been executed yet, and when the consensus key is fresh enough that we don't need to rotate.

Yes.

## Test Plan

Running the key manager locally seems to work and produce the correct logging/metrics and responses.

## Related PRs

None.

### DIFF
--- a/secure/key-manager/src/lib.rs
+++ b/secure/key-manager/src/lib.rs
@@ -60,18 +60,20 @@ pub enum Action {
 #[allow(clippy::large_enum_variant)]
 #[derive(Debug, Error, PartialEq)]
 pub enum Error {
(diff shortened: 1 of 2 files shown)


davidiw (approved):
looks good!

PR #6071, opened 2020-09-17
[state sync] log instead of return error on reconfig notification failure

christinacelee:
## Motivation

On failure to publish reconfig subscription, returning error causes entire process_chunk_response flow to fail. To address this, log the error instead of returning it to bubble up
Will convert to structured logging later on eventually

### DIFF
--- a/state-synchronizer/src/executor_proxy.rs
+++ b/state-synchronizer/src/executor_proxy.rs
@@ -5,6 +5,7 @@ use crate::{counters, SynchronizerState};
 use anyhow::{format_err, Result};
 use executor_types::{ChunkExecutor, ExecutedTrees};
 use itertools::Itertools;
+use libra_logger::prelude::*;
 use libra_types::{
     account_state::AccountState,
     contract_event::ContractEvent,
@@ -156,7 +157,13 @@ impl ExecutorProxyTrait for ExecutorProxy {
             intermediate_end_of_epoch_li,
         )?;
         timer.stop_and_record();
-        self.publish_on_chain_config_updates(reconfig_events)
+        if let Err(e) = self.publish_on_chain_config_updates(reconfig_events) {
+            error!(
+                "[state sync] failed to publish reconfig notification in execute_chunk: {}",
+                e
+            );
+        }
+        Ok(())
     }
 
     fn get_chunk(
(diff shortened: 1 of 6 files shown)


state-synchronizer/src/executor_proxy.rs
```diff
@@ -156,7 +157,13 @@ impl ExecutorProxyTrait for ExecutorProxy {
             intermediate_end_of_epoch_li,
         )?;
         timer.stop_and_record();
-        self.publish_on_chain_config_updates(reconfig_events)
+        if let Err(e) = self.publish_on_chain_config_updates(reconfig_events) {
+            error!(
+                "[state sync] failed to publish reconfig notification in execute_chunk: {}",
+                e
+            );
+        }
```
davidiw, line 165: I don't think this function returns a result any more, so you can probably just update the function signature

davidiw (approved):

state-synchronizer/src/executor_proxy.rs
```diff
@@ -219,7 +226,12 @@ impl ExecutorProxyTrait for ExecutorProxy {
             if !changed_configs.is_disjoint(&subscribed_items.configs)
                 || !event_keys.is_disjoint(&subscribed_items.events)
             {
-                subscription.publish(new_configs.clone())?;
+                if let Err(e) = subscription.publish(new_configs.clone()) {
+                    error!(
+                        "[state sync] failed to publish individual reconfig subscription: {}",
+                        e
+                    );
```
davidiw, line 233: can we print the subscription? can we make folks specify a name for each subscription?

state-synchronizer/src/executor_proxy.rs
christinacelee: there is an instance earlier in this function where we return result (storage-fetch configs): https://github.com/libra/libra/blob/master/state-synchronizer/src/executor_proxy.rs#L199
I thought about grouping all the per-subscription publishing errors and returning that as a single error so we can unify the location of logging the failure of this fn call here... but it seemed a bit ugly/hassle so making the tradeoff for simplicity, but lmk if you think otherwise!

davidiw:
Thanks for adding the name!!!

PR #6072, opened 2020-09-17
[config-builder] Remove it and other awful stuff

davidiw:
* Note: I have the docker portion defined here: https://github.com/libra/libra/pull/6098 -- so don't bother reviewing that (the first commit).
* Introduced a test_config for genesis tooling, it isn't as great as the config-builder version in terms of portability but it seems to get the job done. Will probably need to follow up to improve the testing code around that. I partially blame our approach to testing.
* Introduce several new ways to produce tests that only involve genesis, lots of tests were using config-builder for just this purpose and it was adding unnecessary barriers to doing useful things
* I went into each case of config_builder::test_config and swapped it for an appropriate backend
* Removed all of config-builder


language/tools/vm-genesis/src/lib.rs
```diff
@@ -543,103 +539,114 @@ fn verify_genesis_write_set(events: &[ContractEvent]) {
...
-            let test_config = n.test.as_ref().unwrap();
-            let owner_key = test_config.owner_key.as_ref().unwrap().public_key();
-            let operator_key = test_config.operator_key.as_ref().unwrap().public_key();
-            let operator_account = account_address::from_public_key(&operator_key);
-            let name = idx.to_string().as_bytes().to_vec();
-            let set_operator_script = transaction_builder::encode_set_validator_operator_script(
-                name.clone(),
-                operator_account,
-            );
+pub fn test_genesis_transaction() -> Transaction {
```
zekun000, line 560: why those functions are not with cfg(test)?

vm-validator/src/unit_tests/vm_validator_test.rs
```diff
@@ -166,15 +167,14 @@ fn test_validate_max_gas_units_above_max() {
...
-    let vm_validator = TestValidator::new(&config);
+    let vm_validator = TestValidator::new();
 
     let address = account_config::libra_root_address();
     let transaction = transaction_test_helpers::get_test_signed_transaction(
         address,
         1,
-        &key,
-        key.public_key(),
+        &vm_genesis::GENESIS_KEYPAIR.0,
```
zekun000, line 176: I guess any key pair works here? using genesis keypair looks weird

state-synchronizer/src/tests/on_chain_config_tests.rs
```diff
@@ -72,18 +75,10 @@ fn test_on_chain_config_pub_sub() {
...
-    let operator_key = config
-        .test
-        .as_ref()
-        .unwrap()
-        .operator_key
-        .as_ref()
-        .unwrap()
-        .private_key();
+    let validator_account = validators[0].owner_address;
+    let operator_key = validators[0].key.clone();
```
zekun000, line 79: not very clear whether it's owner key or operator key or consensus key

execution/executor/tests/db_bootstrapper_test.rs
```diff
@@ -254,21 +247,22 @@ fn restore_state_to_db(
 
 #[test]
 fn test_pre_genesis() {
-    let (mut config, genesis_key) = config_builder::test_config();
+    let genesis = vm_genesis::test_genesis_change_set_and_validators(Some(1));
+    let genesis_key = &vm_genesis::GENESIS_KEYPAIR.0;
```
zekun000, line 251: the GENESIS_KEY is confusing, can we name it something like LIBRA_ROOT_FOR_TEST?

zekun000 (approved):
long live the config builder

language/tools/vm-genesis/src/lib.rs
davidiw: they are used throughout the codebase, we'd have to add flags all over the place for it. I'm slowly getting to the point where I think having test utils exposed isn't that bad. Cargo doesn't play very kindly with these type of features and we aren't pulling in additional crates.

vm-validator/src/unit_tests/vm_validator_test.rs
davidiw: for this one, I guess it needs to be a legitimate account? i dunno, but look at the address, so we get the key for that address... I don't really understand how the test works.

state-synchronizer/src/tests/on_chain_config_tests.rs
davidiw: all keys are the same <3

execution/executor/tests/db_bootstrapper_test.rs
davidiw: hmm... @tzakian  and I were discussing some sort of refactoring here, the naming is bad, the testing configuration is bad... we should separate the testing utility from the underlying builder framework.

davidiw (commented):
@zekun000, genesis tools need a lot of love... some big opportunity for better engineering or a solid contribution from an external contributor.

Thanks for the review!

PR #6076, opened 2020-09-17
[move-prover] Verify everything by default & fixes in specs and prover.

wrwg:
This PR makes verifying all functions of a module the default, both when the prover is called from the command line and in tests. The fact that in tests verification was turned off by default has created confusion. It was useful some time ago when large parts of the framework were not specified/verified. Also, we rarely only want to verify public functions, so the default from command line is now all functions (same as in tests).

With this PR, all transaction scripts and modules are verified in tests, by default. This discovered an inconsistency in `LibraConfig` where an `aborts_if [assume] P` and a global invariant contradicted each other. The spec has been rewritten, and instead of `[assume]` now `[concrete]` is used for aborts conditions we don't want to propagate. This works only with opaque functions, but is a good methodology moving forward to avoid inconsistencies like this.

The `[concrete]` property implementation had a bug which is fixed here as well.

## Motivation

Aligning test and command line behavior.

Yes

## Test Plan

Existing tests.

## Related PRs

NA

### DIFF
--- a/language/move-prover/src/spec_translator.rs
+++ b/language/move-prover/src/spec_translator.rs
@@ -494,39 +494,37 @@ impl<'a> ConditionDistribution<'a> {
         let abstract_ = get_prop(CONDITION_ABSTRACT_PROP);
         let concrete = get_prop(CONDITION_CONCRETE_PROP);
 
-        let external = (!injected || exported) && !concrete;
+        // Determines when a condition will be propagated to the caller. This is the case if
+        // the condition is not injected via a schema apply or it is explicitly marked
+        // as exported, and if it is not marked as concrete. Conditions which are not propagated
+        // are only applied to the verification entry point.
+        let propagated_to_intra_caller = !concrete;
+        let propagated_to_inter_caller = (!injected || exported) && propagated_to_intra_caller;
 
+        use FunctionEntryPoint::*;
         match entry_point {
-            FunctionEntryPoint::DirectInterModule if opaque => {
+            DirectInterModule if propagated_to_inter_caller && opaque => {
                 use ConditionKind::*;
                 match &cond.kind {
-                    Requires if external => self.requires.push(cond),
-                    Ensures if external => self.ensures.push(cond),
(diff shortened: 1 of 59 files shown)


language/move-prover/src/spec_translator.rs
```diff
@@ -494,39 +494,37 @@ impl<'a> ConditionDistribution<'a> {
         let abstract_ = get_prop(CONDITION_ABSTRACT_PROP);
         let concrete = get_prop(CONDITION_CONCRETE_PROP);
 
-        let external = (!injected || exported) && !concrete;
+        // Determines when a condition will be propagated to the caller. This is the case if
+        // the condition is not injected via a schema apply or (if so) it is explicitly marked
+        // as exported, and if it is not marked as concrete. Conditions which are not propagated
+        // or only applied to the verification entry point.
```
shazqadeer, line 500: or -> are

language/stdlib/modules/LibraConfig.move
```diff
@@ -235,18 +235,42 @@ module LibraConfig {
...
+                last_reconfiguration_time, now);
+    }
+    /// The following schema describes aborts conditions which we do not want to be propagated to the verification
+    /// of callers, and which are therefore marked as `concrete` to be only verified against the implementation.
+    /// These conditions are unlikely to happen in reality, and excluding them avoids formal noise.
+    spec schema InternalReconfigureAbortsIf {
+        let config = global<Configuration>(CoreAddresses::LIBRA_ROOT_ADDRESS());
+        let current_time = LibraTimestamp::spec_now_microseconds();
+        aborts_if [concrete] current_time <= config.last_reconfiguration_time with Errors::INVALID_STATE;
+        aborts_if [concrete] config.epoch == MAX_U64 with EXECUTION_FAILURE;
```
shazqadeer, line 265: Ah, I can use EXECUTION_FAILURE as an "abort code".  I didn't know that.

shazqadeer (approved):

language/move-prover/src/spec_translator.rs
wrwg: Done.

PR #6077, opened 2020-09-17
Update JsonRpcAsyncClientError logic to retry on all client errors

ankushagarwal:
## Summary

We were not retrying for certain classes of errors. Generally we should be able to retry for any errors originating in the reqwest Client.

### DIFF
--- a/client/json-rpc/src/client.rs
+++ b/client/json-rpc/src/client.rs
@@ -149,15 +149,9 @@ impl JsonRpcAsyncClientError {
     pub fn is_retriable(&self) -> bool {
         match self {
             JsonRpcAsyncClientError::ClientError(e) => {
-                // TODO: Add following condition when is_request is added to reqwest stable release
-                // // Error sending request
-                //  if e.is_request() {
-                //      return true;
-                //  }
-                if e.is_timeout() {
+                if e.is_timeout() || e.is_request() {
                     return true;
                 }
-
                 if let Some(status) = e.status() {
                     // Returned status code indicates a server error
                     return status.is_server_error();


moeziniaf (approved):

client/json-rpc/src/client.rs
```diff
@@ -148,22 +148,7 @@ pub enum JsonRpcAsyncClientError {
...
-                    return true;
-                }
-
-                if let Some(status) = e.status() {
-                    // Returned status code indicates a server error
-                    return status.is_server_error();
-                }
-                false
-            }
+            JsonRpcAsyncClientError::ClientError(_) => true,
```
dmitri-perelman, line 151: I am not sure this is necessarily true. Are we 100% sure we'd not receive `ClientError` in case there is a problem with the request itself?

dmitri-perelman:
While non-retryable errors are bad, inifinite retries for things that should not be retried are much worse. I'd be extra careful, and would not push this change unless we're 100% sure we understand these error types.

ankushagarwal:
@dmitri-perelman You are right. I updated this, so that we only check for known cases.

thefallentree (approved):

client/json-rpc/src/client.rs
```diff
@@ -149,15 +149,9 @@ impl JsonRpcAsyncClientError {
     pub fn is_retriable(&self) -> bool {
         match self {
             JsonRpcAsyncClientError::ClientError(e) => {
-                // TODO: Add following condition when is_request is added to reqwest stable release
-                // // Error sending request
-                //  if e.is_request() {
-                //      return true;
-                //  }
-                if e.is_timeout() {
+                if e.is_timeout() || e.is_request() {
```
xli, line 152: what is this is_request?

also, please add test

xli (changes requested):

client/json-rpc/src/client.rs
ankushagarwal: is_request() is set to true whenever there is a problem sending the request. According to the reqwest docs `/// Returns true if the error is related to the request`

I don't think its possible to unit test this since the constructor for this error type is private to the crate.
I locally tested this by prematurely disconnecting the upstream server and verified that `e.is_request()` is set to true.

PR #6079, opened 2020-09-17
[dkr] Fix permissions of azcopy

mgorven:
Use `--no-same-owner` so it's owned by root, and `chmod +x ` so
group/world get execute permission.


msmouse (approved):
Thanks!

PR #6081, opened 2020-09-17
Require currency codes to be composed of alphanumeric characters

bob-wilson:
Currency code strings must be valid Move identifiers but that does not rule out underscore characters. ISO 4217 only allows sequences of 3 uppercase alphabetic characters, so this is still more lax. We should consider further restrictions in the future but the current placeholder currencies "Coin1" and "Coin2" prevent us from restricting lowercase, numbers, and the length.

## Motivation

Moving toward a more careful definition of valid currency codes

Yes

## Test Plan

Updated tests for invalid gas specifiers to exercise this new code.


tzakian (approved):
LGTM. Thanks for this!

PR #6084, opened 2020-09-17
[json-rpc] add new integration test suite for better code coverage and adding tests easier

xli:
It is challenge to add more json-rpc tests to cover interface definition (like parent vasp role account) in existing unit tests, because it's hard to create test data by mockdb.
This change creates an integration test with local libra validator node and json-rpc service, so that setup data is accurate and easier by executing specific transactions.

Along with new integration test, I converted existing metadata, currency info, unknown role type account and dd role type account tests, and added new tests to cover: parent vasp role account, child vasp role account, peer to peer transaction with events.


libra-action:

Breaking changes policy:

2. Adding fields is ok.

json-rpc/tests/testing.rs
```diff
@@ -154,8 +154,7 @@ impl Env {
             30,
             ChainId::test(),
         )
-        .unwrap();
-        txn
+        .unwrap()
```
thefallentree, line 157: I think we should use. ` .expect("Unable to create Transaction") `

thefallentree (approved):

json-rpc/Cargo.toml
```diff
@@ -57,6 +60,8 @@ executor = { path = "../execution/executor", version = "0.1.0" }
 executor-types = { path = "../execution/executor-types", version = "0.1.0" }
 scratchpad = { path = "../storage/scratchpad", version = "0.1.0" }
 move-vm-types = { path = "../language/move-vm/types", version = "0.1.0" }
+transaction-builder-generated = { path = "../client/transaction-builder", version = "0.1.0" }
+workspace-builder = { path = "../common/workspace-builder", version = "0.1.0" }
```
bmwill, line 64: Please avoid using this dependency. I have a task in my backlog to remove this as it is harmful to the developer experience.

json-rpc/tests/node.rs
```diff
@@ -0,0 +1,99 @@
...
+}
+
+impl Node {
+    pub fn start(node_dir: &std::path::Path) -> Result<Self> {
+        let config_path = node_dir.join("node.yaml");
+        let struct_log_path = node_dir.join("struct.log");
+        let log_path = node_dir.join("log");
+
+        // TODO: test with default_for_validator_full_node?
+        let builder = ValidatorBuilder::new(1, NodeConfig::default_for_validator(), node_dir);
```
zekun000, line 41: @davidiw your favorite config builder

bmwill (dismissed):
I'm actually going to block the landing of this due to the use of workspace-builder and see if I can quickly revive my PR to remove it

json-rpc/tests/node.rs
xli: is there other way to config node? I'm fine with whatever works, this is the only way I found to configure a node.

zekun000: I misread this, I think the genesis_tool::config_builder is the right way to go. I thought it uses the deprecated one because of the added dependency
```
config-builder = { path = "../config/config-builder", version = "0.1.0" }
```

json-rpc/Cargo.toml
```diff
@@ -44,7 +44,10 @@ proptest = "0.10.1"
 reqwest = { version = "0.10.8", features = ["blocking", "json"], default_features = false }
 rand = { version = "0.7.3" }
 
+config-builder = { path = "../config/config-builder", version = "0.1.0" }
```
zekun000, line 47: so I guess you don't really need this dependency?

xli: it is required for BuildSwarm trait implemented by ValidatorBuilder#build_swarm

json-rpc/tests/node.rs
davidiw: I also refactored this code in an uncommitted PR, in the future @xli can just use libra_genesis_tool::test_config() and be done with it: https://github.com/libra/libra/pull/6072

beaudio530:
How do I download this?
Is there a way for contributing\mining\to increase value before it's live

PR #6087, opened 2020-09-17
[move source language] Compiled module interface generator

tnowacki:
- Generator to build "interface" files for compiled modules
- Will be integrated into other tools to let the source language compile indirectly interact with compiled modules

## Motivation

- Will be very useful for the CLI as it will allow to publish single modules at a time, and actually compile against what is in the mock storage 

## Test Plan

- future PR. Not used now

### DIFF
--- a/language/move-lang/src/interface_generator.rs
+++ b/language/move-lang/src/interface_generator.rs
@@ -1,0 +1,284 @@
+// Copyright (c) The Libra Core Contributors
+// SPDX-License-Identifier: Apache-2.0
+
+use crate::shared::Address;
+use anyhow::{anyhow, Result};
+use move_core_types::language_storage::ModuleId;
+use move_vm::{
+    access::ModuleAccess,
+    file_format::{
+        CompiledModule, FunctionDefinition, Kind, SignatureToken, StructDefinition,
+        StructFieldInformation, StructHandleIndex, TypeParameterIndex,
+    },
+};
+use std::{collections::BTreeMap, fs};
+
+macro_rules! push_line {
+    ($s:ident, $e:expr) => {{
+        $s = format!("{}{}\n", $s, $e);
+    }};
+}
+
+macro_rules! push {
+    ($s:ident, $e:expr) => {{
+        $s = format!("{}{}", $s, $e);
+    }};
+}
+
+/// Generate the text for the "interface" file of a compiled module. This "interface" is the
+/// publically visible contents of the CompiledModule, represented in source language syntax
+/// Additionally, it returns the module id (address+name) of the module that was deserialized
+pub fn write_to_string(compiled_module_file_input_path: &str) -> Result<(ModuleId, String)> {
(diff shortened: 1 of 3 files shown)


language/move-lang/src/interface_generator.rs
```diff
@@ -0,0 +1,284 @@
...
+    for field in fields {
+        push_line!(
+            out,
+            format!(
+                "        {}: {},",
+                ctx.module.identifier_at(field.name),
+                write_signature_token(ctx, &field.signature.0),
+            )
+        )
+    }
```
vgao1996, line 160: Just some random thoughts, but I wonder if we would prefer translating struct defs to something more opaque (like the old native struct) if they ever become available in the future... or maybe not? Technically speaking fields are not accessible from outside module so there is no benefit for the Move program to know them, but maybe a smart compiler can leverage the extra knowledge about the layout in some way?

language/move-lang/src/interface_generator.rs
```diff
@@ -0,0 +1,284 @@
...
+        Kind::Resource => ": resource".to_string(),
+        Kind::Copyable => ": copyable".to_string(),
+    }
+}
+
+fn write_parameters(ctx: &mut Context, params: &[SignatureToken]) -> String {
+    params
+        .iter()
+        .enumerate()
+        .map(|(idx, ty)| format!("a{}: {}", idx, write_signature_token(ctx, ty)))
```
vgao1996, line 211: Our current move code uses capital letters (e.g. `T`, `CoinType`). Do we also want to follow that convention here?

vgao1996 (commented):
> Will be very useful for the CLI as it will allow to publish single modules at a time, and actually compile against what is in the mock storage
Thoughts on making this a standalone tool/crate instead of being part of the compiler?

Would you mind adding a test (preferrab expected out) for this tool? My experience is that tools without tests will get broken and it would be a pity if that actually happens again.

tnowacki:
> Thoughts on making this a standalone tool/crate instead of being part of the compiler?

It started that way, but I didn't really gain much from this. And it was much more difficult to test and integrate into the compiler. It being in the same crate doesn't stop a separate bin executable from being made, so I think its okay for now

> Would you mind adding a test (preferably expected out) for this tool? My experience is that tools without tests will get broken and it would be a pity if that actually happens again.

It will be in the next PR

language/move-lang/src/interface_generator.rs
tnowacki: I originally started with `native` structs (still supported in the source language compiler), but I figured seeing the fields could be helpful for trying to orient oneself when looking at on chain data

In other words, having the fields there might be helpful if you are using this in conjunction with something like the resource viewer

vgao1996: Fair point. I was thinking of this more as a library producing outputs that get consumed by other tools, but now I can see its value to a human reader.

PR #6088, opened 2020-09-17
[move prover] add sequential option to prover task runner

emmazzz:
## Motivation

This PR adds an option `--sequential` to the Move Prover so that when it is turned on, the Boogie instances will run sequentially instead of simultaneously, meaning that the second Boogie instance will only get to run if the first instance failed. This is implemented using a semaphore.

In addition, this PR also sets the number of Boogie instances to be 2 and turns `--sequential` on for testing. For command line, the default number of instances is 1 and sequential off. Verification is turned on for a few move functions which previously timed out to see how they behave with the new feature on in MIRAI.

Yes

### DIFF
--- a/language/move-prover/src/prover_task_runner.rs
+++ b/language/move-prover/src/prover_task_runner.rs
@@ -7,15 +7,18 @@
 use crate::cli::Options;
 use async_trait::async_trait;
 use futures::{future::FutureExt, pin_mut, select};
+use log::debug;
 use rand::Rng;
+use regex::Regex;
 use std::{
     process::Output,
-    sync::Arc,
-    sync::mpsc::{channel, Sender},
+    sync::{
+        mpsc::{channel, Sender},
+        Arc,
+    },
 };
 use tokio::{
     process::Command,
-    runtime::Runtime,
     sync::{broadcast, broadcast::Receiver, Semaphore},
 };
 
@@ -23,6 +26,8 @@ use tokio::{
 enum BroadcastMsg {
     Stop,
 }
+
+const MAX_PERMITS: usize = usize::MAX >> 4;
 
 #[async_trait]
 pub trait ProverTask {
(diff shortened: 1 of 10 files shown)


language/move-prover/src/prover_task_runner.rs
```diff
@@ -32,20 +36,36 @@ pub trait ProverTask {
...
+        num_instances: usize,
+        sequential: bool,
+    ) -> (T::TaskId, T::TaskResult)
     where
         T: ProverTask + Clone + Send + 'static,
     {
-        let rt = Runtime::new().unwrap();
-
+        let rt = tokio::runtime::Builder::new()
+            .core_threads(2)
```
wrwg, line 59: Why do you have hardcoded 2 threads here? This restricts parallelism to two 2, no?

Should this be set to num_instances? Or perhaps removed and let Tokio decide how many threads to use based on available cores on the machine>

language/move-prover/src/prover_task_runner.rs
```diff
@@ -32,20 +36,36 @@ pub trait ProverTask {
...
+            .threaded_scheduler()
+            .enable_all()
+            .build()
+            .unwrap();
+        let sem = if sequential {
+            Arc::new(Semaphore::new(1))
+        } else {
+            Arc::new(Semaphore::new(num_instances))
```
wrwg, line 67: Why would you need to set the semaphore with num_instances? If sequential is false, I would set it to MAX and let the old logic deal with this (e.g. the # of threads tokio allocated for the scheduler).

wrwg, line 67: Are you sure you need Arc around a semaphore? The type seems to implement the Sync trait, so it is save to pass between threads directly.

wrwg (changes requested):

language/move-prover/src/prover_task_runner.rs
emmazzz: Semaphore does implement Sync, but it doesn't implement Copy or Clone trait which is needed for us to pass the semaphore in at each iteration of the loop. If we pass the semaphore directly, Rust compiler complains that it has been moved in the last iteration of the loop.

wrwg (approved):

PR #6089, opened 2020-09-17
[faucet] mint endpoint accepts a new param to return submitted transactions

xli:
## Motivation

Existing mint endpoint returns next dd account sequence number, 
Mint endpoint creates new onchain account and transfer coins from dd account to new account.
Client should wait for transactions executed and validate they are success before it moves forward to do anything with the new account.

Existing return value is next dd account seq number, causes client wait for next dd account seq number shows up in the dd account, but a new seq number showed up in dd account does not mean the transaction for the client is succeed.

This diff adds a new param for server to return all submitted transactions, then client can wait for these transaction executed and validate their execution result for confirm mint action succeed.

(Write your answer here.)

## Test Plan

## Related PRs

### DIFF
--- a/client/faucet/src/mint.rs
+++ b/client/faucet/src/mint.rs
@@ -7,16 +7,38 @@ use libra_types::account_config::{
     testnet_dd_account_address, treasury_compliance_account_address, type_tag_for_currency_code,
     LBR_NAME,
 };
-use std::convert::From;
+use std::{convert::From, fmt};
+
+#[derive(Debug)]
+pub enum Response {
+    DDAccountNextSeqNum(u64),
+    SubmittedTxns(Vec<libra_types::transaction::SignedTransaction>),
+}
+
+impl std::fmt::Display for Response {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        match self {
+            Response::DDAccountNextSeqNum(v1) => write!(f, "{}", v1),
+            Response::SubmittedTxns(v2) => {
+                write!(f, "{}", hex::encode(lcs::to_bytes(&v2).unwrap()))
+            }
+        }
+    }
+}
 
 #[derive(serde_derive::Deserialize)]
 pub struct MintParams {
-    amount: u64,
-    currency_code: move_core_types::identifier::Identifier,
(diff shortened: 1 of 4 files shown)


client/faucet/src/mint.rs
```diff
@@ -7,24 +7,46 @@ use libra_types::account_config::{
...
-use std::convert::From;
+use std::{convert::From, fmt};
+
+#[derive(Debug)]
+pub enum Response {
+    DDAccountNextSeqNum(u64),
+    SubmittedTxns(Vec<libra_types::transaction::SignedTransaction>),
+}
+
+impl std::fmt::Display for Response {
```
thefallentree, line 18: should we do impl FromStr for Response instead ?

types/src/transaction/authenticator.rs
```diff
@@ -272,7 +272,10 @@ impl FromStr for AuthenticationKey {
     type Err = Error;
 
     fn from_str(s: &str) -> Result<Self> {
-        assert!(!s.is_empty());
```
thefallentree, line 275: what.. who added this...

thefallentree: can you add an testcase to make sure this don't appear again?

xli: it does not matter anymore after changed to use ensure, can't avoid people use assert with test, they will only add new use case.

I guess a lint rule will be better.

client/faucet/src/mint.rs
xli: no, because I don't expect create Response from string

types/src/transaction/authenticator.rs
xli: anyway,  test added

client/faucet/src/mint.rs
thefallentree: sorry, what I meant is actually implement to_string() https://doc.rust-lang.org/std/string/trait.ToString.html , Display is used for using in an format string that uses "{}", which I don't think is what we need here.

thefallentree: nvm, found out

thefallentree (approved):

sunmilee:
lgtm

PR #6090, opened 2020-09-17
[Operational Tool] Display validator configs even on validator network address decryption failure.

JoshLind:
## Motivation

This PR updates the operational tool to better handle validator network address decryption failures when running the `validator-set` command. In the case that we fail to decrypt an address, the tool will currently display an error message for each failed decryption and show an empty validator set. Rather, we should log the error, display the values that we can and then display a dummy address in the place of the encrypted one. See the outputs below.

Yes.

## Test Plan

Testing the tool by pointing to it a testnet deployment and forcing a decryption failure, we get the following:

```
joshlind@joshlind-mbp debug % ./libra-operational-tool validator-set --config <CONFIG FILE>
Unable to decode account d4c4fb4956d899e55289083f45ac5d84: Unable to deocde network address: Failed reading validator_network_address_keys from storage
Unable to decode account 57208e640c623b27c6bba704380825ab: Unable to deocde network address: Failed reading validator_network_address_keys from storage
Unable to decode account a2719315a51388bc1f1e1c5afa2daaa9: Unable to deocde network address: Failed reading validator_network_address_keys from storage
Unable to decode account 1e674e850cb4d116babc6f870da9c258: Unable to deocde network address: Failed reading validator_network_address_keys from storage
{
  "Result": []
}
```

After this PR we get:
```
joshlind@joshlind-mbp debug % ./libra-operational-tool validator-set --config <CONFIG FILE>
Unable to decode account d4c4fb4956d899e55289083f45ac5d84: Unable to decode network address: Failed reading validator_network_address_keys from storage. Using a dummy validator network address!
Unable to decode account 57208e640c623b27c6bba704380825ab: Unable to decode network address: Failed reading validator_network_address_keys from storage. Using a dummy validator network address!
Unable to decode account a2719315a51388bc1f1e1c5afa2daaa9: Unable to decode network address: Failed reading validator_network_address_keys from storage. Using a dummy validator network address!
Unable to decode account 1e674e850cb4d116babc6f870da9c258: Unable to decode network address: Failed reading validator_network_address_keys from storage. Using a dummy validator network address!
{
  "Result": [
    {
      "name": "val0",
      "account_address": "d4c4fb4956d89...",
      "consensus_public_key": "0292fbfd7ce56f6888f3ea...",
      "fullnode_network_address": "/dns4/val0-libra-validator-fullnode-lb/tcp/6182/ln-noise-ik/...",
      "validator_network_address": "/dns4/0000"
    },
    ...
  ]
}
```

## Related PRs

None.

### DIFF
--- a/config/management/operational/src/validator_set.rs
+++ b/config/management/operational/src/validator_set.rs
@@ -7,6 +7,7 @@ use libra_management::{config::ConfigPath, error::Error, secure_backend::Validat
 use libra_network_address::NetworkAddress;
 use libra_types::account_address::AccountAddress;
 use serde::Serialize;
+use std::str::FromStr;
 use structopt::StructOpt;
 
 #[derive(Debug, StructOpt)]
@@ -44,11 +45,26 @@ impl ValidatorSet {
                 Ok(config) => config,
                 Err(err) => {
                     println!(
-                        "Unable to decode account {}: {}",
+                        "Unable to decode account {}: {}. Using a dummy validator network address!",
                         info.account_address(),
                         err
                     );
-                    continue;
+
+                    // Return the partially decrypted validator config (using a dummy validator
+                    // network address).
+                    let encrypted_config = info.config();
+                    DecryptedValidatorConfig {
+                        name: "".to_string(),
+                        consensus_public_key: encrypted_config.consensus_public_key.clone(),
+                        fullnode_network_address: encrypted_config
+                            .fullnode_network_addresses()
+                            .map_err(|e| Error::NetworkAddressDecodeError(e.to_string()))?[0]
+                            .clone(),
+                        validator_network_address: NetworkAddress::from_str(
+                            "/dns4/could-not-decrypt",
+                        )
+                        .map_err(|e| Error::NetworkAddressDecodeError(e.to_string()))?,
+                    }
                 }
             };
 
--- a/config/management/src/error.rs
+++ b/config/management/src/error.rs
@@ -23,7 +23,7 @@ pub enum Error {
     JsonRpcReadError(&'static str, String),
     #[error("Failed to write '{0}' from JSON-RPC: {1}")]
     JsonRpcWriteError(&'static str, String),
-    #[error("Unable to deocde network address: {0}")]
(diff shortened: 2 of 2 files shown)


gregnazario (approved):
Do we know it's not going to be able to decrypt it beforehand?  If so we should probably skip the failed to decrypt messages.

config/management/operational/src/validator_set.rs
```diff
@@ -44,11 +45,24 @@ impl ValidatorSet {
...
+                    // network address).
+                    let encrypted_config = info.config();
+                    DecryptedValidatorConfig {
+                        name: "".to_string(),
+                        consensus_public_key: encrypted_config.consensus_public_key.clone(),
+                        fullnode_network_address: encrypted_config
+                            .fullnode_network_addresses()
+                            .map_err(|e| Error::NetworkAddressDecodeError(e.to_string()))?[0]
+                            .clone(),
+                        validator_network_address: NetworkAddress::from_str("/dns4/0000")
```
JoshLind, line 63: @gregnazario, any idea for a better dummy validator network address? The DecryptedValidatorConfig object requires one, and I'd ideally like to avoid refactoring DecryptedValidatorConfig to also take in an encrypted adress in the case of a failure... (hence the dummy address here). Any better suggestions?

gregnazario: `/dns4/could-not-decrypt` or, just make it an `Option<NetworkAddress>`

JoshLind: Ooh, nice! Let's go with "could-not-decrypt" -- I'm in favour of displaying something in the output where the address should be so that the operator knows the decryption failed (e.g., if the validator set is long, they might miss the error lines at the top)

JoshLind:
Thanks, @gregnazario! 😄 

> Do we know it's not going to be able to decrypt it beforehand? If so we should probably skip the failed to decrypt messages.

We can't really tell because we require a storage backend to be present -- so it could work. Although, personally I'm in favour of displaying the failure so the user knows exactly what's happening and there isn't a silent failure.

/land

PR #6091, opened 2020-09-18
[move source language] Integrate interface generator

tnowacki:
- Start using the interface generator to let MV files appear in the dep lists for the source language

## Motivation

- It's real cool (should be very helpful for the CLI)

## Test Plan

- Added sanity test that uses them to compile all of the standard library scripts

## Related PRs

See #6087

### DIFF
--- a/language/move-lang/src/lib.rs
+++ b/language/move-lang/src/lib.rs
@@ -36,6 +36,7 @@ use std::{
     path::{Path, PathBuf},
     str::Chars,
 };
+use tempfile::NamedTempFile;
 
 pub const MOVE_EXTENSION: &str = "move";
 pub const MOVE_COMPILED_EXTENSION: &str = "mv";
@@ -54,8 +55,9 @@ pub fn move_check(
     targets: &[String],
     deps: &[String],
     sender_opt: Option<Address>,
+    interface_files_dir_opt: Option<String>,
 ) -> anyhow::Result<()> {
-    let (files, errors) = move_check_no_report(targets, deps, sender_opt)?;
+    let (files, errors) = move_check_no_report(targets, deps, sender_opt, interface_files_dir_opt)?;
     if !errors.is_empty() {
         errors::report_errors(files, errors)
     }
@@ -67,8 +69,11 @@ pub fn move_check_no_report(
     targets: &[String],
     deps: &[String],
     sender_opt: Option<Address>,
+    interface_files_dir_opt: Option<String>,
 ) -> anyhow::Result<(FilesSourceText, Errors)> {
-    let (files, pprog_and_comments_res) = parse_program(targets, deps)?;
+    let mut deps = deps.to_vec();
+    generate_interface_files_for_deps(&mut deps, interface_files_dir_opt)?;
+    let (files, pprog_and_comments_res) = parse_program(targets, &deps)?;
     let pprog_res = pprog_and_comments_res.map(|(pprog, _)| pprog);
     match check_program(pprog_res, sender_opt) {
         Err(errors) => Ok((files, errors)),
@@ -85,10 +90,11 @@ pub fn move_compile(
     targets: &[String],
     deps: &[String],
     sender_opt: Option<Address>,
+    interface_files_dir_opt: Option<String>,
 ) -> anyhow::Result<(FilesSourceText, Vec<CompiledUnit>)> {
-    let (files, pprog_and_comments_res) = parse_program(targets, deps)?;
-    let pprog_res = pprog_and_comments_res.map(|(pprog, _)| pprog);
-    match compile_program(pprog_res, sender_opt) {
+    let (files, compiled_units_or_errors) =
+        move_compile_no_report(targets, deps, sender_opt, interface_files_dir_opt)?;
+    match compiled_units_or_errors {
         Err(errors) => errors::report_errors(files, errors),
         Ok(compiled_units) => Ok((files, compiled_units)),
     }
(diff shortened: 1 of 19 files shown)


language/move-lang/src/lib.rs
```diff
@@ -197,6 +210,88 @@ pub fn output_compiled_units(
...
+) -> anyhow::Result<Option<String>> {
+    let mv_files = find_move_filenames_by_extension(mv_file_locations, MOVE_COMPILED_EXTENSION)?
+        .into_iter()
+        .filter(|s| {
+            let p = Path::new(s);
+            let mv_file = p.is_file()
+                && p.extension()
+                    .map(|ext| ext == MOVE_COMPILED_EXTENSION)
+                    .unwrap_or(false);
+            mv_file || has_compiled_module_magic_number(s)
```
tnowacki, line 235: @bob-wilson I had some issues with 'find_move_filenames_by_extension' since it always returns specific files. So I had issues with some temp files (as I could not give them custom extensions). 
I think what I have here is okay, but let me know if you have any thoughts

language/move-lang/src/lib.rs
```diff
@@ -197,6 +210,88 @@ pub fn output_compiled_units(
...
+
+        dir
+    };
+
+    for mv_file in mv_files {
+        let (id, interface_contents) = interface_generator::write_to_string(&mv_file)?;
+        let addr_dir = format!("{}/{:#x}", hash_dir, id.address());
+        let file_path = format!("{}/{}.{}", addr_dir, id.name(), MOVE_EXTENSION);
+        // it's possible some files exist but not others due to multithreaded environments
+        if Path::new(&file_path).is_file() {
```
tnowacki, line 270: This hashing story is a bit weird. 
Breakdown of what's going on
- I tried using temp files, but I had issues trying to convince the system not to clean them up before they were used
- I tried just using the timestamp or something to just make a new folder every time, but this generated way too many files for test runs
- The hashing was an attempt to just use one set of files per set of dependencies 
- But by grouping everything together, I had issues when the files had not yet been created. So each test thread would be racing to complete them, and then stepping on each other to write to the file. 
- Tried just checking to see if the directory existed, but that was too fast, as the thread might move on before all the files were populated by a different test thread 
- I tried doing some big hacks by having a `static MUTEX: Lazy<Mutex<()>> = ...` and then wrapping this whole section in a mutex, but I don't think it was very clean, and seemed to cause the tests to be really slow
- My solution was to just skip any file that was already created, and then the threads sort of helped each other in filling in the directory. It feels a bit unsafe as there still is technically a possibility that one thread starts creating the files while the other ones race ahead to use them, but get there before the creation is finished. That hasn't happened yet in my local testing yet. 

Any thoughts? cc @vgao1996

sblackshear (commented):
Couple of probably-dumb questions: 
- what are the pros and cons of actually creating the interface file on disk instead of creating a temporary in-memory data structure for the interface?
- do these interface files continue to stay up to date when their corresponding source and/or bytecode files change (including renaming and deletion of the file)?

language/move-lang/src/lib.rs
bob-wilson: The code seems unnecessarily confusing this way. As I understand it, there are 2 cases:

1. A dependency is specified as a directory. Only the files within that directory with the MOVE_COMPILED_EXTENSION will be added to the list of mv_files. In this case, there's no reason to check the extension again, and it's confusing to see that check again.
2. A filename is specified directly as a dependency. It will get through find_move_filenames_by_extension regardless of the extension match. I assume this is where you're seeing the issue with temp files. Checking the extension is not redundant in this case, but neither is it reliable. How about checking only the magic number here? It would be a bit slower because the file needs to be opened and read, but I'd be surprised if that is even noticeable.

Regardless, a comment would be helpful here to explain what's going on.

language/move-lang/src/lib.rs
bob-wilson: My gut feeling is that there must be a better way, but I haven't come up with anything in particular to suggest yet. This might be something that needs to evolve over time.

bob-wilson (approved):

language/move-lang/src/lib.rs
tnowacki: > Only the files within that directory with the MOVE_COMPILED_EXTENSION will be added to the list of mv_files. In this case, there's no reason to check the extension again, and it's confusing to see that check again.

Ah! So I thought too! But the find files utility returns any specific files passed to it. I'll clean that up a bit

libra-action:
```
target packages:
  M tempfile 3.1.0 (direct third-party, crates.io)
    * status changed from transitive third-party
```

libra-action:
```
target packages:
  M tempfile 3.1.0 (direct third-party, crates.io)
    * status changed from transitive third-party
```

tnowacki:
> what are the pros and cons of actually creating the interface file on disk instead of creating a temporary in-memory data structure for the interface?

The con of creating of doing something in memory is its *really* hard, lol. In order to do this, you would have to splice in a variant in the AST at each step of the way, and handle the fact that these alternative variants do not have source location information. In short, it would be a tremendous undertaking that would provide a worse user experience (because error messages could no longer point to files). 

> do these interface files continue to stay up to date when their corresponding source and/or bytecode files change (including renaming and deletion of the file)?

It will stay up to date for any edits, including deletion but not including renaming. The files are bucketed based on the hash of the contents of all of the `MV` files. I could theoretically do the hash of each individual file, but I think it would make it really confusing for a user to find all of the generated interface files for a given run. 

In short, if you rerun the same command after changing an MV file, the hash will change and you will get a new directory for the generated files.

language/move-lang/src/lib.rs
tnowacki: I've done the following as a better current solution

- Check if the file is there, if it is continue
- Write the file contents to a temp file
- Double check if the file is there, if it is continue
- std::fs::rename the temp file into the proper file. This should stop issues I had with one thread writing to the file as the other is reading from it

libra-action:
```
target packages:
  M tempfile 3.1.0 (direct third-party, crates.io)
    * status changed from transitive third-party
```

sblackshear:
Thanks for the detailed answers, makes sense! Excited to see this go in.

libra-action:
```
target packages:
  M tempfile 3.1.0 (direct third-party, crates.io)
    * status changed from transitive third-party
```

PR #6093, opened 2020-09-18
[libra-framework] Specified more transaction scripts

junkil-park:
Specified the following transaction scripts:
- burn.move
- publish_shared_ed25519_public_key.move
- rotate_dual_attestation_info.move
- rotate_shared_ed25519_public_key.move

Specified the SharedEd25519PublicKey module

Added the following governance spec to DualAttestation:
- the resource `Credential` is not transferrable.
- the invariant that only PareantVASP or DD can have `Credential`

Added the spec function & test cases for the native functions in Signature

## Motivation

Specify the transaction scripts

Yes

## Test Plan

cargo test

### DIFF
--- a/language/move-prover/tests/sources/functional/verify_lcs.move
+++ b/language/move-prover/tests/sources/functional/verify_lcs.move
@@ -1,6 +1,6 @@
 // This file is created to verify the native function in the standard LCS module.
 
-module VerifyVector {
+module VerifyLCS {
     use 0x1::LCS;
 
     public fun verify_to_bytes<MoveValue>(v: &MoveValue): vector<u8>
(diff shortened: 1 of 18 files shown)


language/stdlib/modules/Signature.move
```diff
@@ -23,6 +23,11 @@ module Signature {
         public_key: vector<u8>,
         message: vector<u8>
     ): bool;
+
+    spec module {
+        native define spec_ed25519_validate_pubkey(public_key: vector<u8>): bool;
```
emmazzz, line 28: You can use `Signature::ed25519_validate_pubkey` and `Signature::ed25519_verify` directly in specs, since they are both pure functions and I have defined them in the prelude already. There's no need to add spec versions here and in the prelude.

junkil-park: Dome removing these spec functions.

emmazzz (approved):

PR #6094, opened 2020-09-18
[consensus] block on reconfig notification

zekun000:
## Motivation

This commit fix a race condition when `start_new_epoch` finished but
state sync hasn't send the notification, the loop will continue to
process messages assuming it's still in the old epoch until the notification arrives in the `reconfig_events` channel.

The bug is found by a stress test of 1000 epoch + fault injection, it crashed the node with 
```
Failed to persist commit: Internal error: "Try to commit stale transactions with the last version as 2262"'
```
because the node already syncs to the new epoch (version 2268) but still process consensus messages and result in a commit in the previous epoch.

Y

## Test Plan

ENABLE_FAILPOINTS=1 ./scripts/cti --pr 6094 --run reconfiguration -- --count 1000

## Related PRs

### DIFF
--- a/consensus/src/epoch_manager.rs
+++ b/consensus/src/epoch_manager.rs
@@ -78,6 +78,7 @@ pub struct EpochManager {
     storage: Arc<dyn PersistentLivenessStorage>,
     safety_rules_manager: SafetyRulesManager,
     processor: Option<RoundProcessor>,
+    reconfig_events: libra_channel::Receiver<(), OnChainConfigPayload>,
 }
 
 impl EpochManager {
@@ -90,6 +91,7 @@ impl EpochManager {
         txn_manager: Arc<dyn TxnManager>,
         state_computer: Arc<dyn StateComputer>,
         storage: Arc<dyn PersistentLivenessStorage>,
+        reconfig_events: libra_channel::Receiver<(), OnChainConfigPayload>,
     ) -> Self {
         let author = node_config.validator_network.as_ref().unwrap().peer_id();
         let config = node_config.consensus.clone();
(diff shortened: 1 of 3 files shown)


consensus/src/epoch_manager.rs
```diff
@@ -263,8 +266,10 @@ impl EpochManager {
             .context(format!(
                 "[EpochManager] State sync to new epoch {}",
                 ledger_info
-            ))
-        // state_computer notifies reconfiguration in another channel
+            ))?;
+
+        monitor!("reconfig", self.expect_new_epoch().await);
```
aching, line 271: Since GitHub won't let me comment above, is the flow of how the epoch transition described in detail somewhere - e.g. receiving an EpochChangeProof from another validator -> updating storage and then state sync before processing any more messages and why we do this?  It's a little confusing.

aching (approved):
Am I correct in understanding that by continuing to participate in the previous epoch instead of moving into the new epoch, this validator would become malicious and help with a fork?

As mentioned in my comment, it would be great if the workflow of how and why this occurs is detailed somewhere (preferably nearby).

Nice find though with the testing!

zekun000:
no, it's not malicious and won't help with fork, it'll panic when commit some already committed txn.

epoch transition is detailed in the spec. the specific details is not documented, but it's pretty straightforward in the code.

zekun000:
did I miss anything in the summary? I wrote how and why about the root cause.

aching:
So the spec gives the high level idea of the epoch transition, but the reasoning of why we go through this blocking flow is confusing to me.  If it were simple, this bug would probably not have been occurred in the first place.

PR #6098, opened 2020-09-18
[docker] Remove traces of config-builder

davidiw:
* Remove validator-dynamic, the pinacle of config-builder
* Remove building / copying config-builder
* Remove shell scripts as they almost all use config-builder and we execute commands directly within the container
* Stripped out most of the mint test code but we'll add that back via a proper e2e test in compose in the near future... that test wasn't really ... testing anything


rexhoffman (approved):
Nicely done.   A few files may conflict with what I landed, but just in comments.

docker/mint/Dockerfile
```diff
@@ -39,53 +39,18 @@ RUN mkdir -p /opt/libra/bin  /libra/client/data/wallet/
...
-    COUNTER=$(( $COUNTER - 1 )); \
-    done; \
-    if [ -z $PIDS ]; then exit 1; else kill -9 $PIDS || true; fi
-
 FROM pre-test as prod
 # Mint proxy listening address
 EXPOSE 8000
 
-# Define CFG_SEED, AC_HOST, AC_PORT and CFG_CHAIN_ID environment variables when running
-CMD /opt/libra/bin/docker-run.sh
```
sausagee, line 87: We can remove the config-builder but not this startup script yet. I think the faucet server is using the CLI based implementation.  cc @mgorven

mgorven: k8s doesn't use any of these wrapper scripts, this is fine to remove.

mgorven (approved):

sausagee (approved):

docker/mint/Dockerfile
davidiw: no one is using these :/
they probably have some value in indicating what is going on, but i would propose we just update the Dockerfile to have an explicit command at some point if that's the route we want to support.... alternatively we just leverage docker-compose and slowly get that into CI. That way we ensure that our ops infra has better test coverage.

sausagee: docker compose sounds like a good follow-up.

sausagee:
@davidiw the LBT issue is fixed, you can retry land.

davidiw:
> @davidiw the LBT issue is fixed, you can retry land.

There's a couple other hairy issues that I'm trying to weed out at the same time :P, hopefully this latest tweak gets the rest of them... and then we can finally say good bye to the scary side of config-builder.

bmwill:
Sooooo Bors only listens to one command (the first one) per message. This is something that could easily be changed but we just didn't get around to it. I manually added the hi-pri label though

davidiw:
Does bors rebase this on current master for lbt and other tests or does it use this as is? Because I fear I didn't properly rebase this to pass lbt :/

bmwill:
@davidiw Bors always rebases onto the latest master prior to running tests

davidiw:
sweet

PR #6103, opened 2020-09-18
[libra framework] Specs for prologue, add a couple tests

tzakian:
Add specs for the `LibraAccount::prologue`, and make sure all aborts conditions are covered (line numbers/match branches in the Rust handler are referenced for each non-genesis abort condition).

Adds a couple functional tests, and adds an exists check on an unguarded `borrow_global`.


language/stdlib/modules/LibraAccount.move
```diff
@@ -1238,14 +1357,19 @@ module LibraAccount {
             sender_account.sequence_number < (MAX_U64 as u64),
             Errors::limit_exceeded(ESEQUENCE_NUMBER)
         );
+        assert(
+            sender_account.sequence_number == txn_sequence_number,
+            Errors::invalid_argument(ESEQUENCE_NUMBER)
```
tzakian, line 1362: In the process of specifying the epilogue I wound up coming across this and wonder what connection should `txn_sequence_number` have w.r.t. the the account's sequence number? Should `txn_sequence_number` not be passed in and we bump it manually here? 

What are your thoughts @runtian-zhou  @tnowacki and @bob-wilson?

tnowacki, line 1362: Can this actually be hit?

We should probably comment all of these asserts for which are possible to hit and which are for invariants

tzakian: I ran all the tests and nothing broke with this change. From the Move perspective there's no reason why an invalid `txn_sequence_number` could be passed in. My opinion is that if there's no way this can be hit, then `txn_sequence_number` shouldn't be an argument and we should directly increment the account's sequence number.

tnowacki: My guess is that @sblackshear added this code? I don't know the history of it. If its not used now, I'd rather just not pass it in, but we should probably wait to hear from him

language/stdlib/modules/LibraAccount.move
```diff
@@ -1008,8 +1008,13 @@ module LibraAccount {
 
     /// Return the current balance of the account at `addr`.
     public fun balance<Token>(addr: address): u64 acquires Balance {
+        assert(exists<Balance<Token>>(addr), Errors::not_published(EPAYER_DOESNT_HOLD_CURRENCY));
         balance_for(borrow_global<Balance<Token>>(addr))
     }
+    spec fun balance {
+        pragma verify;
```
wrwg, line 1015: Don't need this pragma here and elsewhere anymore. Please remove.

language/stdlib/modules/LibraAccount.move
```diff
@@ -1110,6 +1115,41 @@ module LibraAccount {
             chain_id,
         )
     }
+    spec fun module_prologue {
+        pragma verify;
+        let transaction_sender = Signer::spec_address_of(sender);
+        let max_transaction_fee = txn_gas_price * txn_max_gas_units;
+        include AbortsIfModulePrologue<Token> {
+            sender,
```
wrwg, line 1123: You don't need to pass this. A name in the context is automatically matched. You would get an error if not.

I'm actually surprised this even works (only name: value pairs should work). I guess it is just ignored where an error should have been produced if you use this notation.

wrwg (approved):

language/stdlib/modules/LibraAccount.move
tzakian: ah! awesome. I'll update :)

language/stdlib/modules/LibraAccount.move
sblackshear: Good question. I think the reasoning for passing in `txn_sequence_number` and using `txn_sequence_number + 1` at 1241 in the old code (1382 in the new code) is that it guarantees (in conjunction with the prologue) that the sequence number increases by exactly 1. If we don't pass it in and increment the account sequence number instead, it is possible that code in the script (or WriteSet!) already increased the sequence number (or rewrote it to a smaller value in the WriteSet case).

I think continuing to pass in `txn_sequence_number`, keeping this assertion, and also keeping the code at 1241 probably makes the most sense.

language/stdlib/modules/LibraAccount.move
tzakian: Updated, but it seems like the name context doesn't take into account local (`let`) bindings. Not sure if this is meant to be the case or not?

PR #6105, opened 2020-09-18
[language] increase deserializer test coverage

dariorussi:
Covering few malformed binaries in deserializer.
This increases the coverage a bit (we are over 92%).
There are few more error case I want to think about, as it
gets harder and harder to come up with test cases

### DIFF
--- a/language/vm/src/unit_tests/signature_token_tests.rs
+++ b/language/vm/src/unit_tests/signature_token_tests.rs
@@ -14,29 +14,32 @@ fn serialize_and_deserialize_nested_types_max() {
     let mut ty = SignatureToken::Struct(StructHandleIndex::new(0));
     for _ in 1..SIGNATURE_TOKEN_DEPTH_MAX {
         ty = SignatureToken::Vector(Box::new(ty));
+        let mut binary = BinaryData::new();
+        serialize_signature_token(&mut binary, &ty).expect("serialization should succeed");
+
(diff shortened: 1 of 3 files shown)


language/vm/src/unit_tests/signature_token_tests.rs
```diff
@@ -14,29 +14,32 @@ fn serialize_and_deserialize_nested_types_max() {
...
+        let mut cursor = Cursor::new(binary.as_inner());
+        load_signature_token(&mut cursor).expect("deserialization should succeed");
+    }
 }
 
 #[test]
 fn serialize_nested_types_too_deep() {
     let mut ty = SignatureToken::Struct(StructHandleIndex::new(0));
-    for _ in 0..SIGNATURE_TOKEN_DEPTH_MAX {
+    for _ in 1..SIGNATURE_TOKEN_DEPTH_MAX {
```
tzakian, line 28: Out of curiosity, is there also another test that makes sure that nested types are allowed up to this limit?

tzakian (approved):
Looks good to me

language/vm/src/unit_tests/signature_token_tests.rs
dariorussi: no and I plan to add one, absolutely good point!
But overall I realized that coverage in deserializer is pretty high so I figured I'd spend the next few days looking for other places where we may not have a good enough coverage

PR #6106, opened 2020-09-18
[backup] `restore auto` able to resume from previous failure

msmouse:
## Motivation

so that a network glitch won't ruin everything.

1. epoch endings and state snapthots are small in V1, this makes sure they can be overwritten.
2. skip transactions when possible, in whole batches though, meaning there will still be some overwriting

Y

## Test Plan

playing locally by Ctrl-C ing
will add fail points in smoke tests as a follow up.

## Related PRs

### DIFF
--- a/storage/jellyfish-merkle/src/restore/mod.rs
+++ b/storage/jellyfish-merkle/src/restore/mod.rs
@@ -155,22 +155,42 @@ where
     S: 'a + TreeReader + TreeWriter,
 {
     pub fn new(store: &'a S, version: Version, expected_root_hash: HashValue) -> Result<Self> {
-        let (partial_nodes, previous_leaf) = match store.get_rightmost_leaf()? {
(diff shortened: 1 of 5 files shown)


storage/jellyfish-merkle/src/restore/mod.rs
```diff
@@ -155,22 +155,42 @@ where
...
-            Some((node_key, leaf_node)) => {
+        Self::new_impl(
+            store,
+            version,
+            expected_root_hash,
+            true, /* try_resume */
+        )
+    }
+
+    pub fn new_overwrite(
```
lightmark, line 166: no caller?

lightmark (approved):
LGTM.

storage/jellyfish-merkle/src/restore/mod.rs
msmouse: force pushed during your review :P

PR #6116, opened 2020-09-21
CI: add cargo audit dependency check

fzzr-:
Add [cargo-audit](https://crates.io/crates/cargo-audit) auto-checks as [gh-action](https://github.com/marketplace/actions/rust-audit-check).

## Motivation

Automated checks vulnerable or stale dependencies is necessary at all by security and stability reasons.

Also the action produces neat report ([example](https://github.com/fzzr-/libra/runs/1143710509?check_suite_focus=true)).

Yup.

## Test Plan

This is small extra part for CI.

### DIFF
--- a/.github/workflows/cargo-audit.yml
+++ b/.github/workflows/cargo-audit.yml
@@ -1,0 +1,24 @@
+name: Audit
+on:
+  pull_request:
+    types:
+      - opened
+      - synchronize
+  pull_request_review:
+    types:
+      - submitted
+      - edited
+
+jobs:
+  rust:
+    name: Audit
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+
+      - name: cargo-audit
+        uses: actions-rs/audit-check@v1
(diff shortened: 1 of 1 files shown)


rexhoffman:
I'm reluctant to bring this in to the repo currently for a few reasons:
1) we don't have agreement w/ gha's for compute capacity, and we are running a cargo audit nightly in circleci.
2)Currently we fail the build if anything other than these 3 violations are found in circleci.  https://github.com/libra/libra/blob/master/.circleci/config.yml#L564-L583
3)I'm also a little reluctant to depend on gha that doesn't have the same config we use in circle, even if it doesn't currently break the build.  Cargo versions/flags/etc.
4) I'd need to look in to making sure versions are fully tied down, ideally to a git hash.

metajack:
Thanks for the contribution @fzzr-!

We do currently run cargo-audit as @rexhoffman has linked to, so there's no need to run another one in GHA. Should we decide to move our current job to GHA instead of CircleCI, we'll definitely keep actions-rs in mind.

PR #6117, opened 2020-09-21
[cli] Show private keys for constructed accounts when listing

metajack:
When working with automatically created accounts for testing, it is useful to be able to inspect the private keys. This PR adds them to the output when accounts are listed.

For example, these are useful if you are working with the API and need to construct accounts for testing that will sign something outside the CLI.

### DIFF
--- a/testsuite/cli/libra-wallet/src/key_factory.rs
+++ b/testsuite/cli/libra-wallet/src/key_factory.rs
@@ -104,6 +104,11 @@ impl ExtendedPrivKey {
         libra_types::account_address::from_public_key(&self.get_public())
     }
 
+    /// Get private key
+    pub fn get_private_key(&self) -> Ed25519PrivateKey {
+        self.private_key.clone()
+    }
+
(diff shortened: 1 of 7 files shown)


davidiw (approved):
it is for testing after all....

testsuite/cli/libra-wallet/src/key_factory.rs
```diff
@@ -104,6 +104,11 @@ impl ExtendedPrivKey {
         libra_types::account_address::from_public_key(&self.get_public())
     }
 
+    /// Get private key
+    pub fn get_private_key(&self) -> Ed25519PrivateKey {
+        self.private_key.clone()
```
davidiw, line 109: looks like you'll need to return a reference or perhaps you want to relabel this to something more insidious such as get_private_key_value and convert it into bytes

metajack: I just enabled `features = ["fuzzing"]` for libra-crypto.

PR #6118, opened 2020-09-21
[move-cli] better explanations when script args don't match parameter…

sblackshear:
… types

Previously, users saw only TYPE_MISMATCH, NUMBER_OF_TYPE_ARGUMENTS_MISMATCH, or LINKER_ERROR, which are not very descriptive.


tzakian (approved):
LGTM!

PR #6119, opened 2020-09-21
[vm] remove TODO for checking script arguments with instantiated gene…

sblackshear:
…ric types

There is an unused `_ty_args: Vec<TypeTag>` argument and a TODO for checking these type arguments against script arguments that refer to them (e.g., `main<T>(t: vector<T>)`. However, the bytecode verifier (rightly) requires that all parameter types are closed, and we do not plan to change this in the future. This PR simply removes the unused argument and the TODO to reflect this.


dariorussi (approved):
looks good

PR #6120, opened 2020-09-21
[move-cli] datatest testing harness + some useful tests

sblackshear:
Introducing a basic datatest testing framework for the CLI. It expects directories that contain an `args.txt` file with arguments that the `move` binary understands (one set of arguments per line). The testing framework runs the commands, compares the result to the expected output, and runs `move clean`.

I added tests for each of the major subcommands (compile, run, view) and their arguments as well as tests for error explanations (both runtime and compile-time) and one fairly meaty Libra smoke test.

In the process, I found a few minor bugs/oddities that I fixed:
- Error explanation sometimes failed to find the relevant module when `commit` is false due to a bug in `resolve_function` (now fixed)
- The CLI create empty move_src and move_data directories in many cases. I fixed the former, but the latter is still a problem.
- Introcuced some constants in lib.rs and moved constants from main.rs there
- Updated the README

## Motivation

(Write your answer here.)

## Test Plan

## Related PRs


sblackshear:
Note: this is the same PR as before, I somehow managed to delete the other one...

language/tools/move-cli/src/lib.rs
```diff
@@ -27,6 +28,13 @@ use std::{
     path::{Path, PathBuf},
 };
 
+pub const MOVE_SRC: &str = "move_src";
+pub const MOVE_DATA: &str = "move_data";
+/// Store modules under move_src without an explicit `address {}` block under 0x2.
```
tnowacki, line 33: I plan on just removing this feature really... 
Just haven't had the time due to the amount of tests that need to be changed

language/tools/move-cli/src/main.rs
```diff
@@ -101,33 +96,46 @@ fn maybe_create_dir(dir_name: &str) -> Result<&Path> {
...
 
     // allow user modules in move_src directory to compile against stdlib and move_data modules, but
     // not Libra framework modules
     // TODO: prevent cyclic dep?
-    let mut user_lib_files: Vec<String> = fs::read_dir(move_src)?
-        .map(|f| f.map_or("".to_string(), |f| f.path().to_string_lossy().into_owned()))
-        .filter(|p| p.ends_with(".move"))
-        .collect();
+    let mut user_lib_files: Vec<String> = if move_src.exists() {
+        fs::read_dir(move_src)?
```
tnowacki, line 111: `find_move_filenames` is public in move-lang if you want to use that

language/tools/move-cli/tests/cli_testsuite.rs
```diff
@@ -0,0 +1,144 @@
...
+// this is useful if you want to look at the `move_data` produced by
+// a test.
+const NO_MOVE_CLEAN: &str = "NO_MOVE_CLEAN";
+// if either of these env vars is set, the test harness overwrites the
+// old .exp files with the output instead of checking them against the
+// output.
+const UPDATE_BASELINE: &str = "UPDATE_BASELINE";
+const UB: &str = "UB";
+
+fn format_diff(expected: String, actual: String) -> String {
```
tnowacki, line 37: perhaps we should make this a public function in test utils. Probably fine for now though

language/tools/move-cli/src/main.rs
```diff
@@ -101,33 +96,46 @@ fn maybe_create_dir(dir_name: &str) -> Result<&Path> {
...
+    } else {
+        vec![]
+    };
 
-    if !user_lib_files.is_empty() {
-        println!("Compiling {:?} user module(s)", user_lib_files.len());
+    let has_user_modules = !user_lib_files.is_empty();
+
+    if has_user_modules {
+        print!("Compiling {:?} module(s)", user_lib_files.len());
```
tnowacki, line 122: Fine for now either way, but should this live behind a --verbose flag?

language/tools/move-cli/tests/cli_testsuite.rs
```diff
@@ -0,0 +1,144 @@
...
+            continue;
+        }
+        let cmd_output = Command::new(cli_binary_path.clone())
+            .current_dir(exe_dir)
+            .args(args_iter)
+            .output()?;
+        output += &format!("Command `{}`:\n", args_line);
+        output += std::str::from_utf8(&cmd_output.stdout)?;
+        output += std::str::from_utf8(&cmd_output.stderr)?;
+    }
```
tnowacki, line 100: This is rather cool. Might be worth breaking it out into a separate entry point in `move_cli`. Could be really neat to easily setup batched commands in the CLI and reuse them easily

tnowacki (approved):
Some feedback but nothing really worth blocking over. Ping me if you want another set of eyes after any changes

language/tools/move-cli/src/main.rs
sblackshear: Yeah, I have it on my plate to add a `verbose` or `quiet` flag. The printing is already a bit out of hand.

language/tools/move-cli/tests/cli_testsuite.rs
sblackshear: Ooh, I like this idea a lot-- a `move test` command would be very cool/useful for folks that want to write expected value tests for their custom modules/scripts. We could probably rewrite a lot of our Move VM functional tests using this (CC @vgao1996).

Another thing I'd like to to is make it possible to run the batched commands in the same process + redirect stderr/stdout to a local buffer instead of spawning a new process for every command in the batch (which is slow).

vgao1996: > We could probably rewrite a lot of our Move VM functional tests using this

Great idea. Functional tests were a copy of LLVM's tests and this is exactly how things work there. Although there was a very specific reason I didn't choose invoke the binaries at the beginning which I can't remember now... Anyways I'll review this and see how we'll make things work. In general I really like the idea of using the cli for testing or make the new move testing flow part of the cli.

language/tools/move-cli/src/lib.rs
sblackshear: Ah, gotcha. I will try to avoid using it in tests to decrease the pain for later...

language/tools/move-cli/tests/cli_testsuite.rs
sblackshear: I like this so much I already added preliminary support for it (see `test` subcommand + test.rs file, which contains most of what used to be in cli_testsuite.rs). Some work needs to be done to make `move test` support running a whole directory of tests instead of a single args.txt file (I couldn't figure out the datatest magic to make this work), but should be possible to fix in the future.

PR #6121, opened 2020-09-21
[move-prover] Aborts-with check mode.

wrwg:
This introduces a new property for `aborts_with` to check whether an aborts specification is complete:

```
aborts_with [check] code1, code2, ...
```

... checks whether the given function aborts with the given codes, independent of any other aborts specifications. That is, if there is already an `aborts_if P with code` it will not influence the outcome of the `aborts_with [check] codes` condition.

By default, the check can only verify whether there are any missing abort codes. In order to check whether they are obsolete codes which can never occur, a *negative* check needs to be performed by passing `--negative` to the prover's command line. Negative checks are based on a feature we added as part of Kevin's internship (we expect a verification error and report an error if it does *not* occur). Using the `--negative` flag is still experimental and may not work in all cases as expected.

Details:
- implements the `aborts_with [check]` functionality.
- introducing checks whether pragmas and properties are used correctly. Previously, a typo in a pragma or property would have been silently ignored; this is not longer the case.
- improvements on the negative verification infrastrucure.

## Motivation

Support for transaction script documentation.

Yes

## Test Plan

Added tests.

## Related PRs

NA

### DIFF
--- a/language/move-prover/spec-lang/src/env.rs
+++ b/language/move-prover/spec-lang/src/env.rs
@@ -37,10 +37,11 @@ use vm::{
 
 use crate::{
     ast::{
-        GlobalInvariant, ModuleName, PropertyBag, Spec, SpecBlockInfo, SpecFunDecl, SpecVarDecl,
-        Value,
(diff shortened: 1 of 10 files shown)


junkil-park (approved):
Great!

PR #6122, opened 2020-09-21
[vm] add new libra_vm_critical_errors counter

bob-wilson:
When the VM hits critical internal errors, we want to trigger alerts. Alerts are currently based on Prometheus counters, so this change adds a counter to use whenever something really bad happens inside the VM.

Yes

## Test Plan

Ran existing tests. No tests for this new counter yet.

### DIFF
--- a/language/libra-vm/src/counters.rs
+++ b/language/libra-vm/src/counters.rs
@@ -69,3 +69,9 @@ pub static TXN_GAS_USAGE: Lazy<Histogram> = Lazy::new(|| {
     )
     .unwrap()
 });
+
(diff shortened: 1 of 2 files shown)


dariorussi (approved):
thanks!
question, should we have  something else too that we alert only if going over a threshold? 
It may be the case that we may want more than one of this depending on the error, but meanwhile please think about it

bob-wilson:
We may indeed want to have alerts when something goes over a threshold, but at least for now, I don't know what those things would be or what threshold we would use. That might be something we need to do based on practical experience after the network is launched and scales up.

PR #6123, opened 2020-09-21
[LBT][compat] propogate validator image rename

sausagee:
## Motivation
Propagate validator image rename. 

## Test Plan
Canary


rustielin (approved):
Looks like it found the right image: 
image

lgtm! thanks for the quick fix

sausagee:
woohoo - canary passed

PR #6124, opened 2020-09-21
Replace rustls with native-tls

mgorven:
LSR can't use `rustls` due to certificate verification limitations. Since
we need `native-tls` anyway just use it everywhere instead of having two
TLS dependencies.

### DIFF
--- a/client/json-rpc/Cargo.toml
+++ b/client/json-rpc/Cargo.toml
@@ -23,4 +23,4 @@ libra-workspace-hack = { path = "../../common/workspace-hack", version = "0.1.0"
 
 [features]
 default = ["tls"]
-tls = ["reqwest/rustls-tls"]
+tls = ["reqwest/native-tls"]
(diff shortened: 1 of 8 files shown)


client/json-rpc/Cargo.toml
```diff
@@ -23,4 +23,4 @@ libra-workspace-hack = { path = "../../common/workspace-hack", version = "0.1.0"
 
 [features]
 default = ["tls"]
-tls = ["reqwest/rustls-tls"]
+tls = ["reqwest/default-tls"]
```
metajack, line 26: This isn't needed. default-tls is in the default feature.

metajack (changes requested):

client/json-rpc/Cargo.toml
mgorven: `reqwest` has `default_features=false` above however?

metajack: Sorry, I misread the diff and thought this was in workspace-hack dir.

metajack (approved):
/land

xli (approved):

PR #6125, opened 2020-09-21
 [vm] tests for passing too many and too few signers

sblackshear:
If the VM gets a script like `main() { }` and is passed one or more `signers`, it silently ignores them. If it is passed too few signers, it errors. These are the expected behaviors, but we didn't have tests for them previously.

### DIFF
--- a/language/move-vm/runtime/src/unit_tests/vm_arguments_tests.rs
+++ b/language/move-vm/runtime/src/unit_tests/vm_arguments_tests.rs
@@ -7,7 +7,7 @@ use move_core_types::{
     account_address::AccountAddress,
     gas_schedule::{GasAlgebra, GasUnits},
     identifier::Identifier,
-    language_storage::{ModuleId, StructTag},
+    language_storage::{ModuleId, StructTag, TypeTag},
 };
 use move_vm_types::{
     gas_schedule::{zero_cost_schedule, CostStrategy},
@@ -166,19 +166,22 @@ impl RemoteCache for RemoteStore {
     }
 }
 
-fn call_script(script: Vec<u8>, args: Vec<Value>) -> VMResult<()> {
+fn call_script_with_args_ty_args_signers(
+    script: Vec<u8>,
+    args: Vec<Value>,
+    ty_args: Vec<TypeTag>,
+    signers: Vec<AccountAddress>,
+) -> VMResult<()> {
     let move_vm = MoveVM::new();
     let remote_view = RemoteStore {};
     let mut session = move_vm.new_session(&remote_view);
     let cost_table = zero_cost_schedule();
     let mut cost_strategy = CostStrategy::system(&cost_table, GasUnits::new(0));
-    session.execute_script(
-        script,
-        vec![],
-        args,
-        vec![AccountAddress::random()],
-        &mut cost_strategy,
-    )
+    session.execute_script(script, ty_args, args, signers, &mut cost_strategy)
+}
+
+fn call_script(script: Vec<u8>, args: Vec<Value>) -> VMResult<()> {
+    call_script_with_args_ty_args_signers(script, args, vec![], vec![])
 }
 
 #[test]
@@ -335,9 +338,11 @@ fn check_main_signature() {
         SignatureToken::Bool,
         SignatureToken::Address,
     ]));
-    call_script(
+    call_script_with_args_ty_args_signers(
         script,
         vec![Value::bool(false), Value::address(AccountAddress::random())],
+        vec![],
+        vec![AccountAddress::random()],
     )
     .expect("&Signer first argument is good");
     let script = make_script(Signature(vec![
@@ -473,4 +478,69 @@ fn check_constant_args() {
             .major_status(),
         StatusCode::TYPE_MISMATCH,
     );
+}
+
+#[test]
+fn check_signer_args() {
+    let two_signer_script = make_script(Signature(vec![
+        SignatureToken::Reference(Box::new(SignatureToken::Signer)),
+        SignatureToken::Reference(Box::new(SignatureToken::Signer)),
+    ]));
+    // too few signers (0)
+    assert_eq!(
+        call_script_with_args_ty_args_signers(two_signer_script.clone(), vec![], vec![], vec![])
+            .err()
+            .unwrap()
+            .major_status(),
+        StatusCode::TYPE_MISMATCH
+    );
+
+    // too few signers (1)
+    let one_signer = vec![AccountAddress::random()];
+    assert_eq!(
+        call_script_with_args_ty_args_signers(
+            two_signer_script.clone(),
+            vec![],
+            vec![],
+            one_signer,
+        )
+        .err()
+        .unwrap()
+        .major_status(),
+        StatusCode::TYPE_MISMATCH
+    );
+
+    // too many signers (3)
+    let three_signers = vec![
+        AccountAddress::random(),
+        AccountAddress::random(),
+        AccountAddress::random(),
+    ];
+    assert_eq!(
+        call_script_with_args_ty_args_signers(
+            two_signer_script.clone(),
+            vec![],
+            vec![],
+            three_signers
+        )
+        .err()
+        .unwrap()
+        .major_status(),
+        StatusCode::TYPE_MISMATCH
+    );
+
+    // correct number of signers (2)
+    let two_signers = vec![AccountAddress::random(), AccountAddress::random()];
+    call_script_with_args_ty_args_signers(two_signer_script, vec![], vec![], two_signers)
+        .expect("Expected two signers, passing two should work");
+
+    // too many signers (1) in a script that expects 0 is ok
+    let no_signer_script = make_script(Signature(vec![SignatureToken::U8]));
+    call_script_with_args_ty_args_signers(
+        no_signer_script,
+        vec![Value::u8(10)],
+        vec![],
+        vec![AccountAddress::random()],
+    )
+    .expect("Ok to pass oo many signers");
 }


vgao1996:
> Use the Move VM (not Libra adapter) for tests that don't require signers (this may already be part of Victor's new testing plan)
Unfortunately the handling of signers is currently a Move VM behavior, not something controlled by the testing infra, so a Move VM testing flow will unlikely to save you. To make matters worse, the Libra/Move VM interfaces are also kinda inconsistent as revealed in #5784: the Move VM will fill in the signers for you in `execute_script`, but not in `execute_function`, and the Libra VM currently creates signer refs (vm values) and pass those in in those cases.

I guess we really need to think about what the desired VM behavior should be. I personally feel weird about signer-filling by the VM, but hasn't yet come up with a better plan.

language/move-vm/runtime/src/unit_tests/vm_arguments_tests.rs
```diff
@@ -474,3 +479,64 @@ fn check_constant_args() {
...
+    assert_eq!(
+        call_script_(two_signer_script.clone(), vec![], vec![], vec![])
+            .err()
+            .unwrap()
+            .major_status(),
+        StatusCode::TYPE_MISMATCH
+    );
+
+    // too few signers (1)
+    let one_signer = vec![AccountAddress::random()];
```
vgao1996, line 499: It probably doesn't matter here, but non-determinism in tests makes me panic

sblackshear:
> Unfortunately the handling of signers is currently a Move VM behavior, not something controlled by the testing infra, so a Move VM testing flow will unlikely save you. 

This PR fixes the Move VM behavior. But what we need is to make it possible to write tests that use the Move VM directly instead of going through the Libra adapter (especially when we're not testing Libra adapter behavior)

> I guess we really need to think about what the desired VM behavior should be. I personally feel weird about signer-filling by the VM, but hasn't yet come up with a better plan.

I think `signer`-filling. by the Move VM makes sense, but that it should fill in exactly the number of `signers` you pass to `execute_script` (instead of letting you pass extras, as this PR demonstrates).

language/move-vm/runtime/src/unit_tests/vm_arguments_tests.rs
```diff
@@ -166,19 +166,22 @@ impl RemoteCache for RemoteStore {
     }
 }
 
-fn call_script(script: Vec<u8>, args: Vec<Value>) -> VMResult<()> {
+fn call_script_(
```
tnowacki, line 169: maybe `call_script_and_specify_all_args` or something. Don't care too much since its for tests
Eventually this should just replace `call_script`

language/move-vm/runtime/src/unit_tests/vm_arguments_tests.rs
sblackshear: I was copying the style of the pre-existing tests here, but agreed that a hardcoded address would probably make more sense.

language/move-vm/runtime/src/runtime.rs
```diff
@@ -124,7 +124,16 @@ impl VMRuntime {
                 .map(Value::transaction_argument_signer_reference)
                 .collect()
         } else {
-            // no signer parameters, don't add to args
+            // no signer parameters
+            if false
+            /* senders.is_empty() */
```
vgao1996, line 129: Shouldn't it be `!senders.is_empty()`?

tnowacki, line 129: This was a crutch for migrating to `signer`s from the `move_to_sender` world, since we had no way of specifying what the signer should be, the Move VM had to infer it.

Eventually we could totally move from "all or nothing" to "all specified" for signers, but this would require adding a new argument structure to the transaction type. 

So I guess my point is I don't see the value in this PR right now since its a known problem (and this code isn't run :p) 
If we want to leave it here as a fancy TODO, I suppose that is okay

vgao1996 (commented):
> Without this, we would have to rewrite a ton of tests to add a dummy _account: &signer parameter.

Ah I get your point. We can make it that the test drive will not pass a default signer to the Move VM in the Move testing flow.

language/move-vm/runtime/src/runtime.rs
sblackshear: > Eventually we could totally move from "all or nothing" to "all specified" for signers, but this would require adding a new argument structure to the transaction type.

I actually don't think we would need to change this as long as we're ok with every Libra tx script having >0 `&signer`s (which seems like a reasonable restriction).

sblackshear: I'm ok with keeping or removing the dead code (perhaps better to kill the code + keep the comment), just wanted to demonstrate the fix.

sblackshear: Oops, good call.

tnowacki: > I actually don't think we would need to change this as long as we're ok with every Libra tx script having >0 &signers (which seems like a reasonable restriction).

I was going to speak to this but deleted the comment out of sadness. Its a lot of work to do that at the moment without a lot of time... I'd be on board for the change, but just not sure if it is worth it at this point

sblackshear: Agreed. Definitely not high-pri to fix (especially given the volume of work), but perhaps good to keep in mind as an item to make progress on as we're refactoring the tests.

dariorussi (changes requested):
this is tricky because getting the signer via the API should be a must in my opinion.
So I don't buy into the TODO. I understand the inconsistency in the behavior but I wonder if that even matters.
So not clear about any of this....

dariorussi:
> this is tricky because getting the signer via the API should be a must in my opinion.
> So I don't buy into the TODO. I understand the inconsistency in the behavior but I wonder if that even matters.
> So not clear about any of this....

to articulate this a bit more here is how I see this:
- the API takes the signers of the transaction. That is completely orthogonal to the function invoked taking the signer or not. That is a requirement of the API and I like that. I appreciate we need to talk about the meaning of the signer but that to me is a sound concept.
- the function called may or may not take signers. That is also perfectly fine. Now the only question left is: if it does indeed take signers what should the behavior be? fill as many as you can? or fail if the number does not match? I think all or nothing is a really sound approach so if that is what the code is doing I'd be fine as is

sblackshear:
>  if it does indeed take signers what should the behavior be? fill as many as you can? or fail if the number does not match

I think the latter. I view `signers` as the same as arguments/type arguments, which we insist match the procedure signature exactly.

> the API takes the signers of the transaction. That is completely orthogonal to the function invoked taking the signer or not. 

I think that makes sense from the perspective of the Libra adapter, but not from the perspective of the Move VM. The Move VM has no concept of a transaction; it just takes a `main` + some arguments and runs it. My position is that it should be the adapter's job to pass exactly the right number of `signers` (e.g., none if the script doesn't have any) and the VM should check that the signers + args + type args match the script.

dariorussi:
> > if it does indeed take signers what should the behavior be? fill as many as you can? or fail if the number does not match
> 
> I think the latter. I view `signers` as the same as arguments/type arguments, which we insist match the procedure signature exactly.
> 
> > the API takes the signers of the transaction. That is completely orthogonal to the function invoked taking the signer or not.
> 
> I think that makes sense from the perspective of the Libra adapter, but not from the perspective of the Move VM. The Move VM has no concept of a transaction; it just takes a `main` + some arguments and runs it. My position is that it should be the adapter's job to pass exactly the right number of `signers` (e.g., none if the script doesn't have any) and the VM should check that the signers + args + type args match the script.

I am not sure about any of this, as I already mentioned above. So let me try to be much more vocal and see where we land. Please read carefully and I hope I am not going to add to some frustration or misunderstanding (mine for instance, lol).

I agree that it is an open conversation and one we should have. I personally mentioned that since we introduced `Signer` so I agree it needs a talk and a decision at some point. I guess I failed to articulate my concerns at the time (if I did fail) so this may all come as a surprise to you and everybody else, who knows.

If the VM was oblivious to the signer that should come directly in the arguments and not via explicit addresses, with the VM doing a merging over something it had no clue. Incidentally we had code that did that in the past and I know some people in the team feel that way (hi @tzakian :))
Overall it is a question of what the `Signer` is to the VM. Today that resource primitive has no semantics to the VM except that provided from the framework, and that is a strange concept for a VM. Or at least strange to me (which may be more of a correct statement, and less of a conceptual and ominous absolute :) )

The VM having a concept of `Signer` with a well understood semantics and behavior is something that I can relate to. I understand that certain checks and operations over the signer are done in the adapter, but that does not change much for me. And an API that exposed that concept would be fine by me, if that existed (as it partially does now). It is clearly and explicitly expressing what those addresses mean, even though there may be no check for them. It would make clear the importance of them and how those could be related to something an adapter did or a VM may decide to do at some point. It would make it clear how they relate to arguments and how critical they are.
Again, a proper API around those would make everything consistent and coherent for me. It would make that primitive a concept in the VM people can easily relate to.

Now, one could look at the VM as purely a service for an adapter and a framework. Something that contains whatever hack (read it in the most positive sense please) is needed from one side or the other to operate properly. That view, though not shared by me **at all**, is an absolutely possible and fine view. It is something we need to talk about, characterize and decide on at some point soon (and later).
I already have issue in how `Vector` is exposed and `Signer` just adds another bad spin to what a VM knows and operates on.

I hope this makes some sense to you and I hope it is clear how that API evolved (and again possibly in an inconsistent way given only scripts take vector of senders). I understand (I am not even sure to be honest) your view but I hope we can all accept the fact that the issue needs resolution (through discussions and reflections) and not anyone locking down behavior now according to their understanding.

tnowacki:
I agree with @dariorussi. After sitting down and thinking about what we would likely want to do with the signer API for the VM (I think @dariorussi and I are on the same page here), the TODO is a bit nonsense, as all of this code would be refactored. I don't think its worth landing anything here besides maybe the test (I like more tests).

In other words: The TODO could be nice if we were progressing from the current design in the VM (it's a step forward in some sense). But we want to take two steps back and clean things up a bit before making any "forward" progress

sblackshear:
Fair enough; agreed that it makes sense to clean up and rethink the VM design at a higher level instead of leaving a TODO for a small fix like this. I will fix the PR to dump the TODO but keep the new tests.

dariorussi (approved):
thanks, sorry just saw it. And great test to add overall!

To add to the conversation I also realized that requiring clients to correctly pass the signers implies clients have to know the signature of `main` which is not what happens now and something we have avoided so far.
It would also require a client to deploy validators (rust code) if a main signature (a script) were to change.
None of that sound particularly friendly...

sblackshear:
> thanks, sorry just saw it. And great test to add overall!

Thanks!

> To add to the conversation I also realized that requiring clients to correctly pass the signers implies clients have to know the signature of `main` which is not what happens now and something we have avoided so far.

I would argue that clients already need to know the signature of `main` in order to pass the correct args and type args...

> It would also require a client to deploy validators (rust code) if a main signature (a script) were to change.
> None of that sound particularly friendly...

I view this as the same process a client would need to go through if we change the args or type args that a script requires.

PR #6128, opened 2020-09-21
[types] utility function for extracting coin name from a TypeTag

sblackshear:
Returns the name if the given TypeTag is a StructTag representing one of the coins accepted by Libra, None otherwise.


metajack (approved):
Thanks for the quick turnaround!

/land

PR #6129, opened 2020-09-21
[twins] general cleanup/reorganize

zekun000:
## Motivation

as title.

Y

## Test Plan

cargo xtest -p consensus

## Related PRs


consensus/src/lib.rs
```diff
@@ -29,7 +29,7 @@ mod state_replication;
 #[cfg(any(test, feature = "fuzzing"))]
 mod test_utils;
 #[cfg(test)]
-mod twins_test;
+mod twins;
```
davidiw, line 32: unless twin needs access to consensus internal bits, recommend making it an integration test, see: https://github.com/libra/libra/blob/master/consensus/safety-rules/Cargo.toml#L48

but do try to avoid requiring any feature flags ... I really need to refactor that one to avoid feature flags, sigh.....

consensus/src/twins/twins_node.rs
```diff
@@ -0,0 +1,214 @@
...
+        let (consensus_tx, consensus_rx) =
+            libra_channel::new(QueueStyle::FIFO, NonZeroUsize::new(8).unwrap(), None);
+        let (_conn_mgr_reqs_tx, conn_mgr_reqs_rx) = channel::new_test(8);
+        let (_, conn_notifs_channel) = conn_notifs_channel::new();
+        let network_sender = ConsensusNetworkSender::new(
+            PeerManagerRequestSender::new(network_reqs_tx),
+            ConnectionRequestSender::new(connection_reqs_tx),
+        );
+        let network_events = ConsensusNetworkEvents::new(consensus_rx, conn_notifs_channel);
+
```
davidiw, line 71: oh man... channels? for a single threaded framework? Is there no way to make this less complex :P

have you thought about making twins a wrapper around RoundProcessor (or whatever it is called these days?).

davidiw (approved):
I think we could make this simpler and much more accessible. Do we have a design doc on goals for this work? We could also consider this as a motivation for refactoring for an easier to test consensus. My opinion is that we should make this as simple and procedural as possible. No channels, no channels, no channels!

Anyway let's have this conversation in parallel to you doing whatever you're attempting to accomplish in the immediate.

consensus/src/lib.rs
zekun000: consensus doesn't have a binary to run like safety-rules.

consensus/src/twins/twins_node.rs
zekun000: you know networks are all about channels lol

zekun000:
\land

consensus/src/twins/twins_node.rs
davidiw: Ideally we shouldn't be using networks for this type of test unless you intend on running this on a real network

consensus/src/lib.rs
davidiw: my point was that keeping twins as a unit test is weird :P

PR #6130, opened 2020-09-21
[xbuild] install cargo based tooling as specified versions.

rexhoffman:
## Motivation

The first step of xbuild is making sure sccache is installed.   So implemented tooling in x to allow automatic installation of specific versions of cargo managed tooling.   

Right not tracking sccache and grcov, but more can easily be added.   Supports automatic install with:
```cargo x tools```, and can be checked with ```cargo x tools --check```.

Yes

## Test Plan

Currently tested manually, but looking at ways to simulate cargo commands, may add in the near future.

## Related PRs

None

### DIFF
--- a/devtools/x/src/config.rs
+++ b/devtools/x/src/config.rs
@@ -25,6 +25,7 @@ pub struct Config {
     fix: Fix,
     /// Cargo configuration
     cargo: CargoConfig,
+    tools: Vec<(String, String)>,
 }
 
 #[derive(Clone, Serialize, Deserialize, Debug, PartialEq, Eq)]
@@ -151,5 +152,9 @@ impl Config {
 
     pub fn warn_clippy_lints(&self) -> &[String] {
         &self.clippy.warn
+    }
+
+    pub fn tools(&self) -> &[(String, String)] {
+        &self.tools
     }
 }
--- a/devtools/x/src/installer.rs
+++ b/devtools/x/src/installer.rs
@@ -1,0 +1,53 @@
+// Copyright (c) The Libra Core Contributors
+// SPDX-License-Identifier: Apache-2.0
+
+use crate::{cargo::Cargo, config::CargoConfig};
+use log::info;
+use std::process::Command;
+
+pub fn install_if_needed(name: &str, version: &str) -> bool {
+    //This is a bit of a hack, given Cargo wants the CargoConfig struct, but we shouldn't be using the nightly
+    //cargo's tooling to build our 3rd party cargo tools.  Looking to address that in a refactoring pr once all the needed
+    //components are in place -- perhaps by seperating xcontext in to constituent parts.
+    let config = CargoConfig {
+        toolchain: "stable".to_owned(),
+        flags: Option::None,
+    };
+    install_if_needed_config(&config, name, version)
+}
+
+pub fn install_if_needed_config(config: &CargoConfig, name: &str, version: &str) -> bool {
+    if !check_installed(name, version) {
+        info!("Installing {} {}", name, version);
+        let installers = Cargo::new(config, "install")
+            .arg(name)
+            .arg("--version")
+            .arg(version)
+            .run();
+        if installers.is_err() {
+            info!(
+                "Could not install {} {}, check x.toml to ensure tool exists and is not yanked.",
+                name, version
+            );
+        }
+        installers.is_ok()
+    } else {
+        true
+    }
+}
+
+pub fn check_installed(name: &str, version: &str) -> bool {
+    let result = Command::new(name).arg("--version").output();
+    let found = match result {
+        Ok(output) => format!("{} {}", name, version)
+            .eq(String::from_utf8_lossy(&output.stdout.as_slice()).trim()),
+        _ => false,
+    };
+    info!(
+        "{} of version {} is{} installed",
+        name,
+        version,
+        if !found { " not" } else { "" }
+    );
+    found
+}
--- a/devtools/x/src/tools.rs
+++ b/devtools/x/src/tools.rs
@@ -1,0 +1,28 @@
+// Copyright (c) The Libra Core Contributors
+// SPDX-License-Identifier: Apache-2.0
+
+use crate::{context::XContext, installer, Result};
+use structopt::StructOpt;
+
+#[derive(Debug, StructOpt)]
+pub struct Args {
+    #[structopt(long)]
+    /// Run in 'check' mode. Exits with 0 if all tools installed. Exits with 1 and if not, printing failed
+    check: bool,
+}
+
+pub fn run(args: Args, xctx: XContext) -> Result<()> {
+    let tools = xctx.config().tools();
+    let mut failed: bool = false;
+    for (key, value) in tools {
+        let success = match args.check {
+            false => installer::install_if_needed(key, value),
+            true => installer::check_installed(key, value),
+        };
+        failed = failed || !success;
+    }
+    if failed {
+        std::process::exit(1);
+    }
+    Ok(())
+}
--- a/devtools/x/src/main.rs
+++ b/devtools/x/src/main.rs
@@ -19,8 +19,10 @@ mod diff_summary;
 mod fix;
 mod fmt;
 mod generate_summaries;
+mod installer;
 mod lint;
 mod test;
+mod tools;
 mod utils;
 
 type Result<T> = anyhow::Result<T>;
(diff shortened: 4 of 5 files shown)

### CODE REVIEW

File: devtools/x/src/build.rs [hunk 1]
```diff
@@ -0,0 +1,140 @@
...
+    #[structopt(long, short)]
+    /// Number of parallel jobs, defaults to # of CPUs
+    jobs: Option<u16>,
+    #[structopt(long, parse(from_os_str))]
+    /// Directory to output HTML coverage report to (using grcov)
+    html_cov_dir: Option<OsString>,
+    #[structopt(name = "TESTNAME", parse(from_os_str))]
+    testname: Option<OsString>,
+    #[structopt(name = "ARGS", parse(from_os_str), last = true)]
+    args: Vec<OsString>,
```
bmwill, line 44: This looks like it was mostly copied from `xtest` and there's a couple of these params that maybe don't make sense wrt wrapping `cargo build`.

devtools/x/src/build.rs
```diff
@@ -0,0 +1,140 @@
...
+}
+
+pub fn run(mut args: Args, xctx: XContext) -> Result<()> {
+    args.args.extend(args.testname.clone());
+    let config = xctx.config();
+
+    let env_vars: &[(&str, &str)] = if args.html_cov_dir.is_some() {
```
bmwill, line 51: Does `xbuild` need to know anything about code coverage output? seems like that's specific to `xtest`

devtools/x/src/installer.rs
```diff
@@ -0,0 +1,38 @@
...
+        toolchain: "stable".to_owned(),
+        flags: Option::None,
+    };
+    install_if_needed_or_die_config(&config, name, version);
+}
+
+pub fn install_if_needed_or_die_config(config: &CargoConfig, name: &str, version: &str) {
+    if !check_installed(name, version) {
+        info!("Installing {} {}", name, version);
+        let installres = Cargo::new(config, "install")
```
bmwill, line 19:
```suggestion
        let installers = Cargo::new(config, "install")
```

devtools/x/src/installer.rs
```diff
@@ -0,0 +1,38 @@
+// Copyright (c) The Libra Core Contributors
+// SPDX-License-Identifier: Apache-2.0
+
+use crate::{cargo::Cargo, config::CargoConfig};
+use log::info;
+use std::process::Command;
+
+pub fn install_if_needed_or_die(name: &str, version: &str) {
```
bmwill, line 8: Is this only supposed to install a different cargo version? (I'm assuming yes)

devtools/x/src/installer.rs
```diff
@@ -0,0 +1,38 @@
...
+    } else {
+        info!("Not re-installing {} {}", name, version);
+    }
+}
+
+pub fn check_installed(name: &str, version: &str) -> bool {
+    let result = Command::new(name).arg("--version").output();
+    let comp = format!("{} {}", name, version);
+    return result.is_ok()
+        && comp.eq(String::from_utf8_lossy(&result.unwrap().stdout.as_slice()).trim());
```
bmwill, line 37: This could probably be simplified a little bit using a `match` or `if let`

devtools/x/src/tools.rs
```diff
@@ -0,0 +1,36 @@
...
+    /// Run in 'check' mode. Exits with 0 if all tools installed. Exits with 1 and if not, printing failed
+    check: bool,
+}
+
+pub fn run(args: Args, xctx: XContext) -> Result<()> {
+    let tools = xctx.config().tools();
+    let mut failed: bool = false;
+    for (key, value) in tools {
+        if !args.check {
+            installer::install_if_needed_or_die(key, value);
```
bmwill, line 20: Why or die? might be better to have this return an error and handle the exit code in 1 single place (like you are doing below)

devtools/x/src/build.rs
rexhoffman: ah didn't mean to include this yet.

devtools/x/src/build.rs
rexhoffman: grcov is text, yes, seemed reasonable to allow it all be installed up front to potentially replace parts of dev_setup.sh.

devtools/x/src/tools.rs
rexhoffman: The plan is to use the installer in multiple locations, I suppose I can try to find a good one liner to wrap the calls when used in xtext/xbuild.    The `x tools` is mostly about pre-installation in CI environments, and if a dev wants to setup for offline work.

rexhoffman:
@bmwill addressed your concerns, but the logging is a little less user friendly, and a little more verbose now as a bit of code golf has to be played to make the use of return values you seemed to want with matches/if lets.   Ultimately I'm okay with since the extra logging every time someone will run xtest with coverage of xbuild with sccache will just inform them of the tools they are actually using.   Not attempting to address the cargo version mismatch we are in, as @metajack seems to think it will be resolved with the next release of rust.

devtools/x/src/tools.rs
rexhoffman: removed, the or_die bit.   will need a good one line when used in xtest/xbuild.

devtools/x/src/installer.rs
rexhoffman: Don't want to use the nightly version that could cause breakage in these released version of tools.

devtools/x/src/installer.rs
rexhoffman: used match, doesn't feel simpler to be, but probably more canonically rust.

devtools/x/src/installer.rs
rexhoffman: if we don't have the current cargo stable locally, totally fine with using that to build the latest tooling, since the stable version we're using isn't in x.toml -- though weirdly the nightly version is.   This all seems to need a reworking but not one I wanted to take on now with the redundant info in x.toml and cargo-toolchain.

devtools/x/src/installer.rs
```diff
@@ -0,0 +1,53 @@
...
+    }
+}
+
+pub fn check_installed(name: &str, version: &str) -> bool {
+    let result = Command::new(name).arg("--version").output();
+    let found = match result {
+        Ok(_) => format!("{} {}", name, version)
+            .eq(String::from_utf8_lossy(&result.unwrap().stdout.as_slice()).trim()),
+        Err(_) => false,
+    };
```
bmwill, line 45:
```suggestion
    let found = match result {
        Ok(output) => format!("{} {}", name, version)
            .eq(String::from_utf8_lossy(&output.stdout).trim()),
        Err(_) => false,
    };
```

bmwill: In order to remove the need to unwrap

devtools/x/src/config.rs
```diff
@@ -152,4 +153,8 @@ impl Config {
     pub fn warn_clippy_lints(&self) -> &[String] {
         &self.clippy.warn
     }
+
+    pub fn tools(&self) -> &[(String, String)] {
+        &self.tools.as_slice()
```
bmwill, line 158:
```suggestion
        &self.tools
```

devtools/x/src/installer.rs
```diff
@@ -0,0 +1,53 @@
...
+        Ok(output) => format!("{} {}", name, version)
+            .eq(String::from_utf8_lossy(&output.stdout).trim()),
+        Err(_) => false,
+    };
+    info!(
+        "{} of version {} is{} installed",
+        name,
+        version,
+        if !found { " not" } else { "" }
+    );
```
bmwill, line 51: So I don't think we want this to print out every time anyone ever runs xbuild or xtest. So this should either be a lower level or only output in the error case

rexhoffman: It wont, it'll only print out if we check for it.   Which will occur only when needed.   So I'm okay with it printing out before use.

bmwill: What I'm saying is I don't want a messages saying "sccache is installed" every time I run  `cargo xbuild`, assuming xbuild uses sccache. It makes sense to print a message out the first time something is installed, but not every time the tool is checked for

bmwill (approved):
I would just fix the last nit I had before landing.

devtools/x/src/installer.rs
rexhoffman: I'm invisioning the log line stating that sccache is installed at a certain version occurring right before the env vars are set and logged, and the build line is logged.   Ideally I want folks to know exactly what's going on underneath the covers when xbuild runs, and seeing the sccache log line makes sense to me.   Young has used the log line of what's being built in the past to split out tests in to different runs....  I just want people to be able to drop to the command line and do exactly the same thing as the build, so I don't get called in to debug someone's build if I can avoid it.

PR #6131, opened 2020-09-21
[language] reorganize move-lang and IR functional-tests

vgao1996:
This reorganizes the directory structure of move lang functional tests, putting them into three buckets:
1. `libra`: Tests for the Libra infrastructure.
2. `move`: Tests for the Move language (incl. parser, compiler, verifier, vm, stdlib).
3. `testing-infra`: Dummy tests ensuring the testing infra is functioning as intended.

This distinction will help us understand the coverage better now and will make it easier for us to move the tests to where they belong to when the new testing infra becomes available in the future.


dariorussi (approved):
this is good and a good distinction to have, thanks!

PR #6133, opened 2020-09-22
[breaking][vm] Fix issue where OUT_OF_GAS errors weren't being bubbled up correctly

tzakian:
This was causing scripts that ran `OUT_OF_GAS` to not be bubbled up to the `KeptVMStatus` correctly leading to the wrong error type in the `TransactionOutput`.

This also fixes an issue around location tracking in execute main where the location was being reported based off of the frame above where the issue was happening and not on the current frame.

### DIFF
--- a/language/move-lang/functional-tests/tests/epilogue/out_of_gas_in_nested_module.move
+++ b/language/move-lang/functional-tests/tests/epilogue/out_of_gas_in_nested_module.move
@@ -1,0 +1,34 @@
+module Swapper {
+    use 0x1::Vector;
+    public fun call(x: u64) {
+        swap_it_up(x)
+    }
+    public fun swap_it_up(vec_len: u64) {
+        let v = Vector::empty();
+
+        let i = 0;
+        while (i < vec_len) {
+          Vector::push_back(&mut v, i);
+          i = i + 1;
+        };
+
+        i = 0;
+
+        while (i < vec_len / 2) {
+            Vector::swap(&mut v, i, vec_len - i - 1);
+            i = i + 1;
+        };
+    }
+}
+
+//! new-transaction
+//! max-gas: 621
+script {
+use {{default}}::Swapper;
+fun main() {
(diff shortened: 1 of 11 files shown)


language/move-lang/functional-tests/tests/epilogue/out_of_gas_in_nested_module.move
```diff
@@ -0,0 +1,34 @@
...
+        let i = 0;
+        while (i < vec_len) {
+          Vector::push_back(&mut v, i);
+          i = i + 1;
+        };
+
+        i = 0;
+
+        while (i < vec_len / 2) {
+            Vector::swap(&mut v, i, vec_len - i - 1);
```
sblackshear, line 18: Lol, did you write the test this way because `swap` has a high gas cost?

sblackshear (approved):

language/move-vm/runtime/src/interpreter.rs
```diff
@@ -45,6 +45,12 @@ macro_rules! debug_writeln {
     };
 }
 
+macro_rules! set_err_info {
+    ($frame:ident, $e:expr) => {{
+        $e.at_code_offset($frame.function.index(), $frame.pc).finish($frame.location())
```
tnowacki, line 50: Looks good, lets look at the other uses of set_location before landing

language/move-lang/functional-tests/tests/epilogue/out_of_gas_in_nested_module.move
tzakian: I did indeed!

language/move-vm/runtime/src/interpreter.rs
tzakian: Had a look at the other uses of `set_location` and they all seem correct to me. Additionally any issues happening in the main interpreter loop for a frame have the correct location since it's being derived from the frame that's being executed.

PR #6135, opened 2020-09-22
[management] allow owner/operator support for adding new validator

davidiw:
I cannot directly write into release-0.21 so I need to do this... sigh.

We need this for Wednesday.


mgorven (approved):

mgorven:
This will need an update to the wrapper script: https://github.com/libra/partners/blob/master/scripts/libra-operations.sh

davidiw:
sigh... one time uses .... sigh...

davidiw:
well crap.... I don't know what to do here now...

PR #6139, opened 2020-09-22
[Logging] adding security logging of byzantine votes

AnomalRoil:
## Motivation

This adds a way to detect votes that display byzantine behavior, as was first attempted in #4997.

If two non-NIL block end up voting for two different consensus_data_hash at the same round, then it means that either:

 - the proposer sent different blocks to different voters
 - voters executed the same block differently

To detect this, we simply add a cache in pending_votes to save non-NIL votes by consensus_data_hash.
But this also require adding logic to differentiate NIL and non-NIL votes at the vote_data level.

Yes

## Test Plan

Ideally this could be tested using Twin tests to have it trigger the logs, I guess. But I need to investigate this possibility further.

### DIFF
--- a/common/logger/src/security.rs
+++ b/common/logger/src/security.rs
@@ -35,6 +35,9 @@ pub enum SecurityEvent {
     /// Consensus received an invalid message (not well-formed or incorrect signature)
     ConsensusInvalidMessage,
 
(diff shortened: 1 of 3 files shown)


davidiw:

PR #6143, opened 2020-09-22
[Libra Node] Refactor layout of logic and move library functionality out of main.

JoshLind:
## Motivation

This PR offers some small refactors to the layout of logic in Libra Node:
1. In the first commit we remove the `main_node.rs` file and merge it directly into the `lib.rs` file. It seems unnecessary to have any empty lib file that just imports main_node and adds a layer of indirection to imports.
2. In the second commit we move library code out of the `main.rs` file and into the `lib.rs` file. This will better enable future PRs to access the library code in the `main.rs` file (e.g., when wanting to test or call the `lib.rs` functionality without invoking `main()` directly)

Yes.

## Test Plan

All tests pass locally.

## Related PRs

None.


libra-action:

Breaking changes policy:

2. Adding fields is ok.

davidiw (commented):
does it make sense to move all of that logic out of main.rs? Are we really going to use structopt? I would think the library should expect a node config and that's it. We probably want main.rs also initializing logging. I also would like to get rid of this silly disable logging since that's something that rust_log already provides...

JoshLind:
> does it make sense to move all of that logic out of main.rs? Are we really going to use structopt? 

Well my understanding is that structopt is currently used for spawning nodes, right? e.g., libra swarm uses it when we call the binary? Were you thinking of another way to invoke binaries?

>I would think the library should expect a node config and that's it. We probably want main.rs also initializing logging. I also would like to get rid of this silly disable logging since that's something that rust_log already provides...

Yeah this is my mental mode, too. I think there's a few changes that we can make to help split things up better between main.rs and lib.rs. If it's okay with you, I'll tackle these in a follow up PR (when I have more time to sit down and split things cleanly). I was hoping for this PR to be more a refactor of code location and not mix in logical changes.

davidiw (approved):

JoshLind:
Thanks, @davidiw! 😄

/land

PR #6145, opened 2020-09-22
[twins] avoid timeout and support test case

zekun000:
## Motivation

- Avoid timeout and embed the timeout to the RoundProposer, in order to get rid of the timeout we have the following hacks.
  - Calculate timeout based on the partitions, if the leader partitions don't have enough votes, every one time out, else no one timeout.
  - In order to synchronize the rounds, playground doesn't drop timeout messages, and still sends the sync info from proposal if it's dropped.
- Support the TestCase format which is output by the generator.
- Add a test for the timeout from RoundProposer.
- Implement the execution.

(Write your answer here.)

## Test Plan

## Related PRs

### DIFF
--- a/config/src/config/consensus_config.rs
+++ b/config/src/config/consensus_config.rs
@@ -4,8 +4,10 @@
 use crate::config::SafetyRulesConfig;
 use libra_types::{account_address::AccountAddress, block_info::Round};
 use serde::{Deserialize, Serialize};
-use std::collections::HashSet;
-use std::{collections::HashMap, path::PathBuf};
+use std::{
+    collections::{HashMap, HashSet},
+    path::PathBuf,
+};
 
 #[derive(Clone, Debug, Deserialize, PartialEq, Serialize)]
 #[serde(default, deny_unknown_fields)]
--- a/config/src/config/consensus_config.rs
+++ b/config/src/config/consensus_config.rs
@@ -4,6 +4,7 @@
 use crate::config::SafetyRulesConfig;
 use libra_types::{account_address::AccountAddress, block_info::Round};
 use serde::{Deserialize, Serialize};
+use std::collections::HashSet;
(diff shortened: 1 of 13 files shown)


davidiw (commented):
we're changing prod code to support a testing case :O

zekun000:
@davidiw if you mean the RoundProposer, yes, but it's common across the codebase when we have test-only feature that's dispatched by config.

config/src/config/consensus_config.rs
```diff
@@ -55,7 +56,7 @@ pub enum ConsensusProposerType {
     // Pre-specified proposers for each round,
     // or default proposer if round proposer not
     // specified
-    RoundProposer(HashMap<Round, AccountAddress>),
+    RoundProposer(RoundProposerConfig),
```
aching, line 61: We should probably add to the comment, this is only for testing and not for prod like you did for RoundProposer below.

consensus/src/liveness/round_proposer_election.rs
```diff
@@ -4,31 +4,47 @@
...
 
-/// The round proposer maps a round to author
+/// The round proposer maps a round to author (Twins only testing)
 pub struct RoundProposer {
     // A pre-defined map specifying proposers per round
     proposers: HashMap<Round, Author>,
     // Default proposer to use if proposer for a round is unspecified.
     // We hardcode this to the first proposer
     default_proposer: Author,
+    // The rounds that're supposed to timeout (when it's in a different partition than the proposer)
```
aching, line 17:
```suggestion
    // The rounds that are supposed to timeout (when it's in a different partition than the proposer)
```

aching (commented):
It's an interesting set of hacks to make these tests more reliable.  How are you measuring this is more reliable than it was previously?

zekun000:
it's more reliable in theory, it's either deterministic pass or fail since there's no timeout.

PR #6146, opened 2020-09-22
[move-cli] support saving and viewing events

sblackshear:
- Added `events` subdirectory of move_data/<addr> for storing emitted events. Subdirectories of `events` store event handles named by their "creation number". For example, if account address 0xA creates two handles and emits two events to each, the CLI will create two files `0xA/events/0` and `0xA/events/1` and store the two contract events emitted to each handle in the appropriate file.
- Added similar `move_data/<addr>/resources` and `move_data/<addr>/modules` subdirectories to partition the storage under each address more cleanly
- `move run` now stores emitted events on disk instead of discarding them
- `move view` now supports viewing events


language/tools/move-cli/tests/testsuite/module_run_view/args.exp
```diff
@@ -6,10 +6,10 @@ Changed resource(s) under 2 address(es):
   Changed 1 resource(s) under address 0000000000000000000000000000000b:
     Added type 00000000::Test::R: [U64(7)]
 Discarding changes; re-run with --commit if you would like to keep them.
-Command `view move_data/0x0000000000000000000000000000000a/0x00000000000000000000000000000002::Test::R`:
-Resource not found.
-Command `view move_data/0x0000000000000000000000000000000b/0x00000000000000000000000000000002::Test::R`:
-Resource not found.
+Command `view move_data/0x0000000000000000000000000000000a/resources/0x00000000000000000000000000000002::Test::R`:
+Not found.
```
tzakian, line 10: Possible regression? Shouldn't this say "Resource not found."?

tzakian (approved):
This look great! One possible issue, but happy for this to land as-is.

language/tools/move-cli/tests/testsuite/module_run_view/args.exp
sblackshear: Ah, good call. Not sure if that check is needed; let me try eliminating it.

PR #6149, opened 2020-09-22
[Consensus] Verify vote data upon vote verification

AnomalRoil:
The insert_vote function claimed that it "Insert a vote and if the vote is valid, return a QuorumCertificate preferentially over a TimeoutCertificate if either can can be formed", however the verification of the validity was not performed.

This is making it verify the vote_data as it should according to https://github.com/libra/libra/tree/master/specifications/consensus#vote

yes

## Test Plan

--

### DIFF
--- a/common/logger/src/security.rs
+++ b/common/logger/src/security.rs
@@ -32,7 +32,7 @@ pub enum SecurityEvent {
 
     // Consensus
     // ---------
-    /// Consensus received an invalid message (not well-formed or incorrect signature)
+    /// Consensus received an invalid message (not well-formed, invalid vote data or incorrect signature)
     ConsensusInvalidMessage,
 
     /// Consensus received an equivocating vote
@@ -40,9 +40,6 @@ pub enum SecurityEvent {
 
     /// Consensus received an invalid proposal
     InvalidConsensusProposal,
-
-    /// Consensus received an invalid vote
-    InvalidConsensusVote,
 
     /// Consensus received an invalid new round message
     InvalidConsensusRound,
--- a/consensus/consensus-types/src/vote.rs
+++ b/consensus/consensus-types/src/vote.rs
@@ -138,6 +138,8 @@ impl Vote {
                 .verify(self.author(), &self.timeout(), timeout_signature)
                 .context("Failed to verify Timeout Vote")?;
         }
+        // Let us verify the vote data as well
+        self.vote_data().verify()?;
         Ok(())
     }
 }


AnomalRoil:
@zekun000 @mimoo I think this is ready for a pre-review and to discuss if it should be done or not.

Notice that currently the vote and vote data are only verified during the QC validation I believe. This is correcting this to also validate the vote upon insertion, to avoid wasting more CPU later and wasting whole QC on invalid votes.

zekun000 (changes requested):
the validation is performed at https://github.com/libra/libra/blob/master/consensus/src/round_manager.rs#L50-L66

all messages are first verified before processing. make sure you understand the flow before making changes in the future to avoid redundant work.

AnomalRoil:
Oh, right, I missed that check. The signatures and the vote data hash are verified, I'll remove the first check then.

But on the other hand, the vote data themselves are not verified as far as I can say:
https://github.com/libra/libra/blob/441d1e4b24bb7fc6bbd93d948e1ee148f812d07f/consensus/src/round_manager.rs#L57-L60 
is verifying the signatures by calling  
https://github.com/libra/libra/blob/441d1e4b24bb7fc6bbd93d948e1ee148f812d07f/consensus/consensus-types/src/vote_msg.rs#L52-L61 
which in turns calls 
https://github.com/libra/libra/blob/441d1e4b24bb7fc6bbd93d948e1ee148f812d07f/consensus/consensus-types/src/vote.rs#L126-L142
which verifies the signatures and the hash.

So I think this means the vote data themselves are not verified until the QC is being checked and it then checks the vote_data:
https://github.com/libra/libra/blob/441d1e4b24bb7fc6bbd93d948e1ee148f812d07f/consensus/consensus-types/src/quorum_cert.rs#L123

zekun000 (commented):
yes, I think the audit also pointed out we didn't verify the vote data from the vote msg, but why verify it here, just add it to the VoteMsg.verify()?

consensus/consensus-types/src/vote_msg.rs
```diff
@@ -54,6 +55,19 @@ impl VoteMsg {
             self.vote().epoch() == self.sync_info.epoch(),
             "VoteMsg has different epoch"
         );
+
+        // Let us verify the vote data
+        self.vote.vote_data().verify().map_err(|e| {
```
mimoo, line 60: why not have that in Vote.verify() ?

zekun000: + 1 to this, also I don't think we should send the SecurityEvent here, it's unified at the upper layer to be InvalidConsensusMsg

mimoo: yup no need for a log here

AnomalRoil: Yeah, I knew this last week and had time to forget about it over the weekend.

It's nice it allows me to remove the Security event from the list and also to avoid introducing a new dependency in consensus-type. :)

mimoo (approved):
nice, thanks for adding the check :)

AnomalRoil:
I'll let you land it if you approve it too @zekun000 :)

common/logger/src/security.rs
```diff
@@ -32,7 +32,7 @@ pub enum SecurityEvent {
 
     // Consensus
     // ---------
-    /// Consensus received an invalid message (not well-formed or incorrect signature)
+    /// Consensus received an invalid message (not well-formed, invalid vote data or incorrect signature)
```
zekun000, line 35: this is not only vote data, but also other msg types

zekun000 (commented):
/land

mimoo (approved):
LGTM

mimoo:
@zekun000 looks like it misses your approval

zekun000 (approved):

PR #6150, opened 2020-09-22
[network] clean up channel counter spam

phlip9:
+ Previously, we were using a unique counter for each peer for each channel that gets spun up for the `Peer` actor. This lead to a _ton_ of counters, even for very limited use cases like the validator network. For example, there are ~10 channel counters per `Peer`. In premainnet, each validator would be producing ~ 10 * 30 = 300 unique counters for these metrics. When collected to the shared metrics, you get ~ 10 * 30 * 30 = 9000 counters.

+ This diff removes the peer_gauge in favor of a single counter for each channel. Unfortunately this loses information (we can't isolate congestion to a single channel x peer); however, it's probably fine...

+ Clean up the channel docs and remove the shared global TEST_COUNTER so the unit tests aren't as clunky.

+ Remove some unused functionality that would allow users to configure a "timeout" so messages older than "timeout" would just get silently dropped. No one was using this, so I took the liberty of removing this.

+ Network no longer uses the deprecated OpMetrics : )

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -576,12 +576,9 @@ version = "0.1.0"
 dependencies = [
  "anyhow 1.0.32 (registry+https://github.com/rust-lang/crates.io-index)",
  "futures 0.3.5 (registry+https://github.com/rust-lang/crates.io-index)",
- "libra-logger 0.1.0",
  "libra-metrics 0.1.0",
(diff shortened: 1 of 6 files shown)


bmwill (approved):
Looks good!

PR #6151, opened 2020-09-22
[logging] Add support for display/debug structured log fields

gregnazario:
Many times the identifier based one off structured logs require
conversion of objects to debug or strings before using, and cannot be
displayed with the serializer.  Debug can now be used by putting a ?
before the identifier, and a % before the identifier for display.  

This will also prevent extra allocations of strings.

Examples
```
        error!(name = ?Vec::new(), "Debug");
        error!(name = %DisplayOnlyStruct {}, "Display");
```

### DIFF
--- a/common/logger/src/libra_logger.rs
+++ b/common/logger/src/libra_logger.rs
@@ -509,5 +509,35 @@ mod tests {
         handler.join().unwrap();
         let entry = receiver.recv().unwrap();
         assert_eq!(entry.thread_name.as_deref(), Some("named thread"));
+
+        // Test Debug and Display inputs
+        let debug_struct = DebugStruct {};
+        let display_struct = DisplayStruct {};
+
+        error!(identifier = ?debug_struct, "Debug test");
+        error!(identifier = ?debug_struct, other = "value", "Debug2 test");
+        error!(identifier = %display_struct, "Display test");
+        error!(identifier = %display_struct, other = "value", "Display2 test");
+        error!("Literal" = ?debug_struct, "Debug test");
+        error!("Literal" = ?debug_struct, other = "value", "Debug test");
+        error!("Literal" = %display_struct, "Display test");
+        error!("Literal" = %display_struct, other = "value", "Display2 test");
+        error!("Literal" = %display_struct, other = "value", identifier = ?debug_struct, "Mixed test");
+    }
+
+    struct DebugStruct {}
+
+    impl std::fmt::Debug for DebugStruct {
+        fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+            write!(f, "DebugStruct!")
+        }
+    }
+
+    struct DisplayStruct {}
+
+    impl std::fmt::Display for DisplayStruct {
+        fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+            write!(f, "DisplayStruct!")
+        }
     }
 }
--- a/common/logger/src/macros.rs
+++ b/common/logger/src/macros.rs
@@ -99,6 +99,34 @@ macro_rules! schema {
         )
     };
 
+    // Identifier Keys debug
+    (@ { $(,)* $($out:expr),* }, $($k:ident).+ = ?$val:expr, $($args:tt)*) => {
+        $crate::schema!(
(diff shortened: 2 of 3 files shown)


common/logger/src/libra_logger.rs
```diff
@@ -366,6 +366,7 @@ mod tests {
         Schema, Value, Visitor,
     };
     use chrono::{DateTime, Utc};
+    use serde::export::Formatter;
```
bmwill, line 369: What is this needed for?

common/logger/src/macros.rs
```diff
@@ -99,6 +99,34 @@ macro_rules! schema {
         )
     };
 
+    // Identifier Keys debug
+    (@ { $(,)* $($out:expr),* }, ?$($k:ident).+ = $val:expr, $($args:tt)*) => {
+        $crate::schema!(
+            @ { $($out),*, &$crate::KeyValue::new($crate::__log_stringify!($($k).+), $crate::Value::from_serde(&format!("{:?}", $val))) },
```
bmwill, line 105:
```suggestion
            @ { $($out),*, &$crate::KeyValue::new($crate::__log_stringify!($($k).+), $crate::Value::from_debug(&$val)) },
```

Don't allocate, instead use the from_debug/from_display functions on the Value type

bmwill (changes requested):
I would probably prefer to have the sigil on the value itself, instead of the key. At least that is how both tracing and slog handle it. And it makes it clearer that the value is what needs to use Display or Debug, not the key.

common/logger/src/macros.rs
```diff
@@ -99,6 +99,34 @@ macro_rules! schema {
...
+    };
+
+    (@ { $(,)* $($out:expr),* }, !$($k:ident).+ = $val:expr) => {
+        $crate::schema!(
+            @ { $($out),*, &$crate::KeyValue::new($crate::__log_stringify!($($k).+), $crate::Value::from_serde(&$val.to_string())) },
+        )
+    };
+
     // Literal Keys
     (@ { $(,)* $($out:expr),* }, $k:literal = $val:expr, $($args:tt)*) => {
```
bmwill, line 131: We probably also want to handle the literal case

common/logger/src/libra_logger.rs
gregnazario: Hmm, my IDE must have put this in

common/logger/src/macros.rs
gregnazario: sound good.

common/logger/src/macros.rs
gregnazario: Btw, what's the reason we support literals?

bmwill: I mean we don't have to (i don't know if we actually use them anywhere) but its for the case where you want to name a field with a symbol that's not part of an identifier. E.g. "@timestamp"

gregnazario:
> I would probably prefer to have the sigil on the value itself, instead of the key. At least that is how both tracing and slog handle it. And it makes it clearer that the value is what needs to use Display or Debug, not the key.

This is actually handled through where an inner macro puts the `sigil` before the key.  The issue is having the `sigil` in front of an expression causes the compiler to complain.

https://docs.rs/tracing/0.1.19/src/tracing/macros.rs.html#637-644

gregnazario:
> I would probably prefer to have the sigil on the value itself, instead of the key. At least that is how both tracing and slog handle it. And it makes it clearer that the value is what needs to use Display or Debug, not the key.

Done.

common/logger/src/macros.rs
```diff
@@ -99,6 +99,34 @@ macro_rules! schema {
         )
     };
 
+    // Identifier Keys debug
+    (@ { $(,)* $($out:expr),* }, $($k:ident).+ = ?$val:expr, $($args:tt)*) => {
+        $crate::schema!(
+            @ { $($out),*, &$crate::KeyValue::new($crate::__log_stringify!($($k).+), $crate::Value::from_debug(&format!("{:?}", $val))) },
```
bmwill, line 105: You should remove the format macro and the to_string calls here. The values themselves should impl Debug/Display so we shouldn't have to allocate before the call

gregnazario: Hmm, I thought I did, but they got themselves back in there when I had to revert because I couldn't figure out what was wrong with the macros

bmwill (approved):
Looks great!

/land

PR #6154, opened 2020-09-22
swarm: better error messages when failing to start processes

bmwill:
Also require the binary paths passed via command line instead of assuming they're installed.


xli:
Do we have a script to wrap what need to be built and set arguments for the libra-swarm? 
That way new arguments will not required. 

Also, we may need update doc at: https://developers.libra.org/docs/run-move-locally.

davidiw (approved):

xli (approved):

bmwill:
/land

Yeah we probably want to update some of the docs with @davidiw fancy new `libra-node --test` for most cases.

PR #6157, opened 2020-09-23
[secure-net] logs and counters

davidiw:
since secure net hides internal issues so that clients and servers cannot distinguish between running within the same thread, across threads, or across processes, we need a little internal information, this surfaces that. I don't see this as critical from an operators perspective, but it is definitely critical from a debugging perspective.

### DIFF
--- a/secure/net/src/lib.rs
+++ b/secure/net/src/lib.rs
@@ -15,13 +15,122 @@
 //! Internally both the client and server leverage a NetworkStream that communications in blocks
 //! where a block is a length prefixed array of bytes.
 
-use libra_logger::{debug, trace};
+use libra_logger::{info, trace, warn, Schema};
+use libra_secure_push_metrics::{register_int_counter_vec, IntCounterVec};
+use once_cell::sync::Lazy;
+use serde::Serialize;
 use std::{
     io::{Read, Write},
     net::{Shutdown, SocketAddr, TcpListener, TcpStream},
     thread, time,
 };
 use thiserror::Error;
+
+#[derive(Schema)]
+struct SecureNetLogSchema<'a> {
+    service: &'static str,
+    mode: NetworkMode,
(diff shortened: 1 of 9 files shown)


libra-action:
```
target packages:
  A libra-secure-push-metrics 0.1.0 (workspace, path 'secure/push-metrics')
    * features: [none]
```

secure/net/src/lib.rs
```diff
@@ -15,14 +15,123 @@
...
 };
 use thiserror::Error;
 
+#[derive(Schema)]
+struct SecureNetLogSchema<'a> {
+    service: &'static str,
+    mode: NetworkMode,
+    event: LogEvent,
+    #[schema(debug)]
+    remote_peer: Option<&'a SocketAddr>,
```
gregnazario, line 35: So, this is something that is not great about ES, is if you have two things that serialize to the same field in different ways, it could reject one of them.  As long as both `SocketAddr` and `PeerId` serialize to a single string it should be fine, but just a weird caveat.

secure/net/src/lib.rs
```diff
@@ -38,106 +147,187 @@ pub enum Error {
...
+                self.service,
+                NetworkMode::Client,
+                LogEvent::DisconnectedPeerOnWrite,
+            )
+            .error(&err)
+            .remote_peer(&self.server));
+
             self.stream = None;
+        } else {
+            self.increment_counter(Method::Write, MethodResult::Success);
```
gregnazario, line 223: Love to see the counters

gregnazario (approved):

davidiw:
/canary

lemme verify before landing

libra-action:
```
target packages:
  A libra-secure-push-metrics 0.1.0 (workspace, path 'secure/push-metrics')
    * features: [none]
```

PR #6158, opened 2020-09-23
[Smoke Tests] Split up the massive smoke-tests.rs file into separate component files.

JoshLind:
## Motivation

The `smoke-tests.rs` file has become very large (~2.3k lines of code!). It has also become very messy: the file contains everything from smoke tests themselves, to test utility functions, test environment logic, and everything in between. This, unfortunately, makes it difficult to reason about what type of smoke tests we have, and what functionality is actually being tested.

To address this, this PR splits up the single `smoke-tests.rs` file into separate individual files, grouped by test functionalities and different components. To achieve this, the PR offers two commits:
1. First, we separate the single `smoke-tests.rs` file into several smaller test files, grouped by test component/functionalities (e.g., `operational_tooling.rs`, `state_sync.rs`, `storage.rs` and `full_nodes.rs` etc.). In addition, we also create a `test_environment.rs` file and a `test_utils.rs` file to contain the test environment logic and frequently used helper functions.
2. Second, we rename the "TestEnvironment" struct to "SmokeTestEnvironment".

Note: nothing should semantically change in this PR, except for the renaming and layout. Having split the tests up, we can now identify further refactors and improvements that should be done. However, these will follow in subsequent PRs. Splitting things up should land first.

Yes.

## Test Plan

All tests pass locally. Moreover, by examining the number of tests being performed before the refactor, as well as after the refactor, we can see that no test are lost or added by this change.

## Related PRs

None.


testsuite/smoke-test/src/client.rs
```diff
@@ -0,0 +1,223 @@
...
+};
+use libra_config::config::NodeConfig;
+use libra_trace::trace::trace_node;
+use libra_types::{
+    account_address::AccountAddress, account_config::testnet_dd_account_address,
+    ledger_info::LedgerInfo, waypoint::Waypoint,
+};
+
+#[test]
+fn smoke_test_multi_node() {
```
zekun000, line 15: why this file is called `client`?

JoshLind: Yeah, this file is a bit of a mixed bag of random tests. The only reason I called them client is because all the tests seem to be heavily focussed around performing client proxy operations and testing functionality (e.g., testing client sends, client account creation, client waypoints, client balance fetching, client event tracing etc.) If you have a better suggestion for how to group them, please let me know 😄 

One possibility would be to have several finer grained files:
1. `basic.rs` for basic functionality (e.g., fault tolerance and restartability)
2. `client.rs` for the client waypoint test and client trace support test.
3. `accounts_and_balances.rs` for tests that do basic things like create accounts, transfer funds and check balances.

zekun000: maybe 1&3 can combine together

zekun000 (approved):
thanks for taking the initiatives

bmwill (approved):
This is awesome! Let's just get this massive thing landing shall we?

/land

JoshLind:
Hah! Thanks @bmwill! 😄

@zekun000, I'll follow up with some clean up PRs, so in one of those I'll update the naming of the test files.

PR #6163, opened 2020-09-23
[storage] ensure client known_version no newer than leger version, earlier

msmouse:
## Motivation

1. avoids some in vein operations if it's gonna fail the `ledger_store.get_consistency_proof()` anyways later.
2. more readable error message than lower in the stack.

Y

## Test Plan

existing coverage


zekun000 (approved):
is this the only place we miss the check?

msmouse:
> is this the only place we miss the check?
yes the other calls this.

msmouse:
> is this the only place we miss the check?

yes the other calls this.

/land

PR #6165, opened 2020-09-23
[language] fix error codes in libra vm writeset generation

vgao1996:
This corrects a few errors codes during writeset generation in the the Libra VM. See my comments below for details.

### DIFF
--- a/language/libra-vm/src/libra_vm.rs
+++ b/language/libra-vm/src/libra_vm.rs
@@ -442,9 +442,9 @@ pub fn txn_effects_to_writeset_and_events_cached<C: AccessPathCache>(
             let op = match val_opt {
                 None => WriteOp::Deletion,
                 Some((ty_layout, val)) => {
-                    let blob = val
-                        .simple_serialize(&ty_layout)
-                        .ok_or_else(|| VMStatus::Error(StatusCode::VALUE_SERIALIZATION_ERROR))?;
+                    let blob = val.simple_serialize(&ty_layout).ok_or_else(|| {
+                        VMStatus::Error(StatusCode::UNKNOWN_INVARIANT_VIOLATION_ERROR)
+                    })?;
 
                     WriteOp::Value(blob)
                 }
@@ -459,7 +459,7 @@ pub fn txn_effects_to_writeset_and_events_cached<C: AccessPathCache>(
 
     let ws = WriteSetMut::new(ops)
         .freeze()
-        .map_err(|_| VMStatus::Error(StatusCode::DATA_FORMAT_ERROR))?;
+        .map_err(|_| VMStatus::Error(StatusCode::UNKNOWN_INVARIANT_VIOLATION_ERROR))?;
 
     let events = effects
         .events
@@ -467,7 +467,7 @@ pub fn txn_effects_to_writeset_and_events_cached<C: AccessPathCache>(
         .map(|(guid, seq_num, ty_tag, ty_layout, val)| {
             let msg = val
                 .simple_serialize(&ty_layout)
-                .ok_or_else(|| VMStatus::Error(StatusCode::DATA_FORMAT_ERROR))?;
+                .ok_or_else(|| VMStatus::Error(StatusCode::UNKNOWN_INVARIANT_VIOLATION_ERROR))?;
             let key = EventKey::try_from(guid.as_slice())
                 .map_err(|_| VMStatus::Error(StatusCode::EVENT_KEY_MISMATCH))?;
             Ok(ContractEvent::new(key, seq_num, ty_tag, msg))


language/libra-vm/src/libra_vm.rs
```diff
@@ -459,15 +459,15 @@ pub fn txn_effects_to_writeset_and_events_cached<C: AccessPathCache>(
 
     let ws = WriteSetMut::new(ops)
         .freeze()
-        .map_err(|_| VMStatus::Error(StatusCode::DATA_FORMAT_ERROR))?;
+        .map_err(|_| VMStatus::Error(StatusCode::UNKNOWN_INVARIANT_VIOLATION_ERROR))?;
```
vgao1996, line 462: IMO the Move and Libra VM combined should never fail the writeset generation process.

language/libra-vm/src/libra_vm.rs
```diff
@@ -442,9 +442,9 @@ pub fn txn_effects_to_writeset_and_events_cached<C: AccessPathCache>(
             let op = match val_opt {
                 None => WriteOp::Deletion,
                 Some((ty_layout, val)) => {
-                    let blob = val
-                        .simple_serialize(&ty_layout)
-                        .ok_or_else(|| VMStatus::Error(StatusCode::VALUE_SERIALIZATION_ERROR))?;
+                    let blob = val.simple_serialize(&ty_layout).ok_or_else(|| {
+                        VMStatus::Error(StatusCode::UNKNOWN_INVARIANT_VIOLATION_ERROR)
```
vgao1996, line 446: The Move VM should never return a layout that doesn't match the value.

language/libra-vm/src/libra_vm.rs
```diff
@@ -459,15 +459,15 @@ pub fn txn_effects_to_writeset_and_events_cached<C: AccessPathCache>(
...
+        .map_err(|_| VMStatus::Error(StatusCode::UNKNOWN_INVARIANT_VIOLATION_ERROR))?;
 
     let events = effects
         .events
         .into_iter()
         .map(|(guid, seq_num, ty_tag, ty_layout, val)| {
             let msg = val
                 .simple_serialize(&ty_layout)
-                .ok_or_else(|| VMStatus::Error(StatusCode::DATA_FORMAT_ERROR))?;
+                .ok_or_else(|| VMStatus::Error(StatusCode::UNKNOWN_INVARIANT_VIOLATION_ERROR))?;
```
vgao1996, line 470: The Move VM should never return a layout that doesn't match the value.

language/libra-vm/src/libra_vm.rs
```diff
@@ -459,15 +459,15 @@ pub fn txn_effects_to_writeset_and_events_cached<C: AccessPathCache>(
...
             let key = EventKey::try_from(guid.as_slice())
                 .map_err(|_| VMStatus::Error(StatusCode::EVENT_KEY_MISMATCH))?;
```
vgao1996, line 472: @dariorussi @tnowacki @tzakian It seems to me that it's technically possible for someone to emit an event from Move with a bad guid and trigger this and therefore I'm keeping it as a user error. Let me know if this sounds correct.

On the other hand, I think events in their current form are definitely a Libra concept and will need to be redesigned when we separate Move from Libra. I feel like instead of having
```
native fun write_to_event_store<T: copyable>(guid: vector<u8>, count: u64, msg: T);
```
we'll have
```
native fun write_to_event_store<T: Copyable>(msg: T);
```
In the case of Libra, we will just have another wrapper function that creates an emits a `struct LibraEvent<T: copyable> { guid: ..., seq_num: ..., msg: T }`.

However it's also possible this will simply no longer be a problem if we happen to find a way to solve the general issues with native functions that are closely associated with the VM runtime.

language/libra-vm/src/libra_vm.rs
dariorussi: out of curiosity, isn't the case that both of those generate a discarded transaction with an invariant violation?
I guess what I am asking is "would this change makes it easier or harder to understand what is going on?",
or maybe you are saying that this being an impossible condition it may be better to return a generic invariant error?

language/libra-vm/src/libra_vm.rs
dariorussi: same as above

language/libra-vm/src/libra_vm.rs
dariorussi: same as above

language/libra-vm/src/libra_vm.rs
dariorussi: more of a question for @sblackshear, @runtian-zhou and the others mentioned that for me :)

language/libra-vm/src/libra_vm.rs
vgao1996: None of `VALUE_SERIALIZATION_ERROR`, `VALUE_DESERIALIZATION_ERROR` and `DATA_FORMAT_ERROR` is in the invariant violation category and they will incorrectly label transaction outputs as Kept.
```
                match code.status_type() {
                    // Any unknown error should be discarded
                    StatusType::Unknown => Err(code),
                    // Any error that is a validation status (i.e. an error arising from the prologue)
                    // causes the transaction to not be included.
                    StatusType::Validation => Err(code),
                    // If the VM encountered an invalid internal state, we should discard the transaction.
                    StatusType::InvariantViolation => Err(code),
                    // A transaction that publishes code that cannot be verified will be charged.
                    StatusType::Verification => Ok(KeptVMStatus::MiscellaneousError),
                    // If we are able to decode the`SignedTransaction`, but failed to decode
                    // `SingedTransaction.raw_transaction.payload` (i.e., the transaction script),
                    // there should be a charge made to that user's account for the gas fees related
                    // to decoding, running the prologue etc.
                    StatusType::Deserialization => Ok(KeptVMStatus::MiscellaneousError),
                    // Any error encountered during the execution of the transaction will charge gas.
                    StatusType::Execution => Ok(KeptVMStatus::ExecutionFailure {
                        location: AbortLocation::Script,
                        function: 0,
                        code_offset: 0,
                    }),
                }
```

language/libra-vm/src/libra_vm.rs
tnowacki: > and therefore I'm keeping it as a user error. Let me know if this sounds correct.

`EVENT_KEY_MISMATCH` is an invariant violation. So if a user can hit this, this would be very bad. 
Can you write a test where this is hit?

vgao1996: Oh yeah, NVM. `fun write_to_event_store` is private and other Move code needs to use `fun emit_event`, which will always generate a valid guid if working properly. So in this case the infeasibility to trigger this error is guaranteed by the Move implementation of the Libra framework.

language/libra-vm/src/libra_vm.rs
dariorussi: got it, fantastic!

dariorussi (approved):
approving as I assume you are going to fix (if not fixed already) the issue raised by @tnowacki and we are good. 
As for adding a test for that case feel free to do in another PR

vgao1996:
@dariorussi If I'm understanding correctly, I mistaken `EVENT_KEY_MISMATCH ` as a user error and @tnowacki pointed that out to me so there's nothing to fix in this case. 

I'll consider adding a test, but just like every other invariant violation we have -- we don't have a good way to trigger them if the code is functioning as intended. Anyways I'll keep thinking about it though.

PR #6166, opened 2020-09-23
[breaking][vm][libra framework] Admin scripts no longer trigger reconfigurations by default

tzakian:
This PR changes the default behavior of admin scripts. Specifically, the execution of a `WriteSetPayload::Script` no longer triggers a reconfiguration by default, but reconfigurations can be triggered by an admin script if it triggers it manually, or calls code that triggers one.

The behavior of `WriteSetPayload::Direct` scripts remains unchanged. 

The motivation of this change is that administrative scripts triggering reconfigurations poses issues to AOS, and additionally makes the registration of new currencies on-chain more difficult.


zekun000 (approved):
lgtm, but I know nothing about vm

language/move-lang/functional-tests/tests/libra/libra/add_currency_to_system.move
```diff
@@ -0,0 +1,159 @@
+//! account: vivian, 1000000, 0, validator
```
tzakian, line 1: This is an example of adding a currency to the system using admin scripts, and making sure that it's fully operational. This wasn't possible previously since the writeset of the admin script would intersect with the write set of the `writeset_epilogue` at a `LibraConfig::NewEpochEvent`.

language/libra-vm/src/libra_transaction_executor.rs
```diff
@@ -319,12 +319,12 @@ impl LibraVM {
         remote_cache: &StateViewCache<'_>,
         writeset_payload: &WriteSetPayload,
         txn_sender: Option<AccountAddress>,
-    ) -> Result<ChangeSet, Result<(VMStatus, TransactionOutput), VMStatus>> {
+    ) -> Result<(ChangeSet, bool), Result<(VMStatus, TransactionOutput), VMStatus>> {
```
sblackshear, line 322: Nit: instead of returning an extra bool value, this code might be more readable if we add 
```needs_reconfig(&self): bool {
  match self {
    WriteSetPayload::Direct(_) => true,
    ... 
  }
}```
 to `impl WriteSetPayload` + use that at line 472 instead?

Comment (tzakian) [reply language/libra-vm/src/libra_transaction_executor.rs:322]: I like it! I'll update

File: language/e2e-testsuite/src/tests/admin_script.rs [hunk 3]
```diff
@@ -65,6 +65,71 @@ fn admin_script_rotate_key_single_signer() {
...
+
+        let compiler = Compiler {
+            address: account_config::CORE_CODE_ADDRESS,
+            extra_deps: vec![],
+            ..Compiler::default()
+        };
+        compiler
+            .into_script_blob("file_name", code)
+            .expect("Failed to compile")
+    };
```
davidiw, line 116: some alignment weirdness here?

language/e2e-testsuite/src/tests/admin_script.rs
```diff
@@ -65,6 +65,71 @@ fn admin_script_rotate_key_single_signer() {
...
+    main(lr_account: &signer, account: &signer, auth_key_prefix: vector<u8>) {
+      let rotate_cap: LibraAccount.KeyRotationCapability;
+      rotate_cap = LibraAccount.extract_key_rotation_capability(copy(account));
+      LibraAccount.rotate_authentication_key(&rotate_cap, move(auth_key_prefix));
+      LibraAccount.restore_key_rotation_capability(move(rotate_cap));
+
+      LibraConfig.reconfigure(move(lr_account));
+      return;
+    }
+";
```
davidiw, line 106: here's how I generated a multiline string: https://github.com/libra/libra/blob/master/config/src/config/secure_backend_config.rs#L206-L212

davidiw (commented):
awesome

language/e2e-testsuite/src/tests/admin_script.rs
tzakian: I don't think so since this is what `cargo x fmt` spits out? The alignment makes sense to me since this is closing the `let` on line 92.

moeziniaf (approved):

PR #6167, opened 2020-09-23
[logging] Add thread name to text log output

gregnazario:
Thread name was added, but it was missing from the text log output, so now it will exist.

### DIFF
--- a/common/logger/src/libra_logger.rs
+++ b/common/logger/src/libra_logger.rs
@@ -336,18 +336,23 @@ impl Writer for FileWriter {
 }
 
 /// Converts a record into a string representation:
-/// UNIX_TIMESTAMP LOG_LEVEL FILE:LINE MESSAGE
+/// UNIX_TIMESTAMP LOG_LEVEL [thread_name] FILE:LINE MESSAGE JSON_DATA
 /// Example:
-/// 2020-03-07 05:03:03 INFO common/libra-logger/src/lib.rs:261 Hello
+/// 2020-03-07 05:03:03 INFO [thread_name] common/libra-logger/src/lib.rs:261 Hello { "world": true }
 fn format(entry: &LogEntry) -> Result<String, fmt::Error> {
     use std::fmt::Write;
 
     let mut w = String::new();
+    write!(w, "{}", entry.timestamp)?;
+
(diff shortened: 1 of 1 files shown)


common/logger/src/libra_logger.rs
```diff
@@ -348,9 +348,13 @@ fn format(entry: &LogEntry) -> Result<String, fmt::Error> {
         "{} {} {}",
         entry.metadata.level(),
         entry.timestamp,
-        entry.metadata.location()
+        entry.metadata.location(),
     )?;
 
+    if let Some(thread_name) = &entry.thread_name {
+        write!(w, " [{}]", thread_name)?;
```
zekun000, line 355: can we move it to the beginning like INFO [thread-name]?

common/logger/src/libra_logger.rs
```diff
@@ -336,9 +336,9 @@ impl Writer for FileWriter {
 }
 
 /// Converts a record into a string representation:
-/// UNIX_TIMESTAMP LOG_LEVEL FILE:LINE MESSAGE
+/// UNIX_TIMESTAMP LOG_LEVEL FILE:LINE [thread_name] MESSAGE JSON_DATA
 /// Example:
-/// 2020-03-07 05:03:03 INFO common/libra-logger/src/lib.rs:261 Hello
+/// 2020-03-07 05:03:03 INFO common/libra-logger/src/lib.rs:261 [thread_name] Hello { "world": true }
```
zekun000, line 341: btw I think the order of timestamp and level is wrong

common/logger/src/libra_logger.rs
zekun000: just use &entry.thread_name.unwrap_or("") above

zekun000 (approved):
thanks! I have this in my massive pr https://github.com/libra/libra/pull/6145/files too

common/logger/src/libra_logger.rs
gregnazario: I've fixed it so the timestamp is at the beginning.

gregnazario:
> thanks! I have this in my massive pr https://github.com/libra/libra/pull/6145/files too

Yeah, it probably doesn't belong in that massive PR

PR #6170, opened 2020-09-23
[dkr] Install ca-certificates in client images

mgorven:
These are needed to actually verify the server cert.

Fix build on FB devservers with fwdproxy authentication.


davidiw (approved):
/land

PR #6172, opened 2020-09-23
[state sync] structured logging + some counters

christinacelee:
## Motivation

Convert to structured logging (+ some refactoring for better error catching), add some error counters that go with error logging. Will clean up counters in later PR with dashboards

### DIFF
--- a/state-synchronizer/src/chunk_request.rs
+++ b/state-synchronizer/src/chunk_request.rs
@@ -39,13 +39,24 @@ pub enum TargetType {
     Waypoint(Version),
 }
 
+impl TargetType {
+    pub fn version(&self) -> Option<u64> {
+        match self {
+            TargetType::TargetLedgerInfo(li) => Some(li.ledger_info().version()),
+            TargetType::HighestAvailable { target_li, .. } => {
+                target_li.as_ref().map(|li| li.ledger_info().version())
+            }
+            TargetType::Waypoint(version) => Some(*version),
+        }
+    }
+}
+
 impl fmt::Debug for TargetType {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
         write!(f, "{}", self)
     }
 }
 
-// TODO: This cuts out LedgerInfo, do we need it logging?
 impl fmt::Display for TargetType {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
         match self {
@@ -57,12 +68,11 @@ impl fmt::Display for TargetType {
                 timeout_ms,
             } => write!(
                 f,
-                "HighestAvailable(timeout:{}, target_li_version:{})",
+                "HighestAvailable(timeout:{}, target_li:{})",
                 timeout_ms,
-                target_li.as_ref().map_or_else(
-                    || String::from("None"),
-                    |li| li.ledger_info().version().to_string()
-                )
+                target_li
+                    .as_ref()
+                    .map_or_else(|| String::from("None"), |li| li.to_string())
             ),
             TargetType::Waypoint(version) => write!(f, "Waypoint({})", version),
         }
@@ -106,7 +116,7 @@ impl fmt::Display for GetChunkRequest {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
         write!(
             f,
-            "[ChunkRequest: known version: {}, epoch: {}, limit: {}, target: {:?}]",
+            "[ChunkRequest: known version: {}, epoch: {}, limit: {}, target: {}]",
             self.known_version,
             self.current_epoch,
             self.limit,
--- a/state-synchronizer/src/coordinator.rs
+++ b/state-synchronizer/src/coordinator.rs
@@ -6,6 +6,7 @@ use crate::{
     chunk_response::{GetChunkResponse, ResponseLedgerInfo},
     counters,
     executor_proxy::ExecutorProxyTrait,
+    logging::{LogEntry, LogEvent, LogSchema},
     network::{StateSynchronizerEvents, StateSynchronizerMsg, StateSynchronizerSender},
     request_manager::{PeerScoreUpdateType, RequestManager},
     SynchronizerState,
@@ -96,8 +97,10 @@ impl PendingLedgerInfos {
     /// Adds `new_li` to the queue of pending LI's
     fn add_li(&mut self, new_li: LedgerInfoWithSignatures) {
         if self.pending_li_queue.len() >= self.max_pending_li_limit {
-            debug!(
-                "[state sync] pending LI store reached max capacity, failed to add LI {}",
+            warn!(
+                LogSchema::new(LogEntry::ProcessChunkResponse),
+                "pending LI store reached max capacity {}, failed to add LI {}",
+                self.max_pending_li_limit,
                 new_li
             );
             return;
@@ -196,6 +199,7 @@ impl<T: ExecutorProxyTrait> SyncCoordinator<T> {
         executor_proxy: T,
         initial_state: SynchronizerState,
     ) -> Self {
+        info!(LogSchema::event_log(LogEntry::Waypoint, LogEvent::Initialize).waypoint(waypoint));
         let retry_timeout_val = match role {
             RoleType::FullNode => config.tick_interval_ms + config.long_poll_timeout_ms,
             RoleType::Validator => 2 * config.tick_interval_ms,
@@ -233,6 +237,7 @@ impl<T: ExecutorProxyTrait> SyncCoordinator<T> {
             StateSynchronizerEvents,
         )>,
     ) {
+        info!(LogSchema::new(LogEntry::RuntimeStart));
         let mut interval = interval(Duration::from_millis(self.config.tick_interval_ms)).fuse();
 
         let events: Vec<_> = network_handles
@@ -250,7 +255,8 @@ impl<T: ExecutorProxyTrait> SyncCoordinator<T> {
                                 .with_label_values(&[counters::SYNC_MSG_LABEL, counters::CONSENSUS_SENDER_LABEL])
                                 .start_timer();
                             if let Err(e) = self.request_sync(*request) {
-                                error!("[state sync] request sync fail: {}", e);
+                                error!(LogSchema::new(LogEntry::SyncRequest).error(&e));
+                                counters::PROCESS_SYNC_REQUEST_FAILURE.inc();
                             }
                         }
                         CoordinatorMessage::Commit(txns, events, callback) => {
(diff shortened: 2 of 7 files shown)


state-synchronizer/src/chunk_request.rs
```diff
@@ -39,13 +39,24 @@ pub enum TargetType {
...
+    }
+}
+
 impl fmt::Debug for TargetType {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
         write!(f, "{}", self)
     }
 }
 
-// TODO: This cuts out LedgerInfo, do we need it logging?
```
gregnazario, line 48: So, the reason I add this TODO, is `LedgerInfo` is a huge object last I checked.  I would think about whether we need the entire thing or not.

state-synchronizer/src/request_manager.rs
```diff
@@ -254,17 +259,29 @@ impl RequestManager {
...
             let peer_id = peer.peer_id();
             let send_result = sender.send_to(peer_id, msg.clone());
-            let result_label = if send_result.is_err() {
-                failed_peer_sends.push((peer.clone(), send_result));
+            let curr_log = log.clone().peer(peer.clone());
+            let result_label = if let Err(e) = send_result {
+                failed_peer_sends.push(peer.clone());
+                error!(curr_log
+                    .event(LogEvent::NetworkSendError)
+                    .error(&format_err!("{:?}", e)));
```
gregnazario, line 267: If it's a well known type you can use that, and set the schema to print the debug version.  Might be a little cleaner / easier

In Networking, I couldn't do this, because there were like 3-4 different error types.

state-synchronizer/src/logging.rs
```diff
@@ -0,0 +1,153 @@
...
+};
+use serde::Serialize;
+#[derive(Clone, Schema)]
+pub struct LogSchema<'a> {
+    name: LogEntry,
+    event: Option<LogEvent>,
+    #[schema(debug)]
+    error: Option<&'a Error>,
+    #[schema(display)]
+    peer: Option<PeerNetworkId>,
```
gregnazario, line 19: You should be able to do `Option<&'a PeerNetworkId>` then you don't have to have the custom clone function you have below.

state-synchronizer/src/request_manager.rs
```diff
@@ -232,9 +233,12 @@ impl RequestManager {
     }
 
     pub fn send_chunk_request(&mut self, req: GetChunkRequest) -> Result<()> {
+        let log = LogSchema::new(LogEntry::SendChunkRequest).chunk_req(&req);
+
         // update internal state
         let peers = self.pick_peers();
         if peers.is_empty() {
+            warn!(log.event(LogEvent::MissingPeers));
```
gregnazario, line 241: I like the way you use the schema with these enums.  It's very clean.

gregnazario (commented):
I still have to look over `state-synchronizer/src/coordinator.rs` it's a big one

state-synchronizer/src/chunk_request.rs
christinacelee: the Display format of ledger info with sigs omits the signatures, and I think that might not be that big? last I checked we wouldn't really need the signatures for debugging

state-synchronizer/src/coordinator.rs
```diff
@@ -280,20 +289,31 @@ impl<T: ExecutorProxyTrait> SyncCoordinator<T> {
...
-                                    debug!("[state sync] new peer {}", peer);
+                                    debug!(LogSchema::new(LogEntry::NewPeer).peer_id(&peer).is_upstream_peer(self.request_manager.is_upstream_peer(&peer, origin)));
                                     self.request_manager.enable_peer(peer, origin);
                                     self.check_progress();
                                 }
-                                Event::LostPeer(peer_id, _origin) => {
+                                Event::LostPeer(peer_id, origin) => {
                                     let peer = PeerNetworkId(network_id, peer_id);
-                                    debug!("[state sync] lost peer {}", peer);
+                                    debug!(LogSchema::new(LogEntry::LostPeer).peer_id(&peer).is_upstream_peer(self.request_manager.is_upstream_peer(&peer, origin)));
```
gregnazario: nice this `is_upstream_peer` will definitely help

state-synchronizer/src/logging.rs
```diff
@@ -0,0 +1,153 @@
...
+
+    pub fn peer_id(mut self, peer_id: &PeerNetworkId) -> Self {
+        self.peer = Some(peer_id.clone());
+        self
+    }
+}
+
+#[derive(Clone, Copy, Serialize)]
+#[serde(rename_all = "snake_case")]
+pub enum LogEntry {
```
zekun000, line 108: do you need this? there's already those metadata in logs 
```
            module_path!().split("::").next().unwrap(),
            module_path!(),
            file!(),
            line!(),
            concat!(file!(), ':', line!()),
```

christinacelee: my reasoning was that the LogEntry (and LogEvent) would help search for specific logs when we debug, can sort of serve like IDs for logs

state-synchronizer/src/coordinator.rs
```diff
@@ -394,39 +414,31 @@ impl<T: ExecutorProxyTrait> SyncCoordinator<T> {
     /// StateSynchronizer assumes that it's the only one modifying the storage (consensus is not
     /// trying to commit transactions concurrently).
     fn request_sync(&mut self, request: SyncRequest) -> Result<()> {
+        let local_li_version = self.local_state.highest_local_li.ledger_info().version();
+        let target_version = request.target.ledger_info().version();
+        debug!(
+            LogSchema::event_log(LogEntry::SyncRequest, LogEvent::Received)
+                .target_version(target_version)
+                .local_li_version(local_li_version)
+        );
```
gregnazario, line 423: I don't think this is really important, but I **think** that it shouldn't execute anything in the `debug!()` if debug is disabled.  Though, this reads so much easier.

gregnazario (approved):
There's a lot of changes here, I didn't see anything big standout particularly.  I suggest next time we break it down into a couple commits so we can see the changes like the refactors, where there's a couple functions flipped in space, or the split between the metric changes and the logging changes.

As far as the logging amount, I'll leave it up to you to decide if it logs too much, there were a lot of logs and I can't really keep track, but we should keep an eye on it once we deploy it / run it in cluster test to see if it's excessive.

state-synchronizer/src/coordinator.rs
christinacelee: hm i think this comment might be on the version before my force-push...?

state-synchronizer/src/chunk_request.rs
davidiw: I guess the question is what are we looking for? ledger info itself has a lot of info only relevant to consensus, perhaps not critical but we should probably make a to_client_string or something like that such that we only show version and hash maybe?

state-synchronizer/src/coordinator.rs
```diff
@@ -96,8 +97,10 @@ impl PendingLedgerInfos {
     /// Adds `new_li` to the queue of pending LI's
     fn add_li(&mut self, new_li: LedgerInfoWithSignatures) {
         if self.pending_li_queue.len() >= self.max_pending_li_limit {
-            debug!(
-                "[state sync] pending LI store reached max capacity, failed to add LI {}",
+            warn!(
+                LogSchema::new(LogEntry::ProcessChunkResponse),
+                "pending LI store reached max capacity {}, failed to add LI {}",
+                self.max_pending_li_limit,
```
davidiw, line 103: what does this mean? is this warning covered in your runbook? it isn't clear to me why this would happen or what if anything we as an operator can do ... should we expect to be patient or...?

state-synchronizer/src/coordinator.rs
```diff
@@ -233,6 +237,7 @@ impl<T: ExecutorProxyTrait> SyncCoordinator<T> {
             StateSynchronizerEvents,
         )>,
     ) {
+        info!(LogSchema::new(LogEntry::RuntimeStart));
```
davidiw, line 240: oh clever

state-synchronizer/src/coordinator.rs
```diff
@@ -280,20 +289,37 @@ impl<T: ExecutorProxyTrait> SyncCoordinator<T> {
                             match event {
                                 Event::NewPeer(peer_id, origin) => {
                                     let peer = PeerNetworkId(network_id, peer_id);
-                                    debug!("[state sync] new peer {}", peer);
+                                    debug!(LogSchema::new(LogEntry::NewPeer)
+                                        .peer(&peer)
+                                        .is_upstream_peer(self.request_manager.is_upstream_peer(&peer, origin))
+                                    );
                                     self.request_manager.enable_peer(peer, origin);
```
davidiw, line 296: it is interesting that you have the is_upstream_peer above but regardless you pass it to the request manager

state-synchronizer/src/counters.rs
```diff
@@ -26,6 +26,21 @@ pub const COMMITTED_VERSION_LABEL: &str = "committed"; // Version of latest ledg
 pub const SYNCED_VERSION_LABEL: &str = "synced"; // Version of most recent txn that was synced (even if it is not backed by an LI)
 pub const TARGET_VERSION_LABEL: &str = "target"; // Version a node is trying to catch up to
 
+// failed channel send type labels
```
davidiw, line 29: if you want to have a lot, lot of fun with type safety: https://github.com/libra/libra/blob/master/secure/net/src/lib.rs#L106-L133

state-synchronizer/src/logging.rs
```diff
@@ -0,0 +1,148 @@
...
+            count: None,
+            local_li_version: None,
+            local_synced_version: None,
+            local_epoch: None,
+            ledger_info: None,
+            new_epoch: None,
+            old_epoch: None,
+            target_version: None,
+        }
+    }
```
davidiw, line 88: recommend you have a new_event or something that takes in name: LogEntry and event: Option<LogEvent> and both of these call into that... this avoid having 2x20 lines of code and two places that need to be updated if you add another field.

state-synchronizer/src/logging.rs
davidiw: it is a mixed bag, the advantage of having this well defined is that even if you move around the log lines or rename the functions these fields are static and very descriptive of the intent!

state-synchronizer/src/request_manager.rs
davidiw: should fix that in networking or make a single unified type that they can all Into into

davidiw (commented):
beautiful! I don't know how quickly you want to move on this, but I'd love to see what this looks like in steady state given the experience that PFNs and VFNs currently having with catching up. if you're not in a hurry, we can run canary now, if you are, we can look at the logs after CT runs.

davidiw:
if you want to cancel the canary, feel free to

state-synchronizer/src/coordinator.rs
christinacelee: ah can probably move new/lost peer logging inside request_manager entirely!

PR #6174, opened 2020-09-23
[network] Make NetworkContext serialize consistently

gregnazario:
### Overview
There was a mismatch on serialization which caused issues in the underlying ES storage.

* Private was serializing like `{"network_id":{"private":"vfn"}}`
* Public was serializing like `{"network_id":"public"}`

Now all serialize to a string.  This does cause a problem if you name the private network `public` or `validator`.

### DIFF
--- a/config/src/network_id.rs
+++ b/config/src/network_id.rs
@@ -2,7 +2,7 @@
 // SPDX-License-Identifier: Apache-2.0
 use crate::config::RoleType;
 use libra_types::PeerId;
-use serde::{Deserialize, Serialize};
+use serde::{Deserialize, Serialize, Serializer};
 use std::fmt;
 
 /// A grouping of common information between all networking code for logging.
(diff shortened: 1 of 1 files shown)


config/src/network_id.rs
```diff
@@ -2,19 +2,43 @@
...
 /// for logging accordingly.
-#[derive(Clone, Eq, PartialEq, Serialize)]
+#[derive(Clone, Eq, PartialEq)]
 pub struct NetworkContext {
     network_id: NetworkId,
     role: RoleType,
     peer_id: PeerId,
 }
 
+/// This custom serializer is to make sure that there is a consistent output type of
```
davidiw, line 18: might be easier to use a serde field attribute serialize_with and just point to how you want the network_id to look. https://serde.rs/field-attrs.html

davidiw (approved):

config/src/network_id.rs
gregnazario: Yeah, that probably was easier :P

Issue #5615, opened 2020-08-13
[logging] Update key manager logs with leveled logs


gregnazario:
### Overview
Let's migrate send_struct_log! to use leveled macros, and remove `crit!` logs

### Tasks
- [ ] Remove any reference to `.warning()` or `.critical()`
- [ ] Change `send_struct_log!` to `sl_info!`, `sl_error!`, etc... accordingly similar to the text log macros
- [ ] replace any `crit!` with `error!`

gregnazario:
@JoshLind :) if you could take some time to appropriately fix these logs, that would be great!

I'm tracking all of these in the logging project, to ensure I know how things are working out.

JoshLind:
Sounds good, @gregnazario! I'll schedule some time to do them this week!

gregnazario:
Resolved and tracked in a different way.

PR #6175, opened 2020-09-23
[Key Manager] Update log levels in key manager and improve counter descriptions.

JoshLind:
## Motivation

This PR updates the logging in the key manager to use different log levels (e.g., for errors, warnings and general informative logs). It also updates some of the counter descriptions to be clearer. This PR should close out: https://github.com/libra/libra/issues/5615

Yes.

## Test Plan

Spawning the key manager locally shows the different log levels now being supported.

## Related PRs

None.

### DIFF
--- a/secure/key-manager/src/lib.rs
+++ b/secure/key-manager/src/lib.rs
@@ -147,13 +147,13 @@ where
                 Err(Error::LivenessError(last_value, current_value)) => {
                     // Log the liveness error and continue to execute.
                     let error = Error::LivenessError(last_value, current_value);
-                    info!(LogSchema::new(LogEntry::CheckKeyStatus)
+                    error!(LogSchema::new(LogEntry::CheckKeyStatus)
                         .event(LogEvent::Error)
                         .liveness_error(&error));
                 }
                 Err(e) => {
                     // Log the unexpected error and continue to execute.
-                    info!(LogSchema::new(LogEntry::CheckKeyStatus)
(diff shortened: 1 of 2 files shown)


secure/key-manager/src/lib.rs
```diff
@@ -147,13 +147,13 @@ where
                 Err(Error::LivenessError(last_value, current_value)) => {
                     // Log the liveness error and continue to execute.
                     let error = Error::LivenessError(last_value, current_value);
-                    info!(LogSchema::new(LogEntry::CheckKeyStatus)
+                    error!(LogSchema::new(LogEntry::CheckKeyStatus)
                         .event(LogEvent::Error)
                         .liveness_error(&error));
```
gregnazario, line 152: Could this be temporary / go away?  Mostly, just trying to determine `error!` vs `warn!`

gregnazario (approved):

secure/key-manager/src/lib.rs
JoshLind: Given the frequency at which the key manager checks the state of the keys and blockchain (which is minutes, at most), if this case is hit, it more than likely means consensus is down and something is really wrong. Or another way to look at it: I'd be really surprised if the timestamp on-chain hasn't increased at all in a space of minutes, so I think an error should be logged.

JoshLind:
Thanks @gregnazario! 😄 

/land

PR #6181, opened 2020-09-24
[libra framework] specify p2p and create child vasp scripts

emmazzz:
## Motivation

This PR adds specifications for 
- `peer_to_peer_with_metadata` and `create_child_vasp_account` transaction scripts
- underlying functions in LibraAccount including `pay_from`, `make_account`, `add_currencies_for_account` and `create_child_vasp_account`.

Yes.


language/stdlib/modules/LibraAccount.move
```diff
@@ -238,11 +238,11 @@ module LibraAccount {
         if (is_withdrawal) {
             has_published_account_limits<Token>(payer) &&
             VASP::is_vasp(payer) &&
-            (!VASP::is_vasp(payee) || !VASP::is_same_vasp(payer, payee))
+            !VASP::is_same_vasp(payer, payee)
```
emmazzz, line 241: deleted `!VASP::is_vasp(payee)` because `!VASP::is_vasp(payee)` implies `!VASP::is_same_vasp(payer, payee)`.

DavidLDill (approved):

tzakian:
@matbd It looks like the transaction-builder test is failing because the `EXPECTED_OUTPUT` differs for Rust compilation, but I can't seem to find any reason why that would be the case. Any ideas?

thefallentree:
1. you need to manually run language/transcation-builder cli to refresh the generated code and check it in , that will fix the unit test
2. since this is breaking change, you need to add [breaking] prefix in commit title, to get past the compat check

emmazzz:
> 1. you need to manually run language/transaction-builder cli to refresh the generated code and check it in , that will fix the unit test
> 2. since this is breaking change, you need to add [breaking] prefix in commit title, to get past the compat check

Thanks @thefallentree ! And how do I run the language/transaction-builder cli? I skimmed at the README file in that directory and tried running the commands in it but it didn't seem to regenerating the code.

PR #6182, opened 2020-09-24
[move-cli] support scripts in move_src

sblackshear:
Previously, `move compile` got upset if there was >1 script inside `move_src`. An arbitrary number of scripts can now be placed anywhere inside `move_src`.


tnowacki (commented):
On a PR I'm working on, move_src is gone due to the addition of interface files, let's see how that shakes out before moving on this (if that works for you, happy to land this and just kill it :P)

sblackshear:
There are some other improvements in (e.g.) printing here, so I think I'll land it, but feel free to kill whatever we don't need anymore after relaxing the `move_src` req!

tzakian (approved):
LGTM. Let's get this in and then we can kill `move_src` later on.

tnowacki:
> On a PR I'm working on, move_src is gone due to the addition of interface files, let's see how that shakes out before moving on this (if that works for you, happy to land this and just kill it :P)

:(

PR #6185, opened 2020-09-24
[docgen] Support for overview documentation.

wrwg:
This PR adds support for overview documentation to the documentation generator.

An overview is provided via a template file in markdown format. This markdown is subject to all the processing we do for Move documentation comments (automatic referencing, highlighting, and section organization). The template can also contain placeholders for generated documentation content:

```
> {{move-include NAME_OF_MODULE_OR_SCRIPT}}
> {{move-toc}}
> {{move-index}}
```

See a strawmann [overview template](https://github.com/wrwg/libra/blob/script_doc/language/stdlib/transaction_scripts/overview_template.md) for transaction scripts ([rendered form](https://github.com/wrwg/libra/blob/script_doc/language/stdlib/transaction_scripts/doc/overview.md)) for more details.

The implementation also works for overviews for modules in the Libra framework, and a strawman template has been added there as well. Even without further documentation, the automatically genearted index of all scripts/modules is useful.

Technical details:
- New default for documentation generation is to place specs together with implementations instead of a separate section. This is for now a flag-controlled experiment. Bugs needed to be fixed to make this unused feature work.
- Boilerplate for script documentation has been removed, exploiting that they can contain only one function.
- Removed the restriction that the spec-lang enviroment can contain only one script. This makes generation of script documentation more efficient because they can be all done in one run.

## Motivation

Better documentation.

Yes

## Test Plan

Adapted existing tests.

## Related PRs

NA


language/move-prover/docgen/tests/testsuite.rs
```diff
@@ -24,11 +24,35 @@ const FLAGS: &[&str] = &[
...
+    if !is_root_template {
+        args.push(path.to_string_lossy().to_string());
+    } else {
+        // add as sources all .notest_move files starting with the name of the template.
+        let base_name = path.file_stem().unwrap().to_string_lossy().to_string();
+        let dir_name = path.parent().unwrap();
+        dir_name.read_dir().unwrap().for_each(|res| {
+            let f = res.unwrap();
+            let name = f.file_name().to_string_lossy().to_string();
+            if name.starts_with(&base_name) && name.ends_with(".notest_move") {
```
sblackshear, line 43: I think this can be simplified to: `if f.starts_with(path.with_extension("notest_move"))` (which should let you kill 38 and 42 also).

language/move-prover/spec-lang/src/ast.rs
```diff
@@ -504,6 +505,13 @@ impl ModuleName {
     pub fn name(&self) -> Symbol {
         self.1
     }
+
+    pub fn is_script(&self) -> bool {
```
sblackshear, line 509: Is this magic `ff...f` address also used elsewhere? A comment might help explain what's going on here.

sblackshear (approved):
This is awesome!

tzakian (approved):
This is amazing! Thank you so much for doing this. 

Only question I have is whether or not we want included documents to be "expandable" instead of inlined? I can see arguments for both (and am fine with either!), but am interested in your and @sblackshear 's thoughts on this.

language/move-prover/spec-lang/src/ast.rs
tzakian: I found this also a bit confusing so a comment would definitely be helpful. Also, would it be better to possibly use `0x0` instead here possibly? That's the VM reserved address so no account can be created there and ostensibly because of this no module could be published at that address.

sblackshear:
> This is amazing! Thank you so much for doing this.
> 
> Only question I have is whether or not we want included documents to be "expandable" instead of inlined? I can see arguments for both (and am fine with either!), but am interested in your and @sblackshear 's thoughts on this.

Good question. It seems nice to support both (e.g., `{{doc}}` could mean expandable and `{{{doc}}}` could mean inlined).

language/move-prover/docgen/tests/testsuite.rs
wrwg: This can match `<basename>*.notest_move`, so I would need a regular expression. Seemed to be overkill so I think I stay with this more brute force approach.

language/move-prover/spec-lang/src/ast.rs
wrwg: The magic ff...f comes from the lang framework, I believe the into_module function on scripts creates this. I added a comment to clarify this.

wrwg:
> > This is amazing! Thank you so much for doing this.
> > Only question I have is whether or not we want included documents to be "expandable" instead of inlined? I can see arguments for both (and am fine with either!), but am interested in your and @sblackshear 's thoughts on this.
> 
> Good question. It seems nice to support both (e.g., `{{doc}}` could mean expandable and `{{{doc}}}` could mean inlined).

One problem is that the collapsing doesn't seem to work on Github for navigation. So if you have something like a section within the collapsed part, and click on it in TOC, nothing seem to happen. We can investigate this further as a followup. Another option would be to include only a summary and the link for the full doc in a standalone file.

tzakian:
> One problem is that the collapsing doesn't seem to work on Github for navigation. So if you have something like a section within the collapsed part, and click on it in TOC, nothing seem to happen. We can investigate this further as a followup. Another option would be to include only a summary and the link for the full doc in a standalone file.

Sounds good to me and also a reasonable alternative for now. Happy to get this in as-is, and we can circle back later to see if there are any alternatives if we want to.

PR #6186, opened 2020-09-24
[language] add move-vm-test-utils

vgao1996:
This introduces a new crate `move-vm-test-utils`, offering a few fundamental components that can be used to test the Move VM:
  - `lib move-vm-test-utils`
  - `mod effects`
    - `ChangeSet`: a representation of a collection of effects produced by the Move VM.
  - `mod storage`: mock storages implementing `RemoteCache`.
    - `BlankStorage`: an empty storage.
    - `DeltaStorage`: a storage adapter that stacks a `ChangeSet` on top of an existing storage without actually modifying it.
    - `InMemoryStorage`: general purpose in-memory storage that allows resources and modules to be published and unpublished.

This library may look very similar to the existing `e2e-tests`, especially when you compare `InMemoryStorage` and `FakeDataStore`. However there is a fundamental difference: `e2e-tests` is a testing framework crafted for the Libra system, while `move-vm-test-utils` operates solely on the Move level. This separation is an important step towards a more independent and complete testing story for the Move VM and other Move tools/infrastructures. 

Note there's nothing using this library right now, but I have a bunch of new tests relying on it coming soon. (#6187 being the first one in the row)

The majority of this PR is salvaged from #6055, #5972 and #5969 as we've decided to not pursue the more aggressive changes to the critical paths.

PR #6187, opened 2020-09-24
[language] test Move VM with corrupted storage 

vgao1996:
This tests the VM's handling of corrupted storage backends. Also moves the existing loader tests into this the newly created `move-vm-integration-test` crate as they share a lot of the infrastructure.

**Note: depending on #6186.**


language/move-vm/integration-tests/src/tests/loader_tests.rs
```diff
@@ -129,82 +128,178 @@ impl Adapter {
     }
 }
 
-#[derive(Clone, Debug)]
-struct DataStore {
-    modules: HashMap<ModuleId, Vec<u8>>,
-}
+const MODULES: &str = r#"
```
dariorussi, line 131: I had another test (when I was trying to commit the PR that did not load dependent modules in a `publish_module`) that was using the same set of modules and I think that is a useful way to do things, rather than having only a string to the API

dariorussi (commented):
as implied in one of the comment I find the usage of an external `.move` file (or multiple) very helpful and to me the usage of strings only in the API for compilation seems restrictive. I think it would be helpful to have both compiling via string and via file, probably even multiple files

language/move-vm/integration-tests/src/tests/bad_storage_tests.rs
```diff
@@ -0,0 +1,159 @@
...
+        let mut sess = vm.new_session(&storage);
+        sess.execute_script(
+            script_blob.clone(),
+            vec![],
+            vec![],
+            vec![TEST_ADDR],
+            &mut cost_strategy,
+        )
+        .unwrap();
+    }
```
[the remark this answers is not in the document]
vgao1996: Yes I want to ensure the original code does work as intended and the only factor triggering the invariant violation is the change to the storage. We should be fine without this, but still I feel more comfortable having this extra check.

language/move-vm/integration-tests/src/tests/loader_tests.rs
vgao1996: Yeah certainly no harm to have another option. I'll add that back and revert the changes to the loader tests.

language/move-vm/integration-tests/src/tests/bad_storage_tests.rs
```diff
@@ -0,0 +1,551 @@
...
+            &module_id,
+            &fun_name,
+            vec![],
+            vec![],
+            TEST_ADDR,
+            &mut cost_strategy,
+            |x| x,
+        )
+        .unwrap();
+    }
```
vgao1996, line 358: There's quite a bit of copy-n-pasting going on here and I'm aware of it... For now I'd prefer not trying to make them look pretty/optimized as they'll likely be rewritten in a much more concise way once the new YAML based functional tests come.

language/move-vm/integration-tests/src/tests/bad_storage_tests.rs
```diff
@@ -0,0 +1,159 @@
...
+
+            fun main(s: &signer) {
+                M::check(s);
+            }
+        }
+    "#;
+    let code = code.replace("{{ADDR}}", &format!("0x{}", TEST_ADDR.to_string()));
+    let mut units = compile_units(TEST_ADDR, &code).unwrap();
+
+    let s2 = as_script(units.pop().unwrap());
```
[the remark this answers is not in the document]
vgao1996: As we already about this offline I tried using expect  instead of unwrap and it ended up being a lot more verbose and didn't help clarify the intention here. I added comments explaining what each step is doing instead.

language/move-vm/integration-tests/src/tests/bad_storage_tests.rs
dariorussi: sounds good and no problem with it

dariorussi (approved):
this is great! and a good start, I think we have more cases to add in terms of bogus modules and we can talk about them and also all contribute to them.
For now this is great and let's get it in

Issue #1325, opened 2019-10-14
[consensus] byzantine OOM attacks on block tree


zekun000:
Attack: Today we accept arbitrary proposals for the current round as long as it's from primary proposer, malicious proposers could send as many proposals as they want and eventually make honest nodes out of memory.

Solution: We should only keep the block voted for and reject any proposals for the same round, but we're also willing to replace it with another one which has the QuorumCert. Note only one block at the same round could have a QuorumCert.

bmaurer:
If a proposer proposes a block in round R, they have to present a proof of round R+1 being over, right? This seems to suggest an even stronger property -- that just based on the incoming RPC (without making a sync / etc) once we see a proposal we can safely discard proposals from previous rounds, as long as we haven't voted on them.

It might also be interesting to think about situations in which there is a long growing block tree. One thought on how byzantine validators might trick a victim into seeing an ever growing tree: imagine a victim V is isolated from the network (so to the rest of the network they are offline). The adversary controls f-1 nodes. Whenever the adversary proposes a block on the real network in round R, and it also controls the R+1 proposer, it doesn't make a proposal, causing the round to timeout. Thus the attacker gets a round certificate on round R. The attacker proposes blocks in round R+1 to the victim. This seems like it'd create an infinitely growing chain.

Admittedly this is an extreme case but it's interesting to think about how bad resource exhaustion could get. I wonder if some simple heuristics could help. For example:

1) If you get a proposal B and B.round > B.qc.last_committed_round + 100, only accept B if it proposes no transactions (since proposers have to follow this it would be a protocol constant)
2) Have a config value max_round_delta_to_vote that refuses to vote for a block if B.round > B.qc.round + max_round_delta_to_vote

This would mean that your worst case memory/disk usage would be 100*max_block_size + max_round_delta_to_vote*empty_block_size.

zekun000:
@bmaurer I don't get how to get an ever-growing chain, you can have an ever growing **tree** if you control **all** the proposers because of no qc formed.

my gut feel is any attempt to limit the block tree growth would kill the theoretic liveness.

1. doesn't help the block growth, empty blocks are still blocks.
2. will kill liveness if we timeout 100 rounds.

bmaurer:
Sorry -- you're right it'd be a tree not a chain. But as a validator you'd have to keep every block you've voted on (i.e. the leafs of the trees). You can't discard blocks until you commit at a higher round.

> doesn't help the block growth, empty blocks are still blocks.

[...]

PR #6194, opened 2020-09-24
[cosensus] prevent multiple proposals from the same author in the same round

zekun000:
## Motivation

Address https://github.com/libra/libra/issues/1325

Y

## Test Plan

Added test.

## Related PRs

### DIFF
--- a/consensus/src/liveness/leader_reputation.rs
+++ b/consensus/src/liveness/leader_reputation.rs
@@ -2,12 +2,16 @@
 // SPDX-License-Identifier: Apache-2.0
 
 use crate::liveness::proposer_election::{next, ProposerElection};
-use consensus_types::common::{Author, Round};
+use consensus_types::{
+    block::Block,
+    common::{Author, Round},
+};
+use libra_crypto::HashValue;
 use libra_logger::prelude::*;
 use libra_types::block_metadata::{new_block_event_key, NewBlockEvent};
 use std::{
     cmp::Ordering,
-    collections::HashSet,
+    collections::{HashMap, HashSet},
     sync::{Arc, Mutex},
 };
 use storage_interface::{DbReader, Order};
@@ -130,6 +134,7 @@ pub struct LeaderReputation {
     proposers: Vec<Author>,
     backend: Box<dyn MetadataBackend>,
     heuristic: Box<dyn ReputationHeuristic>,
+    already_proposed: Mutex<(Round, HashMap<Author, HashValue>)>,
 }
 
 impl LeaderReputation {
@@ -142,6 +147,7 @@ impl LeaderReputation {
             proposers,
             backend,
             heuristic,
+            already_proposed: Mutex::new((0, HashMap::new())),
         }
     }
 }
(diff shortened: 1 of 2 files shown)


consensus/src/liveness/leader_reputation.rs
```diff
@@ -171,4 +176,38 @@ impl ProposerElection for LeaderReputation {
...
+                        false
+                    } else {
+                        already_proposed.1.insert(author);
+                        true
+                    }
+                }
+                Ordering::Less => false,
+            }
+        })
+    }
```
davidiw, line 212: why would we ever want more than one round? perhaps we just keep track of the current round and reset it if we see a more recent round and ignore earlier ones?
also do we want to keep track of the block_id?

consensus/src/liveness/leader_reputation_test.rs
```diff
@@ -141,4 +141,6 @@ fn test_api() {
         &signers[unexpected_index],
     );
     assert!(!proposer_election.is_valid_proposal(&bad_proposal));
+    // Second attempt should fail
+    assert!(!proposer_election.is_valid_proposal(&good_proposal));
```
davidiw, line 145: why would sending out a good proposal twice fail? seems a little interesting and not idempotent

consensus/src/liveness/leader_reputation.rs
aching: What if a proposer proposes a different block in the same round for instance?  Given that a proposer should only send the exact sample proposal in the same round, there is no point in accepting any other proposal from the correct round proposer if you already got one from them.

consensus/src/liveness/leader_reputation_test.rs
aching: Same as above^^^

consensus/src/liveness/leader_reputation.rs
davidiw: I think we said the same thing.

aching: I think I misunderstood your first sentence.  If you are saying "more than one proposal from a valid proposer in a round", then yes, I agree.

aching: My comment probably applies more towards the question about "why would sending out a good proposal twice fail? seems a little interesting and not idempotent"

zekun000: it's implementation details, but I mean we never send the proposal more than once, I can change it to track the block id as well

zekun000: > why would we ever want more than one round? perhaps we just keep track of the current round and reset it if we see a more recent round and ignore earlier ones?

@davidiw are you trolling? I thought that's exactly what I did lol

zekun000: ah I misunderstood it too 😄

zekun000: but this assumes the valid proposer for a given round won't change which is not necessarily true since it depends on committed round.

if the node is at round 10 but committed round is at 3, the leader maybe X, but later it synchronized the last committed round to 5, the leader maybe Y even it may not be able to vote for Y.

consensus/src/liveness/leader_reputation.rs
```diff
@@ -130,6 +133,7 @@ pub struct LeaderReputation {
     proposers: Vec<Author>,
     backend: Box<dyn MetadataBackend>,
     heuristic: Box<dyn ReputationHeuristic>,
+    already_proposed: Mutex<(Round, HashSet<Author>)>,
```
davidiw, line 136: Humm, I saw HashMap here or maybe I'm losing my mind :trollface: all good

davidiw: So let's make the entry in HashSet<(Author, HashValue (block id))>

davidiw (approved):

PR #6198, opened 2020-09-24
simplify documentation to point folks to the single node testnet

davidiw:
We probably want to also update the doc to discuss the docker setup. There's probably also room to bring back or even keep swarm or move it to an advance uses. I dunno. This is somewhat of an RFC but I would also like to avoid the amount of questions we get about swarm.


ericnakagawa (approved):
LGTM

bmwill (approved):
This is great! Most folks only want a single node anyway. We can take another look in the future if we want to better support multi-node local networks later.

/land

PR #6199, opened 2020-09-24
[consensus] better logging

zekun000:
## Motivation

add more context to the logs leveraging structure logs

Y

## Test Plan

it compiles.

## Related PRs

### DIFF
--- a/consensus/src/block_storage/sync_manager.rs
+++ b/consensus/src/block_storage/sync_manager.rs
@@ -3,6 +3,7 @@
 
 use crate::{
     block_storage::{BlockReader, BlockStore},
+    logging::{LogEvent, LogSchema},
     network::NetworkSender,
     network_interface::ConsensusMsg,
     persistent_liveness_storage::{PersistentLivenessStorage, RecoveryData},
@@ -21,7 +22,6 @@ use libra_types::{account_address::AccountAddress, epoch_change::EpochChangeProo
 use mirai_annotations::checked_precondition;
 use rand::{prelude::*, Rng};
 use std::{clone::Clone, sync::Arc, time::Duration};
-use termion::color::*;
 
 #[derive(Debug, PartialEq)]
 /// Whether we need to do block retrieval if we want to insert a Quorum Cert.
@@ -171,7 +171,11 @@ impl BlockStore {
         )
         .await?
         .take();
-        debug!("{}Sync to{} {}", Fg(Blue), Fg(Reset), root.0);
+        debug!(
+            LogSchema::new(LogEvent::CommitViaSync).round(self.root().round()),
+            committed_round = root.0.round(),
+            block_id = root.0.id(),
+        );
         self.rebuild(root, root_metadata, blocks, quorum_certs)
             .await;
 
@@ -194,8 +198,8 @@ impl BlockStore {
         state_computer: Arc<dyn StateComputer>,
     ) -> anyhow::Result<RecoveryData> {
         debug!(
-            "Start state sync with peer: {}, to block: {}",
-            retriever.preferred_peer.short_str(),
+            LogSchema::new(LogEvent::StateSync).remote_peer(retriever.preferred_peer),
+            "Start state sync with peer to block: {}",
             highest_commit_cert.commit_info(),
         );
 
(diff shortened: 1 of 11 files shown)


consensus/src/state_computer.rs
```diff
@@ -85,7 +85,7 @@ impl StateComputer for ExecutionProxy {
                 .commit(committed_txns, reconfig_events)
                 .await
         ) {
-            error!("failed to notify state synchronizer: {:?}", e);
+            error!(error = "Failed to notify state synchronizer", "{:?}", e);
```
gregnazario, line 88: Isn't `e` the error you want?

consensus/src/round_manager.rs
```diff
@@ -607,9 +643,17 @@ impl RoundManager {
     /// 1) fetch missing dependencies if required, and then
     /// 2) call process_certificates(), which will start a new round in return.
     async fn process_vote(&mut self, vote: &Vote) -> anyhow::Result<()> {
+        let round = vote.vote_data().proposed().round();
+        debug!(
+            round = round,
+            remote_peer = vote.author(),
+            event = LogEvent::ReceiveVote,
```
bmwill, line 650: Did you consider having a defined schema and using the generated setters to ensure that these fields all have the same type whenever they're logged?

consensus/src/state_computer.rs
zekun000: what I understand is the `error` field would be indexed but the message won't. so I'd like to query the same type of error but the actual error message may vary

consensus/src/round_manager.rs
zekun000: Does that mean I have to have one Schema per event type? or I'll have a giant schema with all possible fields optional?

bmwill: In order to improve maintainability I've been suggesting folks have 1 single schema that has all fields optional. Note I'm not suggesting you actually make a change just pointing out that there's other options

consensus/src/state_computer.rs
bmwill: There's a lot of extra data that is indexed with each log line. E.g. Log level, file name, line number, as well as the message itself. So what Greg is suggesting you do is:

```
error!(error = e, "Failed to notify state synchronizer");
```

Will result in structured data (that is indexed in Kibana) like:

```
{
  level: Error,
  file: "consensus/src/state_computer.rs",
  line: 88,
  message: "Failed to notify state synchronizer",
  data: {
    error: <Error formatted either by Serde/Debug/Display>,
  }
}
```

So everything is still indexable

zekun000: So I'm still able to filter by message = "Failed to notify state synchronizer"?

gregnazario: You will, it'll to a text search, it just won't be quite as easy to filter, and won't be able to be used to be searched like that in visualizations.

gregnazario: The main purpose of adding the extra fields is for indexing addition information, like identifiers, or in this case errors.  So, now I could have a visualization that says, show me a pie chart of all errors when `Failed to notify state synchronizer` and now I can see the percentage of errors by content

bmwill: You'd even be able to filter off the location, 'file:line-number' (which imo sounds like you're actually trying to do?)

consensus/src/block_storage/sync_manager.rs
```diff
@@ -171,7 +171,11 @@ impl BlockStore {
         )
         .await?
         .take();
-        debug!("{}Sync to{} {}", Fg(Blue), Fg(Reset), root.0);
+        debug!(
+            LogSchema::new(LogEvent::CommitViaSync).round(self.root().round()),
+            committed_round = root.0.round(),
+            block_id = root.0.id(),
```
bmwill, line 177: It looks like you have `committed_round` and `block_id` in a couple different log lines, maybe it would be useful to push these up into the LogSchema type?

bmwill (approved):
Looks good to me. I would personally try to push as many fields and things into the LogSchema you defined but that's not a reason to have another round of review.

PR #6206, opened 2020-09-24
[gha][lbt] add PR num to error msg

rustielin:
Makes it easier to see whether flaky land-blocking CT is from the same PR or not.


andll (approved):
Thank you!

PR #6208, opened 2020-09-25
[executor] add latency counter for syncing

lightmark:
We don't need a latency metric for vm in sync but for the whole function call.


msmouse (approved):

msmouse:
doesn't seem to harm in anyway if we keep the other one though..

lightmark:
@bors-libra land
The other one seems not that useful.

PR #6209, opened 2020-09-25
[libra-node] remove the disable logging flag

davidiw:
You can disable logging via config and env, why have yet another means
to do it?

### DIFF
--- a/testsuite/libra-swarm/src/swarm.rs
+++ b/testsuite/libra-swarm/src/swarm.rs
@@ -58,8 +58,6 @@ impl LibraNode {
         role: RoleType,
         config_path: &Path,
         log_path: PathBuf,
-        struct_log_path: PathBuf,
-        disable_logging: bool,
     ) -> Result<Self> {
         let config = NodeConfig::load(&config_path)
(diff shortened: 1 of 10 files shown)


testsuite/libra-swarm/src/swarm.rs
```diff
@@ -74,14 +73,10 @@ impl LibraNode {
...
-        if disable_logging {
-            node_command.arg("-d");
-        } else {
-            node_command.env("STRUCT_LOG_FILE", struct_log_path);
-        }
         node_command
             .stdout(log_file.try_clone()?)
-            .stderr(log_file.try_clone()?);
+            .stderr(log_file.try_clone()?)
+            .env("STRUCT_LOG_FILE", struct_log_path);
```
bmwill, line 79: We don't even respect this env variable anymore...I think?

bmwill (approved):

testsuite/libra-swarm/src/swarm.rs
davidiw: more to delete!

bmwill:
Looks like there's just a single lint that needs to be taken care of

PR #6210, opened 2020-09-25
[metrics] add a common metrics crate

davidiw:


davidiw:
@andll let's see what happens with this...

libra-action:
```
target packages:
  A libra-metrics-core 0.1.0 (workspace, path 'common/metrics-core')
    * features: [none]
```

libra-action:
```
target packages:
  A libra-metrics-core 0.1.0 (workspace, path 'common/metrics-core')
    * features: [none]
```

libra-action:
```
target packages:
  A libra-metrics-core 0.1.0 (workspace, path 'common/metrics-core')
    * features: [none]
```

davidiw:
works: http://grafana.ct-1-k8s-testnet.aws.hlw3truzy4ls.com/explore?orgId=1&left=%5B%22now-1h%22,%22now%22,%22Prometheus%22,%7B%22expr%22:%22libra_secure_net_events%7Bservice%3D%5C%22safety-rules%5C%22%7D%22%7D,%7B%22mode%22:%22Metrics%22%7D,%7B%22ui%22:%5Btrue,true,true,%22none%22%5D%7D%5D

libra-action:
```
target packages:
  A libra-metrics-core 0.1.0 (workspace, path 'common/metrics-core')
    * features: [none]
```

libra-action:
```
target packages:
  A libra-metrics-core 0.1.0 (workspace, path 'common/metrics-core')
    * features: [none]
```

libra-action:
```
target packages:
  A libra-metrics-core 0.1.0 (workspace, path 'common/metrics-core')
    * features: [none]
```

common/metrics-core/src/lib.rs
```diff
@@ -0,0 +1,9 @@
+// Copyright (c) The Libra Core Contributors
+// SPDX-License-Identifier: Apache-2.0
+
+// Re-export counter types from prometheus crate
+pub use prometheus::{
+    gather, register_histogram, register_histogram_vec, register_int_counter,
+    register_int_counter_vec, register_int_gauge, register_int_gauge_vec, Encoder, Histogram,
+    HistogramVec, IntCounter, IntCounterVec, IntGauge, IntGaugeVec, TextEncoder,
+};
```
bmwill, line 9: If this is all the crate does, then why don't we just have folks use prometheus directly instead of adding an extra crate which just exports prometheus types?

davidiw: Look at the links... Without this, it appears we're losing metrics.

bmwill: Something else must be going on, because this is literally just re-exporting types from the prometheus crate instead of using prometheus directly. I'm not saying we can't do this, just saying that doing this shouldn't "fix" lost metrics

metajack: Some context: I originally had reexported prometheus because all the calling code imported both libra-metrics and prometheus. So I reexported the needed prometheus types from libra-metrics to make it easier to use and remove the need for all callers to declare both dependencies directly.

davidiw: I agree, there is definitely a quirk here. At the same time, this definitely fixes the quirk albeit maybe not in the sexiest of fashions. I'm willing to make our metrics repo depend on push-metrics, is that agreeable?

bmwill (approved):

PR #6212, opened 2020-09-25
[xbuild] cleaner install and initial pass at xbuild wiring

rexhoffman:
## Motivation

With this pr:
1) the cargo installer now handles a full type representing a CargoInstallation, allowing installation from a githash on a git repo.
2) x.toml is configured to install a githash from rexhoffman/sccache, allowing anonymous read from v4 api aws s3 public buckets.
3) all x commands now have a configured sccache environment unless request not to, "SKIP_SCCACHE=true" will disable sccache.
4) SCCACHE_AWS_ACCESS_KEY_ID & SCCACHE_AWS_SECRET_ACCESS_KEY being present at execution time will cause the sccache to upload compilation results to the s3 bucket configured in the x.toml. (ci-artifacts.libra.org/sccache/libra)
5) Cuts fresh builds in about half.
6) Will not attempt to run sccache unless CARGO_HOME is set and equals /opt/cargo, and the git repo live in /opt/git/libra (due to sccache limitations).

Yes

## Test Plan

A lot more work... and aws.

## Related PRs

None


.cargo/config
```diff
@@ -1,3 +1,7 @@
+#dont know if I want a build section or a ENV yet....
+#[build]
+#rustc-wrapper = "sccache"
+
```
bmwill, line 4: We probably would want this in the x config not in cargo's, that way we aren't breaking folks running bare cargo commands.

devtools/x/src/build.rs
```diff
@@ -0,0 +1,155 @@
...
+    #[structopt(long)]
+    /// Build all tests
+    tests: bool,
+    #[structopt(long, number_of_values = 1)]
+    /// Build only the specified bench target
+    bench: Vec<String>,
+    #[structopt(long)]
+    ///  Build all benches
+    benchs: bool,
+    #[structopt(long = "all-targets")]
```
bmwill, line 67: You probably can't drop the "all-targets" part and just leave it saying `#[structopt(long)]` this is because the default is kebab-case

devtools/x/src/build.rs
```diff
@@ -0,0 +1,155 @@
...
+    #[structopt(long = "build-plan")]
+    /// Output the build plan in JSON (unstable)
+    build_plan: bool,
+    //TODO: support -vv?
+    #[structopt(long, short)]
+    ///  Use verbose output (-vv very verbose/build.rs output)
+    verbose: bool,
+    #[structopt(long, possible_values = &Coloring::variants(), default_value="Auto")]
+    ///  Coloring: auto, always, never
+    coloring: Coloring,
```
bmwill, line 109: We probably don't need to support 100% of these things from the start, but this looks like a complete list

devtools/x/src/build.rs
```diff
@@ -0,0 +1,155 @@
...
+    ///  Run without accessing the network
+    offline: bool,
+    #[structopt(name = "ARGS", parse(from_os_str), last = true)]
+    args: Vec<OsString>,
+}
+
+pub fn run(args: Args, xctx: XContext) -> Result<()> {
+    xctx.installer().install_if_needed("sccache");
+
+    let env_vars: &[(&str, &str)] = { &[("RUSTC_WRAPPER", "sccache")] };
```
bmwill, line 126: I don't think you need the braces here.

devtools/x/src/installer.rs
```diff
@@ -51,3 +94,19 @@ pub fn check_installed(name: &str, version: &str) -> bool {
     );
     found
 }
+
+fn install_all(tools: &[(String, String)]) -> bool {
+    let mut success: bool = true;
+    for (key, value) in tools {
+        success = success && install_if_needed(key, value);
```
bmwill, line 101: This should also work.

```suggestion
        success &= install_if_needed(key, value);
```

devtools/x/src/installer.rs
```diff
@@ -51,3 +94,19 @@ pub fn check_installed(name: &str, version: &str) -> bool {
...
+    }
+    success
+}
+
+fn check_all(tools: &[(String, String)]) -> bool {
+    let mut success: bool = true;
+    for (key, value) in tools {
+        success = success && check_installed(key, value);
```
bmwill, line 109: Same here

devtools/x/src/main.rs
```diff
@@ -3,13 +3,17 @@
 
 #![forbid(unsafe_code)]
 
+#[macro_use]
+extern crate clap;
+
```
bmwill, line 8: Don't use `extern crate` that's an old, deprecated syntax. You should be able to explicitly bring into scope the macros you want with `use clap::<macro>`

devtools/x/src/test.rs
```diff
@@ -49,6 +49,9 @@ pub fn run(mut args: Args, xctx: XContext) -> Result<()> {
     let config = xctx.config();
 
     let env_vars: &[(&str, &str)] = if args.html_cov_dir.is_some() {
+        if !xctx.installer().install_if_needed("grcov") {
+            std::process::exit(1);
```
bmwill, line 53: I'm thinking that a better way to handle exiting the process would be to throw an error (well return one) and have some logic at main that prints the error and exits. This way we only terminate the process from a single place cleanly.

devtools/x/src/tools.rs
```diff
@@ -12,17 +12,12 @@ pub struct Args {
...
-            true => installer::check_installed(key, value),
-        };
-        failed = failed || !success;
+    let success = match args.check {
+        false => xctx.installer().install_all(),
+        true => xctx.installer().check_all(),
+    };
+    match success {
+        false => std::process::exit(1),
+        true => Ok(()),
```
bmwill, line 21: I would probably use if/else statements instead of matching on a bool but its not an issue if you'd prefer this

bmwill (commented):
Overall looks like a good first stab at this. :)

x.toml
```diff
@@ -10,6 +8,23 @@ smoke-test = { path = "smoke-test" }
...
 flags = "-Zfeatures=all"
 
+[cargo.sccache]
+bucket = "ci-artifacts.libra.org"
+prefix = "sccache/libra/"
+public = true
+envs = [
+   #To debug sccache uncomment the two lines below.
+   ["SCCACHE_ERROR_LOG","/tmp/sccache_log"],
+   ["SCCACHE_LOG", "sccache::compiler::compiler=trace"],
```
rexhoffman, line 18: I'll comment these two lines before merge.   But, soo useful.....

devtools/x/src/config.rs
```diff
@@ -25,7 +25,38 @@ pub struct Config {
...
+    pub envs: Option<Vec<(String, String)>>,
+}
+
+#[derive(Clone, Serialize, Deserialize, Debug, PartialEq, Eq)]
+#[serde(rename_all = "kebab-case")]
+pub struct CargoInstallation {
+    pub version: String,
+    pub git: Option<String>,
+    pub git_rev: Option<String>,
+    pub features: Option<Vec<String>>,
```
bmwill, line 59: Some documentation on what this is trying to describe would be useful. It sort of seems like the version and git parts are mutually exclusive (given how its used below) perhaps this data structure should be better defined to describe that (ie using an enum)

bmwill (commented):
Left a few comments, overall I think this is looking pretty good. Maybe we can get @sunshowers to do a quick pass and provide their opinion too

.circleci/config.yml
```diff
@@ -961,18 +874,23 @@ workflows:
           requires:
             - prefetch-crates
       - build-dev:
+          context: libra_ci
```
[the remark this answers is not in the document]
rexhoffman: it passes in secrets, notable AWS credentials, only users who have access to the context can trigger the build.  So this means that builds tiggered by bors in the auto branch will uploads to aws.

devtools/x/src/config.rs
rexhoffman: Unfortunately the git revision and the version aren't quite mutually exclusive. I have to check the version to know whether to update or not.  Ideally pre release versions would contain their git hash.   But that's probably a harder sell to the authors of projects.

devtools/x/src/cargo.rs
```diff
@@ -201,13 +237,27 @@ impl Cargo {
...
-                .iter()
-                .for_each(|t| info!("Env {:?}: {:?}", t.0, t.1));
+            self.env_additions.iter().for_each(|t| {
+                if SECRET_ENVS
+                    .to_vec()
+                    .contains(&t.0.to_str().unwrap_or_default())
+                {
+                    info!("Env {:?}: ********", t.0);
+                } else {
+                    info!("Env {:?}: {:?}", t.0, t.1);
```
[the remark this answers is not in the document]
rexhoffman: It's important to see that they are there for debugging purposes.

devtools/x/src/build.rs
```diff
@@ -0,0 +1,276 @@
...
+    }
+
+    let cmd = CargoCommand::Build {
+        cargo_config: xctx.config().cargo_config(),
+        direct_args: direct_args.as_slice(),
+        args: &[],
+        env: &[],
+    };
+
+    if !args.package.is_empty() {
```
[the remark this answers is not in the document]
rexhoffman: It would, and did.   I removed it to make the tests pass yesterday -- probably around the same time as the review.  Good eyes though.

devtools/x/src/cargo.rs
```diff
@@ -201,13 +237,27 @@ impl Cargo {
 
         // once all the arguments are added to the command we can log it.
         if log {
-            self.env_additions
-                .iter()
-                .for_each(|t| info!("Env {:?}: {:?}", t.0, t.1));
+            self.env_additions.iter().for_each(|t| {
+                if SECRET_ENVS
+                    .to_vec()
```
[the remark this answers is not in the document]
rexhoffman: huh, yeah, not sure why I even thought I needed to convert to a vec, as slice has contains function as well.....

devtools/x/src/cargo.rs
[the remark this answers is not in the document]
sunshowers: Do you want to print out the None/Some status of secret env vars?

devtools/x/src/cargo.rs
```diff
@@ -15,15 +19,20 @@ use std::{
 const RUST_TOOLCHAIN_VERSION: &str = include_str!("../../../rust-toolchain");
 const RUSTUP_TOOLCHAIN: &str = "RUSTUP_TOOLCHAIN";
 const CARGO: &str = "CARGO";
+const SECRET_ENVS: &[&str] = &["AWS_ACCESS_KEY_ID", "AWS_SECRET_ACCESS_KEY"];
 
 pub struct Cargo {
     inner: Command,
     pass_through_args: Vec<OsString>,
-    env_additions: Vec<(OsString, OsString)>,
+    env_additions: Vec<(OsString, Option<OsString>)>,
```
sunshowers, line 27: I think the right data structure is a BTreeMap for this, not a Vec.

devtools/x/src/config.rs
sunshowers, line 59: Agreed with @bmwill, could these be documented?

devtools/x/src/context.rs
```diff
@@ -20,6 +21,11 @@ impl XContext {
     pub fn with_config(config: Config) -> Self {
         Self {
             core: XCoreContext::new(project_root()),
+            //not happy with the cloning action here....
```
sunshowers, line 24: Could you mark this as XXX? The way to separate it out would probably be to have data structures with/without.

devtools/x/src/context.rs
```diff
@@ -20,6 +21,11 @@ impl XContext {
...
+            installer: Installer {
+                cargo_config: config.cargo_config().clone(),
+                cargo_installations: config.tools(),
+            },
```
sunshowers, line 28: Could this be made a constructor?

devtools/x/src/installer.rs
```diff
@@ -1,42 +1,107 @@
...
-    //cargo's tooling to build our 3rd party cargo tools.  Looking to address that in a refactoring pr once all the needed
-    //components are in place -- perhaps by seperating xcontext in to constituent parts.
-    let config = CargoConfig {
-        toolchain: "stable".to_owned(),
-        flags: Option::None,
-    };
-    install_if_needed_config(&config, name, version)
+pub struct Installer {
+    pub cargo_config: CargoConfig,
+    pub cargo_installations: Vec<(String, CargoInstallation)>,
```
sunshowers, line 13: Could these be made private + a constructor?

devtools/x/src/installer.rs
```diff
@@ -1,42 +1,107 @@
...
+            cmd.arg("--version").arg(&installation.version);
+        }
+        cmd.arg(name);
+
+        let result = cmd.run();
+        if result.is_err() {
             info!(
-                "Could not install {} {}, check x.toml to ensure tool exists and is not yanked.",
-                name, version
+                "Could not install {} {}, check x.toml to ensure tool exists and is not yanked, or provide a git revision.",
```
sunshowers, line 93: How would a git revision be provided? Could the message be clearer?

devtools/x/src/installer.rs
```diff
@@ -1,42 +1,107 @@
...
+                cmd.arg(git_rev);
+            }
+        } else {
+            cmd.arg("--version").arg(&installation.version);
+        }
+        cmd.arg(name);
+
+        let result = cmd.run();
+        if result.is_err() {
             info!(
```
sunshowers, line 92: Should this be `err!`

devtools/x/src/main.rs
```diff
@@ -38,6 +39,9 @@ enum Command {
     #[structopt(name = "bench")]
     /// Run `cargo bench`
     Bench(bench::Args),
+    #[structopt(name = "build")]
+    /// Run `cargo build`
+    Build(Box<build::Args>),
```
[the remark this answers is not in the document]
sunshowers: Could the reason for boxing be documented?

x.toml
sunshowers: Oh this is nice :)

sunshowers (commented):
Looks quite good overall! Just a few comments.

devtools/x/src/cargo.rs
rexhoffman: It would need to be a BTreeMap<OsString, Option<OsString>>, so I could remove environment variables that are set to None.

Is that what you are expecting @sunshowers ?

sunshowers: Yep :)

rexhoffman: Do you want to print out the None/Some status of secret env vars? 

I couldn't think of a good way to textually represent a removal of a variable....   I suppose I could output the bash commands directly.....
```
export FOO="blah"
unset BIZ
```

rexhoffman: So I went off and started converting this, but I guess I disagree.   I want to retain the order of the envs are added (not sort them), and don't mind seeing an env added twice (and override) if the code generates it, a user would see a slightly non-sensical double addition of an environment variable.   But as you would expect the last one written wins.

The order is  a useful hint to the flow.   Having ordering matters in places where grcov/sccache are both run -- or other tools that may need env flags set.    Ideally I'd have a data type that retained order of insertion, but removed any overwritten value and appended it at the end of the sequence and removing it from the prior location......   

I think a linked list backed order preserving set would be the ideal abstraction.   I wouldn't mind making this a linked list instead and performing the removes myself it you're okay with that.

Or we can leave it as is until we hit that situation?

devtools/x/src/context.rs
rexhoffman: I'm not sure what you mean by "Could you mark this as XXX?"

I'll simply remove the line.

.circleci/config.yml
```diff
@@ -398,16 +322,15 @@ jobs:
     steps:
       - build_setup
       - restore_cargo_package_cache
-      - restore_sccache
       - run:
           name: cargo lint
-          command: $CARGO $CARGOFLAGS x lint
+          command: cargo x lint
```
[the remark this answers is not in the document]
rexhoffman: Currently we have cargo audit and cargo fetch being used in circle ci config without x equivalents.   I could add them, but would prefer not to in this pr, and which point these flags could be killed off (as well as the cargo-toolchain file).  Though work would need to be done to make sure the x.toml was sourced for the cargo versions to have installed in docker.

devtools/x/src/cargo.rs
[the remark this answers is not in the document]
rexhoffman: Yeppers,  need to see the None status to know that it will be removed if available.

devtools/x/src/installer.rs
rexhoffman: updated

mimoo:
sorry for the mirai-bot spam ^^ working on fixing this

devtools/x/Cargo.toml
```diff
@@ -23,3 +23,4 @@ glob = "0.3.0"
 x-core = { version = "0.1.0", path = "../x-core" }
 x-lint = { version = "0.1.0", path = "../x-lint" }
 libra-workspace-hack = { path = "../../common/workspace-hack", version = "0.1.0" }
+clap = "2.33.3"
```
sunshowers, line 26: I'd recommend using the version of `clap` that structopt re-exports, to minimize version skew issues: https://docs.rs/structopt/0.3.20/structopt/#reexports

devtools/x/src/cargo.rs
sunshowers: I like the idea of printing out the shell commands, makes it easier to copy-paste as well.

re the map, https://docs.rs/indexmap/1.6.0/indexmap/ may work as we discussed.

devtools/x/src/utils.rs
```diff
@@ -16,19 +21,102 @@ pub fn project_root() -> &'static Path {
...
+    }
+}
+
+pub fn apply_sccache_if_possible<'a>(
+    cargo_config: &'a CargoConfig,
+) -> Result<Vec<(&'a str, Option<&'a str>)>> {
+    let mut envs = vec![];
+
+    if option_env!("SKIP_SCCACHE").is_none() && cargo_config.sccache.is_some() {
+        if let Some(sccach_config) = &cargo_config.sccache {
```
sunshowers, line 58: Should the variable name be `sccache_config`?

devtools/x/src/utils.rs
```diff
@@ -16,19 +21,102 @@ pub fn project_root() -> &'static Path {
...
+            // Are we work on items in the right location:
+            // See: https://github.com/mozilla/sccache#known-caveats
+            let correct_location = option_env!("CARGO_HOME").unwrap_or_default()
+                == sccach_config.required_cargo_home
+                && sccach_config.required_git_home == project_root().to_str().unwrap_or_default();
+            if !correct_location {
+                warn!("You will not benefit from sccache in this build!!!");
+                warn!(
+                    "To get the best experience, please move your libra source code to {} and your set your CARGO_HOME to be {}, simply export it in your .profile or .bash_rc",
+                    &sccach_config.required_git_home, &sccach_config.required_cargo_home
```
sunshowers, line 68: Oh wow, had no idea sccache had such exacting requirements. Does this mean it's going to be hard for developers to set it up locally?

devtools/x/src/utils.rs
```diff
@@ -16,19 +21,102 @@ pub fn project_root() -> &'static Path {
...
+pub fn apply_sccache_if_possible<'a>(
+    cargo_config: &'a CargoConfig,
+) -> Result<Vec<(&'a str, Option<&'a str>)>> {
+    let mut envs = vec![];
+
+    if option_env!("SKIP_SCCACHE").is_none() && cargo_config.sccache.is_some() {
+        if let Some(sccach_config) = &cargo_config.sccache {
+            // Are we work on items in the right location:
+            // See: https://github.com/mozilla/sccache#known-caveats
+            let correct_location = option_env!("CARGO_HOME").unwrap_or_default()
```
sunshowers, line 61: To confirm, `option_env!` will be evaluated at compile time. Is this what we want? If so could you add a comment explaining this?

sunshowers (approved):
Looks great! Just a few comments.

devtools/x/src/utils.rs
rexhoffman: Took me about 3-4 minutes.   Big thing is to ensure your IDE has the same CARGO_HOME.   I'll probably get around to adding that to the dev_setup.sh as well.

devtools/x/Cargo.toml
rexhoffman: excellent, and done, didn't know that was a thing.

devtools/x/src/cargo.rs
rexhoffman: done. thanks for the pointer.

devtools/x/src/utils.rs
rexhoffman: I don't think so, since it can only turn sccache on and off at build time (you were correct to point out option_env! was producing code embedded in the binary.   That was not my intent in the least.

devtools/x/src/utils.rs
rexhoffman: Thank you so much for this pointer, it would have confused me to no end when I would have had to fix this in the future.   I've corrected the code and moved everything to var_os, and changed the value storage from Option<&'a str> to Option<String>.

PR #6215, opened 2020-09-25
Run dep-summaries workflow on auto pushes

metajack:
This adds dep-summaries workflow to push events to the auto branch in order to provide a hook to bors to gate landing on TCB dependency changes. It also reorganizes the workflow a bit to cut down on conditionals propagating through all the individual steps.

### DIFF
--- a/.github/workflows/dep-summaries.yml
+++ b/.github/workflows/dep-summaries.yml
@@ -11,36 +11,36 @@ on:
       - edited
 
 jobs:
-  annotate-dep-summaries:
+  calc-summaries:
     runs-on: ubuntu-latest
-    name: Annotate PRs with dependency summary changes
+    name: Calculate dependency summary changes
+    if: ${{ github.actor != 'bors-libra' }}
+    outputs:
+      lsr-summary-diff: ${{ steps.verify-lsr.diff }}
+      lec-summary-diff: ${{ steps.verify-lec.diff }}
+      release-summary-diff: ${{ steps.verify-release.diff }}
     steps:
       - name: checkout
-        if: ${{ github.actor != 'bors-libra' }}
         uses: actions/checkout@v2
         with:
           ref: ${{ github.event.pull_request.head.sha }}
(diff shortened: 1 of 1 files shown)


bmwill (approved):
I sort of wish there was a way to make these actions even simpler but this is much better than we had before :)

metajack:
Yeah, I started to want some abstraction capability, and checkout doesn't span job boundaries. We only need checkout to get the workflows, so I think this is the final nail in the coffin for in-repo action hosting. I'll likely move them to libra/actions soon.

bmwill:
One other thought, How does this check work if the required reviews are instead authors? Since they're authors they wouldn't be able to bypass this check by providing their own review.

metajack:
I believe the GitHub API just doesn't set that reviewer, which means the other reviewer will have to do it. GH prevents authors from getting review requests or approving reviews I believe.

metajack:
For example, I can pull_request_review to this PR, but only `comment` not `approve` or `request-changes`.

bmwill:
What I was trying to get at is, if @davidiw submits a PR to add a dep to saftey-rules then he'll have to rely on @mimoo to actually approve the change. Is this what we want? Or do we want to skip the check if david is the author. I don't have a preference either way, just pointing out a potential usability issue.

rexhoffman:
Would it be worth building a scripts to instead of add/removing messages just ensure that a message exists or doesn't with a specific content?

With a script like the one I built to handle the difference between auto/master pushes and prs (https://github.com/libra/libra/blob/master/.circleci/get_pr_info.sh) the comparisons would be easier to perform, even though we don't want comments on auto/master, the above could be used to simplify further, and decouple more gha.

rexhoffman:
Also looks like if a dep is removed from a pr, the required review will still be there?

metajack:
> Also looks like if a dep is removed from a pr, the required review will still be there?

The review request is not removed, but the requirement that a dep auditor approve a review will not trigger.

PR #6217, opened 2020-09-25
rust: convert Cargo.lock to v2 format

bmwill:
Use `cargo lock translate` to convert the project's Cargo.lock file to
the v2 format.


metajack (approved):
/land

Issue #6200, opened 2020-09-24
[logging] Consider putting JSON of data in message field


gregnazario:
At request of @davidiw 

### Current behavior
Lack of a message means no message
```
info!(key=value)
=> {"data":{"key":"value"}}
```
### Requested behavior
Lack of a message means put JSON of data in the message field
```
info!(key=value)
=> {"data":{"key":"value"}, "message":"{"key":value"}"}
```

bmwill:
I'm not quite sure what this is asking for. @davidiw could you elaborate a bit more?

gregnazario:
> I'm not quite sure what this is asking for. @davidiw could you elaborate a bit more?

I've updated with what david and I talked about

bmwill:
I'm not quite sure I understand how this is useful, other than doubling the number of bytes logged

davidiw:
so when I hop into kibana, the default log line is empty. That makes my logs useless without hitting the carrot so that you can see more information in it.

bmwill:
I would suggest you just add a message to your log lines then :P

davidiw:
my log lines are the message, it would literally be duplicating itself

davidiw:
https://github.com/libra/libra/blob/master/secure/net/src/lib.rs#L260 or https://github.com/libra/libra/blob/master/consensus/safety-rules/src/safety_rules.rs#L150

davidiw:
maybe this is a trivial tweak in kibana?

davidiw:
cc: @sherry-x thoughts?

sherry-x:
Yeah I've been having this issue as well, lots of logs does not have "message" field and if you try to just show message field it's empty. but I also agree setting message field is just duplication. we can probably do different filter for different module so that the output message is useful, example: https://85bd0f3d76314c71869bce3ae260f6d2.us-west1.gcp.cloud.es.io:9243/goto/3f58e3783a9f4bd352591287df1b9281

davidiw:
That doesn't really work well, @sherry-x, usually I'm not debugging a specific component but I'm debugging an issue and jumping across multiple components. @bmwill / @gregnazario, we really need to have a default string otherwise it makes kibana a pain to use. If there isn't an easy fix to duplicate the data upstream, I propose we just bite the bullet for now and duplicate the data in our code.

gregnazario:
I'm fine with duplicating the data, but we've got to be careful that the data isn't too large

sherry-x:
[...]

PR #6220, opened 2020-09-25
logger: add hack to ensure log lines aren't empty in Kibana

bmwill:
Closes: #6200

### DIFF
--- a/common/logger/src/libra_logger.rs
+++ b/common/logger/src/libra_logger.rs
@@ -245,20 +245,26 @@ impl LoggerService {
         for entry in self.receiver {
             PROCESSED_STRUCT_LOG_COUNT.inc();
 
-            if let Some(writer) = &mut writer {
-                Self::write_to_logstash(writer, &entry);
-            }
-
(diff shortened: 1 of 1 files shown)


common/logger/src/libra_logger.rs
```diff
@@ -245,20 +245,26 @@ impl LoggerService {
...
-            }
-
             if let Some(printer) = &self.printer {
                 let s = format(&entry).expect("Unable to format");
                 printer.write(s)
             }
+
+            if let Some(writer) = &mut writer {
+                Self::write_to_logstash(writer, entry);
+            }
```
davidiw, line 255: interesting, I hadn't realized that these were two different things... should we actually have a Vec of printers?

davidiw (approved):

davidiw:
Thanks man!

common/logger/src/libra_logger.rs
bmwill: Yeah we send things out to stderr as well as to a TCP endpoint.

PR #6221, opened 2020-09-25
[libra framework] move TransactionFee resource under treasury complia…

sblackshear:
…nce account

Previously, this resource lived under the LibraRoot account, which is rather large (~10 KB). Because paying gas fees requires writing to this resource and the granularity of storage writes is "entire account", a transaction with a nonzero gas price ends up being a lot more expensive than the same transaction with zero gas price.

This PR addresses the problem by placing the TransactionFee resource under the TC account instead. This account is much smaller, so this should be faster. In addition, this simplifies the TransactionFee code a bit because TC was already responsible for burning gas fees. Now, the code no longer needs to mention the LR account at all.

 cluster-test results with gas_price=1 before this PR:

`all up : 635 TPS, 7158 ms latency, 8500 ms p99 latency`
 
and after:

`all up : 874 TPS, 5189 ms latency, 6200 ms p99 latency`


testsuite/cluster-test/src/tx_emitter.rs
```diff
@@ -121,7 +121,7 @@ pub static GAS_UNIT_PRICE: Lazy<u64> = Lazy::new(|| {
     if let Ok(v) = env::var("GAS_UNIT_PRICE") {
         v.parse().expect("Failed to parse GAS_UNIT_PRICE")
     } else {
-        0_u64
+        1_u64
```
andll, line 124: Can you revert it now, before actually merging the code?

language/stdlib/modules/TransactionFee.move
```diff
@@ -28,20 +28,18 @@ module TransactionFee {
...
-        add_txn_fee_currency<Coin2>(lr_account, tc_account);
-        add_txn_fee_currency<LBR>(lr_account, tc_account);
+        add_txn_fee_currency<Coin1>(tc_account);
+        add_txn_fee_currency<Coin2>(tc_account);
+        add_txn_fee_currency<LBR>(tc_account);
     }
     spec fun initialize {
         include LibraTimestamp::AbortsIfNotGenesis;
-        include CoreAddresses::AbortsIfNotLibraRoot{account: lr_account};
+//        include CoreAddresses::AbortsIfNotLibraRoot{account: lr_account};
```
tzakian, line 42: Remove?

tzakian (approved):
LGTM!

language/stdlib/modules/TransactionFee.move
sblackshear: Oops, good catch.

PR #6225, opened 2020-09-25
[libra framework] specify cancel_burn script

emmazzz:
## Motivation

This PR specifies `cancel_burn` transaction script along with related functions in `LibraAccount` and `Libra`.

Yes.


junkil-park (approved):
LGTM!

language/stdlib/modules/LibraAccount.move
```diff
@@ -482,10 +482,9 @@ module LibraAccount {
         include Libra::AbortsIfNoCurrency<Token>;
     }
     spec schema DepositEnsures<Token> {
-        payer: address;
         payee: address;
         amount: u64;
-        ensures global<Balance<Token>>(payee).coin.value == old(global<Balance<Token>>(payee).coin.value) + amount;
+        ensures balance<Token>(payee) == old(balance<Token>(payee)) + amount;
```
junkil-park, line 487: I didn't know that we can do this. It's apparent here, but just to make sure, can we omit "global" when accessing a resource in the global state?

emmazzz: No we can't omit "global". `balance<Token>` is a move function defined in the module around line 1215. We can use `balance<Token>` in specs because it's pure.

junkil-park: Ah, I was confused. Thanks for you answer.

PR #6228, opened 2020-09-25
[move-prover] Added install script for cvc4

cbarrettfb:
This adds an install script for cvc4 so it can be used as a back-end for the move prover


junkil-park (approved):
LGTM!

PR #6230, opened 2020-09-25
[move-cli] options for gas metering

sblackshear:
- The --gas-budget flag enables gas metering with the given budget. Libra's genesis gas schedule is used, but in the future we can support custom ones
- Metering is off unless --gas-budget is specified


tzakian (approved):
LGTM!

PR #6234, opened 2020-09-26
[network] Add Error Metrics and On Chain Discovery metrics to Network

gregnazario:
### Overview
I've made error metrics and onchain discovery metrics for networking.  There seemed to be a lot of chatter about the way that I was doing the metrics before attached to logs, so I've removed it.

### DIFF
--- a/network/simple-onchain-discovery/src/lib.rs
+++ b/network/simple-onchain-discovery/src/lib.rs
@@ -6,12 +6,15 @@ use futures::{sink::SinkExt, StreamExt};
 use libra_config::{config::RoleType, network_id::NetworkContext};
 use libra_crypto::x25519;
 use libra_logger::prelude::*;
-use libra_metrics::{register_histogram, DurationHistogram};
+use libra_metrics::{
+    register_histogram, register_int_counter_vec, DurationHistogram, IntCounterVec,
+};
 use libra_network_address::NetworkAddress;
 use libra_network_address_encryption::{Encryptor, Error as EncryptorError};
 use libra_types::on_chain_config::{OnChainConfigPayload, ValidatorSet, ON_CHAIN_CONFIG_REGISTRY};
 use network::{
     connectivity_manager::{ConnectivityRequest, DiscoverySource},
+    counters::inc_by_with_context,
     logging::NetworkSchema,
 };
 use once_cell::sync::Lazy;
@@ -46,6 +49,15 @@ pub static EVENT_PROCESSING_LOOP_BUSY_DURATION_S: Lazy<DurationHistogram> = Lazy
     )
 });
 
+pub static DISCOVERY_COUNTS: Lazy<IntCounterVec> = Lazy::new(|| {
+    register_int_counter_vec!(
+        "libra_simple_onchain_discovery_counts",
+        "Histogram of busy time of spent in event processing loop",
+        &["role_type", "network_id", "peer_id", "metric"]
+    )
+    .unwrap()
+});
+
 /// Listener which converts published  updates from the OnChainConfig to ConnectivityRequests
 /// for the ConnectivityManager.
 pub struct ConfigurationChangeListener {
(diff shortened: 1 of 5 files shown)


network/src/logging.rs
```diff
@@ -72,6 +73,18 @@ impl<'a> NetworkSchema<'a> {
...
+        counter
+            .with_label_values(&[
+                self.network_context.role().as_str(),
+                self.network_context.network_id().as_str(),
+                self.network_context.peer_id().short_str().as_str(),
+            ])
+            .inc_by(amount);
+        self
+    }
+
```
bmwill, line 87: This is an interesting idea :)

gregnazario: After all the requests from others, I was seeing if we could automate the count dimensions in this way.  It's a little harder to have it not custom made, at least in this way.  If there are no dimensions, or at least static dimensions, they're probably pretty easy to set.

network/src/logging.rs
```diff
@@ -69,4 +70,33 @@ impl<'a> NetworkSchema<'a> {
...
+    pub fn with_count(self, counter: &IntCounterVec, amount: i64) -> Self {
+        counter
+            .with_label_values(&[
+                self.network_context.role().as_str(),
+                self.network_context.network_id().as_str(),
+                self.network_context.peer_id().short_str().as_str(),
+            ])
+            .inc_by(amount);
+        self
+    }
```
phlip9, line 84: this seems kind of awkward to me... tbh the problem with these metrics is taking an array of `&str`, which makes it kind of hard to compose functions that just add context

gregnazario: It's awkward, and very specific to networking.  I was just exploring ways to integrate counters into the logs.

Many people have been asking for counters integrated with logs

davidiw:
cool idea

network/simple-onchain-discovery/src/lib.rs
```diff
@@ -161,7 +176,11 @@ impl ConfigurationChangeListener {
             match self.conn_mgr_reqs_tx.send(update).await {
                 Ok(()) => (),
                 Err(e) => warn!(
-                    NetworkSchema::new(&self.network_context),
+                    NetworkSchema::new(&self.network_context).with_dimension_count(
+                        &DISCOVERY_COUNTS,
```
davidiw, line 180: hmm... I guess you don't need this as a formal field, since it is defined as the module in the logs

network/src/counters.rs
```diff
@@ -171,6 +171,33 @@ pub fn rpc_bytes(
     ])
 }
 
+pub static INVALID_RPC_MESSAGES: Lazy<IntCounterVec> = Lazy::new(|| {
```
davidiw, line 174: this is outbound, self-initiated, right?

network/src/counters.rs
```diff
@@ -171,6 +171,33 @@ pub fn rpc_bytes(
...
+pub static INVALID_DIRECT_SEND_MESSAGES: Lazy<IntCounterVec> = Lazy::new(|| {
+    register_int_counter_vec!(
+        "libra_network_invalid_ds_messages",
+        "Number of invalid direct send messages",
+        &["role_type", "network_id", "peer_id"]
+    )
+    .unwrap()
+});
+
+pub static INVALID_PEER_SEND_FAILURES: Lazy<IntCounterVec> = Lazy::new(|| {
```
davidiw, line 192: seems somewhat higher level than the above, why not include protocol_id in the previous two? or is this lower level (closer to the wire)

davidiw (approved):

network/src/counters.rs
gregnazario: Yes.

network/src/counters.rs
gregnazario: Oops, this actually doesn't look like it's used.

network/simple-onchain-discovery/src/lib.rs
gregnazario: I'm not sure I get what you mean, but I've gone ahead and split these metrics off the logs.

network/src/counters.rs
gregnazario: Fixed

gregnazario:
I've gone ahead and refactored this into a smaller and more sane PR, as well as I redid all the commits to make logical sense.

PR #6235, opened 2020-09-26
logger: don't panic when we're unable to serialize

bmwill:


msmouse (approved):
Are we confident this `unwrap()` is pretty impossible to hit?

https://github.com/libra/libra/blob/fc1a16af2667ae87fe3965553c080e8c74b9f0bf/common/logger/src/kv.rs#L34

bmwill:
Yes, generally it shouldn't ever be hit. And in this case it was good that we hit it because we were able to identify that there was an issue and fix it.

/land

gregnazario (approved):

common/logger/src/libra_logger.rs
```diff
@@ -51,7 +51,13 @@ impl LogEntry {
                 let v = match value {
                     Value::Debug(d) => serde_json::Value::String(format!("{:?}", d)),
                     Value::Display(d) => serde_json::Value::String(d.to_string()),
-                    Value::Serde(s) => serde_json::to_value(s).unwrap(),
+                    Value::Serde(s) => match serde_json::to_value(s) {
+                        Ok(value) => value,
+                        Err(e) => {
+                            eprintln!("error serializing structured log: {}", e);
```
gregnazario, line 57: Ah, I see what you mean by "silent", logging this way is truly silent.  I guess we can't do like a `warn!` in here because of the possible recursiveness?

bmwill: I'd probably need to do a bit of refactoring on the `Visitor` trait in order to be able to get that information out. And yeah I'm hesitant to log directly in the logging infra just to be careful and avoid recursing (although I'm sure it would be fine)

gregnazario: Another less fun way would be to force something that implements serialize to also implement Debug, and write it as a Debug value instead.

Let's ignore the location, but could we at least add the key?

PR #6236, opened 2020-09-26
[Key Manager] Force counter initialization at startup.

JoshLind:
## Motivation

This PR updates the key manager to initialize all metric counters at startup (so that the key manager dashboard shows zero values, rather than just being empty). This helps us to better distinguish between missing data, and counters with zero counts.

Yes.

## Test Plan

The key manager seems to run fine and all tests pass; once this lands, I'll spawn up a custom k8s testnet to verify that the dashboard panels are now all showing the correct values (seems to be the easiest way to check this change given that the key manager is not already running in the k8s testnet.)

## Related PRs

None.

### DIFF
--- a/secure/key-manager/src/counters.rs
+++ b/secure/key-manager/src/counters.rs
@@ -4,7 +4,8 @@
 use libra_secure_push_metrics::{register_int_counter_vec, IntCounterVec};
 use once_cell::sync::Lazy;
 
-static STATE_COUNTER: Lazy<IntCounterVec> = Lazy::new(|| {
+/// The metrics counter for the key manager.
+static COUNTER: Lazy<IntCounterVec> = Lazy::new(|| {
     register_int_counter_vec!(
         "libra_key_manager_state",
         "Outcome for key operations",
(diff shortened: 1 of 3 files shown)


secure/key-manager/src/counters.rs
```diff
@@ -13,6 +14,42 @@ static STATE_COUNTER: Lazy<IntCounterVec> = Lazy::new(|| {
...
+    let metric_counter_states = vec![
+        KEYS_STILL_FRESH,
+        LIVENESS_ERROR_ENCOUNTERED,
+        NO_ACTION,
+        ROTATED_IN_STORAGE,
+        SUBMITTED_ROTATION_TRANSACTION,
+        WAITING_ON_RECONFIGURATION,
+        WAITING_ON_TRANSACTION_EXECUTION,
+        UNEXPECTED_ERROR_ENCOUNTERED,
+    ];
```
bmwill, line 44: Small nit, this doesn't need to be a Vec, `&[]` would be sufficient, not that an allocation once is really going to hurt anything.

bmwill (approved):

secure/key-manager/src/counters.rs
JoshLind: Aah, sweet! Missed this -- thanks @bmwill 😄

secure/key-manager/src/counters.rs
```diff
@@ -13,6 +14,40 @@ static STATE_COUNTER: Lazy<IntCounterVec> = Lazy::new(|| {
...
+    let _ = metric_counter_states.iter().map(|metric_counter_state| {
+        COUNTER.with_label_values(metric_counter_state).reset();
+    });
+}
```
bmwill, line 48: Oh! Sorry I missed this. This won't actually initialize anything because this is just creating a lazy iterator. You should be able to change `map` to `for_each` or use a normal for loop to get the desired behavior.

/cancel

JoshLind: Ooh, thanks @bmwill, used "for_each". Kinda annoying that I can't actually test the functionality until it lands because spawning up a k8s environment from scratch is a real pain.. 😄

PR #6237, opened 2020-09-26
[language] Structured logging in the VM

dariorussi:
This is to integrate with structured logging and the rest of the system. 
It should allow us to connect errors to other part of the overall system.
It implies passing a logger to the VM, where the logger is a client controlled instance.
That ends up being a very intrusive change, yet inevitable.
This is the initial PR that allows us to hook up the logger in all critical place, 
more PRs will be coming shortly to add more info.
I also need to make sure everything works as expected and will do it in the next day or 
two once I understand what a good way to test this in the overall system is.


language/libra-vm/src/data_cache.rs
```diff
@@ -74,7 +74,20 @@ impl<'block> StateView for StateViewCache<'block> {
                 Ok(remote_data) => Ok(remote_data),
                 // TODO: should we forward some error info?
                 Err(e) => {
-                    error!("[VM] Error getting data from storage for {:?}", access_path);
+                    // create a LibraLogger from the `data_view` in scope. This logger
```
dariorussi, line 77: curious about opinions here, though it is a true pain to hook this up.
As the comment says though I wonder if it even matters to hook it up at all

language/libra-vm/src/logger.rs
```diff
@@ -0,0 +1,123 @@
...
+use crate::counters::CRITICAL_ERRORS;
+use libra_crypto::HashValue;
+use libra_logger::{prelude::*, Schema};
+use libra_state_view::StateViewId;
+use libra_types::transaction::Version;
+use move_vm_types::logger::Logger;
+use serde::Serialize;
+
+#[derive(Schema, Clone)]
+pub struct LogSchema {
```
dariorussi, line 13: this may change a bit shortly depending on what we need.
The connection to the executor and the `StateView` will be there but we may describe things differently in the VM. We'll see...

language/move-vm/runtime/src/data_cache.rs
```diff
@@ -189,11 +188,6 @@ impl<'r, 'l, C: RemoteCache> DataStore for TransactionDataCache<'r, 'l, C> {
                         None => {
                             let msg =
                                 format!("Failed to deserialize resource {} at {}!", ty_tag, addr);
-                            error!(
```
dariorussi, line 192: callers of this function log themselves so there is no point in logging here and it would be a pin to hook up the logger at this level

language/move-vm/runtime/src/interpreter.rs
```diff
@@ -56,14 +56,16 @@ macro_rules! set_err_info {
 ///
 /// An `Interpreter` instance is a stand alone execution context for a function.
 /// It mimics execution on a single thread, with an call stack and an operand stack.
-pub(crate) struct Interpreter {
+pub(crate) struct Interpreter<L: Logger> {
     /// Operand stack, where Move `Value`s are stored for stack operations.
     operand_stack: Stack,
     /// The stack of active functions.
     call_stack: CallStack,
+    // Logger to report information to clients
```
dariorussi, line 64: it makes it easier and more sane to save the logger in the `Interpreter` and use it from here. We would have to pass it around all over otherwise

language/testing-infra/e2e-tests/src/executor.rs
```diff
@@ -315,6 +315,7 @@ impl FakeExecutor {
             let vm = MoveVM::new();
             let remote_view = RemoteStorage::new(&self.data_store);
             let mut session = vm.new_session(&remote_view);
+            let logger = NoLogLogger;
```
dariorussi, line 318: I am not sure a no logger is right here, and have to play with it a bit to see what is the best solution.
We may change later. Opinions welcome.
Overall for tests it's not clear any logging is meaningful as if there are issues it's for the test to assert them...

language/testing-infra/test-generation/src/lib.rs
```diff
@@ -106,7 +106,8 @@ fn execute_function_in_module<S: StateView>(
 
         let internals = libra_vm.internals();
 
-        let gas_schedule = internals.gas_schedule()?;
+        let logger = NoLogLogger;
```
dariorussi, line 109: here too, as the previous one, not sure if `NoLogLogger` is good. Thoughts?

language/tools/move-cli/src/main.rs
```diff
@@ -220,6 +221,7 @@ fn run(
         })
         .collect();
 
+    let logger = StdErrLogger;
```
dariorussi, line 224: I suppose this is a question for @sblackshear: is a std error logger a good solution for the cli?

language/tools/transaction-replay/src/lib.rs
```diff
@@ -165,12 +166,14 @@ impl LibraDebugger {
         let is_version_ok = |version| {
             self.run_session_at_version(version, |session| {
                 let mut cost_strategy = CostStrategy::system(&gas_table, GasUnits::new(0));
+                let logger = StdErrLogger;
```
dariorussi, line 169: here also we are going to stderror, is that ok? another way to do all of those is to turn on logging on a normal log.
Thoughts? @runtian-zhou

language/tools/vm-genesis/src/lib.rs
```diff
@@ -119,6 +122,7 @@ pub fn encode_genesis_change_set(
 
     let move_vm = MoveVM::new();
     let mut session = move_vm.new_session(&data_cache);
+    let logger = NoLogLogger;
```
dariorussi, line 125: should genesis get a stderror logger just for good luck?

dariorussi:
in being a big and ugly diff this is simply threading a logger through the system.
Please review the questions in the comment and let's try to move it forward quick.

For the Rust experts here (@vgao1996, @tnowacki, @tzakian, maybe all of you except me), please review the usage of traits and suggest better ways to do it if you see good alternatives

language/tools/move-cli/src/main.rs
sblackshear: StdErrLogger sounds great. I am planning to add a `quiet` mode in the future, but I can choose `NoLogLogger` instead in that mode.

language/libra-vm/src/logger.rs
```diff
@@ -0,0 +1,126 @@
...
+    }
+
+    fn info(&self, message: &str) {
+        info!(self.schema.clone(), "{}", message);
+    }
+
+    fn debug(&self, message: &str) {
+        debug!(self.schema.clone(), "{}", message);
+    }
+}
```
bmwill, line 114: @dariorussi: This is sort of an anti-pattern w.r.t. logging and will probably result in things being much more difficult to understand and debug.

* The logging macros expand to capture the filename, module, and line number of where they're invoked. Wrapping the macros in a function call completely removes all of this information from the instrumentation location as now all logs in the VM will appear as if they're coming from this file instead of where they were invoked.

* We recently removed the `crit` log level from the rest of the system, it would probably be best to keep things similar across the stack by removing `crit` and supporting `trace` levels.

* Is there a particular reason why you wanted to thread through a logger via passing it in as a parameter instead of relying on the globally initialized one supplying by the logging facade in `libra-logger`?

More over, given you've started to go this route, is there something lacking from the logging framework that we could add that would better support your needs?

dariorussi: > @dariorussi: This is sort of an anti-pattern w.r.t. logging and will probably result in things being much more difficult to understand and debug.
> 
> * The logging macros expand to capture the filename, module, and line number of where they're invoked. Wrapping the macros in a function call completely removes all of this information from the instrumentation location as now all logs in the VM will appear as if they're coming from this file instead of where they were invoked.

I know and we have talked about this in the channel. We can think to go down a different path but it's not an easy one and later you will find an explanation. Talking to @gregnazario we concluded that, though this is not the ideal for the reason you just mentioned it may not be a deal breaker as the line is not used in a critical way. To me I understand that for certain generic errors (any of those where the string is not enough to identify the location) having the line number matters. In the VM case often the line number is insignificant because it's the line in the Move code that matters and that is not and cannot be reflected in the macro.
Having said so, suggestions welcome

> * We recently removed the `crit` log level from the rest of the system, it would probably be best to keep things similar across the stack by removing `crit` and supporting `trace` levels.

As far as we are concerned the difference between critical and error is in the incrementing a counter that we pick up in monitoring for an alert. Se that difference is vital to us and as you probably noticed both `crit` and `error` go to `error!`. I would like to keep it that way for the VM as it is significant. Again suggestion welcome.

> * Is there a particular reason why you wanted to thread through a logger via passing it in as a parameter instead of relying on the globally initialized one supplying by the logging facade in `libra-logger`?
> 

There are information that the Move VM is unaware of. Specifically everything Libra related is unknown to the Move VM. In that sense it gets hard to push knowledge of Libra into the VM and having a logger controlled from the outside is the only reasonable way.
Thoughts?

> More over, given you've started to go this route, is there something lacking from the logging framework that we could add that would better support your needs?

To be honest the way I have done it makes a lot of sense to me. So I don't know that there is much missing. It may be more for you to evaluate...

dariorussi:
I am going to put up and alternative PR for @bmwill and @gregnazario to look at and evaluate... stay tuned... and sorry I did not add you both to the review initially :(
Also for the VM people, the other PR is not going to be much different in what it touches (in fact most likely not at all) so if someone could familiarize themselves with the PR and get ready to accept that would be priceless!

dariorussi:
> I am going to put up and alternative PR for @bmwill and @gregnazario to look at and evaluate... stay tuned... and sorry I did not add you both to the review initially :(
> Also for the VM people, the other PR is not going to be much different in what it touches (in fact most likely not at all) so if someone could familiarize themselves with the PR and get ready to accept that would be priceless!

turns out my attempt to another PR ended up being much more complex for reason I explained to @bmwill offline.
We are going to continue with this and think about "better" ways in the coming days and iterate over it

language/libra-vm/src/logger.rs
```diff
@@ -0,0 +1,126 @@
...
+
+// Helper `Logger` implementation that does nothing
+#[derive(Clone)]
+pub struct NoLogLogger;
```
vgao1996, line 118: Should this be in `language/move-vm/types/src/logger.rs`?

language/move-vm/runtime/src/debug.rs
```diff
@@ -96,14 +99,14 @@ impl DebugContext {
...
-    pub(crate) fn debug_loop(
+    pub(crate) fn debug_loop<L: Logger>(
         &mut self,
         function_desc: &Function,
         locals: &Locals,
         pc: u16,
         instr: &Bytecode,
         resolver: &Loader,
-        interp: &Interpreter,
+        interp: &Interpreter<L>,
```
vgao1996, line 109: Nit & really no strong opinion on this, but `impl Logger` does seem a little bit more succinct and more consistent with the other function signatures in the Move VM.

language/move-vm/runtime/src/interpreter.rs
```diff
@@ -674,10 +683,10 @@ impl Frame {
     }
 
     /// Execute a Move function until a return or a call opcode is found.
-    fn execute_code(
+    fn execute_code<L: Logger>(
         &mut self,
         resolver: &Resolver,
-        interpreter: &mut Interpreter,
+        interpreter: &mut Interpreter<L>,
```
vgao1996, line 689: Same as above, maybe `&mut Interpreter<impl Logger>`?

language/tools/vm-genesis/src/lib.rs
vgao1996: Yeah I think it's a good idea to use stderr here as the information could help us understand what goes wrong when debugging. The bottom line is, this is an internal tool we use ourselves so I'd personally not worry too much about this spamming the out.

vgao1996 (commented):
I'll likely take another closer look later but overall the changes to the Move VM seem good to me.

language/libra-vm/src/logger.rs
dariorussi: good deal, done

language/move-vm/runtime/src/debug.rs
dariorussi: I did not know that and I like it! :)

language/move-vm/runtime/src/interpreter.rs
dariorussi: changed!

language/e2e-testsuite/src/tests/failed_transaction_tests.rs
```diff
@@ -17,6 +20,7 @@ fn failed_transaction_cleanup_test() {
     let sender = AccountData::new(1_000_000, 10);
     fake_executor.add_account_data(&sender);
 
+    let logger = NoLogLogger;
```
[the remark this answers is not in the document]
dariorussi: done

dariorussi:
> I'll likely take another closer look later but overall the changes to the Move VM seem good to me.

thanks @vgao1996! Much appreciated your feedback, keep that coming :)

dariorussi:
we talked about an alternative solution with @davidiw and @bmwill so something different is coming

dariorussi:
@bmwill @gregnazario @davidiw this removes the previous model and brings the macro back to where the "call" is. So effectively should be what we wanted. Take a look and see if it makes sense to you. 
I may change names a bit but this should be it

language/tools/transaction-replay/src/lib.rs
runtian-zhou: Sounds good to me!

davidiw (commented):
good for me, curious for @bmwill

language/libra-vm/src/errors.rs
```diff
@@ -35,18 +36,22 @@ fn error_split(code: u64) -> (u8, u64) {
...
         VMStatus::MoveAbort(location, code)
             if location != known_locations::account_module_abort() =>
         {
             let (category, reason) = error_split(code);
+            logger.alert();
             error!(
-                "[libra_vm] Unexpected prologue Move abort: {:?}::{:?} \
-                (Category: {:?} Reason: {:?})",
-                location, code, category, reason
+                logger.clone(),
```
bmwill, line 52: Remove all these clones, You shouldn't need to clone this. If you do then I need to fix the macros :)

dariorussi: well let's talk about it.....

bmwill: Turns out there's an issue because the trait isn't seen as being implemented for `&impl Trait`. This is something we could fix at a later point though

bmwill (commented):
Thanks for making these changes I'm much happier with this

dariorussi:
> Thanks for making these changes I'm much happier with this

well good we kept talking about it

dariorussi:
removing all `context.clone()` in favor of `*context` which hopefully is not going to copy or anything and it should be much cheaper and better. Thanks @bmwill

language/libra-vm/src/errors.rs
dariorussi: correct, and agree we can and should fix at some point.
However I think the way we have things now should be good enough

bob-wilson (approved):
This looks great to me.

language/libra-vm/src/logger.rs
tzakian: cc @vgao1996

PR #6239, opened 2020-09-27
[language][resource-viewer] use access path -> type map generated by …

sblackshear:
…static analysis

Previously, resource viewer used an access path -> type map generated by dynamic analysis of the genesis transaction. This means that resource viewer can only parse resource types that happen to be published in genesis, which is quite limiting for debugging.

This PR instead uses the comprehensive packed types map introduced in c8e6e2c91e1b18b548b345f2fbc9a9990b75f007. This map is computed via overapproximate static analysis of the Libra framework thus captures all possible types that can be instantiated in the Libra system. To put some numbers on the improvement: the map generated via dynamic analysis contains 48 types, whereas the map generated via static analysis contains 93.


language/move-prover/src/lib.rs
```diff
@@ -235,6 +236,15 @@ fn run_packed_types_gen(
...
+    let flattened_map = access_path_type_map
+        .into_iter()
+        .map(|(k, v)| (hex::encode(&k), v))
+        .collect::<Vec<_>>();
+    let types_json = serde_json::to_string_pretty(&flattened_map)?;
+    let mut types_file = File::create(options.output_path.clone())?;
+    types_file.write_all(&types_json.as_bytes())?;
+    types_file.write_all(b"\n")?;
+
     // TODO: save to disk, resource-viewer and an LCS schema extractor should look at this
```
wrwg, line 248: Is this TODO now obsolete?

wrwg (approved):

language/move-prover/src/lib.rs
sblackshear: Good call, will delete.

PR #6246, opened 2020-09-28
[mempool] structured logging + some counters

christinacelee:
## Motivation

(Write your answer here.)

## Test Plan

## Related PRs

### DIFF
--- a/mempool/src/shared_mempool/tasks.rs
+++ b/mempool/src/shared_mempool/tasks.rs
@@ -442,6 +442,9 @@ pub(crate) async fn process_state_sync_request(
     req: CommitNotification,
 ) {
     let start_time = Instant::now();
+    debug!(
+        LogSchema::event_log(LogEntry::StateSyncCommit, LogEvent::Received).state_sync_msg(&req)
+    );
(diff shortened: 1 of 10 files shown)


bmwill (approved):
This is really great! Only sort of nit is I wonder if you can figure out a way to construct a `TxnLog` without having to allocate a couple of Vecs, but that's probably not worth spinning extra cycles on right now

mempool/src/shared_mempool/tasks.rs
```diff
@@ -422,14 +400,20 @@ where
...
-        if vm_status.is_some() {
+    for (txn, (mempool_status, maybe_vm_status)) in results.iter() {
+        if let Some(vm_status) = maybe_vm_status {
             // log vm validation failure
+            error!(
+                SecurityEvent::InvalidTransactionMempool,
+                failed_transaction = txn,
+                vm_status = vm_status,
+                sender = sender,
+            );
```
davidiw, line 416: why is this an error? it seems to indicate that if Alice sent Bob a blob of vm invalid transactions, Bob would spam his error log? Maybe we should make this a histogram counter with a debug log?

PR #6247, opened 2020-09-28
[move-vm-types]: remove dependency on libra-types

lerencao:
## Motivation

Remove dependency on libra-types from move-vm-types.

Yes

## Test Plan

No test needed.

## Related PRs

None.

-------------------
const `MAX_TRANSACTION_SIZE_IN_BYTES` is moved from libra-types to nearby of `calculate_intrinsic_gas` in gas_schedule.
The const is used to check txn's size. 
I'm not sure whether we should use `max_transaction_size_in_bytes` of GasConstants arguments and remove the const once and for all.

### DIFF
--- a/language/move-vm/types/src/gas_schedule.rs
+++ b/language/move-vm/types/src/gas_schedule.rs
@@ -6,10 +6,13 @@
 //! It is important to note that the cost schedule defined in this file does not track hashing
 //! operations or other native operations; the cost of each native operation will be returned by the
 //! native function itself.
-use libra_types::{transaction::MAX_TRANSACTION_SIZE_IN_BYTES, vm_status::StatusCode};
 use mirai_annotations::*;
-use move_core_types::gas_schedule::{
-    AbstractMemorySize, CostTable, GasAlgebra, GasCarrier, GasConstants, GasCost, GasUnits,
+use move_core_types::{
+    gas_schedule::{
+        AbstractMemorySize, CostTable, GasAlgebra, GasCarrier, GasConstants, GasCost, GasUnits,
+        MAX_TRANSACTION_SIZE_IN_BYTES,
+    },
+    vm_status::StatusCode,
 };
 use vm::{
     errors::{Location, PartialVMError, PartialVMResult, VMResult},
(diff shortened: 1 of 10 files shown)


language/move-vm/types/src/gas_schedule.rs
```diff
@@ -21,6 +23,8 @@ use vm::{
     file_format_common::{instruction_key, Opcodes},
 };
 
+pub const MAX_TRANSACTION_SIZE_IN_BYTES: usize = 4096;
```
sblackshear, line 26: This feels like it should eventually be a parameter of `CostTable` to allow customization. But moving it here + killing the dep on `libra_types` is a great step in the right direction for now.

tzakian: This constant is already held in the `CostTable` :) In fact, as I was looking at it, there are a number of constants defined in that file already related to gas/transaction size, so that seems like a better place for this than here. 

Could you possibly move this constant to `move-core/types/src/gas_schedule.rs` and add it in with the other constant definitions there?

dariorussi: yeah I like this and removing the dependency is great! thanks!

lerencao: > This constant is already held in the `CostTable` in the `GasConstants` struct :) In fact, as I was looking at it, there are a number of constants defined in that file already related to gas/transaction size, so that seems like a better place for this than here.
> 
> Could you possibly move this constant to `move-core/types/src/gas_schedule.rs` and add it in with the other constant definitions there?

Updated. Now, it lives in move-core-types. Please take a look.

tzakian (approved):
This looks good to me.

Thanks for the contribution!

/land

lerencao:
@tzakian  thanks.  The bot failed to land, it asked my to `allow edit from maintainers`, which I think is a missing feature for forks from organizations. See https://github.com/isaacs/github/issues/1681 and https://github.community/t/how-can-we-enable-allow-edits-from-maintainers-by-default/2847/8. Didn't know this.
So I'm going to fire another PR from my own fork, and close this.

lerencao:
New PR is here: https://github.com/libra/libra/pull/6277

PR #6248, opened 2020-09-28
[ci] Fix up dep-summaries workflow after libra/actions move

metajack:
These two steps still had the old location which doesn't exist anymore.


bmwill (approved):
/land

PR #6250, opened 2020-09-28
[move cli] Refactor to utilize interface files

tnowacki:
- Added .mv file extension on modules
- Removed compile command
- Added check command for just checking a bunch of Move files
- Added publish command for publishing modules
- Flipped commit flags to a --dry-run flag

## Motivation

- Actually using the MV files should be a speedup for big cases
- A +/- is the stdlib now has to be specified and published as a first step
- We might want to make `check` and `build` commands that just delegate to the source language, could be neat. But this brings us closer to that world

## Test Plan

- in progress


language/tools/move-cli/src/main.rs
```diff
@@ -31,26 +31,35 @@ struct Move {
...
+        /// The source files containing modules to publish
+        #[structopt(
+            name = "PATH_TO_SOURCE_FILE",
+            default_value = MOVE_SRC,
+        )]
+        source_files: Vec<String>,
+        /// If set, the effects of executing `script_file` (i.e., published, updated, and
+        /// deleted resources) will NOT be committed to disk.
+        #[structopt(long = "dry-run")]
+        dry_run: bool,
```
sblackshear, line 62: I like the idea of making commits the default! Wondering about `dry-run` name though... would be nice to have something with a logical `short` (`-d` isn't great because it typically means something else). What about `read-only`?

language/tools/move-cli/src/main.rs
```diff
@@ -107,69 +116,64 @@ enum Command {
...
+}
 
-    // TODO: prevent cyclic dep?
-    let mut user_move_src_files = if Path::new(&args.move_src).exists() {
-        move_lang::find_move_filenames(&[args.move_src.clone()], true)?
-    } else {
-        vec![]
-    };
+fn interface_files_dir(args: &Move) -> Result<String> {
+    let dir = format!("{}/{}", &args.move_data, MOVE_COMPILED_INTERFACES_DIR);
```
sblackshear, line 135: Nit: use `Path` + `push` to avoid depending on `/` as a directory delimiter.

language/tools/move-cli/src/main.rs
```diff
@@ -107,69 +116,64 @@ enum Command {
...
 }
 
-/// Compile the user modules in `move_src` and the script in `script_file`
-fn compile(
-    args: &Move,
-    script_file: &Option<String>,
-) -> Result<(OnDiskStateView, Option<CompiledScript>)> {
-    let move_data = maybe_create_dir(&args.move_data)?;
+/// Generate interface files for published files
+fn generate_interface_files(args: &Move) -> Result<()> {
```
sblackshear, line 125: Maybe this should just take `move_data: Path` as an arg?

language/tools/move-cli/src/lib.rs
```diff
@@ -132,6 +126,7 @@ impl OnDiskStateView {
         let mut path = self.get_addr_path(module_id.address());
         path.push(MODULES_DIR);
         path.push(module_id.name().to_string());
+        path.set_extension(MOVE_COMPILED_EXTENSION);
```
sblackshear, line 129: Lol @me storing this with no extension before. We should probably make up an extension for resource files and event log files too...

sblackshear (commented):
We don't have to do it in this PR, but I think it's worth considering a setup where the CLI automatically starts off by publishing the stdlib (or a subset of it) unless you pass a flag that tells it not to. Otherwise, the extra `publish` step will get old quickly...

language/tools/move-cli/src/main.rs
```diff
@@ -107,69 +116,64 @@ enum Command {
...
+    move_lang::generate_interface_files(
+        &[args.move_data.clone()],
+        Some(args.move_data.clone()),
```
sblackshear, line 128: My mental model of `move_data` is "all the stuff that gets stored on-chain". Putting the interface files in `move_data` messes with that model a bit... Would you be amenable to create a separate directory (`generated`? `move_metadata`?) for artifacts like this that don't live on-chain? I expect that we will soon have a lot of other stuff to put in such a directory (e.g., the generated error descriptions, generated docs, generated tx script bindings).

language/tools/move-cli/src/main.rs
tnowacki: Very fancy. Will be doing another PR in move source lang land to clean that up

language/tools/move-cli/src/main.rs
tnowacki: Let me move it into the move_build dir. We are already pull in constants from the source language, so another won't hurt

language/move-lang/src/lib.rs
```diff
@@ -254,22 +257,19 @@ pub fn generate_interface_files(
             std::fs::read(mv_file)?.hash(&mut hasher)
```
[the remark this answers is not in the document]
tnowacki: I'll include it with the PR on Path stuff

sblackshear (approved):

language/tools/move-cli/src/main.rs
tnowacki: I like dry-run as I've personally seen it more often with the command line tools I've interacted with. 

Thoughts on `-n` for the short? It seems to be what git and ninja use

language/move-lang/src/lib.rs
```diff
@@ -40,6 +40,7 @@ use tempfile::NamedTempFile;
 
 pub const MOVE_EXTENSION: &str = "move";
 pub const MOVE_COMPILED_EXTENSION: &str = "mv";
+pub const MOVE_COMPILED_INTERFACES_DIR: &str = "mv_interfaces";
```
sblackshear, line 43: Alternative name ideas: `.mvi` or `.movei` (since it's basically a source file).

sblackshear (approved):

language/move-lang/src/lib.rs
tnowacki: movei isn't bad, but right now the compiler doesn't actually treat them differently than a move file, so for now I don't think its worthwhile to split

sblackshear: Oh, I somehow missed that this was a dir! Totally agree with `.move` to keep it simple for the compiler.

PR #6251, opened 2020-09-28
[json-rpc] endpoint and method processing log

xli:
1. log info for all http request & response status.
2. log info for all JSON-RPC request (method + params, batch all requests)
3. log debug for response success or non server internal error
4. log error for response server internal error
5. generate uniq trace id for all logs within one http request.

### DIFF
--- a/json-rpc/src/runtime.rs
+++ b/json-rpc/src/runtime.rs
@@ -9,7 +9,7 @@ use crate::{
 };
 use futures::future::join_all;
 use libra_config::config::{NodeConfig, RoleType};
-use libra_logger::{debug, error, info, Schema};
+use libra_logger::{info, log, Level, Schema};
 use libra_mempool::MempoolClientSender;
 use libra_types::{chain_id::ChainId, ledger_info::LedgerInfoWithSignatures};
 use rand::{rngs::OsRng, RngCore};
@@ -53,6 +53,27 @@ struct RpcResponseLog<'a> {
     is_batch: bool,
     response_error: bool,
     response: &'a JsonRpcResponse,
+}
+
+#[macro_export]
+macro_rules! log_response {
+    ($trace_id: expr, $resp: expr, $is_batch: expr) => {
+        let mut level = Level::Debug;
+        if let Some(ref error) = $resp.error {
+            if is_internal_error(error) {
+                level = Level::Error
+            }
+        }
+        log!(
+            level,
+            RpcResponseLog {
+                trace_id: $trace_id,
+                is_batch: $is_batch,
+                response_error: $resp.error.is_some(),
+                response: $resp,
+            }
+        );
+    };
 }
 
 /// Creates HTTP server (warp-based) that serves JSON RPC requests
@@ -180,11 +201,10 @@ async fn rpc_endpoint_without_metrics(
 
     let mut rng = OsRng;
     let trace_id = rng.next_u64();
-    let request_log = RpcRequestLog {
+    info!(RpcRequestLog {
         trace_id,
         request: data.clone(),
-    };
-    info!(request_log);
+    });
 
     let resp = Ok(if let Value::Array(requests) = data {
         match service.validate_batch_size_limit(requests.len()) {
@@ -202,7 +222,7 @@ async fn rpc_endpoint_without_metrics(
                 });
                 let responses = join_all(futures).await;
                 for resp in &responses {
-                    log_response(trace_id, &resp, true);
+                    log_response!(trace_id, &resp, true);
                 }
                 warp::reply::json(&responses)
             }
@@ -213,7 +233,7 @@ async fn rpc_endpoint_without_metrics(
                     ledger_info.ledger_info().timestamp_usecs(),
                 );
                 set_response_error(&mut resp, err);
-                log_response(trace_id, &resp, true);
+                log_response!(trace_id, &resp, true);
 
                 warp::reply::json(&resp)
             }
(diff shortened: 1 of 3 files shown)


libra-action:

Breaking changes policy:

2. Adding fields is ok.

json-rpc/src/runtime.rs
```diff
@@ -26,6 +28,33 @@ const LABEL_SUCCESS: &str = "success";
...
+    status: u16,
+    referer: Option<&'a str>,
+    user_agent: Option<&'a str>,
+    #[schema(debug)]
+    elapsed: std::time::Duration,
+}
+
+#[derive(Schema)]
+struct RpcRequestLog {
+    trace_id: u64,
```
gregnazario, line 46: I'd call this a `request_id`.

json-rpc/src/runtime.rs
```diff
@@ -39,11 +68,11 @@ pub fn bootstrap(
     chain_id: ChainId,
 ) -> Runtime {
     let runtime = Builder::new()
-        .thread_name("rpc-")
+        .thread_name("json-rpc")
```
gregnazario, line 71: So, that's why all the logs had thread names of `<component>-`

json-rpc/src/runtime.rs
```diff
@@ -26,6 +28,33 @@ const LABEL_SUCCESS: &str = "success";
...
+    path: String,
+    status: u16,
+    referer: Option<&'a str>,
+    user_agent: Option<&'a str>,
+    #[schema(debug)]
+    elapsed: std::time::Duration,
+}
+
+#[derive(Schema)]
+struct RpcRequestLog {
```
gregnazario, line 45: We've been mostly doing 1 schema per component, but tbh I think what you've done here is very clean for the code below.

json-rpc/src/runtime.rs
```diff
@@ -138,6 +178,14 @@ async fn rpc_endpoint_without_metrics(
         .get_latest_ledger_info()
         .map_err(|_| reject::custom(DatabaseError))?;
 
+    let mut rng = OsRng;
+    let trace_id = rng.next_u64();
+    let request_log = RpcRequestLog {
+        trace_id,
+        request: data.clone(),
+    };
+    info!(request_log);
```
gregnazario, line 187: Minor, you can put this into the macro, and then if the log level doesn't log this, it won't create the struct and do the cloning of data.

json-rpc/src/runtime.rs
```diff
@@ -161,19 +213,37 @@ async fn rpc_endpoint_without_metrics(
...
+        response_error: resp.error.is_some(),
+        response: resp,
+    };
+    if is_internal_error(resp.error.clone()) {
+        error!(log);
+    } else {
+        debug!(log);
+    }
+}
+
```
bmwill, line 246: I highly suggest avoiding this anit-pattern and use the logging macros directly where you want to log. This will ensure that you can get the proper file/line number/module included in the logs themselves.

json-rpc/src/runtime.rs
```diff
@@ -291,6 +363,14 @@ fn set_response_error(response: &mut JsonRpcResponse, error: JsonRpcError) {
     response.error = Some(error);
 }
 
+fn is_internal_error(error: Option<JsonRpcError>) -> bool {
+    if let Some(e) = error {
+        e.code <= -32000 && e.code >= -32099
+    } else {
+        false
+    }
+}
```
gregnazario, line 372: You should probably be able to pass in a reference to the `Option<JsonRpcError>` and then you won't need to clone it.

json-rpc/src/runtime.rs
gregnazario: Yeah, in this case, it'll tell you whether it was in the batch / single API call  or error etc.

I think this might be one of the more acceptable spots of this anti-pattern, because it's highly generalized.

bmwill: I still suggest changing if possible, or write a short macro that just expands to this instead.

json-rpc/src/runtime.rs
bmwill: Specifically you'll probably want the signature to be `Option<&JsonRpcError>` instead of `&Option<JsonRpcError>`

json-rpc/src/runtime.rs
xli: suggestions are welcome, but I need more hint about what you mean, I'm trying my best with this doc: https://developers.libra.org/docs/rustdocs/libra_logger/index.html

json-rpc/src/runtime.rs
xli: yes, after checked log, I think remove last "-" looks better, what you think?

json-rpc/src/runtime.rs
xli: request_id maybe confusing because json-rpc protocol has request { id ... }, hence give trace_id.
also, for batch request, we will only have one trace id for all json-rpc requests

json-rpc/src/runtime.rs
xli: I'm guessing you are asking me log where the error created. 
this log is not for that purpose, this is at endpoint level to understand an internal unexpected error maybe returned, and we need to know what is that error.
For the case where error created, it's up to internal component to decide whether we need log, because some errors maybe handled at high level, for example, let's say account not found in database, if it is considered as an error at db level, but at get_account method, we return None instead of return error. 
Hence technically at the place we create the error, we don't know whether we returned the error to client as internal error.

Ideally, we should not have any error log here, but I'm sure we will see some in production.

bmwill: I'm asking that you remove and stop using a "log" function and instead just use the logging macros ("info", "error", etc) in-place of using a function like this.

json-rpc/src/runtime.rs
gregnazario: Ah interesting okay.  So, not telling any difference between any requests on a batch command

json-rpc/src/runtime.rs
```diff
@@ -26,6 +28,54 @@ const LABEL_SUCCESS: &str = "success";
...
+            level,
+            RpcResponseLog {
+                trace_id: $trace_id,
+                is_batch: $is_batch,
+                response_error: $resp.error.is_some(),
+                response: $resp,
+            }
+        );
+    };
+}
```
bmwill, line 77: Thanks for making this change!

bmwill (approved):

json-rpc/src/runtime.rs
bmwill: Only nit here is that you may want to use fully qualified names to name the macro you're using (so you don't have to bring it in scope elsewhere). e.g `::libra_logger::log!(...)` instead of `log!(..)`

PR #6253, opened 2020-09-28
[transaction-replay] Refactored out the api needed for communicating …

runtian-zhou:
…with validator nodes

## Motivation

Refactored out the api for communicating with validator/full node software. This new `libra-validator-interface` will become a basic building block in the future for building writesets for on-chain administrative operations.

Yes

## Test Plan

`cargo build`


Cargo.toml
```diff
@@ -91,6 +91,7 @@ members = [
     "language/testing-infra/test-generation",
     "language/tools/disassembler",
     "language/tools/genesis-viewer",
+    "language/tools/libra-validator-interface",
```
sblackshear, line 94: Does this libra-specific tool belong in `language/tools`? 

As we think about splitting Move into it's own repo, it seems easiest if all the tools are Move-related and not Libra-related (which I think is mostly true except for transaction-replay).

runtian-zhou: Ahhh I was just thinking about putting thi new cratge inside the `tool` and didn't think about this complication. I would actually argue that it's the transaction-replay and the vm-genesis that is more libra-related than move related. Should we move both crates to a new folder under `language/`? OR `libra-vm/`

sblackshear: Perhaps something like `language/libra-tools` or `libra/tools`? Not sure.

runtian-zhou:
Update: Moved `transaction-replay` and `libra-validator-interface` to `language/libra-tools` so that we can have a unified folder for all libra network related transactions.

sblackshear (approved):
LGTM!

PR #6254, opened 2020-09-28
[dockerhub] always prune on image push to dockerhub

rexhoffman:
## Motivation

Dockhub old image pruning hadn't occurred in a while, since it was gated behind the ci_base image getting updated on a file change.   Old bug, didn't notice while working on the base image after having switched it over to this mechanism from a daily build.

Yes

## Test Plan
CI

## Related PRs
None

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -717,10 +717,6 @@ jobs:
                 command: |
                   #signs the pushed image
                   docker push --disable-content-trust=false libra/build_environment:<< parameters.target-image >>-latest
-            - run:
-                name: docker image pruning.
-                command: |
-                  scripts/dockerhub_prune.sh -u "${DOCKERHUB_USERNAME}"" -p ${DOCKERHUB_PASSWORD}" -x
 
   docker-update-base-images:
     executor: vm-medium-executor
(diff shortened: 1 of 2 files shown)


davidiw:
Is there a way to verify this before you land? Like did the trim happen.

rexhoffman:
@davidiw I had to implement pagination, just pushed that up now, it's running in branch test-2 which should in about 20 minutes delete a good number of old releases, the main problem you're going to face is all the master releases you'll see which are mostly just noise

rexhoffman:
That build is here:   https://app.circleci.com/pipelines/github/libra/libra/31811/workflows/5252c076-3202-49b0-9655-338e4800e482/jobs/224294

Expecting a lot of cruft to be removed from pages 3-4.   Master builds are all pretty new though, so bumped the deletion to be 7 days for non-release images.   @sausagee

rexhoffman:
Well ****, looks like rustc can't be downloaded, and failing out the builds, both last one and latest:

https://app.circleci.com/pipelines/github/libra/libra/31811/workflows/227457ea-59dc-4451-81cc-b7a5a1256a95/jobs/224333

I'll retry the job in a few hours and see if rust infra is back up.   Heh, at least you can still do builds.

rexhoffman:
Here, it's working, deleting around 200 or so images from each repo.

https://app.circleci.com/pipelines/github/libra/libra/31811/workflows/227457ea-59dc-4451-81cc-b7a5a1256a95/jobs/224333/parallel-runs/0/steps/0-114

davidiw (approved):
Awesome! What happens if we trim one but then want to relabel it (master -> testnet). Is that even reasonable?

rexhoffman:
I had to bump the credentials to admin mode, so the deletes would succeed.. rerunning.

rexhoffman:
What happens if we trim one but then want to relabel it (master -> testnet). Is that even reasonable?   We'd have to rebuild it.   I don't think it's likely to happen, but it could.  Depends on the lag for test net, if it gets from 4 hours to 7 days we might be in trouble?

sausagee:
Testnet is currently released from master/trunk. Once we cut the release branch, testnet will be released from the release branch. In that case, we would relabel release-* --> testnet.  The testnet labeling should be fine as the release-* images are kept for 90 days.

rexhoffman:
Actually that didn't do the delete, just claimed it did, but failed.  And...  pagination was screwed up by the deletes.   So fixed all the above.   Not sure if it was a bad u/p passed to the script from circleci, that I overlooked, or if I really needed to change the login mechanism.... oh well.

Proof of it working is here:
https://app.circleci.com/pipelines/github/libra/libra/31892/workflows/5ef73c36-fed4-4b3a-b182-fc9c083c44d2/jobs/224996/parallel-runs/0/steps/0-113
coupled with better logging and the repos having shrunk to less than 1/2 the size.
https://hub.docker.com/repository/registry-1.docker.io/libra/validator/tags?page=6

PR #6255, opened 2020-09-28
[language] test logging

dariorussi:
trying canary for logging....

### DIFF
--- a/language/move-vm/types/src/logger.rs
+++ b/language/move-vm/types/src/logger.rs
@@ -1,0 +1,49 @@
+// Copyright (c) The Libra Core Contributors
+// SPDX-License-Identifier: Apache-2.0
+
+// Trait used by the VM to log interesting data.
+// Clients are responsible for the implementation.
+pub trait Logger: Clone {
+    // A critical error occurred. Typically an invariant violation, and generally anything
+    // that can compromise the behavior of the system.
+    fn crit(&self, message: &str);
+    // An error occurred, severity is high. The error should be investigated but it
+    // does not compromise the health of the system.
+    fn error(&self, message: &str);
+    // An error occurred, severity is low. The error is unexpected under normal conditions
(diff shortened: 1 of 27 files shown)


language/move-vm/types/src/logger.rs
```diff
@@ -0,0 +1,49 @@
+// Copyright (c) The Libra Core Contributors
```
davidiw, line 1: why are we introducing a new logging framework :O?

cc: @bmwill

language/move-vm/types/src/logger.rs
```diff
@@ -0,0 +1,49 @@
...
+
+// A logger for the VM that prints to stderr
+#[derive(Clone)]
+pub struct StdErrLogger;
+
+impl Logger for StdErrLogger {
+    fn crit(&self, message: &str) {
+        eprintln!("{}", format!("crit: {}", message).as_str());
+    }
+
```
davidiw, line 33: if you want to do your own logger, it will ultimately need to hook into the libra logger or it won't make it to our logging framework.

also recommend consider looking at rust log crate

dariorussi: this is just for testing or cli tools and something we could avoid too if needed.
This is not what the runtime/production code will use.
Also this is just a toy PR I will add you to the proper one

language/move-vm/types/src/logger.rs
dariorussi: LOL, that is not quite accurate as mentioned offline as well, also I have added you to the right PR #6237 sorry I did not do it sooner. This is just a test to verify that, in fact, the structure logging that is in common/logger is in fact working as expected

dariorussi: also in this specific PR please take a look at `language/libra-vm/src/logger.rs` to get an answer to your question (or your concerns)

PR #6256, opened 2020-09-28
[json-rpc] validate version params, return client error code instead of internal error code

xli:
1. added test to cover error cases, and enforce return code -32602
2. created parse_version_param to handle version param parsing and validation.

### DIFF
--- a/json-rpc/src/tests/unit_tests.rs
+++ b/json-rpc/src/tests/unit_tests.rs
@@ -278,6 +278,22 @@ fn test_json_rpc_protocol_invalid_requests() {
             }),
         ),
         (
+            "get_metadata invalid arguments: version is too large",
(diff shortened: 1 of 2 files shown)


libra-action:

Breaking changes policy:

2. Adding fields is ok.

json-rpc/src/tests/unit_tests.rs
```diff
@@ -277,6 +277,22 @@ fn test_json_rpc_protocol_invalid_requests() {
                 "libra_ledger_version": version
             }),
         ),
+        (
+            "get_metadata invalid arguments: version is too large",
```
thefallentree, line 281: too high?

thefallentree: out of range?

thefallentree (approved):

PR #6258, opened 2020-09-29
[language/tools] Add account diff to transaction-replay tool

tzakian:
A pretty simple PR that adds a subcommand to the `transaction-replay` tool to allow diffing an account state at two different versions. 

```
$ cargo run -- -u https://testnet.libra.org/v1 diff-account 000000000000000000000000000000dd 0 2340288
```
Screen Shot 2020-09-28 at 5 42 53 PM


sblackshear (approved):
This is awesome!! Should also add this for the CLI at some point.

PR #6260, opened 2020-09-29
[network] Add proxy protocol support

gregnazario:
### Overview

Proxy protocol allows for retrieving the original address associated with a proxied call through a proxy like HAProxy.  It can be configured to be turned on / off from `NetworkConfig`

https://www.haproxy.org/download/1.8/doc/proxy-protocol.txt

### Testing
We now have some pretty comprehensive unit testing for both positive and negative cases for the ProxyProtocol receiver.

Currently, all other integration tests and unit tests do not use the proxy protocol, and I want to run it on a testnet to test it out.

### Next Steps
I may need to write a ProxyProtocol sender, so we can send ProxyProtocol wrapped messages to test in smoke tests or some other integration test.


davidiw:
should we test on a testnet first? cc: @mgorven

davidiw:
AWESOME!

davidiw:
What's the plan for connecting this? Can we detect if the haproxy itself enabled this or not?

mgorven:
Sweet, awesome stuff!

This should be configured in `NetworkConfig`; the k8s configs will enable it for the networks fronted by HAproxy.

gregnazario:
> What's the plan for connecting this? Can we detect if the haproxy itself enabled this or not?

So, the tricky thing here is we could accept both proxy protocol and non-proxy protocol data, but it becomes confusing and a headache.  I'm trying to think of a way we could connect it in through the NetworkConfig so, we enable a K8s config to set up both.

> Sweet, awesome stuff!
> 
> This should be configured in `NetworkConfig`; the k8s configs will enable it for the networks fronted by HAproxy.

I'll take a look into working through that.


Additionally, we should check if we need to turn on any proxy protocol handling on HA proxy from the load balancers in front of it:
* AWS NLB [preserves the client ip without needing proxy protocol.  Though, also supports proxy protocol V2](https://docs.aws.amazon.com/elasticloadbalancing/latest/network/load-balancer-target-groups.html)
* GCP LB [does not preserve the client IP, and only supports Proxy Protocol V1](https://cloud.google.com/load-balancing/docs/tcp/setting-up-tcp#proxy-protocol)
* Azure ELB [does preserve the client IP](https://docs.microsoft.com/en-us/azure/load-balancer/concepts)

mgorven:
> So, the tricky thing here is we could accept both proxy protocol and non-proxy protocol data, but it becomes confusing and a headache. I'm trying to think of a way we could connect it in through the NetworkConfig so, we enable a K8s config to set up both.

You don't want to accept proxy-protocol if the service isn't fronted by a trusted load balancer, because it would allow clients to forge their IP. You could get fancier with the configs with `no`, `accept` and `require` options, but I think a simple on/off config is fine here.

> Additionally, we should check if we need to turn on any proxy protocol handling on HA proxy from the load balancers in front of it:

I believe the source IP is retained in all cloud providers because we're doing L4 balancing.

gregnazario:
> > Additionally, we should check if we need to turn on any proxy protocol handling on HA proxy from the load balancers in front of it:
> 
> I believe the source IP is retained in all cloud providers because we're doing L4 balancing.

You would think that.  But, I know better :). Most of them should do it correctly, but it's implemented differently per cloud provider

mgorven:
I'm not assuming, I mean that I'm fairly sure I've verified this before :-)

network/src/protocols/wire/proxy_protocol.rs
```diff
@@ -0,0 +1,369 @@
+// Copyright (c) The Libra Core Contributors
```
phlip9, line 1: let's move this into netcore, since it's pretty generic and not libranet specific

network/src/protocols/wire/proxy_protocol.rs
```diff
@@ -0,0 +1,369 @@
...
+//! - Rejects ProxyProtocol V1
+//! - Only supports IPv4 & IPv6 TCP mode
+//! - All other valid connections, will just drop the address information and use the proxy's one
+//!
+//! ## Interpetations not in the spec
+//! - An address space that doesn't match the size expected is rejected e.g. too big for IPv4
+//! - Address space that's larger than the current supported requests is rejected
+
+use crate::transport::TSocket;
+use byteorder::{BigEndian, ReadBytesExt};
```
phlip9, line 18: i don't think we need byteorder any more, pretty sure this is included in std now

for example [`std::u16::from_be_bytes`](https://doc.rust-lang.org/std/primitive.u16.html#method.from_be_bytes)

network/src/protocols/wire/proxy_protocol.rs
```diff
@@ -0,0 +1,369 @@
...
+const UDP_UNIX: u8 = 0x32;
+
+// Address sizes
+const IPV4_SIZE: u16 = 12;
+const IPV6_SIZE: u16 = 36;
+const UNIX_SIZE: u16 = 216;
+
+/// Read a proxy protocol event and unwrap the address information associated.
+///
+pub async fn read_proxy_protocol_event<T: TSocket>(
```
phlip9, line 53: we can relax the generic here to just `TSocket: AsyncRead`

network/src/protocols/wire/proxy_protocol.rs
```diff
@@ -0,0 +1,369 @@
...
+    original_addr: &NetworkAddress,
+    stream: &mut T,
+) -> io::Result<NetworkAddress> {
+    // This is small enough that it should not be fragmented by TCP
+    let header: &mut [u8; 12] = &mut [0; 12];
+    stream.read_exact(header).await?;
```
phlip9, line 59: this is kind of weird to me, would prefer

```rust
let header = [0_u8; 12];
stream.read_exact(&mut header).await?;
```

network/src/protocols/wire/proxy_protocol.rs
```diff
@@ -0,0 +1,369 @@
...
+    // If it's not proxy protocol, let's stop
+    if header != &PPV2_SIGNATURE {
+        return Err(io::Error::new(
+            io::ErrorKind::InvalidInput,
+            "Invalid proxy protocol signature",
+        ));
+    }
+
+    // High 4 bits is version, low 4 bits is command
+    let version_and_command: &mut [u8; 1] = &mut [0; 1];
```
phlip9, line 70: note, fwiw, at this layer the underlying TcpSocket isn't buffered, so we're gonna be making like 5 separate read syscalls

network/src/protocols/wire/proxy_protocol.rs
```diff
@@ -0,0 +1,369 @@
...
+                "Unsupported Address Family or Protocol",
+            ));
+        }
+    };
+
+    Ok(source_address)
+}
+
+#[cfg(test)]
+mod test {
```
phlip9, line 162: do you know if proxy protocol lists any test vectors we could also sanity check against?

network/builder/src/dummy.rs
```diff
@@ -154,6 +154,7 @@ pub fn setup_network() -> DummyNetwork {
         listener_addr,
         authentication_mode,
         constants::MAX_FRAME_SIZE,
+        false,
```
phlip9, line 157: let's add block comments like `false, /* enable_proxy_protocol */` to aid readability

network/Cargo.toml
```diff
@@ -11,6 +11,7 @@ edition = "2018"
 
 [dependencies]
 anyhow = "1.0.32"
+byteorder = { version = "1.3.4", default-features = false }
```
phlip9, line 14: delete

network/src/protocols/wire/proxy_protocol.rs
phlip9, line 53: maybe rename to `read_header` and then caller can use `proxy_protocol::read_header`

network/src/protocols/wire/proxy_protocol.rs
gregnazario: I actually did have this, and it didn't compile or something, I don't remember why I switched back to byteorder

network/src/protocols/wire/proxy_protocol.rs
gregnazario: Yeah, I can change it to 2 reads instead of the 5.

network/src/protocols/wire/proxy_protocol.rs
gregnazario: It does not, but it does list this code in C++ at the bottom of the spec, it's just kinda messy.

I read through the whole spec and basically enumerated the cases we want to cover in these tests.

network/src/protocols/wire/proxy_protocol.rs
phlip9: yeah gotta make sure you pass in a `[u8; X]` to `from_be_bytes` instead of a `&[u8]` slice. note, if you pull in `use bytes::Buf;` you can also read everything into a big buffer and then pull out the pieces like

```rust
let mut buf = [0_u8; 32];
socket.read_exact(&mut buf).await?;

let mut cursor = &mut buf[..];

if cursor.bytes()[..HEADER_SIZE] == &PPV2_SIGNATURE {
    cursor.advance(HEADER_SIZE);
} else {
    Err(..)
}
let version_and_command = cursor.get_u8();
let family_and_protocol = cursor.get_u16();
```

gregnazario: I've updated it to not use the `bytes::buf`, but it uses a `try_into` I don't like.

Tbh, this is a lot of overhead and boiler plate just to convert a couple bytes into a number....

network/src/protocols/wire/proxy_protocol.rs
gregnazario: AsyncRead + Unpin

network/src/protocols/wire/proxy_protocol.rs
gregnazario: Done

network/src/protocols/wire/proxy_protocol.rs
gregnazario: Changed to 2 calls now.  One to read the header, and another to read the address space.

network/src/protocols/wire/proxy_protocol.rs
gregnazario: Moved.  I'm not really clear on all the location distinctions still, but I'll go with it

network/builder/src/dummy.rs
gregnazario: Done

network/src/transport.rs
```diff
@@ -97,6 +100,7 @@ pub struct ConnectionMetadata {
     pub remote_peer_id: PeerId,
     pub connection_id: ConnectionId,
     pub addr: NetworkAddress,
+    pub proxy_addr: Option<NetworkAddress>,
```
phlip9, line 103: nit: it looks like this is unused, let's remove it?

phlip9 (approved):
LGTM, land after fixing nits

we could probably also write a quick fuzz test for this, but let's leave that for another PR

network/src/transport.rs
gregnazario: Oops yeah, should remove.

I was trying to think of ways to add both the original address (address of the proxy), and the address of the client (the one that's being encoded).  But, I think for now we can simplify it by leaving it just as the best case closest to the client address.

gregnazario:
> LGTM, land after fixing nits
> 
> we could probably also write a quick fuzz test for this, but let's leave that for another PR

I first want to make sure that I can get it working in a testnet before I push this PR

gregnazario:
/land

We need to test this in a testnet, but my testnet attempted testing has been broken for 2 days.

If proxy protocol isn't enabled there shouldn't be any changes @mgorven hopefully we can test this before we enable it.

mgorven:
Yup, looks fine to land and I can then help test it.

config/src/config/network_config.rs
```diff
@@ -59,6 +59,8 @@ pub struct NetworkConfig {
     // in case some peers don't have well defined addresses.
     pub seed_pubkeys: SeedPublicKeys,
     pub max_frame_size: usize,
+    // Enables proxy protocol on incoming connections to get original source addresses
```
davidiw, line 62: can you point to the spec?

gregnazario: The proxy protocol spec? https://www.haproxy.org/download/1.8/doc/proxy-protocol.txt

PR #6261, opened 2020-09-29
[move-prover][stdlib] Abstract FixedPoint32, avoiding non-linear arithmetics

wrwg:
This PR introduces an abstraction for the module FixedPoint32 with the goal to avoid non-linear arithmetics in verification conditions. It appears that those are a major source of butterflies.

Instead of arbitrary rationals, we use fixed ones for the value of 0.5 and 1.0. The abstraction is performed in a way that the abortion behavior is preserved. Therefore verification of callers is orthogonal (as long as it does not reflect directly over FixedPoint32 values). More precisely, if f is a function in FP32, and A the abstraction, then we have A(f(x)) ~ A(f)(A(x)) (considering aborts as special values).

All existing specifications verify with the abstraction (specifically, no abort conditions are missed, which the prover would complain about). A few timeouts around non-linear arithmetics apparently can be removed. Unfortunately, `LibraAccount::unstaple_lbr` does still timeout.

The specification (see [here](https://github.com/wrwg/libra/blob/abstract-fixedpoint/language/stdlib/modules/doc/FixedPoint32.md)) makes use of the `[abstract]` and `[concrete]` properties of conditions in a non-trivial way for the first time. It appeared necessary to enable the syntax `include [property] Schema` to make this approach practical, so this was added as a side-effect of this PR.

## Motivation

Eliminate timeouts.

Yes

## Test Plan

Re-enabled verifications which did timeout.

## Related PRs

NA

### DIFF
--- a/language/stdlib/modules/Authenticator.move
+++ b/language/stdlib/modules/Authenticator.move
@@ -64,10 +64,11 @@ module Authenticator {
     }
     spec fun ed25519_authentication_key {
         pragma opaque = true;
-        pragma verify = false;
         aborts_if false;
-        ensures result == spec_ed25519_authentication_key(public_key);
+        ensures [abstract] result == spec_ed25519_authentication_key(public_key);
     }
+    /// We use an uninterpreted function to represent the result of key construction. The actual value
+    /// does not matter for the verification of callers.
     spec define spec_ed25519_authentication_key(public_key: vector<u8>): vector<u8>;
 
     /// Compute a multied25519 account authentication key for the policy `k`
--- a/language/stdlib/modules/FixedPoint32.move
+++ b/language/stdlib/modules/FixedPoint32.move
@@ -44,19 +44,50 @@ module FixedPoint32 {
         assert(product <= MAX_U64, Errors::limit_exceeded(EMULTIPLICATION));
         (product as u64)
     }
+
+    /// We specify the concrete semantics of the implementation but use
+    /// an abstracted, simplified semantics for verification of callers. For the verification outcome of
+    /// callers, the actual result of this function is not relevant, as long as the abstraction behaves
+    /// homomorphic.
     spec fun multiply_u64 {
         pragma opaque;
-        include MultiplyAbortsIf;
-        ensures result == spec_multiply_u64(val, multiplier);
+        include [concrete] ConcreteMultiplyAbortsIf;
+        ensures [concrete] result == spec_concrete_multiply_u64(val, multiplier);
+        include [abstract] MultiplyAbortsIf;
+        ensures [abstract] result == spec_multiply_u64(val, multiplier);
     }
+    spec schema ConcreteMultiplyAbortsIf {
+        val: num;
+        multiplier: FixedPoint32;
+        aborts_if spec_concrete_multiply_u64(val, multiplier) > MAX_U64 with Errors::LIMIT_EXCEEDED;
+    }
+    spec define spec_concrete_multiply_u64(val: num, multiplier: FixedPoint32): num {
+        (val * multiplier.value) >> 32
+    }
+
+    /// # Abstract Semantics
+
     spec schema MultiplyAbortsIf {
         val: num;
         multiplier: FixedPoint32;
         aborts_if spec_multiply_u64(val, multiplier) > MAX_U64 with Errors::LIMIT_EXCEEDED;
     }
+
     spec define spec_multiply_u64(val: num, multiplier: FixedPoint32): num {
-        (val * multiplier.value) >> 32
+        if (multiplier.value == 0)
+            // Zero value
+            0
+        else if (multiplier.value == 1)
+            // 1.0
+            val
+        else if (multiplier.value == 2)
+            // 0.5
+            val / 2
+        else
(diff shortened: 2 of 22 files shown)


language/stdlib/modules/Authenticator.move
```diff
@@ -64,10 +64,11 @@ module Authenticator {
     }
     spec fun ed25519_authentication_key {
         pragma opaque = true;
-        pragma verify = false;
         aborts_if false;
-        ensures result == spec_ed25519_authentication_key(public_key);
+        ensures [abstract] result == spec_ed25519_authentication_key(public_key);
     }
+    /// We use an uninterpreted function to represent the result of key construction. The actual value
+    /// does not matter for the verification of calles.
```
DavidLDill, line 71: Typo: calles ==> calls

DavidLDill (approved):

language/stdlib/modules/FixedPoint32.move
```diff
@@ -44,20 +44,51 @@ module FixedPoint32 {
...
+            0
+        else if (multiplier.value == 1)
+            // 1.0
+            val
+        else if (multiplier.value == 2)
+            // 0.5
+            val / 2
+        else
+            // overflow
+            MAX_U64 + 1
```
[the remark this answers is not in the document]
wrwg: There can't be any more arithmetic after the overflow case because than it aborts. Indeed as always the aborts is not in the spec function which just is partial, but in the Move function which has the abstracted semantics. So I think this is in fact a sound abstraction in the sense that A(f(x)) == A(f)(f(A(x)), that is if you only observe via the abstracted viewpoint.

language/stdlib/modules/FixedPoint32.move
```diff
@@ -44,20 +44,51 @@ module FixedPoint32 {
         assert(product <= MAX_U64, Errors::limit_exceeded(EMULTIPLICATION));
         (product as u64)
     }
+
+    /// We specify the concrete semantics of the implementation but use
+    /// an abstracted, simplified semantics for verification of callers. For the verification outcome of
+    /// callers, the actual result of this function is not relevant, as long as the abstraction behaves
+    /// homomorphic.
```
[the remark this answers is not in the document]
wrwg: Yes, in the sense as stated above, I think it is homomorphic.

language/stdlib/modules/FixedPoint32.move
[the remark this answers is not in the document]
wrwg: Well, I think I get your point. Let's say f(x) = x - c. Then the above equation might not hold: A(f(x)) might be defined but A(x) - A(c) is not. It's though interesting to see that this is only module definedness, and I guess there is some kind of theory which captures this kind of refinement module definedness.

Practically, it means that not all computations theoretically possible with Fixedpoint32 are homomorphic for this abstraction. But this would be the case for any kind of abstraction we do perform. For example, we can't prevent someone from doing `fp.value`. We need to lookout that those computations aren't performed in the code we verify.

language/stdlib/modules/FixedPoint32.move
```diff
@@ -44,20 +44,51 @@ module FixedPoint32 {
...
     }
+
     spec define spec_multiply_u64(val: num, multiplier: FixedPoint32): num {
-        (val * multiplier.value) >> 32
+        if (multiplier.value == 0)
+            // Zero value
+            0
+        else if (multiplier.value == 1)
+            // 1.0
+            val
```
junkil-park, line 82: It seems that the multiplication result does not match to the multiplier. If `multiplier.value == 1`, then the result of the multiplication is `val * (1 / 2^32)` because `multiplier` is equivalent to `(1 / 2^32)`.

`else if (multiplier.value == (2 << 32))` seems the right branching condition for the case of "val * 1".

language/stdlib/modules/FixedPoint32.move
```diff
@@ -44,20 +44,51 @@ module FixedPoint32 {
...
+        else if (multiplier.value == 2)
+            // 0.5
+            val / 2
```
junkil-park, line 85: Similarly, `else if (multiplier.value == (2 << 31))` seems the right branching condition for the case of "val * 0.5".

language/stdlib/modules/FixedPoint32.move
```diff
@@ -44,20 +44,51 @@ module FixedPoint32 {
...
+        else
```
junkil-park, line 86: The followings can be the conditions for overflow and underflow.
```
else if (multiplier.value > (2 << 32))
            // overflow
            MAX_U64 + 1
else if (multiplier.value < (2 << 31))
            // underflow
            0
```

language/stdlib/modules/FixedPoint32.move
wrwg: It does intentionally not match the multiplier because it is an abstraction and the representation is different than in the concrete case.

language/stdlib/modules/FixedPoint32.move
wrwg: No, this is not the intent here. It is a different representation.

PR #6262, opened 2020-09-29
[language] use system temp dir in interface generator

guangyuz:
- DEFAULT_OUTPUT_DIR may be inaccessible in some cases

## Motivation

(Write your answer here.)

## Test Plan

## Related PRs

### DIFF
--- a/language/move-lang/src/lib.rs
+++ b/language/move-lang/src/lib.rs
@@ -37,6 +37,7 @@ use std::{
     str::Chars,
 };
 use tempfile::NamedTempFile;
+use libra_temppath::TempPath;
 
 pub const MOVE_EXTENSION: &str = "move";
 pub const MOVE_COMPILED_EXTENSION: &str = "mv";
@@ -242,7 +243,11 @@ pub fn generate_interface_files(
     }
 
     let interface_files_dir =
-        interface_files_dir_opt.unwrap_or_else(|| command_line::DEFAULT_OUTPUT_DIR.to_string());
+        interface_files_dir_opt.unwrap_or_else(|| {
+            let temp_path = TempPath::new();
+            temp_path.create_as_dir().expect("Create temp dir fail.");
+            temp_path.path().to_str().unwrap().to_owned()
+        });
     let hash_dir = {
         use std::{
             collections::hash_map::DefaultHasher,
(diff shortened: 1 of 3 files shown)


guangyuz:
https://github.com/libra/libra/issues/6263

language/move-lang/src/lib.rs
```diff
@@ -242,7 +243,11 @@ pub fn generate_interface_files(
     }
 
     let interface_files_dir =
-        interface_files_dir_opt.unwrap_or_else(|| command_line::DEFAULT_OUTPUT_DIR.to_string());
+        interface_files_dir_opt.unwrap_or_else(|| {
+            let temp_path = TempPath::new();
+            temp_path.create_as_dir().expect("Create temp dir fail.");
+            temp_path.path().to_str().unwrap().to_owned()
+        });
```
davidiw, line 250: if we do this, the files will all be deleted... perhaps instead we should do a check on DEFAULT_OUTPUT_DIR and make sure it is accessible, writable, and a dir?

sblackshear: Agreed. @tnowacki will know for sure, but I believe the whole point of interface files is to avoid recompilation by putting the interface files somewhere where subsequent runs of the compiler will see them. That won't work if we put them in a randomly generated temporary directory....

tnowacki: Yeah, this isn't really good solution as all of the files might be blown away. I'm particularly surprised the tests don't fail with this. I tried using temp dirs, but the tests tended to fail in multithreaded environments because interface files would be deleted in one thread before the test finished. 

"command_line::DEFAULT_OUTPUT_DIR is used to save interface files. However this dir may be inaccessible in some cases, for example the user has no write permission to it." 

I sort of feel like this whole thing is a non-issue. But I'll post in #6262

tnowacki (changes requested):
TempDirs might get blown away, meaning the interface files might be gone before they are used. 

Let's talk some more on the issue page :)

language/move-lang/src/lib.rs
guangyuz: As I explained in the issue page, my test environment has a different configuration. When I run the cargo test, the case sanity_check_testsuite will try to create the directory move_build_output under language/move-lang. Since my environment is write-protected on the source code directory, the test will fail. So there is no problem and I'll close the bug.

PR #6269, opened 2020-09-29
update boogie to 2.7.32 and small fix in dev-setup.sh

shazqadeer:
## Motivation

Updating Boogie to pull in two new features:
- shadowing of bound variables
- augmented error traces

Also fixed some minor issues in prover tool installation.

Yes.

## Test Plan

Tested move prover locally after updating boogie.

### DIFF
--- a/scripts/dev_setup.sh
+++ b/scripts/dev_setup.sh
@@ -20,7 +20,7 @@ HELM_VERSION=3.2.4
 VAULT_VERSION=1.5.0
 Z3_VERSION=4.8.9
 DOTNET_VERSION=3.1
-BOOGIE_VERSION=2.7.30
+BOOGIE_VERSION=2.7.32
 
 SCRIPT_PATH="$( cd "$( dirname "$0" )" >/dev/null 2>&1 && pwd )"
 cd "$SCRIPT_PATH/.." || exit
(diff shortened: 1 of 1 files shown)


scripts/dev_setup.sh
```diff
@@ -346,7 +345,7 @@ function install_boogie {
   if [[ "$("$HOME"/.dotnet/dotnet tool list -g)" =~ .*boogie.*${BOOGIE_VERSION}.* ]]; then
     echo "Boogie $BOOGIE_VERSION already installed"
   else
-    "$HOME/.dotnet/dotnet" tool install --global Boogie --version "$BOOGIE_VERSION"
+    "$HOME/.dotnet/dotnet" tool update --global Boogie --version $BOOGIE_VERSION
```
wrwg, line 348: I intentionally changed this to install because update does it everytime even if the right version is already installed (which is checked above). Any reason you changed it back? I was assuming the check above deals with avoiding failure of install.

shazqadeer: install wasn't working.  When I updated BOOGIE_VERSION to 2.7.32 and ran dev_setup.sh, it didn't update the version.  In  other words, it executes the install command but does not update saying that boogie is already installed.

wrwg: Strange. It does work in CI and all different images we test (those are on a clean machine). Anyway, fine with me to change this.

wrwg (approved):

PR #6270, opened 2020-09-29
[Logging] Avoiding possible DoS through logs

AnomalRoil:
## Motivation

This is adding a check that the data is coming from a somewhat trusted or controlled network when we log the messages themselves, to prevent a public full node to be able to DoS our log system with super large messages.

We only log the possibly big part if we are on the Private or Validator network.

## Related PRs
 
See the discussions in #5905

### DIFF
--- a/mempool/src/shared_mempool/coordinator.rs
+++ b/mempool/src/shared_mempool/coordinator.rs
@@ -23,7 +23,10 @@ use futures::{
     stream::{select_all, FuturesUnordered},
     StreamExt,
 };
-use libra_config::{config::PeerNetworkId, network_id::NodeNetworkId};
+use libra_config::{
+    config::PeerNetworkId,
+    network_id::{NetworkId, NodeNetworkId},
+};
 use libra_logger::prelude::*;
 use libra_trace::prelude::*;
 use libra_types::{on_chain_config::OnChainConfigPayload, transaction::SignedTransaction};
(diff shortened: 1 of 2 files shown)


AnomalRoil:
I guess I should tag @davidiw @mimoo and @christinacelee for having your opinion if this is needed or not?

I've noticed that it seems some of what was discussed in #5905 was done in #6246

davidiw:
It might be worth exploring some form of conditional logging where I can use log!( level = $is_public ? debug : error or some other fancy macro. Also it might be good to be holistic in this effort and avoid one offs or at least establish an agreed paradigm so we don't end up with with too many repeat efforts.

See: https://github.com/libra/libra/pull/6246/files#r497640751

cc: @bmwill for thoughts.

AnomalRoil:
I love the idea of a conditional logging macro :D

bmwill:
I don't know if I really understand wanting a "conditional" log macro. To me it seems like just using a clear `if` statement would be clearer about what is going on. If you want to rate limit a log line you can wrap it with the `sample!` macro though.

mimoo:
you definitely don't want to use sample! on byzantine logging

mimoo (approved):
LGTM!

davidiw (changes requested):
let's discuss this first, we're logging a lot of stuff on a public network... I don't think this is what we want. I'm not sure why we'd want to log a message or an error.

mempool/src/shared_mempool/coordinator.rs
```diff
@@ -171,11 +174,22 @@ pub(crate) async fn coordinator<V>(
...
+                                            peer_id = peer_id,
+                                        );
+                                    }
+                                    _ => {
+                                        error!(
+                                            SecurityEvent::InvalidNetworkEventMempool,
+                                            peer_id = peer_id,
+                                        );
+                                    }
+                                }
```
davidiw, line 191: 1) why is this an error? some remote peer sent us a flaky message, considering this can be easily triggered from the remote side, it seems a debug at best
2) the msg is interesting, I would consider using a sampling here if you really want to log this but I'm not convinced
3) perhaps we should just get a counter? because in general we won't be able to do anything with this except learn information

davidiw:

I think next week we should spend some time wearing a VFN hat and exploring our logs and coming up with a unified interface for making this convenient.

Thanks for initiating this!

PR #6271, opened 2020-09-29
[libra framework] Add `0x1` as a reserved address.

### LINKED ISSUE #5813: add 0x1 as a reserved address

tzakian:
Adds `0x1` as a reserved address where an account cannot be created. 

Closes #5813


language/stdlib/modules/CoreAddresses.move
```diff
@@ -34,6 +34,13 @@ module CoreAddresses {
         0x0
     }
 
+    /// The reserved address where all core modules are published. When module
```
sblackshear, line 37: Maybe remove comment about "when module publishing is closed" to avoid confusion? We can update it if publishing opens up.

sblackshear (commented):
I am in favor of this by I remember that @dariorussi had some concerns (though maybe his concerns were about reserving more addresses, can't remember).

vgao1996:
I also feel like maybe we should reserve a whole block of addresses rather than just 1. (Kinda like port numbers?)

sblackshear:
> I also feel like maybe we should reserve a whole block of addresses rather than just 1. (Kinda like port numbers?)

I think this was also discussed but was even more controversial.

Unfortunately, even if we wanted to do this I don't think there's an easy way to covert an `address` to a u128 so we can efficiently reserve the range. It's an easy native function to write, but we're nearly out of time...

dariorussi (commented):
I am not sure I follow completely but reserving addresses is fine by me. Particularly if they are really "reserved".
I had issues with using an address value (it was 0x0 at the time and thus the "default" value for an address) as a privileged address.
But maybe `Signer` and other things have resolved that concern?

sblackshear (approved):
Do we want to consider the address to u128 native function + reserve a block of addrs as @vgao1996 suggests? Or are we ok going forward with this?

tzakian:
I think with how close things are to code freeze that we shouldn't implement the native option and only reserve 0x1 for now. At a later date we may want to go for the range-based solution though.

PR #6275, opened 2020-09-29
[libra framework] add TransactionChecks pre condition and specify …

emmazzz:
…more scripts

## Motivation

This PR
- adds `LibraAccount::TransactionChecks` schema specifying properties checked by the prologue
(currently there are just two: account exists and is not frozen),
- insert the schema mentioned above to all the specified scripts,
- adds specifications for `create_parent_vasp_account` and `create_designated_dealer` scripts, as well as corresponding functions in `LibraAccount`.

Yes.

### DIFF
--- a/language/stdlib/modules/LibraAccount.move
+++ b/language/stdlib/modules/LibraAccount.move
@@ -452,6 +452,7 @@ module LibraAccount {
         modifies global<LibraAccount>(payee);
         modifies global<AccountLimits::Window<Token>>(VASP::spec_parent_address(payee));
         ensures exists<LibraAccount>(payee);
+        ensures exists<Balance<Token>>(payee);
         ensures global<LibraAccount>(payee).withdrawal_capability
             == old(global<LibraAccount>(payee).withdrawal_capability);
         include DepositAbortsIf<Token>{amount: to_deposit.value};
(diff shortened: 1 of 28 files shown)


language/stdlib/modules/LibraAccount.move
```diff
@@ -1422,6 +1491,17 @@ module LibraAccount {
         aborts_if !LibraTransactionPublishingOption::spec_is_script_allowed(sender, script_hash) with Errors::INVALID_STATE;
     }
 
+    spec define prologue_guarantees(sender: signer) : bool {
+        let addr = Signer::spec_address_of(sender);
+        LibraTimestamp::is_operating() && exists_at(addr) && !AccountFreezing::account_is_frozen(addr)
+    }
+
+    /// Used in transaction script to specify pre-conditions checked by the prologue.
+    spec schema TransactionCertified {
```
DavidLDill, line 1500: Is "TransactionCertified" a good name?  I don't understand it.  PrologueGuarantees or something like that would be more immediately understandable.

DavidLDill (approved):

language/stdlib/modules/LibraAccount.move
wrwg: What about `IsValidTransaction`? Emma and I discussed the "guarantees" yesterday, and it is also not to good. The `IsValid` seems to be the most neutral.

wrwg (approved):
Looks great modulo renaming of the schema. Let's try to land this quickly to avoid merge conflicts. I think @DavidLDill 's PR is in now.

PR #6276, opened 2020-09-30
[move cli] a few small improvements

sblackshear:
- Use .lcs extension for resource and event files
- Fix off-by-1 error in disassembler + remove TODO
- Mention invalid shifts in arithmetic error message


language/tools/disassembler/src/disassembler.rs
```diff
@@ -781,8 +780,10 @@ impl<Location: Clone + Eq> Disassembler<Location> {
             .into_iter()
             .enumerate()
             .map(|(instr_index, dis_instr)| {
+                // +1 because bytecode instructions are 1-indexed
```
tnowacki, line 783: Is that really true? Or maybe I'm confused on this context

language/tools/move-cli/src/lib.rs
```diff
@@ -44,6 +44,9 @@ pub const MOVE_SRC_ADDRESS: Address = Address::new([
     0u8, 0u8, 0u8, 0u8, 0u8, 0u8, 0u8, 0u8, 0u8, 0u8, 0u8, 0u8, 0u8, 0u8, 0u8, 2u8,
 ]);
 
+/// Extension for resource and event files, which are in LCS format
+const LCS_EXTENSION: &str = "lcs";
```
tnowacki, line 48: 👍

tnowacki (approved):

language/tools/disassembler/src/disassembler.rs
sblackshear: Good question; this made me realize that what's I'm doing is wrong. There is definitely an off-by-1 error somewhere, but I'm not sure where it's coming from. Here's a repro (inside `language/tools/move-cli/tests/testsuite/explain_arithmetic_failure`):

```
λ movec script.move
λ movep move_build_output/scripts/main.mv --script 
module ffffffff.<SELF> {


public <SELF>() {
B0:
	0: LdU64(1)
	1: LdU64(2)
	2: Sub
	3: Pop
	4: Ret
}
}
λ move run script.move
Compiling 1 source file(s)
Execution failed because of an arithmetic error (i.e., integer overflow, underflow, or divide-by-zero) in script at code offset 3
```

The failure should be at offset 2, not offset 3, I think perhaps what is happening is that the the VM gives you the offset of the instruction *after* the failure occurred. Does that seem right/is it expected @dariorussi / @vgao1996 ?

PR #6277, opened 2020-09-30
[move-vm-types]: remove dependency on libra-types

lerencao:
## Motivation

See https://github.com/libra/libra/pull/6247


tzakian (approved):
Looks good! Hopefully the land will work :) 

/land

PR #6278, opened 2020-09-30
[move-cli] allow all tests in a directory to be run

mengxu-fb:
## Motivation

Right now, we need to run `move-cli test path/to/args.txt`, which is rather inconvenient. This pull request additionally allows `move-cli test <dir_containing_args.txt>` or `move-cli test <dir containing_several_tests>`.

Yes

## Test Plan

Backward compatibility: `./target/release/move-cli test language/tools/move-cli/tests/testsuite/events_emit_view/args.txt`
Expected output: 1 / 1 tests passed.

Run all tests under the directory: `./target/release/move-cli test language/tools/move-cli/tests/testsuite`
Expected output: 14 / 14 tests passed.

## Related PRs

### DIFF
--- a/language/tools/move-cli/Cargo.toml
+++ b/language/tools/move-cli/Cargo.toml
@@ -35,4 +35,3 @@ lcs = { path = "../../../common/lcs", version = "0.1.0", package = "libra-canoni
 
 [[test]]
 name = "cli_testsuite"
-harness = false
--- a/language/tools/move-cli/src/test.rs
+++ b/language/tools/move-cli/src/test.rs
@@ -139,3 +139,19 @@ pub fn run_one(args_path: &Path, cli_binary: &str) -> anyhow::Result<()> {
         Ok(())
     }
 }
+
+pub fn run_all(args_path: &str, cli_binary: &str) -> anyhow::Result<()> {
+    let mut test_total = 0;
+    let mut test_passed = 0;
+    for entry in move_lang::find_filenames(&[args_path.to_owned()], |fpath| {
+        fpath.file_name().expect("unexpected file entry path") == "args.txt"
+    })? {
+        match run_one(Path::new(&entry), cli_binary) {
+            Ok(_) => test_passed += 1,
+            Err(ex) => eprintln!("Test {} failed with error: {}", entry, ex),
+        }
+        test_total += 1;
+    }
+    println!("{} / {} test(s) passed.", test_total, test_passed);
+    Ok(())
+}
--- a/language/tools/move-cli/tests/cli_testsuite.rs
+++ b/language/tools/move-cli/tests/cli_testsuite.rs
@@ -3,13 +3,33 @@
 
 use move_cli::test;
 
-use std::path::Path;
+use std::path::PathBuf;
 
-pub const CLI_BINARY: &str = "../../../target/debug/move-cli";
-pub const CLI_TESTSUITE_DIR: &str = "tests/testsuite";
+pub const CLI_BINARY_PATH: [&str; 6] = ["..", "..", "..", "target", "debug", "move-cli"];
+pub const CLI_METATEST_PATH: [&str; 3] = ["tests", "metatests", "args.txt"];
+pub const CLI_TESTSUITE_PATH: [&str; 2] = ["tests", "testsuite"];
 
-fn run_all(args_path: &Path) -> datatest_stable::Result<()> {
-    Ok(test::run_one(args_path, CLI_BINARY)?)
+fn get_cli_binary_path() -> String {
+    let pb: PathBuf = CLI_BINARY_PATH.iter().collect();
+    pb.to_str().unwrap().to_owned()
 }
 
-datatest_stable::harness!(run_all, CLI_TESTSUITE_DIR, r"args.txt$");
+fn get_metatest_path() -> String {
+    let pb: PathBuf = CLI_METATEST_PATH.iter().collect();
+    pb.to_str().unwrap().to_owned()
+}
+
+fn get_testsuite_path() -> String {
+    let pb: PathBuf = CLI_TESTSUITE_PATH.iter().collect();
+    pb.to_str().unwrap().to_owned()
+}
+
+#[test]
+fn run_metatest() {
+    assert!(test::run_all(&get_metatest_path(), &get_cli_binary_path()).is_ok());
+}
+
+#[test]
+fn run_testsuite() {
+    assert!(test::run_all(&get_testsuite_path(), &get_cli_binary_path()).is_ok());
+}
--- a/language/move-lang/src/lib.rs
+++ b/language/move-lang/src/lib.rs
@@ -445,7 +445,7 @@ pub fn find_move_filenames(
 /// - For each directory in `paths`, it will return all files that satisfy the predicate
 /// - Any file explicitly passed in `paths`, it will include that file in the result, regardless
 ///   of the file extension
-fn find_filenames<Predicate: FnMut(&Path) -> bool>(
+pub fn find_filenames<Predicate: FnMut(&Path) -> bool>(
     paths: &[String],
     mut is_file_desired: Predicate,
 ) -> anyhow::Result<Vec<String>> {
--- a/language/tools/move-cli/src/main.rs
+++ b/language/tools/move-cli/src/main.rs
@@ -105,10 +105,9 @@ enum Command {
     /// Run expected value tests using the given batch file
     #[structopt(name = "test")]
     Test {
-        // TODO: generalize this to support running all the tests in a given directory
-        /// File containing batch of commands to run
-        #[structopt(name = "file")]
-        file: String,
+        /// a directory path in which all the tests will be executed
+        #[structopt(name = "path")]
+        path: String,
     },
     /// View Move resources, events files, and modules stored on disk
     #[structopt(name = "view")]
@@ -589,10 +588,7 @@ fn main() -> Result<()> {
             *gas_budget,
             *dry_run,
         ),
-        Command::Test { file } => test::run_one(
-            &Path::new(file),
-            &std::env::current_exe()?.to_string_lossy(),
-        ),
+        Command::Test { path } => test::run_all(path, &std::env::current_exe()?.to_string_lossy()),
         Command::View { file } => view(&move_args, file),
         Command::Clean {} => {
             // delete move_data
--- a/language/tools/move-cli/tests/metatests/args.exp
+++ b/language/tools/move-cli/tests/metatests/args.exp
@@ -1,0 +1,6 @@
+Command `test dummy/test_1/args.txt`:
+1 / 1 test(s) passed.
+Command `test dummy/test_2`:
+1 / 1 test(s) passed.
+Command `test dummy `:
+2 / 2 test(s) passed.
--- a/language/tools/move-cli/tests/metatests/args.txt
+++ b/language/tools/move-cli/tests/metatests/args.txt
@@ -1,0 +1,3 @@
+test dummy/test_1/args.txt
+test dummy/test_2
+test dummy


language/tools/move-cli/src/test.rs
```diff
@@ -10,6 +10,7 @@ use std::{
     path::Path,
     process::Command,
 };
+use walkdir::WalkDir;
```
sblackshear, line 13: It would be nice to avoid taking this dependency on `walkdir` if we can. I think we use `datatest::iterate_directory` for something similar to this [elsewhere](https://github.com/libra/libra/blob/master/language/stdlib/src/main.rs#L79); maybe that will work here?

language/tools/move-cli/tests/cli_testsuite.rs
```diff
@@ -8,8 +8,8 @@ use std::path::Path;
...
 pub const CLI_TESTSUITE_DIR: &str = "tests/testsuite";
 
-fn run_all(args_path: &Path) -> datatest_stable::Result<()> {
-    Ok(test::run_one(args_path, CLI_BINARY)?)
+fn test_run_all(args_path: &Path) -> datatest_stable::Result<()> {
+    Ok(test::run_all(args_path, CLI_BINARY)?)
 }
 
-datatest_stable::harness!(run_all, CLI_TESTSUITE_DIR, r"args.txt$");
+datatest_stable::harness!(test_run_all, CLI_TESTSUITE_DIR, r"args.txt$");
```
sblackshear, line 15: Now that we have `run_all`, I wonder if it would work to kill lines 11--15 and just do something like

```
#[cfg(test)]
fn go() {
  test::run_all(CLI_TESTSUITE_DIR, CLI_BINARY)?
}
```

language/tools/move-cli/src/test.rs
```diff
@@ -135,3 +136,28 @@ pub fn run_one(args_path: &Path, cli_binary: &str) -> anyhow::Result<()> {
...
+pub fn run_all(args_path: &Path, cli_binary: &str) -> anyhow::Result<()> {
+    let mut test_total = 0;
+    let mut test_passed = 0;
+    for entry in WalkDir::new(args_path)
+        .into_iter()
+        .filter_map(Result::ok)
+        .filter(|e| e.file_type().is_file())
+    {
+        let fname = entry.file_name().to_string_lossy();
+        if fname == "args.txt" {
```
sblackshear, line 149: Nit: add this check to the `filter`?

sblackshear (commented):
That was fast!!

language/tools/move-cli/src/test.rs
mengxu-fb: I think we can avoid this`walkdir`. Let me emulate its functionality.

language/tools/move-cli/tests/cli_testsuite.rs
mengxu-fb: Thanks for the suggestion! I'll fix it.

language/tools/move-cli/src/test.rs
mengxu-fb: I thought of that before, but since later we need `fname` in the `eprintln`, I decided to leave this check out of the filter.

sblackshear: Ah, that makes sense. Although: we might want to print the whole path (or at least the parent dir) at 155 given that `fname` will always be "args.txt" (with no prefix).

mengxu-fb: Ah, silly of me. You are right. We should print `entry.path()`. Then this one should go to the filter.

sblackshear (approved):
Looks excellent, thank you!

Would you mind adding two `#[test]`s for the behavior in the summary (i.e., `move test <path/to/args.txt>` and `move test <path/to/single/dir/with/args.txt>`? No need to add new test directories/files; pointing them at existing paths is fine.

Otherwise G2G! Once you've added the tests, comment `/land` on the PR to land it (the GitHub UI will not work).

language/tools/move-cli/tests/cli_testsuite.rs
```diff
@@ -7,9 +7,21 @@ use std::path::Path;
...
-    Ok(test::run_one(args_path, CLI_BINARY)?)
+#[test]
+fn test_run_all() {
+    assert!(test::run_all(Path::new(CLI_TESTSUITE_DIR), CLI_BINARY).is_ok());
 }
 
-datatest_stable::harness!(run_all, CLI_TESTSUITE_DIR, r"args.txt$");
+#[test]
+fn test_run_one_dir() {
+    let path_dir = format!("{}/{}", CLI_TESTSUITE_DIR, CLI_TESTSUITE_SAMPLE);
```
sblackshear, line 19: Nit: using `p = Path::new(CLI_TESTSUITE_DIR); p.join(CLI_TESTSUITE_SAMPLE)` instead will ensure that this works on Windows etc.

mengxu-fb: Shall I change the path for CLI_BINARY and CLI_TESTSUITE_DIR too?

language/tools/move-cli/src/test.rs
```diff
@@ -135,3 +135,24 @@ pub fn run_one(args_path: &Path, cli_binary: &str) -> anyhow::Result<()> {
         Ok(())
     }
 }
+
+pub fn run_all(args_path: &Path, cli_binary: &str) -> anyhow::Result<()> {
+    let mut test_total = 0;
+    let mut test_passed = 0;
+    for entry in datatest_stable::utils::iterate_directory(args_path) {
```
tnowacki, line 142: I'd use `move_lang::find_filenames`. We already indirectly use that logic in the CLI and I think it will make things easier for this

language/tools/move-cli/tests/cli_testsuite.rs
tnowacki: I think we should eventually make a `move_lang::file_utils` or maybe even a separate crate

I have a `path_buf_dir` which makes doing these things a lot easier

language/tools/move-cli/tests/cli_testsuite.rs
```diff
@@ -7,9 +7,21 @@ use std::path::Path;
...
+    assert!(test::run_all(Path::new(&path_dir), CLI_BINARY).is_ok());
+}
+
+#[test]
+fn test_run_one_file() {
+    let path_file = format!("{}/{}/args.txt", CLI_TESTSUITE_DIR, CLI_TESTSUITE_SAMPLE);
```
tnowacki, line 25: > Shall I change the path for CLI_BINARY and CLI_TESTSUITE_DIR too?

We should change this one too, yes :)

language/tools/move-cli/src/test.rs
mengxu-fb: Thanks. I'll make the change.

language/tools/move-cli/tests/cli_testsuite.rs
mengxu-fb: Will do.

language/tools/move-cli/src/test.rs
mengxu-fb: Second thought: `find_filenames` is not `pub` in `lib.rs`, shall I make it `pub`?

tnowacki: > shall I make it pub?

Go for it

language/tools/move-cli/tests/cli_testsuite.rs
tnowacki: (Don't need `path_dir` macro for this PR was more a general comment)

language/tools/move-cli/Cargo.toml
```diff
@@ -12,6 +12,7 @@ edition = "2018"
 [dependencies]
 anyhow = "1.0.31"
 difference = "2.0.0"
+serial_test = "0.5.0"
```
mengxu-fb, line 15: I realized that if we run the tests concurrently, we may see an error like
thread 'test_run_one_file' panicked at '`move clean` failed to eliminate move_data directory', language/tools/move-cli/src/test.rs:114:

My guess is that this is caused by the concurrently running the same test creates some conflicts, so I added the `#[serial]` tags in the tests.

This `serial_test` is used in `network/src/protocols` so is not technically a new introduction to the whole project. But please let me know if this is not preferred.

tnowacki: I don't understand the issue. All of the tests should have their own move_data directories so I'm confused as to why there is any concurrency issue

language/tools/move-cli/src/test.rs
```diff
@@ -135,3 +135,23 @@ pub fn run_one(args_path: &Path, cli_binary: &str) -> anyhow::Result<()> {
...
+        fpath.file_name().expect("unexpected file entry path") == "args.txt"
+    })? {
+        match run_one(Path::new(&entry), cli_binary) {
+            Ok(_) => {
+                test_passed += 1;
+            }
+            Err(ex) => {
+                eprintln!("Test {} failed with error: {}", entry, ex);
+            }
+        }
```
tnowacki, line 152: nit, 
```
match run_one(Path::new(&entry), cli_binary) {
    Ok(_) => test_passed += 1,
    Err(ex) => eprintln!("Test {} failed with error: {}", entry, ex),
}
```

language/tools/move-cli/Cargo.toml
sblackshear: I think the problem is that I suggested that we reuse dirs tested in `run_all` in `test_run_one_dir` and `test_run_one_file`. I think we should probably either delete these tests (I suggested them but am not wedded to them) or create separate directories for them that are outside of the paths covered by `test_run_all`.

tnowacki (dismissed):
Let's figure out this concurrency thing before it goes in :)

language/tools/move-cli/Cargo.toml
sblackshear: (i.e., the race is `run_all` || `test_run_one_dir`)

mengxu-fb: Sam is right, it is actually three way: `run_all` || `test_run_one_dir` || `test_run_one_file` since that particular file also resides in that dir.

tnowacki: Ah! I see... but am torn about what to do. 

I'd probably say its best to have separate directories. Otherwise, we basically are rerunning all of the tests 4 separate times on each CI/`cargo test` run, which seems overly paranoid. 

If I had to pick, I'd say either
- Make seprate files/dirs for these 3 functions
- kill all 3: `test_run_all`, `test_run_one_dir`, and `test_run_one_file` and just add tests that call out to the `test` CLI command

mengxu-fb: Per each CI/`cargo test` it should be running all tests in `testsuite` plus 2 extra, the one for `run_one_dir` and the one for `run_one_file`.

tnowacki: That still seems crazy. Especially as these tests grow larger... Anything over just one time seems bad

mengxu-fb: Right. So shall I just create a few mock dirs under, say `move-cli/tests/unit-test` to cover the three cases?

tnowacki: You could... 

I think it would be fine to just use the existing test system and not have these three functions at all. 

So you would add an `args.txt`  that would have
```
test --path <path to subdir>
```

sblackshear: That sounds good! I don't lock `mocktests` but can't think of a better name (I thought of `metatests` because they're testing the testing functionality, but that is maybe confusing)...

mengxu-fb: Question: the approach of adding an `args.txt` with `test --path <path to subdir>` will not be scheduled by `cargo test`? We still need sth in `cli_testsuite.rs` for `cargo` to run the test

language/tools/move-cli/src/test.rs
```diff
@@ -135,3 +135,19 @@ pub fn run_one(args_path: &Path, cli_binary: &str) -> anyhow::Result<()> {
...
+    for entry in move_lang::find_filenames(&[args_path.to_owned()], |fpath| {
+        fpath.file_name().expect("unexpected file entry path") == "args.txt"
+    })? {
+        match run_one(Path::new(&entry), cli_binary) {
+            Ok(_) => test_passed += 1,
+            Err(ex) => eprintln!("Test {} failed with error: {}", entry, ex),
+        }
+        test_total += 1;
+    }
+    println!("{} / {} tests passed.", test_total, test_passed);
```
sblackshear, line 151:
```suggestion
    println!("{} / {} test(s) passed.", test_total, test_passed);
```

mengxu-fb: Updated this and the expected test results too :)

language/tools/move-cli/Cargo.toml
sblackshear: Correct. I was thinking that the `test_run_one_dir` and `test_run_one_file` #[test]` functions could live in `cli_testsuite.rs` (similar to what you had before). That will let `cargo test` pick them up.

mengxu-fb: I added an `args.txt` in the new `metatests` folder which outlines effectively the three cases
```
test dummy/test_1/args.txt
test dummy/test_2
test dummy
```
And left a pointer in `cli_testsuite.rs` to invoke the `args.txt`.

language/tools/move-cli/tests/cli_testsuite.rs
```diff
@@ -3,13 +3,22 @@
...
 }
 
-datatest_stable::harness!(run_all, CLI_TESTSUITE_DIR, r"args.txt$");
+fn get_metatest_path() -> String {
+    let pb: PathBuf = CLI_METATEST_PATH.iter().collect();
+    pb.to_str().unwrap().to_owned()
+}
+
+#[test]
+fn run_test() {
```
sblackshear, line 22: This looks perfect. But don't we need another `#test` for cargo to run `tests/testsuite`? You can try changing a `.exp` file under `tests/testsuite` and running `cargo xtest` to see if it's running those tests.

mengxu-fb: Ah, I think I might had a misunderstanding from the above discussion that the whole `test/testsuite` should not be involved in the CI process. In this case, I can add another `#test` for the `testsuite`, should be easy.

sblackshear:
Thanks! Landing this now.

PR #6280, opened 2020-09-30
[json-rpc] add more tests

xli:
1. added more corner cases test cases
2. added tests for all different types event data views
3. removed a wrapper when parsing event, the original version returns "unknown" event if parsing failed, which hides real error, changed to return server internal error to expose problem.

Still missing EventDataView::UpgradeEvent test, don't know how to generate it, will need talk to @runtian-zhou about it and add test later.

New event CreateAccountEvent is added, it is created recently (https://github.com/libra/libra/pull/6156).
Added new test to ensure there is no unknown events in integration tests at the end, to make test pass, created blank CreateAccountEvent for now (tried to decoded it, but got an error, so try to keep this diff simpler, will talk to @sblackshear  to add created account address field if necessary)

### DIFF
--- a/json-rpc/types/src/views.rs
+++ b/json-rpc/types/src/views.rs
@@ -194,7 +194,7 @@ pub enum EventDataView {
     #[serde(rename = "createaccount")]
     CreateAccount {},
     #[serde(rename = "unknown")]
-    Unknown { module_name: String },
+    Unknown {},
 }
 
 impl TryFrom<ContractEvent> for EventDataView {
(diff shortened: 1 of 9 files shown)


libra-action:

Breaking changes policy:

2. Adding fields is ok.

json-rpc/types/src/views.rs
```diff
@@ -188,174 +191,148 @@ pub enum EventDataView {
...
+
+impl TryFrom<(u64, ContractEvent)> for EventView {
+    type Error = Error;
+
+    fn try_from((txn_version, event): (u64, ContractEvent)) -> Result<Self> {
+        Ok(EventView {
             key: BytesView::from(event.key().as_bytes()),
             sequence_number: event.sequence_number(),
             transaction_version: txn_version,
-            data: event_data.unwrap_or(EventDataView::Unknown {}),
```
matbd, line 357: indeed

matbd (approved):
LGTM

Issue #6287, opened 2020-09-30
[logging] Consider adding hostname to structured logs


gregnazario:
Requested by @mgorven

PR #6291, opened 2020-09-30
logger: include hostname in log events

bmwill:
Closes: #6287

### DIFF
--- a/common/logger/src/libra_logger.rs
+++ b/common/logger/src/libra_logger.rs
@@ -11,6 +11,7 @@ use crate::{
     Event, Filter, Level, Metadata,
 };
 use chrono::{SecondsFormat, Utc};
+use once_cell::sync::Lazy;
 use serde::Serialize;
 use std::{
     collections::BTreeMap,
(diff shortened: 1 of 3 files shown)


common/logger/src/libra_logger.rs
```diff
@@ -67,6 +69,9 @@ impl LogEntry {
         let metadata = *event.metadata();
         let thread_name = thread_name.map(ToOwned::to_owned);
         let message = event.message().map(fmt::format);
+        let hostname = hostname::get()
```
mgorven, line 72: You could potentially cache this at startup since it shouldn't change, but I can't think of any problems calling this every time.

mgorven (approved):

common/logger/src/libra_logger.rs
bmwill: That's a good idea, let me do that before landing

common/logger/src/libra_logger.rs
```diff
@@ -68,6 +71,14 @@ impl LogEntry {
         let thread_name = thread_name.map(ToOwned::to_owned);
         let message = event.message().map(fmt::format);
 
+        static HOSTNAME: Lazy<Option<String>> = Lazy::new(|| {
+            hostname::get()
+                .ok()
+                .and_then(|name| name.into_string().ok())
+        });
+
+        let hostname = HOSTNAME.clone();
```
gregnazario, line 80: Nit: Static variables with all caps names, especially since it is Lazy initialized should probably go outside of the function.

gregnazario (approved):

common/logger/src/libra_logger.rs
bmwill: Generally I would agree with you but this is something that is scoped to this specific function and not intended for use outside similar to the visitor struct definition in the same function.

PR #6292, opened 2020-09-30
[vm] fix off-by-1 error in PC reported in aborting instructions

sblackshear:
The bytecode interpreter increments the PC before executing an instruction. This means that if the instruction at bytecode offset `i` aborts, the PC location associated with the `ExecutionFailure` is `i + 1` (off-by-1).

This PR fixes the issue by moving the increment to the end of the interpreter loop and adding PC increments when:
- pushing a new frame on when returning from a call
- returning from a native function call

This enforces the following invariant: the PC advances to instruction `i+1` iff instruction `i` executed without aborting.

### DIFF
--- a/language/tools/move-cli/tests/testsuite/explain_arithmetic_failure/args.exp
+++ b/language/tools/move-cli/tests/testsuite/explain_arithmetic_failure/args.exp
@@ -1,2 +1,2 @@
 Command `run script.move`:
-Execution failed because of an arithmetic error (i.e., integer overflow/underflow, div/mod by zero, or invalid shift) in script at code offset 3
+Execution failed because of an arithmetic error (i.e., integer overflow/underflow, div/mod by zero, or invalid shift) in script at code offset 2
--- a/language/ir-testsuite/tests/testing_infra/show_all_optional_info.exp
+++ b/language/ir-testsuite/tests/testing_infra/show_all_optional_info.exp
@@ -5,8 +5,8 @@
 [4] Write Set: WriteSet(WriteSetMut { write_set: [(AccessPath { address: a4a46d1b1421502568a4a6ac326d7250, path: 0185aa1f79e25ca70b8f54f8a0b673259fc19b1af35acb98885d99e160600387c9 }, Value(20975741f6b37b591141e96e0c60a57ce8a4a46d1b1421502568a4a6ac326d725001a4a46d1b1421502568a4a6ac326d725001a4a46d1b1421502568a4a6ac326d72500000000000000000180000000000000000a4a46d1b1421502568a4a6ac326d72500000000000000000180000000000000000a4a46d1b1421502568a4a6ac326d72500100000000000000))] })
 [5] Events: []
 [6] Transaction(1)
-[7] Move VM Status: EXECUTION_FAILURE { status_code: ARITHMETIC_ERROR, location: Script, function_definition: 0, code_offset: 3 }
-[8] Transaction Status: Keep(EXECUTION_FAILURE { location: Script, function_definition: 0, code_offset: 3 })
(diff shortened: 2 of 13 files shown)


tnowacki (commented):
Things look good (I think) from the errors side of things. I probably should have checked these offsets more closely when adding this feature. 

Leaving for someone else (probably @vgao1996 or @dariorussi) to accept :)

dariorussi (approved):
this looks fine, I still need to check that the dump on error is not adversely affected but I am doing that as part of the fault injection so I am thinking we could get this in and it would make it even easier for me to check that. But just to make sure we have extra-eyes on what I am talking about, please take a look at maybe_core_dump and the function called that builds the call stack and see if you can spot anything.
In that respect also take a look at the debug stuff (print stack trace) and see if you can spot something there too. 
Those functions I believe use something like `[0..pc]` which is exclusive on the right side and as such may be missing one element now.
Could you please take a look at that too?

dariorussi:
I suppose it would be any usage of `pc` in a slice, thanks!!

vgao1996:
Also looks good to me!

language/tools/move-cli/tests/testsuite/explain_arithmetic_failure/args.exp
```diff
@@ -1,2 +1,10 @@
 Command `run script.move`:
+<<<<<<< HEAD
```
dariorussi, line 2: this cannot be right is it? 😄

dariorussi:
thinking about this some more, I am feeling a better fix would have been to call `pc` `next_inst` or something (just to make it more obvious), leave everything as it was and fix the error reporting.
Anyway, whatever we all want to do at this point

language/tools/move-cli/tests/testsuite/explain_arithmetic_failure/args.exp
sblackshear: The CLI tests were broken but have now been fixed, I need to rebase :).

sblackshear:
> thinking about this some more, I am feeling a better fix would have been to call `pc` `next_inst` or something (just to make it more obvious), leave everything as it was and fix the error reporting.
> Anyway, whatever we all want to do at this point

I think that would definitely be less invasive, but it's harder (at least for me) to understand the invariants on PC is in that case. For example: "`next_inst` always points to the next instruction to be executed" seems like an invariant you'd want, but it doesn't hold because `Branch` instructions overwrite it. In addition, "`next_inst` - 1 always points to the currently executing instruction" (which is the invariant you'd want to do the fixup in error reporting) doesn't hold for the same reason.

dariorussi:
> I think that would definitely be less invasive, but it's harder (at least for me) to understand the invariants on PC is in that case. For example: "`next_inst` always points to the next instruction to be executed" seems like an invariant you'd want, but it doesn't hold because `Branch` instructions overwrite it. In addition, "`next_inst` - 1 always points to the currently executing instruction" (which is the invariant you'd want to do the fixup in error reporting) doesn't hold for the same reason.

in some ideal and conceptual way yes that is true, and as discussed in depth offline that idea of changing the code to have a pointer to the instruction in question is a good long term idea. I think we could review the code at some point and make it so we don't have to increment that pointer in multiple place if possible.
The branch is definitely a bit of a strange case though it is also one that does not really matter right now.
So don't know what to say, except the fact that this is more of a change that I would care to have right now

sblackshear:
>  think we could review the code at some point and make it so we don't have to increment that pointer in multiple place if possible.

Agreed that this is the right long-term path. I think we can probably do it by combining `execute_main` and `execute_code_impl` (or via some other scheme), but that's for later.

> So don't know what to say, except the fact that this is more of a change that I would care to have right now

Fair point. The only thing that matters to me now is having the correct location in aborts; the "strong invariants on PC" stuff is unrelated and can wait. I'm happy to do your suggested fix instead if you prefer.

PR #6295, opened 2020-09-30
[cluster-test] Add load test framework which introduce ability to con…

snakech:
…struct load test with different components

## Motivation
Version 1 for load test framework:
Add load test framework which introduce ability to construct load test with different components
Currently we have interface for tx_emitter, mempool and state_sync

Phase one: Added interface and framework
Phase two: Refactory current load test with new framework

scope for each component owner: display, metrics and any features are needed

(Write your answer here.)

## Test Plan

## Related PRs

### DIFF
--- a/testsuite/cluster-test/src/load_test/mod.rs
+++ b/testsuite/cluster-test/src/load_test/mod.rs
@@ -1,0 +1,40 @@
+// Copyright (c) The Libra Core Contributors
+// SPDX-License-Identifier: Apache-2.0
+
+#![forbid(unsafe_code)]
+
+use anyhow::Result;
+use async_trait::async_trait;
+use futures::future::try_join_all;
+use std::fmt::Display;
+use std::time::Duration;
+
+mod mempool;
+mod state_sync;
+mod tx_emitter;
+
+#[async_trait]
+pub trait Handler: Display + Send {
+    async fn start(&mut self, duration: Duration) -> Result<()>;
+    async fn stop(&mut self) -> Result<()>;
+}
+
+pub struct HandlerRunner {
+    handlers: Vec<Box<dyn Handler>>,
+}
+
+impl HandlerRunner {
+    pub fn new(handlers: Vec<Box<dyn Handler>>) -> Self {
+        Self { handlers }
+    }
+
+    pub async fn start_all<T: Handler>(handler: &mut Vec<T>, duration: Duration) -> Result<()> {
+        try_join_all(handler.iter_mut().map(|h| h.start(duration))).await?;
+        Ok(())
+    }
+
+    pub async fn stop_all<T: Handler>(handler: &mut Vec<T>) -> Result<()> {
+        try_join_all(handler.iter_mut().map(Handler::stop)).await?;
+        Ok(())
+    }
+}
--- a/testsuite/cluster-test/src/lib.rs
+++ b/testsuite/cluster-test/src/lib.rs
@@ -12,6 +12,7 @@ pub mod genesis_helper;
 pub mod github;
 pub mod health;
 pub mod instance;
+pub mod load_test;
 pub mod prometheus;
 pub mod report;
 pub mod slack;
--- a/testsuite/cluster-test/src/load_test/mempool.rs
+++ b/testsuite/cluster-test/src/load_test/mempool.rs
@@ -1,0 +1,87 @@
+// Copyright (c) The Libra Core Contributors
+// SPDX-License-Identifier: Apache-2.0
+
+#![forbid(unsafe_code)]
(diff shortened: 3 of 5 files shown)


testsuite/cluster-test/src/load_test/mod.rs
```diff
@@ -0,0 +1,40 @@
...
+use std::fmt::Display;
+use std::time::Duration;
+
+mod mempool;
+mod state_sync;
+mod tx_emitter;
+
+#[async_trait]
+pub trait Handler: Display + Send {
+    async fn start(&mut self, duration: Duration) -> Result<()>;
```
zekun000, line 18: the interface seems too general, what I had in mind is just
```
enum {
  JSON_RPC,
  LibraNet,
}
trait LoadTest: {
  type Load;
  fn endpoint() -> Endpoint; 
  fn next_load() -> Self::Load;
}
```
and implement the general handling like
```
fn run(load_test: Box<dyn LoadTest>) {
  while let Some(load) = load_test.generate_load() {
    endpoint.submit(load);
  }
}
```

andll:
It would be good to understand how proposed framework simplify things

I am against just producing yet another trait and trying to fit existing tx emitter into it just for abstract generalization

E.g. what we can do - if we have multiple experiments with similar patterns, demonstrate how they become shorter / reuse code with proposed framework

snakech:
> It would be good to understand how proposed framework simplify things
> 
> I am against just producing yet another trait and trying to fit existing tx emitter into it just for abstract generalization
> 
> E.g. what we can do - if we have multiple experiments with similar patterns, demonstrate how they become shorter / reuse code with proposed framework

Yes, I agree with u, from CT side, we can only supply common and general apis/interfaces. This version 1 is mostly based on abstract common procedure from mempool, state_sync and json-rpc.

I think once we have more user cases(components) we can make extension in future, but I think we should avoid over designing at this stage.

zekun000:
as I commented above, the current trait is not very useful, the common part I think worth pulled out is 
```
fn run(load_test: Box<dyn LoadTest>) {
  while let Some(load) = load_test.generate_load() {
    endpoint.submit(load);
  }
}
```
but I don't play with the code and not sure if it's possible.

testsuite/cluster-test/src/load_test/mod.rs
snakech: I like the enum and load test type, but my thinking is for each type, we leave a handler and CT only needs care about handler start and end within an input duration.

for example user can pass in multi flags stand for each component which they wanna include in load test. And test param will build up a vec of handlers. We give flexibility  of each component load generation part to user since CT side does not have full sight of how does each component actually generate load, do they have common points to abstract. So this is why I called it version 1, which it looks a little general, but we can extend or refactory easier

snakech:
> as I commented above, the current trait is not very useful, the common part I think worth pulled out is
> 
> ```
> fn run(load_test: Box<dyn LoadTest>) {
>   while let Some(load) = load_test.generate_load() {
>     endpoint.submit(load);
>   }
> }
> ```
> 
> but I don't play with the code and not sure if it's possible.

technically this is possible for mempool and state_sync but not worth for json-rpc which it just simply run tx_emitter and we already have full set of apis

zekun000:
I can dig more into the code, but imo TxEmitter should only define what txns to generate, not how to send the txns. The endpoint we have today is the Instance::json_rpc_client() which can share across all json rpc load test. And we can have similar endpoint as Instance::libra_net_client().

snakech:
> libra_net_client

I think current load test @christinacelee generate tx among fullnodes with standard ```EmitJobRequest::for_instances```, please correct me if I am wrong

andll:
@zekun000 to be honest that generalization does not make much sense to me

90% of tx_emitter code is related to transaction generation and accounts preparation. Very small portion is related to what appear to be generic in proposed load test framework.
I don't think it makes sense to reorganize tx_emitter just to fit certain pattern of other load tests.

If we think that other future load tests have some common pattern, lets write them first, and then we can generalize it.

zekun000:
fine, without that part, there's not much value for this abstraction, basically the remaining work is just to add some metrics for the existing ones and add some json-rpc read tests.

snakech:
> fine, without that part, there's not much value for this abstraction, basically the remaining work is just to add some metrics for the existing ones and add some json-rpc read tests.

ok, I will stash this change once we have more user cases, and more abstractions is needed, we can do the refactory at that time.

bmwill:
Closing this so we can remove the branch in the libra repo. If you want to reopen this in the future you should first push this to a branch in your own repo and open a PR from there.

PR #6296, opened 2020-09-30
Upper case hex for address display/debug

tnowacki:
- Changed display of address to upper hex

## Motivation

- We canonically use upper case hex in Move programs
- I will drop this PR if we use display/debug in anywhere critical

## Test Plan

- updated source lang/cli tests


sblackshear:
I like this *much* better aesthetically but am unsure if it will have downstream effects on the CLI/json-rpc/elsewhere. CC @xli @kchalkias @davidiw to see if there are any anticipated bad effects

tnowacki:
The spots for access paths seem to call out to lower hex directly, via `{:#x}`, so we should be in the clear there? 

Definitely ready to drop this quickly if it turns out to be too entrenched.

language/move-lang/src/shared/mod.rs
```diff
@@ -72,19 +72,19 @@ impl AsRef<[u8]> for Address {
...
 
 impl fmt::Debug for Address {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        write!(f, "0x{:#x}", self)
+        write!(f, "0x{:#X}", self)
     }
 }
 
-impl fmt::LowerHex for Address {
+impl fmt::UpperHex for Address {
```
sblackshear, line 85: Does implementing `UpperHex` give you `LowerHex` for free? And if not, can we keep the `LowerHex` impl?

sblackshear:
> The spots for access paths seem to call out to lower hex directly, via `{:#x}`, so we should be in the clear there?

I really hope the hashing for access paths uses the raw byes of the address instead of a stringified version...

language/move-lang/src/shared/mod.rs
tnowacki: I'll keep both, yeah

tnowacki:
> > The spots for access paths seem to call out to lower hex directly, via `{:#x}`, so we should be in the clear there?
> 
> I really hope the hashing for access paths uses the raw byes of the address instead of a stringified version...

It probably does? No idea, its just the one spot I right away saw a build error when I removed the lower hex trait impl

xli:
I believe it is only used in test, as long as all tests pass, should be fine, although probably will have some client sdk tests broken as comparing with lower cases.

tnowacki:
Found a bunch of weird stuff in the faucet tests where they used string representations of account addresses. After fixing that it looks like everything is running

sblackshear (approved):
The stronger typing in the faucet looks like an improvement to me.

dmitri-perelman:
As an anecdote, AOS had some (unexpected and non-planned) dependency on address being lower-case... (the HSMs would typically use the addresses written in the approval policies that come with addresses written in in lower case, but during the initialization we'd use `AccountAddress::to_string()`, which changed to be upper case after the display changed). 
Obviously, an AOS bug, we should have not depended on that, but it was fun to investigate :)

metajack:
This broke our Rosetta implementation as it didn't change the output everywhere it appears.

I'll note that Coinbase considers addresses case sensitive due to [Ethereum's checksumming](https://coincodex.com/article/2078/ethereum-address-checksum-explained/), which is some support for us moving to all lowercase as canonical in case we ever want to do similar checksumming.

PR #6298, opened 2020-09-30
[vm] Fix gauge to be a histogram for block transaction count

gregnazario:
The metric was a gauge, which causes it to be sampled and have strange behavior.

Options are a histogram, or a counter.  I set it as a histogram, since that's what the comment was describing it as.

### DIFF
--- a/language/libra-vm/src/libra_transaction_executor.rs
+++ b/language/libra-vm/src/libra_transaction_executor.rs
@@ -40,7 +40,7 @@ use move_vm_types::{
 use rayon::prelude::*;
 use std::{
     collections::HashSet,
-    convert::{AsMut, AsRef, TryFrom},
+    convert::{AsMut, AsRef},
 };
 
 pub struct LibraVM(LibraVMImpl);
@@ -670,10 +670,7 @@ impl LibraVM {
         }
 
         // Record the histogram count for transactions per block.
-        match i64::try_from(count) {
-            Ok(val) => BLOCK_TRANSACTION_COUNT.set(val),
-            Err(_) => BLOCK_TRANSACTION_COUNT.set(std::i64::MAX),
-        }
+        BLOCK_TRANSACTION_COUNT.observe(count as f64);
 
         Ok(result)
     }
--- a/language/libra-vm/src/counters.rs
+++ b/language/libra-vm/src/counters.rs
@@ -2,8 +2,8 @@
 // SPDX-License-Identifier: Apache-2.0
 
 use libra_metrics::{
-    register_histogram, register_int_counter, register_int_counter_vec, register_int_gauge,
-    Histogram, IntCounter, IntCounterVec, IntGauge,
+    register_histogram, register_int_counter, register_int_counter_vec, Histogram, IntCounter,
+    IntCounterVec,
 };
 use once_cell::sync::Lazy;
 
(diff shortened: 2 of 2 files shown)


davidiw:
should it be a histogram or a counter? where are the histogram buckets defined? https://docs.rs/prometheus/0.10.0/prometheus/struct.HistogramOpts.html#structfield.buckets
you likely don't want the default buckets

gregnazario:
> should it be a histogram or a counter? where are the histogram buckets defined? https://docs.rs/prometheus/0.10.0/prometheus/struct.HistogramOpts.html#structfield.buckets
> you likely don't want the default buckets

These are the defined default buckets

https://docs.rs/prometheus/0.10.0/src/prometheus/histogram.rs.html#21-27

bob-wilson:
I was already testing the same change.... You guys are too quick for me

bob-wilson:
I'm still looking at this, mostly because I'm concerned that all the other histograms in Libra are using the default buckets, and that unless I'm missing something, that doesn't seem like it will work well

gregnazario:
> I'm still looking at this, mostly because I'm concerned that all the other histograms in Libra are using the default buckets, and that unless I'm missing something, that doesn't seem like it will work well

I didn't do a lot of looking into histograms as a whole, so perhaps we should look into changing the buckets across the board.  The default buckets really seem like they only work well for time :)

bob-wilson:
> The default buckets really seem like they only work well for time :)

Even that part doesn't make sense to me.... E.G., if I look at the libra_vm_txn_validation_seconds histogram, the premainnet dashboard shows a bunch of average times around 400us. How is that even possible if the smallest bucket is 5ms? I suppose it's possible that anything less than 5ms is going into a zero bucket and then there are a few that take longer than 5ms so the average works out to be in the range of 400us?

gregnazario:
> > The default buckets really seem like they only work well for time :)
> 
> Even that part doesn't make sense to me.... E.G., if I look at the libra_vm_txn_validation_seconds histogram, the premainnet dashboard shows a bunch of average times around 400us. How is that even possible if the smallest bucket is 5ms? I suppose it's possible that anything less than 5ms is going into a zero bucket and then there are a few that take longer than 5ms so the average works out to be in the range of 400us?

Hmm yeah you're right.  This is a little annoying to have to predefine the buckets, but we should come up with buckets that work for different scenarios as best we can.

bob-wilson:
I remember looking at this when I set up metrics for the VM but it has been a while and I had to go back and re-read the docs. The key is that in addition to the bucket counts, histograms also record the sum as an f64 value. If you're only using the histogram to display the average values, the buckets don't matter -- they're only needed for quantile info. So, tl;dr: we might still want to adjust the buckets for various metrics, but it's not essential if you don't care about the quantiles.

gregnazario:
Added @mimoo and @davidiw as per the dependency notifier

language/libra-vm/src/libra_transaction_executor.rs
```diff
@@ -617,9 +618,8 @@ impl LibraVM {
         }
 
         // Record the histogram count for transactions per block.
-        match i64::try_from(count) {
-            Ok(val) => BLOCK_TRANSACTION_COUNT.set(val),
-            Err(_) => BLOCK_TRANSACTION_COUNT.set(std::i64::MAX),
+        if let Some(val) = f64::from_usize(count) {
+            BLOCK_TRANSACTION_COUNT.observe(val)
```
bmwill, line 622: I would just avoid the extra dependency and just do an `as` conversion:

`BLOCK_TRANSACTION_COUNT.observe(count as f64)`

This conversion wont ever fail, but will loose precision in the process (but IMO that's fine for something like this)

gregnazario: Sounds fine to me

PR #6299, opened 2020-09-30
[json-rpc] improve metrics: tag metrics with consistent label name

xli:
1. tag metrics with consistent label name
2. removed unused parse error
3. name metrics more consistent: add rpc prefix for the case it's rpc endpoint level, others are method level


libra-action:

Breaking changes policy:

2. Adding fields is ok.

thefallentree (approved):

PR #6301, opened 2020-10-01
[breaking][global refactor]  remove "Coin2", stop minting of LBR, make Coin1 default

tzakian:
This PR removes `Coin2` and stops minting of LBR. It also changes the "default" currency to "Coin1".


client/faucet/src/mint.rs
```diff
@@ -134,7 +134,7 @@ impl Service {
             seq,
             1_000_000,
             0,
-            LBR_NAME.to_owned(),
+            TMPCOIN_NAME.to_owned(),
```
xli, line 137: why we change LBR to TMPCOIN?

dmitri-perelman:
This is obviously quite a breaking change for AOS and the DDs :).
Why do we change `Coin1` to `TMPCOIN`, if later on we're going to change `TMPCOIN` to something else? Why not to leave `Coin1` for now in order to reduce the number of changes?

tzakian:
The changes to LBR in the modules is a rough over-approximation right now, I'll be updating to add back in more logic and verification for it that existed previously.

client/faucet/src/mint.rs
tzakian: Since minting of LBR is no longer permitted allowed after this PR the "default coin" is switching from LBR to TMPCOIN

wrwg (approved):
Less move changes than expect, still quite a slaughter. LGTM from spec view; notice wrong renaming in spec-lang.md.

language/move-lang/functional-tests/tests/libra/libra/stop_minting.move
```diff
@@ -1,25 +1,88 @@
...
+//! new-transaction
+//! sender: libraroot
+//! execute-as: blessed
+script {
+use 0x1::TransactionFee;
+use 0x1::COIN::{Self, COIN};
+fun main(lr_account: &signer, tc_account: &signer) {
+    COIN::initialize(lr_account, tc_account);
+    TransactionFee::add_txn_fee_currency<COIN>(tc_account);
+}
```
moeziniaf, line 60: nice

language/move-lang/functional-tests/tests/libra/libra_account/add_currency.move
```diff
@@ -1,14 +1,79 @@
+//! account: validator, 1000000, 0, validator
 //! account: vasp, 0,0, address
 //! account: child, 0,0, address
+//! account: vivian, 0, 0, address
+//! account: otto, 0, 0, address
+
+
+// BEGIN: registration of a currency
```
moeziniaf, line 8: :( If only there was a way to invoke some test helper to avoid replicating this code

moeziniaf (approved):

language/move-lang/functional-tests/tests/libra/libra_account/add_currency.move
tzakian: I wish too 😢  But! I think in the new and improved testing framework that Victor is building this should be possible :)

PR #6304, opened 2020-10-01
[move cli] add verbose arg

sblackshear:
Printing was getting a bit out of hand. This suppresses all printing except for `move view` and error messages by default. Using the `-v` flag brings it all back.

### DIFF
--- a/language/tools/move-cli/src/main.rs
+++ b/language/tools/move-cli/src/main.rs
@@ -32,16 +32,19 @@ use structopt::StructOpt;
 #[structopt(name = "Move", about = "CLI frontend for Move compiler and VM")]
 struct Move {
     /// Directory storing Move resources, events, and module bytecodes produced by script execution.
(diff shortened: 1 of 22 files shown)


language/tools/move-cli/src/main.rs
```diff
@@ -32,16 +32,19 @@ use structopt::StructOpt;
 #[structopt(name = "Move", about = "CLI frontend for Move compiler and VM")]
 struct Move {
     /// Directory storing Move resources, events, and module bytecodes produced by script execution.
-    #[structopt(name = "move-data", long = "move-data", default_value = MOVE_DATA)]
+    #[structopt(name = "move-data", long = "move-data", default_value = MOVE_DATA, global = true)]
```
sblackshear, line 35: This nifty `global` option allows you to use the flag either before or after the subcommand (e.g., `move -v check` or `move check -v`). Without the flag, only the first one works. I added this for all the generic flags because it's pretty annoying without it.

tnowacki (approved):
Cool!

PR #6305, opened 2020-10-01
[network] purge deprecated gossip discovery protocol

phlip9:
The first diff is made in preparation for removing the gossip discovery protocol from network.

Previously, this state sync integration test framework would add the previously generated peer as a seed peer and let gossip discovery do the rest of the work to learn all other peer addresses and then connect them. With the first diff, we "pre-bind" the `/memory/<port>` addresses and just build the seed peer set from the beginning.

On a slightly tangential note, it seems strange to me that this framework puts all peers (both validators and full nodes) on the same network (when it's not mocking the network interface), which doesn't really reflect reality. Something to change later I suppose.

The second diff purges all gossip discovery protocol code, removes gossip options from the network config, and fixes some random tests in connectivity_manager that used `DiscoverySource::Gossip`.


gregnazario (commented):
I know we used Gossip discovery for testnets or cluster test, but I guess that was through config builder?

phlip9:
> I know we used Gossip discovery for testnets or cluster test, but I guess that was through config builder?

Yep, config builder's gone, no more gossip discovery dependents : p

phlip9:
passes cluster test no problem

gregnazario (approved):

davidiw (commented):
good stuff, good stuff

PR #6313, opened 2020-10-01
[xtest] no fail fast, and lcov support

rexhoffman:
## Motivation

replace code_coverage.sh script which takes a very long time due to massive recompilation, with a call to xtest.
Seems to save about 2-3 hours of computation.
Capture failed tests and send them to slack.

Fixed up xtest to not fail fast, but instead still generate coverage reports if the xtest run ends in errors.
xtest now supports lcov as well with --html-lcov-dir.
the lcov.info file will get generated to that dir as well as the html report based on lcov.

Yes

## Test Plan

Run in a test-1 branch with the once a day gating logic turned off....
Here's a coverage run:
https://app.circleci.com/pipelines/github/libra/libra/32146/workflows/570bd495-9afb-4eb8-b64d-d1d985580e08/jobs/227368
Here's my latest test run with the test reporting baked in, and running the test-suite as well.  
https://app.circleci.com/pipelines/github/libra/libra/32177/workflows/1aef8d46-623e-4f9f-9e40-2b3888418de7/jobs/227680
Need to check slack when it's done to verify the message, though I tested the shell locally with a different, less time consuming command.

## Related PRs

None.


devtools/x/src/test.rs
```diff
@@ -92,49 +100,136 @@ pub fn run(mut args: Args, xctx: XContext) -> Result<()> {
...
+        let html_cov_path = &html_cov_dir.canonicalize()?;
+        info!("created {}", &html_cov_path.to_string_lossy());
+        exec_grcov(&html_cov_path)?;
+    }
+    if let Some(html_lcov_dir) = &args.html_lcov_dir {
+        create_dir_all(&html_lcov_dir)?;
+        let html_lcov_path = &html_lcov_dir.canonicalize()?;
+        info!("created {}", &html_lcov_path.to_string_lossy());
+        exec_lcov(&html_lcov_path)?;
+        exec_lcov_genhtml(&html_lcov_path)?;
```
sausagee, line 133: Not sure if other people would get confused in which report to look at.  Maybe the grcov report is enough?  Will leave this you.

sausagee (commented):
Looks good.

Will stamp but want to give @bob-wilson heads up on the coverage script firstk.

bob-wilson (commented):
As @sausagee suggests, it isn't clear when one should use lcov vs. grcov. If we're going to support both, it would be good to have some guidance, probably in the xtest --help message. I know I've tried both at different times, but I don't remember the differences.

There is also a reference to the coverage_report script in language/move-prover/tests/README.md. Can you please update that?

rexhoffman:
So a full 4 h 58 min run succeed today.
https://app.circleci.com/pipelines/github/libra/libra/32306/workflows/aeff4077-c2fa-46ad-b33f-9e29148c4c43/jobs/228889
Coverage reported at
Overall coverage rate:
  lines......: 80.4% (98013 of 121855 lines)
  functions..: 16.6% (59851 of 359478 functions)
This run in 2 hrs 4min
https://app.circleci.com/pipelines/github/libra/libra/32341/workflows/4e38ae2d-ff4f-485e-96ef-f06390598c00/jobs/229198
Overall coverage rate:
  lines......: 77.7% (99606 of 128206 lines)
  functions..: 34.4% (37750 of 109625 functions)

Compiling crates individually vs compiling them together.  Larger code coverage difference than I'd expect, but looks like the number of functions generated 3.5 times less.

Updated movelang doc.

rexhoffman:
also the use of piping to a tee breaks in circleci and hangs indefinitely.   Good times.

sausagee (approved):
Remember to uncomment out the block before land.

bmwill:
> also the use of piping to a tee breaks in circleci and hangs indefinitely. Good times.

I'm glad that you were able to figure out what the root cause of that hang was, and i'm glad it wasn't actually anything in our tooling (e.g. `x`)

PR #6314, opened 2020-10-01
[libra framework][docgen] Cleanup AccountFreezing and AccountLimits.

wrwg:
This PR cleans up documentation of mentioned modules. It also fixes specification incompleteness in AccountLimits and re-activates an invariant there. The incompleteness stemmed from turning some specs opaque but not determining all fields of a struct in post conditions.

Moreover, an issue in the doc generator with wrong indentation of `spec define` style helper functions is fixed.

Also, we add a new option `references_file` to the generator which allows to add a file to each generated output containing reference definitions. This is used for transaction script and module documentation; the file is at `stdlib/modules/references_template.md`.

This is a first PR in a series; thought better split this down instead of one big PR to avoid merge conflicts.

## Motivation

Cleanup

Yes

## Test Plan

NA

## Related PRs

NA

### DIFF
--- a/language/move-prover/docgen/src/docgen.rs
+++ b/language/move-prover/docgen/src/docgen.rs
@@ -77,6 +77,7 @@ const WEAK_KEYWORDS: &[&str] = &[
     "include",
     "mut",
     "old",
+    "pragma",
     "pack",
     "unpack",
     "update",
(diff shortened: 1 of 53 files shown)


junkil-park (approved):
LGTM.

PR #6318, opened 2020-10-01
[genesis] disable test dd except for well known test chains

davidiw:


davidiw:
let's see what if anything blows up.... and if nothing, this will be reviewable and not a draft

tzakian (approved):
LGTM! Thanks fo doing this

PR #6325, opened 2020-10-01
Enabling Int Overflow Protection for Release Builds

mickayz:
## Motivation

The Rust compiler does not enable overflow protection for release builds by default.  Enabling this protection will add a check that validates whether an arithmetic operation will overflow the resulting value and panic if so. There is a tradeoff here since panics should generally be avoided, but given that arithmetic errors in Libra could silently lead to severe vulnerabilities including inconsistencies in account balances, consensus errors, liveliness errors, and other unknown unknowns, this tradeoff seems like the right one to make.  There should be no use case where allowing int overflows to occur is necessary.

Yes

## Test Plan

CI tests

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -181,6 +181,7 @@ default-members = [
 
 [profile.release]
 debug = true
+overflow-checks = true
 
 [profile.bench]
 debug = true


bmwill:
IIRC these checks are enabled by default in debug builds but disabled in release builds. Given overflow can cause some interesting corner cases if not handled properly I agree that we should have these checks enabled for our production builds.

dariorussi:
fwiw I believe all interesting places in the VM where using `add_checked` and similar. Of course this PR is more than just the VM so it's fine particularly given it does not seem to have any perf impact.
This was just a fyi

davidiw:
Actually there seems to be a 5% perf regression :O
Probably worth it though

davidiw:
Would anyone be up to starting a 1 week or 2 week long longevity test on this immediately? If not, I vote we postpone until next year.

bmwill:
> Would anyone be up to starting a 1 week or 2 week long longevity test on this immediately? If not, I vote we postpone until next year.

That's actually a really good idea, given we haven't had any long running tests with this enabled we should ideally make sure its battle tested until decide to flip it on for everything

davidiw (changes requested):
Let's postpone and get this into its own longevity test running on a dedicated testnet with lots of cluster testing going on.

davidiw:

PR #6326, opened 2020-10-01
[libra framework] Linking the access control spec to the LIP-2 document

junkil-park:
- Added the hyperlinks from the access control specs in the generated docs to the LIP-2 document
  - Currently, it's linked to the temporary LIP-2 draft in the move-prover directory.
  - This can be easily updated when the official LIP-2 is updated.

- Each tag (e.g., [A1], [H1]) are labeled with either ROLE or PERMISSION.
  - The tag labeled with ROLE is linked to the "roles" table in the LIP-2 doc.
  - The tag labeled with PERMISSION is linked to the "permissions" table in the LIP-2 doc.

## Motivation

To make the formal specs traceable to the access control policy document

Yes

## Test Plan

cargo test

### DIFF
--- a/language/stdlib/modules/AccountLimits.move
+++ b/language/stdlib/modules/AccountLimits.move
@@ -169,8 +169,10 @@ module AccountLimits {
         lr_account: signer;
         to_limit: signer;
         limit_address: address;
+        /// Only ParentVASP and ChildVASP can have the account limits [[E1]][ROLE][[E2]][ROLE][[E3]][ROLE][[E4]][ROLE][[E5]][ROLE][[E6]][ROLE][[E7]][ROLE].
+        include Roles::AbortsIfNotParentVaspOrChildVasp{account: to_limit};
+
         include Roles::AbortsIfNotLibraRoot{account: lr_account};
-        include Roles::AbortsIfNotParentVaspOrChildVasp{account: to_limit}; // Only ParentVASP and ChildVASP can have the account limits [E1][E2][E3][E4][E5][E6][E7].
         aborts_if !exists<LimitsDefinition<CoinType>>(limit_address) with Errors::NOT_PUBLISHED;
         aborts_if exists<Window<CoinType>>(Signer::spec_address_of(to_limit)) with Errors::ALREADY_PUBLISHED;
     }
--- a/language/stdlib/modules/references_template.md
+++ b/language/stdlib/modules/references_template.md
@@ -1,2 +1,2 @@
-
-[]: # (File containing markdown style reference definitions to be included in each generated doc)
+[ROLE]: https://github.com/libra/libra/blob/master/language/move-prover/doc/user/access-control.md#roles
+[PERMISSION]: https://github.com/libra/libra/blob/master/language/move-prover/doc/user/access-control.md#permissions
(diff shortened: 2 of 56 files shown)


language/stdlib/modules/references_template.md
```diff
@@ -1,2 +1,2 @@
-
-[]: # (File containing markdown style reference definitions to be included in each generated doc)
```
wrwg, line 2: This is actually a comment. It looks funny, but this is the recommended way to make comments in Markdown. Please leave it.

language/stdlib/modules/references_template.md
```diff
@@ -1,2 +1,2 @@
...
+[ROLE]: https://github.com/libra/libra/blob/master/language/move-prover/doc/user/access-control.md#roles
```
wrwg, line 1: So we are planning to link only to the table as a whole? Probably OK given that this tables aren't very large.

language/stdlib/modules/AccountLimits.move
```diff
@@ -170,7 +170,7 @@ module AccountLimits {
         to_limit: signer;
         limit_address: address;
         include Roles::AbortsIfNotLibraRoot{account: lr_account};
-        include Roles::AbortsIfNotParentVaspOrChildVasp{account: to_limit}; // Only ParentVASP and ChildVASP can have the account limits [E1][E2][E3][E4][E5][E6][E7].
+        include Roles::AbortsIfNotParentVaspOrChildVasp{account: to_limit}; // Only ParentVASP and ChildVASP can have the account limits [[E1]][ROLE][[E2]][ROLE][[E3]][ROLE][[E4]][ROLE][[E5]][ROLE][[E6]][ROLE][[E7]][ROLE].
```
wrwg, line 173: This doesn't fit on a line even if rendered. Please fix.

wrwg (approved):
I assume you have tested whether links work.

language/stdlib/modules/references_template.md
junkil-park: Added it back.

language/stdlib/modules/AccountLimits.move
junkil-park: Done.

language/stdlib/modules/references_template.md
junkil-park: I think it's OK too. The "roles" table is small enough to fit in a page/screen. The "permissions" table can be a bit over one page/screen, but it may be still good to link the top of the table where the column labels (e.g., H, I, J) are visible. @sblackshear , @tzakian , do you have an opinion on this?

language/stdlib/modules/references_template.md
wrwg: Oops! Just seeing now that this is printed in the documentation, unexpectedly. Yes please remove it, I care later for possibly bringing it back. I hope it doesn't appear in the way you have added it.

language/stdlib/modules/references_template.md
wrwg: Let's not block on this waiting for Sam and Tim, I suggest submit and we can still change it later if desired.

language/stdlib/modules/references_template.md
junkil-park: Ah, OK. I removed it.

PR #6328, opened 2020-10-01
[backup] backup coordinator: epoch ending before other types

msmouse:
## Motivation
So that we will never hit the edge case where a proof in the other backup types fails to verify because required validator set is not yet backed up.

Y

## Test Plan

existing coverage (smoke test) and manual verification on my laptop

### DIFF
--- a/storage/backup/backup-cli/src/backup_types/epoch_ending/restore.rs
+++ b/storage/backup/backup-cli/src/backup_types/epoch_ending/restore.rs
@@ -206,7 +206,7 @@ impl EpochHistory {
         ensure!(!self.epoch_endings.is_empty(), "Empty epoch history.",);
         ensure!(
             epoch <= self.epoch_endings.len() as u64,
-            "History until epoch {} can't varify epoch {}",
+            "History until epoch {} can't verify epoch {}",
             self.epoch_endings.len(),
             epoch,
         );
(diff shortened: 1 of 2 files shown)


storage/backup/backup-cli/src/backup_types/epoch_ending/restore.rs
```diff
@@ -206,7 +206,7 @@ impl EpochHistory {
         ensure!(!self.epoch_endings.is_empty(), "Empty epoch history.",);
         ensure!(
             epoch <= self.epoch_endings.len() as u64,
-            "History until epoch {} can't varify epoch {}",
+            "History until epoch {} can't verify epoch {}",
```
lightmark, line 209: lmao

storage/backup/backup-cli/src/coordinators/backup.rs
```diff
@@ -255,7 +258,7 @@ impl BackupCoordinator {
     ) -> impl StreamExt<Item = ()> + 'a
     where
         S: Copy + Debug + 'a,
-        W: Worker<'a, S, Fut> + Copy + 'static,
+        W: Worker<'a, S, Fut> + Copy + 'a,
```
lightmark, line 261: no idea what you are doing here.

lightmark (approved):

storage/backup/backup-cli/src/coordinators/backup.rs
msmouse: Now there's a closure passed in, so not always `'static` any more..

PR #6329, opened 2020-10-01
[transaction-replay] Added override stdlib option for transaction-replay

runtian-zhou:
## Motivation

Added an `rebuild_stdlib` option so that the bisect transaction will be run against the newly built stdlib file per requested from @sblackshear. This will allow on call to modify the stdlib module as needed to look into the behavior of the network.

Yes
## Test Plan

Run the command locally by spawning up a libra node and add a magic public function to LibraAccount, and try to invoke such function from bisect script. With this flag on, it would be executed successfully. If the flag is off, it would abort with an internal linker error.

### DIFF
--- a/language/libra-tools/transaction-replay/src/lib.rs
+++ b/language/libra-tools/transaction-replay/src/lib.rs
@@ -17,6 +17,7 @@ use libra_vm::{
 use move_core_types::gas_schedule::{GasAlgebra, GasUnits};
 use move_lang::{compiled_unit::CompiledUnit, move_compile_no_report, shared::Address};
 use move_vm_runtime::{logging::NoContextLog, move_vm::MoveVM, session::Session};
+use move_vm_test_utils::{ChangeSet as MoveChanges, DeltaStorage};
 use move_vm_types::gas_schedule::{zero_cost_schedule, CostStrategy};
 use resource_viewer::{AnnotatedAccountStateBlob, MoveValueAnnotator};
 use std::{convert::TryFrom, path::Path};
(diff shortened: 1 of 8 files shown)


language/libra-tools/transaction-replay/src/lib.rs
```diff
@@ -155,11 +157,12 @@ impl LibraDebugger {
         sender: AccountAddress,
         begin: Version,
         end: Version,
+        override_changeset: Option<MoveChanges>,
     ) -> Result<Option<Version>> {
         let predicate = compile_move_script(code_path, sender)?;
```
sblackshear, line 162: Unrelated, but: this is compiling against the local Move source, not the on-chain bytecodes. Probably worth adding a TODO to fix that later.

language/libra-tools/transaction-replay/src/lib.rs
```diff
@@ -132,13 +133,14 @@ impl LibraDebugger {
...
+    pub fn run_session_at_version<F>(&self, version: Version, override_changeset: Option<MoveChanges>, f: F) -> Result<ChangeSet>
     where
-        F: FnOnce(&mut Session<RemoteStorage<DebuggerStateView>>) -> VMResult<()>,
+        F: FnOnce(&mut Session<DeltaStorage<RemoteStorage<DebuggerStateView>>>) -> VMResult<()>,
     {
         let move_vm = MoveVM::new();
         let state_view = DebuggerStateView::new(&*self.debugger, version);
-        let remote_storage = RemoteStorage::new(&state_view);
+        let state_view_storage = RemoteStorage::new(&state_view);
+        let remote_storage = DeltaStorage::new(&state_view_storage, override_changeset.unwrap_or_else(|| MoveChanges::new()));
```
sblackshear, line 143: Might make sense to use `RemoteStorage` if `override_changeset` is `None`?

language/libra-tools/transaction-replay/src/lib.rs
```diff
@@ -155,11 +157,12 @@ impl LibraDebugger {
         sender: AccountAddress,
         begin: Version,
         end: Version,
+        override_changeset: Option<MoveChanges>,
     ) -> Result<Option<Version>> {
         let predicate = compile_move_script(code_path, sender)?;
         let gas_table = zero_cost_schedule();
         let is_version_ok = |version| {
-            self.run_session_at_version(version, |session| {
+            self.run_session_at_version(version, override_changeset.clone(), |session| {
```
sblackshear, line 165: `clone` needed?

sblackshear (commented):
Looks great, thank you for adding this feature! 

Could you add a test for this?

language/libra-tools/transaction-replay/src/lib.rs
runtian-zhou: We actually need this clone as the function signature requires an `Fn` rather than `FnOnce`. You could imagine that this function will be run against multiple versions. Each run will require a copy.

language/libra-tools/transaction-replay/src/lib.rs
runtian-zhou: It might break the typing... The type for the function is `Session<DeltaStorage<RemoteStorage<DebuggerStateView>>> -> VMResult<()>`, so it has to be `DeltaStorage` unless we use some dynamic dispatching over there, which I'm not sure how to do that.

language/libra-tools/transaction-replay/src/lib.rs
runtian-zhou: I'll add a TODO in the binary.

sblackshear (approved):

dariorussi (approved):
good to go!

PR #6331, opened 2020-10-02
[move source language] Check for overlapping targets/deps

tnowacki:
- Added check for something that bit me a lot in the CLI
- Deps and Targets cannot overlap

## Motivation

- I got very confused and want to prevent others from being confused

## Test Plan

- A bit hard to test currently, as none of the move testing infra uses specified files
- Might add a CLI test

### DIFF
--- a/language/move-lang/src/lib.rs
+++ b/language/move-lang/src/lib.rs
@@ -29,7 +29,7 @@ use move_ir_types::location::*;
 use parser::syntax::parse_file_string;
 use shared::Address;
 use std::{
-    collections::{BTreeMap, HashMap},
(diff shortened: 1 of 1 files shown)


bob-wilson (approved):
Seems like a good idea to check this. A test would certainly be nice if you think of a way to do that in the future.

sblackshear:
I think it should be pretty easy to write a move CLI expected value test for this

tnowacki:
> I think it should be pretty easy to write a move CLI expected value test for this

I can't do it yet because the setup for check/build is a bit off in the CLI. If I get around to fixing that, I'll add a test

PR #6337, opened 2020-10-02
[backup] backup coordinator: new strategy for txn batching

msmouse:
## Motivation

With txn batch size at 100, now we produce txn batches (0, 99), (100, 199), ..., (900, 999), (1000, 1099), ...
Which is not ideal.
Consider this:  With state snapshot interval at 1000, we have state snapshots at 0, 1000, 2000, ...
Assume the chain is at version 1090, so the (1000, 1099) txn batch hasn't showed up yet. Now, let's restore to the latest possible versions supported by the backups -- surprisingly, the latest snapshot can't be used, because there's no txn at that version or on top of it.

This commit shifts the txn batches by 1 -- make the genesis txn into it's own backup -- and there is no longer such issue. I think the new strategy makes more sense.

Y

## Test Plan
existing coverage.  play locally.

## Related PRs

### DIFF
--- a/storage/backup/backup-cli/src/coordinators/backup.rs
+++ b/storage/backup/backup-cli/src/coordinators/backup.rs
@@ -289,13 +289,15 @@ where
 }
 
 fn get_batch_range(last_in_backup: Option<u64>, batch_size: usize) -> (u64, u64) {
-    // say, 5 is already in backup, and we target batches of size 10, we will return (6, 9) in this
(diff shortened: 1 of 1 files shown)


lightmark:
lol, it is more like a bug imo.

lightmark (approved):

msmouse:
> lol, it is more like a bug imo.

🏃‍♀️

PR #6340, opened 2020-10-02
[scripts] Added convenience script start_cli_swarm.sh

villesundell:
## Motivation
Recently swarm was changed so, that --libra-node and --cli-path are
madatory, raising two issues:
    - there were no easy way to lauch swarm with CLI, because the
    documentation was out-of-date, and
    - the new way of lauching would not fit well to the documentation
      because the user might have "debug" or "release" version.

This script solves both problems by making lauching the swarm nice and
 easy. It also forwards aguments after "--" straight to the swarm, so
 user can configure their swarm if so desired.

Default is to build and use "debug" build, but "release" build can be
specified by issuing "-r" or "--release".

Yes.

## Test Plan

Manually tested by running:
`./scripts/cli/start_cli_swarm.sh`

Also tested release build by running:
`./scripts/cli/start_cli_swarm.sh -r`

## Related PRs

No related PRs, the documentation change is part of this PR as a separate commit.

### DIFF
--- a/developers.libra.org/docs/reference/libra-cli.md
+++ b/developers.libra.org/docs/reference/libra-cli.md
@@ -18,9 +18,9 @@ To connect to the testnet through the CLI, a convenience script can be used to i
 ### Run a Local Libra Network and Spawn a CLI Client
 To start a local Libra network and spawn a CLI client that connects to this local network, run:
 ```bash
-cargo run -p libra-swarm -- -s
+./scripts/cli/start_cli_swarm.sh
 ```
-The `-s` option causes the CLI to be run after the local Libra network is launched.  Note that this may take a few minutes to build and then start.
+Note that this may take a few minutes to build and then start.
 
 ### Run a CLI Client to Connect to Any Libra Network
 To invoke the CLI client and configure it yourself, run:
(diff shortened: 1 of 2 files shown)


developers.libra.org/docs/reference/libra-cli.md
```diff
@@ -18,9 +18,9 @@ To connect to the testnet through the CLI, a convenience script can be used to i
 ### Run a Local Libra Network and Spawn a CLI Client
 To start a local Libra network and spawn a CLI client that connects to this local network, run:
 ```bash
-cargo run -p libra-swarm -- -s
```
davidiw, line 21: Thanks for this great contribution1

These instructions are really old, we've been updating the docs to point to this new process: https://github.com/libra/libra/blob/master/developers.libra.org/docs/run-local-network.md

I'd be hesitant to add back swarm at the moment and would rather we move toward a unified process of using the `libra-node --test` and go from there.

Will leave the call up to @bmwill

bmwill (approved):
Thank you for adding this wrapper script. I should have added something similar back when I made those changes to libra-swarm. I agree that with @davidiw that most folks should probably just use the `--test` feature of libra-node but we still need a way to start a local network of more than 1 node and this is that mechanism!

/land

PR #6342, opened 2020-10-02
[language][vm] add tests for exec_function and fix error code

vgao1996:
Currently if you call `execute_function` with a nonexistent module id, the VM correctly produces an invariant violation. However if  calling with a nonexistent function name (while the module exists), it returns `LINKER_ERROR`, which is classified as a user error. This gets it fixed and adds tests for both cases.


dariorussi (approved):
great thanks!!!

PR #6344, opened 2020-10-02
[libra framework] Update some transaction docs

tzakian:
Updates transaction docs to include missing error categories that were found during verification.


wrwg (approved):

junkil-park (approved):
LGTM. There is a small issue, which doesn't need to block this PR. There is one error code left undocumented in `burn.move`. It's probably not documented because it seems to be impossible to happen. However, Prover does not prove that fact currently. I will take a look at the issue once this PR lands.

PR #6350, opened 2020-10-02
[language][vm] add tests for storage backends returning errors and fix error handling

vgao1996:
Currently `trait RemoteCache` allows storage backends to return errors. However in those cases, the VM simply propagates whatever error the storage returns, instead of catching it and returning an invariant violation instead. This gets it fixed and adds tests.

### DIFF
--- a/language/move-vm/runtime/src/data_cache.rs
+++ b/language/move-vm/runtime/src/data_cache.rs
@@ -180,8 +180,8 @@ impl<'r, 'l, C: RemoteCache> DataStore for TransactionDataCache<'r, 'l, C> {
             };
             let ty_layout = self.loader.type_to_type_layout(ty)?;
 
-            let gv = match self.remote.get_resource(&addr, &ty_tag)? {
-                Some(blob) => {
+            let gv = match self.remote.get_resource(&addr, &ty_tag) {
+                Ok(Some(blob)) => {
                     let ty_kind_info = self.loader.type_to_kind_info(ty)?;
                     let val = match Value::simple_deserialize(&blob, &ty_kind_info, &ty_layout) {
                         Some(val) => val,
@@ -197,7 +197,19 @@ impl<'r, 'l, C: RemoteCache> DataStore for TransactionDataCache<'r, 'l, C> {
 
                     GlobalValue::cached(val)?
                 }
-                None => GlobalValue::none(),
+                Ok(None) => GlobalValue::none(),
+                Err(err) => {
+                    let msg = format!("Unexpected storage error: {:?}", err);
+                    // REVIEW: better way to get info out of a PartialVMError?
+                    let (_old_status, _old_sub_status, _old_message, indices, offsets) =
+                        err.all_data();
+                    return Err(
+                        PartialVMError::new(StatusCode::UNKNOWN_INVARIANT_VIOLATION_ERROR)
+                            .with_message(msg)
+                            .at_indices(indices)
+                            .at_code_offsets(offsets),
+                    );
+                }
             };
 
             account_cache.data_map.insert(ty.clone(), (ty_layout, gv));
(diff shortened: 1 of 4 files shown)


language/move-vm/runtime/src/data_cache.rs
```diff
@@ -197,7 +199,20 @@ impl<'r, 'l, C: RemoteCache> DataStore for TransactionDataCache<'r, 'l, C> {
 
                     GlobalValue::cached(val)?
                 }
-                None => GlobalValue::none(),
+                Ok(None) => GlobalValue::none(),
+                Err(err) => {
+                    println!("HHIIIIIT");
```
dariorussi, line 204: remove? 😉

dariorussi (approved):
Looks good!
Please checkout the `println` and if you could take a look at our runtime code base and see if we have any left that would be wonderful!

language/move-vm/runtime/src/data_cache.rs
vgao1996: Yeah will get it removed. Agreed it's a good idea to check if there are any `println`s we forgot to delete.

language/move-vm/runtime/src/data_cache.rs
```diff
@@ -197,7 +199,20 @@ impl<'r, 'l, C: RemoteCache> DataStore for TransactionDataCache<'r, 'l, C> {
...
+                    let msg = format!("Unexpected storage error: {:?}", err);
+                    // REVIEW: better way to get info out of a PartialVMError?
```
tnowacki, line 206: I think it might be fair to add an all_data equivalent to `PartialVMError`. I never needed it so I never added it

language/move-vm/runtime/src/data_cache.rs
```diff
@@ -197,7 +199,20 @@ impl<'r, 'l, C: RemoteCache> DataStore for TransactionDataCache<'r, 'l, C> {
...
+                    let (_old_status, _old_sub_status, _old_message, _location, indices, offsets) =
+                        err.finish(Location::Undefined).all_data();
+                    return Err(
+                        PartialVMError::new(StatusCode::UNKNOWN_INVARIANT_VIOLATION_ERROR)
```
tnowacki, line 210: Why is this an invariant violation? I don't doubt it, just am confused

vgao1996: Here's my reasoning: when the storage backend returns you an error, it means something is really really wrong in the sense that you have no idea what has happened at all, and you want to fail fast and discard the transaction. This is not a user error and you should never charge the user for it.

IMO what makes this less obvious and error prone is that `trait RemoteCache` requires the storage to use `VMError` and `PartialVMError`, the same stuff the VM uses internally itself, instead of a dedicated `VMStorageError` or maybe an associated error type. This enables a shortcut where the VM can just propagates whatever errors the storage returns without handling them, which is very wrong.

tnowacki: This makes sense.
Maybe give it it's own invariant violation code? Or a bucket for storage invariant violations? Fine either way though

vgao1996: > a bucket for storage invariant violations
I think this is definitely something we should keep thinking about. At the moment I don't think we fully understand what kind of interfaces we want for storage integration, and if the there's anything special about storage to justify a separate bucket of invariant violations. Also I feel like it's a little bit too late to introduce changes like this.

> give it it's own invariant violation code?
We could, but I guess it's not high priority at the moment -- the Libra VM only cares about the bucket so as long as the error returned is in the correct bucket, we're good. Error codes in the invariant violation bucket are a little bit inconsistent and unorganized in general, so we'll definitely revisit them later.

PR #6352, opened 2020-10-02
[safety-rules] add latency metrics

davidiw:
* individual get and set queries to vault
* time for client to receive a response (including serialization to and from)
* time for service to handle the query internally including vault time

### DIFF
--- a/consensus/safety-rules/src/persistent_safety_storage.rs
+++ b/consensus/safety-rules/src/persistent_safety_storage.rs
@@ -92,6 +92,7 @@ impl PersistentSafetyStorage {
     }
 
     pub fn author(&self) -> Result<Author, Error> {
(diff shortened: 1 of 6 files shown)


consensus/safety-rules/src/persistent_safety_storage.rs
```diff
@@ -92,19 +92,22 @@ impl PersistentSafetyStorage {
     }
 
     pub fn author(&self) -> Result<Author, Error> {
+        let _ = counters::start_timer("get", OWNER_ACCOUNT);
```
bmwill, line 95: Change all of these to be _timer instead of _ as they have different semantics. _ drops immediately while _timer will drop at the end of the scope

bmwill (changes requested):

bmwill (approved):

PR #6354, opened 2020-10-03
[Vault] Automatically trim old/stale keys after key rotations.

JoshLind:
## Motivation

This PR updates the vault client to automatically trim the number of keys held in vault after each key rotation. This prevents vault from collecting old/stale keys that should be discarded (and might otherwise be cause for security vulnerabilities, e.g., key gathering attacks). To achieve this, the PR offers two commits:

1. First, we update the vault client to automatically trim the key versions held in storage after each key rotation. The maximum number of keys is hard-coded at 4 (a rough guesstimate: a higher number might be problematic, a lower number might be too strict and make failure recovery more difficult).
2. Second, we add a unit test to verify this functionality.

Note that before keys can be trimmed in vault, the minimum encryption/decryption versions must be updated, otherwise the call will fail. The vault API for trimming keys can be found here: https://www.vaultproject.io/api-docs/secret/transit

Yes.

## Test Plan

All tests pass locally, including the newly added one.

## Related PRs

None.

### DIFF
--- a/secure/storage/src/vault.rs
+++ b/secure/storage/src/vault.rs
@@ -350,7 +350,7 @@ impl CryptoStorage for VaultStorage {
     fn rotate_key(&mut self, name: &str) -> Result<Ed25519PublicKey, Error> {
         let ns_name = self.crypto_name(name);
         self.client().rotate_key(&ns_name)?;
-        self.get_public_key(name).map(|v| v.public_key)
+        Ok(self.client().trim_key_versions(&ns_name)?)
     }
 
     fn sign<T: CryptoHash + Serialize>(
--- a/secure/storage/src/vault.rs
+++ b/secure/storage/src/vault.rs
@@ -20,6 +20,9 @@ use std::{
     },
 };
 
+#[cfg(any(test, feature = "testing"))]
+use libra_vault_client::ReadResponse;
+
 const LIBRA_DEFAULT: &str = "libra_default";
 
 /// VaultStorage utilizes Vault for maintaining encrypted, authenticated data for Libra. This
@@ -125,6 +128,14 @@ impl VaultStorage {
     #[cfg(any(test, feature = "testing"))]
     pub fn revoke_token_self(&self) -> Result<(), Error> {
         Ok(self.client.revoke_token_self()?)
+    }
+
+    #[cfg(any(test, feature = "testing"))]
+    pub fn get_all_key_versions(
+        &self,
+        name: &str,
+    ) -> Result<Vec<ReadResponse<Ed25519PublicKey>>, Error> {
+        Ok(self.client().read_ed25519_key(name)?)
     }
 
     /// Creates a token but uses the namespace for policies
--- a/secure/storage/vault/src/lib.rs
+++ b/secure/storage/vault/src/lib.rs
@@ -22,6 +22,10 @@ use ureq::Response;
 #[cfg(any(test, feature = "fuzzing"))]
 pub mod fuzzing;
 
+/// The max number of key versions held in vault at any one time.
+/// Keys are trimmed in FIFO order.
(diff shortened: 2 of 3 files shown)


davidiw (changes requested):
Awesome first stab and glad to see we have all the pieces. The problem with this approach is that you assume that the retrieve of a public key should have write abilities on that key, that's dangerous. I would equate rotate privileges as ability to delete / trim. Let's unify it at that layer.

JoshLind:
Thanks @davidiw!

> The problem with this approach is that you assume that the retrieve of a public key should have write abilities on that key, that's dangerous. 

Can you clarify exactly what you mean here? My understanding is that this approach requires: (i) read capabilities (to fetch all the versions); (ii) rotate capabilities; (iii) write capabilities (to set the minimum versions for encryption/decryption and to trim). 
- Are you proposing splitting this up so that trims don't occur automatically on a rotate, but instead we require a separate call trim explicitly? Hence the split in privileges. Or something else?
- Generally, trimming without read capabilities is difficult because you need to be able to provide fault tolerance (e.g., you can try and hold all version numbers in memory so you don't need to read on a trim, but that approach falls apart when the process needs to restart).

secure/storage/src/vault.rs
```diff
@@ -350,7 +361,7 @@ impl CryptoStorage for VaultStorage {
     fn rotate_key(&mut self, name: &str) -> Result<Ed25519PublicKey, Error> {
         let ns_name = self.crypto_name(name);
         self.client().rotate_key(&ns_name)?;
-        self.get_public_key(name).map(|v| v.public_key)
+        Ok(self.client().trim_key_versions(&ns_name)?)
     }
```
davidiw, line 366: sigh, I screwed up...

secure/storage/src/vault.rs
```diff
@@ -127,6 +130,14 @@ impl VaultStorage {
...
 
+    #[cfg(any(test, feature = "testing"))]
+    pub fn get_all_key_versions(
+        &self,
+        name: &str,
+    ) -> Result<Vec<ReadResponse<Ed25519PublicKey>>, Error> {
+        Ok(self.client().read_ed25519_key(name)?)
+    }
+
```
davidiw, line 140: just hold on to the first key, rotate 5 times, and ensure you cannot get the key.

davidiw (approved):
let's avoid the code that adds additional "features" if we can

JoshLind:
Thanks @davidiw! 😄

> just hold on to the first key, rotate 5 times, and ensure you cannot get the key.

I originally thought of that approach but my concern was that it leaves too much "unknown" about the internal state -- e.g., that test will pass if the trimming operation just trims all versions of the key. I know we could get around it by then remembering all versions we fetch and checking they they all exist after rotations etc. but I wanted to be extra neurotic about testing this change because we're doing it so late.. 😟 

I'll land this for now (just to get it in). If you still feel strongly about addressing this, feel free to push back and I'll update the test to be less intrusive.

/land

secure/storage/vault/src/lib.rs
```diff
@@ -318,6 +317,76 @@ impl Client {
...
+    /// Trims the key versions according to the minimum available version specified.
+    /// This operation deletes any older keys and cannot be undone.
+    fn set_minimum_available_version(
+        &self,
+        name: &str,
+        min_available_version: u32,
+    ) -> Result<(), Error> {
+        let request = self
+            .agent
+            .post(&format!("{}/v1/transit/keys/{}/trim", self.host, name));
```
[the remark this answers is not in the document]
JoshLind: @mgorven, yeah, I was going to ping you this morning! Seems like you're already on it 😄 Let us know what you need from our side!

PR #6357, opened 2020-10-03
[RFC][libra-framework][breaking] Change reconfiguration rule

runtian-zhou:
## Motivation

We would like to be able to perform mulitple reconfigurations within one transaction that will emit only one reconfiguration event. To achieve this, I made following modification:

1. Relax the LibraConfig module so that when the last reconfiguration time is equal to current time, we would perform nil op rather than abort
2. Relax the writeset cross check such that if the writeset transaction’s trying to touch the value modified by epilogue, as long as they are modifying to the same WriteOp, such writeset should be accepted

Justification: These two changes will make changing config so much easier without having to change the code drastically. We used to have a lot of code in the LibraConfig to suppress such behavior in the genesis, which could be all gone. The reason why we couldn’t do this thing before was because we didn’t mark the transactions after the first reconfiguration to be Retry, thus the semantic of that transaction would be problematic.

yes

## Test Plan

modified existing test suites. Will add other tests to e2e test to test this behavior.


sblackshear:
What would go wrong if we did (1), but not (2)?

language/libra-vm/src/libra_transaction_executor.rs
```diff
@@ -687,6 +672,22 @@ impl LibraVM {
     }
 }
 
+fn check_intersection<K: Eq + std::hash::Hash, V: Eq>(
+    lhs: impl Iterator<Item = (K, V)>,
+    rhs: impl Iterator<Item = (K, V)>,
```
tzakian, line 677: Would passing a `&HashMap<K, V>` for both of these simplify things (wouldn't need the `.collect`s and the iter/maps above too.

language/stdlib/modules/LibraConfig.move
```diff
@@ -224,6 +224,12 @@ module LibraConfig {
 
         let config_ref = borrow_global_mut<Configuration>(CoreAddresses::LIBRA_ROOT_ADDRESS());
         let current_time = LibraTimestamp::now_microseconds();
+
+        // Do not do anything if a reconfiguration event is already emitted within this transaction.
+        if (current_time == config_ref.last_reconfiguration_time) {
+            return ()
+        };
+
```
tzakian, line 232: Is the early return at the start of this function needed still after this change?

tzakian (commented):
This looks good to me, and I think would be a very nice relaxation of the current rules.

runtian-zhou:
> What would go wrong if we did (1), but not (2)?

Going (1) only will prevent us from doing multiple reconfigs in one admin script. It would also be harder for us to generate an offline writeset as the first reconfig event will guarenteed to be emitted. I'm not too worried about the first use case because the libra root should be able to send arbitrary script, which could utilize such feature. The second one is a bit troublesome.

language/stdlib/modules/LibraConfig.move
runtian-zhou: This will not be needed. Although I'm trying to minimize the change at the moment. To get rid of those lines we will need to reorganize the genesis code a little bit, which I'm not planning to work on in this PR.

sblackshear:
> Going (1) only will prevent us from doing multiple reconfigs in one admin script. 

Ok, got it. I keep forgetting about the "epilogue runs on initial state instead of post-state" thing. Make it so harder to reason about the behavior...

language/move-lang/functional-tests/tests/libra/reconfiguration/reconfiguration.move
```diff
@@ -25,19 +25,28 @@ use 0x1::LibraConfig;
...
 
 fun main(account: &signer) {
     LibraConfig::reconfigure(account);
 }
 }
-// check: "Keep(ABORTED { code: 1025,"
+// check: NewEpochEvent
+// check: event_data: "0300000000000000"
+// check: "Keep(EXECUTED)"
+ß
```
sblackshear, line 52: wat

language/stdlib/modules/LibraConfig.move
```diff
@@ -224,6 +224,12 @@ module LibraConfig {
 
         let config_ref = borrow_global_mut<Configuration>(CoreAddresses::LIBRA_ROOT_ADDRESS());
         let current_time = LibraTimestamp::now_microseconds();
+
+        // Do not do anything if a reconfiguration event is already emitted within this transaction.
+        if (current_time == config_ref.last_reconfiguration_time) {
+            return ()
```
sblackshear, line 230: Nit: `return;` instead?

language/stdlib/modules/LibraConfig.move
```diff
@@ -224,6 +224,12 @@ module LibraConfig {
 
         let config_ref = borrow_global_mut<Configuration>(CoreAddresses::LIBRA_ROOT_ADDRESS());
         let current_time = LibraTimestamp::now_microseconds();
+
+        // Do not do anything if a reconfiguration event is already emitted within this transaction.
```
sblackshear, line 228: I think it would be worth explaining this in a lot more detail because it's pretty subtle:
- The time changes in every non-empty block
- A block automatically ends after a transaction that emits a reconfiguration event
- Each transaction must emit at most one reconfiguration event
- Thus, this check ensures that a transaction that does multiple "reconfiguration required" actions emits only one reconfiguration event

language/libra-vm/src/libra_transaction_executor.rs
```diff
@@ -687,6 +672,22 @@ impl LibraVM {
     }
 }
 
+fn check_intersection<K: Eq + std::hash::Hash, V: Eq>(
```
sblackshear, line 675: Could you add a comment to say what this is doing + why?

language/libra-vm/src/libra_transaction_executor.rs
```diff
@@ -523,33 +523,18 @@ impl LibraVM {
...
-            .is_disjoint(
-                &change_set
-                    .events()
-                    .iter()
-                    .map(|event| event.key())
-                    .collect::<HashSet<_>>(),
-            )
-        {
+
+        if !check_intersection(
```
sblackshear, line 534: I don't understand how this will prevent the duplicate reconfiguration events. For example:

```
epilogue.events: [Reconfig { seq: 1, time: 2}]
change_set.events: `[Reconfig { seq: 1, time: 2 }, Reconfig { seq: 2, time 3 }]
```

I don't think this can be prevented without hardcoding this check

language/libra-vm/src/libra_transaction_executor.rs
sblackshear: Also: might make sense to add this as a function on `WriteSet` (I know you need a separate one for events, but still).

language/libra-vm/src/libra_transaction_executor.rs
sblackshear: Set of <K,V> pairs might also work (+ then you could use the built-in intersection trait).

sblackshear (changes requested):

language/stdlib/modules/LibraConfig.move
runtian-zhou: Just added the comments with these notes and a few extra details. Lemme know how they look to you!

language/move-lang/functional-tests/tests/libra/validator_set/reconfigure_twice_in_block.move
```diff
@@ -1,7 +1,5 @@
-// Checks that only one change to the validator set can be made per block:
-// if the first transaction removes a validator and the second transaction
-// removes a validator in the same block, then the first transaction will succeed
-// and the second will fail.
+// Checks that only two reconfigurations can be done within the same transaction and will only emit one reconfiguration
+// event.
```
zekun000, line 2: nit: the file name should be `reconfigure_twice_in_transaction` not `in_block`

language/libra-vm/src/libra_transaction_executor.rs
sblackshear: Nit: naming this something like `intersection_matches` might make the meaning of the bool return clearer.

sblackshear (approved):

runtian-zhou:
Closing this PR as the other one is less intrusive.

PR #6358, opened 2020-10-03
[json-rpc] add header forwarded into http request log

xli:
In case the request is forwarded from a proxy server, log forwarded for client address info

### DIFF
--- a/json-rpc/src/runtime.rs
+++ b/json-rpc/src/runtime.rs
@@ -18,6 +18,7 @@ use std::{net::SocketAddr, sync::Arc};
 use storage_interface::DbReader;
 use tokio::runtime::{Builder, Runtime};
 use warp::{
+    http::header,
     reject::{self, Reject},
     Filter,
 };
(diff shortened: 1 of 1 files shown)


libra-action:

Breaking changes policy:

2. Adding fields is ok.

json-rpc/src/runtime.rs
```diff
@@ -122,6 +124,10 @@ pub fn bootstrap(
                 referer: info.referer(),
                 user_agent: info.user_agent(),
                 elapsed: info.elapsed(),
+                forwarded: info
+                    .request_headers()
+                    .get(header::FORWARDED)
```
mgorven, line 129: What's the header name?

xli: forwarded

xli: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Forwarded

bmwill: So we know that HAproxy sets this header with the actual IP of the remote?

mgorven: I'll get that part done.

mgorven (approved):

PR #6365, opened 2020-10-03
[easy][GHA] fix typo in post-land test retry limit

sausagee:
## Motivation
This fixes the typo in the polling retry limit in the post-land test. 

## Test Plan
CI


davidiw (approved):
that's a lot of retries :O

sausagee:
> that's a lot of retries :O

It seems a lot because it's a polling loop. It polls the repo once per minute for up to 30 minutes. Once we get the GHA runners up, we can migrate the build over and this poll/retry won't be necessary.

PR #6367, opened 2020-10-04
[network] improve context and clarity of noise handshake errors

phlip9:
[network][1/n] move transport into its own module directory 

[network][2/n] improve context and clarity of noise handshake errors

+ The errors now clearly state

    (1) that the error came from the noise handshake
    (2) which side you are (client || server)
    (3) if you're the server, then the client peerid when it's known
    (4) the short peerid when it makes sense

+ For example, the infamous "unexpected eof" error will show much more clearly in the logs as (approximately):

    + client logs: "noise client: error reading server handshake response message, server probably rejected our handshake message: unexpected eof"

    + server logs: "noise server: client f9b3451a: knwon client connecting to us with unauthenticated public key: 82c835c08806a8660e49d9932ffc84f47abd80c1e4eb884890697354c69b08fe"

+ Removed some useless security logs that were really just spamming redundant information. There's an initial attempt at classifying certain errors as "should security log"; for example, a client failing to even build their initial handshake message or detecting a replay attack. I'll probably expand this later.

+ Made some debug! logs trace! and removed structured logging from them, since they're pretty much purely a local debugging utility and the structured logs were adding too much noise.

[network][3/n] add some key noise handshake tests that weren't covered

[network][4/n] make x25519 pubkey display less noisy

Display:

before: "[X25519 public key 82c835c08806a8660e49d9932ffc84f47abd80c1e4eb884890697354c69b08fe]"
after: "82c835c08806a8660e49d9932ffc84f47abd80c1e4eb884890697354c69b08fe"

Debug:

before: "[X25519 public key 82c835c08806a8660e49d9932ffc84f47abd80c1e4eb884890697354c69b08fe]"
after: "x25519::PublicKey(82c835c08806a8660e49d9932ffc84f47abd80c1e4eb884890697354c69b08fe)"


mimoo:
I don't see why we should remove the security logs

gregnazario (dismissed):
Looks a lot cleaner not to have a bunch of error messages thrown around the code.

Mostly, just some clarification on error messages so that the reader can know what's going on without having to dig into the code

network/src/noise/error.rs
```diff
@@ -0,0 +1,99 @@
...
+    Other(#[from] io::Error),
+}
+
+impl NoiseHandshakeError {
+    /// Errors that are either clear bugs or indicate some security issue. Should
+    /// immediately alert an engineer if we hit one of these errors.
+    // TODO(philiphayes): remove
+    #[allow(dead_code)]
+    pub fn should_security_log(&self) -> bool {
+        use NoiseHandshakeError::*;
```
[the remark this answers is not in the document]
gregnazario: Actually, I now realize this is dead code, were you going to use this somewhere?

crypto/crypto/src/x25519.rs
```diff
@@ -256,13 +256,15 @@ impl traits::ValidCryptoMaterial for PublicKey {
...
     }
 }
 
 impl std::fmt::Debug for PublicKey {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
-        write!(f, "[X25519 public key: {}]", hex::encode(self.0))
+        f.debug_tuple("x25519::PublicKey")
+            .field(&hex::encode(&self.0))
+            .finish()
     }
```
davidiw, line 268: not clear the value add of these... but doesn't it make it asymmetric from the other crypto keys?

network/src/noise/handshake.rs
```diff
@@ -373,53 +341,38 @@ impl NoiseUpgrader {
...
         if let Some(trusted_peers) = self.auth_mode.trusted_peers() {
             match trusted_peers
                 .read()
-                .map_err(|_| {
-                    io::Error::new(
-                        io::ErrorKind::Other,
-                        "noise: unable to read trusted_peers lock",
-                    )
-                })?
+                .map_err(|_| NoiseHandshakeError::LockPoisoned(remote_peer_short))?
```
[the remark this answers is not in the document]
davidiw: why are we logging on this? shouldn't this be an expect / unwrap?

network/src/noise/error.rs
```diff
@@ -0,0 +1,99 @@
...
+        match self {
+            MissingRemotePublicKey
+            | BuildClientHandshakeMessage(_)
+            | SelfDial
```
davidiw, line 93: why is this a security error?

network/src/noise/error.rs
```diff
@@ -0,0 +1,99 @@
...
+            | LockPoisoned(_)
+            | ServerReplayDetected(_, _) => true,
```
davidiw, line 95: won't logging this cost more than the attacker?

cc: @AnomalRoil / @mimoo should we consider some sort of built-in rate limiter for security logs such that they can fire at most 1/second or 1/5 seconds?

davidiw (commented):
Really happy to see clearer / more useful errors! I think this PR is still a work in progress, so going to dismiss @gregnazario's approval

network/src/noise/error.rs
gregnazario: Have we thought about using the `sample!`?

bmwill: `sample!` does exactly this :)

network/src/transport/mod.rs
```diff
@@ -244,12 +237,6 @@ async fn upgrade_inbound<T: TSocket>(
...
         .map_err(|err| {
-            warn!(
-                SecurityEvent::InvalidNetworkHandshakeMsg,
-                error = err.to_string(),
-                origin = "inbound",
-                remote_peer = remote_peer_id
-            );
             let err = format!(
                 "handshake negotiation with peer {} failed: {}",
                 remote_peer_id, err
```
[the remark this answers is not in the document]
phlip9: right, we already log failed inbound and outbound connection upgrades with warn! level. in fact, I think we double log outbound upgrades errors, one in peer_manager and one in connectivity_manager

network/src/noise/error.rs
```diff
@@ -0,0 +1,99 @@
...
+    SelfDial,
+
+    #[error("noise server: client {0}: client is expecting us to have the wrong public key: {1}")]
+    ClientExpectingWrongPubkey(ShortHexStr, String),
+
+    #[error("noise server: client {0}: error parsing handshake init message: {1}")]
+    ServerParseClient(ShortHexStr, NoiseError),
+
+    #[error("noise server: client {0}: lock is poisoned")]
+    LockPoisoned(ShortHexStr),
```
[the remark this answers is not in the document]
phlip9: personally, I don't think we should return Err on lock poisoend, instead we should probably log, abort, and restart the node...

network/src/noise/error.rs
```diff
@@ -0,0 +1,99 @@
...
+
+    #[error(
+        "noise server: client {0}: known client connecting to us with \
+         unauthenticated public key: {1}"
+    )]
```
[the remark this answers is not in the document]
phlip9: this means a client is connecting with a self-reported peer id I recognize (e.g., it's one of the validators' peer ids) but the public key they provided is not one of the public keys that we trust for that peer id. this could be a transient error during a network key rotation or someone attempting to impersonate a validator for example.

network/src/noise/error.rs
```diff
@@ -0,0 +1,99 @@
...
+    UnauthenticatedClientPubkey(ShortHexStr, String),
+
+    #[error("noise server: client {0}: client connecting with unauthenticated peer id: {1}")]
+    UnauthenticatedClient(ShortHexStr, PeerId),
```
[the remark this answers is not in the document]
phlip9: "unauthenticated peer id" pretty much means "not a validator peer id" in the validator network context

network/src/noise/error.rs
```diff
@@ -0,0 +1,99 @@
...
+    ServerReplayDetected(ShortHexStr, u64),
+
+    #[error("noise server: client {0}: error building handshake response message: {1}")]
+    BuildServerHandshakeMessage(ShortHexStr, NoiseError),
+
+    #[error("noise server: client {0}: error sending server handshake response message: {1}")]
+    ServerWrite(ShortHexStr, io::Error),
+
+    #[error("noise: {0}")]
+    Other(#[from] io::Error),
```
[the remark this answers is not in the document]
phlip9: Yeah, this is pretty much just catching the `socket.flush()?`, which is iirc infallible on unbuffered `TcpStream`. I'll rename this

network/src/noise/error.rs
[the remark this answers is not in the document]
phlip9: Yeah I have some more stuff I'll push out after this PR. I'm still kind of experimenting with the best way to do security logs here

network/src/noise/error.rs
mimoo: someone could have found a way to make you self dial, perhaps to DoS you

crypto/crypto/src/x25519.rs
phlip9: actually the ed25519 looks like this

```
Ed25519PublicKey(..)
```

so this makes it consistent

phlip9:
rebased, re-added the security logs inside a sample!, renamed some errors, x25519::PublicKey Display/Debug is 1000% consistent with other crypto/ types

network/src/transport/mod.rs
```diff
@@ -286,7 +287,22 @@ async fn upgrade_outbound<T: TSocket>(
...
+                    SampleRate::Duration(Duration::from_secs(15)),
+                    error!(
+                        SecurityEvent::NoiseHandshake,
+                        NetworkSchema::new(&ctxt.noise.network_context)
+                            .network_address(&addr)
+                            .connection_origin(&origin),
+                        error = %err,
+                    )
+                );
+            }
```
gregnazario, line 303: @mimoo Is this sufficient for what you want to see out of security logging?  Should we use metrics for these security events?

network/src/noise/handshake.rs
```diff
@@ -429,31 +385,23 @@ impl NoiseUpgrader {
...
             client_timestamp.copy_from_slice(&payload);
             let client_timestamp = u64::from_le_bytes(client_timestamp);
 
             // check the timestamp is not a replay
             let mut anti_replay_timestamps = anti_replay_timestamps.write().map_err(|_| {
-                io::Error::new(
-                    io::ErrorKind::Other,
-                    "noise: unable to read anti_replay_timestamps lock",
-                )
+                NoiseHandshakeError::LockPoisoned(remote_peer_short, "anti_replay_timestamps")
```
gregnazario, line 399: Sorry, I'm making you iterate on this, but let's just do `expect` for poisoned locks, otherwise we have to take manual intervention when it dies.

bmwill: Poisoned locked? IMO Noise shouldn't even produce an Error like this and should just `unwrap` it

network/src/noise/error.rs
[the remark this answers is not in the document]
gregnazario: Maybe I wasn't being clear, but I'm saying if I'm some operator, and I see `"unauthenticated peer id"`.  I would be thinking, does this mean the peer id has to be authenticated with some key etc.  Rather than the client is connecting with an `unrecognized`, or `unexpected` peer id.

Similar to the `"unauthenticated public key"`, how do I authenticate a `public key`, when it's a recognized peer.  It makes more sense to be a `unrecognized` or `unexpected` public key.  Especially, since the authentication would be done with that public key.

network/src/noise/handshake.rs
```diff
@@ -167,7 +166,7 @@ impl NoiseUpgrader {
         socket: TSocket,
         origin: ConnectionOrigin,
         remote_public_key: Option<x25519::PublicKey>,
-    ) -> io::Result<(x25519::PublicKey, NoiseStream<TSocket>)>
+    ) -> Result<(x25519::PublicKey, NoiseStream<TSocket>), NoiseHandshakeError>
```
mimoo, line 169: 👍

network/src/noise/handshake.rs
```diff
@@ -177,12 +176,7 @@ impl NoiseUpgrader {
                 let remote_public_key = match remote_public_key {
                     Some(key) => key,
                     None if cfg!(any(test, feature = "fuzzing")) => unreachable!(),
-                    None => {
-                        return Err(std::io::Error::new(
-                            std::io::ErrorKind::Other,
-                            "noise: SHOULD NOT HAPPEN: missing server's key when dialing",
-                        ));
-                    }
+                    None => return Err(NoiseHandshakeError::MissingRemotePublicKey),
```
mimoo, line 179: can it be `MissingServerPublicKey`?

network/src/noise/error.rs
```diff
@@ -0,0 +1,99 @@
...
+    ServerWriteFailed(ShortHexStr, io::Error),
+}
+
+impl NoiseHandshakeError {
+    /// Errors that are either clear bugs or indicate some security issue. Should
+    /// immediately alert an engineer if we hit one of these errors.
+    pub fn should_security_log(&self) -> bool {
+        use NoiseHandshakeError::*;
+        match self {
+            MissingRemotePublicKey
```
mimoo, line 92: that shouldn't be a security log, but a MUST_FIX log, we should really build detection on MUST_FIX type of logs : o

network/src/noise/handshake.rs
```diff
@@ -249,49 +243,45 @@ impl NoiseUpgrader {
                 Some(&payload),
                 &mut client_noise_msg,
             )
-            .map_err(|e| io::Error::new(io::ErrorKind::Other, e))?;
+            .map_err(NoiseHandshakeError::BuildClientHandshakeMessageFailed)?;
```
mimoo, line 246: why the "build" here?

network/src/noise/handshake.rs
```diff
@@ -249,49 +243,45 @@ impl NoiseUpgrader {
...
             self.network_context,
-            remote_public_key
+            remote_public_key,
         );
-        socket.write_all(&client_message).await?;
-        socket.flush().await?;
+        socket
+            .write_all(&client_message)
+            .await
+            .map_err(NoiseHandshakeError::ClientWriteFailed)?;
```
mimoo, line 257: crazy idea, have a `NoiseHandshakeError::IO(std::io::Error)` type for these

network/src/transport/mod.rs
AnomalRoil: This should be enough, IMO. 
We should have metrics and even alerts depending on the security log, not sure any of this is implemented yet. But for a Handshake failure, we probably don't need alerts unless we are seeing a lot of them, hence the need for metrics, I'd say.

mimoo: cc @AnomalRoil 

what about:

* ClientExpectingDifferentPubkey
* ServerParseClient

At the end of the day, perhaps it really doesn't matter if someone connects to us and makes a mistake? We're supposed to address all edge-cases, we might just want to sample on any type of error and alert when we see too many errors of the same type?

network/src/noise/handshake.rs
gregnazario: `expect` or `unwrap` works, something where we need to restart anyways

bmwill: This is a case where `unwrap` is probably preferred, just so that you don't have to provide a generic "poisoned lock" message which doesn't add any value.

network/src/noise/error.rs
phlip9: that's a good point

@gregnazario do we have any way to alert on logs? like "data.error: 'MUST_FIX'" => alert in kibana or something?

network/src/transport/mod.rs
phlip9: I'm actually wondering now, what is a good criteria for security logging? Some of these error types are, as @mimoo pointed out, MUST_FIX programmer errors, which should hopefully get caught in CI. Other than that, there's no clear "this error case can _only_ be a malicious attack" except for maybe `ServerReplayDetected`. We've seen pretty much all other cases as a result of configuration error, and I don't know if I want to spam security-logging when someone doesn't configure their validator properly.

Actually, we probably want different policies here for validator vs vfn... We can definitely be more aggressive about security-logging for validators (the network address shouldn't be known, so this could be early signal that it leaked?) vs. vfn's we probably don't want to security-log anything unless it's a super clear vuln/issue

network/src/noise/handshake.rs
phlip9: we're constructing the client message (vs sending it idk) : p

network/src/noise/handshake.rs
phlip9: I want a more specific message, especially for the ClientRead case, which is like 90% of our errors. for the server errors, the read doesn't know the client peerid yet while the write and flush do

network/src/noise/error.rs
gregnazario: Yeah, but it would have to be a low level alert.  Probably better to make a metric for it, and review it periodically.  If this were to happen, you still have whatever the next release cycle is to get the code out, so the alarm may go off a bunch before then.

It sounds like you want `critical` back haha.

network/src/transport/mod.rs
gregnazario: Basically what you want is a field that is the same as `unreachable` but doesn't panic.

We could create a `Schema` that all it does is add the field `MUST_FIX=true` or whatever which would keep consistency across these types of shouldn't happen situations.

Now, this is one where we definitely don't want to panic, so I do support having it just an error.  But, if you do see this log, what would you do?

network/src/noise/error.rs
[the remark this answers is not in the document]
davidiw: would be good to clarify here why this is an error... such that the network is expecting only authenticated connections or that the key is unknown for this peer id

network/src/noise/error.rs
```diff
@@ -0,0 +1,99 @@
...
+            | BuildClientHandshakeMessageFailed(_)
+            | LockPoisoned(_, _)
```
davidiw, line 94: let's kill all these lock poisoning and replace them with unwraps per coding guidelines. (in another PR)

network/src/transport/mod.rs
```diff
@@ -217,16 +221,19 @@ async fn upgrade_inbound<T: TSocket>(
...
-            network_address = addr,
-            "security: InvalidNetworkPeer {}: '{}'",
-            addr,
-            err
-        );
-        err
+        if err.should_security_log() {
+            sample!(
+                SampleRate::Duration(Duration::from_secs(15)),
+                error!(
```
davidiw, line 227: In my opinion, an error should be reserved for internal issues that need attention, this definitely feels more like a warn or maybe an info.

network/src/transport/mod.rs
```diff
@@ -286,7 +287,22 @@ async fn upgrade_outbound<T: TSocket>(
     let mut socket = ctxt
         .noise
         .upgrade_outbound(socket, remote_pubkey, AntiReplayTimestamps::now)
-        .await?;
+        .await
+        .map_err(|err| {
+            if err.should_security_log() {
+                sample!(
+                    SampleRate::Duration(Duration::from_secs(15)),
+                    error!(
```
davidiw, line 295: same as above

davidiw (approved):
I think there's some room for improvement on policies for logging on different types of networks, but this is seems to address all of the major concerns and then we just pile concerns on existing issues.

and honestly ... I just had to open up someone else's code to understand what an "invalid license id" meant.... so even if I think we could make clearer error messages... these blow that type of nonsense out of the water :P

network/src/transport/mod.rs
mimoo: Yeah I think you're right.
We should probably rename security logs -> byzantine logging
If we do this, we can see that noise errors are probably not security logs, but probably something you still want to monitor/alert on at the association level

phlip9:
take two...

/land

PR #6368, opened 2020-10-04
[libra-framework] Cleanup of Event, FixedPoint32, LibraTimestamp, and Roles

wrwg:
This cleans and reviews specs and documentation of the mentioned modules.

- Event was totally mocked out, not only emitting events, but also creating generators and handlers. I tried for a while to bring it back at least partially, but that causes issues in many specs because of  the new aborts introduced via `Event::publish_generator` and `Event::new_handler` which are ubiquitous in the framework. So it stays mocked out  but has been cleaned up.
- Roles had a lot of mentioning of concepts in the documentation which we removed a while ago (specifically privileges). Good example why we absolutely have to do this cleanup everywhere.

## Motivation

Cleanup

Yes

## Test Plan

NA

## Related PRs

NA

### DIFF
--- a/language/stdlib/modules/Event.move
+++ b/language/stdlib/modules/Event.move
@@ -1,43 +1,42 @@
 address 0x1 {
 
-/**
-The Event module defines an `EventHandleGenerator` that is used to create
-`EventHandle`s with unique GUIDs. It contains a counter for the number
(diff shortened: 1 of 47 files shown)


emmazzz (approved):
LGTM

PR #6374, opened 2020-10-05
logger: have separate filters for local and remote logging

bmwill:


common/logger/src/libra_logger.rs
```diff
@@ -254,6 +295,7 @@ struct LoggerService {
     receiver: Receiver<LogEntry>,
     address: Option<String>,
     printer: Option<Box<dyn Writer>>,
+    facade: Arc<LibraLogger>,
```
gregnazario, line 298: Why not just call it `logger`?

common/debug-interface/src/node_debug_service.rs
```diff
@@ -29,21 +29,43 @@ impl NodeDebugService {
...
+                    logger.set_remote_filter(Filter::builder().parse(filter).build());
                 }
 
                 warp::reply::reply()
             });
 
+        // Post /log
+        let log = warp::post()
+            .and(warp::path("log"))
+            .and(local_filter.or(remote_filter));
```
gregnazario, line 67: So this does both?

common/logger/src/libra_logger.rs
```diff
@@ -132,6 +134,11 @@ impl LibraLoggerBuilder {
         self
     }
 
+    pub fn remote_level(&mut self, level: Level) -> &mut Self {
+        self.central_level = level;
+        self
```
gregnazario, line 139: Could we use the same terms for the `level` / `central_level` with the `local_level` / `remote_level`?

gregnazario (commented):
It looks like everything should work correctly.  I'm just mostly concerned about settling on 2 names for everything here.  What I've got are there two types:
* `local`, `printer`, `level`
* `remote`, `writer`, `central_level`

I'd prefer we just say `local_level` and `remote_level` so it's consistent.  Or going with printer & writer everywhere.

gregnazario (approved):

common/debug-interface/src/node_debug_service.rs
bmwill: `/log/filter` does the local and `/log/remote-filter` does the remote part

common/logger/src/libra_logger.rs
bmwill: oops thanks for catching this

common/debug-interface/src/node_debug_service.rs
gregnazario: Oh, I just wasn't sure what `filter.or(filter)` did both or what.

common/logger/src/libra_logger.rs
bmwill: to be more clear since there's technically two different parts of the logger, the Service and the facade

PR #6383, opened 2020-10-05
[libra framework] remove unused Debug and Compare modules from Libra …

sblackshear:
…Framework

Both of these modules are useful, but not called by any current transaction scripts. This PR moves them to a nursery/ directory under stdlib/ that is not compiled or published alongside the other Libra Framework modules.
This moves the tests for `Compare` and `Debug` to Move CLI expected value tests to avoid the headache of maintaining multiple genesis blobs in the stdlib. In addition, this tests the exact output of `Debug`, whereas before the tests only made sure it didn't crash.

This will delete the copies of the `Compare` and `Debug` modules currently on-chain, but this is not a client or validator software-breaking change because this code is dead. The deletion can be performed via a writeset.


wrwg (approved):

tzakian (approved):
Awesome! I think you might need to regenerate the stdlib though to remove these files from the docs, and to update the module ordering names in the compiled stdlib.

PR #6385, opened 2020-10-05
[docker] fix typo in docker-compose command

brianstrauch:
## Motivation

The README uses `docker-compose -up` when it should be `docker-compose up`


davidiw (approved):
good catch!

/land

Issue #6218, opened 2020-09-25
[network] Observability Deep Dive followup


gregnazario:
## Overview
These are some tasks as followup from the deep dive on 9/21.  I have them in general priority order

### Observability
- [x] Error metrics on important errors (I forgot to do this on the first pass) (@gregnazario will integrate with logs)
- [x] Improve error and log message to be more clear
- [x] Histograms on aggregate metrics (done on latencies)
- [x] Periodic state that needs to be sampled (https://github.com/libra/libra/pull/6387 done for eligible peers and connected peers)
- [ ] Per application throughput metrics, e.g. Consensus or Mempool throughput rather than DirectSend / RPC throughput alone.

### Hardening
- [x] Limit number of connections / throughput for clients
- [ ] Possibly look into how we can identify traffic patterns / users across multiple connections to help better implement the limit

### Testing
- [ ] Injected failure testing, to simulate connections that drop every packet, network partitions etc.
- [x] Fuzzing (low priority) on parts that we are currently missing it on

### Runbooks and Alarming
- [x] Alarms for operators, probably no specifics for blockchain oncall
- [x] Runbooks that cover other scenarios e.g. network partitions, flaky connections that drop every other packet, increased latency

### Tooling
- [x] Noise handshake checker (https://github.com/libra/libra/pull/6415)
- [x] Network config checker (this would have to be per infrastructure, so it's a lot of work unless we ignore per cloud specifics)

PR #6387, opened 2020-10-06
[network] Sample eligible peers used for connections

gregnazario:
This allows us to see what the current state for a node is in thinking
which peers it should be able to connect to.  It will help with
debugging on why a node is rejecting peers from connecting.

As the last sample change I will make for https://github.com/libra/libra/issues/6218 with the connected peers one I made earlier.

Will print out in the form of
```
2020-10-06T18:30:15.014306Z [tokio-runtime-worker] INFO network/src/connectivity_manager/mod.rs:486 Current eligible peers {"eligible_addrs":"{fc6a4571: [[/ip4/127.0.1.1/tcp/8080], []]}","eligible_keys":"{fc6a4571: [{}, {[X25519 public key: bc01d55dca171aea242e1a2d6be28a7975c46407331cd8478167138605122863]}]}","network_context":{"network_id":"Validator","peer_id":"07ff66cbd17377603e329855f3d17582","role":"validator"}}
```

### DIFF
--- a/network/src/connectivity_manager/mod.rs
+++ b/network/src/connectivity_manager/mod.rs
@@ -483,6 +483,16 @@ where
             self.network_context
         );
 
+        // Log the eligible peers with addresses from discovery
+        sample!(SampleRate::Duration(Duration::from_secs(60)), {
+            info!(
+                NetworkSchema::new(&self.network_context),
+                eligible_addrs = ?self.peer_addrs,
+                eligible_keys = ?self.peer_pubkeys,
+                "Current eligible peers"
+            )
+        });
+
         // Cancel dials to peers that are no longer eligible.
         self.cancel_stale_dials().await;
         // Disconnect from connected peers that are no longer eligible.
--- a/network/src/connectivity_manager/mod.rs
+++ b/network/src/connectivity_manager/mod.rs
@@ -555,7 +555,6 @@ where
 
         // Keep track of if any peer's addresses have actually changed, so we can
         // log without too much spam.
-        let mut have_any_changed = false;
         let self_peer_id = self.network_context.peer_id();
 
         // 3. add or update intersection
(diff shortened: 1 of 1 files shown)


bmwill (approved):

network/src/connectivity_manager/mod.rs
```diff
@@ -483,6 +483,17 @@ where
...
         );
 
+        // Log the eligible peers state periodically so we see what our current state is
+        sample!(SampleRate::Duration(Duration::from_secs(60)), {
+            let eligible = self.eligible.read().unwrap().clone();
+            let elig_peers: Vec<_> = eligible.keys().collect();
+            info!(
+                NetworkSchema::new(&self.network_context),
+                peers = ?elig_peers,
+                "Current eligible peers"
```
davidiw, line 493: why not print the values too? I think we expect this to be pretty static but the network addresses could change

gregnazario: Yeah, we could stick the addresses in too.

network/src/connectivity_manager/mod.rs
```diff
@@ -483,6 +483,17 @@ where
             self.network_context
         );
 
+        // Log the eligible peers state periodically so we see what our current state is
+        sample!(SampleRate::Duration(Duration::from_secs(60)), {
```
phlip9, line 487: We can probably ramp the sample rate down to every 5 minutes or something

network/src/connectivity_manager/mod.rs
phlip9: the source of truth is pretty much `self.peer_addrs`, which covers the peer ids, the network addresses, and the public keys (inside the network addresses). so we could probably just dump that whole thing.

gregnazario: So, now I'm confused.  Why do we have `peer_keys` and `eligible` if they're the same but one has a lock?  Does one just not include the self address?

Yeah, I can change it to just dump the network addresses.

gregnazario: I've changed it to print `peer_keys` and `peer_addrs` see overview

network/src/connectivity_manager/mod.rs
gregnazario: I'd rather err on logging more often than less for now.  Especially since I removed the preprocessing, it should be much more lightweight.

PR #6391, opened 2020-10-06
[libra framework] Specified the access control properties of the txn scripts

junkil-park:
- added the access control specs for the P1/P2/P3 transaction scripts

## Motivation

To specify the access control specs of the transaction scripts

Yes

## Test Plan

cargo test

### DIFF
--- a/language/stdlib/transaction_scripts/rotate_authentication_key_with_nonce_admin.move
+++ b/language/stdlib/transaction_scripts/rotate_authentication_key_with_nonce_admin.move
@@ -43,6 +43,7 @@ fun rotate_authentication_key_with_nonce_admin(lr_account: &signer, account: &si
 spec fun rotate_authentication_key_with_nonce_admin {
     use 0x1::Signer;
     use 0x1::Errors;
+    use 0x1::Roles;
 
     include LibraAccount::TransactionChecks{sender: account}; // properties checked by the prologue.
     let account_addr = Signer::spec_address_of(account);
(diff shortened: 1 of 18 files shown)


language/stdlib/transaction_scripts/rotate_authentication_key_with_nonce_admin.move
```diff
@@ -60,6 +61,8 @@ spec fun rotate_authentication_key_with_nonce_admin {
         Errors::NOT_PUBLISHED; // TOOD: Undocumented error code. Added due to the possible absence of SlidingNonce in SlidingNonce::try_record_nonce;
 
     /// Access Control
+    /// Only the Libra Root account can process the admin scripts [[H8]][PERMISSION].
+    requires Roles::has_libra_root_role(lr_account); /// This is ensured by LibraAccount::writeset_prologue.
```
sblackshear, line 65: Would it make more sense to require the entire `spec fun` for `writeset_prologue` here? That would make the comment unnecessary + let the script assume other properties that are true and may be needed in the future.

language/stdlib/transaction_scripts/set_validator_operator_with_nonce_admin.move
```diff
@@ -80,5 +82,11 @@ spec fun set_validator_operator_with_nonce_admin {
         Errors::INVALID_ARGUMENT,
         Errors::NOT_PUBLISHED,
         Errors::REQUIRES_ROLE;
-    }
+
+    /// Access Control
+    /// Only the Libra Root account can process the admin scripts [[H8]][PERMISSION].
+    requires Roles::has_libra_root_role(lr_account); /// This is ensured by LibraAccount::writeset_prologue.
```
sblackshear, line 88: Same question re: prologue here.

sblackshear (approved):
This is excellent, thank you!!

language/stdlib/transaction_scripts/rotate_authentication_key_with_nonce_admin.move
junkil-park: That's a good point. Thanks to @emmazzz , each txn script already has the following line in order to "require" other properties that the prologue function ensures:
```
include LibraAccount::TransactionChecks{sender: account}; // properties checked by the prologue.
```

language/stdlib/transaction_scripts/set_validator_operator_with_nonce_admin.move
junkil-park: Same as above!

language/stdlib/transaction_scripts/rotate_authentication_key_with_nonce_admin.move
sblackshear: The problem is that admin scripts *do not* run the ordinary transaction prologue, so it is not correct to assume this. Instead, admin scripts should assume that the writeset prologue is run.

PR #6392, opened 2020-10-06
add vm script hash allow list to get_metadata response

xli:
## Motivation

VM script hash allow list is useful for client to understand what scripts are allowed by validators. 
Get metadata returns allow list for client to check local scripts. Client SDK libraries will use this to validate stdlib scripts and ensure released version is compatible with premainnet and testnet.

yes

## Test Plan

Unit tests and Integration test

## Related PRs

### DIFF
--- a/json-rpc/types/src/views.rs
+++ b/json-rpc/types/src/views.rs
@@ -348,6 +348,8 @@ pub struct BlockMetadata {
     pub version: u64,
     pub timestamp: u64,
     pub chain_id: u8,
+    pub script_hash_allow_list: Option<Vec<BytesView>>,
+    pub module_publishing_allowed: Option<bool>,
 }
 
 #[derive(Clone, Serialize, Deserialize, Debug, PartialEq)]
--- a/json-rpc/src/methods.rs
+++ b/json-rpc/src/methods.rs
@@ -5,8 +5,8 @@
 use crate::{
     errors::JsonRpcError,
     views::{
-        AccountStateWithProofView, AccountView, BlockMetadata, CurrencyInfoView, EventView,
-        StateProofView, TransactionView,
+        AccountStateWithProofView, AccountView, BlockMetadata, BytesView, CurrencyInfoView,
+        EventView, StateProofView, TransactionView,
     },
 };
 use anyhow::{ensure, format_err, Error, Result};
@@ -25,6 +25,7 @@ use libra_types::{
     event::EventKey,
     ledger_info::LedgerInfoWithSignatures,
     mempool_status::MempoolStatusCode,
+    on_chain_config::VMPublishingOption,
     transaction::SignedTransaction,
 };
 use network::counters;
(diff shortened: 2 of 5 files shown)


libra-action:

Breaking changes policy:

2. Adding fields is ok.

types/src/on_chain_config/vm_publishing_option.rs
```diff
@@ -4,14 +4,15 @@
...
 use serde::{Deserialize, Serialize};
+use std::default::Default;
 
 /// Defines and holds the publishing policies for the VM. There are three possible configurations:
 /// 1. No module publishing, only allowlisted scripts are allowed.
 /// 2. No module publishing, custom scripts are allowed.
 /// 3. Both module publishing and custom scripts are allowed.
 /// We represent these as an enum instead of a struct since allowlisting and module/script
 /// publishing are mutually exclusive options.
-#[derive(Clone, Debug, Deserialize, PartialEq, Serialize)]
```
runtian-zhou, line 14: Can we avoid this default by changing the previous api to be `Option<VMPublishingOption>`?

json-rpc/types/src/views.rs
```diff
@@ -339,6 +339,8 @@ pub struct BlockMetadata {
     pub version: u64,
     pub timestamp: u64,
     pub chain_id: u8,
+    pub script_hash_allow_list: Vec<BytesView>,
```
runtian-zhou, line 342: In general I'm a little bit hesitant to put those items into `BlockMetadata`, the reason is that it will make its definition start to deviate from the original representation of its Move definition in https://github.com/libra/libra/blob/f5f8fa1769cd3cf04cb669b19cb97cf75540915e/language/stdlib/modules/LibraBlock.move#L10. I feel like there is a need for fetching all runtime critical on-chain configs but squashing those different resources might bring some confusion and will harm the future upgrade path in general.

thefallentree: the view here should not be named BlockMetaData, it is not related, maybe we should just rename it to MetadataView instead

sblackshear: In addition, the `script_hash_allowlist` is pretty big (34 32-byte hashes and growing), which may bloat this view...

I wonder if it would make sense to expose different structs for fairly static metadata (e.g., `chain_id`, allowlist, set of allowed currencies) and highly dynamic metadata (e.g., version, timestamp)?

xli: we return chain_id, version & timestamp in every response body, so technically, there is no need to call get_metadata for these information anymore.
Instead of keeping current get_metadata useless, adding onchain configuration is not bad.

I consider it's metadata because the returned data can be used for setup client. 
So splitting into multiple endpoints may not really helpful for client.

runtian-zhou: If that's the case `ConfigurationView` sounds like a better name to me. The point was that the `BlockMetadata` is a confusing name especially with the data we are trying to put over there.

xli: BlockMetadata is hidden to client side, if you feel confusing, I can rename it to something else, like MetadataView.
this is doc for the api: https://github.com/libra/libra/blob/master/json-rpc/docs/method_get_metadata.md

it's blockchain / ledger metadata.

add a new "XxxView" does not mean client side sees same name.
Also, for every server side concept, I don't think client need to understand them, most global static stuff can be inside get metadata response.

runtian-zhou (approved):
LGTM! Will rename `BlockMetadataView` in a separate PR

PR #6394, opened 2020-10-06
[Operational Tooling] Add smoke tests for operational tool commands.

JoshLind:
## Motivation

Note: this PR is based on PR: https://github.com/libra/libra/pull/6386. As such, to review the PR, ignore the first 3 commits.

This PR adds missing smoke tests for the various commands offered by the operational tool. This should help to ensure that the operational tool is not silently broken or seeing regressions. The PR offers 7 commits, each of which adds a smoke test (or two) for the different commands being tested by the commit:
1. print-account
2. account-resource
3. extract-public-key
4. extract-private-key
5. validate-transaction
6. set-validator-config and validator-config
7. validator-set

Note, there's still a few remaining commands that are not yet tested. These will be tested in a follow up PR (to be submitted shortly!)

Yes.

## Test Plan

All tests pass locally.

## Related PRs

This PR is based on PR: https://github.com/libra/libra/pull/6386. As such, PR https://github.com/libra/libra/pull/6386 needs to land first.

### DIFF
--- a/testsuite/smoke-test/src/operational_tooling.rs
+++ b/testsuite/smoke-test/src/operational_tooling.rs
@@ -14,7 +14,10 @@ use libra_types::{
     account_address::AccountAddress, chain_id::ChainId,
     transaction::authenticator::AuthenticationKey,
 };
-use std::convert::{TryFrom, TryInto};
+use std::{
+    convert::{TryFrom, TryInto},
+    fs,
+};
 
 #[test]
 fn test_account_resource() {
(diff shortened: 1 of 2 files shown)


testsuite/smoke-test/src/operational_tooling.rs
```diff
@@ -22,7 +24,51 @@ use libra_types::{
...
+
+    // Load a node config
+    let node_config =
+        NodeConfig::load(swarm.validator_swarm.config.config_files.first().unwrap()).unwrap();
+
+    // Connect the operator tool to the first node's JSON RPC API
+    let op_tool = swarm.get_op_tool(0);
+
+    // Load validator's on disk storage
+    let backend = load_backend_storage(&node_config);
```
davidiw, line 46: can we make this a function instead? just to get a backend or a storage for a given validator node?

testsuite/smoke-test/src/operational_tooling.rs
```diff
@@ -69,6 +117,243 @@ fn test_consensus_key_rotation() {
     assert_eq!(rotated_consensus_key, new_consensus_key);
 }
 
+#[test]
+fn test_e2e_reconfiguration() {
```
davidiw, line 121: please delete this... please...

davidiw (approved):

testsuite/smoke-test/src/operational_tooling.rs
JoshLind: Done in the previous PR 😄

testsuite/smoke-test/src/operational_tooling.rs
JoshLind: Yeah, I actually have a bigger refactor that I'm planning to do which will take care of all these annoying repeated patterns. So I'm gonna land this for now, and then in the next PR use the refactor 😄

PR #6399, opened 2020-10-06
[JSON-RPC] Adding a get_transactions_with_proofs API

AnomalRoil:
This is necessary in order to have a verifying client able to not trust a full node, and verifying all the data provided by the full node.

### DIFF
--- a/json-rpc/docs/method_get_transactions_with_proofs.md
+++ b/json-rpc/docs/method_get_transactions_with_proofs.md
@@ -1,0 +1,37 @@
+## Method get_transactions_with_proofs
+
+**Description**
+
+Get the transactions on the blockchain along with the proofs necessary to verify the said transactions.
+
+
+### Parameters
+
+| Name           | Type           | Description                                                          |
+|----------------|----------------|----------------------------------------------------------------------|
+| start_version  | unsigned int64 | Start on this transaction version for this query                     |
+| limit          | unsigned int64 | Limit the number of transactions returned, the max value is 1000     |
+
+### Returns
+
+
+
+| Name                      | Type               | Description                   |
+|---------------------------|--------------------|-------------------------------|
+| serialized_transactions   | List<string>       | An array of hex encoded strings with the raw bytes of the returned `Transaction` |
+| proofs                    | TransactionsProofs | The proofs, see below.   |
+
+
+The proofs:
+
+
+| Name           | Type           | Description                                                          |
+|----------------|----------------|----------------------------------------------------------------------|
+| ledger_info_to_transaction_infos_proof  | string | An hex encoded string of raw bytes of a `Vec<AccumulatorRangeProof<TransactionAccumulatorHasher>>` that contains the proofs of the returned transactions |
+| transaction_infos          | string | An hex encoded string of raw bytes of a `Vec<TransactionInfo>` that corresponds to returned transcations    |
+
+Notice that all raw bytes encoded strings are containing LCS encoded data.
+
+Also, please notice that you need the ledger_info at the time of the request in order to have the correct accumulator_hash to verify the ledger_info_to_transaction_infos_proof produced at that time, so you should do a batched call to `get_state_proof` whenever you call `get_transactions_with_proofs`.
+
+In order to see an example of how to verify the proofs, please refer to the [integration tests](json-rpc/tests/integration_test.rs).
--- a/json-rpc/src/methods.rs
+++ b/json-rpc/src/methods.rs
@@ -6,7 +6,8 @@ use crate::{
     errors::JsonRpcError,
     views::{
         AccountStateWithProofView, AccountView, BytesView, CurrencyInfoView, EventView,
-        MetadataView, StateProofView, TransactionView,
+        MetadataView, StateProofView, TransactionView, TransactionsProofsView,
+        TransactionsWithProofsView,
     },
 };
 use anyhow::{ensure, format_err, Error, Result};
@@ -393,6 +394,43 @@ async fn get_transactions(
     Ok(result)
 }
 
+/// Returns transactions by range with proofs
+async fn get_transactions_with_proofs(
+    service: JsonRpcService,
+    request: JsonRpcRequest,
+) -> Result<Option<TransactionsWithProofsView>> {
+    let start_version: u64 = request.parse_param(0, "start_version")?;
+    let limit: u64 = request.parse_param(1, "limit")?;
+
+    // Notice limit is a u16 normally, but some APIs require u64 below
+    service.validate_page_size_limit(limit as usize)?;
+
+    if start_version > request.version() {
+        return Ok(None);
+    }
+
+    // We do not fetch events since they don't come with proofs.
+    let txs = service
+        .db
+        .get_transactions(start_version, limit, request.version(), false)?;
+
+    let mut blobs = vec![];
+    for t in txs.transactions.iter() {
+        let bv = lcs::to_bytes(t)?;
+        blobs.push(BytesView::from(bv));
+    }
+
+    let (proofs, tx_info) = txs.proof.unpack();
+
+    Ok(Some(TransactionsWithProofsView {
+        serialized_transactions: blobs,
+        proofs: TransactionsProofsView {
+            ledger_info_to_transaction_infos_proof: BytesView::from(&lcs::to_bytes(&proofs)?),
+            transaction_infos: BytesView::from(&lcs::to_bytes(&tx_info)?),
+        },
+    }))
+}
+
 /// Returns account transaction by account and sequence_number
 async fn get_account_transaction(
     service: JsonRpcService,
@@ -620,7 +658,9 @@ pub(crate) fn build_registry() -> RpcRegistry {
     );
     register_rpc_method!(registry, "get_events", get_events, 3, 0);
     register_rpc_method!(registry, "get_currencies", get_currencies, 0, 0);
+    register_rpc_method!(registry, "get_network_status", get_network_status, 0, 0);
 
+    // Proof APIs
     register_rpc_method!(registry, "get_state_proof", get_state_proof, 1, 0);
     register_rpc_method!(
         registry,
@@ -629,7 +669,13 @@ pub(crate) fn build_registry() -> RpcRegistry {
         3,
         0
     );
-    register_rpc_method!(registry, "get_network_status", get_network_status, 0, 0);
+    register_rpc_method!(
+        registry,
+        "get_transactions_with_proofs",
+        get_transactions_with_proofs,
+        2,
+        0
+    );
 
     registry
 }
--- a/json-rpc/src/tests/unit_tests.rs
+++ b/json-rpc/src/tests/unit_tests.rs
@@ -477,6 +477,66 @@ fn test_json_rpc_protocol_invalid_requests() {
             }),
         ),
         (
+            "get_transactions_with_proofs: invalid start_version param",
+            json!({"jsonrpc": "2.0", "method": "get_transactions_with_proofs", "params": ["helloworld", 1], "id": 1}),
+            json!({
+                "error": {
+                    "code": -32602,
+                    "message": "Invalid param start_version(params[0]): should be unsigned int64",
+                    "data": null
+                },
+                "id": 1,
+                "jsonrpc": "2.0",
+                "libra_chain_id": ChainId::test().id(),
+                "libra_ledger_timestampusec": timestamp,
+                "libra_ledger_version": version
+            }),
+        ),
+        (
+            "get_transactions_with_proofs: start_version is too big, returns empty array",
+            json!({"jsonrpc": "2.0", "method": "get_transactions_with_proofs", "params": [version+1, 1], "id": 1}),
+            json!({
+                "id": 1,
+                "jsonrpc": "2.0",
+                "libra_chain_id": ChainId::test().id(),
+                "libra_ledger_timestampusec": timestamp,
+                "libra_ledger_version": version,
+                "result": null
+            }),
+        ),
+        (
+            "get_transactions_with_proofs: invalid limit param",
+            json!({"jsonrpc": "2.0", "method": "get_transactions_with_proofs", "params": [1, false], "id": 1}),
+            json!({
+                "error": {
+                    "code": -32602,
+                    "message": "Invalid param limit(params[1]): should be unsigned int64",
+                    "data": null
+                },
+                "id": 1,
+                "jsonrpc": "2.0",
+                "libra_chain_id": ChainId::test().id(),
+                "libra_ledger_timestampusec": timestamp,
+                "libra_ledger_version": version
+            }),
+        ),
+        (
+            "get_transactions_with_proofs: limit is too big",
+            json!({"jsonrpc": "2.0", "method": "get_transactions_with_proofs", "params": [1, 1001], "id": 1}),
+            json!({
+                "error": {
+                    "code": -32600,
+                    "message": "Invalid Request: page size = 1001, exceed limit 1000",
+                    "data": null
+                },
+                "id": 1,
+                "jsonrpc": "2.0",
+                "libra_chain_id": ChainId::test().id(),
+                "libra_ledger_timestampusec": timestamp,
+                "libra_ledger_version": version
+            }),
+        ),
+        (
             "get_events: invalid event_key type",
             json!({"jsonrpc": "2.0", "method": "get_events", "params": [false, 1, 10], "id": 1}),
             json!({
--- a/json-rpc/tests/integration_test.rs
+++ b/json-rpc/tests/integration_test.rs
@@ -9,8 +9,10 @@ use libra_types::{
     access_path::AccessPath,
     account_address::AccountAddress,
     account_config::coin1_tmp_tag,
+    epoch_change::EpochChangeProof,
     ledger_info::LedgerInfoWithSignatures,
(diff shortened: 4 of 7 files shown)


libra-action:

Breaking changes policy:

2. Adding fields is ok.

AnomalRoil:
@mimoo @xli  What do you think?

json-rpc/src/methods.rs
```diff
@@ -502,6 +495,33 @@ async fn get_account_transactions(
...
 
+//
+// Proofs API
+//
+
+/// Returns transactions by range with proofs
+async fn get_transactions_with_proofs(
+    service: JsonRpcService,
+    request: JsonRpcRequest,
+) -> Result<Option<TransactionListWithProof>> {
```
xli, line 506: you need return view, add a test to see what is json looks like

json-rpc/src/methods.rs
```diff
@@ -502,6 +495,33 @@ async fn get_account_transactions(
...
+    let start_version: u64 = request.parse_param(0, "start_version")?;
+    let limit: u64 = request.parse_param(1, "limit")?;
+    let include_events: bool = request.parse_param(2, "include_events")?;
```
xli, line 509: if you include event, you need include events proof too

json-rpc/src/methods.rs
xli: need return onchain blob, instead of decoded Transaction data, imaging you are creating client, how you do the validation, you use blob to create hash and validate with proof, then decode the blob for the data.

xli (changes requested):

AnomalRoil:
@xli @mimoo Okay, so it seems MockDB is not implementing valid transactions with proofs, so I cannot verify them in the unit tests.

The code has changed a lot, please let me know what you think now.

json-rpc/src/tests/unit_tests.rs
```diff
@@ -1192,6 +1239,47 @@ fn test_get_transactions() {
...
+        for (i, blob) in received_proof.transactions_blobs.clone().iter().enumerate() {
+            let v = base_version + i as u64;
+            let rtx: Transaction = lcs::from_bytes(&blob.clone().into_bytes().unwrap()).unwrap();
+            let (tx, _status) = &mock_db.all_txns[v as usize];
+
+            assert_eq!(rtx.hash().to_hex(), tx.hash().to_hex());
+        }
+
+        // Notice that we cannot verify the proofs in the test since the MockDB is not implementing correct transactions.
+        // For the same reason we cannot test the proof's transaction_infos field since it is containing default valiues.
```
mimoo, line 1279: "valiues"

json-rpc/types/src/views.rs
```diff
@@ -464,6 +464,18 @@ pub struct TransactionView {
     pub vm_status: VMStatusView,
     pub gas_used: u64,
 }
+#[derive(Clone, Debug, Deserialize, Serialize, PartialEq)]
+pub struct TransactionsWithProofsView {
+    pub first_transaction_version: Option<u64>,
+    pub transactions_blobs: Vec<BytesView>,
```
mimoo, line 470: what's a transaction blob? My guess is that it's an LCS serialized transaction, in this case why not `serialized_transactions`?

json-rpc/types/src/views.rs
```diff
@@ -464,6 +464,18 @@ pub struct TransactionView {
...
+    pub proofs: TransactionsProofsView,
+}
+
+#[derive(Clone, Debug, Serialize, Deserialize, PartialEq)]
+pub struct TransactionsProofsView {
+    pub ledger_info_to_transaction_infos_proof: BytesView,
+    pub transaction_infos: BytesView,
```
mimoo, line 477: same here, are these LCS serialized transaction infos? If so then perhaps rename to `serialized_transaction_infos`?

json-rpc/types/src/views.rs
mimoo: (or just transactions)

AnomalRoil: It is, I just used the same terminology as for AccountStateWithProofView. I can rename it.

mimoo: but in the case of account state, the term account blob is the technical term we use : o

json-rpc/src/tests/unit_tests.rs
```diff
@@ -475,6 +475,50 @@ fn test_json_rpc_protocol_invalid_requests() {
...
+                    "message": "Invalid param limit(params[1]): should be unsigned int64",
+                    "data": null
+                },
+                "id": 1,
+                "jsonrpc": "2.0",
+                "libra_chain_id": ChainId::test().id(),
+                "libra_ledger_timestampusec": timestamp,
+                "libra_ledger_version": version
+            }),
+        ),
```
xli, line 521: could you add test case for limit is too big

json-rpc/src/tests/unit_tests.rs
```diff
@@ -1193,6 +1237,58 @@ fn test_get_transactions() {
...
+        .into_iter()
+    {
+        println!(
+            "Working on base_version {} out of {}",
+            base_version, version
+        );
+        let mut batch = JsonRpcBatch::default();
+        batch.add_get_transactions_with_proofs_request(base_version, page as u64);
+        let result = execute_batch_and_get_first_response(&client, &mut runtime, batch);
+        println!("No panic {}", base_version);
```
xli, line 1260: remove print

json-rpc/src/tests/unit_tests.rs
```diff
@@ -1193,6 +1237,58 @@ fn test_get_transactions() {
...
+
+        let received_proof = TransactionsWithProofsView::from_response(result).unwrap();
+
+        // if it's empty then no need to process it.
+        if received_proof.serialized_transactions.is_empty() {
```
xli, line 1265: this is not a good way to do test, what if all of responses' serialized_transactions is empty?

I think we can remove this test, add one test to json-rpc/tests/integration_test instead.
and the test should validate the proof works with respond transaction blobs

json-rpc/types/src/views.rs
```diff
@@ -465,6 +465,18 @@ pub struct TransactionView {
     pub vm_status: VMStatusView,
     pub gas_used: u64,
 }
+#[derive(Clone, Debug, Deserialize, Serialize, PartialEq)]
+pub struct TransactionsWithProofsView {
+    pub first_transaction_version: Option<u64>,
```
xli, line 470: I wonder what is the case we get first_transaction_version is None, could you write a test for it?

json-rpc/src/methods.rs
```diff
@@ -390,6 +391,52 @@ async fn get_transactions(
...
+    request: JsonRpcRequest,
+) -> Result<Option<TransactionsWithProofsView>> {
+    let start_version: u64 = request.parse_param(0, "start_version")?;
+    let limit: u64 = request.parse_param(1, "limit")?;
+
+    // Notice limit is a u16 normally, but some APIs require u64 below
+    service.validate_page_size_limit(
+        limit
+            .try_into()
+            .expect("Unexpected usize overflow by limit"),
```
xli, line 406: Do not use expect, which panics. map_err into requestError

json-rpc/src/tests/unit_tests.rs
```diff
@@ -475,6 +475,66 @@ fn test_json_rpc_protocol_invalid_requests() {
...
+                    "code": -32600,
+                    "message": "Invalid Request: page size = 1001, exceed limit 1000",
+                    "data": null
+                },
+                "id": 1,
+                "jsonrpc": "2.0",
+                "libra_chain_id": ChainId::test().id(),
+                "libra_ledger_timestampusec": timestamp,
+                "libra_ledger_version": version
+            }),
```
xli, line 536: Could you add one more case: limit number is bigger than usize, ensure we get invalid request param error

json-rpc/tests/integration_test.rs
```diff
@@ -751,6 +754,74 @@ fn create_test_cases() -> Vec<Test> {
                 }
             },
         },
+        Test {
+            name: "get_transactions_with_proofs",
+            run: |env: &mut testing::Env| {
+                let resp = env.send("get_metadata", json!([]));
```
xli, line 760: Assert libra_ledger_version > 0, so that we know the for loop at least run once

json-rpc/src/tests/unit_tests.rs
AnomalRoil: That would be bigger than u64 on 64 bits, and u128 makes the json! macro panic anyway, which makes it a bit difficult to test, no? 
IDK if we have 32 bits specific tests?

AnomalRoil: Postponed for now since usize is 64 bits on 64bits systems and we don't have 32 bits test systems AFAIK.

json-rpc/tests/integration_test.rs
```diff
@@ -858,6 +860,71 @@ fn create_test_cases() -> Vec<Test> {
...
+                    // we need to find a workaround to get the validator set. This works since it's a test
+                    // and the validator set is not changing, but it wouldn't work in practice.
+                    let state_data = env.send("get_state_proof", json!([base_version])).result.unwrap();
+                    let ep_cp = state_data["epoch_change_proof"].as_str().unwrap();
+                    let epoch_proofs:EpochChangeProof = lcs::from_bytes(&hex::decode(&ep_cp).unwrap()).unwrap();
+                    let latest_epoch = epoch_proofs.ledger_info_with_sigs.first().unwrap();
+                    assert!(li.verify_signatures(&latest_epoch.ledger_info().next_epoch_state().unwrap().verifier).is_ok());
+
+                    // and we eventually verify the proofs for the transactions
+                    assert!(li_to_tip.verify(expected_hash, Some(base_version), &hashes).is_ok());
```
xli, line 924: The expected_hash is also from server. Shouldn't we validate expected_hash first?

xli: also, you returned latest ledger info, but the hashes are from transactions base_version to limit, how could we use the latest ledger info for the case if limit is < ledger info.

json-rpc/tests/integration_test.rs
```diff
@@ -858,6 +860,71 @@ fn create_test_cases() -> Vec<Test> {
                 }
             },
         },
+        Test {
+            name: "get_transactions_with_proofs",
+            run: |env: &mut testing::Env| {
+                let resp = env.send("get_metadata", json!([]));
+                assert!(resp.libra_ledger_version > 0);
+                for base_version in 0..resp.libra_ledger_version {
```
xli, line 868: remove this for loop, be specific about what you want to test, set base_version, and I think we should test base_version + limit < resp.libra_ledger_version

AnomalRoil: I'm testing all the possible queries from version 0 to the libra_ledger_version. That's why I have the for loop. I guess I could replace it with just a test for the base_version = 5 or so, but it would cover less stuff, no?

I could add a test where base_version + limit > resp.libra_ledger_version and check it works as intended and returns less than limit txs.
But as long as base_version < resp.libra_ledger_version, this should still work I think. It's a limit, but if the limit is 400 and there are only 100 txs in the response because the ledger_version is equal to base_version+99, then that would still be a valid response, no?

json-rpc/tests/integration_test.rs
```diff
@@ -858,6 +860,71 @@ fn create_test_cases() -> Vec<Test> {
...
+                    assert!(!no_proof_txns.as_array().unwrap().is_empty());
+                    assert_eq!(no_proof_txns.as_array().unwrap().len(), actual_txs.len());
+                    for (index, tx) in no_proof_txns.as_array().unwrap().iter().enumerate() {
+                        assert_eq!(tx["hash"].as_str().unwrap(), actual_txs[index].hash().to_hex());
+                    }
+
+                    // We need to get the details required to verify the proof
+                    let li_raw = data["ledger_info"].as_str().unwrap();
+                    let li:LedgerInfoWithSignatures = lcs::from_bytes(&hex::decode(&li_raw).unwrap()).unwrap();
+                    let expected_hash = li.ledger_info().transaction_accumulator_hash();
```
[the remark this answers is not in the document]
AnomalRoil: If we don't return the whole ledger_info, we cannot verify that the ledger_info is valid.

json-rpc/tests/integration_test.rs
AnomalRoil: It's the case, expected_hash is validated by  `li.verify_signatures` in the assert above.

The latest ledger_info contains the necessary hash to prove the proofs that are generated at that ledger_info, but these proofs can be for any version < ledger_info

xli: one thing here, above "li.verify_signatures" validates signature, is it also validating transaction_accumulator_hash is the one expected?

AnomalRoil: Yes

json-rpc/src/methods.rs
```diff
@@ -393,6 +394,44 @@ async fn get_transactions(
...
+    for t in txs.transactions.iter() {
+        let bv = lcs::to_bytes(t)?;
+        blobs.push(BytesView::from(bv));
+    }
+
+    let (proofs, tx_info) = txs.proof.unpack();
+
+    Ok(Some(TransactionsWithProofsView {
+        serialized_transactions: blobs,
+        ledger_info: BytesView::from(&lcs::to_bytes(&request.ledger_info)?),
```
xli, line 427: We talked about this last time: we should consider returning ledger info with proof or just let client batch call with state_with_proof method.
I think in case we want batch call to get more txns with proof, or get account state with proof, maybe batch call is better.

json-rpc/tests/integration_test.rs
```diff
@@ -858,6 +860,82 @@ fn create_test_cases() -> Vec<Test> {
...
+                let some_li:Vec<_> = epoch_proofs.ledger_info_with_sigs;
+                assert!(!some_li.is_empty());
+                // Let's use the first one since the validator set does not change in the tests.
+                let validator_set = &some_li.first().unwrap().ledger_info().next_epoch_state().unwrap().verifier;
+
+                // We test 2 cases:
+                //      1. base_version + limit < resp.libra_ledger_version
+                //      2. base_version + limit > resp.libra_ledger_version
+                for base_version in &[0, resp.libra_ledger_version] {
+                    let limit = 50;
```
xli, line 885: I'm not sure case 1 is covered here, because I don't know whether server latest ledger version > 50

AnomalRoil: I can add another assert, but [the assert here](https://github.com/libra/libra/pull/6399/files#diff-f1a808a01f2b2b1b2b8b8ab82534890eff0039c5e41cd409ed37bc8d07a1619cR901-R902) ensure we have either 1 or 50 txs in the results, so it works. 
But I can duplicate the test if you prefer.

https://github.com/libra/libra/pull/6399/files#diff-f1a808a01f2b2b1b2b8b8ab82534890eff0039c5e41cd409ed37bc8d07a1619cR901-R902

json-rpc/src/methods.rs
AnomalRoil: How can you do a batch call currently to be sure that both reqwest's ledger_info would be the same?

AnomalRoil: Done

client/json-rpc/src/client.rs
```diff
@@ -71,6 +71,13 @@ impl JsonRpcBatch {
         );
     }
 
+    pub fn add_get_transactions_with_proofs_request(&mut self, start_version: u64, limit: u64) {
```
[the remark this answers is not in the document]
AnomalRoil: I'll do it in a follow-up PR :)

json-rpc/docs/method_get_transactions_with_proofs.md
```diff
@@ -0,0 +1,38 @@
...
+| limit          | unsigned int64 | Limit the number of transactions returned, the max value is 1000     |
+
+### Returns
+
+
+
+| Name                      | Type               | Description                   |
+|---------------------------|--------------------|-------------------------------|
+| serialized_transactions   | List<string>       | An array of hex encoded strings with the raw bytes of the returned `Transaction` |
+| ledger_info               | string             | An hex encoded string of raw bytes of the `LedgerInfoWithSignature` at the moment of the returned proofs |
```
xli, line 22: remove ledger_info

json-rpc/tests/integration_test.rs
xli: then add an assertion abort the for loop: assert resp.libra_ledger_version > limit
move limit = 50 above of the assert, 

also, I think we can change 50 to smaller number like 5 or 10, in case test is unreliable if limit is too much and server does not generate enough versions.

xli (approved):

client/json-rpc/src/response.rs
```diff
@@ -197,3 +202,13 @@ impl ResponseAsView for AccountStateWithProofView {
...
+
+impl ResponseAsView for TransactionsWithProofsView {
+    fn from_response(response: JsonRpcResponse) -> Result<Self> {
+        if let JsonRpcResponse::TransactionsWithProofsResponse(resp) = response {
+            Ok(resp)
+        } else {
+            Self::unexpected_response_error::<Self>(response)
+        }
+    }
+}
```
xli, line 214: could you remove changes to this response.rs file? I don't plan to continue maintain this client, I don't think we need add more to here, if someone uses it, it will increase migration cost.

PR #6400, opened 2020-10-06
[move-prover] Using new Boogie `{:print}` feature for obtaining model values

wrwg:
This PR completes switching to the new `{:print}` feature for extracting model values for prover diagnosis display which @shazqadeer added in Boogie version 2.7.35.

Previously, we used uninterpreted functions and assumptions over its mapping to collect the values being displayed. This added additional load to Z3. The new approach handles this entirely via Boogie, not bothering Z3.

The expected advantage is (a) faster and less butterflies (b) more precise; specifically values with 'dummy' assignments which Z3 actually doesn't need to create a counterexample should vanish. Those values previously appeared in the output with some `<? s-expr>` representation or some nonsense value.

All tests are passing. A very few diagnosis output has changed in the test suites.

*Note*: after this PR lands, prover only works with Boogie 2.7.35 or later.

## Motivation

Harden release against butterfly effect.

Yes

## Test Plan

NA

## Related PRs

#6373

### DIFF
--- a/language/move-prover/src/boogie_helpers.rs
+++ b/language/move-prover/src/boogie_helpers.rs
@@ -421,10 +421,10 @@ pub fn boogie_debug_track_local_via_function(
     var_idx: &str,
     value: &str,
 ) -> String {
-    format!(
-        "if (true) {{ assume $DebugTrackLocal({}, {}, {}, {}); }}",
+    ensure_trace_info(format!(
+        "assume $DebugTrackLocal({}, {}, {}, {});",
         file_idx, pos, var_idx, value
-    )
+    ))
 }
 
 /// Construct a statement to debug track a local based on the Boogie attribute approach.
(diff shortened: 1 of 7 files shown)


wrwg:
Additional comment: this PR also cleans up the code in boogie_wrapper around extracting values from the model, by breaking down a few gigantic functions into multiple ones. This appeared advisable to keep complexity lower, as now both old and new style debug tracing are supported (option based -- the new approach is the default).

language/move-prover/tests/sources/functional/arithm.exp
```diff
@@ -9,7 +9,7 @@ error: abort not covered by any of the `aborts_if` clauses
      │ ╰─────^
      ·
  126 │         x / y
-     │           - abort happened here with execution failure
+     │           - abort happened here
```
shazqadeer, line 12: Flagging this to make sure the change in error message is intended.

shazqadeer (approved):

language/move-prover/tests/sources/functional/arithm.exp
wrwg: Yep, unfortunately the code is more often not available than before, but it seems acceptable because in other cases it is.

language/move-prover/src/boogie_helpers.rs
```diff
@@ -447,17 +461,25 @@ pub fn boogie_debug_track_abort_via_function(
...
-    )
+    ))
+}
+
+/// Wraps a statement such that Boogie does not elimiate it in execution traces. Boogie
+/// does not seem to account for `assume` statements in execution traces, so in order to
+/// let our tracking statements above not be forgotten, this trick ensures that they always
+/// appear in the trace.
+fn ensure_trace_info(s: String) -> String {
+    format!("if (true) {{\n {}\n}}", s.replace("\n", "\n  "))
```
bmwill, line 484: You might not care too much about performance but if you did, and all calls to this function are of the form `ensure_trace_info(format!(...))` you could have this function take in `std::fmt::Arguments` and change callers to do `ensure_trace_info!(format_args!(..))` in order to avoid the extra allocation. Of course if you are required to replace `"\n"` with `"\n "` then you may not even be able to do what I've suggested

wrwg: Thanks for the hint, will remember this next time!

PR #6401, opened 2020-10-06
[RFC][libra-framework][breaking] Change reconfiguration rule

runtian-zhou:
## Motivation

This is yet another version of #6357 but a simplified version: only the `LibraConfig` module was changed and the writeset logic remains unchanged. The original reason for the later one was that we want to be able to do multi reconfiguration in WriteSet transaction. However, this is not essential due to the following reasoning:

There are two different types WriteSet:
   a. AdminScript, which refactoring the WriteSet epilogue logic will fix this issue intrinsically
   b. DirectWriteSet, the extra logic here is only buying us the convenience of constructing it offline. At this moment I think it's really not worth harming the on-chain component of the system to benefit the off chain generation process.

Thus we might only need the changes in move code.

Yes

## Test Plan

Editted the related functional tests.

## Related PRs

#6357


language/stdlib/modules/LibraConfig.move
```diff
@@ -224,6 +224,26 @@ module LibraConfig {
 
         let config_ref = borrow_global_mut<Configuration>(CoreAddresses::LIBRA_ROOT_ADDRESS());
         let current_time = LibraTimestamp::now_microseconds();
+
+        // Do not do anything if a reconfiguration event is already emitted within this transaction.
+        //
+        // This is OK because:
+        // - The time changes in every non-empty block
+        // - A block automatically ends after a transaction that emits a reconfiguration event, which is guarenteed by
```
sblackshear, line 232:
```suggestion
        // - A block automatically ends after a transaction that emits a reconfiguration event, which is guaranteed by
```

language/stdlib/modules/LibraConfig.move
```diff
@@ -224,6 +224,26 @@ module LibraConfig {
...
+        //   LibraVM spec that all transactions comming after a reconfiguration transaction will be returned as Retry
+        //   status.
+        // - Each transaction must emit at most one reconfiguration event
+        //
+        // Thus, this check ensures that a transaction that does multiple "reconfiguration required" actions emits only
+        // one reconfiguration event.
+        //
+        // Another footnote is that the reasoning above hinders reconfiguration in a NIL block after a reconfiguration
```
sblackshear, line 240: I don't understand this last part. What configuration changes are proposed in a block prologue?

sblackshear (approved):
This looks less invasive than the other PR to me... Stamping modulo comment fixes

language/stdlib/modules/LibraConfig.move
runtian-zhou: No configuration change is currently in the block prologue. Writing this down just to remind us that we couldn't do that in the future

sblackshear: I think I would leave this out--for me it raises more questions than it answers.

PR #6402, opened 2020-10-06
[consensus] slow down if no user txns

zekun000:
## Motivation

(Write your answer here.)

## Test Plan

## Related PRs

### DIFF
--- a/consensus/src/txn_manager.rs
+++ b/consensus/src/txn_manager.rs
@@ -8,6 +8,7 @@ use executor_types::StateComputeResult;
 use fail::fail_point;
 use futures::channel::{mpsc, oneshot};
 use itertools::Itertools;
+use libra_logger::prelude::*;
 use libra_mempool::{
     CommittedTransaction, ConsensusRequest, ConsensusResponse, TransactionExclusion,
 };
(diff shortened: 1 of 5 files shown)


consensus/src/txn_manager.rs
```diff
@@ -50,28 +52,37 @@ impl TxnManager for MempoolProxy {
...
-            .map_err(anyhow::Error::from)?;
-        // wait for response
-        match monitor!(
-            "pull_txn",
-            timeout(Duration::from_secs(1), callback_rcv).await
-        ) {
-            Err(_) => {
-                Err(format_err!("[consensus] did not receive GetBlockResponse on time").into())
+        let no_pending_txns = exclude_txns.is_empty();
+        for _ in 0..50 {
```
davidiw, line 56: a small doc block here would be priceless!

consensus/src/txn_manager.rs
```diff
@@ -50,28 +52,37 @@ impl TxnManager for MempoolProxy {
...
+                ))),
+                Ok(resp) => match resp.map_err(anyhow::Error::from)?? {
+                    ConsensusResponse::GetBlockResponse(txns) => Ok(txns),
+                    _ => Err(MempoolError::from(anyhow::anyhow!(
+                        "[consensus] did not receive expected GetBlockResponse"
+                    ))),
+                },
+            }?;
+            if txns.is_empty() && no_pending_txns {
+                delay_for(Duration::from_millis(NO_TXN_DELAY)).await;
```
davidiw, line 80: 500 ms delay, how will this translate in practice on testnet or a local run for example?
can we make this configurable? and if we do, could we make the configuration call into two different functions?
that way we can disable it if it becomes problematic on mainnet
also could a 500 ms delay ever translate into a timeout? should instead should we aim for a 1/10 or 1/4 time of the expected timeout (assuming that is relatively universal)
final thought... should we do 5 * 100 how much cpu are we burning doing this 50 times versus 5?

davidiw (commented):
cool idea

davidiw:
Someone said this is one line of code :P

zekun000:
it's just for experiment (unless you really want it)

davidiw:
> it's just for experiment (unless you really want it)

I think this is super useful for running a local testnet, our global testnet, and devnets as well... curious about @aching's thoughts.

davidiw:
also curious about @metajack's thoughts as he's been working with partners.

zekun000:
this seems acceptable now "fixed tps 10 : 10 TPS, 762 ms latency, 1000 ms p99 latency, no expired txns" with poll_count = 10, I'll default the value to 1 and delegate to each deployment to decide what value to use

consensus/src/txn_manager.rs
```diff
@@ -50,28 +62,40 @@ impl TxnManager for MempoolProxy {
...
+                    "[consensus] did not receive GetBlockResponse on time"
+                ))),
+                Ok(resp) => match resp.map_err(anyhow::Error::from)?? {
+                    ConsensusResponse::GetBlockResponse(txns) => Ok(txns),
+                    _ => Err(MempoolError::from(anyhow::anyhow!(
+                        "[consensus] did not receive expected GetBlockResponse"
+                    ))),
+                },
+            }?;
+            if txns.is_empty() && no_pending_txns {
```
davidiw, line 90: problem is you're gonna hit this on i == self.poll_count - 1, and you don't want to

davidiw (approved):
looks good but if we set this to 1, we ideally don't have an arbitrary sleep at the end ;)

consensus/src/txn_manager.rs
aching: Why didn't we make this configurable btw?

metajack:
This looks excellent for local testing where you don't need high TPS and you'd rather your machine not sound like an airplane.

PR #6403, opened 2020-10-06
[breaking] Add 5-min rate limit to validators calling the reconfiguration

valerini:
This PR adds 5-min rate-limits to the reconfigurations called by the Validator from the Validator Set.
Affects the LibraSystem's module and move functional tests.

[Issue 5117](https://l.workplace.com/l.php?u=https%3A%2F%2Fgithub.com%2Flibra%2Flibra%2Fissues%2F5117&h=AT0yM1Ss0P2gXMkVt8O6V4OvZukdtsr1EzDE66O6RGfCAUCqGBPe87A317WMWfZNM0Y4QyaECsKUoHBn6rh9BmntenQwOb8fC2UcprvmyUZdH0zaUcq2xRo00kgwrp4NRWS4QFLwVtVhkl-Tq_qTGsSETLauua48lqd7Gg)

## Motivation

This change prevents a malicious validator from inducing reconfiguration events too frequently and thus slowing down of halting the system. A malicious validator may still do so, but with the rate limit of 5 minutes and the Association will have the ability to remove this validator from the set if it reconfigures too often.

Yes.

## Test Plan

cargo xtest


execution/executor/tests/storage_integration_test.rs
```diff
@@ -109,7 +110,14 @@ fn test_reconfiguration() {
...
     // txn2 = a dummy block prologue to bump the timer.
-    let txn2 = encode_block_prologue_script(gen_block_metadata(1, validator_account));
+    let txn2 = encode_block_prologue_script(BlockMetadata::new(
+        gen_block_id(1),
+        1,
+        300001,
+        vec![],
+        validator_account,
+    ));
+    // let txn2 = encode_block_prologue_script(gen_block_metadata(1, validator_account));
```
davidiw, line 120: delete?

davidiw (commented):
cc: @JoshLind to help address breakage in smoke tests
@zekun000, @sblackshear, @aching for awareness.

I'm happy with this PR with the fixes to smoke tests. I suspect the smoke tests can be simplified while retaining their value.

language/stdlib/modules/LibraSystem.move
```diff
@@ -451,13 +466,19 @@ module LibraSystem {
         if (!ValidatorConfig::is_valid(validator_info.addr)) {
             return false
         };
+        // This should is ensured by the code calling this function, but left here for safety.
```
sblackshear, line 469: I worry that redundant checks like this can actually reduce safety by causing confusion over whose responsibility it is to enforce the invariant... For example, someone might delete the other check, continue to see tests passing, and think that it is ok (when in fact, the other check is the "real" one).

Thus, my preference would be to omit the check here and keep only the one at 276.

language/stdlib/modules/LibraSystem.move
```diff
@@ -262,6 +272,11 @@ module LibraSystem {
         let to_update_index_vec = get_validator_index_(&libra_system_config.validators, validator_addr);
         assert(Option::is_some(&to_update_index_vec), Errors::invalid_argument(ENOT_AN_ACTIVE_VALIDATOR));
         let to_update_index = *Option::borrow(&to_update_index_vec);
+
+        assert(LibraTimestamp::now_microseconds() >
```
sblackshear, line 276: Should this go inside the branch at 281?

I think the only difference in behavior is that a "no-op" config update within 5 minutes of the last config update will 
1. abort as the code is currently written
2. succeed, but do nothing if we put this inside the branch

language/stdlib/modules/LibraSystem.move
sblackshear: A rule of thumb I use for check redundancy: if you comment out any check, a test should fail. Otherwise, the check is redundant (or you don't have enough tests :) ).

valerini: Totally makes sense!

language/stdlib/modules/LibraSystem.move
valerini: Indeed, would makes sense to succeed if nothing has changed even if the rate limit is not met.

sblackshear (approved):
LGTM! Would you mind adding a test for the "too soon to reconfig, but ValidatorConfig didn't change" corner case?

zekun000:
is key manager aware of this? afaik, every validator tries to rotate the keys nearly at the same time @JoshLind

davidiw:
> is key manager aware of this? afaik, every validator tries to rotate the keys nearly at the same time @JoshLind

each validator has their own 5 min timer.

language/stdlib/modules/LibraSystem.move
```diff
@@ -27,6 +27,9 @@ module LibraSystem {
         /// Validator Operator, human name, and info such as consensus key
         /// and network addresses.
         config: ValidatorConfig::Config,
+        /// The time of last recofiguration invoked by this validator
```
mimoo, line 30: typo: "recofiguration"

language/stdlib/modules/LibraSystem.move
```diff
@@ -27,6 +27,9 @@ module LibraSystem {
...
+        /// in miscroseconds
```
mimoo, line 31: "miscroseconds"

language/stdlib/modules/LibraSystem.move
```diff
@@ -70,6 +73,11 @@ module LibraSystem {
     const EINVALID_TRANSACTION_SENDER: u64 = 4;
     /// An out of bounds index for the validator set was encountered
     const EVALIDATOR_INDEX: u64 = 5;
+    /// Rate limited when trying to update config
+    const ECONFIG_UPDATE_RATE_LIMITED: u64 = 6;
+
+    /// Number of microseconds in 5 minutes
+    const FIVE_MINUTES: u64 = 300000;
```
mimoo, line 80: that's not right no? don't you mean "5000000"?

mimoo: nevermind :D it's late here

aching, line 80: If this really is five minutes in microseconds, shouldn't this be 300000000?

aching (changes requested):
If you want 5 minutes, please fix the microseconds conversion.

language/stdlib/modules/LibraSystem.move
valerini: True, thank you so much for catching this!

language/stdlib/modules/LibraSystem.move
```diff
@@ -277,7 +293,7 @@ module LibraSystem {
         // The validator set is not updated if: (1) `validator_addr` is not in the validator set, or
         // (2) the validator info would not change. The code only does a reconfiguration if the
         // validator set would change.  `ReconfigureAbortsIf` complains if the block time does not
-        // advance, but block time only advances if there is a reconfiguration.
+        // advance, but block time only advances if there is a reconfiguration.x
```
aching, line 296:
```suggestion
        // advance, but block time only advances if there is a reconfiguration.
```

types/src/validator_info.rs
```diff
@@ -29,6 +29,9 @@ pub struct ValidatorInfo {
     consensus_voting_power: u64,
     // Validator config
     config: ValidatorConfig,
+    // The time of last recofiguration invoked by this validator
+    // in miscroseconds
```
aching, line 33:
```suggestion
    // in microseconds
```

aching (approved):

PR #6406, opened 2020-10-06
[json-rpc] clean up TransactionDataView mapper and add tests

xli:
## Motivation

* Remove unnecessary error check in TransactionDataView#from(Transaction), we already handled all types
* Add more test cases to cover corner cases.

Yes

## Test Plan

Unit test and integration test

## Related PRs

### DIFF
--- a/json-rpc/types/src/views.rs
+++ b/json-rpc/types/src/views.rs
@@ -520,13 +520,11 @@ impl ScriptView {
 
 impl From<Transaction> for TransactionDataView {
     fn from(tx: Transaction) -> Self {
-        let x = match tx {
-            Transaction::BlockMetadata(t) => {
-                t.into_inner().map(|x| TransactionDataView::BlockMetadata {
-                    timestamp_usecs: x.1,
-                })
-            }
-            Transaction::GenesisTransaction(_) => Ok(TransactionDataView::WriteSet {}),
+        match tx {
+            Transaction::BlockMetadata(t) => TransactionDataView::BlockMetadata {
+                timestamp_usecs: t.timestamp_usec(),
+            },
+            Transaction::GenesisTransaction(_) => TransactionDataView::WriteSet {},
             Transaction::UserTransaction(t) => {
                 let script_hash = match t.payload() {
                     TransactionPayload::Script(s) => HashValue::sha3_256_of(s.code()),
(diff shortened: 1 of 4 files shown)


libra-action:

Breaking changes policy:

2. Adding fields is ok.

json-rpc/types/src/views.rs
```diff
@@ -520,13 +520,11 @@ impl ScriptView {
...
-                t.into_inner().map(|x| TransactionDataView::BlockMetadata {
-                    timestamp_usecs: x.1,
-                })
-            }
-            Transaction::GenesisTransaction(_) => Ok(TransactionDataView::WriteSet {}),
+        match tx {
+            Transaction::BlockMetadata(t) => TransactionDataView::BlockMetadata {
+                timestamp_usecs: t.timestamp_usec(),
+            },
+            Transaction::GenesisTransaction(_) => TransactionDataView::WriteSet {},
```
thefallentree, line 527: should we rename this too?

json-rpc/types/src/views.rs
```diff
@@ -554,11 +552,9 @@ impl From<Transaction> for TransactionDataView {
                     script_hash,
                     script_bytes,
                     script: t.into_raw_transaction().into_payload().into(),
-                })
+                }
             }
-        };
-
-        x.unwrap_or(TransactionDataView::UnknownTransaction {})
```
thefallentree, line 561: why were these removed ?  It is good for backward compatibility

json-rpc/types/src/views.rs
xli: GenesisTransaction(WriteSetPayload),

I think keep it as WriteSet is better, GenesisTransaction seems too specific, although I don't mind to rename, but will need add new type to TransactionDataView, maybe another diff if we prefer.

json-rpc/types/src/views.rs
xli: because we don't need it, there is no "unknown" transaction, you can see all transaction type is mapped.
it is possible in the future, we will have unknown one, but that will be breaking change to existing code.
in above code, you can see I remove "Ok", so there is no "unwrap_or" anymore.

Technically, I'd like to return error if unwrap failed, instead of return unknown transaction type, which hides real error.
But here we really don't have unknown transaction anymore.

thefallentree (approved):

PR #6410, opened 2020-10-06
Removing unwrap and unreachable macros inside fns of type result

xvschneider:
## Motivation

Using _unwrap_ in a function that returns result is not recommended in the libra codebase, as it promotes recoverable errors to unrecoverable errors and it leads to panic. (More info [here](https://rust-lang.github.io/rust-clippy/master/#unwrap_in_result)). 

Furthermore, other macros that can lead to panics and should not be used in this repo (when inside a function of type result) are _unreachable!, unimplemented!, panic!_ and _todo!_ ( More info [here](https://rust-lang.github.io/rust-clippy/master/#panic_in_result_fn) ). 

Hence I am removing the usage of _unwrap_ and _unreachable!_ in such functions 

In the future, this lint will be added to clippy to prevent this from happening in the first place.

Yes

## Test Plan
No new functionality added, so no new tests needed, but I ran existing tests to ensure nothing is breaking.
cargo build
cargo xtest


crypto/crypto/src/multi_ed25519.rs
```diff
@@ -430,7 +430,10 @@ impl TryFrom<&[u8]> for MultiEd25519Signature {
             return Err(CryptoMaterialError::WrongLengthError);
         }
 
-        let bitmap = bytes[length - BITMAP_NUM_OF_BYTES..].try_into().unwrap();
```
mimoo, line 433: that was a weird unwrap! Can it be re-written as a simple .into()? Or maybe just by copying into an array of the right size?

common/logger/src/struct_log.rs
```diff
@@ -94,13 +94,19 @@ impl Write for TcpWriter {
...
+            let result = stream.write(buf);
+            if result.is_err() {
+                self.stream = None;
+            }
+            result
+        } else {
             self.stream = None;
+            Err(io::Error::new(
+                io::ErrorKind::NotConnected,
+                "Can't write to strean, not connected",
```
mimoo, line 107: @gregnazario ?

gregnazario: So, this was using the `Option<Stream>` as a poor man's lock on whether to refresh the connection or not.  This is used in a single threaded fashion, and I'm not sure that the situation here that seems to be concerned about can happen.

Also `s/strean/stream/g`

network/memsocket/src/lib.rs
```diff
@@ -305,11 +315,11 @@ impl MemorySocket {
         let (socket_a, socket_b) = Self::new_pair();
         // Send the socket to the listener
         if let Err(e) = sender.unbounded_send(socket_a) {
-            if e.is_disconnected() {
-                return Err(ErrorKind::AddrNotAvailable.into());
+            if !e.is_disconnected() {
+                error!("MUST_FIX e should not be disconnected");
             }
+            return Err(ErrorKind::AddrNotAvailable.into());
```
gregnazario, line 321: I'm not sure this message adds value?  If I see this message in the logs, what is `e` without then looking deep into the code?

network/memsocket/src/lib.rs
```diff
@@ -292,7 +299,10 @@ impl MemorySocket {
     /// # Ok(())}
     /// ```
     pub fn connect(port: u16) -> Result<MemorySocket> {
-        let mut switchboard = (&*SWITCHBOARD).lock().unwrap();
+        let mut switchboard =  match (&*SWITCHBOARD).lock() {
+            Ok(switchboard) => switchboard,
+            Err(_e) => return Err(ErrorKind::Other.into()),
+        };
```
gregnazario, line 305: Is there a reason to throw the error away?

Also, what's wrong with panicing on a poisoned lock?  If the lock is poisoned, this whole code basically stops working without manual intervention.

network/memsocket/src/lib.rs
```diff
@@ -90,8 +91,11 @@ impl MemoryListener {
     /// ```
     ///
     /// [`local_addr`]: #method.local_addr
-    pub fn bind(port: u16) -> Result<Self> {
-        let mut switchboard = (&*SWITCHBOARD).lock().unwrap();
+    pub fn bind(port: u16) -> Result<Self,> {
+        let mut switchboard =  match (&*SWITCHBOARD).lock() {
+            Ok(switchboard) => switchboard,
+            Err(_e) => return Err(ErrorKind::AddrInUse.into()),
+        };
```
gregnazario, line 98: Have we seen this be a problem in fuzzing or something?

`MemSocket` isn't used in the real world because we use `TcpSocket`s

gregnazario (commented):
Is there something that's driving the change of all panics into logs?  Are we afraid that it's going to panic over and over and over again?  Some of these situations we probably want it to actually panic like the poisoned lock.

crypto/crypto/src/multi_ed25519.rs
```diff
@@ -502,8 +505,16 @@ impl Signature for MultiEd25519Signature {
...
         // Public keys should be validated to be safe against small subgroup attacks, etc.
         precondition!(has_tag!(public_key, ValidatedPublicKeyTag));
-        let last_bit = bitmap_last_set_bit(self.bitmap);
-        if last_bit == None || last_bit.unwrap() as usize > public_key.length() {
+        let last_bit = match bitmap_last_set_bit(self.bitmap) {
+            Some(last_bit) => last_bit,
+            None => {
+                return Err(anyhow!(
+                    "{}",
+                    CryptoMaterialError::BitVecError("Signature index is out of range".to_string())
```
gregnazario, line 513: This message needs to be updated?

network/memsocket/src/lib.rs
mimoo: We've had this discussion before, but we never came to an actual agreement on if we should panic when locks are poisoned.
My point of view is that: if we are still running at this point, and we didn't crash yet, it's because we decided to not crash the whole process when the relevant thread panicked.
If we made that decision, it's because we wanted the rest of the system to continue working. Perhaps because we might still be able to do useful work. For example, if consensus was running in a thread that crashed, state-sync or mempool could still help the network.
Wdyt?

cc @davidiw @bmwill

bmwill: Yes we should panic on poisoned locks, preferably with `unwrap`. The only time locks are ever poisoned is if a thread panics while still holding the lock. Such cases are not really recoverable (they can be but I guarantee we are not prepared to recover from such cases.

This is one such case where blindly applying the "Don't panic in a function that can return a result" is the wrong thing to do, we do indeed want to panic here.

bmwill (dismissed):
Please revert the changes which unwrap on lock poisoning per our [coding guidelines](https://github.com/libra/libra/blob/master/documentation/coding_guidelines.md#error-handling). Our codebase is not equipped to handle poisoned locks, and the only way a lock becomes poisoned is if a thread panics while holding the lock.

network/memsocket/src/lib.rs
bmwill: We need to be very careful when making changes like this. This code was constructed under the assumption that only an error of `is_disconnected` is possible. blindly changing this without verifying we can handle other error cases would be a mistake.

crypto/crypto/src/multi_ed25519.rs
xvschneider: ` let bitmap = bytes[length - BITMAP_NUM_OF_BYTES..].into();
`
results into `the trait `std::convert::From<&[u8]>` is not implemented for `[u8; 4]``

xvschneider: I think returning a DeserializationError is most appropriate as the caller already expects this one in some cases.

network/memsocket/src/lib.rs
xvschneider: I've reverted the changes for the .lock().unwrap()

common/logger/src/struct_log.rs
xvschneider: > Please revert the changes which unwrap on lock poisoning per our [coding guidelines](https://github.com/libra/libra/blob/master/documentation/coding_guidelines.md#error-handling). Our codebase is not equipped to handle poisoned locks, and the only way a lock becomes poisoned is if a thread panics while holding the lock.

done

xvschneider: even if the situation in question won't happen (panicking on this unwrap), we still want to enable the lint to block all usage of unwrap in functions of type result , so I think making the changes now before the lint becomes available is the right thing to do. We want to reach a baseline of 0 violations of the lint before updating the lint, so that we won't have to fix all the violations in 1 massive PR in the future.

crypto/crypto/src/multi_ed25519.rs
```diff
@@ -502,8 +505,16 @@ impl Signature for MultiEd25519Signature {
...
+        let last_bit = match bitmap_last_set_bit(self.bitmap) {
+            Some(last_bit) => last_bit,
+            None => {
+                return Err(anyhow!(
+                    "{}",
+                    CryptoMaterialError::BitVecError("Failed to find last set bit".to_string())
+                ))
+            }
+        };
+        if last_bit as usize > public_key.length() {
```
bmwill, line 517: The rest of the changes look good, but we need to have a crypto expert review this as it is changing the expected invariants. cc @huitseeker

crypto/crypto/src/multi_ed25519.rs
mimoo: looked a bit more at it, yeah I agree that it's the correct thing to do. BTW that function could be re-written shorter as:

```rust
fn try_from(bytes: &[u8]) -> std::result::Result<MultiEd25519Signature, CryptoMaterialError> {
      let bytes = bytes
        .chunks_exact(ED25519_SIGNATURE_LENGTH);

        // check correct number of signatures
        if bytes.len() < 0 || bytes.len() > MAX_NUM_OF_KEYS {
          return Err(CryptoMaterialError::WrongLengthError);
        }

        // get bitmap
        let bitmap: [u8; BITMAP_NUM_OF_BYTES] = bytes.remainder().try_into().map_err(Err(CryptoMaterialError::WrongLengthError))?;
        if bitmap_count_ones(bitmap) != num_of_sigs as u32 {
          return Err(CryptoMaterialError::DeserializationError);
        }

        // get signatures
        let signatures: Result<Vec<Ed25519Signature>, _> = bytes
          .map(Ed25519Signature::try_from)
          .collect()?;
        
        // return multi-sig
        Ok(MultiEd25519Signature { signatures, bitmap })
    }
```

crypto/crypto/src/multi_ed25519.rs
mimoo, line 517: I would add this in the match case:

```rust
None => return Err(...),
Some(last_bit) if last_bit as usize > public_key.length => return Err(...),
Some(last_bit) => last_bit,
```

crypto/crypto/src/multi_ed25519.rs
```diff
@@ -502,8 +505,16 @@ impl Signature for MultiEd25519Signature {
     ) -> Result<()> {
         // Public keys should be validated to be safe against small subgroup attacks, etc.
         precondition!(has_tag!(public_key, ValidatedPublicKeyTag));
-        let last_bit = bitmap_last_set_bit(self.bitmap);
-        if last_bit == None || last_bit.unwrap() as usize > public_key.length() {
+        let last_bit = match bitmap_last_set_bit(self.bitmap) {
+            Some(last_bit) => last_bit,
+            None => {
+                return Err(anyhow!(
```
mimoo, line 511: cc @kchalkias I think I remember reading that anyhow is more for applications, and thiserror is more for libraries. Shouldn't we use thiserror here instead then?

mimoo: (I guess @phlip9 would know also)

crypto/crypto/src/multi_ed25519.rs
mimoo: it looks fine to me : )

xvschneider: done

crypto/crypto/src/multi_ed25519.rs
phlip9: if something is a library and there's any reason to ever match on the `Err`, then you should have a specific Error type for your library and return e.g. `Result<_, CryptoMaterialError>`. although, the calculus might change for a crypto library since you might want to reduce potential side-channel leakage. I know RustCrypto does something like this (iirc they just return an opaque error type, effectively `type Error = ();`).

mimoo: in this case we're only using wrappers around actual cryptographic primitives, so there shouldn't be any side channel possible (at least not for this file)

mimoo:
crypto looks good to me, @gregnazario if you can check the other stuff I'm good

common/logger/src/struct_log.rs
```diff
@@ -94,13 +94,19 @@ impl Write for TcpWriter {
...
-        let result = stream.write(buf);
-        if result.is_err() {
+        if let Some(mut stream) = self.stream.as_ref() {
+            let result = stream.write(buf);
+            if result.is_err() {
+                self.stream = None;
+            }
+            result
+        } else {
             self.stream = None;
```
gregnazario, line 104: This line isn't needed, because the stream at this point is `None`.

gregnazario (commented):
There's one unnecessary line, where the stream is set to None, after it already is `None`.  But, otherwise, I see no issues in the structured log part.

xvschneider:
> There's one unnecessary line, where the stream is set to None, after it already is `None`. But, otherwise, I see no issues in the structured log part.

thanks for reviewing!

mimoo (approved):
let's get this in then :) thanks for the PR and the patience @xvschneider !

common/logger/src/struct_log.rs
```diff
@@ -94,13 +94,18 @@ impl Write for TcpWriter {
...
+            if result.is_err() {
+                self.stream = None;
+            }
+            result
+        } else {
+            Err(io::Error::new(
+                io::ErrorKind::NotConnected,
+                "Can't write to stream, not connected",
+            ))
         }
```
davidiw, line 108: I think this would be more cleanly written as a set of Rust functors:

```
self.stream
  .as_mut()
  .ok_or_else(|| io::Error::new(...))
  .and_then(|stream| stream.write(...).map_err(|e| { self.stream = None; e })
```

xvschneider:
oops, things went wrong when rebasing, trying to fix

common/logger/src/struct_log.rs
mimoo: that's the annoying things with rust, you can write everything in imperative or functional style, so which do you choose :P?

gregnazario:
Was this merged in and how?  I saw an email something about 59 commits?

xvschneider:
> Was this merged in and how? I saw an email something about 59 commits?

Hi. I'll have to re-create this PR . There were merge conflicts so I had to rebase and something went wrong there that caused all these new commits to appear. Hence why I abandoned the PR

gregnazario:
> > Was this merged in and how? I saw an email something about 59 commits?
> 
> Hi. I'll have to re-create this PR . There were merge conflicts so I had to rebase and something went wrong there that caused all these new commits to appear. Hence why I abandoned the PR

Okay, let us know if you need any help with the process :). I've already had to wrangle through a couple weird rebase situations myself as well.

bmwill:
Even if a rebase or a merge went wrong, everything can be resolved/fixed without needing to abandon and then reopen a new PR.

PR #6412, opened 2020-10-06
[libra framework] cleanup RecoveryAddress and Option

emmazzz:
## Motivation

This PR cleans up the specs and documentation for `RecoveryAddress` and `Option` modules.

Yes.


language/stdlib/modules/Option.move
```diff
@@ -1,16 +1,10 @@
 address 0x1 {
```
DavidLDill, line 1: I noticed in the formatted doc that EOPTION_IS_SET and EOPTION_NOT_SET have exactly the same descriptions.  The descriptions may be correct ("The Option is in an invalid state for the operation attempted."), but it's not as helpful as it could be.

DavidLDill (approved):

language/stdlib/modules/Option.move
emmazzz: Will make it better.

PR #6417, opened 2020-10-07
[language][vm] add comments to better specify the error semantics of the public interfaces

vgao1996:
This is an attempt to specify the error/invariant semantics for the Move VM in a clearer way, which helps us avoid mistakes when implementing and using the Move VM and reason about tests and coverage.

### DIFF
--- a/language/move-vm/runtime/src/move_vm.rs
+++ b/language/move-vm/runtime/src/move_vm.rs
@@ -14,6 +14,20 @@ impl MoveVM {
         }
     }
 
+    /// Create a new Session backed by the given storage.
+    ///
+    /// Right now it is the caller's responsibility to ensure cache coherence of the Move VM Loader
+    ///   - When a module gets published in a Move VM Session, and then gets used by another
+    ///     transaction, it will be loaded into the code cache and stay there even if the resulted
+    ///     effects do not get commited back to the storage when the Session ends.
+    ///   - As a result, if one wants to have multiple sessions at a time, one needs to make sure
+    ///     none of them will try to publish a module. In other words, if there is a module publishing
+    ///     Session it must be the only Session existing.
+    ///   - In general, a new Move VM needs to be created whenever the storage gets modified by an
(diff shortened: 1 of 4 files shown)


language/move-vm/runtime/src/session.rs
```diff
@@ -61,6 +88,18 @@ impl<'r, 'l, R: RemoteCache> Session<'r, 'l, R> {
...
+    ///
+    /// The Move VM MUST return a user error (in other words, an error that's not an invariant violation) if
+    ///   - The module fails to deserialize or verify.
+    ///   - A module with the same ModuleId already exists in the environment.
+    ///   - The sender address does not match that of the module.
+    ///
+    /// The Move VM should not be able to produce other user errors.
+    /// Besides, no user input should cause the Move VM to return an invariant violation.
+    ///
+    /// In case an invariant violation occurs, the whole Session should be considered curropted and one shall
```
dariorussi, line 101: "curropted" -> corrupted

language/move-vm/runtime/src/move_vm.rs
```diff
@@ -14,6 +14,17 @@ impl MoveVM {
         }
     }
 
+    /// Create a new Session backed by the given storage.
+    ///
+    /// Right now it is the caller's responsibility to ensure cache coherence of the Move VM Loader
+    ///   - In general, a new Move VM needs to be created whenever changes to the published code
```
dariorussi, line 20: I don't see how publishing a Module requires a VM to be recreated. In fact, that should not be the case. Maybe I am reading this wrong, but I should be able to publish and keep going, as a matter of fact that is what we do today, so hopefully that is correct behavior

dariorussi (approved):
LGTM

language/move-vm/runtime/src/move_vm.rs
vgao1996: Oops, looks like I generalized the rules a little bit too much (on the conservative side). Got it updated.

PR #6419, opened 2020-10-07
Improve safety rules observability

davidiw:
* Log the config at startup -- it is a boring config, so only print it once
* Log internal state each time consensus_state is queried
* Log internal state at startup
* Log backtrace on panic

### DIFF
--- a/consensus/safety-rules/src/remote_service.rs
+++ b/consensus/safety-rules/src/remote_service.rs
@@ -4,7 +4,7 @@
 use crate::{
     persistent_safety_storage::PersistentSafetyStorage,
     serializer::{SafetyRulesInput, SerializerClient, SerializerService, TSerializerClient},
-    Error, SafetyRules,
+    Error, SafetyRules, TSafetyRules,
 };
 use libra_logger::warn;
 use libra_secure_net::{NetworkClient, NetworkServer};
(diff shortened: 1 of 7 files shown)


consensus/safety-rules/src/safety_rules.rs
```diff
@@ -219,6 +219,16 @@ impl SafetyRules {
     // Internal functions mapped to the public interface to enable exhaustive logging and metrics
 
     fn guarded_consensus_state(&mut self) -> Result<ConsensusState, Error> {
+        let waypoint = self.persistent_storage.waypoint()?;
+        let safety_data = self.persistent_storage.safety_data()?;
+
+        info!(SafetyLogSchema::new(LogEntry::State, LogEvent::Update)
```
zekun000, line 225: this is only called together with initialize, maybe log it there?

zekun000 (approved):

consensus/safety-rules/src/safety_rules.rs
davidiw: I also call it in when the remote service starts up. Considering this can be called arbitrarily whereas initialize cannot, it seemed safer to put here.

consensus/safety-rules/src/remote_service.rs
```diff
@@ -33,13 +33,17 @@ pub fn execute(
     verify_vote_proposal_signature: bool,
     network_timeout_ms: u64,
 ) {
-    let safety_rules = SafetyRules::new(storage, verify_vote_proposal_signature);
+    let mut safety_rules = SafetyRules::new(storage, verify_vote_proposal_signature);
+    if let Err(e) = safety_rules.consensus_state() {
```
davidiw, line 37: here @zekun000

davidiw:
@metajack, it seems this doesn't ping @mimoo for a review if I am the author?

PR #6420, opened 2020-10-07
[libra framework][move-prover] Abstraction of auth key construction in LibraAccount

wrwg:
This pulls auth key construction from make_account into its own private function, which can then be abstracted for verification. This avoids the use of vector concatenation of `LCS:to_bytes(x)` results in verification conditions, which appear to be a source of butterflies.

This doesn't enable new verifications (everything in LibraAccount verifies today), but hardens against timeouts with future critical changes to code and specs.

## Motivation

Hardening for release.

Yes

## Test Plan

NA

## Related PRs

NA


language/stdlib/modules/LibraAccount.move
```diff
@@ -971,12 +965,44 @@ module LibraAccount {
...
+        /// ever become relevant for verification of callers.
+        pragma opaque;
+        include [abstract] CreateAuthenticationKeyAbortsIf;
+        ensures [abstract]
+            result == spec_abstract_create_authentication_key(auth_key_prefix) &&
+            len(result) == 32;
+    }
+    spec schema CreateAuthenticationKeyAbortsIf {
+        auth_key_prefix: vector<u8>;
+        aborts_if 8 + len(auth_key_prefix) != 32 with Errors::INVALID_ARGUMENT;
```
shazqadeer, line 1002: @tzakian : Can you comment on this aborts_if specification?  This is written assuming that the byte representation of the address of an account is 8 bytes.  But, in our conversation just now you indicated that an account is 16 bytes.

wrwg: To add to this, there is no LCS spec where this would be written down AFAICT. I followed the link in LCS.move but it just points to the code.

wrwg: > his is written assuming that the byte representation of the address of an account is 8 bytes. 

More precise, it is assuming that the actual length does not matter for verification for callers, as is written now in the doc comment (and as has been assumed already implicitly before this PR).

tzakian: I'm not fully sure about the semantics of `abstract` so can't fully comment on this. 

To answer Shaz's question, the address is `16` bytes. I am fairly certain the LCS of an account address will be 16 bytes as well since the underlying address is just a `Vec<u8>`, so the 8 looks odd here, but I leave it up to you to decide if this change is needed though since it might change the abort behavior.

wrwg: Ok, I updated this to 16 (before this PR, I think the way how we defined LCS it was actually 4. ;-)).

Notice that there is still the issue I mentioned above: there is no easily discoverable spec of the "canonical" LCS format. At least, I did not find anything then the code when following the link you find in `LCS.move`. This might be OK for this first release, but we need to have this at some point. @matbd @sblackshear

shazqadeer (approved):

PR #6424, opened 2020-10-07
[mempool] add periodic mempool snapshot

christinacelee:
## Motivation

Adding cron job that logs snapshot of internal mempool state, to improve observabiity of txns in mempool.


bmwill (approved):
This is great! Another option would have been to use the `sample!` macro and instrument the main mempool event loop but I like this approach much better (considering `CoreMempool` type is already wrapped in an `Arc<Mutex<>>`)

PR #6426, opened 2020-10-07
[state sync] raise standards for recovering from multicast/failover mode

christinacelee:
## Motivation

In vfn failover cluster test experiment, the VFN makes progress fine when its upstream validator goes down by syncing with its failover VFNs. However, when the upstream validator comes back up, in the current code, the VFN immediately stops casting to its failovers and reverts back to casting to its validator only. However, because when the validator is coming back up, it is still catching up/behind the VFN's state, so the VFN's long poll requests to V have to timeout before the VFN fails back over to the failover VFNs. While this is not technically a bug because the VFN does eventually continue to make progress, the timeout before the failover distracts the VFN's sync progress, which isn't ideal. 

This PR changes the meaning of `multicast_level` to indicate the largest preference of the network (defined by `UpstreamConfig`) to multicast to, not the number of first live networks available to multicast to. When we select peers to broadcast to in all upstream networks with preference <= `multicast_level`, we will failover to the next available network with live peers to broadcast to and increase `multicast_level` accordingly. 

## Test Plan

Above issue is resolved in vfn failover cluster test experiment - the validator coming back up and catching up does not interrupt/distract the VFN's state sync efforts, and VFN makes uninterrupted progress by continuously sending requests to failover VFNs even when V is up but still catching up.


state-synchronizer/src/request_manager.rs
```diff
@@ -234,12 +234,47 @@ impl RequestManager {
...
+            return peers;
+        }
+
+        // if no peers were found from current multicast level above,
+        // do best effort search for next network with live peers to fail over to and pick a peer from it
+        // and update multicast level accordingly
+        let failover = self
+            .eligible_peers
+            .range((Excluded(self.multicast_level), Bound::Unbounded))
+            .next()
```
zekun000, line 258: this doesn't seem doing what the comment suggests? is the `.next()` guaranteed to have live peers or you actually want to use `.find_map()`?

christinacelee: `eligible_peers` is always going to contain live peers - see `update_peer_selection_data`

zekun000 (commented):
I'm confused by the change and the summary in the PR. The code doesn't achieve `This PR makes getting out of failover mode more difficult` since the only way to reduce multicast level is still via processing a successful response. but instead it makes getting into the failover mode more easier by auto failover to the next live peers instead of waiting for timeout.

state-synchronizer/src/request_manager.rs
zekun000: I think this fn can be simpler
```
  while peers.is_empty() || level < self.multicast_level {
    let (level, p) = eligible_peers_iter.next().map(..);
    peers.extend(&mut p);
  }
  if level > self.multicast_level {
    ...
  }
  peers
```

christinacelee: pushed new commit with a variation of the above (reorganize the nesting levels to avoid iter/borrow craziness)

state-synchronizer/src/request_manager.rs
```diff
@@ -234,12 +233,33 @@ impl RequestManager {
...
+            if !chosen_peers.is_empty() && *level >= self.multicast_level {
+                if *level > self.multicast_level {
+                    info!(
+                        LogSchema::event_log(LogEntry::Multicast, LogEvent::Failover)
+                            .old_multicast_level(self.multicast_level)
+                            .new_multicast_level(*level),
+                    );
+                    self.multicast_level = *level;
+                    counters::MULTICAST_LEVEL.set(self.multicast_level as i64);
+                }
```
davidiw, line 257: I think this logic is a little nuanced and not well documented...

Effectively what you want is one function to make your preference less strict (more open) and that is dependent on lack of availability of a node.

Then you want another function to increase that preference level based upon availability of actual data.

This does have shortcomings, which is namely these two properties are not symmetric and thus hard to cleanly reason about. At any rate, a doc block on this would be great and then you can reference that doc block in the code.

"This relaxes our multicast requirements so that we can incorporate additional peers for making state sync requests."

Also larger level is less preferred

state-synchronizer/src/request_manager.rs
```diff
@@ -329,12 +349,12 @@ impl RequestManager {
...
             == Some(0);
-        if is_primary_upstream_peer && self.multicast_level != MIN_UPSTREAM_NETWORK_CT {
+        if is_primary_upstream_peer && self.multicast_level != PRIMARY_NETWORK_PREFERENCE {
             // if chunk from a primary upstream is successful, stop multicasting the request to failover networks
             info!(LogSchema::event_log(LogEntry::Multicast, LogEvent::Recover)
                 .old_multicast_level(self.multicast_level)
-                .new_multicast_level(MIN_UPSTREAM_NETWORK_CT));
-            self.multicast_level = MIN_UPSTREAM_NETWORK_CT;
+                .new_multicast_level(PRIMARY_NETWORK_PREFERENCE));
+            self.multicast_level = PRIMARY_NETWORK_PREFERENCE;
```
davidiw, line 357: wouldn't it be better to just say

```
if peer.level < self.multicast_level {
  // update to a lower level
}

zekun000: +1 to ^, also probably worth having a function to save the boilerplate of info + set + counter.

zekun000 (approved):

PR #6428, opened 2020-10-07
Make most network constants configurable and remove TODOs that won't be done

gregnazario:
### Motivation
As a part of doing some cleanup of TODOs and ensuring the code is in a good state, I've gone ahead and removed a bunch of TODOs that were not going to get done / didn't apply.

In addition, I've made it such that `NetworkConfig` is the *almost* one true source of all constants and configuration for Networking.  There were also some test NetworkBuilder improvements as a result.

The new location of the constants is less than ideal, but it keeps the constants related to their insertion point (defaults for the network config)

### DIFF
--- a/config/src/config/network_config.rs
+++ b/config/src/config/network_config.rs
@@ -23,13 +23,23 @@ use std::{
     string::ToString,
 };
 
+// TODO: We could possibly move these constants somewhere else, but since they are defaults for the
+//   configurations of the system, we'll leave it here for now.
 /// Current supported protocol negotiation handshake version.
 ///
 /// See [`perform_handshake`] in `network/src/transport.rs`
-// TODO(philiphayes): ideally these constants live somewhere in network/ ...
-// might need to extract into a separate network_constants crate or something.
 pub const HANDSHAKE_VERSION: u8 = 0;
-pub const MAX_FRAME_SIZE: usize = 8 * 1024 * 1024;
+pub const NETWORK_CHANNEL_SIZE: usize = 1024;
+pub const PING_INTERVAL_MS: u64 = 1000;
+pub const PING_TIMEOUT_MS: u64 = 10_000;
+pub const PING_FAILURES_TOLERATED: u64 = 5;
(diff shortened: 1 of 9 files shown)


config/src/config/network_config.rs
```diff
@@ -23,21 +23,42 @@ use std::{
...
 /// See [`perform_handshake`] in `network/src/transport.rs`
-// TODO(philiphayes): ideally these constants live somewhere in network/ ...
-// might need to extract into a separate network_constants crate or something.
 pub const HANDSHAKE_VERSION: u8 = 0;
-pub const MAX_FRAME_SIZE: usize = 8 * 1024 * 1024;
+pub const NETWORK_CHANNEL_SIZE: usize = 1024;
+pub const PING_INTERVAL_MS: u64 = 1000;
+pub const PING_TIMEOUT_MS: u64 = 10_000;
+pub const CONNECTIVITY_CHECK_INTERVAL_MS: u64 = 5000;
+pub const PING_FAILURES_TOLERATED: u64 = 10;
```
davidiw, line 36: this is really a silly / high number... can we make this 5?

config/src/config/network_config.rs
```diff
@@ -74,7 +104,12 @@ impl Default for NetworkConfig {
 impl NetworkConfig {
     pub fn network_with_id(network_id: NetworkId) -> NetworkConfig {
         let mut config = Self {
-            connectivity_check_interval_ms: 5000,
+            max_connection_delay_ms: MAX_CONNECTION_DELAY_MS,
```
davidiw, line 107: btw I feel the mixing of consts and na default config seem kinda duplicitous but ... that's my style

davidiw (approved):

config/src/config/network_config.rs
gregnazario: Yeah, I'm much more a readability thing to see the constants.  I know it's kinda a lot of repetitiveness.

config/src/config/network_config.rs
gregnazario: Ha, the funny thing is this is what we had before.  Let's do it and make it 5!

PR #6429, opened 2020-10-07
[toolchain] bump default toolchain for latest stable cargo in calls  `cargo xtest` and `cargo xbuild`

rexhoffman:
## Motivation

So when running cargo xbuild/cargo xtest, even though x calls a nightly cargo, the older cargo used to invoke x runs in to this problem:   https://github.com/rust-lang/cargo/pull/8473
So upgrading to avoid it.   Also, trimming the environment variable of its trailing newline.

Yes

## Test Plan

CI, local builds

## Related PRs

None


metajack (approved):

devtools/x/src/cargo.rs
```diff
@@ -62,7 +62,7 @@ impl Cargo {
         // The environment is inherited for child processes so we only need to set RUSTUP_TOOLCHAIN
         // if it isn't already present in the environment
         if env::var_os(RUSTUP_TOOLCHAIN).is_none() {
-            inner.env(RUSTUP_TOOLCHAIN, RUST_TOOLCHAIN_VERSION);
+            inner.env(RUSTUP_TOOLCHAIN, RUST_TOOLCHAIN_VERSION.trim());
```
bmwill, line 65: This is a constant and should not need to be "trimmed" just update the constant to not have excess whitespace if needed

rexhoffman: Can't call trim in a constant block, at least that's what the compiler tells me.

bmwill: I'm not suggesting you call trim when creating the const, I'm suggesting you make sure the const has a value that doesn't need to be trimmed, which it currently doesn't.

rexhoffman: As per offline communication, since this is sourced from a file, this is fine to land.

bmwill: Yep, I didn't take the time I should have to actually look at how the constant itself was initialized. I never intended to block this PR from landing though (I would have requested changes if I had). I also suppose my initial question wasn't worded well enough to get at what I really wanted to know: What issue you were actually seeing without trimming and why trimming seemed to solve the issue.

rexhoffman: while doing some tests with sccache, I observed some versions of cargo (rustup wrapped cargo) fail when reading the environment variable.   Can't recall the exact setup now - but end result was a broken compilation.   Not entirely sure of all the impact that extra line could have had, as, when I saw that error the first time -- I just fixed and carried on fighting another bug in cargo.

language/transaction-builder/generator/src/lib.rs
```diff
@@ -33,6 +33,7 @@ pub fn read_abis<P: AsRef<std::path::Path>>(dir_path: P) -> anyhow::Result<Vec<S
         abis.push(lcs::from_bytes(&buffer)?);
     }
     // Sorting scripts by alphabetical order.
+    #[allow(clippy::unnecessary_sort_by)]
```
rexhoffman, line 36: The borrow checker fails with sort_by_key as is suggested by this lint.   Probably indicates there are more issues with the borrowing occurring in this method.

language/move-prover/src/lib.rs
```diff
@@ -326,8 +326,7 @@ fn create_and_process_bytecode(options: &Options, env: &GlobalEnv) -> FunctionTa
         Some(
             options
                 .move_sources
-                .iter()
```
rexhoffman, line 329: as per lint

PR #6436, opened 2020-10-07
[network] Improve NetworkAddress spec

gregnazario:
Now, adds more restrictions on actual usage e.g. It must have a Layer3 and a Layer4 protocol, can't end in a `/` etc.

### DIFF
--- a/specifications/network/network-address.md
+++ b/specifications/network/network-address.md
@@ -1,11 +1,14 @@
 # NetworkAddress
 
-`NetworkAddress` is a compact, efficient, self-describing and future-proof network address represented as a stack of protocols, inspired by libp2p's [multiaddr](https://multiformats.io/multiaddr/) format. The primary differences include using [LCS] to describe the binary format and reducing the set of supported protocols.
+`NetworkAddress` is a compact, efficient, self-describing and future-proof network address represented as a stack of
+protocols, inspired by libp2p's [multiaddr](https://multiformats.io/multiaddr/) format. The primary differences include
+using [LCS] to describe the binary format and reducing the set of supported protocols.
 
 (TODO(philiphayes): we should rename this to make it less generic sounding...)
 (TODO(philiphayes): include `EncNetworkAddress` spec)
 
-In particular, a `NetworkAddress` is intended to be a fully self-contained description of _how_ to dial a [LibraNet](spec.md) peer, describing both the base transport protocol and all subsequent connection upgrade protocols.
+In particular, a `NetworkAddress` is intended to be a fully self-contained description of _how_ to dial a
+[LibraNet](spec.md) peer, describing both the base transport protocol and all subsequent connection upgrade protocols.
 
 ## Data Structures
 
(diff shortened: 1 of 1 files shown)


davidiw (approved):

specifications/network/network-address.md
```diff
@@ -92,6 +84,17 @@ NetworkAddress([
...
 
+A `NetworkAddress` as a concatenation of `Protocol` segments must have the following characteristics:
+1. The address must contain a Layer3 protocol (e.g. `Ip4` or `Dns`).
+2. The address must contain a Layer4 protocol (e.g. `Tcp` or `Memory`).
+3. A protocol may be used at most once in an address.
+4. The address must not end in a forward slash `/`
+
+Example possible combinations:
+* `/ip4/127.0.0.1/tcp/6080`
+* `/ip4/127.0.0.1/memory/6080`
```
phlip9, line 95: should just be `/memory/6080` since `Memory` is ~ L3+L4

specifications/network/network-address.md
```diff
@@ -49,25 +46,20 @@ pub enum Protocol {
...
-(TODO(philiphayes): hardening to limit NetworkAddress and RawNetworkAddress max size)
+(TODO(philiphayes): hardening to limit NetworkAddress max size)
 
 ## Human-readable Format
 
-All `NetworkAddress` sent and received over-the-wire or stored on-chain are in their compact `RawNetworkAddress` [`lcs`]-serialized format. However, implementations may also wish to implement the optional human-readable format to aid reading `NetworkAddress`es from configuration or printing `NetworkAddress` in logs.
+All `NetworkAddress` sent and received over-the-wire or stored on-chain are in their compact [`u8`] [`lcs`]-serialized format. However, implementations may also wish to implement the optional human-readable format to aid reading `NetworkAddress`es from configuration or printing `NetworkAddress` in logs.
 
-Each `Protocol` segment is formatted like:
+Each `Protocol` segment is formatted of the form `/<Protocol>/<Value>`, and applies to the following rules:
```
phlip9, line 58: personally, I view these as `/<Protocol>[/<Argument>]*`, where you can have zero-or-more "arguments" for each protocol. for example, `/memory/<port>` takes 1 argument. of course all of our protocols happen to only take one argument atm : p

specifications/network/network-address.md
```diff
@@ -92,6 +84,17 @@ NetworkAddress([
 ]) => "/ip4/127.0.0.1/tcp/6080/ln-noise-ik/080e287879c918794170e258bfaddd75acac5b3e350419044655e4983a487120/ln-handshake/0"
 ```
 
+A `NetworkAddress` as a concatenation of `Protocol` segments must have the following characteristics:
+1. The address must contain a Layer3 protocol (e.g. `Ip4` or `Dns`).
+2. The address must contain a Layer4 protocol (e.g. `Tcp` or `Memory`).
```
phlip9, line 89: `Memory` is both L3+L4

specifications/network/network-address.md
gregnazario: Yeah, that makes sense

specifications/network/network-address.md
gregnazario: Well, I didn't say anywhere that `<Value>` had to be a single item, it could be encoded however we like, but there's currently no specification on how you would separate multiple arguments because there's no example.

I have a feeling that's up to the protocol to specify?

gregnazario: I've gone ahead and changed the spec to say "ProtocolConfig" instead of "Value"

gregnazario:
There seems to be some problem with whitespace that was fixed in the final commit, but in one commit, it was there, so I had to remove the spaces and rebase

PR #6437, opened 2020-10-07
[move lang] Make dep structs not native

tnowacki:
- Small spelling fix
- Quick fix to not make structs in deps native. it might make some things tiny bit slower, but it prevents really bizarre error messages

## Motivation

- Current edge case of trying to instantiate a struct declared in a dep produces some weird error messages

## Test Plan

- cargo test

### DIFF
--- a/language/move-lang/functional-tests/tests/move/dependencies/pack_unpack_private.exp
+++ b/language/move-lang/functional-tests/tests/move/dependencies/pack_unpack_private.exp
@@ -1,0 +1,5 @@
+[0] Transaction(0)
+[1] Move VM Status: EXECUTED
+[2] Transaction Status: Keep(EXECUTED)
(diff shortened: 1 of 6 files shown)


sblackshear (approved):
Add a test case that would have produced the weird error message if it's easy (though guessing testing anything with deps files is not)?

tnowacki:
Can't yet. Will do once I have some cli tests (future PR)

davidiw:
looks abandoned

bob-wilson:
@tnowacki I don't know if you intended to drop this, but at least the typo in the error message should get fixed. I checked that it is still there.

tnowacki:
This was stale due to not being able to land changes to the bytecode instructions. Specifically caused by this PR due to issues described in #7289

PR #6440, opened 2020-10-07
[network] Errors & Error Messages

gregnazario:
* Update some error messages to be a little more helpful after looking through the logs and seeing messages that don't help the reader too much
* Removed Unused `NetworkErrorKind` to decrease the amount of random hanging dead code

### DIFF
--- a/network/src/peer_manager/mod.rs
+++ b/network/src/peer_manager/mod.rs
@@ -587,11 +587,13 @@ where
         self.sample_connected_peers();
         match request {
             PeerManagerRequest::SendMessage(peer_id, msg) => {
+                let protocol_id = msg.protocol_id;
                 if let Some((conn_metadata, sender)) = self.active_peers.get_mut(&peer_id) {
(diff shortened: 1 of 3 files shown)


network/src/peer_manager/mod.rs
```diff
@@ -587,40 +587,47 @@ where
...
                                 .connection_metadata(conn_metadata),
+                            protocol_id = %protocol_id,
+                            error = ?err,
                             "{} Failed to forward outbound rpc to downstream actor. Error: {:?}",
                             self.network_context,
                             err
                         );
                     }
                 } else {
-                    warn!(
```
davidiw, line 621: not sure I understand the downgrade, but :shrug:
same with the above

davidiw (approved):

network/src/peer_manager/mod.rs
gregnazario: I thought that it wasn't that important if it was already disconnected, because what are you going to do?  I'll just change it back to warn.  In a previous PR, I actually lowered it from `error` lol...

PR #6442, opened 2020-10-07
[libra-mutex] introduce wrapper around std mutex

mimoo:
This PR takes the same approach @AnomalRoil took with libra-time: hide the panic under an API that is infallible. The same way unsafe interfaces can be hidden behind safe interfaces.

Thus, we remove a number of undocumented unwrap in the code, and we also move forward a state where it's easier to enable lints (cc @xvschneider) that will prevent us from writing fallible functions (that return `Result`) that can panic directly.


bmwill:
I personally am not a fan of the changes that were made wrt to the `libra-time` crate but that's neither here nor there. I understand the end goal of this PR is to try and limit to location where we're "unwrapping" and potentially causing a panic across the codebase. I think that if we're going through this effort that instead of making our own "mutex" crate that we should instead directly use the locking primitives in the [`parking_lot`](https://docs.rs/parking_lot/0.11.0/parking_lot/) crate. This crate should already be somewhere in our dependency tree so it wouldn't be adding a new dep and it has some advantages over the stdlib mutex's already.

mimoo:
why are you not a fan of the libra-time approach? I thought about alternatives but I couldn't think of much. I think it's much better with it than without it.

I didn't know parking_lot, I'm not opposed to switching to it : o but I'm wondering what others think? I've also seen people doing a hybrid approach: https://docs.rs/shkeleton/0.3.2/src/shkeleton/shkeleton.rs.html#7 (so keeping our own wrapper, but we can easily swap the underlying implementation)

davidiw:
Two thoughts:
1) I'd love to see a common crate for infallible types rather than spread across multiple
2) It is worth exploring things like porking_lot, but maybe a little later ... I could see this PR maybe making it in soon but a parking lot refactor might need more scrutiny.

bmwill:
Fair enough @davidiw, I think its something we could evaluate later.

While we're doing this, though. Instead of having a set of one-off crates maybe we can have a single crate with all of these goodies in them?

mimoo:
> I'd love to see a common crate for infallible types rather than spread across multiple

as discussed privately, I think common could benefit from some cleaning:

* as indicated by @davidiw libra-time, libra-mutex can be under the same libra-infallible crate
* why is LCS here but not types? Considering the noise in common I'd be in for moving LCS outside as it's a pretty critical component of the system (something that any implementation in any language would have to re-implement)
* lots of test-only crates could potentially be regrouped?
* some crates like temppath seems to be some sort of "our own stdlib" crate, which we might want to publish as their own crates at some point?

davidiw (approved):
are you gonna do rwlock next?

mimoo:
for the love of humanity I just can't figure out why this doc building is failing...

mimoo:
nevermind, my fork was too late behind =.= had to rebase

PR #6443, opened 2020-10-07
experiment on std mutex -> parking_lot

mimoo:


mimoo:
indeed no improvement x) but that's without enabling it for tokio and else...

PR #6446, opened 2020-10-07
[cluster-test] removing infinite recursion in Debug implementation

mimoo:


andll (approved):
Thank you. (I hope we checked that Display does not call debug :D)

mimoo:
yeah Display does the right thing :D

PR #6449, opened 2020-10-08
[libra-framework][breaking] Rename upgrade event

runtian-zhou:
## Motivation

Renamed `UpgradeEvent` to `AdminTransactionEvent`. Change the field in the event to include a committed time instead of the writeset bytes, as writeset bytes could be obtained in the raw transaction.

Yes

## Test Plan

Updated test + spec doc


bob-wilson:
Can you explain more about why this is important to do right now?

runtian-zhou:
> Can you explain more about why this is important to do right now?

Yea there are two main reasons: 
1. The name `UpgradeEvent` itself is confusing, as the name implies upgrade in either Libra Core or Libra Framework, which is not necessary the case for WriteSet.
2. The event payload is redundant as querying the transaction that emit this event will give you such data. `committed_time` is a more useful info that is not easily derivable in the current system.

json-rpc/types/src/views.rs
```diff
@@ -316,11 +316,11 @@ impl TryFrom<ContractEvent> for EventDataView {
                 created_address,
                 role_id,
             }
-        } else if event.type_tag() == &TypeTag::Struct(UpgradeEvent::struct_tag()) {
+        } else if event.type_tag() == &TypeTag::Struct(AdminTransactionEvent::struct_tag()) {
             // TODO: missing test
```
xli, line 320: could you remove this todo, I forgot to remove it when I added test.

json-rpc/types/src/views.rs
```diff
@@ -316,11 +316,11 @@ impl TryFrom<ContractEvent> for EventDataView {
...
-            let upgrade_event = UpgradeEvent::try_from(&event)?;
-            EventDataView::Upgrade {
-                write_set: BytesView::from(upgrade_event.write_set()),
+            let upgrade_event = AdminTransactionEvent::try_from(&event)?;
```
xli, line 321: upgrade_event => event

types/src/account_config/events/upgrade.rs
```diff
@@ -5,24 +5,24 @@ use anyhow::Result;
 use move_core_types::move_resource::MoveResource;
```
xli, line 5: probably also rename this file

json-rpc/tests/integration_test.rs
```diff
@@ -619,20 +619,23 @@ fn create_test_cases() -> Vec<Test> {
...
+                assert!(committed_time != 0);
                 assert_eq!(
                     result["events"],
                     json!([
                         {
                             "data":{
-                                "type": "upgrade",
-                                "write_set": write_set_hex
+                                "type": "admintransaction",
+                                "committed_time_seconds": committed_time,
```
xli, line 638: just realized it seems you created a new format to represent timestamp in seconds, we used timestamp, timestampusec, timestamp_secs and timestamp_usecs

I suggest we name it committed_timestamp_secs

xli (approved):

dariorussi (approved):
good to go!

sausagee (approved):
LGTM
/land

PR #6455, opened 2020-10-08
[Vault] Fix error message handling for http status codes and text.

JoshLind:
### Motivation

This PR updates the vault client to better display http status error codes and text. Previously, the error code was not displayed and the status text was derived from 'into_string()' which produces a string value of the response body (which may be empty, and thus not helpful). Instead, we now display the error code as well as the status text directly and the response body.

Yes.

## Test Plan

Running the operational tool locally to reproduce and fix the error helped verify the correct behavior. 

This is the output without the fix (on a 404 key not found error):

```
{
  "Error": "Failed to read 'treasury_complianc' from validator storage: Internal error: Http error: {\"errors\":[]}\n"
}
```

This is the output after the fix:

```
{
  "Error": "Failed to read 'treasury_complianc' from validator storage: Internal error: Http error, status code: 404, status text: Not Found, body: {\"errors\":[]}\n"
}
```

## Related PRs

None.

### DIFF
--- a/secure/storage/src/error.rs
+++ b/secure/storage/src/error.rs
@@ -57,7 +57,7 @@ impl From<libra_vault_client::Error> for Error {
     fn from(error: libra_vault_client::Error) -> Self {
         match error {
             libra_vault_client::Error::NotFound(_, key) => Self::KeyNotSet(key),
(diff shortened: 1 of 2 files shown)


davidiw (approved):
Thanks!

PR #6457, opened 2020-10-08
[mempool] count and reject unrecognized ACKs

christinacelee:
## Motivation

there is a chance that random peers on the network send random ACKs to mempool. This PR helps mempool to be defensive against it by counting and making less assumptions about ACKs it receives

### DIFF
--- a/mempool/src/shared_mempool/peer_manager.rs
+++ b/mempool/src/shared_mempool/peer_manager.rs
@@ -427,14 +427,21 @@ impl PeerManager {
             id
         } else {
             counters::INVALID_ACK_RECEIVED_COUNT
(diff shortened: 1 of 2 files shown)


mempool/src/shared_mempool/peer_manager.rs
```diff
@@ -455,7 +455,14 @@ impl PeerManager {
             .lock()
             .expect("failed to acquire peer_info lock");
 
-        let sync_state = peer_info.get_mut(&peer).expect("missing peer sync state");
```
davidiw, line 458: :O how did I miss this in my review

good find!

davidiw (approved):
/land

PR #6462, opened 2020-10-08
removing unwrap in a function of type result. using unstable sort ins…

xvschneider:
**Motivation**
Using unwrap in a function that returns result is not recommended in the libra codebase, as it promotes recoverable errors to unrecoverable errors and it leads to panic. (More info [here](https://rust-lang.github.io/rust-clippy/master/index.html#unwrap_in_result)).

Hence I am removing the usage of unwrap.

In the future, this lint will be added to clippy to prevent this from happening in the first place.

**Have you read the Contributing Guidelines on pull requests?**
Yes

Furthermore, other I replaced the usage of sort with unstable sort in a function as per this [guideline ](https://rust-lang.github.io/rust-clippy/master/#stable_sort_primitive)

**Test Plan**
No new functionality added, so no new tests needed, but I ran existing tests to ensure nothing is breaking.
cargo build
cargo xtest

### DIFF
--- a/common/logger/src/struct_log.rs
+++ b/common/logger/src/struct_log.rs
@@ -94,13 +94,14 @@ impl Write for TcpWriter {
         // Attempt to write, and if it fails clear underlying stream
         // This doesn't guarantee a message cut off mid send will work, but it does guarantee that
         // we will connect first
-        let mut stream = self.stream.as_ref().unwrap();
-        let result = stream.write(buf);
-        if result.is_err() {
-            self.stream = None;
-        }
-
-        result
+        self.stream
+            .as_mut()
+            .ok_or_else(|| io::Error::new(io::ErrorKind::NotConnected, "No stream"))
+            .and_then(|stream| stream.write(buf))
+            .map_err(|e| {
+                self.stream = None;
+                e
+            })
     }
 
     fn flush(&mut self) -> io::Result<()> {
--- a/common/proptest-helpers/src/repeat_vec.rs
+++ b/common/proptest-helpers/src/repeat_vec.rs
@@ -121,7 +121,7 @@ impl<T> RepeatVec<T> {
     /// Ignores any out of bounds indexes.
     pub fn remove_all(&mut self, logical_indexes: impl IntoIterator<Item = usize>) {
         let mut logical_indexes: Vec<_> = logical_indexes.into_iter().collect();
-        logical_indexes.sort();
+        logical_indexes.sort_unstable();
         logical_indexes.dedup();
         self.remove_all_impl(logical_indexes)
     }
--- a/crypto/crypto/src/multi_ed25519.rs
+++ b/crypto/crypto/src/multi_ed25519.rs
@@ -430,7 +430,10 @@ impl TryFrom<&[u8]> for MultiEd25519Signature {
             return Err(CryptoMaterialError::WrongLengthError);
         }
 
-        let bitmap = bytes[length - BITMAP_NUM_OF_BYTES..].try_into().unwrap();
+        let bitmap = match bytes[length - BITMAP_NUM_OF_BYTES..].try_into() {
+            Ok(bitmap) => bitmap,
+            Err(_) => return Err(CryptoMaterialError::DeserializationError),
+        };
         if bitmap_count_ones(bitmap) != num_of_sigs as u32 {
             return Err(CryptoMaterialError::DeserializationError);
         }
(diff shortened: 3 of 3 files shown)


common/proptest-helpers/src/repeat_vec.rs
```diff
@@ -121,7 +121,7 @@ impl<T> RepeatVec<T> {
     /// Ignores any out of bounds indexes.
     pub fn remove_all(&mut self, logical_indexes: impl IntoIterator<Item = usize>) {
         let mut logical_indexes: Vec<_> = logical_indexes.into_iter().collect();
-        logical_indexes.sort();
+        logical_indexes.sort_unstable();
```
mimoo, line 124: why?

crypto/crypto/src/multi_ed25519.rs
```diff
@@ -502,13 +505,21 @@ impl Signature for MultiEd25519Signature {
...
-            ));
-        }
+        match bitmap_last_set_bit(self.bitmap) {
+            Some(last_bit) if last_bit as usize > public_key.length() => {
+                return Err(anyhow!(
+                    "{}",
+                    CryptoMaterialError::BitVecError("Signature index is out of range".to_string())
+                ))
+            }
+            Some(last_bit) => last_bit,
```
mimoo, line 515: since you don't return anything, you don't want to return `last_bit` here but `()`

mimoo (approved):
left you some comments, otherwise LGTM

xvschneider:
> left you some comments, otherwise LGTM

thanks David 👍

crypto/crypto/src/multi_ed25519.rs
```diff
@@ -502,13 +505,21 @@ impl Signature for MultiEd25519Signature {
...
+                ))
+            }
+            Some(_) => (),
+            None => {
+                return Err(anyhow!(
+                    "{}",
+                    CryptoMaterialError::BitVecError("Failed to find last set bit".to_string())
+                ))
+            }
+        };
```
mimoo, line 522: you could actually simplify this even more by compressing this into two cases:

```rust
Some(last_bit) if last_bit as usize <= public_key.length() => (),
_ => return Err(...),
```

xvschneider: yep good idea

mimoo:
cc @davidiw can we get this in?

common/logger/src/struct_log.rs
```diff
@@ -94,13 +94,18 @@ impl Write for TcpWriter {
...
+            let result = stream.write(buf);
+            if result.is_err() {
+                self.stream = None;
+            }
+            result
+        } else {
+            Err(io::Error::new(
+                io::ErrorKind::NotConnected,
+                "Can't write to stream, not connected",
+            ))
```
davidiw, line 107: can we replace this with the functor example I gave in the previous PR?

https://github.com/libra/libra/pull/6410#discussion_r501445867

mimoo: I'm wondering why functional over imperative? The problem with rust is that you can write both, and that it ends up being a personal preference. I'd be more interested in a principled approach where we advise one style over the other.

davidiw: Well look how much more verbose this is versus what was provided, that's one principle I use. Another principle is to avoid nesting. Functional has no statement based nesting whereas this has if within an if.

xvschneider: I'm trying to do this:
  self.stream
        .as_mut()
        .ok_or_else(|| io::Error::new(io::ErrorKind::NotConnected,"Can't write to stream, not connected"))
        .and_then(|stream|  stream.write(buf).map_err(|e| { self.stream = None; e }));

but I'm getting this

mismatched types
  expected enum `std::result::Result<usize, std::io::Error>`
found unit type `()`rustc(E0308)
struct_log.rs(88, 8): implicitly returns `()` as its body has no tail or `return` expression

That's because the expected return value is a result<usize> but we're not returning in that functional statement. I tried looking online on how to do it but nothing worked. do you have any suggestions?

xvschneider: I've also tried:
```
        self.stream
        .as_mut()
        .ok_or_else(|| io::Error::new(io::ErrorKind::NotConnected,"Can't write to stream, not connected"))
        .and_then(|stream|  { let result = stream.write(buf).map_err(|e| { self.stream = None; e }) ; result } );
```
but it also has the same problem

mimoo: > Well look how much more verbose this is versus what was provided

I think the trade-off is verbosity vs readability. So I don't find it that clear cut.

mimoo: @xvschneider I don't think you can write `self.X` in the closure you pas to `map_err`. You also don't return anything to your function (due to the `;` at the end of your statement)

davidiw: this worked no problem:
```
        self.stream
            .as_mut()
            .ok_or_else(|| io::Error::new(io::ErrorKind::NotConnected, "No stream"))
            .and_then(|stream| stream.write(buf))
            .map_err(|e| {
                self.stream = None;
                e
            })
```

I think your issue was adding the ";" at the end of the call.

let's get a 3rd vote, cc: @bmwill

bmwill: Yeah I'm ok with this, I believe Greg already signed off on this in the last iteration. Lets just make sure to either squash our commits or apply the `bors-squash` label

xvschneider: I can't add the label, if you have permissions can you add bors-squash?

davidiw (approved):

davidiw:
hint: always run cargo xfmt before submitting a PR

xvschneider:
> hint: always run cargo xfmt before submitting a PR

yep you're right, forgot to do that. maybe we can make a git hook that runs xfmt ?

xvschneider:
@davidiw  can you run land again please? Thanks!

PR #6465, opened 2020-10-08
[breaking] use mask for block metadata

zekun000:
## Motivation

Breaking nature

This commit changes LibraBlock in stdlib as well as the Transaction::BlockMetadata which requires a data wipe but has no impact on clients.

Justification

The BlockMetadata uses account address to represent proposer and voters, which is not efficient given there's all known to be in validator set. This change proposes to use a vec<bool> as bitmask to represent the same information but achieves better efficiency.

With a 30 nodes network, the raw txn size changes from 403 bytes to 48 bytes (~10x saving). The compressed rocksdb size changes from ~67 bytes to ~18 bytes (~3x saving).

The major benefit from this change would be the network bandwidth required to serve public fullnode for state sync.

Y

## Test Plan

existing tests.

## Related PRs


consensus/src/liveness/leader_reputation.rs
```diff
@@ -111,15 +107,18 @@ impl ActiveInactiveHeuristic {
...
-            set
-        });
-        candidates
-            .iter()
-            .map(|author| {
-                if set.contains(&author) {
+        let mut mask = vec![false; candidates.len()];
+        // set mask if it's in any of the event and ignore those with different length
+        for event in history {
+            if event.participants().len() == mask.len() {
```
sblackshear, line 113: Will this do the right thing if `history` contains validator sets of the same size, but with different addresses? For example, if

- `candidates` is `[0xB, 0xC]`
- `history` is `[NewBlockEvent { votes: [0xA, 0xC] }, NewBlockEvent { votes: [0xA, 0xD] }]`

, it seems like `0xB` will get `active_weight` even though they have not voted on anything in `history`.

zekun000: great question! it's possible but unlikely since it needs at least two epochs to achieve same set with one different addresses. given it's just heuristic I don't think it matters much, if 0XB is not active its weight will drop soon.

sblackshear: Makes sense. I think we could get the exact answer by comparing `candidates` and `event.participants`? But might be overkill if it's unlikely in practice + a heuristic anyway.

zekun000: it's also unlikely because we filtered the history by round, and round is reset every epoch. so most times it'd be 
`NewBlockEvent { round: 123456789, votes: [0xA, 0xC] } NewBlockEvent { round: 1, votes: [0xA, 0xD]}` and the first event will be filtered out.

zekun000: `event.participants` is just a vector of bool without actual addresses, in order to get the exact answer we'd have to pass the epoch number in NewBlockEvent so that we can filter by that but it seems overkill to me

types/src/block_metadata.rs
```diff
@@ -22,54 +20,38 @@ use serde::{Deserialize, Serialize};
...
 pub struct BlockMetadata {
-    id: HashValue,
     round: u64,
     timestamp_usecs: u64,
-    // The vector has to be sorted to ensure consistent result among all nodes
-    previous_block_votes: Vec<AccountAddress>,
-    proposer: AccountAddress,
+    // masks on the current validator set
+    participants: Vec<bool>,
+    is_nil: bool,
```
aching, line 30: Just pointing out that one thing we lose here is the ability to know the proposer from the voter from on-chain state.  This means that externally there isn't visibility into the fairness of proposers in an epoch.  How can the public ensure that it isn't just one validator proposing all the committed blocks?

One option could be to keep the AccountAddress (16 bytes) for the proposer or an index (4 bytes for u32) into the current validator set to retain this information.

consensus/src/liveness/leader_reputation.rs
aching: I agree these are edge cases, but that behavior Sam pointed out is odd (and basically random), whereas before we relied on correct historical data in all cases.

@zekun000 in any case, we'll have to update https://github.com/libra/libra/tree/master/specifications/consensus#leaderreputation with your point about the round reset by epochs have an impact on the filtering.

zekun000: I don't think it changes anything in the spec, it's all implementation details. The spec doesn't talk about how to represent a committed block at all.

types/src/block_metadata.rs
zekun000: the index can't handle the nil case well. I can only think of using u32::max to represent a nil. @davidiw pointed out it too, but I don't think we need to expose this information, it's just nice to have.

aching: While I appreciate the goal of saving space (network/storage), this current proposal loses information on the proposer, adds more work for the consumer to match against the validator set, and adds the possibility of unexpected edge cases.

Is it worth it?  Given that these txns should be a small percent of the workload once this network is more widely used (e.g. 10s or 100s of tps) and that in the future we don't need the full txn history to sync state, I prefer the existing method.

Network compression can be added to mimic the same improvements as in rocksdb and we can experiment with libstdz and other compression techniques.

A compromise might be to just keep everything in indices of the validator set but as you mention, it's a bit unclean as well.

davidiw:
Thank you for driving this exciting initiative.

types/src/block_metadata.rs
kchalkias: Re the index and null, I’ve seen similar models where the bitmask starts counting from index=1 and the first but always be zero simulating an imaginary “null” participant. Then  leader_index=0 represents the null.

PR #6467, opened 2020-10-08
[docgen] Update documentation generator to support multiple root templates

wrwg:
This updates the documentation generator to support not only one root template but multiple ones. This allows us to write different markdown files for top-level documentation of different aspects of the framework which are fully processed for syntax highlighting and cross-references.

Also adds a new `./language/stdlib/transaction_scripts/spec_documentation_template.md` file which is the top-level page for overview documentation of specifications in the framework. This is a strawman which will be populated in the next few days.

*Why this change is needed*

We want to create a distinct document which introduces readers into the specification and verification approach of the framework, for audiences like auditors. While we do have already technical documentation, this appears to low-level for these readers.

*Why this change is safe*

Only code in `./language/move-prover/docgen` and `./language/stdlib/src/lib.rs` is effected. This code does not run in production.

## Motivation

Finalize documentation

Yes

## Test Plan

NA

## Related PRs

NA

### DIFF
--- a/language/move-prover/docgen/src/docgen.rs
+++ b/language/move-prover/docgen/src/docgen.rs
@@ -112,10 +112,10 @@ pub struct DocgenOptions {
     pub output_directory: String,
     /// In which directories to look for references.
     pub doc_path: Vec<String>,
-    /// An optional path to a file containing a template for a root document for the generated
+    /// A list of paths to files containing templates for root documents for the generated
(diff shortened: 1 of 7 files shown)


DavidLDill (approved):

tzakian (approved):

dariorussi (approved):
good to go!

PR #6469, opened 2020-10-08
[Smoke Tests] Some cleans around operator tooling tests and naming.

JoshLind:
## Motivation

* Note: this PR affects the smoke tests only. Moreover, it doesn't change smoke test behaviors, but rather only offers refactors and clean ups.

There's still a number of missing smoke tests that need to be added to the operational tooling. However, before doing this, there's a bunch of tech-debt, messy/duplicate code and inconsistencies in the smoke tests that we should probably clean up first. This PR offers several small clean ups and refactors that'll make adding the additional tests less cumbersome in the future. 

This PR offers several commits:
1. First, we refactor a lot of duplicate logic in the operational tooling tests into a helper function to prevent redundant code patterns. We also ensure that multiple nodes (> 1) are spawned only when required.
2. Second, we provide a set of libra swarm related utils to help provide a consistent and cleaner interface for performing common operations in tests (e.g., loading configs, creating clients, accessing storage). This also helps clean up the SmokeTestEnvironment interface.
3. Third, we provide some additional cleanups to the genesis smoke test. This test is still quite messy, but hopefully a little clearer now.
4. Finally, we clean up some naming around public and validator full nodes in tests.

Note: while this PR makes our code somewhat cleaner.. there's still a lot of improvements to do. 

Yes.

## Test Plan

All smoke tests still pass.

## Related PRs

None.

### DIFF
--- a/testsuite/smoke-test/src/client.rs
+++ b/testsuite/smoke-test/src/client.rs
@@ -2,9 +2,11 @@
 // SPDX-License-Identifier: Apache-2.0
 
 use crate::test_utils::{
-    compare_balances, load_libra_root_storage, setup_swarm_and_client_proxy, test_smoke_script,
+    compare_balances,
+    libra_swarm_utils::{get_op_tool, load_libra_root_storage},
+    setup_swarm_and_client_proxy, test_smoke_script,
 };
-use libra_config::config::NodeConfig;
+use debug_interface::NodeDebugClient;
 use libra_trace::trace::trace_node;
 use libra_types::{
     account_address::AccountAddress, account_config::testnet_dd_account_address,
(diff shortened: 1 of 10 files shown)


testsuite/smoke-test/src/client.rs
```diff
@@ -101,7 +103,7 @@ fn test_basic_restartability() {
 
 #[test]
 fn test_client_waypoints() {
-    let (env, mut client_proxy) = setup_swarm_and_client_proxy(3, 1);
+    let (env, mut client_proxy) = setup_swarm_and_client_proxy(4, 1);
```
davidiw, line 106: I think 1 validator is sufficient here

testsuite/smoke-test/src/full_nodes.rs
```diff
@@ -14,146 +14,138 @@ fn test_full_node_basic_flow() {
     // launch environment of 4 validator nodes and 2 full nodes
     let mut env = SmokeTestEnvironment::new(4);
```
davidiw, line 15: not gonna get much more out of 4 nodes than 1

davidiw (approved):
would love to see a PR reasoning about the number of nodes used in each test.

JoshLind:
Thanks @davidiw! I'll follow up with a PR going through the number of nodes per test. Definitely seems like something that needs to be looked at! 😄

/land

PR #6473, opened 2020-10-08
[consensus][observability] Capture time as a gauge

aching:
Continually capture the time of consensus process via the event loop to ensure that clock skew between validators is reasonable and to find any unusual (possibly byzantine) clock behavior.


davidiw (approved):
:trollface:

PR #6474, opened 2020-10-08
[move-cli] print compressed addresses in 'move view'

skrtskrtfb:
## Motivation

When running 'move view', addresses are shortened from left to right. This can often lose information, so this PR instead shortens the addresses by stripping all leading zeros. For example, 0x0000000000000000000000000000000a becomes 0xa.

Yes

## Test Plan

Steps to repro:

1. cd ~/libra/language/tools/move-cli/tests/testsuite/module_run_view
2. NO_MOVE_CLEAN=1 cargo xtest
3. move view move_data/0x0000000000000000000000000000000a/resources/0x00000000000000000000000000000002::Test::R.lcs

Old output:
resource 00000000::Test::R {
  i: 6
}

New output:
resource 0x2::Test::R {
  i: 6
}

## Discussion

I wrote a separate function, short_str_lossless, since it seems like short_str is used elsewhere in the codebase and the functionality of short_str_lossless is somewhat different. Furthermore, the ShortHexStr struct seems to enforce a fixed size, and for the purposes of this PR I need a variable display string size, so it didn't seem correct to modify this. I would appreciate feedback on my approach here, in case there is a better way to implement this.

### DIFF
--- a/language/move-core/types/src/account_address.rs
+++ b/language/move-core/types/src/account_address.rs
@@ -47,7 +47,11 @@ impl AccountAddress {
 
     pub fn short_str_lossless(&self) -> String {
         let hex_str = hex::encode(&self.0).trim_start_matches('0').to_string();
-        if hex_str.is_empty() { "0".to_string() } else { hex_str }
+        if hex_str.is_empty() {
+            "0".to_string()
+        } else {
+            hex_str
+        }
     }
 
     pub fn to_vec(&self) -> Vec<u8> {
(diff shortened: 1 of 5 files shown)


sblackshear:
> short_str is used elsewhere in the codebase and the functionality of short_str_lossless is somewhat different

Took a look around and wow--I didn't realize `short_str` was quite so pervasive! I think we probably want to replace every use of `short_str` under `libra/language` with `short_str_lossless`, but that can be done separately. This is a great start.

skrtskrtfb:
StructTag is also used fairly regularly, so it might even make sense to do all the formatting in resource-viewer and avoid causing any compatibility problems, but this seems like the least scalable option. If this is the preferred format scheme throughout the codebase, we can probably leave the changes in StructTag itself.

language/move-core/types/src/account_address.rs
```diff
@@ -45,6 +45,11 @@ impl AccountAddress {
         )
     }
 
+    pub fn short_str_lossless(&self) -> String {
```
sblackshear, line 48: It would probably be good to add some Rust unit tests for this (putting `#[test]`'s at the bottom of the file will work just fine for this).

tzakian (approved):
This looks great to me! As Sam mentioned, this will be a super useful thing to use throughout the code in `language` once this gets in :)

sblackshear (approved):

PR #6479, opened 2020-10-09
[cli] Add input checking to verify waypoint provided

gregnazario:
Now returns the slightly confusing error message:
```
error: The following required arguments were not provided:
    --waypoint <waypoint>
    --waypoint_url <waypoint_url>
```

### DIFF
--- a/testsuite/cli/src/main.rs
+++ b/testsuite/cli/src/main.rs
@@ -64,13 +64,15 @@ struct Args {
     #[structopt(
         name = "waypoint",
         long,
-        help = "Explicitly specify the waypoint to use"
+        help = "Explicitly specify the waypoint to use",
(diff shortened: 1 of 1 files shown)


mgorven (approved):

testsuite/cli/src/main.rs
```diff
@@ -109,7 +109,7 @@ fn main() {
                     panic!("Failure to retrieve a waypoint from {}: {}", url_str, e)
                 })
             })
-            .unwrap()
+            .expect("Must provide either '--waypoint' or '--waypoint-url'")
```
davidiw, line 112: Use this instead: https://github.com/libra/libra/blob/master/config/management/src/config.rs#L21

davidiw (approved):

PR #6484, opened 2020-10-09
[Key Manager] Log the key manager config on startup for better observability.

JoshLind:
## Motivation

This PR updates the key manager to log its config on startup. In many cases, this is useful to provide better observability to parties who don't have access to the key manager deployment directly but do have access to the logs (e.g., the blockchain on-call who might help debug issues, or subject matter experts). 

Note: it seems this is already happening in most of the processes we spawn (e.g., safety rules, libra-node). It might make sense for this to be a requirement in general (just as long as we don't think anything sensitive will be logged -- but I'd think this is unlikely).

Yes.

## Test Plan

Running the key manager locally shows the config now being logged.

## Related PRs

None.

### DIFF
--- a/secure/key-manager/src/main.rs
+++ b/secure/key-manager/src/main.rs
@@ -51,6 +51,11 @@ fn main() {
 }
 
 fn create_and_execute_key_manager(key_manager_config: KeyManagerConfig) -> Result<(), Error> {
+    info!(
+        LogSchema::new(LogEntry::Initialized).event(LogEvent::Pending),
+        key_manager_config = key_manager_config
+    );
+
     let json_rpc_endpoint = key_manager_config.json_rpc_endpoint;
     let libra_interface = JsonRpcLibraInterface::new(json_rpc_endpoint.clone());
     let storage: Storage = (&key_manager_config.secure_backend)
@@ -71,5 +76,6 @@ fn create_and_execute_key_manager(key_manager_config: KeyManagerConfig) -> Resul
     info!(LogSchema::new(LogEntry::Initialized)
         .event(LogEvent::Success)
         .json_rpc_endpoint(&json_rpc_endpoint));
+
     key_manager.execute()
 }


secure/key-manager/src/logging.rs
```diff
@@ -14,6 +15,8 @@ pub struct LogSchema<'a> {
     consensus_key: Option<&'a Ed25519PublicKey>,
     json_rpc_endpoint: Option<&'a str>,
     #[schema(display)]
+    key_manager_config: Option<&'a KeyManagerConfig>,
+    #[schema(display)]
```
gregnazario, line 19: You can save yourself the Display above, by doing `#[schema(debug)]`

secure/key-manager/src/main.rs
```diff
@@ -51,6 +51,10 @@ fn main() {
 }
 
 fn create_and_execute_key_manager(key_manager_config: KeyManagerConfig) -> Result<(), Error> {
+    info!(LogSchema::new(LogEntry::Initialized)
+        .event(LogEvent::Pending)
+        .key_manager_config(&key_manager_config));
```
gregnazario, line 56: I'm of the opinion to not stick it in the Schema struct since you're using it exactly once, but @bmwill might have a different opinion.  You could just do a  

```
info!(
  LogSchema(...).event(...),
  key_manager_config = ?key_manager_config
);
```

gregnazario (approved):
Just stylistic choices as comments, but you should probably just not implement display since it's just running the debug implementation.

secure/key-manager/src/logging.rs
JoshLind: Aah, nice! Didn't know we supported using the debug format. Let's go with that 😄

secure/key-manager/src/main.rs
JoshLind: Yeah, don't feel super strongly here either. I'll follow the pattern of avoiding a new schema struct because it requires less code.. Seems this is what LSR and LibraNode do anyway. 😄

JoshLind:
Thanks @gregnazario 😄 I'll wait now wait for the gatekeepers... 🚢

davidiw (approved):
Surprised we don't want to be consistent with how we are logging these, whether that means modifying the other two or having this one mimic their behavior.

JoshLind:
Thanks @davidiw! 😄 

/land

JoshLind:
> Surprised we don't want to be consistent with how we are logging these, whether that means modifying the other two or having this one mimic their behavior.

Yeah.. I don't have strong feelings here, but if we do settle on a standard, I'll happily follow along!

gregnazario:
> > Surprised we don't want to be consistent with how we are logging these, whether that means modifying the other two or having this one mimic their behavior.
> 
> Yeah.. I don't have strong feelings here, but if we do settle on a standard, I'll happily follow along!

I'm pretty sure I created the other two :D

PR #6485, opened 2020-10-10
[Smoke Tests] Various small clean ups to smoke tests

JoshLind:
## Motivation

This PR offers some clean ups and refactors to the smoke tests through several commits:
1. First, we move the "test_smoke_script" helper function into the client file and make it private. This avoids having it in the shared utils file. In addition, we remove the call to "test_smoke_script" from one of the fullnode tests, as the functionality being tested is redundant for this test.
2. Second, we go through all smoke tests to ensure that either only 1 or 3f+1 nodes are spawned.
3. Third, we clean up the client.rs file by renaming some tests and refactoring a few variables. (where f is the number of faulty or killed nodes in the test).
4. Finally, we go through all the tests and rename "swarm" to "env" (to prevent confusion between the smoke test environment and the various libra swarms that it contains), and "client_proxy" to "client".

(This PR takes us a few small steps towards having cleaner, less redundant and more helpful smoke tests).

Yes.

## Test Plan

All tests pass locally.

## Related PRs

None.

### DIFF
--- a/testsuite/smoke-test/src/client.rs
+++ b/testsuite/smoke-test/src/client.rs
@@ -15,21 +15,20 @@ use libra_types::{
 };
 
 #[test]
-fn smoke_test_single_node() {
-    let (_swarm, client_proxy) = setup_swarm_and_client_proxy(1, 0);
-    test_smoke_script(client_proxy);
+fn test_create_mint_transfer() {
+    let (_env, client) = setup_swarm_and_client_proxy(1, 0);
+    check_create_mint_transfer(client);
 }
 
 // Test if we commit not only user transactions but also block metadata transactions,
 // assert committed version > # of user transactions
 #[test]
-fn smoke_test_single_node_block_metadata() {
-    let (swarm, mut client_proxy) = setup_swarm_and_client_proxy(1, 0);
-    // just need an address to get the latest version
+fn test_create_mint_transfer_block_metadata() {
+    let (env, mut client) = setup_swarm_and_client_proxy(1, 0);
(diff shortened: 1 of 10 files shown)


testsuite/smoke-test/src/client.rs
```diff
@@ -14,27 +15,20 @@ use libra_types::{
...
-    // just need an address to get the latest version
+fn test_create_mint_transfer_block_metadata() {
+    let (env, mut client) = setup_swarm_and_client_proxy(1, 0);
+    check_create_mint_transfer(env.get_validator_client(0, None));
+
     let address = AccountAddress::from_hex_literal("0xA550C18").unwrap();
-    // this script does 4 transactions
-    test_smoke_script(swarm.get_validator_client(0, None));
-    let (_account, version) = client_proxy
+    let (_account, version) = client
```
davidiw, line 31: but now you have a magic 4 for version below....

davidiw: frankly I don't see thae value of that check below (40-44) you would have failed in the function call, but if you're using this function call more than once, we should just kill this test.

davidiw (approved):

testsuite/smoke-test/src/client.rs
JoshLind: Yeah... I felt the test was a bit unnecessary, too. What I've done is added the comment back into the test and merged the two overlapping smoke tests to avoid the redundancy.

JoshLind:
Thanks @davidiw!

/land

PR #6494, opened 2020-10-12
[state sync] fault injection

christinacelee:
# Motivation

Add fault injection for state sync

Needed to comment out json-rpc fault-injection in `fullnode.yaml`...
Also had to increase some health check intervals but not checking these changes in, just as FYI for future

Results for fault-injection for validator-only:
Experiment Result: all up : 717 TPS, 6344 ms latency, 18100 ms p99 latency, no expired txns

Results for fault-injection for validator and fullnode:
Experiment Result: all up : 716 TPS, 6309 ms latency, 19050 ms p99 latency, (!) expired 79 out of 172065 txns

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6115,6 +6115,7 @@ dependencies = [
  "executor",
  "executor-test-helpers",
  "executor-types",
+ "fail",
  "futures 0.3.6",
  "itertools 0.9.0",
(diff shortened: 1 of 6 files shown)


zekun000 (approved):

davidiw (approved):
/land

PR #6495, opened 2020-10-12
[mempool] fix counter for pending broadcasts

christinacelee:
## Motivation

fix counter for counting pending broadcasts to count number of outstanding batches instead of number of total broadcasts (the current way double-counts retried broadcasts)


davidiw (approved):
/land

PR #6497, opened 2020-10-12
[mempool] inject failure

christinacelee:
## Motivation

Fault injection for mempool

Result for benchmark with network send failure enabled (+ consensus + execution error): Experiment Result: all up : 726 TPS, 6255 ms latency, 9400 ms p99 latency, (!) expired 3 out of 174375 txns

For the other errors (Db error, vm validation error), cluster test benchmark doesn't run because mint txns fail to be committed) - this is expected in current CT setup since there is no retry functionality in CT client


zekun000 (approved):

dariorussi:
what is the error you are getting?
I run into the minting and non retry story for the VM fault injection too and we need to fix that or it would be difficult to play with fault injection well.
Please let us know about the error and we go from there

christinacelee:
ah yes should have provided more context!
so this PR adds 3 failpoints for mempool: network send error, DB read errors (vm_validator::get_account_sequence_number), and vm validation error. 

the cluster test benchmark experiment, iirc, runs by first submitting mint transactions, checking whether those mint transactions were committed, and then proceeds to submit p2p transactions. when I run the experiment with errors added with DB read error and vm validation error enabled in the corresponding config yaml files, it fails to get passed the minting phase - which is expected because the injected DB read and vm validation error would reject such transactions from mempool. In the current CT setup, the actor submitting such transactions only checks whether these transactions were committed and returns early from the experiment too soon. to me, this result is more of a symptom of how CT works (no resubmit functionality after tx is not committed)

I think for now this PR can add the failpoints and later enable them in the yaml file (currently i only enable the network send error which doesn't face this issue) once CT has better support for automatically retrying failed txns - but we can open this point up for discussion 😄 

cc @snakech for thoughts on the tx-emitter

testsuite/cluster-test/src/cluster_swarm/configs/validator.yaml
```diff
@@ -66,7 +66,8 @@ failpoints:
   executor::vm_execute_chunk: 1%return
   executor::commit_chunk: 1%return
   executor::vm_execute_block: 1%return
-  #executor::commit_blocks: 0.01%return
+  executor::commit_blocks: 0.01%return
```
dariorussi, line 69: should this be uncommented?

davidiw (approved):
approval based on @zekun000's approval

dariorussi (commented):
no more context needed but thanks! With my comment I was just confirming the issue you run into and saying we should fix that at some point, certainly not implying we should fix it in this PR. I adopted the same strategy of adding fail point and we will enable them at a later point. So all good here. 
There still seem to be error in testing (or similar) but I assume you will take care of that

testsuite/cluster-test/src/cluster_swarm/configs/validator.yaml
christinacelee: i'm not sure.. @lightmark

lightmark: let's chain our @... 
This line was intentional based on some injection glitches. @zekun000 solved?

christinacelee: the uncomment was accidental, left over from when i was playing around with it locally. will revert for now

zekun000: I think it caused the cluster test to fail because this panics the pod (with 1% maybe?)

snakech:
> ah yes should have provided more context!
> so this PR adds 3 failpoints for mempool: network send error, DB read errors (vm_validator::get_account_sequence_number), and vm validation error.
> 
> the cluster test benchmark experiment, iirc, runs by first submitting mint transactions, checking whether those mint transactions were committed, and then proceeds to submit p2p transactions. when I run the experiment with errors added with DB read error and vm validation error enabled in the corresponding config yaml files, it fails to get passed the minting phase - which is expected because the injected DB read and vm validation error would reject such transactions from mempool. In the current CT setup, the actor submitting such transactions only checks whether these transactions were committed and returns early from the experiment too soon. to me, this result is more of a symptom of how CT works (no resubmit functionality after tx is not committed)
> 
> I think for now this PR can add the failpoints and later enable them in the yaml file (currently i only enable the network send error which doesn't face this issue) once CT has better support for automatically retrying failed txns - but we can open this point up for discussion 😄
> 
> cc @snakech for thoughts on the tx-emitter

My I ask some clarification about resubmit? does it means a tx from same sender to receiver with same seq number?

christinacelee:
> My I ask some clarification about resubmit? does it means a tx from same sender to receiver with same seq number?

yes, so in the minting phase in the CT experiment flow, after the tx-emitter submits txns and checks whether they are committed or not, the experiment returns early. I think it'd be cool if there was a feature if a tx was not committed, the tx-emitter can try submitting that same txn again! this would help with testing resilience in vm-validation failures especially in CT

PR #6500, opened 2020-10-12
[libra framework] Add additional tests to get coverage back up

tzakian:
Adds a number of tests to increase testing coverage of the Libra Framework.

A couple small changes to comments in module documentation, but the compiled bytecode is unchanged.


dariorussi (approved):
good to go!

PR #6504, opened 2020-10-13
Add client side redirect from old rust crate docs

JoelMarcey:
Also, remove some unused config options and scripts

## Motivation

Google hasn't totally [indexed](https://bit.ly/34QmcUL) the new links for the Rust crate docs. Set up a client side redirect until that fully takes place.

Also, there were some commented out configs we weren't using. Remove those.

Yes

## Test Plan

Live

## Related PRs

N/A


developers.libra.org/website/pages/404.html
```diff
@@ -1,7 +1,21 @@
-You will be redirected shortly.
+<html>
+  <head>
+    <script type = "text/javascript">
+      var redirects = {
+        'libra.github.io/libra': 'developers.libra.org/docs/rustdocs',
```
bmwill, line 5: hmm this might be difficult because I really think we only want to redirect on libra.github.io/libra/<CRATE_NAME> and not just on libra.github.io/libra since that is now the landing page for the website

JoelMarcey: @bmwill Actually, developers.libra.org is the landing page, right?

If you try `libra.github.io/libra` now, it does the current redirect that is in the `404.html` that I am updating here.

bmwill: IIUC if you hit a 404 then it will redirect to `developers.libra.org/blog`. When I go to `libra.github.io/libra` I do not hit a 404 but instead just immediately go to the landing page `developers.libra.org` (because gh-pages is setup to do that on the backend, right?).

JoelMarcey: Yeah, you might be right here. I know `developers.libra.org/blog` doesn't exist any longer, so that gets redirected to https://libra.org/en-US/blog/. 

So the question is --

Given that https://libra.github.io/libra is not the canonical domain, do we care if we override the default redirection of that domain from the landing page to the Rust crate docs, at least in the short-term until we figure out why Google is taking so long to update the link indices?

bmwill: Yeah my only concern is that if we do add a redirect in, then if you do browse off of `https://libra.github.io/libra` then it wont match the actual, current state of the website. I don't have a strong opinion but just wanted to make sure we understand what this will do to the dev site (at least in the short term)

JoelMarcey: This is a tough one -- honestly, I am not sure it really is a huge issue that needs resolving yet. On the other hand, we want to make sure people can get the info they need and not be misled.

I am going to meet with @ericnakagawa this afternoon to try to figure out the right way to do this.

bmwill: Sounds good, FWIW I personally think its fine to just wait till Google reindexes since we made this change a while ago and this is the first issue we've heard

JoelMarcey: I tend to agree that it is not a showstopping issue.

The weird thing is that if you look at this search: https://www.google.com/search?q=libra_jellyfish_merkle

Notice how the first result is a link to the old site. But the 3rd and 4th results link to the new site.

It's like Google has indexed content on the newly linked pages, but not the actual links themselves.

davidiw:
Any plans here?

davidiw:
Closing out, feel free to re-open.

PR #6505, opened 2020-10-13
support genesis waypoint

zekun000:
## Motivation

(Write your answer here.)

## Test Plan

## Related PRs

### DIFF
--- a/libra-node/src/lib.rs
+++ b/libra-node/src/lib.rs
@@ -248,10 +248,10 @@ pub fn setup_environment(node_config: &NodeConfig, logger: Option<Arc<Logger>>) 
         Arc::clone(&libra_db),
     );
 
-    let waypoint = node_config.base.waypoint.waypoint();
+    let genesis_waypoint = node_config.base.waypoint.genesis_waypoint();
     // if there's genesis txn and waypoint, commit it if the result matches.
     if let Some(genesis) = get_genesis_txn(&node_config) {
-        maybe_bootstrap::<LibraVM>(&db_rw, genesis, waypoint)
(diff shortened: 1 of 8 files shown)


libra-node/src/lib.rs
```diff
@@ -384,7 +384,7 @@ pub fn setup_environment(node_config: &NodeConfig, logger: Option<Arc<Logger>>)
         Arc::clone(&db_rw.reader),
         chunk_executor,
         &node_config,
-        waypoint,
+        genesis_waypoint,
```
davidiw, line 387: I guess longer term ... if we want... this could be the real waypoint, but whatever

davidiw (commented):
Couple other places:

https://github.com/libra/libra/blob/master/docker/compose/public_full_node/public_full_node.yaml
https://github.com/libra/libra/blob/master/config/src/config/test_data/
https://github.com/libra/libra/blob/master/testsuite/cluster-test/src/cluster_builder.rs#L410

cc: @mgorven could you help on the partners side?

davidiw:
btw thanks for tackling this

libra-node/src/lib.rs
zekun000: yeah I was thinking if I want to use waypoint or just this, but since every time insert-waypoint will change both I don't think it matters.

LSR shouldn't be able to change the waypoint to a future value unless something weird happens (db wiped)

davidiw (approved):

zekun000:
> So the validator and fullnode will only read genesis-waypoint, and LSR will only read and write waypoint?

yes

PR #6506, opened 2020-10-13
[writeset-generator] Implement a primitive tooling for generating genesis writeset 

runtian-zhou:
## Motivation

This implements an offchain binary tool that would be used by node operators to generate genesis writeset in case of some catestropic scenarios. In case of a majority of validator lossed, this tools, together with the db bootstrapper will be used to remove bad validators to kick validators out from the network.

The flow should be following: when we lost 1/3 of the node in the network, each node operator would:
1. Pause their own network
2. Sync their node to the latest committed state.
3. Use this tool to generate waypoint transaction. e.g: 
run `cargo run --bin libra-writeset-generator -- --output <path-to-genesis-transaction> remove-validators <addresses to be removed>`
4. Spawn up the db bootstrapper with the genesis transaction generated in step 4 provided.

Yes

## Test Plan

Added e2e test for the generated transaction.


language/libra-tools/writeset-transaction-generator/src/lib.rs
```diff
@@ -0,0 +1,69 @@
...
+use tempfile::NamedTempFile;
+
+fn compile_admin_script(input: &str) -> Result<Script> {
+    let mut temp_file = NamedTempFile::new()?;
+    temp_file.write_all(input.as_bytes())?;
+    let cur_path = temp_file.path().to_str().unwrap().to_owned();
+    Ok(Script::new(compile_script(cur_path), vec![], vec![]))
+}
+
+pub fn encode_remove_validators_transaction(validators: Vec<AccountAddress>) -> Transaction {
```
tzakian, line 21: What would you think of storing these scripts in a separate directory e.g., `admin_scripts` and then use something like [mustache](https://github.com/nickel-org/rust-mustache) to templatize these scripts in a similar manner to what we do in the functional tests? Mustache can encode loops, so it could recreate this logic pretty succinctly I believe. 

The command line could then pass the path to the script along with a comma-separated list of arguments that would be picked up substituted in to the script body?

language/libra-tools/writeset-transaction-generator/src/lib.rs
```diff
@@ -0,0 +1,69 @@
...
+            {}
+        }}
+    }}",
+            script_body
+        )
+        .as_str(),
+    )
+    .expect("Cannot compile script");
+
+    Transaction::GenesisTransaction(WriteSetPayload::Script {
```
zekun000, line 49: there's no authentication for GenesisTransaction, is there somewhere we stop processing it if it's sent through consensus instead of applying manually?

zekun000: oh I think consensus only accepts `SignedTransaction`, halt_network seems more useful as signed writeset? or in general do we want the tooling to be used by AOS too?

runtian-zhou: That is exactly right! The removing validator functionablity would be only doable in the GenesisTransaction setting if we lose enough validators to form a quorem. The `halt_network`, as you suggested, can be sent from AOS as well and should be how it would be used in general.

language/libra-tools/writeset-transaction-generator/src/lib.rs
```diff
@@ -0,0 +1,69 @@
...
+    let script = "
+    script {
+        use 0x1::LibraTransactionPublishingOption;
+        fun main(libra_root: &signer) {
+            LibraTransactionPublishingOption::set_open_script(libra_root);
+            LibraTransactionPublishingOption::add_to_script_allow_list(libra_root, x\"0000000000000000000000000000000000000000000000000000000000000000\");
+        }
+    }";
+
+    Transaction::GenesisTransaction(WriteSetPayload::Script {
```
zekun000, line 65: those expected to be signed by AOS should be RawTransaction instead of Transaction::GenesisTransaction? otherwise AOS has to unpack it first

also maybe we can add a libra version script too?

runtian-zhou: There's a slight gotcha here. The admin script that needs to be signed by AOS will be slight different from the script here: The script here will contain one `signer` instead of two in regular AOS signed admin script. Thus the code here are intended to be only used by the GenesisTransaction use case.

sblackshear:
Do you think it would be possible to generate a single Move script for the transaction and execute it via an AdminScript instead of a WriteSet transaction?

I was thinking about/looking at this the other day and it seems like the main reason we don't do that today is that the number of validators and their initial auth keys is variable. Today, we read them from a config in Rust and then call Move's `add_validator` in a loop. But I believe we could avoid doing that if we had a tool to generate a Move script with the appropriate number of calls to `add_validator`.

runtian-zhou:
> Do you think it would be possible to generate a single Move script for the transaction and execute it via an AdminScript instead of a WriteSet transaction?
> 
> I was thinking about/looking at this the other day and it seems like the main reason we don't do that today is that the number of validators and their initial auth keys is variable. Today, we read them from a config in Rust and then call Move's `add_validator` in a loop. But I believe we could avoid doing that if we had a tool to generate a Move script with the appropriate number of calls to `add_validator`.

We are still executing it as admin script in this case. The main problem this is trying to solve is how we can bring the system back when no consensus could be made (due to a loss of 1/3 node for example). In such scenario, no transactions, including WriteSet/AdminScript will not be able to be committed. The only way to get out of this scenario is we have each ofthe remaining node to apply this transaction (either in the form of WriteSet or AdminScript) to their state independently, and restart the system.

language/libra-tools/writeset-transaction-generator/templates/halt_transactions.move
```diff
@@ -0,0 +1,8 @@
+script {
+    use 0x1::LibraTransactionPublishingOption;
+    fun main(libra_root: &signer) {
+        LibraTransactionPublishingOption::set_open_script(libra_root);
+        // Add a null hash to the script allow list to disable execution of all scripts sent to the validator.
+        LibraTransactionPublishingOption::add_to_script_allow_list(libra_root, x"0000000000000000000000000000000000000000000000000000000000000000");
```
tzakian, line 6: TIL! I had a look around in the Move code, but can't seem to find where this check is being performed, is it happening at the VM level and not at the Move level?

tzakian (commented):
This looks good to me and I'd be fine with it going in as long as the gatekeepers are happy with it :)

language/libra-tools/writeset-transaction-generator/templates/halt_transactions.move
runtian-zhou: It is invoked in the prologue by `is_script_allowed` function!

tzakian (approved):
LGTM!

language/libra-tools/writeset-transaction-generator/src/main.rs
```diff
@@ -0,0 +1,44 @@
...
+}
+
+#[derive(Debug, StructOpt)]
+enum Command {
+    /// List of addresses to remove from validator set
+    #[structopt(name = "remove-validators")]
+    RemoveValidators { addresses: Vec<AccountAddress> },
+    /// Block the execution of any transaction in the network
+    #[structopt(name = "halt-network")]
+    HaltNetwork,
```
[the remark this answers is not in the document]
runtian-zhou: Just implemented this line!

dariorussi (approved):
good to go!

PR #6509, opened 2020-10-13
[language] Add trace! to failing transactions during execution

dariorussi:
Add trace! for transactions that get discarded during execution.
When validating transactions, mempool already reports failures so
we don't need to do it in the VM or Adapter.
However execution was not reporting discarded transaction unless
coming from invariant violation. And discarded transactions are
not charged and a system cost.
Given that trace is not logged to the server and can be enabled
"on the fly" this is a good way to debug possible problems
without penalizing the system.

### DIFF
--- a/language/libra-vm/src/libra_transaction_executor.rs
+++ b/language/libra-vm/src/libra_transaction_executor.rs
@@ -644,17 +644,25 @@ impl LibraVM {
                 debug!(log_context, "Retry after reconfiguration");
                 continue;
             };
-            let (vm_status, output) = match txn {
+            let (vm_status, output, sender) = match txn {
                 Ok(PreprocessedTransaction::BlockPrologue(block_metadata)) => {
                     execute_block_trace_guard.clear();
                     current_block_id = block_metadata.id();
                     trace_code_block!("libra_vm::execute_block_impl", {"block", current_block_id}, execute_block_trace_guard);
-                    self.process_block_prologue(data_cache, block_metadata, &log_context)?
+                    let (vm_status, output) =
+                        self.process_block_prologue(data_cache, block_metadata, &log_context)?;
+                    (vm_status, output, Some("block_prologue".to_string()))
                 }
                 Ok(PreprocessedTransaction::WaypointWriteSet(write_set_payload)) => {
-                    self.process_waypoint_change_set(data_cache, write_set_payload, &log_context)?
+                    let (vm_status, output) = self.process_waypoint_change_set(
+                        data_cache,
+                        write_set_payload,
+                        &log_context,
+                    )?;
+                    (vm_status, output, Some("waypoint_write_set".to_string()))
                 }
                 Ok(PreprocessedTransaction::UserTransaction(txn)) => {
+                    let sender = txn.sender().to_string();
                     let _timer = TXN_TOTAL_SECONDS.start_timer();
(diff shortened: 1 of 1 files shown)


dariorussi:
tested on e2e tests with logging enabled and we get all the proper tracing with good info, including `INVALID_SIGNATURE` which seems a good thing to report in execution

bmwill (approved):

sausagee (approved):

runtian-zhou:
Are we afraid that the "false negative" is going to bother us in the future? I would imagine that those lines will be triggered when the mempool didn't reject such transaction (e.g: when the second transaction used to have enough balance to pay for gas but not until the first transaction gets committed). Is the intention to record transactions as such? Or we just want to know what are transactions that get discarded.

language/libra-vm/src/libra_transaction_executor.rs
```diff
@@ -668,15 +676,30 @@ impl LibraVM {
...
                 }
                 Ok(PreprocessedTransaction::WriteSet(txn)) => {
-                    self.process_writeset_transaction(data_cache, *txn, &log_context)?
+                    let (vm_status, output) =
+                        self.process_writeset_transaction(data_cache, *txn, &log_context)?;
+                    (vm_status, output, Some("write_set".to_string()))
+                }
+                Err(e) => {
+                    let (vm_status, output) = discard_error_vm_status(e);
+                    (vm_status, output, None)
```
runtian-zhou, line 688: This should be a pure signature error. Should it be `None`?

bob-wilson (approved):
This looks good to me

dariorussi:
> Are we afraid that the "false negative" is going to bother us in the future? I would imagine that those lines will be triggered when the mempool didn't reject such transaction (e.g: when the second transaction used to have enough balance to pay for gas but not until the first transaction gets committed). Is the intention to record transactions as such? Or we just want to know what are transactions that get discarded.

I think we are somewhat concerned in an open system that false positive can become a problem. So in some future we may change that. But right now, and to be honest also in the future, one could argue that either those errors do not happen often (as they should not), in which case the noise is not that high anyway, and if they happen a lot being able to know about them may be a good thing. Thoughts?
In any case this is effectively logging nothing and it will only log if the level is kicked down on a given validator. And that is also the intent here, to have a debug story that is completely off in normal operation.
Let me know if you have concerns but this seems a fine story for now, particularly if we are all aware of it

language/libra-vm/src/libra_transaction_executor.rs
dariorussi: yes correct, that is what it looks like, just figured I'd bundle any error under None and print out a "malformed transaction" which for now we know to be only an invalid signature (and maybe forever?).
I somewhat did not want to be too explicit in the face of the `Err(e)`, I can change that if we think is important

bob-wilson:
> In any case this is effectively logging nothing and it will only log if the level is kicked down on a given validator. And that is also the intent here, to have a debug story that is completely off in normal operation.
> Let me know if you have concerns but this seems a fine story for now, particularly if we are all aware of it

That matches my understanding of the situation. The last time we discussed logging these failed transactions, we decided not to do it for exactly the reason that @runtian-zhou mentioned. But, with the trace level of logging, we can selectively enable this for debugging and get that extra level of visibility only when we want it.

PR #6513, opened 2020-10-14
[mempool] config tweaks

christinacelee:
## Motivation

Making garbage-collection interval and system TTL default config values lower - also open to other suggestions

### DIFF
--- a/config/src/config/mempool_config.rs
+++ b/config/src/config/mempool_config.rs
@@ -34,8 +34,8 @@ impl Default for MempoolConfig {
             capacity: 1_000_000,
             capacity_per_user: 100,
             default_failovers: 0,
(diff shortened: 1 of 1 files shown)


config/src/config/mempool_config.rs
```diff
@@ -34,8 +34,8 @@ impl Default for MempoolConfig {
             capacity: 1_000_000,
             capacity_per_user: 100,
             default_failovers: 0,
-            system_transaction_timeout_secs: 86400,
-            system_transaction_gc_interval_ms: 180_000,
+            system_transaction_timeout_secs: 900,
```
davidiw, line 37: I heard something along the line of 5 minutes, but I trust you know better than I do.

davidiw (approved):

PR #6520, opened 2020-10-14
Add NamedChain for pre-mainnet

mgorven:
We should be able to start using a stable chain ID for pre-mainnet, add
a named chain.


davidiw (approved):
exciting times!

PR #6524, opened 2020-10-14
Cargo.toml fix (versino -> version)

mimoo:


msmouse (approved):

davidiw (approved):
is there a way to enable strict checking of Cargo.toml to prevent this in the future?

mimoo:
@davidiw I actually don't understand why we specify a version when we import via path, I'm almost 100% sure that it does nothing.

PR #6525, opened 2020-10-14
[security] adding checked arithmetic to the coding guidelines

mimoo:
rationale: it is undeniable that every time we perform an integer operation, there are implied edge-cases that we MUST think about, and I think we can also agree that not everybody thinks about integer overflows when they write code. Rust gives us a great tool, **checked operations**, which forces us to explicitly document how we want to handle these edge cases.

### DIFF
--- a/documentation/coding_guidelines.md
+++ b/documentation/coding_guidelines.md
@@ -227,6 +227,18 @@ impl Foo {
 }
 ```
 
+### Integer Arithmetic
+
+As every integer operation (`+`, `-`, `/`, `*`, etc.) implies edge-cases (e.g. overflows `u64::MAX + 1`, underflows `0u64 -1`, division by zero, etc.),
+we use checked arithmetic instead of directly using math symbols.
+It forces us to think of edge-cases, and handle them explicitely.
+This is a brief and simplified mini guide of the different functions that exist to handle integer arithmetic:
+
+* [checked_](https://doc.rust-lang.org/std/primitive.u32.html#method.checked_add): use this function if you want to handle overflows and underflows as a special edge-case. It returns `None` if an underflow or overflow has happened, and `Some(operation_result)` otherwise.
+* [overflowing_](https://doc.rust-lang.org/std/primitive.u32.html#method.overflowing_add): use this function if you want the result of an overflow to potentially wrap around (e.g. `u64::MAX.overflow_add(10) == (9, true)`). It returns the underflowed or overflowed result as well as a flag indicating if an overflow has occured or not.
+* [wrapping_](https://doc.rust-lang.org/std/primitive.u32.html#method.wrapping_add): this is similar to overflowing operations, except that it returns the result directly. Use this function if you are sure that you want to handle underflows and overflows by wrapping around.
+* [saturating_](https://doc.rust-lang.org/std/primitive.u32.html#method.saturating_add): if an overflow occurs, the result is kept within the boundary of the type (e.g. `u64::MAX.saturating_add(1) == u64::MAX`).
+
 ### Logging
 
 We currently use [log](https://docs.rs/log/) for logging.


davidiw:
Adding more people to give their opinions.

documentation/coding_guidelines.md
```diff
@@ -227,6 +227,16 @@ impl Foo {
 }
 ```
 
+### Integer Arithmetic
+
+As every integer operation (`+`, `-`, `/`, `*`, etc.) implies edge-cases (e.g. `u64::MAX + 1`),
```
davidiw, line 232: divide can result in edge cases in integers :O?

documentation/coding_guidelines.md
```diff
@@ -227,6 +227,16 @@ impl Foo {
...
+we use checked arithmetic instead of directly using the math symbols (`+`, `-`, etc.):
```
davidiw, line 233: why not multiply?

documentation/coding_guidelines.md
```diff
@@ -227,6 +227,16 @@ impl Foo {
...
+
+* [checked_](https://doc.rust-lang.org/std/primitive.u32.html#method.checked_add): returns `None` if an overflow has happened, `Some(operation_result)` otherwise
+* [overflowing_](https://doc.rust-lang.org/std/primitive.u32.html#method.overflowing_add): returns the result, potentially wrapping around (e.g. `u64::MAX.overflow_add(1) == 0`), as well as a flag indicating if an overflow has occured or not
+* [wrapping_](https://doc.rust-lang.org/std/primitive.u32.html#method.wrapping_add): similar to overflowing operations, except that it returns the result directly
+* [saturating_](https://doc.rust-lang.org/std/primitive.u32.html#method.saturating_add): if an overflow occurs, the result is kept within the boundary of the type (e.g. `u64::MAX.saturating_add(1) == u64::MAX`)
```
davidiw, line 238: this isn't really indicating what to chose or how to chose it, it seems to be a copy of a doc block, perhaps we could pivot more toward recommendations

* choose X if you're not sure what to choose
* if roll over is okay do this
* if roll over would result in safety concerns do this

davidiw (changes requested):

documentation/coding_guidelines.md
mimoo: dividing by 0

documentation/coding_guidelines.md
mimoo: to avoid repeating the list :P

documentation/coding_guidelines.md
mimoo: > The sane default is to turn on overflow-checking 

we had this discussion on workplace, and @sblackshear had a pretty good point there. I thought about it more and, I think it is somewhat similar to debug_asserts. For these we deliberately decided not to crash in prod, so why enable them in prod? The same could be thought of integer overflows, but worse: I don't think anyone who wrote `+` or `-` thought about the edge-cases, and decided that they should crash or not crash in prod. Worse, if most integer overflows end up being benign, then we created a lot of DoS attack vectors in prod.

> This generally means "normal" Rust syntax will get you the checked version

I think you mean that it'll get you the "checked + unwrap version" which from my look at the lints, we can avoid in most cases.

metajack:
> we want to enforce the lint!

Ok, I didn't have the context of the other PR. Sorry.

zekun000:
do we consider the perspective of developer experience/velocity? the syntax is much less friendly than +, -

is there any other open source projects adopt such a lint rule?

mimoo:
> do we consider the perspective of developer experience/velocity? the syntax is much less friendly than +, -

I had a private convo on this subject, my point of view is the same as the reason we adopted rust over go or python. Even though there is a learning curve, we're writing high-assurance software and we need to accept this friction if we want to write secure code. (Maybe it looks a bit uglier, but if it's worth it in terms of security then it's a no-brainer.)

davidiw (commented):
I'd still like to see a generally acceptable approach here... maybe adding something to infallible that wraps checked but does so in a way that logs something meaningful but still gets us out.

also wrapping is what happens by default without any of this magic, probably good to call that out. at the same time you should probably say what the current code base default is if nothing is done.

I think this is still missing somewhat of a guide for this rather nuanced space.

mimoo:
> generally acceptable approach

You're looking for something that doesn't exist unfortunately : o
I understand that it's not great to have to think about how you want to handle edge cases EVERY TIME you use integer arithmetic, but again this is not new and this is already something you had to do before.

> also wrapping is what happens by default without any of this magic

this might be surprising as I think rust is the only language that actually specifies the behavior. I think the idea is still that a wrapping around is a programmer error, unless it was explicitly defined to do so (in other words, if you didn't use the `wrapping_add` function then it means you did not think of edge-cases)

documentation/coding_guidelines.md
```diff
@@ -227,6 +227,18 @@ impl Foo {
...
 
+### Integer Arithmetic
+
+As every integer operation (`+`, `-`, `/`, `*`, etc.) implies edge-cases (e.g. overflows `u64::MAX + 1`, underflows `0u64 -1`, division by zero, etc.),
+we use checked arithmetic instead of directly using math symbols.
+It forces us to think of edge-cases, and handle them explicitely.
+This is a brief and simplified mini guide of the different functions that exist to handle integer arithmetic:
+
+* [checked_](https://doc.rust-lang.org/std/primitive.u32.html#method.checked_add): use this function if you want to handle overflows and underflows as a special edge-case. It returns `None` if an underflow or overflow has happened, and `Some(operation_result)` otherwise.
+* [overflowing_](https://doc.rust-lang.org/std/primitive.u32.html#method.overflowing_add): use this function if you want the result of an overflow to potentially wrap around (e.g. `u64::MAX.overflow_add(1) == 0`). It returns the underflowed or overflowed result as well as a flag indicating if an overflow has occured or not.
```
msmouse, line 238: I think it'll be more clear to give the example `u64::MAX.overflow_add(10 == (0, true)`, so for someone lazy as me who skims through, when reading "except that it returns the result DIRECTLY" doesn't get confused.

mimoo: oops, good point!

davidiw (approved):
congrats on this initiative, I've already started seeing folks apply these guidelines thanks to your linters... even folks that love simplicity like @zekun000, now that's what I call impact!

bmwill:
/land

We need a way to skip cluster test for doc only changes :/

metajack:
> We need a way to skip cluster test for doc only changes :/

Can we do this with target determinator?

sunshowers:
Yeah should be possible. (also e2e test). Will work on this today.

PR #6512, opened 2020-10-14
[mempool] specs/README

christinacelee:
## Motivation

(Write your motivation for proposed changes here.)

### Have you read the [Contributing Guidelines on pull requests](https://github.com/libra/libra/blob/master/CONTRIBUTING.md#pull-requests)?

(Write your answer here.)

## Test Plan

(Share your test plan here. If you changed code, please provide us with clear instructions for verifying that your changes work.)

## Related PRs

(If this PR adds or changes functionality, please take some time to update the docs at https://github.com/libra/libra/developers.libra.org, and link to your PR here.)

PR #6531, opened 2020-10-15
[specifications][mempool] Add initial mempool specs

aching:
This was written by @christinacelee.  It would be great to get it in before the release branch cut today.  This is only documentation with no effect on production code.


mimoo (approved):
let's get this in!

aching:
Oh whoops, https://github.com/libra/libra/pull/6512 already exists.

PR #6532, opened 2020-10-15
[Managament Tooling] Unify the insert-waypoint command logic and add a smoke test.

JoshLind:
## Motivation

This PR unifies the logic for the "insert-waypoint" command offered by both the operational tool and the genesis tool. Currently, the logic for these commands is copy-pasted (and thus has incorrectly diverged). Instead, the logic for these commands should be unified. This PR does this in 2 commits:
1) First, we unify the logic for these commands by having both tools to point to a shared struct.
2) Second, we add a smoke test for the "insert-waypoint" command of the operational tool. The genesis tool is already tested in existing smoke tests.

Yes.

## Test Plan

All tests seem to pass locally.

## Related PRs

None.


davidiw (approved):
/land

PR #6533, opened 2020-10-15
[TCB] building the TCB with overflow checks

mimoo:
This proposes to build the TCB with [overflow checks](https://rust-lang.github.io/rfcs/1535-stable-overflow-checks.html).

How can I test this PR?

### DIFF
--- a/docker/build-common.sh
+++ b/docker/build-common.sh
@@ -14,16 +14,22 @@ export CARGO=$(rustup which --toolchain $RUST_NIGHTLY cargo)
 export CARGOFLAGS=$(cat cargo-flags)
 export CARGO_PROFILE_RELEASE_LTO=thin # override lto setting to turn on thin-LTO for release builds
 
-# Build release binaries
+# Build release binaries except TCB
 ${CARGO} ${CARGOFLAGS} build --release \
          -p libra-genesis-tool \
          -p libra-operational-tool \
          -p libra-node \
-         -p libra-key-manager \
-         -p safety-rules \
          -p db-bootstrapper \
          -p backup-cli \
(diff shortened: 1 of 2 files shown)


docker/build-common.sh
```diff
@@ -14,17 +14,23 @@ export CARGO=$(rustup which --toolchain $RUST_NIGHTLY cargo)
...
          -p libra-operational-tool \
          -p libra-node \
-         -p libra-key-manager \
-         -p safety-rules \
          -p db-bootstrapper \
          -p backup-cli \
          "$@"
 
+# Build release TCB binaries
+git apply docker/tcb_build.patch
```
davidiw, line 27: what is tcb-build.patch?

mimoo: sorry, refresh!

mimoo: you're too fast!

mimoo:
@davidiw looks like nothing is happening in longevity testing so far in terms of integer overflow, perhaps a good time to revisit this PR?

davidiw:
Fix the linters, I think you need to either eliminate the last line in the patch. Let's verify all is good with a cluster test canary too. PR looks pretty interesting but good enough. Though might be good to get a dev-infra person to weigh in.

mimoo:
It's ugly but without custom profiles (cf https://github.com/rust-lang/cargo/issues/6988) I don't see a better way to do it : o would be interested to know what the rust experts think : )

davidiw (approved):
god save us all

davidiw:
let's not leave approved, stale PRs unclosed.

PR #6537, opened 2020-10-16
[state sync] verify waypoint's end-of-epoch li matches end-of-chunk v…

christinacelee:
…ersion before execution

## Motivation

In the waypoint sync protocol, when processing the waypoint chunk response, we don't check for whether the end_of_epoch_li actually matches the end of the chunk of txns. This throws an error in execution, where this version check doesn't happen. This PR adds that version check

Note: IDEALLY also on when we process a waypoint chunk request, we also don't return the end_of_epoch_li if the chunk of txns don't match that end_of_epoch_li, but that requires a larger refactor of the code

## Test Plan

Updated integration test, added new smoke test, both fail without fix impl

### DIFF
--- a/state-synchronizer/src/coordinator.rs
+++ b/state-synchronizer/src/coordinator.rs
@@ -731,6 +731,13 @@ impl<T: ExecutorProxyTrait> SyncCoordinator<T> {
             None
         };
         if let Some(li) = end_of_epoch_li.as_ref() {
+            ensure!(
+                li.ledger_info().version() >= request.known_version,
+                "waypoint request's current_epoch (epoch {}, version {}) < waypoint request's known_version {}",
+                li.ledger_info().epoch(),
+                li.ledger_info().version(),
+                request.known_version,
+            );
             let num_txns_until_end_of_epoch = li.ledger_info().version() - request.known_version;
             limit = std::cmp::min(limit, num_txns_until_end_of_epoch);
         }
@@ -826,6 +833,7 @@ impl<T: ExecutorProxyTrait> SyncCoordinator<T> {
         debug!(
             LogSchema::event_log(LogEntry::ProcessChunkResponse, LogEvent::Received)
                 .chunk_resp(&response)
+                .peer(peer)
         );
         fail_point!("state_sync::apply_chunk", |_| {
             Err(anyhow::anyhow!("Injected error in apply_chunk"))
@@ -1043,12 +1051,12 @@ impl<T: ExecutorProxyTrait> SyncCoordinator<T> {
         // Optimistically fetch the next chunk.
         let new_version =
             self.local_state.highest_version_in_local_storage() + txn_list_with_proof.len() as u64;
-        // The epoch in the optimistic request should be the next epoch if the current chunk
+        // The epoch in the optimistic request (= new_epoch) should be the next epoch if the current chunk
         // is the last one in its epoch.
         let new_epoch = end_of_epoch_li
             .as_ref()
             .map_or(self.local_state.epoch(), |li| {
-                if li.ledger_info().version() == new_version {
+                if li.ledger_info().version() == new_version && li.ledger_info().ends_epoch() {
                     self.local_state.epoch() + 1
                 } else {
                     self.local_state.epoch()
@@ -1064,6 +1072,22 @@ impl<T: ExecutorProxyTrait> SyncCoordinator<T> {
             }
         }
 
+        // verify the end-of-epoch LI for the following before passing it to execution:
+        // * verify end-of-epoch-li against local state
+        // * verify end-of-epoch-li's version corresponds to end-of-chunk version before
+        // passing it to execution
+        // * Executor expects that when it is passed an end-of-epoch LI, it is going to execute/commit
+        // transactions leading up to that end-of-epoch LI
+        // * verify end-of-epoch-li actually ends an epoch
+        let end_of_epoch_li = end_of_epoch_li
+            .map(|li| {
+                // verify end-of-epoch-li against local state
+                self.local_state.trusted_epoch.verify(&li).map(|_| li)
+            })
+            .transpose()?
+            .filter(|li| {
+                li.ledger_info().version() == new_version && li.ledger_info().ends_epoch()
+            });
         self.waypoint.verify(waypoint_li.ledger_info())?;
         self.validate_and_store_chunk(txn_list_with_proof, waypoint_li, end_of_epoch_li)
     }
--- a/state-synchronizer/src/tests/integration_tests.rs
+++ b/state-synchronizer/src/tests/integration_tests.rs
@@ -690,14 +690,21 @@ fn catch_up_with_waypoints() {
         false,
         None,
     );
-    for epoch in 1..10 {
-        env.commit(0, epoch * 100);
+    let mut curr_version = 0;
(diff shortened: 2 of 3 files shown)


zekun000 (commented):
btw I think we never verify the end_of_epoch_li

zekun000:
I'm also not very clear what'd happen if a malicious server tries to do bad things, e.g. skip the end_of_epoch_li, send multiple epochs in one chunk etc.

state-synchronizer/src/coordinator.rs
```diff
@@ -731,6 +731,12 @@ impl<T: ExecutorProxyTrait> SyncCoordinator<T> {
             None
         };
         if let Some(li) = end_of_epoch_li.as_ref() {
+            ensure!(
+                li.ledger_info().version() >= request.known_version,
+                "end-of-epoch LI version {} < request waypoint version {}",
+                li.ledger_info().version(),
+                request.known_version,
```
zekun000, line 738: should be able to use local_state.trusted_epoch to verify it?

christinacelee:
> I'm also not very clear what'd happen if a malicious server tries to do bad things, e.g. skip the end_of_epoch_li, send multiple epochs in one chunk etc.

I would expect we just fail in processing the chunk response and the sync requester tries to find a better peer to request sync from?

state-synchronizer/src/coordinator.rs
christinacelee: this is on the end of sending the chunk response, not processing it. This checks that request's current_epoch and request's known_version match up and doesn't cause overflow below.

zekun000:
I don't see executor enforcing the rules to fail processing chunk (not cross multiple epoch, must have epoch_end_li)

christinacelee:
that's surprising to me (that executor doesn't handle that)..  seems like out-of-scope for this PR / is that really expectation of state sync?

zekun000 (approved):

state-synchronizer/src/coordinator.rs
```diff
@@ -731,6 +731,12 @@ impl<T: ExecutorProxyTrait> SyncCoordinator<T> {
             None
         };
         if let Some(li) = end_of_epoch_li.as_ref() {
+            ensure!(
+                li.ledger_info().version() >= request.known_version,
+                "end-of-epoch LI version {} < request waypoint version {}",
```
davidiw, line 736: should this be local end-of-epoch li version?

davidiw: so this is really just a sanity check on the waypoint?

state-synchronizer/src/coordinator.rs
```diff
@@ -1064,6 +1071,20 @@ impl<T: ExecutorProxyTrait> SyncCoordinator<T> {
...
+            self.local_state.trusted_epoch.verify(&li)?;
+            if li.ledger_info().version() == new_version {
+                Some(li)
+            } else {
+                None
+            }
+        } else {
+            None
+        };
+
```
davidiw, line 1087: can we ad a comment such that we do not pass in an end of epoch li unless we have all the transactions for that epoch?

davidiw: I think there is some quirkiness on the optimistic send starting on line 1048/1055, I believe that should be moved /after/ this check....

state-synchronizer/src/coordinator.rs
```diff
@@ -1064,6 +1071,20 @@ impl<T: ExecutorProxyTrait> SyncCoordinator<T> {
             }
         }
 
+        // verify end-of-epoch-li against local state
+        // verify end-of-epoch-li's version corresponds to end-of-chunk version before
+        // passing it to execution
+        let end_of_epoch_li = if let Some(li) = end_of_epoch_li {
+            self.local_state.trusted_epoch.verify(&li)?;
```
davidiw, line 1078: move this out of the this blob and put it at the end and do the filter earlier

state-synchronizer/src/coordinator.rs
```diff
@@ -1043,15 +1051,24 @@ impl<T: ExecutorProxyTrait> SyncCoordinator<T> {
...
+        // * Verify end-of-epoch-li's version corresponds to end-of-chunk version before
+        // passing it to execution
+        // * Executor expects that when it is passed an end-of-epoch LI, it is going to execute/commit
+        // transactions leading up to that end-of-epoch LI
+        let (new_epoch, end_of_epoch_li) = end_of_epoch_li
+            .map(|li| {
+                // verify end-of-epoch-li against local state
+                self.local_state.trusted_epoch.verify(&li).map(|_| li)
+            })
+            .transpose()?
```
davidiw, line 1066: I thought the plan was to take the li and do the optimistic follow up here

down below where we have

`self.waypoint.verify(`

we could do

end_of_epoch_li.map(|li| self...).transpose()?;

state-synchronizer/src/coordinator.rs
davidiw: I'm a little concerned about adding extra checks at this point in time. Is it that bad if we just ... get rid of this? or make it a warning instead of a Result?

christinacelee: if this check isn't included, we will encounter subtraction overflow right in the line below (line 741)

state-synchronizer/src/coordinator.rs
```diff
@@ -1043,12 +1051,17 @@ impl<T: ExecutorProxyTrait> SyncCoordinator<T> {
...
-        // The epoch in the optimistic request should be the next epoch if the current chunk
+        // The epoch in the optimistic request (= new_epoch) should be the next epoch if the current chunk
         // is the last one in its epoch.
         let new_epoch = end_of_epoch_li
             .as_ref()
+            .map(|li| {
+                // verify end-of-epoch-li against local state
+                self.local_state.trusted_epoch.verify(&li).map(|_| li)
+            })
+            .transpose()?
```
davidiw: delete, you already do this portion below.

state-synchronizer/src/coordinator.rs
```diff
@@ -1043,12 +1051,17 @@ impl<T: ExecutorProxyTrait> SyncCoordinator<T> {
...
             .map_or(self.local_state.epoch(), |li| {
-                if li.ledger_info().version() == new_version {
+                if li.ledger_info().version() == new_version && li.ledger_info().ends_epoch() {
```
davidiw, line 1059: good call

zekun000: maybe we should verify the end_of_epoch_li early and add the ends_epoch check too, if it fails, it's verification error and should be logged.

davidiw: there's a balance between the optimistic request... I kinda don't wanna rock the boat more than we already are

davidiw (approved):
looks good, let's get a cherry pick commit up too :)

state-synchronizer/src/coordinator.rs
christinacelee: > maybe we should verify the end_of_epoch_li early and add the ends_epoch check too, if it fails, it's verification error and should be logged.

lol i had this before too, but asked to change for better readability. also i think in the current code, we also verify the waypoint_li after the network send too - i think the idea was to parallelize network send and sig verification, etc

PR #6539, opened 2020-10-16
[JSON-RPC] Adding a get_events_with_proof API

AnomalRoil:
## Motivation
This is necessary to build a verifying clients and thus to ensure a secure ecosystem.
Otherwise we are encouraging people to trust the full nodes.

This is adding a JSON-RPC call to get events with proofs, along with the necessary backend changes to get these.

## Test Plan

I'm working on adding unit and integration tests right now.

## Related PRs
#6399

### DIFF
--- a/json-rpc/API-CHANGELOG.md
+++ b/json-rpc/API-CHANGELOG.md
@@ -13,6 +13,11 @@ Please add the API change in the following format:
 
 ```
 
+## 2020-12-22 Add `get_events_with_proof` APIs
+
+This allows to verify events since transactions and their proofs are not covering all events.
+
+
 # 2020-12-07 Add a `X-Diem-Chain-Id`, `X-Diem-Ledger-Version` and `X-Diem-Ledger-TimestampUsec` to http response header
 
 The added headers are same with JSON-RPC response json same name fields. They are added for client to know chain id and server's latest block version and timestamp without decoding body json.
--- a/json-rpc/src/tests/unit_tests.rs
+++ b/json-rpc/src/tests/unit_tests.rs
@@ -613,6 +613,86 @@ fn test_json_rpc_protocol_invalid_requests() {
             }),
         ),
         (
+            "get_events_with_proofs: invalid event_key type",
+            json!({"jsonrpc": "2.0", "method": "get_events_with_proofs", "params": [false, 1, 10], "id": 1}),
+            json!({
+                "error": {
+                    "code": -32602,
+                    "message": "Invalid param event key(params[0]): should be hex-encoded string",
+                    "data": null
+                },
+                "id": 1,
+                "jsonrpc": "2.0",
+                "diem_chain_id": ChainId::test().id(),
+                "diem_ledger_timestampusec": timestamp,
+                "diem_ledger_version": version
+            }),
+        ),
+        (
+            "get_events_with_proofs: event_key is not hex-encoded string",
+            json!({"jsonrpc": "2.0", "method": "get_events_with_proofs", "params": ["helloworld", 1, 10], "id": 1}),
+            json!({
+                "error": {
+                    "code": -32602,
+                    "message": "Invalid param event key(params[0]): should be hex-encoded string",
+                    "data": null
+                },
+                "id": 1,
+                "jsonrpc": "2.0",
+                "diem_chain_id": ChainId::test().id(),
+                "diem_ledger_timestampusec": timestamp,
+                "diem_ledger_version": version
+            }),
+        ),
+        (
+            "get_events_with_proofs: invalid start param",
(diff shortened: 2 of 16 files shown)


libra-action:

Breaking changes policy:

2. Adding fields is ok.

AnomalRoil:
@msmouse As discussed, since I've touched LibraDB, it'd be great if you could take a look at this PR.

thefallentree (approved):

storage/libradb/src/lib.rs
```diff
@@ -620,16 +620,29 @@ impl DbReader for LibraDB {
...
+
+    fn get_events_with_proofs(
+        &self,
+        event_key: &EventKey,
+        start: u64,
+        order: Order,
+        limit: u64,
+    ) -> Result<Vec<EventWithProof>> {
+        gauged_api("get_events_with_proofs", || {
             let version = self
```
msmouse, line 640: I think the version against which the proof is provided should be passed in -- so that proofs within the same client request are against a same latest version that the client knows about.

AnomalRoil: This is a good UX question which I've been trying to figure out.
Currently the client is expected to batch its call with a get_state_proof call like it's done here in the integration test: it allows the client to upgrade its state to the latest state (and thus the latest set of validators in case of an epoch change) and to verify the proofs.

You can see in the integration test how I've done it. (Although it's a bit hack-ish since we don't have a state with the validator set in the integration tests...)

Notice this is the same as I've done for the transaction_with_proofs UX, but this is maybe terrible UX and I should instead have the client specify the version for which it wants to get proofs, indeed.

If we agree that it's better to have the client specify the version for which it wants the proofs, then I should change the recent `get_transactions_with_proofs` API too:
https://github.com/libra/libra/blob/8332031c482cdbc678b16310971e8572a2cfb995/json-rpc/src/methods.rs#L397-L415 

What do you think @xli  ? On my side I think this would be cleaner if the client could specify its desired version, indeed since then it allows to have non-batched calls with the API :)

msmouse: There are two matters here: 
1. Client provided version or server latest version. Allowing the client to optionally pass in a version, defaulting to the server latest version is fine.
2. all batched queries in the same request should return a consistent view. i.e. add `ledger_version` param to the DB method you add. Check out other DB methods use it, for example `get_transactions()` -- it should not return things beyond that version, and the proofs should be relative to that version.  (I think in your case the underlying `get_events_by_event_key()` has already dealt with that.

AnomalRoil: Done, we now support both batched and non-batched queries thanks to a `known_version` optional parameter

json-rpc/src/tests/unit_tests.rs
```diff
@@ -612,6 +612,70 @@ fn test_json_rpc_protocol_invalid_requests() {
...
+                    "message": "Invalid param limit(params[2]): should be unsigned int64",
+                    "data": null
+                },
+                "id": 1,
+                "jsonrpc": "2.0",
+                "diem_chain_id": ChainId::test().id(),
+                "diem_ledger_timestampusec": timestamp,
+                "diem_ledger_version": version
+            }),
+        ),
```
xli, line 678: may add a test for validation error for the 4th param

json-rpc/docs/method_get_events_with_proofs.md
```diff
@@ -0,0 +1,25 @@
...
+
+Note:
+1. For `sentpayment` and `receivedpayment` events, call [get_account](method_get_account.md) to get the event key of the event treams for a given user account.
+2. For currency related events, call [get_currencies](method_get_currencies.md) to get the event keys (for example: to LBR xchange rate change event key).
+3. In the case where the default is used for the version parameter, then a batched call with `get_state_proof` is recommended in order to be able to validate the proofs. See the integration tests for an example.
+
+
+### Returns
+
+Returns array of hex encoded string of raw bytes of the event_with_proof type for the requested version (or the latest known version at the moment of the call if none is provided).
```
xli, line 25: could you be more specific about return data type, may add an example here

json-rpc/docs/method_get_events_with_proofs.md
```diff
@@ -0,0 +1,49 @@
...
+        })
+    ])), 
+    error: None }
+```
```
xli, line 49: may check other methods, we should use json text as example

json-rpc/API-CHANGELOG.md
```diff
@@ -13,6 +13,11 @@ Please add the API change in the following format:
 
 ```
 
+## 2020-12-09 Add `get_events_with_proof` APIs
```
xli, line 16: could you update this date to the date you land before you land this PR, as we have release 1.1 cut recently, it's better to use the date you are going to land.

Issue #6161, opened 2020-09-23
[Bug] CLI should handle JSON-RPC server stable response error


xli:
# 🐛 Bug

Client receives server error: The other accumulator is bigger than this one.

## To reproduce

* Start ./scripts/cli/start_cli_testnet.sh
*  a c
* a mb 0 11111 LBR

```
libra% a c
>> Creating/retrieving next local account from wallet
Created/retrieved local account #1 address 8799bcb7a41fc2b5f64dbcdae87b1c57
libra% a mb 0 22322 LBR
>> Sending coins from faucet
waiting for 000000000000000000000000000000dd with sequence number 6795
.......
Response with error: Failed to get state proof with error: JsonRpcError { code: -32000, message: "Server error: The other accumulator is bigger than this one. self.num_leaves: 5079983. sub_acc_leaves: 5079984.", data: None }
....
Response with error: Failed to get state proof with error: JsonRpcError { code: -32000, message: "Server error: The other accumulator is bigger than this one. self.num_leaves: 5079983. sub_acc_leaves: 5079984.", data: None }
..
transaction executed!
Finished sending coins from faucet!
libra%
```

Other commands related to calling server will receive similar error

## Expected Behavior

Client should retry to connect to another server for latest ledger info, and no error output.

## System information

latest testnet

Issue #6025, opened 2020-09-14
[Bug] Testnet CLI prints a lot of error messages when executing commands.


xli:
# 🐛 Bug

CLI connects to testnet show the following errors when executing commands, user has no idea what's going wrong and what need to do for the error:

> Failed to get state proof with error: JsonRpcError { code: -32000, message: "Server error: The other accumulator is bigger than this one. self.num_leaves: 227723. sub_acc_leaves: 227724.", data: None }

## To reproduce

connect to testnet by cli: scripts/cli/start_cli_testnet.sh
do account create, mint coins, query status.

## Expected Behavior

Error should be handled, or print with information about what user should do, or do not print if nothing need to do from user, or downgrade the log level to info, so that it only prints for someone cares about the info.

PR #6545, opened 2020-10-19
[cli] fix CLI problems with new json-rpc client

xli:
## Motivation

CLI problems: (#6161, #6025)
1. It won't handle server stale response, causes client side printing error messages when received stale response.
2. Wait transaction execution logic is not correct, should check transaction hash for making sure the transaction got by account address and seq is the one we are waiting for.
3. It supports non-blocking minting coins, when minting coins, it is easy to get into problems because of Faucet service can't handle concurrent minting requests. As TC / DD accounts need to know right sequence number to submit new transactions.
4. Waiting for minting on testnet Faucet service is not correct: waiting for dd account by checking next account sequence may give wrong information when there is another person running CLI and requested minting coins at same time (used same account sequence).

This PR fixes above problems by the following changes:
1. A new json-rpc client for addressing problem 1 and 2:
  * Handle stale response by retrying the operation.
  * Implement correct wait transaction logic.
2. Removed non-blocking mint, this will make sure user of CLI won't submit multiple minting request in sequence which is likely to be failed and confusing to user.
3. Update CLI faucet minting logic to wait for minting transaction executed, instead of only checking next dd account seq.

Note: new json-rpc client is built on top of protobuf defined response types, the plan is to define jsonrpc response types once in protobuf, then generate code for all client sdk languages. Python, golang, java clients are using the protobuf definition added in this PR already. Later we will also update server side to use same protobuf generated code. After that, when server added new fields / types, client can generate response types from same definition.

Yes

## Test Plan

Manual test.

## Related PRs

### DIFF
--- a/client/json-rpc/src/async_client/client.rs
+++ b/client/json-rpc/src/async_client/client.rs
@@ -1,0 +1,332 @@
+// Copyright (c) The Libra Core Contributors
+// SPDX-License-Identifier: Apache-2.0
+
+use crate::async_client::{
+    defaults, types as jsonrpc, Error, JsonRpcResponse, RetryStrategy, State,
+    WaitForTransactionError,
+};
+use libra_crypto::hash::CryptoHash;
+use libra_types::{
+    account_address::AccountAddress,
+    transaction::{SignedTransaction, Transaction},
+};
+use serde::de::DeserializeOwned;
+use serde_json::json;
+use std::time::Duration;
+
+#[derive(Debug)]
+pub struct Response<R> {
+    pub result: R,
+    pub state: State,
+}
+
+impl<R> std::ops::Deref for Response<R> {
+    type Target = R;
+
+    fn deref(&self) -> &Self::Target {
+        &self.result
+    }
+}
+
+#[derive(Debug)]
+pub struct Client<R> {
+    pub http_client: reqwest::Client,
+    pub server_url: reqwest::Url,
+    pub last_known_state: std::sync::RwLock<Option<State>>,
+    pub retry: R,
+}
+
+impl<R: RetryStrategy> Client<R> {
+    pub fn from_url<T: reqwest::IntoUrl>(server_url: T, retry: R) -> Result<Self, reqwest::Error> {
+        Ok(Self {
+            http_client: reqwest::ClientBuilder::new()
+                .use_native_tls()
+                .timeout(defaults::HTTP_REQUEST_TIMEOUT)
+                .build()
+                .expect("Unable to build Client."),
+            server_url: server_url.into_url()?,
+            last_known_state: std::sync::RwLock::new(None),
+            retry,
+        })
+    }
+
+    pub async fn wait_for_signed_transaction(
+        &self,
+        txn: &SignedTransaction,
+        timeout: Option<Duration>,
+        delay: Option<Duration>,
+    ) -> Result<Response<jsonrpc::Transaction>, WaitForTransactionError> {
+        self.wait_for_transaction(
+            &txn.sender(),
+            txn.sequence_number(),
+            txn.expiration_timestamp_secs(),
+            &Transaction::UserTransaction(txn.clone()).hash().to_hex(),
+            timeout,
+            delay,
+        )
+        .await
+    }
+
+    pub async fn wait_for_transaction(
+        &self,
+        address: &AccountAddress,
+        seq: u64,
+        expiration_time_secs: u64,
+        txn_hash: &str,
+        timeout: Option<Duration>,
+        delay: Option<Duration>,
+    ) -> Result<Response<jsonrpc::Transaction>, WaitForTransactionError> {
+        let start = std::time::Instant::now();
+        while start.elapsed() < timeout.unwrap_or(defaults::TIMEOUT) {
+            let txn_resp = self
+                .get_account_transaction(address, seq, true)
+                .await
+                .map_err(WaitForTransactionError::GetTransactionError)?;
+            if let Some(txn) = txn_resp.result {
+                if txn.hash != txn_hash {
+                    return Err(WaitForTransactionError::TransactionHashMismatchError(txn));
+                }
+                if let Some(ref vm_status) = txn.vm_status {
+                    if vm_status.r#type != jsonrpc::VM_STATUS_EXECUTED {
+                        return Err(WaitForTransactionError::TransactionExecutionFailed(txn));
+                    }
+                }
+                return Ok(Response {
+                    result: txn,
+                    state: txn_resp.state,
+                });
+            }
+            if let Some(state) = self.last_known_state() {
+                if expiration_time_secs <= state.timestamp_usecs / 1_000_000 {
+                    return Err(WaitForTransactionError::TransactionExpired(
+                        state.timestamp_usecs,
+                    ));
+                }
+            }
+            tokio::time::delay_for(delay.unwrap_or(defaults::WAIT_DELAY)).await;
+        }
+        Err(WaitForTransactionError::Timeout(start.elapsed()))
+    }
+
+    pub fn last_known_state(&self) -> Option<State> {
+        let data = self.last_known_state.read().unwrap();
+        data.clone()
+    }
+
+    pub async fn get_metadata(&self) -> Result<Response<jsonrpc::Metadata>, Error> {
+        self.send("get_metadata", json!([])).await
+    }
+
+    pub async fn get_metadata_by_version(
+        &self,
+        version: u64,
+    ) -> Result<Response<jsonrpc::Metadata>, Error> {
+        self.send("get_metadata", json!([version])).await
+    }
+
+    pub async fn get_account(
+        &self,
+        address: &AccountAddress,
+    ) -> Result<Response<Option<jsonrpc::Account>>, Error> {
+        self.send_opt("get_account", json!([address])).await
+    }
+
+    pub async fn get_account_transaction(
+        &self,
+        address: &AccountAddress,
+        seq: u64,
+        include_events: bool,
+    ) -> Result<Response<Option<jsonrpc::Transaction>>, Error> {
+        self.send_opt(
+            "get_account_transaction",
+            json!([address, seq, include_events]),
+        )
+        .await
+    }
+
+    pub async fn get_account_transactions(
+        &self,
+        address: &AccountAddress,
+        start_seq: u64,
+        limit: u64,
+        include_events: bool,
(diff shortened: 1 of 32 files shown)


libra-action:

Breaking changes policy:

2. Adding fields is ok.

davidiw:
Oh man... proto? Is there no other way to do this :(

Bad memories of trying to remove proto from the code base :P

This is very clever though...

Curious, would it be possible to reuse @matbd's code generator framework for LCS?

xli:
> Oh man... proto? Is there no other way to do this :(
> 
> Bad memories of trying to remove proto from the code base :P
> 
> This is very clever though...
> 
> Curious, would it be possible to reuse @matbd's code generator framework for LCS?

Had discussion related to this https://fb.workplace.com/groups/673683343488644/permalink/767450187445292/

Basically, the code generator framework still has a long way to support more languages, and @matbd has no interest to support JSON format serialize & deserialize.

Here the protobuf is used to define interface data structure for making sure client sdk languages' can get limited but good enough backward and forward compatible support, which reduces headache for anyone implementing JSON-API client with strong type for response data, and this is what protobuf built for. 

I understand it maybe bring back something not ideal, but it is really exactly what we need for a PRC interface.

client/json-rpc/src/state.rs
```diff
@@ -0,0 +1,21 @@
...
+
+#[derive(Debug, Clone, PartialEq, PartialOrd)]
+pub struct State {
+    pub chain_id: u8,
+    pub version: u64,
+    pub timestamp_usecs: u64,
+}
+
+impl State {
+    pub fn new_from_response(resp: &response::JsonRpcResponse) -> Self {
```
metajack, line 14: This would usually be called `from_response`.

client/json-rpc/src/client/async_client.rs
```diff
@@ -0,0 +1,1104 @@
...
+        address: &AccountAddress,
+    ) -> Result<Option<jsonrpc::Account>, Error> {
+        self.send_opt("get_account", json!([address])).await
+    }
+
+    pub async fn get_account_transaction(
+        &self,
+        address: &AccountAddress,
+        seq: u64,
+        include_event: bool,
```
metajack, line 126: Suggest plural `include_events`.

client/json-rpc/src/blocking/client.rs
```diff
@@ -0,0 +1,1078 @@
...
+        loop {
+            let ret = self.send_without_retry(method, &params);
+            match ret {
+                Ok(t) => return Ok(t),
+                Err(err) => {
+                    if !self.retry.is_retriable(&err) {
+                        return Err(err);
+                    }
+                    if retries < self.retry.max_retries(&err) {
+                        retries += 1;
```
mimoo, line 216: match retries.checked_add(1) {
  Some(i) if i < self.retry.max_retries(&err) => { 
     retries = i;
     // ...
  }
  _ => return Err(err)
};

client/json-rpc/src/client/async_client.rs
```diff
@@ -0,0 +1,1104 @@
...
+                }
+                if let Some(ref vm_status) = txn.vm_status {
+                    if vm_status.r#type != jsonrpc::VM_STATUS_EXECUTED {
+                        return Err(WaitForTransactionError::TransactionExecutionFailed(txn));
+                    }
+                }
+                return Ok(txn);
+            }
+            if let Some(state) = self.last_known_state() {
+                if expiration_time_secs <= state.timestamp_usecs / 1_000_000 {
```
mimoo, line 91: I don't think there's any problem in writing this, my understanding is that integer division's only edge-case is division by 0, which clearly can't happen here. I created an issue for that here: https://github.com/rust-lang/rust-clippy/issues/6209

testsuite/cli/src/client_proxy.rs
```diff
@@ -316,28 +319,35 @@ impl ClientProxy {
...
+            .map(|amt_view| {
+                let info = currency_info.get(&amt_view.currency).ok_or_else(|| {
+                    format_err!(
+                        "Unable to get currencyy info for balance {}",
+                        amt_view.currency
+                    )
+                })?;
+                if info.scaling_factor > 0 {
+                    let whole_num = amt_view.amount / info.scaling_factor;
+                    let remainder = amt_view.amount % info.scaling_factor;
```
mimoo, line 335:
```rust
match amt_view.amount.checked_div(info.scaling_factor) {
  Some(whole_num) => { ... }
  _ => Err(...)
}
```

testsuite/cli/src/client_proxy.rs
```diff
@@ -679,60 +665,83 @@ impl ClientProxy {
...
+    }
+
+    /// Submit transaction and waits for the transaction executed
+    pub fn submit_and_wait(&mut self, txn: &SignedTransaction, is_blocking: bool) -> Result<()> {
+        self.client.submit_transaction(&txn)?;
+        if is_blocking {
+            self.wait_for_signed_transaction(txn)?;
+        } else {
+            self.update_account_seq(&txn.sender(), txn.sequence_number() + 1);
+        }
```
mimoo, line 696: this is testsuite, and it's also client code, so it's not the biggest deal. You can either ignore this line by adding `#[allow(clippy::integer_arithmetic)]` before it, or just replace the addition with `txn.sequence_number().checked_add(1).expect("seqnum can't reach u64::max");`

client/json-rpc/src/client/async_client.rs
```diff
@@ -0,0 +1,1104 @@
...
+};
+use serde::de::DeserializeOwned;
+use serde_json::json;
+use std::time::Duration;
+
+#[derive(Debug)]
+pub struct Client {
+    pub http_client: reqwest::Client,
+    pub server_url: reqwest::Url,
+    pub retry: std::boxed::Box<dyn RetryStrategy>,
```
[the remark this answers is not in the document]
xli: I think different people may have different opinion about what error can be retried depending on their environment, hence which errors are retriable is part of RetryStrategy, which is also general design for a retry component.

client/json-rpc/src/async_client/client.rs
```diff
@@ -0,0 +1,324 @@
...
+};
+use serde::de::DeserializeOwned;
+use serde_json::json;
+use std::time::Duration;
+
+#[derive(Debug)]
+pub struct Response<R> {
+    pub result: R,
+    pub state: State,
+}
```
metajack, line 21: I think we should `impl<R> Deref for Response<R>` that returns the interior. Then you can do

```
 let result = client.get_account(address).await?;
let version = result.state;
if let Some(acc) = *result {
  // ....
}
```

testsuite/cli/src/client_proxy.rs
```diff
@@ -679,60 +677,85 @@ impl ClientProxy {
...
-                    } else {
-                        break Err(format_err!(
-                            "transaction failed to execute; status: {:?}!",
-                            txn_view.vm_status
-                        ));
-                    }
+    /// Wait for transaction, this function is not safe for waiting for a specific transaction,
+    /// should use wait_for_signed_transaction instead.
+    /// TODO: rename to wait_for_account_seq or remove
+    pub fn wait_for_transaction(&self, address: AccountAddress, seq: u64) -> Result<()> {
```
metajack, line 683: Since now this isn't used, can we just delete it completely?

client/json-rpc/src/lib.rs
```diff
@@ -13,3 +13,7 @@ pub use client::{
 pub use libra_json_rpc_types::{errors, views};
 pub use libra_types::{account_address::AccountAddress, transaction::SignedTransaction};
 pub use response::{JsonRpcResponse, ResponseAsView};
+
+// new implementation module
+
+pub mod async_client;
```
metajack, line 19: Maybe collapse async_client/* into this dir? Or if you don't like that another suggestion would be to pub use the relevant types directly from here.

metajack (approved):
This looks pretty good now! Few comments below.

testsuite/cli/src/client_proxy.rs
xli: it is used by smoke test suite, and in smoke test suite case it does not know the transaction details, only know account address and sequence.
may need think how to handle it better in smoke test, but probably worry about it later.

client/json-rpc/src/lib.rs
xli: I intend to make everything under async_client, it will be easier to fix later by search this word later after we finalize the rust-client codebase.

because some names may conflict if we direct pub in here, like JsonRpcResponse, 
in my mind, as long as it is a clean mod import, then next step is decide whether we move the whole async_client to a separate codebase, if so, we delete this one later and change the use path.
if we don't move to separate codebase, then we need figure out how to clean up legacy code in this module, maybe move all of them in to a legacy module and let everyone know (at least AOS uses this), or we move legacy code into separate codebase and ask people don't migrate to new client to use that instead.
Let me know what you think.

client/json-rpc/src/async_client/client.rs
xli: interesting, let me check it out.

testsuite/smoke-test/src/state_sync.rs
```diff
@@ -236,7 +236,7 @@ fn test_startup_sync_state_with_empty_consensus_db() {
     client_1
         .transfer_coins(&["tb", "0", "1", "10", "Coin1"], true)
         .unwrap();
-    client_0.wait_for_transaction(sender_address, 2).unwrap();
+    client_0.wait_for_transaction(sender_address, 1).unwrap();
```
davidiw, line 239: I see, so indeed we do wait on arbitrary transactions

davidiw (commented):
looks good! just a few nits here and there

testsuite/cli/src/client_proxy.rs
xli: Let me think about it, current change just keep it live to limit the scope of this PR.
Our smoke test suite uses it, I need think about whether we can update smoke test to avoid use this function and remove it.

json-rpc/types/src/lib.rs
```diff
@@ -4,3 +4,38 @@
...
+
+#[cfg(test)]
+mod tests {
+    use crate::{is_default, proto::types::Amount};
+
+    #[test]
+    fn test_is_default() {
+        let amount = Amount {
+            amount: 0,
+            currency: "Coin1".to_string(),
```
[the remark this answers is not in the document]
xli: I change it something else, this is mainly for testing is_default function, so only need a string.

PR #6548, opened 2020-10-19
[Tooling] Update the genesis and operator tooling to better support hex key formats and data trimming.

JoshLind:
## Motivation

This PR fixes an issue with the current genesis and operator tooling that reads keys from files. Currently, we support two file formats: lcs and hex. In the case of lcs, we want to read raw bytes from the file (non-utf8 encoded and files not created by hand). However, with hex, we want to support utf8 encoded files that may be created by hand. To do this, we handle both cases separately and execute whitespace trimming for hex files (e.g., to remove the unwanted characters from files created by hand).

Yes.

## Test Plan

Tested the tool locally using a currently failing example. Also updated the smoke tests to append a newline character to each hex encoded file to ensure that trimming works (as this is how the issue was flagged).

## Related PRs

None.

### DIFF
--- a/testsuite/smoke-test/src/test_utils.rs
+++ b/testsuite/smoke-test/src/test_utils.rs
@@ -202,10 +202,13 @@ fn fetch_backend_storage(
 }
 
 /// Writes a given public key to a file specified by the given path using hex encoding.
+/// Contents are written using utf-8 encoding and a newline is appended to ensure that
+/// whitespace can be handled by tests.
 pub fn write_key_to_file_hex_format(key: &Ed25519PublicKey, key_file_path: PathBuf) {
     let hex_encoded_key = hex::encode(key.to_bytes());
+    let key_and_newline = hex_encoded_key + "\n";
     let mut file = File::create(key_file_path).unwrap();
-    file.write_all(&hex_encoded_key.as_bytes()).unwrap();
+    file.write_all(&key_and_newline.as_bytes()).unwrap();
 }
 
 /// Writes a given public key to a file specified by the given path using lcs encoding.
--- a/config/management/src/lib.rs
+++ b/config/management/src/lib.rs
@@ -41,12 +41,16 @@ macro_rules! execute_command {
 use libra_crypto::ed25519::Ed25519PublicKey;
 use std::{convert::TryInto, fs, path::PathBuf};
 
+/// Reads a given ed25519 public key from file. Attempts to read the key using
+/// lcs encoding first. If this fails, attempts reading the key using hex.
 pub fn read_key_from_file(path: &PathBuf) -> Result<Ed25519PublicKey, String> {
-    let data = fs::read(path).map_err(|e| e.to_string())?;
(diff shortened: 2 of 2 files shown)


ankushagarwal:
Tested this and it worked!
https://github.com/libra/pre-mainnet-genesis/commit/0a479b654eca827c91de700a707c40e5493a1c7a

https://github.com/libra/pre-mainnet-genesis/commit/0339aceafabe069b0d6218d967cc5b677bf5e987

ankushagarwal (approved):
Thank you!!

davidiw (commented):
Thanks for resolving this quickly. What I don't understand is how this snuck in given that you added some pretty detailed testing of this. Should as amend the tests? Was the input weird? How do we prevent any quirks like this in the future?

Also verifying with the #gatekeepers before accepting.

@avery, @dariorussi, @sausagee

JoshLind:
Thanks @davidiw! From what I understand (and was able to reproduce), @ankushagarwal had a simple example of a file with a hex encoded public key in it (length = 64 chars). When the file is read (using fs::read_to_string(..)[1]), the resulting string is then passed to hex::decode[2], which internally calls "string.len()", which actually returns 65 (not 64), hence it was throwing an "Odd number of digits error". From what I can tell, this seems to be the result of the file containing whitespace (e.g., a newline character at the end of the file). To avoid this, we call trim() on the string to remove whitespace, which then results in a new string that actually has 64 characters, which can then be decoded successfully. 

The reason this wasn't caught in existing tests is because we weren't creating input files by hand, and thus no whitespace was being appended to them (this is in contrast to when I created a file by hand using VIM, where the newline was automatically appended). To prevent this from being an issue, I've updated the current smoke tests to append whitespace to the hex format file writer, to ensure we support trimming. Note: I haven't done the same for LCS formatted files, as these files should only ever be produced by the op tool (e.g., running "extract-public-key" using the op tool). Does this make sense?

[1] https://doc.rust-lang.org/std/fs/fn.read_to_string.html
[2] https://docs.rs/hex/0.3.1/hex/fn.decode.html

testsuite/smoke-test/src/test_utils.rs
```diff
@@ -202,10 +202,13 @@ fn fetch_backend_storage(
...
 
 /// Writes a given public key to a file specified by the given path using hex encoding.
+/// Contents are written using utf-8 encoding and a newline is appended to ensure that
+/// whitespace can be handled by tests.
 pub fn write_key_to_file_hex_format(key: &Ed25519PublicKey, key_file_path: PathBuf) {
     let hex_encoded_key = hex::encode(key.to_bytes());
+    let key_and_newline = hex_encoded_key + "\n";
     let mut file = File::create(key_file_path).unwrap();
-    file.write_all(&hex_encoded_key.as_bytes()).unwrap();
+    file.write_all(&key_and_newline.as_bytes()).unwrap();
```
davidiw, line 211: ahh here is the test modification :O

davidiw (approved):
/land

please submit another pr for cherry picking into 1.0

PR #6553, opened 2020-10-20
[backup-cli] removing dirs crate due to RUSTSEC-2020-0053

mimoo:
Removing that dependency to satisfy cargo-audit and [RUSTSEC-2020-0053](https://rustsec.org/advisories/RUSTSEC-2020-0053).

I also simplified the code as we probably don't care if `$HOME` is not set (as in, we'll just panic).


msmouse (approved):
Works for me. Thanks!

davidiw (commented):
Has this been tested in our "production" environments? Is this a replication of the underlying dirs::home_dir? Overall this makes sense, but I cannot imagine how this was or could have been tested. Also why is this going into the home directory instead of a temp directory?

@msmouse

msmouse:
@davidiw  we specify the dir explicitly in "production". It used to be under /tmp (and that's why the dependency was first introduced, because MacOs doesn't have $TMP etc), and moved to the home dir because it's more likely to be deleted under /tmp.

mimoo:
What this affects:
* the backup-cli now only relies on the `$HOME` environment variable being set, instead of previously checking `getpwuid_r` if `$HOME` is not set. It also won't run on windows anymore.

Why this is critical:
* the `dirs` crate triggered a [RUSTSEC Advisory](https://rustsec.org/advisories/RUSTSEC-2020-0053) advising not to rely on this crate anymore due to the lack of maintenance. If we want a clean output from [cargo-audit](https://github.com/RustSec/cargo-audit) we need to get this in.

What is the workaround (even if it is unintuitive / horrible):
* live with an unclean cargo-audit output.

davidiw:
so I suspect given the new guidelines, we can land this in master but we'll need to make /another/ PR that we would submit to be cherry-picked in 1.0

/land

mimoo:
> but we'll need to make /another/ PR that we would submit to be cherry-picked in 1.0

is it for me to do?

davidiw:
@mimoo, if you want this included in 1.0.

mimoo:
like this? https://github.com/libra/libra/pull/6572

PR #6557, opened 2020-10-20
[RFC][writeset-generator] Implement a unified api for writeset generation…

runtian-zhou:
… process.

## Motivation

Build an unified api to help genearte future upgrade transaction. The core api allows developers to:
1. Execute script/functions on top of existing (or even shadowed) states
2. Publish new modules

### Have you read the [Contributing Guidelines on pull requests (https://github.com/libra/libra/blob/master/CONTRIBUTING.md#pull-requests)?

Yes

## Test Plan

Added a test in e2e testsuite that:
1. Publishes a new module
2. Incremented the LibraVersion config.

### DIFF
--- a/language/libra-tools/writeset-transaction-generator/src/writeset_builder.rs
+++ b/language/libra-tools/writeset-transaction-generator/src/writeset_builder.rs
@@ -1,0 +1,199 @@
+// Copyright (c) The Libra Core Contributors
+// SPDX-License-Identifier: Apache-2.0
+
+use anyhow::format_err;
+use compiled_stdlib::{stdlib_modules, StdLibOptions};
+use libra_state_view::StateView;
+use libra_types::{
+    account_address::AccountAddress,
+    account_config::{self, libra_root_address},
+    transaction::{ChangeSet, Script, TransactionArgument, Version},
+};
+use libra_vm::{data_cache::RemoteStorage, txn_effects_to_writeset_and_events};
+use move_core_types::{
+    gas_schedule::{CostTable, GasAlgebra, GasUnits},
+    identifier::Identifier,
+    language_storage::{ModuleId, TypeTag},
+};
+use move_vm_runtime::{
+    data_cache::RemoteCache, logging::NoContextLog, move_vm::MoveVM, session::Session,
+};
+use move_vm_test_utils::{ChangeSet as MoveChanges, DeltaStorage};
+use move_vm_types::{
+    gas_schedule::{zero_cost_schedule, CostStrategy},
+    values::Value,
+};
+use once_cell::sync::Lazy;
+use vm::CompiledModule;
+
+pub static ZERO_COST_SCHEDULE: Lazy<CostTable> = Lazy::new(zero_cost_schedule);
+
+pub struct GenesisSession<'r, 'l, R>(Session<'r, 'l, R>);
+
+impl<'r, 'l, R: RemoteCache> GenesisSession<'r, 'l, R> {
+    pub fn exec_func(
+        &mut self,
+        sender: AccountAddress,
+        module_name: &str,
+        function_name: &str,
+        ty_args: Vec<TypeTag>,
+        args: Vec<Value>,
+    ) {
+        self.0
+            .execute_function(
+                &ModuleId::new(
+                    account_config::CORE_CODE_ADDRESS,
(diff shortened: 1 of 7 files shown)


language/libra-tools/writeset-transaction-generator/src/writeset_builder.rs
```diff
@@ -0,0 +1,199 @@
...
+    override_move_changes: Option<MoveChanges>,
+    procedure: F,
+    to_be_published_modules: &[CompiledModule],
+) -> ChangeSet
+where
+    F: FnOnce(&mut GenesisSession<DeltaStorage<RemoteStorage<S>>>),
+{
+    let move_vm = MoveVM::new();
+    let mut move_changes = override_move_changes.unwrap_or_else(MoveChanges::new);
+    shadow_reconfiguration(state_view, &mut move_changes);
```
runtian-zhou, line 165: A reconfiguration is executed in shadow mode so that no real reconfiguration will be triggered during the execution phase.

language/libra-tools/writeset-transaction-generator/src/writeset_builder.rs
```diff
@@ -0,0 +1,199 @@
...
+            }
+        }
+    }
+}
+
+pub fn build_changeset<S: StateView, F>(
+    state_view: &S,
+    override_move_changes: Option<MoveChanges>,
+    procedure: F,
+    to_be_published_modules: &[CompiledModule],
```
runtian-zhou, line 158: One question is if we would like to shadow the existing modules on chain with the modules that we are planning to publish by default. I can see arguments for both points:
- Shadow because we want to be able to invoke init functions in those modules.
- Not shadow, as it will give more flexibility in case if some functions in the old modules need to be invoked.

The current api uses the none shadow version.

language/libra-tools/writeset-transaction-generator/src/writeset_builder.rs
```diff
@@ -0,0 +1,199 @@
...
+            .map_err(|err| format_err!("Unexpected VM Error: {:?}", err))
+            .unwrap()
+    };
+
+    for module in to_be_published_modules {
+        let mut module_bytes = vec![];
+        module.serialize(&mut module_bytes).unwrap();
+        // TODO: Check compatibility between old and new modules.
+        effect.modules.push((module.self_id(), module_bytes));
+    }
```
vgao1996, line 183: I don't think the check here is sufficient, and conceptually I wouldn't think of it as "we need to check whether the new modules are compatible" but rather "when the effects are applied to the base state, will the new state be compatible", which includes
- Checking if the modules are compatible.
- Checking if the resources are in good state -- we must not publish a resource that's defined in a module that's part of the overriding changes.

language/libra-tools/writeset-transaction-generator/src/writeset_builder.rs
vgao1996: I wonder why you're taking the modules to be published as an array instead of allowing them to be published by the procedure? In other words, what's your concern about adding a `fn publish_module` to `GenesisSession`? To me this seems to be both simpler and more expressive, as now you can use these modules during the process.

I get that you want to check if the new modules are compatible. However it's still possible to do that if they are published by the procedure (see my comment below).

runtian-zhou: Great question! The reason why I deliberatly removed `publish_module` from `GenesisSession` is because `publish_module` requires that module to be not existent on chain in VM currently. If we change the semantic of that function we will need to change the way how Libra Adapter works, which is sth I don't want to touch at the moment.

vgao1996: > because publish_module requires that module to be not existent on chain in VM currently
Technically this is not true. `move_vm_test_utils::ChangeSet` allows you to pretend a module is deleted through shadowing. You also have the option to generate a `move_vm_test_utils::ChangeSet` (or WriteSet) and then merge it directly with another ChangeSet that contains the module you want to overwrite.

runtian-zhou:
Closing this PR as we suggest the other PR is a more preferrable due to safety reasons. Will explain the reasoning in the other PR. There's still some remaining comments made by @vgao1996 and I'll address them in the other PR as well

PR #6559, opened 2020-10-21
[gk] turn off code freeze

sausagee:
Code freeze complete. Turn off gk.


davidiw (approved):
/land

at last, winter is over!

PR #6563, opened 2020-10-21
[gha][lbt] always run compat unless ks

rustielin:
To prevent breaking changes from slipping into `master`. This PR disables the check for `[breaking]` in commit messages to bypass compatibility test. The only way to skip that will be to activate the kill switch `KILL_SWITCH_LAND_BLOCKING_COMPAT` now

### DIFF
--- a/.github/workflows/land-blocking.yml
+++ b/.github/workflows/land-blocking.yml
@@ -79,20 +79,14 @@ jobs:
           fi;
       - name: Determine which cluster-test suite to run
         if: steps.check_ks.outputs.should_run == 'true'
-        # Runs land_blocking if KS activated or [breaking] in commit message heading.
-        # Runs land_blocking_compat otherwise. Compat needs to find an image with PREV_TAG
-        # to test against, otherwise will create one from BASE_GIT_REV in the next step,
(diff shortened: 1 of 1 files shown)


andll:
I would probably vote for smth more flexible, like introducing new kill switch like allow_breaking_change(and having it off by default)

rustielin:
> I would probably vote for smth more flexible, like introducing new kill switch like allow_breaking_change(and having it off by default)

How is that different than the current kill switch `KILL_SWITCH_LAND_BLOCKING_COMPAT`? The current KS will bypass the compat test (to allow breaking change) but still run the rest of the land-blocking cluster test suite.

andll:
Difference is that activating KILL_SWITCH_LAND_BLOCKING_COMPAT will disable compat test for any PR, it is hard to target it to only one specific change

sausagee:
> Difference is that activating KILL_SWITCH_LAND_BLOCKING_COMPAT will disable compat test for any PR, it is hard to target it to only one specific change

The goal is to use KS to target specific PR such that only thoroughly reviewed breaking change can land.

sausagee (approved):

Issue #6564, opened 2020-10-21
[logging] Prevent high value u64s from being logged


gregnazario:
## Problem
Unsigned 64 bit longs are [not really supported](https://github.com/elastic/elasticsearch/issues/32434).  When a `u64` is serialized into JSON, and the highest bit is set (the bit that in signed would make it negative), it is converted to a `BigInt` in JSON serialization (e.g. 1e9).  This causes Elasticsearch to not know what to do with it, because of an indexing mismatch (long before then BigInt after).  As a result, the logs won't be accepted by Elasticsearch, and clog up the logging system.

### Prevention
* Make something that possibly checks that we don't log `u64` in the system?

### Fixing logs that log u64
* Ensure that all `u64` are serialized into `BigInt` format or a string if it's used as an identifier e.g. `BigInt`

## Known Examples
* `trace_id` in JSON-RPC is a randomly generated u64 number, which doesn't work when the highest bit is randomly set.

PR #6565, opened 2020-10-21
[json-rpc] Change trace-id to be a hex string

gregnazario:
### Overview
As mentioned in https://github.com/libra/libra/issues/6564, unsigned
longs are not supported by Elasticsearch.  This just changes the trace
id to a hex string to make it more friendly, and prevent these issues.

I won't push this immediately

### DIFF
--- a/json-rpc/src/methods.rs
+++ b/json-rpc/src/methods.rs
@@ -124,7 +124,7 @@ type RpcHandler =
 pub(crate) type RpcRegistry = HashMap<String, RpcHandler>;
 
 pub(crate) struct JsonRpcRequest {
-    pub trace_id: u64,
+    pub trace_id: String,
     pub params: Vec<Value>,
     pub ledger_info: LedgerInfoWithSignatures,
 }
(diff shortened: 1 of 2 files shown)


libra-action:

Breaking changes policy:

2. Adding fields is ok.

xli (approved):

xli:
Thanks, just saw this in the log.

json-rpc/src/methods.rs
```diff
@@ -124,7 +124,7 @@ type RpcHandler =
 pub(crate) type RpcRegistry = HashMap<String, RpcHandler>;
 
 pub(crate) struct JsonRpcRequest {
-    pub trace_id: u64,
+    pub trace_id: String,
```
davidiw, line 127: any point in changing this? it allows weird stuff to be inserted here

can we not just convert the trace id to a string in the logger?

gregnazario: I figured the hex string would be easier to read and more compact / consistent.  When stringifying a number with the highest bit set, it changes the number from a base 10 to a big int e.g. 1e10 or whatever.

PR #6566, opened 2020-10-21
[libra framework] specify initialize in LibraAccount

emmazzz:
## Motivation

This PR specified and made opaque some functions in `LibraAccount` including `initialize` and `make_account`, as well as `create_libra_root_account` and `create_treasury_compliance_account`.

Yes.


language/stdlib/modules/LibraAccount.move
```diff
@@ -1040,6 +1065,48 @@ module LibraAccount {
         make_account(lr_account, auth_key_prefix)
     }
 
+    spec fun create_libra_root_account {
+        pragma opaque;
+        include CreateLibraRootAccountModifies;
+        include CreateLibraRootAccountAbortsIf;
```
emmazzz, line 1071: Maybe instead of having all the `aborts_if exists<Resource>`s here, we should have a predicate in `Genesis` saying that before genesis no resource exists at any address?

wrwg: That is an interesting question. One problem is that there are intermediate states after the initial empty state anyway. And in fact, someone may call accidentally initialize functions twice.

wrwg (approved):

PR #6571, opened 2020-10-22
[Operational Tool] Add --disable-address-validation flag to avoid spurious address errors

JoshLind:
## Motivation

This PR adds a flag `--disable-address-validation` to the commands in the op tool that update the validator config. It also disables address validation for any key rotation operations (consensus key, validator network key, and full node network key), as these operations are orthogonal to setting the network address. Without this PR, we currently get spurious address validation errors when trying to rotate keys (e.g., when my machine can't resolve an address locally, but that address is indeed valid in the deployment environment).

To achieve this, the PR offers several commits:
1. Add the --disable-address-validation flag to the tool, and disable address validation for all key rotation operations.
2. Add a missing smoke test for the full node network key rotation command (orthogonal, but needed).
3. Add a smoke test for the new flag, as well as update the test helper
4. Update auto-validate to use a checked add

Yes.

## Test Plan

All smoke tests pass locally, including the newly added ones. Moreover, running the tool locally now allows me to rotate the consensus key, whereas it previously fails (because my machine can't connect to the AWS address that's only available within the k8s deployment...)

Before this PR:
```
% ./libra-operational-tool rotate-consensus-key --config k8_config_file                             
{
  "Error": "Invalid arguments: validator address: Failed to resolve address '/dns4/val0-libra-validator-validator-lb/tcp/6180': failed to lookup address information: nodename nor servname provided, or not known"
}
```

After this PR:
```
% ./libra-operational-tool rotate-consensus-key --config k8_config_file
{
  "Result": "Executed"
}

```

## Related PRs

This PR is built on the auto validate PR: https://github.com/libra/libra/pull/6560

### DIFF
--- a/config/management/operational/src/command.rs
+++ b/config/management/operational/src/command.rs
@@ -118,7 +118,7 @@ impl std::fmt::Display for CommandName {
             CommandName::RemoveValidator => "remove-validator",
             CommandName::RotateConsensusKey => "rotate-consensus-key",
             CommandName::RotateOperatorKey => "rotate-operator-key",
-            CommandName::RotateFullNodeNetworkKey => "rotate-fullnode-network-key",
+            CommandName::RotateFullNodeNetworkKey => "rotate-full-node-network-key",
             CommandName::RotateValidatorNetworkKey => "rotate-validator-network-key",
             CommandName::SetValidatorConfig => "set-validator-config",
             CommandName::SetValidatorOperator => "set-validator-operator",
--- a/config/management/operational/src/test_helper.rs
+++ b/config/management/operational/src/test_helper.rs
@@ -225,6 +225,7 @@ impl OperationalTool {
         fullnode_address: Option<NetworkAddress>,
         backend: &config::SecureBackend,
         disable_validate: bool,
+        disable_address_validation: bool,
     ) -> Result<TransactionContext, Error> {
         let args = format!(
             "
@@ -235,6 +236,7 @@ impl OperationalTool {
                 --json-server {host}
                 --validator-backend {backend_args}
                 {disable_validate}
+                {disable_address_validation}
             ",
             command = command(TOOL_NAME, CommandName::SetValidatorConfig),
             host = self.host,
@@ -243,6 +245,8 @@ impl OperationalTool {
             validator_address = optional_arg("validator-address", validator_address),
             backend_args = backend_args(backend)?,
             disable_validate = optional_flag("disable-validate", disable_validate),
+            disable_address_validation =
+                optional_flag("disable-address-validation", disable_address_validation),
         );
 
         let command = Command::from_iter(args.split_whitespace());
--- a/config/management/src/validator_config.rs
+++ b/config/management/src/validator_config.rs
@@ -42,10 +42,13 @@ impl ValidatorConfig {
         fullnode_address: NetworkAddress,
         validator_address: NetworkAddress,
         reconfigure: bool,
+        disable_address_validation: bool,
     ) -> Result<Transaction, Error> {
-        // Verify addresses
-        validate_address("validator address", &validator_address)?;
-        validate_address("fullnode address", &fullnode_address)?;
+        if !disable_address_validation {
+            // Verify addresses
+            validate_address("validator address", &validator_address)?;
+            validate_address("fullnode address", &fullnode_address)?;
+        }
 
         let config = self.config()?;
         let mut storage = config.validator_backend();
--- a/testsuite/smoke-test/src/operational_tooling.rs
+++ b/testsuite/smoke-test/src/operational_tooling.rs
@@ -149,6 +149,52 @@ fn test_create_validator_lcs_file() {
 }
 
 #[test]
+fn test_disable_address_validation() {
+    let num_nodes = 1;
+    let (_env, op_tool, backend, _) = launch_swarm_with_op_tool_and_backend(num_nodes, 0);
+
+    // Try to set the validator config with a bad address and verify failure
+    let bad_network_address = NetworkAddress::from_str("/dns4/127.0.0.1/tcp/1234").unwrap();
+    op_tool
+        .set_validator_config(
+            Some(bad_network_address.clone()),
+            None,
+            &backend,
+            false,
+            false,
+        )
+        .unwrap_err();
+
+    // Now disable address verification to set the validator config with a bad network address
+    let txn_ctx = op_tool
+        .set_validator_config(Some(bad_network_address), None, &backend, false, true)
+        .unwrap();
+    assert_eq!(VMStatusView::Executed, txn_ctx.execution_result.unwrap());
+
+    // Rotate the consensus key and verify that it isn't blocked by a bad network address
+    let _ = op_tool.rotate_consensus_key(&backend, false).unwrap();
+
+    // Rotate the validator network key and verify that it isn't blocked by a bad network address
+    let _ = op_tool
+        .rotate_validator_network_key(&backend, false)
+        .unwrap();
+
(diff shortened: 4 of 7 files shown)


config/management/operational/src/account_resource.rs
```diff
@@ -121,8 +123,14 @@ impl RotateOperatorKey {
...
+        let mut transaction_context =
             client.submit_transaction(rotate_key_txn.as_signed_user_txn().unwrap().clone())?;
-        Ok((txn_ctx, new_storage_key))
+
+        // Perform auto validation if required
+        transaction_context = self
+            .auto_validate
+            .execute(config.json_server, transaction_context)?;
+
+        Ok((transaction_context, new_storage_key))
```
gregnazario, line 134: Btw, the build on top of the previous commits thing is kinda weird, but I do like you separated it out by which commits

config/management/operational/src/test_helper.rs
```diff
@@ -295,6 +305,7 @@ impl OperationalTool {
         self.rotate_key(
             backend,
             disable_validate,
+            false,
```
gregnazario, line 308: I think we might want to either not validate the addresses, or allow it to be configurable.  Makes sense to me for this to be `true` because the operator key doesn't affect the network addresses (therefore if the addresses were wrong already, we shouldn't stop operator key rotation)

config/management/operational/src/test_helper.rs
```diff
@@ -278,10 +286,12 @@ impl OperationalTool {
         &self,
         backend: &config::SecureBackend,
         disable_validate: bool,
+        disable_address_validation: bool,
     ) -> Result<(TransactionContext, Ed25519PublicKey), Error> {
         self.rotate_key(
             backend,
             disable_validate,
+            disable_address_validation,
```
gregnazario, line 294: I think this is another instance where, basically we should only validate the addresses when someone specifically changes them?

config/management/src/validator_config.rs
```diff
@@ -42,10 +42,13 @@ impl ValidatorConfig {
...
         reconfigure: bool,
+        disable_address_validation: bool,
     ) -> Result<Transaction, Error> {
-        // Verify addresses
-        validate_address("validator address", &validator_address)?;
-        validate_address("fullnode address", &fullnode_address)?;
+        if !disable_address_validation {
+            // Verify addresses
+            validate_address("validator address", &validator_address)?;
+            validate_address("fullnode address", &fullnode_address)?;
```
gregnazario, line 50: Okay, so now that I've looked at this and I'm thinking about it.  We should probably not validate addresses when we rotate the keys.  The addresses were already put in previously, and rotating a key is not at the fault of the current operator.  Only when an operator puts in the non-key information, should it validate that it's correct.  Thoughts?

mimoo:
for the #integer-overflow finding, I just suggest that you mark the whole crate as less risky via `#![allow(clippy::integer_arithmetic)]`

(or you can also do the right thing and use a checked operation, but I presume we don't care that much for operational toolings)

config/management/operational/src/account_resource.rs
JoshLind: Thanks @gregnazario! Yeah, I was kind of hoping that would be reviewed first.. so we'd ultimately not need to worry about those commits :P

config/management/operational/src/test_helper.rs
JoshLind: See my comment below.

config/management/operational/src/test_helper.rs
JoshLind: See my comment below.

config/management/src/validator_config.rs
JoshLind: Yeah, I basically did this change with the mindset of making as few code changes as possible. Given that they all eventually call into the set-validator-config command, I figured we could just propagate the flag.  I was trying to avoid refactoring everything. However, it might make sense to consider this now. Let's hold off until the original has landed.

JoshLind:
Thanks all, let's hold off until the base PR has landed first. Will return to this aftwards.

config/management/src/validator_config.rs
davidiw: actually, I would argue it is nice to have this validation... maybe it is over kill, I dunno, but if we have a flag to disable it... I think it is good.

config/management/operational/src/command.rs
```diff
@@ -118,7 +118,7 @@ impl std::fmt::Display for CommandName {
             CommandName::RemoveValidator => "remove-validator",
             CommandName::RotateConsensusKey => "rotate-consensus-key",
             CommandName::RotateOperatorKey => "rotate-operator-key",
-            CommandName::RotateFullNodeNetworkKey => "rotate-fullnode-network-key",
+            CommandName::RotateFullNodeNetworkKey => "rotate-full-node-network-key",
```
davidiw, line 121: if you do this, you're going to have to update the scripts and the runbooks.....

davidiw: I think

JoshLind:
Now that the auto validate PR has landed, I've updated this PR:
- I've disabled all network address validations for key rotations. As @gregnazario pointed out, they aren't that helpful. Moreover, failing on an invalid address when performing a key rotation is really confusing from a user perspective. Instead it's just the set-validator-config command in the op tool, and the validator-config command in the genesis tool that should do validation.
- I've updated the previously flagged potential integer overflow/underflow to use a checked add. If this happens, the tool should just panic. I don't want to ignore it for the whole crate because there might be other interesting cases ('cc @mimoo).

config/management/operational/src/command.rs
JoshLind: So there's two points to make here:
- The existing runbooks don't actually talk about this command at all. So we need to improve our runbooks.
- The command is actually "rotate-full-node-network-key" (note: RotateFullNodeNetworkKey -- "Node" has a capital, so this is currently a bug which wasn't detected because there was no smoke test for it). I don't mind which we use, so I chose the variant with the smallest number of code changes. Overall, it's probably worth pointing out that there's a lot of places in our codebase where we use "full node" and "fullnode". It's not great and it's kinda ugly. Perhaps we should discuss this with the networking folks and settle on something so we can clean up the codebase?

testsuite/smoke-test/src/operational_tooling.rs
```diff
@@ -148,6 +148,52 @@ fn test_create_validator_lcs_file() {
...
+    // Rotate the operator key and verify that it isn't blocked by a bad network address
+    let _ = op_tool.rotate_operator_key(&backend, false).unwrap();
+
+    // Update the validator network address with a valid address
+    let new_network_address = NetworkAddress::from_str("/ip4/10.0.0.16/tcp/80").unwrap();
+    let _ = op_tool
+        .set_validator_config(Some(new_network_address), None, &backend, false, false)
+        .unwrap();
+}
+
```
davidiw, line 196: I don't really think we need a smoke test for this, when we can get everything we need from an integration test.

davidiw (approved):
consider deleting the smoke test.

testsuite/smoke-test/src/operational_tooling.rs
JoshLind: Yeah, I think that's fair, although we do check that transactions are being executed by the blockchain allowing us to set bad networking addresses. But for the other operations, we could check the tool works using integration/unit style testing. I'll leave this as-is for now though as it's simpler to check all this functionality in once place. Once we start to think about our testing principles we can take a look at all the tests to figure out where things should live.

JoshLind:
Thanks, David! 😄 

/land

PR #6572, opened 2020-10-22
[backup-cli] removing dirs crate due to RUSTSEC-2020-0053

mimoo:
What this affects:
* the backup-cli now only relies on the `$HOME` environment variable being set, instead of previously checking `getpwuid_r` if `$HOME` is not set. It also won't run on windows anymore.

Why this is critical:
* the `dirs` crate triggered a [RUSTSEC Advisory](https://rustsec.org/advisories/RUSTSEC-2020-0053) advising not to rely on this crate anymore due to the lack of maintenance. If we want a clean output from [cargo-audit](https://github.com/RustSec/cargo-audit) we need to get this in.

What is the workaround (even if it is unintuitive / horrible):
* live with an unclean cargo-audit output.


davidiw:
Not of a critical enough nature to pull in at this time.

PR #6573, opened 2020-10-22
[stdlib] Require owner to set the operator in move modules

valerini:
Currently ValidatorConfig::get_operator() returns the address of validator owner if operator is unset, then the owner may set the config locally but may not call LibraSystem::update_and_reconfigure() which checks that the caller has operator role.
This PR makes ValidatorConfig::get_operator() return an error in case operator is not set, so the owner cannot set it's own config and be added to the set, the owner will have to set the operator who sets the config.

## Motivation

The owner being its own operator was an inconsistency in the code that is an artifact of how the tests were set-up, now the operator is set everywhere in the tests, so we can require the operator to always be set for the owner.

Yes

## Test Plan

cargo xtest


davidiw:
This is awesome

language/stdlib/modules/ValidatorConfig.move
```diff
@@ -157,7 +157,8 @@ module ValidatorConfig {
         include Roles::AbortsIfNotValidator{validator_addr: sender};
         include AbortsIfNoValidatorConfig{addr: sender};
         ensures !spec_has_operator(Signer::spec_address_of(validator_account));
-        ensures get_operator(sender) == sender;
+        // TODO(ddill): perhaps ensure that get_operator returns an error
+        // ensures get_operator(sender) == sender;
```
sblackshear, line 161: Should we delete this? It's no longer a postcondition of `remove_operator`.

language/stdlib/modules/ValidatorConfig.move
```diff
@@ -275,16 +278,18 @@ module ValidatorConfig {
...
         assert(exists<ValidatorConfig>(addr), Errors::not_published(EVALIDATOR_CONFIG));
         let t_ref = borrow_global<ValidatorConfig>(addr);
-        *Option::borrow_with_default(&t_ref.operator_account, &addr)
+        assert(Option::is_some(&t_ref.operator_account), Errors::invalid_argument(EVALIDATOR_CONFIG));
+        *Option::borrow(&t_ref.operator_account)
     }
 
     spec fun get_operator {
         pragma opaque;
+        pragma verify_duration_estimate = 100;
```
sblackshear, line 292: I think this now needs an `aborts_if` for case where `operator_account` is None. Actually, I'm surprised the prover tests are passing without this. Do you know why @wrwg?

sblackshear (commented):
LGTM modulo the minor prover fixes

mickayz (approved):
LGTM!

davidiw:
Let's try to get this in, in the next couple weeks or so.

davidiw (approved):
/land

this may require setting the breaking changes flag

PR #6576, opened 2020-10-22
[consensus] implementing more granular logging when receiving a proposal

mimoo:
This change turns the `ReceiveProposal` debug log to an info log,
and uses fields to log the data instead of a message.
This is useful to create metrics and alerts in Kibana,
specifically detect when we see tuples of the form {epoch, round, proposer, proposal_hash}
where we have {e, r, p, ph}, {e, r, p, ph'}, and ph != ph'

### DIFF
--- a/consensus/src/round_manager.rs
+++ b/consensus/src/round_manager.rs
@@ -489,6 +489,16 @@ impl RoundManager {
         let author = proposal
             .author()
             .expect("Proposal should be verified having an author");
+
+        info!(
+            self.new_log(LogEvent::ReceiveProposal).remote_peer(author),
+            proposal = %proposal,
+            block_epoch = proposal.epoch(),
+            block_round = proposal.round(),
+            block_hash = proposal.id(),
+            block_parent_hash = proposal.quorum_cert().certified_block().id(),
+        );
+
         ensure!(
             self.proposer_election.is_valid_proposal(&proposal),
             "[RoundManager] Proposer {} for block {} is not a valid proposer for this round",
@@ -507,11 +517,6 @@ impl RoundManager {
         );
 
         let proposal_round = proposal.round();
-
-        debug!(
-            self.new_log(LogEvent::ReceiveProposal).remote_peer(author),
-            "{}", proposal
-        );
 
         if let Some(time_to_receival) = duration_since_epoch().checked_sub(block_time_since_epoch) {
             counters::CREATION_TO_RECEIVAL_S.observe_duration(time_to_receival);


mimoo:
@gregnazario can you confirm that I can turn this into an info log? @zekun000 can you confirm that this is fine :) (moving the log earlier)

consensus/src/round_manager.rs
```diff
@@ -489,6 +489,16 @@ impl RoundManager {
...
             .author()
             .expect("Proposal should be verified having an author");
+
+        info!(
+            self.new_log(LogEvent::ReceiveProposal).remote_peer(author),
+            "block_epoch" = proposal.epoch(),
+            "block_round" = proposal.round(),
+            "block_hash" = proposal.id(),
+            "block_parent_hash" = proposal.quorum_cert().certified_block().id(),
+            "block_proposer" = author,
```
gregnazario, line 499: You can drop the `"` around the names

gregnazario (commented):
I'm fine with this moving to `info!` as long as @zekun000 signs off on the volume and size of these logs won't be too big (I don't have a good estimation to give you on amount per second).

zekun000:
the volume is between 3 to 40 per second, each log size should be relatively small (~100 bytes)

gregnazario:
> the volume is between 3 to 40 per second, each log size should be relatively small (~100 bytes)

So, the current volume for the logging system is roughly ~60 per second.  So, in the worst case it's a 66% increase, in total logging traffic.  Given our current testing with cluster tests, we may lose all of these logs.  I think we need to work with @coryhill and see if we up some of the logging infrastructure before we push this

coryhill:
I believe so far with cluster_test we've only been losing logs on the nodes themselves, right? I don't think the central logging has been a bottleneck.

gregnazario:
> I believe so far with cluster_test we've only been losing logs on the nodes themselves, right? I don't think the central logging has been a bottleneck.

In what I've seen in the past, we lose logs because the central logging falls behind, so everything else falls behind, but I could be wrong.  I could investigate writing multiple write threads to the logging system to see if that's the bottleneck

davidiw:
> This is useful to create metrics and alerts in Kibana,
specifically detect when we see tuples of the form {epoch, round, proposer, proposal_hash}
where we have {e, r, p, ph}, {e, r, p, ph'}, and ph != ph'

Really? I didn't think we could have / had alerts for kibana?

mimoo:
we don't have alerts but we can have them :)

zekun000:
I don't think we can rely on this to detect byzantine behavior, a byzantine node can forge a fake log to accuse an innocent node.

unless we log the entire block including the signature to prove the author indeed signs it, but that will probably overwhelm the logging infra.

mimoo:
That's a good point, two things on that:

* That is still a signal we can get, even if it's not a "proof"
* We have the same issue on state-sync, I think we might want to come up with some per-valdiator local ephemeral storage to store proposals & votes that we see

wdyt?

mimoo:
Actually, thinking more about it, we could just pass the hash of the blocktype, that should be enough no?

consensus/src/round_manager.rs
```diff
@@ -489,6 +489,15 @@ impl RoundManager {
         let author = proposal
             .author()
             .expect("Proposal should be verified having an author");
+
+        info!(
+            self.new_log(LogEvent::ReceiveProposal).remote_peer(author),
+            block_epoch = proposal.epoch(),
+            block_round = proposal.round(),
+            block_hash = proposal.id(),
+            block_parent_hash = proposal.quorum_cert().certified_block().id(),
```
mimoo, line 498: @gregnazario 

* author: 32 bytes
* epoch: 8 bytes
* round: 8 bytes
* id: 32 bytes
* parent id: 32 bytes

gregnazario (approved):

PR #6540, opened 2020-10-17
[draft][language] implement new testing framework

vgao1996:
## Summary
This implements a new testing framework and a Move e2e testing flow. Its modular design will allow us to add new functionalities, tweak behaviors, reuse existing code, and create new testing flows with ease. 

The modularity is achieved by modeling the tests around the following core concepts:
  - State: a state object to be manipulated by tasks through a testing flow.
  - Operations: traits to specify the capabilities of the states.
  - `Task<S>`: essentially a trait describing a function object that takes a state of type S and returns a new state. 

Tasks are usually modeled against a generic state type, using the operation traits to interact with the states. This decouples the states and the tasks, allowing certain tasks to be reused in another testing flow. (For example, this PR only adds a Move testing flow, but when we add a Libra testing flow later, that Libra testing flow will be able to use the Move Compile Task we already have.

Eventually these new flows will replace the existing functional and e2e tests.

## Progress (Still a WIP, Not Ready for Review Yet)
- [x] Make error messages more user-friendly
- [x] Better error messages
- [ ] Missing features
  - [x] Support for Move Standard Lib
  - [ ] Support for Gas Metering
- [ ] Dummy tests for the Move testing flow itself
- [ ] Documentation

PR #6578, opened 2020-10-23
[e2e-testsuites] Make tests deterministic, add golden file

runtian-zhou:
## Motivation

Remove the non-determinism for tests in e2e-testsuites and checkin the golden file so that the outputs won't be off in the future.

The golden file check is disabled for proptests and some more sophisticated p2p payment tests.

Yes

## Test Plan

`cargo test`

Generated the fresh golden files and re-run the checks again locally to make sure golden file check won't fail in the second run.


language/testing-infra/e2e-tests/src/executor.rs
```diff
@@ -38,13 +40,17 @@ use move_vm_types::{
...
+
 /// Provides an environment to run a VM instance.
 ///
 /// This struct is a mock in-memory implementation of the Libra executor.
 #[derive(Debug)]
 pub struct FakeExecutor {
     data_store: FakeDataStore,
     block_time: u64,
+    executed_output: Option<GoldenOutputs>,
+    rng: KeyGen,
```
vgao1996, line 53: It feels a little bit awkward to me to add these to the FakeExecutor. Would it be better if we move this logic to the outside by implementing a wrapper or a few helper functions? Hardcoding this into the FakeExecutor will make it harder for other tests/libraries to depend on it.

runtian-zhou: The main reason is to reduce the changes needed for each individual tests. You could imagine that otherwise we will need to create a seed for each tests. I also tried to have one global seed, but it didn't work either as cargo will run the tests together concurrently.

language/testing-infra/e2e-tests/src/golden_outputs.rs
```diff
@@ -0,0 +1,47 @@
+// Copyright (c) The Libra Core Contributors
+// SPDX-License-Identifier: Apache-2.0
+
+use goldenfile::Mint;
```
runtian-zhou, line 4: @vgao1996 this is the newly added file

language/testing-infra/e2e-tests/src/lib.rs
```diff
@@ -20,6 +20,7 @@ pub mod executor;
 pub mod gas_costs;
 pub mod keygen;
 mod proptest_types;
+mod golden_outputs;
```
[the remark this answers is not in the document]
runtian-zhou: This line is needed for including that file into the build process. See comment below

language/testing-infra/e2e-tests/src/lib.rs
```diff
@@ -55,3 +56,15 @@ macro_rules! assert_prologue_disparity {
         assert!(transaction_status_eq($e3, &$e4));
     };
 }
+
+#[macro_export]
+macro_rules! function_name {
```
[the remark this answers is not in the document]
runtian-zhou: Yea grepping the name of the current function  is so far the only way I could think of to get a deterministic name for each individual tests. Rust does not support getting the name of the current running test case.

language/testing-infra/e2e-tests/src/executor.rs
vgao1996: > The main reason is to reduce the changes needed for each individual tests.
Understood and agreed. 

> You could imagine that otherwise we will need to create a seed for each tests.
No doubt about this.

Read the code again and I'm starting to understand what's causing our "disagreement" here. I feel like we were seeing `FakeExecutor` in two different ways:
- Me: it's a mock LibraExecutor/LibraVM, so it should do nothing other than executing Libra transactions
- You: it's modeling a testing flow so it should own the RNG and do other stuff like injecting accounts into the storage

I think your understanding is right considering the actual ownership model and what the methods are doing, so I'd be happy to stamp this diff. However I do realize this exposes a few other problems and we should try to get them solved:
- Rename `FakeExecutor` to something like `xxxTestingFlow` or `xxxTestingEnv`. With `Executor` being part of its name, it made me think this is more like an instance implementing the `VMExecutor` trait, mocking the real `LibraExecutor`, which is not true.
- For the immediate future, functional tests should not depend on this as it models a different testing flow. Instead it should use the fake storage and the LibraVM APIs directly. (Note: don't bother fixing the existing implementation. There's a rewriting coming in a month or two, in a style similar to #6540)

vgao1996 (approved):

runtian-zhou:
Fixed the naming for `function_name`, I didn't touch the FakeExecutor Part. IMHO a better way to implement this is to add a wrapper that wraps the FakeExecutor with Rng seed and implement `Deref<FakeExecutor>` for the wrapper. But I'll let @vgao1996  decide how we should proceed on this.

vgao1996:
@runtian-zhou Yes you could add a wrapper if you want more modularity but I'm totally fine if you don't do it. In fact, having looked at the methods I'm not sure a wrapper layer will buy us much. Functional tests and other clients could theoretically benefit from this, but that part will be rewritten anyway so I if I were you I wouldn't spend too much time on it right now.

runtian-zhou:
SlidingNonce module was modified in 0038fffeb and cause the golden file to fail. Updated the golden file to reland.

PR #6580, opened 2020-10-23
[state sync] prune outdated LIs from PendingLedgerInfo

christinacelee:
## Motivation

`PendingLedgerInfo` prunes out LIs that we can potentially set to as a target LI in chunk requests based on committed version, not synced version. This causes problems in the following FN long-poll scenario:

1. A FN sends an upstream node a chunk request with known version X and target version Y (X < Y)
2. The upstream that receives this is behind version X when it receives the FN's request, so it caches it for long-poll subscription.
3. Within the long-poll time window, the upstream reaches a version Z > X, and it fulfills the cached long-poll request with a target version of Z and a chunk of transactions X .. W, where W > Y
4. FN processes that response successfully and the resulting local synced version is W, but because of the way PendingLedgerInfos is pruned, the outdated LI Y is still set as the target LI.
5. The next chunk request the FN sends is for known_version W and target version still Y. Since W > Y, the response it gets for such a request is empty, so the FN does not make any state sync progress. 

This PR fixes this issue by making sure no target LI behind the local synced version is a candidate for a target LI (s.t. target version > known_version). 

Will land this PR first and put out a follow-up cherry-pick PR

## Test Plan

Added test to simulate above scenario, fails without fix, passes with fix

### DIFF
--- a/state-synchronizer/src/coordinator.rs
+++ b/state-synchronizer/src/coordinator.rs
@@ -122,8 +122,8 @@ impl PendingLedgerInfos {
         let highest_committed_li = sync_state.highest_local_li.ledger_info().version();
         let highest_synced = sync_state.highest_version_in_local_storage();
 
-        // prune any pending LIs that were successfully committed
-        self.pending_li_queue = self.pending_li_queue.split_off(&(highest_committed_li + 1));
+        // prune any pending LIs that are older than the latest local synced version
+        self.pending_li_queue = self.pending_li_queue.split_off(&(highest_synced + 1));
 
         // pick target LI to use for sending ProgressiveTargetType requests.
         self.target_li = if highest_committed_li == highest_synced {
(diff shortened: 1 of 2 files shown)


zekun000:
is it able to commit the LI with empty chunks?

christinacelee:
LI only gets committed if all txns leading up to it are synced, so would expect no

christinacelee:
or you're saying if the txns are already there and we try to commit the LI afterwards separately? not sure

zekun000:
yeah, the empty chunk sent back by a honest validator means the txns are already there and should be able to commit the LI with empty chunks, not sure if the API supports it

davidiw:
> yeah, the empty chunk sent back by a honest validator means the txns are already there and should be able to commit the LI with empty chunks, not sure if the API supports it

It might be worth trying because I don't see anything in the executor that would prevent this, but there are a lot of proofs that would have to verify correctly even on an empty proof.

Though I am okay even if we just move forward. I do like the attempt to optimize.

christinacelee:
I'd prefer we tackle that issue in a separate future PR - for now keep the scope of this PR as is

davidiw (approved):
Awesome

PR #6583, opened 2020-10-23
[cherry-pick][state sync] verify waypoint's end-of-epoch li matches end-of-chunk v…

christinacelee:
### Justification
This cherry-pick fixes a state-sync bug encountered in premainnet - state sync getting stuck during the waypoint sync phase. These are triggered by state sync not performing certain checks on an end-of-epoch-li before passing it to execution/commit. This change affects Libra nodes - they will have to restart with the updated software.

#### Technical details

In the waypoint sync protocol, when a node processes the waypoint chunk response, it passes a chunk of transactions and an optional end-of-epoch LI to the executor for execution/commit. The executor expects that if it is given an end-of-epoch LI, it meets various requirements (the chunk of txns passed in with it ends at the epoch, the LI ends in epoch) - otherwise, execution/commit will fail, but state sync wasn't performing these checks of the end-of-epoch LI, so state sync was stuck. This PR adds those verifications fo the end-of-epoch LI. 

### Testing
- Added a integration and smoke test simulating the above bug scenario (waypoint sync across multiple epochs that span multiple chunks), both fail without the fix, both pass with the fix
- No regression detected

### Workarounds
- If a node gets stuck due to this unsolved problem, the mitigation is to restart the node by setting its waypoint to an earlier one

- Not fixing this will put severe limitations on the range of waypoints a node can use and lose the benefits a waypoint provides.

### DIFF
--- a/state-synchronizer/src/coordinator.rs
+++ b/state-synchronizer/src/coordinator.rs
@@ -731,6 +731,13 @@ impl<T: ExecutorProxyTrait> SyncCoordinator<T> {
             None
         };
         if let Some(li) = end_of_epoch_li.as_ref() {
+            ensure!(
+                li.ledger_info().version() >= request.known_version,
+                "waypoint request's current_epoch (epoch {}, version {}) < waypoint request's known_version {}",
+                li.ledger_info().epoch(),
+                li.ledger_info().version(),
+                request.known_version,
+            );
             let num_txns_until_end_of_epoch = li.ledger_info().version() - request.known_version;
             limit = std::cmp::min(limit, num_txns_until_end_of_epoch);
         }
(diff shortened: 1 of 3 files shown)


davidiw:
A few addendums:
* The underlying issue is that the executor has different and inconsistent checks for a target li and end of epoch li. StateSync was unaware of this assumption and hence when receiving a chunk that didn't end an epoch but came with an end of epoch li, the executor would return an error.
* On the note of workarounds, finding an appropriate waypoint is a non-trivial task even for experienced folks in the space, let alone operators that have "broken" systems.
* This issue is further a pain if a validator node swaps between operators or the operator loses access to their historical consensus keys, which can and will happen. The proposed solution was to state sync them via a waypoint into a state where their keys could be reconciled (onchain to storage), but right now there is no clean solution here except a very ambiguous backup and restore procedure that has failed more times than it has succeeded.

davidiw (approved):
/land

PR #6584, opened 2020-10-23
[fuzzing] split fuzzer scripts in multiple scripts for usability, also remove dead-link flag

mimoo:


sausagee (commented):
The change looks fine. 

Is the intent to decouple the build/fuzz scripts so that we don't need to open a PR in oss-fuzz for changes we make here?

sausagee (approved):

mimoo:
> Is the intent to decouple the build/fuzz scripts so that we don't need to open a PR in oss-fuzz for changes we make here?

no it's for us (lionhead wanted that, but it's also useful if we want to run tools (like Instruments) on a fuzzer

PR #6586, opened 2020-10-23
Even if docker hub signing is down, let's push images to dockerhub if possible, and if not, just novi ecr.

rexhoffman:
## Motivation

During a hot fix release dockerhub signing was down, revealing a limitation in the flow of build and pushing docker images.

Yes

## Test Plan

CI

## Related PRs

This will get cherry picked in to master as well.


mgorven (approved):

.circleci/config.yml
```diff
@@ -774,13 +774,15 @@ jobs:
...
-            docker/build_push.sh -u -p -b ${BRANCH} -n validator-tcb
-            docker/build_push.sh -u -p -b ${BRANCH} -n cluster-test
+            success=0
+            docker/build_push.sh -u -p -b ${BRANCH} -n client || success=-1
+            docker/build_push.sh -u -p -b ${BRANCH} -n init || success=-1
+            docker/build_push.sh -u -p -b ${BRANCH} -n faucet || success=-1
+            docker/build_push.sh -u -p -b ${BRANCH} -n tools || success=-1
+            docker/build_push.sh -u -p -b ${BRANCH} -n validator || success=-1
+            docker/build_push.sh -u -p -b ${BRANCH} -n validator-tcb || success=-1
+            docker/build_push.sh -u -p -b ${BRANCH} -n cluster-test || success=-1
```
davidiw, line 784: shouldn't we evaluate success before performing these operations?

.circleci/config.yml
```diff
@@ -797,19 +799,22 @@ jobs:
...
-            docker/build_push.sh -u -b ${BRANCH} -n validator-tcb
-            docker/build_push.sh -u -b ${BRANCH} -n cluster-test
+            success=0
+            docker/build_push.sh -u -b ${BRANCH} -n client || success=-1
+            docker/build_push.sh -u -b ${BRANCH} -n init || success=-1
+            docker/build_push.sh -u -b ${BRANCH} -n faucet || success=-1
+            docker/build_push.sh -u -b ${BRANCH} -n tools || success=-1
+            docker/build_push.sh -u -b ${BRANCH} -n validator || success=-1
+            docker/build_push.sh -u -b ${BRANCH} -n validator-tcb || success=-1
+            docker/build_push.sh -u -b ${BRANCH} -n cluster-test || success=-1
```
davidiw, line 809: same here

.circleci/config.yml
```diff
@@ -797,19 +799,22 @@ jobs:
...
+            exit $success
       - run:
           name: docker image pruning.
           command: |
             scripts/dockerhub_prune.sh -u "${DOCKERHUB_USERNAME}" -p "${DOCKERHUB_PASSWORD}" -x
       - run:
           name: push to novi ecr
+          when: always
```
davidiw, line 817: consider this says always... why do we need the rest?

.circleci/config.yml
mgorven: Image signing is failing in this script, but it can still push an unsigned image. The point is to attempt all images so we at least get unsigned images rather than no images.

.circleci/config.yml
mgorven: We want to always attempt the push to Novi ECR; it might fail (if say the Rust build is failing), but if the issue is just with Dockerhub it at least gets us images in ECR.

davidiw: Comment was more that the PR title and code ... aren't quite aligned... I agree on the approach taken.

davidiw (approved):

.circleci/config.yml
rexhoffman: if the command fails, we set success to -1 so it can be reported at the end of all the steps in the flow.   Several things may have gone wrong.   maybe the image couldn't be built, maybe it couldn't be signed (the problem this worked around) and maybe it couldn't be pushed to dockerhub.   Regardless if they could be built we want to continue and build them all (in the current case they could all be built and pushed to dockerhub but docker signing appears to be down down/unreachable).    Errors will still be reported at the end of the build to allow for investigation.

.circleci/config.yml
rexhoffman: the prior steps build the images, this merely pushes them to novi aws ecr.

PR #6588, opened 2020-10-24
[Safety Rules] Avoid exporting the consensus private key

JoshLind:
## Motivation

This PR updates Libra Safety Rules (LSR) to support two modes of operation: (i) exporting the consensus key from vault on each epoch change (which is the default today); and (ii) delegating the signing of data to vault directly, so that LSR doesn't need to export the key (the new mode of operation). The choice of which operational mode is defined by a new boolean in SafetyRulesConfig, "export_consensus_key", which this PR defaults to false.

There are several caveats to this change:
1. If vault is expected to do the signing, on each epoch change, to be sure vault actually holds the consensus private key for the current key version, we sign a piece of test data. If this fails, we immediately know that the key doesn't exist. This prevents us from having to wait until our first sign attempt to realize we don't have the key.
2. There will be a slight performance hit from delegating signatures to vault. We'll need to confirm this exact number in canary/cluster test (old results showed 5% - 15%, but we've since optimized the storage path, this may change).

This PR offers 3 commits to achieve delegated consensus key signatures:
1. First, we replace the ValidatorSigner in LSR with a ConfigurableValidatorSigner and add a new configuration option to SafetyRulesConfig.
2. Next, we update the existing unit tests to support the different operational modes.
3. Finally, we add two new smoke tests to ensure that both operational modes continue to work.

Yes.

## Test Plan

All tests pass.

## Related PRs

This PR is based on an old proof of concept PR that we previously closed out (https://github.com/libra/libra/pull/6004)

### DIFF
--- a/config/src/config/safety_rules_config.rs
+++ b/config/src/config/safety_rules_config.rs
@@ -23,6 +23,7 @@ pub struct SafetyRulesConfig {
     pub service: SafetyRulesService,
     pub test: Option<SafetyRulesTestConfig>,
     pub verify_vote_proposal_signature: bool,
+    pub export_consensus_key: bool,
     // Read/Write/Connect networking operation timeout in milliseconds.
     pub network_timeout_ms: u64,
     pub enable_cached_safety_data: bool,
@@ -36,7 +37,8 @@ impl Default for SafetyRulesConfig {
             service: SafetyRulesService::Thread,
             test: None,
             verify_vote_proposal_signature: true,
-            // Default value of 30seconds for a timeout
+            export_consensus_key: false,
+            // Default value of 30 seconds for a timeout
             network_timeout_ms: 30_000,
             enable_cached_safety_data: true,
         }
--- a/consensus/safety-rules/src/configurable_validator_signer.rs
+++ b/consensus/safety-rules/src/configurable_validator_signer.rs
@@ -1,0 +1,99 @@
+// Copyright (c) The Libra Core Contributors
+// SPDX-License-Identifier: Apache-2.0
+
+use crate::{Error, PersistentSafetyStorage};
+use libra_crypto::{
+    ed25519::{Ed25519PrivateKey, Ed25519PublicKey, Ed25519Signature},
+    hash::CryptoHash,
+};
+use libra_global_constants::CONSENSUS_KEY;
+use libra_types::{account_address::AccountAddress, validator_signer::ValidatorSigner};
+use serde::Serialize;
+
+/// A ConfigurableValidatorSigner is a ValidatorSigner wrapper that offers either
+/// a ValidatorSigner instance or a ValidatorHandle instance, depending on the
+/// configuration chosen. This abstracts away the complexities of handling either
+/// instance, while offering the same API as a ValidatorSigner.
+pub enum ConfigurableValidatorSigner {
+    Signer(ValidatorSigner),
+    Handle(ValidatorHandle),
+}
+
+impl ConfigurableValidatorSigner {
+    /// Returns a new ValidatorSigner instance
+    pub fn new_signer(author: AccountAddress, consensus_key: Ed25519PrivateKey) -> Self {
+        let signer = ValidatorSigner::new(author, consensus_key);
+        ConfigurableValidatorSigner::Signer(signer)
+    }
+
+    /// Returns a new ValidatorHandle instance
+    pub fn new_handle(author: AccountAddress, key_version: Ed25519PublicKey) -> Self {
+        let handle = ValidatorHandle::new(author, key_version);
+        ConfigurableValidatorSigner::Handle(handle)
+    }
+
+    /// Returns the author associated with the signer configuration.
+    pub fn author(&self) -> AccountAddress {
+        match self {
+            ConfigurableValidatorSigner::Signer(signer) => signer.author(),
+            ConfigurableValidatorSigner::Handle(handle) => handle.author(),
+        }
+    }
+
+    /// Returns the public key associated with the signer configuration.
+    pub fn public_key(&self) -> Ed25519PublicKey {
+        match self {
+            ConfigurableValidatorSigner::Signer(signer) => signer.public_key(),
+            ConfigurableValidatorSigner::Handle(handle) => handle.key_version(),
+        }
+    }
+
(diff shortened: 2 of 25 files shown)


davidiw:
Is this ready for review? Not clear looking at the latest canary.

JoshLind:
> Is this ready for review? Not clear looking at the latest canary.

Technically this is ready for review (so if you have early feedback, go for it). However, it's failing on CT because we've renamed the consensus key in vault (so I still need to look into either updating CT to avoid looking for the old key, or find a meaningful work around where we can support both keys).

consensus/safety-rules/src/safety_rules.rs
```diff
@@ -249,26 +260,31 @@ impl SafetyRules {
...
-                        );
-
-                        self.validator_signer = None;
-                        error
-                    })?;
-
-                self.validator_signer = Some(ValidatorSigner::new(author, consensus_key));
+                let validator_handle = Some(ValidatorHandle::new(
+                    author,
+                    CONSENSUS_KEY.into(),
```
davidiw, line 267: this is a const, why are we introducing it as a field?

config/global-constants/src/lib.rs
```diff
@@ -9,7 +9,7 @@
 #![forbid(unsafe_code)]
 
 /// Definitions of global cryptographic keys (e.g., as held in secure storage)
-pub const CONSENSUS_KEY: &str = "consensus";
+pub const CONSENSUS_KEY: &str = "consensus_v2";
```
davidiw, line 12: why are we introducing a new key?

davidiw (changes requested):
At this point, I don't think we should be dictating behavioral changes in deployment. We can make recommendations and provide configuration to do so. We can then work with our partners to roll out improved security but everything should be incremental and reversible. TL;DR, keep names the same and rely on a config to dictate which behavior is desirable.

config/global-constants/src/lib.rs
mgorven: Let's handle this as a separate migration.

consensus/safety-rules/src/safety_rules.rs
JoshLind: Fair enough -- if we're always assuming a validator is identified by the CONSENSUS_KEY, I'll merge this into the ValidatorHandle 😄

config/global-constants/src/lib.rs
JoshLind: > why are we introducing a new key?

@davidiw, as discussed offline, this was because vault doesn't allow a key to be non-exportable after previously being exportable (see the PR description).

> Let's handle this as a separate migration.

Okay, thanks @mgorven, sounds good, I'll update the PR.

JoshLind:
Thanks @davidiw!

> TL;DR, keep names the same and rely on a config to dictate which behavior is desirable.

As discussed offline, I'll update the PR to adopt a configurable approach. This not only makes sense for this controlled roll-outs and fail safes, but it makes sense for the TCB design going forward. I'll update the PR when it's ready for another look.

JoshLind:
@davidiw, this PR is ready for another look, I've finally updated it to be configurable (although, we now configure this to be the default). Hopefully it's not too bad -- although the "configuration" aspect of it definitely introduced a lot more code 😕

consensus/safety-rules/src/configurable_validator_signer.rs
```diff
@@ -0,0 +1,101 @@
+// Copyright (c) The Libra Core Contributors
+// SPDX-License-Identifier: Apache-2.0
+
+#![forbid(unsafe_code)]
```
davidiw, line 4: this isn't required unless it isn't in main.rs and lib.rs files.

consensus/safety-rules/src/safety_rules.rs
```diff
@@ -166,7 +179,7 @@ impl SafetyRules {
     }
 
     /// This verifies whether the author of one proposal is the validator signer
-    fn verify_author(&self, author: Option<Author>) -> Result<(), Error> {
+    fn verify_author(&mut self, author: Option<Author>) -> Result<(), Error> {
```
davidiw, line 182: Why is this mutable? I think we should fish out whether or not the crypto APIs should be mutable for signing rather than polluting this. If, for whatever reason, they must be, we should hide that because in general signing shouldn't be mutable. Hiding via -- rwlock -- for example.

davidiw (commented):
overall looks good, but before we pull in "tech debt" of mut, can we try to improve the hygiene of the code base and try to eliminate it?

consensus/safety-rules/src/configurable_validator_signer.rs
JoshLind: Ooh, thanks! Copy pasta 😄

consensus/safety-rules/src/safety_rules.rs
JoshLind: This is not strictly required (must have been legacy?), so it's been fixed here: https://github.com/libra/libra/pull/6762. I've rebased the PR and it looks better 😄

JoshLind:
Thanks, @davidiw! 😄 

> overall looks good, but before we pull in "tech debt" of mut, can we try to improve the hygiene of the code base and try to eliminate it?

Done, hopefully it looks a little cleaner?

config/src/config/safety_rules_config.rs
```diff
@@ -36,7 +37,8 @@ impl Default for SafetyRulesConfig {
             service: SafetyRulesService::Thread,
             test: None,
             verify_vote_proposal_signature: true,
-            // Default value of 30seconds for a timeout
+            export_consensus_key: false,
```
davidiw, line 40: hmm....

@mgorven, is this good for you?

davidiw (approved):
wait for Michael's response and let's also ensure that perf isn't affected too much... if it is, you'll likely need to bother with that side of things :/

config/src/config/safety_rules_config.rs
mgorven: Defaulting to remote signing? Fine with me!

JoshLind: Sounds good! I'll push it :smile

PR #6593, opened 2020-10-26
Update README.md

davidiw:
this tool doesn't support adding new validators to an existing blockchain that's handled by the operational tool


JoshLind (approved):
LGTM! Thanks @davidiw 
/land

PR #6595, opened 2020-10-26
[consensus] more granular logging for votes

mimoo:
see https://github.com/libra/libra/pull/6576

we want to use this to detect equivocating votes

### DIFF
--- a/consensus/src/round_manager.rs
+++ b/consensus/src/round_manager.rs
@@ -626,11 +626,17 @@ impl RoundManager {
     /// 2) call process_certificates(), which will start a new round in return.
     async fn process_vote(&mut self, vote: &Vote) -> anyhow::Result<()> {
         let round = vote.vote_data().proposed().round();
-        debug!(
+
+        info!(
             self.new_log(LogEvent::ReceiveVote)
                 .remote_peer(vote.author()),
-            "{}", vote
+            vote = %vote,
+            vote_epoch = vote.vote_data().proposed().epoch(),
+            vote_round = vote.vote_data().proposed().round(),
+            vote_id = vote.vote_data().proposed().id(),
+            vote_state = vote.vote_data().proposed().executed_state_id(),
         );
+
         if !vote.is_timeout() {
             // Unlike timeout votes regular votes are sent to the leaders of the next round only.
             let next_round = round + 1;


consensus/src/round_manager.rs
```diff
@@ -621,11 +621,12 @@ impl RoundManager {
     /// 2) call process_certificates(), which will start a new round in return.
     async fn process_vote(&mut self, vote: &Vote) -> anyhow::Result<()> {
         let round = vote.vote_data().proposed().round();
-        debug!(
+        info!(
             self.new_log(LogEvent::ReceiveVote)
                 .remote_peer(vote.author()),
-            "{}", vote
+            vote_data = vote,
```
mimoo, line 627: no variable-length field here, we should expect as many such logs as the received proposal log here: https://github.com/libra/libra/pull/6576 (so N per round, for N validators) **unless we're timing out**. In rounds that timeout, that event should be observed N times more (so NxN per round, for N validators).

gregnazario: How big is the `Vote` object?  I'm pretty sure the `LedgerInfo` part of it is pretty big?
https://github.com/libra/libra/blob/master/consensus/consensus-types/src/vote.rs#L18-L30

gregnazario: So, at minimum it's (N^2 + N) * the size of the block?

mimoo: Screen Shot 2020-10-27 at 12 02 09 PM

mimoo: A `Vote` is basically 3 `BlockInfo`s and some other fields. 
A `BlockInfo` is ~100 bytes if there is no reconfiguration, Nx32 bytes more if there's a reconfiguration.
So that's around 400 bytes per `Vote` when no reconfiguration is occurring, and much more if there's a reconfiguration.

We can take the same strategy that I'm proposing with the ReceiveProposal and only log the signa (epoch, round, id, executed_state_id, author).l, but not the proof, which would bring that down to ~100 bytes

gregnazario: Yeah, just trying to keep it under ~1024 bytes if possible, if we're constantly logging it at this rate.  Over optimizing to less than 100 bytes is not too important

gregnazario: So the current log line, logs any where from 30-300 in a second, averaging usually at the 60 mark [here](https://85bd0f3d76314c71869bce3ae260f6d2.us-west1.gcp.cloud.es.io:9243/app/discover#/?_g=(filters:!(),refreshInterval:(pause:!t,value:0),time:(from:'2020-10-29T20:15:54.357Z',to:'2020-10-29T20:23:01.365Z'))&_a=(columns:!(_source),filters:!(('$state':(store:appState),meta:(alias:!n,disabled:!f,index:'44a07b90-f794-11ea-af4e-cf9e04c7cc93',key:location.keyword,negate:!f,params:(query:'consensus%2Fsrc%2Fround_manager.rs:624'),type:phrase),query:(match_phrase:(location.keyword:'consensus%2Fsrc%2Fround_manager.rs:624')))),index:'44a07b90-f794-11ea-af4e-cf9e04c7cc93',interval:s,query:(language:kuery,query:''),sort:!()))

mimoo: for 30 validators I predicted 90/s without timeout and 900/round with timeout (not sure how long a timeout is), so that's about right! Forgot we were already logging these in debug :)

mimoo: so I realize that the log already truncates a number of fields via the use of custom Display implementation, so to avoid losing any data I instead add some new data in the schema and keep what we've been logging so far.

consensus/src/round_manager.rs
```diff
@@ -621,11 +621,17 @@ impl RoundManager {
     /// 2) call process_certificates(), which will start a new round in return.
     async fn process_vote(&mut self, vote: &Vote) -> anyhow::Result<()> {
         let round = vote.vote_data().proposed().round();
-        debug!(
+
+        info!(
             self.new_log(LogEvent::ReceiveVote)
                 .remote_peer(vote.author()),
-            "{}", vote
+            vote = %vote,
```
gregnazario, line 628: nice, good to know this was helpful :)

consensus/src/round_manager.rs
```diff
@@ -621,11 +621,17 @@ impl RoundManager {
...
+            vote_epoch = vote.vote_data().proposed().epoch(),
+            vote_round = vote.vote_data().proposed().round(),
+            vote_id = vote.vote_data().proposed().id(),
+            vote_state = vote.vote_data().proposed().executed_state_id(),
```
gregnazario, line 632: So these are just so you can query by them?  Otherwise it's in the `vote` field already right?

gregnazario (approved):

consensus/src/round_manager.rs
mimoo: yep! although some of these fields are truncated in the `vote` field (vote_id and vote_state) and here we don't truncate them (on purpose)

PR #6594, opened 2020-10-26
[json-rpc] get_metadata only return move version information when requesting latest version

xli:
## Motivation

get_metadata may give an earlier version, and the node may not have the account state data cached anymore, hence we got the following error in premainnet:

```
 curl -X POST https://fn.premainnet.aosdev.libra.org -H "Content-Type: application/json" --data '{"jsonrpc":"2.0","method":"get_metadata","params":[4],"id":1}' | jq
{
  "libra_chain_id": 19,
  "libra_ledger_version": 65734695,
  "libra_ledger_timestampusec": 1603745486824887,
  "jsonrpc": "2.0",
  "id": 1,
  "error": {
    "code": -32000,
    "message": "Server error: Missing node at NodeKey { version: 4, nibble_path:  }.",
    "data": null
  }
}
```

This PR changes to only return version info when requesting latest version, as these data mostly needed for latest version instead of old versions.

### Have you read the [Contributing Guidelines on pull requests](https://github.com/libra/libra/blob/master/CONTRIBUTING.md#pull-requests)?

(Write your answer here.)

## Test Plan

Integration test

## Related PRs

(If this PR adds or changes functionality, please take some time to update the docs at https://github.com/libra/libra/developers.libra.org, and link to your PR here.)

PR #6600, opened 2020-10-26
cherry-pick [json-rpc] get_metadata only return move version info when requesting latest version

xli:
## Motivation

JSON-RPC get_metadata method responses server error when client request an old version that storage layer dropped the account state for the version.
This PR fixes the error by only return libra version and move script hashing information only when client request latest ledger version.

Why we must have it for V1 launch: without this fix, get_metadata is likely can only get recent transaction metadata, it seems lots of people use it to get transaction timestamp when got a transaction version somewhere that has no timestamp with it.

What workarounds and alternative we have if we do not push the PR: call get_transactions for the transaction timestamp by version.

## Related PRs

#6594

### DIFF
--- a/json-rpc/docs/method_get_metadata.md
+++ b/json-rpc/docs/method_get_metadata.md
@@ -15,6 +15,8 @@ Get the blockchain / ledger metadata.
 
 [Metadata](type_metadata.md)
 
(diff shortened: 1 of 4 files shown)


libra-action:

Breaking changes policy:

2. Adding fields is ok.

thefallentree (approved):

sausagee (approved):

PR #6601, opened 2020-10-27
[network][fuzzing] Fuzz inbound `NetworkMessage` handling and fix memory leak

phlip9:
+ Add fuzzing for the `Peer` actor's inbound message handling. For each fuzzer iteration, we spin up a new `Peer` actor and pipe the raw fuzzer data into it. This mostly tests that the `Peer` inbound message handling doesn't panic or leak memory when reading, deserializing, and handling messages from remote peers.

+ Fix stream-ratelimiter memory leak. From the network fuzzing we discovered a small memory leak (~8 bytes / inbound libranet message) in the stream-ratelimiter crate which was incorrectly using the futures-semaphore crate.

+ The root problem was using `ManuallyDrop` on the `futures-semaphore` permits as a work-around for forgetting a permit. Of course, no handles to the permits were stored and so the permits were never dropped, which would leak an `Arc` (1 machine word) every time a permit was acquired (once per inbound network message).

+ Remove futures-semaphore wrapper. Previously, tokio's semaphore implementation had some limitations that pushed us to add a small wrapper around it, namely `common/futures-semaphore`. Today, tokio's semaphore has finally added all of the functionality we were missing so it's no longer necessary to maintain this wrapper.


network/src/lib.rs
```diff
@@ -10,22 +10,18 @@ pub use interface::NetworkProvider;
 
 pub mod connectivity_manager;
 pub mod constants;
+pub mod counters;
 pub mod error;
 pub mod interface;
 pub mod logging;
+pub mod noise;
```
mimoo, line 17: is this a pattern we want to avoid? By having it `pub` we signal that this is something that other crates here can use, whereas this is not true (besides fuzzing, but this is part of testing)

network/src/peer/mod.rs
```diff
@@ -140,7 +153,7 @@ where
 
     pub async fn start(mut self) {
         let remote_peer_id = self.remote_peer_id();
-        info!(
+        trace!(
```
gregnazario, line 156: I can see value in changing info to trace, but I'm wondering the logic behind changing the info to trace?  Do we have a large number of logs coming from our testing?

Also, we should put this into a separate commit around improving the logging.

network/src/peer/mod.rs
```diff
@@ -78,6 +81,16 @@ enum State {
...
+/// Returns a fully configured length-delimited codec for writing and reading
+/// [`NetworkMessage`]es to a socket.
+pub fn message_codec(max_frame_size: usize) -> LengthDelimitedCodec {
+    LengthDelimitedCodec::builder()
+        .max_frame_length(max_frame_size)
+        .length_field_length(4)
+        .big_endian()
+        .new_codec()
+}
+
```
gregnazario, line 93: I actually do like putting things in separate functions.  Though, this is another refactor that should be in a separate commit.  

After I scrolled down some, I see you're trying to make use of this function in tests, probably to make a more realistic test.  Let's make it another commit around improving the unit tests.

network/src/peer/mod.rs
```diff
@@ -190,6 +198,13 @@ where
                             }
                         },
                         maybe_message = reader.next() => {
+                            trace!(
+                                NetworkSchema::new(&self.network_context).connection_metadata(&self.connection_metadata),
+                                "{} maybe_message: {:?}",
+                                self.network_context,
+                                maybe_message,
+                            );
```
gregnazario, line 206: Might be helpful to say, `{} Peer received maybe_message: {:?}`

network/src/peer/mod.rs
```diff
@@ -492,13 +507,11 @@ where
...
             }
-            PeerRequest::CloseConnection => {
-                self.close_connection(DisconnectReason::Requested).await;
-            }
+            PeerRequest::CloseConnection => self.close_connection(DisconnectReason::Requested),
         }
     }
 
-    async fn close_connection(&mut self, reason: DisconnectReason) {
+    fn close_connection(&mut self, reason: DisconnectReason) {
```
gregnazario, line 514: Haha, an async function that didn't have to be.  I think though, this is another one of those refactors that could be in a separate commit, because it doesn't necessarily have to do with fuzzing.

network/src/testutils/fake_socket.rs
```diff
@@ -40,13 +40,11 @@ impl<'a> ReadOnlyTestSocket<'a> {
     }
 
     /// reads will have to be done byte by byte
-    #[allow(dead_code)]
```
gregnazario, line 43: Woo less dead code

testsuite/libra-fuzzer/README.md
```diff
@@ -115,10 +115,42 @@ fuzz/google-oss-fuzz/build_fuzzer.sh ConsensusProposal .
 
 ### Troubleshooting
 
-#### linking with CC failed
+#### My backtrace does not contain file names and line numbers
 
-Are you on MacOS? Have you checked that Xcode is up-to-date?
+You need to use `llvm-symbolizer`, see https://github.com/rust-fuzz/cargo-fuzz/issues/160
```
gregnazario, line 120: Thanks for updating the README!

common/stream-ratelimiter/src/lib.rs
```diff
@@ -62,7 +62,7 @@ impl<T: Stream> RateLimiter<T> {
             inner: stream,
             interval: interval(refill_interval).fuse(),
             // Start with 0 permits because tokio::interval ticks immediately on setup.
-            permits: Semaphore::new(0),
+            permits: Arc::new(Semaphore::new(0)),
```
gregnazario, line 65: Does the first fuzzing change depend on this commit to succeed?  If so, let's put it as the first commit to fix the bug :)

common/stream-ratelimiter/src/lib.rs
```diff
@@ -76,23 +76,27 @@ impl<T: Stream> Stream for RateLimiter<T> {
     fn poll_next(mut self: Pin<&mut Self>, cx: &mut task::Context<'_>) -> Poll<Option<Self::Item>> {
         // If the interval is ready, add permits. This is safe to do in all calls to poll_next().
         if let Poll::Ready(Some(_)) = self.as_mut().project().interval.poll_next(cx) {
-            self.permits
-                .add_permits(self.capacity - self.permits.available_permits());
+            let num_new_permits = self
+                .capacity
+                .saturating_sub(self.permits.available_permits());
+            self.permits.add_permits(num_new_permits);
```
gregnazario, line 82: Okay, so this always guarantees that the `number of permits < capacity`.  Since, this is the only place we add permits, and without locking the semaphore, the number of permits is always less than or equal to the observed value.

common/bounded-executor/src/lib.rs
```diff
@@ -8,20 +8,24 @@
...
+    semaphore: Arc<Semaphore>,
     executor: Handle,
 }
 
 impl BoundedExecutor {
     /// Create a new `BoundedExecutor` from an existing tokio [`Handle`]
     /// with a maximum concurrent task capacity of `capacity`.
     pub fn new(capacity: usize, executor: Handle) -> Self {
-        let semaphore = Semaphore::new(capacity);
+        let semaphore = Arc::new(Semaphore::new(capacity));
```
gregnazario, line 28: Is this remove `futures-semaphore` something found in the fuzzing, or just for simplification?

network/src/peer/fuzzing.rs
```diff
@@ -126,12 +125,12 @@ pub fn fuzz(data: &[u8]) {
         // Wait for "remote" to disconnect (we read all data and socket read
         // returns EOF), we read a disconnect request, or we fail to deserialize
         // something.
-        let connection_notifs = connection_notifs_rx.collect::<Vec<_>>().await;
+        let _connection_notifs = connection_notifs_rx.collect::<Vec<_>>().await;
```
gregnazario: Minor, but do we need `connection_notifs` to be dropped at the end of the scope or, immediately.  If it doesn't have to be dropped at the end of the scope, we could just replace it with `_`

network/src/peer/fuzzing.rs
```diff
@@ -0,0 +1,150 @@
...
+    // generate unbounded metrics otherwise and OOM during fuzzing.
+    let peer_id = PeerId::ZERO;
+    // However, we want to choose a random _remote_ peer id to ensure we _don't_
+    // have metrics logging the remote peer id (which would eventually OOM in
+    // production for public-facing nodes).
+    let remote_peer_id = PeerId::random();
+
+    // Mock data
+    let network_context = NetworkContext::mock_with_peer_id(peer_id);
+    let socket = ReadOnlyTestSocketVec::new(data.to_vec());
```
mimoo, line 86: glad that's useful :D can't remember why there is a `Vec` at the end of the name though

network/src/peer/fuzzing.rs
```diff
@@ -0,0 +1,150 @@
...
+    let metadata = ConnectionMetadata::new(
+        remote_peer_id,
+        ConnectionId::from(123),
+        NetworkAddress::mock(),
+        ConnectionOrigin::Inbound,
```
mimoo, line 91: will this impact the handling in any ways?

network/src/peer/fuzzing.rs
```diff
@@ -0,0 +1,150 @@
...
+        MessagingProtocolVersion::V1,
+        SupportedProtocols::from(
+            [
+                ProtocolId::ConsensusRpc,
+                ProtocolId::ConsensusDirectSend,
+                ProtocolId::MempoolDirectSend,
+                ProtocolId::StateSynchronizerDirectSend,
+                ProtocolId::DiscoveryDirectSend,
+                ProtocolId::HealthCheckerRpc,
```
mimoo, line 100: sucks that we can't just add all enums easily, looks like we'd have to use a macro for that (https://docs.rs/strum_macros/0.19.4/strum_macros/derive.EnumIter.html)

network/src/peer/fuzzing.rs
```diff
@@ -0,0 +1,150 @@
...
+            ]
+            .iter(),
+        ),
+    );
+    let connection = Connection { socket, metadata };
+
+    let (connection_notifs_tx, connection_notifs_rx) = channel::new_test(8);
+    let max_concurrent_reqs = 8;
+    let max_concurrent_notifs = 8;
+    let channel_size = 8;
```
mimoo, line 110: what's the implication of 8 here : o

network/src/peer/fuzzing.rs
```diff
@@ -0,0 +1,150 @@
...
+pub fn generate_corpus(gen: &mut ValueGenerator) -> Vec<u8> {
+    let network_msgs = gen.generate(vec(any::<NetworkMessage>(), 1..20));
+
+    let network_msgs = network_msgs
+        .into_iter()
+        .map(|msg| lcs::to_bytes(&msg).unwrap())
+        .collect::<Vec<_>>();
+
+    let (write_socket, mut read_socket) = MemorySocket::new_pair();
+    let mut writer = FramedWrite::new(IoCompat::new(write_socket), message_codec(MAX_FRAME_SIZE));
```
mimoo, line 38: how will this change the data? add 2-byte length prefix?

network/src/peer/fuzzing.rs
```diff
@@ -0,0 +1,150 @@
...
+        .unwrap();
+    let executor = rt.handle().clone();
+
+    // We want to choose a constant peer id for _our_ peer id, since we will
+    // generate unbounded metrics otherwise and OOM during fuzzing.
+    let peer_id = PeerId::ZERO;
+    // However, we want to choose a random _remote_ peer id to ensure we _don't_
+    // have metrics logging the remote peer id (which would eventually OOM in
+    // production for public-facing nodes).
+    let remote_peer_id = PeerId::random();
```
mimoo, line 82: I'm not sure I understand why you want a random remote_peer_id here, and I would also suggest against it since this might trip the fuzzer (we want to avoid any non-determinism)

network/src/peer/fuzzing.rs
```diff
@@ -0,0 +1,150 @@
...
+                ProtocolId::StateSynchronizerDirectSend,
+                ProtocolId::DiscoveryDirectSend,
+                ProtocolId::HealthCheckerRpc,
+            ]
+            .iter(),
+        ),
+    );
+    let connection = Connection { socket, metadata };
+
+    let (connection_notifs_tx, connection_notifs_rx) = channel::new_test(8);
```
mimoo, line 107: I'm confused as to what this channel is here, where is it imported?

network/src/peer/fuzzing.rs
```diff
@@ -0,0 +1,150 @@
...
+        // Wait for "remote" to disconnect (we read all data and socket read
+        // returns EOF), we read a disconnect request, or we fail to deserialize
+        // something.
+        let connection_notifs = connection_notifs_rx.collect::<Vec<_>>().await;
+
+        // ACK the "remote" d/c and drop our handle to the Peer actor. Then wait
+        // for all network notifs to drain out and finish.
+        drop(network_reqs_tx);
+        let network_notifs = network_notifs_rx.collect::<Vec<_>>().await;
+    });
```
mimoo, line 134: so two things here:

* If more than 8 messages are sent to any channel, might it block the fuzzer?
* I wish we could create a fake channel that just does nothing on send, this would be useful in other places too : o

network/src/peer/fuzzing.rs
```diff
@@ -0,0 +1,150 @@
...
+
+    // All tasks should be complete. The 60s timer here will at least save this
+    // fuzzer run as a "slow" run if we leak a task. Unfortunately there doesn't
+    // seem to be a way to Err/panic if there are leaked tasks...
+    rt.shutdown_timeout(Duration::from_secs(60));
```
mimoo, line 139: this prevents us from detecting real timeouts : o what do you mean by leak a task?

network/src/peer/fuzzing.rs
gregnazario: This should be the `libra-channel`

network/src/peer/fuzzing.rs
```diff
@@ -0,0 +1,149 @@
...
+            ]
+            .iter(),
+        ),
+    );
+    let connection = Connection { socket, metadata };
+
+    let (connection_notifs_tx, connection_notifs_rx) = channel::new_test(8);
+    let max_concurrent_reqs = 8;
+    let max_concurrent_notifs = 8;
+    let channel_size = 8;
```
gregnazario, line 110: Maybe just a `channel_size = 8` and reuse it like 4 times?

network/src/peer/fuzzing.rs
mimoo: also, why would the peer disconnect here? I feel like this might run forever

network/src/peer/mod.rs
phlip9: ahh, I wanted to extract this so I could use it in the fuzzer since it needs to serialize NetworkMessage's into wire format

common/stream-ratelimiter/src/lib.rs
phlip9: it doesn't fail, just leaks some bytes : p

common/stream-ratelimiter/src/lib.rs
phlip9: i'm not sure what you mean about locking the semaphore, but we actually have exclusive access in `poll_next` (since we have a `Pin<&mut Self>`), so only one thing can `poll_next` at a time. i didn't try to change any logic here, just changed the sub to saturating sub to ensure we never accidentally underflow.

as for the logic itself, it just refills the whole token bucket at every timer tick.

common/bounded-executor/src/lib.rs
phlip9: just some refactoring after stream-ratelimiter no longer could use it and it turned out to be unnecessary now in bounded-executor as well

network/src/peer/fuzzing.rs
phlip9: random number chosen by fair dice roll!

more seriously, our typical channel size is ~1024, which would never get hit in this fuzzer (we only ever sample <= 20 messages per iter), so I chose a lower number here

common/stream-ratelimiter/src/lib.rs
gregnazario: Yeah, guess I was just commenting for myself.

This saturating sub is not the leak fix, but the `Arc` change is.

network/src/peer/fuzzing.rs
phlip9: it writes each serialized `NetworkMessage` as a frame like `{ 4-byte length prefix } || { serialized network message }` into the io stream. the `MAX_FRAME_SIZE` actually limits the size of each serialized network message. so if we try to read or write a serialized network message to the io stream with length > MAX_FRAME_SIZE, it'll return an Err

network/src/peer/fuzzing.rs
phlip9: i'm using this as a low-tech assertion that we don't have any per-remote-peer metrics, since it'll eventually OOM if we do : p

unfortunately, we're already going to get some nasty non-determinism since we're using async stuff (the basic scheduler should limit this at least) and internal timers : (

network/src/peer/fuzzing.rs
phlip9: oh, there's literally a crate name `channel` in common/channel

network/src/peer/fuzzing.rs
phlip9: > If more than 8 messages are sent to any channel, might it block the fuzzer?

nope! just imagine this is like node.js when we use basic scheduler. it's all single-threaded (all tasks are running on one thread) but tokio still multiplexes across them (when one task can't make progress, it'll switch to another pending task). so if this task is waiting on a network notif, tokio will switch to another pending task (e.g. the one reading network messages) to continue making progress.

network/src/peer/mod.rs
mimoo, line 156: 👍

network/src/peer/fuzzing.rs
mimoo: ah I see that [in the spec](https://github.com/libra/libra/blob/master/specifications/network/messaging-v1.md#framing), we might want to test that in combo with the noise fragmentation as well : o

network/src/peer/fuzzing.rs
mimoo: wait, but isn't it a legit OOM then? Remember that the fuzzer resets the state of everything between runs

network/src/peer/fuzzing.rs
mimoo: oh that's why xD yeah it sounds like we should rename this to libra_channel or something no? At a glance it is hard to understand what type of channel this is

network/src/peer/fuzzing.rs
mimoo: Ah, I missed that the last read will return `Poll::Ready(Ok(0))` which is interpreted as `EOF`.

And to your answer, my understanding here is that `collect()` here will continuously read the stream, potentially unblocking the sender. Isn't there a `drain()` function to avoid collecting :P?

network/src/peer/mod.rs
mimoo: also: since we already have an `info!` if the message is empty, you could move this log bellow in the match case `Some(Ok(message)) =>`

network/src/peer/mod.rs
mimoo: it's also a one-liner function, I'm against these :P

testsuite/libra-fuzzer/README.md
```diff
@@ -22,7 +22,7 @@ a corpus with `cargo run --bin libra-fuzzer generate <target>`.
 Once a corpus has been generated, the fuzzer is ready to use, simply run:
 
 ```
-RUSTC_BOOTSTRAP=1 cargo run --bin libra-fuzzer fuzz <target> -- --release --debug-assertions
+RUSTC_BOOTSTRAP=1 cargo run --bin libra-fuzzer --release fuzz <target>
```
mimoo, line 25: can you actually use `--debug` instead :D? The rationale could be included in the README as well: we want a number of checks to be enabled when we fuzz. For example, [overflow-checks](https://rust-lang.github.io/rfcs/1535-stable-overflow-checks.html).

testsuite/libra-fuzzer/README.md
```diff
@@ -115,10 +115,42 @@ fuzz/google-oss-fuzz/build_fuzzer.sh ConsensusProposal .
 
 ### Troubleshooting
 
-#### linking with CC failed
+#### My backtrace does not contain file names and line numbers
 
-Are you on MacOS? Have you checked that Xcode is up-to-date?
```
mimoo, line 120: can you leave that one that one? That's actually a fairly common one : o

testsuite/libra-fuzzer/README.md
```diff
@@ -115,10 +115,42 @@ fuzz/google-oss-fuzz/build_fuzzer.sh ConsensusProposal .
...
+You need to use `llvm-symbolizer`, see https://github.com/rust-fuzz/cargo-fuzz/issues/160
 
-#### My backtrace does not contain file names and line numbers
+#### macOS: Linking with `cc` failed
+
+If you get a linker error like
```
mimoo, line 124: thanks for adding on that issue :D !!

network/src/lib.rs
phlip9: hmm, i don't know what the right tradeoff is here, but I feel the conditional compilation is maybe getting a bit out of hand imo

network/src/peer/mod.rs
phlip9: I don't think these per-actor init/stop messages are buying us enough. especially for public-facing nodes, emitting 10+ logs every time someone connects to us seems a bit crazy to me

network/src/peer/mod.rs
phlip9: it looks like `handle_inbound_message` already does this trace log effectively, I'll just remove this

common/stream-ratelimiter/src/lib.rs
phlip9: err, sorry my comment above was not clear, the leak is the `ManuallyDrop`

```rust
*self.as_mut().project().permit_fut =
    Some(Box::pin(semaphore.into_permit().map(|permit| {
        // Permits should not be added back automatically. They are added on each
        // interval tick.
        ManuallyDrop::new(permit);    // <~~~~~~ this is the memory leak
    })));
```

network/src/peer/fuzzing.rs
phlip9: the llvm fuzzer doesn't spin up a separate process for each iter, so static/global state is preserved between iters. i don't know about the google OSS fuzzer, maybe you can comment on that.

in this specific case, we have a global metrics collector (basically a big global `HashMap<MetricName, Counter>`). if every fuzzer iter pokes a counter with a different name/set of labels this big hashmap will blow up and we'll OOM.

as noted in the comment above this comment, this is expected behaviour in some cases. for example, if we use a different self-peer-id every fuzzer iter, we'll quickly OOM since we include the self-peer-id as a label in all network metrics.

network/src/peer/fuzzing.rs
phlip9: yeah, this is definitely confusing 😅 

there's a crate `channel` in `common/channel` that contains two channel implementations: (1) a wrapper around the blocking `futures::mpsc` channel and (2) `libra-channel`, a keyed, non-blocking, multi-channel with different eviction policies

network/src/peer/fuzzing.rs
phlip9: yeah `collect()` will keep polling the stream until it finishes. no `drain()` function though : (

network/src/peer/fuzzing.rs
mimoo: ah, any ways to disable these globals when fuzzing :D? Where is this being hit? Like, does that impact other fuzzers as well?

testsuite/libra-fuzzer/README.md
phlip9: definitely a trade-off here, in `--debug` mode we'll have much lower iters/sec, cover less of the search space, and probably miss more hard-to-find bugs. the binary will also look more different than prod, which runs in release mode

phlip9:
Ok, just removed stream-ratelimiter completely, following the release-1.0 cherry-pick

network/src/lib.rs
mimoo: yep, pretty ugly :/ it'd be nice to have a macro for writing stuff like

```
#[pub(not(any(feature = "testing", feature = "fuzzing")))]
mod noise;
```

network/src/peer/fuzzing.rs
phlip9: nope! no way to disable them that i know of. for metrics, the global is the default metrics registry: https://docs.rs/prometheus/0.10.0/src/prometheus/registry.rs.html#309

this is one reason why globals suck, but it probably beats passing around a `&Metrics` handle everywhere

mimoo:
fuzzing stuff looks good to me, @gregnazario or @davidiw can you guys approve?

network/src/peer/fuzzing.rs
mimoo: Damn that looks weird to me, I'd imagine a better way would be to have an initialized static + counters function only when initialized = true

testsuite/libra-fuzzer/README.md
mimoo: yeah, we've talked about that before and we chose to fuzz things in debug mode instead. We are definitely missing on a bunch of bugs if we don't do this :/ we could run fuzzers in both mode as well...

network/src/peer/mod.rs
gregnazario: Could you separate out the refactor bits from the fuzzing of this commit into a separate refactor?  Looks like they're completely file separate, so it shouldn't be too bad.

phlip9: done!

gregnazario (approved):

bothra90:
Nice find on the memory leak, @phlip9. Seems like I didn't fully grok what `ManuallyDrop` did and used it incorrectly here and at least [one other place](https://github.com/libra/libra/search?q=manuallydrop). You might want to fix that as well. 

I'm also curious about the fix you had in mind and the alternative token-bucket implementation you are considering - I wasn't really happy with this one either, but it was the best I could come up with back then 😅

mimoo:
come on @bothra90 we're waiting for your PR for the other one :P

PR #6602, opened 2020-10-27
[ci] support for versioned ci images

rexhoffman:
In prep to prebuild a version of sccache from github that is a different version than that used by the release-1 branch, this is an initial setup to version the release of ci docker images.

## Motivation

While this may not be strictly necessary it seems better to future proof our environment by versioning our build environments, so that we could grab old code if we need, build the image and push to docker if the image had been deleted and rerun the build.

This will be cherry picked to the release branch, and a subsequent pr on master will bump the version and sccache version.

Yes

## Test Plan

CI

## Related PRs

None, yet

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -10,6 +10,7 @@ aliases:
         - auto
         - canary
   - &working_directory /opt/git/libra
+  - &image libra/build_environment:circleci-1
 
 version: 2.1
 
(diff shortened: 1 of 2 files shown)


scripts/dev_setup.sh
```diff
@@ -14,6 +14,7 @@
 SHELLCHECK_VERSION=0.7.1
 HADOLINT_VERSION=1.17.4
 SCCACHE_VERSION=0.2.13
+#SCCACHE_GIT='https://github.com/rexhoffman/sccache.git@19fef99c15765ea73460fd9ecb209c35313eac41'
```
metajack, line 17: Did you mean to leave this in and commented out?

metajack: Maybe just leave it blank and have a comment about what it does?

scripts/dev_setup.sh
```diff
@@ -564,6 +571,7 @@ if [[ "$INSTALL_BUILD_TOOLS" == "true" ]]; then
   rustup component add rustfmt
   rustup component add clippy
 
+  install_old_sccache
```
metajack, line 574: `install_sccache`? I don't see `install_old_sccache` defined in here.

.circleci/config.yml
```diff
@@ -10,6 +10,7 @@ aliases:
         - auto
         - canary
   - &working_directory /opt/git/libra
+  - &image libra/build_environment:circleci-1
```
metajack, line 13: Is the `docker.io` prefix unnecessary? That was removed.

metajack (commented):
Look good aside from notes below.

.circleci/config.yml
rexhoffman: it's not.  hoping this will help if multiple dns names exist for dockerhub (they seem to).

scripts/dev_setup.sh
rexhoffman: good catch

metajack (approved):

PR #6604, opened 2020-10-27
[cluster test] gracefully terminate stubbed node in load test

christinacelee:
## Motivation

to remove the following panic when currently running load test:
```
thread 'main' panicked at 'Cannot drop a runtime in a context where blocking is not allowed. This happens when a runtime is dropped from within an asynchronous context.', /root/.cargo/registry/src/github.com-1ecc6299db9ec823/tokio-0.2.22/src/runtime/blocking/shutdown.rs:49:21
thread 'tokio-runtime-worker' panicked at 'SelectNextSome polled after terminated', /rustc/d3fb005a39e62501b8b0b356166e515ae24e2e54/src/libstd/macros.rs:13:23
```

## Test Plan

ran load test and checked panic no longer happens and experiment successfully runs to completion

### DIFF
--- a/testsuite/cluster-test/src/experiments/load_test.rs
+++ b/testsuite/cluster-test/src/experiments/load_test.rs
@@ -29,6 +29,8 @@ use std::{
 use structopt::StructOpt;
 use tokio::runtime::{Builder, Runtime};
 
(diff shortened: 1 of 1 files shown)


snakech (approved):
Thx for such quick fix of this @christinacelee

PR #6605, opened 2020-10-27
[move-prover] Updated the documentation generator to be able to incude diagrams

junkil-park:
- added the parameter "with-diagram" for the `stdlib`'s cli
- by default, the "with-diagram" flag is false
- when the flag "with-diagram" is set to true,
  - the images (dependency diagrams) are generated in `modules/doc/img`,
  - the markdown documents include the image files accordingly

If you'd like to take a look at the generated documents with diagrams:
- https://github.com/junkil-park/libra/tree/diagram-demo/language/stdlib/modules/doc

TODO:
- add test cases to docgen for this feature
- update `dev_setup.sh` to install the external graph visualization tool `graphviz`

## Motivation

To generate the documentation with diagrams included

yes

## Test Plan

cargo test

### DIFF
--- a/language/move-prover/docgen/src/docgen.rs
+++ b/language/move-prover/docgen/src/docgen.rs
@@ -15,18 +15,19 @@ use spec_lang::{
     code_writer::{CodeWriter, CodeWriterLabel},
     emit, emitln,
     env::{
-        FunctionEnv, GlobalEnv, Loc, ModuleEnv, ModuleId, NamedConstantEnv, Parameter, StructEnv,
-        TypeConstraint, TypeParameter,
+        FunId, FunctionEnv, GlobalEnv, Loc, ModuleEnv, ModuleId, NamedConstantEnv, Parameter,
+        QualifiedId, StructEnv, TypeConstraint, TypeParameter,
     },
     symbol::Symbol,
     ty::TypeDisplayContext,
 };
 use std::{
     cell::RefCell,
-    collections::{BTreeMap, BTreeSet},
-    fs::File,
-    io::Read,
+    collections::{BTreeMap, BTreeSet, VecDeque},
+    fs::{self, File},
+    io::{Read, Write},
     path::PathBuf,
+    process::{Command, Stdio},
     rc::Rc,
 };
 
@@ -139,6 +140,10 @@ pub struct DocgenOptions {
     /// An optional file containing reference definitions. The content of this file will
     /// be added to each generated markdown doc.
     pub references_file: Option<String>,
+    /// Whether to include dependency diagrams in the generated docs.
+    pub include_dep_diagrams: bool,
+    /// Whether to include call diagrams in the generated docs.
+    pub include_call_diagrams: bool,
 }
 
 impl Default for DocgenOptions {
@@ -155,6 +160,8 @@ impl Default for DocgenOptions {
             doc_path: vec!["doc".to_string()],
             root_doc_templates: vec![],
             references_file: None,
+            include_dep_diagrams: false,
+            include_call_diagrams: false,
         }
     }
 }
@@ -236,8 +243,7 @@ impl<'env> Docgen<'env> {
         }
     }
 
-    /// Generate documentation, returning a pair of output file names and generated content per
-    /// file.
+    /// Generate document contents, returning pairs of output file names and generated contents.
     pub fn gen(mut self) -> Vec<(String, String)> {
         // Compute missing information about schemas.
         self.compute_declared_schemas();
@@ -612,6 +618,27 @@ impl<'env> Docgen<'env> {
         }
         self.end_code();
 
+        if self.options.include_dep_diagrams {
+            let module_name = module_env.get_name().display(module_env.symbol_pool());
+            self.gen_dependency_diagram(module_env.get_id(), true);
+            self.begin_collapsed(&format!(
+                "Show all the modules that \"{}\" depends on directly or indirectly",
+                module_name
+            ));
+            self.image(&format!("img/{}_forward_dep.svg", module_name));
+            self.end_collapsed();
+
+            if !module_env.is_script_module() {
+                self.gen_dependency_diagram(module_env.get_id(), false);
+                self.begin_collapsed(&format!(
+                    "Show all the modules that depend on \"{}\" directly or indirectly",
+                    module_name
+                ));
+                self.image(&format!("img/{}_backward_dep.svg", module_name));
+                self.end_collapsed();
+            }
+        }
+
         let spec_block_map = self.organize_spec_blocks(module_env);
 
         if !module_env.get_structs().count() > 0 {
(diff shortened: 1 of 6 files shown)


language/move-prover/diagen/src/lib.rs
```diff
@@ -0,0 +1,230 @@
...
+        Self {
+            generate_dot: false,
+            generate_pdf: false,
+            generate_svg: false,
+            generate_global_graph: false,
+        }
+    }
+}
+
+pub fn move_files_in(path: &PathBuf) -> Vec<PathBuf> {
```
wrwg, line 36: I think the tool should not generate diagrams for every .move file in the folder when called from docgen. Rather you should have an entry point into this crate where it generates a diagram file at a specific file location,

language/move-prover/diagen/src/lib.rs
```diff
@@ -0,0 +1,230 @@
...
+    fs::read_dir(path)
+        .unwrap()
```
wrwg, line 38: This tool will just crash if the path does not exists. This is not what we expect from tools in the Move platform. You should use Rust Result type for most of those functions here and generate error messages based on this.

language/move-prover/diagen/src/lib.rs
```diff
@@ -0,0 +1,230 @@
...
+                Some(path)
+            } else {
+                None
+            }
+        })
+        .collect()
+}
+
+// Locate the path of the "language" directory in the Libra repository.
+pub fn locate_language_dir() -> io::Result<PathBuf> {
```
wrwg, line 51: Docgen is a general tool which may be run outside of the libra tree. Even if its in the Libra tree, the location of diagen may change. This here is not very clean. It crashes if somewhere else in the file system. As said, the path to where to store diagen should be passed into a function in this crate, then this might not be longer necessary.

language/move-prover/diagen/src/lib.rs
```diff
@@ -0,0 +1,230 @@
...
+) -> anyhow::Result<()> {
+    if (options.generate_pdf || options.generate_svg) && !is_command_available("dot") {
+        error!("Command not found: dot\nTo install, `brew install graphviz`");
+        return Err(anyhow::anyhow!("Command not found: dot"));
+    }
+
+    let mut dep_graph: BTreeMap<String, Vec<String>> = BTreeMap::new();
+    let mut dep_graph_inverse: BTreeMap<String, Vec<String>> = BTreeMap::new();
+
+    for path in inp_files.iter() {
```
wrwg, line 76: Scanning of files should be unnecessary. As you see in the docgen code, the usage relation is already known (that is how we generate the `use` statements).

language/move-prover/docgen/src/docgen.rs
```diff
@@ -236,9 +240,37 @@ impl<'env> Docgen<'env> {
         }
     }
 
-    /// Generate documentation, returning a pair of output file names and generated content per
-    /// file.
-    pub fn gen(mut self) -> Vec<(String, String)> {
+    /// Generate documentation.
+    pub fn generate(self) -> anyhow::Result<()> {
+        if self.options.include_diagrams {
```
wrwg, line 245: I think you come to a much cleaner design if you include the call to dot/img generation here and do not use the diagen crate for this. You do not need to scan files for the usage relation because this is already calculated here. You do not need to scan directories for .move files. All this is currently very bridle and depends on that this is in the libra tree, and duplicates functionality which we already have.

You can just add a function to generate a diagram for a single module here, it should be rather simple function. Then you call this function at the same place where the `use` declarations are generated in this module.

wrwg (changes requested):

language/move-prover/diagen/src/lib.rs
```diff
@@ -0,0 +1,230 @@
...
+use serde::{self, Deserialize, Serialize};
+use std::{
+    collections::{btree_map::Entry, BTreeMap, BTreeSet, VecDeque},
+    env, fs,
+    io::{self, Write},
+    path::{Path, PathBuf},
+    process::{Command, Stdio},
+};
+
+/// Options passed into the documentation generator.
```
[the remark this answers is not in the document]
junkil-park: done.

language/move-prover/diagen/src/lib.rs
junkil-park: Refactored accordingly.

language/move-prover/diagen/src/lib.rs
junkil-park: Refactored accordingly.

language/move-prover/diagen/src/lib.rs
junkil-park: Refactored. This functions is no longer used.

language/move-prover/diagen/src/lib.rs
junkil-park: Refactored accordingly.

language/move-prover/docgen/src/docgen.rs
junkil-park: The newer commit refactored the code based on the comments here.

language/move-prover/docgen/src/docgen.rs
```diff
@@ -664,7 +688,76 @@ impl<'env> Docgen<'env> {
...
+                if !visited.contains(dep_id) {
+                    visited.insert(*dep_id);
+                    queue.push_back(*dep_id);
+                }
+            }
+        }
+        dot_src.push_str("}}\n");
+
+        let mut out_path = PathBuf::from(&self.options.output_directory).join("img");
+        fs::create_dir_all(&out_path).unwrap(); // TODO: error handling
```
junkil-park, line 732: In this function, there are a few places where error handling needs to be done properly. What would be a good way to handle the errors in this case? `gen_dependency_diagram` and its callers (e.g., `gen_modules(...)` and `gen()`) don't return the Result type (e.g., anyhow::Result<()>).

wrwg: In general, we add errors to the global env, and then either continue and compensate, or gracefully terminate (preferable the first one). There are examples, e.g. search for the string ""cannot read root template" to see one. The overall driver will print all the errors at the end of the run.

language/move-prover/spec-lang/src/env.rs
```diff
@@ -1475,6 +1509,12 @@ impl<'env> ModuleEnv<'env> {
         module_env.into_function(FunId::new(self.env.symbol_pool.make(view.name().as_str())))
     }
 
+    /// Gets a FunctionEnv by FunctionInstantiationIndex.
+    /// The returned function might be from this or another module.
+    pub fn get_function_by_inst_idx(&self, idx: FunctionInstantiationIndex) -> FunctionEnv<'_> {
```
wrwg, line 1514: This function does not need to be public, and perhaps consider unfolding it directly at the caller side.

We very rarely need to work with the handles, since the original bytecode will be translated to our own, which is not longer handle based.

wrwg (commented):
This looks generally good. For new reviewers, the changes to diagen code may be confusing, because they are not needed for this PR.

language/move-prover/spec-lang/src/env.rs
junkil-park: I unfolded the function at the caller side, and removed this function.

language/move-prover/docgen/src/docgen.rs
```diff
@@ -864,6 +1060,19 @@ impl<'env> Docgen<'env> {
                 spec_block_map,
             )
         }
+        if self.options.include_diagrams {
```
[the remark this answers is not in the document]
junkil-park: To `DocgenOption`, I added a new flag `include_call_diagrams: bool` in addition to the existing one `include_dep_diagrams: bool`. However, Stdlib's cli still has one flag called `with_diagrams`. When `with_diagrams` is true, both `include_call_diagrams` and `include_dep_diagrams` would become true currently.

language/move-prover/docgen/src/docgen.rs
```diff
@@ -664,7 +688,179 @@ impl<'env> Docgen<'env> {
...
+
+        let mut dot_src: String = String::from("digraph G {\n");
+        let mut visited: BTreeSet<QualifiedId<FunId>> = BTreeSet::new();
+        let mut queue: VecDeque<QualifiedId<FunId>> = VecDeque::new();
+
+        visited.insert(fun_id);
+        queue.push_back(fun_id);
+
+        while let Some(id) = queue.pop_front() {
+            let env = self.env.get_function(id);
```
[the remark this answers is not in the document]
junkil-park: I renamed it not to use the name `env`.

language/move-prover/docgen/src/docgen.rs
```diff
@@ -612,6 +615,27 @@ impl<'env> Docgen<'env> {
         }
         self.end_code();
 
+        if self.options.include_diagrams {
+            let module_name = module_env.get_name().display(module_env.symbol_pool());
+            self.gen_dependency_diagram(module_env.get_id(), true);
+            self.begin_collapsed(&format!(
+                "Show all the modules that \"{}\" depends on directly or indirectly",
+                &module_name
```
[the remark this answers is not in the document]
junkil-park: Done.

language/move-prover/diagen/src/lib.rs
```diff
@@ -0,0 +1,203 @@
+// Copyright (c) The Libra Core Contributors
```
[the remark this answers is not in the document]
junkil-park: Done.

language/move-prover/spec-lang/src/env.rs
```diff
@@ -1462,9 +1496,9 @@ impl<'env> ModuleEnv<'env> {
             })
     }
 
-    /// Get FunctionEnv for a function used in this module, via the FunctionHandleIndex. The
-    /// returned function might be from this or another module.
-    pub fn get_called_function(&self, idx: FunctionHandleIndex) -> FunctionEnv<'_> {
+    /// Gets a FunctionEnv by FunctionHandleIndex.
+    /// The returned function might be from this or another module.
+    pub fn get_function_by_handle_idx(&self, idx: FunctionHandleIndex) -> FunctionEnv<'_> {
```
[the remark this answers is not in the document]
junkil-park: I revived the previous comment, and renamed it to `get_used_function`, which is similar to the previous one, and distinct from some other function like `FunctionEnv::get_called_functions`. Please let me know what you think.

language/move-prover/docgen/src/docgen.rs
```diff
@@ -664,7 +691,179 @@ impl<'env> Docgen<'env> {
...
+        let fun_env = self.env.get_function(fun_id);
+        let name_of = |env: &FunctionEnv| {
+            if fun_env.module_env.get_id() == env.module_env.get_id() {
+                env.get_simple_name_string()
+            } else {
+                Rc::from(format!("\"{}\"", env.get_name_string()))
+            }
+        };
+
+        let mut dot_src: String = String::from("digraph G {\n");
```
tzakian, line 705: Please correct me if I'm wrong, but it looks like all of the strings being pushed on to `dot_src` are newline-suffixed. What would you think of having `dot_src` be a `Vec<String>` and then when `gen_svg_file` is called passing `dot_src.join("\n")`?

language/move-prover/docgen/src/docgen.rs
```diff
@@ -664,7 +691,179 @@ impl<'env> Docgen<'env> {
...
+
+        self.gen_svg_file(&out_file_path, &dot_src);
+    }
+
+    /// Generate a forward (or backward) dependency diagram (.svg) for the given module.
+    fn gen_dependency_diagram(&self, module_id: ModuleId, is_forward: bool) {
+        let module_env = self.env.get_module(module_id);
+        let module_name = module_env.get_name().display(module_env.symbol_pool());
+
+        let mut dot_src: String = String::from("digraph G {\n");
```
tzakian, line 761: Same question/possible update to `dot_src` here as in `gen_call_diagram`

wrwg (approved):
Great work.

language/move-prover/docgen/src/docgen.rs
```diff
@@ -664,7 +691,179 @@ impl<'env> Docgen<'env> {
...
+        visited.insert(fun_id);
+        queue.push_back(fun_id);
+
+        while let Some(id) = queue.pop_front() {
+            let caller_env = self.env.get_function(id);
+            let caller_name = name_of(&caller_env);
+            let callee_list = caller_env.get_called_functions();
+
+            if fun_env.module_env.get_id() == caller_env.module_env.get_id() {
+                dot_src.push_str(&format!("    {}\n", caller_name));
```
tzakian, line 718: Might be personal preference, but what would you think of using `\t` instead of the fully written-out spaces in these strings?

language/move-prover/docgen/src/docgen.rs
```diff
@@ -664,7 +691,179 @@ impl<'env> Docgen<'env> {
...
+        while let Some(id) = queue.pop_front() {
+            let mod_env = self.env.get_module(id);
+            let mod_name = mod_env.get_name().display(mod_env.symbol_pool());
+            let dep_list = if is_forward {
+                mod_env.get_used_modules(false)
+            } else {
+                mod_env.get_using_modules(false)
+            };
+            dot_src.push_str(&format!("    {}\n", mod_name));
+            for dep_id in dep_list.iter() {
```
tzakian, line 777: Iterating over `dep_list.iter().filter(|dep_id| dep_id != id) { ....}` could make this cleaner and would be more "Rusty" as you shouldn't need the `if .. { continue }` below.

tzakian (approved):
Looks good to me! A couple style nits, but feel free to accept/ignore as you see fit.

language/move-prover/docgen/src/docgen.rs
junkil-park: Done.

language/move-prover/docgen/src/docgen.rs
junkil-park: Done.

language/move-prover/docgen/src/docgen.rs
junkil-park: Done.

language/move-prover/docgen/src/docgen.rs
junkil-park: Done.

PR #6606, opened 2020-10-27
[json-rpc] ensure response data use lower case for hex-encoded string

xli:
## Motivation

JSON-RPC api response for account address case format is inconsistent: some uses lower case, some uses upper case.

Consistent lower case format can avoid people make mistake like: string compare without ignoring case.
Especially for account address, instead of comparing bytes, developers may direct compare hex-encoded string, which should be case insensitive.
This PR also changes other hex-encoded data type to use BytesView for ensuring lower case across board.

## Test Plan

unit tests

## Related PRs

#6296


libra-action:

Breaking changes policy:

2. Adding fields is ok.

json-rpc/src/methods.rs
```diff
@@ -534,7 +536,7 @@ async fn get_account_transactions(
 
         all_txs.push(TransactionView {
             version: tx.version,
-            hash: tx.transaction.hash().to_hex(),
+            hash: BytesView::from(tx.transaction.hash().to_vec()),
             bytes: lcs::to_bytes(&tx.transaction)?.into(),
```
thefallentree, line 540: see here,  if we declare everything in response field as bytesview directly we could sae all these BytesView::from()

thefallentree (approved):
Great change! Thanks for making the codebase more consistent and clean!

xli:
\land

PR #6610, opened 2020-10-27
[cherry-pick][network] fix memory leak, remove stream-ratelimiter from network

phlip9:
### Justification

This small cherrypicked diff removes stream-ratelimiter from network, which contained a memory leak. Every time a peer connected, sent at least one message, and then disconnected, we would leak a `tokio::sync::Semaphore`. We will be introducing a more robust token-bucket filter in the near future to replace stream-ratelimiter; for now, the stream-ratelimiter is causing more problems than it fixes.

The memory leak will eventually cause a node (validator, VFN, PFN) to OOM after too many peers connect, send at least one message, and then disconnect.

### DIFF
--- a/network/src/peer/mod.rs
+++ b/network/src/peer/mod.rs
@@ -26,14 +26,9 @@ use libra_logger::prelude::*;
 use libra_types::PeerId;
 use netcore::compat::IoCompat;
 use serde::{export::Formatter, Serialize};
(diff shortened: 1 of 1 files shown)


gregnazario (approved):
No rate limiting.  Let's hope we can get our next rate limiter out soon

sausagee (approved):

PR #6620, opened 2020-10-29
[debug printing] use short_str_lossless printing for addresses in lan…

sblackshear:
…guage and types code

The AccountAddress:short_str() function truncates account addresses, which can make debugging difficult. This PR replaces uses of this function in language/ and types/ with short_str_losslesses(), which compresses account addresses without losing information.


n4ss (commented):
<3 awesome

tnowacki (approved):

PR #6621, opened 2020-10-29
cherry-pick [json-rpc] ensure response data use lower case for all hex-encoded bytes data

xli:
#### Why we must have it for V1 launch

JSON-RPC api response for account address case format is inconsistent: some uses lower case, some uses upper case.
It is not a bug, as hex-encoded string is case insensitive. 

But it may become a bug source as if some developers do not know the hex-encoded string is case insensitive and compare the account address returned from server without ignoring cases.
Recently @metajack found Coinbase treated hex-encoded account address case sensitive.

Hence includes this in V1 launch can prevent some simple but easy make bugs, improving developer experience.

Note: it is unintentional change (#6296) tried to make all account address to hex becomes upper cases recently, but the change didn't update all account address hex-encoded string in json-rpc response. 

We choose lower case for json-rpc response across board, because most of hex-encoded string in current response is lower case, and Ethereum uses capitalization for checksum.

#### What workarounds and alternative we have if we do not push the PR

1. Our client sdk code should downcase hex-encoded string before handing over data to caller. This is limited, as we only provide sdk in 4 languages.

## Test Plan

N/A

## Related PRs

#6606


aching:
Can we update the relevant specifications to be lower-case?

thefallentree (approved):

metajack:
The followup actions from the gatekeepers to get this landed on release branch were:

1. Fix the remaining case where hex encoded addresses are uppercase in JSON-RPC
2. Update the spec to mandate lower case hex addresses in the JSON-RPC API
3. Update coding conventions to warn about use of BytesView vs. String to mitigate future mistakes

xli:
close for now

PR #6622, opened 2020-10-29
[move-cli] Rename binary and reorganize project tree

metajack:
The move-cli binary is now just named move. The crate did not change names,
only the binary it produces.

The following directories were renamed:

move_src -> src/modules and src/scripts
move_data -> storage
move_build_output -> build


sblackshear (approved):
LGTM, thanks!

metajack:
Minor changes for lint passing and an addition to .gitignore were added.

/land

PR #6627, opened 2020-10-30
[config] set chunk size for state sync to 1k

davidiw:
state syncs txn rate is approximately:
chunk size / (RTT to upstream + execution time)
consensus txn rate is approximately:
chunk size / (2 x slowest latency + execution time)

in addition consensus chunk size is set to 1000, whereas state sync is
currently at 250... as long as these two chunk sizes are the same, we shouldn't see a case for state sync to fall behind to the point it cannot catch up.


davidiw:
Sigh... tests that depend on hardcoded values....

zekun000 (approved):
the tests...

PR #6629, opened 2020-10-30
[management] If network address decryption fails don't error

davidiw:
Much like we did for validator sets, put in a dummy text instead of erroring out...


JoshLind (approved):
LGTM! Thanks David. 

Note: As this was also on my roadmap I'll follow this PR up with something that unifies the error handling (across validator config and validator set), and also adds tests for this functionality.

davidiw:
ahh, I was looking into something, so I added it to unblock myself :P

PR #6630, opened 2020-10-30
[move source language] Fixed two bugs 

### LINKED ISSUE #6607: [Bug] Type checker crashes if not enough type arguments specified for the return type of a function and any field accessed
# 🐛 Bug

## To reproduce

**Code snippet to reproduce**
```rust
module M {
    struct Deal<T> {
        value: u128,
    }

    fun get_deal(): Deal {}

    fun main() {
        get_deal().value
    }
}
```

**Stack trace/error message**
```
thread 'main' panicked at 'assertion failed: tps.len() == args.len()', language/move-lang/src/typing/core.rs:1038:5
```

## Expected Behavior

It should show a compiler error about "Not enough type arguments" for the return type. 

## System information

**Please complete the following information:**
- latest master
- 1.46
- linux

tnowacki:
- Fixes #6607 
  - Moved arity checking into naming for structs. Which prevents having to instantiate the type each time you call a function 
- Fixes a bug with loops
  - label renumbering was not applied to infinite loop heads 

## Motivation

- Fixes two bugs 

## Test Plan

- Tests for the bug cases

### DIFF
--- a/language/move-lang/src/naming/translate.rs
+++ b/language/move-lang/src/naming/translate.rs
@@ -37,7 +37,7 @@ impl ResolvedType {
 struct Context {
     errors: Errors,
     current_module: Option<ModuleIdent>,
-    scoped_types: BTreeMap<ModuleIdent, BTreeMap<String, (Loc, ModuleIdent, Option<Kind>)>>,
+    scoped_types: BTreeMap<ModuleIdent, BTreeMap<String, (Loc, ModuleIdent, Option<Kind>, usize)>>,
     unscoped_types: BTreeMap<String, ResolvedType>,
     scoped_functions: BTreeMap<ModuleIdent, BTreeMap<String, Loc>>,
     unscoped_constants: BTreeMap<String, Loc>,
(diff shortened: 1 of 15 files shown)


language/move-lang/src/naming/translate.rs
```diff
@@ -669,6 +679,37 @@ fn type_(context: &mut Context, sp!(loc, ety_): E::Type) -> N::Type {
...
+    name_f: F,
+    mut ty_args: Vec<N::Type>,
+    arity: usize,
+) -> Vec<N::Type> {
+    let args_len = ty_args.len();
+    if args_len != arity {
+        context.error(vec![(
+            loc,
+            format!(
+                "Invalid instantiation of '{}'. Expected {} type arguments but got {}",
```
sblackshear, line 694:
```suggestion
                "Invalid instantiation of '{}'. Expected {} type argument(s) but got {}",
```

sblackshear (approved):

PR #6636, opened 2020-10-31
[github actions] fix summaries on prs.

rexhoffman:
## Motivation

Get dep change summaries posted to PRs again, which have been broken for a while.

Yes

## Test Plan

Demo it working:  https://github.com/libra/libra/pull/6616

## Related PRs

None

### DIFF
--- a/.github/workflows/dep-summaries.yml
+++ b/.github/workflows/dep-summaries.yml
@@ -19,9 +19,10 @@ jobs:
     name: Calculate dependency summary changes
     if: ${{ github.event_name == 'push' || github.actor != 'bors-libra' }}
     outputs:
-      lsr-summary-diff: ${{ steps.verify-lsr.diff }}
-      lec-summary-diff: ${{ steps.verify-lec.diff }}
-      release-summary-diff: ${{ steps.verify-release.diff }}
+      lsr-summary-diff: ${{ steps.verify-lsr.outputs.diff }}
+      lec-summary-diff: ${{ steps.verify-lec.outputs.diff }}
+      release-summary-diff: ${{ steps.verify-release.outputs.diff }}
+      dep-auditors-list: ${{ steps.dep-auditors.outputs.list }}
     steps:
       - name: checkout
         uses: actions/checkout@v2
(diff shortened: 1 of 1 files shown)


.github/workflows/dep-summaries.yml
```diff
@@ -90,19 +91,11 @@ jobs:
...
           output="${output//$'\r'/'%0D'}"
           echo "::set-output name=diff::${output}"
-
       - name: read dep-auditors list
         id: dep-auditors
-        if: ${{ steps.verify-lsr.outputs.diff || steps.verify-lec.outputs.diff }}
         run: |
           output=$(cat ./.github/dep-auditors.json | jq -r 'join(",")')
           echo "::set-output name=list::${output}"
-      - name: require dep-audit review if TCB deps changed
```
metajack, line 100: You can't move this to the other job because only this job runs on push events. So this will remove the review requirement from the auto branch PR and so bors won't be able to enforce this.

The change above that removes the `if` line also can be restored if this remains here and you won't need to pass dep-auditor output around.

metajack (changes requested):
I must have made the mistake of not input the inputs when I refactored into multiple jobs. Thanks for catching this.

.github/workflows/dep-summaries.yml
rexhoffman: It looks to me like both jobs fire on:

if: ${{ github.event_name == 'push' || github.actor != 'bors-libra' }} (well one is flipped from the other)
And the on: section applies to all jobs....

So they should both fire everywhere.

I was viewing calc summaries as a data provider, and annotate-pr as the worker.   I originally had 3 different jobs but that felt wasteful.    I could see mergeing both jobs in to one so bors wouldn't have to change.

Thoughts?

rexhoffman: Restoring the if line above would cause the original line 173 (if it were pointed at the right value) to fail, and the use of that output is already guarded by an if statement.

https://github.com/libra/libra/pull/6636/files#diff-877b1dac10c8b67e30963c98116d9dfa92a541bc15ea8d7ca78b3c8c80ad8700L171-L173

rexhoffman: And as you can see here if the failing step is moved back in to the calculate job, the annotation update job never gets fired off:  https://github.com/libra/libra/runs/1342165650?check_suite_focus=true

rexhoffman: Splitting the job in to three, calc, enforce, and update pr seem the best approach then.   But the enforcement step can't occur before another step/job that you want to run.

https://github.com/libra/libra/runs/1342269715

rexhoffman: but if calc fails, and enforce never runs, what will bors choose to do?

.github/workflows/dep-summaries.yml
```diff
@@ -90,19 +91,24 @@ jobs:
...
       - name: read dep-auditors list
         id: dep-auditors
-        if: ${{ steps.verify-lsr.outputs.diff || steps.verify-lec.outputs.diff }}
         run: |
           output=$(cat ./.github/dep-auditors.json | jq -r 'join(",")')
           echo "::set-output name=list::${output}"
+
+  requires-dep-audit-review:
+    runs-on: ubuntu-latest
+    name: fail if dep audit review required.
```
metajack, line 102: Place change this field to "Require review for TCB dep summary changes" so that it is more consistent with the other top level names.

metajack (changes requested):
Splitting out the enforcement makes it clearer, even if the logic was correct. I like that `calc-summaries` no longer does double duty.

metajack:
Please squash as well.

metajack (approved):

PR #6637, opened 2020-10-31
Add building of Python Client SDK Docs

JoelMarcey:
## Motivation

Build the docs and link to them from developers.libra.org

(similar to what we do with the Rust Crate Docs)

Yes

## Test Plan

Local build w/ Docusaurus

Screen Shot 2020-10-30 at 7 10 46 PM

## Related PRs

N/A

### DIFF
--- a/developers.libra.org/scripts/build_docs.sh
+++ b/developers.libra.org/scripts/build_docs.sh
@@ -6,13 +6,15 @@
 # run this script from the project root using `./scripts/build_docs.sh`
 
 function usage() {
-  echo "Usage: $0 [-b] [-r]"
+  echo "Usage: $0 [-b] [-r] [-p]"
   echo ""
   echo "Build Libra documentation."
   echo ""
   echo "  -b   Build static version of documentation (otherwise start server)"
   echo ""
   echo "  -r   Build Libra Rust crate documentation"
+  echo ""
+  echo "  -p   Build Libra Python Client SDK documentation"
   echo ""
 }
 
@@ -26,8 +28,19 @@ function install_rustup {
   fi
 }
 
+function install_python {
+  echo "Installing Python 3......"
+  if python3 --version &>/dev/null; then
+    echo "Python 3 is already installed"
+  else
+    echo "Install Python 3 from https://www.python.org/"
+    exit 1
+  fi
+}
+
 BUILD_STATIC=false
 BUILD_RUSTDOCS=false
+BUILD_PYTORCH_SDK_DOCS=false
 
 if [[ "$(basename $PWD)" != "developers.libra.org" ]]; then
   echo "Didn't pass directory check."
@@ -40,7 +53,7 @@ if [[ "$(basename $PWD)" != "developers.libra.org" ]]; then
   exit 1
 fi
 
-while getopts 'hbr' flag; do
+while getopts 'hbrp' flag; do
   case "${flag}" in
     h)
       usage;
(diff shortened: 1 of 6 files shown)


xli (approved):

developers.libra.org/scripts/build_docs.sh
```diff
@@ -26,8 +28,19 @@ function install_rustup {
   fi
 }
 
+function install_python {
```
bmwill, line 31: This function seems in properly named since it's not actually installing python but just checking that it is installed

developers.libra.org/scripts/build_docs.sh
```diff
@@ -26,8 +28,19 @@ function install_rustup {
...
+    echo "Python 3 is already installed"
+  else
+    echo "Install Python 3 from https://www.python.org/"
+    exit 1
+  fi
+}
+
 BUILD_STATIC=false
 BUILD_RUSTDOCS=false
+BUILD_PYTORCH_SDK_DOCS=false
```
bmwill, line 43: pytorch? Do you mean pylibra or something else?

developers.libra.org/scripts/build_docs.sh
```diff
@@ -117,6 +133,30 @@ if [[ $BUILD_RUSTDOCS == true ]]; then
   cp -r $RUSTDOC_DIR $DOCUSAURUS_RUSTDOC_DIR
 fi
 
+if [[ $BUILD_PYTORCH_SDK_DOCS == true ]]; then
+  echo "-----------------------------------"
+  echo "Generating PyTorch Client SDK Docs"
+  echo "-----------------------------------"
```
bmwill, line 139: same here, I don't think we're buliding pytorch

bmwill (commented):
Couple of nits and a question:
This just seems to be downloading a published package and including the docs. Would it make more sense to do one of the following:
1. link to where the published docs should already exist for this package (does pip even host documentation for published packages?)?
2. Build and host the documentation for the python sdk in the python sdk repo and just link to that from the website

developers.libra.org/scripts/build_docs.sh
JoelMarcey: LOL - I always do that. Yes, it should be `PYTHON`

developers.libra.org/scripts/build_docs.sh
JoelMarcey: Good point, can change.

JoelMarcey:
> Would it make more sense to do one of the following:

> 1. link to where the published docs should already exist for this package (does pip even host documentation for published packages?)?
> 2. Build and host the documentation for the python sdk in the python sdk repo and just link to that from the website

I don't think so for (1). I would be ok with (2) if @xli is, but I think a point here is to make the Python SDK a canonical SDK for client developers. And, we are probably going to move the other SDK docs into developers.libra.org as well.

xli:
> > Would it make more sense to do one of the following:
> 
> > 1. link to where the published docs should already exist for this package (does pip even host documentation for published packages?)?
> > 2. Build and host the documentation for the python sdk in the python sdk repo and just link to that from the website
> 
> I don't think so for (1). I would be ok with (2) if @xli is, but I think a point here is to make the Python SDK a canonical SDK for client developers. And, we are probably going to move the other SDK docs into developers.libra.org as well.

for (1), no docs hosting.
for (2), I'm fine with either way, but if we think eventually want all sdk docs together in developers.libra.org, then seems no need to publish somewhere else.

bmwill:
yeah I figured (1) may not be an option. For (2) the only reason I bring it up is to try and avoid building the python docs on every libra commit, but given we probably want all the docs hosted in one place I can see how that wouldn't be an attractive option.

Anyway, I'm not blocking this PR from landing, my comments were more thought provoking than actionable anyway :)

xli (approved):

PR #6638, opened 2020-10-31
[Operational Tools] Add missing functionality and smoke tests for operator tool

JoshLind:
## Motivation

This PR adds missing functionality and smoke tests for the operator tool. Much of this functionality is motivated by previous experience when working with operators to debug/resolve issues. 

This PR offers several commits:
1. First, we add a "--set-genesis" flag to the insert-waypoint command. This allows operators to insert a new waypoint into storage, without overwriting the genesis waypoint needed by full nodes.
2. Next, we add a "print-waypoint" command to display the values of waypoints in storage. This is useful to debug waypoint related issues in storage.
3. Next, we add a smoke test to ensure that failed validator address decryption doesn't prevent the validator-config and validator-set commands from working/being displayed.
4. Next, we remove the redundant error handling for failed decryption of validator addresses in the validator-set command. This handling already happens in the DecryptedValidatorConfig struct.
5. Finally, we add a print-key command to print public keys held in storage. This is also useful for debugging issues without direct access to storage.

Yes.

## Test Plan

All tests pass locally. Moreover, running the tool locally also works. For example:

```
% ./libra-operational-tool print-key --key-name consensus --config config
{
  "Result": "d1e8cb3e32a9449c703cc6af9195b92c7044325dd70895b1137bee11e7732d9f"
}
```

```
 % ./libra-operational-tool print-waypoint --waypoint-name waypoint --config config
{
  "Result": "43636526:c77e83c40d7038c27f4de3f0b53427273f5c92ac341e38c50b69a6747020454a"
}

```

```
% ./libra-operational-tool print-waypoint --waypoint-name genesis-waypoint --config config
{
  "Result": "0:0fb6651cf471cdbe2318a2ad3e670a633580613a89b736d3d9db630a2c0dc0bc"
}
```

## Related PRs

None.

### DIFF
--- a/config/management/operational/src/command.rs
+++ b/config/management/operational/src/command.rs
@@ -7,7 +7,7 @@ use crate::{
 };
 use libra_crypto::{ed25519::Ed25519PublicKey, x25519};
 use libra_management::{error::Error, execute_command};
-use libra_types::account_address::AccountAddress;
+use libra_types::{account_address::AccountAddress, waypoint::Waypoint};
 use serde::Serialize;
 use structopt::StructOpt;
 
@@ -31,7 +31,9 @@ pub enum Command {
     #[structopt(about = "Set the waypoint in the validator storage")]
     InsertWaypoint(libra_management::waypoint::InsertWaypoint),
     #[structopt(about = "Prints an account from the validator storage")]
-    PrintAccount(crate::account::PrintAccount),
+    PrintAccount(crate::print::PrintAccount),
+    #[structopt(about = "Prints a waypoint from the validator storage")]
+    PrintWaypoint(crate::print::PrintWaypoint),
     #[structopt(about = "Remove a validator from ValidatorSet")]
     RemoveValidator(crate::governance::RemoveValidator),
     #[structopt(about = "Rotates the consensus key for a validator")]
@@ -65,6 +67,7 @@ pub enum CommandName {
     ExtractPublicKey,
     InsertWaypoint,
     PrintAccount,
+    PrintWaypoint,
     RemoveValidator,
     RotateConsensusKey,
     RotateOperatorKey,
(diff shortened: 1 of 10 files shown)


mgorven:
> First, we add a "--skip-genesis-waypoint" flag to the insert-waypoint command.

Maybe this should be inverted (`--genesis` to set genesis-waypoint otherwise not). Are we planning to fix these issues requiring setting a recent waypoint, or is this something operators will have to do?

config/management/operational/src/command.rs
```diff
@@ -31,7 +31,11 @@ pub enum Command {
     #[structopt(about = "Set the waypoint in the validator storage")]
     InsertWaypoint(libra_management::waypoint::InsertWaypoint),
     #[structopt(about = "Prints an account from the validator storage")]
-    PrintAccount(crate::account::PrintAccount),
+    PrintAccount(crate::print::PrintAccount),
+    #[structopt(about = "Prints a public key from the validator storage")]
+    PrintKey(crate::print::PrintKey),
+    #[structopt(about = "Prints a waypoint from the validator storage")]
+    PrintWaypoint(crate::print::PrintWaypoint),
```
davidiw, line 38: hahaha, you're already taking away part of my project proposal for this upcoming half... I wonder if we would rather have a generic tool for reading / writing storage rather than this?

davidiw (approved):

config/management/operational/src/command.rs
JoshLind: Hah! Indeed -- it might be good to create a more general tool. Right now I'd argue that if we want more access to vault, we could always query vault directly, but if we create a more general secure storage (not vault specific) a tool would be ideal! 😄

JoshLind:
Thanks, all!

> Maybe this should be inverted (--genesis to set genesis-waypoint otherwise not). 

Aah, agreed! I've changed it to "--set-genesis", so that it doesn't do it by default.

> Are we planning to fix these issues requiring setting a recent waypoint, or is this something operators will have to do?

This is a good question. I'm not entirely sure a plan has been scoped out for this -- we'll need to circle back on this to see if it's in the consensus roadmap.

PR #6640, opened 2020-11-02
[json-rpc-client] implement async_client::Client#batch_send

xli:
## Motivation

Our old client has batch send requests feature, implement batch_send for new async client, so that anyone uses old client and prefer to continue use batch requests, they can migrate over to new client.

The batch send has limit abilities on error handling comparing with other non-batch request in the new async client, 
because batch response may contain multiple different errors.
To keep the program model simple, we returns first error we hit when we validate responses, as most likely we should not get any errors.
It also simplifies retry logic.
However, when submit transaction, is maybe better to use batch_send_without_retry, as re-submit transaction will cause server return request error.

Yes

## Test Plan

Unit test

### DIFF
--- a/client/json-rpc/src/async_client/client.rs
+++ b/client/json-rpc/src/async_client/client.rs
@@ -354,25 +354,8 @@ impl<R: RetryStrategy> Client<R> {
         loop {
             let ret = self.send_without_retry(request).await;
             match ret {
-                Ok(r) => {
-                    return Ok(r);
-                }
-                Err(err) => {
-                    if !self.retry.is_retriable(&err) {
-                        return Err(err);
-                    }
-                    if retries < self.retry.max_retries(&err) {
-                        match retries.checked_add(1) {
-                            Some(i) if i < self.retry.max_retries(&err) => {
-                                retries = i;
-                            }
-                            _ => return Err(err),
-                        };
-                        tokio::time::delay_for(self.retry.delay(&err, retries)).await;
-                        continue;
-                    }
-                    return Err(err);
-                }
+                Ok(r) => return Ok(r),
+                Err(err) => retries = self.handle_retry_error(retries, err).await?,
             }
         }
     }
@@ -418,8 +401,25 @@ impl<R: RetryStrategy> Client<R> {
     /// For get_account and get_account_transaction request, the above convert will return
     /// result not found error instead of None.
     ///
+    /// This function calls to batch_send_with_retry, when you batch submit transactions,
+    /// to avoid re-submit errors, it is better to use batch_send_without_retry and handle errors
+    /// by yourself.
     pub async fn batch_send(&self, requests: Vec<Request>) -> Result<Vec<JsonRpcResponse>, Error> {
-        self.batch_send_without_retry(&requests).await
+        self.batch_send_with_retry(&requests).await
+    }
+
+    pub async fn batch_send_with_retry(
+        &self,
+        requests: &[Request],
+    ) -> Result<Vec<JsonRpcResponse>, Error> {
+        let mut retries: u32 = 0;
+        loop {
+            let ret = self.batch_send_without_retry(requests).await;
+            match ret {
+                Ok(r) => return Ok(r),
+                Err(err) => retries = self.handle_retry_error(retries, err).await?,
+            }
+        }
     }
 
     pub async fn batch_send_without_retry(
@@ -513,5 +513,23 @@ impl<R: RetryStrategy> Client<R> {
             return Err(Error::JsonRpcError(err.clone()));
         }
         Ok(id)
+    }
+
+    async fn handle_retry_error(&self, mut retries: u32, err: Error) -> Result<u32, Error> {
+        if !self.retry.is_retriable(&err) {
+            return Err(err);
+        }
+        if retries < self.retry.max_retries(&err) {
+            match retries.checked_add(1) {
+                Some(i) if i < self.retry.max_retries(&err) => {
+                    retries = i;
+                }
+                _ => return Err(err),
+            };
+            tokio::time::delay_for(self.retry.delay(&err, retries)).await;
+            Ok(retries)
+        } else {
+            Err(err)
+        }
     }
 }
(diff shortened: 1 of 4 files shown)


client/json-rpc/src/async_client/tests.rs
```diff
@@ -36,18 +40,29 @@ macro_rules! assert_err {
...
+    setup_multi_requests(vec![(request, response)])
+}
+
+fn setup_multi_requests(inouts: Vec<((&'static str, Value), Value)>) -> Client<Retry> {
+    setup_with_server(
+        inouts
+            .iter()
+            .map(|((method, params), response)| {
+                (
+                    json!({"id": 1, "jsonrpc": "2.0", "method": method, "params": params}),
```
thefallentree, line 52: the id here should be different for each request.

client/json-rpc/src/async_client/client.rs
```diff
@@ -280,53 +376,61 @@ impl<R: RetryStrategy> Client<R> {
...
-    pub fn update_state(&self, resp_state: State) -> bool {
-        let mut state_writer = self.last_known_state.write().unwrap();
-        if let Some(state) = &*state_writer {
-            if &resp_state < state {
-                return false;
-            }
-        }
-        *state_writer = Some(resp_state);
-        true
+        Ok(resp)
```
thefallentree, line 434: I think you should add an req_id: id  here so you can verify  the id is same  for request and response. 

it is part of the jsonrpc protocol.

but in this case, if this client always send out id  0-X  for request in an batch you could also have an  "batch_validate()" wrapper that verify the response in an batch ,  I think you can also check to make sure responses all have same libra_timstamp  and libra_version etc , which may prevent bug from server side to affect client

client/json-rpc/src/async_client/tests.rs
xli: each request is same because they are standalone request, not batch request in this func.

client/json-rpc/src/async_client/client.rs
xli: updated: batch will check id, as response order may not match requests order.
for non-batch call, still ignores response id for now.

xli: validate don't need response id, as you find response by request id, hence once you know a response for a request in batch case, no need validate response id.

client/json-rpc/src/async_client/client.rs
```diff
@@ -280,53 +376,121 @@ impl<R: RetryStrategy> Client<R> {
...
+            .map(|r| {
+                Err(Error::unexpected_response_not_found_error(format!(
+                    "request: {}",
+                    r
+                )))
+            })
+            .collect::<Vec<_>>();
+        for resp in rpc_resps {
+            if let Some(ref id) = resp.id {
+                if let Ok(index) = serde_json::from_value::<usize>(id.clone()) {
```
thefallentree, line 447: don't you want to throw errors if. fail to deserialize?

client/json-rpc/src/async_client/client.rs
```diff
@@ -280,53 +376,121 @@ impl<R: RetryStrategy> Client<R> {
...
+                    if index < result.len() {
+                        result[index] = self.validate(resp);
```
thefallentree, line 449: it is pretty dangerous to use response id as index to array, also that's why you needed this check,  I would just change the outer loop instead

for (i, resp) in rpc_resps.enumeration() { 

  // check i == index
result[i] = self.validate(resp) 
}

thefallentree: yeah, in protocol the response could be randomly ordered, I think we probably should use an hash map (id as key) instead.

client/json-rpc/src/async_client/client.rs
xli: it means response id does not match request id, as request id is index of requests array

client/json-rpc/src/async_client/client.rs
xli: it's same.
the primary thing here is what is request id, as interface accept an array of requests, we should expect response is an array and index matches.
anything does not match request id, should not be put in result, although, maybe put into result is better for checking later, but I don't think this is something we really need care too much, as if server side has bug causes response id != request id, it will be a huge problem and basically you can't use batch send anymore.

xli: also note the result is pre-init with Err(response not found error with request data), hence no response in result for a specific request will report error

client/json-rpc/src/async_client/client.rs
xli: the problem with reporting error is, we don't know where to put the unmatched response, as batch response may contain valid response. and id mismatched response can't be put into result.

client/json-rpc/src/async_client/client.rs
xli: I think the question here is when one response has invalid id, do we fail the whole batch send call, or we return valid ones and ignore responses that have invalid/mismatched ids.
Either way is fine to me.

thefallentree: IMHO to be on the safe side, we should really start with an all-or-nothing approach ..

be careful if the response contains a invalid id (could be anything) , you don't want it to trigger client crash though.

client/json-rpc/src/async_client/client.rs
```diff
@@ -248,85 +348,175 @@ impl<R: RetryStrategy> Client<R> {
...
+
+        let rpc_resps: Vec<JsonRpcResponse> = self.send_json_request(json![json_requests]).await?;
+        if rpc_resps.len() != requests.len() {
+            return Err(Error::unexpected_invalid_batch_response(rpc_resps));
+        }
+
+        let mut result: Vec<JsonRpcResponse> = rpc_resps.clone();
+        for resp in rpc_resps {
+            let id = self.validate(&resp, 0, requests.len() - 1)?;
+            result[id] = resp;
```
thefallentree, line 442: this won't catch issues where two response have same id . I still think the logic would be easier if you were to to use an hashmap then fill result by looking up id as key

```
resp_maps = HashMap::new() 
for resp in. rpc_resps {
 id = self.validate(0, requests.len() -1)?
 if resp_maps.contains(id) {
    // error
 } 
}
for i in 0..requests.len()-1 {
   result.append(resp_maps.get(i)?)
}
```

thefallentree (approved):
looks great! Thanks!

PR #6641, opened 2020-11-03
[sccache] anonymous reads must be from the s3-us-west-2.amazonaws.com endpoint (aws force redirects)

rexhoffman:
## Motivation

Noticed pr build times were longer than they should be, debugged, found sccache was failing on every read of the s3 bucket, so time for a fix.

The failure was that each response was a redirect to the http://s3-us-west-2.amazonaws.com/ci-artifacts.libra.org/

Yes

## Test Plan

Local debugging.   While calling s3 in public mode verified the correct locations were being called.

```
DEBUG 2020-11-03T00:03:47Z: rusoto_core::request: Full request: 
 method: GET
 final_uri: https://s3-us-west-2.amazonaws.com/ci-artifacts.libra.org/sccache/libra/1/8/b/18ba6a2b9eb67729ce07f22b9ea4f9a4738928f3d825b90f0d919c23bd8fc6ed
Headers:
```

## Related PRs

None

### DIFF
--- a/devtools/x/src/cargo.rs
+++ b/devtools/x/src/cargo.rs
@@ -237,15 +237,15 @@ impl Cargo {
 
         // once all the arguments are added to the command we can log it.
         if log {
-            self.env_additions.iter().for_each(|t| {
-                if let Some(env_val) = t.1 {
-                    if SECRET_ENVS.contains(&t.0.to_str().unwrap_or_default()) && t.1.is_some() {
-                        info!("export {:?}=********", t.0);
+            self.env_additions.iter().for_each(|(name, value_option)| {
+                if let Some(env_val) = value_option {
+                    if SECRET_ENVS.contains(&name.to_str().unwrap_or_default()) {
+                        info!("export {:?}=********", name);
                     } else {
-                        info!("export {:?}={:?}", t.0, env_val);
+                        info!("export {:?}={:?}", name, env_val);
                     }
                 } else {
-                    info!("unset {:?}", t.0);
+                    info!("unset {:?}", name);
                 }
             });
             info!("Executing: {:?}", &self.inner);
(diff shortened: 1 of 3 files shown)


devtools/x/src/cargo.rs
```diff
@@ -240,7 +240,7 @@ impl Cargo {
             self.env_additions.iter().for_each(|t| {
                 if let Some(env_val) = t.1 {
                     if SECRET_ENVS.contains(&t.0.to_str().unwrap_or_default()) && t.1.is_some() {
-                        info!("export {:?}=********", env_val);
```
sunshowers, line 243: A more idiomatic way to write this would be to destructure:

```rust
self.env_additions.iter().for_each(|(key, value)| {
```

sunshowers (approved):

x.toml
```diff
@@ -12,12 +12,14 @@ flags = "-Zfeatures=all"
 bucket = "ci-artifacts.libra.org"
 prefix = "sccache/libra/"
 public = true
+region = "us-west-2"
```
rexhoffman, line 15: need when authenticating agains the endpoint below.

x.toml
```diff
@@ -12,12 +12,14 @@ flags = "-Zfeatures=all"
...
+endpoint = "https://s3-us-west-2.amazonaws.com"
```
rexhoffman, line 16: needed when not using authentication as normally aws would redirect an unauthenticated read of https://ci-artifacts.libra.org/ to https://ci-artifacts.libra.org.s3-us-west-2.amazonaws.com/ which doesn't have a valid cert.  So instead forcing the use of https://s3-us-west-2.amazonaws.com/ci-artifacts.libra.org/ as the bucket location.

devtools/x/src/cargo.rs
```diff
@@ -237,15 +237,15 @@ impl Cargo {
 
         // once all the arguments are added to the command we can log it.
         if log {
-            self.env_additions.iter().for_each(|t| {
-                if let Some(env_val) = t.1 {
-                    if SECRET_ENVS.contains(&t.0.to_str().unwrap_or_default()) && t.1.is_some() {
```
rexhoffman, line 242: line 242,  `&& t.1.is_some()` is redundant due to the if let on 241.

PR #6649, opened 2020-11-03
[e2e-testsuite] Remove golden file checker for genesis.

runtian-zhou:
## Motivation

Remove the golden file checker for genesis. The reason is that we should view genesis as the initial writeset that is needed to spawn a network from MASTER branch head, instead of the release branch. Thus what's in the genesis should not be fixed and thus does not require a golden checker.

Yes

## Test Plan

cargo test 


dariorussi (approved):
thanks!

PR #6653, opened 2020-11-04
T78662332 - DesignatedDealer::add_tier(...) should be private

moeziniaf:
add_tier can only be called when the address has a TierInfo resource.  This resource can only be created in add_currency(...), which automatically sets 4 tiers. Since a DD can only have 4 tiers (checked against a constant), there is no case where add_tier will succeed if called from outside the module.


sblackshear (approved):
Thanks!

mickayz (approved):
LGTM

moeziniaf:
Any idea why `execute_genesis_and_drop_other_transaction` is failing?

PR #6654, opened 2020-11-04
[network] fix small issue with network rpc fuzzer

phlip9:
+ Sampling a new self-peer-id every fuzzer iteration causes metrics to
blow up after ~ 200,000 iters.

+ After this fix, the fuzzer runs pretty much indefinitely. I got up to
at least 16,000,000 iters locally with constant memory.

```
INFO: Seed: 1165517548
190    INITED cov: 611 ft: 611 corp: 1/1b exec/s: 0 rss: 143Mb
8388608        pulse  cov: 612 ft: 613 corp: 3/555b lim: 4096 exec/s: 4459 rss: 1077Mb
16777216       pulse  cov: 612 ft: 613 corp: 3/555b lim: 4096 exec/s: 4464 rss: 1078Mb
```

### DIFF
--- a/network/src/protocols/rpc/fuzzing.rs
+++ b/network/src/protocols/rpc/fuzzing.rs
@@ -29,7 +29,8 @@ const MAX_UVI_PREFIX_BYTES: usize = 19;
 const MAX_SMALL_MSG_BYTES: usize = 32;
 const MAX_MEDIUM_MSG_BYTES: usize = 280;
 
-const MOCK_PEER_ID: PeerId = PeerId::ZERO;
+const MOCK_PEER_ID_1: PeerId = PeerId::new([1u8; PeerId::LENGTH]);
+const MOCK_PEER_ID_2: PeerId = PeerId::new([2u8; PeerId::LENGTH]);
 const MOCK_PROTOCOL_ID: ProtocolId = ProtocolId::ConsensusRpc;
 
 #[test]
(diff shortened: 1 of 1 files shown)


network/src/protocols/rpc/fuzzing.rs
```diff
@@ -63,8 +64,8 @@ pub fn generate_corpus(gen: &mut ValueGenerator) -> Vec<u8> {
 
 // Fuzz the inbound rpc protocol.
 pub fn fuzzer(data: &[u8]) {
-    let network_context = NetworkContext::mock();
-    let connection_metadata = ConnectionMetadata::mock(MOCK_PEER_ID);
+    let network_context = NetworkContext::mock_with_peer_id(MOCK_PEER_ID_1);
+    let connection_metadata = ConnectionMetadata::mock(MOCK_PEER_ID_2);
```
gregnazario, line 68: Interesting, so Prometheus counters blow up from having high cardinality of peer ids in the tests?

I wonder if we want that in this kind of testing in general, or to find a way to shut off the metrics system for things like fuzzing.

I think though, in general, testing with mock hardcoded values like this is probably a better strategy, unless we're trying to specifically fuzz the peer ids

phlip9: Yep, every unique {counter name, label set} adds an entry to the global metrics registry which never goes away.

I don't think this is actually a problem necessarily. The issue here is non-representative usage causing metrics blowup. Specifically, sampling a new `NetworkContext::mock()` every _iteration_ is like testing if our node software can handle running 200,000 concurrent network stacks without emitting too many metrics, which is obviously not representative of actual usage where we have a O(1) # of networks.

In contrast, there's an argument to be made that it's actually _good_ to sample different peer id's for the `ConnectionMedadata` since that would effectively assert that we don't have any metrics specifically tied to the remote peer id.

gregnazario (approved):

mimoo:
@sausagee do we need to cherry-pick this so we can fuzz it?

Issue #6644, opened 2020-11-03
[Bug] Memory usage increases with chain size


mgorven:
# 🐛 Bug

The RSS memory usage of validators and fullnodes increases as the size of the chain increases. This chart shows RSS of the Novi pre-mainnet validator along with the committed version:

![firefox-20201102-153016](https://user-images.githubusercontent.com/183923/98019426-95d5c400-1db6-11eb-9e5d-136e928aaef3.png)

After a process restart memory usage is back at the previous baseline within 1-2h, so this this isn't related to process uptime. It also happens on fullnodes.

## To reproduce

Run a public fullnode against pre-mainnet, mainnet. Or run a k8s testnet with cluster-test to generate a large chain state.

## Expected Behavior

Memory usage should not continually increase.

## System information

release-1.0_41e23b4aff

davidiw:
@lightmark / @msmouse, does this make sense? I would imagine that as our merkle trees become larger, they will take more RSS. Also if we use the premise that RSS increases with the chain size, then it most likely is storage related. Do we have any info about how storage uses RSS over time? Is there max?

mgorven:
Managed to get `jeprof` working by installing inside the running container.
[jeprof.pdf](https://github.com/libra/libra/files/5490512/jeprof.pdf)

lightmark:
I guess it about default option keeps all the tables open. Let's try to limit it.

PR #6656, opened 2020-11-04
[schemadb] put max_open_files into storage-config

lightmark:
The number of files keeps increasing so that the RSS increases unlimitedly due to index/filter_block per file.

I'll later refactor it.

## Motivation

https://github.com/libra/libra/issues/6644

### DIFF
--- a/config/src/config/storage_config.rs
+++ b/config/src/config/storage_config.rs
@@ -8,6 +8,29 @@ use std::{
     path::PathBuf,
 };
 
+/// Port selected RocksDB options for tuning underlying rocksdb instance of LibraDB.
+/// see https://github.com/facebook/rocksdb/blob/master/include/rocksdb/options.h
+/// for detailed explanations.
+#[derive(Copy, Clone, Debug, Deserialize, PartialEq, Serialize)]
+#[serde(default, deny_unknown_fields)]
+pub struct RocksdbConfig {
+    pub max_open_files: i32,
+    pub max_total_wal_size: u64,
+}
+
+impl Default for RocksdbConfig {
+    fn default() -> Self {
+        Self {
+            // Set max_open_files to 10k instead of -1 to avoid keep-growing memory in corridance
+            // with the number of files.
+            max_open_files: 10_000,
+            // For now we set the max total WAL size to be 1G. This config can be useful when column
+            // families are updated at non-uniform frequencies.
+            max_total_wal_size: 1u64 << 30,
+        }
+    }
+}
+
 #[derive(Clone, Debug, Deserialize, PartialEq, Serialize)]
 #[serde(default, deny_unknown_fields)]
 pub struct StorageConfig {
@@ -22,6 +45,8 @@ pub struct StorageConfig {
     data_dir: PathBuf,
     /// Read, Write, Connect timeout for network operations in milliseconds
     pub timeout_ms: u64,
+    /// Rocksdb-specific configurations
+    pub rocksdb_config: RocksdbConfig,
 }
 
 impl Default for StorageConfig {
@@ -36,6 +61,7 @@ impl Default for StorageConfig {
             data_dir: PathBuf::from("/opt/libra/data"),
             // Default read/write/connection timeout, in milliseconds
             timeout_ms: 30_000,
+            rocksdb_config: RocksdbConfig::default(),
         }
     }
 }
--- a/consensus/src/consensusdb/mod.rs
+++ b/consensus/src/consensusdb/mod.rs
@@ -18,7 +18,7 @@ use consensus_types::{block::Block, quorum_cert::QuorumCert};
 use libra_crypto::HashValue;
 use libra_logger::prelude::*;
 use schema::{BLOCK_CF_NAME, QC_CF_NAME, SINGLE_ENTRY_CF_NAME};
-use schemadb::{ReadOptions, SchemaBatch, DB, DEFAULT_CF_NAME};
+use schemadb::{Options, ReadOptions, SchemaBatch, DB, DEFAULT_CF_NAME};
 use std::{collections::HashMap, iter::Iterator, path::Path, time::Instant};
 
 pub struct ConsensusDB {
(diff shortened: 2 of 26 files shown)


mgorven:
How complex is it to make this configurable? I'm just thinking there might be unintended consequences (reduced read perf?) and we might want to tune this in config.

lightmark:
It is easy. But I guess in the future we may want a option files exclusively for rocksdb and here `max_open_files` is just one item.

davidiw:
> @mgorven It is easy. But I guess in the future we may want an option file exclusively for rocksdb and `max_open_files` is just one item in several options.

I think it is okay to create a struct in the storage config call rocks config that way you don't have to worry about this.

msmouse:
The memory consumption 1% of the DB size on disk, which is well affordable and most probably worth it if people actually query the full history on fullnodes. And I guess we are not too sure in or two years we will actually hit a larger DB in production.

I have no problem making it an option that's tunable (okay, to be honest I'd avoid if you ask me), but I'd make our default value a pretty large number (5k or 10k).

@mgorven  How much memory do you think we can afford for this purpose?

mgorven:
> I have no problem making it an option that's tunable (okay, to be honest I'd avoid if you ask me), but I'd make our default value a pretty large number (5k or 10k).
> 
> @mgorven How much memory do you think we can afford for this purpose?

I agree that we can go larger than 1k (based on the current test fullnode which hasn't exceeded 1G). The lowest memory allocation currently is 5.25G (on Azure). We need some space for the OS page cache and transaction handling under load, so somewhere around 3G should be fine.

Given the large impact this has on memory usage I think this needs to be configurable. For public fullnodes users may want to reduce memory usage, and if they're only querying recent data setting a low limit shouldn't affect read performance.

mgorven:
I think 10k is a good default.

lightmark:
Then let's add a struct in the option.

config/src/config/storage_config.rs
```diff
@@ -8,6 +8,33 @@ use std::{
     path::PathBuf,
 };
 
+#[derive(Copy, Clone, Debug, Deserialize, PartialEq, Serialize)]
+#[serde(default, deny_unknown_fields)]
+pub struct RocksdbConfig {
+    /// Port selected RocksDB options for tuning underlying rocksdb instance.
+    /// see https://github.com/facebook/rocksdb/blob/master/include/rocksdb/options.h
+    /// for detailed explanations.
```
msmouse, line 16: should be before the whole struct?

config/src/config/storage_config.rs
```diff
@@ -8,6 +8,33 @@ use std::{
...
+    pub create_if_missing: bool,
+    pub create_if_missing_column_families: bool,
```
msmouse, line 18: `open()`, `open_readonly()` and `open_as_secondary()` use different values for these options, I don't think we should pass them in?

config/src/config/storage_config.rs
```diff
@@ -8,6 +8,33 @@ use std::{
...
+    pub max_open_files: i32,
+    pub max_total_wal_size: u64,
+}
+
+impl Default for RocksdbConfig {
+    fn default() -> Self {
+        Self {
+            create_if_missing: true,
+            create_if_missing_column_families: true,
```
mgorven, line 27: When would these ever be changed?

config/src/config/storage_config.rs
```diff
@@ -8,6 +8,33 @@ use std::{
...
+#[derive(Copy, Clone, Debug, Deserialize, PartialEq, Serialize)]
+#[serde(default, deny_unknown_fields)]
+pub struct RocksdbConfig {
+    /// Port selected RocksDB options for tuning underlying rocksdb instance.
+    /// see https://github.com/facebook/rocksdb/blob/master/include/rocksdb/options.h
+    /// for detailed explanations.
+    pub create_if_missing: bool,
+    pub create_if_missing_column_families: bool,
+    pub max_open_files: i32,
+    pub max_total_wal_size: u64,
```
davidiw, line 20: what's a wal?

consensus/src/consensusdb/mod.rs
```diff
@@ -36,7 +36,10 @@ impl ConsensusDB {
 
         let path = db_root_path.as_ref().join("consensusdb");
         let instant = Instant::now();
-        let db = DB::open(path.clone(), "consensus", column_families)
+        let mut opts = Options::default();
+        opts.create_if_missing(true);
+        opts.create_missing_column_families(true);
```
davidiw, line 41: these default as true, why are we modifying it?

storage/libradb/src/lib.rs
```diff
@@ -93,6 +94,15 @@ fn error_if_too_many_requested(num_requested: u64, max_allowed: u64) -> Result<(
...
 
+fn gen_rocksdb_options(config: &RocksdbConfig) -> Options {
+    let mut db_opts = Options::default();
+    db_opts.create_if_missing(config.create_if_missing);
+    db_opts.create_missing_column_families(config.create_if_missing_column_families);
+    db_opts.set_max_open_files(config.max_open_files);
+    db_opts.set_max_total_wal_size(config.max_total_wal_size);
+    db_opts
+}
+
```
davidiw, line 105: can we use an std::convert::Into instead?

config/src/config/storage_config.rs
lightmark: write_ahead_log

config/src/config/storage_config.rs
lightmark: for val/fullnode, they don't. But for my tool to access db, or open_read_only(), I don't want to create a db if it doesn't exist there. It should be an error "DB not found"

consensus/src/consensusdb/mod.rs
lightmark: default is false. Here ConsensusDB take raw schemaDB so it doesn't do the conversion.

storage/libradb/src/lib.rs
lightmark: I don't think so. Please tell me where it should be. I weighed different solutions. 
the `Into` trait must be implemented in either crate, rocksdb binding (which is third party) or libra-config (do you want to incorporate rocksdb dependency into `libra-config`?)

msmouse (approved):

PR #6660, opened 2020-11-05
[cli/libra framework] Remove Offer module from being published in genesis

tzakian:
Moves the `Offer` module to the nursery so it is no longer published in genesis. Tests have been moved to a Move CLI smoke test.

This move doesn't have any effects to scripts or on-chain modules as it wasn't being used. 

This PR would be a proposed cherry-pick to release.


mickayz (approved):

sblackshear (approved):
Thanks Tim!! And good call on moving the tests to CLI expected value tests.

junkil-park:
> One thing I was wondering: do the verifier tests know to run on `stdlib/nursery`? CC @junkil-park @emmazzz

(1) Yes. `cargo test -p move-prover` verifies each .move files under `language/stdlib/` and its sub-directories. So, it should verify the .move files under `language/stdlib/nursery` as well.

(2) `language/stdlib/nursery` is not included in the dependency list. If a module "uses" one of the nursery modules, Prover will not connect them, but raise an error (e.g., Invalid 'use'. Unbound module: ...).

PR #6663, opened 2020-11-06
[Tooling] Crucial bug fixes and better debugging support for operators

JoshLind:
## What this affects:

This PR affects the genesis tool (./libra-genesis-tool) and the operational tool (./libra-operational-tool), both located in libra/config/management.

## Why is this critical:

This PR provides 4 critical bug fixes for these tools:
1. First, it fixes a bug related to whitespace handling in cryptographic key files when performing crucial operations, such as creating a new validator account or a validator operator account. See here for more information: https://github.com/libra/libra/pull/6548
2. Second, it provides a way to disable network address validation checks when performing operations that update the on-chain ValidatorConfig of a validator account. This avoids spurious errors that currently prevent an operator from updating their ValidatorConfig (including key rotations!). See here for more information: https://github.com/libra/libra/pull/6571
3. Third, it prevents the operational tool from panicking when it fails to decrypt a validator network address held in a ValidatorConfig on-chain. Instead, the tool should still print out the other fields of the ValidatorConfig. See here for more information: https://github.com/libra/libra/pull/6629
4. Finally, it provides a way for the operational tool to specify whether or not to update the genesis waypoint when executing the "insert-waypoint" command. Currently, "insert-waypoint" overwrites both the waypoint and genesis waypoint in storage. This can break validator full nodes in the deployment environments. See here for more information: https://github.com/libra/libra/pull/6638

## What is the workaround (even if it is unintuitive / horrible):

The workarounds for each of these issues are:
1. Try and strip off any whitespace from a cryptographic key file before running a command. However, whitespace is secretly appended to any file when created by hand (e.g., vim appends a newline character). You could try and create these files programmatically to avoid whitespace at all, however, for most users, this isn't great. I've yet to find a good way personally to strip these hidden whitespace characters manually..
2. Ensure that network address validation doesn't fail. Normally, this shouldn't happen. However, it can happen if the network address being stored in the ValidatorConfig can't be looked up by the person running the tool. For example, if you're running your node in AWS, the network address won't be accessible locally, so the update will fail and the operator won't be able to change their ValidatorConfig (which prevents key rotations..).
3. Ensure that decryption doesn't fail. Normally, this shouldn't happen if the person using the tool has a storage backend that contains the shared network encryption key. However, this key is not currently distributed to folks who aren't operators, (e.g., the blockchain oncall). So to ensure these users aren't prevented from retrieving a ValidatorConfig on-chain, they need to be given access to a secure storage with this key (or create one themselves, which isn't really feasible..)
4. Try and connect manually to secure storage (Vault), and set the genesis waypoint by hand. However, because of the way waypoints are stored in Vault (using json), this isn't super straight forward, so it's likely inaccessible for many operators.

## Why are we not just cherry-picking the individual commits for the fixes?

This PR contains 13 commits. Ideally, we'd only cherry-pick the commits that make the bug fixes. However, the genesis and operational tools in master have diverged significantly from the release branch, so re-implementing these fixes and/or handling the merge conflicts would increase the risk of further errors significantly. Moreover, many of the commits in this PR include updates to the smoke tests to ensure these changes/fixes are truly correct and we don't hit regressions. 

Here's a breakdown of the 13 commits in this PR:
- 4 of the commits are direct bug fixes to the issues above.
- 3 commits update the smoke tests to ensure correct functionality regarding the fixes and avoid regressions.
- 3 commits add new features for better operator tooling debugging (e.g., the ability to print out keys, accounts and waypoints held in an operators Vault). These are all very useful features to have to help operators resolve issues.
- The 3 remaining commits include small code refactors and updates to the README. 

## How have we tested these changes?

Thankfully, all of the changes in this PR are tested using smoke tests. Our smoke tests offer pretty comprehensive coverage for the fixes in this PR. Moreover, the tool has also been tested manually, by pointing it at several different k8 test deployments and performing operations.

### DIFF
--- a/config/management/genesis/README.md
+++ b/config/management/genesis/README.md
@@ -1,8 +1,8 @@
 # Libra Config Manager
 
-The `libra-genesis-tool` provides a tool for the genesis ceremony of the Libra blockchain as well as adding new validators to an existing blockchain. The functionality of the tool is dictated by the organization of nodes within the system:
+The `libra-genesis-tool` provides a tool for the genesis ceremony of the Libra blockchain. The functionality of the tool is dictated by the organization of nodes within the system:
 
-* A libra root account that maintains the set of validator owners and validator operators.
+* A libra root account that maintains the set of validator owners, validator operators, and the active validator set.
 * A treasury compliance account that maintains VASPs, DDs, and other related topics.
 * Validator owners (OW) that have accounts on the blockchain. These accounts contain a validator configuration and specify a validator operator.
 * Validator operators (OP) that have accounts on the blockchain. These accounts have the ability to manipulate validator configuration.
--- a/config/management/genesis/src/storage_helper.rs
+++ b/config/management/genesis/src/storage_helper.rs
@@ -127,6 +127,7 @@ impl StorageHelper {
                     path={path};\
                     namespace={validator_ns}
                 --waypoint {waypoint}
+                --set-genesis
             ",
             backend = DISK,
             path = self.path_string(),
--- a/config/management/genesis/src/validator_config.rs
+++ b/config/management/genesis/src/validator_config.rs
@@ -19,6 +19,8 @@ pub struct ValidatorConfig {
     fullnode_address: NetworkAddress,
     #[structopt(flatten)]
     shared_backend: SharedBackend,
+    #[structopt(long, help = "Disables network address validation")]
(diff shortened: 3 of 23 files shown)


sausagee (approved):

JoshLind:
Thanks @sausagee! 😄 

/land

PR #6664, opened 2020-11-06
[consensus] removing redundant info in log

mimoo:
my bad, some of the information was already there : o removing the redundant stuff

Screen Shot 2020-11-05 at 11 58 16 PM

on a positive side I managed to trigger the alerts in kibana!

### DIFF
--- a/consensus/src/round_manager.rs
+++ b/consensus/src/round_manager.rs
@@ -492,9 +492,6 @@ impl RoundManager {
 
         info!(
             self.new_log(LogEvent::ReceiveProposal).remote_peer(author),
-            proposal = %proposal,
-            block_epoch = proposal.epoch(),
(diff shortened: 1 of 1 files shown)


consensus/src/round_manager.rs
```diff
@@ -493,8 +493,6 @@ impl RoundManager {
         info!(
             self.new_log(LogEvent::ReceiveProposal).remote_peer(author),
             proposal = %proposal,
```
gregnazario, line 495: You should be able to remove this, because the 2 things you were missing, are below as block_hash and block_parent_hash.

mimoo: woop, very good point!

gregnazario (approved):

PR #6665, opened 2020-11-06
[logger] Opening logs for write+append, instead of only write

villesundell:
## Motivation

Logs used to be opened with write() only, meaning that if the logfile
already existed, logs would not be produced.

Changing write() to append() (which is effectively write() +
append()).

Yes.

## Test Plan

Ran the following command twice: `cargo run -p libra-node -- --test --config /tmp/efd6dc5656ee8edf92569c330b8de8bb/0/` and verified that:

- Second run also produced logs

- The logfile contained logs from the first run

## Related PRs
No related PRs.


bmwill (approved):
Looks good, thanks!
/land

PR #6671, opened 2020-11-07
[Consensus] Don't panic on failure to initialize LSR

JoshLind:
## Motivation

This PR updates consensus to prevent panicking when safety rules (LSR) fails to initialize due to the validator signer key not being found in secure storage for the current epoch. The reason for this change is to primarily handle cases where a validator node falls behind the head of the blockchain and gets blocked trying to catch up due to not having the old key in secure storage (e.g., due to key trimming).

To achieve this change, we update:
1. The initialize() method in safety rules to emit explicit error types for the validator key not being found, and the validator not being in the validator set.
2. Epoch manager to check for these new error types and only log an error in these cases (instead of previously panicking for the case of the validator key not being found). The idea here is that alerts can be set up when these errors are logged, to notify the operator about a failing initialize().
3. The initialize() method in safety rules now updates the waypoint before checking the key can be found for the validator. This allows safety rules to catch up with the blockchain despite not having old/lost keys.

Yes.

## Test Plan

All tests pass locally.

## Related PRs

None.

### DIFF
--- a/consensus/safety-rules/src/safety_rules.rs
+++ b/consensus/safety-rules/src/safety_rules.rs
@@ -247,44 +247,7 @@ impl SafetyRules {
             .cloned()
             .ok_or(Error::InvalidLedgerInfo)?;
 
-        let author = self.persistent_storage.author()?;
-        if let Some(expected_key) = epoch_state.verifier.get_public_key(&author) {
-            let curr_key = self.signer().ok().map(|s| s.public_key());
-            if curr_key != Some(expected_key.clone()) {
-                let consensus_key = self
-                    .persistent_storage
-                    .consensus_key_for_version(expected_key)
-                    .map_err(|e| {
-                        let error = Error::InternalError(format!(
-                            "Validator key not found: {:?}",
-                            e.to_string()
-                        ));
-                        error!(
-                            SafetyLogSchema::new(LogEntry::KeyReconciliation, LogEvent::Error)
-                                .error(&error),
-                        );
-
-                        self.validator_signer = None;
-                        error
-                    })?;
-
-                self.validator_signer = Some(ValidatorSigner::new(author, consensus_key));
-            }
-
-            debug!(
-                SafetyLogSchema::new(LogEntry::KeyReconciliation, LogEvent::Success),
-                "in set",
-            );
-        } else {
-            debug!(
-                SafetyLogSchema::new(LogEntry::KeyReconciliation, LogEvent::Success),
-                "not in set",
-            );
-            self.validator_signer = None;
-        }
-
         let current_epoch = self.persistent_storage.safety_data()?.epoch;
-
         if current_epoch < epoch_state.epoch {
             // This is ordered specifically to avoid configuration issues:
             // * First set the waypoint to lock in the minimum restarting point,
@@ -303,9 +266,42 @@ impl SafetyRules {
 
             info!(SafetyLogSchema::new(LogEntry::Epoch, LogEvent::Update).epoch(epoch_state.epoch));
         }
-        self.epoch_state = Some(epoch_state);
+        self.epoch_state = Some(epoch_state.clone());
 
-        Ok(())
+        let author = self.persistent_storage.author()?;
+        let expected_key = epoch_state.verifier.get_public_key(&author);
+        let initialize_result = match expected_key {
+            None => Err(Error::ValidatorNotInSet(author.to_string())),
+            Some(expected_key) => {
+                let current_key = self.signer().ok().map(|s| s.public_key());
+                if current_key == Some(expected_key.clone()) {
+                    debug!(
+                        SafetyLogSchema::new(LogEntry::KeyReconciliation, LogEvent::Success),
+                        "in set",
+                    );
+                    Ok(())
+                } else {
+                    match self
+                        .persistent_storage
+                        .consensus_key_for_version(expected_key)
+                    {
+                        Err(error) => Err(Error::ValidatorKeyNotFound(error.to_string())),
+                        Ok(consensus_key) => {
+                            self.validator_signer =
+                                Some(ValidatorSigner::new(author, consensus_key));
+                            Ok(())
+                        }
+                    }
+                }
+            }
+        };
+        initialize_result.map_err(|error| {
+            info!(
+                SafetyLogSchema::new(LogEntry::KeyReconciliation, LogEvent::Error).error(&error),
+            );
+            self.validator_signer = None;
+            error
+        })
     }
 
     fn guarded_construct_and_sign_vote(
--- a/consensus/safety-rules/src/tests/suite.rs
+++ b/consensus/safety-rules/src/tests/suite.rs
@@ -664,7 +664,10 @@ fn test_validator_not_in_set(safety_rules: &Callback) {
     proof
         .ledger_info_with_sigs
         .push(a2.block().quorum_cert().ledger_info().clone());
-    safety_rules.initialize(&proof).unwrap();
+    assert!(matches!(
+        safety_rules.initialize(&proof),
+        Err(Error::ValidatorNotInSet(_))
+    ));
 
     let state = safety_rules.consensus_state().unwrap();
     assert_eq!(state.in_validator_set(), false);
--- a/consensus/src/epoch_manager.rs
+++ b/consensus/src/epoch_manager.rs
@@ -40,7 +40,7 @@ use libra_types::{
     on_chain_config::{OnChainConfigPayload, ValidatorSet},
 };
 use network::protocols::network::Event;
-use safety_rules::SafetyRulesManager;
+use safety_rules::{Error, SafetyRulesManager};
 use std::{cmp::Ordering, sync::Arc, time::Duration};
 
 /// RecoveryManager is used to process events in order to sync up with peer if we can't recover from local consensusdb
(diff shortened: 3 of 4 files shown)


consensus/safety-rules/src/tests/suite.rs
```diff
@@ -664,7 +664,10 @@ fn test_validator_not_in_set(safety_rules: &Callback) {
     proof
         .ledger_info_with_sigs
         .push(a2.block().quorum_cert().ledger_info().clone());
-    safety_rules.initialize(&proof).unwrap();
+    match safety_rules.initialize(&proof) {
```
zekun000, line 667: you can use matches! macro

zekun000 (approved):

consensus/safety-rules/src/tests/suite.rs
JoshLind: Ooh, nice, thanks! Didn't know we had matches on stable 😄

JoshLind:
Thanks @zekun000! 😄 

/land

consensus/src/epoch_manager.rs
```diff
@@ -304,9 +304,25 @@ impl EpochManager {
...
             MetricsSafetyRules::new(self.safety_rules_manager.client(), self.storage.clone());
-        safety_rules
-            .perform_initialize()
-            .expect("Unable to initialize SafetyRules");
+        match safety_rules.perform_initialize() {
+            Ok(()) => { /* Expected */ }
+            Err(Error::ValidatorKeyNotFound(error)) => {
+                error!(
+                    error = error,
+                    "Unable to find the validator consensus key during initialize",
```
aching, line 312: It's unclear to me what happens in this case.  Should the operator just let the validator keep running and then it might catch up to the point where the history matches the keys in vault and then restart the validator?

consensus/src/epoch_manager.rs
```diff
@@ -304,9 +304,25 @@ impl EpochManager {
...
+                );
+            }
+            Err(Error::ValidatorNotInSet(error)) => {
```
aching, line 315: Same questions as above.

consensus/src/epoch_manager.rs
zekun000: It's basically the same as sync_only mode. It doesn't require restart since it tries to initialize every epoch.

JoshLind: Yes, so the current plan is to add an alert for this error, so that when it fires, operators will be aware and manually inspect the error. Upon manual inspection, there's two cases:
1. Assuming the validator is simply behind the blockchain head, the operator will let the validator keep running until it catches up with the current head of the blockchain, at which point LSR will be able to get the keys from storage and things will continue as normal (no restart required).
2. If something is very broken (e.g., secure storage is dead), the operator will need to escalate and try to figure out what's gone wrong.

From previous experience, most operators who run into 1 are already experiencing problems (e.g., they've tried to migrate a node across operators, failed in doing an upgrade, or seen a previous failure which meant their node is significantly behind).

'cc @davidiw who is going to be adding the alerts for this change.

consensus/src/epoch_manager.rs
JoshLind: The plan here is to also issue an alert, so that the operators are aware their nodes are not in the current validator set. Note that currently, if we hit this case in production, all that happens is that we log "not in set" in LSR and set the validator signer to nothing -- things continue otherwise, and the operator is probably none the wiser.

'cc @davidiw

consensus/src/epoch_manager.rs
aching: Thanks for the explanation.

Regarding 1,  we could use the current time compared to the blockchain head time to give a hint at least of how far behind it is - e.g. it's more of an info if it's just really out-of-date compared to being an error.  Also, it might be nice to add a comment that this is a potentially transient error....now I'm starting to wonder if this should be warn rather than error since it should self-heal.

Regarding 2, if secure storage is dead, shouldn't we just propagate that error rather than ValidatorKeyNotFound?

consensus/safety-rules/src/safety_rules.rs
```diff
@@ -303,9 +266,42 @@ impl SafetyRules {
 
             info!(SafetyLogSchema::new(LogEntry::Epoch, LogEvent::Update).epoch(epoch_state.epoch));
         }
-        self.epoch_state = Some(epoch_state);
+        self.epoch_state = Some(epoch_state.clone());
 
-        Ok(())
+        let author = self.persistent_storage.author()?;
```
davidiw, line 271: This would flag that secure storage is unavailable, we don't mask the error in safety rules. @JoshLind / @aching

consensus/safety-rules/src/safety_rules.rs
```diff
@@ -303,9 +266,42 @@ impl SafetyRules {
...
+                        SafetyLogSchema::new(LogEntry::KeyReconciliation, LogEvent::Success),
+                        "in set",
+                    );
+                    Ok(())
+                } else {
+                    match self
+                        .persistent_storage
+                        .consensus_key_for_version(expected_key)
+                    {
+                        Err(error) => Err(Error::ValidatorKeyNotFound(error.to_string())),
```
davidiw, line 288: Though we slightly mask it here.

consensus/src/epoch_manager.rs
davidiw: This should simply be an Error that SafetyRules failed to be initialized as there can be lots of reasons why it didn't.

consensus/src/epoch_manager.rs
davidiw: I think there's something already in the run book as is, but we never implemented the alerts as we said that the validator would be crash looping, which it won't any more.

davidiw (commented):
Overall PR looks good but the internal message on consensus isn't quite right and maybe we want to reduce how much we hide the error for retrieving the key... such that if the value isn't present return key not found otherwise return the unexpected error that wraps secure storage error?

consensus/src/epoch_manager.rs
aching: My concern is that an operator that sees this error won't understand that on a subsequent epoch that initialize might succeed from this error message.  Error generally means something bad happened and the operator might need to take action, but I don't see how this message would communicate what needs to be done or not done.  It would probably just be escalated to a SME in this area when it doesn't need to be with more clear documentation.  I don't mind more verbose error messages as they are supposed to be quite uncommon if we can help operators understand what is going on and what actions to take (even if it's just pointing at other documentation).

consensus/src/epoch_manager.rs
JoshLind: Yeah, we'll need to update this.

consensus/safety-rules/src/safety_rules.rs
JoshLind: True.. although this is exactly what we do today -- we catch the error and say "Validator key not found: {error}"

consensus/src/epoch_manager.rs
JoshLind: > My concern is that an operator that sees this error won't understand that on a subsequent epoch that initialize might succeed from this error message... 

My general opinion here is that the alert triggered for this error should contain all the information for the operator to react to it, i.e., they won't really know if an error is logged in LSR, unless an alert fires, right? (Operators won't be tailing the logs in practice 😄) In which case I was envisioning us having the alert that triggers here point directly to the runbook, which should entail exactly how to react to the alert. Note: I don't mind having verbose error logs, but I was always under the impression that our runbooks were the place for this information to be (hence why we link to runbooks in alerts?)

> Regarding 1, we could use the current time compared to the blockchain head time to give a hint at least of how far behind it is..

This would be great, but it's difficult to identify how far we are behind the head (without having another node to query).. Hence why, it requires some sort of human reasoning. Unless we think it's practical to be pinging other validators for this information? (I'm personally not in favor of this)

> now I'm starting to wonder if this should be warn rather than error since it should self-heal.

I think this can correctly be dealt with at the alert level, e.g., if the alert fires for a few minutes, issue a warning (it could still be catching up, and the runbook should tell the operator that it might fix itself if the operator believes the node to be behind). If the alert continues to fire for 30 minutes+, then increase the alert warning and say that they're unlikely to catch up and it requires intervention?

JoshLind: Note: once we've settled on the right approach that makes everyone happy, I'll definitely follow up with the improvements/fixes we want for these changes. Next time I also won't be so fast as to land soon after an approval 😄

aching: I agree on the point about notification via an alarm about an issue and a runbook is a more detailed place to describe the process of resolving it.  That being said, it would be helpful for an error to convey the seriousness of the issue and implications of it from a standalone point of view of someone who views it.  

By the way, this PR definitely is an improvement, I'm just thinking what else we could do to make it clearer to someone who sees this message.  If I saw it, the first thing I would do is start reading the code trying to extract what to do next.  Personally, I'd prefer to see something more like "Unable to find the validator consensus key during initialize() for epoch XXX and will not participate in consensus.  Currently, version=YYY, blockchain time=ZZZ, epoch=WWW, and will continue syncing until the validator consensus key matches for a future epoch."

JoshLind: Okay, thanks @aching! I can see the benefit in conveying the urgency/context a little more clearly, like you say. Although, I do have some concerns returning that exact error message, but let's debate it in a follow up PR that cleans things up here 😄

PR #6672, opened 2020-11-09
[network] Decrease Healthchecker logs

gregnazario:
The healthchecker logs don't provide much info, so I've turned them down to
trace for testing purposes only.


phlip9 (approved):
LGTM!

PR #6674, opened 2020-11-09
[libradb][jf] exports node stats every version

lightmark:
## Motivation

Change node_stats to be reported by each version instead of the committing version only.

With this change, the pruning will not break stats. But we may need more refactoring to make backup & restore perfectly works with it.

This also eases getting the available state versions for librarian.

Y

## Test Plan

change the existing coverage.

### DIFF
--- a/storage/libradb/src/state_store/mod.rs
+++ b/storage/libradb/src/state_store/mod.rs
@@ -77,24 +77,22 @@ impl StateStore {
         let (new_root_hash_vec, tree_update_batch) =
             JellyfishMerkleTree::new(self).put_blob_sets(blob_sets, first_version)?;
 
-        cs.counter_bumps.bump(
-            LedgerCounter::NewStateNodes,
-            tree_update_batch.node_batch.len(),
-        );
-        cs.counter_bumps.bump(
-            LedgerCounter::NewStateLeaves,
-            tree_update_batch.num_new_leaves,
-        );
+        let num_versions = new_root_hash_vec.len();
+        assert_eq!(num_versions, tree_update_batch.node_stats.len());
+
+        tree_update_batch
+            .node_stats
+            .iter()
+            .enumerate()
+            .for_each(|(i, stats)| {
+                let counter_bumps = cs.counter_bumps(first_version + i as u64);
+                counter_bumps.bump(LedgerCounter::NewStateNodes, stats.new_nodes);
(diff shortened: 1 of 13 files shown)


storage/libradb/src/state_store/mod.rs
```diff
@@ -77,24 +77,25 @@ impl StateStore {
...
+
+        tree_update_batch
+            .node_stats
+            .iter()
+            .enumerate()
+            .for_each(|(i, stats)| {
+                let counter_bumps = cs
+                    .counter_bumps
+                    .entry(first_version + i as u64)
+                    .or_insert_with(LedgerCounterBumps::new);
```
msmouse, line 91: This is repeated so much I'm thinking properly we should let LedgerCounterBumps or another type to manage an internal hash map, or at least short cut this call (`let mut bumps = cs.counter_bumps.for_version(v)`)

storage/libradb/src/system_store/test.rs
```diff
@@ -40,38 +58,35 @@ fn test_inc_ledger_counters() {
...
-        bumps.bump(LedgerCounter::EventsCreated, 20);
+        let bumps = create_bumps_map(2, vec![5; 9]);
 
         let counters = bump_ledger_counters(
             store, 2,  /* first_version */
             10, /* last_version */
             bumps,
         );
-        assert_eq!(counters.get(LedgerCounter::EventsCreated), 30);
+        assert_eq!(counters.get(LedgerCounter::EventsCreated), 55);
```
msmouse, line 68: ?? thought this should be 24?

msmouse: nvm.

storage/libradb/src/system_store/test.rs
```diff
@@ -40,38 +58,35 @@ fn test_inc_ledger_counters() {
...
-            store, 2, /* first_version */
+            store, 4, /* first_version */
             8, /* last_version */
             bumps,
         );
-        assert_eq!(counters.get(LedgerCounter::EventsCreated), 15);
+        assert_eq!(counters.get(LedgerCounter::EventsCreated), 35);
     }
-    // Base version missing, swallowing the error and adding to zeros.
+    // Base version and some entries are missing, swallowing the error and adding to zeros.
```
msmouse, line 81: could you add a test to show that if a counter at a version doesn't bump, it retains the same value as the last version?

msmouse (approved):

PR #6684, opened 2020-11-11
test for tracing changes

dariorussi:
try alternative to tracing where we do not print right away but collect value and print a summary at the end


davidiw:
Is this outdated? feel free to re-open.

PR #6690, opened 2020-11-12
[libra framework] Address some issues raised from audit

tzakian:
* Update comments and fix typos
* Added an additional assertion in `DualAttestation` to make sure that the `base_url` field has been set when checking a signature.
* Update names of some functions to make their effects/purpose clearer
* Change some error reason numbers so that they are sequential
* Updated `Libra::is_SCS_currency` to not call `assert_is_currency`. Refactored some specs to use this instead of `spec_is_SCS_currency`. 

Some tests needed to be updated to handle the. Golden files changed for `RecoveryAddress` tests since gas usage when up slightly.


mickayz (approved):
LGTM! Reviewed all updates to files in language/stdlib/modules/*.move

xli (approved):

PR #6691, opened 2020-11-12
[network] Decrease Healthchecker logs

gregnazario:
### Justification / Why we must have it for V1 launch
* This removes health checker logs, which are roughly a quarter of all logs, but don't serve any usable purpose.  This should decrease pressure on the logging system.

### Breaking nature
* No breaking changes, decreased health check logs, but they're never used

### Who does this affect?
* Logging in all (validator/full) nodes with healthchecking

### Workarounds if we don't push the PR
We could accomplish the same thing by filtering logs in logstash, but that would just stop the logs from being stored, but continue put pressure on the log throughput.

### Related PRs
* Original PR https://github.com/libra/libra/pull/6672

### DIFF
--- a/network/src/protocols/health_checker/mod.rs
+++ b/network/src/protocols/health_checker/mod.rs
@@ -233,7 +233,7 @@ where
                     match self.sample_random_peer() {
                         Some(peer_id) => {
                             let nonce = self.sample_nonce();
-                            debug!(
+                            trace!(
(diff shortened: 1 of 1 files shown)


libra-action:

Breaking changes policy:

2. Adding fields is ok.

dariorussi:
this looks pretty safe....

sausagee (approved):

PR #6694, opened 2020-11-13
[Consensus] Improve error handling for failure to initialize safety rules

JoshLind:
## Motivation

This PR improves the error handling/logging in consensus under failures to initialize safety rules (LSR). More specifically, it:
- Better distinguishes missing keys/data in secure storage vs. unexpected failures. 
- Improves the errors logged by consensus by adding additional context around how to react to the errors.

(This PR is small but mostly just creates a space for discussion around the best approach, so we can come to some state of consensus)

Yes.

## Test Plan

All tests pass locally.

## Related PRs

This PR follows a previous discussion here: https://github.com/libra/libra/pull/6671.

### DIFF
--- a/consensus/safety-rules/src/error.rs
+++ b/consensus/safety-rules/src/error.rs
@@ -29,8 +29,10 @@ pub enum Error {
     InvalidQuorumCertificate(String),
     #[error("{0} is not set, SafetyRules is not initialized")]
     NotInitialized(String),
-    #[error("Error returned by secure storage: {0}")]
-    SecureStorageError(String),
+    #[error("Data not found in secure storage: {0}")]
+    SecureStorageMissingDataError(String),
+    #[error("Unexpected error returned by secure storage: {0}")]
+    SecureStorageUnexpectedError(String),
     #[error("Serialization error: {0}")]
     SerializationError(String),
     #[error("Validator key not found: {0}")]
@@ -55,18 +57,23 @@ impl From<libra_secure_net::Error> for Error {
 
 impl From<libra_secure_storage::Error> for Error {
     fn from(error: libra_secure_storage::Error) -> Self {
-        // If a storage error is thrown that indicates a permission failure, we
-        // want to panic immediately to alert an operator that something has gone
-        // wrong. For example, this error is thrown when a storage (e.g., vault)
-        // token has expired, so it makes sense to fail fast and require a token
-        // renewal!
-        if libra_secure_storage::Error::PermissionDenied == error {
-            panic!(
-                "A permission error was thrown: {:?}. Maybe the storage token needs to be renewed?",
-                error
-            );
-        } else {
-            Self::SecureStorageError(error.to_string())
+        match error {
+            libra_secure_storage::Error::PermissionDenied => {
+                // If a storage error is thrown that indicates a permission failure, we
+                // want to panic immediately to alert an operator that something has gone
+                // wrong. For example, this error is thrown when a storage (e.g., vault)
+                // token has expired, so it makes sense to fail fast and require a token
+                // renewal!
+                panic!(
+                    "A permission error was thrown: {:?}. Maybe the storage token needs to be renewed?",
+                    error
+                );
+            }
+            libra_secure_storage::Error::KeyVersionNotFound(_, _)
+            | libra_secure_storage::Error::KeyNotSet(_) => {
+                Self::SecureStorageMissingDataError(error.to_string())
+            }
+            _ => Self::SecureStorageUnexpectedError(error.to_string()),
         }
     }
 }
--- a/consensus/src/epoch_manager.rs
+++ b/consensus/src/epoch_manager.rs
@@ -40,7 +40,7 @@ use libra_types::{
     on_chain_config::{OnChainConfigPayload, ValidatorSet},
 };
 use network::protocols::network::Event;
-use safety_rules::{Error, SafetyRulesManager};
+use safety_rules::SafetyRulesManager;
 use std::{cmp::Ordering, sync::Arc, time::Duration};
 
 /// RecoveryManager is used to process events in order to sync up with peer if we can't recover from local consensusdb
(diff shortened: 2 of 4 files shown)


consensus/safety-rules/src/error.rs
```diff
@@ -55,18 +57,21 @@ impl From<libra_secure_net::Error> for Error {
 
 impl From<libra_secure_storage::Error> for Error {
     fn from(error: libra_secure_storage::Error) -> Self {
-        // If a storage error is thrown that indicates a permission failure, we
-        // want to panic immediately to alert an operator that something has gone
-        // wrong. For example, this error is thrown when a storage (e.g., vault)
-        // token has expired, so it makes sense to fail fast and require a token
-        // renewal!
-        if libra_secure_storage::Error::PermissionDenied == error {
+        if matches!(error, libra_secure_storage::Error::PermissionDenied) {
```
davidiw, line 60: matches isn't really that great here, we should use match statement.

consensus/src/epoch_manager.rs
```diff
@@ -308,14 +308,17 @@ impl EpochManager {
             Ok(()) => { /* Expected */ }
             Err(Error::ValidatorKeyNotFound(error)) => {
                 error!(
+                    epoch = epoch,
                     error = error,
-                    "Unable to find the validator consensus key during initialize",
+                    "Unable to find the validator consensus key during initialize() and will not participate in consensus.\
+                     Will continue syncing until the validator consensus key matches in a future epoch."
```
davidiw, line 314: While this is an interesting message an all, I'm not convinced on its utility since we'll see a ton of errors and warnings afterward, so they'll have to look at the run books for consensus and safety rules anyway...

consensus/src/epoch_manager.rs
```diff
@@ -308,14 +308,17 @@ impl EpochManager {
             Ok(()) => { /* Expected */ }
             Err(Error::ValidatorKeyNotFound(error)) => {
                 error!(
+                    epoch = epoch,
                     error = error,
-                    "Unable to find the validator consensus key during initialize",
+                    "Unable to find the validator consensus key during initialize() and will not participate in consensus.\
```
davidiw, line 313: The error is simply "Unable to initialize safety rules" We can consider the second part on "not participate in consensus" see my next comment.

consensus/src/epoch_manager.rs
```diff
@@ -308,14 +308,17 @@ impl EpochManager {
...
+                     Will continue syncing until the validator consensus key matches in a future epoch."
                 );
             }
             Err(Error::ValidatorNotInSet(error)) => {
                 error!(
+                    epoch = epoch,
                     error = error,
-                    "The validator is not in the current validator set",
+                    "The validator is not in the current validator set for the epoch and will not participate in consensus.",
```
davidiw, line 321: I don't think we need these additional errors as it is already defined in the internal error message.

consensus/src/epoch_manager.rs
```diff
@@ -308,14 +308,17 @@ impl EpochManager {
...
                 );
             }
             Err(error) => panic!(
```
davidiw, line 324: I recommend we eliminate the panic / expect statement altogether ... if consensus can make progress, let it, who cares what happened to safety rules.... that's safety rules problem.

davidiw (commented):
Thanks for cleaning this up!

consensus/src/epoch_manager.rs
JoshLind: Interesting.. I don't know if I have enough context on how consensus operates here, but assuming it'll just continue to do its thing, this makes sense to me. I'll wait for the others to comment.

consensus/src/epoch_manager.rs
JoshLind: @aching, I'll let you and @davidiw battle it out 😃 My opinion is similar to David's in that even if this error message is super lengthy/explicit, a non-specialist (e.g., operator), would still need to follow the runbook to know how to react, so I'm not sure we gain a whole lot. Thoughts?

zekun000: code as runbook! :trollface:

davidiw:
If @aching does not comment after 5 days, we override!

consensus/src/epoch_manager.rs
aching: Run books are great but ideally an error can provide some context to be standalone in describing the issue and what to do, even if that is just to point to a runbook.  That's kind of what error codes are for anyway.  I think this would be helpful for someone just skimming errors to figure out how much to panic and why.  It's good for me.

aching:
This looks good to me, thanks @JoshLind for making the errors more informative.

davidiw (approved):
/land

consensus/src/epoch_manager.rs
```diff
@@ -306,22 +306,13 @@ impl EpochManager {
             MetricsSafetyRules::new(self.safety_rules_manager.client(), self.storage.clone());
         match safety_rules.perform_initialize() {
             Ok(()) => { /* Expected */ }
-            Err(Error::ValidatorKeyNotFound(error)) => {
+            Err(error) => {
```
davidiw, line 309: probably could have just done a if let Err(e) = safety_rules.perform_initialize() {

oh well

davidiw: in some ensuing PR we should probably add a counter that stores the LSR current state (enabled or disabled)

JoshLind: > probably could have just done a if let Err(e) = safety_rules.perform_initialize() {

Hah! This is what happen when I refactor too quickly :P

> in some ensuing PR we should probably add a counter that stores the LSR current state (enabled or disabled)

I agree; we can add this to the list of things we should do/fix when we tackle the observer mode. For now, the ugliness is just delegated to the runbook 😄

JoshLind:
@aching, FYI: to move quickly for the cherry-pick, David approved the PR with the short error message "Unable to initialize safety rules". We plan to improve the error messages and expand the information being logged as part of a real consensus observer mode.

PR #6698, opened 2020-11-13
[e2e-testsuite] tighten up tests for WriteSet transactions

bob-wilson:
This PR has a few different changes to improve the tests for WriteSet transactions:

- There were two tests for sending a WriteSet txn from an arbitrary account -- remove one of them.

- Add two new tests for the gas currency. (Even though WriteSet txns are not charged for gas, the gas currency still needs to be valid and correspond to a registered currency.)

- For tests that should fail in validation, check that they are also discarded from execution (except for the two new gas currency tests, which are currently broken in that regard).

- Change some comments to make it easier to see what each test is checking.

## Motivation

Better tests

Yes

## Test Plan

Ran them!


runtian-zhou (commented):
Do you mind update the golden files (remove the unneeded ones)? Maybe we should write a helper to do that.

bob-wilson:
It's just the one for "invalid_write_set_sender", right? I removed that now. It would be nice if we could arrange to put the golden files in the same place as the tests so it's more obvious when something should be removed, but that's a separate project.

runtian-zhou (approved):

runtian-zhou:
> It's just the one for "invalid_write_set_sender", right? I removed that now. It would be nice if we could arrange to put the golden files in the same place as the tests so it's more obvious when something should be removed, but that's a separate project.

That's correct. Or we need a way to batch update the golden files (the current regenerate command provided by the library doesn't deal with test removal).

PR #6706, opened 2020-11-14
[x] add support for testing packages changed since a given commit

sunshowers:
Add:
* an abstraction over the Git CLI
* a model for Cargo package selections
* a `changed-since` command to `x` to print out packages changed since
* a `--changed-since` option to `xtest` (and also `xbuild` and `xcheck`).
* support for file path linters
* a file path linter to ensure that filenames only come from a whitelist of characters
* a file path linter to ensure that every file in the repo is matched by a rule

Also move the exclusions in the whitespace linter to being part of `x.toml`.

See the commit messages for a detailed description of each commit.


libra-action:
```
target packages:
  M once_cell 1.5.2 (direct third-party, crates.io)
    * version upgraded from 1.4.1
    * added features: alloc
    * (unchanged features: default, std)

host packages:
  M once_cell 1.5.2 (transitive third-party, crates.io)
    * version upgraded from 1.4.1
    * added features: alloc
    * (unchanged features: default, std)
```

libra-action:
```
target packages:
  M once_cell 1.5.2 (direct third-party, crates.io)
    * version upgraded from 1.4.1
    * added features: alloc
    * (unchanged features: default, std)

host packages:
  M once_cell 1.5.2 (transitive third-party, crates.io)
    * version upgraded from 1.4.1
    * added features: alloc
    * (unchanged features: default, std)
```

libra-action:
```
target packages:
  M once_cell 1.5.2 (direct third-party, crates.io)
    * version upgraded from 1.4.1
    * added features: alloc
    * (unchanged features: default, std)

host packages:
  M once_cell 1.5.2 (transitive third-party, crates.io)
    * version upgraded from 1.4.1
    * added features: alloc
    * (unchanged features: default, std)
```

libra-action:
```
target packages:
  M once_cell 1.5.2 (direct third-party, crates.io)
    * version upgraded from 1.4.1
    * added features: alloc
    * (unchanged features: default, std)

host packages:
  M once_cell 1.5.2 (transitive third-party, crates.io)
    * version upgraded from 1.4.1
    * added features: alloc
    * (unchanged features: default, std)
```

libra-action:
```
target packages:
  M once_cell 1.5.2 (direct third-party, crates.io)
    * version upgraded from 1.4.1
    * added features: alloc
    * (unchanged features: default, std)

host packages:
  M once_cell 1.5.2 (transitive third-party, crates.io)
    * version upgraded from 1.4.1
    * added features: alloc
    * (unchanged features: default, std)
```

libra-action:
```
target packages:
  M once_cell 1.5.2 (direct third-party, crates.io)
    * version upgraded from 1.4.1
    * added features: alloc
    * (unchanged features: default, std)

host packages:
  M once_cell 1.5.2 (transitive third-party, crates.io)
    * version upgraded from 1.4.1
    * added features: alloc
    * (unchanged features: default, std)
```

devtools/x-core/src/errors.rs
```diff
@@ -34,6 +39,13 @@ impl SystemError {
         }
     }
 
+    pub fn from_hex(context: impl Into<Cow<'static, str>>, err: FromHexError) -> Self {
```
rexhoffman, line 42: Just wondering why this was needed?

devtools/x-core/src/git.rs
```diff
@@ -0,0 +1,207 @@
...
+    fn git_command(&self) -> Command {
+        // TODO: add support for the GIT environment variable?
+        let mut command = Command::new("git");
+        command.current_dir(self.root).stderr(Stdio::inherit());
+        command
+    }
+
+    fn get_or_init_scratch(&self, hash: &GitHash) -> Result<PathBuf> {
+        let mut scratch_dir = self.root.join("target");
+        scratch_dir.extend(&["x-scratch", "tree"]);
```
rexhoffman, line 140: can we get a random string location appended to x-scratch here?

devtools/x-core/src/errors.rs
sunshowers: Hmm, not sure I totally understand your question -- we end up parsing the Git hash that's presented, and it's possible to have errors during that process...

devtools/x-core/src/git.rs
sunshowers: I actually thought about that and decided not to do it that way -- a persistent scratch directory maintained by x is what I ended up going with. Means in the common case the same directory can be reused locally.

devtools/x-lint/src/runner.rs
```diff
@@ -206,67 +197,6 @@ impl<'cfg> LintEngine<'cfg> {
...
-        //    [`rental`](https://docs.rs/rental/).
-        // 3. Reorganize the code to use some sort of arena allocator which is passed into the lint
-        //    context (and/or stored alongside with `rental`.)
-        let stdout = self.files_stdout()?;
-
-        Ok(stdout
-            .split(|b| b == &0)
-            .filter_map(|s| {
-                // TODO: make global exclusions configurable.
-                if !s.is_empty() && !s.starts_with(b"testsuite/libra-fuzzer/artifacts/") {
```
rexhoffman, line 229: I think you still need this filter, otherwise any automation/attempt to add fuzz locations will fail on commit.

devtools/x-core/src/git.rs
```diff
@@ -0,0 +1,207 @@
...
+    // Caches.
+    tracked_files: OnceCell<Paths0>,
+}
+
+impl GitCli {
+    /// Creates a new instance of the Git CLI.
+    pub fn new(root: &'static Path) -> Self {
+        Self {
+            root,
+            tracked_files: OnceCell::new(),
```
rexhoffman, line 34: hmmm, if a dev grabs the source location (rather than a git pull, this wont detect the lack of a .git dir in the workspace and will fail later?   What would that error message look like?

devtools/x/src/cargo.rs
```diff
@@ -437,3 +392,153 @@ impl<'a> CargoCommand<'a> {
...
+            includes: CargoPackageInclude::Workspace,
+            excludes: BTreeSet::new(),
+        }
+    }
+
+    /// Returns a new `CargoPackages` that selects the default set of packages for the current
+    /// working directory. This may either be the entire workspace or a local package.
+    pub fn default_cwd(xctx: &XContext) -> Result<Self> {
+        let includes = if project_is_root(&xctx.config().cargo_config())? {
+            CargoPackageInclude::Workspace
```
rexhoffman, line 471: Hmm, little funky naming... took a while to make it clear how you were returning the full workspace when reading CargoPackages.    Move some of the documentation up to the type?

devtools/x/src/cargo.rs
```diff
@@ -437,3 +392,153 @@ impl<'a> CargoCommand<'a> {
...
+#[derive(Debug, StructOpt)]
+pub struct CargoPackageArgs {
+    #[structopt(long, short, number_of_values = 1)]
+    /// Run on the provided packages
+    pub(crate) package: Vec<String>,
+    #[structopt(long, short)]
+    /// Run on packages changed since the merge base of this commit
+    changed_since: Option<String>,
+    #[structopt(long)]
+    /// Run on all packages in the workspace
```
rexhoffman, line 406: It's taking me a bit longer to grok this since CargoPackages could represent the entire workspace, which seems a pretty large departure from where we were.   Not bad, just different.   I think this section needs more doc.

devtools/x/src/cargo.rs
```diff
@@ -437,3 +392,153 @@ impl<'a> CargoCommand<'a> {
...
+        self
+    }
+
+    // ---
+    // Helper methods
+    // ---
+
+    fn is_empty(&self) -> bool {
+        match &self.includes {
+            CargoPackageInclude::Workspace | CargoPackageInclude::LocalPackage => false,
```
rexhoffman, line 503: I can see for testing purposes why this is has nothing do with what's on disk, not sure if it's a good thing or not yet.

devtools/x/src/build.rs
```diff
@@ -255,14 +242,6 @@ pub fn run(args: Box<Args>, xctx: XContext) -> Result<()> {
...
-    } else if utils::project_is_root(&xctx.config().cargo_config())? {
-        // TODO Maybe only run a subest of builds if we're not inside
-        // a package but not at the project root (e.g. language)
-        cmd.run_on_all_packages(&CargoArgs::default())?;
-    } else {
-        cmd.run_on_local_package(&CargoArgs::default())?;
-    }
-    Ok(())
+    let packages = args.cargo_package.to_cargo_packages(&xctx)?;
+    cmd.run_on_packages(&packages, &CargoArgs::default())
```
rexhoffman, line 246: Does make the code look a lot cleaner.

devtools/x/src/cargo.rs
rexhoffman: naming.....   maybe something like "CommandTargets" instead of CargoPackages.....

devtools/x/src/config.rs
```diff
@@ -26,6 +27,8 @@ pub struct Config {
     /// Cargo configuration
     cargo: CargoConfig,
     grcov: CargoTool,
+    /// Determinator configuration
+    determinator: DeterminatorRules,
```
rexhoffman, line 31: Hope these have sane defaults that capture the standard cargo setup :D

x.toml
```diff
@@ -156,3 +156,26 @@ members = [
...
 ]
+
+# ---
+# Determinator rules
+# ---
+
+# CI-related files. TODO: maybe have separate rules for local and CI?
+[[determinator.path-rule]]
+globs = [".github/**/*", ".circleci/**/*", "codecov.yml"]
+mark-changed = "all"
```
rexhoffman, line 167: I personally think (based on what I implement with maven years ago doing something similar) that all files not explicitly excluded from the root of the project should be treated as changes to the root cargo project and result in a full rebuild.   Basically arguing the general assumption that any file under a Cargo.toml is part of the project unless there is another Cargo.toml between it and the root...   closest wins.   Everything after that is an exclusion.
This would capture changes to text files that are referenced in tests for example.    This rule would be redundant in that case.

x.toml
```diff
@@ -156,3 +156,26 @@ members = [
...
+
+# Core Rust and devtools files.
+[[determinator.path-rule]]
+globs = ["rust-toolchain", "cargo-toolchain", "cargo-flags", "x.toml"]
+mark-changed = "all"
```
rexhoffman, line 172: This rule would be redundant as well.

x.toml
```diff
@@ -156,3 +156,26 @@ members = [
...
+
+[[determinator.path-rule]]
+# Ignore website changes in x.
+globs = ["developers.libra.org/**/*"]
+mark-changed = []
```
rexhoffman, line 177: The complexity implied by this rule structure concerns me.    A file in module x implying that module y should be built that is not a dependent crate means there is a cross talk between modules that should be actively discouraged.    The benefits of discouraging that kind of cross talk is we approach a more buck/bazel like build where the inputs to a crate's build are captured by the structure of the cargo toml files.   I think flexibility here is bad.

x.toml
```diff
@@ -156,3 +156,26 @@ members = [
...
+
+[[determinator.package-rule]]
+on-affected = ["x"]
+mark-changed = "all"
```
rexhoffman, line 181: Is x considered a package of libra?    I'd think the virtual workspace would ignore it for the purposes of say "cargo build" or "cargo test".   I think we'd also want to remove x as da default member? https://github.com/libra/libra/blob/master/Cargo.toml#L45-L47.

rexhoffman (commented):
I'd like a bit of documentation (maybe a name change around CargoPackages), and then happy to approve.    

Bigger and not as critical.  I would like to follow up on the rules the determinator uses to determine changes on virtual modules to get us to encourage more buck/bazel like behavior and thinking about the files as well as the inputs/outputs of a crates/virtual crates.

devtools/x-lint/src/runner.rs
sunshowers: Oh yeah, thanks for catching this. Will fix.

devtools/x-core/src/errors.rs
rexhoffman: Got it, new error type that exist if the hash results in an error in interacting with git.   Is it more than just the hash?  Perhaps the hash is well formed but doesn't exist in the git tree.   In that case is it a git error and not a hash error?

devtools/x-core/src/git.rs
rexhoffman: Comment the function to that effect?

x.toml
rexhoffman: Not exactly true in the case of crates that aren't listed as packages from the virtual cargo toml.   I think we should remove devtools as default targets so the "fall back in to" being part of the root virtual crate.

devtools/x/src/config.rs
sunshowers: You know, I've been meaning to add defaults for e.g. `rust-toolchain` etc. Should do them upstream.

x.toml
sunshowers: Hmm, I've gone back and forth on whether the default for a file that can't be matched to a closer Cargo.toml should be skip or full-rebuild. The arguments for skip are:
* We don't try and represent the virtual package at the root at all -- we only care about the actual packages with real code.
* If the root `Cargo.toml` or `Cargo.lock` changes we *do not* do a full rebuild -- and this isn't some sort of special optimization, it falls naturally out of the design.
* In my experience, overbuilding is harder to notice than underbuilding unless we add some logging and metrics which we don't have at the moment -- missing a bug because of underbuilding will be caught quickly.

The arguments for full-rebuild are:
* It's less likely to lead to bugs than skip.
* Maybe it matches people's intuitions better? Not sure about this one.

Some other options are:
* Add a flag to the determinator config controlling the default.
* Print clear warnings for files that were ignored to logs.

Either way, it makes sense that some files should probably be added to the default set of rules in the determinator upstream, like `rust-toolchain` etc. I'll do that upstream.

>  Basically arguing the general assumption that any file under a Cargo.toml is part of the project unless there is another Cargo.toml between it and the root... closest wins

To be clear, this is already how the determinator works, except we don't have a notion of "part of the project" in the determinator at all.

x.toml
sunshowers: I agree that they should be discouraged, but there are legitimate use cases for this. For example, `serde` has `serde_derive` and `serde_derive_internals` package, and the `internals` package's `src` is a symlink to `serde_derive`. This is basically the only way to make procedural macros testable -- I discovered this while writing proc macros myself.

I can add a comment to the determinator saying that this is discouraged, but I don't think I can get rid of it entirely.

x.toml
sunshowers: > Is x considered a package of libra?

Yes, it's in the same workspace, just like all of our other non-test code.

> I'd think the virtual workspace would ignore it for the purposes of say "cargo build" or "cargo test".

Well, we do want tests in `x` to run in CI...

> I think we'd also want to remove x as default member?

Don't have a strong opinion either way on this at the moment -- it's kind of a production binary in that it is a core part of the production pipeline.

devtools/x/src/cargo.rs
sunshowers: Hmm, could you elaborate? I'm not sure what you mean.

devtools/x/src/build.rs
```diff
@@ -23,15 +23,8 @@ pub struct Args {
     #[structopt(long, short)]
     /// No output printed to stdout
     quiet: bool,
-    #[structopt(long, short, number_of_values = 1)]
-    /// Package to build (see `cargo help pkgid`)
-    package: Vec<String>,
-    #[structopt(long)]
-    /// Alias for --workspace (deprecated)
```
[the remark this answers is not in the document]
sunshowers: Yeah, I don't think it's that big a deal, `--all` has been deprecated for a long time and at this point backcompat is not a huge concern.

devtools/x-core/src/git.rs
sunshowers: Chatting with Rex F2F, what he meant was the source archive -- definitely agreed that it needs to be handled.

devtools/x/src/cargo.rs
sunshowers: TODO: document design requirements and decisions here.

devtools/x/src/cargo.rs
sunshowers: TODO: document design requirements and decisions here.

devtools/x/src/cargo.rs
sunshowers: TODO: change name of method, add logging for the case where we don't try and build anything.

devtools/x-core/src/git.rs
sunshowers: Done.

devtools/x-core/src/git.rs
sunshowers: Added handling for missing .git.

devtools/x/src/cargo.rs
sunshowers: Done.

devtools/x/src/cargo.rs
sunshowers: Done.

devtools/x/src/config.rs
sunshowers: Done, added defaults upstream.

x.toml
sunshowers: After discussion with @rexhoffman, switched the default over to full-rebuild rather than ignore. Also added a linter to ensure that every file in the libra workspace is matched by a rule and/or a package.

x.toml
sunshowers: Even though it would be, per our discussion I've added a linter to ensure that every file in the repo is covered by a rule or package.

devtools/x/src/cargo.rs
sunshowers: Renamed it to `SelectedPackages`.

devtools/x-lint/src/runner.rs
sunshowers: Done.

libra-action:
```
target packages:
  M once_cell 1.5.2 (direct third-party, crates.io)
    * version upgraded from 1.4.1
    * added features: alloc
    * (unchanged features: default, std)

host packages:
  M once_cell 1.5.2 (transitive third-party, crates.io)
    * version upgraded from 1.4.1
    * added features: alloc
    * (unchanged features: default, std)
```

libra-action:
```
target packages:
  M once_cell 1.5.2 (direct third-party, crates.io)
    * version upgraded from 1.4.1
    * added features: alloc
    * (unchanged features: default, std)

host packages:
  M once_cell 1.5.2 (transitive third-party, crates.io)
    * version upgraded from 1.4.1
    * added features: alloc
    * (unchanged features: default, std)
```

libra-action:
```
target packages:
  M once_cell 1.5.2 (direct third-party, crates.io)
    * version upgraded from 1.4.1
    * added features: alloc
    * (unchanged features: default, std)

host packages:
  M once_cell 1.5.2 (transitive third-party, crates.io)
    * version upgraded from 1.4.1
    * added features: alloc
    * (unchanged features: default, std)
```

x.toml
```diff
@@ -156,3 +169,38 @@ members = [
...
+# ---
+
+# CI-related files. TODO: maybe have separate rules for local and CI?
+[[determinator.path-rule]]
+globs = [".github/**/*", ".circleci/**/*", "codecov.yml"]
+mark-changed = "all"
+
+# Core devtools files.
+[[determinator.path-rule]]
+globs = ["cargo-toolchain", "cargo-flags", "scripts/**/*", "x.toml"]
```
rexhoffman, line 184: not sure if I agree with scripts/ being in the all rule.   I would think scripts should be orthogonal except for dev_setup.sh in CI.   Yeah just checked them all think only dev_setup.sh in CI should spark a full rebuild.

x.toml
```diff
@@ -69,6 +69,19 @@ include-dev = true
 proc-macros-on-target = false
 # Don't set target or host platforms, or omitted packages, for the full set.
 
+[workspace]
+
+# Regex for allowed characters in paths. Regex must have ^ and $ anchors.
+allowed-paths = "^([a-zA-Z0-9._\\-/@:]|-)+$"
```
rexhoffman, line 75: I think we should be restrictive as we possible can here.   Not sure if you already have with the fuzzers.

Also I think it's fine to make this a default and remove from the config file and perhaps (or not?) allow for overrides.

devtools/x-lint/src/runner.rs
```diff
@@ -211,61 +238,12 @@ impl<'cfg> LintEngine<'cfg> {
...
-                }
-                Ok(output.stdout)
-            })
-            .map(|stdout| stdout.as_slice())
+    fn file_list(&self) -> Result<impl Iterator<Item = &'cfg Path> + 'cfg> {
+        let tracked_files = self.config.core.git_cli().tracked_files()?;
+        // TODO: make global exclusions configurable
+        Ok(tracked_files
+            .iter()
+            .filter(|f| !f.starts_with("testsuite/libra-fuzzer/artifacts/")))
```
rexhoffman, line 246: Ah this is still needed do to header/final line rules and such.   Good TODO on it.

rexhoffman (approved):
Looks really good, you went all out on it.   Just some lints and a test to cleanup.

x.toml
sunshowers: Ah cool, I can do that then.

sunshowers: Done.

x.toml
sunshowers: Hmm, I'd like to tackle the fuzzer thing at a later point -- I think they should be excluded from content linters but maybe not from any others.

libra-action:
```
target packages:
  M once_cell 1.5.2 (direct third-party, crates.io)
    * version upgraded from 1.4.1
    * added features: alloc
    * (unchanged features: default, std)

host packages:
  M once_cell 1.5.2 (transitive third-party, crates.io)
    * version upgraded from 1.4.1
    * added features: alloc
    * (unchanged features: default, std)
```

libra-action:
```
target packages:
  M once_cell 1.5.2 (direct third-party, crates.io)
    * version upgraded from 1.4.1
    * added features: alloc
    * (unchanged features: default, std)

host packages:
  M once_cell 1.5.2 (transitive third-party, crates.io)
    * version upgraded from 1.4.1
    * added features: alloc
    * (unchanged features: default, std)
```

libra-action:
```
target packages:
  M once_cell 1.5.2 (direct third-party, crates.io)
    * version upgraded from 1.4.1
    * added features: alloc
    * (unchanged features: default, std)

host packages:
  M once_cell 1.5.2 (transitive third-party, crates.io)
    * version upgraded from 1.4.1
    * added features: alloc
    * (unchanged features: default, std)
```

libra-action:
```
target packages:
  M once_cell 1.5.2 (direct third-party, crates.io)
    * version upgraded from 1.4.1
    * added features: alloc
    * (unchanged features: default, std)

host packages:
  M once_cell 1.5.2 (transitive third-party, crates.io)
    * version upgraded from 1.4.1
    * added features: alloc
    * (unchanged features: default, std)
```

libra-action:
```
target packages:
  M once_cell 1.5.2 (direct third-party, crates.io)
    * version upgraded from 1.4.1
    * added features: alloc
    * (unchanged features: default, std)

host packages:
  M once_cell 1.5.2 (transitive third-party, crates.io)
    * version upgraded from 1.4.1
    * added features: alloc
    * (unchanged features: default, std)
```

libra-action:
```
target packages:
  M once_cell 1.5.2 (direct third-party, crates.io)
    * version upgraded from 1.4.1
    * added features: alloc
    * (unchanged features: default, std)

host packages:
  M once_cell 1.5.2 (transitive third-party, crates.io)
    * version upgraded from 1.4.1
    * added features: alloc
    * (unchanged features: default, std)
```

devtools/x/src/context.rs
```diff
@@ -14,16 +15,18 @@ pub struct XContext {
...
     pub fn new() -> Result<Self> {
-        Ok(Self::with_config(Config::from_project_root()?))
+        Self::with_config(Config::from_project_root()?)
     }
 
     /// Creates a new `GlobalContext` based on the given config.
-    pub fn with_config(config: Config) -> Self {
-        Self {
-            core: XCoreContext::new(project_root()),
+    pub fn with_config(config: Config) -> Result<Self> {
```
rexhoffman, line 22: I really like this approach to providing data to a context.   I used similar approaches with spring context for years.

devtools/x/src/utils.rs
```diff
@@ -32,6 +32,7 @@ pub fn locate_project(cargo_config: &CargoConfig) -> Result<PathBuf> {
     Ok(serde_json::from_slice::<LocateProject>(&output)?.root)
 }
 
+#[allow(dead_code)]
```
rexhoffman: any reason to not remove?

rexhoffman (approved):

devtools/x/src/context.rs
sunshowers: Yeah, I was thinking of future testability when I did it that way. Definitely makes it a lot easier.

devtools/x/src/utils.rs
sunshowers: Yeah I was thinking it might be useful in the future, but probably not given this approach. Will remove.

sunshowers: Removed.

libra-action:
```
target packages:
  M once_cell 1.5.2 (direct third-party, crates.io)
    * version upgraded from 1.4.1
    * added features: alloc
    * (unchanged features: default, std)

host packages:
  M once_cell 1.5.2 (transitive third-party, crates.io)
    * version upgraded from 1.4.1
    * added features: alloc
    * (unchanged features: default, std)
```

libra-action:
```
target packages:
  M once_cell 1.5.2 (direct third-party, crates.io)
    * version upgraded from 1.4.1
    * added features: alloc
    * (unchanged features: default, std)

host packages:
  M once_cell 1.5.2 (transitive third-party, crates.io)
    * version upgraded from 1.4.1
    * added features: alloc
    * (unchanged features: default, std)
```

libra-action:
```
target packages:
  M once_cell 1.5.2 (direct third-party, crates.io)
    * version upgraded from 1.4.1
    * added features: alloc
    * (unchanged features: default, std)

host packages:
  M once_cell 1.5.2 (transitive third-party, crates.io)
    * version upgraded from 1.4.1
    * added features: alloc
    * (unchanged features: default, std)
```

sunshowers:
/land

(retrying per @rexhoffman's recommendation)

sunshowers:
/land

last attempt for now

PR #6266, opened 2020-09-29
[Draft] Use a threading model for the key manager smoke test.

JoshLind:
This is a draft PR. Please don't review.

## Motivation

This PR updates the key manager smoke test to use a thread instead of spawning a new process.

### Have you read the [Contributing Guidelines on pull requests](https://github.com/libra/libra/blob/master/CONTRIBUTING.md#pull-requests)?

(Write your answer here.)

## Test Plan

(Share your test plan here. If you changed code, please provide us with clear instructions for verifying that your changes work.)

## Related PRs

(If this PR adds or changes functionality, please take some time to update the docs at https://github.com/libra/libra/developers.libra.org, and link to your PR here.)

PR #6707, opened 2020-11-14
[Smoke Tests] Add smoke test attribute for spawned processes and update key manager test.

JoshLind:
## Motivation

This PR offers two incremental improvements for our smoke test infrastructure:
- First, it adds a new "#[smoke_test]" attribute that can be added to individual smoke tests to spawn them in new processes and automatically fail them after a specific amount of time (e.g., 2 minutes). It does this using the rusty fork crate (https://github.com/AltSysrq/rusty-fork) and a new "smoke-test/attribute" crate.
- Second, it updates the key manager smoke test to adopt a spawned thread approach, as opposed to using workspace builder to build and run the key manager as a separate process binary. The benefit of this is that the key manager test now completes in seconds, as opposed to minutes. Moreover, it also makes the key manager test much less flaky (I've yet to see it flake, so we no longer need to ignore it! 😄).

Why do we need to use the rusty fork library?
- Currently, this is because Rust doesn't have great support for killing threads. If we didn't spawn the test as a new process it would mean that when the test completed, we'd either have to kill the thread manually, which is hacky (e.g., see the draft PR linked below), or just let it run in the background, which isn't great. Instead, now that each test is run using its own process, when the test completes, the process dies and the spawned threads go with it!

Do we plan on using this for other smoke tests?
- Yes, that's the plan. Obviously, it'll require a lot more work to allow us to spawn Libra nodes using threads instead of processes, but once we do that, we'll be able to use the "#[smoke_test]" attribute to help us clean up deployments after each test. This will mean potentially removing the workspace builder and having faster smoke tests. We do this for the key manager test now first, as it's a simple proof of concept that is easy to reason about.

Yes.

## Test Plan

The tests pass locally.

## Related PRs

This PR builds on a previous draft that tried to adopt a spawned thread approach for the key manager test: https://github.com/libra/libra/pull/6266/files. The old PR was closed because we didn't have a great way to kill the key manager thread after the test completed. Now we do!

### DIFF
--- a/testsuite/smoke-test/src/key_manager.rs
+++ b/testsuite/smoke-test/src/key_manager.rs
@@ -3,68 +3,73 @@
 
 use crate::{
     smoke_test_environment::SmokeTestEnvironment,
-    test_utils::libra_swarm_utils::{get_json_rpc_libra_interface, load_node_config},
-    workspace_builder,
+    test_utils::libra_swarm_utils::load_backend_storage,
 };
-use libra_config::config::{Identity, KeyManagerConfig};
+use libra_config::config::NodeConfig;
 use libra_global_constants::CONSENSUS_KEY;
-use libra_key_manager::libra_interface::LibraInterface;
+use libra_key_manager::{
+    libra_interface::{JsonRpcLibraInterface, LibraInterface},
+    KeyManager,
+};
 use libra_secure_storage::{CryptoStorage, Storage};
-use std::{convert::TryInto, process::Command, thread::sleep, time::Duration};
+use libra_secure_time::RealTimeService;
+use libra_smoke_test_attribute::smoke_test;
+use libra_types::chain_id::ChainId;
+use rusty_fork::rusty_fork_test;
+use std::{convert::TryInto, thread, thread::sleep, time::Duration};
 
-const KEY_MANAGER_BIN: &str = "libra-key-manager";
-
-#[test]
-#[ignore]
+#[smoke_test]
 fn test_key_manager_consensus_rotation() {
+    // Create and launch a local validator swarm
     let mut env = SmokeTestEnvironment::new(1);
     env.validator_swarm.launch();
 
-    // Create a node config for the key manager by extracting the first node config in the swarm.
-    let (node_config, config_path) = load_node_config(&env.validator_swarm, 0);
-    let mut key_manager_config = KeyManagerConfig::default();
-    key_manager_config.json_rpc_endpoint =
-        format!("http://127.0.0.1:{}", node_config.json_rpc.address.port());
-    key_manager_config.rotation_period_secs = 10;
-    key_manager_config.sleep_period_secs = 1000; // Large sleep period to force a single rotation
+    // Fetch the first node config in the swarm
+    let node_config_path = env.validator_swarm.config.config_files.get(0).unwrap();
+    let node_config = NodeConfig::load(&node_config_path).unwrap();
 
-    // Load validator's on disk storage and update key manager secure backend in config
-    let storage: Storage = if let Identity::FromStorage(storage_identity) =
-        &node_config.validator_network.as_ref().unwrap().identity
-    {
-        let storage_backend = storage_identity.backend.clone();
-        key_manager_config.secure_backend = storage_backend.clone();
-        (&storage_backend).try_into().unwrap()
-    } else {
-        panic!("Couldn't load identity from storage");
-    };
-
-    // Save the key manager config to disk
-    let key_manager_config_path = config_path.with_file_name("key_manager.yaml");
-    key_manager_config.save(&key_manager_config_path).unwrap();
+    // Load validator's on disk storage
+    let secure_backend = load_backend_storage(&env.validator_swarm, 0);
+    let storage: Storage = (&secure_backend).try_into().unwrap();
 
     // Create a json-rpc connection to the blockchain and verify storage matches the on-chain state.
-    let libra_interface = get_json_rpc_libra_interface(&env.validator_swarm, 0);
+    let json_rpc_endpoint = format!(
+        "http://127.0.0.1:{}",
+        env.validator_swarm.get_client_port(0)
+    );
+    let libra_interface = JsonRpcLibraInterface::new(json_rpc_endpoint.clone());
     let account = node_config.validator_network.unwrap().peer_id();
     let current_consensus = storage.get_public_key(CONSENSUS_KEY).unwrap().public_key;
     let validator_info = libra_interface.retrieve_validator_info(account).unwrap();
     assert_eq!(&current_consensus, validator_info.consensus_public_key());
 
-    // Spawn the key manager and sleep until a rotation occurs.
-    let mut command = Command::new(workspace_builder::get_bin(KEY_MANAGER_BIN));
-    command
-        .current_dir(workspace_builder::workspace_root())
-        .arg(key_manager_config_path);
+    // Create the key manager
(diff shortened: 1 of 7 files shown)


rexhoffman:
So great improvement to test result, but with serious downsides.

1) spawning processes means debugging the test and code under test isn't as straight forward as it might otherwise be, or depending on tooling might not be and option.
2) Sound like the threading model needs handled appropriately then.    If we can't find ways to spawn/kill threads safely then passing work to an existing thread pool like tokio that we can interrupt and kill may be needed....   This crate claims to have figured it out:  https://docs.rs/thread-pool/0.1.1/thread_pool/struct.ThreadPool.html.  Don't have a ton of time to dig in to it, but seems that this underlying problem should be treated as a high priority as it's a fundamental building block of testing and robust systems.

metajack:
Can we also trigger the tests in threaded mode for debugging purposes? Alternatively, is there a way to launch the test process directly for debugging?

bmwill:
@JoshLind I think this is a great step forward and I'm glad that it looks like you've been able to eliminate some of the flakiness from this particular test. My only concern is that I think we probably need to completely rethink the testing architecture (maybe not even using standard rust `#[test]` attributes) in order to make even more improvements to smoketests as a whole (as well as making things easier to share with cluster test)

testsuite/smoke-test/src/key_manager.rs
```diff
@@ -3,68 +3,66 @@
...
+    sleep(Duration::from_secs(15)); // Wait a few seconds for the blockchain to update
 
     // Verify the consensus key has been rotated in secure storage and on-chain.
     let rotated_consensus = storage.get_public_key(CONSENSUS_KEY).unwrap().public_key;
     let validator_info = libra_interface.retrieve_validator_info(account).unwrap();
     assert_eq!(&rotated_consensus, validator_info.consensus_public_key());
     assert_ne!(current_consensus, rotated_consensus);
-
-    // Kill the key manager process
-    key_manager.kill().unwrap();
```
davidiw, line 69: lovely

testsuite/smoke-test/src/key_manager.rs
```diff
@@ -3,68 +3,66 @@
...
+        1000,
+        ChainId::test(),
+    );
 
-    let mut key_manager = command.spawn().unwrap();
-    sleep(Duration::from_secs(20));
+    // Spawn the key manager and execute a rotation
+    sleep(Duration::from_secs(10)); // Wait a few seconds for the key to become stale
+    let _key_manager_thread = thread::spawn(move || key_manager.execute());
+    sleep(Duration::from_secs(15)); // Wait a few seconds for the blockchain to update
```
davidiw, line 61: wow that's a long time...

testsuite/smoke-test/src/key_manager.rs
```diff
@@ -3,68 +3,66 @@
...
+        1,
+        1000, // Large sleep period to force a single rotation
+        1000,
+        ChainId::test(),
+    );
 
-    let mut key_manager = command.spawn().unwrap();
-    sleep(Duration::from_secs(20));
+    // Spawn the key manager and execute a rotation
+    sleep(Duration::from_secs(10)); // Wait a few seconds for the key to become stale
```
davidiw, line 59: this feels awkward, can we change the timeouts in key manager to be aggressive and then here we just poll every second for up to 20 seconds that the rotation occurred instead of this wait 10 and then wait 15 and then check?

testsuite/smoke-test/src/key_manager.rs
```diff
@@ -3,68 +3,66 @@
...
+use rusty_fork::rusty_fork_test;
+use std::{convert::TryInto, thread, thread::sleep, time::Duration};
 
-#[test]
-#[ignore]
+#[smoke_test]
 fn test_key_manager_consensus_rotation() {
+    // Create and launch a local validator swarm
     let mut env = SmokeTestEnvironment::new(1);
     env.validator_swarm.launch();
```
davidiw, line 25: so these are all still running using the legacy method, right?

davidiw (commented):
Overall looks good, will review others comments outside the code itself.

davidiw:
@rexhoffman, I think you make a good point, when we spawn the test fork, does the output and maybe even error propagate to the originating thread? At least well enough so that we could potentially see how far into the "smoke test" the framework got before crashing?

> Can we also trigger the tests in threaded mode for debugging purposes? Alternatively, is there a way to launch the test process directly for debugging?

@JoshLind I think the answer is eventually? Maybe it would be good to create a github issue that lies out your plan?

> @JoshLind I think this is a great step forward and I'm glad that it looks like you've been able to eliminate some of the flakiness from this particular test. My only concern is that I think we probably need to completely rethink the testing architecture (maybe not even using standard rust #[test] attributes) in order to make even more improvements to smoketests as a whole (as well as making things easier to share with cluster test)

@bmwill, agreed, let's make sure that this puts us on the right path before we land it although it does improve the UX quite a bit... I can imagine the B&R smoke test will be an even better cleanup.

testsuite/smoke-test/src/key_manager.rs
JoshLind: Sure, I've updated it to check in a loop for up to 60 seconds (6 seconds sleep period, 10 periods) 😄

testsuite/smoke-test/src/key_manager.rs
JoshLind: Yes, nothing has changed from the perspective of the swarm. These changes will come in a follow up PR.

davidiw (approved):
unblock the good times

JoshLind:
Thanks all! 😄 (returning to this..)

> @bmwill, agreed, let's make sure that this puts us on the right path before we land it although it does improve the UX quite a bit... I can imagine the B&R smoke test will be an even better cleanup.

@davidiw, FYI: @bmwill and I had a sync before the break and it seems that this work should line up pretty well with the bigger testing changes we have in mind -- so nothing should be blocking us for now with these smaller PRs.

> @rexhoffman, I think you make a good point, when we spawn the test fork, does the output and maybe even error propagate to the originating thread? At least well enough so that we could potentially see how far into the "smoke test" the framework got before crashing?

@davidiw + @rexhoffman, regarding test debugging, nothing has changed in this PR. The key manager thread is still logging using the same logger used by the libra node. If a failure occurs, the log shows both the node and the key manager logs, so it's the same to debug as the current model. (Note: the current model has to explicit attach to the parent process, but we no longer have to do this because loggers are a per-process entity).

> Can we also trigger the tests in threaded mode for debugging purposes? Alternatively, is there a way to launch the test process directly for debugging?
> @JoshLind I think the answer is eventually? Maybe it would be good to create a github issue that lies out your plan?

@davidiw, @metajack, This is something that will be more obvious in the next PR (where we actually adopt the threading approach for the libra nodes), however, from a test debugging point of view, nothing changes --the nodes will still log to the log files and the test process will still output any errors to the console, so it's all the same as our current model.
- Regarding supporting both a threading and process model for libra swarm, this is TBD. My feeling is that supporting both will allow us to easily roll back if we encounter any issues. However, FWIW, I've worked through a proof of concept PR that updates our tests to adopt the threading approach and nothing should change from the developers' perspective/experience.
- Regarding an issue to outline the current plans I have, I've created one here: https://github.com/libra/libra/issues/6758. I'll update it as I go along.

JoshLind:
(Will land this tonight -- just giving some time to make sure there isn't any pushback. But given that nothing big is really changing in this PR, I suspect not)

PR #6709, opened 2020-11-16
[LCS] remove test dependency to libra-crypto

matbd:
## Motivation

We are going to move LCS to its own repo soon.

## Test Plan

CI


bmwill (approved):
LGTM to me, though might want to wait for some of the crypto folks to see this before merging

PR #6710, opened 2020-11-16
Cherry-pick PR #6656 [config] rocksdb options to release 1.0

lightmark:
- Justification and breaking nature (who does it affect? validators, full nodes, tooling, operators, AOS, etc.)
All the node will be affected with the new default rocksdb options.

- Comprehensive test results that demonstrate the fix working and not breaking existing workflows.
Theoretically it doesn't affect libra system behavior but db behaviors. The current cluster tests and dry-run longevity test has proven it is working and not breaking anything.

- Why we must have it for V1 launch.
Because for long run it may trigger OOM.

- What workarounds and alternative we have if we do not push the PR.
provide unlimited memory to each node which seems impractical

### DIFF
--- a/language/libra-tools/libra-validator-interface/Cargo.toml
+++ b/language/libra-tools/libra-validator-interface/Cargo.toml
@@ -14,6 +14,7 @@ edition = "2018"
 [dependencies]
 reqwest = { version = "0.10.8", features = ["blocking", "json"] }
 anyhow = "1.0.33"
+libra-config = { path = "../../../config", version = "0.1.0" }
 libra-json-rpc-client = { path = "../../../client/json-rpc", version = "0.1.0" }
 libra-types = { path = "../../../types", version = "0.1.0" }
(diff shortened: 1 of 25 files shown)


dariorussi:
so is it protecting the OOM the only effect of this PR?
Do we have a sense of when the system OOMs and under which conditions?

sausagee (approved):

language/libra-tools/libra-validator-interface/Cargo.toml
```diff
@@ -13,7 +13,8 @@ edition = "2018"
 
 [dependencies]
 reqwest = { version = "0.10.8", features = ["blocking", "json"] }
-anyhow = "1.0.33"
+anyhow = "1.0.34"
```
sausagee, line 16: @lightmark please revert this version bump.

sausagee (changes requested):

lightmark:
@sausagee plz stamp

sausagee (approved):
/land

PR #6711, opened 2020-11-17
[writeset-generator] Implement a unifited api for writeset generation

runtian-zhou:
… process.

## Motivation

This is the mirror version of #6557 trying to come up with the api for generating writesets for future upgrades. The difference here is that this api disallows shadowing the resource: only modules that will be published in the network are allowed to shadow the execution process. We suggest that this is a much safer approach than the original shadowing versioning due to following reasons:

1. All code that will be used to generate writeset will be committed on chain, which allows a good visibility of how writesets are created.
2. There were still some safety concern of if we could handle the access control policy for writeset only functions. We suggest this will become less of an issue once we have better visibility modifier and constraint system in Move.

In the long run however, we suggest that we should minimize the use of this tool and use admin script to mimic what's done in this module. This is because admin script will provide the full visibility of how writeset is generated, and will not suffer to this problem where the writeset generation version is always lagged to the current head state. As a result, we will need to investigate into how we can extend the functionality of admin script to satisfy our need for future upgrades(a good criterion is whether we could use admin script to spawn genesis), and use admin scripts for all known use cases. And we will use the `DirectWriteSet` only on those scenarios where some ad-hoc fix is needed and cannot be achieved via admin script.

Yes

## Test Plan

Added a new test.


runtian-zhou:
For @vgao1996's comment over why not expose `publish_module` in `GenesisSession`, I think the answer wil be much clearer in this PR: the reason is that we don't want module shadowing to happen. i.e: all modules that generates on-chain states will be published. Thus instead of publishing them one by one in the session, publishing them all at once seems the way to go as they will need to be published anyway.

language/stdlib/modules/LibraConfig.move
```diff
@@ -164,6 +167,27 @@ module LibraConfig {
         include SetEnsures<Config>;
     }
 
+    fun disable_reconfiguration(lr_account: &signer) {
```
sblackshear, line 170: Comments on these functions to document that they are private functions that should only be used by the WriteSet generator?

language/stdlib/modules/LibraConfig.move
```diff
@@ -164,6 +167,27 @@ module LibraConfig {
...
+        assert(
+            Signer::address_of(lr_account) == CoreAddresses::LIBRA_ROOT_ADDRESS(),
+            Errors::requires_address(ELIBRA_CONFIG)
+        );
+        Roles::assert_libra_root(lr_account);
+        move_to(lr_account, DisableReconfiguration {} )
```
sblackshear, line 176: Would it make more sense for this to be a field of the existing `Configuration` resource? That seems easier than trying to create a new singleton resource under LIBRA_ROOT.

language/stdlib/modules/LibraConfig.move
```diff
@@ -164,6 +167,27 @@ module LibraConfig {
...
+
+    fun enable_reconfiguration(lr_account: &signer) acquires DisableReconfiguration {
+        assert(Signer::address_of(lr_account) == CoreAddresses::LIBRA_ROOT_ADDRESS(), Errors::requires_address(ELIBRA_CONFIG));
+
+        if(exists<DisableReconfiguration>(Signer::address_of(lr_account))) {
+            DisableReconfiguration {} = move_from<DisableReconfiguration>(Signer::address_of(lr_account));
+        }
+    }
+
+    fun should_reconfigure(): bool {
```
sblackshear, line 187: This name is a bit confusing--maybe `reconfiguration_enabled`?

language/libra-tools/writeset-transaction-generator/src/writeset_builder.rs
```diff
@@ -0,0 +1,180 @@
...
+            .0
+            .finish()
+            .map_err(|err| format_err!("Unexpected VM Error: {:?}", err))
+            .unwrap()
+    };
+
+    for module in modules {
+        let mut module_bytes = vec![];
+        module.serialize(&mut module_bytes).unwrap();
+        // TODO: Check compatibility between old and new modules.
```
sblackshear, line 167: This should be easier to do once https://github.com/libra/libra/pull/6753 lands (see new `compatibility.rs` file). But definitely seems like we should add this sanity check before landing.

language/libra-tools/writeset-transaction-generator/src/writeset_builder.rs
```diff
@@ -0,0 +1,180 @@
...
+        let mut module_bytes = vec![];
+        module.serialize(&mut module_bytes).unwrap();
+        shadow_changeset.publish_or_overwrite_module(id, module_bytes);
+    }
+    shadow_changeset
+}
+
+pub fn build_changeset<S: StateView, F>(
+    state_view: &S,
+    procedure: F,
```
sblackshear, line 142: Would it be possible to make this a `CompiledScript,  the name of a source file that will be compiled against` modules`, or a fully qualified procedure name? A closure feels like a bit of a blank check given how structured the uses of this are (e.g., typically calling one Move function).

My main concern here is transparency--it would be nice to publish a Move-only artifact on GitHub that shows how the WriteSet was produced. But this closure will disappear forever after `build_changeset` is run.

runtian-zhou: There are two main issues here:
1. `CompiledScript` cannot invoke private functions in a module
2. `CompiledScript` cannot call as arbitrary signer.

Those are the two major drawbacks that already occurs in the genesis generation process that prevents us from using pure Move script to setup the states.

language/stdlib/modules/LibraConfig.move
runtian-zhou: My original thought is that we would like to avoid breaking changes like this. The current PR will leave the on-chain state totally unchanged. I'm a bit more confortable with this approach as we don't have additional unused data on chain.

runtian-zhou: Actually after playing with the code, it is not possible for us to put it in `Configuration`. The reason is that changes to `Configuration` during the WriteSet generation process will cause Configuration resource to be part of that writeset, even if we change the value back.

language/stdlib/modules/LibraConfig.move
runtian-zhou: Updated in fixup!

language/stdlib/modules/LibraConfig.move
runtian-zhou: Updated in fixup!

tzakian (commented):
Overall LGTM! Just the one comment

language/stdlib/modules/LibraConfig.move
```diff
@@ -164,6 +167,31 @@ module LibraConfig {
...
+        Roles::assert_libra_root(lr_account);
+        move_to(lr_account, DisableReconfiguration {} )
+    }
+
+    /// Private function to resume reconfiguration.
+    /// This function should only be used for offline WriteSet generation purpose and should never be invoked on chain.
+    fun enable_reconfiguration(lr_account: &signer) acquires DisableReconfiguration {
+        assert(Signer::address_of(lr_account) == CoreAddresses::LIBRA_ROOT_ADDRESS(), Errors::requires_address(ELIBRA_CONFIG));
+
+        if(exists<DisableReconfiguration>(Signer::address_of(lr_account))) {
```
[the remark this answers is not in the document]
runtian-zhou: Fixed!

tzakian (approved):

sausagee:
@runtian-zhou, I wonder if this is actually a breaking change.

runtian-zhou:
> @runtian-zhou, I wonder if this is actually a breaking change.

This PR changes the genesis module. However it should not change the behavior of validator software as we are implementing an offline tooling.

PR #6712, opened 2020-11-17
[network] Simplify Network Healthcheck

gregnazario:
### Overview
The network healtcheck was randomly picking a peer to healthcheck.  If
there are a large number of peers, the likelihood of healthchecking a
peer at a regular rate diminishes.  It was as strange behavior that
really doesn't healthcheck at a regular rate.  This now, healthchecks
all peers at a regular rate, creating different nonces for each.  The
nonces are to prevent spoofs of health check responses.

### Why do this now?
It was actually just bothering me to explain that our health checks are non-deterministic.  Now, it's more akin to a regular health check, where pings go out at a regular rate, and are kept track of for each peer that it's healthchecking.

### But wait, won't this create a lot of extra traffic?
Yes, but with the changes coming up for connection limits, we should have a limited amount of healthchecks.  This will prevent us basically causing excess load just from a large number of connections connected to a peer.

### DIFF
--- a/network/src/protocols/health_checker/mod.rs
+++ b/network/src/protocols/health_checker/mod.rs
@@ -39,7 +39,7 @@ use libra_config::network_id::NetworkContext;
 use libra_logger::prelude::*;
 use libra_metrics::IntCounterVec;
 use libra_types::PeerId;
-use rand::{rngs::SmallRng, seq::SliceRandom, Rng, SeedableRng};
+use rand::{rngs::SmallRng, Rng, SeedableRng};
 use serde::{Deserialize, Serialize};
 use std::{collections::HashMap, sync::Arc, time::Duration};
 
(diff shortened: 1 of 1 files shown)


network/src/protocols/health_checker/mod.rs
```diff
@@ -194,72 +194,72 @@ where
...
-                        None => {
-                            trace!(
-                                NetworkSchema::new(&self.network_context),
-                                round = self.round,
-                                "{} No connected peer to ping round: {}",
-                                self.network_context,
-                                self.round
-                            );
-                        }
+                    if peers.is_empty() {
```
phlip9, line 233: can just `if self.connected.is_empty() {`

network/src/protocols/health_checker/mod.rs
```diff
@@ -194,72 +194,72 @@ where
...
+                        trace!(
+                            NetworkSchema::new(&self.network_context),
+                            round = self.round,
+                            "{} No connected peer to ping round: {}",
+                            self.network_context,
+                            self.round
+                        );
+                    }
+
+                    for peer_id in peers {
```
phlip9, line 243: `for peer_id in self.connected.keys() {`

phlip9 (approved):
LGTM (+/- some nits)!

Maybe let's also reduce the default tick rate a little? Right now it's at 1 "tick" per second, so we make 1 outbound ping per second and expect about 1 inbound ping per second. Currently this diff will make ~#V outbound pings/sec and expect #V inbound pings/sec so like 2 * 25 = 50 health checker msg/sec total. Maybe we instead target ~10/sec, so change the `PING_INTERVAL_MS` to 5000.

davidiw:
Likely a net improvement. Let's check a canary before we land this and validate your sentiment... I think we should also pull in all incoming messages into the health checker as indications of liveness... can you explore how "hard" that would be to do in Rust? I think the other concern is that what happens if we have a toooon of FNs can we argue that the weight of this would be less than a state sync service at steady state?

/canary

gregnazario:
> Likely a net improvement. Let's check a canary before we land this and validate your sentiment... I think we should also pull in all incoming messages into the health checker as indications of liveness... can you explore how "hard" that would be to do in Rust? I think the other concern is that what happens if we have a toooon of FNs can we argue that the weight of this would be less than a state sync service at steady state?

Sure, I can see if we can just use other messages as health checks.

The interesting thing here, is we use these nonces in the health check messages to ensure that health check messages aren't spoofed on the message back, except if we're already connected, I'm not exactly sure where the nonce adds value.

I know if we'll have a ton of FNs we'll have a large number of healthchecks, but they are much lighter weight than anything else we currently send.

gregnazario:
@davidiw Can't run the canary due to docker limits.  We'll have to run this ourselves on cluster test for testing

network/src/protocols/health_checker/mod.rs
gregnazario: Without some rust magic (that i'm not quite aware of yet), I can't do that.

`self.connected.keys()` is an immutable borrow, then `self.nonce()` is a mutable borrow.  So, I'm leaving it with the cloning for now.

PR #6713, opened 2020-11-17
[vm] stop combining all writeset validation errors into one status code

bob-wilson:
Previously most validation errors for WriteSet transactions were all reported with the same REJECTED_WRITE_SET status code. (But, that was not done consistently because errors with INVALID_GAS_SPECIFIER and
CURRENCY_INFO_DOES_NOT_EXIST were still reported separately.) This led to unnecessary special cases in specifying the behavior of WriteSet transactions and also made testing fragile, e.g., the vm-validator's test_validate_non_genesis_write_set was actually failing because of a "too new" sequence number, which is probably not what it was intended to test.

This changes WriteSet transaction errors to be converted to status codes in the same way as any other kind of transaction. The REJECTED_WRITE_SET status is kept but is now only used for EINVALID_WRITESET_SENDER. This leads to two changes in behavior:

1. The status code may be different when an invalid WriteSet transaction fails validation. Since these transactions are invalid, they must be discarded and so the status code change does not break compatibility.

2. More than one WriteSet transaction can be in mempool at once. Previously the special exemption for "too new" sequence numbers during the validation phase did not apply to WriteSet transactions, so that only one of them could be in flight at once. This removes that restriction. This should also be a compatible change.

## Motivation

Make WriteSet transactions more consistent

Yes

## Test Plan

Updated and ran tests


vm-validator/src/unit_tests/vm_validator_test.rs
```diff
@@ -384,16 +384,29 @@ fn test_validate_invalid_arguments() {
 fn test_validate_non_genesis_write_set() {
     let vm_validator = TestValidator::new();
 
+    // Confirm that a correct transaction is validated successfully.
     let address = account_config::libra_root_address();
     let transaction = transaction_test_helpers::get_write_set_txn(
         address,
-        2,
```
runtian-zhou, line 390: Should we check for the txn sequence number too new case as well here?

runtian-zhou (approved):
Seems like a reasonable cleanup. Also since the change would only affect the discarded transaction it would not be a breaking change.

vm-validator/src/unit_tests/vm_validator_test.rs
bob-wilson: There's already a test for that in e2e-testsuite/src/tests/write_set.rs. Also, with this change, "too new" sequence numbers are now OK for validation of WriteSet transactions (just as they are for other kinds of transactions), so there's nothing special to be tested for that. That said, if you think it's important to test here, it wouldn't hurt anything (except time to run the tests) to add another test for that

runtian-zhou: Ah that makes sense! Was just wondering why the test was removed. This sounds good to me then!

PR #6714, opened 2020-11-17
Inbound / Outbound Connection limits

gregnazario:
### Overview

I'm generalizing the inbound / outbound connection limits.  This should let us limit connections inbound to work in tandem with things like rate limiting to prevent too much traffic coming into a single node.  This is generalized across both Validators and Full nodes, as validators do have a limit on their own.

* Inbound limits are handled by PeerManager
* Outbound limits are handled by ConnectivityManager

#### What happens when a connection switches between inbound and outbound
This is problematic because of our P2P view used sometimes for a P2P network and sometimes a Client / Server network.  For now, I'll just assume that it's close enough to the total limits and that it is rare. The tie breaking is always one directional, so we shouldn't have flip flops.  If we find an issue with allowing too many inbound connections become outbound connections and not be counted, we can add either "total" connections, or outbound connections to Peer manager

#### Why put the limit in PeerManager after negotiating connections?
This is for current simplicity and forward thinking, as next add will be to add information from the Eligible peers to be fed into this.  Because ConnectivityManager currently handles this in a separate actor, it becomes unwieldy to have it too far from it.  I may in the future refactor this into PeerManager to help this integration.

#### Next steps
* Add ability to pull in trusted peers if there is a connectivity manager / make it such that connectivity manager doesn't necessarily connect outward on an interface automatically (e.g. public)
* Tune limits
* Optimize lookup of inbound / outbound connected peers

### DIFF
--- a/network/src/peer_manager/mod.rs
+++ b/network/src/peer_manager/mod.rs
@@ -287,6 +287,8 @@ where
     channel_size: usize,
     /// Max network frame size
     max_frame_size: usize,
+    /// Inbound connection limit separate of outbound connections
+    inbound_connection_limit: usize,
 }
 
 impl<TTransport, TSocket> PeerManager<TTransport, TSocket>
@@ -312,6 +314,7 @@ where
         max_concurrent_network_reqs: usize,
         max_concurrent_network_notifs: usize,
         max_frame_size: usize,
+        inbound_connection_limit: usize,
     ) -> Self {
         let (transport_notifs_tx, transport_notifs_rx) = channel::new(
             channel_size,
@@ -350,6 +353,7 @@ where
             max_concurrent_network_notifs,
             channel_size,
             max_frame_size,
+            inbound_connection_limit,
         }
     }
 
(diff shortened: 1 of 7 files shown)


network/src/peer_manager/mod.rs
```diff
@@ -447,9 +451,27 @@ where
...
+                if conn.metadata.origin == ConnectionOrigin::Outbound
+                    || self
+                        .active_peers
+                        .contains_key(&conn.metadata.remote_peer_id)
+                    || inbound_conns < self.inbound_connection_limit
+                {
+                    // Add new peer, updating counters and all
+                    self.add_peer(conn);
+                    self.update_connected_peers_metrics();
+                }
```
phlip9, line 474: Let's move the "New connection established" log into the if branch and add a log to the else branch that tells us the connection was dropped b/c we're at the inbound connection limit.

phlip9 (approved):
LGTM

davidiw (commented):
Curious if we want to push this into a staging repo where we build all of our hardening work and then dump into master or if we want to merge directly into master.

Fantastic work ehre.

gregnazario:
> Curious if we want to push this into a staging repo where we build all of our hardening work and then dump into master or if we want to merge directly into master.
> 
> Fantastic work ehre.

We could do a separate feature branch for the hardening work.  Then we can just rebase the branch and get it back into master

network/src/peer_manager/mod.rs
gregnazario: Now, realized we aren't really closing the connection via the socket, so I'll add that.

PR #6723, opened 2020-11-19
[move-cli] include stdlib sources and binaries in the CLI rust binary to avoid publishing stdlib all the time

mengxu-fb:
## Motivation

In the original workflow, if a CLI needs to check/compile/run a script or module that depends on stdlib, it needs to
`publish ../../../../../stdlib/modules` and `publish ../../../../../stdlib/nursery` first before `run script.move`.

In this PR, we try to simplify the process a bit by incorporating the source and binary files of the stdlib in the CLI rust binary and control dependencies with a `--mode` flag. So if the `script.move` has a dependency to the stdlib, we can simply run it with `run script.move --mode stdlib` without the need to do the `publish` first.

Another flag that is currently supported is `--mode bare` which disables inclusion of the stdlib (or any library) as a whole.

Yes

## Test Plan

Currently we default the `--mode` flag to `bare`, to maintain compatibility with existing test cases. We will change the default to `stdlib` once the PR is approved and the test files are updated.

In the meanwhile, I have added four test cases, `script_check_with_deps`, `script_run_with_deps`, `module_publish_view_with_deps`, and `module_run_view_with_deps` to check that the `--mode stdlib` implementation works as intended.

With these, `cargo xtest` should not fail.


language/tools/move-cli/src/main.rs
```diff
@@ -38,9 +41,20 @@ struct Move {
...
+    /// Directory storing the depended Move packages.
+    #[structopt(long, default_value = DEFAULT_PACKAGE_DIR, global = true)]
+    package_dir: PathBuf,
     /// Directory storing Move resources, events, and module bytecodes produced by script execution.
     #[structopt(long, default_value = DEFAULT_BUILD_DIR, global = true)]
     build_dir: String,
+    /// Dependency inclusion mode
+    #[structopt(
+        long,
+        parse(try_from_str = parse_deps_from_mode_string),
```
tnowacki, line 53: mega nit, I usually like puttin this one at the end

language/tools/move-cli/src/main.rs
```diff
@@ -38,9 +41,20 @@ struct Move {
     /// Directory storing Move resources, events, and module bytecodes produced by script execution.
     #[structopt(long, default_value = DEFAULT_STORAGE_DIR, global = true)]
     storage_dir: String,
+    /// Directory storing the depended Move packages.
+    #[structopt(long, default_value = DEFAULT_PACKAGE_DIR, global = true)]
+    package_dir: PathBuf,
```
tnowacki, line 46: I wouldn't make this an argument right now. No need to be extensible here if we don't need it to be.

language/tools/move-cli/src/main.rs
```diff
@@ -170,8 +185,9 @@ fn publish(args: &Move, files: &[String]) -> Result<OnDiskStateView> {
     if args.verbose {
         println!("Compiling Move modules...")
     }
-    let interface_dir = interface_files_dir(&args.build_dir)?;
-    let (_, compiled_units) = move_lang::move_compile(files, &[interface_dir], None, None)?;
+    let mut deps = args.mode.prepare_source_files(&args.package_dir)?;
```
tnowacki, line 188: I think this might be a bit easier if we make this a function for the `Move` struct, so it can just be `args.mode_source_files()`

language/tools/move-cli/src/main.rs
```diff
@@ -625,6 +642,12 @@ fn main() -> Result<()> {
                 fs::remove_dir_all(&storage_dir)?;
             }
 
+            // delete packages
+            let package_dir = Path::new(&move_args.package_dir);
```
tnowacki, line 646: I'll leave this comment elsewhere too, but we might want to put the package stuff either in storage. 

I like the feeling of it sitting in storage as like "here is this magical elsewhere code that we are preloading for you"

tnowacki: @shb thoughts?

language/tools/move-cli/src/package.rs
```diff
@@ -0,0 +1,227 @@
...
+                let file_path = pkg_path.join(file.path());
+                let mut fp = File::create(file_path)?;
+                fp.write_all(file.contents())?;
+            } else {
+                bail!(
+                    "Unexpected directory name ending with the MOVE extension: {}",
+                    entry.path().display()
+                );
+            }
+        }
```
tnowacki, line 92: We don't need all this code, the source language compiler can find all the files for you, you just need to point it to the right subdirectory, in this case `package/source_files/`

language/tools/move-cli/src/package.rs
```diff
@@ -0,0 +1,227 @@
...
+            })
+            .collect();
+        Ok(compiled_modules)
+    }
+}
+
+/// The dependency interface exposed to CLI main
+pub struct MoveDep(Vec<&'static MovePackage>);
+
+impl MoveDep {
```
tnowacki, line 195: I think this should be called mode? If I'm understanding it's purpose properly

language/tools/move-cli/src/package.rs
```diff
@@ -0,0 +1,227 @@
...
+    pub fn prepare_source_files(&self, dir_path: &Path) -> Result<Vec<String>> {
+        let mut paths = vec![];
+        for pkg in &self.0 {
+            paths.extend(pkg.prepare_source_files(dir_path)?);
+        }
+        Ok(paths)
```
tnowacki, line 201: self.0.iter().map(|pkg| pkg.prepare_source_files()).flatten().collect() should work here

language/tools/move-cli/src/package.rs
```diff
@@ -0,0 +1,227 @@
...
+            modules.extend(pkg.load_modules(dir_path)?);
+        }
+        Ok(modules)
+    }
+}
+
+pub fn parse_deps_from_mode_string(mode: &str) -> Result<MoveDep> {
+    match mode {
+        "bare" => Ok(MoveDep(vec![])),
+        "stdlib" => Ok(MoveDep(vec![&PACKAGE_NURSERY])),
```
tnowacki, line 224: I don't think this is quite the split we want. 

There should be three modes 
"bare" -> empty as it is here today
"stdlib" -> [Vector, Option, Errors, Events, ..Nursery]
"libra" -> [/* Everything else */, ..Stdlib]

language/tools/move-cli/src/package.rs
```diff
@@ -0,0 +1,227 @@
...
+
+    /// Load modules from the given directory
+    pub fn load_modules(&self, dir_path: &Path) -> Result<Vec<CompiledModule>> {
+        let compiled_modules =
+            find_filenames(
+                &[dir_path.to_str().unwrap().to_owned()],
+                |fpath| match fpath.extension().and_then(|s| s.to_str()) {
+                    None => false,
+                    Some(ext) => ext == COMPILED_EXTENSION,
+                },
```
tnowacki, line 181: Let's make `move_lang::extension_equals` public which will clean this up a bit

language/tools/move-cli/src/package.rs
```diff
@@ -0,0 +1,227 @@
...
+    deps: &'static [&'static MovePackage],
+}
+
+impl MovePackage {
+    /// Build the stdlib package statically
+    const fn pkg_stdlib() -> Self {
+        MovePackage {
+            name: "stdlib",
+            source_dir: include_dir!("../../stdlib/modules"),
+            binary_dir: Some(include_dir!("../../stdlib/compiled/stdlib")),
```
tnowacki, line 44: Don't include this, we need to regenerate them

language/tools/move-cli/src/package.rs
```diff
@@ -0,0 +1,227 @@
...
+/// Content for the nursery package
+const PACKAGE_NURSERY: MovePackage = MovePackage::pkg_nursery();
+
+struct MovePackage {
+    /// Name of the package
+    name: &'static str,
+    /// The directory containing all the .move source files
+    source_dir: Dir<'static>,
+    /// The directory containing all the .mv binary files, if available
+    binary_dir: Option<Dir<'static>>,
```
tnowacki, line 33: Don't include this, we will just regenerate them

language/tools/move-cli/src/package.rs
```diff
@@ -0,0 +1,227 @@
...
+                );
+            }
+        }
+
+        // done
+        Ok(dep_dirs)
+    }
+
+    /// Lay down the binary files to a given directory
+    pub fn prepare_binary_files(&self, dir_path: &Path) -> Result<Vec<String>> {
```
tnowacki, line 99: I don't think this should be public at all. We just need `load_modules` to be public

language/tools/move-cli/src/package.rs
```diff
@@ -0,0 +1,227 @@
...
+        MovePackage {
+            name: "nursery",
+            source_dir: include_dir!("../../stdlib/nursery"),
+            binary_dir: None,
+            deps: &[&PACKAGE_STDLIB],
+        }
+    }
+
+    /// Lay down the source files to a given directory
+    pub fn prepare_source_files(&self, dir_path: &Path) -> Result<Vec<String>> {
```
tnowacki, line 60: I think the code for generating the source files/module files can be a lot simpler if we make it less parameterized

We should just assume our directory structure and always write to it. This goes with removing that command line argument I mentioned earlier

tnowacki (commented):
Overall it looks like a great start!

Highlevel things we will need to clean up
- We need 3 modes bare/stdlib/libra 
- Removing the parameterized dir is going to let you clean a lot of things up (likewise on not storing the compiled binaries with include_dir, just keep the sources)
- Slimming down the API and letting some of the compiler do the work should also help too

Thanks for the quick start on this!

language/tools/move-cli/src/main.rs
mengxu-fb: Yup, I thought about that too. Will make the changes.

language/tools/move-cli/src/package.rs
mengxu-fb: I see. The reason I have this glob thing is because in the `language/stdlib/modules` there are several `.md` files which I don't want to splash over to the `package/<name>/source_files`. But you are right that even we have those `.md` files there, the compiler can filter them out.

language/tools/move-cli/src/package.rs
mengxu-fb: Yes, it servers the purpose of mode, but it also has the feeling of being a collection of the dependencies... I am pretty bad at giving it names...

language/tools/move-cli/src/package.rs
mengxu-fb: Thank you! This is precisely the things I would like to ask you and Tim. I also noticed that there is a bundle of transaction scripts sitting in stdlib, should they belong to `libra` too? Or we don't need to worry about them?

language/tools/move-cli/src/package.rs
mengxu-fb: Will do

language/tools/move-cli/src/package.rs
mengxu-fb: You mean, compile them when, say, we first call `load_modules` on this package?

language/tools/move-cli/src/main.rs
tzakian: +1 to keeping these package dirs around for the same reason. I also think in the eventual setting where you can have user-defined modes, having them be actual directories that you can see will be nice to have.

language/tools/move-cli/src/package.rs
tzakian: I don't think you need to worry about the transaction scripts. They're predefined since custom scripts aren't allowed to be sent to the chain, but for the CLI they are.

mengxu-fb: Thanks for the clarification Tim.

mengxu-fb: So, I plan to put everything in the [Module Utility Libraries](https://github.com/libra/libra/blob/master/language/stdlib/modules/doc/overview.md#module-utility-libraries) plus nursery as `stdlib` and the rest as `libra`.

Right now I guess I have to hardcode this information though...

tnowacki: I think we will always have to hard code it unless we split up or directories. Some work for another time @tzakian

language/tools/move-cli/src/main.rs
tnowacki: After sleeping on it, I think for now we could just have a single hardcoded directory, that sits outside of build or storage. We can deal with custom packages later

language/tools/move-cli/src/package.rs
tnowacki: Ah, I think I misunderstood the code. Instead of doing the glob, I would just case on file extension instead of doing this glob filter. So just walk over every entry in the dir and filter out those with the right extension. Is that possible?

language/tools/move-cli/src/package.rs
tnowacki: I would compile them ahead of time to keep it simple, but yeah thats the idea

language/tools/move-cli/src/package.rs
tzakian: Agreed. I think for now this will need to be hardcoded, but eventually we'll want to split this up.

language/tools/move-cli/src/package.rs
```diff
@@ -0,0 +1,227 @@
...
+        // write file content
+        fs::create_dir_all(&pkg_path)?;
+
+        let glob = format!("**/*.{}", MOVE_EXTENSION);
+        for entry in self.source_dir.find(&glob).unwrap() {
+            if let DirEntry::File(file) = entry {
+                let file_path = pkg_path.join(file.path());
+                let mut fp = File::create(file_path)?;
+                fp.write_all(file.contents())?;
+            } else {
```
tnowacki, line 86: I really loathe if-let with an else, I think a `match` would do better here :)

language/tools/move-cli/src/package.rs
mengxu-fb: That is certainly doable, changes on the way.

language/tools/move-cli/src/main.rs
metajack: Did I miss some conversation that happened about this? I was coming to leave a comment that it should probably live in `build`. It is required to be a file on disk because of the compiler API it seems, which is a build-time thing, so that seems the appropriate place to put it. We should avoid creating too many top level directories in the users code to avoid name clashes and ugliness in `ls` output etc.

language/tools/move-cli/src/package.rs
metajack: You'll probably want to either compile them least every time the mtime changes or every time you call `prepare_source_files` if you don't want to deal with tracking mtimes.

language/tools/move-cli/src/package.rs
```diff
@@ -0,0 +1,227 @@
...
+    /// Dependencies
+    deps: &'static [&'static MovePackage],
+}
+
+impl MovePackage {
+    /// Build the stdlib package statically
+    const fn pkg_stdlib() -> Self {
+        MovePackage {
+            name: "stdlib",
+            source_dir: include_dir!("../../stdlib/modules"),
```
metajack, line 43: This will need to be a hardcoded list to not grab all of LF.

language/tools/move-cli/src/package.rs
metajack: I think your opinion might be an outlier in Rust land, although I think I'd agree with it the case where you cared about more than one variant. `if let Some(foo) = bar { x } else { y }` is pretty natural.  But that does not apply to most uses of `Result` at all, where `match` is clearly warranted.

In any case, if you and the rest of the Move team feel strongly about this pattern, I'd suggest creating a coding guidelines doc with these kinds of things, otherwise you'll end up with inconsistent changes based on who the reviewers are.

language/tools/move-cli/src/package.rs
mengxu-fb: I think with `include_dir!`, we are statically include a snapshot of the file content with our CLI binary at build time, so I guess we do not need to worry about `mtimes`? Correct me if I misinterpret the message.

metajack: I guess I was assuming if the files existed on disk, it was to support some workflow where you'd update them without having to recompile the CLI, but I suppose you'd just use `--mode bare` for that. If the files just exist because the compiler API can only take a `Vec<String>` as input, perhaps we should consider support `Vec<Box<dyn trait io::Read>>` to avoid round tripping to disk.

language/tools/move-cli/src/main.rs
tnowacki: Perhaps living in build isn't too bad.. but originally that had intended to be the directory for the output from `move build`

metajack: I was thinking of it like Cargo's `target` where lots of things the "build" process deals with are stored there, not just binary artifacts, but also timestamp data, generated code, etc.

mengxu-fb:
Major changes in this update:

1. Improved the split between `stdlib` and `libra`:
Right now, the following modules + all nursery goes to `--mode stdlib` and the rest goes to `--mode libra`:
```
    "CoreAddresses.move",
    "Errors.move",
    "Event.move",
    "FixedPoint32.move",
    "Hash.move",
    "LCS.move",
    "Option.move",
    "Vector.move",
    "Signer.move",
```

2. Freshly compile the source code to modules instead of loading the pre-built ones.

3. Introduced a mode preparation procedure before relevant commands, i.e., `check`, `publish`, and `run`. For `check` and `publish`, the source files will be populated and for `run`, both the source files and the compiled modules will be loaded. This is controlled by the `Move::prepare_mode(source_only: bool)` function.

4. Using `build/package` as the workspace instead of having `package/` as a top-level directory. Per each package, the sources and binaries are stored in `<package-name>/[source_files | compiled]` respectively.

mengxu-fb:
A major concern I have now is that if we have multiple `run`s in the `args.txt` and interpret it with `move-cli test`, these library modules will be written to `storage` twice.

To be specific, given the following `args.txt`:

```
run scriptA.move --mode stdlib
run scriptB.move --mode stdlib
```

The stdlib modules will be persisted by the `OnDiskStateView` twice, one per each `run`, which is clearly a waste of effort. But currently `OnDiskStateView` cannot differentiate whether a module is a library module or a user-published module...

language/tools/move-cli/src/package.rs
```diff
@@ -0,0 +1,258 @@
...
+const PKG_BINARY_DIR: &str = "compiled";
+
+/// Content for the stdlib directory
+const DIR_STDLIB: Dir = include_dir!("../../stdlib/modules");
+/// Content for the nursery directory
+const DIR_NURSERY: Dir = include_dir!("../../stdlib/nursery");
+
+/// The portion in the stdlib directory that are considered stdlib
+const STDLIB_MODULE_FILES: [&str; 9] = [
+    "CoreAddresses.move",
```
tzakian, line 33: I've remove CoreAddresses from this list as it's really defining the key addresses for the libra side of things.

language/tools/move-cli/src/package.rs
```diff
@@ -0,0 +1,258 @@
...
+    deps: vec![&PACKAGE_STDLIB],
+});
+
+struct SourceFilter<'a> {
+    /// The embedded directory
+    source_dir: &'a Dir<'a>,
+    /// Source files to be included, if set to None, include everything
+    inclusion: Option<HashSet<&'a str>>,
+    /// Source files to be excluded, to exclude nothing, set it to empty
+    exclusion: HashSet<&'a str>,
```
tzakian, line 78: For consistency it might be best to make these both nops if they're the empty hashset, or they're both set to none. I have a personal preference for `None`, but I'm fine either way as long as they're consistent.

language/tools/move-cli/src/package.rs
```diff
@@ -0,0 +1,258 @@
...
+                    CompiledUnit::Module { ident, module, .. } => {
+                        let mut data = vec![];
+                        module.serialize(&mut data)?;
+                        let file_path = pkg_bin_path
+                            .join(ident.0.value.name.0.value)
+                            .with_extension(COMPILED_EXTENSION);
+                        let mut fp = File::create(file_path)?;
+                        fp.write_all(&data)?;
+                    }
+                    CompiledUnit::Script { loc, .. } => println!(
```
tzakian, line 181: Nit: since this is a warning it should probably go to stderr (`eprintln!`)

tzakian (approved):
Overall LGTM! A couple small nits/changes before landing is all.

language/tools/move-cli/src/package.rs
mengxu-fb: I see the consistency argument. I can additionally make `exclusion` an Option as well and setting it to `None` will be the same as giving it an empty set. But that gives a bit inconsistency with the `None` for `inclusion` as well because a `None` in `inclusion` means include all but a `None` in `exclusion` means exclude nothing... So...

mengxu-fb:
Update: added a new commit which now makes `stdlib` as the default mode instead of `bare`. This commit also updates the testcases accordingly:
- if a test case does not depends on `stdlib`, add a `--mode bare` to its command line
- if a test case depends on `stdlib` but not `libra`, remove the `publish ../../../../../stdlib/modules` commands
- if a test case depends on `libra`, remove the `publish ...` and also add a `--mode libra` to its command line

language/tools/move-cli/src/package.rs
tzakian: Good point. I guess I was viewing it from the "what value means you should ignore this field's value" and the `None` made more sense that viewpoint, but I can see what you mean how it could also be confusing.

 I'm fine keeping it as-is if you would prefer to keep it this way, so feel free to land :)

mengxu-fb:
Anyone wants to have a final look and help me with landing the PR? :)

tzakian:
I'll land.

/land

language/tools/move-cli/src/main.rs
sblackshear: Generalizing `build` to a `target`-like directory for generated stuff seems like a good idea to me.

mengxu-fb: In this PR, we did re-use the `build` dir and produces the package data under `build/package/<package-name>/[source_files | compiled]` to reduce the number of top-level directories, as Jack and Todd suggested.

PR #6725, opened 2020-11-19
[Consensus] Add smoke tests for consensus observer mode

JoshLind:
## Motivation

This PR adds a simple smoke test to ensure that consensus observer mode is working (and continues to be supported going forward). To do this, we force a failure in safety rules on a single node in a cluster and ensure that it continues to observe transactions being executed in the cluster.

Yes.

## Test Plan

All tests pass locally (including the newly added one).

## Related PRs

This PR adds tests for functionality that was changed in https://github.com/libra/libra/pull/6671 and https://github.com/libra/libra/pull/6694.

### DIFF
--- a/testsuite/smoke-test/src/consensus.rs
+++ b/testsuite/smoke-test/src/consensus.rs
@@ -1,0 +1,56 @@
+// Copyright (c) The Libra Core Contributors
+// SPDX-License-Identifier: Apache-2.0
+
+use crate::{
+    operational_tooling::launch_swarm_with_op_tool_and_backend,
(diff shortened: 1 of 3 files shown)


JoshLind:
Thanks @zekun000! 😄 

> does the test fail without the observer mode?

Yeah, it does. If safety rules fails to initialize() (e.g., due to the key not being found, or data missing in storage) and it panics, the final assertion in the test fails because the node can't see the correct latest sequence number (i.e., it's blocked from observing). Now that we don't panic on a failure to initialize(), the test passes.

PR #6726, opened 2020-11-19
[Cherry-pick] Consensus shouldn't panic on a failure to initialize safety rules

JoshLind:
## What this affects:

This PR changes the way that consensus interacts with safety rules when safety rules fails to initialize under an epoch change:
- At present, if safety rules fails to initialize when an epoch changes (e.g., due to not being able to find the expected consensus key in storage), consensus will panic and stop operating.
- Instead, consensus should not panic but continue to operate and simply observe the progress being made by the rest of the network. This is what we already do in some cases today (e.g., if the validator is not in the validator set for the new epoch).

## Why is this critical:

There are several practical reasons why safety rules might fail to initialize under a new epoch change. A notable one is when secure storage no longer contains the consensus key expected for that epoch. The reasons for this might include:
- The validator has been migrated from one operator to another, and thus the new operator does not have access to the old consensus keys.
- The validator has been stuck for a temporary amount of time (e.g., due to another issue), and then tries to come back online. However, during the time that it was stuck, the consensus keys were trimmed from secure storage, meaning the old consensus keys (for epochs in the past) are no longer available to safety rules.
- Operators leaving the key manager running during an upgrade process that spans multiple days, thus removing old consensus keys from secure storage.
- ...

Unfortunately, we are seeing validator operators hitting these cases in practice, and thus their validators are becoming blocked until they perform a really heavy handed workaround (see below).

## What is the workaround (even if it is unintuitive / horrible):

The workaround for this issue currently requires 4 steps:
1. First, stop the key manager and perform a consensus key rotation manually by hand. This will force the blockchain to update to a new epoch, E.
2. Next, run a backup and restore operation on the node (to restore the blockchain to epoch version >= epoch E).
3. Next, grab a waypoint (that is >= epoch E) from another node manually, and use the operator tool to insert it into secure storage).
4. Finally, restart the validator, safety rules and the key manager.

## How have we tested these changes?

These changes are tested by existing unit tests and a newly added smoke test.

## Related PRs

This PR cherry-picks the changes implemented by the following PRs:
- https://github.com/libra/libra/pull/6671
- https://github.com/libra/libra/pull/6694
- https://github.com/libra/libra/pull/6725

### DIFF
--- a/consensus/safety-rules/src/error.rs
+++ b/consensus/safety-rules/src/error.rs
@@ -29,10 +29,16 @@ pub enum Error {
     InvalidQuorumCertificate(String),
     #[error("{0} is not set, SafetyRules is not initialized")]
     NotInitialized(String),
-    #[error("Error returned by secure storage: {0}")]
-    SecureStorageError(String),
+    #[error("Data not found in secure storage: {0}")]
+    SecureStorageMissingDataError(String),
+    #[error("Unexpected error returned by secure storage: {0}")]
+    SecureStorageUnexpectedError(String),
     #[error("Serialization error: {0}")]
     SerializationError(String),
+    #[error("Validator key not found: {0}")]
+    ValidatorKeyNotFound(String),
+    #[error("The validator is not in the validator set. Address not in set: {0}")]
+    ValidatorNotInSet(String),
     #[error("Vote proposal missing expected signature")]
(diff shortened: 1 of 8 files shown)


sausagee (approved):
/land

Issue #6082, opened 2020-09-17
Require currency codes to use uppercase characters


bob-wilson:
ISO 4217 requires currency codes to be composed of 3 uppercase alphabetic characters. I am currently proposing that we restrict Libra currency codes to be arbitrary length alphanumeric strings that start with a letter (https://github.com/libra/libra/pull/6081). We cannot easily restrict them more right now because of the "Coin1" and "Coin2" placeholder currencies that use lowercase, numbers, and are more than 3 characters long. I'm not sure we need to restrict ourselves on the length or the use of numbers, but after those placeholders are gone, it would be good to at least require uppercase characters, since there is a potential for confusion and ambiguity with mixing uppercase and lowercase characters. For example, if someone mistakenly specifies "usd" instead of "USD", they may be surprised by the results.

bob-wilson:
@tzakian fixed this in #7252

PR #6727, opened 2020-11-19
[diem framework] Address audit comments for prologue and epilogue

tzakian:
This PR adds the following checks to the prologue:
* That the sequence number is not max u64
* That the currency that is specified for the transaction fee is not only a registered currency, but also a currency that transaction fees can be paid in.

It also adds a check to the `writeset_epilogue` to make check that the account is the LR account.


mickayz (approved):
LGTM!

wrwg (approved):

language/stdlib/modules/DiemAccount.move
```diff
@@ -1650,27 +1652,38 @@ module DiemAccount {
...
 
-        // [PCA10]: Check that the transaction's sequence number matches the
+        // [PCA11]: Check that the transaction's sequence number matches the
         // current sequence number. Otherwise sequence number is too new by [PCA8].
         assert(
             txn_sequence_number == sender_account.sequence_number,
             Errors::invalid_argument(PROLOGUE_ESEQUENCE_NUMBER_TOO_NEW)
         );
 
+        // [PCA12]: Check that the transaction's sequence number will not overflow.
```
bob-wilson, line 1681: It looks like this change missed the "WARNING" comment below. It also missed updating the specification, which I'm working to correct now. I'll reorder this check at the same time.

language/stdlib/modules/DiemAccount.move
```diff
@@ -1650,27 +1652,38 @@ module DiemAccount {
                 balance_amount >= max_transaction_fee,
                 Errors::invalid_argument(PROLOGUE_ECANT_PAY_GAS_DEPOSIT)
             );
+            // [PCA8]: Check that the gas fee can be paid in this currency
+            assert(
+                TransactionFee::is_coin_initialized<Token>(),
+                Errors::invalid_argument(PROLOGUE_ECANT_PAY_GAS_DEPOSIT)
```
bob-wilson, line 1658: IMO, this should really have a separate error code. This one translates to a validation status of `INSUFFICIENT_BALANCE_FOR_TRANSACTION_FEE`. But, this check has nothing to do with the account's balance, and it would be confusing to report that when an account has a sufficient balance of the specified gas currency. I will add a new error code for this.

language/stdlib/modules/DiemAccount.move
tzakian: Good spots. 

My understanding is that this ordering is to prevent the "SEQUENCE_NUMBER_TOO_NEW" from clobbering other errors (cf. https://github.com/diem/diem/issues/6321). At the time I made this change my thinking is that this "sequence number overflows a u64" case would be so uncommon as to warrant being at the end, but reordering it to be above this check for consistency is probably the right call (happy to do this as well). Also, cc @runtian-zhou who made this change to comment.

About the specification, I'm not sure if I follow? The specification for the prologue was updated with this change, since otherwise this wouldn't have been able to land since verification of the modules is a land blocking test.

language/stdlib/modules/DiemAccount.move
tzakian: One other option here would be to have this translate out to an `INVALID_GAS_SPECIFIER` validation status. This seems  in line with the error (that the gas currency provided is not registered as a "gas currency") while not introducing a new validation status code.

language/stdlib/modules/DiemAccount.move
bob-wilson: See #7063 
I was referring to the Move Adapter specification document. I agree that the specific problem with overflowing the sequence number is unlikely to occur in practice, but it's important to have that "WARNING" comment in the right place so that no other checks are inserted after the "too new" sequence number check.

language/stdlib/modules/DiemAccount.move
bob-wilson: `INVALID_GAS_SPECIFIER` describes the case where the specifier is not a valid identifier (or perhaps with https://github.com/libra/libra/issues/6082 that it uses lowercase characters). I don't think it would be good to combine that with checking if the currency is registered for transaction fees.

PR #6736, opened 2020-11-20
[rust] upgrade to 1.48.0

rexhoffman:
## Motivation

Keeping up to date with rust.

Yes

## Test Plan

All the CI

## Related PRs

None


language/move-prover/spec-lang/src/env.rs
```diff
@@ -2553,12 +2553,14 @@ impl<'env> FunctionEnv<'env> {
     }
 
     /// Returns the function name excluding the address and the module name
+    //TODO: return slice rather than owned strings.
+    #[allow(clippy::rc_buffer)]
```
rexhoffman, line 2557: @dariorussi or whomever in move land, some of these lints were a little to large for me to want to roll them in to this update, so marked with allow.    At some point the debt will need addressed.    Seems RC/ARC should be used with slices rather than vecs (and str rather than String).

network/src/peer/test.rs
```diff
@@ -380,7 +381,7 @@ fn peer_terminates_when_request_tx_has_dropped() {
     // Perform yamux handshake in a different task.
     let drop = async move {
         // Do not drop connection.
-        ManuallyDrop::new(connection);
+        forget(connection);
```
rexhoffman, line 384: @metajack @bmwill this is a tad interesting.

language/move-prover/bytecode/src/stackless_bytecode.rs
```diff
@@ -484,6 +484,7 @@ impl<'env> BytecodeDisplay<'env> {
         Ok(())
     }
 
+    #[allow(clippy::rc_buffer)]
```
bmwill, line 487: Can we add this to the x.toml instead of here, that way we know we can go and and clean this up later (if needed)

language/testing-infra/functional-tests/src/common.rs
```diff
@@ -95,10 +95,7 @@ impl<T> AsRef<T> for LineSp<T> {
...
 /// Checks if `s` starts with `prefix`. If yes, returns a reference to the remaining part
 /// with the prefix stripped away.
+/// please just use: `s.strip_prefix(prefix)`
 pub fn strip<'a>(s: &'a str, prefix: &str) -> Option<&'a str> {
-    if s.starts_with(prefix) {
-        Some(&s[prefix.len()..])
-    } else {
-        None
-    }
+    s.strip_prefix(prefix)
```
bmwill, line 100: Can we just replace all instances of `strip` with `strip_prefix`?

network/src/peer/test.rs
bmwill: Which lint did you hit with this? @phlip9 would have context on this part of the codebase

phlip9: Agghh I thought I purged all the `ManuallyDrop`... looks like I missed a spot.

```diff
@@ -365,15 +365,13 @@ fn peer_terminates_when_request_tx_has_dropped() {
     let (
         peer,
         peer_handle,
-        connection,
+        _connection,
         _peer_notifs_rx,
         _peer_rpc_notifs_rx,
         _peer_direct_send_notifs_rx,
     ) = build_test_peer(rt.handle().clone(), ConnectionOrigin::Inbound);
     // Perform yamux handshake in a different task.
     let drop = async move {
-        // Do not drop connection.
-        ManuallyDrop::new(connection);
         // Drop peer handle.
         drop(peer_handle);
     };

```

rexhoffman: This line wasn't a lint, but a warning I cleaned up at the same time as the other,
```
unused return value of `ManuallyDrop::<T>::new` that must be used
`#[warn(unused_must_use)]` on by default
if you don't need the wrapper, you can use `mem::forget` insteadrustc(unused_must_use)
```

network/src/peer/test.rs
```diff
@@ -191,7 +191,8 @@ fn peer_send_message() {
 
     let client = async move {
         peer_handle.send_message(send_msg, PROTOCOL).await.unwrap();
-        ManuallyDrop::new(peer_handle);
+        //ManuallyDrop::new(peer_handle);
+        forget(peer_handle);
```
rexhoffman, line 195: This wasn't a lint, but a compiler warning:

unused return value of `ManuallyDrop::<T>::new` that must be used
`#[warn(unused_must_use)]` on by default
if you don't need the wrapper, you can use `mem::forget` insteadrustc(unused_must_use)

language/move-prover/bytecode/src/stackless_bytecode.rs
rexhoffman: Done, wanted to make it a warn but didn't function as a warn, only error....

language/testing-infra/functional-tests/src/common.rs
rexhoffman: Done.

network/src/peer/test.rs
rexhoffman: @phlip9 done, didn't know you could tell rust to ignore memory....

network/src/peer/test.rs
```diff
@@ -191,8 +191,6 @@ fn peer_send_message() {
 
     let client = async move {
         peer_handle.send_message(send_msg, PROTOCOL).await.unwrap();
-        //ManuallyDrop::new(peer_handle);
-        forget(peer_handle);
```
rexhoffman: @phlip9 I suspect this is safe to remove, neither the ManuallyDrop nor the forget were needed for the test to pass.   Don't know if this introduces any kind of race though?

phlip9: Yeah just change this to

```rust
     let (
         peer,
         peer_handle,
         _connection,
         _peer_notifs_rx,
         _peer_rpc_notifs_rx,
         _peer_direct_send_notifs_rx,
     ) = build_test_peer(rt.handle().clone(), ConnectionOrigin::Inbound);
     // Perform yamux handshake in a different task.
     let drop = async move {
         // Drop peer handle.
         drop(peer_handle);
     };
```

phlip9: Oh that's what you already did, oops 😅 

The original intention was to make sure the peer actor shutdown correctly if you dropped the peer_handle (but not the underlying connection). Of course, we can actually accomplish that without the `ManuallyDrop` by just changing `connection` to `_connection` so it drops at the end of the test : )

bmwill (approved):

PR #6737, opened 2020-11-20
[network] Support HTTP CONNECT proxy for Libranet

mgorven:
Some deployment environments will require use of a proxy for outgoing
connections. This adds support for an HTTP proxy on outgoing Libranet
connections. There don't seem to be any good libraries for doing this so
I've implemented the CONNECT request/response manually. The proxy
configuration is taken from the `https_proxy` environment variable.

Test Plan: Ran public fullnode in environment requiring a proxy,
connected successfully.

### DIFF
--- a/network/netcore/src/transport/tcp.rs
+++ b/network/netcore/src/transport/tcp.rs
@@ -4,12 +4,12 @@
 //! TCP Transport
 use crate::{compat::IoCompat, transport::Transport};
 use futures::{
-    future::{self, Future},
+    future::{self, Either, Future},
     io::{AsyncRead, AsyncWrite},
     ready,
     stream::Stream,
 };
-use libra_network_address::{parse_dns_tcp, parse_ip_tcp, IpFilter, NetworkAddress};
+use libra_network_address::{parse_dns_tcp, parse_ip_tcp, parse_tcp, IpFilter, NetworkAddress};
 use libra_types::PeerId;
 use std::{
     convert::TryFrom,
@@ -20,7 +20,11 @@ use std::{
     task::{Context, Poll},
     time::Duration,
 };
-use tokio::net::{lookup_host, TcpListener, TcpStream};
+use tokio::{
+    io::{AsyncReadExt, AsyncWriteExt},
+    net::{lookup_host, TcpListener, TcpStream},
+};
+use url::Url;
 
 /// Transport to build TCP connections
 #[derive(Debug, Clone, Default)]
@@ -105,8 +109,26 @@ impl Transport for TcpTransport {
             .or_else(|| parse_dns_tcp(protos).map(|_| ()))
             .ok_or_else(|| invalid_addr_error(&addr))?;
 
+        let proxy_addr = std::env::var("https_proxy")
+            .ok()
+            .and_then(|https_proxy| Url::parse(&https_proxy).ok())
+            .and_then(|url| {
+                if url.has_host() && url.scheme() == "http" {
+                    Some(format!(
+                        "{}:{}",
+                        url.host().unwrap(),
+                        url.port_or_known_default().unwrap()
+                    ))
+                } else {
+                    None
+                }
+            });
+
         let f: Pin<Box<dyn Future<Output = io::Result<TcpStream>> + Send + 'static>> =
-            Box::pin(resolve_and_connect(addr));
+            Box::pin(match proxy_addr {
+                Some(proxy_addr) => Either::Left(connect_via_proxy(proxy_addr, addr)),
+                None => Either::Right(resolve_and_connect(addr)),
+            });
 
         Ok(TcpOutbound {
             inner: f,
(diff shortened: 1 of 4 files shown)


network/netcore/src/transport/tcp.rs
```diff
@@ -105,8 +109,29 @@ impl Transport for TcpTransport {
...
+                            url.port_or_known_default().unwrap()
+                        ))
+                    } else {
+                        None
+                    }
+                }
+                Err(_e) => None,
+            },
+            Err(_e) => None,
+        };
```
phlip9, line 128:
```rust
let proxy_addr = std::env::var("https_proxy")
    .ok()
    .and_then(|https_proxy| Url::parse(&https_proxy).ok())
    .and_then(|url| {
        if url.has_host() && url.scheme() == "http" {
            Some(format!(
                "{}:{}",
                url.host().unwrap(),
                url.port_or_known_default().unwrap(),
            ))
        } else {
            None
        }
    });
```

phlip9 (approved):
Awesome, thanks! Feel free to land after fixing nits. Do we need to cherry pick this as well?

network/netcore/src/transport/tcp.rs
```diff
@@ -163,6 +188,37 @@ async fn resolve_and_connect(addr: NetworkAddress) -> io::Result<TcpStream> {
...
+async fn connect_via_proxy(proxy_addr: String, addr: NetworkAddress) -> io::Result<TcpStream> {
+    let protos = addr.as_slice();
+
+    if let Some(((host, port), _addr_suffix)) = parse_tcp(protos) {
+        let mut s = TcpStream::connect(proxy_addr).await?;
+        let mut buffer = [0; 1024];
+
+        s.write_all(&format!("CONNECT {0}:{1} HTTP/1.1\r\n\r\n", host, port).into_bytes())
+            .await?;
+        let len = s.read(&mut buffer[..]).await?;
```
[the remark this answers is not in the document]
mgorven: Yeah, this isn't completely robust but should be good enough for now. The FB proxy server response is ~250B and I made the buffer 1024B; CONNECT responses shouldn't be very verbose.

network/netcore/src/transport/tcp.rs
```diff
@@ -163,6 +188,37 @@ async fn resolve_and_connect(addr: NetworkAddress) -> io::Result<TcpStream> {
...
 }
 
+async fn connect_via_proxy(proxy_addr: String, addr: NetworkAddress) -> io::Result<TcpStream> {
+    let protos = addr.as_slice();
+
+    if let Some(((host, port), _addr_suffix)) = parse_tcp(protos) {
+        let mut s = TcpStream::connect(proxy_addr).await?;
+        let mut buffer = [0; 1024];
+
+        s.write_all(&format!("CONNECT {0}:{1} HTTP/1.1\r\n\r\n", host, port).into_bytes())
```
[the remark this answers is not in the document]
mgorven: Can I make it HTTP 1.0 instead? ;-)

network/netcore/src/transport/tcp.rs
```diff
@@ -105,8 +109,29 @@ impl Transport for TcpTransport {
             .or_else(|| parse_dns_tcp(protos).map(|_| ()))
             .ok_or_else(|| invalid_addr_error(&addr))?;
 
+        let proxy_addr = match std::env::var("https_proxy") {
```
gregnazario, line 112: Do we expect to be changing this on the fly?  Or can we load this on startup via the NetworkConfig?  Or at least just load the environment variable once on startup instead of every time.

network/netcore/src/transport/tcp.rs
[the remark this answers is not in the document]
gregnazario: I'd like to see us make a followup github issue to ensure we at least track the places where we don't handle fragmented reads.

network/netcore/src/transport/tcp.rs
gregnazario, line 198: Nit: I'm not a fan of 1 letter variables, because I have to scroll up until I see where it's initialized to tell what it is.

network/netcore/src/transport/tcp.rs
```diff
@@ -105,8 +109,29 @@ impl Transport for TcpTransport {
...
+                            url.host().unwrap(),
+                            url.port_or_known_default().unwrap()
+                        ))
+                    } else {
+                        None
+                    }
+                }
+                Err(_e) => None,
+            },
+            Err(_e) => None,
```
gregnazario, line 127: Nit: Any reason not to just put `_`?  Looks like we're allowing for `https_proxy` but not being strict that all connections must come from it then?

network/netcore/src/transport/tcp.rs
gregnazario: Now, I see this is loading an environment variable and always parsing it the same.

mgorven: No, it's not expected to change, but I figured reading environment variables should be pretty cheap and outgoing Libranet connections are not made that frequently.

network/netcore/src/transport/tcp.rs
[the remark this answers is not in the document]
gregnazario: Also the [RFC](https://www.ietf.org/rfc/rfc2817.txt) :)
```
5.2 Requesting a Tunnel with CONNECT

   A CONNECT method requests that a proxy establish a tunnel connection
   on its behalf. The Request-URI portion of the Request-Line is always
   an 'authority' as defined by URI Generic Syntax [2], which is to say
   the host name and port number destination of the requested connection
   separated by a colon:

      CONNECT server.example.com:80 HTTP/1.1
      Host: server.example.com:80
```

network/netcore/src/transport/tcp.rs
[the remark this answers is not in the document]
phlip9: I think this is the only place?  Maybe we just bite the bullet and drop a quick loop in there...

network/netcore/src/transport/tcp.rs
```diff
@@ -163,6 +185,43 @@ async fn resolve_and_connect(addr: NetworkAddress) -> io::Result<TcpStream> {
...
+                    format!(
+                        "HTTP proxy CONNECT failed: {}",
+                        String::from_utf8_lossy(&msg)
+                    ),
+                ));
+            } else if msg.len() >= 16
+                && (msg.starts_with(b"HTTP/1.1 200") || msg.starts_with(b"HTTP/1.0 200"))
+                && msg.ends_with(b"\r\n\r\n")
+            {
+                break Ok(stream);
```
phlip9, line 217: I found these `break EXPR;` a bit surprising, didn't realise rust supported this : o

let's just replace them with `return` to keep it simple : )

phlip9 (approved):

mgorven:
Just to confirm: does the client always speak first in Libranet?

gregnazario:
> Just to confirm: does the client always speak first in Libranet?

The one initiating the connection should speak first in the initial connection (during the handshake), which would be viewed as the "client" for that time being.  But it is a P2P protocol, so it's a little weird.

phlip9:
> Just to confirm: does the client always speak first in Libranet?

Yeah, the server (currently) always waits for a client noise handshake message immediately after opening the TCP connection

network/netcore/src/transport/tcp.rs
mgorven: @gregnazario Are you okay with this? I actually can't see a good place to do this at startup since the `TcpTransport` is created in `const LIBRA_TCP_TRANSPORT`.

gregnazario: Yeah, let's just do this for now.  Looks like the best way would probably to make a constructor, and load it where the `const` is originally made.

PR #6738, opened 2020-11-20
[network] Centralize LibraNet NetworkMessage framing and IO logic

phlip9:
I'm going to try to spend some time refactoring the Peer actor to reduce the number of actors/tasks and hopefully simplify the life-cycle a bit. This diff mostly moves some of the socket/IO logic out of the Peer actor and centralizes the logic in one place, where we can use it elsewhere. It's also easier to test/fuzz some of the IO logic in isolation like this. In doing so, we can also replace a lot of the bespoke `LengthDelimitedCodec` and `Framed` usage in tests and fuzzers.

+ Add `NetworkMessageStream` and `NetworkMessageSink` adapters around socket-like objects.

+ `NetworkMessageStream` adapts a read socket (`AsyncRead`) into a stream of `NetworkMessage`s and handles all of the underlying framing, deserializing, and reading of `NetworkMessage`s from the socket.

+ Likewise, `NetworkMessageSink` adapts a write socket (`AsyncWrite`) into a sink of `NetworkMessage`s, i.e., something you can just send `NetworkMessage`s on and it will handle the underlying framing, serializing, and writing of `NetworkMessage`s into the socket.

+ The Sink/Stream is split into two separate structs so we can send and receive messages concurrently without coordination. Right now, our Peer actor splits the underlying socket with a `tokio::io::split`, which uses a mutex under-the-hood to serialize acces to the socket. However, with a trait, we can probably get rid of this `tokio::io::split`.

+ If we want, we can remove our dependency on tokio-util's length-delimited codec and hopefully simplify this a bit.

+ Clarify a missing piece in the LibraNet specs around the maximum frame size.

+ The writeable fake sockets now support an option to enable fragmented writes, which forces writes to happen one byte at a time. Like the fragmented read option, this helps ensure our socket logic is properly handling partial reads and writes.

### DIFF
--- a/network/src/testutils/fake_socket.rs
+++ b/network/src/testutils/fake_socket.rs
@@ -24,8 +24,8 @@ use std::{io, pin::Pin};
 pub struct ReadOnlyTestSocket<'a> {
     /// the content
     content: &'a [u8],
-    /// the socket will read byte-by-byte
-    fragmented: bool,
+    /// fragment reads byte-by-byte
+    fragmented_read: bool,
     /// continue to read 0s once content has been fully read
     trailing: bool,
 }
@@ -34,14 +34,14 @@ impl<'a> ReadOnlyTestSocket<'a> {
     pub fn new(content: &'a [u8]) -> Self {
         Self {
             content,
-            fragmented: false,
+            fragmented_read: false,
             trailing: false,
         }
     }
 
-    /// reads will have to be done byte by byte
+    /// reads will be done byte-by-byte
     pub fn set_fragmented_read(&mut self) {
-        self.fragmented = true;
+        self.fragmented_read = true;
     }
 
     /// reads will never return pending, but 0s
@@ -75,8 +75,15 @@ impl<'a> AsyncRead for ReadOnlyTestSocket<'a> {
     fn poll_read(
         mut self: Pin<&mut Self>,
         mut _context: &mut Context,
-        buf: &mut [u8],
+        mut buf: &mut [u8],
     ) -> Poll<io::Result<usize>> {
+        // read/recv on an empty buffer is underspecified
+        assert!(!buf.is_empty());
+
+        if self.fragmented_read {
+            buf = &mut buf[..1];
+        }
+
         // nothing left to read
         if self.content.is_empty() {
             if self.trailing {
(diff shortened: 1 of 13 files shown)


network/src/testutils/fake_socket.rs
```diff
@@ -75,8 +75,12 @@ impl<'a> AsyncRead for ReadOnlyTestSocket<'a> {
     fn poll_read(
         mut self: Pin<&mut Self>,
         mut _context: &mut Context,
-        buf: &mut [u8],
+        mut buf: &mut [u8],
     ) -> Poll<io::Result<usize>> {
+        if self.fragmented_read {
+            buf = &mut buf[..1];
+        }
+
```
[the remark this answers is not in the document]
gregnazario: I'm just saying callers should handle partial reads, and we might miss that they aren't if we're always trimming the buffer,  but it doesn't affect the original buffer, so it isn't an issue.

specifications/network/messaging-v1.md
```diff
@@ -118,4 +118,26 @@ The serialized `NetworkMsg`s over-the-wire then look like a sequence of length-p
 [u32-length-prefix] || [serialized-message-bytes] || ..
 ```
 
+### Maximum Frame Size
```
davidiw, line 121: wow

davidiw (approved):
thanks man!

network/src/testutils/fake_socket.rs
```diff
@@ -75,8 +75,12 @@ impl<'a> AsyncRead for ReadOnlyTestSocket<'a> {
     fn poll_read(
         mut self: Pin<&mut Self>,
         mut _context: &mut Context,
-        buf: &mut [u8],
+        mut buf: &mut [u8],
     ) -> Poll<io::Result<usize>> {
+        if self.fragmented_read {
```
mimoo, line 80: not sure if important but I think we're missing the edge-case where buf is of length 0 :P

davidiw: test code :P

phlip9: Added an explicit `assert!(!buf.is_empty());` to make this clearer. While it's not strictly an error to `read(&[])`, it's kind of undefined and isn't portable. So I've added this assert to help us avoid depending on this undefined behaviour.

PR #6746, opened 2020-11-21
[circleci] use target determinator to only run a subset of unit tests

sunshowers:
Use the target determinator for unit tests. We can probably also use it
for high-level filtering of builds in the future (stacked with sccache),
but this PR is specifically for unit tests.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -465,7 +465,8 @@ jobs:
       - run:
           name: Run all unit tests
           command: |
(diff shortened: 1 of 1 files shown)


sausagee (approved):
It's great to see the integration work out nicely! Saving the compute is also awesome!

rexhoffman (approved):
Do it !!!

:D

glad to see your hard work pay off.

PR #6747, opened 2020-11-21
[x] don't ignore deleted files in changed-since

sunshowers:
Realized that you can do something silly like just delete a file without
affecting any others.

Tested by deleting storage/libradb/src/metrics.rs, then running `cargo xbuild
--changed-since HEAD` and seeing that libradb and a bunch of other packages
were re-tested.


rexhoffman (approved):
Good catch!

PR #6753, opened 2020-11-25
[move][cli] Support safe module republishing with breaking changes ch…

sblackshear:
…ecks

- If the CLI is trying to compile `M.move`, the logic for finding compilation deps now removes `M.move` from the deps list.
More generally, any files directly or transitively reachable from the compilation targets are removed from the deps list. This
fixes the annoying "duplicate definition" warning that you used to get when trying to re-publish a module
- Before a `move publish`, the CLI runs the bytecode linking and data layout compatibility checks and reports an error if the
code to be published could cause a breaking change to either linking or data layout
- Added tests for republishing, running `move check M.move` after publishing `M.move`, and the breaking changes checks

### DIFF
--- a/language/tools/move-cli/src/main.rs
+++ b/language/tools/move-cli/src/main.rs
@@ -24,8 +24,10 @@ use move_vm_runtime::{data_cache::TransactionEffects, logging::NoContextLog, mov
 use move_vm_types::{gas_schedule, values::Value};
 use vm::{
     access::ScriptAccess,
+    compatibility::Compatibility,
     errors::VMError,
     file_format::{CompiledModule, CompiledScript, SignatureToken},
+    normalized::Module,
 };
 
 use anyhow::{bail, Result};
@@ -91,6 +93,11 @@ enum Command {
         /// If set, modules will not override existing ones in storage
         #[structopt(long = "no-republish")]
         no_republish: bool,
+        /// By default, code that might cause breaking changes for bytecode
+        /// linking or data layout compatibility checks will not be published.
+        /// Set this flag to ignore breaking changes checks and publish anyway
+        #[structopt(long = "ignore-breaking-changes")]
+        ignore_breaking_changes: bool,
         /// If set, the effects of executing `script_file` (i.e., published, updated, and
         /// deleted resources) will NOT be committed to disk.
         #[structopt(long = "dry-run", short = "n")]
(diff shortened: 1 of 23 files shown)


sblackshear:
Can't add @mengxu-fb as a reviewer for some reason, but definitely want your opinion!

language/tools/move-cli/src/main.rs
```diff
@@ -163,6 +170,25 @@ impl Move {
...
+    /// Same as `get_compilation_deps`, but do a best-effort filtering to exclude dep files already present in 
+    /// `to_compile`. This prevents "duplicate definition" warnings from the compiler
+    fn get_filtered_deps(&self, to_compile: &[String]) -> Result<Vec<String>> {
+        fn get_stem(s: &String) -> Option<&str> {
+            Path::new(s).file_stem().map(|stem| stem.to_str()).flatten()
+        }
+        let deps = self.get_compilation_deps()?;
+        let to_compile_filenames = move_lang::find_move_filenames(to_compile, true)?;   
+        // convert /path/to/File.move to File.move
+        let file_names: Vec<&str> = to_compile_filenames.iter().map(|s| get_stem(s)).flatten().collect();
```
mengxu-fb, line 182: Maybe using a `Set` here instead of a `Vec`?

language/tools/move-cli/src/main.rs
```diff
@@ -163,6 +170,25 @@ impl Move {
...
+        // exclude any filenames in `deps` that do not match File.move
+        // Note: this is best-effort in at least two ways:
+        // (1) It will wrongly filter modules in the same name under different addresses (e.g., 0x1/M.move and 0x2/M.move)
+        // (2) It will fail to filter modules with the same filename, but different module names (e.g., 0x1/M in both M1.move and M2.move)
```
mengxu-fb, line 186: (3) It may cause modules to be lost? (e.g., 0x1/M and 0x1/N in the deps' `M.move` but only 0x1/M in the updated `M.move` (although I guess this should not be an issue for stdlib modules).

mengxu-fb:
Thanks Sam for pinning me here. The changes LGTM, and I'll leave it to others to stamp it :)

I think given what the CLI has, the file exclusion can only be best-effort and never perfect. I wish we had compiler-side change to support module updates directly, say, in, `move_compile(targets: &[String], deps: &[String], ...)` modules in `targets` supersedes modules found in `deps`.

language/tools/move-cli/src/main.rs
```diff
@@ -91,6 +93,11 @@ enum Command {
         /// If set, modules will not override existing ones in storage
         #[structopt(long = "no-republish")]
         no_republish: bool,
+        /// By default, code that might cause breaking changes for bytecode
+        /// linking or data layout compatibility checks will not be published.
+        /// Set this flag to ignore breakng changes checks and publish anyway
```
tzakian, line 98: nit

```suggestion
        /// Set this flag to ignore breaking changes checks and publish anyway
```

language/tools/move-cli/src/main.rs
```diff
@@ -336,10 +350,33 @@ fn publish(args: &Move, republish: bool, files: &[String]) -> Result<OnDiskState
...
+    if !ignore_breaking_changes {
+        let view = OnDiskStateView::create(storage_dir.to_path_buf(), &[])?;
+        for m in &modules {
+            let id = m.self_id();
+            if let Ok(old_m) = view.get_compiled_module(&m.self_id()) {
+                let old_api = Module::new(&old_m);
+                let new_api = Module::new(m);
+                let compat = Compatibility::check(&old_api, &new_api);
+                if !compat.is_fully_compatible() {
+                    println!("Breaking change detected--publishing aborted. Re-run with --ignore-breaking-changes to publish anyway.")
```
tzakian, line 363: Possibly a larger question for the CLI as well, but I'm thinking this should probably be an `eprintln!` instead since this is a warning/error.

tzakian (approved):
LGTM! Just a couple small things

language/tools/move-cli/src/main.rs
sblackshear: Yeah, good call. I'll do that for now, and in the future I'd like to get rid of all `print`'s and turn them into writes to a buffer. I think we need to do this in order to run CLI tests concurrently in the same process.

PR #6757, opened 2020-11-30
[RFC] Combine WriteSet and Events

sblackshear:
PR implementing part of @msmouse's proposal by eliminating events as a transaction output and placing them inside the WriteSet instead. This is not intended to be committed, just a proof-of-concept to see if any issues arise.


language/libra-vm/src/libra_vm.rs
```diff
@@ -499,24 +500,34 @@ pub fn txn_effects_to_writeset_and_events_cached<C: AccessPathCache>(
...
                 .ok_or_else(|| VMStatus::Error(StatusCode::UNKNOWN_INVARIANT_VIOLATION_ERROR))?;
             let key = EventKey::try_from(guid.as_slice())
                 .map_err(|_| VMStatus::Error(StatusCode::EVENT_KEY_MISMATCH))?;
-            Ok(ContractEvent::new(key, seq_num, ty_tag, msg))
+            Ok(ContractEvent::new(key, ty_tag, msg))
         })
         .collect::<Result<Vec<_>, VMStatus>>()?;
 
-    Ok((ws, events))
+    for event in events {
```
sblackshear, line 516: This is the key piece of code: putting an event inside the WriteSet. There is a big decision point here: where do we put the `TypeTag` that describes the structure of the event?

types/src/contract_event.rs
```diff
@@ -62,8 +52,6 @@ impl Deref for ContractEvent {
 pub struct ContractEventV0 {
     /// The unique key that the event was emitted to
     key: EventKey,
-    /// The number of messages that have been emitted to the path previously
-    sequence_number: u64,
```
sblackshear, line 66: This is the other big change. Move no longer needs to keep track of this; the storage layer will be responsible for assigning it.

davidiw:
Closing out old PR.

PR #6760, opened 2020-11-30
[move CLI] Update README

tzakian:
Updated CLI README. Added additional sections on modes and testing. Cleanup/updating throughout as well. 

Rendered:
https://github.com/libra/libra/blob/ff09eb2353da6b16511ad95849e7a06fdf6f1ece/language/tools/move-cli/README.md

### DIFF
--- a/language/tools/move-cli/README.md
+++ b/language/tools/move-cli/README.md
@@ -1,19 +1,69 @@
 # Move CLI
 
-This is a tool for experimenting with Move without validators, a blockchain, or transactions. Persistent data is stored on-disk in a directory structure that mimics the Move memory model
+This is a tool for experimenting with Move without validators, a blockchain, or
+transactions. Persistent data is stored on-disk in a directory structure that
+mimics the Move memory model
 
 ## Installation
-```
-λ cargo install --path libra/language/tools/move-cli
+```shell
+$ cargo install --path libra/language/tools/move-cli
 ```
 or
+```shell
+$ cargo install --git https://github.com/libra/libra move-cli
 ```
-λ cargo install --git https://github.com/libra/libra move-cli
+
+This will install the `move` binary in your Cargo binary directory. On
+macOS and Linux this is usually `~/.cargo/bin`. You'll want to make sure
+this location is in your `PATH` environment variable.
+
+Now you should be able to run the Move CLI:
+
+```shell
+$ move
+Move 0.1.0
+CLI frontend for Move compiler and VM
+
+USAGE:
+    move [FLAGS] [OPTIONS] <SUBCOMMAND>
+  ...
+```
+
+We'll go through the most common Move CLI commands and flags here, however
+you can find the complete list of commands available by calling `move
+--help`.  Additionally, the complete list of flags and options available
+for each Move CLI command can be found by passing the `--help` flag to it,
+i.e., `move <command> --help`.
+
+
+## Project structure
+
+Each Move CLI project with a given `name` should have the following structure to
+it:
+
+```
+name/
+└── src
+    ├── modules # Directory containing all Move source modules
+    │   ├ ...
+    │   └── Module.move
+    └── scripts # Directory containing all Move scripts
+        ├ ...
+        └── script.move
+```
+
+Let's now create a Move project that we'll use for the code in this README and `cd` into it:
+
+```shell
+$ mkdir readme
+$ cd readme
+$ mkdir -p src/modules
+$ mkdir -p src/scripts
 ```
 
 ## Compiling and running scripts
 
-Here is a simple script that prints its `signer`:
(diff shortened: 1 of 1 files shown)


language/tools/move-cli/README.md
```diff
@@ -24,21 +31,82 @@ fun main(account: &signer) {
...
-The `--signers 0xf` part indicates which account address(es) have signed off on the script. Omitting `--signers` or passing multiple signers to this single-`signer` script will trigger a type error.
-
-## Adding new modules via `src`
-
-By default, the CLI compiles and includes all files from the Move standard library and Libra Framework. New modules can be added to the `src` directory (or a directory of your choosing specified via `--source-dir`. The `move run` command will compile and publish each module source file in this directory before running the given script.
+The `--signers 0xf` argument indicates which account address(es) have signed
+off on the script. Omitting `--signers` or passing multiple signers to this
+single-`signer` script will trigger a type error.
+
+## Modes and adding new modules
```
sblackshear, line 45: Modes seem like a more advanced topic to me, particularly given that we've chosen a good `default` mode that most users won't need to change... Perhaps move this into the "Using CLI with the Libra Framework" section at the end?

language/tools/move-cli/README.md
```diff
@@ -130,33 +194,126 @@ fun main(account: &signer) {
...
-λ move view storage/0x0000000000000000000000000000000f/resources/0x00000000000000000000000000000002\:\:Test\:\:Resource
-resource 00000000::Test::Resource {
-    i: 10
+λ move view storage/0x0000000000000000000000000000000F/resources/0x00000000000000000000000000000002::Test::Resource.lcs
+resource 0x2::Test::Resource {
+        i: 10
 }
 ```
 
+### Cleaning state
```
sblackshear, line 217: Nice! Might also be useful to explain `--dry-run` in this section.

language/tools/move-cli/README.md
```diff
@@ -130,33 +194,126 @@ fun main(account: &signer) {
...
 
-## Using the CLI with Libra modules and genesis state
+Since state persists from one call to the Move CLI to another, there will
+frequently be times where you want to start again at a clean state.  This can
+be done using the `move clean` command. This command will remove all previously
+published modules and resources from the state, and will start you back at a
+fresh state empty state.
```
sblackshear, line 223: Might be easier to just say: this deletes the contents of the `storage` directory (and whatever else it does).

language/tools/move-cli/README.md
```diff
@@ -130,33 +194,126 @@ fun main(account: &signer) {
...
+λ move clean
+λ move view storage/0x0000000000000000000000000000000F/resources/0x00000000000000000000000000000002::Test::Resource.lcs
+Error: `move view <file>` must point to a valid file under storage
+```
+
+## Testing with the Move CLI
+
+The Move CLI also has a built-in testing framework. Each test is run
+independently in its own sandbox so state does not persist from one test to
+another, or from one test run to another.
```
sblackshear, line 235: This second part of the sentence is a bit confusing--someone might think state disappears between args.txt commands. Maybe kill?

sblackshear (approved):
Looking good! Could we put the project structure guide in here somewhere (maybe after installation)? I think that would help folks get the lay of the land.

language/tools/move-cli/README.md
```diff
@@ -130,33 +194,126 @@ fun main(account: &signer) {
...
+
+Each test is structured as a directory consisting of an `args.txt` file that
+specifies a sequence of Move CLI commands that should be run in that directory.
+Usually, there are two subdirectories, one for `modules`, and the other for
```
sblackshear, line 239: The project structure section would let us kill this sentence.

language/tools/move-cli/README.md
```diff
@@ -130,33 +194,126 @@ fun main(account: &signer) {
...
+
+```
+λ move test readme_test
+...<snipped output>
+1 / 0 test(s) passed.
+Error: 1 / 1 test(s) failed.
+```
+
+However, as we see this test will fail since there is no `args.exp` file for the test
+yet. We can generate this expectation file by setting the `UPDATE_BASELINE`
```
sblackshear, line 283:
```suggestion
yet. We can generate this expectation file by setting the `UPDATE_BASELINE` (or UB for short)
```

language/tools/move-cli/README.md
tzakian: Sounds like a good idea to me! I'll move down.

language/tools/move-cli/README.md
tzakian: I'll kill

language/tools/move-cli/README.md
tzakian: I'll add

PR #6762, opened 2020-11-30
[Secure Storage] Remove mutable self identifier from sign_* methods

JoshLind:
## Motivation

This PR removes the mutable self identifier from the sign_* methods exposed by CryptoStorage. It's unclear to me why these need to be mutable, however, my suspicion is that this requirement is probably a legacy one that was cleaned up a while ago (e.g., due to changes we made around Ed25519 private key signing?).

Yes.

## Test Plan

All tests pass locally.

## Related PRs

None.


davidiw (approved):
good clean up

JoshLind:
Thanks, @davidiw! 😄 

/land

PR #6763, opened 2020-11-30
[move source language] Update compiler API

tnowacki:
- Allow for general start/stop of the compiler
- Simplified and reduced number of entry points

## Motivation

- For supporting republishing, we need the ability to parse the files and check for the address/modules that exist in that file so we know which to remove from the dependency list 
- general start/stop and "state machine"-like control allows the compiler to continue after parsing. 

## Test Plan

- Cargo test

### DIFF
--- a/language/move-lang/src/lib.rs
+++ b/language/move-lang/src/lib.rs
@@ -44,9 +44,43 @@ pub const MOVE_COMPILED_EXTENSION: &str = "mv";
 pub const MOVE_COMPILED_INTERFACES_DIR: &str = "mv_interfaces";
 pub const SOURCE_MAP_EXTENSION: &str = "mvsm";
 
+#[macro_export]
+macro_rules! unwrap_or_report_errors {
+    ($files:ident, $res:ident) => {{
+        match $res {
+            Ok(t) => t,
+            Err(errors) => {
+                assert!(!errors.is_empty());
+                $crate::errors::report_errors($files, errors)
+            }
+        }
+    }};
+}
+
 //**************************************************************************************************
 // Entry
 //**************************************************************************************************
+
+#[derive(Clone, Copy, PartialEq, Eq, PartialOrd, Ord)]
+pub enum Pass {
+    Parser,
+    Expansion,
+    Naming,
+    Typing,
+    HLIR,
+    CFGIR,
+    Compilation,
+}
+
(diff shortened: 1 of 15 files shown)


language/tools/move-cli/src/main.rs
```diff
@@ -213,9 +214,9 @@ fn publish(args: &Move, files: &[String]) -> Result<OnDiskStateView> {
     if args.verbose {
         println!("Compiling Move modules...")
     }
-    let (_, compiled_units) =
+    let (files, compiled_units_res) =
         move_lang::move_compile(files, &args.get_compilation_deps()?, None, None)?;
-
+    let compiled_units = move_lang::unwrap_or_report_errors!(files, compiled_units_res);
```
sblackshear, line 219: It seems like `move_compile` followed by `unwrap_or_report_errors` is the most common pattern. Perhaps we could expose a `move_compile_and_report` (or similar) API that does these two steps for you like the old `move_compile` API did?

tnowacki: I was trying to reduce the number of entry points, but maybe you are right...

tnowacki: At the very least I like the more positive phrasing of `_and_report` than the old negative `_no_report`, makes the functions a bit cleaner to define too

language/move-lang/src/lib.rs
```diff
@@ -93,53 +129,52 @@ pub fn move_compile(
...
-        let (eprog, errors) = expansion::translate::program(pprog, sender_opt);
-        check_errors(errors)?;
-        Ok((eprog, comment_map))
-    });
-    Ok((files, res))
+    let result = pprog_and_comments_res.map(|(pprog, comments)| (comments, sender_opt, pprog));
+    Ok((files, result))
+}
+
+pub fn move_continue_up_to(pass: PassResult, until: Pass) -> Result<PassResult, Errors> {
```
sblackshear, line 176: Nit: comment explaining that `until` is exclusive (i.e., the `until` pass will *not* be run)?

sblackshear (approved):
Nice!

language/move-lang/src/lib.rs
tnowacki: Other way around, it is inclusive, but I will add a comment

PR #6765, opened 2020-12-01
[common] Simple token bucket rate limiter

gregnazario:
### Overview
This token bucket rate limiter is made to be generic to be able to
throttle on arbitrary tokens, with a default bucket size and fill rate
for all buckets.  The default interval in which buckets are being
refilled is 1 second.  It also provides the framework to build into the
builder ways for custom rates for specific keys, and eviction policies
of cached keys in the future.  This version is not thread safe, and must
be single threaded.

## Motivation

We needed a more comprehensive rate limiting solution that allows us to adjust based on bytes, requests, or any other value to rate limit on.  This will allow us to apply it to multiple applications and dimensions to rate limit on.  The first application will be for networking requests and then networking data by bytes.

Yes

## Test Plan

I believe the unit tests are pretty comprehensive at this point.  Since it's not made currently for multi threaded, I added some TODOs around some improvement areas, including let's look at how we should make this multithreaded.

## Related PRs

None

### DIFF
--- a/common/rate-limiter/src/lib.rs
+++ b/common/rate-limiter/src/lib.rs
@@ -1,0 +1,318 @@
+// Copyright (c) The Libra Core Contributors
+// SPDX-License-Identifier: Apache-2.0
+
+#![forbid(unsafe_code)]
+
+use libra_logger::prelude::*;
+use std::{cmp::min, collections::HashMap, fmt::Debug, hash::Hash, time::Instant};
+use tokio::time::Duration;
+
+/// Interface for a keyed rate limiter.  This should allow us to
+/// drop in a different keyed rate limiter if required.
+pub trait KeyedRateLimiter<Key: Eq + Hash + Clone + Debug> {
+    /// Determine if a throttle is needed on the key for the given count
+    /// This does not block, or wait and returns `true` if throttled
+    fn throttle(&mut self, key: Key, count: u64) -> bool;
+
+    /// Fills the throttling mechanism for a number of ticks.  Can be manually called
+    fn fill_throttles(&mut self, intervals: u64);
+}
+
+impl<Key: Eq + Hash + Clone + Debug> KeyedRateLimiter<Key> for TokenBucketRateLimiter<Key> {
+    /// Processes a throttle check, for synchronous bookkeeping it also ensures
+    /// all buckets are filled in a timely manner, and buckets exist.
+    /// TODO: Do we want to add a waiting mechanism, this only returns immediately
+    fn throttle(&mut self, key: Key, count: u64) -> bool {
+        // We could be more strict on 0, but there's nothing to throttle then, exit quickly
+        if count == 0 {
+            return false;
+        }
+
+        // Refill buckets if we need to
+        // TODO: Should we do this separately, or continue doing synchronously here
+        self.check_refill(self.last_refresh_time.elapsed());
+
+        // A request larger than the current number of tokens will be throttled
+        // Note: a request that is larger than the bucket size will ALWAYS be throttled
+        let bucket = self.bucket(
+            key.clone(),
+            self.default_bucket_size,
+            self.default_fill_rate,
+        );
+
+        // Only count against limit if we're able to actually process the request
+        if count <= bucket.tokens {
+            bucket.tokens = bucket.tokens.saturating_sub(count);
+            false
+        } else {
+            trace!("Throttling key: {:?}", key);
+            true
+        }
+    }
+
+    /// Fill the buckets with the given rate, saturating at the max size
+    /// This should occur on an interval basis
+    fn fill_throttles(&mut self, intervals: u64) {
+        self.buckets.values_mut().for_each(|bucket| {
+            // Skip if bucket is already filled
+            if bucket.size != bucket.tokens {
+                let new_tokens = intervals.saturating_mul(bucket.rate);
+
+                // Saturate at bucket size
+                bucket.tokens = min(bucket.size, bucket.tokens.saturating_add(new_tokens));
+            }
+        });
+    }
+}
+
+/// A generic token bucket filter
+///
+/// # Terms
+/// ## Key
+/// A `key` is an identifier of the item being rate limited
+///
+/// ## Token
+/// A `token` is the smallest discrete value that we want to rate limit by.  In a situation involving
+/// network requests, this may represent a request or a byte.  `Tokens` are the counters for the
+/// rate limiting, and when there are no `tokens` left in a `bucket`, the `key` is throttled.
+///
+/// ## Bucket
+/// A `bucket` is the tracker of the number of `tokens`.  It has a `bucket size`, and any additional
+/// tokens added to it will "spill" out of the `bucket`.  The `buckets` are filled at an `interval`
+/// with a given `fill rate`.
+///
+/// ## Interval
+/// The `interval` at which we refill *all* of the `buckets` in the token bucket filter. Configured
+/// across the whole token bucket filter.
+///
+/// ## Fill Rate
+/// The rate at which we fill a `bucket` with tokens. Configured per bucket.
+///
+/// ## Bucket Size
+/// Maximum size of a bucket.  A bucket saturates at this size.  Configured per bucket.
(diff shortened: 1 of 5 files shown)


JoshLind:
Noob question: doing a quick search online showed up a few token bucket filter libraries (as well as non-equivalent alternatives, e.g., leaky bucket implementations): (e.g., there's a couple here: https://crates.io/search?q=token%20bucket)

Did you consider any of these before diving into creating our own one?

gregnazario:
> Noob question: doing a quick search online showed up a few token bucket filter libraries (as well as non-equivalent alternatives, e.g., leaky bucket implementations): e.g., there's a couple here: https://crates.io/search?q=token%20bucket. Did you consider any of these before diving into creating our own one? If so, I'm guessing none fit our requirements?

The only one that accomplished what we wanted was [Governor](https://docs.rs/governor/0.3.1/governor/)

This would introduce several new dependencies, I'm not sure what our process on determining what is accepted yet.

* no-std-compat
* quanta
* dashmap
* futures-timer

common/rate-limiter/src/lib.rs
```diff
@@ -0,0 +1,332 @@
...
+            buckets: HashMap::new(),
+            last_refresh_time: Instant::now(),
+            default_bucket_size,
+            default_fill_rate,
+            interval,
+        }
+    }
+
+    /// Adds a bucket with the given configuration, defaultted to filled
+    fn ensure_bucket(&mut self, key: Key, bucket_size: u64, fill_rate: u64) {
```
phlip9, line 161: can probably remove this

common/rate-limiter/src/lib.rs
```diff
@@ -0,0 +1,332 @@
...
+use libra_logger::prelude::*;
+use std::{collections::HashMap, fmt::Debug, hash::Hash, time::Instant};
+use tokio::time::Duration;
+
+/// Interface for a keyed rate limiter.  This should allow us to
+/// drop in a different keyed rate limiter if required.
+pub trait KeyedRateLimiter<Key: Eq + Hash + Clone + Debug> {
+    /// Determine if a throttle is needed on the key for the given count
+    /// This does not block, or wait and returns `true` if throttled
+    fn throttle(&mut self, key: Key, count: u64) -> bool;
```
phlip9, line 15: is `bool` the right return value here? if I'm throttled, presumably I need either some wait time (so I myself can wait until the next time where I might be able to get a permit) or I need to register my task's Waker with the rate limiter so it can wake me up when it's my turn.

common/rate-limiter/src/lib.rs
```diff
@@ -0,0 +1,332 @@
...
+                    .tokens
+                    .saturating_add(intervals.saturating_mul(bucket.rate));
+
+                // Saturate at bucket size
+                bucket.tokens = if new_tokens > bucket.size {
+                    bucket.size
+                } else {
+                    new_tokens
+                }
+            }
```
phlip9, line 76:
```rust
let new_tokens = intervals.saturating_mul(bucket.rate);
bucket.tokens = min(bucket.size, bucket.tokens.saturating_add(new_tokens));
```

common/rate-limiter/src/lib.rs
```diff
@@ -0,0 +1,332 @@
...
+        // Refill buckets if we need to
+        // TODO: Should we do this separately, or continue doing synchronously here
+        self.check_refill(self.last_refresh_time.elapsed());
+
+        // If there's no bucket, let's create a default one
+        self.ensure_bucket(
+            key.clone(),
+            self.default_bucket_size,
+            self.default_fill_rate,
+        );
```
phlip9, line 40: HashMap `.entry` API works well for this : )

```rust
let bucket = self.buckets.entry(&key).or_insert_with(|| Bucket::new(bucket_size, fill_rate));
```

common/rate-limiter/src/lib.rs
```diff
@@ -0,0 +1,332 @@
...
+
+        // A request larger than the current number of tokens will be throttled
+        // Note: a request that is larger than the bucket size will ALWAYS be throttled
+        if let Some(bucket) = self.buckets.get_mut(&key) {
```
phlip9, line 44: can remove this get_mut and else branch

common/rate-limiter/src/lib.rs
gregnazario: I'm leaving it up to the requesting service to handle it in this first iteration, there's a TODO in there for something like this.

Here's our problem, our downstream services don't retry, but since we currently don't apply backpressure we'd have to just wait until it's my turn.  But, what happens when you've got a ton of requests coming in waiting, how do we load shed so we don't just keep accepting requests that then get immediately thrown into waiting.

I have a feeling that different situations would prescribe different solutions, including how do you then provide fairness on the waking based on order, size, and key.  So, for now I *want* to just return that it's been throttled, and initially drop the request.

common/rate-limiter/src/lib.rs
gregnazario: I separated this out for legibility, but yes it could be removed.

common/rate-limiter/src/lib.rs
gregnazario: Nice, my rust was rusty from having time off.

common/rate-limiter/src/lib.rs
phlip9: Gotcha, maybe we return (like governor), `Result<(), Duration>` to signal "hey you're throttled and here's a hint on when you can probably next get a permit". Then callers can decide to either wait (if they're applying backpressure) or load shed o/w.

common/rate-limiter/src/lib.rs
gregnazario: Ah, I remember why.

So, in the event we add the builder to make custom throttles for specific keys, we could use this function accordingly to add buckets

common/rate-limiter/src/lib.rs
gregnazario: Yeah, that's a little different, where the rate limiter handles the jitter and the delay in which you'll be sending your next request, which is interesting.  I'll take a look into it and see if it'll give us some results.

I highly expect this piece to evolve over time :)

phlip9 (approved):
LGTM! It's a bit hard to review the interface in isolation w/o knowing how we're going to use it, so let's just land this now so we can iterate on it : )

common/rate-limiter/src/lib.rs
```diff
@@ -0,0 +1,318 @@
...
+    fn test_default_fill() {
+        let key_a = "Hello";
+        let default_bucket_size = 1;
+        let default_fill_rate = 1;
+        // Make interval long enough that it won't refill manually in this test
+        let builder = RateLimiterBuilder::new(default_bucket_size, default_fill_rate);
+
+        let mut rate_limiter = builder.build();
+        assert_num_allowed(&mut rate_limiter, key_a, 1);
+        sleep(Duration::from_secs(1));
```
JoshLind, line 313: nit: one way to avoid having to sleep in tests is to pass in some sort of time service, and then in tests use a mock. If I remember correctly, we did this for the key manager: https://github.com/libra/libra/blob/1b32f757673bd16aff5cfd84901573b4eb1766f5/secure/key-manager/src/lib.rs#L105

JoshLind: e.g., in the key manager tests we used, we applied a MockTimeService (https://github.com/libra/libra/blob/342116c954c6645fdaae192c4a448d7a46a08581/secure/key-manager/src/tests.rs#L363)

gregnazario: I agree, should use a mock time service instead of actual time.  I'll fix it up so we I don't waste 1 second * the number of runs of developer time.

gregnazario:
> LGTM! It's a bit hard to review the interface in isolation w/o knowing how we're going to use it, so let's just land this now so we can iterate on it : )

Already iterating on it!  I agree it's hard to review the interface, but I wanted to some early insight

gregnazario:
I'm going to close this and try to use Governor instead

PR #6766, opened 2020-12-01
[breaking] [vm] Remove hash in access path

runtian-zhou:
## Motivation

Remove hash in AccessPath. This will be a huge breaking change as it will change the key for each resources entirely. However, we do believe that this is the right step to take with following reasons:

1. This will greatly help decoding the resources that gets stored in the storage: 
- We use a global static analysis that overapproximates the list of types that may be packed by the standard library, then uses the results to generate a map from each of these types to its hash. We must build this map each time the standard library is updated.
- We ask clients to store this map locally and keep it up to date over time as we add/remove types. Without this map, a client will not know how to interpret account blobs or any other state data stored on-chain.
- When Libra devs debug by looking at the global state, they cannot tell what resources are in a particular account without consulting this map. If the state is historical, the Libra dev may also need to look up a historical version of the map (which we do not provide any way to retrieve or compute). Clients have the same issue with viewing their own historical data.

Removing the hash will solve this problem fundamentally: the layout of a resource will be encoded in the ke.

2. This wil impact the storage by having a slightly bigger storage key size: changing from a 32byte hash to a serialized type name will cause the average key size to grow by 2x. Some access path might be smaller though thanks to our short 16byte address schema. According to @msmouse calculation, it will increase the raw disk consumption of a transaction by 3%. Given that the key section could potentially be compressed efficiently in the future, it is okay to make such change.

3. This will benefit execution slightly: According to my previous benchmark runs, the execution will be faster as we don't need to calculate the hash of Move types any more. The commit time might be slightly longer with bigger key size but its impact is neglegible. Here's the benchmark numbers that I got:
With hash:
execution time: 318.0410959ms, commit time: 89.66438356ms, avg tps: 1236.965753
Without hash:
execution time: 310.5472973ms, commit time: 88.86486486ms , avg tps: 1261.371622

4. This will complicate the gas logic for access path slightly, as longer type name will cause bigger disk consumption. However we could solve this by providing a maximum length of path and charge a constant fee. cc @tzakian.

Given that the potential long term benefit that exposing the type could provide, this seems like a desirable breaking change to make.

The PR is organized in two commits: the first commit removes the hashing part in the AccessPath and fixed the test cases, the second commit did a bit of code refactor so that it would be a lot easier for us to decode the AccessPath in the future.

Yes

## Test Plan

Fixed all existing tests. Ran cluster test & execution-benchmarks with performance number attached.


sblackshear (approved):
LGTM. Nice work, this is actually a much lighter code change than expected!

runtian-zhou:
Had a random error in cluster test unfortunately.

jolestar:
Is there a security problem if someone maliciously constructs a very long key through a nested Generic type structure?

runtian-zhou:
> Is there a security problem if someone maliciously constructs a very long key through a nested Generic type structure?

That is actually a ligit concern. Our repsonse is following:
1. You pay for the data. We will be charged more if you are planning to put a longer name in storage.
2. It is another spot that the system will have to deal with perf implications. For the VM this could mean string interning or other classic string perf solutions

On the other hand, we strongly suggest that the drawbacks over there worth it for the benefits of not having the hash as this will make our storage totally self-describable.

PR #6767, opened 2020-12-01
[state-sync] version counters updates

davidiw:
* Cleaned up the version counters
* Moved target version counter to a saner and earlier location
* Made target version more sensible such that it is always updated otherwise we end up with charts where target < synced | committed
* Added a highest version because this actually gives us an indication of how far behind the node is where as target is more indicative of when the next commit will occur

### DIFF
--- a/state-synchronizer/src/coordinator.rs
+++ b/state-synchronizer/src/coordinator.rs
@@ -144,6 +144,10 @@ impl PendingLedgerInfos {
     fn target_li(&self) -> Option<LedgerInfoWithSignatures> {
         self.target_li.clone()
     }
+
+    fn highest_version(&self) -> Option<Version> {
+        self.pending_li_queue.keys().last().cloned()
+    }
 }
 
 /// Coordination of synchronization process is driven by SyncCoordinator, which `start()` function
@@ -1217,6 +1221,15 @@ impl<T: ExecutorProxyTrait> SyncCoordinator<T> {
             }
         };
 
+        let target_version = target
+            .version()
+            .unwrap_or_else(|| known_version.wrapping_add(1));
+        counters::set_version(counters::VersionType::Target, target_version);
+        let highest_version = self
+            .pending_ledger_infos
+            .highest_version()
+            .unwrap_or(target_version);
+        counters::set_version(counters::VersionType::Highest, highest_version);
         let req = GetChunkRequest::new(known_version, known_epoch, self.config.chunk_limit, target);
         self.request_manager.send_chunk_request(req)
     }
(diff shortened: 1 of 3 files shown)


state-synchronizer/src/coordinator.rs
```diff
@@ -1221,6 +1221,15 @@ impl<T: ExecutorProxyTrait> SyncCoordinator<T> {
...
         };
 
+        let target_version = target
+            .version()
+            .unwrap_or_else(|| known_version.wrapping_add(1));
+        counters::set_version(counters::VersionType::Target, target_version);
+        let highest_version = self
+            .pending_ledger_infos
+            .highest_version()
+            .unwrap_or(target_version);
```
JoshLind, line 1231: Nit: I don't have a lot of scope into how these metrics are used, but would we rather not just ".unwrap_or(known_version)"? e.g., if pending ledger infos is empty and there is no target, highest_version == target_version, so we lose some information in the metrics about the latest known_version? Unless we also log known_version somewhere? (I'm just wondering if we'll be able to discern all the cases using metrics -- but feel free to disregard my comment! 😄)

state-synchronizer/src/counters.rs
```diff
@@ -19,9 +19,33 @@ pub const CHUNK_REQUEST_MSG_LABEL: &str = "chunk_request";
...
+    VERSION
+        .with_label_values(&[version_type.as_str()])
+        .set(version as i64)
+}
+
+pub enum VersionType {
+    /// Version of latest ledger info committed.
+    Committed,
+    /// Highest known version or version proceeding it
+    Highest,
```
JoshLind, line 32: Nit: Perhaps rename "Highest" to "HighestKnown" to make it explicit? "Highest" by itself seems a little ambiguous to me.

JoshLind (approved):
Nice clean ups! 😃 Thanks, @davidiw!

state-synchronizer/src/coordinator.rs
davidiw: committed <= synced <= target <= highest

and then in the code as it is today

committed <= synced < target <= highest

because of the way the counters are hit I don't think nuance would be worth it whereas folks would always expect target to be such that synced < target <= highest otherwise what are you going after?

JoshLind: Okay, this makes sense to me. I wasn't looking for anything specific -- just wanted to call out that if there is nuance in the metrics around edge cases, we should pay attention to them. But it sounds like there isn't anything interesting 😄

PR #6774, opened 2020-12-01
[state-sync] store timestamps for synced, committed, and real to prometheus

davidiw:
The intent here is that we would like to observe how far behind our fullnode (or any node) is behind real time which should give us an indication of how far behind it is in syncing with the chain. Synced gives us insights about how far we are within an epoch and which committed would give us.

This hasn't been tested yet -- poking @aaron and @zekun000 and @msmouse for some early feedback.

If not, I'll just try it by syncing to premainnet or another network that has lots of epochs so I can see some interesting data :D

### DIFF
--- a/state-synchronizer/src/coordinator.rs
+++ b/state-synchronizer/src/coordinator.rs
@@ -461,6 +461,19 @@ impl<T: ExecutorProxyTrait> SyncCoordinator<T> {
         let local_epoch = self.local_state.epoch();
         counters::set_version(counters::VersionType::Synced, synced_version);
         counters::set_version(counters::VersionType::Committed, committed_version);
+        counters::set_timestamp(
+            counters::TimestampType::Synced,
+            self.executor_proxy.get_version_timestamp(synced_version)?,
+        );
+        counters::set_timestamp(
+            counters::TimestampType::Committed,
+            self.executor_proxy
+                .get_version_timestamp(committed_version)?,
+        );
+        counters::set_timestamp(
+            counters::TimestampType::Real,
+            libra_infallible::duration_since_epoch().as_micros() as u64,
+        );
         counters::EPOCH.set(local_epoch as i64);
         debug!(LogSchema::new(LogEntry::LocalState)
             .local_li_version(committed_version)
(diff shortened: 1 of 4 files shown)


davidiw:
Ignore the first 3 commits, they're approved, I just need to do some tweaks and land them at which point I'll update this PR.

lightmark:
it should work. lgtm

state-synchronizer/src/coordinator.rs
```diff
@@ -461,6 +461,20 @@ impl<T: ExecutorProxyTrait> SyncCoordinator<T> {
         let local_epoch = self.local_state.epoch();
         counters::set_version(counters::VersionType::Synced, synced_version);
         counters::set_version(counters::VersionType::Committed, committed_version);
+        counters::set_timestamp(
+            counters::TimestampType::Synced,
+            self.executor_proxy
+                .get_vm_timestamp_for_version(synced_version)?,
```
msmouse, line 467: This calls into storage to get the timestamp for a version? I guess you probably don't want to do it (especially multiple times in the critical path)?

state-synchronizer/src/executor_proxy.rs
```diff
@@ -195,6 +198,25 @@ impl ExecutorProxyTrait for ExecutorProxy {
         self.storage.get_epoch_ending_ledger_info(version)
     }
 
+    /// Returns the ledger's timestamp for the given version in microseconds
+    fn get_vm_timestamp_for_version(&self, version: u64) -> Result<u64> {
+        self.storage
+            .get_account_state_with_proof_by_version(
```
msmouse, line 204: Probably more efficient and reliable to do `storage.get_block_timestamp(version)`

davidiw: oh god... nice

state-synchronizer/src/coordinator.rs
davidiw: I don't this it is multiple times per chunk. Thoughts on alternative approaches?

msmouse: Service thread for logging? 😝😝😝 But probably log it even more less frequently. For example, log commited timestamp only when a ledger_info is written, and when it's the case, the synced timestamp should be the same?

msmouse: K. I was exaggerating. Probably fine doing it twice for each chunk.

davidiw (commented):
crap I need to remove some nonsense that got pulled in here under the docker compose...

zekun000 (approved):

PR #6776, opened 2020-12-01
add new site redirects

oriooctopus:
redirect users from v1 docs site to v2


ericnakagawa (approved):
These work thanks Olliver!

PR #6778, opened 2020-12-02
[Logger] Add backtraces to error logs

JoshLind:
## Motivation

This PR updates the metadata structure for logs to take backtraces when error! logs are created. This is useful for when you need to track down or identify the underlying cause of an error quickly (e.g., if you're an oncall responding to an error alarm in code you are not familiar with). 

### Why can't we just use the other logs around the error to identify/construct the backtrace?
- Ideally, this would be true. However, there is no guarantee that surrounding logs are informative enough to tell us what's happened (I've personally encountered this during my oncall rotation).
- Moreover, it's not completely true that our logs are strictly ordered according to execution ordering in the program. As far as I understand, our logs are ordered by timestamp, however, our timestamps only have a limited amount of precision. Thus, it's possible that multiple logs have the same timestamp, and there is no guarantee about the execution order of logs that share timestamps.

Note: this PR chooses not to take backtraces for all logs, but only for errors. This is because: (i) backtraces can be long, and we don't want to spam our logging pipelines; and (ii) backtraces do have some cost to construct (e.g., symbol resolution).

Yes.

## Test Plan

The existing logger tests have been updated to verify that a backtrace is taken for error logs. Moreover, running the test locally and forcing it to display the metadata contents of the error log shows the backtrace stack clearly, e.g.:

```
 LogEntry { metadata: Metadata { level: Error, target: "libra_logger", module_path: "libra_logger::libra_logger::tests", file: "common/logger/src/libra_logger.rs", line: 575, location: "common/logger/src/libra_logger.rs:575", backtrace: Some(   0: backtrace::backtrace::libunwind::trace
             at /Users/joshlind/.cargo/registry/src/github.com-1ecc6299db9ec823/backtrace-0.3.55/src/backtrace/libunwind.rs:90:5
      backtrace::backtrace::trace_unsynchronized
             at /Users/joshlind/.cargo/registry/src/github.com-1ecc6299db9ec823/backtrace-0.3.55/src/backtrace/mod.rs:66:5
   1: backtrace::backtrace::trace
             at /Users/joshlind/.cargo/registry/src/github.com-1ecc6299db9ec823/backtrace-0.3.55/src/backtrace/mod.rs:53:14
   2: backtrace::capture::Backtrace::create
             at /Users/joshlind/.cargo/registry/src/github.com-1ecc6299db9ec823/backtrace-0.3.55/src/capture.rs:176:9
   3: backtrace::capture::Backtrace::new
             at /Users/joshlind/.cargo/registry/src/github.com-1ecc6299db9ec823/backtrace-0.3.55/src/capture.rs:140:22
   4: libra_logger::metadata::Metadata::new
             at src/metadata.rs:43:34
   5: libra_logger::libra_logger::tests::basic
             at src/libra_logger.rs:575:9
   6: libra_logger::libra_logger::tests::basic::{{closure}}
             at src/libra_logger.rs:519:5
   7: core::ops::function::FnOnce::call_once
             at /Users/joshlind/.rustup/toolchains/1.47.0-x86_64-apple-darwin/lib/rustlib/src/rust/library/core/src/ops/function.rs:227:5
   8: core::ops::function::FnOnce::call_once
             at /rustc/18bf6b4f01a6feaf7259ba7cdae58031af1b7b39/library/core/src/ops/function.rs:227:5
      test::__rust_begin_short_backtrace
             at /rustc/18bf6b4f01a6feaf7259ba7cdae58031af1b7b39/library/test/src/lib.rs:517:5
   9: <alloc::boxed::Box<F> as core::ops::function::FnOnce<A>>::call_once
             at /rustc/18bf6b4f01a6feaf7259ba7cdae58031af1b7b39/library/alloc/src/boxed.rs:1042:9
      <std::panic::AssertUnwindSafe<F> as core::ops::function::FnOnce<()>>::call_once
             at /rustc/18bf6b4f01a6feaf7259ba7cdae58031af1b7b39/library/std/src/panic.rs:308:9
      std::panicking::try::do_call
             at /rustc/18bf6b4f01a6feaf7259ba7cdae58031af1b7b39/library/std/src/panicking.rs:373:40
      std::panicking::try
             at /rustc/18bf6b4f01a6feaf7259ba7cdae58031af1b7b39/library/std/src/panicking.rs:337:19
      std::panic::catch_unwind
             at /rustc/18bf6b4f01a6feaf7259ba7cdae58031af1b7b39/library/std/src/panic.rs:379:14
      test::run_test_in_process
             at /rustc/18bf6b4f01a6feaf7259ba7cdae58031af1b7b39/library/test/src/lib.rs:544:18
      test::run_test::run_test_inner::{{closure}}
             at /rustc/18bf6b4f01a6feaf7259ba7cdae58031af1b7b39/library/test/src/lib.rs:450:39
  10: std::sys_common::backtrace::__rust_begin_short_backtrace
             at /rustc/18bf6b4f01a6feaf7259ba7cdae58031af1b7b39/library/std/src/sys_common/backtrace.rs:137:18
  11: std::thread::Builder::spawn_unchecked::{{closure}}::{{closure}}
             at /rustc/18bf6b4f01a6feaf7259ba7cdae58031af1b7b39/library/std/src/thread/mod.rs:458:17
      <std::panic::AssertUnwindSafe<F> as core::ops::function::FnOnce<()>>::call_once
             at /rustc/18bf6b4f01a6feaf7259ba7cdae58031af1b7b39/library/std/src/panic.rs:308:9
      std::panicking::try::do_call
             at /rustc/18bf6b4f01a6feaf7259ba7cdae58031af1b7b39/library/std/src/panicking.rs:373:40
      std::panicking::try
             at /rustc/18bf6b4f01a6feaf7259ba7cdae58031af1b7b39/library/std/src/panicking.rs:337:19
      std::panic::catch_unwind
             at /rustc/18bf6b4f01a6feaf7259ba7cdae58031af1b7b39/library/std/src/panic.rs:379:14
      std::thread::Builder::spawn_unchecked::{{closure}}
             at /rustc/18bf6b4f01a6feaf7259ba7cdae58031af1b7b39/library/std/src/thread/mod.rs:457:30
      core::ops::function::FnOnce::call_once{{vtable.shim}}
             at /rustc/18bf6b4f01a6feaf7259ba7cdae58031af1b7b39/library/core/src/ops/function.rs:227:5
  12: <alloc::boxed::Box<F> as core::ops::function::FnOnce<A>>::call_once
             at /rustc/18bf6b4f01a6feaf7259ba7cdae58031af1b7b39/library/alloc/src/boxed.rs:1042:9
      <alloc::boxed::Box<F> as core::ops::function::FnOnce<A>>::call_once
             at /rustc/18bf6b4f01a6feaf7259ba7cdae58031af1b7b39/library/alloc/src/boxed.rs:1042:9
      std::sys::unix::thread::Thread::new::thread_start
             at /rustc/18bf6b4f01a6feaf7259ba7cdae58031af1b7b39/library/std/src/sys/unix/thread.rs:87:17
  13: __pthread_start
) 
```

The only open questions I have are how this will render in Kibana (i.e., with or without newline support), and whether or not we should consider trimming off a few of the backtrace frames to remove the various Backtrace::new() and Backtrace::create() calls. Either way, having this information for errors will be super useful!

## Related PRs

None.

### DIFF
--- a/common/logger/src/metadata.rs
+++ b/common/logger/src/metadata.rs
@@ -1,10 +1,11 @@
 // Copyright (c) The Libra Core Contributors
 // SPDX-License-Identifier: Apache-2.0
 
+use backtrace::Backtrace;
 use serde::{Deserialize, Serialize};
 use std::{fmt, str::FromStr};
 
-#[derive(Clone, Copy, Debug, Serialize, Deserialize)]
+#[derive(Clone, Debug, Serialize, Deserialize)]
 pub struct Metadata {
     /// The level of verbosity of the event
     level: Level,
@@ -23,10 +24,14 @@ pub struct Metadata {
 
     /// The file name and line number together 'file:line'
     location: &'static str,
+
+    /// The program backtrace taken when the event occurred. Backtraces are
+    /// only supported for errors.
+    backtrace: Option<Backtrace>,
 }
 
 impl Metadata {
-    pub const fn new(
+    pub fn new(
         level: Level,
         target: &'static str,
         module_path: &'static str,
@@ -34,6 +39,11 @@ impl Metadata {
         line: u32,
         location: &'static str,
     ) -> Self {
+        let backtrace = match level {
+            Level::Error => Some(Backtrace::new()),
+            _ => None,
+        };
+
         Self {
             level,
             target,
@@ -41,6 +51,7 @@ impl Metadata {
             file,
             line,
             location,
+            backtrace,
         }
     }
 
@@ -70,6 +81,10 @@ impl Metadata {
 
     pub fn location(&self) -> &'static str {
         self.location
+    }
+
+    pub fn backtrace(&self) -> Option<Backtrace> {
+        self.backtrace.clone()
     }
 }
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -231,6 +231,7 @@ dependencies = [
  "miniz_oxide",
  "object",
  "rustc-demangle",
+ "serde",
 ]
 
 [[package]]
@@ -3188,6 +3189,7 @@ dependencies = [
 name = "libra-logger"
 version = "0.1.0"
 dependencies = [
+ "backtrace",
  "chrono",
  "erased-serde",
  "hostname",
--- a/common/logger/Cargo.toml
+++ b/common/logger/Cargo.toml
@@ -12,6 +12,7 @@ edition = "2018"
 # Do NOT add any inter-project dependencies.
 # This is to avoid ever having a circular dependency with the libra-logger crate.
 [dependencies]
+backtrace = { version = "0.3", features = ["serde"] }
 chrono = "0.4.19"
 erased-serde = "0.3.12"
 hostname = "0.3.1"
--- a/common/logger/src/libra_logger.rs
+++ b/common/logger/src/libra_logger.rs
@@ -68,7 +68,7 @@ impl LogEntry {
             }
         }
 
-        let metadata = *event.metadata();
+        let metadata = event.metadata().clone();
         let thread_name = thread_name.map(ToOwned::to_owned);
         let message = event.message().map(fmt::format);
 
@@ -544,6 +544,7 @@ mod tests {
         );
         assert_eq!(entry.metadata.module_path(), module_path!());
         assert_eq!(entry.metadata.file(), file!());
+        assert!(entry.metadata.backtrace().is_none());
         assert_eq!(entry.message.as_deref(), Some("This is a log"));
 
         // Log time should be the time the structured log entry was created
@@ -569,6 +570,11 @@ mod tests {
             entry.data.get("category").and_then(JsonValue::as_str),
             Some("name"),
         );
+
+        // Test error logs contain backtraces
+        error!("This is an error log");
+        let entry = receiver.recv().unwrap();
+        assert!(entry.metadata.backtrace().is_some());
 
         // Test all log levels work properly
         // Tracing should be skipped because the Logger was setup to skip Tracing events


common/logger/src/metadata.rs
```diff
@@ -23,24 +24,34 @@ pub struct Metadata {
 
     /// The file name and line number together 'file:line'
     location: &'static str,
+
+    /// The program backtrace taken when the event occurred. Backtraces are
+    /// only supported for errors.
+    backtrace: Option<Backtrace>,
```
gregnazario, line 30: Luckily this makes it at the end of the log lines!  We might want to consider trimming it in some way.  Usually only the first few lines in a stack trace are useful.

common/logger/src/metadata.rs
```diff
@@ -23,24 +24,34 @@ pub struct Metadata {
...
         target: &'static str,
         module_path: &'static str,
         file: &'static str,
         line: u32,
         location: &'static str,
     ) -> Self {
+        let backtrace = match level {
+            Level::Error => Some(Backtrace::new()),
+            _ => None,
+        };
```
gregnazario, line 45: So, this doesn't give the backtrace of the error though, just the backtrace of the error log line.  Is this what we want?  Sometimes the error is passed several places before it's logged.

Additionally, sometimes it can be helpful for errors that were passed into warn.

gregnazario (commented):
It looks good, except I think we need to think about ways to get stack traces attached to the errors rather than the logs.

bmwill:
I agree with Greg in that this probably isn't what we want. This provides backtraces at the log line instead of where the error occurred which is what we actually want.

common/logger/src/metadata.rs
JoshLind: Yeah, I noted this in the PR description. If we decide to go with this approach, I can trim off the first few stack frames to make the backtrace clearer 😄

common/logger/src/metadata.rs
JoshLind: Yeah, exactly -- this takes the backtrace at the line of the "error!" log. See my response below for whether or not this makes sense right now

JoshLind:
Thanks @gregnazario, @bmwill 😄!

>It looks good, except I think we need to think about ways to get stack traces attached to the errors rather than the logs.

I agree. In an ideal world, we'd be able to do this with ease, and pass those bracktraces into the error log and thus have them in Kibana for when we need to debug things. The reason I didn't adopt take approach in this PR is because (as far as I can tell) it's not a trivial PR (see the note* below). 

My thinking here was to provide a starting point for error backtraces, that we can then later iterate on/improve. The reason I thought this was useful is because in my previous oncall shift I had to track down an underlying issue (effectively try to figure out the backtrace of an error log/line in the code), and the process was a lot more painful than it needed to be (it took me about 1.5 hours because the logs were not strictly ordered, and it was code I was unfamiliar with).

*Note: From previous discussions with @bmwill, to get backtraces (automatically) in code, you need stable backtrace support for errors (e.g., https://docs.rs/stable-eyre/0.2.1/stable_eyre/ will do the job). However, the problem comes with adopting the approach required by these error libraries:
- First, for each error, you need to adopt the library approach for grabbing the backtrace (e.g., use "eyre!" in stable). This needs to be adopted everywhere we want a backtrace (or at least we need to come up with a magic macro for doing this and apply it everywhere we want).
- Second, (and probably more difficult to deal with) this causes problems when errors are passed between components (e.g., from secure storage to the key manager), because for most error passing our code does does stuff like: `format!("Error: {}", error.to_string())`, which typically produces a string dump of the error message and backtrace, and then wraps it in another error type before continuing. If this error wrapper then also grabs a backtrace, things become super messy! 

Taking the backtrace where the error is logged, gets around these issues, but I agree, it's not optimal. Overall I think providing good backtrace support for errors is something we should do/think about, but it's not trivial and I figured some backtrace support for now is better than none :P Thoughts?

JoshLind:
Also: if it provides any help/context, I created a previous PR that attempted to get backtraces for all errors in the key manager (https://github.com/libra/libra/pull/5859). I ended up closing it because it requires a lot of code modifications and given the risk/reward, I felt it wasn't worth it at the time.

common/logger/src/metadata.rs
davidiw: while I generally agree I would avoid pre-optimizations, I don't think our error logs are that frequent and we could also look at sampling rather than trying to guess what portion of a log might be valuable.

davidiw:
To echo what others have said, we can likely figure out the line of code where the logging occurred by looking at the log itself and searching the code base. It is a pretty trivial process -- albeit maybe not for folks that don't have the code base locally -- but hey can they really help debug / fix issues if they don't?

thanks for looking into this!

JoshLind:
> To echo what others have said, we can likely figure out the line of code where the logging occurred by looking at the log itself and searching the code base. It is a pretty trivial process -- albeit maybe not for folks that don't have the code base locally -- but hey can they really help debug / fix issues if they don't?

Sure, although my reason for looking into this was *not* about finding the specific log line (which is pretty trivial using code search and it's actually already included in our logging framework), but about finding out *how the program got there* (i.e., the stack trace). I had to do this for an invariant violation in the VM (to see how we got to the place that we did) and it took me a lot longer than I would have liked. 

My general experience when debugging errors and crashes is that backtraces are a lot more useful than one might think. But given the general consensus/negative feeling around this PR, I'm reluctant to think that maybe it's just me who likes having this information.. 😅

bmwill (approved):
I think that ensuring we have sufficient insight into what happens during an error before launch is very important. While I think that would be more accurate if we were able to attached the backtraces to the errors themselves I think this would be an acceptable replacement for that (until backtraces in the Error trait is stabalized)

JoshLind:
Closing this our in favour of the replacement (that has been successfully rebased after the libra -> diem name change): https://github.com/diem/diem/pull/6895

PR #6779, opened 2020-12-02
[vm] refactor validation code to allow calling it from the execution phase

bob-wilson:

In the process of making this change, I discovered that we are charging gas for the number of accounts that are mutated, but the VM's num_mutated_accounts function is incorrectly including accounts that are only read from in that count. We should fix that in the future, but to preserve compatibility with the current behavior, I had to pass around a VM session so that the gas cost from running the prologue (due to the bug) could be included with the gas cost from executing the transaction script.

## Motivation

Fixing a minor inconsistency and avoiding future problems

Yes

## Test Plan

I added tests to ensure the existing inconsistencies are fixed.

### DIFF
--- a/language/e2e-testsuite/src/tests/write_set.rs
+++ b/language/e2e-testsuite/src/tests/write_set.rs
@@ -266,8 +266,9 @@ fn bad_writesets() {
         .sequence_number(1)
         .gas_currency_code("Bad_ID")
         .sign();
-    assert_eq!(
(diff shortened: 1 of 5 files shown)


bob-wilson:
I would much prefer to land the breaking version of this change (#6780) if we have an opportunity to do that.

bob-wilson:
We're going with the breaking version of this, so I'll close this now.

PR #6780, opened 2020-12-02
[breaking][vm] refactor validation code to allow calling it from the execution phase

bob-wilson:

In the process of making this change, I discovered that we are charging gas for the number of accounts that are mutated, but the VM's num_mutated_accounts function is incorrectly including accounts that are only read from in that count. Because of that bug, this is a breaking change. The prologue functions are now run in a separate VM session, so with this change, no gas is charged for the prologue functions (as was originally intended). We should still fix that bug because it continues to affect the gas charges for executing the transaction scripts.

## Motivation

Fixing a minor inconsistency and avoiding future problems

Yes

## Test Plan

I added tests to ensure the existing inconsistencies are fixed.

## Related PRs

#6779 is a non-breaking version of this, but it makes the code more complicated and constrains us for the future.


bob-wilson:
@runtian-zhou offered to manage this PR for me while I'm out of the office. We're hoping that there will be an opportunity to take this breaking change due to other changes that will be needed regardless.

language/libra-vm/src/libra_transaction_validator.rs
```diff
@@ -113,73 +70,38 @@ impl VMValidator for LibraVMValidator {
...
         let txn_sender = transaction.sender();
-        let signature_verified_txn = if let Ok(t) = transaction.check_signature() {
+
+        let txn = if let Ok(t) = transaction.check_signature() {
             t
         } else {
             return VMValidatorResult::new(
                 Some(StatusCode::INVALID_SIGNATURE),
-                gas_price,
+                0,
```
runtian-zhou, line 81: we will have a mempool behavior change here, but I think this one makes sense

language/move-vm/types/src/gas_schedule.rs
```diff
@@ -74,10 +74,9 @@ impl<'a> CostStrategy<'a> {
 
     /// Charge a given amount of gas and fail if not enough gas units are left.
     pub fn deduct_gas(&mut self, amount: GasUnits<GasCarrier>) -> PartialVMResult<()> {
-        if !self.charge {
+        if !self.charge || amount.get() == 0 {
```
runtian-zhou, line 77: Why do we need this extra check here?

runtian-zhou: Ok I'm actually puzzled by this line of code change. Tests still pass even without this line. I guess I'll need some insights from @bob-wilson once he's back :(

language/libra-vm/src/libra_transaction_validator.rs
bob-wilson: Does it actually change something? I was assuming that an invalid transaction will be discarded so that the priority for mempool will be irrelevant.

language/move-vm/types/src/gas_schedule.rs
bob-wilson: The essential change here was to remove the subsequent assertion that the amount is non-zero. Adding the check for an early-exit here seemed most consistent with that assertion, but I don't think it's strictly required. It does save a tiny bit of unnecessary computation, though.

bob-wilson:
This is being replaced by #6891

PR #6786, opened 2020-12-02
[move-lang] allow _x to bind but do not warn as unused local

mengxu-fb:
## Motivation

`_` patterns are used to not introduce a binding, e.g. `let (x, _, z) = (0, 1, 2);`
- In Rust, if you do `_y` instead of `_`, e.g. `let (x, _y, z) = (0, 1, 2);` This introduces a binding `_y = 1` but it will not complain if you don't use it
- In Move currently `_y` is just sugar for `_` which isn't really helpful

The fix:
- make `_y` not remap to `_`, this will automatically fix being able to use it
- add an exclusion for underscore prefixed locals in the unused local warning

## Test Plan

`cargo xtest`

### DIFF
--- a/language/move-lang/src/cfgir/liveness/mod.rs
+++ b/language/move-lang/src/cfgir/liveness/mod.rs
@@ -295,16 +295,11 @@ mod last_usage {
                         DisplayVar::Tmp => (),
                         DisplayVar::Orig(v_str) => {
                             if !v.starts_with_underscore() {
-                                let verb = if context.is_resourceful(&v) {
-                                    // do not suggest prefixing for resource bindings
-                                    "replacing"
-                                } else {
-                                    "replacing / prefixing"
-                                };
                                 let msg = format!(
                                     "Unused assignment or binding for local '{}'. \
-                                    Consider removing or {} it with '_'",
-                                    v_str, verb
+                                    Consider removing, replacing with '_', \
+                                    or prefixing with '_' (e.g., '_{}')",
+                                    v_str, v_str
                                 );
                                 context.error(vec![(l.loc, msg)]);
                             }
--- a/language/move-lang/src/cfgir/liveness/mod.rs
+++ b/language/move-lang/src/cfgir/liveness/mod.rs
@@ -294,12 +294,14 @@ mod last_usage {
                     match display_var(v.value()) {
                         DisplayVar::Tmp => (),
                         DisplayVar::Orig(v_str) => {
-                            let msg = format!(
-                                "Unused assignment or binding for local '{}'. Consider removing \
-                                 or replacing it with '_'",
-                                v_str
-                            );
-                            context.error(vec![(l.loc, msg)]);
+                            if !v_str.starts_with('_') {
+                                let msg = format!(
+                                    "Unused assignment or binding for local '{}'. \
+                                    Consider removing or replacing it with '_'",
+                                    v_str
+                                );
+                                context.error(vec![(l.loc, msg)]);
+                            }
                             if !context.is_resourceful(v) {
                                 l.value = L::Ignore
                             }
--- a/language/move-lang/src/cfgir/liveness/mod.rs
+++ b/language/move-lang/src/cfgir/liveness/mod.rs
@@ -294,11 +294,17 @@ mod last_usage {
                     match display_var(v.value()) {
                         DisplayVar::Tmp => (),
                         DisplayVar::Orig(v_str) => {
-                            if !v_str.starts_with('_') {
+                            if !v.starts_with_underscore() {
+                                let verb = if context.is_resourceful(&v) {
+                                    // do not suggest prefixing for resource bindings
+                                    "replacing"
+                                } else {
+                                    "replacing / prefixing"
+                                };
                                 let msg = format!(
                                     "Unused assignment or binding for local '{}'. \
-                                    Consider removing or replacing it with '_'",
-                                    v_str
+                                    Consider removing or {} it with '_'",
+                                    v_str, verb
                                 );
                                 context.error(vec![(l.loc, msg)]);
                             }
--- a/language/move-lang/tests/move_check/typing/bind_pop_resource.exp
+++ b/language/move-lang/tests/move_check/typing/bind_pop_resource.exp
@@ -14,20 +14,6 @@ error:
 
 error: 
 
-   ┌── tests/move_check/typing/bind_pop_resource.move:6:13 ───
-   │
- 6 │         let _r: R = R{};
-   │             ^^ Cannot ignore resource values. The value must be used
-   ·
- 6 │         let _r: R = R{};
-   │                 - The type: '0x8675309::M::R'
-   ·
- 2 │     resource struct R {}
-   │     -------- Is found to be a non-copyable type here
-   │
-
-error: 
-
    ┌── tests/move_check/typing/bind_pop_resource.move:7:14 ───
    │
  7 │         let (_, _):(R, R) = (R{}, R{});
--- a/language/move-lang/tests/move_check/typing/bind_pop_resource.exp
+++ b/language/move-lang/tests/move_check/typing/bind_pop_resource.exp
@@ -14,12 +14,12 @@ error:
 
 error: 
 
-   ┌── tests/move_check/typing/bind_pop_resource.move:7:14 ───
+   ┌── tests/move_check/typing/bind_pop_resource.move:9:14 ───
    │
- 7 │         let (_, _):(R, R) = (R{}, R{});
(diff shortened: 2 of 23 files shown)


mengxu-fb:
Updated test, `almost_invalid_local_name.move` is still valid. The invalid test is moved to `invalid_local_name.move`.

language/move-lang/src/cfgir/liveness/mod.rs
```diff
@@ -294,12 +294,14 @@ mod last_usage {
                     match display_var(v.value()) {
                         DisplayVar::Tmp => (),
                         DisplayVar::Orig(v_str) => {
-                            let msg = format!(
-                                "Unused assignment or binding for local '{}'. Consider removing \
-                                 or replacing it with '_'",
-                                v_str
-                            );
-                            context.error(vec![(l.loc, msg)]);
+                            if !v_str.starts_with('_') {
```
tnowacki, line 297: `v.starts_with_underscore()` would make me feel a lot less nervous here

language/move-lang/src/cfgir/liveness/mod.rs
```diff
@@ -294,12 +294,14 @@ mod last_usage {
...
+                                let msg = format!(
+                                    "Unused assignment or binding for local '{}'. \
+                                    Consider removing or replacing it with '_'",
```
tnowacki, line 300: We should update this error message to suggest prepending it with `_`

language/move-lang/tests/move_check/typing/bind_pop_resource.exp
```diff
@@ -12,20 +12,6 @@ error:
...
- 6 │         let _r: R = R{};
-   │             ^^ Cannot ignore resource values. The value must be used
-   ·
- 6 │         let _r: R = R{};
-   │                 - The type: '0x8675309::M::R'
-   ·
- 2 │     resource struct R {}
-   │     -------- Is found to be a non-copyable type here
-   │
-
```
tnowacki, line 28: This should still be an error. what's going on here?

tnowacki: Or at least there should be a new error that says "unused resource in local _r"

language/move-lang/src/cfgir/liveness/mod.rs
mengxu-fb: Sure, I'll update it. This will lead to several updates to the `exp` files too.

language/move-lang/tests/move_check/typing/bind_pop_resource.exp
mengxu-fb: You are right. This should still be an error.

If I just put
```
module M {
    resource struct R {}

    fun t0() {
        let _r: R = R{};
    }
}
```

It gives
```
error:

   ┌── test.move:5:24 ───
   │
 5 │         let _r: R = R{};
   │                        ^ Invalid return
   ·
 5 │         let _r: R = R{};
   │             -- The local '_r' still contains a resource value due to this assignment. The resource must be consumed before the function returns
   │

```

But if I put `let _: R = R{};` above or below `let _r`, the error is gone...

Please allow me a bit more time to check this.

tnowacki: I think it's because I just rely on unused assignments to check that. You might need another case for:
- unused _ variables that are resourceful

mengxu-fb: I guess things might be a bit more complex there, even without the underscore, the `r` is not warned.
```
module M {
    resource struct R {}

    fun t0() {
        let _: R = R{};
        let r: R = R{};
    }
}
```

tnowacki: That's very bizarre... If you need help debugging this let me know

mengxu-fb: My suspicion is that the pass `LocalsSafety` is shortcut by unused assignment but let me look more into this. I think I can debug this one.

mengxu-fb:
```
        PassResult::Naming(nprog, errors) => {
            let (tprog, errors) = typing::translate::program(nprog, errors);
            check_errors(errors)?;
            ^^^^^^^^^^^^^^^^^^^^^^
            run(PassResult::Typing(tprog), until)
        }
        PassResult::Typing(tprog) => {
            let (hprog, errors) = hlir::translate::program(tprog);
            run(PassResult::HLIR(hprog, errors), until)
        }
        PassResult::HLIR(hprog, errors) => {
            let (cprog, errors) = cfgir::translate::program(errors, hprog);
            check_errors(errors)?;
            run(PassResult::CFGIR(cprog), until)
        }
        PassResult::CFGIR(cprog) => {
            let compiled_units = to_bytecode::translate::program(cprog)?;
            assert!(until == Pass::Compilation);
            run(PassResult::Compilation(compiled_units), Pass::Compilation)
        }
```

This `check_error()` short-circuited the execution such that the CFGIR (where `LocalSafety` lives) is not executed.

Removing it brings back the error, but I doubt that this is the right way to fix it
```
error:

   ┌── test.move:5:13 ───
   │
 5 │         let r: R = R{};
   │             ^ Unused assignment or binding for local 'r'. Consider removing it or replacing/prefixing it with '_'
   │

error:

   ┌── test.move:6:16 ───
   │
 6 │     let _: R = R{};
   │                   ^ Invalid return
   ·
 5 │         let r: R = R{};
   │             - The local 'r' still contains a resource value due to this assignment. The resource must be consumed before the function returns
   │

```

mengxu-fb: This might not be easy to fix, as reasoning whether a local is actually used or not needs CFGIR, but reasoning whether a `_` binds to a resource just needs `typing`.

This issue no impact on the overall correctness of the compiler. But it will leave certain errors (e.g., those checks that requires HLIR, CFGIR) undetected if there are naming or typing isssues.

tnowacki: Oh if its just an issue with that test its fine. It is well known that you do not get unused resource warnings until after typing has completed. This is normal and expected.

But you will still need to fix that one case for resources in that one test as I mentioned above

mengxu-fb: I think this is also the reason why that error is no longer showing in the `bind_pop_resource.move` case (i.e., the case you mentioned above)
```
$cat bind_pop_resource.move
module M {
    resource struct R {}

    fun t0() {
        let _: R = R{};
        let _r: R = R{};
        let (_, _):(R, R) = (R{}, R{});
    }
}
```

After making `_r` as binding instead of ignore, that error will require CFGIR to check while the others did not pass the `typing` pass. Hence the original error is not printed.

I think it's better for me to put a comment in `bind_pop_resource.move` to mark that this `let _r` is also wrong, but its error won't be shown. Or better still, create another test in `locals` with only `let _r: R = R{};` in `fun t0()`.

tnowacki: I would do both of those things, put a comment and make another test. Thanks!

mengxu-fb:
Rebased past the renaming. @tnowacki would you mind having a final look on the PR?

language/move-lang/src/cfgir/liveness/mod.rs
```diff
@@ -294,12 +294,20 @@ mod last_usage {
...
-                            let msg = format!(
-                                "Unused assignment or binding for local '{}'. Consider removing \
-                                 or replacing it with '_'",
-                                v_str
-                            );
-                            context.error(vec![(l.loc, msg)]);
+                            if !v.starts_with_underscore() {
+                                let verb = if context.is_resourceful(&v) {
+                                    // do not suggest prefixing for resource bindings
+                                    "replacing"
```
tnowacki, line 300: I think this is unnecessary, replacing it with `_` if it is resourceful will result in an error, so it might be okay to just say "replacing / prefixing" always. 

I might be a bit more explicit about prefixing so maybe:

"Unused assignment or binding for local '{:v_str}'. Consider removing, replacing with '_', or prefixing with '_' e.g. '_{:v_str}'"

That might be a bit wordy, but I think it is good to squeeze in the prefix example somehow

mengxu-fb:
Rebased to master.

tnowacki (approved):

mengxu-fb:
Rebased, see if this may avoid the Circle CI glitch (2xlarge docker).

PR #6789, opened 2020-12-02
[network] Integrate Rate limiting on incoming messages

gregnazario:
## Summary
Using `Governor`, we now have a hardcoded rate limiting for incoming
messages.  This will be extended to outbound messages in a future PR.

This doesn't currently provide any backpressure, and instead drops messages on a per frame level.  Future plans are to provide for providing backpressure

## Planning Issue
https://github.com/diem/diem/issues/6859

### DIFF
--- a/network/src/peer_manager/mod.rs
+++ b/network/src/peer_manager/mod.rs
@@ -55,7 +55,7 @@ mod error;
 mod tests;
 
 pub use self::error::PeerManagerError;
-use governor::{clock::DefaultClock, state::keyed::DefaultKeyedStateStore, RateLimiter};
+use crate::rate_limiter::RateLimiter;
 use serde::export::Formatter;
 use std::{net::IpAddr, num::NonZeroU32};
 
@@ -291,8 +291,8 @@ where
     max_frame_size: usize,
     /// Inbound connection limit separate of outbound connections
     inbound_connection_limit: usize,
-    /// Inbound request rate limiter TODO: adjust state store and clock
-    inbound_rate_limiter: Arc<RateLimiter<IpAddr, DefaultKeyedStateStore<IpAddr>, DefaultClock>>,
+    /// Inbound request rate limiter
+    inbound_rate_limiter: Arc<RateLimiter<IpAddr>>,
 }
 
 impl<TTransport, TSocket> PeerManager<TTransport, TSocket>
--- a/network/src/peer_manager/mod.rs
+++ b/network/src/peer_manager/mod.rs
@@ -19,7 +19,7 @@ use crate::{
         direct_send::Message,
         rpc::{error::RpcError, InboundRpcRequest, OutboundRpcRequest},
     },
-    rate_limiter, transport,
+    transport,
     transport::{Connection, ConnectionId, ConnectionMetadata},
     ProtocolId,
 };
@@ -57,7 +57,7 @@ mod tests;
 pub use self::error::PeerManagerError;
 use crate::rate_limiter::RateLimiter;
 use serde::export::Formatter;
-use std::{net::IpAddr, num::NonZeroU32};
+use std::net::IpAddr;
 
 /// Request received by PeerManager from upstream actors.
 #[derive(Debug, Serialize)]
@@ -319,6 +319,7 @@ where
         max_concurrent_network_notifs: usize,
         max_frame_size: usize,
         inbound_connection_limit: usize,
+        inbound_rate_limiter: Option<Arc<RateLimiter<IpAddr>>>,
     ) -> Self {
         let (transport_notifs_tx, transport_notifs_rx) = channel::new(
             channel_size,
@@ -338,7 +339,6 @@ where
                 transport_notifs_tx_clone,
             )
         });
-        let default_rate = NonZeroU32::new(102_400).unwrap(); // 100 KiB
 
         Self {
             network_context,
@@ -360,11 +360,7 @@ where
             channel_size,
             max_frame_size,
             inbound_connection_limit,
-            // FIXME: Use a config configured quota
-            inbound_rate_limiter: Some(rate_limiter::new_per_second_keyed(
-                default_rate,
-                default_rate,
-            )),
+            inbound_rate_limiter,
         }
     }
 
--- a/network/src/peer_manager/mod.rs
+++ b/network/src/peer_manager/mod.rs
@@ -292,7 +292,7 @@ where
     /// Inbound connection limit separate of outbound connections
     inbound_connection_limit: usize,
     /// Inbound request rate limiter
-    inbound_rate_limiter: Arc<RateLimiter<IpAddr>>,
+    inbound_rate_limiter: Option<Arc<RateLimiter<IpAddr>>>,
 }
 
 impl<TTransport, TSocket> PeerManager<TTransport, TSocket>
@@ -361,7 +361,10 @@ where
             max_frame_size,
             inbound_connection_limit,
             // FIXME: Use a config configured quota
-            inbound_rate_limiter: rate_limiter::new_per_second_keyed(default_rate, default_rate),
+            inbound_rate_limiter: Some(rate_limiter::new_per_second_keyed(
+                default_rate,
+                default_rate,
+            )),
         }
     }
 
--- a/network/src/peer_manager/mod.rs
+++ b/network/src/peer_manager/mod.rs
@@ -19,7 +19,7 @@ use crate::{
         direct_send::Message,
         rpc::{error::RpcError, InboundRpcRequest, OutboundRpcRequest},
     },
-    transport,
+    rate_limiter, transport,
     transport::{Connection, ConnectionId, ConnectionMetadata},
     ProtocolId,
 };
@@ -55,7 +55,9 @@ mod error;
 mod tests;
 
 pub use self::error::PeerManagerError;
+use governor::{clock::DefaultClock, state::keyed::DefaultKeyedStateStore, RateLimiter};
 use serde::export::Formatter;
+use std::{net::IpAddr, num::NonZeroU32};
 
 /// Request received by PeerManager from upstream actors.
 #[derive(Debug, Serialize)]
@@ -289,6 +291,8 @@ where
     max_frame_size: usize,
     /// Inbound connection limit separate of outbound connections
     inbound_connection_limit: usize,
+    /// Inbound request rate limiter TODO: adjust state store and clock
+    inbound_rate_limiter: Arc<RateLimiter<IpAddr, DefaultKeyedStateStore<IpAddr>, DefaultClock>>,
 }
 
 impl<TTransport, TSocket> PeerManager<TTransport, TSocket>
@@ -334,6 +338,8 @@ where
                 transport_notifs_tx_clone,
             )
         });
+        let default_rate = NonZeroU32::new(102_400).unwrap(); // 100 KiB
+
         Self {
             network_context,
             executor,
@@ -354,6 +360,8 @@ where
             channel_size,
             max_frame_size,
             inbound_connection_limit,
+            // FIXME: Use a config configured quota
+            inbound_rate_limiter: rate_limiter::new_per_second_keyed(default_rate, default_rate),
         }
     }
 
@@ -771,6 +779,7 @@ where
             self.max_concurrent_network_notifs,
             self.channel_size,
             self.max_frame_size,
+            self.inbound_rate_limiter.clone(),
         );
         // Start background task to handle events (RPCs and DirectSend messages) received from
         // peer.
--- a/network/src/protocols/wire/messaging/v1/mod.rs
+++ b/network/src/protocols/wire/messaging/v1/mod.rs
@@ -9,14 +9,16 @@
 //! describes in greater detail how these messages are sent and received
 //! over-the-wire.
 
-use crate::{protocols::wire::handshake::v1::ProtocolId, rate_limiter::rate_limit_msg};
+use crate::{
+    protocols::wire::handshake::v1::ProtocolId,
+    rate_limiter::{rate_limit_msg, RateLimiter},
+};
 use bytes::Bytes;
 use futures::{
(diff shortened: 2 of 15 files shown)


network/src/peer_manager/mod.rs
```diff
@@ -354,6 +362,11 @@ where
             channel_size,
             max_frame_size,
             inbound_connection_limit,
+            // FIXME: Don't make this quota hard coded, also it doesn't support more than max u32 here
+            // Allow 1MB per second
+            inbound_rate_limiter: Arc::new(governor::RateLimiter::keyed(Quota::per_second(
+                NonZeroU32::new(1_000_000).unwrap(),
```
davidiw, line 368: also 1 MB/s is a looot we probably can suffice 50 KB/s or substantially less...

network/src/peer_manager/mod.rs
```diff
@@ -816,24 +829,44 @@ where
...
         let network_context = self.network_context.clone();
+        let inbound_rate_limiter = self.inbound_rate_limiter.clone();
         self.executor.spawn(network_events.for_each_concurrent(
             self.max_concurrent_network_reqs,
             move |inbound_event| {
                 Self::handle_inbound_event(
                     network_context.clone(),
                     inbound_event,
                     peer_id,
+                    inbound_rate_limiter.clone(),
```
davidiw, line 840: you clone, so that you can clone? what is this weirdness? same with network context :O

does this have something to do with self not being within the async context?

network/src/peer_manager/mod.rs
```diff
@@ -816,24 +829,44 @@ where
...
 
+    fn rate_limit<'a>(
+        inbound_rate_limiter: &'a RateLimiter<PeerId, DefaultKeyedStateStore<PeerId>, DefaultClock>,
+        peer_id: &PeerId,
+        msg_length: usize,
+    ) -> Result<(), NegativeMultiDecision<NotUntil<'a, QuantaInstant>>> {
+        // We can't rate limit something of 0 size
+        if msg_length < 1 {
+            return Ok(());
+        }
```
davidiw, line 856: what does it matter?

network/src/peer_manager/mod.rs
```diff
@@ -816,24 +829,44 @@ where
...
+        let length = NonZeroU32::new(msg_length as u32).unwrap();
```
davidiw, line 857: why unwrap in something that returns a result?

network/src/peer_manager/mod.rs
```diff
@@ -816,24 +829,44 @@ where
...
+
+        // TODO Governor doesn't support larger than u32, this may be problematic for large messages
```
davidiw, line 859: maybe we should just reject larger messages by ensuring that msg <= u32::max first

davidiw (commented):
I think I had in mind that we wouldn't even read from the wire so I think this is currently too high in the stack.

For incoming we should enable TCP back pressure which can only be achieved by letting the queue fill up. This approach would only limit CPU based DoS but not a single TCP stream or IP based.

Likewise, I had thought we might make the rate limiter based upon IP. Btw how does the GC work?

For outgoing, we should probably figure out how to introduce something such that the none of the channels coming into the network are being read so we push back on the applications or drop messages on the floor.

network/src/peer_manager/mod.rs
gregnazario: This has to deal with the `for_each_concurrent` then `move` :(

network/src/peer_manager/mod.rs
gregnazario: Yeah, it was an arbitrary limit, and I was pretty sure this was in draft...

network/src/peer_manager/mod.rs
gregnazario: This is because Governor takes a `NonZeroU32`.  So, if it was of size 0, it would fail on the result down below.

The reason for all of this mess was because I was working around a returned lifetime value that coincides with the `inbound_rate_limiter` lifetime in itself.  So, after a lot of iteration I got here just to get this function to work so I wouldn't have to repeat this block twice.

network/src/peer_manager/mod.rs
gregnazario: This is required in the next line to not be a result?

network/src/peer_manager/mod.rs
gregnazario: Once again, this was a draft I was going to look into that next

gregnazario:
> I think I had in mind that we wouldn't even read from the wire so I think this is currently too high in the stack.

I was playing with where to put the rate limiting, and it's not done.

> For incoming we should enable TCP back pressure which can only be achieved by letting the queue fill up. This approach would only limit CPU based DoS but not a single TCP stream or IP based.

Our system isn't built for backpressure, I think I've discussed this before with you that I was going to drop messages to start, and add the backpressure if necessary.

> Likewise, I had thought we might make the rate limiter based upon IP. Btw how does the GC work?

I did this based on PeerId in this example because of your previous comment to me saying PeerId should be enough instead of IP :/

GC handled by Governor and uses a recently used tracker to get rid of unused keys.  This would probably be run periodically

> For outgoing, we should probably figure out how to introduce something such that the none of the channels coming into the network are being read so we push back on the applications or drop messages on the floor.

Outgoing is not currently captured here as it's only doing incoming.  I'm probably going forward with dropping messages to start and then backpressure similar to the inbound.  This should be simpler than adding the backpressure on inbound (hopefully)

libra-action:
```
target packages:
  A governor 0.3.1 (direct third-party, crates.io)
    * features: dashmap, default, futures, futures-timer, jitter, quanta, rand, std
  A ahash 0.3.8 (transitive third-party, crates.io)
    * features: compile-time-rng, const-random, default, std
  A const-random 0.1.13 (transitive third-party, crates.io)
    * features: [none]
  A dashmap 3.11.10 (transitive third-party, crates.io)
    * features: default
  A futures-timer 3.0.2 (transitive third-party, crates.io)
    * features: [none]
  A hashbrown 0.8.2 (transitive third-party, crates.io) (other versions: 0.9.1)
    * features: ahash, default, inline-more
  A no-std-compat 0.4.1 (transitive third-party, crates.io)
    * features: alloc, compat_hash, hashbrown, std
  A nonzero_ext 0.2.0 (transitive third-party, crates.io)
    * features: std
  A quanta 0.4.1 (transitive third-party, crates.io)
    * features: [none]

host packages:
  A const-random-macro 0.1.13 (transitive third-party, crates.io)
    * features: [none]
  A crunchy 0.2.2 (transitive third-party, crates.io)
    * features: default, limit_128
  A getrandom 0.2.0 (transitive third-party, crates.io) (other versions: 0.1.15)
    * features: [none]
  A tiny-keccak 2.0.2 (transitive third-party, crates.io)
    * features: default, sha3
```

libra-action:
```
target packages:
  A governor 0.3.1 (direct third-party, crates.io)
    * features: dashmap, default, futures, futures-timer, jitter, quanta, rand, std
  A ahash 0.3.8 (transitive third-party, crates.io)
    * features: compile-time-rng, const-random, default, std
  A const-random 0.1.13 (transitive third-party, crates.io)
    * features: [none]
  A dashmap 3.11.10 (transitive third-party, crates.io)
    * features: default
  A futures-timer 3.0.2 (transitive third-party, crates.io)
    * features: [none]
  A hashbrown 0.8.2 (transitive third-party, crates.io) (other versions: 0.9.1)
    * features: ahash, default, inline-more
  A no-std-compat 0.4.1 (transitive third-party, crates.io)
    * features: alloc, compat_hash, hashbrown, std
  A nonzero_ext 0.2.0 (transitive third-party, crates.io)
    * features: std
  A quanta 0.4.1 (transitive third-party, crates.io)
    * features: [none]

host packages:
  A const-random-macro 0.1.13 (transitive third-party, crates.io)
    * features: [none]
  A crunchy 0.2.2 (transitive third-party, crates.io)
    * features: default, limit_128
  A getrandom 0.2.0 (transitive third-party, crates.io) (other versions: 0.1.15)
    * features: [none]
  A tiny-keccak 2.0.2 (transitive third-party, crates.io)
    * features: default, sha3
```

libra-action:
```
target packages:
  A governor 0.3.1 (direct third-party, crates.io)
    * features: dashmap, default, futures, futures-timer, jitter, quanta, rand, std
  A ahash 0.3.8 (transitive third-party, crates.io)
    * features: compile-time-rng, const-random, default, std
  A const-random 0.1.13 (transitive third-party, crates.io)
    * features: [none]
  A dashmap 3.11.10 (transitive third-party, crates.io)
    * features: default
  A futures-timer 3.0.2 (transitive third-party, crates.io)
    * features: [none]
  A hashbrown 0.8.2 (transitive third-party, crates.io) (other versions: 0.9.1)
    * features: ahash, default, inline-more
  A no-std-compat 0.4.1 (transitive third-party, crates.io)
    * features: alloc, compat_hash, hashbrown, std
  A nonzero_ext 0.2.0 (transitive third-party, crates.io)
    * features: std
  A quanta 0.4.1 (transitive third-party, crates.io)
    * features: [none]

host packages:
  A const-random-macro 0.1.13 (transitive third-party, crates.io)
    * features: [none]
  A crunchy 0.2.2 (transitive third-party, crates.io)
    * features: default, limit_128
  A getrandom 0.2.0 (transitive third-party, crates.io) (other versions: 0.1.15)
    * features: [none]
  A tiny-keccak 2.0.2 (transitive third-party, crates.io)
    * features: default, sha3
```

network/src/peer_manager/mod.rs
gregnazario: Done.

libra-action:
```
target packages:
  A governor 0.3.1 (direct third-party, crates.io)
    * features: dashmap, default, futures, futures-timer, jitter, quanta, rand, std
  A ahash 0.3.8 (transitive third-party, crates.io)
    * features: compile-time-rng, const-random, default, std
  A const-random 0.1.13 (transitive third-party, crates.io)
    * features: [none]
  A dashmap 3.11.10 (transitive third-party, crates.io)
    * features: default
  A futures-timer 3.0.2 (transitive third-party, crates.io)
    * features: [none]
  A hashbrown 0.8.2 (transitive third-party, crates.io) (other versions: 0.9.1)
    * features: ahash, default, inline-more
  A no-std-compat 0.4.1 (transitive third-party, crates.io)
    * features: alloc, compat_hash, hashbrown, std
  A nonzero_ext 0.2.0 (transitive third-party, crates.io)
    * features: std
  A quanta 0.4.1 (transitive third-party, crates.io)
    * features: [none]

host packages:
  A const-random-macro 0.1.13 (transitive third-party, crates.io)
    * features: [none]
  A crunchy 0.2.2 (transitive third-party, crates.io)
    * features: default, limit_128
  A getrandom 0.2.0 (transitive third-party, crates.io) (other versions: 0.1.15)
    * features: [none]
  A tiny-keccak 2.0.2 (transitive third-party, crates.io)
    * features: default, sha3
```

network/src/protocols/wire/messaging/v1/mod.rs
```diff
@@ -154,27 +159,58 @@ pub fn network_message_frame_codec(max_frame_size: usize) -> LengthDelimitedCode
...
             Poll::Ready(Some(Ok(frame))) => {
                 let frame = frame.freeze();
 
+                // Handle rate limiting of messages
+                if let Some(rate_limiter) = &stream.rate_limiter {
+                    if let Err(_error) = rate_limit_msg(rate_limiter, &stream.ip_addr, frame.len())
+                    {
+                        // TODO save message until it's ready
+                        // Drop message, it's been throttled
+                        return Poll::Pending;
```
phlip9, line 210: the `Stream` will hang here if we just return `Poll::Pending` without registering our `cx.waker` somewhere. When we return `Poll::Pending`, we're telling the parent future/stream/runtime: "hey, I don't have any more work to do and you don't need to poll me again until someone has triggered my waker`. 

This is why we pass our `cx` (which contains our waker) to our child `framed_read.poll_next(cx)`. When there's no more data on the socket, it'll return `Poll::Pending` and register the waker with the epoll reactor to wake us up when there's more data available.

In this case, we probably want to sleep a bit or maybe just yield immediately back to the scheduler like

```rust
cx.waker.wake();
return Poll::Pending;
```

Sleeping is tougher:

```rust
use futures::future::Future;
use futures::ready;
use std::time::Duration;
use tokio::time::{delay_for, Delay};

#[pin_project]
pub struct NetworkMessageStream {
    #[pin]
    framed_read: /* .. */,
    // When we have Some(delay), we'll always finish polling it (the delay has finished)
    // before polling the framed_read.
    // Note: don't need pin here b/c Delay: Unpin
    delay: Option<Delay>,
}

impl NetworkMessageStream {
    pub fn new(socket, max_frame_size) -> Self {
        // ..
        Self { framed_read, delay: None }
    }
}

impl Stream for NetworkMessageStream {
    fn poll_next(mut self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Option<Self::Item>> {
        if self.delay.is_some() {
            // ready! will return Poll::Pending if delay.poll(cx) returns Poll::Pending
            ready!(Pin::new(self.as_mut().project().delay.as_mut().unwrap()).poll(cx));

            // delay is done, set it to None
            *self.as_mut().project().delay = None;
        }

        // finally read the next frame
        match self.as_mut().project().framed_read.poll_next(cx) {
            Poll::Ready(Some(Ok(frame))) => {
                let frame = frame.freeze();

                // we want to delay before reading the next frame, so we set delay to Some(Delay)
                if rate_limited {
                    let delay_time = /* .. */;
                    *self.as_mut().project().delay = Some(delay_for(Duration::from_millis(delay_time)));
                }
                
                // ..
        }
    }

```

phlip9 (changes requested):
I guess I'd like to see the implementation with backpressure and maybe some more points on how we mitigate the issues with rate limiting at the framing layer

PR #6793, opened 2020-12-03
[rustsec] Bump time to 0.2.23

sausagee:
## Motivation
Bump the time crate to 0.2.23 to address RUSTSEC-2020-0071.

```
RUSTSEC-2020-0071: time: Potential segfault in the time crate
```

## Test Plan
cargo audit locally + CI

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3639,7 +3639,7 @@ dependencies = [
  "serde_yaml",
  "static_assertions",
  "thiserror",
- "time 0.2.22",
+ "time 0.2.23",
  "tokio",
  "url",
 ]
@@ -5621,7 +5621,7 @@ dependencies = [
  "rustc_version",
  "serde",
  "sha2",
- "time 0.2.22",
+ "time 0.2.23",
  "tokio",
 ]
 
@@ -6788,9 +6788,9 @@ dependencies = [
 
 [[package]]
 name = "time"
-version = "0.2.22"
+version = "0.2.23"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "55b7151c9065e80917fbf285d9a5d1432f60db41d170ccafc749a136b41a93af"
+checksum = "bcdaeea317915d59b2b4cd3b5efcd156c309108664277793f5351700c02ce98b"
 dependencies = [
  "const_fn",
  "libc",


Cargo.lock
```diff
@@ -2613,7 +2613,7 @@ dependencies = [
  "serde_yaml",
  "static_assertions",
  "thiserror",
- "time 0.2.22",
+ "time 0.2.23",
```
mimoo, line 2616: so what does it mean here that some dependencies have a version set, and some don't? (cc @sunshowers)

mimoo: (my guess is that dependencies that DON'T have a version set here are using the same version as we use when we directly import them, if this is the case then we should be able to remove the version here.)

mimoo: (interestingly I found https://docs.rs/cargo-lock/6.0.0/cargo_lock/ while searching for the answer)

sunshowers: This is the Cargo.lock file, it's not human editable. To answer your question, versions get set when there are multiple versions of the same package and you need to disambiguate by version.

rexhoffman:
Alright, why didn't dependabot show us this vulnerability?   Let's hold off until I've some time to investigate/change dependabot config today?

Cargo.lock
rexhoffman: Many of the versions in the lock file are set via dependabot prs, which I suspect use cargo-edits, `cargo update` command.

mimoo:
here's the diff for `time` btw:

[report.html.zip](https://github.com/libra/libra/files/5637818/report.html.zip)

rexhoffman:
@sunshowers according to guppy this pr is a no-op....

Cargo.lock
rexhoffman: Okay so this is only for the kube/rusoto crates which I'm pretty sure are test only code.   So maybe ignored by guppy -- but dependabot?

sunshowers:
From https://app.circleci.com/pipelines/github/libra/libra/35277/workflows/9d4e2071-fa0d-40ff-a995-406ba48b8d51/jobs/258063 it looks like the only affected package is `cluster-test`.

Cargo.lock
sunshowers: @rexhoffman I think it might be because time is only a transitive dependency, not a direct one...

this is a pretty bad bug though, and maybe we should prioritize doing our own dependabot

rexhoffman:
Got it, we don't report changes to non-prod code.

metajack:
Is dependabot supposed to do security notifications for Rust code? I know it does for nodejs, etc.

We already have cargo-audit running nightly which will catch this. I assume this is how @sausagee found it in the first place?

metajack:
See https://app.circleci.com/pipelines/github/novifinancial/libra-ops/2609/workflows/e6656ec5-0254-4dd0-8865-e6ecb00c2345/jobs/10205

This gets reported to Novi slack #security channel.

sausagee:
> See https://app.circleci.com/pipelines/github/novifinancial/libra-ops/2609/workflows/e6656ec5-0254-4dd0-8865-e6ecb00c2345/jobs/10205
> 
> This gets reported to Novi slack #security channel.

Yep. I did a spot check of our various alerts yesterday and found it. 

Ideally dependabot should prioritize RUSTSEC update over routine version update.

rexhoffman:
Well dependabot and github's dependabot security updates appear to be two seperate systems with different package system support.    

Rust is not included in the security updates from github's dependabot:   https://docs.github.com/en/free-pro-team@latest/github/visualizing-repository-data-with-graphs/about-the-dependency-graph#supported-package-ecosystems.

But as for dependabot proper I would have expected to see an update for the time crate regardless.    We aren't limiting to direct dependencies.    I'll explicitly set transitive deps to true and see if that changes anything.

https://docs.github.com/en/free-pro-team@latest/github/administering-a-repository/configuration-options-for-dependency-updates#allow

metajack (approved):

rexhoffman:
After updating dependabot to look at transitive dependencies as well this dependency is missed.   It seems that non-release transitive dependencies are ignored?   It's not entirely clear to me why this is missed by dependabot's rust implementation.

rexhoffman:
If I was certain this isn't a result of our use of cargo nightly or feature flags I'd open an issue with dependabot, but will refrain from doing so until I have more time to investigate.

rexhoffman (approved):

PR #6801, opened 2020-12-03
[testnet-faucet] add an option to create dd account by mint service

xli:
## Motivation

Supports creating DD account on testnet by Faucet mint service, so that testnet can be used as initial integration environment for testing business features related to liquidity transferring from DD to VASP.

YES

## Test Plan

Unit test, and local manual test with local validator, faucet services.

### DIFF
--- a/client/faucet/src/main.rs
+++ b/client/faucet/src/main.rs
@@ -234,6 +234,38 @@ mod tests {
         let addr = AccountAddress::try_from("a74fd7c46952c497e75afb0a7932586d".to_owned()).unwrap();
         let account = reader.get(&addr).expect("account should be created");
         assert_eq!(account["balances"][0]["amount"], amount);
+        assert_eq!(account["role"]["type"], "parent_vasp");
+    }
+
+    #[tokio::test]
+    async fn test_mint_dd_account_with_txns_response() {
+        let accounts = genesis_accounts();
+        let service = setup(accounts.clone());
+        let filter = routes(service);
+
+        let auth_key = "44b8f03f203ec45dbd7484e433752efe54aa533116e934f8a50c28bece06d3ac";
+        let amount = 13345;
+        let resp = warp::test::request()
+            .method("POST")
+            .path(
+                format!(
+                    "/mint?auth_key={}&amount={}&currency_code=LBR&return_txns=true&is_designated_dealer=true",
+                    auth_key, amount
+                )
+                .as_str(),
+            )
+            .reply(&filter)
+            .await;
+        let body = resp.body();
+        let txns: Vec<libra_types::transaction::SignedTransaction> =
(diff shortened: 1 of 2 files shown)


client/faucet/src/mint.rs
```diff
@@ -49,7 +50,23 @@ impl MintParams {
...
+    }
+
+    fn create_designated_dealer_account_script(
+        &self,
+        seq: u64,
+    ) -> libra_types::transaction::TransactionPayload {
+        libra_types::transaction::TransactionPayload::Script(
+            transaction_builder_generated::stdlib::encode_create_designated_dealer_script(
+                self.currency_code(),
+                0, // sliding nonce
```
davidiw, line 66: shouldn't this be seq?

xli: encode_create_designated_dealer_script,

this is not creating transaction, it is script parameter.

client/faucet/src/main.rs
```diff
@@ -309,11 +341,17 @@ mod tests {
...
+                            let mut writer = accounts.write();
+                            let previous =
+                                writer.insert(address, create_dd_account(&address.to_string(), 0));
+                            assert!(previous.is_none(), "should not create account twice");
+                        }
                         Some(ScriptCall::PeerToPeerWithMetadata { payee, amount, .. }) => {
                             let mut writer = accounts.write();
                             let account =
                                 writer.get_mut(&payee).expect("account should be created");
-                            *account = create_vasp_account(payee.to_string().as_str(), amount);
```
sunmilee, line 316: How come this is not needed anymore?

client/faucet/src/mint.rs
davidiw: oh interesting, I had been inserting the sequence number here too in all my cases but it looks like that's not required

client/faucet/src/mint.rs
```diff
@@ -99,11 +116,12 @@ impl Service {
...
-                params.create_parent_vasp_account_script(tc_seq),
-                treasury_compliance_account_address(),
-                tc_seq,
-            )?);
+            let script = if params.is_designated_dealer.unwrap_or(false) {
+                params.create_designated_dealer_account_script(tc_seq)
+            } else {
+                params.create_parent_vasp_account_script(tc_seq)
+            };
+            txns.push(self.create_txn(script, treasury_compliance_account_address(), tc_seq)?);
```
davidiw, line 124: @sunmilee looks like it was probably redundant...

davidiw (approved):

client/faucet/src/main.rs
xli: new implementation direct update the balances amount:  account["balances"][0]["amount"] = serde_json::json!(amount);

old implementation recreates vasp account json, but won't work well anymore because depending on payee account role, it maybe DD or VASP.

PR #6802, opened 2020-12-03
[jsonrpc][client] Support multiple endpoints in JSON-RPC client

ankushagarwal:
## Motivation

Support talking to multiple JSON-RPC servers in parallel and return the response with the highest libra_ledger_version

### Summary

* async_client now optionally takes a list of JSON-RPC endpoints instead of a single one
* The user of this client can specify how many requests to make in parallel
* From the responses received from multiple servers, we choose the one with the highest libra_ledger_version
* For batch requests, select the response batch which has the highest libra_ledger_version

## Test Plan

* Added unit tests for methods which select the best response from a pool of responses obtained from multiple servers

### DIFF
--- a/client/json-rpc/src/async_client/client.rs
+++ b/client/json-rpc/src/async_client/client.rs
@@ -2,155 +2,42 @@
 // SPDX-License-Identifier: Apache-2.0
 
 use crate::async_client::{
-    defaults, types as jsonrpc, Error, JsonRpcResponse, RetryStrategy, State,
-    WaitForTransactionError,
+    defaults, types as jsonrpc, BroadcastHttpClient, Error, HttpClient, JsonRpcResponse, Request,
+    Response, RetryStrategy, SimpleHttpClient, State, WaitForTransactionError,
 };
 use diem_crypto::hash::CryptoHash;
 use diem_types::{
     account_address::AccountAddress,
     transaction::{SignedTransaction, Transaction},
 };
+
 use serde::de::DeserializeOwned;
-use serde_json::json;
-use std::{
-    collections::HashMap,
-    convert::{TryFrom, TryInto},
-    time::Duration,
-};
 
-#[derive(Debug)]
-pub struct Request {
-    pub method: &'static str,
-    pub params: serde_json::Value,
-}
+use std::{convert::TryInto, sync::Arc, time::Duration};
 
-impl Request {
-    pub fn new(method: &'static str, params: serde_json::Value) -> Self {
-        Request { method, params }
-    }
-
-    pub fn submit(txn: &SignedTransaction) -> Result<Self, lcs::Error> {
-        let txn_payload = hex::encode(lcs::to_bytes(txn)?);
-        Ok(Self::new("submit", json!([txn_payload])))
-    }
-
-    pub fn get_account_state_with_proof(
-        address: &AccountAddress,
-        from_version: Option<u64>,
-        to_version: Option<u64>,
-    ) -> Self {
-        Self::new(
-            "get_account_state_with_proof",
-            json!([address, from_version, to_version]),
-        )
-    }
-
-    pub fn get_state_proof(from_version: u64) -> Self {
-        Self::new("get_state_proof", json!([from_version]))
-    }
-
-    pub fn get_currencies() -> Self {
-        Self::new("get_currencies", json!([]))
-    }
-
-    pub fn get_events(key: &str, start_seq: u64, limit: u64) -> Self {
-        Self::new("get_events", json!([key, start_seq, limit]))
-    }
-
-    pub fn get_transactions(start_seq: u64, limit: u64, include_events: bool) -> Self {
-        Self::new(
-            "get_transactions",
-            json!([start_seq, limit, include_events]),
-        )
-    }
-
-    pub fn get_account_transactions(
-        address: &AccountAddress,
-        start_seq: u64,
-        limit: u64,
-        include_events: bool,
-    ) -> Self {
-        Self::new(
-            "get_account_transactions",
-            json!([address, start_seq, limit, include_events]),
-        )
-    }
-
-    pub fn get_account_transaction(
-        address: &AccountAddress,
-        seq: u64,
-        include_events: bool,
-    ) -> Self {
-        Self::new(
-            "get_account_transaction",
-            json!([address, seq, include_events]),
-        )
-    }
-
-    pub fn get_account(address: &AccountAddress) -> Self {
-        Self::new("get_account", json!([address]))
-    }
-    pub fn get_metadata_by_version(version: u64) -> Self {
-        Self::new("get_metadata", json!([version]))
-    }
-
-    pub fn get_metadata() -> Self {
-        Self::new("get_metadata", json!([]))
-    }
-
-    pub fn to_json(&self, id: usize) -> serde_json::Value {
-        json!({"jsonrpc": "2.0", "id": id, "method": self.method, "params": self.params})
-    }
-}
-
-#[derive(Debug)]
-pub struct Response<R> {
-    pub result: R,
-    pub state: State,
-}
-
-impl<R> std::ops::Deref for Response<R> {
-    type Target = R;
-
-    fn deref(&self) -> &Self::Target {
-        &self.result
-    }
-}
-
-impl<R: for<'de> serde::Deserialize<'de>> TryFrom<JsonRpcResponse> for Response<R> {
-    type Error = Error;
-
-    fn try_from(resp: JsonRpcResponse) -> Result<Self, Error> {
-        let state = State::from_response(&resp);
-        match resp.result {
-            Some(ret) => Ok(Self {
-                result: serde_json::from_value::<R>(ret)
-                    .map_err(Error::DeserializeResponseJsonError)?,
-                state,
-            }),
-            None => Err(Error::ResultNotFound(resp)),
-        }
-    }
-}
-
-#[derive(Debug)]
 pub struct Client<R> {
-    pub http_client: reqwest::Client,
-    pub server_url: reqwest::Url,
-    pub last_known_state: std::sync::RwLock<Option<State>>,
+    pub http_client: Arc<dyn HttpClient>,
     pub retry: R,
 }
 
 impl<R: RetryStrategy> Client<R> {
     pub fn from_url<T: reqwest::IntoUrl>(server_url: T, retry: R) -> Result<Self, reqwest::Error> {
         Ok(Self {
-            http_client: reqwest::ClientBuilder::new()
-                .use_native_tls()
-                .timeout(defaults::HTTP_REQUEST_TIMEOUT)
-                .build()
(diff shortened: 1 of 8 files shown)


libra-action:
```
target packages:
  M serde-generate 0.16.0 (direct third-party, crates.io)
    * version DOWNGRADED from 0.16.1
```

client/json-rpc/src/async_client/client.rs
```diff
@@ -362,9 +400,27 @@ impl<R: RetryStrategy> Client<R> {
 
     pub async fn send_without_retry(&self, request: &Request) -> Result<JsonRpcResponse, Error> {
         let id = 1;
-        let rpc_resp: JsonRpcResponse = self.send_json_request(request.to_json(id)).await?;
-        self.validate(&rpc_resp, id, id)?;
-        Ok(rpc_resp)
```
xli, line 367: do not change logic in this func

client/json-rpc/src/async_client/client.rs
```diff
@@ -362,9 +400,27 @@ impl<R: RetryStrategy> Client<R> {
...
+        let rpc_resp: Vec<JsonRpcResponse> =
```
xli, line 403: the following specific logic should be inside your ClusterRequestStrategy, this way, you don't need change batch_send_without_retry method

client/json-rpc/src/async_client/client.rs
```diff
@@ -432,39 +488,60 @@ impl<R: RetryStrategy> Client<R> {
...
+                        resp_maps
+                            .remove(&i)
+                            .ok_or_else(|| Error::unexpected_no_response(req.clone()))
+                    })
+                    .collect::<Result<Vec<JsonRpcResponse>, Error>>()
             })
-            .collect()
-    }
-
-    async fn send_json_request<T: for<'de> serde::Deserialize<'de>>(
```
xli, line 461: here, call the HttpClient to send request

         return self.http_client.send(request)

the default httpclient send method should do same with single_json_request.

client/json-rpc/src/async_client/client.rs
```diff
@@ -475,6 +552,50 @@ impl<R: RetryStrategy> Client<R> {
         resp.json().await.map_err(Error::InvalidHTTPResponse)
     }
 
+    async fn multi_send_json_request<T: for<'de> serde::Deserialize<'de>>(
```
xli, line 555: put all these into new LBHttpClient

client/json-rpc/src/async_client/client.rs
```diff
@@ -136,7 +139,8 @@ impl<R: for<'de> serde::Deserialize<'de>> TryFrom<JsonRpcResponse> for Response<
 #[derive(Debug)]
 pub struct Client<R> {
     pub http_client: reqwest::Client,
-    pub server_url: reqwest::Url,
+    pub server_urls: Vec<reqwest::Url>,
+    pub num_parallel_requests: usize,
```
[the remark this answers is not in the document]
xli: the benefit of this is making batch and non-batch same. 

the only thing may need more work is move most of the validate logic (see func validate) into the HttpClient,
the validate func need to be splitted into 2:
1. validate id
2. validate others

the validate id part is different for batch / non-batch, hence we can pull it out to each function.
for others validation, they can be put into ReqwestHttpClient.

Inside LBHttpClient, it should calls to ReqwestHttpClient for sending out one request to one server.

so LBHttpClient may just consider init multiple ReqwestHttpClient during initialization, then the parallel call is a simple fan out and then pick right response.

client/json-rpc/src/async_client/client.rs
```diff
@@ -475,6 +552,50 @@ impl<R: RetryStrategy> Client<R> {
...
+            let url = url.clone();
+            let req = req.clone();
+            async move {
+                match timeout(
+                    defaults::TIMEOUT,
+                    self.single_json_request(url.clone(), req),
+                )
+                .await
+                {
+                    Err(_) => Err(Error::ResponseTimeout(format!(
```
rgamba, line 575: I'm thinking about the case where one of the servers is down, we'd constantly be hitting this timeout. Should we consider caching the bad servers or something like that to avoid this?

rgamba (commented):
This is great, thanks!

client/json-rpc/src/async_client/client.rs
```diff
@@ -475,6 +552,50 @@ impl<R: RetryStrategy> Client<R> {
...
+        &self,
+        req: serde_json::Value,
+    ) -> Result<Vec<T>, Error> {
+        // Choose `num_parallel_requests` from `self.server_urls` to send requests to
+        let chosen_urls: Vec<reqwest::Url> = self
+            .server_urls
+            .choose_multiple(&mut rand::thread_rng(), self.num_parallel_requests)
+            .cloned()
+            .collect();
+        let futures = chosen_urls.iter().map(|url| {
```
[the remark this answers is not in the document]
ankushagarwal: I think its safe because commits are atomic. @xli or @davidiw to confirm

client/json-rpc/src/async_client/client.rs
ankushagarwal: It's a good point. Depends on what down means. If the server is completely down, we should be hitting a connection error and returning immediately, this would really happen if the server is up, but really slow to respond.

The idea of caching bad servers sounds good, but it's beyond the scope of this PR.

ankushagarwal:
Addressed comments, added some unit tests

client/json-rpc/src/async_client/client.rs
```diff
@@ -2,155 +2,64 @@
...
     pub retry: R,
 }
 
 impl<R: RetryStrategy> Client<R> {
     pub fn from_url<T: reqwest::IntoUrl>(server_url: T, retry: R) -> Result<Self, reqwest::Error> {
+        let reqwest_client = reqwest::ClientBuilder::new()
+            .use_native_tls()
+            .timeout(defaults::HTTP_REQUEST_TIMEOUT)
+            .build()
+            .expect("Unable to build Client.");
```
rgamba, line 30: This looks like an impl detail of `SimpleHttpClient` and this layer probably shouldn't know about it. Should we just expose a `SimpleHttpClient::new` method that takes care of this?

client/json-rpc/src/async_client/http_client.rs
```diff
@@ -0,0 +1,435 @@
...
+    fn update_state(&self, resp_state: State) -> bool {
+        self.state_manager.update_state(resp_state)
+    }
+
+    fn last_known_state(&self) -> Option<State> {
+        self.state_manager.last_known_state()
+    }
+}
+
+pub struct LBHttpClient {
```
rgamba, line 265: nit - `LB` might give the wrong impression that this impl does a load balancing between the given urls when in fact it broadcasts.

client/json-rpc/src/async_client/client.rs
ankushagarwal: Updated

client/json-rpc/src/async_client/http_client.rs
ankushagarwal: Updated

ankushagarwal:
diem naming rebase

ankushagarwal:
@xli Gentle ping for this PR review

client/json-rpc/src/async_client/http_client.rs
```diff
@@ -0,0 +1,450 @@
...
+                    Err(_) => Err(Error::ResponseTimeout(format!(
+                        "JSON RPC Timed out waiting for response from server {}",
+                        url
+                    ))),
+                    Ok(result) => result,
+                }
+            }
+        });
+        let mut results: Vec<Result<T, Error>> = join_all(futures).await.into_iter().collect();
+        if results.is_empty() {
```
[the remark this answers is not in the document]
ankushagarwal: Yes. And this should never happen because in the constructor we validate that 1 <= num_parallel_requests <= server_urls.len()

client/json-rpc/src/async_client/client.rs
```diff
@@ -2,155 +2,54 @@
...
+                "num_parallel_requests({}) should be <= length of server_urls({})",
+                num_parallel_requests,
+                server_urls.len()
+            ));
+        }
+        let server_urls = server_urls
+            .into_iter()
+            .map(|x| x.into_url())
+            .collect::<Result<Vec<_>, reqwest::Error>>()
+            .map_err(|x| format!("{}", x))?;
```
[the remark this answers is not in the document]
ankushagarwal: Updated

client/json-rpc/src/async_client/client.rs
```diff
@@ -2,155 +2,54 @@
...
     pub fn from_url<T: reqwest::IntoUrl>(server_url: T, retry: R) -> Result<Self, reqwest::Error> {
         Ok(Self {
-            http_client: reqwest::ClientBuilder::new()
-                .use_native_tls()
-                .timeout(defaults::HTTP_REQUEST_TIMEOUT)
-                .build()
-                .expect("Unable to build Client."),
-            server_url: server_url.into_url()?,
-            last_known_state: std::sync::RwLock::new(None),
+            http_client: Arc::new(SimpleHttpClient::new(server_url)),
```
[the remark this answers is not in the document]
ankushagarwal: Updated

client/json-rpc/src/async_client/http_client.rs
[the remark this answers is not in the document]
xli: I think we can safe remove this if block right?

client/json-rpc/src/async_client/http_client.rs
```diff
@@ -0,0 +1,468 @@
...
+        // Choose `num_parallel_requests` from `self.server_urls` to send requests to
+        let chosen_urls: Vec<reqwest::Url> = self
+            .urls
+            .choose_multiple(&mut rand::thread_rng(), self.num_parallel_requests)
+            .cloned()
+            .collect();
+        let futures = chosen_urls.iter().map(|url| {
+            let url = url.clone();
+            let req = req.clone();
+            let client = self.http_client.clone();
```
xli, line 334: should we clone? or just use the http client reference? 

The Client holds a connection pool internally, so it is advised that you create one and reuse it. see https://docs.rs/reqwest/0.10.9/reqwest/struct.Client.html

xli (approved):

client/json-rpc/src/async_client/http_client.rs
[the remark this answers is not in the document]
ankushagarwal: I would prefer leaving this since we can panic in `results.swap_remove(0)` if there is a bug in the constructor.

client/json-rpc/src/async_client/http_client.rs
ankushagarwal: request::Client is already a reference (Arc<ClientRef>), so cloning it is safe. Without cloning we run into issues because the reference is moved to `async move ` block.

PR #6806, opened 2020-12-04
depbot

rexhoffman:
## Motivation

Dependabot is not currently providing updates for indirect dependencies.
Docker dependabot was misconfigured and was erroring out. It only accepts directories with Dockerfiles in them, non-recursive.
Corrected.

Yes

## Test Plan

## Related PRs

Ran dependabot updates on master of my fork.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,24 +1,10 @@
 version: 2
 updates:
-  - package-ecosystem: "docker"
-    directory: "/docker/"
-    schedule:
-      interval: "daily"
-    labels:
-      - "docker"
-      - "dependencies"
-
-  - package-ecosystem: "github-actions"
-    directory: "/"
-    schedule:
-      interval: "daily"
-    labels:
-      - "github-actions"
-      - "dependencies"
-    open-pull-requests-limit: 15
-
   - package-ecosystem: "cargo"
     directory: "/"
+    allow:
+      # The undocumented default is to open prs for only direct dependencies.
+      - dependency-type: all
     schedule:
       interval: "daily"
     labels:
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -1,31 +1,124 @@
...
+    schedule:
+      interval: "daily"
+    labels:
+      - "cargo"
+      - "dependencies"
+    ignore:
+      - dependency-name: "codespan*"
+      - dependency-name: "libfuzzer-sys"
+      - dependency-name: "bindgen"
+    open-pull-requests-limit: 20
```
sausagee, line 16: Are the oncalls okay with bumping this up another 5?

.github/dependabot.yml
```diff
@@ -1,31 +1,124 @@
...
+
+  - package-ecosystem: "docker"
+    directory: "/docker/tools"
+    schedule:
+      interval: "daily"
+    labels:
+      - "docker"
+      - "dependencies"
+
+  - package-ecosystem: "docker"
```
sausagee, line 108: n00b question - what would this update do for the Dockerfile?

rexhoffman: PRs will look like this:   https://github.com/rexhoffman/libra/pull/242

.github/dependabot.yml
rexhoffman: That's not the interesting part, this is: 

https://github.com/libra/libra/pull/6806/files/3cf3a7187cc6745fc61aedd3d09c34f2cadb0cd9#diff-dd4fbda47e51f1e35defb9275a9cd9c212ecde0b870cba89ddaaae65c5f3cd28R6

It will result in at least a dozen new prs for our transitive dependencies.    There is no documentation to the effect, but the default behavior of dependabot v2 from github is to only update rust direct dependencies.   Dependabot V1 had different behavior, and security update support for rust.    V2 dependabot does not.

metajack (approved):

.github/dependabot.yml
sausagee: got cha. It'd be good to leave this note in the config since there is no documentation.

sausagee (approved):

.github/dependabot.yml
metajack: Testing

PR #6809, opened 2020-12-04
[move-lang] use copy instead of rename for tempfiles

mengxu-fb:
## Motivation

`rename(from, to)` will fail when `from` and `to` are on separate filesystems. This is usually not an issue for macOS but might be for certain Linux distributions where the `/tmp` is in tmpfs and others in `ext4` or sth.

The tempfile will be deleted when it goes out of scope (which is next line after the `copy`).

## Test Plan

`cargo xtest`

### DIFF
--- a/language/move-lang/src/lib.rs
+++ b/language/move-lang/src/lib.rs
@@ -349,7 +349,7 @@ pub fn generate_interface_files(
         if separate_by_hash && Path::new(&file_path).is_file() {
             continue;
         }
-        std::fs::rename(tmp.path(), file_path)?;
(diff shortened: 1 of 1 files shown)


tnowacki:
(Discussion had offline but bringing this in for posterity) 

We use `rename` because it is threadsafe. But it requires the /tmp dir to be in the same filesystem mount as the target file. We could do a `copy`+`rename`, where we first copy from the tmp to some random garbage in the target dir, and then rename from the random garbage to the target file. But that is probably overkill for now

PR #6839, opened 2020-12-05
[easy] a couple small cleanups

sunshowers:
Noticed these while looking at the smoke tests. Also remove `language/test.sh` because it's no longer used and its main functionality is now in `cargo xtest`.


davidiw (approved):
/land

PR #6849, opened 2020-12-08
making separate landblocking step for clippy to run only on tcb and d…

xvschneider:
## Motivation
Making separate landblocking step for clippy to run only on tcb and dependencies.
This will be useful because it will let us first fix security issues discovered by clippy in tcb, then enable the lint for TCB, without affecting non-tcb folders, thus making it easier to prioritize fixes .

Yes

## Test Plan

ran `cargo xclippy -p diem-secure-json-rpc -p diem-key-manager -p diem-secure-net -p diem-secure-push-metrics -p diem-secure-storage -p diem-secure-time -p safety-rules  -p execution-correctness -- -w clippy::integer_arithmetic
`

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -345,7 +345,7 @@ jobs:
           command: cargo xclippy --workspace --all-targets
       - run:
           name: cargo clippy tcb
-          command: cargo xclippy -p diem-secure-json-rpc -p diem-key-manager -p diem-secure-net -p diem-secure-push-metrics -p diem-secure-storage -p diem-secure-time -p safety-rules  -p execution-correctness
+          command: cargo xclippy -p safety-rules  -p execution-correctness
       - run:
           name: cargo fmt
           command: cargo xfmt --check
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -345,8 +345,7 @@ jobs:
           command: cargo xclippy --workspace --all-targets
       - run:
           name: cargo clippy tcb
-          command: |
-          cargo xclippy -p diem-secure-json-rpc -p diem-key-manager -p diem-secure-net -p diem-secure-push-metrics -p diem-secure-storage -p diem-secure-time -p safety-rules  -p execution-correctness
+          command: cargo xclippy -p diem-secure-json-rpc -p diem-key-manager -p diem-secure-net -p diem-secure-push-metrics -p diem-secure-storage -p diem-secure-time -p safety-rules  -p execution-correctness
       - run:
           name: cargo fmt
           command: cargo xfmt --check
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -344,6 +344,10 @@ jobs:
           name: cargo clippy
           command: cargo xclippy --workspace --all-targets
       - run:
+          name: cargo clippy tcb
+          command: |
+          cargo xclippy -p diem-secure-json-rpc -p diem-key-manager -p diem-secure-net -p diem-secure-push-metrics -p diem-secure-storage -p diem-secure-time -p safety-rules  -p execution-correctness
+      - run:
           name: cargo fmt
           command: cargo xfmt --check
       - run:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -345,7 +345,7 @@ jobs:
           command: cargo xclippy --workspace --all-targets
       - run:
           name: cargo clippy tcb
-          command: cargo xclippy -p safety-rules  -p execution-correctness
+          command: cargo xclippy -p safety-rules  -p execution-correctness -p diem-key-manager
       - run:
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -343,6 +343,9 @@ jobs:
       - run:
           name: cargo clippy
           command: cargo xclippy --workspace --all-targets
+      - run:
+          name: cargo clippy tcb
+          command: cargo xclippy -p diem-secure-json-rpc -p diem-key-manager -p diem-secure-net -p diem-secure-push-metrics -p diem-secure-storage -p diem-secure-time -p safety-rules  -p execution-correctness
```
mimoo, line 348: I think this would be clearer:

```
command: |
  cargo xclippy \
    -p diem-secure-json-rpc \
    -p diem-key-manager \
...
```

also wondering:

1) are there any lints that we can enable today? 
2) looking at @jnaulty 's diagram, it looks like none of these are land-blocking (only the bors github-action are)

sunshowers (changes requested):
Hmm, where did you get this list from? The TCBs are defined as `safety-rules` and `execution-correctness`, and their dependencies, only...

In general I think this should be managed by `cargo x`, it's a bit dangerous to have a separate definition of the TCB as another source of truth. Happy to talk tomorrow if you like 😊

davidiw:
> Hmm, where did you get this list from? The TCBs are defined as `safety-rules` and `execution-correctness`, and their dependencies, only, so in theory this should just be `-p safety-rules -p execution-correctness`.
> 
> For example, `diem-key-manager` is not part of the TCB to my knowledge.
> 
> In general I think this should be managed by `cargo x`, it's a bit dangerous to have a separate definition of the TCB as another source of truth. Happy to talk tomorrow if you like 😊

I agree, we should ideally maintain a single definition of TCBs, how would one invoke a TCB build via cargo x? That isn't particularly clear to me. Though I will definitely amplify the point that specifying dependencies of the TCB is unnecessary and may be a bit confusing.

key-manager is actually a tcb and maybe we should have done a better job promoting this. It is responsible for rotating the consensus public key and is the only process that has signing access to the operator's key -- which has the ability to perform key rotations for consensus among other actions (updating any fields within a validator config or even modifying an operator's account).

mimoo:
I suggest:

* A `README` in `secure/`
* A `secure/TCB.toml` listing the workplace crates that are root crates of the TCB. We could then parse it dynamically to do TCB-specific stuff like building, linting, etc.

sunshowers:
> how would one invoke a TCB build via cargo x

Yeah I should add that to x. Give me a bit.

> key-manager is actually a tcb

Ahh ok, do we want it to be a third TCB listed alongside LSR and LEC?

> A secure/TCB.toml listing the workplace crates that are root crates of the TCB

That's at least in principle what the list in x.toml is for, fwiw. The change still seems reasonable to do.

mimoo:
`x.toml` sounds more like a developer tooling, it'd be nice to have an authoritative `tcb.toml` in the `secure/` folder no? I don't really feel strongly about it though :D

xvschneider:
made a mistake while rebasing, will try again tomorrow

sunshowers:
OK once #6943 lands, you can say `cargo xclippy --subset lsr --subset lec` etc. Hope this helps :)

xli (changes requested):
this is not reviewable with so many changes that should not belong to this PR

xvschneider:
> argo xclippy --subset lsr --subset lec

so I should do:
`cargo xclippy --subset lsr --subset lec -- -W clippy::integer_arithmetic 
`
instead of
`cargo xclippy -p safety-rules  -p execution-correctness
`

?

xvschneider:
@xli  yes I'm aware of that, fixed the issue now. Thanks!

xli (approved):
Alright, I approve for removing myself from the blocker.

.circleci/config.yml
rexhoffman: CircleCi is always land blocking.

mimoo:
```
cargo xclippy --subset tcb -W clippy::integer_arithmetic
```

would be better imo, although I'm fine with the current approach as well and let @sunshowers and dev-infra figure out how to properly specify the TCB :)

(Also your current code misses key manager :o)

xvschneider:
@sunshowers do I still need to change anything right now? Thanks!

sunshowers (approved):
Thanks for your patience. I think you also want to add the `key-manager` crate, otherwise this looks fantastic.

I'll take care of switching this to `--members` in #6943.

xvschneider:
> Thanks for your patience. I think you also want to add the `key-manager` crate, otherwise this looks fantastic.
> 
> I'll take care of switching this to `--members` in #6943.

Thank you for reviewing!

PR #6852, opened 2020-12-08
[json-rpc] add chain id, latest ledger version and timestamp to http response header

xli:
## Motivation

These information is in JSON response body too, but when the request is batch request, client side need decode the array of the response first to get single response object for accessing the information. 
These information are useful to determine whether the server response is stale or not, use response header will be easier to decode and can ignore body if client found the ledger version is stale (client side tracks latest known version).

Yes

## Test Plan

Unit and integration test


libra-action:

Breaking changes policy:

2. Adding fields is ok.

thefallentree (approved):

PR #6860, opened 2020-12-08
[storage] get_latest_sequence_number with dedicated index

msmouse:
## Motivation

Add physical event by version index so that it's efficient to answer the question "what's the latest seq for this event at this version?", and it accommodates to the case where a same event stream emits multiple events in the same version.

Y

## Test Plan

unit tests

### DIFF
--- a/storage/diemdb/src/event_store/mod.rs
+++ b/storage/diemdb/src/event_store/mod.rs
@@ -12,7 +12,7 @@ use crate::{
     ledger_counters::{LedgerCounter, LedgerCounterBumps},
     schema::{
         event::EventSchema, event_accumulator::EventAccumulatorSchema,
-        event_by_key::EventByKeySchema,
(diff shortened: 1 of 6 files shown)


msmouse:
* Justification and breaking nature (who does it affect? validators, full nodes, tooling, operators, AOS, etc.)

Added one column family so the DB needs to be wiped, on both validator and fullnode nodes.

* Comprehensive test results that demonstrate the fix working and not breaking existing workflows.

existing and added unit tests at DiemDB level.

* Why we must have it for V1 launch.

This is a more robust version of querying the latest sequence number for an event stream at a specific version. The old one uses binary search, was not well tested.

* What workarounds and alternative we have if we do not push the PR.

The binary search version probably works in reality as well.. (A known issue was it will return wrong answer if multiple events under the same key are emitted by the same transaction version, but it's not the case in reality.)

sausagee (approved):

PR #6873, opened 2020-12-09
fix typo in README

rustielin:


sherry-x:
seems we already have a PR for this from the community  https://github.com/diem/diem/pull/6851

rustielin:
> seems we already have a PR for this from the community #6851

ah didn't catch that!

PR #6877, opened 2020-12-09
[network] Fix flaky test peer_send_message

phlip9:
I recently introduced a small race into the peer_send_message test where the server sends several messages, then the peer_handle drops and the server closes the connection before the client is able to fully read all of the messages.

Instead, the client shuts down the connection only after it has read all 30 messages.

### DIFF
--- a/network/src/peer/test.rs
+++ b/network/src/peer/test.rs
@@ -20,9 +20,14 @@ use bytes::Bytes;
 use diem_config::network_id::NetworkContext;
 use diem_network_address::NetworkAddress;
 use diem_types::PeerId;
-use futures::{future, io::AsyncWriteExt, stream::StreamExt, SinkExt};
+use futures::{
+    future,
+    io::{AsyncRead, AsyncWrite, AsyncWriteExt},
+    stream::StreamExt,
+    SinkExt,
+};
 use memsocket::MemorySocket;
-use netcore::transport::ConnectionOrigin;
+use netcore::{compat::IoCompat, transport::ConnectionOrigin};
 use std::str::FromStr;
 use tokio::runtime::{Handle, Runtime};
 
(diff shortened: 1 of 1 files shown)


network/src/peer/test.rs
```diff
@@ -108,6 +113,18 @@ fn build_test_connected_peers(
...
+    connection: &'a mut MemorySocket,
+) -> (
+    NetworkMessageSink<impl AsyncWrite + 'a>,
+    NetworkMessageStream<impl AsyncRead + 'a>,
+) {
+    let (read_half, write_half) = tokio::io::split(IoCompat::new(connection));
+    let sink = NetworkMessageSink::new(IoCompat::new(write_half), MAX_FRAME_SIZE);
+    let stream = NetworkMessageStream::new(IoCompat::new(read_half), MAX_FRAME_SIZE);
+    (sink, stream)
+}
```
gregnazario, line 126: I see so the issue is that you removed the framing of the messages off the connection?

gregnazario (approved):

network/src/peer/test.rs
phlip9: Ahh no this is just a convenience function so I can get both a sink and a stream. The problem has to do with the order of send, recv, and shutdown.

before: `join({ send all then request peer shutdown }, { recv all }, { run peer })`
after: `join({ send all }, { recv all then close connection }, { run peer })`

Note that the `run peer` will run until either: (1) the client shuts down the connection or (2) we request peer to shutdown when all the `peer_handle`s drop.

Before, we race between receiving all the messages and the peer shutting down. Sometimes the peer shuts down and closes the connection before the "client" finishes receiving all the messages.

It's kind of tricky to see because drops are implicit, but I changed `async move` to just `async` so we don't move the `peer_handle` into the `server` task, which would cause it to drop and request shutdown after sending all of the messages.

gregnazario: I feel like all the moves we have break a lot of the visibility for things like the tests.  Like for example having to send debug network messages in tests just to get information about pieces that were moved

phlip9: Yeah I agree, it's kind of annoying that we can't easily get a ref to e.g. the `Peer` actor after it gets consumed by `start(self)`... 

We could implement `Future` for `Peer` itself, but then we can't use async await in the "event loop" part. The `start(self)` pattern also let's us construct unnamed types easily which we don't have to put in the struct. Maybe we do a `async fn step(&mut self) -> bool` and `start(mut self) { while self.step().await {} }` so we can drive the actor ourselves in these unit tests? Not sure... would have to play around with it a bit to validate

Issue #6790, opened 2020-12-02
[network] Allow for known trusted peers in untrusted networks


gregnazario:
# Overview
### Types of Networks
Currently, there are two types of networks:
* Trusted networks, with mutual authentication (e.g. Validator network, both sides are known and authenticated)
* Untrusted networks, server only authentication (e.g. Public network, only the node being connected to authenticates)

### Trusted Peers Discovery and Connections
With a Trusted network, we retrieve trusted peers via On-Chain discovery in the `ConnectivityManager`.  This type of discovery is built upon the idea of the Peer-to-Peer(P2P) network or an upstream network.  Thus, when we discover the trusted peers, we try to connect to trusted peers with outbound dials, and include keeping track of inbound dials from trusted peers.

This has led to us adding an outbound connection limit on the `Connectivity Manager` for full nodes to ensure they have only 3 upstream connections.

### Untrusted Peers Discovery and Connections
With an Untrusted network, we retrieve the first peers to connect to by seed peers, and do not retrieve on-chain discovery data or use it for mutual authentication.  Therefore, all inbound connections on this network are treated equally for better or worse.

# Problem
We currently are building some work arounds for providing prioritized or reliable access to the network for preferred nodes.  Additionally, we're working on adding connection limits to prevent having too many nodes connected.  On public untrusted networks, we currently don't treat known trusted peers differently from any other peer, which means that we could block known trusted nodes from connecting to the network.  This could block purposes such as fallback to public network when a node has fallen far enough behind that it isn't allowed in the trusted set.

The main issues are:
* How do we prevent known peers from being blocked?
* How do we allow known peers to given reliable access?
* How do we retrieve the known peers information without automatically making outbound calls?

# Approaches
tl;dr to provide better access to some prioritized peers, we should be able to take advantage of the current "trusted peers" infrastructure

davidiw:
Support mutual and server-only auth at the same time -- we would upgrade to mutual auth if the incoming peer is a well known peer. Some code pointers:

https://github.com/libra/libra/blob/master/network/src/transport/mod.rs#L407

So in this case, we would make it such that HandshakeAuthMode should not be inferred

Then during the noise handshake, if a peer is within the trusted peer set, we use mutual auth and if it is not we use server-only, so this needs some tweaking: https://github.com/libra/libra/blob/master/network/src/noise/handshake.rs#L350

Finally a peer that is recognized as trusted must receive an appropriate designation so that is clear upstream: 
[...]

PR #6878, opened 2020-12-10
network: enable having trusted peers even in non-validator networks

bmwill:
Enable having "trusted peers" even in non-validator networks.

Ref #6790

### DIFF
--- a/network/src/noise/handshake.rs
+++ b/network/src/noise/handshake.rs
@@ -11,6 +11,7 @@
 //! [stream]: network::noise::stream
 
 use crate::noise::{error::NoiseHandshakeError, stream::NoiseStream};
+use crate::transport::TrustLevel;
 use diem_config::network_id::NetworkContext;
 use diem_crypto::{noise, x25519};
 use diem_infallible::{duration_since_epoch, RwLock};
@@ -182,7 +183,7 @@ impl NoiseUpgrader {
                     .await?
             }
             ConnectionOrigin::Inbound => {
-                let (socket, _peer_id) = self.upgrade_inbound(socket).await?;
+                let (socket, _peer_id, _) = self.upgrade_inbound(socket).await?;
                 socket
             }
         };
@@ -298,7 +299,7 @@ impl NoiseUpgrader {
     pub async fn upgrade_inbound<TSocket>(
         &self,
         mut socket: TSocket,
-    ) -> Result<(NoiseStream<TSocket>, PeerId), NoiseHandshakeError>
+    ) -> Result<(NoiseStream<TSocket>, PeerId, TrustLevel), NoiseHandshakeError>
     where
         TSocket: AsyncRead + AsyncWrite + Debug + Unpin,
     {
@@ -347,7 +348,7 @@ impl NoiseUpgrader {
             .map_err(|err| NoiseHandshakeError::ServerParseClient(remote_peer_short, err))?;
 
         // if mutual auth mode, verify the remote pubkey is in our set of trusted peers
-        if let Some(trusted_peers) = self.auth_mode.trusted_peers() {
+        let trust_level = if let Some(trusted_peers) = self.auth_mode.trusted_peers() {
             match trusted_peers.read().get(&remote_peer_id) {
                 Some(remote_pubkey_set) => {
                     if !remote_pubkey_set.contains(&remote_public_key) {
@@ -364,6 +365,7 @@ impl NoiseUpgrader {
                     ))
                 }
             };
+            TrustLevel::Trusted
         } else {
             // if not, verify that their peerid is constructed correctly from their public key
             let derived_remote_peer_id = PeerId::from_identity_public_key(remote_public_key);
@@ -374,7 +376,8 @@ impl NoiseUpgrader {
                     derived_remote_peer_id,
                 ));
             }
-        }
+            TrustLevel::Untrusted
+        };
 
         // if on a mutually authenticated network,
         // the payload should contain a u64 client timestamp
@@ -430,7 +433,7 @@ impl NoiseUpgrader {
             self.network_context,
             remote_peer_short,
         );
-        Ok((NoiseStream::new(socket, session), remote_peer_id))
+        Ok((NoiseStream::new(socket, session), remote_peer_id, trust_level))
     }
 }
 
@@ -513,7 +516,7 @@ mod test {
         server_public_key: x25519::PublicKey,
     ) -> (
         Result<NoiseStream<MemorySocket>, NoiseHandshakeError>,
-        Result<(NoiseStream<MemorySocket>, PeerId), NoiseHandshakeError>,
+        Result<(NoiseStream<MemorySocket>, PeerId, TrustLevel), NoiseHandshakeError>,
     ) {
         // create an in-memory socket for testing
         let (dialer_socket, listener_socket) = MemorySocket::new_pair();
@@ -584,7 +587,7 @@ mod test {
 
         let (client_res, server_res) = perform_handshake(&client, &server, server_public_key);
         let client_stream = client_res.unwrap();
-        let (server_stream, _) = server_res.unwrap();
+        let (server_stream, _, _) = server_res.unwrap();
 
         assert_eq!(client_stream.get_remote_static(), server_public_key);
         assert_eq!(server_stream.get_remote_static(), client_public_key);
--- a/network/src/noise/handshake.rs
+++ b/network/src/noise/handshake.rs
@@ -10,8 +10,10 @@
 //!
 //! [stream]: network::noise::stream
 
-use crate::noise::{error::NoiseHandshakeError, stream::NoiseStream};
-use crate::transport::TrustLevel;
+use crate::{
+    noise::{error::NoiseHandshakeError, stream::NoiseStream},
+    transport::TrustLevel,
+};
 use diem_config::network_id::NetworkContext;
 use diem_crypto::{noise, x25519};
 use diem_infallible::{duration_since_epoch, RwLock};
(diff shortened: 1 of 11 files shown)


network/src/noise/handshake.rs
```diff
@@ -93,7 +94,7 @@ pub enum HandshakeAuthMode {
     /// In `ServerOnly` mode, the dialer authenticates the server. However, the
     /// server does not care who connects to them and will allow inbound connections
     /// from any peer.
-    ServerOnly,
+    MaybeMutual(Arc<RwLock<HashMap<PeerId, HashSet<x25519::PublicKey>>>>),
```
gregnazario, line 97: hmm, do we need the anti-replay timestamps here as well?

network/src/transport/mod.rs
```diff
@@ -392,7 +404,7 @@ where
         base_transport: TTransport,
         network_context: Arc<NetworkContext>,
         identity_key: x25519::PrivateKey,
-        trusted_peers: Option<Arc<RwLock<HashMap<PeerId, HashSet<x25519::PublicKey>>>>>,
+        auth_mode: HandshakeAuthMode,
```
gregnazario, line 407: That's a lot better :P

network/src/noise/handshake.rs
```diff
@@ -93,7 +94,7 @@ pub enum HandshakeAuthMode {
     /// In `ServerOnly` mode, the dialer authenticates the server. However, the
```
gregnazario, line 94: Probably should update this comment

gregnazario (commented):
Looks mostly good as far as I can tell.  Were we expecting from the external usage that it should be fully Noise compatible with other implementations?

bmwill:
So this PR doesn't really change anything about the noise implementation, all its doing is allowing you to have a "trusted peers" list for the "MaybeMutual" case and allows you to mark connections as trusted if an inbound happens to be in the trusted peers set.

I haven't dug into it too much yet, but do you know what the trusted peers list is initialized to when starting up a full node?

gregnazario:
> So this PR doesn't really change anything about the noise implementation, all its doing is allowing you to have a "trusted peers" list for the "MaybeMutual" case and allows you to mark connections as trusted if an inbound happens to be in the trusted peers set.
> 
> I haven't dug into it too much yet, but do you know what the trusted peers list is initialized to when starting up a full node?

I believe it's initialized to the seed peers, but I'll have to double check

network/src/noise/handshake.rs
bmwill: Well given the comment on it for the Mutual auth case I'm not convinced its what we want on public networks

davidiw: I'd say they wouldn't be bad but are not P0. I think in order to take advantage of them, you would /only/ use them for mutual authentication in the maybe mutual path. Without looking at the code, I don't know how complicated that would be, but I can imagine messing it up and creating a really ugly bug. I'm curious from @mimoo who probably remembers better if this is something we should consider prioritizing...

network/src/noise/handshake.rs
```diff
@@ -347,34 +351,53 @@ impl NoiseUpgrader {
...
-                            hex::encode(remote_public_key.as_slice()),
-                        ));
+                            remote_peer_id,
+                        ))
+                    }
+                };
+                TrustLevel::Trusted
+            }
+            HandshakeAuthMode::MaybeMutual(trusted_peers) => {
+                match trusted_peers.read().get(&remote_peer_id) {
```
davidiw, line 375: should we consider a slight inversion here to minimize code:

```
match trusted_peers.read().get(&remote_peer_id) {
  Some(...) => {
    ...
  }
  None => {
    match &self.auth_mode {
      ...
    }
}
```

davidiw (approved):
fantastic!!!!!

mimoo:
Does it mean the network spec should be updated o.o?

davidiw:
> I haven't dug into it too much yet, but do you know what the trusted peers list is initialized to when starting up a full node?

Trusted peers would be defined by either those in the config: 
https://github.com/diem/diem/blob/master/config/src/config/network_config.rs#L77-L84

or via discovery methods (on-chain configs)

we have use cases for both (vasps for the former and reduction of vfn fail-over networks for the second)

network/src/noise/handshake.rs
davidiw: also @phlip9

davidiw:
> Does it mean the [network spec](https://github.com/diem/diem/blob/master/specifications/network/README.md) should be updated o.o?

I don't see this as a wire specification change but I can see where we define authentication modes, which should not be part of the specification. I think that should probably be stripped out and instead we should state a more primitive statement on expectations. 1) keys are exchanged via noise, 2) keys may have been exchanged apriori to establish trust -- then ideally another doc would have defined these expectations -- maybe we file an issue to track updates to the spec.

network/src/noise/handshake.rs
davidiw: Deleted my earlier comments, there are a few issues with timestamps and why they won't be that great for this case:
1) They have no real relationship with time, one could simply choose from 1 to infinity and the server will never verify they are within the current time window -- this could also have reduced or completely eliminated maintained state depending on the implementation
2) A maybe mutual auth already supports arbitrary peers pinging it, so even if we use replay prevention that would only prevent it on trusted peers as anyone can generate arbitrary keys (e.g., random bytes) to cause wasted DHs

So yeah, definitely not worth adding and maybe even worth fixing the comments

mimoo: If `trusted_peers` is always needed, it should probably be part of another data structure instead (like the state).
About the timestamp, the [network spec](https://github.com/diem/diem/blob/master/specifications/network/noise.md#replay-attacks) notes:

> We do not protect against this in the FN since an attacker can just create as many valid handshakes as they want.

network/src/noise/handshake.rs
```diff
@@ -348,35 +351,52 @@ impl NoiseUpgrader {
...
+                    Some(remote_pubkey_set) => {
+                        if !remote_pubkey_set.contains(&remote_public_key) {
+                            return Err(NoiseHandshakeError::UnauthenticatedClientPubkey(
```
mimoo, line 378: shouldn't you return `TrustLevel::Untrusted` instead here?

network/src/noise/handshake.rs
```diff
@@ -105,20 +105,23 @@ impl HandshakeAuthMode {
         }
     }
 
+    pub fn maybe_mutual(
+        trusted_peers: Arc<RwLock<HashMap<PeerId, HashSet<x25519::PublicKey>>>>,
+    ) -> Self {
+        HandshakeAuthMode::MaybeMutual(trusted_peers)
+    }
```
mimoo, line 112: what's the point of this function if you can write directly `HandshakeAuthMode::MaybeMutual(trusted_peers)` instead of `HandshakeAuthMode::maybe_mutual(trusted_peers)`?

network/src/noise/handshake.rs
davidiw: Great question, let's brainstorm on that briefly
1) We assume that the peer id is not derived in the mutual auth case, so we would need to add that check back in
2) If a peer id is derived but it assumes mutual auth and does not get it, they won't get an error and may assume they are privileged

I think the key question is in how we instruct folks to share their id, key pairs -- I suspect we'll ask them to use their VASP (whatever partner) account ids and share a public key

mimoo: I would have thought this would be more correct:

```rust
let is_trusted = if let Some(remote_pubkey_set) = trusted_peers.read().get(&remote_peer_id) {
  if remote_pubkey_set.contains(&remote_public_key) {
    TrustLevel::Trusted
  }
} else {
  TrustLevel::Untrusted
};

let derived_remote_peer_id =
    PeerId::from_identity_public_key(remote_public_key);
if derived_remote_peer_id != remote_peer_id {
    return Err(NoiseHandshakeError::ClientPeerIdMismatch(
        remote_peer_short,
        remote_peer_id,
        derived_remote_peer_id,
    ));
}

is_trusted
```

network/src/noise/handshake.rs
phlip9: We could only do anti-replay detection for trusted peers, even on the public network. Then we can put off garbage collection for another day : )

network/src/transport/mod.rs
```diff
@@ -96,6 +96,12 @@ impl ConnectionIdGenerator {
     }
 }
 
+#[derive(Clone, Debug, PartialEq, Serialize)]
+pub enum TrustLevel {
+    Trusted,
+    Untrusted,
+}
```
phlip9, line 102: Add a small comment explaining TrustLevel::Trusted ==> peer is in trusted_peers set

davidiw: I suspect TrustLevel may change to Role, but I haven't had a chance to discuss the concept in great depth with folks. In the ideal case, we would actually distinguish a random downstream peer from a priority upstream (a private VFN), upstream (a public VFN or validator), privileged downstream (VASP, DD) and then funneling that into the connection data so that applications can learn this information.

network/src/noise/handshake.rs
davidiw: I would actually argue that the current approach implemented is correct. If I have a known peer id slipping into an untrusted level -- it is weird. Also if a peer is untrusted, the get statement will fail resulting in the remote_pub_keyset being None.

trusted_peers.read().get(...)

would be None

so maybe "it's late" :P

network/src/noise/handshake.rs
phlip9: To clarify a bit, the threat here is a passive MitM that observes handshakes on the wire. Given DiemNet's current 1-connection 1-peer design, without replay attack detection, the MitM can evict honest peers' connections by replaying their handshakes. In lieu of more work implementing a more sophisticated `AntiReplayTimestamps` structure, we could just do anti replay detection for trusted peers.

Or maybe we bite the bullet and just do anti replay detection for everything. A simple approach might be for `AntiReplayTimestamps` to instead contain _two_ maps. One for trusted peers and one for untrusted peers. We also add a maximum capacity and just evict a random entry at capacity.

davidiw (commented):
To @mimoo's point, it might be useful to add appropriate tests that ensure that maybe mutual works for allowing untrusted and trusted peers to connect.

PR #6879, opened 2020-12-10
[async_client] Introduce protobuf to jsonrpc type conversions

rgamba:
## Motivation

There is currently no easy way to transform protobuf structs into the legacy rust jsonrpc types. This PR solves this by implementing `TryFrom` traits for all types.

## Test Plan

Added unit tests to cover all conversions.

### DIFF
--- a/json-rpc/types/src/conversion.rs
+++ b/json-rpc/types/src/conversion.rs
@@ -1,0 +1,581 @@
+// Copyright (c) The Diem Core Contributors
+// SPDX-License-Identifier: Apache-2.0
+
+use crate::{
+    proto::types as jsonrpc,
+    views::{
+        AccountRoleView, AccountStateProofView, AccountStateWithProofView, AccountView, AmountView,
+        BytesView, EventDataView, EventView, MetadataView, MoveAbortExplanationView, ScriptView,
+        TransactionDataView, TransactionView, VMStatusView,
+    },
+};
+use anyhow::{bail, Result};
+use diem_types::account_state::AccountState;
+use std::convert::{TryFrom, TryInto};
+
+impl TryFrom<jsonrpc::Event> for EventView {
+    type Error = anyhow::Error;
+
+    fn try_from(item: jsonrpc::Event) -> Result<EventView> {
+        Ok(EventView {
+            key: item.key.clone().into(),
+            sequence_number: item.sequence_number,
+            transaction_version: item.transaction_version,
+            data: match item.data {
+                Some(event_data) => event_data.try_into()?,
+                None => bail!("event.data is empty for event={:?}", item),
+            },
+        })
+    }
+}
+
+impl TryFrom<jsonrpc::EventData> for EventDataView {
+    type Error = anyhow::Error;
+
+    fn try_from(item: jsonrpc::EventData) -> Result<Self> {
+        match item.r#type.as_str() {
+            "burn" => Ok(Self::Burn {
+                amount: item.amount.unwrap().into(),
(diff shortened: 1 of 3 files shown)


libra-action:

Breaking changes policy:

2. Adding fields is ok.

ankushagarwal:
LGTM, Thanks!

xli (changes requested):
I don't think this is good path moving forward by providing this converter.  I understand why you need it, hence you may keep it in where use it.

rgamba:
> I don't think this is good path moving forward by providing this converter in diem/json-rpc. I understand why you need it, hence you may keep it in where use it.

Can you please elaborate? I do see value on having an easy conversion mechanism as part of the public api because they are ultimately mirror structs and given both struct sets are part of this crate, this seems the appropriate place to impl the conversion traits.

xli:
> > I don't think this is good path moving forward by providing this converter in diem/json-rpc. I understand why you need it, hence you may keep it in where use it.
> 
> Can you please elaborate? I do see value on having an easy conversion mechanism as part of the public api because they are ultimately mirror structs and given both struct sets are part of this crate, this seems the appropriate place to impl the conversion traits.

These view types should be removed eventually, instead of we maintain 2 types and convert them. the problem with converter is, when we need add a new field, the field will be added into the view type before we delete them, hence it also need update the converter, which makes the converter to be tight to a specific version of the code.

rgamba:
> hence it also need update the converter

I think that's a reasonable tradeoff for the convenience :) IMO as long as both coexist then the user will likely have the need to convert, unless they want to couple their domain code with protobuffer structs.

xli:
> > hence it also need update the converter
> 
> I think that's a reasonable tradeoff for the convenience :) IMO as long as both coexist then the user will likely have the need to convert, unless they want to couple their domain code with protobuffer structs.

Like I said, I understand why you want to do this, but it's just not right place for these, you are free to put these into where use them.

ankushagarwal:
The protobuf types are not rust friendly and require a lot of boilerplate code to be maintained by clients trying to use them. This code is agnostic to any usage of the client and hence it makes sense to stay here along with tests. Anybody else trying to write a rust client would have to go through the same ordeal and write a similar implementation.

As a pure rust project, we should provide first class support for rust-types to clients of the project rather than requiring everyone to write fragile boilerplate code.

I do understand that its a tradeoff to maintain it here, but the benefits significantly outweight the costs.

davidiw:
I think the point is that the Views are legacy code and will be stripped out sooner rather than later.

xli:
> The protobuf types are not rust friendly and require a lot of boilerplate code to be maintained by clients trying to use them. This code is agnostic to any usage of the client and hence it makes sense to stay here along with tests. Anybody else trying to write a rust client would have to go through the same ordeal and write a similar implementation.
> 
> As a pure rust project, we should provide first class support for rust-types to clients of the project rather than requiring everyone to write fragile boilerplate code.
> 
> I do understand that its a tradeoff to maintain it here, but the benefits significantly outweight the costs.

The point is I can't promise no breaking change if this is merged.
Once release 1.1 is settled, I will switch server code to use protobuf types, and move view types to where legacy code may use it and likely completely remove them. It can happen anytime in Jan, hence I don't see a reason to merge this change.

I understand you're not interested in other languages, but I don't think rust is special client application language because we implement server by Rust.
It may change when we have a maintainer can focus on Rust language client SDK development.
You are free to maintain a copy of rust view types for a specific version of JSON-RPC API response, which I don't recommend.

PR #6881, opened 2020-12-10
[backup] backup events with transactions

msmouse:
## Motivation

Events as part of the "ledger history" should accompany the transactions in a transaction backup.

Y

## Test Plan

unit tests. smoke tests.

### DIFF
--- a/storage/backup/backup-cli/src/backup_types/tests.rs
+++ b/storage/backup/backup-cli/src/backup_types/tests.rs
@@ -154,10 +154,20 @@ fn test_end_to_end_impl(d: TestData) {
     )
     .unwrap();
     assert_eq!(
-        d.db.get_transactions(d.txn_start_ver, num_txns_to_backup, d.target_ver, false)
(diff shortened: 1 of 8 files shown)


msmouse:
* Justification and breaking nature (who does it affect? validators, full nodes, tooling, operators, AOS, etc.)

Backups needs to be re-made since the binary format has changed to include events in a transaction backup.

* Comprehensive test results that demonstrate the fix working and not breaking existing workflows.

unit tests and smoke tests.

* Why we must have it for V1 launch.

This is to make sure in V1 every node have all events available for querying, even if it's recovered from a backup. This was the consensus out of a discussion with @dmitri-perelman and folks.

* What workarounds and alternative we have if we do not push the PR.

The AOS requires the nodes to have at least a few days of events when it had issues and was behind. When it does happen and unfortunately the fullnode it talks to was recovered from a backup recently, it needs to talk to another fullnode that has the history.

sausagee (approved):

PR #6891, opened 2020-12-10
[vm] Revalidate txns

runtian-zhou:
## Motivation

Now that #6918 is landed, this refactor will no longer be a breaking change. 

Yes

## Test Plan

The first commit changes the gas charging and thus changes the golden files. The second one should only change one golden file.

## Related PRs
#6780

### DIFF
--- a/language/diem-vm/src/diem_transaction_executor.rs
+++ b/language/diem-vm/src/diem_transaction_executor.rs
@@ -4,8 +4,9 @@
 use crate::{
     counters::*,
     data_cache::StateViewCache,
+    diem_transaction_validator::validate_signature_checked_transaction,
     diem_vm::{
-        charge_global_write_gas_usage, get_currency_info, get_transaction_output,
+        charge_global_write_gas_usage, get_transaction_output,
         txn_effects_to_writeset_and_events_cached, DiemVMImpl, DiemVMInternals,
     },
     errors::expect_only_successful_execution,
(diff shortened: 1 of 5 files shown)


language/diem-vm/src/diem_vm.rs
```diff
@@ -516,27 +515,6 @@ pub fn txn_effects_to_writeset_and_events_cached<C: AccessPathCache>(
     Ok((ws, events))
 }
 
-pub(crate) fn charge_global_write_gas_usage<R: RemoteCache>(
```
tzakian, line 519: I'm not sure if I quite follow why this is going away in this PR? My understanding of the change was that it should only be fixing the logic in `session.num_mutated_accounts` to only return the number of accounts that were mutated, and not include the read/non-mutated accounts in the session?

language/diem-vm/src/diem_transaction_executor.rs
```diff
@@ -198,8 +185,6 @@ impl DiemVM {
                 )
                 .map_err(|e| e.into_vm_status())?;
 
-            charge_global_write_gas_usage(cost_strategy, &session)?;
```
tzakian, line 201: I think this should stay no? We still want to keep charging for global writes is my understanding?

bob-wilson (commented):
I've reviewed the 2nd commit and it looks good. I feel like I shouldn't be the one to approve it, though, since it is just my change from #6780. The first commit is going in separately as #6897

language/diem-vm/src/diem_vm.rs
```diff
@@ -37,7 +37,6 @@ use move_vm_types::{
     values::Value,
 };
 use std::{convert::TryFrom, sync::Arc};
-use vm::errors::Location;
```
bob-wilson, line 40: I don't know where this change came from. It wasn't in my #6780 and regardless it doesn't work since there is still a reference to Location::Undefined in this file.

bob-wilson:
I tested this locally and confirmed that it is no longer a breaking change.

@runtian-zhou Can you please edit the git commit message to remove the 2nd paragraph about the gas issues?

bob-wilson (approved):
Thanks, Runtian. If you've reviewed this and think it looks OK to you, then I'm definitely fine with it.

runtian-zhou:
Looks good to me. It's nice for us to refactor out the common pieces of the code to prevent errors like missing checks.

PR #6896, opened 2020-12-11
[async_client] Add custom retry strategy per request

rgamba:
## Motivation

This PR introduces the option of passing in a custom retry strategy on a per-request basis rather than having to use the retry strategy used at time of instantiation. This allows clients to leverage the client's retry mechanism while having the flexibility to use different strategies when needed.

## Test Plan

Current tests cover all the code paths introduced by this PR.

### DIFF
--- a/client/json-rpc/src/async_client/client.rs
+++ b/client/json-rpc/src/async_client/client.rs
@@ -206,14 +206,16 @@ impl<R: RetryStrategy> Client<R> {
     }
 
     pub async fn send<T: DeserializeOwned>(&self, request: Request) -> Result<Response<T>, Error> {
-        self.send_with_retry(&request).await?.try_into()
+        self.send_with_retry(&request, &self.retry)
+            .await?
+            .try_into()
     }
 
     pub async fn send_opt<T: DeserializeOwned>(
         &self,
         request: Request,
     ) -> Result<Response<Option<T>>, Error> {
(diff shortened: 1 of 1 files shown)


client/json-rpc/src/async_client/client.rs
```diff
@@ -225,12 +225,20 @@ impl<R: RetryStrategy> Client<R> {
     }
 
     pub async fn send_with_retry(&self, request: &Request) -> Result<JsonRpcResponse, Error> {
```
xli, line 227: add retry: &RS, to this method param instead of creating new one.

client/json-rpc/src/async_client/client.rs
```diff
@@ -279,29 +287,43 @@ impl<R: RetryStrategy> Client<R> {
     pub async fn batch_send_with_retry(
         &self,
         requests: &[Request],
```
xli, line 289: add retry: &RS, to this fn instead of creating new one

rgamba: Sure. Do we need to bump any major semantic version since it'll be a breaking change?

rgamba: Synced offline. No users depend on this so it's fine to update. Done

xli (approved):

PR #6897, opened 2020-12-11
[breaking][vm] Fix global write charge

runtian-zhou:
## Motivation

This PR fixes the gas charing logic for number of account mutated in the transaction. The previous logic will overcharge transaction. This will unfortunately be a breaking change as it changes the default gas charging logic.

Yes
## Test Plan
Added a new test to make sure read-only accounts will not be charged.

## Related PRs

#6891


runtian-zhou:
On the other hand this PR may have a drawback of undercharging gas price: In the current schema, we get total # of mutated accounts before epilogue. Thus if the Account is unchanged in the previous transaction we will not charge that piece of write.

language/move-vm/integration-tests/src/tests/mutated_accounts_tests.rs
```diff
@@ -0,0 +1,131 @@
...
+                Some((layout, value)) => {
+                    changes.publish_resource(addr, ty, value.simple_serialize(&layout).unwrap())
+                }
+                None => changes.unpublish_resource(addr, ty),
+            }
+            .unwrap();
+        }
+    }
+    changes
+}
```
vgao1996, line 35: There's already a similar function at `move-vm/test-utils/src/misc.rs`.

language/move-vm/runtime/src/data_cache.rs
```diff
@@ -158,7 +158,13 @@ impl<'r, 'l, R: RemoteCache> TransactionDataCache<'r, 'l, R> {
...
     pub(crate) fn num_mutated_accounts(&self) -> u64 {
-        self.account_map.keys().len() as u64
+        let mut total_mutate_accounts: u64 = 0;
+        for (_, entry) in self.account_map.iter() {
+            if entry.data_map.values().any(|(_, v)| v.is_mutated()) {
+                total_mutate_accounts += 1;
+            }
+        }
+        total_mutate_accounts
     }
```
vgao1996, line 168: To me, `num_mutated_accounts` has always been a relatively weird API to expose from the VM/Session, but I understand there's some tricky logic we need to deal with at the call site...

@runtian-zhou could you remind me why we don't calculate this from the transaction effects? Is it because we need to know the number before going into the epilogue, where gas distribution is handled?

vgao1996 (commented):
Overall LGTM. Are we also charging for the number of accounts read? In theory that could also be an expensive operation in case of a cache miss.

dariorussi:
let's talk some more about this and we may not have this ready by today, which sucks but better put in the right fix than a broken one.
@runtian-zhou makes a very valid point that the epilogue is guaranteed to change the sender account and so that should be accounted for the transactions changed. The way this is now, if the sender is untouched before the prologue it would not contribute to the set of written accounts. That is bogus and it does not make sense to push changes that do not do the right thing. To be honest it feels more correct to overcharge at that point.
We will look at this today and we will try to make sense of it but given important parties are on PTO this may go to next week.
Sorry. Also important to point out that - I think,  but could be wrong -  this breaking change is unlikely to break anything for now and at the beginning. In that respect, as bad as it is to have a breaking change coming later it should be fine

bob-wilson:
It seems to me like we should pass the sender's account as an argument to num_mutated_accounts so that it will always be included, even if it wasn't already modified. That seems like it ought to be straightforward.

language/move-vm/integration-tests/src/tests/mutated_accounts_tests.rs
bob-wilson: In the revised version of this PR in #6918, I changed to use the existing function.

bob-wilson:
See #6918 where I implemented my proposed modification to account for epilogue changes to the sender's account.

bob-wilson:
The revised version of this landed in #6918

PR #6917, opened 2020-12-12
[ci] switch build-dev to use xcheck rather than xbuild + more improvements

sunshowers:
The main goal here is to cut down on build-dev times.

Previously we were doing a full check (without --all-targets), then checks on a few packages. Do this in a more principled manner:
* a check of all production code
* a check of all code in the workspace

Also refactor a bunch of code to be shared across build, check, test and a couple other commands.

### DIFF
--- a/devtools/x/src/cargo.rs
+++ b/devtools/x/src/cargo.rs
@@ -2,25 +2,23 @@
 // SPDX-License-Identifier: Apache-2.0
 
 use crate::{
-    changed_since::changed_since_impl,
+    cargo::selected_package::{SelectedInclude, SelectedPackages},
     config::CargoConfig,
-    context::XContext,
     utils::{apply_sccache_if_possible, project_root},
     Result,
 };
 use anyhow::anyhow;
-use guppy::graph::DependencyDirection;
 use indexmap::map::IndexMap;
 use log::{info, warn};
 use std::{
-    collections::BTreeSet,
     env,
     ffi::{OsStr, OsString},
     path::Path,
     process::{Command, Output, Stdio},
     time::Instant,
 };
-use structopt::StructOpt;
+
+pub mod selected_package;
 
 const RUST_TOOLCHAIN_VERSION: &str = include_str!("../../../rust-toolchain");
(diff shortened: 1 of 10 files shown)


sunshowers:
[Before](https://app.circleci.com/pipelines/github/diem/diem/35914/workflows/08e3c1cf-0bcf-4a30-8a68-f93fadf71ce5/jobs/264971)

[After](https://app.circleci.com/pipelines/github/diem/diem/35916/workflows/401e6627-3d9f-46df-8d74-4454fcf921c2/jobs/264990)

Looks like times are down from ~16 minutes to ~6 minutes.

devtools/x/src/cargo.rs
```diff
@@ -315,23 +317,22 @@ impl<'a> CargoCommand<'a> {
     pub fn cargo_config(&self) -> &CargoConfig {
         match self {
             CargoCommand::Bench { cargo_config, .. } => cargo_config,
-            CargoCommand::Check(config) => config,
```
rexhoffman, line 318: appreciate standardization of the commands...

rexhoffman (approved):
Looks really good.   Appreciate the refactoring to standardize command behavior/pass throughs.

Only concern is we might miss feature dependence on some bin's build when run separately, rather than as part of workspace xbuild/xcheck --production.

rexhoffman:
Hmmm also wonder if by not fully exercising the compiler we're opening our selves up to nasty surprises with compilation/linking especially with rocksdb.

sunshowers:
> Only concern is we might miss feature dependence on some bin's build when run separately, rather than as part of workspace xbuild/xcheck --production.

Well we weren't doing a good job before either I think, given that we weren't testing building just our production binaries at all (just a few test-only binaries that matter less). So this is an improvement. Besides, hopefully my upcoming workspace-hack work will make this better.

> Hmmm also wonder if by not fully exercising the compiler we're opening our selves up to nasty surprises with compilation/linking especially with rocksdb. Ehh, tests should catch problems here.

Yeah, we do go through linking etc with tests.

PR #6918, opened 2020-12-12
[breaking][vm] Fix global write charge

bob-wilson:
This is a revised version of #6897 to address feedback from the review.

## Motivation

This PR fixes the gas charging logic for the number of accounts mutated in a transaction. The previous logic will overcharge because it fails to distinguish accounts that are only read and not modified. This will unfortunately be a breaking change as it changes the default gas charging logic.

Yes

## Test Plan

Added a new test to make sure read-only accounts will not be charged.

## Related PRs

#6897
#6891


tzakian (approved):
This LGTM!

PR #6928, opened 2020-12-14
[infallible] Reexport lock guard types

phlip9:
Minor code cleanliness change. Reexport lock guard types `MutexGuard`,
`RwLockReadGuard`, and `RwLockWriteGuard` through `diem-infallible` so
you can do:

```rust
use diem_infallible::{Mutex, MutexGuard};
```

instead of:

```rust
use diem_infallible::Mutex;
use std::sync::MutexGuard;
```


bmwill (approved):
/land

PR #6929, opened 2020-12-14
[time-service] Add TimeService for mocking async time

phlip9:
### Motivation

TL;DR: got fed up with flaky tests for Peer rpc timeouts... wrote this as an experiment to mock async time.

Testing time-related things in network is kind of a pain. We get around this problem for interval-like things using a `ticker` mpsc channel, but that's kind of a hack and doesn't cover sleeping, timeouts, or querying the clock.

This problem is not unique to network; two other attempts exist in secure/time and consensus. The first is not async-aware, but works great for normal threading-only code.

Compared to the consensus time service, this one returns concrete `Future`s, which integrate more cleanly with the async ecosystem. The consensus time service also mandates spawning every sleep as a task, which sucks when we're setting/registering/polling timers on every rpc and socket read/write. It's also more convenient for the socket rate limiting work we're doing, which needs manual `AsyncRead`/`AsyncWrite` poll implementations.

This implementation also comes as a convenient enum so we don't have to infect literally every struct in network/ with a `T: TimeService` trait bound. Though perf is not really a goal here, it should be noted that the interfaces here are zero cost when we're running without testing features.

With this mock time service, I think it's actually not too far of a step to mock out the whole runtime so we can start fuzzing the full execution trace deterministically.

As an example, here's a test using the `MockTimeService`:

```rust
#[test]
fn peer_recv_rpc_timeout() {
    let mut rt = Runtime::new().unwrap();
    let time = MockTimeService::new();
    let /* .. */ = build_test_peer(rt.handle().clone(), time.clone().into(), ConnectionOrigin::Inbound);
    let (mut client_sink, mut _client_stream) = build_network_sink_stream(&mut connection);

    let send_msg = NetworkMessage::RpcRequest(RpcRequest { /* .. */ });
    let recv_msg = PeerNotification::RecvRpc(InboundRpcRequest { /* .. */ });

    let test = async move {
        // Client sends the rpc request.
        client_sink.send(&send_msg).await.unwrap();

        // Server receives the rpc request from client.
        let received = peer_notifs_rx.next().await.unwrap();
        assert_eq!(received, recv_msg);

        // Pull out the request completion handle.
        let mut res_tx = match received {
            PeerNotification::RecvRpc(req) => req.res_tx,
            _ => panic!("Unexpected PeerNotification: {:?}", received),
        };

        // The rpc response channel should still be open since we haven't timed out yet.
        assert!(!res_tx.is_canceled());

        // Advance time to timeout the inbound rpc.
        time.advance(INBOUND_RPC_TIMEOUT_MS).await;

        // The rpc response channel should be canceled from the timeout.
        assert!(res_tx.is_canceled());
        res_tx.cancellation().await;

        // Client then closes connection.
        client_sink.close().await.unwrap();
    };
    rt.block_on(future::join(peer.start(), test));
}
```

### DIFF
--- a/common/time-service/src/mock.rs
+++ b/common/time-service/src/mock.rs
@@ -1,0 +1,467 @@
+// Copyright (c) The Diem Core Contributors
+// SPDX-License-Identifier: Apache-2.0
+
+use crate::{Sleep, SleepTrait, TimeServiceTrait, ZERO_DURATION};
+use diem_infallible::Mutex;
+use futures::future::Future;
+use std::{
+    cmp::max,
+    collections::hash_map::{DefaultHasher, HashMap},
+    fmt::Debug,
+    hash::BuildHasherDefault,
+    pin::Pin,
+    sync::{Arc, MutexGuard},
+    task::{Context, Poll, Waker},
+    time::Duration,
+};
+
+// TODO(philiphayes): use more robust solution that'll let us control the seed.
+type DeterministicHasher = BuildHasherDefault<DefaultHasher>;
+
+type SleepIndex = usize;
+
+/// A [`TimeService`] that simulates time and allows for fine-grained control over
+/// advancing time and waking up sleeping tasks.
+#[derive(Clone, Debug)]
+pub struct MockTimeService {
+    inner: Arc<Mutex<Inner>>,
+}
+
+#[derive(Debug)]
+struct Inner {
+    /// The current simulated time.
+    now: Duration,
+    /// The next free index in the `pending` queue that we can allocate to a new waiter.
+    next_sleep_index: SleepIndex,
+    /// A queue of pending `SleepEntry`s, each with a deadline for when we should
+    /// wake them.
+    ///
+    /// This `HashMap` also uses a hasher with constant seed (in the future the
+    /// seed should be configurable) to help improve test determinism.
+    pending: HashMap<SleepIndex, SleepEntry, DeterministicHasher>,
+}
+
+/// A timer entry owned by the `MockTimeService` in the `pending` queue.
+///
+/// These are indexed by `SleepIndex`, which you get when you `register_sleep` a
+/// new waiter.
+#[derive(Debug)]
+struct SleepEntry {
+    /// The time when this sleep task expires and should be woken up.
+    deadline: Duration,
+    /// When the corresponding `MockSleep` future polls and isn't expired
+    /// yet, it will register its `Waker` here for the `MockTimeService`
+    maybe_waker: Option<Waker>,
+}
+
+/// A [`Future`] that resolves when the simulated time in the [`MockTimeService`]
+/// advances past the deadline. When these are pending, they have a `SleepEntry`
+/// in the `MockTimeService`'s `pending` queue.
+#[derive(Debug)]
+pub struct MockSleep {
+    /// A handle to the `MockTimeService` so we can get the underlying `SleepEntry`.
+    time_service: MockTimeService,
+    /// The index of this `MockSleep`'s `SleepEntry` in the `MockTimeService`'s
+    /// `pending` queue. If the `pending` queue doesn't contain our entry, then
+    /// this sleep must be completed.
+    entry_index: SleepIndex,
+}
+
+/////////////////////
+// MockTimeService //
+/////////////////////
+
+impl MockTimeService {
+    pub fn new() -> Self {
+        Self {
+            inner: Arc::new(Mutex::new(Inner {
+                now: ZERO_DURATION,
+                next_sleep_index: 0,
+                pending: HashMap::default(),
+            })),
+        }
+    }
+
+    fn lock(&self) -> MutexGuard<'_, Inner> {
+        self.inner.lock()
+    }
+
+    /// Return the number of pending `Sleep` waiters.
+    pub fn num_waiters(&self) -> usize {
+        self.lock().pending.len()
+    }
+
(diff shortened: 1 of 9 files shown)


davidiw (commented):
This is critical for us building a simulation friendly library!

One question before doing a more serious review -- is there anyway to make this crate supersede secure/time such that async and/or tokio is built on top?

phlip9:
> One question before doing a more serious review -- is there anyway to make this crate supersede secure/time such that async and/or tokio is built on top?

My understanding of secure/time is that it assumes only one thread right? So in secure/time, when we call `MockTimeService::sleep` we can just increment the current time because our thread is the only one running and the only one that can observe the time change.

Here, we're managing time across many concurrent tasks. You could get the equivalent functionality like in secure/time with:

```rust
impl TimeServiceTrait {
    fn sleep_blocking(&self, duration: Duration) {
        let delay = self.sleep(duration);
        futures::executor::block_on(delay);
    }
}
```

Compared to secure/time, this would require some other thread calling `time.advance`. This is why consensus mock time service has an "auto advance with limit" option, but I haven't added that yet (this PR was already getting big : p). The auto advance just automatically completes _and_ advances time on a sleep (while time is under some limit). That way, the above `sleep_blocking` would work exactly the same as `secure/time` does today.

rexhoffman:
Conceptually far better than what we're currently doing.

I will throw out there though, that what we'll want to eventually have is an api that models the behavior of a network of nodes, not just controls the clock....   ideally this clock would only be used to drive state changes in transaction (awaiting -> timed out).... which would have a better api represented as something like:

let cluster = cluster::Builder().hasSlowNode("x").build();
cluster.accept(message).timeOutOnNode("x").proceedOnNodes().assertQuorumContains(message).assertNode.....

But you get the idea...  we effective want a queryable monad as our api and when executing in process we should use direct state manipulations ignoring the clock, or only moving it forward to keep the transition logic happy....

Each step in the monad/builder above could be a separate struct with separate capabilities, but after describing the state you want the app to transition to you should end up back at the cluster type and be able to query it's state.... did the transaction reach quorum, did the lagging leader get evicted from leadership?  etc....

Hopefully most tests doent

phlip9:
> I will throw out there though, that what we'll want to eventually have is an api that models the behavior of a network of nodes, not just controls the clock.... ideally this clock would only be used to drive state changes in transaction (awaiting -> timed out)....
>
> But you get the idea... we effective want a queryable monad as our api and when executing in process we should use direct state manipulations ignoring the clock, or only moving it forward to keep the transition logic happy....
>
> Each step in the monad/builder above could be a separate struct with separate capabilities, but after describing the state you want the app to transition to you should end up back at the cluster type and be able to query it's state.... did the transaction reach quorum, did the lagging leader get evicted from leadership? etc....
>
> Hopefully most tests don't need to access the clock, aside from the transition applied by the monad auto updating the clock for them. But one step at a time.

Oh definitely, that would be super useful! A few points here:

1. In the short term, this time service is more about increasing test determinism. Along those lines, the next source of non-determinism is the tokio scheduler itself; mocking the scheduler would allow us to execute most of network's tests deterministically.
2. Of course, a time service is also useful for full cluster simulation. It would make sense to me that each virtual node gets its own mock runtime (including a mock time service) and the cluster harness drives each virtual node according to the test profile.
3. I think the big thing missing that's blocking your proposal is actually full message intermediation hooked into the network layer. Consensus has a `NetworkPlayground` which does this, but at a layer above network. In this case, the cluster harness would probably init each virtual node with a mock transport. The mock transport would intermediate new connections through the cluster harness, who can now control the full message flow and network topology. Some API validating constraints probably include: how to support twins (how do we id nodes in the harness)? do we support json-rpc? do we serialize/encrypt messages across the wire? can we ensure determinism? etc...

common/time-service/src/tokio.rs
```diff
@@ -0,0 +1,42 @@
...
+
+    /// See [`tokio::time::delay_for`]
+    fn sleep(&self, duration: Duration) -> Sleep {
+        tokio::time::delay_for(duration).into()
+    }
+}
+
+impl SleepTrait for TokioSleep {
+    fn is_elapsed(&self) -> bool {
+        self.is_elapsed()
```
davidiw, line 35: recursive function? or is this calling the TokioSleep(self)::is_elapsed?

common/time-service/src/mock.rs
```diff
@@ -0,0 +1,467 @@
...
+// MockTimeService //
+/////////////////////
+
+impl MockTimeService {
+    pub fn new() -> Self {
+        Self {
+            inner: Arc::new(Mutex::new(Inner {
+                now: ZERO_DURATION,
+                next_sleep_index: 0,
+                pending: HashMap::default(),
```
davidiw, line 80: BTreeMap is probably more performant here or a BinaryTree either as a min heap -- from experience playing with simulators...

davidiw (approved):
Yeah, this kinda sucks that we cannot write this in a Tokio free fashion without over-engineering...

common/time-service/src/tokio.rs
phlip9: Ahh, this calls `TokioSleep::elapsed(self)`. I'll make that clearer

common/time-service/src/mock.rs
phlip9: Right, I think I can do `BTreepMap<(Duration, SleepIndex), Option<Waker>>` instead of this `HashMap<SleepIndex, SleepEntry>`.

Issue #6800, opened 2020-12-03
[network] Refactor and consolidate per-peer actors (Peer, Rpc, DirectSend, NetworkProvider)


phlip9:
### Overview

The [Peer](https://github.com/libra/libra/blob/100662e20f13aefaafb3a9be679166fa366ded2e/network/src/peer/mod.rs) actor (and associated actors [Rpc](https://github.com/libra/libra/blob/100662e20f13aefaafb3a9be679166fa366ded2e/network/src/protocols/rpc/mod.rs), [DirectSend](https://github.com/libra/libra/blob/100662e20f13aefaafb3a9be679166fa366ded2e/network/src/protocols/direct_send/mod.rs), [NetworkProvider](https://github.com/libra/libra/blob/100662e20f13aefaafb3a9be679166fa366ded2e/network/src/interface/mod.rs)) are getting a bit unwieldy and can be simplified and consolidated.

### Current State

1. There are too many actors, too many message types, too much actor communication.
2. We currently spawn 11 tasks per peer connection! (2 (Peer) + 1 (DS) + 1 (RPC) + 7 (NetworkProvider)). With this refactor we can reduce that down to 1.
3. DirectSend actor does almost no useful work.
4. Shutdown is complicated by having to manage proper shutdown of 11 concurrent tasks.
5. NetworkProvider actor does no useful work and can be removed once we consolidate to 1 task.

### Goals

1. Consolidate Peer + DS + RPC into 1 Future (than can be spawned as 1 task).
2. Remove NetworkProvider and the intermediate queues/channels between Peer and PeerManager
3. Simplify the main Peer event loop; all async stuff happens as Futures polled across a `futures::select`, inside each branch is just normal blocking code.
4. Simplify the shutdown process.

### Incremental Steps

1. Move raw socket logic out; centralize framing and IO logic: #6738
2. Fold DirectSend logic into Peer. Remove DirectSend task(s) and channels: #6807
3. Split out InboundRpcs into a field owned by Peer: #6939 
4. Refactor OutboundRpcs into a field owned by Peer. Remove task(s) and channels in NetworkProvider: #7225 
5. Remove intermediate tasks and channels in NetworkProvider. Remove NetworkProvider: #7233 
6. Remove async from `futures::select` branches.
7. Refactor separate writer task into a Future that gets driven in the event loop.

### End State Sketch

[...]

PR #6939, opened 2020-12-14
[network] Split out InboundRpcs into a field owned by Peer

phlip9:
See #6800

This PR completes half of the planned refactoring for Peer + Rpc by driving the inbound rpc handling from the Peer task instead of a separate Rpc task.

I tried to preserve the counters and logs as much as possible, though I'd like to eventually clean them up a bit.

This PR also removes the inbound rpc fuzzer. Back when we used yamux, each inbound rpc would read and deserialize the rpc message off the substream, i.e., deserializing untrusted input. Consequently, it was much more important to fuzz the rpc handling directly. Today, the rpc logic is really just shuffling messages around and updating counters. Finally, `peer::fuzzing` actually covers everything `rpc::fuzzing` did, in addition to fuzzing the wire protocol.


network/src/peer/mod.rs
```diff
@@ -420,7 +445,23 @@ where
...
+                        error = %err,
+                        "{} Error handling inbound rpc request: {}",
+                        self.network_context,
+                        err
+                    );
+                }
+            }
+            NetworkMessage::RpcResponse(_) => {
                 let notif = PeerNotification::NewMessage(message);
                 self.rpc_notifs_tx.send(notif).await.map_err(|err| {
```
gregnazario, line 466: Interesting that we used to notify new inbound RPC responses even on requests...  Either that or we just had a poorly commented notification channel.

gregnazario (approved):
Is moving OutboundRPCs next?

PR #6941, opened 2020-12-14
[network] Add AsyncRead Byte Rate Limiting

gregnazario:
### Overview
This adds inbound and outbound rate limiting at a byte throughput controlled level.  Backpressure will occur when the reader is rate limiting *and* the `TcpSocket` buffer has filled.  Therefore, rate limit throughputs of reading on the application side will be the expected throughput, but the actual wire read throughput would be maximum `bufferSize` over the rate limit.  This would then even out as if the rate limiting continues, the buffer would already be full.

To add Inbound and Outbound rate limiting, we now have a configuration in the `TokenBucketConfig` of `NetworkConfig` to change each independently.  In addition, each one can be turned on or off independently.

Garbage collection is handled at disconnects.  If there are no references to the `Bucket` at disconnect time, it will release the bucket.  Buckets are initialized to 1/4 of their full size to prevent abusing this behavior.

I've also spent time to simplify down some of the extra features I had added in the previous implementation, to just be the following:
* One bucket size / rate for all buckets
* Only use the top level for management of buckets, use the bucket for all rate limiting
* Allow for making an `open` rate limiter, which just lets all traffic through for testing purposes
* Garbage collection, based on reference counting
* Fill the buckets only when you try to use them
* Allow for adding back unused tokens (for something like a `read` where you might not use all of them)

### Why not use a third party library?
I spent time looking at Governor, and after some back and forth, it didn't give us quite what we wanted, plus had to pull in more dependencies that we didn't necessarily need like Dashmap.  Now, the map operations are entirely off of the critical path, just garbage collection remains, which should only be periodic.

### Testing
From the integration test in the RateLimiting main function, here's an output of one of the runs
```
Starting experiments
Bytes to test: 1000000, Expected Throughput(actions/s): 1, Expected Throughput(bytes/s): 10000, 

== No Read or Write Limiting ==
Write stats	|	Total actions: 100	|	Total bytes: 1000000	|	Throughput(actions/s): 20000	|	Throughput(bytes/s): 200000000	|	Time elapsed: 5.46656ms
Read stats	|	Total actions: 100	|	Total bytes: 1000000	|	Throughput(actions/s): 20000	|	Throughput(bytes/s): 200000000	|	Time elapsed: 5.534616ms

== Read Limited ==
Write stats	|	Total actions: 100	|	Total bytes: 1000000	|	Throughput(actions/s): 1.8828491272994297	|	Throughput(bytes/s): 18828.491272994295	|	Time elapsed: 53.111312588s
Read stats	|	Total actions: 100	|	Total bytes: 1000000	|	Throughput(actions/s): 1.0101112132445782	|	Throughput(bytes/s): 10101.112132445784	|	Time elapsed: 98.99958156s

== Write Limited ==
Write stats	|	Total actions: 100	|	Total bytes: 1000000	|	Throughput(actions/s): 1.010070401907013	|	Throughput(bytes/s): 10100.70401907013	|	Time elapsed: 99.003092829s
Read stats	|	Total actions: 100	|	Total bytes: 1000000	|	Throughput(actions/s): 1.010070401907013	|	Throughput(bytes/s): 10100.70401907013	|	Time elapsed: 99.003324322s

== Read & Write Limited ==
Write stats	|	Total actions: 100	|	Total bytes: 1000000	|	Throughput(actions/s): 1.0101010101010102	|	Throughput(bytes/s): 10101.0101010101	|	Time elapsed: 99.000364369s
Read stats	|	Total actions: 100	|	Total bytes: 1000000	|	Throughput(actions/s): 1.0101010101010102	|	Throughput(bytes/s): 10101.0101010101	|	Time elapsed: 99.000589412s
```

And a bigger one to show that average throughput approaches the rate limiting with a higher sample size.
```
Starting experiments
Bytes to test: 10000000, Expected Throughput(actions/s): 1, Expected Throughput(bytes/s): 10000, 

== No Read or Write Limiting ==
Write stats	|	Total actions: 1000	|	Total bytes: 10000000	|	Throughput(actions/s): 21276.59574468085	|	Throughput(bytes/s): 212765957.44680852	|	Time elapsed: 47.88036ms
Read stats	|	Total actions: 1000	|	Total bytes: 10000000	|	Throughput(actions/s): 21276.59574468085	|	Throughput(bytes/s): 212765957.44680852	|	Time elapsed: 47.960291ms

== Read Limited ==
Write stats	|	Total actions: 1000	|	Total bytes: 10000000	|	Throughput(actions/s): 1.04808180068838	|	Throughput(bytes/s): 10480.818006883801	|	Time elapsed: 954.124338916s
Read stats	|	Total actions: 1000	|	Total bytes: 10000000	|	Throughput(actions/s): 1.001001001001001	|	Throughput(bytes/s): 10010.01001001001	|	Time elapsed: 999.000938076s

== Write Limited ==
Write stats	|	Total actions: 1000	|	Total bytes: 10000000	|	Throughput(actions/s): 1.00099398702912	|	Throughput(bytes/s): 10009.9398702912	|	Time elapsed: 999.007847151s
Read stats	|	Total actions: 1000	|	Total bytes: 10000000	|	Throughput(actions/s): 1.0009929850411607	|	Throughput(bytes/s): 10009.929850411609	|	Time elapsed: 999.008094395s

== Read & Write Limited ==
Write stats	|	Total actions: 1000	|	Total bytes: 10000000	|	Throughput(actions/s): 1.001001001001001	|	Throughput(bytes/s): 10010.01001001001	|	Time elapsed: 999.00025048s
Read stats	|	Total actions: 1000	|	Total bytes: 10000000	|	Throughput(actions/s): 1.001001001001001	|	Throughput(bytes/s): 10010.01001001001	|	Time elapsed: 999.000346919s
```

### Other Reading
* https://en.wikipedia.org/wiki/Token_bucket

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -158,6 +158,7 @@ members = [
 # For more, see the "Conditional compilation for tests" section in documentation/coding_guidelines.md.
 default-members = [
     "common/trace",
+    "common/rate-limiter",
     "config/generate-key",
     "config/management/genesis",
     "config/management/operational",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -23,6 +23,7 @@ members = [
     "common/num-variants",
     "common/proptest-helpers",
     "common/proxy",
+    "common/rate-limiter",
     "common/retrier",
     "common/short-hex-str",
     "common/subscription-service",
--- a/common/rate-limiter/src/async_lib.rs
+++ b/common/rate-limiter/src/async_lib.rs
@@ -4,7 +4,7 @@
 use crate::rate_limit::{Bucket, SharedBucket};
 use diem_infallible::Mutex;
 use futures::{
-    io::AsyncRead,
+    io::{AsyncRead, Error},
     ready,
     task::{Context, Poll},
     AsyncWrite, Future,
@@ -131,10 +131,46 @@ impl<T: AsyncWrite> AsyncWrite for AsyncRateLimiter<T> {
     }
 }
 
+impl<T: tokio::io::AsyncRead> tokio::io::AsyncRead for AsyncRateLimiter<T> {
+    fn poll_read(
+        self: Pin<&mut Self>,
+        cx: &mut Context<'_>,
+        buf: &mut [u8],
+    ) -> Poll<io::Result<usize>> {
+        let this = self.project();
+        this.rate_limiter
+            .poll_limited(this.inner, cx, buf.len(), |resource, cx, allowed| {
+                resource.poll_read(cx, &mut buf[..allowed])
+            })
+    }
+}
+
+impl<T: tokio::io::AsyncWrite> tokio::io::AsyncWrite for AsyncRateLimiter<T> {
+    fn poll_write(
+        self: Pin<&mut Self>,
+        cx: &mut Context<'_>,
+        buf: &[u8],
+    ) -> Poll<io::Result<usize>> {
+        let this = self.project();
+        this.rate_limiter
+            .poll_limited(this.inner, cx, buf.len(), |resource, cx, allowed| {
+                resource.poll_write(cx, &buf[..allowed])
+            })
+    }
+
+    fn poll_flush(self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<io::Result<()>> {
+        self.project().inner.poll_flush(cx)
+    }
+
+    fn poll_shutdown(self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Result<(), Error>> {
+        self.project().inner.poll_shutdown(cx)
+    }
+}
+
 #[cfg(test)]
 mod tests {
     use super::*;
-    use futures::AsyncReadExt;
+    use tokio::io::AsyncReadExt as TokioAsyncReadExt;
 
     #[tokio::test]
     async fn test_async_read() {
--- a/common/rate-limiter/src/async_lib.rs
+++ b/common/rate-limiter/src/async_lib.rs
@@ -21,7 +21,8 @@ struct PollRateLimiter {
 
 impl PollRateLimiter {
     fn new(bucket: Option<SharedBucket>) -> Self {
-        let bucket = bucket.unwrap_or_else(|| Arc::new(Mutex::new(Bucket::open())));
+        let bucket =
+            bucket.unwrap_or_else(|| Arc::new(Mutex::new(Bucket::open(String::from("None")))));
         PollRateLimiter {
             bucket,
             delay: None,
@@ -70,7 +71,7 @@ impl PollRateLimiter {
             Poll::Ready(Ok(actual)) => allowed.saturating_sub(*actual),
             _ => allowed,
         };
-        self.bucket.lock().add_tokens(tokens_to_return);
+        self.bucket.lock().return_tokens(tokens_to_return);
 
         result
     }
@@ -175,7 +176,7 @@ mod tests {
     #[tokio::test]
     async fn test_async_read() {
         let source: &[u8] = b"12345678901234567890123456";
-        let rate_limiter = Arc::new(Mutex::new(Bucket::new(15, 15, 5)));
+        let rate_limiter = Arc::new(Mutex::new(Bucket::new("test".to_string(), 15, 15, 5)));
         let mut reader = AsyncRateLimiter::new(source, Some(rate_limiter));
 
         let mut buf: [u8; 30] = [0; 30];
--- a/common/rate-limiter/src/async_lib.rs
+++ b/common/rate-limiter/src/async_lib.rs
@@ -1,0 +1,166 @@
+// Copyright (c) The Diem Core Contributors
+// SPDX-License-Identifier: Apache-2.0
+
+use crate::rate_limit::{Bucket, SharedBucket};
+use diem_infallible::Mutex;
+use futures::{
+    io::AsyncRead,
+    ready,
+    task::{Context, Poll},
+    AsyncWrite, Future,
+};
+use pin_project::pin_project;
+use std::{io, pin::Pin, sync::Arc};
+use tokio::time::{delay_until, Delay};
+
+/// An inner struct for keeping track of the delay, and bucket of rate limiting
+struct PollRateLimiter {
+    bucket: SharedBucket,
+    delay: Option<Delay>,
+}
+
+impl PollRateLimiter {
+    fn new(bucket: Option<SharedBucket>) -> Self {
+        let bucket = bucket.unwrap_or_else(|| Arc::new(Mutex::new(Bucket::open())));
+        PollRateLimiter {
+            bucket,
+            delay: None,
+        }
+    }
+
+    /// Poll and attempt to acquire the `requested` amount of tokens.
+    /// Keep trying until some amount of tokens are acquired.  Note: This doesn't provide
+    /// fairness so if two pollers hold the same bucket, one could continually lose.
+    fn poll_acquire(&mut self, cx: &mut Context<'_>, requested: usize) -> Poll<usize> {
+        loop {
+            // Wait until the delay is finished.
+            if let Some(ref mut delay) = self.delay {
+                ready!(Pin::new(delay).poll(cx));
+                self.delay = None;
+            }
+            // Try to acquire some tokens. If we're rate limited, we have to wait
+            // before trying again.
+            match self.bucket.lock().acquire_tokens(requested) {
+                Ok(allowed) => return Poll::Ready(allowed),
+                Err(wait_time) => {
+                    self.delay = Some(delay_until(tokio::time::Instant::from_std(wait_time)));
+                }
+            }
+        }
+    }
+
+    /// Poll an inner reader or writer, rate limited by the `bucket`.  This will provide an amount
+    /// of bytes allowed ot be read including partial reads.  It will return unused tokens back to
+    /// the bucket on partial reads / writes.
+    pub fn poll_limited<
+        T,
+        Action: FnOnce(Pin<&mut T>, &mut Context<'_>, usize) -> Poll<io::Result<usize>>,
+    >(
+        &mut self,
+        resource: Pin<&mut T>,
+        cx: &mut Context<'_>,
+        requested: usize,
+        poll_resource: Action,
+    ) -> Poll<io::Result<usize>> {
+        let allowed = ready!(self.poll_acquire(cx, requested));
+        let result = poll_resource(resource, cx, allowed);
+
+        // In order to have an accurate throttle rate, we remove tokens, and then add ones we don't use
+        let tokens_to_return = match &result {
+            Poll::Ready(Ok(actual)) => allowed.saturating_sub(*actual),
+            _ => allowed,
+        };
+        self.bucket.lock().add_tokens(tokens_to_return);
+
+        result
+    }
+}
+
+/// A rate limiter for `AsyncRead` or `AsyncWrite` interfaces to rate limit read/write bytes
+///
+/// This will pause and wait to send any future bytes until it's permitted to in the future
+#[pin_project]
+pub struct AsyncRateLimiter<T> {
+    #[pin]
+    inner: T,
+    rate_limiter: PollRateLimiter,
+}
+
+impl<T> AsyncRateLimiter<T> {
+    pub fn new(inner: T, bucket: Option<SharedBucket>) -> Self {
+        Self {
+            inner,
+            rate_limiter: PollRateLimiter::new(bucket),
+        }
+    }
+}
+
+impl<T: AsyncRead> AsyncRead for AsyncRateLimiter<T> {
+    fn poll_read(
+        self: Pin<&mut Self>,
+        cx: &mut Context<'_>,
+        buf: &mut [u8],
+    ) -> Poll<io::Result<usize>> {
+        let this = self.project();
+        this.rate_limiter
+            .poll_limited(this.inner, cx, buf.len(), |resource, cx, allowed| {
+                resource.poll_read(cx, &mut buf[..allowed])
+            })
+    }
+}
+
+impl<T: AsyncWrite> AsyncWrite for AsyncRateLimiter<T> {
+    fn poll_write(
+        self: Pin<&mut Self>,
+        cx: &mut Context<'_>,
+        buf: &[u8],
+    ) -> Poll<io::Result<usize>> {
+        let this = self.project();
+        this.rate_limiter
+            .poll_limited(this.inner, cx, buf.len(), |resource, cx, allowed| {
+                resource.poll_write(cx, &buf[..allowed])
+            })
+    }
+
+    fn poll_flush(self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<io::Result<()>> {
+        self.project().inner.poll_flush(cx)
+    }
+
+    fn poll_close(self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<io::Result<()>> {
+        self.project().inner.poll_close(cx)
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use futures::AsyncReadExt;
+
+    #[tokio::test]
+    async fn test_async_read() {
+        let source: &[u8] = b"12345678901234567890123456";
+        let rate_limiter = Arc::new(Mutex::new(Bucket::new(15, 15, 5)));
+        let mut reader = AsyncRateLimiter::new(source, Some(rate_limiter));
+
+        let mut buf: [u8; 30] = [0; 30];
+        assert_eq!(15, reader.read(&mut buf).await.expect("Successful read"));
+        assert_eq!(
+            5,
+            reader.read(&mut buf[15..]).await.expect("Successful read")
+        );
+        assert_eq!(
+            5,
+            reader.read(&mut buf[20..]).await.expect("Successful read")
+        );
+        assert_eq!(
+            1,
+            reader.read(&mut buf[25..]).await.expect("Successful read")
+        );
+        assert_eq!(
+            0,
+            reader.read(&mut buf[26..]).await.expect("Successful read")
+        );
+
+        assert_eq!(&buf[..26], source);
+    }
+}
--- a/common/rate-limiter/src/lib.rs
+++ b/common/rate-limiter/src/lib.rs
@@ -1,0 +1,6 @@
+// Copyright (c) The Diem Core Contributors
+// SPDX-License-Identifier: Apache-2.0
+
+#![forbid(unsafe_code)]
+
+pub mod rate_limit;
--- a/common/rate-limiter/src/lib.rs
+++ b/common/rate-limiter/src/lib.rs
@@ -3,4 +3,5 @@
 
 #![forbid(unsafe_code)]
 
+pub mod async_lib;
 pub mod rate_limit;
--- a/common/rate-limiter/src/main.rs
+++ b/common/rate-limiter/src/main.rs
@@ -39,6 +39,7 @@ struct RunStats {
 
 /// A main function for testing rate limiting throughput and backpressure
 fn main() {
+    diem_logger::DiemLogger::init_for_testing();
     println!("Starting experiments");
     println!(
         "Bytes to test: {}, Expected Throughput(actions/s): {}, Expected Throughput(bytes/s): {}, ",
@@ -204,8 +205,8 @@ async fn test_rate_limiter<
     results
 }
 
-fn simple_shared_bucket(size: usize) -> SharedBucket {
-    Arc::new(Mutex::new(Bucket::new(size, size, size)))
+fn simple_shared_bucket(label: &'static str, size: usize) -> SharedBucket {
+    Arc::new(Mutex::new(Bucket::new(label.to_string(), size, size, size)))
 }
 
 /// For comparison, no rate limiting
@@ -217,7 +218,7 @@ async fn test_no_rate_limit(num_bytes: usize) -> Vec<RunStats> {
 /// that causes the backpressure only to occur after a large amount of traffic first. (> 500k bytes)
 /// > 55 chunks could cause backpressure, for a proper test > 100 or > 200 give good results.
 async fn test_rate_limit_read(num_bytes: usize, throughput: usize) -> Vec<RunStats> {
-    let inbound_rate_limiter = simple_shared_bucket(throughput);
+    let inbound_rate_limiter = simple_shared_bucket("read", throughput);
     test_rate_limiter(
         num_bytes,
         |reader| AsyncRateLimiter::new(reader, Some(inbound_rate_limiter)),
@@ -228,7 +229,7 @@ async fn test_rate_limit_read(num_bytes: usize, throughput: usize) -> Vec<RunSta
 
 /// Tests only if the writer is rate limited, which should provide a smoother read rate
 async fn test_rate_limit_write(num_bytes: usize, throughput: usize) -> Vec<RunStats> {
-    let outbound_rate_limiter = simple_shared_bucket(throughput);
+    let outbound_rate_limiter = simple_shared_bucket("write", throughput);
     test_rate_limiter(
         num_bytes,
         |reader| reader,
@@ -240,8 +241,8 @@ async fn test_rate_limit_write(num_bytes: usize, throughput: usize) -> Vec<RunSt
 /// Tests if both are rate limited.  Results should be roughly the same in the long term as
 /// the other two
 async fn test_rate_limit_read_write(num_bytes: usize, throughput: usize) -> Vec<RunStats> {
-    let inbound_rate_limiter = simple_shared_bucket(throughput);
-    let outbound_rate_limiter = simple_shared_bucket(throughput);
+    let inbound_rate_limiter = simple_shared_bucket("read", throughput);
+    let outbound_rate_limiter = simple_shared_bucket("write", throughput);
     test_rate_limiter(
         num_bytes,
         |reader| AsyncRateLimiter::new(reader, Some(inbound_rate_limiter)),
--- a/common/rate-limiter/src/main.rs
+++ b/common/rate-limiter/src/main.rs
@@ -21,11 +21,15 @@ use tokio::{
 /// Where to run the test read/write
 const ADDRESS: &str = "localhost:7777";
 /// Total # of bytes to transfer.  May transfer more, but won't transfer less
-const BYTES_TO_TRANSFER: usize = 1_000_000;
+const BYTES_TO_TRANSFER: usize = 100_000_000;
 /// Expected throughput to be configured on the rate limiter
 const ALLOWED_BYTES_PER_SEC: usize = 10_000;
 /// Chunk size to write bytes to the stream at a time
 const CHUNK_SIZE: usize = 10_000;
+/// Buffer size used by Tokio when reading, backpressure only occurs after this is full
+const READ_BUFFER_SIZE: usize = 100;
+/// Buffer size used by Tokio when writing
+const WRITE_BUFFER_SIZE: usize = 100;
 
 /// Holder for statistics of a single experiment and reader/writer
 struct RunStats {
@@ -42,9 +46,9 @@ fn main() {
     diem_logger::DiemLogger::init_for_testing();
     println!("Starting experiments");
     println!(
-        "Bytes to test: {}, Expected Throughput(actions/s): {}, Expected Throughput(bytes/s): {}, ",
+        "Bytes to test: {}, Expected Throughput(actions/s): {}, Expected Throughput(bytes/s): {}",
         BYTES_TO_TRANSFER,
-        ALLOWED_BYTES_PER_SEC / CHUNK_SIZE,
+        ALLOWED_BYTES_PER_SEC as f64 / CHUNK_SIZE as f64,
         ALLOWED_BYTES_PER_SEC
     );
 
@@ -68,7 +72,29 @@ fn main() {
     run_experiment(
         "Read & Write Limited",
         &runtime,
-        test_rate_limit_read_write(BYTES_TO_TRANSFER, ALLOWED_BYTES_PER_SEC),
+        test_rate_limit_read_write(
+            BYTES_TO_TRANSFER,
+            ALLOWED_BYTES_PER_SEC,
+            ALLOWED_BYTES_PER_SEC,
+        ),
+    );
+    run_experiment(
+        "Write 2x Read Limited",
+        &runtime,
+        test_rate_limit_read_write(
+            BYTES_TO_TRANSFER,
+            ALLOWED_BYTES_PER_SEC,
+            ALLOWED_BYTES_PER_SEC.saturating_mul(2),
+        ),
+    );
+    run_experiment(
+        "Read 2x Write Limited",
+        &runtime,
+        test_rate_limit_read_write(
+            BYTES_TO_TRANSFER,
+            ALLOWED_BYTES_PER_SEC.saturating_mul(2),
+            ALLOWED_BYTES_PER_SEC,
+        ),
     );
 }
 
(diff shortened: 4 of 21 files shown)


libra-action:
```
target packages:
  A diem-rate-limiter 0.1.0 (workspace, path 'common/rate-limiter')
    * features: [none]
```

gregnazario:
TBD garbage collection & Async waking wasn't finished

common/rate-limiter/src/lib.rs
```diff
@@ -0,0 +1,392 @@
...
+        mut self: Pin<&mut Self>,
+        cx: &mut Context<'_>,
+        buf: &mut [u8],
+    ) -> Poll<std::io::Result<usize>> {
+        let requested = buf.len();
+        let allowed = Pin::new(&mut self.rate_limiter.lock().unwrap()).acquire_tokens(requested);
+
+        // Stop now if we're completely throttled
+        if allowed == 0 {
+            return Poll::Pending;
```
phlip9, line 275: if we return `Poll::Pending` like this the task may hang. from the docs (https://doc.rust-lang.org/std/task/enum.Poll.html):

> When a function returns Pending, the function must also ensure that the current task is scheduled to be awoken when progress can be made.

you could get around this by yielding like

```rust
if allowed == 0 {
    cx.waker.wake();
    return Poll::Pending;
}
```

but that's just barely worse than a busy loop. adding a backoff delay before the next poll_read would incidentally solve this issue. maybe we change the interface of `acquire_tokens` to return `Result<usize, Duration>`, interpreted as: "Err(Duration) implies there's no capacity; you should wait this long before asking again".

common/rate-limiter/src/lib.rs
```diff
@@ -0,0 +1,392 @@
...
+}
+
+impl<T: AsyncRead + Unpin> AsyncRead for AsyncReadRateLimiter<T> {
+    fn poll_read(
+        mut self: Pin<&mut Self>,
+        cx: &mut Context<'_>,
+        buf: &mut [u8],
+    ) -> Poll<std::io::Result<usize>> {
+        let requested = buf.len();
+        let allowed = Pin::new(&mut self.rate_limiter.lock().unwrap()).acquire_tokens(requested);
```
phlip9, line 271: the `Pin::new` stuff isn't necessary. should just be able to do `self.rate_limiter.lock().acquire_tokens(requested)`

common/rate-limiter/src/lib.rs
```diff
@@ -0,0 +1,392 @@
...
+use futures::{
+    task::{Context, Poll},
+    AsyncRead,
+};
+use std::{
+    cmp::min,
+    collections::HashMap,
+    fmt::Debug,
+    hash::Hash,
+    sync::{Arc, Mutex, RwLock},
```
phlip9, line 15: use the fancy new `diem_infallible::{Mutex, RwLock}` : p

don't need to unwrap after `.lock()`!

common/rate-limiter/src/lib.rs
```diff
@@ -0,0 +1,392 @@
...
+        // The other option is to just throttle on how much we request anyways, which is less accurate, but
+        // faster.
+        let result = Pin::new(&mut self.reader).poll_read(cx, buf);
+        let tokens_to_return = match &result {
+            Poll::Ready(Ok(actual)) => allowed - *actual,
+            _ => allowed,
+        };
+
+        // Return unused tokens
+        Pin::new(&mut self.rate_limiter.lock().unwrap()).add_tokens(tokens_to_return);
```
phlip9, line 291: same here don't need the `Pin::new`

common/rate-limiter/src/lib.rs
```diff
@@ -0,0 +1,392 @@
...
+            .for_each(|bucket| {
+                bucket.lock().unwrap().refill();
+            });
+    }
+
+    /// Garbage collects unused references (there's only one reference, the map)
+    fn garbage_collect(&mut self) {
+        let mut write = self.buckets.write().unwrap();
+
+        let keys_to_remove: Vec<_> = write
```
phlip9, line 149: clever trick:

can use [`HashMap::retain`](https://doc.rust-lang.org/std/collections/struct.HashMap.html#method.retain) here : )

```rust
write.retain(|key, bucket| Arc::strong_count(bucket) > 1);
```

common/rate-limiter/src/lib.rs
```diff
@@ -0,0 +1,392 @@
...
+    /// This should occur on an interval basis
+    fn refill(&mut self) {
+        self.buckets
+            .write()
+            .unwrap()
+            .values_mut()
+            .for_each(|bucket| {
+                bucket.lock().unwrap().refill();
+            });
+    }
```
phlip9, line 143: if `acquire_tokens` does a `refill`, is this necessary?

common/rate-limiter/src/lib.rs
```diff
@@ -0,0 +1,392 @@
...
+        Self {
+            buckets: RwLock::new(HashMap::new()),
+            default_bucket_size: std::usize::MAX,
+            default_fill_rate: std::usize::MAX,
+            open: true,
+        }
+    }
+
+    /// Retrieve bucket, or create a new one
+    pub fn bucket(&mut self, key: Key) -> Arc<Mutex<Bucket>> {
```
phlip9, line 99: if we're using a `RwLock`, then most of these methods should be `&self`. if we're using `&mut self` then there's no point to the `RwLock` since we would have exclusive access to the `TokenBucketRateLimiter` : p

common/rate-limiter/src/lib.rs
```diff
@@ -0,0 +1,392 @@
...
+/// This can be used as a standalone rate limiter; however, to make it more useful
+/// it should be wrapped in an `Arc` and a `Mutex` to be shared across threads.
+#[derive(Debug)]
+pub struct Bucket {
+    /// The current number of available tokens to be used
+    tokens: usize,
+    /// Maximum number of `tokens` in the bucket
+    size: usize,
+    /// The fill rate of the bucket.  Amount added to `tokens` on a `refill`
+    rate: usize,
```
phlip9, line 177: probably clarify the units on `rate`. it looks like `tokens/sec`?

common/rate-limiter/src/lib.rs
```diff
@@ -0,0 +1,392 @@
...
+/// A rate limiter for `AsyncRead` interfaces to rate limit read bytes
+///
+/// This will pause and wait to send any future bytes until it's permitted to in the future
+pub struct AsyncReadRateLimiter<T: AsyncRead + Unpin> {
+    reader: T,
+    rate_limiter: Arc<Mutex<Bucket>>,
+}
+
+impl<T: AsyncRead + Unpin> AsyncReadRateLimiter<T> {
+    pub fn new(reader: T, rate_limiter: Arc<Mutex<Bucket>>) -> Self {
```
phlip9, line 256: it looks like this is supposed to be a `rate_limiter: Option<..>`?

common/rate-limiter/src/lib.rs
gregnazario: Since we always provide tokens back (even if it's not the requested size), we can wait the interval period instead.  If we wanted to put jitter, or other information, we could put the `Duration` and give some jitter outside of the second wait periodl.

common/rate-limiter/src/lib.rs
gregnazario: Yeah, tokens/sec

common/rate-limiter/src/lib.rs
gregnazario: This was originally to have periodic fills of the buckets at a regular interval.  But, now since it's doing it on acquire tokens it's not needed.

common/rate-limiter/src/lib.rs
phlip9: right, `acquire_tokens` happens to tell you to wait the interval period ; )

common/rate-limiter/src/lib.rs
gregnazario: I don't think so?  I was first playing around with using `Option<..>` everywhere, then just made the rate limiter let everything through instead.

libra-action:
```
target packages:
  A diem-rate-limiter 0.1.0 (workspace, path 'common/rate-limiter')
    * features: [none]
```

libra-action:
```
target packages:
  A diem-rate-limiter 0.1.0 (workspace, path 'common/rate-limiter')
    * features: [none]
```

common/rate-limiter/src/lib.rs
```diff
@@ -0,0 +1,485 @@
...
+        mut self: Pin<&mut Self>,
+        cx: &mut Context<'_>,
+        buf: &mut [u8],
+    ) -> Poll<std::io::Result<usize>> {
+        let requested = buf.len();
+        let allowed = self.rate_limiter.lock().acquire_tokens(requested);
+
+        // Stop now if we're completely throttled
+        if allowed == 0 {
+            sleep(self.rate_limiter.lock().time_until_next_refill());
```
phlip9, line 320: don't want to use `thread::sleep` here, since it'll block the actual tokio runtime worker thread (which will block the entire Peer actor and prevent other futures from running on this thread; there are usually at most `num_cores` runtime worker threads). instead use the `tokio::time::Delay` future (created with `tokio::time::delay_for`).

unfortunately, we can't just await the delay here, since we're writing a poll function. when we write poll functions, it's more like writing a state machine. it might be helpful to think about the states and when you transition between them:

Screen Shot 2020-12-16 at 10 31 30

you can encode the state like

```rust
use tokio::time::Delay;

struct AsyncReadRateLimiter {
    reader: T,
    rate_limiter: Arc<Mutex<Bucket>>,
    waiting: Option<Delay>,
}
```

you might need a `loop { .. }`

common/rate-limiter/src/lib.rs
```diff
@@ -0,0 +1,485 @@
...
+}
+
+impl<T: AsyncRead + Unpin> AsyncRead for AsyncReadRateLimiter<T> {
+    fn poll_read(
+        mut self: Pin<&mut Self>,
+        cx: &mut Context<'_>,
+        buf: &mut [u8],
+    ) -> Poll<std::io::Result<usize>> {
+        let requested = buf.len();
+        let allowed = self.rate_limiter.lock().acquire_tokens(requested);
```
phlip9, line 316: probably want to get the tokens or time to refill in one call so we can just avoid any weird parallel interleavings

common/rate-limiter/src/lib.rs
gregnazario: I "think" I fixed this, there's now a test to check it out

libra-action:
```
target packages:
  A diem-rate-limiter 0.1.0 (workspace, path 'common/rate-limiter')
    * features: [none]
```

libra-action:
```
target packages:
  A diem-rate-limiter 0.1.0 (workspace, path 'common/rate-limiter')
    * features: [none]
```

libra-action:
```
target packages:
  A diem-rate-limiter 0.1.0 (workspace, path 'common/rate-limiter')
    * features: [none]
```

libra-action:
```
target packages:
  A diem-rate-limiter 0.1.0 (workspace, path 'common/rate-limiter')
    * features: [none]
```

common/rate-limiter/src/rate_limit.rs
```diff
@@ -0,0 +1,446 @@
...
+            self.buckets
+                .write()
+                .entry(key)
+                .or_insert_with(|| bucket_create(size, rate))
+                .clone()
+        }
+    }
+
+    /// Garbage collects a single key, if we know what it is
+    pub fn garbage_collect_key(&self, key: &Key) -> bool {
```
davidiw, line 120: maybe we should call this maybe_garbage_collect_key?

common/rate-limiter/src/rate_limit.rs
```diff
@@ -0,0 +1,446 @@
...
+        if !self.open {
+            self.bucket_inner(key, |size, rate| {
+                Arc::new(Mutex::new(Bucket::new(size, rate)))
+            })
+        } else {
+            self.bucket_inner(key, |_, _| Arc::new(Mutex::new(Bucket::open())))
+        }
+    }
+
+    fn bucket_inner<F: FnOnce(NonZeroUsize, NonZeroUsize) -> Arc<Mutex<Bucket>>>(
```
davidiw, line 96: Maybe typedef Arc<Mutex<Bucket>> to SharedBucket?

common/rate-limiter/src/rate_limit.rs
```diff
@@ -0,0 +1,446 @@
...
+        &self,
+        key: Key,
+        bucket_create: F,
+    ) -> Arc<Mutex<Bucket>> {
+        // Attempt to do a weaker read lock first, followed by a write lock if it's missing
+        // For the common (read) case, there should be higher throughput
+        let maybe_value = { self.buckets.read().get(&key).cloned() };
+
+        if let Some(value) = maybe_value {
```
davidiw, line 105: why not merge these two lines?

common/rate-limiter/src/rate_limit.rs
```diff
@@ -0,0 +1,446 @@
...
+            size: max,
+            rate: max,
+            last_refresh_time: Instant::now(),
+            open: true,
+        }
+    }
+
+    /// Refill tokens based on how many seconds have passed since last refresh
+    pub(crate) fn refill(&mut self) {
+        let num_intervals = self.last_refresh_time.elapsed().as_secs();
```
davidiw, line 179: how would we simulate / test this?

common/rate-limiter/src/rate_limit.rs
```diff
@@ -0,0 +1,446 @@
...
+        }
+    }
+
+    /// Retrieve the maximum amount of tokens up to `count`
+    /// Tells us how much of the requested size we can send
+    fn deduct_tokens(&mut self, requested: usize) -> usize {
+        let tokens = self.tokens;
+        let tokens_left = tokens.saturating_sub(requested);
+        let tokens_allowed = min(tokens, requested);
+        self.tokens = tokens_left;
```
davidiw, line 234: why are we setting tokens_left and self.tokens, feels like this should just be set once and erase line 232... specifically
```
let tokens_allowed = min(self.tokens, requested)
self.tokens = self.tokens.saturating_sub(tokens_allowed)
```

common/rate-limiter/src/rate_limit.rs
```diff
@@ -0,0 +1,446 @@
...
+            .acquire_tokens(1)
+            .expect_err("Expected time to wait");
+    }
+
+    // Check number of keys in the rate limiter
+    fn assert_num_keys(rate_limiters: &TokenBucketRateLimiter<&str>, num_keys: usize) {
+        assert_eq!(num_keys, rate_limiters.buckets.read().len())
+    }
+
+    macro_rules! NonZeroUsize {
```
davidiw, line 291: hahaha, why not move this to infallible and use it more prominently

common/rate-limiter/src/rate_limit.rs
```diff
@@ -0,0 +1,446 @@
...
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use std::{sync::MutexGuard, thread::sleep};
+    use tokio::time::Duration;
+
+    // Helper function, checks the throttle limit
+    fn assert_num_allowed(rate_limiter: &mut MutexGuard<Bucket>, num_allowed: usize) {
```
davidiw, line 274: feels like this should be called assert_acquire

davidiw: also why not just pass in the arc bucket instead of the mutex guard?

common/rate-limiter/src/rate_limit.rs
```diff
@@ -0,0 +1,446 @@
...
+        let rate_limiter =
+            TokenBucketRateLimiter::new(NonZeroUsize!(bucket_size), NonZeroUsize!(bucket_rate));
+
+        let bucket_arc = rate_limiter.bucket(key);
+        let mut bucket = bucket_arc.lock();
+        assert_num_allowed(&mut bucket, bucket_size);
+
+        // After 1 refill period, we should be at least 1 rate change if not more
+        // TODO: Put in a mock time service
+        sleep(bucket.time_of_next_refill().duration_since(Instant::now()));
```
davidiw, line 362: might be a good idea to adapt Philip's simulated time.

network/src/peer_manager/mod.rs
```diff
@@ -533,6 +539,13 @@ where
                     );
                     self.send_conn_notification(peer_id, notif);
                 }
+
+                // Garbage collect unused rate limit buckets
```
davidiw, line 543: this implicitly makes an assumption that we have some sort of rate limiting on incoming connections ... otherwise the attacker is incentivized to just create a new connection to drive in more traffic... maybe we should start buckets out as empty or 1/4 full?

davidiw (commented):
Overall looks good, what's missing is a solid e2e test that really shows that this works as expected whether that be an AsyncReader with simulated time or something else. Your thoughts?

davidiw:
it might also be worth setting up a benchmark / main function that uses this and actually shows bandwidth with and without this across either just a null async pipe or a tcp socket.

common/rate-limiter/src/rate_limit.rs
gregnazario: Yeah, or `try_garbage_collect_key`, except the try interface usually gives an `Option`.  Thoughts?

common/rate-limiter/src/rate_limit.rs
gregnazario: I actually did that, and I guess it disappeared :/.  It'll make it cleaner

common/rate-limiter/src/rate_limit.rs
gregnazario: I separated it to ensure the read lock was let go immediately, if a write lock needed to take place after.

Let me circle back and see if I really needed it like this.

common/rate-limiter/src/rate_limit.rs
gregnazario: Lack of sleep :P

common/rate-limiter/src/rate_limit.rs
gregnazario: I thought about that, just didn't want to stretch this out too much across how many crates I'm changing.

common/rate-limiter/src/rate_limit.rs
gregnazario: > also why not just pass in the arc bucket instead of the mutex guard?

I think I was trying to avoid having to write `.lock()` everywhere

common/rate-limiter/src/rate_limit.rs
gregnazario: Yeah, originally it hadn't been pushed through, but now it is so I can adapt it to it.

network/src/peer_manager/mod.rs
gregnazario: So, originally what I wanted to do was have the buckets garbage collected on a periodic basis, ignoring when connections are created and deleted, which made more sense to be full.

It is a good idea to start empty, except it does slow *all* connections by the period to start.

So, what I'll probably do is make it configurable, as a fraction of the bucket fill rate.

common/rate-limiter/src/rate_limit.rs
gregnazario: It's currently tested in the unit tests below.  With a mock time service, it then won't have to wait seconds.

gregnazario:
> it might also be worth setting up a benchmark / main function that uses this and actually shows bandwidth with and without this across either just a null async pipe or a tcp socket.

Yeah, I agree it would be worth setting up benchmarks for this.  I wanted to see how cluster test would do, but a pure rate limiting only test would be useful.

common/rate-limiter/src/rate_limit.rs
gregnazario: Here you go: https://github.com/diem/diem/pull/7011

common/rate-limiter/src/rate_limit.rs
gregnazario: Yep, basically without it being in a separate line, the lifetime of the read lock holds through the lower line write lock, which causes it to lock up on the same thread by never letting go of the read lock before trying to request a write lock.

davidiw:
I think the only outstanding question is whether or not this really works -- especially with back pressure? what's the plan?

gregnazario:
> I think the only outstanding question is whether or not this really works -- especially with back pressure? what's the plan?

I'm working on a E2E test right now to see if I can lower the throttle rate, and force traffic through higher than it such that we see back pressure or not

libra-action:
```
target packages:
  A diem-rate-limiter 0.1.0 (workspace, path 'common/rate-limiter')
    * features: [none]
```

libra-action:
```
target packages:
  A diem-rate-limiter 0.1.0 (workspace, path 'common/rate-limiter')
    * features: [none]
```

gregnazario:
> I think the only outstanding question is whether or not this really works -- especially with back pressure? what's the plan?

I've added a test in the latest push that shows that backpressure does work.  However, it needs to fill the socket buffer before that will occur.  I have the test disabled because it takes too long, but feel free to check it out.  Also, I haven't really done benches in tokio, so it's just a regular test

libra-action:
```
target packages:
  A diem-rate-limiter 0.1.0 (workspace, path 'common/rate-limiter')
    * features: [none]
```

libra-action:
```
target packages:
  A diem-rate-limiter 0.1.0 (workspace, path 'common/rate-limiter')
    * features: [none]
```

Cargo.toml
```diff
@@ -23,6 +23,8 @@ members = [
     "common/num-variants",
     "common/proptest-helpers",
     "common/proxy",
+    "common/rate-limiter",
+    "common/rate-limiter/integration-test",
```
davidiw, line 27: You shouldn't need to add this as a top level crate, see 
https://github.com/diem/diem/blob/master/consensus/safety-rules/Cargo.toml#L49
that points to https://github.com/diem/diem/blob/master/consensus/safety-rules/tests/binary.rs

common/rate-limiter/integration-test/src/main.rs
```diff
@@ -0,0 +1,213 @@
+// Copyright (c) The Diem Core Contributors
+// SPDX-License-Identifier: Apache-2.0
+
+use futures::executor::block_on;
+
+fn main() {
```
davidiw, line 6: oh, I see you went with a main function, then let's just embed that in as the default main for rate limiter or rename it using something like this: https://doc.rust-lang.org/cargo/reference/cargo-targets.html#binaries

common/rate-limiter/integration-test/src/main.rs
```diff
@@ -0,0 +1,213 @@
...
+    }
+
+    fn simple_shared_bucket(size: usize) -> SharedBucket {
+        Arc::new(Mutex::new(Bucket::new(size, size, size)))
+    }
+
+    /// Tests backpressure, where only the reader is limited.  Using a socket adds a buffer
+    /// that causes the backpressure only to occur after a large amount of traffic first. (> 500k bytes)
+    /// > 55 chunks could cause backpressure, for a proper test > 100 or > 200 give good results.
+    pub async fn test_rate_limit_read(num_chunks: usize) -> Vec<RunStats> {
```
davidiw, line 180: would recommend replacing num_chunks with num_bytes as chunks requires digging into the code to understand what's going on.

config/src/config/network_config.rs
```diff
@@ -96,6 +98,10 @@ pub struct NetworkConfig {
     pub max_outbound_connections: usize,
     // Maximum number of outbound connections, limited by PeerManager
     pub max_inbound_connections: usize,
+    // Maximum number of bytes/s for an IP
+    pub inbound_ip_byte_bucket_rate: usize,
+    // Maximum burst of bytes for an IP
+    pub inbound_ip_byte_bucket_size: usize,
```
davidiw, line 104: maybe make this an option -- if both aren't set than we don't instantiate the rate limiter?

alternatively make this an optional struct, since we'll need the same values for writing

davidiw (commented):
overall looks good, would like to get consensus on the config before landing

config/src/config/network_config.rs
gregnazario: Optional struct done.

It's just kinda annoying because now there's a struct in common (for the rate limiter), and one in config (for the config)

libra-action:
```
target packages:
  A diem-rate-limiter 0.1.0 (workspace, path 'common/rate-limiter')
    * features: [none]
```

libra-action:
```
target packages:
  A diem-rate-limiter 0.1.0 (workspace, path 'common/rate-limiter')
    * features: [none]
```

libra-action:
```
target packages:
  A diem-rate-limiter 0.1.0 (workspace, path 'common/rate-limiter')
    * features: [none]
```

libra-action:
```
target packages:
  A diem-rate-limiter 0.1.0 (workspace, path 'common/rate-limiter')
    * features: [none]
```

libra-action:
```
target packages:
  A diem-rate-limiter 0.1.0 (workspace, path 'common/rate-limiter')
    * features: [none]
```

libra-action:
```
target packages:
  A diem-rate-limiter 0.1.0 (workspace, path 'common/rate-limiter')
    * features: [none]
```

libra-action:
```
target packages:
  A diem-rate-limiter 0.1.0 (workspace, path 'common/rate-limiter')
    * features: [none]
```

libra-action:
```
target packages:
  A diem-rate-limiter 0.1.0 (workspace, path 'common/rate-limiter')
    * features: [none]
```

common/rate-limiter/src/async_lib.rs
```diff
@@ -0,0 +1,203 @@
...
+        PollRateLimiter {
+            bucket,
+            delay: None,
+        }
+    }
+
+    /// Poll and attempt to acquire the `requested` amount of tokens.
+    /// Keep trying until some amount of tokens are acquired.  Note: This doesn't provide
+    /// fairness so if two pollers hold the same bucket, one could continually lose.
+    fn poll_acquire(&mut self, cx: &mut Context<'_>, requested: usize) -> Poll<usize> {
```
davidiw, line 35: I know why this bothered me now... there's no logging and it is an infinite loop that makes a deep assumption about the bucket implementation -- that is that the bucket is nonzero

can we perhaps make a max iterations in the loop? (in another PR)

common/rate-limiter/src/async_lib.rs
```diff
@@ -0,0 +1,203 @@
...
+    }
+}
+
+impl<T: AsyncWrite> AsyncWrite for AsyncRateLimiter<T> {
+    fn poll_write(
+        self: Pin<&mut Self>,
+        cx: &mut Context<'_>,
+        buf: &[u8],
+    ) -> Poll<io::Result<usize>> {
+        let this = self.project();
```
davidiw, line 119: why have this as a separate line? same as above

common/rate-limiter/src/main.rs
```diff
@@ -0,0 +1,283 @@
...
+/// A main function for testing rate limiting throughput and backpressure
+fn main() {
+    diem_logger::DiemLogger::init_for_testing();
+    println!("Starting experiments");
+    println!(
+        "Bytes to test: {}, Expected Throughput(actions/s): {}, Expected Throughput(bytes/s): {}",
+        BYTES_TO_TRANSFER,
+        ALLOWED_BYTES_PER_SEC as f64 / CHUNK_SIZE as f64,
+        ALLOWED_BYTES_PER_SEC
+    );
```
davidiw, line 53: nice

common/rate-limiter/src/rate_limit.rs
```diff
@@ -0,0 +1,498 @@
...
+            open: false,
+        }
+    }
+
+    pub fn test(default_bucket_size: usize, default_fill_rate: usize) -> Self {
+        Self::new("test", 100, default_bucket_size, default_fill_rate)
+    }
+
+    /// Used for testing and to not have a rate limiter
+    pub fn open(label: &'static str) -> Self {
```
davidiw, line 89: uncapped?

common/rate-limiter/src/rate_limit.rs
```diff
@@ -0,0 +1,498 @@
...
+            new_bucket_start_percentage: 100,
+            default_bucket_size: std::usize::MAX,
+            default_fill_rate: std::usize::MAX,
+            open: true,
+        }
+    }
+
+    /// Retrieve bucket, or create a new one
+    pub fn bucket(&self, key: Key) -> SharedBucket {
+        if !self.open {
```
davidiw, line 102: maybe invert? and I think uncapped is a little clearer than open but up to you

common/rate-limiter/src/rate_limit.rs
```diff
@@ -0,0 +1,498 @@
...
+            self.bucket_inner(key, |label, initial, size, rate| {
+                Arc::new(Mutex::new(Bucket::new(label, initial, size, rate)))
+            })
+        } else {
+            self.bucket_inner(key, |label, _, _, _| {
+                Arc::new(Mutex::new(Bucket::open(label)))
+            })
```
davidiw, line 109: why not...

```
self.bucket(key, |label, initial, size, rate|) {
  if open {
    Arc::new(Mutex::new(
  } else {
  }
})
```
you can remove 2 lines :P

common/rate-limiter/src/rate_limit.rs
```diff
@@ -0,0 +1,498 @@
...
+                        rate,
+                    )
+                })
+                .clone()
+        }
+    }
+
+    /// Garbage collects a single key, if we know what it is
+    pub fn try_garbage_collect_key(&self, key: &Key) -> bool {
+        let mut write = self.buckets.write();
```
davidiw, line 147: maybe just call this buckets or mut_buckets? not write

common/rate-limiter/src/rate_limit.rs
```diff
@@ -0,0 +1,498 @@
...
+        let arc = write.get(key);
```
davidiw, line 148: might be better to make a single map that does all this?
but also arc => maybe key
```
buckets.get(key).map(|bucket| if ... { write.remove(key); true } else { false }).unwrap_or(false);

**Review by davidiw — APPROVED** [review]
did we cluster test this enabled yet?

File: common/rate-limiter/src/async_lib.rs (hunk 25 above)
Comment (gregnazario) [reply common/rate-limiter/src/async_lib.rs:35]: We could make max iterations on the loop in another PR.

File: common/rate-limiter/src/async_lib.rs (hunk 26 above)
Comment (gregnazario) [reply common/rate-limiter/src/async_lib.rs:119]: I think it was so I didn't type self.project() twice?  Though, it might have been because `self.project()` didn't like being moved twice or something.

File: common/rate-limiter/src/rate_limit.rs (hunk 28 above)
Comment (gregnazario) [reply common/rate-limiter/src/rate_limit.rs:89]: I used the term `open` like an open throttle

File: common/rate-limiter/src/rate_limit.rs (hunk 32 above)
Comment (gregnazario) [reply common/rate-limiter/src/rate_limit.rs:148]: Yeah, these terms aren't the greatest let me make a followup to fix some of these.

Comment (gregnazario) [reply common/rate-limiter/src/rate_limit.rs:148]: I remember now, I tried to use a map statement, and it doesn't work because dealing with making a mutable borrow within another mutable borrow.

**Comment by libra-action** [comment]
```
target packages:
  A diem-rate-limiter 0.1.0 (workspace, path 'common/rate-limiter')
    * features: [none]
```

File: common/rate-limiter/src/async_lib.rs (hunk 25 above)
Comment (phlip9) [reply common/rate-limiter/src/async_lib.rs:35]: is the loop really an issue? it's like polling a lock with a delay if it's already held

```rust
loop {
    if let Some(permit) = lock.try_acquire() {
        return permit;
    } else {
        sleep(delay);
    }
}
```

if we're concerned with starvation, then i'd agree; we don't currently guarantee any fairness across waiters. of course, for the current network use-case where waiters are all from the same IP / client, I don't think fairness is that important. we should definitely document the lack of fairness somewhere clearly though : )

libra-action:
```
target packages:
  A diem-rate-limiter 0.1.0 (workspace, path 'common/rate-limiter')
    * features: [none]
```

gregnazario:
Both a flaky E2E test and a flaky unit test (unrelated) have stopped me from trying to get this to be pushed, so let's just try to land it.

/land

libra-action:
```
target packages:
  A diem-rate-limiter 0.1.0 (workspace, path 'common/rate-limiter')
    * features: [none]
```

libra-action:
```
target packages:
  A diem-rate-limiter 0.1.0 (workspace, path 'common/rate-limiter')
    * features: [none]
```

PR #6943, opened 2020-12-15
[x] allow workspace subsets to be specified for xbuild etc

sunshowers:
Add a new option `--members` which allows for specifying workspace subsets.

Also make `--changed-since` work with `--members`, so that you can say e.g. build anything in LEC that changed since origin/master.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -394,7 +394,7 @@ jobs:
           command: cargo xclippy --workspace --all-targets
       - run:
           name: cargo clippy tcb
-          command: cargo xclippy -p safety-rules  -p execution-correctness -p diem-key-manager
+          command: cargo xclippy --members lec --members lsr --members key-manager
       - run:
           name: cargo fmt
           command: cargo xfmt --check
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -421,7 +421,7 @@ jobs:
       - build_setup
       - restore_cargo_package_cache
       - run:
-          command: RUST_BACKTRACE=1 cargo xcheck -j 8 --production
+          command: RUST_BACKTRACE=1 cargo xcheck -j 16 --members production
       - run:
           command: RUST_BACKTRACE=1 cargo xcheck -j 8 --workspace --all-targets
       - run:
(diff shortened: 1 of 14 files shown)


x.toml
```diff
@@ -157,19 +157,19 @@ members = [
 # Interesting subsets of the workspace, These are used for generating and
 # checking dependency summaries.
 
-[workspace.subsets.lsr]
+[subsets.lsr]
```
rexhoffman, line 160: These really aren't subsets.   They are nodes used to define a subset via a transformation of finding all in workspace dependencies, correct.   The naming sits a little wrong with me and think this needs to be aggressively documented at the various points where this could be misused...   still chewing on this though.

sunshowers: Yeah, good point. Maybe it should be `root-members` instead of `members` to signal this.

rexhoffman: Alright, been chewing on this.   If we want to use the subset terminology we should probably have something like

default-members.subset.lsr to indicate these configurations behave the same as default members would....  

Happy to be convinced otherwise.

.circleci/config.yml
```diff
@@ -369,7 +369,7 @@ jobs:
       - build_setup
       - restore_cargo_package_cache
       - run:
-          command: RUST_BACKTRACE=1 cargo xcheck -j 16 --production
+          command: RUST_BACKTRACE=1 cargo xcheck -j 16 --subset production
```
rexhoffman, line 372: perhaps --members so it's clear what this is a subset of?   Will make more sense if you read the comments below this.

x.toml
rexhoffman: heh didn't see your comment before I made my second one.... wierd.

.circleci/config.yml
sunshowers: Yeah I like `--members`.

sunshowers: Done.

x.toml
sunshowers: I've changed it to `root-members` -- my concern with `default-members` is that it's a bit too specific. I think `root-members` is the right level of generality and communicates this information well. (We can also add `members` in the future if we *don't* want to traverse dependencies.

rexhoffman (approved):
Thanks for the walk through of the guppy logic.   I think I've a better grasp of it now.

PR #6945, opened 2020-12-15
[diem framework] Update hex in comments to reflect new "XDX" name

tzakian:
Fixes a couple places where the ASCII hex encoding of "LBR" wasn't updated to the hex encoding of "XDX" in comments


bob-wilson (approved):
Looks good. Thanks, Tim!

PR #6950, opened 2020-12-16
logger: ensure Metadata is Copy and add backtrace test

bmwill:


common/logger/src/metadata.rs
```diff
@@ -24,42 +23,24 @@ pub struct Metadata {
...
     location: &'static str,
-
-    /// The program backtrace taken when the event occurred. Backtraces are
-    /// only supported for errors.
-    backtrace: Option<String>,
 }
 
 impl Metadata {
-    pub fn new(
+    pub const fn new(
```
gregnazario, line 29: interesting, didn't know you could do that

gregnazario (approved):

PR #6952, opened 2020-12-17
[diem framework] Limit the number of validators in the network to 256 or less 

tzakian:
This PR bounds the maximum number of validators that can participate in consensus to be less than or equal to 256. 

This bounds the vector so that we don't run into gas issues in the future where portions of the validator set may become inaccessible.  We can increase this number later if we need to, but this was agreed as a sensible limit for the validator set at this time.


language/stdlib/modules/DiemSystem.move
```diff
@@ -75,10 +75,15 @@ module DiemSystem {
...
     /// Rate limited when trying to update config
     const ECONFIG_UPDATE_RATE_LIMITED: u64 = 6;
+    /// There was an issue in the addition/removal of a validator to/from the validator set
+    const EVALIDATOR_SET: u64 = 7;
 
     /// Number of microseconds in 5 minutes
     const FIVE_MINUTES: u64 = 300000000;
 
+    /// The maximum number of allowed validators in the validator set
+    const MAX_NUMBER_ALLOWED_VALIDATORS: u64 = 256;
```
sblackshear, line 85: Nit: `MAX_VALIDATORS` is slightly simpler?

language/stdlib/modules/DiemSystem.move
```diff
@@ -75,10 +75,15 @@ module DiemSystem {
     const EVALIDATOR_INDEX: u64 = 5;
     /// Rate limited when trying to update config
     const ECONFIG_UPDATE_RATE_LIMITED: u64 = 6;
+    /// There was an issue in the addition/removal of a validator to/from the validator set
+    const EVALIDATOR_SET: u64 = 7;
```
sblackshear, line 79: `EMAX_VALIDATORS` might be a bit clearer? "limit exceeded: max validators" clearly conveys what's happening, but "limited exceeded: validator set" sounds like it could potentially be too few validators or some other problem.

sblackshear (approved):

language/stdlib/modules/DiemSystem.move
tzakian: Good suggestion! I'll make the change

davidiw:
Just want to call out that this introduces genesis-breaking changes. How do we plan on incorporating this upstream?

tzakian:
> Just want to call out that this introduces genesis-breaking changes. How do we plan on incorporating this upstream?

Good question.  The plan isn't on incorporating this upstream immediately for the next release. The plan that I'm aware of is that this would be deployed in the next release along with any other updates to Move modules/Diem core code.

mickayz (approved):
LGTM

davidiw:
Closing out old PR.

mickayz:
Hi @davidiw, can you share context on why this was closed? My understanding was this mitigated part of T77811081.

tzakian:
I think it was closed because I forgot to land....I'll reopen and work on getting this landed today.

davidiw:
I actually thought this had landed in another space, I guess that was the VASP limit. As this is a breaking change at this point, perhaps we can bring this to a conversation with the gatekeepers?

tzakian:
I'm probably wrong and confused about this all, but I don't think this is a breaking change? The Diem Framework modules will be updated, but it shouldn't break clients in any way or change the data representation of any on-chain value, and this should be rolled out with the normal release schedule for the Diem Framework. I'm happy to bring to the gatekeepers, but not sure if it's necessary?

davidiw:
let's get this landed!!!!!

tzakian:
/land

at last!

PR #6955, opened 2020-12-17
[diem framework] Increase max number of child accounts to 2^16 from 2^8

tzakian:
This PR increases the maximum allowed number of child accounts for a parent VASP account from 2^8 to 2^16.

## `RecoveryAddress` and gas considerations 

In order to prevent possible out-of-gas issues when rotating an authentication key registered with a single `RecoveryAddress` we bound the maximum allowed number of registered keys to the same number as before (256). This is a short-term measure, and in the future we're planning on removing this limitation and changing the underlying data structure to allow more accounts to be registered with a single `RecoveryAddress` (e.g., a hashmap like structure).

## Testing

A new test has been added to make sure that we can rotate all keys in a maxed out recovery address. The test for the bounding of accounts has been removed, since this test takes far too long to be included in the test suite, and we already have this checked in the specification for adding child VASP accounts.


sblackshear (approved):
LGTM--thanks Tim!!

mickayz (commented):
LGTM

PR #6957, opened 2020-12-17
[move cli] remove dry_run flag from publish

sblackshear:
`publish --dry-run` and `check` are the same, so this removes the `--dry-run` flag.
In addition, this fixes a few incorrect/missing comments

### DIFF
--- a/language/tools/move-cli/src/main.rs
+++ b/language/tools/move-cli/src/main.rs
@@ -49,8 +49,8 @@ struct Move {
     /// Directory storing Move resources, events, and module bytecodes produced by script execution.
     #[structopt(long, default_value = DEFAULT_STORAGE_DIR, global = true)]
     storage_dir: String,
-    /// Directory storing Move resources, events, and module bytecodes produced by script execution.
-    #[structopt(long, default_value = DEFAULT_BUILD_DIR, global = true)]
+    /// Directory storing build artifacts produced by compilation
+    #[structopt(long, short = "d", default_value = DEFAULT_BUILD_DIR, global = true)]
     build_dir: String,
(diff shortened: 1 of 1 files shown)


sblackshear:
GH still won't let me add @mengxu-fb as a reviewer, but please add yourself!

language/tools/move-cli/src/main.rs
```diff
@@ -750,11 +748,11 @@ fn main() -> Result<()> {
         Command::Publish {
             source_files,
             no_republish,
-            dry_run,
         } => {
             move_args.prepare_mode(true)?;
             let state = publish(&move_args, !*no_republish, source_files)?;
-            maybe_commit_effects(&move_args, !*dry_run, None, &state)
+            let dry_run = false; // publish is the same as check if dry_run is set
+            maybe_commit_effects(&move_args, dry_run, None, &state)
```
mengxu-fb, line 755: Would it be better if we just have `maybe_commit_effects(&move_args, false, None, &state)` here?

mengxu-fb (approved):
Thanks @sblackshear for tagging me. The changes LGTM!

language/tools/move-cli/src/main.rs
sblackshear: It would be shorter, but I find code with anonymous boolean literals hard to read/update. Ideally, Rust would have named parameters to help us with this...

mengxu-fb:
It seems that my approval does not count because I don't have write-access (which is also the reason I cannot be a reviewer). Someone else needs to stamp this PR :(

tzakian (approved):

PR #6966, opened 2020-12-17
[move-prover] Added cvc4 to move-prover developer setup

cbarrettfb:
Add cvc4 to the developer install script for move-prover


wrwg (approved):
Nice!

PR #6969, opened 2020-12-17
[move CLI] Add "create" command to create test scaffold

tzakian:
This PR adds the `create` command to the Move CLI test sub-command to create a test scaffold as described in the README:
```
move-test 0.1.0
Run expected value tests using the given batch file

USAGE:
    move test [FLAGS] [OPTIONS] <path>

FLAGS:
        --create       Create a new test directory scaffold with the specified <path>
    -h, --help         Prints help information
        --track-cov    Show coverage information after tests are done. By default, coverage will not be tracked nor
                       shown
    -V, --version      Prints version information
    -v                 Print additional diagnostics

OPTIONS:
        --build-dir <build-dir>        Directory storing Move resources, events, and module bytecodes produced by script
                                       execution [default: build]
        --mode <mode>                  Dependency inclusion mode [default: stdlib]
        --storage-dir <storage-dir>    Directory storing Move resources, events, and module bytecodes produced by script
                                       execution [default: storage]

ARGS:
    <path>    a directory path in which all the tests will be executed
```

```
> move test test_dir --create
> tree test_dir
test_dir
├── args.txt
└── src
    ├── modules
    └── scripts
```

### DIFF
--- a/language/tools/move-cli/src/main.rs
+++ b/language/tools/move-cli/src/main.rs
@@ -142,6 +142,9 @@ enum Command {
         /// By default, coverage will not be tracked nor shown.
         #[structopt(long = "track-cov")]
         track_cov: bool,
+        /// Create a new test directory scaffold with the specified <path>
+        #[structopt(long = "create")]
+        create: bool,
     },
     /// View Move resources, events files, and modules stored on disk
     #[structopt(name = "view")]
@@ -839,7 +842,16 @@ fn main() -> Result<()> {
                 move_args.verbose,
             )
         }
-        Command::Test { path, track_cov } => test::run_all(
+        Command::Test {
+            path,
+            track_cov: _,
+            create: true,
+        } => test::create_test_scaffold(path),
+        Command::Test {
+            path,
+            track_cov,
+            create: false,
+        } => test::run_all(
             path,
             &std::env::current_exe()?.to_string_lossy(),
             *track_cov,
--- a/language/tools/move-cli/README.md
+++ b/language/tools/move-cli/README.md
@@ -325,6 +325,19 @@ Command `run scripts/debug_script.move --signers 0xf --mode bare`:
 ...
 ```
 
+The scaffolding for a new test that follows the above structure for tests can be created
+by passing the `--create` flag to `move test` along with the name of the test that you wish to create:
+
+```
+$ move test new_test_name --create
+$ tree new_test_name
+new_test_name
+├── args.txt
+└── src
+    ├── modules
+    └── scripts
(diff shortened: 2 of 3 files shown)


tzakian:
cc @mengxu-fb since I can't seem to add you as a reviewer

language/tools/move-cli/src/main.rs
```diff
@@ -775,7 +778,33 @@ fn main() -> Result<()> {
                 *dry_run,
             )
         }
-        Command::Test { path, track_cov } => test::run_all(
+        Command::Test {
+            path,
+            track_cov: _,
+            create: true,
+        } => {
+            let dirs = ["src", "src/modules", "src/scripts"];
```
mengxu-fb, line 786: I think we have a `pub const DEFAULT_SOURCE_DIR: &str = "src";` in `lib.rs`. Not sure if it will be better to use the const instead of hardcoding it here.

mengxu-fb (approved):
LGTM, but I think you might need another stamp to land it :)

sblackshear (approved):
Looks great! This (or a variant) also be useful as a top-level command for creating a new Move project dir (I think there is a `cargo init` command that does something similar). But probably want to wait until we can generate a default config file/figure out what to with `args.txt`.

language/tools/move-cli/src/main.rs
```diff
@@ -775,7 +778,33 @@ fn main() -> Result<()> {
...
+            let files = ["args.txt"];
```
sblackshear, line 787: Might be useful to put a comment or two at the top of args.txt that explains what to do; e.g.,

```
# This is a batch file. To write an expected value test that runs `move <command`> <args1>; move <command2> <args2>, write <command1> <args1>\n<command2> <args2>`. # is a comment
```

sblackshear:
Ah, might make sense to add an example of this handy command to the testing section of the README too.

language/tools/move-cli/src/main.rs
tzakian: Good idea! I've added it in.

tzakian:
> Ah, might make sense to add an example of this handy command to the testing section of the README too.

Agreed! Added an example of using it in the README along with the files/directories that are created.

language/tools/move-cli/src/main.rs
tzakian: Good spot. I've updated. I've also moved this logic in to the `test` module since that seems to be a better place for it. I also added a check so that we don't overwrite an existing directory/file when create a scaffold.

PR #6971, opened 2020-12-17
[move cli] doctor command for detecting inconsistencies in storage

sblackshear:
This command does a bunch of sanity checks to ensure that the storage directory is well-formed. For now:
- All modules verify
- All modules link
- All resources deserialize w.r.t the current version of their declaring module

It does this in the obvious way: iterating through the global storage, collecting each module/resource, and doing the check.
This is useful for detecting breaking changes (both code and data layout) after-the-fact, as well as bugs in the CLI.

I was motivated to create this check when I noticed a bug where `move publish` sometimes publishes a module `M` without publishing library modules.
If you run `move doctor` on such a state, it would catch the issue. I am planning to add `move doctor` to a bunch of the publish tests once I fix this bug.
It will also be useful for testing the `--ignore-breaking-changes` flag in https://github.com/diem/diem/pull/6753.

### DIFF
--- a/language/tools/move-cli/src/lib.rs
+++ b/language/tools/move-cli/src/lib.rs
@@ -24,7 +24,7 @@ use vm::{
 
 use anyhow::{anyhow, bail, Result};
 use std::{
-    collections::HashMap,
+    collections::{BTreeMap, HashMap},
     convert::TryFrom,
     fs,
     path::{Path, PathBuf},
@@ -187,6 +187,8 @@ impl OnDiskStateView {
         })
     }
 
+    /// Returns a deserialized representation of the resource value stored at `resource_path`.
+    /// Returns Err if the path does not hold a resource value or the resource cannot be deserialized
     pub fn view_resource(&self, resource_path: &Path) -> Result<Option<AnnotatedMoveStruct>> {
         if resource_path.is_dir() {
             bail!("Bad resource path {:?}. Needed file, found directory")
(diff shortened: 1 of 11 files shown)


sblackshear:
@reviewer @mengxu-fb

language/tools/move-cli/src/lib.rs
```diff
@@ -325,6 +327,44 @@ impl OnDiskStateView {
...
+    pub fn module_paths(&self) -> impl Iterator<Item = PathBuf> + '_ {
+        self.iter_paths(move |p| self.is_module_path(p))
+    }
+
+    pub fn event_paths(&self) -> impl Iterator<Item = PathBuf> + '_ {
+        self.iter_paths(move |p| self.is_event_path(p))
+    }
+
+    /// Return a map of module ID -> module for all modules in the self.storage_dir.
+    /// Returns an Err if a module does not deserailize
```
mengxu-fb, line 355:
```suggestion
    /// Returns an Err if a module does not deserialize or if there are duplications
```

language/tools/move-cli/src/lib.rs
```diff
@@ -187,6 +187,8 @@ impl OnDiskStateView {
         })
     }
 
+    /// Returns a deserialized representatio of the resource value stored at `resource_path`.
```
mengxu-fb, line 190:
```suggestion
    /// Returns a deserialized representation of the resource value stored at `resource_path`.
```

sblackshear: speling is dificelt

mengxu-fb (approved):
LGTM! `pub fn event_paths()` are not used although it should go with check *(4) all events can be deserialized (TODO)*.

PR #6972, opened 2020-12-18
[move-prover] Adds cvc4 to tests

barrettcw:
This adds a set of unit tests to be run with cvc4 when running "cargo test"

### DIFF
--- a/language/move-prover/tests/testsuite.rs
+++ b/language/move-prover/tests/testsuite.rs
@@ -60,6 +60,30 @@ fn test_runner(path: &Path) -> datatest_stable::Result<()> {
         Err(err) => format!("Move prover returns: {}\n", err),
     };
     if baseline_valid {
+        if let Some(ref path) = baseline_path {
+            diags += &String::from_utf8_lossy(&error_writer.into_inner()).to_string();
+            verify_or_update_baseline(path.as_path(), &diags)?
+        } else if !diags.is_empty() {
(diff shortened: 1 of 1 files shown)


language/move-prover/tests/testsuite.rs
```diff
@@ -72,9 +96,94 @@ fn test_runner(path: &Path) -> datatest_stable::Result<()> {
             .into());
         }
     }
+
     Ok(())
 }
 
+fn cvc4blacklisted(path: &Path) -> bool {
```
wrwg, line 103: This could be done nicer if you like so. See the use of extract_test_directives in the `let baseline_valid = ...`. You could have `!extract_test_directive("no-cvc4")?` and then just have a comment `// no-cvc4` + explanations why in the source of the test. But I leave it to you, not sure how much time pressure there is for getting this done.

wrwg (approved):

language/move-prover/tests/testsuite.rs
DavidLDill: Clark can't land this, I think, because his affiliation with Facebook ended.  So, I'm going to try to land it and we can fix this up in a future PR.

wrwg: The PR workflow did not pass. I tried to rerun it yesterday because the failure looked like a flake, but it again failed. Otherwise I would have already landed.

PR #6974, opened 2020-12-18
[ct] update max child vasp in test

sausagee:
## Motivation
The max supported child account is increased to 2^16 in ab391a0bd1. Update the limit in cluster test setup to be in sync.

## Test Plan
Canary

### DIFF
--- a/testsuite/cluster-test/src/tx_emitter.rs
+++ b/testsuite/cluster-test/src/tx_emitter.rs
@@ -50,7 +50,7 @@ use tokio::{task::JoinHandle, time};
 
 const MAX_TXN_BATCH_SIZE: usize = 100; // Max transactions per account in mempool
                                        // Please make 'MAX_CHILD_VASP_NUM' consistency with 'MAX_CHILD_ACCOUNTS' constant under VASP.move
-const MAX_CHILD_VASP_NUM: usize = 256;
+const MAX_CHILD_VASP_NUM: usize = 65536;
 const MAX_VASP_ACCOUNT_NUM: usize = 16;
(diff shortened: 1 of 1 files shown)


snakech (commented):
I think simply change this number will not work, this number dependents on how many VASP accounts. Currently we only have 16 vasps for CT so this number can not be changed more than 256

sausagee:
> I think simply change this number will not work, this number dependents on how many VASP accounts. Currently we only have 16 vasps for CT so this number can not be changed more than 256

Maybe I have a misunderstanding of cluster test.  I thought cluster test creates up to `MAX_VASP_ACCOUNT_NUM` (currently set to 16) * MAX_CHILD_VASP_NUM accounts during testing.  Since the child account limit is bumped in the Diem framework, I supposed we can bump it in the testing, no?

snakech:
> > I think simply change this number will not work, this number dependents on how many VASP accounts. Currently we only have 16 vasps for CT so this number can not be changed more than 256
> 
> Maybe I have a misunderstanding of cluster test. I thought cluster test creates up to `MAX_VASP_ACCOUNT_NUM` (currently set to 16) * MAX_CHILD_VASP_NUM accounts during testing. Since the child account limit is bumped in the Diem framework, I supposed we can bump it in the testing, no?

oh, sorry its my fault misunderstood your pr... MAX_CHILD_VASP_NUM is previously hard limit of max number for each VASP account, if child account limit is bumped in Diem, I think it's ok to bump it to latest value in CT as well.

snakech (approved):

PR #6983, opened 2020-12-18
   [prover] trait for compositional interproc analysis, use for packed types analysis and usage analysis

sblackshear:
This PR introduces a `CompositionalAnalysis` trait that facilitates bottom-up interprocedural analysis. It uses the same mechanism for storing and retrieving summaries that UsageAnalysis already uses.
    
To demonstrate the generality of this trait, the PR splits the two analyses in usage_analysis.rs (packed types and memory usage) into separate files and uses the `CompositionalAnalysis` trait to implement both of them.
    
The motivation for this is making it easier to build other summary-based analyses that are flow-sensitive (e.g., fine-grained read/write set analysis to support transaction parallelization and AOS approvals, inference of post/aborts conditions, ...). There is more work to be done first, such as allowing the summary and abstract state to be different types, supporting recursion, decreasing the boilerplate `impl trait for`'s, and so on. But I think this is a good start.
    
Note: `CompositionalAnalysis` implements a flow-sensitive propagation, which is overkill for the flow-insensitive packed types and memory usage analyses. However, I measured the performance before/after conversion (benchmark: building packed types map for stdlib) and can't observe a difference. So I think it is ok to use CompositionalAnalysis everywhere, even on flow-insensitive summary-based analyses. We can always add a separate trait for flow-insensitive analyses if perf becomes a problem.


language/move-prover/bytecode/src/borrow_analysis.rs
```diff
@@ -250,7 +250,7 @@ impl FunctionTargetProcessor for BorrowAnalysisProcessor {
             BorrowAnnotation(BTreeMap::new())
         } else {
             let func_target = FunctionTarget::new(func_env, &data);
-            let mut analyzer = BorrowAnalysis::new(&func_target);
+            let analyzer = BorrowAnalysis::new(&func_target);
```
wrwg, line 253: The mut here was intended to allow the Analysis to accumulate some state, but I'm fine with removing it for now.

wrwg (approved):

PR #6987, opened 2020-12-18
[cluster-test] Added alert for DD account balance

snakech:
Added a warning for DD account if requested coins is higher than current balance

## Motivation

(Write your answer here.)

## Test Plan

## Related PRs

### DIFF
--- a/testsuite/cluster-test/src/tx_emitter.rs
+++ b/testsuite/cluster-test/src/tx_emitter.rs
@@ -408,10 +408,17 @@ impl TxEmitter {
         let balance = retrieve_account_balance(&client, faucet_account.address).await?;
         for b in balance {
             if b.currency.eq(XUS_NAME) {
-                info!(
-                    "DD account current balances are {}, requested {} coins",
-                    b.amount, coins_total
-                );
+                if b.amount < coins_total {
+                    warn!(
+                        "DD account current balance {} is low, requested {} coins",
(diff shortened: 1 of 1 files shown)


testsuite/cluster-test/src/tx_emitter.rs
```diff
@@ -408,10 +408,14 @@ impl TxEmitter {
         let balance = retrieve_account_balance(&client, faucet_account.address).await?;
         for b in balance {
             if b.currency.eq(XUS_NAME) {
-                info!(
-                    "DD account current balances are {}, requested {} coins",
-                    b.amount, coins_total
-                );
+                if b.amount < coins_total {
+                    warn!("DD account current balance {} is low, requested {} coins", b.amount, coins_total);
```
sausagee, line 412: or error!.  Will leave this to you.

sausagee (approved):

testsuite/cluster-test/src/tx_emitter.rs
snakech: we do not need error to crash the system, because if this is not the first time mint to child vasps, those vasps are reusable and I assume they already have some coins and balance is not 0. So for this case we only need to care about the first time mint to new child accounts

rexhoffman:
Failed lint so I bounced it out of the queue rather than let it continue for 30 minutes.

PR #6988, opened 2020-12-19
move: create temp files in the directory they are needed in

bmwill:
This is an alternate solution to the one proposed in #6809

### DIFF
--- a/language/move-lang/src/lib.rs
+++ b/language/move-lang/src/lib.rs
@@ -338,10 +338,11 @@ pub fn generate_interface_files(
             continue;
         }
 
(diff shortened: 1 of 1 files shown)


mengxu-fb (commented):
Thanks for the PR. I think the code LGTM. But I will leave it to @tnowacki for the final saying :)

tnowacki (approved):
I think I didn't do this because I didn't think about it.

These few lines of code have been refactored a few times, perhaps it was possible originally, but it definitely looks like the right thing to do now
Thanks!

PR #6990, opened 2020-12-19
[ci] use vm executors

sausagee:
## Summary
The bigger docker executors are not available due to billing issue.  This works around it by switching the affected jobs to the vm executors.  This should unblock the PR workflow and resume verify and land. 

## Test
CI in place

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -32,7 +32,7 @@ executors:
     docker:
       - image: *image
     resource_class: xlarge
-  premainnet-cluster-test-executor:
+  xlarge-build-executor:
     docker:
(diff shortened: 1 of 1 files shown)


rexhoffman:
makes sense though the xlarge machine is 8 cpu/ 32gig ram and the docker xlarge is 8 cpu 16 gig ram…. and since we’re ram constrained…..

I'll work on the CARGO_HOME to get sccache working for us.

rexhoffman (approved):

rexhoffman:
/land

unblocking for now.  hopefully we revert is access to 2xlarge machines is restored.

PR #6991, opened 2020-12-19
[file-format][bytecode-verifier] new bytecodes for vector operations

mengxu-fb:
Added the following new bytecodes to support basic vector operations:

- `VecPack(N)` Create a vector of length N by consuming the last N elements from stack
- `VecLen` Get the length of a vector
- `VecImmBorrow` Acquire an immutable reference to the element at a given index of the vector
- `VecMutBorrorw` Acquire a mutable reference to the element at a given index of the vector.
- `VecPushBack` Add an element to the end of the vector
- `VecPopBack` Pop an element from the end of vector
- `VecUnpack(N)` Destroy a vector of length N and unpack its elements onto the stack.
- `VecSwap` Swaps the elements at two indices in the vector

Several related bytecode verifiers are updated too, including

- Bounds checker: the signature index in the bytecode is within bounds
- Signature: the signature index in the bytecode corresponds to a single-token, non-reference signature
- Stack usage: the vector operations consumes and also produces the correct number of values on the stack
- Type safety: vector and the argument types are correct. For `VecPack(N)` the last N elements on the stack must have the element type.
- Reference safety: vector are in a correct imm/mut state when related operations happen 

Other misc but related updates include
- Added new bytecodes to the zero-cost gas schedule table
- Added new bytecodes to the initial gas schedule table, copying values from the native op table
- Proptest fuzzers are updated to produce valid and invalid inputs regarding these bytecode

## Motivation

Not all native functions are equal. Some native functions are more applicable in the Diem context while other are more generic and may be better supported at the language level. `Vector` is one example that may be promoted to have some support at the bytecode level.

A larger context that is considered is the separation of Move stdlib and Diem framework. Promoting native vector operations into bytecode will ease the separation.

As a small side benefit, this will also improve the user experience of using `Vector` in Move. At least we don’t have to write `use 0x1::Vector` all the time for basic vector operations. And also shrink `Vector.move` from `stdlib`.

Yes

## Test Plan

unit and prop tests are updated accordingly. `cargo xtest` should not fail.


language/vm/src/file_format.rs
```diff
@@ -1224,6 +1224,55 @@ pub enum Bytecode {
...
+    /// Stack transition:
+    ///
+    /// ```... -> ..., vec[]```
+    VecEmpty(SignatureIndex),
+    /// Return the length of the vector,
+    ///
+    /// Stack transition:
+    ///
+    /// ```..., vector_reference -> ..., u64_value```
+    VecLen,
```
sblackshear, line 1238: I think that we should include a `SignatureIndex` on all of these operations, not just `Vec::empty`.
- This makes the operations more consistent with each other 
- A bytecode-to-native compiler will have a much easier time compiling vector operations if it doesn't have to repeat the type inference of the bytecode verifier. Attaching the type to the operations makes it simple to allocate the appropriate amount of space
- It reduces our reliance on the magic of type inference in the bytecode verifier

language/bytecode-verifier/src/reference_safety/mod.rs
```diff
@@ -324,6 +324,62 @@ fn execute_inner(
             let struct_def = verifier.resolver.struct_def_at(struct_inst.def)?;
             unpack(verifier, struct_def)
         }
+
+        Bytecode::VecEmpty(idx) => {
+            // since reference safety is scheduled after type safety, we know that the element type
+            // must be valid, i.e., the unwrap is safe here.
```
sblackshear, line 330: @tnowacki/ @shazqadeer: do we `unwrap` or propagate an internal invariant violation in these sorts of cases elsewhere in the verifier?

language/bytecode-verifier/src/type_safety.rs
```diff
@@ -344,6 +353,36 @@ fn move_to(
...
+fn borrow_vector_element(
+    verifier: &mut TypeSafetyChecker,
+    offset: CodeOffset,
+    mut_: bool,
+) -> PartialVMResult<()> {
+    let operand_idx = verifier.stack.pop().unwrap();
+    let operand_vec = verifier.stack.pop().unwrap();
+
+    // check index
+    // TODO: shall we allow U8 and U128 (i.e., is_integer()) as index too?
```
sblackshear, line 365: I don't think so-- I think the index type should always be U64

language/vm/src/file_format.rs
tnowacki: Agreed

language/bytecode-verifier/src/reference_safety/mod.rs
tnowacki: We should never panic if possible (though that isn't always the case in the verifier right now)

Should raise a `StatusCode::VERIFIER_INVARIANT_VIOLATION`

language/bytecode-verifier/src/type_safety.rs
```diff
@@ -344,6 +353,36 @@ fn move_to(
...
+    if !matches!(operand_idx, ST::U64) {
```
tnowacki, line 366: != ST::U64 is a bit more canonical here

language/bytecode-verifier/src/type_safety.rs
```diff
@@ -696,6 +735,76 @@ fn verify_instr(
...
+                return Err(verifier.error(StatusCode::VECTOR_OP_TYPE_MISMATCH_ERROR, offset));
+            }
+        }
+
+        Bytecode::VecSwap => {
+            let operand_idx2 = verifier.stack.pop().unwrap();
+            let operand_idx1 = verifier.stack.pop().unwrap();
+            let operand_vec = verifier.stack.pop().unwrap();
+
+            // TODO: shall we allow U8 and U128 (i.e., is_integer()) as index too?
```
tnowacki, line 799: As mentioned above, no

language/bytecode-verifier/src/type_safety.rs
```diff
@@ -696,6 +735,76 @@ fn verify_instr(
...
+            if !matches!(operand_idx1, ST::U64) || !matches!(operand_idx2, ST::U64) {
```
tnowacki, line 800: != ST::U64

language/bytecode-verifier/src/type_safety.rs
```diff
@@ -741,3 +850,52 @@ fn instantiate(token: &SignatureToken, subst: &Signature) -> SignatureToken {
...
+    match ty {
+        U8 | U64 | U128 | Bool | Address | Signer | Struct(_) | TypeParameter(_) => Ok(()),
+        Reference(_) | MutableReference(_) => {
+            Err(PartialVMError::new(StatusCode::INVALID_VECTOR_ELEMENT_TYPE)
+                .with_message("reference not allowed as vector element type".to_string()))
+        }
+        Vector(ty) => check_vector_element_sig_token(ty),
+        StructInstantiation(_, type_arguments) => check_vector_element_sig_tokens(type_arguments),
+    }
+}
```
tnowacki, line 876: Should be done in instruction concistency, or signatures.rs

language/bytecode-verifier/src/type_safety.rs
```diff
@@ -741,3 +850,52 @@ fn instantiate(token: &SignatureToken, subst: &Signature) -> SignatureToken {
...
+
+fn get_vector_element_type(vector_ref_ty: SignatureToken, mut_: bool) -> Option<SignatureToken> {
```
tnowacki, line 878: You will likely need to rework this once every vector operation takes a signature, so I will leave comments until that refactor

language/vm/src/file_format.rs
```diff
@@ -1224,6 +1224,55 @@ pub enum Bytecode {
...
+    /// Stack transition:
+    ///
+    /// ```..., vector_reference -> ..., element```
+    VecPopBack,
+    /// Destroy the vector. Aborts if the vector is not empty
+    ///
+    /// Stack transition:
+    ///
+    /// ```..., vec[] -> ...`
+    VecDestroy,
```
tnowacki, line 1270: I think we should name this `VecDestroyEmpty`

language/bytecode-verifier/src/reference_safety/mod.rs
```diff
@@ -324,6 +324,62 @@ fn execute_inner(
...
+        Bytecode::VecLen => {
+            let vec_ref = verifier.stack.pop().unwrap();
+            state.vector_op(offset, vec_ref, false)?;
+            verifier.stack.push(state.value_for(&SignatureToken::U64));
+        }
+
+        Bytecode::VecImmBorrow => {
+            checked_verify!(verifier.stack.pop().unwrap().is_value());
+            let vec_ref = verifier.stack.pop().unwrap();
+            state.vector_element_borrow(offset, vec_ref, false)?;
```
tnowacki, line 352: We might want to relax this should be relaxed to match `ImmBorrowField` in that it should also work over mutable vector references.

That being said... the relaxation doesn't do anything here. There are times where ImmBorrowField is valid but freeze is not, hence that relaxed rule. Thoughts, @shazqadeer @sblackshear ?

tnowacki:
As a general thing, we should keep this PR up to date, but it probably shouldn't go in until the ability changes are done

mengxu-fb:
This PR will need some rebasing but I'll keep it up to date from now on :)

language/bytecode-verifier/src/reference_safety/mod.rs
sblackshear: I'm not sure I understand "the relaxation doesn't do anything here"...can you elaborate?

tnowacki: `ImmBorrowField` can work over mutable or immutable references. This is due to adding expressibility where a given field can be borrowed immutably, but the parent ref cannot be frozen. 

As written here, `VecImmBorrow` can work only over immutable references

mengxu-fb:
Rebased on to latest master and will keep it udpated.

language/bytecode-verifier/src/type_safety.rs
```diff
@@ -696,6 +735,76 @@ fn verify_instr(
             let type_args = verifier.resolver.signature_at(struct_inst.type_parameters);
             move_to(verifier, offset, struct_def, type_args)?
         }
+
+        Bytecode::VecEmpty(idx) => {
+            let element_sig = verifier.resolver.signature_at(*idx);
+            if element_sig.len() != 1 {
+                return Err(verifier.error(StatusCode::INVALID_VECTOR_ELEMENT_TYPE, offset));
+            }
```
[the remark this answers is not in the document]
mengxu-fb: Check relocated to `signature.rs`, `instruction_consistency.rs` seems to be dealing with instructions that have both a generic and non-generic versions, so I guess `signature.rs` could be a better place.

language/bytecode-verifier/src/type_safety.rs
mengxu-fb: Check relocated to `signature.rs`, and hopefully, using the ability checking in the correct way.

language/bytecode-verifier/src/type_safety.rs
mengxu-fb: Done with the change that every vector operation takes a signature.

language/bytecode-verifier/src/signature.rs
```diff
@@ -156,7 +156,75 @@ impl<'a> SignatureChecker<'a> {
...
+                | ImmBorrowLoc(_)
+                | MutBorrowField(_)
+                | MutBorrowFieldGeneric(_)
+                | ImmBorrowField(_)
+                | ImmBorrowFieldGeneric(_)
+                | MutBorrowGlobal(_)
+                | ImmBorrowGlobal(_)
+                | Exists(_)
+                | MoveTo(_)
+                | MoveToGeneric(_)
```
mengxu-fb, line 224: Question: is there a reason why `MoveToGeneric`, `MutBorrowFieldGeneric`, `ImmBorrowFieldGeneric` were not checked originally?

language/bytecode-verifier/src/signature.rs
```diff
@@ -241,6 +309,15 @@ impl<'a> SignatureChecker<'a> {
         constraints: &[AbilitySet],
         global_abilities: &[AbilitySet],
     ) -> PartialVMResult<()> {
+        if type_arguments.len() != constraints.len() {
```
mengxu-fb, line 312: Question: I don't seem to find where this check was performed... Did I miss anything in the bounds check or in the signature check?

language/vm/src/file_format.rs
mengxu-fb: Shall we do `VecEmpty` -> `VecCreateEmpty`?

language/bytecode-verifier/src/signature.rs
vgao1996: Looks like a bug to me. Good catch!

mengxu-fb: I can put another PR to address this then. But let me know if you are already working on this.

vgao1996: Go ahead and put up a PR! I'm not currently touching this part of the codebase so no need to worry about conflicts.

davidiw:
Old PR, feel free to re-open.

mengxu-fb:
Reviving the work here

language/bytecode-verifier/src/reference_safety/mod.rs
mengxu-fb: Updated, now both `VecImmBorrow` and `VecLen` accept a mutable reference of a vector type as well.

mengxu-fb:
Support in the Move VM is ready in PR #8821

mengxu-fb:
Bumped the bytecode format version to 4 since the branch has cut, hence the last two commits.

language/move-ir/types/src/ast.rs
```diff
@@ -392,6 +392,23 @@ pub enum Builtin {
...
+    /// Acquire an immutable reference to the element at a given index of the vector
+    VecImmBorrow(Vec<Type>),
+    /// Acquire a mutable reference to the element at a given index of the vector
+    VecMutBorrow(Vec<Type>),
+    /// Push an element to the end of the vector
+    VecPushBack(Vec<Type>),
+    /// Pop and return an element from the end of the vector
+    VecPopBack(Vec<Type>),
+    /// Destroy an empty vector
+    VecDestroyEmpty(Vec<Type>),
```
vgao1996, line 408: IIRC @tnowacki mentioned the idea that instead of having `VecDestroyEmpty`, we add a more generalized `VecDestroyN`. Do we wanna consider this right now, or do we want to do it after the current implementation lands?

mengxu-fb: I like the idea of `VecPack(n)` and `VecUnpack(n)`, but my personal preference would be to implement it in a separate PR.

tnowacki: Could you just remove VecDestroyEmpty and VecEmpty from this PR then? feels weird to land them just to change them

mengxu-fb: Hmm, that can be a bit ugly I guess. Removing `VecDestroyEmpty` and `VecEmpty` will break some of the IR tests because these tests depends on `VecEmpty` to produce a valid vector.

tnowacki: Then you can continue to use the native functions no? Those aren't going anywhere anytime soon

mengxu-fb: Yes, but need to additionally import the `0x1.Vector;`. Not a big deal though, but just some extra work. If this is the route, I think maybe adding `VecPack(n)` and `VecUnpack(n)` in this PR would make more sense.

tnowacki: Either way is fine by me :) I think the size of the review will be roughly the same either way so whatever is easier for you

mengxu-fb: I would prefer to have this landed first :) It has been here for so long and rebased for so many times :)

tnowacki: Sure... I get that, just very awkward to add bytecode operations just to remove them.

mengxu-fb: Yup, I understand the awkwardness. I owe a big thank you to you and whomever else that helped reviewing this giant PR.

mengxu-fb: @tnowacki @vgao1996 This PR is now ported to using the `VecPack(type, num)` and `VecUnpack(type, num)` bytecode (the porting is done in the last three commits).

`VecEmpty` and `VecDestroyEmpty` are gone as they are now just special cases of `VecPack(type, 0)`, `VecUnpack(type, 0)` respectively.

tnowacki (dismissed):
Awesome! I think you will have to fix some tests, but let's roll with it!

mengxu-fb:
The forge-tests are not passing but for reasons I don't think is related to this PR:
```
rpc error: error sending request for url (http://localhost:42985/metrics): error trying to connect: tcp connect error: Cannot assign requested address (os error 99)
```

I started to see this same error in my other unrelated PRs as well and they don't seem to block landing, so I'll give landing a try.

/land

mengxu-fb:
Ah, the cluster test is failing because we bumped up the file format version number. I need to fix that.

mengxu-fb:
The latest rebase
- distributed the ir-based tests into `bytecode-verifier/tests` and `move-vm/tests` respectively, using the new transactional testing infra
- unbumped the file version number from 4 to 3, because the compatibility test always uses a genesis at the tip of this PR, which a version 4 bytecode cannot be deserialized properly by the base version before this PR. We may need to work out a testing plan for this change in the future.

tnowacki (approved):

PR #7001, opened 2020-12-22
[json-rpc] fixing flaky test

mimoo:
There are cases where the state will change in between queries to the full node,
this is not the case if we batch the queries.

### DIFF
--- a/json-rpc/json-rpc-spec.md
+++ b/json-rpc/json-rpc-spec.md
@@ -64,7 +64,11 @@ You can use this information to verify liveness / status of nodes in the network
 
 The JSON-RPC protocol allows requests to be batched. An arbitrary number of requests (maximum 20 by default) can be combined into a single batch and submitted to the server. These requests will be processed together under a single request context.
 
+For example:
 
+```
+curl -X POST -H "Content-Type: application/json" --data '[{"jsonrpc":"2.0","method":"get_metadata","params":[1],"id":1},{"jsonrpc":"2.0","method":"get_metadata","params":[9],"id":2}]' "https://client.testnet.diem.com/"
+```
 ### Errors
 
 If errors occur during a request, they are returned in an error object, as defined in: [https://www.jsonrpc.org/specification#error_object](https://www.jsonrpc.org/specification#error_object).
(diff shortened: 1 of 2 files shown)


libra-action:

Breaking changes policy:

2. Adding fields is ok.

json-rpc/json-rpc-spec.md
```diff
@@ -64,7 +64,11 @@ You can use this information to verify liveness / status of nodes in the network
 
 The JSON-RPC protocol allows requests to be batched. An arbitrary number of requests (maximum 20 by default) can be combined into a single batch and submitted to the server. These requests will be processed together under a single request context.
 
+For example:
 
+```
+curl -X POST -H "Content-Type: application/json" --data '[{"jsonrpc":"2.0","method":"get_metadata","params":[1],"id":1},{"jsonrpc":"2.0","method":"get_metadata","params":[9],"id":2}]' "https://client.testnet.libra.org/"
+```
```
xli, line 71: diem.com

xli (approved):

json-rpc/json-rpc-spec.md
mimoo: ooch ^^

json-rpc/tests/integration_test.rs
```diff
@@ -81,23 +81,40 @@ fn create_test_cases() -> Vec<Test> {
...
                 assert_eq!(metadata["module_publishing_allowed"], true);
                 assert_eq!(metadata["diem_version"], 1);
                 assert_eq!(metadata["dual_attestation_limit"], 1000000000);
-                assert_ne!(resp.diem_ledger_timestampusec, 0);
-                assert_ne!(resp.diem_ledger_version, 0);
+                assert_ne!(diem_ledger_timestampusec, 0);
+                assert_ne!(diem_ledger_version, 0);
 
                 // prove the accumulator_root_hash
-                let sp_resp = env.send("get_state_proof", json!([resp.diem_ledger_version]));
```
AnomalRoil, line 99: Actually this line was handling the "state can change inbetween" since it receives the "last known version from client" as being `resp.diem_ledger_version`.
So this is a unnecessary PR.

AnomalRoil: But there's a problem here since we check the provided proof against `metadata["accumulator_root_hash"]` below without batching the queries, agreed.

PR #7002, opened 2020-12-22
[mempool][easy] fix node creation in tests

mimoo:


mempool/src/tests/shared_mempool_test.rs
```diff
@@ -204,8 +205,10 @@ impl SharedMempoolNetwork {
         let mut smp = Self::default();
         let mut peers = vec![];
 
-        for _ in 0..validator_nodes_count {
-            let mut config = NodeConfig::random();
+        let mut rng = StdRng::from_seed([0u8; 32]);
+        for idx in 0..validator_nodes_count {
+            let mut config =
+                NodeConfig::random_with_template(idx as u32, &NodeConfig::default(), &mut rng);
```
davidiw, line 211: would we benefit from adding a new NodeConfig::new random_in_set(idx: u32) or is this the only place we use this?

mempool/src/tests/shared_mempool_test.rs
```diff
@@ -237,7 +240,12 @@ impl SharedMempoolNetwork {
         let full_node = PeerId::random();
 
         // validator config
-        let mut config = NodeConfig::random();
+        let mut rng = StdRng::from_seed([0u8; 32]);
```
davidiw, line 243: I don't believe this is necessary and it conflicts with how we've setup other networks (to some extent).

mempool/src/tests/shared_mempool_test.rs
mimoo: So, in all other places where we use `NodeConfig::random()` we only use it once, which is fine. If you use it twice, it ends up creating two nodes with the same keys and addresses. This doesn't look like what we want to test in these situations, so I fixed it. If you're talking about removing the `NodeConfig::default()` argument then we can't because it is used in one other place with an actual argument there. If you're talking about removing the `rng` argument, we could but this then would produce different keys every time we run this test. (Probably not be a big deal in this scenario, but if you're debugging it can be annoying to see different values every time.)

mempool/src/tests/shared_mempool_test.rs
mimoo: this is actually what the `NodeConfig::random()` does, I just make sure that the two calls in this function produce two different node configs instead of the same one.

davidiw: I mean I don't think this change is necessary -- was this test failing?

mempool/src/tests/shared_mempool_test.rs
davidiw: I agree, my question was does it make sense to make something that would automatically get you the ith configuration. Looking at the code, it looks like probably not as you would create an O(n) function called n times :D

mimoo: Not sure what type of function you have in mind, but I could create a self-updatable closure which I think would be a good idea to avoid someone forgetting to increase the idx : o

mempool/src/tests/shared_mempool_test.rs
mimoo: Yeah this test is flaky, see https://github.com/diem/diem/pull/6826#issuecomment-749220605

I'll leave it up to you if you think it's not necessary :) I saw that as an easy fix of something that didn't look right.

davidiw (approved):

PR #7008, opened 2020-12-22
[move-prover] Allow to manage different classes of tests in `./tests`

wrwg:
This modifies the test runner to support a new category of "extended" tests under `./tests/xsources`.  Those tests aren't run by default or under CI when `cargo test` is called. Instead, to run those tests, one uses `MVP_TEST_X=1 cargo test`.

## Motivation

Performance testing

## Test Plan

cargo test

## Related PRs

NA


emmazzz (approved):
LGTM

PR #7010, opened 2020-12-22
[move-cli] allow move doctor to pickup the library dependencies

mengxu-fb:
## Motivation

Currently the `move doctor` command is not aware of the library dependencies when checking modules. This PR fills the information.

Yes

## Test Plan

A new test case (credit to Sam) is added.

### DIFF
--- a/language/tools/move-cli/src/main.rs
+++ b/language/tools/move-cli/src/main.rs
@@ -169,20 +169,41 @@ impl Move {
     /// If `source_only` is true, only the source files will be populated. The modules will
     /// not be compiled nor loaded.
     ///
-    /// Currently, `source_only` is set to
-    ///  - true for "check" and "publish" and
-    ///  - false for "run" and "doctor"
-    fn prepare_mode(&self, source_only: bool) -> Result<()> {
-        self.mode.prepare(&self.get_package_dir(), source_only)
+    /// Currently, `source_only` is set to true for `check` only and false for `publish` and `run`
+    pub fn prepare_mode(&self, source_only: bool) -> Result<()> {
+        self.mode.prepare(&self.get_package_dir(), source_only)?;
+
+        // prepare the storage if are not only caring about the source files
+        if !source_only {
+            self.preload_storage()?;
+        }
+
+        Ok(())
     }
 
-    /// This collect the dependencies for compiling a script or module. The dependencies
-    /// include not only the loaded libraries, but also the interface files generated from
-    /// prior "publish" commands.
-    fn get_compilation_deps(&self) -> Result<Vec<String>> {
-        let mut src_dirs = self.mode.source_files(&self.get_package_dir())?;
-        src_dirs.push(interface_files_dir(&self.build_dir)?);
-        Ok(src_dirs)
+    /// Generate interface files for published files
+    pub fn generate_interface_files(&self) -> Result<()> {
+        move_lang::generate_interface_files(
+            &[self.storage_dir.clone()],
+            Some(self.build_dir.clone()),
+            false,
+        )?;
+        Ok(())
+    }
+
+    pub fn interface_files_dir(&self) -> Result<String> {
+        let mut path = PathBuf::from(self.build_dir.clone());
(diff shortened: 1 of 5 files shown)


language/tools/move-cli/tests/testsuite/doctor_with_stdlib/src/modules/M.move
```diff
@@ -0,0 +1,10 @@
+address 0x2 {
+module M {
+    use 0x1::Debug;
+    struct Coin { f: u64, g: u64 }
```
sblackshear, line 4: Nit: simplify test by removing `Coin`?

language/tools/move-cli/tests/testsuite/doctor_with_stdlib/src/modules/M.move
```diff
@@ -0,0 +1,10 @@
...
+
+    fun dangle_after_move(_c: Coin) {
```
sblackshear, line 6: Nit: `f` or some other name? I think this silly name came from my repro, which was not very well minified :)

sblackshear (changes requested):
I think the problem here is with `publish`, not with `doctor`. IMO, the CLI should always create a `storage` directory that will pass bytecode verification and linking in isolation (e.g., without looking at `build` or anything else outside of storage). The only exception should be when `--ignore-breaking-changes` is used).

mengxu-fb:
I see. Yeah, having a self-contained `storage` makes sense. This probably means some additional changes on how the library packages are splashed over. I'll get back with an update. I'll minimize the test case as well :)

mengxu-fb:
Updated the `prepare_mode()` function so that besides splashing the source files and compiling the modules from libraries, it also saves library modules to `storage` if those libraries modules do not exist yet.

language/tools/move-cli/src/main.rs
```diff
@@ -189,6 +212,25 @@ impl Move {
     fn get_library_modules(&self) -> Result<Vec<CompiledModule>> {
         self.mode.compiled_modules(&self.get_package_dir())
     }
+
+    /// Preload the storage with library modules (if such modules are not available in the storage)
+    fn preload_storage(&self) -> Result<()> {
+        let storage_dir = maybe_create_dir(&self.storage_dir)?;
+        let view = OnDiskStateView::create(storage_dir.to_path_buf(), &[])?;
```
sblackshear, line 219: I worry that the `prepare_mode`/`preload_storage` flow is getting a bit too imperative--seems easy to set it up on one control-flow path but not another, or to accidentally set it up multiple times.

Maybe we could make things a bit more functional by changing `OnDiskStateView` to take a `mode` + do what `preload_storage` is doing here (or something else)? We can also address this in a follow-up if  it's too big for here.

language/tools/move-cli/src/main.rs
```diff
@@ -377,7 +401,12 @@ fn publish(
         }
     }
 
-    OnDiskStateView::create(storage_dir.to_path_buf(), &modules)
+    let state = OnDiskStateView::create(storage_dir.to_path_buf(), &modules)?;
+    if state.save_modules()? {
+        args.generate_interface_files()?;
```
sblackshear, line 406: Similar comment re: imperativeness here--it seems like we now want to call `generate_interface_files` every time we do `save_modules`, might want to make it impossible to forget to do that.

language/tools/move-cli/src/main.rs
mengxu-fb: Yes, this is a bit concerning. Shall I suggest to do it in a follow-up PR as this will be addressing a separate issue.

language/tools/move-cli/src/main.rs
mengxu-fb: Same as above, will do it in a separate PR.

language/tools/move-cli/src/main.rs
sblackshear: Sounds good!

sblackshear (approved):
Thanks for fixing this!

mengxu-fb:
Rebased

PR #7011, opened 2020-12-22
Remove NonZeroUsize from most places in the code

gregnazario:
## Motivation

After looking through the code base, we're using `NonZeroUsize` everywhere just because we're instantiating the `diem-channel` a ton of times.  Now, we can cut that out and just pass in the `usize` by itself, cutting down on a lot of unnecessary boilerplate.

### DIFF
--- a/common/infallible/src/nonzero.rs
+++ b/common/infallible/src/nonzero.rs
@@ -1,0 +1,35 @@
+// Copyright (c) The Diem Core Contributors
+// SPDX-License-Identifier: Apache-2.0
+
+/// A wrapper around `std::num::NonZeroUsize` to no longer worry about `unwrap()`
+#[macro_export]
+macro_rules! NonZeroUsize {
+    ($num:expr) => {
+        NonZeroUsize!($num, "Must be non-zero")
+    };
+    ($num:expr, $message:literal) => {
+        std::num::NonZeroUsize::new($num).expect($message)
+    };
+}
+
+#[cfg(test)]
+mod tests {
+
+    #[test]
+    fn test_nonzero() {
+        assert_eq!(1, NonZeroUsize!(1).get());
+        assert_eq!(std::usize::MAX, NonZeroUsize!(std::usize::MAX).get());
+    }
+
+    #[test]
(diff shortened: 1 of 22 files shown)


davidiw (approved):
good cleanup

common/infallible/src/nonzero.rs
```diff
@@ -0,0 +1,35 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+/// A wrapper around `std::num::NonZeroUsize` to no longer worry about `unwrap()`
+#[macro_export]
+macro_rules! NonZeroUsize {
+    ($num:expr) => {
+        NonZeroUsize!($num, "Must be non-zero")
+    };
+    ($num:expr, $message:literal) => {
+        std::num::NonZeroUsize::new($num).expect($message)
```
phlip9, line 11: one big reason for `nonzero_ext` is to get compile time proof that a literal is nonzero. you can see the assertion here: https://docs.rs/nonzero_ext/0.2.0/src/nonzero_ext/lib.rs.html#311

that way you can write `nonzero!(123)` without any unwraps because it checks that the compile-time constant is nonzero. if you write `nonzero!(0)` it will error at compile time.

here we don't communicate that and instead just silently panic inside at runtime, which seems suboptimal.

bmwill: Yeah. I think that this change is more a step backward in terms of type safety.

gregnazario: As I said earlier in Slack, if you look at where these are being used, its either A: a test, which is not that big of a difference, except the tests have to run to fail.  Or B: With information loaded from a Config, which is not possible to error at compile time, because it at startup of runtime.

I get your concern that this isn't quite as quick at telling us if a number is NonZero.  But it's added a lot of boilerplate in here.  Let's talk outside this.

davidiw: Good arguments.

* Unix has /dev/null but we don't as we don't allow zero sized queues
* The proposal as it stands would pull in another dependency for just a macro :/

I do love type safety so it might be worth bringing back something in the future if folks really want to enforce non-zero queues.

PR #7013, opened 2020-12-23
[move cli] finish move doctor by detecting event breaking changes + a…

sblackshear:
…dding docs

Previously, `move doctor` checked that resources deserialize correctly, but not events. This PR adds the event check + updates the README with documentation for breaking changes detection via `move publish` and `move doctor`. Finally, added a `doctor` command to the diem smoke test as a stress-test for `doctor`.

PR #7031, opened 2020-12-23
[move-cli] simplify the mode preparation and module saving logic for OnDiskStateView

mengxu-fb:
## Motivation

This PR is a follow-up of #7010 to further simplify the code. In particular, it does four simplifications

1. The logic for saving modules to the `OnDiskStateView` is simplified to one function call. The old flow is
```
let state = OnDiskStateView::create(..., &[modules])?;
if state.save_modules()? {
    generate_interface_files(...)?;
}
```
Which is repeated in several places.

The new flow is `state.save_modules(&[modules], Some(build_dir))?`;

2. Simplified `OnDiskStateView` creation. The old flow is
```
let storage_dir = maybe_create_dir(&args.storage_dir)?.canonicalize()?;
let state = OnDiskStateView::create(storage_dir, /* compiled_modules */ ...)?;
```
Now we can simplify create it with `let state = OnDiskStateView::create(&args.storage_dir)?;`.

3. Removed the `modules` and `resources` field in the `OnDiskStateView`. `resources` are not used in the old flow, and `modules` becomes stale after the simplification in 1.

4. Simplified mode preparation. The `source_only` option is dropped, as it does not seem to be useful. And the procedure for preloading the library modules to on-disk storage is also inlined into the `prepare_mode()` function.

This PR also fixes several comments which are not consistent with the code.

Yes

## Test Plan

`cargo xtest`

### DIFF
--- a/language/tools/move-cli/src/lib.rs
+++ b/language/tools/move-cli/src/lib.rs
@@ -24,7 +24,7 @@ use vm::{
 
 use anyhow::{anyhow, bail, Result};
 use std::{
-    collections::{BTreeMap, HashMap},
+    collections::BTreeMap,
     convert::TryFrom,
     fs,
     path::{Path, PathBuf},
@@ -60,33 +60,20 @@ const EVENTS_DIR: &str = "events";
 
 #[derive(Debug)]
 pub struct OnDiskStateView {
-    modules: HashMap<ModuleId, Vec<u8>>,
-    resources: HashMap<(AccountAddress, StructTag), Vec<u8>>,
     storage_dir: PathBuf,
 }
 
 impl OnDiskStateView {
-    /// Create an `OnDiskStateView` that reads/writes resource data in `storage_dir` and can
-    /// execute code in `compiled_modules`.
-    pub fn create(storage_dir: PathBuf, compiled_modules: &[CompiledModule]) -> Result<Self> {
-        if !storage_dir.exists() || !storage_dir.is_dir() {
-            bail!(
-                "Attempting to create OnDiskStateView from bad data directory {:?}",
-                storage_dir
-            )
+    /// Create an `OnDiskStateView` that reads/writes resource data and modules in `storage_dir`.
+    pub fn create<P: Into<PathBuf>>(storage_dir: P) -> Result<Self> {
+        let storage_dir = storage_dir.into();
+        if !storage_dir.exists() {
+            fs::create_dir_all(&storage_dir)?;
         }
-
-        let mut modules = HashMap::with_capacity(compiled_modules.len());
-        for module in compiled_modules {
-            let mut module_bytes = vec![];
-            module.serialize(&mut module_bytes)?;
-            modules.insert(module.self_id(), module_bytes);
-        }
-        let resources = HashMap::new();
         Ok(Self {
-            modules,
-            resources,
-            storage_dir,
+            // it is important to canonicalize the path here because `is_data_path()` relies on the
+            // fact that storage_dir is canonicalized.
+            storage_dir: storage_dir.canonicalize()?,
         })
     }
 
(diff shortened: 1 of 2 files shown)


language/tools/move-cli/src/main.rs
```diff
@@ -842,15 +797,15 @@ fn main() -> Result<()> {
             source_files,
             no_republish,
         } => {
-            move_args.prepare_mode(true)?;
+            move_args.prepare_mode()?;
```
sblackshear, line 800: It looks like almost every command (except `test` and `clean`) does at least one of the following two things:
- call `prepare_mode` (which creates an `OnDiskStateView`)
- create a subsequent `OnDiskStateView`

In light of this, might make sense to create a single `OnDiskStateView` before the `match` (e.g., it could be returned by `prepare_mode`) and pass it to each of the relevant command functions?

sblackshear (approved):
Excellent simplifications!

language/tools/move-cli/src/lib.rs
```diff
@@ -315,22 +299,50 @@ impl OnDiskStateView {
...
-    /// Save all the modules in the local cache
-    /// Returns true if any modules were saved
-    pub fn save_modules(&self) -> Result<bool> {
-        for (id, bytes) in &self.modules {
-            self.save_module(id, &bytes)?
+    /// Save all the modules in the local cache, re-generate mv_interfaces if required.
+    pub fn save_modules<P: Into<PathBuf>>(
+        &self,
+        modules: &[CompiledModule],
+        build_dir_opt: Option<P>,
```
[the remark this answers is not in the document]
mengxu-fb: It would definitely simplify the code and did come into my mind. The only reason I did not do this in the PR is that I kind of have a mental model of `storage_dir` == `OnDiskStateView` mapping while adding `build_dir` to `OnDiskStateView` also means that the `OnDiskStateView` needs to manage the things produced in the build process (i.e., the mv_interfaces).

language/tools/move-cli/src/main.rs
mengxu-fb: It makes sense! I'll make an update.

mengxu-fb: Hmm.... having a second thought now... It seems that `test`, `clean`, and `view` do not involve any dependent libraries so we might not want to create a `storage` directory and splash the library files there.

But despite of this, it still makes sense to have `prepare_mode()` return an `OnDiskStateView` and pass it to other handlers (i.e., `publish`, 'run`, and `doctor`).

sblackshear: Nice, that seems like a good middle ground.

language/tools/move-cli/src/lib.rs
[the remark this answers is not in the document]
mengxu-fb: In fact, there are quite some benefits by merging `build_dir` into `OnDiskStateView`, not only here for the code simplification, but but also for avoid carrying the `args: &Move` (which is a `structopt`) all over the places.

Currently the `args:&Move` is passed to all functions because we need the information of `args.verbose` and `args.build_dir`. Now with commit e198fd7, `OnDiskStateView` knows `build_dir` and `verbose` is passed in as an additional function argument. We no longer need to carry `args:&Move` everywhere.

This makes an important distinction. In Move CLI, only `args:&Move` can `prepare_state()` and to get an `OnDiskStateView`. Given that `args:&Move` disappears before control-flow gets into the command handlers, there is no way re-create another `state` when running the command handlers (e.g., `publish`, `run`, `check`, etc). Hence, reducing the chance of messing up the `storage`.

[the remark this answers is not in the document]
sblackshear: Good points!

mengxu-fb:
Rebased past #7013 with conflicts resolved

PR #7039, opened 2020-12-24
Add constructor and getters for GeneralMetadataV0 as inner fields are private

riyazdf:
## Motivation
I was following along with the DIP-4 source snippets and realized that the samples as-written do not work because the inner fields for the metadata structs are private. From discussion with @sblackshear and @n4ss, we agreed that a constructor for this data (and perhaps we should also consider the same for Travel Rule metadata, etc) would be useful. An alternative that I'm fine with is to make the fields `pub`, but I defer to the maintainers if they have a preference :) 

✅ 

## Test Plan
Added simple unit test, and ran:
`cargo x lint && cargo xfmt && cargo xclippy --all-targets`

### DIFF
--- a/types/src/transaction/metadata.rs
+++ b/types/src/transaction/metadata.rs
@@ -42,6 +42,32 @@ pub struct GeneralMetadataV0 {
     referenced_event: Option<u64>,
 }
 
+impl GeneralMetadataV0 {
+    pub fn new(
+        to_subaddress: Option<Vec<u8>>,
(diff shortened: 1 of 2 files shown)


sblackshear (approved):
LGTM!

n4ss (approved):
Awesome! Sorry that's my bad that we were missing this :< Is there a way for the code generation to detect private VS public fields @matbd ?

PR #7057, opened 2020-12-28
[move-prover/spec-lang] Enable identifiers as pragma values.

wrwg:
This PR enables pragmas and properties to have identifiers as values, in addition to literals. For example:

```
spec fun f {
  pragma fully_qualified = 0x1::Genesis::init; // This is fine even if not declared
  pragma simple_qualified = Genesis::init; // This will result in move-lang error if Genesis is not imported
  pragma simple = init; // This is fine even if not declared
}
```

## Motivation

Prepares for being able to implement `pragma called_only_by = <ident>`

Yes

## Test Plan

Added test

## Related PRs

NA


language/move-prover/spec-lang/tests/sources/pragmas_err.move
```diff
@@ -0,0 +1,22 @@
+module M {
+    spec module {
+        // Should be fine
+        pragma verify = false;
+
+        // Should produce error that bar is not valid (not defined)
+        pragma bar = true;
+
+        // Should be ok (we do not check types of values)
+        pragma verify = a_valid_id;
```
emmazzz, line 10: It seems that we allow `pragma verify = false;` and `pragma verify = a_valid_id;` to appear together in a spec block. Maybe it would be less confusing if we generate an error saying something like "pragma verify has conflicting types".

emmazzz: Reading the code again I see that we might not be able to check if they are of the same type since we don't know the type of `a_valid_id`, which could be a boolean as well.

emmazzz (approved):
LGTM

language/move-prover/spec-lang/tests/sources/pragmas_err.move
wrwg: Pragmas aren't type checked in general. The only thing we check is if the name is a known pragma and valid in the given context. Adding pragma type checking would be an independent, larger change than this one.

emmazzz: Another thing: this only allows the pragma value to be one identifier right? What if we want to say some function is called only from both f and g?

wrwg: That won't be possible for now. We can add it later, but perhaps the full Move-based solution might be earlier here than expected.

PR #7060, opened 2020-12-28
[state-sync][bug-fix] removing integer overflow

mimoo:


state-synchronizer/src/tests/helpers.rs
```diff
@@ -131,11 +131,14 @@ impl ExecutorProxyTrait for MockExecutorProxy {
         limit: u64,
         target_version: u64,
     ) -> Result<TransactionListWithProof> {
+        let start_version = known_version
```
JoshLind, line 134: Weird that we have to fix this in the test code.. I'm guessing it's because the fuzzer issue found was actually here (in test code) as opposed to non-test code, but the issue that it highlighted is more generic than just test code? 😄

state-synchronizer/src/executor_proxy.rs
```diff
@@ -182,8 +182,11 @@ impl ExecutorProxyTrait for ExecutorProxy {
         limit: u64,
         target_version: u64,
     ) -> Result<TransactionListWithProof> {
+        let start_version = known_version
+            .checked_add(1)
+            .ok_or_else(|| format_err!("Known version too high"))?;
```
JoshLind, line 187: nit: maybe "Known version has overflown"?

JoshLind (approved):
LGTM, thanks @mimoo! 😄

JoshLind:
FWIW, I'll follow this up with another PR that handles more of these cases in state sync soon.

state-synchronizer/src/tests/helpers.rs
mimoo: yeah I wrote that in the report, the issue was in the test code, but looking at the real code it's there too : (

state-synchronizer/src/executor_proxy.rs
mimoo: it hasn't yet at this point, its value is still `u64::max`, which is a-normally high :D

PR #7061, opened 2020-12-28
[move-lang] move the shadow compilation logic from move-cli to source compiler

mengxu-fb:
## Motivation

The feature to allow Move module compilation with duplicating/conflicting definitions against library modules is experimented in Move CLI. But ideally, this feature should live in the source compiler. This PR moves the logic there.

It also simplifies the filtering logic by only looking at the `parser::ast::Program` instead of going through the `mv_interfaces` directory with filename matching.

Yes

## Test Plan

`cargo xtest`

### DIFF
--- a/language/move-lang/src/bin/move-check.rs
+++ b/language/move-lang/src/bin/move-check.rs
@@ -45,14 +45,14 @@ pub struct Options {
     )]
     pub out_dir: Option<String>,
 
-    /// If set, allow modules defined in source_files to shadow modules of the same id that exist
-    /// in dependencies
+    /// If set, do not allow modules defined in source_files to shadow modules of the same id that
+    /// exist in dependencies. Checking will fail in this case.
     #[structopt(
-        name = "SOURCES_SHADOW_DEPS",
-        short = cli::SHADOW_SHORT,
-        long = cli::SHADOW,
+        name = "SOURCES_DO_NOT_SHADOW_DEPS",
+        short = cli::NO_SHADOW_SHORT,
+        long = cli::NO_SHADOW,
     )]
-    pub shadow: bool,
+    pub no_shadow: bool,
 }
 
 pub fn main() -> anyhow::Result<()> {
@@ -61,10 +61,15 @@ pub fn main() -> anyhow::Result<()> {
         dependencies,
         sender,
         out_dir,
-        shadow,
+        no_shadow,
     } = Options::from_args();
 
-    let _files =
-        move_lang::move_check_and_report(&source_files, &dependencies, sender, out_dir, shadow)?;
+    let _files = move_lang::move_check_and_report(
+        &source_files,
+        &dependencies,
+        sender,
+        out_dir,
+        !no_shadow,
+    )?;
     Ok(())
 }
--- a/language/move-lang/src/bin/move-check.rs
+++ b/language/move-lang/src/bin/move-check.rs
@@ -44,6 +44,15 @@ pub struct Options {
         long = cli::OUT_DIR,
     )]
     pub out_dir: Option<String>,
+
+    /// If set, allow modules defined in source_files to shadow modules of the same id that exist
+    /// in dependencies
+    #[structopt(
+        name = "SOURCES_SHADOW_DEPS",
+        short = cli::SHADOW_SHORT,
+        long = cli::SHADOW,
+    )]
+    pub shadow: bool,
 }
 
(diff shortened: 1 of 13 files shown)


language/move-lang/src/lib.rs
```diff
@@ -180,6 +180,33 @@ pub fn move_continue_up_to(pass: PassResult, until: Pass) -> Result<PassResult,
     run(pass, until)
 }
 
+pub fn move_compile_up_to(
```
sblackshear, line 183: A comment here would be useful

language/move-lang/src/lib.rs
```diff
@@ -355,6 +382,52 @@ pub fn generate_interface_files(
     Ok(Some(all_addr_dir.into_os_string().into_string().unwrap()))
 }
 
+fn shadow_lib_module_definitions(
```
sblackshear, line 385: Same here

sblackshear (approved):
Nice!

mengxu-fb:
Comments added, would like @tnowacki to have a final look before I land it :)

language/move-lang/src/lib.rs
```diff
@@ -180,6 +180,41 @@ pub fn move_continue_up_to(pass: PassResult, until: Pass) -> Result<PassResult,
...
+    } else {
+        pprog
+    };
+
+    Ok((
+        files,
+        move_continue_up_to(PassResult::Parser(sender_opt, pprog), until),
+    ))
+}
+
```
tnowacki, line 217: I just went through a big refactor trying to slim this down, so I'm not really keen on adding another entry point. 

I think `sources_shadow_deps` should just be a flag on all compiler entry points

tnowacki (changes requested):
Looks simple enough! I'm just trying to keep things slim, so let's see how passing the flag in works

language/move-lang/src/lib.rs
```diff
@@ -180,6 +180,41 @@ pub fn move_continue_up_to(pass: PassResult, until: Pass) -> Result<PassResult,
...
+/// `move_continue_up_to`, the stopping point is inclusive, meaning that the pass specified by
+/// `until: Pass` will be run
+pub fn move_compile_up_to(
+    targets: &[String],
+    deps: &[String],
+    sender_opt: Option<Address>,
+    interface_files_dir_opt: Option<String>,
+    allow_shadow: bool,
+    until: Pass,
+) -> anyhow::Result<(FilesSourceText, Result<PassResult, Errors>)> {
```
tzakian, line 198: Maybe a larger question, but I'm wondering why we are using two different versions of `Result` here...It seems like other places in this file are doing a similar thing, so this doesn't block this PR, but it might be useful to mark this as a future cleanup unless there's a good reason for this. @tnowacki curious if you think this cleanup would make sense.

mengxu-fb: In fact, I had the same wondering here :)

language/move-lang/src/lib.rs
mengxu-fb: Just a heads up, this will lead to many changes across the `language/` dir if every entry points of the compiler need to be updated...

language/move-lang/src/bin/move-check.rs
```diff
@@ -44,6 +44,15 @@ pub struct Options {
...
     pub out_dir: Option<String>,
+
+    /// If set, allow modules defined in source_files to shadow modules of the same id that exist
+    /// in dependencies
+    #[structopt(
+        name = "SOURCES_SHADOW_DEPS",
+        short = cli::SHADOW_SHORT,
+        long = cli::SHADOW,
+    )]
+    pub shadow: bool,
```
tnowacki, line 55: can you set the default to be true? 

I feel like it is more intuitive than the alternative. But thoughts?

mengxu-fb: I don't have a strong opinion either way.

mengxu-fb: Made an update to default it to be true.

language/move-lang/src/lib.rs
tnowacki: Its tricky....

The outer result is for random IO nonsense, the inner one is for errors from the compiler itself. 
IF you wanted to clean this up, you could handroll an enum for the 3 states. It would be clearer but I'm not sure if it would be simpler

tnowacki (approved):
LGTM. Thanks for doing this!

PR #7063, opened 2020-12-29
[language] follow up to latest prologue/epilogue changes

bob-wilson:
PR #6727 added 2 checks to the prologue and 1 check to the epilogue. This PR updates the Move Adapter specification document to match that change and also fixes a number of minor problems and inconsistencies:

- Add a new BAD_TRANSACTION_FEE_CURRENCY error and status code for the case where the transaction fee currency has not been registered in the TransactionFee module. Otherwise, the status is reported as INSUFFICIENT_BALANCE_FOR_TRANSACTION_FEE, which does not make sense in this case that has nothing to do with any balance.

- Reorder the check in the prologue for BAD_TRANSACTION_FEE_CURRENCY so that it comes before checking the sender's balance. If both checks would fail, we should report this one instead of giving the misleading impression that the sender could fix the problem by increasing their balance.

- Reorder the check for PROLOGUE_ESEQUENCE_NUMBER_TOO_BIG so that it does not come after the "too new" sequence number check. The latter must be the last check in the prologue, since it is non-fatal during the initial validation (but not during execution).

- Reorder the prologue specification to match the order in the code.

- Remove the ESEQUENCE_NUMBER error code. It describes the same situation as the new PROLOGUE_ESEQUENCE_NUMBER_TOO_BIG that was added in #6727 and we do not need both. (It was also being used
in a redundant check in the epilogue that the txn sequence number matches the sending account, but the prologue's version of that check returns PROLOGUE_ESEQUENCE_NUMBER_TOO_NEW and the epilogue should be consistent with that. Due to the prologue check, I do not believe this is a breaking change, because it is impossible to hit this error in the epilogue.)

- Rename the PROLOGUE_INVALID_WRITESET_SENDER constant to "PROLOGUE_EINVALID_WRITESET_SENDER", matching all the other prologue error codes that have a "PROLOGUE_E" prefix.

- Remove the EPILOGUE_INVALID_WRITESET_SENDER constant that was added in #6727. We do not need separate error codes for the epilogue and should use the existing prologue error code for this.

- Fix some inaccurate comments.

## Motivation

Document and refine the previous change

Yes

## Test Plan

Updated existing tests and ran them.

## Related PRs

#6727


bob-wilson:
One question here is whether I am correct in assuming that the epilogue can also use the various "PROLOGUE_E*" error codes. I didn't just make that up. There is a precedent in using PROLOGUE_ECANT_PAY_GAS_DEPOSIT in the epilogue for the case where the account balance has dropped. If that precedent is incorrect and there's an important reason to only use those error codes from within the prologue functions, please speak up!

tzakian (approved):
Thanks for fixing these issues! And my apologies for making them in the first place.

This PR looks good to me. 

To your question about reusing the prologue error codes in the epilogue, this shouldn't be an issue -- outside of `ECANT_PAY_GAS_DEPOSIT` -- these will all result in an `UNEXPECTED_ERROR_FROM_KNOWN_MOVE_FUNCTION` if they are ever raised.

language/stdlib/modules/DiemAccount.move
```diff
@@ -1641,22 +1637,22 @@ module DiemAccount {
 
         // Don't grab the balance if the transaction fee is zero
         if (max_transaction_fee > 0) {
-            // [PCA6]: Check that the account has a balance in this currency
+            // [PCA6]: Check that the gas fee can be paid in this currency
```
bob-wilson, line 1640: Another question: could this check be performed even when the max fee is zero? It would be simpler to specify that the gas currency must always be registered for transaction fees, instead of having to say that the check is conditional upon a non-zero max fee. I left it here for 2 reasons: (1) it would be a breaking change for a txn with an invalid gas currency and a gas price of zero, and (2) this gives an automatic exception to WriteSet transactions, which could be important if the TransactionFee currencies ever get broken and need to be fixed. It's not terrible to leave this here, but I welcome suggestions for anything better.

tzakian: The check could be performed when the max fee is zero. 

However, I would be in favor of not doing that for the reasons you mention. I would also add two additional reasons for this:
1. Not checking that the currency is a gas-fee currency seems most in-line (at least to me) with the logic that we have around gas in the "fee = 0" case in the prologue. In that case we don't check that the sending account has a balance in the currency. 
2. If we view the prologue as the minimal set of preconditions for successful execution of the epilogue (with the exception of someone lowering their balance to below the transaction fee during the transaction), not checking the gas currency is a gas-fee currency also makes sense with the logic in https://github.com/diem/diem/blob/master/language/stdlib/modules/DiemAccount.move#L1783

PR #7067, opened 2020-12-29
Fix semantic error response checks for JSON RPC

aching:
The JSON-RPC tests currently use string matching that can vary based on the JSON stringification that is subject to out-of-order fields:

---- async_client::tests::test_batch_send_requests_and_response_id_not_matched_error stdout ----
thread 'async_client::tests::test_batch_send_requests_and_response_id_not_matched_error' panicked at 'assertion failed: `(left == right)`
  left: `"Err(UnexpectedError(InvalidResponseId(JsonRpcResponse { diem_chain_id: 4, diem_ledger_version: 1, diem_ledger_timestampusec: 1602888396000000, jsonrpc: \"2.0\", id: Some(Number(2)), result: Some(Object({\"chain_id\": Number(4), \"timestamp\": Number(234234), \"version\": Number(1)})), error: None })))"`,
 right: `"Err(UnexpectedError(InvalidResponseId(JsonRpcResponse { diem_chain_id: 4, diem_ledger_version: 1, diem_ledger_timestampusec: 1602888396000000, jsonrpc: \"2.0\", id: Some(Number(2)), result: Some(Object({\"version\": Number(1), \"timestamp\": Number(234234), \"chain_id\": Number(4)})), error: None })))"`', client/json-rpc/src/async_client/tests.rs:766:9
stack backtrace:

This change does the equivalence prior to stringification to avoid these types of false, flaky test failures while preserving the full checking of the response.

### DIFF
--- a/client/json-rpc/src/async_client/tests.rs
+++ b/client/json-rpc/src/async_client/tests.rs
@@ -2,6 +2,12 @@
 // SPDX-License-Identifier: Apache-2.0
 
 use crate::async_client::{
+    error::{
+        Error::UnexpectedError,
+        UnexpectedError::{
+            DuplicatedResponseId, InvalidResponseId, InvalidResponseIdType, ResponseIdNotFound,
+        },
+    },
     types as jsonrpc, BroadcastHttpClient, Client, Error, JsonRpcResponse, Request, Response,
     Retry, State, WaitForTransactionError,
 };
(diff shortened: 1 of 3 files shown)


libra-action:

Breaking changes policy:

2. Adding fields is ok.

xli:
we should not test error details, only error type is good enough, because error details may change (like nit message update or a new json field is added due to other changes), the test should focus on asserting what it cares, which is error type in the test.

aching:
I think it's important that we verify that the error response is providing correct information right?  Is that checked elsewhere in the test suite?  We don't want to accidentally provide incorrect information to clients.

aching:
Discussed with @xli offline.  JSON-RPC tests are covered in async (here, for mocking everything to the client), unit_tests.rs (mock db, real server) where the json responses are fully check, and also integration_test.rs (real node) where the json responses are fully check.

This response checking is not needed since it does not actually verify real server code, only a mocked server.  Abandoning.

PR #7068, opened 2020-12-29
[language] Move crate language/move-prover/spec-lang to language/move-model

wrwg:
The spec-lang crate has become a general tool for reflecting over move code, used by tools other than the prover. This PR is a 1st step to reflect this correctly and move it out of the prover. In subsequent PRs, some tools currently in move-prover will be moved out there as well and decoupled from the prover command line driver.

The suugested name is `move-model`, reflecting that this crate's modules provide a model of the Move language which is on a higher abstraction level than file format. It nevertheless contains also the logic for type checking the spec language, which may itself at some point be moved to a different place.

This is a zero-code-change PR, just renaming.

## Motivation

Clean up code base.

Yes

## Test Plan

Existing tests

## Related PRs

NA


libra-action:
```
target packages:
  A move-model 0.1.0 (workspace, path 'language/move-model')
    * features: [none]
  R spec-lang 0.1.0 (workspace, path 'language/move-prover/spec-lang')
    * (old features: [none])
```

bob-wilson:
On the name "move-model": I can't think of anything better than "model" but should we consider dropping the "move-" prefix? I'm envisioning a potential future where the current "language" directory is the top-level of standalone Move project, presumably in a directory named "move", and it seems unnecessary to have a bunch of things named "move/move-*". On the other hand, perhaps "model" alone is too vague to suggest its purpose.

On the content: there's a lot of stuff here that doesn't seem to fit in a "model" (e.g., the specification ASTs, translation from the compiler, and type checking). You mention that the type checking may move elsewhere. That definitely seems like it should happen, and I would say the same for the ASTs and translation. This is just the first step, right?

wrwg:
> On the name "move-model": I can't think of anything better than "model" but should we consider dropping the "move-" prefix? I'm envisioning a potential future where the current "language" directory is the top-level of standalone Move project, presumably in a directory named "move", and it seems unnecessary to have a bunch of things named "move/move-*". On the other hand, perhaps "model" alone is too vague to suggest its purpose.
> 
> On the content: there's a lot of stuff here that doesn't seem to fit in a "model" (e.g., the specification ASTs, translation from the compiler, and type checking). You mention that the type checking may move elsewhere. That definitely seems like it should happen, and I would say the same for the ASTs and translation. This is just the first step, right?

Yes, agreed with removing the move prefix, but I guess we can do this in one later step for all other crates as well. And yes this is only the first step, though some things will be difficult to move out of there. For instance the spec lang AST really is part of the model (a Move program 'model' is bytecode + spec AST + metadata).

emmazzz (approved):
Approved conditional on agreement about name and contents.

bob-wilson:
> Yes, agreed with removing the move prefix, but I guess we can do this in one later step for all other crates as well. And yes this is only the first step, though some things will be difficult to move out of there. For instance the spec lang AST really is part of the model (a Move program 'model' is bytecode + spec AST + metadata).

I see. At some point, we might want to use this model for tools that don't need the specs, but we can consider that later.

bob-wilson:
I'm OK to go ahead with this for now, but I'm adding @tnowacki as a reviewer to make sure he sees it and has a chance to weigh in later.

wrwg:
> > Yes, agreed with removing the move prefix, but I guess we can do this in one later step for all other crates as well. And yes this is only the first step, though some things will be difficult to move out of there. For instance the spec lang AST really is part of the model (a Move program 'model' is bytecode + spec AST + metadata).
> 
> I see. At some point, we might want to use this model for tools that don't need the specs, but we can consider that later.

We use it already for such tools today, namely abigen, diagen, and errmapgen. The API provided in env.rs is a swiss-knife for multiple use cases. If you don't need specs, don't use the functions related to it.

PR #7077, opened 2020-12-30
[move-model] Further cleanup of move-model.

wrwg:
This PR cleans up the `move-model` crate after it was moved out of `move-prove`r:

- Splitting the gigantic previous translate.rs module in multiple modules and moving those into a new subdirectory (model) `building`. This is the code which contains the spec type checker and which we may eventually integrate closer with move-lang.
- Renaming `env` module into `model` and moving some stuff out of there into independent modules.
- Rewriting some documentation to remove move prover bias from this crate.

There is no logical change in this PR even though the diff may create the impression a lot of code has changed.

## Motivation

Cleanup

Yes

## Test Plan

Existing tests

## Related PRs

#7068


language/move-model/src/building/exp_rewriter.rs
```diff
@@ -0,0 +1,159 @@
+// Copyright (c) The Diem Core Contributors
+// SPDX-License-Identifier: Apache-2.0
+
+use std::collections::{BTreeMap, BTreeSet, VecDeque};
+
+use crate::{
+    ast::Exp,
+    building::module_builder::ModuleBuilder,
```
shazqadeer, line 8: I wonder if "builder" is a better name than "building".  I don't have any deep reason for it. "building" just seemed odd to me.

shazqadeer (approved):

language/move-model/src/building/exp_rewriter.rs
wrwg: It does not work because there is already a builder.rs. I next tried `build` but this also does not work because it is in .gitignore ;-)

wrwg: > It does not work because there is already a builder.rs. I next tried `build` but this also does not work because it is in .gitignore ;-)

Found some renaming schema so its now `builder` as it supposed to be.

PR #7089, opened 2021-01-04
Gha

rexhoffman:
## Motivation

Okay due to machine size decrease on CircleCI taking build times from 30 minutes to 60+ action had to be taken.  So here's the start of gha workflow for Diem.

Still in draft form.

Done:
 * Build logic to inspect changes in PRs, Bors, and branch pushes to allow for avoidant/faster builds and land times in bors.
 * Build a new base image for github actions (used debian).
 * Get the unit test flow working.
 *  Find a way to store secrets that only some users can access (environments) and use that in docker build.
 * Simplify the build by pulling out actions for: Detecting Changes, Determining if globs have changed to trigger builds, and docker hub login.

Todo:
 * Nasty learning github actions create a new user in the docker image.   So all of dev-setup.sh's ~./bin/ tools can't be found as they were in /root/bin/ -- gotta fix this.
 * Clean up the environment in the build for tests (pull out all the env setting in to actions, clean up unneeded env settings).
 * timeout executables is in the dockerhub image, but not callable in gha.  Is it caching old images?  Screwing with the shell? Dunno yet
 * Gha wont let you move the home/working dir, which sucks since this is needed to share sccache builds with developer desktops (once we turn on mac builds).
 * we need a way to automate the manual approval step for environments and allow approvers to automatically be approved on their pushes (so the bors bot can push to master for us an get images built immediately).    Mentioned this in gha slack.
 * need to write our own script to clean up seriously old github action runs in the ui, some over 9 months old.    Seems this isn't expect behavior for gha -- also mentioned in gha slack.

At some far distant point in the past.

## Test Plan

Plan to get the tests working in gha.

## Related PRs

None


.github/actions/changes/action.yml
```diff
@@ -0,0 +1,52 @@
...
+    value: ${{ steps.calc.outputs.changes-target-branch }}
+  changes-pull-request-number:
+    description: If a pullrequest number can be found, either in bors, or in a pr, the number.
+    value: ${{ steps.calc.outputs.changes-pull-request-number }}
+runs:
+  using: "composite"
+  steps:
+    - id: calc
+      run: |
+        if [[ $(git log --pretty=format:"%H" -n 26 | wc -l) < 2 ]]; then
```
rexhoffman, line 22: This could use a little bit of love, like all the actions I put together over break.
1. how do we want to log?
2. flag for logging?
3. The shell inside these actions scripts needs refactoring.   I just got it working but not happy with the state.  Targeting getting it working before getting it pretty.

That said, put real effort in to the API exposed by the actions to the workflows.   Look at that as primary take away from this for now.

.github/actions/changes/action.yml
```diff
@@ -0,0 +1,52 @@
...
+          echo It is likely that checkout did not return more than one commit, recommend calling checkout with fetch-depth: 25 1>&2
+          echo BASE_GITHASH is likely to not be found even if it would otherwise be the immediate prior revision. 1>&2
+        fi
+
+        eval $(./.github/actions/changes/get_pr_info.sh -w "${WORKFLOW}" -b -d)
+
+        echo "CHANGES_TARGET_BRANCH=${CHANGES_TARGET_BRANCH}" >> $GITHUB_ENV
```
rexhoffman, line 29: You can expose both environment variables to future builds in a job, and expose a variable as action output.   I think it's smart to do this in general to allow for simpler files as long as we document the behavior.

.github/actions/changes/get_pr_info.sh
```diff
@@ -0,0 +1,150 @@
...
+  esac
+done
+
+if [[ -z "${WORKFLOW_FILE}" ]]; then
+  echoerr Workflow file must be sepecified.
+  usage
+  exit 1
+fi
+
+#calculate the login for curl
```
rexhoffman, line 73: This logic all needs refactoring.   Not happy.  But it works in prs and pushs, need to test in bors, but that section is from the equivalent file in circleci.    Called only by the changes action.

.github/actions/changes/readme.md
```diff
@@ -0,0 +1,55 @@
+# Docker Hub Login and Image Signing #
```
rexhoffman, line 1: I want to see docs like this for every action we write.   Make future user/devs (even if ourselves) hate us less.

.github/actions/dockerhub_login/readme.md
```diff
@@ -0,0 +1,111 @@
...
+
+* First there is an **org key** held by a member of the Diem Association in our case.
+* Second there is a **repositiory key**, aka a key that can must be signed by the **org key** and can be used to allow users to sign docker hub images bound for one **repository** ie:  validator, tools, mint, etc.
+* Third the key used by this action will be neither of the above, it will be a third **user key** that must be signed by the holder of the
+     **repository keys** to allow the signing of the docker image to be considered valid by this **user key**.
+
+Finally, the user key's privilage to sign any repo can be revoked/granted at any time, and is information stored by the dockerhub notary service, not in the key file.
+
+## Example Docker Key Commands ##
+
```
rexhoffman, line 88: @coryhill I no longer have access to the slack channel where I gave you instructions on how to do the key signing for dockerhub, if you could dig it up and comment/update this file once landed I'm sure future versions of us would appreciate it.

.github/actions/matches/action.yml
```diff
@@ -0,0 +1,50 @@
+name: "File Matching"
+description: |
```
rexhoffman, line 2: needs a readme, but pretty simple and works pretty well, after all it is just a grep.

.github/workflows/ci-publish-base-image.yml
```diff
@@ -0,0 +1,37 @@
+name: Publish Base CI Image To Dockerhub
```
rexhoffman, line 1: I built this first to build the image.   Once we have automated environment support I could add this to the main ci-test.yml and conditionally trigger it on lands to main/master.   Kinda want to get down to one workflow file with condition triggers for jobs.   That way we can share state with caches easily.

.github/workflows/ci-test.yml
```diff
@@ -0,0 +1,143 @@
+name: ci-test
+
+on:
+  push:
+    branches: [auto, gha-test-*]
+  pull_request:
+    branches: [main, master, release-*, gha-test-*]
+
+jobs:
+  prepare:
```
rexhoffman, line 10: this determines which jobs should run by finding the changes and setting a test-* output to be picked up by dependent jobs.   Right now only file based.   Could also be branch based.    Work around from gha putting the file change attribute only in the "on" section.   So annoying total the wrong place to put that config, should be on the job.   Would make this prepare section unneeded.

.github/workflows/ci-test.yml
```diff
@@ -0,0 +1,143 @@
...
+        name: find website changes.
+        uses: ./.github/actions/matches
+        with:
+          pattern: "documentation/**"
+
+  dev-setup-sh-test:
+    runs-on: ubuntu-latest-xl
+    continue-on-error: false
+    needs: prepare
+    if: ${{ needs.prepare.outputs.test-dev-setup == 'true' }}
```
rexhoffman, line 55: and here's why we have the prepare section.   Only test the dev-setup.sh script in various docker environments if relevant files have changed, otherwise dont even start up the machines.

.github/workflows/ci-test.yml
```diff
@@ -0,0 +1,143 @@
...
+    steps:
+      - uses: actions/checkout@v2
+      - name: build image with dev-setup.sh
+        run: docker build -f docker/ci/${{ matrix.target_os }}/Dockerfile -t libra/build_environment:test .
+
+  non-rust-lint:
+    runs-on: ubuntu-latest
+    continue-on-error: false
+    needs: prepare
+    if: ${{ needs.prepare.outputs.test-non-rust-lint == 'true'  }}
```
rexhoffman, line 68: same thing for shellcheck/dockerhub

.github/workflows/ci-test.yml
```diff
@@ -0,0 +1,143 @@
...
+    runs-on: ubuntu-latest-xl
+    continue-on-error: false
+    needs: prepare
+    #if: ${{ needs.prepare.outputs.test-rust == 'true' }}
+    container:
+      image: libra/build_environment:github-1
+    env:
+      RUST_BACKTRACE: 1
+    steps:
+      - name: stats
```
rexhoffman, line 97: okay past this is a mess and needs refactoring, but it's passing right now with move tools config out of the way.   see notes about build locations in the github action/home folders and dev's ~/bin/ dir.    Need to fix this to get full working tests.   Not sure about sccache yet.

.github/workflows/ci-test.yml
```diff
@@ -0,0 +1,143 @@
...
+        # echo 'BOOGIE_EXE="$HOME/.dotnet/tools/boogie"' >> $GITHUB_ENV
+      #- name: Setup Env
+      - name: run unit tests
+        run: |
+          set | grep GITHUB
+          set | grep RUST
+          set | grep CARGO
+          echo
+          echo CI_TIMEOUT=$CI_TIMEOUT
+          RUST_BACKTRACE=1 cargo x test --jobs 16 --unit --changed-since "origin/$TARGET_BRANCH"
```
rexhoffman, line 141: Target branch is only set in the prepare job on prs/bors builds.   This will fail on pushes to master/main/release/etc.  Maybe we should use the changes-base-githash instead.

docker/ci/github/Dockerfile
```diff
@@ -0,0 +1,43 @@
...
+    && [ -x "$(command -v cargo)" ] \
+    && [ -x "$(command -v sccache)" ] \
+    && [ -x "$(command -v grcov)" ] \
+    && [ -x "$(command -v helm)" ] \
+    && [ -x "$(command -v aws)" ] \
+    && [ -x "$(command -v z3)" ] \
+    && [ -x "$(command -v "$HOME/.dotnet/tools/boogie")" ] \
+    && [ -x "$(xargs rustup which cargo --toolchain < /diem/rust-toolchain )" ] \
+    && [ -x "$(xargs rustup which cargo --toolchain < /diem/cargo-toolchain)" ]
+
```
rexhoffman, line 38: should test for more tools, needed to add git and jq so far.   also still having issues with timeout.

scripts/dev_setup.sh
```diff
@@ -582,6 +582,8 @@ fi
 if [[ "$PACKAGE_MANAGER" == "apt-get" ]]; then
 	[[ "$BATCH_MODE" == "false" ]] && echo "Updating apt-get......"
 	"${PRE_COMMAND[@]}" apt-get update
+  [[ "$BATCH_MODE" == "false" ]] && echo "Installing ca-certificates......"
```
rexhoffman, line 585: oh and cacerts too...

scripts/dev_setup.sh
```diff
@@ -616,6 +618,12 @@ if [[ "$OPERATIONS" == "true" ]]; then
   install_pkg yamllint "$PACKAGE_MANAGER"
   install_pkg python3 "$PACKAGE_MANAGER"
   install_pkg unzip "$PACKAGE_MANAGER"
+  install_pkg jq "$PACKAGE_MANAGER"
+  install_pkg git "$PACKAGE_MANAGER"
+  #for timeout
+  if [[ "$PACKAGE_MANAGER" == "apt-get" ]]; then
+    install_pkg coreutils "$PACKAGE_MANAGER"
```
rexhoffman, line 625: I should delete this line about coretools, already installed, just timeout isn't working.

PR #7102, opened 2021-01-04
[network] emit metadata on new and lost peers

davidiw:
When a new connection is formed, there is a lot of information available
and rather than just expose what we need for now, let's expose all that
information and let the recipients take what they need.

1. We have already extended the data here once if not twice
2. We have another requirement to extend it to include trusted peer / role

The interesting thing is that state sync and mempool already translate
these events into yet another event so there will need to be ensuing PRs
to address that.

## Motivation

(Write your answer here.)

## Test Plan

## Related PRs


bmwill (approved):

gregnazario (approved):
Yeah, now that I remember it did seem weird to me that we were cutting out part of the metadata, if we used most of it

davidiw:
feeling good about this one...

/land

PR #7113, opened 2021-01-05
[storagte] state pruner: offload the first seek to the worker thread

msmouse:
## Motivation

PR #6803 avoided repeatedly seeking to the beginning in the stale node index, which was costly, by doing it beforehand and only once at construction time. However that one time yet costly call sit in the critical path of the node start up, which can result in a slow node start if the pruning window was large.

This moves the initial seek to the work thread.

Y

## Test Plan

existing coverage.

## Related PRs

#6803

### DIFF
--- a/storage/diemdb/src/pruner/mod.rs
+++ b/storage/diemdb/src/pruner/mod.rs
@@ -19,8 +19,6 @@ use diem_jellyfish_merkle::StaleNodeIndex;
 use diem_logger::prelude::*;
 use diem_types::transaction::Version;
 use schemadb::{ReadOptions, SchemaBatch, SchemaIterator, DB};
-#[cfg(test)]
-use std::thread::sleep;
 use std::{
     iter::Peekable,
     sync::{
@@ -28,7 +26,7 @@ use std::{
         mpsc::{channel, Receiver, Sender},
         Arc,
     },
-    thread::JoinHandle,
+    thread::{sleep, JoinHandle},
     time::{Duration, Instant},
 };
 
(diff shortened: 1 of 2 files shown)


storage/diemdb/src/pruner/mod.rs
```diff
@@ -223,6 +202,51 @@ impl Worker {
...
+    /// Find out the first undeleted item in the stale node index.
+    ///
+    /// Seeking from the beginning (version 0) is potentially costly, we do it once upon worker
+    /// thread start, record the progress and seek from that position afterwards.
+    fn initialize(&mut self) {
+        loop {
+            match self.get_least_readable_version() {
+                Ok(least_readable_version) => {
+                    info!(
+                        least_readable_version = least_readable_version,
```
lightmark, line 214: ?

lightmark (approved):
only one line is confusing

storage/diemdb/src/pruner/mod.rs
```diff
@@ -223,6 +202,51 @@ impl Worker {
...
+                    self.target_least_readable_version = least_readable_version;
+                    self.record_progress(least_readable_version);
+                    return;
+                }
+                Err(e) => {
+                    error!(
+                        error = ?e,
+                        "[state pruner worker] Error on first seek. Retrying in 1 second.",
+                    );
+                    sleep(Duration::from_secs(1));
```
sausagee, line 226: Wonder if longer window is needed - would there be a flood of logs during the time a node restarts and works its way through finding the right postiion?

msmouse: No, it's seeking the local DB and failure is extremely unlikely. It's not something that prevents the node to run, hence a retry loop, otherwise I'd just panic!.

PR #7009, opened 2020-12-22
[move prover] compile all sources in prep for new invariant approach

emmazzz:
## Motivation

This PR is the first of a series of PRs to address the cross-module invariants problem in the Move Prover. This PR
- Compiles all Move code in the provided folder into the environment
- Adds placeholder analysis phase for later analyzing the functions/modules to be translated/verified
- Writes out TODOs for work items in the subsequent PRs 
- Moves the warning of unused schema to the end of spec checker so only one warning will be generated for each unused schema

### Have you read the [Contributing Guidelines on pull requests](https://github.com/diem/diem/blob/master/CONTRIBUTING.md#pull-requests)?

Yes.

PR #7120, opened 2021-01-05
[move prover] add VC for functions modifying target resources

emmazzz:
## Motivation

This PR is the second step of the implementation of the new invariant verification approach. In this PR, we filled in the verification analysis where, in addition to functions in target modules, a verification condition is added for all functions in non-target modules if they directly modify target resources(resources mentioned in target invariants).

A [test file](https://github.com/diem/diem/compare/master...emmazzz:inv2?expand=1#diff-6f883142cd37eba1fe67939369af66f7ff40bd6d7f357f26538a2e83c4216800) is added to show the difference made by this change. With the previous approach the test module would pass the prover with no errors, but with the current approach prover will generate an error since a function in the dependency module violates the invariant. See more details in the comments in the test.

The verification has to be turned off for functions `AccountLimits::publish_unrestricted_limits`, `DiemConfig::set`, `DiemConfig::set_with_capability_and_reconfigure` and `DiemConfig::publish_new_config_and_get_capability`, which are all generic functions violating invariants about specific types of resources. They will be dealt with and the verification will be turned back on when later features become available. 

Yes

## Test Plan

cargo test

## Related PRs

#7009

### DIFF
--- a/language/move-model/src/model.rs
+++ b/language/move-model/src/model.rs
@@ -352,6 +352,8 @@ pub struct GlobalEnv {
     file_id_is_dep: BTreeSet<FileId>,
     /// A set indicating whether a module should be translated or not.
     module_id_should_translate: RefCell<BTreeSet<ModuleId>>,
+    /// A set indicating whether a function not in target module should be verified or not.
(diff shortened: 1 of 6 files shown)


wrwg (approved):
Nice!

PR #7062, opened 2020-12-28
[json-rpc] fix flaky tests and remove unstable tests

xli:
## Motivation

* should not assert json string same, which is unstable due to json to string may have different order to render field values
* remove get_events_with_proofs tests for now, as it failed on build randomly, and don't know root cause.

Failure logs for the flaky tests removed:

```
run get_events_with_proofs with known version: thread 'test_interface' panicked at 'assertion failed: !some_li.is_empty()', json-rpc/tests/integration_test.rs:1097:17
stack backtrace:
   0: std::panicking::begin_panic
             at /rustc/7eac88abb2e57e752f3302f02be5f3ce3d7adfb4/library/std/src/panicking.rs:505
   1: integration_test::create_test_cases::{{closure}}
             at ./tests/integration_test.rs:1097
   2: core::ops::function::FnOnce::call_once
             at /rustc/7eac88abb2e57e752f3302f02be5f3ce3d7adfb4/library/core/src/ops/function.rs:227
   3: integration_test::test_interface
             at ./tests/integration_test.rs:40
   4: integration_test::test_interface::{{closure}}
             at ./tests/integration_test.rs:29
   5: core::ops::function::FnOnce::call_once
             at /rustc/7eac88abb2e57e752f3302f02be5f3ce3d7adfb4/library/core/src/ops/function.rs:227
   6: core::ops::function::FnOnce::call_once
             at /rustc/7eac88abb2e57e752f3302f02be5f3ce3d7adfb4/library/core/src/ops/function.rs:227
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.
failures:
    test_interface
```

```
run get_transactions_with_proofs: thread 'test_interface' panicked at 'assertion failed: hashes.len() == 10 || hashes.len() == 1', json-rpc/tests/integration_test.rs:931:21
stack backtrace:
   0: std::panicking::begin_panic
             at /rustc/7eac88abb2e57e752f3302f02be5f3ce3d7adfb4/library/std/src/panicking.rs:505
   1: integration_test::create_test_cases::{{closure}}
             at ./tests/integration_test.rs:931
   2: core::ops::function::FnOnce::call_once
             at /rustc/7eac88abb2e57e752f3302f02be5f3ce3d7adfb4/library/core/src/ops/function.rs:227
   3: integration_test::test_interface
             at ./tests/integration_test.rs:40
   4: integration_test::test_interface::{{closure}}
             at ./tests/integration_test.rs:29
[...]

PR #6914, opened 2020-12-11
[json-rpc] minimum doc for expiremental verifying APIs

xli:
## Motivation

We didn't document these APIs because they are not ready for public usage, and maybe changed in near future.
This PR just specifically call it out, so that anyone discovered them won't be surprised when we change these APIs.

### Have you read the [Contributing Guidelines on pull requests](https://github.com/diem/diem/blob/master/CONTRIBUTING.md#pull-requests)?

YES

PR #7121, opened 2021-01-05
[json-rpc] Fix Verifying API

AnomalRoil:
This contains fixes so that the verifying API has a uniform API.

Now the `get_transaction_with_proofs` function will behave like the `get_events_with_proofs` and neither have an optional `known_version` parameter that allows to ask for proofs against a specific version since past specific version are likely to have been pruned.

The recommended way of using both APIs is to batch it with a call to `get_state_proof`.

It also adds back an integration test that shouldn't be flaky and should work. 

## Related PRs
#7062
#6539
#6914 
#6399


libra-action:

Breaking changes policy:

2. Adding fields is ok.

xli:
could you add an item to API-CHANGELOG for the change of api params?

xli (approved):

PR #7123, opened 2021-01-05
[secure] Include Move abort code explanation in the VMStatusView

tzakian:
The `VMStatusView` in the `secure-json-rpc` client was dropping the Move abort explanation when deserializing from the json-rpc server response. This adds these fields in so that we now get them. One of the smoke tests for the operational tooling has been updated to make sure that these abort code explanations are being created and returned back out. In the updated test we now get the following back as the MoveAbort status:

```
MoveAbort { 
    location: "00000000000000000000000000000001::DiemSystem",
    abort_code: 519,
    explanation: Some(MoveAbortExplanationView {
        category: "INVALID_ARGUMENT",
        category_description: " An argument provided to an operation is invalid. Example: a signing key has the wrong format.", 
        reason: "EALREADY_A_VALIDATOR", 
        reason_description: " Tried to add a validator to the validator set that was already in it" 
    })
 }
```
We only check the `category` and `reason` as these should be fairly stable, whereas the descriptions are subject to some possible variability.

### DIFF
--- a/testsuite/smoke-test/src/operational_tooling.rs
+++ b/testsuite/smoke-test/src/operational_tooling.rs
@@ -197,166 +197,28 @@ fn test_disable_address_validation() {
 
 #[test]
 fn test_set_operator_and_add_new_validator() {
-    let num_nodes = 3;
-    let (env, op_tool, _, _) = launch_swarm_with_op_tool_and_backend(num_nodes, 0);
+    let status = set_operator_and_add_new_validator_helper();
+    assert_ne!(VMStatusView::Executed, status);
+}
 
-    // Create new validator and validator operator keys and accounts
-    let (validator_key, validator_account) = create_new_test_account();
-    let (operator_key, operator_account) = create_new_test_account();
+#[test]
+fn test_move_explain_operational_tooling() {
+    let status = set_operator_and_add_new_validator_helper();
 
-    // Write the validator key to a file and create the validator account
-    let validator_key_path = write_key_to_file(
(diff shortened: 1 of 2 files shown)


testsuite/smoke-test/src/operational_tooling.rs
```diff
@@ -356,7 +356,23 @@ fn test_set_operator_and_add_new_validator() {
     let txn_ctx = op_tool
         .add_validator(validator_account, &diem_backend, false)
         .unwrap();
-    assert_ne!(VMStatusView::Executed, txn_ctx.execution_result.unwrap());
+    let execution_result = txn_ctx.execution_result.unwrap();
```
JoshLind, line 359: nit: any chance we could move this verification into a new smoke test dedicated to verifying the error category and reason? The issue with the test_set_operator_and_add_new_validator() test is that it's already pretty long and difficult to reason about

JoshLind (approved):
Thanks @tzakian! LGTM 😄

FYI: Although I don't think we should have any big issues with this change, I'd normally want to verify the tool manually by running it. However, given that it looks like we'll need to make some additional modifications to the tool's output shortly (as well as the fact that these changes won't land in the release branch), I can verify the tool manually when handling these changes.

testsuite/smoke-test/src/operational_tooling.rs
tzakian: Sure thing! I'll move it into a separate one.

PR #7124, opened 2021-01-05
[writeset-builder]  Implement a command to output WriteSetPayload

runtian-zhou:
…he stdlib.

## Motivation

Fixed a number of issues around the writeset-builder tool:
1.  Added a flag to output either the WriteSetPayload(required by AOS writeset portal) or the Transaction (required by the db-bootstrapper tool)
2. Fixed the json argument for building custom script.

There will be more follow up items. The most important TODO item is to hook the tool up with real state and generate the release patch for the updateed stdlib.

Yes

## Test Plan

Make sure `buid_custom_script` works with `remove_validators.move` via following command:
`cargo run -- -o out.bin build-custom-script remove_validators.move '{"addresses": ["00000000000000000000000000000000"]}'`
Fixed existing tests.

### DIFF
--- a/language/diem-tools/writeset-transaction-generator/src/main.rs
+++ b/language/diem-tools/writeset-transaction-generator/src/main.rs
@@ -4,17 +4,20 @@
 use anyhow::{format_err, Result};
 use diem_types::{account_address::AccountAddress, transaction::Transaction};
 use diem_writeset_generator::{
-    encode_custom_script, encode_halt_network_transaction, encode_remove_validators_transaction,
+    encode_custom_script, encode_halt_network_payload, encode_remove_validators_payload,
 };
 use std::path::PathBuf;
 use structopt::StructOpt;
 
 #[derive(Debug, StructOpt)]
 struct Opt {
-    /// Path to the local DiemDB file
+    /// Path to the output serialized bytes
     #[structopt(long, short, parse(from_os_str))]
     output: PathBuf,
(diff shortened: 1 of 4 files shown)


language/diem-tools/writeset-transaction-generator/src/main.rs
```diff
@@ -29,22 +39,70 @@ enum Command {
...
+        Command::CherryPickModules { names } => {
+            let modules = build_stdlib()
+                .into_iter()
+                .filter(|(module_name, _)| names.iter().any(|name| module_name.contains(name)))
+                .map(|(name, module)| {
+                    println!("Selecting module {:?}", name);
+                    module
+                })
+                .collect::<Vec<_>>();
+            // TODO: Add Move doctor test here to make sure the state with this new module is
```
mengxu-fb, line 69: Just curious here, will there be a case when we are required to generate an invalid / corrupted module?

mengxu-fb (approved):
Thank you!

language/diem-tools/writeset-transaction-generator/src/main.rs
runtian-zhou: Actually I would like to rework on this a bit more to make sure we can guard against such case!

runtian-zhou:
After a discussion with @sblackshear, we suggest that it is safer to generate the writeset patch by computing the diff between remote stdlib in production and stdlib file on disk. Thus I got rid of the cherry-picking part in this PR.

PR #7127, opened 2021-01-06
[move source language] Add default aliases for standard library

tnowacki:
- Adds aliases for "standard library" modules
- Adds aliases for standard library types, Option and FixedPoint32
- TODO ideally this should be configurable somehow

## Motivation

It's real annoying to have to write use 0x1::Vector and use 0x1::Signer and use 0x1::Errors everywhere

## Test Plan

- new tests

### DIFF
--- a/language/stdlib/modules/DiemAccount.move
+++ b/language/stdlib/modules/DiemAccount.move
@@ -19,16 +19,13 @@ module DiemAccount {
     use 0x1::DiemConfig;
     use 0x1::DiemTimestamp;
     use 0x1::DiemTransactionPublishingOption;
-    use 0x1::Signer;
     use 0x1::SlidingNonce;
     use 0x1::TransactionFee;
     use 0x1::ValidatorConfig;
     use 0x1::ValidatorOperatorConfig;
     use 0x1::VASP;
-    use 0x1::Vector;
     use 0x1::DesignatedDealer;
     use 0x1::Diem::{Self, Diem};
-    use 0x1::Option::{Self, Option};
     use 0x1::Roles;
 
     /// An `address` is a Diem Account iff it has a published DiemAccount resource.
--- a/language/move-lang/src/expansion/aliases.rs
+++ b/language/move-lang/src/expansion/aliases.rs
@@ -2,7 +2,7 @@
 // SPDX-License-Identifier: Apache-2.0
 
 use crate::{
-    parser::ast::ModuleIdent,
+    parser::ast::{ModuleIdent, ModuleIdent_, ModuleName},
     shared::{remembering_unique_map::RememberingUniqueMap, *},
 };
 use move_ir_types::location::*;
@@ -20,6 +20,57 @@ pub struct AliasMap {
     members: RememberingUniqueMap<Name, (ModuleIdent, Name)>,
     current_scope: AliasSet,
 }
+
+// unsafe function for creating ModuleIdent for auto-imported modules
+// This usage is "safe" as the location should never be used
+fn unsafe_no_loc_diem_module_ident(s: &str) -> ModuleIdent {
+    ModuleIdent(Spanned::unsafe_no_loc(ModuleIdent_ {
+        name: ModuleName(Spanned::unsafe_no_loc(s.to_string())),
+        address: Address::DIEM_CORE,
+    }))
+}
+
+macro_rules! prelude_modules {
+    ($(const $const:ident = $v:tt; fn $fn:ident),*) => {
+        $(
+            const $const: &'static str = $v;
+            fn $fn() -> ModuleIdent {
+                unsafe_no_loc_diem_module_ident($const)
+            }
+        )*
+
+        pub(super) fn all_prelude_modules() -> Vec<ModuleIdent> {
+            vec![$($fn(),)*]
+        }
+    };
+}
+
+macro_rules! prelude_members {
+    ($(mod $module_fn:ident; const $const:ident = $v:tt; fn $fn:ident),*) => {
+        $(
+            const $const: &'static str = $v;
+            fn $fn() -> Name {
+                Spanned::unsafe_no_loc($const.to_string())
+            }
+        )*
+
+        pub(super) fn all_prelude_module_members() -> Vec<(ModuleIdent, Name)> {
+            vec![$(($module_fn(), $fn()),)*]
+        }
+    };
+}
+
+// auto included Diem modules
+prelude_modules!(
+    const OPTION_NAME = "Option"; fn option_ident,
+    const SIGNER_NAME = "Signer"; fn signer_ident,
+    const VECTOR_NAME = "Vector"; fn vector_ident
+);
+
+// auto included Diem module members
+prelude_members!(
+    mod option_ident; const OPTION_STRUCT_NAME = OPTION_NAME; fn option_struct_name
+);
 
 impl AliasSet {
     pub fn new() -> Self {
@@ -51,7 +102,8 @@ impl AliasMap {
             current_scope,
         } = self;
         let is_empty = modules.is_empty() && members.is_empty();
-        assert!(current_scope.is_empty() == is_empty);
+        // is_empty ==> current_scope.is_empty();
+        assert!(!is_empty || current_scope.is_empty());
         is_empty
     }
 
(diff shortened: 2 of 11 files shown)


tnowacki:
It's definitely kind of a hack, but a really neat hack...

The idea and implementation is sound, the hack is just in the hardcoding of Diem modules

language/stdlib/modules/DiemAccount.move
```diff
@@ -11,24 +11,19 @@ module DiemAccount {
     use 0x1::AccountLimits::{Self, AccountLimitMutationCapability};
     use 0x1::XUS::XUS;
     use 0x1::DualAttestation;
-    use 0x1::Errors;
```
tnowacki, line 14: I will update the other stdlib modules if this is approved

tzakian:
I like this! A couple questions:
1. Are these module names reserved? e.g., what happens if we have two `0x1::Errors` modules? Or if I have an `0x2::Errors` module and import that? Is it the same as today?
2. What if we add another resource/type to one of these modules? Will this new type be automatically picked up? Does name resolution/errors behave in the same way with them? 
2a. If ^ what if I want to alias a type from the stdlib? i.e., I've written my code and use a currently unused typename `Foo`, but then we add a `Foo` type to the `Errors` stdlib module, do I have to change my code? Or is there some way of explicitly turning off this import?

tnowacki:
1) This wouldn't do anything really. Or nothing new. The check is just "If `0x1::Errors` exists add an implicit alias `use 0x1::Errors as Errors`. Implicit aliases 
2) Autoimporting types from the stdlib might not be a bad idea, but for now its hard coded
2a) There are no issues with your own `Foo` it will shadow the stdlib import and you will never know.

tzakian:
Sounds good! Only concern I might have is with 2a since that could create some ambiguity/non-obvious behavior around which `Foo` is being used possibly (i.e., we now have some kind of binding precedence to these identifiers).

What would you think of: in the 2a case it would be a hard error if there's a name conflict, but we would respect aliases if you imported the stdlib module directly? For example:

```
resource struct Foo { ... }
```

=> ERROR

```
use 0x1::Errors::{Foo as ErrorsFoo};
...
resource struct Foo { ... }
```

=> OK

tnowacki:
I'm not sure about your examples, but both of them would not error 

think about it as if someone had written

```
use 0x1::Errors::Foo;
address 0x42 {
module  M {
  ...
  // Foo resolves to 0x1::Errors::Foo
}
}
```

```
use 0x1::Errors::Foo;
address 0x42 {
module  M {
  // Foo resolves to 0x42::M::Foo;
  resource struct Foo {} // shadows outer use
}
}
```

You can't actually write this today, but it works that way

tzakian:
I guess what I'm saying is that today if I have the following two modules:

```rust
address 0x2 {
module A {
    struct Foo {}
    public fun x() { }
}
}
```

```rust
address 0x2 {
module B {
    use 0x2::A::{Self, Foo};

    struct Foo {}

    public fun y(_x: Foo) { A::x() }
}
}
```

This results in the following error
```
   ┌── src/modules/B.move:5:12 ───
   │
 5 │     struct Foo {}
   │            ^^^ Duplicate module member or alias 'Foo'. Top level names in a namespace must be unique
   ·
 3 │     use 0x2::A::{Self, Foo};
   │                        --- Previously defined here
   │
```

What I'm saying is that if the types in the  `Errors` module are implicitly opened up, I would prefer if name conflicts between a type in `Errors` and the using module resulted in an error similar to this as opposed to an implicit shadowing of the name. 

The question then is: if you find yourself in this situation how should you be able to get out of it? My proposal is that you can by explictly `use`ing the module (and its members) and renaming away the name conflict just as you would in this case.

tnowacki:
You would not get that error, and should not get that error. That error only occurs if you declare the same name in the same scope, as there is no way to resolve which one you meant.

The shadowing is the behavior we already perform for scopes with use. Nothing new there. Scopes can shadow declarations from previous scopes. In this case, it is clear what was intended 

```
use 0x42::A as Foo;
fun ex() {
  Foo::bar(); // A::bar();
  { 
    use 0x42::B as Foo;
    Foo::bar(); // B::bar();
  }
}
```
this is similar to the example I showed
```
use 0x1::Errors::Foo; // the compiler is essentially inserting this alias at the top level outside of all modules
address 0x42 {
module  M {
  // Foo resolves to 0x42::M::Foo;
  resource struct Foo {} // shadows outer use
}
}
```

I understand that it might be confusing that you are silently shadowing an alias that wasn't declared anywhere. 
In the future, I think these aliases should be set up in some config file (maybe with a default setting that we have here). That config will be the sort of explicit declaration.

sblackshear:
> I understand that it might be confusing that you are silently shadowing an alias that wasn't declared anywhere.

+1 to the concern on this. We might also think about banning shadowing of aliases altogether. Adding a single innocent-looking  `use` that sneakily changes the meaning of your entire program seems like a bad thing from a security/auditing perspective.

bob-wilson:
I love the feature but hate the hardcoded hack in the compiler. Can we at least sketch out a plan for a proper way to configure this from something in the Diem Framework before we go ahead with this?

tzakian:
> > I understand that it might be confusing that you are silently shadowing an alias that wasn't declared anywhere.
> 
> +1 to the concern on this. We might also think about banning shadowing of aliases altogether. Adding a single innocent-looking `use` that sneakily changes the meaning of your entire program seems like a bad thing from a security/auditing perspective.

Agreed! I think allowing shadowing is generally something we should examine. Implicit/silent shadowing to me seems like it would go against a lot of what we want in Move source and probably something we don't want (knowing exactly what data structures, functions etc. you're using from the code in the Move module is incredibly important IMO).

tnowacki:
> > > I understand that it might be confusing that you are silently shadowing an alias that wasn't declared anywhere.
> > 
> > 
> > +1 to the concern on this. We might also think about banning shadowing of aliases altogether. Adding a single innocent-looking `use` that sneakily changes the meaning of your entire program seems like a bad thing from a security/auditing perspective.
> 
> I agree, allowing shadowing is generally something we should examine. Implicit/silent shadowing to me seems even harder to understand, and (IMO) seems like it would go against a lot of what we want in Move source and probably something we don't want (knowing exactly what data structures, functions etc. you're using from the code in the Move module is incredibly important).

As discussed yesterday, I would disagree with the characterization that this is implicit/silent shadowing. It is very loud and explicit shadowing. The only possible way to shadow it, is by declaring a new name inside your module. It is then very clear which item it refers to, as you just declared the name. 
Here are the possible cases with `Option` the module and `Option` the struct type as an example
- Declare your own `module Option` inside it will refer to `0xYOUR_ADDR::Option`. The declaration shadowed `0x1::Option as Option` 
- Declare you own alias `use 0xADDR::MODULE as Option`. This use shadows `0x1::Option as Option` 
- Declare your own type `struct Option { ... }`. This declaration shadows `0x1::Option::Option as Option`
- Declare your own alias `use 0xADDR::MODULE::TYPE as Option`. This use shadows `0x1::Option::Option as Option`

There is no implicit shadowing here. The only implicit is the top level `use 0x1::Option` and `use 0x1::Option::Option`

---

Talking about shadowing as its own thing is a separate issue. While I see possible confusion coming from shadowing as a general concept (whether it be with modules, functions, structs, or locals), I think it is a necessary evil that is left up to the programmers discretion. 
- For modules, functions, and/or structs: I feel that in some cases an alias can be confusing and some cases it will be very clear and obvious. It is up to the programmer and the reviewers to ensure the code makes sense. Overall footguns are minimal as any new name is explicitly introduced. In that file, the programmer just needs to search for the name, and they will find the declaration of that alias or type. 
- For local variables: Classically, there are more concerns about footguns because a programmer might (1) shadow a local and then (2) mutate that local and (3) be confused when the mutation isn't seen in the outer scope. This obviously can still happen but we have some tools in the language already that mitigate this risk
  - The compiler warns on unused assignments
  - Move's type system will error if any resource values are unused. This is the really dangerous case for accidentally shadows and the type system will prevent it from happening (in most cases)

tnowacki:
> I love the feature but hate the hardcoded hack in the compiler. Can we at least sketch out a plan for a proper way to configure this from something in the Diem Framework before we go ahead with this?

I think at some point in the future we will need a bunch of different top-level names introduced. Especially if we want named addresses, something like `address Diem = 0x1;`. These sort of names will be top-level and shared across any files the project consumes. 

Perhaps these sort of top level names will be restricted to being used in a config file of sorts. 
In that config file, we could allow top-level `use` declarations. Which will work exactly as the hack works today. 

Concretely, the hack is equivalent to doing 
```
use 0x1::Option::{Self, Option}
...
address MYADDR { module MYCODE { ... } ... }
``` 
where the use declaration is outside of any address/script block. 

#### TLDR
So what is missing is
- Some notion of top level aliases/configs
- Parsing said top level aliases

In other words, this is adding no new *mechanisms*, only a hacked place where existing mechanisms can't really be used.

tnowacki:
@tzakian @sblackshear I reduced the aliases introduced to just Option, Vector, and Signer
@bob-wilson I updated the design a bit to be more conducive to allowing configuration later

davidiw:
Old PR, feel free to re-open.

PR #7128, opened 2021-01-06
[move-prover] Fix hanging if boogie is not found and print which path was tried.

wrwg:
If the BOOGIE_EXE is not found, the current prover_task_runner hangs. That is because a panicking thread does not terminate the program, but just dies, blocking other threads waiting for it. This PR propagates the error out of the task runner to the caller and panics in the main thread. It also prints the path where the boogie program was looked for.

## Motivation

Better UX

Yes

## Test Plan

manual
## Related PRs

NA


rexhoffman (approved):
Stamp to further testing in CI

PR #7129, opened 2021-01-06
[gha] move lang test fix

rexhoffman:
## Motivation

1) Move lang tests are now working in gha.
2) Copies the contents of /root/ installed by the dev-setup.sh in to the new home dir set by gha on execution /github/home/ in the build setup action.
3) Sets needed env variables.
4) Enables triggers that determine when a job should run based on files changed.

Yes

## Test Plan

In gha

## Related PRs

None

### DIFF
--- a/.github/workflows/ci-test.yml
+++ b/.github/workflows/ci-test.yml
@@ -36,7 +36,7 @@ jobs:
         name: find rust/cargo changes.
         uses: ./.github/actions/matches
         with:
-          pattern: '**.rs\|Cargo.toml\|Cargo.lock'
+          pattern: '**.rs\|**.move\|Cargo.toml\|Cargo.lock'
       - id: non-rust-lint-changes
         name: find shell/dockerfile changes
         uses: ./.github/actions/matches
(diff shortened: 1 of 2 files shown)


.github/workflows/ci-test.yml
```diff
@@ -120,15 +121,10 @@ jobs:
...
     needs: prepare
-    #if: ${{ needs.prepare.outputs.test-rust == 'true' }}
+    if: ${{ needs.prepare.outputs.test-rust == 'true' }}
     container:
       image: libra/build_environment:github-1
     steps:
-      - name: stats
-        run: |
-          set x
-          cat /proc/meminfo
```
sausagee, line 130: nit - it would be good to keep this and do another stat dump at the end of the job - we can use the delta to tell mem usage at job level and check if there is regression in it.

sausagee (approved):

.github/workflows/ci-test.yml
rexhoffman: This has been in for a while, we're getting the machine types we're asking for....

PR #7131, opened 2021-01-06
[tools] connect CLI to transaction replay

sblackshear:
Add save_write_sets option to transaction-replay CLI that converts each write set to the on-disk representation that the Move CLI
understands. This makes it easy to get a local mirror of a state of interest, then use the CLI to inspect the state, run `move doctor`,
experiment with running txes on top of the latest state, etc.

## Test Plan

```
cargo run --bin diem-transaction-replay -- --url https://mainnet.diem.com -s replay-transactions 0 1
move view storage/0x00000000000000000000000000000001/modules/XUS.mv 
move doctor
```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2237,6 +2237,7 @@ dependencies = [
  "reqwest",
  "scratchpad",
  "storage-interface",
+ "vm",
 ]
 
 [[package]]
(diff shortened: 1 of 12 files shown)


tzakian (approved):
Looks great to me! I'm looking forward to trying this out :) 

An extension that would be cool to implement at some point would be if we extended the Move CLI to integrate in with this/json-rpc through the OnDiskStateView to fetch account blobs that haven't already been downloaded. My thinking being if you want to work over accounts at some version number but you encounter an account blob that hasn't been downloaded yet it would be nice for that fetching to "just work".

PR #7132, opened 2021-01-06
[language] add vector<address> and vector<u64> for transaction_argument.rs

liangping:
## Motivation

Currently, the TransactionArgument/move script don't have `vector<address>` and `vector<64>` as its input types; However, there are numbers of use cases that need input a sequence of address and numbers. such as Airdrop, One-To-Many transfer(multi-send). So I add address and u64 implementations in TransactionArgument and related places.

Yes

## Test Plan

1, add test script in `language/stdlib/transaction_scripts`
```
script {
    use 0x1::Signer;
    fun test_hello<CoinType>(
        account: &signer,
        _designated_dealer_address: vector<address>,
        _mint_amount: vector<u64>,
        _tier_index: u64
    ) {
        assert(Signer::borrow_address(account)== &0x0, 12)
    }
}
```

2, run `cargo run -p stdlib` 
3. it should work if no error raised.
4. Here we can see the result from stdlib generation:
```
pub fn encode_test_hello_script(
    coin_type: TypeTag,
    _designated_dealer_address: Vec<AccountAddress>,
    _mint_amount: Vec<u64>,
    _tier_index: u64,
) -> Script {
    Script::new(
        TEST_HELLO_CODE.to_vec(),
        vec![coin_type],
        vec![
            TransactionArgument::AddressVector(_designated_dealer_address),
            TransactionArgument::U64Vector(_mint_amount),
            TransactionArgument::U64(_tier_index),
        ],
    )
}
```
5, Here is the result of `cargo x lint && cargo xfmt && cargo xclippy --all-targets`
```
    Finished dev [unoptimized + debuginfo] target(s) in 5m 03s
        INFO [11:18:50.759] - Completed in 306250ms: "cargo" "clippy" "--all-targets" "--workspace" "--" "-D" "warnings" "-A" "clippy::unit_arg" "-A" "clippy::mutable_key_type" "-A" "clippy::eval-order-dependence" "-A" "clippy::new-without-default" "-A" "clippy::rc_buffer" "-W" "clippy::wildcard_dependencies"

```
all tests passed

## Related PRs

### DIFF
--- a/language/diem-tools/writeset-transaction-generator/src/writeset_builder.rs
+++ b/language/diem-tools/writeset-transaction-generator/src/writeset_builder.rs
@@ -123,7 +123,7 @@ fn convert_txn_args(args: &[TransactionArgument]) -> Vec<Value> {
             TransactionArgument::Bool(b) => Value::bool(*b),
             TransactionArgument::AddressVector(v) => Value::vector_address(v.clone()),
             TransactionArgument::U8Vector(v) => Value::vector_u8(v.clone()),
-            TransactionArgument::U64Vector(v)=>Value::vector_u64(v.clone()),
+            TransactionArgument::U64Vector(v) => Value::vector_u64(v.clone()),
         })
         .collect()
 }
--- a/language/diem-tools/writeset-transaction-generator/src/writeset_builder.rs
+++ b/language/diem-tools/writeset-transaction-generator/src/writeset_builder.rs
@@ -121,7 +121,9 @@ fn convert_txn_args(args: &[TransactionArgument]) -> Vec<Value> {
             TransactionArgument::U128(i) => Value::u128(*i),
             TransactionArgument::Address(a) => Value::address(*a),
             TransactionArgument::Bool(b) => Value::bool(*b),
+            TransactionArgument::AddressVector(v) => Value::vector_address(v.clone()),
             TransactionArgument::U8Vector(v) => Value::vector_u8(v.clone()),
+            TransactionArgument::U64Vector(v)=>Value::vector_u64(v.clone()),
         })
         .collect()
 }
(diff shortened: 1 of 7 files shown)


vgao1996:
Hi @liangping, thanks for putting up this PR! Indeed `TransactionArgument` is quite limiting right now and we are planning to further extend it!

However we're considering a more general approach that would allow values of practically any types to passed in from the client side, barring the verifier restriction on the signature of main:
```
pub enum TransactionArgument {
    U8(u8),
    U64(u64),
    U128(u128),
    Address(AccountAddress),
    Bool(bool),
    BCS(Vec<u8>),
}
```
The idea is that clients will serialize the values using [BCS](https://github.com/diem/bcs) before passing them to the VM, and then the VM will try to deserialize the values following the type declarations. Eventually after a grace period, we will make a breaking change removing all other variants so that arguments are represented in a uniform way.
```
pub struct TransactionArgument(pub Vec<u8>);
```

Feel free to implement this if you have time! However in order for the VM to support it, we need to first revive https://github.com/diem/diem/pull/5784 and get it checked in, which our team will do in the next 2 or 3 days.

Let me know if this sounds good to you!

liangping:
Hi, @vgao1996 , sounds interesting for me. 

Currently, It's not possible to deserialize the BCS bytes to a Value if we don't know the Type, I guest this is the one of parts you might need to improve in VM side.

Let me know, when it's ready.

vgao1996:
> Currently, It's not possible to deserialize the BCS bytes to a Value if we don't know the Type, I guest this is the one of parts you might need to improve in VM side.
Indeed. Although from my understanding, this is not a big issue to the VM as the type information is available (in the form of Move code). Maybe there's a use case you are particularly worried about?

It will be problematic however, if you've been using `TransactionArgument` in your own system and relying on the fact that they are self-descriptive. If this is the case, I would recommend you implement your own version of `TransactionArgument` and only convert that into BCS bytes when you are ready to send it to the VM.

vgao1996:
FYI the draft PR is up: #7170. Sorry for the delay. It took longer than expected for me to fix the VM Loader. We should be able to get this in next week.

liangping:
I saw you codes, it's a great improvement.

davidiw:
this PR has been open for nearly 2 months without any activity, closing

PR #7143, opened 2021-01-06
[move-vm] module republish

mengxu-fb:
## Motivation

The Move CLI does not use the VM's API for module publishing as of now. Instead, it directly writes to the on-disk data store. There are two reasons behind: 1) convenience, and 2) module-republishing is not supported in the VM side.

This PR adds module-republishing logic to the VM side, and the current criteria is: module publishing is allowed if the new module is backward compatible with the old module. (Compatibility is defined by the `Compatibility::check()` function). If the new module is not backward compatible, the VM errors out with `StatusCode::BACKWARD_INCOMPATIBLE_MODULE_UPDATE`.

Within this PR, the module re-publishing flow is (if module-republishing is allowed)
1. Build a transaction with publishing module mode together with the module bytecode
2. the VM session execute the transaction and, if all checks are passing, finishes with a transaction effect that contains the bytes of modules to be republished.
3. The storage can then call whatever the procedure is to save modules to storage. In the case of the Move CLI, it is `OnDiskStateView::save_modules()` to save these bytes to storage, regardless whether the same module id already exists or not.

And since the Move CLI also supports publishing modules that break backward compatibility, we cannot re-use the VM's `publish_module` API for that flow, so we force the Move CLI to the on-disk storage override directly.

Yes

## Test Plan

`cargo xtest`

Both the ir-testsuite and e2e-testsuite are updated with test cases to cover module republishing scenarios.


language/move-vm/runtime/src/session.rs
```diff
@@ -110,6 +110,29 @@ impl<'r, 'l, R: RemoteCache> Session<'r, 'l, R> {
...
+        )
+    }
+
+    // TODO: merge with publish_module?
+    pub fn publish_module_ext(
+        &mut self,
+        module: Vec<u8>,
+        sender: AccountAddress,
+        allow_republish: bool,
+        allow_breaking_changes: bool,
```
sblackshear, line 127: I feel uncomfortable with having this flag in the public API even if we plan to remove it later. Thoughts on eliminating it and supporting it via CLI changes earlier instead of later?

Similar with `allow_republish`-- I think it's reasonable to assume that all modules can be republished for now, then add restrictions on republishing inside the VM if/when we change the bytecode format. In general, I don't think we should extend the VM's public API with functionality that we plan to remove later.

mengxu-fb: Totally agreed, in fact, changes on the way!

mengxu-fb: Done, the recent commit should address the discomfort :)

vgao1996: > I think it's reasonable to assume that all modules can be republished for now
I don't think we should do this. Republishing a module is always unsafe as it can change the contract in ways we couldn't detect, even if the old and new signatures match. We must make sure that one can not update a module by simply submitting another Diem module publishing transaction. I feel like at least in the short term, the power to republish will need to kept in check.

language/move-vm/runtime/src/runtime.rs
```diff
@@ -73,14 +74,32 @@ impl VMRuntime {
...
+            let old_m = normalized::Module::new(&old_module);
+            let new_m = normalized::Module::new(&compiled_module);
+            let compat = Compatibility::check(&old_m, &new_m);
+            if !compat.is_fully_compatible() {
+                return Err(
+                    PartialVMError::new(StatusCode::BACKWARD_INCOMPATIBLE_MODULE_UPDATE)
+                        .finish(Location::Undefined),
+                );
+            }
+        }
```
vgao1996, line 102: As I mentioned above, I don't think this is safe. You previous version (boolean flag in the inner layer + separate interfaces for publishing and republishing in the outer layer) seems more favorable to me.

vgao1996 (commented):
In general, I like where this is leading us. In the past if we had to republish a module we would have to directly hack the DB, which was very error prone. Adding something like a `fn republish` or `fn unpublish` to the VM actually makes sense to me, and I think they will make the process safer by imposing additional compatibility checks. We could even add some new (privileged) Diem transactions to make this process official.

language/move-vm/runtime/src/session.rs
sblackshear: I agree with this in the long term. But I think we do not have any mechanism for enforcing this until we add an `updgradability` byte to the module format. Absent that, we must either assume that every module is updatable or no module is updatable. It's going to be quite hard to use/test the update feature if we assume the latter :).

language/move-vm/runtime/src/runtime.rs
sblackshear: Curious: do you think there should be separate transaction types for publish and republish? I was thinking that these should be the same transaction type and same VM API because (1) they require exactly the same data/permissions and (2) I'm not sure what separating them buys us. We have tried fairly hard to keep the transaction types and VM public APIs minimal, so I think we should only add new surface area if there is a strong motivation.

language/move-vm/runtime/src/session.rs
vgao1996: Ah I see. The upgradability byte makes sense to me, especially in an open publishing setup. I would also imagine having more fine-grained control over individual functions or structs. Definitely a feature we should look more into next!

> It's going to be quite hard to use/test the update feature if we assume the latter :).
I still don't quite get this part. The current version of this PR will change the way the existing module publishing transactions work, and allowing contracts to be modified unilaterally. I'm totally fine adding the ability to republish a module, but want to see it being isolated from the normal flow and require special privilege on the Diem layer.

> Absent that, we must either assume that every module is updatable or no module is updatable.
The way I'm understanding this is that in the short term, there will be no open publishing, and since there are a limited number of modules and participants in the network, the association will be able to make folks agree on an upgrade or even change of semantics in certain cases and apply the changes. In other words, all modules are updatable in a while. However since such updates are still extremely dangerous, I just feel like we should use a separate transaction type for it.

language/move-vm/runtime/src/runtime.rs
vgao1996: Ah I see where our disagreement comes from. To me
(1) publishing and republishing require totally different permissions -- republishing is changing an already-agreed contract and thus unsafe and at least in the near future, require association approval. Normal publishing isn't going to break anything so you just need to prove you are the owner of the account (and that you are allowed to publish onto the network).
(2) Another benefit the separation might buy us is reduced chance to make a mistake -- republishing is a lot more risky and it sounds like a bad thing to me you could accidentally trigger this when only meant to publish a new module.

language/move-vm/runtime/src/session.rs
sblackshear: > We must make sure that one can not update a module by simply submitting another Diem module publishing transaction.

> allowing contracts to be modified unilaterally. I'm totally fine adding the ability to republish a module, but want to see it being isolated from the normal flow

What kind of isolation do you have in mind?

>  and require special privilege on the Diem layer.
> However since such updates are still extremely dangerous, I just feel like we should use a separate transaction type for it.

In the Diem layer, module publishing transactions are already restricted to 0xA550C18 (which is the strictest permission we have). With this PR, 0xA550C18 will also have the ability to update modules. I don't think a new transaction type adds any extra safety on top of this already-strictest level.

language/move-vm/runtime/src/runtime.rs
sblackshear: > republishing is changing an already-agreed contract and thus unsafe and at least in the near future, require association approval
> Normal publishing isn't going to break anything so you just need to prove you are the owner of the account (and that you are allowed to publish onto the network).

Today, only the Association can publish and change modules. Letting another party either publish new modules or update new modules are equally risky in my book. For example: the new module may implement a use-case we do not want or are not legally allowed to have, and an updrage to an existing module has the same issue.

So I do not agree that these necessarily require different permissions, but I do agree that once we allow non-Association parties to publish, we must think carefully about what the rules are. I would suggest something like: we can gate module publishing by Association approval, but the value of the upgradability byte is one of the criteria considered in the approval process (e.g., it's easier to get an immutable module approved than one that allows arbitrary upgrades).

However, let's not get too far ahead of ourselves here: this PR is just allowing the Association to republish its own modules, which I think is a very safe and reasonable place to start.
> it sounds like a bad thing to me you could accidentally trigger one when you only meant to publish a new module.

I think that this is quite unlikely because of the backward compatibility checks. Specifically, in order to do this, you new module would need to coincidentally:
1. have the same name as the old one
2. be published at the same address as the old one
3. contain a superset of the structs of the old one
4. contain a superset of the public functions of the old one

If (1) - (4) were all true, I would question why the author was trying to write a new module in the first place instead of extending the existing one :).

language/move-vm/runtime/src/session.rs
mengxu-fb: > The current version of this PR will change the way the existing module publishing transactions work

I am not sure that I get this fully. If the meaning is now publishing and re-publishing follow the same VM API, yes, it is. But I don't think other parts are affected. In particular, both with and without this PR, the transaction for publishing a module will lead to a `TransactionEffect` containing serialized module bytes and the state/storage layer can pick up the bytes and save them accordingly. This part of the module publishing flow is not affected.

vgao1996: > the transaction for publishing a module will lead to a TransactionEffect containing serialized module byte
It's when a module already exists things are different:
- Old behavior: the VM will return an error and won't generate a TransactioEffect a all
- New behavior: the VM will generate a TransactionEffects including an updated version of the module

But as Sam mentioned, the Move VM APIs sit behind the Diem ones, so it's not like this can be invoked arbitrarily by a client -- you need association approval to publish anything right now.

language/move-vm/runtime/src/runtime.rs
vgao1996: Let's continue the discussion in this thread :)

I get your points. I agree that for the Diem side of things, since we currently impose the strictest access control for both publishing and republishing, the overwriting behavior will likely be fine and there's no need to create a new transaction type just for the sake of minimizing risk in the short term. And as you mentioned, if one day we open up publishing to partners or even the public, things will need to be revisited.

I now realize that my "opposition" was more of a matter of Move VM API design: it felt weird to me that the VM is making the choice for its clients, always allowing modules to be updated silently. I feel like this choice should be made by the adapter and the VM's role is to make options available. To me the VM's interfaces should be expressive and somewhat generic, and should not assume the client works in a particular way. I think we're now in a weird time when Diem is still the single monolithic client of Move, but other clients are also emerging, hence the confusion over how far in the future we're looking into. Anyways, let's compare the options we have:
- (current impl) only `fn publish` (overwriting)
- `fn publish(..., allow_republishing: bool)` (error when module exists and flag set to false; overwrite if otherwise)
- `fn publish(...)` (error when module exists) + `fn update(...)` (publish the module if one doesn't exist) -- equivalent to above, just that we now have two entries instead of a boolean flag
- `fn publish(...)` (error when module exists) + `fn update(...)` (error when module doesn't exist)

> I think that this is quite unlikely because of the backward compatibility checks.

This is a very good point and I think following this logic, we can say that it's not well justified that the reason why the VM would have separate entries or the boolean flag is because it wants to protect its clients. Still, we need to consider whether the VM should make these options available to the clients.

mengxu-fb: That is a nice summary of options!

But the only reason I can see for having multiple entry points (i.e., `fn publish`, `fn update`, and even the `fn publish(..., allow_republish: bool)`) in the VM is to support different transaction types. So the underlying question behind these options is: shall we have different transaction types for:

- Publish_NoRepublishing
- Publish_AllowRepublishingWhenCompatible
- Publish_UpdateOnly

And how do we encode that in the transaction types...

sblackshear: >  To me the VM's interfaces should be expressive and somewhat generic, and should not assume the client works in a particular way. 

I agree with this principle, and think that a single `publish(M)` API for both publishing and updates is in keeping with it.

Note that this minimal API can support all three transaction types mentioned by Meng. Keep in mind that an adapter has access to the global storage, which gives it a *lot* of flexibility in how it uses the VM APIs. For example: a client adapter can implement a `PublishAllowRepublishing(M)` transaction type as `if  M does not exists in global storage call VM::publish(M) else err` and a `PublishUpdateOnly(M)` transaction type as `if M exists in global storage call VM::publish(M) else err`, and an `PublishAllowRepublishingWhenCompatible(M)` transaction type as `call VM::publish(M)`.

Now, to answer Meng's question about which transaction types we want in Diem: I think the single module publish transaction type we have is just fine for now unless we find a strong motivation for the other types. We are already considering some changes to module publishing (e.g., multiple modules at a time, multiple signers), and that will be more painful if we duplicate publishing functionality across multiple tx types.

language/tools/move-cli/src/lib.rs
```diff
@@ -315,42 +315,43 @@ impl OnDiskStateView {
...
             fs::create_dir_all(path.parent().unwrap())?
         }
-
-        let mut module_bytes = vec![];
-        module.serialize(&mut module_bytes)?;
-
         Ok(fs::write(path, &module_bytes)?)
     }
 
+    fn sync_interface_files(&self) -> Result<()> {
```
sblackshear, line 326: Nit: a comment would be helpful here.

language/tools/move-cli/src/lib.rs
```diff
@@ -315,42 +315,43 @@ impl OnDiskStateView {
...
-                Some(
-                    self.build_dir
-                        .clone()
-                        .into_os_string()
-                        .into_string()
-                        .unwrap(),
-                ),
-                false,
-            )?;
+            self.sync_interface_files()?;
```
sblackshear, line 354: Not related to this PR, but this would feel less magical if it took a list of `ModuleId`'s...

language/tools/move-cli/src/main.rs
```diff
@@ -251,31 +258,51 @@ fn publish(
...
+
+        if !has_error {
+            let effects = session.finish().map_err(|e| e.into_vm_status())?;
+            if verbose {
+                explain_publish_effects(&effects, &state)?
+            }
+            state.save_modules(&effects.modules)?;
+        }
+    } else {
+        // NOTE: the VM enforces the most strict way of module republishing and does not allow
```
sblackshear, line 293: Nice!

language/tools/move-cli/src/main.rs
```diff
@@ -529,8 +567,54 @@ fn explain_type_error(
...
+                "Module {} exists already. Re-run without --no-republish to publish anyway.",
+                module_id
+            );
+        }
+        VMStatus::Error(BACKWARD_INCOMPATIBLE_MODULE_UPDATE) => {
+            if let Ok(old_m) = state.get_compiled_module(&module_id) {
+                let old_api = Module::new(&old_m);
+                let new_api = Module::new(module);
+                let compat = Compatibility::check(&old_api, &new_api);
+                if !compat.is_fully_compatible() {
```
sblackshear, line 590: Should we assert this instead of using an `if`? I think it should always be true if we get a `BACKWARD_INCOMPATIBLE_MODULE_UPDATE`

language/tools/move-cli/src/main.rs
```diff
@@ -391,14 +406,39 @@ fn run(
...
+        CostStrategy::system(gas_schedule, GasUnits::new(0))
+    };
+    Ok(cost_strategy)
+}
+
+fn explain_publish_effects(effects: &TransactionEffects, _state: &OnDiskStateView) -> Result<()> {
+    // publish effects should contain no events and resources
+    assert!(effects.events.is_empty());
+    assert!(effects.resources.is_empty());
+    // TODO
```
sblackshear, line 436: Finish explaining what's left to be done + how it will use `state`?

sblackshear (commented):
The code LGTM at this point, but don't want to accept it until we've reached consensus on the right VM API(s)!

language/tools/move-cli/src/lib.rs
mengxu-fb: I guess I might need another PR for this, this PR already has several commits.

language/tools/move-cli/src/main.rs
mengxu-fb: Right, we should assert here. I'll make a change.

language/tools/move-cli/src/main.rs
mengxu-fb: Seems that this PR is in the right direction, I'll finish up with the explanation.

mengxu-fb: Done, with the recent commit. For each module in the `TransactionEffects`, it explains whether this is `Publishing a new module <module_id>` or `Updating an existing module <module_id>`.

language/move-vm/runtime/src/runtime.rs
vgao1996: > a PublishUpdateOnly(M) transaction type as if M exists in global storage call VM::publish(M) else err
This makes me feel a lot better about the current solution! And I agree that the module publishing APIs will likely go through more changes in the incoming days, so let's not worry too much about the detailed differences here.

jolestar: I can provide some Move module republish/upgrade practical experience, maybe can help this discussion.

We use MoveVM in starcoin project, and starcoin supports Module upgrade.

1. We remove the repeat module limit in move_vm [runtime.rs](https://github.com/starcoinorg/diem/blob/starcoin_01_05/language/move-vm/runtime/src/runtime.rs)
2. Implement compatibility check in starcoin move VM(https://github.com/starcoinorg/starcoin/blob/master/vm/vm-runtime/src/starcoin_vm.rs#L274).
3. Define [ModuleUpgradeStrategy](https://github.com/starcoinorg/starcoin/blob/master/vm/stdlib/modules/PackageTxnManager.move#L52 ) in stdlib,  module publisher can write an onchain config resource to choose a ModuleUpgradeStragety for all module at the publisher's address.
5. The transaction payload supports a [Package](https://github.com/starcoinorg/starcoin/blob/master/vm/types/src/transaction/mod.rs#L245) type, it contains a Vec<Module> and an init Script(for genesis transaction).
6. When executing the Package transaction, VM checks the Compatibility and UpgradeStragety if the module already exists.

So,  I have the following suggestions:

1. Provider a VM config that allows developers to configure the behavior of the VM, such as check compatibility,  allow publish, enable debug, etc.
2. The VM public API keep simple, do not care about Republish/Compatible, let the upper layer applications to decide how to use the Move VM, such as in Dev env, it allows republish module without check Compatibility.

sblackshear (approved):

language/move-vm/runtime/src/runtime.rs
sblackshear: @jolestar Thanks for sharing your experience and advice!

> We help Move lang support an approach to marker a method is deprecated.

Sounds interesting--can you say more about how this part works?

mengxu-fb: @jolestar Thank you too for the sharing and advice!

I am also curious about this part:
> When executing the Package transaction, VM checks the Compatibility and UpgradeStragety if the module already exists.

Does this implies that the VM knows where the `UpgradeStragety` is stored in the account address? Well, since this `UpgradeStragety` is a resource, the VM should know, but just want to confirm that whether the VM is hardcoded to look for a `UpgradeStragety` resource when doing the `Package` transaction.

jolestar: > @jolestar Thanks for sharing your experience and advice!
> 
> > We help Move lang support an approach to marker a method is deprecated.
> 
> Sounds interesting--can you say more about how this part works?

typo s/help/hope/ ， expect Move lang support @Deprecated annotation like java or rust.

jolestar: > @jolestar Thank you too for the sharing and advice!
> 
> I am also curious about this part:
> 
> > When executing the Package transaction, VM checks the Compatibility and UpgradeStragety if the module already exists.
> 
> Does this implies that the VM knows where the `UpgradeStragety` is stored in the account address? Well, since this `UpgradeStragety` is a resource, the VM should know, but just want to confirm that whether the VM is hardcoded to look for a `UpgradeStragety` resource when doing the `Package` transaction.

VM needs to read on chain state from state store, like VMConfig, if No UpgradeStragety in the Package's module address, use default strategy.

mengxu-fb:
Resolved merge conflict

PR #7149, opened 2021-01-07
[secure] Replace secure/time with common/time-service

phlip9:
+ Add "blocking" support for common/time-service. Users writing non-async code can call `time_service.sleep_blocking(wait)`, which will block the current thread until `wait` duration has elapsed. In the real time service, this just does a `std::thread::sleep`.

+ Add `MockTimeService` auto advance feature. `MockTimeService` now has the option to auto advance time when new `Sleep`s are created. When we're auto advancing, new `Sleep`s are immediately resolved and the simulation time is advanced to the wait deadline. New `Sleep`s whose deadlines are after the auto advance deadline will be pending and must be manually woken by the `MockTimeService::advance_{..}` methods. A version of auto advance is already used by the consensus time service and is necessary to support blocking clients.

+ Replace secure/time with common/time-service. Some visible benefits include: (1) removes the `T: TimeService` trait that otherwise infected most of the types in `secure/storage` and (2) can now use the same time service across modules, including async modules like consensus and network.

+ Add `async` feature gating `tokio` in `MockTimeService` to prevent `tokio` from infecting the TCB.


bmwill:
Hopefully I don't hate life too much if this gets merged in before the tokio upgrade. The time service crate was probably the hardest to upgrade due to the tokio Sleep type no longer being `Unpin`

phlip9:
> Hopefully I don't hate life too much if this gets merged in before the tokio upgrade. The time service crate was probably the hardest to upgrade due to the tokio Sleep type no longer being Unpin

Yeah I noticed that haha. I'm fine landing this after the tokio upgrade goes through : )

secure/key-manager/src/tests.rs
```diff
@@ -540,7 +528,10 @@ fn verify_init_and_basic_rotation<T: DiemInterface>(mut node: Node<T>) {
     // Verify correct initialization (on-chain and in storage)
     node.key_manager.compare_storage_to_config().unwrap();
     node.key_manager.compare_info_to_config().unwrap();
-    assert_eq!(node.time.now(), node.key_manager.last_rotation().unwrap());
+    assert_eq!(
+        node.time.now().as_secs(),
```
gregnazario, line 532: Just wondering, we have `time.now().as_secs()` a lot of places here.  Should we just have a convenience function `now_secs()` to get the seconds directly?  When I worked on my own changes with the time service, I found myself doing the `now().as_secs()` too.

secure/key-manager/src/tests.rs
```diff
@@ -497,15 +485,15 @@ fn verify_manual_rotation_on_chain<T: DiemInterface>(mut node: Node<T>) {
     let new_privkey = Ed25519PrivateKey::generate(&mut rng);
     let new_pubkey = new_privkey.public_key();
     // Increment time to 5min + 1sec, so that the rotation passes through rate limits
-    node.time.increment_by(301);
+    node.time.advance(Duration::from_secs(301));
```
gregnazario, line 488: Similarly, since this is for mock only, maybe `advance_secs()` might be more user friendly as a convenience function :)

secure/key-manager/src/tests.rs
```diff
@@ -15,8 +15,8 @@ use diem_crypto::{ed25519::Ed25519PrivateKey, HashValue, PrivateKey, Uniform};
 use diem_global_constants::{
     CONSENSUS_KEY, OPERATOR_ACCOUNT, OPERATOR_KEY, OWNER_ACCOUNT, OWNER_KEY,
 };
-use diem_secure_storage::{InMemoryStorageInternal, KVStorage};
-use diem_secure_time::{MockTimeService, TimeService};
+use diem_secure_storage::{InMemoryStorage, KVStorage};
+use diem_time_service::{MockTimeService, TimeServiceTrait};
```
gregnazario, line 19: Can we document (maybe I missed it), that the `TimeServiceTrait` needs to be imported to use the `TimeService` functions directly?

common/time-service/src/lib.rs
```diff
@@ -94,8 +101,12 @@ pub trait TimeServiceTrait: Send + Sync + Clone + Debug {
     ///
     /// Canceling a sleep instance is done by dropping the returned future. No
     /// additional cleanup work is required.
+    #[cfg(any(test, feature = "async"))]
     fn sleep(&self, duration: Duration) -> Sleep;
 
+    /// Blocks the current thread until `duration` time has passed.
+    fn sleep_blocking(&self, duration: Duration);
```
gregnazario, line 108: Awesome, I was looking for this in my refactor

common/time-service/src/mock.rs
```diff
@@ -14,12 +14,25 @@ use std::{
     time::Duration,
 };
 
+/// TODO(philiphayes): Use `Duration::MAX` once it stabilizes.
+fn duration_max() -> Duration {
+    Duration::new(std::u64::MAX, 1_000_000_000 - 1)
+}
+
```
gregnazario, line 21: Always wondered why this isn't stable yet.  Could you make it a const or static, so this is only executed once?

common/time-service/src/mock.rs
```diff
@@ -29,6 +42,8 @@ pub struct MockTimeService {
 struct Inner {
     /// The current simulated time.
     now: Duration,
+    /// If `Some`, then auto advance until the simulation time passes the deadline.
+    auto_advance_deadline: Option<Duration>,
```
gregnazario, line 46: Also, something I always wanted in my tests :P

common/time-service/src/real.rs
```diff
@@ -0,0 +1,50 @@
...
+#[cfg(any(test, feature = "async"))]
+use crate::{Sleep, SleepTrait};
+use std::{thread, time::Duration};
+
+/// The real production tokio [`TimeService`].
+///
+/// Note: `RealTimeService` is just a zero-sized type whose methods only delegate
+/// to the respective [`tokio::time`] functions.
+#[derive(Copy, Clone, Debug, Default)]
+pub struct RealTimeService;
```
gregnazario, line 14: I guess `tokio` is `"Real"`

gregnazario (approved):
I'm glad I tried to use this in my own PR, and also glad that I didn't finish it, because there are clearly some rough edges you found how to update in this one.

Mostly just some minor comments here about how some rough edges could be easier for mocking, otherwise lgtm!

testsuite/smoke-test/src/key_manager.rs
```diff
@@ -47,7 +47,7 @@ fn test_key_manager_consensus_rotation() {
     let mut key_manager = KeyManager::new(
         JsonRpcDiemInterface::new(json_rpc_endpoint),
         key_manager_storage,
-        RealTimeService::new(),
+        RealTimeService::new().into(),
```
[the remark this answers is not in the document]
phlip9: You're right, let's standardize on `TimeService::real()` and `TimeService::mock()`

common/time-service/src/mock.rs
phlip9: Duration::new is not a const fn :'(

I'll add an `#[inline]` hint here for now : p

gregnazario: What about static?  Since it's copyable.

phlip9:
rebased

/land

libra-action:
```
target packages:
  A diem-time-service 0.1.0 (workspace, path 'common/time-service')
    * features: default
  R diem-secure-time 0.1.0 (workspace, path 'secure/time')
    * (old features: [none])
```

libra-action:
```
target packages:
  A diem-time-service 0.1.0 (workspace, path 'common/time-service')
    * features: default
  R diem-secure-time 0.1.0 (workspace, path 'secure/time')
    * (old features: [none])
```

libra-action:
```
target packages:
  A diem-time-service 0.1.0 (workspace, path 'common/time-service')
    * features: default
  R diem-secure-time 0.1.0 (workspace, path 'secure/time')
    * (old features: [none])
```

PR #7150, opened 2021-01-07
[secure-storage] support github branches

davidiw:
* add branch support which is different for get versus post
* moved the storage reset on tests to be the first operation rather than
  only after tests to mitigate cases where storage is dirty
* improved error handling on ureq

### DIFF
--- a/config/management/src/secure_backend.rs
+++ b/config/management/src/secure_backend.rs
@@ -85,6 +85,7 @@ impl TryInto<config::SecureBackend> for SecureBackend {
                     .parameters
                     .remove("repository")
                     .ok_or_else(|| Error::BackendParsingError("missing repository".into()))?;
+                let branch = self.parameters.remove("branch");
                 let token = self
                     .parameters
                     .remove("token")
@@ -93,6 +94,7 @@ impl TryInto<config::SecureBackend> for SecureBackend {
                     namespace: self.parameters.remove("namespace"),
                     repository_owner,
                     repository,
+                    branch,
                     token: Token::FromDisk(PathBuf::from(token)),
                 })
             }
@@ -147,6 +149,7 @@ pair: "k0=v0;k1=v1;...".  The current supported formats are:
         an optional namespace: "namespace=NAMESPACE"
         an optional server certificate: "ca_certificate=PATH_TO_CERT"
     GitHub: "backend=github;repository_owner=REPOSITORY_OWNER;repository=REPOSITORY;token=PATH_TO_TOKEN"
+        an optional branch: "branch=BRANCH", defaults to master
         an optional namespace: "namespace=NAMESPACE"
     InMemory: "backend=memory"
     OnDisk: "backend=disk;path=LOCAL_PATH"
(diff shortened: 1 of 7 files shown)


secure/storage/github/src/lib.rs
```diff
@@ -228,12 +243,19 @@ impl Client {
...
 
-    fn url(&self, path: &str) -> String {
+    fn post_url(&self, path: &str) -> String {
         format!(
             "{}/repos/{}/{}/contents/{}",
             URL, self.owner, self.repository, path
         )
     }
+
+    fn get_url(&self, path: &str) -> String {
```
JoshLind, line 253: Aah.. that's annoying that get_url != post_url 😢

JoshLind (approved):
Thanks, @davidiw! 😄 My only suggestion is about adding tests to ensure that different branches indeed don't overlap (see below).

config/management/src/secure_backend.rs
```diff
@@ -147,6 +149,7 @@ pair: "k0=v0;k1=v1;...".  The current supported formats are:
         an optional namespace: "namespace=NAMESPACE"
         an optional server certificate: "ca_certificate=PATH_TO_CERT"
     GitHub: "backend=github;repository_owner=REPOSITORY_OWNER;repository=REPOSITORY;token=PATH_TO_TOKEN"
+        an optional branch: "branch=BRANCH", defaults to master
```
mgorven, line 152: So the caller has to explicitly specify the branch? What about defaulting to `genesis` when doing `genesis` or `create-waypoint` operations?

config/src/config/secure_backend_config.rs
```diff
@@ -24,6 +24,8 @@ pub struct GitHubConfig {
     pub repository_owner: String,
     /// The repository where storage will mount
     pub repository: String,
+    /// The branch containing storage, defaults to master
+    pub branch: Option<String>,
```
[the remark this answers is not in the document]
davidiw: This is required so that the field can be used in a config file... otherwise we have to check for empty and maybe git supports empty branches?

secure/storage/src/tests/github.rs
```diff
@@ -16,6 +18,7 @@ fn github_storage() {
     let mut storage = Storage::from(GitHubStorage::new(
         OWNER.into(),
         REPOSITORY.into(),
+        BRANCH.into(),
```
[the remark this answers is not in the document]
davidiw: It would if these tests could actually be run regularly... but it is all very manual. I did verify this though :)

config/management/src/secure_backend.rs
davidiw: I thought about this and would prefer we introduce this nuance to the tool as testing it from the tooling side is a bit more cumbersome, would require additional testing, and just be a pain in the butt... whereas fixing it in your script is a lot "easier" ... though happy to discuss more if that's not the case.

davidiw:
@land

JoshLind:
@davidiw, did you mean /land? 😄

JoshLind:
😱  @bors-libra .. why you no listen?

PR #7153, opened 2021-01-07
[Dependabot] Disable transitives

rexhoffman:
## Motivation

Given that dependabot transitive prs just clog the CI pipeline without offering any substantial benefit to CI (unless devs mistakenly code against the transitive dep, or devs of the dep code against internals of a transitive dep) lets turn off transitive prs from dependabot for crates.

Sec reviews can still occur in or out of band for the deps prior to release.

Yes

## Test Plan

CI

## Related PRs

None


sausagee (approved):
/land

PR #7165, opened 2021-01-08
[writeset-generator] Implement a command to generate a release patch.

runtian-zhou:
## Motivation

Create a command to create the writeset needed for release in `writeset-generator`. This command will take a json-rpc endpoint and a blockchain height as an input, comparing the diem framework over there with the local version, and generate a WriteSet. This will be a critical step in our release process as we need to check-in this generated binary for AOS to deploy the modified Diem Framework into production.

Yes

## Test Plan

Added a test suite to make sure the diff logic is correct.

### DIFF
--- a/language/diem-tools/writeset-transaction-generator/src/main.rs
+++ b/language/diem-tools/writeset-transaction-generator/src/main.rs
@@ -3,10 +3,13 @@
 
 use anyhow::{format_err, Result};
 use diem_types::{account_address::AccountAddress, transaction::Transaction};
+use diem_validator_interface::{DiemValidatorInterface, JsonRpcDebuggerInterface};
 use diem_writeset_generator::{
-    encode_custom_script, encode_halt_network_payload, encode_remove_validators_payload,
+    create_release_writeset, encode_custom_script, encode_halt_network_payload,
+    encode_remove_validators_payload,
 };
 use std::path::PathBuf;
+use stdlib::build_stdlib;
 use structopt::StructOpt;
 
 #[derive(Debug, StructOpt)]
@@ -36,6 +39,14 @@ enum Command {
         args: String,
         execute_as: Option<AccountAddress>,
     },
+    /// Create a release writeset by comparing local Diem Framework against a remote blockchain state.
+    #[structopt(name = "create-release")]
+    CreateDiemFrameworkRelease {
+        /// Public JSON-rpc endpoint URL.
+        url: String,
+        /// Blockchain height
+        version: u64,
+    },
 }
 
 fn save_bytes(bytes: Vec<u8>, path: PathBuf) -> Result<()> {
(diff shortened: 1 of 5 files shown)


language/diem-tools/writeset-transaction-generator/src/release.rs
```diff
@@ -0,0 +1,72 @@
...
+use anyhow::Result;
+use diem_types::{
+    access_path::AccessPath,
+    transaction::{ChangeSet, WriteSetPayload},
+    write_set::{WriteOp, WriteSetMut},
+};
+use std::collections::{BTreeMap, BTreeSet};
+use vm::CompiledModule;
+
+pub fn create_release_writeset(
```
sblackshear, line 13: Would it make sense to add this as a function of `ChangeSet`? (e.g., `ChangeSet::from_modules(old: Vec<CompiledModule>, new: Vec<CompiledModule>`)? This seems like it might be useful in other contexts too

language/diem-tools/writeset-transaction-generator/src/release.rs
```diff
@@ -0,0 +1,72 @@
...
+        .collect::<BTreeMap<_, _>>();
+    let remote_ids = remote_framework_map.keys().collect::<BTreeSet<_>>();
+    let local_framework_map = local_frameworks
+        .into_iter()
+        .map(|m| (m.self_id(), m))
+        .collect::<BTreeMap<_, _>>();
+    let local_ids = local_framework_map.keys().collect::<BTreeSet<_>>();
+
+    let mut framework_changes = BTreeMap::new();
+    // 1. Insert news modules to be published.
```
sblackshear, line 29:
```suggestion
    // 1. Insert new modules to be published.
```

language/diem-tools/writeset-transaction-generator/src/main.rs
```diff
@@ -36,6 +39,14 @@ enum Command {
         args: String,
         execute_as: Option<AccountAddress>,
     },
+    /// Create a release writeset by comparing local stdlib against a remote blockchain state.
+    #[structopt(name = "create-release")]
+    CreateRelease {
```
sblackshear, line 44: Nit: `CreateDiemFrameworkRelease`?

sblackshear (approved):

language/diem-tools/writeset-transaction-generator/src/release.rs
runtian-zhou: My question is if `diem_types` should import `vm::CompiledModule` as a dependency. Thanks to our previous separation `vm` no longer depends on `diem_types` so it is doable. Although I would vote against it to keep the dependency of `diem_types` to be minimal as it is the core dependency for most of our codebase.

runtian-zhou:
Fixed Sam's comment!

language/diem-tools/writeset-transaction-generator/src/release.rs
sblackshear: Fair point. If we find other uses for this, we can find somewhere more general to put it.

PR #7167, opened 2021-01-08
[move-prover] Refining translation of specs into bytecode.

wrwg:
This is an incremental step in moving the prover translation logic into bytecode transformations. It refines the existing transformation as follows:

- Aborts is now fully handled, including abort codes.
- Propagation of aborts from calls to functions or builtin operators is made explicit via a new instruction `OnAbort(label, local)`. This instruction appears after a call which has the potential to abort. If the abort happened, it branches to the label after storing the abort code in the local. This maintains the flexibility to store abort codes and flag in global state or make them function parameters depending on the backend translation scheme.
- Abort and return handling is moved to a unique block for each function.
- We now generate two variants for each function during specification instrumentation:
  - `FunctionVariant::Baseline`: code intended for inlining, no assertions injected
  - `FunctionVariant::Verification`: code intended for verification. assertions injected
- On call of functions with opaque specifications, the call is rewritten by injecting the specification

This still does not deal with invariants (global and structs) which happens in future PRs.

## Motivation

Prover refactoring

Yes

## Test Plan

Baseline tests

## Related PRs

NA


language/move-prover/bytecode/tests/borrow/basic_test.exp
```diff
@@ -609,13 +609,13 @@ fun TestBorrow::test8(b: bool, n: u64, r_ref: TestBorrow::R): TestBorrow::R {
      var r2: TestBorrow::R
      var t_ref: &mut TestBorrow::R
      var $t6: bool
-     var $t7: TestBorrow::R
-     var $t8: &mut TestBorrow::R
-     var $t9: u64
+     var $t7: u64
```
wrwg, line 612: Those changes come from a bug fix in the reaching definitions processor. The Join operator was broken. Apparently the bug did not effect verification results, but was discovered while testing the new spec_instrumenter.

wrwg:
This PR was significantly updated to now include treatment of calls to opaque functions (which are translated to injecting pre/post conditions instead of an actual call). With those changes, the treatment of function specifications should be fairly complete. What remains are struct and global invariants (and indeed a new -- much simpler than today -- Boogie backend).

language/move-model/src/ast.rs
```diff
@@ -331,41 +331,88 @@ impl Exp {
...
+                shadowed.extend(decls.iter().map(|d| d.name));
+            } else {
+                for sym in decls.iter().map(|d| d.name) {
+                    if let Some(pos) = shadowed.iter().position(|s| *s == sym) {
+                        // Remove one instance of this symbol. The same symbol can appear
+                        // multiple times in `shadowed`.
+                        shadowed.remove(pos);
+                    }
+                }
+                if let Exp::LocalVar(_, sym) = e {
```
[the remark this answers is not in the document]
wrwg: I wish there would be a clippy for this. Will fix with next commit.

language/move-prover/bytecode/tests/spec_instrumentation/opaque_call.exp
```diff
@@ -0,0 +1,240 @@
+============ initial translation from Move ================
+
+[variant baseline]
```
shazqadeer, line 3: What is the purpose of the variant attribute?

language/move-prover/bytecode/src/stackless_bytecode.rs
```diff
@@ -173,6 +224,7 @@ impl BorrowNode {
 pub enum PropKind {
     Assert,
     Assume,
+    Modifies,
```
shazqadeer, line 227: Is this the statement for "havoc"?  If yes, what is the argument to this statement?

shazqadeer (approved):
Lots of good stuff!

language/move-prover/bytecode/src/stackless_bytecode.rs
wrwg: Currently I'm passing in the same expression as we have on spec lang level (global<T>(a))

language/move-prover/bytecode/tests/spec_instrumentation/opaque_call.exp
wrwg: After spec instrumentation, each function comes into variants, one for verification, one for inlining. ("baseline" is used for the inling)

PR #7170, opened 2021-01-08
[language][vm] take serialized values as arguments

vgao1996:
This makes the VM take serialized values as arguments. This change
- Allows us to pass in values of virtually all types, except for references. Generic types are also supported.
- Greatly simplifies the check we need to perform on the VM side.

Note: In order to support signers, `execute_function` now takes a vec of addresses and automatically fills in the signer arguments for the caller, similar to `execute_script`.

This is a reimplementation of #5784.


language/move-vm/runtime/src/runtime.rs
```diff
@@ -111,58 +113,127 @@ impl VMRuntime {
...
-
-        // Build the arguments list for the main and check the arguments are of restricted types.
+        // Build the arguments list and check the arguments are of restricted types.
         // Signers are built up from left-to-right. Either all signer arguments are used, or no
         // signer arguments can be be used by a script.
-        let parameters = &main.parameters().0;
-        let has_signer_parameters = parameters.get(0).map_or(false, is_signer_reference);
-        let mut signers_and_args = if has_signer_parameters {
-            if parameters.len() != args.len() + senders.len() {
+        let has_signer_parameters = tys.get(0).map_or(false, is_signer_reference);
```
sblackshear, line 189: Pre-existing, but: will this do the right thing for (e.g.) `main(i: u64, a: &signer)`? It seems funky to only look at the first type.

language/move-vm/runtime/src/runtime.rs
```diff
@@ -111,58 +113,127 @@ impl VMRuntime {
...
     }
 
-    // See Session::execute_script for what contracts to follow.
-    pub(crate) fn execute_script(
+    fn deserialize_args(&self, tys: &[Type], args: Vec<Vec<u8>>) -> PartialVMResult<Vec<Value>> {
+        // Check if the number of arguments matches the number of types.
+        // REVIEW: is `TYPE_MISMATCH` the most suitable error code here?
+        if tys.len() != args.len() {
+            return Err(
+                PartialVMError::new(StatusCode::TYPE_MISMATCH).with_message(format!(
```
sblackshear, line 121: There is a `NUMBER_OF_TYPE_ARGUMENTS_MISMATCH`. Feels like it would also make sense to have a `NUMBER_OF_ARGUMENTS_MISMATCH`?

language/move-vm/runtime/src/runtime.rs
```diff
@@ -111,58 +113,127 @@ impl VMRuntime {
...
+
+        // Try to deserialize the arguments according to the layouts.
+        let mut vals = vec![];
+        for ((layout, kinfo), arg) in layouts.iter().zip(kinfos.iter()).zip(args.iter()) {
+            match Value::simple_deserialize(arg, kinfo, layout) {
+                Some(val) => vals.push(val),
+                None => {
+                    warn!("[VM] failed to deserialize argument");
+                    return Err(PartialVMError::new(
+                        StatusCode::FAILED_TO_DESERIALIZE_RESOURCE,
```
sblackshear, line 163: `VALUE_DESERIALIZATION_ERROR` or `UNEXPECTED_DESERIALIZATION_ERROR` feel like better fits here given that the argument isn't a resource?

language/move-vm/runtime/src/loader.rs
```diff
@@ -82,14 +82,23 @@ impl ScriptCache {
         }
     }
 
-    fn get(&self, hash: &HashValue) -> Option<Arc<Function>> {
-        self.scripts.get(hash).map(|script| script.entry_point())
+    fn get(&self, hash: &HashValue) -> Option<(Arc<Function>, Vec<Type>)> {
```
vgao1996, line 85: I really really hate the way I got the Loader to return parameter types for main and other functions. Tried a few other methods but they either didn't work due to the quirks of the Loader implementation or were massively more complicated than the current one.

I'll try a little bit harder and see if I can find a cleaner solution, but I think the fundamental issue here is the Loader's lack of modularity and well defined interfaces.  Let's also include "simplifying/remodeling the loader" as part of our plan to better understand the VM data flow.

language/move-vm/runtime/src/runtime.rs
vgao1996: > will this do the right thing for (e.g.) main(i: u64, a: &signer)?

No, since it is currently required that signer arguments come before the regular ones, but I think we can definitely change it if we wish. Thoughts @tnowacki @tzakian?

sblackshear: Ah, by "right thing" I meant reject the code. I am totally ok with requiring the signer arguments to come first (and think it should stay that way).

vgao1996: Oh I see. My understanding is that the current code will reject them, but the error messages may look weird and the errors code may not be correct. I'll do another round of review on errors after I fix our existing tests, and will write additional tests to cover inputs we didn't allow in the past.

language/move-vm/runtime/src/runtime.rs
```diff
@@ -111,58 +130,126 @@ impl VMRuntime {
         data_store.publish_module(&module_id, module)
     }
 
+    fn deserialize_args(&self, tys: &[Type], args: Vec<Vec<u8>>) -> PartialVMResult<Vec<Value>> {
+        // Check if the number of arguments matches the number of types.
+        // REVIEW: is `TYPE_MISMATCH` the most suitable error code here?
```
sblackshear, line 135: Can delete this comment now.

language/move-vm/runtime/src/runtime.rs
```diff
@@ -111,58 +130,126 @@ impl VMRuntime {
...
-        } else {
-            // no signer parameters, don't add to args
-            vec![]
-        };
-        signers_and_args.append(&mut args);
-        check_args(&signers_and_args).map_err(|e| e.finish(Location::Script))?;
-
+        let signers_and_args = self
+            .create_signers_and_arguments(&params, senders, args)
+            .map_err(|err| err.finish(Location::Undefined))?;
```
sblackshear, line 248:
```suggestion
            .map_err(|err| err.finish(Location::Script))?;
```
or perhaps read the start location from `main`?

language/move-vm/runtime/src/runtime.rs
```diff
@@ -177,24 +264,27 @@ impl VMRuntime {
...
-        let (func, type_params) =
+        let (func, ty_args, params) =
             self.loader
                 .load_function(function_name, module, &ty_args, data_store, log_context)?;
 
         // check the arguments provided are of restricted types
-        check_args(&args).map_err(|e| e.finish(Location::Module(module.clone())))?;
+        let args = self
+            .create_signers_and_arguments(&params, senders, args)
+            .map_err(|err| err.finish(Location::Undefined))?;
```
sblackshear, line 282: Is it possible to read the start location from `func`?

language/move-vm/runtime/src/runtime.rs
```diff
@@ -177,24 +264,27 @@ impl VMRuntime {
         module: &ModuleId,
         function_name: &IdentStr,
         ty_args: Vec<TypeTag>,
-        args: Vec<Value>,
+        args: Vec<Vec<u8>>,
```
sblackshear, line 267: Should we rename this to `execute_function_ext` or something else to distinguish it from the "normal", internal call case?

language/move-vm/runtime/src/runtime.rs
```diff
@@ -177,24 +264,27 @@ impl VMRuntime {
...
         // its dependencies if the module was not loaded
-        let (func, type_params) =
+        let (func, ty_args, params) =
             self.loader
                 .load_function(function_name, module, &ty_args, data_store, log_context)?;
 
         // check the arguments provided are of restricted types
-        check_args(&args).map_err(|e| e.finish(Location::Module(module.clone())))?;
+        let args = self
+            .create_signers_and_arguments(&params, senders, args)
```
sblackshear, line 281: What will happen if I try to call a function with a signature like `f(i: u64, s: &signer)`? Although we force the `signer`s to occur first in scripts, functions do not have this restriction.

language/move-vm/runtime/src/runtime.rs
tnowacki: Its an interesting one. 
Typically "Undefined" is used for anything that is VM specific and doesn't really live in any location. For example if you hit an error in the loader, you get Undefined even if that was the result of loading some module. 

So here I could see it going eitherway, since its for the script args, but its not really *in* the script since the script has not started executing yet

sblackshear: Yeah, good point. This is really like a bytecode verifier error; keeping locations for errors that occur during concrete execution seems very sensible.

vgao1996: Shall we consider adding something like a `Location::TransactionArgument`? To me that's also a common source of user errors.

language/diem-vm/src/diem_vm.rs
```diff
@@ -227,16 +225,25 @@ impl DiemVMImpl {
...
-                    Value::u64(txn_sequence_number),
-                    Value::vector_u8(txn_public_key),
-                    Value::u64(txn_gas_price),
-                    Value::u64(txn_max_gas_units),
-                    Value::u64(txn_expiration_timestamp_secs),
-                    Value::u8(chain_id.id()),
-                    Value::vector_u8(txn_data.script_hash.clone()),
+                    MoveValue::U64(txn_sequence_number)
+                        .simple_serialize()
+                        .unwrap(),
```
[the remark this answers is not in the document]
vgao1996: - They are guaranteed to succeed.
- I'll look into this option.
- Will also look into this.

language/move-vm/runtime/src/runtime.rs
vgao1996: Good question. In order to address this I added a few new error codes and slightly changed the way the checks were performed.

Here's how we do it now:
- First we look at the signature of the function and split the params into two parts: `[prefix signer refs] [rest]`
- If `len([prefix signer refs])` is non-zero and not equal to `len([signer args])` then `Err(NUMBER_OF_SIGNER_ARGUMENTS_MISMATCH)`
- if `len(args)` != `len([rest])` then `Err(NUMBER_OF_ARGUMENTS_MISMATCH)`
- We try to obtain the layouts of the types in `[rest]`, and if one fails, we return `Err(INVALID_PARAM_TYPE_FOR_DESERIALIZATION)`. This error code covers not only `&signer` but also other reference or even invalid types.
- Lastly we try to deserialize the values and return `Err(FAILED_TO_DESERIALIZE_ARGUMENT)` upon failure.

BTW I fixed a few existing tests but that was just to see how it runs. Still planning to review the tests again and will likely rewrite a few of them and add more. Will let you know when that is done.

language/move-vm/runtime/src/runtime.rs
vgao1996: Not sure I understand what "normal" internal call case we're trying to disambiguate from? Could you elaborate?

libra-action:
```
target packages:
  M backtrace 0.3.55 (direct third-party, crates.io)
    * version DOWNGRADED from 0.3.56
    * (unchanged features: addr2line, default, gimli-symbolize, miniz_oxide, object, serde, std)
  M bytes 1.0.0 (direct third-party, crates.io)
    * version DOWNGRADED from 1.0.1
    * (unchanged features: default, std)
  M serde 1.0.120 (direct third-party, crates.io)
    * version DOWNGRADED from 1.0.122
    * (unchanged features: default, derive, rc, serde_derive, std)
  M fiat-crypto 0.1.5 (transitive third-party, crates.io)
    * version DOWNGRADED from 0.1.6
  M object 0.22.0 (transitive third-party, crates.io)
    * version DOWNGRADED from 0.23.0
    * (unchanged features: archive, coff, elf, macho, pe, read_core, unaligned)
  M parking_lot 0.11.1 (transitive third-party, crates.io)
    * status changed from direct third-party
    * (unchanged features: default)

host packages:
  M syn 1.0.58 (direct third-party, crates.io)
    * version DOWNGRADED from 1.0.60
    * (unchanged features: clone-impls, default, derive, extra-traits, full, parsing, printing, proc-macro, quote, visit)
  M serde_derive 1.0.120 (transitive third-party, crates.io)
    * version DOWNGRADED from 1.0.122
    * (unchanged features: default)
```

libra-action:
```
target packages:
  M backtrace 0.3.55 (direct third-party, crates.io)
    * version DOWNGRADED from 0.3.56
    * (unchanged features: addr2line, default, gimli-symbolize, miniz_oxide, object, serde, std)
  M bytes 1.0.0 (direct third-party, crates.io) (other versions: 0.5.6)
    * version DOWNGRADED from 1.0.1
    * (unchanged features: default, std)
  M serde 1.0.120 (direct third-party, crates.io)
    * version DOWNGRADED from 1.0.122
    * (unchanged features: default, derive, rc, serde_derive, std)
  M fiat-crypto 0.1.5 (transitive third-party, crates.io)
    * version DOWNGRADED from 0.1.6
  M object 0.22.0 (transitive third-party, crates.io)
    * version DOWNGRADED from 0.23.0
    * (unchanged features: archive, coff, elf, macho, pe, read_core, unaligned)
  M parking_lot 0.11.1 (transitive third-party, crates.io)
    * status changed from direct third-party
    * (unchanged features: default)
  M tokio-stream 0.1.1 (transitive third-party, crates.io)
    * version DOWNGRADED from 0.1.2
    * (unchanged features: default, time)
  M tokio-util 0.6.1 (transitive third-party, crates.io)
    * version DOWNGRADED from 0.6.2
    * (unchanged features: codec, default)
  R async-stream 0.3.0 (transitive third-party, crates.io)
    * (old features: [none])

host packages:
  M syn 1.0.58 (direct third-party, crates.io)
    * version DOWNGRADED from 1.0.60
    * (unchanged features: clone-impls, default, derive, extra-traits, full, parsing, printing, proc-macro, quote, visit, visit-mut)
  M serde_derive 1.0.120 (transitive third-party, crates.io)
    * version DOWNGRADED from 1.0.122
    * (unchanged features: default)
  R async-stream-impl 0.3.0 (transitive third-party, crates.io)
    * (old features: [none])
```

libra-action:
```
target packages:
  A state-synchronizer 0.1.0 (workspace, path 'state-synchronizer')
    * features: default
  M backtrace 0.3.55 (direct third-party, crates.io)
    * version DOWNGRADED from 0.3.56
    * (unchanged features: addr2line, default, gimli-symbolize, miniz_oxide, object, serde, std)
  M bytes 1.0.0 (direct third-party, crates.io) (other versions: 0.5.6)
    * version DOWNGRADED from 1.0.1
    * (unchanged features: default, serde, std)
  M serde 1.0.120 (direct third-party, crates.io)
    * version DOWNGRADED from 1.0.122
    * (unchanged features: alloc, default, derive, rc, serde_derive, std)
  M serde-generate 0.17.0 (direct third-party, crates.io)
    * version DOWNGRADED from 0.18.0
  M tokio-stream 0.1.1 (direct third-party, crates.io)
    * version DOWNGRADED from 0.1.2
    * (unchanged features: default, time)
  M tokio-util 0.6.1 (direct third-party, crates.io)
    * version DOWNGRADED from 0.6.2
    * (unchanged features: codec, compat, default, futures-io, io)
  M warp 0.3.0 (direct third-party, crates.io)
    * removed features: tls, tokio-rustls
    * (unchanged features: default, multipart, tokio-tungstenite, websocket)
  M fiat-crypto 0.1.5 (transitive third-party, crates.io)
    * version DOWNGRADED from 0.1.6
  M getrandom 0.2.1 (transitive third-party, crates.io) (other versions: 0.1.15)
    * version DOWNGRADED from 0.2.2
    * (unchanged features: std)
  M object 0.22.0 (transitive third-party, crates.io)
    * version DOWNGRADED from 0.23.0
    * (unchanged features: archive, coff, elf, macho, pe, read_core, unaligned)
  M parking_lot 0.11.1 (transitive third-party, crates.io)
    * status changed from direct third-party
    * (unchanged features: default)
  R state-sync 0.1.0 (workspace, path 'state-sync')
    * (old features: default)
  R async-stream 0.3.0 (transitive third-party, crates.io)
    * (old features: [none])
  R ring 0.16.19 (transitive third-party, crates.io)
    * (old features: alloc, default, dev_urandom_fallback, once_cell)
  R rustls 0.19.0 (transitive third-party, crates.io)
    * (old features: default, log, logging)
  R sct 0.6.0 (transitive third-party, crates.io)
    * (old features: [none])
  R spin 0.5.2 (transitive third-party, crates.io)
    * (old features: [none])
  R tokio-rustls 0.22.0 (transitive third-party, crates.io)
    * (old features: [none])
  R untrusted 0.7.1 (transitive third-party, crates.io)
    * (old features: [none])
  R webpki 0.21.4 (transitive third-party, crates.io)
    * (old features: default, std, trust_anchor_util)

host packages:
  M syn 1.0.58 (direct third-party, crates.io)
    * version DOWNGRADED from 1.0.60
    * (unchanged features: clone-impls, default, derive, extra-traits, full, parsing, printing, proc-macro, quote, visit, visit-mut)
  M bytes 1.0.0 (transitive third-party, crates.io)
    * version DOWNGRADED from 1.0.1
  M getrandom 0.2.1 (transitive third-party, crates.io)
    * version DOWNGRADED from 0.2.2
    * (unchanged features: std)
  M serde_derive 1.0.120 (transitive third-party, crates.io)
    * version DOWNGRADED from 1.0.122
    * (unchanged features: default)
  R async-stream-impl 0.3.0 (transitive third-party, crates.io)
    * (old features: [none])
```

language/move-vm/runtime/src/runtime.rs
tnowacki: > Shall we consider adding something like a Location::TransactionArgument? 
I feel like that bucket is a bit narrow, but maybe

> Yeah, good point. This is really like a bytecode verifier error; keeping locations for errors that occur during concrete execution seems very sensible.
From a bytecode verifier perspective, this should probably be Script, since the location is set for the verifier. 
But I think it depends on what this error really is semantically. 

So script might make sense here

vgao1996:
UPDATE: added all but 1 test for script. Still working on tests for functions.

language/move-vm/runtime/src/runtime.rs
sblackshear: E.g., calling `M::foo()` from `M::bar()`

language/move-vm/runtime/src/runtime.rs
sblackshear: I see. That seems a bit funky, because a signer can be both in `senders` and `args`...

I wonder if decoupling the script/function cases and checking signature validity from deserialization would make life easier. For example:

- Replace `create_signers_and_arguments(params, senders, args)` with `senders.extend(args); let all_args = senders; create_arguments(senders, args)`. All `create_arguments` does is match types to args + serialize, and it treats `&signer` and `address` the same.
- Add a separate function for checking that the `signer`s are the first part of the signature + call it only from the `script` case.

vgao1996:
Had a discussion with @sblackshear and we realized we could further simplify the VM public interfaces by treating signers more like normal arguments. Here's a new proposal:

## Move VM
`fn execute_script` and `fn execute_function` will no longer take a Vec of signer addresses:
```
    fn execute_script(
        &self,
        script: Vec<u8>,
        ty_args: Vec<TypeTag>,
        args: Vec<Vec<u8>>,
        senders: Vec<AccountAddress>,
        cost_strategy: &mut CostStrategy,
        log_context: &impl LogContext,
    ) -> VMResult<()>

    fn execute_function(
        &self,
        module: &ModuleId,
        function_name: &IdentStr,
        ty_args: Vec<TypeTag>,
        args: Vec<Vec<u8>>,
        senders: Vec<AccountAddress>,
        cost_strategy: &mut CostStrategy,
        log_context: &impl LogContext,
    ) -> VMResult<()>
```
The Move VM will simply match the arguments against parameters and deserialize them. It will no longer require signers to appear before other arguments, perform signer auto-filling, or allow signer args to be ignored. For now, we'll implement a special rule allowing an `&signer` to be constructed from bytes with a layout identical to that of `address` (or `signer`). This will enable a smooth transition to a model of owned signers under the new constraint system.
```
fun foo(x: &signer, y: u64, z: &signer);
// Args: 0x1, 42, 0x2
// OK

fun main(x: u64, y: &signer);
// Args: 42, 0x1
// OK

fun main(x: u64);
// Args: 0x1, 42
// Err: NUMBER_OF_ARGUMENTS_MISMATCH
// The Move VM will not allow excessive signers args.

fun main(x: bool);
// Args: 0
// Err: FAILED_TO_DESERIALIZE_ARGUMENT

fun main(x: &u64);
// Err: INVALID_PARAM_TYPE_FOR_DESERIALIZATION
```

## Bytecode Verifier
Currently, the bytecode verifier only allows `&signer`s to appear at the beginning of the list and restrict the remaining params to certain primitive types. This limitation will remain unchanged.

## Diem Adapter
When executing a script, the Diem Adapter will insert signers into the beginning of the arg list if the signature of main starts with 1 or more `&signer`. It will error if the number of signer params (non-zero) does not match the number of signer args. (Personally I think it's the adapter's job to check this, but technically it could also be done by the Move VM and I'm open to both options.)

Let me know how you guys think about this!

vgao1996:
@sblackshear I implemented some of the changes we discussed offline. Here's how it works now:
- `fn execute_function` no longer takes `senders: Vec<AccountAddress>`. Instead you'll have to pass them as normal arguments. There is a special rule permitting `&signer` to be created from an argument serialized as `signer`.
  - Possible errors: 
    - NUMBER_OF_ARGUMENTS_MISMATCH
    - INVALID_PARAM_TYPE_FOR_DESERIALIZATION
    - FAILED_TO_DESERIALIZE_ARGUMENT
- `fn execute_script` still takes in `senders: Vec<AccountAddress>` and propagate the signer args for you.
  - Possible errors:
    - INVALID_MAIN_FUNCTION_SIGNATURE (by the verifier)
    - NUMBER_OF_SIGNER_ARGUMENTS_MISMATCH
    - NUMBER_OF_ARGUMENTS_MISMATCH
    - FAILED_TO_DESERIALIZE_ARGUMENT

language/move-vm/runtime/src/runtime.rs
vgao1996: Internal function calls are handled by the interpreter loop. `fn execute_function` is only used when you are calling from an outer environment.

vgao1996:
Update: rebased onto latest master

sblackshear (approved):
LGTM!

PR #7175, opened 2021-01-08
[diem-vm] fix performance regression

bob-wilson:
PR #6891 introduced an unintended performance regression by using a separate TransactionDataCache for the validation step during execution. Account data that was deserialized while validating a transaction had to be deserialized again while executing it. This changes to pass around the VM session so that the same cache is used in both steps.

Yes

## Test Plan

No behavioral change; ran existing tests

### DIFF
--- a/language/diem-vm/src/diem_transaction_executor.rs
+++ b/language/diem-vm/src/diem_transaction_executor.rs
@@ -151,9 +151,9 @@ impl DiemVM {
         ))
     }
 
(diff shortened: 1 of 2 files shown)


vgao1996:
The code looks good to me, but I wonder why certain ci tests (most notably lint, unit-test, e2e-test) got skipped. Shall we check with dev infra about it?

vgao1996 (approved):

PR #7176, opened 2021-01-08
[rate-limiter] Turn rate limiting off by default

gregnazario:
### Overview
Currently, there's a requirement that there is configuration for
FullNodes for rate limiting.  Now, we leave it up to the operator to turn on rate limiting, and allow an easier way to turn it off.

### DIFF
--- a/config/src/config/network_config.rs
+++ b/config/src/config/network_config.rs
@@ -187,18 +187,6 @@ impl NetworkConfig {
             ));
         }
 
-        // Fullnodes require a rate limiting config
-        if role == RoleType::FullNode {
-            self.inbound_rate_limit_config = Some(
-                self.inbound_rate_limit_config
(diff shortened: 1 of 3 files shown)


config/src/config/network_config.rs
```diff
@@ -331,6 +331,8 @@ pub struct RateLimitConfig {
     pub ip_byte_bucket_size: usize,
     /// Initial amount of tokens initially in the bucket
     pub initial_bucket_fill_percentage: u8,
+    /// Allow for an open throttle
+    pub open: bool,
```
phlip9, line 335: maybe rename this to `enabled`? seems more intuitive to me when writing the config, like

```yaml
inbound_rate_limit_config:
    enabled: false
```

phlip9 (approved):
LGTM!

PR #7177, opened 2021-01-08
[rate-limiter] Add Optional Rate limit metrics

gregnazario:
Adding rate limit metrics at the rate limiter level.  Logging provides more detail, but to not cause excess cardinality issues in the metrics system, metrics are not at a key level.

### DIFF
--- a/common/rate-limiter/src/rate_limit.rs
+++ b/common/rate-limiter/src/rate_limit.rs
@@ -53,6 +53,7 @@ const ONE_SEC: Duration = Duration::from_secs(1);
 ///
 pub struct TokenBucketRateLimiter<Key: Eq + Hash + Clone + Debug> {
     label: &'static str,
+    log_info: String,
     buckets: RwLock<HashMap<Key, SharedBucket>>,
     new_bucket_start_percentage: u8,
     default_bucket_size: usize,
@@ -64,6 +65,7 @@ pub struct TokenBucketRateLimiter<Key: Eq + Hash + Clone + Debug> {
 impl<Key: Eq + Hash + Clone + Debug> TokenBucketRateLimiter<Key> {
     pub fn new(
         label: &'static str,
+        log_info: String,
         new_bucket_start_percentage: u8,
         default_bucket_size: usize,
         default_fill_rate: usize,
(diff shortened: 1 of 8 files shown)


JoshLind (approved):
Overall, this makes sense to me. I'm still a little iffy on exactly how these metrics and logs are intended to be used (but I suspect this is because I've yet to see it work in practice and explore it myself 😄).

common/rate-limiter/src/rate_limit.rs
```diff
@@ -239,19 +256,27 @@ impl Bucket {
...
+                    self.allowed_in_period
+                        .saturating_add(self.throttled_in_period),
                 );
             }
 
             // Optional metrics
             if let Some(metrics) = self.metrics.as_ref() {
                 metrics
-                    .with_label_values(&[self.label.as_str(), self.key.as_str(), "allowed"])
+                    .with_label_values(&[self.label.as_str(), "allowed"])
```
JoshLind, line 276: Any reason "key" was removed from the metrics?

gregnazario: Too high cardinality :(

That would create a metric for every incoming IP ever connected to a node * the number of network interfaces.

I think in the future if we work some of those kinks out, this won't be an issue.

JoshLind: :O Aah.. makes sense! Thanks @gregnazario 😄

gregnazario:
> Overall, this makes sense to me. I'm still a little iffy on exactly how these metrics and logs are intended to be used (but I suspect this is because I've yet to see it work in practice and explore it myself 😄).

So, normally a metric like this would tell us the `actual throughput` of the system, as well as the `blocked throughput` of the system.  It can be helpful to tell you if the limits are wrong, e.g. all the traffic is legitimate, and all of the services are working fine, but 2/3 of the traffic is being dropped.

The logs provide a way to construct these metrics in a non-aggregated way because it's the raw data per refill period.  Maybe we could add a switch to turn on and off these logs / metrics.

Additionally, if you see a large spike of throttles, you can correlate that in the logs to an IP, and block that IP if they're DoSing you.

PR #7178, opened 2021-01-09
[writeset-generator] Implement a verificiation logic for release flow

runtian-zhou:
## Motivation

Make sure after applying the writeset, the Diem Frame on chain will be exactly the same as the Diem Framework on disk.

This PR requires a slight refactor of the `DiemVM` to be able to invoke the `execute_writeset` payload logic from outside of this module.

Yes

## Test Plan
Tested the tool against mainnet via following command:
`cargo run -- --output out.bin --output-payload create-release https://mainnet.diem.com 380066305`

### DIFF
--- a/language/diem-tools/writeset-transaction-generator/src/release.rs
+++ b/language/diem-tools/writeset-transaction-generator/src/release.rs
@@ -1,13 +1,23 @@
 // Copyright (c) The Diem Core Contributors
 // SPDX-License-Identifier: Apache-2.0
 
-use anyhow::Result;
+use anyhow::{anyhow, bail, Result};
 use diem_types::{
-    access_path::AccessPath,
-    transaction::{ChangeSet, WriteSetPayload},
+    access_path::{AccessPath, Path},
+    account_config::diem_root_address,
+    account_state::AccountState,
+    transaction::{ChangeSet, TransactionStatus, Version, WriteSetPayload},
     write_set::{WriteOp, WriteSetMut},
 };
-use std::collections::{BTreeMap, BTreeSet};
+use diem_validator_interface::{DebuggerStateView, DiemValidatorInterface};
+use diem_vm::{data_cache::StateViewCache, transaction_metadata::TransactionMetadata, DiemVM};
+use move_core_types::vm_status::{KeptVMStatus, VMStatus};
+use move_vm_runtime::logging::NoContextLog;
+use std::{
+    collections::{BTreeMap, BTreeSet},
+    convert::TryFrom,
+};
+use stdlib::build_stdlib;
 use vm::CompiledModule;
 
 pub fn create_release_writeset(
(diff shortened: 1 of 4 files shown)


language/diem-tools/writeset-transaction-generator/src/release.rs
```diff
@@ -71,3 +81,95 @@ pub fn create_release_writeset(
         vec![],
     )))
 }
+
+pub fn verify_payload_change<I: DiemValidatorInterface>(
+    validator: &I,
+    block_height_opt: Option<Version>,
+    payload: WriteSetPayload,
```
sblackshear, line 88: Nit: Looks like `&WriteSetPayload` will work here?

language/diem-tools/writeset-transaction-generator/src/release.rs
```diff
@@ -71,3 +81,95 @@ pub fn create_release_writeset(
         vec![],
     )))
 }
+
+pub fn verify_payload_change<I: DiemValidatorInterface>(
+    validator: &I,
+    block_height_opt: Option<Version>,
```
sblackshear, line 87: Maybe get rid of the option here? It's not used at any call site, and ideally this should always return the same result given the same parameters.

language/diem-tools/writeset-transaction-generator/src/release.rs
```diff
@@ -71,3 +81,95 @@ pub fn create_release_writeset(
...
+                WriteOp::Deletion => {
+                    println!("Deleting deprecated module: {:?}", module_id);
+                    if old_modules.remove(&module_id).is_none() {
+                        bail!("Removing non-existant module")
+                    }
+                }
+                WriteOp::Value(v) => {
+                    let updated_module = match CompiledModule::deserialize(v.as_slice()) {
+                        Ok(m) => m,
+                        Err(e) => bail!("Unexpected module deserialzie error {:?}", e),
```
sblackshear, line 150:
```suggestion
                        Err(e) => bail!("Unexpected module deserialize error {:?}", e),
```

language/diem-tools/writeset-transaction-generator/src/release.rs
```diff
@@ -71,3 +81,95 @@ pub fn create_release_writeset(
...
+                    };
+                    println!(
+                        "Updating new module: {:?} \n {:#?}",
```
sblackshear, line 153:
```suggestion
                        "Adding new module: {:?} \n {:#?}",
```

language/diem-tools/writeset-transaction-generator/src/release.rs
```diff
@@ -71,3 +81,95 @@ pub fn create_release_writeset(
...
+        .collect::<BTreeMap<_, _>>();
+
+    for (access_path, write_op) in output.write_set() {
+        let path = bcs::from_bytes::<Path>(access_path.path.as_slice())?;
+        if let Path::Code(module_id) = path {
+            match write_op {
+                WriteOp::Deletion => {
+                    println!("Deleting deprecated module: {:?}", module_id);
+                    if old_modules.remove(&module_id).is_none() {
+                        bail!("Removing non-existant module")
```
sblackshear, line 144:
```suggestion
                        bail!("Removing non-existent module")
```

language/diem-tools/writeset-transaction-generator/src/release.rs
```diff
@@ -71,3 +81,95 @@ pub fn create_release_writeset(
         vec![],
     )))
 }
+
+pub fn verify_payload_change<I: DiemValidatorInterface>(
```
sblackshear, line 85: Nit: comment (can probably just lift some of the already-good code comments to doc ones).

language/diem-tools/writeset-transaction-generator/src/release.rs
```diff
@@ -71,3 +81,95 @@ pub fn create_release_writeset(
...
+    let state_view = DebuggerStateView::new(validator, block_height);
+    let mut vm = DiemVM::new(&state_view);
+    let cache = StateViewCache::new(&state_view);
+    let log_context = NoContextLog::new();
+    let mut txn_data = TransactionMetadata::default();
+    txn_data.sequence_number =
+        match validator.get_account_state_by_version(diem_root_address(), block_height)? {
+            Some(account) => AccountState::try_from(&account)?
+                .get_account_resource()?
+                .ok_or_else(|| anyhow!("Diem root account doesn't exists"))?
```
sblackshear, line 103:
```suggestion
                .ok_or_else(|| anyhow!("Diem root account doesn't exist"))?
```

language/diem-tools/writeset-transaction-generator/src/release.rs
```diff
@@ -71,3 +81,95 @@ pub fn create_release_writeset(
...
+    }
+
+    // Should contain a reconfiguration event
+    let new_epoch_event_key = diem_types::on_chain_config::new_epoch_event_key();
+    if !output
+        .events()
+        .iter()
+        .any(|e| *e.key() == new_epoch_event_key)
+    {
+        bail!("Outputed WriteSet won't trigger a reconfiguration")
```
sblackshear, line 127:
```suggestion
        bail!("Output WriteSet won't trigger a reconfiguration")
```

language/diem-tools/writeset-transaction-generator/src/release.rs
```diff
@@ -71,3 +81,95 @@ pub fn create_release_writeset(
...
+            }
+        }
+    }
+
+    let local_modules = build_stdlib()
+        .into_iter()
+        .map(|(_, m)| (m.self_id(), m))
+        .collect::<BTreeMap<_, _>>();
+    if local_modules.len() != old_modules.len() {
+        bail!("Remote module length differs from local module length")
```
sblackshear, line 167:
```suggestion
        bail!("Found {:?} modules locally but {:?} in remote storage", local_modules.len(), old_modules.len())
```

sblackshear (approved):
Love it!

language/diem-tools/writeset-transaction-generator/src/release.rs
runtian-zhou: Ah! There's a specific reason why I would like to make it optional. The reason is I want this function to be invoked, it would be nice if we just run these checks against the latest state known.

sblackshear: I understand that, but I'm not sure it's a good idea in this case. It's definitely annoying to have to enter the version manually, but I think in this case the friction might be a feature. Verification is always relative to a version, and the version is always moving upward. If we try to hide this from the user, it might lead to incorrect assumptions and/or hard-to-debug failures.

For example: imagine the Move oncall creates a DF release with`--latest`. AOS grabs the bytes from the release and attempts to re-validate the write set before signing it, but now validation fails. Because there's no record of the version the oncall and AOS validated against, it will be very hard to determine what went wrong. By contrast, if both know the versions, it is easy to bisect and find out where validation started failing. 

If we force the user to enter the version manually, it should always be possible to do this.

PR #7182, opened 2021-01-09
[consensus] avoid mempool long poll if root block has user txns

zekun000:
## Motivation

This commit fixes the case when the leader commits immediately while others wait for
additional mempool long poll period.

(Write your answer here.)

## Test Plan

## Related PRs


davidiw (approved):
Very clever, look around that code area for comments that don't make sense now, especially about "root (excluding)".

Also if you want to have fun in rust, why isn't this a map function: https://github.com/diem/diem/blob/master/consensus/src/txn_manager.rs#L85-L86

PR #7189, opened 2021-01-11
[transaction-replay] Implement a few commands for txn-replay tool

runtian-zhou:
…ove CLI

## Motivation

Implement three commands:
1.  An option to save account states to local storage so that they can be later verified by the Move CLI.
2. Refactored the logic for executing unsigned writeset for verification purpose, and integrate such functionality to txn-replay tool
3. A command to fetch all the key accounts on chain: including core code address, diem_root, treasuary_compliance and all validators.
4. A checked-in release artifact along with a `verify-release` for `writeset-generator` to make sure the blob generated is indeed built from the checked in artifact.

I also updated these commands in the release process runbook.

Yes

## Test Plan

1. Tested the `annotate-account` command locally for both core_code_address and assoc_root address.
2. Tested the `writeset-generator`'s release command still works, and the generated blob can be re-executed by the txn replay tool.
3. Tested the `annotate-key-accounts` with premainnet and make sure states are fetched and passes the `move-doctor`
4. Tested the generated blob from `create-release` command for `writeset-generator` can pass the `verify-release` command. The command for creating a release and verifying the release will now look like following:
`cargo run -- --output out.bin --output-payload create-release PREMAINNET <BLOCK_HEIGHT>`
`cargo run -- verify-release PREMAINNET out.bin` 

## Related PRs

#7165


.gitignore
```diff
@@ -51,3 +52,4 @@ target-out-docker
 
 # Doc generation output
 *.md.old
+/language/diem-tools/transaction-replay/storage/
```
mengxu-fb, line 55: Same as line 41?

runtian-zhou: oops. My bad

sblackshear (approved):

runtian-zhou:
Update: Fixed Sam's comment in the first commit. Add a second commit to implement a new command for `writeset-generator` to checkin and verify the release artifact.

language/diem-tools/diem-validator-interface/src/lib.rs
```diff
@@ -53,6 +50,48 @@ pub trait DiemValidatorInterface {
         }
         Ok(acc)
     }
+
+    fn get_key_accounts_by_version(
```
sblackshear, line 54: Nit: `get_admin_accounts`, `get_governance_accounts` or similar might be better? The term "key" is so overloaded that I'd prefer not to use it...

sblackshear: Also, a comment saying what this does seems useful.

language/diem-tools/writeset-transaction-generator/src/main.rs
```diff
@@ -47,6 +51,50 @@ enum Command {
...
     },
+    /// Verify if a blob is generated by the checked-in artifact.
+    #[structopt(name = "verify-release")]
+    VerifyDiemFrameworkRelease {
+        /// Path to the serialized bytes of WriteSet.
+        #[structopt(parse(from_os_str))]
+        writeset_path: PathBuf,
+        /// True if the serialized byte has the type of WriteSetPayload.
+        #[structopt(long)]
+        is_payload: bool,
```
sblackshear, line 62: This seems funky. It would probably be simpler not to allow polymorphism in the type of the blob. Is there a reason this tries to support validating both a WriteSet transaction and a payload?

language/diem-tools/writeset-transaction-generator/src/main.rs
```diff
@@ -47,6 +51,50 @@ enum Command {
...
+    },
+}
+
+/// Checkin the release artifact to a local file to make sure the release process is reproducible
+#[derive(Serialize, Deserialize)]
+struct ReleaseArtifact {
+    remote_json_rpc_url: String,
```
sblackshear, line 69: I think it would be more future-proof to use a ChainID instead of URL (both here and in the inputs to the CLI). The problem: the URLs are not stable identifiers (e.g., they have changed several times already due to the Diem renaming), whereas a chain ID is a stable identifier by definition.

language/diem-tools/writeset-transaction-generator/src/main.rs
```diff
@@ -69,24 +117,64 @@ fn main() -> Result<()> {
             execute_as,
         ),
         Command::CreateDiemFrameworkRelease { url, version } => {
-            let remote = JsonRpcDebuggerInterface::new(url.as_str())?;
-            let remote_modules = remote.get_diem_framework_modules_by_version(version)?;
-            let local_modules = build_stdlib().into_iter().map(|(_, m)| m).collect();
-            let payload = create_release_writeset(remote_modules, local_modules)?;
-            verify_payload_change(&remote, Some(version), &payload)?;
+            if let Ok(artifact) = load_artifact() {
+                if artifact.remote_json_rpc_url != url || artifact.version > version {
```
sblackshear, line 121: I think this will fail if we try to create releases for multiple deployments from the same git branch. We probably want `load_artifact` to take a `chain_id` or `url` + have separate directories for release artifacts of each deployment.

sblackshear (changes requested):

language/diem-tools/writeset-transaction-generator/src/main.rs
runtian-zhou: The main problem here is this: writeset used for AOS has different type from the writeset used for waypoint. That's why we need the output-payload flag and the is-payload flag here. Although this command is mostly used for the AOS for release process, in rare cases where we want to roll out emergency changes via waypoint this flag will be needed. E.g: if we find a critical secruity breach in Move modules on chain

language/diem-tools/writeset-transaction-generator/src/main.rs
runtian-zhou: That's right. Lemme change that

language/diem-tools/diem-validator-interface/src/lib.rs
runtian-zhou: Fixed!

runtian-zhou:
Upadte: Replace the json_rpc url with a ChainID with a builtin url mapping and save the artifact file per ChainID. Also got rid of the `is_payload` flag by using a or_else logic.

language/diem-tools/writeset-transaction-generator/src/main.rs
runtian-zhou: I replaced this flag with a either or logic with reasons given above: we want to be able to verify both types because in extreme cases we might want to create a release when the system is down, then we have to rely on the Waypoint Transaction to bring us back.

runtian-zhou: Another question is should we distinguish the two types of payload by enforcing a extension name for a blob? That we would be able to get rid of the polymorphism over there.

sblackshear: I like the extension-based approach.

sblackshear (approved):
LGTM. 

Not needed for this PR, but in the future we should create some `ValidatorInterface` mocks that will allow us to test all these flows...

runtian-zhou:
> LGTM.
> 
> Not needed for this PR, but in the future we should create some `ValidatorInterface` mocks that will allow us to test all these flows...

I've been thinking about this for a while. To test these flows we really need a mock node setup (like `diem-swarm`) locally so that we have a working json-rpc endpoint we can talk to. However we need to create some changes to the Diem Framework to test this flow and that part is non-trivial for me to automate.

sblackshear:
One idea for that: you can have a for-testing-only copy of a very stable file (e.g., `Signature.move`) that contains a small change or two. The test can copy the changed file in, make a few other easy changes (e.g., delete a module, add a new module) rebuild everything, and make sure it works.

language/diem-tools/writeset-transaction-generator/src/main.rs
```diff
@@ -68,25 +126,78 @@ fn main() -> Result<()> {
...
-            let remote_modules = remote.get_diem_framework_modules_by_version(version)?;
-            let local_modules = build_stdlib().into_iter().map(|(_, m)| m).collect();
-            let payload = create_release_writeset(remote_modules, local_modules)?;
-            verify_payload_change(&remote, Some(version), &payload)?;
+        Command::CreateDiemFrameworkRelease {
+            chain_id,
+            url,
+            version,
+        } => {
+            if let Ok(artifact) = load_artifact(&chain_id) {
```
sblackshear, line 134: It might be better to use `load_artifact?` here and manually create release `version` 0 artifacts for each chain ID?

language/diem-tools/writeset-transaction-generator/src/main.rs
```diff
@@ -42,11 +47,64 @@ enum Command {
...
+        url: String,
+        /// Path to the serialized bytes of WriteSet.
+        #[structopt(parse(from_os_str))]
+        writeset_path: PathBuf,
+    },
+}
+
+/// Checkin the release artifact to a local file to make sure the release process is reproducible
+#[derive(Serialize, Deserialize)]
+struct ReleaseArtifact {
```
sblackshear, line 73: I'm wondering if we also want something that commits to the framework code (e.g., a hash) here. Thoughts?

language/diem-tools/writeset-transaction-generator/src/main.rs
```diff
@@ -42,11 +47,64 @@ enum Command {
...
+    chain_id: ChainId,
+    // TODO: Deprecate after release
```
sblackshear, line 75: Do we need to keep this?

runtian-zhou: I thought keeping this item will be desirable because we can prevent the manual error from inputting a chain_id with a different url.

language/diem-tools/writeset-transaction-generator/src/main.rs
runtian-zhou: A hash of the current stdlib would be nice. Will add that.

language/diem-tools/writeset-transaction-generator/src/main.rs
runtian-zhou: Added a maunal flag to create the artifact for the first release.

language/diem-tools/writeset-transaction-generator/src/main.rs
runtian-zhou: PR Updated in fixup commit!

runtian-zhou:
I'm working on a e2e test of this flow in smoke test. Will put it out in a separate PR.

runtian-zhou:
Updated per @sblackshear requests:
1. Added hash for the aritfact
2. Got rid of the URL in artifact
3. Change aritfact to json for better readability
4. Create a `first-release` option to create the first release artifact on disk. If the flag is not set the release process will abort if no artifact file is found.

PR #7194, opened 2021-01-11
[for discussion only] helper code on exploring Move visibility design options

mengxu-fb:
## Motivation

There are four levels of granularity we should consider when designing the visibility scheme:

- module-to-module
  Any function in module B can access any protected function in module A
- module-to-function
  Function B::bar() can access any protected function in module A
- function-to-module
  Protected function A::foo() can be accessed by any function in module B
- function-to-function
  Protected function A::foo() can only be accessed by B::bar()

This PR does a what-if analysis, i.e., within the Diem framework, who should be declared as a friend of whom should we adopt one of the four schemes?

The results are shown in the attached files:
[module-to-module](https://github.com/diem/diem/files/5798256/viz_m2m.txt)
[module-to-function](https://github.com/diem/diem/files/5798258/viz_m2f.txt)
[function-to-module](https://github.com/diem/diem/files/5798259/viz_f2m.txt)
[function-to-function](https://github.com/diem/diem/files/5798261/viz_f2f.txt)

Here is an example (in the function-to-module file) to illustrate how to interpret these files:
```
AccountFreezing::account_is_frozen:
	DiemAccount
```
reads as: the function `AccountFreezing::account_is_frozen` should declare, and only declare, module `DiemAccount` as a friend. And if this is done, then `AccountFreezing::account_is_frozen` does not need to be `public` anymore.

You may run the code to produce these files yourself,
```
cd language/tools/move-visibility
cargo run
```


jolestar:
Can define a visibility scope, like rust's pub(crate)? Treat the modules under an account address as a crate.

mengxu-fb:
Thanks for the comment @jolestar. Yes, that is indeed one option we are exploring. But depending on the use cases, we may want sth a bit more flexible than that, e.g., with support for cross-account module linking.

mengxu-fb:
Friend visibility implementation adopts the `module-to-module` design, this PR is no longer needed.

PR #7196, opened 2021-01-11
[diem] Don't use serde::export private types

phlip9:
Serde had a `serde::export` module exposed for use in its proc-macro derive crates, though it was never intended to be used publicly.

With the latest serde release (1.0.119) the `serde::export` module was finally renamed to `serde::__private` to more clearly express that downstream crates should not depend on it: (https://github.com/serde-rs/serde/blob/master/serde/src/lib.rs#L275).

This diff removes all occurances of `use serde::export::{..}` and replaces them with their `use std::{..}` equivalents.

### DIFF
--- a/language/move-model/src/builder/module_builder.rs
+++ b/language/move-model/src/builder/module_builder.rs
@@ -1,11 +1,13 @@
 // Copyright (c) The Diem Core Contributors
 // SPDX-License-Identifier: Apache-2.0
 
-use std::collections::{BTreeMap, BTreeSet};
+use std::{
+    collections::{BTreeMap, BTreeSet},
+    fmt,
+};
 
 use itertools::Itertools;
 use regex::Regex;
-use serde::export::{fmt, Formatter};
 
 use bytecode_source_map::source_map::SourceMap;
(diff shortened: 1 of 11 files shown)


language/move-model/src/builder/module_builder.rs
```diff
@@ -1,11 +1,13 @@
...
 
-use std::collections::{BTreeMap, BTreeSet};
+use std::{
+    collections::{BTreeMap, BTreeSet},
+    fmt,
+};
 
 use itertools::Itertools;
 use regex::Regex;
-use serde::export::{fmt, Formatter};
```
gregnazario, line 8: I think it's funny because my IDE seems to default to this (probably why there's a bunch of these)

gregnazario (approved):
Thanks for cleaning it up and making it more consistent.

language/move-model/src/builder/module_builder.rs
dtolnay: That needs to get reported as a bug to whatever IDE or language plugin you are using. It should not be inserting imports of hidden paths.

PR #7214, opened 2021-01-12
[gha] testing

rexhoffman:
## Motivation

Check tests on gha in master w/o changes.

Y

## Test Plan

CI

## Related PRs

None


rexhoffman:
So looks like removing the slash I added to work around another problem now works...   I was very bleary eyed when I updated that to use that slash, but hard to argue with the passing tests (as far as we trust the test).

.github/workflows/ci-test.yml
```diff
@@ -14,6 +14,7 @@ env:
   max_threads: 16
   pre_command: cd /opt/git/diem/
 
+
```
bmwill, line 17: remove this extra line before merging

bmwill (approved):

PR #7216, opened 2021-01-12
[for discussion only][vm] fully backword compatible bytecode format with function visibility info

mengxu-fb:
## Motivation

Here is a fully backward compatible bytecode format for the function visibility scheme I have in mind. I'll explain later what do I mean by fully-backward compatible.

But before that, this PR is based on the following assumptions:
- We prefer callee-side friend list declaration (so we don’t need to look at multiple files to know who can call a particular function).
- Module-level friend list seems to be a reasonable granularity to start with.

In other words, the friend list will be embedded in the function definition bytecode. This is similar to what is currently done (i.e., the `is_public` bit is stored in the function definition bytecode).

Now, here is the explanation on why this change is backward compatible:

The current layout for the bytecode of a function definition:
```
| ULEB128: function_index | U8: flags | Vec<ULEB128>: acquires_list | Optional<Vec<U8>>: code_unit |
```
Note that this `flags` is a bit mysterious here, as it is essentially capturing two information in the current scheme
1. Whether this function `is_public`
2. Whether this function `is_native`

Decomposing this flag, we get essentially the following:
- For public functions:
```
| ULEB128: function_index | U8: (0x1 | is_native) | Vec<ULEB128>: acquires_list | Optional<Vec<U8>>: code_unit |
```
- For private functions:
```
| ULEB128: function_index | U8: (0x0 | is_native) | Vec<ULEB128>: acquires_list | Optional<Vec<U8>>: code_unit |
```

Now here is the new layout in this PR:
```
| ULEB128: function_index | U8: flags | Optional<Vec<ULEB128>>: friends | Vec<ULEB128>: acquires_list | Optional<Vec<U8>>: code_unit |
```

Note that a new section, `| Optional<Vec<ULEB128>>: friends |` is added, which maps to a vector of `ModuleHandleIndex`. But this section is Optional, meaning that it only exists when `flags & IS_PROTECTED_BIT != 0`, and if we further unroll the layout casing on the visibility flag, it means:

- For public functions:
```
| ULEB128: function_index | U8: (0x1 | is_native) | Vec<ULEB128>: acquires_list | Optional<Vec<U8>>: code_unit |
```
- For private functions:
```
| ULEB128: function_index | U8: (0x0 | is_native) | Vec<ULEB128>: acquires_list | Optional<Vec<U8>>: code_unit |
```
- For protected functions:
```
| ULEB128: function_index | U8: (0x4 | is_native) | Vec<ULEB128>: friends | Vec<ULEB128>: acquires_list | Optional<Vec<U8>>: code_unit |
```

Note that in the `public` and `private` case, the new and old layout are exactly the same! Hence, the scheme is fully backward compatible.

### DIFF
--- a/language/vm/src/file_format.rs
+++ b/language/vm/src/file_format.rs
@@ -363,6 +363,7 @@ impl StructDefinition {
         }
     }
 }
+
 /// A `FieldDefinition` is the definition of a field: its name and the field type.
 #[derive(Clone, Debug, Eq, PartialEq)]
 #[cfg_attr(any(test, feature = "fuzzing"), derive(Arbitrary))]
@@ -374,9 +375,28 @@ pub struct FieldDefinition {
     pub signature: TypeSignature,
 }
 
+/// `VisibilityScope` restricts who may call a given function.
+#[derive(Clone, Debug, Eq, PartialEq)]
+#[cfg_attr(any(test, feature = "fuzzing"), derive(Arbitrary))]
+#[cfg_attr(any(test, feature = "fuzzing"), proptest(no_params))]
+pub enum VisibilityScope {
+    /// The function can only be called within the module that defines it.
+    Private,
+    /// The function can be called by any module and script.
+    Public,
+    /// Similar to "private", but in addition, the function can also be called
+    /// by any module specified in the list of module handles.
+    Protected(Vec<ModuleHandleIndex>),
+}
+
+impl Default for VisibilityScope {
+    fn default() -> Self {
+        VisibilityScope::Private
+    }
+}
+
 /// A `FunctionDefinition` is the implementation of a function. It defines
 /// the *prototype* of the function and the function body.
-
 #[derive(Clone, Debug, Default, Eq, PartialEq)]
 #[cfg_attr(any(test, feature = "fuzzing"), derive(Arbitrary))]
 #[cfg_attr(any(test, feature = "fuzzing"), proptest(params = "usize"))]
@@ -384,7 +404,7 @@ pub struct FunctionDefinition {
     /// The prototype of the function (module, name, signature).
     pub function: FunctionHandleIndex,
     /// Flag to indicate if this function is public.
-    pub is_public: bool,
+    pub visibility: VisibilityScope,
     /// List of nominal resources (declared in this module) that the procedure might access
     /// Either through: BorrowGlobal, MoveFrom, or transitively through another procedure
     /// This list of acquires grants the borrow checker the ability to statically verify the safety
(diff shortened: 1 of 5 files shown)


tnowacki:
Why is it `Optional<Vec<ULEB128>>: friends` and not `Vec<ULEB128>` ?

mengxu-fb:
The reason is because `Public` and `Private` do not have a friend list, and hence, do not have the `Vec<ULEB128>`. Adding an empty `Vec<ULEB128>` will break compatibility because in the old format, that is the position for `Vec<ULEB128>: acquires_list`.

mengxu-fb:
Ah, maybe I missed this, serializing an empty `Vec<ULEB128>` does not yield zero bytes. It actually yield one byte, which is the count (although it is 0) of the `Vec`.

language/vm/src/file_format.rs
```diff
@@ -374,17 +375,36 @@ pub struct FieldDefinition {
...
-
 #[derive(Clone, Debug, Default, Eq, PartialEq)]
 #[cfg_attr(any(test, feature = "fuzzing"), derive(Arbitrary))]
 #[cfg_attr(any(test, feature = "fuzzing"), proptest(params = "usize"))]
 pub struct FunctionDefinition {
     /// The prototype of the function (module, name, signature).
     pub function: FunctionHandleIndex,
     /// Flag to indicate if this function is public.
-    pub is_public: bool,
+    pub visibility: VisibilityScope,
```
mengxu-fb, line 407: Another clarification is that: the description of this PR shows how the bytecode layout looks like. The actual field in the `FunctionDefinition` Rust struct is here.

tnowacki:
I think the Option the option thing is starting to make sense, thanks!

I'll comment on the code for some other things

language/vm/src/file_format.rs
```diff
@@ -374,17 +375,36 @@ pub struct FieldDefinition {
     pub signature: TypeSignature,
 }
 
+/// `VisibilityScope` restricts who may call a given function.
+#[derive(Clone, Debug, Eq, PartialEq)]
+#[cfg_attr(any(test, feature = "fuzzing"), derive(Arbitrary))]
+#[cfg_attr(any(test, feature = "fuzzing"), proptest(no_params))]
+pub enum VisibilityScope {
```
tnowacki, line 382: nit, I'd just name this `Visibility`

language/vm/src/file_format.rs
```diff
@@ -374,17 +375,36 @@ pub struct FieldDefinition {
...
+    /// The function can only be called within the module that defines it.
+    Private,
+    /// The function can be called by any module and script.
+    Public,
+    /// Similar to "private", but in addition, the function can also be called
+    /// by any module specified in the list of module handles.
+    Protected(Vec<ModuleHandleIndex>),
```
tnowacki, line 389: I'd probably break this out to 

`Protected(VisibilityListIndex)`

and then somewhere else
```
define_index! {
    name: VisibilityListIndex,
    kind: VisibilityList,
    doc: "Index into the `VisibilityList` table.",
}
struct VisibilityList {
    modules: Vec<ModuleHandleIndex>
}
```

I imagine a lot of functions will have the same set of valid callers. 
It will also reduce the diff size on updates to the list

language/vm/src/file_format.rs
```diff
@@ -404,19 +424,27 @@ pub struct FunctionDefinition {
...
+    /// Returns whether the FunctionDefinition can be called from a module other than the module
+    /// that defines it.
+    ///
+    /// TODO: following the definition above (which is inherited from the old public-private binary
+    /// model), the `protected` scope should also be considered as `is_public`. We need to refactor
+    /// the function name as well as its call stack.
     pub fn is_public(&self) -> bool {
-        self.is_public
+        matches!(&self.visibility, VisibilityScope::Public | VisibilityScope::Protected(_))
     }
```
tnowacki, line 435: If you were actually going to land this bit by bit, it should return false on protected.

In actuality, I'd prefer to see a larger PR that refactored this into `pub fn visiblity(&self) -> &Visibility` 

This function is probably called in just one spot in the whole codebase (at least just a handful) so the refactor shouldn't be bad

mengxu-fb: No, I am not going to land this :)

This is just for a proof-of-concept only and you are absolutely right that we need to make sure things are all well-defined and cleared up before making the changes.

PR #7217, opened 2021-01-13
[network] Metric if network identity doesn't match onchain

gregnazario:
It's possible that a user could set their network identity onchain and
then continue communicating on the network with the wrong network
identity.  This causes a strange 1 way communication, which makes it
very confusing.  The metric allows us to know that there's an issue and that we can mitigate it faster

### DIFF
--- a/network/simple-onchain-discovery/src/builder.rs
+++ b/network/simple-onchain-discovery/src/builder.rs
@@ -4,6 +4,7 @@
 use crate::ConfigurationChangeListener;
 use channel::diem_channel;
 use diem_config::network_id::NetworkContext;
+use diem_crypto::x25519::PublicKey;
 use diem_network_address_encryption::Encryptor;
 use diem_types::on_chain_config::OnChainConfigPayload;
 use network::connectivity_manager::ConnectivityRequest;
@@ -12,6 +13,7 @@ use tokio::runtime::Handle;
 
 struct ConfigurationChangeListenerConfig {
     network_context: Arc<NetworkContext>,
+    expected_pubkey: PublicKey,
     encryptor: Encryptor,
     conn_mgr_reqs_tx: channel::Sender<ConnectivityRequest>,
     reconfig_events: diem_channel::Receiver<(), OnChainConfigPayload>,
@@ -20,12 +22,14 @@ struct ConfigurationChangeListenerConfig {
 impl ConfigurationChangeListenerConfig {
     fn new(
         network_context: Arc<NetworkContext>,
+        expected_pubkey: PublicKey,
         encryptor: Encryptor,
         conn_mgr_reqs_tx: channel::Sender<ConnectivityRequest>,
         reconfig_events: diem_channel::Receiver<(), OnChainConfigPayload>,
     ) -> Self {
         Self {
             network_context,
+            expected_pubkey,
             encryptor,
             conn_mgr_reqs_tx,
             reconfig_events,
@@ -49,6 +53,7 @@ pub struct ConfigurationChangeListenerBuilder {
 impl ConfigurationChangeListenerBuilder {
     pub fn create(
         network_context: Arc<NetworkContext>,
+        expected_pubkey: PublicKey,
         encryptor: Encryptor,
         conn_mgr_reqs_tx: channel::Sender<ConnectivityRequest>,
         reconfig_events: diem_channel::Receiver<(), OnChainConfigPayload>,
@@ -56,6 +61,7 @@ impl ConfigurationChangeListenerBuilder {
         Self {
             config: Some(ConfigurationChangeListenerConfig::new(
                 network_context,
+                expected_pubkey,
                 encryptor,
                 conn_mgr_reqs_tx,
                 reconfig_events,
@@ -71,6 +77,7 @@ impl ConfigurationChangeListenerBuilder {
         let config = self.config.take().expect("Listener must be configured");
         self.listener = Some(ConfigurationChangeListener::new(
             config.network_context,
+            config.expected_pubkey,
             config.encryptor,
             config.conn_mgr_reqs_tx,
             config.reconfig_events,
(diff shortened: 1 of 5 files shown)


davidiw (approved):
nice, it is really sad that it is so much more work to improve the listener to handle migrating to a new network key :/

gregnazario:
The two E2E tests that are failing in question, don't seem to fail locally for me, and actually always succeed.  So, it seems there's some sort of race condition or flakiness in these tests around this new functionality.  I'm concerned that it may cause the nodes to panic over and over again unless, it's just something wrong with how the tests are set up @davidiw thoughts?
* `genesis::test_genesis_transaction_flow`
* `operational_tooling::test_network_key_rotation_recovery`

network/simple-onchain-discovery/src/builder.rs
```diff
@@ -71,6 +77,7 @@ impl ConfigurationChangeListenerBuilder {
         let config = self.config.take().expect("Listener must be configured");
         self.listener = Some(ConfigurationChangeListener::new(
             config.network_context,
+            config.expected_pubkey,
```
phlip9, line 80: can't comment on the exact line, but here:

```rust
    pub fn start(&mut self, executor: &Handle) -> &mut Self {
        assert_eq!(self.state, State::BUILT);
        self.state = State::STARTED;
        let listener = self.listener.take().expect("Listener must be built");
>>>>    executor.spawn(listener.start());    <<<<
        self
    }
```

you can see we're spawning a task `listener.start()` but not holding onto the `JoinHandle`. This means if the change listener panics, the task will print the error and shutdown, but the panic won't propagate and bring down the node. In fact, it will be worse, because the change listener will never restart and we'll just never get discovery updates ever again...

FWIW, you have the exact same problem if you use `std::thread::spawn` and just drop the `JoinHandle`; it won't propagate any panics crossing the thread boundary since you've detached the thread.

to be honest, our graceful shutdown process in diem-node... doesn't really exist : p

it seems the "proper" approach would be to look at each of the `executor::spawn` and `thread::spawn` points across the codebase, and decide "is a panic in this module recoverable?" if yes, it's probably fine to just drop the `JoinHandle` as we do already. in some places, we would probably want to restart the task. Otherwise, we need to hold onto the `JoinHandle`s and `std::panic::resume_unwind` the panics.

gregnazario: Yeah, in this case we want to propagate the panic.

As a whole, I believe we're avoiding unwinding panics, and just taking a panic as a reason to restart the whole process.  Otherwise, the `panic!` -> `unwind` pattern is basically like using a try catch in another language, which I'm pretty sure rust was trying to avoid.

phlip9: Right, I would personally err on just propagating the panics everywhere, with the possible exception of a few isolated places like the `Peer` actor or transport upgrades, which is our first line-of-defense against untrusted input. Most of all, catching panics makes it a lot harder for our test infra to work properly, since it hides most latent bugs.

network/simple-onchain-discovery/src/lib.rs
```diff
@@ -215,3 +234,107 @@ impl ConfigurationChangeListener {
...
+            .await
+            .expect("Shouldn't timeout");
+        };
+
+        // Ensure we get the right panic
+        if let Err(err) = block_on(runtime.spawn(listener_future)) {
+            assert_eq!(
+                "Onchain pubkey differs from local pubkey",
+                *err.into_panic().downcast::<&'static str>().unwrap()
+            );
```
phlip9, line 302: I think you can decorate the test to expect a specific panic message like:

```rust
#[test]
#[should_panic(expected = "")]
fn restart_if_key_mismatch() {
    // ..
    block_on(listener_future);
}
```

then the test will fail if it doesn't panic or the message doesn't match : )

network/simple-onchain-discovery/src/lib.rs
```diff
@@ -215,3 +234,107 @@ impl ConfigurationChangeListener {
...
+            peer_id,
+            consensus_pubkey,
+            different_pubkey,
+            &mut reconfig_tx,
+        );
+
+        let listener_future = async move {
+            // Run the test, ensuring we actually stop after a couple seconds in case it fails to fail
+            timeout_at(
+                tokio::time::Instant::from(Instant::now() + Duration::from_secs(2)),
```
phlip9, line 290: nit: let's increase the timeout here to 10sec just to be safe

network/simple-onchain-discovery/src/lib.rs
```diff
@@ -167,6 +170,22 @@ impl ConfigurationChangeListener {
...
 
+        // Ensure that the public key matches what's onchain for this peer
+        if let Some(ConnectivityRequest::UpdateEligibleNodes(_, peer_updates)) = updates
+            .iter()
+            .find(|requests| matches!(requests, ConnectivityRequest::UpdateEligibleNodes(_, _)))
+        {
+            if let Some(pubkeys) = peer_updates.get(&self.network_context.peer_id()) {
+                if !pubkeys.contains(&self.expected_pubkey) {
+                    error!(
+                        "Onchain pubkey {:?} differs from local pubkey {}",
```
phlip9, line 181: nit: add network context to error log

network/simple-onchain-discovery/src/lib.rs
gregnazario: Originally, I was doing a join which was wrapping the panic (why I did it this way), so I probably don't have to do it anymore, now that I don't have the join.

network/simple-onchain-discovery/src/lib.rs
gregnazario: If it takes 10 seconds to run this test by itself I think we've got bigger problems?

davidiw (approved):
at some point, we're probably just going to have to introduce secure storage into this and figure out a nice encapsulation around onchain config updates and secure storage so that we can upgrade noise listener more conveniently....

PR #7218, opened 2021-01-13
[move-prover] Add syntax for first-class quantifiers

matbd:
## Motivation

Currently, quantified formulas are represented as a call of the form `$spec_{all,any}(range_expr, lambda_expr)` for each quantified variable. This is limiting our ability to generate accurate triggers.

In this PR, we make quantifiers a proper AST node in Move specs and the Move model AST.
The new AST nodes support quantification of multiple variables but this is not parsed or translated as such yet.

## Test Plan

CI

### DIFF
--- a/language/move-lang/src/expansion/ast.rs
+++ b/language/move-lang/src/expansion/ast.rs
@@ -3,7 +3,7 @@
 
 use crate::{
     parser::ast::{
-        BinOp, ConstantName, Field, FunctionName, FunctionVisibility, Kind, ModuleIdent,
+        BinOp, ConstantName, Field, FunctionName, FunctionVisibility, Kind, ModuleIdent, QuantKind,
         ResourceLoc, SpecApplyPattern, SpecBlockTarget, SpecConditionKind, StructName, UnaryOp,
         Var,
     },
@@ -215,6 +215,11 @@ pub type LValue = Spanned<LValue_>;
 pub type LValueList_ = Vec<LValue>;
 pub type LValueList = Spanned<LValueList_>;
 
+pub type LValueWithRange_ = (LValue, Exp);
+pub type LValueWithRange = Spanned<LValueWithRange_>;
+pub type LValueWithRangeList_ = Vec<LValueWithRange>;
+pub type LValueWithRangeList = Spanned<LValueWithRangeList_>;
+
 #[derive(Debug, Clone, PartialEq)]
 #[allow(clippy::large_enum_variant)]
 pub enum ExpDotted_ {
@@ -256,7 +261,8 @@ pub enum Exp_ {
     While(Box<Exp>, Box<Exp>),
     Loop(Box<Exp>),
     Block(Sequence),
-    Lambda(LValueList, Box<Exp>), // spec only
+    Lambda(LValueList, Box<Exp>),                    // spec only
+    Quant(QuantKind, LValueWithRangeList, Box<Exp>), // spec only
 
     Assign(LValueList, Box<Exp>),
     FieldMutate(Box<ExpDotted>, Box<Exp>),
@@ -806,6 +812,13 @@ impl AstDebug for Exp_ {
                 w.write(" ");
                 e.ast_debug(w);
             }
+            E::Quant(kind, sp!(_, rs), e) => {
+                kind.ast_debug(w);
+                w.write(" ");
+                rs.ast_debug(w);
+                w.write(": ");
+                e.ast_debug(w);
+            }
             E::ExpList(es) => {
                 w.write("(");
                 w.comma(es, |w, e| e.ast_debug(w));
@@ -950,5 +963,26 @@ impl AstDebug for LValue_ {
                 w.write("}");
             }
         }
+    }
+}
+
+impl AstDebug for Vec<LValueWithRange> {
+    fn ast_debug(&self, w: &mut AstWriter) {
+        let parens = self.len() != 1;
+        if parens {
+            w.write("(");
+        }
+        w.comma(self, |w, b| b.ast_debug(w));
+        if parens {
+            w.write(")");
+        }
+    }
+}
+
+impl AstDebug for (LValue, Exp) {
+    fn ast_debug(&self, w: &mut AstWriter) {
+        self.0.ast_debug(w);
+        w.write(" in ");
+        self.1.ast_debug(w);
     }
 }
--- a/language/move-model/src/builder/exp_translator.rs
+++ b/language/move-model/src/builder/exp_translator.rs
@@ -16,7 +16,7 @@ use move_lang::{
 };
 
 use crate::{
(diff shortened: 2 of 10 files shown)


language/move-lang/src/expansion/ast.rs
```diff
@@ -256,7 +261,8 @@ pub enum Exp_ {
     While(Box<Exp>, Box<Exp>),
     Loop(Box<Exp>),
     Block(Sequence),
-    Lambda(LValueList, Box<Exp>), // spec only
+    Lambda(LValueList, Box<Exp>),                    // spec only
+    Quant(QuantKind, LValueWithRangeList, Box<Exp>), // spec only
```
wrwg, line 265: Wondering whether we should right away add triggers here. That would be another Vec<Exp>. I think we want to have the power to do triggers manually, though hopefully, in most cases the user  can rely on some automatic generation.

language/move-model/src/builder/exp_translator.rs
```diff
@@ -1785,6 +1788,94 @@ impl<'env, 'translator, 'module_translator> ExpTranslator<'env, 'translator, 'mo
...
+                }
+                EA::LValue_::Unpack(..) | EA::LValue_::Var(..) => self.error(
+                    &loc,
+                    "[current restriction] tuples not supported in quantifiers",
+                ),
+            }
+        }
+        let rty = self.check_type(
+            loc,
+            &Type::new_prim(PrimitiveType::Bool),
```
wrwg, line 1865: nit: there is a constant BOOL_TYPE in type.rs

language/move-prover/src/spec_translator.rs
```diff
@@ -1890,15 +1902,15 @@ impl<'env> SpecTranslator<'env> {
             return self.translate_exp(exp);
         }
         let loc = self.module_env().env.get_node_loc(node_id);
-        if let Some((name, binding)) = self.get_decl_var(&loc, vars) {
-            let name_str = self.module_env().symbol_pool().string(name);
+        if let [var] = vars {
```
wrwg, line 1905: Did not know this is possible!

language/move-prover/src/spec_translator.rs
```diff
@@ -2134,7 +2144,14 @@ impl<'env> SpecTranslator<'env> {
...
-    fn translate_all_or_exists(&self, loc: &Loc, is_all: bool, args: &[Exp]) {
+    fn translate_all_or_exists(
+        &self,
+        loc: &Loc,
+        kind: QuantKind,
+        var: &LocalVarDecl,
+        range: &Exp,
+        exp: &Exp,
+    ) {
         // all(v, |x| x > 0) -->
```
wrwg, line 2155: Those comments are now obsolete

language/move-prover/src/spec_translator.rs
```diff
@@ -2147,112 +2164,90 @@ impl<'env> SpecTranslator<'env> {
...
+        };
+        let var_name = self.module_env().symbol_pool().string(var.name);
+        let quant_var = self.fresh_var_name("i");
+        let mut is_vector = false;
+        let mut is_domain: Option<Type> = None;
+        match quant_ty {
+            Type::Vector(..) => is_vector = true,
+            Type::TypeDomain(t) => is_domain = Some(t.as_ref().clone()),
+            Type::Primitive(PrimitiveType::Range) => (),
+            Type::Reference(_, b) => {
```
wrwg, line 2180: nit: you can do `match quant_ty.skip_references()`. In any case, you could also assert on that this is not happening, because the spec language does not allow it. (We may have references from locals coming from the Move lang only.)

wrwg (approved):
Wow! I left a comment to perhaps add manual triggers right away, but perhaps can be done in a followup PR, because one further pass to support multiple bindings is needed anyway.

shazqadeer:
Great stuff, @matbd !!

language/move-prover/src/spec_translator.rs
matbd: Me neither, I found it by reading this file! Slice patterns were added in Rust 1.26 apparently.

language/move-lang/src/expansion/ast.rs
matbd: I wanted to discuss some options with you all, now that I understand the details a bit more.
I also need to finish parsing/translating quantifications over multiple variables.

So it seemed reasonable to review/land this chunk of commits and open new PR(s).

language/move-prover/src/spec_translator.rs
matbd: Thanks, I meant to ask about this actually. The current type checking (before and after PR) doesn't allow references indeed, so maybe I'll just remove this case.

language/move-lang/src/expansion/ast.rs
```diff
@@ -215,6 +215,11 @@ pub type LValue = Spanned<LValue_>;
 pub type LValueList_ = Vec<LValue>;
 pub type LValueList = Spanned<LValueList_>;
 
+pub type LValueWithRange_ = (LValue, Exp);
```
matbd, line 218: If that's ok, I'm going to rename `LValueWithRange` into `ConstrainedLValue`. (I realize now that number range is just a special case, next to type domain and vectors)

language/move-model/src/exp_rewriter.rs
```diff
@@ -67,6 +67,21 @@ impl<'env, 'rewriter> ExpRewriter<'env, 'rewriter> {
...
             }
+            Quant(id, kind, ranges, body) => {
+                self.shadowed
+                    .push_front(ranges.iter().map(|(decl, _)| decl.name).collect());
+                let res = Quant(
+                    self.rewrite_attrs(*id),
+                    *kind,
+                    ranges
+                        .iter()
+                        .map(|(decl, range)| (decl.clone(), self.rewrite(range)))
```
matbd, line 78: Actually, the "ranges" probably need to be computed before `self.shadowed.push_front(..)`

language/move-model/src/builder/exp_translator.rs
matbd: It looks like this constant is (uniformly) not used in this file yet.

language/move-prover/src/spec_translator.rs
matbd: Turns out the `.skip_reference()` is useful.

language/move-model/src/exp_rewriter.rs
matbd: (fixed)

language/move-lang/src/expansion/ast.rs
matbd: Looking at the [wikipedia page for Quantifiers](https://en.wikipedia.org/wiki/Quantifier_(logic)), the two appropriate expressions seem to be "domain of discourse" and "range of quantification".

matbd: I guess the current name is fine.

PR #7221, opened 2021-01-13
[move prover] implement pragma friend restricting calling context

emmazzz:
## Motivation
This PR implements a new function level spec `pragma friend`, hoping to simulate the friend visibility feature in Move before it is available. Right now this pragma is only used in `DiemTimestamp`.

If a function `f` has `pragma friend = g` in its spec block, then:
- `f` will not be verified on its own(the same effect as `pragma verify = false`).
- When `f` is supposed to be verified against some global invariants, its friend will be verified instead.
- `f` can't be opaque.
- Functions other than `g` can't call `f`.

Yes.

## Test Plan

Added two new tests.

### DIFF
--- a/language/move-prover/src/lib.rs
+++ b/language/move-prover/src/lib.rs
@@ -69,10 +69,11 @@ pub fn run_move_prover<W: WriteColor>(
 ) -> anyhow::Result<()> {
     let now = Instant::now();
     let target_sources = find_move_filenames(&options.move_sources, true)?;
-    let other_sources = remove_sources(
+    let all_sources = collect_all_sources(
         &target_sources,
-        find_move_filenames(&options.move_deps, true)?,
-    );
+        &find_move_filenames(&options.move_deps, true)?,
+    )?;
+    let other_sources = remove_sources(&target_sources, all_sources);
     let address = Some(options.account_address.as_ref());
     debug!("parsing and checking sources");
     let mut env: GlobalEnv = run_model_builder(target_sources, other_sources, address)?;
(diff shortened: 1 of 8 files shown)


language/move-prover/src/lib.rs
```diff
@@ -313,18 +317,50 @@ fn verification_analysis(env: &mut GlobalEnv, targets: &FunctionTargetsHolder) {
...
+                        {
+                            env.error(&loc, &format!("function {} is called by other functions while it can only be called by its friend {}",
+                                                     func_env.get_name_string(),
+                                                     func_env.get_friend_name().unwrap())
+                            );
+                        }
+                    } else {
+                        // The function's friend is not in the environment, either because
+                        // the friend is not yet published or the user made a typo.
+                        env.error(&loc, &format!("function {} is called by other functions while it can only be called by its friend {}",
```
wrwg, line 344: This should be a different error message.

language/move-prover/src/lib.rs
```diff
@@ -313,18 +317,50 @@ fn verification_analysis(env: &mut GlobalEnv, targets: &FunctionTargetsHolder) {
...
+                    env.error(&loc, "function with a friend cannot be declared as opaque");
+                }
+                let calling_functions = func_env.get_calling_functions();
+
+                // If a function is called by some caller, check that the caller
+                // is indeed its friend.
+                if !calling_functions.is_empty() {
+                    if let Some(friend_env) = func_env.get_friend_env() {
+                        // The function's friend is in the environment.
+                        if calling_functions.len() != 1
```
wrwg, line 333: This is interesting. We currently can have only one `pragma friend`, though (I think) the syntax allows for repeated friend declarations. This restriction is OK until we have full visibility, but perhaps we should provide an error message at the friend declaration when there are multiple ones.

language/move-prover/tests/sources/functional/friend_error.move
```diff
@@ -0,0 +1,28 @@
+module TestFriendError {
```
wrwg, line 1: Perhaps have a test with more than one friend declaration.

wrwg (approved):
Nice!

PR #7231, opened 2021-01-14
[network] Fix incorrect usage of `select_next_some` in onchain discovery 

phlip9:
`stream.select_next_some()` returns a `FusedFuture`, which expects callers to check `is_terminated` before polling to ensure callers don't poll the future after it's completed.

In this case, `simple-onchain-discovery` was incorrectly using `select_next_some` inside a loop, expecting that the `reconfig_events` channel would never close. However, it appears that we get occasional panics during node shutdown, as referenced by this comment in diem-node:

> Shutdown network runtimes to avoid panic error log after DiemHandle is dropped:
> thread ‘network-’ panicked at ‘SelectNextSome polled after terminated’,...
> stack backtrace:
>    ......
>    8: network_simple_onchain_discovery::ConfigurationChangeListener::start::{{closure}}
>      at network/simple-onchain-discovery/src/lib.rs:175
>    ......
> Other runtimes don't have same problem.

Then, remove the `DiemHandle::shutdown`, which is no longer necessary after this change.

### DIFF
--- a/diem-node/src/lib.rs
+++ b/diem-node/src/lib.rs
@@ -48,26 +48,10 @@ pub struct DiemHandle {
     _rpc: Runtime,
     _mempool: Runtime,
     _state_synchronizer: StateSynchronizer,
-    network_runtimes: Vec<Runtime>,
+    _network_runtimes: Vec<Runtime>,
     _consensus_runtime: Option<Runtime>,
     _debug: NodeDebugService,
     _backup: Runtime,
-}
-
-impl DiemHandle {
-    pub fn shutdown(&mut self) {
-        // Shutdown network runtimes to avoid panic error log after DiemHandle is dropped:
-        // thread ‘network-’ panicked at ‘SelectNextSome polled after terminated’,...
-        // stack backtrace:
-        //    ......
-        //    8: network_simple_onchain_discovery::ConfigurationChangeListener::start::{{closure}}
-        //      at network/simple-onchain-discovery/src/lib.rs:175
-        //    ......
-        // Other runtimes don't have same problem.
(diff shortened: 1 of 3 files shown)


libra-action:

Breaking changes policy:

2. Adding fields is ok.

diem-node/src/lib.rs
```diff
@@ -48,28 +48,12 @@ pub struct DiemHandle {
     _rpc: Runtime,
     _mempool: Runtime,
     _state_synchronizer: StateSynchronizer,
-    network_runtimes: Vec<Runtime>,
+    _network_runtimes: Vec<Runtime>,
     _consensus_runtime: Option<Runtime>,
     _debug: NodeDebugService,
     _backup: Runtime,
 }
```
gregnazario, line 55: What's the reason to keep around DiemHandle, if everything put into it is an unused variable?

gregnazario (commented):
Could this be the reason why there are a bunch of flaky tests around shutdown of the DiemNode?

diem-node/src/lib.rs
phlip9: the `DiemHandle` is like an Arc; as long as you own it, the node will keep running. Drop it and it will shutdown.

phlip9:
> Could this be the reason why there are a bunch of flaky tests around shutdown of the DiemNode?

Definitely a possibility. We also don't really have any graceful shutdown logic and lots of our code relies on channels staying live at all times, which isn't true during tear-down. : p

gregnazario (approved):

xli (approved):
Thank you for fixing it!

PR #7233, opened 2021-01-14
[network] Remove `NetworkProvider`.

phlip9:
1. Remove the intermediate `NetworkProvider` channels and tasks forwarding between channels. This reduces the number of tasks per peer to 2 (Peer actor, Peer writer task).

2. Replace the `Peer{Request,Notification}` types with `Network{Request,Notification}`. Then rename `Network{Request,Notification}` to `Peer{Request,Notification}`.

3. Remove `NetworkProvider` and `interface` module. Just use the `Peer` constructor directly.

4. Move peer shutdown logic into separate method.


network/src/peer/mod.rs
```diff
@@ -46,14 +46,22 @@ mod test;
...
+/// Requests [`Peer`] receives from the `PeerManager`.
 #[derive(Debug)]
 pub enum PeerRequest {
-    // Temporary message types until we can remove NetworkProvider and use the
-    // diem_channel::Receiver<ProtocolId, NetworkRequest> directly.
-    SendDirectSend(Message),
+    /// Send an RPC request to peer.
     SendRpc(OutboundRpcRequest),
+    /// Fire-and-forget style message send to peer.
+    SendMessage(Message),
```
gregnazario, line 55: Any reason to rename DirectSend to just message?

network/src/peer/mod.rs
```diff
@@ -64,27 +72,14 @@ pub enum DisconnectReason {
...
-            match self {
-                DisconnectReason::Requested => "Requested",
-                DisconnectReason::ConnectionLost => "ConnectionLost",
-            }
-        )
+        let s = match self {
+            DisconnectReason::Requested => "Requested",
+            DisconnectReason::ConnectionLost => "ConnectionLost",
+        };
+        write!(f, "{}", s)
```
gregnazario, line 79: I don't know if this was my fault earlier, but really this should probably be the `as_str` then `display` functionality.  But, let's not bother changing that.

network/src/peer/mod.rs
```diff
@@ -195,120 +194,76 @@ where
...
-                        maybe_response = self.inbound_rpcs.next_completed_response() => {
-                            if let Err(err) = self.inbound_rpcs.send_outbound_response(&mut write_reqs_tx, maybe_response).await {
+        let reason = loop {
+            if let State::ShuttingDown(reason) = self.state {
+                break reason;
+            }
+
+            futures::select! {
+                // Handle a new outbound request from the PeerManager.
+                maybe_request = self.peer_reqs_rx.next() => {
```
gregnazario, line 216: This isn't particular to this PR, but I like to call these `futures::select!` blocks we have nesting hell.  Since we have so much nesting it goes on and on and on.

After all these refactors, we should probably come up with some standardization on how to generally do these, e.g. each block is a function, or whatever.

network/src/peer/mod.rs
```diff
@@ -447,16 +402,14 @@ where
                 }
                 ReadError::IoError(_) => {
                     // IoErrors are mostly unrecoverable so just close the connection.
-                    self.close_connection(DisconnectReason::ConnectionLost);
+                    self.shutdown(DisconnectReason::ConnectionLost);
```
gregnazario, line 405: So, do we no longer close the connection?  Oh, nevermind I see when it comes around on the shutdown it closes it.

network/src/peer/mod.rs
phlip9: `SendDirectSend` is definitely clearer, i'll rename : )

network/src/protocols/rpc/mod.rs
```diff
@@ -191,9 +192,9 @@ impl InboundRpcs {
     }
 
     /// Handle a new inbound `RpcRequest` message off the wire.
-    pub async fn handle_inbound_request(
+    pub fn handle_inbound_request(
         &mut self,
-        peer_notifs_tx: &mut channel::Sender<PeerNotification>,
+        peer_notifs_tx: &mut diem_channel::Sender<ProtocolId, PeerNotification>,
```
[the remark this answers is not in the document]
phlip9: this is definitely not super clear... `channel` is a single async FIFO queue that provides backpressure, i.e., `tx.send(msg).await` will wait until there is capacity before finally enqueuing. In contrast, `diem_channel` is a set of non-async queues (one per `ProtocolId` in this case). these queues _don't_ provide backpressure, i.e., `tx.send(msg)` (note the lack of `.await`) will just drop a message (the current msg if it's in FIFO mode) and immediately return.

of course, the previous state wasn't providing any backpressure either, since there was a no-backpressure `diem_channel` feeding into a backpressured `channel`

```
              diem_channel                          channel 
PeerManager <--------------> NetworkProvider task <---------> Peer task
```

much of this diff is removing the intermediate `channel`s

```
              diem_channel
PeerManager <--------------> Peer task
```

network/src/peer/mod.rs
phlip9: definitely. I wonder if we could take it even further and enforce a stricter separation between the outer event loop and the inner state machine. Maybe this is too much boilerplate, but you could imagine an inner state machine that only exposes like `handle_{event}` methods and then the outer event loop that just looks like

```rust
futures::select! {
  event = async_event_source => self.inner.handle_<...>(event),
  // ..
}
```

it would probably make it easier to test the inner state machine without forcing an implicit dependency on the order that the event loop processes its events.

would also make it easier to fuzz the state machine; you could sample a list of events and feed them into the state machine directly, then make some assertions about the end state.

the big caveat here is not testing a single state machine in isolation, but it doesn't really support testing many interacting state machines. we would need to mock out the event "transport" between state machines more effectively somehow.

network/src/protocols/rpc/mod.rs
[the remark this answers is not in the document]
gregnazario: Okay, that makes more sense to me now.  We'll have to visit later if we want backpressure again or not, but since this didn't do that really anyways it didn't make a difference :P

network/src/peer/mod.rs
gregnazario: Let's keep it separate scope of this PR, but its definitely something I want to tackle on a one off day.  Just to at least clean it up so it's always of that form.  Especially since the formatter really doesn't like dealing with whatever is inside the `futures::select!` since it's a macro.

gregnazario (approved):

PR #7234, opened 2021-01-14
[vm] avoid cyclic dependencies that may be introduced by module republishing

mengxu-fb:
## Motivation

#7143 allows the VM to republish modules. But it has a loose end: cyclic module dependencies can be introduced by updating a module.

This is showcased in the the `republish_cyclic` test case in the Move CLI.

If we first publish two new modules N and M,
```
address 0x43 {
module N {
  public fun bar() {}
}
}
```

```
address 0x42 {
module M {
  use 0x43::N;
  public fun foo() {
    N::bar()
  }
}
}
```

And subsequently update N to be
```
address 0x43 {
module N {
  use 0x42::M;
  public fun bar() {
    M::foo()
  }
}
}
```

Before this PR, the VM will allow the republishing as the new module N is fully compatible (i.e., linking and data layout) with the old module N.

But now N and M make a cyclic dependency, which is a property we would like to avoid.

This PR adds a check for cyclic dependencies. The check is added in the loader entry point, `verify_module_verify_no_missing_dependencies`, that is designated to be invoked upon module publishing. Details on why the check works are explained in the code comments.

Yes

## Test Plan

The `republish_cyclic` should fail at the republish step with error `CYCLIC_MODULE_DEPENDENCY`.

Another test case, `republish_cyclic_transitive`, brings a longer dependency chain. The dependency changes from `A->B->C->D` into `A->B->C->{A,D}`. Having a longer chain has no effect on the checking algorithm as of now because the loader will load *all* dependent modules into cache for verification purpose. But just in case in the future, if the loader is changed to loading *immediate* dependencies for verifications only, this test case will fail and give an early warning on that.

## Related PRs

#7143


language/vm/src/compatibility.rs
```diff
@@ -98,22 +98,4 @@ impl Compatibility {
             struct_layout,
         }
     }
-
-    /// Return true if `new_module` can safely update `old_module`
-    pub fn can_update(
```
mengxu-fb, line 103: The reason we can remove this function (which is not used anywhere) is that its purpose can be served by the `Self:: is_fully_compatible()` function, which addresses point (3), and the updated version of `verify_module_verify_no_missing_dependencies()` in the loader (which addresses point (1) and (2) that are marked as TODO),

language/move-vm/runtime/src/loader.rs
```diff
@@ -590,7 +590,41 @@ impl Loader {
         data_store: &mut impl DataStore,
         log_context: &impl LogContext,
     ) -> VMResult<()> {
-        self.verify_module(module, data_store, true, log_context)
+        // A standard DFS algorithm that follows the dependency chain from the `source_module_id`
```
tnowacki, line 593: I don't think this check should be put here, let me take a look

tnowacki: Yeah, I would put all of this logic in a new file, and invoke it in `verify_module_dependencies`. SImilar to the style where we have a lot of small separate checks in `verify_module`

mengxu-fb: Will do!

mengxu-fb: After a deeper look at the code I realized that there is a difference between putting a check at the `verify_module_verify_no_missing_dependencies` vs `verify_module_dependencies` or `verify_module`:

1) It happens that `verify_module_verify_no_missing_dependencies` is the entry point for module publishing (and it is only called by the VM for module publishing), hence, having the cyclic dependencies check there only takes info effect when a new module is published.

2) However, putting the cyclic dependencies check in `verify_module_dependencies` or `verify_module` means that whenever the module is first loaded into the code cache, the cyclic dependencies check will be performed. And in the case of `A deps on B deps on C deps on D`, the check will run on D, then C -> D, then B -> C -> D, and then A -> B -> C -> D.

I am not so sure which way is better. To me option 2 feels safer although it means a bit longer time for code cache loading.

Also, I would prefer to invoke the check in `verify_module` instead of `verify_module_dependencies` as the latter only checks for *immediate* dependencies of a module while what we need to look at is all dependencies.

tnowacki: I wrote this up already somewhere but it got squashed. So sorry if you get this twice. 

This code needs to be changed to be brought inline with the style of all of the other various checkers. Which is (2) but with some other changes
1. Make a new file in bytcode-verifier/src. Name it something like cyclic_dependencies.rs
  * alternatively you could add it as a pass in `DependencyChecker::verify`
2. Move all the code into that new module. The newer checkers that I wrote (or rewrote) have all of the entry points as a `verify` function in the module, e.g. `type_safety::verify`. Other older checks have it as an entry point in a type, e.g. `DependencyChecker::verify`
3. Invoke the `verify` function after `DependencyChecker::verify` in `verify_module`

language/move-vm/runtime/src/loader.rs
```diff
@@ -681,6 +686,37 @@ impl Loader {
...
 
+    // We do not allow cyclic dependencies between modules, i.e., module A depends on module B
+    // while at the same time, B (transitively) depends on A.
+    fn verify_no_cyclic_module_dependencies(&self, module: &CompiledModule) -> VMResult<()> {
+        // A standard DFS algorithm that follows the dependency chain from the `source_module_id`
+        // to see whether the `dst_module_id` is in its dependencies. This function can only be
+        // called when the module is fully verified and the `module_cache` is properly prepared.
+        fn is_reachable_via_dependencies(
+            dst_module_id: &ModuleId,
+            src_module_id: &ModuleId,
```
tnowacki, line 697: nit, I might name this `cur_module_id`

language/move-vm/runtime/src/loader.rs
```diff
@@ -681,6 +686,37 @@ impl Loader {
...
+            loader: &Loader,
+        ) -> bool {
+            if dst_module_id == src_module_id {
+                return true;
+            }
+            let src_module = loader.get_module(src_module_id);
+            module_dependencies(src_module.module())
+                .iter()
+                .any(|dep| is_reachable_via_dependencies(dst_module_id, dep, loader))
```
tnowacki, line 706: I would then maybe name these as next, so `|next| ...`

Just trying to make the DFS more apparent, but it's up to you

language/move-vm/runtime/src/loader.rs
```diff
@@ -681,6 +686,37 @@ impl Loader {
...
     }
 
+    // We do not allow cyclic dependencies between modules, i.e., module A depends on module B
+    // while at the same time, B (transitively) depends on A.
+    fn verify_no_cyclic_module_dependencies(&self, module: &CompiledModule) -> VMResult<()> {
+        // A standard DFS algorithm that follows the dependency chain from the `source_module_id`
+        // to see whether the `dst_module_id` is in its dependencies. This function can only be
+        // called when the module is fully verified and the `module_cache` is properly prepared.
+        fn is_reachable_via_dependencies(
+            dst_module_id: &ModuleId,
```
tnowacki, line 696: maybe `target_module_id` ?

Or something to indicate it is the module we are verifying? `dst_module_id` feels weird to me

language/move-vm/runtime/src/loader.rs
```diff
@@ -681,6 +686,37 @@ impl Loader {
...
+            src_module_id: &ModuleId,
+            loader: &Loader,
+        ) -> bool {
+            if dst_module_id == src_module_id {
+                return true;
+            }
+            let src_module = loader.get_module(src_module_id);
+            module_dependencies(src_module.module())
```
tnowacki, line 704: Probably way worse, but something to consider 
```
dst_module_id == src_module_id  || module_dependencies(loader.get_module(src_module_id).module()).iter.any(...)
```

language/move-vm/runtime/src/loader.rs
mengxu-fb: Thank you Todd, I thought about what you suggested before and one thing that hinders me from taking that approach is the need for the `loader`, or to be more precisely, a map from `module_id` to `module`, and the `module_dependencies` function. I also tried to piggyback on `DependencyChecker` but later realized that it only captures the immediate dependencies.

Of course these are not unsolvable problems so I am happy to refactor it to make it inline with the style of other checkers :)

tnowacki: I see what you are saying now about the loader. I'd still like to try to refactor it if possible. 

Perhaps we can just have the function take something like `F: FnMut(ModuleID) -> &CompiledModule` or something similar. And this function can then be populated by the loader. Thoughts?

mengxu-fb: I am not sure about the difficulty of having loader to provide a `F: FnMut(ModuleID) -> &CompiledModule`, but one thing we can do to avoid the problem is to pre-calculate `all_deps: BTreeMap<ModuleId, &CompiledModule>` and pass it to `CyclicDependencyChecker::verify(module, &all_deps)`. Then we can do our DFS there.

language/tools/move-cli/tests/testsuite/republish_cyclic/args.txt
```diff
@@ -0,0 +1,6 @@
+publish src1 -v
+view storage/0x00000000000000000000000000000042/modules/M.mv
+view storage/0x00000000000000000000000000000043/modules/N.mv
+publish src2 -v
```
sblackshear, line 4: Not needed for this PR, but in a follow-up would be great to:
- add cyclic dep checks in `move doctor`
- show that after `publish src2` fails`, `publish src2 --ignore-breaking-changes` succeeds + running `move doctor` after detects the problem.

mengxu-fb: I'll add a commit to this PR for this (otherwise I might forgot to create a new PR...)

language/move-vm/runtime/src/loader.rs
mengxu-fb: Done, relocated the check over to `bytecode-verifier`, with the help of `F: Fn(&ModuleID) -> Vec<ModuleId>` that can be supplied by the `loader` (and later, in the `move doctor` case, by sth from the `OnDiskStateView`).

language/bytecode-verifier/src/dependencies.rs
```diff
@@ -338,3 +338,46 @@ impl<'a> DependencyChecker<'a> {
...
+impl CyclicModuleDependencyChecker {
+    pub fn verify_module<F>(module: &CompiledModule, dep_loader: F) -> VMResult<()>
+    where
+        F: Fn(&ModuleId) -> Vec<ModuleId>,
+    {
+        Self::verify_module_impl(module, dep_loader)
+            .map_err(|e| e.finish(Location::Module(module.self_id())))
+    }
+
+    pub fn verify_module_impl<F>(module: &CompiledModule, dep_loader: F) -> PartialVMResult<()>
```
tnowacki, line 353: Why is this pub?

language/move-vm/runtime/src/loader.rs
```diff
@@ -681,6 +686,16 @@ impl Loader {
...
+    // We do not allow cyclic dependencies between modules, i.e., module A depends on module B
+    // while at the same time, B (transitively) depends on A.
+    fn verify_no_cyclic_module_dependencies(&self, module: &CompiledModule) -> VMResult<()> {
+        CyclicModuleDependencyChecker::verify_module(module, |module_id| {
+            self.get_module(module_id)
+                .module()
+                .immediate_module_dependencies()
+        })
+    }
+
```
tnowacki, line 698: I'd remove this function. It should really be a part of `verify_module_dependencies`

language/bytecode-verifier/src/dependencies.rs
mengxu-fb: Followed-suit..., it should not be...

language/bytecode-verifier/src/dependencies.rs
```diff
@@ -338,3 +338,46 @@ impl<'a> DependencyChecker<'a> {
...
+    where
+        F: Fn(&ModuleId) -> Vec<ModuleId>,
+    {
+        // A standard DFS algorithm that follows the dependency chain from the `cursor_module_id`
+        // to see whether the `target_module_id` is in its dependencies.
+        fn is_reachable_via_dependencies<F>(
+            target_module_id: &ModuleId,
+            cursor_module_id: &ModuleId,
+            dep_loader: &F,
```
tnowacki, line 362: I think you can replace  `&F` with `F`. 

&dep_loader should implement the desired trait here

language/bytecode-verifier/src/dependencies.rs
```diff
@@ -338,3 +338,46 @@ impl<'a> DependencyChecker<'a> {
...
+    where
+        F: Fn(&ModuleId) -> Vec<ModuleId>,
+    {
+        Self::verify_module_impl(module, dep_loader)
+            .map_err(|e| e.finish(Location::Module(module.self_id())))
+    }
+
+    pub fn verify_module_impl<F>(module: &CompiledModule, dep_loader: F) -> PartialVMResult<()>
+    where
+        F: Fn(&ModuleId) -> Vec<ModuleId>,
```
tnowacki, line 355: I usually like putting these inline, but its up to you

language/bytecode-verifier/src/dependencies.rs
mengxu-fb: I thought so, but the compiler does not seem to be happy with it:
`error[E0275]: overflow evaluating the requirement [closure@language/move-vm/runtime/src/loader.rs:649:62: 653:10]: Fn<(&ModuleId,)>`

language/tools/move-cli/tests/testsuite/republish_cyclic/args.txt
mengxu-fb: Updated

language/bytecode-verifier/src/dependencies.rs
```diff
@@ -338,3 +338,46 @@ impl<'a> DependencyChecker<'a> {
         }
     }
 }
+
+pub struct CyclicModuleDependencyChecker {}
+
+impl CyclicModuleDependencyChecker {
+    pub fn verify_module<F>(module: &CompiledModule, dep_loader: F) -> VMResult<()>
```
tnowacki, line 345: Naming this thing "dep_loader" feels a bit leaky as this bytecode verifier knows nothing about the loader. 

Perhaps a more declarative name, something like "module_dependencies"?

mengxu-fb: How about, `immediate_module_dependencies`?

tnowacki: Works for me!

language/bytecode-verifier/src/dependencies.rs
tnowacki: Thats a new one...

language/bytecode-verifier/src/dependencies.rs
```diff
@@ -338,3 +338,46 @@ impl<'a> DependencyChecker<'a> {
...
+        fn is_reachable_via_dependencies<F>(
+            target_module_id: &ModuleId,
+            cursor_module_id: &ModuleId,
+            dep_loader: &F,
+        ) -> bool
+        where
+            F: Fn(&ModuleId) -> Vec<ModuleId>,
+        {
+            target_module_id == cursor_module_id
+                || (dep_loader)(cursor_module_id)
```
tnowacki, line 368: Do you need the parens still?

mengxu-fb: Technically no, it's more like a personal preference to indicate that this is a function pointer...

tnowacki: They aren't function pointers though... at least not really. I mean at some point sure they are, but as far as Rust is concerned you have some value that implements the trait `Fn(&ModuleId) -> Vec<ModuleId>`

I think the consistent style with the rest of the codebase would be to not include them

language/vm/src/access.rs
```diff
@@ -183,6 +183,20 @@ pub trait ModuleAccess: Sync {
...
+        let self_handle = self.self_handle();
+        self.module_handles()
+            .iter()
+            .filter_map(|handle| {
+                if handle == self_handle {
+                    None
+                } else {
+                    Some(self.module_id_for_handle(handle))
+                }
+            })
```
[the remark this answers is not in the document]
mengxu-fb: Absolutely right on the reading! Just a question, does it mean that .filter() and .map() require two traversals of the list while filter_map() requires only one? I am not sure how Rust compiler will optimize this.

language/tools/move-cli/tests/testsuite/republish_cyclic_transitive/args.exp
```diff
@@ -0,0 +1,16 @@
...
+Compiling Move modules...
+Found and compiled 4 modules
+Publishing a new module 00000000000000000000000000000042::A
+Publishing a new module 00000000000000000000000000000042::B
+Publishing a new module 00000000000000000000000000000042::C
+Publishing a new module 00000000000000000000000000000042::D
+Command `publish src2 -v`:
+Compiling Move modules...
+Found and compiled 1 modules
+Publishing failed with unexpected error CYCLIC_MODULE_DEPENDENCY
```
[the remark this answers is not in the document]
mengxu-fb: Yes, I am a bit surprised too. I guess sth is wrong with the `shadow` logic? But I haven't check it.

language/bytecode-verifier/src/dependencies.rs
```diff
@@ -338,3 +338,52 @@ impl<'a> DependencyChecker<'a> {
...
+        module: &CompiledModule,
+        immediate_module_dependencies_getter: F,
+    ) -> VMResult<()> {
+        Self::verify_module_impl(module, immediate_module_dependencies_getter)
+            .map_err(|e| e.finish(Location::Module(module.self_id())))
+    }
+
+    fn verify_module_impl<F: Fn(&ModuleId) -> Vec<ModuleId>>(
+        module: &CompiledModule,
+        immediate_module_dependencies_getter: F,
```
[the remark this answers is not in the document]
mengxu-fb: No worries, I share your point, yes now I recall that the `_getter` is almost a unique OOP thing...

language/vm/src/access.rs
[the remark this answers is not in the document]
tnowacki: AFAIK the iterator calls are "lazy" and should be deforested (hence the need to collect at the end)

I doubt you would notice a perf difference except in the most extreme cases

language/bytecode-verifier/src/dependencies.rs
```diff
@@ -338,3 +338,48 @@ impl<'a> DependencyChecker<'a> {
...
+            .map_err(|e| e.finish(Location::Module(module.self_id())))
+    }
+
+    fn verify_module_impl<F: Fn(&ModuleId) -> Vec<ModuleId>>(
+        module: &CompiledModule,
+        immediate_module_dependencies: F,
+    ) -> PartialVMResult<()> {
+        // A standard DFS algorithm that follows the dependency chain from the `cursor_module_id`
+        // to see whether the `target_module_id` is in its dependencies.
+        fn is_reachable_via_dependencies<F: Fn(&ModuleId) -> Vec<ModuleId>>(
```
shazqadeer, line 359: I don't see the visited set optimization being used (or am I mistaken?).  You could get exponential behavior without this optimization.

language/move-vm/runtime/src/runtime.rs
```diff
@@ -101,7 +101,7 @@ impl VMRuntime {
             }
         }
 
-        // perform bytecode and loading verification
+        // perform bytecode, loading verification, and checking of cyclic dependencies
```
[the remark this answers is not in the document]
mengxu-fb: Right, I think I should probably leave the comment unchanged. Now checking cyclic dependencies is part of the loading verification.

language/bytecode-verifier/src/dependencies.rs
mengxu-fb: Right, I see. I'll give it a tweak.

shazqadeer:
Few questions:

1. Is it true that the module cache is closed under the use relation?  That is, if a module M is present in the cache so will any module N used by M? If not, it must be the case that modules are loaded on demand when a called function is needed.  Could it happen that the cyclic dependency checker may ask to see the dependencies of a module that is not in the cache?  What happens in such a case?
2. The cyclic dependency checker may do work (in the worst case) proportional to the total number of modules in the Diem blockchain?  Are we comfortable with that?

mengxu-fb:
To answer Shaz's questions about code cache: Yes, the module cache is closed under the use relation --- If a module M is present in the cache, any module that M depends on, directly or transitively, will be materialized in the cache.

language/tools/move-cli/tests/testsuite/republish_cyclic_transitive/args.exp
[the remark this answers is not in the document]
tnowacki: I think I found the issue, and not really sure it is truly fixable. 

The source language does not introspect the bodies of functions for dependency checking (since for perf it doesn't want to look at their function bodies). So the source language doesn't see the circularity
(in this case it would be even harder since it is using the generated "interface" file for 0x42::M)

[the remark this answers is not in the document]
mengxu-fb: Ah, right, in the interface files, `use 0x42::N` is stripped, which is there in M's source code.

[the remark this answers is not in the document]
tnowacki: I don't think uses are ever considered for dependencies (if they are its just for error message purposes).

Conceptually, uses are just aliases. 

The real issue here is that the interface files (1) don't have function bodies and (2) even if they did they are marked as dep files and the function bodies would be ignored for perf reasons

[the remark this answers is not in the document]
mengxu-fb: Sorry, I misunderstood `use`. Yes, those are the real issues.

But does this mean that the compiler may produce bytecode that may not pass the bytecode verifier? Well, it might be wrong to say that the bytecode *does not pass* the bytecode verifier because the per-module checkings are all fine, and the linkings are fine too. It is only this cyclic dep that causes some trouble.

sblackshear:
2. The cyclic dependency checker may do work (in the worst case) proportional to the total number of modules in the Diem blockchain? Are we comfortable with that?

For now, we are comfortable with this because of the answer to (1): the loader already downloads every module in the blockchain in the worst case.

In the future, I imagine we will have to limit the amount of code loaded eagerly and bound the exploration of the dependency checker. But for now, the framework is so small that this isn't an issue.

tnowacki:
I think we should be okay with all the perf concerns here. But it is an interesting longer term issue. I think we will just have to appropriately cost the "republish" functionality. 

Another thought: if we are really concerned about perf, we could turn this check off for normal publishing.

mengxu-fb:
I think appropriately costing the "republish" function is the right thing to consider even without this check.

Consider the case when the loader cache is cold (e.g., publishing a module in a fresh VM instance): given the fact that the loader needs to download every module in the blockchain in the worst case, which is DFS by itself, the perf penalty stays almost the same with or without this check --- cycle detection is way lightweight than deserializing all the dependency modules.

language/tools/move-cli/tests/testsuite/republish_cyclic_transitive/args.exp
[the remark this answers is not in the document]
tnowacki: The compiler will always produce code that might not pass the being able to being published on chain. In other words, the compiler has no on chain view, so it could always produce code that is inconsistent or wrong with what is in a given "chain"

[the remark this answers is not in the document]
mengxu-fb: Right! I think what I was trying to say is:
1) for module compilation we pass it the module source + dependencies for compilation, and
2) even the dependencies are exactly the same as the ones on-chain,
3) the compiler may output the module bytecode that will be rejected by the bytecode verifier.

I am not sure whether this may happen before republishing is allowed.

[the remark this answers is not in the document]
tnowacki: It sort of could have happened before.

I think it's fine for now and there are ways of dealing with it in the future if it is more of an issue. Its quite the edge case and if you are really concerned about it, you can make all of your files "source files", which might be the common case anyway

mengxu-fb:
@shazqadeer @tnowacki @sblackshear I think I have addressed the comments but do let me know if there are more changes I need to implement.

language/bytecode-verifier/src/dependencies.rs
```diff
@@ -338,3 +338,65 @@ impl<'a> DependencyChecker<'a> {
...
+    ) -> VMResult<()> {
+        Self::verify_module_impl(module, immediate_module_dependencies)
+            .map_err(|e| e.finish(Location::Module(module.self_id())))
+    }
+
+    fn verify_module_impl<F: Fn(&ModuleId) -> Vec<ModuleId>>(
+        module: &CompiledModule,
+        immediate_module_dependencies: F,
+    ) -> PartialVMResult<()> {
+        // A variant of the three-color DFS, visit each node at max once
```
sblackshear, line 357: Say what return value means?

language/tools/move-cli/src/main.rs
```diff
@@ -608,6 +609,52 @@ fn explain_publish_error(
                 println!("Linking API for structs/functions of module {} has changed. Need to redeploy all dependent modules.", module_id)
             }
         }
+        VMStatus::Error(CYCLIC_MODULE_DEPENDENCY) => {
+            println!(
+                "Publishing module {} introduces cyclic dependencies.",
+                module_id
+            );
+            // find all cycles with an iterative DFS
```
sblackshear, line 617: Might make sense to extract this useful logic to a `print_cycle` helper function + use it below in `move doctor`? That way, you can do the cyclic dep check once in `doctor` instead of for each module.

sblackshear (approved):
LGTM, thanks for iterating on this so much!

language/tools/move-cli/src/main.rs
mengxu-fb: Yes, this will be helpful. I guess the reason I did not do so is that it kinds of breaks the mental model --- `move doctor` calls exactly the same `bytecode_verifier::*::verify_module()` checks consecutively as the VM would do, and I kind of want to preserve the property of keeping checks in `move doctor` and the VM side in-sync.

shazqadeer (approved):

language/tools/move-cli/src/main.rs
sblackshear: Fair point! It's not like it's important for `move doctor` to be fast anyway... I like your way better.

PR #7239, opened 2021-01-14
minimum gas should be scaled

guangyuz:
## Motivation

(Write your answer here.)

## Test Plan

## Related PRs

### DIFF
--- a/language/benchmarks/src/move_vm.rs
+++ b/language/benchmarks/src/move_vm.rs
@@ -57,7 +57,7 @@ fn execute(
     let data_cache = StateViewCache::new(&state);
     let log_context = NoContextLog::new();
     let mut session = move_vm.new_session(&data_cache);
-    let mut cost_strategy = CostStrategy::system(&gas_schedule, GasUnits::new(100_000_000));
+    let mut cost_strategy = CostStrategy::system(&gas_schedule, ScaledGasUnits::new(100_000_000));
 
     let mut mod_blob = vec![];
     module
--- a/language/diem-tools/transaction-replay/src/lib.rs
+++ b/language/diem-tools/transaction-replay/src/lib.rs
@@ -15,7 +15,7 @@ use diem_validator_interface::{
 };
 use diem_vm::{data_cache::RemoteStorage, txn_effects_to_writeset_and_events, DiemVM, VMExecutor};
 use move_cli::OnDiskStateView;
-use move_core_types::gas_schedule::{GasAlgebra, GasUnits};
+use move_core_types::gas_schedule::{GasAlgebra, ScaledGasUnits};
 use move_lang::{compiled_unit::CompiledUnit, move_compile, shared::Address};
 use move_vm_runtime::{logging::NoContextLog, move_vm::MoveVM, session::Session};
 use move_vm_test_utils::{ChangeSet as MoveChanges, DeltaStorage};
@@ -301,7 +301,7 @@ impl DiemDebugger {
         let gas_table = zero_cost_schedule();
         let is_version_ok = |version| {
             self.run_session_at_version(version, override_changeset.clone(), |session| {
-                let mut cost_strategy = CostStrategy::system(&gas_table, GasUnits::new(0));
+                let mut cost_strategy = CostStrategy::system(&gas_table, ScaledGasUnits::new(0));
                 let log_context = NoContextLog::new();
                 session.execute_script(
                     predicate.clone(),
--- a/language/diem-tools/writeset-transaction-generator/src/writeset_builder.rs
+++ b/language/diem-tools/writeset-transaction-generator/src/writeset_builder.rs
@@ -11,7 +11,7 @@ use diem_types::{
 };
 use diem_vm::{data_cache::RemoteStorage, txn_effects_to_writeset_and_events};
 use move_core_types::{
-    gas_schedule::{CostTable, GasAlgebra, GasUnits},
+    gas_schedule::{CostTable, GasAlgebra, ScaledGasUnits},
     identifier::Identifier,
     language_storage::{ModuleId, TypeTag},
 };
@@ -49,7 +49,7 @@ impl<'r, 'l, R: RemoteCache> GenesisSession<'r, 'l, R> {
                 ty_args,
                 args,
                 sender,
-                &mut CostStrategy::system(&ZERO_COST_SCHEDULE, GasUnits::new(100_000_000)),
+                &mut CostStrategy::system(&ZERO_COST_SCHEDULE, ScaledGasUnits::new(100_000_000)),
                 &NoContextLog::new(),
             )
             .unwrap_or_else(|e| {
@@ -69,7 +69,7 @@ impl<'r, 'l, R: RemoteCache> GenesisSession<'r, 'l, R> {
                 script.ty_args().to_vec(),
                 convert_txn_args(script.args()),
                 vec![sender],
-                &mut CostStrategy::system(&ZERO_COST_SCHEDULE, GasUnits::new(100_000_000)),
+                &mut CostStrategy::system(&ZERO_COST_SCHEDULE, ScaledGasUnits::new(100_000_000)),
                 &NoContextLog::new(),
             )
             .unwrap()
--- a/language/diem-vm/src/diem_transaction_executor.rs
+++ b/language/diem-vm/src/diem_transaction_executor.rs
@@ -31,7 +31,7 @@ use diem_types::{
 use fail::fail_point;
 use move_core_types::{
     account_address::AccountAddress,
-    gas_schedule::{CostTable, GasAlgebra, GasCarrier, GasUnits},
+    gas_schedule::{CostTable, GasAlgebra, GasCarrier, ScaledGasUnits},
     identifier::IdentStr,
 };
 use move_vm_runtime::{data_cache::RemoteCache, logging::LogContext, session::Session};
@@ -62,7 +62,7 @@ impl DiemVM {
         &self,
         error_code: VMStatus,
         gas_schedule: &CostTable,
-        gas_left: GasUnits<GasCarrier>,
+        gas_left: ScaledGasUnits<GasCarrier>,
         txn_data: &TransactionMetadata,
         remote_cache: &StateViewCache<'_>,
         account_currency_symbol: &IdentStr,
@@ -84,7 +84,7 @@ impl DiemVM {
         &self,
         error_code: VMStatus,
         gas_schedule: &CostTable,
-        gas_left: GasUnits<GasCarrier>,
+        gas_left: ScaledGasUnits<GasCarrier>,
         txn_data: &TransactionMetadata,
         remote_cache: &StateViewCache<'_>,
         account_currency_symbol: &IdentStr,
@@ -125,7 +125,7 @@ impl DiemVM {
         &self,
         mut session: Session<R>,
         gas_schedule: &CostTable,
-        gas_left: GasUnits<GasCarrier>,
+        gas_left: ScaledGasUnits<GasCarrier>,
         txn_data: &TransactionMetadata,
         account_currency_symbol: &IdentStr,
         log_context: &impl LogContext,
@@ -337,7 +337,7 @@ impl DiemVM {
         log_context: &impl LogContext,
     ) -> Result<ChangeSet, Result<(VMStatus, TransactionOutput), VMStatus>> {
         let gas_schedule = zero_cost_schedule();
-        let mut cost_strategy = CostStrategy::system(&gas_schedule, GasUnits::new(0));
+        let mut cost_strategy = CostStrategy::system(&gas_schedule, ScaledGasUnits::new(0));
 
         Ok(match writeset_payload {
             WriteSetPayload::Direct(change_set) => change_set.clone(),
@@ -424,7 +424,7 @@ impl DiemVM {
         txn_data.sender = account_config::reserved_vm_address();
 
         let gas_schedule = zero_cost_schedule();
-        let mut cost_strategy = CostStrategy::system(&gas_schedule, GasUnits::new(0));
+        let mut cost_strategy = CostStrategy::system(&gas_schedule, ScaledGasUnits::new(0));
         let mut session = self.0.new_session(remote_cache);
 
         let (round, timestamp, previous_vote, proposer) = block_metadata.into_inner();
--- a/language/diem-vm/src/diem_transaction_validator.rs
+++ b/language/diem-vm/src/diem_transaction_validator.rs
@@ -18,7 +18,7 @@ use diem_types::{
     vm_status::{StatusCode, VMStatus},
 };
 use move_core_types::{
-    gas_schedule::{GasAlgebra, GasUnits},
+    gas_schedule::{GasAlgebra, ScaledGasUnits},
     identifier::{IdentStr, Identifier},
     move_resource::MoveResource,
 };
@@ -148,7 +148,7 @@ pub(crate) fn validate_signature_checked_transaction<R: RemoteCache>(
     let txn_data = TransactionMetadata::new(transaction);
     let log_context = AdapterLogSchema::new(remote_cache.id(), 0);
     let mut cost_strategy =
-        CostStrategy::system(vm.get_gas_schedule(&log_context)?, GasUnits::new(0));
+        CostStrategy::system(vm.get_gas_schedule(&log_context)?, ScaledGasUnits::new(0));
     let prologue_status = match transaction.payload() {
         TransactionPayload::Script(_script) => {
             vm.check_gas(&txn_data, &log_context)?;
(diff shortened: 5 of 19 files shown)


guangyuz:
https://github.com/diem/diem/issues/7238

tzakian:
Hi @guangyuz, thanks for reporting this and for the PR! 

You bring up a really good point that we should probably decrease this to be more in-line with the average gas consumption. I don't think we want to necessarily scale the minimum gas amount using the `gas_unit_scaling` factor however; depending on the fluctuations of the gas scaling factor, and the `min_transaction_gas_units` this could lead to a minimum amount of zero being required for a non-zero gas price (which we definitely don't want). 

It would probably be better to update the `min_transaction_gas_units` in both the [DiemVMConfig](https://github.com/diem/diem/blob/master/language/stdlib/modules/DiemVMConfig.move#L85) and in the [Rust code](https://github.com/diem/diem/blob/master/language/move-core/types/src/gas_schedule.rs#L203) with a lower value (100) to be more in-line with the average gas consumption of a transaction.

guangyuz:
> 
> It would probably be better to update the `min_transaction_gas_units` in both the [DiemVMConfig](https://github.com/diem/diem/blob/master/language/stdlib/modules/DiemVMConfig.move#L85) and in the [Rust code](https://github.com/diem/diem/blob/master/language/move-core/types/src/gas_schedule.rs#L203) with a lower value (100) to be more in-line with the average gas consumption of a transaction.

Hi @tzakian,

I have some doubts, not sure if the correct value of "min_transaction_gas_units" is 100. I did some experiment to understand the logic behind.

I add log at [charge_intrinsic_gas](https://github.com/diem/diem/blob/master/language/diem-vm/src/diem_transaction_executor.rs#L175) and [charge_global_write_gas_usage](https://github.com/diem/diem/blob/master/language/diem-vm/src/diem_transaction_executor.rs#L188) to measure the gas used in each step. Then ran two stdlib scripts: add_currency_to_account and peer_to_peer_with_metadata and collected the following data:

add_currency_to_account:
default initial gas: GasUnits(4000000)
initial gas after scaling up: GasUnits(4000000000)
...step1.charge_intrinsic_gas: GasUnits(600)
...step2.charge_instr: GasUnits(89763)
...step3.charge_global_write_gas_usage: GasUnits(7200)
gas left: GasUnits(3999902437)
gas used(after scaling down): GasUnits(97)

peer_to_peer_with_metadata:
default initial gas: GasUnits(4000000)
initial gas after scaling up: GasUnits(4000000000)
...step1.charge_intrinsic_gas: GasUnits(600)
...step2.charge_instr: GasUnits(470541)
...step3.charge_global_write_gas_usage: GasUnits(14400)
gas left: GasUnits(3999514459)
gas used(after scaling down): GasUnits(485)

So, actually the intrinsic cost of the transactions is always less than 1 (after scaling down). If, like I did in the patch, we only guarantee that the submitted transactions max gas units covers the intrinsic cost of the transaction, then this will lead to a strange situation as you mentioned - minimum amount of zero being required for a non-zero gas price. But set "min_transaction_gas_units" to 100 is also somewhat inappropriate. We still can't run a script like add_currency_to_account although we have enough gas (97).

I can’t think of a better solution, maybe we should increase this "min_transaction_gas_units" value to greater than 1000？

guangyuz:
Another idea is to scaling up max_gas_amount instead of scaling down min_txn_fee to avoid minimum amount of zero being required. See the new patch, this time the minimum amount 1 is required. ;-)

bob-wilson:
@tzakian I disagree. It seems to me like we have 2 different kinds of gas measurements: scaled and unscaled. (It would be a good idea to introduce separate types to clarify when each should be used!) The VM operates on the scaled gas (* 1000) and the external world sees the unscaled gas (/ 1000). In this code in the adapter, we're comparing the unscaled max_gas_amount from the txn with the scaled result of calculate_intrinsic_gas(). That doesn't make sense to me, and I think @guangyuz had the right idea for fixing it in this PR. If we need to adjust gas constants to make that work, then so be it, but that part isn't clear to me.

tzakian:
After the discussion above, I agree that we want to scale the min/max gas constants.

I think the updated methodology of multiplying by the scaling factor as opposed to dividing by it makes sense to me, and I would be happy with this type of change, as long as it's applied consistently for both min and max gas constants. 

This then leaves the question of what the `min_gas_units` and `max_gas_units` should be set at. I will take a look at the gas consumption across the scripts and reply back here on what seems to be best/data from that so we can determine a new `min_gas_units`. For the `max_gas_units`,  I think we should just scale the current set `max_gas_units` by the scaling factor amount.

bob-wilson:
I'm getting confused. I really think we should introduce separate types for scaled vs. unscaled gas units, because I'm having trouble keeping them straight. I hadn't looked at `max_gas_units` before. Unlike `min_transaction_gas_units`, `max_gas_units` is a field in the transaction, not in the gas schedule. Based on my explanation above, this should be using unscaled gas, because it is a value passed in from the outside world. However, that value is getting passed directly through to the VM in calls in execute_function without getting scaled. That seems like a bug to me. @tzakian do you agree?

jolestar:
@bob-wilson  We need to reconsider why scaled gas was introduced? Is it because of the precision of currency is too small? Then why not increase the precision of the currency?

tzakian:
@bob-wilson 
> I'm getting confused. I really think we should introduce separate types for scaled vs. unscaled gas units, because I'm having trouble keeping them straight.

Agreed! Pretty much all of the type-level mechanics are all there already around gas units, so this change shouldn't be too hard. In fact it's probably just doing a bit of refactoring to make the boundary and usage of the `GasUnits` type fit scaled vs unscaled. 

>  I hadn't looked at `max_gas_units` before. Unlike `min_transaction_gas_units`, `max_gas_units` is a field in the transaction, not in the gas schedule. Based on my explanation above, this should be using unscaled gas, because it is a value passed in from the outside world. However, that value is getting passed directly through to the VM in calls in execute_function without getting scaled. That seems like a bug to me. @tzakian do you agree?

As it stands right now, I don't believe so. This is because:
1. All the fields in the user-submitted transaction are unscaled (and I think it should remain that way)
2. When they are being passed to the prologue functions they are also unscaled. This data is used to compute the max transaction fee liability for the transaction for balance checking purposes.
3. In the epilogues we pass in the (unscaled) remaining gas. See the implementation of `remaining_gas` for this. This unscaled number is used to compute the final transaction fee.

So in all of those cases that I saw in the diem_vm, we do want to be dealing with the unscaled amount of gas units.

Also I think there was confusion between  scaling the `max_gas_units` in the transaction, and the `maximum_number_of_gas_units` field in the gas constants struct, I meant the latter. I think the transaction gas should remain unscaled. 


@jolestar 
> We need to reconsider why scaled gas was introduced? Is it because of the precision of currency is too small? Then why not increase the precision of the currency?

There are a couple reasons we decided on scaled gas vs increasing the currency precision:
1.  If you make the currency precision smaller, you are reducing the largest amount of money that can be represented in that currency. We also probably don't want to have the representation of currency in the system dictated by gas, and instead give that flexibility to others (e.g., economists) so they can decide. 
2. We can change the scaling factor as time goes on as, e.g., the instructions become less expensive computationally, whereas once a given precision is picked for a currency that can't really ever be changed. And in general, the scaling factor gives us a knob to turn to scale the transaction fees in the system as time goes on.

bob-wilson:
> So in all of those cases that I saw in the diem_vm, we do want to be dealing with the unscaled amount of gas units.

Agreed with all of those cases. The potential bug that concerns me is when the transaction's max_gas_amount is passed as an argument to execute_function. I think it needs to be scaled there. See, for example, the code in run_script_prologue, etc.

tzakian:
> > So in all of those cases that I saw in the diem_vm, we do want to be dealing with the unscaled amount of gas units.
> 
> Agreed with all of those cases. The potential bug that concerns me is when the transaction's max_gas_amount is passed as an argument to execute_function. I think it needs to be scaled there. See, for example, the code in run_script_prologue, etc.

For clarity, I think we are both discussing the [same code](https://github.com/diem/diem/blob/master/language/diem-vm/src/diem_vm.rs#L234). Let me know if I'm wrong though. I would say the code in this example is fine. I say this because the transaction fee is calculated based on the _unscaled_ gas usage in the epilogue, therefore we want to compute the maximum transaction fee here based off of the unscaled gas units passed in the transaction -- so we want to pass the unscaled value.

bob-wilson:
Ah, I get it. Sorry for the confusion. I was reading that as a Rust argument to VM, but it is a Move argument to the prologue/epilogue function. I completely agree that this should be unscaled.

guangyuz:
> @bob-wilson
> 
> > I'm getting confused. I really think we should introduce separate types for scaled vs. unscaled gas units, because I'm having trouble keeping them straight.
> 
> Agreed! Pretty much all of the type-level mechanics are all there already around gas units, so this change shouldn't be too hard. In fact it's probably just doing a bit of refactoring to make the boundary and usage of the `GasUnits` type fit scaled vs unscaled.

The patch has been updated. `ScaledGasUnits` is introduced to describe gas units on the user side, and `GasUnits` is still used in the engine side.

bob-wilson (changes requested):
This is along the general lines of what I was thinking, but after reviewing this, I would be more inclined to keep the "GasUnits" name for the externally visible amounts and use a new "InternalGasUnits" type for the amounts used within the VM. Otherwise, it's not very clear which way the scaling goes: are the "scaled" units the internal or external ones?

It really seems to me like the `maximum_number_of_gas_units` gas constant should be expressed in the "external" units, especially when I see code like the `get_transaction_parameters` function in language/testing-infra/functional-tests/src/evaluator.rs. But, changing that might be more trouble than it's worth, since it is stored on-chain in the gas schedule. @tzakian WDYT? If we don't change it, the new type should help avoid mistakes.

It also looks like the `calculate_intrinsic_gas` function might be better returning "external" units. Besides the comparison that is adjusted in `diem_vm.rs`, see the code using that in `language/testing-infra/e2e-tests/src/account_universe/bad_transaction.rs`.

Finally, the Move Adapter specification (in `specifications/move_adapter/README.md`) should be adjusted if we change the way the max gas amount checked.

bob-wilson:
By the way, @guangyuz, if you'd prefer to hand this off to someone else, feel free to ask. I really appreciate you pointing out this problem regardless. If you don't mind iterating on the PR, that is even better!

tzakian:
> This is along the general lines of what I was thinking, but after reviewing this, I would be more inclined to keep the "GasUnits" name for the externally visible amounts and use a new "InternalGasUnits" type for the amounts used within the VM. Otherwise, it's not very clear which way the scaling goes: are the "scaled" units the internal or external ones?

I agree with keeping the `GasUnits` name as-is. I was thinking that we wouldn't even have a newtype for non-scaled gas units (these would just be `u64`s), that way there's no confusion, and since metering should never been done against anything other than a `GasUnit` type. But I'm open to having a newtype for unscaled gas units as well.


> It really seems to me like the `maximum_number_of_gas_units` gas constant should be expressed in the "external" units, especially when I see code like the `get_transaction_parameters` function in language/testing-infra/functional-tests/src/evaluator.rs. But, changing that might be more trouble than it's worth, since it is stored on-chain in the gas schedule. @tzakian WDYT? If we don't change it, the new type should help avoid mistakes.

I think we should scale this if we are scaling the `min_transaction_gas_units`. I think we could keep this as-is as and just treat it as scaled now, since it's already set quite high and I don't think dividing it by, e.g., 1000 will cause any issues.

To be precise my thoughts are:
* Any representation of "units of gas" in the gas constants stored on-chain should be defined in a scaled form (i.e., both `min_transaction_gas_units` and `maximum_number_of_gas_units`. 
* Any representation of units of gas in the transaction submitted should be unscaled.

The main thing for me is consistency; either the struct constants need to all be scaled, or all be unscaled.

> It also looks like the `calculate_intrinsic_gas` function might be better returning "external" units. Besides the comparison that is adjusted in `diem_vm.rs`, see the code using that in `language/testing-infra/e2e-tests/src/account_universe/bad_transaction.rs`.

At least from a preliminary look this doesn't look to be a problem to me, but I may be missing something..

> By the way, @guangyuz, if you'd prefer to hand this off to someone else, feel free to ask. I really appreciate you pointing out this problem regardless. If you don't mind iterating on the PR, that is even better!

I also want to second this -- and wanted to also thanks as well for bringing this up! Let us know how you want to proceed.

bob-wilson:
@tzakian and I just chatted.

> > This is along the general lines of what I was thinking, but after reviewing this, I would be more inclined to keep the "GasUnits" name for the externally visible amounts and use a new "InternalGasUnits" type for the amounts used within the VM. Otherwise, it's not very clear which way the scaling goes: are the "scaled" units the internal or external ones?
> 
> I agree with keeping the `GasUnits` name as-is. I was thinking that we wouldn't even have a newtype for non-scaled gas units (these would just be `u64`s), that way there's no confusion, and since metering should never been done against anything other than a `GasUnit` type. I'm open to having a newtype for unscaled gas units as well, just I'm not sure what that buys us is worth the extra complexity.

I think I convinced Tim about the need for a newtype. As we were talking, I found it really tough to keep straight between the "scaled" and "unscaled" terminology, so at least for me, "internal" (e.g., InternalGasUnits) and "external" (e.g., GasUnits) works a lot better.

> > It really seems to me like the `maximum_number_of_gas_units` gas constant should be expressed in the "external" units, especially when I see code like the `get_transaction_parameters` function in language/testing-infra/functional-tests/src/evaluator.rs. But, changing that might be more trouble than it's worth, since it is stored on-chain in the gas schedule. @tzakian WDYT? If we don't change it, the new type should help avoid mistakes.
> 
> I think we should scale this if we are scaling the `min_transaction_gas_units`. I think we could keep this as-is as and just treat it as scaled now, since it's already set quite high and I don't think dividing it by, e.g., 1000 will cause any issues.

One thing that Tim and I realized is that changing the on-chain gas schedule or the interpretation of the existing values will be difficult. We can do it if we have to, but it will add a lot of complexity since we need the ability to re-execute old transactions and get identical results -- including for gas fees. One thing we can do without too much complexity is to fix this bug where transactions are being incorrectly discarded during validation, as long as the fix doesn't change the behavior for transactions that were previously treated as valid.

> 
> To be precise my thoughts are:
> 
> * Any representation of "units of gas" in the gas constants stored on-chain should be defined in a scaled form (i.e., both `min_transaction_gas_units` and `maximum_number_of_gas_units`.
> * Any representation of units of gas in the transaction submitted should be unscaled.
> 
> The main thing for me is consistency; either the struct constants need to all be scaled, or all be unscaled.

Nope. See above. What we need is to classify the existing constants as either internal or external units. Here's my take on that:

Internal
- `global_memory_per_byte_cost`
- `global_memory_per_byte_write_cost`
- `min_transaction_gas_units` (note inconsistency where this is used in `e2e-testsuite/src/tests/verify_txn.rs`)
- `intrinsic_gas_per_byte`

External
- `maximum_number_of_gas_units` (i.e., the conversion in the current PR is unnecessary)

> > It also looks like the `calculate_intrinsic_gas` function might be better returning "external" units. Besides the comparison that is adjusted in `diem_vm.rs`, see the code using that in `language/testing-infra/e2e-tests/src/account_universe/bad_transaction.rs`.
> 
> At least from a preliminary look this doesn't look to be a problem to me, but I may be missing something..

Tim pointed out the root problem with the `calculate_intrinsic_gas` function: we're using the result in one place as "external" units and in another place as "internal" units. We should leave the function returning internal units and convert when comparing against the transaction's `max_gas_amount` (as @guangyuz proposed in this PR).

guangyuz:
> @tzakian and I just chatted.
> 
> > > This is along the general lines of what I was thinking, but after reviewing this, I would be more inclined to keep the "GasUnits" name for the externally visible amounts and use a new "InternalGasUnits" type for the amounts used within the VM. Otherwise, it's not very clear which way the scaling goes: are the "scaled" units the internal or external ones?
> > 
> > 
> > I agree with keeping the `GasUnits` name as-is. I was thinking that we wouldn't even have a newtype for non-scaled gas units (these would just be `u64`s), that way there's no confusion, and since metering should never been done against anything other than a `GasUnit` type. I'm open to having a newtype for unscaled gas units as well, just I'm not sure what that buys us is worth the extra complexity.
> 
> I think I convinced Tim about the need for a newtype. As we were talking, I found it really tough to keep straight between the "scaled" and "unscaled" terminology, so at least for me, "internal" (e.g., InternalGasUnits) and "external" (e.g., GasUnits) works a lot better.
> 
I am also in favor of using GasUnits / InternalGasUnits, which allows us not to change the TransactionMetadata type.

guangyuz:
> By the way, @guangyuz, if you'd prefer to hand this off to someone else, feel free to ask. I really appreciate you pointing out this problem regardless. If you don't mind iterating on the PR, that is even better!

This is an incompatible modification. If you are willing to take over, it is undoubtedly the best.

bob-wilson:
> If you are willing to take over, it is undoubtedly the best.

Sure, I'll volunteer, since I already spent a bunch of time digging into it yesterday. Thanks again, @guangyuz !!

PR #7240, opened 2021-01-14
[developers.diem.com] Switch README from the old Libra logo to the Diem logo

bob-wilson:
Looks like we missed this one.... Thanks to @shazqadeer for noticing.

Yes


shazqadeer (approved):

JoelMarcey (approved):

JakeSilverman:
Is this missing a period? Perhaps should be "../.assets/diem.png"

bob-wilson:
> Is this missing a period? Perhaps should be "../.assets/diem.png"

This PR adds a separate copy of the image inside the developers.diem.com directory. I wasn't sure if this directory hierarchy is ever published separately, and thought it would be best to make it independent. If that's not necessary, we could remove the copied image and make that change.

JakeSilverman:
Ah I see. Only saw the modified file. Didn't see the new one. Probably a similar issue that the file "https://github.com/diem/diem/blob/master/diem.png" likely shouldn't exist.

bob-wilson:
It looks like this failed to land because of a flaky test, so we have an opportunity to revise this before it goes in. @JoelMarcey can you give some guidance on whether we need 2 copies of the image? Any reason not to also delete the top-level diem.png image that actually contains the Libra logo? (It appears to be unused.)

JoelMarcey:
I think it can be removed. @oriooctopus @jjwill, agree?

JakeSilverman:
Since discussing diem.png (the libra logo), want to also point out that developers.diem.com/libra.png will also be unused after this PR is applied and so can probably be deleted also.

oriooctopus:
I agree that we can delete the image in the base directory, but I don't think we should create a new folder... we already have a `static/img` folder... can we just use that? If so, please put it in at `static/img/logo.png`

bob-wilson:
> If so, please put it in at `static/img/logo.png`

How about static/img/diem-logo.png? There are already some similar files in that directory (diem-logomark-white.png and diem-header-logo-white.png) so I think it would be good to have the file name start with "diem".

I updated the PR to do that and also remove the unused files that @JakeSilverman pointed out.

oriooctopus:
That sounds good to me!

PR #7242, opened 2021-01-14
[vm] minimize vec allocation and de-allocation in calls to native functions

mengxu-fb:
## Motivation

(part of the effort in porting [George's PR](https://github.com/gdanezis/libra/pull/1))

In current implementation of native functions,
1) allocation of a `Vec` (even we know it is going to be an empty `Vec`) is needed every time a native function returns, and
2) a `Vec` de-allocation is needed when the return values get consumed soon afterwards.

However, the allocation and de-allocation is not necessary. Currently, the native functions in the VM return either zero or one value, and none of them returns multiple values. This creates an optimization opportunity and there are two ways for this:

1) Converting the native return values from a `Vec` into an `enum` that has three  variants: `None`, `One`, and `Many` and only the `Many` variant will lead to vector allocation and de-allocation (which does not exist yet as of now). (We could make it an `Optional`, but to maintain compatibility and future-proof, we need the three-variant enum). This option is implemented up to commit ab07220

2) Replacing `Vec` with a `SmallVec<[Value; 1]>` that stores one value on stack and spills over automatically onto heap. This option is implemented in commit dad1f0f. 

Yes

## Test Plan

Running the `execution/executor-benchmark` does not show any difference in terms of TPS because the testing scope of the benchmark is too large. A new benchmark is added in `diem/language/benchmark` to benchmark VM's perf on native functions only.

However, the benchmark is very sensitive to fluctuations, and I haven't found a reliable way of measure the performance.

The most reliable number I can get is given through the following command, and by running it on an idle linux machine....:
`taskset -c 4 cargo bench --bench benchmarks -- --sample-size 1000 --measurement-time 20`
where `taskset` pins the process to a particular CPU core (CPU 4 in this case) and the sample size is increased to 1000 (default is 100).

Running it before and after optimization commits yields the following result:
```
arith                   time:   [14.989 ms 15.005 ms 15.024 ms]
                        change: [-0.7628% -0.6246% -0.4829%] (p = 0.00 < 0.05)
                        Change within noise threshold.
Found 87 outliers among 1000 measurements (8.70%)

call                    time:   [14.162 ms 14.166 ms 14.169 ms]
                        change: [+2.6035% +2.6754% +2.7435%] (p = 0.00 < 0.05)
                        Performance has regressed.
Found 49 outliers among 1000 measurements (4.90%)

natives                 time:   [17.454 ms 17.464 ms 17.475 ms]
                        change: [-10.551% -10.471% -10.395%] (p = 0.00 < 0.05)
                        Performance has improved
Found 102 outliers among 1000 measurements (10.20%)
```

This profile shows in general 10% improvement on VM's performance on native functions, but I won't read too much into the number, as the benchmark is highly sensitive and I have seen better and worse numbers as well. Let me know if you have better ways of making the test runs stable!

## Related PRs

[George's original PR](https://github.com/gdanezis/libra/pull/1)

### DIFF
--- a/language/move-vm/runtime/src/interpreter.rs
+++ b/language/move-vm/runtime/src/interpreter.rs
@@ -291,11 +291,7 @@ impl<L: LogContext> Interpreter<L> {
         match return_values {
             NativeReturnValues::None => {}
             NativeReturnValues::One(value) => self.operand_stack.push(value)?,
-            NativeReturnValues::Many(values) => {
-                for value in values {
-                    self.operand_stack.push(value)?;
-                }
-            }
+            NativeReturnValues::Many(values) => self.operand_stack.pushn(values)?,
         }
         Ok(())
     }
@@ -594,6 +590,20 @@ impl Stack {
         if self.0.len() < OPERAND_STACK_SIZE_LIMIT {
             self.0.push(value);
             Ok(())
+        } else {
+            Err(PartialVMError::new(StatusCode::EXECUTION_STACK_OVERFLOW))
+        }
+    }
+
+    /// Push n values on the stack
+    fn pushn(&mut self, mut values: Vec<Value>) -> PartialVMResult<()> {
+        if let Some(new_len) = self.0.len().checked_add(values.len()) {
+            if new_len <= OPERAND_STACK_SIZE_LIMIT {
+                self.0.append(&mut values);
+                Ok(())
+            } else {
+                Err(PartialVMError::new(StatusCode::EXECUTION_STACK_OVERFLOW))
+            }
         } else {
             Err(PartialVMError::new(StatusCode::EXECUTION_STACK_OVERFLOW))
         }
--- a/language/move-vm/runtime/src/interpreter.rs
+++ b/language/move-vm/runtime/src/interpreter.rs
@@ -18,6 +18,7 @@ use move_vm_types::{
     data_store::DataStore,
     gas_schedule::CostStrategy,
     loaded_data::runtime_types::Type,
+    natives::function::NativeReturnValues,
     values::{
         self, GlobalValue, IntegerValue, Locals, Reference, Struct, StructRef, VMValueCast, Value,
     },
@@ -284,11 +285,17 @@ impl<L: LogContext> Interpreter<L> {
         let native_function = function.get_native()?;
         let result = native_function.dispatch(&mut native_context, ty_args, arguments)?;
         cost_strategy.deduct_gas(result.cost)?;
-        let values = result
+        let return_values = result
             .result
             .map_err(|code| PartialVMError::new(StatusCode::ABORTED).with_sub_status(code))?;
-        for value in values {
-            self.operand_stack.push(value)?;
+        match return_values {
+            NativeReturnValues::None => {}
+            NativeReturnValues::One(value) => self.operand_stack.push(value)?,
+            NativeReturnValues::Many(values) => {
+                for value in values {
+                    self.operand_stack.push(value)?;
+                }
+            }
         }
         Ok(())
     }
--- a/language/move-vm/runtime/src/interpreter.rs
+++ b/language/move-vm/runtime/src/interpreter.rs
@@ -18,7 +18,6 @@ use move_vm_types::{
     data_store::DataStore,
     gas_schedule::CostStrategy,
     loaded_data::runtime_types::Type,
-    natives::function::NativeReturnValues,
     values::{
         self, GlobalValue, IntegerValue, Locals, Reference, Struct, StructRef, VMValueCast, Value,
     },
@@ -288,10 +287,8 @@ impl<L: LogContext> Interpreter<L> {
         let return_values = result
             .result
             .map_err(|code| PartialVMError::new(StatusCode::ABORTED).with_sub_status(code))?;
-        match return_values {
-            NativeReturnValues::None => {}
-            NativeReturnValues::One(value) => self.operand_stack.push(value)?,
-            NativeReturnValues::Many(values) => self.operand_stack.pushn(values)?,
+        for value in return_values {
+            self.operand_stack.push(value)?;
         }
         Ok(())
     }
@@ -590,20 +587,6 @@ impl Stack {
         if self.0.len() < OPERAND_STACK_SIZE_LIMIT {
             self.0.push(value);
             Ok(())
-        } else {
-            Err(PartialVMError::new(StatusCode::EXECUTION_STACK_OVERFLOW))
-        }
-    }
-
-    /// Push n values on the stack
-    fn pushn(&mut self, mut values: Vec<Value>) -> PartialVMResult<()> {
-        if let Some(new_len) = self.0.len().checked_add(values.len()) {
-            if new_len <= OPERAND_STACK_SIZE_LIMIT {
-                self.0.append(&mut values);
-                Ok(())
-            } else {
-                Err(PartialVMError::new(StatusCode::EXECUTION_STACK_OVERFLOW))
-            }
         } else {
             Err(PartialVMError::new(StatusCode::EXECUTION_STACK_OVERFLOW))
         }
--- a/language/move-vm/types/src/natives/function.rs
+++ b/language/move-vm/types/src/natives/function.rs
@@ -50,6 +50,17 @@ pub trait NativeContext {
     fn type_to_type_layout(&self, ty: &Type) -> PartialVMResult<Option<MoveTypeLayout>>;
 }
 
+/// The return value(s) of a native function
+///
+/// None - The native function does not return any thing
+/// One  - The native function returns exactly one value
+/// Many - The native function returns a vector of values
+pub enum NativeReturnValues {
+    None,
+    One(Value),
+    Many(Vec<Value>),
+}
+
 /// Result of a native function execution requires charges for execution cost.
 ///
 /// An execution that causes an invariant violation would not return a `NativeResult` but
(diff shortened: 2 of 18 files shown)


mengxu-fb:
@gdanezis, I can't find you in the reviewer list but would definitely need your opinion on this.

language/move-vm/runtime/src/interpreter.rs
```diff
@@ -288,11 +289,17 @@ impl<L: LogContext> Interpreter<L> {
...
             .result
             .map_err(|code| PartialVMError::new(StatusCode::ABORTED).with_sub_status(code))?;
-        for value in values {
-            self.operand_stack.push(value)?;
+        match return_values {
+            NativeReturnValues::None => {}
+            NativeReturnValues::One(value) => self.operand_stack.push(value)?,
+            NativeReturnValues::Many(values) => {
+                for value in values {
+                    self.operand_stack.push(value)?;
```
gdanezis, line 300: Here can we first extend the operand stack to have a capacity the length of the values, so that it does not increase incrementally? Also I see we are already mapping the error above (line 292). Can we  exit there and only do an append to the operand stack vector here?

gdanezis (approved):
Small comment to avoid double copy of array -- otherwise all good (was also check before by Dario).

language/move-vm/runtime/src/interpreter.rs
mengxu-fb: Right, we should do an append instead of the for loop. Will update it.

tnowacki:
Do we have numbers for this yet? I'm guessing the goal is to reduce memory overhead? I'm just intrigued by how much of a difference it will make, I do not have a good intuition for performance differences for a change like this in Rust

mengxu-fb:
Yes, I am going to add some actual number later. Got trapped in the module republishing PRs :(

sunshowers:
(note than an empty vec doesn't allocate, but the single element vs multi element result might make a difference—woule smallvec be a good tool for that?)

mengxu-fb:
`smallvec` would be a nice fit but I don't see it being used anywhere in the repo. Is there a reason to not use it? Or it is just not necessary yet?

runtian-zhou (approved):
LGTM modulo smallvec. But I guess we can worry about that later

language/move-vm/types/src/natives/function.rs
```diff
@@ -52,6 +52,17 @@ pub trait NativeContext {
     fn is_resource(&self, ty: &Type) -> bool;
 }
 
+/// The return value(s) of a native function
```
runtian-zhou, line 55: Maybe add a comment here to mention that we did this intentionaaly to avoid allocation overhead of vectors, and suggests using smallvec later if necessary?

mengxu-fb: Will do.

libra-action:
```
target packages:
  M backtrace 0.3.55 (direct third-party, crates.io)
    * version DOWNGRADED from 0.3.56
    * (unchanged features: addr2line, default, gimli-symbolize, miniz_oxide, object, serde, std)
  M fiat-crypto 0.1.5 (transitive third-party, crates.io)
    * version DOWNGRADED from 0.1.6
  M object 0.22.0 (transitive third-party, crates.io)
    * version DOWNGRADED from 0.23.0
    * (unchanged features: archive, coff, elf, macho, pe, read_core, unaligned)
  M parking_lot 0.11.1 (transitive third-party, crates.io)
    * status changed from direct third-party
    * (unchanged features: default)
```

libra-action:
```
target packages:
  M backtrace 0.3.55 (direct third-party, crates.io)
    * version DOWNGRADED from 0.3.56
    * (unchanged features: addr2line, default, gimli-symbolize, miniz_oxide, object, serde, std)
  M fiat-crypto 0.1.5 (transitive third-party, crates.io)
    * version DOWNGRADED from 0.1.6
  M object 0.22.0 (transitive third-party, crates.io)
    * version DOWNGRADED from 0.23.0
    * (unchanged features: archive, coff, elf, macho, pe, read_core, unaligned)
  M parking_lot 0.11.1 (transitive third-party, crates.io)
    * status changed from direct third-party
    * (unchanged features: default)
  M tokio-stream 0.1.1 (transitive third-party, crates.io)
    * version DOWNGRADED from 0.1.2
    * (unchanged features: default, time)
  M tokio-util 0.6.1 (transitive third-party, crates.io)
    * version DOWNGRADED from 0.6.2
    * (unchanged features: codec, default)
  R async-stream 0.3.0 (transitive third-party, crates.io)
    * (old features: [none])

host packages:
  R async-stream-impl 0.3.0 (transitive third-party, crates.io)
    * (old features: [none])
```

libra-action:
```
target packages:
  A state-synchronizer 0.1.0 (workspace, path 'state-synchronizer')
    * features: default
  M backtrace 0.3.55 (direct third-party, crates.io)
    * version DOWNGRADED from 0.3.56
    * (unchanged features: addr2line, default, gimli-symbolize, miniz_oxide, object, serde, std)
  M tokio-stream 0.1.1 (direct third-party, crates.io)
    * version DOWNGRADED from 0.1.2
    * (unchanged features: default, time)
  M tokio-util 0.6.1 (direct third-party, crates.io)
    * version DOWNGRADED from 0.6.2
    * (unchanged features: codec, compat, default, futures-io, io)
  M warp 0.3.0 (direct third-party, crates.io)
    * removed features: tls, tokio-rustls
    * (unchanged features: default, multipart, tokio-tungstenite, websocket)
  M fiat-crypto 0.1.5 (transitive third-party, crates.io)
    * version DOWNGRADED from 0.1.6
  M getrandom 0.2.1 (transitive third-party, crates.io) (other versions: 0.1.15)
    * version DOWNGRADED from 0.2.2
    * (unchanged features: std)
  M object 0.22.0 (transitive third-party, crates.io)
    * version DOWNGRADED from 0.23.0
    * (unchanged features: archive, coff, elf, macho, pe, read_core, unaligned)
  M parking_lot 0.11.1 (transitive third-party, crates.io)
    * status changed from direct third-party
    * (unchanged features: default)
  R state-sync 0.1.0 (workspace, path 'state-sync')
    * (old features: default)
  R async-stream 0.3.0 (transitive third-party, crates.io)
    * (old features: [none])
  R ring 0.16.19 (transitive third-party, crates.io)
    * (old features: alloc, default, dev_urandom_fallback, once_cell)
  R rustls 0.19.0 (transitive third-party, crates.io)
    * (old features: default, log, logging)
  R sct 0.6.0 (transitive third-party, crates.io)
    * (old features: [none])
  R spin 0.5.2 (transitive third-party, crates.io)
    * (old features: [none])
  R tokio-rustls 0.22.0 (transitive third-party, crates.io)
    * (old features: [none])
  R untrusted 0.7.1 (transitive third-party, crates.io)
    * (old features: [none])
  R webpki 0.21.4 (transitive third-party, crates.io)
    * (old features: default, std, trust_anchor_util)

host packages:
  M getrandom 0.2.1 (transitive third-party, crates.io)
    * version DOWNGRADED from 0.2.2
    * (unchanged features: std)
  R async-stream-impl 0.3.0 (transitive third-party, crates.io)
    * (old features: [none])
```

bob-wilson:
> LGTM modulo smallvec. But I guess we can worry about that later

If we use smallvec, this PR won't be needed, will it? I'm not sure it makes sense to land this now if we're thinking to use smallvec

mengxu-fb:
We still need the benchmarking part of this PR, but the {None, One, Many} enum should be replaced with `SmallVec<Value; 1>`.

mengxu-fb:
Second thought, let me just experiment with `smallvec` and let's see how the code will look like.

runtian-zhou:
> > LGTM modulo smallvec. But I guess we can worry about that later
> 
> If we use smallvec, this PR won't be needed, will it? I'm not sure it makes sense to land this now if we're thinking to use smallvec

I thought we can still 

> > LGTM modulo smallvec. But I guess we can worry about that later
> 
> If we use smallvec, this PR won't be needed, will it? I'm not sure it makes sense to land this now if we're thinking to use smallvec

My thought is that we could keep the `NativeResult ` and replace the inner definition  with a smallvec later if we want.

mimoo:
> smallvec would be a nice fit but I don't see it being used anywhere in the repo. Is there a reason to not use it? Or it is just not necessary yet?

(smallvec is actually used transitively, not directly)

libra-action:
```
target packages:
  M smallvec 1.6.1 (direct third-party, crates.io)
    * status changed from transitive third-party
```

libra-action:
```
target packages:
  M smallvec 1.6.1 (direct third-party, crates.io)
    * status changed from transitive third-party
```

mengxu-fb:
Benchmarking the changes before and after `SmallVec` (e.g., Option 1 vs Option 2) does not show much difference:

```
arith                   time:   [15.591 ms 15.607 ms 15.624 ms]
                        change: [-2.7872% -2.6375% -2.4789%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 117 outliers among 1000 measurements (11.70%)

call                    time:   [14.240 ms 14.258 ms 14.278 ms]
                        change: [+0.1942% +0.3088% +0.4558%] (p = 0.00 < 0.05)
                        Change within noise threshold.
Found 39 outliers among 1000 measurements (3.90%)

natives                 time:   [17.465 ms 17.486 ms 17.508 ms]
                        change: [+0.4599% +0.6053% +0.7593%] (p = 0.00 < 0.05)
                        Change within noise threshold.
Found 94 outliers among 1000 measurements (9.40%)
```

Benchmarking the changes of `SmallVec` against the un-optimized version yields the following result:
```
arith                   time:   [14.838 ms 14.845 ms 14.854 ms]
                        change: [-1.9848% -1.8837% -1.7846%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 32 outliers among 1000 measurements (3.20%)

call                    time:   [14.369 ms 14.373 ms 14.377 ms]
                        change: [-0.2469% -0.1915% -0.1439%] (p = 0.00 < 0.05)
                        Change within noise threshold.
Found 17 outliers among 1000 measurements (1.70%)

natives                 time:   [17.541 ms 17.553 ms 17.565 ms]
                        change: [-5.5716% -5.4962% -5.4121%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 89 outliers among 1000 measurements (8.90%)
```

mengxu-fb:
Rebased to master, let's go with the `smallvec` optimization.

mehrab2585 (approved):

libra-action:
```
target packages:
  M smallvec 1.6.1 (direct third-party, crates.io)
    * status changed from transitive third-party
```

libra-action:
```
target packages:
  M smallvec 1.6.1 (direct third-party, crates.io)
    * status changed from transitive third-party
```

libra-action:
```
target packages:
  M smallvec 1.6.1 (direct third-party, crates.io)
    * status changed from transitive third-party
```

libra-action:
```
target packages:
  M smallvec 1.6.1 (direct third-party, crates.io)
    * status changed from transitive third-party
```

PR #7252, opened 2021-01-15
[diem-types] Update currency code restrictions now that we have the real codes.

tzakian:
This updates the currency code restrictions now that we have the actual currency codes, and to be in line with the restrictions described in the [DIP-20 draft](https://github.com/diem/dip/blob/master/dips/dip-20.md#operational-restrictions). 

## Testing
Adds two tests:
* A simple handwritten test to test out corner cases.
* A proptest to make sure all identifiers of the form `[A-Z][A-Z0-9]*` are accepted as valid currency codes.


bob-wilson (approved):
Thanks! This fixes https://github.com/diem/diem/issues/6082

PR #7254, opened 2021-01-15
Gha release 1.1

rexhoffman:
## Motivation

Prepare for gha use in release 1.1 branch for pr/bors usage.
 
Yes

## Test Plan

CI

## Related PRs

None outstanding.


docker/ci/github/Dockerfile
```diff
@@ -0,0 +1,44 @@
+FROM debian:buster-20201209@sha256:5b5fa7e155b1f19dffb996ea64e55520b80d5bd7a8fdb5aed1acabd217e9ed59 AS setup_ci
+#FROM debian:buster-20201117-slim@sha256:21ce4b82ed3425e757b1ac98ff9cbeb83540034c0cbc32766a59c3eca9e0daf8 AS setup_ci
```
sunshowers: Remove this line?

.github/workflows/ci-test.yml
```diff
@@ -7,35 +7,103 @@ on:
...
+      - id: changes
+        name: determine changes
+        uses: ./.github/actions/changes
+        with:
+          workflow-file: ci-test.yml
+      - id: rust-changes
+        name: find rust/cargo changes.
+        uses: ./.github/actions/matches
+        with:
+          pattern: '**.rs\|Cargo.toml\|Cargo.lock'
```
sunshowers: I'm kinda of concerned about this as I mentioned elsewhere -- this is missing a lot of different files that affect the Rust build. Could we at least unconditionally use the determinator for Rust files for now?

.github/workflows/ci-test.yml
```diff
@@ -86,6 +86,7 @@ jobs:
 
   unit-test:
     runs-on: ubuntu-latest-xl
+    timeout-minutes: 70
```
sunshowers: Could you add a comment explaining how 70 was chosen? For future ours' sake :)

sausagee (approved):

.github/workflows/ci-test.yml
```diff
@@ -21,6 +21,9 @@ jobs:
     steps:
       - uses: actions/checkout@v2
         with:
+          # This ensures that the tip of the PR is checked out instead of the merge between the base ref and the tip
+          # On `push` this value will be empty and will "do-the-right-thing"
+          ref: ${{ github.event.pull_request.head.sha }}
```
sunshowers, line 33: Huh, I don't know what this means at all... what does "checkout tip of PR instead of merge commit" mean? What's the "merge commit" in this context? I'm quite shocked checking out the head of the PR isn't the default behavior of `actions/checkout`.

https://github.com/actions/checkout#usage says:

```
    # The branch, tag or SHA to checkout. When checking out the repository that
    # triggered a workflow, this defaults to the reference or SHA for that event.
    # Otherwise, uses the default branch.
```

This seems like the same thing as the head commit? What's the difference here? Could you help by drawing a DAG?

sunshowers (commented):
The PR looks really nice overall. A few things I spotted.

.github/actions/changes/readme.md
```diff
@@ -0,0 +1,55 @@
+# Docker Hub Login and Image Signing #
```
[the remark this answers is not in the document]
rexhoffman: heh, yep... will fix in master.

.github/workflows/ci-test.yml
rexhoffman: It was updated:   I agree it was risky before.

        with:
          pattern: '^documentation\|^docker\|^scripts'
          invert: "true"

.github/workflows/ci-test.yml
rexhoffman: This is because gha checkout be default doesn't pull a sha from the repo is makes it's own squash - and has no history at all in the pull.

.github/workflows/ci-test.yml
rexhoffman: It was dropped to 40, 70 was way overkill, should 32-33 at most.

PR #7257, opened 2021-01-15
[move][testing] Make dependencies in IR tests unique

tnowacki:
- Keyed compiled module dependencies in the IRCompiler instance of the testing Compiler trait

## Motivation

- Uniqueness is needed in an upcoming IR update

## Test Plan

- ran tests
- Theoretically a perf increase if there were a TON of duplicate modules, as the IR compiler generates the dep before adding it to it's own map (and shadowing the old one). In practice, no change as there really aren't duplicate modules except for about 5 IR tests

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3528,6 +3528,7 @@ dependencies = [
  "diem-workspace-hack",
  "functional-tests",
  "ir-to-bytecode",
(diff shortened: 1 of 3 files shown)


vgao1996 (approved):
LGTM

PR #7258, opened 2021-01-15
readme: update rustdoc link and remove circleci status

bmwill:


JoelMarcey (approved):
LGTM - we can see if we can create a sub domain later that is crates.diem.com or something maybe. But let's do this for now in the interim.

PR #7265, opened 2021-01-16
[gha] single job if needed for bors

rexhoffman:
## Motivation

GHA doesn't do a good job of providing the identity of a job that has succeeded/failed.   Less so for a workflow file.
In order for bors to have something to hold on to, creating a success job that'll fast fail if any job fails.

So this was a bit tricky.   The result from a job is a string of "failure","skipped","success" or "cancelled", see:

https://docs.github.com/en/actions/reference/context-and-expression-syntax-for-github-actions#needs-context

Fun to figure out.

Also if a needed job was skipped by default, the dependent job is skipped as well (hence the "always()" in the if line of the job).  Commented inline.

Need to test this by creating failure cases and testing, as well as success case.  Created a PR for each and tested there:

Success PR: https://github.com/diem/diem/actions/runs/494420181 -- notice it works with skips as well.

And failure PR:  https://github.com/diem/diem/actions/runs/494729561.

This is as easy as gha can make it for bors.

Y

## Test Plan

Two draft PRs one forcing success and another failure.  Feel free to inspect, will close when merged/cancelled.
https://github.com/diem/diem/pull/7267
https://github.com/diem/diem/pull/7266

## Related PRs

https://github.com/diem/diem/pull/7267
https://github.com/diem/diem/pull/7266

### DIFF
--- a/.github/actions/matches/action.yml
+++ b/.github/actions/matches/action.yml
@@ -45,7 +45,7 @@ runs:
         if [[ -f "${TARGET_FILE}" ]]; then
           echo found: 1>&2
           cat "${TARGET_FILE}" | grep "${INV}" "${PATTERN}" || echo No matching files. 1>&2
-          count="$( cat "${TARGET_FILE}" | grep -c "${INV}" "${PATTERN}" )" || true
+          count="$( cat "${TARGET_FILE}" | grep -c "${INV}" "${PATTERN}" || echo 0)"
           if [[ "$count" == 0 ]]; then
             result=false
           fi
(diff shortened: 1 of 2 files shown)


rexhoffman:
Eh, not the best.   The success job doesn't get marked as skipped when the first failure of a job it depends occurs.  

Maybe also need the tear down step to emit a job failure message in some secure way?

image

rexhoffman:
If we decide to use this anyway, drop last commit before merging.

bmwill (approved):

PR #7268, opened 2021-01-18
[move-prover] Refactoring: forking boogie translator and refining bytecode translator

wrwg:
This is an incremental step towards prover refactoring for bytecode transformation architecture:

- Forking the boogie translator into a new crate `move-prover/boogie-backend`. This contains a copy of the (much simplified) version of bytecode and spec expressions to boogie translation. All logic about translating function spec blocks is gone as this happens now as bytecode transformation. According to line count, the overall code size is shrinking down to 30%. *Unfortunately we cannot see the diff* of this fork with github, and everything in the new boogie-backend appears as new, but it is not.
- The old backend is still there and runs by default. To use the new backend and bytecode from the command line, one can use `--v2`.
- Moving `ProverOptions` and `BoogieOptions` out of `cli.rs` and into the according crates, so it can be accessed there.
- Refining `SpecInstrumenter` to more precisely deal with the different abort condition failures and generate the same diagnosis (or better) than the old Boogie backend.
- Implementing tracing for execution traces as a bytecode transformation `DebugInstrumenter`.
- The tests `arithm.move`, `aborts_if.move`, and `aborts_if_with_code.move` are now operating with the v2 compilation scheme.

Invariants are still not supported, as well as many functional tests haven't been ported to v2 yet (specifically around resource access).

## Motivation

Simplify prover architecture and make it less dependent from Boogie.

Yes

## Test Plan

Ported some existing tests.

## Related PRs

NA


libra-action:
```
target packages:
  A boogie-backend 0.1.0 (workspace, path 'language/move-prover/boogie-backend')
    * features: [none]
```

language/move-model/src/model.rs
```diff
@@ -2519,3 +2553,23 @@ impl<'env> FunctionEnv<'env> {
...
+// =================================================================================================
+/// # Formatting
+
+impl fmt::Display for ConditionTag {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        use ConditionTag::*;
+        match self {
+            Requires => write!(f, "requires"),
+            Ensures => write!(f, "ensures"),
+            NegativeTest => write!(f, "negative-test"),
```
shazqadeer, line 2566: Is NegativeTest for specifications that we expect to fail (similar to Kevin's project)?  If yes, I was expecting it to be a sub-classification of requires/ensures.

language/move-prover/bytecode/src/debug_instrumentation.rs
```diff
@@ -0,0 +1,91 @@
+// Copyright (c) The Diem Core Contributors
+// SPDX-License-Identifier: Apache-2.0
+
+//! Transformation injection trace instructions which are used to visualize execution.
```
shazqadeer, line 4: injection -> injects?

language/move-model/src/builder/module_builder.rs
```diff
@@ -1411,6 +1411,11 @@ impl<'env, 'translator> ModuleBuilder<'env, 'translator> {
                 .error(loc, "decreases specification not supported currently");
             return;
         }
+        if matches!(kind, ConditionKind::SucceedsIf) {
+            self.parent
+                .error(loc, "condition kind is not longer supported");
```
shazqadeer, line 1416: Drop "longer"?

"condition kind is not supported"

shazqadeer (approved):

language/move-model/src/model.rs
wrwg: Yes. But I expect to refactor this after the old backend is retired. There is a lot of mess right now how verification conditions are associated with assert/assume in the code (the ConditionTag and ConditionKind) which needs to be overhauled.

language/move-prover/bytecode/src/debug_instrumentation.rs
wrwg: Done

language/move-model/src/builder/module_builder.rs
wrwg: Done

libra-action:
```
target packages:
  A boogie-backend 0.1.0 (workspace, path 'language/move-prover/boogie-backend')
    * features: [none]
```

libra-action:
```
target packages:
  A boogie-backend 0.1.0 (workspace, path 'language/move-prover/boogie-backend')
    * features: [none]
```

PR #7272, opened 2021-01-19
[prover] separate functions for is_native and is_intrinsic

sblackshear:
Previously, is_native conflated these two concepts, which frustrates analysis clients that want to treat these differently.
The motivation for this change is to support the read/write set analysis, which needs to analyze the body of functions marked as intrinsic.

The PR changes all uses of `is_native` in bytecode_translator.rs and spec_instrumentation.rs to `is_native && is_intrinsic`. There are more uses of `is_native` in some of the other bytecode analyses/transformations (e.g., borrow_analysis), but I left them as is because it seems like they also want to ignore the intrinsic pragma. Let me know if this wrong and I can fix them.

### DIFF
--- a/language/move-model/src/model.rs
+++ b/language/move-model/src/model.rs
@@ -2248,11 +2248,15 @@ impl<'env> FunctionEnv<'env> {
         }
     }
 
-    /// Returns true if this function is native. The function is also marked as native
-    /// if it has the pragma intrinsic set to true.
+    /// Returns true if this function is native.
     pub fn is_native(&self) -> bool {
         let view = self.definition_view();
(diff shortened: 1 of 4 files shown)


shazqadeer:
How is is_intrinsic different from is_native?

sblackshear:
> How is is_intrinsic different from is_native?

Native is a Move VM feature that allows Move functions to be implemented in the host language (e.g., Rust), whereas intrinsic is a Move prover feature that (to my understanding) allows the prover to use a handwritten model of a Move function instead of its Move bytecode implementation. 

A function `foo` is marked native via `public native fun foo() { ... }` and intrinsic via `spec module { pragma intrinsic = true }`

wrwg (approved):

PR #7275, opened 2021-01-19
[json-rpc] allow no params field in JSON-RPC API call request

### LINKED ISSUE #7208: JSON-RPC server requires params
The JSON-RPC server implementation violates the [JSON-RPC 2.0 specification](https://www.jsonrpc.org/specification) by rejecting requests without the `params` field.

The spec says:

> params
> A Structured value that holds the parameter values to be used during the invocation of the method. This member MAY be omitted.

However, there is a unit test that asserts the server returns an error when this field is omitted.

https://github.com/diem/diem/blob/e7654920e355498055ff2979c8029d2571b2109d/json-rpc/src/tests/unit_tests.rs#L280-L281

This may cause some JSON-RPC client implementations to fail to send certain requests to the server. For example, it's common practice to set the `omitempty` flag in the Go language, where empty arrays and nulls are synonymous.

xli:
## Motivation

Resolve #7208

Yes

## Test Plan

Unit test


libra-action:

Breaking changes policy:

2. Adding fields is ok.

bmwill (approved):
lgtm!

PR #7282, opened 2021-01-20
[RFC][vm] Reduce unneeded copy for obtaining resource bytes

runtian-zhou:
## Motivation

In our code there's multiple reads from the cache which doesn't really require owning the bytes: only those reads straight from the `StateView` will return the owned instance of vector bytes. This PR cleans these unnecessary allocations up by using a referenence when possible.

Yes

## Test Plan

I ran the `cargo bench` on `txn_benchs`. The perf was improved by 2%.

### DIFF
--- a/language/move-vm/runtime/src/data_cache.rs
+++ b/language/move-vm/runtime/src/data_cache.rs
@@ -16,7 +16,7 @@ use move_vm_types::{
     loaded_data::runtime_types::Type,
     values::{GlobalValue, GlobalValueEffect, Value},
 };
-use std::collections::btree_map::BTreeMap;
+use std::{borrow::Cow, collections::btree_map::BTreeMap};
 use vm::errors::*;
 
 /// Trait for the Move VM to abstract storage operations.
@@ -44,7 +44,7 @@ pub trait RemoteCache {
         &self,
         address: &AccountAddress,
         tag: &StructTag,
-    ) -> PartialVMResult<Option<Vec<u8>>>;
+    ) -> PartialVMResult<Option<Cow<[u8]>>>;
 }
 
 pub struct AccountDataCache {
--- a/language/diem-vm/src/data_cache.rs
+++ b/language/diem-vm/src/data_cache.rs
@@ -19,7 +19,7 @@ use move_core_types::{
     language_storage::{ModuleId, StructTag},
 };
 use move_vm_runtime::data_cache::RemoteCache;
-use std::collections::btree_map::BTreeMap;
+use std::{borrow::Cow, collections::btree_map::BTreeMap};
 use vm::errors::*;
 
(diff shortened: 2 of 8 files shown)


language/move-vm/runtime/src/data_cache.rs
```diff
@@ -42,7 +44,24 @@ pub trait RemoteCache {
         &self,
         address: &AccountAddress,
         tag: &StructTag,
-    ) -> PartialVMResult<Option<Vec<u8>>>;
+    ) -> PartialVMResult<Option<RefBytes>>;
+}
+
+
+pub enum RefBytes<'life> {
```
sblackshear, line 51: Nit: `'life` is less idiomatic than `'a` and no more descriptive...

sblackshear (commented):
> I ran the cargo bench on txn_benchs. The perf was improved by 2%.

Is this run in CI? That would be very useful for preventing regressions by a well-meaning person trying to simplify this code.

runtian-zhou:
> > I ran the cargo bench on txn_benchs. The perf was improved by 2%.
> 
> Is this run in CI? That would be very useful for preventing regressions by a well-meaning person trying to simplify this code.

I'm working with @sausagee and @rexhoffman to get some metric from the CI code.

zekun000:
maybe worth considering using some abstraction like Bytes crate for memory sharing

language/move-vm/runtime/src/data_cache.rs
```diff
@@ -42,7 +42,23 @@ pub trait RemoteCache {
         &self,
         address: &AccountAddress,
         tag: &StructTag,
-    ) -> PartialVMResult<Option<Vec<u8>>>;
+    ) -> PartialVMResult<Option<RefBytes>>;
+}
+
+pub enum RefBytes<'life> {
```
phlip9, line 48: is that a bootleg `Cow<'a, [u8]>` I see? : )
https://doc.rust-lang.org/std/borrow/enum.Cow.html

zekun000: ah I was trying to find something in std but ended up AsRef which requires 'static

phlip9: yeah, Cow is pretty cool! lets you return either an `Owned(Vec<u8>)` or a `Borrowed(&'a [u8])`. iirc there is also `supercow` (https://altsysrq.github.io/rustdoc/supercow/index.html), which `guppy` also uses.

tnowacki: Should we just be using `Cow` here?

runtian-zhou: I think we should use `Cow`.

davidiw:
Old PR, feel free to re-open.

PR #7291, opened 2021-01-20
[network][docs] Fixup network documentation. Fix rustdocs links.

phlip9:
+ Docs now reflect the current state of the network module after the `Peer` refactor: #6800

+ Fixed the rustdoc documentation links so they resolve correctly.

+ Link to the specifications where relevant.

### DIFF
--- a/network/README.md
+++ b/network/README.md
@@ -4,45 +4,43 @@ title: Network
 custom_edit_url: https://github.com/diem/diem/edit/master/network/README.md
 ---
 
-
-The network component provides peer-to-peer communication primitives to other
-components of a validator.
-
 ## Overview
 
-For more detailed info, see the [Network Specification](../specifications/network/README.md).
+For more detailed info, see the [DiemNet Specification](../specifications/network/README.md).
 
-The network component is specifically designed to facilitate the consensus and
-shared mempool protocols. Currently, it provides these consumers with two
-primary interfaces:
-* RPC, for Remote Procedure Calls; and
-* DirectSend, for fire-and-forget style message delivery to a single receiver.
+DiemNet is the primary protocol for communication between any two nodes in the
+Diem ecosystem. It is specifically designed to facilitate the consensus, shared
+mempool, and state sync protocols. DiemNet tries to maintain at-most one connection
+with each remote peer; the application protocols to that remote peer are then
+multiplexed over the single peer connection.
+
+Currently, it provides application protocols with two primary interfaces:
+
+* DirectSend: for fire-and-forget style message delivery.
+* RPC: for unary Remote Procedure Calls.
 
 The network component uses:
+
 * TCP for reliable transport.
-* [Noise](https://noiseprotocol.org/noise.html) for authentication and full
- end-to-end encryption.
-* On-chain addresses for discovery, with a seed peers file for initial discovery.
+* [NoiseIK] for authentication and full end-to-end encryption.
+* On-chain [`NetworkAddress`](./network-address/src/lib.rs) set for discovery, with
+  optional seed peers in the [`NetworkConfig`](../config/src/config/network_config.rs)
+  as a fallback.
 
-Each new substream is assigned a *protocol* supported by both the sender and
-the receiver. Each RPC and DirectSend type corresponds to one such protocol.
+Validators will only allow connections from other validators. Their identity and
+public key information is provided by the [`simple-onchain-discovery`] protocol,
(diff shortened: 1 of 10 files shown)


network/src/protocols/wire/handshake/v1/mod.rs
```diff
@@ -2,13 +2,15 @@
 // SPDX-License-Identifier: Apache-2.0
 
 //! This module defines the structs transported during the network handshake protocol v1.
-//! These should serialize as per [link](TODO: Add ref).
+//! These should serialize as per the [DiemNet Handshake v1 Specification].
 //!
 //! During the v1 Handshake protocol, both end-points of a connection send a serialized and
-//! length-prefixed `HandshakeMsg` to each other. The handshake message contains a map from
+//! length-prefixed [`HandshakeMsg`] to each other. The handshake message contains a map from
```
gregnazario, line 8: What's the purpose of the brackets btw?  Will it link back to the original type?

network/src/protocols/rpc/mod.rs
```diff
@@ -318,6 +319,8 @@ impl InboundRpcs {
 }
 
 /// `OutboundRpcs` handles new outbound rpc requests made from the application layer.
+///
+/// There is one `OutboundRpcs` handler per connection.
```
gregnazario, line 323: Perhaps per Peer is more appropriate?  Or maybe just drop this comment since we have only one peer per connection?  And I'm not sure if we had a multiple connections to a single peer in the future, we'd need multiple outbound RPC handlers.

network/README.md
```diff
@@ -4,45 +4,45 @@ title: Network
...
+public key information is provided by the [`simple-onchain-discovery`] protocol,
+which updates the eligible member information on each consensus reconfiguration.
+
+In contrast, Validator Full Node (VFNs) servers will only prioritize connections
+from more trusted peers in the on-chain discovery set; they will still service
+any public clients. Public Full Nodes (PFNs) connecting to VFNs will always
+authenticate the VFN server using the available discovery information.
+
+Each member of the validator network maintains a full membership view and connects
+directly to all other validators in order to maintain a full-mesh network.
```
gregnazario, line 41: Move this up next to the `Validator` section to keep it together with the other validator comments

network/README.md
```diff
@@ -51,80 +51,85 @@ partial membership views, sophisticated failure detectors, or network overlays.
...
+| Consensus | Mempool | State Sync | Health |  Application Modules
++-----------+---------+------------+--------+
+      ^          ^          ^           ^
+      |          |          |           |      Network Interface
+      v          v          v           v
++----------------+--------------------------+
+| Peer Manager  <->   Connectivity Manager  |  Network Module
++----------------+--------------------------+
+|    Peer(s)     |
++----------------+
```
gregnazario, line 65: Might be more appropriate to put the connectivity manager directly to the side, everything really kind of talks to Peer Manager except for discovery

```
+-----------+---------+------------+--------+
| Consensus | Mempool | State Sync | Health |  Application Modules
+-----------+---------+------------+--------+
      ^          ^          ^           ^
      |          |          |           |      Network Interface
      v          v          v           v
+------------------------------------------+   +----------------------+
|              Peer Manager                |<->| Connectivity Manager |  Network Module
+------------------------------------------+   +----------------------+
|    Peer(s)     |
+----------------+
```

network/src/noise/handshake.rs
```diff
@@ -693,8 +693,8 @@ mod test {
         client.network_context = NetworkContext::mock_with_peer_id(PeerId::random());
         let (client_res, server_res) = perform_handshake(&client, &server, server_public_key);
 
-        trace!("client_res: {}", client_res.unwrap_err());
-        trace!("server_res: {}", server_res.unwrap_err());
+        client_res.unwrap_err();
+        server_res.unwrap_err();
```
gregnazario, line 697: This test doesn't check the error?

Also, let's stick to docs, or code fixes at once

network/src/protocols/wire/handshake/v1/mod.rs
phlip9: Right, it'll link to the definition in the rendered rustdocs. You can view the rendered version with:

```
cargo doc --all-features -p network --open
```

network/src/noise/handshake.rs
phlip9: `unwrap_err` will panic if it's not an `Err` : )

gregnazario (approved):

PR #7297, opened 2021-01-21
[json-rpc] Support TLS on JSON-RPC port

mgorven:
If `tls_cert_path` and `tls_key_path` are provided, enable TLS on the
JSON-RPC port. This may be necessary in deployments without a load
balancer, or to provide additional security between a load balancer and
the node.

Test Plan: Ran fullnode with and without TLS, tested JSON-RPC request.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5641,6 +5641,21 @@ dependencies = [
 ]
 
 [[package]]
+name = "ring"
+version = "0.16.19"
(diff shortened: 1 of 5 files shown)


libra-action:

Breaking changes policy:

2. Adding fields is ok.

xli (approved):

mgorven:
Warp only supports RusTLS, so that gets pulled in again unfortunately.

PR #7305, opened 2021-01-21
[writeset-generator] Filter out Genesis module from release process.

runtian-zhou:
## Motivation

Currently `Genesis` will be included in the `stdlib_modules`. We will need to manually filter those modules out from the release process so that they won't be published on-chain during the new release process.

Yes

## Test Plan
Manually tested the binary and make sure genesis module isn't included.


mengxu-fb (commented):
The changes LGTM. Just want to comment that the flow mimics the `publish_stdlib` function in `vm-genesis` (or rather, summarizes its net effects in the WriteSet).

mengxu-fb (approved):

PR #7306, opened 2021-01-21
[x] adding json/toml output to generate-summaries

mimoo:


devtools/x/src/generate_summaries.rs
```diff
@@ -18,6 +18,9 @@ pub struct Args {
     #[structopt(name = "OUT_DIR")]
     /// Directory to output summaries to (default: target/summaries)
     out_dir: Option<PathBuf>,
+    #[structopt(name = "OUTPUT_FORMAT")]
+    /// Output in text, toml or json
+    output_format: Option<String>,
```
sunshowers, line 23: This should be an enum using `arg_enum!`, here's an example: https://github.com/diem/diem/blob/39e4d7a064ea4b9d071c56d9df5099b655c4b8f8/devtools/x/src/cargo/build_args.rs#L7

In general, input validation should occur as close to the edges of the program as possible.

devtools/x/src/generate_summaries.rs
```diff
@@ -77,25 +98,39 @@ fn write_summary(
...
+    let (out, path) = match output_format {
+        Some("json") => {
+            let out = serde_json::to_string(&summary)?;
+            let summary_path = out_dir.join(format!("summary-{}.json", name));
+            (out, summary_path)
+        }
+        Some("toml") => {
+            let out = toml::to_string(&summary)?;
+            let summary_path = out_dir.join(format!("summary-{}.toml", name));
+            (out, summary_path)
```
sunshowers, line 115: Please continue to use the `write_to_string` method.

sunshowers (changes requested):

devtools/x/src/generate_summaries.rs
mimoo: you mean to have the text at the beginning of the doc? The problem with that is that it makes the json file invalid, but it works with toml so I can have it for toml if that's what you're asking?

mimoo: oh wait, the default text was already toml 🤦🏻‍♂️

devtools/x/src/generate_summaries.rs
mimoo: done!

devtools/x/src/generate_summaries.rs
mimoo: done!

devtools/x/src/generate_summaries.rs
```diff
@@ -11,13 +11,24 @@ use std::{
...
+}
 
 #[derive(Debug, StructOpt)]
 pub struct Args {
     #[structopt(name = "OUT_DIR")]
     /// Directory to output summaries to (default: target/summaries)
     out_dir: Option<PathBuf>,
+    #[structopt(name = "OUTPUT_FORMAT")]
+    /// Output in text, toml or json
+    output_format: Option<OutputFormat>,
```
sunshowers, line 31: I think the only supported output formats are toml and json? If so I think you can just use a non-Option enum here and set a default: https://github.com/diem/diem/blob/39e4d7a064ea4b9d071c56d9df5099b655c4b8f8/devtools/x/src/cargo/build_args.rs#L85-L87

mimoo: simplified : o tested, it works!

devtools/x/src/diff_summary.rs
```diff
@@ -15,9 +24,9 @@ pub struct Args {
     #[structopt(name = "COMPARE_SUMMARY")]
     /// Path to the comparison summary
     compare_summary: PathBuf,
-    #[structopt(name = "OUTPUT_FORMAT")]
+    #[structopt(name = "OUTPUT_FORMAT", default_value = "Text")]
```
sunshowers, line 27: This should also have a `possible_values` argument: https://github.com/diem/diem/blob/39e4d7a064ea4b9d071c56d9df5099b655c4b8f8/devtools/x/src/cargo/build_args.rs#L85

devtools/x/src/generate_summaries.rs
```diff
@@ -11,13 +11,24 @@ use std::{
...
+        Json,
+    }
+}
 
 #[derive(Debug, StructOpt)]
 pub struct Args {
     #[structopt(name = "OUT_DIR")]
     /// Directory to output summaries to (default: target/summaries)
     out_dir: Option<PathBuf>,
+    #[structopt(name = "OUTPUT_FORMAT", default_value = "Toml")]
```
sunshowers, line 29: as should this :)

sunshowers (approved):
Looks great, thanks! Just needs a couple more changes before it can land.

devtools/x/src/diff_summary.rs
mimoo: I don't think we need this for an enum:

```sh
> cargo x generate-summaries out_dir bad_value
error: Invalid value for '<OUTPUT_FORMAT>': valid values: Toml, Json

> cargo x generate-summaries -h
Generate build summaries for important subsets

USAGE:
    x generate-summaries [ARGS]

FLAGS:
    -h, --help       Prints help information
    -V, --version    Prints version information

ARGS:
    <OUT_DIR>          Directory to output summaries to (default: target/summaries)
    <OUTPUT_FORMAT>    Output in text, toml or json [default: Toml]
```

sunshowers: Ahh perfect! Thanks.

